PR #27793, opened 2026-06-23
feat(front): sound notification preference + feature flag "sound-notification" in user settings

Inesvvv:
## Description
This PR adds **only** the user-settings UI and a new `sound_notification` feature flag. The mechanism that actually plays the sound on manual agent actions follows in a separate PR.

New **"Sound notification"** section in the Notifications panel of user settings:
  - enable/disable toggle
  - sound picker (11 built-in sounds: Pluck, Wood, Skeumorphic, Game, Xylophone,
    Digital Synth, Horn, Bell, Kalimba, Marimba, Tuba)
  -**Play** button to preview the selected sound before saving
  - unified Save / dirty-state with the existing notification preferences
Preference persisted via user metadata.
New `sound_notification` feature flag (stage `dust_only`); the settings section is gated behind it (`hasFeature("sound_notification")`), so it's hidden unless the workspace is flagged.

Screenshot 2026-06-23 at 15 04 42

## Tests

- **Flag on** → the "Sound notification" section appears; toggle, sound picker, and Play preview work; Save persists the preference and clears the dirty state.
- **Flag off** → the section is hidden, and the unified Save/dirty logic safely  no-ops for the sound panel (existing ref guards).

## Risk

Low since affects UI in settings only with feature flag activated.


PopDaph (commented):
As discussed this afternoon front/components/me/SoundNotificationPreferences.tsx needs to be rewritten to simplify a lot > we're mixing states and ref and the complexity is not needed 🙏🏻

Inesvvv:
I reviewed the code of SoundNotificationPreferences.tsx and simplified it down to just two refs that track the saved values of enabled and sound. The difference between the states and the refs allows us to know when to allow the user to save the changes. I think its as simple as possible now :crossed_fingers:

front/components/UserSettingsPopover.tsx
```diff
@@ -554,30 +556,57 @@ function CustomizationSection() {
...
 function NotificationsSection({ owner }: { owner: WorkspaceType }) {
   const { user } = useUser();
+  const { hasFeature } = useFeatureFlags();
   const notificationPreferencesRef =
     useRef<NotificationPreferencesRefProps>(null);
-  const [isDirty, setIsDirty] = useState(false);
+  const soundNotificationPreferencesRef =
+    useRef<SoundNotificationPreferencesRefProps>(null);
+  const [soundDirty, setSoundDirty] = useState(false);
+  const [notifDirty, setNotifDirty] = useState(false);
```
PopDaph, line 575: Why do we need to "dirty" states?

front/components/UserSettingsPopover.tsx
```diff
@@ -554,30 +556,57 @@ function CustomizationSection() {
...
-    if (notificationPreferencesRef.current) {
-      await notificationPreferencesRef.current.savePreferences();
-      setIsDirty(false);
-    }
+    await Promise.all([
+      soundNotificationPreferencesRef.current?.savePreferences(),
+      notificationPreferencesRef.current?.savePreferences(),
+    ]);
+    setSoundDirty(false);
+    setNotifDirty(false);
```
PopDaph: Here it means if the updates fails the user is stuck with a disabled saved button 🙏🏻

front/components/me/SoundNotificationPreferences.tsx
```diff
@@ -0,0 +1,217 @@
...
+  const [selectedSound, setSelectedSound] = useState<SoundNotificationType>(
+    DEFAULT_SOUND_NOTIFICATION
+  );
+
+  const originalEnabledRef = useRef(false);
+  const originalSoundRef = useRef<SoundNotificationType>(
+    DEFAULT_SOUND_NOTIFICATION
+  );
+
+  useEffect(() => {
```
PopDaph, line 56: The toggle and dropdown are clickable *before* the saved settings finish loading, and these effects then overwrite the UI state with the server value. Repro: saved sound is Marimba → UI shows the default (Pluck) for a moment → user picks Bell → saved data arrives and snaps the UI back to Marimba, throwing away their choice.

Simplest fix: don't let the controls be used until the data has loaded. `useUserMetadata` gives you `isMetadataLoading`. Disable the SliderToggle (line 138) and the dropdown (line 181) and show a loading state until both settle. Bonus: that's also a loading state we're meant to have ([REACT3]).

PopDaph (commented):
@ykmsd I think it's overly complex and I think part of this complexity could be killed if we remove all the isDirty logic to make sure we activate the Save button only of something has changes. I'd kill it cause I don't like this pattern but WDYT, 🙏🏻

ykmsd:
So I did a quick pass and I'm having difficult time to read. Usually you want to avoid using ref unless you know what you are doing, and I never personally need to use `useImperativeHandle` but do we really need it? If code can be simplified by removing "dirty" part I would definitely go for it 👍

Inesvvv:
I used the `useImperativeHandle` and the isDirty because that was the implementation used in the `NotificationPreferences.tsx` so to be consistent I thought this was better, but I'll try to simplify 👍

ykmsd:
@Inesvvv it's unfortunate but don't trust our codebase, we don't write good react code sadly. If you are not sure if you should do the same for other part of the codebase, feel free to come to me to double check!

Inesvvv:
Okay will do! Thank you :)

front/components/UserSettingsPopover.tsx
Inesvvv: I had to keep the dirty state here because the NotificationPreferences file still uses that and since the save button is shared by both.

front/components/me/SoundNotificationPreferences.tsx
```diff
@@ -0,0 +1,203 @@
...
+  SliderToggle,
+} from "@dust-tt/sparkle";
+import { zodResolver } from "@hookform/resolvers/zod";
+import { useEffect } from "react";
+import type { Control } from "react-hook-form";
+import { useController, useForm } from "react-hook-form";
+import { z } from "zod";
+
+const SoundNotificationFormSchema = z.object({
+  enabled: z.boolean(),
```
PopDaph, line 27: NIT: for next PR, since "enabled" is related to the sound, I'll rename to "isSoundEnabled" 🙏🏻

front/components/me/SoundNotificationPreferences.tsx
```diff
@@ -0,0 +1,203 @@
...
+  const {
+    metadata: enabledMetadata,
+    mutateMetadata: mutateEnabled,
+    isMetadataLoading: isEnabledLoading,
+  } = useUserMetadata(SOUND_NOTIFICATION_METADATA_KEYS.enabled);
+  const {
+    metadata: soundMetadata,
+    mutateMetadata: mutateSound,
+    isMetadataLoading: isSoundLoading,
+  } = useUserMetadata(SOUND_NOTIFICATION_METADATA_KEYS.sound);
```
PopDaph, line 44: In a new PR also, ideally we have a way to load multiple medatada at once, and save them all at once -> not blocking for this PR

front/components/me/SoundNotificationPreferences.tsx
```diff
@@ -0,0 +1,203 @@
...
+  const handlePlay = () => {
+    const audio = new Audio(
+      `/sounds/${encodeURIComponent(soundField.value)}.mp3`
+    );
+    void audio.play();
+  };
+
+  return (
+    <div className="rounded-xl border border-border dark:border-border-night">
+      <div className="flex items-center p-4">
```
ykmsd, line 124: I think you are repeating the same code here and below, I would probably create another component (something like SectionLayout etc) and reuse it for both, so if we ever want to update the style you only need to update it once place 😄

front/components/UserSettingsPopover.tsx
```diff
@@ -564,43 +568,73 @@ function CustomizationSection() {
...
-          disabled={!isDirty}
-        />
-      }
+      description="Manage the notifications of your Dust workspace"
+      footer={saveButton}
     >
+      {hasFeature("sound_notification") && (
+        <SoundNotificationPreferences
+          control={sound.control}
+          disabled={sound.isLoading}
```
PopDaph, line 632: The two render branches disagree here: the no-subscriberHash branch passes disabled={sound.isLoading || isSubmitting}, but this one only passes disabled={sound.isLoading}.

That leaves the sound toggle/dropdown interactive during a save. If the user flips the toggle while the request is in flight, handleSubmit has already captured data, so on success form.reset(data) resets to the pre-toggle value and the user's mid-save change is silently dropped (and the panel looks clean).

Can we align it with the other branch?

          `disabled={sound.isLoading || isSubmitting}`


(again, not blocking)

PopDaph (approved):

front/components/me/SoundNotificationPreferences.tsx
```diff
@@ -0,0 +1,203 @@
...
+        <div className="flex flex-1 flex-col gap-1">
+          <span className="text-sm font-semibold text-foreground dark:text-foreground-night">
+            Manual actions sound notification
+          </span>
+          <span className="text-sm text-muted-foreground dark:text-muted-foreground-night">
+            Play a sound when a manual action is required
```
ykmsd, line 130: `when a manual action is required` is kind of sad? Can be something like `when an agent needs your attention`?

ykmsd: i will review copy with @ClementAupiais

ykmsd (commented):
👍

PR #27794, opened 2026-06-23
feat(scripts): add eu/us cluster and prodbox connect helpers

ElPicador:
## Description

Two improvements to the migration and cluster tooling:

1. **Post-deploy confirmation gate** (`scripts/run-migrations.sh`): when running `post-deploy` migrations, the script now pauses and asks the operator to confirm that `front` and `front-sse` are deployed with the latest code before proceeding. Post-deploy migrations can reference DB models that the new code depends on, so running them against old pods is unsafe.

2. **Cluster/prodbox connect helpers** (`scripts/connect-cluster.sh`, `scripts/connect-prodbox.sh`, `scripts/lib/gcp.sh`): shared GCP helpers extracted into `lib/gcp.sh` and two convenience scripts for pointing `kubectl` at a Dust production cluster (`eu`/`us`) or dropping into a prodbox shell — without mutating the caller's active kubectl context.

## Tests

Manually verified the confirmation prompt blocks when answering `N` and proceeds when answering `y`.

### DIFF
--- a/scripts/lib/gcp.sh
+++ b/scripts/lib/gcp.sh
@@ -1,0 +1,88 @@
+# shellcheck shell=bash
+#
+# Shared helpers for connecting to Dust production GKE clusters.
+#
+# Source this file from a script; do not execute it directly. All log/status
+# output goes to stderr so that helpers which "return" a value via stdout
+# (e.g. gcp_region_for_alias, get_prodbox_pod) stay safe inside $(...).
+
+# Map the short region alias used on the command line to the GCP region, which
+# is also the name of the gcloud configuration to use. Keep in sync with the
+# REGIONS list in run-migrations.sh as new regions are added.
+gcp_region_for_alias() {
+  local alias="$1"
+  case "$alias" in
+    us | us-central1) echo "us-central1" ;;
+    eu | europe-west1) echo "europe-west1" ;;
+    *)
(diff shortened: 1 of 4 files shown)


ElPicador:

- #27794 ◀

scripts/lib/gcp.sh
```diff
@@ -0,0 +1,88 @@
...
+}
+
+# Fetch cluster credentials for a region into the given KUBECONFIG path.
+# --configuration uses the region-named gcloud config without activating it
+# globally, keeping the caller's active configuration intact.
+connect_cluster() {
+  local region="$1"
+  local kubeconfig="$2"
+
+  echo "🌍 ${region} — fetching cluster credentials..." >&2
```
flvndvd, line 56: This feels somewhat a duplicate of what we have in the init infra script [here](https://github.com/dust-tt/dust-infra/blob/main/scripts/setup_infra.sh). Can we keep infra logic in the infra repo 🙏?

ElPicador: We need this for the `run-migration.sh` script 

So either we:
* accept the duplication (this PR)
* create a repo/cli to share this (directionality what we want, but longer). 

If it's ok for you, I'll merge this as is, and we can clean this later

Fraggle (approved):
LGTM

PR #27797, opened 2026-06-23
feat(llm): add xAI provider scaffolding to the new LLM router

adrsimon:
## Description

First of two PRs adding Grok support to the new `model_constructors` LLM router. This one lays the provider scaffolding only; model-config mixins, endpoint classes, tests, and registration follow in PR2.

xAI is OpenAI-Responses-compatible (`client.responses.create`, base URL `https://api.x.ai/v1`), so the new client is a thin clone of `OpenAIResponsesStream` that reuses the shared `sdk/openai_responses` converters — no new converter code.

What's included:
- **Type registrations** — `XAI_PROVIDER_ID` (`xai`), `XAI_API` (`xai`, mirroring the fireworks/togetherai pattern), `XAI_API_KEY` credential, and the three model ids: `grok-4-latest`, `grok-4-1-fast-reasoning-latest`, `grok-4-1-fast-non-reasoning-latest`.
- **`providers/xai/reasoning_efforts.ts`** — `XAI_SUPPORTED_REASONING_EFFORTS = none/low/medium/high`, kept ISO with the legacy router's effort mapping (`none→none, light→low, medium→medium, high→high`).
- **`stream/clients/xai.ts`** — `XaiStream` abstract client (apiKey + baseURL only, matching legacy — no extra OpenAI-client knobs).

Kept 100% ISO with the legacy `XaiLLM` on every constant: base URL, client options, credential name, provider id, Responses-API surface, reasoning-effort set, and model-id strings.

## Tests

No runtime behavior is wired yet (no endpoints registered), so there's nothing to exercise at the API level in this PR. `npx tsgo --noEmit` is clean and biome passes (both also enforced by the pre-commit hook). Live-API integration tests for each endpoint land in PR2 via the model_constructors test harness.

## Risk

Very low / inert. This PR only adds new type entries and an abstract client that nothing instantiates yet — no endpoint is added to `STREAM_ENDPOINTS`, so the router's behavior is unchanged. Trivially rollback-able.

## Deploy Plan

Standard deploy. No migrations, no config, no feature-flag changes. Safe to ship independently of PR2.

### DIFF
--- a/front/lib/model_constructors/stream/clients/xai.ts
+++ b/front/lib/model_constructors/stream/clients/xai.ts
@@ -1,0 +1,72 @@
+import { XAI_SUPPORTED_REASONING_EFFORTS } from "@app/lib/model_constructors/providers/xai/reasoning_efforts";
+import { WithOpenAIResponsesInputConverter } from "@app/lib/model_constructors/sdk/openai_responses/converters/input";
+import { WithOpenAIResponsesOutputConverter } from "@app/lib/model_constructors/sdk/openai_responses/converters/output";
+import { rawOutputToEvents } from "@app/lib/model_constructors/sdk/openai_responses/converters/output/utils";
+import { StreamEndpoint } from "@app/lib/model_constructors/stream/endpoint";
+import type { Credentials } from "@app/lib/model_constructors/types/credentials";
+import { inputConfigSchema } from "@app/lib/model_constructors/types/input/configuration";
+import type { ModelResponseEvent } from "@app/lib/model_constructors/types/output/events";
+import { XAI_API } from "@app/lib/model_constructors/types/provider_apis";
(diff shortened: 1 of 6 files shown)


front/lib/model_constructors/stream/clients/xai.ts
```diff
@@ -0,0 +1,72 @@
...
+
+const configSchema = inputConfigSchema.extend({
+  reasoning: z
+    .object({
+      effort: z.enum(XAI_SUPPORTED_REASONING_EFFORTS),
+    })
+    .optional(),
+});
+
+type XaiInputConfig = z.infer<typeof configSchema>;
```
davidebbo, line 30: nit: this reverse casing of `xAI` bothers me a bit, but feel free to ignore :)

davidebbo (approved):
LGTM

PR #27800, opened 2026-06-23
Reduce excel tool output size

Chrisjow:
## Description

Reduces the output size of the Microsoft Excel `read_worksheet` tool.

The tool now returns worksheet data as CSV instead of the full Microsoft Graph JSON response, which makes outputs much more compact and easier for agents to consume. It also adds a 25,000-cell safety limit to avoid overly large tool results.

When no range is provided, the tool first checks the worksheet used range size. If the used range is within the limit, it reads it; otherwise it returns an error asking the caller to provide a smaller range. 
When a range is provided, the tool validates that it is a bounded A1 range and rejects ranges above the same limit.

## Tests

Tested the tool manually.

## Risk

Low risk. The main behavior change is the response format for `read_worksheet`: callers now receive CSV text instead of the raw Graph API JSON payload. This should reduce context usage and improve agent usability, but any code relying on the previous JSON shape may need to be updated.

The 25,000-cell limit may also reject very large worksheets that were previously returned in full.

## Deploy Plan

Deploy front

### DIFF
--- a/front/lib/api/actions/servers/microsoft_excel/metadata.ts
+++ b/front/lib/api/actions/servers/microsoft_excel/metadata.ts
@@ -50,7 +50,7 @@ export const MICROSOFT_EXCEL_TOOLS_METADATA = createToolsRecord({
   },
   read_worksheet: {
     description:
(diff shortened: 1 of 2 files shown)


ykmsd (approved):
ty!

PR #27801, opened 2026-06-23
perf(analytics): source members table consumed credits from the analytics index

rfrenoy:
## Description

The members usage table read per-user consumed AWU credits from Metronome at page load, which required a usage scan that scaled with the number of members and made the table slow to load. Source the consumed value from the analytics Elasticsearch index instead (sum of cost.full_awu over the current billing cycle), in a single aggregation per page. Seat data and spend limits stay on their existing workspace-level cached Metronome calls.

Limit: we can no longer hide free seat consumed credits when user upgraded to a paid seat on the current billing cycle. It can be addressed in a follow up PR by storing the timestamp when a user upgrade to a paid plan and compute credit from `max(billing cycle start, migration to paid plan)`

## Tests

Local

## Risk

Low

## Deploy Plan

Front

### DIFF
--- a/front/lib/api/credits/members_usage.ts
+++ b/front/lib/api/credits/members_usage.ts
@@ -1,3 +1,4 @@
+import { bucketsToArray, searchAnalytics } from "@app/lib/api/elasticsearch";
 import type { Authenticator } from "@app/lib/auth";
 import { listPerUserCreditBalanceAlertsForWorkspace } from "@app/lib/metronome/alerts/per_user_credit_balance";
 import type {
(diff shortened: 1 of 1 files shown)


tdraier (approved):

rfrenoy:
@tdraier what about merging this, closing https://github.com/dust-tt/dust/pull/27828 (without merging) and accepting that it's wrong on free seat usage on the immediate short term. And we shortly add a tag on ES so that in the future we don't count free seat usage once the user is on a paid seat? seems better fitted than relying on the DB + logic

tdraier (approved):

tdraier:
agree, we'll fix the usage of free seats afters

PR #27802, opened 2026-06-23
[marketing] fix: update sonnet model ID in academy chat

id13:
## Description

A bug prior to marketing code migration what that we hardcoded the model-id in academy/chat endpoint.
When that model-id stopped to work, so did the endpoint.
Now we use a constant from llm-router's that will make typing catch any regression (what we should have done in the first place tbh)

Also we were swallowing errors, we now throw them, 🫶 oncall eng.

## Tests

- Tested locally

## Risk

Low

## Deploy Plan

- Deploy front & marketing


id13:
PMRR

davidebbo (approved):
LGTM

PR #27754, opened 2026-06-22
[Skills] Restrict enable_skill to available skills

philipperolet:
## Description

Fixes https://github.com/dust-tt/tasks/issues/8989

Context: [Slack thread](https://dust4ai.slack.com/archives/C0BA8R2CJ3U/p1781720554284579?thread_ts=1781720207.115419&cid=C0BA8R2CJ3U)

Follow-up: https://github.com/dust-tt/dust/pull/27807 splits out the adjacent discoverable-skill agent-loop-context fix.

Agents could call `enable_skill` with any active workspace skill name because the handler resolved skills with a workspace-wide active-name lookup. The prompt only lists skills available in the current agent loop, but execution did not enforce that boundary.

This changes `enable_skill` to resolve the requested skill from the current agent-loop allow-list instead: already enabled skills, equipped/available skills, active skills explicitly tagged in the current user message, and skills referenced one level down by active/available roots. Those roots include system, enabled, and equipped skills; enabled skills are included because the current API returns them separately from `equippedSkills`.

The referenced-skill expansion is intentionally not transitive. If Skill A is available and references Skill B, Skill B can be enabled. Skill C, referenced only by Skill B, cannot be enabled until Skill B has first been enabled; once Skill B is enabled, its direct references become valid one-hop enable targets. This supports nested skill authoring without re-opening arbitrary workspace-wide enablement. References normalized to `<unavailable_skill>` are not part of the enable allow-list.

The fix also threads the current `userMessage` into tool execution and makes it required in the agent-loop tool run context. The current user message is required to support explicit `<skill>` tags in the message and to apply run-specific code-defined skill restrictions when resolving user-tagged or referenced skills by id/name.

## Risks

Blast radius: agent skill activation through the `skill_management` MCP tool.
Risk: low

## Deploy Plan
- deploy front

## Test

- [x] `NODE_ENV=test npm test lib/api/actions/servers/skill_management/tools/index.test.ts lib/actions/mcp_actions.test.ts`
- [x] `npx tsgo --noEmit`

PR #27807, opened 2026-06-23
[Skills] Filter discoverable skills with run context

philipperolet:
## Description

Follows merged https://github.com/dust-tt/dust/pull/27754.

When an agent has `discover_skills`, `SkillResource.listForAgentLoop()` adds discoverable skills to the agent loop. That lookup did not receive the current agent-loop context, so a built-in skill defined in code could still be exposed as discoverable even when it is disabled for the current message origin.

This passes `agentLoopData` to `listDiscoverable()` from `listForAgentLoop()`. For example, `mention_users` is disabled for Slack, Slack Workflow, and Teams messages based on `userMessage.context.origin`; after this change it is not included as a discoverable skill in those runs.

## Risks

Blast radius: discoverable skills included in agent loop availability.  
Risk: low

## Deploy Plan

- deploy front

## Test

- [x] `NODE_ENV=test npm test lib/api/assistant/jit_actions.test.ts`


aubin-tchoi (approved):
LGTM

PR #27811, opened 2026-06-23
[front] Extend advanced model access to credit-priced plans

thervier:
## Description

Until now, access to some advanced models was based on feature flags or if the plan was enterprise (based on availableIfOneOf in model declaration that could define "enterprise" or "featureFlag")
=> turn "enterprise" into "plansWithAvancedModels" and check this against a function isPlanForAvancedModels that would include CP plans

* Renames the enterprise field in availableIfOneOf to plansWithAvancedModels for clarity — it now reflects that access is tied to plan type, not just enterprise status.
* Introduces isPlanForAvancedModels in assistant.ts, which extends the previous isEnterpriseOrDust check to also include credit-priced plans (CP_*), giving those workspaces access to Opus models.
*  Moves isEnterpriseOrDust (the narrower enterprise+dust check, still used by the LLM router) to plan_codes.ts alongside the other plan predicates.

## Tests

Tested locally

## Risk

Low, should be only additive for CP plans

## Deploy Plan

Deploy front

### DIFF
--- a/front/lib/assistant.ts
+++ b/front/lib/assistant.ts
@@ -1,4 +1,5 @@
 import {
+  isCreditPricedPlanPrefix,
   isDustCompanyPlan,
   isEnterprisePlanPrefix,
   isUpgraded,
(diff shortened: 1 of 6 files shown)


front/lib/assistant.ts
```diff
@@ -12,10 +13,12 @@ import type { PlanType } from "@app/types/plan";
...
 import type { WhitelistableFeature } from "@app/types/shared/feature_flags";
 
-export function isEnterpriseOrDust(plan: PlanType | null): boolean {
+export function isPlanForAdvancedModels(plan: PlanType | null): boolean {
   return (
     plan !== null &&
-    (isEnterprisePlanPrefix(plan.code) || isDustCompanyPlan(plan.code))
+    (isEnterprisePlanPrefix(plan.code) ||
+      isCreditPricedPlanPrefix(plan.code) ||
+      isDustCompanyPlan(plan.code))
```
tdraier, line 21: shouldn't we have a new boolean in the plan instead of hardcoding plan names ?

tdraier (approved):

PR #27819, opened 2026-06-23
[Builder] Gate admin editing on skills

philipperolet:
## Description
Fixes https://github.com/dust-tt/tasks/issues/9066 (1 on 2)

Adds a reusable builder editor gate message and applies it to existing skill editing. Workspace admins who are not skill editors can view the skill builder, but mutation controls stay disabled while editor membership is loading or while they remain non-editors. The Add me CTA adds the current user as a skill editor and waits for the editors SWR refresh before unlocking.

## Risks

Blast radius: skill builder edit screens for workspace admins who are not skill editors.
Risk: standard

## Deploy Plan
- pmrr
- deploy front

### DIFF
--- a/front/components/shared/BuilderEditorGateMessage.tsx
+++ b/front/components/shared/BuilderEditorGateMessage.tsx
@@ -4,6 +4,7 @@ import {
   ContentMessageAction,
   InfoCircle,
   RefreshCw02,
+  UsersPlus,
 } from "@dust-tt/sparkle";
 
 type BuilderType = "agent" | "skill";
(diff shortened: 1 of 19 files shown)


philipperolet:
cc @pinotalexandre

pinotalexandre:
tyty!

pinotalexandre (commented):
@philipperolet LGTM for this part, but it's missing the part where the fields are grayed out before switching to edit mode (basically setting the fields and button to disabled), or did you plan to do that in a second part?

philipperolet:
Hey, yes it was adding  too much code for an edge case
I went with the faster solution of preventing save in that case (but fields that are not "save-dependent" e.g. suggestions / file uploads are grayed or removed

PR #27821, opened 2026-06-23
Fix agents resolving to models that are unavailable in the workspace region

flvndvd:
## Description

This PR fixes https://github.com/dust-tt/tasks/issues/9079.

Some workspaces restricted to regionally hosted models could not start a conversation with certain agents and instead hit a generic "the model is not supported" error. The root cause was a disagreement between how a model is chosen and how it is later validated. Selection only checked that the model's provider was enabled for the workspace, whereas the check that runs when a message is posted also enforces plan and, crucially, region constraints. An agent could therefore resolve to a model that is not offered in the workspace region, and the request was then refused downstream.

The fix routes all agent model selection through a single chokepoint that asks the same complete question the downstream validation asks: is this model both provider-enabled and actually available for this workspace given its plan and region. Selection now walks its ordered list of preferred models and returns the first one that fully satisfies those constraints, transparently falling through to the next best option when a preferred model is not offered in the region. The previously scattered selection points that each only checked provider eligibility now all defer to this one place, so the guarantee is enforced uniformly instead of being re-derived, and occasionally forgotten, at every site. That is what makes the patch systemic rather than a one-off fix for the agent where the bug surfaced.

### DIFF
--- a/front/lib/api/assistant/global_agents/configurations/dust/deep-dive.ts
+++ b/front/lib/api/assistant/global_agents/configurations/dust/deep-dive.ts
@@ -43,6 +43,7 @@ import {
   type ModelProviderIdType,
   type ReasoningEffort,
 } from "@app/types/assistant/models/types";
+import type { WhitelistableFeature } from "@app/types/shared/feature_flags";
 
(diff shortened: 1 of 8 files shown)


spolu (approved):
LGTM

PR #27531, opened 2026-06-17
Render scoped file preview directives

fontanierh:
## Description

Fixes the rendering half of dust-tt/tasks#8891.

Adds a `:preview_file{...}` markdown directive for scoped conversation and Pod file paths. Agent markdown renders it as a previewable file citation with the file name, file type, and file icon. Clicking the citation opens the existing file preview dialog, which already exposes the download action.

Sparkle also hides incomplete trailing text directives during streaming so users do not see partial directive syntax before it turns into the component.

The follow-up PR wires file tools to emit the directive.

## Tests

Local checks:
- `source /Users/henryfontanier/.dust-hive/envs/pr27542-dl-124434/env.sh && NODE_ENV=test npm run test -- lib/markdown/file_preview.test.ts components/assistant/AgentMessageMarkdown.integration.test.tsx`
- `npm --prefix front run tsgo`
- `npm --prefix sparkle run build:esm`

## Risk

Low. Malformed directives are ignored and valid directives use the existing authenticated scoped file preview/download flow.

## Deploy Plan

Standard deploy.

PR #27823, opened 2026-06-23
Render file previews as inline links

fontanierh:
## Description

Follow-up to #27531. The first renderer used the existing file citation card, which made an inline file reference look like a large attachment card instead of a hyperlink.

Render `:preview_file` as an inline link-style citation. Normal click still opens the existing preview dialog, and the link carries the preview URL so open-in-new-tab behavior has a real target.

This PR is intentionally between #27531 and #27542. #27542 should be rebased onto this branch.

## Tests

- `NODE_ENV=test npm run test -- components/assistant/AgentMessageMarkdown.integration.test.tsx`
- `npx biome check components/assistant/conversation/FilePreviewContext.tsx components/markdown/FilePreviewBlock.tsx components/assistant/AgentMessageMarkdown.integration.test.tsx`
- `npm --prefix front run tsgo`

## Risk

Low. This only changes the presentation of `:preview_file` citations and keeps the preview dialog path intact.

## Deploy Plan

Standard deploy.

### DIFF
--- a/front/components/assistant/AgentMessageMarkdown.integration.test.tsx
+++ b/front/components/assistant/AgentMessageMarkdown.integration.test.tsx
@@ -123,11 +123,21 @@ describe("AgentMessageMarkdown - Integration Tests", () => {
         </FilePreviewProvider>
       );
 
-      expect(screen.getByText("booklet.pdf")).toBeInTheDocument();
+      const link = screen.getByRole("link", {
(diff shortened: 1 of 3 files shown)


fontanierh:
Closing this because the inline preview rendering landed on main via #27809, so PR #27542 no longer needs this intermediary branch.

PR #27828, opened 2026-06-23
fix(members): reset consumed credits on free to paid seat upgrade

rfrenoy:
## Description

The members table sourced consumed credits from the analytics index keyed by sId for everyone. Metronome instead records free-seat usage under a separate user id (free-<sId>), so a free→paid upgrade resets the user's billed consumption to 0 — behavior the ES-backed table lost. Paid→paid changes (e.g. pro→max) keep the same id and must not reset.

Add a freeSeatExitedAt timestamp stamped once when a user leaves the one-shot free seat (centralized in updateMembershipSeat, so all upgrade paths are covered; never set on paid→paid). The members table counts a free→paid upgrader's consumed credits only from that timestamp (clamped into the cycle) via a keyed filters aggregation, while everyone else is summed from cycle start in the existing terms aggregation. Prior-cycle free usage is already outside the window, so no backfill is needed.
## Tests

local

## Risk

low
## Deploy Plan

merge, pre-deploy mig, deploy

### DIFF
--- a/front/lib/api/credits/members_usage.ts
+++ b/front/lib/api/credits/members_usage.ts
@@ -183,25 +183,50 @@ type ConsumedCreditsAggs = {
   by_user?: estypes.AggregationsMultiBucketAggregateBase<ConsumedCreditsBucket>;
 };
 
+type ConsumedCreditsFilterBucket = {
+  credits?: estypes.AggregationsSumAggregate;
+};
+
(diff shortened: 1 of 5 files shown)


spolu:
👍 on data-model given simple change.

rfrenoy:
Closing, after more thinking I'd rather index the free seat info in ES and avoid querying the db every time we want to display usage.

PR #27829, opened 2026-06-23
Skip reconcile on non cp plan

tdraier:
## Description

Two optimizations for the `[UserCreditStateMachine] State reconciled` log noise:

**1. Skip reconcile entirely for non-CP plans**

`reconcileWorkspaceUserCreditStates` was being triggered for all Metronome workspaces (every workspace has a shadow contract), including those not on credit-priced plans where per-user credit state is meaningless. It now takes a `planCode` parameter and returns immediately if the plan isn't a `CP_*` plan — short-circuiting all Metronome API calls (seat balances, per-user usage) and the membership loop.

**2. Only log when state actually changes**

`setUserCreditStateReconciled` was logging at `info` level for every member on every reconcile, even when no-op (same state, only Redis cache refreshed). Now only logs when `fromState !== toState`.

Together these eliminate the log flood for non-CP workspaces entirely, and reduce it to meaningful state-change events for CP workspaces.

## Tests

Updated existing tests in `reconcile_credit_state.test.ts` to pass the new required `planCode` field (`CP_BUSINESS_PLAN`). No behavior changes for CP plans.

## Risk

Low. Non-CP workspaces are cleanly skipped before any Metronome API calls. The CP guard uses the same `isCreditPricedPlanPrefix` (`CP_` prefix check) used elsewhere. The log reduction for CP workspaces is purely observational — the Redis cache refresh still happens unconditionally.

## Deploy Plan

### DIFF
--- a/front/lib/api/metronome/process_webhook.ts
+++ b/front/lib/api/metronome/process_webhook.ts
@@ -1428,14 +1428,15 @@ export async function processMetronomeWebhook({
       // onto a business plan) leaves users stuck in their previous state.
       // Mirrors the pool reconcile above; pass the new contract id directly
       // since the subscription swap below may not have happened yet.
+      const targetPlanCode =
+        contractResult.value.custom_fields?.[PLAN_CODE_CUSTOM_FIELD_KEY];
+
       await reconcileWorkspaceUserCreditStates({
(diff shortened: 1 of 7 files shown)


tdraier:

- #27829 ◀

rfrenoy (approved):

PR #27830, opened 2026-06-23
refactor(front): move reclassified-internal DTO types out of lib/api into types/api

davidebbo:
Part of the migration to move HTTP request/response DTO types and zod schemas out of `front/lib/api` and into `front/types/api/*`. Design doc: https://app.notion.com/p/38128599d94180848990e566ae472970

These 9 files were initially deferred as "public" because v1 routes import from them — but symbol-level analysis showed the v1 routes import only their **functions** (which stay in lib/api), never their DTO *types*. The DTO types are internal (the public counterparts already live separately in `@dust-tt/client`), so they get the normal `types/api` treatment.

- Files: `apps`, `assistant/user_relation`, `dust_app_secrets`, `files/upload_metadata`, `provider_credentials`, `tables`, `user`, `data_sources`, `data_source_view`.
- All business functions stay in lib/api; v1 routes importing them are untouched.
- 38 consumers updated; `lib/api/run.ts` re-export shim repointed to `types/api/apps`.

**Partial:** `data_source_view.ts` keeps `GetContentNodesOrChildrenRequestBody` (references `SortingParamsCodec`, an io-ts codec in `lib/api/pagination` — a real `types → lib` gotcha) for a later pass.

Verified: `tsgo --noEmit` clean in both `front` and `front-api`.

### DIFF
--- a/extension/shared/tools/takeScreenshotOrAttachFileTool.ts
+++ b/extension/shared/tools/takeScreenshotOrAttachFileTool.ts
@@ -1,7 +1,7 @@
 import { MCPError } from "@app/lib/actions/mcp_errors";
 import type { ToolHandlerResult } from "@app/lib/actions/mcp_internal_actions/tool_definition";
-import type { FileUploadRequestResponseBody } from "@app/lib/api/files/upload_metadata";
 import { clientFetch } from "@app/lib/egress/client";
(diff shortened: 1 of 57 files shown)


matteotrab (commented):
LGTM

PR #27834, opened 2026-06-23
Fix: do not completely ignore the AGENTS.md file if it's too big.

Fraggle:
## Description

Two bugs with oversized `AGENTS.md` files:

- **TipTap was silently trimming content** when `CharacterCount.configure({ limit })` was set — loading a file above the limit caused TipTap to cut the content at load time, so the editor appeared to show (and potentially save) a truncated version. Fixed by dropping the `limit` from `CharacterCount` so it counts only, and enforcing the cap explicitly in `MarkdownFileEditor`'s save handler instead. A warning banner (`ContentMessage`) now surfaces when content exceeds the limit, and saving is blocked until the file is trimmed.
- **`POD_AGENTS_MD_MAX_CHARACTER_COUNT` was too low** (4096 → 8192), causing valid AGENTS.md files to be ignored in prompt injection via `readPodAgentsMdContent`.

## Tests

Tested locally: opened an oversized AGENTS.md in Pod settings — warning banner appears, save is blocked with an error notification. Trimming below 8192 chars restores normal save behavior.

## Risk

Low. Changes are confined to the `MarkdownFileEditor` component and the shared `POD_AGENTS_MD_MAX_CHARACTER_COUNT` constant. No data is deleted or migrated; the limit increase only allows more content through, and the editor-side enforcement is additive (warning + blocked save).

## Deploy Plan

Deploy `front` only.

### DIFF
--- a/front/components/editor/MarkdownFileEditor.tsx
+++ b/front/components/editor/MarkdownFileEditor.tsx
@@ -2,6 +2,7 @@ import {
   MarkdownEditor,
   type MarkdownEditorProps,
 } from "@app/components/editor/MarkdownEditor";
+import { useSendNotification } from "@app/hooks/useNotification";
 import {
   getFilePathContentApiPath,
   useFileContentByUrl,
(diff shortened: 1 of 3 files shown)


Fraggle:

- #27834 ◀

ElPicador (commented):
👍

PR #27835, opened 2026-06-23
feat(llm): restore flash-lite `none` reasoning effort in the new router

adrsimon:
## Description

Gemini 3 has no `ThinkingLevel.NONE`/`OFF` member, so the new `google_genai` converter had no way to express the legacy "minimal thinking, hidden thoughts" floor that the old router applied for the `none` reasoning effort.

This change:
- Maps `none` → `{ thinkingBudget: 128, includeThoughts: false }` in `effortToThinkingConfig` (replacing the previous `effortToThinkingLevel`, which could only return a level). This matches the legacy `POST_GEMINI_3_THINKING_CONFIG_MAPPING.none` exactly.
- Re-adds `none` as a supported reasoning effort for Gemini 3.1 Flash-Lite (`GEMINI_FLASH_LITE_SUPPORTED_REASONING_EFFORTS` = none/minimal/low/medium/high). Flash-3.5 and Pro are unchanged.

Net effect: the existing `google_ai_studio/global/gemini-3.1-flash-lite` endpoint reaches parity with the legacy router on the `none` effort. Behaviour for every other effort is byte-identical to before.

This is the first of a 3-PR stack adding Gemini-on-Vertex to the new router; it is independent and shippable on its own.

## Tests

`dust-llm-class-tdd` live integration suite against the real API:
- Flash-Lite `r-none` cases now run with default checkers (previously asserted `INPUT_CONFIGURATION_ERROR`); `reasoning/.../r-none` → `HAS_NO_REASONING` confirms `includeThoughts: false` surfaces no thoughts. 7/7 green.
- Full `google_ai_studio_global_gemini_3_1_flash_lite` suite green.
- `tsgo --noEmit` clean; biome clean (pre-commit).

## Risk

Low. Converter change is behaviour-preserving for all non-`none` efforts (same `ThinkingConfig` object). `none` is additive for flash-lite. Easy to roll back.

## Deploy Plan

Standard. No migration, no flag.

### DIFF
--- a/front/lib/model_constructors/providers/google_ai_studio/inputConfig.ts
+++ b/front/lib/model_constructors/providers/google_ai_studio/inputConfig.ts
@@ -1,4 +1,4 @@
-import { GEMINI_SUPPORTED_REASONING_EFFORTS } from "@app/lib/model_constructors/providers/google_ai_studio/reasoning_efforts";
+import { GEMINI_FLASH_LITE_SUPPORTED_REASONING_EFFORTS } from "@app/lib/model_constructors/providers/google_ai_studio/reasoning_efforts";
 import {
   inputConfigSchema,
   temperatureSchema,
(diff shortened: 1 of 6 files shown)


davidebbo (approved):
LGTM

PR #27836, opened 2026-06-23
feat(llm): add Gemini-on-Vertex stream client and endpoints

adrsimon:
## Description

Adds the Vertex AI (agent-platform) transport for Gemini to the new `model_constructors` router, for parity with the legacy `GoogleLLM` `useVertex` path.

- New client `stream/clients/agent_platform_google.ts` (`AgentPlatformGoogleStream`): constructs `new GoogleGenAI({ vertexai: true, project, location: "global" })`, `providerId = google_ai_studio`, `api = agent-platform`. Reuses the existing `google_genai` input/output converters and the `WithGoogleAiStudioGemini*Config` model mixins unchanged — the Vertex request contract is identical to AI Studio.
- Three endpoint classes `stream/endpoints/agent_platform_eu_gemini_{3_1_flash_lite,3_1_pro,3_5_flash}.ts` with per-endpoint pricing and `region: "eu"`.
- Three live integration tests mirroring the AI Studio sibling key sets.

**Why `location: "global"` and not `"eu"`/`"europe-west1"`:** the native `@google/genai` `:streamGenerateContent` API does not serve the `eu` multi-region alias (returns a generic 404), unlike Anthropic's `:rawPredict` partner API which does — that's why Anthropic-on-Vertex can use `region: "eu"` but Gemini can't. Gemini 3.x preview models are also not published to `europe-west1`. The legacy router passes no location, which `@google/genai` v1.43 defaults to `"global"` — so `"global"` is exact legacy parity. `region: "eu"` is kept as the Dust-facing residency label (matches legacy's hardcoded `inferenceRegion: "eu"`).

Stacked on #27835. **Not wired into the registries here** — registration lands in the follow-up so each PR typechecks independently (`DUST_STREAM_ENDPOINTS` is exhaustive over `StreamEndpointId`, so the two registrations must land atomically).

> ⚠️ **Review callout — pricing:** `tokenPricing` on the three endpoints currently mirrors the AI Studio sibling values. Vertex EU prices may differ; please confirm against the Vertex pricing page before launch.

## Tests

`dust-llm-class-tdd` live suite against Vertex (requires `gcloud` ADC + `VERTEX_AI_PROJECT_ID`):
- flash-lite 40/40, pro 40/40, 3.5-flash 40/40 (one transient 429 on a full flash-lite run, passes on retry — known project quota gap, not a code issue).
- `tsgo --noEmit` clean **in isolation** (verified on this branch tip without the follow-up's registrations); biome clean.

## Risk

Low and inert: the endpoints are not registered in this PR, so nothing routes to them yet. New client is additive.

## Deploy Plan

Standard. The endpoints become reachable only after the follow-up registration PR.

### DIFF
--- a/front/lib/model_constructors/stream/clients/agent_platform_google.ts
+++ b/front/lib/model_constructors/stream/clients/agent_platform_google.ts
@@ -1,0 +1,76 @@
+import type { BaseEndpointConfiguration } from "@app/lib/model_constructors/configuration";
+import {
+  type GoogleAiStudioInputConfig,
+  googleAiStudioConfigSchema,
+} from "@app/lib/model_constructors/providers/google_ai_studio/inputConfig";
+import { WithGoogleGenAIInputConverter } from "@app/lib/model_constructors/sdk/google_genai/converters/input";
+import { WithGoogleGenAIOutputConverter } from "@app/lib/model_constructors/sdk/google_genai/converters/output";
+import { rawOutputToEvents } from "@app/lib/model_constructors/sdk/google_genai/converters/output/utils";
+import { StreamEndpoint } from "@app/lib/model_constructors/stream/endpoint";
+import type { Credentials } from "@app/lib/model_constructors/types/credentials";
+import type { ModelResponseEvent } from "@app/lib/model_constructors/types/output/events";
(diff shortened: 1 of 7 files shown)


davidebbo (approved):
LGTM

PR #27837, opened 2026-06-23
feat(llm): register Gemini-on-Vertex endpoints in the LLM routers

adrsimon:
## Description

Activates the Gemini-on-Vertex endpoints added in #27836 by wiring them into both routers:
- Registers the three `agent-platform/eu` Gemini endpoints in `STREAM_ENDPOINTS` (`model_constructors`).
- Adds the three Dust routing wrappers `llms/stream/endpoints/agent_platform_eu_gemini_*.ts` and registers them in `DUST_STREAM_ENDPOINTS`.

Both registrations land together by necessity: `DUST_STREAM_ENDPOINTS satisfies Record<StreamEndpointId, ...>` is exhaustive, so adding a `STREAM_ENDPOINTS` key without its Dust wrapper would fail to typecheck.

Stacked on #27836 (← #27835).

## Tests

- Full live suites green via the endpoint classes (see #27836).
- `tsgo --noEmit` clean on this tip (full state, both registries populated); biome clean.

## Risk

Medium — this is the change that makes traffic actually route to Vertex for matching workspaces. Mitigated by the `endpointFilter` gate (flag / credit-priced only) and easy rollback (revert the registration). Validate the gate audience (callout above) before merge.

## Deploy Plan

Merge after #27835 and #27836. Roll out by enabling `use_vertex_for_supported_models` for the target workspaces; monitor Vertex EU error/latency before broadening.


davidebbo (approved):
LGTM

PR #27843, opened 2026-06-23
feat(analytics): index is_free_seat to reset consumed credits on free to paid upgrade

rfrenoy:
## Description

The members usage table read per-user consumed AWU credits from Metronome at
page load, which required a usage scan that scaled with the number of members
and made the table slow to load. Source the consumed value from the analytics
Elasticsearch index instead (sum of cost.full_awu over the current billing
cycle), in a single aggregation per page. Seat data and spend limits stay on
their existing workspace-level cached Metronome calls.

## Tests

Unit + local

## Risk

Low
## Deploy Plan

Run ES migration (`npx tsx scripts/execute_elasticsearch_http.ts --file lib/analytics/migrations/20260623_add_is_free_seat_to_agent_message_analytics_2.http --execute
`), then deploy

### DIFF
--- a/front/lib/api/credits/members_usage.ts
+++ b/front/lib/api/credits/members_usage.ts
@@ -184,25 +184,38 @@ type ConsumedCreditsBucket = {
   credits?: estypes.AggregationsSumAggregate;
 };
 
-type ConsumedCreditsAggs = {
+type ConsumedCreditsGroup = {
   by_user?: estypes.AggregationsMultiBucketAggregateBase<ConsumedCreditsBucket>;
+};
+
+type ConsumedCreditsAggs = {
+  paid_users?: ConsumedCreditsGroup;
+  free_users?: ConsumedCreditsGroup;
 };
 
 // Per-user consumed AWU credits for the current billing cycle, summed from the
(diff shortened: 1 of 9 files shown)


front/lib/api/credits/members_usage.ts
```diff
@@ -244,11 +259,46 @@ async function fetchConsumedAwuCreditsByUserId({
...
-        // Scoped to `userIds`, so `size` covers every distinct bucket and the
-        // per-bucket sums are exact (no cross-shard top-N approximation).
-        by_user: {
-          terms: { field: "user_id", size: userIds.length },
-          aggs: { credits: { sum: { field: "cost.full_awu" } } },
+        // Paid (and seatless) users: count only paid-seat usage, excluding any
+        // free-seat usage from before an upgrade. `must_not is_free_seat=true`
+        // (rather than `is_free_seat=false`) so historical docs indexed before
+        // this field existed — which can't be backfilled — count as paid.
+        paid_users: {
```
tdraier: can't we make a single aggregate on the 2 fields user_id + is_free_user ?

tdraier (approved):

PR #27851, opened 2026-06-23
Fix Gong connectors on non-default regions hitting the wrong API host

flvndvd:
## Description

This PR fixes https://github.com/dust-tt/tasks/issues/9123.

Some Gong customers are provisioned on a region-specific API host rather than the default one, and Gong returns that host during the OAuth token exchange. We were already capturing it at connector creation and storing it on the connector configuration, but the API client ignored it and always talked to the default host. For customers on another region this meant every request landed on the wrong host and came back as a forbidden response, which we interpret as a revoked token, so these connectors surfaced a misleading external OAuth token error even though their credentials were perfectly valid.

This change makes the API client use the per-customer host that we already store on the configuration instead of the hardcoded default. No migration is needed since that value has always been captured and stored as a required field, so every existing connector already has the correct host available and simply starts using it.

### DIFF
--- a/connectors/src/connectors/gong/lib/gong_api.ts
+++ b/connectors/src/connectors/gong/lib/gong_api.ts
@@ -265,12 +265,19 @@ export function clampRetryAfterSeconds(
 }
 
 export class GongClient {
-  private readonly baseUrl = "https://api.gong.io/v2";
+  private readonly baseUrl: string;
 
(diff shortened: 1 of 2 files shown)


flvndvd:
PMRR.

adrsimon (commented):
LGTM

PR #27856, opened 2026-06-24
Unify files and knowledge search slash command.

Fraggle:
## Description

Merges the separate "reference file" and "attach knowledge" slash commands into a single unified context search entry point. The old `FileSearchNode` / `KnowledgeSearchNode` TipTap atoms and their views are replaced by a single `ContextSearchNode` (`insertContextSearchNode`), backed by the new `ContextSlashSearch` component that renders both file results (via the extracted `useContextFileSlashSearchItems` hook) and knowledge results (via `useUnifiedSearch`) in one dropdown.

- `ContextFileSlashSearch` is refactored from a self-contained component into a `useContextFileSlashSearchItems` hook, controlled by an `includeFiles` flag.
- `INSERT_CONTEXT_FILE_SLASH_COMMAND_ACTION` and the `"reference-file"` slash command id are removed; the "attach-knowledge" command now inserts `contextSearchNode`.
- `includeContextFilesRef` gate no longer requires `actions.includes("attachment")` — it only checks that a conversation or space is present.

## Tests

Updated `InputBarSlashSuggestionDropdown.test.ts` to reflect the removal of `includeSelectContextFile` and `"reference-file"` from the dropdown items, including new query-match cases (`"reference"` and `"company"` both resolve to `INSERT_KNOWLEDGE_SLASH_COMMAND_ACTION`). Tested manually in the input bar.

## Risk

Low. Front-only change. The `ContextSearchNode` replaces two separate nodes; any in-flight editor state with `fileSearchNode` or `knowledgeSearchNode` atoms would be unrecognised by TipTap and dropped, but these nodes are ephemeral (never persisted). No API or data model changes.

## Deploy Plan

Deploy `front` only.


Fraggle:

- #27856 ◀
    - #27859
        - #27860
            - #27861

ElPicador (approved):

PR #27866, opened 2026-06-24
refactor(mcp): move google_calendar meeting defaults from server instructions to create_event description

ElPicador:
## Description

Move the three meeting-creation defaults out of `GOOGLE_CALENDAR_SERVER.serverInfo.instructions` and into the `create_event` tool description directly.

**Why:** `serverInfo.instructions` is injected into the system prompt on every agent loop turn (via `generation.ts`), which bloats the prompt prefix and busts Anthropic's stable cache block. With BM25 tool search, server instructions are also not indexed — only tool names, descriptions, and input schemas are — so the guidance was invisible to tool search. Putting the defaults on the tool description means they are cache-stable, BM25-searchable, and only surfaced when the model considers calling `create_event`.

## Tests

No logic changes — description string and null instructions only.

### DIFF
--- a/front/lib/api/actions/servers/google_calendar/metadata.ts
+++ b/front/lib/api/actions/servers/google_calendar/metadata.ts
@@ -106,7 +106,8 @@ export const GOOGLE_CALENDAR_TOOLS_METADATA = createToolsRecord({
     },
   },
   create_event: {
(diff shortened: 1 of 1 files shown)


ElPicador:

- #27866 ◀

flvndvd (approved):
🐎

PR #27868, opened 2026-06-24
fix(extension): load PCM AudioWorklet from static file to comply with Chrome MV3 CSP

ElPicador:
## Description

Fixes live STT in both Chrome and Firefox extensions. Several MV3 CSP constraints had to be worked around:

**AudioWorklet (both browsers)**
- The PCM processor worklet was loaded via a `blob:` URL — blocked by MV3 `script-src` CSP in both Chrome and Firefox (Chrome refuses to load the manifest if `blob:` is declared; Firefox rejects the connection)
- Added `pcm-processor.js` as a static asset copied to the extension build dir, and updated `useVoiceLiveTranscriberService` to detect `chrome.runtime.getURL` and load it from the extension origin instead of a blob URL

**ElevenLabs WebSocket**
- Added `wss://api.elevenlabs.io` to `connect-src` in all Chrome and Firefox manifests (dev/production/release)

### DIFF
--- a/extension/platforms/chrome/manifests/manifest.development.json
+++ b/extension/platforms/chrome/manifests/manifest.development.json
@@ -6,7 +6,7 @@
     "128": "images/dust32green.png"
   },
   "content_security_policy": {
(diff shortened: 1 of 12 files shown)


ElPicador:

- #27866
    - #27868 ◀

matteotrab (approved):

PR #27870, opened 2026-06-24
Add workspace opt-out for Computer features

fontanierh:
## Description

Computer is moving toward a broader sandbox rollout, but some customers still need workspace-level opt-out while the regular sandbox flags may be enabled globally or per workspace.

This adds `disable_computer_feature`, an on-demand workspace flag that leaves the opt-out visible while making Computer feature gates treat `sandbox_tools`, `sandbox_dsbx_tools`, and `sandbox_workspace_admin` as disabled. The JS SDK feature flag union is updated too.

## Tests

Added auth tests covering workspace-level sandbox flags and globally rolled-out sandbox flags under `disable_computer_feature`, plus a front-api upload test for the sandbox CSV limit opt-out.

## Risk

Worst case, a workspace with the disable flag loses Computer access until the flag is removed. Safe to rollback.

## Deploy Plan

Standard deploy. Enable `disable_computer_feature` only for workspaces that need the opt-out.

### DIFF
--- a/front-api/middlewares/sandbox_auth.ts
+++ b/front-api/middlewares/sandbox_auth.ts
@@ -5,6 +5,7 @@ import {
   getFeatureFlags,
 } from "@app/lib/auth";
 import { getClientIp } from "@app/lib/utils/request";
+import { isComputerFeatureEnabled } from "@app/types/shared/feature_flags";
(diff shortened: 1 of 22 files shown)


aubin-tchoi (commented):
LGTM

PR #27872, opened 2026-06-24
feat: enable live STT to everyone

ElPicador:
## Description

Makes live speech-to-text the only transcription path in the input bar, removing the `live_speech_to_text` feature flag gate that limited it to specific workspaces.

Since the legacy record-then-upload `useVoiceTranscriberService` is no longer reachable, it's deleted outright. Its shared helpers (`requestMicrophone` and the browser-extension microphone permission flow) move into `hooks/utils/voice.ts` so `useVoiceLiveTranscriberService` can keep using them.


ElPicador:

- #27872 ◀

Fraggle (approved):
LGTM

PR #27873, opened 2026-06-24
 Gate whitelabel frames behind a feature flag instead of the plan

flvndvd:
## Description

Whitelabel frames were gated on a dedicated plan entitlement column. That made every enablement an operational chore: the GTM team had to get a plan edited in Poke to turn the feature on for a workspace, which is heavier than it should be for what is effectively a toggle. This moves the gate to a workspace feature flag so it can be flipped on demand without touching plans, and drops the now-unused entitlement column from the plan. We also settled on the name "whitelabel frames" over "branded frames", so the concept is renamed across the gate, the test factory, and user-facing copy. The broader workspace branding surface keeps its existing name and routes, since branding is wider than just this feature.

💡 One thing worth being explicit about: a feature flag is not where this should live long term. Flags are meant for rollout and experimentation, not for permanent entitlement, and using one as a durable commercial gate is a smell. We are doing it deliberately because we do not yet have an add-on mechanism to attach this to, and a flag is the lowest-friction option available today that unblocks GTM. When add-ons exist, this should migrate onto a proper entitlement and the flag should be retired.

## Deploy Plan

- [ ] Deploy front
- [ ] Confirm no workspaces were granted the feature
- [ ] Run migration

### DIFF
--- a/extension/ui/components/auth/AuthProvider.tsx
+++ b/extension/ui/components/auth/AuthProvider.tsx
@@ -87,7 +87,6 @@ const EXTENSION_SUBSCRIPTION: SubscriptionType = {
     },
     trialPeriodDays: 0,
     isByok: false,
-    isBrandedFramesAllowed: false,
     isAuditLogsAllowed: false,
   },
   requestCancelAt: null,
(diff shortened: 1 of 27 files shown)


flvndvd:
PMRR

ykmsd (commented):
sorry for the delay LGTM

PR #27878, opened 2026-06-24
Unify Computer feature flags

fontanierh:
## Description

Follow up of #27870.

Fold the Computer rollout flags into `sandbox_tools`. Workspaces with `sandbox_tools` and without `disable_computer_feature` get the full Computer feature set: sandbox tools, dsbx, and workspace admin configuration. The disable flag wins at each Computer call site.

SDK compatibility note: the old SDK feature literals stay typed for client compatibility. The backend no longer whitelists or emits those flags.

## Tests

Tested locally:
- `npx biome check --write --error-on-warnings` on changed files
- `npm test -- lib/auth.global_feature_flags.test.ts lib/resources/skill/code_defined/sandbox.test.ts lib/api/actions/servers/sandbox/tools/index.test.ts` from `front`
- `npm run build:types` from `sdks/js`

Could not run the `front-api` route slice in this checkout because Vite fails before test execution on missing `rehype-raw` from the built Sparkle package. Full `front` and `front-api` typechecks also fail on existing Sparkle/API type drift unrelated to this change.

## Risk

Worst case, Computer access is gated too broadly or too narrowly for a workspace. Safe to rollback.

## Deploy Plan

Deploy front, front-sse, and front-api together through the standard pipeline.

- [ ] Deploy front.
- [ ] Deploy front-sse with front because `front/lib/auth.ts` changed.
- [ ] Deploy front-api.
- [ ] No Swagger schema change needed, only route gating changed.
- [ ] No sandbox image rebuild needed, only sandbox image helper comments changed.


spolu (approved):
LGTM

PR #27874, opened 2026-06-24
refactor(mcp): move slack formatting rules from server instructions to posting tool descriptions

ElPicador:
## Description

Move Markdown formatting rules (`[text](url)`, `**bold**`, `<@USER_ID>` mentions, `#CHANNEL` references) and the bot "must be added to channel" constraint from `serverInfo.instructions` into the `message` parameter descriptions of the relevant posting tools (`post_message`, `edit_message`, `schedule_message`).

**Why:** `serverInfo.instructions` is injected into the system prompt on every agent loop turn, bloating the prefix and busting Anthropic's stable cache block. BM25 tool search indexes tool names, descriptions, and input schemas — not server instructions — so the guidance was invisible to tool search. Moving the rules to the `message` parameter description makes them cache-stable, BM25-searchable, and only surfaced when the model is about to call a posting tool.

The bot-channel precondition was already present on the `post_message` and `read_channel_history` tool descriptions; no new content was needed there.

PR #27879, opened 2026-06-24
refactor(mcp): move schedules_management server instructions to create_schedule description

ElPicador:
## Description

Move the three constraints from `schedules_management` `serverInfo.instructions` onto the `create_schedule` tool description:
- user-scoping ("each user can only view and manage their own schedules")
- trigger behavior ("runs this agent with the specified prompt")
- rate limit ("20 schedule creations per user per day")

Same motivation as the prior Slack and Google Calendar PRs: server instructions are injected into the system prompt on every turn, bloating the prompt prefix and busting the stable cache block. They are also not indexed by BM25 tool search. Moving them to the tool description makes them cache-stable, searchable, and only surfaced when the model is considering `create_schedule`.


ElPicador:

- #27874
    - #27879 ◀
        - #27881

flvndvd (approved):
Should we also the update the `list_schedules` to also surface that it returns only the schedules for agent and current user?

PR #27881, opened 2026-06-24
refactor(mcp): move snowflake server instructions to query tool description

ElPicador:
## Description

Move the two constraints from `snowflake` `serverInfo.instructions` into the `query` tool description:
- schema-exploration precondition ("use list_databases, list_schemas, list_tables, describe_table / describe_semantic_view before writing queries")
- SELECT-only restriction ("only SELECT queries are allowed")

The `query` description already said "read-only" and "use dedicated tools" — the server instructions added the specific tool names and the SELECT framing. The merged description avoids duplication while preserving all guidance.

Same motivation as the prior Slack, Google Calendar, and schedules_management PRs: server instructions inject into the system prompt on every turn, busting the stable cache block and staying invisible to BM25 tool search.

### DIFF
--- a/front/lib/api/actions/servers/snowflake/metadata.ts
+++ b/front/lib/api/actions/servers/snowflake/metadata.ts
@@ -6,8 +6,16 @@ import { zodToJsonSchema } from "zod-to-json-schema";
 
 export const MAX_QUERY_ROWS = 1000;
 
+export const SNOWFLAKE_LIST_DATABASES_TOOL_NAME = "list_databases" as const;
+export const SNOWFLAKE_LIST_SCHEMAS_TOOL_NAME = "list_schemas" as const;
+export const SNOWFLAKE_LIST_TABLES_TOOL_NAME = "list_tables" as const;
+export const SNOWFLAKE_DESCRIBE_TABLE_TOOL_NAME = "describe_table" as const;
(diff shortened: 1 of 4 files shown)


ElPicador:

- #27874
    - #27879
        - #27881 ◀

front/lib/api/actions/servers/snowflake/metadata.ts
```diff
@@ -77,7 +77,7 @@ export const SNOWFLAKE_TOOLS_METADATA = createToolsRecord({
   },
   query: {
     description:
-      "Execute a read-only SQL query against Snowflake. Write operations are not allowed. Use the dedicated tools for listing databases, schemas, tables, and describing table structure.",
+      "Execute a read-only SQL query against Snowflake. Only SELECT queries are allowed — write operations are not permitted. Before writing a query, use list_databases, list_schemas, list_tables, and describe_table (or describe_semantic_view for semantic views) to explore the schema.",
```
flvndvd: Can we avoid the emdashes 🙏

flvndvd: And we usually try to use constant over magic string for the tool names as we sometimes miss tool names in description when refactoring them.

flvndvd (approved):

PR #27883, opened 2026-06-24
fix(gmail): stop garbling HTML thread history in reply quotes

TimotheeCrouzet:
## Description

Fixes garbled quoted thread history in Gmail reply drafts. When replying to an email with an HTML body (HTML-only messages), create_draft / send_mail flattened the body with a naive regex and ran it through escape(), which entity-encoded the markup (raw tags shown), leaked <style> CSS and left tag fragments.

decodeMessageBody now returns the decoded body with its mimeType. createQuoteSection branches on it: plain-text is escaped (as before), HTML is embedded as-is after sanitization (sanitize-html). The regex strip is removed.

Fixes: [link](https://github.com/dust-tt/tasks/issues/9095)

**Note:** multipart emails carry both a plain-text and an HTML copy of the body. We quote the plain-text copy, which has no formatting (bold/italic don't exist in plain text), so the quote shows content (links, symbols) but not rich formatting. Only HTML-only emails produce a formatted quote. Quoting the HTML copy everywhere is a possible follow-up.

## Tests

Manual. Replied (draft and direct send) to an HTML-only email: formatting rendered, links and < > & preserved, no entity-encoded tags. Replied to a multipart email: plain-text quote rendered correctly.

## Risk

Low, scoped to Gmail reply-quote construction. sanitize-html is already a dependency. Safe to roll back by revert.

## Deploy Plan

### DIFF
--- a/front/lib/api/actions/servers/gmail/helpers.ts
+++ b/front/lib/api/actions/servers/gmail/helpers.ts
@@ -1,4 +1,5 @@
 import { escape } from "html-escaper";
+import sanitizeHtml from "sanitize-html";
 
 export interface GmailHeader {
   name: string;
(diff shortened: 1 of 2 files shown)


frankaloia (approved):
LGTM

PR #27889, opened 2026-06-24
Relax schema validation for MCP servers (as does MCP Inspector)

Fraggle:
## Description

Since MCP SDK 1.22, `Client.listTools()` pre-compiles AJV validators for each tool's `outputSchema`. Remote servers like Google Stitch declare `$ref`s in `outputSchema` without in-document `$defs`, which causes `listTools()` to throw — even though the RPC response is valid and the data is usable.

Introduces `NoOpJsonSchemaValidator` (implementing the `jsonSchemaValidator` interface) and passes it via `{ jsonSchemaValidator: NO_OP_MCP_JSON_SCHEMA_VALIDATOR }` to the `Client` constructor in `connectToMCPServer`. Dust only consumes `inputSchema` from `tools/list` and never validates structured tool output, so skipping output-schema compilation has no functional impact. This mirrors MCP Inspector's `try/catch` around `ajv.compile`.

This fixes google stitch MCP sever ([we are not alone](https://www.google.com/search?client=firefox-b-d&q=%22can%27t+resolve+reference+%23%2F%24defs%2FScreenInstance+from+id+%23%22)).

## Tests

Tested manually against a Google Stitch MCP server that was previously failing on `listTools()`.

## Risk

Low. No change to tool execution or input handling — only output-schema compilation is bypassed. The `NoOpJsonSchemaValidator` always returns `valid: true`, matching the behavior Dust already relies on (output schema was never enforced before SDK 1.22).

## Deploy Plan

Deploy front.

### DIFF
--- a/front/lib/actions/mcp_helper.ts
+++ b/front/lib/actions/mcp_helper.ts
@@ -32,6 +32,34 @@ import {
   asDisplayToolName,
 } from "@app/types/shared/utils/string_utils";
 import { McpError } from "@modelcontextprotocol/sdk/types.js";
+import type {
+  JsonSchemaType,
+  JsonSchemaValidator,
+  jsonSchemaValidator,
+} from "@modelcontextprotocol/sdk/validation/types.js";
+
+/**
+ * Disables MCP SDK output-schema validation.
+ *
+ * Since SDK 1.22, `Client.listTools()` pre-compiles AJV validators for each tool's
+ * `outputSchema`. Some remote servers (e.g. Google Stitch) declare `$ref`s in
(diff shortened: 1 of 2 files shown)


Fraggle:

- #27889 ◀

ElPicador (approved):
could we add a unit test on that?

Or at least a check on when we upgrde the SDK that we don't break this?

Fraggle:
> could we add a unit test on that?

We don't have a system yet to test the sdk.

> Or at least a check on when we upgrde the SDK that we don't break this?

Shouldn't, we are not doing a hack, just passing a custom validator.

PR #27897, opened 2026-06-24
feat(front): always show message credits in header behind feature flag

Nils-Fedrigo:
## Description

Adds an opt-in display of the credit cost next to the timestamp in agent message headers (e.g. `14:18 · 1,287 credits`), gated behind the new `always_show_message_credits` feature flag (stage `dust_only`, credit-priced plans only).

- New `credits` slot on `ConversationMessageTitle` (sparkle), rendered after the timestamp with a `·` separator in the same muted style.
- When the flag is on, the credit cost is shown inline in the header instead of in the message menu (the menu entry is dropped to avoid duplication).
- The cost is written asynchronously after a message finishes, so the header polls the single-message endpoint until the value lands, then stops:
  - A subtle `LoadingBlock` skeleton (less prominent than the spinner) is shown while the cost is still pending.
  - Polling and the loader are bounded to a 2-minute window after completion, so legacy messages that never recorded a cost don't spin or poll forever.
- Extended `useConversationMessage` to accept an optional `refreshInterval`.

When the flag is off, behavior is unchanged (cost stays in the message menu, fetched on menu open only).

## Tests

Manual, with the flag enabled on a credit-priced workspace:
- Send a message and confirm the loader appears on completion, then resolves to the credit value.
- Confirm the credit cost no longer appears in the message dropdown when the flag is on.
- Confirm older/historical messages show their recorded cost (or nothing) without a perpetual loader.
- Confirm no credits are shown for sub-agent (`parentAgent`) messages or while streaming.

### DIFF
--- a/front/components/assistant/conversation/AgentMessage.tsx
+++ b/front/components/assistant/conversation/AgentMessage.tsx
@@ -57,6 +57,7 @@ import { isImageProgressOutput } from "@app/lib/actions/mcp_internal_actions/out
 import { CONTEXT_WINDOW_DOC_URL } from "@app/lib/api/assistant/errors";
 import config from "@app/lib/api/config";
 import { useAuth, useFeatureFlags } from "@app/lib/auth/AuthContext";
+import { formatCredits } from "@app/lib/client/credits";
 import { clientFetch } from "@app/lib/egress/client";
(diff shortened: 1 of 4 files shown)


Nils-Fedrigo:

- #27897 ◀

PR #27898, opened 2026-06-24
[Agent Builder] Block save while data is loading

philipperolet:
## Description
Fixes https://github.com/dust-tt/tasks/issues/8990

Agent Builder loads parts of an existing agent form asynchronously. This blocks saving while that data is still loading, so an edit cannot persist an incomplete payload before the current skills/tools/triggers/editors have hydrated.

This is frontend-only and does not change API behavior.

## Risks
Blast radius: Agent Builder save button state while editing or duplicating agents.
Risk: low

## Deploy Plan
- pmrr
- deploy front


ykmsd (commented):
👍

PR #27781, opened 2026-06-23
[Skills] Remove customized skill code

philipperolet:
## Description
Follows https://github.com/dust-tt/dust/pull/27106. Related to https://github.com/dust-tt/tasks/issues/8616.

Removes the legacy customized-skill runtime now that production skills were backfilled to explicit skill references. This deletes the one-shot backfill script, removes private create handling for `extendedSkillId`, removes extended-skill relation fetching and UI badges/form plumbing, and stops injecting former base-skill instructions/tools through `extendedSkill` at agent-loop time. Migrated skills now rely on their explicit `<skill ... />` references.

The public skill response keeps deprecated compatibility fields (`isExtendable: false`, `extendedSkillId: null`) and the Sequelize model keeps the nullable legacy column until a later schema migration drops it, so this cleanup does not break the public response shape or make generated migrations drift.

## Risks
Blast radius: skill creation/editing/list/detail API responses, skill builder UI, and enabled-skill prompt/tool rendering.
Risk: standard. The legacy compatibility path is removed, while regular skill references remain unchanged.

## Deploy Plan
- Confirm the customized-skill backfill has run in all production regions.
- Deploy `front-api` and `front` in the same rollout. If sequenced, deploy `front-api` first and `front` immediately after.
- No DB migration; legacy columns remain as inert schema history.

## Tests
- [ ] `cd front-api && CI=1 NODE_ENV=test npm test -- routes/w/[wId]/skills/index.test.ts` (blocked in this hive: missing `@novu/framework` in node_modules)
- [ ] `cd front && NODE_ENV=test npm test lib/resources/skill/skill_resource.test.ts` (blocked in this hive: missing `@novu/framework` in node_modules)

PR #27899, opened 2026-06-24
[Skills] Drop legacy skill extension columns

philipperolet:
## Description
Related to https://github.com/dust-tt/tasks/issues/8616. Follows https://github.com/dust-tt/dust/pull/27781.

Drops the legacy customized-skill DB columns now that customized skills have been backfilled and the runtime code no longer reads them. The public skill response compatibility fields stay unchanged: `isExtendable` remains `false` and `extendedSkillId` remains `null`.

The post-deploy migration was generated with the new `front/migrations/post-deploy` workflow, then reduced to the two intended `DROP COLUMN` statements because the local DB has older pending migrations and the raw generator output included unrelated local drift.

## Risks
Blast radius: skill Sequelize models and post-deploy schema migration for legacy columns.
Risk: low. Runtime code no longer reads these DB columns, and public response compatibility stays unchanged.

## Deploy Plan
- Deploy `front`
- run the front post-deploy migrations


spolu (approved):
LGTM

PR #27901, opened 2026-06-24
Gate conversation sandbox status behind Computer feature

fontanierh:
## Description

Gate the conversation sandbox status endpoint behind the Computer feature middleware, so workspaces without Computer enabled cannot access sandbox state. Also stop the files panel from polling sandbox status when Computer is disabled, and add route coverage for enabled, disabled, missing sandbox, and unsupported method cases.

## Tests

- Added `front-api` route tests for enabled, no sandbox, disabled Computer, and unsupported methods.
- Ran the focused `front-api` sandbox route test in the Hive environment.
- Ran `npm run lint:swagger-annotations` in `front-api`.
- Ran `npm run tsgo -- --noEmit` in `front` and `front-api`.

## Risk

Low. The API change only gates the sandbox status endpoint with the existing Computer feature middleware. The UI change avoids a fetch that would be rejected when Computer is disabled.

## Deploy Plan

Standard deploy.

### DIFF
--- a/front-api/routes/w/[wId]/assistant/conversations/[cId]/sandbox/index.test.ts
+++ b/front-api/routes/w/[wId]/assistant/conversations/[cId]/sandbox/index.test.ts
@@ -60,3 +60,20 @@ describe("GET /api/w/:wId/assistant/conversations/:cId/sandbox", () => {
     });
   });
 });
+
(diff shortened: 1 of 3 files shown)


fontanierh:
Closing this, it adds a new feature gate to a route that was not previously gated and changes legacy mounted-files behavior.

PR #27910, opened 2026-06-24
feat(front): polish notification settings UX (save toast, gated customize, sparkle typography)

Inesvvv:
## Description
Stacked on top of `sound-notification-settings` (PR 1). UX polish for the sound notification settings panel:

- **Save confirmation** — shows a success toast ("Notification preferences saved") when the save succeeds.
- **Gate "Customize" on the toggle** — the sound picker and Play button are disabled when the toggle is off, and the "Customize" row dims to `faint` to signal you must toggle on first.
- **Typography** — use the Sparkle tokens from Figma: rows use `heading-sm`/`copy-sm`, and the section header uses `heading-2xl`/`copy-sm` (shared `SectionContent`, so this also applies to the other settings section headers).

## Tests

Tested locally:
- Toggle off → picker + Play disabled, "Customize" row greyed (`faint`).
- Toggle on → controls enabled, row at normal colors.
- Change a value + Save → success toast, Save disables. Failed save → no success toast, Save stays enabled.

## Risk

Low. Frontend-only, behind the `sound_notification` flag from PR 1

## Deploy Plan

Merge after PR 1

### DIFF
--- a/front/components/UserSettingsPopover.tsx
+++ b/front/components/UserSettingsPopover.tsx
@@ -12,6 +12,7 @@ import { useTheme } from "@app/components/sparkle/ThemeContext";
 import { AwuUsageBar } from "@app/components/workspace/MembersUsageTable";
 import { useFileUploaderService } from "@app/hooks/useFileUploaderService";
 import { useIsMac } from "@app/hooks/useKeyboardShortcutLabel";
+import { useSendNotification } from "@app/hooks/useNotification";
 import { useAuth, useFeatureFlags } from "@app/lib/auth/AuthContext";
 import { isSubmitMessageKey } from "@app/lib/keymaps";
 import { useMyUsage, useSeatPlan } from "@app/lib/swr/credits";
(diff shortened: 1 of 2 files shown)


front/components/UserSettingsPopover.tsx
```diff
@@ -566,12 +568,18 @@ function NotificationsSection({ owner }: { owner: WorkspaceType }) {
...
 
   const handleSave = async () => {
-    await Promise.all([
+    const results = await Promise.all([
       soundNotificationPreferencesRef.current?.savePreferences(),
       notificationPreferencesRef.current?.savePreferences(),
     ]);
-    setSoundDirty(false);
-    setNotifDirty(false);
+    if (results.every((result) => result !== false)) {
```
PopDaph: Else we show the error notif?

PopDaph (approved):

front/components/UserSettingsPopover.tsx
Inesvvv: I looked into the error notif (toast). If I add an else with an error notif it appears several times (for both the toggle and the selected sound) so instead I added it into the UserSettingsPopover so that it appears only once whenever there is an issue saving notifications. Should I update the push or merge it like this?

PR #27911, opened 2026-06-24
Remove slideshow MCP server

flvndvd:
## Description

This PR fixes https://github.com/dust-tt/tasks/issues/9198.

Removes the slideshow internal MCP server. It was a preview, feature-flag-gated tool whose functionality is now fully covered by the interactive content server, which already produces and renders slideshow presentations through its own content type. This deletes the server implementation and all of its wiring: the registry entry, the global-agent attachment, the deep-dive instruction, the display label, the action weight, the marketing exclusion, and the feature flag.

The server's numeric id is preserved as a legacy id so that historical actions in existing conversations keep resolving and rendering. Past actions are intentionally not deleted. The slideshow content type and all of its rendering, file categorization, and PDF export behavior are left untouched because they belong to the interactive content server, not to the removed slideshow server.

A migration is included to clean up the live wiring that was created per workspace while the server existed: the MCP server views, the user tool approvals, and the remote tool metadata. It runs across all workspaces in dry-run mode by default, and with execute to apply. Run it after deploy.


ElPicador (approved):
I love the smell of burned code in the afternoon

PR #27494, opened 2026-06-17
Introduce conversation sandbox ownership link

fontanierh:
## Description

Shared Computer S1 refactor slice. Context: [Shared Computer v0](https://app.notion.com/p/38128599d9418127b78ff7285cb722e4).

This PR moves existing conversation sandbox ownership toward a join-table model instead of adding more ownership fields directly to `sandboxes`.

The schema/model changes are:

- add `conversation_sandboxes` as the conversation-to-sandbox ownership table;
- add a standalone data migration script to backfill `conversation_sandboxes` from existing `sandboxes.conversationId` rows and verify both models are ISO;
- add uniqueness for one sandbox per conversation and one conversation owner per sandbox;
- keep writing the legacy `sandboxes.conversationId` column for deploy compatibility until a later PR drops it.

The runtime changes are limited to existing conversation sandboxes:

- `SandboxResource.makeNew` writes both `sandboxes.conversationId` and `conversation_sandboxes`;
- sandbox deletion paths delete the `conversation_sandboxes` row alongside the legacy sandbox row;
- reads remain on the legacy `sandboxes.conversationId` model in this PR;
- no Pod-scoped sandbox creation/fetch path is introduced in this PR.

## Follow-ups

- Switch conversation sandbox reads to `conversation_sandboxes`.
- Introduce the Pod/space sandbox ownership table and model.
- Update Pod-scoped runtime paths to use that table.
- Remove the legacy fallback once all writers use ownership tables.
- Drop `sandboxes.conversationId` in a later post-deploy migration.

## Tests

- `npm run format:changed`
- `source ~/.dust-hive/envs/shared-computer-s1/env.sh && cd front && FRONT_DATABASE_URI="$TEST_FRONT_DATABASE_URI" REDIS_URI="$TEST_REDIS_URI" REDIS_CACHE_URI="$TEST_REDIS_URI" NODE_ENV=test npx tsx admin/db.ts`
- `source ~/.dust-hive/envs/shared-computer-s1/env.sh && cd front && FRONT_DATABASE_URI="$TEST_FRONT_DATABASE_URI" NODE_ENV=test npx tsx migrations/20260624_backfill_conversation_sandboxes.ts`
- `source ~/.dust-hive/envs/shared-computer-s1/env.sh && cd front && FRONT_DATABASE_URI="$TEST_FRONT_DATABASE_URI" NODE_ENV=test npx tsx migrations/20260624_backfill_conversation_sandboxes.ts --execute`
[...]

PR #27916, opened 2026-06-24
Switch sandbox reads to ownership table

fontanierh:
## Description

Follow up of #27494.

Switch sandbox lookup paths to prefer `conversation_sandboxes` while keeping `sandboxes.conversationId` as the fallback during the migration window. The reaper now resolves conversation owners through the ownership table too.

## Tests

Added resource tests for preferred ownership reads, legacy fallback, the ownership-map helper, and its workspace guard.

Locally: `npm test -- lib/resources/sandbox_resource.test.ts`.

## Risk

Worst case, a bad ownership row points a lookup at the wrong sandbox. Safe to rollback because the legacy column remains, fallback still exists, and #27494 keeps writes mirrored.

## Deploy Plan

- [x] Merge and deploy #27494 first.
- [x] Run #27494's standalone data migration with `--execute` in every production region and confirm it completes with no missing or extra ownership rows.
- [x] Deploy this PR after the ownership table and legacy column are ISO.
- [ ] Monitor sandbox lifecycle and reaper warnings after deploy.

### DIFF
--- a/front/lib/resources/sandbox_resource.ts
+++ b/front/lib/resources/sandbox_resource.ts
@@ -53,6 +53,8 @@ export interface SandboxResource extends ReadonlyAttributesType<SandboxModel> {}
 // eslint-disable-next-line @typescript-eslint/no-unsafe-declaration-merging
 export class SandboxResource extends BaseResource<SandboxModel> {
   static model: ModelStaticWorkspaceAware<SandboxModel> = SandboxModel;
+  private static conversationSandboxModel: ModelStaticWorkspaceAware<ConversationSandboxModel> =
+    ConversationSandboxModel;
 
   private static deleteEgressPolicyAfterDestroy(
     sandbox: SandboxResource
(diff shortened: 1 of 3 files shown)


front/lib/resources/sandbox_resource.ts
```diff
@@ -232,6 +270,48 @@ export class SandboxResource extends BaseResource<SandboxModel> {
...
+        sandboxModelIdsByWorkspaceModelId.get(sandbox.workspaceId) ?? [];
+      sandboxModelIds.push(sandbox.id);
+      sandboxModelIdsByWorkspaceModelId.set(
+        sandbox.workspaceId,
+        sandboxModelIds
+      );
+    }
+
+    const rows = (
+      await Promise.all(
```
flvndvd: Arf, can we avoid the `Promise.all()` here? Especially as we hit the DB, it could fan out in many queries.

flvndvd (approved):

PR #27918, opened 2026-06-24
Move HTTP client server instructions into the send_request tool description

flvndvd:
## Description

This PR fixes https://github.com/dust-tt/tasks/issues/9199. 

The HTTP client server _(soon to be retired, so best effort here)_ carried a long block of server-level instructions covering how and when to use it. Server instructions are injected once for the whole server and sit far from the moment the model actually decides to make a request, so the guidance was easy to lose track of and heavier on tokens than it needed to be.

This moves the guidance to where it matters, directly into the tool description, and trims it down to what actually changes behavior.


aubin-tchoi (approved):
LGTM

ElPicador (approved):

PR #27920, opened 2026-06-24
Fix selected spaces migration FK

fontanierh:
## Description

Fix the selected spaces pre-deploy migration by pointing `spaceId` at `vaults`, the physical table behind `SpaceModel`. PR #27539 referenced `spaces`, which does not exist in the front database.

## Tests

- Ran `npm run migration:status` in `front` and confirmed the migration is pending.
- Checked local DB metadata: `public.vaults` exists and `public.spaces` does not.
- Ran the migration file with `psql -f` in a throwaway schema with minimal FK tables.

## Risk

Low. SQL-only fix for a migration that currently fails before creating `conversation_selected_spaces`.

## Deploy Plan

- [ ] Merge this fix.
- [ ] Rerun `./scripts/run-migrations.sh pre-deploy front`.
- [ ] Confirm `pre-deploy/20260624143314_add_conversation_selected_spaces.sql` is no longer pending.


spolu (commented):
LGTM

PR #27921, opened 2026-06-24
feat(models): add workspace default model foundations

Nils-Fedrigo:
## Description

First PR of a stack introducing a **workspace-level default model**: a model an admin can pick to back the standard agents (Dust, dust-task, …) and to be the default for newly created custom agents. When unset, it resolves live to the best available model (today Sonnet 4.6); an admin can pin a model to freeze it.

This PR lands the foundations only — **no behavior change**, nothing consumes the setting yet:

- New sentinel model `workspace-default` (mirrors the `noop` virtual model) for custom agents that follow the default. Stored in the existing `STRING` columns, so no DB migration. Added to `STATIC_MODEL_IDS`, the token-pricing map (zero entry), and the `@dust-tt/client` SDK unions (model id + feature flag).
- Workspace setting plumbing: `workspaceDefaultModel` in `WorkspaceMetadata`, `getWorkspaceDefaultModelSetting` reader, and `LightWorkspaceType.metadata` typing.
- Resolution helpers in `lib/api/assistant/models.ts`: `getPinnedWorkspaceDefaultModel`, `getWorkspaceDefaultModel` (pinned-if-enabled, else live best large model), and `isModelEnabledForWorkspace`.
- `usesWorkspaceDefault` output flag on `AgentModelConfigurationSchema`.
- `workspace_default_model` feature flag (on-demand) — gates the whole feature across the stack.
- Client helper `getResolvedWorkspaceDefaultModel` for "currently X" display.

Stack:
1. **#27921 — foundations (this PR)**
2. #27922 — standard agents honor the default
3. #27923 — admins can configure it (endpoint + UI)
4. #27924 — new agents follow it (builder + display)

## Tests

Covered by later PRs in the stack (endpoint tests in #27923). This PR is type-only scaffolding; `tsgo` passes for `front`, `front-api`, and `sdks/js`.

## Risk

None. No code path reads the setting yet, and the SDK changes are additive union members (the `dist` is regenerated at build time).

PR #27922, opened 2026-06-24
feat(assistant): standard agents honor workspace default model

Nils-Fedrigo:
## Description

Part of the **workspace default model** stack (builds on #27921).

Wires the read path so the standard/global agents resolve their model through the workspace default:

- `getModelForAgentConfiguration` (the single agent-config serialization boundary) resolves the `workspace-default` sentinel to a concrete model via `getWorkspaceDefaultModel`, so it never reaches execution, and flags `usesWorkspaceDefault` for the UI.
- The Dust agent (and its reasoning-effort variants) and dust-task / deep-dive prefer the admin-pinned model when one is set, falling back to the existing preference list + `getLargeWhitelistedModel` otherwise. All existing special cases (free-tier small model, GPT-5.5 default flag, deep-dive enterprise Opus) are preserved when nothing is pinned.

**No behavior change in practice until #27923 lands** — there is no way to set a default value yet, so `getPinnedWorkspaceDefaultModel` always returns `null` and every agent keeps today's resolution.

## Tests

Manual once the full stack is in: pin a model in #27923 → confirm Dust and dust-task switch; clear it → confirm they revert. No standalone tests here (pure resolution logic exercised end-to-end via the stack).

## Risk

Low. With no pinned value the candidate lists and fallbacks are identical to today.

### DIFF
--- a/front/lib/api/assistant/configuration/helpers.ts
+++ b/front/lib/api/assistant/configuration/helpers.ts
@@ -1,6 +1,8 @@
 import { fetchMCPServerActionConfigurations } from "@app/lib/actions/configuration/mcp";
 import { getFavoriteStates } from "@app/lib/api/assistant/get_favorite_states";
+import { getWorkspaceDefaultModel } from "@app/lib/api/assistant/models";
 import type { Authenticator } from "@app/lib/auth";
+import { getFeatureFlags } from "@app/lib/auth";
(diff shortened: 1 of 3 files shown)


Nils-Fedrigo:

- #27921
    - #27922 ◀
        - #27923
            - #27924

PR #27923, opened 2026-06-24
feat(workspace): let admins configure the default model

Nils-Fedrigo:
## Description

Part of the **workspace default model** stack (builds on #27921, #27922).

Lets admins configure the workspace default model:

- `POST /api/w/[wId]` gains a `workspaceDefaultModel` body variant (`{ providerId, modelId }` or `null` to clear). Admin-only, gated by the `workspace_default_model` flag, validates the model is real and enabled for the workspace (`isModelEnabledForWorkspace`), then persists via `updateWorkspaceMetadata`. Mirrors the existing `workspaceDefaultAgentId` handler.
- **Model Providers** admin page gets a "Default model" section (`DefaultModelSelect`) showing `Automatic (currently X)` and the list of available models, saved through the new `useWorkspaceDefaultModel` hook. Picking a model shows a **cost/latency confirmation warning** ("this powers Dust, dust-task and every agent set to follow the workspace default"). No dollar figures — model configs carry no price data.

After this PR the standard agents (#27922) are fully controllable. New custom agents don't follow the default yet — that's #27924.

## Tests

Functional endpoint tests added in `front-api/routes/w/[wId]/index.test.ts`: pin a valid model, clear with `null` (preserving other metadata), 400 on unknown model, 403 when the flag is off, 403 for non-admins.

## Risk

Low. Endpoint and UI are flag-gated; clearing/invalid input handled. The setting only takes effect for agents wired in #27922 / #27924.

PR #27924, opened 2026-06-24
feat(agent-builder): new agents follow the workspace default model

Nils-Fedrigo:
## Description

Final PR of the **workspace default model** stack (builds on #27921–#27923).

Makes newly created custom agents follow the workspace default dynamically, and surfaces it in the builder/details:

- `createPendingAgentConfiguration` stores the `workspace-default` sentinel for new agents when the flag is on (otherwise the historical hardcoded Sonnet default).
- Agent builder initializes new agents to the sentinel and round-trips it: the model picker gains a "Workspace default (currently X)" entry, and `transformAgentConfigurationToFormData` maps the `usesWorkspaceDefault` flag back to the sentinel so saving keeps following the default.
- Agent details (`AgentInfoTab`) renders "Workspace default (currently X)" for agents that follow it.

Because the sentinel is resolved at the serialization boundary (#27922), every existing execution path keeps seeing a concrete model. Existing agents with pinned models are unaffected — only new agents follow by default.

## Tests

Manual end-to-end (full stack + flag on): create a new agent → shows "Workspace default"; set the workspace default in Model Providers → the new agent and Dust both switch; clear it → both revert to the live default; pin then un-whitelist that provider → graceful fallback. Existing pinned agents unchanged.

## Risk

Low and flag-gated. With `workspace_default_model` off, new agents keep the previous concrete Sonnet default and the picker entry is hidden.

### DIFF
--- a/front/components/agent_builder/AgentBuilder.tsx
+++ b/front/components/agent_builder/AgentBuilder.tsx
@@ -26,6 +26,7 @@ import { submitAgentBuilderForm } from "@app/components/agent_builder/submitAgen
 import {
   getDefaultAgentFormData,
   getDefaultModel,
+  getResolvedWorkspaceDefaultModel,
   transformAgentConfigurationToFormData,
   transformDuplicateAgentToFormData,
   transformTemplateToFormData,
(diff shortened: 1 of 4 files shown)


Nils-Fedrigo:

- #27921
    - #27922
        - #27923
            - #27924 ◀

PR #27925, opened 2026-06-24
Revert "[Agent Builder] Block save while data is loading"

philipperolet:
Reverts dust-tt/dust#27898

Deploy
- pmrr


ykmsd:
👍 (but don't trust my 👍 🥶)

PR #27926, opened 2026-06-24
[Agent Builder] Restore save hydration guard

philipperolet:
## Description
Related to https://github.com/dust-tt/tasks/issues/8990

Reapplies https://github.com/dust-tt/dust/pull/27898 after https://github.com/dust-tt/dust/pull/27925.

Save stays disabled until existing agent data has loaded/revalidated, but create and duplicate flows do not wait on editors.

## Risks
Blast radius: Agent Builder save button state.
Risk: low

## Deploy Plan
- pmrr (this time it does not block save)
- deploy front

### DIFF
--- a/front/components/agent_builder/AgentBuilder.tsx
+++ b/front/components/agent_builder/AgentBuilder.tsx
@@ -561,8 +561,17 @@ export default function AgentBuilder({
     });
   };
 
+  const { isDirty, isSubmitting } = form.formState;
+
+  const isSaveDisabled =
+    isSubmitting ||
+    isActionsLoading ||
+    isSkillsLoading ||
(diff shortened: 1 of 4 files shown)


philipperolet:
Context on the removed duplicateAgentId save-button exception: it was introduced when duplicating an agent did not necessarily make the form dirty, while Save was disabled for non-dirty forms. Duplication needed a bypass so users could save the copy.

That dirty-state check is gone now, so the bypass is no longer needed. Keeping it would also skip the loading guard for actions/skills that duplication needs to copy. The current fix keeps duplicate save blocked until that data is loaded, while avoiding editor loading for create/duplicate flows.

PR #27927, opened 2026-06-24
Move Salesforce server instructions into the tool descriptions

flvndvd:
## Description

This PR fixes https://github.com/dust-tt/tasks/issues/9207.

The Salesforce server carried a long block of server-level instructions that the model only saw once, far from the moment it picks a tool. This moves that guidance onto the tools themselves so it sits right at the decision point, and trims it to the parts that actually change how a call is made.

The read query tool now explains that it is read-only, the order in which to discover an object before querying it, how custom names are spelled, how to join related records, and how to read the inline field listing and its limit. The describe and list tools each state their role in that flow. No behavior on the request side changes, this is purely about where the guidance lives and keeping it lean.

After further analysis, we opt for changing the tools rather than creating a global or custom skill for one main reason: users have already created skills to explain how to use Salesforce, with instructions that sometimes overlap with the instructions we do put. Will be monitored through https://app.datadoghq.eu/notebook/323356/salesforce-tools-usage?refresh_mode=sliding&from_ts=1782395309872&to_ts=1782481709872


aubin-tchoi (approved):
sounds good to me but looks very hard to test and confirm this doesn't break the specific use cases some customers may have

ElPicador (approved):
trusting you on this one!

PR #27928, opened 2026-06-24
Move JIRA server instructions into the tool descriptions

flvndvd:
## Description

This PR fixes https://github.com/dust-tt/tasks/issues/9208.

The JIRA server carried a block of server-level instructions that the model only sees once, far from the point where it actually picks a tool. Most of that guidance was already duplicated in the individual tool descriptions, so it added length without adding signal. This folds the two pieces that were not already present at a call site, the authentication step and the get_issue default fields with the built-in versus custom field naming convention, into the descriptions of the tools they concern, then removes the server instructions and the now unused constant. Behavior is unchanged and the guidance now sits next to the decision it informs.

### DIFF
--- a/front/lib/actions/mcp_internal_actions/instructions.ts
+++ b/front/lib/actions/mcp_internal_actions/instructions.ts
@@ -43,27 +43,3 @@ The output includes:
 -   Other object-level properties.
 This is the most reliable way to discover the correct names for fields and relationships before writing an \`execute_read_query\`.
 `;
-
-export const JIRA_SERVER_INSTRUCTIONS = `
-      You have access to the following tools: get_issue, get_projects, get_project, get_transitions, create_comment, get_issues, get_issue_types, get_issue_create_fields, get_connection_info, transition_issue, create_issue, update_issue, create_issue_link, delete_issue_link, get_issue_link_types, get_users, get_issue_read_fields.
-
-      # General Workflow for JIRA Data:
-      1.  **Authenticate:** Use \`get_connection_info\` to authenticate with JIRA if you are not authenticated ("No access token found").
-      2.  **For Create/Update Operations:** Always use \`get_issue_types\` and \`get_issue_create_fields\` with the specific issue typename to get its create/update-time metadata (fields available on the Create/Update screens).
-      3.  **Execute Read Query:** Use \`get_issues\` to retrieve data using JQL. Construct your JQL queries based on the information obtained from \`get_issue_types\` to ensure you are using correct field and relationship names.
-
-      **Best Practices for Querying:**
-      1.  **Discover Object Structure First:** Use \`get_issue_create_fields\` to understand fields available for create/update in a given project and issue type. Alternatively, for a quick field list directly in a query, use \`get_issues\`.
-      2.  **Verify Field and Relationship Names:** If you encounter JIRA 400 errors suggesting that the field or relationship does not exist, use \`get_issue_types\` for the relevant object(s) to confirm the exact names and their availability.
-
-      **Field Selection for get_issue (Optional Performance Optimization):**
-      - By default, \`get_issue\` returns essential fields: summary, issuetype, priority, assignee, reporter, labels, duedate, parent, project, status
-      - Only use \`get_issue_read_fields\` if you need additional custom fields or want to optimize performance by requesting specific fields
-      - For built-in fields use the \`key\` (e.g., "summary", "issuetype", "status"). For custom fields use the \`id\` (e.g., "customfield_10020").
-
-      **User Lookup (get_users):**
-      - Provide emailAddress for exact match on email. Example: { "emailAddress": "jane.doe@acme.com" }
-      - Or provide name for case-insensitive contains on display name. Example: { "name": "Jane" }
-      - If neither emailAddress nor name is provided, the tool lists the first maxResults users.
-      - For pagination, pass startAt using the previous result's nextStartAt. Example: { "name": "Jane", "startAt": 100 }
-    `;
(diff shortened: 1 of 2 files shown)


thervier:
@flvndvd there is a risk here that the usage of the tools is less efficient right as we lose here some global instructions that integrate some best practice, no?

flvndvd:
> @flvndvd there is a risk here that the usage of the tools is less efficient right as we lose here some global instructions that integrate some best practice, no?

I tried my best to move all the relevant pieces onto the tool description and I'm counting on your expert eye on Jira to flag anything I'll have drop that was added on purpose or following some feedback. FWIW most servers don't have any instructions so I'd assume Jira should be pretty easy to align on this pattern.

front/lib/actions/mcp_internal_actions/instructions.ts
```diff
@@ -44,30 +44,6 @@ The output includes:
...
-      **Field Selection for get_issue (Optional Performance Optimization):**
-      - By default, \`get_issue\` returns essential fields: summary, issuetype, priority, assignee, reporter, labels, duedate, parent, project, status
-      - Only use \`get_issue_read_fields\` if you need additional custom fields or want to optimize performance by requesting specific fields
-      - For built-in fields use the \`key\` (e.g., "summary", "issuetype", "status"). For custom fields use the \`id\` (e.g., "customfield_10020").
-
-      **User Lookup (get_users):**
-      - Provide emailAddress for exact match on email. Example: { "emailAddress": "jane.doe@acme.com" }
-      - Or provide name for case-insensitive contains on display name. Example: { "name": "Jane" }
-      - If neither emailAddress nor name is provided, the tool lists the first maxResults users.
-      - For pagination, pass startAt using the previous result's nextStartAt. Example: { "name": "Jane", "startAt": 100 }
```
thervier, line 68: this part has not been moved to the get_users tool description ?

front/lib/actions/mcp_internal_actions/instructions.ts
```diff
@@ -44,30 +44,6 @@ The output includes:
 This is the most reliable way to discover the correct names for fields and relationships before writing an \`execute_read_query\`.
 `;
 
-export const JIRA_SERVER_INSTRUCTIONS = `
-      You have access to the following tools: get_issue, get_projects, get_project, get_transitions, create_comment, get_issues, get_issue_types, get_issue_create_fields, get_connection_info, transition_issue, create_issue, update_issue, create_issue_link, delete_issue_link, get_issue_link_types, get_users, get_issue_read_fields.
-
-      # General Workflow for JIRA Data:
-      1.  **Authenticate:** Use \`get_connection_info\` to authenticate with JIRA if you are not authenticated ("No access token found").
-      2.  **For Create/Update Operations:** Always use \`get_issue_types\` and \`get_issue_create_fields\` with the specific issue typename to get its create/update-time metadata (fields available on the Create/Update screens).
-      3.  **Execute Read Query:** Use \`get_issues\` to retrieve data using JQL. Construct your JQL queries based on the information obtained from \`get_issue_types\` to ensure you are using correct field and relationship names.
```
thervier, line 53: not moved to "get_issues" tool description ?

front/lib/actions/mcp_internal_actions/instructions.ts
```diff
@@ -44,30 +44,6 @@ The output includes:
 This is the most reliable way to discover the correct names for fields and relationships before writing an \`execute_read_query\`.
 `;
 
-export const JIRA_SERVER_INSTRUCTIONS = `
-      You have access to the following tools: get_issue, get_projects, get_project, get_transitions, create_comment, get_issues, get_issue_types, get_issue_create_fields, get_connection_info, transition_issue, create_issue, update_issue, create_issue_link, delete_issue_link, get_issue_link_types, get_users, get_issue_read_fields.
-
-      # General Workflow for JIRA Data:
-      1.  **Authenticate:** Use \`get_connection_info\` to authenticate with JIRA if you are not authenticated ("No access token found").
-      2.  **For Create/Update Operations:** Always use \`get_issue_types\` and \`get_issue_create_fields\` with the specific issue typename to get its create/update-time metadata (fields available on the Create/Update screens).
```
thervier, line 52: not moved to `get_issue_types\` and \`get_issue_create_fields\` tools descriptions?

front/lib/actions/mcp_internal_actions/instructions.ts
```diff
@@ -44,30 +44,6 @@ The output includes:
...
-      3.  **Execute Read Query:** Use \`get_issues\` to retrieve data using JQL. Construct your JQL queries based on the information obtained from \`get_issue_types\` to ensure you are using correct field and relationship names.
-
-      **Best Practices for Querying:**
-      1.  **Discover Object Structure First:** Use \`get_issue_create_fields\` to understand fields available for create/update in a given project and issue type. Alternatively, for a quick field list directly in a query, use \`get_issues\`.
```
thervier, line 56: to be moved in get_issue_create_fields tool description?

front/lib/actions/mcp_internal_actions/instructions.ts
flvndvd: So this one was actually wrong, it says "Use get_issues to retrieve data using JQL," but `get_issues` takes structured filters. JQL lives in `get_issues_using_jql`.

front/lib/actions/mcp_internal_actions/instructions.ts
flvndvd: `get_issue_create_fields` already says:
> "Create/update-time field metadata for a specific project and issue type. Returns only fields available on the Create/Update screens (subset). Use get_issue_types to get the issue type ID."

front/lib/actions/mcp_internal_actions/instructions.ts
flvndvd: Same for the `get_users` description:
>       "Search for JIRA users. Provide emailAddress for exact email match, or name for display name contains. If neither is provided, returns the first maxResults users. Use startAt for pagination (pass the previous result's nextStartAt).",

thervier (approved):
Thank you!

PR #27929, opened 2026-06-24
[Skills] Allow enabling skills from earlier user messages

philipperolet:
## Description

Follows https://github.com/dust-tt/dust/pull/27754.

That PR allowed `enable_skill` to enable skills explicitly tagged in the current user message. A skill tag from an earlier visible user message can still represent user intent for the conversation, but the handler did not include those tags in the allow-list. In that case, the agent could see or refer back to `<skill id="..." name="..." />` from the conversation and still fail to enable the skill by name.

This extends the same user-message allow-list to visible user messages at or before the current user message rank. It keeps the current-user-message path, ignores later messages, and still resolves tagged skills through the agent-loop-aware active-skill fetch.

The PR also removes the redundant `listActiveByNameForAgentLoop` helper from the previous PR and lets `fetchActiveByName` receive optional agent-loop context.

## Risks

Blast radius: skill activation through the `skill_management` MCP tool.
Risk: low

## Deploy Plan
- pmrr
- deploy front

## Test

- [x] `NODE_ENV=test npm test lib/api/actions/servers/skill_management/tools/index.test.ts`

### DIFF
--- a/front/lib/api/actions/servers/skill_management/tools/index.test.ts
+++ b/front/lib/api/actions/servers/skill_management/tools/index.test.ts
@@ -40,8 +40,16 @@ describe("skill_management enable_skill tool", () => {
   const auth = {};
   const agentConfiguration = { sId: "agent-id" };
   const agentMessage = { sId: "agent-message-id" };
-  const conversation = { sId: "conversation-id" };
-  const userMessage = { content: "", sId: "user-message-id" };
+  const userMessage = {
+    content: "",
+    sId: "user-message-id",
+    type: "user_message",
+    visibility: "visible",
+  };
+  const conversation: {
+    content: (typeof userMessage)[][];
+    sId: string;
+  } = { content: [], sId: "conversation-id" };
   const skill = {
     enableForAgent: mockEnableForAgent,
     getFileAttachments: mockGetFileAttachments,
@@ -75,8 +83,10 @@ describe("skill_management enable_skill tool", () => {
   });
 
   function makeExtra({
+    conversationOverride = conversation,
     userMessageOverride = userMessage,
   }: {
+    conversationOverride?: typeof conversation;
     userMessageOverride?: typeof userMessage;
   } = {}) {
     return {
@@ -85,7 +95,7 @@ describe("skill_management enable_skill tool", () => {
         runContext: {
           agentConfiguration,
           agentMessage,
-          conversation,
+          conversation: conversationOverride,
           userMessage: userMessageOverride,
         },
       },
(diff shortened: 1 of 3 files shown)


front/lib/api/actions/servers/skill_management/tools/index.ts
```diff
@@ -22,9 +26,42 @@ async function findAvailableSkillForAgentLoop({
 }): Promise<SkillResource | null> {
   const { enabledSkills, equippedSkills, systemSkills } =
     await SkillResource.listForAgentLoop(auth, agentLoopData);
+  const userMessageSkillIds = new Set(
```
philipperolet, line 18: can you extract that part into a function, extractSkillIdsFromConversationMessages?
also, what is the product reason to stop at latest compaction, vs just adding all messages?
if no reason good enough, remove that part

philipperolet: Addressed.

- Extracted the conversation scan into `extractSkillIdsFromConversationMessages`.
- Removed the compaction cutoff. Product-wise, compaction is internal context management; an explicit user skill tag remains user intent in the conversation.
- Kept the rank guard so later messages cannot authorize the current run.
- Folded in Aubin’s previous comment by removing `listActiveByNameForAgentLoop` and adding optional agent-loop context to `fetchActiveByName`.

front/lib/api/actions/servers/skill_management/tools/index.test.ts
```diff
@@ -28,20 +28,45 @@ vi.mock("@app/lib/api/skills/conversation_files", () => ({
...
+    type: "user_message";
+    visibility: "visible";
+  };
+  type TestCompactionMessage = {
+    content: string | null;
+    sId: string;
+    status: "succeeded";
+    type: "compaction_message";
+    visibility: "visible";
+  };
```
aubin-tchoi, line 53: OOC why not reuse `UserMessageType`

front/lib/api/actions/servers/skill_management/tools/index.test.ts
```diff
@@ -6,17 +6,17 @@ import { beforeEach, describe, expect, it, vi } from "vitest";
...
+  mockFetchActiveByName,
   mockFetchActiveByIdsForAgentLoop,
   mockGetFileAttachments,
-  mockListActiveByNameForAgentLoop,
   mockListForAgentLoop,
   mockLoadSkillFilesToConversation,
 } = vi.hoisted(() => ({
   mockEnableForAgent: vi.fn(),
   mockBatchFetchUsedBySkills: vi.fn(),
+  mockFetchActiveByName: vi.fn(),
```
aubin-tchoi, line 17: OOC why mock this?

aubin-tchoi (commented):
LGTM

PR #27930, opened 2026-06-24
[Skills] Drop deprecated skill response fields

philipperolet:
## Description
Related to https://github.com/dust-tt/tasks/issues/8616. Follows https://github.com/dust-tt/dust/pull/27781 and https://github.com/dust-tt/dust/pull/27899.

Removes the deprecated skill response compatibility fields `isExtendable` and `extendedSkillId` from `SkillType`, `SkillResource.toJSON()`, and the public OpenAPI schema. These fields were legacy customized-skill compatibility shims and have only returned `false` / `null` since the customized-skill cleanup.

This is a breaking response-shape cleanup for clients still reading those deprecated fields.

## Risks
Blast radius: public skill API responses and internal code typed against `SkillType`.
Risk: standard. The removed fields were deprecated constant values, but API consumers that still read them must stop depending on them.

## Deploy Plan
- Deploy `front`

- ⚠️ For this API change we could wonder about backwards compatibility. IMO we don't need to address it, as it's 1. a cleanup, 2. very small product surface; ifever it breaks a few api calls, we can address directly via support

### DIFF
--- a/front-api/public/swagger.json
+++ b/front-api/public/swagger.json
@@ -13776,20 +13776,9 @@
             "type": "boolean",
             "description": "Whether the authenticated actor can edit the skill"
           },
-          "isExtendable": {
-            "type": "boolean",
-            "deprecated": true,
-            "description": "Deprecated compatibility field. Always false."
-          },
           "isDefault": {
             "type": "boolean",
             "description": "Whether this skill is enabled by default"
-          },
-          "extendedSkillId": {
-            "type": "string",
-            "nullable": true,
-            "deprecated": true,
-            "description": "Deprecated compatibility field. Always null."
           },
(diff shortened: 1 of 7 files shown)


flvndvd (approved):

flvndvd:
Might be good to confirm that old versions of the SDK won't break on type validation.

philipperolet:
> Might be good to confirm that old versions of the SDK won't break on type validation.

Good point. checked @dust-tt/client@1.0.57, 1.1.22, and 1.2.6 tarballs. None contain isExtendable, extendedSkillId, /skills, or Skill response handling in the SDK source/bundles. So for the official JS SDK, we're good

PR #27948, opened 2026-06-25
Add seats info in poke

tdraier:
## Description

Shows seat commitments (min seats, max seats) and current seat usage in the poke admin UI, under the Subscriptions tab.

**What's added:**
- `seatPlan` field on `PokeWorkspaceInfo` — fetched in parallel with the members count, best-effort (returns `null` for non-Metronome or non-seat-billed workspaces)
- `SeatCommitmentsSection` component in the subscription card: one row per seat type showing `{name} | {min} min / {max} max / {used} used`
- Section is hidden when there are no seat types (free plans, MAU contracts, etc.)

## Tests

Manually verified type-check passes. The `getSeatPlan` function is already unit-tested; this change only adds a display layer on top of existing logic.

## Risk

Low. Read-only display change. `getSeatPlan` failures degrade gracefully to `null` (section hidden), so no impact on workspaces without a seat-billed Metronome contract.

## Deploy Plan

### DIFF
--- a/front/components/poke/pages/WorkspacePage.tsx
+++ b/front/components/poke/pages/WorkspacePage.tsx
@@ -141,6 +141,7 @@ export function WorkspacePage() {
     defaultAlerts,
     programmaticCreditState,
     programmaticWarningReached,
+    seatPlan,
     stripeSubscription,
     stripeCustomerId,
     subscriptions,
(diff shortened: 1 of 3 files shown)


tdraier:

- #27948 ◀

thervier (approved):
LGTM

PR #27949, opened 2026-06-25
Improvements and fixes in switch_contract

tdraier:
## Description

Significant refactor and bug fixes to the `switch_contract` poke flow.

### `switch_contract.ts` — refactor

- **Extracted all steps into module-level functions**, split into two groups:
  - **Pre-provision** (hard failures, abort cleanly before any Metronome state exists): eligibility check, Stripe customer resolution, Metronome customer resolution, package validation, swap timing, seat billing floors, WorkOS org provisioning, PAYG/credit config persistence, cancel existing pending contract.
  - **Post-provision** (best-effort, failures collected as warnings): contract edits, seat remap, seat sync, pending subscription creation, Stripe cancellation scheduling.
- **Single Metronome edit call**: `stepNetPaymentTerms`, `stepInitialCredits`, and `stepSeatConfiguration` (commits + overrides) are merged into one `stepContractEdits` — a single `v2.contracts.edit` call instead of three.
- **`persistCreditConfig`** (renamed from `stepPaygConfig`): now writes all `CreditUsageConfigurationResource` fields (discount, PAYG, usage cap, balance threshold, pool cap, programmatic cap, auto-upgrade, top-up, invoice finalization) before provisioning.
- **Post-provision error handling**: all post-provision failures are collected as warnings; if any exist, returns `provision_inconsistent` with a bulleted list instead of `Ok`.
- **Stripe `cancel_at` fix**: backdated contracts have `alignedStart` in the past; Stripe rejects past `cancel_at`. Now uses `max(alignedStart, now + 60s)` so the subscription is cancelled as soon as possible.
- **Stripe cancellation is best-effort**: failure no longer rolls back the contract — it surfaces as a warning.
- Removed `provision_inconsistent` from pre-provision error paths (no Metronome state to roll back at that point). Removed dead `dispatchPaygEnabled/Disabled` calls (the `contract.start` webhook already calls `syncPoolCreditStateFromBalance`).

### `SwitchContractDialog.tsx` — UI improvements

Screenshot 2026-06-25 at 10 58 21

Screenshot 2026-06-25 at 10 58 39

- **Credit configuration section**: all `CreditUsageConfigurationResource` fields (previously only in the separate "Manage Credit Usage Configuration" plugin) are now inline at the end of the form.
- **Two-column grid layout** throughout: label left, field right — applied to the top fields, initial credits section, and credit configuration section.
- **Commitment price auto-fill**: if left blank, defaults to `minSeats × rate` on submit. The placeholder shows the calculated value dynamically.
- **Annual seat monthly hint**: for `_yearly` seat types, shows `≈ X/mo` below the rate field.
- **Period defaults on frequency change**: switching a payment schedule from one-time to monthly/quarterly/semi-annually/annually auto-fills the periods field (12/4/2/1). Uses `form.watch(callback)` for reliable per-field change detection (fixes a previous ref-based approach that didn't work).
- **Number field clear fix** (`fields.tsx`): clearing a `type="number"` input now emits `undefined` instead of `0`, so the placeholder reappears.

## Tests

Tested manually via the Switch Contract dialog in the poke UI.

## Risk

Low — changes are scoped to the operator poke interface (`switch_contract` endpoint + dialog). The Metronome contract provisioning logic is unchanged; only the surrounding orchestration and UI are affected.

Backdated `cancel_at` fix is new behavior: previously the whole switch failed on backdated contracts; now it succeeds and schedules Stripe cancellation at `now + 60s`.

## Deploy Plan

Standard deploy. No migration needed.

### DIFF
--- a/front/components/poke/subscriptions/SwitchContractDialog.tsx
+++ b/front/components/poke/subscriptions/SwitchContractDialog.tsx
@@ -20,6 +20,7 @@ import {
   usePokePlans,
   usePokeStripeCustomerCurrency,
 } from "@app/lib/swr/poke";
+import { usePokePluginAsyncArgs } from "@app/poke/swr/plugins";
 import assert from "@app/lib/utils/assert";
 import { isCreditPricedPlan } from "@app/types/plan";
 import type { WorkspaceType } from "@app/types/user";
@@ -39,7 +40,7 @@ import {
   Spinner,
 } from "@dust-tt/sparkle";
 import { zodResolver } from "@hookform/resolvers/zod";
-import { useCallback, useEffect, useMemo, useState } from "react";
+import { useCallback, useEffect, useMemo, useRef, useState } from "react";
 import { useForm } from "react-hook-form";
 import { z } from "zod";
 
@@ -70,12 +71,20 @@ const SwitchContractFormSchema = z
       .min(0, "Net payment terms must be ≥ 0")
       .max(365, "Net payment terms must be ≤ 365")
       .optional(),
+    // Credit usage configuration.
     paygEnabled: z.boolean().default(false),
     usageCapCredits: z
       .number()
       .int("Usage cap must be an integer number of credits")
       .min(1, "Usage cap must be at least 1 credit")
       .optional(),
+    defaultDiscountPercent: z.number().int().min(0).max(100).default(0),
+    balanceThresholdCredits: z.number().int().min(0).optional(),
+    defaultPoolCapCredits: z.number().int().min(0).optional(),
+    programmaticMonthlyCapCredits: z.number().int().min(0).optional(),
+    autoSeatUpgradeEnabled: z.boolean().default(false),
+    topUpEnabled: z.boolean().default(false),
+    autoInvoiceFinalizationEnabled: z.boolean().default(true),
     // One-off initial credits (contract-level prepaid commit). Toggled on via
     // `showInitialCredits`; both fields are then required together and assembled
     // into `initialCredits` on submit.
@@ -165,6 +174,14 @@ function snapDatetimeLocalToHour(value: string): string {
 
 const isLegacyPackageName = (name: string) => /\blegacy\b/i.test(name);
 
+const DEFAULT_PERIODS_FOR_FREQUENCY: Record<string, number | undefined> = {
+  monthly: 12,
+  quarterly: 4,
+  semi_annually: 2,
+  annually: 1,
+  one_time: undefined,
+};
+
 interface SwitchContractDialogProps {
   owner: WorkspaceType;
   stripeCustomerId: string | null;
@@ -228,6 +245,13 @@ export default function SwitchContractDialog({
       netPaymentTermsDays: undefined,
       paygEnabled: false,
       usageCapCredits: undefined,
+      defaultDiscountPercent: 0,
+      balanceThresholdCredits: undefined,
+      defaultPoolCapCredits: undefined,
+      programmaticMonthlyCapCredits: undefined,
+      autoSeatUpgradeEnabled: false,
+      topUpEnabled: false,
+      autoInvoiceFinalizationEnabled: true,
       showInitialCredits: false,
       initialCreditsAmount: undefined,
       initialCreditsInvoiceAmount: undefined,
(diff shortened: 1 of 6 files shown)


tdraier:

- #27949 ◀

front/lib/api/poke/switch_contract.ts
```diff
@@ -562,462 +518,601 @@ export async function switchContract({
...
+    defaultDiscountPercent: body.defaultDiscountPercent,
+    paygEnabled: body.paygEnabled,
+    usageCapCredits: body.usageCapCredits ?? null,
+    balanceThresholdAwuCredits: body.balanceThresholdCredits ?? null,
+    defaultPoolCapAwuCredits: body.defaultPoolCapCredits ?? null,
+    programmaticMonthlyCapAwuCredits: body.programmaticMonthlyCapCredits ?? null,
+    autoSeatUpgradeEnabled: body.autoSeatUpgradeEnabled,
+    topUpEnabled: body.topUpEnabled,
+    autoInvoiceFinalizationEnabled: body.autoInvoiceFinalizationEnabled,
+  };
```
rfrenoy, line 652: this means we will override many config elements with default values, is that intended?

front/lib/api/poke/switch_contract.ts
```diff
@@ -562,462 +518,601 @@ export async function switchContract({
...
-      priority: AWU_PRIORITY_PURCHASED_COMMIT,
-      applicableProductTags: ["usage"],
+    addCommits.push({
+      product_id: getProductPrepaidCommitId(),
+      type: "PREPAID",
       name: `Initial credits: ${body.initialCredits.amountCredits.toLocaleString()} credits`,
-      // Scope the key to the freshly provisioned contract: a retroactive
-      // start can otherwise collide with a prior switch that shared the same
-      // workspace, start moment, and amount.
-      uniquenessKey: `initial-credits-${metronomeContractId}-${body.initialCredits.amountCredits}`,
```
rfrenoy, line 704: is it ok to drop the uniqueness key?

front/lib/api/poke/switch_contract.ts
tdraier: good catch we should keep previous values if they were set

front/lib/api/poke/switch_contract.ts
tdraier: I would say yes, since now it's one single edit

front/lib/api/poke/switch_contract.ts
tdraier: now loading the previosu value

front/lib/api/poke/switch_contract.ts
```diff
@@ -562,462 +522,641 @@ export async function switchContract({
...
-          "provision_inconsistent",
-          `Provisioned Metronome contract ${metronomeContractId} but failed to ` +
-            `set net payment terms to ${body.netPaymentTermsDays} days: ` +
-            `${netTermsResult.error.message}. Manual cleanup may be required.`
+          "payg_config_failed",
+          `Failed to create credit configuration: ${result.error.message}`
         )
       );
     }
   }
```
rfrenoy, line 676: here we store the config in the db but we don't call Metronome. Where/when is the config synced?

front/lib/api/poke/switch_contract.ts
```diff
@@ -562,462 +522,641 @@ export async function switchContract({
...
-    additionalCustomFields: body.hubspotDealId
-      ? { [HUBSPOT_DEAL_ID_CUSTOM_FIELD_KEY]: body.hubspotDealId }
-      : undefined,
-  });
-  if (provisionResult.isErr()) {
+// Ensure the workspace has a WorkOS organization for any paid tier.
+async function ensureWorkOSOrg(
+  ownerLight: LightWorkspaceType,
+  pkgTier: MetronomePackageTier,
+  workspaceId: string
```
rfrenoy: we can drop that arg?

front/components/poke/subscriptions/SwitchContractDialog.tsx
```diff
@@ -248,6 +272,79 @@ export default function SwitchContractDialog({
...
+    form.setValue(
+      "topUpEnabled",
+      typeof c.topUpEnabled === "boolean" ? c.topUpEnabled : false
+    );
+    form.setValue(
+      "autoInvoiceFinalizationEnabled",
+      typeof c.autoInvoiceFinalizationEnabled === "boolean"
+        ? c.autoInvoiceFinalizationEnabled
+        : true
+    );
```
rfrenoy: nit: we could use a zod schema?

rfrenoy (approved):
One concern about Metronome sync, and two nits, otherwise LGTM

front/lib/api/poke/switch_contract.ts
tdraier: All sync happen on `contract.start` webhook

rfrenoy: ok, and reading the webhook, we do that before provisioning the contract so I guess there is no risk of race condition (at least as long as we keep calls in this order). So fine by me 👍

tdraier: yes all config is stored before provisioning contract

PR #27950, opened 2026-06-25
Preserve Snowflake selections on read-only failures

fontanierh:
## Description

Fixes https://github.com/dust-tt/tasks/issues/9137.

Snowflake `NOT_READONLY` failures used the remote database garbage collector with an empty tree, which removed both Core-visible tables and the connector rows storing admin selections. This keeps the cleanup of Core content, preserves selected DB/schema/table rows for the next successful sync, and pauses the connector with `remote_database_connection_not_readonly` until grants are fixed.

## Tests

Added regression coverage for preserving selected permissions during no-tree garbage collection and for pausing Snowflake connectors on read-only failures.

Ran locally:
- `NODE_ENV=test CONNECTORS_DATABASE_URI=postgres://test:test@localhost:5433/dust_connectors_test_issue_9137_20260625 npx vitest run src/lib/remote_databases/activities.test.ts src/lib/temporal_monitoring.test.ts`
- `npx tsgo --noEmit`

## Risk

Worst case, selected Snowflake permissions are retained when the remote object was intentionally removed. The next successful sync still reconciles the catalog and recreates only selected reachable content. This does not restore selections that were already deleted before this deploy; those still need backup/log-based recovery or manual reselection. Safe to rollback.

## Deploy Plan

Standard deploy.

### DIFF
--- a/connectors/src/lib/remote_databases/activities.ts
+++ b/connectors/src/lib/remote_databases/activities.ts
@@ -21,6 +21,17 @@ import logger from "@connectors/logger/logger";
 import type { ConnectorResource } from "@connectors/resources/connector_resource";
 import type { DataSourceConfig, INTERNAL_MIME_TYPES } from "@connectors/types";
 
+type RemoteDatabaseHierarchyModel =
+  | RemoteDatabaseModel
+  | RemoteSchemaModel
+  | RemoteTableModel;
+
+type CleanupUnusedRemoteDatabaseHierarchyModel = {
+  destroy: () => Promise<unknown>;
+  permission: RemoteDatabaseHierarchyModel["permission"];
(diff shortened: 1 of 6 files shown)


connectors/src/lib/remote_databases/activities.ts
```diff
@@ -602,9 +611,19 @@ export async function sync({
       dataSourceConfig,
       folderId: unusedSchema.internalId,
     });
-    await unusedSchema.destroy();
 
-    if (unusedSchema.permission === "selected") {
+    if (unusedSchema.permission === "selected" && preserveSelectedPermissions) {
```
flvndvd: `nit` could be factorized.

flvndvd (approved):
Thank you!

PR #27951, opened 2026-06-25
fix(poke): raise negotiated free credits cap to $5,000

rfrenoy:
## Description

Bump MAX_FREE_CREDITS_DOLLARS from $2,000 to $5,000 so contracts with higher monthly API credit allowances can be provisioned via the Poke Upgrade/Downgrade flow.

Ticket: https://github.com/dust-tt/tasks/issues/9220


thervier (approved):
LGTM

PR #27954, opened 2026-06-25
[front] fix: align MarkdownFilePreview tests with double-click handler

davidebbo:
## Problem

The "switches to edit mode when clicking preview content" test was failing:

```
AssertionError: expected "vi.fn()" to be called with arguments: [ 'edit' ]
Number of calls: 0
```

Root cause: `MarkdownFilePreview` binds its edit-mode handler to `onDoubleClick`, but the test fired a single `click`, so the handler never ran. The sibling "does not switch to edit mode when clicking a link" test was passing vacuously for the same reason — a single click can never trigger edit mode.

## Fix

Fire `doubleClick` in both tests (and rename them to "double-clicking") to match the component's intentional double-click behavior. The link test now genuinely verifies that double-clicking a link does not switch to edit mode.

## Test plan

- [x] `NODE_ENV=test npm run test -- components/file_explorer/MarkdownFilePreview.test.tsx` — 6 passed

### DIFF
--- a/front/components/file_explorer/MarkdownFilePreview.test.tsx
+++ b/front/components/file_explorer/MarkdownFilePreview.test.tsx
@@ -72,7 +72,7 @@ describe("MarkdownFilePreview", () => {
     expect(link).toHaveAttribute("rel", "noopener noreferrer");
   });
 
-  it("switches to edit mode when clicking preview content", () => {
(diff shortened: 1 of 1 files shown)


Fraggle (approved):
I did switch to double click, totally reasonnable.

PR #27957, opened 2026-06-25
[front] chore: move Productboard instructions to custom skill

aubin-tchoi:
## Description

- Removes the inline Productboard MCP server instructions.
- Adds a backfill script that adds the skill `Work with Productboard` and updates agents that were using the skill.
- The editors of the skill are derived from the agents that use it.

## Tests

- Tested locally.

## Risk

- Low.

## Deploy Plan

- Deploy front.
- Run backfill.


front/lib/api/actions/servers/productboard/metadata.ts
```diff
@@ -369,151 +369,7 @@ export const PRODUCTBOARD_TOOLS_METADATA = createToolsRecord({
   },
 });
 
-export const PRODUCTBOARD_SERVER_INSTRUCTIONS = `
```
flvndvd, line 372: Just checking, none of this had to be ported to the tools despcription?

front/migrations/20260625_backfill_productboard_skill.ts
```diff
@@ -0,0 +1,529 @@
...
+  await AgentSkillModel.bulkCreate(
+    productboardAgentModelIds.map((agentConfigurationModelId) => ({
+      agentConfigurationId: agentConfigurationModelId,
+      customSkillId: skill.id,
+      globalSkillId: null,
+      workspaceId: workspace.id,
+    }))
+  );
+
+  logger.info(
```
flvndvd, line 513: Aren't we missing to replace the actual tool by the skill directly in their agents?

aubin-tchoi: ah I missed the removal of the tools indeed, fixing

aubin-tchoi: done

front/lib/api/actions/servers/productboard/metadata.ts
aubin-tchoi: we could try to also make these instructions leaner and move part of them in tools descriptions, but my goal here is really to stay iso once the skill is enabled

flvndvd (approved):

PR #27958, opened 2026-06-25
Improve google_drive tool descriptions for tool-search recall

flvndvd:
Tightens the `google_drive` MCP tool descriptions so they retrieve well under BM25 tool-search (which indexes tool name, description, and input schema only, not server-level text).

Each edited tool now carries its platform token and the action verbs users actually type:
- `list_drives`: add "Google Drive" (was untagged, collided with microsoft_drive)
- `search_files`: add find/locate + document/spreadsheet/presentation/folder nouns
- `get_file_content`: add read/open
- `copy_file`: add clone/duplicate
- `revoke_file_sharing`: add unshare/stop sharing

Validated with a local BM25 harness over realistic queries: top-1 accuracy 12/18 → 16/18, 5 improved, 0 regressed. String-only change, no behavior change.

### DIFF
--- a/front/lib/api/actions/servers/google_drive/metadata.ts
+++ b/front/lib/api/actions/servers/google_drive/metadata.ts
@@ -26,7 +26,7 @@ const capabilitiesSchema = z
 
 export const GOOGLE_DRIVE_TOOLS_METADATA = createToolsRecord({
   list_drives: {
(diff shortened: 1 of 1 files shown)


adrsimon (approved):
LGTM

PR #27959, opened 2026-06-25
Add missing indices

tdraier:
## Summary

Add missing leading-field indexes on FK columns that were causing `deleteSpacesActivity` to take 10+ minutes in production.

**Root cause:** `DELETE FROM vaults WHERE id = ?` and `DELETE FROM mcp_server_views WHERE workspaceId = ? AND id = ?` each triggered Postgres FK constraint scans on referencing tables that lacked a leading-field index on the FK column — forcing full sequential scans even after all rows had already been cleaned up.

**Slow queries observed in APM (4–8s each):**
- `DELETE FROM vaults` → Postgres scanned `conversations` (14M rows) to verify `spaceId` FK — only had `(workspaceId, spaceId)`, not usable for `WHERE spaceId = ?`
- `DELETE FROM mcp_server_views` → Postgres scanned `conversation_mcp_server_views` and `skill_mcp_server_configurations` on `mcpServerViewId` — no index at all

**Fixes:**
- `conversations.spaceId` — new index (14M rows, built `CONCURRENTLY`)
- `conversation_mcp_server_views.mcpServerViewId` — new index (`CONCURRENTLY`)
- `skill_mcp_server_configurations.mcpServerViewId` — new index (`CONCURRENTLY`)

## Deploy Plan

Pre-deploy migrations only. All three indexes use `CONCURRENTLY` so no table locks.

### DIFF
--- a/front/lib/models/agent/actions/conversation_mcp_server_view.ts
+++ b/front/lib/models/agent/actions/conversation_mcp_server_view.ts
@@ -105,6 +105,10 @@ ConversationMCPServerViewModel.init(
         name: "conversation_mcp_server_views_conversation_id",
         concurrently: true,
       },
+      {
+        fields: ["mcpServerViewId"],
+        concurrently: true,
+      },
     ],
   }
 );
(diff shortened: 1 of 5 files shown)


tdraier:

- #27959 ◀

flvndvd (approved):

PR #27960, opened 2026-06-25
Fix: better scroll handling on mobile for main page and conversations.

Fraggle:
## Description

**Everything is scoped to mobile only.**

Fixes mobile scroll by switching from a fixed-height viewport model to native document/window scroll on mobile. Updated only the main and conversation pages, can extends the approach.

Notice below how the content uses the whole screen.

Before

https://github.com/user-attachments/assets/f4bdc8fa-3eed-4c36-a821-3d2c74fd08b3

After

https://github.com/user-attachments/assets/cf5197a4-8cec-4328-af7a-d24fbfc95683

## Tests

Tested manually on mobile. No automated tests — pure layout/scroll behaviour change.

## Risk

Low. All changes are gated on `isMobile` (`useIsMobile` hook) and the `document-scroll-mode` CSS class, leaving desktop layout untouched. Rollback is a single front deploy revert.

## Deploy Plan

Deploy `front`.

### DIFF
--- a/front/styles/global.css
+++ b/front/styles/global.css
@@ -64,11 +64,11 @@ body {
 
 /*
  * Mobile document scroll (body.document-scroll-mode via useDocumentScrollMode).
- * Unlocks height chain, document/window scroll, panel background through full range.
+ * Top-level height chain + panel background. Component layout overrides live in
+ * MOBILE_DOCUMENT_SCROLL_CLASSES (see AppContentLayout, DropzoneContainer).
  */
 html:has(body.document-scroll-mode),
 html:has(body.document-scroll-mode) body,
-html:has(body.document-scroll-mode) #__next,
 html:has(body.document-scroll-mode) #root {
   height: auto;
   min-height: 100dvh;
@@ -102,36 +102,6 @@ body.document-scroll-mode #root {
   width: 100%;
 }
 
-body.document-scroll-mode .dropzone-container {
-  height: auto;
-  min-height: 0;
-}
-
-body.document-scroll-mode .app-content-root {
-  height: auto;
-  min-height: 100dvh;
-  width: 100%;
-}
-
-body.document-scroll-mode .app-content-row {
-  flex: 0 1 auto;
-  min-height: auto;
-  width: 100%;
-  max-width: 100%;
-  overflow-x: clip;
-}
-
-body.document-scroll-mode .app-content-main {
-  height: auto;
-  min-height: var(--panel-height);
-  flex: 0 1 auto;
-  width: 100%;
-  max-width: 100%;
-  /* Contain wide message content; table scroll is handled via #BlockWrapper rules below */
-  overflow-x: clip;
-  background-color: transparent;
-}
-
 /*
  * Markdown tables (TableBlock) use ScrollArea (overflow-hidden) + sticky copy
  * actions (ContentBlockWrapper). Nested scroll + sticky inside document scroll
@@ -150,16 +120,6 @@ body.document-scroll-mode #BlockWrapper .s-overflow-hidden {
 body.document-scroll-mode #BlockWrapper .s-sticky {
   position: relative;
   top: auto;
-}
-
-body.document-scroll-mode .app-content-pane,
-body.document-scroll-mode .app-content-scroll {
-  height: auto;
-  flex: 0 1 auto;
-  width: 100%;
-  max-width: 100%;
-  overflow-x: clip;
-  overflow-y: visible;
 }
 
 #__next {
--- a/front/styles/global.css
+++ b/front/styles/global.css
@@ -62,6 +62,106 @@ body {
   overscroll-behavior: none;
 }
 
+/*
+ * Mobile document scroll (body.document-scroll-mode via useDocumentScrollMode).
+ * Unlocks height chain, document/window scroll, panel background through full range.
+ */
+html:has(body.document-scroll-mode),
+html:has(body.document-scroll-mode) body,
+html:has(body.document-scroll-mode) #__next,
+html:has(body.document-scroll-mode) #root {
+  height: auto;
+  min-height: 100dvh;
+}
+
+html:has(body.document-scroll-mode),
+html:has(body.document-scroll-mode) #__next,
+body.document-scroll-mode,
+body.document-scroll-mode #root {
+  @apply bg-panel-background;
+}
+
+html.dark:has(body.document-scroll-mode),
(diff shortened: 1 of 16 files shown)


Fraggle:

- #27960 ◀

front/styles/global.css
```diff
@@ -62,6 +62,106 @@ body {
...
+  overflow-y: visible;
+}
+
+body.document-scroll-mode #BlockWrapper .s-sticky {
+  position: relative;
+  top: auto;
+}
+
+body.document-scroll-mode .app-content-pane,
+body.document-scroll-mode .app-content-scroll {
```
ykmsd: why we have different name when the styles are the same?

ykmsd: I assume this is to avoid duplication we can probably have const for this and apply everywhere?

front/styles/global.css
```diff
@@ -62,6 +62,106 @@ body {
   overscroll-behavior: none;
 }
 
+/*
+ * Mobile document scroll (body.document-scroll-mode via useDocumentScrollMode).
+ * Unlocks height chain, document/window scroll, panel background through full range.
+ */
+html:has(body.document-scroll-mode),
+html:has(body.document-scroll-mode) body,
+html:has(body.document-scroll-mode) #__next,
```
ykmsd: is it intentional to have one for #__next?

front/styles/global.css
```diff
@@ -62,6 +62,106 @@ body {
...
+html:has(body.document-scroll-mode) #root {
+  height: auto;
+  min-height: 100dvh;
+}
+
+html:has(body.document-scroll-mode),
+html:has(body.document-scroll-mode) #__next,
+body.document-scroll-mode,
+body.document-scroll-mode #root {
+  @apply bg-panel-background;
```
ykmsd, line 81: ideally we want to have css variable for bg-app-background which conditionally change the color based on the viewport size so we don't have to define like this, but I will do it in the future

front/styles/global.css
```diff
@@ -62,6 +62,106 @@ body {
...
+  overflow-x: clip;
+  background-color: transparent;
+}
+
+/*
+ * Markdown tables (TableBlock) use ScrollArea (overflow-hidden) + sticky copy
+ * actions (ContentBlockWrapper). Nested scroll + sticky inside document scroll
+ * clips content and breaks iOS window scroll once the table leaves the viewport.
+ */
+body.document-scroll-mode #BlockWrapper {
```
ykmsd: can't we pass it as isMobile props and conditionally change the classnames?

front/styles/global.css
```diff
@@ -62,6 +62,106 @@ body {
...
+}
+
+body.document-scroll-mode #root {
+  flex: 1 1 auto;
+  display: flex;
+  flex-direction: column;
+  width: 100%;
+}
+
+body.document-scroll-mode .dropzone-container {
```
ykmsd: I'm not a big fan of using global css to override the mobile style. Global css is supposed to be something for generic (and for overriding styles from external components we don't have control). For example where we have `app-content-pane`, we have tailwind classes too like this: 
```
<div className="app-content-pane flex min-h-0 flex-1 flex-col h-panel overflow-y-auto">
```
and when you look at this code you cannot tell which classnames will be applied in the end. Even if you come to check out global.css file, it's still hard to know which ones will be applied (some styles are conflicting, I assume global css will win though)

If .document-scroll-mode = isMobile, can't we conditionally apply classnames in each component (something like isMobile ? "h-auto ..." : "h-full...")? I guess there is no way to do that for top level nodes like html body, but for components it would be clearer what styles we want to apply for different scenario?

front/styles/global.css
Fraggle: Nope, thanks for catching.

front/styles/global.css
Fraggle: Yepp.

front/styles/global.css
Fraggle: It's a component within Sparkle, used by multiple other compoenents.
Passing isMobile in all the call sites is not a good idea imho.

Alternatively, I could add useIsMobile to sparkle but it doesn't feel appropriate as I think sparkle should not have anything specific to our decision regarding mobile threshold etc..

Any other idea ?

front/styles/global.css
Fraggle: I see your POV of not overloading global.css however we already have custom css rules and animations there. I can create a separate file but I don't really see the real value ?

On injecting the class on the fly, as you mention, it doesn't work for top level nodes so it would be a mix anyway. But it's also already a mix so I guess I can do it on the fly for .dropzone-container .app-content-root & .app-content-row

front/styles/global.css
Fraggle: Will inline.

front/styles/global.css
ykmsd: Is there no way to pass a custom classname to the component to accomplish the same result? Because if we ever refactor BlockWrapper I would never expect that there is another styling contract in global.css that can break things

If there is really no way I'm fine with

Fraggle: Look at the code and let me know, feels to me that it's worth to pass down the classname down multiple components.

Maybe it's just fine to change the global styling of the affected component to this one (so, no override) but it's hard to guarantee I'll not break anything (even if i test thorougly). Might be the best option 'tho.

front/styles/global.css
ykmsd: It's okay to mix if there is no option and you don't need to create another file for sure. But if there is an option I would rather have it inside the  component so you can see everything together in once place, and easy to refactor. If I just look at `app-content-root`, I would rather think it's a semantic classname we define in tailwind, I would never guess that this is used for mobile only. 

sidenote: animation class names etc should be actually defined as custom class names in tailwind 🙃

ykmsd (approved):
I approve since I don't want to block you, I don't agree with the implementation but the overall change is a really cool improvement so thank you!

Fraggle:
@ykmsd I just pushed a few commits adressing your comments. I'm still looking if applying globally the #BlockWrapper changes is fine on desktop.

Fraggle:
@ykmsd this one in particular https://github.com/dust-tt/dust/pull/27960/changes/86d85258e5b6aaf00c055191eeb91b9cf834fc8f

ykmsd:
👍 Probably we can improve but I don't have super clean solution in my mind right now, but I thiiiiink at least it's easier to refactor than the first version? ty!

PR #27962, opened 2026-06-25
Use separate tool to list attachments

tdraier:
## Description

Fixes https://github.com/dust-tt/tasks/issues/8923

Outlook's `get_attachments` tool was silently failing for attachments, reporting them as "too large to process". Two bugs:

1. **`handleBase64Upload` size limit was too small for non-image files.** The function used `FILE_OFFLOAD_IMAGE_SIZE_BYTES = 2MB` as the cap for all file types. Since base64 adds ~33% overhead, any binary file >~1.5MB was rejected with "The generated file was too large to be stored." A 2MB PDF produces a ~2.67MB base64 string, which exceeded the cap. Fixed by adding `FILE_OFFLOAD_FILE_SIZE_BYTES = 20MB` and using it for non-image types in `handleBase64Upload`.

2. **`get_attachments` downloaded all attachments at once, making it fragile.** For messages with multiple attachments (PDFs, PPTs, etc.) the tool returned all content inline in a single result, which put pressure on context and made it impossible for the agent to select which attachments to retrieve. For attachments >4MB, the Graph API omits `contentBytes` entirely and the old tool silently returned "No content available" with no fallback.

**New tools** (following the Gmail pattern):
- `list_attachments` — returns metadata only (id, name, contentType, size, isInline); excludes inline images and signatures by default. Lets the agent inspect what's there before downloading.
- `get_attachment` — fetches a single attachment by ID. Uses `contentBytes` when present (≤4MB); falls back to the `/$value` binary download endpoint for larger files.

`get_attachments` is kept for backward compatibility but its description now recommends the new pair.

## Tests

Tested manually against the workspace from the original report. Also covered by existing MCP tool type-checking.

## Risk

Low. New tools are additive. The `handleBase64Upload` fix widens an existing size gate from 2MB to 20MB for non-image files — this was always the intended behavior for files (the 2MB constant was named `FILE_OFFLOAD_IMAGE_SIZE_BYTES` and was clearly meant for images only).

## Deploy Plan


tdraier:

- #27962 ◀

Fraggle (approved):
LGTM

PR #27964, opened 2026-06-25
feat(seats): promote none-seat members into spare committed paid seats on contract switch

rfrenoy:
## Description

When a contract switch lands members on `none`, promote them into committed paid seats the new contract still has spare capacity for. Committed seats (minSeats > 0) are billed whether assigned or not, so filling otherwise-empty slots adds no cost and the seat can be reassigned to a new joiner later.

All-or-nothing: either there is enough committed capacity to seat every `none` member (all promoted) or none of them are.

Restrict auto-promotion to the committed pro and workspace tiers (and yearly variants) via PROMOTABLE_SEAT_TYPES. max is excluded: it's a premium seat an admin grants deliberately, and correcting an over-eager auto-grant is a downgrade that only takes effect at the next billing period. So we never auto-assign a seat the admin can't immediately walk back.

## Tests
Unit

## Risk

Low
## Deploy Plan

Front

### DIFF
--- a/front/lib/metronome/seats.test.ts
+++ b/front/lib/metronome/seats.test.ts
@@ -4,6 +4,7 @@ import {
   computeSeatCreditTransfers,
   getSeatCreditNameForSeatType,
   hasContractSeatSubscription,
+  promoteNoneSeatTypesForContract,
   resolveRemappedSeatType,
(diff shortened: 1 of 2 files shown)


thervier (approved):
LGTM

PR #27967, opened 2026-06-25
Introduce `runtimeToolStakeLevelOverride` for dynamic tool stake definition (based on plan for now)

spolu:
## Description

Introduce `runtimeToolStakeLevelOverride` to enable dynamic runtime tool stake overrides (based on `PlanType` as an example).

Motivation is move wake-ups tool stake to low if the user is on a credit-based subscription.

## Tests

N/A will test locally with wake-ups in a follow-up PR. Tested non breaking locally.

## Risk

Low

## Deploy Plan

- deploy `front`

### DIFF
--- a/front/lib/actions/mcp_actions.ts
+++ b/front/lib/actions/mcp_actions.ts
@@ -36,7 +36,6 @@ import {
   getInternalMCPServerNameAndWorkspaceId,
   getInternalMCPServerToolStakes,
   INTERNAL_MCP_SERVERS,
-  type InternalMCPServerNameType,
   resolveInternalMCPServerToolStakeLevel,
 } from "@app/lib/actions/mcp_internal_actions/constants";
 import { findMatchingSubSchemas } from "@app/lib/actions/mcp_internal_actions/input_configuration";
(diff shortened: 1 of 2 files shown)


front/lib/actions/mcp_internal_actions/constants.ts
```diff
@@ -1156,12 +1179,19 @@ type IsRestrictedCallback = (params: {
   isDeepDiveDisabled: boolean;
 }) => boolean;
 
+type RuntimeToolStakeLevelOverride = (params: {
```
Fraggle: Very nit, i think we want "Callback" in the name, like we do for isRestricted ?

front/lib/actions/mcp_internal_actions/constants.ts
```diff
@@ -1396,6 +1426,18 @@ export function getInternalMCPServerToolStakes(
   return server.metadata.tools_stakes;
 }
 
+export function resolveInternalMCPServerToolStakeLevel(
+  name: InternalMCPServerNameType,
+  params: Parameters<RuntimeToolStakeLevelOverride>[0]
```
Fraggle: nit: [0] to get "params" is weird (and a bit flimsy)

front/lib/actions/mcp_actions.ts
```diff
@@ -1262,26 +1264,47 @@ export async function buildToolConfigurationsFromRawTools(
   } = r.value;
 
   const availability = getAvailabilityOfInternalMCPServerById(mcpServerId);
+  const { serverType } = getServerTypeAndIdFromSId(mcpServerId);
```
Fraggle: nit: getInternalMCPServerNameAndWorkspaceId() return Err if the mcpServerId is not internal, you can avoid some loc here.

Fraggle (approved):
Excellent addition !

spolu:
Cool! Will update and merge 👍

PR #27968, opened 2026-06-25
[front] jira/confluence: fix use of extra config with cloud_url

thervier:
## Description

Fixes https://github.com/dust-tt/tasks/issues/8932

* Add getUpdatedExtraConfig to the Jira and Confluence Tools OAuth providers: when a user authenticates via personal_actions with an mcp_server_id, the workspace connection metadata is fetched to populate jira_cloud_url / confluence_cloud_url in the extra config, enabling instance-specific API routing.
* Fix instance URL matching in the Jira and Confluence API helpers: the user-supplied cloud URL is now normalized (trailing slash stripped, https:// added if missing) before comparing it against Atlassian's accessible-resources API response, which always returns canonical URLs. Without this, a trailing slash in the stored URL would cause the find to miss and fall back to the wrong instance.

## Tests

Tested locally with an account that is not the admin that setup the tool => OK

## Risk

Low, jira/confluence only and improved personal connections

## Deploy Plan

Deploy front

### DIFF
--- a/front/lib/api/actions/servers/confluence/helpers.ts
+++ b/front/lib/api/actions/servers/confluence/helpers.ts
@@ -19,6 +19,7 @@ import {
   CreatePagePayloadSchema,
 } from "@app/lib/api/actions/servers/confluence/types";
 import logger from "@app/logger/logger";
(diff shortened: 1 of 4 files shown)


Fraggle (approved):
LGTM

PR #27969, opened 2026-06-25
fix(front): keep credit cost visible with loader while fetching

Nils-Fedrigo:
## Description

On credit-priced plans, the "Credit cost" item in the agent message menu could flicker or pop in late while the per-message cost was still being fetched.

Two fixes:

- Latch the menu-open state (`hasOpenedMenu`) and use it — instead of the live `isMenuOpen` — to gate the cost fetch. This keeps the SWR query enabled after the menu closes, so `refreshedMessage` (which carries `subAgentCostCredits`) isn't dropped back to the prop values during the close animation, removing a flicker of the displayed total.
- Keep the cost section visible with a loading placeholder while the fetch is in flight, rather than having it pop in once the value resolves. `useCreditCostMenuItem` now returns `{ creditCostItem, isCreditPriced }` so the menu can decide to render the section (with a loader) on credit-priced plans even before the cost is available.

Only affects credit-priced plans; no behavior change for other plans (`isCreditPriced` is `false`, section stays hidden).

### DIFF
--- a/front/components/assistant/conversation/AgentMessage.tsx
+++ b/front/components/assistant/conversation/AgentMessage.tsx
@@ -27,7 +27,10 @@ import {
   isUserMessage,
   makeInitialMessageStreamState,
 } from "@app/components/assistant/conversation/types";
-import { useCreditCostMenuItem } from "@app/components/assistant/conversation/useCreditCostMenuItem";
+import {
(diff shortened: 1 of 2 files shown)


Nils-Fedrigo:

- #27969 ◀

rfrenoy (approved):

PR #27970, opened 2026-06-25
Improve inline file preview links

flvndvd:
## Description

This updates `:preview_file` directives to render as themed inline file links while preserving the preview dialog and tooltip behavior. It also fixes alignment with surrounding text, prevents nested buttons, and instructs agents to place file directives naturally within sentences.

image


fontanierh (approved):
LGTM !

PR #27971, opened 2026-06-25
[front] fix: map "none" reasoning effort to "minimal" for gpt-5 models

adrsimon:
## Description

When running behind `use_new_llm_router`, the per-model reasoning schemas for `gpt-5`, `gpt-5-mini` and `gpt-5-nano` rejected a `"none"` reasoning effort outright, throwing a raw `ZodError`:

```
ZodError: reasoning.effort received "none", expected 'minimal' | 'low' | 'medium' | 'high'
```

This crashes any agent that carries a `"none"` effort into one of these models — most commonly a stale value left over from a model swap (an agent previously on a `none`-supporting model, then switched without clearing the effort). The old router never hit this because it mapped `none -> minimal` per-model; the new router validated strictly instead.

These three models' API has no `"none"` level (the next-lowest is `"minimal"`). They are exactly the three models that previously declared a `none -> minimal` mapping, so this PR restores that behavior in the new router: each model now accepts `"none"` as input and maps it to `"minimal"` in its config schema, leaving every other supported effort untouched. `"xhigh"` and the universal `"maximal"` remain input configuration errors.

## Tests

Updated the live endpoint tests for all three models so the `r-none` cases now expect a normal run (mapped to `minimal`) instead of an input configuration error; `r-maximal` stays an input error.

Ran the live suites against the OpenAI API for all three:

```
RUN_LLM_TEST=true npm run test -- --config lib/model_constructors/test/vite.config.js \
  lib/model_constructors/test/endpoints/openai_responses_global_gpt_five.test.ts \
  lib/model_constructors/test/endpoints/openai_responses_global_gpt_five_nano.test.ts \
  lib/model_constructors/test/endpoints/openai_responses_global_gpt_five_mini.test.ts
```

All `r-none` cases pass (including forced-tool and json-schema variants).

## Risk

Low. Scoped to three model config schemas in the new router; no change for any effort other than `none`. Worst case, agents that sent `none` to these models run at `minimal` instead of erroring. Safe to rollback.

## Deploy Plan

- [ ] Deploy `front`

### DIFF
--- a/front/lib/model_constructors/providers/openai/models/gpt_five.ts
+++ b/front/lib/model_constructors/providers/openai/models/gpt_five.ts
@@ -12,15 +12,25 @@ const CONTEXT_SIZE = 400_000;
 const MAX_OUTPUT_TOKENS = 128_000;
 const DEFAULT_REASONING_EFFORT = "medium";
 
-// gpt-5 accepts minimal/low/medium/high. Unlike gpt-5.5 it does NOT support
-// "none" or "xhigh"; the universal "maximal" (maps to xhigh) is unsupported
-// too. They all surface as an input configuration error.
-const GPT_5_REASONING_EFFORTS = ["minimal", "low", "medium", "high"] as const;
+// gpt-5 accepts minimal/low/medium/high. It has no "none"; we accept it and map
+// it to the nearest supported effort ("minimal"). "xhigh" and the universal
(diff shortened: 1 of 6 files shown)


matteotrab (approved):

davidebbo (approved):
LGTM

PR #27952, opened 2026-06-25
fix: disable reasoning for process data sources

matteotrab:
## Description

Fixes a regression introduced by #27783, which changed the default `reasoningEffort` from `"none"` to `modelConfig.defaultReasoningEffort` when omitted.

`process_data_sources.ts` calls `runMultiActionsAgent` with `forceToolCall`. Claude models whose `defaultReasoningEffort` is non-null have thinking enabled by default, which Anthropic rejects when `tool_choice` forces a specific tool:

```
Invalid request to anthropic. 400 {"type":"error","error":{"type":"invalid_request_error","message":"Thinking may not be enabled when tool_choice forces tool use."}}
```

See logs [here](https://app.datadoghq.eu/logs?query=%40llmEventType%3Aerror%20%40modelId%3Aclaude-sonnet-4-6%20%40errorContent.originalError.error.error.message%3A%22Thinking%20may%20not%20be%20enabled%20when%20tool_choice%20forces%20tool%20use.%22&agg_m=count&agg_m_source=base&agg_t=count&clustering_pattern_field_path=message&cols=%40errorContent.type%2C%40context.conversationId%2C%40context.workspaceId&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&from_ts=1781081033340&to_ts=1782377033340&live=true)

This PR explicitly passes `reasoningEffort: null` in the `process_data_sources` call, restoring the previous behavior of no reasoning for this structured extraction path.

## Risk

Low. This restores the behavior that existed before #27783 for this specific call site. No other callers are affected.

## Deploy Plan

Standard deployment.

PR #27783, opened 2026-06-23
fix(llm): default reasoning effort to the model default instead of "none"

adrsimon:
## Description

The base LLM constructor hardcoded `reasoningEffort = "none"` when callers omitted it, conflating "unspecified" with "explicitly none". Some models — e.g. `gpt-5-mini` — do not support `"none"` (their config has `supportedReasoningEfforts.none = false` and the provider API rejects it), so the request failed with a `ZodError` against the model's config schema:

```
ZodError: reasoning.effort received "none", expected 'minimal' | 'low' | 'medium' | 'high'
```

Every utility caller that goes through `runMultiActionsAgent` / `getStreamLLM` (title, name, description, tag and emoji suggestions, candidate deduplication, web summarization, etc.) and the batch reinforcement caller via `getBatchLLM` omits `reasoningEffort`, so all of them crashed whenever the selected model didn't accept `"none"` (notably `getSmallWhitelistedModel`, which returns `gpt-5-mini` first).

Fix: fall back to the model's own `defaultReasoningEffort` when the caller doesn't specify one, instead of a global `"none"`. Callers that explicitly pass an effort (including `"none"` on models that support it) are unaffected.

## Tests

Type-checked (`tsgo --noEmit`) and validated through the pre-commit typecheck/biome hooks. Manually traced the failing path (`merge_into_project` → `deduplicateCandidates` → `runMultiActionsAgent` → `getStreamLLM`) to confirm `gpt-5-mini` now resolves to its `"medium"` default and passes the config schema.

## Risk

Low. The change only affects callers that omit `reasoningEffort` (previously `"none"`). Those callers now run at each model's intended `defaultReasoningEffort` — a small cost/latency increase for utility calls on reasoning models, in exchange for not crashing. The main agent runner passes an explicit effort and is unchanged. Safe to roll back by reverting the single commit.

## Deploy Plan

Standard `front` deploy. No migration, no config change.

PR #27972, opened 2026-06-25
revert: roll back reasoning effort default changes

adrsimon:
## Description

Reverts the following three commits related to reasoning effort defaults:

- #27952 — fix: disable reasoning for process data sources
- #27880 — fix: preserve null reasoningEffort overwrites
- #27783 — fix(llm): default reasoning effort to the model default instead of "none"

The real fix lands in #27971, which supersedes this line of changes.

## Tests

Reverts applied cleanly with no conflicts.

## Risk

Low — straight revert of recent commits. Safe to roll back if needed.

## Deploy Plan

Standard deploy.


matteotrab (approved):
Thank you!

PR #27978, opened 2026-06-25
Upg: do not wait between events fetch if we are paginating history.

Fraggle:
## Description

`streamConversationEventsForRoute` and `streamMessageEventsForRoute` now opt in to `writeDoneSentinel: true`, enabling the server to send `data: done` after each Redis history pagination batch (and on idle timeout). Previously these routes did not send the sentinel, so when `RedisHybridManager` triggered a reconnect via `callback("close")` during history replay, clients saw it as a connection error and applied backoff before retrying. Now `useEventSource` catches `done` and reconnects immediately, draining paginated history as fast as possible.

## Tests

Tested locally by observing SSE reconnect behaviour during history pagination — no backoff delay between pages.

## Risk

Low. `writeDoneSentinel` was already used by MCP routes with no issues. The `done` sentinel is already filtered by the SDK before JSON-parsing; `useEventSource` was already handling it. No consumer treats it as a data event.

## Deploy Plan

Deploy front.

### DIFF
--- a/front-api/lib/api/sse/conversation_events.ts
+++ b/front-api/lib/api/sse/conversation_events.ts
@@ -52,5 +52,6 @@ export async function streamConversationEventsForRoute(
         signal,
       }),
     transform: (event) => opts.transformEvent(auth, event),
+    writeDoneSentinel: true,
   });
 }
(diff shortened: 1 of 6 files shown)


Fraggle:

- #27978 ◀

flvndvd (approved):

PR #27985, opened 2026-06-25
feat(front): sound notification when an agent requires manual action

Inesvvv:
## Description

Plays a sound notification when an agent pauses and requires a manual action (tool approval / decline), so the user is alerted even when they're not looking at the conversation.

Follow-up to #27793, which added the user preference (enable + sound choice) and the `sound_notification` feature flag. This PR adds :

- New Novu `manual-action-required` workflow (silent in-app signal, intercepted client-side by tag) + `notifyManualActionRequired` helper, registered on the  front-api Novu bridge.
- Fires at the three points where a conversation is marked action-required: sandbox child action, `run_tool`, and the run-model wrapper.
- Client: `useSoundNotification` hook + `useSetupNotifications` plays the user's configured sound when the notification arrives and they aren't already viewing that conversation.

Gated behind the `sound_notification` feature flag. 

## Tests

Manual testing locally using Novu bridge to test notifications.

## Risk
Low. Entirely gated behind the `sound_notification` flag (off by default).

### DIFF
--- a/front/hooks/useSetupNotifications.ts
+++ b/front/hooks/useSetupNotifications.ts
@@ -1,21 +1,29 @@
 import { useBrowserNotification } from "@app/hooks/useBrowserNotification";
 import { useSendNotification } from "@app/hooks/useNotification";
 import { useNovuClient } from "@app/hooks/useNovuClient";
+import { useSoundNotification } from "@app/hooks/useSoundNotification";
 import config from "@app/lib/api/config";
 import { ConversationsUpdatedEvent } from "@app/lib/notifications/events";
 import { useAppRouter } from "@app/lib/platform";
 import { workspaceAuthContextUrl } from "@app/lib/swr/workspaces";
-import { PROVIDER_CREDENTIALS_HEALTH_UPDATED_TAG } from "@app/types/notification_preferences";
+import {
+  MANUAL_ACTION_REQUIRED_TAG,
+  PROVIDER_CREDENTIALS_HEALTH_UPDATED_TAG,
+} from "@app/types/notification_preferences";
 import { isString } from "@app/types/shared/utils/general";
 import type { Novu } from "@novu/js";
 import { useEffect } from "react";
 import { mutate } from "swr";
+
+const isActivelyViewing = (isOnConversationPage: boolean): boolean =>
+  isOnConversationPage && window.document.hasFocus();
 
 export const useSetupNotifications = () => {
   const { push } = useAppRouter();
   const { novuClient } = useNovuClient();
   const sendNotification = useSendNotification();
   const { allowBrowserNotification, notify } = useBrowserNotification();
+  const { requestManualActionSound } = useSoundNotification();
 
   useEffect(() => {
     const setupNotifications = async (novuClient: Novu) => {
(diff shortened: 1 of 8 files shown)


PopDaph (commented):
**Notification granularity (nit)**

One thing I wanted to confirm on the firing points: as far as I can tell, the wrapper and `run_tool` paths are mostly mutually exclusive, not double-firing for the same approval.

The wrapper fires once per step for creation-time approvals (`someActionsRequireManualApproval` is a single check, so N approval tools in a step -> one event), and the loop breaks on `needsApproval` before `runToolActivity` runs. So `run_tool` should only fire for execution-time blocks that the wrapper doesn’t cover.

I don’t think there’s a same-action double-fire to worry about there. The residual case is multiple parallel tools each hitting an execution-time block in the same step, which could emit several events and therefore several chimes. Minor, but if we want at-most-one-
chime-per-turn, the simplest fix may be a client-side debounce keyed on `conversationId` in `useSoundNotification.ts`, rather than trying to force every server path through the same guard.

Not blocking, but I think the intended granularity should be explicit.

**Server-side opt-out gate**

The Novu workflow exists only to make the client play a local sound, but today the call sites trigger it unconditionally. The only server gate is `auth.user()` existing, and `sound_notification_enabled` is checked client-side.

Could we add a server-side early return when `sound_notification_enabled !== "true"` before triggering Novu, so we don’t emit Novu traffic for users who never enabled it? Absence of the metadata should be treated as disabled, matching the current client default.

`auth.user()` is available and `UserResource.getMetadata(...)` exists, so this should be a small change.

Two notes:
- The “don’t play if I’m already focused on this conversation” suppression still has to stay client-side, since the server can’t know
that.
- This gate naturally no-ops for agent/API-triggered runs where `auth.user()` is null, which seems fine.

Where feasible, we should avoid doing the metadata lookup repeatedly for multiple events in the same blocked turn.

front/hooks/useSetupNotifications.ts
```diff
@@ -40,6 +45,26 @@ export const useSetupNotifications = () => {
             return;
           }
 
+          if (
+            notification.result.tags?.includes(MANUAL_ACTION_REQUIRED_TAG) &&
```
Fraggle, line 52: nit: some of the code look exactly like the block below.

Makes me wonder if you want to share and show a notification to redirect ?

front/hooks/useSoundNotification.ts
```diff
@@ -0,0 +1,158 @@
...
+import {
+  DEFAULT_SOUND_NOTIFICATION,
+  isSoundNotificationType,
+  SOUND_NOTIFICATION_METADATA_KEYS,
+} from "@app/types/notification_preferences";
+import { useCallback, useEffect, useState } from "react";
+
+let audioContext: AudioContext | null = null;
+const audioBufferCache = new Map<string, AudioBuffer>();
+let unlockListenerAttached = false;
```
Fraggle: Having all theses things stored in global memory isn't great :(
I would scope them to the hook.

front/temporal/agent_loop/activities/run_model_and_create_actions_wrapper.ts
```diff
@@ -292,6 +293,12 @@ async function _runModelAndCreateActionsActivity({
     await ConversationResource.markAsActionRequired(auth, {
       conversation: runAgentData.conversation,
     });
+
+    if (!runAgentData.conversation.actionRequired) {
+      notifyManualActionRequired(auth, {
```
Fraggle, line 298: it's werid to call "notifyManualActionRequired" when the flag is false.

front/temporal/agent_loop/activities/run_tool.ts
```diff
@@ -396,6 +397,13 @@ async function executeToolStreaming(
           conversation,
         });
 
+        if (!conversation.actionRequired) {
+          notifyManualActionRequired(auth, {
```
Fraggle, line 401: same here, it's werid to call "notifyManualActionRequired" when the flag is false.

Fraggle (approved):
It works but did you try to encapsulate everything in the hook instead of a mix of global memory and hook ?

PR #27987, opened 2026-06-25
[front] enh: Display code-defined skills instructions

id13:
## Description

In a skills in skills world, we might want people to start building on top of the code-defined, dust-curated skills (at least in my case for pptx).

Right now, you can't look at their instructions (except if you look at our code)

This is a 80/20 fix for that.

Not strongly held on the solution, somewhat held on the necessity to do it though.

## Tests

Tested locally

## Risk

Low

## Deploy Plan

Deploy front


aubin-tchoi (approved):
LGTM

aubin-tchoi:
Maybe we should decide on a per-skill basis whether instructions should be shown (e.g. for Frames it's better to make it opaque)
At least should probably remain hidden for system skills

id13:
True, then maybe I make it opted-out by default and only enable it for the ones I own

aubin-tchoi (approved):

PR #27988, opened 2026-06-25
Poke plugin to assign seats

tdraier:
## Description

Adds a new workspace-scoped Poke plugin, **Batch Update Seats**, that lets an operator assign a seat type to many users at once.

- Operator picks a **seat type** (dropdown over `MEMBERSHIP_SEAT_TYPES`) and pastes a list of **emails**, one per line.
- For each email we look up the user and their active membership in the workspace, then call `updateMembershipSeatAndTrack` (concurrency 10) — the same path used everywhere else, so all existing rules (free-seat eligibility, paid-seat-on-free-plan, internal `maxSeats` cap, Metronome scheduling, audit logging, debounced seat-count sync) apply unchanged.
- For paid seat types, an upfront **contract entitlement** check (`getSeatSubscriptionsFromContract`) rejects the whole batch if the requested seat isn't sold on the workspace's contract. This is checked once (the contract is workspace-level). It only enforces when the contract actually sells seats; contracts with no seat subscription (or non-Metronome workspaces) fall through to existing behavior.
- Returns simple JSON with a per-email result list. Status is one of:
  - `updated` — `previousSeatType`, `newSeatType`, `scheduledSeatChangeAt` (ISO string or `null` if applied immediately)
  - `unchanged` — already the requested `seatType`, no write performed
  - `user_not_found` — no user matched that email
  - `failed` — carries the `error` type (`not_found`, `metronome_error`, `free_seat_not_allowed`, `paid_seat_not_allowed_on_free_plan`, `seat_limit_reached`)

Gated to `requiredRoles: ["support"]`. The operator is recorded as the seat-change author.

## Tests

Type-check (`tsgo`) and biome pass clean. No automated tests added — this is an operator-only Poke plugin that composes existing, already-tested helpers (`updateMembershipSeatAndTrack`, `getSeatSubscriptionsFromContract`); behavior to verify manually via Poke against a test workspace (mixed list of members / non-members / already-correct seats / a seat type not on the contract).

## Risk

Low. Operator-only tool behind the `support` role. It does not introduce new seat-mutation logic — it delegates to `updateMembershipSeatAndTrack`, so billing/Metronome behavior is identical to existing single-user seat changes. The entitlement guard only adds a rejection path; it never widens what can be assigned. Fully revertible (additive plugin file + one index export line).

## Deploy Plan

Standard `front` deploy. No migrations, env vars, or feature flags. Plugin becomes available in Poke immediately on deploy.

### DIFF
--- a/front/lib/api/poke/plugins/workspaces/batch_update_seats.ts
+++ b/front/lib/api/poke/plugins/workspaces/batch_update_seats.ts
@@ -1,0 +1,142 @@
+import { updateMembershipSeatAndTrack } from "@app/lib/api/membership";
+import { createPlugin } from "@app/lib/api/poke/types";
+import { getActiveContract } from "@app/lib/metronome/plan_type";
+import {
+  getProductSeatTypes,
+  getSeatSubscriptionsFromContract,
+} from "@app/lib/metronome/seat_types";
+import { UserResource } from "@app/lib/resources/user_resource";
+import { concurrentExecutor } from "@app/lib/utils/async_utils";
+import {
+  isMembershipSeatType,
+  isPaidSeatType,
+  MEMBERSHIP_SEAT_TYPES,
+} from "@app/types/memberships";
+import { mapToEnumValues } from "@app/types/poke/plugins";
+import { Err, Ok } from "@app/types/shared/result";
+
+export const batchUpdateSeatsPlugin = createPlugin({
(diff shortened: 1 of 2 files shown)


tdraier:

- #27988 ◀

rfrenoy:
LGTM

PR #27989, opened 2026-06-25
Fix blank-line/blank-list-item artifacts in suggestion diff preview

smb2268:
fix https://github.com/dust-tt/tasks/issues/9219
## Description
The suggestion diff preview rendered spurious blank lines and blank list items: swapping a straight apostrophe for a curly one added a phantom line, and editing a list item rendered an empty row before the change. The accepted content was always correct — this was purely a decoration rendering artifact.

Root cause: diffBlockContent() returns change ranges in the block's full content space, which includes the open/close tokens of nested blocks (paragraphs, list items). buildBlockDecorations then cut newNode at those positions and serialized the result, so an inline-level change inside a block-with-block-children got re-wrapped in its containing <p>/<li> and rendered inline as a phantom block.

Fix: take a real Slice and unwrap single-chain block wrappers when the change is contained within one text block (open on both ends), and drop a leading empty block left by an empty-paragraph insertion. Genuine cross-block / new-block additions keep their structure. acceptSuggestion semantics are unchanged.

## Tests
Tested manually
Before:
Screenshot 2026-06-25 at 3 19 15 PM

After:
Screenshot 2026-06-25 at 3 07 28 PM

## Risk
Low

## Deploy Plan

- [ ] deploy front

### DIFF
--- a/front/components/editor/extensions/agent_builder/InstructionSuggestionExtension.test.ts
+++ b/front/components/editor/extensions/agent_builder/InstructionSuggestionExtension.test.ts
@@ -569,6 +569,244 @@ describe("InstructionSuggestionExtension", () => {
     });
   });
 
+  describe("diff rendering artifacts (blank line / blank list item)", () => {
+    // Addition widgets should render inline content; a block-level element
+    // (<p>, <li>) inside one means a wrapper leaked and renders as a phantom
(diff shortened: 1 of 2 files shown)


flvndvd (commented):
LGTM

PR #27990, opened 2026-06-25
Fix: make compact bar UI more visible now with the optmimized scrolling, fix the live transcribe on mobile.

Fraggle:
## Description

Two independent fixes, both stemming from the document-scroll mobile layout introduced in PR27960:

- **Compact bar visibility**: `inputBarCompactStyles.ts` bumps the pill to `h-10`, `copy-md font-medium` preview text, stronger `backdrop-blur-xl`/`backdrop-saturate-150`, and `bg-background/75` — making the floating compact bar clearly legible against the conversation background.
- **Live transcribe on mobile**: `useEditorService` now receives `isMobileViewport` and uses `insertContentAt(..., { updateSelection: false })` instead of `chain().focus("end")` when inserting `voicePartial`/committed text on mobile, preventing the editor from stealing focus mid-transcription.
- **`sm:` → `md:` breakpoint alignment**: All input bar and layout responsive classes updated to `md:` (768px) to match the `useIsMobile` threshold and prevent breakpoints from firing on mid-size mobile viewports.
- `enforceStickyFooterAtBottom` unconditionally enabled now that mobile scroll is handled at the document level.

## Tests

Tested locally on mobile viewport. Verified compact bar appearance and live transcription flow.

## Risk

Low. Pure front-end styling and focus-management changes. No data model or API changes. Safe to rollback.

## Deploy Plan

Deploy `front`.

### DIFF
--- a/front/components/assistant/conversation/input_bar/InputBar.tsx
+++ b/front/components/assistant/conversation/input_bar/InputBar.tsx
@@ -440,7 +440,7 @@ export const InputBar = React.memo(function InputBar({
         }}
         className={classNames(
           isShaking && "animate-shake",
-          "relative flex flex-col items-stretch gap-0 sm:flex-row",
+          "relative flex flex-col items-stretch gap-0 md:flex-row",
           INPUT_BAR_COMPACT_MORPH_TRANSITION_CLASSES,
           !effectiveIsCompact && "w-full flex-1 self-stretch",
           effectiveIsCompact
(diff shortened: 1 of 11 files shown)


Fraggle:

- #27990 ◀

front/components/assistant/conversation/input_bar/InputBar.tsx
```diff
@@ -440,7 +440,7 @@ export const InputBar = React.memo(function InputBar({
         }}
         className={classNames(
           isShaking && "animate-shake",
-          "relative flex flex-col items-stretch gap-0 sm:flex-row",
+          "relative flex flex-col items-stretch gap-0 md:flex-row",
```
Fraggle, line 443: aligning to use md: instead of sm: for determining between mobile / desktop (as useIsMobile does)

ykmsd: not really related but since we are more investing in mobile UI, I wonder if it's better to have something semantic like `mobile:` instead of `md`?

Fraggle: actually I _think_ it's the opposite md: will trigger the rule if we pass the width threshold. So it's more like desktop:....

ykmsd: ah yes 🙃

PR #28001, opened 2026-06-26
[front] metronome: paginate listMetronomeCustomerCommits/Credits

rfrenoy:
## Description

`listMetronomeCustomerCommits` and `listMetronomeCustomerCredits` returned only `page.data` from a single `v1.customers.*.list(...)` call, silently dropping every entry past the first page. Iterate the SDK's async-paginated result instead so callers see the full set.


tdraier (approved):
LGTM

PR #28002, opened 2026-06-26
Surface document size and nudge toward grep in the company data cat tool

flvndvd:
## Description

This PR aims at fixing (to some extend) https://github.com/dust-tt/tasks/issues/9245.

When an agent reads a document through the company data cat tool, it previously received only the raw text with no indication of how large the document was or whether more content remained. The result was a common failure mode where the agent would keep paginating by incrementing the offset to pull the rest of a large document, accumulating every page in its context and exhausting the window for no good reason.

This change appends a short footer to the returned text so the agent always knows the total size of the document and, when content remains, exactly where to continue from. Once the agent has paged to the end, it now gets an explicit end of document marker so a looping read stops deterministically instead of nudging the offset indefinitely. For documents above a large threshold the footer also steers the agent toward filtering with a regular expression rather than reading everything linearly. The total size and offset values were already returned by core but had been discarded, so no new backend work is required, and the window is computed from the offset and limit rather than the returned text length so it stays correct even when a regular expression filter is applied after slicing.

The tool and parameter descriptions were updated to match, telling the model that the output reports total size and continuation offset and that filtering is preferred over paging a large document to exhaustion. The behavior change is purely additive: a bracketed footer in the text plus richer descriptions, with no change to the API contract.

### DIFF
--- a/front/lib/actions/mcp_internal_actions/types.ts
+++ b/front/lib/actions/mcp_internal_actions/types.ts
@@ -211,20 +211,24 @@ export const DataSourceFilesystemCatInputSchema = z.object({
     .optional()
     .describe(
       "The character position to start reading from (0-based). If not provided, starts from " +
-        "the beginning."
+        "the beginning. The tool output reports the total document size and, when more content " +
+        "remains, the offset to use to continue reading."
     ),
   limit: z
     .number()
     .optional()
     .describe(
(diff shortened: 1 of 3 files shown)


fontanierh (approved):
LGTM

aubin-tchoi (approved):
LGTM

PR #28003, opened 2026-06-26
[front] billing: expire contract commits/credits one year after start

rfrenoy:
## Description

Ticket: https://github.com/dust-tt/tasks/issues/9225

Contract commits and credits (initial credits, AWU top-ups, business-activation seat prepayment) were granted a far-future (2999) "forever" access window and a `DUST_CARRY_ON_RENEWAL=forever` custom field, so they never expired and carried their full balance forever across renewals.

Grant them a one-year window instead:
- add `oneYearAfter(start)` and use it at the three creation sites for both the Metronome access `ending_before` and the `DUST_CARRY_ON_RENEWAL` value, so the expiry is absolute (one year from the original start) and preserved — not reset — when a balance is carried onto a renewed contract.
- add `scripts/backfill_contract_credit_expiry.ts` to rewrite existing forever entries to expire one year after they start (dry-run by default; flags retroactively-expired entries before --execute).
## Tests

Local

## Risk

Low
## Deploy Plan

Run `cd front && npx tsx scripts/backfill_contract_credit_expiry.ts --execute`


tdraier (approved):
🙏

PR #28004, opened 2026-06-26
Derive sandbox MCP request timeout from max command timeout plus buffer

flvndvd:
## Description

The sandbox bash tool enforces a command timeout inside the container, returning captured output when the limit is reached. The MCP layer applies its own hard deadline to the whole tool call. Both were hardcoded to the same two minute value, so when the model requested the maximum command timeout the two deadlines fired at the same moment. Because the in-container path needs extra time to tear down the process, flush output, and round-trip to the host, the MCP deadline tended to win and abort the call before the graceful timeout could return its partial output, surfacing a hard error instead of a clean result.

This introduces a single source of truth for the default and maximum command timeouts and derives the MCP deadline from the maximum plus a buffer, so the in-container timeout always resolves first and the MCP deadline acts purely as an outer backstop. Raising the maximum command timeout now moves the MCP ceiling with it.


fontanierh (approved):
Thank you for fixing

PR #28005, opened 2026-06-26
Remove legacy sandbox ownership fallback

fontanierh:
Follow up of #27494 and #27916.

## Description

Remove the remaining sandbox ownership read fallback to the legacy `sandboxes.conversationId` column. Conversation sandbox reads and reaper ownership resolution now depend on `conversation_sandboxes` only, while writes still keep the legacy column populated until the later drop-column PR.

## Tests

Tested locally with explicit Biome on the changed files, `NODE_ENV=test npm test -- lib/resources/sandbox_resource.test.ts`, and `nvm use && NODE_OPTIONS=\"--max-old-space-size=8192\" npx tsc --noEmit` in `front`.

## Risk

A sandbox missing its `conversation_sandboxes` row will no longer be resolved from the legacy column. Safe to rollback by reverting this PR while the legacy column still exists.

## Deploy Plan

- [ ] Before deploying this PR, verify `conversation_sandboxes` is ISO with `sandboxes.conversationId` in prod (`missing=0`, `extra=0`) using `front/migrations/20260624_backfill_conversation_sandboxes.ts`.
- [ ] Deploy front.
- [ ] Monitor sandbox lifecycle/reaper errors.

### DIFF
--- a/front/lib/resources/sandbox_resource.test.ts
+++ b/front/lib/resources/sandbox_resource.test.ts
@@ -235,6 +235,39 @@ describe("SandboxResource.dangerouslyDestroyIfSleeping", () => {
     );
     expect(reloaded?.status).toBe("deleted");
   });
+
+  it("does not fall back to the legacy conversationId column when ownership is missing", async () => {
+    const sandbox = await SandboxFactory.create(
+      authenticator,
+      conversationResource.toJSON(),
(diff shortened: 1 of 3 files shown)


flvndvd (commented):
LGTM

PR #28006, opened 2026-06-26
Route conversation sandbox calls through conversations

fontanierh:
## Description

Route conversation-owned sandbox operations through `ConversationResource` instead of direct external calls to `SandboxResource`. This adds conversation-level methods for fetch, ensure active, pause for approval, deletion, and reaper lifecycle actions while keeping fleet-level sandbox scans on `SandboxResource`.

Also updates the sandbox lifecycle helper, sandbox-child blocking, conversation deletion, reaper activities, and test helpers to use the conversation boundary.

## Tests

- `npx biome check --write --error-on-warnings front/lib/api/assistant/conversation/destroy.ts front/lib/api/sandbox/lifecycle.test.ts front/lib/api/sandbox/lifecycle.ts front/lib/api/sandbox/sandbox_child_block.ts front/lib/resources/conversation_resource.ts front/lib/resources/sandbox_resource.ts front/temporal/sandbox_reaper/activities.ts front/temporal/sandbox_reaper/kill_requester/activities.test.ts front/tests/utils/SandboxFactory.ts`
- `cd front && NODE_ENV=test npm test -- lib/api/sandbox/lifecycle.test.ts temporal/sandbox_reaper/kill_requester/activities.test.ts lib/resources/sandbox_resource.test.ts lib/api/sandbox/sandbox_child_block.test.ts lib/resources/conversation_resource.test.ts`
- `cd front && NODE_OPTIONS="--max-old-space-size=8192" npx tsc --noEmit`
- GitHub CI is green on the latest commit.

## Risk

Low. This is a boundary refactor: behavior remains delegated to the existing sandbox resource implementation. The main risk is missing a call site, mitigated by search for direct conversation-owned `SandboxResource` calls plus focused lifecycle, reaper, and conversation resource tests.

## Deploy Plan

Standard deploy. No schema migration, data migration, feature flag, or backfill is required.

### DIFF
--- a/front/lib/resources/sandbox_resource.ts
+++ b/front/lib/resources/sandbox_resource.ts
@@ -42,11 +42,16 @@ import assert from "assert";
 import type { Attributes, ModelStatic, Transaction } from "sequelize";
 import { Op } from "sequelize";
 
-interface EnsureSandboxResult {
+export interface EnsureSandboxResult {
   freshlyCreated: boolean;
   sandbox: SandboxResource;
   wokeFromSleep: boolean;
 }
+
+export type ConversationSandboxOwner = Pick<
+  ConversationWithoutContentType,
+  "id" | "sId"
+>;
 
 // eslint-disable-next-line @typescript-eslint/no-unsafe-declaration-merging
 export interface SandboxResource extends ReadonlyAttributesType<SandboxModel> {}
@@ -229,7 +234,7 @@ export class SandboxResource extends BaseResource<SandboxModel> {
 
   static async fetchByConversation(
     auth: Authenticator,
-    conversation: ConversationWithoutContentType
+    conversation: ConversationSandboxOwner
   ): Promise<SandboxResource | null> {
     const link = await this.conversationSandboxModel.findOne({
       where: {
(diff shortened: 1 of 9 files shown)


front/lib/resources/sandbox_resource.ts
```diff
@@ -440,7 +445,7 @@ export class SandboxResource extends BaseResource<SandboxModel> {
   // cannot shadow a system var like CONVERSATION_ID.
   private static async buildSandboxEnvVars(
     auth: Authenticator,
-    conversation: ConversationWithoutContentType,
+    conversation: ConversationSandboxOwner,
```
flvndvd, line 448: `nit` it would even better if we never leak conversation here, since we seem to only use it for some env vars. It could be a `Record<string, string>`, WDYT?

flvndvd (approved):
Sad that we have some code paths still relying on type over the resource that constraint to add the static variation of the method (not on you).

front/lib/resources/sandbox_resource.ts
fontanierh: Yep 100%

Doing next PR but in any case we have to because next it can be for a pod

PR #27984, opened 2026-06-25
Remove enterprise shadow contracts

tdraier:
## Description

Removes enterprise shadow contract provisioning and all associated dead code. Enterprise workspaces were previously provisioned with a Metronome shadow contract (running in parallel with Stripe for invoice comparison). This is no longer needed.

**Removed from the enterprise upgrade flow:**
- `provisionShadowEnterpriseMetronomeContract` — the function that created shadow Metronome contracts for enterprise on upgrade
- The `stripeSubscription` parameter from `pokeUpgradeWorkspaceToEnterprise` (was only used for shadow provisioning)
- The shadow branch in the Poke upgrade endpoint

**Deleted dead code that was only needed for shadow enterprise contracts:**
- `extractEnterprisePricing`, `buildEnterpriseOverrides`, `applyEnterpriseOverrides`, `countMauForWorkspace` and all related types/helpers from `contracts.ts` (~600 lines)
- `mau_sync.ts` entirely (MAU tier parsing, MAU sync logic)
- `mau_sync.test.ts`, `mau_contract.test.ts`
- `scripts/migrate_metronome_contracts.ts`
- All MAU product ID constants (`getProductMauId`, `getProductMauTierIds`, 14 backing DEV/PROD constants) from `constants.ts`
- MAU fields (`mau`, `mauUnitPriceCents`, `mauTierUnitPricesCents`) from the invoice endpoint and `MetronomeInvoiceSummary` type
- `mauTiers` from `MetronomeContractSummary` type and the `getMetronomeContractSummary` function
- MAU tier and simple MAU rendering blocks from `MetronomeSubscriptionPanel`

## Tests

Updated `contracts.test.ts`: removed stale mocks for deleted modules (`mau_sync`, MAU constants), removed tests for deleted functions (`extractEnterprisePricing`, `buildEnterpriseOverrides`, the old `syncContractQuantities`-based seat sync tests), and updated the `provisionMetronomeContract` test to match the simplified seat-only sync path.

## Risk

Low. This removes provisioning logic that is no longer exercised — existing enterprise workspaces are either Stripe-only or fully Metronome-billed. No runtime behavior changes for workspaces already in production. Rollback is safe: the removed code paths are not reachable from any currently deployed client.

## Deploy Plan

Deploy as normal — no migration needed.

PR #27965, opened 2026-06-25
Remove MAU sync

tdraier:
## Description

Shadow contracts for legacy enterprise plans are not used anymore (neither for testing or for revrec), we can remove the MAU sync which was completely specific to these contract types. Simplify the syncSeats* methods and remove duplicate code.

## Risk

low

## Deploy Plan

deploy front
delete schedule manually (`temporal schedule delete --schedule-id metronome-gauge-schedule` or ui)

PR #28008, opened 2026-06-26
Cleaned subscription page

tdraier:
## Description

Removes `MetronomeSubscriptionPanel` — a dead code path on the Subscription page that could never render (any workspace with `isMetronomeOnlyBilled = true` lands on `/billing`, not `/subscription`).

- Deletes `MetronomeSubscriptionPanel.tsx`
- Removes the `useMetronomePanel` branch from `SubscriptionPage`
- Moves `CancelMetronomeSubscriptionDialog` and `ReactivateMetronomeSubscriptionDialog` (still live on the Billing page) into `SubscriptionActionButtons.tsx` where they're used

## Tests

No behavior change — the panel was unreachable. Type-checks clean.

## Risk

None — dead code removal only.

## Deploy Plan

Deploy as normal.


tdraier:

- #27965
    - #27984
        - #28008 ◀

thervier (approved):
LGTM

small reflection: maybe we could return directly when isSubscriptionMetronomeBilled to avoid displaying the page at all if a user finds it in a way via url?

tdraier:
> LGTM
> 
> small reflection: maybe we could return directly when isSubscriptionMetronomeBilled to avoid displaying the page at all if a user finds it in a way via url?

actually we can directly redirect to /billing in that case

thervier:
@tdraier yes even better, you are right !

PR #28011, opened 2026-06-26
feat: enable alerting for sandbox tool

matteotrab:
## Description

Enables `enableAlerting` on the sandbox `bash` tool so that infrastructure-level errors are picked up by the tool datadog monitors.

Notice that bash command failures are returned as successful results with the output, not as tracked errors. Tracked errors should only fire on infrastructure failures (sandbox not starting, lifecycle issues, etc.) which are the kind of errors that should be surfaced on the monitors.

Yesterday's incident revealed that sandbox errors were going undetected because no monitor was wired up.

## Risk

Low. Only adds signal, doesn't change behavior.

## Deploy Plan

Standard deployment.


matteotrab:

- #28011 ◀

flvndvd (approved):
Let's see how it goes, hopefully not too noisy!

PR #28012, opened 2026-06-26
Ultra temporary debug log.

Fraggle:
## Description

Adds a temporary `logger.info` call in the `POST /api/v1/w/:wId/assistant/conversations` handler, gated on workspace sId `ba367d3014`. Logs the full validated request body to help debug an issue on that specific workspace. Must be reverted once the investigation is done.

## Tests

Targeted debug log — no automated tests. Verified manually that the guard sId prevents any impact on other workspaces.

## Risk

Low. Entirely gated behind a hardcoded workspace sId check; no behavior change for any other workspace. Log output may include request body content, so should not stay in prod longer than needed.

## Deploy Plan

Deploy front. Revert and redeploy once debugging is complete.


Fraggle:

- #28012 ◀

davidebbo (commented):
LGTM, but track removing it :)

PR #28015, opened 2026-06-26
[front] enh: show enabled skill instructions in action details

aubin-tchoi:
## Description

This PR adds a skill instructions preview to the MCP `enable_skill` action details. The sidebar now resolves the enabled skill from the tool output, keeps the original message first (smth like "The tool was successfully enabled"), and renders the skill guidelines below it with the same readonly instructions editor used by the Skill info page.

Conversation inline details stay unchanged, and the instructions section is hidden when the skill has no instructions.

image

## Tests

- Tested locally.

## Risk

- Low.

## Deploy Plan

- Deploy front.


pinotalexandre (approved):
LGTM

PR #28017, opened 2026-06-26
[front] feat: add skill favorite data model

aubin-tchoi:
## Description

- Adds the data model for per-user skill favorites, which comes down to a new `skill_favorites` table.
- We add a table that has at most 1 row per user with a list of their favorite skills. We accept a weaker consistency on references compared to a fully normalized data model; stale star IDs are harmless.
 we need to be able to efficiently fetch the skills the current user set as favorite" (in the agent loop, in the UI) but also the users who set a certain skill as favorite (to show the number of stars on a skill). There is no specific join, all queries will be direct queries by skill ID or by user ID.

### Overall goal

The goal of this feature is to allow setting skills as favorite.
The main entry point is `SkillInfoPage` via Manage Skills, from where a user can set/unset a skill as favorite (star button).
Favorite skills appear first in skill lists (capabilities picker, slash command results) but do not affect sorting (invariant: two users typing the same search query should get the same result).
Agents that can discover skills (i.e. have the skill "Discover Skills") are aware of favorite skills, they see their descriptions and can enable them.

This will allow cleaning up the default skills on our workspace, which have been used for the exact use case of personal favorite skills, but affecting everyone, amounting to 10k tokens currently :).

## Tests

- Tested locally.

## Risk

- Low. Additive schema/model change.

## Deploy Plan

- Run migration.
- Deploy front.

### DIFF
--- a/front/types/shared/feature_flags.ts
+++ b/front/types/shared/feature_flags.ts
@@ -246,6 +246,11 @@ export const WHITELISTABLE_FEATURES_CONFIG = {
       "Enable the Plan Mode skill: agents maintain a live plan.md for non-trivial tasks, with an optional human-approval checkpoint.",
     stage: "dust_only",
   },
+  skill_favorites: {
+    description:
+      "Enable user favorites for skills, including favorite controls and runtime skill availability.",
+    stage: "dust_only",
+  },
   allow_old_notion_mcp: {
     description:
       "Allow individual workspaces to keep using the old internal Notion MCP server alongside the official one",
--- a/front/types/shared/feature_flags.ts
+++ b/front/types/shared/feature_flags.ts
@@ -246,11 +246,6 @@ export const WHITELISTABLE_FEATURES_CONFIG = {
       "Enable the Plan Mode skill: agents maintain a live plan.md for non-trivial tasks, with an optional human-approval checkpoint.",
     stage: "dust_only",
   },
-  skill_favorites: {
-    description:
-      "Enable user favorites for skills, including favorite controls and runtime skill availability.",
-    stage: "dust_only",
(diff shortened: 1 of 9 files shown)


id13:
Highly against normalizing this relationship, would rather store favorited skill ids in a column on user and a counter on skill table :)

fontanierh (approved):

aubin-tchoi:
> Highly against normalizing this relationship, would rather store favorited skill ids in a column on user and a counter on skill table :)

the two unique indexes are a good sign this is not the right pattern, so I'm in favor of denormalizing but would still have a separate table. 99% of the times where we fetch users (e.g. `fromSession`) we don't need to pull this list of favorites

aubin-tchoi:
so this would be 1 row per user with a list of skill IDs in this case
we do lose the FK constraint, so cleaning up this list of favorites can get really tricky

aubin-tchoi:
and as for the favorite count, recovering from divergence between the two tables seems more or less impossible

aubin-tchoi:
settling on one new table with 1 row per user and an array of skills 🤝

front/types/shared/feature_flags.ts
```diff
@@ -246,6 +246,11 @@ export const WHITELISTABLE_FEATURES_CONFIG = {
       "Enable the Plan Mode skill: agents maintain a live plan.md for non-trivial tasks, with an optional human-approval checkpoint.",
     stage: "dust_only",
   },
+  skill_favorites: {
```
aubin-tchoi: moving this to a separate PR

id13 (approved):
LGTM

PR #28024, opened 2026-06-26
Introduce SandboxFunction model

spolu:
## Description

Introduce a new SandboxFunctionModel to back the upcoming PodEndpointResource.

Naming discussion [here](https://dust4ai.slack.com/archives/C0AGY9MTNJV/p1782484696707239)

Points to:
- A pod (a sandbox function lives in a pod for now, maybe not in the future)
- A FileResource (the deployed sandbox function version)
- A JSONSchema for input and outputs (to present sandbox functions as tools to agents)

JSONSchema storage is aligned with https://github.com/dust-tt/dust/blob/main/front/lib/models/agent/actions/mcp.ts

## Tests

N/A

## Risk

Low

## Deploy Plan

- Run `npm run migration:apply:pre-deploy` on `prodbox`
- Deploy `front`


front/lib/resources/storage/models/sandbox_function.ts
```diff
@@ -0,0 +1,110 @@
...
+    modelName: "pod_endpoint",
+    sequelize: frontSequelize,
+    indexes: [
+      {
+        fields: ["workspaceId", "podId", "fileId"],
+        unique: true,
+        concurrently: true,
+      },
+      {
+        fields: ["podId"],
```
spolu, line 79: I presume this one can be removed?

front/lib/resources/storage/models/sandbox_function.ts
```diff
@@ -0,0 +1,110 @@
...
+        concurrently: true,
+      },
+      {
+        fields: ["fileId"],
```
spolu, line 83: This one is good to keep I suspect but should I add workspaceId?

front/lib/resources/storage/models/pod_endpoint.ts
```diff
@@ -0,0 +1,110 @@
...
+    throw new Error("JSON schema is not an object or a string");
+  }
+
+  const validationResult = validateJsonSchema(value);
+  if (!validationResult.isValid) {
+    throw new Error(`Invalid JSON schema: ${validationResult.error}`);
+  }
+}
+
+export class PodEndpointModel extends WorkspaceAwareModel<PodEndpointModel> {
```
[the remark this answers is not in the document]
spolu: T'es con toi (resource is coming)

front/lib/resources/storage/models/sandbox_function.ts
flvndvd: It's not an art, so it really depends about the volume/stats of the table, it might go down the path of `["workspaceId", "podId", "fileId"]` if they are many for a given workspace, or just use podId if it assumes that it's a small enough resultset. IMHO it's not too heavy so we can keep it around.

front/lib/resources/storage/models/sandbox_function.ts
flvndvd: We could make this one unique, since we might not want to have two functions relying on the same file, right?

flvndvd (approved):

front/lib/resources/storage/models/sandbox_function.ts
spolu: done

front/lib/resources/storage/models/sandbox_function.ts
spolu: as per IRL keeping. low cost.

front/lib/resources/storage/models/sandbox_function.ts
id13: WorkspaceId is counter productive here
Good to keep single index

front/lib/resources/storage/models/sandbox_function.ts
id13: @spolu per the run book about data model: this index is absolutely needed so good thing to keep , and the composite one is most probably useless, much better to start without and add if really necessary.

flvndvd: Let's keep it though cause we all love too much the `SELECT` and `DELETE` on the whole workspace 🙃

id13: Fair enough, since workspaceId does not have a single-column index on this model afaik

PR #28026, opened 2026-06-26
Remove available-tool-servers listing from the system prompt

flvndvd:
## Description

The system prompt enumerated every MCP server and listed its tool names under per-server headings. That listing is redundant with the actual tool specifications the model already receives, and the per-server instructions it also carried have moved into the individual tool descriptions. It added prompt bloat and per-request entropy that hurt prompt-cache stability for no benefit.

This drops the listing block entirely. Tool availability is unchanged since tools are passed to the model through their full specifications independently of this prompt section.

## Risk

Could break model understanding of server and connection between tools. Safe to rollback.

### DIFF
--- a/front/lib/api/assistant/generation.ts
+++ b/front/lib/api/assistant/generation.ts
@@ -164,38 +164,6 @@ function constructToolsSection({
 
   toolsSection += toolUseDirectives;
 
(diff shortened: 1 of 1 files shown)


fontanierh (approved):

aubin-tchoi (approved):
LGTM

PR #28029, opened 2026-06-26
Add an offline BM25 diagnostic for MCP tool-search recall

flvndvd:
## Description

We are about to turn on tool search, which surfaces a tool by lexically matching a user's intent against that tool's name, description, and input parameters. This change _reproduces_ that same ranking offline so we can verify, as we revise tool wording one server at a time, that a realistic phrasing actually surfaces the intended tool first, and that the read and search tools that look alike across different services stay distinguishable from one another.

It reads the live tool definitions directly rather than copying any wording, so what it scores is exactly what the real index would see, and it cannot fall out of sync with the source. It is deliberately a manual check for now: you run it on demand while iterating on descriptions, and it gates nothing.

This is intentionally temporary in that form. Once the wording passes settle, we will consolidate the same checks into a regression test that fails the build when a future edit pushes a tool out of its expected rank. At that point the manual step is no longer needed and goes away. Until then, this gives us a fast, faithful way to tune recall without guessing.


fontanierh (approved):
looks v reasonable

aubin-tchoi (approved):
LGTM

PR #28030, opened 2026-06-26
Improve Confluence tool descriptions for tool-search recall

flvndvd:
## Description

Part of https://github.com/dust-tt/tasks/issues/9168.

This sharpens the wording of the Confluence tools so the right one is surfaced when tool search ranks them by how well a user's phrasing matches each tool's name, description, and parameters. The main confusion was between searching for pages and reading one page by its identifier, which read almost identically and competed for the same intents. Searching now leads clearly with finding and locating a page by topic when you do not yet know its identifier, while reading leads with retrieving a single page once you already have its identifier. A second issue was that the current-user tool used a common word that pulled unrelated questions toward it; its wording now centres on telling you who you are, which also lets it answer that question directly. The create tool no longer borrows vocabulary that belonged to reading. No behaviour changes, only descriptions.


aubin-tchoi (approved):
LGTM

PR #28032, opened 2026-06-26
fix(analytics) stop get_agent_details leaking private agent details

rfrenoy:
## Description

The get_agent_details workspace analytics tool printed the description and full instructions of agents the admin cannot read (hidden-scope agents returned by getAgentConfigurations with canRead false). Return a redacted view for those agents: name, scope, and model only, with the description masked and instructions/skills/tools withheld. Readable agents are unchanged.


fabiencelier (approved):
LGTM

PR #28035, opened 2026-06-26
[front] Poke plugin to grant free credits to a member + free-seat allowance fixes

tdraier:
## Description

Adds a "Grant Free Credits to Member" poke plugin, invokable per-row from the members table in a workspace's Usage tab (free seats only): it raises the member's existing per-user free-seat credit's granted amount in Metronome via `v2.contracts.editCredit` (no invoice, no second credit) and refreshes their balance alerts. Because free-seat allowances had been a fixed `FREE_SEAT_LIFETIME_AWU_CREDITS` constant in several places, the poke members table, the customer usage-page "Credits usage" column, and the personal-settings "Your Credits" bar now all read each free member's real allowance/balance from their per-user credit (fixing displays like `1798/500`, `500/500`, and `NaN/500` → `2/1800`). Also hardens `AwuUsageBar` against a missing balance (`!= null`) and hides the "Edit spend limit" row action for free/none/seatless members, who have no pool to cap.

## Tests

Manual testing in poke and the customer Usage/Settings surfaces (grant flow, balance + allowance/cap consistency across all three gauges, menu visibility per seat type); type-check (`tsgo`) and biome pass.

## Risk

Low and free-seat-scoped: paid/none seat rows produce byte-identical display output, the plugin and its Metronome helpers only run for free-seat credits, and the only broader change is one extra `credits.list` read on the customer usage page when free seats are present (skipped otherwise). Safe to roll back (revert the branch).

## Deploy Plan

Standard front deploy, no migrations or config changes.

### DIFF
--- a/front/lib/api/poke/plugins/workspaces/grant_user_free_credits.ts
+++ b/front/lib/api/poke/plugins/workspaces/grant_user_free_credits.ts
@@ -1,7 +1,7 @@
 import { createPlugin } from "@app/lib/api/poke/types";
 import { upsertPerUserCreditBalanceAlerts } from "@app/lib/metronome/alerts/per_user_credit_balance";
 import {
-  adjustCustomerCreditBalance,
+  editCustomerCreditSegmentAmount,
   findPerUserCustomerCreditSegment,
 } from "@app/lib/metronome/client";
 import {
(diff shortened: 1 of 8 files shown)


front/lib/api/poke/plugins/workspaces/grant_user_free_credits.ts
```diff
@@ -0,0 +1,194 @@
...
+      return new Err(new Error(fromError(validationResult.error).toString()));
+    }
+    const { userId, amountCredits } = validationResult.data;
+
+    const workspaceResource = await WorkspaceResource.fetchById(workspace.sId);
+    if (!workspaceResource) {
+      return new Err(new Error(`Workspace not found: wId='${workspace.sId}'`));
+    }
+
+    const { metronomeCustomerId } = workspaceResource;
```
rfrenoy: isn't that already available in the workspace arg?

rfrenoy (approved):
Small comment, otherwise LGTM

front/lib/api/poke/plugins/workspaces/grant_user_free_credits.ts
tdraier: Good catch — `metronomeCustomerId` is on `LightWorkspaceType`, which is exactly the plugin's `workspace` arg type. Dropped the redundant `WorkspaceResource.fetchById` and now read it straight from `workspace` in bcfdf7ecf1.

PR #28042, opened 2026-06-26
[front] chore: add index to speedup agent message skill lookups

aubin-tchoi:
## Description

- Adds an index on `agent_message_skills` to speed up this query: https://github.com/dust-tt/dust/blob/main/front/lib/resources/skill/skill_resource.ts#L3410 (`listAgentMessageSkillsByCustomSkills`, used for skill reinforcement).
- Breakdown of affected queries, scan cost change and write overhead in https://app.pganalyze.com/databases/-642267661/checks/index_advisor/indexing_engine/36722381.
- A partial index here gives the same usefulness for the query, with a lower storage/cache/write overhead, plus we always query with a non null custom skill ID.

## Tests

- Migration tested locally.

## Risk

- Low.

## Deploy Plan

- Run migration.
- Deploy front.

### DIFF
--- a/front/lib/models/skill/conversation_skill.ts
+++ b/front/lib/models/skill/conversation_skill.ts
@@ -189,9 +189,9 @@ AgentMessageSkillModel.init(
         concurrently: true,
       },
       {
-        fields: ["workspaceId", "customSkillId"],
+        fields: ["customSkillId"],
         where: { customSkillId: { [Op.ne]: null } },
-        name: "idx_agent_message_skills_workspace_custom_skill",
+        name: "idx_agent_message_skills_custom_skill",
         concurrently: true,
       },
     ],
--- a/front/lib/models/skill/conversation_skill.ts
+++ b/front/lib/models/skill/conversation_skill.ts
@@ -188,6 +188,12 @@ AgentMessageSkillModel.init(
         name: "idx_agent_message_skills_agent_message_id",
         concurrently: true,
       },
+      {
+        fields: ["workspaceId", "customSkillId"],
+        where: { customSkillId: { [Op.ne]: null } },
+        name: "idx_agent_message_skills_workspace_custom_skill",
+        concurrently: true,
+      },
     ],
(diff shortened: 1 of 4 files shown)


flvndvd (approved):

front/lib/models/skill/conversation_skill.ts
```diff
@@ -188,6 +188,12 @@ AgentMessageSkillModel.init(
         name: "idx_agent_message_skills_agent_message_id",
         concurrently: true,
       },
+      {
+        fields: ["workspaceId", "customSkillId"],
+        where: { customSkillId: { [Op.ne]: null } },
```
flvndvd, line 193: I approved a bit too fast.

If we are trying to improve this query:
```
    const records = await AgentMessageSkillModel.findAll({
      attributes: [
        "createdAt",
        "conversationId",
        "customSkillId",
        "agentConfigurationId",
      ],
      where: {
        workspaceId: workspace.id,
        customSkillId: { [Op.in]: [...skillsById.keys()] },
      },
    });
```

Then we need to add the not null to the SQL query as well otherwise PG won't use the index IIRC.

id13:
@aubin-tchoi 🔥

I think workspaceId can be ditched, pganalyze doesn’t know that a custom skill only belongs to one workspace

front/lib/models/skill/conversation_skill.ts
aubin-tchoi: added

aubin-tchoi:
@id13 don't you think that in this case it can be useful for pg to filter by workspace first if there are many custom skill IDs?

id13:
pg creates a concatenation of columns in the index for its keys, and the btree will be split according to this key, so not really going to speed  up imo

id13:
Depends on the access pattern ofc but for an fk col it’s always going to be =

aubin-tchoi:
makes sense actually (forgot that was an FK), updated

PR #28043, opened 2026-06-26
Improve Google Sheets tool descriptions for tool-search recall

flvndvd:
## Description

⚠️ Server soon to be deprecated. Seen with @smb2268 and @frankaloia.

This sharpens the Google Sheets tool descriptions so the right one surfaces when tool search ranks them by how well a user's phrasing matches each tool's name, description, and parameters, and it registers the server in the offline diagnostic so we can keep checking that. Reading a range of values now leads with reading values, formatting now names the things people actually ask for such as bold text and colors, and finding a spreadsheet by name now leads with searching and no longer carries an oversized parameter explanation that was pushing it down the ranking. One sibling that repeated the word spreadsheet was trimmed so it stops crowding out the listing tool. With these changes every Google Sheets phrasing we test retrieves the intended tool. A handful of intents around creating and listing spreadsheets remain genuinely ambiguous because the Drive integration exposes its own overlapping spreadsheet tools, and since the ranking sees only the tool and not which integration it belongs to, those cannot be separated by wording alone. Those are marked as tolerant in the diagnostic with a note, and the durable fix is to consolidate the overlapping spreadsheet tools rather than to keep tuning text.

### DIFF
--- a/front/lib/api/actions/servers/google_sheets/metadata.ts
+++ b/front/lib/api/actions/servers/google_sheets/metadata.ts
@@ -9,13 +9,13 @@ export const GOOGLE_SHEETS_TOOL_NAME = "google_sheets" as const;
 export const GOOGLE_SHEETS_TOOLS_METADATA = createToolsRecord({
   list_spreadsheets: {
     description:
-      "List Google Sheets spreadsheets accessible by the user from both personal drive and shared drives. Supports pagination and search.",
+      "Search for and list your Google Sheets spreadsheets by name, across your personal and shared drives, to find a spreadsheet when you do not know its ID.",
     schema: {
       nameFilter: z
         .string()
         .optional()
         .describe(
-          "The text to search for in file names. Uses Google Drive's 'contains' operator which is case-insensitive and performs prefix matching only. For example, searching 'hello' will match 'HelloWorld' but not 'WorldHello'."
+          "Text to search for in spreadsheet names (case-insensitive prefix match)."
         ),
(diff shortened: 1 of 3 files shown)


front/lib/api/actions/servers/google_sheets/metadata.ts
```diff
@@ -9,13 +9,13 @@ export const GOOGLE_SHEETS_TOOL_NAME = "google_sheets" as const;
...
   list_spreadsheets: {
     description:
-      "List Google Sheets spreadsheets accessible by the user from both personal drive and shared drives. Supports pagination and search.",
+      "Search for and list your Google Sheets spreadsheets by name, across your personal and shared drives, to find a spreadsheet when you do not know its ID.",
     schema: {
       nameFilter: z
         .string()
         .optional()
         .describe(
-          "The text to search for in file names. Uses Google Drive's 'contains' operator which is case-insensitive and performs prefix matching only. For example, searching 'hello' will match 'HelloWorld' but not 'WorldHello'."
```
aubin-tchoi, line 18: shouldn't we keep the example?

aubin-tchoi (approved):
LGTM

front/lib/api/actions/servers/google_sheets/metadata.ts
flvndvd: since it's soon to be retired (see thread [here](https://dust4ai.slack.com/archives/C050SM8NSPK/p1782487972544409)) won't sweat it too much,

PR #28045, opened 2026-06-26
Introduce SandboxFunctionResource

spolu:
## Description

Follow-up from: https://github.com/dust-tt/dust/pull/28024

Introduce basic scaffolding for the `SandboxFunctionResource`.

Deletion by pods is not super optimized but at least fully implemented. Optimization is likely slightly involved given the number of models hidden behind FileResource. Right now, sequential delete which will be peaceful on the DB but potentially a bit slow for the user. The deleteByPod should only happen on pod deletion which is rare enough.
 
## Tests

Some tests, but mostly N/A

## Risk

None

## Deploy Plan

N/A

### DIFF
--- a/front/lib/resources/sandbox_function_resource.ts
+++ b/front/lib/resources/sandbox_function_resource.ts
@@ -1,7 +1,7 @@
 import type { Authenticator } from "@app/lib/auth";
 import { BaseResource } from "@app/lib/resources/base_resource";
 import { FileResource } from "@app/lib/resources/file_resource";
-import type { SpaceResource } from "@app/lib/resources/space_resource";
+import { SpaceResource } from "@app/lib/resources/space_resource";
 import { SandboxFunctionModel } from "@app/lib/resources/storage/models/sandbox_function";
 import type { ReadonlyAttributesType } from "@app/lib/resources/storage/types";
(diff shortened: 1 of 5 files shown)


front/lib/resources/sandbox_function_resource.ts
```diff
@@ -0,0 +1,193 @@
...
+    options?: ResourceFindOptions<SandboxFunctionModel>
+  ): Promise<SandboxFunctionResource[]> {
+    const { where, ...rest } = options ?? {};
+    const sandboxFunctions = await this.model.findAll({
+      where: {
+        ...where,
+        workspaceId: auth.getNonNullableWorkspace().id,
+      },
+      ...rest,
+    });
```
flvndvd, line 120: Shouldn't we fetch the associated pod and ensure you can read at least?

flvndvd (approved):

front/lib/resources/sandbox_function_resource.ts
spolu: Fair. Will add that 👍

spolu:
@flvndvd PTAL

Wil rename podId / pod to spaceId / space in subsequent PR because it just make sense to not use pod name here!

spolu:
PMRR

PR #28062, opened 2026-06-26
Instrument Anthropic BM25 tool search queries

flvndvd:
## Description

We recently introduced Anthropic's tool search tool behind a FF, deferring cold MCP tools so the model discovers them on demand via BM25. We had no visibility into what the model actually searches for, so we couldn't tell whether deferred tools are discoverable or how to tune their names and descriptions for recall.

This adds logging of the natural-language query the model issues on each BM25 search, the tools it discovers in return, and search errors, plus a StatsD counter of search volume broken down by model and provider. Search-result blocks are now consumed properly instead of being treated as something to surface to the client.

### DIFF
--- a/front/lib/model_constructors/sdk/anthropic_ai/converters/output/tool_search_logging.ts
+++ b/front/lib/model_constructors/sdk/anthropic_ai/converters/output/tool_search_logging.ts
@@ -46,11 +46,11 @@ export function logToolSearchQuery({
 
   logger.info(
     {
+      ...logFields,
       toolName,
       query,
       // Keep the raw payload only when parsing failed, to debug malformed input.
       rawInput: query === undefined ? rawInput : undefined,
-      ...logFields,
     },
     "Anthropic tool search query"
   );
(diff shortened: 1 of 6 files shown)


front/lib/model_constructors/sdk/anthropic_ai/converters/output/tool_search_logging.ts
```diff
@@ -0,0 +1,90 @@
...
+    query = parsed.value.query;
+  }
+
+  logger.info(
+    {
+      toolName,
+      query,
+      // Keep the raw payload only when parsing failed, to debug malformed input.
+      rawInput: query === undefined ? rawInput : undefined,
+      ...logFields,
```
aubin-tchoi: nit: could destructure first before the other args

aubin-tchoi (approved):

flvndvd:
Works pretty well, see [here](https://app.datadoghq.eu/notebook/320610?cell_id=hlkxqblp).

PR #28069, opened 2026-06-26
Add a tool-search prompt hint in the Anthropic client when the search tool is present

flvndvd:
## Description

Behind the `anthropic_tool_search` flag, cold MCP tools are deferred and the model is supposed to discover them through Anthropic's tool search tool. In practice it often skips the search and guesses or answers from memory, so this PR adds a system-prompt hint telling it to search for tools before answering.

The hint is injected in the Anthropic client layer and gated on whether a tool search tool actually ends up in the request, derived from the converted tools array rather than re-derived from a feature flag upstream. This is the only place that already knows for certain whether the search tool is being sent, including the case where a forced tool gets un-deferred and no search tool is added. As a result the hint is present if and only if the search tool is present in the same request, so it never reaches providers that expose all tools at once and have no search tool, and it cannot drift out of sync with the deferral decision.

The hint sits in the shared cache tier for now. With tool search on there is effectively always a deferred tool, so it could live in the longer-lived instructions tier, but conditional attachment-driven tools are hot and still force that tier to downgrade. A TODO marks moving it up once the hot tool set is reduced to a minimal curated set.

Applied to both the live client and the new model_constructors client for parity.

### DIFF
--- a/front/lib/api/llm/clients/anthropic/index.test.ts
+++ b/front/lib/api/llm/clients/anthropic/index.test.ts
@@ -1,0 +1,49 @@
+import { buildSystemBlocks } from "@app/lib/api/llm/clients/anthropic/index";
+import type { StructuredSystemPrompt } from "@app/lib/api/llm/types/options";
+import { TOOL_SEARCH_INSTRUCTION } from "@app/lib/model_constructors/sdk/anthropic_ai/converters/input/tool_search";
+import { describe, expect, it } from "vitest";
+
+const prompt: StructuredSystemPrompt = {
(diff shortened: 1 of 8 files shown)


flvndvd:
PMRR.

spolu:
ooc what is tool_search_tool_bm25 ? an Anthropic specific construct?

flvndvd:
> ooc what is tool_search_tool_bm25 ? an Anthropic specific construct?

Yes, more details [here](https://platform.claude.com/docs/en/agents-and-tools/tool-use/tool-search-tool).

spolu:
neat

PR #28071, opened 2026-06-26
Improve Frame tool descriptions for tool-search recall

flvndvd:
## Description

This improves how the Frame server's tools are retrieved by Anthropic's tool search, which ranks tools lexically over each tool's name, description, and input schema. The most important problem was that users (and us) call these artifacts "Frame"everywhere in the product, yet every tool description called them "Interactive Content file", a phrase nobody types. Because the word Frame only lived in the display labels, which the index never sees, requests phrased around a frame had nothing distinctive to match and the whole server ranked low. Each description now leads with its action verb and the word Frame so intent maps to the right tool.

The second problem was internal competition. The create tool had a long, keyword-dense description that acted as a magnet, and it even mentioned editing, so edit, read, and rename requests were all being pulled toward create instead. The edit tool was the longest in the server and repeated the same generic words many times, which under length normalization buried it even though it held the only terms that actually distinguished it, like changing a chart or its colors. Tightening create so it stops advertising other actions, and trimming edit down to its meaningful vocabulary while keeping the exact-match editing mechanics intact, resolved the collisions.

With these changes every Frame intent now retrieves its correct tool as the top result, up from roughly two thirds before, with no relaxed ranking tolerance needed. The change also registers the Frame server and its labeled queries in the existing retrieval diagnostic so this coverage is exercised alongside the other reviewed servers.

### DIFF
--- a/front/lib/api/actions/servers/interactive_content/metadata.ts
+++ b/front/lib/api/actions/servers/interactive_content/metadata.ts
@@ -32,11 +32,11 @@ export const EXPORT_INTERACTIVE_CONTENT_FILE_TOOL_NAME =
 export const INTERACTIVE_CONTENT_TOOLS_METADATA = createToolsRecord({
   [CREATE_INTERACTIVE_CONTENT_FILE_TOOL_NAME]: {
     description:
-      "Create a new Interactive Content file that users can execute or interact with. Use this for " +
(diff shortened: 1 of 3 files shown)


flvndvd:
PMRR.

PR #28072, opened 2026-06-26
Default tools to deferred with a per-tool eager opt-in for Anthropic tool search

flvndvd:
## Description

See data points in https://github.com/dust-tt/tasks/issues/9161.

> [!NOTE]
> Will add some tests to enforce number of eager tools in a follow up PR.

Previously the agent loop decided per server whether a tool was loaded upfront, derived from each internal server's availability. That conflated "attached by default" with "kept in the cached prefix" and made the loaded-upfront set impossible to curate, since it was whatever happened to be auto availability.

Tools now default to deferred behind tool search, and a tool opts in to being loaded upfront with an explicit, provider-agnostic property on its own definition. This makes the upfront set a deliberate, minimal choice instead of a side effect of availability. The opt-in set is the web search and browse tools, file reading and listing, data source search, skill enabling, the sandbox shell, and the toolset listing and enabling tools so the agent can always discover and turn on more capabilities while everything else loads on demand. All tools part of the eager set, are **ALWAYS** available.

Whether a non-eager tool is actually deferred is an Anthropic concern, so that decision now lives entirely in the Anthropic clients rather than leaking into provider-agnostic code. Specifications carry only the intrinsic eager property, and the feature gate is threaded down as a neutral signal that providers without tool search ignore. The wire-level deferral field no longer appears anywhere outside the Anthropic clients.

Behavior is unchanged when tool search is off, where every tool is sent upfront as before.

## Risk

Worst case, agent misses obvious tool usage, will monitor, gated behind a FF and safe to disabled if needed.


front/lib/actions/mcp_internal_actions/tool_definition.ts
```diff
@@ -47,6 +47,9 @@ export interface ToolDefinition<
 > {
   name: TName;
   enableAlerting?: boolean;
+  // When true, the tool is kept in the cached tools prefix (loaded upfront)
+  // instead of being deferred behind tool search. Defaults to deferred.
+  eager?: boolean;
```
ElPicador, line 52:
```suggestion
  eagerLoaded?: boolean;
```

to have a clearer name

ElPicador (approved):
Nice!

front/lib/model_constructors/sdk/anthropic_ai/converters/input/utils.ts
```diff
@@ -376,7 +376,10 @@ export function outputFormatToOutputConfig(outputFormat: OutputFormat): {
   };
 }
 
-export function toolSpecToAnthropicAITool(tool: ToolSpecification): Tool {
+export function toolSpecToAnthropicAITool(
+  tool: ToolSpecification,
+  toolSearchEnabled: boolean
```
aubin-tchoi, line 381: nit: would put this boolean in a structured object

front/lib/api/actions/servers/search/metadata.ts
```diff
@@ -22,6 +22,7 @@ export const SEARCH_TOOLS_METADATA = createToolsRecord({
     description: SEARCH_TOOL_DESCRIPTION,
     schema: SearchWithDataSourcesInputSchema.shape,
     stake: "never_ask" as const,
+    eager: true,
```
aubin-tchoi: I think you also want the one in `data_source_filesystem`

aubin-tchoi (approved):
LGTM

PR #28075, opened 2026-06-27
Fix Gemini fallback models that are no longer routable

flvndvd:
## Description

This PR fixes https://github.com/dust-tt/tasks/issues/9270.

The fast and small model selectors, conversation title generation, web page summarization, and the `dust-quick` global agents all fell back to Gemini 2.5 Flash (and 3 Flash in a couple of spots) for workspaces where only Google was whitelisted.

Since Google routing was narrowed to Vertex, those model ids no longer resolve to a client and any request that selected them failed at runtime with "Model not supported". This swaps every one of those fallbacks to Gemini 3.5 Flash, which is routable through both the legacy Vertex path and the new router, restoring builder suggestions, title generation, and summarization for Google-only workspaces. Note this only fixes the internal selectors.

@adrsimon @pmilliotte some thoughts:
There is still an underlying gap here: a model can still pass the enablement check while having no routable endpoint, so an agent explicitly configured on an unroutable Gemini will still fail. Worth a follow-up to align model availability with actual routability.


flvndvd:
PMRR.

PR #28081, opened 2026-06-28
Add sandbox_upload.ts operator script for uploading files/folders

davidebbo:
## What

Adds `front/scripts/sandbox_upload.ts`, an operator-only helper (mirroring `front/scripts/sandbox_exec.ts`) for uploading a single file or an entire folder into a running E2B sandbox.

The stock `e2b` CLI has no file-transfer command. Piping a file through `e2b sandbox exec` works for text but is **not binary-safe** (raw binary piping hangs; base64-over-exec works but inflates ~33% and is awkward for folders). This script uses the E2B SDK's `sandbox.files.write(entries[], { user })` directly:

- **Binary-safe** — uploads exact `ArrayBuffer` bytes, no base64.
- **Single file or recursive folder** — folders are walked recursively (symlinks/sockets skipped), written in batches of 25 to bound memory; parent dirs are created automatically.
- **Sensible default user** — defaults to `agent` (the workload user) so uploaded files are usable by the sandbox workload and readable via `e2b sandbox exec` (which also runs as `agent`). Defaulting to `root` would land files in `/home/agent` owned by root and cause "Permission denied" for the agent. Pass `-u root` for root-only locations like `/etc`.

## Usage

```bash
# Single file -> explicit destination
npx tsx scripts/sandbox_upload.ts -s "$SANDBOX_ID" ./local.txt /home/agent/dest.txt
# Single file -> into a directory (trailing slash keeps basename)
npx tsx scripts/sandbox_upload.ts -s "$SANDBOX_ID" ./local.txt /home/agent/dir/
# Entire folder (recursive)
npx tsx scripts/sandbox_upload.ts -s "$SANDBOX_ID" ./localdir /home/agent/dest
# Root-only location
npx tsx scripts/sandbox_upload.ts -s "$SANDBOX_ID" -u root ./ca.pem /etc/ca.pem
```

## Testing

Verified end-to-end against a live sandbox:
- Single file round-trips; folder upload sha256 matches byte-for-byte.
- Files owned by `agent` (mode 660) with the default; `-u root` writes to `/etc` and reads back correctly.

## Notes / limitations

- Upload-only and content-only: does not delete remote files absent from the source (not a true rsync) and does not preserve local file modes (sandbox umask applies).

### DIFF
--- a/front/scripts/sandbox_upload.ts
+++ b/front/scripts/sandbox_upload.ts
@@ -1,0 +1,150 @@
+#!/usr/bin/env tsx
+import { readFileSync, statSync } from "node:fs";
+import { readdir } from "node:fs/promises";
+import { basename, join as joinLocal, posix } from "node:path";
+import { parseArgs } from "node:util";
+
+import config from "@app/lib/api/config";
+import logger from "@app/logger/logger";
+import { Sandbox } from "e2b";
+
+// Operator-only escape hatch for uploading files/folders into a sandbox during manual debugging.
+// Mirrors scripts/sandbox_exec.ts. Runtime app code must route sandbox file writes through
+// SandboxProvider, not this script.
+
+// Number of files written per batch. Bounds how much file data we hold in memory at once when
+// uploading large folders, while still amortizing request overhead across many small files.
+const UPLOAD_BATCH_SIZE = 25;
+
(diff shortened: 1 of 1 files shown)


id13:
LGTM

PR #28082, opened 2026-06-28
Add dust-functions: serverless-style handler runner + schema discovery

davidebbo:
## Description

Adds `cli/dust-functions`, a Bun-based system for running Web-standard HTTP handlers as isolated, serverless-style invocations (no server) and discovering their I/O contracts for model tool-use.

Two CLIs:

- **`run_request <handler.ts> < request.json`** — reads a request as JSON on stdin, builds a real `Request`, calls the handler's `export default { fetch }`, and serializes the `Response` to stdout. The output is `{ ok: true, response }` for any handler that ran (a 404/500 is still `ok: true`) or `{ ok: false, error }` for an invocation failure (`import_failed` / `threw` / `bad_return` / `bad_input`). If the handler exports a Zod `schema.input`, the request body is validated against it first and a `400` is returned on failure — handlers can trust their input.
- **`discover <folder>`** — scans a folder of handlers and emits each one's input/output JSON Schema (converted from the handler's Zod `schema` export), ready to drop into a tool-use / function-calling API (`name` / `description` / `input_schema`). Never crashes on a bad file: non-handlers, missing/invalid schemas, and import errors are reported under `skipped`.

The same Zod schema drives both `discover` (model-facing contract) and `run_request` runtime validation — one source of truth.

Runs on Bun (handlers are plain `.ts`, executed natively), consistent with `cli/dust-sandbox/e2e`. Also enrolls `cli/dust-functions` in the biome linter/assist `includes` (mirroring `cli/dust-cli`).

## Tests

`bun test` in `cli/dust-functions` — 39 tests across protocol encoding/decoding, the invoker (success, 404, throw, missing/broken handler, bad return, binary bodies, body pass-through, relative-path resolution, input validation), and discovery (cataloging, name derivation, all skip reasons, `*.test.ts` ignoring, non-directory rejection), including CLI end-to-end tests that spawn the tools. `biome check` is clean.

## Risk

Low / isolated. New self-contained package under `cli/`; no existing code imports it and it is not wired into any build or runtime path. The only shared-file change is adding `cli/dust-functions/**` to `biome.json` lint/assist `includes`, which only affects linting of the new package. Fully rollback-safe (delete the directory + revert the biome.json hunk). Not added to root npm `workspaces` (it is a standalone Bun package with its own lockfile).

## Deploy Plan

None — developer tooling only, nothing deployed.


fontanierh (commented):
I think this should be part of our `dsbx` binary that we already ship on the sandbox base image: it's a rust codebase that contains all the dust code that we run on sandboxes.

I would recommend creating a new subcommand named `functions`

cli/dust-functions/run_request.ts
```diff
@@ -0,0 +1,196 @@
...
+      : resolve(process.cwd(), handlerPath);
+    const mod = await import(spec);
+    const def = mod.default;
+    if (typeof def?.fetch !== "function") {
+      throw new Error(
+        "handler must `export default { fetch(req) {...} }` with a fetch function"
+      );
+    }
+    handler = def;
+    const declared = (mod.schema as { input?: unknown } | undefined)?.input;
```
flvndvd, line 45: IMHO we should enforce this before a function can be published!

cli/dust-functions/fixtures/catalog/greet.ts
```diff
@@ -0,0 +1,23 @@
+import { z } from "zod";
+
+export const schema = {
```
flvndvd, line 3: OOC, how good was the model at writing this web-standard pattern?

davidebbo: I asked it to write simple add/multiply functions and it got it right. My prompt had this function as a full sample. Didn't try anything too fancy though.

flvndvd: Do we plan to also expose result schema? so the down stream client knows how to consume the function results?

cli/dust-functions/run_request.ts
davidebbo: Agreed, this should fail the publish

cli/dust-functions/fixtures/catalog/greet.ts
davidebbo: Yes, it's there in the `output` field.

davidebbo:
Now using https://github.com/dust-tt/dust/pull/28088 instead.

PR #28084, opened 2026-06-28
Add generic module bundler engine with Frame wrapper

flvndvd:
## Description

This PR introduces a storage- and consumer-agnostic bundler that takes a multi-file source tree and produces a single self-contained module. It resolves and inlines relative imports (with extension and directory-index probing), keeps non-relative specifiers external for the runtime to resolve, rejects imports that escape the module root, and reads sources through an async interface so it can run over any backend at any nesting depth.

The engine is deliberately decoupled from frames. The two things that genuinely differ per consumer are explicit inputs: the esbuild output options (a frame builds for the browser with JSX preserved and no minification, a sandbox Function will build for node) and an optional per-file transform, which is the only consumer-specific seam. This is what lets us reuse the exact same bundling logic for the upcoming sandbox Functions without forking it.

Frames are layered on top as a thin wrapper that supplies the viz esbuild options and a JSX source-location transform, so a published frame keeps the origin of every element for live editing.

This PR is the inert foundation: nothing in production imports the engine yet. The publish pipeline and render path that consume it land in the next PR.

### DIFF
--- a/front/lib/api/bundler/bundle_module.ts
+++ b/front/lib/api/bundler/bundle_module.ts
@@ -56,7 +56,10 @@ export interface BundleModuleParams {
   transform?: (relPath: string, content: string) => string;
 }
 
-// Extensions probed when a relative import omits one (e.g. `./Chart` -> `./Chart.tsx`).
+// Extensions probed, in order, when a relative import omits one (e.g. `./Chart` -> `./Chart.tsx`).
+// Resolution is first-match-wins by this order, never an ambiguity error, so the order is the
+// tie-breaker when sibling files share a base name (`./Chart` resolves to Chart.tsx over Chart.ts).
+// The leading "" matches a specifier that already carries its extension.
 const RESOLVE_EXTENSIONS = ["", ".tsx", ".ts", ".jsx", ".js", ".json"] as const;
 const BUNDLE_NAMESPACE = "bundle";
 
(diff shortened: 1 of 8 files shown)


front/lib/api/bundler/bundle_module.ts
```diff
@@ -0,0 +1,262 @@
...
+  /**
+   * Per-file transform applied before a file is handed to esbuild, the engine's only
+   * consumer-specific seam. Frames use it to stamp JSX source-location tags. Receives the
+   * root-relative path and raw contents, returns the (possibly rewritten) contents.
+   */
+  transform?: (relPath: string, content: string) => string;
+}
+
+// Extensions probed when a relative import omits one (e.g. `./Chart` -> `./Chart.tsx`).
+const RESOLVE_EXTENSIONS = ["", ".tsx", ".ts", ".jsx", ".js", ".json"] as const;
```
fontanierh, line 63: is array order tie breaking here ? Or is it an error if ambiguous?  comment could clarify

front/lib/api/bundler/bundle_module.ts
```diff
@@ -0,0 +1,262 @@
...
+  }
+
+  if (relPath.endsWith(".json")) {
+    return "json";
+  }
+
+  return "tsx";
+}
+
+function isEsbuildFailure(
```
fontanierh, line 90: no zod :p ?

front/lib/api/bundler/bundle_module.ts
```diff
@@ -0,0 +1,262 @@
...
+  // file. Returns the resolved root-relative path or an error marker.
+  const resolveRelative = (
+    importerRel: string,
+    spec: string
+  ): { rel: string } | { error: "escape" | "missing" } => {
+    const joined = path.posix.normalize(
+      path.posix.join(path.posix.dirname(importerRel), spec)
+    );
+    // TODO(FRAMES BUNDLE): Consider relaxing this to allow imports from pods/ in the conversation.
+    if (joined.startsWith("..") || path.posix.isAbsolute(joined)) {
```
fontanierh, line 156: oh so this is actually going to run _on_ the sandbox ? nice !

fontanierh (approved):
Looks very solid

front/lib/api/bundler/bundle_module.ts
flvndvd: Yes 🚀

front/lib/api/bundler/bundle_module.ts
flvndvd: Indeed updated comment and code to add more clarity.

spolu:
How is the "module root" of "escape the module root" defined. If I have a file.ts as root of the build that imports ../lib/cool.ts that seems like a good use case for us? would it get blocked here? 

LGTM!

flvndvd:
> How is the "module root" of "escape the module root" defined. If I have a file.ts as root of the build that imports ../lib/cool.ts that seems like a good use case for us? would it get blocked here?

Yes for now it's only partially supported, so we use the frame root folder as temporary root, the goal is to open up more broadly. It requires some extra plumbing so that we can resolve both `/files/pods` and `/files/conversations`.

spolu:
> Yes for now it's only partially supported, so we use the frame root folder as temporary root, the goal is to open up more broadly. It requires some extra plumbing so that we can resolve both /files/pods and /files/conversations.

Roger that!

PR #28085, opened 2026-06-28
Add Frame publish pipeline and render update behind frame_publish flag

flvndvd:
## Description

Stacked on the bundler PR (see https://github.com/dust-tt/dust/pull/28084). Wires the bundler into a real publish flow for Frames and makes the rendered Frame serve the built output.

Today a model edits a Frame's files directly in the Computer mount, but those edits never reach the canonical file that renders, so they are effectively lost. This change adds a publish step that reads the Frame's sources from the mount by following the entry's import graph, validating each file as it loads, builds them into a single bundle, and stores that bundle as the rendered version while refreshing the canonical source. Only files the Frame actually imports are read, so unrelated files
in the directory neither slow publishing nor block it. One file stays the unit of sharing, so the existing share token, URL, and access allowlist keep working unchanged. Rendering now serves the built bundle once a Frame has been published and falls back to the original source otherwise, with an explicit version request still taking precedence.

Publishing is exposed as a new tool on the interactive content server and is gated behind a new `frame_publish` feature flag. The tool is filtered out of the toolset for any workspace without the flag, so this can roll out gradually with no effect on existing workspaces. Gating is runtime only: the statically declared tool stake is unchanged.

The human live-edit path that lets someone edit rendered text and have it flow back through the
bundle is intentionally not in this PR. It builds on this pipeline and lands next.

### DIFF
--- a/front/lib/actions/mcp_internal_actions/__snapshots__/mcp_servers_metadata.test.ts.snap
+++ b/front/lib/actions/mcp_internal_actions/__snapshots__/mcp_servers_metadata.test.ts.snap
@@ -327,6 +327,7 @@ exports[`MCP Servers Metadata Snapshot > should have stable tool stakes across a
     "edit_interactive_content_file": "never_ask",
     "export_interactive_content_file": "never_ask",
     "get_interactive_content_file_share_url": "never_ask",
+    "publish_interactive_content_file": "never_ask",
     "rename_interactive_content_file": "never_ask",
     "retrieve_interactive_content_file": "never_ask",
     "revert_interactive_content_file": "never_ask",
--- a/front/lib/api/actions/servers/interactive_content/metadata.ts
+++ b/front/lib/api/actions/servers/interactive_content/metadata.ts
@@ -254,14 +254,14 @@ export const INTERACTIVE_CONTENT_TOOLS_METADATA = createToolsRecord({
   },
   [PUBLISH_INTERACTIVE_CONTENT_FILE_TOOL_NAME]: {
     description:
-      "Publish a Frame from its source files in the Computer. After creating or editing a " +
-      "Frame's `.tsx` source files directly in the Computer, call this to build them into the " +
-      "live Frame. It resolves the Frame's dependency tree from the given directory (inlining " +
-      "relative imports), validates TypeScript and JSX, then updates the canonical Frame so " +
-      "viewers and shares see the new version. A syntax error blocks publishing and is reported " +
-      "back so you can fix it. Tailwind warnings are returned but do not block. Pass the " +
-      "`file_id` of the Frame you created via the Frames tools and the scoped path of the " +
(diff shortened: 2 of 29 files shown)


front/lib/api/actions/servers/interactive_content/metadata.ts
```diff
@@ -250,6 +252,41 @@ export const INTERACTIVE_CONTENT_TOOLS_METADATA = createToolsRecord({
       done: "Export Frame",
     },
   },
+  [PUBLISH_INTERACTIVE_CONTENT_FILE_TOOL_NAME]: {
```
flvndvd, line 255: Will add the new instructions in the frame skill when removing FF.

front/lib/api/viz/publish_frame.ts
```diff
@@ -0,0 +1,189 @@
...
+      const entrySource = cache.get(file.fileName);
+      if (entrySource !== undefined) {
+        await file.uploadContent(auth, entrySource);
+      }
+
+      // 3. Store the bundle as the rendered version and mark the frame published.
+      await file.uploadProcessed(auth, buildResult.value.code);
+      // frameBundleRootPath flips rendering to the bundle, and live edits later reuse it to
+      // relocate the source tree in the mount and rebuild without a model in the loop.
+      await file.setUseCaseMetadata(auth, {
```
flvndvd, line 164: Definitely not a big fan of doing this, will consider adding a dedicated light data model for frames (a bit like what's done for the `SandboxFunction`.

front/lib/api/actions/servers/interactive_content/metadata.ts
flvndvd: Done in: https://github.com/dust-tt/dust/pull/28087

front/lib/actions/mcp_internal_actions/__snapshots__/mcp_servers_metadata.test.ts.snap
```diff
@@ -327,6 +327,7 @@ exports[`MCP Servers Metadata Snapshot > should have stable tool stakes across a
     "edit_interactive_content_file": "never_ask",
     "export_interactive_content_file": "never_ask",
     "get_interactive_content_file_share_url": "never_ask",
+    "publish_interactive_content_file": "never_ask",
```
fontanierh, line 330: Interesting that we're going for never_ask. I guess that we don't really have a choice because you can't preview before publishing

fontanierh (approved):
LGTM

front/lib/actions/mcp_internal_actions/__snapshots__/mcp_servers_metadata.test.ts.snap
flvndvd: Yes exactly, and since existing behavior already "publish" without permission.

PR #28086, opened 2026-06-28
Add live editing of published Frames through the bundle

flvndvd:
## Description

Stacked on the publish PR (see https://github.com/dust-tt/dust/pull/28085). Lets a person double-click rendered text in a published Frame, edit it in place, and have the change flow back to the source and re-render, with no model in the loop.

> [!NOTE]
> Will send a follow up PR for some performance improvements on the whole build and publish. 

When the publisher bundles a Frame it stamps each element with its source origin (file, line, column). The viz runtime reads that origin from the clicked element and sends it with the old and new text to the edit endpoint. The server routes the edit back to the source file by content, matching the text and using the location only to break ties between duplicates, so it stays correct even as earlier edits shift positions. It writes the updated source back to the mount and republishes
so the rendered bundle and the source stay in step. Frames that were never published keep using the existing text-matching edit path unchanged.

Editing by source location is the only robust option once a Frame is served as a bundle, because the rendered bytes no longer line up with any single source file. Matching by content rather than by position is what makes repeated edits safe.

### DIFF
--- a/viz/app/components/EditableFrame.tsx
+++ b/viz/app/components/EditableFrame.tsx
@@ -136,25 +136,44 @@ export function EditableFrame({ children }: EditableFrameProps) {
         setTimeout(() => target.classList.remove(...cls), FLASH_DURATION_MS);
       };
 
-      // Wrap rawText with surrounding source context so the server string-search is unique.
+      // Published (bundled) frames carry a `data-source` ("<relPath>:<line>:<col>") on each JSX
+      // element. When present, route the edit to the source file by location (robust against
+      // duplicated text and works through the bundle). Otherwise, fall back to the legacy
+      // context-string match against the rendered code.
+      const sourceEl = target.closest<HTMLElement>("[data-source]");
+      const source = sourceEl?.dataset.source;
+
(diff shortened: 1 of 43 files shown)


viz/app/components/EditableFrame.tsx
```diff
@@ -136,25 +136,44 @@ export function EditableFrame({ children }: EditableFrameProps) {
...
       const trailingWs = rawText.match(/\s*\n\s*$/)?.[0] ?? "";
       const newRawText = leadingWs + newVisibleText + trailingWs;
-      const oldText = ctxBefore + rawText + ctxAfter;
-      const newText = ctxBefore + newRawText + ctxAfter;
+
+      // Location edits send the visible (trimmed) text. The server preserves surrounding
+      // whitespace and disambiguates among the element's own text children by oldText.
+      const editParams = source
+        ? { oldText: originalVisibleText, newText: newVisibleText, source }
+        : (() => {
```
fontanierh, line 163: who are you and what did you do to the real flavien ?

fontanierh (approved):

PR #28088, opened 2026-06-28
dsbx function run/get: execute sandbox functions via embedded Bun runner

davidebbo:
## Description

Adds `dsbx function run <name>` and `dsbx function get <name>` to the sandbox CLI, so functions (Web-standard `fetch` handlers, shipped as self-contained Bun bundles) can be executed and introspected from within a sandbox.

- **`dsbx function run <name>`** — reads an HTTP request envelope as JSON on stdin (`{method?,url?,headers?,body?,encoding?}`; `url` defaults to `http://localhost/`), runs the handler, and writes the response as JSON on stdout (`{ok:true,response:{status,headers,body,encoding}}`, or `{ok:false,error:{kind,message,stack?}}` for an invocation failure — `import_failed|threw|bad_return|bad_input`). Any HTTP status (incl. 4xx/5xx) is a successful `ok:true`. If the function exports a Zod `schema.input`, the body is validated before the handler runs (→ `400`).
- **`dsbx function get <name>`** — prints the function's JSON-Schema contract `{name, description, input_schema, output_schema}` (for model tool-use). Enumeration is intentionally omitted (the function list is known by other means).

`name` resolves to `${DUST_FUNCTIONS_DIR}/<name>.ts` and is restricted to `[A-Za-z0-9_-]+` (path-traversal guard). Exit codes: `0` ok / `1` invocation-failure / `2` bad usage (missing dir/name/file). User-facing failures emit `{error}` JSON on stdout.

### How it works

Rust can't execute TypeScript, so the execution engine is a small Bun/TypeScript package (`cli/dust-sandbox/functions-runner/`: `protocol`, `invoke` + validation, single-function `schema` extraction, a `runner.ts` CLI). It is bundled with Zod inlined into `runner.js` (a **generated artifact, not committed** — `.gitignore`d), which `dsbx` embeds via `include_str!`, writes to a per-invocation temp file (random name, mode 0600), and runs with `bun`. The bundle is built with `bun run build` on the host before compiling `dsbx` — CI, the release workflow, and the upsert script all run that step first, and `cross` mounts the result, so `bun` is **not** needed inside the cross/musl build container. A `build.rs` fails early with instructions if the bundle is missing. `bun` is otherwise only a runtime dependency of `function run/get`.

## Tests

- **Rust (`cargo test`)**: 240 unit + 1 e2e passing — clap parsing of `function run|get`, name validation (rejects `../x`, `/abs`, `a/b`, empty), `DUST_FUNCTIONS_DIR` handling, missing-file errors (env-mutating tests serialized behind a mutex).
- **Runner (`bun test`)**: 31 passing across protocol (encode/decode), invoke (success, 404, throw, bad import, bad return, binary body, body pass-through, body validation incl. 400/non-JSON/non-Zod-skip), schema extraction, and the runner CLI (run/get/bad_input→exit 2/unknown-subcommand) via spawned processes.
- Manual e2e verified: `get` prints schema; `run` valid → 200; missing field → 400; missing function → `{error}` exit 1.
- biome clean (the package's TS sources are enrolled in the linter; the generated `runner.js` is hard-ignored).

## Risk

Moderate, well-contained. New self-contained module + a new clap subcommand; no existing `dsbx` behavior changes. The only shared-file edit is `biome.json` (enrolling the runner sources, ignoring the generated bundle). Security reviewed: path-traversal guard on `name`, and the runner temp file is written to a random 0600 path (no predictable-path/TOCTOU/symlink exposure). Rollback = revert the branch.

`runner.js` is no longer committed: CI and the release build regenerate it from source on every run, so there is no stale-bundle risk in CI. Local builds rebuild it via `bun run build` (a `build.rs` presence check fails the build with instructions if it's missing).

## Deploy Plan

No deploy. Developer/sandbox tooling shipped inside `dsbx`; the new subcommands require `bun` and `DUST_FUNCTIONS_DIR` to be present in the sandbox at runtime.


cli/dust-sandbox/functions-runner/protocol.ts
```diff
@@ -0,0 +1,96 @@
...
+  | { ok: false; error: InvocationError };
+
+export class BadInputError extends Error {}
+
+export function parseInput(raw: string): RequestInput {
+  let parsed: unknown;
+  try {
+    parsed = JSON.parse(raw);
+  } catch (e) {
+    throw new BadInputError(`stdin is not valid JSON: ${(e as Error).message}`);
```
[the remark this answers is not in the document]
davidebbo: Fixed in cef603b8b4 — now uses `e instanceof Error ? e.message : String(e)`. (Committed bundle `runner.js` regenerated too.)

cli/dust-sandbox/functions-runner/runner.ts
```diff
@@ -0,0 +1,57 @@
...
+}
+
+async function getHandler(handlerPath: string): Promise<number> {
+  try {
+    const schema = await getFunctionSchema(handlerPath);
+    process.stdout.write(`${JSON.stringify(schema)}\n`);
+    return 0;
+  } catch (e) {
+    process.stdout.write(
+      `${JSON.stringify({ error: (e as Error).message })}\n`
```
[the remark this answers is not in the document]
davidebbo: Fixed in cef603b8b4 — now uses `e instanceof Error ? e.message : String(e)`. (Committed bundle `runner.js` regenerated too.)

cli/dust-sandbox/src/commands/function/run.rs
```diff
@@ -0,0 +1,26 @@
...
+
+use anyhow::Result;
+use tokio::process::Command;
+
+use super::{emit_error, ensure_runner, resolve_existing};
+
+/// Execute a function. stdin (request envelope) and stdout (response JSON) are
+/// inherited so they stream straight through the runner; the runner's exit code
+/// becomes ours.
+pub async fn cmd_function_run(name: &str) -> Result<()> {
```
fontanierh, line 9: Can we make it run as skuid 1003? that's what we use for the agent code: runs unprivileged, with all network going through our egress proxy (which prevents reaching non-allowlisted domains, and also enables substitution of DSEC secrets).

we might need to run `dsbx function run` as a privileged user since it can access the bundles

davidebbo: Slightly confused here, as you seem to say both that it should run as untrusted and as trusted. 🤔 

I would add that `run` should only get readonly access to the bundles. Only publish should have full access.

fontanierh: I'm saying that `dsbx function run` would maybe be run as root (will be called by sandbox resource probably). But bun (harness + bundle) should run as 1003 🙏

fontanierh: So `Command::new("bun")...` should be downgraded to 1003

davidebbo: Makes sense, updated.

cli/dust-sandbox/src/commands/function/mod.rs
```diff
@@ -0,0 +1,312 @@
...
+#[cfg(target_os = "linux")]
+fn resolve_drop_target() -> Result<Option<DropTarget>> {
+    // SAFETY: geteuid only reads the caller's effective uid.
+    if unsafe { libc::geteuid() } != 0 {
+        return Ok(None);
+    }
+
+    // SAFETY: getpwuid returns a pointer into static storage valid until the
+    // next getpw* call; we read the fields we need immediately.
+    let pw = unsafe { libc::getpwuid(AGENT_UID as libc::uid_t) };
```
fontanierh, line 82: no way around the unsafes ?

fontanierh (approved):
Exciting 🔥

cli/dust-sandbox/src/commands/function/mod.rs
davidebbo: Let's leave this to a follow-up. Claude says it could be done with added dependencies:

geteuid/getpwuid/getgrouplist: these are the FFI unsafe blocks, and they can be made safe, but only with a crate:
- rustix is already in our tree (v1.1.4) and safely wraps geteuid — but it does not cover getpwuid/getgrouplist (those are libc database functions, not raw syscalls), so it can't replace the lookup.
- nix does wrap them: nix::unistd::User::from_uid(uid)?  and nix::unistd::getgrouplist(name, gid)?. That's a new dependency, but it'd make the whole lookup safe.

fontanierh: Thanks ! I think it's valuable if dsbx doesn't have any unsafe itself

davidebbo: Done in https://github.com/dust-tt/dust/pull/28174

PR #28090, opened 2026-06-28
Add sandbox_functions MCP server with list tool

flvndvd:
## Description

Introduces an internal MCP server for sandbox functions, exposing a single read-only list tool (for now) that returns the schema-typed functions published in the current pod together with their input and output schemas.

The server is hidden by default and gated behind the `sandbox_functions` feature flag, so it only surfaces for workspaces that have opted in. It resolves the pod from the conversation context and reads functions scoped to that pod.

This is the first step toward the broader sandbox functions capability. Publishing, which bundles a function on the pod's own sandbox and persists the built artifact, is intentionally left out and will land in a follow-up so this PR stays small and focused on read access.

**🔮 What's next**
- Add a `description` and `name` field on the `SanboxFunctionModel`
- Consolidate around exposing a single invocation handle (I guess sId)
- Adopt the lean-list plus detailed-get pattern: keep list lightweight (name, description, a compact signature) and move full JSON schemas behind a dedicated detail tool, so discovery stays cheap as the number and size of functions grow.


fontanierh (approved):
LGTM

PR #28099, opened 2026-06-29
Add scripts to seed provisioned groups.

Fraggle:
## Description

Adds `seed_directory_provisioned_groups.ts` to create fake SCIM/directory-provisioned `GroupResource` entries for local dev and testing. The script generates groups with fake `workOSGroupId` values, assigns random active workspace members via `dangerouslyAddMembers`, and optionally sets a fake `workOSOrganizationId` (prefixed `org_fake_local_dev_`) on the workspace via `WorkspaceResource.updateWorkOSOrganizationId`.

Extends `reset_directory_created_groups.ts` with an `--add-fake-work-os-organization-id` flag to also clear that fake org ID on cleanup. Extracts `getFakeWorkOSOrganizationId` and `isFakeWorkOSOrganizationId` helpers into `workspace_helpers.ts` to keep the fake-ID logic consistent across both scripts.

## Tests

Tested locally — ran seed script in dry-run then with `--execute`, verified groups appeared in DB, ran reset script to clean up.

## Risk

No production risk — scripts are local/dev tooling only. The `org_fake_local_dev_` prefix on fake org IDs makes them easily identifiable; `isFakeWorkOSOrganizationId` guards against accidentally clearing real WorkOS org IDs.

## Deploy Plan

No deploy needed — scripts-only change.

PR #28098, opened 2026-06-29
Add: methods to bucket models in tiers.

Fraggle:
## Description

Adds `front/lib/api/models_picker/tiers.ts`, the foundational tier-classification module for the model picker pricing initiative. Models are bucketed into four tiers (`fast`, `balanced`, `powerful`, `frontier`) based on blended cost (average of input + output pricing per million tokens), with thresholds at $0.5, $3, and $15 USD.

Key exports:
- `getModelTier(modelId)` — classifies a model via `MODEL_PRICING` lookup + `getModelTierForBlendedCost`
- `groupModelsByTier()` — partitions all picker models (`STATIC_MODEL_IDS` minus `NOOP_MODEL_ID`) into tier buckets
- `getModelsAllowedForTier(maxTier)` — returns a cumulative allowlist of all models at or below a tier ceiling, enabling workspace-level model access control

## Tests

Added Vitest unit tests in `tiers.test.ts` covering all exports: boundary conditions for `getModelTierForBlendedCost`, representative model classification for `getModelTier`, exhaustive partition coverage for `groupModelsByTier`, and allowlist ceiling enforcement for `getModelsAllowedForTier`.

## Risk

Low. Pure utility module — no DB changes, no API routes, no existing code modified. The only risk is incorrect tier assignment if `MODEL_PRICING` entries are missing or stale; `getModelTier` throws on missing entries, and the tests guard against misclassification.

## Deploy Plan

Deploy front.

PR #28100, opened 2026-06-29
Add models for model tier picker.

Fraggle:
## Description

Lays the data-model foundation for the model tier picker. Adds two new `WorkspaceAwareModel` tables — `group_model_tiers` (`GroupModelTierModel`) and `user_model_tiers` (`UserModelTierModel`) — that store per-(workspace, group) and per-(workspace, user) `ModelTier` overrides. Adds `defaultModelsTier: ModelTier | null` to `WorkspaceModel` and `WorkspaceResource` as the workspace-level fallback. The effective tier resolution order will be: user-level → group-level → workspace default.

## Tests

No business logic in this PR — pure schema and model registration. Tested that migrations apply cleanly locally.

## Risk

Low. The SQL migration is additive only (new tables + nullable column with `DEFAULT NULL`). No existing queries or code paths are affected. `onDelete: RESTRICT` on both FKs prevents orphaned tiers. Tables are empty until write logic is wired in a follow-up PR.

## Deploy Plan

1. Run pre-deploy SQL migration `20260626174515_add_model_tier_tables_and_workspace_default.sql` against the front DB.
2. Deploy front.

### DIFF
--- a/front/lib/resources/storage/models/group_model_tier.ts
+++ b/front/lib/resources/storage/models/group_model_tier.ts
@@ -1,0 +1,76 @@
+import { MODEL_TIERS, type ModelTier } from "@app/lib/api/models_picker/tiers";
+import { frontSequelize } from "@app/lib/resources/storage";
+import { DataTypes } from "@app/lib/resources/storage/data_types";
+import { GroupModel } from "@app/lib/resources/storage/models/groups";
+import { WorkspaceAwareModel } from "@app/lib/resources/storage/wrappers/workspace_models";
+import type { CreationOptional, ForeignKey } from "sequelize";
+
+/**
+ * Per-(workspace, group) model tier override.
+ *
+ * When absent, members of the group inherit the workspace default tier unless
+ * they have a user-level override.
+ */
+export class GroupModelTierModel extends WorkspaceAwareModel<GroupModelTierModel> {
+  declare createdAt: CreationOptional<Date>;
+  declare updatedAt: CreationOptional<Date>;
+
+  declare groupId: ForeignKey<GroupModel["id"]>;
+  declare tier: ModelTier;
+}
+
+GroupModelTierModel.init(
+  {
+    createdAt: {
+      type: DataTypes.DATE,
+      allowNull: false,
+      defaultValue: DataTypes.NOW,
+    },
+    updatedAt: {
+      type: DataTypes.DATE,
+      allowNull: false,
+      defaultValue: DataTypes.NOW,
+    },
+    groupId: {
+      type: DataTypes.BIGINT,
+      allowNull: false,
+      references: {
+        model: GroupModel,
+        key: "id",
+      },
+    },
+    tier: {
+      type: DataTypes.STRING,
+      allowNull: false,
+      validate: {
+        isIn: [MODEL_TIERS],
+      },
+    },
+  },
+  {
+    modelName: "group_model_tiers",
+    sequelize: frontSequelize,
+    indexes: [
+      {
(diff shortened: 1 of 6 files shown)


Fraggle:

- #28098
    - #28099
        - #28100 ◀
            - #28101
                - #28125

flvndvd (approved):

front/lib/resources/storage/models/group_model_tier.ts
```diff
@@ -0,0 +1,76 @@
...
+ * Per-(workspace, group) model tier override.
+ *
+ * When absent, members of the group inherit the workspace default tier unless
+ * they have a user-level override.
+ */
+export class GroupModelTierModel extends WorkspaceAwareModel<GroupModelTierModel> {
+  declare createdAt: CreationOptional<Date>;
+  declare updatedAt: CreationOptional<Date>;
+
+  declare groupId: ForeignKey<GroupModel["id"]>;
```
Nils-Fedrigo, line 18: have you checked the painful scenario that you mentioned that groups could potentially de-sync and that their group id could change?

front/lib/resources/storage/models/group_model_tier.ts
```diff
@@ -0,0 +1,76 @@
...
+      allowNull: false,
+      validate: {
+        isIn: [MODEL_TIERS],
+      },
+    },
+  },
+  {
+    modelName: "group_model_tiers",
+    sequelize: frontSequelize,
+    indexes: [
```
Nils-Fedrigo, line 53: potentially add an index on workspaceId, needed when deleting or relocating?

Nils-Fedrigo: I don't know

Nils-Fedrigo (commented):
thanks!!

front/lib/resources/storage/models/group_model_tier.ts
Fraggle: I think it's a more general problem but thanks for remind me :)

front/lib/resources/storage/models/group_model_tier.ts
Fraggle: You got it via workspaceId, groupId.
I just checked, it's aligned with prior art.

spolu:
For the reader at home, MODEL_TIERS are defined here: https://github.com/dust-tt/dust/pull/28098/changes#diff-126d1a15381c363177e65b3ebcb64e84263c1348c20bc4aa781d9666d63b79fbR9

spolu:
Small question here on model tiers: are we sure about the expressibility? Where would we put GLM 5.2 as an example?

I might want to allow GLM 5.2 since OSS as "frontier" or "powerful" but not Opus 4.8? Will I be able to express that?

Fraggle:
> Small question here on model tiers: are we sure about the expressibility? Where would we put GLM 5.2 as an example?
> I might want to allow GLM 5.2 since OSS as "frontier" or "powerful" but not Opus 4.8? Will I be able to express that?

We are probably moving to express as cost instead of "power" since that's actually what it's based on and what is the reason why admins would want to restrict something.

And we are still iterating on the tiers.

Does that answer you question ?

spolu:
Does not completely answer the question as it creates new questions: looks like we have two dimensions in reality, cost AND capability right?

I understand that the tiers proposed here are not final. Looks like we may need two dimensions in the end?

Fraggle:
Never mind, in light of discussions today, the whole tiered approach is going away. Will cancel the stack.

spolu:
ack!

PR #28101, opened 2026-06-29
Add: resource to handle models tiers

Fraggle:
## Description

Introduces `ModelTierResource`, a static facade for reading and writing model tier configuration at three scopes: workspace, user, and group. Workspace tier is stored on `workspaces.defaultModelsTier` (null falls back to `DEFAULT_MODEL_TIER = "frontier"`). User and group tiers are stored as sparse override rows in `user_model_tiers` / `group_model_tiers`.

Also adds `isModelTier` type guard and `resolveWorkspaceDefaultModelTier` helper to `tiers.ts`, and threads an optional `transaction` parameter through `WorkspaceResource.updateWorkspaceSettings`.

## Tests

Added Vitest unit tests in `model_tier_resource.test.ts` covering set/get/clear for workspace, user, and group tiers, and an isolation test that verifies workspace defaults and user overrides don't bleed into each other.

## Risk

Low. Pure additive change — no existing call sites modified beyond making `updateWorkspaceSettings`'s `transaction` parameter optional. `ModelTierResource` is not wired to any route yet.

Dependency: `GroupModelTierModel` and `UserModelTierModel` (imported but not in this diff) must exist as Sequelize models with their backing DB tables before this code runs. Confirm the companion migration is in place.

## Deploy Plan

Deploy front. Ensure `user_model_tiers` and `group_model_tiers` tables are created (migration) before or alongside deployment.

### DIFF
--- a/front/lib/api/models_picker/tiers.ts
+++ b/front/lib/api/models_picker/tiers.ts
@@ -14,6 +14,18 @@ export const MODEL_TIERS = [
 ] as const;
 export type ModelTier = (typeof MODEL_TIERS)[number];
 
+export const DEFAULT_MODEL_TIER: ModelTier = "frontier";
+
+export function isModelTier(value: string): value is ModelTier {
+  return (MODEL_TIERS as readonly string[]).includes(value);
+}
+
+export function resolveWorkspaceDefaultModelTier(
+  defaultModelsTier: ModelTier | null
+): ModelTier {
+  return defaultModelsTier ?? DEFAULT_MODEL_TIER;
+}
+
 /**
  * Blended cost thresholds in USD per million tokens (average of input + output).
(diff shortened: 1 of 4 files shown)


Fraggle:

- #28098
    - #28099
        - #28100
            - #28101 ◀
                - #28125

front/lib/api/models_picker/tiers.ts
```diff
@@ -14,6 +14,18 @@ export const MODEL_TIERS = [
 ] as const;
 export type ModelTier = (typeof MODEL_TIERS)[number];
 
+export const DEFAULT_MODEL_TIER: ModelTier = "frontier";
```
Nils-Fedrigo, line 17: so by default everyone has access to frontier? Better to restrict it to Powerful no?

PR #28106, opened 2026-06-29
feat(llm): always route GA Gemini through agent platform in the new router

adrsimon:
## Description

Makes every streaming call for GA Gemini models go through agent platform (Vertex) in the new LLM router, and removes the preview model from agent platform entirely.

- Broaden the agent-platform `endpointFilter` for `gemini-3.5-flash` and `gemini-3.1-flash-lite` from `{ or: [use_vertex_for_supported_models, isCreditPriced] }` to `{}`, so all workspaces stream these GA models through Vertex rather than just flagged/credit-priced ones.
- Remove the now-redundant AI Studio (`google_ai_studio_global`) stream endpoints for those two GA models — Dust + `model_constructors` endpoint files, registry entries, and integration tests. Agent platform is now the only registered stream route for them.
- Remove the agent-platform (Vertex) stream endpoint for `gemini-3.1-pro-preview` — endpoint files, registry entries, and integration test. The preview model now streams exclusively through AI Studio, never Vertex.

- `gemini-3.5-flash` → agent platform only
- `gemini-3.1-flash-lite` → agent platform only
- `gemini-3.1-pro-preview` → AI Studio only

Batch is out of scope: there is no agent-platform/Vertex Gemini batch client, so Gemini batch continues to go through AI Studio.

## Tests

- Removed the integration tests for the deleted endpoints; the remaining agent-platform GA endpoint integration tests are unchanged.

## Risk

- Low. The new router is gated behind `use_new_llm_router`; without it, the legacy router is unchanged. When enabled, GA Gemini resolves to the agent-platform endpoint for every workspace and the preview model resolves to AI Studio.
- The GA filter is unconditional (`{}`), unlike Claude's agent-platform endpoints which stay gated, and the new router has no region/credential guard in endpoint selection. This assumes `AGENT_PLATFORM_PROJECT_ID` is configured wherever `use_new_llm_router` is enabled (legacy gated Vertex on `europe-west1`). Rollback is a straightforward revert.

## Deploy Plan

- Standard deploy. Ensure `AGENT_PLATFORM_PROJECT_ID` is set in any environment where `use_new_llm_router` is on before/at rollout.


flvndvd (approved):
LGTM

PR #28113, opened 2026-06-29
Add slug and description to sandbox functions

flvndvd:
## Description

Sandbox functions had no identity of their own (outside of the opaque sId). There was no human-readable description, which gave a model little to choose on.

This gives each sandbox function a model-provided slug and a required description. The slug is a lowercase, hyphenated identifier, unique within a pod, and serves as the stable handle the model reasons about. It's the canonical way to reference a function. The description is a short summary that helps the model pick the right function. The list output now reads as slug and description only, dropping both the bundle filename and the internal identifier, neither of which the model needs.

A pre-deploy migration adds the two columns and a per-pod uniqueness constraint on the slug. Both columns are required, which is safe because the feature is still behind a flag with no creation path live yet, so the table is empty.

## Deploy Plan

- [x] Apply migration
- [ ] Deploy front

### DIFF
--- a/front-api/routes/w/[wId]/sandbox-functions/[functionId]/invocations.test.ts
+++ b/front-api/routes/w/[wId]/sandbox-functions/[functionId]/invocations.test.ts
@@ -48,6 +48,8 @@ async function setupSandboxFunction({
   const sandboxFunction = await SandboxFunctionResource.makeNew(adminAuth, {
     space,
     file,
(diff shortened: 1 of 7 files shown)


davidebbo (approved):
LGTM

flvndvd:
Migration applied in both regions.

spolu (commented):
LGTM \o/

PR #28114, opened 2026-06-29
[front] tool credit cost: centralize cost computation logic

thervier:
## Description

In preparation of evolution of credit cost per tool => centralize all the logic for credit cost computation at the same place

Move the free-origin check (isFreeOrigin) and the per-tool AWU weight lookup (toolAwuFromServerName) down into the low-level functions in metronome/events.ts, so all callers (intelligenceAwuFromRunUsages, toolAwuFromActions, etc.) apply the same billing rules automatically.

This removes the need for callers to know about isFreeUsage or contextOrigin handling — they just pass the origin and the right thing happens.

computeAgentMessageCredits now takes a contextOrigin instead of a boolean isFreeUsage, aligning it with the Metronome event functions it delegates to.

## Tests

Tested locally

## Risk

Medium, code refactoring but still impacts the cost computation which is key for us

## Deploy Plan

Deploy front

### DIFF
--- a/front/lib/metronome/events.ts
+++ b/front/lib/metronome/events.ts
@@ -152,9 +152,7 @@ const TOOL_CATEGORY_MAP: Record<InternalMCPServerNameType, ToolCategory> = {
   clari_copilot: "advanced",
 };
 
-export function getToolCategory(
-  internalMCPServerName: string | null
-): ToolCategory {
+function getToolCategory(internalMCPServerName: string | null): ToolCategory {
   if (
     !internalMCPServerName ||
     !isInternalMCPServerName(internalMCPServerName)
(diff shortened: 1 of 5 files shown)


front/lib/metronome/events.ts
```diff
@@ -71,16 +71,18 @@ export function isToolCategory(value: string): value is ToolCategory {
 // Exhaustive map — TypeScript will error if a new internal MCP server is added
 // without being categorized here.
 const TOOL_CATEGORY_MAP: Record<InternalMCPServerNameType, ToolCategory> = {
+  // Set as basic but overriden as free
```
rfrenoy:
```suggestion
  // Set as basic but overridden as free
```

rfrenoy (approved):

PR #28117, opened 2026-06-29
Add pod sandbox readiness lifecycle

fontanierh:
## Description

Follow up of [#28048](https://github.com/dust-tt/dust/pull/28048). This adds the pod sandbox readiness path without wiring product call sites yet.

- Adds `ensurePodSandboxReady`, backed by `PodSandboxAdapter` and `DustFileSystem.forPod`.
- Threads sandbox owner context into telemetry, Fluent Bit enrichment, and the egress env manifest so pods expose `SPACE_ID`, not `CONVERSATION_ID`.
- Keeps conversation-created pod mounts backward compatible, while pod-only filesystem setup does not request the legacy `/files/pod` sandbox alias.
- Updates existing conversation sandbox tool callers to use `ensureConversationSandboxReady` explicitly.

Next PR: wire pod sandbox callers to this lifecycle helper.

## Tests

- `NODE_ENV=test npm test -- lib/api/sandbox/lifecycle.test.ts lib/api/sandbox/egress.test.ts lib/api/sandbox/env_manifest.test.ts lib/api/file_system/dust_file_system.test.ts lib/api/actions/servers/sandbox/tools/index.test.ts lib/api/sandbox/image/registry.test.ts`
- `NODE_OPTIONS="--max-old-space-size=8192" npx tsgo --noEmit`

## Risk

Low. There are no pod sandbox call sites yet; existing conversation callers still use the same readiness path through `ensureConversationSandboxReady`. Worst case, pod-only readiness is unused until the next PR. Safe to rollback.

## Deploy Plan

- [ ] Standard deploy.
- [ ] After deploy, open `/poke/kill` (Kill Switches) and request a kill of older `dust-base` sandbox image versions so live sandboxes pick up `0.8.38`.

PR #28118, opened 2026-06-29
Support sandbox function invocation tokens

fontanierh:
## Description

Follow up of #28117 and prep for sandbox function invocation.

Adds the token/auth plumbing needed by pod sandbox function invocations:
- Sandbox tokens now support function invocation claims under the existing `sbt-` prefix, with a required invocation ID claim and optional conversation claim.
- Sandbox-token auth scopes function invocation tokens to the pod space, plus conversation spaces when a conversation claim is present.
- Existing sandbox action callback middleware continues to accept only complete conversation action token claims.

No endpoint behavior changes in this PR. The actual invoke path is stacked separately in #28123.

## Tests

Tested locally:
- `npx biome check` on touched files
- `front`: `lib/api/sandbox/access_tokens.test.ts`
- `front-api`: `routes/v1/w/[wId]/sandbox/actions/index.test.ts`

## Risk

Existing conversation action tokens keep the same payload shape. Function invocation tokens are duck-typed from complete invocation claims instead of a discriminator, and the action callback middleware rejects tokens that do not include complete conversation action claims. Safe to rollback.

## Deploy Plan

Standard deploy. No migration or manual step.

### DIFF
--- a/front/lib/api/sandbox/access_tokens.ts
+++ b/front/lib/api/sandbox/access_tokens.ts
@@ -16,22 +16,91 @@ import { z } from "zod";
 
 export const SANDBOX_TOKEN_PREFIX = "sbt-";
 
-const SandboxExecTokenPayloadSchema = z.object({
-  wId: z.string(),
-  cId: z.string(),
-  // Optional: omitted when the conversation is driven by a non-human actor
-  // (e.g. a Slack bot user with no associated Dust user).
-  uId: z.string().optional(),
-  aId: z.string(),
-  mId: z.string(),
-  sbId: z.string(),
-  execId: z.string(),
-  actionId: z.string(),
-});
+function isDefined<T>(value: T | undefined): value is T {
+  return value !== undefined;
+}
 
-export type SandboxExecTokenPayload = z.infer<
-  typeof SandboxExecTokenPayloadSchema
->;
+const SandboxTokenPayloadSchema = z
+  .object({
+    wId: z.string(),
+    uId: z.string().optional(),
+    sbId: z.string(),
+    execId: z.string(),
+    cId: z.string().optional(),
+    aId: z.string().optional(),
+    mId: z.string().optional(),
+    actionId: z.string().optional(),
+    spaceId: z.string().optional(),
+    sandboxFunctionId: z.string().optional(),
+    invocationId: z.string().optional(),
+  })
+  .superRefine((payload, ctx) => {
+    const actionClaims = [payload.aId, payload.mId, payload.actionId];
+    const invocationClaims = [
+      payload.spaceId,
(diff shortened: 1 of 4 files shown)


front/lib/auth.ts
```diff
@@ -632,6 +661,72 @@ export class Authenticator {
...
+        api_error: {
+          type: "invalid_sandbox_token_error",
+          message: "The sandbox token pod space is invalid.",
+        },
+      });
+    }
+
+    const allowedSpaceIds = new Set<ModelId>([podSpaceModelId]);
+    if (claims.cId) {
+      const conversation = await ConversationModel.findOne({
```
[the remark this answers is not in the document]
fontanierh: Restored this to the pre-existing auth-layer lookup and removed the ad-hoc ConversationAuthResource. We are intentionally keeping this direct model lookup in Authenticator for now because fromSandboxToken is constructing the Authenticator itself, and importing ConversationResource here currently creates an auth <-> conversation_resource runtime cycle (ConversationResource imports auth helpers such as hasFeatureFlag). I added an inline comment to make that constraint explicit.

front/lib/api/sandbox/access_tokens.ts
```diff
@@ -16,22 +16,91 @@ import { z } from "zod";
...
-});
-
-export type SandboxExecTokenPayload = z.infer<
-  typeof SandboxExecTokenPayloadSchema
->;
+function isDefined<T>(value: T | undefined): value is T {
+  return value !== undefined;
+}
+
+const SandboxTokenPayloadSchema = z
```
flvndvd, line 23: Should we do a discriminated union type here? So we can remove the conditional check later in the code?

flvndvd: Maybe just adding a discriminated type?

front/lib/api/sandbox/access_tokens.ts
```diff
@@ -137,12 +206,53 @@ export async function generateSandboxExecToken(
...
 
+export async function generateSandboxFunctionInvocationToken(
+  auth: Authenticator,
+  {
+    conversationId,
+    sandbox,
+    sandboxFunction,
+    invocationId,
+    execId,
+    expiryMs = 2 * 60 * 1000, // Default to 2 minutes
```
flvndvd, line 217: No magic variable 😛

front/lib/api/sandbox/access_tokens.ts
```diff
@@ -137,12 +206,53 @@ export async function generateSandboxExecToken(
...
+  }: {
+    conversationId?: string;
```
flvndvd, line 219: Why do we need a conversation Id here?

PR #28120, opened 2026-06-29
fix(llm): route non-BYOK Gemini through Vertex via api-level filter

adrsimon:
## Description

Non-BYOK workspaces couldn't reach the agent-platform (Vertex) Gemini endpoints in the new router. `getProviderIdFilter` stripped `google_ai_studio` entirely, which also excluded the Vertex endpoints (they share that provider id), so non-BYOK fell back to legacy.

- Keep `google_ai_studio` in the provider filter for non-BYOK.
- Exclude only the direct AI-Studio API (`api = google-ai-studio`) for non-BYOK, so Vertex stays reachable but Dust's managed AI-Studio key isn't used.
- Rename `EndpointConfig.providerApi` → `api` to match the endpoint's static field (api-level filtering was a no-op before).

### Legacy equivalent

In the legacy router this is the `useVertex: !plan.isByok` line in the Google branch of `getLegacyLLM` (`front/lib/api/llm/index.ts`): non-BYOK → Vertex (GCP project creds, not the direct AI-Studio key), BYOK → direct AI Studio. The same model maps to two new-router endpoints sharing `providerId: "google_ai_studio"` — one `api: "vertex"`, one `api: "google-ai-studio"`. Legacy picks between them with the `useVertex` flag; the new router picks via endpoint filtering. This change makes the `api`-level exclusion the direct analog of `!plan.isByok`.

## Tests

`router_parity.test.ts` unaffected. `getWorkspaceFilter` isn't unit-tested yet — verified manually against the endpoint configs.

## Risk

Low. The change widens routing for non-BYOK to the Vertex Gemini endpoints; the direct AI-Studio key stays excluded. Rollback is reverting the filter change — non-BYOK falls back to legacy as before.

## Deploy Plan

Standard deploy, no ordering constraints.

### DIFF
--- a/front/lib/api/llm/index.ts
+++ b/front/lib/api/llm/index.ts
@@ -34,12 +34,14 @@ import { getModelConfigByModelId } from "@app/lib/llms/model_configurations";
 import { getStreamEndpoints } from "@app/lib/llms/stream";
 import type {
   EndpointConfig,
+  LogicalFilters,
   ValueFilter,
   Where,
   WorkspaceConfig,
 } from "@app/lib/llms/types/filter";
 import { sortEndpointsByPreferredRegion } from "@app/lib/llms/utils/sort_endpoints";
 import { isModelId } from "@app/lib/model_constructors/types/model_ids";
+import { GOOGLE_AI_STUDIO_API } from "@app/lib/model_constructors/types/provider_apis";
 import {
   isProviderId,
   type ProviderId,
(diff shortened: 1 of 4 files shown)


front/lib/api/llm/index.ts
```diff
@@ -305,19 +307,29 @@ function getProviderIdFilter(auth: Authenticator): ValueFilter<ProviderId> {
   const byok = auth.getNonNullablePlan().isByok;
   const providerIds = byok
     ? intersection(whitelistedProviderIds, BYOK_MODEL_PROVIDER_IDS)
-    : whitelistedProviderIds.filter(
-        // We route all non-byok gemini requests to agent platform
```
flvndvd, line 309: IF workspace uses their own key, they should still hit google_ai_studio. If we want to align on agent_platform, IMHO this requires comms.

PR #28122, opened 2026-06-29
Add get tool for sandbox function schemas and slim down list

flvndvd:
## Description

The list tool returned each function's full input and output JSON schemas inline, which is verbose and scales poorly as the number and size of functions grow. This splits discovery from detail: list now returns only each function's slug and description, and a new get tool returns the full schemas for a single function looked up by its slug. This keeps listing cheap for the model and follows the usual list-and-get pattern, with list pointing the model at get when it needs a function's interface.

The get tool resolves the function within the current pod and fails clearly when no function matches the slug. Access is scoped the same way as listing, so a function in an inaccessible space is not returned.


davidebbo (approved):
LGTM

davidebbo:
Should it be added to the Computer skill?

flvndvd:
> Should it be added to the Computer skill?

Will add it to a dedicated global skill.

spolu:
Exciting!

PR #28123, opened 2026-06-29
Invoke sandbox functions on pod sandboxes

fontanierh:
## Description

Follow up of #28118.

Adds the first sandbox function invocation path for pod sandboxes:
- `SandboxFunctionResource.invoke` creates an ephemeral invocation id, ensures the pod sandbox is ready, writes a valid temporary `dsbx function run` stub in a per-invocation staging directory, mints a short-lived `sbt-` token with invocation claims, and runs it in the pod sandbox.
- The invocation endpoint delegates to the resource and keeps returning `status: "created"`.
- No invocation row is persisted yet. The token remains valid for its short TTL so a future durable callback can post results and revoke it after acceptance.

This intentionally does not read from the sandbox function file or a pod bundle mount yet. We will wire the mounted bundle path once that mount exists.

## Tests

Tested locally:
- `npx biome check` on touched files
- `front`: `lib/resources/sandbox_function_resource.test.ts`
- `front-api`: `routes/w/[wId]/sandbox-functions/[functionId]/invocations.test.ts`
- `npm -w front-api run docs` (no generated OpenAPI diff)

## Risk

Worst case, sandbox function invocation returns a 500 before any persisted state is created. Safe to rollback.

## Deploy Plan

Standard deploy. No migration or manual step.

### DIFF
--- a/front-api/routes/w/[wId]/sandbox-functions/[functionId]/invocations.ts
+++ b/front-api/routes/w/[wId]/sandbox-functions/[functionId]/invocations.ts
@@ -1,4 +1,3 @@
-import { randomUUID } from "node:crypto";
 import { SandboxFunctionResource } from "@app/lib/resources/sandbox_function_resource";
 import type {
   PostSandboxFunctionInvocationRequestBody,
@@ -54,17 +53,24 @@ app.post(
       });
     }
 
-    // TODO(spolu): implement sandbox function invocation
-    void body;
+    const invocationResult = await sandboxFunction.invoke(auth, body);
+    if (invocationResult.isErr()) {
+      return apiError(
+        ctx,
+        {
+          status_code: 500,
+          api_error: {
+            type: "internal_server_error",
+            message: "Sandbox function invocation failed.",
+          },
+        },
+        invocationResult.error
+      );
+    }
 
     return ctx.json(
       {
-        invocation: {
-          id: randomUUID(),
-          functionId: sandboxFunction.sId,
-          status: "created",
-          createdAt: new Date().toISOString(),
-        },
+        invocation: invocationResult.value,
       },
       201
     );
--- a/front/lib/resources/sandbox_function_resource.ts
+++ b/front/lib/resources/sandbox_function_resource.ts
@@ -1,3 +1,12 @@
+import { randomUUID } from "node:crypto";
+import path from "node:path";
+import config from "@app/lib/api/config";
+import {
+  generateExecId,
+  generateSandboxFunctionInvocationToken,
+} from "@app/lib/api/sandbox/access_tokens";
+import { ensurePodSandboxReady } from "@app/lib/api/sandbox/lifecycle";
+import { shellEscape } from "@app/lib/api/sandbox/shell";
 import type { Authenticator } from "@app/lib/auth";
 import { BaseResource } from "@app/lib/resources/base_resource";
 import { FileResource } from "@app/lib/resources/file_resource";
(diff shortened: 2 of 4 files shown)


front/lib/resources/sandbox_function_resource.ts
```diff
@@ -242,6 +301,125 @@ export class SandboxFunctionResource extends BaseResource<SandboxFunctionModel>
...
+      const execResult = await ensureResult.value.sandbox.exec(auth, command, {
+        workingDirectory: SANDBOX_FUNCTION_WORKING_DIRECTORY,
+        envVars: {
+          DUST_API_URL: `${sandboxAPIBaseUrl()}/api/v1/w/${auth.getNonNullableWorkspace().sId}`,
+          DUST_FUNCTIONS_DIR: stagedFunctionsDir,
+          DUST_SANDBOX_TOKEN: token,
+        },
+        stdin: JSON.stringify(inputEnvelope),
+        timeoutMs: SANDBOX_FUNCTION_EXEC_TIMEOUT_MS,
+        user: "agent-proxied",
```
davidebbo, line 352: Not sure if the logic that saves the bundled harness to disk will work under that user. But we'll see.

front/lib/resources/sandbox_function_resource.ts
```diff
@@ -14,14 +23,64 @@ import {
...
 import assert from "assert";
 import type { JSONSchema7 as JSONSchema } from "json-schema";
 import type { Attributes, Transaction } from "sequelize";
 
+const SANDBOX_FUNCTION_WORKING_DIRECTORY = "/home/agent";
+const SANDBOX_FUNCTION_EXEC_TIMEOUT_MS = 2 * 60 * 1000;
+const SANDBOX_FUNCTION_STAGING_ROOT = "/tmp/dust-sandbox-functions";
+const DSBX_BIN_PATH = "/opt/bin/dsbx";
+
+function sandboxAPIBaseUrl(): string {
```
davidebbo: Slightly confusing name, as it's really the Dust API Url, to be used *from* the sandbox.

front/lib/resources/sandbox_function_resource.ts
```diff
@@ -14,14 +23,64 @@ import {
...
+}: {
+  sourcePath: string;
+  stagedFunctionsDir: string;
+  slug: string;
+}): string {
+  return [
+    "set -euo pipefail",
+    `rm -rf -- ${shellEscape(stagedFunctionsDir)}`,
+    `mkdir -p -- ${shellEscape(stagedFunctionsDir)}`,
+    `cp -- ${shellEscape(sourcePath)} ${shellEscape(
```
davidebbo: Is that because we don't want to run directly off the bundle FUSE dir?

davidebbo: Maybe could use a comment here

front/lib/resources/sandbox_function_resource.ts
```diff
@@ -242,6 +301,125 @@ export class SandboxFunctionResource extends BaseResource<SandboxFunctionModel>
...
+        invocationId
+      );
+      const command = buildSandboxFunctionCommand({
+        sourcePath: mountedFilePath.value,
+        stagedFunctionsDir,
+        slug: this.slug,
+      });
+      const inputEnvelope = {
+        method: "POST",
+        url: `https://dust.local/sandbox-functions/${this.sId}/invocations/${invocationId}`,
```
davidebbo, line 328: So this is the URL that the running function will get in its request, right? I think in most cases it will ignore it, as we don't expect it to do routing.

Is there any significance to dust.local, or is it just for the sake of having a host name?

davidebbo (approved):
LGTM

front/lib/resources/sandbox_function_resource.ts
fontanierh: Yes probably not but since I didn't test yet, I don't really know what is needed.

Once we know precisely I'll create a specific user `function-runner` that can do all that's needed (including switch to agent-proxied). That way we can fully avoid root

front/lib/resources/sandbox_function_resource.ts
fontanierh: Fair point

front/lib/resources/sandbox_function_resource.ts
fontanierh: It's more that I was trying to align with what dsbx expects currently.

But maybe we should align dsbx with whatever the mount will be. I can just remove all thsi I think

front/lib/resources/sandbox_function_resource.ts
fontanierh: No just inventing something plausible but that doesn't actually resolve to anything

front/lib/resources/sandbox_function_resource.ts
fontanierh: Removed the mounted-bundle copy path for now. Since bundles are not mounted yet, the invoke path now writes a tiny per-invocation stub function and runs that through `dsbx function run`; we will wire the mounted bundle path once the mount exists.

fontanierh:
Checked the Danger docs warning: this route is annotated with `@ignoreswagger`, and `npm -w front-api run docs` completes cleanly with no generated `front-api/public/swagger.json` diff. No OpenAPI deploy needed for this PR.

front/lib/resources/sandbox_function_resource.ts
fontanierh: Renamed the helper to `dustAPIBaseUrlForSandbox()` so the code now says this is the Dust API base URL that sandbox code should use, not a sandbox API URL.

front/lib/resources/sandbox_function_resource.ts
```diff
@@ -14,14 +23,54 @@ import {
...
+
+  return [
+    "set -euo pipefail",
+    `rm -rf -- ${shellEscape(stagedFunctionsDir)}`,
+    `mkdir -p -- ${shellEscape(stagedFunctionsDir)}`,
+    `cat > ${shellEscape(functionPath)} <<'DUST_SANDBOX_FUNCTION_EOF'`,
+    "export default {",
+    "  async fetch() {",
+    "    return Response.json({ ok: true });",
+    "  },",
```
davidebbo, line 67: Maybe no comma after brace?

fontanierh: why not ?

fontanierh: Its valid syntax

front-api/routes/w/[wId]/sandbox-functions/[functionId]/invocations.ts
```diff
@@ -54,17 +53,24 @@ app.post(
       });
     }
 
-    // TODO(spolu): implement sandbox function invocation
-    void body;
+    const invocationResult = await sandboxFunction.invoke(auth, body);
```
Comment (spolu) [inline front-api/routes/w/[wId]/sandbox-functions/[functionId]/invocations.ts:56]: 🔥

spolu (commented):
Wonderful

front/lib/resources/sandbox_function_resource.ts
davidebbo: Yeah, it's fine.

front/lib/resources/sandbox_function_resource.ts
davidebbo: From initial tests, it seems to work.

fontanierh: yeah but we're not accessing any protected folder just yet. Once we have the bundles we will

PR #28124, opened 2026-06-29
feat(front): redesign notification settings page

Inesvvv:
## Description
Rebuild the notifications panel to match the new design using Sparkle:
- Split into "Inbox notifications" and "Other channels" sections via Page.SectionHeader
- Move channel preferences onto SettingsList rows with SliderToggle and dropdowns, replacing the inline sentence-style controls
- Spinner while data loads 
-  on load failure, show an inline warning message instead of the rows
- Notification preferences are now fetched via a dedicated SWR hook `useConversationNotificationPreferences` in the parent instead of an imperative effect in the child, and passed down as props.
- `NotificationPreferences` is now a controlled `react-hook-form` form (`useNotificationPreferencesForm` + zod schema). Dropped the `forwardRef` / `useImperativeHandle` / `onChanged`-in-`useEffect` plumbing, the parent reads `isDirty` / `save` off the hook directly. Saves only write `dirtyFields`.

Screenshot 2026-06-29 at 14 29 43

## Tests
Tested locally 
Add sound_notification feature flag to be able to view the full notification page.

## Risk

Low. Frontend-only, scoped to the notifications settings panel and the underlying data wiring is unchanged.

## Deploy Plan

`front` deploy

### DIFF
--- a/front/components/UserSettingsPopover.tsx
+++ b/front/components/UserSettingsPopover.tsx
@@ -1,6 +1,8 @@
 import { UsageUpgradeButton } from "@app/components/credits/UsageUpgradeButton";
-import type { NotificationPreferencesRefProps } from "@app/components/me/NotificationPreferences";
-import { NotificationPreferences } from "@app/components/me/NotificationPreferences";
+import {
+  NotificationPreferences,
+  useNotificationPreferencesForm,
+} from "@app/components/me/NotificationPreferences";
 import { PendingInvitationsTable } from "@app/components/me/PendingInvitationsTable";
 import {
   SoundNotificationPreferences,
@@ -32,6 +34,7 @@ import {
   BarChart01,
   Bell01,
   Button,
+  ContentMessageInline,
   cn,
   Dialog,
   DialogClose,
@@ -43,12 +46,14 @@ import {
   DropdownMenuShortcut,
   DropdownMenuTrigger,
   Edit04,
+  InfoCircle,
   Input,
   Label,
   Mail01,
   Moon01,
   NavigationList,
   NavigationListItem,
+  Page,
   Separator,
   Settings01,
   ShapesPlus,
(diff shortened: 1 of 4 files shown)


front/components/me/NotificationPreferences.tsx
```diff
@@ -365,47 +349,29 @@ export const NotificationPreferences = forwardRef<
...
-    }
-    return displayedChannels.length > 0 ? displayedChannels.join(", ") : "none";
-  };
+  const isPreferencesReady = conversationPreferences !== undefined;
 
-  if (isLoadingPreferences || isSlackSetupLoading) {
-    return <Spinner />;
-  }
-
-  if (!conversationPreferences) {
```
PopDaph, line 389: Now when Novu fails (session error / request error / missing workflow — all three of which you still datadogLogger.error), conversationPreferences stays undefined, so the rows render permanently disabled with no explanation to the user, and savePreferences() silently return falses. Dropping the blocking spinner is the right call, but losing the error message is a regression — the user just sees dead toggles. Worth keeping a lightweight inline message or toast on the persisted-error case (e.g. when not loading and conversationPreferences === undefined). wdyt?

front/components/UserSettingsPopover.tsx
```diff
@@ -612,40 +612,49 @@ function NotificationsSection({ owner }: { owner: WorkspaceType }) {
     />
   );
 
+  const inboxSection = hasFeature("sound_notification") && (
```
PopDaph: Let's avoid putting jsx into a var, we prefer do a shared component if needed 🙏🏻

front/components/me/NotificationPreferences.tsx
```diff
@@ -365,47 +349,29 @@ export const NotificationPreferences = forwardRef<
...
-      displayedChannels.push("Slack");
-    }
-    if (preference.channels.email) {
-      displayedChannels.push("email");
-    }
-    return displayedChannels.length > 0 ? displayedChannels.join(", ") : "none";
-  };
+  const isPreferencesReady = conversationPreferences !== undefined;
 
-  if (isLoadingPreferences || isSlackSetupLoading) {
```
PopDaph, line 385: Why removing the spinner when data is loading? 🙏🏻

Inesvvv: If we keep the spinner I'll also add it for the sound notification settings so that everything loads at once

front/components/me/NotificationPreferences.tsx
Inesvvv: I was going to add a toast but in a separate pr. I'll add to this pr 👍

Inesvvv:
@PopDaph I implemented inline message for Novu fail and removed the JSX from var. I left the no spinner as is, lmk if you want to reimplement it

front/lib/swr/notifications.ts
```diff
@@ -0,0 +1,123 @@
...
+  } else if (data) {
+    status = "ready";
+  } else if (error) {
+    status = "error";
+  }
+
+  const saveConversationPreferences = useCallback(
+    async (preference: Preference) => {
+      if (!novuClient) {
+        throw new Error("Novu client is not ready.");
```
PopDaph, line 104: Does the throw shows the raw error to the user?

PopDaph: (here and other occurrences in this file 🙏🏻 )

Inesvvv:
Screenshot 2026-07-01 at 10 26 10
Screenshot 2026-07-01 at 10 36 18

PopDaph (approved):
I did not review thoroughly but I know you tested it well. LGTM 🙏🏻

Inesvvv:
I asked flavify to review it :) All good

PR #28065, opened 2026-06-26
[front] API key credit cap (10/10): tests

tdraier:
Part of the **per-API-key credit spend limit** stack. Builds on #28064. Final PR.

### This PR — tests (11 passing)
- **`api_key_caps.test.ts`** — `upsertMetronomeApiKeyCapAlert` builds a `spend_threshold_reached` alert scoped to `group_values: [{ key: "api_key_name", value }]` with the right threshold + uniqueness key; `clearMetronomeApiKeyCapAlert` archives by uniqueness key; both propagate Metronome errors.
- **`api_key_credit_state_machine.test.ts`** — `expectedApiKeyCreditStateFromUsage` (no cap → `on_pool`; spend < cap → `on_pool`; spend ≥ cap → `capped`).
- **`api_key_credit_state_machine.transition.test.ts`** (DB-backed, `KeyFactory`) — `on_pool → capped` on `api_key_cap_reached` (persisted to the row), `capped → on_pool` on `api_key_cap_resolved` and `admin_cap_cleared`.

Mocks only the Metronome alert layer / Redis cache writer; uses the real DB per the test guidelines.

### Stack complete
1 schema · 2 alerts · 3 state · 4 lib/reconcile · 5 webhook · 6 enforcement · 7 API · 8 UI · 9 backfill · 10 tests.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #28064, opened 2026-06-26
[front] API key credit cap (9/10): backfill poke plugin

tdraier:
Part of the **per-API-key credit spend limit** stack. Builds on #28063.

### This PR — backfill
- **`backfillApiKeyCreditCapsForWorkspace`** — for a credit-priced workspace: converts any legacy USD per-key cap to a credit cap (microUSD / 10,000, clamped to ≥ 1), (re)creates the Metronome per-key cap alerts, and reconciles each key's credit state from live usage. Idempotent.
- **"Sync API Key Credit Cap Alerts"** poke plugin (billing role, credit-priced workspaces) wrapping it — run on credit-pricing adoption or to repair missing alerts.

This closes the migration gap: workspaces already on credit pricing with keys that had a legacy cap get their alerts/state created without waiting for a manual edit.

### Next in stack
Tests.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #28063, opened 2026-06-26
[front] API key credit cap (8/10): UI for per-key credit cap

tdraier:
Part of the **per-API-key credit spend limit** stack. Builds on #28061.

### This PR — UI
Surfaces the per-key credit cap on credit-priced plans (gated on `isCreditPricedPlan`, the inverse of the legacy USD-cap gate so the two never show together):

- **APIKeysList** — shows the monthly credit cap and a **capped** badge derived from `key.creditState`; the *Edit cap* button now appears on credit plans too.
- **NewAPIKeyDialog** — a credit-denominated cap input on credit plans, passed through create as `monthly_cap_awu_credits`.
- **EditKeyCreditCapDialog** — new credits-denominated edit sheet (mirrors the USD one).
- **APIKeysPage** — create/update handlers send `monthly_cap_awu_credits`.
- **utils** — credit-cap zod schema + parse/format helpers.

Follows the file's existing inline `clientFetch` pattern for consistency.

### Next in stack
Backfill plugin → tests.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #28061, opened 2026-06-26
[front] API key credit cap (7/10): API contract (create + update)

tdraier:
Part of the **per-API-key credit spend limit** stack. Builds on #28059.

### This PR — API contract
Both the create (`POST /api/w/:wId/keys`) and update (`PATCH /api/w/:wId/keys/:id`) routes now accept **`monthly_cap_awu_credits`** (AWU credits; `null` = unlimited) alongside the legacy `monthly_cap_micro_usd`. Both fields are optional → backward compatible ([BACK12]).

When the credit field is present, the route delegates to `setApiKeySpendLimit` (credit-priced gate + Metronome alert sync + reconcile) and maps the domain `Result` error to HTTP. The legacy USD path is untouched for non-credit plans. On create, the cap is validated up front (credit-priced + range) so a key is never created with an unappliable cap.

Adds `previous_spending_cap_awu_credits` / `new_spending_cap_awu_credits` to the `api_key.updated` audit schema.

Routes are `@ignoreswagger` (internal), so no Swagger schema changes.

### ⚠️ Post-merge action ([AUDIT10])
Re-register the audit schema before deploy:
`npx tsx front/admin/register_audit_log_schemas.ts --execute --changed`

### Next in stack
UI + SWR → backfill plugin → tests.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #28127, opened 2026-06-29
[front] API key credit cap: poke per-key reconcile target

tdraier:
Adds an **`api_key`** target to the existing **"Check & Reconcile Credit State"** poke plugin, so an admin can reconcile a single API key's credit state from live Metronome usage — the per-key analogue of the existing `user` target.

- New `keyName` arg (shown when target = `api_key`).
- Reconciles **every active key with that name** (names aren't unique; each is reconciled per its own cap via `reconcileApiKey`, matching the webhook/backfill behavior), reporting before/after state per key.
- `check` reports drift without writing; `execute` writes through the same authoritative reconcile setter.

Stacked on top of #28065 (the cap stack). Pure addition — no schema/contract change.

### DIFF
--- a/front/lib/api/metronome/reconcile_credit_state.ts
+++ b/front/lib/api/metronome/reconcile_credit_state.ts
@@ -66,6 +66,7 @@ export const RECONCILE_CREDIT_STATE_TARGETS = [
   "pool",
   "programmatic",
   "user",
+  "api_key",
 ] as const;
 
 export type ReconcileCreditStateTarget =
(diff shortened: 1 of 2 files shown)


tdraier:

- #28050
    - #28051
        - #28054
            - #28057
                - #28058
                    - #28059
                        - #28061
                            - #28063
                                - #28064
                                    - #28065
                                        - #28127 ◀
                                            - #28176

fabiencelier (approved):

PR #28128, opened 2026-06-29
Do not hide the steps when conversation errors

PopDaph:
## Description

When an agent message errored, we dropped all inline activity steps — you could no longer see what the agent did before it failed.

**Root cause:** `failed` was the only terminal status that hit an early `return <ErrorMessage>` before the main render, so `<InlineActivitySteps>` never mounted. The step data was never lost (loaded on reload, flushed on live error) — purely a render gap.

**What this PR does:** drops the `failed` early-return and renders `<ErrorMessage>` at the bottom of the normal render, like `cancelled` / `interrupted` already do.

> [!NOTE]
> Side-effects of the fall-through, both consistent with `cancelled`: partial content the agent wrote before erroring now renders, and a tool-less error shows a short "Errored after Xs" line above the error card.

## Tests

Locally. 

## Risk

Can be rolled back. 

## Deploy Plan

Deploy front.

### DIFF
--- a/front/components/assistant/conversation/AgentMessage.tsx
+++ b/front/components/assistant/conversation/AgentMessage.tsx
@@ -1309,23 +1309,6 @@ function AgentMessageContent({
     );
   }
 
-  if (agentMessage.status === "failed") {
-    return (
-      <ErrorMessage
-        error={
-          agentMessage.error ?? {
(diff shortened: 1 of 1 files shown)


ykmsd (approved):
nice 🫶

PR #28129, opened 2026-06-29
Persist and replay provider tool-search blocks so Anthropic thinking signatures stay valid

flvndvd:
## Description

With tool search enabled, Anthropic runs the search inside a single assistant turn and emits its own server-side blocks interleaved with the model's thinking. We were dropping those blocks when reconstructing the turn for the next request, which invalidated the thinking-block signatures and caused intermittent 400s whenever the model searched, reasoned, and then called a discovered tool.

These blocks are now captured and stored in stream order alongside the rest of the turn, and replayed verbatim to Anthropic so the reconstructed turn matches what was originally returned and the signatures validate.

The blocks are stored as an opaque, provider-tagged passthrough so the provider-agnostic layer never learns anything Anthropic-specific. Every other model provider, token accounting, and the message rendering simply skip a passthrough that isn't theirs, which also keeps cross-provider conversations safe. All knowledge of what the blocks actually are stays inside the Anthropic client, which parses them back with full typing and runtime validation.

## Deploy Plan

No migration required just some sequelize validation logic.

### DIFF
--- a/front/lib/api/llm/clients/anthropic/utils/anthropic_to_events.ts
+++ b/front/lib/api/llm/clients/anthropic/utils/anthropic_to_events.ts
@@ -232,7 +232,8 @@ function* handleContentBlockStart(
         accumulatorType: blockType === "text" ? "text" : "reasoning",
       };
       return;
-    case "tool_use": {
+
+    case "tool_use":
       stateContainer.state = {
         currentBlockIndex: event.index,
         accumulator: "",
@@ -252,7 +253,7 @@ function* handleContentBlockStart(
         metadata,
       };
       return;
-    }
+
     case "redacted_thinking":
       // "Redacted thinking" provides no actionable information, as everything is encrypted
       return;
@@ -300,6 +301,7 @@ function* handleContentBlockStart(
     case "fallback":
       // We don't use these Anthropic tools
       return;
+
     default:
       // New content block types may appear (e.g. via a new beta) before the
       // SDK types and this client are updated. Ignore rather than crash.
--- a/front/lib/api/llm/clients/anthropic/utils/anthropic_to_events.ts
+++ b/front/lib/api/llm/clients/anthropic/utils/anthropic_to_events.ts
@@ -9,6 +9,7 @@ import type {
   Message,
   MessageStreamEvent,
 } from "@anthropic-ai/sdk/resources/messages/messages.mjs";
+import { ANTHROPIC_PROVIDER_ID } from "@app/lib/api/llm/clients/anthropic/types";
 import { validateContentBlockIndex } from "@app/lib/api/llm/clients/anthropic/utils/predicates";
 import type { StreamState } from "@app/lib/api/llm/clients/anthropic/utils/types";
 import { SuccessAggregate } from "@app/lib/api/llm/types/aggregates";
@@ -16,6 +17,7 @@ import type {
   CacheMissReason,
   LLMEvent,
   LLMOutputItem,
+  ProviderPassthroughEvent,
   ReasoningDeltaEvent,
   ReasoningGeneratedEvent,
   TextDeltaEvent,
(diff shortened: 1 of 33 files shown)


front/lib/api/llm/clients/anthropic/utils/anthropic_to_events.ts
```diff
@@ -264,16 +266,25 @@ function* handleContentBlockStart(
         accumulator: "",
         accumulatorType: "tool_search",
         toolName: event.content_block.name,
+        toolId: event.content_block.id,
       };
       return;
 
     case "tool_search_tool_result":
```
aubin-tchoi, line 274: nit: why no curly braces here like the next one (or no curly brace in the next one)

front/lib/api/llm/clients/anthropic/utils/tool_search_passthrough.ts
```diff
@@ -0,0 +1,96 @@
...
+import { z } from "zod";
+
+// Anthropic runs tool search server-side inside a single assistant turn, emitting
+// `server_tool_use` and `tool_search_tool_result` blocks interleaved with the
+// thinking blocks. Those blocks must be replayed verbatim on the next request or
+// the thinking-block signatures are rejected. We persist them opaquely in the
+// generic content layer (as provider passthrough) and parse them back here with
+// full typing.
+//
+// The schemas mirror the Anthropic SDK param shapes so a parsed block is directly
```
aubin-tchoi, line 15: IIUC the schemas are not exposed by the SDK, only the types? is there a way to assert that `z.infer<serverToolUseSchema>` (and the result one) are compatible with the SDK's type?

front/lib/api/llm/clients/anthropic/utils/tool_search_passthrough.ts
```diff
@@ -0,0 +1,96 @@
...
+export const anthropicToolSearchBlockSchema = z.discriminatedUnion("type", [
+  serverToolUseSchema,
+  toolSearchToolResultSchema,
+]);
+
+export type AnthropicToolSearchBlock = z.infer<
+  typeof anthropicToolSearchBlockSchema
+>;
+
+function hasBlockType(value: unknown): value is { type: unknown } {
```
aubin-tchoi, line 60: nit: could be inlined into `block?.type` if it's just for loggin

front/lib/api/llm/transitionLLM.ts
```diff
@@ -83,7 +84,7 @@ function mapReasoningEffort(
 /**
  * Converts an old-system message to new BaseMessage(s).
  */
-function toBaseMessages(
+export function toBaseMessages(
```
aubin-tchoi, line 87: nit but why do we need to export this function and the one below? IMHO it's good that it stays local to the class

front/lib/model_constructors/types/input/messages.ts
```diff
@@ -62,10 +64,22 @@ export type BaseAssistantToolCallRequestMessage = {
   signature?: string;
 };
 
+// Opaque, provider-specific block carried verbatim so the producing provider
+// can replay it. Every other consumer skips it. The block is kept opaque here.
+export type BaseAssistantProviderPassthroughMessage = {
```
aubin-tchoi, line 69: why do I feel like this type is defined 4 times?

aubin-tchoi (approved):
LGTM

front/lib/api/llm/clients/anthropic/utils/anthropic_to_events.ts
flvndvd: Removed the other ones.

front/lib/api/llm/transitionLLM.ts
flvndvd: It was to add some tests. We had none...

front/lib/model_constructors/types/input/messages.ts
flvndvd: Only one actually, but very similar to the legacy router.

front/lib/api/llm/clients/anthropic/utils/tool_search_passthrough.ts
flvndvd: Good idea, added it in https://github.com/dust-tt/dust/pull/28129/changes/c2b4932fcbf8685fbccda2648869648fb0a3cef3.

PR #28136, opened 2026-06-29
Limit skill description suggestion to 750 characters

fabiencelier:
## Description

Closes https://github.com/dust-tt/tasks/issues/9299

Throw if the description suggestion is more than 750 characters
The reinforcement LLM will see the error and will be able to retry.

## Risk
low

## Deploy Plan

deploy front


davidebbo (approved):
LGTM clean!

PR #28137, opened 2026-06-29
Add bun to the sandbox base image

flvndvd:
## Description

Adds the `bun` runtime to the sandbox base image so agents and upcoming sandbox functions have a fast JavaScript and TypeScript runtime available alongside the existing node tooling.

`bun` is installed from a pinned release (1.3.14) using the same download-and-verify pattern already used for the other prebuilt binaries in the image: the release archive is fetched, checked against the published sha256 sums, and the binary is placed in a root-owned location that sandbox workloads cannot write to. The release assets, the path inside the archive, and the checksum step were all verified against the actual published release before pinning.

### DIFF
--- a/front/lib/api/sandbox/image/registry.ts
+++ b/front/lib/api/sandbox/image/registry.ts
@@ -19,7 +19,7 @@ import fs from "fs";
 import path from "path";
 
 const DUST_BEDROCK_IMAGE_VERSION = "1.10.0";
-const DUST_BASE_IMAGE_VERSION = "0.8.38";
+const DUST_BASE_IMAGE_VERSION = "0.8.39";
 const DSBX_CLI_VERSION = "0.1.26";
 // Identity, not coverage list: agent-proxied is a specific Linux user. The
 // nftables ruleset covers SANDBOX_UNTRUSTED_UIDS as a set; reordering that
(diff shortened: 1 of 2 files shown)


front/lib/api/sandbox/image/registry.ts
```diff
@@ -371,6 +374,20 @@ SHELLEOF`,
     runtime: "system",
     profile: "openai",
   })
+  .runCmd(
+    `curl -fsSL https://github.com/oven-sh/bun/releases/download/bun-v${BUN_VERSION}/bun-linux-x64.zip -o /tmp/bun.zip && ` +
+      `curl -fsSL https://github.com/oven-sh/bun/releases/download/bun-v${BUN_VERSION}/SHASUMS256.txt -o /tmp/bun-checksums.txt && ` +
+      "grep 'bun-linux-x64.zip' /tmp/bun-checksums.txt | awk '{print $1 \"  /tmp/bun.zip\"}' | sha256sum -c - && " +
```
davidebbo, line 380: A little complex but I assume it works :)

davidebbo (approved):
LGTM

fontanierh (approved):

front/lib/api/sandbox/image/registry.ts
flvndvd: They steer toward the install script (`curl -fsSL https://bun.com/install | bash -s "bun-v1.3.14"`) for pinned versions. I took inspiration from above with the direct-download-plus-checksum approach instead cause I did not want to pipe a remote script to bash as root 🙃

PR #28139, opened 2026-06-29
fix(llm): correct new-router OpenAI reasoning replay

adrsimon:
## Description

Under `use_new_llm_router`, gpt-5.5 failed with `400 Invalid 'input[N].id': string too long ... got a string with length 4044`. The old router was fine. Two bugs in the new router's reasoning round-trip; this fixes both.

Reasoning replay state is persisted as top-level `{ id, encrypted_content }` (the old router's shape). The new router carries it nested under `metadata.content` and bridges to/from that shape in `transitionLLM`.

**Read side (the 400):** on replay, `signature` was filled with the `encrypted_content` blob for every provider, and the OpenAI converter uses `signature` as the item `id` (64-char max) → 4044-char id. Now branches on provider: OpenAI puts the short `id` in `signature` and the blob in a new `encryptedContent` field (and skips cross-region items); Anthropic/Gemini keep their signature in `signature` as before.

**Write side (silent loss):** the bridges only lifted `content.signature` to top-level `encrypted_content`. OpenAI emits `content.{id, encryptedContent}`, so its reasoning persisted with neither key and was dropped on replay. New shared `reasoningContentToLegacyMetadata` helper writes the old-router shape, used by both the streaming and batch bridges.

## Tests

`transitionLLM.test.ts` (read + write mappings) and `openai_responses/converters/input/utils.test.ts` (converter). 14 tests, all green; `tsgo` clean.

## Risk

Low — behind `use_new_llm_router`. Non-OpenAI paths unchanged; OpenAI paths now mirror the legacy router. Safe to rollback.

## Deploy Plan

- [ ] Deploy `front`. No migration.

### DIFF
--- a/front/lib/api/llm/transitionLLM.ts
+++ b/front/lib/api/llm/transitionLLM.ts
@@ -20,7 +20,7 @@ import type {
 import { normalizePrompt } from "@app/lib/api/llm/types/options";
 import {
   extractEncryptedContentFromMetadata,
-  extractIdFromMetadata,
+  parseReasoningMetadata,
   parseResponseFormatSchema,
 } from "@app/lib/api/llm/utils";
 import type { Authenticator } from "@app/lib/auth";
@@ -165,15 +165,16 @@ export function toBaseMessages(
                   },
                 ];
               }
+              const { id, encryptedContent } = parseReasoningMetadata(
+                c.value.metadata
+              );
               return [
                 {
                   role: "assistant",
                   type: "reasoning",
                   content: { value: c.value.reasoning },
-                  signature: extractIdFromMetadata(c.value.metadata),
-                  encryptedContent: extractEncryptedContentFromMetadata(
-                    c.value.metadata
-                  ),
+                  signature: id,
+                  encryptedContent,
                 },
               ];
             }
--- a/front/lib/api/llm/transitionLLM.ts
+++ b/front/lib/api/llm/transitionLLM.ts
@@ -20,6 +20,7 @@ import type {
 import { normalizePrompt } from "@app/lib/api/llm/types/options";
 import {
   extractEncryptedContentFromMetadata,
+  extractIdFromMetadata,
   parseResponseFormatSchema,
 } from "@app/lib/api/llm/utils";
 import type { Authenticator } from "@app/lib/auth";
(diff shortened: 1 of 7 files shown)


front/lib/api/llm/transitionLLM.ts
```diff
@@ -141,20 +149,46 @@ function toBaseMessages(
...
+                  },
+                ];
+              }
+              // OpenAI encrypted content is only usable in the region that
+              // generated it (per-region keys), so skip cross-region items
+              // rather than send an unusable one.
+              const region = c.value.region ?? "us";
+              if (region !== REGION_MAPPING[regionConfig.getCurrentRegion()]) {
+                return [];
+              }
```
aubin-tchoi: I think this may be duplicated somewhere else, or in any case should probably not live here

front/lib/api/llm/transitionLLM.ts
```diff
@@ -186,6 +220,23 @@ function toBaseMessages(
...
+// The new router nests reasoning replay state under `metadata.content`: OpenAI
+// uses `id` + `encryptedContent`, Anthropic/Gemini use `signature`. Persist it in
+// the legacy top-level shape (`id` / `encrypted_content`) the replay path reads,
+// matching what the old router writes.
+export function reasoningContentToLegacyMetadata(
+  content: Record<string, unknown> | undefined
+): { id?: string; encrypted_content?: string } {
+  const id = typeof content?.id === "string" ? content.id : undefined;
+  const encryptedContent = [content?.encryptedContent, content?.signature].find(
+    (v): v is string => typeof v === "string"
```
aubin-tchoi: nit: could use `isString`, or even `content?.encryptedContent ?? content?.signature`

aubin-tchoi (approved):
LGTM except for the region check, which should be separated IMO

front/lib/api/llm/transitionLLM.ts
```diff
@@ -141,20 +149,46 @@ function toBaseMessages(
...
+                    ),
+                  },
+                ];
+              }
+              // OpenAI encrypted content is only usable in the region that
+              // generated it (per-region keys), so skip cross-region items
+              // rather than send an unusable one.
+              const region = c.value.region ?? "us";
+              if (region !== REGION_MAPPING[regionConfig.getCurrentRegion()]) {
+                return [];
```
davidebbo: Might make sense to log a warning? In theory, this should be transitional and go away after a while (unless people use really old convos)

davidebbo (approved):
LGTM

PR #27974, opened 2026-06-25
Polish plan mode card & add ability to close plan from there

PopDaph:
## Description

Refines the plan mode card that sits above the input bar and lets the user close the plan directly from it. Design to map Figma and align on the Card for wakeup.

<kbd>
Screenshot 2026-06-25 at 16 36 37
</kbd>

## Tests

Locally. 

## Risk

Gated. 

## Deploy Plan

Deploy front.

PR #28140, opened 2026-06-29
Fix plan card and panel intermittently not appearing after plan creation

PopDaph:
## Plan card and panel sometimes don't show up (and the panel won't close)

Follow up of #27974. After creating a plan the card and side panel sometimes never appeared, and after closing a plan the panel could stay open showing "No active plan". You had to refresh.

### Root cause

The plan card and panel react to a single signal: the `plan_updated` event. It rides the conversation-channel SSE, which keeps a replay buffer of exactly `5` events (`front/lib/api/assistant/streaming/events.ts:51`) and drops/reconnects constantly (`ERR_INCOMPLETE_CHUNKED_ENCODING`). When it dies around the moment `create_plan` fires, the event
is lost, the reconnect does not replay it, and nothing recovers until a page refresh. Close is on the same single thread of fate: nothing closes the panel except that one flaky event.

### Fix

Stop leaning on the unreliable channel, and stop driving the UI off a specific event. Revalidate the plan on the per-message `agent_action_success` event (`internalMCPServerName === "plan_mode"`), which rides the channel that delivers the tokens you can see (10 min retention, `lastEventId` replay + dedup). Then let the UI react to plan content:

- `PlanCard` (always mounted, the stable SWR subscriber) owns open and close via a pure `planPanelDecision` helper tracking one `PlanPresence` state (`unknown | empty | present`): open on `empty -> present`, close on `present -> empty`, `unknown` suppresses both on first load. This pulls open/close out of the `ConversationViewer` event handler, dropping the refs it needed.
- `handle_plan_updated` slimmed to `isClosed -> write null, else revalidate`.

Open and close now work however the change arrived, not just when one event happens to land. Net front only.

### Risk

Front only, gated by the `plan_mode` flag. Worst case the timing regresses to the old behavior.

Safe to rollback.

### DIFF
--- a/front/components/assistant/conversation/AgentMessage.tsx
+++ b/front/components/assistant/conversation/AgentMessage.tsx
@@ -52,6 +52,7 @@ import {
 import { useConversationAttachments } from "@app/hooks/conversations/useConversationAttachments";
 import { useConversationSandboxFiles } from "@app/hooks/conversations/useConversationSandboxFiles";
 import { useConversationSandboxStatus } from "@app/hooks/conversations/useConversationSandboxStatus";
+import { planFileKey } from "@app/hooks/conversations/usePlanFile";
 import { useAgentMessageStream } from "@app/hooks/useAgentMessageStream";
 import { useDeleteAgentMessage } from "@app/hooks/useDeleteAgentMessage";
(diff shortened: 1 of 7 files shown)


fontanierh (approved):
LGTM

PR #28143, opened 2026-06-29
feat(mobile): real-time voice transcription via ElevenLabs Scribe

adrsimon:
## Description

Replaces the file-upload voice transcription with **live streaming speech-to-text**, mirroring the web app's recently added real-time voice mode.

Audio is captured with `AVAudioEngine`, resampled to 16 kHz mono PCM, and streamed over a WebSocket to ElevenLabs Scribe v2 (`scribe_v2_realtime`, VAD commit strategy). Partial and committed transcripts arrive as the user speaks and fill the input bar in real time. A single-use token is minted by the backend (`/services/transcribe/get-token`) so the device never holds the workspace's ElevenLabs key.

Voice capture now happens in a dedicated **full-screen view**:
- White canvas that floods with blue as speaking amplitude rises, plus a pulsing orb.
- Auto-scrolling live transcript that keeps the latest tokens in view.
- Three controls: exit (down arrow, keeps the transcript), record/stop (the only control active while recording), and send.

Removes the obsolete `TranscriptionService` (file-upload path) and the inline recording UI.

## Tests

Built for the simulator (`make build`, green). Manually verified the mic flow: start → words stream in live → stop → final text lands in the input bar → send / exit. On-device verification of resampling quality and VAD segmentation feel is recommended since the simulator mic differs from a real device.

## Risk

Low and self-contained — touches only the voice/input-bar path. If the live transcription regresses, the feature is reachable only via the mic button and can be reverted independently of the other mobile changes. The backend token endpoint is already in production for the web client.

## Deploy Plan

Standard mobile build. No backend or schema changes.


aubin-tchoi (commented):
LGTM

PR #28144, opened 2026-06-29
dsbx function run: deliver response to the Dust result API

davidebbo:
## Description

After `dsbx function run` executes a function, deliver its response to the Dust result API (**`POST /api/v1/sandbox/sandbox-functions/result`**, relative to `DUST_API_URL`) instead of writing it to stdout. This is the (A) direction discussed: `dsbx` itself brokers the call, reusing the existing `DustApiClient` (Bearer `DUST_SANDBOX_TOKEN`, JSON, timeout) — the same plumbing `dsbx tools` uses.

- **`spawn_function`** (refactored from `run_bun`): optionally captures the runner's stdout and returns `(exit_code, Option<stdout>)` rather than `process::exit`ing, so the caller decides what to do with the response. Only stdout is piped (stderr/stdin inherited), so there's no deadlock. `get` behavior is unchanged.
- **`cmd_function_run`**:
  - **`DUST_SANDBOX_TOKEN` set** → POST the captured response to the result API; exit `0` on delivery, `{ "error": … }` on stdout + nonzero on failure. Nothing else on stdout.
  - **unset/empty (local/testing)** → fall back to writing the response to stdout and exit with the function's code (the previous behavior), so functions can still be run/tested locally without the API.
- **`DustApiClient::post_function_result(function, result)`**: posts `{ "function": <name>, "result": <runner Output JSON> }` and treats any 2xx as success (response body ignored, so an empty ack is fine).

## Status / assumptions

The endpoint **does not exist yet** — this is the client side, looking ahead. Provisional and easy to adjust when the API is designed:
- `DUST_API_URL` is assumed to include the `/api/v1` prefix (consistent with the existing `sandbox/actions` calls), so the relative path resolves to `/api/v1/sandbox/sandbox-functions/result`.
- Request body shape `{ function, result }` and the 2xx-ignored-body response.
- Exit `0` on successful delivery (the function's own ok/error is carried in the delivered payload).

## Tests

- `cargo fmt --check`, `cargo clippy --all-targets --all-features -- -D warnings`, `cargo test` (240) all pass.
- Manually verified: **no token** → response on stdout, exit 0; **token set + dead API URL** → response captured (not on stdout), POST attempted to `…/api/v1/sandbox/sandbox-functions/result`, failure surfaced as `{error}` + nonzero exit.

## Risk

Low / self-contained, behind the `function run` path. No-token behavior is unchanged, so local/testing flows are unaffected. Builds on the merged `dsbx function` feature.

## Deploy Plan

None — developer/sandbox tooling in `dsbx`. Becomes live once the result endpoint exists and `DUST_SANDBOX_TOKEN`/`DUST_API_URL` are present in the sandbox.

### DIFF
--- a/cli/dust-sandbox/src/api/client.rs
+++ b/cli/dust-sandbox/src/api/client.rs
@@ -234,4 +234,32 @@ impl DustApiClient {
 
         self.poll_action_result(&action_id).await
     }
+
+    /// Deliver a completed function's response to the sandbox-functions result
+    /// endpoint (`/api/v1/sandbox/sandbox-functions/result`, relative to
+    /// `DUST_API_URL`). The body carries the function name and the runner's
+    /// response JSON. Returns `Ok` on a 2xx; the response body is ignored, so an
+    /// empty/ack response is fine.
+    pub async fn post_function_result(
+        &self,
+        function: &str,
+        result: &serde_json::Value,
+    ) -> anyhow::Result<()> {
+        let url = self.url("sandbox/sandbox-functions/result");
+        let body = serde_json::json!({ "function": function, "result": result });
+        let resp = self
+            .client
+            .post(&url)
+            .json(&body)
+            .send()
+            .await
+            .context(format!("POST {url}"))?;
+
+        let status = resp.status();
+        if !status.is_success() {
+            let text = resp.text().await.unwrap_or_default();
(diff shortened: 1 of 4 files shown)


fontanierh (approved):
LGTM

PR #28145, opened 2026-06-29
fix(mobile): conversation, auth, and tool-approval improvements

adrsimon:
## Description

A batch of independent mobile fixes bundled into one PR. **Two of these are security/authorization-relevant** and worth a closer look:

- **🔒 Auth — ephemeral web session.** Logout only clears our tokens; the in-app browser previously kept its WorkOS cookies, so the next login silently re-authenticated the same account. Switches to an ephemeral cookie store so logout actually signs the user out.
- **🔒 Tool approvals — scope to the triggering user.** Approval and "ask user a question" prompts are now scoped to the user whose turn triggered the action. The triggering user's sId is carried through the stream events (`ToolApproveExecutionEvent`, `ToolAskUserQuestionEvent`, `BlockedAction`) and only that user may respond — other participants in a shared conversation can no longer approve someone else's tool call. API-key runs (null userId) are unaffected.

The rest are behavioral/cosmetic:

- **Messages — inline image previews** for image attachments in the thread.
- **Conversations — mark action-required as read.** `actionRequired` conversations also surface in the unread inbox; the server's read path self-heals a stale flag, so we mark those read too and optimistically clear the list's unread indicator via a notification.
- **Conversation list — reload guard.** The list re-triggered its initial load on every workspace switch, re-fetching `/api/user` and reverting the active workspace to the session default. Guarded so the initial load runs once.
- **Generated files — path-only files.** Handle files with a `filePath` and null `fileId` (oversized tool output offloaded to disk); they no longer crash the id/viewer path.

## Tests

Built for the simulator (`make build`, green). Manually exercised: logout/login no longer auto-signs-in; tool-approval prompt only actionable by the triggering user in a shared conversation; image attachments render inline; action-required conversations clear from the unread list; workspace switching stays on the selected workspace.

## Risk

The two authorization changes are the main risk surface:
- Ephemeral session: users will need to re-enter credentials more often (expected, and the point of the fix). Safe to roll back.
- Approval scoping relies on the server populating `userId` on the stream events; when null (API-key runs) the prompt remains actionable, preserving current behavior. No state migration; revertable per-file.

The remaining changes are additive/local to their views and low-risk.

## Deploy Plan

Standard mobile build. No backend or schema changes; depends on the server already emitting `userId` on tool-approval stream events (already present in the web flow).

### DIFF
--- a/x/adrsimon/mobile/Dust/Models/Attachment.swift
+++ b/x/adrsimon/mobile/Dust/Models/Attachment.swift
@@ -38,6 +38,10 @@ struct Attachment: Identifiable {
         contentType.hasPrefix(frameContentTypePrefix)
     }
 
+    static func isImage(_ contentType: String) -> Bool {
+        contentType.hasPrefix("image/")
+    }
+
     static func sfSymbol(for contentType: String) -> String {
         if contentType.hasPrefix(frameContentTypePrefix) { return "rectangle.on.rectangle" }
         if contentType.hasPrefix("image/") { return "photo" }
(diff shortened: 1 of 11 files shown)


aubin-tchoi (commented):
LGTM

PR #28147, opened 2026-06-29
feat(usage): bulk per-user spend-limit endpoint + temporal workflow

rfrenoy:
## Description

Add POST `/api/w/:wId/members/bulk-spend-limit`: resolves a selection (explicit userIds or "all matching filter minus excludes") into member sIds, then launches a bulk_spend_limit Temporal workflow and returns { workflowId, memberCount }.

The workflow chunks userIds (25/batch) over setSpendLimitForUsersActivity, which fans out the existing setUserSpendLimit with bounded concurrency (2) via concurrentExecutor, attributing per-member audit events to the acting admin. Idempotent, so Temporal retries are safe.

Extract resolveMatchingMemberUserIds from members_usage (shared seat/group restriction helper) to expand "select all matching" server-side.

## Tests

Unit + local

## Risk

Low, feature flagged

## Deploy Plan

Front


front/temporal/worker_registry.ts
```diff
@@ -61,6 +63,7 @@ export const workerFunctions: Record<WorkerName, () => Promise<void>> = {
   agent_trigger_webhook: runAgentTriggerWebhookWorker,
   analytics_queue: runAnalyticsWorker,
   conversation_fork_queue: runConversationForkQueueWorker,
+  bulk_spend_limit: runBulkSpendLimitWorker,
```
tdraier: adding a worker is maybe a bit overkill ?  we can use metronome worker

rfrenoy: fine by me, did not know what our policy was

rfrenoy: Fixed in fec4b7021785a22b475a6a895128f68f8faa8ad0

tdraier (approved):

PR #28153, opened 2026-06-29
Upgrade banner for free plans on Usage page

fabiencelier:
## Description

Closes https://github.com/dust-tt/tasks/issues/9001

Use the same banner as the billing page but with a different button to upgrade yourself directly.

Screenshot 2026-06-29 at 17 06 45
Screenshot 2026-06-29 at 17 06 50

## Tests

tested locally 

## Risk

low, it's only UI

## Deploy Plan

deploy front

### DIFF
--- a/front/components/pages/workspace/UsagePage.tsx
+++ b/front/components/pages/workspace/UsagePage.tsx
@@ -721,16 +721,16 @@ export function UsagePage() {
 
         {!isReadOnly && isCreditPricedFreePlan(subscription.plan.code) && (
           <FreePlanUpgradeSection
-              action={
-                <Button
-                  label="Change my seat"
-                  variant="highlight"
-                  size="sm"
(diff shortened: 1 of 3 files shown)


front/components/pages/workspace/UsagePage.tsx
```diff
@@ -744,21 +720,16 @@ export function UsagePage() {
 
         {!isReadOnly &&
           (myUsage?.seatType === "free" || myUsage?.seatType === "none") && (
```
thervier: @fabiencelier do we keep this condition? For me as soon as we are not anymore on a free plan we are good (because we dont want to force an admin to have a paid seat)

fabiencelier: changed the check to display the banner on whether the plan is a free plan or not

thervier (approved):
LGTM

PR #28154, opened 2026-06-29
feat(front): sticky cross-group filtering on usage credit chart

rfrenoy:
## Description

Make the credit usage chart's user/agent filter sticky across groupBy changes and shared with the credit tables.

## Tests

Local
## Risk

Low
## Deploy Plan

Front

### DIFF
--- a/front/components/workspace/AwuUsageFromAnalyticsChart.tsx
+++ b/front/components/workspace/AwuUsageFromAnalyticsChart.tsx
@@ -12,6 +12,15 @@ import {
 import { ChartContainer } from "@app/components/charts/ChartContainer";
 import type { LegendItem } from "@app/components/charts/ChartLegend";
 import { ChartTooltipCard } from "@app/components/charts/ChartTooltip";
+import type { AnalyticsFilter } from "@app/components/workspace/analytics/analyticsFilter";
+import {
+  isScopeDimension,
+  removeScopeEntity,
+  SCOPE_DIMENSION_LABEL,
+  SCOPE_DIMENSIONS,
+  scopeFilterToIds,
+  toggleScopeEntity,
+} from "@app/components/workspace/analytics/analyticsFilter";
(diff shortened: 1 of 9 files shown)


front/components/workspace/AwuUsageFromAnalyticsChart.tsx
```diff
@@ -104,6 +117,13 @@ export interface BaseAwuUsageFromAnalyticsChartProps {
...
+  // Sticky scope filters. When set, the chart is restricted to that user/agent
+  // and a removable chip is shown. Legend clicks in "user"/"agent" mode set the
+  // matching filter when the corresponding setter is provided.
+  userFilter?: AnalyticsEntityFilter | null;
+  agentFilter?: AnalyticsEntityFilter | null;
+  onUserFilterChange?: (filter: AnalyticsEntityFilter | null) => void;
+  onAgentFilterChange?: (filter: AnalyticsEntityFilter | null) => void;
 }
 
 export function BaseAwuUsageFromAnalyticsChart({
```
rfrenoy: will rework in a dedicated PR

tdraier (approved):

tdraier:
LGTM, but why specific to user/agent ? ppul graph was gathering any number of groups

rfrenoy:
@tdraier I will add `source`. Usage type is trickier because it's derived from the documents, it's not a field we retrieve directly. And I'm not sure it makes sense as a sticky filter

tdraier:
> I will add source. Usage type is trickier because it's derived from the documents, it's not a field we retrieve directly. And I'm not sure it makes sense as a sticky filter

my point was not to necessarily add all fields but rather why using dedicated custom states for every field we want to add

rfrenoy:
@tdraier Made changes to use the common logic with ppul chart. They're not targeting the same things so I had to create an abstraction layer. React doctor comments are legit but issues were already there and are surfaced because I moved the existing logic. Worth refactoring at some point

PR #28164, opened 2026-06-29
Initialize OpenTelemetry so Langfuse traces start at the HTTP request

flvndvd:
## Description

Since we moved to Hono, the `front-api` never initialized OpenTelemetry, unlike the Temporal workers which set it up at startup. As a result, any observation created while serving an HTTP request had no registered span processor and was silently dropped, so Langfuse traces appeared to begin only once execution reached the agent-loop worker, missing the originating request.

This wires up the same initialization the workers already use, registering the OTEL SDK and Langfuse span processor under the `dust-front` service name at boot, before requests are served. It is a no-op when Langfuse is disabled and uses the proven dd-trace + OTEL ordering that already runs in production. No explicit shutdown flush is needed since the prestop drain period comfortably exceeds the batch span processor's flush interval.

### DIFF
--- a/front-api/lib/instrumentation-config.ts
+++ b/front-api/lib/instrumentation-config.ts
@@ -1,0 +1,8 @@
+import { initializeOpenTelemetryInstrumentation } from "@app/lib/api/instrumentation/init";
+
+// Initialize OpenTelemetry / Langfuse instrumentation for the front-api process.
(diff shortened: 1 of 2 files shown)


flvndvd:
@ykmsd partially fixing your OTEL issue, will send another one to fix the tags.

ykmsd (approved):

PR #28165, opened 2026-06-29
[dsbx] Add function build subcommand and bake the sandbox function toolchain

flvndvd:
## Description

This PR adds the build half of the sandbox functions publish flow. A new `dsbx` function build subcommand bundles a function's source with its relative imports into a single self-contained module and extracts its JSON-Schema input/output contract, writing both to files. External packages like `zod` are left as imports rather than inlined, so bundles stay small and leverage harness at runtime. Bundling and the schema-extracting import run unprivileged (agent uid) when dsbx is invoked as root, matching run and get.

Results are written to files rather than returned over stdout, since sandbox command output can be truncated. stdout only carries a small ok/error envelope, and the caller (MCP tool) reads the artifacts back.

The function toolchain now resolves from the image's shared global `node_modules` rather than a bespoke directory. We can challenge this later but for now since we expect models to create function from the conversation's sandbox, it's nice to share the same libs. zod is added to the existing global install the conversation and pod sandboxes already use, and the runner's `NODE_PATH` points there so external imports resolve wherever a bundle lives.

This bumps the `dsbx` CLI to 0.1.27 and the sandbox base image to 0.8.40.

> [!WARNING]
> Note the image references the new CLI version, so the dsbx 0.1.27 release must be cut before the 0.8.40 image is built.

### DIFF
--- a/cli/dust-sandbox/functions-runner/build.ts
+++ b/cli/dust-sandbox/functions-runner/build.ts
@@ -1,0 +1,77 @@
+// Build a sandbox function: bundle its source with its relative imports into a single module and
+// extract its JSON-Schema I/O contract.
+//
+// External packages (zod and the rest of the sandbox "harness") are left as imports, not inlined:
+// the sandbox provides them at invocation time, which keeps bundles small and shares a single zod
+// instance with the runner.
+//
+// The bundle and schema are written to files rather than returned on stdout:
+// `dsbx function build` invocations are read back out-of-band by the caller
+// (sandbox results can be truncated). stdout only carries a small
+// `{ ok: true }` / `{ ok: false, error }` envelope.
+
+import { getFunctionSchema } from "./schema.ts";
+
+export type BuildErrorKind =
+  | "bad_args"
+  | "build_failed"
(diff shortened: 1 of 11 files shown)


cli/dust-sandbox/functions-runner/build.ts
```diff
@@ -0,0 +1,77 @@
...
+      return {
+        ok: false,
+        error: {
+          kind: "build_failed",
+          message: result.logs.map((log) => String(log)).join("\n"),
+        },
+      };
+    }
+
+    bundled = await result.outputs[0].text();
```
davidebbo: So the whole bundle is in memory? I would have thought that it would write it directly to a file

davidebbo (approved):
LGTM. Nice and simple!

cli/dust-sandbox/functions-runner/build.ts
flvndvd: The bundle is already fully in memory the moment `Bun.build()` returns, I updated it to use `Blob` over string so we don't copy it when getting the text.

PR #28167, opened 2026-06-29
Add sandbox readFile provider and route function bundles to a dedicated prefix

flvndvd:
## Description

On the same line as https://github.com/dust-tt/dust/pull/28165.

This PR implements file reads from the e2b sandbox provider, which previously threw, so callers can pull artifacts off a sandbox after they are written. Also introduces a dedicated storage prefix for published sandbox function bundles, separate from the read-write pod files prefix. Bundles are written only by front via the GCS API into a location that will be mounted as read-only on the function's sandbox. This is the storage and read plumbing the publish flow builds on. The publish tool and the read-only mount of this prefix for invocation come in the next PR.

### DIFF
--- a/front/lib/api/files/mount_path.test.ts
+++ b/front/lib/api/files/mount_path.test.ts
@@ -4,6 +4,7 @@ import {
   getConversationFilePath,
   getConversationFilesBasePath,
   getPodFilesBasePath,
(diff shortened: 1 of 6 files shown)


fontanierh (approved):
LGTM

PR #28169, opened 2026-06-29
feat(usage): wire bulk spend-limit modal with optimistic update

rfrenoy:
## Description

Replace the placeholder batch action with BulkEditSpendLimitModal + useBulkSetUserSpendLimit. On validate: POST the selection (explicit ids or filter+excludes) to the bulk endpoint, then optimistically patch the visible selected rows in the members-usage cache (effective = pool cap + seat allowance), show a toast, clear selection, and schedule invalidateMembersUsage so the async workflow's DB writes reconcile.

One limit: we can currently select and update limits on free/none seats, which should be prevented. As the feature is behind a disabled feature flagged, this is fine to ship as long as it gets sorted out. I'll fix it in a later PR today.

## Tests

Local

## Risk

Low, does not affect existing logic + feature flagged.

## Deploy Plan

Front

### DIFF
--- a/front/components/pages/workspace/UsagePage.tsx
+++ b/front/components/pages/workspace/UsagePage.tsx
@@ -43,7 +43,6 @@ import {
 } from "@app/lib/swr/credits";
 import { useGroups } from "@app/lib/swr/groups";
 import {
-  invalidateMembersUsage,
   useBulkSetUserSpendLimit,
   useMembersUsage,
   useUpdateMemberSeatType,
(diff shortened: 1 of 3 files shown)


front/components/pages/workspace/UsagePage.tsx
```diff
@@ -487,6 +512,55 @@ export function UsagePage() {
...
+      }
+
+      // The change is applied asynchronously by the workflow, and the effective
+      // cap (pool cap + the per-seat allowance) is resolved server-side, which
+      // the client can't reproduce. So we don't optimistically fabricate it.
+      // Clear the selection and revalidate a few times so the table reflects the
+      // true server values once the workflow lands.
+      selection.clearSelection();
+      for (const delayMs of [8000, 16000]) {
+        window.setTimeout(() => {
```
Fraggle: This smells :)

Also, we are not cancelling it on unmount.

Fraggle (approved):
LGTM

front/components/pages/workspace/UsagePage.tsx
rfrenoy: What would you suggest? I can remove it, but then we accept that data is stale until the admin refreshes the page. Which is also fine by me

Fraggle: Does it really need to be async in the backend? (i lack context)

rfrenoy: I haven't tested since using Temporal, I will check. Previously it was taking seconds, and so it felt weird to me that admins could not see the impact of their actions before seconds, and adding a loading state felt poor. Maybe it's better now. I'll come back to you

Fraggle: > the effective cap (pool cap + the per-seat allowance) is resolved server-side

Could we return the resolved value immediately while the temporal update is applying async ?

rfrenoy: Sorry, did not answer your question but discussed optimistic vs pessimistic. So async in the backend is necessary yes, since we can't parallelize calls to Metronome or we will get rate limited, and we can't wait for everything to settle. I'll just remove the cache invalidation logic

PR #28171, opened 2026-06-29
Accept sandbox function result tokens

fontanierh:
## Description

Follow up of #28123.

Root cause: the result callback endpoint was still authenticated like a sandbox action callback, so a sandbox function invocation token could be rejected before the stub result handler ran. The callback path also inherited the `sandbox_tools`/Computer gate even though invocation is controlled before token minting, and the strict result body schema did not accept the `function` field now posted by `dsbx function run`.

This PR mounts sandbox auth with an explicit token kind for each sandbox subtree: `/sandbox/actions/*` requires sandbox exec tokens and keeps the Computer gate, while `/sandbox/sandbox-functions/*` requires sandbox function invocation tokens and relies on the signed token. It also accepts the optional `function` field on result callbacks.

## Tests

- `front-api`: `NODE_ENV=test npm test -- routes/v1/w/[wId]/sandbox/sandbox-functions/result.test.ts routes/v1/w/[wId]/sandbox/actions/index.test.ts`
- `front-api`: `npx tsgo --noEmit`
- `front-api`: `npm run lint:swagger-annotations`
- Local smoke in `sandbox-function-smoke`: callback contract returned 200 and full `SandboxFunctionResource.invoke` returned `status: "created"` with `SBX_DEV_UNRESTRICTED_EGRESS=true` and `SBX_DEV_FRONT_URL` set to an HTTPS tunnel.

## Risk

Worst case, sandbox function result callbacks keep failing. Safe to rollback.

## Deploy Plan

- [ ] Merge this PR.
- [ ] Deploy front.

### DIFF
--- a/front-api/routes/v1/w/[wId]/sandbox/sandbox-functions/result.ts
+++ b/front-api/routes/v1/w/[wId]/sandbox/sandbox-functions/result.ts
@@ -6,6 +6,7 @@ import { z } from "zod";
 
 const PostSandboxFunctionResultRequestBodySchema = z
   .object({
+    function: z.string().optional(),
     result: z.unknown(),
   })
   .strict();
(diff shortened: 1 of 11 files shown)


front-api/routes/v1/w/[wId]/sandbox/sandbox-functions/result.ts
```diff
@@ -6,6 +6,7 @@ import { z } from "zod";
 
 const PostSandboxFunctionResultRequestBodySchema = z
   .object({
+    function: z.string().optional(),
```
Comment (davidebbo) [inline front-api/routes/v1/w/[wId]/sandbox/sandbox-functions/result.ts:9]: I think I'll remove this from the payload, we don't need it, as we should have everything encapsulated in the claims.

davidebbo (approved):
LGTM

PR #28172, opened 2026-06-29
fix: wire Metronome contract for CP_FREE_PLAN workspaces created via Poke

frankaloia:
## Summary

- `createWorkspaceInternal` (used by the Poke create_workspace plugin) was calling `internalSubscribeWorkspaceToFreePlan` for `CP_FREE_PLAN` workspaces, which only ensured a Metronome customer but never provisioned a contract. This left `metronomeContractId` null on the subscription, causing users who joined those workspaces to land in `on_pool` state.
- When `planCode` is `CP_FREE_PLAN` and Metronome billing is enabled, `createWorkspaceInternal` now calls `activateCreditPricedFreePlan` directly, which provisions the customer + contract + subscription with the contract ID wired in.
- Made the user seat-assignment step in `activateCreditPricedFreePlan` conditional on `auth.user()` being present. When called from `internalAdminForWorkspace` (no user, as at workspace creation time), the seat step is skipped — seat types are assigned when members join via the normal membership flow. Existing callers (e.g. `trial/start`) always have a user and are unaffected.

## Risk

Low. The change only affects Poke-created workspaces on `CP_FREE_PLAN` with Metronome enabled

## Testing

Use Poke create_workspace with `CP_FREE_PLAN` on staging

### DIFF
--- a/front/lib/api/subscription.ts
+++ b/front/lib/api/subscription.ts
@@ -19,6 +19,7 @@ import { CREDIT_PRICED_FREE_PLAN_CODE } from "@app/lib/plans/plan_codes";
 import { KillSwitchResource } from "@app/lib/resources/kill_switch_resource";
 import { SubscriptionResource } from "@app/lib/resources/subscription_resource";
 import { TriggerResource } from "@app/lib/resources/trigger_resource";
+import type { UserResource } from "@app/lib/resources/user_resource";
 import { renderLightWorkspaceType } from "@app/lib/workspace";
 import logger from "@app/logger/logger";
 import { terminateScheduleWorkspaceScrubWorkflow } from "@app/temporal/scrub_workspace/client";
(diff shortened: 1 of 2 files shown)


front/lib/api/subscription.ts
```diff
@@ -88,17 +88,19 @@ export async function activateCreditPricedFreePlan(
...
-    workspace: lightWorkspace,
-    newSeatType: "free",
-    author: "no-author",
-  });
-  if (seatResult.isErr()) {
-    throw new Error(
-      `Failed to update user to free seat: ${seatResult.error.type}`
-    );
+  const user = auth.user();
+  if (user) {
```
tdraier, line 92: In which case can we activate a free plan without user ?

front/lib/iam/workspaces.ts
```diff
@@ -95,20 +103,27 @@ export async function createWorkspaceInternal({
...
-    if (isUpgraded(newSubscription.getPlan())) {
-      const orgRes = await getOrCreateWorkOSOrganization(lightWorkspace);
-      if (orgRes.isErr()) {
-        logger.error(
-          { error: orgRes.error, workspaceId: workspace.sId },
-          "Failed to create WorkOS organization during workspace creation"
-        );
+    if (
+      isCreditPricedFreePlan(planCode) &&
+      (await isMetronomeBillingEnabled(auth))
```
tdraier: you should not be able to CreditPricedFreePlan with non metronome billed workspace. I would keep isCreditPricedFreePlan only .

tdraier (approved):

front/lib/api/subscription.ts
frankaloia: When it is created in poke, technically there is no user at that point of hte workflow. I will make this more clear in the code

front/lib/iam/workspaces.ts
frankaloia: The reason I kept that is so we still check consistently against our global kill switch before going through the workflow

PR #28173, opened 2026-06-30
fix(teams): list_messages pagination, createdDateTime range, and tool description

smb2268:
Fix https://github.com/dust-tt/tasks/issues/9298#event-27339318795
## Description
Three bugs in list_messages, all in the pagination/date path:

1. Empty results when toDate is in the past. Pagination only continued if the oldest message on the page was in-range; since Graph returns newest-first, page 1 is all newer than the window → stopped on page 1, never walked back. → Fix: continue until the oldest message on the page is older than fromDate (or the message limit is hit).
2.  Date-ranged on lastModifiedDateTime. A message sent in-range but edited/reacted-to later was dropped; an old message edited in-range was wrongly included. → Fix: range on createdDateTime (filter, stop condition, and server-side $filter/$orderby).
3.  messageId documented as valid for chats. Agents passed it on chats, where it has no meaning. → Fix: metadata now states messageId is channels-only.

Plus three hardening changes:
1. Push the toDate upper bound server-side for chats (skips the now→toDate over-fetch; channels can't filter server-side).
2. Replace the unbounded pagination loop with a transparent scan backstop (MAX_MESSAGES_TO_SCAN) that tells the caller when results are truncated.
3. Validate fromDate/toDate, guard empty pages, and unit-test the stop logic (pagination.test.ts).

## Tests
Manually verify

## Risk
Low, no schema/contract change. Worst case a huge channel range scans up to the backstop and returns partial history with a note saying so.

## Deploy Plan

- [ ] deploy front

### DIFF
--- a/front/lib/api/actions/servers/microsoft_teams/metadata.ts
+++ b/front/lib/api/actions/servers/microsoft_teams/metadata.ts
@@ -99,7 +99,7 @@ export const MICROSOFT_TEAMS_TOOLS_METADATA = createToolsRecord({
   },
   list_messages: {
     description:
-      "List messages from a Teams channel or chat (1:1, group, or meeting chat). Provide either chatId (for chats) or both teamId + channelId (for channels). For channels, returns top-level messages only. To read a full thread, call again with a messageId. For chats, returns all messages. Supports pagination and date range filtering.",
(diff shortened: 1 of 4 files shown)


frankaloia (approved):
LGTM

PR #28174, opened 2026-06-30
dsbx function: drop privileges via runuser (remove all unsafe)

davidebbo:
## Description

Addresses the review on #28088 ("no way around the unsafes?"). Removes **all** `unsafe` from the function privilege-drop path.

Previously, running the function as the unprivileged `agent-proxied` user required: `libc::getpwuid`/`getgrouplist` (FFI) to resolve its gid + supplementary groups, and a `pre_exec` closure doing `setgroups`/`setgid`/`setuid` in the forked child — four `unsafe` blocks, plus a `cfg(target_os = "linux")` gate because of libc signature differences.

This replaces that with a fully-safe approach: when `dsbx` runs as root, the function is executed via **`runuser -u agent-proxied -- bun …`** (util-linux), which resolves the user's uid/primary-gid/supplementary-groups and drops privileges before exec. No `unsafe`, no FFI, no fork/exec hook.

- **Root check**: `rustix::process::geteuid().is_root()` (safe). `rustix` was already in the tree (via `tempfile`); this just enables its `process` feature — no new crate downloaded.
- **Bundle access**: bundles can be root-only, so `dsbx` (still root) stages the bundle into a temp dir and makes it + the runner **world-readable** (`0644`/`0755`) rather than `chown`ing — so the dropped child can read them without a uid lookup. Contents aren't secret (the runner is embedded in the binary; the bundle is the function being executed) and temp names are random.
- The code is **no longer `cfg(linux)`-only** and has **no `libc`** usage, so it compiles and is testable on macOS dev too. Net −57 lines.

## Why an external tool

Setting supplementary groups (needed so the dropped child doesn't inherit root's groups) can't be done with a safe builder method: `std`'s `Command::groups` is unstable (`setgroups`, #90747) and `tokio` doesn't expose it — only `pre_exec` (unsafe) could. So the only `pre_exec`-free, fully-safe route is to delegate the whole drop to `runuser`, which also handles initgroups.

## Tests

- `cargo fmt --check`, `cargo clippy --all-targets --all-features -- -D warnings`, `cargo test` (240) all pass.
- `grep` confirms no `unsafe`/`libc::`/`getpwuid`/`pre_exec`/`cfg(target_os)` remain in the function module (only the word "unsafe" in a doc comment).
- Non-root e2e (local): runs `bun` directly, function executes, `cwd` = `$DUST_FUNCTIONS_DIR`.

## Risk / things to confirm

- **Runtime dependency on `runuser`** (util-linux). The sandbox image is apt-based (Debian/Ubuntu), which ships it; if the image were ever slimmed, the root path would fail-closed (function won't run as root rather than running privileged). Worth an explicit guarantee in the image build if you want certainty — happy to add it.
- The actual root→`agent-proxied` drop **can't be exercised on a non-root dev machine**; I verified compile/lint/tests and the non-root path locally. Needs a sandbox run to confirm the drop, egress containment, and group-based `/files` access.
- `runuser` (non-login) preserves the environment and cwd we set; flag if your image's PAM config (`/etc/pam.d/runuser`) differs.

## Deploy Plan

None — `dsbx` tooling. Behavior change only when `dsbx function run/get` is invoked as root in the sandbox.

### DIFF
--- a/cli/dust-sandbox/Cargo.lock
+++ b/cli/dust-sandbox/Cargo.lock
@@ -326,6 +326,7 @@ dependencies = [
  "lru",
  "rcgen",
  "reqwest",
+ "rustix",
  "rustls",
  "rustls-native-certs",
  "serde",
--- a/cli/dust-sandbox/Cargo.lock
+++ b/cli/dust-sandbox/Cargo.lock
@@ -313,7 +313,7 @@ checksum = "92773504d58c093f6de2459af4af33faa518c13451eb8f2b5698ed3d36e7c813"
 
 [[package]]
 name = "dust-sandbox"
-version = "0.1.27"
+version = "0.1.28"
 dependencies = [
  "anyhow",
  "base64",
--- a/cli/dust-sandbox/Cargo.toml
+++ b/cli/dust-sandbox/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "dust-sandbox"
-version = "0.1.27"
+version = "0.1.28"
 edition = "2021"
 
 [[bin]]
--- a/cli/dust-sandbox/Cargo.toml
+++ b/cli/dust-sandbox/Cargo.toml
@@ -21,6 +21,7 @@ httparse = "1"
 serde = { version = "1", features = ["derive"] }
 serde_json = "1"
 reqwest = { version = "0.12", default-features = false, features = ["json", "rustls-tls-native-roots"] }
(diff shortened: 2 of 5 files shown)


fontanierh (approved):
Nice! Thanks!

I think you also want to bump
Version for dsbx and sandbox img

davidebbo:
> I think you also want to bump version for dsbx and sandbox img

Good catch, done!

PR #28175, opened 2026-06-30
Remove temporary debug

Fraggle:
## Description

Removes the temporary workspace-gated `logger.info` block from `front-api/routes/v1/w/[wId]/assistant/conversations/index.ts` that was logging both the validated JSON body and raw body (`ctx.req.text()`) for workspace `ba367d3014`. Also drops the now-unused `logger` import.

## Tests

No functional change — deletion only. Verified the surrounding `app.post` handler logic is unaffected.

## Risk

Low. Pure removal of a debug-only code path that was never executed in production for any workspace other than `ba367d3014`. No behaviour change for any other tenant.

## Deploy Plan

Deploy front-api.


Fraggle:

- #28175 ◀

davidebbo (approved):
LGTM

PR #28059, opened 2026-06-26
[front] API key credit cap (6/10): enforcement gate

tdraier:
Part of the **per-API-key credit spend limit** stack. Builds on #28058.

### This PR — enforcement
`checkProgrammaticUsageLimits` previously skipped the per-key cap entirely for credit-priced plans. This adds the credit-priced branch: when the request is made with an API key whose credit state is `capped` (`isApiKeyCapped`, Redis-cached with DB fallback), it returns a `rate_limit_error` with a "credit spend limit" message. Legacy plans keep the existing ES usage-tally path unchanged.

This is the read side of the state the webhook (#28058) and reconcile (#28057) write.

### Next in stack
API contract + Swagger → UI + SWR → backfill plugin → tests.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #28058, opened 2026-06-26
[front] API key credit cap (5/10): webhook routing

tdraier:
Part of the **per-API-key credit spend limit** stack. Builds on #28057.

### This PR — webhook routing
`processMetronomeWebhook` now routes `alerts.spend_threshold_reached` / `alerts.spend_threshold_resolved` events that carry an `api_key_name` group value to the per-API-key dispatchers:

- **reached** → `dispatchApiKeyCapReached` (key → `capped`)
- **resolved** → `dispatchApiKeyCapResolved` (billing-cycle renewal resets `current_spend` → `on_pool`)

The new branch sits between the per-user (`user_id`) and programmatic (`usage_type`) branches — the three are mutually exclusive by group key, so a per-API-key alert no longer falls through to the workspace PAYG handler.

### Next in stack
Enforcement gate → API/UI/Swagger/audit → backfill plugin → tests.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #28051, opened 2026-06-26
[front] API key credit cap (2/10): Metronome per-key cap alerts

tdraier:
Part of the **per-API-key credit spend limit** stack. Builds on #28050.

### This PR — Metronome alert wiring
New `lib/metronome/alerts/api_key_caps.ts`, mirroring the per-user cap alert helpers:

- `upsertMetronomeApiKeyCapAlert({ metronomeCustomerId, workspaceId, keyName, awuCredits })` — idempotent `spend_threshold_reached` alert, threshold in AWU credits, scoped to the key via `group_values: [{ key: "api_key_name", value: keyName }]`.
- `clearMetronomeApiKeyCapAlert(...)` — archives the alert (releasing its uniqueness key).

### Why no Metronome product change is needed
`api_key_name` is already a billable-metric group key on the AWU products (`setup_new_pricing.ts`) and is emitted on every usage event, so spend alerts can filter on it directly — confirmed via the Metronome alert UI (it lists `api_key_name` as a selectable group key). No drop/recreate of products or `presentation_group_key` change required.

Spend is anchored to the billing period, so the `resolved` webhook fires at monthly renewal — matching the legacy per-key monthly cap semantics.

No warning/80%-threshold alert (the per-key state machine is 2-state: `on_pool` / `capped`).

### Next in stack
State machine + state read/write/cache → set-cap lib + reconcile → webhook routing → enforcement gate → API/UI/Swagger/audit → tests.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #28050, opened 2026-06-26
[front] API key credit cap (1/10): keys schema + credit state

tdraier:
First PR in the stack adding **per-API-key credit spend limits** on new-pricing (credit) plans, mirroring the existing per-user spend-limit machinery (Metronome `spend_threshold_reached` alerts → credit-state transitions).

### This PR — data model only
- **`keys.monthlyCapAwuCredits`** (nullable int): admin-set cap in AWU credits (1 credit = $0.01 = 10,000 microUSD). `NULL` = unlimited.
- **`keys.creditState`** (`"on_pool" | "capped"`, default `"on_pool"`): per-key credit state; the enforcement source on credit-priced plans (added in a later PR).
- Plumbs both fields through `KeyModel`, `KeyResource` (secret cache, `toJSON`, update helpers), `KeyType`, and adds the `ApiKeyCreditState` type + guard.

### Key names are NOT unique — the cap is per-name
Production data has many active keys sharing a `(workspaceId, name)` pair, and Metronome aggregates spend by `api_key_name`. So the cap is effectively **per-name**: all active keys sharing a name share the limit and are blocked/unblocked together (handled in the webhook PR). No uniqueness is introduced or enforced.

### Deploy note
The migration is `pre-deploy` (two plain `ADD COLUMN`s, no index) — safe to apply before the code ships.

> Migration SQL was hand-written in the generator's format (this Conductor workspace has no live front DB to diff against). Please re-validate / `migration:apply` in a warm env.

### Next in stack
Metronome alert wiring → state machine → set-cap lib/reconcile → webhook routing → enforcement gate → API/UI/Swagger/audit → backfill → tests.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #28176, opened 2026-06-30
[front] API key credit cap: reconcile key state at end of agent loop

tdraier:
Part of the **per-API-key credit spend limit** work. Dedicated PR on top of the stack (builds on #28127).

### Why
Metronome `spend_threshold_reached` alerts don't attribute spend by `api_key_name` (it's only a billable-metric group key, not the AWU products' **presentation** group key — `user_id` is). Empirically the alert stays `customer_status: "ok"` even when a key is well over its cap, so we can't rely on the alert→webhook path to flip a key to `capped`.

### What
Switch to a **pull** trigger in the Metronome usage activity (`emitMetronomeUsageEventsActivity`, end of agent loop). Right after `ingestMetronomeEvents`, if the request used a **capped, non-system** API key on a **credit-priced** plan with an active contract, call `reconcileApiKey` — which reads live per-`api_key_name` usage (the usage API *does* attribute correctly) and flips `creditState`. The next API call with that key is then blocked by the enforcement gate (#28059).

- Non-fatal: a reconcile failure only logs; the next message (or the poke `api_key` reconcile / a manual one) converges.
- Reuses the `KeyResource` already fetched for `api_key_name` attribution, and only runs for keys that actually have a cap (no extra work for uncapped keys).

### Note
This runs once per agent message that uses a capped key (a usage-API call each). It's the trigger the alert was supposed to be; if it proves chatty we can add a short cache/throttle. Pairs with the set-time reconcile (#28057) and the poke per-key reconcile (#28127).

### DIFF
--- a/front/temporal/usage_queue/activities.ts
+++ b/front/temporal/usage_queue/activities.ts
@@ -1,4 +1,5 @@
 import { isToolExecutionStatusFinal } from "@app/lib/actions/statuses";
+import { reconcileApiKey } from "@app/lib/api/metronome/reconcile_credit_state";
 import { syncMetronomeSeatCountForWorkspace } from "@app/lib/api/metronome/seat_sync";
 import {
   isProgrammaticUsage,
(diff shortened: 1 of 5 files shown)


tdraier:

- #28050
    - #28051
        - #28054
            - #28057
                - #28058
                    - #28059
                        - #28061
                            - #28063
                                - #28064
                                    - #28065
                                        - #28127
                                            - #28176 ◀

fabiencelier (approved):
Let's try this, until metronome can do the job

PR #28185, opened 2026-06-30
[lint] chore: remove MCP instructions grit rule

aubin-tchoi:
## Description

- Removes the Grit rule that warned on non-null `serverInfo.instructions` in MCP server metadata, plus its examples.
- This is not used anymore: the MCP server type does not allow setting an `instructions` field, and the remaining servers with instructions were cleaned up in https://github.com/dust-tt/dust/pull/27957 for Productboard and https://github.com/dust-tt/dust/pull/27927 for Salesforce.

## Tests

- `npm run format:ci`

## Risk

- Low.

## Deploy Plan

- No deploy.


ElPicador (approved):
Thanks for the clean up!

PR #28186, opened 2026-06-30
[Pods] Pods Default skills agent loop injection

xebeohp:
## Description
The code that deals with injecting the default skills into the agentic loop for the default skills feature in Pods. This way, conversations activated not through the input bar (Slack, API, triggers) also get the default skills that the user selected.

## Tests
Tests added to skill_resource.test.ts:
- injects a pod's default skills into enabledSkills
- does not inject pod defaults into a non-pod conversation
- dedups a skill that is both a pod default and conversation-enabled

Test added to project_metadata_resource.test.ts:
- Deleting a custom skill scrubs its sId from every pod's defaultSkillsIds array

## Risk
Low risk, behind the feature flag "pod_default_skills".

## Deploy Plan
- Last step of Pods default skills feature to merge
- Turn on feature flag for Dust-only and monitor

### DIFF
--- a/front/lib/resources/project_metadata_resource.ts
+++ b/front/lib/resources/project_metadata_resource.ts
@@ -91,17 +91,11 @@ export class ProjectMetadataResource extends BaseResource<ProjectMetadataModel> 
       },
     });
 
-    // Default skills ride along on each row (`defaultSkillsIds`), so no extra
-    // query is needed.
     return models.map((model) =>
       ProjectMetadataResource.fromModel(model, model.spaceId)
     );
   }
 
-  // Prune a skill's sId from every pod's inline default-skill list in the
-  // workspace. Default skills are stored as an sId array on `project_metadata`,
-  // so when a skill is deleted we strip its sId from the arrays that contain
-  // it via `array_remove`.
   static async removeSkillFromAllDefaultSkills(
     auth: Authenticator,
     skillSId: string,
@@ -109,8 +103,6 @@ export class ProjectMetadataResource extends BaseResource<ProjectMetadataModel> 
   ): Promise<void> {
     await ProjectMetadataModel.update(
       {
-        // Collapse to null (not an empty `{}` array) when this was the pod's last
-        // default skill, keeping the "no defaults" state consistent with setDefaultSkills.
         defaultSkillsIds: fn(
           "nullif",
           fn("array_remove", col("defaultSkillsIds"), skillSId),
--- a/front/lib/resources/project_metadata_resource.ts
+++ b/front/lib/resources/project_metadata_resource.ts
@@ -98,20 +98,20 @@ export class ProjectMetadataResource extends BaseResource<ProjectMetadataModel> 
 
   static async removeSkillFromAllDefaultSkills(
     auth: Authenticator,
-    skillSId: string,
+    skillId: string,
     transaction?: Transaction
   ): Promise<void> {
     await ProjectMetadataModel.update(
       {
         defaultSkillsIds: fn(
           "nullif",
-          fn("array_remove", col("defaultSkillsIds"), skillSId),
+          fn("array_remove", col("defaultSkillsIds"), skillId),
           literal("'{}'")
         ),
       },
       {
         where: {
-          defaultSkillsIds: { [Op.contains]: [skillSId] },
+          defaultSkillsIds: { [Op.contains]: [skillId] },
           workspaceId: auth.getNonNullableWorkspace().id,
         },
         transaction,
--- a/front/lib/resources/project_metadata_resource.ts
+++ b/front/lib/resources/project_metadata_resource.ts
@@ -4,11 +4,12 @@ import type { SkillResource } from "@app/lib/resources/skill/skill_resource";
 import { SpaceResource } from "@app/lib/resources/space_resource";
 import { ProjectMetadataModel } from "@app/lib/resources/storage/models/project_metadata";
 import type { ReadonlyAttributesType } from "@app/lib/resources/storage/types";
-import { makeSId } from "@app/lib/resources/string_ids";
+import { getResourceIdFromSId, makeSId } from "@app/lib/resources/string_ids";
 import type { PodMetadataType } from "@app/types/project_metadata";
 import type { ModelId } from "@app/types/shared/model_id";
 import type { Result } from "@app/types/shared/result";
 import { Ok } from "@app/types/shared/result";
+import { removeNulls } from "@app/types/shared/utils/general";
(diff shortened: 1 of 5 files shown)


Fraggle (approved):
LGTM but I would ask for @aubin-tchoi 's eyes on this one.

front/lib/resources/project_metadata_resource.ts
```diff
@@ -90,13 +91,34 @@ export class ProjectMetadataResource extends BaseResource<ProjectMetadataModel>
...
+      {
+        defaultSkillsIds: fn(
+          "nullif",
+          fn("array_remove", col("defaultSkillsIds"), skillSId),
+          literal("'{}'")
+        ),
+      },
+      {
+        where: {
+          defaultSkillsIds: { [Op.contains]: [skillSId] },
```
aubin-tchoi: unrelated to the PR but `defaultSkillIds` would have been more adequate

front/lib/resources/project_metadata_resource.ts
```diff
@@ -90,13 +91,34 @@ export class ProjectMetadataResource extends BaseResource<ProjectMetadataModel>
...
-    // Default skills ride along on each row (`defaultSkillsIds`), so no extra
-    // query is needed.
     return models.map((model) =>
       ProjectMetadataResource.fromModel(model, model.spaceId)
     );
   }
 
+  static async removeSkillFromAllDefaultSkills(
+    auth: Authenticator,
+    skillSId: string,
```
aubin-tchoi: should be `skillId` (`xxxId` -> string identifier, `xxxModelId` -> model ID)

front/lib/resources/project_metadata_resource.ts
```diff
@@ -90,13 +91,34 @@ export class ProjectMetadataResource extends BaseResource<ProjectMetadataModel>
...
+    transaction?: Transaction
+  ): Promise<void> {
+    await ProjectMetadataModel.update(
```
aubin-tchoi, line 119: nit:
```suggestion
    await this.update(
```

the `update` is overloaded on the base resource to also update the current instance in-place (no change here since you don't return the new value, but quite foot-gunny)

front/lib/resources/skill/skill_resource.ts
```diff
@@ -1439,10 +1441,36 @@ export class SkillResource extends BaseResource<SkillConfigurationModel> {
     });
   }
 
-  /**
-   * List skills for the agent loop, returning system skills, enabled skills, and
-   * equipped skills.
-   */
+  static async listPodDefaultSkillsForConversation(
```
aubin-tchoi, line 1453: OOC why not add this in `listEnabledByConversation`?

front/lib/resources/skill/skill_resource.ts
```diff
@@ -1458,16 +1486,31 @@ export class SkillResource extends BaseResource<SkillConfigurationModel> {
...
+
+    const conversationEnabledById = new Map(
+      conversationEnabledSkillsRaw.map((s) => [s.sId, s])
+    );
+    for (const skill of podDefaultSkills) {
+      if (!conversationEnabledById.has(skill.sId)) {
+        conversationEnabledById.set(skill.sId, skill);
+      }
+    }
+    const conversationEnabledSkills = [...conversationEnabledById.values()];
```
aubin-tchoi: they are enabled by default, not only available?

aubin-tchoi (commented):
LGTM, left a few nits + one question

front/lib/resources/skill/skill_resource.ts
xebeohp: Yes, would having them only be available and under `equippedSkills` be better?

front/lib/resources/project_metadata_resource.ts
xebeohp: Fixed.

front/lib/resources/skill/skill_resource.ts
xebeohp: Fixed.

front/lib/resources/skill/skill_resource.ts
aubin-tchoi: depends on what you want to do, what are those skills? are they selected by users?

aubin-tchoi: what kind of expectation do we create for them?

xebeohp: Yes, they are selected by users in the Pod settings tab and we expect them to act like manually added skills in conversations (and in this case they are enabled by default).

aubin-tchoi: by manually added skills in conversations you mean the ones you type in with the slash command?
if so they are not enabled by default, the agent enables them (and is prompted to)

aubin-tchoi: actually you need to let the agent enable them, pretty sure the instructions won't be rendered in your PR

xebeohp: Yes, I meant the skills you add with the slash command/dropdown menu. That clears things up. I will fix this, thanks!

front/lib/resources/skill/skill_resource.ts
```diff
@@ -1439,10 +1447,38 @@ export class SkillResource extends BaseResource<SkillConfigurationModel> {
...
+    if (!isPodConversation(conversation)) {
+      return [];
+    }
+
+    const spaceModelId = getResourceIdFromSId(conversation.spaceId);
+    if (spaceModelId === null) {
+      return [];
+    }
+
+    const [projectMetadata] = await ProjectMetadataResource.fetchBySpaceIds(
```
aubin-tchoi, line 1467: so this method is actually `fetchBySpaceModelIds`, and you should fetch by id instead of having to get the model ID from the string identifier (i.e. `ProjectMetadataResource` should do the job)

front/lib/resources/skill/skill_resource.ts
```diff
@@ -1458,7 +1494,6 @@ export class SkillResource extends BaseResource<SkillConfigurationModel> {
     equippedSkills: SkillResource[];
   }> {
     const { agentConfiguration, conversation } = params;
-    // Light type-guard to check whether we have a full AgentLoopExecutionData.
```
aubin-tchoi, line 1461: nit: let's keep this comment

aubin-tchoi (approved):
LGTM

front/lib/resources/skill/skill_resource.ts
xebeohp: Ok.

PR #28190, opened 2026-06-30
sandbox-functions: run real functions from bundles mount

davidebbo:
## Description

`SandboxFunctionResource.invoke()` previously wrote a hard-coded stub function into a per-invocation staging dir before running it. Now that function bundles are mounted at **`/bundles`**, run them directly:

- Set `DUST_FUNCTIONS_DIR` to `/sandbox-functions/pods/<podId>` (was a `/tmp/...` per-invocation staging dir). dsbx resolves `function run <slug>` to `/sandbox-functions/pods/<podId>/<slug>.ts`.
- Removed the staging dir entirely (no more writing/`rm -rf`/`mkdir`/heredoc) — the bundle is run directly from the mounted, read-only folder.
- `buildSandboxFunctionCommand` is reduced to `dsbx function run <slug>`; dropped the now-unused `node:path` import.

The invocation still passes the request envelope on stdin and a short-lived `DUST_SANDBOX_TOKEN`; the response is delivered by dsbx to the result API (this just checks the exit code and returns `status: "created"`).

## Tests

biome clean on the changed file. No behavioral change to the invoke contract beyond where the bundle is read from.

## Risk

Low. Single-file change to where functions are read; relies on the `/bundles` mount being present and readable by the `agent-proxied` user the function runs as.

## Deploy Plan

None beyond the `/bundles` mount being provisioned in the sandbox.

### DIFF
--- a/front/lib/metronome/api_key_credit_state_machine.transition.test.ts
+++ b/front/lib/metronome/api_key_credit_state_machine.transition.test.ts
@@ -29,7 +29,7 @@ describe("transitionApiKeyCreditState", () => {
     const result = await transitionApiKeyCreditState(
       key,
       { type: "api_key_cap_reached" },
-      { workspaceId: workspace.sId, keyId: key.id }
+      { workspaceId: workspace.sId, keyModelId: key.id }
     );
 
     expect(result.isOk()).toBe(true);
@@ -54,7 +54,7 @@ describe("transitionApiKeyCreditState", () => {
     const result = await transitionApiKeyCreditState(
       key,
       { type: "api_key_cap_resolved" },
-      { workspaceId: workspace.sId, keyId: key.id }
+      { workspaceId: workspace.sId, keyModelId: key.id }
     );
 
     expect(result.isOk()).toBe(true);
@@ -68,7 +68,7 @@ describe("transitionApiKeyCreditState", () => {
     const result = await transitionApiKeyCreditState(
       key,
       { type: "admin_cap_cleared" },
-      { workspaceId: workspace.sId, keyId: key.id }
+      { workspaceId: workspace.sId, keyModelId: key.id }
     );
 
     expect(result.isOk()).toBe(true);


front/lib/resources/sandbox_function_resource.ts
```diff
@@ -38,7 +37,9 @@ import type { Attributes, Transaction } from "sequelize";
 
 const SANDBOX_FUNCTION_WORKING_DIRECTORY = "/home/agent";
 const SANDBOX_FUNCTION_EXEC_TIMEOUT_MS = 2 * 60 * 1000;
-const SANDBOX_FUNCTION_STAGING_ROOT = "/tmp/dust-sandbox-functions";
+// Function bundles are mounted read-only here; dsbx runs them directly from this
+// folder (no staging), resolving `function run <slug>` to `<dir>/<slug>.ts`.
+const SANDBOX_FUNCTION_BUNDLES_DIR = "/bundles";
```
davidebbo: This is the wrong folder. Just using this for testing until we know the right folder.

front/lib/resources/sandbox_function_resource.ts
```diff
@@ -47,29 +46,11 @@ function dustAPIBaseUrlForSandbox(): string {
...
-    "    return Response.json({ ok: true });",
-    "  },",
-    "};",
-    "DUST_SANDBOX_FUNCTION_EOF",
-    `${DSBX_BIN_PATH} function run ${shellEscape(slug)}`,
-  ].join("\n");
+// A pod's function bundles are mounted read-only at
+// `/sandbox-functions/pods/<spaceSId>`; dsbx runs them directly from there (no
+// staging), resolving `function run <slug>` to `<dir>/<slug>.ts`.
+function bundlesDirForPod(podId: string): string {
```
fontanierh: nit: kind of a useless function IMHI

fontanierh (approved):
exciting

front/lib/resources/sandbox_function_resource.ts
davidebbo: Yeah, I'll inline

flvndvd (approved):

PR #28192, opened 2026-06-30
Publish sandbox functions from the pod sandbox

flvndvd:
## Description

This PR adds the publish half of sandbox functions: a model working in a pod writes a TypeScript source onto the pod-mounted filesystem, then publishes it by path. The source is bundled on the pod's own sandbox and its input and output JSON schemas are extracted from the function's schema export, so a published function carries a self-contained bundle plus a typed contract.

The build runs on the sandbox but never writes storage itself. The sandbox produces the bundle and schema in a non-mounted scratch directory and MCP tool reads them back, then front is the sole writer of the stored bundle. This is deliberate: if the sandbox could write bundle storage, a malicious function running on that same sandbox could forge or overwrite any bundle. The bundle is the only persisted file, stored in a dedicated prefix separate from the read-write pod files, so nothing inside a sandbox can reach or tamper with a published bundle. The source the model wrote stays on the mount and never becomes a tracked file.

Bundling executes untrusted code (extracting the schema imports the module and runs its top-level code), so it runs as the same egress-controlled, unprivileged user the invocation path already uses, rather than as root. The model-supplied path is resolved and traversal-checked through the existing file system layer rather than by rebuilding that logic. Publishing is idempotent on the function slug within a pod: re-publishing swaps the bundle and refreshes the contract in place, and a failed cleanup of the superseded bundle does not fail an otherwise successful publish.

The capability is exposed as a single MCP tool gated on pod write permission, which validates the slug and surfaces build and schema errors back to the caller so it can correct the source. This stacks on the read-file and dedicated-prefix groundwork in the prior change and depends on the bundler shipping in the sandbox image.

## Deploy Plan

No migrations to run only some refactoring around type imports.

### DIFF
--- a/front/lib/api/actions/servers/sandbox_functions/metadata.ts
+++ b/front/lib/api/actions/servers/sandbox_functions/metadata.ts
@@ -1,6 +1,6 @@
 import type { ServerMetadata } from "@app/lib/actions/mcp_internal_actions/tool_definition";
 import { createToolsRecord } from "@app/lib/actions/mcp_internal_actions/tool_definition";
-import { SANDBOX_FUNCTION_SLUG_REGEX } from "@app/types/api/sandbox/functions";
+import { SANDBOX_FUNCTION_SLUG_REGEX } from "@app/types/api/sandbox_functions";
 import type { JSONSchema7 as JSONSchema } from "json-schema";
 import { z } from "zod";
 import { zodToJsonSchema } from "zod-to-json-schema";
--- a/front/lib/api/actions/servers/sandbox_functions/metadata.ts
+++ b/front/lib/api/actions/servers/sandbox_functions/metadata.ts
@@ -1,5 +1,6 @@
 import type { ServerMetadata } from "@app/lib/actions/mcp_internal_actions/tool_definition";
 import { createToolsRecord } from "@app/lib/actions/mcp_internal_actions/tool_definition";
+import { SANDBOX_FUNCTION_SLUG_REGEX } from "@app/types/api/sandbox/functions";
 import type { JSONSchema7 as JSONSchema } from "json-schema";
 import { z } from "zod";
 import { zodToJsonSchema } from "zod-to-json-schema";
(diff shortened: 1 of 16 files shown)


spolu:
Neat.

front/lib/api/actions/servers/sandbox_functions/metadata.ts
```diff
@@ -36,7 +37,41 @@ export const SANDBOX_FUNCTIONS_TOOLS_METADATA = createToolsRecord({
       done: "Got sandbox function",
     },
   },
-  // TODO(SANDBOX_FUNCTION) Add publish tool once we have pod's sandboxes.
+  publish: {
+    description:
+      "Publish a sandbox function from a TypeScript source file in the current pod. The source " +
+      "must default-export a `fetch(request: Request): Promise<Response>` handler and export a " +
+      "`schema` with zod `input` and `output`. It is bundled on the pod sandbox (only `zod` is " +
```
fontanierh, line 44: only zod is available to import could mislead the agent into thinking that functions cannot import anything no ?

front/lib/api/sandbox_functions/build_on_sandbox.ts
```diff
@@ -0,0 +1,208 @@
...
+    "set -euo pipefail",
+    `rm -rf -- ${shellEscape(buildDir)}`,
+    `mkdir -p -- ${shellEscape(buildDir)}`,
+    // `--` stops the model-supplied source path from being read as a dsbx flag.
+    `${DSBX_BIN_PATH} function build -- ${shellEscape(srcSandboxPath)} ${shellEscape(bundlePath)} ${shellEscape(schemaPath)}`,
+  ].join("\n");
+
+  const execResult = await sandbox.exec(auth, command, {
+    timeoutMs: BUILD_EXEC_TIMEOUT_MS,
+    user: "agent-proxied",
```
fontanierh, line 104: We'll add a new user for this I think (since we don't want agent-proxied to be able to do this)

fontanierh: (I'll take)

front/lib/api/sandbox_functions/build_on_sandbox.ts
```diff
@@ -0,0 +1,208 @@
...
+ * Runs as `agent-proxied` (the egress-controlled invocation user) because extracting the schema
+ * imports the module and runs its untrusted top-level code.
+ */
+export async function buildSandboxFunctionOnSandbox(
+  auth: Authenticator,
+  {
+    sandbox,
+    srcSandboxPath,
+  }: {
+    sandbox: SandboxResource;
```
fontanierh: We don't ensureActive from here ?

fontanierh (approved):
Exciting !

front/lib/api/sandbox_functions/build_on_sandbox.ts
flvndvd: checked right before this call, but let me move it closer to the actual use.

front/lib/api/actions/servers/sandbox_functions/metadata.ts
flvndvd: Use this mainly to test, will add a skill so I can dry this.

PR #28195, opened 2026-06-30
[front] feat: sync coupons cross-region on creation

thervier:
## Description

Fixes https://github.com/dust-tt/tasks/issues/8912

CouponRedemptionModel.couponId is a numeric FK. When a workspace is relocated (US → EU), the coupon must exist in the destination DB with the same numeric id — otherwise the FK is broken. Today coupons are created manually per-region, which has already caused incidents.

What this PR does:
* When a coupon is created in Poke (US), it is synchronously pushed to EU via a new POST /api/lookup/coupons/sync endpoint using the existing REGION_RESOLVER_SECRET auth pattern. The push uses CouponModel.upsert with the explicit numeric id to guarantee both regions share the same row id.
* If the push fails, the local coupon is deleted and the caller gets a 500 — a coupon that exists in only one region is never allowed.
* In development, the push is skipped.
* Creating a coupon from EU is blocked at the API level (400). The Poke UI hides the create button in EU and shows an informational banner instead (driven by a canCreateCoupon field returned by GET /api/poke/coupons).
* No DB migration needed — reuses existing CouponModel columns and REGION_RESOLVER_SECRET.

## Tests

* Tested locally for display in page
* Added endpoint tests in front-api/routes/poke/coupons/index.test.ts

Will need to be testes in prod to make sure the mechanism works as expected from US to EUR region

## Risk

Low, impacts only coupon creation in Poke

## Deploy Plan

Deploy front

### DIFF
--- a/front-api/routes/lookup/index.ts
+++ b/front-api/routes/lookup/index.ts
@@ -1,8 +1,9 @@
 import { createHono } from "@front-api/lib/hono";
-
 import resource from "./[resource]";
+import coupons from "./coupons";
 
 const app = createHono();
+app.route("/coupons", coupons);
 app.route("/:resource", resource);
 
 export default app;
--- a/front/lib/resources/coupon_resource.ts
+++ b/front/lib/resources/coupon_resource.ts
@@ -167,7 +167,14 @@ export class CouponResource extends BaseResource<CouponModel> {
     };
   }
 
-  static async upsertBySId(
+  // Used exclusively for cross-region coupon sync. When a coupon is created in
+  // the main region (US) and pushed to the secondary region (EU), the EU row
+  // must share the exact same numeric primary key as the US row. This is
+  // required because CouponRedemptionModel.couponId is a numeric FK: if a
+  // workspace relocates cross-region, its redemptions must still resolve to the
+  // correct coupon row. INSERT ... ON CONFLICT (id) DO UPDATE preserves the PK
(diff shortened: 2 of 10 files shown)


front-api/routes/lookup/index.ts
```diff
@@ -1,8 +1,9 @@
 import { createHono } from "@front-api/lib/hono";
-
 import resource from "./[resource]";
+import coupons from "./coupons";
 
 const app = createHono();
+app.route("/coupons", coupons);
 app.route("/:resource", resource);
```
tdraier, line 7: can't coupon be added in "generic" resource ?

front/lib/resources/coupon_resource.ts
```diff
@@ -167,6 +167,35 @@ export class CouponResource extends BaseResource<CouponModel> {
     };
   }
 
+  static async upsertBySId(
```
tdraier:
```suggestion
  static async upsertById(
```

tdraier (approved):

front-api/routes/lookup/index.ts
thervier: I checked it as well (as it seemed weird) but the resources have a generic behavior that is readonly and here it is semantically different as we write as well => Claude recommended to add it separately

front/lib/resources/coupon_resource.ts
thervier: fixed

PR #28203, opened 2026-06-30
 feat(Slack MCP) add workspace toggle to allow Slack personal MCP footer removal

LeandreLeBizec:
## Description

Adds a workspace-level admin toggle to allow removal of the _"Sent via [Agent] on Dust"_ footer that is automatically appended to Slack messages posted by the Slack personal MCP tools (`post_message`, `schedule_message`). The setting (`slackPersonalAllowFooterRemoval`) is stored in workspace metadata and exposed via a new `SlackPersonalFooterRemovalToggle` component in the workspace Capabilities settings page. The `executePostMessage` and `executeScheduleMessage` helpers now accept an optional `showSentByFooter` parameter (defaults to `true`) and skip footer injection when explicitly set to `false`.

## Tests

Tested manually: toggling the setting in workspace settings and verifying that messages posted via the Slack personal MCP do/don't include the footer accordingly.

## Risk

Low. The footer is shown by default (`showSentByFooter !== false`), so existing behaviour is unchanged unless the admin explicitly enables the toggle.

## Deploy Plan

Deploy `front`.


front/lib/api/poke/plugins/workspaces/toggle_slack_personal_footer_removal.ts
```diff
@@ -0,0 +1,46 @@
+import { createPlugin } from "@app/lib/api/poke/types";
+import { updateWorkspaceMetadata } from "@app/lib/api/workspace";
+import { Err, Ok } from "@app/types/shared/result";
+
+export const toggleSlackPersonalFooterRemovalPlugin = createPlugin({
```
davidebbo: Do we need this if they have their own UI?

front/lib/api/actions/servers/slack_personal/tools/index.ts
```diff
@@ -555,7 +554,15 @@ export function createSlackPersonalTools(
...
     post_message: async (
-      { to, message, threadTs, fileId, unfurlLinks, unfurlMedia },
+      {
+        to,
+        message,
+        threadTs,
+        fileId,
+        unfurlLinks,
+        unfurlMedia,
+        show_sent_by_footer,
```
davidebbo, line 568: So one challenge is that model will always see this, but it will in most cases be ineffective since the toggle is off by default. Right?

davidebbo: Oh nm, I see you strip it, all good 👍

front/lib/api/poke/plugins/workspaces/toggle_slack_personal_footer_removal.ts
LeandreLeBizec: good catch thanks! It is legacy logic from a test I've made with a poke plugin instead of the workspace admin toggle!

davidebbo: So you're planning to remove it?

davidebbo:
Seems you have a failing test, but maybe unrelated? Try rebasing.

front/lib/api/poke/plugins/workspaces/toggle_slack_personal_footer_removal.ts
LeandreLeBizec: Yes, done, thx 🙏

davidebbo (approved):
LGTM

PR #28206, opened 2026-06-30
dsbx function: run bundles in place, any extension (drop staging)

davidebbo:
## Description

Functions are now mounted at `$DUST_FUNCTIONS_DIR` (`/sandbox-functions/pods/<podId>`, an agent-readable mount), and `bun` runs `.ts`/`.js`/`.mjs`/`.cjs` natively — so there's no reason to copy a bundle into a temp dir before running it. This drops staging and stops assuming `.ts`.

- **No staging:** removed `stage_bundle` and the staging branch. The bundle is run **in place** at its mount path. The root/`runuser` path now only chmods the *embedded-runner* temp file (created by root, 0600) so the dropped `agent-proxied` child can read it — the bundle itself is already agent-readable on the mount.
- **Extension-agnostic resolution:** instead of `format!("{name}.ts")`, scan `$DUST_FUNCTIONS_DIR` for a file whose stem is `<name>` (one match → run it; zero → `function not found`; multiple → ambiguous error). So a `<slug>.js` / `.mjs` / `.cjs` bundle works without renaming.
- **Schema name:** `get` derives `name` via `basename(path, extname(path))` rather than stripping a hard-coded `.ts`.
- Rewrote the resolution unit tests against a real temp dir (any extension, missing, ambiguous, unset env, bad name).

## Tests

- `cargo fmt --check`, `cargo clippy --all-targets --all-features -- -D warnings`, `cargo test` — all pass (7 function tests incl. the new resolution cases).
- Runner `bun test` — 32 pass.
- e2e (local, non-root): `.js` and `.mjs` bundles run and return 200; an ambiguous `<name>.js` + `<name>.ts` errors with a clear message.

## Risk

Low. dsbx-only. Removes a copy step and broadens resolution; relies (as before) on `$DUST_FUNCTIONS_DIR` being the agent-readable bundles mount.

## Deploy Plan

None directly. Ships in the next dsbx release (version bump + base-image bump can ride along when cutting it).


fontanierh (approved):

flvndvd:
Manually deployed both "a la mano". Merging.

PR #28207, opened 2026-06-30
[front] enh: tweak slack bot tools description for user ID formatting

aubin-tchoi:
## Description

We have seen examples where the `slack_bot` agent sometimes fails to correctly fill in Slack user IDs in mentions and instead puts an ID from another system (e.g. a HubSpot ID), producing broken `<@...>` mentions in posted/edited messages.

This PR changes the description copy only; no behavior change. It's split into two commits, first a pure line-wrap of the existing strings (content identical), then the wording changes so the second commit's diff is just the new guidance.

## Tests

- Tested locally.

## Risk

- Low.

## Deploy Plan

- Deploy front.

### DIFF
--- a/front/lib/api/actions/servers/slack_bot/metadata.ts
+++ b/front/lib/api/actions/servers/slack_bot/metadata.ts
@@ -17,7 +17,12 @@ export const SLACK_BOT_TOOLS_METADATA = createToolsRecord({
       message: z
         .string()
         .describe(
-          "The message to post, using standard Markdown formatting (e.g., [text](url) for links, **bold**, *italic*). Do NOT use Slack-specific markup like <url|text> for links. The system converts Markdown to Slack format automatically. To mention a user, use <@USER_ID>. To reference a channel, use #CHANNEL or <#CHANNEL_ID>."
+          "The message to post, using standard Markdown formatting " +
+            "(e.g., [text](url) for links, **bold**, *italic*). Do NOT " +
+            "use Slack-specific markup like <url|text> for links. The " +
+            "system converts Markdown to Slack format automatically. To " +
+            "mention a user, use <@USER_ID>. To reference a channel, use " +
+            "#CHANNEL or <#CHANNEL_ID>."
         ),
       threadTs: z
         .string()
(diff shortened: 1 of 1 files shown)


front/lib/api/actions/servers/slack_bot/metadata.ts
```diff
@@ -17,7 +17,18 @@ export const SLACK_BOT_TOOLS_METADATA = createToolsRecord({
...
+            "(e.g., [text](url) for links, **bold**, *italic*). Do NOT " +
+            "use Slack-specific markup like <url|text> for links. The " +
+            "system converts Markdown to Slack format automatically. To " +
+            "mention a user, use <@USER_ID>, where USER_ID is a Slack " +
+            "user ID that always starts with 'U' and contains only " +
+            "uppercase letters and digits (e.g., <@U01234ABCD>), so " +
+            "every valid mention starts with '<@U'. Never use an ID " +
+            "from another system (e.g. a HubSpot, Notion, CRM, or " +
+            "database ID) as the Slack user ID; resolve it with the " +
+            "search_user tool if needed. To reference a channel, use " +
```
LeandreLeBizec, line 30: > resolve it with the search_user tool if needed.

This one is a bit ambigious + refering another tool? With the precise description of the slack user id, is it needed?

front/lib/api/actions/servers/slack_bot/metadata.ts
```diff
@@ -63,7 +74,18 @@ export const SLACK_BOT_TOOLS_METADATA = createToolsRecord({
...
```
LeandreLeBizec, line 87: same

LeandreLeBizec (approved):
LGTM, minor concern about refering another tool in a tool desciption

front/lib/api/actions/servers/slack_bot/metadata.ts
aubin-tchoi: from what I saw, when it worked it had to use this other tool to find the ID

aubin-tchoi: didn't seem to be able to find it otherwise (even if you know what it's supposed to look like you need to know where to look)

PR #28209, opened 2026-06-30
SandboxFunctions invocation events endpoint

spolu:
## Description

Introduce a new events endpoint for sandbox function invocations to track creation and result events from the frame wrapper.

## Tests

Covered by some tests

## Risk

None

## Deploy Plan

- deploy `front`
- deploy `front-sse`


front/lib/resources/sandbox_function_invocation_resource.ts
```diff
@@ -68,6 +72,39 @@ export class SandboxFunctionInvocationResource extends BaseResource<SandboxFunct
...
+    if (!isResourceSId("sandbox_function_invocation", invocationId)) {
+      return null;
+    }
+
+    const invocationModelId = getResourceIdFromSId(invocationId);
+    if (invocationModelId === null) {
+      return null;
+    }
+
+    const invocation = await this.model.findOne({
```
flvndvd: no baseFetch on this resource?

flvndvd (approved):

fontanierh (approved):
Looks great

front/lib/resources/sandbox_function_invocation_resource.ts
spolu: ah yeah good catch. will fix

spolu: done

PR #28225, opened 2026-06-30
Mount published sandbox functions read-only into the pod sandbox

flvndvd:
## Description

This makes a pod's published function bundles available inside its own sandbox so they can be executed, while keeping the backend the only writer of them.

The dedicated, backend-only storage location that holds a pod's bundles is now mounted into that pod's sandbox as a read-only filesystem, and the function runtime reads bundles directly from it instead of materializing a placeholder into a scratch directory at invocation time.

Read-only is enforced where it actually matters, which is the storage credential rather than the mount itself. The credential backing the mount is served to the filesystem over loopback and is therefore reachable by untrusted code running in the sandbox, so a write-capable credential would let a function overwrite or forge a published bundle and defeat the integrity guarantee the dedicated location exists to provide. The downscoped credential now grants only object reads for the bundle location, while the writable pod-files location keeps read-write as before. The mount also carries the operating-system read-only flag as a second layer, but the credential scope is the real protection.

The bundle mount sits under a per-pod path, mirroring how a pod's regular files are already mounted, so a single sandbox could carry more than one pod's functions without collision later on.

The generic filesystem abstraction stays generic: it only gains the notion of a mount that lives inside the sandbox but is never exposed through the agent-visible path API. The specific decision that a pod's sandbox mounts its functions read-only lives with the pod sandbox lifecycle, which already owns what gets mounted there.

image

### DIFF
--- a/front/lib/api/files/mount_path.ts
+++ b/front/lib/api/files/mount_path.ts
@@ -84,7 +84,7 @@ export function getPodSandboxFunctionsBasePath({
   workspaceId: string;
   podId: string;
 }): string {
-  return `${getBaseMountPathForWorkspace({ workspaceId })}pods/${podId}/sandbox_functions/`;
+  return `${getBaseMountPathForWorkspace({ workspaceId })}pods/${podId}/sandbox-functions/`;
 }
 
 /**
@@ -93,7 +93,7 @@ export function getPodSandboxFunctionsBasePath({
  * collision.
  */
 export function getPodSandboxFunctionsMountPoint(podId: string): string {
-  return `/sandbox_functions/pods/${podId}`;
+  return `/sandbox-functions/pods/${podId}`;
 }
 
 /**
--- a/front/lib/api/files/mount_path.ts
+++ b/front/lib/api/files/mount_path.ts
@@ -88,6 +88,15 @@ export function getPodSandboxFunctionsBasePath({
 }
 
 /**
+ * Absolute in-sandbox path the pod's published bundles are mounted at. Pod-scoped like the
+ * `/files/pod-<id>` files mount, so one sandbox could carry several pods' functions without
+ * collision.
+ */
+export function getPodSandboxFunctionsMountPoint(podId: string): string {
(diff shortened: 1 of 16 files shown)


front/lib/api/files/mount_path.ts
```diff
@@ -87,6 +87,15 @@ export function getPodSandboxFunctionsBasePath({
   return `${getBaseMountPathForWorkspace({ workspaceId })}pods/${podId}/sandbox_functions/`;
 }
 
+/**
+ * Absolute in-sandbox path the pod's published bundles are mounted at. Pod-scoped like the
+ * `/files/pod-<id>` files mount, so one sandbox could carry several pods' functions without
+ * collision.
+ */
+export function getPodSandboxFunctionsMountPoint(podId: string): string {
+  return `/sandbox_functions/pods/${podId}`;
```
flvndvd: Used the new pattern on which we aligned. Not exposed to the model anyway.

fontanierh (approved):
clean

front/lib/api/files/mount_path.ts
spolu: A bit distaste full to have sandbox_functions then pods but I undertstand it's to avoid confusion with /pods/{podId} and /pods/sandbox-functions/{podIds} so I guess it's good.

Meta question: from a "product standpoint" we won't call it "sandbox functions" so we should not call it as such when presenting to the model. 

We don't know what will be the final name but expect this path to change

spolu (approved):
Looks reasonable

front/lib/api/files/mount_path.ts
fontanierh: I think final name likely `Pod Functions` no ?

fontanierh: maybe `Pod Actions`

flvndvd: For now this path isn't exposed anywhere to the model happy to change in the future.

PR #28227, opened 2026-06-30
[ci] Cancel superseded workflow runs on feature branches

flvndvd:
## Description

See thread [here](https://dust4ai.slack.com/archives/C050SM8NSPK/p1782818734673559).

Our CI runs on GitHub-hosted runners, so we're capped at a fixed number of concurrent jobs. The heavy push-triggered workflows had no concurrency control, so every push to a branch started a fresh full matrix without cancelling the previous in-flight one. When several branches push repeatedly, runs stack up and queue behind the concurrent-job ceiling, which is why we periodically run short on runners.

This adds a per-branch concurrency group to those workflows so that pushing a new commit to a feature branch cancels the superseded run of the same workflow on that same branch. Cancellation is scoped strictly to the branch being pushed, so one branch never affects another.

`main` is deliberately exempt: every commit that lands on main keeps a full CI record rather than being cancelled by the next merge. The tradeoff is that main can still stack runs during merge bursts, which we accept in exchange for per-commit verification.

This is a steady-state improvement for repeated pushes to the same branch. It does not prevent spikes caused by many distinct branches being pushed at once, which remain a separate capacity question.

### DIFF
--- a/.github/workflows/build-and-test-core.yml
+++ b/.github/workflows/build-and-test-core.yml
@@ -4,6 +4,12 @@ on:
   push:
   workflow_dispatch:
 
+# Cancel superseded runs on feature branches when a new commit is pushed; never
+# cancel on main so every mainline commit keeps a full CI record.
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: ${{ github.ref != 'refs/heads/main' }}
+
 permissions:
   contents: read
   pull-requests: write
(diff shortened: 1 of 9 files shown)


ElPicador (approved):

rfrenoy:
@flvndvd I'm not sure this would solve the issue, as @ElPicador I'm assuming in your stack you have n branches but only 1 commit per branch?

flvndvd:
From the math, we had 33% of our running actions resulting from this. 

Out of 303 active runs, only 202 has distinct branch + workflow pairs (so 101 redundant stacked runs).

PR #28230, opened 2026-06-30
fix(slack): descriptive error when channel blocks bot posting

rfrenoy:
## Description

Ticket: https://github.com/dust-tt/tasks/issues/9215

When a Slack channel restricts who can post, the bot's reply is rejected and users only saw the generic "An unexpected error occurred" message, giving no hint that it was a posting-permission denial.

Detect the relevant Slack platform error codes (checked relevant error codes from https://docs.slack.dev/reference/methods/chat.postMessage) and surface a descriptive message pointing admins at the channel posting permissions.

## Tests

Unit

## Risk

Low
## Deploy Plan

Connectors


spolu (approved):
LGTM \o/

PR #28237, opened 2026-06-30
chore(mcp): improve common_utilities tool descriptions for BM25 tool-search recall

aubin-tchoi:
## Description

This PR tweaks the `common_utilities` tool descriptions so each ranks better to the intended tool in tool search and is better segregated under the BM25 norm ([design doc](https://app.notion.com/p/dust-tt/Design-Doc-Anthropic-Token-Consumption-Cache-Efficiency-38728599d94180a2b7f8e101a776f564?source=copy_link#a7d3294f78444b25966d56efe3f3bd51), [tool-search doc](https://platform.claude.com/docs/en/agents-and-tools/tool-use/tool-search-tool)).

- `generate_random_number`: now says "integer (whole number)" and "pick a random number in a range" so it segregates from the `generate_random_float` sibling (both previously matched the bare token "number").
- `math_operation`: was just "Perform mathematical operations.", now leads with "Calculate" and names arithmetic/percentages to match typical use cases (this tool was not showing up for "calculate 15 percent of 240").

The other four tools already ranked #1 and were left untouched. Also registers `common_utilities` in the BM25 testing script with one labeled query per tool.

Note: some of the tools here may be deprecated in a near future (e.g. you can just rush bash on a sandbox to do computations) so not spending too much time here.

No behavior change.

## Tests

- Tested locally with BM25 samples.

## Risk

- Low.

## Deploy Plan

- Deploy front.

### DIFF
--- a/front/lib/api/actions/servers/common_utilities/metadata.ts
+++ b/front/lib/api/actions/servers/common_utilities/metadata.ts
@@ -12,7 +12,7 @@ const MAX_WAIT_DURATION_MS = 3 * 60 * 1_000;
 export const COMMON_UTILITIES_TOOLS_METADATA = createToolsRecord({
   generate_random_number: {
     description:
-      "Generate a random positive number between 1 and the provided maximum (inclusive).",
+      "Generate a random integer (whole number) between 1 and the provided maximum (inclusive). Pick a random number in a range.",
     schema: {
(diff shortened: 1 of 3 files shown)


aubin-tchoi:
PMRR

PR #28242, opened 2026-06-30
fix(input-bar): center "Search knowledge" empty state in its container

rfrenoy:
## Description

Ticket: https://github.com/dust-tt/tasks/issues/9320

The empty-state block is rendered inside sparkle's DropdownMenuContent, whose Radix ScrollArea wraps children in a content-height `display:table` div. The block's `h-full` centering resolved against that collapsed height, so it rendered at the top and crowded the separator. Force the Radix scroll wrapper to fill the viewport so the existing centering resolves correctly.

Screenshot 2026-06-30 at 15 38 19

## Tests

Local

## Risk

Low

## Deploy Plan

Front


ykmsd (approved):
ty!

PR #28246, opened 2026-06-30
chore(mcp): improve files tool descriptions for BM25 tool-search recall

aubin-tchoi:
## Description

- Tweaks the `files` MCP server tool descriptions so they rank and segregate better under the BM25 norm used by tool-search. The `grep` tool description is reworded to lead with the "regular expression pattern" intent vocabulary and to keep the line-number/`cat` cross-reference compact.
- Registers the `files` server in the BM25 testing script (`scripts/mcp_bm25/run.ts`) and adds a labeled `// --- files ---` query section (`scripts/mcp_bm25/queries.ts`) so each `files.*` tool has a representative query covered by the harness.
- No behavior change: descriptions are metadata only, and the script edits are diagnostic tooling.

## Tests

- Ran the BM25 harness locally; all 10 `files.*` queries rank 1, no `files.*` misses.

## Risk

- Low.

## Deploy Plan

- Deploy front.

### DIFF
--- a/front/lib/api/actions/servers/files/metadata.ts
+++ b/front/lib/api/actions/servers/files/metadata.ts
@@ -217,9 +217,9 @@ const FILES_TOOLS_COMMON_METADATA = {
   },
   [FILES_GREP_ACTION_NAME]: {
     description:
-      "Search a text file for lines matching a regular expression. " +
-      "Returns matching lines with their line numbers. " +
-      `Use the line numbers with \`${getPrefixedToolName(FILES_SERVER_NAME, FILES_CAT_ACTION_NAME)}\` to read surrounding context. ` +
+      "Search a text file for lines matching a regular expression pattern. " +
+      "Returns each matching line with its number, which you can pass to " +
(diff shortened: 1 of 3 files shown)


front/lib/api/actions/servers/files/metadata.ts
```diff
@@ -217,9 +217,9 @@ const FILES_TOOLS_COMMON_METADATA = {
   },
   [FILES_GREP_ACTION_NAME]: {
     description:
-      "Search a text file for lines matching a regular expression. " +
-      "Returns matching lines with their line numbers. " +
-      `Use the line numbers with \`${getPrefixedToolName(FILES_SERVER_NAME, FILES_CAT_ACTION_NAME)}\` to read surrounding context. ` +
+      "Search a text file for lines matching a regular expression pattern. " +
```
ElPicador, line 220: wondering if we shouldn't steal the description from claude code

ElPicador (approved):

front/lib/api/actions/servers/files/metadata.ts
aubin-tchoi: theirs doesn't work quite the same as ours, but there are def some potential improvements to do here!

PR #28255, opened 2026-06-30
fix(front): classify Anthropic mid-stream errors like the old router

adrsimon:
## Description

Since the `use_new_llm_router` rollout, mid-stream Anthropic failures (socket reset / premature close) surface as terminal `unknown_error` ("Unknown error from Anthropic") across sonnet/opus/haiku. Zero before 2026-06-17, ~284 on 2026-06-30.

The SDK's `messages.stream()` rewraps non-`APIError` body-read failures into a bare `AnthropicError` (it's not an `APIConnectionError` — that's only built at connect time). `classifyStreamError` matched none of its cases and fell through to a non-retryable `unknown_error`. The old router (`lib/api/llm`) classified the same errors via `categorizeLLMError` and kept them retryable, so they were silently retried and never surfaced.

This restores the old router's classification:
- `apiErrorToErrorEvent`: default missing status to 500 (mirrors `categorizeAnthropicError`'s `status ?? 500`).
- New `bareStreamErrorToErrorEvent`: ports `categorizeLLMError`'s substring matching to the new `ErrorType` union.

Substring matching is a deliberate 1:1 replica of the old router for migration parity (GEN1); commented to revisit once it's retired.

## Tests

Added parity cases in `output/utils.test.ts` (bare `AnthropicError` → network/rate-limit/overloaded/timeout/stream/server, unmatched → unknown; statusless `APIError` → server_error). 86 pass; tsgo + biome clean.

## Risk

Low, confined to Anthropic stream-error classification. Worst case: a terminal error containing a matched keyword gets retried a few times before failing. Matches old-router behavior. Revert to roll back.

## Deploy Plan

Standard deploy, no migration. Watch `@errorContent.message:"Unknown error from Anthropic"` drop after rollout.

### DIFF
--- a/front/lib/model_constructors/sdk/anthropic_ai/converters/output/utils.ts
+++ b/front/lib/model_constructors/sdk/anthropic_ai/converters/output/utils.ts
@@ -23,6 +23,7 @@ import {
 import type { EndpointMetadata } from "@app/lib/model_constructors/types/endpoint_metadata";
 import type {
   ErrorEvent,
+  ErrorType,
   ModelResponseEvent,
   NonDeltaResponseEvent,
   ProviderPassthroughEvent,
@@ -41,6 +42,7 @@ import {
   assertNever,
   assertNeverAndIgnore,
 } from "@app/types/shared/utils/assert_never";
+import { normalizeError } from "@app/types/shared/utils/error_utils";
 import { isRecord } from "@app/types/shared/utils/general";
 import { safeParseJSON } from "@app/types/shared/utils/json_utils";
 
@@ -401,7 +403,9 @@ function apiErrorToErrorEvent(
   metadata: EndpointMetadata,
   error: APIError
 ): ErrorEvent {
-  const status = error.status;
+  // Mid-stream SSE `error` events surface as an `APIError` with no HTTP status;
+  // the old router defaulted those to 500, so mirror that here.
+  const status = error.status ?? 500;
   switch (status) {
     case 400:
     case 422:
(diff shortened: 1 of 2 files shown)


aubin-tchoi (approved):
LGTM

front/lib/model_constructors/sdk/anthropic_ai/converters/output/utils.ts
```diff
@@ -493,17 +497,129 @@ export function streamErrorToErrorEvent(
...
+// retryability-equivalent type (network_error / invalid_request_error).
+function bareStreamErrorToErrorEvent(
+  metadata: EndpointMetadata,
+  error: unknown
+): ErrorEvent {
+  const message = normalizeError(error).message;
+  const lower = message.toLowerCase();
+
+  const build = (type: ErrorType, text: string): ErrorEvent =>
+    buildErrorEvent({ metadata, type, message: text, originalError: error });
```
aubin-tchoi, line 523: nit: would inline it

front/lib/model_constructors/sdk/anthropic_ai/converters/output/utils.ts
```diff
@@ -493,17 +497,129 @@ export function streamErrorToErrorEvent(
...
+// here verbatim for migration parity (see CODING_RULES GEN1); revisit once the
+// old router is fully retired. The new `ErrorType` union has no
+// `terminated_error`/`context_length_exceeded`, so those map to the closest
+// retryability-equivalent type (network_error / invalid_request_error).
+function bareStreamErrorToErrorEvent(
+  metadata: EndpointMetadata,
+  error: unknown
+): ErrorEvent {
+  const message = normalizeError(error).message;
+  const lower = message.toLowerCase();
```
davidebbo, line 520: Pretty horrible, but maybe that's the best we can do! :)

PR #28256, opened 2026-06-30
feat: bubble up ask_user_question for sub-agents

adrsimon:
## Description

When an agent spawns a sub-agent through the `run_agent` tool, the sub-agent's blocking events (tool approvals, personal/file auth) already bubble up to the parent conversation so the user can respond there. `ask_user_question` was the one blocking event left out of that chain — a sub-agent's question never surfaced in the parent, so the user couldn't answer and the run stalled.

This makes a sub-agent's question behave like a sub-agent's file-authorization request:

- collect `tool_ask_user_question` in the `run_agent` blocking-event loop and route it to the parent channel via `pubsubMessageId` (`getEventMessageChannelId`)
- stamp `isLastBlockingEventForStep` on the deferred question event in `publish_deferred_events` — auth events already carried it, questions didn't, so the parent step never finalized
- on answer, resume both runs through the existing FE `retryHandler` (`retryHandlerWithResetState`), the same two-call path file authorization uses: the child retry is a no-op once the answer is registered (the action is no longer blocked), and the parent's `blocked_child_action_input_required` retry relaunches the parent run. No `resume_ancestor_conversations` flag on the answer endpoints.
- submit the answer to the child conversation/message ids carried by the blocked action rather than the parent props (backward compatible: in the non-bubbled case those equal the current conversation)

## Tests

- `UserAnswerRequired.test.tsx` green (10/10) with the new `retryHandler` wiring.
- Manual end-to-end via the hive app: agent A (`run_agent` → agent B with `ask_user_question`) — question renders in A's conversation, answering targets the child id, both B and A resume to completion, and the pending question survives a parent reload.
- Regression: direct (non-sub-agent) `ask_user_question` and sub-agent approvals/file-auth unaffected.

## Risk

Low. The answer endpoints' request schema is unchanged. The resume is driven by the same FE retry handler that already powers sub-agent file authorization, so no new backend coordination path is introduced. Worst case the question doesn't bubble (current behavior). Safe to rollback.

## Deploy Plan

- [ ] Deploy the SDK (`@dust-tt/client` gains the optional `isLastBlockingEventForStep` field) and front-api before/with front so the bubbled event carries the field the parent step relies on.

### DIFF
--- a/front/components/assistant/conversation/BlockedAction.tsx
+++ b/front/components/assistant/conversation/BlockedAction.tsx
@@ -11,7 +11,6 @@ interface BlockedActionProps {
   triggeringUser: UserType | null;
   owner: LightWorkspaceType;
   conversationId: string;
-  messageId: string;
   retryHandler: (params: {
     conversationId: string;
     messageId: string;
@@ -23,7 +22,6 @@ export function BlockedAction({
   triggeringUser,
   owner,
   conversationId,
-  messageId,
   retryHandler,
 }: BlockedActionProps) {
   switch (blockedAction.status) {
@@ -72,8 +70,12 @@ export function BlockedAction({
           blockedAction={blockedAction}
           triggeringUser={triggeringUser}
           owner={owner}
-          conversationId={conversationId}
-          messageId={messageId}
+          retryHandler={() =>
+            retryHandler({
+              conversationId: blockedAction.conversationId,
+              messageId: blockedAction.messageId,
+            })
+          }
         />
       );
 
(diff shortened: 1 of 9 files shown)


front/components/assistant/conversation/BlockedAction.tsx
```diff
@@ -23,7 +22,6 @@ export function BlockedAction({
   triggeringUser,
   owner,
   conversationId,
-  messageId,
```
aubin-tchoi, line 26: seems odd that we drop the message ID but not the convo ID

aubin-tchoi: actually comes from `PodTasksCreateValidationDetails` that uses the convo ID, so unrelated to this PR

aubin-tchoi (approved):
LGTM

PR #28257, opened 2026-06-30
Archive the active plan when forking a conversation

philipperolet:
## Description

Forking a conversation copies the source conversation's file system into the new branch, including the active Plan Mode `plan.md`. As a result, a freshly branched conversation shows the parent's todo card, which is stale and confusing (reported by Victor in #shipped).

Instead of dropping the plan, this archives it on the fork: after the mount copy, we call the existing `closeActivePlan` on the new conversation. This moves the copied `plan.md` into `archived_plans/` (the same behavior as closing/trashing a plan), so the branch starts with no active todo card while the plan history is preserved (the source's `archived_plans/` are still copied as before).

Change is a single file: the conversation fork activity calls `closeActivePlan(auth, dest.toJSON())` after the GCS mount copy. A failure is logged and the fork is unblocked anyway (consistent with the surrounding copy-failure handling). `closeActivePlan` is idempotent (no-op when there is no active plan), so forks without a plan are unaffected.

## Tests

- No new test added: `closeActivePlan` is already covered in `front/lib/api/assistant/plan_mode.test.ts` (including the idempotent no-active-plan path), and there is no existing harness for the fork temporal activity.
- Local validation was not run in the Computer; validation is delegated to PR CI.

## Risk

Low. Single, additive call scoped to the fork activity. No schema or API changes. `closeActivePlan` is idempotent and already used by the close-plan tool and HTTP endpoint. Easily reversible by removing the call.

## Deploy Plan

Standard deploy. No migration or special ordering required.

### DIFF
--- a/front/lib/api/actions/servers/plan_mode/metadata.ts
+++ b/front/lib/api/actions/servers/plan_mode/metadata.ts
@@ -21,6 +21,9 @@ export const CLOSE_PLAN_TOOL_NAME = "close_plan" as const;
 
 export const PLAN_FILE_NAME = "plan.md" as const;
 
+// Directory under the conversation file system where closed plans are archived.
+export const ARCHIVED_PLANS_DIR_NAME = "archived_plans" as const;
+
 export const PLAN_MODE_TOOLS_METADATA = createToolsRecord({
   create_plan: {
     description:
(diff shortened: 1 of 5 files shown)


philipperolet:
closing, actually not needed

PR #28270, opened 2026-06-30
[viz] Send functionIdOrSlug to match front callFunction RPC schema

flvndvd:
## Description

The viz `callFunction` RPC posted functionId while front validates incoming RPC messages against a zod schema that requires `functionIdOrSlug` since #28258. The parent gates every `postMessage` through `isVisualizationRPCRequest`, so the mismatched param made the schema parse fail, the message got silently dropped, and `createSandboxFunctionInvocation` never ran.


fontanierh (approved):
lgtm

spolu (commented):
My bad! Thanks!

PR #28273, opened 2026-06-30
[dsbx] Set NODE_PATH when running functions so zod resolves at invocation

flvndvd:
## Description

Testing pod function invocation I got:
```
{"eventId":"1782849201661-0","data":{"type":"sandbox_function_invocation_result","created":1782849201660,"invocationId":"sfi_nWqafL1MPG","functionId":"sfn_q03Y14OQ9a","result":{"error":{"kind":"import_failed","message":"ResolveMessage: Cannot find package 'zod' from '/sandbox-functions/pods/vlt_nWqafL1MP7oRW/list-comments.ts'","stack":"Error: ResolveMessage: Cannot find package 'zod' from '/sandbox-functions/pods/vlt_nWqafL1MP7oRW/list-comments.ts'\n at fail (/tmp/dsbx-functions-runner-CoF82j.js:14492:49)\n at async runHandler (/tmp/dsbx-functions-runner-CoF82j.js:14517:27)"},"ok":false}}}
```

Published sandbox functions keep zod and the rest of the sandbox harness as external imports rather than inlining them, so the bundle stays small and shares a single zod instance with the runner. Resolving those externals at runtime relies on `NODE_PATH` pointing at the image's global npm modules. The build path set that environment variable but the run path did not, so building a function succeeded while invoking it failed with "Cannot find package 'zod'". This was confirmed directly on a sandbox as the agent user: importing a published bundle fails with no NODE_PATH and succeeds once it points at the global modules.

### DIFF
--- a/cli/dust-sandbox/Cargo.lock
+++ b/cli/dust-sandbox/Cargo.lock
@@ -313,7 +313,7 @@ checksum = "92773504d58c093f6de2459af4af33faa518c13451eb8f2b5698ed3d36e7c813"
 
 [[package]]
 name = "dust-sandbox"
-version = "0.1.30"
+version = "0.1.31"
 dependencies = [
  "anyhow",
  "base64",
--- a/cli/dust-sandbox/Cargo.toml
+++ b/cli/dust-sandbox/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "dust-sandbox"
-version = "0.1.30"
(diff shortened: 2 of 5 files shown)


flvndvd:
PMRR.

davidebbo (commented):
LGTM

fontanierh (commented):
LGTM

PR #28276, opened 2026-07-01
Add function to extract advanced models.

Fraggle:
## Description

Adds `isAdvancedModel(m: ModelConfigurationType)` and `getAdvancedModels()` to `front/lib/assistant.ts`. Both functions identify models gated by `availableIfOneOf.plansWithAdvancedModels === true` (e.g. Claude Opus 4.x variants). These helpers centralise the "advanced model" predicate so callers don't inline the `availableIfOneOf` check — a building block for the model tier enforcement logic.

## Tests

Added Vitest unit tests for both `isAdvancedModel` and `getAdvancedModels` in `assistant.test.ts`, covering the `plansWithAdvancedModels` flag, absent flag, undefined `availableIfOneOf`, and exclusion of feature-flag-only models.

## Risk

Low. Pure additive export — no existing behaviour changes. No consumers yet; nothing breaks if the functions are unused.

## Deploy Plan

Deploy front.

PR #28277, opened 2026-07-01
Add: models to store users, groups and workspaces allowed advanced models.

Fraggle:
## Description

Adds three Sequelize models — `UserAllowedAdvancedModel`, `GroupAllowedAdvancedModel`, and `WorkspaceAllowedAdvancedModel` — backed by `user_allowed_advanced_models`, `group_allowed_advanced_models`, and `workspace_allowed_advanced_models` tables. Each row stores an explicit `(workspaceId, entityId, providerId, modelId)` allowlist entry, enabling fine-grained per-user, per-group, and per-workspace control over which advanced models are accessible. All three extend `WorkspaceAwareModel` with `ON DELETE RESTRICT` FKs to `users`/`groups`/`workspaces`. Unique indexes enforce no duplicate grants per entity/model pair.

## Tests

No tests — schema-only addition with no business logic in this PR.

## Risk

Low. Pre-deploy migration is purely additive: three new tables with `CREATE INDEX CONCURRENTLY`. No existing tables or data are modified. Rollback means dropping the new tables, which is safe while they're empty.

## Deploy Plan

1. Run pre-deploy SQL migration against the front DB: `20260630160031_add_user_and_group_allowed_advanced_models.sql`
2. Deploy front.

### DIFF
--- a/front/admin/db.ts
+++ b/front/admin/db.ts
@@ -45,6 +45,11 @@ import { WebhookRequestModel } from "@app/lib/models/agent/triggers/webhook_requ
 import { WebhookRequestTriggerModel } from "@app/lib/models/agent/triggers/webhook_request_trigger";
 import { WebhookSourceModel } from "@app/lib/models/agent/triggers/webhook_source";
 import { WebhookSourcesViewModel } from "@app/lib/models/agent/triggers/webhook_sources_view";
+import {
+  GroupAllowedAdvancedModel,
+  UserAllowedAdvancedModel,
+  WorkspaceAllowedAdvancedModel,
+} from "@app/lib/models/allowed_advanced_model";
 import { DustAppSecretModel } from "@app/lib/models/dust_app_secret";
 import { ExtensionConfigurationModel } from "@app/lib/models/extension";
 import { FeatureFlagModel } from "@app/lib/models/feature_flag";
@@ -198,6 +203,9 @@ export function loadAllModels() {
     PlanModel,
     SubscriptionModel,
     ProviderCredentialModel,
+    UserAllowedAdvancedModel,
+    GroupAllowedAdvancedModel,
+    WorkspaceAllowedAdvancedModel,
     TemplateModel,
     CreditModel,
     CouponModel,
--- a/front/lib/models/allowed_advanced_model.ts
+++ b/front/lib/models/allowed_advanced_model.ts
@@ -1,0 +1,164 @@
+import { frontSequelize } from "@app/lib/resources/storage";
+import { DataTypes } from "@app/lib/resources/storage/data_types";
+import { GroupModel } from "@app/lib/resources/storage/models/groups";
+import { UserModel } from "@app/lib/resources/storage/models/user";
+import { WorkspaceAwareModel } from "@app/lib/resources/storage/wrappers/workspace_models";
+import type {
+  ModelIdType,
+  ModelProviderIdType,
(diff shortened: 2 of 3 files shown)


Fraggle:

- #28276
    - #28277 ◀
        - #28278
            - #28279
                - #28280
                    - #28281
                        - #28333
                            - #28334

spolu (commented):
Do we believe in the "Advance" term? why not juste AllowedModels? Seems more generic?

LGTM

Fraggle:
> Do we believe in the "Advance" term? why not juste AllowedModels? Seems more generic?

I re-used the existing terminology that is everywhere already (inc. plan property).
No strong opinions.

To be clear, it's about very specific models that will likely block agents you if you don't have access to them (no fallback).

spolu:
Yep I don't have the full context was just a comment triggered by seeing just this PR 👍

front/lib/models/allowed_advanced_model.ts
```diff
@@ -0,0 +1,162 @@
...
+import { GroupModel } from "@app/lib/resources/storage/models/groups";
+import { UserModel } from "@app/lib/resources/storage/models/user";
+import { WorkspaceAwareModel } from "@app/lib/resources/storage/wrappers/workspace_models";
+import type {
+  ModelIdType,
+  ModelProviderIdType,
+} from "@app/types/assistant/models/types";
+import type { CreationOptional, ForeignKey } from "sequelize";
+
+export class UserAllowedAdvancedModel extends WorkspaceAwareModel<UserAllowedAdvancedModel> {
```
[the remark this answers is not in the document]
Fraggle: wrong.

spolu (approved):

front/admin/db.ts
```diff
@@ -199,6 +204,9 @@ export function loadAllModels() {
     PlanModel,
     SubscriptionModel,
     ProviderCredentialModel,
+    UserAllowedAdvancedModel,
```
Nils-Fedrigo, line 206: Would argue for UserAllowedModel too -> we already get requests to restrict some models for users

Nils-Fedrigo (commented):
lgtm, just a nit

front/lib/models/allowed_advanced_model.ts
```diff
@@ -0,0 +1,164 @@
...
+  declare id: CreationOptional<number>;
+  declare createdAt: CreationOptional<Date>;
+  declare updatedAt: CreationOptional<Date>;
+
+  declare providerId: ModelProviderIdType;
+  declare modelId: ModelIdType;
+}
+
+WorkspaceAllowedAdvancedModel.init(
+  {
```
Nils-Fedrigo, line 132: maybe add a createdByUser field? just an idea

front/admin/db.ts
Nils-Fedrigo: (your call)

Nils-Fedrigo (approved):

front/lib/models/allowed_advanced_model.ts
Fraggle: Actually, a audit log makes total sense.
Will add (in another PR)

front/admin/db.ts
Fraggle: It changes quite a bit the scope and create other issues. I would tackle separately.

PR #28279, opened 2026-07-01
Add: api endpoints for advanced models access.

Fraggle:
## Description

Adds admin-only front-api routes for managing which advanced models are accessible, completing the API surface for the model tier access control initiative. Three new route groups are mounted under `/api/w/:wId/advanced_models/`:

- `GET /advanced_models` — returns the catalog of advanced models (from `AdvancedModelResource.getAdvancedModels()`)
- `GET/POST/DELETE /advanced_models/allowed/workspace` — workspace-level model allowlist
- `GET/POST/DELETE /advanced_models/allowed/users` — per-user model allowlist
- `GET/POST/DELETE /advanced_models/allowed/groups` — per-group model allowlist

All routes are `@ignoreswagger` and gated by `ensureIsAdmin()`. Shared Zod schemas and response types are extracted into `front/types/api/advanced_models.ts` (types previously inlined in `AdvancedModelResource` are moved there). Also adds explicit `name` fields to the unique indexes in `UserAllowedAdvancedModel` and `GroupAllowedAdvancedModel` Sequelize models.

## Tests

Added integration tests in `front-api/routes/w/[wId]/advanced_models/index.test.ts` covering the full add/list/remove lifecycle for workspace, user, and group allowed models, plus a 403 guard for non-admin callers.

## Risk

Low. All routes are admin-only and `@ignoreswagger`. No new DB tables or SQL migrations — the Sequelize index `name` additions are metadata-only and don't alter the actual DB schema at runtime. Rollback is a straightforward redeploy.

## Deploy Plan

Deploy front-api.

PR #28278, opened 2026-07-01
Add: resource for advanced models.

Fraggle:
## Description

Introduces `AdvancedModelResource` in `front/lib/resources/advanced_model_resource.ts` as the canonical layer for managing per-workspace, per-user, and per-group advanced model allowlists. Moves `isAdvancedModel` and `getAdvancedModels` from free functions in `assistant.ts` into static methods on the resource class, and adds admin-only CRUD — `addWorkspaceAllowedAdvancedModel`, `addUserAllowedAdvancedModel`, `addGroupAllowedAdvancedModel` and their `remove`/`list` counterparts. All mutations validate the target is an actual advanced model (`plansWithAdvancedModels === true`), enforce `auth.isAdmin()`, and return `Result<T, DustError<...>>`. Uses `findOrCreate` for idempotent adds.

## Tests

Full Vitest coverage in `advanced_model_resource.test.ts`: `isAdvancedModel`, `getAdvancedModels`, add/list/remove for workspace, user, and group; guard tests for non-admin callers, non-advanced models, and non-member users.

## Risk

Low. No new API routes. No schema changes in this PR — the `WorkspaceAllowedAdvancedModel`, `UserAllowedAdvancedModel`, and `GroupAllowedAdvancedModel` Sequelize models are consumed but not introduced here. If the backing tables (`workspace_allowed_advanced_models`, `user_allowed_advanced_models`, `group_allowed_advanced_models`) haven't been migrated yet, that migration must land before this PR is deployed.

## Deploy Plan

Ensure the `allowed_advanced_model` DB migration is applied first (if not already), then deploy front.

PR #28280, opened 2026-07-01
Create hooks for advanced models.

Fraggle:
## Description

Adds SWR hooks in `front/lib/swr/advanced_models.ts` for the advanced models REST endpoints introduced in #28279. Four read hooks — `useAdvancedModels`, `useUserAllowedAdvancedModels`, `useGroupAllowedAdvancedModels`, `useWorkspaceAllowedAdvancedModels` — and three mutation hook pairs — `useUserAllowedAdvancedModelMutations`, `useGroupAllowedAdvancedModelMutations`, `useWorkspaceAllowedAdvancedModelMutations` — covering POST/DELETE for each scope. All mutation hooks follow the standard `clientFetch` + `isMutating` + `sendNotification` + SWR invalidation pattern used elsewhere in the codebase.

## Tests

Tested locally against the advanced models API endpoints.

## Risk

Low. Front-only, no new API routes, no DB changes. Hooks are not yet wired to any UI component.

## Deploy Plan

Deploy front.

### DIFF
--- a/front/lib/swr/advanced_models.ts
+++ b/front/lib/swr/advanced_models.ts
@@ -1,0 +1,426 @@
+import { useSendNotification } from "@app/hooks/useNotification";
+import { clientFetch } from "@app/lib/egress/client";
+import {
+  emptyArray,
+  getErrorFromResponse,
+  useFetcher,
+  useSWRWithDefaults,
+} from "@app/lib/swr/swr";
+import type {
+  AllowedAdvancedModelType,
+  GetAdvancedModelsResponseBody,
+  GetGroupAllowedAdvancedModelsResponseBody,
+  GetUserAllowedAdvancedModelsResponseBody,
+  GetWorkspaceAllowedAdvancedModelsResponseBody,
+  GroupAllowedAdvancedModelBody,
+  GroupAllowedAdvancedModelsType,
+  UserAllowedAdvancedModelBody,
+  UserAllowedAdvancedModelsType,
+} from "@app/types/api/advanced_models";
+import { normalizeError } from "@app/types/shared/utils/error_utils";
+import type { LightWorkspaceType } from "@app/types/user";
+import { useCallback, useState } from "react";
+import type { Fetcher } from "swr";
+
+export const advancedModelsUrl = (workspaceId: string) =>
+  `/api/w/${workspaceId}/advanced_models`;
+
+export const userAllowedAdvancedModelsUrl = (workspaceId: string) =>
+  `${advancedModelsUrl(workspaceId)}/allowed/users`;
+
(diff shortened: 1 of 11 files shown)


Fraggle:

- #28276
    - #28277
        - #28278
            - #28279
                - #28280 ◀
                    - #28281
                        - #28333
                            - #28334

front/lib/swr/advanced_models.ts
```diff
@@ -0,0 +1,426 @@
...
+        });
+        return false;
+      } finally {
+        setIsMutating(false);
+      }
+    },
+    [owner.sId, mutateUserAllowedAdvancedModels, sendNotification]
+  );
+
+  const removeUserAllowedAdvancedModel = useCallback(
```
Nils-Fedrigo, line 187: worth creating a helper to avoid rewriting everything 3 times? (thank god for llms)

front/lib/swr/advanced_models.ts
```diff
@@ -0,0 +1,426 @@
...
+          const error = await getErrorFromResponse(response);
+          sendNotification({
+            type: "error",
+            title: "Failed to allow advanced model for user",
+            description: error.message,
+          });
+          return false;
+        }
+
+        await mutateUserAllowedAdvancedModels();
```
Nils-Fedrigo, line 171: nit- instead of getting the mutate from the hook get the one from swr directly

a friend put it in these words:


Mounting the read hook with disabled: true solely to extract mutate. That couples the mutation hook to the read hook's internals and creates a second (inert) SWR subscription. The more robust idiom is the global keyed mutate from useSWRConfig, invalidating by the same key builder:


const { mutate } = useSWRConfig();
// after success:
await mutate(userAllowedAdvancedModelsUrl(owner.sId));
This decouples the mutation from the read hook entirely and invalidates any mounted subscriber to that key — which is what you actually want, since the component showing the list is a different consumer than this mutation hook. It's the same revalidation semantics, just sourced without a parasitic hook mount.

Nils-Fedrigo (approved):
looks good to me

front/lib/swr/advanced_models.ts
Fraggle: >  This decouples the mutation from the read hook

Is it really ? To me, since we need to share the URL/key somewhere (here it's the same file, but not always),  we are still coupling both (read and write hooks) via the shared URL/key in the middle, it's less direct but still a indirect coupling.

I don't have a super strong opinion but at least this is consistent with our existing pattern in the codebase.

front/lib/swr/advanced_models.ts
Fraggle: It's not exactly the same, different Url, different notif, different mutation hooks.
If you don't might I prefer straightforward code :)

PR #28285, opened 2026-07-01
[front] fix: scope `peakConcurrency` by route

aubin-tchoi:
## Description

Looking at [these logs](https://app.datadoghq.eu/logs?query=%22Processed%20request%22%20service%3Afront%20%40peakConcurrency%3A%3C150%20%40peakConcurrency%3A%3E6&agg_m=%40peakConcurrency&agg_m_source=base&agg_q=%40route&agg_t=max&analyticsOptions=%5B%22bars%22%2C%22dog_classic%22%2Cnull%2Cnull%2C%22value%22%5D&clustering_pattern_field_path=message&cols=host%2Cservice%2C%40peakConcurrency&flat_group_bys=true&fromUser=true&messageDisplay=inline&refresh_mode=sliding&saved-view-id=221639&sort_m=%40peakConcurrency&sort_t=max&storage=hot&stream_sort=time%2Cdesc&top_n=50&top_o=top&viz=stream&x_missing=true&from_ts=1776327440318&to_ts=1776413840318&live=true), it looks like the peak concurrency we log is not scoped to each route, but instead shows the per-process (thus per-pod) concurrency.

This PR changes the in-flight tracking to be by HTTP method + Hono route, which is what we had before (before Hono).

This change only affects the logging.

## Tests

- Tested locally.

## Risk

- Low.

## Deploy Plan

- Deploy front.

### DIFF
--- a/front-api/middlewares/request_logger.ts
+++ b/front-api/middlewares/request_logger.ts
@@ -42,13 +42,15 @@ const SKIP_LOGGER_PATHS = new Set([
   "/api/kill",
 ]);
 
+type RouteConcurrencyState = {
+  activeRequests: number;
+  inFlightPeaks: Set<{ peak: number }>;
+};
+
 // In-flight request tracking, used to report (per request) the peak number of
-// requests processed concurrently during its lifetime. `activeRequests` is the
-// live count; each in-flight request holds a mutable `peak` that we raise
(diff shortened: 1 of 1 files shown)


id13 (approved):
LGTM

PR #28287, opened 2026-07-01
fix(llm-router): recover malformed tool-call JSON instead of failing

adrsimon:
## Description

With `use_new_llm_router`, some turns fail with `unknown_error` / "Unknown error from Anthropic: Expected ',' or '}' after property value in JSON at position N" (and similar `JSON.parse` messages). These are malformed tool-call arguments: with `eager_input_streaming` the model streams tool JSON and Anthropic no longer validates it server-side.

The new router streams via the non-beta `messages.stream()` accumulator, which lets the raw `JSON.parse` `SyntaxError` escape. `getInvalidToolJsonMessage` only matches the wrapped form, so the bare error fell through to a terminal `unknown_error` and the in-progress tool-call recovery was skipped.

Now, when a `tool_use` block is open and the abort is a JSON parse failure (`partialParse` only throws on genuinely malformed input, never on truncation), we recover the raw JSON from the block's own accumulator and emit a `tool_call` wrapping it — the same self-correction path already used for the wrapped-error case — so the agent loop re-samples instead of failing the turn.

## Tests

Added coverage in `utils.test.ts`: recovery from the accumulator on a bare parse error, and the guard that a bare parse error with no open `tool_use` block still surfaces as an error event. Full suite passes (88).

## Risk

Low. Change is scoped to the stream error path; a genuine parse abort now becomes a recoverable `tool_call` instead of a terminal error. Non-parse errors are unchanged. Easy to roll back.

## Deploy Plan

Standard.

### DIFF
--- a/front/lib/model_constructors/sdk/anthropic_ai/converters/output/utils.ts
+++ b/front/lib/model_constructors/sdk/anthropic_ai/converters/output/utils.ts
@@ -101,6 +101,38 @@ export function getInvalidToolJsonMessage(err: unknown): string | null {
   return null;
 }
 
+// The SDK surfaces a malformed tool-call JSON as a raw `JSON.parse` SyntaxError,
+// either bare or as the `cause` of an AnthropicError. `partialParse` tolerates
+// truncation and only throws on genuinely malformed input, so this is always bad
+// model output, never a transport cut.
+function isBareToolJsonParseError(err: unknown): boolean {
+  if (err instanceof SyntaxError) {
+    return true;
+  }
+  return (
+    err instanceof AnthropicError &&
+    "cause" in err &&
+    err.cause instanceof SyntaxError
+  );
+}
+
+// The raw malformed tool JSON when a stream aborted on a tool-call parse
+// failure, or null if unrelated. When the SDK embeds it in the error message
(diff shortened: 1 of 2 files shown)


front/lib/model_constructors/sdk/anthropic_ai/converters/output/utils.ts
```diff
@@ -101,6 +101,38 @@ export function getInvalidToolJsonMessage(err: unknown): string | null {
...
+// either bare or as the `cause` of an AnthropicError. `partialParse` tolerates
+// truncation and only throws on genuinely malformed input, so this is always bad
+// model output, never a transport cut.
+function isBareToolJsonParseError(err: unknown): boolean {
+  if (err instanceof SyntaxError) {
+    return true;
+  }
+  return (
+    err instanceof AnthropicError &&
+    "cause" in err &&
```
davidebbo, line 114: Maybe show an example of what the full error looks like in a comment

davidebbo (approved):
LGTM

front/lib/model_constructors/sdk/anthropic_ai/converters/output/utils.ts
adrsimon: the tests have some examples

PR #28290, opened 2026-07-01
Re-use the new plan field for advanced models + increase the cache version

fabiencelier:
## Description

 - re-revert https://github.com/dust-tt/dust/pull/28272
 - increase the subscription cache version so we always have the new field availalbe

## Risk

may break the same customers if cache was not the reason.
I will monitor

## Deploy Plan

deploy front

### DIFF
--- a/front/types/shared/cache_resource_registry.ts
+++ b/front/types/shared/cache_resource_registry.ts
@@ -44,7 +44,7 @@ export function buildCacheKeyPattern(
 }
 
 export const WORKSPACE_CACHE_KEY_VERSION = 2;
-export const SUBSCRIPTION_CACHE_KEY_VERSION = 1;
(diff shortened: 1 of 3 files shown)


davidebbo (approved):
LGTM

front/types/shared/cache_resource_registry.ts
```diff
@@ -44,7 +44,7 @@ export function buildCacheKeyPattern(
 }
 
 export const WORKSPACE_CACHE_KEY_VERSION = 2;
-export const SUBSCRIPTION_CACHE_KEY_VERSION = 1;
+export const SUBSCRIPTION_CACHE_KEY_VERSION = 2;
```
davidebbo, line 49: Would be good to have comment here explaining when it should be bumped. Might help models get it right.

PR #28292, opened 2026-07-01
Run sandbox functions outside gcsfuse cwd

fontanierh:
## Description

Follow up of #28123.

Sandbox function runs were launching Bun with `$DUST_FUNCTIONS_DIR` as cwd. On pod sandboxes that directory is gcsfuse-backed, and the smoke test showed the same mounted handler takes about 1.0-1.4s when cwd is the mount vs 0.2-0.3s from `/home/agent`.

This keeps importing bundles from `$DUST_FUNCTIONS_DIR`, but launches Bun from `/home/agent` by default. `DUST_FUNCTION_WORKING_DIR` can override it for local runs or future experiments. It also bumps `dsbx` to 0.1.32 and `dust-base` to 0.8.45 so the fix can be released.

## Tests

- Tested locally with `cargo test --all`, `cargo clippy --all-targets --all-features -- -D warnings`, and `bun test` in `cli/dust-sandbox/functions-runner`.
- Local `dsbx function run` smoke with `DUST_FUNCTION_WORKING_DIR=/tmp`; output confirmed `cwd` moved while `import.meta.dir` stayed at the bundle dir.
- E2B smoke on `dust-base_0-8-44`: same handler on gcsfuse, cwd on mount vs `/home/agent`; output confirmed `cwd` moved while `import.meta.dir` stayed on the mount.
- Tried the front registry vitest locally, but Hive test DB services could not start because Git fetch hit case-conflicting remote refs on this case-insensitive filesystem. CI should cover it.

## Risk

Risk is low but not zero: function code that implicitly used `process.cwd()` as the bundle directory will now see `/home/agent`. The handler still imports from the mount, and `import.meta.dir` is unchanged. Safe to rollback by restoring the previous `dsbx` and `dust-base` versions.

## Deploy Plan

- [ ] Merge this PR.
- [ ] Run the `Release dsbx CLI` workflow from main.
- [ ] Run the `Sandbox Image Registry` workflow to build `dust-base` 0.8.45 in both regions.
- [ ] Deploy front so new pod sandboxes use `dust-base_0-8-45`.
- [ ] After front deploy, open `/poke/kill` and kill older `dust-base` versions so existing sandboxes cycle onto `dust-base_0-8-45`.

### DIFF
--- a/cli/dust-sandbox/Cargo.lock
+++ b/cli/dust-sandbox/Cargo.lock
@@ -313,7 +313,7 @@ checksum = "92773504d58c093f6de2459af4af33faa518c13451eb8f2b5698ed3d36e7c813"
 
 [[package]]
 name = "dust-sandbox"
-version = "0.1.31"
+version = "0.1.32"
 dependencies = [
  "anyhow",
  "base64",
--- a/cli/dust-sandbox/Cargo.toml
+++ b/cli/dust-sandbox/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "dust-sandbox"
-version = "0.1.31"
+version = "0.1.32"
 edition = "2021"
 
 [[bin]]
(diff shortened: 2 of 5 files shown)


davidebbo (approved):
LGTM

flvndvd:
🔥 🔥

PR #28295, opened 2026-07-01
feat(analytics): add data export panel to the analytics page

rfrenoy:
## Description

Add an export panel at the bottom of the analytics page that lets admins download any of the analytics tables (usage metrics, active users, source, agents, users, skills, skill/tool usage, messages, feedback) as CSV. It calls the new session-authed /api/w/{wId}/analytics/export route via the shared useDownloadCsv hook, restoring the export capability lost when the legacy analytics page was removed.

The panel has its own dedicated period selector, independent from the page-level time range selector, so changing one does not affect the other.

https://github.com/user-attachments/assets/d90e7e1b-ab31-48db-b614-b4e3ade82be9

## Tests

Local

## Risk

Low
## Deploy Plan

Front


fabiencelier (approved):
LGTM
Thank you

PR #28296, opened 2026-07-01
Keep sandbox functions active for reaper

fontanierh:
## Description

Follow up of #28123 and #28292. Function invocations now refresh `sandboxes.lastActivityAt` after pod sandbox setup, so pod sandboxes running functions stay visible to the Reaper through the existing activity signal.

> [!NOTE]
> This keeps current short function invocations visible to the Reaper through the existing `lastActivityAt` signal. It does not yet model active long-running invocations; that will need either a heartbeat or Reaper-side active invocation state once result persistence lands.

## Tests

Tested locally:
- `npx biome check --write --error-on-warnings front/lib/resources/sandbox_function_resource.ts front/lib/resources/sandbox_function_resource.test.ts`
- `NODE_ENV=test npm run test -- lib/resources/sandbox_function_resource.test.ts` from `front`
- `npm run tsgo -- --noEmit` from `front`

## Risk

Worst case, one extra `lastActivityAt` write on sandbox function invocation start. Safe to rollback.

## Deploy Plan

Standard deploy.


flvndvd (commented):
LGTM

PR #28302, opened 2026-07-01
Analytics export: return all editors of the agent in analytics

fabiencelier:
## Description

Adds an `editorEmails` field to the agents analytics export listing everyone who authored any version of an agent (sourced like lastAuthors). authorEmails is unchanged, and it should be deprecated then removed

  - JSON: editorEmails is a string array.
  - CSV: rendered as a `; `-joined cell (via `toAgentExportCsvRow`).

## Tests

added unit tests
tested locally

## Risk

low, may slow down the query but seems OK

## Deploy Plan

deploy front

### DIFF
--- a/front/lib/api/analytics/agents_export.ts
+++ b/front/lib/api/analytics/agents_export.ts
@@ -136,7 +136,8 @@ export async function fetchAgentExportRows(
     ])
   );
 
-  const scopeFilter = includeHiddenAgents ? "" : `AND ac."scope" != 'hidden'`;
+  const scopeFilter = (alias: string) =>
+    includeHiddenAgents ? "" : `AND ${alias}."scope" != 'hidden'`;
 
   // TODO(BACK5): Migrate to AgentConfigurationResource when a suitable method exists.
   const readReplica = getFrontReplicaDbConnection();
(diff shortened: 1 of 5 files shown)


front/lib/api/analytics/agents_export.ts
```diff
@@ -137,12 +152,24 @@ export async function fetchAgentExportRows(
...
+      -- agent (matching how "lastAuthors" is sourced in recent_authors.ts:
+      -- distinct authorIds across all versions, not just the active one).
+      LEFT JOIN (
+        SELECT av."sId",
+               ARRAY_AGG(DISTINCT edt."email") AS "editorEmails"
+        FROM "agent_configurations" av
+          JOIN "users" edt ON av."authorId" = edt."id"
+        WHERE av."workspaceId" = :wId
+        GROUP BY av."sId"
+      ) editors ON editors."sId" = ac."sId"
```
fabiencelier, line 184: @id13 @flvndvd are you ok with this new part of the query performance wise ?
Any insight to make it better ?

agent configurations with versions for one workspace would be a few thousands rows so it's not small but not gigantic either 🤷

id13: to know what needs to be optimized, you need to run a few `EXPLAIN(ANALYZE, BUFFERS) SELECT ...` on the read replica (very important) on a few workspaces and see what the explain look like, are the plans consistent, where's the bottleneck etc.

I would create a script for that, very easy with claude, and inspect some by hand and ask claude to check the consistency.

Worth doing if this is in a synchronous codepath (e.g I click, the browser waits for the answer => I only have 2 minutes so perf matters a bit vs it's async and I don't care if it takes 10 minutes)

rfrenoy (approved):

PR #28304, opened 2026-07-01
Expose linked file ids in path metadata

fontanierh:
## Description

Follow-up of #28946, which is merged.

Frame paths need their linked FileResource id before the interactive side panel can render them. Return the id from the existing HEAD path endpoint by looking up the canonical mount path.

Stack: 1 of 3. Next: #28894.

## Tests

Added endpoint coverage for returning the linked file id while preserving existing HEAD metadata behavior.

## Risk

Low. Existing HEAD metadata remains unchanged except for the additional header when a linked resource exists.

## Deploy Plan

Deploy normally.

### DIFF
--- a/front-api/routes/w/[wId]/files/path/[...canonicalPath].test.ts
+++ b/front-api/routes/w/[wId]/files/path/[...canonicalPath].test.ts
@@ -3,6 +3,8 @@ import { getPrivateUploadBucket } from "@app/lib/file_storage";
 import { FileResource } from "@app/lib/resources/file_resource";
 import { FileFactory } from "@app/tests/utils/FileFactory";
 import { createPrivateApiMockRequest } from "@app/tests/utils/generic_private_api_tests";
+import { fileStorageMock } from "@app/tests/utils/mocks/file_storage";
+import { DUST_FILE_ID_HEADER, frameContentType } from "@app/types/files";
 import { honoApp } from "@front-api/app";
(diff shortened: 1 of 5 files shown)


front-api/routes/w/[wId]/files/path/[...canonicalPath].test.ts
```diff
@@ -36,8 +37,11 @@ function makeBucket(
   const fileDelete =
     overrides.fileDelete ?? vi.fn().mockResolvedValue(undefined);
   const save = overrides.save ?? vi.fn().mockResolvedValue(undefined);
+  const bucket: ReturnType<typeof getPrivateUploadBucket> = Object.create(
+    FileStorage.prototype
+  );
 
-  return {
+  return Object.assign(bucket, {
```
Comment (flvndvd) [inline front-api/routes/w/[wId]/files/path/[...canonicalPath].test.ts]: Hum not a big fan of using `Object.assign` here. I shipped a global mock recently, you should be able to use it properly.

flvndvd (approved):
LGTM

front-api/routes/w/[wId]/files/path/[...canonicalPath].test.ts
Comment (fontanierh) [reply front-api/routes/w/[wId]/files/path/[...canonicalPath].test.ts]: Updated the endpoint tests to use the global `fileStorageMock` and removed the custom `Object.assign`/`FileStorage` construction. I also added the missing read-stream tracking and empty prefix-list behavior to the shared mock.

PR #28311, opened 2026-07-01
[front] contract.start: swap based on shadow-contract, not current sub billing

tdraier:
The contract.start fallback (no pending subscription staged, e.g. an immediate
switch) refused to swap whenever the current subscription wasn't Metronome-only
billed. That blocked legitimate switches of shadow-billed (Stripe + Metronome)
workspaces onto a new contract.

Decide from the STARTED contract instead: only skip when it is a shadow
contract — no `customer_billing_provider_configuration` (Stripe owns billing)
and the workspace is still Stripe-billed. Any contract that delivers its own
invoices takes over the subscription, whatever the current sub's rail. The
Stripe-billed guard keeps free / Metronome-only contracts (which also lack a
delivery config) from being mistaken for shadows.

Co-Authored-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>

PR #28310, opened 2026-07-01
[front] swapMetronomeContract: new sub is Metronome-only (no Stripe id)

tdraier:
`swapMetronomeContract` copied the current sub's `stripeSubscriptionId` onto
the new subscription. Since the new sub is billed by its Metronome contract,
that produced an invalid credit-priced (CP_*) subscription carrying a Stripe
subscription id — a shadow-billed state — when swapping from a shadow-billed
(Stripe + Metronome) subscription. Set it to null.

Co-Authored-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>

PR #28134, opened 2026-06-29
[front] Migrate legacy Pro plans (monthly + yearly) to Business (new pricing)

tdraier:
## Description

Migrates legacy Pro workspaces (any `PRO_*` plan, **monthly or yearly**) to the new credit-priced Business plan, and adds the user-facing **cancel / resume** flow for the migration window. Dry-run-by-default script + supporting lib/API/UI.

Depends on **#28393** (adds the `promoteNoneSeatsTo` switch-contract field) — now **merged into `main`**, along with the earlier billing-path fixes (#28310, #28311, #28313).

### Per workspace (batch script)
1. **Migration date** inside the rollout window (default 2026-07-23 → 2026-08-23), floored to the hour:
   - **monthly** → the workspace's own monthly renewal boundary in the window (staggered per workspace);
   - **yearly** → **all yearly plans end at the cutover** (fixed at the window **start** date, at the subscription's billing-anchor hour so they spread across the day). Active yearly subs migrate to Business; already-cancelled yearly subs ending *after* the cutover have their Stripe cancellation pulled in to it (no Business contract — they opted out); those ending on/before it are left alone. Cut-over yearly subs are **refunded** the unused prepaid time (prorated on remaining days) by the `customer.subscription.deleted` webhook.
2. **Plan selection**: `PRO_PLAN_SEAT_39` (the enterprise seat-based legacy plan) **always** migrates to `CP_BUSINESS_LEGACY_LARGE_PLAN`, bypassing the fit check. Every other `PRO_*` plan uses `checkWorkspaceFitsPlanLimits`: standard `CP_BUSINESS_PLAN` when the workspace fits its limits (seats / spaces / data sources); otherwise `CP_BUSINESS_LEGACY_LARGE_PLAN` (limits ≥ any `PRO_*` plan). Skips + logs if it exceeds even the large plan.
3. **Default payment method** (`ensureStripeCustomerDefaultPaymentMethod`): promote the subscription's card (or the customer's most recent card) to the customer default — Metronome bills the customer default, which legacy subscription-billed customers often don't have set.
4. **Switch** via `switchContract` (future-dated contract → pending subscription + Stripe cancellation scheduled at the migration date), forcing seat-less members onto a paid seat via `promoteNoneSeatsTo` — `pro` (monthly) / `pro_yearly` (yearly).
[...]

PR #28312, opened 2026-07-01
[front] contract.start: create subscription when none active (webhook ordering)

tdraier:
The stripe `customer.subscription.deleted` and metronome `contract.start`
webhooks arrive in random order. When `deleted` is processed first it ends the
active subscription (and schedules a scrub), so `contract.start` then found no
active subscription and bailed — leaving the workspace with no subscription and
a pending scrub. This is also the normal "cancel then re-register" flow.

`contract.start` now tolerates a missing active subscription: it swaps the
active sub onto the new contract when present, otherwise creates a new active
subscription (SubscriptionResource.createActiveMetronomeSubscription). Either
path then runs restoreWorkspaceAfterSubscription, which cancels the scheduled
scrub. previousPlanCode falls back to FREE_NO_PLAN_CODE for the audit event.


tdraier:

- #28310
    - #28311
        - #28312 ◀
            - #28313
                - #28134

thervier (commented):
As discussed not needed right?

PR #28313, opened 2026-07-01
[front] Seat mid-period new members on both current and pending contracts

tdraier:
## Description

When a workspace has a scheduled future contract switch (e.g. a legacy→Business migration in its window), a member added or changed **mid-window** must be seated on **both** contracts, with the correct seat on each:

- the **current** contract with the seat they actually take now (e.g. a legacy `workspace`/`none` seat) — the current contract still bills / finalizes a proration invoice for the added seat; and
- the **pending** contract with the seat that will bill them once it starts (legacy `workspace`/`none` → `pro` on Business), so its first invoice is correct.

The seat invoice is cut at contract start, so the pending contract's seats must be staged **before** it starts — reconciling in the `contract.start` webhook is too late.

### Changes
- **`resolveSeatTypeForNewMembership`** (`membership.ts`): resolves the active seat against the **active** contract only (the plan billing the member right now). A pending switch does not change the active seat.
- **`syncMetronomeSeatCountForWorkspace`** (`seat_sync.ts`): keeps reconciling the **active** contract exactly as before, and additionally **stages the pending contract** when one exists (`stagePendingContractSeats`): schedules a future-dated seat change per member (`remapMembershipSeatTypesForContract`, `swapAt: "next-hour"`, `startingAt` = pending start) and pre-provisions its seat counts (`syncSeatCount` at the pending start). Best-effort (logged, self-heals on the next membership change); skips the live-balance credit reconcile since the contract isn't active yet. Every membership change already routes through this fn, so the pending contract stays correct across adds, revokes, and seat changes.
- **`remapMembershipSeatTypesForContract` / `promoteNoneSeatTypesForContract`** (`seats.ts`): gain a `forceSeatType` (surfaced as `promoteNoneSeatType` on the remap wrapper) that, when set and billed by the contract, forces **every** `none` member onto it — preempting the committed-spare promotion. `stagePendingContractSeats` passes `"pro"` for the current legacy→Business migration (Business has no committed seats, so behavior is unchanged there).

The scheduled future rows are provisioned to Metronome by `syncSeatCount`, which reads active + scheduled-future memberships and writes each future segment with its own `starting_at` (Metronome flips it on the date; no scheduler on our side).

## Tests

Manual: on a workspace with an active legacy contract + a pending Business contract, adding a member (including via the invitation flow) creates the active `none`/`workspace` row on the current contract **plus** a scheduled `pro` row on the pending contract, and both contracts sync (the mid-window sync runs via the debounced `syncMetronomeSeatCountWorkflow`). `lib/metronome/seats.test.ts` green; `tsgo` clean on changed files.

## Risk

Low. Additive to the seat-sync path — the active-contract reconcile is unchanged. Pending-contract staging is best-effort and idempotent, and only runs when a pending contract with a seat subscription exists.

## Deploy Plan

No migration or config change. Base is `main`; also the base of the migration PR (#28134).

### DIFF
--- a/front/lib/api/membership.ts
+++ b/front/lib/api/membership.ts
@@ -95,6 +95,12 @@ async function resolveSeatTypeForNewMembership(
   if (!workspace.metronomeCustomerId) {
     return "none";
   }
+  // The new member's active seat is always resolved against the workspace's
+  // ACTIVE contract — the plan billing them right now. A pending contract switch
+  // (e.g. a legacy→Business migration in its window) does NOT change this: the
+  // member takes the current contract's seat now, and is separately scheduled
+  // onto the pending contract at its start by the seat sync (see
+  // `syncMetronomeSeatCountForWorkspace`).
   const subscription = await SubscriptionResource.fetchActiveByWorkspaceModelId(
     workspace.id
   );
--- a/front/lib/api/metronome/seat_sync.ts
+++ b/front/lib/api/metronome/seat_sync.ts
@@ -4,9 +4,11 @@ import {
 } from "@app/lib/api/metronome/reconcile_credit_state";
 import { syncDefaultPoolCapAlertsForWorkspace } from "@app/lib/api/workspace/default_user_spend_limit";
 import { Authenticator } from "@app/lib/auth";
+import { getMetronomeContractById } from "@app/lib/metronome/client";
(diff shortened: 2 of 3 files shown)


tdraier:

- #28313 ◀
    - #28134

rfrenoy (approved):

PR #28314, opened 2026-07-01
Drop legacy agent_step_contents type check constraint

flvndvd:
## Description

The database has a stale check constraint on the type column of agent step contents that only permits `text_content`, `reasoning`, `function_call`, and `error`. When `provider_passthrough` was later introduced, no migration updated this constraint, so any attempt to persist a `provider_passthrough` step content fails at insert time with a check constraint violation. The constraint exists in both EU and US.

This drops the constraint entirely rather than extending it. It's extremely confusing as this constraint does not live anywhere in the code. 

The migration is hand-written and idempotent because the constraint was never expressed in the model and is therefore not something the schema diff tooling manages.

### DIFF
--- a/front/migrations/pre-deploy/20260701133528_drop_agent_step_contents_type_check.sql
+++ b/front/migrations/pre-deploy/20260701133528_drop_agent_step_contents_type_check.sql
@@ -1,0 +1,15 @@
+/*
+Drop the legacy CHECK constraint on agent_step_contents.type.
+
+The constraint only allowed ('text_content','reasoning','function_call','error')
+and was never updated when 'provider_passthrough' was added (PR #28129), causing
+inserts of provider_passthrough rows to fail. Allowed values are enforced in code
+via the Sequelize `isIn` validate on AgentStepContentModel, so the DB-level check
+is redundant. This constraint is not generated from the model, hence this
+hand-written migration rather than a pg-schema-diff one.
+
+Statement 0
+*/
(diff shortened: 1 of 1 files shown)


davidebbo (approved):
LGTM

front/migrations/pre-deploy/20260701133528_drop_agent_step_contents_type_check.sql
```diff
@@ -0,0 +1,15 @@
...
+inserts of provider_passthrough rows to fail. Allowed values are enforced in code
+via the Sequelize `isIn` validate on AgentStepContentModel, so the DB-level check
+is redundant. This constraint is not generated from the model, hence this
+hand-written migration rather than a pg-schema-diff one.
+
+Statement 0
+*/
+SET SESSION statement_timeout = 3000;
+SET SESSION lock_timeout = 3000;
+ALTER TABLE "public"."agent_step_contents" DROP CONSTRAINT IF EXISTS "agent_step_contents_type_check";
```
davidebbo, line 15: But don't you need to create a new one with provider_passthrough added to the list?

davidebbo: Ah ok, saw PR descr.

aubin-tchoi (approved):
LGTM
OOC do we know if the script generates them automatically? we should maybe remove it in the Sequelize model init

PR #28317, opened 2026-07-01
fix(sparkle): remove bullet marker on markdown task list items

Inesvvv:
## Description
 
Fixes a rendering bug in the Sparkle markdown renderer where task lists displayed **both** a bullet point and a checkbox.

`remark-gfm` renders task items as `<li class="task-list-item">` containing a checkbox, but `UlBlock` applies `list-disc` to every `<ul>`, so the browser drew its default bullet on top of the checkbox. The fix detects the `task-list-item` class in `LiBlock` and applies `list-none` to those items only, suppressing the bullet while leaving regular bulleted/numbered lists untouched.

   **Before**

   - • ☐ to do
   - • ☑ done

   **After**

   - ☐ to do
   - ☑ done

## Tests

Verified manually in Storybook 
Screenshot 2026-07-01 at 14 04 30
--> no longer see the bullet points in front

## Risk

 Very low, modifies just `List.tsx`

## Deploy Plan

### DIFF
--- a/sparkle/src/components/markdown/List.tsx
+++ b/sparkle/src/components/markdown/List.tsx
@@ -64,8 +64,17 @@ export const LiBlock = memo(
   ({ children, className }: LiBlockProps) => {
     const { textColor, forcedTextSize } = useMarkdownStyle();
     const textSize = forcedTextSize ?? markdownParagraphSize;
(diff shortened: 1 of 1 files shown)


PopDaph (approved):
LGTM, ok for you @ykmsd? 🙏🏻

PR #28321, opened 2026-07-01
Add call tool to the sandbox_functions MCP server

PopDaph:
## Description

Pod functions can be published to a Pod, but an agent had no native way to call one. S9 fixes that: `call` is a 4th tool on the existing `sandbox_functions` server (`list` / `get` / `publish` / `call`), so any agent in a Pod can invoke a published function and get a typed result back in the loop.

Design doc: [S9: Endpoint Discoverability by Agents](https://app.notion.com/p/dust-tt/S9-Endpoint-Discoverability-by-Agents-38b28599d9418088a984c4cb0d0cdc90). 

## What this PR does

1. Adds the `call` tool (`metadata.ts` + `tools/call.ts` + `tools/index.ts`): takes `slug` + `input`, returns the function's output.
2. Adds `callSandboxFunction` (sibling to `publish_sandbox_function.ts`): calls `SandboxFunctionResource.invoke()`, then consumes the existing invocation event stream for the result.

## Worth knowing

- **Delivery is reused, not rebuilt.** The result callback -> Redis pub/sub -> SSE path already exists on main; `callSandboxFunction` just subscribes with `lastEventId: null`, so history replay returns the result even though `invoke()` publishes it mid-exec. No polling.
- A function's response is relayed with its HTTP status. dsbx exits 0 regardless, so we read the envelope's ok, not the exit code. A non-2xx comes back as a normal result (HTTP <status> + body, like http_client); only a runner failure (import error, thrown handler, bad return) is surfaced as a tool error.
- **`input` is typed as an object (`z.record`), not `z.unknown()`.** `z.unknown()` compiles to `{}`, which the model sends as a string that double-encodes into a `req.json()` failure.

## Risk

Worst case, `call` times out or returns a structured tool error and the loop continues. No persisted state touched. Additive, behind the existing `sandbox_functions` FF. Safe to roll back.

## Deploy plan

Deploy front.

### DIFF
--- a/front/lib/api/actions/servers/sandbox_functions/metadata.ts
+++ b/front/lib/api/actions/servers/sandbox_functions/metadata.ts
@@ -72,6 +72,31 @@ export const SANDBOX_FUNCTIONS_TOOLS_METADATA = createToolsRecord({
       done: "Published sandbox function",
     },
   },
+  call: {
+    description:
+      "Call a sandbox function published in the current pod by its slug, passing its input " +
+      "payload, and get back the function's output. Use the get tool first to see the function's " +
+      "input schema. Input is validated inside the sandbox.",
+    schema: {
+      slug: z
+        .string()
+        .min(1)
+        .describe(
+          "The slug of the sandbox function to call, as shown by the list tool."
+        ),
+      input: z
+        .record(z.unknown())
+        .optional()
+        .describe(
+          "The function's input payload as a JSON object, matching its input schema (see the get tool)."
+        ),
+    },
+    stake: "low",
+    displayLabels: {
+      running: "Calling sandbox function...",
+      done: "Called sandbox function",
+    },
+  },
 });
 
 export const SANDBOX_FUNCTIONS_SERVER = {
--- a/front/lib/api/actions/servers/sandbox_functions/tools/call.ts
+++ b/front/lib/api/actions/servers/sandbox_functions/tools/call.ts
@@ -1,0 +1,60 @@
+import { MCPError } from "@app/lib/actions/mcp_errors";
+import type {
+  ToolHandlerExtra,
+  ToolHandlerResult,
+} from "@app/lib/actions/mcp_internal_actions/tool_definition";
+import { getPod } from "@app/lib/api/actions/servers/pod_manager/helpers";
+import { callSandboxFunction } from "@app/lib/api/sandbox_functions/call_sandbox_function";
+import { SandboxFunctionResource } from "@app/lib/resources/sandbox_function_resource";
+import { Err, Ok } from "@app/types/shared/result";
+
+export async function callHandler(
+  { slug, input }: { slug: string; input?: Record<string, unknown> },
+  { auth, agentLoopContext }: ToolHandlerExtra
+): Promise<ToolHandlerResult> {
+  const podResult = await getPod(auth, { agentLoopContext });
+  if (podResult.isErr()) {
+    return new Err(podResult.error);
+  }
+
+  const sandboxFunction = await SandboxFunctionResource.fetchBySpaceAndSlug(
+    auth,
+    podResult.value.pod,
+    slug
+  );
+  if (!sandboxFunction) {
+    return new Err(
+      new MCPError(`No sandbox function with slug "${slug}" in this pod.`, {
+        tracked: false,
+      })
+    );
+  }
+
+  const result = await callSandboxFunction(auth, sandboxFunction, input);
+  if (result.isErr()) {
+    return new Err(new MCPError(result.error.message));
+  }
+
+  const outcome = result.value;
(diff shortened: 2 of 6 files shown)


front/lib/api/actions/servers/sandbox_functions/metadata.ts
```diff
@@ -72,6 +72,31 @@ export const SANDBOX_FUNCTIONS_TOOLS_METADATA = createToolsRecord({
       done: "Published sandbox function",
     },
   },
+  call: {
+    description:
+      "Call a sandbox function published in the current pod by its slug, passing its input " +
+      "payload, and get back the function's output. Use the get tool first to see the function's " +
+      "input schema. Input is validated inside the sandbox.",
+    schema: {
+      slug: z
```
spolu, line 81: This include podId? Should we separate it?

front/lib/api/actions/servers/sandbox_functions/tools/call.ts
```diff
@@ -0,0 +1,60 @@
...
+import { callSandboxFunction } from "@app/lib/api/sandbox_functions/call_sandbox_function";
+import { SandboxFunctionResource } from "@app/lib/resources/sandbox_function_resource";
+import { Err, Ok } from "@app/types/shared/result";
+
+export async function callHandler(
+  { slug, input }: { slug: string; input?: Record<string, unknown> },
+  { auth, agentLoopContext }: ToolHandlerExtra
+): Promise<ToolHandlerResult> {
+  const podResult = await getPod(auth, { agentLoopContext });
+  if (podResult.isErr()) {
```
spolu, line 16: So that means we can't call from outside the pod?

front/lib/api/sandbox_functions/call_sandbox_function.ts
```diff
@@ -0,0 +1,102 @@
...
+  sandboxFunction: SandboxFunctionResource,
+  input: unknown
+): Promise<Result<SandboxFunctionCallOutcome, Error>> {
+  const invocationResult = await sandboxFunction.invoke(auth, { input });
+  if (invocationResult.isErr()) {
+    return invocationResult;
+  }
+  const { sId: invocationId } = invocationResult.value;
+
+  for await (const { data } of getSandboxFunctionInvocationEvents({
```
spolu, line 66: This may get killed by deploy?

What happens then ooc?

spolu: Should we have a blocking version of the invoke to mimick what we do with regular sandbox tools? With ability to reconnect to the execution based on the existence of a tool approval?

spolu (approved):
LGTM as first functional implementation, I thiiiink we'll need to revisit aspects as we introduce tool calls right?

PopDaph:
Thanks @spolu merging already indeed and will apply all feedback in a follow up 👍🏻

front/lib/api/actions/servers/sandbox_functions/metadata.ts
PopDaph: Slug doesn't include the podId, the pod comes from the agent loop context via getPod, same as list/get/publish, so there's nothing bundled to separate. If we want to call across pods later we'd add an explicit pod param (like getPod's dustPod variant), but that's out of scope for v0 no? I'm unsure!

front/lib/api/actions/servers/sandbox_functions/tools/call.ts
PopDaph: I thought it was v0 to only have this as part of pod, was that wrong?? 🙏🏻

front/lib/api/actions/servers/sandbox_functions/metadata.ts
spolu: Can def be tackled later 👍

front/lib/api/actions/servers/sandbox_functions/tools/call.ts
spolu: Can def be tackled later 👍

front/lib/api/sandbox_functions/call_sandbox_function.ts
PopDaph: Yes I think it'll get killed by deploy, the activity is no_retry so the turn just fails and a re-entry would re-run the function 😬 

From what I could trace, the reconnect seems to ride the approval-pause flow (persist the invocationId at pause, re-subscribe to the event stream on resume), my instinct is to leave it for after S5 rather than build it blind now? Does it make sense?

front/lib/api/actions/servers/sandbox_functions/metadata.ts
spolu: And agreed on your analysis of it being tied to approval 👌

PR #28322, opened 2026-07-01
chore(front-api): remove deprecated v1 workspace-usage endpoint

rfrenoy:
## Description

The public `GET /api/v1/w/{wId}/workspace-usage` endpoint was deprecated with a removal date of 2026-07-01 in favor of `GET /api/v1/w/{wId}/analytics/export`. It has no remaining callers, so drop the route and regenerate the swagger spec.
## Deploy Plan

OpenAPI + Front


fabiencelier (approved):
🔥 🔥 🔥

PR #28324, opened 2026-07-01
[Sparkle] Refactor color token architecture, revamp Foundations stories, and migrate semantic colors in `front`

ClementAupiais:
## Description

Consolidates the Sparkle color system around a single source of truth and brings the Storybook documentation in sync with it. In `tokens.css`, all hardcoded `oklch` values in semantic and structural tokens are replaced by `var()` references to their primitive palette counterparts (e.g. `--color-highlight-500: var(--color-blue-500)`), and the file is reorganized into labeled sections (Primitive palettes → Semantic → Structural & surfaces → Brand). A new `foundations-helpers.tsx` module drives the revamped `ColorPalette`, `Typography`, `Shadows`, and `Motion` stories: token values are read live from computed CSS variables instead of a hardcoded JS map, color families are discovered from `theme.css` at runtime, and a `withThemedSurface` decorator fixes dark-mode propagation. Separately, scattered `gray-*` Tailwind classes in `front/` are migrated to their semantic equivalents (`primary-*`, `muted`, `muted-foreground`), and the loading-screen inline styles in `front-spa/index.html` are updated to match the new token values.

## Tests

Verified manually in Storybook: all 508 (token, context) pairs resolve to their original value in both light and dark themes. Toggle the theme in the toolbar to confirm semantic tokens flip correctly; hover swatches to check WCAG contrast grades.

## Risk

Low — `tokens.css` changes are color-preserving (pure reorganization + `var()` indirection). The `front/` class renames (`gray-*` → semantic) are straightforward visual-equivalent swaps. Storybook files are dev-only.

## Deploy Plan

Deploy sparkle + front.

### DIFF
--- a/front-spa/index.html
+++ b/front-spa/index.html
@@ -10,7 +10,7 @@
     <!-- PWA: web app manifest enables installation and standalone (no browser chrome) launch. -->
     <link rel="manifest" href="/manifest.webmanifest" />
     <meta name="theme-color" content="#FBFAF9" media="(prefers-color-scheme: light)" />
-    <meta name="theme-color" content="#1C222D" media="(prefers-color-scheme: dark)" />
+    <meta name="theme-color" content="#191715" media="(prefers-color-scheme: dark)" />
 
     <!-- PWA: iOS Safari standalone support ("Add to Home Screen"). -->
     <meta name="apple-mobile-web-app-capable" content="yes" />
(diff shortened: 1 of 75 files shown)


smb2268 (approved):
Did some smoke testing and ran this through fable which both found a couple small spots in dark mode where bg was flipping but text wasn't or vice versa, pushed a few small commits to fix those, for example: Screenshot 2026-07-02 at 11 49 56 PM

If we're worried about recurrance, maybe we should add a biome rule that flags:
text-white / text-black co-occurring with a bg-primary-* class in the same className
text-foreground / text-muted-foreground on bg-primary-600–950 backgrounds
Inline backgroundColor: in style={{}} in product code

Otherwise I feel good about it :)

PR #28326, opened 2026-07-01
fix(capabilities): align skill and tool avatar sizes in capabilities picker

rfrenoy:
## Description

Ticket: https://github.com/dust-tt/tasks/issues/9346

Skill icons were rendered as components, so DropdownMenuItem's renderIcon wrapped them in <Icon size="sm"> and forced h-5 w-5 (20px) onto the avatar, while tool icons were passed as already-rendered elements that ignored the injected className and stayed at 36px. Render both as elements at a consistent xs size so they align.

Screenshot 2026-07-01 at 15 06 08

## Tests

Local

## Risk

Low

### DIFF
--- a/front/components/assistant/CapabilitiesPicker.tsx
+++ b/front/components/assistant/CapabilitiesPicker.tsx
@@ -351,11 +351,13 @@ export function CapabilitiesPicker({
           continue;
         }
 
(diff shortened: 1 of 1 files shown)


ykmsd (approved):
ty!

PR #21322, opened 2026-02-09
[FIX] Preprocess Markdown For Instruction Block Tags

flvndvd:
## Description

Fix `InstructionBlockExtension` failing to parse XML-like instruction blocks (e.g. `<CRITICAL_INFORMATION>...</CRITICAL_INFORMATION>`) from markdown content in the Agent Builder editor.

### Root Cause

The `InstructionBlockExtension` was moved from `paragraph` to `block+` content type. This change meant the marked tokenizer now requires block-level separation (double newline `\n\n`) to recognize instruction blocks. 
However:
- Single `\n` before tags: Most agent instructions in the DB use a single `\n` before XML-like tags. The marked block tokenizer never fires because it treats single `\n` as a soft break within the same paragraph, not a block separator.
- `OPENING_TAG_BEGINNING_REGEX` too strict: The regex inherited a $ end-of-string anchor from `OPENING_TAG_REGEX` (used for InputRules), making `OPENING_TAG_BEGINNING_REGEX` = /^<TAG>$/i. This only matches strings that are exactly an opening tag. The `start()` function could never match when there's content after the tag.
- `escapeUnrecognizedHtmlTags` regex incomplete: The regex [A-Za-z][A-Za-z0-9]* didn't support underscores in tag names (e.g. `CRITICAL_INFORMATION`) and didn't match closing tags (`</TAG>`).

The preview mode was unaffected because it uses react-markdown with a custom `instructionBlockDirective` that does its own regex matching on the raw string. No dependency on marked's block tokenization.

### Solution

Preprocess markdown before `setContent`:
- Added logic to ensure blank lines (\n\n) around matched XML tag pairs so marked treats them as block-level tokens.
- Fixed `OPENING_TAG_BEGINNING_REGEX`: Defined independently without `$` anchor so the tokenizer's `start()` function returns the correct index.
Standalone tag handling: Unrecognized standalone tags (e.g. <URL> without a closing pair) have brackets stripped to prevent marked crashes, while matched pairs are left intact for the instruction block tokenizer.

### Related
- Incident: Agent Builder not showing instructions [thread](https://dust4ai.slack.com/archives/C05B529FHV1/p1770625713287079)

## Tests

Tested on front-edge.

PR #21254, opened 2026-02-06
Fix TipTap Range Error on Invalid Angle Bracketed Tags

flvndvd:
## Description

It resolves https://github.com/dust-tt/tasks/issues/6413.

This PR:
- fixes `RangeError: Invalid content for node type paragraph` crash when markdown contains angle-bracketed tokens
like `<URL>` that marked lexes as inline HTML
- adds `escapeUnrecognizedHtmlTags` helper that derives recognized tags from the editor's `ProseMirror` schema and
escapes unrecognized ones with HTML entities before passing to the markdown parser

### Context

There is a known tiptap bug https://github.com/ueberdosis/tiptap/issues/7256. The lib `@tiptap/markdown`'s `parseHTMLToken` calls `generateJSON` on unrecognized tags which produces block nodes that violate the inline schema. Upstream fix `https://github.com/ueberdosis/tiptap/pull/7260` not yet merged. This workaround can be removed once it is.

PR #28327, opened 2026-07-01
chore(deps): upgrade tiptap to 3.27.1 across front, extension, and sparkle

ElPicador:
## Description

Upgrades `@tiptap/*` from 3.19.0 (3.22.1 for `@tiptap/html`/`@tiptap/static-renderer`, 3.15.3 in the
sparkle playground) to 3.27.1 across `front`, `extension`, and `sparkle/playground`, and pins the
versions exactly instead of via caret ranges so all three workspaces stay in lockstep.

The main upstream bug fix this pulls in is [ueberdosis/tiptap#7256](https://github.com/ueberdosis/tiptap/issues/7256):
unrecognized angle-bracket/XML-like tags in Markdown (e.g. `<agent>`, `<rules>`, or a stray
`<enter existing CID here if available>`) were silently swallowed or could crash `setContent`
with a ProseMirror `contentMatchAt`/"Invalid content for node paragraph" error, especially with
schemas that use Mention/Variable/Table nodes. That's exactly the bug we hit and worked around
ourselves in #21254 and #21322 — `front/components/editor/lib/preprocessMarkdownForEditor.ts` has
been escaping `<` with a zero-width space before handing content to `@tiptap/markdown`, with a
`TODO: Remove when tiptap merges ueberdosis/tiptap#7260` note. Tiptap landed a proper fix upstream
(preserving unrecognized tags as literal text instead of dropping/crashing) in `@tiptap/markdown`
3.24.0. Our local workaround is left in place for now — removing it is a separate, independently
testable follow-up.

## Tests

Dependency bump only; existing editor unit tests (`InstructionSuggestionExtension.test.ts` and
friends) and the `front`/`extension`/`sparkle` test suites cover the affected editor paths.


ElPicador:

- #28327 ◀

PR #28333, opened 2026-07-01
Add: unify method to resolve allowed models.

Fraggle:
## Description

Extracts the allowed-models resolution logic into a shared pure function `resolveAllowedAdvancedModels` in `front/lib/advanced_models/resolve_allowed.ts`, and adds a server-side `AdvancedModelResource.resolveAllowedAdvancedModels` static method that fetches from the DB and delegates to it.

Previously the union/dedup logic lived inline in `resolveAdvancedModelsForUser` (client-side, pre-fetched maps). This PR:
- Moves `advancedModelKey` and the core resolution logic to a shared module, reusable by both client and server paths.
- Adds `AdvancedModelResource.resolveAllowedAdvancedModels(auth, { user?, groupIds? })` for single-call server-side resolution (parallel DB fetches via `Promise.all`), with private helpers `loadWorkspaceAllowedAdvancedModels`, `loadGroupAllowedAdvancedModelsByGroupId`, `loadUserAllowedAdvancedModels`.
- Makes `ResolvedAdvancedModelsForUser` a type alias for `ResolvedAllowedAdvancedModels`.

## Tests

Added integration tests in `advanced_model_resource.test.ts` covering: union across all three scopes, deduplication, `hasUserLevelOverride` flag, explicit `groupIds` override, and non-admin access.

## Risk

Low. Pure refactor — no behavior change to existing `resolveAdvancedModelsForUser`. New server-side method is additive. No DB schema changes.

## Deploy Plan

Deploy front.

PR #28281, opened 2026-07-01
Add: column and modal for advanced models.

Fraggle:
## Description

Wires the advanced models allowlist UI into the admin Usage page, gated on `models_picker` feature flag + admin role.

- **`MembersUsageTable`**: new `advancedModels` column showing resolved model names per member (tooltip for multi-model, `(custom)` suffix for user-level overrides via `resolveAdvancedModelsForUser`).
- **`EditAdvancedModelsModal`**: new `Dialog` with `SliderToggle` per model, scoped to `user | group | workspace`. Calls add/remove mutations from `useUser/Group/WorkspaceAllowedAdvancedModelMutations` with per-model pending state.
- **`UsagePage`**: loads catalog + allowed models from all four SWR hooks, derives `userAllowedAdvancedModelsByUserId`, `groupAdvancedModelsByGroupId`, and `groupNameToId` lookup maps, passes them to `MembersUsageTable`, renders `EditAdvancedModelsModal` for user/group/workspace targets, adds `AdvancedModelsSettingsCard` in the settings section, and exposes a group-filter button to open the group scope modal.

file-d26eed69087daa6ef3454ef89adface8

file-8fba08f178c4c2e3bd8f3754ebad6687

file-67485704efb0eb80656b3cc8f2ee13bc

file-f7cb74f2638ddaa9fc19cc40dcea01cd

file-fd6cdb4d0f148bce8eab50598797eb82

file-91047aa0ba46a5df43cdcab00043254d

Note that design wise we can def. iterate.

## Tests

Tested locally with `models_picker` feature flag enabled. Verified toggle add/remove for user, group, and workspace scopes; column display with single model, multiple models (tooltip), and empty state.

## Risk

Low. Entirely gated behind `models_picker` FF + admin check. All SWR hooks disable when the flag is off. No DB changes.

## Deploy Plan

Deploy front.

PR #28334, opened 2026-07-01
Restrict access to advanced model in the agentic loop

Fraggle:
## Description

Enforces advanced model access restrictions server-side in the agentic loop, not just the UI. At `step === 0` in `runModel`, calls `getAdvancedModelAccessErrorForAgentConfiguration` — which checks `models_picker` FF, `AdvancedModelResource.isAdvancedModel`, and the user/workspace allowlist — and short-circuits with `publishAgentError` if access is denied. Extracts the check into `front/lib/advanced_models/access.ts` with `buildAdvancedModelAccessDeniedError` producing a `GenericErrorContent` with code `advanced_model_not_enabled`.

## Tests

Added Vitest integration tests in `access.test.ts` covering: FF disabled → null, non-advanced model → null, advanced model not allowlisted → error, workspace allowlist grant → null, user allowlist grant → null.

file-5327cd53288b2169293d0176879ed1d2

## Risk

Low. Gated entirely on `models_picker` feature flag; no behaviour change for workspaces without it. Check runs only at `step === 0` to avoid redundant DB calls per loop step. Rollback is a deploy revert.

## Deploy Plan

Deploy front.

### DIFF
--- a/front/temporal/agent_loop/lib/run_model.ts
+++ b/front/temporal/agent_loop/lib/run_model.ts
@@ -7,6 +7,7 @@ import type { StepContext } from "@app/lib/actions/types";
 import type { AgentActionSpecification } from "@app/lib/actions/types/agent";
 import { isServerSideMCPServerConfigurationWithName } from "@app/lib/actions/types/guards";
 import { computeStepContexts } from "@app/lib/actions/utils";
+import { getAdvancedModelAccessErrorForAgentConfiguration } from "@app/lib/advanced_models/access";
 import { createClientSideMCPServerConfigurations } from "@app/lib/api/actions/mcp_client_side";
 import { getAgentConfigurationsForView } from "@app/lib/api/assistant/configuration/views";
 import { renderConversationForModel } from "@app/lib/api/assistant/conversation_rendering";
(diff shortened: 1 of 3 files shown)


Fraggle:

- #28276
    - #28277
        - #28278
            - #28279
                - #28280
                    - #28281
                        - #28333
                            - #28334 ◀

front/temporal/agent_loop/lib/run_model.ts
```diff
@@ -227,6 +228,23 @@ export async function runModel(
     return null;
   }
 
+  const featureFlags = await getFeatureFlags(auth);
+
+  if (step === 0) {
```
Nils-Fedrigo, line 228: so we leave access to already started convos? 
It's better to just test every step no ? Worst case they @ another agent

Nils-Fedrigo (approved):
lgtm

front/temporal/agent_loop/lib/run_model.ts
Fraggle: If they @another agent it will be a separate agentic loop.
(here step is not the first message but the first iteration of the loop)

PR #28336, opened 2026-07-01
feat(seats): auto-assign workspace/workspace_yearly committed seats

rfrenoy:
## Description

Add `workspace` and `workspace_yearly` to the AUTO_ASSIGNABLE set in getDefaultSeatTypeForContract so new members on pooled Enterprise contracts fill committed workspace seats (up to the minSeats floor) instead of always landing on "none". `max`/`max_yearly` remain manual-only.
## Tests

Unit

## Risk

Low
## Deploy Plan

Front

### DIFF
--- a/front/lib/metronome/seat_types.test.ts
+++ b/front/lib/metronome/seat_types.test.ts
@@ -173,6 +173,88 @@ describe("getDefaultSeatTypeForContract — committed seats", () => {
     ).toBe("free");
   });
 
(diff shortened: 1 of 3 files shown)


tdraier (approved):
For the same reason as in https://github.com/dust-tt/dust/pull/28347 , this should not be called on shadow contracts ( non CP plans ) - the resolveSeatTypeForNewMembership (only caller ? ) should return none for these plans

PR #28338, opened 2026-07-01
[front] metronome contracts: add displayedName + use it for free plan

thervier:
## Description

Add an optional displayed name when creating a new metronome contract + use if for free plans

* Add an optional displayedName param to createMetronomeContract and provisionMetronomeContract so callers can set a human-readable name on a Metronome contract.
* Metronome's contracts.create rejects name when package_alias/package_id is set (our only creation path), so displayedName is applied via a follow-up v2.contracts.edit (update_contract_name) once the contract exists, alongside the existing custom-fields/Stripe-billing post-create steps.
* Use it in provisionCreditPricedFreePlan to name free-plan contracts Free Business USD / Free Business EUR based on the workspace's billing currency, making them identifiable in the Metronome dashboard.

## Tests

Tested locally:
* creating a free plan contract uses the displayed name
* creating a contract without a displayed name works as before

## Risk

Medium, impacts free plan creation but more generally the creation of new contracts with Metronome

## Deploy Plan

Deploy front

### DIFF
--- a/front/lib/api/subscription.ts
+++ b/front/lib/api/subscription.ts
@@ -114,6 +114,7 @@ async function provisionCreditPricedFreePlan(
     swapAt: "current-hour",
     enableStripeBilling: false,
     planCode: CREDIT_PRICED_FREE_PLAN_CODE,
+    displayedName: `Free Business ${currency.toUpperCase()}`,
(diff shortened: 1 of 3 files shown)


tdraier (approved):
LGTM

PR #28343, opened 2026-07-01
Pricing: cap per group 2: Always consider max group cap when computing effective cap

fabiencelier:
Related to https://github.com/dust-tt/tasks/issues/9339

## Description

Effective cap now account for the max group cap everywhere it is computed.
Note that there is no way to currently set the cap so far so nothing will change for now.

## Risk

low

## Deploy Plan

deploy front

### DIFF
--- a/front/lib/resources/group_resource.ts
+++ b/front/lib/resources/group_resource.ts
@@ -1246,17 +1246,16 @@ export class GroupResource extends BaseResource<GroupModel> {
   }
 
   // For each user, the highest per-group pool cap (excluding seat allowance)
-  // among the cap-eligible groups they belong to. Users with no capped group are
-  // absent from the map (the caller falls back to the workspace default). Used to
-  // resolve the "max(group caps)" term of a user's effective spend limit.
+  // among the cap-eligible (provisioned) groups they belong to. Users with no
+  // capped group are absent from the map (the caller falls back to the workspace
+  // default). Used to resolve the "max(group caps)" term of a user's effective
+  // spend limit.
   static async listMaxPoolCapAwuCreditsByUserModelIdInWorkspace({
     workspace,
     userModelIds,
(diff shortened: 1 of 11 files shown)


front/lib/resources/group_resource.ts
```diff
@@ -1244,6 +1245,65 @@ export class GroupResource extends BaseResource<GroupModel> {
...
+    userModelIds: ModelId[];
+    groupKinds?: Exclude<GroupKind, "system">[];
+  }): Promise<Map<ModelId, number>> {
+    const result = new Map<ModelId, number>();
+    if (userModelIds.length === 0) {
+      return result;
+    }
+
+    const now = new Date();
+    const memberships = await GroupMembershipModel.findAll({
```
tdraier, line 1266: nit : this should be available in group_resource

front/types/groups.ts
```diff
@@ -36,6 +36,12 @@ export const GROUP_KINDS = [
 ] as const;
 export type GroupKind = (typeof GROUP_KINDS)[number];
 
+// Group kinds that can carry a per-group usage spend limit and be surfaced in
+// the Usage > Groups admin table. Real user-facing groups only: "global" is the
+// whole workspace (that's the default), "system" is internal, and the
+// agent/skill/space editor groups are per-resource, not membership groups.
+export const CAP_ELIGIBLE_GROUP_KINDS = ["regular", "provisioned"] as const;
```
tdraier: regular are space groups only iirc, do we want to see them here ?

tdraier (approved):
LGTM

front/types/groups.ts
fabiencelier: Indeed, switching to only provisioned

front/lib/resources/group_resource.ts
fabiencelier: There is no exact equivalent function to do this

PR #28049, opened 2026-06-26
feat(front): add model picker to conversation input bar

Nils-Fedrigo:
## Description

https://github.com/user-attachments/assets/68ebb81c-6433-4f8c-ba7a-c4f446f89490

Adds the model-picker dropdown to the conversation input bar and mounts it next to the agent chip, gated behind the `models_picker` feature flag (`dust_only` stage).

- Introduces `InputBarModelPicker`: keep **Auto** (the agent's configured model) or pick a specific model, with an optional reasoning-effort override and a set of suggested pins.
- Mounts the picker in `InputBarButtons` (when the `model-picker` action is enabled) and feeds its selection up through `InputBar` / `InputBarContainer` into the send path plumbed in #28489, which the backend (#28346) applies as a per-message override.

The send-path plumbing (#28489) and the agent-chip rework (#28488) it depends on are split into their own PRs below in the stack; this PR is just the picker component and its wiring into the input bar.

Top of the model-picker stack (bottom → top): #28344 → #28346 → #28488 → #28489 → **#28049 (this)**.

## Tests

Manual: with `models_picker` enabled, open the picker, keep Auto and confirm the agent's configured model runs; pick a specific model (and a reasoning effort) and confirm the sent message runs on the override. With the flag off, the picker is not shown.

## Risk

Low. Frontend is gated by `models_picker`; the backend override only takes effect when a selection is sent, so with no selection behavior is unchanged. Rollback by disabling the flag.

PR #28344, opened 2026-07-01
feat(front): add per-message requested-model columns to agent messages

Nils-Fedrigo:
## Description

First PR of a stack allowing to change the model + reasoning effort level at runtime for the messages. 
This uses a model picker in the input bar. 
We also lay the groundwork necessary to have different models picked according to the message.

In this PR, we add 3 columns to AgentMessageModel
- requestedProviderId
- requestedModelId
- requestedReasoningEffort

Why is this stored? 
- front display
- allow for "same model" retry
- auditability

Why store this triplet 
- modelId: pretty logical
- providerId : needed to differentiate providers of same models to not break EU-only inference, of providers of OS models. Cost might differ. Overall, the whole code disambiguates by provider&model
- reasoningEffort : cost might greatly differ depending on the reasoning effort

Other things we can potentially store
- a reason for override: model picker or auto mode later on. We'll see when we get to auto mode if that is needed.

## Tests

Schema-only change; covered by existing model/migration tooling. Migration applied locally.

## Risk

Low. Additive, nullable columns with a pre-deploy migration — no backfill, no read/write path yet,
trivially rollback-safe.

## Deploy Plan

Pre-deploy migration: must run before the code that reads/writes these columns (#28346) is live.
Standard expand pattern; columns are nullable so no backfill is required.

### DIFF
--- a/front/lib/models/agent/conversation.ts
+++ b/front/lib/models/agent/conversation.ts
@@ -459,12 +459,9 @@ export class AgentMessageModel extends WorkspaceAwareModel<AgentMessageModel> {
   declare prunedContext: boolean | null;
   declare costCredits: number | null;
 
-  // Per-message model override from the input-bar model picker. `requestedModelTier`
-  // records the abstract tier the user picked ("efficient" | "balanced" | "powerful",
-  // null for an explicit "advanced" pick or no pick); the provider/model/effort
-  // triplet is the concrete model that tier (or advanced pick) resolved to at
-  // send time. All null when the message runs the agent's configured model.
-  declare requestedModelTier: string | null;
+  // Per-message model override from the input-bar model picker: the concrete
+  // provider/model/effort triplet the user picked at send time. All null when
+  // the message runs the agent's configured model.
   declare requestedProviderId: string | null;
   declare requestedModelId: string | null;
   declare requestedReasoningEffort: string | null;
@@ -555,11 +552,6 @@ AgentMessageModel.init(
     },
     costCredits: {
       type: DataTypes.INTEGER,
-      allowNull: true,
-      defaultValue: null,
-    },
-    requestedModelTier: {
-      type: DataTypes.STRING,
       allowNull: true,
       defaultValue: null,
     },
--- a/front/lib/models/agent/conversation.ts
+++ b/front/lib/models/agent/conversation.ts
@@ -460,7 +460,7 @@ export class AgentMessageModel extends WorkspaceAwareModel<AgentMessageModel> {
   declare costCredits: number | null;
 
   // Per-message model override from the input-bar model picker. `requestedModelTier`
-  // records the abstract tier the user picked ("fast" | "balanced" | "powerful",
+  // records the abstract tier the user picked ("efficient" | "balanced" | "powerful",
   // null for an explicit "advanced" pick or no pick); the provider/model/effort
   // triplet is the concrete model that tier (or advanced pick) resolved to at
   // send time. All null when the message runs the agent's configured model.
--- a/front/lib/models/agent/conversation.ts
+++ b/front/lib/models/agent/conversation.ts
@@ -458,6 +458,16 @@ export class AgentMessageModel extends WorkspaceAwareModel<AgentMessageModel> {
   declare completedAt: Date | null;
   declare prunedContext: boolean | null;
   declare costCredits: number | null;
+
+  // Per-message model override from the input-bar model picker. `requestedModelTier`
+  // records the abstract tier the user picked ("fast" | "balanced" | "powerful",
+  // null for an explicit "advanced" pick or no pick); the provider/model/effort
+  // triplet is the concrete model that tier (or advanced pick) resolved to at
+  // send time. All null when the message runs the agent's configured model.
+  declare requestedModelTier: string | null;
+  declare requestedProviderId: string | null;
+  declare requestedModelId: string | null;
+  declare requestedReasoningEffort: string | null;
 }
 
 AgentMessageModel.init(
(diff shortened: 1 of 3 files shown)


Nils-Fedrigo:

- #28344 ◀
    - #28346
        - #28488
            - #28489
                - #28049

front/lib/models/agent/conversation.ts
```diff
@@ -458,6 +458,13 @@ export class AgentMessageModel extends WorkspaceAwareModel<AgentMessageModel> {
   declare completedAt: Date | null;
   declare prunedContext: boolean | null;
   declare costCredits: number | null;
+
+  // Per-message model override from the input-bar model picker: the concrete
+  // provider/model/effort triplet the user picked at send time. All null when
```
Fraggle, line 463: Is provider really needed in general ?

And further more, shall we collapse the reasonning effort in the modelId as we discussed IRL ?

(goal is to only store one column)

Nils-Fedrigo: The provider-agnostic lookup breaks the moment two providers serve the same open-weight modelId (fireworks / togetherai / deepseek)

It seems weird to me to not save that for completeness

Nils-Fedrigo: > shall we collapse the reasonning effort in the modelId as we discussed IRL ?

That's a pretty structuring change, and there is no urgency as I see it to go for it, I'd prefer to stay like that for now

Nils-Fedrigo: >Is provider really needed in general ?


Imagine a scenario where we want anthropic only EU then we would default to the US provider and break EU-only inference

Fraggle: > The provider-agnostic lookup breaks the moment two providers serve the same open-weight modelId (fireworks / togetherai / deepseek)

I missed that one :)

spolu:
I don't have enough context to review this one. Happy to review a design doc / proposal of course 👍

Nils-Fedrigo:
@spolu @Fraggle I added more thorough context in the description of this PR, can you please come at it again? Many thanks!

@spolu I added you as a reviewer as I believe you are the member of [data-model-owners](https://github.com/orgs/dust-tt/teams/data-model-owners) with the most context on this. 
Let me know if you need more tokens.

spolu:
Same cause same effects :) 

I don't have enough context to review this one. Happy to review a design doc / proposal of course 👍

Nils-Fedrigo:
@spolu no problem! 

I thought that expanding the description of this PR would do it.

What in particular would you like to see and what part of the PR description still has holes?
Or could you provide more tokens as to what's still bugging you?

Here was the [design doc](https://app.notion.com/p/dust-tt/Introduce-default-restricted-model-on-Dust-38728599d94180b49431cf27119ad5a0?source=copy_link#38a28599d94180a88ce8e0e012658e65) that you reviewed at the beginning of the project, I try to keep up to date with our current exchanges.

spolu:
Nothing bugs me it's just that I generally don't have sufficient context to review single PRs from initiatives I'm not a part of 👍

front/migrations/pre-deploy/20260701165547_add_requested_model_to_agent_message.sql
```diff
@@ -0,0 +1,20 @@
+/*
+Statement 0
+*/
+SET SESSION statement_timeout = 3000;
+SET SESSION lock_timeout = 3000;
+ALTER TABLE "public"."agent_messages" ADD COLUMN "requestedProviderId" character varying(255) COLLATE "pg_catalog"."default" DEFAULT NULL::character varying;
```
Fraggle, line 6: I'm surprised by the COLLATE thing, never saw that in other case.

Fraggle (approved):
LGTM, worst case scenario, easy to rollback.

front/lib/models/agent/conversation.ts
```diff
@@ -548,6 +555,21 @@ AgentMessageModel.init(
...
+    requestedProviderId: {
+      type: DataTypes.STRING,
+      allowNull: true,
+      defaultValue: null,
+    },
+    requestedModelId: {
+      type: DataTypes.STRING,
+      allowNull: true,
+      defaultValue: null,
+    },
```
id13, line 567: I would rename Id to Slug, as these are loosely held Ids, and Id is an extremely charged word (let alone the eternal debate around sID vs SID vs ModelId 🫠 )

front/migrations/pre-deploy/20260701165547_add_requested_model_to_agent_message.sql
id13, line 6: what are the access patterns?

front/migrations/pre-deploy/20260701165547_add_requested_model_to_agent_message.sql
```diff
@@ -0,0 +1,20 @@
...
+
+/*
+Statement 1
+*/
+SET SESSION statement_timeout = 3000;
+SET SESSION lock_timeout = 3000;
+ALTER TABLE "public"."agent_messages" ADD COLUMN "requestedModelId" character varying(255) COLLATE "pg_catalog"."default" DEFAULT NULL::character varying;
```
id13, line 13: same

front/migrations/pre-deploy/20260701165547_add_requested_model_to_agent_message.sql
```diff
@@ -0,0 +1,20 @@
...
+
+/*
+Statement 2
+*/
+SET SESSION statement_timeout = 3000;
+SET SESSION lock_timeout = 3000;
+ALTER TABLE "public"."agent_messages" ADD COLUMN "requestedReasoningEffort" character varying(255) COLLATE "pg_catalog"."default" DEFAULT NULL::character varying;
```
id13, line 20: same

id13: My underlying question being around indexing. This is a big table.
Which means brainlessly adding indexes for everything is not a good strategy, but not adding one is probably going to take forever at query time

id13 (approved):
LGTM, some questions but nothing blocking

front/lib/models/agent/conversation.ts
Nils-Fedrigo: consistently bad is better than inconsistently good I'd say :/ currently the corresponding columns are (badly) named providerId and modelId

front/migrations/pre-deploy/20260701165547_add_requested_model_to_agent_message.sql
Nils-Fedrigo: just read-along, never filter-by! So no index needed

We hardly ever hard delete model so not even deletion prompts an index

PR #28346, opened 2026-07-01
feat(front): resolve per-message model override at send and apply at run

Nils-Fedrigo:
## Description

Backend for the input-bar model picker (on top of the #28344 migration that adds the `requested*`
columns). Wires a per-message model override end-to-end while staying a no-op until a client sends
`modelSelection` (the picker UI lands in #28049).

- Adds an optional `modelSelection` (`providerId` / `modelId` / optional `reasoningEffort`) to the
  message-post request schema — covering both the create-message and create-conversation paths — and
  documents it in swagger.
- `resolveModelSelection` (new, in `lib/api/assistant/models.ts`) turns a raw picker selection into a
  concrete model at send time, validating it against the workspace's enabled models via the same
  enablement gate used everywhere else. Unknown or disabled models resolve to `null`, so the agent
  falls back to its configured model.
- `postUserMessage` resolves the selection, then `createAgentMessages` persists the resolved
  provider/model/effort on the agent message. The override is preserved on retry so a retried message
  re-runs the same picked model.
- `requestedAgentModelFromColumns` rebuilds the resolved model from the stored columns (defensive
  validation) and serializes it onto `AgentMessageType.requestedModel`.
- `getAgentLoopDataWithAuth` is the single choke point that applies the override: it overrides
  `agentConfiguration.model` (provider/model/effort only; temperature and other settings are
  inherited) so every downstream consumer runs the picked model.

When no selection is sent, the agent runs its configured model exactly as before.

## Tests

Exercised by the existing conversation message-post endpoint tests, which pass unchanged (default
no-op path). Resolution is guarded by the workspace enablement gate, and column deserialization
validates provider/model/effort defensively.

## Risk

Low. Behavior only changes for requests that carry `modelSelection`, and no client sends it until
#28049 ships. An override can never select a model the workspace can't run (reuses the enablement
gate). Rollback-safe.

## Deploy Plan

Depends on the #28344 migration being deployed first — this PR reads and writes the `requested*`
columns on agent messages.

### DIFF
--- a/front/lib/api/assistant/conversation.ts
+++ b/front/lib/api/assistant/conversation.ts
@@ -161,7 +161,7 @@ import {
 } from "@app/types/assistant/mentions";
 import type {
   ModelSelectionType,
-  RequestedAgentModel,
+  ResolvedRequestedModel,
 } from "@app/types/assistant/models/types";
 import type {
   ContentFragmentContextType,
@@ -600,14 +600,8 @@ export async function postUserMessage(
   // Resolve the picker selection to a concrete model for this workspace. If it
   // cannot be honored (unknown/disabled model), we leave `requestedModel` null
   // and the agent's configured model is used.
-  const resolvedRequestedModel = modelSelection
+  const requestedModel: ResolvedRequestedModel | null = modelSelection
     ? resolveModelSelection(auth, modelSelection, { featureFlags })
-    : null;
-  const requestedModel: RequestedAgentModel | null = resolvedRequestedModel
-    ? {
-        ...resolvedRequestedModel,
-        tier: modelSelection?.type === "tier" ? modelSelection.tier : null,
-      }
     : null;
 
   const isPartOfPod = isPodConversation(conversation);
--- a/front/lib/api/assistant/conversation.ts
+++ b/front/lib/api/assistant/conversation.ts
@@ -27,7 +27,10 @@ import {
   batchRenderMessages,
   batchRenderUserMessagesWithoutMentions,
 } from "@app/lib/api/assistant/messages";
-import { isProviderWhitelisted } from "@app/lib/api/assistant/models";
+import {
+  isProviderWhitelisted,
+  resolveModelSelection,
+} from "@app/lib/api/assistant/models";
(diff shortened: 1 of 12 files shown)


Nils-Fedrigo:

- #28344
    - #28346 ◀
        - #28488
            - #28489
                - #28049
                    - #28595

front/lib/api/assistant/conversation.ts
```diff
@@ -560,6 +568,10 @@ export async function postUserMessage(
     skipToolsValidation: boolean;
     doNotAssociateUser?: boolean;
     skipDustAutoMention?: boolean;
+    // Optional per-message model override from the input-bar model picker,
+    // resolved to a concrete model below and applied to every agent message
+    // created for this user message.
```
Nils-Fedrigo:
```suggestion
```

front/lib/api/assistant/models.ts
```diff
@@ -199,3 +208,74 @@ function _getLargeWhitelistedModel(
...
+  requestedModelId: string | null;
+  requestedReasoningEffort: string | null;
+}): ResolvedRequestedModel | null {
+  const { requestedProviderId, requestedModelId, requestedReasoningEffort } =
+    row;
+
+  if (
+    !requestedProviderId ||
+    !requestedModelId ||
+    !requestedReasoningEffort ||
```
Fraggle, line 292: this one is never null ?

Fraggle (approved):
LGTM

front/types/assistant/agent_run.ts
```diff
@@ -318,8 +319,25 @@ export async function getAgentLoopDataWithAuth(
...
+          providerId: requestedModel.providerId,
+          modelId: requestedModel.modelId,
+          reasoningEffort: requestedModel.reasoningEffort,
+        }),
+      }
+    : agentConfiguration;
+
   return new Ok({
-    agentConfiguration,
+    agentConfiguration: finalAgentConfiguration,
```
Fraggle: As discussed IRL, this part is likely superseeded by https://github.com/dust-tt/dust/pull/28574

PR #28348, opened 2026-07-01
Fix: create child conversations in the same pod as the main conversation.

Fraggle:
## Description

`getOrCreateConversation` was not forwarding the parent conversation's `spaceId` when spawning child conversations via `run_agent`. As a result, sub-conversations created by an agent running inside a Pod were placed outside the Pod. Fixed by passing `spaceId: mainConversation.spaceId ?? undefined` to `DustAPI.createConversation`.

## Tests

Added Vitest unit tests for `getOrCreateConversation` covering two cases: `spaceId` is forwarded when the parent conversation has one, and omitted (`undefined`) when the parent has `spaceId: null`.

## Risk

Low. One-line change that adds a previously missing field. No data migration. Conversations already created without the correct `spaceId` are unaffected going forward.

## Deploy Plan

Deploy front.

### DIFF
--- a/front/lib/api/actions/servers/run_agent/conversation.test.ts
+++ b/front/lib/api/actions/servers/run_agent/conversation.test.ts
@@ -1,0 +1,202 @@
+import type { AgentLoopRunContextType } from "@app/lib/actions/types";
+import { getOrCreateConversation } from "@app/lib/api/actions/servers/run_agent/conversation";
+import type { Authenticator } from "@app/lib/auth";
+import { generateRandomModelSId } from "@app/lib/resources/string_ids_server";
+import type { AgentConfigurationType } from "@app/types/assistant/agent";
+import type {
+  AgentMessageType,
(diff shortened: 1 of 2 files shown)


Fraggle:

- #28348 ◀

front/lib/api/actions/servers/run_agent/conversation.test.ts
```diff
@@ -0,0 +1,202 @@
...
+    reactions: [],
+    modelInteractionDurationMs: null,
+    completionDurationMs: null,
+    rank: 1,
+    branchId: null,
+    richMentions: [],
+    costCredits: null,
+  } as unknown as AgentMessageType;
+
+  const mainAgent = {
```
flvndvd, line 80: Can e use the factories here? We have for agents, conversations, messages, ... So it's a bit more robust to changes.

flvndvd (approved):
LGTM

PR #28351, opened 2026-07-01
fix(conversation): stop duplicated inline activity steps on Temporal retries

rfrenoy:
### Context
In a streaming agent turn, the inline "Thinking / activity" steps could be rendered dozens of times within a single turn, flooding the conversation view ([tasks#9287](https://github.com/dust-tt/tasks/issues/9287), [tasks#8022](https://github.com/dust-tt/tasks/issues/8022)). The duplication only appeared while the turn was streaming and disappeared once it completed. A hard refresh mid-stream did not help.

An agent turn runs as a Temporal loop of steps (`agentLoopWorkflow`). Each step is one `runModelAndCreateActionsActivity` (the LLM call, emitting `generation_tokens` for chain-of-thought/text and `tool_params` for tool calls) followed by `runToolActivity`. The model activity is configured with `maximumAttempts: 5`, so on failure, Temporal re-runs the whole step's model call.

Every emitted event takes two paths:
  - Redis stream (`stream:message-<sId>`): append-only log, tagged with its `step`, replayed to the client over SSE. Retried attempts are appended as brand-new entries.
  - DB step content (`AgentStepContentResource.createNewVersion`): versioned per (step, index), latest-wins. A retry overwrites the previous version.

While streaming, the client (`useAgentMessageStream`) rebuilds the timeline by reducing the Redis event stream into `inlineActivitySteps` (append-based). Once the turn ends, it replaces that array wholesale with the server-rendered `contentView`, which is computed from the collapsed DB step content, which is why the bug self-healed at "Done" and why reload was always clean.

The client reducer could not absorb a retry's re-emitted events:
  - `seenEventIds` only drops exact stream-id replays. Retried events have new ids.
  - `appendThinkingStep` deduped only against the most-recent thinking step `appendContentStep` did not dedupe at all
  - `retryCoTBuffer` only suppressed an *in-progress* (unflushed) CoT. Once a segment had been flushed into the timeline (because a tool call followed), the retry re-appended it.

So each of the up-to-5 attempts piled its steps on top of the previous ones, and a fresh mount mid-stream replayed the same log and rebuilt the same duplicates.

### Fix
Mirror the server's per-step, latest-wins semantics in the live reducer:

  - Tag every inline activity step with the agent-loop `step` that produced it (`InlineActivityStep.step`, set both client-side and in the server builder).
  - Track `currentStep` (advanced only by `generation_tokens`, the only events that fill the CoT/text buffer).
  - When a `generation_tokens` CoT event arrives with a fresh `traceId` on the same step (a new step also gets a fresh traceId, but its step number advances), treat it as a Temporal retry: drop the inline steps already committed for that step and rebuild them from the retry's events. This generalizes the old `retryCoTBuffer` beyond the unflushed case. The buffer is kept only to de-flicker the active CoT.

Note: it looks like `useChildAgentStream` has the same Temporal-retry duplication, but is out of scope for this PR. Will be done in a follow-up PR. Added a TODO comment about it.

## Tests

Locally, by injecting code in the temporal worker to force retries:

### Before the fix, see duplicated attempts
Screenshot 2026-07-01 at 21 56 40

### After the fix, only latest attempt is displayed
Screenshot 2026-07-01 at 21 54 53

## Risk

Low

## Deploy Plan

Front

### DIFF
--- a/front/hooks/useAgentMessageStream.test.ts
+++ b/front/hooks/useAgentMessageStream.test.ts
@@ -249,9 +249,14 @@ describe("appendThinkingStep", () => {
   it("appends a thinking step when none exists", () => {
     const steps: InlineActivityStep[] = [];
     expect(
-      appendThinkingStep(steps, "Enable the toolset", "thinking-1")
+      appendThinkingStep(steps, "Enable the toolset", "thinking-1", 0)
     ).toEqual([
-      { type: "thinking", content: "Enable the toolset", id: "thinking-1" },
+      {
+        type: "thinking",
+        content: "Enable the toolset",
+        id: "thinking-1",
+        step: 0,
+      },
     ]);
   });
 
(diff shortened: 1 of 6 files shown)


PopDaph (approved):
As discussed IRL this won't fix the problem reported in the card (we see there's more than 5 occurrences), but this will fix the problem you describe. Thanks very much for digging into this one 🙏🏻

PR #28353, opened 2026-07-02
feat(skills): add Activation global skill [S1-1]

frankaloia:
## Summary
* Implements [dust-tt/tasks#9352](https://github.com/dust-tt/tasks/issues/9352).
* Consider this an initial skeleton for the activation skill (to be given a better name at a future date)
* Behind FF
* Will be iterating on this continuously in initiative

## Testing
* manually went through workflow in local

## Risk

Low, behind FF

### DIFF
--- a/front/types/shared/feature_flags.ts
+++ b/front/types/shared/feature_flags.ts
@@ -355,6 +355,10 @@ export const WHITELISTABLE_FEATURES_CONFIG = {
       "Model picker in the conversation input bar: pick a model tier (Fast, Balanced, Powerful, Frontier) or a specific model.",
     stage: "dust_only",
   },
+  activation_skill: {
+    description: "Enable the Activation skill for agentic user activation pods",
+    stage: "dust_only",
+  },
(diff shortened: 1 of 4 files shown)


front/types/shared/feature_flags.ts
```diff
@@ -355,6 +355,10 @@ export const WHITELISTABLE_FEATURES_CONFIG = {
       "Model picker in the conversation input bar: pick a model tier (Fast, Balanced, Powerful, Frontier) or a specific model.",
     stage: "dust_only",
   },
+  activation_skill: {
+    description: "Enable the Activation skill for agentic user activation pods",
+    stage: "dust_only",
+  },
```
adrsimon, line 361: nit: do we want a ff for just the skill or rather the whole feature ? will there be another one ?

adrsimon (approved):
LGTM

front/types/shared/feature_flags.ts
frankaloia: Just the skill was intentional as I can see a likely world where we deliver it first

PR #28356, opened 2026-07-02
fix(conversations): let pod members read agent messages from other spaces

ElPicador:
## Description

Moving a conversation into a Pod is supposed to make it readable by every Pod member, but when the conversation includes a message from an agent scoped to a different (e.g. private) space, that promise breaks: `GET .../conversations/{cId}` returns 200, while `GET .../conversations/{cId}/messages` 403s with `conversation_with_unavailable_agent` for any Pod member who isn't also a member of the agent's original space.

**The bug:** moving a conversation into a Pod (`moveConversationToProject`) rewrites the *conversation's* `requestedSpaceIds` to just the Pod's space, so the conversation record itself becomes readable Pod-wide. It never touches the *agent's* own `requestedSpaceIds`, which keeps pointing at wherever the agent originally lived. Rendering the message history resolves each agent message's configuration via `getAgentConfigurations`, which filters agents by the agent's own space permissions — so for a Pod member outside that original space, the agent config comes back missing and rendering fails.

**The fix:** when rendering messages that are already part of a Pod conversation's history, skip the per-agent space check — Pod-level read access (already verified before we get here) is sufficient to vouch for that history, consistent with how Pod conversations are designed to work elsewhere in the codebase. This is implemented as an explicit opt-in flag (`dangerouslySkipSpacePermissionCheck`) threaded through `getAgentConfigurations`, so every other caller (agent pickers, mention suggestions, etc.) is unaffected.

Fixes dust-tt/tasks#9344.

## Tests

Added a functional test reproducing the exact scenario (agent restricted to a private space, conversation moved into a Pod, a Pod member outside that private space) — verified it fails with 403 against the pre-fix code and passes with the fix. Added a second test confirming a user with no access to the Pod at all is still correctly denied, to guard against the bypass being too broad.

### DIFF
--- a/front/lib/api/assistant/messages.ts
+++ b/front/lib/api/assistant/messages.ts
@@ -226,7 +226,10 @@ function renderUserMessage(
 async function batchRenderUserMessages(
   auth: Authenticator,
   messages: MessageModel[],
-  mentionsByMessageId?: Map<ModelId, MentionModel[]>
+  mentionsByMessageId: Map<ModelId, MentionModel[]> | undefined,
+  // Used in pods: Pod conversations are, by design, readable by anyone with read access to the pod's
+  // space (see front/lib/api/assistant/conversation/permissions.ts).
+  bypassAgentSpacePermissionCheck: boolean
 ): Promise<UserMessageType[]> {
   const userMessages = messages.filter(
     (m) => m.userMessage !== null && m.userMessage !== undefined
@@ -267,6 +270,7 @@ async function batchRenderUserMessages(
       ? await getAgentConfigurations(auth, {
           agentIds: agentConfigurationIds,
           variant: "extra_light",
+          dangerouslySkipSpacePermissionCheck: bypassAgentSpacePermissionCheck,
         })
       : [];
   const reactionsByMessageId = await getMessagesReactions(auth, {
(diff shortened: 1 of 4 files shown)


ElPicador:

- #28356 ◀

front/lib/api/cancel.ts
```diff
@@ -72,7 +72,8 @@ export async function terminateMessageGeneration(
     messageRows,
     "full",
     null,
-    new Map()
+    new Map(),
+    !!conversation.spaceId
```
Fraggle, line 76: `isPodConversation()`

front/lib/api/assistant/messages.ts
```diff
@@ -906,17 +915,26 @@ export async function batchRenderMessages<V extends RenderMessageVariant>(
     }
   }
 
+  // Pod conversations are, by design, readable by anyone with read access to the pod's
+  // space (see front/lib/api/assistant/conversation/permissions.ts). `conversation` here
+  // was already fetched through a permission-checked path, so pod-level access to it also
+  // vouches for the agents mentioned in its history, even if those agents' own
+  // `requestedSpaceIds` still point at a different (e.g. private) space.
+  const bypassAgentSpacePermissionCheck = !!conversation.spaceId;
```
Fraggle, line 923: `isPodConversation()`

Fraggle (approved):
Looks reasonable

ElPicador:
closing in favor of https://github.com/dust-tt/dust/pull/28395/changes

PR #28357, opened 2026-07-02
chore(debug): run profiler without tsx + add front-api variant

davidebbo:
## What

Makes `profile_k8s_pods.ts` work against the front-api pods.

### 1. Build the profiler bundle in CI (not on the pod)

`front`'s `debug:profiler` ran `run_profiler.ts` with plain `node`, which failed with `ERR_MODULE_NOT_FOUND` on `@app/lib` — Node can't resolve the `@app/*` tsconfig alias or extensionless imports.

Following review feedback, instead of transforming TypeScript at runtime on production pods, the script is **compiled during the Docker build**, mirroring the existing `migrate.js` pattern:

- `dockerfiles/front.Dockerfile` bundles `scripts/debug/run_profiler.ts` → `dist/run_profiler.js` (esbuild, same flags as `migrate.js`), and copies the artifact into the workers and front-api runtime images.
- `front/scripts/debug/run-profiler.cjs` — a tiny shim mirroring `run-migrate.cjs`: runs the pre-built `dist/run_profiler.js` with plain `node` when present (production pods — no runtime transform), and falls back to `tsx` on developer machines where the bundle isn't built.
- Both `debug:profiler` scripts point at the shim.

### 2. Add `debug:profiler` to `front-api`

`scripts/debug/profile_k8s_pods.ts` execs `npm run debug:profiler` in the pod's WORKDIR. The front-facing deployment runs the `front-api` image (`WORKDIR /app/front-api`), and that pod is the one actually serving the `/api/debug/profiler` endpoint (front/Next has no such endpoint). `front-api/package.json` had no such script, so the exec failed there.

It reuses front's script via `../front/scripts/debug/run-profiler.cjs` (no duplication); the shim anchors its paths to its own location so it resolves `front/dist/run_profiler.js` from either workspace.

### 3. Target the server's actual host instead of `localhost`

The script previously fetched a hardcoded `http://localhost:3000`, failing with `fetch failed` (ECONNREFUSED). `front-api/server.ts` binds to `process.env.HOSTNAME ?? "localhost"`, and in Kubernetes `HOSTNAME` is the pod name (mapped to the pod IP) — so the server never listens on `127.0.0.1`.

Added `config.getSelfServerBaseUrl()` mirroring the server's binding (`HOSTNAME`/`PORT` with the same defaults) and used it in `run_profiler.ts`. Falls back to `http://localhost:3000` locally.

## Testing

- `npm run debug:profiler` in both `front/` and `front-api/` runs the pre-built bundle when present, and falls back to `tsx` when it isn't — both print the expected dry-run message.
- With `-- --execute`, the bundle resolves the full `@app/*` graph and reaches the fetch.
- `config.getSelfServerBaseUrl()` returns `http://<pod-hostname>:3000` when `HOSTNAME` is set, and `http://localhost:3000` locally.


front/scripts/esbuild-loader.mjs
```diff
@@ -0,0 +1,76 @@
+// Node module customization hooks that let `node` run our TypeScript scripts
+// without `tsx`. They cover the two things plain Node lacks:
+//   1. Resolving the `@app/*` tsconfig path alias and extensionless imports.
+//   2. Transforming TypeScript (types, enums, ...) via the already-present esbuild.
+//
+// Registered from `register-esbuild.mjs` and used e.g. by `npm run debug:profiler`.
+import { existsSync, readFileSync, statSync } from "node:fs";
```
flvndvd: @davidebbo why not just building it in the CI, so we avoid running this on a production pod?

davidebbo: Good call — switched to compiling the script in the Docker build (`dist/run_profiler.js`, same as `migrate.js`) with a `run-profiler.cjs` shim that mirrors `run-migrate.cjs`: production pods run the pre-built bundle with plain `node` (no runtime transform), and it falls back to `tsx` only on dev machines. Dropped the runtime esbuild loader.

davidebbo: @flvndvd ok, I changed it to be like the migrate script.

dockerfiles/front.Dockerfile
```diff
@@ -49,6 +49,10 @@ COPY /scripts/db /app/scripts/db
 # Compile migration script so all runtime images have dist/migrate.js without needing TypeScript sources
 RUN npx esbuild scripts/migrate.ts --bundle --platform=node --target=node22 --alias:@app=. --packages=external --outfile=dist/migrate.js --sourcemap
 
+# Compile the profiler script so `npm run debug:profiler` runs a pre-built bundle
+# on production pods instead of transforming TypeScript at runtime.
+RUN npx esbuild scripts/debug/run_profiler.ts --bundle --platform=node --target=node22 --alias:@app=. --packages=external --outfile=dist/run_profiler.js --sourcemap
```
flvndvd: Let's use `node24` to align on the base deps image.

front/lib/api/config.ts
```diff
@@ -463,6 +463,16 @@ const config = {
   getProfilerSecret: (): string | undefined => {
     return EnvironmentConfig.getOptionalEnvVariable("DEBUG_PROFILER_SECRET");
   },
+  // Base URL of this process's own HTTP server, used by in-pod callers such as
+  // the profiler script to reach the server. Must mirror how front-api/server.ts
+  // binds: in Kubernetes HOSTNAME is the pod name (mapped to the pod IP), so the
+  // server does not listen on localhost.
+  getSelfServerBaseUrl: (): string => {
```
flvndvd: Not sure to understand why we need this? Is it to override the pod? Ideally those debug script does not leak env vars in the prod config. We can always change the CLI we use locally, to provide port and local host.

front/scripts/debug/run-profiler.cjs
```diff
@@ -0,0 +1,22 @@
+'use strict';
```
flvndvd: Why do have cjs here?

flvndvd (commented):
@

dockerfiles/front.Dockerfile
davidebbo: Done — using `--target=node24` now.

front/lib/api/config.ts
davidebbo: Root issue was that localhost doesn't work on pod. But I changed it to take host/port as CLI args.

front/scripts/debug/run-profiler.cjs
davidebbo: Mostly, I was matching what we have for run-migrate.cjs. But you're right, it's not necessary.

PR #28358, opened 2026-07-02
Fix invoice cleanup, add check on total to avoid double billing

tdraier:
## Description

`cleanMetronomeInvoiceLines` (used to strip commit/credit-covered lines from
Metronome-pushed Stripe draft invoices before finalization) relied on a
`metronome_commit_id` Stripe line-item metadata key to detect lines whose cost
was already covered by a commit. That key does not exist — Metronome never
sets it, so the check always evaluated to `false` and those lines were never
removed.

In practice this double-billed customers: Metronome represents a fully-applied
commit as a pair of lines on the same invoice with equal and opposite amounts
— a negative line (`"<label> applied"`) and the positive usage/subscription
line it offsets (description containing `(<label>)`). Only the negative line
was ever caught (via the `isNegative` filter); the matching positive line
stayed on the invoice, so the customer was charged for it a second time on top
of the actual commitment charge.

Fix:
- Added `findCommitAppliedLineIds`, which pairs the negative `"<label>
  applied"` line with the positive line whose description contains
  `(<label>)` and whose amount/currency match exactly, and marks the positive
  line for removal too.
- As a safety net, `cleanMetronomeInvoiceLines` now returns the invoice total
  before and after cleaning. If cleaning changed the total (i.e. our matching
  didn't net out as expected), `cleanAndFinalizeMetronomeDraftInvoice` logs a
  `panic: true` error and leaves the invoice as an unfinalized draft
  (`totals_mismatch` outcome) instead of finalizing a possibly-wrong invoice.
  It returns `Ok` rather than throwing, since the `lines_cleaned` metadata
  flag is already set by that point — a Temporal retry would just no-op into
  `skipped`, not re-attempt anything useful.

Found while investigating a production invoice where a "Pro Seat (Yearly)"
commitment line was double-billed; confirmed via logs that the
`metronome_commit_id` metadata key never appears on any Stripe line item.

## Tests

Type-checked (`npx tsgo --noEmit`, clean). No functional test added — this is
a narrow fix to a Temporal-invoked billing helper with no HTTP endpoint;
verified the new pairing logic against the two real production log samples
that surfaced the bug (matching negative/positive line pairs by description +
amount).

## Risk

Billing-critical path (Stripe invoice finalization for Metronome customers).
Mitigated by the new total-invariant check: if the fix's matching logic ever
misfires, the invoice is held as a draft and alerted on (`panic: true`) rather
than finalized incorrectly, so the worst case is a delayed invoice, not a
wrong one.

Only 2 known affected invoices found in production logs so far — need a
follow-up to check whether they were already finalized/paid and require a
credit note.

## Deploy Plan

Standard deploy. No migration, no config change, no feature flag.

### DIFF
--- a/front/lib/plans/stripe.ts
+++ b/front/lib/plans/stripe.ts
@@ -1351,34 +1351,95 @@ export async function finalizeInvoice(
 const METRONOME_INVOICE_LINES_CLEANED_FLAG = "lines_cleaned";
 
 /**
+ * Metronome represents a fully-applied commit/credit as a pair of lines on the
+ * same invoice with equal and opposite amounts: a negative line whose
+ * description is "<label> applied", and the positive usage/subscription line
+ * it offsets, whose description contains "(<label>)" (e.g. a negative
+ * "Platform Seat (Yearly) commitment: 53 seats applied" line offsetting a
+ * positive "Platform Seat (Yearly) (Platform Seat (Yearly) commitment: 53
+ * seats)" line). Neither line carries metadata identifying the pairing (there
+ * is no such thing as `metronome_commit_id` on Stripe line items — Metronome
+ * does not document or set one), so we match them by description + amount
+ * instead.
+ */
+function findCommitAppliedLineIds(
+  lines: Stripe.InvoiceLineItem[]
+): Set<string> {
+  const appliedSuffix = " applied";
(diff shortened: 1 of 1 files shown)


tdraier:

- #28358 ◀

rfrenoy (approved):

PR #28377, opened 2026-07-02
[front] poke members table: show active seat + scheduled seat change

tdraier:
## Description

The Poke members list fed the table via `getMembers(auth)` (no `activeOnly`),
which uses `getLatestMemberships` — the row with the greatest `startAt` per
user, with no `startAt <= now` filter. For a member with a scheduled future
seat change (e.g. seated mid-migration, flipping to `pro` at the pending
contract start), that returned the not-yet-active future row, so the table
showed `pro` as if it were current. The seat dropdown posts changes with
`immediate: true`, so the displayed value wasn't even what a change would diff
against.

Fetch members by their ACTIVE seat (`activeOnly: true`) and surface any
scheduled future seat change separately: the endpoint reads
`getScheduledFutureMemberships`, attaches `scheduledSeatType` /
`scheduledSeatChangeAt` (soonest per member) to each member, and the table
renders the active seat with a `→ <seat> on <date>` indicator beside it.

## Tests

locally 

## Risk

low, poke

## Deploy Plan

deploy front

### DIFF
--- a/front-api/routes/poke/workspaces/[wId]/memberships.ts
+++ b/front-api/routes/poke/workspaces/[wId]/memberships.ts
@@ -4,8 +4,10 @@ import type {
 } from "@app/lib/api/poke/memberships";
 import { getMembers } from "@app/lib/api/workspace";
 import { MembershipInvitationResource } from "@app/lib/resources/membership_invitation_resource";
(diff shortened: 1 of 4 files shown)


tdraier:

- #28377 ◀

thervier (approved):
LGTM

PR #28379, opened 2026-07-02
Add a path-keyed edit tool to the files server and treat Frame sources as regular files

flvndvd:
## Description

Frames are moving to an edit-source-then-publish flow: the model edits a Frame's source files in place on the Computer mount and explicitly publishes to rebuild the rendered Frame. Until now the files server could not participate in that flow. It had no way to make a surgical change to a file, and it refused to touch anything with a Frame content type, forcing every Frame edit through the file-id based edit tool that writes the canonical file directly.

This adds an edit tool to the files server that replaces an exact string match in a text file, with the same matching and occurrence-count semantics as the existing Frame edit tool. It also removes the Frame-specific rejections from file creation: a Frame file on the mount is now handled like any other file. This matches reality, since the model can already write these files freely through the Computer, so the rejection never gated a capability, it only made the files server inconsistent with it. Creating a file with a Frame content type just writes source to the mount. An actual shareable Frame is still minted through the interactive content server, which can take that path as a template source.

Writes to a Frame source preserve its content type on the mount and never touch the rendered Frame. The tool result says how to apply the change: workspaces with frame publishing are told to publish, others are pointed at the file-id edit tool. This is the only place the feature flag is consulted, availability of the tools themselves is not gated.

Downloading remote content over a Frame source stays rejected, as does the mirrored create tool on the separate MCP server surface, both untouched by this change.

## Risk

Low. New tool plus relaxed guards on the files server, no change to how Frames render or publish. Worst case a model writes a Frame source that only lands on the mount, which is already possible through the Computer today.

## Deploy Plan

Deploy front.

### DIFF
--- a/front/lib/api/actions/servers/files/tools/create.test.ts
+++ b/front/lib/api/actions/servers/files/tools/create.test.ts
@@ -4,6 +4,7 @@ import { createHandler } from "@app/lib/api/actions/servers/files/tools/create";
 import { createConversation } from "@app/lib/api/assistant/conversation";
 import { Authenticator } from "@app/lib/auth";
 import { getPrivateUploadBucket } from "@app/lib/file_storage";
+import { FeatureFlagFactory } from "@app/tests/utils/FeatureFlagFactory";
 import { createResourceTest } from "@app/tests/utils/generic_resource_tests";
 import { SpaceFactory } from "@app/tests/utils/SpaceFactory";
 import type { ConversationType } from "@app/types/assistant/conversation";
@@ -48,8 +49,16 @@ async function setupProjectConversation(): Promise<{
 }
 
 describe("createHandler", () => {
-  it("returns Err when content_type is a frame MIME type", async () => {
+  it("creates a new frame-typed file as a regular mount write", async () => {
     const { auth, conversation } = await setupProjectConversation();
+
+    const saveMock = vi.fn().mockResolvedValue(undefined);
+    vi.mocked(getPrivateUploadBucket).mockReturnValue({
+      file: vi.fn(() => ({
+        exists: vi.fn().mockResolvedValue([false]),
+        save: saveMock,
+      })),
+    } as unknown as ReturnType<typeof getPrivateUploadBucket>);
 
     const result = await createHandler(
       {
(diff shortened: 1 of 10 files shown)


front/lib/api/actions/servers/files/tools/create.test.ts
```diff
@@ -48,9 +49,17 @@ async function setupProjectConversation(): Promise<{
 }
 
 describe("createHandler", () => {
-  it("returns Err when content_type is a frame MIME type", async () => {
+  it("creates a new frame-typed file as a regular mount write", async () => {
     const { auth, conversation } = await setupProjectConversation();
 
+    const saveMock = vi.fn().mockResolvedValue(undefined);
+    vi.mocked(getPrivateUploadBucket).mockReturnValue({
```
aubin-tchoi: OOC don't we have a more generic way to mock this (or at least defined once in the file)?

front/lib/api/actions/servers/files/tools/create.ts
```diff
@@ -97,6 +95,20 @@ export async function createHandler(
   const sizeKb = Math.ceil(contentBuffer.byteLength / 1024);
   const verb = exists ? "Updated" : "Created";
 
+  if (isFrameSourceOverwrite) {
+    const flags = await getFeatureFlags(auth);
+    return new Ok([
+      {
+        type: "text",
+        text:
+          `Updated \`${path}\` (${sizeKb} KB). ` +
```
aubin-tchoi, line 104: nit: I think what you have here is in KiB instead of KB

front/lib/api/actions/servers/files/tools/edit.ts
```diff
@@ -0,0 +1,165 @@
...
+        );
+
+      default:
+        return new Err(
+          new MCPError(`Failed to write file \`${path}\`: ${err.message}`)
+        );
+    }
+  }
+
+  const pluralSuffix = occurrences === 1 ? "" : "s";
```
aubin-tchoi: nit: we have `pluralize`

front/lib/api/actions/servers/files/tools/utils.ts
```diff
@@ -156,13 +157,37 @@ export function frameFileCreateRejectedError(): MCPError {
...
+ */
+export function frameSourceUpdatedNotice({
+  hasFramePublishFF,
+}: {
+  hasFramePublishFF: boolean;
+}): string {
+  if (hasFramePublishFF) {
+    return (
+      "This updated the Frame's source only. The rendered Frame is unchanged until you " +
+      `publish it with \`${getPrefixedToolName(INTERACTIVE_CONTENT_SERVER_NAME, PUBLISH_INTERACTIVE_CONTENT_FILE_TOOL_NAME)}\`.`
```
aubin-tchoi, line 180: IIUC the model understands whether the frame is live or not by looking at the history and checking whether this message does not appear after the latest publish tool call

this may not survive well compactions, and interactions made by sub agents if any
How about a state on the frame that will surface this more directly? something like changing the file name to `MyFrame.tsx.tmp` (and maybe even keep the original file for backup)

aubin-tchoi (approved):
LGTM, mostly one comment

front/lib/api/actions/servers/files/tools/utils.ts
flvndvd: Seen IRL, will keep as-is and add a message on the publish tool.

front/lib/api/actions/servers/files/tools/create.test.ts
flvndvd: Update the global mock, this one isnt ideal since it's used to mock FileResource direct use of GCS SDK over the actual `DustFileSystem`.

PR #28381, opened 2026-07-02
Ungate Frame publishing by removing the frame_publish feature flag

flvndvd:
## Description

Frame publishing has been running gated for long enough and only a single workspace carries the flag, so this removes it and makes publishing available everywhere.

The publish tool is now always registered on the interactive content server. The Frames skill teaches the computer-first flow, edit the mounted source in place and republish, whenever the Computer is available, with the retrieve and edit loop remaining the fallback when it is not. The flag itself is deleted from the feature flag registry and the SDK.

This also simplifies the files server guidance that landed with the path-keyed edit tool. Since publish now always exists, the notice after a write to a Frame source no longer branches on the flag and simply tells the model to publish for the change to reach the rendered Frame.

No cleanup migration is included, the single flag row left in the database is inert once the flag is gone from the registry and can be deleted by hand.

## Risk

Low. Publishing becomes available to all workspaces at once, behavior for the previously flagged workspace is unchanged.

## Deploy Plan


aubin-tchoi (approved):
LGTM

spolu (approved):
LGTM

PR #28384, opened 2026-07-02
 Surface runner output when a sandbox function invocation fails

PopDaph:
When `dsbx function run` exits non-zero, we only returned "Sandbox function invocation failed
with exit code N" and discarded the runner's output, making failures (e.g. the runner unable to
POST its result back to front) impossible to diagnose, both in logs and for the agent.

This PR changes `SandboxFunctionResource.invoke()` so that on a non-zero exit code:

- we log a `logger.error` with the full `stdout`/`stderr` plus identifiers (workspace, pod,
  function, invocation, exit code)
- the returned error message includes the runner's stderr (falling back to stdout, truncated to
  2048 chars), which flows through the `call` tool's `MCPError` to the agent so it can see the
  actual cause

Same pattern as sandbox root commands in `egress.ts` (`stderr || stdout` appended to the error
message).

## Risk

Low. Error path only; no behavior change on success.

## Deploy

Deploy front.

### DIFF
--- a/front/lib/resources/sandbox_function_resource.test.ts
+++ b/front/lib/resources/sandbox_function_resource.test.ts
@@ -490,6 +490,92 @@ describe("SandboxFunctionResource", () => {
     });
   });
 
+  async function setupInvokeTest() {
+    const { authenticator, workspace } = await createResourceTest({
+      role: "admin",
+    });
+    const space = await SpaceFactory.project(workspace);
+    const file = await FileFactory.create(authenticator, null, {
+      contentType: sandboxFunctionContentType,
+      fileName: "comments.ts",
+      fileSize: 100,
+      status: "created",
+      useCase: "project_context",
+      useCaseMetadata: { spaceId: space.sId },
+    });
+    const sandboxFunction = await SandboxFunctionResource.makeNew(
+      authenticator,
+      {
+        space,
+        file,
+        slug: "add-comment",
+        description: "Add a comment.",
+        inputSchema,
+        outputSchema,
+      }
+    );
+    const sandbox = await SandboxResource.makeNew(authenticator, {
+      providerId: "test-provider-id",
+      status: "running",
+      baseImage: "dust-base",
+      version: "0.0.0-test",
+    });
+    vi.mocked(ensurePodSandboxReady).mockResolvedValue(
+      new Ok({ sandbox, freshlyCreated: false })
+    );
+    vi.mocked(generateSandboxFunctionInvocationToken).mockResolvedValue(
+      "sbt-function-token"
+    );
+
+    return { authenticator, sandboxFunction, sandbox };
+  }
+
+  it("surfaces the runner stderr when the invocation exits non-zero", async () => {
+    const { authenticator, sandboxFunction, sandbox } = await setupInvokeTest();
+    vi.spyOn(sandbox, "exec").mockResolvedValue(
+      new Ok({
+        exitCode: 1,
+        stdout: "some stdout",
+        stderr: "dsbx command failed: connection refused",
+      })
+    );
+
+    const result = await sandboxFunction.invoke(authenticator, {
+      input: { message: "hello" },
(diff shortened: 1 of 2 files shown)


front/lib/resources/sandbox_function_resource.test.ts
```diff
@@ -490,6 +490,92 @@ describe("SandboxFunctionResource", () => {
...
+    const space = await SpaceFactory.project(workspace);
+    const file = await FileFactory.create(authenticator, null, {
+      contentType: sandboxFunctionContentType,
+      fileName: "comments.ts",
+      fileSize: 100,
+      status: "created",
+      useCase: "project_context",
+      useCaseMetadata: { spaceId: space.sId },
+    });
+    const sandboxFunction = await SandboxFunctionResource.makeNew(
```
flvndvd, line 506: Could be nice to create a factory for this one 🙃.

front/lib/resources/sandbox_function_resource.ts
```diff
@@ -386,9 +392,31 @@ export class SandboxFunctionResource extends BaseResource<SandboxFunctionModel>
...
+        // Surface the runner's stderr (stdout when empty) so the agent sees the actual cause,
+        // not just the exit code.
+        const detail = truncate(
+          stderr || stdout,
+          SANDBOX_FUNCTION_ERROR_DETAIL_MAX_CHARS
+        ).trim();
         return new Err(
           new Error(
-            `Sandbox function invocation failed with exit code ${execResult.value.exitCode}.`
+            `Sandbox function invocation failed with exit code ${exitCode}${
```
flvndvd, line 417: Great that we also expose it to the model.

front/lib/resources/sandbox_function_resource.ts
```diff
@@ -386,9 +392,31 @@ export class SandboxFunctionResource extends BaseResource<SandboxFunctionModel>
...
+              detail ? `:\n${detail}` : "."
```
PopDaph, line 418: Could this leak any sensitive info??

flvndvd (approved):
🙏 Mind creating a Dashboard in DD with Dust to start centralizing instrumentation?

front/lib/resources/sandbox_function_resource.ts
flvndvd: Fair. Seen IRL, not exposing for now.

PopDaph: Discussed IRL with @flvndvd cc @fontanierh @spolu @zmarouf 
-> We merge as is I'll create a DD dashboard to monitor, it's internal only, and then we can assess.
It's super convenient that the agent has access to the error so worth the shot!

front/lib/resources/sandbox_function_resource.ts
```diff
@@ -386,9 +392,31 @@ export class SandboxFunctionResource extends BaseResource<SandboxFunctionModel>
...
+        const { exitCode, stdout, stderr } = execResult.value;
+        logger.error(
+          {
+            workspaceId: auth.getNonNullableWorkspace().sId,
+            spaceId: this.space.sId,
+            sandboxFunctionId: this.sId,
+            slug: this.slug,
+            invocationId: invocation.sId,
+            exitCode,
+            stdout: truncate(stdout, SANDBOX_FUNCTION_ERROR_LOG_MAX_CHARS),
```
spolu, line 404: Aren't we leaking potentially user data to DD 😬

zmarouf: @spolu Yeah. That's what @PopDaph means in the comment 🆙 
The goal is to understand what Sandbox function errors look like on our workspace before figuring out next steps. We can for example enable Sensitive Data Scanner in Datadog and just redact all PII on ingest.

spolu: Roger that 💯

PR #28385, opened 2026-07-02
[front] enh: track web AgentBrowser usage

aubin-tchoi:
## Description

This PR adds lightweight PostHog tracking to the web AgentBrowser so we can tell whether users interact with the panel and what agent-oriented action they take from it.

Goal is to have more analytics on this to make an informed decision on how to rework it.

It tracks selecting an agent and opening agent details from either the grid or the search dropdown, with source and basic agent metadata. It intentionally does not track mobile, tag, tab, or sort clicks to not add too much noise.

## Tests

- Tested locally.

## Risk

- Low. Client-side analytics only.

## Deploy Plan

- Deploy front.


fabiencelier (approved):
LGTM

PR #28388, opened 2026-07-02
 Fix Anthropic input token under-reporting by restoring cache-creation tokens in usage accounting

flvndvd:
## Description

We recently started differentiating cache TTL pricing on Anthropic, this opened a breach. The transition layer between the new model constructors router and the legacy usage pipeline only forwarded the flat total, which is zero whenever the breakdown is present, so every cache-write token silently disappeared from reported usage. Reported input was dramatically lower than what Anthropic actually billed, most visibly on the first message of a conversation where the entire prompt is a cache write, and the cache-write premium was never included in our cost tracking. Langfuse input totals were off by the same amount.

This change makes the transition layer report the true cache-write total and carries the short-lived versus long-lived split through the usage pipeline, since Anthropic bills one-hour cache writes at twice the base input
rate versus 1.25x for five-minute ones. The pricing tables gain a long-lived cache write rate for all Claude
models, including the EU Vertex surcharge and the Sonnet 5 intro pricing, and the cost computation now bills
long-lived writes at that premium. Providers that only report a flat cache-write total keep the exact same
behavior as before. The generic types and pricing model stay provider-agnostic and speak only of short and
long-lived cache tiers.

Langfuse keeps receiving the flat total, which is now correct, so its input numbers are fixed as well. Per-tier
cost inside Langfuse would require custom model definitions on their side and is left out of scope.


aubin-tchoi (approved):
LGTM

adrsimon (approved):
LGTM

flvndvd:
@theogz this PR fixes a first identified issue on the tokens used, tokens part of the cache writes were not accounted for properly.

PR #28389, opened 2026-07-02
fix(llm): stop excluding agent-platform gemini endpoints for non-byok workspaces

adrsimon:
## Description

Since enabling `use_new_llm_router` for all workspaces, ~78% of "Falling back to the old router" logs are gemini models, in both regions.

Root cause: `getProviderIdFilter` strips `google_ai_studio` from the allowed providers for every non-byok workspace (intent: route non-byok gemini through agent platform). But the agent-platform gemini endpoints also declare `providerId = google_ai_studio`, so the filter excluded them too. No gemini stream endpoint could ever match for a non-byok workspace.

Changes:
- Keep `google_ai_studio` in the provider filter and instead exclude endpoints whose `api` is the direct Google AI Studio API for non-byok workspaces. Agent-platform gemini endpoints stay eligible.
- Rename `EndpointConfig.providerApi` to `api` to match the endpoint constructors' static property — as written, a `providerApi` filter could never match a real endpoint.

Non-byok `gemini-3.1-pro-preview` still falls back after this: it has no agent-platform endpoint in the registry yet, only the direct google-ai-studio one (byok-only). Follow-up to add it.

## Tests

Added `lib/api/llm/index.test.ts`:
- a non-byok workspace resolves gemini-3.5-flash to agent-platform endpoints only, and never gets a direct google-ai-studio endpoint
- a byok workspace with a configured google credential still gets the direct google-ai-studio endpoint for gemini-3.1-pro

Updated `matches_where.test.ts` for the field rename. Ran the filter/sort/transition suites.

## Risk

Low.

## Deploy Plan

Deploy front.

### DIFF
--- a/front/lib/api/llm/index.test.ts
+++ b/front/lib/api/llm/index.test.ts
@@ -1,5 +1,5 @@
 import { getWorkspaceFilter } from "@app/lib/api/llm";
-import { Authenticator } from "@app/lib/auth";
+import { Authenticator, getFeatureFlags } from "@app/lib/auth";
 import { getStreamEndpoints } from "@app/lib/llms/stream";
 import type { WorkspaceConfig } from "@app/lib/llms/types/filter";
 import {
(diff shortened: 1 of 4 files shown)


front/lib/api/llm/index.test.ts
```diff
@@ -0,0 +1,61 @@
...
+} from "@app/lib/model_constructors/types/model_ids";
+import {
+  AGENT_PLATFORM_API,
+  GOOGLE_AI_STUDIO_API,
+} from "@app/lib/model_constructors/types/provider_apis";
+import { ProviderCredentialFactory } from "@app/tests/utils/ProviderCredentialFactory";
+import { WorkspaceFactory } from "@app/tests/utils/WorkspaceFactory";
+import { describe, expect, it } from "vitest";
+
+const workspaceConfig: WorkspaceConfig = {
```
flvndvd: Can we use the factory here, please?

flvndvd (approved):

PR #28390, opened 2026-07-02
log Anthropic pause_turn stop reason

flvndvd:
## Description

Anthropic ends a turn with a `pause_turn` stop reason when its server-side sampling loop hits the iteration limit while running server tools, which for us means tool search. The documented handling is to send the paused response back so the model can finish its turn. We currently don't do that: the agent loop sees no pending tool calls, finalizes the message, and the answer is silently truncated mid-work.

Before implementing the continuation, this adds a warning log whenever a `pause_turn` comes through the router, so we can measure how often the truncation actually happens in production and prioritize accordingly. No behavior change.


aubin-tchoi (approved):
LGTM

PR #28392, opened 2026-07-02
[front] tool credit cost: moved to internalMCP server and tool levels

thervier:
## Description

Replaces the hardcoded per-server billing maps in events.ts with toolCostCategory and freeUsage declared directly on every tool in each server's metadata, and sets the actual decided cost values for all tools.

* Before: A TOOL_CATEGORY_MAP (a ~75-entry exhaustive record in events.ts) assigned a single "basic" | "advanced" category to each internal MCP server. Free servers were tracked separately in a FREE_TOOL_SERVERS set. Adding a new server required updating two places in events.ts; per-tool pricing was impossible.
* After: toolCostCategory and freeUsage are mandatory on every tool individually in each server's metadata.ts. There are no server-level defaults and no fallback logic. getToolBillingInfo is a direct lookup on the tool entry. Adding a tool requires explicitly declaring its billing intent.

**Two orthogonal billing fields, mandatory at tool level**
* toolCostCategory: "basic" | "advanced" — the intrinsic pricing tier. 1 AWU for basic, 3 AWU for advanced.
* freeUsage: boolean — whether this tool's invocations are free regardless of category. Overrides the AWU charge to 0, while still recording the real tier in Metronome for analytics (so we can track what is being subsidised and switching from free to paid only requires flipping one field).

**Snapshot test for billing values**
A Vitest snapshot test (mcp_servers_metadata.test.ts) asserts toolCostCategory and freeUsage for every tool on every internal MCP server. Any accidental change to a tool's billing values or any new tool added without billing fields will break the snapshot, making regressions visible in CI.

**Changes**
* front/lib/api/mcp.ts: Renamed ToolCategory → ToolCostCategory, TOOL_CATEGORIES → TOOL_COST_CATEGORIES. Removed toolCategory and freeUsage from InternalMCPServerDefinitionType.
* front/lib/actions/mcp_internal_actions/tool_definition.ts: toolCostCategory: ToolCostCategory and freeUsage: boolean added as mandatory fields to ToolDefinition, ClientToolDefinition, and InternalMCPToolType.
* front/lib/metronome/events.ts: Deleted TOOL_CATEGORY_MAP, FREE_TOOL_SERVERS. getToolBillingInfo(serverName, toolName) replaces the old map lookup with a direct read from metadata. Renamed TOOL_CATEGORY_AWU_WEIGHTS → TOOL_COST_CATEGORY_AWU_WEIGHTS, isToolCategory → isToolCostCategory.
* ~77 servers/*/metadata.ts files: toolCostCategory and freeUsage added to every tool entry with explicit decided values (e.g. poke tools → basic/freeUsage: true, files/search/sandbox → advanced/freeUsage: false, formerly-free servers like agent_memory and toolsets → basic/freeUsage: true per tool).
* Consumers (per_user_usage.ts, per_api_key_usage.ts, programmatic_awu_usage.ts, awu_usage.ts, debug_user_awu_spend.ts, setup_new_pricing.ts): updated to renamed exports.

## Tests

 * npx tsgo --noEmit passes with 0 errors
 * credit_cost.test.ts passes — covers free-server (0 AWU), basic (1 AWU), advanced (3 AWU) cases, and the toolName threading
 * Tested locally with:
   * a free origin (sidekick) => all free => OK
   * a tool with freeUsage = true => tool usage = 0 => OK
   * a tool with freeUsage = false and basic or advanced => tool usage based on category => OK

## Risk

Medium, as it impacts the usage computation via credits

## Deploy Plan

Deploy front

### DIFF
--- a/front/lib/metronome/events.ts
+++ b/front/lib/metronome/events.ts
@@ -4,11 +4,7 @@ import {
   isInternalMCPServerName,
 } from "@app/lib/actions/mcp_internal_actions/constants";
 import type { InternalMCPToolType } from "@app/lib/actions/mcp_internal_actions/tool_definition";
-import {
-  type InternalMCPServerDefinitionType,
-  TOOL_CATEGORIES,
-  type ToolCategory,
-} from "@app/lib/api/mcp";
+import { TOOL_COST_CATEGORIES, type ToolCostCategory } from "@app/lib/api/mcp";
 import type { RunUsageType } from "@app/lib/resources/run_resource";
 import type { UserMessageOrigin } from "@app/types/assistant/conversation";
 import { createHash } from "crypto";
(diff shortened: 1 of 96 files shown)


tdraier (approved):

front/lib/metronome/events.ts
```diff
@@ -65,106 +71,29 @@ export const TOOL_CATEGORY_AWU_WEIGHTS: Record<ToolCategory, number> = {
...
+  const metadata = getInternalMCPServerMetadata(serverName);
+  // Both casts are safe: each metadata file satisfies ServerMetadata which uses InternalMCPToolType
+  // and InternalMCPServerDefinitionType. The narrow as-const type drops optional fields absent from
+  // that specific object, so we cast to access them uniformly.
+  const serverInfo = metadata.serverInfo as InternalMCPServerDefinitionType;
+  const tool = (metadata.tools as InternalMCPToolType[]).find(
+    (t) => t.name === toolName
+  );
+  return {
+    toolCategory: tool?.toolCategory ?? serverInfo.toolCategory,
```
rfrenoy: today `tool?.toolCategory` is always undefined, is that expected? I think we need to update `createToolsRecord` to take an optional toolCategory?

rfrenoy (commented):
Currently my understanding is `tool?.toolCategory` is always undefined (see comment).  

It is unclear to me that once defined, it will properly populate to `toolAwuFromAction`, which is what is effectively used to send data to Metronome and to Elastic Search. So there is a risk we define these tool categories but effectively keep on billing the server info. I think we should update the `tools` array in the servers metadata, and add a test to make sure `toolAwuFromAction` is correct with the new info

tdraier (approved):

PR #28394, opened 2026-07-02
[front] poke: add optional forceSeatType to SwitchContractDialog

tdraier:
## Description

Exposes the switch-contract seat-type override in Poke: an optional **"Force seat type"** selector (billable seat types + a "— No override —" default) in the seats section of `SwitchContractDialog`. On submit it sends `promoteNoneSeatsTo` when a valid seat is chosen, which forces every seat-less (`none`) member onto that seat on the new contract (preempting committed-seat placement).

Stacked on #28393, which adds the `promoteNoneSeatsTo` backend field.

Off by default (empty = default seat assignment), so existing switches are unaffected.

## Tests

`tsgo` clean; manual via the Poke subscriptions dialog.

## Risk

Low — UI-only, optional field.

## Deploy Plan

Merge after #28393.

### DIFF
--- a/front/components/poke/subscriptions/SwitchContractDialog.tsx
+++ b/front/components/poke/subscriptions/SwitchContractDialog.tsx
@@ -18,6 +18,10 @@ import {
   usePokeStripeCustomerCurrency,
 } from "@app/lib/swr/poke";
 import { usePokePluginAsyncArgs } from "@app/poke/swr/plugins";
+import {
+  BILLABLE_SEAT_TYPES,
+  isMembershipSeatType,
+} from "@app/types/memberships";
 import { isCreditPricedPlan } from "@app/types/plan";
 import { CreditUsageConfigurationSchema } from "@app/types/poke/credit_usage_configuration";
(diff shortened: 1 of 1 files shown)


front/components/poke/subscriptions/SwitchContractDialog.tsx
```diff
@@ -249,6 +258,7 @@ export default function SwitchContractDialog({
       autoSeatUpgradeEnabled: false,
       topUpEnabled: false,
       autoInvoiceFinalizationEnabled: true,
+      forceSeatType: "",
```
rfrenoy, line 261: nit: could we have a real seat type as default? I think it would make the logic clearer

rfrenoy (approved):
LGTM with one nit comment

tdraier:

- #28394 ◀

front/components/poke/subscriptions/SwitchContractDialog.tsx
tdraier: by default we want use the regular committed-seats behaviour  ?

PR #28397, opened 2026-07-02
Fix files edit tool after the tool context rename merge race

flvndvd:
## Description

The path-keyed edit tool PR and the tool context rename merged the same day, and since the edit tool files were brand new they landed verbatim with the pre-rename harness types. Main currently fails type checking on them and the edit tool handler cannot resolve its conversation, which also fails its test suite. This renames the context accesses to the new shape, nothing else changes.

## Risk

None, mechanical rename aligning two already-merged changes.

## Deploy Plan

Deploy front.


aubin-tchoi (approved):
LGTM

PR #28399, opened 2026-07-02
Pricing: refresh the limit after bulk edit

fabiencelier:
## Description

Closes https://github.com/dust-tt/tasks/issues/9338

Make the endpoint synchronous.
I think it is is reasonnable as updating the limit should not take too long even with many users.

Alternative would be to have a poll mecanism on the workflow status which adds complexity

https://github.com/user-attachments/assets/d907cbde-6b86-491d-8ea2-9252f3ebe8b8

## Tests

Tested locally
Updated unit tests

## Risk

May make a long request if we update many users at once, but should be reasonnably fast.

## Deploy Plan

deploy front


rfrenoy (approved):
Looking nice! 🔥

PR #28403, opened 2026-07-02
Steer the model away from mixing tool searches with tool calls

flvndvd:
## Description

When the model issues a tool search and a regular tool call in the same turn, Anthropic ends the turn on the tool call and never runs the search. Replaying that un-run search block on the next request is fragile: the API only resumes it when the follow-up message contains nothing but tool results, which our rendering does not guarantee, and rejects the whole request otherwise. This is what has been killing agent turns right after a skill gets enabled.

This adds one sentence to the tool search instruction telling the model to search first, wait for the results, then call tools. It shrinks how often the fragile shape forms and keeps searches effective instead of wasted. Will assess if it keeps happening and might send another PR to parse messages before sending them.


aubin-tchoi (approved):
LGTM

PR #28404, opened 2026-07-02
[front] refactor: cleanup `SkillResource` method naming

aubin-tchoi:
## Description

This PR refactors the interface of `SkillResource` around a few main (pre-existing) design principles:
- `baseFetch` filters on active skills by default, making it implied in any use case that does not require archived skills (meaning that when you need the archived skills, you explicitly know that you need to list archived skills, rest of the time it's always active skills).
- The main entry points are the same regardless of the context, meaning that `fetchActiveByIdsForAgentLoop` is redundant with `fetchByIds`.

## Tests

- Tested locally.

## Risk

- Low.

## Deploy Plan

- Deploy front.


xebeohp (approved):
LGTM!

PR #28405, opened 2026-07-02
db: add activation_recommendations table

frankaloia:
## Summary

Main details on where this will be used is in [S1: Activation Skill](https://app.notion.com/p/dust-tt/S1-Activation-Skill-38628599d94181a5a4f8f94b47b139f0?source=copy_link)

* storing recommendation history so they can be surfaced to the agent in later runs to improve the quality of user recommendations + provide context on what the user already acted on
* Will be used in [S3: Measurement Loop, Metrics, & Dashboards](https://app.notion.com/p/dust-tt/S3-Measurement-Loop-Metrics-Dashboards-38628599d9418132aab3ef6ec99f1083?source=copy_link) to measure internally (1) are people acting on these recommendations (2) what recommendations are leading to the best outcomes. We  might need more metadata in the future, but started relatively simple and just planning to store references of skills/triggers created as a result of the user interaction

### DB schema (`pre-deploy` migration)
- Table with `workspaceId`, `userId`, `status`, `content`, `rationale`, and three nullable artifact FKs (`conversationModelId`, `skillModelId`, `triggerModelId`)
- `ON DELETE RESTRICT` on `workspaceId`/`userId` — a recommendation must not outlive the workspace or user
- `ON DELETE SET NULL` on the artifact FKs — a recommendation survives when a conversation/skill/trigger is deleted; the FK is nulled rather than cascading the delete, avoiding the loss of a recommendation just because the artifact that motivated it was cleaned up
- Every FK column is indexed to avoid table scans on parent-row deletion (BACK13)
- No `sId`: recommendations are internal state, never exposed in URLs or public API (SEC2)

## Testing
1. Ran locally

## Deploy
1. Pre deploy hook should handle migration deployment
2. Deploy front

### DIFF
--- a/front/lib/models/activation/activation_recommendation.ts
+++ b/front/lib/models/activation/activation_recommendation.ts
@@ -1,0 +1,144 @@
+import { ConversationModel } from "@app/lib/models/agent/conversation";
+import { TriggerModel } from "@app/lib/models/agent/triggers/triggers";
+import { SkillConfigurationModel } from "@app/lib/models/skill";
+import { frontSequelize } from "@app/lib/resources/storage";
+import { DANGEROUSLY_UNBOUNDED_TEXT, DataTypes } from "@app/lib/resources/storage/data_types";
+import { UserModel } from "@app/lib/resources/storage/models/user";
+import { WorkspaceAwareModel } from "@app/lib/resources/storage/wrappers/workspace_models";
+import type { CreationOptional, ForeignKey, NonAttribute } from "sequelize";
+
+export type ActivationRecommendationStatus =
+  | "pending"
+  | "accepted"
+  | "rejected"
+  | "saved"
+  | "recurring";
+
+export class ActivationRecommendationModel extends WorkspaceAwareModel<ActivationRecommendationModel> {
+  declare createdAt: CreationOptional<Date>;
+  declare updatedAt: CreationOptional<Date>;
+
+  declare userId: ForeignKey<UserModel["id"]>;
+  declare status: ActivationRecommendationStatus;
+  declare content: string;
+  declare rationale: string;
+
+  declare conversationModelId: ForeignKey<ConversationModel["id"]> | null;
+  declare skillModelId: ForeignKey<SkillConfigurationModel["id"]> | null;
+  declare triggerModelId: ForeignKey<TriggerModel["id"]> | null;
+
+  declare user: NonAttribute<UserModel>;
+  declare conversation: NonAttribute<ConversationModel | null>;
+  declare skillConfiguration: NonAttribute<SkillConfigurationModel | null>;
+  declare trigger: NonAttribute<TriggerModel | null>;
+}
+
+ActivationRecommendationModel.init(
+  {
+    createdAt: {
+      type: DataTypes.DATE,
+      allowNull: false,
+      defaultValue: DataTypes.NOW,
+    },
+    updatedAt: {
+      type: DataTypes.DATE,
+      allowNull: false,
+      defaultValue: DataTypes.NOW,
+    },
+    userId: {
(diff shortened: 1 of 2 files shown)


id13 (approved):

front/lib/models/activation/activation_recommendation.ts
```diff
@@ -0,0 +1,144 @@
...
+  declare updatedAt: CreationOptional<Date>;
+
+  declare userId: ForeignKey<UserModel["id"]>;
+  declare status: ActivationRecommendationStatus;
+  declare content: string;
+  declare rationale: string;
+
+  declare conversationModelId: ForeignKey<ConversationModel["id"]> | null;
+  declare skillModelId: ForeignKey<SkillConfigurationModel["id"]> | null;
+  declare triggerModelId: ForeignKey<TriggerModel["id"]> | null;
```
aubin-tchoi: not consistent in naming with `userId` and what we do elsewhere

front/lib/models/activation/activation_recommendation.ts
```diff
@@ -0,0 +1,144 @@
...
+
+  declare user: NonAttribute<UserModel>;
+  declare conversation: NonAttribute<ConversationModel | null>;
+  declare skillConfiguration: NonAttribute<SkillConfigurationModel | null>;
+  declare trigger: NonAttribute<TriggerModel | null>;
```
aubin-tchoi: will you need this? (ideally separate queries rather than includes)

front/lib/models/activation/activation_recommendation.ts
```diff
@@ -0,0 +1,144 @@
...
+      defaultValue: DataTypes.NOW,
+    },
+    userId: {
+      type: DataTypes.BIGINT,
+      allowNull: false,
+    },
+    status: {
+      type: DataTypes.STRING,
+      allowNull: false,
+      defaultValue: "pending",
```
aubin-tchoi: do you want a default value in db rather than on the application side?

front/lib/models/activation/activation_recommendation.ts
```diff
@@ -0,0 +1,144 @@
...
+});
+
+ActivationRecommendationModel.belongsTo(SkillConfigurationModel, {
+  foreignKey: { name: "skillModelId", allowNull: true },
+  onDelete: "SET NULL",
+  as: "skillConfiguration",
+});
+SkillConfigurationModel.hasMany(ActivationRecommendationModel, {
+  foreignKey: { name: "skillModelId", allowNull: true },
+  onDelete: "SET NULL",
```
aubin-tchoi, line 129: don't you want to restrict instead? I would assume that `skillId` null means that the recommendation was not about a skill, which would be mutated involuntarily on skill deletion here

front/migrations/pre-deploy/20260701100000_create_activation_recommendations.sql
```diff
@@ -0,0 +1,165 @@
...
+);
+
+/*
+Statement 2
+*/
+SET SESSION statement_timeout = 3000;
+SET SESSION lock_timeout = 3000;
+ALTER TABLE "public"."activation_recommendations"
+	ADD CONSTRAINT "activation_recommendations_status_check"
+	CHECK (status IN ('pending', 'accepted', 'rejected', 'saved', 'recurring'));
```
aubin-tchoi: debatable but wouldn't add this db constraint and handle application side instead (via typing for instance)

front/lib/models/activation/activation_recommendation.ts
frankaloia: Fair push back, I am going to leave this out for now given the uncertainty of if they will def be needed

PR #28407, opened 2026-07-02
fix(conversation): stop inline-step duplication from remount replay

rfrenoy:
## Description

Tickets:
* https://github.com/dust-tt/tasks/issues/9287
* https://github.com/dust-tt/tasks/issues/8022

Make inline steps idempotent instead of fighting the replay. Step ids are now derived from the triggering event id (stable across replays) rather than Date.now(), and appendThinkingStep/appendContentStep skip on insert if a step with that id already exists.

This complements the earlier retry-collapse fix in [PR28351](https://github.com/dust-tt/dust/pull/28351).

## Tests

Local + unit

## Risk

Can eventually break UX if there are unexpected side effects. Did not see any issue while testing locally

## Deploy Plan

Front

### DIFF
--- a/front/hooks/useAgentMessageStream.test.ts
+++ b/front/hooks/useAgentMessageStream.test.ts
@@ -817,7 +817,7 @@ describe("useAgentMessageStream", () => {
       );
     });
 
-    // All three attempts are the SAME step (Temporal retries), so only the final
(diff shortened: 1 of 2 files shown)


PopDaph (approved):
LGTM thanks vm 🙏🏻 🤞🏻

PR #28408, opened 2026-07-02
feat(llm): route the noop model through the new LLM router

adrsimon:
## Description

Registers the `noop` model as a first-class endpoint in the new per-endpoint LLM router so it is served through `use_new_llm_router` instead of always falling back to the legacy per-provider router.

`noop` is unusual: it has no external provider API, and its usefulness comes from two behaviors the generic `StreamEndpointTransition` cannot express:

- **`staticResponse`** — the `sidekick` and `dust` global agents run on `NOOP_MODEL_CONFIG` and pass `metaData: { staticResponse }` to emit canned replies (new-agent-from-scratch first turn, skill-suggestion echo). The generic transition forwards only the conversation payload + parsed config, dropping `metaData`.

- **`consume $X`** — records an explicit simulated credit cost. The new router derives cost from `token_usage × tokenPricing`, so a flat dollar amount can't be represented that way.

A dedicated `NoopStreamTransition` preserves both: it injects the static response into the noop request and records the simulated cost through the base `getSimulatedRunUsages` hook. So turning the flag on does not regress the global-agent static replies.

## Tests

- `npx tsgo --noEmit` — clean.
- `lib/api/llm/transitionLLM.test.ts` — 12/12 pass.
- Biome + typecheck pre-commit hooks pass.
- `router_parity.test.ts` excludes `noop` (no provider SDK request to diff against) — it synthesizes its stream in-process.
- Live test

## Risk

Low.

## Deploy Plan

Standard deploy.

### DIFF
--- a/front/lib/api/llm/index.ts
+++ b/front/lib/api/llm/index.ts
@@ -20,6 +20,7 @@ import { isXaiWhitelistedModelId } from "@app/lib/api/llm/clients/xai/types";
 import type { LLM } from "@app/lib/api/llm/llm";
 import {
   BatchEndpointTransition,
+  NoopStreamTransition,
   StreamEndpointTransition,
 } from "@app/lib/api/llm/transitionLLM";
 import type { LLMParameters } from "@app/lib/api/llm/types/options";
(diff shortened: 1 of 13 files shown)


davidebbo:
Yeah, sorry, noop is a bit of an oddball, and I contributed to that. But it is useful :)

davidebbo (approved):
LGTM

PR #28409, opened 2026-07-02
feat(conversations): add visible/hidden visibilities with backward-compatible readers and backfill

ElPicador:
## Description

First phase of renaming `ConversationVisibility` values: `unlisted` → `visible` and `test` → `hidden`. This is the backward-compatible "expand" step, so nothing is renamed on the wire yet.

- The `ConversationVisibility` union now carries both the old and the new values, with transitional `LISTED_CONVERSATION_VISIBILITIES` / `HIDDEN_CONVERSATION_VISIBILITIES` helpers.
- All conversation readers/queries tolerate a mixed dataset (`visibility IN (unlisted, visible)`, exclude `(test, hidden)`), so old rows and future rows both resolve correctly.
- The private POST conversations schema accepts both old and new values.
- The public API boundary keeps the legacy wire names: `toPublicConversationVisibility` maps `visible` → `unlisted` and `hidden` → `test`, so the published `@dust-tt/client` SDK is unaffected.
- Adds a batched backfill migration to rename existing rows (`unlisted` → `visible`, `test` → `hidden`) per workspace.

Writers still emit the old values in this PR — flipping the writers and dropping the old union members come in follow-up PRs.

## Tests

- Added a `fetchByModelIds` test that seeds all five visibilities and asserts the reader tolerates the mix: `deleted` is always excluded, and `excludeTest` drops both `test` and `hidden` while keeping both `unlisted` and `visible`.
- Backfill migration runs dry-run by default; execute with `--execute` (optionally `--workspaceId`).

## Deploy Plan

Deploy this PR first (readers become tolerant), then run the backfill migration to rename existing rows. Only after that do the follow-up PRs flip the writers and remove the legacy values.

### DIFF
--- a/front/lib/api/v1/backward_compatibility.ts
+++ b/front/lib/api/v1/backward_compatibility.ts
@@ -53,6 +53,29 @@ export function normalizeConversationVisibility(
 }
 
 /**
+ * Translates an internal conversation visibility to the value exposed on the
+ * public API. During the "unlisted" -> "visible" / "test" -> "hidden" rename we
+ * keep the public wire contract on the legacy names so existing SDK clients are
+ * unaffected: "visible" is reported as "unlisted" and "hidden" as "test".
+ */
+export function toPublicConversationVisibility(
+  visibility: ConversationVisibility
+): ConversationPublicType["visibility"] {
+  switch (visibility) {
+    case "unlisted":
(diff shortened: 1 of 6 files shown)


ElPicador:

- #28409 ◀
    - #28419

PR #28411, opened 2026-07-02
feat: add Claude Fable 5 to new LLM router

adrsimon:
## Description

Adds Claude Fable 5 via EAP KEY to the new stream endpoint router, including model-constructor metadata, the Anthropic global stream endpoint, Dust endpoint metadata, registry wiring, and endpoint coverage.

Also applies the existing EAP Anthropic credential routing to new stream and batch router transitions so models configured with `useEapKey` use the dedicated key consistently.

## Tests

- Pre-commit hook: secret scan, front typecheck, and Biome passed.
- `NODE_ENV=test npm run test -- --config lib/model_constructors/test/vite.config.js lib/model_constructors/test/endpoints/anthropic_claude_fable_five.test.ts` (imports cleanly; live cases skipped without `RUN_LLM_TEST`).
- `npx tsgo --noEmit 2>&1 | grep -v node_modules` produced no output.

## Risk

Low to moderate. The new endpoint is feature-gated behind `claude_fable_5_feature`, uses the existing Anthropic stream client, and keeps Fable 5 aligned with the existing 250k context / 64k output rollout values. 

## Deploy Plan

Deploy normally.

### DIFF
--- a/front/lib/api/llm/index.ts
+++ b/front/lib/api/llm/index.ts
@@ -391,7 +391,16 @@ function getStreamEndpointLLM(
     return null;
   }
 
-  return new StreamEndpointTransition(auth, llmParameters, endpoint);
+  const modelConfig = getModelConfigByModelId(llmParameters.modelId);
+  const credentials = modelConfig?.useEapKey
+    ? withEapAnthropicKey(llmParameters.modelId, llmParameters.credentials)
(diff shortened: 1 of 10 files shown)


aubin-tchoi (approved):
LGTM, as per IRL it's only for us (no ZDR + pricing topic)

PR #28413, opened 2026-07-02
Deprecate the file-id Frame edit tool in favor of editing the source and publishing

flvndvd:
## Description

Frames are now updated by editing their source files by path and publishing. This deprecates the tool that edits a Frame through its file id: conversations with the conversation file system no longer expose it, while conversations created before the file system defaulted on keep it since they have no path-based alternative. The tool that reads a Frame's content by file id stays available everywhere, though: it reads the canonical file directly rather than through the mount, so it remains the only way to reach a Frame whose mount path hasn't been resolved yet, such as one created before the mount system existed and not yet covered by the backfill below.

The Frames skill instructions teach the new flow. With the Computer available the model edits the mounted source, without it the model edits through the files tools, and legacy conversations keep the retrieve-then-edit guidance. Across all three, the instructions and the relevant tool descriptions now also steer the model away from recreating a Frame it was only asked to change, since doing so loses the Frame's identity and share URL and burns tokens regenerating content that didn't need to change.

The rejection message shown when another tool touches a Frame file now points at the source-edit and publish flow as well.

A backfill script gives Frames created before the mount dual-write a mount path by copying their canonical content, so publishing and path-based editing work for them too.

Reverting a Frame now also refreshes its mount copy from the restored version, so the revert is visible to a source read and isn't silently undone by the next publish. This refresh is best-effort: a failure here still lets the revert succeed and is only logged, since the canonical restore has already happened by that point.

## Risk

Low. Agents in file-system conversations lose the file-id edit tool on deploy; the toolset is recomputed at each step and the updated instructions and rejection messages redirect to the source-edit and publish flow. The file-id read tool is unaffected everywhere, and legacy conversations are unaffected. The revert mount refresh is a data-integrity fix with no change to revert's success or failure contract.

## Deploy Plan

Deploy front, then run the backfill script (dry run first, then with execute).

### DIFF
--- a/front/lib/api/actions/servers/files/metadata.ts
+++ b/front/lib/api/actions/servers/files/metadata.ts
@@ -4,6 +4,7 @@ import { getPrefixedToolName } from "@app/lib/actions/tool_name_utils";
 import {
   CREATE_INTERACTIVE_CONTENT_FILE_TOOL_NAME,
   INTERACTIVE_CONTENT_SERVER_NAME,
+  PUBLISH_INTERACTIVE_CONTENT_FILE_TOOL_NAME,
 } from "@app/lib/api/actions/servers/interactive_content/metadata";
 import { FILE_PREVIEW_DIRECTIVE_EXAMPLE } from "@app/lib/markdown/file_preview";
 import { frameContentType, frameSlideshowContentType } from "@app/types/files";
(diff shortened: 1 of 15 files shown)


front/lib/api/actions/servers/files/metadata.ts
```diff
@@ -4,6 +4,7 @@ import { getPrefixedToolName } from "@app/lib/actions/tool_name_utils";
 import {
```
flvndvd, line 4: Instructions are a bit heavy as we have to cover 3 cases (computer-first, file-first, legacy). Will clean up once backfill has been run.

front/lib/api/actions/servers/files/tools/edit.ts
```diff
@@ -104,9 +109,14 @@ export async function editHandler(
 
   if (occurrences === 0) {
     return new Err(
-      new MCPError(`String not found in file: "${old_string}"`, {
-        tracked: false,
-      })
+      new MCPError(
+        `String not found in file: "${old_string}". The file may have changed since you last ` +
```
aubin-tchoi: nit: the current wording makes it sound like old_string is the name of the file rather than the string to replace, maybe `String ${old_string} not found in file"`

aubin-tchoi (approved):
LGTM

front/lib/api/actions/servers/files/tools/edit.ts
flvndvd: Good catch.

PR #28414, opened 2026-07-02
fix(front): focus composer when clicking "New" on the home page

rfrenoy:
## Description

Ticket: https://github.com/dust-tt/tasks/issues/9376

Clicking "New" while already on the new-conversation page did not focus the input bar. Two issues:

1. handleNewClick read the conversation id from `router.query.cId`, but in the react-router SPA the id is a path param (usePathParam), never present in `router.query`. So `isNewConversation` was always false and setAnimate(true) was never called. Derive it from useActiveConversationId (null on "new").

2. The `animate` focus flag was never reset, so once true (e.g. after agent selection) later setAnimate(true) calls were swallowed as no-ops. Consume the flag after focusing so each request re-triggers the effect.
## Tests

Local

## Risk

Low
## Deploy Plan

Front

### DIFF
--- a/front/components/assistant/conversation/input_bar/InputBarContainer.tsx
+++ b/front/components/assistant/conversation/input_bar/InputBarContainer.tsx
@@ -1130,9 +1130,10 @@ const InputBarContainer = ({
     isSpacesLoading,
   ]);
 
-  // When input bar animation is requested, it means the new button was clicked (removing focus from
-  // the input bar), we grab it back.
-  const { animate, setAnimate, captureActions } = useContext(InputBarContext);
+  // When an input bar refocus is requested (e.g. the "New" button was clicked, removing focus from
+  // the input bar), we grab focus back.
+  const { shouldFocusInput, setShouldFocusInput, captureActions } =
+    useContext(InputBarContext);
 
   const handleSingleAgentSelect = useCallback(
     (mention: RichMention) => {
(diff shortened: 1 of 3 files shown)


front/components/assistant/conversation/input_bar/InputBarContainer.tsx
```diff
@@ -1183,8 +1183,11 @@ const InputBarContainer = ({
     if (animate) {
       // Schedule focus to avoid flushing during render lifecycle.
       queueMicrotask(() => editorService.focusEnd());
+      // Consume the flag so a later request (e.g. clicking "New" again) re-runs
+      // this effect instead of being swallowed as a no-op state update.
+      setAnimate(false);
```
ykmsd: I'm pretty sure we can kill this animate (I think we removed the animation effect from the input bar) 🤔

ykmsd (approved):

front/components/assistant/conversation/input_bar/InputBarContainer.tsx
ykmsd: hm but this is needed for refocus??

rfrenoy: Yes. Added [this commit](https://github.com/dust-tt/dust/pull/28414/changes/a9622699378306c0d36dbeae65c5dc0846861011) to rename `animate` to `shouldFocusInput`

PR #28415, opened 2026-07-02
feat(global-agents): fall back to Sonnet 4.6 for claude-sonnet in EU-only workspaces

adrsimon:
## Description

The `claude-sonnet` global agent (`CLAUDE_5_SONNET`) hardcoded Claude Sonnet 5 as its model. Sonnet 5 declares `regionalAvailability: { "europe-west1": false }`, so for EU-models-only workspaces the agent pointed at a model that is not enabled in the region.

This changes the agent to serve **Sonnet 5** by default and fall back to **Sonnet 4.6** when Sonnet 5 is not enabled for the workspace. It reuses the existing `selectEnabledModel` picker (already used by the deep-dive / dust / sidekick global agents), which runs the same `isModelEnabled` predicate enforced at message-post time — so the selected model can never be rejected later as "not supported". EU-only workspaces (`regionalModelsOnly` + `europe-west1`) fall through to Sonnet 4.6 automatically; everyone else gets Sonnet 5.

`featureFlags` is threaded into the builder so the availability check uses the workspace's real flags, matching the other global-agent builders.

## Tests

- `npx tsgo --noEmit` passes.
- biome + front typecheck pass via the pre-commit hook.

## Risk

Low. The change is scoped to the model selection for a single global agent. Non-EU workspaces keep the same model (Sonnet 5); only EU-models-only workspaces change (from an unavailable Sonnet 5 to the available Sonnet 4.6, which is the intended fix).

## Deploy Plan

Standard `front` deploy. No migrations or config changes required.

### DIFF
--- a/front/lib/api/assistant/global_agents/configurations/anthropic.ts
+++ b/front/lib/api/assistant/global_agents/configurations/anthropic.ts
@@ -339,7 +339,7 @@ export function _getClaude5SonnetGlobalAgent({
     versionCreatedAt: null,
     versionAuthorId: null,
     name: metadata.name,
-    description: metadata.description,
+    description: modelConfig.description,
     instructions: `${globalAgentGuidelines}\n${globalAgentWebSearchGuidelines}`,
     instructionsHtml: null,
     pictureUrl: metadata.pictureUrl,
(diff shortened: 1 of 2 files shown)


front/lib/api/assistant/global_agents/configurations/anthropic.ts
```diff
@@ -316,26 +320,37 @@ export function _getClaude5SonnetGlobalAgent({
...
+      [
+        CLAUDE_SONNET_5_DEFAULT_MODEL_CONFIG,
+        CLAUDE_SONNET_4_6_DEFAULT_MODEL_CONFIG,
+      ],
+      { featureFlags }
+    ) ?? CLAUDE_SONNET_4_6_DEFAULT_MODEL_CONFIG;
+
   return {
     id: -1,
     sId,
```
aubin-tchoi, line 337: we should also swap the `sId` no?

aubin-tchoi (commented):
LGTM

PR #27382, opened 2026-06-15
Fix forked convos so they show the correct title

davidebbo:
## Description

  When branching a conversation, the sidebar optimistically shows a `Branched from '…'` title, but after a page refresh it reverted to **New Conversation**.

  ### Root cause

  A branched conversation is created with `title: null` — the `Branched from '…'` label is *derived*, not stored. `getConversationDisplayTitle()` derives it from the conversation's `forkingData`, and `toListItem()` relies on that.

  The Elasticsearch-backed listing path loads `forkingData`, but the **DB listing path** (`fetchPrivateConversationsPaginated`, the fallback used when the `conversation_search_read` flag is off — e.g. local/dev) fetched conversations *without* `includeForkingData`. So
  `forkingData` was `undefined`, and the title derivation fell through to the date-based "New Conversation" fallback.

  ### Fix

  Pass `{ includeForkingData: true }` to `baseFetchWithAuthorization` in `fetchPrivateConversationsPaginated`, so `toListItem()` can derive the branched title — consistent with the ES path and the per-conversation fetch path. This is a single `hasOne` join over the
  bounded page (not an N+1).

  ## Testing

  Added a regression test in `conversation_resource.test.ts` that forks an untitled child, lists via `listPrivateConversationsForUserPaginatedFromDB`, and asserts the item's title is `Branched from '<parent title>'`. Confirmed it **fails without the fix** and **passes with it**.

PR #28417, opened 2026-07-02
Revert forkingData load in DB conversation list (revert #27382)

davidebbo:
## Description

Reverts #27382.

#27382 passed `{ includeForkingData: true }` to `baseFetchWithAuthorization` in the DB-backed private conversation listing (`fetchPrivateConversationsPaginated`) so that untitled forked conversations would display their derived `Branched from '...'` title, consistent with the Elasticsearch listing path.

The problem: loading fork data turns out to be **costly when grabbing a list of conversations**. On large lists, the query plan appears to degrade significantly (the extra join is not the cheap bounded `hasOne` it looked like in isolation).

The bug #27382 originally fixed was minor — in the DB listing path (the fallback used when the `conversation_search_read` flag is off, e.g. local/dev), untitled forked conversations fell back to the date-based "New Conversation" title instead of "Branched from '...'". Given the cost, it's better to revert this fix for now rather than pay the query-plan penalty on every conversation list.

We can revisit later by optimizing the query if we decide restoring the derived title in the DB path is worth it.

## Changes

- Revert `{ includeForkingData: true }` back to `{}` in `fetchPrivateConversationsPaginated`.
- Remove the regression test that asserted the derived "Branched from '...'" title in the DB paginated list.

## Risk

Low. This restores the prior behavior; the only user-visible effect is that untitled forked conversations in the DB listing path again fall back to the default title. Safe to roll back.


philipperolet (approved):
LGTM

PR #28419, opened 2026-07-02
feat(conversations): flip conversation visibility writers to visible/hidden

ElPicador:
## Description

Second phase of renaming `ConversationVisibility` values: flips every internal writer from `unlisted`/`test` to `visible`/`hidden`. Stacked on #28409, which made all readers tolerant of both old and new values — this PR is safe to deploy on top of it.

- `ConversationModel`'s default value and all direct `createConversation` / `ConversationResource.makeNew` call sites (onboarding, email trigger, project tasks, temporal activities, reinforcement, MCP tool servers, sidekick/preview conversations, batch LLM, dev seed scripts) now write `visible` / `hidden`.
- `ConversationResource.updateVisibilityToUnlisted` renamed to `updateVisibilityToVisible` and updated accordingly (only caller: the poke "restore conversation" plugin).
- The public API write path is the one place that needed real logic, not just a literal swap: `normalizeConversationVisibility` (called from the public v1 POST conversations handler) now maps the legacy wire values it accepts forever (`workspace`, `unlisted`, `test`) onto the new internal storage values (`visible`, `hidden`), while the read-side `toPublicConversationVisibility` (added in #28409) keeps reporting the old names back to callers. The public wire contract is unchanged.
- Server-to-server calls that go through the public API SDK (the `run_agent` MCP tool's `DustAPI.createConversation`, and the Slack/Discord/Teams connectors) intentionally keep sending the legacy wire value `unlisted` — it gets normalized on the way in, same as any other API client.
- Test factory defaults (`ConversationFactory`, `mockConversation`) also flipped so new test data matches production's new default.

## Tests

- Added two functional tests on `POST /api/v1/w/[wId]/assistant/conversations` asserting the full round trip: sending the legacy `test`/default wire values stores `hidden`/`visible` internally while the response still reports the legacy `test`/`unlisted` values, so existing SDK clients see no change.
- Full existing suite for conversations, reinforcement, temporal triggers/labs activities, and batch LLM re-run and green after the flip.

## Deploy Plan

Deploy after #28409 is live and the backfill migration from that PR has been run (or at least deployed — readers tolerate both values regardless of ordering). No further migration needed here; this only changes what new writes look like.

### DIFF
--- a/front-api/routes/v1/w/[wId]/assistant/conversations/index.test.ts
+++ b/front-api/routes/v1/w/[wId]/assistant/conversations/index.test.ts
@@ -1,6 +1,7 @@
+import { ConversationResource } from "@app/lib/resources/conversation_resource";
 import { createPublicApiMockRequest } from "@app/tests/utils/generic_public_api_tests";
 import { honoApp } from "@front-api/app";
-import { describe, expect, it, vi } from "vitest";
+import { assert, describe, expect, it, vi } from "vitest";
 
 vi.mock("@app/lib/api/programmatic_usage/tracking", () => ({
   isProgrammaticUsage: () => false,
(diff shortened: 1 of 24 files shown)


ElPicador:

- #28409
    - #28419 ◀

PR #28420, opened 2026-07-02
Retrieve agent memories on demand instead of injecting them into the prompt

flvndvd:
## Description

Agent memories were injected into the system prompt on every request to the dust agents. Memories are per-user and sat ahead of everything shared in the prompt cache prefix, so two users of the same workspace always diverged before the largest common part of the prompt, and any memory write mid-conversation invalidated the cached prefix. With this change the system prompt no longer contains memories at all and becomes identical across users of a workspace, which is a prerequisite for sharing the cached prefix between them.

The model is instead instructed to fetch memories with the memory tool when prior context about the user is likely to change the answer.

Might be revisited later.

### DIFF
--- a/front/lib/api/assistant/global_agents/configurations/dust/dust.ts
+++ b/front/lib/api/assistant/global_agents/configurations/dust/dust.ts
@@ -31,9 +31,7 @@ import {
   isDustCompanyPlan,
   isEnterprisePlanPrefix,
 } from "@app/lib/plans/plan_codes";
-import type { AgentMemoryResource } from "@app/lib/resources/agent_memory_resource";
 import type { MCPServerViewResource } from "@app/lib/resources/mcp_server_view_resource";
-import { formatTimestampToFriendlyDate } from "@app/lib/utils";
 import type {
   AgentConfigurationType,
   AgentModelConfigurationType,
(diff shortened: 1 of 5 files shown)


front/lib/api/assistant/global_agents/configurations/dust/dust.ts
```diff
@@ -136,8 +134,7 @@ Do not enable it for generic help requests, non-Dust products, or ambiguous ment
 You have access to a persistent, user-specific memory system. Each user has their own private memory store.
 
 <critical_behavior>
-Existing memories are critical to your success. Always use them to tailor your responses and improve your performance over time.
-They are available to you directly in the <existing_memories> section.
+Memories are not included in your context automatically. Retrieve them with the \`agent_memory\` tool when prior context about the user is likely to change your answer: recurring workflows, personal preferences, ongoing projects, or requests that assume context you don't have. Do not retrieve memories for self-contained requests that any user would want answered the same way.
```
aubin-tchoi: nit: I don't think we need to mention that memories are not here (no expectation for them to be)

aubin-tchoi (approved):
LGTM

PR #28422, opened 2026-07-02
feat(activation): add user_analytics MCP server and data context [S1-1]

frankaloia:
## Summary
* https://github.com/dust-tt/tasks/issues/9353
* Adding data required to activation skill in order to make a decision
* Using fetchInstructions to include user type + available tools as this will be needed every loop and does not add any new dependency on hot path
* Adding new mcp server to get user analytics: top usage by this user + top usage in workspace for agents/skills the user has access to. This is just a raw rankings of the agents/skills. We did not use the existing workspace_analytics server as this is restricted to admins-only as it includes more sensitive information

## Testing

Screenshot 2026-07-02 at 2 14 54 PM

## Risk

Low - not customer facing yet

## Deploy
* Deploy front


front/lib/actions/mcp_internal_actions/internal_mcp_server_availability.snapshot.json
```diff
@@ -1,83 +1,318 @@
 {
   "auto": [
-    { "name": "image_generation", "id": 2 },
```
adrsimon, line 3: any idea why it was reformated ? seems odd but whatever

adrsimon (approved):

PR #28423, opened 2026-07-02
fix(conv rendering): drop errored steps entirely to unblock conversations

rfrenoy:
## Description

Ticket: https://github.com/dust-tt/tasks/issues/9222

An upstream provider error left a truncated function_call in the errored step's content. renderConversationForModelMultiActions only skipped the error content item, replaying the malformed call to the model which rejected every retry with multi_actions_error, permanently blocking the conversation. Drop the whole errored step (contents + actions) instead.

## Tests

Unit

## Risk

Low
## Deploy Plan

Front

### DIFF
--- a/front/lib/api/assistant/conversation_rendering/helpers.test.ts
+++ b/front/lib/api/assistant/conversation_rendering/helpers.test.ts
@@ -339,6 +339,136 @@ The following skills are available for use with the skill_management__enable_ski
   });
 });
 
+describe("error content handling in getSteps", () => {
+  async function buildErroredMessage(contents: AgentMessageType["contents"]) {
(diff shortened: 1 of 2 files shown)


flvndvd:
@rfrenoy looking that the associated conversations, they are hitting errors because the signed url we provide somehow has expired, did you find otherwise?

rfrenoy:
@flvndvd I think you are correct. I followed the Eng-runner analysis to reproduce the errors, but these are not the actual errors the conversations in the tickets are getting. I'll close this one and look for a fix to keep the conversations working on attachment urls that the model can't download. Will keep you posted if I find anything.

flvndvd:
I have another card about this, I'll probably do some work as well.

PR #28428, opened 2026-07-03
[front] chore: cleanup `constructPromptMultiActions` signature

aubin-tchoi:
## Description

This PR removes `equippedSkills` and `enabledSkills` from the signature of `constructPromptMultiActions`.
These two fields are unused now that we moved skill rendering to user messages.
Goal is to have `constructPromptMultiActions` as lean as possible to avoid creating ways to add entropy to the system prompt.

## Tests

- Tested locally.

## Risk

- Low.

## Deploy Plan

- Deploy front.


aubin-tchoi:
PMRR (got another one coming up)

PR #28430, opened 2026-07-03
Pricing: display period and prorated period below each seat line of the billing 

fabiencelier:
## Description

closes https://github.com/dust-tt/tasks/issues/9399

When we add seat mid contract, they are billed prorated at the end of the period.
This means the bill next month will be higher (prorated seat + next seat) and without the period it is confusing to understand (bill is higher than normal bill afterwards).

This adds the period in the UI to clarify this

image

## Tests

tested locally

## Risk

low

## Deploy Plan

deploy front

### DIFF
--- a/front-api/routes/w/[wId]/metronome/invoice.ts
+++ b/front-api/routes/w/[wId]/metronome/invoice.ts
@@ -201,10 +201,18 @@ app.get(
           totalCents: itemCurrency
             ? amountCents(item.total, itemCurrency)
             : item.total,
+          isProrated: item.is_prorated ?? false,
+          periodStartMs: item.starting_at
+            ? new Date(item.starting_at).getTime()
+            : null,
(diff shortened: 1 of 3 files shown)


front-api/routes/w/[wId]/metronome/invoice.ts
```diff
@@ -201,10 +201,18 @@ app.get(
...
+            : null,
+          // Exclusive end of the period covered by the line item.
+          periodEndMs: item.ending_before
+            ? new Date(item.ending_before).getTime()
+            : null,
         };
       });
 
-    return ctx.json({ currency, lineItems, items: invoice.line_items });
+    return ctx.json({ currency, lineItems });
```
Comment (fabiencelier) [inline front-api/routes/w/[wId]/metronome/invoice.ts:215]: debug leftover I think ?
It was returning the whole line, potentially leaking some other information to the frontend.

thervier (approved):
LGTM

PR #28437, opened 2026-07-03
fix(triggers): pass owner as prop to TriggerPodSelector instead of context

ElPicador:
## Description

`TriggerPodSelector` read `owner` from `useAgentBuilderContext()`, which threw `useAgentBuilderContext must be used within an AgentBuilderProvider` when the component was rendered inside `AgentDetailsSheet` → `TriggerEditView` — a tree with no `AgentBuilderProvider`.

This removes the context dependency and passes `owner` down as a prop. Both `ScheduleEditionSheetContent` and `WebhookEditionSheetContent` already have `owner`, so it's threaded through their `*PodSelector` sub-components into `TriggerPodSelector`.

## Tests

Open a schedule/webhook trigger edit sheet from the agent details sheet (outside the agent builder) and confirm the Pod selector renders without crashing.


ElPicador:

- #28437 ◀

Fraggle (commented):
LGTM

PR #28439, opened 2026-07-03
feat: allow business admins to see usage page

matteotrab:
## Description

This PR extends access to the usage page and related credits/analytics APIs to users with the `business admin` role, in addition to regular admins.

We want business admins to see and interact with the members tab. Business admins should not however be able to edit the spend limits of groups or the settings in the settings tab.

## Tests

Manually.

## Risks
Low. The `ensureIsBusinessAdmin` middleware should be verified to correctly scope access and not inadvertently grant more permissions than intended.

## Deploy Plan

Standard deployment.

### DIFF
--- a/front/components/pages/workspace/UsagePage.tsx
+++ b/front/components/pages/workspace/UsagePage.tsx
@@ -239,8 +239,6 @@ export function UsagePage() {
       }),
     []
   );
-  // Admin-only Requests tab: pending member-initiated upgrade requests, resolved
-  // by approving (via the seat / spend-limit modals) or denying.
   const isWorkspaceAdmin = isAdmin(owner);
   const modelsPickerEnabled = hasFeature("models_picker") && isWorkspaceAdmin;
   const [membersTab, setMembersTab] = useState<"members" | "requests">(
@@ -248,7 +246,6 @@ export function UsagePage() {
   );
   const { upgradeRequests, isUpgradeRequestsLoading } = useUpgradeRequests({
     workspaceId: owner.sId,
-    disabled: !isWorkspaceAdmin,
   });
 
   const filteredUpgradeRequests = useMemo(() => {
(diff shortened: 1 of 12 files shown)


matteotrab:

- #28439 ◀

rfrenoy (approved):
lgtm

front/components/pages/workspace/UsagePage.tsx
```diff
@@ -239,16 +239,13 @@ export function UsagePage() {
       }),
     []
   );
-  // Admin-only Requests tab: pending member-initiated upgrade requests, resolved
```
philipperolet, line 242: ooc why not keep that com and use !!isBusinessAdmin?

philipperolet (approved):
LGTM 🙏  left a com

front/components/pages/workspace/UsagePage.tsx
matteotrab: I had removed it because the `isWorkspaceAdmin` was not being used only for what the comment was saying and I found it confusing. Also the page was accessible to admins only so it did not make much sense

philipperolet (commented):
Actually, a few additional coms, would be great to handle in a follow-up PR
  - business admins can now resolve upgrade requests, but request emails still only go to full admins. The PR exposes the Requests UI/actions to business admins and authorizes PATCH with ensureIsBusinessAdmin() but request-created notifications still fetch recipients with roles: ["admin"] (upgrade_requests.ts:65). Business admins won’t know requests exist unless they manually visit Usage.

  - maybe worth functional tests. This PR changes authorization on several sensitive endpoints, including seat changes, spend limits, bulk spend limits, purchase/config reads, and upgrade-request resolution, but no test files were updated. Existing tests mostly assert user is rejected and admin works; they don’t prove business_admin can do the intended actions or is still blocked from admin-only writes like purchase POST, usage-configuration PATCH, and group spend-limit PUT.

PR #28440, opened 2026-07-03
[front,sparkle] fix: prevent panel content reflow during open/close animations

aubin-tchoi:
## Description

This PR fixes resizable side-panel animations that caused content to reflow while opening or closing.

Animated resizable groups transition sibling panels together while opted-in panels preserve their content layout. Both use 300ms out-quint motion and respect reduced-motion preferences.

This affects in a particularly visible way the conversation side panel (the one for credits for instance).

## Tests

- Tested locally.

## Risk

- Low.

## Deploy Plan

- Deploy front.


ClementAupiais (approved):
The modal drawer (agent details) feels a bit slow, let's just make sure it's using the same speed and velocity (might just be me)

PR #28441, opened 2026-07-03
Strengthen the tool search instruction against mixing searches with skill activations

flvndvd:
## Description

Follow-up to #28403. The single sentence added there was not enough: the model still frequently issues a tool search alongside a skill activation in the same turn. That combination is the worst case for us, because enabling a skill injects a composite user message with the skill instructions into the next request, which guarantees the un-run search cannot be resumed and the whole request gets rejected.

If this still does not eliminate the failure, the robust fix remains stripping unresumable dangling search blocks from the request before sending, as mentioned in the previous PR.


aubin-tchoi (approved):
LGTM

PR #28443, opened 2026-07-03
Add: audit log when manipulating workspace advanced models access + default is on.

Fraggle:
## Description

Changes workspace advanced model access to **default ON**: rather than an opt-in allowlist (empty table = no access), `workspace_allowed_advanced_models` now acts as a deny-list using an `enabled` boolean column. `loadWorkspaceAllowedAdvancedModels` returns all advanced models when no rows exist via `getDefaultWorkspaceAllowedAdvancedModels()`. `removeWorkspaceAllowedAdvancedModel` now upserts a row with `enabled=false` instead of deleting; `addWorkspaceAllowedAdvancedModel` destroys the disabled row to re-enable.

Adds WorkOS audit log events for `workspace.advanced_model_access_updated`, `user.advanced_model_access_updated`, and `group.advanced_model_access_updated` emitted by the new `emitAdvancedModelAccessUpdatedAuditLog` private helper in `AdvancedModelResource`, covering add/remove at all three scopes with `provider_id`, `model_id`, and `enabled` metadata.

Pre-deploy SQL migration adds `enabled BOOLEAN NOT NULL DEFAULT TRUE` to `workspace_allowed_advanced_models` and removes the `dust_internal/advanced_models_allowlist_configured` sentinel row.

## Tests

- Updated Vitest tests in `advanced_model_resource.test.ts` and `access.test.ts` to reflect the new default-all-enabled semantics.
- Added cases for: workspace default returning all models, disabling a specific model at workspace level, disabling all workspace models, and re-enabling after removal.
- Integration test in `index.test.ts` now validates that the initial list returns all models before any explicit workspace add.

## Risk

Medium. Workspaces with no rows in `workspace_allowed_advanced_models` will now have all advanced models enabled by default (previously none). The migration is additive (new column with `DEFAULT TRUE`) plus deletes only the sentinel row — safe to apply without downtime. Feature is gated on the `models_picker` FF. Rollback requires reverting the column removal and re-inserting the sentinel row if needed.

## Deploy Plan

1. Run pre-deploy SQL on the front DB: `front/migrations/pre-deploy/20260703134500_add_enabled_to_workspace_allowed_advanced_models.sql`
2. Deploy front.


Fraggle:

- #28443 ◀

front/lib/resources/advanced_model_resource.ts
```diff
@@ -268,14 +306,27 @@ export class AdvancedModelResource {
...
         providerId: modelRes.value.providerId,
         modelId: modelRes.value.modelId,
       },
     });
 
+    if (!existingRow || existingRow.enabled) {
+      return new Ok(undefined);
+    }
+
+    await existingRow.destroy();
```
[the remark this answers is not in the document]
Fraggle: False positive, it's intentional (and covered by test)

front/admin/audit_log_schemas/group.advanced_model_access_updated.json
```diff
@@ -0,0 +1,16 @@
...
+  "targets": [
+    {
+      "type": "workspace"
+    },
+    {
+      "type": "group"
+    }
+  ],
+  "metadata": {
+    "enabled": "string",
```
Nils-Fedrigo, line 12: enabledBy?

Nils-Fedrigo: or do we have that already?

Nils-Fedrigo (approved):
thank you!

front/admin/audit_log_schemas/group.advanced_model_access_updated.json
Fraggle: I'm not sure, looking into it.

Fraggle: Yep, all audit log have [an actor that is built from the authenticator](https://github.com/dust-tt/dust/blob/4487e543107c2a3b3cd5e8e314244479acde6bf6/front/lib/api/audit/workos_audit.ts#L238).

PR #28446, opened 2026-07-03
Add SandboxFunctionMCPAction model & migration

PopDaph:
## Description

Sandbox functions cannot call MCP tools today: the `/v1/w/:wId/sandbox/actions/*` endpoints require a conversation-bound token and persist an `AgentMCPAction`, which mandates a conversation, an agent message and a step content. A function invocation has none of those.

This PR lays the foundations, with **no behavior change**:

1. `SandboxFunctionMCPActionModel` + pre-deploy migration: an MCP tool call made from a sandbox function invocation.
2. `ToolExecutionStatus` split into a base lifecycle + agent-loop extension.
3. RESTRICT-FK deletion wiring: view hard-delete flows clean up action rows (`destroyMCPServerViewDependencies`); the invocation-deletion side is a dated TODO, landing with the resource in PR2 — safe because nothing writes to the table until PR3.

The `agent_loop_context_required` manifest flag (whether a server's tools need an agent loop to run) ships as its own mechanical PR, consumed by PR3.

## The model, vs `AgentMCPActionModel`

| `AgentMCPActionModel` | Ours | Why |
|---|---|---|
| `agentMessageId` FK (NOT NULL) | → `sandboxFunctionInvocationId` FK | the parent swap, that's the whole point of the new model |
| `stepContentId` FK (+ join row) | dropped | ties the action to the LLM's function-call step content; no LLM step exists here |
| `mcpServerConfigurationId` (string) | → `mcpServerViewId` FK | theirs is the sId of a virtual agent-config object (can't FK); ours is a real view row |
| `version` (default 0) | dropped | versioning of re-executed actions within an agent message step; no such concept here |
| `status` (`ToolExecutionStatus`, nullable) | `ToolExecutionBaseStatus`, NOT NULL | theirs carries `ready_*` scheduling + conversation `blocked_*` states; ours is the base lifecycle (nullable was a backfill artifact) |
| `citationsAllocated` (default 0) | dropped | citation ref accounting per agent message; no conversation to cite into |
| `augmentedInputs` JSONB | → `inputs` JSONB | same role; nothing gets "augmented" without an agent config injecting configured values |
| `toolConfiguration` JSONB | same (minimal, synthesized) | execution snapshot (timeout, retry, wire name); derived + gated at creation, executed from the row, same pattern as `create_child_action.ts` |
| `stepContext` JSONB | dropped | citations offsets / resume state / sandbox-child info, all agent-loop step machinery |
| `AgentMCPActionOutputItemModel` (per-block rows) | → single `outputGcsPath` column | per-block rows exist for conversation rendering; dsbx reads the output exactly once as one array, written once to one GCS object |

## Statuses

`ToolExecutionBaseStatus` (`running | blocked_validation_required | succeeded | errored | denied`) is the lifecycle any tool execution has; `ToolExecutionStatus` extends it with the agent-loop-only states. Same values as before, now with the relationship explicit and `satisfies`-pinned.

Note: `blocked_validation_required` / `denied` have no writer yet: v1 rejects approval-requiring tools at call time. They are in the base because approvals will be bubbled up to the user later (gated on invocation durability).

## Follow-ups

- `agent_loop_context_required` manifest flag (mechanical, already branched)
- PR2: `SandboxFunctionMCPActionResource` (GCS output helpers, deletion wiring + regression tests)
- PR3: endpoints + Temporal execution, first e2e with `common_utilities`
- PR4: pod file outputs, flipping real servers

## Risk

Worst case, a dead table. Safe to rollback. Migration is addition-only.

## Deploy Plan

- [ ] Apply pre-deploy migration
- [ ] Deploy front

### DIFF
--- a/front/admin/db.ts
+++ b/front/admin/db.ts
@@ -132,6 +132,7 @@ import {
   SandboxFunctionInvocationModel,
   SandboxFunctionModel,
 } from "@app/lib/resources/storage/models/sandbox_function";
+import { SandboxFunctionMCPActionModel } from "@app/lib/resources/storage/models/sandbox_function_mcp_action";
 import { SpaceModel } from "@app/lib/resources/storage/models/spaces";
 import {
   TakeawaySourcesModel,
@@ -223,6 +224,7 @@ export function loadAllModels() {
     RemoteMCPServerToolMetadataModel,
     InternalMCPServerCredentialModel,
     ConversationMCPServerViewModel,
+    SandboxFunctionMCPActionModel,
     AgentMCPServerConfigurationModel,
     AgentTablesQueryConfigurationTableModel,
     AgentDataSourceConfigurationModel,
--- a/front/lib/actions/statuses.ts
+++ b/front/lib/actions/statuses.ts
@@ -6,13 +6,41 @@ const TOOL_EXECUTION_FINAL_STATUSES = [
 
 type ToolExecutionFinalStatus = (typeof TOOL_EXECUTION_FINAL_STATUSES)[number];
 
+// Base lifecycle shared by every MCP tool execution, regardless of context: the tool runs, may be
+// blocked awaiting user approval, and ends succeeded, errored, or denied. AgentMCPActionModel and
+// SandboxFunctionMCPActionModel both draw from this vocabulary; the agent loop extends it with its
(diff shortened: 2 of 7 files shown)


spolu (approved):
💯 beautiful

PR #28448, opened 2026-07-03
Check free credits

tdraier:
## Description

Follow-up for #28450.

Original reason: some legacy `PRO_`/`ENT_` workspaces had missing free programmatic-usage credits because their Metronome contract had been archived — with no live contract, the `credit.segment.start` webhook never fires, so nothing was creating the recurring `free` credit anymore. #28450 fixes this going forward by granting the credit directly from the Stripe `customer.subscription.updated`/`created` webhook instead of depending on Metronome. This PR is the one-off reconciliation script needed to catch up workspaces that already missed grants (or whose existing credit's cadence had drifted from Stripe's actual billing period) before that fix landed.

This adds `scripts/realign_free_credits_to_stripe_periods.ts`, an operator-run script that reconciles the DB `free` credit ledger against Stripe, treating Stripe's period as the source of truth (Metronome can no longer be trusted/queried for these workspaces).

### What it does

For every active, non-free-test-plan subscription whose plan code starts with `PRO_` or `ENT_` (optionally further filtered to one exact `--planCode`) and that has a live Stripe subscription:

1. Computes the current Stripe period, floored to the hour (`epochSecondsToFloorHourISO`) to match the convention existing credits were created under.
2. Looks only at `free` credits whose `invoiceOrLineItemId` starts with `free-renewal-` — the recurring-renewal family — explicitly excluding one-off manual `free-poke-*` grants from the poke plugin, which must never be read as, or realigned into, the renewal cadence.
3. If no such credit currently covers today: **backfills** one spanning the current Stripe period end-to-end.
4. If one does cover today but its `expirationDate` doesn't match Stripe's period end (drift, e.g. from a Metronome contract that was re-provisioned at a different time than the Stripe subscription): **snaps its end date** to the Stripe boundary and **provisions the next period's credit** starting exactly there, so future renewals stay aligned.
5. If the active credit already matches Stripe's period end: no-op.

The credit amount uses the workspace's `ProgrammaticUsageConfiguration.freeCreditMicroUsd` override when configured (negotiated deals), otherwise the standard bracket formula (`calculateFreeCreditAmountMicroUsd`), scaled ×12 (`YEARLY_MULTIPLIER`) for yearly subscriptions — mirroring the amount logic in `handleFreeCreditSegmentGrant`. Credits created by this script use the idempotency key format `free-renewal-{workspaceId}-{startEpochSeconds}`, so re-running the script is a no-op once a workspace is aligned.

Dry-run by default (logs `[DRY RUN] ...` for every intended action); pass `--execute` to actually create/update credits.

### Changes

- **`lib/resources/credit_resource.ts`**: add `updateExpirationDate(auth, expirationDate)` — no prior method could mutate a started credit's `expirationDate`.
- **`scripts/realign_free_credits_to_stripe_periods.ts`** (new): the realignment/backfill script described above.

## Tests

- `npx tsgo --noEmit` clean.
- `npm run format:changed` clean.
- No automated tests added — this is an operator-run one-off script, run in dry-run mode against production data to inspect output before any `--execute` run.

## Risk

Medium (billing — creates/edits real credit records). Mitigations:
- Dry-run by default; requires explicit `--execute` to write.
- Idempotent creates (deterministic `free-renewal-{workspaceId}-{startEpochSeconds}` key) — safe to re-run.
- Scoped to `PRO_`/`ENT_` plan codes only, even when `--planCode` is passed, so it can never touch new-pricing (`CP_*`) workspaces that never had this free-credit mechanism.
- Only reads/writes credits already in the `free-renewal-*` family — never touches manually-granted `free-poke-*` credits.

## Deploy Plan

Should land and run after #28450 is merged/deployed, so newly-realigned/backfilled credits keep renewing correctly going forward via the Stripe webhook instead of drifting again. No deploy required to run the script itself (it's an ad hoc `npx tsx` invocation, not wired into any request path or webhook). Recommended rollout: dry-run first with a narrow `--planCode`, review the logged actions, then re-run with `--execute` once satisfied; broaden to other plan codes as needed.

### DIFF
--- a/front/lib/resources/credit_resource.ts
+++ b/front/lib/resources/credit_resource.ts
@@ -452,6 +452,19 @@ export class CreditResource extends BaseResource<CreditModel> {
     );
   }
 
+  async updateExpirationDate(
+    auth: Authenticator,
+    expirationDate: Date,
+    { transaction }: { transaction?: Transaction } = {}
+  ): Promise<[affectedCount: number]> {
+    return this.update(
+      {
+        expirationDate,
+      },
+      transaction
+    );
+  }
+
   async setMetronomeCreditId(
     metronomeCreditId: string,
     { transaction }: { transaction?: Transaction } = {}
--- a/front/scripts/realign_free_credits_to_stripe_periods.ts
+++ b/front/scripts/realign_free_credits_to_stripe_periods.ts
@@ -1,0 +1,345 @@
+import { Authenticator } from "@app/lib/auth";
+import {
+  calculateFreeCreditAmountMicroUsd,
+  countEligibleUsersForFreeCredits,
+  YEARLY_MULTIPLIER,
+} from "@app/lib/credits/free";
+import { getStripeSubscription } from "@app/lib/plans/stripe";
+import { CreditResource } from "@app/lib/resources/credit_resource";
+import { ProgrammaticUsageConfigurationResource } from "@app/lib/resources/programmatic_usage_configuration_resource";
+import { SubscriptionResource } from "@app/lib/resources/subscription_resource";
+import { WorkspaceResource } from "@app/lib/resources/workspace_resource";
+import { concurrentExecutor } from "@app/lib/utils/async_utils";
+import type { ArgumentSpecs } from "@app/scripts/helpers";
+import { makeScript } from "@app/scripts/helpers";
+import type Stripe from "stripe";
+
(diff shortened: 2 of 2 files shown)


tdraier:

- #28450
    - #28448 ◀

rfrenoy (approved):
lgtm

PR #28451, opened 2026-07-03
Ignore h1-pentest test connectors in last-sync production check

davidebbo:
## Summary

The "Connectors have not synced in the last week" production check was alerting on connectors `55901` and `55902`, which belong to the **h1-pentest** workspace — a test we don't want to alert on.

This adds an `IGNORED_CONNECTOR_IDS` list and filters those connectors out before the freshness check.

## Test plan

- Typecheck / biome pass via pre-commit hooks.
- The two connectors will no longer be considered by `checkConnectorsLastSyncSuccess`, so they can't trigger the alert.


ykmsd (approved):
alright it feels a bit weird to me but we've got an approval from @zmarouf so this is safe 🫡

PR #28457, opened 2026-07-03
Pricing: filter invoices on the actual workspace

fabiencelier:
## Description

closes https://github.com/dust-tt/tasks/issues/9403
Fetch more than the 12 invoices and filter on the one marked as belonging to the workspace.

## Tests

TODO: test locally

## Risk

risk: if we have more than 88% of invoices belonging to other workspaces we may see less than 12 invoices.
No really probably but possible. In that case we can change to do multiple queries until we have 12 for the given workspace.
Easy to revert if needed

## Deploy Plan

deploy front


thervier (approved):
LGTM

PR #28461, opened 2026-07-03
Tool outputs: resolve DustFileSystem from tool context

spolu:
## Description

`writeToToolOutputsFolder` / `persistToolOutput` now take a `ToolContextType` instead of a conversation. The file system is resolved inside `action_output_fs`: conversation file system for agent loop run contexts, pod file system for sandbox function run contexts (`invocation.sandboxFunction.space`). Outputs land in `conversation-{cId}/.tool_outputs/...` or `pod-{pId}/.tool_outputs/...` accordingly.

Call sites (`processToolResults`, `web_search_browse`, `run_dust_app`, `extract_data`) thread their existing `toolContext` through. No behavior change for the agent loop path.

This is plumbing for now we still need to get rid of `conversation` in `front/lib/actions/mcp_execution.ts`

## Tests

Updated `mcp_execution.test.ts` setup to pass a tool context; existing offload tests pass.

## Risk

Low. Pod path is not exercised yet (no caller runs tools in a sandbox function context); agent loop path is unchanged.

## Deploy Plan

- deploy `front`

### DIFF
--- a/front/lib/actions/mcp_execution.test.ts
+++ b/front/lib/actions/mcp_execution.test.ts
@@ -83,17 +83,38 @@ async function setupTest() {
     retryPolicy: "no_retry",
   };
 
-  const { action } = await ConversationFactory.createAgentMessage(auth, {
-    workspace,
-    conversation,
-    agentConfig,
-    mcpAction: { toolConfiguration },
-  });
+  const { action, agentMessage } = await ConversationFactory.createAgentMessage(
+    auth,
+    {
+      workspace,
+      conversation,
+      agentConfig,
+      mcpAction: { toolConfiguration },
+    }
+  );
   assert(action, "MCP action should be created");
 
-  const toolContext = {
-    runContext: { contextType: "agent_loop", conversation },
-  } as unknown as ToolContextType;
+  // The agent message above is created with rank 0, use rank 1 for the user message.
+  const { userMessage } = await ConversationFactory.createUserMessage({
+    auth,
+    workspace,
+    conversation,
+    content: "Test message",
+    rank: 1,
+  });
+
+  const toolContext: ToolContextType = {
+    runContext: {
+      contextType: "agent_loop",
+      agentConfiguration: agentConfig,
+      agentMessage,
+      currentAction: action.toJSON(),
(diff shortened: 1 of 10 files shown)


front/lib/actions/mcp_execution.test.ts
```diff
@@ -90,7 +91,11 @@ async function setupTest() {
   });
   assert(action, "MCP action should be created");
 
-  return { auth, conversation, action, toolConfiguration };
+  const toolContext = {
+    runContext: { contextType: "agent_loop", conversation },
+  } as unknown as ToolContextType;
```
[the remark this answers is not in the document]
spolu: Fixed in 2b401445df: the cast is gone, the test now builds a fully-typed `ToolContextType` from the factories (`createAgentMessage` for the agent message/action, `createUserMessage` for the user message).

spolu:
Tested locally e2e

PMRR

front/lib/actions/mcp_execution.ts
```diff
@@ -145,18 +148,24 @@ export async function processToolResults(
...
     toolCallResultContent,
-    toolConfiguration,
+    toolContext,
   }: {
     action: AgentMCPActionResource;
     conversation: ConversationType;
     localLogger: Logger;
     toolCallResultContent: CallToolResult["content"];
-    toolConfiguration: LightMCPToolConfigurationType;
+    toolContext: ToolContextType;
```
flvndvd, line 157: Looking at the call site it's seems doable to enforce the runcontext variant here? Can we try to leverage TS instead of runtime assert?

flvndvd (commented):
LGTM

front/lib/actions/mcp_execution.ts
spolu: Yeah I will start pushing ToolContextType[runContext] (and give it a name) wherever possible 👍 Which should be most places.

spolu: Ideally we even remove the listContext entirely at some point and have two separate types for 2 separate code paths but not sure we'll get all the way there

PR #28463, opened 2026-07-03
fix(conversations): don't 500 on userless auth when listing private-by-default conversations

davidebbo:
## Problem

Workspaces with **"Private conversation URLs by default"** enabled were emitting `User not found while auth method is not api key, system api key, or internal` errors (500s).

`ConversationResource.baseFetchWithAuthorization` — the central path behind nearly all conversation list/fetch calls — throws whenever a **non-bypassed** authenticator (`session`, `oauth`, or `sandbox_token`) has a `null` user while the workspace flag is on.

A null user reaches that point in three legitimate ways:

1. **Userless sandbox token** (`Authenticator.fromSandboxToken`) — a conversation driven by a non-human actor (e.g. a Slack bot / triggered conversation). Most likely culprit for bot-active workspaces.
2. **`session`** — a valid WorkOS session with no matching Dust `User` row (deleted user, mid-onboarding, cross-region/stale session).
3. **Bearer `oauth` via `fromSession`** — same missing-Dust-user case (`fromWorkOSToken` guards this, but the `fromSession` bearer branch does not).

## Fix

The sibling single-access path, `canUserAccessPrivateByDefaultConversation`, already handles a null user **gracefully by denying access** (returns `false`, no throw). The list path was simply inconsistent — it threw a 500 instead of filtering.

This change aligns the two: a userless authenticator can never be a conversation participant, so it sees **none** of the participant-restricted conversations instead of crashing.

**No capability change** — userless auth was already denied these conversations via `canAccess`; this only stops the 500 on the list path.

## Test plan

- [x] `npx tsgo --noEmit` passes
- [x] `biome check` clean
- [ ] Reviewer: confirm behavior via a workspace with the flag on + a bot/triggered conversation

### DIFF
--- a/front/lib/resources/conversation_resource.test.ts
+++ b/front/lib/resources/conversation_resource.test.ts
@@ -1811,6 +1811,75 @@ describe("baseFetchWithAuthorization with space-based permissions", () => {
     );
   });
 
+  it("should not throw and should filter participant-restricted conversations for a userless sandbox token when private conversation URLs are private by default", async () => {
+    const updateResult = await WorkspaceResource.updateMetadata(workspace.id, {
+      privateConversationUrlsByDefault: true,
+    });
+    assert(updateResult.isOk(), "Failed to enable private conversation URLs");
+
+    const refreshedAdminAuth = await Authenticator.fromUserIdAndWorkspaceId(
+      adminAuth.getNonNullableUser().sId,
+      workspace.sId
+    );
+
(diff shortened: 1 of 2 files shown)


Fraggle (approved):
LGTM, would add / update tests :)

davidebbo:
> LGTM, would add / update tests :)

Good call. Test added.

PR #28464, opened 2026-07-03
Block selection of non available models and prevent save.

Fraggle:
## Description

Adds model selectability enforcement for the `models_picker` feature flag. When the FF is on, advanced models are shown in the agent builder picker but disabled (`isSelectable: false`) if the user/workspace hasn't been granted access via `AdvancedModelResource.resolveAllowedAdvancedModels`. A new `EnabledModelConfigurationType` (= `ModelConfigurationType & { isSelectable: boolean }`) threads through the UI. The currently-selected model — if now restricted — stays visible in `ModelSelectionSubmenu` with a "Not enabled for you. Choose another model to save." warning. `createOrUpgradeAgentConfiguration` enforces access server-side via `getAdvancedModelAccessErrorForAgentConfiguration` and returns an `Err` on save if the model isn't allowed. `getDefaultModel` filters to selectable models only.

file-31dc8a74f7a9bc1d32f63ee0f938ec9d

image

## Tests

- New Vitest suite for `withModelSelectability` in `front/lib/advanced_models/enabled_models.test.ts` covering: FF off (all selectable), non-advanced (always selectable), advanced + disallowed (not selectable), advanced + allowed (selectable).
- Updated `isModelAvailable` and `filterEnabledModels` tests in `front/lib/assistant.test.ts` for `models_picker` + no plan `hasAdvancedModelAccess` cases.
- Tested manually in the agent builder UI.

## Risk

All changes are gated behind the `models_picker` FF. When the FF is off, `withModelSelectability` marks every model selectable and behaviour is unchanged. Low risk, trivially rollback-safe.

## Deploy Plan

Deploy front.

### DIFF
--- a/front-api/routes/w/[wId]/models.ts
+++ b/front-api/routes/w/[wId]/models.ts
@@ -1,10 +1,6 @@
-import { USED_MODEL_CONFIGS } from "@app/components/providers/model_configs";
-import { getWhitelistedProviders } from "@app/lib/api/assistant/models";
-import { config as regionConfig } from "@app/lib/api/regions/config";
-import { filterEnabledModels } from "@app/lib/assistant";
-import { getFeatureFlags } from "@app/lib/auth";
+import { withModelSelectability } from "@app/lib/advanced_models/enabled_models";
+import { getAvailableModelsForWorkspace } from "@app/lib/api/assistant/workspace_capabilities";
 import type { GetEnabledModelsResponseType } from "@app/types/api/assistant/models";
-import { CUSTOM_MODEL_CONFIGS } from "@app/types/assistant/models/custom_models.generated";
 import { workspaceApp } from "@front-api/middlewares/ctx";
 import type { HandlerResult } from "@front-api/middlewares/utils";
 
@@ -15,20 +11,10 @@ const app = workspaceApp();
 app.get("/", async (ctx): HandlerResult<GetEnabledModelsResponseType> => {
   const auth = ctx.get("auth");
 
-  const featureFlags = await getFeatureFlags(auth);
-  const owner = auth.getNonNullableWorkspace();
-  const plan = auth.plan();
-  const region = regionConfig.getCurrentRegion();
-  const whitelistedProviders = getWhitelistedProviders(auth);
+  const availableModels = await getAvailableModelsForWorkspace(auth);
 
-  // Include both standard models and custom models (from GCS at build time).
-  const allUsedModels = [...USED_MODEL_CONFIGS, ...CUSTOM_MODEL_CONFIGS];
-  const models = filterEnabledModels(allUsedModels, {
-    featureFlags,
-    plan,
-    regionalModelsOnly: owner.regionalModelsOnly,
-    region,
-    whitelistedProviders,
+  const models = await withModelSelectability(auth, {
+    models: availableModels,
   });
 
   return ctx.json({ models });
--- a/front/components/agent_builder/instructions/ModelSelectionSubmenu.tsx
+++ b/front/components/agent_builder/instructions/ModelSelectionSubmenu.tsx
@@ -8,8 +8,8 @@ import { RegionalFlag } from "@app/components/shared/RegionalFlag";
 import { useTheme } from "@app/components/sparkle/ThemeContext";
 import { useAuth, useFeatureFlags } from "@app/lib/auth/AuthContext";
 import { useRegionContext } from "@app/lib/auth/RegionContext";
+import type { EnabledModelConfigurationType } from "@app/types/api/assistant/models";
 import { getProviderDisplayName } from "@app/types/assistant/models/providers";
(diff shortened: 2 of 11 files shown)


Fraggle:

- #28464 ◀

front/lib/assistant.ts
```diff
@@ -32,7 +32,10 @@ export function isModelAvailable(
     region: RegionType;
   }
 ) {
-  if (m.largeModel && !isUpgraded(plan)) {
+  const includeAdvancedModelInPicker =
```
Fraggle, line 36: otherwise the models gets filtered too early

front-api/routes/w/[wId]/models.ts
```diff
@@ -23,14 +24,16 @@ app.get("/", async (ctx): HandlerResult<GetEnabledModelsResponseType> => {
 
   // Include both standard models and custom models (from GCS at build time).
   const allUsedModels = [...USED_MODEL_CONFIGS, ...CUSTOM_MODEL_CONFIGS];
-  const models = filterEnabledModels(allUsedModels, {
+  const enabledModels = filterEnabledModels(allUsedModels, {
```
Comment (Nils-Fedrigo) [inline front-api/routes/w/[wId]/models.ts]: so logic is "all models from enabled providers are displayed, and then we check if you have access to it?"
Maybe it's better to either 
- rename this function getModelsFromEnabledProviders or sth
- fold the withModelSelectability in it
to avoid confusion

Comment (Nils-Fedrigo) [reply front-api/routes/w/[wId]/models.ts]: also this currently does not align with disabled providers (eg when picking an agent done on Sonnet 4.6, opening the agent for modification after disabling A\ yields this)

Image

Would be cool if we could have the same behaviour either way while we're at it! (with different tooltips)

Comment (Fraggle) [reply front-api/routes/w/[wId]/models.ts]: Good idea, will TAL.

front/components/agent_builder/instructions/ModelSelectionSubmenu.tsx
```diff
@@ -24,29 +25,45 @@ import type React from "react";
 import { useController } from "react-hook-form";
 
 interface ModelSelectionSubmenuProps {
-  models: ModelConfigurationType[];
+  models: EnabledModelConfigurationType[];
 }
 
+const NOT_SELECTABLE_MODEL_DESCRIPTION =
+  "Not enabled for you. Choose another model to save.";
```
Nils-Fedrigo, line 31: We have limited real estate but I think it's not clear enough, lmk
```suggestion
  "No access. To save, try another model or ask an admin.";
```

Nils-Fedrigo: let's not split hair in half :o)

front/components/agent_builder/instructions/ModelSelectionSubmenu.tsx
```diff
@@ -24,29 +25,45 @@ import type React from "react";
...
 }
 
 function ModelRadioItem({
   modelConfig,
   isDark,
+  isSelected,
   onModelSelection,
   regionalComponent,
 }: ModelRadioItemProps) {
+  const isSelectable = isEnabledModelSelectable(modelConfig);
```
Nils-Fedrigo: I feel like this is creating confusion for nothing while it is jst 
`model.isSelectable`

front/components/agent_builder/instructions/ModelSelectionSubmenu.tsx
```diff
@@ -24,29 +25,45 @@ import type React from "react";
...
+      description={
+        !isSelectable && isSelected
+          ? NOT_SELECTABLE_MODEL_DESCRIPTION
+          : modelConfig.shortDescription
+      }
       label={modelConfig.displayName}
-      onClick={() => onModelSelection(modelConfig)}
+      disabled={!isSelectable}
+      onClick={() => {
+        if (isSelectable) {
```
Nils-Fedrigo: mayyybe not needed ?

front/lib/assistant.ts
Nils-Fedrigo: I'd argue for a comment in code then

Nils-Fedrigo (approved):

front/lib/assistant.ts
Fraggle: yep

front-api/routes/w/[wId]/models.ts
Comment (Fraggle) [reply front-api/routes/w/[wId]/models.ts]: As seen IRL, will adjust the name but not do the provider thing right now because it might affect other callers of the function.

front/components/agent_builder/instructions/ModelSelectionSubmenu.tsx
Fraggle: checking.

Fraggle: removed

PR #25903, opened 2026-05-20
Migrate various handlers: stripe, debug, email and more

davidebbo:
## Description

Did some refactoring, and migrated the following to Hono:

front/pages/api/email/validate-action.ts
front/pages/api/email/webhook.ts
front/pages/api/enrichment/company.ts
front/pages/api/metronome/webhook.ts
front/pages/api/stripe/portal.ts
front/pages/api/stripe/webhook.ts
front/pages/api/debug/profiler.ts
front/pages/api/templates/index.ts
front/pages/api/templates/[tId]/index.ts

## Risk

Some. Could not easily test all the scenarios involved, and not all have test coverage.

## Deploy Plan

Front

PR #28466, opened 2026-07-03
fix(email): parse form-encoded body in email validate-action endpoint

philipperolet:
## Problem

Clicking an Allow/Decline link in an agent tool-validation email fails in the browser with:

```
{"error":{"type":"invalid_request_error","message":"Missing or invalid token parameter."}}
```

Reported in dust-tt/tasks#9410.

## Root cause

The validation page (`ValidationPage.tsx`) auto-submits a plain HTML `<form method="POST">`, which sends `application/x-www-form-urlencoded`. The old Next.js handler worked because Next's body parser transparently decodes form bodies. The Hono migration (#25903) replaced it with `ctx.req.json().catch(() => ({}))`, which fails on form-encoded bodies, silently falls back to `{}`, and 400s. Every email validation link click has been broken since.

## Fix

Parse the body according to content-type (`parseBody()` for form-encoded, JSON otherwise), matching the existing pattern in `front-api/routes/workos/[action].ts` which fixed the same migration bug for the Chrome extension token endpoint.

Adds functional tests covering form-encoded, JSON, and missing-token cases.

Fixes dust-tt/tasks#9410

### DIFF
--- a/front-api/routes/email/validate-action.test.ts
+++ b/front-api/routes/email/validate-action.test.ts
@@ -1,0 +1,49 @@
+import { generateValidationToken } from "@app/lib/api/email/validation_token";
+import { honoApp } from "@front-api/app";
+import { describe, expect, it } from "vitest";
+
+process.env.EMAIL_VALIDATION_SECRET ||= "test-email-validation-secret";
+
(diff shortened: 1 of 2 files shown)


davidebbo (commented):
LGTM

PR #28470, opened 2026-07-05
[front] Legacy Pro → Business migration: plan targeting, skip-reason triage, seat-sync perf

tdraier:
## Description

Follow-up refinements to the legacy Pro → Business migration (script + shared seat-sync), driven by observations from prod dry runs. Builds on the migration merged in #28134.

### Plan selection
- **More legacy plans forced to legacy-large.** Beyond `PRO_PLAN_SEAT_39`, the plans `PRO_PLAN_LARGE_FILES`, `PRO_PLAN_LARGE_FILES_10SPACES`, and `PRO_PLAN_PLUS_SEAT_29` now always migrate to `CP_BUSINESS_LEGACY_LARGE_PLAN` (their limits/features exceed the standard Business plan), regardless of current usage. Added the two missing legacy plan-code constants.
- **Plan-selection logging.** Each workspace now logs the chosen plan and *why* — `workspace fits standard Business limits`, `standard Business limits exceeded (<violations>)` (so it's clear why legacy-large was picked), or `PRO_PLAN_SEAT_39 always uses legacy-large Business`.
- Extended the doc-size limit on the large plan.

### Skip-reason accuracy (dry-run triage)
`getPerSeatPricing` returns null for several reasons; the script used to lump them all under `could not resolve per-seat Stripe pricing`. Now classified by the Stripe subscription's **actual seat quantity**:
- `quantity === 0` → benign skip **`workspace has no active seats`** (empty workspace; anchored on Stripe quantity, not our member count, since the two can diverge).
- anything else (metered / enterprise MAU price, missing currency option, unreadable sub) → returned as an **error** so it surfaces as a `Migration failed` line to investigate, not a silent skip.

### Performance / rate limiting (bulk run)
- **Skip seat-state reads on freshly provisioned contracts.** `switchContract`'s seat sync read current Metronome seat state per seat product to compute a diff — but a newly created contract has no prior assignments, so those reads (5/workspace for the Business package, 4 of them empty no-ops) are pure overhead. A gated `assumeEmptySeats` flag is threaded from `switchContract` (set only when the provisioned contract was **not recovered**) through `syncSeatCount` → `reconcileSeatBasedSegment`, which then treats every segment as empty and skips `getMetronomeSubscriptionSeatState`. Recovered contracts (rare) still read, so correctness is preserved. To support the gate, `createMetronomeContract` / `provisionMetronomeContract` now surface the `recovered` flag they already computed.
- **Raised Metronome SDK `maxRetries` 2 → 5** so bulk jobs survive sustained 429 rate-limit pressure (the SDK already backs off and honors `Retry-After`).

## Tests

- Existing `switch_contract` / `contracts` test mocks updated for the new `recovered` return field.
- `assumeEmptySeats` defaults to `false` (read-everything), so every other `syncSeatCount` caller and their tests are unchanged.
- `tsgo` clean on both `front` and `front-api`. (Two unrelated pre-existing hive-env `dist` skews — `@dust-tt/sparkle` `YoutrustLogo`, `@dust-tt/client` `activation_skill` — were resolved locally by rebuilding those workspaces; both are green in CI.)

## Risk

Low–medium. The seat-read skip touches shared `switchContract` seat-sync code (used by all contract switches, not just this migration), but it's gated on `recovered === false` and defaults off, so behavior is unchanged for every existing caller. `maxRetries` is a global client bump (more resilience, slower failure on persistent errors). The migration-script changes only affect dry-run classification/logging and plan targeting.

## Deploy Plan

- Ship with the rest of the migration tooling; no schema or config change required.
- For the bulk run: use `--concurrency ~4–8` (fewer concurrent Metronome calls) — combined with the per-workspace read reduction and the higher `maxRetries`, this keeps rate-limit pressure manageable.

### DIFF
--- a/front-api/routes/poke/workspaces/[wId]/switch_contract.test.ts
+++ b/front-api/routes/poke/workspaces/[wId]/switch_contract.test.ts
@@ -264,7 +264,7 @@ beforeEach(() => {
     ])
   );
   vi.mocked(provisionMetronomeContract).mockResolvedValue(
-    new Ok({ metronomeContractId: NEW_CONTRACT_ID })
+    new Ok({ metronomeContractId: NEW_CONTRACT_ID, recovered: false })
   );
   vi.mocked(remapMembershipSeatTypesForContract).mockResolvedValue(
     new Ok(undefined)
@@ -501,7 +501,7 @@ describe("POST /api/poke/workspaces/[wId]/switch_contract — Pro / Business", (
     // Second schedule with a different (business) target → ends P1, creates P2.
     const SECOND_CONTRACT_ID = "contract_new_zzz";
     vi.mocked(provisionMetronomeContract).mockResolvedValueOnce(
-      new Ok({ metronomeContractId: SECOND_CONTRACT_ID })
+      new Ok({ metronomeContractId: SECOND_CONTRACT_ID, recovered: false })
     );
     const secondResponse = await postSwitchContract(
       workspace.sId,
--- a/front/lib/api/billing/migrate_to_business.ts
+++ b/front/lib/api/billing/migrate_to_business.ts
@@ -3,7 +3,6 @@ import {
   SwitchContractBodySchema,
   switchContract,
 } from "@app/lib/api/poke/switch_contract";
-import { getMembersCount } from "@app/lib/api/workspace";
 import type { Authenticator } from "@app/lib/auth";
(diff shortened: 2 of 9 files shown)


tdraier:

- #28470 ◀

thervier (approved):
LGTM

PR #28472, opened 2026-07-05
Add new metrics version with compound keys, remove dead code for analytics

tdraier:
## Description

Adds `v2` variants of the two AWU billing metrics ("LLM Provider Cost AWU", "Tool Invocations") carrying a compound `api_key_name` + `usage_type` (+ `tool_category`) group key, and cleans up dead analytics code found while investigating.

**Context:** Metronome only evaluates a spend alert if every property used in the alert's `group_filter` and in the billed product's `pricing_group_key` is present together in *one* compound group key on the metric — confirmed with Metronome support after an `api_key_name`-filtered spend alert silently never fired despite real spend over threshold ([Slack thread](https://dust4ai.slack.com/archives/C0AGNMLBRB5/p1782740593425549)). The existing metrics only had `api_key_name` and `usage_type` as separate bare group keys, never combined, which is why the alert never fired.

**Why new metrics instead of editing the existing ones:** the two live billing metrics are already at Metronome's group-key cap, and changing a metric's `group_keys` forces Metronome to archive + recreate it, cascading to dependent products/rate cards and requiring a historical-usage reflow negotiated with Metronome support. Since the live metrics can't be recreated without that reflow, `LLM Provider Cost AWU v2` / `Tool Invocations v2` are added as new, separate metric definitions in `setup_new_pricing.ts`. `v2` also carries the bare `[usage_type(, tool_category)]` and `[api_key_name(, tool_category)]` compound keys (matching what the two live metrics already declare today) so that `programmatic_awu_usage.ts` and `per_api_key_usage.ts` — both left completely unchanged in this PR — keep working identically whichever metric ID `getMetricLlmProviderCostAwuId()` / `getMetricToolInvocationsId()` resolve to.

**`front/scripts/metronome_setup.ts`:** a product's `billable_metric_id`, `pricing_group_key`, and `presentation_group_key` are now reconciled in place via `products.update` instead of forcing an archive/recreate (confirmed supported directly in the Metronome SDK — only `type` genuinely requires recreation). Previously, swapping a product's billing metric would recreate the product, which cascaded into recreating every rate card referencing it (seen in sandbox: `Standard EUR: product AI Usage ... not found in existing rates`). This is needed for the eventual product cutover (see Deploy Plan) to be a safe in-place edit rather than a recreate.

**Dead code removal**, found while auditing which of the metrics' group keys were actually used:
- Deleted `front/lib/api/analytics/metronome_usage.ts` — `getMetronomeUsage` and its supporting types/schemas had zero callers anywhere in the codebase (confirmed by exhaustive grep). The two still-used helpers (`getMetronomeWindowSize`, `aggregateToWindowBuckets`) were moved into `awu_usage.ts` (since removed entirely by #28477, migrated to ES-backed analytics).

## Current deployment state

- `v2` metrics are **created in dev only** (`DEV_METRIC_TOOL_INVOCATIONS_V2` / `DEV_METRIC_LLM_PROVIDER_COST_AWU_V2` in `constants.ts`). `PROD_METRIC_TOOL_INVOCATIONS` / `PROD_METRIC_LLM_PROVIDER_COST_AWU` are untouched — **prod is still fully on the original metrics**.
- `setup_common.ts`'s `AI Usage` / `Tool Usage` products still declare `billable_metric_name: "LLM Provider Cost AWU"` / `"Tool Invocations"` (the originals) in this PR — no product is repointed to `v2` in code.
- In dev/sandbox only, the `AI Usage` / `Tool Usage` products were manually repointed to the `v2` metric IDs directly against Metronome (outside of this PR's code) to end-to-end test: seat-based credit consumption, invoicing, per-user usage, per-API-key usage, the `api_key_name` spend alert firing/resolving, the programmatic cap, and the per-user cap. All confirmed working against `v2`.
- Since prod hasn't moved, `per_api_key_usage.ts` and `programmatic_awu_usage.ts` (which query `getMetricToolInvocationsId()` / `getMetricLlmProviderCostAwuId()` directly) were double-checked to still resolve correctly against the **original** metrics in prod today — confirmed unchanged from `main`.

## Tests

Typechecked (`tsgo --noEmit`, both `front` and `front-api`) and formatted (`biome check`) clean. Manually end-to-end tested against dev/sandbox Metronome as described above (dry-run + `--execute` against `metronome_setup.ts`, real usage traffic, real alert firing). No automated tests added: the `v2` metrics aren't wired to any product in code yet (no prod billing behavior change from this PR alone).

## Risk

Low for this PR as merged: the two live billing metrics and the products that reference them are completely untouched in prod — verified line-by-line against `main`. The `v2` metrics are net-new and unattached in code, so they have zero effect on prod until a follow-up explicitly repoints the products.

## Deploy Plan

1. Merge this PR (creates the `v2` metrics in dev only, no prod change).
2. Run `metronome_setup.ts --execute` in **prod** to create the `v2` metrics there too (still unattached — no billing impact).
3. Ask Metronome support to backfill/reflow historical usage into the prod `v2` metrics before any cutover, so historical reporting stays continuous once switched.
4. Only after the backfill is confirmed: follow-up PR repoints `AI Usage` / `Tool Usage`'s `billable_metric_name` to the `v2` metrics in prod (a safe in-place `products.update` edit now, not a recreate, thanks to the `metronome_setup.ts` change in this PR) and points the `api_key_name` spend alert at the new compound key.


tdraier:

- #28472 ◀

rfrenoy (approved):
lgtm

PR #28473, opened 2026-07-06
[front] chore: allow `ask_user_question` for all origins consumed via webapp

aubin-tchoi:
## Description

This PR allows using the `ask_user_question` tool from user message origins that are consumed via the webapp.
Goal is to reduce entropy on the list of tools available, which causes errors notably when using tool search.
Also turns the allow-list into a blacklist to prevent getting similar issues when a new origin is added.

## Tests

- Checked locally (wakeup)

## Risk

- Low.

## Deploy Plan

- Deploy front.

### DIFF
--- a/front/temporal/agent_loop/lib/run_model.ts
+++ b/front/temporal/agent_loop/lib/run_model.ts
@@ -84,11 +84,26 @@ import { startActiveObservation } from "@langfuse/tracing";
 import { Context, heartbeat } from "@temporalio/activity";
 import assert from "assert";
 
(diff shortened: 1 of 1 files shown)


front/temporal/agent_loop/lib/run_model.ts
```diff
@@ -89,6 +89,10 @@ const ASK_USER_QUESTION_ALLOWED_ORIGINS: UserMessageOrigin[] = [
   "slack",
   "extension",
   "agent_sidekick",
+  "project_kickoff",
```
flvndvd: Can we do a black list approach instead? To ensure we don't forget to add when adding new origin?

flvndvd (approved):

PR #28476, opened 2026-07-06
[front-api] Add early startup "signs of life" logging

davidebbo:
## Problem

We sometimes get startup probe failures when `front` / `front-sse` boot:

```
Startup probe failed: Get "http://.../api/healthz/startup": dial tcp ...:3000: connect: connection refused
```

When that happens we see **no log whatsoever** for the process in Datadog.

Both `front` and `front-sse` run the same binary — the esbuild-bundled Hono server `front-api/dist/server.js` (from `front-api/server.ts`). The only existing startup log, `"front-api server listening"`, fires inside the `serve()` callback — i.e. **after** OpenTelemetry init and the entire `./app` module graph have loaded. Because ESM imports are hoisted, no statement in `server.ts` runs before `./app` is fully imported, so a hang or crash during module init produces zero logs.

## Change

Add a boot checkpoint ladder (each fires once per process start). The **last** log seen localizes where a stuck/crashing boot died:

| Log | Where | Fires when |
|---|---|---|
| `front-api process starting up` (`elapsedMs` since process start) | new `lib/startup-log.ts`, imported **first** | before OTEL setup + app graph — only depends on pino |
| `front-api instrumentation initialized` | end of `lib/instrumentation-config.ts` | OTEL init done, app graph about to load |
| `front-api starting HTTP server` | `server.ts`, right before `serve()` | full app graph imported OK, about to bind |
| `front-api server listening` (existing) | `serve()` callback | socket bound |

Interpretation:
- **none** → died before/at Node bootstrap, the dd-trace preload, or pino import
- **only (1)** → tracer/OTEL instrumentation init hung or threw
- **(1) + (2)** → the `./app` route/module graph import hung or threw
- **(1)–(3), no (4)** → socket bind stalled

No Dockerfile/CMD changes needed — these are plain in-bundle log statements.

### DIFF
--- a/front-api/lib/instrumentation-config.ts
+++ b/front-api/lib/instrumentation-config.ts
@@ -1,4 +1,5 @@
 import { initializeOpenTelemetryInstrumentation } from "@app/lib/api/instrumentation/init";
+import logger from "@app/logger/logger";
 
 // Initialize OpenTelemetry / Langfuse instrumentation for the front-api process.
 // Without this, any `startActiveObservation` created while handling an HTTP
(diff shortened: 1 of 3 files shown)


id13 (commented):
LGTM

id13 (approved):

PR #28477, opened 2026-07-06
[front] Remove Metronome-backed getAwuUsage (migrate read-only usage summary to ES analytics)

tdraier:
## Summary

`getAwuUsage` (the Metronome usage-API path, incl. the dead `includeFreeUsage` opt-in) had a **single** consumer: the workspace Usage page's read-only "credits spent this period" summary, gated behind `!isCreditPriced && usage_page_read_only`.

That read-only mode targets legacy / non-Metronome workspaces (which may have no Metronome usage data), and the chart it sits next to already reads the ES-backed analytics endpoint. So this migrates the summary to `useAwuUsageFromAnalytics` (ungrouped `total`, summed over the current billing cycle) and removes the now-unused Metronome path entirely — cascading down to the metric definitions that only that path used.

## Changes

- Migrate the read-only period-spend summary in `UsagePage.tsx` to `useAwuUsageFromAnalytics`.
- Delete `front/lib/api/analytics/awu_usage.ts` (`getAwuUsage`, `AwuUsageQuerySchema`, response types, `includeFreeUsage`).
- Delete the `/api/w/:wId/analytics/awu-usage` front-api route (+ test) and its mount in `analytics/index.ts`.
- Delete the `useAwuUsage` SWR hook.
- Remove the now-orphaned `(non-free)` AWU / tool-invocation **metric getters** and their DEV/PROD id constants from `constants.ts` (`awu_usage.ts` was their only consumer).
- Drop the two `(non-free)` **metric definitions** from `setup_new_pricing.ts` — they only ever powered the free-excluding graph via those getters, carry no billing meaning, and are not attached to any product/rate.

## Notes

- No behavior change for credit-priced workspaces — the affected summary is read-only / legacy only.
- The all-usage AWU metric getters in `constants.ts` are untouched (still used by per-user / per-api-key / programmatic usage).


rfrenoy (approved):
LGTM, though I was wondering if we should keep Metronome as a fallback in case we have an issue with ES? It's not considered an infra critical component (we should be able to live without it for a short while) and following our increasing use of ES to display usage, I think that is no longer true

tdraier:
> LGTM, though I was wondering if we should keep Metronome as a fallback in case we have an issue with ES? It's not considered an infra critical component (we should be able to live without it for a short while) and following our increasing use of ES to display usage, I think that is no longer true

I'm cleaning that in order to be able to create clean metric and not having to maintain too many - metronome as fallback for analytics is not really a viable option ( need to have all configured group keys)

PR #28479, opened 2026-07-06
[front] chore: log missing replayed tools

aubin-tchoi:
## Description

This PR adds logging in the agent loop for replayed tools that are missing from the current action specifications. Replayed tools are collected from conversation-history function calls and Anthropic `tool_search_tool_result` passthrough references.

Next step is to analyze these logs, and potentially add dummy tool specifications for these.
End goal is to make sure the list of tools is append-only.

One issue we face when removing tool specifications is that a replayed `tool_search_tool_result` whose `tool_reference` points to a missing tool errors with 400: `Tool reference 'archived_lookup_tool' not found in available tools`.

Tested Anthropic tool-search replay behavior locally with the API: adding a dummy tool with the same name and dummy description/schema is accepted by Anthropic.

## Tests

- Tested locally.

## Risk

- Low. Logging only.

## Deploy Plan

- Deploy front.

### DIFF
--- a/front/temporal/agent_loop/lib/run_model.ts
+++ b/front/temporal/agent_loop/lib/run_model.ts
@@ -135,7 +135,7 @@ function getReplayedToolNames(
     }
   }
 
-  return Array.from(toolNames);
+  return [...toolNames];
 }
 
 // This method is used by the multi-actions execution loop to pick the next action to execute and
--- a/front/temporal/agent_loop/lib/run_model.ts
+++ b/front/temporal/agent_loop/lib/run_model.ts
@@ -74,7 +74,10 @@ import type {
   AgentMessageType,
   UserMessageOrigin,
 } from "@app/types/assistant/conversation";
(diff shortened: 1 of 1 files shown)


front/temporal/agent_loop/lib/run_model.ts
```diff
@@ -124,6 +112,32 @@ function concatWithNewlineBoundary(
...
 }
 
+function getReplayedToolNames(
+  modelConversation: ModelConversationTypeMultiActions
+): string[] {
+  const toolNames = new Set<string>();
+
+  for (const message of modelConversation.messages) {
+    switch (message.role) {
+      case "assistant":
```
flvndvd, line 124: We should also capture the tools surfaced through the tool server search results used for searching. If my understanding is correct, those are the one messing up.

aubin-tchoi: indeed, added!

flvndvd (approved):

PR #28480, opened 2026-07-06
fix(labs-transcripts): don't delete retrieval schedule on transient errors

rfrenoy:
## Description

Ticket: https://github.com/dust-tt/tasks/issues/9004

Meeting Transcript Processing silently stops for users when a transient error deletes their Temporal retrieval schedule with no auto-recovery. A Temporal schedule keeps firing regardless of whether individual runs fail, so the only reason a momentary failure becomes permanent is the explicit stopRetrieveTranscriptsWorkflow() teardown.

This PR removes the teardown from the two every-5-min external-call paths:

- getTranscriptsGoogleAuth: only tear down when the token is genuinely revoked (token_revoked_error). Transient OAuth/network errors now throw without deleting the schedule, which self-heals on the next tick.
- retrieveNewTranscriptsActivity: the Drive files.list failure path no longer deletes the schedule. It fails the run (schedule survives).

Also adds two read-only admin diagnostics for on-call:
- check_transcripts_connections.ts: per-config token validity, scope, history recency, and live-schedule status, bucketed in a summary.
- check_transcripts_pending.ts: previews the retrieval query (Drive listing + history dedup, no processing) to count transcripts a config would pick up on its next run.

## Tests

Local

## Risk

Low

## Deploy Plan

Front

### DIFF
--- a/front/admin/check_transcripts_connections.ts
+++ b/front/admin/check_transcripts_connections.ts
@@ -1,0 +1,215 @@
+import config from "@app/lib/api/config";
+import { getOAuthConnectionAccessToken } from "@app/lib/api/oauth_access_token";
+import { Authenticator } from "@app/lib/auth";
+import { LabsTranscriptsConfigurationResource } from "@app/lib/resources/labs_transcripts_resource";
+import { getTemporalClientForFrontNamespace } from "@app/lib/temporal";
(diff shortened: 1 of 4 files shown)


Fraggle (approved):
LGTM

PR #28481, opened 2026-07-06
[Admin Gov] Phase 0.1 — group_permissions model + migration

philipperolet:
## Description

Admin Governance — Phase 0, PR 1/9. Foundations of the single `group_permissions` table: the Sequelize model (`WorkspaceAwareModel`), the permission/resource vocabulary types (`front/types/group_permissions.ts`), and the pre-deploy migration. Nothing is wired into existing code, so there is **zero production behavior change**.

Schema: unique index `(groupId, permissionType, resourceType, resourceId)` — `groupId` already scopes the workspace and doubles as the FK index (BACK13); a `(workspaceId, resourceType, resourceId)` index for the "who can act on X" direction; and a CHECK that the instance-less domains (`billing`/`identity`/`audit_log`) and the `*` wildcard require `resourceId = -1`.

First of a git-spice stack (registry → resource → governance toggle → auth check → shadow helper → seeding hook).

Context: [Design Doc](https://app.notion.com/p/dust-tt/Design-Doc-Admin-Governance-38a28599d941817292d0e94f8dfafe48) · [Implementation Plan](https://app.notion.com/p/dust-tt/Implementation-Plan-Admin-Governance-39528599d9418153aa9bf3dd69d5448d)

## Risks

Blast radius: adds an unused table + model; no reads or writes from existing code.
Risk: none

## Deploy Plan

- run migration `20260706113819_create_group_permissions.sql` (pre-deploy, additive)
- deploy front

PR #28484, opened 2026-07-06
[Admin Gov] Phase 0.2 — group_permissions validity registry

philipperolet:
## Description

Admin Governance — Phase 0, PR 2/9. Follows #28481.

Adds the static validity registry (`front/lib/resources/group_permission_registry.ts`) — the single source of truth for which `(permissionType, resourceType, resourceId)` combinations are representable. `assertValidGrant(...)` **fails fast** (throws) on invalid combinations; invalid writes are programmer errors, so no `Result`. Encodes the validity table: `space` (read/write/admin), `agent`/`skill` (read/write/create/publish, with `create` type-level only), `frame` (invite/publish, instance-less), `billing`/`identity` (admin, instance-less), `audit_log` (read, instance-less). `"*"` wildcards are always type-level (`-1`).

Not wired into any write path yet (that's PR 3) — no behavior change.

Context: [Design Doc](https://app.notion.com/p/dust-tt/Design-Doc-Admin-Governance-38a28599d941817292d0e94f8dfafe48) · [Implementation Plan](https://app.notion.com/p/dust-tt/Implementation-Plan-Admin-Governance-39528599d9418153aa9bf3dd69d5448d)

## Risks

Blast radius: new pure-function module, not yet called.
Risk: none

## Deploy Plan
- pmrr
- deploy front

### DIFF
--- a/front/lib/resources/group_permission_registry.ts
+++ b/front/lib/resources/group_permission_registry.ts
@@ -1,0 +1,115 @@
+import type {
+  GroupPermissionResourceType,
+  PermissionType,
+} from "@app/types/group_permissions";
+import { WHOLE_TYPE_RESOURCE_ID } from "@app/types/group_permissions";
+import assert from "assert";
+
+/**
+ * Static source of truth for `group_permissions` validity.
+ *
+ * The table is polymorphic, so its representable space is larger than its valid space. Every write
+ * path in `GroupPermissionResource` validates against this registry and throws on violation — invalid
+ * writes are programmer errors, so we fail fast rather than returning a `Result`. The DB CHECK only
+ * covers the coarse instance-less-domain rule; finer per-verb rules (e.g. `create` ⇒ `-1` only) live
+ * here.
+ *
+ * The `"*"` wildcards in the vocabulary are intentionally not part of the per-type table: a wildcard
+ * grant applies to the whole type (or all types) and is only ever a type-level (`-1`) grant.
+ */
+
+// Concrete (non-wildcard) vocabulary — the registry describes these.
+type ConcreteResourceType = Exclude<GroupPermissionResourceType, "*">;
+type ConcretePermissionType = Exclude<PermissionType, "*">;
+
+interface ResourceTypeRule {
+  // Verbs grantable on a specific instance (resourceId > 0).
+  instanceLevelPermissions: ConcretePermissionType[];
+  // Verbs grantable type-wide (resourceId = -1). For types with instances this includes the
+  // instance verbs (a -1 grant covers all resources) plus type-only verbs like "create"; the two
+  // lists therefore overlap. Instance-less domains have an empty instanceLevelPermissions and list
+  // their verbs here only.
+  typeLevelPermissions: ConcretePermissionType[];
+}
+
+const REGISTRY: Record<ConcreteResourceType, ResourceTypeRule> = {
+  space: {
+    instanceLevelPermissions: ["read", "write", "admin"],
+    typeLevelPermissions: ["read", "write", "admin"],
+  },
+  agent: {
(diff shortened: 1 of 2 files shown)


philipperolet:

- #28481
    - #28484 ◀
        - #28485
            - #28539
                - #28486
                    - #28487
                        - #28501
                    - #28502

front/lib/resources/group_permission_registry.ts
```diff
@@ -0,0 +1,117 @@
...
+ *
+ * The `"*"` wildcards in the vocabulary are intentionally not part of the per-type table: a wildcard
+ * grant applies to the whole type (or all types) and is only ever a type-level (`-1`) grant.
+ */
+
+// Concrete (non-wildcard) vocabulary — the registry describes these.
+type ConcreteResourceType = Exclude<GroupPermissionResourceType, "*">;
+type ConcretePermissionType = Exclude<PermissionType, "*">;
+
+interface ResourceTypeRule {
```
philipperolet, line 25: would it make sense to name fields "instanceLevelPermissions", "typeLevelPermissions" (with possible non-null intersection) and to enforce "instanceLevelPermissions = [] if hasInstances = false?

philipperolet:
FYI @zmarouf @ykmsd

front/lib/resources/group_permission_registry.ts
philipperolet: Done — reworked to `instanceLevelPermissions` + `typeLevelPermissions` (overlapping, as you suggested). `hasInstances` is gone: instance-less domains are exactly the ones with an empty `instanceLevelPermissions`, so it was redundant. Validation is now a direct membership check keyed on the resourceId kind:

- `resourceId > 0` → must be in `instanceLevelPermissions`
- `resourceId = -1` → must be in `typeLevelPermissions`

e.g. `agent` = instance `[read, write, publish]`, type-wide `[read, write, publish, create]` (the `-1`-only `create` is the intersection difference). Tests still green.

front/lib/resources/group_permission_registry.ts
```diff
@@ -0,0 +1,115 @@
...
+  audit_log: {
+    instanceLevelPermissions: [],
+    typeLevelPermissions: ["read"],
+  },
+};
+
+interface GrantSpec {
+  permissionType: PermissionType;
+  resourceType: GroupPermissionResourceType;
+  resourceId: number;
```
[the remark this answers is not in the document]
philipperolet: Keeping `resourceId`. It's a Sequelize column/attribute, and the codebase names numeric id/FK attributes `xId` (in this same table `groupId` and `workspaceId` are numeric `ForeignKey`s named `xId`, not `xModelId`). BACK10's `ModelId` suffix targets business-logic variables (to disambiguate from string sIds), not column names. Renaming just this param would diverge from the `group_permissions.resourceId` column (already applied in prod); the only consistent alternative is a column-rename migration across the stack, which isn't worth it for an internal, never-URL-exposed column. Keeping it consistent with its sibling columns.

front/lib/resources/group_permission_registry.ts
```diff
@@ -0,0 +1,115 @@
...
+  space: {
+    instanceLevelPermissions: ["read", "write", "admin"],
+    typeLevelPermissions: ["read", "write", "admin"],
+  },
+  agent: {
+    instanceLevelPermissions: ["read", "write", "publish"],
+    typeLevelPermissions: ["read", "write", "publish", "create"],
+  },
+  skill: {
+    instanceLevelPermissions: ["read", "write", "publish"],
```
matteotrab, line 45: When do we plan to use the publish permission with a specific skill instance?

Same question for agents

matteotrab (commented):
LGTM left a comment

PR #28486, opened 2026-07-06
[Admin Gov] Phase 0.3b — type-wide and batch group_permission writes

philipperolet:
## Description

Admin Governance — Phase 0, PR 3b/9. Follows #28485.

Adds the dedicated type-wide (`resourceId = -1`) writers to `GroupPermissionResource`, kept separate from the instance CRUD so the "grant the whole type" path is explicit and reviewable:
- `grantTypeWide(...)` — single-group convenience over `grantTypeWideForGroups`; the only sanctioned way to write a `-1` row (never a defaulted param). Idempotent, registry-validated.
- `grantTypeWideForGroups(...)` — batch of the above (one INSERT, unique index dedupes); backs the governance `setGroups` transition without an N+1.
- `revokeTypeWide(...)` — removes a group's `-1` row.
- `grantMany(...)` — batch instance grants; each validated, `-1` rejected as in `grant`.

No call sites — no behavior change.

Context: [Design Doc](https://app.notion.com/p/dust-tt/Design-Doc-Admin-Governance-38a28599d941817292d0e94f8dfafe48) · [Implementation Plan](https://app.notion.com/p/dust-tt/Implementation-Plan-Admin-Governance-39528599d9418153aa9bf3dd69d5448d)

## Risks

Blast radius: new resource methods, not yet called from existing code.
Risk: none

## Deploy Plan

- deploy front

### DIFF
--- a/front/lib/resources/group_permission_resource.ts
+++ b/front/lib/resources/group_permission_resource.ts
@@ -30,6 +30,13 @@ interface InstanceGrantSpec {
   transaction?: Transaction;
 }
 
+interface TypeWideGrantSpec {
+  group: GroupResource;
+  permissionType: PermissionType;
+  resourceType: GroupPermissionResourceType;
+  transaction?: Transaction;
+}
+
 interface ListForGroupsSpec {
   groupModelIds: ModelId[];
   permissionType?: PermissionType;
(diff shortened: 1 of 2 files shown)


philipperolet:

- #28481
    - #28484
        - #28485
            - #28539
                - #28486 ◀
                    - #28669
                        - #28501
                    - #28502

front/lib/resources/group_permission_resource.ts
```diff
@@ -192,6 +199,133 @@ export class GroupPermissionResource extends BaseResource<GroupPermissionModel>
     });
   }
 
+  // Grant a type-wide permission (resourceId = -1: the whole type). Dedicated, explicitly named so a
+  // defaulted -1 can never silently reach `grant`. Idempotent. Used for type-level verbs (e.g.
+  // "create") and for governance capabilities (see the governance-toggle methods).
+  static async grantOnAllResourcesOfType(
```
philipperolet: can this just be a special case of the function below, called with [group]?

front/lib/resources/group_permission_resource.ts
```diff
@@ -30,6 +30,13 @@ interface InstanceGrantSpec {
   transaction?: Transaction;
 }
 
+interface TypeWideGrantSpec {
+  group: GroupResource;
+  permissionType: PermissionType;
+  resourceType: GroupPermissionResourceType;
+  transaction?: Transaction;
```
philipperolet, line 37: what about extracting transaction from the spec, as it's a cross concern? iirc a common pattern we have is functionName(auth, { mainFunctionArgs}, {options}) with options containing transaction?

front/lib/resources/group_permission_resource.ts
philipperolet: the name is a bit edgy since e.g. for "billing" it's not "all resources", there's no resource instance...
wdyt? what about grantResourceTypeWideAccess? or maybe there are better options?

front/lib/resources/group_permission_resource.ts
```diff
@@ -192,6 +199,133 @@ export class GroupPermissionResource extends BaseResource<GroupPermissionModel>
...
+        resourceType,
+        resourceId: WHOLE_TYPE_RESOURCE_ID,
+      },
+      transaction,
+    });
+  }
+
+  // Batch of instance-level grants (one INSERT, unique index dedupes). Each is validated; -1 is
+  // rejected here as in `grant` — type-wide grants use the dedicated methods above.
+  static async grantMany(
```
philipperolet, line 275: same here maybe we can simplify grant so it calls grantMany, and maybe we can rename batchgrant? not sure, assess

front/lib/resources/group_permission_resource.ts
philipperolet: Done — `grantTypeWide` is now a thin wrapper over `grantTypeWideForGroups([group])`, so validation + write live in one place.

philipperolet: Renamed to `grantTypeWide` / `grantTypeWideForGroups` / `revokeTypeWide` — reads correctly for instance-less domains too (resourceId = -1 = the whole type).

front/lib/resources/group_permission_resource.ts
philipperolet: Keeping `transaction` in the spec: the codebase convention is transaction-in-args (21 resources do this; none use a separate options arg), so matching it per GEN1. Happy to revisit if there's a specific precedent you had in mind.

front/lib/resources/group_permission_resource.ts
philipperolet: Leaving as-is: `grant` is a single-row `findOrCreate` (returns the row) vs `grantMany`'s `bulkCreate`/`ignoreDuplicates` — genuinely different ops; and `grant`/`grantMany` are on the already-merged PR3a, so consolidating would be a separate PR against main. `grantMany` reads clearly, so keeping the name.

matteotrab (approved):

PR #28488, opened 2026-07-06
refactor(front): move agent deselect into AgentPicker selected section

Nils-Fedrigo:
## Description

After
Screenshot 2026-07-06 at 13 34 24

Before
image

Reworks the agent chip in the conversation input bar to move deselection into the `AgentPicker` dropdown, in preparation for the model picker that sits next to it in the stack.

- The currently selected agent is pinned at the top of the `AgentPicker` list under a **Selected** label, with an inline toggle to deselect it.
- Drops the separate `XClose` button that used to sit on the agent chip — deselection now happens from inside the picker. **Only 3% of convos did not ping a user as the first message and anyway when you @ someone, the agent deselects.**
- On mobile, the agent label is hidden so the chip shows only the avatar, leaving room for the other input-bar buttons.
Screenshot 2026-07-06 at 13 36 42

Part of the model-picker stack (bottom → top): #28344 → #28346 → **#28488 (this)** → #28489 → #28049.

## Tests

Manual: open the input bar, select an agent, reopen the picker and confirm it appears under "Selected" with a working deselect toggle. Check the chip renders avatar-only at mobile widths and full-label on desktop.

## Risk

Low, but user-facing and **not** feature-flagged: the agent-chip interaction changes for everyone. Purely presentational, safe to roll back by reverting.

## Deploy Plan

None.

### DIFF
--- a/front/components/assistant/AgentPicker.tsx
+++ b/front/components/assistant/AgentPicker.tsx
@@ -35,8 +35,6 @@ interface AgentPickerProps {
   disabled?: boolean;
   mountPortal?: boolean;
   onOpenChange?: (open: boolean) => void;
-  // When set, the matching agent is pinned at the top of the list under
-  // "Selected" with a toggle to deselect it (via `onDeselect`).
   selectedAgentId?: string | null;
   onDeselect?: () => void;
 }
--- a/front/components/assistant/AgentPicker.tsx
+++ b/front/components/assistant/AgentPicker.tsx
@@ -7,13 +7,16 @@ import type { LightWorkspaceType } from "@app/types/user";
 import {
   Avatar,
   Button,
+  Check,
   DotsHorizontal,
   DropdownMenu,
   DropdownMenuContent,
   DropdownMenuItem,
+  DropdownMenuLabel,
   DropdownMenuSearchbar,
   DropdownMenuSeparator,
   DropdownMenuTrigger,
+  Icon,
   Robot,
 } from "@dust-tt/sparkle";
 import { useState } from "react";
@@ -32,6 +35,10 @@ interface AgentPickerProps {
   disabled?: boolean;
   mountPortal?: boolean;
   onOpenChange?: (open: boolean) => void;
+  // When set, the matching agent is pinned at the top of the list under
+  // "Selected" with a toggle to deselect it (via `onDeselect`).
+  selectedAgentId?: string | null;
+  onDeselect?: () => void;
 }
 
 export function AgentPicker({
@@ -47,13 +54,22 @@ export function AgentPicker({
   isLoading = false,
   disabled = false,
   onOpenChange,
+  selectedAgentId,
+  onDeselect,
 }: AgentPickerProps) {
   const clientType = useClientType();
   const isMobile = useIsMobile();
   const [searchText, setSearchText] = useState("");
   const [isOpen, setIsOpen] = useState(false);
 
-  const searchedAgents = filterAndSortAgents(agents, searchText);
+  const selectedAgent = selectedAgentId
+    ? (agents.find((a) => a.sId === selectedAgentId) ?? null)
+    : null;
+  // The selected agent is pinned separately at the top, so keep it out of the
+  // searchable list below to avoid showing it twice.
+  const searchedAgents = filterAndSortAgents(agents, searchText).filter(
+    (a) => a.sId !== selectedAgent?.sId
+  );
 
   return (
     <DropdownMenu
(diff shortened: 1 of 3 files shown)


Nils-Fedrigo:

- #28344
    - #28346
        - #28488 ◀
            - #28489
                - #28049
                    - #28595

front/components/assistant/AgentPicker.tsx
```diff
@@ -32,6 +35,10 @@ interface AgentPickerProps {
   disabled?: boolean;
   mountPortal?: boolean;
   onOpenChange?: (open: boolean) => void;
+  // When set, the matching agent is pinned at the top of the list under
+  // "Selected" with a toggle to deselect it (via `onDeselect`).
```
Nils-Fedrigo:
```suggestion
```

PopDaph (commented):
@Nils-Fedrigo Could you give me some context on the rationale behind this change? 🙏🏻

I'm coming in with limited context on the project, so from where I stand, this feels like it adds friction to removing the agent without making that action very discoverable. It's not immediately obvious that you need to re-select the same agent to remove it. If the goal is intentionally to add some friction, would it make sense to at least include a cross icon inside the dropdown? Was a designer involved in this decision? 🙏🏻

On the 3% figure: does that exclude users who deselected the agent but then had it added back as the default in a new version of the message?

spolu:
@Nils-Fedrigo do we have a figma to discuss the exact experience we want to build here?

Nils-Fedrigo:
@PopDaph
thanks for your comments
Working version of figma is [here](https://www.figma.com/design/wJJMfVF6bluurSKfrEuysc/Product---WIP?node-id=12045-28674&t=nVytqqYnL09z7RPf-4) but probably does not address your concerns
So you can check the final state in the video [here](https://github.com/dust-tt/dust/pull/28049)

The designer in the loop was @ClementAupiais 

more tokens as to why I think it's a net positive change and does not result in too much friction
- If I ping someone in a convo in a first message, it deselects the agent automatically (no added click)
- So the only real tradeoff is someone deselecting agent, then pinging someone else (adds 1 click). Granted this is 3% of users, I argue this is an ok tradeoff
- We really really need this lateral real estate to add the model picker in the input bar (🙏)

query

WITH total_messages AS (
    SELECT COUNT(*) AS cnt
    FROM DUST_MODELS.FACT_USER_MESSAGES fum
    JOIN FRONT.MESSAGES m
        ON m.usermessageid = fum.user_message_id
    WHERE m.contentfragmentid IS NULL
        AND fum.source NOT IN ('api', 'cli_programmatic', 'gsheet', 'make', 'n8n', 'zapier', 'triggered_programmatic', 'slack_workflow')
        AND fum.created_at >= '2026-01-01'
),
no_agent_messages AS (
    SELECT COUNT(*) AS cnt
    FROM DUST_MODELS.FACT_USER_MESSAGES fum
    JOIN FRONT.MESSAGES m
        ON m.usermessageid = fum.user_message_id
    LEFT JOIN FRONT.MENTIONS men
        ON men.messageid = m.id
        AND men.agentconfigurationid IS NOT NULL
    WHERE men.id IS NULL
        AND m.contentfragmentid IS NULL
        AND fum.source NOT IN ('api', 'cli_programmatic', 'gsheet', 'make', 'n8n', 'zapier', 'triggered_programmatic', 'slack_workflow')
        AND fum.created_at >= '2026-01-01'
)
SELECT
    t.cnt                                                          AS total_messages,
    n.cnt                                                          AS messages_without_agent,
    t.cnt - n.cnt                                                  AS messages_with_agent,
    ROUND(n.cnt * 100.0 / NULLIF(t.cnt, 0), 2)                    AS pct_without_agent,
    ROUND((t.cnt - n.cnt) * 100.0 / NULLIF(t.cnt, 0), 2)          AS pct_with_agent
FROM total_messages t, no_agent_messages n;
Screenshot 2026-07-07 at 14 57 20

Nils-Fedrigo:
Discussed live with @PopDaph, I'll add a different bg color to improve discoverability that you can remove the agent, and on hover change to X the tick. 

We concluded it's not that big of an issue to remove the X button (if I understood correctly)

Nils-Fedrigo:
Screenshot 2026-07-07 at 16 49 48
on hover
Screenshot 2026-07-07 at 16 50 00

front/lib/utils.ts
```diff
@@ -318,6 +319,19 @@ export function filterAndSortAgents(
     );
   }
 
+  // Pin the selected agent to the front so it always displays first.
```
PopDaph, line 322: Question: 

`filterAndSortAgents` is also called from backend code (`lib/api/assistant/email/email_trigger.ts`, `lib/api/mcp_server/tools/agents/list.ts`).It's a shared name-matching helper, not a picker helper. "Pin the selected agent" is pure dropdown presentation, so it feels like it doesn't belong in the shared util, even opt-in.

Should we do the pinning in `AgentPicker` after the call instead?

```ts
const searched = filterAndSortAgents(agents, searchText);
const selected = searched.find((a) => a.sId === selectedAgentId);
const searchedAgents = selected
  ? [selected, ...searched.filter((a) => a.sId !== selectedAgentId)]
  : searched;
```

Also simpler than a full `.sort()` to move one element to the front. I'd say keep the util generic, does it make sense? 

Somewhat related: should the selected agent stay pinned while actively searching? If I type a query looking for a *different* agent, the current one overriding fuzzy relevance at slot #1 reads as noise. Pinning only when `searchText` is empty might be the better behavior, wdyt?

PopDaph (approved):

front/lib/utils.ts
Nils-Fedrigo: yep, claude did it like that in the beginning, but I opinionated otherwise.

Anyway, I'll move it out!

PopDaph: Feel free to keep as is if it was intentional, it's your PR you decide!! 🙏🏻

PopDaph: (PR is already approved 😬 )

PR #28490, opened 2026-07-06
[front/viz] enable callFunction for shared frames for authenticated members

TimotheeCrouzet:
## Description

Enables `callFunction` (sandbox function RPC) on shared/public frames when the
viewer is an authenticated workspace member. Previously it was hard-blocked for
everyone on shared frames.

The authenticated-member status is passed **explicitly** instead of being decoded
from the viz access token client-side.

- **front-api `/public/frames`**: `resolveOptionalAuth` resolves the viewer from
  the session cookie; the response returns `isAuthenticatedMember` (`!!user`).
  Front uses it to derive `frameAccess` → the host-side guard.
- **front-api `/viz/content`**: this endpoint already verifies the access token to
  serve the frame code; it now also returns `isAuthenticatedMember`
  (`!!tokenPayload.userId`) in the same response.
- **viz**: `ServerSideVisualizationWrapper` (server component) reads
  `isAuthenticatedMember` from the `/viz/content` fetch it already makes
  (server-to-server), and passes it to `ServerVisualizationWrapperClient`, which
  selects `HybridDataAPI` (SSR cache for reads, `callFunction` over RPC) instead
  of `CacheDataAPI`. No URL param, no token decode.

Both fields (optional, append-only) are added to the SDK schemas. The flag is a
UI hint only; real authorization stays server-side (session cookie + pod access)
at invocation.

## Tests

- Authenticated member on a shared frame: `callFunction` executes end-to-end.
- Anonymous / non-member (private window / other Google session): returns
  "Sandbox function <id> is not supported in public frames."

## Risk

**Low.**

- Anonymous/non-member behavior unchanged (safe default `false`).
- The `isAuthenticated` URL param is only a UX hint (unverified, tamper-able);
  actual authorization is server-side (cookie + pod access), untouched. Tampering
  the URL at worst exposes `callFunction` in viz, but the invocation is still
  rejected by the host guard (server-authoritative `frameAccess`) and the
  authenticated endpoint (401/404).
- A member without pod access gets 404.
- safe to revert

## Deploy Plan

Standard deploy.

### DIFF
--- a/viz/app/content/ServerVisualizationWrapper.tsx
+++ b/viz/app/content/ServerVisualizationWrapper.tsx
@@ -97,6 +97,7 @@ interface ServerSideVisualizationWrapperProps {
   identifier: string;
   isFullHeight?: boolean;
   isPdfMode?: boolean;
+  isAuthenticatedMember?: boolean;
 }
 
 /**
@@ -116,6 +117,7 @@ export async function ServerSideVisualizationWrapper({
   identifier,
   isFullHeight = false,
   isPdfMode = false,
+  isAuthenticatedMember = false,
 }: ServerSideVisualizationWrapperProps) {
   let prefetchedCode: string | undefined;
   let preFetchedFiles: PreFetchedFile[] = [];
@@ -168,6 +170,7 @@ export async function ServerSideVisualizationWrapper({
       isPdfMode={isPdfMode}
       prefetchedCode={prefetchedCode}
       prefetchedFiles={preFetchedFiles}
+      isAuthenticatedMember={isAuthenticatedMember}
     />
   );
 }
--- a/viz/app/content/ServerVisualizationWrapper.tsx
+++ b/viz/app/content/ServerVisualizationWrapper.tsx
@@ -97,7 +97,6 @@ interface ServerSideVisualizationWrapperProps {
   identifier: string;
   isFullHeight?: boolean;
   isPdfMode?: boolean;
-  isAuthenticatedMember?: boolean;
 }
 
 /**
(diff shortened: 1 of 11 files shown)


front/package.json
```diff
@@ -165,6 +165,7 @@
     "io-ts-types": "^0.5.19",
     "iron-session": "^8.0.4",
     "isomorphic-dompurify": "^2.32.0",
+    "jose": "^6.1.0",
```
flvndvd: Why adding a new package, given that we already have `jsonwebtoken` (see line 171)?

viz/app/content/ServerVisualizationWrapper.tsx
```diff
@@ -163,6 +169,7 @@ export async function ServerSideVisualizationWrapper({
   return (
     <ServerVisualizationWrapperClient
       allowedOrigins={allowedOrigins}
+      canCallFunctions={canCallFunctions}
```
flvndvd: This seems pretty "local" and might fall short tomorrow if need to expose other tools when authenticated as a workspace member. Could this be driven by the interface we expose?

viz/app/content/page.tsx
```diff
@@ -37,6 +44,7 @@ export default function RenderVisualization({
         identifier={identifier!}
         isFullHeight={isFullHeight}
         isPdfMode={isPdfMode}
+        isAuthenticatedMember={isAuthenticated === "true"}
```
flvndvd: `nit` follow existing pattern, and add it.

flvndvd (approved):

PR #28492, opened 2026-07-06
[Agentic User Activation] Migration for Poke provisioning plugin

xebeohp:
## Description (Edited after discussion with Frank and David about Poke plugin design)
Add a column to the `project_metadata` table which stores the pod type. Used to differentiate an activation pod from a typical user-created pod and remains null unless an Activation Pod is created.

This migration is the first step of creating the poke provisioning plugin for Agentic User Activation.

## Tests
Manually tested in my local DB.

## Risk
Low risk, no code depends on this column yet.

## Deploy Plan
Merge then deploy migration.

### DIFF
--- a/front/lib/resources/storage/models/project_metadata.ts
+++ b/front/lib/resources/storage/models/project_metadata.ts
@@ -7,6 +7,8 @@ import { SpaceModel } from "@app/lib/resources/storage/models/spaces";
 import { WorkspaceAwareModel } from "@app/lib/resources/storage/wrappers/workspace_models";
 import type { CreationOptional, ForeignKey } from "sequelize";
 
+export type PodType = "activation";
+
 export class ProjectMetadataModel extends WorkspaceAwareModel<ProjectMetadataModel> {
   declare id: CreationOptional<number>;
   declare createdAt: CreationOptional<Date>;
@@ -24,7 +26,7 @@ export class ProjectMetadataModel extends WorkspaceAwareModel<ProjectMetadataMod
   /** sId of the agent pre-selected for new conversations in this pod. Null = @dust. */
   declare defaultAgentId: CreationOptional<string | null>;
   declare defaultSkillsIds: CreationOptional<string[] | null>;
-  declare activationPodMemberEmail: CreationOptional<string | null>;
+  declare podType: CreationOptional<PodType | null>;
 }
 
 ProjectMetadataModel.init(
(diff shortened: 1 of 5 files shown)


id13 (commented):
Why not using a workspace membership id instead ? 

also, you probably want an index :)

front/lib/resources/storage/models/project_metadata.ts
```diff
@@ -75,6 +76,11 @@ ProjectMetadataModel.init(
       defaultValue: null,
       field: "defaultSkillsIds",
     },
+    activationPodMemberEmail: {
```
frankaloia: I think we can't assume there will be only one member of the activation pod in the future. I'm also generally a bit hesitant to add params for activation that aren't extensible for other use cases.

I think there is a bit of an open question on how do we identify existing activation pods we want to solve, but for the initial poke tool, can we just use the existing mechanisms (ie. name for idempotency)

xebeohp: Ok, maybe we discuss this in our meeting? I assumed that we wanted 1 user to 1 pod for personalized onboarding, but if not I'll need to rework it.

Also isn't name not the best way since the user can easily rename the pod, then the search won't work anymore?

frankaloia: I'm quite confident that we will want to (at some point) experiment with activations pods that are made for n team members.

Perhaps we should make this slightly more generic and go with a `provisioningSource` enum and a provisionedForUserId so that we can look up something like WHERE workspaceId = ? AND provisioningSource = 'activation' AND provisionedForUserId = ?

frankaloia: (then if we do target provisioned for a different ID than user, we can always add it sometime in the future)

xebeohp:

- #28492 ◀

front/lib/resources/storage/models/project_metadata.ts
```diff
@@ -75,6 +78,11 @@ ProjectMetadataModel.init(
       defaultValue: null,
       field: "defaultSkillsIds",
     },
+    podType: {
```
frankaloia: I'm hesitant to add just a generic "type" to the pod as this implies a behavioral contract without explicitly defining it in the code. I prefer that we pin the semantics to the variable name, which is why I was initially thinking `provisioningSource` would be clear

xebeohp: So just rename to `provisioningSource`?

frankaloia: I think so, yes

frankaloia (approved):
LGTM

id13:
LGTM, don't forget to add a partial index on provisioningSource where provisioningSource IS NOT NULL

id13 (approved):

PR #28493, opened 2026-07-06
[front] Normalize sub-cent unit prices on Metronome-pushed Stripe invoices

fabiencelier:
closes https://github.com/dust-tt/tasks/issues/9401

## Description

Metronome pushes invoice items with unit_amount_decimal (sub-cent precision), which Stripe renders verbatim on the customer-facing invoice (e.g. "€19.6451612903"). When cleaning a Metronome draft invoice, rewrite each kept line's unit price to the shortest decimal that multiplies back (by the unchanged quantity) to the exact line total, preserving both the displayed quantity and the invoice total.

Checking the unit test help understand the behaviour if needed

## Tests

added unit tests

## Risk

break some amounts displayed on the invoice.
Should not happen as we have acheck for that.

## Deploy Plan

deploy front

### DIFF
--- a/front/lib/plans/stripe.ts
+++ b/front/lib/plans/stripe.ts
@@ -1705,6 +1705,127 @@ function findCommitAppliedLineIds(
 }
 
 /**
+ * Stripe accepts at most 12 decimal places on `unit_amount_decimal`.
+ */
+const STRIPE_MAX_UNIT_AMOUNT_DECIMAL_PLACES = 12;
+
+/**
+ * Round-half-up integer division for positive bigints — the bigint equivalent
+ * of Math.round(numerator / denominator), which is also how Stripe rounds
+ * recomputed line totals. Bigint `/` truncates, so rounding is expressed as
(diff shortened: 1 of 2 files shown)


tdraier (approved):

front/lib/plans/stripe.ts
```diff
@@ -1703,6 +1703,127 @@ function findCommitAppliedLineIds(
...
+        invoiceItemId,
+        unitAmountDecimalCents,
+        amountCents,
+        quantity,
+        update,
+      },
+      "[Stripe] Normalizing sub-cent unit price on Metronome invoice line"
+    );
+
+    await stripe.invoiceItems.update(invoiceItemId, update);
```
tdraier, line 1810: bit lost here with recalculation .. just to be sure, it does not change the invoice total ? if totalsMatch is false, this will go to panic -> engrunner

fabiencelier: Yes !

PR #28497, opened 2026-07-06
fix(builder): replace editors control with "Become an editor" for non-editor admins

rfrenoy:
## Description

Ticket: https://github.com/dust-tt/tasks/issues/9418

Replace the editors dialog with "Become an editor" button when current user is an admin without edit rights on the agent or skill. Moved the common logic into a shared `BuilderEditorGateContext.tsx` helper.

## Tests

Local
## Risk

Low

## Deploy Plan

Front


philipperolet (commented):
👋 thanks for looking at that
Admins are supposed to click the "Become an editor" button
Imo it's a bit trapdoory to have an "isCurrentEditor" depend on the form control, I'd assume using dev tools you could make yourself an editor in the form and then save, so non-admins could add themselves as editor no?

Maybe a 80/20 here would be to repeat the banner from the agent inside the editors dialog

rfrenoy:
@philipperolet 

> Imo it's a bit trapdoory to have an "isCurrentEditor" depend on the form control, I'd assume using dev tools you could make yourself an editor in the form and then save, so non-admins could add themselves as editor no?

It is gated on the backend, so I don't see a risk there, a user using dev tools would get a 403 when saving.  

We can move the banner to the editors dialog, but it feels unnatural to me that on the "manage editors" dialog, I can select/unselect myself but it does nothing and there's an extra button for that. To go with your 80/20 suggestion, I would remove the current user to the list of editors in the dialog, so that it's obvious there is no other way than clicking the "become an editor" button.

philipperolet:
> @philipperolet
> 
> > Imo it's a bit trapdoory to have an "isCurrentEditor" depend on the form control, I'd assume using dev tools you could make yourself an editor in the form and then save, so non-admins could add themselves as editor no?
> 
> It is gated on the backend, so I don't see a risk there, a user using dev tools would get a 403 when saving.
> 
> We can move the banner to the editors dialog, but it feels unnatural to me that on the "manage editors" dialog, I can select/unselect myself but it does nothing and there's an extra button for that. To go with your 80/20 suggestion, I would remove the current user to the list of editors in the dialog, so that it's obvious there is no other way than clicking the "become an editor" button.

That'd be surprising to the user I think. What about disabling the editors button in the main form, with a tooltip saying to click "Become an editor"? Happy to discuss IRL if needed

philipperolet (commented):
Thanks 🙏 Looks directionally good 👍 
However, I think we don't need a react context here,  we can just reuse/extract the “Become an editor” button UI and pass the few props down explicitly
IMO we can discard BuilderEditorGateContext and have a smaller PR
(although I might have missed something? happy to discuss IRL)

rfrenoy:
@philipperolet done in c97025189ade94d4a81ba3d276440cb523ae2655  

It's 171 added LOC for 171 removed LOC, so not smaller. It's one context vs prop drilling, so a matter of taste I guess?

philipperolet:
> @philipperolet done in [c970251](https://github.com/dust-tt/dust/commit/c97025189ade94d4a81ba3d276440cb523ae2655)
> 
> It's 171 added LOC for 171 removed LOC, so not smaller. It's one context vs prop drilling, so a matter of taste I guess?

fair enough; I thought it'd be smaller without the context. Both are fine by me then, feel free to merge your preferred commit 🙏

philipperolet (approved):

rfrenoy:
No pb, thank you for the discussion 🙂

PR #28499, opened 2026-07-06
chore: remove unused feature flags

davidebbo:
## Summary

Cleans up feature flags in two ways:

### 1. Delete five flags unused across the whole codebase

They appeared only in their declaration sites, never in any `hasFeature` / `featureFlags.includes` check:

- `anthropic_reasoning_token_count`
- `force_us_api_url`
- `gemini_3_1_pro_feature`
- `hootl_subscriptions`
- `new_file_explorer`

Removed from both `front/types/shared/feature_flags.ts` and the `WhitelistableFeature` union in `sdks/js/src/types.ts`.

### 2. Remove drifted entries from the SDK union

The SDK `WhitelistableFeature` union carried 13 entries that were **never** in the front config and are unused as flags anywhere:

`agent_builder_copilot`, `agent_builder_copilot_builders`, `agent_builder_shrink_wrap`, `dust_academy`, `dust_no_spa`, `dust_spa`, `gpt_image_2_feature`, `metronome_billing_usage_page`, `sandbox_dsbx_tools`, `sandbox_workspace_admin`, `skills_as_user_messages`, `slack_bot_mcp`, `slideshow`

The SDK union and the front config now match exactly (**74 flags each**, no drift).

## Why it's safe

- `WhitelistableFeaturesSchema` in the SDK uses `FlexibleEnumSchema`, which accepts **any string at runtime** — the union is compile-time typing only, so nothing that parses `feature_flags` breaks.
- The front config is the source of truth for which flags exist; removed flags can no longer be enabled or emitted by the server, so no SDK consumer can receive them.
- The drift entries were never in the front config to begin with, so they could never have been enabled.

## Verification

- `npx tsgo --noEmit` (front) — clean
- `npx tsc --noEmit` (sdks/js) — clean
- front config ↔ SDK union: exact match, 74 flags each


aubin-tchoi (approved):
LGTM

PR #28506, opened 2026-07-06
Reduce HubSpot MCP tools: replace redundant read tools with search_crm_objects

Inesvvv:
## Description

Reduce number of tools from 43 to 38 by removing redundant `get` tools. Removing the get tools was the simplest way of reducing the number of tools because there is no workspace permission complications. This pr replaced: 
- `get_contact`
 - `get_company`
 -  `get_deal`
 - `get_object_by_email`
 - `get_latest_objects`
All replaced by pre-existing `search_crm_objects``

## Tests

Tested manually on the local dev hubspot account to ensure that behaviour hadn't changed. + compared the use of tools before and after

## Risk

Low-to-moderate. Capability is preserved (everything the removed tools did is doable via search_crm_objects), but there are behavioral trade-offs to be aware of:

- Eventual consistency on read-after-write. The removed get_* tools used basicApi.getById, which is strongly consistent. search_crm_objects goes through HubSpot's search index, which lags seconds-to-minutes behind writes. An agent that creates/updates a record and then immediately fetches it by id may briefly get no result or a stale version. Fetching pre-existing records is unaffected.
- Removing tools affects existing configs. Per-tool permission settings are keyed by tool name (remote_mcp_server_tool_metadata), so any workspace that customized get_contact/etc. will have those settings orphaned, and agents that pinned these specific tools lose them. They must switch to search_crm_objects.
- Property coverage differs slightly. The get tools returned trimmed default property sets; search_crm_objects returns all properties (more complete, but more tokens).

## Deploy Plan
Standard


matteotrab (approved):

front/lib/api/actions/servers/bm25_tool_search.test.ts
```diff
@@ -634,22 +634,6 @@ const QUERIES: LabeledQuery[] = [
   },
 
   // --- hubspot ---
-  {
-    query: "find a hubspot contact by email address",
-    expected: "hubspot.get_object_by_email",
```
aubin-tchoi, line 639: Let's keep them but replace the expected tool with search_crm_objects

front/lib/api/actions/servers/hubspot/metadata.ts
```diff
@@ -145,76 +131,6 @@ export const HUBSPOT_TOOLS_METADATA = createToolsRecord({
...
-    },
-    toolCostCategory: "advanced",
-    freeUsage: false,
-  },
-  get_company: {
-    description:
-      "Open a single HubSpot company record by its ID and read its properties. " +
-      "Returns default properties plus any additional properties specified.",
-    schema: {
-      companyId: z.string().describe("The ID of the company to retrieve."),
```
aubin-tchoi, line 181: Direct lookup by id is still quite different from relying on their search, are we sure agents are able to use the search as a drop in replacement for these lookups?

front/lib/api/actions/servers/hubspot/metadata.ts
```diff
@@ -145,76 +131,6 @@ export const HUBSPOT_TOOLS_METADATA = createToolsRecord({
...
   },
-  get_latest_objects: {
-    description: `Get latest objects from Hubspot. Supports ${SIMPLE_OBJECTS.join(", ")}. Limit is 200.`,
-    schema: {
-      objectType: z.enum(SIMPLE_OBJECTS),
-      limit: z.number().optional(),
-    },
-    stake: "never_ask",
-    displayLabels: {
-      running: "Retrieving latest HubSpot objects",
```
aubin-tchoi, line 156: We lose the richness of these labels, is there a way to add it back via dynamic display labels?

front/lib/api/actions/servers/hubspot/metadata.ts
aubin-tchoi, line 179: Let's enrich the description here to provide a bit of guidance on what it can search and how

PR #28509, opened 2026-07-06
[LLM] Add OpenAI Responses EU endpoints for GPT-5 models

pmilliotte:
## Description

Register OpenAI Responses EU stream endpoints for the GPT-5 model family, with a per-endpoint base URL and a 10% regional pricing uplift for GPT-5.4+.

## Tests

Added stream endpoint tests for each new EU endpoint; every stream endpoint is now mapped to its test setup for type-checked coverage.

## Risk

Low.

## Deploy Plan

- Deploy `main` on `front`

### DIFF
--- a/front/lib/model_constructors/providers/openai/base_url.ts
+++ b/front/lib/model_constructors/providers/openai/base_url.ts
@@ -1,0 +1,3 @@
+export const OPENAI_GLOBAL_BASE_URL = "https://api.openai.com/v1";
+
+export const OPENAI_EU_BASE_URL = "https://eu.api.openai.com/v1/";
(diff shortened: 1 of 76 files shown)


front/lib/model_constructors/providers/openai/base_url.ts
```diff
@@ -0,0 +1,3 @@
+export const OPENAI_GLOBAL_BASE_URL = "https://api.openai.com/v1";
+
+export const OPENAI_EU_BASE_URL = "https://eu.api.openai.com/v1/";
```
adrsimon, line 3: isn't it stored in an env variable ? DUST_MANAGED_OPENAI_BASE_URL ?

adrsimon (commented):
looks like a lot of code to put the same endpoint in two regions (bc conceptually we call the same endpoint, stored in the DUST_MANAGED_OPENAI_BASE_URL)

i still don't get why we can't just reuse the same code bc it's only one endpoint

adrsimon (approved):

PR #28511, opened 2026-07-06
[front] fix: inject placeholder specifications for historic tool references

aubin-tchoi:
## Description

Follow up on https://github.com/dust-tt/dust/pull/28479.
Injects tool specifications for tools that are found in the convo history (either `tool_references` from tool search or previous tool calls) but not in the current specifications.

This is a little more exhaustive than only Anthropic tool search uses, for regular tool calls this comes in addition to the `missing_action_catcher` and also allows detecting issues due to tools having been removed vs the model inventing a one from scratch.

## Tests

- Updated existing Anthropic converter tests.
- Tested locally with a direct Anthropic API smoke using `DUST_MANAGED_ANTHROPIC_API_KEY`.

## Risk

- Medium.

## Deploy Plan

- Deploy front.

### DIFF
--- a/front/lib/api/llm/clients/anthropic/utils/conversation_to_anthropic.test.ts
+++ b/front/lib/api/llm/clients/anthropic/utils/conversation_to_anthropic.test.ts
@@ -30,6 +30,20 @@ const baseSpec: AgentActionSpecification = {
   inputSchema: { type: "object", properties: {} },
 };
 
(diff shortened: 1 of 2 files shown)


flvndvd (approved):
Just to confirm that it does not break anything the fact that the tools' specifications are changing.

aubin-tchoi:
👍🏻 adding a test

PR #28517, opened 2026-07-06
Strip unreplayable Anthropic tool search blocks at request build

flvndvd:
## Description

With Anthropic tool search enabled, assistant turns carry server side search blocks that we persist opaquely and replay verbatim on later requests. The API validates those blocks on every request and rejects the whole request with a 400 when the replayed shape is not one it can honor. In production this bricked conversations: once the bad shape was persisted, every subsequent step failed deterministically with the same error, and retries could not help.

Live probing and a [chat](https://dust4ai.slack.com/archives/C0735P3JH9T/p1783006234708799?thread_ts=1782917065.661639&cid=C0735P3JH9T) with Anthropic established the exact contract, which the public documentation does not cover. A search the API never ran, which happens when the model issues searches and a client tool call in the same turn and the turn ends on the client call, or when a turn is interrupted, is accepted in exactly one shape: it must sit on the final assistant message and the continuation must consist exclusively of tool results, in which case the API resumes the search on the follow up request. Every other replay is rejected: a text block appended to the continuation, which is what enabling a skill does with its instructions and was the trigger observed in production, a pending search sitting in the middle of history after the conversation moved on, or a stopped or errored turn followed by a new user message. Separately, requests that carry no tool search tool at all, such as title generation and other auxiliary calls that replay conversation history without tools, reject any tool search block because the tool references inside the results cannot be expanded.

The fix is a single pure sanitation step where the provider request is assembled, applied in both the new router and the legacy client and therefore covering streaming and batch. It keeps everything the API accepts and strips only what it would reject. Completed search pairs are never touched. A pending search in the resumable shape is kept so the API can still run it, which is strictly better than stripping since the search is not lost. Every other pending search is removed, as are all tool search blocks on requests that do not carry the search tool. When stripping empties a message entirely the message is dropped and its same role neighbors are re merged, since the API also rejects consecutive same role messages. When nothing needs stripping the input is returned untouched, so the common path allocates nothing and stays byte identical for prompt caching.

> [!NOTE]
> The main risk assessed was the interaction with signed thinking blocks, since the same assistant message can interleave reasoning with searches and the API rejects modified thinking. Both the documentation and live probes confirm the signature covers the thinking blocks themselves and not their siblings: removing a pending search from the middle of a message that carries signed thinking is accepted, including with an unresumable continuation and with tool choice steering on the follow up request.

### DIFF
--- a/front/lib/model_constructors/sdk/anthropic_ai/tool_search_dangling_server_tool_use.test.ts
+++ b/front/lib/model_constructors/sdk/anthropic_ai/tool_search_dangling_server_tool_use.test.ts
@@ -1,0 +1,365 @@
+import type {
+  MessageParam,
+  RawMessageStartEvent,
+  RawMessageStreamEvent,
+} from "@anthropic-ai/sdk/resources/messages/messages";
+import {
+  parseAnthropicToolSearchBlock,
+  stripUnreplayableToolSearchBlocks,
+} from "@app/lib/api/llm/clients/anthropic/utils/tool_search_passthrough";
+import type { MessageBlockConverters } from "@app/lib/model_constructors/sdk/anthropic_ai/converters/input/utils";
+import {
+  assistantProviderPassthroughMessageToBlocks,
+  assistantReasoningMessageToThinkingBlocks,
+  assistantTextMessageToTextBlock,
+  assistantToolCallRequestToToolUseBlock,
+  conversationToMessages,
+  imageUrlToImageBlock,
+  systemMessageToTextBlock,
+  userTextMessageToTextBlock,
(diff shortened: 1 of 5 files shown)


front/lib/model_constructors/sdk/anthropic_ai/tool_search_dangling_server_tool_use.test.ts
```diff
@@ -0,0 +1,362 @@
+import type {
```
flvndvd, line 1: Script used to repro, committing for posterity.

aubin-tchoi (approved):
LGTM

PR #28518, opened 2026-07-06
Pricing: Add a new 'top-ups history' tab in usage page

fabiencelier:
## Description

closes https://github.com/dust-tt/tasks/issues/8837

There is currently no way to see the top-ups contributing to the pool.
This PRs pull the information for metronome and displays it as a new tab in the "Usage" page.

Screenshot 2026-07-06 at 16 10 05

There is a question of where to place this between the "Usage" page and the "Billing" page. IMHO it is more adapted to be in the Usage page because it is the only place where we talk about the credit pool, and it is the place where you can actually top up the pool, so where you will want to see the history.

## Tests

Tested locally + unit tests

## Risk

low, easy to revert

## Deploy Plan

deploy front


thervier (approved):
LGTM

PR #28520, opened 2026-07-06
Pricing: Add contract id to pool credits coupon 

fabiencelier:
## Description

closes https://github.com/dust-tt/tasks/issues/9428

To be consistent with the other kind of coupons

Also rename for clarity:
 -  createCouponCredit -> createSeatCouponCredit
 - createCreditsCouponCredit -> createPoolTopupCouponCredit
 - redeemCreditsCoupon -> redeemPoolTopupCoupon
 - RedeemCreditsCouponError -> RedeemPoolTopupCouponError

## Tests

tested locally: coupon is still counted in pool and still displayed in top up history.

## Risk

low

## Deploy Plan

deploy front

### DIFF
--- a/front/lib/metronome/coupons.ts
+++ b/front/lib/metronome/coupons.ts
@@ -234,7 +234,7 @@ export type RedeemPoolTopupCouponError =
 
 // Redeem a "credit_pool_top_up" coupon as a standalone, synchronous action (the "Use
 // coupon" Top-Up tab): no payment is involved, the coupon simply grants free
-// AWU credits to the workspace pool. Mirrors `redeemCoupon` (the subscription
+// AWU credits to the workspace pool. Mirrors `redeemSeatCoupon` (the subscription
 // path) end-to-end: validate → check for an existing redemption → create the
 // pending redemption (incrementing the count) → grant the credit → mark active.
 // On a Metronome failure the pending redemption is rolled back so the count is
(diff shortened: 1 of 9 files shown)


front/lib/metronome/coupons.ts
thervier: should be renamed as well, no ?

front/lib/metronome/coupons.ts
```diff
@@ -396,7 +422,7 @@ export async function redeemCoupon(
   }
   const redemption = pendingResult.value;
 
-  const creditResult = await createCouponCredit({
+  const creditResult = await createSeatCouponCredit({
```
thervier, line 426: we should use the contractId here as well

fabiencelier: done

front/lib/metronome/coupons.ts
fabiencelier: did a batch of additional renaming, trying to always say "seat coupon" or "pool topup coupon"

front/lib/metronome/coupons.ts
```diff
@@ -396,8 +423,9 @@ export async function redeemCoupon(
...
     metronomeCustomerId,
+    metronomeContractId: contract ? contract.id : undefined,
```
thervier: nit: `metronomeContractId: contract?.id` instead ?

thervier (approved):
thx a lot!
LGTM

PR #28521, opened 2026-07-06
[front] enh: show tool-search in Poke conversation page

aubin-tchoi:
## Description

This PR makes the Poke conversation debug view show provider passthrough blocks persisted in agent message contents. The row uses a light best-effort label for tool-search calls/results when the obvious fields are present, but always keeps the raw provider block expandable so the debug view remains useful even if the provider shape changes.

It also keeps all Poke trace and Langfuse links visible by rendering them in a wrapping trace row instead of the truncated metadata item.

The active-message status strip now stays neutral and uses a spinner for generating agent messages instead of showing a warning-style callout.

## Tests

- Tested locally.

## Risk

- Low. Poke-only debug UI.

## Deploy Plan

- Deploy front.

### DIFF
--- a/front/components/poke/pages/ConversationPage.tsx
+++ b/front/components/poke/pages/ConversationPage.tsx
@@ -91,6 +91,115 @@ function formatDurationMs(durationMs: number) {
     : `${durationMs}ms`;
 }
 
+interface ProviderPassthroughEntry {
+  block: unknown;
+  key: string;
+  provider: string;
+  step: number;
+}
+
+function isObjectRecord(value: unknown): value is Record<string, unknown> {
+  return typeof value === "object" && value !== null && !Array.isArray(value);
+}
+
+function getProviderPassthroughEntries(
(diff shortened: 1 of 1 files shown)


front/components/poke/pages/ConversationPage.tsx
```diff
@@ -908,12 +1120,12 @@ export function ConversationPage() {
             </div>
           )}
           {(pendingUserCount > 0 || createdAgentCount > 0) && (
-            <div className="flex flex-wrap items-center gap-2 rounded-md border border-border-warning bg-background px-3 py-2 text-sm text-warning">
-              <span className="text-sm font-medium text-warning">
+            <div className="flex flex-wrap items-center gap-2 rounded-md border border-separator bg-muted-background px-3 py-2 text-sm text-muted-foreground">
```
aubin-tchoi, line 1123: unrelated change: just changing colors here

adrsimon (approved):

PR #28522, opened 2026-07-06
[front] Fix crash in getStripePricingData when a currency option is missing

davidebbo:
## Problem

`GET /api/w/:wId/credits/purchase` was throwing a 500 in production:

```
TypeError: Cannot read properties of undefined (reading 'unit_amount')
    at getStripePricingData (/app/front/lib/plans/stripe.ts:103:22)
    at getCreditPurchaseInfo (/app/front/lib/api/credits/purchase.ts:157:25)
```

`getStripePricingData` iterates over `SUPPORTED_CURRENCIES` (`usd`, `eur`) and read
`price.currency_options[currency].unit_amount` without first checking that the option
existed. When a Stripe price has an option configured for `usd` but not `eur`,
`price.currency_options["eur"]` is `undefined`, so dereferencing it throws.

TypeScript didn't catch this because `noUncheckedIndexedAccess` is not enabled.

## Fix

The loop's existing `else` branch already logs `"[Stripe] Currency option not configured"`
and skips such currencies — the intent was always to tolerate a missing option, it just
dereferenced before checking. The fix treats a missing option as `unitAmount = 0`, which
falls into that existing log-and-skip path. The `assert(currencyOptions.usd.unitAmount > 0)`
still guards the critical case where USD itself is misconfigured.

## Tests

Added a `getStripePricingData` test block (with a `prices.retrieve` mock). The
"missing currency option" test reproduced the exact production error before the fix and
passes after. All 34 tests in the file pass; typecheck and biome are clean.

### DIFF
--- a/front/lib/plans/stripe.test.ts
+++ b/front/lib/plans/stripe.test.ts
@@ -3,6 +3,7 @@ import {
   finalizeInvoice,
   getCreditAmountFromInvoice,
   getCreditPurchaseCouponId,
+  getStripePricingData,
   getSubscriptionInvoices,
   isCreditPurchaseInvoice,
   isEnterpriseSubscription,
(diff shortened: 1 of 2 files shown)


id13 (approved):
LGTM

PR #28528, opened 2026-07-06
sparkle: use lighter stone track color in dark mode for SliderToggle …

pinotalexandre:
## Description
Changed the background color of the slider toggle in dark mode

## Risk
low


ykmsd (approved):
There is a conflict so I cannot test it but I approved, I trust you that you tested it 🫡
 
Also ideally we want to avoid using `dark:` prefix and use semantic token something like slide-toggle-bg-idle)

pinotalexandre:
Updated @ykmsd

PR #28536, opened 2026-07-06
[front] enh: tweak discover knowledge instructions

aubin-tchoi:
## Description

This PR changes the instructions of  the `discover_knowledge` skill to explain how to use its company data and data warehouse tools directly in the skill prompt rather than relying on the assumption that the tool descriptions are also exposed (not always true since they can now be deferred).

Intentionally keeps the tool descriptions as is, they will add a bit of redundancy but are necessary to prevent breaking direct consumers of the tool that do not go through the skill (most custom agents, this account for approximately half the semantic searches).

Also updates the query sample for bm25 search tests; it only makes sense that the query do not contain specific knowledge that you would access only after accessing the tool descriptions.

No runtime behavior changes.

## Tests

- Tested locally.

## Risk

- Medium.

## Deploy Plan

- Deploy front.


front/lib/resources/skill/code_defined/discover_knowledge.ts
```diff
@@ -1,46 +1,98 @@
...
+  "Company data filesystem tools:\n" +
+  `- \`${filesystemToolName("semantic_search")}\`: best for topic, intent, ` +
+  "or meaning-based discovery. " +
+  "Use it for broad questions, policies, project context, scattered facts, or " +
+  "anything where the exact page title is unknown.\n" +
+  `- \`${filesystemToolName("list")}\`: best for understanding structure ` +
+  "and coverage. Use it when a folder or data source matters as a collection, " +
+  "when search feels too broad, or when nearby sibling documents may be " +
+  "relevant.\n" +
+  `- \`${filesystemToolName("find")}\`: best for title-based discovery from ` +
```
flvndvd, line 40: `ni` should we explain the "exact match" constraint?

flvndvd (approved):

front/lib/resources/skill/code_defined/discover_knowledge.ts
aubin-tchoi: I'd keep anything related to what each parameter does in the tool description as it does not help with understanding whether it's useful (here we cover it lightly by mentioning that the match can be partial)

PR #28538, opened 2026-07-06
Fix tool search pair detection across assistant messages

flvndvd:
## Description

The replay sanitizer from #28517 decided whether an Anthropic tool search was still pending by looking for its result inside the same assistant message. That assumption breaks in the flow the sanitizer was built to preserve: when a pending search is kept and the API resumes it, the result opens the next assistant turn, so a completed pair legitimately spans two assistant messages with the tool result continuation in between. On the following replay the call half was misclassified as pending and stripped while the result half was kept, and the API rejected the request with a 400 for a search result lacking its corresponding call. This is the production error seen after deploy.

Pairing is now computed across the whole conversation: a call counts as pending only when no result matches it anywhere, and a result is kept only when its call was kept, which also removes genuinely orphaned results defensively. The resumable-pending rule, the byte-identical fast path for prompt caching, and the strip warning log are unchanged. The sanitizer is stateless at request build time, so affected conversations kept their full history and replay correctly once this is deployed. A regression test reproduces the exact production topology and asserts the messages pass through untouched, with a second test covering the orphaned result case.

### DIFF
--- a/front/lib/api/llm/clients/anthropic/utils/tool_search_passthrough.test.ts
+++ b/front/lib/api/llm/clients/anthropic/utils/tool_search_passthrough.test.ts
@@ -202,6 +202,34 @@ describe("stripUnreplayableToolSearchBlocks", () => {
     ]);
   });
 
+  it("keeps a pair split across assistant messages after a resumed search completed on the next turn", () => {
+    const messages = [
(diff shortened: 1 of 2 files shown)


aubin-tchoi (approved):
LGTM

PR #28543, opened 2026-07-06
fix(transcripts): read non-Google-Docs transcript files via alt=media

rfrenoy:
## Description

Google Meet/Gemini transcripts now land in Drive as text files rather than native Google Docs. drive.files.export only works on Docs Editors files, so it 403s on these, which surfaced as "[processTranscriptActivity] File content is not accessible. Stopping."

Fetch the mimeType and branch: export Google Docs, download text/* files with files.get({ alt: "media" }), and throw with the mime type otherwise. The failure log now includes the mimeType so unexpected types are diagnosable rather than silent.
## Risk

Low, the feature does not work without it anyway 😬 
## Deploy Plan

Front

### DIFF
--- a/front/temporal/labs/transcripts/utils/google.ts
+++ b/front/temporal/labs/transcripts/utils/google.ts
@@ -111,6 +111,46 @@ export async function retrieveGoogleTranscripts(
   return new Ok(fileIdsToProcess);
 }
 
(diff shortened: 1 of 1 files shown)


thervier (commented):
LGTM

PR #28553, opened 2026-07-07
test(transcripts): add drive.readonly to labs_transcripts scope

rfrenoy:
## Description

Diagnostic for the export 404 on Meet transcript Docs: drive.meet.readonly allows metadata reads but not files.export (Google returns 404). Adding drive.readonly to confirm the scope is the cause. Requires re-consent to take effect.
## Risk

Low
## Deploy Plan

Front + user must reauth


thervier (approved):
LGTM

PR #28556, opened 2026-07-07
[front] enh: make img-generation tool credit-based

id13:
## Description

Currently, we do not bill LLM usage of image tool at all.
With the advent of credit-based pricing, it makes sense (and is desirable) to bill using the same logic, which means an all or nothing, we only bill if the img tool call was successful AND the agent-loop succeeded.

We do this by introducing runIds to tool call answer

### DIFF
--- a/front/lib/actions/types/index.ts
+++ b/front/lib/actions/types/index.ts
@@ -122,6 +122,10 @@ export type StepContext = {
   // Dust run IDs of LLM runs made by the tool itself (e.g. image generation).
   // Flowed back to the agent loop so the runs are billed like any other LLM
   // usage of the agent message.
+  // /!\ Append-only and owned by the DB: written exclusively through
+  // AgentMCPActionResource.appendToolRunIds (atomic JSONB append), and
+  // preserved by updateStepContext so stale in-memory spreads cannot clobber
+  // runs recorded mid-execution.
   toolRunIds?: string[];
 };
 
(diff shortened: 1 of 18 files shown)


front/lib/api/actions/servers/image_generation/helpers.ts
```diff
@@ -181,25 +194,69 @@ export async function checkImageGenerationRateLimit(
...
+    if (action) {
+      await action.updateStepContext({
+        ...action.stepContext,
+        runIds: [...(action.stepContext.runIds ?? []), dustRunId],
+      });
+    }
+  }
+
+  return dustRunId;
 }
```
id13, line 252: @thervier or @pmilliotte would like your eyes on this 😇

front/lib/actions/types/index.ts
```diff
@@ -115,6 +115,10 @@ export type StepContext = {
   resumeState: Record<string, unknown> | null;
   retrievalTopK: number;
   websearchResultCount: number;
+  // Dust run IDs of LLM runs made by the tool itself (e.g. image generation).
+  // Flowed back to the agent loop so the runs are billed like any other LLM
+  // usage of the agent message.
+  runIds?: string[];
```
id13: Will revert this, not a big fan

front/lib/api/actions/servers/image_generation/helpers.ts
thervier: invoking @Nils-Fedrigo here as he worked recently on the usage in the agent loop for the pricing initiative :) 

But I have tested the PR and checked with message cost, and it works as expected: the run is added to the list of runs and taken into account during credit computation for the message

front/lib/actions/types/index.ts
id13: Meditating, alternatives are worse, maybe worth keeping ... looking at dustRunIds

front/lib/api/actions/servers/image_generation/helpers.ts
pmilliotte: LGTM

PR #28557, opened 2026-07-07
Truncate tool error message in 'Tool execution error' log

davidebbo:
## Description

The tool's error message can be very large, bloating our logs. This caps it at 200 chars in the `"Tool execution error"` log context in `wrappers.ts`, while preserving the `tracked` and `code` fields that were previously logged via pino's `err` serializer.

### DIFF
--- a/front/lib/actions/mcp_internal_actions/wrappers.ts
+++ b/front/lib/actions/mcp_internal_actions/wrappers.ts
@@ -193,10 +193,7 @@ function withToolLogging<T>(
           ),
           code: result.error.code,
           cause: result.error.cause
-            ? truncate(
-                errorToString(result.error.cause),
-                MAX_LOGGED_ERROR_MESSAGE_LENGTH
-              )
+            ? errorToString(result.error.cause)
             : undefined,
         },
       };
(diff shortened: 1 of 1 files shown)


front/lib/actions/mcp_internal_actions/wrappers.ts
```diff
@@ -183,7 +186,14 @@ function withToolLogging<T>(
       const logContext = {
         ...loggerArgs,
         duration: elapsed,
-        error: result.error,
+        error: {
+          message: truncate(
+            result.error.message,
+            MAX_LOGGED_ERROR_MESSAGE_LENGTH
+          ),
+          tracked: result.error.tracked,
```
aubin-tchoi: nit: we already know whether tracked is true based on the logging level (error/warn)

front/lib/actions/mcp_internal_actions/wrappers.ts
```diff
@@ -183,7 +186,14 @@ function withToolLogging<T>(
       const logContext = {
         ...loggerArgs,
         duration: elapsed,
-        error: result.error,
+        error: {
```
aubin-tchoi, line 189: IIRC we also have a `cause` we can log

aubin-tchoi (approved):
LGTM

front/lib/actions/mcp_internal_actions/wrappers.ts
davidebbo: Ok, I removed 'tracked'

front/lib/actions/mcp_internal_actions/wrappers.ts
davidebbo: Ok, I added the cause.

zmarouf:
LGTM

PR #28560, opened 2026-07-07
feat: add governance page

matteotrab:
## Description

This PR introduces a first version of the Governance page that allows business admins and admins to manage governance settings.

Known missing features to be done in followup PRs:
- use real hooks
- mobile responsiveness
- groups only sections
- copy
- loading state

Note that the prototype is using an old version of the design system.

https://github.com/user-attachments/assets/af638aa9-7cd6-4e61-8050-7096350201cc

## Tests

Manually.

## Risks
Low. The page is hidden behind the feature flag.

## Deploy Plan

Standard deployment.

### DIFF
--- a/front/components/navigation/config.ts
+++ b/front/components/navigation/config.ts
@@ -32,6 +32,7 @@ import {
   Shapes,
   Stars02,
   Terminal,
+  Toggle01Left,
   Users01,
   Zap,
 } from "@dust-tt/sparkle";
@@ -91,6 +92,7 @@ export type SubNavigationAdminId =
   | "subscription"
   | "billing"
   | "workspace"
+  | "governance"
   | "workspace_branding"
   | "model_providers"
   | "members"
@@ -108,6 +110,7 @@ export const ADMIN_ROUTE_PATTERNS: Record<SubNavigationAdminId, string[]> = {
   members: ["/w/[wId]/members"],
   identity_and_provisioning: ["/w/[wId]/identity-and-provisioning"],
   workspace: ["/w/[wId]/workspace"],
+  governance: ["/w/[wId]/governance"],
   workspace_branding: ["/w/[wId]/brand"],
   model_providers: ["/w/[wId]/model-providers"],
   analytics: ["/w/[wId]/analytics"],
@@ -210,7 +213,7 @@ export const getTopNavigationTabs = (
     ref: spaceMenuButtonRef,
   });
 
-  if (isAdmin(owner) || isOnlyBusinessAdmin(owner)) {
+  if (isBusinessAdmin(owner)) {
     nav.push({
       id: "settings",
       label: "Admin",
@@ -221,6 +224,7 @@ export const getTopNavigationTabs = (
           "/w/[wId]/members",
           "/w/[wId]/identity-and-provisioning",
           "/w/[wId]/workspace",
+          "/w/[wId]/governance",
           "/w/[wId]/branding",
           "/w/[wId]/model-providers",
           "/w/[wId]/subscription",
@@ -297,6 +301,18 @@ export const subNavigationAdmin = ({
         current: isCurrent("workspace"),
         disabled: !hasAdminRole,
       },
+      ...(featureFlags.includes("admin_governance")
+        ? [
+            {
+              id: "governance" as const,
+              label: "Governance",
+              icon: Toggle01Left,
+              href: `/w/${owner.sId}/governance`,
+              current: isCurrent("governance"),
+              disabled: !hasBusinessAdminRole,
+            },
+          ]
+        : []),
       ...(featureFlags.includes("whitelabel_frames")
         ? [
             {
--- a/front/components/pages/workspace/governance/GovernancePage.tsx
+++ b/front/components/pages/workspace/governance/GovernancePage.tsx
@@ -1,0 +1,132 @@
+import { GovernanceSettingSection } from "@app/components/pages/workspace/governance/GovernanceSettingSection";
+import { useWorkspace } from "@app/lib/auth/AuthContext";
+import { useGovernancePermissions } from "@app/lib/swr/governance";
+import { useGroups } from "@app/lib/swr/groups";
+import type {
+  GovernancePermission,
+  GroupPermissionResourceType,
+  PermissionType,
+} from "@app/types/group_permissions";
+import type { LightWorkspaceType } from "@app/types/user";
+import {
+  ActionFrame,
+  Page,
+  PuzzlePiece01,
+  Robot,
+  Toggle01Left,
+} from "@dust-tt/sparkle";
+import groupBy from "lodash/groupBy";
+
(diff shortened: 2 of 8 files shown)


matteotrab:

- #28560 ◀

front/lib/swr/governance.ts
```diff
@@ -0,0 +1,40 @@
+import type { GovernancePermission } from "@app/types/group_permissions";
+import type { LightWorkspaceType } from "@app/types/user";
+
+export function useGovernancePermissions(owner: LightWorkspaceType): {
+  governancePermissions: GovernancePermission[];
+  isLoading: boolean;
+} {
+  const governancePermissions: GovernancePermission[] = [
```
ykmsd: let's not inline it and take it out from the function 🥶

ykmsd: even if it's fake 😄

front/components/pages/workspace/governance/GroupSelector.tsx
```diff
@@ -0,0 +1,93 @@
...
+          {filteredGroups.length === 0 && (
+            <DropdownMenuItem
+              label={groupSearch ? "No groups found" : "All groups added"}
+              disabled
+            />
+          )}
+        </DropdownMenuContent>
+      </DropdownMenu>
+      {selectedGroups.map((group) => (
+        <span key={group.sId}>
```
ykmsd: why do we need span here? is it for key? Can't we use React.fragment?

front/components/pages/workspace/governance/GovernancePage.tsx
```diff
@@ -0,0 +1,132 @@
...
+          groups={groups}
+          onPermissionChange={onPermissionChange}
+        />
+        <GovernanceSettingSection
+          label="Frames"
+          icon={ActionFrame}
+          governanceSettings={frameSettings}
+          groups={groups}
+          onPermissionChange={onPermissionChange}
+        />
```
ykmsd: nit but we are repeating many codes here, wouldn't it be cleaner to have const for setting section and map over?

front/components/pages/workspace/governance/GovernancePage.tsx
```diff
@@ -0,0 +1,132 @@
...
+  );
+
+  const agentSettings = governanceSettingsMap.agent ?? [];
+  const skillSettings = governanceSettingsMap.skill ?? [];
+  const frameSettings = governanceSettingsMap.frame ?? [];
+
+  if (isLoading) {
+    return (
+      <Page>
+        <Page.Header title="Governance" description="Loading..." />
```
ykmsd, line 67: description loading sounds strange but I cannot see the UI, are we showing it in UI?

front/components/pages/workspace/governance/GovernancePage.tsx
```diff
@@ -0,0 +1,132 @@
...
+
+function useUpdateGovernancePermission(owner: LightWorkspaceType) {
+  return (input: GovernancePermission) => {
+    return true;
+  };
+}
+
+export const GovernancePage = () => {
+  const owner = useWorkspace();
+  const { groups } = useGroups({ owner, kinds: ["provisioned"] });
```
ykmsd: note for anybody who wonder: `["provisioned"]` is not stable so we are re-creating this on every re-render but we are not re-fetching on every re-render because we flatten it to string inside the useGroups 
```
if (kinds && kinds.length > 0) {
      kinds.forEach((k) => params.append("kind", k));
    } 
```

front/components/pages/workspace/governance/GovernancePage.tsx
```diff
@@ -0,0 +1,132 @@
...
+    description: "Publish Skills workspace-wide for all members to use",
+  },
+  "invite:frame": {
+    label: "Members + email invites",
+    description:
+      "Frames can be shared with workspace members or via email invite",
+  },
+};
+
+function toGovernanceSettings(
```
ykmsd: discussed IRL, I think it's simpler to separate label + description and actually configurations

front/components/pages/workspace/governance/GroupSelector.tsx
matteotrab: No need at all for a span actually, we can give the key directly to the chip

front/components/pages/workspace/governance/GovernancePage.tsx
matteotrab: For the moment it never appears as we are using fake data. I will make a proper loading page later

front/types/group_permissions.ts
```diff
@@ -51,3 +51,31 @@ export function isGroupPermissionResourceType(
...
+  scope: string
+): scope is PermissionConfigurationScope => {
+  return PERMISSION_CONFIGURATION_SCOPES.some(
+    (validScope) => validScope === scope
+  );
+};
+
+export type GovernancePermissionConfiguration = {
+  scope: PermissionConfigurationScope;
+  groupIds?: string[];
```
philipperolet: nit: some typing-fu could have us do { scope: groups, groupIds } OR { scope: PermConfig excluding groups} I think?

front/components/navigation/config.ts
```diff
@@ -210,7 +213,7 @@ export const getTopNavigationTabs = (
     ref: spaceMenuButtonRef,
   });
 
-  if (isAdmin(owner) || isOnlyBusinessAdmin(owner)) {
+  if (isBusinessAdmin(owner)) {
```
philipperolet, line 216: nit: the page is not actually hidden behind admin_governance: anyone with the role can open /w/<wId>/governance
we might feature flag here too, even if low risk? not sure it's required but flagging

front/components/pages/workspace/governance/GovernanceSettingRow.tsx
```diff
@@ -0,0 +1,80 @@
...
+  groups,
+  onChange,
+}: GovernanceSettingRowProps) => {
+  const [configuration, setConfiguration] =
+    useState<GovernancePermissionConfiguration>(
+      governanceSetting.configuration
+    );
+
+  const selectedGroupIds = new Set(configuration.groupIds ?? []);
+  const selectedGroups = groups.filter((g) => selectedGroupIds.has(g.sId));
```
philipperolet: we might want to make sure (assert?) that all groups in configuration.groupIds are indeed in selectedGroups?
if for some reason one is not available, then current code silently drops it forever IIUC?

philipperolet (approved):
Thanks 🙏 left a few coms

front/components/navigation/config.ts
matteotrab: Good point, I'll do it

front/components/pages/workspace/governance/GovernanceSettingRow.tsx
matteotrab: Added a check to display an error in that case

PR #28562, opened 2026-07-07
fix(labs-transcripts): skip file instead of deleting schedule on config drift

rfrenoy:
## Description

Ticket: https://github.com/dust-tt/tasks/issues/9004
Follow-up to https://github.com/dust-tt/dust/pull/28480.

processTranscriptActivity deleted the whole retrieval schedule when a single file could not be stored or processed (missing data source view, no write permission, missing data source, missing/archived agent). A transient permission/fetch blip or a since-removed resource would therefore permanently disable processing for the user with no auto-recovery. Replace the six per-file `stopRetrieveTranscriptsWorkflow() + return` blocks with a plain `return` (skip this file, log it). The schedule survives, so the file is retried on the next tick and recovers on its own once the underlying config issue is resolved. Logs updated from "Stopping." to "Skipping file." to match the new behavior. The remaining teardown calls (workspace/user/config not found) are left as-is: those reflect a genuinely orphaned config and are out of scope for this fix.

### DIFF
--- a/front/temporal/labs/transcripts/activities.ts
+++ b/front/temporal/labs/transcripts/activities.ts
@@ -383,9 +383,8 @@ export async function processTranscriptActivity(
     if (!dataSourceViewId) {
       localLogger.error(
         {},
-        "[processTranscriptActivity] No datasource view id found. Stopping."
+        "[processTranscriptActivity] No datasource view id found. Skipping file."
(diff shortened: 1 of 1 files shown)


Fraggle (approved):
LGTM

PR #28563, opened 2026-07-07
Pricing: Add a Coupons tab in billing to list redeemed coupons

fabiencelier:
## Description

Closes https://github.com/dust-tt/tasks/issues/9433

Add a tab to show the coupons redeemed and the remaining money on it, plus the invoiced period corresponding to the spending.
The tab is only shown to user with redeemed coupons.

Changes:
 - new API to retrieve the coupons: seat coupons fetches the data from metronome, and topup coupons are straightforward.
 - new endpoint + SWR + tests
 - new UI component

image

image

Final design may be subject to change when looping back with design team.

## Tests

 - tested locally
 - added some unit test for the endpoint

## Risk

low

## Deploy Plan

deploy front

### DIFF
--- a/front-api/routes/w/[wId]/coupon/index.ts
+++ b/front-api/routes/w/[wId]/coupon/index.ts
@@ -1,6 +1,7 @@
 import { workspaceApp } from "@front-api/middlewares/ctx";
 
 import redeem from "./redeem";
+import redemptions from "./redemptions";
 import validate from "./validate";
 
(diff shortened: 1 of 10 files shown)


thervier (approved):
LGTM
nit: there could a misunderstanding when reading the code between the already existing objects for Coupons and Redemptions (resources) and the new objects that corresponds to the "usage" view of a coupon based on metronome.

PR #28564, opened 2026-07-07
[front] fix: move pods skill's instructions in system prompt within pods

aubin-tchoi:
## Description

This PR fixes how pod's instructions are exposed to the model.

- Move pods instruction to the system prompt when in a pod (this is the only actual behavior change).
- Keep Pods available as a normal global skill outside Pod conversations (i.e. model sees it and can enable).
- Other unrelated janitorial work: keep auto-equipped skills like Computer and Pod Functions as global skills that the model can enable on demand (this was always the case and was the intent, it was just typed in a confusing way).

This PR is mostly about fixing the behavior for pods. A follow up refactor-only PR will move the files around and strengthen the contract. 

## Tests

- Updated existing tests.

## Risk

- Low.

## Deploy Plan

- Deploy front.

### DIFF
--- a/front/lib/resources/skill/code_defined/pod_functions.ts
+++ b/front/lib/resources/skill/code_defined/pod_functions.ts
@@ -1,7 +1,7 @@
 import { SANDBOX_FUNCTIONS_SERVER_NAME } from "@app/lib/api/actions/servers/sandbox_functions/metadata";
 import type { Authenticator } from "@app/lib/auth";
 import { getFeatureFlags } from "@app/lib/auth";
-import type { SystemSkillDefinition } from "@app/lib/resources/skill/code_defined/shared";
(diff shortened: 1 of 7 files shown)


front/lib/resources/skill/code_defined/pod_functions.ts
```diff
@@ -81,4 +81,4 @@ function's contract before relying on it. See each tool's own description for it
     !isPodConversation(agentLoopData.conversation),
   // Equipped in Pod conversations but not auto-enabled.
   isAutoEquippedForAgentLoop: (): boolean => true,
-} as const satisfies SystemSkillDefinition;
+} as const satisfies GlobalSkillDefinition;
```
aubin-tchoi, line 84: will strengthen the type and reorganize a bit in a follow up PR (will probably move system and global skills in 2 separate folders)

flvndvd (approved):

PR #28566, opened 2026-07-07
[front] fix: deduplicate Go Deep company data tools

aubin-tchoi:
## Description

This PR fixes Go Deep exposing a separate `data_sources_file_system` set of tools from Discover Knowledge.

Go Deep now uses the same `company_data` server name override as Discover Knowledge, so when both skills are enabled the shared company data filesystem tools deduplicate through the existing MCP server configuration dedupe path.

## Tests

- Added a test.

## Risk

- Low.

## Deploy Plan

- Deploy front.

### DIFF
--- a/front/lib/resources/skill/code_defined/go_deep.ts
+++ b/front/lib/resources/skill/code_defined/go_deep.ts
@@ -2,7 +2,10 @@ import { getDeepDiveInstructions } from "@app/lib/api/assistant/global_agents/co
 import { isDeepDiveDisabledByAdmin } from "@app/lib/api/assistant/global_agents/configurations/dust/utils";
 import type { Authenticator } from "@app/lib/auth";
 import { getFeatureFlags } from "@app/lib/auth";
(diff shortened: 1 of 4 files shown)


front/lib/resources/skill/code_defined/go_deep.ts
```diff
@@ -42,7 +42,7 @@ export const goDeepSkill = {
       childAgentId: GLOBAL_AGENTS_SID.DUST_PLANNING,
       serverNameOverride: "planning_agent",
     },
-    { name: "data_sources_file_system" },
+    { name: "data_sources_file_system", serverNameOverride: "company_data" },
```
flvndvd: Can we share a constant name here? It seems pretty easy to break.

flvndvd (approved):

front/lib/resources/skill/code_defined/go_deep.ts
aubin-tchoi: sure but hopefully no one really touches this lightly

PR #28568, opened 2026-07-07
Stop eagerly loading replayed tools behind tool search

flvndvd:
## Description

Until now, any tool that had been called or discovered earlier in a conversation and missing in the last turn was promoted to eager loading on every subsequent model call, pulling it out of the deferred tool-search catalog.

### DIFF
--- a/front/temporal/agent_loop/lib/run_model.ts
+++ b/front/temporal/agent_loop/lib/run_model.ts
@@ -200,7 +200,34 @@ function buildReplayOnlyToolSpecification(
       required: [],
       additionalProperties: true,
     },
-    eager: true,
+  };
+}
+
+// Replayed tools keep their intrinsic `eager` flag: providers resolve deferred
+// tools from the replayed history, and promoting them would invalidate the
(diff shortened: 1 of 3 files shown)


front/temporal/agent_loop/lib/run_model.ts
```diff
@@ -200,7 +200,35 @@ function buildReplayOnlyToolSpecification(
...
+// must have a definition.
+export function buildSpecificationsWithReplayPlaceholders(
+  baseSpecifications: AgentActionSpecification[],
+  modelConversation: ModelConversationTypeMultiActions
+): {
+  specifications: AgentActionSpecification[];
+  missingReplayedToolNames: string[];
+} {
+  const currentToolNames = new Set(baseSpecifications.map((spec) => spec.name));
+  const missingReplayedToolNames = getReplayedToolNames(
```
flvndvd, line 220: Dropped the set since `getReplayedToolNames` already uses a `Set` internally.

aubin-tchoi (approved):
LGTM

PR #28572, opened 2026-07-07
[front] chore: unify skill auto enabled/equipped callback

aubin-tchoi:
## Description

This PR fixes and refactors how equipped/enabled is defined for skills, notably rationalizing the `isAutoXxx` callbacks, that proved to add several edge cases.

Also adds quite a few tests: the goal is really to make sure we do not alter what ends up in each bucket (system, enabled, equipped). We used to have a very simple rule (added in agent builder -> equipped, enabled by agent -> enabled) but have grown to have more needs in terms of customization (some skills are always equipped, some conditionally on the current conversation) so the few invariants are now encoded in tests rather than in implementation, which can outgrow them.

`listForAgentLoop` now builds the three buckets as follows:
- system skills come from agent-added system skills plus auto-enabled code-defined skills;
- enabled skills come from `listEnabledByConversation`, excluding skills already owned by system prompt content;
- equipped skills come from agent skills, auto-equipped skills, Pod defaults, and discoverable skills, excluding system prompt skills and de-duping across equipped sources.

The definition types now prevent system skills from returning `"equipped"`. `projects` remains the narrow global exception because it is equip-able outside Pods and system prompt content inside Pod conversations.

Also sorts skills in a stable way within the list of equipped skills.

## Tests

- Added a bunch of tests.
- Tested locally with pods/computer/go deep (all edge cases of auto smth).

## Risk

- Low.

## Deploy Plan

- Deploy front.

### DIFF
--- a/front/lib/resources/skill/code_defined/shared.ts
+++ b/front/lib/resources/skill/code_defined/shared.ts
@@ -15,16 +15,16 @@ export type MCPServerDefinition = {
   serverNameOverride?: string;
 };
 
-export type AutoAgentLoopAvailability = "enabled" | "equipped";
+export type AutoSkillModeForAgentLoop = "enabled" | "equipped";
 
-type AutoAgentLoopAvailabilityParams = {
+type AutoSkillModeForAgentLoopParams = {
   agentConfiguration: AgentLoopExecutionData["agentConfiguration"];
   conversation: ConversationWithoutContentType;
 };
 
-type AutoAgentLoopAvailabilityCallback<
-  T extends AutoAgentLoopAvailability = AutoAgentLoopAvailability,
-> = (params: AutoAgentLoopAvailabilityParams) => T | undefined;
+type AutoSkillModeForAgentLoopCallback<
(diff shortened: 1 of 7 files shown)


front/lib/resources/skill/code_defined/shared.ts
```diff
@@ -67,18 +62,18 @@ type WithDynamicInstructions<T extends BaseSkillDefinition> = T & {
   ) => Promise<string>;
 };
 
-export type SkillDefinition =
-  | WithStaticInstructions<BaseSkillDefinition>
-  | WithDynamicInstructions<BaseSkillDefinition>;
+export type SkillDefinition<
+  T extends "enabled" | "equipped" = "enabled" | "equipped",
```
flvndvd: `nit` given number of repetition, a type would be nice.

flvndvd (approved):

front/lib/resources/skill/code_defined/shared.ts
aubin-tchoi: fair enough

PR #28579, opened 2026-07-07
fix: use "Computer" feature name in user-facing Sandbox strings

adrsimon:
## Description

A few customer-facing strings still said "Sandbox" instead of the "Computer" feature name. Renames them in the Computer admin toasts and the egress-domain approval prompt. Code, variables and backend keep the `sandbox` naming.

Fixes https://github.com/dust-tt/tasks/issues/9263

## Tests

Manual.

## Risk

None, copy-only change.

## Deploy Plan

Standard.


aubin-tchoi (approved):
LGTM

PR #28581, opened 2026-07-07
Print applied migration SQL for #engineering_papertrail

Nils-Fedrigo:
## Description

When running `./scripts/run-migrations.sh pre-deploy front` (and its siblings: `post-deploy`, `connectors`), the operator had no easy way to see *what* SQL was actually applied so they can paste it into `#engineering_papertrail`.

This change makes the migration runner print, for every migration it just applied, the migration **name** followed by the **full SQL contents** (`cat`) in a clearly delimited, copy-paste-friendly block.

- `scripts/db/migration-runner.ts`: after `runUp` logs `"Migrations applied."`, it dumps each applied migration's name and file contents to stdout via the `printAppliedMigrationSql` helper. This now runs **unconditionally** — no env var is required.
- `scripts/run-migrations.sh`: no longer needs to opt in via an env var. Also restores the executable bit (`755`) on the script.

Because the runner only prints migrations that were *actually* applied, the output naturally appears once (in the first region that still had pending migrations) and stays empty in regions where nothing was pending.

Output format (name, then cat):

```
========================================================================
pre-deploy/20240101_add_col.sql
========================================================================
SET SESSION statement_timeout = 0;
ALTER TABLE "users" ADD COLUMN "nickname" VARCHAR(255);
========================================================================
```

## Tests

Manually validated the `printAppliedMigrationSql` logic in isolation:
- Prints the phase-prefixed name, separators, and the full SQL for each applied migration.
- Handles a missing migration file gracefully with a `(migration file not found: ...)` line instead of throwing.
- Confirmed the `migrations/<name>` path reconstruction works because the applied name is phase-prefixed (`pre-deploy/<file>.sql`) and the runner `cd`s into the component directory.

Note that the SQL now prints on every migration run, including local `npm run migration:apply`.

### DIFF
--- a/scripts/db/migration-runner.ts
+++ b/scripts/db/migration-runner.ts
@@ -1,5 +1,5 @@
 import { execFileSync } from "child_process";
-import { existsSync } from "fs";
+import { existsSync, readFileSync } from "fs";
 import { QueryTypes } from "sequelize";
 import type { Sequelize } from "sequelize";
 import type { MigrationParams } from "umzug";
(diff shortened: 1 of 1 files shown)


Nils-Fedrigo:

- #28581 ◀

PopDaph (approved):
\o/

PR #28582, opened 2026-07-07
Always cache the system prompt instructions block for 1h

flvndvd:
## Description

The instructions block of the system prompt was only given the 1h cache TTL when the conversation had no conditional just in time tools, falling back to the default 5 minutes otherwise. That guard is obsolete: the content that varies with conversation state, namely the tools, and its directives and the server listing, has since moved to the shared context tier / deferred tools search, so the instructions block is now stable per agent version and workspace settings. This change makes the 1h TTL unconditional in both the model constructors path and the legacy Anthropic client, and removes the flag that was threaded through the agent loop solely to drive the downgrade, since nothing consumes it anymore.

### DIFF
--- a/front/lib/api/assistant/jit_actions.ts
+++ b/front/lib/api/assistant/jit_actions.ts
@@ -162,10 +162,7 @@ export async function getJITServers(
     conversation: ConversationWithoutContentType;
     attachments: ConversationAttachmentType[];
   }
-): Promise<{
-  servers: ServerSideMCPServerConfigurationType[];
-  hasConditionalJITTools: boolean;
-}> {
+): Promise<ServerSideMCPServerConfigurationType[]> {
   const mcpServersToFetch = new Set<AutoInternalMCPServerNameType>(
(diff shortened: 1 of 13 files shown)


front/lib/api/assistant/jit_actions.ts
```diff
@@ -204,8 +201,5 @@ export async function getJITServers(
     }),
   ]);
 
-  return {
-    servers: [...baseServers, ...conditionalServers],
-    hasConditionalJITTools: conditionalServers.length > 0,
```
aubin-tchoi, line 209: ideally we double check that every eager tool is on an `auto_hidden_builder` MCP server, otherwise this can burst the cache quite heavily. may not be too hard to constrain in typing either

aubin-tchoi (approved):
LGTM

front/lib/api/assistant/jit_actions.ts
flvndvd: Outside of the bash tool, they are all always accessible.

PR #28584, opened 2026-07-07
fix(front): guard against undefined pod members to avoid page crash

adrsimon:
## Description

Fixes https://github.com/dust-tt/tasks/issues/9420

The pod page (`/w/:wId/pods/:podId`) crashed for some pods with:

```
TypeError: Cannot read properties of undefined (reading 'filter')
```

The failing code is `PodHeaderActions` running `members.filter((m) => m.isEditor)` with `members` being `undefined`. The header renders first in the tree, so its `members.filter` threw before the sibling `members.length` accessors — hence the `'filter'` message. A point-fix in the header would just move the crash to the next consumer.

While the space GET endpoint always returns `members`, a partial/stale cache entry (e.g. the bare `SpaceType` shape) can lack it. The sibling `PodMenu` already guards defensively (`podInfo?.members?.filter(...) ?? []`), confirming the field is known to be occasionally absent.

This normalizes `members` to an array at the `useSpaceInfo` boundary so every consumer is protected at once, keeping the original object reference when `members` is already present to avoid re-renders.

## Tests

Typechecked and linted.

## Risk

Very low.

## Deploy Plan

Standard front deploy, no migration.

### DIFF
--- a/front/lib/swr/spaces.ts
+++ b/front/lib/swr/spaces.ts
@@ -38,7 +38,7 @@ import { MIN_SEARCH_QUERY_SIZE } from "@app/types/core/utils";
 import type { DataSourceViewType } from "@app/types/data_source_view";
 import { isString } from "@app/types/shared/utils/general";
 import type { PodType, SpaceKind, SpaceType } from "@app/types/space";
-import type { LightWorkspaceType } from "@app/types/user";
+import type { LightWorkspaceType, SpaceUserType } from "@app/types/user";
 import { useCallback, useMemo } from "react";
 import type { Fetcher, KeyedMutator } from "swr";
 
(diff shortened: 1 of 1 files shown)


Fraggle (approved):
LGTM but I don't understand how this can happen.

adrsimon:
tbh i'm quite puzzled to and it's scoped to a very little amount of pods 

https://app.datadoghq.eu/logs?query=%2A%3A"TypeError%3A%20Cannot%20read%20properties%20of%20undefined%20%28reading%20%27filter%27%29"&agg_m=count&agg_m_source=base&agg_q=%40view.url_details.path&agg_t=count&analyticsOptions=%5B"bars"%2C"dog_classic"%2Cnull%2Cnull%2C"value"%5D&clustering_pattern_field_path=message&cols=host%2Cservice&flat_group_bys=true&messageDisplay=inline&refresh_mode=sliding&sort_m=count&sort_t=count&storage=hot&stream_sort=desc&top_n=10&top_o=top&viz=toplist&x_missing=true&from_ts=1783255927845&to_ts=1783428727845&live=true

Capture d’écran 2026-07-07 à 14 54 28

this is the only .filter i found though

PR #28586, opened 2026-07-07
Analytics/pricing: Add API keys in the usage graph

fabiencelier:
## Description

Closes https://github.com/dust-tt/tasks/issues/9427

Add the possibility to filter and groupby api key in the credits usage graph.
Add a "Not API" key to group values not matching an API key so the total is exact (also allow to filter on not API keys while drilling down on other fields).

There is a migration to remove the "DustSystemKey" value from the analytics (and not put it again) because it is not an actual value which should be shown to customers.

## Tests

Tested locally

image

image

## Risk

low

## Deploy Plan

Deploy then run ES migration
```
npx tsx scripts/execute_elasticsearch_http.ts --file lib/analytics/migrations/20260707_remove_system_api_key_name.http--execute
```

### DIFF
--- a/front/lib/analytics/indices/agent_message_analytics_2.mappings.json
+++ b/front/lib/analytics/indices/agent_message_analytics_2.mappings.json
@@ -24,6 +24,12 @@
     "context_origin": {
       "type": "keyword"
     },
+    "auth_method": {
+      "type": "keyword"
+    },
+    "api_key_name": {
+      "type": "keyword"
+    },
     "timestamp": {
       "type": "date"
     },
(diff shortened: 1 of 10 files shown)


front/lib/analytics/indices/agent_message_analytics_2.mappings.json
```diff
@@ -24,6 +24,12 @@
     "context_origin": {
       "type": "keyword"
     },
+    "auth_method": {
+      "type": "keyword"
+    },
+    "api_key_name": {
+      "type": "keyword"
+    },
```
fabiencelier, line 32: Change already applied in prod by this migration:

https://github.com/dust-tt/dust/blob/c9f16e96bfb5c9fea4ac169f3c5fa1f6010336c6/front/lib/analytics/migrations/20251118_add_api_key.http#L1-L12

rfrenoy (approved):

PR #28587, opened 2026-07-07
fix: RFC 2047-encode Gmail recipient display names

adrsimon:
## Description

The `To`/`From`/`Cc`/`Bcc` headers were built with raw UTF-8 display names, while only `Subject` was RFC 2047-encoded. When the message is base64-encoded, the header carries undeclared 8-bit bytes, which mail clients interpret as Latin-1.

Encode the display-name portion of each address header when it contains non-ASCII characters, leaving the `<email>` part and all-ASCII names untouched. Already-encoded ASCII words pass through unchanged, so there's no double-encoding regardless of whether Gmail returns names decoded or pre-encoded.

## Tests

Added unit tests in `helpers.test.ts` covering accented names, quoted names, all-ASCII passthrough, bare addresses, and empty display names. All gmail helper tests pass; typecheck and biome clean.

## Risk

Low.

## Deploy Plan

### DIFF
--- a/front/lib/api/actions/servers/gmail/helpers.test.ts
+++ b/front/lib/api/actions/servers/gmail/helpers.test.ts
@@ -1,5 +1,6 @@
 import type { GmailMessagePayload } from "@app/lib/api/actions/servers/gmail/helpers";
 import {
+  encodeEmailAddressHeader,
   findAttachmentData,
   findAttachmentIdByPartId,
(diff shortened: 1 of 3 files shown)


fabiencelier (approved):
LGTM, assuming the mail client can actually decode this

adrsimon:
we're already using this same code for subject so subject would be broken if it wasn't supported

PR #28588, opened 2026-07-07
Eagerly load custom agents' configured tools

flvndvd:
## Description

When tool search is enabled, a custom agent's hand-picked tools were deferred behind search like everything else. Those tools are the agent's identity: the builder chose them deliberately, and making the model discover them through search adds a round trip and a recall risk, for no cache benefit since the set _is expected to be small_ and stable per agent version. This change keeps every tool coming from the agent's configured actions in the eagerly loaded set when the agent is not a global one. Global agents keep their curated hot set, and tools added by skills, conversation state or client-side servers stay deferred for everyone, so mid-conversation additions still append to the deferred catalog instead of rewriting the cached prefix.

### DIFF
--- a/front/temporal/agent_loop/lib/run_model.test.ts
+++ b/front/temporal/agent_loop/lib/run_model.test.ts
@@ -161,9 +161,9 @@ describe("buildBaseSpecifications", () => {
       }
     );
 
-    expect(specifications.find((s) => s.name === "configured_tool")?.eager).toBe(
-      true
-    );
+    expect(
+      specifications.find((s) => s.name === "configured_tool")?.eager
+    ).toBe(true);
     expect(
       specifications.find((s) => s.name === "jit_tool")?.eager
     ).toBeUndefined();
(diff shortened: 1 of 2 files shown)


front/temporal/agent_loop/lib/run_model.test.ts
```diff
@@ -66,6 +71,174 @@ function makeConversation(
   return { messages };
 }
 
+function makeServerSideToolConfiguration(
```
aubin-tchoi, line 74: nit: could be shared somewhere, maybe in a factory class

front/temporal/agent_loop/lib/run_model.ts
```diff
@@ -203,6 +212,38 @@ function buildReplayOnlyToolSpecification(
...
+      .filter(isServerSideMCPServerConfiguration)
+      .map((action) => action.mcpServerViewId)
+  );
+
+  return availableActions.map((action) => {
+    const specification = buildToolSpecification(action);
+    if (
+      isCustomAgent &&
+      isServerSideMCPToolConfiguration(action) &&
+      agentActionViewIds.has(action.mcpServerViewId)
```
aubin-tchoi, line 238: what happens for tools added from the input bar, they don't appear here, right?

aubin-tchoi (approved):
LGTM

front/temporal/agent_loop/lib/run_model.ts
flvndvd: yep, they will be deferred.

front/temporal/agent_loop/lib/run_model.test.ts
flvndvd: Coming in the next wave of PRs.

PR #28589, opened 2026-07-07
feat: snapshot browser tab state at message-send time

adrsimon:
## Description

`list_browser_tabs` read live tab state at tool-execution time. The agent can take several seconds to reach the tool, so if the user switched or navigated tabs in the meantime, the stale tab leaked into the result and the agent acted on the wrong one.

This snapshots the tab list when the user sends a message and serves `list_browser_tabs` from that snapshot (falling back to a live query if none exists yet).

- `GET_ACTIVE_TAB` stays live on purpose — its only no-`tabId` caller is the manual "attach current tab" button, an immediate user action that should reflect the current tab.
- The send-time trigger goes through an optional `onBeforeSubmit` callback on `InputBarContext`, so no `chrome.*` APIs leak into shared `front` code.

## Tests

Manual. No test harness exists for the extension background message handlers.

## Risk

Low. Worst case, `list_browser_tabs` shows a slightly stale list within a turn; falls back to a live query when no snapshot exists. `onBeforeSubmit` is a no-op outside the extension. Safe to rollback.

## Deploy Plan

Standard front + extension deploy, no ordering constraints.

### DIFF
--- a/extension/shared/background.ts
+++ b/extension/shared/background.ts
@@ -28,8 +28,11 @@ import type {
   GetSessionInfoMessage,
   GetSessionInfoResponse,
   InputBarStatusMessage,
+  SnapshotTabStateMessage,
+  SnapshotTabStateResponse,
   TabActionMessage,
   TabActionResponse,
+  TabInfo,
   TypeTextMessage,
   TypeTextResponse,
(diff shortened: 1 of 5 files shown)


tdraier (approved):
👍

PR #28593, opened 2026-07-07
[front] fix: defer monthly→yearly seat switches to end of period

tdraier:
## What

Switching a seat from monthly to yearly billing now **defers to the end of the current period** in all cases, rather than applying immediately.

## Why

`classifySeatChange` (`front/lib/metronome/seats.ts`) decides whether a seat change is applied immediately or deferred. It deferred only true downgrades (new AWU allocation < previous) and applied everything equal-or-higher immediately.

A monthly→yearly switch of the same tier (e.g. `pro → pro_yearly`) shares the **same** allocation, so it was applying mid-period — the same for `pro → max_yearly`. Committing to annual billing should take effect at the next period boundary, mirroring how downgrades (e.g. `max → pro`) already behave.

## Change

- Added an `isMonthlyToYearlySwitch` check: previous seat is a paid **monthly** seat (`pro`/`max`/`workspace`) and the new seat is a `_yearly` variant.
- Such switches now take the **deferred** branch, anchored to the previous seat's next AWU credit renewal (same end-of-period anchor downgrades use), **even when the target tier is higher**.
- Additions from `free`/`none` to a yearly seat stay immediate (fresh subscription); yearly→monthly is unchanged.

Applies to the customer-facing seat-change path (`PATCH /api/w/:wId/members/:uId/seat-type` → `updateMembershipSeatAndTrack`). The forced-`immediate` internal path (e.g. auto seat upgrade) bypasses the classifier by design and is unaffected.

## Tests

Added 3 cases to `seats.test.ts`:
- `pro → pro_yearly` → deferred to next renewal
- `pro → max_yearly` (higher tier) → deferred
- `pro_yearly → pro` → immediate (reverse direction unaffected)

All 57 tests pass; `tsgo --noEmit` clean.

### DIFF
--- a/front/lib/metronome/seats.test.ts
+++ b/front/lib/metronome/seats.test.ts
@@ -625,6 +625,81 @@ describe("classifySeatChange", () => {
     expect(classify("pro", "max")).toEqual({ kind: "immediate" });
   });
 
+  it("defers a monthly→yearly switch of the same tier (pro → pro_yearly) to the next credit renewal", () => {
+    // Same allowance, so not a downgrade — but committing to annual billing is
+    // deferred to end of period rather than applied mid-period.
+    const { contract: c, productSeatTypes: types } = makeContract({
+      seats: [
+        { seatType: "pro", awu: 100 },
+        { seatType: "pro_yearly", awu: 100, frequency: "ANNUAL" },
+      ],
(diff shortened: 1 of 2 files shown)


thervier (approved):
LGTM

PR #28594, opened 2026-07-07
[front] business_activation: add panic logs if failure after payment

thervier:
## Description

Add panic logs for business activation (when upgrading to a Business plan) when there is a failure in handleSubscriptionActivationSuccess because we want to have an alert when payment was successful but activation of business plan or seat change has actually failed.

## Tests

N/A

## Risk

Low, logs only

## Deploy Plan

Deploy front


fabiencelier (approved):
✅

PR #28489, opened 2026-07-06
feat(front): thread optional modelSelection through the message-send path

Nils-Fedrigo:
## Description

Threads an optional per-message model override through the message-send path, so the input-bar model picker (#28049) has a channel to reach the backend.

- Carries an optional `ModelSelectionType` from the conversation submit handlers (`ConversationContainer`, `ConversationViewer`) down into `useSubmitMessage` and `useCreateConversationWithMessage`.
- Adds a `disabled` option to the `useModels` SWR hook so the picker can skip fetching when it isn't shown.

No producer exists yet — the value is always `undefined` at this point in the stack — so this PR is behavior-neutral on its own. The picker that actually sets a selection lands in #28049.

Part of the model-picker stack (bottom → top): #28344 → #28346 → #28488 → **#28489 (this)** → #28049.

## Tests

No behavior change to exercise; covered by typecheck. End-to-end selection is tested in #28049.

## Risk

None — pure optional plumbing, no runtime behavior change until a selection is produced.

## Deploy Plan

None.

PR #28595, opened 2026-07-07
feat(front): persist model picker selection and surface per-message model

Nils-Fedrigo:
## Description

Builds on the input-bar model picker. Three related changes so that a per-message model override behaves like the agent selection and stays visible:

- **Persist the picker selection across messages.** The model-picker selection now lives in `InputBarContext` (like the sticky agent) instead of local component state, so it no longer resets to "Auto" when the input bar remounts (sending a message, navigating between conversations). Removed the now-redundant `onModelSelectionChange` callback chain — the context is the single source of truth, read at submit time.
- **Show the model used on the agent message header.** When a message ran on a per-message override, the message header now displays `with <model>` (e.g. "with GPT 5.5 High") next to the agent name, in a muted style. This required threading the already-existing `requestedModel` through `getLightAgentMessageFromAgentMessage` and adding it (optional) to `LightAgentMessageType`.
- **Restore the picker from the conversation on reload.** On loading an existing conversation, the picker is seeded from the model the latest agent message ran on (its `requestedModel`, or "Auto" if none), mirroring how the sticky agent is re-derived from history. Applied once per conversation so it does not clobber the user's manual picker changes.

## Tests

Manual, in an existing conversation with the `models_picker` flag enabled:
- Pick a model override, send several messages → picker keeps the override (does not reset to Auto).
- Each agent message that ran on an override shows "with <model>" next to the agent name.
- Reload the page → picker is set to the latest message's model; a conversation whose last message ran on the agent's own model shows "Auto".

### DIFF
--- a/front/components/assistant/conversation/AgentInputBar.tsx
+++ b/front/components/assistant/conversation/AgentInputBar.tsx
@@ -2,6 +2,7 @@ import { useBlockedActionsContext } from "@app/components/assistant/conversation
 import { ContextUsageWarningBanner } from "@app/components/assistant/conversation/ContextUsageWarningBanner";
 import { useGenerationContext } from "@app/components/assistant/conversation/GenerationContextProvider";
 import { InputBar } from "@app/components/assistant/conversation/input_bar/InputBar";
+import { InputBarContext } from "@app/components/assistant/conversation/input_bar/InputBarContext";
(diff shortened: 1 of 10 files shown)


Nils-Fedrigo:

- #28344
    - #28346
        - #28488
            - #28489
                - #28049
                    - #28595 ◀

PR #28600, opened 2026-07-07
[front] feat: publish flow compat gate + reconcile integration

id13:
## Description

**Stack 8/11 (pod-state), top of the publish sub-stack.** Wires the gate into `publishSandboxFunction`, per the design's publish pipeline: build → compat check → reconcile → store.

- **Per-pod lock, deliberately narrow**: the Redis lock helper has a hard 5s TTL with no renewal, so the ~2min build runs OUTSIDE the lock; only compat-check → reconcile → store is serialized (`sandbox_function:publish:{podId}`). Because the TTL can still expire mid-section (reconcile execs into a sandbox), correctness rests on: a **fencing re-read** of sibling manifests + pure re-diff immediately before store, an **optimistic `updatedAt` guard** at the update path (mismatch → typed `publish_conflict`; first publishes are covered by the unique slug index), and a metric+warn (`sandbox_functions.publish_lock_ttl_exceeded.count`) so we can see how often the section outlives the TTL. The lock helper throws on acquisition timeout — mapped to `publish_conflict` via an acquired-flag so genuine bugs still propagate (ERR1).
- **Reconcile step** (`dsbx_db.ts`): `dsbx db reconcile` per declared database via sandbox exec, cloning the build exec pattern (absolute dsbx path, shellEscape + `--` before model-influenced operands, `agent-proxied`, 60s timeout, last-line zod-validated envelope). Runner error kinds map to two codes: model-correctable refusals → `reconcile_blocked` (tracked:false at the MCP boundary — the agent self-corrects), infrastructure → `reconcile_failed` (tracked).
- **Blocks are actionable**: the `compat_blocked` message lists each `(db.table.column, breaks: functions)` plus the additive migrate path (keep old columns declared, add alongside, dual-write, republish siblings) — the design's exact error-message contract.
- GCS bundle upload stays inside the section on purpose: re-publish overwrites the live bundle in place, so uploading before the gates pass would corrupt the published function on a late block.
- The bundle + manifests are stored **only after a clean reconcile apply**; the Ok payload carries warnings + stale-sibling notes for the publish tool upstack.

## Tests

9 publish vitest cases at this commit (mocked sandbox boundary per front convention): manifests stored + reconcile invoked per db with source-relative schema paths, compat block before any reconcile with nothing stored, reconcile refusal propagation (single- and multi-db, stop at first failure), warnings + stale-sibling outcome. `tsgo` clean.

## Risk

Medium: this is the write path for published functions. Mitigations above (narrow lock + fencing + optimistic guard + additive-only reconcile). Residual, documented: the 5s lock is best-effort; true serialization would need a renewable lock/fencing token, which front doesn't have today. Feature-flagged surface (`sandbox_functions`).

## Deploy Plan

After the manifests migration (two PRs down). Front-only; tolerates older sandbox images (dsbx errors surface as typed publish errors).

### DIFF
--- a/front/lib/api/actions/servers/sandbox_functions/metadata.ts
+++ b/front/lib/api/actions/servers/sandbox_functions/metadata.ts
@@ -45,9 +45,12 @@ export const SANDBOX_FUNCTIONS_TOOLS_METADATA = createToolsRecord({
     description:
       "Publish a sandbox function from a TypeScript source file in the current pod. The source " +
       "must default-export a `fetch(request: Request): Promise<Response>` handler and export a " +
-      "`schema` with zod `input` and `output`. It is bundled on the pod sandbox (only `zod` is " +
-      "available to import) and its input and output JSON schemas are extracted from the `schema` " +
-      "export. Re-publishing the same slug replaces the previous version.",
+      "`schema` with zod `input` and `output` (plus an optional `databases` list of the pod " +
+      "databases it opens, each backed by a `databases/{db}.db.ts` drizzle schema file). It is " +
+      "bundled on the pod sandbox (`zod`, `drizzle-orm` and `@dust/pod` are available to import) " +
+      "and its input and output JSON schemas are extracted from the `schema` export. Publish " +
+      "neither validates nor touches pod databases — schema files are applied and checked by " +
+      "`dsbx db reconcile`. Re-publishing the same slug replaces the previous version.",
     schema: {
       slug: z
         .string()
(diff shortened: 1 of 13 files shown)


id13:

- #28611
    - #28613
        - #28614
            - #28599
                - #28718
                    - #28600 ◀
                        - #28720
                            - #28601

spolu:
Hello! Someone said sandbox function?? 

I'm not making much sense out of the PR since marked as draft ignoring for now 👍

id13:
Hey @spolu you can ignore indeed 😄 

(especially because most of the code will be removed)

PR #28603, opened 2026-07-07
fix(front): resolve conversation files in convert_file_format

adrsimon:
## Description

`convert_file_format` only knew how to handle two kinds of input: a URL or a `fil_` sId. But `generate_file` writes text/CSV outputs straight to GCS via `DustFileSystem` and creates no `FileResource` row, so those files are referenced only by a canonical scoped path (`conversation-{id}/x.csv`) with a null `fileId`.

That path matched neither `validateUrl` nor `getResourceNameAndIdFromSId`, so it fell into the else branch that uploaded the raw `file_id_or_url` **string** to ConvertAPI as the file body. Result: the converted file contained the path text (e.g. `conversation-<id>/x.csv` in cell A1) instead of the actual data. User-uploaded files (real `fil_` sIds) hit the working branch and were unaffected.

Fix: resolve any non-URL input through `resolveConversationFileRef` (the same resolver already used by `image_generation` and the mail tools) which handles legacy fileIds, legacy scoped paths, and canonical scoped paths. We then hand ConvertAPI a signed URL via the existing `{ File: url }` convert call. Unresolvable inputs now return a proper "File not found" error instead of silently producing garbage.

Fixes dust-tt/tasks#9342

## Tests

- `npx tsgo --noEmit` and biome clean (pre-commit hooks pass).

## Risk

Low.

## Deploy Plan

Standard `front` deploy.

### DIFF
--- a/front/lib/api/actions/servers/file_generation/tools/index.ts
+++ b/front/lib/api/actions/servers/file_generation/tools/index.ts
@@ -3,6 +3,7 @@ import { Readable } from "node:stream";
 import { MCPError } from "@app/lib/actions/mcp_errors";
 import type { ToolHandlers } from "@app/lib/actions/mcp_internal_actions/tool_definition";
 import { buildTools } from "@app/lib/actions/mcp_internal_actions/tool_definition";
+import { resolveConversationFileRef } from "@app/lib/actions/mcp_internal_actions/utils/file_utils";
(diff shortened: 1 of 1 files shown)


flvndvd (approved):
IMHO we should change a bit the metadata for the tool.

PR #28604, opened 2026-07-07
Add the resource for `SandboxFunctionMCPActionModel`

PopDaph:
## Description

Follow up of #28446. Adds the resource for `SandboxFunctionMCPActionModel`:

- `makeNew` / `fetchById` / `markAsSucceeded` / `markAsErrored` — the row is born `running`,  executed from its persisted `toolConfiguration` snapshot.
- `writeOutput` / `readOutput`: the full `CallToolResult` content array goes to a **single GCS object** (`w/{wId}/mcp_output_items/{sId}/output.json`), written once at tool completion.
- Deletion with GCS cleanup as a single owner (`deleteAllForInvocation` / `deleteAllForSandboxFunction` / `deleteAllForMCPServerViews`), wired into the two paths that were TODO'd in #28446: invocation/function deletion and MCP server view hard-delete. The RESTRICT FKs are now safe ahead of any writer.
- GCS objects are deleted **after** the rows, post-commit via `transaction.afterCommit` when a transaction is provided (same shape as `invalidateAfterCommit` in `lib/utils/cache.ts`). A rollback leaves rows and objects consistent. Overkill??

Tests cover the GCS round trip, both RESTRICT regressions (invocation delete + view hard-delete with actions present → no FK violation, no orphaned GCS object) and transaction rollback (rows and objects both survive).

Follow-ups
- PR3: endpoints + Temporal execution, first e2e with `common_utilities`
- PR4: pod file outputs, flipping real servers

## Tests

Tests cover the GCS round trip and both RESTRICT regressions (invocation delete + view hard-delete with actions present → no FK violation, no orphaned GCS object).

## Risk

No caller creates actions yet. Safe to rollback.

## Deploy Plan

- [ ] Deploy front

### DIFF
--- a/front/lib/resources/sandbox_function_mcp_action_resource.test.ts
+++ b/front/lib/resources/sandbox_function_mcp_action_resource.test.ts
@@ -1,0 +1,267 @@
+import { SandboxFunctionInvocationResource } from "@app/lib/resources/sandbox_function_invocation_resource";
+import { SandboxFunctionMCPActionResource } from "@app/lib/resources/sandbox_function_mcp_action_resource";
+import { SandboxFunctionResource } from "@app/lib/resources/sandbox_function_resource";
+import { frontSequelize } from "@app/lib/resources/storage";
+import { FileFactory } from "@app/tests/utils/FileFactory";
+import { createResourceTest } from "@app/tests/utils/generic_resource_tests";
+import { MCPServerViewFactory } from "@app/tests/utils/MCPServerViewFactory";
+import { RemoteMCPServerFactory } from "@app/tests/utils/RemoteMCPServerFactory";
+import { SandboxFunctionMCPActionFactory } from "@app/tests/utils/SandboxFunctionMCPActionFactory";
+import { SpaceFactory } from "@app/tests/utils/SpaceFactory";
+import { sandboxFunctionContentType } from "@app/types/files";
+import type { JSONSchema7 as JSONSchema } from "json-schema";
+import { beforeEach, describe, expect, it, vi } from "vitest";
+
+// In-memory GCS mock: writes persist content that reads can return.
+const gcsStore = new Map<string, Buffer>();
+
+vi.mock("@app/lib/file_storage", () => ({
+  getPrivateUploadBucket: vi.fn(() => ({
+    file: vi.fn((path: string) => ({
+      save: vi.fn(async (data: Buffer) => {
+        gcsStore.set(path, data);
+      }),
+      download: vi.fn(async () => {
+        const buf = gcsStore.get(path);
+        if (!buf) {
+          throw new Error(`GCS file not found: ${path}`);
+        }
+        return [buf];
+      }),
+    })),
+    delete: vi.fn(async (path: string, opts?: { ignoreNotFound?: boolean }) => {
+      if (!gcsStore.has(path) && !opts?.ignoreNotFound) {
+        throw new Error(`GCS file not found: ${path}`);
+      }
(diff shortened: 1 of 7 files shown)


spolu (approved):
LGTM

front/lib/resources/sandbox_function_mcp_action_resource.test.ts
```diff
@@ -0,0 +1,267 @@
...
+import { createResourceTest } from "@app/tests/utils/generic_resource_tests";
+import { MCPServerViewFactory } from "@app/tests/utils/MCPServerViewFactory";
+import { RemoteMCPServerFactory } from "@app/tests/utils/RemoteMCPServerFactory";
+import { SandboxFunctionMCPActionFactory } from "@app/tests/utils/SandboxFunctionMCPActionFactory";
+import { SpaceFactory } from "@app/tests/utils/SpaceFactory";
+import { sandboxFunctionContentType } from "@app/types/files";
+import type { JSONSchema7 as JSONSchema } from "json-schema";
+import { beforeEach, describe, expect, it, vi } from "vitest";
+
+// In-memory GCS mock: writes persist content that reads can return.
```
flvndvd, line 15: I added a global mock, you should be able to use it here, it also expose a spy to assert it works as expected.

front/lib/resources/sandbox_function_mcp_action_resource.test.ts
```diff
@@ -0,0 +1,267 @@
...
+    delete: vi.fn(async (path: string, opts?: { ignoreNotFound?: boolean }) => {
+      if (!gcsStore.has(path) && !opts?.ignoreNotFound) {
+        throw new Error(`GCS file not found: ${path}`);
+      }
+      gcsStore.delete(path);
+    }),
+  })),
+}));
+
+const inputSchema: JSONSchema = {
```
flvndvd, line 41: Can we create a factory for those.

front/lib/resources/sandbox_function_mcp_action_resource.ts
```diff
@@ -0,0 +1,390 @@
...
+    } catch (err) {
+      return new Err(normalizeError(err));
+    }
+  }
+
+  // GCS objects are deleted best-effort AFTER the rows (post-commit when a transaction is
+  // provided): a failed or rolled-back row deletion must not leave rows pointing at deleted
+  // objects, while a failed GCS deletion only logs — the orphaned object is harmless.
+  private static async deleteAllWithOutputs(
+    where: {
```
flvndvd, line 249: Can we pick attributes from the `WhereOptions`?

front/lib/resources/sandbox_function_mcp_action_resource.ts
```diff
@@ -0,0 +1,390 @@
...
+          { err: gcsResult.error, pathCount: gcsPaths.length },
+          "Failed to delete sandbox function MCP action outputs from GCS"
+        );
+      }
+    };
+
+    if (transaction) {
+      // Same shape as `invalidateAfterCommit` in `lib/utils/cache.ts`: the callback never rejects
+      // today, but an unhandled rejection post-commit would be unattributable.
+      transaction.afterCommit(() =>
```
flvndvd, line 283: OOC why are we doing this? It seems that if we fail deleting, those files will remain dangling forever, right? Since not in DB anymore.

flvndvd (approved):

PopDaph:
Seen IRL merging as is but will apply all feedback in a follow up 🙏🏻

PR #28602, opened 2026-07-07
[front] feat: pass DUST_POD_DATABASES_DIR to sandbox function invocations

id13:
## Description

Set the env var for pod database

## Tests

None

## Risk

Low

## Deploy Plan

Deploy front

PR #28605, opened 2026-07-07
[dsbx] feat: add @dust/pod package

id13:
## Description

Adds `@dust/pod`, the runtime library for code running in a pod sandbox. First surface: `db(name)` — a cached Drizzle handle over the pod's live SQLite database at `${DUST_POD_DATABASES_DIR}/{name}.db`.

Design decisions:

- **Must-exist open** — `db()` never creates database files; the set of databases is fixed at publish time (`dsbx db reconcile`), so a typo'd name errors clearly instead of minting an empty database.
- **Per-database size cap** (1 GiB in production) via `PRAGMA max_page_count`, with `SQLITE_FULL` translated to `PodDatabaseFullError` on every execution path Drizzle uses. The database stays readable and rows can still be deleted, so the agent recovers in place by reclaiming space. *Not* a security boundary — workload code can write to the filesystem directly (or rewrite the env var); the sandbox's own disk is the hard limit.
- **Required env contract, no defaults below front** — front owns both the databases dir and the quota and passes them per exec through dsbx; `db()` fails loudly when either is absent or the quota is not a canonical positive decimal, instead of masking a broken launch context with a baked-in value.
- **`wal_autocheckpoint = 0`** — Litestream owns checkpointing: an application checkpoint slipping in between Litestream's own can make it miss WAL frames and force a full re-snapshot.

## Tests

`bun test` in `cli/dust-sandbox/pod` (pod guard, name contract, must-exist open, pragmas, quota clamp, instance caching, error translation), wired into `build-dust-sandbox.yml` along with the bundle build.

## Risk

It's a no-op

## Deploy Plan

- Merge

### DIFF
--- a/cli/dust-sandbox/pod/db.ts
+++ b/cli/dust-sandbox/pod/db.ts
@@ -67,7 +67,12 @@ export const POD_DATABASE_BUSY_TIMEOUT_MS = 5000;
 /** Valid database names (also the manifest/publish contract). */
 export const POD_DATABASE_NAME_REGEX = /^[a-z][a-z0-9_]{0,63}$/;
 
-export class PodDatabaseError extends Error {}
+export class PodDatabaseError extends Error {
+  constructor(message: string) {
+    super(message);
+    this.name = "PodDatabaseError";
+  }
+}
 
 export class PodDatabaseInvalidNameError extends PodDatabaseError {
   constructor(name: string) {
--- a/cli/dust-sandbox/pod/db.ts
+++ b/cli/dust-sandbox/pod/db.ts
@@ -1,0 +1,338 @@
+import type { Changes, SQLQueryBindings, Statement } from "bun:sqlite";
+import { Database } from "bun:sqlite";
+import type { BunSQLiteDatabase } from "drizzle-orm/bun-sqlite";
+import { drizzle } from "drizzle-orm/bun-sqlite";
+
+/**
+ * Pod state databases.
+ *
+ * `db(name)` returns a cached Drizzle instance over the pod's live SQLite
+ * database at `${DUST_POD_DATABASES_DIR}/{name}.db`. Databases are created by
+ * the publish pipeline (`dsbx db reconcile`), never here: the file is opened
+ * must-exist so a typo'd name errors clearly instead of minting an empty
+ * database. Functions that never call `db()` pay nothing.
+ *
+ * Disk guardrails — pod state must not fill the sandbox disk by accident:
+ *
+ * - `db()` cannot create database files (must-exist open), so the set of
+ *   databases is fixed at publish time. Total pod-state footprint is the
+ *   number of declared databases times the per-database cap; bounding the
+ *   count is the publish pipeline's job.
+ * - Each database is capped at {@link DEFAULT_POD_DATABASE_MAX_SIZE_BYTES}
+ *   (1 GiB) via `PRAGMA max_page_count`. Writes past the cap fail with
+ *   {@link PodDatabaseFullError}; the database stays readable and rows can
+ *   still be deleted, so the agent recovers in place by reclaiming space.
+ * - The cap override env var is clamped down-only, since it is readable and
+ *   writable by the workload itself (see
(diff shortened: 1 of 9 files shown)


id13:

- #28605 ◀
    - #28606
        - #28602

cli/dust-sandbox/pod/index.ts
```diff
@@ -0,0 +1,282 @@
+import type { Changes, SQLQueryBindings, Statement } from "bun:sqlite";
```
id13: probably moving most of these to a db.ts, probably needs split

cli/dust-sandbox/pod/package.json
```diff
@@ -0,0 +1,14 @@
+{
+  "name": "@dust/pod",
+  "version": "0.1.0",
+  "private": true,
+  "description": "Pod state runtime: db(name) returns a Drizzle instance over the pod's bun:sqlite database.",
```
id13: too narrow

cli/dust-sandbox/pod/index.ts
```diff
@@ -0,0 +1,282 @@
...
+  const row = sqlite.query<{ page_size: number }, []>("PRAGMA page_size").get();
+  if (row === null) {
+    throw new PodDatabaseError("PRAGMA page_size returned no row");
+  }
+  // The quota in pages of this database's actual page size. max_page_count is
+  // per-connection, so it is re-applied on every open; SQLite clamps it up to
+  // the current page count if the database already exceeds the quota.
+  const maxPageCount = Math.max(1, Math.floor(maxSizeBytes / row.page_size));
+  sqlite.exec(`PRAGMA max_page_count = ${maxPageCount}`);
+}
```
id13: This should explain the design choices and the reason for each pragma, a bit too lite for now

id13:
PMRR

cli/dust-sandbox/pod/db.ts
```diff
@@ -0,0 +1,350 @@
...
+}
+
+/**
+ * A bun:sqlite Database that translates `SQLITE_FULL` (the quota surface of
+ * `PRAGMA max_page_count`) into {@link PodDatabaseFullError} on every
+ * execution path Drizzle uses: `exec`/`run` directly, and statements obtained
+ * through `prepare` (which `query()` and `transaction()` also go through).
+ */
+class PodSqliteDatabase extends Database {
+  readonly #dbName: string;
```
flvndvd, line 187: Is this new to prefix with `#`?

cli/dust-sandbox/pod/db.ts
```diff
@@ -0,0 +1,350 @@
...
+
+  constructor(path: string, dbName: string, maxSizeBytes: number) {
+    // readwrite without create: opening a missing file throws SQLITE_CANTOPEN
+    // instead of minting an empty database (verified against bun 1.3.14).
+    super(path, { readwrite: true });
+    this.#dbName = dbName;
+    this.#maxSizeBytes = maxSizeBytes;
+  }
+
+  run<ParamsType extends SQLQueryBindings[]>(
```
flvndvd, line 198: If the goal is to override the parent function, you can use the `override` keyword that ensures that the prototype remains the same.

cli/dust-sandbox/pod/index.test.ts
```diff
@@ -0,0 +1,356 @@
...
+  POD_DATABASE_BUSY_TIMEOUT_MS,
+  POD_DATABASE_MAX_SIZE_BYTES_ENV,
+  POD_DATABASES_DIR_ENV,
+  POD_SPACE_ID_ENV,
+  PodDatabaseError,
+  PodDatabaseFullError,
+  PodDatabaseInvalidNameError,
+  PodDatabaseNotDeclaredError,
+  PodDatabasesUnavailableError,
+} from "./index.ts";
```
flvndvd, line 20: nit use alias import.

flvndvd (commented):
LGTM

cli/dust-sandbox/pod/db.ts
id13: I don't know, but it's disgusting, will fix that

PR #28607, opened 2026-07-07
fix(wakeups): delete cron schedule when a wake-up is cancelled at fire time

rfrenoy:
## Description

When runWakeUpActivity hit a terminal condition (conversation gone/inaccessible, or agent_configuration_not_found / agent_inaccessible / model_disabled) it called markCancelled(), which only flips the DB row. For cron wake-ups the backing Temporal schedule was never deleted, so Temporal kept spawning daily workflows that no-op forever (until the maxFires cap).

The fire-time terminal paths now cancel the wake-up AND delete its cron schedule via cleanupTemporalScheduleIfCronTerminal (generalized from the former cleanupTemporalIfCronExpired to also cover the cancelled state). We deliberately do not use cancel(): for one-shot wake-ups that would cancel the very workflow this activity runs in. Deleting the cron schedule is safe from within a scheduled run.

Refs dust-tt/tasks#8477

## Tests

Unit
## Risk

Low

## Deploy Plan

Front

### DIFF
--- a/front/lib/resources/wakeup_resource.ts
+++ b/front/lib/resources/wakeup_resource.ts
@@ -658,10 +658,17 @@ export class WakeUpResource extends BaseResource<WakeUpModel> {
     });
   }
 
-  async cleanupTemporalIfCronExpired(
+  async cleanupTemporalScheduleIfCronTerminal(
     auth: Authenticator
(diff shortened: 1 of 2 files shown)


spolu (approved):
LGTM 🙏

PR #28597, opened 2026-07-07
[front] feat: pod-state lifecycle

id13:
## Description

Implement the new pod sandbox lifecycle hooks to on sleep, to do one last sync before.
 
* On cold start
   - restore all dbs in pod-state gcs folder to local sqlite dbs
   - start litestream replication
* On sleep
   - Do one last sync from litestream, on happy path it's a no-op, it's more of a sanity check to alert in case of data corruption than anything else.

## Tests

- Unit tests
- Tested locally (see #28596 )

## Risk

Risk: Low
Blast radius: conversation sandboxes (if we broken somehow their lifecycle)

## Deploy Plan

- [ ] Deploy front
- [ ] killswitch sandbox

PR #28596, opened 2026-07-07
[front] feat: add litestream and pod-state layout to sandbox image

id13:
## Description

🪨 Img foundations for app/pod state db

- **litestream v0.5.13 binary** at `/opt/bin/litestream` (root:root 755).
- **`dust-state` system user** (primary group `dust-state`, supplementary `agent`, nologin) runs the replication daemon. **3-location split** of `/pod-state`:
  - `/pod-state/databases`: live SQLite files, `dust-state:agent` 2770 + default ACLs (same hardening block as `/files`): uid-1003 function code and the daemon both need rw. SQLite never touches gcsfuse.
  - `/pod-state/replica`: gcsfuse mount point for the litestream file replica, `dust-state:dust-state` 0700: must be invisible to the workload.
- **`litestream.service`** (User=dust-state, Restart=always, hardening modeled on dust-egress-resolver.service) + a **static `/etc/litestream.yml`** using litestream's **directory watcher** (`dir` + `pattern: "*.db"` + `watch: true`): databases created at any point, including publish-time `dsbx db reconcile` are discovered and replicated within seconds. The unit is **deliberately not enabled at boot**: front starts it at runtime after the replica mount + restore at boot the daemon would write to the unmounted local directory (the design's silent-unmount failure mode) and manage files mid-restore. Also, because we're still sharing imgs between pod/conv sandboxes, so conv would be paying for something they're not using.

- **npm pins** `drizzle-orm@0.45.2` / `drizzle-kit@0.31.10`, and **`@dust/pod` vendored** by a build-time bun bundle into `/opt/npm-global/lib/node_modules/@dust/pod

- no need to killswitch since no-op 

## Tests

- `registry.test.ts` (17): version/sha pins, dust-state user, `/pod-state` modes + ACLs, unit content + **not-enabled-at-boot**, static watcher-config content, all new ops ordered **before** the final hardening re-run.
- `pod_package.test.ts` (2): repo-root resolution pinned against markers (the source dir lands in a sibling PR).
- `sandbox_security_check` unit tests (12) + new **live probes** run by CI against the built template: `/pod-state` ownership audit, uid-1003 rw on `databases/`, uid-1003 denied on `replica/`, litestream binary + unit not workload-writable.

Manually tested using this script

<details><summary>Manually tested using this script (from the top of the stack)</summary>
<p>

```typescript

import { ensurePodSandboxReady } from "@app/lib/api/sandbox/lifecycle";
[...]

PR #28608, opened 2026-07-07
[front] feat: pod-state gcs prefix, replica mount profile and path helper

id13:
## Description

**PR 2/4 of the pod-state stack** (`design_docs/POD_STATE_DESIGN.md`, "GCS architecture and auth"): the GCS plumbing for the litestream replica — token scope, mount profile, path helper.

**What & why:**
- New **`pod_state` SandboxOnlyMount kind** → rw CAB prefix `w/{wId}/pods/{podId}/state` (no trailing slash: the CAB conditions append the `/` themselves). Pod sandboxes now mint **7 rules** (1 unconditional + 2×3 prefixes: files rw, sandbox-functions ro, state rw) of the CAB ceiling of 10 / adapter cap of 4 targets.
- **`GCSMountTarget.mountProfile`** (`"workload" | "pod_state_replica"`). The replica profile is mounted **as `dust-state` via runuser** instead of root-with-`allow_other`: with no `allow_other`, the FUSE layer only admits the *mounting* user — one mechanism that simultaneously keeps the replica usable by the litestream daemon and kernel-denied to the untrusted workload uid 1003 (and even root). A root-made mount without `allow_other` would lock out the daemon itself. Also per-profile: `--kernel-list-cache-ttl-secs=0` (restore must never see a stale LTX listing — design failure mode 3), file/dir modes 600/700.
- `pod_state_replica` targets are **skipped entirely (mount + CAB rules) on images without the `pod_state` capability**, so existing paused sandboxes keep refreshing tokens exactly as today.
- `getPodStateBasePath` path helper (trailing-slash form for object writers/prefix deletes, consistent with its siblings).

Stack (bottom-up): #28596 image → **this** → restore → sleep barrier (#28597).

## Tests

- `token.test.ts` (8 + manual CAB suite): rule shape, the 7-rule pod set, rw scope on the state prefix.
- `mount_path.test.ts` (52): helper family incl. the new state prefix.
- (The end-to-end wiring test that derives the CAB set from the real pod mount set lands with the barrier PR, which introduces the shared mount factory.)

## Risk

Pure plumbing, inert until the follow-ups: the state target only materializes for pod sandboxes on the `pod_state`-capable image. CAB budget checked twice (comment'd formula + adapter constructor throw above 4 targets).

## Deploy Plan

No ordering constraints; deploy with the rest of front. Does nothing until the restore/barrier PRs and the 0.8.51 image exist.

### DIFF
--- a/front/lib/api/file_system/backends/gcs_file_system_backend.ts
+++ b/front/lib/api/file_system/backends/gcs_file_system_backend.ts
@@ -608,18 +608,24 @@ export class GCSFileSystemBackend implements FileSystemBackend {
   ): SandboxMountAdapter {
     const bucket = fileStorageConfig.getGcsPrivateUploadsBucket();
     const targets: GCSMountTarget[] = [
-      ...mounts.map((mount) => ({
-        gcsPrefix: this.mountRootGCSPrefix(mount),
-        sandboxMountPoint: mount.sandboxMountPoint,
-        legacySandboxMountPoint: mount.legacySandboxMountPoint,
-        readOnly: false,
-      })),
-      ...sandboxOnlyMounts.map((mount) => ({
-        gcsPrefix: this.sandboxOnlyMountGCSPrefix(mount),
-        sandboxMountPoint: mount.sandboxMountPoint,
-        legacySandboxMountPoint: null,
-        readOnly: mount.readOnly,
-      })),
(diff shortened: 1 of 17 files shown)


id13:
Ai slope, ignore

PR #28609, opened 2026-07-07
[front] feat: pod-state cold-start restore and litestream daemon start

id13:
## Description

**PR 3/4 of the pod-state stack** (`design_docs/POD_STATE_DESIGN.md`, "Cold start: create, or destroy → recreate"): restore the pod's SQLite databases from the GCS replica on fresh sandbox creation, then start the litestream daemon.

**What & why:**
- **Cold-start sequence** (`setupPodStateOnColdStart`, wired in `ensureOwnerSandboxReady`'s freshlyCreated branch, strictly AFTER the gcsfuse mounts, awaited — `invoke` awaits readiness, which is what keeps functions off half-restored databases): enumerate `{db}.db` replica subdirs (the directory watcher names them by database *filename*) → per DB: `litestream restore` into a **same-dir temp file** (atomic rename at the end), `PRAGMA quick_check` integrity gate, `chmod 660` (group `agent` = function code rw) → `mv` into place → **then** start `litestream.service` (the watcher discovers the restored files and everything created later).
- `-if-replica-exists` + an explicit output-file check: a replica dir with **no restorable backup** (crash during the very first LTX upload leaving a stray `.tmp` object — design failure mode 5) skips the database with a warning instead of bricking every cold start of the pod.
- Everything touching the replica mount runs **as `dust-state` via runuser** (no `allow_other` ⇒ FUSE denies all other uids, including root); all root execs use `execRoot` + `RootCommand` absolute paths with `--` before workload-influenced operands (SEC3); enumeration output is filtered through the contract name regex (drops dotfiles/hostile names).
- **Failure self-heals via `requestKill()`**: `status=running` is committed before this step, so a plain error would make the next call take the warm path and silently serve a half-initialized sandbox (no restored DBs, no daemon). Kill-requesting instead makes the next access destroy + recreate and re-run setup from scratch.
- **Capability gate is a version threshold** (`dust-base` ≥ 0.8.51), NOT a registry lookup: the registry only holds the single *current* image, so exact-tag matching would silently turn pod-state handling OFF for every existing sandbox at the next image bump. Anomalous image identities alert (statsd + `logger.error`) rather than silently gating off; null/pre-versioning rows are legitimately pre-pod-state.
- Pod deletion / workspace scrub now delete the GCS state prefix (`deletePodStatePrefix` in `scrubSpaceActivity`): LTX objects are never FileResources, so per-file deletion paths can't reach them — without this the replica chain leaks forever.
- New startup phases (`pod_state_setup` + enumerate/restore_db/start_daemon) follow the existing `traceSandboxStartupPhase` pattern.

Known v1 tradeoff: a *concurrent* invoke against a cold pod can reach exec while the first caller is still mid-restore (no in-sandbox readiness sentinel); worst case is a transient must-exist-open failure. Accepted for v1.

Stack (bottom-up): #28596 image → mounts → **this** → sleep barrier (#28597).

## Tests

- `pod_state.test.ts`: name allowlist, replica/live enumeration parsers ({db}.db layout), dotted-version compare + threshold gate (incl. "0.8.50 = last image without pod state" and bump-survival cases).
- `sandbox_resource.test.ts` (28): lifecycle behavior unchanged with the new `requestKill`.
- Boundary typechecks standalone (`tsgo --noEmit`).

## Risk

Gated on `runtimeOwner.kind === "pod"` + the `pod_state` capability ⇒ strict no-op until 0.8.51 pod sandboxes exist. Restore failures block pod readiness by design (fail-loud, then self-heal via recreate); the empty-replica tolerance removes the main persistent-failure cause.

## Deploy Plan

Deploy with front, before or after the image build (no-op without the 0.8.51 template). No migration.

### DIFF
--- a/front/lib/api/sandbox/instrumentation.ts
+++ b/front/lib/api/sandbox/instrumentation.ts
@@ -54,7 +54,14 @@ export type SandboxStartupPhase =
   // the token, starts the token server, and polls it ready (was three execs).
   | "gcs.mint_token"
   | "gcs.token_server"
-  | "gcs.gcsfuse_mount";
+  | "gcs.gcsfuse_mount"
+  // Pod state bring-up (cold start only, after gcs_mount): restore replicated
+  // SQLite databases, then start the daemon (its static directory-watcher
+  // config is baked in the image).
+  | "pod_state_setup"
+  | "pod_state.enumerate"
+  | "pod_state.restore_db"
+  | "pod_state.start_daemon";
 
 // Opens a parent APM span for a startup phase. The provider.* child spans nest
 // underneath automatically, so this adds semantic grouping (and a per-phase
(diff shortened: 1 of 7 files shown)


id13:
Ai slope, ignore

PR #28611, opened 2026-07-07
[dsbx] feat: dsx db helpers

id13:
## Description

Add the base helpers to be used by the various dsbx db command that will be turned into `sandbox_function` MCP server tools (see [here](https://app.notion.com/p/dust-tt/S2-App-Pod-DB-38128599d94181ca9f44f8288a5ff2c5?source=copy_link#39828599d94180cdb7b0c8e77236a72d))

## Tests

Unit tests based on fixtures.

## Risk

None, this is dead-code so far

## Deploy Plan

- Merge

PR #28601, opened 2026-07-07
[front] feat: pod_functions skill v2

id13:
## Description

Update the sandbox function skill

## Tests

Tested locally end to end

## Risk

Low

## Deploy Plan

Merge

PR #28612, opened 2026-07-07
[front] feat: mirror sandbox function build manifests

id13:
## Description

**Stack 2/10 (pod-state, T2).** Front reads back the manifests that `dsbx function build` (#28611) writes into the schema file, and threads them through the build result (`manifests: FunctionManifests | null`, null = the function declares no databases). Storage and the compat gate land upstack (#28615, #28616).

The schema file comes out of the sandbox, where model-authored code runs, so front revalidates it rather than trusting the runner's build-time validation:

- `manifests.ts` is a zod parser for manifest.v1 — but not a second authored schema: the shape is authored once in `cli/dust-sandbox/functions-runner/manifest_types.ts` (dependency-free), and this file type-imports it (erased at compile time — front never bundles cli code, and the runner's zod 4 never meets front's zod 3) with two-way assignability tripwires, so any drift fails `tsgo` in the PR that introduces it. Runtime constants can't be type-checked, so `manifests.test.ts` asserts regex equality with the runner's `DB_NAME_REGEX` and rejection of every runner-reserved `Object.prototype`-colliding key (prototype-pollution defense — table/column/index names become record keys in front and JSONB).
- Manifest build-error kinds map to the model-correctable `invalid_contract` via a `Record<ManifestErrorKind, …>` keyed by the runner's union, so a renamed or added kind fails the typecheck instead of silently degrading to `internal` (`hasOwnProperty` guard, not `in`: the envelope kind is sandbox-influenced).
- `databases` stays optional in the schema-file parser so files produced by older dsbx images keep parsing.

## Tests

`npm run test -- lib/api/sandbox_functions/`: 27/27 (manifest parsing happy path, hostile `__proto__` manifest rejected, malformed shape rejected, error-kind mapping, mirror drift checks). `npx tsgo --noEmit` clean; verified the tripwire fires (TS2322) when a field is added runner-side.

## Risk

Low: no behavior change for functions without databases; `manifests` is threaded but unconsumed until upstack.

## Deploy Plan

None beyond the front deploy; independent of the image bump (older dsbx images simply produce no `databases` field).


id13:

- #28611
    - #28612 ◀
        - #28613
            - #28614
                - #28599
                    - #28718
                        - #28615
                            - #28616
                                - #28600
                                    - #28618
                                        - #28719
                                            - #28720
                                                - #28601

id13:
Folded into the restructured stack: the front pipeline now lands in its final shape in #28600 (the stored-schema mirror, manifests column and compat differ never exist in the new history), on top of the dsbx stack #28611→#28613→#28614→#28599→#28718. See #28600.

PR #28615, opened 2026-07-07
[front] feat: manifests column on sandbox functions (model + migration)

id13:
## Description

**Stack 6/10 (pod-state), first of the publish sub-stack.** Storage for the build-produced manifests: a nullable `manifests` JSONB column on `sandbox_functions` (design "publish 3": manifests are "stored verbatim … at publish"; null = the function declares no databases).

- Model: `inputSchema`/`outputSchema` JSONB pattern + a zod validate hook against the manifest.v1 mirror (null-safe — Sequelize runs custom validators on null).
- Migration generated with the real generator (`npm run migration:generate:pre-deploy`, pg-schema-diff against a pristine models-synced baseline): exactly one `ALTER TABLE … ADD COLUMN "manifests" jsonb`.
- Resource threading with a deliberate asymmetry: `makeNew` takes `manifests` **optional (default null)** — creation without databases genuinely means none — while `updateContent` **requires** it, because omitting it on a re-publish would silently null the stored manifests.
- `publishSandboxFunction` threads the build result's manifests through both store paths (the compat gate + reconcile step that consume them land two PRs up).

## Tests

`front`: `npx tsgo --noEmit` clean at this commit; `npm run test -- lib/resources/sandbox_function_resource.test.ts lib/api/sandbox_functions/publish_sandbox_function.test.ts` — 19/19. Migration applied for real against a dev DB (dust-hive env): `\d sandbox_functions` shows `manifests jsonb` nullable/no-default, `schema_migrations` records it, `migration:status` clean.

## Risk

Low: additive nullable column, no reader yet. Instant DDL (no default, no rewrite).

## Deploy Plan

Pre-deploy migration (additive) before the front deploy, per the standard migration flow.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #28614, opened 2026-07-07
[dsbx] feat: pod db schema introspection

id13:
## Description

- Add db command tests + fixtures, should have been in the previous PR 🤷 
- Add schema introspection (it's glorified drizzle-kit pull)

## Tests

Tested manually

## Risk

Low

## Deploy Plan

- Deploy front
- Killswitch sandboxes

PR #28616, opened 2026-07-07
[front] feat: pod database manifest compat differ

id13:
## Description

**Stack 7/10 (pod-state).** The pure manifest compatibility differ — the heart of the publish gate (design: "How to handle breaking changes? … we do NOT handle breaking changes by forbidding them, raising errors to the agent").

`diffManifestsAgainstSiblings` compares the manifests being published against the stored manifests of every other function of the pod declaring the same database, plus the function's own previous manifests. Semantics (frozen in manifest.v1):

- **BLOCK (structural)** — a sibling references something removed/changed: table missing, column missing, `type`/`notNull`/`primaryKey`/`autoIncrement` differ.
- **BLOCK (additive-but-breaking)** — new `NOT NULL` column without default on a table that already exists in any sibling manifest **or in this function's own previous manifests** (live rows predate the column even with no siblings).
- **WARN (mode drift)** — same column, same storage type, different mode: value-level divergence only (stored bytes identical), which is exactly why manifests carry modes — SQLite can't see this. Warn, never block; publish proceeds.
- **WARN (unique tightening)** — new `unique: true` index on a table a sibling uses (UNIQUE is allowed per the manifest.v1 deviation; tightening makes sibling writes fail, so it warns).
- **Indexes never block** (queries don't reference indexes); `hasDefault` changes never block.
- `computeStaleSiblings` — post-publish notes: siblings whose stored manifest for a shared db now differs (their bundles keep working; their embedded schema files are stale until republished — the design's accepted no-transactional-publish tradeoff).

Pure function, no I/O — blocks dedupe across siblings and carry the affected function list so the publish error can say *"removes chat.messages.body (breaks: list-messages)"*.

## Tests

18 vitest cases at this commit (`compat.test.ts`), deliberately including the design doc's worked examples: the `body → content` rename (blocked, names the sibling), the `report-activity` mode-drift warning (proceeds with the alignment message), and the `channel NOT NULL` addition (blocked with the additive-migrate hint). `tsgo` clean.

## Risk

None standalone — nothing calls it until the next PR.

## Deploy Plan

Standard front deploy (inert until wired).

### DIFF
--- a/front/lib/api/sandbox_functions/compat.test.ts
+++ b/front/lib/api/sandbox_functions/compat.test.ts
@@ -1,20 +1,20 @@
 import type {
   CompatDiff,
-  SiblingManifests,
+  SiblingState,
 } from "@app/lib/api/sandbox_functions/compat";
 import {
   computeStaleSiblings,
-  diffManifestsAgainstSiblings,
+  diffStateAgainstSiblings,
 } from "@app/lib/api/sandbox_functions/compat";
 import type {
-  FunctionStateManifest,
-  ManifestColumn,
-  ManifestIndex,
-  ManifestTable,
+  DatabaseColumn,
+  DatabaseIndex,
+  DatabaseTable,
+  FunctionState,
 } from "@app/lib/api/sandbox_functions/manifests";
 import { describe, expect, it } from "vitest";
 
-function col(overrides: Partial<ManifestColumn> = {}): ManifestColumn {
+function col(overrides: Partial<DatabaseColumn> = {}): DatabaseColumn {
   return {
     type: "text",
     mode: null,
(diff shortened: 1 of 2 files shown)


id13:

- #28611
    - #28612
        - #28613
            - #28614
                - #28599
                    - #28718
                        - #28615
                            - #28616 ◀
                                - #28600
                                    - #28618
                                        - #28719
                                            - #28720
                                                - #28601

id13:
Folded into the restructured stack: the front pipeline now lands in its final shape in #28600 (the stored-schema mirror, manifests column and compat differ never exist in the new history), on top of the dsbx stack #28611→#28613→#28614→#28599→#28718. See #28600.

PR #28622, opened 2026-07-07
Match eager tool promotion on configuration id instead of server view id

flvndvd:
## Description

Follow up of https://github.com/dust-tt/dust/pull/28588.

The recent change that eagerly loads a custom agent's configured tools matched tools to their configured actions through the server view id. That key is not unique: several configurations of the same internal server share one view, and runtime-built conversation servers reuse those views too. In practice, an agent configured with table query tools saw the conversation-level table query tools, which appear as soon as a queryable file lands in the conversation, wrongly promoted to the eagerly loaded set mid-conversation, rewriting the cached tool prefix and defeating the purpose of the deferred catalog. The promotion now matches on the persisted id of the configuration, which runtime-built servers never carry, so only the tools the builder actually configured stay eager. A regression test covers the shared-view scenario observed in production.

### DIFF
--- a/front/temporal/agent_loop/lib/run_model.test.ts
+++ b/front/temporal/agent_loop/lib/run_model.test.ts
@@ -73,10 +73,20 @@ function makeConversation(
 
 function makeServerSideToolConfiguration(
   name: string,
-  { mcpServerViewId, eager }: { mcpServerViewId: string; eager?: boolean }
+  {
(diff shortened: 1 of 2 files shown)


flvndvd:
PMRR.

aubin-tchoi (commented):
LGTM

PR #28624, opened 2026-07-07
[core] Remove version_hash/latest_hash document filtering & index

id13:
## Description

`data_source_documents` have roughly 2x more data in its index than its heap, which is a bit inefficient, but also makes updates and insert very costly i/o wise (and we have a nefarious update pattern that is beyond the scope of this PR, one day maybe).

`idx_data_sources_documents_data_source_document_id_hash` itself is roughly 66% of the table, and is currently not being used (in the last 10 days) 

So this index drop should be completely invisible (along with the codepaths that use it), and reduce the cost of updating this table (it's the most updated table on core, by a big margin).

But given how ancient this code is, I lack the context about why it was introduced in the first place, and whether it's useful.
Ran into @adrsimon who worked on qdrant recently, didn't used, and @aubin-tchoi couldn't recall a use for it recently.

## Tests

Existing tests.

## Risk

Low

## Deploy Plan

- deploy core
- run migration

### DIFF
--- a/core/src/stores/postgres.rs
+++ b/core/src/stores/postgres.rs
@@ -1423,7 +1423,6 @@ impl Store for PostgresStore {
         project: &Project,
         data_source_id: &str,
         document_id: &str,
-        version_hash: &Option<String>,
     ) -> Result<Option<Document>> {
         let project_id = project.project_id();
         let data_source_id = data_source_id.to_string();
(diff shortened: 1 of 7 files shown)


core/src/stores/postgres.rs
```diff
@@ -1445,34 +1444,18 @@ impl Store for PostgresStore {
...
-                c.query(
-                    "SELECT dsd.id, dsd.created, dsd.timestamp, dsn.tags_array, dsn.parents, \
-                       dsn.source_url, dsd.hash, dsd.text_size, dsd.chunk_count, dsn.title, \
-                       dsn.mime_type, dsn.provider_visibility \
-                       FROM data_sources_documents dsd \
-                       INNER JOIN data_sources_nodes dsn ON dsn.document=dsd.id \
-                       WHERE dsd.data_source = $1 AND dsd.document_id = $2 \
-                       AND dsd.hash = $3 LIMIT 1",
-                    &[&data_source_row_id, &document_id, &version_hash],
-                )
```
id13, line 1472: Looked up in front, this is a dead codepath, so removing

id13: But please keep me honest

flvndvd (approved):

id13:
So after discussing with @flvndvd it was used for doc tracker, which is retired

aubin-tchoi (approved):
LGTM

PR #28627, opened 2026-07-07
fix: surface an agent error for unroutable models instead of hanging

adrsimon:
## Description

Fixes https://github.com/dust-tt/tasks/issues/9277

Setting an agent's model (via `PATCH /api/v1/w/{wId}/assistant/agent_configurations/{sId}`) to one that is listed in `SUPPORTED_MODEL_CONFIGS` but that no LLM client can actually serve returned HTTP 200, but starting a conversation hung forever on "Thinking…".

Root cause: `runModel` treated `llm === null` as "should not happen", logged, and `return null` **without** emitting a terminal `agent_error` / `agent_message_success` event. The FE then waited on pubsub indefinitely.

- `run_model.ts`: publish a `model_not_available` agent error when `llm` is null, so the message resolves and the UI shows an actionable error. This also rescues models already stuck in this state (e.g. Haiku 3, `sonnet-3.5-0620`).
- `create_or_upgrade.ts` / `conversation.ts`: reject an unknown `(modelId, providerId)` pair at save/post time with a clear error, so bad pairs never reach the runtime.
- `gemini-2.5-flash`: drop the incorrect `isLatest: true` flag, newer gemini 3s supersede it.

No model configs are removed: unroutable-but-listed models stay in `SUPPORTED_MODEL_CONFIGS` so existing conversations keep resolving their metadata.

## Tests

Manual: reproduced the hang by patching an agent onto an unroutable model. Typecheck + biome clean.

## Risk

Low.

## Deploy Plan

Standard front + agent-loop worker deploy.

### DIFF
--- a/front/lib/api/assistant/configuration/create_or_upgrade.ts
+++ b/front/lib/api/assistant/configuration/create_or_upgrade.ts
@@ -149,18 +149,25 @@ export async function createOrUpgradeAgentConfiguration({
 
   const featureFlags = await getFeatureFlags(auth);
   const modelConfig = getSupportedModelConfig(assistant.model);
-  if (modelConfig) {
-    const accessError = await getAdvancedModelAccessErrorForAgentConfiguration(
-      auth,
-      {
-        agentName: assistant.name,
-        model: modelConfig,
-        featureFlags,
-      }
+  if (!modelConfig) {
+    return new Err(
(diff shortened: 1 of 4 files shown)


aubin-tchoi (approved):
LGTM

PR #28628, opened 2026-07-07
Show checkout steps

tdraier:
## Description

Improves the checkout waiting experience. Previously the whole activation showed a single static "Processing payment…" spinner for ~30s, so users had no sense of progress and often thought it had stalled.

Screenshot 2026-07-07 at 17 23 03

**Multi-step progress indicator.** The `confirming` / `waiting_for_payment` phases now render a 3-step checklist that reflects the real backend phases, each advancing with a spinner → checkmark:
1. **Setting up your subscription** — the `POST /business-activation` (Metronome customer + contract + commit creation).
2. **Processing payment** — waiting for the Stripe charge + Metronome `payment_gate.payment_status: paid` webhook.
3. **Activating your workspace** — the (workflow-driven) plan swap, seat update, and seat sync.

**Real "activating" signal via polling.** Added an optional `progress: "activating"` field to the polled checkout-payment Redis record. The Metronome webhook endpoint marks it the moment a `payment_gate.payment_status: paid` event is received — the earliest reliable "payment cleared" point — so the UI advances to step 3 instead of showing "Processing payment" for the whole post-payment window (which is dominated by deferred Temporal activation work).

**Non-blocking receipt URL.** The Stripe hosted invoice ("View receipt") URL is a slow (~3s) round-trip. The status poll no longer fetches it, so the success screen appears immediately on `succeeded`; the receipt URL is fetched lazily afterwards (`?receipt=true`) and the button appears when ready.

## Tests

- Manual: ran a full checkout and inspected server logs to confirm the phase timings and that the stepper advances (confirming → processing payment → activating your workspace → success).
- `front/lib/credits/checkout_payment_status.test.ts` passes (covers the Redis record incl. the new optional `progress` field).
- `tsgo` clean in `front` and `front-api`.

## Risk

Low.
- The `progress` field is additive and optional on a private, onboarding-only endpoint (`GetBusinessActivationResponseBody`); older records without it are handled.
- Marking `activating` in the webhook endpoint is idempotent and a no-op unless a checkout is pending on that contract; it does not replace the workflow that performs (and completes) the real activation.
- Frontend uses `assertNeverAndIgnore` on the phase switch, so unknown values won't crash.
- Safe to roll back — no schema/data migration.

## Deploy Plan

Standard. `front` (webhook endpoint + status lib) and `front-api` (route) deploy together; no migration, no worker/queue changes.

### DIFF
--- a/front-api/routes/metronome/webhook.ts
+++ b/front-api/routes/metronome/webhook.ts
@@ -1,4 +1,5 @@
 import apiConfig from "@app/lib/api/config";
+import { markCheckoutPaymentActivating } from "@app/lib/credits/checkout_payment_status";
 import { unwrapMetronomeWebhook } from "@app/lib/metronome/client";
 import {
   getCustomerIdFromEvent,
@@ -109,6 +110,21 @@ app.post("/", async (ctx): HandlerResult<ResponseBody> => {
 
   if (!workspace) {
     return ctx.json({ success: true });
+  }
+
+  // Payment has cleared. The activation itself is deferred to the Temporal
+  // workflow below and can take several seconds (workflow scheduling + contract
+  // swap, seat sync, ...), during which the checkout polling UI would otherwise
+  // still show "processing payment". Advance it to "activating" here, at the
+  // earliest reliable point. No-op unless a checkout is pending on this contract.
+  if (
+    event.type === "payment_gate.payment_status" &&
+    event.properties.payment_status === "paid"
+  ) {
+    await markCheckoutPaymentActivating({
+      workspaceId: workspace.sId,
+      contractId: event.properties.contract_id,
+    });
   }
 
   // Hand the event off to a Temporal workflow for durable processing.
--- a/front-api/routes/metronome/webhook.ts
+++ b/front-api/routes/metronome/webhook.ts
@@ -1,5 +1,4 @@
 import apiConfig from "@app/lib/api/config";
-import { markCheckoutPaymentActivating } from "@app/lib/credits/checkout_payment_status";
 import { unwrapMetronomeWebhook } from "@app/lib/metronome/client";
 import {
   getCustomerIdFromEvent,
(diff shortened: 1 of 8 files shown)


tdraier:

- #28628 ◀

front-api/routes/metronome/webhook.ts
```diff
@@ -111,6 +112,21 @@ app.post("/", async (ctx): HandlerResult<ResponseBody> => {
...
+  // Payment has cleared. The activation itself is deferred to the Temporal
+  // workflow below and can take several seconds (workflow scheduling + contract
+  // swap, seat sync, ...), during which the checkout polling UI would otherwise
+  // still show "processing payment". Advance it to "activating" here, at the
+  // earliest reliable point. No-op unless a checkout is pending on this contract.
+  if (
+    event.type === "payment_gate.payment_status" &&
+    event.properties.payment_status === "paid"
+  ) {
+    await markCheckoutPaymentActivating({
```
thervier: feel a bit weird to have this alone here and not in the workflow (we could wonder if there are others elements in the workflow that should happen synchronously before it for similar reasons

front/lib/api/checkout/business_activation.ts
thervier, line 290: specific corner case: here, in the case where the payment if 0 (thanks to coupons), this is called directly and the webhook is not triggered, only handleSubscriptionActivationSuccess directly.

In this case do the implemented UI changes work as expected

front-api/routes/metronome/webhook.ts
tdraier: yes we could have other stuff handled synchronously indeed.
We could also keep  it in handleSubscriptionActivationSuccess but move it much earlier in the method

front/lib/api/checkout/business_activation.ts
tdraier: probably not 😬

front-api/routes/metronome/webhook.ts
tdraier: moved it back to handleSubscriptionActivationSuccess

thervier (commented):
I have tested for both the normal path and the one with 0 amount => they work :) 

2 remarks for the "zero amount activation" (with a coupon)
* we jump directly from the first point loading to the success screen (we dont we steps 2 and 3)
* we get a warning (that was there already): 
```
WARN (93535): [Business Activation] Failed to fetch Stripe hosted invoice URL {"metronomeCustomerId":"25ab8030-170c-49f5-a1c6-ebe499a98bf6","metronomeInvoiceId":"free-activation"}
    err: {
      "type": "NotFoundError",
      "message": "404 The specified invoice was not found",
```

tdraier:
> I have tested for both the normal path and the one with 0 amount => they work :)
> 
> 2 remarks for the "zero amount activation" (with a coupon)
> 
> * we jump directly from the first point loading to the success screen (we dont we steps 2 and 3)
> * we get a warning (that was there already):
> 
> ```
> WARN (93535): [Business Activation] Failed to fetch Stripe hosted invoice URL {"metronomeCustomerId":"25ab8030-170c-49f5-a1c6-ebe499a98bf6","metronomeInvoiceId":"free-activation"}
>     err: {
>       "type": "NotFoundError",
>       "message": "404 The specified invoice was not found",
> ```

both should be fixed now

thervier (approved):
LGTM
(and tested it in all the different possible cases)

PR #28631, opened 2026-07-07
Tool context: carry the MCP action resource in the agent loop run context

spolu:
## Description

Step towards conversation-free tool execution (sandbox function invocations).

- `AgentLoopRunContextType.currentAction: AgentMCPActionType` becomes `action: AgentMCPActionResource`: the run context carries the action resource itself instead of its serialized form.
- `SandboxFunctionRunContextType` gains `action: SandboxFunctionMCPActionResource`, so `runContext.action` resolves the right resource on both context types.
- `processToolNotification` and `processToolResults` no longer take an `action` parameter; they resolve it from the tool context. Agent-loop-only behaviors (per-block output items, step-context resume state) assert an agent loop run context; sandbox output persistence (`writeOutput`) is left as a `TODO(SANDBOX_FUNCTIONS)` for the sandbox function runner.
- `ToolNotificationEvent`: `action` moves from the shared base onto each variant — `AgentMCPActionWithOutputType` on the agent loop event, `SandboxFunctionMCPActionType` on the sandbox function event.
- The sandbox bash tool (sole reader of `currentAction`) now destructures the resource and passes `sandboxAction.toJSON()` to `generateSandboxExecToken`, whose signature is unchanged.

## Tests

`lib/actions/mcp_execution.test.ts`, `lib/actions/mcp_actions.test.ts`, `lib/api/actions/servers/sandbox/tools/index.test.ts`, `lib/api/sandbox/access_tokens.test.ts`, `lib/resources/sandbox_function_mcp_action_resource.test.ts` pass.

## Risk

Low. Agent loop behavior unchanged: the same resource flows through, just via the context. The sandbox function paths assert until the sandbox runner lands.

## Deploy Plan

- deploy `front`

### DIFF
--- a/front/lib/actions/mcp_execution.ts
+++ b/front/lib/actions/mcp_execution.ts
@@ -105,12 +105,12 @@ export async function processToolNotification(
 
   // Specific handling for run_agent notifications indicating the tool has
   // started and can be resumed: the action is updated to save the resumeState.
-  if (isRunAgentQueryProgressOutput(output)) {
-    // Sandbox function actions carry no step context to persist a resume state on.
-    assert(
-      isAgentLoopRunContext(runContext),
-      "run_agent notifications require an agent loop run context."
-    );
+  // Sandbox function actions carry no step context to persist a resume state on, so this only
+  // applies to agent loop run contexts.
+  if (
+    isRunAgentQueryProgressOutput(output) &&
+    isAgentLoopRunContext(runContext)
+  ) {
     await runContext.action.updateStepContext({
       ...runContext.action.stepContext,
       resumeState: {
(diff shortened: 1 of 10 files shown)


front/lib/actions/mcp_execution.ts
```diff
@@ -176,6 +172,7 @@ export async function processToolResults(
   const { runContext } = toolContext;
   assert(runContext, "processToolResults requires a tool run context.");
   const { toolConfiguration } = runContext;
+  const action = getMCPActionFromRunContext(runContext);
```
spolu: @PopDaph if you look at the full code here you'll see that that action is used for `action.createOutputItems(...)`

We'll need to have the equivalent on SandboxFunctionMCPActionResource or a shared static that work on both 👍

spolu: This thing calls `AgentMCPActionOutputItemModel.bulkCreate` so I guess we'll have an equivalent for it

PopDaph (approved):

PR #28634, opened 2026-07-07
fix(connectors): bound Google Drive file downloads to prevent OOM

davidebbo:
## Problem

The Google Drive connector worker was OOM-restarting. Investigation of a killed pod showed multiple 300–524 MB presentations being **fully buffered in memory** before being rejected, several at once. Two independent gaps let this happen, both rooted in the fact that **Google-native files (Docs/Slides) report no `size` in their Drive metadata**:

1. **`getAdaptiveConcurrency` gave unknown-size files maximum concurrency.** It filtered out files with no `size` before computing the largest file in a batch. A page of only Google-native docs therefore yielded `maxFoundFileSize = 0` → `Math.floor(maxFileSize / 0) = Infinity` → the default concurrency (8). So up to 8 large exports were held in memory simultaneously per activity.

2. **The pre-download size guard never fires for native docs.** `sync_one_file.ts` skips files over `MAX_FILE_SIZE_TO_DOWNLOAD` only `if (file.size && ...)`, which is always false for Docs/Slides. They went straight to `drive.files.export`, whose entire response was buffered in memory with only a reactive "too large to export" catch.

## Changes

- **`getAdaptiveConcurrency`** (`lib/async_utils.ts`): treat unknown-size files as worst-case (`maxFileSize`) so batches containing them drop to low concurrency instead of getting the default. Files with a known size *over* the limit are still ignored (they're skipped before download). This also fixes the same latent bug in the Microsoft connector, which shares this helper.

- **Cap the download/export itself** (`handle_file_export.ts`, `handle_google_drive_export.ts`): pass `maxContentLength: MAX_FILE_SIZE_TO_DOWNLOAD` to both `drive.files.get({alt:"media"})` and `drive.files.export(...)`. This maps to node-fetch's `size` limit, which **aborts the stream mid-download** — an oversized file never fully lands in memory. Matches the existing `maxContentLength` pattern in `dust_project`.

- **`isFileTooLargeToDownloadError`** (`temporal/utils.ts`): a type-safe helper that detects the resulting `max-size` error (and the older `ERR_OUT_OF_RANGE` case), so oversized files are **skipped gracefully** instead of throwing and triggering ~20 Temporal retries that each re-download. It also replaces a non-type-safe `e as { code: string }` cast (rule GEN4).

## Testing

- New unit tests, written test-first: `async_utils.test.ts` (6) and `utils.test.ts` (3).
- `incremental_sync.test.ts` (6) re-run to confirm no regression.
- **15/15 passing**, `tsgo --noEmit` clean.

### DIFF
--- a/connectors/src/connectors/google_drive/temporal/utils.ts
+++ b/connectors/src/connectors/google_drive/temporal/utils.ts
@@ -348,6 +348,27 @@ export async function _getLabels(
   return [];
 }
 
+// Detects the errors raised when a file exceeds the download size cap. Google-native docs
+// (Slides/Docs) report no size in their metadata, so the pre-download guard cannot catch them; we
+// cap the download itself via gaxios `maxContentLength`, which makes node-fetch abort mid-stream
+// with a `max-size` error. `ERR_OUT_OF_RANGE` is the older reactive case where Node fails to
+// allocate a buffer/string for a huge response.
+export function isFileTooLargeToDownloadError(error: unknown): boolean {
+  if (!(error instanceof Error)) {
+    return false;
+  }
+
+  if ("type" in error && error.type === "max-size") {
+    return true;
+  }
+
+  if ("code" in error && error.code === "ERR_OUT_OF_RANGE") {
+    return true;
+  }
+
+  return false;
+}
+
 export function isSharedDriveNotFoundError(error: unknown): boolean {
   if (!(error instanceof GaxiosError)) {
     return false;
--- a/connectors/src/connectors/google_drive/temporal/utils.ts
+++ b/connectors/src/connectors/google_drive/temporal/utils.ts
@@ -348,22 +348,37 @@ export async function _getLabels(
   return [];
 }
 
-// Detects the errors raised when a file exceeds the download size cap. Google-native docs
-// (Slides/Docs) report no size in their metadata, so the pre-download guard cannot catch them; we
-// cap the download itself via gaxios `maxContentLength`, which makes node-fetch abort mid-stream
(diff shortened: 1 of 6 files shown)


connectors/src/connectors/google_drive/temporal/utils.ts
```diff
@@ -348,6 +348,27 @@ export async function _getLabels(
...
+// (Slides/Docs) report no size in their metadata, so the pre-download guard cannot catch them; we
+// cap the download itself via gaxios `maxContentLength`, which makes node-fetch abort mid-stream
+// with a `max-size` error. `ERR_OUT_OF_RANGE` is the older reactive case where Node fails to
+// allocate a buffer/string for a huge response.
+export function isFileTooLargeToDownloadError(error: unknown): boolean {
+  if (!(error instanceof Error)) {
+    return false;
+  }
+
+  if ("type" in error && error.type === "max-size") {
```
philipperolet, line 357: could be worth checking gaxios, I think a maxContentLength overflow might be thrown as a GaxiosError with the original FetchError stored on error.error, so the top-level type is undefined

so the new checks might not catch the capped-download failure, and oversized native docs will still throw and trigger Temporal retries instead of being skipped

philipperolet (commented):
Thanks a lot 🙏 Great fixing this
Left a com worth checking

connectors/src/connectors/google_drive/temporal/utils.ts
davidebbo: Good point! I changed it so it looks for those two things both in error and error.error.

philipperolet (commented):
LGTM

PR #28639, opened 2026-07-07
feat(skill_authoring): add pagination to list_skills tool

frankaloia:
## Summary

- This list_skills tool was failing in practice because the content was super bloated. Should just be a lightweight pagination and rely on the get_skill tool to get rest of details
- Adding ability to filter by writable, agent_discoverable. Previously it was hardcoded to included only writable skills, which is not helpful for all cases

## Testing

* Validated locally

## Risk

Low — additive change, existing callers that pass no args get the same first-page behavior as before.

## Deploy

1. deploy front

### DIFF
--- a/front/lib/api/actions/servers/skill_authoring/metadata.ts
+++ b/front/lib/api/actions/servers/skill_authoring/metadata.ts
@@ -13,8 +13,22 @@ export const UPDATE_SKILL_TOOL_NAME = "update_skill" as const;
 export const SKILL_AUTHORING_TOOLS_METADATA = createToolsRecord({
   [LIST_SKILLS_TOOL_NAME]: {
     description:
-      "List active custom Skills in this workspace. Returns lightweight summaries so you can find the skill id to inspect or update.",
-    schema: {},
+      "List active custom Skills in this workspace. Returns lightweight summaries (sId, name, agentFacingDescription) paginated. Pass the returned `nextCursor` to fetch the next page.",
+    schema: {
+      cursor: z
+        .string()
+        .optional()
+        .describe(
+          "Pagination cursor from a previous call. Omit for the first page."
+        ),
+      limit: z
+        .number()
+        .int()
+        .positive()
+        .max(50)
+        .optional()
+        .describe("Skills per page. Default 20, max 50."),
+    },
     stake: "never_ask",
     displayLabels: {
       running: "Listing skills",
--- a/front/lib/api/actions/servers/skill_authoring/metadata.ts
+++ b/front/lib/api/actions/servers/skill_authoring/metadata.ts
@@ -13,13 +13,13 @@ export const UPDATE_SKILL_TOOL_NAME = "update_skill" as const;
 export const SKILL_AUTHORING_TOOLS_METADATA = createToolsRecord({
   [LIST_SKILLS_TOOL_NAME]: {
     description:
-      "List active custom Skills in this workspace. Returns lightweight summaries (sId, name, agentFacingDescription, canWrite) paginated",
+      "List active custom Skills in this workspace. Returns lightweight summaries (sId, name, agentFacingDescription, canWrite) paginated.",
     schema: {
       filter: z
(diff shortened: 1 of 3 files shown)


smb2268 (approved):
lgtm

front/lib/api/actions/servers/skill_authoring/tools/index.ts
```diff
@@ -218,35 +218,54 @@ function findDisallowedSpecialTagChanges(
...
-        name: skill.name,
-        agentFacingDescription: skill.agentFacingDescription,
-        userFacingDescription: skill.userFacingDescription,
-        icon: skill.icon,
-      }));
+    let skills;
+    if (resolvedFilter === "agent_discoverable") {
+      skills = await SkillResource.listDiscoverable(auth);
+    } else {
+      const allSkills = await SkillResource.listByWorkspace(auth, {
```
philipperolet, line 233: nit: maybe worth enforcing an order here? this way is backend order is not guaranteed or shifts, pagination stays the same?

front/lib/api/actions/servers/skill_authoring/metadata.ts
```diff
@@ -13,8 +13,28 @@ export const UPDATE_SKILL_TOOL_NAME = "update_skill" as const;
...
+      "List active custom Skills in this workspace. Returns lightweight summaries (sId, name, agentFacingDescription, canWrite) paginated.",
+    schema: {
+      filter: z
+        .enum(["all", "writable", "agent_discoverable"])
+        .optional()
+        .describe(
+          "Scope of skills to return. `writable` (default): only skills the user can edit. `all`: every skill in the workspace. `agent_discoverable`: only skills the agent can invoke in this conversation."
+        ),
+      cursor: z
+        .string()
```
philipperolet, line 25: ooc why not an int directly?

front/lib/api/actions/servers/skill_authoring/tools/index.ts
```diff
@@ -218,35 +218,54 @@ function findDisallowedSpecialTagChanges(
...
+      });
+      skills =
+        resolvedFilter === "writable"
+          ? allSkills.filter((skill) => skill.canWrite(auth))
+          : allSkills;
+    }
+
+    const pageSize = Math.min(limit ?? 20, 50);
+    const offset = cursor
+      ? parseInt(Buffer.from(cursor, "base64").toString(), 10)
```
philipperolet, line 247: should we handle irrelevant cursors here if we keep using a string? make a tool error so the agent corrects itself?
if we move to int, then positive is enforced at schema level; and we could also make a tool error if cursor is over total skill length, but that's less critical

front/lib/api/actions/servers/skill_authoring/metadata.ts
```diff
@@ -13,8 +13,28 @@ export const UPDATE_SKILL_TOOL_NAME = "update_skill" as const;
...
     description:
-      "List active custom Skills in this workspace. Returns lightweight summaries so you can find the skill id to inspect or update.",
-    schema: {},
+      "List active custom Skills in this workspace. Returns lightweight summaries (sId, name, agentFacingDescription, canWrite) paginated.",
+    schema: {
+      filter: z
+        .enum(["all", "writable", "agent_discoverable"])
+        .optional()
+        .describe(
+          "Scope of skills to return. `writable` (default): only skills the user can edit. `all`: every skill in the workspace. `agent_discoverable`: only skills the agent can invoke in this conversation."
```
philipperolet, line 22: here agent_discoverable is not actually conversation-scoped. the handler calls listDiscoverable(auth) without agentLoopData, while baseFetch only applies isDisabledForAgentLoop when that context exists. we say “only skills the agent can invoke in this conversation”, but this can include skills disabled for the current loop and miss agent/conversation-enabled non-default skills. 

we can either pass the run context through (likely best option), or change the filter/docs to “workspace default discoverable skills”.

philipperolet (commented):
🙏

PR #28649, opened 2026-07-07
[front] fix: split write and admin access on skills

aubin-tchoi:
## Description

Following up on the [following thread](https://dust4ai.slack.com/archives/C0B2L7PT56C/p1783337776146159).

This PR separates the concerns of admin rights and edit writes, currently merged into a single `canWrite` check.
Admins have the right to add themselves as editors (gated behind `canAdministrate`), but lose direct write access.
This does not change the behavior in Skill Builder, which additionally implemented this behavior client-side and forced users to add themselves as editors before making any change.

> [!NOTE]
>  This does change the behavior for the skill authoring tools, which now prevents admins to make change to any skill on the workspace.
> These tools intentionally do not allow editing editors lists, which remains a manual, very intentional action.

## Tests

- Tested locally.

## Risk

- Medium as it touches permissions, but technically should not get any user to get stuck as admins can always add themselves as editors.

## Deploy Plan

- Deploy front.

### DIFF
--- a/front/lib/resources/group_resource.ts
+++ b/front/lib/resources/group_resource.ts
@@ -2357,8 +2357,8 @@ export class GroupResource extends BaseResource<GroupModel> {
     return auth.canWrite(this.requestedPermissions());
   }
 
-  canAdmin(auth: Authenticator): boolean {
-    return auth.canAdmin(this.requestedPermissions());
+  canAdministrate(auth: Authenticator): boolean {
+    return auth.canAdministrate(this.requestedPermissions());
   }
 
   isSystem(): boolean {
--- a/front/lib/resources/group_resource.ts
+++ b/front/lib/resources/group_resource.ts
@@ -2283,9 +2283,11 @@ export class GroupResource extends BaseResource<GroupModel> {
    * 2. Role-based: Workspace admins get read and write access
    *
    * For agent_editors and skill_editors groups, the permissions are:
-   * 1. Group-based: The group's members get read and write access
-   * 2. Role-based: Workspace admins get read and write access. All users can
+   * 1. Group-based: The group's members get full access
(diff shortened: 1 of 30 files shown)


front/lib/resources/group_resource.ts
```diff
@@ -2301,11 +2303,11 @@ export class GroupResource extends BaseResource<GroupModel> {
           groups: [
             {
               id: this.id,
-              permissions: ["read", "write"],
+              permissions: ["read", "write", "admin"],
```
aubin-tchoi, line 2306: main change here, most of the other changes are call site changes from `canWrite` -> `canAdministrate` if needed

front/lib/resources/skill/skill_resource.ts
```diff
@@ -1800,6 +1800,19 @@ export class SkillResource extends BaseResource<SkillConfigurationModel> {
     return this.editorGroup.canWrite(auth);
   }
 
+  canAdministrate(auth: Authenticator): boolean {
+    // API keys with at least builder role can administrate any skill.
```
aubin-tchoi, line 1804: this is to avoid regressions, clearly not ideal

ykmsd (approved):

philipperolet (commented):
Thanks! However I think there's an issue: admins don't get write permissions anymore, but editing editor groups (agent/skill) is gated behind canWrite, e.g. in assistant/agent_configurations/[aId]/editors.ts
So admins cannot become editors anymore (despite the description); IMO that needs to be fixed in a follow up

philipperolet (commented):
Also @aubin-tchoi more minor: I'm told the swagger schema needs updating (currently documents canWrite but not canAdministrate in /api/v1/w/{wId}/skills return type

PR #28549, opened 2026-07-07
feat(sandbox): pod network settings UI [dedicated-file variant]

smb2268:
> **Superseded** by the owner-keyed egress policy re-layout stack (`egress-relayout-*` branches, PRs `28652`–`28655`). Branch kept for reference.

## Pod egress allowlist — settings UI

**DEDICATED-FILE variant.** Byte-identical to the projection variant's UI (#28515): the API contract is unchanged, so the Network section, settings-tab wiring, and SWR hooks carry over untouched.

### Stack (DEDICATED-FILE variant, GCS-only)

Alternative implementation of pod-level egress allowlists that mirrors the workspace egress policy mechanism **exactly** — GCS-only, the policy file is the store, no DB record. Built for side-by-side comparison with the PROJECTION stack (#28352 → #28512 → #28513 → #28514 → #28515). **Do not merge both.**

1. #28544 — egress-proxy: pod space policy file (deploys first)
2. #28650 — backend: GCS trio + spaceId JWT claim
3. #28547 — admin API (route mirrors the workspace route; contract identical to projection variant)
4. #28548 — audit (emit in route, like the workspace route)
5. #28549 — settings UI (byte-identical to projection variant)

_(#28545, a dedicated DB table, was dropped in favor of exact workspace mimicry; #28546 was auto-closed with it and replaced by #28650.)_

---

## Comparison: DEDICATED-FILE (GCS-only) vs PROJECTION

| | PROJECTION (#28352–#28515) | DEDICATED-FILE (this stack) |
|---|---|---|
| Proxy (Rust) changes | **None** | spaceId claim + `pods/` lookup + `p:` cache key + invalidation case (~120 lines + tests) |
| Deploy sequencing | None (front-only) | **Two-phase: proxy ships first**, then front |
| Storage | DB column on `ProjectMetadata` (source of truth) + per-sandbox GCS projection | **GCS file only** (`pods/{spaceSId}.json`) — exact workspace-policy architecture |
| DB migration | Required | **None** |
| Front diff size | model + resource + lib + adapter + route (~600 lines) | egress_policy trio + JWT + route (~350 lines) |
| Activation work | Re-project at every fresh create **and wake** | **None** |
| `add_egress_domain` shared-file collision | Latent (projection replaces the file the tool appends to) | **None** — separate files |
| Works with no sandbox ever created | DB yes; GCS file deferred to first activation | Yes, immediately |
[...]

PR #28548, opened 2026-07-07
feat(sandbox): pod egress policy audit [dedicated-file variant]

smb2268:
> **Superseded** by the owner-keyed egress policy re-layout stack (`egress-relayout-*` branches, PRs `28652`–`28655`). Branch kept for reference.

## Pod egress allowlist — audit logging

**DEDICATED-FILE variant.** Same `pod_egress_policy.updated` event, schema (incl. `actor_type`), and metadata as the projection variant (#28514). The emit sits in the route handler after the successful write — exactly where the workspace route puts it.

> [!NOTE]
> The task spec referred to this event as `pod_sandbox.allowlist.updated`; the actual action (both variants) is `pod_egress_policy.updated`.

### Stack (DEDICATED-FILE variant, GCS-only)

Alternative implementation of pod-level egress allowlists that mirrors the workspace egress policy mechanism **exactly** — GCS-only, the policy file is the store, no DB record. Built for side-by-side comparison with the PROJECTION stack (#28352 → #28512 → #28513 → #28514 → #28515). **Do not merge both.**

1. #28544 — egress-proxy: pod space policy file (deploys first)
2. #28650 — backend: GCS trio + spaceId JWT claim
3. #28547 — admin API (route mirrors the workspace route; contract identical to projection variant)
4. #28548 — audit (emit in route, like the workspace route)
5. #28549 — settings UI (byte-identical to projection variant)

_(#28545, a dedicated DB table, was dropped in favor of exact workspace mimicry; #28546 was auto-closed with it and replaced by #28650.)_

### Deploy plan

- [ ] Register the schema with WorkOS before deploy: `npx tsx front/admin/register_audit_log_schemas.ts --execute --changed` (not needed if the projection variant's registration already ran — same schema)

PR #28547, opened 2026-07-07
feat(sandbox): pod egress policy API [dedicated-file variant]

smb2268:
> **Superseded** by the owner-keyed egress policy re-layout stack (`egress-relayout-*` branches, PRs `28652`–`28655`). Branch kept for reference.

## Pod egress allowlist — admin API

**DEDICATED-FILE variant.** `GET`/`PUT` `/api/w/:wId/spaces/:spaceId/sandbox/egress-policy`, mirroring the workspace egress-policy route verbatim: handlers read/write the GCS file directly (no intermediate lib layer), 500 on storage errors, parse-level 400 on invalid domains. On top: the pod gates (workspace-admin + `sandbox_functions` FF via the sandbox sub-app, Pod-only spaces).

**Contract is byte-identical to the projection variant (#28513)** — the settings UI carries over with zero changes.

Route tests mirror the workspace route's test approach (full module mock; storage behavior covered by unit tests in #28650), including the 500-on-read-failure case the projection variant couldn't have (its GET read the DB).

### Stack (DEDICATED-FILE variant, GCS-only)

Alternative implementation of pod-level egress allowlists that mirrors the workspace egress policy mechanism **exactly** — GCS-only, the policy file is the store, no DB record. Built for side-by-side comparison with the PROJECTION stack (#28352 → #28512 → #28513 → #28514 → #28515). **Do not merge both.**

1. #28544 — egress-proxy: pod space policy file (deploys first)
2. #28650 — backend: GCS trio + spaceId JWT claim
3. #28547 — admin API (route mirrors the workspace route; contract identical to projection variant)
4. #28548 — audit (emit in route, like the workspace route)
5. #28549 — settings UI (byte-identical to projection variant)

_(#28545, a dedicated DB table, was dropped in favor of exact workspace mimicry; #28546 was auto-closed with it and replaced by #28650.)_

PR #28546, opened 2026-07-07
feat(sandbox): pod egress policy backend [dedicated-file variant]

smb2268:
## Pod egress allowlist — backend (space-file render)

**DEDICATED-FILE variant.** The render/enforcement layer:

- `writePodSpacePolicy` / `deletePodSpacePolicy` render the DB record to `pods/{spaceSId}.json` (wildcards, 100-domain cap) and invalidate the proxy cache by spaceId.
- `mintEgressJwt` gains an optional `spaceId` claim, set for pod sandboxes from the runtime owner in `setupEgressForwarder`; `mintEgressInvalidationJwt` gains the matching key. Back-compatible with proxy builds that predate the claim.
- `setPodEgressDomains`: DB write, then render. A failed render **surfaces an error** (the proxy reads the file; silent success would hide stale enforcement) — unlike the projection variant's best-effort write, whose per-activation re-projection makes silent failure self-healing.
- `addPodEgressDomain` (manifest-approve path): idempotent append with **no shared-file collision** with the `add_egress_domain` agent tool — the pod policy no longer lives in `sandboxes/{providerId}.json`.
- Activation: **no per-activation projection and no wake rewrite** (the space-keyed file survives sandbox lifecycles). Fresh creation runs a best-effort self-heal render, covering lost bucket objects and workspace relocation (the projection variant's wake-time rewrite disappears entirely).
- `hardDeleteSpace` deletes the policy row (FK RESTRICT) and best-effort scrubs the file — note the workspace policy file has no such scrub today; this variant does not copy that gap.

### Stack (DEDICATED-FILE variant)

Alternative implementation of pod-level egress allowlists, built for side-by-side comparison with the PROJECTION stack (#28352 → #28512 → #28513 → #28514 → #28515). **Do not merge both.**

1. #28544 — egress-proxy: pod space policy file (deploys first)
2. #28545 — data layer (dedicated table)
3. #28546 — backend (space-file render + spaceId JWT claim)
4. #28547 — admin API (contract identical to projection variant)
5. #28548 — audit (emit on the canonical DB write)
6. #28549 — settings UI (byte-identical to projection variant)

### Risk

Worst case, the pod file is missing/stale → fail-closed to the workspace allowlist. Requires #28544 deployed first.

PR #28650, opened 2026-07-07
feat(sandbox): pod egress policy backend [dedicated-file variant]

smb2268:
> **Superseded** by the owner-keyed egress policy re-layout stack (`egress-relayout-*` branches, PRs `28652`–`28655`). Branch kept for reference.

## Pod egress allowlist — backend (GCS-only, workspace mirror)

**DEDICATED-FILE variant.** The pod section in `egress_policy.ts` is a verbatim mirror of the workspace trio, keyed by the pod's space sId:

- `readPodSpacePolicy` / `writePodSpacePolicy` / `deletePodSpacePolicy` on `pods/{spaceSId}.json` — same normalization (wildcards supported), same missing-file-is-empty semantics, same shapes (`write(key, { policy })`), cache invalidation keyed by spaceId. Like `writeWorkspacePolicy`, there is **no domain-count cap** (the projection variant's 100-cap protected the shared per-sandbox file; that constraint doesn't exist here).
- `mintEgressJwt` gains an optional `spaceId` claim (set for pod sandboxes via the runtime owner in `setupEgressForwarder`); `mintEgressInvalidationJwt` gains the matching invalidation key. Back-compatible with older proxy builds.

The space-keyed file survives sandbox destroy/recreate and sleep/wake: **no per-activation projection, no wake rewrite, zero sandbox-adapter changes** — total front diff is 3 files.

> [!NOTE]
> Mirroring the workspace policy exactly means inheriting its gaps knowingly: no file scrub on space deletion (the workspace file has none on workspace deletion either) and no relocation handling. Stale files are inert — space sIds are never reused and no sandbox carries the claim after deletion.

### Stack (DEDICATED-FILE variant, GCS-only)

Alternative implementation of pod-level egress allowlists that mirrors the workspace egress policy mechanism **exactly** — GCS-only, the policy file is the store, no DB record. Built for side-by-side comparison with the PROJECTION stack (#28352 → #28512 → #28513 → #28514 → #28515). **Do not merge both.**

1. #28544 — egress-proxy: pod space policy file (deploys first)
2. #28650 — backend: GCS trio + spaceId JWT claim
3. #28547 — admin API (route mirrors the workspace route; contract identical to projection variant)
4. #28548 — audit (emit in route, like the workspace route)
5. #28549 — settings UI (byte-identical to projection variant)

_(#28545, a dedicated DB table, was dropped in favor of exact workspace mimicry; #28546 was auto-closed with it and replaced by #28650.)_

### Risk

Worst case, the pod file is missing → fail-closed to the workspace allowlist. Requires #28544 deployed first.

### DIFF
--- a/front/lib/api/sandbox/egress.ts
+++ b/front/lib/api/sandbox/egress.ts
@@ -106,13 +106,22 @@ async function runSuccessfulRootCommand(
   return new Ok(undefined);
 }
 
-export function mintEgressJwt(providerId: string, workspaceId: string): string {
+// spaceId is set for pod (Shared Computer) sandboxes so the proxy evaluates
+// the pod's space policy file (`pods/{spaceId}.json`) in addition to the
+// workspace and sandbox policies. Optional and back-compatible: proxy builds
+// that predate the claim ignore it.
+export function mintEgressJwt(
+  providerId: string,
+  workspaceId: string,
+  { spaceId }: { spaceId?: string } = {}
+): string {
   return jwt.sign(
     {
       iss: "dust-front",
       aud: "dust-egress-proxy",
       sbId: providerId,
       wId: workspaceId,
+      ...(spaceId ? { spaceId } : {}),
     },
     config.getEgressProxyJwtSecret(),
     {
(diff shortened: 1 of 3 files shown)


smb2268:

- #28544
    - #28650 ◀
        - #28547
            - #28548
                - #28549

PR #28651, opened 2026-07-07
[front] move logger for unsupported file type 

ykmsd:
## Description
This PR is to change the logger position I added in [this PR](https://github.com/dust-tt/dust/pull/28561). Inside the `processToolResults`, we first run `persistToolOutput` and this offload text block (except "conversation_files",  "files", "interactive_content", "sandbox" tools, iiuc they are capped somewhere) so we don't need to log them. I moved it to after `(res.value !== null) ` check, null equals persistToolOutput doesn't do anything (which I find it a bit weird but ok)

### DIFF
--- a/front/lib/actions/mcp_execution.ts
+++ b/front/lib/actions/mcp_execution.ts
@@ -414,6 +414,19 @@ export async function processToolResults(
                 file: null,
               };
             }
+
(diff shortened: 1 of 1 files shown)


flvndvd (commented):
LGTM

PR #28652, opened 2026-07-07
feat(egress-proxy): workspace-prefixed, owner-keyed policy layout (dual-read)

smb2268:
## Egress policy re-layout — proxy dual-read

First phase of moving the egress policy storage to a workspace-prefixed, **owner-keyed** layout:

```
w/{wId}/sandbox-egress-policy.json           workspace-wide policy (was workspaces/{wId}.json)
w/{wId}/sandboxes/{ownerId}.json owner policy (was sandboxes/{providerId}.json)
                                 ownerId = conversation sId | pod space sId
```

**Why:** keying the per-sandbox policy by its *owner* instead of the ephemeral provider id makes it stable across sandbox destroy/recreate cycles — agent-approved domains persist for the conversation, and pods get a policy file with **no third storage location** (their spaceId is just another ownerId). The `w/{wId}/` prefix makes workspace relocation and scrubbing prefix operations.

**This PR is the dual-read phase** so the proxy can deploy before front changes anything:

- Optional `ownerId` JWT claim; `sbId` stays the required identity claim (and stays in logs). Old tokens keep validating.
- Workspace lookup prefers the new path and **falls back** to the legacy path when absent (dated TODO).
- Evaluation **unions** workspace + owner + legacy per-sandbox policies, all fail-closed — in-flight approvals in legacy sandbox files keep working until those sandboxes age out (dated TODO).
- `/invalidate-policy` accepts: `wId` (evicts both workspace layouts), `wId`+`ownerId` (owner entry), `sbId` (legacy entry).

### Stack / rollout

1. #28652 (this PR) — proxy dual-read. **Deploys and soaks first.**
2. #28653 — front cutover (new paths, ownerId claim, conversation-lifetime approvals)
3. #28654 — backfill script: copy legacy workspace files, skip-if-exists; runs after front deploys
4. #28655 — pod egress-policy route + UI on the owner file
5. (follow-up, dated) cleanup: drop fallbacks/unions/legacy objects

### Risk

Worst case, a lookup fails closed to a stricter policy. New paths are unread until front mints the claim.

> [!WARNING]
> **Rollback is only clean before #28653 deploys.** Once front writes owner files, rolling this proxy back makes those approvals invisible while the tool's "already allowed" pre-check still sees them — a deny loop re-approval cannot fix. Roll front back first, then the proxy.

### Tests

[...]

PR #28653, opened 2026-07-07
feat(sandbox): cut front over to owner-keyed egress policy layout

smb2268:
## Egress policy re-layout — front cutover

Second phase (proxy dual-read is #28652 — **must deploy first**). Front now writes the owner-keyed layout:

- **Workspace policy** → `w/{wId}/sandbox-egress-policy.json`. Reads fall back to the legacy path when the new object is absent (covers un-backfilled workspaces); writes go to the new path and **dual-write the legacy path** so a front rollback keeps seeing — and safely editing — the same policy. Dated TODOs on both.
- **Owner policy** replaces the per-providerId sandbox policy: `readOwnerPolicy` / `addOwnerPolicyDomain` / `deleteOwnerPolicy` on `w/{wId}/sandboxes/{ownerId}.json`.
- **Behavior change (intended):** agent-approved domains now persist for the **conversation's lifetime**, across sandbox destroy/recreate cycles — previously they silently died with the sandbox. The `add_egress_domain` tool copy is updated to say so.
- `mintEgressJwt` carries `ownerId` (conversation sId or pod space sId, from the runtime owner); `sbId` stays for identity/log tracing. Owner cache invalidation sends `workspaceId` + `ownerId`.
- **Lifecycle:** owner files are deleted with their owner — conversation destruction scrubs the file (pod space deletion follows in the pod PR). Sandbox destroy now only cleans up legacy per-providerId files, with no invalidation (the sandbox and its token are gone; cache entries age out).
- The existing `sandbox_egress_policy.sandbox_updated` audit event gains a `conversation_id` metadata field: the provider id it logged no longer identifies the approval's durable scope. Schema version bumped — **re-register with WorkOS before deploy** (`npx tsx front/admin/register_audit_log_schemas.ts --execute --changed`). No new audit actions in this stack.
- The tool's "already allowed?" pre-check keeps exact-string membership, with a comment documenting the assumption (this function is the only writer of conversation owner files and only appends exact domains); it must become wildcard-aware before the tool is ever enabled on pod sandboxes, whose files are admin-managed.

### Stack / rollout

See #28652. After this deploys: run the backfill script (follow-up PR) to copy legacy workspace files; the read fallback covers the gap until then.

### Risk

Worst case, an owner file write fails → fail-closed to the workspace policy, same as today. Requires #28652 deployed and healthy first — on an old proxy, the ownerId claim is ignored and new-path files are unread, degrading to "conversation/pod grants not applied" (never an over-grant). Front rollback is covered by the legacy dual-write. **Proxy rollback stops being clean once this deploys** (owner-file approvals go invisible to an old proxy while the tool's pre-check still sees them) — if both must roll back, front first.

### Tests

Workspace: new-layout read, legacy fallback, new-preferred, dual-write (incl. legacy-write-failure tolerated), delete-both. Owner: read/append/dedupe/cap/wildcard-reject, workspace+owner invalidation, delete, legacy-file deletion without invalidation. Tool: owner-keyed call expectations. front-api workspace route suite unchanged and green.

### DIFF
--- a/front/lib/api/assistant/conversation/destroy.ts
+++ b/front/lib/api/assistant/conversation/destroy.ts
@@ -1,4 +1,5 @@
 import { hardDeleteDataSource } from "@app/lib/api/data_sources";
+import { deleteOwnerPolicy } from "@app/lib/api/sandbox/egress_policy";
 import type { Authenticator } from "@app/lib/auth";
 import { AgentSuggestionModel } from "@app/lib/models/agent/agent_suggestion";
 import {
@@ -28,6 +29,7 @@ import {
   ProjectTaskSourceModel,
 } from "@app/lib/resources/storage/models/project_task";
 import { WakeUpResource } from "@app/lib/resources/wakeup_resource";
+import logger from "@app/logger/logger";
 import type { ConversationWithoutContentType } from "@app/types/assistant/conversation";
 import type { ModelId } from "@app/types/shared/model_id";
 import type { Result } from "@app/types/shared/result";
@@ -184,6 +186,20 @@ export async function destroyConversation(
   }
 ): Promise<Result<void, Error>> {
   const owner = auth.getNonNullableWorkspace();
+
+  // Best-effort scrub of the conversation's sandbox egress allowlist file
+  // (owner-keyed, so it is not deleted with individual sandboxes). A failure
+  // logs but does not block the destroy.
+  const deleteOwnerPolicyRes = await deleteOwnerPolicy(auth, conversation.sId);
+  if (deleteOwnerPolicyRes.isErr()) {
+    logger.warn(
+      {
+        err: deleteOwnerPolicyRes.error,
+        conversationId: conversation.sId,
+      },
+      "Failed to delete conversation egress policy file on destroy."
+    );
+  }
 
   await ConversationForkResource.deleteForConversationModelId(auth, {
     conversationModelId: conversation.id,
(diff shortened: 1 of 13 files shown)


smb2268:

- #28784
    - #28653 ◀
        - #28654
            - #28655

front/lib/api/assistant/conversation/destroy.ts
```diff
@@ -185,6 +187,24 @@ export async function destroyConversation(
 ): Promise<Result<void, Error>> {
   const owner = auth.getNonNullableWorkspace();
 
+  // Best-effort scrub of the conversation's sandbox egress allowlist file
+  // (owner-keyed, so it is not deleted with individual sandboxes).
+  void deleteOwnerPolicy(auth, conversation.sId)
```
flvndvd: Why not awaiting here?

front/lib/api/sandbox/egress.ts
```diff
@@ -106,13 +106,24 @@ async function runSuccessfulRootCommand(
   return new Ok(undefined);
 }
 
-export function mintEgressJwt(providerId: string, workspaceId: string): string {
+// ownerId is the sandbox owner's stable sId (conversation sId for
+// conversation sandboxes, space sId for pod sandboxes); it selects the owner
+// policy file `w/{wId}/sandboxes/{ownerId}.json`. sbId stays for identity and
+// log tracing. Older proxy builds ignore the ownerId claim.
+export function mintEgressJwt(
```
flvndvd: Can we avoid the flat attributes here and use an object so we never get confused with the keys?

flvndvd (approved):

PR #28661, opened 2026-07-08
[front] enh: auto-enable Computer skill for dust-like agents

aubin-tchoi:
## Description

This PR auto-enables the Computer skill for dust-like agents.
The rationale is that dust-like agents are the primary users of this tool, and it is used in a large variety of use cases.
Deferring the bash tool is thus not desirable, it increases significantly the number of tool searches required for situations where it is very clear that the tool is required (ideally tool search is only used for tools that are used in specific situations, not ones used very commonly). However, since the bash tool is hidden behind a skill, not deferring it would cause the skill activation to burst the cache, hence making it auto-enabled.

Will wait for https://github.com/dust-tt/dust/pull/28572 to merge as the current behavior for auto-enabled may be bugged.

## Tests

- Tested locally.

## Risk

- Low.

## Deploy Plan

- Deploy front.


flvndvd (approved):
LGTM

PR #28662, opened 2026-07-08
Load the files edit tool eagerly instead of behind tool search

flvndvd:
## Description

Editing a file is a common case, so it is worth keeping the files edit tool in the model's cached prefix instead of hiding it behind tool search, the same treatment already given to listing and reading files. This also gives it a stable spot in the model's toolset instead of leaving it to tool search ranking, which can be sensitive to wording changes elsewhere in the tool corpus.

This is required to align frames on the regular file handling operations.


flvndvd:
PMRR.

aubin-tchoi (commented):
LGTM

PR #28435, opened 2026-07-03
Introduce SandboxFunctionRunContextType, minimal fix of servers/tools

spolu:
## Description

- Introduces `SandboxFunctionRunContextType` 
  - Fixes servers that are easy to fix
  - Asserts for servers that are agent run loop dependant
  - Asserts and leaves TODO for the ones that need more involved fixing

Files of particular interest:
- `front/lib/actions/mcp_actions.ts`
- `front/lib/actions/types/index.ts`
- `front/lib/api/mcp/run_tool.ts`
- `x/spolu/tool_context/WORK.md` for the blueprint and follow-ups

First high-level review is eased by collapsing on the left menu the `front/lib/api/actions/servers` directory.

r? @PopDaph I am so sorry. Really. Very hard to not do in one go :(
cc @flvndvd interactive_content TODO left
cc @adrsimon for visibility given you reviewed the previous one

No agent was involved in this one.

## Tests

Covered by TS and tests.

- tested locally that tool calling is operational 
- tsgo on ALL `front` also passing locally

## Risk

Low. Can be reverted at any time

## Deploy Plan

- deploy `front`

PR #28664, opened 2026-07-08
Sandbox functions: MCP tool execution workflow

PopDaph:
## Description

Sandbox functions (code running via `dsbx function run` on a pod) will be able to call MCP tools without a conversation. The model landed in #28446 and the resource in #28604. This PR adds the execution layer: given a `SandboxFunctionMCPAction` row, run the tool on Temporal and persist its output. The endpoints that create these rows come in the next PR (stacked).

### Output persistence in `processToolResults`

Fills the `TODO(SANDBOX_FUNCTIONS)` left by #28631: `processToolResults` now branches on the run context for persistence.

| run context | persistence |
|---|---|
| `agent_loop` | one output item row per content block (unchanged) |
| `sandbox_function` | full content array as a single GCS object, `outputGcsPath` on the action row |

A single object because the consumer (dsbx) reads the output exactly once, as a whole array, at completion. There are no per-block consumers (rendering, citations, per-block file FKs and retention are conversation concepts). Everything upstream of persistence was already context-aware (#28435): generated files land as `project_context` on the pod's space, large
text offloads to the pod file system.

### Temporal execution

`runSandboxFunctionToolWorkflow` → `runSandboxFunctionToolActivity` on `agent-loop-queue-v2` (same queue as `runSandboxChildToolWorkflow`), `maximumAttempts: 1` (tool calls are not idempotent). The activity drives `tryCallMCPTool` with the `sandbox_function` run context, persists through `processToolResults` (error content included — the poll endpoint returns it alongside the errored status), then marks the action succeeded/errored. Heartbeats come from
`tryCallMCPTool` during the call and `withPeriodicHeartbeat` during result processing, mirroring `run_tool`.

`makeToolNotificationEvent` becomes optional in `tryCallMCPTool`: there is no conversation surface to stream progress notifications to.

## Tests

- `mcp_execution.test.ts`: sandbox run context writes the full content array to a single GCS
  object, sets `outputGcsPath`, returns no output items.
- Existing `processToolResults` suite passes unchanged (agent-loop path untouched).

## Risk

`launchSandboxFunctionToolWorkflow` has no caller in this PR: the `/sandbox/actions/call`  endpoint that invokes it is the next PR of the stack. So dead code until the follow-up PR lands (no caller). Worst case, revert. Safe to rollback.

## Deploy Plan

Deploy front.

### DIFF
--- a/front/lib/actions/mcp_actions.ts
+++ b/front/lib/actions/mcp_actions.ts
@@ -352,11 +352,7 @@ export async function* tryCallMCPTool(
     signal,
   }: {
     progressToken: ModelId;
-    // Optional: callers without a conversation surface (sandbox function invocations) skip it,
-    // and progress notifications are consumed but not yielded. Without a callback,
-    // `store_resource` notification contents are NOT persisted (persistence happens in the
-    // callback via `processToolNotification`). Acceptable while only `run_agent` emits them.
-    makeToolNotificationEvent?: (
+    makeToolNotificationEvent: (
       notification: MCPProgressNotificationType
     ) => Promise<ToolNotificationEvent>;
     signal?: AbortSignal;
@@ -597,9 +593,7 @@ export async function* tryCallMCPTool(
           break;
         }
         notificationPromise = notificationStream.next();
-        if (makeToolNotificationEvent) {
-          yield makeToolNotificationEvent(iteratorResult.value);
-        }
+        yield makeToolNotificationEvent(iteratorResult.value);
       }
     }
 
--- a/front/lib/actions/mcp_actions.ts
+++ b/front/lib/actions/mcp_actions.ts
@@ -353,7 +353,11 @@ export async function* tryCallMCPTool(
   }: {
     progressToken: ModelId;
     // Optional: callers without a conversation surface (sandbox function invocations) have no
-    // notification event to build — progress notifications are consumed but not yielded.
+    // notification event to build — progress notifications are consumed but not yielded. Note
+    // that this drops more than UI events: `store_resource` notification contents are persisted
+    // by the caller-provided callback (via `processToolNotification`), so without a callback they
+    // are NOT persisted. Acceptable while no tool callable in these modes emits them (only
+    // `run_agent` does today).
     makeToolNotificationEvent?: (
       notification: MCPProgressNotificationType
     ) => Promise<ToolNotificationEvent>;
--- a/front/lib/actions/mcp_actions.ts
+++ b/front/lib/actions/mcp_actions.ts
@@ -352,12 +352,10 @@ export async function* tryCallMCPTool(
     signal,
   }: {
     progressToken: ModelId;
-    // Optional: callers without a conversation surface (sandbox function invocations) have no
-    // notification event to build — progress notifications are consumed but not yielded. Note
-    // that this drops more than UI events: `store_resource` notification contents are persisted
-    // by the caller-provided callback (via `processToolNotification`), so without a callback they
-    // are NOT persisted. Acceptable while no tool callable in these modes emits them (only
-    // `run_agent` does today).
+    // Optional: callers without a conversation surface (sandbox function invocations) skip it,
+    // and progress notifications are consumed but not yielded. Without a callback,
+    // `store_resource` notification contents are NOT persisted (persistence happens in the
+    // callback via `processToolNotification`). Acceptable while only `run_agent` emits them.
     makeToolNotificationEvent?: (
       notification: MCPProgressNotificationType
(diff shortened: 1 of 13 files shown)


front/temporal/agent_loop/activities/run_sandbox_function_tool.ts
```diff
@@ -0,0 +1,121 @@
...
+    invocation,
+    toolConfiguration: action.toolConfiguration,
+  };
+
+  const startTime = performance.now();
+
+  // Without `makeToolNotificationEvent` the generator yields nothing (there is no conversation
+  // surface to stream progress notifications to); it only carries the tool result, which is its
+  // return value, hence the drain.
+  const toolCallResult = await drainAsyncGenerator(
```
PopDaph, line 72: Since we wanted the switch inside `tryCallMCPTool`, this is a bit cracra 😬

front/lib/actions/mcp_actions.ts
```diff
@@ -352,7 +352,11 @@ export async function* tryCallMCPTool(
     signal,
   }: {
     progressToken: ModelId;
-    makeToolNotificationEvent: (
+    // Optional: callers without a conversation surface (sandbox function invocations) skip it,
+    // and progress notifications are consumed but not yielded. Without a callback,
+    // `store_resource` notification contents are NOT persisted (persistence happens in the
+    // callback via `processToolNotification`). Acceptable while only `run_agent` emits them.
+    makeToolNotificationEvent?: (
```
spolu: Why skip it? Any cost in yielding it? We might leverage it eventually in the viz wrapper to show progress?

front/lib/actions/mcp_execution.ts
```diff
@@ -488,7 +476,33 @@ export async function processToolResults(
     })
   );
 
-  return { outputItems, generatedFiles };
+  switch (runContext.contextType) {
```
spolu: I would favor my refactor in https://github.com/dust-tt/dust/pull/28665 which aligns a bit more the two resources approaches. Should be an easy rebase.

front/lib/resources/sandbox_function_resource.ts
```diff
@@ -264,6 +264,41 @@ export class SandboxFunctionResource extends BaseResource<SandboxFunctionModel>
     return sandboxFunction ?? null;
   }
 
+  // Lives here rather than on SandboxFunctionInvocationResource: the invocation resource is
+  // constructed with its function, and importing this resource from there would close an import
+  // cycle (this file already imports the invocation resource for deletion).
+  static async fetchInvocationByModelIdWithAuth(
```
spolu: Can we have it on `SandboxFunctionMCPActionResource` so that we don't have to pass a ModelID around? Sould we actually load `SandboxFunctionInvocationResource` automatically on all `SandboxFunctionMCPActionResource` ?

front/temporal/agent_loop/activities/run_sandbox_function_tool.ts
spolu: Not sure I understand that comment? 😬

spolu: Ideally whatever event is emitted here should be posted to the invocation event stream no?

spolu (commented):
Left a few comments/questions.

Generally LGTM otherwise!

front/lib/actions/mcp_actions.ts
PopDaph: Fair!

front/lib/resources/sandbox_function_resource.ts
PopDaph: I went with fetchInvocationForAction(auth, action) on SandboxFunctionResource, it takes the action so no ModelId gets passed around, which was the main thing.

I couldn't put it directly on SandboxFunctionMCPActionResource: it would have to value-import the invocation resource to build one, but the invocation resource already value-imports the action resource for its RESTRICT cascade deletes, so action -> invocation -> action trips noImportCycles. Auto-loading on every fetch is blocked the same way, plus it's an extra query nothing needs today (only the activity fetches the action, and toJSON already derives invocationId without loading the invocation).

I can properly move it onto the action resource as a follow-up, it means pulling the cascade-delete out of the invocation resource so it stops importing the action. Lmk if you want that, otherwise this keeps the ModelId out of the call.

spolu (approved):
LGTM

front/lib/resources/sandbox_function_resource.ts
spolu: No ModelId. I'm happy! :)

PR #28665, opened 2026-07-08
Tool outputs: return the generic content shape from action output persistence

spolu:
## Description

Step towards conversation-free tool execution (sandbox function invocations): both action resources expose the same `createOutputItems(auth, contents: Array<{ content, fileId? }>): Promise<Result<ToolOutputItemType[], Error>>`, and `processToolResults` persists through a single polymorphic call on `runContext.action`.

- New `ToolOutputItemType` (`{ content, fileId, file, workspaceId }`): the generic persisted-output shape shared by both resources — the stored content plus the file linkage required to render the model-facing view (`hideFileFromActionOutput`).
- `AgentMCPActionResource.createOutputItems` persists as before (DB rows, GCS, Redis warm), returns the stored items in the generic shape, and returns a `Result` instead of throwing on GCS failure (resolves the standing TODO).
- `SandboxFunctionMCPActionResource.writeOutput` is renamed `createOutputItems` with the same signature, storing the raw content array.
- The resources stay storage-only: `hideFileFromActionOutput` (model-facing presentation) remains where it was, applied in `runToolWithStreaming` when building the `tool_success` output — `ToolOutputItemType` carries exactly what it needs.
- `processToolNotification`'s `store_resource` handling is polymorphic too (no more agent-loop assert), and `AgentMCPActionOutputItemModel` disappears from the tool execution pipeline.

## Tests

- `lib/actions/mcp_execution.test.ts`, `lib/actions/mcp_actions.test.ts`, `lib/actions/mcp_internal_actions/exit_events.test.ts`, `lib/resources/agent_mcp_action_resource.test.ts`, `lib/resources/sandbox_function_mcp_action_resource.test.ts`, `temporal/agent_loop/activities/common.test.ts` pass.
- Tested locally

## Risk

Low. Agent loop behavior unchanged (same hidden-content view flows into `tool_success` / `agent_action_success`, GCS failures still abort the step); the sandbox function path has no caller yet. Can be reverted.

## Deploy Plan

- deploy `front`

### DIFF
--- a/front/lib/actions/mcp_execution.test.ts
+++ b/front/lib/actions/mcp_execution.test.ts
@@ -265,7 +265,7 @@ describe("processToolResults", () => {
     });
 
     expect(outputItems).toHaveLength(1);
-    const stored = outputItems[0];
+    const stored = outputItems[0].content;
 
     // The large text block should be converted to a resource with a truncated snippet.
     expect(stored.type).toBe("resource");
@@ -298,7 +298,7 @@ describe("processToolResults", () => {
     });
 
     expect(outputItems).toHaveLength(1);
-    const stored = outputItems[0];
+    const stored = outputItems[0].content;
 
     expect(stored.type).toBe("resource");
     if (stored.type === "resource" && "text" in stored.resource) {
@@ -324,7 +324,7 @@ describe("processToolResults", () => {
     });
 
     expect(outputItems).toHaveLength(1);
-    const stored = outputItems[0];
+    const stored = outputItems[0].content;
 
     expect(stored.type).toBe("text");
     if (stored.type === "text") {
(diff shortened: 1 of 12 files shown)


spolu:
r? @PopDaph cc @flvndvd 

Note: @flvndvd we're still storing content in pg and the `createOutputItems` implementation still has a bunch of TODOs on that note.

PopDaph (approved):

flvndvd:
> r? @PopDaph cc @flvndvd
> 
> Note: @flvndvd we're still storing content in pg and the `createOutputItems` implementation still has a bunch of TODOs on that note.

Yes I've been running migrations in the past few week ends to backfill the missing rows, so we can drop it for good. Still need a few cycles as it's pretty slow.

PR #28666, opened 2026-07-08
Sandbox functions: list, call and poll MCP tools from invocations

PopDaph:
Follow up of #28664.
Code running in a sandbox function invocation can now list, call and poll MCP tools through the sandbox API using its invocation token, no conversation involved.

The three `/v1/w/:wId/sandbox/actions` endpoints previously hard-required a conversation-bound exec token. `sandboxAuth` now takes `allowedTokenKinds` and the handlers branch on the claims type (exec-token behavior unchanged; the computer feature-flag check still applies to exec tokens only).

| endpoint | invocation-token behavior |
|---|---|
| `GET /` | internal server views of the pod space + global space |
| `POST /call` | resolves the view within the pod + global space, creates a `SandboxFunctionMCPAction` (status `running`), launches its workflow, `202 {status: "pending", actionId}` |
| `GET /:aId` | scoped to the token's invocation; running → `202 pending`, succeeded/errored → `200` with the output array read from GCS, denied → `403 rejected` |

Same wire contract as the exec-token flavor: dsbx cannot tell which storage produced its answer.

`createSandboxFunctionMCPAction` gates at creation: the server view must be in the pod or global space and readable by the caller (same confinement as the listing: pod + global space selection plus a `canReadOrAdministrate` check, so a revoked pod access within the token TTL is honored); internal servers only (remote servers have their own auth story); the tool exists and is enabled; effective stake `never_ask`, since there is no approval surface without a conversation, so approval-requiring tools are rejected with a clear error (approval bubbling is a follow-up, gated on invocation durability). An out-of-scope view id reports as not-found so the sandbox cannot probe other spaces. The resolved tool configuration is snapshotted on the row (same derive-then-snapshot pattern as `create_child_action`), keeping activity re-runs deterministic. There is deliberately no per-server gating beyond that: tools that require an agent-loop context error at execution based on the run context.

Security-relevant change to call out: `restrictGroupsToSandboxFunctionInvocationSpaces` now also grants invocation tokens the global space. The auto internal server views (`common_utilities`, `search`, …) live there, so without it nothing is listable or callable. Still intersected with the token user's base groups.

Known edges, deferred: invocation tokens expire after 2 minutes (durability follow-up), and a token minted without a `uId` (non-human actor) gets minimal groups, so `/call` resolves nothing and 404s. File outputs from tools land as `project_context` files on the pod's space; surfacing them on the pod filesystem is a follow-up.

## Tests

Endpoint tests: listing (invocation vs exec branch), call (happy path, unknown view, view outside the pod + global spaces → 404, remote server, unknown tool, conversation-coupled server), poll (pending, success with GCS output, cross-invocation scoping 404).

Manual e2e on a local pod: a published function shelling out to `dsbx tools` listed 33 servers and called `common_utilities.generate_random_number` (result returned through call → workflow → GCS → poll, 392ms). Also driven raw with a minted invocation token and `curl` only, no agent, no conversation: `202 pending` → `200 success` with the output array inline.

## Risk

New token kind on an existing surface: exec-token requests go through the same middleware. Regression risk is covered by the existing exec-token tests; behavior for them is unchanged. Worst case, invocation-token tool calls fail. Safe to rollback.

## Deploy Plan

Deploy front.

### DIFF
--- a/front-api/routes/v1/w/[wId]/sandbox/actions/[aId]/index.ts
+++ b/front-api/routes/v1/w/[wId]/sandbox/actions/[aId]/index.ts
@@ -1,13 +1,19 @@
 import { isToolExecutionStatusFinal } from "@app/lib/actions/statuses";
 import { isSandboxChildActionInfo } from "@app/lib/actions/types";
-import { isSandboxExecTokenPayload } from "@app/lib/api/sandbox/access_tokens";
+import {
+  isSandboxExecTokenPayload,
+  isSandboxFunctionInvocationTokenPayload,
+} from "@app/lib/api/sandbox/access_tokens";
 import { AgentMCPActionResource } from "@app/lib/resources/agent_mcp_action_resource";
+import { SandboxFunctionMCPActionResource } from "@app/lib/resources/sandbox_function_mcp_action_resource";
 import type { AgentMCPActionWithOutputType } from "@app/types/actions";
+import type { SandboxFunctionMCPActionType } from "@app/types/api/sandbox_functions";
 import { assertNever } from "@app/types/shared/utils/assert_never";
 import { sandboxApp } from "@front-api/middlewares/ctx";
 import type { HandlerResult } from "@front-api/middlewares/utils";
 import { apiError } from "@front-api/middlewares/utils";
 import { validate } from "@front-api/middlewares/validator";
+import type { CallToolResult } from "@modelcontextprotocol/sdk/types.js";
 import { z } from "zod";
 
 const ParamsSchema = z.object({
@@ -28,8 +34,18 @@ type CallToolSuccessResponse = {
   action: AgentMCPActionWithOutputType;
 };
 
-export type FetchConversationMessageActionResponse =
+type CallToolSandboxFunctionSuccessResponse = {
+  status: "success";
+  action: SandboxFunctionMCPActionType & {
+    output: CallToolResult["content"] | null;
+  };
+};
+
+// Flavor-neutral on purpose: exec tokens poll AgentMCPActions (conversation sandboxes) and
+// invocation tokens poll SandboxFunctionMCPActions, through the same handler and wire contract.
+export type GetSandboxActionResponseType =
   | CallToolSuccessResponse
+  | CallToolSandboxFunctionSuccessResponse
   | CallToolPendingResponse
   | CallToolRejectedResponse;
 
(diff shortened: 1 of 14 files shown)


front-api/routes/v1/w/[wId]/sandbox/actions/call.ts
```diff
@@ -25,6 +30,65 @@ app.post(
...
+          default:
+            assertNever(result.error.type);
+        }
+      }
+
+      return ctx.json(
+        { status: "pending" as const, actionId: result.value.actionId },
+        202
+      );
+    }
```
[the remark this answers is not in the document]
Comment (PopDaph) [reply front-api/routes/v1/w/[wId]/sandbox/actions/call.ts:82]: @zeropath-ai the suggested fix would 403 all sandbox-function tool calls (removing the feature) so not applying it. 

The real issue seemed to be upstream: pod_manager's getPod resolved a caller-supplied dustPod without a canRead check, allowing cross-pod reads. Fixed by enforcing pod.canRead(auth) in getPod (also closes the same hole on the agent-loop path), with regression tests. Good for you?

front-api/routes/v1/w/[wId]/sandbox/actions/[aId]/index.ts
```diff
@@ -28,8 +34,16 @@ type CallToolSuccessResponse = {
...
 };
 
+type CallToolSandboxFunctionSuccessResponse = {
+  status: "success";
+  action: SandboxFunctionMCPActionType & {
+    output: CallToolResult["content"] | null;
+  };
+};
+
 export type FetchConversationMessageActionResponse =
```
Comment (spolu) [inline front-api/routes/v1/w/[wId]/sandbox/actions/[aId]/index.ts]: This one looks misnamed?

front-api/routes/v1/w/[wId]/sandbox/actions/index.ts
```diff
@@ -32,6 +66,30 @@ app.use("*", sandboxAuth({ tokenKind: "action" }));
...
+    // views not yet materialized in the global space stay invisible until another surface
+    // hydrates them.
+    const views = await MCPServerViewResource.listBySpaceIds(
+      auth,
+      [claims.spaceId],
+      { includeGlobalSpace: true }
+    );
+
+    const serverViews = views
+      .filter((view) => view.serverType === "internal")
```
Comment (spolu) [inline front-api/routes/v1/w/[wId]/sandbox/actions/index.ts]: Why this internal fitler?

front-api/routes/v1/w/[wId]/sandbox/actions/[aId]/index.ts
Comment (PopDaph) [reply front-api/routes/v1/w/[wId]/sandbox/actions/[aId]/index.ts]: Renamed to GetSandboxActionResponseType (matches GetSandboxToolsResponseType on the list route) with a comment on why it stays flavor-neutral rather than naming the function flavor.

front-api/routes/v1/w/[wId]/sandbox/actions/index.ts
Comment (PopDaph) [reply front-api/routes/v1/w/[wId]/sandbox/actions/index.ts]: I initially gated the first version on internal + never_ask tools so we'd only expose tools we knew worked, but flipped it the other way around: allow everything and work our way up to supporting them. 

The rule is now: a sandbox function can call any tool on any server view in its pod + global space, as long as the tool exists and isn't admin-disabled. Stakes are still snapshotted on the action so the approval bubbling follow-up can enforce them later.

front-api/routes/v1/w/[wId]/sandbox/actions/[aId]/index.ts
```diff
@@ -43,9 +59,60 @@ const app = sandboxApp();
...
   "/",
   validate("param", ParamsSchema),
-  async (ctx): HandlerResult<FetchConversationMessageActionResponse> => {
+  async (ctx): HandlerResult<GetSandboxActionResponseType> => {
     const auth = ctx.get("auth");
     const claims = ctx.get("sandboxClaims");
+
+    const { aId } = ctx.req.valid("param");
+
+    if (isSandboxFunctionInvocationTokenPayload(claims)) {
```
Comment (PopDaph) [inline front-api/routes/v1/w/[wId]/sandbox/actions/[aId]/index.ts:68]: Is this cracra and better to create a new route?

Comment (spolu) [reply front-api/routes/v1/w/[wId]/sandbox/actions/[aId]/index.ts:68]: I'm fine personally with one route. Maybe have a switch instead of the if would make it look a tiny bit better?

front-api/routes/v1/w/[wId]/sandbox/actions/index.ts
Comment (spolu) [reply front-api/routes/v1/w/[wId]/sandbox/actions/index.ts]: >  allow everything and work our way up to supporting them.

LGTM

front-api/routes/v1/w/[wId]/sandbox/actions/[aId]/index.ts
Comment (spolu) [reply front-api/routes/v1/w/[wId]/sandbox/actions/[aId]/index.ts]: 🙏

spolu (approved):
LGTM

PR #28592, opened 2026-07-07
Add: auto model and update agent builder picker

Fraggle:
## Description

Introduces `AUTO_MODEL_ID` / `AUTO_MODEL_CONFIG` (`front/types/assistant/models/auto.ts`) — a virtual model gated on `models_picker` FF that represents dynamic routing to the best model for the task. Registers `auto` across the type system: `STATIC_MODEL_IDS`, `MODEL_PROVIDER_IDS`, `SUPPORTED_MODEL_CONFIGS`, `USED_MODEL_CONFIGS`, token pricing (zero-cost), error labels, sandbox profile fallback, and the JS SDK's `KnownModelLLMId`.

In the agent builder's `AdvancedSettings`, adds a `SliderToggle` row for "Auto" at the top of the model dropdown. Selecting it stores the current `modelSettings` and switches to `AUTO_MODEL_CONFIG`; deselecting restores the previous selection. `ModelSelectionSubmenu` and `ReasoningEffortSubmenu` are hidden while Auto is active. Also fixes `ModelSelectionSubmenu` regional flag logic: removes the `SHOULD_DISPLAY_FLAG` region map so `RegionalFlag` shows in all regions when `use_vertex_for_supported_models` is on and not BYOK.

## Tests

Tested locally in the agent builder. No new automated tests added (UI-only change).

## Risk

Auto model is gated on `models_picker` FF — no impact for workspaces without the flag. The virtual model is excluded from `ModelProvidersPageContent` and priced at zero to avoid side effects. Low risk, straightforward rollback.

## Deploy Plan

Deploy front.

PR #28670, opened 2026-07-08
Add: requestXXX columns to user message, handle model resolution correctly.

Fraggle:
## Description

Adds `requestedProviderId`, `requestedModelId`, `requestedReasoningEffort` columns to `user_messages` so the raw model selection from the picker is persisted on the user message (`UserMessageType.requestedModel: ModelSelectionType | null`).

Alongside this, model resolution is refactored:
- `resolveModelSelection` → `resolveModel`, which now takes both the user's `selection` and the agent's `configuration`, resolving in priority order: user selection → agent config → workspace-supported fallback. It always returns a resolved model and a `modelResolutionMethod` ('user' | 'agent').
- Resolution is moved from `postUserMessage` (single, pre-mention) to `createAgentMessages` (per agent-mention), so each mention is resolved against its own agent's configured model.
- `AgentMessageType.requestedModel` (optional) is replaced by `resolvedModel: ResolvedRequestedModel | null` (required), reflecting the actual model used.
- The retry path now copies the original `AgentMessageModel` row columns verbatim instead of re-validating at retry time.
- `ModelSelectionSchema.modelId` tightened from `z.string()` to `z.enum(MODEL_IDS)`.

## Tests

- Extended `models.test.ts` with a full `resolveModel` suite: user selection honored, fallback to agent model, fallback to workspace large model, disabled selection fallthrough, reasoning effort pass-through and fallback.
- Existing `createUserMessage`, `createAgentMessages`, and MCP action tests updated to include the new required fields.
- Tested locally.

## Risk

Medium. This is a DB schema addition (nullable columns, no data backfill) on `user_messages` — safe to run while the app is live. The resolution logic change moves `resolveModel` to per-mention scope; behavior is equivalent for single-agent messages and improves correctness for multi-mention threads. Retry path now copies original row values instead of re-resolving, which removes an edge case where a model disabled post-send would silently change retry behavior.

## Deploy Plan

1. Run pre-deploy SQL migration (`20260708083845_add_columns_for_requested_models_on_user_message.sql`) on the front DB.
2. Deploy front.

### DIFF
--- a/front/components/assistant/conversation/lib.ts
+++ b/front/components/assistant/conversation/lib.ts
@@ -54,6 +54,7 @@ export function createPlaceholderUserMessage({
       origin: "web",
     },
     reactions: [],
+    requestedModel: null, // TODO(models_picker): add the requested model
     contentFragments: [
       ...(contentFragments?.uploaded ?? []).map(
         (cf) =>
--- a/front/lib/api/assistant/conversation.ts
+++ b/front/lib/api/assistant/conversation.ts
@@ -27,10 +27,7 @@ import {
   batchRenderMessages,
   batchRenderUserMessagesWithoutMentions,
 } from "@app/lib/api/assistant/messages";
-import {
-  isProviderWhitelisted,
-  resolveModelSelection,
-} from "@app/lib/api/assistant/models";
+import { isProviderWhitelisted } from "@app/lib/api/assistant/models";
 import { gracefullyStopAgentLoop } from "@app/lib/api/assistant/pubsub";
 import {
   MESSAGE_RATE_LIMIT_PER_ACTOR_PER_HOUR,
@@ -159,10 +156,7 @@ import {
   isUserMention,
   toMentionType,
 } from "@app/types/assistant/mentions";
-import type {
-  ModelSelectionType,
-  ResolvedRequestedModel,
-} from "@app/types/assistant/models/types";
+import type { ModelSelectionType } from "@app/types/assistant/models/types";
 import type {
   ContentFragmentContextType,
   ContentFragmentType,
@@ -596,14 +590,6 @@ export async function postUserMessage(
   }
 
   const featureFlags = await getFeatureFlags(auth);
-
-  // Resolve the picker selection to a concrete model for this workspace. If it
-  // cannot be honored (unknown/disabled model), we leave `resolvedModel` null
-  // and the agent's configured model is used.
-  const resolvedModel: ResolvedRequestedModel | null = modelSelection
-    ? resolveModelSelection(auth, modelSelection, { featureFlags })
-    : null;
-
   const isPartOfPod = isPodConversation(conversation);
 
   if (isPartOfPod) {
@@ -891,6 +877,7 @@ export async function postUserMessage(
                 ...context,
                 origin: "branch_anchor",
               },
+              requestedModel: modelSelection ?? null,
             },
             transaction: t,
           });
@@ -1001,6 +988,7 @@ export async function postUserMessage(
         context: enrichedContext,
         agenticMessageData,
         visibility,
+        requestedModel: modelSelection ?? null,
       },
       transaction: t,
     });
@@ -1045,7 +1033,6 @@ export async function postUserMessage(
             skipToolsValidation,
             nextMessageRank,
             userMessage: userMessageWithoutMentions,
-            resolvedModel,
           },
           transaction: t,
         });
@@ -1898,6 +1885,7 @@ export async function retryAgentMessage(
           type: "retry",
           parentId: messageRow.parentId,
           agentMessage: message,
+          agentMessageRow: messageRow.agentMessage,
         },
         transaction: t,
       });
--- a/front/lib/api/assistant/conversation/messages.ts
+++ b/front/lib/api/assistant/conversation/messages.ts
@@ -6,7 +6,7 @@ import {
 import { getCompletionDuration } from "@app/lib/api/assistant/messages";
 import {
   resolvedModelFromAgentMessageRow,
-  resolveModelSelection,
+  resolveModel,
 } from "@app/lib/api/assistant/models";
 import { type Authenticator, getFeatureFlags } from "@app/lib/auth";
 import {
@@ -39,7 +39,7 @@ import type {
 import { isPodConversation } from "@app/types/assistant/conversation";
 import type { MentionType } from "@app/types/assistant/mentions";
 import { isAgentMention } from "@app/types/assistant/mentions";
-import type { ResolvedRequestedModel } from "@app/types/assistant/models/types";
+import type { ModelSelectionType } from "@app/types/assistant/models/types";
 import type { ModelId } from "@app/types/shared/model_id";
 import { assertNever } from "@app/types/shared/utils/assert_never";
 import { removeNulls } from "@app/types/shared/utils/general";
@@ -97,6 +97,7 @@ export async function createUserMessage(
           context: UserMessageContext;
           agenticMessageData?: AgenticMessageData;
           visibility?: MessageVisibility;
+          requestedModel: ModelSelectionType | null;
         };
     transaction: Transaction;
   }
@@ -110,6 +111,7 @@ export async function createUserMessage(
   let context: UserMessageContext | null = null;
   let agenticMessageData: AgenticMessageData | undefined = undefined;
   let visibility: MessageVisibility = "visible";
+  let requestedModel: ModelSelectionType | null = null;
 
   switch (metadata.type) {
     case "edit":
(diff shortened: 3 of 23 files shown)


Fraggle:

- #28670 ◀
    - #28684
        - #28592

front/components/assistant/conversation/lib.ts
```diff
@@ -54,6 +54,7 @@ export function createPlaceholderUserMessage({
       origin: "web",
     },
     reactions: [],
+    requestedModel: null, // TODO(models_picker): add the requested model
```
Fraggle, line 57: this one is front end and related to the model picker, can you own it ?

front/lib/api/assistant/conversation/messages.ts
```diff
@@ -292,26 +296,20 @@ export const createAgentMessages = async (
     case "retry":
       {
         const agentConfiguration = metadata.agentMessage.configuration;
-        // Preserve the per-message model override, but re-validate it: a model
-        // enabled at first send may have since been disabled for the workspace,
-        // in which case the retry falls back to the agent's configured model.
```
Fraggle, line 297: Could be added back but in practice I don't think it's worth it.

front/lib/api/assistant/conversation/messages.ts
```diff
@@ -465,18 +464,26 @@ export const createAgentMessages = async (
               { transaction }
             );
 
+            const { resolvedModel, modelResolutionMethod } = resolveModel(
```
Fraggle, line 471: this is the key part, we always resolve and store on the agent message.

front/lib/api/assistant/conversation/messages.ts
```diff
@@ -465,18 +464,26 @@ export const createAgentMessages = async (
...
+              auth,
+              {
+                selection: metadata.userMessage.requestedModel ?? undefined,
```
Fraggle, line 474: resilient to pending message as it takes the selection from what was stored on the user message.

front/lib/api/assistant/conversation/messages.ts
```diff
@@ -199,6 +202,9 @@ export async function createUserMessage(
       userContextAuthMethod: context.authMethod ?? null,
       agenticMessageType,
       agenticOriginMessageId,
+      requestedModelId: requestedModel?.modelId ?? null,
+      requestedProviderId: requestedModel?.providerId ?? null,
+      requestedReasoningEffort: requestedModel?.reasoningEffort ?? null,
```
Fraggle, line 208: Here we store on the user message.

Nils-Fedrigo:
>The retry path now copies the original AgentMessageModel row columns verbatim instead of re-validating at retry time.

don't we want to revalidate?

Fraggle:
>> The retry path now copies the original AgentMessageModel row columns verbatim instead of re-validating at retry time.

> don't we want to revalidate?

I hesitated, I ultimately felt on this side because it would be a "silent" change in what we are actually retrying.

front/components/assistant/conversation/lib.ts
Nils-Fedrigo: just need to thread it right? No need for a display?

front/lib/api/assistant/conversation/messages.ts
Nils-Fedrigo: ok ok

front/components/assistant/conversation/lib.ts
Fraggle: It's used for displaying the message while we wait for the backend response. But it would also be useful for you as it would already support the selected model by the user (... to default to it for the next message).

front/lib/api/assistant/conversation/messages.ts
Nils-Fedrigo: worth making it an in-code comment?

front/lib/api/assistant/conversation.ts
```diff
@@ -596,14 +590,6 @@ export async function postUserMessage(
   }
 
   const featureFlags = await getFeatureFlags(auth);
-
-  // Resolve the picker selection to a concrete model for this workspace. If it
-  // cannot be honored (unknown/disabled model), we leave `resolvedModel` null
-  // and the agent's configured model is used.
-  const resolvedModel: ResolvedRequestedModel | null = modelSelection
-    ? resolveModelSelection(auth, modelSelection, { featureFlags })
```
Nils-Fedrigo, line 604: so we only guard in front as a result?

Nils-Fedrigo: ok my bad I understand now => we give the "raw" requestedModel to createUserMessage then guard against abuse there bc we want to store the raw in the user message

front/lib/api/assistant/models.ts
```diff
@@ -230,74 +236,115 @@ function toResolvedModel(
...
 }
 
-// Resolves a raw picker selection to a concrete model, or null when it cannot be
-// honored for this workspace (unknown or disabled model), in which case the
-// caller falls back to the agent's configured model.
-export function resolveModelSelection(
+// Resolves the model for an agent message according to these rules:
+// 1. Pick the user's selection
+// 2. If the user did not select a model, pick the agent's configured model
+// 3. Finally fallback to the a supported model by the workspace.
```
Nils-Fedrigo, line 242:
```suggestion
// 3. Finally fallback to a supported model by the workspace.
```

front/lib/api/assistant/models.ts
```diff
@@ -230,74 +236,115 @@ function toResolvedModel(
   };
 }
 
-// Resolves a raw picker selection to a concrete model, or null when it cannot be
-// honored for this workspace (unknown or disabled model), in which case the
-// caller falls back to the agent's configured model.
-export function resolveModelSelection(
+// Resolves the model for an agent message according to these rules:
+// 1. Pick the user's selection
```
Nils-Fedrigo, line 240:
```suggestion
// 1. Pick the user's selection if it's part of their allowlist
```

front/lib/api/assistant/models.ts
```diff
@@ -230,74 +236,115 @@ function toResolvedModel(
...
+}): m is ResolvedRequestedModel {
+  return (
+    m.providerId !== null &&
+    m.modelId !== null &&
+    m.reasoningEffort !== null &&
+    isModelProviderId(m.providerId) &&
+    isModelId(m.modelId) &&
+    isReasoningEffort(m.reasoningEffort as ReasoningEffort)
   );
-  if (config?.supportsResponseFormat) {
```
Nils-Fedrigo, line 273: was this lost in translation ?

Nils-Fedrigo: the idea that we check if the new model supports the response format and if not, drop it from the call

front/types/assistant/models/types.ts
```diff
@@ -23,7 +24,7 @@ export type CustomThinkingType = (typeof CUSTOM_THINKING_TYPES)[number];
 // provider/model pick, with an optional reasoning-effort override.
 export const ModelSelectionSchema = z.object({
   providerId: z.enum(MODEL_PROVIDER_IDS),
-  modelId: z.string(),
+  modelId: z.enum(MODEL_IDS),
```
Nils-Fedrigo, line 27: no idea of the implications

front/types/assistant/agent_run.ts
```diff
@@ -325,38 +324,40 @@ export async function getAgentLoopDataWithAuth(
...
   }
 
   return new Ok({
     agentConfiguration: agentConfigurationWithoutModel,
     model: {
-      ...agentModel,
-      ...model,
+      // This contains general configuration for the model, like the provider and model ID.
+      ...modelConfig,
+      // This contains specific configuration for reasoning effort, temperature, etc.
```
Nils-Fedrigo, line 355:
```suggestion
      // This contains specific configuration for model, reasoning effort.
```

front/types/assistant/models/types.ts
Fraggle: Ensure the type we receive in modelSelection is correct.

front/lib/api/assistant/models.ts
Fraggle: It was moved to the place where we store the model config for the execution loop.

Nils-Fedrigo: ok ok I see it below you're good

front/types/assistant/agent_run.ts
```diff
@@ -325,38 +324,40 @@ export async function getAgentLoopDataWithAuth(
...
+      ...resolvedModelConfig,
+      // Cleanup unsupported settings
+      ...(modelConfig?.supportsResponseFormat
+        ? { responseFormat: resolvedModelConfig.responseFormat }
```
Nils-Fedrigo, line 359: I find it a bit dégeu and prefer and little if above but ok

Nils-Fedrigo (approved):
clean, thanks

front/types/assistant/agent_run.ts
Fraggle: :-)
You made me laugh.

I figure that it was the right place since we were building the final shape and it was a one liner.
I wouldn't take it bad if you want to move it elsewhere (reconcileXXX + the templated type was not super clear to me).

PR #28671, opened 2026-07-08
[marketing] update team members

sfriquet:
## Description

Add members that joined since 2026-6-9
Remove departed members

## Tests

glanced at http://localhost:3009/home/about


adrsimon (approved):
LGTM!

PR #28673, opened 2026-07-08
Improve messages when changing seat billing period

tdraier:
## Description

1. ChangeSeatModal.tsx — isDeferredChange now also accounts for the monthly→yearly commitment case (mirroring the backend's isMonthlyToYearlySwitch in classifySeatChange), so the "The change will take effect at the next credit refresh." message now shows up for monthly→yearly switches too.
2. MembersUsageTable.tsx — the scheduled-change tooltip now detects the same-tier monthly→yearly case and shows "This user will switch to annual billing at the end of the billing period (date)" instead of the confusing "This user will be changed to Pro..." (since pro and pro_yearly share the same tier rank, it previously fell into the generic "changed" verb).

## Deploy Plan

deploy front


tdraier:

- #28673 ◀

fabiencelier (approved):

PR #28675, opened 2026-07-08
[dsbx] fix: address @dust/pod review feedback

id13:
Addresses @flvndvd's post-merge review comments on #28605:

- `#dbName`/`#maxSizeBytes` → `private readonly` fields, matching the codebase convention (native `#` private fields were used nowhere else in the repo).
- `override` keyword on the `Database` method overrides, with `noImplicitOverride` added to the package tsconfig so the compiler enforces it.
- The test now imports the package by name (`@dust/pod`) instead of `./index.ts`, backed by an `exports` entry in package.json (required for self-referencing imports, and pins the public entry point).

## Tests
- `bun test`: 22 pass, 0 fail.
- `bunx tsc --noEmit`: clean.


flvndvd (approved):
Thank you 🙏

PR #28677, opened 2026-07-08
feat(front): attach remote MCP servers via agent toolset API

adrsimon:
## Description

The `toolset` field of `PATCH`/`POST /api/v1/w/{wId}/assistant/agent_configurations` only resolved internal auto-available servers by name. Any other `mcp_server_name` was silently skipped, so remote tools could never be attached to an agent programmatically.

Non-auto names are now matched against the workspace's MCP server views by display name:
- system-space views are excluded (agents attach from global/regular spaces);
- when a server is shared to several spaces, the global-space view wins (same as the auto-internal path);
- an ambiguous name resolves to nothing and is reported in `skippedActions` rather than guessing.

Views are fetched once per request, so resolving many toolset entries doesn't fan out into an N+1. Name resolution lives on `MCPServerViewResource.resolveAttachableByName`. 

## Tests

`front/lib/api/assistant/configuration/yaml_import.test.ts` added cases for: remote server attached by name, remote server present only in the system space (skipped), and a name matching several servers (skipped as ambiguous). Existing skip-reason expectation updated.

## Risk

Safe to rollback.

## Deploy Plan

Standard front deploy.

### DIFF
--- a/front-api/public/swagger.json
+++ b/front-api/public/swagger.json
@@ -715,6 +715,7 @@
                   },
                   "toolset": {
                     "type": "array",
+                    "description": "Replaces the full set of tools on the agent. Any tool not present in this array is removed, so send the complete desired toolset. Each entry resolves an MCP server by name (see configuration.mcp_server_name). Entries that cannot be resolved are not applied and are returned in the skippedActions field of the response rather than causing the whole request to fail.",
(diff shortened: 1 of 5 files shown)


rfrenoy (approved):
lgtm

PR #28678, opened 2026-07-08
fix(front): let unstick plugin force-unlock multi-step blocked conversations

adrsimon:
## Description

The `Unstick Conversation` poke plugin fails with `plugin_execution_failed: All blocked actions must be from the same step, got 3, 1`.

**Root cause:** unstick calls `updateAgentMessageWithFinalStatus("failed")` → (status is unresumable) → `denyBlockedActionsForAgentMessage`, which fetches the message's blocked actions via `listBlockedActionsForAgentMessage` and inherits its `assert(uniqueSteps.length === 1)`. When a stuck message has blocked actions spanning several steps, the assert throws and rolls back the whole finalize transaction, so the conversation stays stuck.

The single-step invariant is sound, the agent loop should pause on the first blocking step, so a message should never accumulate blocked actions across steps. We want to keep enforcing it *by default* so violations surface as errors. But the unstick plugin is an explicit admin escape hatch: it needs to finalize a genuinely broken conversation even when the invariant is violated.

**What this PR does:**
1. Adds a `force` option to `updateAgentMessageWithFinalStatus`, threaded through to `denyBlockedActionsForAgentMessage`.
2. Default (`force: false`, every existing caller): the deny path enforces the single-step invariant and throws on a violation — unchanged behavior, bugs still surface loudly.
3. `force: true` (the unstick plugin only): skips the invariant check and denies blocked actions across all steps, and `logger.warn`s the anomaly so we get prod signal on how often it happens.

> [!NOTE]
> `force` only lets the admin tool finalize an already-broken conversation. It does **not** fix *why* an agent message ends up with blocked actions on multiple steps, that is a separate agent-loop bug. A follow-up PR will fix the origin of the broken single-step invariant; the warn log added here is the thread to pull on.

## Tests

Added to `blocked_actions.test.ts` (shared `setupMultiStepBlockedMessage` helper):
- `force: true` on a multi-step blocked message denies all actions and clears `actionRequired` (reproduces the unstick path).
- Without `force`, the same setup throws `All blocked actions must be from the same step` (invariant still enforced by default).

Verified the force test fails against the pre-fix code and both pass now. 12/12 in the file.

## Risk

Pure logic change, no schema. Safe to rollback.

## Deploy Plan

- [ ] Deploy front.

### DIFF
--- a/front/lib/resources/agent_mcp_action_resource.ts
+++ b/front/lib/resources/agent_mcp_action_resource.ts
@@ -718,12 +718,24 @@ export class AgentMCPActionResource extends BaseResource<AgentMCPActionModel> {
     });
   }
 
+  /**
+   * A message should never have blocked actions from more than one step, and resume paths rely
+   * on that to resume the agent loop from a single, unambiguous step. When `force` is false this
+   * enforces the invariant and throws on a violation, surfacing the bug. `force` (used by the
+   * unstick-conversation poke plugin) logs the anomaly instead of throwing, so a genuinely stuck
+   * conversation can still be finalized.
+   */
   static async listBlockedActionsForAgentMessage(
     auth: Authenticator,
     {
       agentMessageId,
       transaction,
-    }: { agentMessageId: ModelId; transaction?: Transaction }
+      force = false,
+    }: {
+      agentMessageId: ModelId;
+      transaction?: Transaction;
+      force?: boolean;
+    }
   ): Promise<AgentMCPActionResource[]> {
     const actions = await this.baseFetch(
       auth,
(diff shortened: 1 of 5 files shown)


front/lib/resources/agent_mcp_action_resource.ts
```diff
@@ -759,17 +772,23 @@ export class AgentMCPActionResource extends BaseResource<AgentMCPActionModel> {
...
+   * single-step invariant; the unstick-conversation poke plugin passes true to finalize an
+   * anomalous, genuinely stuck conversation instead of throwing.
    */
   static async denyBlockedActionsForAgentMessage(
     auth: Authenticator,
     {
       agentMessageId,
       transaction,
-    }: { agentMessageId: ModelId; transaction: Transaction }
+      force = false,
```
aubin-tchoi: how about a name more explicit to make sure we only use it for this? something like `bypassSameStepCheck` or even `dangerouslyBypass...`?

aubin-tchoi (approved):
LGTM

PR #28679, opened 2026-07-08
Fix: do not show depth > 0 conversations in pods.

Fraggle:
## Description

Sub-conversations (agent-spawned child conversations with `depth > 0`) were leaking into Pod conversation lists. Added `depth: { [Op.eq]: 0 }` filters to both `listSpaceUnreadConversationsAndActivityForUser` and `listConversationsInSpacePaginated` in `ConversationResource` so only root conversations are returned.

## Tests

Added unit tests for both query paths in `conversation_resource.test.ts`: create a conversation, force-set `depth: 1` via `ConversationModel.update`, assert it's excluded from the result while root conversations are still returned.

## Risk

Low. Two additive `WHERE` clause filters on an existing indexed column (`depth`). No schema change. Existing conversations at `depth: 0` are unaffected.

## Deploy Plan

Deploy front.


Fraggle:

- #28679 ◀

Fraggle:
(pmrr as it adds visual noise)

adrsimon (commented):
LGTM

PR #28681, opened 2026-07-08
Fix: deployment instruction danger rules detection

Fraggle:
## Description

Two fixes in `checkDeployPlanSection` in `front/dangerfile.ts`:
- `danger.github.pr.body ?? ""` — prevents a crash when the PR body is `null`.
- `\r?\n` instead of `\r\n` in the Deploy Plan section regex — makes the match work on both CRLF and LF line endings (GitHub delivers either depending on the client).

These were silently causing the danger rule to fail to detect a filled-in Deploy Plan section, so valid PRs would get a false-positive danger warning.

## Tests

Tested locally by running the danger check against PRs with LF-only bodies and null bodies.

## Risk

Low. Change is limited to the dangerfile CI check; no production code affected. Fully safe to rollback.

## Deploy Plan

No deploy needed — dangerfile change only, takes effect on the next CI run.

### DIFF
--- a/front/dangerfile.ts
+++ b/front/dangerfile.ts
@@ -148,10 +148,10 @@ function checkSDKLabel() {
 }
 
 function checkDeployPlanSection() {
-  const PRDescription = danger.github.pr.body;
+  const PRDescription = danger.github.pr.body ?? "";
 
   const deployPlanSectionRegex =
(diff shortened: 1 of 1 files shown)


Fraggle:

- #28681 ◀

PR #28683, opened 2026-07-08
[SPACES] Handle group name unique constraint race in createSpaceAndGroup

pmilliotte:
## Description

Catch `UniqueConstraintError` when creating the members group in `createSpaceAndGroup` and return a `space_already_exists` `DustError` instead of a 500, fixing a race when two space creations collide on the same group name.

## Tests

Tested locally.

## Risk

Low.

## Deploy Plan

- Deploy `main` on `front`

### DIFF
--- a/front/lib/api/spaces.ts
+++ b/front/lib/api/spaces.ts
@@ -39,7 +39,7 @@ import { Err, Ok } from "@app/types/shared/result";
 import { assertNever } from "@app/types/shared/utils/assert_never";
 import assert from "assert";
 import uniq from "lodash/uniq";
-import { Op } from "sequelize";
+import { Op, UniqueConstraintError } from "sequelize";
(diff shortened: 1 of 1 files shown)


Fraggle (approved):
Thanks, this fixes one task in my todo :)

front/lib/api/spaces.ts
```diff
@@ -522,14 +522,27 @@ export async function createSpaceAndGroup(
...
+          kind: "regular",
+        },
+        { transaction: t }
+      );
+    } catch (err) {
+      if (err instanceof UniqueConstraintError) {
+        return new Err(
+          new DustError(
+            "space_already_exists",
+            "This space name is already used."
```
Fraggle: nit: space/pod ?

PR #28684, opened 2026-07-08
Rework default model selection for agent builder.

Fraggle:
## Description

Default model selection in the agent builder was split across client-side logic: `getDefaultModel()` computed the best model from a hardcoded `PREFERRED_LARGE_MODEL_IDS` list after models loaded asynchronously, causing a render race where the form could initialize with a stale `CLAUDE_SONNET_4_6` fallback before being corrected by a `useEffect`.

- `getModelsForAuth(auth)` in `enabled_models.ts` now returns both `models` and `defaultModel` together, replacing the two-step `getAvailableModelsForWorkspace` + `withModelSelectability` call in the `/models` route.
- `AgentBuilder` is split: a thin outer component waits for `defaultModel` (showing a spinner or error state), then passes it as a resolved prop to `AgentBuilderForm`.
- `getDefaultModel` and `PREFERRED_LARGE_MODEL_IDS` are removed from `transformAgentConfiguration.ts`; `getDefaultAgentFormData` and `transformTemplateToFormData` now require `defaultModel` explicitly.
- `pickPreferredLargeModel` is extracted as a reusable exported fn and used in both `getDefaultModelFromEnabledModels` and `resolveModel`.
- `createPendingAgentConfiguration` now uses `getModelsForAuth` instead of hardcoding `CLAUDE_SONNET_4_6`.

## Tests

- Added Vitest unit tests for `getDefaultModelFromEnabledModels` (preferred order, non-selectable skip, hardcoded fallback) in `enabled_models.test.ts`.
- Added Vitest unit tests for `pickPreferredLargeModel` (preferred order, large model fallback, empty list fallback) in `models.test.ts`.
- Tested locally: new agent creation picks the correct workspace-level default model without a flash of the hardcoded Sonnet fallback.

## Risk

Low. Pure front-end refactor — no DB changes, no API contract changes (the `/models` route response shape is unchanged, now just also includes `defaultModel`). The hardcoded `CLAUDE_SONNET_4_6` fallback is preserved for the edge case where no selectable models exist. Rollback is safe.

## Deploy Plan

Deploy front.

### DIFF
--- a/front/components/agent_builder/AgentBuilder.tsx
+++ b/front/components/agent_builder/AgentBuilder.tsx
@@ -25,7 +25,6 @@ import { useSidekickMCPServer } from "@app/components/agent_builder/sidekick/use
 import { submitAgentBuilderForm } from "@app/components/agent_builder/submitAgentBuilderForm";
 import {
   getDefaultAgentFormData,
-  getDefaultModel,
   transformAgentConfigurationToFormData,
   transformDuplicateAgentToFormData,
   transformTemplateToFormData,
@@ -59,12 +58,20 @@ import { emptyArray, useFetcher } from "@app/lib/swr/swr";
 import { getConversationRoute } from "@app/lib/utils/router";
 import { removeParamFromRouter } from "@app/lib/utils/router_util";
 import datadogLogger from "@app/logger/datadogLogger";
+import type { EnabledModelConfigurationType } from "@app/types/api/assistant/models";
 import type { AgentConfigurationType } from "@app/types/assistant/agent";
 import type { TemplateInfo } from "@app/types/assistant/templates";
 import { normalizeError } from "@app/types/shared/utils/error_utils";
 import { isString, removeNulls } from "@app/types/shared/utils/general";
 import { pluralize } from "@app/types/shared/utils/string_utils";
 import { isBuilder } from "@app/types/user";
+import {
+  ContentMessage,
+  ContentMessageAction,
+  InfoCircle,
+  RefreshCw02,
+  Spinner,
+} from "@dust-tt/sparkle";
 import { zodResolver } from "@hookform/resolvers/zod";
 import set from "lodash/set";
 import {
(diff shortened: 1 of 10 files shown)


Fraggle:

- #28670
    - #28684 ◀
        - #28592

front/components/agent_builder/transformAgentConfiguration.ts
```diff
@@ -59,52 +50,13 @@ export function transformAgentConfigurationToFormData(
   };
 }
 
-const PREFERRED_LARGE_MODEL_IDS: ModelIdType[] = [
```
Fraggle, line 62: move to the backend's response.

front/components/agent_builder/AgentBuilder.tsx
```diff
@@ -110,12 +117,63 @@ interface AgentBuilderProps {
...
+      {...props}
+      newAgentDefaultModel={
+        props.agentConfiguration ? undefined : defaultModel!
+      }
+    />
+  );
+}
+
+interface AgentBuilderFormProps extends AgentBuilderProps {
+  newAgentDefaultModel?: EnabledModelConfigurationType;
```
Nils-Fedrigo, line 167: rename agentDefaultModel?

Nils-Fedrigo: also why optional?

Nils-Fedrigo: it's a component you use only there and then you need if(newAgentDefaultModel) and newAgentDefaultModel!

front/components/agent_builder/AgentBuilder.tsx
```diff
@@ -317,17 +385,19 @@ export default function AgentBuilder({
...
         slackChannels: agentSlackChannels,
       },
-      // For new agents, update model settings once available models are loaded.
-      ...(!agentConfiguration && {
-        generationSettings: {
-          ...currentValues.generationSettings,
-          modelSettings: {
-            modelId: defaultModel.modelId,
-            providerId: defaultModel.providerId,
+      // Templates may preset a model; override it with the backend default.
```
Nils-Fedrigo, line 388: not sure if templates can currently include model, but  maybe that'd be future proof to switch modelSettings first then ...currentValues.generationSettings,

front/lib/advanced_models/enabled_models.ts
```diff
@@ -35,3 +37,31 @@ export async function withModelSelectability(
       allowedAdvancedModelKeys.has(advancedModelKey(model)),
   }));
 }
+
+export async function getEnabledModelsForAuth(
+  auth: Authenticator
+): Promise<EnabledModelConfigurationType[]> {
+  const availableModels = await getAvailableModelsForWorkspace(auth);
+  return withModelSelectability(auth, { models: availableModels });
```
Nils-Fedrigo, line 45: just to be sure, everywhere "model"= "model+effort level" ?

front/lib/api/assistant/configuration/agent.ts
```diff
@@ -122,11 +123,10 @@ export async function createPendingAgentConfiguration(
...
         description: PENDING_AGENT_PLACEHOLDER_DESCRIPTION,
         instructions: null,
-        providerId: CLAUDE_SONNET_4_6_DEFAULT_MODEL_CONFIG.providerId,
-        modelId: CLAUDE_SONNET_4_6_DEFAULT_MODEL_CONFIG.modelId,
+        providerId: defaultModel.providerId,
+        modelId: defaultModel.modelId,
         temperature: 0.7,
-        reasoningEffort:
-          CLAUDE_SONNET_4_6_DEFAULT_MODEL_CONFIG.defaultReasoningEffort,
+        reasoningEffort: defaultModel.defaultReasoningEffort,
```
Nils-Fedrigo, line 131: ok so I have my answer, in the future getModelsForAuth() should return a reasoning answer too

Nils-Fedrigo (approved):
thanks!

front/lib/advanced_models/enabled_models.ts
Fraggle: Nope, here for example it's models + available effort levels.

front/lib/api/assistant/configuration/agent.ts
Fraggle: Yep, i agree.

front/components/agent_builder/AgentBuilder.tsx
Fraggle: Good point.

PR #28638, opened 2026-07-07
[Agentic User Activation] Poke provisioning plugin skeleton

xebeohp:
## Description
The skeleton version of the poke provisioning tool to create activation pods. What this skeleton version does:
- Create a new pod for the user from poke (`Join Activation Pod` plugin) and invite them to the pod (works for multiple users as well)
- From poke set the default skills for the activation pod (can select multiple since it calls the default skills feature code)
- From poke add instructions to the [AGENTS.md](http://agents.md/) file
- Store that this is an activation pod in the `provisioningSource` column in `project_metadata`

## Tests
Manual tests on local DB, poke, and dust.

## Risk
Low risk. Only activated by a manual Poke plugin.

## Deploy Plan
Normal deployment.

PR #28686, opened 2026-07-08
[Agentic User Activation] Poke plugin: create initial conversation

xebeohp:
## Description
When a Pod is created from the Join Activation Pod Poke plugin, an initial conversation is pre-loaded in the pod which calls the Activation Skill using a message hidden from the user.

There is no UI in the Poke plugin for this.

## Tests
Manual local tests

## Risk
Low risk, all controlled through a Poke plugin.

## Deploy Plan
Normal deployment. Part of Agentic User Activation.


xebeohp:

- #28638
    - #28686 ◀

front/lib/api/poke/plugins/spaces/join_activation_pod.ts
```diff
@@ -292,12 +296,56 @@ export const joinActivationPodPlugin = createPlugin({
...
+      workspace.sId
+    );
+    const podMemberAuth = await Authenticator.fromUserIdAndWorkspaceId(
+      creator.sId,
+      workspace.sId
+    );
+
+    let activationTriggerCreated = false;
+    const podViewResult = await ensureActivationPodView(adminAuth, pod);
+    if (podViewResult.isErr()) {
```
frankaloia, line 333: Should we throw these errors (this one and below) and stop the plugin flow instead of catching?

front/lib/api/activation/trigger.ts
```diff
@@ -0,0 +1,174 @@
+import type { Authenticator } from "@app/lib/auth";
```
frankaloia, line 1: We do have a /activation subfolder that I think this file should go in

front/lib/api/assistant/activation.ts
```diff
@@ -0,0 +1,174 @@
...
+  try {
+    const source = await WebhookSourceResource.fetchByName(
+      adminAuth,
+      ACTIVATION_WEBHOOK_SOURCE_NAME
+    );
+    if (!source) {
+      return new Err(
+        new Error(
+          `Webhook source "${ACTIVATION_WEBHOOK_SOURCE_NAME}" is not provisioned ` +
+            "for this workspace. Create it via the webhook sources admin UI (or " +
```
frankaloia: nit: Not possible to configure via the admin UI today

front/lib/api/assistant/activation.ts
```diff
@@ -0,0 +1,174 @@
...
+function activationEventBody(podSId: string): Record<string, unknown> {
+  return { [ACTIVATION_POD_ID_FIELD]: podSId };
+}
+
+export async function ensureActivationPodView(
+  adminAuth: Authenticator,
+  pod: SpaceResource
+): Promise<Result<WebhookSourcesViewResource, Error>> {
+  try {
+    const source = await WebhookSourceResource.fetchByName(
```
frankaloia: For this to work in their workspace, do we need to create the webhook source too?

front/lib/api/assistant/activation.ts
```diff
@@ -0,0 +1,174 @@
...
+import { Err, Ok } from "@app/types/shared/result";
+import { normalizeError } from "@app/types/shared/utils/error_utils";
+
+// A single workspace-level webhook source is shared by all Activation Pods.
+export const ACTIVATION_WEBHOOK_SOURCE_NAME = "Activation";
+
+const ACTIVATION_POD_ID_FIELD = "podId";
+
+export const ACTIVATION_CUSTOM_PROMPT =
+  "Welcome me to my new Pod and recommend the next best action to get more value from Dust." +
```
frankaloia: nit: Lots of logic is baked into the skill, so I think we can just shorten to the first sentence

front/lib/api/assistant/activation.ts
```diff
@@ -0,0 +1,174 @@
...
+
+function activationTriggerFilter(podSId: string): string {
+  return `(eq "${ACTIVATION_POD_ID_FIELD}" "${podSId}")`;
+}
+
+function activationEventBody(podSId: string): Record<string, unknown> {
+  return { [ACTIVATION_POD_ID_FIELD]: podSId };
+}
+
+export async function ensureActivationPodView(
```
frankaloia: nit: I think this function can be better named to let us know it is referring to the webhook source

front/lib/api/activation/trigger.ts
```diff
@@ -0,0 +1,174 @@
...
+      includePayload: false,
+      filter: activationTriggerFilter(pod.sId),
+    },
+    naturalLanguageDescription: null,
+    customPrompt: ACTIVATION_CUSTOM_PROMPT,
+    editor: creator.id,
+    webhookSourceViewId: podView.id,
+    executionPerDayLimitOverride: 1,
+    executionMode: "fair_use",
+    origin: "user",
```
frankaloia, line 150: I think we actually want a new origin for this given it's not technically "user"?

front/lib/api/activation/trigger.ts
```diff
@@ -0,0 +1,174 @@
...
+    });
+
+    return new Ok(podView);
+  } catch (err) {
+    return new Err(normalizeError(err));
+  }
+}
+
+// Creates the pod's user-owned activation trigger. Runs under the creator's
+// authenticator so the trigger is owned by the user.
```
frankaloia, line 108: Can we add a TODO that this type of creation is temporary for testing and we will add a consent path for users? Don't want readers to be concerned about this approach

frankaloia (commented):
Thanks, worked great, just some clean up items.

frankaloia:
Pushed changes addressing the review comments (commit 6d833f9). Branch also updated with latest `main`.

- **Move to `/activation` subfolder** — `lib/api/assistant/activation.ts` → `lib/api/activation/trigger.ts`.
- **Rename** — `ensureActivationPodView` → `getOrCreateActivationWebhookSourceView` (makes clear it's the webhook source).
- **Auto-create the webhook source** — the `getOrCreate…` helper now provisions the `Activation` webhook source if it doesn't exist, so this works in any workspace (no admin-UI/seed prerequisite). Pod-view lookup is scoped by space, which also avoids resolving orphaned views from deleted pods.
- **Shorten prompt** — trimmed to the first sentence; `customPrompt` uses `""` (the product's "no prompt" convention for webhook triggers), not `null`.
- **Stop the flow on error** — the poke plugin now returns `Err` on webhook-view/trigger failures instead of catch-and-log.
- **Temporary on-behalf-of creation** — added a `TODO(activation)` note; kept `origin: "user"` (with a comment) rather than introducing a new origin, since it's transient and becomes accurate once the consent path lands.

Also added while testing:
- Optional **Pod name** arg on the plugin (falls back to `"<creator>'s Activation Pod"`) so re-runs don't collide on a duplicate name.
- The plugin now **emits the activation event** at the end of provisioning so the trigger fires the initial conversation immediately.

Note: with the new Pod name field, the description's "There is no UI in the Poke plugin" is now slightly out of date.

frankaloia:
Added (commit c8c5686): the provisioned pod is now **starred for every member** so it surfaces in their sidebar — via a new batch `UserProjectPreferencesResource.setStarredForUsers` called from the poke plugin after the pod is created.

frankaloia (approved):
LGTM (I updated PR after my initial comments)

PR #28688, opened 2026-07-08
refactor(agents): add AgentUserRelationResource

rfrenoy:
## Description

Introduces a Resource wrapper over the agent_user_relations table (deleteForAgent / countForAgent) so callers stop touching the AgentUserRelationModel Sequelize model directly ([BACK3]). Consumed by the agent hard-delete cascade PR that will be submitted as a follow-up.

Broader migration of the other existing direct AgentUserRelationModel usages (favorites views, favorite states, user scrub) is intentionally out of scope.

## Tests

None

## Risk

None
## Deploy Plan

Front


flvndvd (approved):
Fine as a transitory step, end goal, this one should probably end up in the `AgentConfigurationResource`.

PR #28692, opened 2026-07-08
[poke] improve plan type filtering

tdraier:
## Description

Follow-up to the poke workspaces plan-type filter. The filter previously over-fetched up to 100 workspaces per query and did the filtering/pagination in JS — which silently broke once a bucket had more candidates than that ahead of it in `createdAt DESC` order (e.g. filtering to "Enterprise" among ~500k free workspaces could come back empty or with a wrong `hasMore`).

This moves the filtering into SQL:

- Every bucket except `free` (enterprise, legacy_enterprise, legacy_pro, business, friends_and_family, dust) is now filtered via an inner join on `subscriptions.plan.code` directly in the main query, so the DB does the filtering *and* the pagination in one go — we only ever fetch the requested page, never the full matching set.
- `free` has no plan-code pattern of its own (it's "no active subscription, or one that isn't any other bucket"), so it stays an exclude-list: fetch the ids of the non-free buckets (expected to stay in the low thousands — ~500 enterprise + ~5000 pro/business + F&F/dust — even as the free tier grows into the hundreds of thousands) and `NOT IN` them.
- The bucket → plan-code mapping now lives in one place (`POKE_PLAN_CODE_MATCHERS` in `plan_codes.ts`), consumed by `SubscriptionResource.buildPokePlanCodeWhere` / `listActiveWorkspaceIdsWithNonFreePlanType` — no more duplicated prefix logic between a JS classifier and the SQL.
- Worked around a Sequelize limitation along the way: `limit` combined with a nested `required: true` include triggers Sequelize's automatic subquery-splitting for hasMany pagination, which generated invalid SQL (a join clause referencing an alias never brought into that subquery's scope). Disabled with `subQuery: false`, which is safe here since a workspace has at most one active subscription (enforced everywhere a subscription is switched), so the join can't duplicate rows.

## Tests

Extended `front-api/routes/poke/workspaces/index.test.ts`:
- Added coverage for the `free` bucket correctly including a workspace with no active subscription, and excluding that same workspace from other buckets.
- Existing plan-type filter and pagination tests (offset/hasMore, search combined with a filter) continue to pass end-to-end against the new SQL-based implementation.

Ran `npx tsgo --noEmit` on `front` and `front-api`, and the full `subscription_resource.test.ts` suite.

## Risk

Low — poke is an internal, super-user-only admin tool, and this is entirely behind the existing opt-in `planType` query param. No change to the response contract. The `subQuery: false` addition only applies when a plan-type filter is set, and relies on an invariant (≤1 active subscription per workspace) already enforced by every subscription-switching code path.

## Deploy Plan

Standard rollout — no migration, backfill, or feature flag needed.

### DIFF
--- a/front-api/routes/poke/workspaces/index.test.ts
+++ b/front-api/routes/poke/workspaces/index.test.ts
@@ -1,3 +1,4 @@
+import { SubscriptionResource } from "@app/lib/resources/subscription_resource";
 import { WorkspaceResource } from "@app/lib/resources/workspace_resource";
 import { createPrivateApiMockRequest } from "@app/tests/utils/generic_private_api_tests";
 import { WorkspaceFactory } from "@app/tests/utils/WorkspaceFactory";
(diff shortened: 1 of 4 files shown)


tdraier:

- #28692 ◀

adrsimon (approved):
LGTM 👍

PR #28693, opened 2026-07-08
fix(agents): cancel wake-ups when an agent is archived

rfrenoy:
## Description

archiveAgentConfiguration disabled the agent's triggers but never cancelled its wake-ups, so a cron wake-up kept firing after the agent was archived. It now cancels every wake-up for the agent (deleting the cron schedule / pending one-shot workflow, and reconciling a leaked schedule for already-terminal cron wake-ups), with bounded concurrency.

Adds:
- WakeUpResource.cancelForCascade (permission-checkless system cancel that also reconciles terminal-cron leaked schedules) and listByAgentConfigurationId.
- A (workspaceId, agentConfigurationId) index on wake_ups
## Tests

Unit

## Risk

Low
## Deploy Plan

pre-deploy mig + Front

### DIFF
--- a/front/lib/api/assistant/configuration/agent.test.ts
+++ b/front/lib/api/assistant/configuration/agent.test.ts
@@ -21,6 +21,7 @@ import { ConversationFactory } from "@app/tests/utils/ConversationFactory";
 import { createResourceTest } from "@app/tests/utils/generic_resource_tests";
 import { MembershipFactory } from "@app/tests/utils/MembershipFactory";
 import { UserFactory } from "@app/tests/utils/UserFactory";
+import { WakeUpFactory } from "@app/tests/utils/WakeUpFactory";
 import { Ok } from "@app/types/shared/result";
 import { describe, expect, it, vi } from "vitest";
 
@@ -352,32 +353,20 @@ describe("archiveAgentConfiguration and restoreAgentConfiguration", () => {
       messagesCreatedAt: [new Date()],
     });
 
-    const wakeUpResult = await WakeUpResource.makeNew(
+    const wakeUp = await WakeUpFactory.cron(
       authenticator,
-      {
-        scheduleType: "cron",
-        fireAt: null,
-        cronExpression: "0 7 * * *",
-        cronTimezone: "Europe/Paris",
-        reason: "Daily wake-up",
-      },
       conversation,
-      agent
+      agent,
+      {
+        reason: "Daily wake-up",
+      }
     );
-    if (wakeUpResult.isErr()) {
-      throw new Error(
-        `Failed to create test wake-up: ${wakeUpResult.error.message}`
-      );
-    }
 
     const archived = await archiveAgentConfiguration(authenticator, agent.sId);
     expect(archived).toBe(true);
 
     expect(cancelSpy).toHaveBeenCalled();
-    const refetched = await WakeUpResource.fetchById(
-      authenticator,
-      wakeUpResult.value.sId
-    );
+    const refetched = await WakeUpResource.fetchById(authenticator, wakeUp.sId);
     expect(refetched?.status).toBe("cancelled");
 
     launchSpy.mockRestore();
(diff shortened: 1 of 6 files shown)


front/lib/api/assistant/configuration/agent.test.ts
```diff
@@ -331,6 +335,107 @@ describe("archiveAgentConfiguration and restoreAgentConfiguration", () => {
...
+
+    const { authenticator } = await createResourceTest({ role: "admin" });
+    const agent =
+      await AgentConfigurationFactory.createTestAgent(authenticator);
+    const conversation = await ConversationFactory.create(authenticator, {
+      agentConfigurationId: agent.sId,
+      messagesCreatedAt: [new Date()],
+    });
+
+    const wakeUpResult = await WakeUpResource.makeNew(
```
flvndvd: `nit` would be nice to have a Factory for wake ups.

front/lib/resources/wakeup_resource.ts
```diff
@@ -338,6 +338,23 @@ export class WakeUpResource extends BaseResource<WakeUpModel> {
...
+    agentConfigurationId: string,
+    { status }: { status?: WakeUpStatus | WakeUpStatus[] } = {}
+  ): Promise<WakeUpResource[]> {
+    return this.baseFetch(auth, {
+      where: {
+        agentConfigurationId,
+        ...(status !== undefined ? { status } : {}),
+      },
+      order: [
+        ["createdAt", "ASC"],
```
flvndvd, line 352: just to make sure that we have proper index coverage on those.

front/lib/resources/wakeup_resource.ts
```diff
@@ -486,8 +503,28 @@ export class WakeUpResource extends BaseResource<WakeUpModel> {
...
+   * Ensures a wake-up leaves no live Temporal footprint as part of a
+   * system-initiated cascade (agent archive / hard-delete), without the
+   * owner/admin permission check, since the agent it targets is going away.
+   *
+   * - If still scheduled: cancels it (deletes the cron schedule / cancels the
+   *   pending one-shot workflow) and marks the row cancelled.
+   * - If already terminal: reconciles a cron schedule that may have been left
+   *   orphaned by a prior failed cleanup (no-op for one-shots / non-cron).
+   */
+  async cancelForCascade(
```
flvndvd: Not sure to get the meaning of cascade here?

front/migrations/pre-deploy/20260708120858_index_wake_ups_on_agent_configuration_id.sql
```diff
@@ -0,0 +1,6 @@
+/*
+Statement 0
+*/
+SET SESSION statement_timeout = 1200000;
+SET SESSION lock_timeout = 3000;
+CREATE INDEX CONCURRENTLY "wake_ups_workspace_id_agent_configuration_id_idx" ON public.wake_ups USING btree ("workspaceId", "agentConfigurationId");
```
flvndvd, line 6: 👌

flvndvd (approved):

front/lib/resources/wakeup_resource.ts
rfrenoy: it's because it cascades from the agent being archived/deleted, but I can rename to something like `cancelForRemovedAgent` if it's clearer. I'll see what I can do

front/migrations/pre-deploy/20260708120858_index_wake_ups_on_agent_configuration_id.sql
id13: just passing by, workspaceId is necessary at all for an agentConfigurationId equality lookup, and if we have an agentConfigurationId index already (which we should !) then this whole index is not necessary

front/lib/api/assistant/configuration/agent.test.ts
rfrenoy: done in 236cd8115fb038420c217b4c18b3dc8755daeeea

front/migrations/pre-deploy/20260708120858_index_wake_ups_on_agent_configuration_id.sql
rfrenoy: I thought as the model is "workspace aware" we must use workspace id as the first field in a compound index for it to work? So don't we need a (workspaceId, agentConfigurationId) index?

front/lib/resources/wakeup_resource.ts
rfrenoy: will simply rename to `forceCancel`

rfrenoy: done in 662a1fc557a22d3880a5cd762b8748ebdfd31c3b

front/migrations/pre-deploy/20260708120858_index_wake_ups_on_agent_configuration_id.sql
id13: discussed irl, agentConfigurationId is NOT an FK so my comment is not relevant

PR #28695, opened 2026-07-08
fix(front): resolve Jira issue type by project-scoped id when creating issues

adrsimon:
## Description

The built-in Jira `create_issue` tool rejected sub-task creation by type name (e.g. `"Sous-tâche"`) with a 400 `Le type de ticket sélectionné n'est pas valide`, even though `get_issue_types` returned that same type as valid for the project.

Root cause: Jira's `POST /rest/api/3/issue` resolves `issuetype.name` **instance-wide**, not within the target project. When several projects/schemes share a localized type name — as sub-task types do — that lookup is ambiguous and Jira rejects an otherwise valid type. The tool only accepted `issuetype: { name }` and sent it verbatim, so there was no way to disambiguate.

Changes:
- `create_issue` now resolves the type name against the **project-scoped** `get_issue_types` list (`createmeta/{projectKey}/issuetypes`) and sends the resolved `id`. If resolution fails, it falls back to the name so behavior is no worse than before.
- The `issuetype` schema now also accepts an explicit `id` (preferred), so callers can bypass name resolution entirely. Passing `id` avoids the extra lookup.

## Tests

Type-check and biome pass. Verified the resolution logic against the reported repro (projects IN / SAL, type `Sous-tâche` id `10006`): callers passing the name get the project-scoped id substituted; callers passing an id use it directly. No existing test harness for this internal MCP server, so no functional test added.

## Risk

Low. Change is scoped to `create_issue`; `update_issue` is untouched. Name-based creation adds one `get_issue_types` call per create (skipped when an id is passed). Fully backward compatible — existing callers keep passing `name`. Safe to roll back.

## Deploy Plan

Standard `front` deploy, no migration.

### DIFF
--- a/front/lib/api/actions/servers/jira/jira_api_helper.ts
+++ b/front/lib/api/actions/servers/jira/jira_api_helper.ts
@@ -822,6 +822,32 @@ async function processFieldsWithMetadata(
   return processFieldsForJira(fields, fieldsMetadata);
 }
 
+// Jira resolves `issuetype.name` instance-wide, which is ambiguous when projects share a
+// localized name (e.g. sub-tasks named "Sous-tâche") and gets rejected. Resolve to the
+// project-scoped id instead.
+async function resolveIssueType(
+  baseUrl: string,
+  accessToken: string,
+  projectKey: string,
+  issuetype: { id?: string; name?: string }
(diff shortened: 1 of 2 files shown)


thervier (approved):
LGTM

PR #28697, opened 2026-07-08
Publish Frames by their actual current path, not by name

flvndvd:
## Description

Two Frames in the same conversation could end up sharing the same file name. Publishing resolved which file to build from that name alone, so the second Frame to collide on a name could silently build from the first Frame's content instead of its own, or fail to find anything at all, with no way for the caller to tell it apart.

Rather than trusting a name that can collide or go stale after a rename, publishing now takes the entry's actual current location directly, the same thing already visible when listing the conversation's files. This removes the dependency on the file's own name entirely.

The one place that still needs to know the entry without asking again is a live edit made directly on the rendered Frame, since no model is involved there to supply it fresh. For that case, the entry is now remembered from the last successful publish and reused, rather than guessed from the file's name.

Added a test that publishes two Frames sharing a name and confirms each one builds from its own content, independently of the other.

### DIFF
--- a/front/lib/api/actions/servers/interactive_content/metadata.ts
+++ b/front/lib/api/actions/servers/interactive_content/metadata.ts
@@ -302,9 +302,7 @@ export const INTERACTIVE_CONTENT_TOOLS_METADATA = createToolsRecord({
             "`conversation-<conversationId>/<filename>`, or `pod-<id>/<filename>` for a Frame " +
             "stored in a project's shared space. If the file was renamed, use its current " +
             "name, listing the directory first if unsure. This path's directory becomes the " +
-            "bundling root: keep every source file under it. Anything reaching outside it, " +
-            "whether `../` above the root or into a different scoped path, is not bundled and " +
-            "will not resolve at render."
+            "bundling root for relative imports."
         ),
     },
     enableAlerting: true,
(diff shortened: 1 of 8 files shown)


adrsimon (approved):

front/lib/api/actions/servers/interactive_content/metadata.ts
```diff
@@ -282,11 +282,12 @@ export const INTERACTIVE_CONTENT_TOOLS_METADATA = createToolsRecord({
...
-      "imports), validates TypeScript and JSX, then updates the canonical Frame so viewers and " +
-      "shares see the new version. A syntax error blocks publishing and is reported back so you " +
-      "can fix it. Tailwind warnings are returned but do not block. Pass the `file_id` of the " +
-      "Frame and `path` set to `conversation-<conversationId>`, the directory that holds the file.",
+      "live Frame. It resolves the Frame's dependency tree from the entry file's directory " +
+      "(inlining relative imports), validates TypeScript and JSX, then updates the canonical " +
+      "Frame so viewers and shares see the new version. A syntax error blocks publishing and is " +
+      "reported back so you can fix it. Tailwind warnings are returned but do not block. Pass " +
+      "the `file_id` of the Frame and `path` set to the entry file's current scoped path, e.g. " +
+      "`conversation-<conversationId>/<filename>`.",
```
aubin-tchoi, line 290: not sure I understand what "the entry file's current scoped path" means

aubin-tchoi (approved):

front/lib/api/actions/servers/interactive_content/metadata.ts
aubin-tchoi: ah it's the current file path (even if it was moved, etc...)?

flvndvd: yes

flvndvd: Will do some more tweaks afterward.

PR #28701, opened 2026-07-08
feat(front): show radio btn or checkbox when asking question

sfriquet:
## Description

Fixes [dust-tt/tasks#8739](https://github.com/dust-tt/tasks/issues/8739). 

This adds an explicit trailing selection indicator to each option row, radio button for single select, checkbox for multi select.

## Tests

Screenshot 2026-07-08 at 16 14 49
Screenshot 2026-07-08 at 16 14 54
Screenshot 2026-07-08 at 16 16 49
Screenshot 2026-07-08 at 16 17 30

### DIFF
--- a/front/components/assistant/conversation/UserAnswerRequired.tsx
+++ b/front/components/assistant/conversation/UserAnswerRequired.tsx
@@ -296,6 +296,7 @@ export function UserAnswerRequired({
               counterValue={index + 1}
               selected={answerDraft.selectedOptions.includes(index)}
               disableHover={isKeyboardNavigating}
+              selectionIndicator={question.multiSelect ? "checkbox" : "radio"}
               onFocusCapture={() => activateOption(index)}
               onMouseEnter={() => activateOption(index)}
               className={cn(
--- a/sparkle/src/components/OptionCard.tsx
+++ b/sparkle/src/components/OptionCard.tsx
@@ -1,10 +1,14 @@
 import { Card, type CardVariantType } from "@sparkle/components/Card";
-import { Checkbox } from "@sparkle/components/Checkbox";
+import {
+  checkboxIndicatorStyles,
+  checkboxStyles,
+} from "@sparkle/components/Checkbox";
 import { Counter } from "@sparkle/components/Counter";
 import {
   radioIndicatorStyles,
   radioStyles,
 } from "@sparkle/components/RadioGroup";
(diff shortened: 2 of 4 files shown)


sparkle/src/components/OptionCard.tsx
```diff
@@ -69,6 +78,24 @@ export function OptionCard({
...
+      {selectionIndicator === "radio" && (
+        <div
+          aria-hidden="true"
+          className={cn(radioStyles(), "pointer-events-none shrink-0")}
+        >
+          {selected && <div className={radioIndicatorStyles()} />}
+        </div>
+      )}
+      {selectionIndicator === "checkbox" && (
+        <Checkbox
```
PopDaph: Question:  Why not exporting the checkbox styles from Checkbox and rendering a decorative div, like the radio branch does with radioStyles (and like Dropdown / DataTable)? Rendering the real Checkbox puts a <button> inside the card's role="button" and forces the 4 disarm props.

PopDaph (approved):
Thanks!!

front/components/assistant/conversation/UserAnswerRequired.tsx
```diff
@@ -296,6 +296,7 @@ export function UserAnswerRequired({
               counterValue={index + 1}
               selected={answerDraft.selectedOptions.includes(index)}
               disableHover={isKeyboardNavigating}
+              selectionIndicator={question.multiSelect ? "checkbox" : "radio"}
```
aubin-tchoi, line 299: keep me honest but I think you'll never see the radio button filled in no? given that when you click on an option row it submits immediately

aubin-tchoi (approved):
LGTM 🙏🏻

sparkle/src/components/OptionCard.tsx
sfriquet: Right! Updated to use `div` + `Check`

front/components/assistant/conversation/UserAnswerRequired.tsx
sfriquet: correct, with the current submit behavior we won't display the `selected` state. Imo it's worth handling it anyway since the submit behavior isn't the responsibility of `OptionCard`

aubin-tchoi: agreed, from the component's perspective it does make sense
works for me!

PR #28702, opened 2026-07-08
feat(assistant): show default and per-message model in conversation picker

Nils-Fedrigo:
## Description

Improvements to the conversation input-bar model picker and message header (all behind the `models_picker` feature flag). Stacked on top of #28592.

- **"Default from agent" section**: when the selected agent has a fixed configured model (i.e. it is not on `auto`), the picker shows a **Default from agent** section above **Suggested** with the agent's real model, and dedupes that model out of the Suggested pins.
- **Picker defaults to the agent model, not Auto**: for a fixed-model agent the picker now opens on **Model: Default** (agent's model, Auto toggle off) instead of **Model: Auto**; agents on the auto model still show **Model: Auto**. "Default" and "Auto" both send no per-message override — the agent runs its configured model.
- **Per-message model label**: agent messages that ran on a model different from the agent's preset now show it next to the agent name in secondary color (e.g. `Dust with GPT-5.5 Light`). This covers both per-message picker overrides and `auto` agents that resolved to a concrete model. Comparison is a strict triplet match (provider/model/effort) against the agent's configured model.
- **Picker persistence**: the picker seeds its selection from the last message's requested model, so an explicit model choice sticks across sends and reloads.

To support the message label, `resolvedModel` and the agent's configured `model` are now carried on `LightAgentMessageType` (and its swagger schema), populated in `getLightAgentMessageFromAgentMessage` and left null on optimistic placeholders.

## Tests

Typecheck (`front` + `front-api`) and the affected unit tests pass (`useAgentMessageStream`, `InlineActivitySteps`). UI behavior was not exercised end-to-end (requires the `models_picker` flag and a live workspace/conversation).

## Deploy Plan

Standard deploy, no migration needed

### DIFF
--- a/front-api/routes/swagger_private_schemas.ts
+++ b/front-api/routes/swagger_private_schemas.ts
@@ -408,10 +408,10 @@
  *           type: number
  *           nullable: true
  *           description: Aggregated credit cost of all sub-agents (run_agent / agent_handover) spawned recursively by this message. Computed only on single-message fetches; null otherwise.
- *         requestedModel:
+ *         resolvedModel:
  *           type: object
  *           nullable: true
- *           description: Per-message model override from the input-bar model picker. Null when the agent ran its configured model.
+ *           description: Model triplet used to generate the message. Null when the agent ran its configured model (legacy).
  *           properties:
  *             providerId:
  *               type: string
@@ -419,6 +419,11 @@
  *               type: string
  *             reasoningEffort:
  *               type: string
+ *         modelResolutionMethod:
+ *           type: string
+ *           nullable: true
+ *           enum: [agent, user, auto]
+ *           description: How resolvedModel was chosen - agent (configured model), user (per-message picker), or auto (routed through the auto model). Null (legacy).
  *     PrivateLightAgentMessage:
  *       type: object
  *       description: A lighter agent message used in paginated message list responses.
@@ -524,6 +529,22 @@
  *           type: number
  *           nullable: true
  *           description: Aggregated credit cost of all sub-agents (run_agent / agent_handover) spawned recursively by this message. Computed only on single-message fetches; null otherwise.
+ *         resolvedModel:
+ *           type: object
+ *           nullable: true
+ *           description: Model triplet used to generate the message. Null when the agent ran its configured model (legacy).
+ *           properties:
+ *             providerId:
+ *               type: string
+ *             modelId:
+ *               type: string
+ *             reasoningEffort:
+ *               type: string
+ *         modelResolutionMethod:
+ *           type: string
+ *           nullable: true
+ *           enum: [agent, user, auto]
+ *           description: How resolvedModel was chosen - agent (configured model), user (per-message picker), or auto (routed through the auto model). Null (legacy).
  *         activitySteps:
  *           type: array
  *           items:
--- a/front/components/assistant/conversation/AgentMessage.tsx
+++ b/front/components/assistant/conversation/AgentMessage.tsx
@@ -14,6 +14,7 @@ import type { FeedbackSelectorBaseProps } from "@app/components/assistant/conver
 import { FeedbackSelector } from "@app/components/assistant/conversation/FeedbackSelector";
 import { useAutoOpenFilesPanel } from "@app/components/assistant/conversation/files_panel/useAutoOpenFilesPanel";
 import { useGenerationContext } from "@app/components/assistant/conversation/GenerationContextProvider";
+import { getModelWithReasoningEffortLabel } from "@app/components/assistant/conversation/input_bar/modelPickerUtils";
 import { useAutoOpenInteractiveContent } from "@app/components/assistant/conversation/interactive_content/useAutoOpenInteractiveContent";
 import type {
   AgentMessageStateWithControlEvent,
@@ -65,7 +66,7 @@ import { useAuth, useFeatureFlags } from "@app/lib/auth/AuthContext";
 import { clientFetch } from "@app/lib/egress/client";
 import type { DustError } from "@app/lib/error";
 import { FILE_ID_PATTERN } from "@app/lib/files";
-import { getResolvedModelLabel } from "@app/lib/llms/model_configurations";
+import { getSupportedModelConfig } from "@app/lib/llms/model_configurations";
 import { getFilePreviewDirectivePaths } from "@app/lib/markdown/file_preview";
 import { extractFromString } from "@app/lib/mentions/format";
 import { useUnifiedAgentConfigurations } from "@app/lib/swr/assistants";
@@ -1050,12 +1051,17 @@ export function AgentMessage({
     canShowAgentConversationActions(agentConfiguration.sId);
   const isArchived = agentConfiguration.status === "archived";
 
-  // When this message ran on a per-message model picked from the input-bar
-  // picker (rather than the agent's preset or an auto resolution), surface the
-  // model next to the agent name (e.g. "Dust with GPT-5.5 Light").
-  const perMessageModelLabel =
+  const perMessageModel =
     agentMessage.modelResolutionMethod === "user" && agentMessage.resolvedModel
-      ? getResolvedModelLabel(agentMessage.resolvedModel)
+      ? getSupportedModelConfig(agentMessage.resolvedModel)
+      : null;
+  const perMessageModelLabel =
+    perMessageModel && agentMessage.resolvedModel
+      ? getModelWithReasoningEffortLabel({
+          kind: "model",
+          model: perMessageModel,
+          effort: agentMessage.resolvedModel.reasoningEffort,
+        })
       : null;
 
   const renderName = useCallback(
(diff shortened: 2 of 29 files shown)


Nils-Fedrigo:

- #28592
    - #28702 ◀
        - #28703

front-api/routes/swagger_private_schemas.ts
```diff
@@ -408,17 +408,22 @@
  *           type: number
  *           nullable: true
  *           description: Aggregated credit cost of all sub-agents (run_agent / agent_handover) spawned recursively by this message. Computed only on single-message fetches; null otherwise.
- *         requestedModel:
+ *         resolvedModel:
```
Nils-Fedrigo, line 411: was stale from previous change

front/components/assistant/conversation/AgentMessage.tsx
```diff
@@ -1060,6 +1075,11 @@ export function AgentMessage({
           canMention={canMention}
           isDisabled={isArchived}
         />
+        {perMessageModelLabel && (
+          <span className="pl-1 font-normal text-muted-foreground">
+            with {perMessageModelLabel}
```
Fraggle: nit: i wonder if user would be confused by this.
nit2: are you aware of the `Chip` component?

front/components/assistant/conversation/input_bar/modelPickerUtils.ts
```diff
@@ -77,10 +79,14 @@ export interface ProviderGroup {
   models: { model: ModelConfigurationType; efforts: ReasoningEffort[] }[];
 }
 
-export type Selection =
+export type UserModelSelection =
   | { kind: "auto" }
   | { kind: "model"; model: ModelConfigurationType; effort: ReasoningEffort };
```
Fraggle: Why not re-use the auto model configuration here ?

front/components/assistant/conversation/input_bar/modelPickerUtils.ts
```diff
@@ -113,24 +120,35 @@ export function getModelWithReasoningEffortKey(
...
+    default:
+      assertNeverAndIgnore(selection);
+      return "";
   }
-
-  return `${model.displayName} ${capitalize(effort)}`;
 }
 
-// Converts the picker's local selection into the API model selection
 export function toModelSelection(
```
Fraggle: Could get rid of this.

front/components/assistant/conversation/input_bar/modelPickerUtils.ts
```diff
@@ -143,3 +161,51 @@ export function toModelSelection(
...
+  models: ModelConfigurationType[],
+  selection: { providerId: string; modelId: string }
+): ModelConfigurationType | undefined {
+  return models.find(
+    (m) =>
+      m.providerId === selection.providerId && m.modelId === selection.modelId
+  );
+}
+
+export function resolveDefaultSelection({
```
Fraggle, line 176: these 2 methods feels a bit redondant with what was added yesterday. Maybe you can reuse what we already have if we remove intermediaries and have auto use [the auto ModelConfigurationType](https://github.com/dust-tt/dust/blob/a85fa6743e58539ae78a3d1bc33a6d284feaff02/front/types/assistant/models/auto.ts#L4) ?

In general, feels like we could infer:
- auto because we use the auto model configuration
- agent because the model configuration is the same as the agent's
- mode because the model configuration is not the same as the agent's

Fraggle (approved):
Looks reasonnable :-)

front/components/assistant/conversation/AgentMessage.tsx
Nils-Fedrigo: added a tooltip

front/components/assistant/conversation/input_bar/modelPickerUtils.ts
Nils-Fedrigo: we need the kind to infer what is the state of the picker!

front/components/assistant/conversation/input_bar/modelPickerUtils.ts
```diff
@@ -112,34 +136,80 @@ export function getModelWithReasoningEffortKey(
...
-  return `${model.displayName} ${capitalize(effort)}`;
+function findAvailableModel(
+  models: ModelConfigurationType[],
+  selection: { providerId: string; modelId: string }
+): ModelConfigurationType | undefined {
+  return models.find(
+    (m) =>
+      m.providerId === selection.providerId && m.modelId === selection.modelId
+  );
 }
```
Nils-Fedrigo, line 175:
```suggestion

// `models` always includes the auto sentinel (AUTO_MODEL_CONFIG) whenever the
// models picker feature is on, so a plain lookup already resolves an "auto"
// selection — no need to special-case AUTO_MODEL_ID before looking it up.
```

front/components/assistant/conversation/input_bar/modelPickerUtils.ts
```diff
@@ -112,34 +136,80 @@ export function getModelWithReasoningEffortKey(
...
+  agentModel,
+  lastRequestedModel,
+  models,
+}: {
+  agentModel: AgentModelConfigurationType | null;
+  lastRequestedModel: ModelSelectionType | null;
+  models: ModelConfigurationType[];
+}): Selection {
+  if (lastRequestedModel) {
+    if (lastRequestedModel.modelId === AUTO_MODEL_ID) {
```
Nils-Fedrigo:
```suggestion
  const requestedModel = lastRequestedModel
    ? findAvailableModel(models, lastRequestedModel)
    : undefined;
  if (requestedModel) {
    if (requestedModel.modelId === AUTO_MODEL_ID) {
```

front/components/assistant/conversation/input_bar/modelPickerUtils.ts
```diff
@@ -112,34 +136,80 @@ export function getModelWithReasoningEffortKey(
...
-        reasoningEffort: selection.effort,
+        kind: "model",
+        model,
+        effort,
+        toSend: buildModelSelection(model, effort),
       };
-    default:
-      assertNeverAndIgnore(selection);
-      return undefined;
+    }
```
Nils-Fedrigo:
```suggestion
    const effort =
      lastRequestedModel?.reasoningEffort ??
      requestedModel.defaultReasoningEffort;
    return {
      kind: "model",
      model: requestedModel,
      effort,
      toSend: buildModelSelection(requestedModel, effort),
    };
```

front/components/assistant/conversation/input_bar/modelPickerUtils.ts
```diff
@@ -112,34 +136,80 @@ export function getModelWithReasoningEffortKey(
...
+  }
+
+  if (!agentModel || agentModel.modelId === AUTO_MODEL_ID) {
+    return { kind: "auto", toSend: AUTO_MODEL_SELECTION };
+  }
+  const model = findAvailableModel(models, agentModel);
+  if (!model) {
```
Nils-Fedrigo:
```suggestion
  const agentDefaultModel = agentModel
    ? findAvailableModel(models, agentModel)
    : undefined;
  if (!agentDefaultModel || agentDefaultModel.modelId === AUTO_MODEL_ID) {
```

front/components/assistant/conversation/input_bar/modelPickerUtils.ts
```diff
@@ -112,34 +136,80 @@ export function getModelWithReasoningEffortKey(
...
+    return { kind: "auto", toSend: AUTO_MODEL_SELECTION };
   }
+  return {
+    kind: "agent",
+    model,
+    effort: agentModel.reasoningEffort ?? model.defaultReasoningEffort,
```
Nils-Fedrigo:
```suggestion
    model: agentDefaultModel,
    effort:
      agentModel?.reasoningEffort ?? agentDefaultModel.defaultReasoningEffort,
```

front/components/assistant/conversation/input_bar/modelPickerUtils.ts
Nils-Fedrigo: It's a bit harder than that because we need to handle nicely the switch from agent to another so that it follows their config.

However I can try to make it work by 
- checking if same as agent first 
- then checking if auto 
- then checking model

front/components/assistant/conversation/input_bar/modelPickerUtils.ts
Fraggle: See my comment below about inferring the kind.

front/components/assistant/conversation/input_bar/modelPickerUtils.ts
Nils-Fedrigo: that's more or less what I do no?

Fraggle:
> that's more or less what I do no?

It's more like I'd love to avoid having another intermediary type and converting from one to the other and vice versa.
Same for the backend, I'd love to return to it a bit later and simplify the typing.

Nils-Fedrigo:
@Fraggle okok I propose to merge this and then we check together how to clean it further, I clearly spent way too much time turning this in every direction and probably need to discuss it live

Fraggle:
> @Fraggle okok I propose to merge this and then we check together how to clean it further, I clearly spent way too much time turning this in every direction and probably need to discuss it live

I already approved and I have zero problem if you merge and discard all my comments (for now or forever).
We can totally revisit later as well.

PR #28703, opened 2026-07-08
feat(assistant): default @dust to the auto model and refresh auto branding

Nils-Fedrigo:
## Description

Makes the `auto` model the default for @dust, refreshes the auto branding, and reworks the input-bar model picker around the agent's configured default.

- **@dust defaults to `auto`**: `_getDustGlobalAgent` now uses `AUTO_MODEL_CONFIG` as its preferred model, so Dust picks the best model per message. When `auto` isn't available for the workspace (no `models_picker`), the builder falls back to a whitelisted model as before. `deep-dive` also adds `AUTO_MODEL_CONFIG` to its candidate list.
- **Auto branding**: the `auto` model now renders the @dust agent avatar (`DUST_AVATAR_URL`) via a small `DustAvatarIcon` component instead of the generic Dust logo, its display name becomes `Auto: Dust handles model selection`, the tooltip is trimmed, and `light`/`medium`/`high` reasoning efforts are enabled.
- **Model picker rework**: the picker now derives what it shows from `resolveDefaultSelection` (agent's configured model + the last requested model in the conversation) rather than always starting on Auto. A manual user override is tracked separately and cleared when switching agents. A new "Default from agent" section surfaces the agent's configured model (and drops it from Suggested to avoid duplication).
- **Per-message model surfacing**: adds `modelResolutionMethod` (`"agent" | "user" | "auto"`) alongside `resolvedModel` on agent messages, plumbed through the types, the Sequelize model, message rendering, and the private Swagger schema. Agent messages now show a `with <model>` label when a per-message model was picked from the input bar.

## Tests

Typecheck passes; `global_agents`, `models`, and `enabled_models` suites pass, along with the message-rendering and stream-hook tests updated for the new field.

The picker UI, the per-message `with <model>` label, and the auto-avatar branding were not exercised end-to-end (require the `models_picker` flag and a live workspace) — worth a manual pass on a flagged workspace.

### DIFF
--- a/front/components/providers/DustAvatarIcon.tsx
+++ b/front/components/providers/DustAvatarIcon.tsx
@@ -1,0 +1,18 @@
+import { DUST_AVATAR_URL } from "@app/types/assistant/avatar";
+import { cn } from "@dust-tt/sparkle";
+
+interface DustAvatarIconProps {
+  className?: string;
+}
+
+// Renders the @dust agent's avatar as an icon so the "auto" model shows the same
+// visual as the Dust agent rather than the generic Dust logo.
+export function DustAvatarIcon({ className }: DustAvatarIconProps) {
+  return (
+    <img
+      src={DUST_AVATAR_URL}
+      alt=""
+      className={cn("rounded-full object-cover", className)}
+    />
+  );
+}
--- a/front/lib/api/assistant/global_agents/configurations/dust/dust.ts
+++ b/front/lib/api/assistant/global_agents/configurations/dust/dust.ts
@@ -48,6 +48,7 @@ import {
   CLAUDE_SONNET_4_6_DEFAULT_MODEL_CONFIG,
   CLAUDE_SONNET_5_DEFAULT_MODEL_CONFIG,
 } from "@app/types/assistant/models/anthropic";
+import { AUTO_MODEL_CONFIG } from "@app/types/assistant/models/auto";
 import { CUSTOM_MODEL_CONFIGS } from "@app/types/assistant/models/custom_models.generated";
 import {
   FIREWORKS_DEEPSEEK_V4_PRO_MODEL_CONFIG,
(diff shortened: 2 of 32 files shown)


Nils-Fedrigo:

- #28592
    - #28702
        - #28703 ◀

front/components/providers/DustAvatarIcon.tsx
```diff
@@ -0,0 +1,18 @@
...
+
+interface DustAvatarIconProps {
+  className?: string;
+}
+
+// Renders the @dust agent's avatar as an icon so the "auto" model shows the same
+// visual as the Dust agent rather than the generic Dust logo.
+export function DustAvatarIcon({ className }: DustAvatarIconProps) {
+  return (
+    <img
```
Fraggle, line 12: I'm not an expert on this but i think we have an Avatar component for this, no ?

front/lib/api/assistant/global_agents/configurations/dust/dust.ts
```diff
@@ -462,13 +463,13 @@ export function _getDustGlobalAgent(
   auth: Authenticator,
   args: DustLikeGlobalAgentArgs
 ): AgentConfigurationType | null {
+  // Default @dust to the "auto" model so Dust picks the best model per message.
+  // When "auto" isn't available for the workspace (no models_picker), the builder
```
Fraggle, line 467: what do you mean by "builder" here ?

front/lib/api/assistant/global_agents/configurations/dust/dust.ts
```diff
@@ -462,13 +463,13 @@ export function _getDustGlobalAgent(
...
+  // falls back to a whitelisted model.
   return _getDustLikeGlobalAgent(auth, args, {
     agentId: GLOBAL_AGENTS_SID.DUST,
     name: "dust",
-    preferredModelConfiguration: args.preferGpt55DefaultModel
```
Fraggle: we are completely ignoring preferGpt55DefaultModel right ?

Fraggle (approved):
LGTM but for that you likely want another pair of eyes. @spolu ?

PR #28704, opened 2026-07-08
fix(public-api): expose path-backed generated files in conversations

adrsimon:
## Description

The public API previously filtered out **path-backed** generated files (those with `fileId: null`) from agent message actions and stream events. These are files written by the agent file system tools, and dropping them meant API consumers never saw them and had no way to download them.

This PR:

- Keeps path-backed generated files in the public compatibility layer (`front/lib/api/v1/backward_compatibility.ts`) for both the conversation response and the v1 SSE stream, instead of stripping them out.
- Extends the SDK `ActionGeneratedFileSchema` to a union: a DB-backed file (`fileId: string`) or a path-backed file (`fileId: null` with `filePath`).
- Updates the conversation file download route to accept the canonical `conversation-{cId}/...` path that current file tools surface in `generatedFiles` (the legacy `conversation/...` form is still accepted), while rejecting canonical paths that target a different conversation (403).
- Updates the Swagger/OpenAPI schemas accordingly.

## Tests

Added regression coverage:

- Path-backed generated files (`fileId: null` + `filePath`) appear in public conversation responses (`.../[cId]/index.test.ts`).
- Canonical conversation-scoped file paths resolve and download correctly (`.../files/[...rel].test.ts`).
- A canonical path for another conversation is rejected with 403 (`.../files/[...rel].test.ts`).

All pre-commit checks (biome, typecheck, swagger annotations, docs check) pass.

## Risk

Low. The change is additive on the read/serialization path: previously-hidden files now surface, and the download route accepts an additional path form without loosening cross-conversation scoping (canonical paths for other conversations are explicitly rejected). Safe to roll back by reverting.

## Deploy Plan

Standard deploy. Ship `front`/`front-api` together; the SDK schema change is backward compatible (existing DB-backed files still validate).


aubin-tchoi (approved):
LGTM, @flvndvd no other changes expected here?

flvndvd:
> LGTM, @flvndvd no other changes expected here?

We are going to change the scoped path actually to align on `conversations/<cId>/<file>`

spolu:
@adrsimon looks like it broke main on connectors? 

https://app.dust.tt/w/0ec9852c2f/conversation/bSEJWvFMMV

Example build failures:

https://github.com/dust-tt/dust/actions/runs/29012787669/job/86100248838?pr=28712
https://github.com/dust-tt/dust/actions/runs/29012277867/job/86098632556

PR #28708, opened 2026-07-08
fix(front): clean requestedSpaceIds on all skills referencing a deleted space

adrsimon:
## Description

Fixes https://github.com/dust-tt/tasks/issues/9500

Deleting a space could leave skills and agents referencing the now-deleted space in their `requestedSpaceIds`, which hides them from every user (the visibility gate drops any agent whose requested spaces can't all be resolved).

Root cause: on deletion, affected skills were discovered **only** through the MCP server views / data source views located in the space. A skill can reference a space via `requestedSpaceIds` without holding a live view in it, so those skills were never cleaned. Because an agent's `requestedSpaceIds` is recomputed as the union of its own capabilities plus each attached skill's `requestedSpaceIds`, the stale space stuck to the skill and got re-applied to agents on the next recompute.

Changes:
- Discover affected skills directly by `requestedSpaceIds` as well (new `SkillResource.listByRequestedSpaceId`), mirroring the existing agent lookup, so the skill (the durable source) is always cleaned.
- Strip agents with a fresh in-transaction query after the skill loop instead of a snapshot captured before it, so agents whose refs change during skill cleanup are caught.
- Add a migration to strip a given space id from every `requestedSpaceIds` column (`agent_configurations`, `skill_configurations`, `skill_versions`) to repair rows already affected by past deletions.

## Tests

Typechecked the changed files. 

## Risk

Low.

## Deploy Plan

Deploy normally. Run the migration once as a dry run, confirm the affected counts, then `--execute` with the orphaned space ids to repair existing rows.

### DIFF
--- a/front/migrations/20260708_strip_space_from_requested_space_ids.ts
+++ b/front/migrations/20260708_strip_space_from_requested_space_ids.ts
@@ -1,0 +1,108 @@
+import { QueryTypes } from "sequelize";
+
+import { frontSequelize } from "@app/lib/resources/storage";
+import type { Logger } from "@app/logger/logger";
+import { makeScript } from "@app/scripts/helpers";
+
+// Every table that carries a `requestedSpaceIds` bigint[] column. `skill_versions`
+// inherits the column from `SkillConfigurationModel` via SKILL_MODEL_ATTRIBUTES, so
+// it must be included alongside `skill_configurations`.
+const TABLES = [
+  "agent_configurations",
+  "skill_configurations",
+  "skill_versions",
+] as const;
+
+type CountRow = { count: string };
+
+// True when the row's requestedSpaceIds overlaps the target spaces, i.e. it still
+// references at least one of them.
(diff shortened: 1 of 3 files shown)


aubin-tchoi (approved):
LGTM 🙏

front/migrations/20260708_strip_space_from_requested_space_ids.ts
```diff
@@ -0,0 +1,108 @@
...
+     SET "requestedSpaceIds" = ${requestedSpaceIdsWithoutTargets(spaceIdsLiteral)}
+     WHERE ${referencesTargetSpaces(spaceIdsLiteral)}`,
+    { type: QueryTypes.UPDATE }
+  );
+
+  logger.info(
+    { table, affected },
+    "Stripped target spaces from requestedSpaceIds"
+  );
+}
```
id13, line 89: I would wrap this in a transaction and do a select for update, just to be safe

id13 (approved):
LGTM

PR #28709, opened 2026-07-08
Pricing: correctly display price of message whe using ask user question

fabiencelier:
## Description

closes https://github.com/dust-tt/tasks/issues/9301

Credit cost is computed in the finalize activity after the terminal stream event is published, so when a message resumes after an ask_user graceful stop, the second run's success event pushes the first run's stored total into the client state: a non-null value that disabled the authoritative refetch, so the menu showed only run 1's cost until a page refresh.

This PR marks messages that complete a run live in this session: the cost fetch stays enabled for them (the streamed value can't be trusted), and the value is revalidated once per completion when the actions menu opens — by which point the recompute has landed.

Historical messages are unaffected and still open the menu with zero network calls.

## Tests

tested locally 

## Risk

low

## Deploy Plan

deploy front

### DIFF
--- a/front/components/assistant/conversation/AgentMessage.tsx
+++ b/front/components/assistant/conversation/AgentMessage.tsx
@@ -129,6 +129,7 @@ import {
   useContext,
   useEffect,
   useMemo,
+  useRef,
   useState,
 } from "react";
 import type { Components } from "react-markdown";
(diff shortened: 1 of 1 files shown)


front/components/assistant/conversation/AgentMessage.tsx
```diff
@@ -261,24 +261,44 @@ export function AgentMessage({
...
-  // Re-fetch (on menu open) only if a cost we want to show is still missing:
+  // Set once this message finishes a run live in this session (terminal stream
+  // event). Credits are computed in the finalize activity *after* the terminal
+  // event is published, so the streamed `costCredits` is never the authoritative
+  // total: it is `null` on a first run, and the *previous* run's total after an
+  // `ask_user` resume. So we keep the authoritative single-message fetch enabled
+  // and revalidate it when the menu opens (see below), rather than trusting the
+  // streamed value. Never reset: flipping `needsCostFetch` back to false would
+  // null the SWR key and revert the display to the stale streamed prop.
+  const [hasCompletedRunThisSession, setHasCompletedRunThisSession] =
```
Nils-Fedrigo, line 272: I'm not entirely sure why you need 2 different pointers  (hasCompletedRunThisSession and pendingCostRevalidationRef), but I trust you that nullifying needsCostFetch would then revert to the stale streamed prop. Maybe only one ref would do as this would not re-render state? This is possibly a hack

Nils-Fedrigo (approved):
thanks tons.

Left a clean code comment but quite a small one

front/components/assistant/conversation/AgentMessage.tsx
fabiencelier: not that easy to do it with a single 
if needsCostFetch becolme false then the whole SWR data become null and we use the data received initially so we canot have a single true/false state.
Claude managed to do it a with an interger incremented when we need a refresh but overal it is more complex

PR #28710, opened 2026-07-08
Add cancel subscription for free plans

tdraier:
## Description

Free-plan (`CP_FREE_PLAN`) workspaces had no way to cancel/end their subscription from the billing page — the "Cancel subscription" button was only shown for Metronome-billed Business plans (`isCancellablePlan` required `isBusinessPlanPrefix`). This extends the existing cancel flow to the free plan too.

Every `CP_FREE_PLAN` subscription is backed by a real Metronome contract (provisioned at signup via `provisionMetronomeContract`, or attached via the `contract.start` webhook), so the free plan goes through **the exact same cancellation flow** as `CP_BUSINESS_PLAN` — `cancelWorkspaceContractAtPeriodEnd` schedules the real contract to end via Metronome, and the existing `contract.end` webhook handler picks it up from there (including launching the workspace scrub), with no new code path needed for that part.

The only difference for the free plan: since nothing is being billed, there's no reason to wait for a billing period end — and free-plan contracts may not reliably have a draft invoice to look one up from anyway (no billed seat subscription generating one). So `cancelWorkspaceContractAtPeriodEnd` skips the draft-invoice lookup for the free plan and schedules the contract to end immediately (`now`) instead of at the current period's end.

- `FreePlanBilling.tsx`: renders `<SubscriptionActionButtons />` next to the plan name/status chip, same as the paid billing page.
- `SubscriptionContext.tsx`: `isCancellablePlan` now also returns `true` for Metronome-billed `CP_FREE_PLAN` subscriptions (previously required `isBusinessPlanPrefix`).
- `lib/metronome/contract_lifecycle.ts`: `cancelWorkspaceContractAtPeriodEnd` now branches only on how to compute the end date (`now` for free plan vs. current-period-end from Metronome draft invoices for Business) — everything downstream (`scheduleMetronomeContractEnd`, `markAsCanceled`) is shared.
- `SubscriptionActionButtons.tsx`: cancel dialog shows "Your subscription will end immediately" for the free plan instead of the period-end copy.
- `useMetronomeContractLifecycleAction.ts`: fixed the success toast, which previously always claimed "will end at the end of the period" regardless of plan — now it just confirms the cancellation happened.

No route/API changes: this reuses the existing `PATCH /api/w/[wId]/metronome/contract` (`action: "cancel"`) endpoint end-to-end.

## Tests

- `npx tsgo --noEmit` passes.
- `biome check` passes on all touched files.
- Manually not verified in a running app (no local Metronome-backed free-plan workspace at hand); verified by code path tracing, cross-checked against how the existing Business cancel flow reaches the `contract.end` webhook and scrub launch.

## Risk

- Low blast radius: gating change is additive, existing Business-plan cancel behavior and code path are unchanged.
- Relies on Metronome accepting an immediate (`now`) `ending_before` for `scheduleMetronomeContractEnd` — this is an already-supported default elsewhere in the codebase (e.g. `SubscriptionResource.endActiveSubscription`, contract-swap-on-plan-change), not a novel usage.
- Contract end is rounded up to the next hour boundary by Metronome (`ceilToHourISO`), so "immediate" cancellation may take up to ~1h to actually reflect via the webhook/scrub trigger — acceptable given nothing is billed in the meantime.

## Deploy Plan

### DIFF
--- a/front/components/workspace/billing/SubscriptionActionButtons.tsx
+++ b/front/components/workspace/billing/SubscriptionActionButtons.tsx
@@ -1,4 +1,5 @@
 import { useSubscriptionContext } from "@app/components/workspace/billing/SubscriptionContext";
+import { isCreditPricedFreePlan } from "@app/lib/plans/plan_codes";
 import { TRACKING_AREAS, withTracking } from "@app/lib/tracking";
 import {
   Button,
@@ -15,12 +16,16 @@ import {
 
 function CancelMetronomeSubscriptionDialog() {
   const {
+    subscription,
     periodEndLabel,
     isCancellingSubscription,
     cancelSubscription,
     showCancelDialog,
     setShowCancelDialog,
   } = useSubscriptionContext();
+  const isImmediateCancellation = isCreditPricedFreePlan(
+    subscription.plan.code
+  );
   // "July 12, 2026" → "July 12"
   const shortDate = periodEndLabel ? periodEndLabel.split(",")[0] : null;
 
@@ -37,7 +42,9 @@ function CancelMetronomeSubscriptionDialog() {
         <DialogHeader>
           <DialogTitle>Cancel your subscription</DialogTitle>
           <DialogDescription>
-            {periodEndLabel ? (
+            {isImmediateCancellation ? (
+              "Your subscription will end immediately."
+            ) : periodEndLabel ? (
               <>
                 Your plan will remain active until{" "}
                 <span className="font-bold">{periodEndLabel}</span>.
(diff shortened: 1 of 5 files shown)


tdraier:

- #28710 ◀

front/components/workspace/billing/SubscriptionActionButtons.tsx
```diff
@@ -52,6 +59,11 @@ function CancelMetronomeSubscriptionDialog() {
             <div className="flex justify-center py-8">
               <Spinner variant="dark" size="md" />
             </div>
+          ) : isImmediateCancellation ? (
+            <ContentMessage size="sm" variant="highlight">
+              This ends your subscription right away. You can resubscribe at any
+              time.
```
rfrenoy: Is it relevant to say they can resubscribe at any time? I would expect something closer to "This ends your subscription right away. This operation is irreversible".

rfrenoy (approved):
Non-blocking comment about working, otherwise LGTM

front/components/workspace/billing/SubscriptionActionButtons.tsx
tdraier: Yes .. actually they can only resubscribe with a paid plan.

PR #28712, opened 2026-07-08
Tool runner: make runToolWithStreaming tool-context dependent

spolu:
## Description

Step towards conversation-free tool execution (sandbox function invocations): the tool runner is now driven entirely by the tool context.

- `runToolWithStreaming(auth, { toolContext }, options)`: no more `action` / `agentConfiguration` / `model` / `agentMessage` / `conversation` / `userMessage` parameters. Inputs (`augmentedInputs` vs `inputs`), logger fields, and the initial `updateStatus("running")` (agent loop only — sandbox actions are created running) are resolved from the run context. `executeToolStreaming` keeps its signature and builds the context.
- `ToolExecution` is split into `AgentLoopToolExecution` (conversation-scoped identifiers) and `SandboxFunctionToolExecution` (`sandboxFunctionId` / `invocationId`) over a shared base, and all exit/pause events get both variants: `ToolPersonalAuthRequiredEvent`, `ToolFileAuthRequiredEvent`, `ToolAskUserQuestionEvent`, `ToolEarlyExitEvent`, `ToolPausedEvent`. Both variants share the single validation metadata type (`agentName` is the hardcoded `"agent"` placeholder from #28717, pending removal). The publish path (`AgentMessageEvents`, `DeferrableEvent`) and `BlockedToolExecution` stay pinned to the agent loop variants: no wire shape changes.
- `getExitOrPauseEvents(auth, { outputItems, toolContext })` builds the variant matching the run context (identifiers, tool call name, and inputs derived through an exhaustive switch). For auth and user-question exits, the agent-loop-only mutations (blocked statuses, resume step context, sandbox bash pause) are guarded but the event is returned in both contexts: sandbox function invocations observe the returned event instead. For `tool_blocked_awaiting_input` the blocked status and resume state are likewise persisted in agent loop contexts only; its emitters (sandbox bash, run_agent) are unavailable to sandbox functions today, and supporting run_agent there requires invocation-level pause/resume semantics first (TODO in place).
- `handleMCPActionError` accepts both action resources (all calls are polymorphic).

## Tests

`lib/actions/mcp_execution.test.ts`, `lib/actions/mcp_actions.test.ts`, `lib/actions/mcp_internal_actions/exit_events.test.ts`, `temporal/agent_loop/activities/common.test.ts` pass.

## Risk

Medium. The agent loop tool execution path is refactored in place: behavior is intended to be identical (same events, same payloads, same mutations), but this is the core tool execution flow. The sandbox function paths have no caller yet.

## Deploy Plan

- deploy `front`

### DIFF
--- a/front/lib/actions/mcp_internal_actions/events.ts
+++ b/front/lib/actions/mcp_internal_actions/events.ts
@@ -5,15 +5,12 @@ import type {
 import type { UserQuestion } from "@app/lib/actions/types";
 import type { OAuthProvider } from "@app/types/oauth/lib";
 
-export interface ToolExecution<
+interface ToolExecutionBase<
   T extends MCPValidationMetadataType = MCPValidationMetadataType,
 > {
-  conversationId: string;
-  messageId: string;
   actionId: string;
   // User might be undefined if the run was initiated through the public API using an API key.
   userId?: string;
-  configurationId: string;
   created: number;
 
   stake?: MCPToolStakeLevelType;
@@ -28,6 +25,24 @@ export interface ToolExecution<
   approvalArgsLabel?: string;
 }
 
+// Tool execution scoped to an agent loop run: carries the conversation-scoped identifiers used on
+// the conversation message channel.
+export interface AgentLoopToolExecution<
+  T extends MCPValidationMetadataType = MCPValidationMetadataType,
+> extends ToolExecutionBase<T> {
+  conversationId: string;
+  messageId: string;
+  configurationId: string;
+}
+
+// Tool execution scoped to a sandbox function invocation.
+export interface SandboxFunctionToolExecution<
+  T extends MCPValidationMetadataType = MCPValidationMetadataType,
+> extends ToolExecutionBase<T> {
+  sandboxFunctionId: string;
+  invocationId: string;
+}
+
 type ToolPersonalAuthError = {
   mcpServerId: string;
   provider: OAuthProvider;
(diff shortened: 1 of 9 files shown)


spolu:
(connectors error unrelated being worked on here: https://dust4ai.slack.com/archives/C050SM8NSPK/p1783594802312989)

spolu:
Asking 3 reviewers because this one is a bit finicky.

r? @flvndvd  types any better idea?
r? @adrsimon  exit_events in particular the fact that the assert is rightfully an assert: we want to support calling a sub-agent as a tool from a sandbox function so we need this file logic in general but want to confirm tool_blocked_awaiting_input is sandbox bash specific and hence not applicable outside of an agent loop
r? @PopDaph  as master reviewer :pray:

spolu:
(assuming @adrsimon is a GO as per IRL)

front/lib/actions/mcp_internal_actions/events.ts
```diff
@@ -60,37 +85,65 @@ type ToolFileAuthError = {
...
-  configurationId: string;
-  messageId: string;
-  conversationId: string;
   text: string;
   isError: boolean;
   reason?: "deploy_interruption" | "user_cancellation" | "none";
 };
 
+export type AgentLoopToolEarlyExitEvent = ToolEarlyExitEventBase & {
+  configurationId: string;
```
PopDaph: To remove some duplication, do we want a base event type for agent loop containing configurationId/messageId. conversationId and a base event type for sbx fn with sandboxFunctionId/invocationId?

front/lib/actions/mcp_internal_actions/events.ts
```diff
@@ -60,37 +85,65 @@ type ToolFileAuthError = {
...
 }
 
-export interface ToolAskUserQuestionEvent extends ToolExecution {
+export interface AgentLoopToolAskUserQuestionEvent
+  extends AgentLoopToolExecution {
+  type: "tool_ask_user_question";
+  question: UserQuestion;
+}
+
+export interface SandboxFunctionToolAskUserQuestionEvent
```
flvndvd, line 122: Do we plan to support this one in sandbox?

front/lib/actions/mcp_internal_actions/events.ts
```diff
@@ -60,37 +85,65 @@ type ToolFileAuthError = {
...
 // Pauses agent execution to prompt user for file access consent (e.g., Google Drive).
 // Non-terminal because the tool can resume once the user authorizes the file.
-export interface ToolFileAuthRequiredEvent
-  extends ToolExecution<
-    MCPValidationMetadataType & {
-      mcpServerId: string;
-      mcpServerDisplayName: string;
-    }
-  > {
+export interface AgentLoopToolFileAuthRequiredEvent
```
flvndvd, line 96: Small DRY suggestion: the two "scope" shapes (agent-loop's configurationId/messageId/conversationId vs sandbox-function's sandboxFunctionId/invocationId) exist 4 times in this file. We could pull them into two named types so we have one definition for each, and mirrors the `contextType` pattern already used for `AgentLoopRunContextType`/`SandboxFunctionRunContextType`:

```
export type AgentLoopEventScope = {
  configurationId: string;
  messageId: string;
  conversationId: string;
};

export type SandboxFunctionEventScope = {
  sandboxFunctionId: string;
  invocationId: string;
};
```

PopDaph (commented):
LGTM 🙏🏻

front/lib/actions/mcp_internal_actions/events.ts
spolu: done

front/lib/actions/mcp_internal_actions/events.ts
spolu: Did the above to dry, almost equivalent 👍

front/lib/actions/mcp_internal_actions/events.ts
spolu: I told you no but eventually I think yes possibly so keeping it 👍

PR #28717, opened 2026-07-09
Deprecate agentName in tool validation metadata

spolu:
## Description

First step of removing `agentName` from tool validation metadata (`MCPValidationMetadataType`) which is conversation specific in a world where tool lifecycle should not.

This means we won't show agent name in confirmation windows in product but I believe this is fine loss for more generic events.

The field is part of the public event schemas (`tool_approve_execution`, personal/file auth, ask user question) and already-deployed CLI and extension versions require it at parse time, so removal is staged:

- `front`: `agentName` is typed as the literal `"agent"`, forcing every emit site to the placeholder constant while the wire keeps satisfying old clients' required-string parsing. All emit sites updated accordingly.
- `sdks/js`: `agentName` becomes optional in `MCPValidationMetadataSchema` (kept `z.string()`, not a literal, so events published before the deploy and replayed from Redis still parse).
- All reader sites will survive disappearance of the field from the server once this code is live: web approval dialogs (`MCPToolValidationRequired`, `ToolValidationDetails`), the CLI approval prompt and approval cache keys (approvals are no longer scoped per agent), the Slack/Teams connectors' approval and auth cards and the extension.

Note: `RunAgentBlockingEvent` is now fully client-typed (the events are collected from the child agent stream parsed with the public client schemas and serialized back into a client-shaped resource); `getExitOrPauseEvents` normalizes them to the internal shape when forwarding.

We will be able to drop agentName entirely in a couple days once the clients are all cycled out and in the meantime we'll have the hard-coded `agentName: "agent"` even in non conversation-tied tool execution related events.

Obviously this could break external clients subscribed to these events and exepecting an agentName in the future once we remove it.... but probability is low IMHO. No breakage possible for now.

## Tests

- `lib/resources/agent_mcp_action_resource.test.ts`, `lib/actions/mcp_internal_actions/exit_events.test.ts`, `lib/api/assistant/email/email_trigger.test.ts`, `lib/actions/mcp_execution.test.ts` pass. Typecheck across `front`, `sdks/js`, `cli`, `connectors`.
- tested locally

## Risk

Low. The field is still emitted (as `"agent"`), so old clients keep parsing; only display strings change ("agent" instead of the agent name in approval dialogs, CLI prompts, and Slack/Teams cards), and CLI cached approvals are now keyed per tool instead of per agent-tool pair.

## Deploy Plan

- deploy `front`
- deploy `connectors`
- publish `@dust-tt/client`, release the CLI, and publish the extension (so clients tolerant of the field's absence start cycling in)

### DIFF
--- a/cli/dust-cli/src/ui/commands/Chat.tsx
+++ b/cli/dust-cli/src/ui/commands/Chat.tsx
@@ -2052,7 +2052,7 @@ const CliChat: FC<CliChatProps> = ({
                   : inlineSelector.mode === "approval" &&
                       pendingApproval &&
                       pendingApproval.type === "tool_approve_execution"
-                    ? `Tool Approval Required: agent wants to use ${pendingApproval.metadata.toolName}, what do you want to do? `
+                    ? `Tool Approval Required: the agent wants to use ${pendingApproval.metadata.toolName}, what do you want to do? `
                     : inlineSelector.mode === "diff" && pendingDiffApproval
                       ? `Changes Preview: ${pendingDiffApproval.filePath} `
                       : mentionPrefix
--- a/cli/dust-cli/src/ui/commands/Chat.tsx
+++ b/cli/dust-cli/src/ui/commands/Chat.tsx
@@ -433,7 +433,7 @@ const CliChat: FC<CliChatProps> = ({
       // For low stake tools, check cache first
       if (event.stake === "low") {
         const cachedApproval = await toolsCache.getCachedApproval({
-          agentName: event.metadata.agentName,
+          agentName: "agent",
           mcpServerName: event.metadata.mcpServerName,
           toolName: event.metadata.toolName,
         });
@@ -490,7 +490,7 @@ const CliChat: FC<CliChatProps> = ({
         // Cache the approval if requested and it's a low stake tool
         if (cacheApproval && pendingApproval.stake === "low") {
           await toolsCache.setCachedApproval({
-            agentName: pendingApproval.metadata.agentName,
+            agentName: "agent",
             mcpServerName: pendingApproval.metadata.mcpServerName,
             toolName: pendingApproval.metadata.toolName,
           });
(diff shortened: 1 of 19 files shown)


cli/dust-cli/src/ui/commands/Chat.tsx
```diff
@@ -2055,7 +2052,7 @@ const CliChat: FC<CliChatProps> = ({
                   : inlineSelector.mode === "approval" &&
                       pendingApproval &&
                       pendingApproval.type === "tool_approve_execution"
-                    ? `Tool Approval Required: ${pendingApproval.metadata.agentName} wants to use ${pendingApproval.metadata.toolName}, what do you want to do? `
+                    ? `Tool Approval Required: agent wants to use ${pendingApproval.metadata.toolName}, what do you want to do? `
```
PopDaph: Here (there) and everywhere wouldn't it be much better to use "the agent" and not "agent"? 🙏🏻

spolu:
Example impact:

Screenshot 2026-07-09 at 12 18 48

Diff is hardly noticeable IMHO

front/components/assistant/conversation/tool_validation/PodTasksUpdateValidationDetails.tsx
```diff
@@ -273,9 +266,7 @@ function TaskUpdateRow({
                   Marked done by
                 </span>
                 <span className="text-sm font-medium text-foreground">
-                  {taskInput.markAsDoneByType === "user"
-                    ? "You"
-                    : `@${agentName}`}
+                  {taskInput.markAsDoneByType === "user" ? "you" : "agent"}
```
PopDaph, line 269: Here and below we are loosing the first letter in uppercase 🙏🏻

PopDaph (commented):
I would change all the wordings before merging to have more natural phrasing. The bare 'agent' reads like a placeholder where a name used to be, which is exactly what it is :D

Also, pit of success, I would already remove this line: connectors/src/connectors/slack/chat/stream_conversation_handler.ts:521

PopDaph (approved):
Nothing is blocker so approving already 🙇🏻‍♀️

front/components/assistant/conversation/tool_validation/PodTasksUpdateValidationDetails.tsx
spolu: the uppercase was a mistake as far as I can tell?

cli/dust-cli/src/ui/commands/Chat.tsx
spolu: Yep this one can be tweaked

spolu: (Also this is the CLI :))

spolu: (changed most instances 👍 )

PR #28718, opened 2026-07-09
[dsbx] feat: db query runs DML, refuses DDL

id13:
db query moves from read-only to SELECT + DML
(INSERT/UPDATE/DELETE/REPLACE, optionally WITH): the database opens
read-write must-exist, the raw statement's first keyword is gated BEFORE
sqlite3_prepare (some pragmas take effect at compile time), and every
write runs in one BEGIN IMMEDIATE transaction with a schema_version
re-check as the second barrier — DDL, PRAGMA, ATTACH, transaction control
and VACUUM are refused, so the schema still only evolves through
reconcile and the WAL journal mode litestream depends on cannot be broken
from a query. Plain DML reports the affected-row count in a new
`changes` envelope field; RETURNING flows through the spill-capped row
path. queryReadonly is renamed runQuery accordingly.

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>

### DIFF
--- a/cli/dust-sandbox/functions-runner/db/commands.test.ts
+++ b/cli/dust-sandbox/functions-runner/db/commands.test.ts
@@ -10,7 +10,7 @@ import { DbCommandError } from "./common.ts";
 import {
   QUERY_INLINE_PAYLOAD_CAP_BYTES,
   QUERY_INLINE_ROW_CAP,
-  queryReadonly,
+  runQuery,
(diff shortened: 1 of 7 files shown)


id13:

- #28611
    - #28613
        - #28614
            - #28599
                - #28718 ◀
                    - #28600
                        - #28720
                            - #28601

PR #28618, opened 2026-07-07
[front] feat: surface pod databases in sandbox_functions tools

id13:
## Description

**Stack 9/10 (pod-state).** Surfaces pod databases in the `sandbox_functions` tools and completes the publish narrative (design "MCP 1 + 3", after the pod_databases server itself was dropped by design decision — the manifest data carries the same information).

- `get`/`list` now show each function's declared databases from its stored manifest.
- The `publish` tool response carries the gate's story: **mode-drift / unique-tightening warnings**, **stale-sibling notes** ("published against an older schema … republish when convenient"), and an **untracked-leftover note** (live databases no function declares anymore). Untracked computation is deliberately cheap: one `dsbx db list` after store (via a new `listDatabasesOnSandbox` exec helper — same shellEscape/`--`/agent-proxied/envelope pattern as reconcile), only on database-declaring publishes, best-effort — a listing failure logs a warning and never fails the publish.
- `destructive_change` reconcile errors gain a recovery hint for the reconciled-but-store-failed drift case (live schema ahead of stored manifests → republish the wider-schema function).
- New publish error codes surface `tracked:false` (compat blocks, reconcile refusals, publish conflicts are agent-correctable); publish tool description rewritten for the databases contract (descriptions feed the tool-search corpus).

Note: the sandbox-side `dsbx db schema` and `db query` subcommands (previous PRs) currently have **no front consumer** — kept as sandbox-side plumbing per the scope ruling.

## Tests

`tsgo` clean at this commit; `npm run test -- lib/api/actions/servers/sandbox_functions lib/api/sandbox_functions/publish_sandbox_function.test.ts` — 17/17 (get/list database lines, untracked-leftover outcome, note rendering paths).

## Risk

Low: response-shape additions to existing text tools; no behavior change for functions without databases.

## Deploy Plan

Standard front deploy, after the publish-flow PR below it.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #28719, opened 2026-07-09
[front] feat: drop stored schemas and the compat differ — reconcile is the gate

id13:
The per-function database shapes were mirrored from the build envelope,
stored verbatim in a sandbox_functions.manifests JSONB column, and diffed
against every sibling's stored copy at publish (compat.ts). All of that
goes: the schemas' single source of truth is the authored
databases/{db}.db.ts files plus the live SQLite file, and the publish
gate is solely the in-sandbox `dsbx db reconcile` refusal of destructive
or non-additive DDL. Front now parses only the declared database names +
schemaFile paths from the build envelope (the reconcile inputs; names are
still regex- and reserved-key-checked since the envelope comes back from
the sandbox where model-authored code runs).

Consequences: the manifests column, its migration (never deployed) and
the resource/model plumbing are deleted; compat_blocked and the
warnings/stale-sibling/untracked-database publish notes disappear
(convention now lives in the pod_functions skill text); list/get lose the
stored databases display; dsbx db list loses its only front consumer.

The publish flow keeps its Redis lock and gets the review fixes that were
entangled here: executeWithLock throws a typed
LockAcquisitionTimeoutError (a Redis outage no longer masquerades as a
retryable publish_conflict), lock.ts exports LOCK_TTL_MS instead of the
mirrored constant, updateContent claims the row with a conditional
UPDATE ... WHERE updatedAt = expected BEFORE overwriting the live bundle
(the first-publish path re-checks before creating the file so a lost race
is a typed conflict, not a unique-index 500 plus an orphaned bundle),
applied reconcile DDL is logged, and build's envelope parse reuses
parseDbEnvelope.

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>


id13:

- #28611
    - #28612
        - #28613
            - #28614
                - #28599
                    - #28718
                        - #28615
                            - #28616
                                - #28600
                                    - #28618
                                        - #28719 ◀
                                            - #28720
                                                - #28601

PR #28734, opened 2026-07-09
[front] fix: defer eager skill tools

aubin-tchoi:
## Description

The `bash` tool was made eager in https://github.com/dust-tt/dust/pull/28661 and always added to Dust-like agents.
We are seeing examples of custom agents using the Computer skill and bursting the cache due to tools changes (example trace [here](https://langfuse.dust.tt/project/cmi4ft2qm00061707q4a9azyc/traces?filter=metadata%3BstringObject%3BconversationId%3B%3D%3BhWzzLtMKO4&peek=d80163aa0828b8c62293d7e5b3ef8132&timestamp=2026-07-09T07%3A08%3A27.626Z))

This PR makes skill-provided tools stay deferred even when their server metadata marks the tool as eager.
Agent-configured tools keep their existing eager promotion behavior (but you can't add bash to an agent).

## Tests

- Added a test.

## Risk

- Low.

## Deploy Plan

- Deploy front.

### DIFF
--- a/front/lib/actions/mcp_actions.ts
+++ b/front/lib/actions/mcp_actions.ts
@@ -959,16 +959,6 @@ export async function disambiguateServerNamesBySpace(
  * Deduplicates MCP server configurations by view ID and name.
  * Priority order: agent actions > client-side > skill servers > JIT servers.
  */
-function getMCPServerConfigurationDeduplicationKey(
-  config: MCPServerConfigurationType
-): string {
-  const viewId = isServerSideMCPServerConfiguration(config)
-    ? config.mcpServerViewId
-    : config.clientSideMcpServerId;
-
-  return `${viewId}:${slugify(config.name)}`;
-}
-
 export function deduplicateMCPServerConfigurations({
   agentActions,
   clientSideActions,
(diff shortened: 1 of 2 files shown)


front/lib/actions/mcp_actions.ts
```diff
@@ -1018,11 +1018,24 @@ export async function tryListMCPTools(
...
     jitServers,
   });
+  const nonSkillServers = [
+    ...agentLoopListToolsContext.agentConfiguration.actions,
+    ...(agentLoopListToolsContext.clientSideActionConfigurations ?? []),
+    ...jitServers,
+  ];
+  const isSkillServerConfig = deduplicatedConfigs.map(
+    (config) =>
+      skillServers.includes(config) && !nonSkillServers.includes(config)
```
flvndvd: Using object reference for comparison? Are we sure about this? Can't we use specific fields?

flvndvd (approved):
LGTM

front/lib/actions/mcp_actions.ts
aubin-tchoi: figured that was very local and pointer comparison works but indeed it does look brittle, fixing

PR #28740, opened 2026-07-09
feat: check feature flag when inviting business admins

matteotrab:
## Description

This PR gates the assignment of the `business_admin` role during invitations behind the `admin_governance` feature flag.

- In the invitation update endpoint (`[iId]/index.ts`), a check is added: if the requested `initialRole` is `business_admin` and the `admin_governance` feature flag is not enabled for the workspace, the API returns a `403 workspace_auth_error`.
- The same feature flag import and guard is added to the invitation creation endpoint (`invitations/index.ts`), preventing new invitations with the `business_admin` role from being sent if the flag is absent.

Note that it was already not possible from the UI to assign the business admin role without the feature flag. This adds a check on the endpoints.

## Tests

Manually.

## Risks

Low — the change only adds a guard around an existing code path. Workspaces without the `admin_governance` flag will no longer be able to assign the `business_admin` role via invitations, which is the intended behaviour.

## Deploy Plan

Standard deployment.

### DIFF
--- a/front-api/routes/w/[wId]/invitations/[iId]/index.ts
+++ b/front-api/routes/w/[wId]/invitations/[iId]/index.ts
@@ -3,6 +3,7 @@ import {
   emitAuditLogEvent,
   getAuditLogContext,
 } from "@app/lib/api/audit/workos_audit";
+import { getFeatureFlags } from "@app/lib/auth";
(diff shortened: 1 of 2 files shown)


matteotrab:

- #28740 ◀

ykmsd (approved):

PR #28741, opened 2026-07-09
Sandbox functions: run tools through runToolWithStreaming instead of tryCallMCPTool

PopDaph:
Follow up of #28712, which made `runToolWithStreaming` tool-context dependent. This PR makes `runSandboxFunctionToolActivity` actually use it: the activity was still hand-rolling the orchestration (`tryCallMCPTool` + notification persistence + heartbeat-wrapped `processToolResults` + status transitions), duplicating the shared layer. It now fetches the action and invocation, builds the run context, and drains `runToolWithStreaming` (no invocation event stream to forward events to yet).

One behavior this PR must own: for a sandbox function context, `getExitOrPauseEvents` emits invocation-scoped events without setting a status (there is no pause surface for invocations yet), so a pause resource (personal auth required, blocked awaiting input, early exit) would leave the action `running` with the poll hanging until token expiry. The activity fails closed: any run ending without a terminal status marks the action errored, with the pause resource persisted in the output so the function sees what the tool needs and handles it on its side. Approval bubbling will replace this with `blocked_validation_required` + invocation stream events.

Also:
- The activity passes the temporal cancellation + shutdown signal to the run, like the agent loop activity.
- `markAsSucceeded` / `markAsErrored` on the sandbox action resource now round `executionDurationMs` like the agent resource does: the shared layer passes floats into an INTEGER column (the old activity rounded at call sites).
- On worker shutdown, `retry_on_interrupt` tool throws are caught and mark the action errored: the workflow runs with `maximumAttempts: 1`, so a rethrow would fail the workflow and leave the action `running`. Dated TODO to honor the policy with a retryable proxy like the agent loop workflow.

## Tests

New: `lib/api/mcp/run_tool.test.ts` drives `runToolWithStreaming` with a sandbox function run context through the real persistence path (only `tryCallMCPTool` mocked): success and tool error. New: `temporal/agent_loop/activities/run_sandbox_function_tool.test.ts` covers the activity end-to-end (temporal context mocked): normal success, fail-closed on personal-auth-required with the resource persisted, fail-closed on early exit. Existing: `mcp_execution` and all agent-loop temporal suites pass.

## Risk

The agent loop path is untouched (the shared layer landed in #28712). Worst case, sandbox function tool calls fail. Safe to rollback.

## Deploy Plan

- [ ] Deploy front

### DIFF
--- a/front/lib/api/mcp/run_tool.test.ts
+++ b/front/lib/api/mcp/run_tool.test.ts
@@ -1,0 +1,130 @@
+import { tryCallMCPTool } from "@app/lib/actions/mcp_actions";
+import type { SandboxFunctionRunContext } from "@app/lib/actions/types";
+import { runToolWithStreaming } from "@app/lib/api/mcp/run_tool";
+import { InternalMCPServerInMemoryResource } from "@app/lib/resources/internal_mcp_server_in_memory_resource";
+import { SandboxFunctionMCPActionResource } from "@app/lib/resources/sandbox_function_mcp_action_resource";
+import { MCPServerViewFactory } from "@app/tests/utils/MCPServerViewFactory";
+import { fileStorageMock } from "@app/tests/utils/mocks/file_storage";
+import { SandboxFunctionMCPActionFactory } from "@app/tests/utils/SandboxFunctionMCPActionFactory";
+import { createPersistedSandboxFunctionInvocationTokenTestContext } from "@app/tests/utils/SandboxTokenFactory";
+import type { CallToolResult } from "@modelcontextprotocol/sdk/types.js";
+import { describe, expect, it, vi } from "vitest";
+
+vi.mock("@app/lib/actions/mcp_actions", async (importOriginal) => {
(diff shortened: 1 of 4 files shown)


front/lib/api/mcp/run_tool.ts
```diff
@@ -69,45 +58,49 @@ export async function* runToolWithStreaming(
...
+  );
 
-  await action.updateStatus("running");
+  // Sandbox function actions are created `running`; agent actions transition here.
+  let inputs: Record<string, unknown>;
+  if (runContext.contextType === "agent_loop") {
+    await runContext.action.updateStatus("running");
+    inputs = runContext.action.augmentedInputs;
+  } else {
+    inputs = runContext.action.inputs;
```
PopDaph: 🙈

spolu (approved):
LGTM after rebase on https://github.com/dust-tt/dust/pull/28712

spolu (approved):
LGTM

PR #28742, opened 2026-07-09
Point the Pods skill at pod_manager's semantic search tool

flvndvd:
## Description

See Slack thread:  [here](https://dust4ai.slack.com/archives/C0BCDAL5ZTM/p1783588699670459) 

The Pods skill instructions told agents to check the Pod overview, read files directly, or fall back to company-wide search, but never named the semantic search tool that `pod_manager` exposes for querying Pod files, linked context nodes, and conversation transcripts by meaning. Because that tool is loaded lazily through tool search rather than always present in context, an agent had no way to discover it existed unless it happened to guess the right search terms. This adds an explicit step naming the tool and describing when to reach for it.

### DIFF
--- a/front/lib/resources/skill/code_defined/global/projects.ts
+++ b/front/lib/resources/skill/code_defined/global/projects.ts
@@ -1,6 +1,10 @@
 import { SEARCH_SERVER_NAME } from "@app/lib/actions/mcp_internal_actions/constants";
+import { getPrefixedToolName } from "@app/lib/actions/tool_name_utils";
 import { FILES_SERVER_NAME } from "@app/lib/api/actions/servers/files/metadata";
-import { POD_MANAGER_SERVER_NAME } from "@app/lib/api/actions/servers/pod_manager/metadata";
+import {
+  POD_MANAGER_SERVER_NAME,
+  SEMANTIC_SEARCH_TOOL_NAME,
+} from "@app/lib/api/actions/servers/pod_manager/metadata";
 import {
   formatPodAgentsMdPromptSection,
   readPodAgentsMdContent,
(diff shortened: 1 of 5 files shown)


front/lib/resources/skill/code_defined/global/projects.ts
```diff
@@ -57,12 +61,13 @@ Use the \`sId\` from \`pod_tasks\` tools (e.g. \`list_tasks\`, \`create_tasks\`)
 
 When you need to find information, use this order (skip steps if the relevant tools are not in your tool list):
 1. **Pod overview**: \`${POD_MANAGER_SERVER_NAME}\` \`get_information\` returns the Pod URL, description, and what is attached to the Pod.
-2. **Pod files**: read and search \`pod-{podId}/<rel>\` files through the sandbox or the \`${FILES_SERVER_NAME}\` MCP tools.
-3. **Company-wide**: If still insufficient, use \`company_data_*\` tools and \`${SEARCH_SERVER_NAME}\` for broader company data sources.
+2. **Topic or question**: \`${getPrefixedToolName(POD_MANAGER_SERVER_NAME, SEMANTIC_SEARCH_TOOL_NAME)}\` searches Pod files, linked context nodes, and Pod conversation transcripts by meaning. Use it when you have a topic or question rather than a specific file path.
```
aubin-tchoi: is "context nodes" a term that is used elsewhere?

aubin-tchoi (approved):
LGTM

front/lib/resources/skill/code_defined/global/projects.ts
aubin-tchoi: updated

Fraggle (commented):
thanks !

PR #28744, opened 2026-07-09
fix(sparkle): right-align NavigationListItem suffix

spolu:
## Description

The wake-up indicator in the conversation list sidebar sat mid-row (right after the title) and disappeared on hover when the more-menu button appears, which read as a glitch.

This right-aligns the `NavigationListItem` suffix so it sits flush against the right edge — exactly where the more-menu button shows up — so the hover swap reads as an intentional replacement. Also right-aligns the "Archived" chip on pods in sidebar search results (the only other suffix usage), which reads better there too.

Screenshot 2026-07-09 at 12 44 18

## Tests

N/A, tested locally.
Available in SPA preview below

## Risk

Low. One-line CSS change scoped to the `NavigationListItem` suffix container.

## Deploy Plan

- publish `sparkle`
- deploy `front`

### DIFF
--- a/sparkle/src/components/NavigationList.tsx
+++ b/sparkle/src/components/NavigationList.tsx
@@ -173,7 +173,7 @@ const NavigationListItem = React.forwardRef<
             {suffix && (
               <div
                 className={cn(
-                  "flex grow flex-shrink-0 items-center",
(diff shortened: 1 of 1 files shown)


ykmsd (approved):
👍

PR #28749, opened 2026-07-09
Sandbox functions: scope pod tool output files by invocation

PopDaph:
Follow up of #28664 / #28666. 

Tools called from sandbox function invocations offload their outputs (registered resources, oversized text) to the pod's `.tool_outputs/` folder, which the function reads through the `/files/pod-{pId}` mount. That folder was flat: a pod outlives its invocations, so outputs from successive or concurrent invocations mixed together and a function could not tell its own apart.

Offloads from a sandbox function run context now land under `.tool_outputs/{invocationId}/`, visible in-sandbox at `/files/pod-{pId}/.tool_outputs/{invocationId}/`. The output blocks keep carrying the scoped path, so the function reads exactly its own files. A fresh per-invocation directory also avoids gcsfuse negative-stat-cache staleness on repeated paths. Conversation flavor unchanged. User-facing generated files (images, blobs) keep landing at the pod root per `writeToPodFolder`'s convention.

Note: Pre-existing flat files under `.tool_outputs/` are left in place, no move or backfill: readers use the path embedded in their output blocks, so old blocks keep resolving. The flat folder just stops accumulating.

Drive-by 🚗: `createPersistedSandboxFunctionInvocationTokenTestContext` created the pod space with no members, so `DustFileSystem.forPod` silently failed `canRead` under test and offloads degraded to a "Failed to save the tool output" block, making the offload path effectively untestable. The user now joins the pod editor group and the auth is refreshed, matching production where the invocation-token auth is granted the pod space groups.

## Tests

New `mcp_execution` test asserting the invocation-scoped path for an offloaded resource block in a sandbox function run context. Existing: `mcp_execution` (12) and all front-api sandbox suites (41) pass with the factory change.

## Risk

Path change only affects the sandbox function flavor, which shipped days ago; readers use the path embedded in the output blocks, so no stored data references break in either direction. Worst case, functions look for outputs at the old flat path. Safe to rollback.

## Deploy Plan

- [ ] Deploy front

### DIFF
--- a/front/lib/actions/mcp_execution.test.ts
+++ b/front/lib/actions/mcp_execution.test.ts
@@ -478,7 +478,7 @@ describe("processToolResults", () => {
   });
 
   async function setupSandboxFunctionTest() {
-    const { auth, workspace, invocation, globalSpace } =
+    const { auth, workspace, invocation, globalSpace, podSpace } =
       await createPersistedSandboxFunctionInvocationTokenTestContext();
     const server = await InternalMCPServerInMemoryResource.makeNew(auth, {
       name: "common_utilities",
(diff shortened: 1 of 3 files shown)


spolu (approved):
@flvndvd was team no `invoicationId` directory because they are long (esp in EU) and high entropy which can be error prone / slow for the model to generate as argument to follow-up actions.

I'm quite neutral.

PopDaph:
@flvndvd I'll not merge that if you are against it, so let me know what you think 🙏🏻

flvndvd (approved):
LGTM

flvndvd:
> @flvndvd was team no `invoicationId` directory because they are long (esp in EU) and high entropy which can be error prone / slow for the model to generate as argument to follow-up actions.
> 
> I'm quite neutral.

I guess we don't have any other unique identifier here, right?

flvndvd:
> > @flvndvd was team no `invoicationId` directory because they are long (esp in EU) and high entropy which can be error prone / slow for the model to generate as argument to follow-up actions.
> > I'm quite neutral.
> 
> I guess we don't have any other unique identifier here, right?

Seen IRL, we can use the pod function slug.

spolu:
> Seen IRL, we can use the pod function slug.

Not bad

spolu:
\o/

PR #28751, opened 2026-07-09
Make previous-interactions pruning budget-driven instead of positional

flvndvd:
## Description

Anthropic prompt caching was missing billions of tokens, and the root cause traced back to how we render previous interactions on every LLM call. The existing logic always redacted tool results **beyond the last three interactions** regardless of how much context budget was actually available, and a separate selection step could drop whole interactions based on how big the current turn's own new content happened to be. Both of these are driven by a moving reference point that shifts on every single turn, which means the request sent to the model changes in ways that break the cached prefix almost every time, even in conversations nowhere near their context limit.

The fix reworks pruning so previous interactions get a fixed budget claim that never depends on the size of the newest turn, and redaction is now driven by actual token pressure rather than position, with decisions snapped to periodic checkpoints so the same content survives unchanged across many consecutive turns instead of shifting each time. The current turn absorbs whatever budget is left, which is safe since it's new content that was never cached to begin with. A safety fallback still exists for the rare case where even the protected recent interactions don't fit, but it no longer fires as the everyday mechanism.

This does change behavior in one visible way worth calling out: previously, anything past the last three interactions was always stripped down regardless of how much room was left. Now that only happens when the budget genuinely requires it, so in conversations with headroom the model will see more of its own recent tool history in full than it did before.

### DIFF
--- a/front/lib/api/assistant/conversation_rendering/pruning.ts
+++ b/front/lib/api/assistant/conversation_rendering/pruning.ts
@@ -203,12 +203,17 @@ export function prunePreviousInteractions(
       let effectiveFrontier = neededFrontier;
       for (let i = neededFrontier; i < floorStart; i++) {
         effectiveFrontier = i;
+        if (i === 0) {
+          // Index 0 is trivially always a checkpoint: it's the start of the conversation.
+          break;
+        }
         const bucketAtI = Math.floor(prefixSum[i] / PRUNING_CHECKPOINT_TOKENS);
-        const bucketBefore =
-          i === 0
-            ? -1
-            : Math.floor(prefixSum[i - 1] / PRUNING_CHECKPOINT_TOKENS);
-        if (bucketAtI > bucketBefore) {
+        const bucketBefore = Math.floor(
+          prefixSum[i - 1] / PRUNING_CHECKPOINT_TOKENS
+        );
+        if (bucketAtI !== bucketBefore) {
+          // We fell into a different bucket than the previous interaction, so this is a
+          // checkpoint too.
           break;
         }
       }
--- a/front/lib/api/assistant/conversation_rendering/pruning.ts
+++ b/front/lib/api/assistant/conversation_rendering/pruning.ts
@@ -6,6 +6,12 @@ const PRUNED_TOOL_RESULT_PLACEHOLDER =
   "</dust_system>";
 const PRUNED_TOOL_RESULT_TOKENS = 24;
 
+// Batch granularity for the redaction checkpoint in prunePreviousInteractions. Same idea as the
+// `clear_at_least` setting on provider-native context-editing features. Don't invalidate a cached
+// prefix for a handful of tokens. Only redact once enough new content has accumulated to make the
+// cache-write cost worthwhile. Starting point, tune against production cache-miss metrics.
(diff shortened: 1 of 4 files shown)


front/lib/api/assistant/conversation_rendering/pruning.ts
```diff
@@ -111,40 +117,135 @@ export function progressivelyPruneInteraction(
...
+  for (let i = 0; i < n; i++) {
+    totalIfNothingRedacted += originalTokens[i];
+  }
+
+  if (totalIfNothingRedacted > maxTokens) {
+    // Redact starting from the OLDEST eligible interaction forward, tracking the actual running
+    // total (not just whether original sizes would exceed budget), until it fits. This is the
+    // minimal redaction needed. Nothing above this frontier is touched.
+    let total = totalIfNothingRedacted;
+    let neededFrontier = -1;
```
aubin-tchoi, line 189: nit: would call that the lowest/earliest working frontier

front/lib/api/assistant/conversation_rendering/pruning.ts
```diff
@@ -111,40 +117,135 @@ export function progressivelyPruneInteraction(
...
+      total -= originalTokens[i] - redactedTokens[i];
+      neededFrontier = i;
     }
 
-    if (shouldPrune) {
-      // We are pruning all previous interactions. We simply prune all tool results and update the
-      // available tokens.
-      interactions[i] = pruneAllToolResults(interactions[i]);
-      interactionTokens = getInteractionTokenCount(interactions[i]);
+    if (neededFrontier >= 0) {
```
aubin-tchoi, line 195: this can't happen since we're in an `if totalIfNothingRedacted > maxTokens`, right?

front/lib/api/assistant/conversation_rendering/pruning.ts
```diff
@@ -111,40 +117,135 @@ export function progressivelyPruneInteraction(
...
+      let effectiveFrontier = neededFrontier;
+      for (let i = neededFrontier; i < floorStart; i++) {
+        effectiveFrontier = i;
+        const bucketAtI = Math.floor(prefixSum[i] / PRUNING_CHECKPOINT_TOKENS);
+        const bucketBefore =
+          i === 0
+            ? -1
+            : Math.floor(prefixSum[i - 1] / PRUNING_CHECKPOINT_TOKENS);
+        if (bucketAtI > bucketBefore) {
+          break;
```
aubin-tchoi, line 217: nit: would rewrite this into
```
if i === 0 break
if we fall into a different bucket between i and i - 1, also break
```
to highlight that we check whether we fall into the same bucket

aubin-tchoi (approved):
LGTM 🔥

front/lib/api/assistant/conversation_rendering/pruning.ts
flvndvd: Actually it can, it just means the previous interactions we do have, all of which are protected, already add up to more than the budget on their own.

front/lib/api/assistant/conversation_rendering/pruning.ts
flvndvd: Indeed, clearer.

PR #28758, opened 2026-07-09
fix(front-api): restrict API key full-secret visibility to its creator

adrsimon:
closes https://github.com/dust-tt/tasks/issues/9532

## Description

In the workspace **API keys** admin page, a freshly generated key's full secret is shown for the first 10 minutes after creation. Until now that window was purely time-based: **any** admin who opened the page during those 10 minutes could see the full secret of a key they did not create.

`KeyResource.toJSON()` decided secret visibility from `createdAt` alone and had no notion of who was asking. This changes `toJSON()` to take the requesting user's model id and only return the full secret when the requester is the key's creator (`key.userId`) **and** the request is within the 10-minute window. Every other admin — and the creator once the window closes — gets the redacted value.

All five call sites in `front-api` (list, create, patch ×2, disable) already have the authenticated user in scope and now pass `user.id`. The `KeyType` shape is unchanged (`secret` stays a redacted string, never null), so no frontend changes are required.

## Tests

Added `front-api/routes/w/[wId]/keys/index.test.ts` covering the list endpoint:
- The admin who created a key sees the full secret within the window.
- A different admin in the same workspace sees the redacted secret.

Both pass. `front` and `front-api` typecheck clean; biome and the pre-commit hooks pass.

## Risk

Low. The change only tightens what is returned — it never exposes more than before. Edge case: keys whose creator has since been removed (`userId` is null via `onDelete: SET NULL`) will always be redacted, which is the correct posture (a key's full secret is only recoverable at creation time). Easy to roll back (revert-only, no migration).

## Deploy Plan

Standard `front` + `front-api` deploy. No migration, no config, no coordinated client release.

### DIFF
--- a/front-api/routes/w/[wId]/keys/[id]/disable.ts
+++ b/front-api/routes/w/[wId]/keys/[id]/disable.ts
@@ -25,6 +25,7 @@ app.post(
   validate("param", KeyIdParamSchema),
   async (ctx): HandlerResult<PostKeysResponseBody> => {
     const auth = ctx.get("auth");
+    const user = auth.getNonNullableUser();
     const owner = auth.getNonNullableWorkspace();
 
     const { id } = ctx.req.valid("param");
@@ -61,7 +62,7 @@ app.post(
 
     return ctx.json({
       key: {
-        ...key.toJSON(),
+        ...key.toJSON(user.id),
         status: "disabled",
       },
     });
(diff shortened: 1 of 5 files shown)


Fraggle (approved):
Looks reasonnable but likely a bit hacky :)

PR #28761, opened 2026-07-09
refactor(mcp): narrow tool handler context to runContext

spolu:
Cadeau de Noel

## Description

`ToolHandlerExtra` carried the full `ToolContextType` union (`runContext` | `listToolsContext`), but tool handlers only ever execute on a connection established with a run context — the listing phase registers handlers without invoking them (the single `callTool` path always connects with `{ runContext }`). The union invited dead optional chaining and unreachable `listToolsContext` fallbacks in handlers.

This narrows the handler-facing contract:

- New `ToolRunContextType = AgentLoopRunContextType | SandboxFunctionRunContextType` alias in `lib/actions/types`, also used inside `ToolContextType` and the run-context type guards.
- `ToolHandlerExtra` now carries a non-optional `runContext: ToolRunContextType` instead of `toolContext?: ToolContextType`.
- `registerTool` is the narrowing boundary: `createServer` factories keep receiving `ToolContextType` (dynamic servers like `run_agent`/`run_dust_app`/`extract_data` legitimately need both phases to register tools), and the run context is asserted and extracted when a handler actually fires.

Dead code this exposed and removed: the unreachable `?? listToolsContext?.conversation` fallback in `set_conversation_title`, "no context" error branches in jira/gmail/microsoft_drive/web_search_browse/toolsets, and a dead `?? "google_drive"` connectionId fallback. Helpers that legitimately take `ToolContextType` (searchFunction, getPod, getUserTimezone, ...) are unchanged; handler call sites pass `{ runContext }`, which satisfies the union with no casts.

Also fixes test fakes whose `as unknown as` casts hid the shape change from the compiler (files, sandbox, skill_management, mcp_actions search mock) — those would have silently kept passing a `toolContext` property nobody reads.

## Tests

`tsgo --noEmit` clean. Ran all touched test suites (~250 tests: files, sandbox, sandbox_functions, skill_management, skill_authoring, sidekick servers, poke, google_drive, extract_data, interactive_content, common_utilities, workspace_analytics, mcp_actions, mcp_execution) — all passing.

Tested locally

## Risk

Low. Behavior-preserving refactor: the removed guards and fallbacks were unreachable (handlers always run with a run context), and a runtime assert in `registerTool` backstops the invariant. Type-only change for the ~50 server files.

## Deploy Plan

- deploy `front`

### DIFF
--- a/front/lib/actions/mcp_internal_actions/tool_definition.ts
+++ b/front/lib/actions/mcp_internal_actions/tool_definition.ts
@@ -1,6 +1,6 @@
 import type { MCPToolStakeLevelType } from "@app/lib/actions/constants";
 import type { MCPError } from "@app/lib/actions/mcp_errors";
-import type { ToolContextType } from "@app/lib/actions/types";
+import type { ToolRunContextType } from "@app/lib/actions/types";
 import type {
   InternalMCPServerDefinitionType,
   MCPToolType,
@@ -22,7 +22,7 @@ export type ToolHandlerExtra = RequestHandlerExtra<
   ServerNotification
 > & {
   auth: Authenticator;
-  toolContext?: ToolContextType;
+  runContext: ToolRunContextType;
 };
 
 export type ToolHandlerResult = Result<CallToolResult["content"], MCPError>;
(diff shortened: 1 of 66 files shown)


front/lib/actions/mcp_internal_actions/tool_definition.ts
```diff
@@ -22,7 +22,7 @@ export type ToolHandlerExtra = RequestHandlerExtra<
   ServerNotification
 > & {
   auth: Authenticator;
-  toolContext?: ToolContextType;
+  runContext: ToolRunContextType;
```
spolu, line 25: The big win

aubin-tchoi (approved):
amazing 💪🏻

flvndvd (commented):
🔥

PR #28763, opened 2026-07-09
Add: convenient method to give user permission on a resource.

Fraggle:
## Description

Adds `GroupPermissionResource.grantToUser` and `revokeFromUser` — two high-level static helpers that manage per-user instance-level permissions without callers needing to know about the backing `regular_auto` group mechanics.

- `grantToUser` finds the existing `regular_auto` group for a `(permissionType, resourceType, resourceId)` tuple, or creates one and calls `grant()` on first use, then adds the user via `dangerouslyAddMember`. Idempotent.
- `revokeFromUser` removes the user from the backing group. If they were the last member, it revokes the grant and deletes the group. No-op if the user is not a member.

Both methods are fully transactional via `withTransaction` and accept an optional outer `transaction`.

## Tests

Added Vitest unit tests in `group_permission_resource.test.ts` covering: create-on-first-grant, group reuse for a second user, revoke-with-group-cleanup when last member removed, and revoke-while-other-members-remain.

## Risk

Low. Purely additive — new static methods on an existing resource, no schema changes, no existing call sites modified.

## Deploy Plan

Deploy front.

### DIFF
--- a/front/lib/resources/group_permission_resource.test.ts
+++ b/front/lib/resources/group_permission_resource.test.ts
@@ -1,7 +1,9 @@
 import { Authenticator } from "@app/lib/auth";
 import { GroupPermissionResource } from "@app/lib/resources/group_permission_resource";
-import type { GroupResource } from "@app/lib/resources/group_resource";
+import { GroupResource } from "@app/lib/resources/group_resource";
 import { GroupFactory } from "@app/tests/utils/GroupFactory";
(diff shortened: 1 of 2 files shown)


Fraggle:

- #28763 ◀
    - #28779
        - #28781
            - #28791
                - #28792
                    - #28793
                        - #28794
                            - #28795
                                - #28796

PR #28764, opened 2026-07-09
[front] Fix Agent Builder editor margins and update save/cancel/leave buttons

rfrenoy:
## Description

Fixes https://github.com/dust-tt/tasks/issues/9010

Screenshot 2026-07-09 at 14 16 53

Align the left-panel content with the top/bottom bars. The content was inset by two stacking layers of horizontal padding (the wrapper plus each section's own px-6), pushing it ~24px further right than the header and footer on both sides.

Make the content wrapper the single source of horizontal padding (p-4), drop the redundant px-6 from AgentBuilderSectionContainer and AgentBuilderSpacesBlock, and remove the mx-4 margins from the bars so the header, content, and footer all sit at a consistent 16px inset.
## Tests

Local

## Risk

Low

## Deploy Plan

Front

### DIFF
--- a/front/components/agent_builder/AgentBuilderLeftPanel.tsx
+++ b/front/components/agent_builder/AgentBuilderLeftPanel.tsx
@@ -9,6 +9,7 @@ import {
   BarFooter,
   BarHeader,
   Button,
+  cn,
   ScrollArea,
   XClose,
 } from "@dust-tt/sparkle";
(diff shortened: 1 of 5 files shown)


front/components/agent_builder/instructions/AgentBuilderInstructionsEditor.tsx
```diff
@@ -247,16 +247,25 @@ export function AgentBuilderInstructionsEditor({
...
+          if (!isInstructionDiffMode && !editor.isDestroyed) {
+            field.onChange(editor.getMarkdown());
+            // Strip style/class/id attributes to store clean HTML structure.
+            instructionsHtmlField.onChange(
+              stripHtmlAttributes(editor.getHTML())
+            );
+          }
+        },
+        INSTRUCTIONS_DEBOUNCE_MS,
+        { leading: true, trailing: true }
```
ykmsd, line 267: TIL this exists

front/components/agent_builder/AgentBuilderLeftPanel.tsx
```diff
@@ -50,19 +53,20 @@ export function AgentBuilderLeftPanel({
...
+              onClick={handleCancel}
+              variant="ghost"
+              type="button"
+            />
+          )
         }
       />
       <ScrollArea className="flex-1">
-        <div className="mx-auto space-y-10 p-8 2xl:max-w-5xl">
+        <div className="mx-auto space-y-10 p-4 2xl:max-w-5xl">
```
ykmsd, line 69: I think x padding is too tight unless you open it in a big monitor, how about `px-6 py-4`?

CleanShot 2026-07-09 at 17 27 17@2x

rfrenoy:
@ykmsd It does not exactly behave in prod as it does locally (once we start editing, if we undo the changes we don't come back to the initial state), I'm investigating.

front/components/agent_builder/AgentBuilderLeftPanel.tsx
ykmsd: I guess we want to align the padding with header so we need to adjust the header too 🙏

ykmsd (commented):
I will wait your investigation but it's kind of weird that this close button doesn't appear for the first time you open an agent builder no? I expected isDirty to be `false` on initialization but seems like it's not 🤔

rfrenoy:
@adrsimon @ykmsd I'm sorry I did not have time to merge this before my leave, can one of you take over?  

The feature mostly works, I just have an issue when they are ">" or "<" characters in the instructions, then even if you undo your changes the cross button never reappears and the cancel/save buttons never leave. I think that's acceptable, but I'll let you decide. The issue probably comes from the fact that it messes with _stripHtmlAttributes_ if you want to dig further.

ykmsd:
@rfrenoy no worries I can take over, anyway I was supposed to do it 🙈 @adrsimon I'm off on Monday but I promise I will take a look sometime next week 🫡

ykmsd:
commented here but I'm not super convinced with hiding the button approach https://github.com/dust-tt/tasks/issues/9010#issuecomment-5031757902 I will only adjust the spacing for now

PR #28765, opened 2026-07-09
Mark the trailing tool result with a cache breakpoint on Vertex in the legacy Anthropic client

flvndvd:
## Description

On Vertex AI, prompt caching requires an explicit breakpoint since automatic caching isn't supported there. The legacy Anthropic client already handled this for the last message when it was plain user text, but silently dropped the flag whenever the last message was a tool result, which is exactly the shape of every iteration of a tool-calling loop. As a result, conversations that ended a turn on a tool result were reprocessing the full conversation tail uncached on every iteration on Vertex, adding latency and cost with no benefit. This fix threads the same "is this the last message" signal through to the tool result path so the breakpoint gets set there too, matching the behavior already used for plain text turns. Direct Anthropic API traffic is unaffected since it relies on automatic caching rather than this explicit marker.

### DIFF
--- a/front/lib/api/llm/clients/anthropic/utils/conversation_to_anthropic.ts
+++ b/front/lib/api/llm/clients/anthropic/utils/conversation_to_anthropic.ts
@@ -181,11 +181,20 @@ async function toolResultToParam(
 
 async function functionMessage(
   message: FunctionMessageTypeModel,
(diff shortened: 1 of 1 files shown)


aubin-tchoi (approved):
LGTM

pmilliotte (approved):

PR #28767, opened 2026-07-09
Rename ToolContext* types (remove -Type) suffix

spolu:
## Description

Rename of `TooContext*Type` types to `ToolContext*` as these are not really serialised or sent over the network.

## Tests

Covered by existing tests and type-checker. Tested locally.

## Risk

None. Can be reverted.

## Deploy Plan

N/A pure type.

### DIFF
--- a/front/lib/actions/types/index.ts
+++ b/front/lib/actions/types/index.ts
@@ -193,9 +193,7 @@ export type AgentLoopListToolsContext = {
 
 // Context available to tool handlers at execution time: tools only ever run on a connection
 // established with a run context, never on a listing-phase connection.
-export type ToolRunContext =
-  | AgentLoopRunContext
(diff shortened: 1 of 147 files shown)


front/lib/actions/types/index.ts
```diff
@@ -193,28 +193,26 @@ export type AgentLoopListToolsContextType = {
...
 
 export function isAgentLoopRunContext(
-  value: ToolRunContextType | undefined
-): value is AgentLoopRunContextType {
+  value: ToolRunContext | undefined
+): value is AgentLoopRunContext {
   return value?.contextType === "agent_loop";
 }
 
-export type ToolContextType =
```
spolu, line 212: This is the crux of the rename

flvndvd (approved):
LGTM

PR #28770, opened 2026-07-09
[front] refactor: move group pool cap to a dedicated group_pool_caps table

fabiencelier:
## Description

Moves the per-group usage spend limit off the `groups` table into a new `group_pool_caps` table storing just `workspaceId`, `groupId` and `poolCapAwuCredits`. A row means the group carries a cap; the absence of a row means no cap (falls back to the workspace default), so the column's `NULL` state maps to "no row" and the new column is `NOT NULL`.

## Migrations (expand/contract)

- **pre-deploy**: create `group_pool_caps` and backfill it from `groups."poolCapAwuCredits"`.
- **post-deploy**: catch-up backfill with `ON CONFLICT ("groupId") DO NOTHING` (covers caps created by old code during the deploy window, without clobbering rows the new code wrote), then drop the `groups."poolCapAwuCredits"` column.

Known edge: a cap *modified or cleared* by old code during the deploy window keeps its pre-deploy value in the new table. Caps change rarely, so this window is accepted.

## Deploy Plan

1. Run the pre-deploy migration (creates + backfills `group_pool_caps`).
2. Deploy `front` / `front-api`.
3. Run the post-deploy migration (catch-up backfill + drop the old column).

### DIFF
--- a/front-api/routes/w/[wId]/groups/index.ts
+++ b/front-api/routes/w/[wId]/groups/index.ts
@@ -16,7 +16,10 @@ import groupDetail from "./[groupId]";
 import spendLimit from "./[groupId]/spend_limit";
 
 export type GetGroupsResponseBody = {
-  groups: (GroupType & { memberCount: number })[];
+  groups: (GroupType & {
+    memberCount: number;
+    poolCapAwuCredits: number | null;
+  })[];
 };
 
 const GetGroupsQuerySchema = z.object({
@@ -49,10 +52,15 @@ app.get(
       auth,
       groups
     );
+    const poolCaps = await GroupResource.getPoolCapAwuCreditsForGroups(
+      auth,
+      groups
+    );
 
     const groupsWithMemberCount = groups.map((group) => ({
       ...group.toJSON(),
       memberCount: memberCounts.get(group.id) ?? 0,
+      poolCapAwuCredits: poolCaps.get(group.id) ?? null,
     }));
 
     return ctx.json({ groups: groupsWithMemberCount });
--- a/front-api/routes/w/[wId]/groups/index.ts
+++ b/front-api/routes/w/[wId]/groups/index.ts
@@ -32,6 +32,12 @@ const GetGroupsQuerySchema = z.object({
   // When "true", each group also carries its member sIds (one extra batched
   // query) instead of just memberCount.
   withMembers: z.enum(["true", "false"]).optional(),
+  // When "true", each group also carries its pool cap (one extra batched
+  // query). Accepted but not yet honored: the cap is still returned
+  // unconditionally so front-end bundles that predate the flag keep working.
+  // A follow-up makes the cap conditional on this flag, once the bundles
+  // sending it are deployed everywhere.
+  withPoolCaps: z.enum(["true", "false"]).optional(),
 });
 
 // Mounted at /api/w/:wId/groups.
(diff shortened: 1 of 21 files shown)


front/lib/resources/storage/models/group_pool_caps.ts
```diff
@@ -0,0 +1,47 @@
...
+    indexes: [
+      { unique: true, fields: ["groupId"] },
+      { fields: ["workspaceId"], concurrently: true },
+    ],
+  }
+);
+
+GroupPoolCapModel.belongsTo(GroupModel, {
+  foreignKey: { name: "groupId", allowNull: false },
+  onDelete: "RESTRICT",
```
philipperolet, line 46: we need to update workspace deletion to delete groupPoolCaps before groups

front/migrations/post-deploy/20260720114500_drop_pool_cap_awu_credits_from_groups.sql
```diff
@@ -0,0 +1,24 @@
...
+already written to group_pool_caps by the new code (which no longer touches
+the groups column) as the source of truth.
+ */
+SET SESSION statement_timeout = 1200000;
+SET SESSION lock_timeout = 3000;
+INSERT INTO "public"."group_pool_caps" ("createdAt", "updatedAt", "workspaceId", "groupId", "poolCapAwuCredits")
+SELECT now(), now(), "workspaceId", "id", "poolCapAwuCredits"
+FROM "public"."groups"
+WHERE "poolCapAwuCredits" IS NOT NULL
+ON CONFLICT ("groupId") DO NOTHING;
```
philipperolet, line 19: here, if people changed caps between the pre-deploy and post deploy, that'd be lost no?

philipperolet: I think we can ignore if there are less than 10 caps change per day (would mean 95% chance of everything fine) but leaving that to you

front/lib/resources/group_resource.ts
philipperolet, line 1: don't forget to deploy front-sse too

front/lib/resources/group_resource.ts
```diff
@@ -2643,7 +2704,6 @@ export class GroupResource extends BaseResource<GroupModel> {
       workspaceId: this.workspaceId,
       kind: this.kind,
       memberCount,
-      poolCapAwuCredits: this.poolCapAwuCredits,
```
philipperolet, line 2646: nit: as we remove that, some api callers (group create, detail, patch) on old browser bundles may fail. imo not too big an issue because a refresh solves it and I don't think anything ends in a broken state, but flagging

philipperolet (commented):
Looks good, thanks for the work 🙏 Left a few coms

front/lib/resources/storage/models/group_pool_caps.ts
fabiencelier: ah true, thank you for catching this

front-api/routes/w/[wId]/groups/index.ts
Comment (id13) [inline front-api/routes/w/[wId]/groups/index.ts:49]: Maybe add an include there instead of adding one more query

front/lib/resources/group_resource.ts
id13, line 1314: How often is this method called ?

id13 (approved):
A little bit worried by access path and N+1 potential

PR #28771, opened 2026-07-09
Batch the tool-output-content fetch window instead of sliding it every turn

flvndvd:
## Description

When loading a conversation for the agent to run on, we only fetch tool-output content from the database for a limited number of recent interactions, both to bound memory usage and to avoid unnecessary reads of large or file-backed tool results. Today that window is a fixed size that always covers the most recent handful of interactions, which means as a conversation grows, exactly one older interaction loses its tool-output content on every single turn, the boundary just slides forward by one each time.

With the recent changes from https://github.com/dust-tt/dust/pull/28751, that interaction's content changes from what the model actually produced to a generic "no output" placeholder, which breaks Anthropic's prompt cache the same way our earlier pruning fix addressed, except this happens upstream of that fix entirely, before rendering ever gets a say, so it was quietly capping how much of that improvement could actually land.

This change keeps the same guaranteed floor of always-fresh recent interactions, but anchors the boundary for everything older to a fixed position counted from the start of the conversation rather than distance from the latest turn. In practice that means the window stays exactly the same for many consecutive turns and only moves forward once a full batch of new interactions has accumulated, instead of creeping by one every time. No caller of this function needed to change. The fix is entirely internal to how the window is computed.

The tradeoff worth naming: instead of losing one interaction's content every turn forever, the conversation now goes through longer stable stretches punctuated by an occasional larger jump, several interactions' worth of content disappearing at once when a batch boundary is crossed. That's a real change in shape, not just frequency, and the batch size is a tunable starting guess rather than something empirically measured yet.

> [!NOTE]
> Not a big fan but will see how we can improve later.

### DIFF
--- a/front/lib/api/assistant/conversation/fetch.test.ts
+++ b/front/lib/api/assistant/conversation/fetch.test.ts
@@ -1,0 +1,104 @@
+import {
+  computeMessagesWithToolOutputContent,
+  TOOL_OUTPUT_FETCH_BATCH_SIZE,
+} from "@app/lib/api/assistant/conversation/fetch";
+import type { Interaction } from "@app/lib/api/assistant/conversation_rendering/pruning";
+import { describe, expect, it } from "vitest";
+
+// Builds n interactions, each a user message followed by an agent message. Agent message ids are
+// 1, 2, 3, ... in order, matching the interaction's position, so tests can assert on exactly which
+// interactions' content ended up included.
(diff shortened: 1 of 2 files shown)


flvndvd:
PMRR.

adrsimon (approved):

aubin-tchoi (commented):
LGTM

PR #28773, opened 2026-07-09
[front] fix fair-use limits when cost overshoots the limit

thervier:
## Description

Fix fair-use AWU credits silently dropped when cost overshoots the limit

* computeAndStoreAgentMessageCredits (credit_cost.ts) persisted a message's AWU credit cost by calling rateLimiter, but rateLimiter's Lua script is a gatekeeper: it drops the write entirely when count + incrementBy > limit. Enforcement already happens earlier, pre-message, via getRateLimiterCount + a >= maxAwuCredits check — so gating the recording write too meant the counter could get stuck just below the limit (e.g. 9/10) and never actually reach it, since any message costing more than the remaining headroom silently failed to record. Users could keep sending messages indefinitely without ever tripping the fair-use limit.
* Added addRateLimiterCount to rate_limiter.ts: an unconditional-write variant with no limit guard, for post-hoc recording of a cost that already happened. Enforcement stays exactly where it was (getRateLimiterCount pre-message).
* Swapped the rateLimiter call in computeAndStoreAgentMessageCredits for addRateLimiterCount.

## Tests

 * npx tsgo --noEmit clean
 * biome check clean
 * Added regression tests in rate_limiter.test.ts against real Redis, including one that reproduces the exact 9+2>10 overshoot from the bug and confirms the full amount is now persisted (previously the write would've been dropped)
 * credit_cost.test.ts and rate_limiter.test.ts pass (22 tests)

## Risk

Low, impacts only AWU fair use limits (that are not used yet)

## Deploy Plan

Deploy front

### DIFF
--- a/front/lib/api/assistant/credit_cost.ts
+++ b/front/lib/api/assistant/credit_cost.ts
@@ -14,8 +14,8 @@ import { ConversationResource } from "@app/lib/resources/conversation_resource";
 import type { RunUsageType } from "@app/lib/resources/run_resource";
 import { RunResource } from "@app/lib/resources/run_resource";
 import {
+  addRateLimiterCount,
   getTimeframeSecondsFromLiteral,
-  rateLimiter,
 } from "@app/lib/utils/rate_limiter";
 import logger from "@app/logger/logger";
 import {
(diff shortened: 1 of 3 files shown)


spolu (approved):
LGTM

PR #28774, opened 2026-07-09
Document the larger Frame source size cap in the files edit and create tool descriptions

flvndvd:
## Description

Follow up of https://github.com/dust-tt/dust/pull/28713.

The edit and create tool descriptions still told agents that files over 50 KB could not be edited or created, but a recent change gave Frame source files a much higher cap, 1024 KB, so large templates stay editable and creatable. Agents were seeing stale guidance and could incorrectly assume a Frame template was too large to work with. The descriptions now call out the higher cap that applies when editing a Frame source file or overwriting an existing one through create, matching the actual runtime behavior.


adrsimon (approved):

aubin-tchoi (approved):
LGTM

PR #28775, opened 2026-07-09
[front] fix: allow admins to become agent editors

aubin-tchoi:
## Description

Follow-up to https://github.com/dust-tt/dust/pull/28649, which split write/admin permissions for editor groups. This PR fixes the agent editors endpoint so workspace admins use the admin permission path when modifying agent editors, including adding themselves as editors.

It also adds regression coverage for both skills and agents to keep the "Become an editor" path working for workspace admins who are not already in the resource editor group.

## Tests

- Added regression tests.

## Risk

- Low. Narrow editor-management permission check.

## Deploy Plan

- Deploy front.

### DIFF
--- a/front-api/routes/w/[wId]/assistant/agent_configurations/[aId]/editors.ts
+++ b/front-api/routes/w/[wId]/assistant/agent_configurations/[aId]/editors.ts
@@ -203,7 +203,7 @@ app.patch(
     }
 
     const editorGroup = editorGroupRes.value;
-    if (!editorGroup.canWrite(auth)) {
+    if (!editorGroup.canAdministrate(auth)) {
       return apiError(ctx, {
         status_code: 403,
         api_error: {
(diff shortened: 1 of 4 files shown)


front-api/routes/w/[wId]/assistant/agent_configurations/[aId]/editors.ts
```diff
@@ -203,7 +203,7 @@ app.patch(
     }
 
     const editorGroup = editorGroupRes.value;
-    if (!editorGroup.canWrite(auth)) {
+    if (!editorGroup.canAdministrate(auth)) {
```
Comment (philipperolet) [inline front-api/routes/w/[wId]/assistant/agent_configurations/[aId]/editors.ts:206]: i don't recall, do non-admin editors have canAdministrate access? otherwise they'll stop being able to edit editors?

philipperolet (commented):
Hey, little question otherwise LGTM

front-api/routes/w/[wId]/assistant/agent_configurations/[aId]/editors.ts
Comment (aubin-tchoi) [reply front-api/routes/w/[wId]/assistant/agent_configurations/[aId]/editors.ts:206]: They do (the name makes it a bit confusing though)

Comment (aubin-tchoi) [reply front-api/routes/w/[wId]/assistant/agent_configurations/[aId]/editors.ts:206]: adding a test 👍🏻

Comment (aubin-tchoi) [reply front-api/routes/w/[wId]/assistant/agent_configurations/[aId]/editors.ts:206]: the test already exists actually, it's `editor should successfully add another editor`

philipperolet (approved):

PR #28766, opened 2026-07-09
Extract presentational cores from tool approval and personal auth components

spolu:
## Description

Refactor in preparation for tool approval / personal auth support in the viz wrapper (sandbox function tool calls). No behavior change.

- Extract `ToolValidationCard` (title/always-allow/decline-allow UI, `MCP_TOOL_OVERRIDES`) out of `MCPToolValidationRequired`, which becomes a thin conversation wrapper keeping `useValidateAction`, the `BlockedActionsProvider` context and the always-allow cascade.
- Extract `PersonalAuthenticationCard` (connect/skip flow, credential overrides, mid-connection cancellation guard) out of `MCPServerPersonalAuthenticationRequired`, which becomes a thin wrapper keeping `useResolveAuthentication` and the context.
- The cards take the narrowed display shape (`Pick<ToolExecutionBase, ...>`) plus an `onValidate` / `onResolve` submit callback, so both agent-loop and sandbox-function blocked tool executions can be rendered. `ToolValidationDetails` props narrowed accordingly.

Follow-ups (separate PRs): SSE consumption restructure in the viz wrapper (blocking events, reconnect), then the sandbox wrappers + validate/resolve hooks.

## Tests

- Existing `MCPServerPersonalAuthenticationRequired.test.tsx` passes unchanged (mocks are module-level, extraction is transparent to it).
- Tested locally

## Risk

Low. Pure extraction; conversation flows unchanged.

## Deploy Plan

- deploy `front`

PR #28776, opened 2026-07-09
Reconnect and lifetime handling for the sandbox function invocation event stream

spolu:
## Description

Third sandbox function approval slice, after #28766 and #28788.

Sandbox function invocation events now use `useEventSource`, reusing its `done` sentinel reconnects, `lastEventId` resumes, bounded backoff, heartbeat, and page lifecycle recovery. Each in-flight invocation mounts a component that settles the iframe call and will host approval and authentication UI.

`useEventSource` accepts an optional terminal-error callback so invocation failures settle directly without an error-watching effect. The fixed client timeout is removed because the server publishes a terminal result or error once an invocation exists.

The personal authentication event is added to the invocation event union ahead of the follow-up UI.

## Tests

N/A, tested locally.

## Risk

Low. Limited to feature-flagged sandbox function invocations.

## Deploy Plan

- deploy `front`

### DIFF
--- a/front/components/assistant/conversation/actions/VisualizationActionIframe.tsx
+++ b/front/components/assistant/conversation/actions/VisualizationActionIframe.tsx
@@ -101,12 +101,8 @@ interface SandboxFunctionInvocationProps {
   ) => void;
 }
 
-// Owns the client side of one in-flight invocation: consumes its SSE stream through
-// useEventSource (which handles the `done` idle sentinel, lastEventId resumes, reconnection
-// backoff and bfcache recovery) and settles the pending `callFunction` promise held by the
-// parent. Will also render the blocking-event UI (tool approval, personal authentication) for the
-// invocation; renders nothing until then. No client-side deadline: the server guarantees a
-// terminal event (result or error) once the invocation exists.
+// Consumes one invocation's event stream, settles the pending iframe call, and will render
+// approval or authentication UI when needed.
 function SandboxFunctionInvocation({
   workspaceId,
   functionId,
@@ -170,22 +166,19 @@ function SandboxFunctionInvocation({
     [invocationId, onSettle]
   );
 
-  const { isError } = useEventSource(
+  const onTerminalError = useCallback(() => {
+    onSettle(invocationId, {
+      result: null,
+      error: "Failed to listen to function invocation events.",
+    });
+  }, [invocationId, onSettle]);
+
+  useEventSource(
     buildEventSourceURL,
     onEventCallback,
-    `sandbox-function-invocation-${invocationId}`
+    `sandbox-function-invocation-${invocationId}`,
+    { onTerminalError }
   );
-
-  // useEventSource only exposes its terminal failure (reconnect attempts exhausted) as state, so
-  // an effect is the only place to propagate it into the pending call settlement.
-  useEffect(() => {
-    if (isError) {
-      onSettle(invocationId, {
-        result: null,
-        error: "Failed to listen to function invocation events.",
-      });
-    }
-  }, [isError, invocationId, onSettle]);
 
   return null;
 }
--- a/front/components/assistant/conversation/actions/VisualizationActionIframe.tsx
+++ b/front/components/assistant/conversation/actions/VisualizationActionIframe.tsx
@@ -1,6 +1,7 @@
 import { useVisualizationRetry } from "@app/hooks/conversations";
+import { useEventSource } from "@app/hooks/useEventSource";
 import { useSendNotification } from "@app/hooks/useNotification";
-import { clientEventSource, clientFetch } from "@app/lib/egress/client";
+import { clientFetch } from "@app/lib/egress/client";
 import { getErrorFromResponse } from "@app/lib/swr/swr";
 import datadogLogger from "@app/logger/datadogLogger";
 import type {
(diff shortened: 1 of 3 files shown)


spolu:
Tested end to end

PMRR

front/hooks/useEventSource.ts
PopDaph, line 157: I'm not sure about the timeout removal, what happens if the user close her laptop for more than 10min and missed the terminal event? It looks like they would try to reconnect forever cause there's not check on MAX_RECONNECT_ATTEMPTS on the done reconnects path?

spolu: Well same as with conversation state we will attempt reconnect etc and reonnect with lastEventId to recover the events and possibly receive the result!

spolu: We will need ~ all that is in useEventSource over time so it's pretty natural to use it directly which renders the timeout meaningless right?

front/components/assistant/conversation/actions/VisualizationActionIframe.tsx
```diff
@@ -90,116 +91,109 @@ const getExtensionFromBlob = (blob: Blob): string => {
...
-        } = JSON.parse(event.data);
+        } = JSON.parse(eventStr);
 
         switch (eventPayload.data.type) {
           case "sandbox_function_invocation_created":
             // NO-OP
             break;
           case "sandbox_function_invocation_result":
-            finish({ result: eventPayload.data.result });
+            onSettle(invocationId, { result: eventPayload.data.result });
```
flvndvd, line 138: What's the plan here? Don't we want to stop listening when we get this event?

front/components/assistant/conversation/actions/VisualizationActionIframe.tsx
```diff
@@ -433,6 +429,48 @@ export const VisualizationActionIframe = forwardRef<
   const [isCodeDrawerOpen, setCodeDrawerOpened] = useState(false);
   const vizIframeRef = useRef<HTMLIFrameElement | null>(null);
 
+  // In-flight sandbox function invocations. Each entry mounts a
+  // SandboxFunctionInvocation; the resolver of the pending `callFunction` promise is kept
+  // in a ref so settling stays a pure state update.
+  const [activeInvocations, setActiveInvocations] = useState<
+    { functionId: string; invocationId: string }[]
+  >([]);
+  const invocationResolversRef = useRef<Map<
```
flvndvd, line 438: I wish we had sockets here? The split by `invocationId` makes this code pretty sad.

front/components/assistant/conversation/actions/VisualizationActionIframe.tsx
```diff
@@ -433,6 +429,48 @@ export const VisualizationActionIframe = forwardRef<
...
+    ({
+      functionId,
+      invocationId,
+    }: {
+      functionId: string;
+      invocationId: string;
+    }) =>
+      new Promise<CommandResultMap["callFunction"]>((resolve) => {
+        invocationResolvers.set(invocationId, resolve);
+        setActiveInvocations((prev) => [...prev, { functionId, invocationId }]);
```
flvndvd, line 457: Returning a new array here will change all the references, should we just push instead and return the `prev` reference (@ykmsd keep me honest here)?

front/components/assistant/conversation/actions/VisualizationActionIframe.tsx
```diff
@@ -433,6 +429,48 @@ export const VisualizationActionIframe = forwardRef<
...
+
+  const waitForSandboxFunctionInvocationResult = useCallback(
+    ({
+      functionId,
+      invocationId,
+    }: {
+      functionId: string;
+      invocationId: string;
+    }) =>
+      new Promise<CommandResultMap["callFunction"]>((resolve) => {
```
flvndvd, line 455: One issue with this pattern is that in case component remounts, we leave dangling promises.

front/components/assistant/conversation/actions/VisualizationActionIframe.tsx
spolu: The event is terminal so the stream will end no?

front/components/assistant/conversation/actions/VisualizationActionIframe.tsx
spolu: Open to suggestions here 😬

front/components/assistant/conversation/actions/VisualizationActionIframe.tsx
flvndvd: It depends on the stream implementation on the server side but IIRC it's slightly overfitted around our conversation needs, message drops it once status is "finish".

front/components/assistant/conversation/actions/VisualizationActionIframe.tsx
flvndvd: Wondering if just wiring those events directly to viz, and letting viz keep it's "promise" opened, could not abstract most of this.

spolu: We do need to get the tool related events here for approval and stuff so there is no "pure wiring" right?

front/components/assistant/conversation/actions/VisualizationActionIframe.tsx
spolu: Will double check we have similar logic here (I do believe so) AND we do do drop the stream and don't reconnect etc... once a terminal event is reached

front/components/assistant/conversation/actions/VisualizationActionIframe.tsx
ykmsd: state is immutable so you cannot really do that no?

PR #28783, opened 2026-07-09
fix(mcp): update cat and search tool handlers to use runContext

frankaloia:
## Summary

- After #28761 renamed `ToolHandlerExtra.toolContext` → `runContext`, `cat.ts` and `search.ts` were not updated.
- Because `toolContext` was typed optional, TypeScript accepted the mismatch silently — at runtime `toolContext` was always `undefined`, causing every `cat` call to return `"No conversation context available"`.
- Fix: both handlers now destructure `runContext: ToolRunContext` directly from `extra`, matching the current `ToolHandlerExtra` shape.

## Testing

Incident fix — `search.ts` had the same stale signature but the missing guard was a `throw` rather than an `Err` return; still needs the update.


smb2268 (approved):
Nice, looks good

PR #28786, opened 2026-07-10
fix(skill_authoring): address list_skills pagination review comments

frankaloia:
## Summary

- **Stable ordering**: skills are now sorted by `sId` before slicing, so pagination is consistent regardless of DB return order
- **cursor as integer**: dropped the opaque base64 string in favour of a plain `z.number().int().nonnegative()` offset — schema enforces validity, simpler for the agent to reason about, and out-of-bounds cursor returns a tool error so the agent self-corrects
- **`agent_discoverable` description**: clarified that this filter returns workspace-default discoverable skills regardless of conversation context — this is the intended behavior, the description was just misleading

## Testing
Local validation

## Risk
Fixing existing issue

## Deploy
1. Deploy front

### DIFF
--- a/front/lib/api/actions/servers/skill_authoring/tools/index.ts
+++ b/front/lib/api/actions/servers/skill_authoring/tools/index.ts
@@ -364,16 +364,22 @@ const handlers: ToolHandlers<typeof SKILL_AUTHORING_TOOLS_METADATA> = {
           : allSkills;
     }
 
+    skills = [...skills].sort((a, b) => a.sId.localeCompare(b.sId));
+
     const pageSize = Math.min(limit ?? 20, 50);
-    const offset = cursor
-      ? parseInt(Buffer.from(cursor, "base64").toString(), 10)
-      : 0;
+    const offset = cursor ?? 0;
+
+    if (offset > skills.length) {
+      return new Err(
+        new MCPError(
+          `cursor ${offset} is out of range (total: ${skills.length})`
+        )
+      );
+    }
+
     const page = skills.slice(offset, offset + pageSize);
     const nextOffset = offset + pageSize;
-    const nextCursor =
-      nextOffset < skills.length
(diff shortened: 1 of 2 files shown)


front/lib/api/actions/servers/skill_authoring/tools/index.ts
```diff
@@ -242,16 +243,22 @@ const handlers: ToolHandlers<typeof SKILL_AUTHORING_TOOLS_METADATA> = {
...
+
     const pageSize = Math.min(limit ?? 20, 50);
-    const offset = cursor
-      ? parseInt(Buffer.from(cursor, "base64").toString(), 10)
-      : 0;
+    const offset = cursor ?? 0;
+
+    if (offset > skills.length) {
+      return new Err(
+        new MCPError(
```
philipperolet, line 374: what about tracked = false here? IIUC it's expected that there are those errors sometimes and the agent self-corrects? in which case we want to avoid monitors?

front/lib/api/actions/servers/skill_authoring/tools/index.ts
```diff
@@ -242,16 +243,22 @@ const handlers: ToolHandlers<typeof SKILL_AUTHORING_TOOLS_METADATA> = {
...
 
+    skills = [...skills].sort((a, b) => a.sId.localeCompare(b.sId));
+
     const pageSize = Math.min(limit ?? 20, 50);
-    const offset = cursor
-      ? parseInt(Buffer.from(cursor, "base64").toString(), 10)
-      : 0;
+    const offset = cursor ?? 0;
+
+    if (offset > skills.length) {
```
philipperolet: nit: maybe >= ? 
as per L261, a valid returned offset is always < length, so offset === length is the same error case but here treated differently (returns empty page VS throwing)

philipperolet (approved):
LGTM, thanks 🙏 maybe worth acting on tracked, but up to you 👍

PR #28789, opened 2026-07-10
Raise the pruning checkpoint granularity to 20k tokens

flvndvd:
## Description

We've observed in production that a single agent turn running a couple of web searches can produce around five thousand tokens on its own, which was enough to cross the previous checkpoint threshold almost every turn and defeat the batching benefit the pruning refactor was meant to provide. Rather than scale the checkpoint as a percentage of each model's context window, which would tie a stability-critical constant to a live, per-model quantity, we picked a single flat value informed by our primary production model's context size. We verified against the actual pruning function that a checkpoint far larger than a small model's entire budget degrades gracefully: the budget is still always respected and there are no reversals, it just loses the batching benefit for that model, which is an acceptable tradeoff since minimizing cache misses matters most for the large-context models that see the longest conversations.


flvndvd:
PMRR.

aubin-tchoi (commented):
LGTM

PR #28781, opened 2026-07-09
Add: model tier resource and user / group / workspace management.

Fraggle:
## Description

Wires model tier access into the group permission system. Adds `"use"` to `PERMISSION_TYPES` and `"models_tier"` to `GROUP_PERMISSION_RESOURCE_TYPES`, registers the type in the permission registry (instance-level only — no type-wide grants), and extends `ModelsTierResource` with `grantToUser`, `revokeFromUser`, `grantToGroup`, and `revokeFromGroup`, all delegating to `GroupPermissionResource` using the tier's numeric ID as `resourceId`. User grants go through a `regular_auto` group (existing pattern from `grantToUser`); group grants target the group directly. Also adds `grantToEverybody`/`revokeFromEverybody` to `GroupPermissionResource` for instance-level grants to the workspace global group.

## Tests

Added `models_tier_resource.test.ts` covering user and group grant/revoke flows end-to-end. Updated `group_permission_registry.test.ts` to assert `models_tier` validates correctly and rejects type-wide grants. Updated `group_permission_resource.test.ts` to cover `grantToEverybody`/`revokeFromEverybody`.

## Risk

Low. No DB migration — uses the existing `group_permissions` table. No new API routes. The `models_tier` resource type is inert until callers use it.

## Deploy Plan

Deploy front.

PR #28779, opened 2026-07-09
Add model tiers.

Fraggle:
## Description

Introduces a static `(modelId, reasoningEffort) → ModelsTierName` classification system as the data layer for model tier enforcement. Three tiers: `cost_efficient`, `balanced`, `frontier`.

- `STATIC_MODEL_SUPPORTED_REASONING_EFFORTS` — per-model const map of supported reasoning efforts, extracted separately from `SUPPORTED_MODEL_CONFIGS` to keep typing sharp and avoid circular imports.
- `STATIC_MODEL_TIERS` in `tiers.ts` — explicit editorial table assigning each `(StaticModelIdType, effort)` pair to a tier. Replaces the cost-threshold approach from `models_picker/tiers.ts` with a curated static lookup.
- `ModelsTierResource` — static facade exposing `getTierForModel` / `getTierForSelection` / `listTiers` as the query interface for downstream enforcement and API routes.

## Tests

New Vitest suite in `tiers.test.ts`:
- Asserts `STATIC_MODEL_TIERS` covers every `StaticModelIdType`.
- Asserts `STATIC_MODEL_SUPPORTED_REASONING_EFFORTS` stays in sync with `SUPPORTED_MODEL_CONFIGS` (catch-all guard for future model additions).
- Spot-checks tier assignments for key models: opus/fable → `frontier`, sonnet light → `cost_efficient`, sonnet medium → `balanced`, GPT-5.5 light → `balanced`, GPT-5-nano light → `cost_efficient`.

## Risk

Low. New files only — no DB migrations, no API routes, no UI changes. No existing logic modified.

## Deploy Plan

Deploy front.

PR #28791, opened 2026-07-10
Complete model tier resource and resolution policy.

Fraggle:
## Description

Completes `ModelsTierResource` with a clean CRUD surface and wires in the three-level resolution policy (user > groups > workspace).

- **`front/lib/model_tiers/tier_order.ts`**: New module with `getTierIndex`, `expandTiersUpTo`, `getMaxTierName`, `isTierWithinMax`, and label/description formatters. `DEFAULT_MAX_MODEL_TIER = "frontier"`.
- **`front/lib/model_tiers/resolve_allowed.ts`**: `resolveAllowedModelTiers` implements the priority policy and returns `{ tiers, source }` (`source` ∈ `"workspace" | "groups" | "user"`). Global-group grants are excluded from the groups layer by convention.
- **`front/lib/model_tiers/group_kinds.ts`**: `MODEL_TIER_OVERRIDE_GROUP_KINDS = ["regular", "provisioned"]` — only these kinds can carry tier overrides; `global` and `regular_auto` are excluded.
- **`ModelsTierResource`**: Replaces the old permission-based `grantToUser`/`revokeFromUser`/`grantToGroup`/`revokeFromGroup` with `set/clear` per user/group/workspace and a top-level `resolveAllowedTierNames(auth)`.
- **`STATIC_MODEL_TIERS` reclassification**: `gpt-4o`, `gpt-4-turbo`, `gpt-4.1` family, and mini/nano models at `medium`/`high` reasoning effort downgraded from `balanced` to `cost_efficient`.

## Tests

Vitest unit tests for `tier_order`, `resolve_allowed`, and `group_kinds`. Integration tests in `models_tier_resource.test.ts` updated to cover workspace defaults, set/clear lifecycle, and all three resolution priority cases (workspace baseline, group override, user override).

## Risk

Low. Front-only, no DB migration. The `ModelsTierResource` API rename only affects callers within the same PR (prior API from PR28781). The `STATIC_MODEL_TIERS` reclassifications lower the tier of several models — users with a workspace default of `frontier` are unaffected; only enforced paths would see tighter classification.

## Deploy Plan

Deploy front.

### DIFF
--- a/front/lib/model_tiers/tier_order.ts
+++ b/front/lib/model_tiers/tier_order.ts
@@ -1,0 +1,55 @@
+import {
+  MODELS_TIER_NAMES,
+  type ModelsTierName,
+} from "@app/lib/api/assistant/token_pricing/tiers";
+import { formatAsDisplayName } from "@app/types/shared/utils/string_utils";
+
+export const DEFAULT_MAX_MODEL_TIER: ModelsTierName = "premium";
+
+export function getTierIndex(tierName: ModelsTierName): number {
+  return MODELS_TIER_NAMES.indexOf(tierName);
+}
+
+export function expandTiersUpTo(maxTierName: ModelsTierName): ModelsTierName[] {
+  const maxIndex = getTierIndex(maxTierName);
+  if (maxIndex < 0) {
+    return [];
+  }
+
+  return [...MODELS_TIER_NAMES.slice(0, maxIndex + 1)];
+}
+
+export function getMaxTierName(
+  tierNames: readonly ModelsTierName[]
+): ModelsTierName | null {
+  if (tierNames.length === 0) {
+    return null;
+  }
+
+  let maxTierName: ModelsTierName = tierNames[0];
+  for (const tierName of tierNames) {
+    if (getTierIndex(tierName) > getTierIndex(maxTierName)) {
+      maxTierName = tierName;
+    }
+  }
+
+  return maxTierName;
+}
+
+export function isTierWithinMax(
+  tierName: ModelsTierName,
+  maxTierName: ModelsTierName
+): boolean {
+  return getTierIndex(tierName) <= getTierIndex(maxTierName);
+}
+
+export function formatMaxTierDescription(
+  maxTierName: ModelsTierName
+): string | undefined {
+  const lowerTiers = expandTiersUpTo(maxTierName).slice(0, -1);
+  if (lowerTiers.length === 0) {
+    return undefined;
+  }
+
+  return `Includes ${lowerTiers.map(formatAsDisplayName).join(", ")}`;
+}
--- a/front/lib/resources/models_tier_resource.ts
+++ b/front/lib/resources/models_tier_resource.ts
@@ -7,18 +7,42 @@ import {
   STATIC_MODEL_TIERS,
 } from "@app/lib/api/assistant/token_pricing/tiers";
 import type { Authenticator } from "@app/lib/auth";
+import { DustError } from "@app/lib/error";
+import { isModelTierOverrideGroupKind } from "@app/lib/model_tiers/group_kinds";
+import { resolveAllowedModelTiers } from "@app/lib/model_tiers/resolve_allowed";
+import {
+  DEFAULT_MAX_MODEL_TIER,
+  expandTiersUpTo,
+  getMaxTierName,
+} from "@app/lib/model_tiers/tier_order";
 import { GroupPermissionResource } from "@app/lib/resources/group_permission_resource";
-import type { GroupResource } from "@app/lib/resources/group_resource";
+import { GroupResource } from "@app/lib/resources/group_resource";
+import { MembershipResource } from "@app/lib/resources/membership_resource";
+import { GroupPermissionModel } from "@app/lib/resources/storage/models/group_permissions";
+import { makeSId } from "@app/lib/resources/string_ids";
+import { UserResource } from "@app/lib/resources/user_resource";
+import type {
+  GroupAllowedModelTiersType,
+  UserAllowedModelTiersType,
+} from "@app/types/api/model_tiers";
+import type { ModelId } from "@app/types/shared/model_id";
 import type { Result } from "@app/types/shared/result";
+import { Err, Ok } from "@app/types/shared/result";
 import type { UserType } from "@app/types/user";
 import assert from "assert";
 import type { Transaction } from "sequelize";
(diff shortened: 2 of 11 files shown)


Fraggle:

- #28763
    - #28779
        - #28781
            - #28791 ◀
                - #28792
                    - #28793
                        - #28794
                            - #28795
                                - #28796

front/lib/model_tiers/tier_order.ts
```diff
@@ -0,0 +1,71 @@
...
+export function getTierTitle(tierName: ModelsTierName): string {
+  switch (tierName) {
+    case "cost_efficient":
+      return "Cost efficient";
+    case "balanced":
+      return "Balanced";
+    case "premium":
+      return "Premium";
+    default:
+      return tierName;
```
Nils-Fedrigo: assertNever?

front/lib/resources/models_tier_resource.ts
```diff
@@ -63,12 +90,125 @@ export class ModelsTierResource {
     return STATIC_MODEL_TIERS[modelId][reasoningEffort] ?? null;
   }
 
+  static getTierNamesForModelConfig(model: {
```
Nils-Fedrigo: I'm not sure I understand the purpose of this jsut yet

front/lib/resources/models_tier_resource.ts
```diff
@@ -63,12 +90,125 @@ export class ModelsTierResource {
...
+  }
+
   private static getTierResourceId(tierName: ModelsTierName): number {
     const tier = this.getTier(tierName);
     assert(tier, `Unknown models tier: ${tierName}`);
     return tier.id;
   }
 
+  private static tierNameFromResourceId(
+    resourceId: number
```
Nils-Fedrigo, line 101: ressourceId = tierId ?

Nils-Fedrigo: ah ok it's the ressourceId of the GroupPermissionModel

front/lib/resources/models_tier_resource.ts
```diff
@@ -63,12 +90,125 @@ export class ModelsTierResource {
...
+    for (const tierName of MODELS_TIER_NAMES) {
+      await this.revokeFromGroup(auth, { group, tierName });
+    }
+  }
+
+  private static async getWorkspaceMaxAllowedTierName(
+    auth: Authenticator
+  ): Promise<ModelsTierName> {
+    const explicit = await this.getExplicitWorkspaceTierNames(auth);
+    return getMaxTierName(explicit) ?? DEFAULT_MAX_MODEL_TIER;
```
Nils-Fedrigo, line 162: getMaxTierName(explicit)
should never be null no?

front/lib/resources/models_tier_resource.ts
```diff
@@ -120,4 +260,439 @@ export class ModelsTierResource {
...
+  static async setUserMaxAllowedTier(
+    auth: Authenticator,
+    { userId, tierName }: { userId: string; tierName: ModelsTierName }
+  ): Promise<
+    Result<
+      undefined,
+      DustError<"invalid_request_error" | "user_not_found" | "user_not_member">
+    >
+  > {
+    this.assertIsAdmin(auth);
```
Nils-Fedrigo, line 248: nit:use a standard function?

front/lib/resources/models_tier_resource.ts
```diff
@@ -120,4 +260,439 @@ export class ModelsTierResource {
...
+          "user_not_member",
+          "User is not an active member of the workspace."
+        )
+      );
+    }
+
+    const userJson = user.toJSON();
+    await this.clearUserTierGrants(auth, userJson);
+
+    const result = await this.grantToUser(auth, {
```
Nils-Fedrigo, line 273: I wonder if (later) we should make grantToUser an upsert instead of an insert

front/lib/model_tiers/tier_order.ts
Fraggle: yep

front/lib/resources/models_tier_resource.ts
Nils-Fedrigo: same for grantToGroup

front/lib/resources/models_tier_resource.ts
Fraggle: `explicit` might be null here if you haven't set a tier for the workspace.
getMaxTierName() can be used for user, groups or workspace so it can't have a default.

front/lib/resources/models_tier_resource.ts
```diff
@@ -120,4 +260,439 @@ export class ModelsTierResource {
...
+  ): Promise<ModelsTierName[]> {
+    this.assertIsAdmin(auth);
+    return this.loadWorkspaceTierGrants(auth);
+  }
+
+  static async listWorkspaceMaxAllowedTierName(
+    auth: Authenticator
+  ): Promise<ModelsTierName> {
+    this.assertIsAdmin(auth);
+    return this.getWorkspaceMaxAllowedTierName(auth);
```
Nils-Fedrigo, line 389: not sure if that makes sense to gate it behind assertIsAdmin but we don't care that much

Nils-Fedrigo: because you already can have it from the user instead, anyway.

front/lib/resources/models_tier_resource.ts
Fraggle: You can't because the underlying system do not enforce it so you need to clear potential rows in our case because you are only allowed to use one tier per layer.

front/lib/resources/models_tier_resource.ts
Fraggle: ..only if it's not overriden either at the user or groups level, no ?

Nils-Fedrigo (approved):
overall seems there are a lot of helpers in all directions that we should aim to simplify. the system seems maybe too convoluted for what it should be. Let's refacto when you come back?

front/lib/resources/models_tier_resource.ts
Fraggle: let me check, maybe something lost in refactoring

Fraggle: indeed, dead code, removing

Fraggle:
> overall seems there are a lot of helpers in all directions that we should aim to simplify

Yep, that's what i was trying to say in my comment in the channel.

PR #28792, opened 2026-07-10
Switch runtime model enforcement to model tiers.

Fraggle:
## Description

Migrate model enforcement from the `advanced_models` access layer to the new `model_tiers` module. Creates `front/lib/model_tiers/enabled_models.ts` and `front/lib/model_tiers/access.ts` replacing their `advanced_models/` counterparts across all consumers: the `/models` route, `createOrUpgradeAgentConfiguration`, the Temporal `runModel` activity, and `resolveModel`.

Key behavioral changes over the old `advanced_models` approach:
- `withModelSelectability` now filters `supportedReasoningEfforts` per-tier rather than blocking entire models — a user capped at `cost_efficient` sees only the `light` effort on Sonnet instead of the model being hidden.
- `getModelTierAccessErrorForAgentConfiguration` takes `reasoningEffort` into account when resolving the tier, so the save-time guard matches the run-time guard exactly.
- Both paths are no-ops when the `models_picker` FF is off; legacy `advanced_models` code remains untouched.

## Tests

Added Vitest integration tests in `front/lib/model_tiers/enabled_models.test.ts` covering: FF disabled (full selectability), `cost_efficient` cap (only `light` effort on Sonnet), `balanced` cap (light/medium/high on Sonnet), and frontier-only models marked not selectable when capped at balanced.

## Risk

Low. Fully gated behind the `models_picker` feature flag. Legacy `advanced_models` code is left in place. No DB changes, no new migrations.

## Deploy Plan

Deploy front.

### DIFF
--- a/front/lib/model_tiers/access.ts
+++ b/front/lib/model_tiers/access.ts
@@ -1,0 +1,66 @@
+import type { Authenticator } from "@app/lib/auth";
+import { ModelsTierResource } from "@app/lib/resources/models_tier_resource";
+import type { GenericErrorContent } from "@app/types/assistant/agent";
+import type {
+  ModelConfigurationType,
+  ReasoningEffort,
+} from "@app/types/assistant/models/types";
+import { getMinimumReasoningEffort } from "@app/types/assistant/models/types";
+import type { WhitelistableFeature } from "@app/types/shared/feature_flags";
+
+export const MODEL_TIER_NOT_ENABLED_ERROR_CODE = "model_tier_not_enabled";
+
+export function buildModelTierAccessDeniedError(
+  agentName: string
+): GenericErrorContent {
+  return {
+    code: MODEL_TIER_NOT_ENABLED_ERROR_CODE,
+    message:
+      `Assistant ${agentName} uses a model tier that is not enabled for you. ` +
+      `Please contact your workspace admin or use another assistant.`,
+    metadata: {
+      errorTitle: "Model tier not enabled",
+      category: "unknown_error",
+    },
+  };
+}
+
+export async function getModelTierAccessErrorForAgentConfiguration(
+  auth: Authenticator,
+  {
+    agentName,
+    model,
+    reasoningEffort,
+    featureFlags,
+  }: {
+    agentName: string;
+    model: ModelConfigurationType;
+    reasoningEffort?: ReasoningEffort;
+    featureFlags: WhitelistableFeature[];
+  }
+): Promise<GenericErrorContent | null> {
+  if (!featureFlags.includes("models_picker")) {
+    return null;
+  }
+
+  const effectiveReasoningEffort =
+    reasoningEffort ??
+    getMinimumReasoningEffort(model.supportedReasoningEfforts);
+  const tierName = ModelsTierResource.getTierForModel(
+    model.modelId,
+    effectiveReasoningEffort
+  );
+
+  if (!tierName) {
+    return null;
+  }
+
+  const { tiers: allowedTierNames } =
+    await ModelsTierResource.resolveAllowedTierNames(auth);
+
+  if (allowedTierNames.includes(tierName)) {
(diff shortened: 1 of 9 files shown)


Fraggle:

- #28763
    - #28779
        - #28781
            - #28791
                - #28792 ◀
                    - #28793
                        - #28794
                            - #28795
                                - #28796

front/lib/model_tiers/access.ts
```diff
@@ -0,0 +1,66 @@
...
+    featureFlags: WhitelistableFeature[];
+  }
+): Promise<GenericErrorContent | null> {
+  if (!featureFlags.includes("models_picker")) {
+    return null;
+  }
+
+  const effectiveReasoningEffort =
+    reasoningEffort ??
+    getMinimumReasoningEffort(model.supportedReasoningEfforts);
```
Nils-Fedrigo, line 48: I'm not 100% sure but should this be the default reasoning effort? I'll cehck where this is used

Nils-Fedrigo: ok this should be a case that does not arrive in practise I guess🤞

Fraggle: It could if you have an old agent set on a model + reasoning effort that is no longer available to you.
I also hesistated but seems like the right fallback.

front/lib/model_tiers/enabled_models.test.ts
```diff
@@ -0,0 +1,105 @@
...
+    await MembershipFactory.associate(workspace, user, { role: "user" });
+    await ModelsTierResource.setUserMaxAllowedTier(adminAuth, {
+      userId: user.sId,
+      tierName,
+    });
+
+    return Authenticator.fromUserIdAndWorkspaceId(user.sId, workspace.sId);
+  }
+
+  it("keeps full reasoning efforts when models_picker is disabled", async () => {
```
Nils-Fedrigo, line 34: here it wuold have worked too with models_picker enabled ? because we don't restrict the user to a certain group ? worth having it in the chck

Nils-Fedrigo: like what is done in the next one

front/lib/model_tiers/enabled_models.ts
```diff
@@ -0,0 +1,145 @@
...
+    const tierName = ModelsTierResource.getTierForModel(
+      model.modelId as keyof typeof STATIC_MODEL_TIERS,
+      effort
+    );
+    if (tierName && allowedTierNamesSet.has(tierName)) {
+      supportedReasoningEfforts[effort] = true;
+    }
+  }
+
+  const allowedEfforts = getAvailableReasoningEfforts(
```
Nils-Fedrigo: a bit redundant with what you built above?

front/lib/model_tiers/enabled_models.ts
```diff
@@ -0,0 +1,145 @@
...
+
+  return models.map((model) => {
+    const restricted = restrictModelConfigToAllowedTiers(
+      model,
+      allowedTierNamesSet
+    );
+
+    return {
+      ...model,
+      ...restricted,
```
Nils-Fedrigo: I can't compute this pattern in my head but what would be the goal? Just return restricted?

Nils-Fedrigo (approved):
should this PR also delete the files on advanced that are not used anymore?

front/lib/model_tiers/enabled_models.ts
Fraggle: it return a list of ReasonningEffort from the list of true / false as defined in the configuration.
WDYM exactly ?

front/lib/model_tiers/enabled_models.ts
Fraggle: Hum, indeed maybe just restricted if enough, let me check.

PR #28798, opened 2026-07-10
[front] fix: stable sort LLM tool specification

aubin-tchoi:
## Description

This PR adds a stable sorting on the order of tools specifications sent to LLM providers to avoid bursting cache (example [here](https://langfuse.dust.tt/project/cmi4ft2qm00061707q4a9azyc/traces?filter=metadata%3BstringObject%3BconversationId%3B%3D%3BxnwzwGSBbN&peek=7b2df67767c1e73a5c237c81df5d39c9&timestamp=2026-07-09T18%3A54%3A27.649Z&observation=c07ff238fa87b0fb)).

Replay-only placeholder specs are sorted as well, they are sorted separately; replacing a spec with a placeholder will burst the cache anyway since the JSON schema and descriptions are different.

## Tests

- Tested locally.

## Risk

- Low.

## Deploy Plan

- Deploy front.

### DIFF
--- a/front/temporal/agent_loop/lib/run_model.test.ts
+++ b/front/temporal/agent_loop/lib/run_model.test.ts
@@ -268,6 +268,47 @@ describe("buildBaseSpecifications", () => {
     expect(specifications[0].eager).toBeUndefined();
   });
 
+  it("sorts specifications by tool name before model calls", () => {
+    const specifications = buildBaseSpecifications(
+      [
+        makeServerSideToolConfiguration("zeta_tool", {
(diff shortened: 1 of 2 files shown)


flvndvd (approved):
> replacing a spec with a placeholder will burst the cache anyway since the JSON schema and descriptions are different.

Only true if eagerly mounted, right? Which IIRC is not the case at all for placeholders we add?

aubin-tchoi:
yes correct, they should now be always deferred

PR #28799, opened 2026-07-10
Compact resource-shaped tool output before sending it to the model

flvndvd:
## Description

Despite improving our pruning strategy, we are still pruning way too often. Zooming-in, we noticed that MCP tools that return resource-type output, websearch results being the clearest example, currently have their entire raw structure serialized verbatim into the tool-result content sent to the model, including an internal `type` wrapper and a `mimeType` field that only exist so our own code can tell tool-output shapes apart.
Every resource schema in the codebase uses a fixed constant for that field, and wherever a real content type genuinely matters, like an actual file's format, it already lives under a differently named field that's left untouched, so nothing informative is lost. This only touches the branch that already falls back to a raw dump for the model, after existing text and vision-image handling has run, so tool-generated files and vision images are unaffected.

On a full websearch call this cuts a little over 1/6 of the tokens with no change in the actual information the model sees, which matters given a couple of web searches alone can produce enough tokens to force a cache-invalidating redaction almost every turn.

### DIFF
--- a/front/lib/api/assistant/conversation_rendering/helpers.test.ts
+++ b/front/lib/api/assistant/conversation_rendering/helpers.test.ts
@@ -474,7 +474,7 @@ describe("vision image rendering in getSteps", () => {
     }
   });
 
-  it("falls back to JSON when the model does not support vision", async () => {
+  it("falls back to JSON when the model does not support vision, flattened and without the resource wrapper or mimeType", async () => {
     const { auth, message, model, workspaceId, conversationId } =
       await buildVisionTest();
     const nonVisionModel = { ...model, supportsVision: false };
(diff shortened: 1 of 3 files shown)


front/lib/api/assistant/conversation_rendering/helpers.test.ts
```diff
@@ -516,3 +430,82 @@ describe("vision image rendering in getSteps", () => {
...
+      {
+        name: "Websearch Agent",
+        description: "An agent that searches the web",
+        model: { providerId: "anthropic", modelId: "claude-sonnet-4-6" },
+      }
+    );
+    const model = getSupportedModelConfig(agentConfig.model);
+    assert(model, "Expected a supported model configuration.");
+
+    const websearchResults = [
```
flvndvd, line 449: 🇫🇷

front/lib/api/assistant/conversation_rendering/helpers.ts
```diff
@@ -45,9 +45,25 @@ import type {
...
 const RENDER_ACTIONS_CONCURRENCY = 5;
 
+/**
+ * Drops the internal "type": "resource" wrapper and the top-level mimeType discriminator from a
+ * resource-shaped tool output item before it's serialized for the model. Every resource schema's
+ * mimeType is a fixed literal used only to distinguish shapes in our own code (see
+ * output_schemas.ts); it's never a real content type, and never what the model reads. Other item
+ * types (text, image) are returned unchanged.
+ */
+function compactResourceItem(item: CallToolResult["content"][number]) {
```
flvndvd, line 59: Actual changes

aubin-tchoi (approved):
LGTM

PR #28801, opened 2026-07-10
fix(front): let workspace admins manage agent editors

adrsimon:
Fixes https://github.com/dust-tt/tasks/issues/9541

## Description

Fixes workspace admins getting a 403 ("Failed to update editors" toast) when clicking "Become an editor" on the *"You are not an editor of this agent"* gate in the agent builder.

#28649 split write and admin access on editor groups: admins keep `admin` but no longer have `write`, and managing the editor list became a `canAdministrate` action. That PR updated the **skill** editors endpoint but missed the **agent** side, which shares the same `agent_editors` permission block. The agent editors PATCH and the agent editor member-management paths in `agent.ts` still gated on `canWrite`, so admins were blocked.

This gates agent editor management on `canAdministrate`, mirroring the skill side. Members still pass (they have `admin`); content-write checks are unchanged.

Confirmed in prod: `PATCH …/editors` returned 403 176× in 14 days with `usr.role: admin` and message *"Only editors of the agent or workspace admins can modify editors."*

## Tests

Added a regression test (admin who isn't an editor adds self → 200). It fails with the old `canWrite` gate and passes with `canAdministrate`. Agent + skill editors suites green; typecheck and biome clean.

## Risk

Low. Broadens editor management from `write` to `admin` on editor groups, which is exactly the pre-#28649 behavior for admins. Members are unaffected. Rollback is a straight revert.

## Deploy Plan

### DIFF
--- a/front-api/routes/w/[wId]/assistant/agent_configurations/[aId]/editors.test.ts
+++ b/front-api/routes/w/[wId]/assistant/agent_configurations/[aId]/editors.test.ts
@@ -137,6 +137,27 @@ describe("PATCH /api/w/:wId/assistant/agent_configurations/:aId/editors", () => 
     expect(editorIds).toContain(newEditor.sId);
   });
 
+  it("admin who is not an editor should be able to add self as editor", async () => {
+    // Regression test for the "Become an editor" flow: a workspace admin who is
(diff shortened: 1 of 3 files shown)


adrsimon:
dup from https://github.com/dust-tt/dust/pull/28775

PR #28804, opened 2026-07-10
fix: upgrade Temporal TS SDK 1.15.0 → 1.15.1 (workflow isolation fix)

adrsimon:
## Description

Fixes dust-tt/tasks#9558.

Temporal published an advisory for a **severe workflow-isolation bug** in the TypeScript SDK. State can leak across workflow executions (module-level variables retaining another execution's values, queries returning the wrong execution's data, workflows stuck on stale promises) when **both** hold:

1. Worker uses `reuseV8Context: true`.
2. Workflow bundle uses `webpack >= 5.108.0`.

Affected SDK releases include **1.15.0**, which we pin exactly across `connectors`, `front` and `front/workers-build`. The fix is **1.15.1**.

We are **not exposed today** (resolved webpack was `< 5.108.0`), but `@temporalio/worker` declares an unpinned transitive webpack range, so a routine lockfile refresh could silently flip both conditions true. Bumping the SDK removes the risk at the source.

Changes:
- Bump `@temporalio/{activity,client,common,worker,workflow,interceptors-opentelemetry}` `1.15.0 → 1.15.1` in `connectors/package.json`, `front/package.json`, `front/workers-build/package.json`.
- Regenerate lockfiles. `front`/`connectors` share the root workspaces `package-lock.json`; `front/workers-build/package-lock.json` was also reconciled off its stale `1.12.1` pin (itself an affected release, per the issue notes).
- Kept exact pins to match the existing convention in these files. The issue suggested a `^1.15.1` caret; happy to switch if reviewers prefer patches to flow in automatically.

⚠️ **`.npmrc`: `min-release-age` temporarily lowered `7 → 1`.** `@temporalio/*` 1.15.1 was published 2026-07-08, so the repo's 7-day supply-chain hold blocked it. There's a `# TODO` to rebump to `7` after 2026-07-15 (once 1.15.1 clears the 7-day window).

## Tests

- `npm install --package-lock-only` resolves cleanly in `front` and `connectors` (workspaces off root lockfile) and in `front/workers-build`.
- `front/workers-build` resolves webpack `5.108.4`, which is safe here because 1.15.1 patches the isolation bug in the SDK itself.
- `package-lock-check` pre-commit hook passes.

## Risk

Low. Patch-level SDK bump (1.15.0 → 1.15.1) with no API changes. Rollback is a straight revert of the version bumps and lockfiles.

## Deploy Plan

Standard deploy of `front` and `connectors`. Follow-up: revert `.npmrc` `min-release-age` back to `7` after 2026-07-15.


front/workers-build/package-lock.json
```diff
@@ -3296,9 +3275,9 @@
       }
     },
     "node_modules/terser": {
-      "version": "5.44.1",
```
aubin-tchoi, line 3299: the version of terser changes?

aubin-tchoi: maybe via temporal client > webpack actually

aubin-tchoi (approved):
LGTM but there are many changes, let's test on front-edge first

adrsimon:
tested on front-edge
- can launch a conversation
- can set up a trigger and create the temporal schedule 

temporal seems working 👍 merging

philipperolet:
hey @adrsimon sorry I missed that one and thanks @aubin-tchoi for reviewing

PR #28805, opened 2026-07-10
Analytics: expose parent message id in message table export

fabiencelier:
## Description

Closes https://github.com/dust-tt/tasks/issues/9443

Add parent message id for sub-agent in sub-conversations so user can track what triggerred the subconversation.
This value is only present for sub-agent in background mode (sub conversation).
For normal messages and sub-agent in handoff mode there is no need for the parent message id as there is a unique conversation id for them (and we don't currently track these parent id in ES anyway ^^)

## Tests

tested locally

Here is an export event, we can see the 2 calls to the agent "FrenchDelegateInBakcground" generated a call to a subagent "WebInfoSearcher" in a background conversation and these can be linked by the message parent id `iy1ZpKGdPl`

image

## Risk

low, this is a new field in the analytics export.

## Deploy Plan

deploy front

### DIFF
--- a/front/lib/api/analytics/messages_export.ts
+++ b/front/lib/api/analytics/messages_export.ts
@@ -21,6 +21,9 @@ interface AgentMessageDocument extends ElasticsearchBaseDocument {
   timestamp: string;
   agent_id: string;
   conversation_id: string;
+  // SIds of the agent messages that triggered this message through `run_agent`,
+  // direct parent first. Empty or absent for user-initiated messages.
+  ancestor_message_ids?: string[];
   user_id: string;
   context_origin: string;
   status: string;
@@ -37,6 +40,7 @@ export interface MessageExportRow {
   assistantName: string;
   assistantSettings: string;
   conversationId: string;
+  parentMessageId: string;
   userId: string;
   userEmail: string;
   source: string;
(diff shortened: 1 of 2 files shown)


front/lib/api/analytics/messages_export.ts
```diff
@@ -21,6 +21,9 @@ interface AgentMessageDocument extends ElasticsearchBaseDocument {
   timestamp: string;
   agent_id: string;
   conversation_id: string;
+  // SIds of the agent messages that triggered this message through `run_agent`,
```
aubin-tchoi: nit
```suggestion
  // Ids of the agent messages that triggered this message through `run_agent`,
```

front/lib/api/analytics/messages_export.ts
```diff
@@ -21,6 +21,9 @@ interface AgentMessageDocument extends ElasticsearchBaseDocument {
...
+  // direct parent first. Empty or absent for user-initiated messages.
```
aubin-tchoi, line 25: can't we pick between empty array or undefined?

aubin-tchoi (approved):
LGTM but don't you need to update the mapping?

front/lib/api/analytics/messages_export.ts
fabiencelier: We use empty array now days but some old docs may have undefined
This pre-exist this PR

PR #28807, opened 2026-07-10
fix(front): show agent default model as "Default" in picker even when outside the shortlist

Nils-Fedrigo:
## Description

When you switch the input bar to an agent whose default model is not part of the workspace's picker shortlist (e.g. an agent configured with GPT-5.5), the model picker used to display **Auto** instead of the agent's actual model. This happened because the picker only resolved the agent's model against the shortlist returned by `/api/w/[wId]/models`, and fell back to `Auto` when it wasn't found there.

The picker now falls back to the full supported-model catalog (`getSupportedModelConfig`) when the agent's configured model isn't in the shortlist. The model is then shown as **Default** in the trigger label and surfaced in the "Default from agent" section of the dropdown, exactly like a shortlisted default.

This only affects display/resolution of the *agent's* default. The `Auto` fallback still applies when the agent has no model or its model is `auto`, and the send-time payload is unchanged (the server already resolves models against the full catalog).

## Tests

Added `modelPickerUtils.test.ts` covering `resolveDefaultSelection`:
- Agent default that IS in the shortlist → resolves to `agent`.
- Agent default that is NOT in the shortlist (GPT-5.5) → falls back to the catalog and resolves to `agent` (previously `auto`), preserving reasoning effort.
- No agent model → `auto`.


Nils-Fedrigo:

- #28807 ◀

Fraggle (approved):

PR #28812, opened 2026-07-10
Add: ability to mark as done a task directly from the popup.

Fraggle:
## Description

Adds a `Checkbox` to `TaskDirectivePopoverBodyLoaded` so users can toggle a task done/todo directly from the task popup, without navigating to the Pod. When toggled, the task text renders with `line-through` + `text-faint` styling. The toggle is gated on `canEdit` (pod member + not archived).

The update uses `useUpdatePodTask` with an optimistic mutation: `mutateWorkspacePodTask` is called immediately with the projected state, then reverted on API error. `mutateWorkspacePodTask` is now threaded through as a typed prop (replacing the opaque `onTaskUpdated?: () => void`), which also lets the "start conversation" button trigger a full revalidation.

> 📸 Please attach a screenshot or GIF showing the checkbox interaction in the popup.

## Tests

Tested locally — toggled tasks done and back to todo from the popup, verified optimistic update and error rollback behavior.

## Risk

Low. Change is isolated to `TaskDirectiveBlock.tsx` and the pod task popup UI. The optimistic update pattern rolls back on API failure. No new API routes or DB changes.

## Deploy Plan

Deploy front.

### DIFF
--- a/front/components/markdown/TaskDirectiveBlock.tsx
+++ b/front/components/markdown/TaskDirectiveBlock.tsx
@@ -6,6 +6,7 @@ import { useUnifiedAgentConfigurations } from "@app/lib/swr/assistants";
 import {
   usePodMetadata,
   useStartPodTaskConversation,
+  useUpdatePodTask,
   useWorkspacePodTask,
 } from "@app/lib/swr/pods";
 import { timeAgoFrom } from "@app/lib/utils";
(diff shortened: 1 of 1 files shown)


Fraggle:

- #28812 ◀

matteotrab (approved):

PR #28815, opened 2026-07-10
Add: new tools to move conversation in/out of pods and change access.

Fraggle:
## Description

Three additions to the `pod_manager` MCP server:

- **`move_conversation_to_pod` / `move_conversation_out_of_pod`**: new tools (stake `never_ask`) that reassign a conversation's space to a target Pod or back to the user's personal space. Both default to the current agent conversation when no `conversationId` is provided. Backed by `moveConversationToProject` / `moveConversationOutOfProject` in `projects/conversations`.
- **`edit_information` gains `access` field**: agents can now flip a Pod between `restricted` and `open`. Respects the `areOpenPodsAllowed` workspace policy. Uses the new `getPodMemberAndEditorSIds` helper to preserve existing memberships when calling `pod.updatePermissions()`.
- **`create_pod` renames `visibility` → `access`** (`private` → `restricted`): aligns terminology with `edit_information`. This is a breaking schema change for any agent already calling `create_pod` with `visibility`.

UI: `PodEditInformationValidationDetails` shows a before/after diff card for each field being changed (title, description, access, pinned frame). `get_information` response now includes `access`.

## Tests

- Updated BM25 tool search test with queries for the three new intents.
- Snapshot updated for tool billing info and stakes.
- Tested manually.

## Risk

Medium. The `create_pod` `visibility` → `access` rename is a breaking schema change — agents currently generating `visibility` in their tool call will get a validation error until they re-read the updated tool definition. The `edit_information` access change calls `pod.updatePermissions()` which is the same path used by the UI; safe to rollback.

## Deploy Plan

Deploy front.

### DIFF
--- a/front/lib/api/actions/servers/pod_manager/metadata.ts
+++ b/front/lib/api/actions/servers/pod_manager/metadata.ts
@@ -21,6 +21,8 @@ export const POD_MANAGER_SERVER_NAME = "pod_manager" as const;
 export const UPDATE_MEMBERS_TOOL_NAME = "update_members" as const;
 export const LIST_MEMBERS_TOOL_NAME = "list_members" as const;
 export const SEMANTIC_SEARCH_TOOL_NAME = "semantic_search" as const;
+export const EDIT_INFORMATION_TOOL_NAME = "edit_information" as const;
+export const MOVE_CONVERSATION_TOOL_NAME = "move_conversation" as const;
 
 export const POD_MANAGER_TOOLS_METADATA = createToolsRecord({
   add_content_node: {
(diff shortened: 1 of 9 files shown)


Fraggle:

- #28815 ◀

front/lib/api/actions/servers/pod_manager/metadata.ts
```diff
@@ -440,6 +452,53 @@ export const POD_MANAGER_TOOLS_METADATA = createToolsRecord({
     toolCostCategory: "basic",
     freeUsage: true,
   },
+  [MOVE_CONVERSATION_TO_POD_TOOL_NAME]: {
```
matteotrab: What do you think of a single move conversation tool to move conversations out and in pods?

matteotrab (approved):

front/lib/api/actions/servers/pod_manager/metadata.ts
Fraggle: Yep good idea, less tools is always good !

PR #28816, opened 2026-07-10
Gate Anthropic tool search on real per-model support

flvndvd:
## Description

The `anthropic_tool_search` feature flag alone was enough to make the agent loop send `defer_loading`/the `tool-search-tool` entry on every Anthropic request, with no check for whether the specific model in use actually supports the feature. Anthropic only supports tool search on a handful of recent models, so a workspace with the flag on but an agent configured for an older Claude model would have gotten a request Anthropic likely rejects outright.


aubin-tchoi (approved):
LGTM

PR #28817, opened 2026-07-10
Exclude deferred tool schemas from the token-budget estimate

flvndvd:
## Description

Based on https://github.com/dust-tt/dust/pull/28816. Anthropic's tool-search feature excludes deferred tool definitions from the model's active context until they're actually discovered mid-conversation, confirmed directly from their own documentation: only tool definitions loaded into context by a search count as input tokens, and deferred ones sit outside the prefix entirely. Our own token-budget estimate didn't reflect that at all, it tokenized every available tool's full schema regardless of eager or deferred status, which meant the budget looked meaningfully tighter than it really was on any workspace with tool search active and a large tool catalog, since most tools in a typical setup end up deferred.

The fix only changes what gets fed into the existing token-counting call for the tools portion of the budget: eager specs plus the tool-search tool itself when tool search is active, everything as before when it isn't. 
## Risk

Low risk, since as of today, we use very conservative context window for most models.

### DIFF
--- a/front/temporal/agent_loop/lib/run_model.test.ts
+++ b/front/temporal/agent_loop/lib/run_model.test.ts
@@ -5,6 +5,7 @@ import { ANTHROPIC_PROVIDER_ID } from "@app/lib/api/llm/clients/anthropic/types"
 import {
   buildBaseSpecifications,
   buildSpecificationsWithReplayPlaceholders,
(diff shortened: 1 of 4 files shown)


aubin-tchoi (approved):
LGTM

PR #28818, opened 2026-07-10
Share tool approval handling across run contexts

spolu:
## Description

Decouple tool stake resolution from agent messages and centralize approval-event payload construction across agent-loop and sandbox child calls. This prepares sandbox functions to reuse the same approval policy and event shape without changing their execution behavior yet.

## Tests

- `cd front && NODE_ENV=test npm test lib/api/sandbox/create_child_action.test.ts lib/actions/tool_status.test.ts`
- tested locally

## Risk

Low. This is a behavior-preserving refactor covered by the existing approval tests.

## Deploy Plan

- deploy `front`

### DIFF
--- a/front/lib/actions/tool_approval_events.ts
+++ b/front/lib/actions/tool_approval_events.ts
@@ -1,0 +1,67 @@
+import type { MCPToolConfigurationType } from "@app/lib/actions/mcp";
+import {
+  getInternalMCPServerDisplayedAs,
+  getInternalMCPServerNameFromSId,
+} from "@app/lib/actions/mcp_internal_actions/constants";
+import type { MCPApproveExecutionEventData } from "@app/lib/actions/mcp_internal_actions/events";
+import { getApprovalArgsLabel } from "@app/lib/actions/tool_approval_labels";
+import type { Authenticator } from "@app/lib/auth";
+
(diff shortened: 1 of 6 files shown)


front/lib/actions/tool_approval_events.ts
```diff
@@ -0,0 +1,67 @@
...
+      toolName: toolConfiguration.originalName,
+      mcpServerName: toolConfiguration.mcpServerName,
+      agentName: "agent",
+      icon: toolConfiguration.icon,
+      displayedAs: getInternalMCPServerDisplayedAs(
+        toolConfiguration.toolServerId
+      ),
+    },
+    argumentsRequiringApproval,
+    approvalArgsLabel: await getApprovalArgsLabel({
```
flvndvd: `nit` can we extract await into it's own operation before the return. It's better to debug when you use breakpoints.

flvndvd (approved):

front/lib/actions/tool_approval_events.ts
spolu: Sure thing!

spolu: done

spolu:
cc @PopDaph for viz

PR #28820, opened 2026-07-10
[front] fix: keep system skill tools eager

aubin-tchoi:
## Description

Tools exposed through skills are deferred by clearing `eager`, even when their tool metadata marks them eager. System skills should be exempt from that behavior.

This PR keeps system-skill server keys out of the `isFromSkillServer` classification used to clear `eager`. Regular skill-only tools remain deferred, and system origin wins when the same server is exposed by both a system skill and an enabled skill.

## Tests

- Updated existing testts.

## Risk

- Low.

## Deploy Plan

- Deploy front.

### DIFF
--- a/front/lib/api/assistant/skill_actions.ts
+++ b/front/lib/api/assistant/skill_actions.ts
@@ -299,12 +299,6 @@ export async function resolveSkillMCPServers(
     }
   );
 
-  const activeSkills = [...systemSkills, ...enabledSkills];
-
-  if (activeSkills.length === 0) {
-    return { skillServers: [], systemSkillServers: [] };
-  }
-
   return getSkillServers(auth, {
     agentConfiguration,
     enabledSkills,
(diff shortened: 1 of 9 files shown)


front/lib/api/assistant/skill_actions.ts
```diff
@@ -67,15 +76,25 @@ export async function getSkillServers(
...
+        }),
+        isSystemSkill: systemSkillIds.has(skill.sId),
+      };
     })
   );
 
+  const skillServers = mcpServersWithOrigin
+    .filter(({ isSystemSkill }) => !isSystemSkill)
+    .map(({ server }) => server);
+  const systemSkillServers = mcpServersWithOrigin
```
flvndvd: Can't we get the skill scope form the skill itself?

flvndvd (approved):

PR #28821, opened 2026-07-10
Rewrite the Pod Functions skill for accuracy and clarity

flvndvd:
## Description

The Pod Functions skill previously led with implementation detail rather than explaining what the capability actually unblocks, and got the calling convention for Frames wrong. This rewrite leads with the three concrete reasons to reach for a pod function: giving a Frame a server side capability it cannot run itself, collapsing a repeated multi step operation behind one typed call, and persisting data that a Frame should keep reflecting even when it changes elsewhere.

The instructions now correctly explain that a Frame must address a function by its fully qualified identifier, since the bare slug the conversation side tools accept does not resolve from a Frame's own runtime, and they describe the actual response shape a Frame gets back today, which is the raw sandbox runner envelope rather than the parsed output the function declares.

@id13 left a TODO for you on DB interactions.

### DIFF
--- a/front/lib/resources/skill/code_defined/global/pod_functions.ts
+++ b/front/lib/resources/skill/code_defined/global/pod_functions.ts
@@ -9,38 +9,33 @@ export const podFunctionsSkill = {
   kind: "global",
   name: "Pod Functions",
   userFacingDescription:
-    "Give Frames a typed backend, or a way to turn a multi-step task into one reusable function.",
+    "Run hosted functions on the Pod's Computer that can persist data and call other tools.",
   agentFacingDescription:
-    "A pod function is a versioned, typed server-side function stored on the Pod, callable by " +
-    "slug from this conversation or from a Frame's own runtime. It unblocks a Frame from having " +
-    "to run something itself: reaching a server-side capability its browser sandbox cannot (an " +
-    "external API call, a workspace secret, logic no client code should hold), reusing a " +
-    "multi-step operation (several API calls, some branching, a non-trivial computation) behind " +
-    "one typed contract instead of redoing it by hand every time, or reading and writing data " +
-    "that outlives a single conversation, so a Frame calling the function keeps showing current " +
-    "state even when that state was last written from somewhere else.",
-  // TODO(POD_FUNCTION: JD): the "Persisting state across calls" section below is a placeholder header only;
-  // fill in the SQLite/db() story once it's finalized.
-  instructions: `Pod functions are versioned, typed callables stored on the Pod. Each one is a
-TypeScript module with zod-typed input and output, bundled on the Pod's sandbox at publish time and
-reusable across conversations in the same Pod, callable from a Frame's own runtime or directly
-from this conversation with the call tool described below.
+    "A pod function is a hosted function that runs on the Pod's own Computer, shared across " +
+    "every conversation in the Pod, with the ability to persist data and call other tools. It's " +
+    "callable by slug from this conversation or from a Frame's own runtime.",
+  // TODO(POD_FUNCTION: JD/spolu): the SQLite/db() story for "Persisting state across calls"
+  // below still needs to be filled in.
+  instructions: `Pod functions are versioned, typed functions published on the Pod's own
+Computer: a persistent environment shared across every conversation in the Pod, not the one
+scoped to this conversation. Each one is a TypeScript module with zod-typed input and output,
+bundled at publish time and reusable across conversations in the same Pod, callable from a
+Frame's own runtime or directly from this conversation with the call tool described below.
 
 Reach for a pod function instead of inline code or an ad hoc tool call when any of these apply:
 
-- A Frame needs a server-side capability it cannot run inside its own browser sandbox: an
-  external API call, a workspace secret, or logic no client-side code should hold.
-- The same multi-step operation (several API calls, some branching, a non-trivial computation)
-  keeps getting redone by hand and is worth collapsing behind one typed input/output contract.
-- Data needs to persist beyond one call and be shared across writers: written from this
-  conversation or another conversation in the same Pod, then read back by a Frame that always
-  shows the latest state without needing the conversation that wrote it (see "Persisting state
-  across calls" below).
+- A Frame needs a server-side capability it cannot run inside its own browser sandbox: calling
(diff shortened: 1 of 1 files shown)


id13 (approved):
LGTM

front/lib/resources/skill/code_defined/global/pod_functions.ts
```diff
@@ -9,21 +9,38 @@ export const podFunctionsSkill = {
   kind: "global",
   name: "Pod Functions",
   userFacingDescription:
-    "Author and publish reusable, schema-typed functions that run on the Pod's sandbox.",
+    "Give Frames a typed backend, or a way to turn a multi-step task into one reusable function.",
```
spolu: "multi-step task into reusable funciton" seems misleading? that's not quite the goal since there is no way to do "one step" anyway today?

=> Author and publish reusable, schema-typed functions that run on the Pod's sandbox, providing frames with a backend

Maybe?

front/lib/resources/skill/code_defined/global/pod_functions.ts
```diff
@@ -9,21 +9,38 @@ export const podFunctionsSkill = {
...
   agentFacingDescription:
-    "Publish reusable TypeScript functions to the Pod and inspect the ones already published. " +
-    "A function is a typed input/output handler bundled and stored on the Pod to back Frames: a " +
-    "Frame can call a published function to reach a server-side capability behind a typed contract. " +
-    "Enable this skill to turn a sandbox script into a published, named function a Frame can call.",
+    "A pod function is a versioned, typed server-side function stored on the Pod, callable by " +
+    "slug from this conversation or from a Frame's own runtime. It unblocks a Frame from having " +
```
spolu: by slug (pod-{podId}/{slugName})

spolu: > It unblocks a Frame from having to run something itself

Weird really. The real thing that it provides are hosted functions executed on a shared sandbox (with stateful storage and access to dsbx to run tools)

front/lib/resources/skill/code_defined/global/pod_functions.ts
```diff
@@ -9,21 +9,38 @@ export const podFunctionsSkill = {
...
+    "to run something itself: reaching a server-side capability its browser sandbox cannot (an " +
+    "external API call, a workspace secret, logic no client code should hold), reusing a " +
+    "multi-step operation (several API calls, some branching, a non-trivial computation) behind " +
```
spolu: The multi-step trope I find really misleading as there is no non multi-step alternatives :)

front/lib/resources/skill/code_defined/global/pod_functions.ts
```diff
@@ -9,21 +9,38 @@ export const podFunctionsSkill = {
...
+    "one typed contract instead of redoing it by hand every time, or reading and writing data " +
+    "that outlives a single conversation, so a Frame calling the function keeps showing current " +
+    "state even when that state was last written from somewhere else.",
```
spolu: Overall I think the description can be shortened. Looks like a LOT of tokens?

front/lib/resources/skill/code_defined/global/pod_functions.ts
```diff
@@ -9,21 +9,38 @@ export const podFunctionsSkill = {
...
+  // TODO(POD_FUNCTION: JD): the "Persisting state across calls" section below is a placeholder header only;
+  // fill in the SQLite/db() story once it's finalized.
   instructions: `Pod functions are versioned, typed callables stored on the Pod. Each one is a
 TypeScript module with zod-typed input and output, bundled on the Pod's sandbox at publish time and
```
spolu: What's the pod sandbox? :)

front/lib/resources/skill/code_defined/global/pod_functions.ts
```diff
@@ -9,21 +9,38 @@ export const podFunctionsSkill = {
...
-reusable across conversations in the same Pod. They exist to back Frames: a Frame can call a
-published function to reach a server-side capability behind a typed contract.
+reusable across conversations in the same Pod, callable from a Frame's own runtime or directly
+from this conversation with the call tool described below.
+
+Reach for a pod function instead of inline code or an ad hoc tool call when any of these apply:
+
+- A Frame needs a server-side capability it cannot run inside its own browser sandbox: an
+  external API call, a workspace secret, or logic no client-side code should hold.
+- The same multi-step operation (several API calls, some branching, a non-trivial computation)
```
spolu: meh

front/lib/resources/skill/code_defined/global/pod_functions.ts
```diff
@@ -9,21 +9,38 @@ export const podFunctionsSkill = {
...
+  keeps getting redone by hand and is worth collapsing behind one typed input/output contract.
+- Data needs to persist beyond one call and be shared across writers: written from this
```
spolu: I would add precise exampels such as storing files or accessing a database through sqlite

spolu (commented):
Some comments on descirption and instructions a bit.

The rest on "calling a function" LGTM

front/lib/resources/skill/code_defined/global/pod_functions.ts
id13: It's on me to do that

front/lib/resources/skill/code_defined/global/pod_functions.ts
flvndvd: What about the model does always the same 10 steps to pull the data. Wouldn't it be a great use case to abstract those 10 calls behind one function?

front/lib/resources/skill/code_defined/global/pod_functions.ts
flvndvd: I needed a way to still expose in the description that it can be used through a frame, will reword.

front/lib/resources/skill/code_defined/global/pod_functions.ts
spolu: Sure but it's a side benefit. Here the goal is to explain what is the core feature. If we use the side-benefit to do so I think it reduce the clarity of the explanation

PR #28823, opened 2026-07-10
feat(analytics): index agent tag ids

sfriquet:
## Description

Refs https://github.com/dust-tt/tasks/issues/8167

Adds an `agent_tag_ids` field to the `front.agent_message_analytics` Elasticsearch
index so analytics can be sliced/aggregated by agent tag (e.g. "message volume
across all agents tagged `post-sales).

## Tests
- added tests
- checked locally that ES index is populated with the tags ids, e.g
```
GET front.agent_message_analytics/_search
{
  "size": 5,
  "sort": [{ "timestamp": "desc" }],
  "query": { "exists": { "field": "agent_tag_ids" } },
  "_source": ["message_id","agent_id","agent_version","agent_tag_ids","timestamp"]
}
```

## Deploy Plan

1. apply ES migration
2. merge & deploy

### DIFF
--- a/front/lib/resources/tags_resource.ts
+++ b/front/lib/resources/tags_resource.ts
@@ -1,4 +1,5 @@
 import type { Authenticator } from "@app/lib/auth";
+import { AgentConfigurationModel } from "@app/lib/models/agent/agent";
 import { TagAgentModel } from "@app/lib/models/agent/tag_agent";
 import { TagModel } from "@app/lib/models/tags";
 import { BaseResource } from "@app/lib/resources/base_resource";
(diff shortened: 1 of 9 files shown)


front/lib/resources/tags_resource.ts
```diff
@@ -220,6 +221,43 @@ export class TagResource extends BaseResource<TagModel> {
     );
   }
 
+  /**
+   * List the tags attached to a specific agent configuration version, identified
+   * by its sId and version.
+   */
+  static async listForAgentVersion(
```
fabiencelier, line 228: Can you please add a small unit test for this new method if it's not too long ?
I see there is no existing test for tags_resource.ts it's usually good practice to test the resource
It's not a strict rule but it's often easy to test the resource methods + they are shared by several so it's worth making sure they work.

fabiencelier (approved):
LGTM
great work

PR #28824, opened 2026-07-10
Enforce sandbox function tool stakes

spolu:
## Description

Apply the shared tool stake policy when sandbox functions call MCP tools. Approval-required actions are persisted as blocked and publish a scoped approval event to the invocation stream; implicitly allowed actions continue to launch immediately.

Workflow launch after approval remains in the follow-up validation endpoints.

## Tests

- `cd front-api && NODE_ENV=test npm test 'routes/v1/w/[wId]/sandbox/actions/call.test.ts'`

## Risk

Approval-required sandbox function calls remain blocked until resolved by the follow-up endpoints.
None since SandboxFunction is not live yet.

## Deploy Plan

- deploy `front`

### DIFF
--- a/front/lib/api/sandbox_functions/create_sandbox_function_mcp_action.ts
+++ b/front/lib/api/sandbox_functions/create_sandbox_function_mcp_action.ts
@@ -1,7 +1,11 @@
 import type { ServerSideMCPToolConfigurationType } from "@app/lib/actions/mcp";
 import { MCP_TOOL_CONFIGURATION_FIELDS_TO_OMIT } from "@app/lib/actions/mcp";
 import { buildToolConfigurationsFromRawTools } from "@app/lib/actions/mcp_actions";
+import type { SandboxFunctionMCPApproveExecutionEvent } from "@app/lib/actions/mcp_internal_actions/events";
 import { validateToolInputs } from "@app/lib/actions/mcp_utils";
+import { makeMCPApproveExecutionEventBase } from "@app/lib/actions/tool_approval_events";
+import { getExecutionStatusFromConfig } from "@app/lib/actions/tool_status";
+import { publishSandboxFunctionInvocationEvent } from "@app/lib/api/sandbox_functions/events";
 import type { Authenticator } from "@app/lib/auth";
 import { MCPServerViewResource } from "@app/lib/resources/mcp_server_view_resource";
 import { SandboxFunctionInvocationResource } from "@app/lib/resources/sandbox_function_invocation_resource";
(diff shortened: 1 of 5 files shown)


front/lib/api/sandbox_functions/create_sandbox_function_mcp_action.ts
```diff
@@ -181,18 +184,60 @@ export async function createSandboxFunctionMCPAction(
     );
   }
 
+  const toolConfiguration = toolConfigurationRes.value;
+  const { status: executionStatus } = await getExecutionStatusFromConfig(auth, {
```
PopDaph, line 188: Two questions on the stake evaluation here:

1. Since we don't pass context, medium-stake tools will always block (per-agent approvals have no agent to key on), and low auto-approval only works when the invocation token carries a uId. That's invisible at the call site so worth a one-line comment?

````
  const toolConfiguration = toolConfigurationRes.value;
  // No agent context here: medium-stake tools always block (per-agent approvals have no agent to
  // key on), and low-stake auto-approval only applies when the invocation token carries a user.
  const { status: executionStatus } = await getExecutionStatusFromConfig(auth, {
    actionConfiguration: toolConfiguration,
  });
````

2. OOC, do we eventually want medium approvals to be rememberable in this context, keyed per (user, MCP server, tool, sandbox function, argument values) i.e. the function takes the agent's place as the approval subject? Or is always-ask the intended long-term behavior for sandbox functions?

PopDaph (approved):
👍🏻

front/lib/api/sandbox_functions/create_sandbox_function_mcp_action.ts
spolu: On 2 I think we will want some form of memory yes. Maybe the frameId should be the subject?

spolu: On 1/2 will dig more 👍🙏

spolu: Okay so there's a rabbit hole with agentId here :)

spolu:
@PopDaph moving forward with this one will treat the context separately 👍

spolu:
TODO: clean-up agentId + context passing for sandbox functions

front/lib/api/sandbox_functions/create_sandbox_function_mcp_action.ts
spolu: Handled here: https://github.com/dust-tt/dust/pull/28859

PR #28813, opened 2026-07-10
Surface tool approval and personal authentication in the viz wrapper

spolu:
## Description

Surface sandbox-function tool approval and personal-authentication requests over the active Frame.

- Queue blocking SSE events per invocation and render the existing approval and authentication cards over the Frame.
- Key and remove queue entries by SSE event ID so follow-up events for the same action are preserved and card state resets correctly.
- Post validation and authentication resolutions through colocated `lib/swr/sandbox_functions` hooks.

Backend event emission and resolution endpoints land separately; until then this UI remains unreachable.

## Tests

N/A, tested locally.

## Risk

Low. Limited to the feature-flagged `sandbox_functions` flow.

## Deploy Plan

- deploy `front`

PR #28825, opened 2026-07-10
Colocate tool approval hooks in SWR

spolu:
## Description

Move tool authentication and validation mutations into a shared SWR module. A discriminated `agent_loop` / `sandbox_function` context lets both run types use the same two hooks while preserving their route payloads, notifications, and already-resolved handling.

The authentication outcome type now comes from its canonical backend definition.

## Tests

- `cd front && NODE_ENV=test npm test components/assistant/conversation/MCPServerPersonalAuthenticationRequired.test.tsx`
- tested locally for `agent_loop`

## Risk

Low. Request routing remains exhaustive and behavior-preserving.

## Deploy Plan

- deploy `front`

PR #28829, opened 2026-07-10
Sandbox functions: tool approval validation endpoint

PopDaph:
Follow up of #28824, which blocks approval-requiring sandbox function tool actions at creation (no workflow launched) and publishes the approval event to the invocation stream. This PR adds the resolution half: the validate endpoint the merged approval card (#28813) already calls.

`POST /api/w/:wId/sandbox-functions/:functionIdOrSlug/invocations/:invocationId/actions/:actionId/validate` with `{ approved: "approved" | "rejected" | "always_approved" }`:

- Approved: compare-and-swap `blocked_validation_required` -> `running`, then the first `launchSandboxFunctionToolWorkflow` for the action. dsbx keeps polling (blocked reads as `202 pending`), so resumption needs no client cooperation.
- Rejected: CAS -> `denied`; the poll surfaces it as `403 rejected` and the in-sandbox tool call fails.
- `always_approved`: records a global always-approve for low-stake tools (same keying as the conversation flavor). Medium-stake records are keyed on an agent, which a sandbox function has none of: treated as a one-time approval.
- The pending approval event is removed from the invocation Redis channel so SSE history replay does not resurface the approval card.
- Concurrent resolutions have exactly one winner through the CAS (new `updateStatusFromExpected` on the action resource, mirroring the agent one); losers succeed silently, and a late duplicate gets `action_not_blocked`, which the client treats as success.

Authorization note: anyone who can read the pod space can resolve an approval, unlike the conversation flavor where only the parent message author may. Invocations record no initiating user, and this matches the invocation SSE route's access model. `always_approved` records the low-stake approval for the resolving user.

Audit is deferred with a dated TODO: the `tool.approval_resolved` schema requires agent and conversation targets that do not exist for functions.

Next PR: `resolve-authentication` for `tool_personal_auth_required`, pending the auth-blocked status addition to the base status union (to sync with the event delivery slice).

## Tests

Endpoint tests: approve (status, workflow launch, event removal), reject (denied, no launch), always-approve recording for low stake, double validation -> `action_not_blocked`, cross-invocation action -> 404, non-blocked action -> 400. Manual e2e locally: blocked action approved through the endpoint, workflow executed the tool and the poll reached `succeeded`; rejected action polled as `403 rejected`.

## Risk

None, SandboxFunction is not live yet. The endpoint is unreachable until #28824 creates blocked actions.

## Deploy Plan

- [ ] Deploy front

### DIFF
--- a/front-api/routes/w/[wId]/sandbox-functions/[functionId]/invocations.ts
+++ b/front-api/routes/w/[wId]/sandbox-functions/[functionId]/invocations.ts
@@ -148,7 +148,7 @@ app.post(
 
 /** @ignoreswagger */
 app.post(
-  "/:invocationId/actions/:actionId/validate",
+  "/:invocationId/actions/:actionId/validate-action",
   validate("param", ValidateActionParamsSchema),
   validate("json", ValidateActionBodySchema),
   async (ctx): HandlerResult<{ success: boolean }> => {
--- a/front-api/routes/w/[wId]/sandbox-functions/[functionId]/invocations.ts
+++ b/front-api/routes/w/[wId]/sandbox-functions/[functionId]/invocations.ts
@@ -1,9 +1,12 @@
+import { MCP_VALIDATION_OUTPUTS } from "@app/lib/actions/constants";
 import { publishSandboxFunctionInvocationEvent } from "@app/lib/api/sandbox_functions/events";
+import { validateSandboxFunctionAction } from "@app/lib/api/sandbox_functions/validate_action";
 import { SandboxFunctionResource } from "@app/lib/resources/sandbox_function_resource";
 import type {
   PostSandboxFunctionInvocationRequestBody,
   PostSandboxFunctionInvocationResponseBody,
 } from "@app/types/api/sandbox_functions";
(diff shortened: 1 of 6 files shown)


front-api/routes/w/[wId]/sandbox-functions/[functionId]/invocations.ts
```diff
@@ -131,4 +146,63 @@ app.post(
   }
 );
 
+/** @ignoreswagger */
+app.post(
+  "/:invocationId/actions/:actionId/validate",
```
Comment (spolu) [inline front-api/routes/w/[wId]/sandbox-functions/[functionId]/invocations.ts]: to be perfectly aligned shouldn't the name be validate-action? I see I put `validate` here: https://github.com/dust-tt/dust/pull/28825/changes#diff-b4553a14fa149a51e7e7266a099e4c83150ce6208b7d0f1933b4893fee514f7bR95   but can be easily changed.

spolu (approved):
LGTM

front-api/routes/w/[wId]/sandbox-functions/[functionId]/invocations.ts
Comment (PopDaph) [reply front-api/routes/w/[wId]/sandbox-functions/[functionId]/invocations.ts]: done!

PR #28830, opened 2026-07-10
[front] fix: render skill avatars in pod settings

aubin-tchoi:
## Description

Skill avatars in Pod settings were passed through generic `Avatar` and `Icon` wrappers, causing nested rendering and distorted sizing.

Render the avatars directly at the intended `xs` size in both the selected skill pills and the default skills picker, aligning with the existing capabilities picker.

We can probably share even more of the UI with the capabilities picker, will tackle as a follow up.

Before
image

After
image

## Tests

- Tested locally.

## Risk

- Low.

## Deploy Plan

- Deploy front.


xebeohp (approved):
LGTM

PR #28832, opened 2026-07-10
feat: add regular manual groups

matteotrab:
## Description

This PR introduces a new `regular_manual` group kind that can be created manually from the UI to grant specific permissions to users, distinct from `regular_auto` groups which are managed programmatically by workspace admins.

- Adds `regular_manual` to `GROUP_KINDS`, `GroupKindCodec`, and the `isRegularManualGroupKind()` helper in `front/types/groups.ts`.
- Adds an `isRegularManual()` method to `GroupResource` and defines its permission model (members get `read`, admins and business admins get `read/write/admin`).
- Extends `addMembers` and `removeMembers` to allow `regular_manual` groups
- Includes `regular_manual` in the default group kinds fetched by `listAllWorkspaceGroups` and exposes it in the poke API endpoints (groups listing and user groups).

## Tests

Manually.

## Risks
Regular manual groups might be fetched when not needed.

## Deploy Plan

Standard deployment.

### DIFF
--- a/front-api/routes/poke/workspaces/[wId]/groups/index.ts
+++ b/front-api/routes/poke/workspaces/[wId]/groups/index.ts
@@ -1,5 +1,11 @@
 import type { PokeListGroups } from "@app/lib/api/poke/groups";
 import { GroupResource } from "@app/lib/resources/group_resource";
+import {
+  GROUP_KINDS,
+  isAgentEditorGroupKind,
+  isSkillEditorGroupKind,
+  isSystemGroupKind,
+} from "@app/types/groups";
 import { pokeApp } from "@front-api/middlewares/ctx";
 import type { HandlerResult } from "@front-api/middlewares/utils";
 
@@ -13,13 +19,12 @@ app.get("/", async (ctx): HandlerResult<PokeListGroups> => {
   const auth = ctx.get("auth");
 
   const groups = await GroupResource.listAllWorkspaceGroups(auth, {
-    groupKinds: [
-      "global",
-      "regular_auto",
-      "regular_manual",
-      "space_editors",
-      "provisioned",
-    ],
+    groupKinds: GROUP_KINDS.filter(
+      (kind) =>
+        !isSystemGroupKind(kind) &&
+        !isAgentEditorGroupKind(kind) &&
+        !isSkillEditorGroupKind(kind)
+    ),
   });
 
   return ctx.json({ groups: groups.map((g) => g.toJSON()) });
--- a/front-api/routes/poke/workspaces/[wId]/groups/index.ts
+++ b/front-api/routes/poke/workspaces/[wId]/groups/index.ts
@@ -13,7 +13,13 @@ app.get("/", async (ctx): HandlerResult<PokeListGroups> => {
   const auth = ctx.get("auth");
 
   const groups = await GroupResource.listAllWorkspaceGroups(auth, {
-    groupKinds: ["global", "regular_auto", "space_editors", "provisioned"],
+    groupKinds: [
+      "global",
+      "regular_auto",
+      "regular_manual",
+      "space_editors",
+      "provisioned",
+    ],
   });
 
   return ctx.json({ groups: groups.map((g) => g.toJSON()) });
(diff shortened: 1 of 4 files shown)


matteotrab:

- #28832 ◀
    - #28833
        - #28837
            - #28839
                - #28840
                    - #28843

front/lib/resources/group_resource.ts
```diff
@@ -1554,24 +1557,6 @@ export class GroupResource extends BaseResource<GroupModel> {
       );
     }
 
-    // Users can only be added to regular_auto, space_editors, agent_editors, skill_editors or provisioned groups.
-    if (
```
matteotrab, line 1558: We have an assert above, so this if should not be needed

front/lib/resources/group_resource.ts
```diff
@@ -1725,24 +1711,6 @@ export class GroupResource extends BaseResource<GroupModel> {
       );
     }
 
-    // Users can only be removed from regular_auto, space_editors, agent_editors, skill_editors or provisioned groups.
-    if (
```
matteotrab, line 1729: Same here, there is an assert above so this if should not be needed

front/lib/resources/group_resource.ts
```diff
@@ -2337,6 +2305,24 @@ export class GroupResource extends BaseResource<GroupModel> {
       ];
     }
 
+    if (this.isRegularManual()) {
```
matteotrab, line 2303: Permissions will then be moved to the group_permissions table, but for the moment we define them here.

ykmsd (approved):
looks good to me but I'm newbie so let's get a review from @philipperolet 🙏

front/types/groups.ts
```diff
@@ -11,9 +11,12 @@ import { isRoleType } from "./user";
  * global group: Contains all users from the workspace. Has access to the global
  * Space which holds all existing datasource created before spaces.
  *
- * regular group: Contains specific users added by workspace admins. Has access
+ * regular_auto group: Contains specific users added by workspace admins. Has access
```
philipperolet: this copy needs changing
"regular" groups are groups for which users are selected one by one (vs provisionned)
- auto means the group was created implicitely by dust (e.g. agent editors, space members)
- manual means it was created by the user via the UI (btw maybe regular_user would be better?)

front-api/routes/poke/workspaces/[wId]/groups/index.ts
```diff
@@ -13,7 +13,13 @@ app.get("/", async (ctx): HandlerResult<PokeListGroups> => {
   const auth = ctx.get("auth");
 
   const groups = await GroupResource.listAllWorkspaceGroups(auth, {
-    groupKinds: ["global", "regular_auto", "space_editors", "provisioned"],
+    groupKinds: [
```
Comment (philipperolet) [inline front-api/routes/poke/workspaces/[wId]/groups/index.ts]: what about using GROUP_KINDS from types directly here?

front/lib/api/actions/servers/poke/tools/users.ts
```diff
@@ -42,7 +42,13 @@ export const userHandlers: UserHandlers = {
     const targetAuth = targetAuthResult.value;
 
     const groups = await GroupResource.listAllWorkspaceGroups(targetAuth, {
-      groupKinds: ["global", "regular_auto", "space_editors", "provisioned"],
+      groupKinds: [
```
philipperolet: what about using GROUP_KINDS from types directly here?

front/lib/resources/group_resource.ts
```diff
@@ -637,6 +638,7 @@ export class GroupResource extends BaseResource<GroupModel> {
     groupKinds = [
       "global",
       "regular_auto",
+      "regular_manual",
```
philipperolet: what about using GROUP_KINDS from types directly here?

front-api/routes/poke/workspaces/[wId]/groups/index.ts
Comment (philipperolet) [inline front-api/routes/poke/workspaces/[wId]/groups/index.ts:1]: what about the non-poke api? GET /api/w/:wId/groups I think should return regular manual too no? so that useGroups({owner}) works?

philipperolet (approved):
Thanks @matteotrab looks good, left a few coms, pre-approving, feel free to merge once addressed (you can decide which to adress or reject ofc) 👍

front/types/groups.ts
matteotrab: Updated the copy. I prefer regular_manual to regular_user because it's explicit about the fact that the group was created manually and it's opposed to the regular_auto group

front-api/routes/poke/workspaces/[wId]/groups/index.ts
Comment (matteotrab) [reply front-api/routes/poke/workspaces/[wId]/groups/index.ts:1]: We can get regular manual groups by using the kind parameter of the `useGroups` hook. I'm not sure if we actually want `useGroups({owner})` to return regular_manual groups too. Today the only place where we use the hook without passing an argument for `kind` is in the API Keys page. I'll need to check better how API keys work to make sure we are displaying the correct group kinds there

PR #28836, opened 2026-07-10
handle unsupported file type from tool result and upload to gcs

ykmsd:
## Description
This PR is to avoid inlining base64 as a tool output. While investigating the memory usage of front agent loop worker, we noticed that when a tool returns unsupported file type, we embed full base64 blob as a tool output, and re-renders for the model every step. 

Example from one snapshot: google drive get file content tool returned 76MB xlsb file, and we inline it in every step, and that ended up taking 57% of total heap: 
```
153.66MB  {"texts":["user system <dust_system>...   (a request payload)
153.61MB  function google_drive__get_file_content [{"text":"{...XXX.xlsb...
153.61MB  [{"text":"{...fileId... XXX.xlsb...   (a tool result)
 76.80MB  UEsDBB...                          (a base64 blob)
```
This is because we only handle supported file types. This PR now removes the supported file type check, any blob resource goes through handleBase64Upload and upload to GCS instead of inlining it. 

You can see all the mimeType we are currently inlining from [here](https://app.datadoghq.eu/logs?query=%22MCP%20tool%20returned%20an%20embedded%20resource%20with%20an%20unsupported%20or%20missing%20mimeType%3B%20storing%20it%20inline%20in%20the%20conversation%22%20%40blobBytes%3A%3E0&agg_m=count&agg_m_source=base&agg_q=%40mimeType&agg_t=count&clustering_pattern_field_path=message&cols=host%2Cservice%2C%40apiError.api_error.type%2C%40apiError.api_error.message&flat_group_bys=true&fromUser=true&messageDisplay=inline&refresh_mode=sliding&sort_m=count&sort_t=count&storage=hot&stream_sort=time%2Cdesc&top_n=100&top_o=top&viz=toplist&x_missing=true&from_ts=1747296641466&to_ts=1747901441466&live=true).

## Tests

Will test it on front edge if I can fetch unsupported file somehow 
## Deploy Plan
Type change only, we don't need to do anything


front/lib/actions/mcp_execution.ts
```diff
@@ -302,32 +304,39 @@ export async function processToolResults(
...
-          } else if (
-            block.resource.mimeType &&
-            // File generated by the tool, not upserted yet.
-            isSupportedFileContentType(block.resource.mimeType)
-          ) {
-            if (isBlobResource(block)) {
+          }
+
+          // File generated by the tool, not upserted yet.
+          if (isBlobResource(block)) {
```
ykmsd, line 325: I'm not sure when ` block.resource.mimeType` will be empty, do I still need to add this check here? (but if so it will go to inline path?)

front/lib/actions/mcp_utils.ts
```diff
@@ -51,16 +48,13 @@ import { basename, extname } from "path";
...
   | { type: "image"; contentType: SupportedImageContentType }
-  | { type: "file"; contentType: SupportedFileContentType };
+  | { type: "file"; contentType: string };
 
-function getResourceInfo(mimeType: string): ResourceInfo | null {
+function getResourceInfo(mimeType: string): ResourceInfo {
   if (isSupportedImageContentType(mimeType)) {
     return { type: "image", contentType: mimeType };
   }
-  if (isSupportedFileContentType(mimeType)) {
```
ykmsd, line 61: or should we do differently for non supported file? 🤔

front/lib/actions/mcp_internal_actions/output_schemas.ts
```diff
@@ -93,12 +93,7 @@ const ToolGeneratedFilePathSchema = z.object({
   mimeType: z.literal(INTERNAL_MIME_TYPES.TOOL_OUTPUT.FILE_PATH),
   path: z.string(),
   title: z.string(),
-  contentType: z.enum(
-    Object.keys(ALL_FILE_FORMATS) as [
-      AllSupportedFileContentType,
-      ...AllSupportedFileContentType[],
-    ]
-  ),
+  contentType: z.string(),
```
ykmsd, line 96: I have zero idea if it's safe, from what I understood with my discussion with claude code it's not dangerous but I'm too newbie to tell

id13: Looks safe to me, it just means we are going to upload a lot of files instead of inlining them, which is the goal of this PR anyway

main codepath : https://github.com/dust-tt/dust/blob/main/front/lib/actions/mcp_execution.ts#L282-L282

id13 (approved):
LGTM

front/lib/actions/mcp_execution.ts
```diff
@@ -302,32 +304,39 @@ export async function processToolResults(
...
+              fileName,
+              toolContext,
+            });
+          }
+
+          // Non-blob resource generated by the tool (referenced by URI, not inline
+          // base64), not upserted yet.
+          if (
+            block.resource.mimeType &&
+            isSupportedFileContentType(block.resource.mimeType)
```
id13, line 352: Is this going to work ?

ykmsd: do you mean if we will ever come to this code path after my changes or generally if it's working or not? 🤔

so I'm not familiar with this area at all, but this logic was already [here](https://github.com/dust-tt/dust/blob/3ac6eb18648d21bf2983da6e5fa0683cca3f3262/front/lib/actions/mcp_execution.ts#L343-L393) and I moved out `if (isBlobResource(block))` so the order changed:
```
  } else if (
            block.resource.mimeType &&
            // File generated by the tool, not upserted yet.
            isSupportedFileContentType(block.resource.mimeType)
          ) {
            if (isBlobResource(block)) {
          ...
            }

            const fileName = isResourceWithName(block.resource)
              ? block.resource.name
              : (block.resource.uri.split("/").pop() ?? "generated-file");

            // Files land on the conversation in an agent loop, on the pod's shared project
            // context in a sandbox function invocation.
```

and we will come here when all of them are true, and apparently for the internal tools it's only for browse tool taking a page screenshot?? (but I'm not sure if this is true)
```
1. It's not a file-path-only resource (isToolGeneratedFilePath = false)
2. It's not an already-upserted tool file (isToolGeneratedFile = false)
3. It's not an inline base64 blob (isBlobResource = false) — i.e. no blob field
4. It has a mimeType that is a supported file content type
```

front/lib/actions/mcp_execution.ts
```diff
@@ -317,32 +319,39 @@ export async function processToolResults(
...
+          if (isBlobResource(block)) {
+            const { mimeType } = block.resource;
+
+            let fileName: string;
+            if (mimeType && isSupportedFileContentType(mimeType)) {
               const extensionFromContentType =
-                extensionsForContentType(
-                  block.resource.mimeType as SupportedFileContentType
-                )[0] || "";
+                extensionsForContentType(mimeType)[0] || "";
```
flvndvd, line 331: If file's mime type is not supported, I guess the name won't contain any extension. Do we know if in the occurrences we've seen, the name returned by MCP was maybe already containing the extension?

front/lib/actions/mcp_execution.ts
```diff
@@ -391,57 +400,56 @@ export async function processToolResults(
...
-                toolName: toolConfiguration.name,
-                serverName: toolConfiguration.mcpServerName,
-                blobBytes: isBlobResource(block)
-                  ? Buffer.byteLength(block.resource.blob, "utf8")
-                  : 0,
-              },
-              "MCP tool returned an embedded resource with an unsupported or missing mimeType; storing it inline in the conversation."
-            );
+          const text =
+            "text" in block.resource && typeof block.resource.text === "string"
```
flvndvd, line 406: `nit` typeguard?

front/lib/actions/mcp_utils.ts
flvndvd: Seems good like this. IIRC, it's only used to surface what happened to the tool output to the model.

flvndvd (approved):
LGTM modulo the widen content type. Let's move forward with this and we will see if we need to narrow it down.

front/lib/actions/mcp_execution.ts
id13: @flvndvd taking over @ykmsd to get this one done


So, logs adjacents to the OOM all follow the same pattern: 

mimetype in `application/vnd.dust.tool-output.tool-marker, .websearch-result, .agent-pause-tool-output`
So there is no extensions in these cases.

see [here](https://app.datadoghq.eu/logs?query=service%3Afront-agent-loop-worker%20pod_name%3Afront-agent-loop-worker-deployment-7cc766cdf7-tzw59%20%22mimeType%22&agg_m=count&agg_m_source=base&agg_t=count&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&from_ts=1784282400000&to_ts=1784289000000&live=false) and [here](https://app.datadoghq.eu/logs?query=service%3Afront-agent-loop-worker%20%28%22mimeType%22%20OR%20%22mime_type%22%20OR%20%22not%20supported%22%20OR%20%22unsupported%20mime%22%29&agg_m=count&agg_m_source=base&agg_t=count&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&from_ts=1783928924728&to_ts=1784533724728&live=false)

flvndvd: IMHO since we own all of those, they should be uploaded directly from their respective MCP.

id13: Ok, will own double hardening both on tool and in this part of the agent-loop

ykmsd: I checked several results of google drive and microsoft and seems like most of the time we can get the file name containing extensions from resource.uri so I guess it's okay? @id13 I'll do the quick test and merge this one but do you want to add something before merging?

id13: Yes I thinks it's ok too, see this : https://dust4ai.slack.com/archives/C09GELMTTRT/p1784535924720819

PR #28839, opened 2026-07-10
feat: endpoint to get regular manual group details

matteotrab:
## Description

This PR adds a new API endpoint to retrieve details of a regular manual group, including its members.

## Tests

Added tests covering:
- Correct response shape (group `sId`, `kind`, and member `sId`s).
- Access control: admin and business admin get `200`, regular user gets `403`.
- `404` returned for non-`regular_manual` groups.

## Risks

Low.

## Deploy Plan

Standard deployment.

PR #28837, opened 2026-07-10
feat: add endpoint to create regular manual groups

matteotrab:
## Description

This PR adds a `POST /api/w/:wId/groups` endpoint to allow admins and business admins to create `regular_manual` groups.

## Tests
Added tests covering: successful creation by admin and business admin, seeding members, invalid member IDs, duplicate names, missing name, and role-based access rejections (user and builder).

## Risks

Low.

## Deploy Plan

Standard deployment.

PR #28833, opened 2026-07-10
feat: add regular manual groups to cap eligible group kinds

matteotrab:
## Description

This PR adds `regular_manual` groups to the `CAP_ELIGIBLE_GROUP_KINDS` constant, enabling them to carry per-group usage spend limits and appear in the Usage > Groups admin table alongside `provisioned` groups.

## Tests

Manually.

## Risks

Low.

## Deploy Plan

Standard deployment.

PR #28840, opened 2026-07-10
feat: endpoint to update regular manual group

matteotrab:
## Description

This PR adds a `PATCH /api/w/:wId/groups/:groupId` endpoint to update a regular manual group's name and/or member list in a single request.

## Tests

Added tests covering: rename, set members, clear members, combined rename + set members, rename-only leaving members unchanged, business admin access, 403 for regular user, 404 for non-`regular_manual` group, 409 for duplicate name, 404 for out-of-workspace member, and 400 for empty name.

## Risks

Low.

## Deploy Plan

Standard deployment.

### DIFF
--- a/front/lib/resources/group_resource.ts
+++ b/front/lib/resources/group_resource.ts
@@ -2211,6 +2211,7 @@ export class GroupResource extends BaseResource<GroupModel> {
         | "user_not_found"
         | "user_not_member"
         | "user_already_member"
+        | "group_not_found"
         | "group_requirements_not_met"
         | "system_or_global_group"
       >
@@ -2223,6 +2224,10 @@ export class GroupResource extends BaseResource<GroupModel> {
           `Only workspace admins and ${BUSINESS_ADMIN_ROLE_NAME}s can update groups.`
         )
       );
+    }
+
+    if (!this.isRegularManual()) {
+      return new Err(new DustError("group_not_found", "Group not found."));
     }
 
     if (name !== undefined) {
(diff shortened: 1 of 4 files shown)


matteotrab:

- #28832
    - #28833
        - #28837
            - #28839
                - #28840 ◀
                    - #28843

front/lib/resources/group_resource.ts
```diff
@@ -2206,6 +2206,74 @@ export class GroupResource extends BaseResource<GroupModel> {
...
+      const uniqueMemberIds = [...new Set(memberIds)];
+      const users = await UserResource.fetchByIds(uniqueMemberIds);
+      if (users.length !== uniqueMemberIds.length) {
+        return new Err(
+          new DustError("user_not_found", "Some users were not found.")
+        );
+      }
+
+      const setResult = await this.dangerouslySetMembers(auth, {
+        users: users.map((u) => u.toJSON()),
```
ykmsd, line 2260: same question here but I know it's not your code

front/types/api/groups/manage.ts
```diff
@@ -17,3 +17,15 @@ export type GetGroupResponseBody = {
   group: GroupType;
   members: UserType[];
 };
+
+export const PatchGroupBodySchema = z.object({
+  name: z.string().min(1).optional(),
+  memberIds: z.array(z.string()).optional(),
+});
```
ykmsd, line 24: does it mean you can pass empty body? shouldn't we return early or make at least to provide one of them?

ykmsd (approved):

front/lib/resources/group_resource.ts
matteotrab: Yes we can improve it but out of scope for this PR

front/types/api/groups/manage.ts
matteotrab: Changed to return early in that case

PR #28844, opened 2026-07-10
Cache signed GCS URLs for conversation-rendered files

flvndvd:
## Description

Every time a conversation is rendered for the model, any image or file it references gets a brand new signed GCS URL, even for history that hasn't changed at all, since generating one embeds a fresh timestamp-based signature and expiry on every call. That means any interaction containing an image can never be part of a stable cached prefix, since the exact bytes at that position diverge on every single turn. Added a cache-aside layer so repeated renders within the signed URL's own validity window reuse the same URL instead of hitting GCS again, applied at both the legacy and canonical file-path code paths that conversation rendering uses, since they converge on the same underlying signer.


aubin-tchoi (approved):
LGTM

PR #28846, opened 2026-07-10
feat(model-picker): remove Dust/auto section from input bar model picker

Nils-Fedrigo:
## Description

Removes the "Dust" / auto provider group from the input bar model picker. The auto model was being bucketed into `moreByProvider` (rendered as a "Dust" section under "More models") and also surfaced as a hit when searching. Since auto selection is already exposed through the dedicated "Auto" toggle at the top of the dropdown, this section was redundant.

The auto model is now filtered out of `allModelsWithEfforts`, so it no longer appears as a browsable model line — neither as its own "Dust" provider group nor as a search result. The "Auto" toggle is unaffected and remains the way to enable auto model selection.

## Tests

Manually verify in the input bar model picker:
- The "Dust" section no longer appears under "More models".
- Searching for "auto" or "dust" no longer returns the auto model as a line item.
- The "Auto" toggle still enables/disables auto model selection as before.

### DIFF
--- a/front/components/assistant/conversation/input_bar/InputBarModelPicker.tsx
+++ b/front/components/assistant/conversation/input_bar/InputBarModelPicker.tsx
@@ -128,8 +128,6 @@ export function InputBarModelPicker({
   const allModelsWithEfforts = useMemo<ModelWithReasoningEffort[]>(
     () =>
       models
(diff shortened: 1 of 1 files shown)


Nils-Fedrigo:

- #28846 ◀

Fraggle (approved):
LGTM

PR #28848, opened 2026-07-10
feat(model-picker): hide suggested section when agent has a default

Nils-Fedrigo:
## Description

Hides the "Suggested" section in the model picker when the current agent has a configured default model. Previously both the "Default from agent" and "Suggested" sections rendered side by side (with the default merely filtered out of the suggestions list), which was redundant — the agent's default already surfaces the recommended pick. When a default is present, the Suggested section is now omitted entirely; when there is no default, suggestions render as before.

## Tests

Manually verified in the model picker dropdown:
- Agent with a configured default model → only the "Default from agent" section shows, no "Suggested" section.
- Agent without a default → "Suggested" section renders as before.


Nils-Fedrigo:

- #28848 ◀

PR #28851, opened 2026-07-10
[Skills] Skill restrictions canRead --> isMember

xebeohp:
## Description
This PR restricts skill space visibility to space members: skills that request a restricted space now use space.isMember(auth) instead of the broader canRead permission check, so they're only visible to members of that space.

## Tests
Tested locally.
Added resource-level tests covering member vs non-member visibility for open Pods, restricted spaces, and open regular spaces.

## Risk
Low risk.

## Deploy Plan
Normal deployment.

### DIFF
--- a/front/lib/resources/skill/skill_resource.test.ts
+++ b/front/lib/resources/skill/skill_resource.test.ts
@@ -20,9 +20,11 @@ import { GroupSpaceFactory } from "@app/tests/utils/GroupSpaceFactory";
 import { createResourceTest } from "@app/tests/utils/generic_resource_tests";
 import { KeyFactory } from "@app/tests/utils/KeyFactory";
 import { MCPServerViewFactory } from "@app/tests/utils/MCPServerViewFactory";
+import { MembershipFactory } from "@app/tests/utils/MembershipFactory";
 import { RemoteMCPServerFactory } from "@app/tests/utils/RemoteMCPServerFactory";
 import { SkillFactory } from "@app/tests/utils/SkillFactory";
 import { SpaceFactory } from "@app/tests/utils/SpaceFactory";
+import { UserFactory } from "@app/tests/utils/UserFactory";
 import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
 
 describe("SkillResource", () => {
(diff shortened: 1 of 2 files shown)


front/components/assistant/CapabilitiesPicker.tsx
```diff
@@ -257,7 +257,6 @@ export function CapabilitiesPicker({
   const { skills, isSkillsLoading } = useSkills({
     owner,
     status: "active",
-    globalSpaceOnly: true,
```
aubin-tchoi, line 274: are we sure the requested space IDs of the convo are updated properly when using a skill from a restricted space?

front/lib/api/skills/space_requirements.ts
```diff
@@ -18,27 +18,27 @@ export async function resolveAdditionalRequestedSpaceModelIds(
   }
 
   const spaces = await SpaceResource.fetchByIds(auth, requestedSpaceIds);
-  const readableSpacesById = new Map(
+  const memberSpacesById = new Map(
     spaces
-      .filter((space) => space.canRead(auth))
```
aubin-tchoi, line 23: are we sure we don't want to stick to the access-base rule?

front/components/assistant/CapabilitiesPicker.tsx
aubin-tchoi, line 274: maybe we should also surface a bit better in the UI which ones are on private spaces? so that ppl are aware of it when using them

front/lib/api/skills/space_requirements.ts
xebeohp: Yeah that makes more sense, fixed.

front/components/assistant/CapabilitiesPicker.tsx
xebeohp: Yes, it does that. Just to confirm you mean when selecting a restricted space as one of the accessible spaces for the skill.

xebeohp: Do you mean UI of the space or skill? I feel like the space UI is intuitive (black = you can access, red = you can't access) and the skill is just unsearchable if its attached to a space you don't have access to.
Screenshot 2026-07-13 at 10 57 24

aubin-tchoi: the scenario I had in mind was: if I use a skill that requests spaces A and B in a convo that only requests C, are the requested spaces of the conversation updated to A, B and C?

xebeohp: No, the requested space ID is not updated.

aubin-tchoi: the UI of the convo i meant, but since it's a much broader topic we can leave it out for this PR

aubin-tchoi: we need to update it, otherwise it's a potential data leak

xebeohp: Ok, so after using the skill that requests A & B in space C, if we add space C to the requestedSpaceIds, is that the equivalent of adding space C as a space that the skill requests?
Then only people with access to spaces A, B, & C can use the skill moving forward, addressing the data leak issue?

If I'm interpreting it correctly, should this also automatically be reflected in the skills setting UI?

aubin-tchoi: wait no, what I'm referring to is the conversation's requested spaces, not the skill's

aubin-tchoi: the conversation's requested spaces are the spaces the conversation pulled data/resources from
when someone tries to open a conversation we check that they have access to all those spaces

aubin-tchoi: in a converstion where a skill that requests spaces A and B is used, the spaces need to at least contain spaces A and B

aubin-tchoi: lmk if unclear

xebeohp:

- #28851 ◀
    - #29020

front/lib/resources/skill/skill_resource.test.ts
```diff
@@ -64,6 +66,77 @@ describe("SkillResource", () => {
...
+      await MembershipFactory.associate(testContext.workspace, outsider, {
+        role: "user",
+      });
+      return Authenticator.fromUserIdAndWorkspaceId(
+        outsider.sId,
+        testContext.workspace.sId
+      );
+    }
+
+    async function listSeesSkill(auth: Authenticator, skillSId: string) {
```
aubin-tchoi, line 81: nit: let's avoid nested functions unless needed (would inline it here)

aubin-tchoi, line 81:
```suggestion
    async function listSeesSkill(auth: Authenticator, skillId: string) {
```

front/lib/resources/skill/skill_resource.ts
```diff
@@ -581,7 +577,6 @@ export class SkillResource extends BaseResource<SkillConfigurationModel> {
       transaction,
     });
 
-    // Check if the user has access to skill requested spaces.
```
aubin-tchoi, line 584: nit: why remove this comment?

front/lib/resources/skill/skill_resource.test.ts
```diff
@@ -64,6 +66,77 @@ describe("SkillResource", () => {
...
+        requestedSpaceIds: [space.id],
+      });
+
+      expect(await listSeesSkill(testContext.authenticator, skill.sId)).toBe(
+        true
+      );
+      expect(await listSeesSkill(await nonMemberAuth(), skill.sId)).toBe(false);
+    });
+
+    it("keeps a skill restricted to an open regular space visible to all members", async () => {
```
aubin-tchoi, line 126: surprising that the rule is different from pods

aubin-tchoi: ah actually you can't really add yourself to an open space

aubin-tchoi (approved):

PR #28853, opened 2026-07-10
feat(assistant): pass requested model to placeholder user message on model picker override

Nils-Fedrigo:
## Description

When the `models_picker` feature flag is enabled, the bare "model agents" — global
agents that merely wrap a single provider model (`gpt-5`, `claude-sonnet`,
`claude-4.5-haiku`, `mistral-large`, `gemini-pro`, the `o*` series, etc.) — are now
hidden from the default agent list. The model picker exposes these underlying models
directly in the input bar, so the wrapper agents are redundant when the flag is on.

The filtering is done in `getGlobalAgents` alongside the existing default-list
filters, and only applies when building the default list (`agentIds === undefined`) —
the same path every picker/list view uses. Explicit fetches by `agentId` still resolve
these agents, so existing conversations that reference them keep working. This mirrors
the existing `RETIRED_GLOBAL_AGENTS_SID` handling.

## Tests

Verify with the `models_picker` flag on: the model-wrapper global agents (e.g.
`claude-sonnet`, `gpt-5`) no longer appear in the agent picker / agent list, while a
pre-existing conversation using one of them still loads and runs. With the flag off,
they appear as before.

### DIFF
--- a/front/components/assistant/conversation/ConversationViewer.tsx
+++ b/front/components/assistant/conversation/ConversationViewer.tsx
@@ -291,7 +291,7 @@ function buildFirstMessagePlaceholders(
   pending: PendingConversationMessage,
   user: UserType
 ): FirstMessagePlaceholders {
-  const { input, mentions, contentFragments } = pending;
+  const { input, mentions, contentFragments, modelSelection } = pending;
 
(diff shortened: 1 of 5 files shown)


Nils-Fedrigo:

- #28670
    - #28853 ◀

PR #28857, opened 2026-07-10
Bump the tool-output-content fetch batch size and add visibility into conversation length

flvndvd:
## Description

We recently confirmed a real cache-busting bug in production: as a conversation grows, old tool results periodically get dropped from what we fetch for the model, and the way that drop is currently batched means it happens in big, sudden jumps rather than gradually. When one of those jumps hits, the model sees content disappear that it saw in full just one turn earlier, which invalidates the whole prompt cache from that point forward and makes the next request far more expensive than it needs to be.

We looked at a more involved fix that tracks, per conversation, exactly how much history is still needed and adjusts the fetch window accordingly, but decided to hold off on landing that until we understand the problem better. Instead of guessing at the right batch size, this change raises the current threshold to a safer default and adds a metric that reports how many interactions a typical conversation actually has. Once we have a few days of real numbers, we can either tune this constant with confidence or revisit the more thorough fix if it turns out long conversations are common enough to matter.

This is a low-risk, mostly observational change. Nothing changes for the vast majority of conversations that stay under the new threshold, and we'll be watching the new metric before deciding what to do next.

### DIFF
--- a/front/lib/api/assistant/conversation/fetch.test.ts
+++ b/front/lib/api/assistant/conversation/fetch.test.ts
@@ -51,31 +51,43 @@ describe("computeMessagesWithToolOutputContent", () => {
   });
 
   it("excludes the oldest batch once the floor's start crosses a batch boundary, and stays put across many turns", () => {
-    // At n=14, floorStart=10, which is the first batch boundary: interactions 1-10 (ids) get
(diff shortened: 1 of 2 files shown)


aubin-tchoi (approved):
LGTM

PR #28860, opened 2026-07-10
feat(sandbox): space-scoped sandbox env vars on a unified table

smb2268:
## Description

Schema foundation for Pod-scoped sandbox secrets — **one unified table**, per the consolidation discussion below: pods and workspaces are two scopes of the same entity.

**What this PR does:**

1. `workspace_sandbox_env_vars` gains a nullable **`spaceId`** (FK `vaults(id)` RESTRICT): `NULL` = workspace scope (all existing rows, untouched — no backfill), set = pod scope (pods are project spaces). Uniqueness becomes per-scope via two partial unique indexes, so a pod var can shadow a workspace name — the pod-wins merge at sandbox boot depends on it.
2. User-FK indexes on `createdByUserId` / `lastUpdatedByUserId` (BACK13 — pre-existing gap on this table); pod-scoped rows die with their space (hard-delete path, reached transitively by workspace scrub); `KNOWN_SPACE_RELATED_MODELS` guard entry.

**Schema delta:**

| Change | Notes |
|---|---|
| `spaceId` bigint NULL, FK → `vaults(id)` RESTRICT | scope discriminator; rows die with their space |
| `sandbox_env_vars_workspace_scope_name_idx` | UNIQUE `("workspaceId", name) WHERE "spaceId" IS NULL` |
| `sandbox_env_vars_pod_scope_name_idx` | UNIQUE `("spaceId", name) WHERE "spaceId" IS NOT NULL`; also serves FK lookups |
| `createdByUserId` / `lastUpdatedByUserId` indexes | BACK13 |

Nothing reads or writes the new column yet — no behavior change.

> [!NOTE]
> The legacy full unique `workspace_sandbox_env_vars_workspace_name_idx` is **intentionally kept**: already-deployed queries filter `(workspaceId, name)` without a `spaceId` predicate and can't use the partial indexes. PR 2 ships scope-aware queries and drops the legacy index **post-deploy**. Until that drop, a pod row cannot shadow a workspace name in prod — fine, pod writes are gated behind PR 2.

> [!NOTE]
> An earlier revision of this stack also carried external-secret-provider columns (`secret_source_kind`, nullable `encrypted_value` + CHECK). That has been removed on purpose: we're meeting with 1Password to solidify the provider structure first, and the provider schema will land in a dedicated, purely additive PR once the shape is grounded in real requirements.

**Stack:**
- PR 1 (this PR): schema. **Merges and applies first.**
- PR 2 (`28861`): scope-aware `SandboxEnvVarResource` consolidating both scopes + legacy index drop post-deploy.
- PR 3 (`28862`): pod sandbox boot-path wiring (egress-secrets merge + owner env injection).
- PR 4 (`29327`): pod routes + shared env vars UI (workspace and pod on one component).

## Risk

Low, safe to rollback

## Deploy Plan

- [ ] Merge; ensure the pre-deploy migration is applied in US + EU before front deploys

### DIFF
--- a/front/lib/resources/space_resource.test.ts
+++ b/front/lib/resources/space_resource.test.ts
@@ -11,11 +11,13 @@ import { frontSequelize } from "@app/lib/resources/storage";
 import { GroupMembershipModel } from "@app/lib/resources/storage/models/group_memberships";
 import { GroupSpaceModel } from "@app/lib/resources/storage/models/group_spaces";
 import { SpaceModel } from "@app/lib/resources/storage/models/spaces";
+import { WorkspaceSandboxEnvVarModel } from "@app/lib/resources/storage/models/workspace_sandbox_env_var";
 import type { UserResource } from "@app/lib/resources/user_resource";
 import { AgentConfigurationFactory } from "@app/tests/utils/AgentConfigurationFactory";
 import { ConversationFactory } from "@app/tests/utils/ConversationFactory";
 import { GroupFactory } from "@app/tests/utils/GroupFactory";
 import { MembershipFactory } from "@app/tests/utils/MembershipFactory";
+import { SandboxEnvVarFactory } from "@app/tests/utils/SandboxEnvVarFactory";
 import { SpaceFactory } from "@app/tests/utils/SpaceFactory";
 import { UserFactory } from "@app/tests/utils/UserFactory";
 import { WorkspaceFactory } from "@app/tests/utils/WorkspaceFactory";
@@ -140,6 +142,44 @@ describe("SpaceResource", () => {
           },
         })
       ).resolves.toBe(0);
+    });
+
+    it("should delete pod-scoped sandbox env vars but keep workspace-scoped ones when hard deleting a space", async () => {
+      const pod = await SpaceFactory.project(workspace, user1.id);
+
+      await SandboxEnvVarFactory.create(adminAuth, {
+        name: "POD_TOKEN",
+        space: pod,
+      });
+      await SandboxEnvVarFactory.create(adminAuth, {
+        name: "WORKSPACE_TOKEN",
+      });
+
+      const softDeleteResult = await pod.delete(adminAuth, {
+        hardDelete: false,
+      });
+      expect(softDeleteResult.isOk()).toBe(true);
+
+      const deletedSpace = await SpaceResource.fetchById(adminAuth, pod.sId, {
+        includeDeleted: true,
+      });
+      if (!deletedSpace) {
+        throw new Error("Deleted space should exist");
+      }
+
+      const hardDeleteResult = await hardDeleteSpace(adminAuth, deletedSpace);
+      expect(hardDeleteResult.isOk()).toBe(true);
+
+      await expect(
+        WorkspaceSandboxEnvVarModel.count({
+          where: { workspaceId: workspace.id, spaceId: pod.id },
+        })
+      ).resolves.toBe(0);
+      await expect(
+        WorkspaceSandboxEnvVarModel.count({
+          where: { workspaceId: workspace.id, name: "WORKSPACE_TOKEN" },
+        })
+      ).resolves.toBe(1);
     });
 
     describe("authorization checks", () => {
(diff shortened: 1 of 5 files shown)


front/lib/resources/storage/models/workspace_sandbox_env_var.ts
```diff
@@ -3,21 +3,45 @@ import {
...
   declare allowedDomains: string[] | null;
-  declare encryptedValue: string;
+  // Null only for external secret sources (none implemented yet): those
+  // store a pointer in secretSourceConfig instead of a value. dust-managed
+  // rows always carry the AES-256-CBC ciphertext — enforced by the
+  // sandbox_env_vars_dust_managed_value_check DB constraint.
+  // /!\ That constraint is hand-written in the migration and invisible to
+  // Sequelize, so pg-schema-diff will propose DROP CONSTRAINT for it in any
+  // migration you generate from this model — remove that statement, do not
+  // accept the drop.
```
flvndvd: We might want to remove this piece, right? It holds true only for this one PR/migration?

front/lib/resources/storage/models/workspace_sandbox_env_var.ts
```diff
@@ -3,21 +3,45 @@ import {
...
+  declare encryptedValue: string | null;
```
flvndvd: I might be missing the full picture and a few tokens, but I'm not sure to see how would remote secret be represented in table?

front/lib/resources/storage/models/workspace_sandbox_env_var.ts
```diff
@@ -57,23 +81,69 @@ WorkspaceSandboxEnvVarModel.init(
...
     },
     encryptedValue: {
+      type: DANGEROUSLY_UNBOUNDED_TEXT,
+      allowNull: true,
+    },
+    secretSourceKind: {
       type: DANGEROUSLY_UNBOUNDED_TEXT,
       allowNull: false,
+      defaultValue: "dust-managed",
+      field: "secret_source_kind",
```
flvndvd: Any reason to override just here?

front/lib/resources/storage/models/workspace_sandbox_env_var.ts
```diff
@@ -57,23 +81,69 @@ WorkspaceSandboxEnvVarModel.init(
...
+    },
+    secretSourceConfig: {
+      type: DataTypes.JSONB,
```
flvndvd: We are not big fan of `JSONB` field, mind sharing a few examples of what we intend to store there?

front/lib/resources/storage/models/workspace_sandbox_env_var.ts
```diff
@@ -57,23 +81,69 @@ WorkspaceSandboxEnvVarModel.init(
...
+      },
+      // User FKs are SET NULL on user deletion — without these indexes,
+      // scrubbing a user would scan the table (BACK13).
+      { fields: ["createdByUserId"], concurrently: true },
+      { fields: ["lastUpdatedByUserId"], concurrently: true },
     ],
   }
 );
 
+WorkspaceSandboxEnvVarModel.belongsTo(SpaceModel, {
```
flvndvd, line 108: We are missing the other direction relationship.

flvndvd (commented):
Left a few questions around the data model

smb2268:
@flvndvd simplified PR ready for eyes

front/lib/resources/space_resource.test.ts
```diff
@@ -142,6 +143,55 @@ describe("SpaceResource", () => {
       ).resolves.toBe(0);
     });
 
+    it("should delete pod-scoped sandbox env vars but keep workspace-scoped ones when hard deleting a space", async () => {
+      const pod = await SpaceFactory.project(workspace, user1.id);
+
+      // Raw model writes on purpose: nothing can create pod-scoped rows
+      // through resources yet (gated behind the resources PR).
+      await WorkspaceSandboxEnvVarModel.create({
```
flvndvd: Nit can we use the resource or a factory here? But ideally no operations directly on the model in the tests.

flvndvd (approved):
10x simpler 😅

PR #28861, opened 2026-07-10
feat(sandbox): scope-aware SandboxEnvVarResource on the unified table

smb2268:
## Description

Follow up of #28860 (unified schema). One scope-aware resource for sandbox env vars — per review, the parallel pod resource is gone along with the pod table.

> [!NOTE]
> The mechanical renames (model class/file, kind type, name helpers, wire types) are extracted to `29433` per review feedback on size — once it merges, this PR rebases and sheds those hunks, leaving only the semantic changes below.

**What this PR does:**

1. **`SandboxEnvVarResource`** replaces `WorkspaceSandboxEnvVarResource`: every method takes a typed `{ kind: "workspace", workspace } | { kind: "pod", pod }` scope, which drives row filtering (`spaceId` NULL vs set), the **encryption key** (workspace sId vs pod space sId — derived via `scopeEncryptionKey` from the same object that scoped the query, so key and row scope cannot diverge), per-scope caps, and the ownership asserts (`belongsToScope` on mutations, sandbox-owner assert on pod boot loads).
2. **Staged index swap**: scope-aware queries ship with this deploy; the **post-deploy** migration drops the legacy `(workspaceId, name)` full unique that #28860's partial per-scope uniques supersede. After the drop, a pod var may shadow a workspace name (pod wins at boot — next PR).
3. **Workspace routes gain a `belongsToScope` guard**: the shared table holds pod-scoped rows this surface must not read or mutate (they are encrypted under a different key) → 404, covered by a route test (PATCH + DELETE against a pod row 404 and leave it untouched).
4. Hardening from earlier review rounds rides along: `createOnly` on `upsert` so `makeNew` rejects an existing row **before any write** (previously overwrote then reported failure 😱, no-clobber regression tests on both scopes), user-name joins.
5. Pod-scoped mutations emit the same `sandbox_env_var.*` audit events through the shared resource code paths; the pod-identifying `space_id` metadata and schema updates land in the dedicated audit PR (`29346`).
6. **Naming sweep** — `Workspace` stays only on genuinely workspace-scoped things. Renamed (all internal, no wire/DDL change): `SandboxEnvVarModel` (class + file; `modelName`/table unchanged), `SandboxEnvVarKind` / `SANDBOX_ENV_VAR_KINDS`, `render/parseSandboxEnvVarName*` helpers. Kept: `deleteAllForWorkspace`, `MAX_VARS_PER_WORKSPACE`, the workspace routes' local schemas, and the table name (a DB rename is real migration work for zero behavior).

External secret providers are deliberately **not** in this PR (see #28860 — provider schema lands after the 1Password discussions); values decrypt inline under the scope key, same pattern as before, one key per scope.

No pod HTTP surface in this PR — the routes (`29348`), shared UI (`29349`), and audit metadata (`29346`) follow in the stack.

## Risk

Worst case, the `createOnly` change rejects a duplicate create that previously (incorrectly) overwrote — the only production caller uses `upsert` without it, verified. Pod scope is dead code until the boot PR. Safe to rollback before the post-deploy index drop; after it, rollback needs the legacy index re-created first (partial uniques keep correctness either way, only pre-cutover query plans regress).

## Deploy Plan

- [ ] Apply the pre-deploy migration (creates `sandbox_env_vars_workspace_id_idx`, both regions)
- [ ] Deploy front (scope-aware queries live)
- [ ] Apply the post-deploy migration (drops the legacy unique, both regions)


smb2268:

- #28861 ◀

front-api/routes/w/[wId]/sandbox/env-vars/[id].ts
```diff
@@ -40,8 +40,14 @@ app.patch(
       });
     }
 
-    const envVar = await WorkspaceSandboxEnvVarResource.fetchById(auth, id);
-    if (!envVar) {
+    const scope = {
+      kind: "workspace" as const,
+      workspace: auth.getNonNullableWorkspace(),
+    };
+    const envVar = await SandboxEnvVarResource.fetchById(auth, id);
```
Comment (fontanierh) [inline front-api/routes/w/[wId]/sandbox/env-vars/[id].ts]: OOC why don't we enforce scope in the resource directly ?

front-api/routes/w/[wId]/sandbox/env-vars/[id].ts
```diff
@@ -122,8 +128,16 @@ app.delete(
     const auth = ctx.get("auth");
     const { id } = ctx.req.valid("param");
 
-    const envVar = await WorkspaceSandboxEnvVarResource.fetchById(auth, id);
-    if (!envVar) {
+    const envVar = await SandboxEnvVarResource.fetchById(auth, id);
+    // The table also holds pod-scoped rows — this workspace route must not
```
Comment (fontanierh) [inline front-api/routes/w/[wId]/sandbox/env-vars/[id].ts]: to make sure I fully understand, we use this endpoint for workspace-level and sandbox-level env vars, then a different endpoint for pod-level?

front/migrations/post-deploy/20260712120000_drop_legacy_sandbox_env_vars_unique.sql
```diff
@@ -0,0 +1,15 @@
...
+(sandbox_env_vars_workspace_scope_name_idx / sandbox_env_vars_pod_scope_name_idx).
+Post-deploy on purpose: the deploy this migration belongs to ships
+scope-aware queries whose predicates match the partial indexes; the legacy
+index must survive until that code is live so older pods' (workspaceId, name)
+lookups stay indexed. Once dropped, a pod-scoped row may shadow a
+workspace-scoped name — which the pod-wins merge requires.
+*/
+SET SESSION statement_timeout = 1200000;
+SET SESSION lock_timeout = 3000;
+DROP INDEX CONCURRENTLY "public"."workspace_sandbox_env_vars_workspace_name_idx";
```
fontanierh, line 19: I think this removes all indexes that lead with the workspace, so eg delete all for workspace will become slow no ?

front-api/routes/w/[wId]/sandbox/env-vars/[id].ts
Comment (smb2268) [reply front-api/routes/w/[wId]/sandbox/env-vars/[id].ts]: There's actually no sandbox-level vars like we have for egress domains, just workspace and pod (space) level. We're storing the vars for workspace & pod scope in the same table, so here we're refactoring the existing WorkspaceSandboxEnvVarResource to strip `Workspace` from the name (started that process in my last PR) and allow it to represent both Workspace & Space scoped env vars.

Routes that handle the space scope are in a PR later in this stack!

front-api/routes/w/[wId]/sandbox/env-vars/[id].ts
Comment (smb2268) [reply front-api/routes/w/[wId]/sandbox/env-vars/[id].ts]: @fontanierh I was matching the pattern used elsewhere here per [GEN1] rule. `fetchById(auth, id)` + check-in-handler is how space-routed resources do it elsewhere (e.g. the apps routes: AppResource.fetchById(auth, aId) then found.space.sId !== space.sId → 404)

I do think it'd be a bit more readable to do upfront so happy to make the change if you prefer, it'd just be a little diff than we have elsewhere

front/migrations/post-deploy/20260712120000_drop_legacy_sandbox_env_vars_unique.sql
smb2268: @fontanierh yep good catch, fixing

front-api/routes/w/[wId]/sandbox/env-vars/[id].ts
Comment (fontanierh) [reply front-api/routes/w/[wId]/sandbox/env-vars/[id].ts]: It's just that this "scope" thing feels like a concept coming from the resource, there are a bunch of places in there where we list by scope etc.. So it would make kind of sense to me if the resources exposed a "fetchByScope" or stg like that

Comment (fontanierh) [reply front-api/routes/w/[wId]/sandbox/env-vars/[id].ts]: wdyt ?

front-api/routes/w/[wId]/sandbox/env-vars/[id].ts
Comment (fontanierh) [reply front-api/routes/w/[wId]/sandbox/env-vars/[id].ts]: Indeed got confused!

fontanierh (approved):

front-api/routes/w/[wId]/sandbox/env-vars/[id].ts
Comment (smb2268) [reply front-api/routes/w/[wId]/sandbox/env-vars/[id].ts]: yep good with me, updated!

flvndvd (approved):
LGTM on the data model front.

PR #28862, opened 2026-07-10
feat(sandbox): inject pod env vars and secrets at pod sandbox boot

smb2268:
## Description

Follow up of #28861 (unified scope-aware resource). Wires Pod-scoped sandbox secrets into the pod sandbox boot path — the only PR of the stack that changes what a running sandbox does.

**What this PR does:**

1. **Pod config applies to every Computer running in the Pod** — mirroring the egress-policy decision: both pod-owned sandboxes *and* conversation-owned sandboxes whose conversation lives in the pod receive the pod's env vars and secrets. The conversation runtime owner now carries its `spaceId` (same linkage the egress policy file already keys on).
2. **`egress-secrets.json` is built per runtime owner**: workspace entries for every sandbox; for sandboxes running in a pod, pod entries are merged in and **pod wins on name collision** (placeholder and allowed domains included, not just the value). Pod entries decrypt inline under the pod scope key against the unified table (`spaceId`-scoped rows) — any failure (missing nonce, missing domains, decrypt error) aborts the whole write, never a partial or empty-valued entry. A missing pod space fails the build for pod-owned sandboxes; for conversations it skips the pod layer (same rule as the env injection, keeping file and env consistent).
3. **`writeEgressSecretsFile` takes the `SandboxRuntimeOwner`** already in scope at `setupEgressForwarder` (the same owner that feeds the `ownerId` JWT claim).
4. **Both adapters inject pod env vars** through the owner env layer: config vars in cleartext (`DST_*`), HTTPS secrets as their DSEC placeholders (dsbx swaps them on the wire). The owner layer beats the workspace layer in `buildSandboxEnvVars`, so env precedence matches the egress-file merge — pod wins in both places. Fail closed: a var that cannot be resolved aborts activation rather than booting without it. The loads ride a lazy `envVars` factory so `ensureActive` on an already-running sandbox pays nothing.
5. **Bash-output redaction covers pod scope**: conversations in a pod now run with pod var values in their sandbox, so `redactSandboxEnvVarsFromOutput` scrubs the pod's config values from tool output alongside the workspace ones.
6. **The env manifest lists pod vars too** (`sandbox-env-manifest.json` — names/placeholders/domains only, never values): sandboxes running in a pod see the merged var set, pod winning on collision, so function/skill authors can discover pod vars. The owner→pod resolution is now one shared rule (`resolvePodForRuntimeOwner`) used by the egress-secrets build and the manifest alike — pod-owned sandboxes hard-fail without their pod space; conversations without a (project) space just get workspace config.

Pod-beats-workspace precedence for plain env vars (owner layer over workspace layer in `buildSandboxEnvVars`) is locked by a test: a colliding `DST_*` var and `DSEC_*` placeholder both surface the pod value in `provider.create`.

> [!NOTE]
> Behavioral no-op in production until the pod secrets API routes (next PR) let admins create pod env vars — until then no rows exist, both builders return workspace-only output, and the owner env layer adds nothing.

## Risk

Worst case, a sandbox running in a pod fails activation on an unresolvable pod var (fail-closed by design). Conversations outside a pod are untouched: their path returns workspace entries unchanged and their env gains nothing. Safe to rollback.

## Tests

Pod/workspace merge with pod-wins collision (placeholder + domains + value) for both pod-owned and conversation-in-pod owners, manifest listing merged for both owner kinds, non-pod-space conversations get workspace entries only, conversation-owner passthrough regression, decrypt-under-pod-scope-key via the real egress consumer, env-layer precedence pinned at `provider.create` for both adapters (including no `SPACE_ID` leak into conversation sandboxes), boot-owner assert matrix (wrong pod / no pod / matching pod conversation), stdin-not-argv write regression.

### DIFF
--- a/front/lib/api/actions/servers/sandbox/tools/index.test.ts
+++ b/front/lib/api/actions/servers/sandbox/tools/index.test.ts
@@ -1,7 +1,9 @@
 import { Authenticator } from "@app/lib/auth";
+import { SpaceResource } from "@app/lib/resources/space_resource";
 import { WorkspaceResource } from "@app/lib/resources/workspace_resource";
 import { FeatureFlagFactory } from "@app/tests/utils/FeatureFlagFactory";
 import { createResourceTest } from "@app/tests/utils/generic_resource_tests";
+import type { ConversationWithoutContentType } from "@app/types/assistant/conversation";
 import { Err, Ok } from "@app/types/shared/result";
 import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
 
@@ -226,7 +228,15 @@ describe("runSandboxBashTool", () => {
     mockFetchActionById.mockResolvedValue(null);
   });
 
-  function makeExtra() {
+  function makeExtra(
+    overrides: {
+      // The fixture is an untyped fake (the runContext is cast as never), so
+      // this Picks only the fields the tools read — typed against the real
+      // conversation shape so renames surface at compile time.
+      conversation?: Pick<ConversationWithoutContentType, "sId"> &
+        Partial<Pick<ConversationWithoutContentType, "spaceId">>;
+    } = {}
+  ) {
     return {
       auth: {
         getNonNullableWorkspace: () => ({
@@ -241,7 +251,7 @@ describe("runSandboxBashTool", () => {
         },
         modelInfo: { endpoint: { modelConfig: { providerId: "openai" } } },
         agentMessage: { sId: "message-id", agentMessageId: 1 },
-        conversation: { sId: "conversation-id" },
+        conversation: overrides.conversation ?? { sId: "conversation-id" },
         action: {
           sId: "sandbox-action-id",
           toJSON: () => ({ sId: "sandbox-action-id" }),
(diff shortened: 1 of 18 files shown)


smb2268:

- #28861
    - #28862 ◀

front/lib/api/actions/servers/sandbox/tools/index.test.ts
```diff
@@ -226,7 +227,9 @@ describe("runSandboxBashTool", () => {
     mockFetchActionById.mockResolvedValue(null);
   });
 
-  function makeExtra() {
+  function makeExtra(
+    overrides: { conversation?: { sId: string; spaceId?: string } } = {}
```
flvndvd: Any reason to not take at least the `LightConversationType` or  `ConversationResource` here?

front/lib/api/sandbox/egress_secrets.ts
```diff
@@ -82,11 +87,128 @@ export async function buildEgressSecretFileEntries(
   return new Ok(entries);
 }
 
+export async function buildPodEgressSecretEntries(
```
flvndvd, line 78: Can we get your eyes on this @fontanierh?

flvndvd (approved):

front/lib/api/actions/servers/sandbox/tools/index.test.ts
smb2268: swapped to Pick<ConversationWithoutContentType, "sId"> & Partial<Pick<…, "spaceId">>,  which seems to be the pattern for similar tests, IMO I don't think mocking out the full structure adds any strength to the test but happy to do it if you disagree!

front/lib/api/sandbox/egress_secrets.ts
```diff
@@ -82,11 +87,128 @@ export async function buildEgressSecretFileEntries(
...
+  const scope = { kind: "pod" as const, pod };
+  const resources = await SandboxEnvVarResource.listHttpsSecretsForEgress(
+    auth,
+    scope,
+    runtimeOwner
+  );
+
+  const entries: EgressSecretFileEntry[] = [];
+  for (const resource of resources) {
+    if (!resource.placeholderNonce) {
```
fontanierh: Why is the type allowing for a nullable placeholder nonce ?

front/lib/api/sandbox/egress_secrets.ts
```diff
@@ -82,11 +87,128 @@ export async function buildEgressSecretFileEntries(
...
+      return new Err(
+        new Error(
+          `Pod HTTPS secret sandbox environment variable ${resource.envName} is missing its placeholder nonce.`
+        )
+      );
+    }
+    if (!resource.allowedDomains) {
```
fontanierh: How would this be possible? Why are we not enforcing this a type level ?

smb2268: It shouldn't be possible, but moved to type narrow to be more clear

PR #28863, opened 2026-07-10
Surface sandbox function authentication events

spolu:
## Description

Sandbox function tool actions now preserve personal-authentication blocks instead of being marked errored. Authentication-required events are published to the invocation stream, and the action polling endpoint keeps returning pending while authentication is required.

Includes an implementation note for moving outer sandbox function execution into Temporal so the visualization can subscribe before a blocking tool call. This follow-up is designed to compose with #28829, which resolves approvals and launches the inner tool workflow.

## Tests

- `NODE_ENV=test npm test temporal/agent_loop/activities/run_sandbox_function_tool.test.ts`
- `NODE_ENV=test npm test routes/v1/w/[wId]/sandbox/actions/[aId]/index.test.ts`

## Risk

Low. Sandbox functions are not live yet; changes are limited to sandbox-function tool authentication state and event delivery.

## Deploy Plan

- deploy `front`

### DIFF
--- a/front/lib/actions/mcp.ts
+++ b/front/lib/actions/mcp.ts
@@ -344,6 +344,12 @@ export function isSandboxFunctionToolEvent<
   return "sandboxFunctionId" in event;
 }
 
+export function isAgentLoopToolEvent<
+  T extends AgentLoopEventScope | SandboxFunctionEventScope,
+>(event: T): event is Extract<T, AgentLoopEventScope> {
+  return "conversationId" in event;
+}
+
 export function isToolFileAuthRequiredEvent(
   event: unknown
 ): event is ToolFileAuthRequiredEvent {
--- a/front/lib/actions/mcp.ts
+++ b/front/lib/actions/mcp.ts
@@ -8,11 +8,12 @@ import type {
   MCPServerAvailability,
 } from "@app/lib/actions/mcp_internal_actions/constants";
 import type {
+  AgentLoopEventScope,
   AgentLoopMCPApproveExecutionEvent,
   AgentLoopToolExecution,
   MCPApproveExecutionEvent,
+  SandboxFunctionEventScope,
   SandboxFunctionToolExecution,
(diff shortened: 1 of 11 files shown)


front/temporal/agent_loop/activities/run_sandbox_function_tool.ts
```diff
@@ -58,31 +60,47 @@ export async function runSandboxFunctionToolActivity(
...
-      )
-    );
+    for await (const event of runToolWithStreaming(
+      auth,
+      { toolContext: { runContext } },
+      { signal: abortSignal }
+    )) {
+      switch (event.type) {
+        case "tool_personal_auth_required":
+          assert(
```
PopDaph, line 77: nit: should we use a proper typeguard instead of `assert("invocationId" in event, ...)` here? Something like `isSandboxFunctionToolPersonalAuthRequiredEvent(event)`? (If someone later adds another field or renames invocationId, the string check silently rots)

PopDaph (approved):

front/temporal/agent_loop/activities/run_sandbox_function_tool.ts
spolu: Yup 👍

spolu:
@PopDaph what do you think of the plan in WORK.md? Are you aligned or am I missing something?

front/temporal/agent_loop/activities/run_sandbox_function_tool.ts
spolu: done 👍

front/lib/actions/mcp.ts
```diff
@@ -342,6 +338,18 @@ export function isToolPersonalAuthRequiredEvent(
   );
 }
 
+export function isSandboxFunctionToolEvent<
```
spolu, line 341: Look at me @flvndvd! I'm Typescript pilled!

PR #28864, opened 2026-07-10
[front] add groups UI

ykmsd:
## Description
This PR is to add groups UI. Not plugged with backend so pure frontend only. 

CleanShot 2026-07-10 at 23 09 59@2x

### DIFF
--- a/front/components/groups/CreateGroupDialog.tsx
+++ b/front/components/groups/CreateGroupDialog.tsx
@@ -1,0 +1,78 @@
+import { MemberSelectionTable } from "@app/components/members/MemberSelectionTable";
+import type { WorkspaceType } from "@app/types/user";
+import {
+  Dialog,
+  DialogContainer,
+  DialogContent,
+  DialogFooter,
+  DialogHeader,
+  DialogTitle,
+  Input,
+} from "@dust-tt/sparkle";
+import { useEffect, useState } from "react";
+
+interface CreateGroupDialogProps {
+  owner: WorkspaceType;
+  isOpen: boolean;
+  onOpenChange: (open: boolean) => void;
+}
+
+export function CreateGroupDialog({
+  owner,
+  isOpen,
+  onOpenChange,
+}: CreateGroupDialogProps) {
+  const [name, setName] = useState("");
+  const [selectedMemberIds, setSelectedMemberIds] = useState<Set<string>>(
+    () => new Set()
+  );
+
+  useEffect(() => {
+    if (isOpen) {
+      setName("");
+      setSelectedMemberIds(new Set());
+    }
+  }, [isOpen]);
+
+  const handleCreate = () => {
+    // TODO: wire backend
+    onOpenChange(false);
+  };
+
+  return (
+    <Dialog open={isOpen} onOpenChange={onOpenChange}>
+      <DialogContent size="xl" height="lg">
+        <DialogHeader>
+          <DialogTitle>New group</DialogTitle>
+        </DialogHeader>
+        <DialogContainer>
+          <div className="flex flex-col gap-5">
+            <Input
+              name="group-name"
+              label="Group name"
+              placeholder="e.g. Sales"
+              value={name}
+              onChange={(e) => setName(e.target.value)}
+              autoFocus
+            />
+            <MemberSelectionTable
+              owner={owner}
+              selectedMemberIds={selectedMemberIds}
+              onSelectionChange={(ids) => setSelectedMemberIds(ids)}
+              searchLabel="Members"
+            />
+          </div>
+        </DialogContainer>
+        <DialogFooter
+          leftButtonProps={{ label: "Cancel", variant: "ghost" }}
+          rightButtonProps={{
+            label: "Create",
+            variant: "primary",
+            onClick: handleCreate,
+            disabled: name.trim().length === 0,
+          }}
+        />
+      </DialogContent>
+    </Dialog>
+  );
+}
--- a/front/components/groups/CreateGroupDialog.tsx
+++ b/front/components/groups/CreateGroupDialog.tsx
@@ -39,7 +39,8 @@ export function CreateGroupDialog({
     onOpenChange(false);
   };
 
-  const shouldDisableButton = name.trim().length === 0 || selectedMemberIds.size === 0;
+  const shouldDisableButton =
+    name.trim().length === 0 || selectedMemberIds.size === 0;
 
   return (
     <Dialog open={isOpen} onOpenChange={onOpenChange}>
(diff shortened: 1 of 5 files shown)


front/components/groups/WorkspaceGroupsList.tsx
```diff
@@ -0,0 +1,145 @@
...
+function getGroupKindChip(kind: GroupKind): {
+  label: string;
+  color: ChipColor;
+} {
+  switch (kind) {
+    case "provisioned":
+      return { label: "Provisioned", color: "success" };
+    default:
+      return { label: kind, color: "primary" };
+  }
```
ykmsd, line 30: this needs to be adjusted later

front/components/groups/CreateGroupDialog.tsx
```diff
@@ -0,0 +1,78 @@
...
+  isOpen,
+  onOpenChange,
+}: CreateGroupDialogProps) {
+  const [name, setName] = useState("");
+  const [selectedMemberIds, setSelectedMemberIds] = useState<Set<string>>(
+    () => new Set()
+  );
+
+  useEffect(() => {
+    if (isOpen) {
```
ykmsd, line 31: it's unfortunate but right now this is the only way to reset the form without flickering

front/components/groups/WorkspaceGroupsList.tsx
```diff
@@ -0,0 +1,151 @@
...
+type ChipColor = NonNullable<React.ComponentProps<typeof Chip>["color"]>;
+
+function getGroupKindChip(kind: GroupKind): {
+  label: string;
+  color: ChipColor;
+} {
+  switch (kind) {
+    case "provisioned":
+      return { label: "Provisioned", color: "success" };
+    default:
```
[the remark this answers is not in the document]
ykmsd: 🤔

front/components/groups/CreateGroupDialog.tsx
```diff
@@ -0,0 +1,78 @@
...
+          <div className="flex flex-col gap-5">
+            <Input
+              name="group-name"
+              label="Group name"
+              placeholder="e.g. Sales"
+              value={name}
+              onChange={(e) => setName(e.target.value)}
+              autoFocus
+            />
+            <MemberSelectionTable
```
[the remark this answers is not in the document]
ykmsd: this is search so we don't fetch anything by default?

front/components/groups/CreateGroupDialog.tsx
```diff
@@ -0,0 +1,77 @@
...
+            />
+          </div>
+        </DialogContainer>
+        <DialogFooter
+          leftButtonProps={{ label: "Cancel", variant: "ghost" }}
+          rightButtonProps={{
+            label: "Create",
+            variant: "primary",
+            onClick: handleCreate,
+            disabled: name.trim().length === 0,
```
ykmsd: left a comment in matteo's pr but can we create a group without adding somebody?

front/components/groups/CreateGroupDialog.tsx
matteotrab: Does it flicker if we reset when the modal is closing?

front/components/groups/CreateGroupDialog.tsx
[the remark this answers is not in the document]
matteotrab: I think it would be nice to show some members even when not searching. Do we have an easy way to do it? If not it's fine like this

front/components/groups/CreateGroupDialog.tsx
matteotrab: I've changed it to make it mandatory to have members when creating a group

front/components/groups/WorkspaceGroupsList.tsx
```diff
@@ -0,0 +1,151 @@
...
+  Plus,
+  SearchInput,
+  Spinner,
+  Users01,
+} from "@dust-tt/sparkle";
+import type { CellContext, ColumnDef } from "@tanstack/react-table";
+import { useMemo, useState } from "react";
+
+// Only provisioned groups are surfaced in the workspace Groups tab for now.
+const GROUP_KINDS: GroupKind[] = ["provisioned"];
```
matteotrab: I think that we will end up using ["provisioned", "regular_manual"] in different places (in this page and in the governance one to start with). I think we could already define the group kinds we need in `front/types/groups.ts` as we do for `CAP_ELIGIBLE_GROUP_KINDS` so we can reuse it where we need. What do you think?

matteotrab (approved):
Thank you!

front/components/groups/CreateGroupDialog.tsx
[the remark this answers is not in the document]
ykmsd: I thought about it but it's purely for cosmetics than improving usability? 🤔 It's inside the modal so we can only show limited number of members and it's very less likely that you have somebody you want to have in a group you are creating 

I can still try and if it's easy let's show something 👍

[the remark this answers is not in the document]
matteotrab: Yes it would be mainly cosmetic. Let's keep it like this and see if we want to change it later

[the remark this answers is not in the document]
ykmsd: okay it's not hard (we need to remove `disabled: !searchText`) but let's keep as it is for now as you suggested 👍

front/components/groups/CreateGroupDialog.tsx
ykmsd: I will update the frontend too 👍

front/components/groups/CreateGroupDialog.tsx
ykmsd: so the problem is we have animation here, we don't want to reset the form state until animation is done (and this is where you would see the flickering). I can use `onAnimationEnd` something like this, but this relies on state being updated before animation ends (animation happens both on entrance and exit). I looked into it in the past and I thiiiink the best way to avoid using useEffect is we unmount the content components on closing, but this means we need to handle exist animation ourselves. I can look into it again since I hate this pattern 🫡

```
  const onAnimationEnd = () => {
    if (!isOpen) {
      setName("");
      setSelectedMemberIds(new Set());
    }
  };
```

matteotrab: I see thank you. A couple of followup questions
- If we reset the name and members state when opening the dialog, then we would have no flicker, correct? But that would imply moving the trigger inside this component
- Can we do
```ts
<DialogContent
    size="xl"
    height="lg"
    onCloseAutoFocus={() => {
        setName("");
        setSelectedMemberIds(new Set());
    }}
>
````
I'm not sure if this is the way that `onCloseAutoFocus` is supposed to be used, but in theory this would avoid the flickering

philipperolet:
Thx @ykmsd :pray:

PR #28870, opened 2026-07-12
Refactor sandbox function invocation lifecycle

spolu:
## Description

- Separate sandbox function invocation creation from execution while keeping execution synchronous (will soon not be called directly from invioke/createAndStartExecution but instead from a workflow/activity)
- Move execution and failure handling onto `SandboxFunctionInvocationResource`, emit the created event when the invocation is created, and persist an `errored` status before emitting terminal errors. 
- Lifecycle is handled by `createAndStartExecution`
- Moves the tests to their right place

This prepares the full invocation lifecycle for durable workflow execution in a follow-up PR.
See: https://github.com/dust-tt/dust/blob/main/x/spolu/tool_context/SANBOX_PLAN.md

## Tests

- `NODE_ENV=test npm test lib/resources/sandbox_function_resource.test.ts lib/api/sandbox_functions/call_sandbox_function.test.ts`
- `NODE_ENV=test npm test 'routes/w/[wId]/sandbox-functions/[functionId]/invocations.test.ts'`

## Risk

Low. The invocation remains synchronous and the API request/response schemas are unchanged.

## Deploy Plan

- deploy `front`

### DIFF
--- a/front-api/routes/w/[wId]/sandbox-functions/[functionId]/invocations.ts
+++ b/front-api/routes/w/[wId]/sandbox-functions/[functionId]/invocations.ts
@@ -113,16 +113,12 @@ app.post(
       });
     }
 
-    const invocation = await SandboxFunctionInvocationResource.createAndPublish(
-      auth,
-      {
+    const invocationResult =
+      await SandboxFunctionInvocationResource.createAndStartExecution(auth, {
         sandboxFunction,
-      }
-    );
-
-    const executionResult = await invocation.execute(auth, body);
-    if (executionResult.isErr()) {
-      await invocation.fail(executionResult.error);
+        body,
+      });
+    if (invocationResult.isErr()) {
       return apiError(
         ctx,
         {
@@ -132,13 +128,13 @@ app.post(
             message: "Sandbox function invocation failed.",
           },
         },
-        executionResult.error
+        invocationResult.error
       );
     }
 
     return ctx.json(
       {
-        invocation: invocation.toJSON(),
+        invocation: invocationResult.value.toJSON(),
       },
       201
     );
--- a/front-api/routes/w/[wId]/sandbox-functions/[functionId]/invocations.ts
+++ b/front-api/routes/w/[wId]/sandbox-functions/[functionId]/invocations.ts
@@ -1,6 +1,5 @@
 import { MCP_VALIDATION_OUTPUTS } from "@app/lib/actions/constants";
 import { validateSandboxFunctionAction } from "@app/lib/api/sandbox_functions/validate_action";
-import { SandboxFunctionInvocationResource } from "@app/lib/resources/sandbox_function_invocation_resource";
 import { SandboxFunctionResource } from "@app/lib/resources/sandbox_function_resource";
 import type {
   PostSandboxFunctionInvocationRequestBody,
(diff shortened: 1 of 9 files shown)


front-api/routes/w/[wId]/sandbox-functions/[functionId]/invocations.ts
```diff
@@ -113,8 +114,22 @@ app.post(
       });
     }
 
-    const invocationResult = await sandboxFunction.invoke(auth, body);
-    if (invocationResult.isErr()) {
+    const invocation = await SandboxFunctionInvocationResource.makeNew(auth, {
+      sandboxFunction,
+    });
+    const invocationType = invocation.toJSON();
+    await publishSandboxFunctionInvocationEvent(
```
Comment (flvndvd) [inline front-api/routes/w/[wId]/sandbox-functions/[functionId]/invocations.ts]: I'd argue that this event should live in the  `SandboxFunctionInvocationResource` itself.

front-api/routes/w/[wId]/sandbox-functions/[functionId]/invocations.ts
```diff
@@ -113,8 +114,22 @@ app.post(
...
+        type: "sandbox_function_invocation_created",
+        created: invocation.createdAt.getTime(),
+        invocation: invocationType,
+      },
+      { invocationId: invocation.sId }
+    );
+
+    const executionResult = await invocation.execute(auth, body);
+    if (executionResult.isErr()) {
+      await invocation.fail(executionResult.error);
```
Comment (flvndvd) [inline front-api/routes/w/[wId]/sandbox-functions/[functionId]/invocations.ts]: OOC why don't execute set the failed state itself?

front/lib/resources/sandbox_function_invocation_resource.ts
```diff
@@ -11,11 +23,38 @@ import {
...
+} from "@app/types/api/sandbox_functions";
+import { isDevelopment } from "@app/types/shared/env";
 import type { ModelId } from "@app/types/shared/model_id";
 import { Err, Ok, type Result } from "@app/types/shared/result";
 import { normalizeError } from "@app/types/shared/utils/error_utils";
+import { truncate } from "@app/types/shared/utils/string_utils";
 import type { Attributes, Transaction } from "sequelize";
 
+const SANDBOX_FUNCTION_WORKING_DIRECTORY = "/home/agent";
+const SANDBOX_FUNCTION_EXEC_TIMEOUT_MS = 2 * 60 * 1000;
```
flvndvd, line 39: Wondering if we should not align this to the bash command we expose where model can provide a timeout. We've noticed that 2 minutes is sometimes too short for `dsbx`.

spolu: Agreed but this is just a code move, belongs to another stream tbh so will leave as is for now 👍

front-api/routes/w/[wId]/sandbox-functions/[functionId]/invocations.ts
Comment (spolu) [reply front-api/routes/w/[wId]/sandbox-functions/[functionId]/invocations.ts]: Yeah... could add an invocation creation + event sending static function there.... didn't feel fully satisfied with it but will reintroduce it

Comment (spolu) [reply front-api/routes/w/[wId]/sandbox-functions/[functionId]/invocations.ts]: Introduced `createAndStartExecution` (this will soon create a workflow vs execute sync)

front-api/routes/w/[wId]/sandbox-functions/[functionId]/invocations.ts
Comment (spolu) [reply front-api/routes/w/[wId]/sandbox-functions/[functionId]/invocations.ts]: Introduced `createAndStartExecution` (this will soon create a workflow vs execute sync)

spolu:
r? @flvndvd PTAL
cc @PopDaph

spolu:
PMRR

flvndvd (commented):
LGTM

PR #28872, opened 2026-07-13
[BILLING] chunk Metronome seat ID edits to respect the 1000-seat cap

pmilliotte:
## Description

Chunk `addSeatIds`/`removeSeatIds` into ≤1000-ID batches across sequential `contracts.edit` calls so large seat reconciles no longer hit Metronome's 1000-seat-per-edit limit (was throwing `400 Number of subscription seat IDs must be less than or equal to 1000` in `syncMetronomeSeatCountActivity`).

## Tests

Added unit tests in `client.test.ts` covering single edit at the cap, chunking above the cap (no batch >1000, each seat sent once, unassigned delta on the final edit), and the no-op case. All 22 pass; typecheck clean.

## Risk

Low.

## Deploy Plan

- Deploy `main` on `front`

### DIFF
--- a/front/lib/metronome/client.test.ts
+++ b/front/lib/metronome/client.test.ts
@@ -6,6 +6,7 @@ import {
   createMetronomeContract,
   createMetronomeCredit,
   findSeatCreditSegmentForPeriod,
+  updateSubscriptionSeats,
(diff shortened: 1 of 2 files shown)


tdraier (approved):
🙏

PR #28873, opened 2026-07-13
[SLACK] Fix syncOneMessageDebounced infinite loop in syncNonThreaded

pmilliotte:
## Description

Fixes the `syncOneMessageDebounced` workflow timing out: `syncNonThreaded` now paginates via Slack's `next_cursor` instead of manually decrementing a `latest` bound truncated to whole seconds (`parseInt(message.ts)`), which could loop forever on a second holding ≥100 messages and burn the activity's 60-min `startToCloseTimeout` on repeated retries.

## Tests

No.

## Risk

Low.

## Deploy Plan

- Deploy `main` on `connectors`

### DIFF
--- a/connectors/src/connectors/slack/temporal/activities.ts
+++ b/connectors/src/connectors/slack/temporal/activities.ts
@@ -485,7 +485,7 @@ export async function syncNonThreaded({
   const slackClient = await getSlackClient(connectorId);
 
   let hasMore: boolean | undefined = undefined;
-  let latestTsSec = endTsSec;
+  let nextCursor: string | undefined = undefined;
   const seenMessagesTs = new Set<string>();
   do {
     let c: ConversationsHistoryResponse | undefined = undefined;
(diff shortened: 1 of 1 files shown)


connectors/src/connectors/slack/temporal/activities.ts
```diff
@@ -507,8 +507,9 @@ export async function syncNonThreaded({
               channel: channelId,
               limit: CONVERSATION_HISTORY_LIMIT,
               oldest: `${startTsSec}`,
-              latest: `${latestTsSec}`,
+              latest: `${endTsSec}`,
```
flvndvd, line 510: IIRC at some point `next_cursor` and `latest`/`oldest` were exclusive parameters. Did they lift the restriction?

flvndvd (approved):

connectors/src/connectors/slack/temporal/activities.ts
pmilliotte: Looks like it
https://docs.slack.dev/reference/methods/conversations.history/#pagination-time

> This form of pagination can be used in conjunction with cursors.

PR #28874, opened 2026-07-13
Add endDate / platform fee / logs

tdraier:
## Description

Extends the poke "Switch contract" dialog (`SwitchContractDialog`) with several operator-requested capabilities, and consolidates the request schema between the dialog and the server.

- **Optional contract end date.** Adds an "Ends at (UTC)" field alongside the existing start date, with the same hour-snapping and local-time label. Sent as `endingAt` and applied post-provision via a dedicated `v1.contracts.updateEndDate` call (`scheduleMetronomeContractEnd`) — left blank, the contract stays open-ended. 
Screenshot 2026-07-13 at 12 12 42

- **Scheduled charge ("Platform Fee").** New toggle section lets an operator add a pure invoice-only charge (no credit grant) with the same payment-schedule shape as commits: one-time or monthly/quarterly/semi-annually/annually with a number of periods, plus an arbitrary amount and optional display name. Applied via Metronome's `add_scheduled_charges` contract edit against a new "Platform Fee" FIXED product.
Screenshot 2026-07-13 at 12 12 36
- **Stripe customer is now mandatory for every switch** (was optional, for a "free tier, no Stripe link" path that's no longer used — there is no free tier package anymore, and the dialog's own submit button already required a resolved currency in practice). `resolvedCurrency` is now non-nullable throughout `switch_contract.ts`, removing a number of now-unreachable "requires a Stripe customer" checks.
- **The dialog now shows nothing beyond the Stripe Customer ID field** (plus its resolution status) until that customer resolves to a valid billing currency — no package, seat, or credit configuration is rendered for an unresolved/invalid customer.
- **Schema consolidation.** Moved the request schema into an isomorphic module (`types/poke/switch_contract.ts`, no server-only imports) shared by both the server (`lib/api/poke/switch_contract.ts`) and the dialog. The dialog's form schema is now built directly on top of the server's `SwitchContractBodySchema` via `.omit()`/`.extend()` instead of hand-duplicating ~20 fields; only the few fields whose *editing* state genuinely differs from the wire payload are overridden. `seats` is now `Record<seatType, ...>` end-to-end (was an array server-side) since there's a single first-party caller. "Force seat type" now reuses `promoteNoneSeatsTo` directly (a `SliderToggle` + `undefined` represents "no override") instead of a separate sentinel-string field.
- **Logging.** Logs the full request body (with `workspaceId`) right before provisioning the contract, to make it easier to debug switch-contract issues from Datadog/logs alone.

## Tests

- `npx tsgo --noEmit` clean on `front` and `front-api`.
- Existing `front-api/routes/poke/workspaces/[wId]/switch_contract.test.ts` (19 tests) pass unchanged.
- Manually exercised the dialog locally (package/tier selection, end date, scheduled charge toggle, seats).

## Risk

Medium — `stripeCustomerId` becoming mandatory and `seats` becoming a record are breaking changes to this private endpoint. Low practical impact: this poke endpoint has a single first-party caller (this dialog), which already required a Stripe customer via its own submit-button gating, and no script constructs a `seats` array literal. The "Platform Fee" Metronome product was provisioned in both sandbox and production via `scripts/metronome_setup.ts` before merge.

## Deploy Plan

Standard deploy. No migration or feature flag required.

### DIFF
--- a/front/components/poke/subscriptions/SwitchContractDialog.tsx
+++ b/front/components/poke/subscriptions/SwitchContractDialog.tsx
@@ -3,7 +3,6 @@ import {
   InputField,
   SelectField,
 } from "@app/components/poke/shadcn/ui/form/fields";
-import type { SwitchContractBodySchema } from "@app/lib/api/poke/switch_contract";
 import { clientFetch } from "@app/lib/egress/client";
 import { amountCents } from "@app/lib/metronome/amounts";
 import { isPaygEligibleTier } from "@app/lib/metronome/types";
@@ -18,12 +17,10 @@ import {
   usePokeStripeCustomerCurrency,
 } from "@app/lib/swr/poke";
 import { usePokePluginAsyncArgs } from "@app/poke/swr/plugins";
-import {
-  BILLABLE_SEAT_TYPES,
-  isMembershipSeatType,
-} from "@app/types/memberships";
+import { BILLABLE_SEAT_TYPES } from "@app/types/memberships";
 import { isCreditPricedPlan } from "@app/types/plan";
 import { CreditUsageConfigurationSchema } from "@app/types/poke/credit_usage_configuration";
+import { SwitchContractBodySchema } from "@app/types/poke/switch_contract";
 import type { WorkspaceType } from "@app/types/user";
 import {
   Button,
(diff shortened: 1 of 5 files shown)


tdraier:

- #28874 ◀

front/components/poke/subscriptions/SwitchContractDialog.tsx
```diff
@@ -667,6 +731,43 @@ export default function SwitchContractDialog({
...
+        if (
+          values.scheduledChargeFrequency !== "one_time" &&
+          (values.scheduledChargePeriods === undefined ||
+            values.scheduledChargePeriods < 2)
+        ) {
+          setError(
+            "Scheduled payments require a number of periods of at least 2."
+          );
+          return;
+        }
```
pmilliotte: Looks like we could handle these checks in the `SwitchContractFormSchema`. Either in `refine` or better with a `union`

pmilliotte (approved):
Screenshots welcomed in the pr description when there are UI changes 🙏

PR #28876, opened 2026-07-13
Add cost breakdown in poke

tdraier:
## Description

Add a message-level cost breakdown to the poke conversation view, showing the **stored** credit cost (`AgentMessageModel.costCredits`, persisted by the billing pipeline) side-by-side with a **recalculated** cost derived fresh from the message's raw token usage and tool actions. Useful for auditing/debugging billing discrepancies without leaving poke.

Screenshot 2026-07-13 at 10 05 12

For each agent message:
- A "Cost" panel showing stored AWU vs. recalculated AWU (split into LLM / tools), with a "mismatch" warning chip if the two disagree.
- An expandable per-model table (provider/model, prompt/completion/cached tokens, $ cost, AWU).
- An inline AWU chip on each tool call row (category + free/billed).

Implementation:
- `lib/api/assistant/credit_cost.ts`: new `computeAgentMessageCreditsBreakdown()`, mirroring the existing `computeAgentMessageCredits()` (used to persist `costCredits`) but split by LLM vs. tool and by model/tool instead of only summed.
- `lib/poke/conversation.ts`: `getPokeConversation()` batch-fetches runs + token usage for the whole conversation (2 queries total, no N+1), regroups by message, and attaches the recalculated breakdown alongside the stored `costCredits`.
- `types/poke/index.ts`: `PokeAgentMessageType` gains an optional `costBreakdown` field.
- `components/poke/pages/ConversationPage.tsx`: renders the new cost panel and per-tool cost chips.

## Tests

Manually verified via `npx tsgo --noEmit` (no new type errors) and `biome check --write` on the touched files. Not exercised against a live poke conversation in this session.

## Risk

Poke-only, read-only addition (no new writes, no changes to the billing pipeline itself). Adds 2 extra batched DB queries per conversation page load (`RunResource.listByDustRunIds` + `listRunUsagesForRuns`), scoped to the conversation's own runs.

## Deploy Plan

Standard deploy, no migration or feature flag needed.

### DIFF
--- a/front/lib/api/assistant/credit_cost.ts
+++ b/front/lib/api/assistant/credit_cost.ts
@@ -3,10 +3,15 @@ import type { ToolExecutionStatus } from "@app/lib/actions/statuses";
 import { isToolExecutionStatusFinal } from "@app/lib/actions/statuses";
 import { getToolNameFromFunctionCallName } from "@app/lib/actions/tool_display_labels";
 import { makeFairUseAwuCreditsRateLimitKeyForUser } from "@app/lib/api/assistant/rate_limits";
+import type { ToolCostCategory } from "@app/lib/api/mcp";
 import type { Authenticator } from "@app/lib/auth";
 import {
+  awuFromMicroUsd,
   computeRunKey,
+  getToolBillingInfo,
   intelligenceAwuFromRunUsagesGroupedByRunKey,
+  isFreeOrigin,
+  toolAwuFromAction,
   toolAwuFromActions,
 } from "@app/lib/metronome/events";
 import { AgentMCPActionResource } from "@app/lib/resources/agent_mcp_action_resource";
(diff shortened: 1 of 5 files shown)


tdraier:

- #28876 ◀

front/lib/api/assistant/credit_cost.ts
```diff
@@ -22,13 +27,145 @@ import {
...
+}
+
+/**
+ * Same computation as `computeAgentMessageCredits`, but split by LLM vs. tool
+ * cost and by model / tool so it can be displayed (e.g. in poke) instead of
+ * only summed. `byModel[].awu` is computed by grouping per runKey first
+ * (matching `intelligenceAwuFromRunUsagesGroupedByRunKey`), so the sum of
+ * `byModel[].awu` always equals `llmAwu` exactly.
+ */
+export function computeAgentMessageCreditsBreakdown({
```
Nils-Fedrigo: nit: refactor the computeAgentMessageCredits to use sum(computeAgentMessageCreditsBreakdown)

front/lib/api/assistant/credit_cost.ts
```diff
@@ -22,13 +27,145 @@ import {
...
+  const finalActions = actions.filter((a) =>
+    isToolExecutionStatusFinal(a.status)
+  );
+
+  const modelUsageByKey = new Map<
+    string,
+    Omit<AgentMessageCreditsModelBreakdown, "awu">
+  >();
+  for (const usage of runUsages) {
+    const key = `${usage.providerId}|${usage.modelId}`;
```
Nils-Fedrigo: with the reasoning effort too in the key? Not sure how that is computed in Metronome

front/lib/api/assistant/credit_cost.ts
```diff
@@ -22,13 +27,145 @@ import {
...
+        existing.cacheCreationTokens + (usage.cacheCreationTokens ?? 0),
+      costMicroUsd: existing.costMicroUsd + usage.costMicroUsd,
+    });
+  }
+
+  const modelAwuByKey = new Map<string, number>();
+  if (!isFreeOrigin(contextOrigin)) {
+    const byRunKey = new Map<string, RunUsageType[]>();
+    for (const usage of runUsages) {
+      const runKey = usage.runKey ?? "__legacy__";
```
Nils-Fedrigo: small nit: I don't remember if I put it that way but that probably shouldn't be a magic constant

front/lib/api/assistant/credit_cost.ts
```diff
@@ -22,13 +27,145 @@ import {
...
+    };
+  });
+
+  const llmAwu = intelligenceAwuFromRunUsagesGroupedByRunKey(
+    runUsages,
+    contextOrigin
+  );
+  const toolAwu = toolAwuFromActions(finalActions, contextOrigin);
+
+  return { llmAwu, toolAwu, totalAwu: llmAwu + toolAwu, byModel, byTool };
```
Nils-Fedrigo: IIRC we have the same plumbing for putting the data in ES, worth not reduplicating it again if possible and have a single source of truth

Nils-Fedrigo: you could even pull the ES data?

Nils-Fedrigo (approved):

Nils-Fedrigo:
I think it makes sense to reuse the same function everywhere!

front/lib/api/assistant/credit_cost.ts
tdraier: i don't think reasoning effort is changing the cost ?

front/lib/api/assistant/credit_cost.ts
tdraier: it was a local const, exporting

front/lib/api/assistant/credit_cost.ts
tdraier: yes, better - still the cost per individual tool is calculated runtime

PR #28856, opened 2026-07-10
[front] enh: remove inline manifest in Computer skill instructions

aubin-tchoi:
## Description

The Computer skill currently injects in its instructions a yaml manifest of all the libraries that are available at runtime. This list started small but grew big over time, now accounting for 2,548 tokens (see trace [here](https://langfuse.dust.tt/project/cmi4ft2qm00061707q4a9azyc/traces?filter=metadata%3BstringObject%3BconversationId%3B%3D%3BH3rmurWjSK&peek=72ca03e416f06d9c0e3ceed4a6882d7c&timestamp=2026-07-13T07%3A44%3A55.589Z&observation=8475ff7dce2c7589)).

This PR replaces the inline manifest with an on-demand call to `describe_toolset`..

## Tests

- Updated existing tests.

## Risk

- Low.

## Deploy Plan

- Deploy front.

PR #28878, opened 2026-07-13
[front] enh: add back concise Computer tool manifest

aubin-tchoi:
## Description

Follow-up on #28856 that removed entirely the YAML manifest where we mention which libs/binaries are available on the sandbox in the skill's instructions. Removing it saves 2,548 tokens, but means the agent has no idea which tools and libraries are available until it calls `describe_toolset`. This is not an issue for the very common ones (`git`, `curl`, `wget`, `jq` for instance), but is more problematic for the non-standard ones we provide like `xlsx_inspect`, `docx_inspect`, etc..

Now amounts to 549 tokens.

This PR adds a compact list with one line per runtime and pinned versions where available, totaling roughly 800 characters. Full descriptions and usage metadata are exposed for the tools we provide that are not industry standards.

## Tests

- Tested locally on a local excel file with formulas.
- Updated existing tests.

## Risk

- Low.

## Deploy Plan

- Deploy front.

### DIFF
--- a/front/lib/api/sandbox/image/tool_manifest.ts
+++ b/front/lib/api/sandbox/image/tool_manifest.ts
@@ -4,6 +4,7 @@ import type {
   ToolManifest,
   ToolRuntime,
 } from "@app/lib/api/sandbox/image/types";
+import { TOOL_RUNTIMES } from "@app/lib/api/sandbox/image/types";
 import * as yaml from "js-yaml";
 
 export function createToolManifest(tools: readonly ToolEntry[]): ToolManifest {
@@ -12,7 +13,6 @@ export function createToolManifest(tools: readonly ToolEntry[]): ToolManifest {
     python: [],
     node: [],
   };
-  const dustTools: ManifestToolEntry[] = [];
 
   for (const tool of tools) {
     const entry: ManifestToolEntry = {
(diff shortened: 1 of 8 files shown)


front/lib/resources/skill/code_defined/global/sandbox.ts
```diff
@@ -320,7 +383,7 @@ export const sandboxSkill = {
     "Run code, scripts, and shell commands in the conversation's Computer (a sandboxed Linux environment).",
   agentFacingDescription:
     "Execute code and commands in an isolated Linux sandbox. Useful to parse lengthy tool outputs, run code, " +
-    "process data, install packages, manipulate files, or perform any task requiring shell access. " +
+    "process data, manipulate files, or perform any task requiring shell access. " +
```
aubin-tchoi, line 388: removed "install packages" because it's not true

front/lib/api/sandbox/image/tool_manifest.ts
```diff
@@ -44,6 +44,24 @@ export function toolManifestToJSON(manifest: ToolManifest): string {
   return JSON.stringify(manifest, null, 2);
 }
 
+export function toolManifestToCompactText(manifest: ToolManifest): string {
+  return TOOL_RUNTIMES.flatMap((runtime) => {
```
flvndvd, line 48: Let's sort to make sure that we don't burst the prompt by mistake, even though it's hardly defined in code, ceinture bretelle.

front/lib/resources/skill/code_defined/global/sandbox.ts
```diff
@@ -280,8 +290,28 @@ value you should not have — apologize, do not retry the command, and do not
...
+  if (tools.size === 0) {
+    return "";
+  }
+
+  const descriptions = [...tools.values()]
+    .map((tool) => {
+      const summary = tool.description.match(/^.*?\.(?=\s|$)/)?.[0];
+      return `- \`${tool.name}\`: ${summary ?? tool.description}`;
+    })
+    .join("\n");
```
flvndvd, line 309: Sort.

flvndvd (approved):

PR #28880, opened 2026-07-13
Fix tier resolution for generated custom models

Nils-Fedrigo:
## Description

Fixes agent executions crashing before the provider call when a generated custom/EAP model is selected while model-tier enforcement is enabled.

- Widen `ModelTierSelection` to include generated custom model IDs.
- Keep `STATIC_MODEL_TIERS` compile-time exhaustive for static models by removing the widening annotation and retaining the `satisfies StaticModelTiersMap` check.
- Classify generated custom models, including `claude-fruitcake-eap`, as `premium` by default instead of indexing the static map with an absent key.
- Route full selections through the defensive lookup so future custom models cannot trigger the same `Cannot read properties of undefined` crash.
- Pass the resolved per-message reasoning effort to the runtime access check.
- Add a focused regression test for the EAP model.

This keeps custom models outside `STATIC_MODEL_TIERS`, since they are generated from GCS at build time and cannot be exhaustively represented in the checked-in static map. The conservative `premium` default ensures new custom models remain access-controlled immediately.

## Tests

- Added a unit regression test covering `claude-fruitcake-eap` as a generated custom model.
- Ran `git diff --check` successfully.
- Local unit tests were not run because the Computer clone does not include installed repository dependencies. Validation is delegated to PR CI.

## Risk

Low to moderate. The behavioral change is that all generated custom models default to the `premium` tier. This is intentionally conservative, but users without premium-tier access will no longer be able to select those models unless an explicit custom-model tier mechanism is added later.

The change is safe to roll back by reverting this PR.

## Deploy Plan

Deploy front normally. After deployment, verify that a `claude-fruitcake-eap` per-message selection reaches model execution or returns the expected model-tier access error, without an activity failure.

### DIFF
--- a/front/lib/api/assistant/token_pricing/tiers.test.ts
+++ b/front/lib/api/assistant/token_pricing/tiers.test.ts
@@ -13,11 +13,11 @@ import {
   STATIC_MODEL_IDS,
   SUPPORTED_MODEL_CONFIGS,
 } from "@app/types/assistant/models/models";
-import type { ModelIdType } from "@app/types/assistant/models/types";
 import {
   GPT_5_5_MODEL_ID,
   GPT_5_NANO_MODEL_ID,
 } from "@app/types/assistant/models/openai";
+import type { ModelIdType } from "@app/types/assistant/models/types";
 import { getAvailableReasoningEfforts } from "@app/types/assistant/models/types";
 import { GROK_4_FAST_REASONING_MODEL_CONFIG } from "@app/types/assistant/models/xai";
 import { describe, expect, it } from "vitest";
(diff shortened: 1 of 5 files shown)


front/lib/api/assistant/token_pricing/tiers.test.ts
```diff
@@ -72,6 +73,16 @@ describe("token_pricing/tiers", () => {
     ).toBe("premium");
   });
 
+  it("classifies generated custom models as premium", () => {
+    // Custom models are generated from GCS at build time and are therefore not
+    // present in the checked-in CUSTOM_MODEL_IDS fixture.
+    const customModelId: ModelIdType = "claude-fruitcake-eap";
```
Nils-Fedrigo:
```suggestion
    const customModelId = "my-custom-model-from-eap" as ModelIdType;
```

Nils-Fedrigo:

- #28880 ◀

Fraggle (commented):
LGTM

PR #28887, opened 2026-07-13
fix: reset member search pagination when search term changes

adrsimon:
## Description

Typing a new search term on page 2+ of the admin members list didn't reset pagination back to page 1, so a search with fewer results than the current offset silently rendered an empty list even though the count was correct.

## Tests

Manual — searched from page 2 and confirmed results now show on page 1.

## Risk

Low, single-file effect dependency fix.

## Deploy Plan

Standard deploy.


aubin-tchoi (approved):
LGTM

PR #28888, opened 2026-07-13
Route dust-oai to GPT-5.6 Sol and add Luna variants

fontanierh:
## Description

The internal `dust-oai` evaluation agents still target GPT-5.5. With the [GPT-5.6 family](https://openai.com/index/previewing-gpt-5-6-sol/) available, we need Sol coverage plus a parallel Luna series for side-by-side evaluation.

- Routes `dust-oai`, `dust-oai-medium`, and `dust-oai-high` to GPT-5.6 Sol.
- Adds `dust-oai-luna`, `dust-oai-luna-medium`, and `dust-oai-luna-high` on GPT-5.6 Luna.
- Keeps all six agents behind `dust_internal_global_agents`.

## Tests

Added focused coverage for model and reasoning-effort resolution across all six variants, plus Luna visibility without the internal-agent flag.

## Risk

Worst case, the internal agents select the wrong GPT-5.6 model or reasoning effort. They remain behind the existing internal flag and the change is safe to rollback.

## Deploy Plan

Standard deploy.


Nils-Fedrigo (approved):
add terra while you are at it 🙏 ?

fontanierh:
@Nils-Fedrigo 

Do you feel it's needed ? My understanding is that there is always a Sol or Luna reasoning effort that is better than terra and that there is ~no reason to use terra

PR #28890, opened 2026-07-13
Prune tool results by recency across the whole conversation, not per turn

flvndvd:
## Description

Follow up of https://github.com/dust-tt/dust/pull/28751.

A single agent turn used to be exempt from pruning as long as it was the one currently being answered, so a long tool-calling loop inside one response could grow the prompt without bound. This changes the redaction mechanism to look at tool results by plain recency across the entire conversation, current turn included, so an old tool call from three steps into the response being written right now gets treated exactly like an old tool call from three turns ago. Similar to how most context-editing feature work natively, having no notion of a current turn either.

Dropping a turn entirely, rather than just blanking out one of its tool results, still only ever happens to a whole turn at once, never partially, so a tool call can never end up separated from its result and an uploaded file can never end up separated from the message it belongs to. The turn actually being answered is never dropped under any circumstance, only ever redacted.

### DIFF
--- a/front/lib/api/assistant/conversation_rendering/index.test.ts
+++ b/front/lib/api/assistant/conversation_rendering/index.test.ts
@@ -208,9 +208,9 @@ describe("renderConversationForModel", () => {
     expect(res.value.prunedContext).toBe(false);
   });
 
-  it("redacts old tool outputs beyond TOOL_RESULTS_TO_PRESERVE but keeps the most recent ones, across separate interactions", async () => {
+  it("prunes old tool outputs beyond TOOL_RESULTS_TO_PRESERVE but keeps the most recent ones, across separate interactions", async () => {
     // 12 interactions (TOOL_RESULTS_TO_PRESERVE is 10), each with one tool call: the first 2 are
-    // outside the preserved window and must be redacted. The last 10 are the protected floor and
+    // outside the preserved window and must be pruned. The last 10 are the protected floor and
     // must survive untouched, regardless of budget-driven checkpoint search.
     const messages = [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12].flatMap((i) => [
       userMessage(`u_${i}`),
@@ -234,7 +234,7 @@ describe("renderConversationForModel", () => {
       prompt: "PROMPT",
       enabledSkills: [],
       tools: "TOOLS",
-      // Total unredacted: 12 * 5020 = 60_240. Redacting the 2 eligible tool results (outside the
+      // Total unpruned: 12 * 5020 = 60_240. Pruning the 2 eligible tool results (outside the
       // floor of 10) saves 2 * (5000 - 24) = 9_952, bringing it to 50_288, comfortably under
       // this budget. Nothing else needs to be touched.
       allowedTokenCount: computeAllowedTokenCount({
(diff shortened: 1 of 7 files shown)


front/lib/api/assistant/conversation_rendering/index.ts
```diff
@@ -23,31 +23,142 @@ import type { CredentialsType } from "@app/types/provider";
...
+ * floor, then force dropping past the protected floor.
+ *
+ * pruneToolResults and dropInteractionsToFit are each called twice, once with their normal floor
+ * and once with a floor of 0. This is safe because both floors work off the CURRENT state: a
+ * message pruneToolResults already redacted is already at the placeholder size, so redacting it
+ * again would save nothing and its own eligibility check excludes it, and an interaction
+ * dropInteractionsToFit already dropped is simply gone from the array, so there's nothing left
+ * for the second call to reconsider. Each call only ever reaches further than the one before it.
+ */
+function pruneConversationToBudget(
```
flvndvd, line 66: Kept in his file to ease this one review, will be moved to a dedicated file in a follow up PR.

front/lib/api/assistant/conversation_rendering/index.ts
```diff
@@ -23,31 +23,136 @@ import type { CredentialsType } from "@app/types/provider";
...
+      PREVIOUS_INTERACTIONS_TO_PRESERVE
+    );
+    if (previousAfter !== previousBefore) {
+      prunedContext = true;
+    }
+    pruned = [...previousAfter, currentInteraction];
+    totalTokens = sumInteractionTokens(pruned);
+  }
+
+  // Layer 3: pruning alone wasn't enough, force it past TOOL_RESULTS_TO_PRESERVE.
```
fontanierh: nit: comment slightly confusing / unclear. Would be helpful for the reader to understand that at this point, even the current interaction doesn't fit.

front/lib/api/assistant/conversation_rendering/pruning.ts
```diff
@@ -1,265 +1,257 @@
...
+ * the same input always prunes the same way, stable enough to survive the model provider's
+ * prompt cache. maxTokens should be stable across turns for that to hold. It doesn't have to be
+ * byte-identical: the caller derives it from prompt and tool-definition sizes that drift a little
+ * between turns, and the checkpoint rounding absorbs drift well below PRUNING_CHECKPOINT_TOKENS.
+ * What it must not be is a fast-moving live quantity (remaining budget, time, message count).
+ */
+export function pruneToolResults(
+  interactions: InteractionWithTokens[],
+  maxTokens: number,
+  toolResultsToPreserve: number
```
aubin-tchoi: nit: could be passed in a structured object

front/lib/api/assistant/conversation_rendering/index.ts
fontanierh: Well, I guess not only current but `PREVIOUS_INTERACTIONS_TO_PRESERVE`

front/lib/api/assistant/conversation_rendering/index.ts
```diff
@@ -23,31 +23,136 @@ import type { CredentialsType } from "@app/types/provider";
...
+  // Layer 1: proactive pruning, within the protected floor, up to pruningBudget.
+  let pruned = pruneToolResults(
+    interactions,
+    pruningBudget,
+    TOOL_RESULTS_TO_PRESERVE
+  );
+  let prunedContext = pruned !== interactions;
+  let totalTokens = sumInteractionTokens(pruned);
+
+  // Layer 2: drop whole previous interactions, oldest first, never the current one.
```
fontanierh: "never the current one" is slightly misleading. It's more than that if PREVIOUS_INTERACTIONS_TO_PRESERVE > 1

front/lib/api/assistant/conversation_rendering/index.test.ts
```diff
@@ -401,55 +382,54 @@ describe("renderConversationForModel", () => {
...
     );
-    expect(currentTool.content).toBe("cur_function");
-    expect(res.value.prunedContext).toBe(false);
+    const tool12 = getFunctionMessage(
+      res.value.modelConversation.messages,
+      "tool_12"
+    );
+    expect(tool1.content).toContain("no longer available");
+    expect(tool12.content).toBe("result_12");
   });
```
aubin-tchoi, line 395: seem like we don't have a test checking layer 3 specifically, something like `"force-prunes protected tool results when dropping old interactions is not enough"`

fontanierh (approved):

aubin-tchoi (approved):
LGTM

PR #28891, opened 2026-07-13
[front] fix: align description size limit with Skill Builder in skill authoring

aubin-tchoi:
## Description

Skill authoring could create or update skills with agent-facing descriptions longer than the 750-character limit enforced by Skill Builder. This PR fixes that.

Also bumps the limit to a round 1k.

Not too convinced we need this limit, but at the very least the xp should be the exact same between the tools and manual actions in Skill Builder.

## Tests

- Tested locally.

## Risk

- Low.

## Deploy Plan

- Deploy front.


front/lib/skills/labels.ts
```diff
@@ -1,2 +1,5 @@
+// Maximum length (in characters) of a skill's agent-facing description.
```
aubin-tchoi, line 1: moved to break a dependency cycle

fontanierh (approved):
Lgtm

PR #28894, opened 2026-07-13
Resolve path-backed interactive file citations

fontanierh:
## Description

Follow-up of #28304, which exposes the linked file id on path metadata responses.

Interactive-content panels use `fileId` as their canonical identity, while markdown file directives initially have only a scoped path. Resolve that path on click before opening the panel, keeping the panel API and URL state `fileId`-only. Non-interactive files keep the existing preview flow.

Stack: 1 of 2. Next: #28895.

## Tests

- Added HEAD resolution coverage for linked, unlinked, missing, and error responses.
- Added component coverage proving a path-backed Frame citation opens the panel with `fileId`.
- Ran the focused Vitest suites and full Front `tsgo --noEmit`.

## Risk

Low. Worst case, a path-backed interactive citation fails to resolve and shows an error notification. Existing file-id panel callers and non-interactive previews are unchanged. Safe to rollback.

## Deploy Plan

Deploy normally.

### DIFF
--- a/front/components/assistant/conversation/attachment/PreviewableCitation.tsx
+++ b/front/components/assistant/conversation/attachment/PreviewableCitation.tsx
@@ -3,9 +3,15 @@ import type {
   FileCitationCardSize,
 } from "@app/components/assistant/conversation/attachment/FileCitationCard";
 import { FileCitationCard } from "@app/components/assistant/conversation/attachment/FileCitationCard";
+import { ConversationSidePanelContext } from "@app/components/assistant/conversation/ConversationSidePanelContext";
 import { useFilePreviewContext } from "@app/components/assistant/conversation/FilePreviewContext";
+import { useSendNotification } from "@app/hooks/useNotification";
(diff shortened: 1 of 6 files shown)


front/components/assistant/conversation/attachment/PreviewableCitation.tsx
```diff
@@ -52,10 +58,37 @@ export function PreviewableCitation({
...
+        sidePanel.openPanel({
+          type: "interactive_content",
+          fileId: resolvedFileId,
+        });
+      } catch (error) {
+        sendNotification({
+          type: "error",
+          title: "Failed to open Frame",
+          description:
+            error instanceof Error ? error.message : "An error occurred.",
```
flvndvd: `nit`` we have `normalizeError`.

flvndvd (approved):

front/components/assistant/conversation/attachment/PreviewableCitation.tsx
fontanierh: good catch, thanks

PR #28897, opened 2026-07-13
Clarified type for fetchActiveByWorkspaceModelId

tdraier:
## Description

fetchActiveByWorkspaceModelId cannot return null/undefined as it was described in the type - this lead to invalid/dead code while testing if subscription is null. "No subscription" is actually always "FREE_NO_PLAN"

## Tests

locally

## Risk

low

## Deploy Plan

deploy front


tdraier:

- #28897 ◀

PopDaph (approved):
🙏🏻

PR #28898, opened 2026-07-13
Include the provider in the cache diagnostics key

flvndvd:
## Description

The prompt-cache diagnostics chain stores the previous response id in Redis keyed by conversation and agent. With the recent addition of tweaking model mid-flight in the conversation we need to account for this, as the next request could send a response id issued by a different provider, which is invalid and would failed the whole API call. This adds the provider to the key so a provider switch simply starts a fresh chain. The model itself is deliberately not part of the key: providers already detect model changes across consecutive requests and reports them in the diagnostics payload, so keeping the chain across model switches is exactly what lets us observe them.


flvndvd:
PMRR.

PR #28902, opened 2026-07-13
Eagerly load more file-related tools

flvndvd:
## Description

See thread [here](https://dust4ai.slack.com/archives/C0BCDAL5ZTM/p1783933116421429).

This PR eagerly loads a few more tools as part of the default set of tools available to the model. Those are part of the most frequently searched/used.


aubin-tchoi (approved):
LGTM

PR #9061, opened 2024-12-02
Public API & SDK get_agent_config fetches with authors + use in extension

PopDaph:
## Description

- Public API: adds query params `withAuthors` to also fetch authors when fetching agents configs, if set to true.
- SDK: Adds `includes` params in `getAgentConfigurations` method.
- Extension: sends the new includes param on the query to fetch Assistant favorites so we can display the authors. 

On the private api includes can be `authors` + `usage`, since `usage` is not used yet I decided to not add it already. It's easier to add it when needed over removing it if we want to. 

Screenshot 2024-12-02 at 16 49 06

Fixes: https://github.com/dust-tt/tasks/issues/1727

## Risk

Can be rolled back. 
The new param is optional in the SDK and in API call so should be safe.

## Deploy Plan

- Deploy front.

PR #28903, opened 2026-07-13
Add HubSpot custom object lookup and write support

Inesvvv:
## Description

Solves [HubSpot MCP] Enable custom object lookup and write  [#9061](https://github.com/dust-tt/tasks/issues/9061)
Adds HubSpot custom object support to the MCP server by extending the existing read tools (`search_crm_objects`, `get_object_properties`) to accept dynamic custom object type IDs, adding a `list_custom_object_schemas` discovery tool, and adding generic create_custom_object / update_custom_object write tools (gated by the new `crm.objects.custom.write` scope).

## Tests

Tested manually locally. 

## Risk

Low 
## Deploy Plan

Add scope crm.objects.custom.write to prod and then deploy

### DIFF
--- a/front/lib/actions/mcp_internal_actions/__snapshots__/mcp_servers_metadata.test.ts.snap
+++ b/front/lib/actions/mcp_internal_actions/__snapshots__/mcp_servers_metadata.test.ts.snap
@@ -985,6 +985,10 @@ exports[`MCP Servers Metadata Snapshot > should have stable tool billing info ac
       "freeUsage": false,
       "toolCostCategory": "advanced",
     },
+    "create_custom_object": {
+      "freeUsage": false,
+      "toolCostCategory": "advanced",
+    },
     "create_deal": {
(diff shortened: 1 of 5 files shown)


front/lib/actions/mcp_internal_actions/__snapshots__/mcp_servers_metadata.test.ts.snap
```diff
@@ -1069,6 +1073,10 @@ exports[`MCP Servers Metadata Snapshot > should have stable tool billing info ac
       "freeUsage": false,
       "toolCostCategory": "advanced",
     },
+    "list_custom_object_schemas": {
+      "freeUsage": false,
+      "toolCostCategory": "advanced",
```
matteotrab, line 1078: Should this be advanced? I'm not sure how we decide the category?

matteotrab (approved):
Thank you!

PR #28904, opened 2026-07-13
[front] fix: constrain Sidekick model suggestions to per-model reasoning efforts

adrsimon:
## Description

Sidekick could suggest `reasoningEffort: "none"` on a model even when the builder UI's reasoning-effort picker doesn't offer "none" for that model, so the suggestion couldn't be replicated manually.

Root cause: `suggest_model`'s schema validated `reasoningEffort` against the global `ORDERED_REASONING_EFFORTS` list rather than the target model's `supportedReasoningEfforts`, which is what the builder picker actually uses (`getAvailableReasoningEfforts`).

Changes:
- `suggest_model` handler now checks the suggested `reasoningEffort` against the model's `supportedReasoningEfforts` and returns an error listing the valid options if it's not supported.
- `formatAvailableModels` (Sidekick's model context) now includes each model's supported reasoning efforts, so it has a chance to pick a valid one up front instead of relying on the guardrail.

Fixes #9578

## Tests

Added a regression test using `claude-sonnet-4-6` (which doesn't support `none`) asserting `suggest_model` now errors out. Ran the full `agent_sidekick_context` test suite (63 tests, all passing).

## Risk

Low. Scoped to the Sidekick `suggest_model` tool; the new validation only rejects an already-invalid input, it doesn't change behavior for valid inputs.

## Deploy Plan

Standard deploy, no migration or feature flag involved.

### DIFF
--- a/front/lib/api/actions/servers/agent_sidekick_context/agent_sidekick_context.test.ts
+++ b/front/lib/api/actions/servers/agent_sidekick_context/agent_sidekick_context.test.ts
@@ -1780,6 +1780,39 @@ describe("agent_sidekick_context tools", () => {
       }
     });
 
+    it("returns error when suggesting an unsupported reasoning effort for the model", async () => {
+      const { authenticator } = await createResourceTest({ role: "admin" });
+
+      const agentConfiguration =
+        await AgentConfigurationFactory.createTestAgent(authenticator);
+
+      const { getAgentConfigurationIdFromContext } = await import(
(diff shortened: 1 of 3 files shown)


frankaloia (approved):
LGTM although for transparency, questioning if we should remove model selection entirely with the intro of auto model

adrsimon:
@frankaloia agreed in the long run but didn't want to take any non-thought-about decision today

PR #27545, opened 2026-06-17
Add selected Space runtime validation helpers

fontanierh:
## Stack

1. #29142 - Allow selecting open Spaces in conversations (merged)
2. **#27545 - Add selected Space runtime validation helpers**
3. #28905 - Use selected Spaces for runtime skill scope
4. #28906 - Inherit selected Spaces in sub-agent conversations
5. #28911 - Cover sub-agent cleanup boundaries

This PR is **399 changed lines** against `main`.

## Why

Selected-Space rows can become stale after access or Space changes, and child conversations may be created by system-key runs without a current user. Runtime consumers need one shared validation path, and child inheritance must preserve who originally selected each Space.

## What changed

- Adds `getValidSelectedSpaceIdsForAgentRun`, which returns active selected regular Spaces that remain readable while the feature is enabled.
- Adds `getEffectiveSpaceIdsForAgentRun`, which combines the agent's configured Spaces with valid conversation selections.
- Adds `copySelectedConversationSpacesToChild` for materializing valid selections into a child conversation.
- Extends selected-Space upserts to accept per-Space selector provenance, including userless and system-key execution.
- Revalidates selections independently, so one stale selection does not discard other valid selections.
- Preserves durable selected-Space ACL rows for system-key runs while treating selector identity as audit provenance.
- Adds focused coverage for feature gating, partial invalidation, open Space transitions, effective scope, child copying, and selector provenance.

## Risk

This PR only adds foundational helpers and provenance support. Runtime skill resolution and `run_agent` adoption are in the next PRs. Invalid selected rows are ignored instead of widening runtime scope. No schema or migration changes are included.

## Deploy plan

- [x] Merge and deploy #29142.
- [ ] Merge this PR and deploy `front` normally.
- [ ] Merge #28905, then #28906, then test-only #28911.

## Verification

- Focused runtime stack suite: 116 tests
- `npm run tsgo -- --noEmit` in `front`
- `npm run tsgo -- --noEmit` in `front-api`
- `npx biome check` on changed TypeScript files
- `git diff --check`

PR #28905, opened 2026-07-13
Use selected Spaces for runtime skill scope

fontanierh:
## Stack

1. #29142 - Allow selecting open Spaces in conversations (merged)
2. #27545 - Add selected Space runtime validation helpers (merged)
3. **#28905 - Use selected Spaces for runtime skill scope**
4. #28906 - Inherit selected Spaces in sub-agent conversations
5. #28911 - Cover sub-agent cleanup boundaries

This PR is **374 changed lines** against `main`.

## Why

Selected Spaces are persisted in the conversation ACL, but runtime skill and tool resolution still uses only the agent configuration's Spaces. Runtime must use the same validated per-conversation scope without recomputing it at each consumer or moving selected-Space instructions into the long-lived prompt cache.

## What changed

- Computes effective Space IDs in `SkillResource.listForAgentLoop` using the merged validation helper.
- Threads that scope through custom and global skill loading, MCP server views, skill instructions, and skill-management lookups.
- Returns `hasSelectedSpacesOutsideAgentScope` from the same computation for prompt-cache placement.
- Applies the scope in the normal agent loop, prompt commands, investigation tooling, and the Poke conversation renderer.
- Requires callers with agent-loop data to provide the already-computed scope.
- Adds focused coverage for selected-Space data sources, subset handling, open Space transitions, and prompt placement.

## Risk

Runtime scope widens only with active selected regular Spaces that remain readable. Agent-configured Spaces remain included. Selected-Space-dependent instructions stay outside the long-lived cache tier only when the conversation adds Spaces outside the agent's configured scope. No API schema or migration changes are included.

## Deploy plan

- [x] Merge and deploy #29142 and #27545.
- [ ] Merge this PR.
- [ ] Deploy `front` and `front-api`.
- [ ] Monitor selected-Space access failures and skill or MCP resolution errors.
- [ ] Merge #28906, then test-only #28911.

## Verification

- Focused runtime stack suite: 116 tests
- `npm run tsgo -- --noEmit` in `front`
- `npm run tsgo -- --noEmit` in `front-api`
- `npx biome check` on changed TypeScript files
- `git diff --check`

### DIFF
--- a/front/lib/api/actions/servers/skill_management/tools/index.ts
+++ b/front/lib/api/actions/servers/skill_management/tools/index.ts
@@ -48,12 +48,12 @@ async function findAvailableSkillForAgentLoop({
   agentLoopData: AgentLoopExecutionData;
   skillName: string;
 }): Promise<SkillResource | null> {
-  const { enabledSkills, equippedSkills, systemSkills } =
+  const { effectiveSpaceIds, enabledSkills, equippedSkills, systemSkills } =
     await SkillResource.listForAgentLoop(auth, agentLoopData);
   const userMessageSkills = await SkillResource.fetchByIds(
     auth,
     extractSkillIdsFromConversationMessages(agentLoopData),
-    { agentLoopData, onlyActive: true }
+    { agentLoopData, effectiveSpaceIds, onlyActive: true }
   );
   const directlyAllowedSkills = [
     ...enabledSkills,
@@ -76,6 +76,7 @@ async function findAvailableSkillForAgentLoop({
   );
   const candidate = await SkillResource.fetchByName(auth, skillName, {
     agentLoopData,
+    effectiveSpaceIds,
   });
   if (!candidate) {
     return null;
--- a/front/lib/api/assistant/generation.ts
+++ b/front/lib/api/assistant/generation.ts
@@ -394,6 +394,7 @@ export function constructPromptMultiActions(
     isNewFileExplorer = false,
     hasSandboxTools = false,
     disableFormattingPrompt = false,
+    hasSelectedSpacesOutsideAgentScope = false,
   }: {
     userMessage: AgentLoopExecutionData["userMessage"];
     agentConfiguration: AgentLoopExecutionData["agentConfiguration"];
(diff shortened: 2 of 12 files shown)


front/lib/api/actions/servers/skill_management/tools/index.ts
```diff
@@ -48,12 +48,12 @@ async function findAvailableSkillForAgentLoop({
   agentLoopData: AgentLoopExecutionData;
   skillName: string;
 }): Promise<SkillResource | null> {
-  const { enabledSkills, equippedSkills, systemSkills } =
+  const { effectiveSpaceIds, enabledSkills, equippedSkills, systemSkills } =
```
Fraggle, line 51: I'm probably missing some context but I wonder if it's possible to put the effective spaces retrieval directly in the agentLoopData ?

front/lib/api/assistant/generation.ts
```diff
@@ -463,28 +465,27 @@ export function constructPromptMultiActions(
     // Structured form with 3 cache tiers, ordered from most stable to most volatile.
     //
     // Instructions (long cache): stable per agent config, covering agent instructions, skills,
-    // format docs, and guidelines.
+    // format docs, and guidelines. If selected conversation Spaces are active, skill
+    // instructions can depend on per-conversation scope and must stay out of this tier.
```
Fraggle, line 469: WDYT of elaborating on how they can depend of per-conversation scope ?

Fraggle (commented):
LGTM

front/lib/api/actions/servers/skill_management/tools/index.ts
fontanierh: directionally i agree but not on `AgentLoopExecutionData` because that one is widely constructed, serialized across Temporal/tool boundaries, and sometimes synthesized in tests

Effective Space IDs are authorization-sensitive derived data that should be revalidated IMO

a cleaner follow-up would introduce an enriched runtime context containing:

```
{
  agentLoopData,
  effectiveSpaceIds
}
```

we compute it once per activity/tool execution, then require downstream consumers to use it

Will see what I can do !

fontanierh: thx for suggesting

front/lib/api/assistant/generation.ts
fontanierh: aligned

front/lib/api/actions/servers/skill_management/tools/index.ts
Fraggle: I'm not 100% sure but I think we already operate under the assumption that we lookup permissions only at the beggining of the loop.

PR #28908, opened 2026-07-13
fix(oauth): audience field + clearer DCR error for MCP static OAuth

adrsimon:
## Description

Two small follow-ups from unblocking a customer on tasks#9540 (Static OAuth against an audience-scoped Auth0 tenant):

1. Adds an optional `resource`/`audience` field to the MCP static OAuth setup form. Some OAuth servers issue audience-scoped tokens (RFC 8707) and reject requests that omit this parameter; the backend already supports passing it through `extraConfig`, it just wasn't exposed in the form.
2. When discovering OAuth metadata for a remote MCP server, skip the dynamic client registration attempt if the server doesn't advertise a `registration_endpoint` at all, and return a clearer message pointing at Static OAuth instead of the generic "failed to register client" error (which reads as a bug rather than an expected limitation).

## Tests

Typecheck passes on the changed files. Not manually tested end-to-end against a real audience-scoped OAuth server.

## Risk

Low 

## Deploy Plan

Standard deploy, no migration or ordering concerns.

### DIFF
--- a/front/lib/resources/remote_mcp_servers_resource.ts
+++ b/front/lib/resources/remote_mcp_servers_resource.ts
@@ -527,21 +527,6 @@ export class RemoteMCPServerResource extends BaseResource<RemoteMCPServerModel> 
       }
     }
 
-    // Some authorization servers only support pre-registered (static) clients and don't
-    // advertise a registration endpoint at all. Attempting DCR in that case always fails and
-    // the resulting error is misleading (it looks like a failed registration attempt rather
-    // than "this server doesn't support automatic setup").
-    if (!metadata.registration_endpoint) {
-      return new Err(
-        new DustError(
-          "internal_error",
-          "This server does not support automatic OAuth setup (no dynamic client " +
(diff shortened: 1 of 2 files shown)


front/lib/resources/remote_mcp_servers_resource.ts
```diff
@@ -527,6 +527,21 @@ export class RemoteMCPServerResource extends BaseResource<RemoteMCPServerModel>
       }
     }
 
+    // Some authorization servers only support pre-registered (static) clients and don't
+    // advertise a registration endpoint at all. Attempting DCR in that case always fails and
+    // the resulting error is misleading (it looks like a failed registration attempt rather
+    // than "this server doesn't support automatic setup").
+    if (!metadata.registration_endpoint) {
```
aubin-tchoi: it always fails when `registration_endpoint` is not defined? if we want to be safe we could also do this check after it fails, when rewriting the error

aubin-tchoi (approved):
LGTM

PR #28913, opened 2026-07-13
Batch update seat types in past

tdraier:
## Description

One-off admin script to correct a batch of memberships onto a target seat type, backdated to a manual effective date, and push the corrected seat counts to Metronome as of that same date (rather than only from now on).

Usage:
```
npx tsx scripts/backdate_membership_seat_types_from_csv.ts \
  --wId <workspaceSId> --csvPath <path-to-csv-with-email-column> \
  --seatType <target> --effectiveDate <ISO date> [--execute]
```

Dry-run by default (omit `--execute`).

## Tests

Manual dry-run against the affected workspace to confirm the resolved user list and current→target seat types before executing.

## Risk

Backdating only works if the affected billing period hasn't been invoiced/finalized in Metronome yet — confirmed this is the case for the target workspace. Script writes membership seat type immediately (no scheduling) and reconciles per-user credit state after the Metronome sync; failures during reconcile are logged but non-fatal.

### DIFF
--- a/front/lib/metronome/seats.ts
+++ b/front/lib/metronome/seats.ts
@@ -1330,6 +1330,7 @@ export async function syncSeatCount({
   startingAt,
   contract,
   assumeEmptySeats,
+  carryConsumptionAcrossStartingAt = false,
 }: {
   metronomeCustomerId: string;
   contractId: string;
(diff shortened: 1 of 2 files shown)


tdraier:

- #28913 ◀

PR #28916, opened 2026-07-13
[front] enh: explain knowledge tags in the skills prompt

aubin-tchoi:
## Description

Agents can find `<knowledge ... />` tags in enabled skill instructions, and sometimes fail to understand exactly how they can leverage them.
Example [here](https://langfuse.dust.tt/project/cmi4ft2qm00061707q4a9azyc/traces?filter=metadata%3BstringObject%3BconversationId%3B%3D%3BE9IA4IBFCQ&peek=83b989b195435612e0a30e06673bb625&timestamp=2026-07-13T13%3A02%3A31.727Z) where the agent runs a semantic search instead of directly reading the file.

This PR adds guidance explaining that the data can be read with the skill-attached `cat` tool using the tag `id` as `nodeId`, and that `list` or `semantic_search` can explore nodes with children.

## Tests

- Updated existing tests.

## Risk

- Low. Prompt-only change.
- Will burst the cache.

## Deploy Plan

- Deploy front.

### DIFF
--- a/front/lib/api/assistant/generation.test.ts
+++ b/front/lib/api/assistant/generation.test.ts
@@ -518,7 +518,10 @@ describe("constructPromptMultiActions - system prompt stability", () => {
       '`<knowledge id="..." title="..." ... />` tags, which point to specific workspace knowledge attached to the skill'
     );
     expect(text).toContain(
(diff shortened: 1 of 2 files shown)


flvndvd (approved):
LGTM

PR #28926, opened 2026-07-13
[LLM] Add gpt-5.6 sol/terra/luna support

pmilliotte:
## Description

Add support for the gpt-5.6 sol/terra/luna models (openai responses API, EU + global regions) across the model_constructors and dust stream layers, bumping the openai SDK to 6.46 to pick up the new response types.

## Tests

Added endpoint tests for all six model constructor stream endpoints; existing suite passes.

## Risk

Low.

## Deploy Plan

- Deploy `main` on `front`


adrsimon (commented):
LGTM

PR #28927, opened 2026-07-13
Batch interaction drops so conversations at the context ceiling stop paying a full cache miss on every call

flvndvd:
## Description

We observed in production that once a conversation fills its token budget, every call evicts the oldest interaction to make room for the new turn. Removing the head of the conversation shifts every message, so the provider's prompt cache misses entirely and the full conversation is re-ingested at full price on every request, often seconds apart, for as long as the conversation stays at the ceiling.

This change batches those evictions. When dropping is needed, instead of removing the strict minimum, we drop forward to a stable boundary derived from the conversation's token situation, the same batching idea already used for tool result pruning. Consecutive calls agree on where the head sits, and each drop buys enough headroom that the next several turns fit without dropping again. One deliberate full miss then amortizes over many calls instead of recurring on each one. When no such boundary exists in the droppable range, the drop stays minimal, since over-dropping a small conversation would erase content without gaining stability. The last-resort path that cuts past the protected floor also keeps minimal drops on purpose: the cache is already lost at that point and what remains is the most recent context, so it should shrink as little as possible.

### DIFF
--- a/front/lib/api/assistant/conversation_rendering/pruning.ts
+++ b/front/lib/api/assistant/conversation_rendering/pruning.ts
@@ -18,11 +18,17 @@ const PRUNED_TOOL_RESULT_PLACEHOLDER =
   "</dust_system>";
 const PRUNED_TOOL_RESULT_TOKENS = 24;
 
-// Batch size for the pruning checkpoint: like provider-native `clear_at_least`, don't move the
-// frontier for a handful of tokens, only once enough has accumulated to be worth invalidating the
-// cached prefix. Flat rather than a percentage of contextSize since no model in our fleet sits
-// between 16k-64k tokens. Starting point, tune against production cache-miss metrics.
+// Pruning advances through history in batches of this size, not message by message. Every move
+// of the pruning frontier invalidates the provider cache from that point on. Moving it for a
+// handful of tokens costs more than it saves, so it only moves once a batch has accumulated.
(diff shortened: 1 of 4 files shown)


front/lib/api/assistant/conversation_rendering/pruning.ts
```diff
@@ -235,26 +242,82 @@ export function pruneToolResults(
...
+ * checkpoint exists before the floor, the drop stays minimal. Extending to the floor would
+ * erase content without gaining any stability.
+ *
+ * With batchToCheckpoint false, the drop is exactly minimal. This is for last-resort callers
+ * that are already cutting into recent interactions and should cut as few as possible.
+ *
+ * One ordering rule keeps the checkpoints meaningful: this function must run after pruning,
+ * with a budget at most as tight as pruning's. Checkpoints only stabilize the drop boundary if
+ * the prefix sums are identical from one call to the next. That holds because by the time a
+ * conversation drops at all, everything prunable in the droppable range was already pruned, so
```
aubin-tchoi, line 263: not literally everything prunable I think, the last `TOOL_RESULTS_TO_PRESERVE ` are not pruned at this point I think

aubin-tchoi (approved):
LGTM

front/lib/api/assistant/conversation_rendering/pruning.ts
flvndvd: Yes, this was covered in `in the droppable range` but let me make it clearer.

PR #12201, opened 2025-04-23
[sparkle] - fix(ScrollArea): responsiveness

JulesBelveze:
## Description

This PR aims at making the `DropdownMenu` component smart enough to expand based on its content and available height. When available height is smaller than its content it becomes scrollable, leveraging our `ScrollArea` component, ensuring style consistency.

## Risk

Low

## Deploy Plan

- Publish sparkle
- Update front and ditch all manually implemented scrolling behaviour

PR #28928, opened 2026-07-13
fix(frame): remove nested ScrollArea so Share modal recipient list scrolls

smb2268:
## Description

Fixes [#9589](https://github.com/dust-tt/tasks/issues/9589): in the Frame Share modal, the manual email recipient list was not scrollable, so only the first ~4 recipients were visible and their "viewed" status couldn't be checked for the rest.

Root cause: the recipient list was wrapped in its own `<ScrollArea className="max-h-96">`, nested inside `SheetContainer`, which already wraps its children in a `ScrollArea` (see `sparkle/src/components/Sheet.tsx`). Nesting a Radix `ScrollArea` inside another scrollable/overflow-hidden ancestor without an explicit height is a known anti-pattern in this codebase (previously called out and cleaned up in #12201) — the inner scrollbar/viewport sizing doesn't reliably pick up the true content overflow, so the list gets visually clipped with no way to scroll to the remaining rows.

Fix: remove the redundant inner `ScrollArea` around the recipient list and let it render as a plain flex column. Scrolling for the whole sheet body (scope options, invite form, recipient list, files-used section) is now owned by the single outer `ScrollArea` already provided by `SheetContainer`.

## Tests

Verified manually

## Risk

Low. May be better UI to have a nested inner scroll area but complex w our current slideout component so I think this is fine for now

## Deploy Plan

- [ ]  deploy front


frankaloia (approved):
LGTM

PR #28929, opened 2026-07-13
feat(microsoft_drive): target upload folders by id, remove folderPath

smb2268:
Fix https://github.com/dust-tt/tasks/issues/8289
## Description

`microsoft_drive__upload_file` resolved its destination by interpreting the `folderPath` name string. Two customer reports (2026-07-10 and 2026-07-13/15) showed uploads reporting success while the file landed in a newly created duplicate folder named after the document library root (e.g. `"Documents partages"` on French sites): the folder walk resolved the path one way, the final upload re-walked the raw path, and Graph implicitly created the misresolved segments. The earlier 403 symptom fixed in `26438` was the same interpretation problem with an exact name collision.

Rather than hardening the name-interpretation heuristics, this aligns the server with the `google_drive` pattern: folders are addressed by id, never by assembling locale-dependent name paths.

- **`upload_file`**: targets `parentFolderId` directly (`PUT .../items/{id}:/{fileName}:/content`), or the drive root when omitted. The `folderPath` param and its folder walk — path resolution, implicit folder creation, library-root stripping — are removed. Implicit creation on misresolved paths is how uploads silently misfiled.
- **New `get_item_from_url`**: resolves any OneDrive/SharePoint URL (including `/:f:/s/...` sharing links) to its drive item via the Graph shares API, returning `id`, `driveId`, `siteId`, `name`, `type`, and `webUrl`.
- **New `create_folder`**: explicit folder creation under a parent id (or the drive root), returning the new folder's id.
- `search_drive_items` / `upload_file` descriptions updated so agents resolve folder ids by URL, browse, search, or creation.

**Behavior change**: `folderPath` no longer exists on `upload_file`. Tool schemas are model-facing and re-read every run, so no clients are pinned to the old shape; agents whose instructions relied on implicit path creation will compose `get_item_from_url` / `list_drive_items` / `create_folder` → `upload_file` instead.

## Tests
Manually verified improvement in behavior

## Risk

Low

## Deploy Plan

Deploy `front`. No feature flag or migration required.

### DIFF
--- a/front/lib/actions/mcp_internal_actions/__snapshots__/mcp_servers_metadata.test.ts.snap
+++ b/front/lib/actions/mcp_internal_actions/__snapshots__/mcp_servers_metadata.test.ts.snap
@@ -1323,7 +1323,15 @@ exports[`MCP Servers Metadata Snapshot > should have stable tool billing info ac
       "freeUsage": false,
       "toolCostCategory": "advanced",
     },
+    "create_folder": {
+      "freeUsage": false,
+      "toolCostCategory": "advanced",
+    },
     "get_file_content": {
+      "freeUsage": false,
+      "toolCostCategory": "advanced",
+    },
+    "get_item_from_url": {
       "freeUsage": false,
       "toolCostCategory": "advanced",
     },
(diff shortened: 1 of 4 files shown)


frankaloia (approved):
Nice, makes sense

PR #28931, opened 2026-07-13
fix(front): create conversation forks at the parent's depth so they stay listed

smb2268:
## Description

Fixes #9586.

Branched/forked conversations (created via the "Branch" UI action / `POST /assistant/conversations/{cId}/forks`) were disappearing from every conversation list (Pod list, unread activity list) while remaining reachable through their direct URL.

Root cause: `createConversationFork` created forks with `depth = parentConversation.depth + 1`. But `depth` was introduced (`13111`) as the **run_agent recursion counter** — the public API rejects creation at `depth >= 4` with "Recursive run_agent calls exceeded depth" — and every `depth` consumer treats `depth > 0` as "agent-internal":

- `listConversationsInSpacePaginated` and `listSpaceUnreadConversationsAndActivityForUser` hard-filter `depth = 0` (added in `28679` to hide `run_agent` sub-conversations), so forks vanished from Pod lists and the sidebar unread activity.
- The `conversation-unread` and `agent-message-feedback` notification workflows skip `depth > 0` conversations, so forks never produced notifications.
- `getMaxActionsPerStep(depth)` halves per-step action budgets as depth grows (8→8→4→2), so agents in fork-of-fork conversations ran with degraded limits — and a run_agent call inside a depth-3 fork chain would hit the depth-4 cap and hard-fail.
- Forks were exempted from the private-by-default conversation URL restriction (`shouldApplyPrivateByDefaultUrlRestriction`).

An earlier revision of this PR fixed only the two list queries (excluding `depth > 0` conversations unless registered in `conversation_forks`). That addressed the visible symptom but kept forks mislabeled, leaving the notification/action-limit/recursion-cap behaviors above broken and requiring every future `depth` consumer to special-case forks.

This revision fixes the data instead: forks are user-facing conversations, so `createConversationFork` now creates them at the **parent's depth** rather than incrementing. Inheriting (rather than hardcoding `0`) is deliberate — in the edge case where someone branches a run_agent sub-conversation from its direct URL, the fork conservatively keeps the parent's agent-internal status. Fork lineage and nesting remain fully represented in `conversation_forks` (parent/child links, indexed both directions); if we later want a denormalized fork-nesting level for UI, it should be a dedicated field, not the recursion guard.

A backfill migration (`20260714_backfill_fork_conversation_depth.ts`) realigns existing forked conversations by propagating `depth = parent.depth` through the `conversation_forks` lineage. It runs in batches of 1,000 ids to bound row locks on the hot `conversations` table, re-selects after each batch so fork-of-fork chains converge, warns if it exits before fixpoint, and deliberately leaves `updatedAt` untouched so backfilled rows keep their list ordering. Standard dry-run/`--execute` flow.

## Tests

- Updated the `createConversationFork` expectation: a fork's depth now equals the parent's depth.
- Regression test for the reported scenario: forking a conversation in a project space and asserting the fork shows up in `listConversationsInSpacePaginated`.
- Regression test for the unread/activity path: a non-participant project member sees the fork in `listSpaceUnreadConversationsAndActivityForUser` (the sidebar badge path).
- The existing `should exclude sub-conversations (depth > 0)` tests are unmodified, confirming genuine `run_agent` sub-conversations stay hidden.

`npx tsgo --noEmit`, `biome check`, and the `forks.test.ts` suite (16/16) pass locally.

## Risk

Low.

- Creation-side change is one line; no schema change.
- Existing forks stay hidden until the backfill runs (status quo, no regression); the backfill is batched and idempotent (safe to re-run).
- Behavior changes beyond list visibility are intentional: forks now emit unread/feedback notifications, get full action budgets, no longer consume the run_agent recursion budget, and respect the private-by-default URL restriction — all consistent with being user-facing conversations.

## Deploy Plan

- [ ] Deploy front
- [ ] Dry-run `npx tsx front/migrations/20260714_backfill_fork_conversation_depth.ts` (record the mismatched count)
- [ ] Run with `--execute`
- [ ] Confirm affected workspaces' branched conversations reappear in Pod lists


philipperolet (approved):
👍 🙏

philipperolet:
quick nit @smb2268 : 
"Fixes #9586" in the PR description unfortunately does not work as it refers to the dust repo instead of the tasks repo where the task you want to close is (here it referred to a merged PR by aubin)
You can do "Fixes https://github.com/dust-tt/tasks/issues/9586"
Thanks again for taking care of the issue

PR #28936, opened 2026-07-14
Fix inline Frame deep links opening a blank file view

smb2268:
## Description

Fixes `dust-tt/tasks#9596`: clicking a Frame file deep-linked inline in conversation text (via the `:preview_file` markdown directive) opened a blank file view, while the generated-file card and the Frame panel worked fine.

**Root cause** — two gaps compounding:
1. Inline directives render through `FilePreviewBlock` → `PreviewableCitation` with only a scoped **path** — no `fileId` — while the interactive-content side panel requires a `fileId`. So no click-side routing fix alone can work: the identifier the panel needs is simply not available at the render site.
2. The generic preview dialog deliberately has no renderer for the `frame` category (`FilePreviewContent.tsx` returns `null`), because every other surface (file explorer, citation cards, generated-file chips) intercepts Frames and opens the `interactive_content` side panel first. The inline path had no such interception, so clicks fell through to a dialog that renders an empty body.

**What this PR does:**
1. Adds `useInteractiveFileResolution`, which resolves an inline scoped-path reference to its interactive-content `FileResource` at render time, in priority order:
   - the message's own `generatedFiles` (in memory, zero fetch — covers Frames created/linked in the same message, the reported repro);
   - the conversation attachments (covers Frames created by earlier messages, e.g. an inline link in a "here's your edit" follow-up) — exact path match first, then a name fallback, since agent-generated attachments carry `path: null`.
   The attachments lookup is SWR-cached and `disabled` unless the reference could denote a Frame (interactive contentType, or a frame-capable extension when the directive omits `contentType`) and the in-memory match missed.
2. Provides message-scoped lookup data (`owner`, `conversationId`, `generatedFiles`) through a new `FilePreviewLookupContext` in `AgentMessage`, mirroring `CitationsContext` — deliberately not via the plugin-closure route, which would recreate component types and remount every directive node on message updates.
3. Makes interactive content **panel-only** in `PreviewableCitation`: it never routes to the file preview dialog (which cannot render Frames). Resolved references open the side panel exactly like the generated-file cards; unresolved ones render without a click affordance instead of a dead link or an empty dialog.

Outside conversations (agent builder preview, user messages) the context is null and behavior is unchanged. Non-interactive content types are unaffected.

## Tests

- Unit tests for the resolution matchers (title/path matching, interactive-only filtering, most-recently-updated tie-breaking).
- Component tests for the full wiring: frame link resolves via message `generatedFiles` and opens the panel (including with the directive's `contentType` omitted); resolution via conversation attachments when the message generated nothing; unresolved frames rendering without a click affordance; non-interactive files keeping the preview dialog (regression).
- `tsgo --noEmit`, `biome check`, and the existing markdown/attachment/file-explorer suites pass.

Suggested manual check before merge: create a Frame and click the inline link in the same message (side panel opens), then ask for an edit in a follow-up message and click that message's inline link (exercises the attachments fallback).

## Risk

Low. Resolution only activates inside conversation messages for Frame-eligible references; everything else keeps the existing dialog path. Worst case, an inline Frame link renders without a click affordance (the card and auto-open panel still work). Safe to rollback.

## Deploy Plan

Standard deploy; no migrations, flags, or ordering.


flvndvd:
Duplicates of recent work done by @fontanierh? See https://github.com/dust-tt/dust/pull/28894.

smb2268:
> Duplicates of recent work done by @fontanierh? See #28894.

@flvndvd ah nice, this wasn't merged in when I tested the bug yesterday. will make sure it's resolved

smb2268:
@flvndvd I'm now seeing this on prod for the test convo I was using yesterday
before @fontanierh change the primary modal opened but was empty, when tested w this PR it worked as expected (pre-rebase)
Screenshot 2026-07-15 at 1 26 46 PM

fontanierh:
on new conversations it should work 👍

fontanierh:
In fact even on old conversations, can you link the conversation where it's not working ? You might need to refresh the page

smb2268:
Closing, addressed in the linked PR

PR #28937, opened 2026-07-14
fix(front): point offloaded tool-output snippets at the archived file

smb2268:
Fix https://github.com/dust-tt/tasks/issues/9595
## Description

Since `27178` (June 10), tool output blocks offloaded to the conversation file system are replaced with an 8,000-character snippet — but for **resource-shaped** blocks (webbrowser `BROWSE_RESULT`, data-source `DATA_SOURCE_NODE_CONTENT`) the snippet dropped the pointer to the offloaded file: the resource keeps the web URL as `uri` and the scoped path returned by `persistToolOutput` was discarded. Plain-text blocks kept their pointer (`uri` = scoped path), resource blocks did not. Net effect: agents could not read any browsed page beyond 8,000 characters, and the `data_sources_file_system` `cat` pagination footer was silently cut off with no recovery path.

The offload + 8,000-char snippet itself is intended (keeps large content out of the model context; the model reads it back on demand) — this PR only restores the missing pointer:

- Offload snippets (both text blocks and resource blocks) are built by a shared `makeOffloadedSnippet` helper that appends `[Full content archived at {scopedPath}]`, so the model can always read the full content back from the conversation file system.
- The `... (truncated)` marker (referenced by prompts) is only emitted when characters were actually dropped — the offload threshold is in bytes while the snippet cut is in characters, so short/multibyte content could previously claim truncation that never happened.

## Tests

- The two existing snippet tests now assert the archived-file pointer is present in the snippet.
- New tests: a small `BROWSE_RESULT` is offloaded with its full text in the snippet, the archive pointer, and no truncation marker; a large `BROWSE_RESULT` is archived to `.tool_outputs/` as `.md` with a truncated snippet pointing at the archived file.
- `npm test lib/actions/mcp_execution.test.ts`: 15/15 pass. `tsgo --noEmit` and `biome check` clean.
Confirmed fix manually:
Screenshot 2026-07-14 at 1 04 30 PM
Screenshot 2026-07-14 at 1 04 23 PM

## Risk

Low. No change to what gets offloaded or when — only the snippet text the model sees gains the archive pointer (and loses a spurious truncation marker when nothing was dropped). Safe to rollback.

## Deploy Plan

deploy front

### DIFF
--- a/front/lib/actions/mcp_execution.test.ts
+++ b/front/lib/actions/mcp_execution.test.ts
@@ -11,7 +11,10 @@ import {
   getAugmentedInputs,
   processToolResults,
 } from "@app/lib/actions/mcp_execution";
-import type { DataSourceNodeContentType } from "@app/lib/actions/mcp_internal_actions/output_schemas";
+import type {
+  BrowseResultResourceType,
+  DataSourceNodeContentType,
+} from "@app/lib/actions/mcp_internal_actions/output_schemas";
 import type { ToolContext } from "@app/lib/actions/types";
 import { TOOL_OUTPUTS_FOLDER_NAME } from "@app/lib/api/files/mount_path";
 import { Authenticator } from "@app/lib/auth";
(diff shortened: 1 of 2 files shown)


flvndvd (commented):
LGTM

PR #9594, opened 2024-12-21
[Keyword search] Index creation: remove previous alias option

philipperolet:
Description
---
1. add `is_write_index` to alias on index creation (required when alias points to multiple indices
2. provide an arg to remove alias on previous index version

Risks
---
na

Deploy
---
update runbook
core

PR #28938, opened 2026-07-14
fix(front): make pruned-context chip 'Learn more' link clickable

smb2268:
Fix https://github.com/dust-tt/tasks/issues/9594
## Description

The "Context limit reached" chip shown on agent messages when context-window pruning occurs used a Sparkle `Tooltip` for its explanation. Sparkle's `Tooltip` hardcodes Radix's `disableHoverableContent`, so the tooltip closes as soon as the pointer leaves the chip — the "Learn more" link inside could never be reached with the mouse. It was also unreachable by keyboard, since Radix tooltips close on blur and can't be tabbed into (this is the real bug behind issue `#9594`, which was misfiled against `ReachedLimitPopup.tsx`).

This switches `PrunedContextChip` to a click-triggered `Popover`, following the existing chip-with-interactive-content pattern in `TaskDirectiveChipInner` (`front/components/markdown/TaskDirectiveBlock.tsx`): the chip is wrapped in a native `<button>` with an `aria-label` and focus-visible ring as the popover trigger. Content is unchanged; the popover width matches the tooltip's old `max-w-sm` and clamps on narrow viewports. The link is now reachable by both mouse and keyboard (Enter opens, Tab reaches the link, Escape closes).

## Tests

- `tsgo --noEmit` and biome pass on `front`.
- Not exercised against a live pruned conversation (reproducing a context-window prune end-to-end is impractical); the change is presentational and mirrors an established pattern.

## Risk

Low — UI-only change to a single informational chip. Interaction changes from hover to click. Safe to rollback.

## Deploy Plan

Standard front deploy.

### DIFF
--- a/front/components/assistant/conversation/AgentMessage.tsx
+++ b/front/components/assistant/conversation/AgentMessage.tsx
@@ -122,6 +122,9 @@ import {
   InfoCircle,
   InteractiveImageGrid,
   Link01,
+  PopoverContent,
+  PopoverRoot,
+  PopoverTrigger,
   RefreshCw02,
   Stop,
   Tooltip,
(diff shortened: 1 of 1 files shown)


frankaloia (approved):
LGTM

PR #28939, opened 2026-07-14
Fix MCP tab isolation: random server IDs + rate limiting

frankaloia:
## Summary
- https://github.com/dust-tt/tasks/issues/9566

Bug:
 The serverId is passed in the user message context (clientSideMCPServerIds) when
  the message is sent. That exact serverId is baked into the agent configuration for that conversation's run — the
  backend dispatches all tool calls to that specific channel for the lifetime of the run.

  So the actual bug:
  1. Tab A sends a message → serverId = mcp-client-side:sidekick.1 is baked into that conversation run
  2. Tab A's registration expires (or it closes and deregisters)
  3. Tab B opens and claims slot .1 (first free slot)
  4. The agent is still dispatching tool calls for Tab A's conversation to the .1 channel
  5. Tab B receives them — it's on the wrong agent's channel

  It doesn't require explicit close+reopen. It happens whenever any tab loses its registration (TTL expiry, SSE
  drop, crash) while another tab simultaneously claims the same slot.

- **Fix**: Replace the numeric suffix with a random 8-byte hex suffix (`randomBytes(8).toString("hex")`). Each registration gets a globally unique channel, so concurrent tabs are fully isolated.
- **Rate limiting**: The old scheme relied on a 256-slot cap to prevent abuse. With random IDs there is no natural slot limit, so added a 100/min per-user rate limit on both register endpoints instead.
- **SSE recovery**: On SSE stream errors, the transport now heartbeats the existing registration first. If still alive it reconnects the stream; if dead it re-registers (getting a new serverId) and reconnects to the new channel. Previously it blindly reconnected the stream to the old channel even after re-registration.

## Testing
* Added unit tests
Screenshot 2026-07-14 at 3 29 40 PM

## Risk

Some risk in unintended consequences with the client tool logic

## Deploy
1. Deploy front

### DIFF
--- a/front-api/routes/w/[wId]/mcp/register.ts
+++ b/front-api/routes/w/[wId]/mcp/register.ts
@@ -1,7 +1,6 @@
-import {
-  MCPServerInstanceLimitError,
-  registerMCPServer,
-} from "@app/lib/api/actions/mcp/client_side_registry";
+import { registerMCPServer } from "@app/lib/api/actions/mcp/client_side_registry";
+import { rateLimiter } from "@app/lib/utils/rate_limiter";
+import logger from "@app/logger/logger";
 import { workspaceApp } from "@front-api/middlewares/ctx";
 import { apiError } from "@front-api/middlewares/utils";
 import { validate } from "@front-api/middlewares/validator";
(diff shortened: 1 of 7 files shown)


front/lib/api/actions/mcp/client_side_registry.test.ts
```diff
@@ -0,0 +1,96 @@
+import {
+  deregisterMCPServer,
+  getBaseServerId,
+  registerMCPServer,
+} from "@app/lib/api/actions/mcp/client_side_registry";
+import { Authenticator } from "@app/lib/auth";
+import { MembershipFactory } from "@app/tests/utils/MembershipFactory";
+import { UserFactory } from "@app/tests/utils/UserFactory";
+import { WorkspaceFactory } from "@app/tests/utils/WorkspaceFactory";
+import { beforeEach, describe, expect, it } from "vitest";
```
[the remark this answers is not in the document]
frankaloia: Will add a unit test for that endpoint, but this unit tests still serves a separate purpose

front-api/routes/w/[wId]/mcp/register.ts
```diff
@@ -36,22 +35,35 @@ app.post(
     const auth = ctx.get("auth");
     const { serverName } = ctx.req.valid("json");
 
+    const userId = auth.getNonNullableUser().id;
+    const remaining = await rateLimiter({
+      key: `mcp:register:${userId}`,
+      maxPerTimeframe: 100,
```
Comment (flvndvd) [inline front-api/routes/w/[wId]/mcp/register.ts]: `nit` but can we share some of those constants between private and public APIs?

front/lib/client/BrowserMCPTransport.ts
```diff
@@ -129,6 +129,16 @@ export class BrowserMCPTransport implements Transport {
       // Setup heartbeat to keep the server registration alive.
       this.setupHeartbeat(data.serverId);
 
+      // If an SSE stream was already opened (re-registration after a lost
+      // registration), it is still attached to the previous serverId's channel
+      // and would keep receiving requests that no longer
+      // belong to this transport. Reconnect to the new serverId's channel. The
+      // lastEventId belongs to the old channel's stream, so drop it.
+      if (this.eventSource) {
```
flvndvd, line 137: This fix should also apply to the sdk's implementation, no?

flvndvd (approved):

PR #28941, opened 2026-07-14
feat(activation): add --execute mode to orchestrator CLI

frankaloia:
## Summary

- Adds `--execute` flag to the activation orchestrator CLI (`front/temporal/activation/admin/cli.ts`) that fires real activation webhook events instead of just dry-running
- Moves `emitActivationEvent` into `orchestrator.ts` alongside the eligibility logic so the CLI is self-contained on main (no dependency on activation trigger provisioning work)

## Testing

```bash
# Dry run (default — prints eligible users, sends nothing)
npx tsx front/temporal/activation/admin/cli.ts run --workspace <sId>

# Execute — fires one activation event per eligible pod
npx tsx front/temporal/activation/admin/cli.ts run --workspace <sId> --execute

# Restrict to a single user
npx tsx front/temporal/activation/admin/cli.ts run --workspace <sId> --execute --user <userSId>
```

## Risk

Low — admin-only CLI script, not wired into any production flow. `emitActivationEvent` requires the workspace to already have an "Activation" webhook source configured; otherwise it returns an Err and logs it without crashing.

### DIFF
--- a/front/temporal/activation/admin/cli.ts
+++ b/front/temporal/activation/admin/cli.ts
@@ -1,4 +1,7 @@
 import { determineEligibleActivationUsers } from "@app/lib/api/activation/orchestrator";
+import { emitActivationEvent } from "@app/lib/api/activation/trigger";
+import { Authenticator } from "@app/lib/auth";
+import { SpaceResource } from "@app/lib/resources/space_resource";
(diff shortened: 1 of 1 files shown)


ElPicador (approved):
👍

PR #28944, opened 2026-07-15
[front] chore: Make GDrive GC production check more reliable

id13:
## Description

Currently, Google Drive production check about garbage dead documents (not garbage collected) reports false positives, i.e rows that have not been yet GCd but will soon as the GC workflow is running.

This new approach instead fires a failed production checks if both condition are met:
- % of dead documents exceeds 5% of data source
- the GC workflow is not running

If the GC workflow for a given datasource runs often enough and long enough, we might have false positives. 
Given current frequencies, I'm not too concerned.

## Tests

N/A

## Risk

Low
Blast radius: some false positives

## Deploy Plan

Deploy front


id13:
PMRR

philipperolet (commented):
LGTM

PR #28945, opened 2026-07-15
[front] fix: show slash skills before tools finish loading

aubin-tchoi:
## Description

The input bar slash menu kept its capability loading state until both skills and MCP tools finished loading, even though one of them will be available sooner (much sooner, on our workspace we have skills after 450ms vs both after 4.5s).

`useInputBarSlashCommandCapabilities` now clears the loading state as soon as a matching capability is available. When no capabilities are available, it keeps loading until tools settle, avoiding a transient empty state before tool results arrive.

## Tests

- Tested locally.

## Risk

- Low.

## Deploy Plan

- Deploy front.


PopDaph (approved):
Could be slightly dangerous if agents start to use this pattern but pragmatic for this use case! LGTM

aubin-tchoi:
agreed, adding a small comment!

PR #28946, opened 2026-07-15
Route file HEAD requests through GET handler

fontanierh:
## Description

Hono 4.12 dispatches HEAD requests through GET routing while preserving the original request method. As a result, the existing `app.on("HEAD", ...)` route is never selected and HEAD requests execute the full GET content path.

see https://hono.dev/docs/guides/best-practices#head-request-best-practices

Move the existing metadata-only behavior into a helper reached early from the GET handler. This does not change the endpoint contract.

Stack: 1 of 4. Next: #28304.

## Tests

The endpoint suite now verifies that HEAD returns metadata without opening a content stream.

## Risk

Low. Response status and headers are unchanged. The implementation avoids unnecessary file reads for HEAD requests.

## Deploy Plan

Standard deploy.

### DIFF
--- a/front-api/routes/w/[wId]/files/path/[...canonicalPath].test.ts
+++ b/front-api/routes/w/[wId]/files/path/[...canonicalPath].test.ts
@@ -243,12 +243,14 @@ describe("HEAD /api/w/:wId/files/path/:canonicalPath", () => {
 
   it("returns metadata headers without a body", async () => {
     const { workspace, conversation } = await setup();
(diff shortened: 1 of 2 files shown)


flvndvd (approved):
🤯 Confusing Hono, OOC how did you find out?

fontanierh:
@flvndvd I am working on allowing to open the frame from the markdown directive and needed to do some stuff with the HEAD endpoint, at which point I realized it was not doing what I thought

PR #28950, opened 2026-07-15
Run sandbox function invocations in Temporal

spolu:
## Description

Follow-up to #28870.

- Launch sandbox function invocations in a dedicated Temporal workflow after creating and publishing the invocation.
- Execute the existing invocation in a non-retrying activity registered on the agent-loop worker.
- Mark the invocation errored and publish its terminal event when workflow launch or execution fails.

A sandbox function can block while a tool waits for approval, but the visualization cannot subscribe to its event stream until the invocation POST returns. Moving the outer execution to Temporal returns the invocation ID first, allowing approval and authentication events to reach the UI while execution is blocked.

## Tests

- Focused front and front-api tests (26 tests)
- Front TypeScript check
- Temporal workflow bundle build (27 workers)
- Biome formatting/checks

## RIsks

- None (not live product)

## Deploy

- deploy `front`

### DIFF
--- a/front/temporal/agent_loop/activities/run_sandbox_function_invocation.ts
+++ b/front/temporal/agent_loop/activities/run_sandbox_function_invocation.ts
@@ -1,41 +1,25 @@
 import type { AuthenticatorType } from "@app/lib/auth";
 import { Authenticator } from "@app/lib/auth";
-import { SandboxFunctionInvocationResource } from "@app/lib/resources/sandbox_function_invocation_resource";
 import { SandboxFunctionResource } from "@app/lib/resources/sandbox_function_resource";
-import type { PostSandboxFunctionInvocationRequestBody } from "@app/types/api/sandbox_functions";
 
 export type RunSandboxFunctionInvocationActivityArgs = {
-  sandboxFunctionId: string;
   invocationId: string;
-  body: PostSandboxFunctionInvocationRequestBody;
 };
 
 export async function runSandboxFunctionInvocationActivity(
   authType: AuthenticatorType,
-  {
-    sandboxFunctionId,
-    invocationId,
-    body,
-  }: RunSandboxFunctionInvocationActivityArgs
+  { invocationId }: RunSandboxFunctionInvocationActivityArgs
 ): Promise<void> {
   const auth = await Authenticator.fromJsonWithRefrehedGroups(authType);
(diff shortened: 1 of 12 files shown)


front/temporal/agent_loop/activities/run_sandbox_function_invocation.ts
```diff
@@ -0,0 +1,38 @@
...
+
+  const invocation = await SandboxFunctionInvocationResource.fetchById(auth, {
+    sandboxFunction,
+    invocationId,
+  });
+  if (!invocation) {
+    throw new Error(`Sandbox function invocation not found: ${invocationId}`);
+  }
+
+  const executionResult = await invocation.execute(auth);
```
PopDaph, line 33: This await blocks for the whole function run (execute → sandbox.exec, up to SANDBOX_FUNCTION_EXEC_TIMEOUT_MS = 2min), and the activity runs on agent-loop-queue-v2, so it holds one of the worker's 75 shared activity slots the entire time, alongside the live agent loop (run_tool / run_model).

Should we move the sandbox activities (this + runSandboxFunctionToolActivity, which blocks too) onto a dedicated queue + worker so the blast radius stays inside sandbox functions. WDYT?

front/temporal/agent_loop/client.ts
```diff
@@ -296,3 +301,51 @@ export async function launchSandboxFunctionToolWorkflow(
...
+        sandboxFunctionId: sandboxFunction.sId,
+        invocationId: invocation.sId,
+        workspaceId,
+      },
+    });
+  } catch (error) {
+    if (error instanceof WorkflowExecutionAlreadyStartedError) {
+      return new Ok(undefined);
+    }
+    return new Err(normalizeError(error));
```
PopDaph, line 347: 👍🏻 cleaner than what I done on `launchSandboxFunctionToolWorkflow` 🙈

PopDaph (approved):

front/temporal/agent_loop/activities/run_sandbox_function_invocation.ts
spolu: Yep that was something I was about to suggest us doing indeed 👍

spolu: I will move forward with this and work on a new queue as next step 👍

PR #28951, opened 2026-07-15
Instrument conversation pruning and prompt-cache usage for Datadog

flvndvd:
## Description

The conversation pruning system now makes cache-relevant decisions on every render, batched pruning of old tool results and batched dropping of old interactions, and until now we had no way to see how often each mechanism fires in production, how much it reclaims, or whether the prompt cache actually benefits. This adds the metrics a dashboard needs to answer those questions and to tune the system's constants against real traffic.

Two families of series are emitted:
- On the decision side, each agent-loop render reports its outcome (nothing needed, pruned, dropped, or the forced last-resort variants), how many tokens each layer reclaimed, the headroom a batched drop bought before the next full cache miss becomes necessary, context utilization, and a saturation flag marking conversations where pruning has exhausted everything it may touch
- On the outcome side, every model call reports its input tokens split into cache hits, cache writes, and uncached, which yields the cache hit ratio, the ground truth the whole pruning effort answers to, and separates streaming from batch traffic since batch never benefits from a warm cache.

The design keeps instrumentation out of the decision logic.

### DIFF
--- a/front/lib/api/assistant/conversation_rendering/instrumentation.test.ts
+++ b/front/lib/api/assistant/conversation_rendering/instrumentation.test.ts
@@ -1,0 +1,109 @@
+import type { ConversationPruningStats } from "@app/lib/api/assistant/conversation_rendering/instrumentation";
+import { computeConversationRenderingMetrics } from "@app/lib/api/assistant/conversation_rendering/instrumentation";
+import { describe, expect, it } from "vitest";
+
+// A render where nothing happened: every snapshot equals the starting total.
+function statsWhereNothingHappened(): ConversationPruningStats {
+  return {
+    totalTokensBefore: 50_000,
+    totalTokensAfterPruning: 50_000,
+    totalTokensAfterDropping: 50_000,
+    totalTokensAfterFloorPruning: 50_000,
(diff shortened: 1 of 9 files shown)


fontanierh (approved):
LGTM

front/lib/api/assistant/conversation_rendering/instrumentation.test.ts
```diff
@@ -0,0 +1,109 @@
...
+    interactionsBefore: 12,
+    interactionsAfterDropping: 12,
+    interactionsAfterFloorDropping: 12,
+    pruningBudget: 100_000,
+    budgetForInteractions: 160_000,
+  };
+}
+
+describe("computeConversationRenderingMetrics", () => {
+  it("reports a clean fit when no layer changed anything", () => {
```
aubin-tchoi, line 22: we're testing the instrumentation?

aubin-tchoi (approved):
LGTM

front/lib/api/assistant/conversation_rendering/instrumentation.test.ts
flvndvd: Ceinture bretelle

PR #28953, opened 2026-07-15
Fix pagination

tdraier:
## Description

Fixes `listMetronomeSeatBalances` returning incomplete results on contracts with a few hundred+ seats, which caused two visible bugs:
- poke's Usage tab showing `- / 8,000` (missing seat balance) for many users
- the credit-state reconcile job defaulting those same users to `on_pool` even though they had plenty of seat balance remaining

**Root cause**: the function queried Metronome's `/v1/contracts/seatBalances/list` unfiltered ("list all seats on this contract"). That unfiltered mode silently omits most seats on large contracts — pagination itself terminates correctly (`next_page` legitimately goes `null`), the underlying list Metronome returns is just incomplete. Querying the same endpoint with an explicit `seat_ids` filter reliably returns the seat.

Confirmed live on a ~910-seat production contract: the unfiltered list returned 536 seats; the same seats queried via `seat_ids` in one pass came back essentially complete.

**Fix**: `listMetronomeSeatBalances` now takes `seatIds?: string[]` (chunked in batches of 100) instead of relying on the unfiltered list, plus two robustness fixes found while validating against production data:
- **Chunk-level resilience**: Metronome rejects an entire `seat_ids` batch with a 400 if even one id in it isn't found on the contract (e.g. a removed/reassigned seat). A chunk failure now falls back to querying that chunk's ids individually instead of losing every seat in the batch.
- **Retry on silent omission**: separately from the above, Metronome was observed to non-deterministically omit a valid, balance-holding seat from an otherwise-successful (200) response — reproducible with a single user across back-to-back identical calls, no error at all. This is now retried (up to 2x, 300ms apart) before a seat is reported as missing, and is tracked separately from a confirmed absence (an explicit 400 "not found in contract subscriptions", which is not retried).

Updated every call site to pass explicit seat ids instead of relying on the unfiltered list:
- `lib/api/credits/members_usage.ts` — poke Usage tab
- `lib/api/metronome/reconcile_credit_state.ts` — the global credit-state reconcile (reordered to fetch memberships before the seat-balance call, since the ids are now needed up front)
- `lib/metronome/seats.ts` (×2) — seat credit transfer logic (upgrade/downgrade between seat types), which had the same unfiltered-list bug and could silently miscompute credit carryover during a seat-type transfer
- `lib/metronome/live_user_credit_inputs.ts`, `scripts/debug_user_awu_spend.ts` — already single-seat callers, just renamed the param

Also adds `scripts/debug_seat_balances.ts`, a read-only diagnostic script to cross-reference a workspace's active pro/max members against Metronome's live seat balances and report exactly who's missing one.

## Tests

- Existing unit tests for all touched files pass (`client.test.ts`, `seats.test.ts`, `reconcile_credit_state.test.ts`, `live_user_credit_inputs.test.ts` — 87 tests).
- Manually verified against a live ~910-seat production contract via `scripts/debug_seat_balances.ts`:
  - Before: unfiltered list found 536/910 seats.
  - After (filtered, no retry yet): found 878-883/910, non-deterministically, matching the "silent omission" theory.
  - After (filtered + retry): found 906/910, with the remaining 4 all confirmed via explicit "not found in contract subscriptions" errors (a real, separate data gap — those 4 users' seats genuinely aren't registered on the contract subscription; needs a seat-count sync, not a code fix).

## Risk

- Behavioral change to a widely-used Metronome read path (`listMetronomeSeatBalances`), but the change only makes it more correct (was silently dropping data before); no callers relied on the old unfiltered behavior on purpose.
- Slightly more Metronome API calls per invocation (chunked + occasional retries) instead of one unfiltered call, but bounded (100 ids/chunk, max 2 retries per unresolved id) and only runs where seat ids are already known.
- Safe to roll back: reverts to the previous (buggy but previously-shipped) behavior.

## Deploy Plan

Standard deploy. No migration, no config change, no feature flag. Once live, worth re-running `scripts/debug_seat_balances.ts --workspaceId <wId>` against the previously-affected workspace to confirm the fix, then using the "Sync Metronome Seat Count" poke plugin to resolve any workspace's remaining confirmed-absent seats (a separate, genuine data-sync action, not part of this PR).

### DIFF
--- a/front/lib/api/credits/members_usage.ts
+++ b/front/lib/api/credits/members_usage.ts
@@ -458,20 +458,25 @@ async function fetchSeatDataForMembersTable({
 // Live per-seat AWU balance remaining for paid (seat-managed) seats, keyed by
 // userId. This is the expensive read (`listMetronomeSeatBalances`) gated to poke
 // — free seats are handled separately by `fetchFreeSeatCreditsForMembersTable`.
+// Always queried by explicit `seatIds`: Metronome's unfiltered seat-balances
+// list silently omits most seats on contracts with a few hundred+ seats.
 // Degrades to an empty map on any read failure so the table still renders.
 async function fetchSeatBalancesForMembersTable({
   metronomeCustomerId,
   metronomeContractId,
+  userIds,
 }: {
   metronomeCustomerId: string | null;
   metronomeContractId: string | null;
+  userIds: string[];
 }): Promise<Map<string, number>> {
-  if (!metronomeCustomerId || !metronomeContractId) {
+  if (!metronomeCustomerId || !metronomeContractId || userIds.length === 0) {
     return new Map();
   }
   const result = await listMetronomeSeatBalances({
     metronomeCustomerId,
     metronomeContractId,
+    seatIds: userIds,
   });
   const balanceByUserId = new Map<string, number>();
   if (result.isErr()) {
@@ -1323,6 +1328,7 @@ export async function getMembersUsage({
       ? fetchSeatBalancesForMembersTable({
           metronomeCustomerId: metronomeCustomerId ?? null,
           metronomeContractId,
+          userIds: users.map((u) => u.sId),
         })
       : Promise.resolve(new Map<string, number>()),
     // Free-seat per-user credit balance + granted total. Needed on every surface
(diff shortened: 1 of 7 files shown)


tdraier:

- #28953 ◀

PR #28955, opened 2026-07-15
[connectors] fix: Make Notion GC more respecful of temporal's limits

id13:
## Description

Big Notion GC workflows can cross the 51k events limit after which temporal terminates it.
This is problematic because in such case, it need to be caught by oncall eng (we have a production check but it can and is missed) AND restarted manually (from temporal)

This PR leverages the `workflowInfo().continueAsNewSuggested` introduced by temporal and proven to work (I used it in some temporal workflows) - to limit the unbounded growth of `garbageCollectBatch`.

## Tests

N/A

## Risk

Low
Blast radius: Notion connector :/

## Deploy Plan

- Deploy connectors
- In 3 days, deprecate the patch
- In one week, remove it and the former activity


fontanierh (approved):
LGTM

PR #28957, opened 2026-07-15
Persist sandbox function invocation data in GCS

spolu:
## Description

Persist sandbox function invocation inputs, results, and errors in GCS and store the object path on the invocation.

Invocation data is loaded with the resource, updated before success or failure events are emitted, and deleted with the invocation. This provides durable inputs and outputs for the upcoming Temporal execution workflow without passing invocation payloads through Temporal.

## Tests

- 14 sandbox function resource tests.
- 20 sandbox function API and SSE endpoint tests.
- Tested locally
```
dust_front=# SELECT * FROM sandbox_function_invocations;
         createdAt          |         updatedAt          | sandboxFunctionId |  status   | workspaceId | id |                                 gcsPath                                  
----------------------------+----------------------------+-------------------+-----------+-------------+----+--------------------------------------------------------------------------
 2026-07-15 11:10:13.898+00 | 2026-07-15 11:10:15.841+00 |                 2 | succeeded |           1 |  1 | w/DevWkSpace/sandbox_functions/sfn_M3noynRxIM/invocations/sfi_Q8Ac4uf35L
(1 row)
```

## Risk

None, not a live product

## Deploy Plan

- Run pre-deploy migration `20260715111106_add_gcs_path_to_sandbox_function_invocations.sql`.
- Deploy `front` and `front-api`.
- Backfill the production workspace with:
  `cd front && npx tsx migrations/20260715_backfill_sandbox_function_invocation_gcs_paths.ts --workspaceId <workspace-sId> --execute`.

### DIFF
--- a/front/lib/resources/sandbox_function_invocation_resource.ts
+++ b/front/lib/resources/sandbox_function_invocation_resource.ts
@@ -11,6 +11,7 @@ import { ensurePodSandboxReady } from "@app/lib/api/sandbox/lifecycle";
 import { shellEscape } from "@app/lib/api/sandbox/shell";
 import { publishSandboxFunctionInvocationEvent } from "@app/lib/api/sandbox_functions/events";
 import type { Authenticator } from "@app/lib/auth";
+import { getPrivateUploadBucket } from "@app/lib/file_storage";
 import { BaseResource } from "@app/lib/resources/base_resource";
 import { SandboxFunctionMCPActionResource } from "@app/lib/resources/sandbox_function_mcp_action_resource";
 import type { SandboxFunctionResource } from "@app/lib/resources/sandbox_function_resource";
@@ -23,6 +24,7 @@ import {
   makeSId,
 } from "@app/lib/resources/string_ids";
 import type { ResourceFindOptions } from "@app/lib/resources/types";
+import { concurrentExecutor, withRetry } from "@app/lib/utils/async_utils";
 import logger from "@app/logger/logger";
 import type {
   PostSandboxFunctionInvocationRequestBody,
@@ -34,6 +36,7 @@ import { Err, Ok, type Result } from "@app/types/shared/result";
 import { normalizeError } from "@app/types/shared/utils/error_utils";
 import { truncate } from "@app/types/shared/utils/string_utils";
 import type { Attributes, Transaction } from "sequelize";
+import { Op } from "sequelize";
 
 const SANDBOX_FUNCTION_WORKING_DIRECTORY = "/home/agent";
 const SANDBOX_FUNCTION_EXEC_TIMEOUT_MS = 2 * 60 * 1000;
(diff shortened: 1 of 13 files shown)


front/lib/resources/storage/models/sandbox_function.ts
```diff
@@ -174,6 +178,11 @@ SandboxFunctionInvocationModel.init(
         isIn: [SANDBOX_FUNCTION_INVOCATION_STATUSES],
       },
     },
+    gcsPath: {
+      type: DANGEROUSLY_UNBOUNDED_TEXT,
+      allowNull: true,
```
spolu, line 183: Will allowNull: false after the migration is run

front/lib/resources/sandbox_function_invocation_resource.ts
```diff
@@ -42,6 +45,42 @@ const DSBX_BIN_PATH = "/opt/bin/dsbx";
 // a larger one for the log fields.
 const SANDBOX_FUNCTION_ERROR_DETAIL_MAX_CHARS = 2_048;
 const SANDBOX_FUNCTION_ERROR_LOG_MAX_CHARS = 16_384;
+const GCS_CONCURRENCY = 4;
+
+type SandboxFunctionInvocationData = {
```
flvndvd: We usually go with zod here 🙏. So we can also version if needed.

front/lib/resources/sandbox_function_invocation_resource.ts
```diff
@@ -224,12 +289,72 @@ export class SandboxFunctionInvocationResource extends BaseResource<SandboxFunct
     return makeSId("sandbox_function_invocation", { id, workspaceId });
   }
 
+  private static async loadDataFromGcs(
```
flvndvd: IMHO all of those should be class methods. Also shouldn't we apply the same logic that we have mcp output items, where it's also saved in Redis to avoid hitting GCS API every time?

front/lib/resources/sandbox_function_invocation_resource.ts
```diff
@@ -335,19 +478,32 @@ export class SandboxFunctionInvocationResource extends BaseResource<SandboxFunct
     sandboxFunction: SandboxFunctionResource,
     { transaction }: { transaction?: Transaction } = {}
   ): Promise<number> {
+    const where = {
+      sandboxFunctionId: sandboxFunction.id,
+      workspaceId: sandboxFunction.workspaceId,
+    };
+    const invocations = await this.model.findAll({
```
flvndvd, line 496: Not needed in this PR but let's stop the big delete for and instead do pagination, we had many DB issues lately because of this.

front/lib/resources/sandbox_function_resource.ts
```diff
@@ -260,11 +260,13 @@ export class SandboxFunctionResource extends BaseResource<SandboxFunctionModel>
...
-    return new SandboxFunctionInvocationResource(
-      SandboxFunctionInvocationModel,
-      invocation.get(),
-      { sandboxFunction }
-    );
+    return SandboxFunctionInvocationResource.fetchById(auth, {
+      sandboxFunction,
+      invocationId: SandboxFunctionInvocationResource.modelIdToSId({
+        id: invocation.id,
+        workspaceId: invocation.workspaceId,
```
flvndvd, line 267: This feels very misleading to forge a `sId` for a function that we are about to fetch 🤔. Can't we expose a `fetchYByModelId`?

flvndvd (approved):

front/lib/resources/sandbox_function_invocation_resource.ts
spolu: Will leave optimization to later 👍 
What is the rationale for class method?

front/lib/resources/sandbox_function_resource.ts
spolu: Not quite. I had the exact same reflex. The problem is that you need the invocation to have the sandboxFunctionId so that you can load it so that you can then fetchById

We could have a fetchByModelId but that adds a fair amount of code. I consider here that SandboxFunctionResource is close enough to SandboxFunctionInocationResource to have the "right" to call modelIdToSid for it.

front/lib/resources/sandbox_function_invocation_resource.ts
spolu: Will tackle in a subsequent PR 👍

front/lib/resources/sandbox_function_invocation_resource.ts
spolu: addressing

front/lib/resources/sandbox_function_invocation_resource.ts
```diff
@@ -42,6 +46,41 @@ const DSBX_BIN_PATH = "/opt/bin/dsbx";
 // a larger one for the log fields.
 const SANDBOX_FUNCTION_ERROR_DETAIL_MAX_CHARS = 2_048;
 const SANDBOX_FUNCTION_ERROR_LOG_MAX_CHARS = 16_384;
+const GCS_CONCURRENCY = 4;
+const SANDBOX_FUNCTION_INVOCATION_DATA_VERSION = 1;
+
+const SandboxFunctionInvocationDataSchema = z.object({
+  version: z.literal(SANDBOX_FUNCTION_INVOCATION_DATA_VERSION),
```
spolu, line 53: @flvndvd r? PTAL good with that approach of versioning?

spolu: I haven't seen much prior art

PR #28958, opened 2026-07-15
[front] fix: prevent user answer submission blink + add animation

aubin-tchoi:
## Description

User answers could briefly show the question options again between the answer request completing and the agent retry finishing.

Keep the submitted state active through the full retry lifecycle, preserve the card layout while the options transition to progress, then animate the card out before removing it. Reduced-motion preferences skip the transitions.

Before
Jul-15-2026 2-10-16 PM

After
Jul-15-2026 2-08-55 PM

## Tests

- Updated component tests to cover the retry gap and reduced-motion completion.

## Risk

- Low. Affects the ask user question response UI.

## Deploy Plan

- Deploy front.

### DIFF
--- a/front/components/assistant/conversation/UserAnswerRequired.tsx
+++ b/front/components/assistant/conversation/UserAnswerRequired.tsx
@@ -7,7 +7,7 @@ import { canCurrentUserRespondToParentUserMessage } from "@app/lib/api/assistant
 import { useAuth } from "@app/lib/auth/AuthContext";
 import type { LightWorkspaceType, UserType } from "@app/types/user";
 import { ArrowUp, Button, cn, OptionCard, Spinner } from "@dust-tt/sparkle";
-import { AnimatePresence, motion, useReducedMotion } from "framer-motion";
+import { useReducedMotion } from "framer-motion";
 import type { KeyboardEvent } from "react";
 import { useEffect, useRef, useState } from "react";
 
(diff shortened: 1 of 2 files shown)


front/components/assistant/conversation/UserAnswerRequired.tsx
```diff
@@ -16,6 +16,7 @@ import {
   OptionCard,
   Spinner,
 } from "@dust-tt/sparkle";
+import { AnimatePresence, motion, useReducedMotion } from "framer-motion";
```
aubin-tchoi: consistent with other imports

front/components/assistant/conversation/UserAnswerRequired.tsx
```diff
@@ -90,7 +94,8 @@ export function UserAnswerRequired({
     answer: UserQuestionAnswer,
     { isSkip = false }: { isSkip?: boolean } = {}
   ) {
-    setIsSkipPending(isSkip);
+    containerRef.current?.focus({ preventScroll: true });
```
ykmsd, line 89: I'm not familiar with this code path, why do we need to focus upon submission? 🤔

front/components/assistant/conversation/UserAnswerRequired.tsx
```diff
@@ -104,11 +109,23 @@ export function UserAnswerRequired({
...
+        await retryHandler();
+        if (shouldReduceMotion) {
+          removeCompletedAction(blockedAction.actionId);
+        } else {
+          shouldWaitForExitAnimation = true;
+          setIsCompleted(true);
+        }
+      } finally {
+        if (!shouldWaitForExitAnimation) {
+          setPendingSubmission(null);
```
ykmsd: I think I'm a bit lost here, so we set it to null only when shouldWaitForExitAnimation == false? when we will reset it to `null`  if it's true? 🤔

front/components/assistant/conversation/UserAnswerRequired.tsx
```diff
@@ -50,19 +51,21 @@ export function UserAnswerRequired({
...
+  const { answerQuestion, errorMessage } = useAnswerUserQuestion({ owner });
 
   const answerDraft = useUserAnswerDraft({
     multiSelect: blockedAction.question.multiSelect,
   });
-  const [isSkipPending, setIsSkipPending] = useState(false);
+  const [pendingSubmission, setPendingSubmission] = useState<
+    "answer" | "skip" | null
+  >(null);
+  const [isCompleted, setIsCompleted] = useState(false);
```
ykmsd: am I correct that we skip to sync this state entirely if shouldReduceMotion is true? I cannot come up with anything better so it's okay as it is but I find it a bit confusing

ykmsd (approved):

front/components/assistant/conversation/UserAnswerRequired.tsx
aubin-tchoi: the controls will become disabled and the form will be aria-hidden, the focus here was to avoid the focus from being on the disabled/hidden content but need to double check what happens in this case (do we fallback to focusing the document?)

front/components/assistant/conversation/UserAnswerRequired.tsx
aubin-tchoi: it's a bit convoluted but we dont need it: `isCompleted` starts the exit animation; then the blocked action is removed and we unmount

aubin-tchoi: trying to simplify

front/components/assistant/conversation/UserAnswerRequired.tsx
aubin-tchoi: yes, correct! we skip quite a lot if shouldReduceMotion is true. tried to put the animation logic in a hook but it got more complex so 🤷🏻

front/components/assistant/conversation/UserAnswerRequired.tsx
aubin-tchoi: ok I think it's better now

PR #28959, opened 2026-07-15
[front] fix: protect equipped skills from context pruning

aubin-tchoi:
## Description

Equipped skills are rendered as part of a leading user message. Our pruning currently sometimes drops it, it's treated as one interaction like any other, leaving the model without the list of skills available to enable.

Keep leading messages outside the prunable interaction history while still counting their tokens against both the hard context ceiling and proactive pruning target. Had considered using `role: "user", name: "system"` as a criterion for not dropping a message, but it's also used by enabled-skill instructions later in the conversation so cannot be used like this (these are dropped at compaction).

## Tests

- Updated tests. Tested locally.

## Risk

- Low.

## Deploy Plan

- Deploy front.

### DIFF
--- a/front/lib/api/assistant/conversation_rendering/index.test.ts
+++ b/front/lib/api/assistant/conversation_rendering/index.test.ts
@@ -47,10 +47,7 @@ function createConversation() {
   } as any;
 }
 
-function userMessage(
-  text: string,
-  name = "user"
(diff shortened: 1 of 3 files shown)


aubin-tchoi:
Not needed anymore since we don't prune user messages (we used to do this when we dropped entire interactions)

PR #28972, opened 2026-07-15
Allow switch to free seat from poke

tdraier:
## Description

Allow switch to free seat from poke.
For users who were incorrectly asisgned to a paid seat before. This won't recreate a new credit if they already had one (so they may be blocked if their free credit is already exhausted)

## Risk

Can create some inconsistencies in their credit spend alert.

## Deploy Plan

deploy plan


tdraier:

- #28972 ◀

ElPicador (approved):

PR #28973, opened 2026-07-15
[extension] fix pod conversation stuck loading and client-side MCP tools

matteotrab:
## Description

This PR fixes two bugs of the chrome extension when creating a conversation from a pod's input bar: the conversation page was stuck loading, and the conversation had no client-side MCP tools.

Root cause: the input-bar context (holding the deferred first message) and the client-side MCP server (`useMcpServer()`) were mounted per page. Navigating from the pod route to the conversation route unmounted them.

Fix: hoist both into `ProtectedRoute`, a layout route that survives the swap between the two pages. `ExtensionInputBarProvider` now mounts once and reads the conversation id via `useActiveConversationId()`; a new `ExtensionClientSideMCPServerProvider` owns the single `useMcpServer()` call and exposes the ids through an extension-only context.

Moreover `PodPageContent` takes an optional `clientSideMCPServerIds?: string[]` prop (same convention as `ConversationContainerVirtuoso`), supplied by the extension. The web app passes nothing, so its behavior is unchanged.

## Tests
Manually.

## Risk

Low, scoped to the extension. Moves provider mount points rather than changing logic; the only `front` change is one optional prop defaulting to today's web behavior.

## Deploy Plan

Standard deployment.

### DIFF
--- a/extension/ui/components/auth/ProtectedRoute.tsx
+++ b/extension/ui/components/auth/ProtectedRoute.tsx
@@ -2,6 +2,8 @@ import { cn, Spinner } from "@dust-tt/sparkle";
 import { usePlatform } from "@extension/shared/context/PlatformContext";
 import type { RouteChangeMesssage } from "@extension/shared/messages";
 import { useExtensionAuth } from "@extension/ui/components/auth/AuthProvider";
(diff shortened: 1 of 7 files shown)


matteotrab:

- #28973 ◀

Nils-Fedrigo (approved):

PR #28975, opened 2026-07-15
[front] fix: Support utf8 in api key name & user email

id13:
## Description

Internal calls (run_agent & co) forward the caller's API key name and user email as headers. fetch throws on header values with chars above 0xFF, so a key named Prod 🔑 or an email containing иван broke every sub-agent call.
Most people don't use utf8 in their email address, because of legacy, but in key names it's already the case, and it will break in any call involving those. So I'm fixing both in one pass.

Values that don't fit in Latin-1 are now sent as RFC 2047 encoded-words (=?utf-8?B?…?=) and decoded in the front-api auth middleware — the only place that reads these headers. Everything Latin-1-safe stays raw, so the wire is unchanged for existing values and no deploy ordering is needed (encoded values were unsendable before).

Notes:
- x-api-user-email drives system-key impersonation (exact-string lookup), so lossy sanitization wasn't an option.
- x-dust-api-key-name is attribution-only; billing resolves key names from the DB via userContextApiKeyId, never from the header.
- No Node built-in for this;
- Checked that the encoded form isn't logged anywhere before decode (request logger, error handlers, statsd, SDK).

## Tests

Unit tests

## Risk

Low
Blast radius - at least all run_agent calls, but probably all api call with impersonation 😬 

## Deploy Plan

- [x] Frontedge it
- [x] merge
- [x] deploy front & front-sse

### DIFF
--- a/connectors/src/types/shared/headers.ts
+++ b/connectors/src/types/shared/headers.ts
@@ -31,6 +31,8 @@ export function getHeaderFromUserEmail(email: string | undefined) {
     return undefined;
   }
 
+  // The email may exceed Latin-1 (internationalized addresses); DustAPI
+  // encodes extra header values on the wire (see @dust-tt/client baseHeaders).
   return {
     [DustUserEmailHeader]: email,
   };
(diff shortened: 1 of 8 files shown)


tdraier (approved):
LGTM

id13:
front-edge test looked good

PR #28976, opened 2026-07-15
[front] Make sandbox invocation GCS paths non-null

spolu:
## Description

Makes sandbox function invocation `gcsPath` mandatory now that the production backfill has completed.

- Adds the post-deploy `NOT NULL` constraint.
- Removes temporary nullable fallbacks from the model and invocation resource.
- Creates invocations transactionally so the database-generated invocation ID can be included in the final GCS path before the row becomes visible.
- Keeps the completed backfill script type-safe against the now non-null model.

## Tests

- Covered by existing tests
- Tested locally

## RIsks

None

## Deploy Plan

- deploy `front`
- run post-deploy migration

(backfill was already run as part of https://github.com/dust-tt/dust/pull/28957)

### DIFF
--- a/front/lib/resources/sandbox_function_invocation_resource.ts
+++ b/front/lib/resources/sandbox_function_invocation_resource.ts
@@ -25,6 +25,7 @@ import {
 } from "@app/lib/resources/string_ids";
 import type { ResourceFindOptions } from "@app/lib/resources/types";
 import { concurrentExecutor, withRetry } from "@app/lib/utils/async_utils";
+import { withTransaction } from "@app/lib/utils/sql_utils";
 import logger from "@app/logger/logger";
 import type {
   PostSandboxFunctionInvocationRequestBody,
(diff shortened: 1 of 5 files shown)


spolu:
PMRR

front/lib/resources/sandbox_function_invocation_resource.ts
```diff
@@ -365,29 +347,33 @@ export class SandboxFunctionInvocationResource extends BaseResource<SandboxFunct
...
+        version: SANDBOX_FUNCTION_INVOCATION_DATA_VERSION,
+        input,
+      };
+      const resource = new this(this.model, invocation.get(), {
+        sandboxFunction,
+        data,
+      });
+      const gcsPath = resource.buildGcsPath(auth);
+      await resource.update({ gcsPath }, t);
+      await resource.writeDataToGcs();
```
flvndvd, line 373: I see the attractiveness of using transaction here, but locking the DB to do HTTP call has never aged well. Can we avoid doing so?

flvndvd (commented):
LGTM

front/lib/resources/sandbox_function_invocation_resource.ts
spolu: Fair fair! But then I need to allow null in DB as I need the id to build the path.

Will refactor accordingly and have a couple asserts when accessing 👍

spolu: Wait no I can just write the data after the transaction. Good catch

PR #28977, opened 2026-07-15
Add support to read frame slideshows

flvndvd:
## Description

This PR widen the support so read file also works on slideshows.


aubin-tchoi (approved):
LGTM

PR #28980, opened 2026-07-15
Allow to set user pool usage to 0

tdraier:
## Description

Allow to disable pool usage by setting spend limit to 0 (individual, group and batch)

Update the usage bar to not show overage (was actually more confusing than helpful). Keep the tooltip to see the actual overage if any,

## Risk

none

## Deploy Plan

deploy front


tdraier:

- #28980 ◀

smb2268 (approved):
ty!

PR #28983, opened 2026-07-15
Sandbox functions: tool personal-authentication resolution endpoint

PopDaph:
## Description 

Follow up of #28863, which blocks a sandbox function tool action on `blocked_authentication_required` and publishes the `tool_personal_auth_required` event to the invocation stream when a tool needs personal OAuth. This PR adds the resolution half, mirroring the validate-action endpoint (#28829): the endpoint the merged personal-auth card already calls.

`POST /api/w/:wId/sandbox-functions/:functionIdOrSlug/invocations/:invocationId/actions/:actionId/resolve-authentication` with `{ outcome: "completed" | "denied" }`:

- Completed: compare-and-swap `blocked_authentication_required` -> `running`, then relaunch `launchSandboxFunctionToolWorkflow`. The tool re-runs from scratch and now finds the personal connection the user just created. Safe to re-run: the block happens at connection time, before the tool call, so there are no partial side effects.
- Denied: CAS -> `denied`; the poll surfaces it as `403 rejected` and the in-sandbox tool call fails.
- The pending `tool_personal_auth_required` event is removed from the invocation Redis channel so SSE history replay does not resurface the authentication card.
- Concurrent resolutions have exactly one winner through the CAS; losers succeed silently, and a late duplicate gets `action_not_blocked`, which the client treats as success.
- Launch failure after the CAS compensates the action to a terminal `errored` (CAS-guarded) then rethrows, so the poll doesn't hang until token expiry with no workflow. Same shape as the validate endpoint; the create/validate/resolve launch sites all share this exposure until launch is made durable.

Authorization: anyone who can read the pod space can resolve, same as the validate endpoint and the invocation SSE route. Invocations record no initiating user.

The params schema is now shared by both action-resolution routes (`ActionResolutionParamsSchema`).

## Tests

Endpoint tests: completed (status -> running, workflow relaunch, event removal), denied (denied, no relaunch), wrong status -> `action_not_blocked` 400, cross-invocation action -> 404.

## Risk

None, SandboxFunction is not live yet. The endpoint is unreachable until a tool call blocks on authentication (#28863).

## Deploy Plan

- [ ] Deploy front

### DIFF
--- a/front-api/routes/w/[wId]/sandbox-functions/[functionId]/invocations.test.ts
+++ b/front-api/routes/w/[wId]/sandbox-functions/[functionId]/invocations.test.ts
@@ -130,8 +130,12 @@ async function setupSandboxFunction({
 // approval-requiring tools (created without a workflow launch).
 async function setupBlockedAction({
   permission = "high",
+  blockedStatus = "blocked_validation_required",
 }: {
   permission?: MCPToolStakeLevelType;
+  blockedStatus?:
+    | "blocked_validation_required"
+    | "blocked_authentication_required";
 } = {}) {
   const { workspace, sandboxFunction, adminAuth, space } =
     await setupSandboxFunction();
@@ -157,7 +161,7 @@ async function setupBlockedAction({
     permission,
   });
   const [blockedCount] = await action.updateStatusFromExpected(adminAuth, {
-    status: "blocked_validation_required",
+    status: blockedStatus,
     expectedStatus: "running",
   });
   expect(blockedCount).toBe(1);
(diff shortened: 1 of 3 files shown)


spolu (approved):
LGTM

The zeropath comments call for a user field on the invocation so that we can check alignment here 👍 

Yours to take?

PopDaph:
Yes that's what I was asking yesterday > opened a slack thread in case someone had a better idea https://dust4ai.slack.com/archives/C0AGY9MTNJV/p1784126820573349 🙏🏻 

My plate yes 👍🏻 
Adding a comment in here to clarify and will handle on my next PR, I think it's okay to merge as is for now since internal only.

PR #28988, opened 2026-07-15
Expose linked file ids to browser clients

fontanierh:
## Description

Expose `X-Dust-File-Id` through the Front API CORS middleware. Without this, the path metadata endpoint returns the linked Frame file ID but `app.dust.tt` cannot read it, so Frame directives report that no linked file exists.

## Tests

Added middleware coverage for normal cross-origin and preflight responses. Reproduced the mismatch against production conversation data before applying the fix.

## Risk

Low. This only exposes an existing response header to allowed browser origins.

## Deploy Plan

Standard deploy.


flvndvd (commented):
LGTM

PR #28990, opened 2026-07-15
feat(pod-manager): split pinned-frame into its own never_ask tool

frankaloia:
## Summary

* Extracts pinned-frame pinning/unpinning out of the `edit_information` pod_manager tool into a dedicated **`set_pinned_frame`** tool with `never_ask` stake, so updating the Pod banner frame no longer prompts for approval.
* Motivation: the activation flow refreshes the pinned Frame every turn; a `low`-stake approval prompt on each refresh is disruptive. Pinning/unpinning a Pod-owned file the agent just created is a safe, high-frequency operation, so it warrants `never_ask` — while `edit_information` (title/description/access) keeps its `low` stake untouched.

## Risk

Low, but note a **behavior change for the `edit_information` tool contract**: it no longer accepts `pinnedFramePath`. Callers that previously pinned via `edit_information` must switch to `set_pinned_frame`. Grep shows no other in-repo callers pass `pinnedFramePath` to this tool.

## Testing
* Locally called tools

## Deployment
1. Deploy front


ElPicador (approved):
great idea!

PR #28991, opened 2026-07-15
feat(schedules): scope schedule tool to the current Pod

frankaloia:
## Summary

Makes the `schedules_management` MCP tool Pod-aware, so schedules created inside a Pod are attached to it

## Risk

Low

## Testing
Screenshot 2026-07-15 at 3 04 47 PM

## Deployment
1. Deploy front

### DIFF
--- a/front/lib/api/actions/servers/schedules_management/tools/index.ts
+++ b/front/lib/api/actions/servers/schedules_management/tools/index.ts
@@ -8,6 +8,7 @@ import {
 import { SCHEDULES_MANAGEMENT_TOOLS_METADATA } from "@app/lib/api/actions/servers/schedules_management/metadata";
 import { generateScheduleRule } from "@app/lib/api/assistant/configuration/triggers";
 import type { Authenticator } from "@app/lib/auth";
+import { SpaceResource } from "@app/lib/resources/space_resource";
 import {
   resolveTriggerSpaceId,
   TriggerResource,
@@ -22,7 +23,10 @@ import { Err, Ok } from "@app/types/shared/result";
 import assert from "assert";
 import { UniqueConstraintError } from "sequelize";
 
-function renderSchedule(schedule: ScheduleTriggerType): string {
+function renderSchedule(
+  schedule: ScheduleTriggerType,
+  podName?: string | null
+): string {
   const config = schedule.configuration;
   const scheduleDesc =
     schedule.naturalLanguageDescription ?? describeScheduleConfig(config);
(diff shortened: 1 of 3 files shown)


front/lib/api/actions/servers/schedules_management/tools/index.ts
```diff
@@ -186,10 +211,25 @@ export function createSchedulesManagementTools(
         ]);
       }
 
-      const scheduleList = schedulesResult.value
-        .map((s) => s.toJSON())
-        .filter(isScheduleTrigger)
-        .map((schedule) => renderSchedule(schedule))
+      // Resolve Pod names for schedules attached to a Pod (single batch query).
+      const podSIds = [
```
aubin-tchoi: nit: 
```suggestion
      const podIds = [
```

front/lib/api/actions/servers/schedules_management/tools/index.ts
```diff
@@ -186,10 +211,25 @@ export function createSchedulesManagementTools(
...
+        ...new Set(
+          schedules
+            .map((s) => s.spaceId)
+            .filter((id): id is string => id !== null)
+        ),
+      ];
+      const pods =
+        podSIds.length > 0 ? await SpaceResource.fetchByIds(auth, podSIds) : [];
+      const podNameBySId = new Map(pods.map((p) => [p.sId, p.name]));
```
aubin-tchoi:
```suggestion
      const podNameById = new Map(pods.map((p) => [p.sId, p.name]));
```

aubin-tchoi (approved):
LGTM

front/lib/api/actions/servers/schedules_management/tools/index.ts
```diff
@@ -65,6 +75,14 @@ export function createSchedulesManagementTools(
 
       const { agentConfiguration } = toolContext.runContext;
 
+      // When a podId is provided, attach the trigger to that Pod so the
+      // scheduled run lands in the Pod (where the pinned view lives).
+      const spaceIdRes = await resolveTriggerSpaceId(auth, podId);
+      if (spaceIdRes.isErr()) {
+        return new Err(new MCPError(spaceIdRes.error));
+      }
+      const spaceId = spaceIdRes.value;
```
aubin-tchoi, line 84: maybe we need to fetch the pod here to confirm we have write access on it (unless it's already done somewhere else)

frankaloia: So I actually think we don't want to check write. We want to check canRead to stay consistent with create conversation (which this does already):

   // Check if the user has access to the space.
    // Note, using canRead because spaces members do not have write access to the space as write is tied with datasources.
    if (space && !space.canRead(auth)) {
      throw new Error(
        "Cannot create conversation in a space you do not have access to."
      );
    }

PR #29004, opened 2026-07-15
feat(slack): invoke a specific model via +model mentions

Nils-Fedrigo:
## Description

Allows invoking a specific model through the Slack bot when the workspace has the `models_picker` feature flag, e.g.:

```
@dust +gpt-5.6-luna-high what is the capital of France?
```

When a `+`/`~`/`=` mention candidate does not exactly match an agent name, it is now matched (exact, case-insensitive) against the workspace's pickable models — either a bare model id (`+gpt-5.6-luna`, run with the model's default reasoning effort) or a model id with a reasoning effort suffix (`+gpt-5.6-luna-high`). On a match, the message is posted with a `modelSelection` and falls through to the regular default-agent resolution (channel routing rule, then default agents), mirroring the web input-bar picker semantics.

Changes:

- **connectors**: `processMentions` resolves model mentions against the models fetched from front (only when a mention candidate is present); an exact agent name match always takes priority over a model match, and with no models available the previous fuzzy agent matching is unchanged. The resolved `modelSelection` is sent in the message request body.
- **front-api**: new system-key-only `GET /api/v1/w/{wId}/models` endpoint listing selectable models with their supported reasoning efforts (empty list when `models_picker` is off, so the connector needs no separate flag check).
- **public API**: optional `modelSelection` field on the v1 messages and conversation-creation endpoints, validated against the flag and known provider/model/effort values, then threaded to `postUserMessage` where it is stored as the user message's `requestedModel` (same path as the web picker, so run-time tier enforcement applies unchanged). `postUserMessageAndWaitForCompletion` gained the same optional param for `blocking: true`.
- **sdks/js**: `PublicModelSelectionSchema`, `AvailableModelSchema` and a `getAvailableModels()` client method. Model/provider ids are plain strings in the SDK so new models don't require an SDK release.

All schema changes are additive and optional (no breaking changes).

## Tests

- `connectors/src/lib/bot/mentions.test.ts`: model mention with effort suffix, bare model id (default effort), case-insensitivity, `=` prefix, unsupported effort suffix falling back to fuzzy agent matching, exact agent name shadowing a model, and empty model list preserving previous behavior.
- `front-api/routes/v1/w/[wId]/models.test.ts`: 403 for non-system keys, empty list without the flag, selectable model shape with the flag.
- `front-api/routes/v1/w/[wId]/assistant/conversations/index.test.ts`: 400 when `modelSelection` is sent without the flag, 400 for unknown model ids.

## Risk

Low. The feature is gated behind the `models_picker` feature flag (dust_only): without it the models endpoint returns an empty list and the public API rejects `modelSelection`, so the Slack mention behavior is unchanged. The model-list fetch in the Slack bot degrades gracefully (logs a warning and keeps agent matching) if the call fails.

## Deploy Plan

Deploy front before connectors so the `/api/v1/w/{wId}/models` endpoint and the `modelSelection` field exist when the connector starts sending them (the connector tolerates the endpoint failing in the interim).


Nils-Fedrigo:

- #29004 ◀

PR #29005, opened 2026-07-15
feat(analytics): expose tags to analytics mcp

sfriquet:
## Description

Expose tags to workspace analytics mcp. Goal is to have the analyst agent be able to leverage tags, e.g: "how many credits are used by agents tagged finance ?"

- creates a new call `get_top_tags`: returns tags by message count (gets the name and the sid)
- adds tagsId to other functions like `get_credit_usage`

refs https://github.com/dust-tt/tasks/issues/8167

## Tests

Screenshot 2026-07-15 at 18 41 42
Screenshot 2026-07-15 at 18 42 27

### DIFF
--- a/front/lib/api/assistant/observability/top_tags.ts
+++ b/front/lib/api/assistant/observability/top_tags.ts
@@ -1,38 +1,37 @@
-import type { WorkspaceTopUserRow } from "@app/lib/api/analytics/workspace_analytics";
 import { buildAgentAnalyticsBaseQuery } from "@app/lib/api/assistant/observability/utils";
 import type { ElasticsearchError } from "@app/lib/api/elasticsearch";
 import { bucketsToArray, searchAnalytics } from "@app/lib/api/elasticsearch";
 import type { Authenticator } from "@app/lib/auth";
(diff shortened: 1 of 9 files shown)


front/lib/api/assistant/observability/top_tags.ts
```diff
@@ -0,0 +1,117 @@
...
+};
+
+// Ranks agent tags by message count over a time window, with the count of
+// distinct agents carrying each tag.
+// Backs the top-tags analytics endpoint.
+// Either `days` or `startDate`/`endDate` bounds the window; source/agent/user
+// filters are optional.
+// Since agents can have multiple tags, counts overlap and can sum to more than
+// the total message volume.
+export async function fetchTopTags(
```
ElPicador: the naming is weird, shouldn't it be `fetchAgentTopTags`?

ElPicador (approved):
👍

front/lib/api/assistant/observability/top_tags.ts
sfriquet: renamed to `agent_tags` in the context of the analytics

PR #29006, opened 2026-07-15
[front] Time-bound workspace seat limits (schedulable commitments)

tdraier:
Give `workspace_seat_limit` a `startAt`/`endAt` window (like memberships) so
seat commitments can be scheduled ahead of time, and add the resource methods
to read/write them.

- Model: add startAt (NOT NULL default NOW) + endAt (nullable); replace the
  (workspaceId, seatType) unique index with a partial unique index scoped to
  open-ended rows so scheduled/historical rows coexist.
- Pre-deploy migration adds the columns (backfilled startAt floored to the
  hour) and swaps the index.
- Resource: fetchByWorkspace reads the active value only; fetchScheduleByWorkspace
  returns current + scheduled-future phases; setScheduledLimit (windowed) and
  setScheduleForSeatType (replace a seat type's whole timeline, deriving
  contiguous windows from start-only phases).
- Resource tests.

## Deploy Plan

run `./scripts/run-migrations.sh pre-deploy front`

### DIFF
--- a/front/lib/resources/workspace_seat_limit_resource.ts
+++ b/front/lib/resources/workspace_seat_limit_resource.ts
@@ -1,5 +1,7 @@
 import type { Authenticator } from "@app/lib/auth";
 import { BaseResource } from "@app/lib/resources/base_resource";
+import { frontSequelize } from "@app/lib/resources/storage";
+import { Op } from "@app/lib/resources/storage/data_types";
 import { WorkspaceSeatLimitModel } from "@app/lib/resources/storage/models/workspace_seat_limit";
 import type { ReadonlyAttributesType } from "@app/lib/resources/storage/types";
 import type { ModelStaticWorkspaceAware } from "@app/lib/resources/storage/wrappers/workspace_models";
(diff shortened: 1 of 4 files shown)


tdraier:

- #29006 ◀
    - #29007
        - #29008

ElPicador (approved):

front/migrations/pre-deploy/20260715120000_add_time_bounds_to_workspace_seat_limit.sql
```diff
@@ -0,0 +1,24 @@
...
+UPDATE "public"."workspace_seat_limits"
+SET
+  "startAt" = date_trunc('hour', "startAt");
+
+-- Relax the old one-row-per-(workspace, seat type) unique constraint and replace
+-- it with a partial unique index scoped to open-ended rows, so that scheduled
+-- and historical rows (endAt IS NOT NULL) can coexist for the same seat type.
+DROP INDEX CONCURRENTLY IF EXISTS "workspace_seat_limits_workspace_seat_type_idx";
+
+CREATE UNIQUE INDEX CONCURRENTLY IF NOT EXISTS "workspace_seat_limits_workspace_seat_type_active_idx" ON "public"."workspace_seat_limits" ("workspaceId", "seatType")
```
PopDaph, line 22: Question: why not first creating the new index and then deleting the old one?

front/lib/resources/workspace_seat_limit_resource.ts
```diff
@@ -38,33 +68,100 @@ export class WorkspaceSeatLimitResource extends BaseResource<WorkspaceSeatLimitM
...
+    for (const bucket of result.values()) {
+      bucket.sort((a, b) => a.startAt.getTime() - b.startAt.getTime());
     }
     return result;
   }
 
   /**
-   * Create or update the configured limit for a (workspace, seat type).
-   * Rejects when `maxSeats` is set and less than `minSeats`.
+   * Create or update the currently effective (open-ended) limit for a
```
PopDaph, line 161: upsert stops meaning "apply now" once a schedule exists. The comment says it targets "the currently effective (open-ended) limit", but after setScheduledLimit the open-ended row is the future one, so upsert silently edits the scheduled value and leaves the current value untouched. 

Is that intended for switch_contract.ts and the manage_seat_limits plugin? Both call it meaning "apply now", so with a schedule in place they'd silently edit the future value instead.

PopDaph (commented):
👍🏻

PR #29011, opened 2026-07-15
[Admin Gov] Rename group_permissions permissionType -> grantType

philipperolet:
## Description

Mechanical rename of the `group_permissions` grant vocabulary from `permissionType` → `grantType`,
ahead of the **role-backed permissions** model. In that model a grant row stores a registry-defined
*grant type* — a role such as `member`/`editor`, or a singleton capability such as `create`/`use` —
rather than a plain verb, so `grantType` is the accurate name.

This PR changes only names: no values change, no behavior changes, the registry still holds verbs.
The semantic reshape (verb lists → role definitions, checks expanding verb→grant-types) lands in
follow-up PRs of this stack.

**No schema change.** `group_permissions` has a live consumer on the model-tier resolution path
(`ModelsTierResource`, reached from assistant model resolution via `lib/model_tiers`), so a physical
`RENAME COLUMN` would break model resolution during the deploy window and on rollback. Instead the
`grantType` attribute maps to the existing physical column `permissionType` via Sequelize `field:`;
the physical rename is deferred to a later expand/contract migration. (Raised in review by @matteotrab
and the external reviewer.)

- Type `PermissionType` → `GrantType`, `PERMISSION_TYPES` → `GRANT_TYPES`, `isPermissionType` →
  `isGrantType`, `GovernancePermission.permissionType` → `grantType`, `MODELS_TIER_PERMISSION_TYPE`
  → `MODELS_TIER_GRANT_TYPE`, propagated through the model, registry (`assertValidGrant`), resource,
  `auth.hasWorkspacePermission`, `models_tier_resource`, the governance page components, the SWR
  layer, and the tests.
- Also removes a pre-existing name collision with the unrelated
  `@app/types/resource_permissions` `PermissionType` (the two are kept distinct in `auth.ts`).

No end-user behavior change.

## Risks

Blast radius: model-tier grants (`group_permissions`), read on the assistant model-resolution path.
This PR does not touch the schema and does not change any query, so that path is unaffected.
Risk: low

## Deploy Plan

- deploy front + front-sse (no migration)


philipperolet:

- #29011 ◀

front/migrations/pre-deploy/20260715120000_rename_group_permissions_grant_type.sql
```diff
@@ -0,0 +1,18 @@
...
+-- role-backed permissions model (a grant row will store a registry-defined grant type, not a plain
+-- verb). Mechanical rename only; values are unchanged by this migration. The sole live consumer is
+-- the model-tier feature (lib/resources/models_tier_resource.ts), whose code ships in lockstep.
+
+/*
+Statement 0
+*/
+SET SESSION statement_timeout = 3000;
+SET SESSION lock_timeout = 3000;
+ALTER TABLE "public"."group_permissions" RENAME COLUMN "permissionType" TO "grantType";
```
matteotrab: This table is currently being used (see the model_tier_resource). Shouldn't we first create a new column and then drop the old one?

matteotrab (approved):
Thank youi Left one comment

front/migrations/pre-deploy/20260715120000_rename_group_permissions_grant_type.sql
philipperolet: Good call — you're right, and the external reviewer flagged the same thing. I confirmed `group_permissions` is read on the assistant model-resolution path (`ModelsTierResource.resolveAllowedTierNames` → `listForGroups`, reached from `lib/model_tiers/access` + `enabled_models` via `resolve_model` / agent configuration), so it's a hot path, not just admin settings. A single-shot `RENAME COLUMN` would break model resolution during the deploy window and on rollback.

I dropped the migration entirely. The `grantType` attribute now maps to the existing physical column via Sequelize `field: "permissionType"`, so there's **no schema change** in this PR — only the code vocabulary is renamed. The physical column rename can be done later as a proper expand/contract migration (or kept as-is; the `field:` mapping is idiomatic here). PTAL.

matteotrab: Perfect, thank you

PR #29013, opened 2026-07-15
Add script to revoke/change a Frame's share scope by token

smb2268:
## Description

Adds a reusable admin script (`front/scripts/revoke_frame_share.ts`) to update a Frame's share scope (`shareable_files.shareScope`) by its share `token`.

Context: a Frame publicly shared by a now-deactivated user is still publicly accessible. Workspace admins can't open the departed creator's conversation via the normal UI to change its sharing, and there's currently no Poke plugin or endpoint to patch a single Frame's share scope, so this needs a manual backend fix (see dust-tt/tasks#9619).

The script:
- Looks up the `shareable_files` row by `tokenId` (the UUID from the share URL).
- Defaults `--scope` to the most restrictive value, `emails_only`, but accepts any valid `FileShareScope` (`emails_only`, `workspace_and_emails`, `public`) via `--scope`.
- Is a dry run by default; requires `--execute` to actually write.
- Is a no-op if the row is already at the target scope.

Lives under `front/scripts/` (not `front/migrations/`) since it's a reusable, parameterized admin script rather than a one-off data migration, alongside similar scripts like `revoke_memberships_from_csv.ts` and `unrevoke_membership.ts` (moved here per review feedback).

Fixes dust-tt/tasks#9619

## Tests

Not run in CI locally beyond a manual read-through against the `ShareableFileModel` definition (`front/lib/resources/storage/models/files.ts`) and the existing `makeScript` helper. No unit tests added, this mirrors the pattern of other scripts in `front/scripts/`. Local validation was not run in the Computer; relying on PR CI (typecheck/lint) for verification.

## Risk

Low risk: the script only touches a single `shareable_files` row, identified by its unique `token`, and defaults to dry-run (no write) unless `--execute` is passed. No schema changes. Rollback is trivial: re-run with `--scope` set back to the previous value.

## Deploy Plan

1. Merge this PR.
2. On prodbox (correct region for the affected workspace), pull `main`.
3. Run in dry-run mode first to confirm the target row:
   `npx tsx front/scripts/revoke_frame_share.ts --tokenId <token>`
4. Run with `--execute` and the intended `--scope` to apply.
5. Post the command and outcome to `#engineering_papertrail`.


frankaloia (approved):
LGTM but prob doesnt need to be under migrations specifically since it's a reusable script

PR #29014, opened 2026-07-15
Move invocation GCS write outside transaction

spolu:
## Description

Move sandbox function invocation data writes to GCS outside the database transaction callback. The transaction now only creates the invocation and persists its generated GCS path.

## Tests

- `NODE_ENV=test npm test -- lib/resources/sandbox_function_invocation_resource.test.ts`

## Risk

Low. This only changes when the initial invocation payload is written to GCS.

## Deploy Plan

- Deploy `front`


spolu:
PMRR

flvndvd (commented):
🙏

PR #29016, opened 2026-07-15
feat(activation): refine recommendation-flow prompt

frankaloia:
## Summary

- Various activation prompt improvements based on user feedback

## Risk

Low - not customer facing yet

## Testing

- Local activaiton runs

## Deployment
1. Deploy front


ElPicador (approved):
I would add a few test cases for this prompt, I'll see with Phoebe to do it!

frankaloia:
> I would add a few test cases for this prompt, I'll see with Phoebe to do it!

@ElPicador Yes, but tbh it's changing so rapidly I think it's worth holding off for now

PR #29018, opened 2026-07-16
[front] remove background color from date indicator

ykmsd:
## Description
This PR is to remove the background color from date indicator in conversation view:

before:
CleanShot 2026-07-16 at 09 17 24@2x

after: 
CleanShot 2026-07-16 at 09 17 49@2x


ClementAupiais (approved):
LGTM

PR #29020, opened 2026-07-16
Surface space-restricted skills in capability pickers

xebeohp:
Removes the globalSpaceOnly filter from useSkills in the capabilities
picker, pod settings, slash-command capabilities, and usePodDefaultSkills
so skills restricted to spaces/Pods the user can access appear in the
pickers. Adds an endpoint test for the default vs globalSpaceOnly listing
behavior.


xebeohp:

- #28851
    - #29020 ◀

PR #29022, opened 2026-07-16
[LLM] Add global Gemini stream endpoints + fix Gemini pricing/regions

pmilliotte:
## Description

Add global-region Gemini stream endpoints (3.1 Pro, 3.1 Flash Lite, 3.5 Flash) across the inference and dust layers, and correct Gemini pricing (implicit caching → no cache-creation charge, EU multi-region markup) while routing EU endpoints to the `eu` region.

## Tests

Tested locally.

## Risk

Low.

## Deploy Plan

- Deploy `main` on `front`


adrsimon (commented):
- LGTM

PR #29024, opened 2026-07-16
[front] perf: skip loading unrequested skill instructions

aubin-tchoi:
## Description

When `SkillResource` callers set `withInstructions: false`, code-defined skill instructions are skipped, but custom skills still `SELECT` `instructions` and `instructionsHtml` from the db. These fields can be large, and the callers do not consume them.

This PR excludes both columns from the `SELECT` when instructions are not requested.

Strongly suspect these text values are stored in TOAST storage. Not selecting these columns should avoid fetching and decompressing their external TOAST values, and it also removes a bit of I/O + memory usage in the node process. 

Should not change anything for single-skill fetches, targets more specifically large lists of skills.

## Tests

- Added tests.

## Risk

- Low.

## Deploy Plan

- Deploy front.

### DIFF
--- a/front/lib/resources/skill/skill_resource.ts
+++ b/front/lib/resources/skill/skill_resource.ts
@@ -566,14 +566,32 @@ export class SkillResource extends BaseResource<SkillConfigurationModel> {
       withInstructions = true,
       withTools = true,
       withFileAttachments = true,
+      attributes,
       ...otherOptions
     } = options;
 
+    const instructionAttributes = ["instructions", "instructionsHtml"];
+    const isInstructionAttribute = (attribute: unknown) =>
+      instructionAttributes.some(
+        (instructionAttribute) => instructionAttribute === attribute
+      );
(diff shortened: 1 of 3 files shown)


front/lib/resources/skill/skill_resource.ts
```diff
@@ -571,6 +571,9 @@ export class SkillResource extends BaseResource<SkillConfigurationModel> {
 
     const customSkills = await this.model.findAll({
       ...otherOptions,
```
pmilliotte, line 573: We're going to have a pb if otherOptions sets the `attributes` property

aubin-tchoi: good catch, fixing

aubin-tchoi: done (narrowed down the type to forbid this, baseFetch controls the `attributes` field based on the other higher-level options like `withInstructions`)

pmilliotte (approved):

PR #29026, opened 2026-07-16
feat(front): notify on model tier changes and show picker loading state

Nils-Fedrigo:
## Description

Improves feedback when changing allowed model tiers from the Usage page. Previously the mutation hooks in `lib/swr/model_tiers.ts` only surfaced a notification on failure, so a successful change was silent.

- Add a success notification after every model tier mutation succeeds — for the user, group, and workspace levels (both set and clear paths).
- Show a visible loading state on the tier picker button while a mutation is in flight: the `ModelTierPickerDropdown` button now renders Sparkle's built-in `isLoading` spinner in addition to being disabled.

## Tests

Manual: on the workspace Usage page (with the `models_picker` flag enabled, as an admin), change the workspace, a group, and a user model tier. Verify the picker shows a spinner while saving and a success toast on completion, and that clearing an override notifies as expected.

### DIFF
--- a/front/components/workspace/ModelTierPickerDropdown.tsx
+++ b/front/components/workspace/ModelTierPickerDropdown.tsx
@@ -42,6 +42,7 @@ export function ModelTierPickerDropdown({
           size="sm"
           isSelect
           label={label}
+          isLoading={isMutating}
           disabled={readOnly || isMutating}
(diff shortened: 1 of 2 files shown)


Nils-Fedrigo:

- #29026 ◀

Fraggle (commented):
LGTM

PR #29031, opened 2026-07-16
feat(temporal): scaffold per-workspace activation scheduler workflow

ElPicador:
## Description

Scaffolds a per-workspace Temporal scheduler for the upcoming Activation feature, following the same pattern as `temporal/reinforcement`: one schedule per workspace, firing at the start of the regional workday (9am local, 1h jitter) with `ScheduleOverlapPolicy.SKIP`, running a short-lived per-workspace workflow.

Gated behind a new `activation_scheduler` feature flag (off by default). This is intentionally just the skeleton — `runActivationForWorkspaceActivity` is a logging placeholder; business logic will be plugged in separately. No cross-workspace reconciliation loop is included: schedules are started/stopped explicitly rather than reconciled against the flag on a cron.

Lifecycle is wired at both ends:
- `startActivationWorkspaceSchedule` is called from `createWorkspaceInternal` when the flag is already enabled for the workspace (workspace-level flag or global rollout) at creation time.
- `deleteActivationWorkspaceSchedule` is called from workspace hard-deletion (`deleteWorkspaceActivity`) so a deleted workspace's schedule doesn't keep firing indefinitely.

## Risk

Low — new code behind a flag that defaults to off, and the worker isn't wired to run anything by default.

### DIFF
--- a/front/lib/iam/workspaces.ts
+++ b/front/lib/iam/workspaces.ts
@@ -1,7 +1,7 @@
 import { seedWorkspaceCapabilities } from "@app/lib/api/permissions/governance_seeding";
 import { activateCreditPricedFreePlanForWorkspace } from "@app/lib/api/subscription";
 import { getOrCreateWorkOSOrganization } from "@app/lib/api/workos/organization";
-import { Authenticator } from "@app/lib/auth";
+import { Authenticator, hasFeatureFlag } from "@app/lib/auth";
 import type { SessionWithUser } from "@app/lib/iam/provider";
 import { PlanModel } from "@app/lib/models/plan";
 import {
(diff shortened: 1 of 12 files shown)


ElPicador:

- #29031 ◀
    - #29033
        - #29059
            - #29065
                - #29067

front/lib/iam/workspaces.ts
```diff
@@ -103,6 +104,18 @@ export async function createWorkspaceInternal({
   // Ensure all auto MCP server views are created for the workspace
   await MCPServerViewResource.ensureAllAutoToolsAreCreated(auth);
 
+  if (await hasFeatureFlag(auth, "activation_scheduler")) {
+    const scheduleRes = await startActivationWorkspaceSchedule({
```
frankaloia, line 108: Do you think we need to do this for every workspace? At least to start, I think this will only be needed for a smaller set of workspace especially given it's initiated via poke

ElPicador: I like "hiding" code behind flags so we can enable them as a whole. 

If you prefer to have the poke plugin to start the workflow that's also fine by me.

frankaloia: I think it's ok for now behind a FF. TBD if we actually fully release it without FF

frankaloia (approved):
LGTM pending lint

PR #29032, opened 2026-07-16
feat(front): show per-message credit cost on all plans

Nils-Fedrigo:
## Description

Show the per-message "Credit cost" row in the agent message action dropdown for all
workspaces, not just those on a credit-priced (`CP_`) plan.

Previously `useCreditCostMenuItem` gated the item behind `isCreditPricedPlan(subscription.plan)`,
so only credit-priced workspaces saw the cost. That plan check is removed, so the row now
appears on every plan whenever a message has a computed credit cost.

Also cleans up the now-redundant `isCreditPriced` flag that was always `true` after removing
the gate: the hook now returns `DropdownMenuItemProps | null` directly, and the two call sites
in `AgentMessage.tsx` (open-revalidation guard and the render condition) are simplified
accordingly.

## Tests

Manual: open an agent message's "..." action menu and confirm the "Credit cost" row shows
with the correct total (and sub-agent breakdown tooltip) on a non-credit-priced workspace,
including the loading placeholder while the authoritative cost is being fetched.

### DIFF
--- a/front/components/assistant/conversation/AgentMessage.tsx
+++ b/front/components/assistant/conversation/AgentMessage.tsx
@@ -319,14 +319,14 @@ export function AgentMessage({
   });
   const refreshedAgentMessage =
     refreshedMessage?.type === "agent_message" ? refreshedMessage : null;
-  const { creditCostItem, isCreditPriced } = useCreditCostMenuItem({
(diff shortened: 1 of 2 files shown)


Nils-Fedrigo:

- #29032 ◀

tdraier (approved):

PR #29034, opened 2026-07-16
feat(front): add allowed models section to model providers page

Nils-Fedrigo:
## Description

Adds an **"Allowed models"** section to the Model Providers page, mirroring the model-tier management feature already available on the Usage page. The section is gated behind the `models_picker` feature flag and is admin-only.

It exposes three tabs (matching the Usage page):

- **Members** — a table listing each member with **Name / Groups / Models tier**, plus a three-dots menu with a **"Models tier ›"** submenu to override a member's allowed tier.
- **Groups** — reuses `GroupsUsageTable` showing **Group / Members / Models tier** (spend-limit column hidden here).
- **Workspace settings** — the workspace-wide model-tier toggle.

Shared logic (tier resolution, tier picker, submenu builder, SWR hooks) is reused across both pages; only the presentational tables differ.

Changes:
- New `AllowedModelsSection` (tabbed section + data fetching) and `AllowedModelsMembersTable` (lightweight members table composing the existing tier helpers).
- `GroupsUsageTable` gains a `showSpendLimitColumn` prop (default `true`) so the section can render Group/Members/Models tier only.
- `ModelTiersSettingsCard` gains a `showHeader` prop (default `true`) so the settings tab shows just the toggle.
- Wired into `ModelProvidersPageContent`, gated by `hasFeature("models_picker") && isAdmin(workspace)` — the same gate used on the Usage page.

## Tests

Manual verification behind the `models_picker` flag on the Model Providers page: each tab renders, the members three-dots "Models tier" submenu sets/clears a per-user override, the groups picker updates per-group tiers, and the workspace toggle updates the workspace-wide tier. Usage page behavior is unchanged (both extended components default to their previous rendering).

### DIFF
--- a/front/components/pages/workspace/model_providers/ModelProvidersPageContent.tsx
+++ b/front/components/pages/workspace/model_providers/ModelProvidersPageContent.tsx
@@ -4,6 +4,7 @@ import { ProvidersConfigurationList } from "@app/components/pages/workspace/mode
 import { ProvidersToggleList } from "@app/components/pages/workspace/model_providers/ProvidersToggleList";
 import { RegionalModelsOnlyToggle } from "@app/components/pages/workspace/model_providers/RegionalModelsOnlyToggle";
 import { USED_MODEL_CONFIGS } from "@app/components/providers/types";
+import { AllowedModelsSection } from "@app/components/workspace/AllowedModelsSection";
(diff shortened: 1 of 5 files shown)


Nils-Fedrigo:

- #29034 ◀

PR #29038, opened 2026-07-16
[sparkle] Center spinner and preserve size in button loading state

ClementAupiais:
## Description

Previously, the `Button` loading state replaced the icon with a spinner and hid other content, causing the button to change size depending on the label/icon widths. This PR fixes that by keeping the resting content (icon, label, chevron, `iconRight`) in the layout but invisible during loading, and overlaying an absolutely-centered spinner on top. The button wrapper gets `position: relative` only when `isLoading` is true to support the overlay positioning.

## Tests

Visual check: toggle `isLoading` on buttons of various sizes/variants and confirm the button does not resize and the spinner is centered.

## Risk

Low — cosmetic-only change to `Button.tsx` in Sparkle. No logic or API surface changes.

## Deploy Plan

Deploy sparkle.

### DIFF
--- a/sparkle/src/components/Button.tsx
+++ b/sparkle/src/components/Button.tsx
@@ -344,13 +344,9 @@ const Button = React.forwardRef<HTMLButtonElement, ButtonProps>(
       return <Icon visual={visual} size={iconSize} className={extraClass} />;
     };
 
-    const content = (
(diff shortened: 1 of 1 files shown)


pinotalexandre (approved):
LGTM

PR #29040, opened 2026-07-16
[connectors] enh: stream gdrive spreadsheets to reduce heap footprint

id13:
## Description

Concurrent google spreadsheet processing on connectors has always been putting pressure on heap memory. Previous mitigations: concurrency down to 8 per activity, cut processing at 50k rows. But we were still buffering each whole sheet three times over (JSON response, rows, CSV) before sending it to core.

This PR streams sheet values instead: one `values.get` call per sheet with `responseType: "stream"`, parsed incrementally with `stream-json` (already a connectors dependency — the Microsoft connector uses it for GCS delta files). Rows are folded into the CSV builder in 1000-row chunks as they arrive on the wire, so live memory is bounded by one chunk + the CSV being built (≤ 50MB cap) + the one CSV being upserted. When a cap trips (50k non-empty rows / 50MB padded CSV), we stop pulling and close the socket — bytes we never download.

Behavior deltas vs main:

- `ERR_STRING_TOO_LONG` can no longer occur — the response body is never buffered into a string; dropped that handling.
- "Unable to parse range" 400s (sheet renamed mid-sync) now skip only that sheet instead of every sheet in the same batched call.
- gaxios attaches non-2xx bodies as raw strings when `responseType: "stream"`; `isUnableToParseError` handles both shapes.
- Heartbeat before each API call and per 1000-row chunk; one log per file with `valuesGetCallCount`.

## Tests

17 unit tests: CSV building/padding byte-identical to the previous pad-then-`stringify` implementation, both caps closing the stream early (the fixture asserts teardown ran and no further chunks were pulled), per-sheet not-readable, file-level skip, plus real-stream tests of the JSON parser (absent `values`, malformed JSON rejects, early exit destroys the source).

## Risk

Low
Blast radius: spreadsheet file processing activity in gdrive sync.
Worst case, gsheet sync misbehaves on some spreadsheet shape. Safe to rollback (deploy-only, no schema change).

## Deploy Plan

- [ ] Deploy connectors
- [ ] Watch connectors heap + Sheets API 429 rate in DD

### DIFF
--- a/connectors/src/connectors/google_drive/temporal/spreadsheets.test.ts
+++ b/connectors/src/connectors/google_drive/temporal/spreadsheets.test.ts
@@ -1,0 +1,393 @@
+import type {
+  Sheet,
+  SheetContentEvent,
+  StreamSheetRows,
+} from "@connectors/connectors/google_drive/temporal/spreadsheets";
+import {
+  fetchSheetContentsAsCsv,
+  sheetValuesRange,
+  streamRowsFromValuesJson,
+} from "@connectors/connectors/google_drive/temporal/spreadsheets";
+import logger from "@connectors/logger/logger";
+import { stringify } from "csv-stringify/sync";
+import { Readable } from "stream";
+import { describe, expect, it } from "vitest";
+
+function makeSheet(overrides: Partial<Sheet> = {}): Sheet {
+  return {
+    id: 1,
+    spreadsheet: { id: "spreadsheet-id", title: "My Spreadsheet" },
+    title: "Sheet1",
+    gridRowCount: 10,
+    ...overrides,
+  };
+}
+
+// Serves per-sheet fixtures as chunked row streams the way the streamed values.get produces
+// them, recording the chunks pulled per sheet and whether each sheet's stream was torn down
+// (its finally ran) — the observable proof that an aborted consumer closes the stream.
+function makeRowStreamer(
+  rowsBySheetTitle: Record<string, string[][]>,
+  { chunkRows = 1000 }: { chunkRows?: number } = {}
+) {
+  const pulledChunks: Record<string, number> = {};
+  const closedSheets: string[] = [];
+  const streamSheetRows: StreamSheetRows = async function* (sheet) {
+    try {
+      const allRows = rowsBySheetTitle[sheet.title];
+      if (!allRows) {
+        throw new Error(`Unknown sheet: ${sheet.title}`);
+      }
+      for (let i = 0; i < allRows.length; i += chunkRows) {
+        pulledChunks[sheet.title] = (pulledChunks[sheet.title] ?? 0) + 1;
+        yield { kind: "rows", rows: allRows.slice(i, i + chunkRows) };
+      }
+    } finally {
+      closedSheets.push(sheet.title);
+    }
+  };
+  return { streamSheetRows, pulledChunks, closedSheets };
+}
+
+async function collectEvents(
+  events: AsyncGenerator<SheetContentEvent, void>
+): Promise<SheetContentEvent[]> {
+  const collected: SheetContentEvent[] = [];
+  for await (const event of events) {
+    collected.push(event);
+  }
+  return collected;
+}
+
+describe("fetchSheetContentsAsCsv", () => {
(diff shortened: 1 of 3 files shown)


connectors/src/connectors/google_drive/temporal/spreadsheets.test.ts
```diff
@@ -0,0 +1,199 @@
...
+    gridRowCount: 10,
+    ...overrides,
+  };
+}
+
+// Serves values from a fixture, slicing by the row range requested (e.g. "'Sheet1'!1:10000"),
+// the way the Sheets API values.get endpoint does, and records the requested ranges.
+function makeFetcher(allRows: string[][], requestedRanges: string[] = []) {
+  return async (range: string) => {
+    requestedRanges.push(range);
```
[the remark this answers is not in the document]
id13: fair

connectors/src/connectors/google_drive/temporal/spreadsheets.ts
```diff
@@ -25,29 +26,57 @@ import type { GoogleDriveObjectType, ModelId } from "@connectors/types";
...
-import type { GaxiosResponse, OAuth2Client } from "googleapis-common";
+import type { OAuth2Client } from "googleapis-common";
 import { GaxiosError } from "googleapis-common";
 
 const MAXIMUM_NUMBER_OF_GSHEET_ROWS = 50000;
 const MAXIMUM_NUMBER_OF_SHEETS_PER_SPREADSHEET = 50;
+// Number of rows fetched per Sheets API call. Sheet contents are fetched in row chunks so that
+// peak memory per activity stays bounded by one chunk plus the accumulated CSV (<= MAX_CSV_SIZE),
+// regardless of the sheet's total row count.
+const SHEET_VALUES_FETCH_CHUNK_ROWS = 1000;
```
philipperolet: IIUC we'd make then at most 50 queries for a large spreadsheet (instead of one)
I'd up that, likely to 5k, to mitigate rate limiting challenges

The quota is 60 reads/user/minute apparently, unless I misunderstands it

connectors/src/connectors/google_drive/temporal/spreadsheets.ts
```diff
@@ -195,119 +253,85 @@ async function processSheet(
...
   }
 
-  localLogger.info(
-    "[Spreadsheet] Failed to import sheet. Will be deleted if already synced."
-  );
-
-  return false;
-}
-
-async function batchGetSheets(
```
philipperolet, line 238: I think batchGet here was a way to go fast on many-spreadsheets with many-sheets: faster grab, less latencies, less consumption of rate limit with the Google Drive API
not sure what waters we get into if we remove it and do all sequentially
@fontanierh @flvndvd who wrote this code years ago might have better insights?
maybe there's a way to batch-grab while limiting at 5k row total to get the best compromise?

connectors/src/connectors/google_drive/temporal/spreadsheets.ts
philipperolet: also if we go this route, things will get slower so maybe heartbeating might make sense (although the 5mn Heartbeat timeout should be enough)

connectors/src/connectors/google_drive/temporal/spreadsheets.ts
```diff
@@ -115,75 +142,106 @@ async function upsertGdriveTable(
...
-    }
+  localLogger.info(
+    { csvSizeBytes, rowCount: csvLines.length },
+    "[Spreadsheet] Fetched sheet content."
+  );
 
-    throw err;
-  }
+  // If a row has less cells than the widest row, fill the gap with empty cells.
+  const csv = `${csvLines
```
philipperolet: should we check the expected size is < MAX_CSV_SIZE here before adding the commas?
if there's a late wide row, we'll fill up the memory no?

connectors/src/connectors/google_drive/temporal/spreadsheets.ts
id13: Will see if it is possible, if so the code will be simpler

PR #29044, opened 2026-07-16
Move sandbox functions to a dedicated Temporal queue

spolu:
## Description

Move sandbox function invocation and tool workflows out of the agent-loop worker and onto a dedicated `sandbox-functions-queue-v1` Temporal queue and worker.

The new worker uses the front Temporal namespace, is registered for production bundle generation and local development, and supports TypeScript path aliases in its workflow bundle.

## Tests

- 45 focused `front` and `front-api` tests
- Built all Temporal workflow bundles
- Verified `sandbox-functions-queue-v1` reaches `RUNNING` locally
- Front typecheck

## Risk

Medium. Sandbox function workflows move to a new task queue. Tasks will wait until the corresponding worker is deployed.

## Deploy Plan

- Deploy `front`
- Apply https://github.com/dust-tt/dust-infra/pull/936 to start the sandbox function workers


front/temporal/sandbox_functions/config.ts
```diff
@@ -0,0 +1,3 @@
+const QUEUE_VERSION = 1;
+
+export const QUEUE_NAME = `sandbox-functions-queue-v${QUEUE_VERSION}`;
```
PopDaph, line 3: I'm wondering if I thought this through correctly: I asked to move sandbox functions to a different worker, but should the invocation and tool activities also be isolated from each other in separate queues, not just from the agent loop?

if both runSandboxFunctionInvocationActivity and runSandboxFunctionToolActivity run on this one queue: the invocation activity blocks in sandbox.exec() for the whole function run while the function is waiting on its tool calls, and those tool calls execute as runSandboxFunctionToolActivity on this same queue. so at ≥ maxConcurrentActivityTaskExecutions concurrent tool-using invocations, every slot could be held by an invocation waiting for a tool that can't get a slot, and they'd all time out at the 2min exec ceiling (with partial starvation before that). Does that hold or am I missing something that prevents it?

if it does, would splitting into two queues (invocations on one, tools on the other) make sense?

PopDaph (approved):

front/temporal/sandbox_functions/config.ts
spolu: Well first thing IMHO is: make sure we don't starve the agent loop with invocations as we're building it. This PR achieves that :)

Then if I'm not mistaken, agent loop run tool is subject to same waits right? When I use the sandbox bash tool and it uses a dsbx tool call we do have the same timeout right?

spolu:
SHAs: 
-`front`/`front-edge`: `a51f5a0`

PR #29046, opened 2026-07-16
[front] enh: wrap readable.fromWeb to have better error msg

id13:
## Description

When AbortSignal.timeout fires while the response body of an upload-from-URL is being streamed to GCS, the resulting DOMException (TimeoutError) was emitted as an 'error' event on the Readable.fromWeb source stream. 

pipe() does not forward source errors to the destination, so nothing routed the error into the write pipeline: the write hung and the exception surfaced as an uncaught exception in the agent loop instead of a tool error.

**Fix: we will no longer log an uncaught exception when a transient upload error happens.**

- limitReadableStream now forwards source-stream errors into the transform (limited.destroy(err)), so an abort mid-stream rejects dustFs.write's internal pipeline and resolves to a normal Err.
- The Readable.fromWeb → limitReadableStream → dustFs.write sequence is wrapped in a try/catch returning Err({ message: "Failed to fetch URL: ..." }) as a double protection for anything else overflowing from node's internals.

## Tests

Added some tests

## Risk

Low
Blast radius: some file uploads codepaths around dust filesystem's upload from url. Mostly agent loop

## Deploy Plan

Deploy front

### DIFF
--- a/front/lib/api/file_system/upload_from_url.ts
+++ b/front/lib/api/file_system/upload_from_url.ts
@@ -62,6 +62,9 @@ function limitReadableStream(
   });
 
   limited.on("error", () => stream.destroy());
+  // pipe() does not forward source errors (e.g. a fetch abort mid-stream);
+  // left unforwarded they throw as uncaught exceptions.
+  stream.on("error", (err) => limited.destroy(err));
 
   return {
     stream: stream.pipe(limited),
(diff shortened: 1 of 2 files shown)


front/lib/api/file_system/upload_from_url.ts
```diff
@@ -177,36 +180,49 @@ export async function uploadFileFromUrlToFileSystem(
...
-    contentType: finalContentType,
-    sizeBytes,
-    existed,
-  });
+    return new Ok({
+      contentType: finalContentType,
+      sizeBytes,
+      existed,
+    });
+  } catch {
```
flvndvd, line 191: What are we aiming to catch with this very wide try/catch?

id13: Readable.fromWeb

id13: Will move up

id13:
PMRR

flvndvd (commented):
LGTM

PR #29052, opened 2026-07-16
Validate Pod function calls when publishing Frames

flvndvd:
## Description

Frame publishing now validates Pod function usage before persisting any published content. Each referenced function
must exist in the Frame’s current Pod, and the supplied input must match the function’s JSON Schema. Frames without
an accessible Pod scope are rejected when they attempt to use Pod functions.

Validation uses an in-memory TypeScript program containing the complete reachable source graph. This supports
references assembled through template literals, constants, imports, and aliases instead of relying on string
matching. Both stable function identifiers and Pod slug references are recognized.

This PR adds `json-schema-to-typescript` as a server-side dependency. It converts the stored function schemas into
TypeScript declarations used only during publishing. This avoids maintaining a partial custom JSON Schema converter
and preserves support for nested objects, required properties, arrays, unions, enums, and schema definitions.

Schema conversion is restricted to in-memory schemas. External HTTP and filesystem references are disabled,
preventing schemas from triggering network requests or local file reads. Conversion failures produce a dedicated
publishing error instead of bypassing validation.

The check runs after bundling, when all reachable sources are cached, but before content, bundle, metadata, or
allowlist writes. Function metadata is loaded in one batch. Diagnostics are grouped by source.

> [!NOTE]
> This will help but model is still crazy at times
> image

**Example:**
image


front/lib/api/viz/validate_frame_pod_functions.ts
```diff
@@ -0,0 +1,544 @@
...
+
+  return ts.ScriptKind.JS;
+}
+
+interface PodFunctionContract {
+  functionReferences: string[];
+  inputSchema: JSONSchema;
+}
+
+async function buildDustReactHooksDeclaration(
```
flvndvd, line 139: @fontanierh once return type is refactored, I'll add validation here as well.

front/lib/api/viz/validate_frame_pod_functions.ts
```diff
@@ -0,0 +1,544 @@
...
+      );
+      entries.push(
+        ...contract.functionReferences.map(
+          (functionReference) =>
+            `  ${JSON.stringify(functionReference)}: ${namespace}.${typeName};`
+        )
+      );
+    }
+  } catch (error) {
+    return new Err(
```
fontanierh: can we wrap only the compile statement ?

fontanierh (approved):

PR #29058, opened 2026-07-16
Fix test invocations.test.ts

PopDaph:
## Description

front-api/routes/w/[wId]/sandbox-functions/[functionId]/invocations.test.ts:718 was calling makeNew(adminAuth, { sandboxFunction }) without the now-required input. This is a classic two-greens-collide: #28983 (resolve-authentication, added that test) and the PR that made makeNew's input required each passed CI alone, but combine to break tsgo once both land. Added input: undefined (matching the pattern used elsewhere). Also fixed my untracked dev script (sandbox_create_blocked_action.ts:93) the same way. 

## Tests

CI

## Deploy Plan

Merge.


spolu (commented):
LGTM 🙏

PR #29059, opened 2026-07-16
feat(activation): add ActivationNudge model to track sent nudges

ElPicador:
## Description

Adds an `ActivationNudge` table (model + migration only, no behavior wired up yet) to record every time we nudge a user's Activation Pod.

We already have `ActivationRecommendation` to record the recommendations shown to a user during a conversation, but nothing tracks *when* a Pod was nudged. To decide whether to nudge a Pod again, the orchestrator needs that history plus some metadata about each send — this table is that record.

One row = one nudge sent to a Pod: `spaceId` (the Pod) and `triggerId` (the trigger that fired it), both required. `triggerId` is used instead of a conversation reference because it's known synchronously when the nudge is fired, and conversations already carry `triggerId`, so nudge → conversation is a simple join for later use (e.g. checking replies). Indexes are `(workspaceId, spaceId)` and `(workspaceId, triggerId)`, matching how this table will actually be queried (always workspace-scoped).

Follow-up PRs on this stack will add the frequency-cap logic (skip a Pod nudged within the last N days) and abandoned-detection.

## Tests

Verified locally: applied the migration against a real Postgres instance and confirmed the generated schema matches the model exactly (regenerating the diff afterward shows zero drift). `npx tsgo --noEmit` and `biome check` pass.

### DIFF
--- a/front/lib/models/activation/activation_nudge.ts
+++ b/front/lib/models/activation/activation_nudge.ts
@@ -1,0 +1,68 @@
+import { TriggerModel } from "@app/lib/models/agent/triggers/triggers";
+import { frontSequelize } from "@app/lib/resources/storage";
+import { DataTypes } from "@app/lib/resources/storage/data_types";
+import { SpaceModel } from "@app/lib/resources/storage/models/spaces";
+import { UserModel } from "@app/lib/resources/storage/models/user";
+import { WorkspaceAwareModel } from "@app/lib/resources/storage/wrappers/workspace_models";
+import type { CreationOptional, ForeignKey } from "sequelize";
+
+// One row = one nudge sent to a Pod (i.e. the activation trigger was fired for it).
+export class ActivationNudgeModel extends WorkspaceAwareModel<ActivationNudgeModel> {
+  declare createdAt: CreationOptional<Date>;
+  declare updatedAt: CreationOptional<Date>;
+
+  // The Pod that was nudged.
+  declare spaceId: ForeignKey<SpaceModel["id"]>;
+  // The trigger that fired the nudge.
+  declare triggerId: ForeignKey<TriggerModel["id"]>;
+  // The user targeted by the nudge.
+  declare userId: ForeignKey<UserModel["id"]>;
+}
+
+ActivationNudgeModel.init(
+  {
+    createdAt: {
+      type: DataTypes.DATE,
+      allowNull: false,
+      defaultValue: DataTypes.NOW,
+    },
+    updatedAt: {
+      type: DataTypes.DATE,
+      allowNull: false,
+      defaultValue: DataTypes.NOW,
+    },
+  },
+  {
+    modelName: "activation_nudge",
+    sequelize: frontSequelize,
+    indexes: [
+      { fields: ["spaceId"], concurrently: true },
+      { fields: ["triggerId"], concurrently: true },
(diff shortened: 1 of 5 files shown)


ElPicador:

- #29031
    - #29033
        - #29059 ◀
            - #29065
                - #29067
                    - #29103
                        - #29105
                        - #29117

front/lib/models/activation/activation_nudge.ts
```diff
@@ -0,0 +1,56 @@
...
+
+// One row = one nudge sent to a Pod (i.e. the activation trigger was fired for it).
+export class ActivationNudgeModel extends WorkspaceAwareModel<ActivationNudgeModel> {
+  declare createdAt: CreationOptional<Date>;
+  declare updatedAt: CreationOptional<Date>;
+
+  // The Pod that was nudged.
+  declare spaceId: ForeignKey<SpaceModel["id"]>;
+  // The trigger that fired the nudge.
+  declare triggerId: ForeignKey<TriggerModel["id"]>;
```
frankaloia, line 17: Do you think we need any user-based index to query by user we are targeting?

front/lib/models/activation/activation_nudge.ts
```diff
@@ -0,0 +1,56 @@
+import { TriggerModel } from "@app/lib/models/agent/triggers/triggers";
+import { frontSequelize } from "@app/lib/resources/storage";
+import { DataTypes } from "@app/lib/resources/storage/data_types";
+import { SpaceModel } from "@app/lib/resources/storage/models/spaces";
+import { WorkspaceAwareModel } from "@app/lib/resources/storage/wrappers/workspace_models";
+import type { CreationOptional, ForeignKey } from "sequelize";
+
+// One row = one nudge sent to a Pod (i.e. the activation trigger was fired for it).
+export class ActivationNudgeModel extends WorkspaceAwareModel<ActivationNudgeModel> {
```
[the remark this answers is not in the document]
ElPicador: added in another PR

front/lib/models/activation/activation_nudge.ts
ElPicador: Adding the userId, indeed that's a good idea!

xebeohp (approved):
LTGM

frankaloia (approved):
LGTM

front/lib/models/activation/activation_nudge.ts
```diff
@@ -0,0 +1,68 @@
...
+  declare createdAt: CreationOptional<Date>;
+  declare updatedAt: CreationOptional<Date>;
+
+  // The Pod that was nudged.
+  declare spaceId: ForeignKey<SpaceModel["id"]>;
+  // The trigger that fired the nudge.
+  declare triggerId: ForeignKey<TriggerModel["id"]>;
+  // The user targeted by the nudge.
+  declare userId: ForeignKey<UserModel["id"]>;
```
frankaloia, line 19: One note on this is that it adds into the model the assumption nudge == one user, which yes, that is absolutely what we are targeting, but 2 thoughts:
1. I can see this changing in the future for team activation
2. That invariant is not enforced by the pod itself. Can easily have multiple people added to these pods. I think this is _fine_ as we can send individual events for each user, so worth noting but not too worried

PR #29061, opened 2026-07-16
feat(front): allow connecting personal tools from user settings

adrsimon:
closes https://github.com/dust-tt/tasks/issues/7353 product feedback

## Description

Small QoL change to the Tools tab in personal settings. It used to only list tools you'd already connected or approved, now personal-auth tools available in the workspace show up there too, with a Connect button.

- Rows for not-yet-connected personal tools get a Connect button, wired to the same `useCreatePersonalConnection` flow used by the agent builder dialog and the blocked-action card.
- The row dropdown items are now conditional: "Clear confirmation preferences" only when there's approval history, "Disconnect" only when connected, no menu at all otherwise.
- On success the shared connections SWR key revalidates, so the row flips to the Connected chip without a refresh.

## Tests

Type-check + biome. Manual pass of the OAuth popup from the settings popover still to do before merge.

## Risk

Low.

## Deploy Plan

Deploy front.

### DIFF
--- a/front/components/me/UserToolsTable.tsx
+++ b/front/components/me/UserToolsTable.tsx
@@ -7,6 +7,7 @@ import { getAvatar } from "@app/lib/actions/mcp_icons";
 import type { MCPServerViewType } from "@app/lib/api/mcp";
 import type { MCPServerConnectionType } from "@app/lib/resources/mcp_server_connection_resource";
 import {
+  useCreatePersonalConnection,
   useDeleteMCPServerConnection,
   useManualMCPServerViewsFromSpaces,
   useMCPServerConnections,
@@ -18,6 +19,7 @@ import type { LightWorkspaceType } from "@app/types/user";
 import {
   Button,
   Chip,
+  CloudArrowLeftRight,
   DataTable,
   DotsHorizontal,
   DropdownMenu,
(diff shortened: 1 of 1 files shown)


front/components/me/UserToolsTable.tsx
```diff
@@ -162,6 +208,22 @@ export function UserToolsTable({ owner }: UserToolsTableProps) {
...
+              {!row.original.connection && row.original.canConnect && (
+                <Button
+                  icon={CloudArrowLeftRight}
+                  size="xs"
+                  variant="outline"
+                  label="Connect"
+                  disabled={connectingServerId !== null}
+                  isLoading={
+                    connectingServerId === row.original.serverView.server.sId
+                  }
```
ykmsd, line 220: I don't have any solution so I think it's okay but for whatever reason if you decide to abandon the auth flow you will be stuck in this state 

Image

ykmsd (approved):
Nice improvement 👍 I wonder if people want to filter out or sort by the tools you already connected vs not connected but can be a follow up if somebody ever complains

front/components/me/UserToolsTable.tsx
adrsimon: yeah... i think already the case when you abandon the auth flow in a conversation
will see if any brilliant idea come up later but for now i found nothing

LeandreLeBizec:
@adrsimon the goat, thanks for handling it!

PR #28699, opened 2026-07-08
feat(skills): handle private repo import with GitHub connection

Inesvvv:
## Description

Handles importing skills from private GitHub repos. Just the NO connection part for now. When a repo can't be reached and the workspace has no GitHub connection, `skills/detect` now returns `{ skills: [], githubConnectionRequired: true }` instead of an error, and the UI branches on role:

- **Admins** get a  "Connect GitHub" card that runs the workspace OAuth flow (shared, all-members connection) and reruns detection.
- **Builders** get a "Repository not found, check the URL or ask an admin to connect GitHub" message.

Switching tabs also clears the repo URL for a clean slate.

Admin: 
Screenshot 2026-07-08 at 17 13 53

Builder : 
Screenshot 2026-07-08 at 16 19 02

## Tests

Tested locally as both builder and admin for the UI but couldn't test the full flow due to OAuth issues 

## Risk

Low

## Deploy Plan

Deploy front and fronnt-api

PR #29062, opened 2026-07-16
feat(skills): show GitHub connection status and allow disconnect

Inesvvv:
## Description

Follow-up to #28699. Adds the connected-state UI for importing skills from a GitHub repository.

- Shows a connection status row (GitHub logo, "Connected" chip, and the avatar of who connected it) whenever the workspace has a GitHub connection.
- When connected but the account can't access the repo, shows "GitHub connection can't access this repository" , admins are prompted to reconnect, builders to ask an admin.
- Admins can disconnect via a "…" menu (builders can't).
- Removes the temporary MCP fallback from the connection read path now that this UI ships.

## Tests

Can't test locally. 

## Risk

Low

## Deploy Plan

Standard deploy of front


front/components/skills/import/ImportFromRepositoryTab.tsx
```diff
@@ -79,28 +88,62 @@ export function ImportFromRepositoryTab({
         disabled={isImporting}
         className="bg-muted-background"
       />
-      {repositoryNotFound ? (
-        isAdmin(owner) ? (
-          <ConnectWorkspaceGitHubMessage
-            owner={owner}
-            onConnected={() => triggerDetect(repoUrlField.value)}
+
+      {(() => {
```
aubin-tchoi: nit: can be rewritten without the IIFE

front/components/skills/import/ImportFromRepositoryTab.tsx
```diff
@@ -79,28 +88,62 @@ export function ImportFromRepositoryTab({
...
-        )
-      ) : (
-        <DetectedSkillsList
-          detectedSkills={detectedSkills}
-          isDetecting={isDetecting}
-          detectError={detectError}
+        );
+      })()}
+
+      {isConnected && connection && (
```
aubin-tchoi: nit: 
```suggestion
      {isConnected && (
```

front/lib/api/skills/github_connection.ts
```diff
@@ -39,5 +75,56 @@ export async function setWorkspaceGitHubConnection(
...
+  }
+
+  const user = await UserResource.fetchById(connection.connectedBy);
+  if (!user) {
+    return { connectedBy: null };
+  }
+
+  const userJSON = user.toJSON();
+  return {
+    connectedBy: { fullName: userJSON.fullName, imageUrl: userJSON.image },
```
aubin-tchoi: nit: no need to call `toJSON` to retrieve those metadata. `toJSON` should stay as close as possible to the interface with the front-end, the purpose of `toJSON` methods is specific to serializing data into objects for the front-end to consume

front/lib/api/skills/github_connection.ts
```diff
@@ -39,5 +75,56 @@ export async function setWorkspaceGitHubConnection(
...
+    action: "skill_import_github_connection.created",
+    targets: [buildAuditLogTarget("workspace", owner)],
+    context: getAuditLogContext(auth),
+    metadata: { connected_by: connectedBy },
+  });
+
+  return new Ok(undefined);
+}
+
+export async function getWorkspaceGitHubConnectionStatus(
```
aubin-tchoi: can't this be a method on `WorkspaceResource`?

front/lib/api/workspace.ts
```diff
@@ -593,6 +593,20 @@ export async function updateWorkspaceMetadata(
   return WorkspaceResource.updateMetadata(owner.id, newMetadata);
 }
 
+export async function removeWorkspaceMetadataKeys(
```
aubin-tchoi: should be on the resource

front/components/skills/import/ImportFromRepositoryTab.tsx
```diff
@@ -79,28 +88,62 @@ export function ImportFromRepositoryTab({
...
+        <GitHubConnectionRow
+          owner={owner}
+          connection={connection}
+          onDisconnected={() => {
+            void mutateConnection();
+            triggerDetect(repoUrlField.value);
```
aubin-tchoi: IIRC this one stores a ref to the last URL, maybe something to clean up here to make sure it does re-trigger

front/components/skills/import/ImportFromRepositoryTab.tsx
```diff
@@ -79,28 +88,62 @@ export function ImportFromRepositoryTab({
...
         className="bg-muted-background"
       />
-      {repositoryNotFound ? (
-        isAdmin(owner) ? (
-          <ConnectWorkspaceGitHubMessage
-            owner={owner}
-            onConnected={() => triggerDetect(repoUrlField.value)}
+
+      {(() => {
+        if (repositoryNotFound && isConnected) {
```
aubin-tchoi: `isConnected` will be false whiles it's loading, I suspect there can be a blink during the loading

front/lib/swr/github_connection.ts
```diff
@@ -1,10 +1,73 @@
...
+  disabled,
+}: {
+  owner: LightWorkspaceType;
+  disabled?: boolean;
+}) {
+  const { fetcher } = useFetcher();
+  const connectionFetcher: Fetcher<GetGitHubConnectionResponseBody> = fetcher;
+
+  const { data, error, mutate } = useSWRWithDefaults(
+    `/api/w/${owner.sId}/skills/github-connection`,
```
aubin-tchoi: nit: in terms of semantics, having the endpoint at `/api/w/:wId/skills/import/github-connection` would be more accurate

aubin-tchoi (commented):
left a few comments, all relative to coding style; everything great otherwise!

front/lib/resources/workspace_resource.ts
```diff
@@ -756,6 +758,54 @@ export class WorkspaceResource extends BaseResource<WorkspaceModel> {
     return this.updateByModelIdAndCheckExistence(id, { metadata });
   }
 
+  async removeMetadataKeys(keys: string[]): Promise<Result<void, Error>> {
```
aubin-tchoi, line 761: FYI this method does not update the current instance, which means that if you call this method, then your current instance will still have the old metadata

front/components/skills/import/ImportFromRepositoryTab.tsx
```diff
@@ -61,6 +70,53 @@ export function ImportFromRepositoryTab({
...
+    repositoryContent = (
+      <ContentMessage
+        variant="warning"
+        size="lg"
+        icon={InfoCircle}
+        title="GitHub connection can't access this repository"
+      >
+        The currently connected GitHub account can't access this repository.{" "}
+        {isAdmin(owner)
+          ? "Reconnect with an account that has access."
```
aubin-tchoi, line 92: nit: if I'm not mistaken, in this case we tell the user to reconnect with a button to disconnect

aubin-tchoi (approved):
LGTM!

PR #29063, opened 2026-07-16
Optimize seat sync

tdraier:
## Description

Optimizes the "Sync Metronome Seat Count" flow (`syncMetronomeSeatCountForWorkspace` → `syncSeatCount` → `reconcileWorkspaceUserCreditStates` → `syncDefaultPoolCapAlertsForWorkspace`), which could take 2+ minutes on workspaces with many staggered scheduled seat changes, and still took ~15.7s on a trivial workspace with nothing to change.

**Diagnostics added first**, to find the actual bottlenecks rather than guess:
- `workspaceId`-tagged `logger.info` calls at every phase boundary (staging a pending contract, the seat-state fetch, the transfer-prep step, each `reconcileSeatBasedSegment` call, the credit-state reconcile, the alert sync), each with a `durationMs`.
- `syncSeatCount` now returns a real `SyncSeatCountSummary` (subscription/timestamp/segment-call/transfer/free-user counts, whether anything mutated, duration) instead of `undefined`, threaded through `SeatSyncOutcome` into the poke plugin's displayed output — running the plugin now shows a multi-line breakdown instead of just "Metronome seat count synced."

**Fixes, found by reading the traces**:

- **`listMetronomeSeatBalances`** (`lib/metronome/client.ts`): `seatIds` is now mandatory — every call site already passed it explicitly, so the dead unfiltered-list code path (previously the root cause of a separate, already-fixed bug — see #28966) is fully removed. When a batch of ids fails (Metronome rejects the *entire* batch if even one id isn't found on the contract), recovery now bisects the batch instead of falling back to querying every id individually — isolates a handful of bad ids in ~log2(N) rounds of batched calls instead of N single-id calls. The outer per-chunk loop now runs concurrently instead of sequentially.

- **`emptyOriginSeatCreditsForTransfers`** (`lib/metronome/seats.ts`): reordered to fetch Metronome's current per-subscription seat state *before* the balance fetch, and skip the balance fetch entirely when there are no actual pro/max seat-type transfer candidates (the common case). When there are candidates, the balance fetch is now scoped to just the transferring subset instead of every eligible user.

- **Eliminated a duplicate Metronome call**: `getMetronomeSubscriptionAssignedSeatIds` was a thin wrapper over `getMetronomeSubscriptionSeatState` — the transfer-prep step and the main reconcile loop's "immediate base" pass were each independently fetching the exact same "now" state per subscription. It's now fetched once in `syncSeatCount` and shared via a `cachedSeatState` param on `reconcileSeatBasedSegment`, extended to also cover the "free" subscription (previously never cached, since transfer-prep never looked at it). With that, the wrapper had no remaining reason to exist beyond one call site (`buildSeatDataByUserId`), so it's been deleted entirely and that call site now uses `getMetronomeSubscriptionSeatState` directly, matching the rest of the file.

- **Free-seat credit revoke**: `revokeFreeSeatCreditsForExFreeUsers`'s `listCustomerPerUserCreditBalances` call is expensive and previously ran unconditionally on every sync. It's now gated on a cheap, already-fetched signal — Metronome's actual "free" seat assignment list vs. our desired free users — and only runs when someone has actually moved away from a free seat. This is a deliberate tradeoff: a missed revoke (e.g. contract dropped "free" entitlement between an old credit and now) is low-stakes and self-corrects on the next sync where it happens to run; it isn't worth an unconditional credit-listing call on every membership operation.

- **`reconcileWorkspaceUserCreditStates`** (`lib/api/metronome/reconcile_credit_state.ts`): two groups of previously-sequential, mutually-independent reads now run via `Promise.all` — (1) seat allowances, credit-usage config, and active memberships; (2) seat balances, per-user free credit balances, per-user usage, and max group caps (all only depend on the memberships from group 1, not on each other).

- **`syncDefaultPoolCapAlertsForWorkspace`** (`lib/api/workspace/default_user_spend_limit.ts`): the 4 independent cap/warning alert upserts (2 seat types × 2 alert kinds) now run concurrently via `concurrentExecutor` instead of one at a time. The cap-vs-warning branching there is an exhaustive `switch` + `assertNever` on the task kind rather than if/else, per [GEN6].

## Tests

- Existing unit tests for all touched files pass (`client.test.ts`, `seats.test.ts`, `seats_sync.test.ts`, `reconcile_credit_state.test.ts`, `live_user_credit_inputs.test.ts`, `default_user_spend_limit.test.ts` — 105 tests).
- Manually verified against a live production workspace by running the "Sync Metronome Seat Count" poke plugin and reading the new phase-by-phase logs before/after each fix:
  - Baseline (nothing to change): ~15.7s total (`emptyOriginSeatCreditsForTransfers` ~3s, 3× `reconcileSeatBasedSegment` ~2.4s combined, free-seat grant/revoke ~3.2s, `reconcileWorkspaceUserCreditStates` ~5.4s, alert sync ~1.7s).
  - After all fixes, a run that included a real seat-count change: ~7.2s total (`reconcileWorkspaceUserCreditStates` ~2.5s, alert sync ~0.3s, `syncSeatCount` down to only the unavoidable upfront seat-state fetch plus the one real write).
  - `Promise.all`-related `pg` "client already executing a query" deprecation warnings observed in the local test run are a test-harness artifact (each test runs inside one shared Sequelize transaction via CLS, per `vite.setup.ts`) — not present in production, where the connection pool hands out separate connections to genuinely concurrent queries.

## Risk

- All changes are either pure performance optimizations with no behavioral difference (caching a value already being fetched, running mutually-independent calls concurrently instead of sequentially) or the one deliberate, discussed tradeoff on free-credit revoke above.
- No schema or migration changes, no feature flag.
- Safe to roll back: reverts to the previous (slower, but already-correct) behavior.

## Deploy Plan

Standard deploy. No migration, no config change. Once live, re-run "Sync Metronome Seat Count" from poke on a previously-slow workspace and check the new logs/plugin output to confirm the improvement.

### DIFF
--- a/front/lib/api/metronome/seat_sync.ts
+++ b/front/lib/api/metronome/seat_sync.ts
@@ -9,6 +9,7 @@ import { getActiveContract } from "@app/lib/metronome/plan_type";
 import {
   hasContractSeatSubscription,
   remapMembershipSeatTypesForContract,
+  type SyncSeatCountSummary,
   syncSeatCount,
 } from "@app/lib/metronome/seats";
 import { isProPlanPrefix } from "@app/lib/plans/plan_codes";
@@ -23,10 +24,17 @@ import type { LightWorkspaceType } from "@app/types/user";
  * Outcome of `syncMetronomeSeatCountForWorkspace`. `synced` means the
  * reconciliation ran; `skipped` means there was nothing to sync (not on
  * Metronome, no active contract, or no seat subscription) — with a
- * human-readable `reason`.
+ * human-readable `reason`. `activeContractSummary` is `null` only when
+ * nothing ran for the active contract (e.g. a pending contract was staged
+ * but there's no active contract to sync yet).
  */
 export type SeatSyncOutcome =
-  | { status: "synced" }
+  | {
+      status: "synced";
+      stagedPendingContract: boolean;
+      activeContractSummary: SyncSeatCountSummary | null;
+      workspaceUserCreditStatesReconciled: boolean;
+    }
   | { status: "skipped"; reason: string };
 
 /**
@@ -48,14 +56,22 @@ export type SeatSyncOutcome =
  * user (e.g. an auto-upgrade unblocking one member mid-flow) and skips the
  * workspace-wide cap-alert sync. The seat-count push itself is always
  * workspace-wide. When omitted, the whole workspace is reconciled.
+ *
+ * `forceFreeCreditRevokeCheck` runs the ex-free-seat credit revoke check
+ * unconditionally instead of only when the cheap gate signals a change (see
+ * `syncSeatCount`). Set by the poke plugin, where an operator explicitly
+ * asked for a thorough pass; left unset on the automatic/debounced path.
  */
 export async function syncMetronomeSeatCountForWorkspace({
   workspace,
   reconcileUserId,
+  forceFreeCreditRevokeCheck,
 }: {
   workspace: LightWorkspaceType;
   reconcileUserId?: string;
+  forceFreeCreditRevokeCheck?: boolean;
 }): Promise<Result<SeatSyncOutcome, Error>> {
+  const workspaceId = workspace.sId;
   if (!workspace.metronomeCustomerId) {
     return new Ok({
       status: "skipped",
@@ -75,10 +91,19 @@ export async function syncMetronomeSeatCountForWorkspace({
   // contract stays correct across adds, revokes, and seat changes alike. This
   // is independent of, and in addition to, reconciling the active contract
   // below (whose seats a legacy shadow contract still bills / finalizes).
+  const stagePendingStartedAt = Date.now();
   const stagedPending = await stagePendingContractSeats({
     workspace,
     currentPlanCode: activeSubscription?.getPlan().code ?? null,
   });
+  logger.info(
+    {
+      workspaceId,
+      stagedPending,
+      durationMs: Date.now() - stagePendingStartedAt,
+    },
+    "[SeatSync] stagePendingContractSeats done"
+  );
 
   const activeContract = activeSubscription?.metronomeContractId
     ? await getActiveContract(workspace.sId)
@@ -89,15 +114,26 @@ export async function syncMetronomeSeatCountForWorkspace({
     !activeContract ||
     !(await hasContractSeatSubscription(activeContract))
   ) {
-    return new Ok(
-      stagedPending
-        ? { status: "synced" }
-        : {
-            status: "skipped",
-            reason:
-              "no active or pending contract with a seat subscription to sync",
-          }
+    if (stagedPending) {
+      logger.info(
+        { workspaceId },
+        "[SeatSync] Done — only the pending contract was staged, no active contract to sync"
+      );
+      return new Ok({
+        status: "synced",
+        stagedPendingContract: true,
+        activeContractSummary: null,
+        workspaceUserCreditStatesReconciled: false,
+      });
+    }
+    logger.info(
+      { workspaceId },
+      "[SeatSync] Skipped — no active or pending contract with a seat subscription to sync"
     );
+    return new Ok({
+      status: "skipped",
+      reason: "no active or pending contract with a seat subscription to sync",
+    });
   }
 
   const result = await syncSeatCount({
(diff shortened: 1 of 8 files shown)


tdraier:

- #29063 ◀

front/lib/metronome/client.ts
```diff
@@ -3756,12 +3727,36 @@ export async function listMetronomeSeatBalances({
       }
     };
 
+    // When a batch fails, isolate the bad seat_id(s) by bisecting rather
+    // than querying every id individually. Metronome rejects a batch
+    // entirely if even one id in it isn't found on the contract, so with a
+    // large batch and only a handful of bad ids, halving repeatedly finds
+    // them in ~log2(N) rounds of (still-batched) calls instead of N
+    // individual single-id calls.
+    const fetchSeatIdsBisect = async (
```
PopDaph, line 3745: When a batch fails, the two recursive calls run through Promise.all, so every failing level doubles the number of Metronome calls in flight, there's no cap. If a chunk fails for something that isn't a bad seat_id (429, 5xx, Metronome incident), it bisects all the way down and we could end up with ~100 single-id calls in flight for one chunk, x4 with the outer concurrency: 4, right when Metronome is already struggling. The old fallback was capped at 8 concurrent calls. Does that hold or am I missing something that bounds it?

If it does, would running the two recursive calls one after the other make sense? We'd keep the log2(N) rounds when only a few ids are bad, and there'd never be more than one call in flight per chunk. WDYT?

front/lib/api/metronome/seat_sync.ts
```diff
@@ -56,7 +64,17 @@ export async function syncMetronomeSeatCountForWorkspace({
   workspace: LightWorkspaceType;
   reconcileUserId?: string;
 }): Promise<Result<SeatSyncOutcome, Error>> {
+  const workspaceId = workspace.sId;
```
PopDaph, line 74: NIT: iiuc this function runs on every membership change (join, leave, seat change), for every workspace, including the ones that aren't on Metronome at all. So these two logs ("starting" and "skipped, not provisioned") will spam for basically every membership event in prod, and they don't tell us much. Could we keep only the logs that carry a durationMs + the final summary, and drop these? 🙏🏻

front/lib/metronome/seats.ts
```diff
@@ -1748,20 +1901,69 @@ export async function syncSeatCount({
...
+    )?.sub.id;
+    const freeSeatState = freeSubscriptionId
+      ? seatStateBySubscriptionId.get(freeSubscriptionId)
+      : undefined;
+    const movedAwayFromFree = freeSeatState
+      ? freeSeatState.assignedSeatIds.filter(
+          (id) => !currentFreeUserIds.has(id)
+        )
+      : [];
+    if (movedAwayFromFree.length > 0) {
```
PopDaph: Question on the revoke gate: the check now only runs when we can see a user still assigned to a free seat in Metronome but no longer free in our DB. 

But if a run crashes after the reconcile loop (which removes their free seat in Metronome) and before the revoke, the next run sees nothing: the seat is already gone on Metronome's side, so the gate stays closed and their free credit is never cleaned up, until some other user happens to leave a free seat later. 

Since that credit grants real AWU, is that window okay? 

If not, maybe keep the unconditional check on the poke plugin path only (operator-triggered, so rare), so there's always a way to catch up without paying the cost on every membership change. 

WDYT? Does that makes sense or not?

PopDaph (approved):

front/lib/metronome/client.ts
tdraier: good catch, it can lead to multiple parallel calls. Serializing the 2 calls wouldn't hurt.

front/lib/metronome/seats.ts
tdraier: yes it can happen, but as explained in the description, keeping the credit has no impact - it can only be used by the user when they have "free" seat in dust. Revoking it is only to make it cleaner but it's not worth the cost. We could potentially add it to the poke path but I'm not sure it's worth it either. I was actually more thinking of removing entirely the revoke path to make code simpler.

front/lib/metronome/client.ts
tdraier: I've also added another optimization, as the error message does contain the first id that breaks - which can be immediately be excluded

front/lib/metronome/seats.ts
tdraier: Can still be ok to add this in poke - revoking the credits is cleaner and it also remove some metronome alerts

PR #29064, opened 2026-07-16
fix(scripts): surface spawn errors in run-migrate, install devDeps in prodbox

ElPicador:
## Description

`scripts/db/run-migrate.cjs` swallowed `spawnSync` failures silently (e.g. when
`tsx` isn't on `PATH`, `result.error` was set but never checked, so the script
just exited with an unhelpful status code). This surfaces `ENOENT` and other
spawn errors, plus signal termination, with an actionable message.

Since `tsx` is a devDependency, this also updates the prodbox image to install
devDependencies (`npm ci --include=dev`) so the migration script can actually
find `tsx` when running against uncompiled TypeScript.

## Tests

Verified locally: running the script with `tsx` removed from `PATH` now prints
a clear error and exits 1, instead of failing without explanation.

### DIFF
--- a/dockerfiles/prodbox.Dockerfile
+++ b/dockerfiles/prodbox.Dockerfile
@@ -14,8 +14,9 @@ RUN npm install -g npm@11.11.0
 
 COPY . .
 
-# Install dependencies
-RUN --mount=type=cache,id=npm-cache,target=/root/.npm npm ci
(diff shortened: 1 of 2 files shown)


ElPicador:

- #29064 ◀

tdraier (approved):
🙏

PR #29065, opened 2026-07-16
feat(activation): return trigger id from emitActivationEvent

ElPicador:
## Description

`emitActivationEvent` currently returns `void` on success. To link an `ActivationNudge` row (added in #29059) to the trigger that fired it, we need that trigger's sId at the call site.

`processWebhookRequest` already computes the matched trigger(s) internally via `filterTriggers` when it processes the activation webhook event, so this just surfaces that id through the return value instead of the caller having to look it up again. `emitActivationEvent` now returns `{ triggerId: string | null }` (a pod has at most one activation trigger, via `activationTriggerFilter`), and `processWebhookRequest` returns `{ triggerIds: string[] }` for the shared webhook-processing path.

No existing call sites needed changes since they only checked `.isErr()`.

## Tests

Pure return-type plumbing; verified with `npx tsgo --noEmit` and `biome check` on the touched files. No new tests needed — behavior is unchanged, only the return value is now populated.

### DIFF
--- a/front/lib/api/activation/trigger.ts
+++ b/front/lib/api/activation/trigger.ts
@@ -158,11 +158,13 @@ export async function createActivationTrigger(
   return new Ok({ triggerId: triggerRes.value.sId });
 }
 
-// Fires the activation trigger for a single pod by emitting an internal webhook event.
+// Fires the activation trigger for a single pod by emitting an internal webhook
+// event. Returns the sId of the pod's activation trigger, if the event matched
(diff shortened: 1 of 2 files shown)


ElPicador:

- #29031
    - #29033
        - #29059
            - #29065 ◀
                - #29067
                    - #29103
                        - #29105
                        - #29117

frankaloia (approved):
LGTM

PR #29070, opened 2026-07-16
fix(front): reorder Auto toggle above search bar in model picker

Nils-Fedrigo:
## Description

Reorders the model picker content in the input bar so the **Auto** toggle sits at the top, above the search bar. The search bar is now hidden while Auto is on, since there is no model to pick in that state (it still shows during the searching state, when the Auto row is hidden).

## Tests

Manual: open the model picker from the input bar, toggle Auto on/off, and confirm the Auto row renders above the search bar and the search bar disappears when Auto is enabled.

## Risk

Low — presentation-only change to `ModelPickerContent`.

## Deploy Plan


Nils-Fedrigo:

- #29070 ◀

Fraggle (commented):
LGTM

PR #29072, opened 2026-07-16
Add user id to invocation model

PopDaph:
## Description

First of 3 PRs adding the initiating user to sandbox function invocations, so we can enforce that only the invoker resolves their own tool auth/approval (fixes the confused-deputy gap flagged on #28983).

- **PR1 (this):** add nullable `userId` FK on `SandboxFunctionInvocationModel` + migration. **Unused** for now, purely additive and deployable on its own.
- **PR2:** populate `userId` at `makeNew` from `auth.user()`.
- **PR3:** enforce resolver == invoker in `resolve-authentication` and `validate-action` via `canCurrentUserRespondToParentUserMessage`.

**Notes:**
- Nullable forever: API-key / scheduled / bot invocations have no human, so it stays nullable (no future non-null backfill).
- `ON DELETE SET NULL`: an invocation is a workspace operation record that should outlive its triggerer. Also required because the action -> invocation FK is `RESTRICT`, so a `CASCADE` from the user would be blocked / bypass GCS cleanup. Same pattern as `resolvedByUserId` / `generatedByUserId`.
- Index on `userId` (single column, not `workspaceId`-first) because it serves the `ON DELETE SET NULL` scan on user deletion, which is not workspace-scoped.

## Tests

- Front TypeScript check
- Migration applied locally, verified column (nullable), index, and FK (`SET NULL`, validated)

## Risk

Worst case, an unused nullable column + index. Safe to rollback.

## Deploy Plan

- [ ] Apply pre-deploy migration
- [ ] Deploy `front`

### DIFF
--- a/front/lib/resources/storage/models/sandbox_function.ts
+++ b/front/lib/resources/storage/models/sandbox_function.ts
@@ -5,6 +5,7 @@ import {
 } from "@app/lib/resources/storage/data_types";
 import { FileModel } from "@app/lib/resources/storage/models/files";
 import { SpaceModel } from "@app/lib/resources/storage/models/spaces";
+import { UserModel } from "@app/lib/resources/storage/models/user";
 import { WorkspaceAwareModel } from "@app/lib/resources/storage/wrappers/workspace_models";
 import { validateJsonSchema } from "@app/lib/utils/json_schemas";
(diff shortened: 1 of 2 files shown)


spolu (approved):
LGTM

PR #29073, opened 2026-07-16
[Admin Gov] Store role names as group_permissions grant types

philipperolet:
## Description

Switches what a `group_permissions` grant row stores from a plain **verb** to a registry-defined
**role name**. Follows #29047 (which reshaped the registry into role definitions).

- **Vocabulary** (`types/group_permissions.ts`): split into two sets.
  - `GRANT_VERBS` — the granular capabilities (`read`/`write`/`admin`/`create`/`publish`/`invite`/`use`).
    Never stored; used only to describe what a role grants and to ask capability questions.
  - `GRANT_TYPES` — the role names actually stored: `reader`/`member`/`admin` (space), `editor`
    (agent+skill), and the singleton capabilities `create`/`publish`/`invite`/`admin`/`read`/`use`
    (+ `"*"`).
- **Validation** (`group_permission_registry.ts`): `assertValidGrant` now checks the grant type is a
  role defined for the resource type at the required level (instance vs type-wide), instead of
  deriving a verb set.

**No schema change and no live-writer change.** The only grant types stored today are singletons —
`models_tier` writes `use`; the governance capabilities are `create`/`publish`/`invite`/`admin` —
whose role name equals the verb, so their stored values are unchanged.

### Why this matters (revoke-collision)

Additive multi-grant is a design invariant: a group can hold more than one role on the same
resource. Storing the role name keeps overlapping roles as distinct rows, so revoking one never
destroys verbs the group still holds via another (e.g. revoking `member` leaves an `admin` grant —
and its `read`/`write` — intact). Under verb storage those roles would collapse onto shared
`read`/`write` rows and revoking `member` would delete them. Adds a regression test for exactly this.

### Deferred

Verb→role expansion at check time is intentionally **not** in this PR. The only check today,
`auth.hasWorkspacePermission`, is type-wide, and every type-level role is a singleton (name == verb),
so it matches the stored value directly with no expansion. Expansion arrives with the first
instance-level, multi-verb capability check that needs it.

## Risks

Blast radius: `group_permissions` grant validation and the stored grant-type vocabulary. No schema
change; no change to any live writer's stored values.
Risk: low

## Deploy Plan

- deploy front + front-sse (no migration)

### DIFF
--- a/front/lib/auth.ts
+++ b/front/lib/auth.ts
@@ -1099,7 +1099,7 @@ export class Authenticator {
    * Whether the caller holds a workspace-level capability. A capability is a
    * (grantType, resourceType) pair whose grants live on the type-wide (-1) group_permissions
    * rows. Admins bypass unconditionally (billing/security are admin-by-default). Otherwise we look
-   * for a -1 grant on any of the caller's groups; "*" grants match any verb / type.
+   * for a -1 grant on any of the caller's groups; "*" grants match any grant type / resource type.
    *
    * Cold path: a query per check is fine — no caching yet (pending auth-resolution decision).
    */
--- a/front/lib/auth.workspace_permission.test.ts
+++ b/front/lib/auth.workspace_permission.test.ts
@@ -111,9 +111,9 @@ describe("Authenticator.hasWorkspacePermission", () => {
   });
 
   it("rejects an invalid capability query, even for admins", async () => {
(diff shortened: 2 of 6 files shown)


philipperolet:

- #29011
    - #29047
        - #29073 ◀

matteotrab (approved):

PR #26375, opened 2026-05-28
 Support canonical scoped paths in conversation file utils

flvndvd:
## Description

External MCP tools (GDrive, Slack, image generation, Jira, etc.) that need to read conversation files call `resolveConversationFileRef` and `getFileFromConversationAttachment`. These functions currently only handle legacy scoped paths (`conversation/file.txt`) and raw file IDs (fil_xxx).

As the files MCP server migrates its tools to `DustFileSystem`, it will start producing canonical scoped paths (`conversation-{id}/file.txt`, `pod-{id}/file.txt`). Without this change, external tools receiving those paths from the agent would fail silently or return errors.

This PR adds a fast-path for canonical scoped paths in both functions. Canonical paths are routed through `DustFileSystem.fromScopedPath`, which resolves the conversation or pod by ID directly. No ambient conversation context required. The existing legacy path and file ID resolution paths are unchanged.

PR #29074, opened 2026-07-16
mcp file utils: fall back to FileResource when a canonical scoped path is missing from GCS

smb2268:
## Description

Fixes dust-tt/tasks#9648.

MCP tools (gmail, google_drive, jira, microsoft_drive, outlook, slack, file_generation, image_generation) fail with `File not found` when the model references a **user-uploaded** conversation file by its **canonical scoped path** (`conversation-{id}/report.pptx`).

**Root cause:** the model sees canonical scoped paths because content fragments render `path` from `FileResource.mountFilePath` — but user uploads are stored at the FileResource canonical key (`files/w/{wId}/{fileId}/...`) and may never have been copied to the deterministic mount path `DustFileSystem` reads from. `resolveConversationFileRef` and `getFileFromConversationAttachment` resolved canonical paths exclusively through the `DustFileSystem` stat, so the miss returned `File not found` without ever trying the pre-existing `FileResource`-based resolution (regression from `#26375`, which added the canonical-path branch without a fallback).

**What this PR does:**

1. On a canonical-path stat miss, resolve the linked `FileResource` via the existing `fetchLinkedFileResource` helper (`lib/api/files/file_system_ops.ts`, same bridge the files move tool uses) and serve the original version from there. Only ready files are served; a genuine miss keeps the original `File not found` error.
2. Normalize the scoped path before the lookup, so the fallback resolves the same key `stat()` just missed on inputs with `.`/`..` segments.
3. Drive-by 🚗: extract `makeFileRefFromResource` so the fallback and the legacy fileId branch share the ref shape.

> [!NOTE]
> The native `files` tool and the `/files/path/[...canonicalPath]` route read through `DustFileSystem` directly and have the same gap — intentionally out of scope here, follow-up to be filed.

Confirmed in prod logs (workspace `35WFegvIrd`, conversation `wRRqT5Izdf`): `microsoft_drive__upload_file` and the native `files` tool both fail on the same canonical path.

## Tests

- New functional tests for the fallback in both functions (real factory-created files whose mount path is claimed in the DB but absent from GCS), plus a test that the path is normalized before translation.
- Existing genuine-miss tests still assert `File not found`.
- `npx tsgo --noEmit` + `npx biome check` clean.

## Risk

Worst case, a canonical path whose GCS mount copy was deleted via `files__delete` resolves to the original version instead of `File not found`. This is consistent with existing behavior, not a resurrection: `files__delete` only removes the GCS object (the `FileResource` row and the content fragment remain), so the file is still a live conversation attachment and legacy `fil_xxx` references already serve it. The coherent fix for delete semantics is wiring the existing-but-uncalled `deleteCanonicalFile` (which deletes the linked `FileResource`) into the delete tool — follow-up, out of scope here. Safe to rollback.

## Deploy Plan

- [ ] Deploy front

### DIFF
--- a/front/lib/actions/mcp_internal_actions/utils/file_utils.test.ts
+++ b/front/lib/actions/mcp_internal_actions/utils/file_utils.test.ts
@@ -42,6 +42,8 @@ vi.mock("@app/lib/api/file_system", async (importOriginal) => {
     DustFileSystem: {
       ...actual.DustFileSystem,
       fromScopedPath: mockFromScopedPath,
+      // Class statics are non-enumerable, so the spread above does not carry them over.
+      normalizeScopedPath: actual.DustFileSystem.normalizeScopedPath,
     },
   };
 });
@@ -110,7 +112,7 @@ describe("getFileFromConversationAttachment", () => {
     });
 
     it("returns Err when the file is not found", async () => {
-      const { authenticator: auth } = await createResourceTest({
+      const { authenticator: auth, workspace } = await createResourceTest({
         role: "admin",
       });
 
@@ -127,7 +129,7 @@ describe("getFileFromConversationAttachment", () => {
           toMountFilePath: vi
             .fn()
             .mockReturnValue(
-              `w/unknown/conversations/${conversation.sId}/files/missing.pdf`
+              `w/${workspace.sId}/conversations/${conversation.sId}/files/missing.pdf`
             ),
         })
       );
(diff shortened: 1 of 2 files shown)


smb2268:
Closing: built on a wrong root cause. The incident's object was present at the deterministic mount path all along — the failing calls retyped the filename with an ASCII apostrophe (U+0027) where the stored name has a typographic one (U+2019), so a FileResource fallback keyed off the same mistyped string returns the identical error. Superseded by `29511` (surface `fil_xxx` ids in listings, prefer ids in tool descriptions, near-match recovery on path misses).

PR #29077, opened 2026-07-16
feat(front): attach files to code-defined skills; ship activation pod Frame template

frankaloia:
## Summary

Adds the ability for **code-defined (global/system) skills** to ship files, and uses it to attach the day-zero **pod overview Frame template** to the Activation skill.

- A global skill definition can now declare `files: [{ fileName, contentType, content }]` (`CodeDefinedSkillFile`).
- On enable, these files are materialized into the conversation file system under `skills/<skillName>/<fileName>` — through the **same** `loadSkillFilesToConversation` path that custom-skill `FileResource` attachments already use.
- The loader now normalizes **both** sources into one `WritableSkillFile` list and writes them in a single loop, so any future global skill can attach files with no extra wiring:
- **Activation skill**: attaches `pod_frame_template.tsx` (bundled in `static_files/`) and instructs the skill to build the pinned pod overview Frame from that template rather than authoring Frame code from scratch.
- This template tsx file is just downloaded from a real Frame i created

Not too concerned on exact frame content yet, we can iterate as we go

## Risk

* Intended to be low and additive

## Testing
Screenshot 2026-07-16 at 6 23 34 PM

## Deplyoment
1. Deploy front

### DIFF
--- a/front/lib/resources/skill/code_defined/global/static_files/activation_pod_frame_template.ts
+++ b/front/lib/resources/skill/code_defined/global/static_files/activation_pod_frame_template.ts
@@ -1,0 +1,367 @@
+export const ACTIVATION_POD_FRAME_TEMPLATE = `import { useEffect, useState } from "react";
+import {
+  Sparkles,
+  ArrowRight,
+  Search,
+  MessageCircleQuestion,
+  Gift,
+  Users,
+  Building2,
+  Lightbulb,
+  Bot,
+  Wrench,
+} from "lucide-react";
+
+/* ============================================================
+   ACTIVATION POD FRAME v7
+   Design sources: beautiful-frames-design skill (icons not
+   emojis, perfect spacing, premium minimalist), Henry's
+   failure-mode analysis (committed visual identity: one accent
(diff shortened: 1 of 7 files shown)


front/lib/resources/skill/code_defined/global/static_files/activation_pod_frame_template.ts
```diff
@@ -0,0 +1,364 @@
+export const ACTIVATION_POD_FRAME_TEMPLATE = `
```
ElPicador, line 1: I'm a bit worried we break this tsx file very quickly. 

I would add a unit test to compile it (or at least check syntax) in the CI

Probably [one of those](https://app.dust.tt/w/0ec9852c2f/conversation/xymhfStX43) could work

ElPicador (approved):

front/lib/resources/skill/code_defined/global/static_files/activation_pod_frame_template.ts
frankaloia: Will do pre-merge

PR #9623, opened 2024-12-23
cleanup(GitHub) - remove old long-running workflows

aubin-tchoi:
## Description

- Follow up on https://github.com/dust-tt/dust/pull/9562
- Removes the old workflows that were kept to avoid non-deterministic errors.
- I will consider renaming the new ones to remove the v2 in an upcoming PR.

## Risk

- Low

## Deploy Plan

- Deploy connectors.

PR #29078, opened 2026-07-17
microsoft: skip worksheets with a previously persisted skip reason instead of re-fetching

smb2268:
## Description

Fixes part of #9623.

Microsoft Graph's Excel API returns a `504 MaxRequestDurationExceeded` when it can't render a specific worksheet's content in time (independent of sheet count — a 61-sheet workbook can be far worse than a 484-sheet one, depending on formula/used-range complexity per tab). We already persist a `skipReason` on the worksheet's `MicrosoftNodeResource` when this happens, but nothing checked that flag before re-attempting the same worksheet on a later sync.

That gap meant every retry of the surrounding delta batch re-ran the full worksheet loop from scratch, re-paying the ~90s Graph timeout on every already-known-bad worksheet, every time. In the incident behind #9623 this multiplied a single file's ~80-minute worst case into a batch that needed 4–8 retries over several hours before finally completing within the 120-minute activity timeout.

This PR adds the missing check: before calling `processSheet` for a worksheet, `handleSpreadSheet` now looks up the already-fetched `syncedWorksheets` list (no extra API/DB call) for an existing `skipReason` and skips re-fetching if found, while keeping the worksheet out of the deletion pass so its skip record is preserved.

Not included here (parked, see issue discussion):
- Any change to `batchSize` in `processDeltaChangesFromGCS`.
- A per-file sheet-count cap — checked against real data, wouldn't have prevented this incident.
- A TTL/retry-after-N-days policy for clearing a persisted skip (currently permanent until a full resync).

## Tests

Manual review + local diff inspection only; no local build/test run in this environment (no local toolchain available to run the connectors test suite). Change is isolated to `handleSpreadSheet` in `connectors/src/connectors/microsoft/temporal/spreadsheets.ts`, brace/paren balance checked against the pre-existing file. Relying on PR CI for lint/type-check/test validation.

## Risk

Low, contained to the Microsoft connector's spreadsheet worksheet loop. Failure mode if something is wrong: a worksheet that was previously skipped due to a transient failure would continue to be skipped instead of retried, which is a slight regression in optimism but not a correctness risk to other data. Safe to roll back (revert this diff) if unexpected behavior appears.

## Deploy Plan

Standard deploy through the connectors pipeline. No migration or config change needed. Suggest monitoring `workflow_timeout_failure` counts for `provider:microsoft` in Datadog after rollout to confirm the retry-storm pattern (see #9623) stops recurring.


id13 (approved):

adrsimon (approved):

adrsimon:
merging as per https://dust4ai.slack.com/archives/C050SM8NSPK/p1784258161467579?thread_ts=1784258109.138279&cid=C050SM8NSPK

PR #18785, opened 2025-12-08
Databricks mcp scaffolding

FlagBenett:
## Description

Scaffolding the databricks mcp in front
- behind a feature flag
- basic architecture and 1 tool (list warehouses)
- personal or shared oauth
- Oauth provider including client_id, client_secret and databricks_workspace

## Tests

locally

## Risk

low

## Deploy Plan

front only

PR #14414, opened 2025-07-17
scaffolding the jira MCP

FlagBenett:
## Description

We are building an internal JIRA MCP server, which is a wrapper around [JIRA's REST API v3](https://developer.atlassian.com/cloud/jira/platform/rest/v3/intro/#version)
We are using an [Oauth app](https://developer.atlassian.com/cloud/jira/platform/oauth-2-3lo-apps/), running on Dust's jira account. 
The app must be published from the [developer console](https://developer.atlassian.com/console/myapps/) so that users can use it.

### Other options considered

Adding the [official JIRA MCP](https://mcp.atlassian.com/v1/sse) as a remote MCP server is currently blocked by a client whitelisting process

# Features

Scaffolding for the JIRA internal MCP
- implementation of the jira_tool feature flag
- config options : personal credentials OR workspace credentials
- MCP server with only one basic tool for now : get_issue (to get a single JIRA issue)
- generic JiraAPICall wrapper function to all jira api endpoints for extension later
- helper functions (getJiraResourceInfo) to retrieve accessible resources from the authentication token

New environment variables OAUTH_JIRA_CLIENT_ID and OAUTH_JIRA_CLIENT_SECRET variables are required to enable the feature globally

## Risk

- sdk update required

## Deploy Plan

Deploy front

PR #29079, opened 2026-07-17
feat: scaffold ServiceNow MCP server integration

thomasvicaire:
## Description

Adds the initial skeleton for a ServiceNow internal MCP server, following the same pattern as the Databricks/Jira integrations:

- Instance-URL-based OAuth provider (Rust `core` + `front`), modeled on the Databricks provider since both are self-hosted/dedicated-instance platforms requiring a per-tenant OAuth app (no shared multi-tenant "Dust app" is possible here, unlike Slack/Notion/Google)
- A single skeleton tool, `list_incidents`, calling ServiceNow's Table API
- Full registration plumbing: `servicenow_tool` feature flag (off by default, `isPreview: true`), MCP server catalog entry, dispatcher wiring, BM25 tool-search coverage, availability snapshot, marketing integrations entry, SDK type unions

This is a first PR intentionally scoped to the skeleton (one read-only tool) — more tools (create/update incident, etc.) will follow in subsequent PRs, matching how Jira (#14414) and Databricks (#18785) were scaffolded.

## Tests

- `npx tsgo --noEmit`: clean
- `cargo check` / `cargo clippy`: clean (only pre-existing warnings shared with `databricks.rs`)
- `npm run test -- bm25_tool_search.test.ts`: 774/774 passing, including new ServiceNow query cases
- `npm run test:update-internal-mcp-availability-snapshot`: snapshot updated and verified stable
- **OAuth verified end-to-end against a live ServiceNow Personal Developer Instance**: registered a real OAuth app, walked the authorization-code flow manually, and confirmed both the token exchange (`finalize`) and refresh-token exchange (`refresh`) return the exact `access_token`/`refresh_token`/`expires_in` shape our Rust provider parses
- Ran an 8-angle automated code review against the diff; fixed two real issues found (missing snapshot entry; `renderIncident` using `??` instead of a falsy check, which missed ServiceNow's empty-string convention for unset fields, plus added `sysparm_display_value=true` so priority/state return human-readable labels)

## Risk

Low. Everything is additive and gated behind the `servicenow_tool` feature flag (off by default) — no existing code path or user-visible behavior changes. Safe to roll back by disabling the flag or reverting.

## Deploy Plan

Standard deploy, no special steps. Icon is a temporary placeholder (`ActionCloudArrowLeftRightIcon`) until a real ServiceNow logo is available from design — marketing will need a redeploy once that lands to show it on the public integrations page.


thomasvicaire:
Splitting this into a stacked pair per review feedback — OAuth first for focused review, then the MCP skeleton on top:

- #29134 (ServiceNow OAuth provider, core + front)
- #29135 (MCP server skeleton, stacked on #29134)

Closing this one in favor of those.

PR #29080, opened 2026-07-17
Stabilize conversation window pruning

flvndvd:
## Description

The existing policy protects a fixed number of recent tool results. Those results can span many interactions,
causing the pruning frontier to advance repeatedly as new tool calls arrive. Whole-interaction dropping also
happens close to the hard context limit, which can produce frequent cache rewrites without creating meaningful
headroom.

This change separates tool-result pruning from whole-interaction dropping. Old tool results are eligible for
pruning across the retained conversation, including long-running current interactions, **while the latest current
tool result remains protected**.

Normal cleanup starts at the soft context limit only when it can reclaim at least 20k tokens. It targets 20k below
that limit. If pruning is insufficient, old complete interactions are dropped in checkpoint-sized batches while the
latest three interactions remain protected from soft dropping. This creates enough headroom to avoid moving the
pruning or dropping frontier on every ordinary interaction.

At the hard limit, checkpoint requirements and previous-interaction protection are lifted. Remaining eligible tool
results are pruned and previous interactions can be dropped until the request fits. The current interaction is
never dropped. Rendering returns a context overflow when its irreducible content cannot fit.

**The conversation is replayed chronologically because the service remains stateless.** Replaying previous interaction
prefixes reproduces earlier cleanup decisions and prevents dropped history from returning as new interactions are
appended (preventing reversals).

This replay introduces an explicit performance tradeoff. Appending interactions adds a loop, and cleanup above the
soft limit may scan retained history to calculate pruning savings and droppable context. The worst case can
therefore approach O(n²), compared with the previous one-shot pass. Retained token totals are tracked
incrementally, so the common path below the soft limit remains linear. Cleanup also reduces the retained window and
normal mutations are separated by 20k checkpoints. A future persisted window state can remove the replay cost and
maintain pruning and dropping frontiers directly.

Monotonicity is currently guaranteed at interaction boundaries. A growing open interaction is still processed
atomically. Moving interaction construction into the window state is intentionally deferred so the class can later
own the open interaction without reconstructing message-level checkpoints in this change.

Existing compaction thresholds and leading-message behavior remain unchanged. The full conversation-rendering and
downstream agent-loop test suites pass.

### DIFF
--- a/front/lib/api/assistant/conversation_rendering/window_state.ts
+++ b/front/lib/api/assistant/conversation_rendering/window_state.ts
@@ -1,21 +1,24 @@
 import type { ConversationPruningStats } from "@app/lib/api/assistant/conversation_rendering/instrumentation";
+import type { InteractionWithTokens } from "@app/lib/api/assistant/conversation_rendering/pruning";
+import {
+  DROP_CHECKPOINT_TOKENS,
+  dropInteractionsToFit,
+  getInteractionTokenCount,
+  getToolResultTokenSavings,
+  PRUNING_CHECKPOINT_TOKENS,
+  pruneToolResults,
+  sumInteractionTokens,
+} from "@app/lib/api/assistant/conversation_rendering/pruning";
 import logger from "@app/logger/logger";
 import type { Result } from "@app/types/shared/result";
 import { Err, Ok } from "@app/types/shared/result";
-import type { InteractionWithTokens } from "./pruning";
-import {
-  dropInteractionsToFit,
-  getInteractionTokenCount,
-  pruneToolResults,
-  sumInteractionTokens,
-  TOOL_RESULTS_TO_PRESERVE,
-} from "./pruning";
+import { assertNever } from "@app/types/shared/utils/assert_never";
 
-// How many of the most recent interactions dropInteractionsToFit never drops entirely, even if
-// their own tool results were already pruned. A different floor from TOOL_RESULTS_TO_PRESERVE
-// (pruning.ts). That one protects tool result content regardless of turn. This one protects whole
-// recent turns from being erased regardless of how many tool calls they made.
+// Existing consumers use this to decide when compaction is available. Keep that behavior
+// independent from the window's soft-drop policy.
 export const PREVIOUS_INTERACTIONS_TO_PRESERVE = 3;
+
+const INTERACTIONS_TO_PRESERVE_AT_SOFT_LIMIT = 3;
 
 type ConversationWindowResult = {
   interactions: InteractionWithTokens[];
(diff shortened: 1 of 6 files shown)


front/lib/api/assistant/conversation_rendering/window_state.ts
```diff
@@ -162,26 +119,259 @@ export class ConversationWindowState {
...
+      return;
+    }
+
+    const eligibleToolResultCount = this.prunableToolResultCount();
+    if (
+      this.toolResultTokenSavings(eligibleToolResultCount) <
+      PRUNING_CHECKPOINT_TOKENS
+    ) {
+      return;
+    }
```
aubin-tchoi, line 169: even if we wouldn't save more than `PRUNING_CHECKPOINT_TOKENS` tokens, we should still prune no?

front/lib/api/assistant/conversation_rendering/window_state.ts
```diff
@@ -162,26 +119,259 @@ export class ConversationWindowState {
...
+
+    const currentInteraction = this.interactions[this.interactions.length - 1];
+    const previousBefore = this.interactions.slice(0, -1);
+    const previousAfter = dropInteractionsToFit(previousBefore, {
+      maxTokens:
+        this.softCleanupTarget() - getInteractionTokenCount(currentInteraction),
+      interactionsToPreserve: 0,
+      batchToCheckpoint: true,
```
aubin-tchoi: for the hard limit we could drop this

front/lib/api/assistant/conversation_rendering/window_state.ts
```diff
@@ -162,26 +119,259 @@ export class ConversationWindowState {
...
+  }
+
+  private prunableToolResultCount(): number {
+    const currentInteraction = this.interactions[this.interactions.length - 1];
+    const currentHasToolResult = currentInteraction.messages.some(
+      (message) => message.role === "function"
+    );
+
+    return (
+      this.countToolResults(this.interactions) - (currentHasToolResult ? 1 : 0)
```
aubin-tchoi, line 315: can be more than 1 tool result in the current interaction no?

front/lib/api/assistant/conversation_rendering/window_state.ts
```diff
@@ -162,26 +119,259 @@ export class ConversationWindowState {
...
+      previousBefore.length - previousAfter.length;
+    this.prunedContext = true;
+    this.interactions = [...previousAfter, currentInteraction];
+    this.retainedTokens = tokensAfter;
+  }
+
+  private pruneEligibleToolResults({
+    eligibleToolResultCount,
+    maxTokens,
+    tier,
```
aubin-tchoi, line 263: shouldn't `maxTokens` be controlled by the `tier` here?

front/lib/api/assistant/conversation_rendering/window_state.ts
```diff
@@ -162,26 +119,259 @@ export class ConversationWindowState {
...
+  }
+
+  private pruneAtHardLimit(): void {
+    if (!this.exceedsHardBudget()) {
+      return;
+    }
+
+    this.pruneEligibleToolResults({
+      eligibleToolResultCount: this.prunableToolResultCount(),
+      maxTokens: this.softCleanupTarget(),
```
aubin-tchoi, line 219: I feel like there's a risk that if we consider the `pruningBudget` here as a limit instead of the actual budget we could end up never really using this step and jumping to the drop directly

aubin-tchoi (approved):
LGTM

front/lib/api/assistant/conversation_rendering/window_state.ts
flvndvd: Not for now, this checkpoint guarantees that we do not prune on every new turn.

front/lib/api/assistant/conversation_rendering/window_state.ts
flvndvd: Yes but we only preserve the last one so the model gets a chance to see the tool that was potentially just executed.

front/lib/api/assistant/conversation_rendering/window_state.ts
flvndvd: The `tier` here does not change the behavior, it's purely for instrumentation purpose.

front/lib/api/assistant/conversation_rendering/window_state.ts
flvndvd: Yeah, fair, set it explicitly to `false`.

front/lib/api/assistant/conversation_rendering/window_state.ts
flvndvd: Yes it's expected, once we pay for a hard cleanup, this creates enough headroom to avoid another cleanup on the next interaction.

PR #29083, opened 2026-07-17
feat: endpoint to get governance permissions

matteotrab:
## Description

This PR introduces a `GET /api/w/:wId/governance-permissions` endpoint to serve workspace governance permissions, replacing the hardcoded test data that was previously used on the frontend.

- Adds a new Hono route handler (`front-api/routes/w/[wId]/governance-permissions.ts`) restricted to business admins and above, calling a new `getWorkspaceGovernancePermissions` service function.
- Admins receive all capabilities (agents, skills, frames, billing/identity); business admins receive the same minus billing/identity.

## Tests

Added tests covering: admin (all capabilities, defaulting to `admins_only`), business admin (no billing/identity), scope reflection (`everyone` and `groups`), and 403 for non-business-admin users.

## Risks

Low.

## Deploy Plan

Standard deployment.

PR #29084, opened 2026-07-17
feat: endpoint to edit governance permissions

matteotrab:
## Description

This PR adds a `PATCH /api/w/:wId/governance-permissions` endpoint to allow admins and business admins to edit workspace governance permissions.

- Implements `setWorkspaceGovernancePermission` in `front/lib/api/permissions/governance.ts`, supporting three scopes: `admins_only`, `everyone`, and `groups` (with validation that only user-manageable groups can be assigned).
- Implements the `useUpdateGovernancePermission` SWR hook in `front/lib/swr/governance.ts`, replacing the previous no-op stub in `GovernancePage.tsx`.

## Tests

Added tests covering: granting to everyone, granting to specific groups, rejecting non-manageable groups, reverting to `admins_only`, role-based access (business admin allowed/forbidden), and empty groups fallback to `admins_only`.

## Risks

Low.

## Deploy Plan

Standard deployment.

### DIFF
--- a/front-api/routes/w/[wId]/governance-permissions.ts
+++ b/front-api/routes/w/[wId]/governance-permissions.ts
@@ -58,7 +58,7 @@ app.patch(
 
     if (result.isErr()) {
       switch (result.error.code) {
-        case "capability_not_managed":
+        case "unauthorized":
           return apiError(ctx, {
             status_code: 403,
             api_error: {
@@ -66,7 +66,15 @@ app.patch(
               message: result.error.message,
             },
           });
-        case "invalid_groups":
+        case "group_not_found":
+          return apiError(ctx, {
+            status_code: 404,
+            api_error: {
+              type: "group_not_found",
+              message: result.error.message,
+            },
+          });
+        case "invalid_id":
           return apiError(ctx, {
             status_code: 400,
             api_error: {
--- a/front-api/routes/w/[wId]/governance-permissions.ts
+++ b/front-api/routes/w/[wId]/governance-permissions.ts
@@ -1,8 +1,21 @@
-import { getWorkspaceGovernancePermissions } from "@app/lib/api/permissions/governance";
-import type { GetGovernancePermissionsResponseBody } from "@app/types/api/governance";
+import {
+  getWorkspaceGovernancePermissions,
+  setWorkspaceGovernancePermission,
+} from "@app/lib/api/permissions/governance";
+import type {
+  GetGovernancePermissionsResponseBody,
+  PatchGovernancePermissionResponseBody,
+} from "@app/types/api/governance";
+import {
+  GRANT_TYPES,
+  GROUP_PERMISSION_RESOURCE_TYPES,
+} from "@app/types/group_permissions";
+import { assertNever } from "@app/types/shared/utils/assert_never";
 import { workspaceApp } from "@front-api/middlewares/ctx";
 import { ensureIsBusinessAdmin } from "@front-api/middlewares/ensure_role";
-import type { HandlerResult } from "@front-api/middlewares/utils";
+import { apiError, type HandlerResult } from "@front-api/middlewares/utils";
+import { validate } from "@front-api/middlewares/validator";
+import { z } from "zod";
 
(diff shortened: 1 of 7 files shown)


matteotrab:

- #29073
    - #29083
        - #29084 ◀
            - #29089
                - #29101
                    - #29107

front-api/routes/w/[wId]/governance-permissions.ts
```diff
@@ -20,4 +33,54 @@ app.get(
...
+              type: "invalid_request_error",
+              message: result.error.message,
+            },
+          });
+        default:
+          assertNever(result.error.code);
+      }
+    }
+
+    return ctx.json({ governancePermission: result.value });
```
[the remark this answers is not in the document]
Comment (philipperolet) [reply front-api/routes/w/[wId]/governance-permissions.ts:90]: true; to do either in this PR or a follow up

front/lib/api/permissions/governance.ts
```diff
@@ -70,3 +92,94 @@ export async function getWorkspaceGovernancePermissions(
...
+      }
+      const groupsRes = await GroupResource.fetchByIds(
+        auth,
+        configuration.groupIds
+      );
+      if (groupsRes.isErr()) {
+        return new Err(
+          new DustError(
+            "invalid_groups",
+            "The groups configuration references unknown groups."
```
philipperolet: nit: why not return groupsRes here? otherwise we lose which error it was in the fetch

front/lib/api/permissions/governance.ts
```diff
@@ -70,3 +92,94 @@ export async function getWorkspaceGovernancePermissions(
...
+
+    case "everyone":
+      await GroupPermissionResource.setForEverybody(auth, capability);
+      break;
+
+    case "groups": {
+      // "Groups" with no groups selected grants the capability to nobody specific, which is
+      // equivalent to admins_only.
+      if (configuration.groupIds.length === 0) {
+        await GroupPermissionResource.disable(auth, capability);
```
philipperolet, line 134: 👍 but make sure client-side we mutate so it reverts to "Admin Only" selector (or alternatively forbid empty group send client side / here)

front/lib/swr/governance.ts
```diff
@@ -31,3 +42,44 @@ export function useGovernancePermissions(
...
+              ...current.governancePermissions,
+              [capabilityKey(governancePermission)]: governancePermission,
+            },
+          },
+        { revalidate: false }
+      );
+
+      return true;
+    },
+    [owner, mutateGovernancePermissions, sendNotification]
```
philipperolet, line 83: ooc do we mutate the setting row on success & failure?

philipperolet (approved):
LGTM 🙏 left a few coms, trusting you on how to adress

Also, related to this PR we need to allow business admins to read provisioned groups => https://github.com/dust-tt/dust/pull/29122

front-api/routes/w/[wId]/governance-permissions.ts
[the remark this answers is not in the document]
Comment (matteotrab) [reply front-api/routes/w/[wId]/governance-permissions.ts:90]: Yes, I'll add audit logs in a follow up. I'm creating an issue on Github so we don't forget

front/lib/api/permissions/governance.ts
matteotrab: Good point, changing to return it

front/lib/api/permissions/governance.ts
matteotrab: We can't always do that. Consider the following
- The toggle is set to "everyone"
- The user clicks on "groups"
- We send the request with no `groupIds` and disable the capability
- We cannot move the toggle to "admins only" or the user would not be able to select any group as the groups dropdown appears only when the toggle is set on "groups".

If we forbid sending empty group then the user would see the toggle on "groups" with no groups selected, but the capability would still be granted to everyone.

We could avoid sending the request when moving the toggle from "admins only" to "groups"

front/lib/swr/governance.ts
matteotrab: We update the setting row optimistically and then revert in case of error. See https://github.com/dust-tt/dust/pull/29089

PR #29085, opened 2026-07-17
Cap GPT 5.6 context at 272k

fontanierh:
## Description

- Cap GPT 5.6 Sol, Terra, and Luna context windows at 272k in both model registries.
- Cap generation at 64k, matching the Anthropic model convention and leaving 208k for input before prompt and tool overhead.
- Keep the standard context budgeting and truncation behavior shared by every model.
- Let the context usage indicator consume the updated registry value.

## Tests

Added coverage for all variants and global/EU endpoints, including the 272k context, 64k output cap, and derived 208k input budget.

## Risk

This only changes GPT 5.6 context-size and generation-limit configuration plus descriptions. It introduces no model-specific budgeting behavior. The PR commits can be reverted directly.

## Deploy Plan

Standard deploy.


aubin-tchoi (approved):
LGTM

front/lib/model_constructors/providers/openai/models/gpt_five_dot_six.test.ts
```diff
@@ -0,0 +1,63 @@
...
+import { OpenAIResponsesGlobalGptFiveDotSixTerraStream } from "@app/lib/model_constructors/stream/endpoints/openai_responses_global_gpt_five_dot_six_terra";
+import {
+  GPT_5_6_LUNA_MODEL_CONFIG,
+  GPT_5_6_SOL_MODEL_CONFIG,
+  GPT_5_6_TERRA_MODEL_CONFIG,
+} from "@app/types/assistant/models/openai";
+import { describe, expect, it } from "vitest";
+
+const EXPECTED_CONTEXT_SIZE = 272_000;
+const EXPECTED_MAX_OUTPUT_TOKENS = 128_000;
```
flvndvd, line 16: IIUC, this means that the actual input size is 272k - 128k => 144k tokens. Two things:
- compaction threshold is based on the full context window so 70% of 250k
- pruning will kick off at 60 % of the allowed token count (so 144k tokens)

This seems a bit low, no?

fontanierh: I reduced to 64k like we do for ANT

PR #29086, opened 2026-07-17
[front] fix: Make impossible to upload file to a managed datasource

id13:
## Description

Currently we can upload files with contentType corresponding to the folder usecase - in datasources managed by a connector 🤯 

This PR makes it impossible to drag and drop from the frontend. And also fails with a 403 on the backend in this case.

## Tests

Tested locally

## Risk

Low

## Deploy Plan

- Deploy front


adrsimon (approved):
ah! interesting
LGTM

PR #29087, opened 2026-07-17
feat(front): replace add tools dropdown with a modal in spaces > tools

adrsimon:
## Description

The "Add Tools" dropdown in spaces > tools got messy, too many entries for a dropdown. This replaces it with a modal: search bar on top, "Add MCP Server" button next to it, and the prefilled MCPs shown as cards with icon + description (2 columns, 3 on large screens).

A few behaviors worth noting:

- The modal stays open after adding a tool, so you can add several in a row. Config dialogs (OAuth etc.) open stacked on top.
- The dialog is now a single controlled instance in `AdminActionsList` (it used to be embedded in the trigger, which unmounted it when the empty-state CTA swapped for the table on first add).

Capture d’écran 2026-07-17 à 11 26 18

Likely next steps are :
- Replicate on triggers page
- Add categories to be able to filter and discover more easily the servers

## Tests

Manual. Typecheck + biome pass.

## Risk

Low, admin-only surface. Worst case the modal misbehaves and we revert to the dropdown. Safe to rollback.

## Deploy Plan

Deploy front.


sparkle/src/components/ActionCard.tsx
```diff
@@ -173,7 +173,7 @@ export const ActionCard = React.forwardRef<HTMLDivElement, ActionCardProps>(
               </div>
             </div>
             <TruncatedText
```
ykmsd, line 175: So this doesn't really work well with link:

Image

I honestly don't think we need to show a tooltip (we should really have a short description but let's not battle for that today). If you can fix TruncatedText that would be great but not sure how it's easy, and I guess this has been here so okay to skip it 🥲

ykmsd (approved):
Fixed a few styling, could you quickly check before you merge? 
Assuming you didn't change any logic and I don't have to test it 🫡

PR #29088, opened 2026-07-17
Add pod function invocation inspection

flvndvd:
## Description

Adds a free, read-only tool for inspecting recent Pod function invocations, including status, timestamps,
input, and result or error. Results are scoped to the current Pod and returned newest-first with a configurable limit.

Also standardizes model-facing terminology around “Pod functions” while preserving internal identifiers.


front/lib/resources/sandbox_function_invocation_resource.ts
```diff
@@ -525,4 +550,25 @@ export class SandboxFunctionInvocationResource extends BaseResource<SandboxFunct
       return new Err(normalizeError(error));
     }
   }
+
+  toJSON(): SandboxFunctionInvocationType {
```
flvndvd, line 554: Just moved around.

spolu (approved):

PR #29090, opened 2026-07-17
[front] refactor: move tool stakes from server-level to tool-level metadata

aubin-tchoi:
## Description

Internal MCP tool stakes are currently duplicated in a server-level `tools_stakes` record, for each single MCP server (the code that collects the stakes and put them in a record is strictly the same). This PR moves this collection to the call sites (few of them). In the process, the `stake` had to be added onto each `InternalMCPToolType` (in the definition of the tool metadata we did drop it).

Also removes some duplicated tool stake definitions that were also on `INTERNAL_MCP_SERVERS`.
The internal MCP server runbook is also updated to use the new metadata shape.

The values of the stakes are not changed, only the wiring around them.

## Tests

- Tested locally, passes existing tests.

## Risk

- Low.

## Deploy Plan

- Deploy front.


flvndvd (approved):
\o/

PR #29092, opened 2026-07-17
[Activation] Create activation nudge convo from poke

xebeohp:
## Description
Create an activation nudge conversation from a poke plugin that only shows up for activation pods (image shown below). A separate conversation is created for each user, and nudge conversations are created for **all** members of the pod at one time.

This reuses the same trigger/webhook flow as "join activation pod" to emit a single activation webhook event and each activation pod member's trigger starts a nudge conversation.

Screenshot 2026-07-17 at 13 48 12

## Tests
Manual tests on Poke and local.

## Risk
Low risk. A Poke Plugin only liked to Activation pods.

## Deploy Plan

### DIFF
--- a/front/lib/api/poke/plugins/spaces/send_activation_nudge.ts
+++ b/front/lib/api/poke/plugins/spaces/send_activation_nudge.ts
@@ -1,31 +1,14 @@
-import { getAgentConfiguration } from "@app/lib/api/assistant/configuration/agent";
-import {
-  createConversation,
-  postUserMessage,
-} from "@app/lib/api/assistant/conversation";
+import { emitActivationEvent } from "@app/lib/api/activation/trigger";
 import { createPlugin } from "@app/lib/api/poke/types";
 import { Authenticator } from "@app/lib/auth";
-import { serializeMention } from "@app/lib/mentions/format";
 import { ProjectMetadataResource } from "@app/lib/resources/project_metadata_resource";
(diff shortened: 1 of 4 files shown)


xebeohp:

- #29092 ◀

front/lib/api/poke/plugins/spaces/send_activation_nudge.ts
```diff
@@ -0,0 +1,251 @@
...
+        profilePictureUrl: userJson.image,
+        // UTC since this runs server-side without the user's timezone.
+        timezone: "UTC",
+        origin: "triggered",
+      };
+
+      // Post the nudge as the user's first message to @dust: the serialized
+      // agent mention followed by the nudge body, exactly like a trigger's
+      // initial message (serializeMention + customPrompt).
+      const postRes = await postUserMessage(recipientAuth, {
```
frankaloia: Instead of create conversation can we reuse the logic we have in the orchestrator where we send the event to the event source?

xebeohp: Yeah that also works. In that case each member of the pod would get a nudge conversation always and theres no subset selection like what I have right now.

xebeohp: Not sure if being able to select a subset of members to get the nudge is useful or if nudging all members of the pod is what we want?

frankaloia (approved):

PR #28450, opened 2026-07-03
feat(front-billing): grant free programmatic credits via Stripe webhook

tdraier:
## Description

Enterprise shadow contracts (#28234) and their packages (#28319) were removed, so the Metronome `credit.segment.start` recurring-credit path no longer fires for enterprise workspaces — they stopped receiving their monthly/yearly free programmatic-usage credits.

This reverts to the pre-#25129 mechanism **for all plans**: grant the free credit on the Stripe `customer.subscription.updated` / `customer.subscription.created` webhook, anchored to the Stripe billing period. Simpler, no shadow contracts, no reprovisioning script.

### Changes
- **`lib/credits/free.ts`**: restore `grantFreeCreditsFromSubscriptionStateChange` / `grantFreeCreditFromSubscriptionStateChangeYearly` (from git `5631abb557~1`) and add a `grantFreeCreditsForSubscription` dispatcher (monthly vs yearly). Idempotent per billing cycle via `free-renewal-{sub}-{periodStart}`.
- **`linkAndSyncMetronomeRecurringCredit`**: best-effort pushes the granted amount into the workspace's Metronome recurring free credit (`updateMetronomeCreditSegmentAmount`) and stamps `metronomeCreditId`, so Metronome-billed plans (legacy pro/business) keep their invoice offset **in sync**. No-op for enterprise / non-Metronome workspaces. DB credit is committed first, so a Metronome hiccup only affects invoice offset (reconciled by the sync script).
- **`lib/api/stripe/webhook_handler.ts`**: wire the grant into `customer.subscription.updated` and `customer.subscription.created`.
- **`lib/api/metronome/process_webhook.ts`**: remove the automatic free-credit grant on `credit.segment.start`. Kept `handleFreeCreditSegmentGrant` (poke plugin) and `grantFreeCreditFromMetronomeSegment` (fix script).
- **`lib/credits/free.test.ts`**: restore the grant unit tests.

## Tests

- `npx tsgo --noEmit` clean.
- `free.test.ts` → 42 passed (includes restored monthly + yearly grant suites).
- `npm run format:changed` clean.

## Risk

Low–medium (billing). No double-grant: legacy pro/business Metronome contracts still fire `credit.segment.start`, but that path no longer creates a DB credit — the Stripe path is the sole grantor and syncs the Metronome amount. Residual: the poke plugin can still be invoked manually (now effectively obsolete).

## Deploy Plan

[...]

PR #29093, opened 2026-07-17
Cleaned up poke API Usage tab

tdraier:
## Description

Now that free programmatic credits are granted via the Stripe `customer.subscription.updated` webhook (#28450) and are no longer backed by a Metronome credit segment, this PR strips the now-dead Metronome-reconciliation paths out of the Poke "API Usage" tab and its supporting code:

- **Poke plugin removal**: deletes the `trigger-free-credit-segment-grant` plugin (`lib/api/poke/plugins/workspaces/trigger_free_credit_segment_grant.ts`) and its UI trigger button (`TriggerFreeCreditSegmentGrantButton.tsx`), along with `handleFreeCreditSegmentGrant` in `lib/api/metronome/process_webhook.ts` (only caller was the plugin). Trims the imports that became unused as a result.
- **API Usage tab / unified credits table**: `front-api/routes/poke/workspaces/[wId]/credits/index.ts` no longer calls `listMetronomeBalances` or merges Metronome balances into the response — it just lists internal DB credits (`CreditResource.listAll`) plus the 30-day excess total. `PokeUnifiedCreditRow` drops its `metronome` field (now DB-only, non-nullable `internal`) and the `hasMetronome` flag is gone from `PokeListCreditsResponseBody`.
- **Columns/table cleanup**: `components/poke/credits/columns.tsx` drops the "Status" column (Metronome vs DB reconciliation chips: "Metronome only" / "Missing in Metronome" / "Matched" / "Mismatch") and all DB-vs-Metronome comparison helpers (`AmountCell`, `pickMetronome`, `getRowType`/`getRowStartDateMs`/`getRowExpirationDateMs` fallbacks) — columns now render the DB values directly.
- **Charts**: removes the "Credit usage" chart (`PokeAwuUsageFromAnalyticsChart`) from the API Usage tab (`table.tsx`), keeping only the "Usage cost graph" (`PokeProgrammaticCostChart`). The chart component itself is untouched since it's still used by the separate Poke "Usage" tab.
- **Dead code removal**: deletes `lib/api/credits/metronome_balances.ts` (`metronomeBalanceToDisplayData`, only used by the route above) and `isMetronomeExcessCredit` in `lib/metronome/types.ts` (same reason). Deletes `scripts/realign_free_credits_to_stripe_periods.ts`, a one-off backfill script that realigned free-credit periods to Stripe billing cycles — no longer needed now that credits aren't Metronome-period-derived.

No behavior change for still-active flows: `isMetronomeFreeCredit`, `calculateFreeCreditAmountMicroUsd`, `countEligibleUsersForFreeCredits`, `updateMetronomeCreditSegmentAmount`, and `grantFreeCreditFromMetronomeSegment` are all still used elsewhere (other scripts/plugins) and are untouched.

## Tests

- `npx tsgo --noEmit` passes clean on both `front` and `front-api`.
- `npm run format:changed` (biome) passes with no fixes needed.
- No functional/endpoint tests existed for the removed plugin or the Poke credits route; none were added since this is a pure removal of dead paths with no new behavior.
- Not manually verified in the browser (Poke admin UI) — recommend a quick manual check of the "API Usage" tab on a workspace after deploy to confirm the credits table and cost graph still render correctly without the removed column/chart.

## Risk

Low. This only removes reconciliation/manual-recovery paths that depended on Metronome data no longer being the source of truth for free credits:
- The removed Poke plugin was a manual admin fallback for a grant path that Stripe now handles automatically; removing it just means there's no manual replay button anymore (none of the other credit-granting flows are affected).
- The Poke "API Usage" tab becomes DB-only — cosmetic/admin-visibility change only, no impact on billing or credit-granting logic.
- The deleted backfill script (`realign_free_credits_to_stripe_periods.ts`) is a manually-invoked one-off script, not called from any scheduled job or other code path.

Rollback is a straightforward revert if the removed admin tooling turns out to still be needed.

## Deploy Plan

Standard deploy, no special sequencing or migration required (no DB schema or data changes).


tdraier:

- #29093 ◀

front/types/api/poke/credits.ts
```diff
@@ -16,12 +16,10 @@ export type PokeCreditType = {
 
 export type PokeUnifiedCreditRow = {
```
PopDaph: PokeUnifiedCreditRow is dead weight now: nothing is "unified" anymore, and rowKey isn't consumed anywhere client-side. 
We could flatten to rows: PokeCreditType[] and rename makeColumnsForUnifiedCredits? Feel free to ignore of course

PopDaph (approved):
Love the diff!

front/types/api/poke/credits.ts
tdraier: missed that one :-)

PR #29094, opened 2026-07-17
[front] fix: Remove user from projectPreferences

id13:
## Description

Currently we derive user from auth to attach it as an exposed user atr on userProjectPreferencesResource.
It's incorrect since the pref row itself belong to a different user from the auth object, and it's making the workflow crash for messages coming from v1 api

## Tests

N/A

## Risk

Low

## Deploy Plan

Deploy front


id13:
PMRR

matteotrab (approved):

PR #29096, opened 2026-07-17
[front] refactor: read internal MCP stakes directly from metadata

aubin-tchoi:
## Description

Follow-up on #29090. Now that stakes live on each internal MCP tool, `getInternalMCPServerToolStakes` only rebuilds a separate record for callers to consume.

There are 2 callers and one of them rebuilds the record just to read 1 value from it. This PR removes the helper and reads stakes directly from the tools metadata.

## Tests

- Covered by existing MCP metadata tests.

## Risk

- Low.

## Deploy Plan

- Deploy front.


aubin-tchoi:
PMRR (I have a bigger one coming)

PR #29100, opened 2026-07-17
Use different metronome customer in dev

tdraier:
## Description

Allow to use metronome on dust-hive workspace , by adding different id on top of the workspace id. This prevent mixing all contracts on the same customer - each hive will have it's own metronome customer.

## Tests

locally

## Risk

dev only


tdraier:

- #29100 ◀

front/lib/metronome/client.ts
```diff
@@ -170,6 +170,28 @@ export async function ingestMetronomeEvents(
 // Customer management
 // ---------------------------------------------------------------------------
 
+/**
+ * The Metronome ingest alias for a workspace. Normally just the workspace
+ * sId, so usage events with `customer_id: workspaceId` match the customer
+ * created with that alias.
+ *
+ * In dust-hive environments, every hive (and its users) shares the same
```
fontanierh: do we want "dust hive" to leak inside front prod code ?

fontanierh (approved):

front/lib/metronome/client.ts
tdraier: haa I renamed the env var but did not check the comments.. removing 🙏

PR #29104, opened 2026-07-17
Performance improvements on seat sync

tdraier:
## Description

Two more seat-sync performance fixes, found from production logs on a ~500-user workspace mid the legacy Pro → Business migration, whose `syncMetronomeSeatCountWorkflow` had been stuck retrying `syncMetronomeSeatCountActivity` (10-minute `startToCloseTimeout`) 1740 times — i.e. the sync had never once completed within the timeout.

- **`remapMembershipSeatTypesForContract` was re-scheduling every not-yet-migrated member on every single membership event** (`lib/metronome/seats.ts`). This is called from `stagePendingContractSeats`, which runs on every membership add/change/revoke while a workspace has a pending contract — not just during the migration script's bulk run. Its no-op check compared the computed target against the *active* membership row's `seatType`, but the scheduling path (`scheduleSeatChange`) never touches that row — it only closes it and creates a separate future-dated row. So a member already correctly scheduled by a previous call still looked unmigrated on the next one, and got re-scheduled (a full destroy+update+create transaction) again — for every not-yet-migrated member, on every membership event in the workspace, for the whole migration window. Now the function also fetches each member's existing scheduled row (`MembershipResource.getScheduledMembershipsByUserIdInWorkspace`, one batched query) and skips re-scheduling when it already matches the target seat type and start date.

- **`syncSeatCount`'s per-timestamp reconcile was quadratic in workspace size** (`lib/metronome/seats.ts`). `desiredSIdsAt` is called once per `(seat subscription × distinct effective timestamp)`, and for each call, `seatTypeAt` walked the *entire* `scheduledChanges` list for every single user to find that user's own changes. With most users mid-migration carrying a pending scheduled change, that's `subscriptions × timestamps × users × scheduledChanges` comparisons — for 500 users this reaches into the millions of iterations, run synchronously. `scheduledChanges` is now grouped into a `Map<userId, ScheduledChange[]>` (sorted ascending) once up front, so each lookup only walks that one user's own changes instead of the whole workspace's.

- Removed two `logger.info` calls in `stagePendingContractSeats` (`lib/api/metronome/seat_sync.ts`) that fired on every sync for every workspace *without* a pending contract — pure log-volume noise, since the vast majority of workspaces hit this path on every membership change.

## Tests

- Added 5 test cases for `remapMembershipSeatTypesForContract` in `seats.test.ts`, covering: no existing schedule (schedules), an existing schedule that already matches (no-op), a mismatched seat type (re-schedules), a mismatched start date (re-schedules), and the immediate-apply path (never looks up scheduled rows, applies directly). Full `lib/metronome` suite (70 tests) passes.
- Not yet verified live against the previously-stuck workflow — waiting to see whether it now completes and clears its retry count once deployed.

## Risk

- Both fixes preserve behavior for the case they don't optimize: the "already scheduled" no-op requires *both* the seat type and the start date to match, so a genuine change (e.g. the pending contract's start date itself moves) still triggers a re-schedule. The `seatTypeAt` grouping produces the exact same per-user/per-timestamp result as before, just without rescanning the whole workspace to get it.
- No schema or migration changes, no feature flag. Safe to roll back — reverts to the previous (slower, but already-correct) behavior.

## Deploy Plan

Standard deploy. Once live, re-run "Sync Metronome Seat Count" from poke (or wait for the next membership event) on the workspace whose workflow was stuck retrying, and confirm it completes within the activity timeout instead of continuing to retry.

### DIFF
--- a/front/lib/api/metronome/seat_sync.ts
+++ b/front/lib/api/metronome/seat_sync.ts
@@ -269,10 +269,6 @@ async function stagePendingContractSeats({
   const pendingSubscription =
     await SubscriptionResource.fetchPendingByWorkspaceModelId(workspace.id);
   if (!pendingSubscription?.metronomeContractId) {
-    logger.info(
-      { workspaceId },
-      "[SeatSync] stagePendingContractSeats — no pending subscription, skipping"
-    );
     return false;
   }
   const pendingContractResult = await getMetronomeContractById({
@@ -283,10 +279,6 @@ async function stagePendingContractSeats({
     pendingContractResult.isErr() ||
     !(await hasContractSeatSubscription(pendingContractResult.value))
   ) {
-    logger.info(
-      { workspaceId },
-      "[SeatSync] stagePendingContractSeats — pending contract has no seat subscription, skipping"
-    );
     return false;
   }
   const pendingContract = pendingContractResult.value;
--- a/front/lib/metronome/seats.test.ts
+++ b/front/lib/metronome/seats.test.ts
@@ -5,6 +5,7 @@ import {
   getSeatCreditNameForSeatType,
   hasContractSeatSubscription,
   promoteNoneSeatTypesForContract,
+  remapMembershipSeatTypesForContract,
   resolveRemappedSeatType,
 } from "@app/lib/metronome/seats";
 import {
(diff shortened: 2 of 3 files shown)


tdraier:

- #29104 ◀

PopDaph (approved):
Looks reasonable :D

PR #29067, opened 2026-07-16
feat(activation): record and gate pod nudges with a frequency cap

ElPicador:
## Description

Builds on the activation scheduler by recording every nudge that's sent and using that history to avoid over-nudging.

- Adds `ActivationNudgeResource`/`ActivationNudgeModel`, and records one each time the scheduler fires a pod's activation trigger.
- Adds `isEligibleForNudge`, called before firing a pod's trigger: skips the pod if it was already nudged within the last `activationNudgeFrequencyCapDays` (workspace `metadata`, defaults to 2 days when unset).

## Tests

Added unit tests for `getActivationNudgeFrequencyCapDays` (default, workspace override, invalid override) and `isEligibleForNudge` (never nudged vs. nudged within the cap window).

PR #29105, opened 2026-07-17
feat(activation): gate nudges on opted-out and dead pods

ElPicador:
## Description

Adds two more eligibility conditions to `isEligibleForNudge`, on top of the existing frequency cap and unanswered-streak checks — both apply even to a pod that's never been nudged before:

- **Opted out**: the pod's activation trigger is missing or its `status` isn't `"enabled"` (the user disabled it).
- **Dead**: the pod is archived, or its target user (the trigger's editor) no longer has an active workspace membership.

`TriggerResource.listBySpace` was added to look up a pod's activation trigger. `TriggerFactory.webhook` in tests now accepts an optional `spaceId` so tests can wire a trigger to a specific pod.

## Tests

Added coverage for: never-nudged pod with an enabled trigger, disabled trigger (opted out), missing trigger, archived pod, and a user who left the workspace (dead).

### DIFF
--- a/front/lib/api/activation/nudge.ts
+++ b/front/lib/api/activation/nudge.ts
@@ -1,7 +1,11 @@
+import { findActivationTrigger } from "@app/lib/api/activation/trigger";
 import type { Authenticator } from "@app/lib/auth";
 import { ActivationNudgeResource } from "@app/lib/resources/activation_nudge_resource";
 import { ConversationResource } from "@app/lib/resources/conversation_resource";
+import { MembershipResource } from "@app/lib/resources/membership_resource";
 import type { SpaceResource } from "@app/lib/resources/space_resource";
+import type { TriggerResource } from "@app/lib/resources/trigger_resource";
+import { UserResource } from "@app/lib/resources/user_resource";
 import {
   DEFAULT_ACTIVATION_NUDGE_FREQUENCY_CAP_DAYS,
   DEFAULT_ACTIVATION_NUDGE_MAX_UNANSWERED_COUNT,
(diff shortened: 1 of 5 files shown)


ElPicador:

- #29031
    - #29033
        - #29059
            - #29065
                - #29067
                    - #29103
                        - #29105 ◀
                            - #29117
                                - #29226
                                    - #29244
                                        - #29260

front/lib/api/activation/nudge.ts
```diff
@@ -78,7 +81,46 @@ async function countUnansweredNudgeStreak(
...
-// Gates re-nudging a pod on two conditions:
+// Finds the pod's activation trigger (the webhook trigger scoped to this
+// pod's space). Null if the pod has no trigger, which shouldn't happen for a
+// pod returned by `ProjectMetadataResource.fetchActivationPods`.
+async function fetchActivationTrigger(
+  auth: Authenticator,
+  pod: SpaceResource
+): Promise<TriggerResource | null> {
+  const triggers = await TriggerResource.listBySpace(auth, pod.id);
+  return triggers.find((trigger) => trigger.kind === "webhook") ?? null;
```
frankaloia: The pod is likely to have other triggers besides activation, so we may need to filter down if we want to pinpoint to that

ElPicador: indeed!

ElPicador:
Fixed in a follow-up commit: `fetchActivationTrigger` now pinpoints the trigger via `webhookSourceViewId` instead of just `kind === "webhook"`.

The activation trigger *is* marked distinctly — each pod gets its own `WebhookSourcesViewResource` view of a single shared, workspace-level "Activation" `WebhookSourceResource` (looked up by name), and `createActivationTrigger` stamps that view's id onto the trigger's `webhookSourceViewId`. New `findActivationTrigger` in `trigger.ts` looks up that source, finds the pod's view of it, then uses the existing `TriggerResource.listByWebhookSourceViewId` to fetch the trigger — so an unrelated webhook trigger a user created in the same pod space can no longer be mistaken for the activation one.

Updated `nudge.test.ts` to wire triggers the same way production does (real webhook source + view), and added `includeDeleted: true` to the view lookup so this still resolves for archived pods (needed since eligibility gating fetches the trigger before it can determine the pod is dead).

front/lib/api/activation/trigger.ts
```diff
@@ -158,6 +158,41 @@ export async function createActivationTrigger(
   return new Ok({ triggerId: triggerRes.value.sId });
 }
 
+// Finds the pod's activation trigger: the trigger on the pod's own view of
+// the shared Activation webhook source (see
+// `getOrCreateActivationWebhookSourceView`). Filtering on `webhookSourceViewId`
+// rather than just `kind === "webhook"` matters because a pod's space can hold
+// other, unrelated webhook triggers. There is at most one activation trigger
+// per pod, since `createActivationTrigger` provisions exactly one per pod view.
```
frankaloia, line 166: Technically as of today there is nothing strictly enforcing this. I think this is ok

frankaloia (approved):
LGTM, one minor comment but i think not worth actioning

PR #29108, opened 2026-07-17
fix: expose agent config instructions as per-block array

adrsimon:
## Description

`get_agent_config` (Sidekick's primary tool) serialized `instructionsHtml` as a single JSON string. For agents with large prompts (we have one that is ~44KB), that value lands on one >20KB line. Once the tool output is offloaded to a conversation file, Sidekick reads it back with `files cat`, whose byte cap truncates the oversized line at 20KB, and since its `offset` is a line number, there's no way to page into the remainder. So Sidekick could never read the full instructions and edits to long prompts broke.

Fix at the source: emit instructions as an array of top-level HTML blocks (`instructionsHtmlBlocks`) instead of one escaped string. Pretty-printed JSON then puts one block per line, keeping the output paginatable line-by-line while preserving each block's `data-block-id`.

- `getAgentConfig.ts`: field `instructionsHtml` → `instructionsHtmlBlocks`, plus updated tool description.
- `shared/utils.ts`: new `splitInstructionsHtmlBlocks` helper (DOMParser, same pattern as `getBlockOuterHtml`).
- `agent_suggestions_shared.ts`: updated the Sidekick prompt to describe the array shape.

## Tests

Typecheck + biome pass (also via pre-commit hooks). No automated test covers this client-side tool's output today. Verified manually that `splitInstructionsHtmlBlocks` returns one entry per top-level block with `data-block-id` intact, and `[]` for empty instructions.

## Risk

Low. 

## Deploy Plan

Standard front deploy.

### DIFF
--- a/front/components/agent_builder/sidekick/tools/getAgentConfig.ts
+++ b/front/components/agent_builder/sidekick/tools/getAgentConfig.ts
@@ -1,4 +1,5 @@
 import type { AgentBuilderFormData } from "@app/components/agent_builder/AgentBuilderFormContext";
+import { splitInstructionsHtmlBlocks } from "@app/components/shared/utils";
 import datadogLogger from "@app/logger/datadogLogger";
 import type { AgentSuggestionType } from "@app/types/suggestions/agent_suggestion";
 import type { McpServer } from "@modelcontextprotocol/sdk/server/mcp.js";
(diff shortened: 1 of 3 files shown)


frankaloia (approved):
LGTM

PR #29110, opened 2026-07-17
Remove seat_limits_schedule plugin, only use it through ui

tdraier:
## Description

`manage-seat-limits` was a poke plugin whose args (`enabled` toggle, single `minSeats`/`maxSeats`/`startDate`/`endDate` window) no longer matched reality: seat limits are edited exclusively through the seat-limit schedule dialog, which posts a full list of phases to `POST /api/poke/workspaces/[wId]/seat_limits_schedule` → `setSeatLimitScheduleForSeatType` → `WorkspaceSeatLimitResource.setScheduleForSeatType`, bypassing the plugin entirely. The plugin was `isHidden: true` and referenced nowhere else — stale, unreachable code that could apply an inconsistent single-window change to seat limits if it were ever invoked directly (e.g. via the generic `/api/poke/plugins/:pluginId/run` route).

This PR:
- Reworks `manage_seat_limits.ts` into a no-op plugin (`execute` always returns `Err("NO_OP")`), matching the existing convention in `upgrade_downgrade.ts` (`upgrade-free-plan`, `downgrade-no-plan`, `switch-contract`, ...) for plugins that exist only to back a `PluginRunResource` audit record for a dedicated endpoint. Its args are reduced to `seatType` + `phasesJson` (a JSON dump of the phases actually submitted).
- Updates `setSeatLimitScheduleForSeatType` to create a `PluginRunResource` via `PluginRunResource.makeNew` and record the outcome with `recordError`/`recordResult`, the same pattern already used by `upgrade.ts`/`downgrade.ts`/`switch_contract.ts`. Every seat-limit schedule change made through the dialog is now captured in the poke plugin-run audit trail, which it wasn't before.

## Tests

- `npx tsgo --noEmit` passes.
- `npx biome check` passes on the changed files.
- No functional test exists yet for `seat_limits_schedule.ts` (none did before this change either).

## Risk

Low. The actual seat-limit mutation logic (`WorkspaceSeatLimitResource.setScheduleForSeatType`, Metronome sync) is unchanged. The only behavior change is that a `PluginRunResource` row is now written per schedule change; a failure to write that row is logged but does not fail the request, since the real change already succeeded by that point.

## Deploy Plan

### DIFF
--- a/front/lib/api/poke/plugins/workspaces/manage_seat_limits.ts
+++ b/front/lib/api/poke/plugins/workspaces/manage_seat_limits.ts
@@ -1,210 +1,44 @@
 import { createPlugin } from "@app/lib/api/poke/types";
-import { WorkspaceSeatLimitResource } from "@app/lib/resources/workspace_seat_limit_resource";
-import { launchMetronomeSeatCountSyncWorkflow } from "@app/temporal/usage_queue/client";
-import {
-  isMembershipSeatType,
-  MEMBERSHIP_SEAT_TYPES,
-} from "@app/types/memberships";
-import { isCreditPricedPlan } from "@app/types/plan";
+import { MEMBERSHIP_SEAT_TYPES } from "@app/types/memberships";
 import { mapToEnumValues } from "@app/types/poke/plugins";
-import type { Result } from "@app/types/shared/result";
-import { Err, Ok } from "@app/types/shared/result";
-import { isString } from "@app/types/shared/utils/general";
-import { z } from "zod";
-import { fromError } from "zod-validation-error";
+import { Err } from "@app/types/shared/result";
 
-// Poke date args arrive as strings (empty when unset). Parse to a Date or
-// null, surfacing a readable error for an unparseable value.
-function parseOptionalDate(
-  value: unknown,
-  label: string
-): Result<Date | null, Error> {
-  if (!isString(value) || value.trim() === "") {
-    return new Ok(null);
-  }
-  const date = new Date(value);
-  if (isNaN(date.getTime())) {
-    return new Err(new Error(`Invalid ${label}.`));
-  }
-  return new Ok(date);
-}
-
-function toDateLabel(date: Date): string {
-  return date.toISOString().slice(0, 10);
-}
-
-const SeatLimitArgsSchema = z.object({
-  minSeats: z
-    .number()
-    .int("Min seats must be an integer")
-    .min(0, "Min seats must be at least 0"),
-  maxSeats: z.preprocess(
-    (v) => (v === 0 ? undefined : v),
-    z
-      .number()
-      .int("Max seats must be an integer")
-      .min(1, "Max seats must be at least 1")
-      .optional()
-  ),
-});
-
+// Seat limits are now edited exclusively through the seat-limit schedule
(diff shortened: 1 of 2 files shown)


tdraier:

- #29110 ◀

front/lib/api/poke/plugins/workspaces/manage_seat_limits.ts
```diff
@@ -1,210 +1,44 @@
...
     description:
-      "Configure the per-seat-type minimum and maximum seat counts for this workspace. " +
-      "The minimum is the billing floor sent to Metronome even when fewer " +
-      "members hold that seat (the shortfall is billed as unassigned seats). " +
-      "The maximum is a hard cap — assignments into an at-cap tier are rejected. " +
-      "Disabling removes the floor and cap for the selected seat type.",
+      "Audit-log record of a seat-limit schedule change applied through the " +
+      "seat-limit schedule dialog.",
     resourceTypes: ["workspaces"],
+    isHidden: true,
```
Nils-Fedrigo, line 19: in the desc you said that the plugin was previously isHidden= true but here you add it?

front/lib/api/poke/plugins/workspaces/manage_seat_limits.ts
```diff
@@ -1,210 +1,44 @@
...
-            `${syncResult.error.message}. Re-run once the issue is resolved.`
-        )
-      );
-    }
-
-    return new Ok({
-      display: "text",
-      value: `${message} Metronome seat sync launched.`,
-    });
+  execute: async () => {
```
Nils-Fedrigo, line 41: I don't understand why you don't delete the plugin instead but you probably know better!

Nils-Fedrigo: ah ok I understand by reading below

front/lib/api/poke/seat_limits_schedule.ts
```diff
@@ -72,26 +74,53 @@ export async function setSeatLimitScheduleForSeatType(
 ): Promise<Result<void, Error>> {
   const workspace = auth.getNonNullableWorkspace();
 
+  // `manage-seat-limits` is a no-op plugin (see manage_seat_limits.ts):
+  // it exists only so this endpoint can save a plugin record for the
+  // change it makes, the same way upgrade.ts/downgrade.ts do.
+  const plugin = pluginManager.getNonNullablePlugin("manage-seat-limits");
+  const pluginRun = await PluginRunResource.makeNew(
```
Nils-Fedrigo, line 81: instead directly log something in the audit log?

Nils-Fedrigo (approved):
just seems a bit weird for me to keep a plugin instead of doing a log directly but you probably know what you are doing!

front/lib/api/poke/plugins/workspaces/manage_seat_limits.ts
tdraier: my bad, it's not hidden, but very confusing to use as is - better to hide it

tdraier:
> just seems a bit weird for me to keep a plugin instead of doing a log directly but you probably know what you are doing!

weird indeed, that'a a pattern we have for all these "plugin" which have dedicated ui .. plugin is here only to keep the audit run entries.

PR #29114, opened 2026-07-17
feat(front): group model picker by model maker instead of serving provider

Nils-Fedrigo:
## Description

In the input-bar model picker, the "More models" section grouped models by their **serving provider**. Several models are served through Fireworks but are actually made by other labs (Z.ai's GLM, Moonshot AI's Kimi, MiniMax's M2.5, DeepSeek's V3.2/V4), so they all bucketed under a single "Fireworks" header — hiding the real creator.

This groups and labels that section by the **real lab/maker** instead. Models on native providers (OpenAI, Anthropic, Google, Mistral, xAI) are unaffected — for them maker == provider.

Changes:
- New optional, display-only `modelMaker` field on `ModelConfigurationType` (`getModelMaker(model)` returns `model.modelMaker ?? model.providerId`). Only the Fireworks-served configs set it (`deepseek`, `moonshot`, `minimax`, `zai`).
- `getModelMakerDisplayName` / `getModelMakerLogo` fall through to the existing provider helpers for shared makers, and add the three new labs.
- New Sparkle logos for Z.ai, Moonshot AI, MiniMax.
- Input-bar picker groups "More models" by maker (`getModelMaker`); search matches maker names.
- Agent builder model selection ("Other models" submenu) groups by maker too, with the "From {maker}" labels and per-row logos resolved from the maker.

The `" (Fireworks)"` suffix on model display names is intentionally kept as-is.

### Why not just change `providerId` to the real provider?

It's tempting to fix this at the root by setting `providerId: "zai"` (etc.) on the Fireworks configs instead of adding a separate `modelMaker` field. We deliberately don't, for two reasons:

1. **`providerId` means "serving provider", and these models really are served by Fireworks.** It is the routing/credentials key: it selects the API client, the API key, region routing, and BYOK/whitelist gating. GLM-5.2 is reached through the Fireworks API with Fireworks credentials — its `providerId` must stay `"fireworks"` or the request would be routed to a provider we don't have a client/keys for. This matches how the rest of the stack already models it: for Anthropic in the EU, `providerId` stays `"anthropic"` while the actual serving backend (Vertex / `agent-platform`) is a separate axis (`api` + `region`) in `model_constructors` — the maker and the serving path are already decoupled everywhere except that Fireworks-served models had no place to record the maker.

2. **`providerId` is on the wire and persisted.** It's part of the `ModelSelectionType` sent from the picker and stored as the last-requested model, and it keys agent model configs. Changing it would be a breaking change to persisted selections and the public/private API contract.

So `modelMaker` is a purely additive, display-only field that restores the maker the Fireworks path drops, without touching routing, credentials, or any persisted/wire contract.

## Tests

Manual: with `models_picker` + `fireworks_new_model_feature` enabled, open the input-bar picker → "More models" and confirm Fireworks-served models now appear under **DeepSeek**, **Moonshot AI**, **MiniMax**, and **Z.ai** headers (with logos) rather than one "Fireworks" group; native-provider groups unchanged; selecting a model still sends the same `{providerId, modelId, reasoningEffort}`; searching "Z.ai" / "MiniMax" surfaces the matching models.

Note: the Sparkle package must be rebuilt (`npm --prefix sparkle run build`) and the front server restarted for the new logos and `modelMaker` field to appear (configs are read server-side).


Nils-Fedrigo:

- #29114 ◀

front/types/assistant/models/providers.ts
```diff
@@ -52,6 +54,42 @@ export function getProviderDisplayName(
...
 }
+/**
+ * MODEL MAKERS
+ *
+ * The "maker" is the lab that actually created a model, as opposed to the
+ * serving `providerId`. For native providers these coincide (an OpenAI model is
+ * served by OpenAI), so a config only sets `modelMaker` when the two diverge —
+ * e.g. Z.ai's GLM or Moonshot AI's Kimi served through Fireworks. Makers reuse
+ * provider ids for the shared cases plus a few lab-only ids.
+ */
```
pmilliotte, line 59: Comment too long

front/types/assistant/models/types.ts
```diff
@@ -72,6 +73,10 @@ const CustomAvailabilityConditionSchema = z.object({
 // This is the source of truth for the structure of ModelConfigurationType.
 export const ModelConfigurationSchema = z.object({
   providerId: z.string(),
+  // The lab that created the model, when it differs from the serving
+  // `providerId` (e.g. Z.ai's GLM served through Fireworks). Defaults to
+  // `providerId` when unset — see `getModelMaker`.
```
pmilliotte: too long

pmilliotte (approved):
🎉

PR #29115, opened 2026-07-17
[front] refactor: simplify internal MCP tool metadata typing

aubin-tchoi:
## Description

Part of a multi step plan to allow defining in which context a tool can be ran on the tool metadata themselves in a declarative way.

The plan goes like this:

- Remove the boilerplate helpers `createToolRecords` and `buildTools`.
- Move the `createServer` one level up.
- Add a metadata to declare which contexts each tool supports (by default all contexts) + bake it in the generic `createServer`.

Follow up on https://github.com/dust-tt/dust/pull/29119

This PR defines tools metadata as an arrays with explicit names and zod schemas that are surfaced one layer up (this will unlock a much richer typing as we can infer a type from a zod schema, unlike a JSON schema. So we're better off sticking with the zod schema as long we can). This removes `createToolsRecord`, `createToolsArray`, and the repeated schema-conversion boilerplate from each server.

Switching from a record to an array creates a risk of allowing duplicate tool names; this PR introduces a function `ensureUniqueToolNames` (only called once, not repeated on each server) that enforces in type that tool names are unique among an MCP server.

No changes are intended to tool behavior or exposed metadata.

## Tests

- Covered by existing tests.

## Risk

- Medium.

## Deploy Plan

- Deploy front.

### DIFF
--- a/front/lib/actions/mcp_internal_actions/constants.ts
+++ b/front/lib/actions/mcp_internal_actions/constants.ts
@@ -108,6 +108,9 @@ import type { WhitelistableFeature } from "@app/types/shared/feature_flags";
 import type { ModelId } from "@app/types/shared/model_id";
 import type { Result } from "@app/types/shared/result";
 import { Err, Ok } from "@app/types/shared/result";
+import type { JSONSchema7 as JSONSchema } from "json-schema";
+import { z } from "zod";
+import { zodToJsonSchema } from "zod-to-json-schema";
 
 export const ADVANCED_SEARCH_SWITCH = "advanced_search";
 export const USE_SUMMARY_SWITCH = "useSummary";
@@ -243,7 +246,35 @@ export const MCP_SERVER_AVAILABILITY = [
 ] as const;
 export type MCPServerAvailability = (typeof MCP_SERVER_AVAILABILITY)[number];
 
-export const INTERNAL_MCP_SERVERS = {
+type HasUniqueNames<
+  T extends readonly { name: string }[],
+  Seen extends string = never,
+> = T extends readonly [
+  infer Head extends { name: string },
+  ...infer Tail extends readonly { name: string }[],
+]
+  ? Head["name"] extends Seen
+    ? false
+    : HasUniqueNames<Tail, Seen | Head["name"]>
+  : true;
+
+function ensureUniqueToolNames<
+  const T extends {
+    [K in InternalMCPServerNameType]: InternalMCPServerEntry<K>;
+  },
+>(
+  servers: T & {
+    [K in InternalMCPServerNameType]: HasUniqueNames<
+      T[K]["metadata"]["tools"]
+    > extends true
+      ? unknown
+      : { __duplicateToolNames: never };
+  }
+): T {
+  return servers;
+}
+
+export const INTERNAL_MCP_SERVERS = ensureUniqueToolNames({
   // Note:
   // ids should be stable, do not change them when moving internal servers to production as it would break existing agents.
 
@@ -1204,7 +1235,7 @@ export const INTERNAL_MCP_SERVERS = {
   // Using satisfies here instead of: type to avoid TypeScript widening the type and breaking the type inference for AutoInternalMCPServerNameType.
 } satisfies {
   [K in InternalMCPServerNameType]: InternalMCPServerEntry<K>;
-};
+});
 
 type IsRestrictedCallback = (params: {
   plan: PlanType;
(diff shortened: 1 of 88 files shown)


front/lib/actions/mcp_internal_actions/constants.ts
```diff
@@ -246,7 +249,35 @@ export const MCP_SERVER_AVAILABILITY = [
 ] as const;
 export type MCPServerAvailability = (typeof MCP_SERVER_AVAILABILITY)[number];
 
-export const INTERNAL_MCP_SERVERS = {
+type HasUniqueNames<
```
flvndvd: IMHO worth a comment, I do see some agents/humans eventually dropping it at some point!

front/lib/actions/mcp_internal_actions/constants.ts
```diff
@@ -246,7 +249,35 @@ export const MCP_SERVER_AVAILABILITY = [
...
+> = T extends readonly [
+  infer Head extends { name: string },
+  ...infer Tail extends readonly { name: string }[],
+]
+  ? Head["name"] extends Seen
+    ? false
+    : HasUniqueNames<Tail, Seen | Head["name"]>
+  : true;
+
+function ensureUniqueToolNames<
```
flvndvd, line 270: IMHO this should be used in the `createServer` directly, you'll save a lot of TS complexity and will be able to remove the recursivity.

front/lib/api/actions/servers/agent_memory/metadata.ts
```diff
@@ -11,8 +11,9 @@ export const AGENT_MEMORY_ERASE_TOOL_NAME = "erase_entries";
 export const AGENT_MEMORY_EDIT_TOOL_NAME = "edit_entries";
 export const AGENT_MEMORY_COMPACT_TOOL_NAME = "compact_memory";
 
-export const AGENT_MEMORY_TOOLS_METADATA = createToolsRecord({
-  [AGENT_MEMORY_RETRIEVE_TOOL_NAME]: {
+export const AGENT_MEMORY_TOOLS_METADATA = [
```
flvndvd, line 11: I'd probably go for using the unique type here and using explicit typing over a `satisfies`.

front/lib/actions/mcp_internal_actions/constants.ts
aubin-tchoi: actually let me remove this altogether. I'll add it on the `createServer` once we have a generic implementation for it

front/lib/actions/mcp_internal_actions/constants.ts
aubin-tchoi: agreed, will add it once we have a generic version of it

front/lib/api/actions/servers/agent_memory/metadata.ts
aubin-tchoi: struggling to add a unique type here without an intermediate function, will add in the generic `createServer`
IMHO we shouldn't have a separate object for the tools and just read from `AGENT_MEMORY_SERVER` (but that's a huge PR)

front/lib/actions/mcp_internal_actions/tool_definition.ts
```diff
@@ -27,9 +26,12 @@ export type ToolHandlerExtra = RequestHandlerExtra<
...
 export type ToolHandlerResult = Result<CallToolResult["content"], MCPError>;
 
-export type ToolHandlers<T extends Record<string, { schema: ZodRawShape }>> = {
-  [K in keyof T]: (
-    params: z.infer<z.ZodObject<T[K]["schema"]>>,
+export type ToolHandlers<ToolsList extends readonly ToolMeta[]> = {
+  [ToolName in ToolsList[number]["name"]]: (
+    // Type the params with the type inferred from the zod schema (z.ZodObject because it's a zod shape, not a schema).
+    params: z.infer<
+      z.ZodObject<Extract<ToolsList[number], { name: ToolName }>["schema"]>
```
aubin-tchoi, line 33: the change here is that we find the tool in the array using an `Extract` since it's now an array instead of a record

front/lib/actions/mcp_internal_actions/tool_definition.ts
```diff
@@ -129,28 +123,20 @@ export function buildClientTools<T extends Record<string, ClientToolMeta>>(
   );
 }
 
-export function buildTools<T extends Record<string, ToolMeta>>(
+export function buildTools<T extends readonly ToolMeta[]>(
```
aubin-tchoi, line 126: this one will be folded into the generic createServer: each tools/index file will export the handlers only, and createServer will find the handler by name

flvndvd (approved):

front/lib/actions/mcp_internal_actions/constants.ts
aubin-tchoi: added back a simpler version
couldn't do it on each tools array in the end, as we need to keep the inferred type for the handlers' inputs

PR #29116, opened 2026-07-17
[perf]: skip telemetry startup for warm sandboxes

flvndvd:
## Description

Zooming on pod functions poor performances, we can definitely speed things up. 

image

_One trace among many trace, order of magnitude is similar across many traces_

Warm sandbox executions restarted Fluent Bit on every readiness check, adding a few ms per command despite telemetry already running. This starts telemetry only after sandbox creation or wake, preserving logging while removing the redundant provider round trip.


id13 (approved):
LGTM

PR #29119, opened 2026-07-17
[front] chore: simplify internal MCP metadata types

aubin-tchoi:
## Description

Part of a multi step plan to allow defining in which context a tool can be ran on the tool metadata themselves in a declarative way.

The plan goes like this:
- Remove the boilerplate helpers `createToolRecords` and `buildTools`.
- Move the `createServer` one level up.
- Add a metadata to declare which contexts each tool supports (by default all contexts) + bake it in the generic `createServer`.

This PR starts by simplifying the typing of internal MCP servers: removes unused generics and types, cleaned up the existing ones (`InternalMCPServerEntryWithoutMetadata` exists in types but we never ever populate an internal mcp server without metadata so this just makes caller's lives harder).

## Tests

- Covered by existing tests.

## Risk

- Low. Type-only cleanup around internal MCP metadata.

## Deploy Plan

- Deploy front.

### DIFF
--- a/front/lib/actions/mcp_internal_actions/mcp_servers_metadata.test.ts
+++ b/front/lib/actions/mcp_internal_actions/mcp_servers_metadata.test.ts
@@ -33,7 +33,6 @@ import {
 } from "@app/lib/actions/mcp_internal_actions/constants";
 import { InMemoryWithAuthTransport } from "@app/lib/actions/mcp_internal_actions/in_memory_with_auth_transport";
 import { getInternalMCPServer } from "@app/lib/actions/mcp_internal_actions/servers";
-import type { InternalMCPToolType } from "@app/lib/actions/mcp_internal_actions/tool_definition";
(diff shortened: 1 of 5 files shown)


front/lib/actions/mcp_internal_actions/mcp_servers_metadata.test.ts
```diff
@@ -270,8 +269,7 @@ describe("MCP Servers Metadata Snapshot", () => {
     const allBilling: Record<string, Record<string, ToolBillingSnapshot>> = {};
 
     for (const serverName of [...AVAILABLE_INTERNAL_MCP_SERVER_NAMES].sort()) {
-      const tools = getInternalMCPServerMetadata(serverName)
-        .tools as InternalMCPToolType[];
+      const tools = getInternalMCPServerMetadata(serverName).tools;
```
flvndvd: `nit`

```suggestion
      const { tools } = getInternalMCPServerMetadata(serverName);
```

flvndvd (approved):

PR #29120, opened 2026-07-17
feat(front): link to Usage page for model access tiers from model providers page

Nils-Fedrigo:
## Description

Users complained that the place to find where we restrict models to users is not easy to find, so we add another path to find it from the Models provider page

Adds a "Model access tiers" section at the bottom of the Model Providers admin page (`/w/[wId]/model-providers`) that points admins to the Usage page, where model access tiers can be configured per workspace, group, or member.

The section is gated on the `models_picker` feature flag — the same flag the Usage page uses to render its tier controls — so the link only appears when there is something to manage. It includes a short explanation and a "Manage in Usage" button (with a trailing arrow) that navigates client-side to `/w/[wId]/usage`.
image


Nils-Fedrigo:

- #29120 ◀

PopDaph (approved):

PR #29121, opened 2026-07-17
Overlay sandbox function approvals over frames

spolu:
## Description

Keep visualization frames visible while sandbox function tools wait for approval or authentication. The blocking card now appears over the frame using the standard translucent dialog backdrop and is centered within the frame.

Screenshot 2026-07-17 at 17 55 42
Screenshot 2026-07-17 at 18 19 51

## Tests

Tested locally

## Risk

None

## Deploy Plan

- Deploy `front`

### DIFF
--- a/front/components/assistant/conversation/actions/VisualizationActionIframe.tsx
+++ b/front/components/assistant/conversation/actions/VisualizationActionIframe.tsx
@@ -249,12 +249,13 @@ function SandboxFunctionInvocation({
       blockedActionCard = null;
   }
 
-  // Covers the frame while the tool is blocked on user input; positioned within the component's
-  // relative root, same layering approach as the loading spinner overlay. Cards are keyed by the
-  // delivery eventId so consecutive events of the same type never reuse local card state.
+  // Overlays the frame while the tool is blocked on user input. Cards are keyed by the delivery
+  // eventId so consecutive events of the same type never reuse local card state.
   return (
-    <div className="absolute inset-0 z-10 flex items-center justify-center overflow-auto bg-panel-background p-4">
-      <div className="w-full max-w-xl">{blockedActionCard}</div>
+    <div className="absolute inset-0 z-10 flex items-center justify-center overflow-auto bg-muted-foreground/75 p-4 dark:bg-muted-background/75">
+      <div className="flex w-full max-w-xl justify-center">
+        {blockedActionCard}
+      </div>
     </div>
   );
 }


spolu:
@PopDaph what was your comment on that bit of CSS the first time I introduced it?

spolu:
cc @ykmsd in case you have any comment on this 🙏

spolu:
PMRR

front/components/assistant/conversation/actions/VisualizationActionIframe.tsx
```diff
@@ -249,12 +249,13 @@ function SandboxFunctionInvocation({
...
 
-  // Covers the frame while the tool is blocked on user input; positioned within the component's
-  // relative root, same layering approach as the loading spinner overlay. Cards are keyed by the
-  // delivery eventId so consecutive events of the same type never reuse local card state.
+  // Overlays the frame while the tool is blocked on user input. Cards are keyed by the delivery
+  // eventId so consecutive events of the same type never reuse local card state.
   return (
-    <div className="absolute inset-0 z-10 flex items-center justify-center overflow-auto bg-panel-background p-4">
-      <div className="w-full max-w-xl">{blockedActionCard}</div>
+    <div className="absolute inset-0 z-10 flex items-center justify-center overflow-auto bg-muted-foreground/75 p-4 dark:bg-muted-background/75">
```
PopDaph, line 255: We don't need the darkmode classes anymore \o/

PopDaph (commented):
👍🏻

front/components/assistant/conversation/actions/VisualizationActionIframe.tsx
spolu: Ah! Will remove!

spolu: uh I do see a difference without it?

PopDaph: Maybe cause one is bg-muted-foreground and the other is bg-muted-background? 
Since we migrated to tailwind v4 we do not put darkmode classes anymore 🙏🏻

spolu: Will dig 👍

ykmsd: indeed we don't need it but do we want to use different colors for light and dark here? Okay to use dark prefix sometimes but ideally we want to use (or add) semantic tokens

spolu: @ykmsd what is the desired end state here happy to address but not sure I understand what I'm supposed to do here? 😬🙏

ykmsd: sorry for not being clear 😆 
my question is: does `bg-muted-foreground/75` work for both light theme and dark theme? did you intentionally set `dark:bg-muted-background/75` because it didn't work for some scenario?

If it works any scenario, let's just have `bg-muted-foreground/75` only. If it doesn't work for some cases, I'm happy to take over from here 👍

spolu: Very clear. Def different but I think it "worked" (lighter fading). Will double check and move to that if looks good 👍

spolu: Done here: https://github.com/dust-tt/dust/pull/29352

PR #29123, opened 2026-07-17
feat(activation): ES-backed activation evaluator, wired into scheduler + CLI

frankaloia:
## Summary

* Adds the "is a user activated" evaluator and gates the activation nudge orchestration on it. The scheduled `activation_scheduler` Temporal workflow and the admin CLI run identical logic through one shared function.
**Definition:** a user is *activated* when, over a trailing **28 days**, they have **≥6 HVUC days spanning ≥3 distinct Monday-based UTC weeks**. An **HVUC day** = a **DAU day** (≥1 human-initiated organic message) on which the user also had **≥1 succeeded tool call that is advanced-cost, a frame touch (`interactive_content`), or `run_agent`**.
* This is generated via a composite Elasticsearch query

## Origin handling

- **Programmatic** origins (`api`, `*_programmatic`, `zapier`, …) are dropped from the query entirely — never part of the organic signal.
- **`triggered`** is organic and still counts toward the **HVUC** signal, but is excluded from the **DAU** half: otherwise a recurring automation (a trigger that runs `run_agent` daily) would activate a user with zero human interaction ("activated by construction"). A DAU day requires a real human message.

## Testing

* Validated against my dust-hives

## Risk

* Unintended ES load (intention was that all limits should be sufficient in preventing this)

## Deployment
1. Deploy front


front/lib/api/activation/queries/user_day_cells.ts
```diff
@@ -0,0 +1,222 @@
...
+import { Err, Ok } from "@app/types/shared/result";
+import type { estypes } from "@elastic/elasticsearch";
+
+// ---------------------------------------------------------------------------
+// Elasticsearch query: per-(user, day) activity facts.
+// ---------------------------------------------------------------------------
+//
+// One composite aggregation over agent_message_analytics_2 returning, per
+// (user, day): whether the user was a DAU, and whether they had an HVUC signal.
+// An HVUC signal = ≥1 succeeded tool call that is either advanced-cost, a frame
```
ElPicador: I would not use acronyms in the code - especially in comments. I think it's better to be explicit so the code can be understood by more people

front/lib/api/activation/evaluator.ts
```diff
@@ -0,0 +1,137 @@
...
+//   (interactive_content), or run_agent.
+
+export const MIN_HVUC_DAYS = 6;
+export const MIN_DISTINCT_WEEKS = 3;
+export const TRAILING_WINDOW_DAYS = 28;
+
+export interface ActivationEvidence {
+  // ISO dates (YYYY-MM-DD, UTC) of the qualifying HVUC days.
+  qualifyingDays: string[];
+  // Distinct Monday-based UTC ISO weeks spanned by the qualifying days.
```
ElPicador, line 22: why not using dates directly?

front/lib/api/activation/evaluator.ts
```diff
@@ -0,0 +1,137 @@
...
+    asOf?: Date;
+  }
+): Promise<Result<Map<string, ActivationResult>, Error>> {
+  // The ES query is scoped by workspace_id; guard that it matches the caller's
+  // authenticated workspace so a mismatched auth can't read another workspace.
+  assert(
+    auth.getNonNullableWorkspace().sId === workspaceId,
+    "evaluateActivation: workspaceId must match the authenticated workspace"
+  );
+
```
ElPicador:
```suggestion
export async function evaluateActivation(
  auth: Authenticator,
  {
    userIds,
    asOf = new Date(),
  }: {
    userIds: string[];
    asOf?: Date;
  }
): Promise<Result<Map<string, ActivationResult>, Error>> {
const workspaceId = auth.getNonNullableWorkspace().sId
```

you don't need that, and it's not consistent to how we use auth/workspaces in the code. 

SeE: https://github.com/dust-tt/dust/pull/29123/changes#diff-ba6489dee0d6d4cc859763f91bf04d02c5752cced95f7428571c38f3ea988c97R76 (it's the same)

ElPicador (approved):
Nice!

front/lib/api/activation/queries/user_day_cells.ts
frankaloia: Agreed

front/lib/api/activation/evaluator.ts
frankaloia: Yea, would be cleaner

front/lib/api/activation/queries/user_day_cells.ts
```diff
@@ -0,0 +1,233 @@
...
+          range: {
+            timestamp: {
+              gte: windowStart.toISOString(),
+              lt: windowEnd.toISOString(),
+            },
+          },
+        },
+        { terms: { status: AGENT_MESSAGE_STATUSES_TO_TRACK } },
+      ],
+      must_not: [
```
flvndvd, line 148: ES is actually offering pretty poor performances on `must_not` clauses, they often advise using a filter instead and looking for specific values.

front/lib/api/activation/queries/user_day_cells.ts
```diff
@@ -0,0 +1,233 @@
...
+// ---------------------------------------------------------------------------
+// Elasticsearch query: per-(user, day) activity facts.
+// ---------------------------------------------------------------------------
+//
+// One composite aggregation over agent_message_analytics_2 returning, per
+// (user, day): whether the user was a daily active user, and whether they had a
+// high-value use case signal (≥1 succeeded tool call that is either
+// advanced-cost, a frame touch (interactive_content), or run_agent).
+
+const FRAME_SERVER_NAME = "interactive_content";
```
flvndvd, line 19: IIRC we already have variable for it, if we don't can we add on in the server and import it here. So we avoid potential drifts + keep clear responsibilities.

front/lib/api/activation/queries/user_day_cells.ts
```diff
@@ -0,0 +1,233 @@
...
+const RUN_AGENT_SERVER_NAME = "run_agent";
```
flvndvd, line 20: Same for run agent.

front/lib/api/activation/queries/user_day_cells.ts
```diff
@@ -0,0 +1,233 @@
...
+    );
+  }
+  for (const userId of uniqueUserIds) {
+    factsByUser.set(userId, []);
+  }
+
+  const query: estypes.QueryDslQueryContainer = {
+    bool: {
+      filter: [
+        { term: { workspace_id: workspaceId } },
```
flvndvd, line 136: We use to have this one function that was always adding the `workspace_id` clause to ensure that all queries must be scoped to a workspace (similar to what we are doing at the DB level), any chance we can re-use it, IIRC, it  was this one:

https://github.com/dust-tt/dust/blob/48e6f055f8bf23a71569ddd52514e370d4dcd30e/front/lib/api/assistant/observability/utils.ts#L90

Maybe it did not age well, and people dropped a few other business concerns in it, but IMHO we must ensure that ALL queries perform on ES are based on this one very important clause.

front/lib/api/activation/queries/user_day_cells.ts
```diff
@@ -0,0 +1,233 @@
...
+              time_zone: "UTC",
+            },
+          },
+        },
+      ],
+      ...(afterKey ? { after: afterKey } : {}),
+    };
+
+    const result = await searchAnalytics<never, UserDayCellsAggs>(query, {
+      size: 0,
```
flvndvd, line 200: We usually also provide the query on the workspace id here to limit the documents on which the aggregations can run if too loose.

front/lib/api/activation/orchestrator.ts
```diff
@@ -44,39 +46,138 @@ export async function determineEligibleActivationUsers({
       spaces
     );
 
-  const eligible: NudgePlan[] = [];
-  const skipped: SkippedUser[] = [];
-
+  // Build the candidate (pod, member) list and the deduped set of user sIds to
+  // evaluate in a single batch.
+  const candidates: { podId: string; spaceId: string; userId: string }[] = [];
+  const userSIds = new Set<string>();
```
flvndvd, line 52: No `SId` 😛

front/lib/api/activation/orchestrator.ts
```diff
@@ -44,39 +46,138 @@ export async function determineEligibleActivationUsers({
...
+      userSIds.add(member.sId);
+    }
+  }
+
+  if (candidates.length === 0) {
+    return new Ok({ eligible: [], skipped: [] });
+  }
+
+  const activationResult = await evaluateActivation(auth, {
+    userIds: [...userSIds],
```
flvndvd, line 77: Any reason for spreading here?

front/lib/api/activation/orchestrator.ts
```diff
@@ -44,39 +46,138 @@ export async function determineEligibleActivationUsers({
...
+  asOf,
+}: {
+  workspaceId: string;
+  userId?: string | null;
+  dryRun: boolean;
+  asOf?: Date;
+}): Promise<Result<OrchestratorResult, Error>> {
+  // Activation conversations live in Pods, which are restricted spaces: request
+  // all groups so admin auth can read/write them.
+  const auth = await Authenticator.internalAdminForWorkspace(workspaceId, {
```
flvndvd, line 134: Couldn't we force the dangerous auth object earlier in the pipeline, and just pass it to all those call sites?

front/lib/api/activation/orchestrator.ts
```diff
@@ -44,39 +46,138 @@ export async function determineEligibleActivationUsers({
...
+  }
+
+  // Emit one activation event per eligible (pod, user), carrying the target
+  // user's sId so only that user's pod trigger matches and fires.
+  if (plan.eligible.length === 0) {
+    return new Ok(plan);
+  }
+  const uniqueSpaceIds = [...new Set(plan.eligible.map((p) => p.spaceId))];
+  const pods = await SpaceResource.fetchByIds(auth, uniqueSpaceIds);
+  const podBySId = new Map(pods.map((pod) => [pod.sId, pod]));
```
flvndvd, line 158:
```suggestion
  const podById = new Map(pods.map((pod) => [pod.sId, pod]));
```

front/lib/api/activation/trigger.ts
```diff
@@ -19,14 +19,23 @@ import { randomUUID } from "crypto";
...
+// Filtering on both podId and userId ensures a given event only fires the target user.
+// Note, these filter values are visible to the user in the trigger's configuration.
+function activationTriggerFilter(podSId: string, userId: string): string {
+  return `(and (eq "${ACTIVATION_POD_ID_FIELD}" "${podSId}") (eq "${ACTIVATION_USER_ID_FIELD}" "${userId}"))`;
 }
 
-function activationEventBody(podSId: string): Record<string, unknown> {
-  return { [ACTIVATION_POD_ID_FIELD]: podSId };
+function activationEventBody(
+  podSId: string,
```
flvndvd, line 32:
```suggestion
  podId: string,
```

flvndvd (commented):
Hopefully not too late to the party 🙈

PR #29124, opened 2026-07-18
Fix sync for free seats on legacy plans

tdraier:
## Description

Fixes a bug found on a workspace with a legacy contract (+ a pending Business contract mid-migration): free-seat credits and per-user credit-balance alerts (`free-xxxx`) were being created for *every* member, even though none of them actually have a `free` seat type.

Root cause: `currentFreeUserIds` was computed via `desiredSIdsAt("free", baseMs)`, but that helper's `userSeatType === subSeatType || legacy` fallback exists so a legacy contract's `none` members get folded into its one flat `workspace` billing subscription — correct for the seat-quantity reconcile loop, but it applied to *any* `subSeatType`, including `"free"`. So on any legacy-plan workspace it always matched everyone, regardless of actual seat type.

Fix: `currentFreeUserIds` is now computed directly from `seatTypeAt(...) === "free"`, without the legacy fallback, so it correctly resolves to nobody on a legacy contract with no free-seat members.

While in there, also scoped `grantFreeSeatCredits`'s alert-upsert step to only newly-free users (those Metronome's "free" subscription doesn't already show as assigned) instead of re-upserting alerts for every currently-free user on every sync — same accepted low-stakes/self-correcting tradeoff already used on the revoke side.

## Tests

Existing `lib/metronome/seats.test.ts` / `seats_sync.test.ts` suites pass (70 tests).

## Risk

Pure correctness fix + a performance scoping change with an accepted, already-precedented tradeoff (see above). No schema/migration changes, no feature flag.

### DIFF
--- a/front/lib/metronome/seats.ts
+++ b/front/lib/metronome/seats.ts
@@ -705,11 +705,13 @@ async function grantFreeSeatCredits({
   metronomeCustomerId,
   workspaceId,
   userIds,
+  alreadyAssignedFreeUserIds,
   startingAt,
 }: {
   metronomeCustomerId: string;
   workspaceId: string;
   userIds: string[];
+  alreadyAssignedFreeUserIds: Set<string>;
   startingAt: Date;
 }): Promise<void> {
   if (userIds.length === 0) {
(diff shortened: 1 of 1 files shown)


tdraier:

- #29124 ◀

tdraier:
PMRR

PopDaph (commented):
👍🏻

PR #27403, opened 2026-06-16
upgrade tailwind v4, compile the way to build css, remove dark and s prefixes

ykmsd:
## Description
This PR is to upgrade tailwind to v4 and change the way to bundle css so that we can get rid of `dark:` and `s-` prefixes. 

Right now we precompiled sparkle css, and then each app ships separate styles and we stitch together in the browser. 

e.g.
```
import "styles/global.css";              // @tailwind base 
import "@dust-tt/sparkle/dist/sparkle.css"; // precompiled sparkle 
import "styles/components.css";   
```

In this PR, we got rid of `tailwind.config.js` (it's still supported in v4 but tailwind encourages to use native css and not js for config). Each app runs a single Tailwind pass that scans sparkle and the app together, and emitting one CSS. 

So for example in front-spa index.css:

```
// import sparkle css entry (containing theme, utilities, tokens etc)
@import "@dust-tt/sparkle/styles/tailwind.css";
// import front specific css styles
@import "@dust-tt/front/styles/theme-extras.css";

// scanning all the relevant files 
@source "./**/*.{ts,tsx}";
@source "../../front/components/**/*.{ts,tsx}";
@source "../../front/lib/**/*.{ts,tsx}";
```
 
The only ugly thing is custom container query definition, I need to inline it everywhere and I'm pretty sure there is a better way but I couldn't figure it out, I will do it as a follow up. 

The bundled CSS file is more or less the same (prod client.css 39.5k, preview 34.7k, which actually I thought it would be much less so there might be an opportunity to optimize)    

> [!NOTE]  
> I didn't update viz's tailwind version. Viz shipped all the available tailwind classnames to be able to render any llm generated code and since v4 changed lots of classnames the old frames can be broken unless we can ship both v3 and v4 classnames.

PR #29125, opened 2026-07-18
[Sparkle] Fix Mermaid graph rendering

philipperolet:
## Description

Fixes the Mermaid regression introduced by https://github.com/dust-tt/dust/pull/27403.

Context: [sample conversation](https://app.dust.tt/w/0ec9852c2f/conversation/5mfZBN1KPJ)

Tailwind v4 changed its exported palette from hex to OKLCH values, which Mermaid cannot parse.
Sparkle now uses the browser to convert the live Tailwind palette to sRGB hex before initializing
Mermaid.

The previous global `mermaid.run()` call could also race with React Strict Mode: the replayed effect
replaced Mermaid's temporary SVG while it was measuring the graph, leaving the source text visible.
The component now renders off-node through Mermaid's direct API and installs only the completed SVG.
A Storybook interaction assertion verifies that Mermaid produces an SVG.

## Risks

Blast radius: Mermaid diagrams rendered in Markdown
Risk: low

## Deploy Plan

- deploy front

## Tests

- [x] Reproduced the Mermaid initialization failure with the Tailwind v4 palette
- [x] Verified Mermaid initializes with an sRGB hex palette
- [x] Reproduced the Strict Mode rendering race in the local sample conversation
- [x] Verified the sample conversation renders an SVG in the Hive

### DIFF
--- a/sparkle/src/components/markdown/CodeBlockWithExtendedSupport.tsx
+++ b/sparkle/src/components/markdown/CodeBlockWithExtendedSupport.tsx
@@ -14,8 +14,24 @@ import {
   sameNodePosition,
 } from "@sparkle/components/markdown/utils";
 import { Stars02, Terminal } from "@sparkle/icons/v2-stroke";
+import { cssColorToHex } from "@sparkle/lib/colors";
 import { cn } from "@sparkle/lib/utils";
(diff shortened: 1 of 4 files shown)


spolu (approved):

ykmsd (approved):
🙏

PR #29128, opened 2026-07-19
[front] perf: limit rendered capability entries

spolu:
## Description

Limit capability menus to the first 50 ranked matches and reuse a normalized search index in the input-bar picker.

Previously, each picker search keystroke rebuilt display metadata and instantiated every matching capability row, including icons, buttons, Radix items, and registry refs. Large workspaces therefore paid rendering cost proportional to their full capability count.

The picker and slash menus still search and rank the complete client-side dataset, but only materialize the first 50 results.

## Tests

- `NODE_ENV=test npm test components/editor/extensions/shared/SlashCommandCapabilitiesItems.test.ts`

## Risk

Low. Users can still find capabilities beyond the first 50 by refining their search, but can no longer browse every match in a single unfiltered list.

## Deploy Plan

- deploy `front`

### DIFF
--- a/front/components/assistant/CapabilitiesPicker.tsx
+++ b/front/components/assistant/CapabilitiesPicker.tsx
@@ -1,7 +1,8 @@
 import { CreateMCPServerDialog } from "@app/components/actions/mcp/create/CreateMCPServerDialog";
 import {
-  matchesSlashCommandCapabilityQuery,
-  sortSlashCommandCapabilityMatches,
+  type CapabilitySearchIndexItem,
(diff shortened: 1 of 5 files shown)


spolu:
PMRR very open to feedback since this was mostly agent coded but the gains are obvious on our workspace.

ykmsd:
I need time to think about it what is the best way to optimize it, but meanwhile if it works 👍

spolu:
Roger that 🫡 

Please feel free to iterate ofc 👍

`/` is back to being usable now which was my top goal :-)

PR #29129, opened 2026-07-19
Fix free seat credits cleanup + ui

tdraier:
## Description

A few fixes around free-seat credits, found while cleaning up after the legacy-contract sync bug fixed in #29124.

- **`ChangeSeatModal` showed no options at all for some workspaces** (`components/workspace/ChangeSeatModal.tsx`): the modal seeds its initial Monthly/Yearly tab from the *current* seat's own billing frequency, but `"free"` is filtered out of the selectable seat types before that — so on a workspace whose only selectable seat types are e.g. `pro_yearly`/`max_yearly` (both "annual"), a free-seat member's own frequency (e.g. "monthly") could seed a tab with nothing in it, and the Monthly/Yearly switch only renders when more than one tab actually has options — so there was no way to escape it. Now the initial tab only trusts the current seat's frequency when it's actually a populated bucket, otherwise falls back to the first available one (the same fallback that already existed, just not reached here).

- **`revokeFreeSeatCreditsForExFreeUsers` never actually revoked or cleared anything** (`lib/metronome/seats.ts`): two stacked bugs.
  1. Its filter re-stripped an already-normalized user id (`listCustomerPerUserCreditBalances` already returns plain sIds), so `fromFreeMetronomeUserId` always returned `null` and the revoke list was always empty.
  2. Even fixed, the alert-clear call used the plain user id, but alerts are created with the free-prefixed id (`grantFreeSeatCredits` → `upsertPerUserCreditBalanceAlerts`) — so it was targeting a uniqueness key that never existed.

  Both fixed: the filter now compares plain ids directly, and the alert clear now re-derives the prefixed id to match how it was created.

- **`listCustomerPerUserCreditIds`** (`lib/metronome/client.ts`, new): a lighter variant of `listCustomerPerUserCreditBalances` that skips Metronome's balance computation (`include_balance: false`) for the two call sites that only ever needed credit ids to revoke, not balances — `revokeFreeSeatCreditsForExFreeUsers` and the new cleanup script below. `listCustomerPerUserCreditUserIds` (used by `grantFreeSeatCredits`'s already-granted check) is now a thin wrapper over it instead of duplicating the same listing/filtering with a second, different (and only coincidentally equivalent) user-id extraction path.

- **`scripts/revoke_stray_free_seat_credits.ts`** (new): one-off cleanup for the stray `free-<sId>` credits/alerts created by the legacy-contract bug — cross-checks live DB membership state against Metronome's per-user free-seat credits and revokes/clears anything that no longer matches a real free-seat member. Iterates every workspace with a `metronomeCustomerId` (or a single one via `--workspaceId`), rate-limited to stay under Metronome's published RPS cap, dry-run by default (`--execute` to apply).

## Tests

`lib/metronome` suite (client.test.ts, seats.test.ts, seats_sync.test.ts — 92 tests) passes. No test coverage for `ChangeSeatModal` (no existing test infra for this component).

## Risk

The credit/alert fixes only change previously-broken (no-op) behavior — nothing to regress. The `listCustomerPerUserCreditUserIds` refactor changes its output format from raw/prefixed to normalized ids; its one caller was updated to match. No schema/migration changes, no feature flag.

## Deploy Plan

Standard deploy. Once live, run `scripts/revoke_stray_free_seat_credits.ts` (dry-run first) against the affected workspace(s) to clean up the stray credits/alerts left by the earlier bug.

### DIFF
--- a/front/lib/metronome/client.ts
+++ b/front/lib/metronome/client.ts
@@ -2631,66 +2631,26 @@ export async function addPerUserCreditToCustomer({
 }
 
 /**
- * Return the set of seat user sIds that already have a per-user customer credit
- * named `creditName`. Includes archived/expired credits so past grants are not
- * re-issued — the uniqueness key also enforces this server-side.
+ * Shared core for listing a customer's per-user credits of a given
+ * `contractCreditType`, keyed by user sId (from the `DUST_PER_USER_CREDIT_USER`
+ * custom field — old plain-sId and new free-prefixed formats collapse to the
+ * same key). Includes archived/expired credits so a past grant is never
+ * mistaken for missing, and a fully-consumed credit still appears (balance 0)
+ * rather than being absent.
+ *
+ * `includeBalances: false` skips requesting `include_balance` from Metronome,
+ * which otherwise computes a live balance for every credit and makes the
+ * call meaningfully heavier — `balanceAwu` is meaningless (always 0) in that
+ * case; only pass `true` when a caller actually reads it.
  */
-export async function listCustomerPerUserCreditUserIds({
+async function listCustomerPerUserCreditEntries({
   metronomeCustomerId,
   contractCreditType,
+  includeBalances,
 }: {
   metronomeCustomerId: string;
   contractCreditType: ContractCreditType;
-}): Promise<Result<Set<string>, Error>> {
-  if (!config.getMetronomeApiKey()) {
-    return new Ok(new Set());
-  }
-
-  const client = getMetronomeClient();
-
-  try {
-    const userIds = new Set<string>();
-    for await (const entry of client.v1.customers.credits.list({
-      customer_id: metronomeCustomerId,
-      include_balance: false,
-      include_archived: true,
-    })) {
-      if (
-        entry.contract ||
-        entry.custom_fields?.[CONTRACT_CREDIT_TYPE_CUSTOM_FIELD_KEY] !==
-          contractCreditType
-      ) {
-        continue;
-      }
-      for (const specifier of entry.specifiers ?? []) {
-        const userId = specifier.presentation_group_values?.user_id;
-        if (userId) {
-          userIds.add(userId);
-        }
-      }
-    }
-    return new Ok(userIds);
-  } catch (err) {
-    const error = normalizeError(err);
-    logger.error(
-      { error, metronomeCustomerId },
-      "[Metronome] Failed to list per-user customer credits"
-    );
-    return new Err(error);
-  }
-}
-
-/**
- * Return the live AWU balance of each free-seat per-user customer credit, keyed
- * by user sId (from the `DUST_PER_USER_CREDIT_USER` custom field). Only active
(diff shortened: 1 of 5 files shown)


tdraier:

- #29129 ◀

front/lib/metronome/seats.ts
```diff
@@ -855,10 +845,14 @@ async function revokeFreeSeatCreditsForExFreeUsers({
...
       }
+      // Alerts are created with the free-prefixed user id (see
+      // `grantFreeSeatCredits`'s `upsertPerUserCreditBalanceAlerts` call) —
+      // must clear with the same form or `clearMetronomeAlert` targets a
+      // uniqueness key that was never created.
       const clearResult = await clearPerUserCreditBalanceAlerts({
         metronomeCustomerId,
         workspaceId,
-        userId,
+        userId: toFreeMetronomeUserId(userId),
```
PopDaph, line 855: Do we have some created before user the free prefixed? And do we need to clear them?

front/lib/metronome/client.ts
```diff
@@ -2631,55 +2631,88 @@ export async function addPerUserCreditToCustomer({
...
+ * `DUST_PER_USER_CREDIT_USER` custom field) — normalized the same way as
+ * `listCustomerPerUserCreditBalances` (old plain-sId and new free-prefixed
+ * formats collapse to the same key), but without requesting balances.
+ * Metronome computes a balance for every credit on `include_balance: true`,
+ * which makes that call meaningfully heavier — use this instead wherever a
+ * caller only needs to know which users have a credit and its id(s) to act
+ * on (e.g. revoke), not the balance itself.
  */
-export async function listCustomerPerUserCreditUserIds({
+export async function listCustomerPerUserCreditIds({
```
PopDaph: NIT: listCustomerPerUserCreditIds and listCustomerPerUserCreditBalances now share their entire body: same pagination loop, same entry.contract skip, same custom-field checks, same fromFreeMetronomeUserId(rawUserId) ?? rawUserId normalization. 
The only real difference is what gets accumulated (ids vs ids + balance sums).
Should we share some code? Feels relevant here since iiuc the original bug came from two "coincidentally equivalent" extraction paths drifting apart.

PopDaph (approved):

front/lib/metronome/seats.ts
tdraier: No credits have been created without the free- prefix, however the revoke was not adding the prefix so not removing them. We have a huge number of credits created (including the ones created because of what has been fixed here : https://github.com/dust-tt/dust/pull/29124 ), that have not been revoked because of this - but that's the goal of the script to clean them up.

front/lib/metronome/client.ts
tdraier: oops I've updated `listCustomerPerUserCreditUserIds` to reuse  `listCustomerPerUserCreditIds`  but not `listCustomerPerUserCreditBalances` .. missed it, thanks

PR #29130, opened 2026-07-19
[front] perf: throttle capability refreshes

spolu:
## Description

Reduce repeated capability-list requests by applying a one-minute SWR deduplication window and a five-minute focus revalidation throttle.

The policy is scoped to the preloaded capabilities picker and input-bar slash menu. The shared skills, spaces, and available MCP server hooks now accept per-consumer SWR options, leaving other screens' refresh behavior unchanged.

## Tests

N/A, tested locally.

## Risk

Low. Initial loads and explicit mutations still revalidate immediately; only automatic duplicate and focus-triggered refreshes are throttled.

## Deploy Plan

- deploy `front`

### DIFF
--- a/front/lib/swr/capabilities.ts
+++ b/front/lib/swr/capabilities.ts
@@ -1,0 +1,9 @@
+import type { SWRConfiguration } from "swr";
+
+const CAPABILITIES_DEDUPING_INTERVAL_MS = 60 * 1000;
+const CAPABILITIES_FOCUS_THROTTLE_INTERVAL_MS = 5 * 60 * 1000;
+
+export const CAPABILITIES_SWR_OPTIONS = {
+  dedupingInterval: CAPABILITIES_DEDUPING_INTERVAL_MS,
+  focusThrottleInterval: CAPABILITIES_FOCUS_THROTTLE_INTERVAL_MS,
+} satisfies SWRConfiguration;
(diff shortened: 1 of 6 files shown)


spolu:
PMRR

front/lib/swr/capabilities.ts
```diff
@@ -0,0 +1,9 @@
+import type { SWRConfiguration } from "swr";
+
+const CAPABILITIES_DEDUPING_INTERVAL_MS = 60 * 1000;
+const CAPABILITIES_FOCUS_THROTTLE_INTERVAL_MS = 5 * 60 * 1000;
+
+export const CAPABILITIES_SWR_OPTIONS = {
+  dedupingInterval: CAPABILITIES_DEDUPING_INTERVAL_MS,
+  focusThrottleInterval: CAPABILITIES_FOCUS_THROTTLE_INTERVAL_MS,
```
aubin-tchoi, line 8: might as well disable revalidateIfStale and revalidateOnFocus tbh

aubin-tchoi (approved):

front/lib/swr/capabilities.ts
spolu: hmmm why?

ykmsd: I don't think we should disable? We still want to revalidate in case something has changed, and iiuc this throttles now controlling the frequency?

ykmsd (commented):
👍

front/lib/swr/capabilities.ts
aubin-tchoi: if we only revalidate once every hour isn't more or less the same as considering the data static?

ykmsd: it's once every minute right? 🤔

spolu: Yes this moves to:

- 1m for dedup (2 hoooks asking for the same thing. Now N hooks asking for the same thing will get the same data for 1mn)
- 5m for focus (if I switch tab back and forth will get refreshed only if stale for more than 5mn)

aubin-tchoi: ahh I misread that, though it was once every hour

aubin-tchoi: nvm then!

PR #29133, opened 2026-07-20
Collect egress forwarder failures in DD

flvndvd:
## Description

The egress health check currently tells us when the forwarder is unavailable and restarted (it happens [pretty often](https://app.datadoghq.eu/logs?query=%40event%3Aegress.health_fail&agg_m=count&agg_m_source=base&agg_t=count&cols=host%2Cservice&event=AwAAAZ98oPi0w_La2QAAABhBWjk4b1A4MkFBQkdxaThKZDFicTl3QWMAAAAkMDE5ZjdjYzItZDc0ZC00NDQ5LThiZGYtMWYwZmFhYWM4NDQ1AAmdTw&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=stream&from_ts=1783929885210&to_ts=1784534685210&live=true)), but the
underlying process output remains in the sandbox and is not shipped. This makes it difficult to understand why the forwarder stopped working.

This change tails the forwarder log and sends failures to Datadog with the existing sandbox, conversation, space, and workspace context. Tail offsets are persisted across Fluent Bit restarts, newly created log files are discovered, truncation caused by forwarder restarts is supported, buffering is capped, and oversized lines are skipped.

The forwarder (through dsbx) emits structured tracing JSON with its message nested under fields.message. The telemetry
enrichment now promotes that message into Datadog’s main message field, maps the forwarder’s log level to the appropriate Datadog status, and preserves raw non-JSON output so startup crashes and fatal errors are still visible.

The sandbox image version is bumped so the updated telemetry configuration is deployed. This does not
change egress enforcement, health checks, restart behavior, or policy handling.

### DIFF
--- a/front/lib/api/sandbox/image/telemetry/enrich.lua
+++ b/front/lib/api/sandbox/image/telemetry/enrich.lua
@@ -3,7 +3,13 @@ function enrich_message(tag, timestamp, record)
     local command = record["command"]
     local pid = record["pid"]
     local exit_code = record["exit_code"]
-    local msg = record["msg"] or record["message"]
+    local structured_fields = record["fields"]
+    local structured_message = nil
+    if type(structured_fields) == "table" then
+        structured_message = structured_fields["message"]
+    end
+    local msg = record["msg"] or record["message"] or structured_message or
+        record["log"]
     local data = record["data"]
     local tool = record["tool"]
     local profile = record["profile"]
@@ -38,6 +44,21 @@ function enrich_message(tag, timestamp, record)
         record["status"] = litestream_level_map[litestream_level] or "warn"
     end
 
+    -- dsbx uses tracing-subscriber's JSON formatter. Its message is nested in
+    -- `fields`, while the severity is the top-level `level` value.
+    local forwarder_level = nil
+    if tag == "sandbox.egress_forwarder" then
+        forwarder_level = record["level"]
+        local forwarder_level_map = {
+            ERROR = "error",
+            WARN = "warn",
+            INFO = "info",
+            DEBUG = "debug",
+            TRACE = "debug"
+        }
+        record["status"] = forwarder_level_map[forwarder_level] or "warn"
+    end
+
     local severity_map = {
         process_start = "info",
         process_exit = "info",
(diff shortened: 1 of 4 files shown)


front/lib/api/sandbox/image/telemetry/enrich.lua
```diff
@@ -3,7 +3,13 @@ function enrich_message(tag, timestamp, record)
     local command = record["command"]
     local pid = record["pid"]
     local exit_code = record["exit_code"]
-    local msg = record["msg"] or record["message"]
+    local structured_fields = record["fields"]
```
flvndvd, line 6: IMHO we should really move away from json logs in `dsbx` it yields to pretty complex parsing...

id13: Not sure I got it right

We parse logs in dsbx ? 

Or you are concerned about this lua code's complexity?

flvndvd: I ended up doing this, because logs from dsbx are formatted as JSON, so we need this plumbing to make them DD friendly. 

I was challenging using JSON on the dsbx side at first.

fontanierh: For forwarder logs JSOn was convenient because we're parsing them to surface some of them in the tool response

front/lib/api/sandbox/image/telemetry/fluent-bit.conf
```diff
@@ -14,6 +14,20 @@
     read_from_tail   off
     mem_buf_limit    20MB
 
+# Collect the dsbx forwarder's structured stdout/stderr. The file is created
```
fontanierh, line 17: Hmmm thinking about it Iam not sure we actually want this going to datadog. Domains might be sensitive info

flvndvd: Any idea on how to capture why it's failing then?

fontanierh: The thing is, the vast majority of times it is failing because domain isn't allow-listed or because the request genuinely failed (you're not authenticated, API failed, 404 etc..)

flvndvd: Ok closing, will address in another way if needed.

PR #29134, opened 2026-07-20
feat: add ServiceNow OAuth provider (core + front)

thomasvicaire:
## Description

First of a 2-PR stack for the ServiceNow MCP integration (split out per review feedback: OAuth first for focused review, then the MCP skeleton on top).

Adds OAuth 2.0 support for ServiceNow, modeled on the existing Databricks provider since both are self-hosted/dedicated-instance platforms requiring a per-tenant OAuth app (instance URL + client_id + client_secret), rather than a single shared multi-tenant app like Slack/Notion/Google.

- `core/src/oauth/providers/servicenow.rs`: authorization-code exchange and refresh against `{instance}/oauth_token.do`
- `front/lib/api/oauth/providers/servicenow.ts`: setup URL construction, credential validation, connection metadata handling
- Registers `servicenow` across the OAuth provider registries (core `ConnectionProvider`/`CredentialProvider` enums, front provider strategy map, SDK `OAuthProviderSchema`)

No MCP server consumes this provider yet - that's `#2` in the stack ([thomasvicaire/servicenow-mcp-skeleton](https://github.com/dust-tt/dust/compare/thomasvicaire/servicenow-oauth...thomasvicaire/servicenow-mcp-skeleton)), so this PR alone adds no user-visible surface.

## ServiceNow OAuth app setup

When creating the OAuth application in **System OAuth -> Application Registry**, add the auth scope **`useraccount`**. The app still needs the Dust redirect URI, client ID, client secret, and the per-instance OAuth endpoints.

## Tests

- `cargo check` / `cargo clippy`: clean
- `npx tsgo --noEmit`: clean
- **Verified end-to-end against a live ServiceNow Personal Developer Instance**: registered a real OAuth app, walked the authorization-code flow manually, and confirmed both the token exchange (`finalize`) and refresh-token exchange (`refresh`) return the exact `access_token`/`refresh_token`/`expires_in` shape our Rust provider parses.

## Risk

Very low. Purely additive (new enum variants with matching match arms, new provider files, new registry entries). Nothing existing changes behavior, and nothing yet references this provider from a live code path.

## Deploy Plan


thomasvicaire:
**Manual OAuth verification — done ✅**

Tested end-to-end against a live ServiceNow Personal Developer Instance (free, self-service at [developer.servicenow.com/dev.do](https://developer.servicenow.com/dev.do) → "Start Building" to spin one up):

- Registered an OAuth app on the instance (System OAuth → Application Registry)
- Ran the actual compiled `ServicenowConnectionProvider::finalize()` and `::refresh()` (not a mock) against the real instance
- `finalize()`: authorization-code exchange returns the expected `access_token` / `refresh_token` / `expires_in`
- `refresh()`: refresh-token exchange also returns the expected shape, confirms token renewal works
- Re-verified after the GEN9 fix (`expires_in` → `expires_in_seconds`) to make sure the rename didn't regress anything

Both pass. OAuth layer is confirmed working end-to-end.

core/src/oauth/providers/servicenow.rs
```diff
@@ -0,0 +1,254 @@
...
+        let client_secret = content
+            .get("client_secret")
+            .and_then(|v| v.as_str())
+            .ok_or_else(|| anyhow!("Missing client_secret in ServiceNow credential"))?;
+
+        Ok((client_id.to_string(), client_secret.to_string()))
+    }
+}
+
+/// ServiceNow OAuth documentation: https://docs.servicenow.com/bundle/latest-release/page/administer/security/concept/c_OAuthApplications.html
```
aubin-tchoi: link doesn't work btw
probably this one? https://www.servicenow.com/docs/r/platform-security/authentication/c_OAuthApplications.html

front/lib/api/oauth/providers/servicenow.ts
```diff
@@ -0,0 +1,206 @@
...
+        const oauthApi = new OAuthAPI(config.getOAuthAPIConfig(), logger);
+        const connectionRes = await oauthApi.getAccessToken({
+          connectionId: oauthConnectionIdRes.value,
+        });
+        if (connectionRes.isErr()) {
+          throw new Error(
+            "Failed to get connection metadata: " + connectionRes.error.message
+          );
+        }
+        const connection = connectionRes.value.connection;
```
aubin-tchoi: nit
```suggestion
        const { connection } = connectionRes.value;
```

front/lib/api/oauth/providers/servicenow.ts
```diff
@@ -0,0 +1,206 @@
...
+
+        return {
+          client_id: connection.metadata.client_id,
+          servicenow_instance_url: connection.metadata.servicenow_instance_url,
+          ...restConfig,
```
aubin-tchoi:
```suggestion
          ...restConfig,
          client_id: connection.metadata.client_id,
          servicenow_instance_url: connection.metadata.servicenow_instance_url,
```

front/lib/api/oauth/providers/servicenow.ts
```diff
@@ -0,0 +1,206 @@
...
+    return getStringFromQuery(query, "state");
+  }
+
+  isExtraConfigValid(extraConfig: ExtraConfigType, useCase: OAuthUseCase) {
+    if (useCase === "personal_actions") {
+      // If we have an mcp_server_id it means the admin already setup the connection and we have
+      // everything we need, otherwise we'll need the client_id, client_secret, and servicenow_instance_url.
+      if (extraConfig.mcp_server_id) {
+        return true;
+      }
```
aubin-tchoi, line 84: the implementation here does not seem consistent with the comment, if we don't have an `mcp_server_id` here we don't check `client_id`, `client_secret` and `servicenow_instance_url`, we actually reject if we do have them all

front/lib/api/oauth/providers/servicenow.ts
```diff
@@ -0,0 +1,206 @@
...
+    } else if (useCase === "platform_actions") {
+      return !!(
+        extraConfig.client_id &&
+        extraConfig.client_secret &&
+        extraConfig.servicenow_instance_url
```
aubin-tchoi: no validation at all on the instance URL?

front/lib/api/oauth/providers/servicenow.ts
```diff
@@ -0,0 +1,213 @@
...
+    }: {
+      extraConfig: ExtraConfigType;
+      useCase: OAuthUseCase;
+    }
+  ): Promise<ExtraConfigType> {
+    if (useCase === "personal_actions") {
+      // For personal actions we reuse the existing connection credential id from the existing
+      // workspace connection (setup by admin) if we have it, otherwise we fallback to assuming
+      // we have client_secret (initial admin setup).
+      const { mcp_server_id, ...restConfig } = extraConfig;
```
aubin-tchoi: we don't strip the `client_secret` here?

aubin-tchoi (approved):
LGTM

thomasvicaire:
Fixed in 67073d9f07 — good catch. Moved the `client_secret` destructuring to the top of `getUpdatedExtraConfig`, before the `useCase === "personal_actions"` branch, so it's stripped unconditionally regardless of which path executes (previously only the fallback path stripped it, not the `mcp_server_id` branch).

Verified with `tsgo --noEmit` and an empirical check that `client_secret` is absent from the returned config on both paths.

Rebased #29135 and #29152 on top so the fix propagates through the stack.

PR #29136, opened 2026-07-20
Self-improving skill: fix permission issue to create skill suggestion

fabiencelier:
## Description

Seems to be broken since https://github.com/dust-tt/dust/pull/28649
The auth needs skill editor groups to be allowed to create skill suggestion, because calling `SkillSuggestionResource.createSuggestionForSkill` requires skill.canWrite

## Tests

Added a unit test

## Risk

low

## Deploy Plan

deploy front


aubin-tchoi (approved):
LGTM, sorry about that 🙏🏻

PR #29138, opened 2026-07-20
[sdks/js] fix: retry fetch once when the connection closes before any response bytes

id13:
## Description

front-api closes idle keep-alive connections after 5s (`front-api/server.ts`), that being said we encounter cases on connector where we expect the connection to be still there past that, and we have an uncaught exception bubble all the way to top.

This PR aims at retrying a fetch in case of a socket hang up on the other side at the dust-api client level

## Risk

Low - at most one extra request, only on pre-response connection death. No-op in browsers.

## Deploy Plan

- [ ] deploy `connectors`
- [ ] dispatch `publish-sdk-client.yml` (`@dust-tt/client@1.2.8`)


tdraier (approved):
LGTM

PR #29139, opened 2026-07-20
perf: throttle sandbox runtime refreshes

flvndvd:
## Description

Warm sandbox executions currently refresh the GCS credential and run the egress health check before every command. Representative traces show ~ 370 ms for GCS and ~ 418 ms for egress, adding hundreds of ms even when both were refreshed moments earlier.

This change stores the last successful runtime refresh timestamp in the database. Warm executions refreshed within the previous five minutes skip both operations and avoid any additional sandbox roundtrip.

Missing or expired timestamps trigger the existing GCS refresh followed by the egress health check. The
timestamp advances only after both operations succeed. If either fails, execution remains blocked and the timestamp is unchanged, ensuring subsequent commands retry instead of treating a partial refresh as valid.

Cold starts and sandbox wakes always bypass the guard. Wakes and provider recreation invalidate any previous timestamp so a recent value from the former runtime cannot suppress required initialization or retries.

Concurrent requests at the five-minute boundary may perform duplicate refreshes, but this is safe and avoids introducing another distributed lock or local cache.

The new database column is nullable, making the migration safe for existing rows. Existing sandboxes perform one refresh on their first execution after deployment.

Moved the tests to use a persisted database-backed sandbox resources.

## Deploy Plan

- [ ] Apply SQL migration
- [ ] Deploy front

### DIFF
--- a/front/lib/api/sandbox/lifecycle.test.ts
+++ b/front/lib/api/sandbox/lifecycle.test.ts
@@ -70,11 +70,23 @@ vi.mock("@app/lib/api/sandbox/telemetry", () => ({
   startTelemetry: mockStartTelemetry,
 }));
 
-vi.mock("@app/lib/resources/conversation_sandbox_adapter", () => ({
-  ConversationSandboxAdapter: {
-    ensureSandboxActive: mockEnsureSandboxActive,
-  },
-}));
+vi.mock(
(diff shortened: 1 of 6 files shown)


id13 (approved):
LGTM

PR #29141, opened 2026-07-20
Self-improving skills: Delete the batch in a separate activity

fabiencelier:
## Description

Run batch data deletion in a separate activity so the activity processing the batch is idempotent and can be retried after failure.
The new activity always run to make sure we delete the data even if the previous activities fail.
We don't retry the deletion if the batch is not found 

## Tests

Tested locally:
The new delete activity happens after the processing activity
image

## Risk

low
The risk is to not correctly delete the user data after usage, but the try/catch activity mecanism make is actaully safer than before

## Deploy Plan

deploy front

```
# Kill the ensure-schedules cron pinned to the old v1 queue, and restart it on v2.
  npx tsx temporal/reinforcement/admin/cli.ts stop-ensure
  npx tsx temporal/reinforcement/admin/cli.ts start-ensure
```

### DIFF
--- a/front/temporal/reinforcement/activities.ts
+++ b/front/temporal/reinforcement/activities.ts
@@ -1062,6 +1062,66 @@ export async function checkBatchStatusActivity({
   return llm.getBatchStatus(batchId);
 }
 
+function isBatchNotFoundError(err: unknown): boolean {
+  return (
+    typeof err === "object" &&
+    err !== null &&
+    "status" in err &&
+    err.status === 404
+  );
+}
+
+/**
+ * Delete a batch's data on the provider once its results have been consumed.
+ * Idempotent: a batch that no longer exists is treated as already deleted, so
+ * the activity can be safely retried.
+ */
+export async function deleteBatchActivity({
+  workspaceId,
+  batchId,
+}: {
+  workspaceId: string;
+  batchId: string;
+}): Promise<void> {
+  const auth = await getAuthForWorkspace(workspaceId);
+
+  const llm = await getReinforcedSkillsLLM(
+    auth,
+    "reinforcement_analyze_conversation"
+  );
+  if (!llm) {
+    throw ApplicationFailure.nonRetryable(
+      "ReinforcedSkills: no LLM available for batch deletion"
+    );
+  }
+
+  try {
+    const deleted = await llm.deleteBatch(batchId);
+    if (!deleted) {
+      const metadata = llm.getMetadata();
+      logger.warn(
+        {
+          workspaceId,
+          providerId: metadata.clientId,
+          modelId: metadata.modelId,
+          batchId,
+        },
+        "ReinforcedSkills: Failed to delete batch"
+      );
+    }
+  } catch (err) {
+    // The batch may already have been deleted by a previous attempt of this activity.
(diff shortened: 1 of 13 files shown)


front/temporal/reinforcement/workflows.ts
```diff
@@ -252,14 +262,25 @@ async function aggregateSkillWithMultiStepBatch({
...
+        batchId: batchResult.batchId,
+        reinforcementConversationIds: batchResult.reinforcementConversationIds,
+      });
+    } finally {
+      // Always delete the batch data once consumed, even if processing failed.
+      await deleteBatchActivity({
+        workspaceId,
+        batchId: batchResult.batchId,
+      });
+    }
```
id13, line 283: NDE guaranteed here

You need to keep both codepaths, with a temporalio `patch(blabla)` if/else to segregate those 


https://app.notion.com/p/dust-tt/Runbook-Temporal-e065e5ccd53443c1a57118d08c7a19f9?source=copy_link#16128599d94180de9d6de8305ee8468a

front/temporal/reinforcement/workflows.ts
```diff
@@ -375,20 +396,29 @@ export async function reinforcementWorkspaceWorkflow({
...
+        // Always delete the batch data once consumed, even if processing failed.
+        await deleteBatchActivity({
           workspaceId,
           batchId: batchResult.batchId,
-          reinforcementConversationMap,
-          conversationSkillMap: Object.fromEntries(
-            pendingConversations.map((c) => [c.conversationId, c.skillIds])
-          ),
         });
+      }
```
id13, line 421: same remark

front/temporal/reinforcement/activities.ts
```diff
@@ -1062,6 +1062,66 @@ export async function checkBatchStatusActivity({
...
+    if (isBatchNotFoundError(err)) {
+      logger.info(
+        { workspaceId, batchId },
+        "ReinforcedSkills: batch to delete not found, it may already be deleted"
+      );
+      return;
+    }
+    throw err;
+  }
+}
```
id13: NIT: can we avoid this big try/catch and do a result pattern at the llm.delete batch level ? 🙏

front/temporal/reinforcement/workflows.ts
fabiencelier: There is no workflow running during the day, they are triggered at night so we should be safe no ?
https://cloud.temporal.io/namespaces/dust-front-prod.gmnlm/workflows?query=%60WorkflowType%60%3D%22reinforcementWorkspaceWorkflow%22

fabiencelier: I will bump the queue to V2 so it is clear that new workflows will be on V2 and different

front/temporal/reinforcement/activities.ts
fabiencelier: done 
It made the PR bigger but changes are mostly trivial and it's cleaner now IMHO

front/temporal/reinforcement/workflows.ts
id13: Ok then no need for patch indeed 👍

id13 (approved):
LGTM

PR #29143, opened 2026-07-20
[front] nit: use typeguard on text

id13:
## Description

Use typeguard to infer text 

## Tests

N/A

## Risk

Low

## Deploy Plan

Deploy front

PR #29144, opened 2026-07-20
[front] enh: Offload oversized semantic search chunks to file system

id13:
## Description

Offload semantic search tool result to a .tool-outputs after a certain threshold.
We are currently inlining a lot of outputs either as b64 blob or text resources over mcp, which has a lot of memory footprint amplification, and can bust conversation context.

So instead this stack aims at introducing a JIT file offload for the biggest offenders according to [logs](https://app.datadoghq.eu/logs?query=inline&agg_m=count&agg_m_source=base&agg_q=%40serverName%2C%40toolName&agg_t=count&analyticsOptions=%5B%22bars%22%2C%22dog_classic%22%2Cnull%2Cnull%2C%22value%22%5D&clustering_pattern_field_path=message&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&storage=hot&stream_sort=desc&viz=toplist&from_ts=1784554496351&to_ts=1784555396351&live=true)

## Tests

Unit tests

## Risk

Low

## Deploy Plan

Deploy front

### DIFF
--- a/front/lib/api/actions/servers/search/tools/index.ts
+++ b/front/lib/api/actions/servers/search/tools/index.ts
@@ -8,6 +8,7 @@ import {
   applyNodeIdsFilterToCoreSearchArgs,
   getCoreSearchArgs,
 } from "@app/lib/actions/mcp_internal_actions/tools/utils";
+import { offloadLargeSearchResultChunks } from "@app/lib/actions/mcp_internal_actions/utils/search_result_offload";
 import {
   isAgentLoopRunContext,
   type ToolContext,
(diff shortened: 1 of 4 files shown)


id13:

- #29143
    - #29144 ◀
        - #29145
            - #29146
                - #29147

front/lib/api/actions/servers/search/tools/index.ts
```diff
@@ -186,8 +187,17 @@ export async function searchFunction(
...
+    ? await offloadLargeSearchResultChunks(
+        auth,
+        toolContext.runContext,
+        results
+      )
+    : results;
+
   return new Ok(
-    results.map((result) => ({
+    offloadedResults.map((result) => ({
```
flvndvd, line 200: This one should never return base64, does it?

id13: No, but why do you think it is ?

id13: The goal is to avoid inlining big stuff, either text or b64

PR #29150, opened 2026-07-20
[connectors] fix: bound memory in Microsoft Excel sheet ingestion

id13:
## Context

`connectors-worker-microsoft` hit 11 Node heap OOMs (`FATAL ERROR: Reached heap limit`) between Jul 7 and Jul 20 — including three crash-restart-crash cycles on one pod resuming the same sync. Strongest lead: an oversized Excel workbook on `conn_zKywRhQMss`.

The Excel path fetched each worksheet's entire `usedRange` (`?$select=text`) in a single Graph call, materializing the full 2D cell array in the heap **before** the only guard (50k rows) ran. A sheet with few rows but many columns — or a sparse sheet with a huge used range — OOMs before the guard can reject it. Worse, the too-many-rows rejection returned a plain `Err` without persisting a skip, and the stale-sheet deletion pass wiped freshly-written `skipReason` rows for previously-synced sheets, so pathological workbooks were re-fetched on every sync.

## Changes

- **Probe first**: new `getWorksheetUsedRangeMetadata` fetches `usedRange(valuesOnly=true)` scalar dimensions only (`rowCount`, `columnCount`, `cellCount`, origin) — no cell values.
- **Pre-fetch guards**: 50k-row cap (unchanged, now enforced before any fetch) plus a new 3M-cell cap. Oversized sheets are persisted via `markInternalIdAsSkipped` (`too_many_rows` / `too_many_cells`) and their stale table is deleted, so they're skipped cheaply on subsequent syncs via the existing `knownSkippedWorksheetIds` path.
- **Chunked reads**: new `getWorksheetRangeText` fetches bounded A1 windows (`range(address='B3:AO2502')`), ~100k cells per request, strictly sequential per workbook (per Microsoft guidance), with a heartbeat per chunk.
- **Incremental CSV assembly**: chunks are stringified and accumulated as CSV parts (byte-identical output to the old single-shot `stringify`); accumulation aborts at 50MB (aligned with `MAX_CSV_SIZE`) with a persisted `csv_too_large` skip. The full cell matrix is never resident.
- **Skip rows now survive the sync that writes them**: `processSheet` returns `Ok` after marking a sheet skipped (including the existing 504 path), so the stale-sheet `batchDelete` pass no longer destroys the skip marker the same sync.
- **Throttling**: `MicrosoftThrottlingError` from probe/chunk reads is rethrown instead of swallowed, letting `MicrosoftCastKnownErrorsInterceptor` honor `Retry-After` via `nextRetryDelay`.

Worst-case heap per in-flight spreadsheet drops from unbounded to ~50MB CSV + one ~1-2MB chunk.

## Tests

New `spreadsheets.test.ts` (14 tests): A1 address math incl. non-A1 origins and 16384-column sheets; probe-time skips with persisted reasons; 3-chunk happy path asserting exact window addresses and byte-identical CSV; byte-cap abort mid-loop; empty sheets; 504 on probe and mid-chunk; throttling rethrow; `handleSpreadSheet` honoring persisted skips without re-fetching.

## Risk

Sheets wider than ~60 columns at 50k rows (>3M cells) stop syncing and get a visible `skipReason` — previously they either OOMed the worker or synced by luck. Skips are permanent until cleared (same trade-off as existing `error_fetching_content` skips).

### DIFF
--- a/connectors/src/connectors/microsoft/lib/graph_api.ts
+++ b/connectors/src/connectors/microsoft/lib/graph_api.ts
@@ -367,23 +367,61 @@ export async function getWorksheets(
   return { results: res.value };
 }
 
-export async function getWorksheetContent(
+// Scalar metadata of a worksheet's used range, without any cell content.
+export type WorkbookRangeMetadata = Pick<
+  WorkbookRange,
+  | "address"
+  | "cellCount"
+  | "columnCount"
+  | "columnIndex"
+  | "rowCount"
+  | "rowIndex"
+>;
+
+export async function getWorksheetUsedRangeMetadata(
   logger: LoggerInterface,
   client: Client,
   internalId: string
+): Promise<WorkbookRangeMetadata> {
+  const { nodeType, itemAPIPath: itemApiPath } =
+    typeAndPathFromInternalId(internalId);
+
+  if (nodeType !== "worksheet") {
+    throw new Error(
+      `Invalid node type: ${nodeType} for getWorksheetUsedRangeMetadata, expected worksheet`
+    );
+  }
+  // valuesOnly=true ignores formatting-only cells so the boundary reflects
+  // actual values. Selecting only scalar properties keeps the heavy 2D arrays
+  // (text, values, formulas) out of the response entirely.
+  const res = await clientApiGet(
+    logger,
+    client,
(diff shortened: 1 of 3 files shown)


id13:
AI slope for now, please ignore

needs self-review + cleanup

PR #29135, opened 2026-07-20
feat: scaffold ServiceNow MCP server skeleton

thomasvicaire:
## Description

Second of a 2-PR stack for the ServiceNow MCP integration. **Depends on #29134** (ServiceNow OAuth provider) — base branch is `thomasvicaire/servicenow-oauth`, not `main`. Diff will shrink to just this PR's files once #29134 merges and this gets rebased onto main.

Adds the initial internal MCP server for ServiceNow: a single `list_incidents` tool against the Table API, plus the full registration plumbing:
- Feature flag `servicenow_tool` (off by default, `isPreview: true`) gating the server
- Server catalog entry (`INTERNAL_MCP_SERVERS`, id `1042`) and dispatcher wiring
- BM25 tool-search coverage
- Availability snapshot update
- Marketing integrations entry (categorized under "Support", alongside Zendesk/Freshservice)

This is intentionally scoped to the skeleton (one read-only tool) — more tools (`get_incident`, `create_incident`, `update_incident`) are in progress as a follow-up stacked PR, matching how Jira (#14414 → #14433 "basic tooling for jira") and Databricks (#18785) were built out incrementally.

## Tests

- `npx tsgo --noEmit`: clean
- `npm run test -- bm25_tool_search.test.ts`: all passing, including new ServiceNow query cases
- `npm run test -- constants.test.ts` (availability snapshot): passing
- Ran an 8-angle automated code review against the diff before the first commit; fixed the two real issues found (a missing snapshot entry, and `renderIncident` using `??` instead of a falsy check — ServiceNow returns `""` not `null` for unset fields)

## Risk

Low. Additive only, gated behind the `servicenow_tool` feature flag (off by default) — no existing code path or user-visible behavior changes.

## Deploy Plan

Standard deploy, no special steps. Icon is a temporary placeholder (`ActionCloudArrowLeftRightIcon`) until a real ServiceNow logo is available from design — marketing needs a redeploy once that lands.

PR #29152, opened 2026-07-20
feat: add get_incident, create_incident, update_incident tools

thomasvicaire:
## Description

Third of the ServiceNow MCP stack. **Depends on #29135** (MCP skeleton) — base branch is `thomasvicaire/servicenow-mcp-skeleton`, not `main`.

Rounds out the incident CRUD surface (`list_incidents` already existed) with:
- `get_incident`: look up a single incident by number
- `create_incident`: create a new incident
- `update_incident`: update/resolve/close an existing incident by `sys_id`

Uses `sysparm_input_display_value` alongside `sysparm_display_value` so callers can read and write choice fields (state, priority, resolution code) as human-readable labels rather than ServiceNow's internal codes.

## Tests

- `npx tsgo --noEmit`: clean
- `npm run test -- bm25_tool_search.test.ts`: all passing, including new query cases for the three tools
- Ran an automated 4-angle code review before committing; fixed two real issues found:
  - `get_incident` interpolated the incident number directly into a `sysparm_query`, letting `^`/`=` turn an exact-number lookup into an arbitrary compound query — now validated and rejected upfront with a clear error.
  - `update_incident` could fire a no-op PATCH when called with zero fields to change — now short-circuits with a message instead.
- **All three tools verified end-to-end against a live ServiceNow Personal Developer Instance** via the real TypeScript code paths: `create_incident` → `update_incident` (state: Resolved, with a valid resolution code) → confirmed via `get_incident`.

## Risk

Low. Additive only (new tools on an already-flagged, already-gated server). The two fixes above tighten input handling; no existing behavior changes.

## Deploy Plan

### DIFF
--- a/front/lib/api/actions/servers/servicenow/client.ts
+++ b/front/lib/api/actions/servers/servicenow/client.ts
@@ -18,6 +18,24 @@ export const IncidentSchema = z.object({
 });
 export type Incident = z.infer<typeof IncidentSchema>;
 
+// Fields writable via the incident CRUD tools, keyed by the ServiceNow
+// field name. Passing sysparm_input_display_value=true on write lets callers
+// use human-readable choice labels (e.g. "Resolved", "1 - Critical") instead
+// of ServiceNow's internal numeric codes.
+export type WritableIncidentFields = {
+  short_description?: string;
+  description?: string;
+  urgency?: string;
+  impact?: string;
+  priority?: string;
+  state?: string;
+  category?: string;
+  assignment_group?: string;
+  work_notes?: string;
+  close_notes?: string;
+  close_code?: string;
+};
+
 function getInstanceUrl(authInfo?: AuthInfo): string | null {
   if (!authInfo?.extra) {
     return null;
@@ -29,31 +47,27 @@ function getInstanceUrl(authInfo?: AuthInfo): string | null {
   return servicenowInstanceUrl.trim().replace(/\/$/, "");
 }
 
-async function servicenowApiCall<T extends z.ZodTypeAny>(
+async function request<T extends z.ZodTypeAny>(
   {
-    endpoint,
+    url,
     accessToken,
-    instanceUrl,
+    method,
+    body,
   }: {
-    endpoint: string;
+    url: string;
     accessToken: string;
-    instanceUrl: string;
-  },
-  schema: T,
-  options: {
-    method?: "GET" | "POST" | "PATCH";
+    method: "GET" | "POST" | "PATCH";
     body?: Record<string, unknown>;
-  } = {}
+  },
+  schema: T
 ): Promise<Result<z.infer<T>, MCPError>> {
-  const url = `${instanceUrl}${endpoint}`;
-
   const response = await untrustedFetch(url, {
-    method: options.method ?? "GET",
+    method,
     headers: {
       Authorization: `Bearer ${accessToken}`,
       "Content-Type": "application/json",
     },
-    ...(options.body && { body: JSON.stringify(options.body) }),
+    ...(body && { body: JSON.stringify(body) }),
   });
 
   if (!response.ok) {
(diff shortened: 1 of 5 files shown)


thomasvicaire:
Fixed the CI test failure on this PR: the MCP metadata snapshot test (tracking tool stakes/billing info across all internal servers) wasn't regenerated for the three new tools added here. Pushed in 64d02c10ee, rebased on top of the equivalent fix on #29135.

thomasvicaire:
Did a manual pass over this PR looking for the same category of issue Aubin caught on #29134 (description/comment not matching reality) before he gets to review it. Found one:

\`update_incident\`'s \`resolutionCode\` field description gave example values (\`'Solved (Permanently)'\`, \`'Solved (Work Around)'\`, \`'Closed/Resolved by Caller'\`) that I'd invented rather than verified — and none of them actually work. When we live-tested this tool earlier in the review, ServiceNow rejected \`"Solved (Permanently)"\` outright; the real valid values on that instance (discovered via \`sys_choice\`) were completely different (\`'Solution provided'\`, \`'Resolved by caller'\`, etc.), because \`close_code\` choices are configured per ServiceNow instance, not a fixed global enum.

Fixed in 0fa65edafe: rewrote the description to say that plainly instead of asserting specific values as universal, with a fallback hint (check \`list_incidents\` for a value the instance accepts).

Rest of the diff (helpers.ts, tools/index.ts) checked line-by-line against the same lens (comment/implementation mismatches, missing validation, copy-paste artifacts from the Databricks template) — nothing else found.

front/lib/api/actions/servers/servicenow/metadata.ts
```diff
@@ -26,6 +52,83 @@ export const SERVICENOW_TOOLS_METADATA = [
...
+    },
+    toolCostCategory: "advanced",
+    freeUsage: false,
+  },
+  {
+    name: "update_incident",
+    description:
+      "Update, resolve, or close an existing ServiceNow incident (ticket) identified by its sys_id (returned by list_incidents/get_incident).",
+    schema: {
+      sysId: z.string().describe("The sys_id of the incident to update."),
```
aubin-tchoi: this tool will be very complex to use for the agent and has way too many parameters
adding only this one tool is probably around 1k token by itself for the tool specification

front/lib/api/actions/servers/servicenow/client.ts
```diff
@@ -111,6 +120,107 @@ export async function listIncidents(
   return new Ok(result.value.result);
 }
 
+export async function getIncidentByNumber(
+  accessToken: string,
+  instanceUrl: string,
+  incidentNumber: string
```
aubin-tchoi: let's pass those within a structured object

front/lib/api/actions/servers/servicenow/client.ts
```diff
@@ -111,6 +120,107 @@ export async function listIncidents(
   return new Ok(result.value.result);
 }
 
+export async function getIncidentByNumber(
```
aubin-tchoi: why not structure these in a class that would store the access token and instance URL?

front/lib/api/actions/servers/servicenow/metadata.ts
aubin-tchoi: I think the agent has no way of finding this `sys_id`, it's not exposed through `renderIncident` by `getIncident`

front/lib/api/actions/servers/servicenow/metadata.ts
```diff
@@ -26,6 +52,83 @@ export const SERVICENOW_TOOLS_METADATA = [
...
+      running: "Creating ServiceNow incident",
+      done: "Create ServiceNow incident",
+    },
+    toolCostCategory: "advanced",
+    freeUsage: false,
+  },
+  {
+    name: "update_incident",
+    description:
+      "Update, resolve, or close an existing ServiceNow incident (ticket) identified by its sys_id (returned by list_incidents/get_incident).",
```
aubin-tchoi: so here we force the agent to do 2 tool calls if it wants to update an incident, one to get the `sys_id` from the incident number, and one to do the update. Why not support passing the incident number here?

front/lib/api/actions/servers/servicenow/client.ts
```diff
@@ -39,24 +39,33 @@ async function servicenowApiCall<T extends z.ZodTypeAny>(
     accessToken: string;
     instanceUrl: string;
   },
-  schema: T
+  schema: T,
+  options: {
+    method?: "GET" | "POST" | "PATCH";
```
aubin-tchoi: why not separate the get from the post to properly type the relation between the method and the body?

front/lib/api/actions/servers/servicenow/client.ts
```diff
@@ -111,6 +120,107 @@ export async function listIncidents(
...
+  instanceUrl: string,
+  incidentNumber: string
+): Promise<Result<Incident | null, MCPError>> {
+  // ServiceNow's encoded query syntax treats "^" as a condition separator (and
+  // "=" as an operator), so passing it through unescaped would let a caller
+  // turn this exact-number lookup into an arbitrary compound query.
+  if (/[\^=]/.test(incidentNumber)) {
+    return new Err(
+      new MCPError(
+        `Invalid incident number: "${incidentNumber}". Expected a plain incident number, e.g. "INC0010001".`
```
aubin-tchoi: this should not be tracked

front/lib/api/actions/servers/servicenow/metadata.ts
thomasvicaire: trimmed the schema

front/lib/api/actions/servers/servicenow/client.ts
thomasvicaire: sure

front/lib/api/actions/servers/servicenow/tools/index.ts
```diff
@@ -38,6 +79,91 @@ const handlers: ToolHandlers<typeof SERVICENOW_TOOLS_METADATA> = {
...
+    const client = clientResult.value;
+
+    const result = await client.getIncidentByNumber(incidentNumber);
+
+    if (result.isErr()) {
+      return result;
+    }
+
+    if (!result.value) {
+      return new Ok([
```
aubin-tchoi: should be exposed with `isError` to true

front/lib/api/actions/servers/servicenow/client.ts
```diff
@@ -81,39 +104,178 @@ async function servicenowApiCall<T extends z.ZodTypeAny>(
...
+};
 
-  const result = await servicenowApiCall(
-    {
-      endpoint: `/api/now/table/incident?${params.toString()}`,
-      accessToken,
-      instanceUrl,
-    },
-    z.object({ result: z.array(IncidentSchema) })
+function get<T extends z.ZodTypeAny>(
```
aubin-tchoi: this should be a private method of the client class

front/lib/api/actions/servers/servicenow/client.ts
```diff
@@ -81,39 +104,178 @@ async function servicenowApiCall<T extends z.ZodTypeAny>(
...
+  { endpoint, accessToken, instanceUrl }: BaseRequestArgs,
+  schema: T
+): Promise<Result<z.infer<T>, MCPError>> {
+  return request(
+    { url: `${instanceUrl}${endpoint}`, accessToken, method: "GET" },
+    schema
+  );
+}
+
+function write<T extends z.ZodTypeAny>(
```
aubin-tchoi: same for this one

aubin-tchoi: nit: `write` is slightly awkward

front/lib/api/actions/servers/servicenow/client.ts
```diff
@@ -81,39 +104,178 @@ async function servicenowApiCall<T extends z.ZodTypeAny>(
...
-  // like state/priority instead of ServiceNow's raw internal codes.
-  params.set("sysparm_display_value", "true");
-  if (query) {
-    params.set("sysparm_query", query);
-  }
+type BaseRequestArgs = {
+  endpoint: string;
+  accessToken: string;
+  instanceUrl: string;
+};
```
aubin-tchoi: you shouldn't need this type since the access token and instance URL are stored by the client

front/lib/api/actions/servers/servicenow/client.ts
```diff
@@ -81,39 +104,178 @@ async function servicenowApiCall<T extends z.ZodTypeAny>(
...
+  }
+
+  async updateIncident({
+    incidentNumber,
+    fields,
+  }: {
+    incidentNumber: string;
+    fields: WritableIncidentFields;
+  }): Promise<Result<Incident, MCPError>> {
+    const existingResult = await this.getIncidentByNumber(incidentNumber);
```
aubin-tchoi: the client should directly map to endpoints, not contain logic on how to chain API calls (should be in the tool instead)

aubin-tchoi (commented):
LGTM, left a few non-critical comments
can you share how many tokens does the tool specifications take + the resulting JSON schemas?

front/lib/api/actions/servers/servicenow/client.ts
```diff
@@ -81,39 +104,142 @@ async function servicenowApiCall<T extends z.ZodTypeAny>(
...
 
-  const result = await servicenowApiCall(
-    {
-      endpoint: `/api/now/table/incident?${params.toString()}`,
-      accessToken,
-      instanceUrl,
-    },
-    z.object({ result: z.array(IncidentSchema) })
-  );
+  private mutate<T extends z.ZodTypeAny>(
```
aubin-tchoi, line 127: nit: `mutate` looks a bit off + introduces a new word. why not 1 method per verb?

aubin-tchoi (approved):
LGTM!

PR #29154, opened 2026-07-20
Ship Computer to all workspaces

fontanierh:
## Description

Remove the `disable_computer_feature` flag and its gated code paths. Computer APIs, admin settings, uploads, tools, prompts, and skills are now available to every workspace.

Upload limits and raw-file processing are now derived directly from the file use case. Prompt construction uses `hasComputerTool` only where the generated agent may genuinely lack access to Computer.

## Tests

Tested locally with 140 focused Vitest cases across the affected Front and front-api paths. Regenerated and validated the OpenAPI document.

## Risk

Low. This intentionally removes the workspace opt-out; legacy flag rows are ignored once the flag leaves the whitelist. The sandbox image artifact is unchanged; edits under its module only remove runtime manifest filtering.

## Deploy Plan

Standard deploy.


front/lib/api/assistant/generation.ts
```diff
@@ -283,11 +283,11 @@ function constructAttachmentsSection({
 }
 
 function constructAttachmentsSectionNewFileExplorer({
-  hasSandboxTools,
+  hasComputerTool,
```
fontanierh, line 286: Will be cleaning this up next (but I want this PR to be mostly no-op)

flvndvd (approved):
FUEGO 🔥 🔥 🔥

PR #29155, opened 2026-07-20
[LLM] Attribute Fireworks-hosted models to their real labs; fix legacy transition provider keying

pmilliotte:
## Description

Reworks the new LLM router's endpoint identity to a `lab/model/region/host` scheme where `lab` is the model's actual developer (so Fireworks-hosted models are attributed to `deepseek`/`z_ai`/`moonshot_ai` and endpoints are filtered by both lab and host), and fixes the legacy transition layer to key each model's legacy config on its serving provider rather than its lab.

## Tests

Tested locally by comparing the payloads sent through the legacy LLM path vs. the new router and confirming parity. Router parity and per-endpoint test suites updated accordingly.

## Risk

Low.

## Deploy Plan

- Deploy `main` on `front`


front/lib/api/llm/index.ts
```diff
@@ -345,6 +347,17 @@ const REGION_MAPPING: Record<RegionType, Region> = {
...
+// Maps a legacy `ModelIdType` (Fireworks ids still carry the
+// `accounts/fireworks/models/` prefix) to the bare `Model` id the router keys
+// endpoints by. Returns null for unknown ids.
+export function legacyModelIdToModel(modelId: string): Model | null {
+  const bare = modelId.startsWith(FIREWORKS_MODEL_PREFIX)
+    ? modelId.slice(FIREWORKS_MODEL_PREFIX.length)
+    : modelId;
+
+  return isModel(bare) ? bare : null;
+}
```
pmilliotte, line 410: Removed the `accounts/fireworks/models/` fireworks prefix to identify models in the new router.

Why ? For example if we want to serve GLM5.2 through another host, we want that `model = glm-5.2`, without fireworks prefix

front-api/routes/marketing/academy/chat.ts
```diff
@@ -1,9 +1,10 @@
 /** @ignoreswagger */
 import Anthropic from "@anthropic-ai/sdk";
 import config from "@app/lib/api/config";
-import { CLAUDE_SONNET_4_6_MODEL_ID } from "@app/lib/model_constructors/types/model_ids";
+
 import { rateLimiter } from "@app/lib/utils/rate_limiter";
 import logger from "@app/logger/logger";
+import { CLAUDE_SONNET_4_6_MODEL_ID } from "@app/types/assistant/models/anthropic";
```
pmilliotte, line 7: Should have been imported from there since the beginning, cannot happen again as now we import CLAUDE_SONNET_4_6 as a Model (not ModelId) when needed in the new router

front/lib/api/llm/index.ts
```diff
@@ -315,24 +322,71 @@ function getRegionFilter(auth: Authenticator): ValueFilter<Region> | undefined {
...
+  );
+  const labFilter: Where<EndpointConfig> = {
+    lab: { in: whitelistedLabs },
+  };
+  const whitelistedHosts = compact(
+    providerIds.map((id) => PROVIDER_ID_TO_HOST[id])
+  );
+  const hostFilter: Where<EndpointConfig> = {
+    host: { in: whitelistedHosts },
+  };
```
pmilliotte, line 378: (soon to be) legacy ModelProviderId is confusing: it includes fireworks which is not a lab.
In the new router,  we distinguish labs from hosts to later be able to filter on labs (user) or hosts (admin)

This filter enables same behavior when retrieving models based on workspace's whitelistedProviders

front/lib/api/llm/transitionLLM.ts
```diff
@@ -579,6 +584,18 @@ function convertBatchEventsToOld(
...
 
+// The endpoint's `lab` is the model's developer ("deepseek" for a
+// fireworks-hosted DeepSeek model), not the legacy `ModelProviderIdType` the LLM
+// base keys its config on — that's the serving provider ("fireworks"). Resolve
+// it from the model config, which is unique by modelId.
+function legacyProviderIdForModel(modelId: ModelIdType): ModelProviderIdType {
+  const modelConfig = getModelConfigByModelId(modelId);
+  if (!modelConfig) {
+    throw new Error(`Model config not found for ${modelId}`);
+  }
```
pmilliotte, line 638: Never happens, will be removed after cleaning phase.

New endpoints will gather all the necessary config, and once we have a new endpointId, we are sure to get the config associated to it

front/lib/api/llm/transitionLLM.ts
```diff
@@ -111,6 +112,45 @@ function mapReasoningEffort(
   }
 }
 
+// The persisted passthrough `provider` uses the legacy provider-id vocabulary
+// (e.g. `google_ai_studio`), while BaseMessage carries the model's `Lab`. Only
+// `google_ai_studio` diverges (-> `google`); the rest are identical.
+function passthroughProviderToLab(
```
pmilliotte, line 118: passthrough is an anthropic feature only.

Later there might have other labs supporting it, but it's not a "host-related" feature so removing non compatible host vs labs for now

PR #29157, opened 2026-07-20
[core] chore: upgrade qdrant to 1.18

id13:
## Description

Upgrade qdrant-client to 1.18
Previous attempts were failing at run step due to a missing peer dep version in libc. (see comment in docker img)

## Tests

Tested locally

## Risk

Low

## Deploy Plan

Deploy core

### DIFF
--- a/dockerfiles/core.Dockerfile
+++ b/dockerfiles/core.Dockerfile
@@ -1,5 +1,7 @@
 # Build stage — full Rust toolchain + cmake (needed for sentencepiece)
-FROM rust:1.94.1 AS builder
+# Keep the builder on bookworm: the runtime stage is bookworm-slim and binaries must not link
+# against a newer glibc than the runtime provides.
+FROM rust:1.94.1-bookworm AS builder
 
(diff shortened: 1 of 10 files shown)


dockerfiles/core.Dockerfile
```diff
@@ -1,5 +1,7 @@
 # Build stage — full Rust toolchain + cmake (needed for sentencepiece)
-FROM rust:1.94.1 AS builder
+# Keep the builder on bookworm: the runtime stage is bookworm-slim and binaries must not link
+# against a newer glibc than the runtime provides.
```
id13, line 3: 🤦

dockerfiles/core.Dockerfile
id13, line 17: 🤦 🤦

dockerfiles/core.Dockerfile
id13: Reading the dockerfile from rust, the 1.94 is from trixie, and we are then copying things to a bookworm

flvndvd (approved):

PR #29159, opened 2026-07-20
Add checkpointed no-drop conversation window

flvndvd:
## Description

Following all our recent attempts at improving our pruning, this new approach considers a way simpler way.
 
This new strategy exclusively prune eligible (already consumed / not current assistant turn) tool results. If despite pruning it does not fit a follow up PR will add auto compaction as a natural escalation. We are not dropping anymore, since we noticed that it might give some headroom but we eventually end up pruning/dropping on every subsequent turns which lead to bursting the cache too often. 

This PR introduces a feature-flagged conversation-window strategy that preserves every interaction and limits
context reduction to consumed tool results. The existing strategy remains the default rollout and rollback path.

Tool results are represented by lightweight internal nodes, with pending and eligible queues referencing the
same payload. A complete result batch remains protected until consumed by the model. Pruning begins at 60%
utilization and advances only in meaningful token checkpoints, creating runway while reducing repeated prompt-cache invalidations.

When pruning cannot recover enough space, the complete history continues beyond the conservative nominal budget instead of failing or silently removing interactions. Existing UI thresholds handle compaction between agent runs, while strategy-specific saturation and budget-overrun metrics provide rollout visibility.

### DIFF
--- a/front/lib/api/assistant/conversation_rendering/checkpointed_window_state.test.ts
+++ b/front/lib/api/assistant/conversation_rendering/checkpointed_window_state.test.ts
@@ -1,0 +1,231 @@
+import { CheckpointedConversationWindowState } from "@app/lib/api/assistant/conversation_rendering/checkpointed_window_state";
+import type { InteractionWithTokens } from "@app/lib/api/assistant/conversation_rendering/pruning";
+import type { ModelMessageTypeMultiActions } from "@app/types/assistant/generation";
+import { describe, expect, it } from "vitest";
+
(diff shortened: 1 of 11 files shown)


fontanierh (approved):
So if this is enabled, and we go over, we get errors for now ?

flvndvd:
> So if this is enabled, and we go over, we get errors for now ?

Only logs, the compaction that prevents user from posting message will constraint you to compact.

aubin-tchoi (approved):
LGTM

PR #29160, opened 2026-07-20
Clarify conversation-scoped skill caching

fontanierh:
## Description

Explain how code-defined system skill instructions depend on effective Space IDs, using Discover Tools as the concrete example, and why conversation-expanded scope keeps those instructions out of the long-cache tier.

## Tests

Not applicable, comment-only change.

## Risk

None, comment-only change.

## Deploy Plan

Standard deploy.


Fraggle (commented):
THanks !

PR #29161, opened 2026-07-20
Download binary files instead of previewing them

fontanierh:
## Description

Treat unknown and opaque MIME types as download-only instead of decoding their bytes as text. Inline citations and explorer entries now download these files directly, and preview navigation skips them.

## Tests

Added focused Vitest coverage for ZIP and octet-stream classification, citation downloads, explorer downloads, and navigation.

## Risk

Low. The preview policy uses a text and renderer allowlist; unsupported types fall back to the existing download paths.

## Deploy Plan

Standard deploy.

### DIFF
--- a/front/components/file_explorer/utils.ts
+++ b/front/components/file_explorer/utils.ts
@@ -6,21 +6,32 @@ import {
   isInteractiveContentType,
   isMarkdownContentType,
   isPdfContentType,
+  isSandboxFunctionContentType,
+  stripMimeParameters,
 } from "@app/types/files";
 
-const VIEWER_CONTENT_TYPES = [
+const VIEWER_CONTENT_TYPES = new Set<string>([
   "application/msword",
   "application/vnd.openxmlformats-officedocument.wordprocessingml.document",
   "application/vnd.ms-powerpoint",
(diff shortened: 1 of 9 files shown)


front/components/file_explorer/utils.ts
```diff
@@ -80,7 +103,11 @@ export function getFilePreviewConfig(contentType: string): FilePreviewConfig {
...
   }
 
-  if (category === "code" || category === "data") {
+  if (
+    category === "code" ||
+    isSandboxFunctionContentType(contentType) ||
+    isTextPreviewContentType(contentType)
+  ) {
     return {
       category: "code",
```
flvndvd, line 112: So we map all plain text supported file to code down the line? Feels a bit misleading, no?

flvndvd (approved):

front/components/file_explorer/utils.ts
fontanierh: That's currently the case I believe no ?

flvndvd: Hum I thought they had a specific behavior where people could edit them alongside markdown.

zmarouf (commented):
LGTM

PR #29162, opened 2026-07-20
feat(front): keep sticky session preference for input bar model picker

Nils-Fedrigo:
## Description

The input-bar model picker (behind the `models_picker` flag) now remembers the last override the user picked for the whole browser-tab session, so it no longer has to be re-selected on every new conversation or agent switch.

The picked override is stored in `InputBarContext` and backed by `sessionStorage`, so it:
- is **global** — one override applies across all agents and conversations;
- **survives page reloads** but stays scoped to the tab (cleared when the tab/session ends — no DB, no `localStorage`);
- **defers to conversation history** — an existing conversation still shows its own last-requested model; the sticky only seeds new conversations and agent switches. Effective order: local pick this render → conversation `lastRequestedModel` → session sticky → agent default → Auto.

Reverting the picker to the agent default clears the sticky. Reads from storage are validated with the existing `ModelSelectionSchema` so a stale value can't break the picker.

## Tests

Manual, with the `models_picker` flag enabled:
- Pick a model in a conversation → start a new conversation → picker pre-selects it and sends it.
- Switch the addressed agent → override is kept (global).
- Open an existing conversation that used a different model → it shows its own last model, not the sticky.
- Revert to Default → new conversations fall back to the agent default.
- Reload the page → sticky is preserved; open/close the tab → sticky is gone.

### DIFF
--- a/front/components/assistant/conversation/input_bar/InputBarContext.tsx
+++ b/front/components/assistant/conversation/input_bar/InputBarContext.tsx
@@ -9,6 +9,7 @@ import type {
   RichMention,
 } from "@app/types/assistant/mentions";
 import type { ModelSelectionType } from "@app/types/assistant/models/types";
+import { ModelSelectionSchema } from "@app/types/assistant/models/types";
 import type { ContentFragmentsType } from "@app/types/content_fragment";
 import {
   createContext,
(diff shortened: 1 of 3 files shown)


Nils-Fedrigo:

- #29162 ◀
    - #29168

Fraggle (approved):
LGTM

PR #29165, opened 2026-07-20
Reduce dependency on metronome webhook

tdraier:
## Description

In dev/local environments Metronome webhooks are never delivered (no `METRONOME_WEBHOOK_SECRET` / no reachable endpoint). Two consequences, fixed here:

**1. `switchContract` created no Subscription row at all for immediate switches.**
Subscription creation/swap for an immediate contract switch only happened in the `contract.start` webhook handler (`process_webhook.ts`). Without the webhook, `switchContract` provisioned the Metronome contract but left the workspace with zero local Subscription rows — not even a pending one.

Fix: extracted the `contract.start` handling into a standalone `applyContractStartSubscriptionSwap({ workspace, contractId, customerId })` (same idempotency/shadow-contract checks, no behavior change for the real webhook path). `switchContract` now calls it synchronously right after provisioning, but **only** when no Metronome webhook secret is configured and the switch is immediate (a future-dated switch still stages a pending row and waits for the real webhook). Production/staging always have the secret set and continue to rely exclusively on the webhook.

**2. Several AWU credit/commit grants depended on the webhook to retroactively stamp `DUST_CONTRACT_CREDIT_TYPE`.**
`credit.create`/`commit.create` (and related) webhook handlers stamp a `pool`/`excess` custom field used by balance alerts and pool-state reconciliation. Without the webhook, newly granted credits/commits stayed unstamped in dev. Since the one-time `add_commits`/`add_credits` (and `v1.customers.credits.create`) Metronome APIs already accept `custom_fields` at creation time, the following call sites now stamp directly instead of waiting on the webhook:
- Legacy credit migration (converted + bonus credits) — always `pool` (never the excess product).
- Coupon credits — the AWU pool top-up coupon stamps `pool`; the seat-subscription coupon credit (fiat-denominated, never AWU) is explicitly **not** stamped.
- Poke "Grant AWU credits" plugin, free-credit path (the paid/commit path already stamped at creation).
- Renewal carry-over (commit + credit regrants) — copies the `DUST_CONTRACT_CREDIT_TYPE` value forward from the source commit/credit rather than reclassifying it, so it works regardless of pool/excess/unstamped.

Excess credits are auto-created by Metronome package templates (never explicitly granted by our code), so they're out of scope here and still rely on the webhook — not carried over on renewal either, so this doesn't affect them.

## Tests

- `npx tsgo --noEmit` clean.
- `biome check` clean on all changed files.
- Existing Vitest suites pass unchanged: `process_webhook.test.ts` (25), `coupons.test.ts` (37), `client.test.ts` (22), `renewal_carry_over.test.ts` (8).
- Manually reasoned through each webhook consumer (`lib/metronome/alerts/balance_threshold.ts`, `lib/api/credits/awu_pool_summary.ts`) to confirm the `pool` stamp is always combined with an explicit AWU `credit_type_id` filter downstream, so no behavior change for non-AWU credits.

## Risk

Low.
- The `switchContract` change is gated on the absence of a webhook secret, so prod/staging (which always have it configured) are unaffected — behavior there is unchanged, still driven exclusively by the real `contract.start` webhook.
- The stamping changes are purely additive (`custom_fields` passed at creation); the webhook's own stamping logic is unchanged and still re-checks/no-ops if already stamped, so webhook re-delivery remains safe.
- Rollback is a straight revert — no data migration involved.

## Deploy Plan

Standard deploy, no feature flag, no migration.

### DIFF
--- a/front/lib/api/metronome/process_webhook.ts
+++ b/front/lib/api/metronome/process_webhook.ts
@@ -571,11 +571,300 @@ async function handlePerUserSpendThresholdEvent({
   return new Ok(undefined);
 }
 
-// If the started contract was stamped for the legacy → Business migration,
-// convert the workspace's convertible legacy credits to AWU and grant the
-// per-user free bonus — computed now, so the amounts reflect the workspace's
-// state at activation. Shared by both `contract.start` swap paths (pending row
-// activation and immediate swap). Best-effort: failures are logged, never throw.
+// Perform the subscription-affecting side effects of a Metronome
+// `contract.start` event: carry over renewal balances, reconcile credit
+// state, and swap (or activate a pending) local Subscription row onto the
+// new contract.
+export async function applyContractStartSubscriptionSwap({
+  workspace,
+  contractId,
+  customerId,
+}: {
+  workspace: WorkspaceResource;
+  contractId: string;
+  customerId: string;
+}): Promise<Result<undefined, ProcessMetronomeWebhookError>> {
+  // Read the PLAN_CODE custom field to know which plan to swap the
+  // workspace subscription onto. The actual swap is gated below on
+  // `isMetronomeOnlyBilled` — other billing paths (shadow, pure
+  // Stripe) handle their own state transitions, and contracts whose
+  // start aligns with a synchronous DB flip get caught by the
+  // idempotency check. Fetched up-front because the carry-over below also
+  // needs the contract's transition lineage and start.
+  const contractResult = await getMetronomeContractById({
+    metronomeCustomerId: customerId,
+    metronomeContractId: contractId,
+  });
+  if (contractResult.isErr()) {
+    logger.error(
+      {
+        contractId,
+        customerId,
+        error: contractResult.error,
+        workspaceId: workspace.sId,
+      },
+      "[Metronome Webhook] contract.start: failed to fetch contract"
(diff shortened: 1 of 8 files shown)


tdraier:

- #29165 ◀

front/lib/api/metronome/process_webhook.ts
```diff
@@ -576,6 +576,304 @@ async function handlePerUserSpendThresholdEvent({
...
 // activation and immediate swap). Best-effort: failures are logged, never throw.
+// Perform the subscription-affecting side effects of a Metronome
+// `contract.start` event: carry over renewal balances, reconcile credit
+// state, and swap (or activate a pending) local Subscription row onto the
+// new contract. Extracted from the `contract.start` webhook case so
+// `switchContract` can replay it synchronously in environments where the
+// Metronome webhook is not configured (e.g. local dev, where `contract.start`
+// would otherwise never be delivered and no Subscription row — not even a
+// pending one — would ever be created).
+export async function applyContractStartSubscriptionSwap({
```
fabiencelier: unnecesary end of comment IMHO

```suggestion
// new contract.
export async function applyContractStartSubscriptionSwap({
```

fabiencelier (approved):

PR #29166, opened 2026-07-20
Fix skill import dialog UI and detection state

Inesvvv:
## Description

Three fixes: 
1. Fixed github avatar for dark mode 
2. Fixed the UI of the connect button, putting the text on one line 
3. Added `clearDetection` to reset everything when the user deletes the url 

Github avatar fix + connect button fix: 
Screenshot 2026-07-20 at 15 02 00

## Tests

Tested locally

## Risk

Low, UI fixes

## Deploy Plan

Deploy front

### DIFF
--- a/front/lib/swr/skill_configurations.ts
+++ b/front/lib/swr/skill_configurations.ts
@@ -474,6 +474,7 @@ export function useDetectSkillsFromRepo({
           if (signal.aborted) {
             return;
           }
+          setDetectedSkills([]);
           if (isAPIErrorResponse(err)) {
             setDetectError(err.error.message);
             setRepositoryNotFound(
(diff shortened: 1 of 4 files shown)


aubin-tchoi (approved):
LGTM

front/lib/swr/skill_configurations.ts
```diff
@@ -501,13 +502,22 @@ export function useDetectSkillsFromRepo({
     [triggerDetect]
   );
 
+  const clearDetection = useCallback(() => {
+    lastDetectedUrl.current = null;
+    setDetectedSkills([]);
+    setDetectError(null);
+    setRepositoryNotFound(false);
+    setIsDetecting(false);
```
aubin-tchoi, line 510: you also need to abort the in-flight request I think (IIRC it's debounced)

Inesvvv: I thought it added too much complexity to the code for something that only happens if the user pastes the url and deletes it within the 1s debounce, which I think might be a very niche edge case. If it's something that's worth the extra code I can definitely add it

aubin-tchoi: wfm

PR #29171, opened 2026-07-20
Add heartbeats for temporal activity

tdraier:
## Description

`syncMetronomeSeatCountWorkflow` got permanently stuck retrying a `syncMetronomeSeatCountActivity` `StartToCloseTimeout` with zero Datadog signal: the activity's code only logs `"Activity failed"` when the activity function itself throws, but a server-side StartToClose timeout is enforced by Temporal without ever notifying the running worker code — so it never threw, never logged, and kept running orphaned in the background while Temporal silently retried a new attempt.

Adds heartbeating (via a guarded `heartbeat()` helper in `lib/temporal.ts`, ported from connectors' `src/lib/temporal.ts` — safe to call outside a Temporal activity context, where it's a no-op) at every loop/checkpoint in `syncMetronomeSeatCountActivity`'s call chain whose duration scales with workspace size: the seat reconcile loop, free-seat grant/revoke, seat-credit transfers, bulk membership remaps, and per-member credit-state reconciliation. Paired with `heartbeatTimeout: "1 minute"` on the activity's Temporal proxy config, this lets Temporal detect and cancel a stalled attempt within ~1 minute instead of only via the coarse 10-minute `startToCloseTimeout` — and because the helper's cancellation checkpoint (`Context.current().sleep(0)`) surfaces that cancellation as a normal thrown error, it now gets caught and logged as `"Activity failed"` by the existing `ActivityInboundLogInterceptor`, which is what the Datadog "Activity Failure" monitor and the "Stuck front-worker" dashboard already key off.

## Tests

Existing unit tests in `seats.test.ts`, `seats_sync.test.ts`, and `reconcile_credit_state.test.ts` cover the touched functions. Added the missing `heartbeat` export to the global `@app/lib/temporal` mock in `vite.setup.ts` (it fully replaces the module, so the new export was coming back `undefined` and failing 7 tests).

## Risk

Low. `heartbeat()` no-ops outside an activity context, so all non-Temporal callers of this shared seat-sync code (poke plugin, membership hooks, checkout flow) are unaffected. Inside the activity, the only behavior change is that a genuinely stalled attempt now gets cancelled after ~1 minute of no progress instead of running silently for up to 10 minutes.

### DIFF
--- a/front/lib/api/metronome/reconcile_credit_state.ts
+++ b/front/lib/api/metronome/reconcile_credit_state.ts
@@ -41,6 +41,7 @@ import { MembershipResource } from "@app/lib/resources/membership_resource";
 import { UserResource } from "@app/lib/resources/user_resource";
 import type { WorkspaceResource } from "@app/lib/resources/workspace_resource";
 import { resolveEffectiveSpendLimitAwuCredits } from "@app/lib/spend_limits/effective";
+import { heartbeat } from "@app/lib/temporal";
 import { renderLightWorkspaceType } from "@app/lib/workspace";
 import logger from "@app/logger/logger";
 import type {
@@ -681,6 +682,7 @@ export async function reconcileWorkspaceUserCreditStates({
         "[ReconcileCreditState] Failed to reconcile a user's credit state"
       );
     }
+    await heartbeat();
   }
 }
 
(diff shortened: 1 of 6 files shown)


tdraier:

- #29171 ◀

PopDaph (approved):
LGTM!

PR #29175, opened 2026-07-20
Extend model image signed URL validity

flvndvd:
## Description

This PR aims at fixing https://github.com/dust-tt/tasks/issues/9236.

It extends model image signed URLs to 15 minutes while retaining the 80% cache buffer. User-facing download validity remains unchanged. This reduces provider download failures caused by near-expiry cached URLs.


aubin-tchoi (approved):
LGTM

PR #29039, opened 2026-07-16
[Skills] Remove 'globalSpaceOnly: true' to fix restricted skills bug

xebeohp:
## Description
Fix the bug with restricted skills becoming unsearchable for all after being restricted to pods/spaces. The fix involves removing the 'globalSpaceOnly: true' flag from the capabilities picker, slash command, and default skills.

## Tests
Added tests to skills/test. Tested locally.

## Deploy Plan

PR #29177, opened 2026-07-20
[Skills] Skills restriction in pods fix

xebeohp:
## Description
Restricting a skill to a Pod/space made it unsearchable everywhere ([#29039](https://github.com/dust-tt/tasks/issues/9407)). The initial fix removed the `globalSpaceOnly: true flag` from the capabilities picker, slash command, and default-skills picker but that over-corrected.

To ensure that restricting skills follows the access rule of pods (everything in a pod must be accessible to everyone in the pod), pod skill selection are now scoped to global skills + skills restricted to that specific Pod. Skills restricted to other spaces/Pods are excluded. Outside a Pod (personal conversations, skill builder, command palette) nothing changes — all accessible skills remain available.

Ex: A skill is eligible in Pod X iff every one of its requestedSpaceIds is either the global space or X's own space. Thus a restricted skill can be used in at most one pod (if you try to restrict a skill to two pods, the logic will fail but the skill can still be used in a non-pod conversation since it passes no `podSpaceId`).

## Tests
Local tests, added tests to index, project_metadata, and project_metadata_resource.

## Risk
Low risk

### DIFF
--- a/front-api/routes/w/[wId]/skills/index.ts
+++ b/front-api/routes/w/[wId]/skills/index.ts
@@ -153,37 +153,7 @@ app.get(
     }
     const skillStatus = statusValidation.data;
 
-    const availabilityValidation =
-      SkillAvailabilitiesSchema.safeParse(availabilityParams);
-    if (!availabilityValidation.success) {
-      return apiError(ctx, {
-        status_code: 400,
-        api_error: {
-          type: "invalid_request_error",
-          message: `Invalid availability: ${availabilityParams}. Expected "editors", "workspace_users", or "users_and_agents".`,
-        },
-      });
-    }
-    // An explicit availability takes priority over the deprecated isDefault alias.
-    const availability =
-      availabilityValidation.data && availabilityValidation.data.length > 0
-        ? availabilityValidation.data
-        : isDefault === "true"
-          ? ["users_and_agents" as const]
-          : undefined;
-
-    // Only admins may list unpublished skills they don't edit (e.g. for governance views).
-    if (bypassEditorVisibility && !auth.isAdmin()) {
-      return apiError(ctx, {
-        status_code: 403,
-        api_error: {
-          type: "app_auth_error",
-          message: "Only admins can bypass editor visibility.",
-        },
-      });
-    }
-
-    const allSkills = await SkillResource.listByWorkspace(auth, {
+    const skills = await SkillResource.listBySpace(auth, {
       status: skillStatus,
       globalSpaceOnly: globalSpaceOnly === "true",
       podSpaceId,
--- a/front-api/routes/w/[wId]/skills/index.ts
+++ b/front-api/routes/w/[wId]/skills/index.ts
@@ -9,6 +9,7 @@ import { DataSourceViewResource } from "@app/lib/resources/data_source_view_reso
 import { FileResource } from "@app/lib/resources/file_resource";
 import { MCPServerViewResource } from "@app/lib/resources/mcp_server_view_resource";
 import { SkillResource } from "@app/lib/resources/skill/skill_resource";
+import { getResourceIdFromSId } from "@app/lib/resources/string_ids";
 import logger from "@app/logger/logger";
 import type {
   GetSkillsResponseBody,
@@ -134,10 +135,11 @@ app.get(
     const onlyCustom = ctx.req.query("onlyCustom");
     // @deprecated Use availability instead. Kept while old clients still send it.
     const isDefault = ctx.req.query("isDefault");
-    const bypassEditorVisibility =
-      ctx.req.query("bypassEditorVisibility") === "true";
-    // Repeatable: ?availability=workspace_users&availability=users_and_agents.
-    const availabilityParams = ctx.req.queries("availability");
+    const podSpaceSId = ctx.req.query("podSpaceId");
+    const podSpaceId =
+      podSpaceSId !== undefined
+        ? (getResourceIdFromSId(podSpaceSId) ?? undefined)
+        : undefined;
 
     const statusValidation = SkillStatusSchema.safeParse(status);
     if (!statusValidation.success) {
@@ -184,6 +186,7 @@ app.get(
     const allSkills = await SkillResource.listByWorkspace(auth, {
       status: skillStatus,
       globalSpaceOnly: globalSpaceOnly === "true",
+      podSpaceId,
       onlyCustom: onlyCustom === "true",
       availability,
       withInstructions: false,
--- a/front-api/routes/w/[wId]/skills/index.ts
+++ b/front-api/routes/w/[wId]/skills/index.ts
@@ -135,10 +135,10 @@ app.get(
     const onlyCustom = ctx.req.query("onlyCustom");
     // @deprecated Use availability instead. Kept while old clients still send it.
     const isDefault = ctx.req.query("isDefault");
-    const podSpaceSId = ctx.req.query("podSpaceId");
-    const podSpaceId =
-      podSpaceSId !== undefined
-        ? (getResourceIdFromSId(podSpaceSId) ?? undefined)
+    const podSpaceId = ctx.req.query("podSpaceId");
+    const podSpaceModelId =
+      podSpaceId !== undefined
+        ? (getResourceIdFromSId(podSpaceId) ?? undefined)
         : undefined;
 
     const statusValidation = SkillStatusSchema.safeParse(status);
@@ -156,7 +156,7 @@ app.get(
     const skills = await SkillResource.listBySpace(auth, {
       status: skillStatus,
       globalSpaceOnly: globalSpaceOnly === "true",
-      podSpaceId,
+      podSpaceId: podSpaceModelId,
       onlyCustom: onlyCustom === "true",
       availability,
       withInstructions: false,
--- a/front/components/skill_builder/SkillBuilderRequestedSpacesSection.tsx
+++ b/front/components/skill_builder/SkillBuilderRequestedSpacesSection.tsx
@@ -238,8 +238,10 @@ export function SkillBuilderRequestedSpacesSection({
             Visibility control and available data
           </h3>
           <p className="text-sm text-muted-foreground">
-            Add a space or pod to restrict usage to its members and make its
-            data available to this skill.
+            Choose which spaces and Pods this skill can access. The skill can
+            use their knowledge and capabilities, and only users with access to
+            all selected spaces and Pods can use it. A skill restricted to a Pod
+            is only available inside that Pod and not in others.
           </p>
         </div>
         <Button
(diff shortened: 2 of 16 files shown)


front/lib/api/actions/servers/pod_manager/metadata.ts
```diff
@@ -24,6 +24,7 @@ export const SEMANTIC_SEARCH_TOOL_NAME = "semantic_search" as const;
 export const EDIT_INFORMATION_TOOL_NAME = "edit_information" as const;
 export const SET_PINNED_FRAME_TOOL_NAME = "set_pinned_frame" as const;
 export const MOVE_CONVERSATION_TOOL_NAME = "move_conversation" as const;
+export const DEFAULT_AGENT_TOOL_NAME = "default_agent" as const;
```
aubin-tchoi: would name it with a verb (`set_default_agent`)

front/lib/api/actions/servers/pod_manager/metadata.ts
```diff
@@ -142,6 +143,33 @@ export const POD_MANAGER_TOOLS_METADATA = createToolsRecord({
     toolCostCategory: "basic",
     freeUsage: true,
   },
+  [DEFAULT_AGENT_TOOL_NAME]: {
+    description:
+      "Set or clear the Pod default agent — the agent that handles new conversations started in this Pod when no agent is picked explicitly. " +
```
aubin-tchoi: nit: let's avoid the em dash here

front/lib/api/actions/servers/pod_manager/metadata.ts
```diff
@@ -142,6 +143,33 @@ export const POD_MANAGER_TOOLS_METADATA = createToolsRecord({
...
+      "Provide agentName to set it (matched by name, best match wins), or pass null to reset to the default (@dust).",
+    schema: {
+      agentName: z
+        .string()
+        .nullable()
+        .describe(
+          "Name of the agent to set as the Pod default. The tool searches matching agent configurations and uses the best match. Pass null to reset to the default (@dust)."
```
aubin-tchoi: how does the agent knows about names of other agents OOC?

front/components/actions/blocked/ToolValidationCard.tsx
```diff
@@ -138,6 +140,18 @@ const MCP_TOOL_OVERRIDES: Partial<
...
+      title: (inputs) => {
+        if (!isPodManagerDefaultAgentInput(inputs)) {
+          return `Allow agent to set the Pod default agent?`;
+        }
+        if (inputs.agentName === null) {
+          return `Allow agent to reset the Pod default agent to @dust?`;
+        }
+        return `Allow agent to set the Pod default agent to @${inputs.agentName}?`;
+      },
+      alwaysAllowLabel: () => `Always allow agent to set the Pod default agent`,
```
aubin-tchoi: nit
```suggestion
      alwaysAllowLabel: () => "Always allow agent to set the Pod default agent",
```

front-api/routes/w/[wId]/skills/index.ts
```diff
@@ -104,6 +105,11 @@ app.get(
     const globalSpaceOnly = ctx.req.query("globalSpaceOnly");
     const onlyCustom = ctx.req.query("onlyCustom");
     const isDefault = ctx.req.query("isDefault");
+    const podSpaceSId = ctx.req.query("podSpaceId");
```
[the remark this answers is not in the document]
Comment (aubin-tchoi) [reply front-api/routes/w/[wId]/skills/index.ts]: 👍🏻

front-api/routes/w/[wId]/skills/index.ts
```diff
@@ -118,6 +124,7 @@ app.get(
     const skills = await SkillResource.listByWorkspace(auth, {
```
Comment (aubin-tchoi) [inline front-api/routes/w/[wId]/skills/index.ts]: let's add a method `listBySpace` instead of reusing `listByWorkspace`

front/lib/api/actions/servers/pod_manager/tools/index.ts
```diff
@@ -411,6 +416,84 @@ export function createProjectManagerTools(
       }, "Failed to set Pod pinned frame");
     },
 
+    [DEFAULT_AGENT_TOOL_NAME]: async (params) => {
+      return withErrorHandling(async () => {
+        const contextRes = await getPod(auth, {
+          toolContext,
+          dustPod: params.dustPod,
```
aubin-tchoi: nit: let's destructure params, this will simplify a bit the handler's code

front/lib/api/actions/servers/pod_manager/tools/index.ts
```diff
@@ -411,6 +416,84 @@ export function createProjectManagerTools(
...
+              { tracked: false }
+            )
+          );
+        }
+
+        const featureFlags = await getFeatureFlags(auth);
+        if (!featureFlags.includes("pod_default_agent")) {
+          return new Err(
+            new MCPError(
+              "Setting a Pod default agent is not enabled for this workspace.",
```
aubin-tchoi: ideally the tool is not shown at all (but I understand if this is a bit hard to do)

front/lib/api/actions/servers/pod_manager/tools/index.ts
```diff
@@ -411,6 +416,84 @@ export function createProjectManagerTools(
...
+                tracked: false,
+              })
+            );
+          }
+          defaultAgentId = resolvedAgentId;
+          const agent = await getAgentConfiguration(auth, {
+            agentId: resolvedAgentId,
+            variant: "extra_light",
+          });
+          defaultAgentName = agent?.name ?? null;
```
aubin-tchoi: if we didn't find the agent we should probably error here

aubin-tchoi: could mean the user does not have the rights to see it for instance

front/lib/api/actions/servers/pod_manager/tools/index.ts
```diff
@@ -411,6 +416,84 @@ export function createProjectManagerTools(
...
+
+        return new Ok(
+          makeSuccessResponse({
+            success: true,
+            defaultAgentId,
+            message: defaultAgentId
+              ? `Pod default agent set to ${
+                  defaultAgentName ? `@${defaultAgentName}` : defaultAgentId
+                }.`
+              : "Pod default agent reset to the default (@dust).",
```
aubin-tchoi: nit: magic variable for `dust`

front/lib/api/actions/servers/pod_manager/metadata.ts
aubin-tchoi: are we sure we want best matches here? wouldn't want to end up with a different agent that happens to have a very similar name, right? depending on the answer to the first comment we could even want an ID here

aubin-tchoi (commented):
directionally LGTM but some details to iron out

meta comment: it seems we have 2 very distinct tasks in one PR here (the restriction fix and the addition of a tool), would split in two 🙏🏻

xebeohp:
I think I accidentally committed the tool code on this branch somehow, it is in a separate PR draft. Will remove the tool code from this branch and apply these comments to that PR after applying the edits and merging the skills restrictions stuff.

front-api/routes/w/[wId]/skills/index.ts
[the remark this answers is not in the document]
Comment (xebeohp) [reply front-api/routes/w/[wId]/skills/index.ts]: Fixed.

front-api/routes/w/[wId]/skills/index.ts
```diff
@@ -120,6 +126,7 @@ app.get(
...
       status: skillStatus,
       globalSpaceOnly: globalSpaceOnly === "true",
+      podSpaceId,
```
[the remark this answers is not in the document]
Comment (xebeohp) [reply front-api/routes/w/[wId]/skills/index.ts]: Fixed.

front-api/routes/w/[wId]/skills/index.ts
Comment (xebeohp) [reply front-api/routes/w/[wId]/skills/index.ts]: Fixed.

front/components/skill_builder/SkillBuilderRequestedSpacesSection.tsx
```diff
@@ -238,7 +238,8 @@ export function SkillBuilderRequestedSpacesSection({
           <p className="text-sm text-muted-foreground">
             Choose which spaces and Pods this skill can access. The skill can
             use their knowledge and capabilities, and only users with access to
-            all selected spaces and Pods can use it.
+            all selected spaces and Pods can use it. A skill restricted to a Pod
+            is only available inside that Pod and not in others.
```
aubin-tchoi, line 244: this is something we want to mirror with a UI pattern IMHO, currently you have a list of checkboxes for the pods

front/lib/resources/skill/skill_resource.ts
```diff
@@ -1332,15 +1332,36 @@ export class SkillResource extends BaseResource<SkillConfigurationModel> {
...
-    if (globalSpaceOnly) {
-      const globalSpace = await SpaceResource.fetchWorkspaceGlobalSpace(auth);
-      return skills.filter((skill) =>
-        skill.requestedSpaceIds.every((id) => id === globalSpace.id)
-      );
+  static async listBySpace(
+    auth: Authenticator,
+    {
+      globalSpaceOnly,
+      podSpaceId,
```
aubin-tchoi, line 1477: let's name it `spaceId` to match the name and not make it specific to pods

front/components/skill_builder/submitSkillBuilderForm.ts
```diff
@@ -124,6 +125,8 @@ export async function submitSkillBuilderForm({
       }
     }
 
+    await invalidateWorkspaceSkills(owner.sId);
```
aubin-tchoi, line 128: why do we need this OOC?

front-api/routes/w/[wId]/skills/index.ts
```diff
@@ -115,9 +121,10 @@ app.get(
     }
     const skillStatus = statusValidation.data;
 
-    const skills = await SkillResource.listByWorkspace(auth, {
+    const skills = await SkillResource.listBySpace(auth, {
       status: skillStatus,
       globalSpaceOnly: globalSpaceOnly === "true",
+      podSpaceId: podSpaceModelId,
```
Comment (aubin-tchoi) [inline front-api/routes/w/[wId]/skills/index.ts:159]: we need to get rid of globalSpaceOnly if we introduce a podSpaceId, the two properties intersect

Comment (aubin-tchoi) [reply front-api/routes/w/[wId]/skills/index.ts:159]: `globalSpaceOnly` should be `listBySpace` with the global space

front/lib/resources/skill/skill_resource.ts
```diff
@@ -1541,11 +1562,28 @@ export class SkillResource extends BaseResource<SkillConfigurationModel> {
       [conversation.spaceId]
     );
 
-    return this.fetchByIds(auth, projectMetadata?.defaultSkillIds ?? [], {
-      agentLoopData,
-      effectiveSpaceIds,
-      onlyActive: true,
-    });
+    const skills = await this.fetchByIds(
```
aubin-tchoi, line 1701: IIUC you're storing the normalized data (i.e. you cannot add a pod default skill that isn't only on the pod), so you probably do not need this

PR #29183, opened 2026-07-20
[front] perf: filter spaces by kind in SQL for useSpaces

spolu:
## Description

Pass requested space kinds from `useSpaces` to the spaces endpoint and apply the filter in PostgreSQL before hydrating space groups.

This prevents the capabilities preloader's global-space request from loading every project and group in the workspace. Project metadata enrichment now only runs when the filtered result contains projects. Requests without a kind keep the existing full response, and SWR mutations continue to invalidate all filtered URL variants.

The private API documentation now describes the repeatable `kind` query parameter.

## Tests

- Added endpoint coverage for repeated kind filtering and conditional project metadata enrichment.
- `NODE_ENV=test npm test -- 'routes/w/[wId]/spaces/index.test.ts'`

## Risk

Medium. Touches spaces listing which is pervasively used.

## Deploy Plan

- deploy `front`

### DIFF
--- a/front-api/public/swagger.json
+++ b/front-api/public/swagger.json
@@ -9637,9 +9637,21 @@
             "in": "query",
             "name": "kind",
             "required": false,
-            "description": "Filter by space kind (e.g. system)",
+            "description": "Filter by one or more space kinds. Repeat the parameter to include several kinds.",
+            "style": "form",
+            "explode": true,
             "schema": {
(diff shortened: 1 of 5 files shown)


spolu:
PMRR but happy to revisit based on your feedback

id13:
LGTM

PR #29187, opened 2026-07-20
[front] perf: avoid system space fetches for MCP views

spolu:
## Description

Avoid fetching and fully hydrating the workspace system space whenever MCP server views are resolved.

The system space was fetched unconditionally by `MCPServerViewResource.baseFetch`, including for empty, remote-only, and auto-internal results. The downstream resolver only needed its ID to verify that manual internal servers still had a canonical system-space view.

Auto and remote servers now skip that lookup entirely. Manual servers are validated with a targeted join between MCP server views and the system space, avoiding the expensive groups hydration. Redundant lookups in the MCP server and approvals routes are removed as well.

## Tests

- `NODE_ENV=test npm test -- lib/resources/mcp_server_view_resource.test.ts`
- `NODE_ENV=test npm test -- routes/w/[wId]/mcp/[serverId]/index.test.ts`

## Risk

Low. Biggest risk is that the new query is expensive, but will stil happen much less often.

## Deploy Plan

- deploy `front`


spolu:
PMRR open to feedback of course!

id13:
LGTM

PR #29190, opened 2026-07-20
[Admin Gov] Rename group_permissions column to grantType (1/2, expand)

philipperolet:
## Description

Renames the `group_permissions` physical column `permissionType` → `grantType` to match the code
vocabulary (the Sequelize attribute has been `grantType` since the rename PR, mapped to the physical
column via `field: "permissionType"`).

`group_permissions` is read on the per-message model-resolution path (`ModelsTierResource.resolveAllowedTierNames`),
so a single-shot `RENAME COLUMN` would break old pods still querying `permissionType` during the
rolling deploy → failed model resolution for every message. The rename is therefore done as
expand/contract. **This is the expand step (1/2):**

- **Pre-deploy migration:** add the `grantType` column, install a bidirectional mirror trigger
  (old pods write `permissionType`, new pods write `grantType`; the trigger fills the counterpart so
  every row is complete during rollout), backfill existing rows, and build the new unique index
  `CONCURRENTLY`.
- **Model:** drop the `field: "permissionType"` mapping (`grantType` now maps to the `grantType`
  column) and point the unique index at `grantType`.
- Folds in two stale verb-storage comment fixes in `group_permission_resource.ts`.

The **contract step (2/2, post-deploy)** — drop `permissionType`, the trigger and old index, and set
`grantType NOT NULL` — is a follow-up PR merged only after this is deployed and pods have cycled.

## Risks

**Blast radius:** the `group_permissions` table (governance capability grants + model-tier grants),
which is read on the model-resolution path exercised by every agent message.

**Risk:** Low. The change is additive + trigger-bridged, so during the deploy window both column
names resolve to complete data and neither old nor new pods break. The migration was verified
end-to-end on a throwaway database: backfill, both write directions (old-code `permissionType`-only
and new-code `grantType`-only inserts), the new unique index build, and idempotent re-apply. The new
column is nullable in this step (trigger + backfill keep it populated); `NOT NULL` is set in the
contract step. tsgo clean; 69 group_permission / models_tier / auth tests pass.

## Deploy Plan

- Standard: the pre-deploy migration runs before the new code. Safe against still-running old pods —
  `grantType` is just an extra column the trigger keeps mirrored.
- After this deploys and pods cycle, merge the contract PR (2/2) to drop `permissionType`, the
  trigger, and the old index, and to set `grantType NOT NULL`.
- Rollback: revert the code; the extra column and trigger are harmless and are removed by the
  contract step (or by re-running once reverted).

### DIFF
--- a/front/lib/resources/group_permission_resource.ts
+++ b/front/lib/resources/group_permission_resource.ts
@@ -363,8 +363,8 @@ export class GroupPermissionResource extends BaseResource<GroupPermissionModel> 
     });
   }
 
-  // Read grants for the given groups, optionally narrowed by verb / type / resource. The
-  // (workspaceId, resourceType, resourceId) index backs type/resource-scoped reads.
+  // Read grants for the given groups, optionally narrowed by grant type / resource type /
+  // resource. The (workspaceId, resourceType, resourceId) index backs type/resource-scoped reads.
   static async listForGroups(
     auth: Authenticator,
     { groupModelIds, grantType, resourceType, resourceId }: ListForGroupsSpec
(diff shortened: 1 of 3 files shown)


philipperolet:

- #29190 ◀
    - #29191

matteotrab (approved):

PR #29192, opened 2026-07-20
[Admin Gov] Sync a manual Builders group from the builder role

philipperolet:
## Description

Related to https://github.com/dust-tt/tasks/issues/9459 — first PR of the builder role deprecation series (backfill  follows in a separate PR). Decisions log: [issue comments](https://github.com/dust-tt/tasks/issues/9459#issuecomment-5032168852).

Design doc: [Admin Governance §5](https://app.notion.com/p/dust-tt/Design-Doc-Admin-Governance-38a28599d941817292d0e94f8dfafe48).

Keeps a per-workspace **"Builders"** group (`regular_manual`) in sync with the `builder` role, so that when the role is removed the group can be granted the builders' governance capabilities (create agents / skills) and former builders keep their rights. Once the role is removed, the sync goes away and the group becomes fully admin-managed.

The group is deliberately **independent from SCIM provisioning**: provisioned `dust-builders` groups proved unreliable in practice (drifted membership, stale groups after deprovisioning). The ~40 workspaces provisioning builders keep their IdP group *and* get the manual group — IdP changes flow through role assignment, which keeps the manual group in sync automatically. They will receive an explanation of the two groups.

`GroupResource.syncBuilderGroupMembership` implements idempotent ensure-state semantics (active membership ends up matching the role) and is called from every membership write path:
- role changes, membership creation (invite acceptance / signup flows) and revocation, via the `lib/api/membership.ts` wrappers;
- SCIM role assignment (`handleRoleAssignmentForGroup`), the poke `apply_group_roles` plugin and the `unrevoke_membership` script, which call the resource methods directly.

Non-obvious points:
- In `createAndTrackMembership`, the sync runs **after** `restoreGroupMembershipsRevokedWith` so a rejoining user's group membership reflects their new role rather than the pre-revoke one.
- Retried writes (`already_on_role`, `already_revoked`) also sync, healing drift when a first attempt failed after the role write but before the group sync.
- Group creation races (two concurrent membership writes) are resolved through the `(workspaceId, name)` unique index.

Also unifies the previously duplicated `ADMIN_GROUP_NAME` / `BUILDER_GROUP_NAME` constants in `workos_events_queue/activities.ts` with the ones exported by `GroupResource`.

## Tests
- [x] Resource tests: lazy group creation, no creation for non-builders, add/remove on role transitions, idempotency, provisioned name-squat skipped, coexistence with a provisioned dust-builders group.

## Risks
Blast radius: membership writes (role change, invite acceptance, revoke, SCIM role sync) now also write to the Builders group.
Risk: low

## Deploy Plan
- deploy front

### DIFF
--- a/front/lib/api/poke/plugins/workspaces/apply_group_roles.ts
+++ b/front/lib/api/poke/plugins/workspaces/apply_group_roles.ts
@@ -104,6 +104,8 @@ export const applyGroupRoles = createPlugin({
             `Failed to update role for user ${user.sId}: ${updateResult.error.type}`
           );
         } else {
+          // Per-changed-member queries, like the role update above: poke-only plugin,
+          // bounded by the number of role changes.
           await GroupResource.syncBuilderGroupMembership({
             workspace,
             user,
--- a/front/lib/api/poke/plugins/workspaces/apply_group_roles.ts
+++ b/front/lib/api/poke/plugins/workspaces/apply_group_roles.ts
@@ -104,6 +104,11 @@ export const applyGroupRoles = createPlugin({
             `Failed to update role for user ${user.sId}: ${updateResult.error.type}`
           );
         } else {
+          await GroupResource.syncBuilderGroupMembership({
+            workspace,
+            user,
+            isBuilder: expectedRole === "builder",
+          });
           updatedCount++;
         }
       }
--- a/front/lib/resources/group_resource.ts
+++ b/front/lib/resources/group_resource.ts
@@ -2648,16 +2648,6 @@ export class GroupResource extends BaseResource<GroupModel> {
       }
     }
 
-    if (group.kind === "provisioned") {
-      // An IdP group squats the name: never mutate provisioned membership. The nightly
-      // consistency check will surface the blocked sync.
-      logger.warn(
-        { workspaceId: workspace.id, groupId: group.id },
-        "Builders group name taken by a provisioned group; skipping builder group sync"
-      );
-      return;
-    }
-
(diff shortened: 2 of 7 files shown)


front/lib/api/poke/plugins/workspaces/apply_group_roles.ts
```diff
@@ -104,6 +104,11 @@ export const applyGroupRoles = createPlugin({
             `Failed to update role for user ${user.sId}: ${updateResult.error.type}`
           );
         } else {
+          await GroupResource.syncBuilderGroupMembership({
```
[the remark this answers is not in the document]
philipperolet: Fine as this method will be deleted once builder role is deprecated

front/lib/resources/group_resource.ts
```diff
@@ -2577,6 +2581,125 @@ export class GroupResource extends BaseResource<GroupModel> {
...
+          where: {
+            workspaceId: workspace.id,
+            name: MANUAL_BUILDERS_GROUP_NAME,
+          },
+        });
+        assert(group, "Builders group missing after unique constraint error");
+      }
+    }
+
+    if (group.kind === "provisioned") {
```
philipperolet: this is very defensive as currently no group named "Builders" exist in prod

philipperolet: Removed the guard and its test (ca67e3b6b2); also dropped the matching guard in the backfill migration for consistency.

tdraier (approved):
LGTM

PR #29195, opened 2026-07-20
feat(triggers): rename schedules_management → triggers_management, add event trigger tools

frankaloia:
## Summary

Renames the `schedules_management` internal MCP server to `triggers_management` and adds event (webhook) trigger support so agents can create event-driven triggers, not just schedules.

- **Rename** `schedules_management` → `triggers_management`. The numeric server id `1020` is unchanged, so persisted sIds are stable — no data migration and existing agents that enabled the server keep working.
- **`list_triggers`** (replaces `list_schedules`) — lists both schedule and event triggers in one view. 
- **`disable_trigger`** (replaces `disable_schedule`) — works on any trigger kind.
- **`list_event_sources`** (new) — dedicated discovery tool listing workspace-configured webhook sources (GitHub, Linear, Jira, Zendesk, …) and the events each emits.
- **`create_event_trigger`** (new) — binds an agent to a webhook source + event. Optional natural-language `filterDescription` is turned into a payload filter via `getWebhookFilterGeneration` and validated with `parseMatcherExpression`. Supports `podId` for Pod attachment. Webhook sources cannot be created by the agent — they must already be configured in the workspace.
- `create_schedule` is unchanged.

## Testing
* Testing all tools locally

## Risk
* Low - should only be additive. SID of the mcp server should not change.

## Deployment
1. Deploy Front


Fraggle (approved):
LGTM

PR #29196, opened 2026-07-20
Update activation recommendation model to support new UX modal

frankaloia:
## Summary
- Drops the rationale field because we don't have a clear use for it and it is not populated for now
- Adds the title field which we require a s short, user-facing label that will be surfaced in the UX in the following design: https://app.dust.tt/share/frame/8afb4896-9522-437a-b25f-a22dd0a278c5
- Adding composite index to cheaply query open recommendations for user

## Testing

- Resource tests: 10/10 pass

## Risk

Low — None of this is surfaced to users yet

## Deployment
1. Run migration
2. Deploy front

### DIFF
--- a/front/lib/models/activation/activation_recommendation.ts
+++ b/front/lib/models/activation/activation_recommendation.ts
@@ -84,6 +84,11 @@ ActivationRecommendationModel.init(
         concurrently: true,
       },
       {
+        name: "idx_activation_recommendations_user",
+        fields: ["userId"],
+        concurrently: true,
+      },
+      {
         fields: ["conversationId"],
         concurrently: true,
       },
--- a/front/lib/models/activation/activation_recommendation.ts
+++ b/front/lib/models/activation/activation_recommendation.ts
@@ -22,8 +22,8 @@ export class ActivationRecommendationModel extends WorkspaceAwareModel<Activatio
 
   declare userId: ForeignKey<UserModel["id"]>;
   declare status: ActivationRecommendationStatus;
+  declare title: string;
   declare content: string;
-  declare rationale: string;
 
   // The conversation in which the recommendation was (originally) made
   declare conversationId: ForeignKey<ConversationModel["id"]> | null;
@@ -53,11 +53,11 @@ ActivationRecommendationModel.init(
       type: DataTypes.STRING,
       allowNull: false,
     },
-    content: {
+    title: {
       type: DataTypes.STRING(4096),
       allowNull: false,
     },
-    rationale: {
+    content: {
       type: DataTypes.STRING(4096),
       allowNull: false,
     },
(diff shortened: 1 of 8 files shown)


front/lib/models/activation/activation_recommendation.ts
```diff
@@ -79,11 +79,8 @@ ActivationRecommendationModel.init(
     sequelize: frontSequelize,
     indexes: [
       {
-        fields: ["userId"],
-        concurrently: true,
```
id13, line 83: userId is a foreign key, this index needs to stay

front/migrations/pre-deploy/20260720120000_add_title_drop_rationale_activation_recommendations.sql
```diff
@@ -0,0 +1,30 @@
...
+SET SESSION statement_timeout = 1200000;
+SET SESSION lock_timeout = 3000;
+DROP INDEX CONCURRENTLY IF EXISTS "activation_recommendations_workspace_id";
+
+/*
+Statement 3
+*/
+SET SESSION statement_timeout = 1200000;
+SET SESSION lock_timeout = 3000;
+CREATE INDEX CONCURRENTLY "activation_recommendations_workspace_user_status_created_at_idx" ON public.activation_recommendations USING btree ("workspaceId", "userId", "status", "createdAt");
```
id13: how many different values of status do you anticipate per workspace x user ? and how many of which type ? 
I assume suggested very few, because they would fall in either dismissed or executed, with most of the rows in executed.

so an index is not going to help much there 🤔 

I would instead do (workspaceId, userId, createdAt DESC) if you know you are going to fetch the latest recommendation, otherwise I would just do (workspaceId, userId)

front/lib/models/activation/activation_recommendation.ts
frankaloia: Yes indeed

front/migrations/pre-deploy/20260720120000_add_title_drop_rationale_activation_recommendations.sql
frankaloia: There are currently 3 statuses: suggested, executed, dismissed. The use case here is getting all recommendations in the suggested state. It's likely that will correlate to the most recently created, but not necessarily guaranteed. If we remove the status, we are forcing us to go through all of the recommendations for the user, which _should_ be on the order of 10s-100s, which feels doable, but there is no strict guarantee on the number of recommendation.

Based on that are you still recommending the simpler composite? I think I'm still leaning towards including status

id13: I still think we should go about workspaceId, userId, 10-100 rows is nothing for an index filter. Especially if we see a big skew towards one of the three value which I'm confident we will :+1:

frankaloia: Ok, no push back

id13 (approved):

PR #29200, opened 2026-07-21
Fix task-list checkbox alignment in markdown

Inesvvv:
## Description

Fixed the alignment of task lists

Result in storybook: 
Screenshot 2026-07-21 at 09 38 53

## Tests

Tested locally result in storybook 

## Risk

Low

## Deploy Plan

Deploy of sparkle

### DIFF
--- a/sparkle/src/components/markdown/List.tsx
+++ b/sparkle/src/components/markdown/List.tsx
@@ -8,22 +8,44 @@ import { cn } from "@sparkle/lib";
 import { cva } from "class-variance-authority";
 import React, { memo } from "react";
 
-export const ulBlockVariants = cva(["list-disc pb-2 pl-6 flex flex-col gap-1"]);
+export const ulBlockVariants = cva("pb-2 flex flex-col gap-1", {
+  variants: {
+    taskList: {
+      false: "list-disc pl-6",
+      true: "",
+    },
+  },
+  defaultVariants: {
+    taskList: false,
+  },
+});
 
 interface UlBlockProps {
   children: React.ReactNode;
+  className?: string;
   node?: MarkdownNode;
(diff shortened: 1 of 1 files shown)


sparkle/src/components/markdown/List.tsx
```diff
@@ -8,22 +8,44 @@ import { cn } from "@sparkle/lib";
...
+  className?: string;
   node?: MarkdownNode;
 }
 
 export const UlBlock = memo(
-  ({ children }: UlBlockProps) => {
+  ({ children, className }: UlBlockProps) => {
     const { textColor, forcedTextSize } = useMarkdownStyle();
     const textSize = forcedTextSize ?? markdownParagraphSize;
+    const isTaskList = className?.includes("contains-task-list");
```
PopDaph, line 33: Should we add a comment to explain that is is a magic class from remark gfm? 
@ykmsd as owner of sparkle, is it okay for you? I think it's safe enough since documented here in their readme https://github.com/remarkjs/remark-gfm and at the same time not the end of the world if it breaks. 

If we want to be a bit cleaner we could do: 

```ts
// Classes added by remark-gfm on GFM task lists.
const CONTAINS_TASK_LIST_CLASS = "contains-task-list";
const TASK_LIST_ITEM_CLASS = "task-list-item";

const hasClass = (className: string | undefined, token: string) =>
  className?.split(" ").includes(token) ?? false;

Then in UlBlock:

const isTaskList = hasClass(className, CONTAINS_TASK_LIST_CLASS);

and in LiBlock:

const isTaskListItem = hasClass(className, TASK_LIST_ITEM_CLASS);
```

PopDaph (approved):

PR #29205, opened 2026-07-21
Prune historical reasoning in checkpointed conversation windows

flvndvd:
## Description

This PR adds historical reasoning pruning to the recently introduced checkpointed conversation window. V2 still never drops interactions: user messages, assistant answers, tool calls, and conversation ordering remain available to the model.

The pruning boundary is the latest real user message. Reasoning generated after that point is preserved unchanged, including reasoning interleaved with tool calls and results. Injected system messages do not advance this boundary. 
About providers:
- Anthropic requires this active tool-use reasoning to be returned unchanged but allows thinking from earlier assistant turns to be omitted. See Anthropic’s extended-thinking documentation
(https://platform.claude.com/docs/en/build-with-claude/extended-thinking#preserving-thinking-blocks).
- OpenAI similarly recommends preserving every reasoning and function-call item since the latest user message, while its `current_turn` mode excludes reasoning from earlier turns. OpenAI’s reasoning guidance
(https://developers.openai.com/api/docs/guides/reasoning#keeping-reasoning-items-in-context).

Historical reasoning is only removed under context pressure. It shares the existing 20k pruning checkpoint with consumed tool results, so both contribute to one predictable cache rewrite. Tool results are pruned first, reasoning is used only when they cannot provide enough headroom. Previously pruned content is never restored.

### DIFF
--- a/front/lib/api/assistant/conversation_rendering/checkpointed_window_state.test.ts
+++ b/front/lib/api/assistant/conversation_rendering/checkpointed_window_state.test.ts
@@ -13,11 +13,15 @@ function withTokens<T extends ModelMessageTypeMultiActions>(
   return { ...message, tokenCount };
 }
 
-function userMessage(name: string, tokenCount: number) {
+function userMessage(
+  name: string,
+  tokenCount: number,
+  { messageName = "user" }: { messageName?: string } = {}
+) {
   return withTokens(
     {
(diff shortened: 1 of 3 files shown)


flvndvd:
Closing at the gain is not clear yet, and current approach is good enough.

PR #29207, opened 2026-07-21
Fix agent reasoning effort resolution

fontanierh:
## Description

Use the reasoning effort configured on an agent when model resolution does not have an explicit user selection. Keep falling back to the resolved model default when the requested effort is unsupported.

This fixes global agents configured for high effort being silently resolved at the model default effort.

## Tests

- Added regression coverage for a supported high agent effort.
- Added coverage for unsupported agent effort fallback.
- `NODE_ENV=test npm run test -- lib/api/assistant/models.test.ts` (22 tests passed)
- `npx tsgo --noEmit`
- `npx biome check --write --error-on-warnings front/lib/api/assistant/resolve_model.ts front/lib/api/assistant/models.test.ts`

## Risk

Low. The change only affects resolution without an explicit model selection. Unsupported efforts retain the current model-default fallback.

## Deploy Plan

Deploy normally. No migration or configuration change is required.

### DIFF
--- a/front/lib/api/assistant/models.test.ts
+++ b/front/lib/api/assistant/models.test.ts
@@ -19,12 +19,14 @@ import { AUTO_MODEL_ID } from "@app/types/assistant/models/auto";
 import {
   GPT_5_4_MINI_MODEL_CONFIG,
   GPT_5_5_MODEL_CONFIG,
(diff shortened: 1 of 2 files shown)


Nils-Fedrigo (approved):
thanks for the catch

PR #29208, opened 2026-07-21
feat(front): rework input bar model picker with tiers and streams

Nils-Fedrigo:
## Description

Reworks the input-bar model picker (behind the `models_picker` flag) around three tiers, and wires the two new curated auto-routing streams from the backend PR #29229 into it.

- **Quick / Standard / Deep** are the primary picks, each with a subtitle (Simple / Everyday / Heavy tasks), a star icon, and a check when active.
- **Quick and Deep now auto-route** among a curated pool (the `auto_quick` / `auto_deep` stream meta-models): the tier sends its sentinel and Dust picks the first available model in the pool at send-time — the same mechanism as Standard/`auto`, but scoped. A tier shows when at least one of its candidates is available to the workspace, and carries its own tooltip. The stream sentinels are filtered out of "More models".
- **Standard behaves like the old "Auto"**: it sends the Auto selection (Dust picks any model). The "Auto" toggle is removed and Standard is the default fallback — an agent whose default is Auto (or no agent default) resolves to Standard.
- **Agent defaults and other non-tier models live under "More models"**: when the active selection is a concrete non-tier model, a check cascades down the path — More models → provider → model.
- **"More models"** nests the search bar plus per-maker submenus; provider sections drop per-model title labels (each row carries its own name + effort chip).
- **Input-bar button is icon-only**: the tier's star when a tier is active, otherwise the model maker's logo, with a tooltip for the current selection.

Stacked on #29229 (backend stream routing). Supersedes #29231, whose picker wiring is folded into this PR.

## Tests

Manual, behind the `models_picker` flag: verify tier selection, that Standard sends Auto, that Quick/Deep route through the streams, the check cascade into More models for agent/non-tier defaults, and search within More models. Typecheck (`tsgo`) and biome pass.


Nils-Fedrigo:

- #29229
    - #29208 ◀

PR #29210, opened 2026-07-21
Clarify additive Space selection in the input bar

fontanierh:
## Description

Clarifies that Spaces selected from the input bar extend the agent's configured Spaces rather than replace them. The picker now shows Agent's Spaces separately from Additional Spaces, and its tooltip uses additive wording.

Screenshot 2026-07-21 at 11 40 08

## Tests

Tested locally in a seeded Hive with open and restricted Spaces. Verified selection, persistent dropdown behavior, filtering, icons, and selection count updates.

## Risk

Low. This only changes picker copy and menu structure.

## Deploy Plan

Standard deploy.


fabiencelier (commented):
LGTM ✅

PR #29211, opened 2026-07-21
Self-improving skills: mark notification as read when suggestion are handled

fabiencelier:
## Description

Closes https://github.com/dust-tt/tasks/issues/8768

The purpose is to avoid being notified by some suggestion another user has already handled.
When the last suggestion of the conversation has been handled, mark the conversation as read for everybody so 

## Tests

Added unit tests for the new resources methods

## Risk

low: just about notifications + easy to revert

## Deploy Plan

deploy front


frankaloia (approved):
Clean

PR #29219, opened 2026-07-21
[front] perf: avoid global space fetches in the agent loop

id13:
## Description

We query a lot spaces x group vault x group just to get the global space id.
This code remove entirely the lookup mcp server view (which is a hot code path) and datasource (side-drive)

Left untouched (genuinely need the resource, LLM-frequency only): the `toolsets` and `pod_manager` tool handlers.

## Tests

Unit tests

## Risk

Standard
Blast radius: agent loop mostly

## Plan

deploy front

### DIFF
--- a/front/lib/resources/data_source_view_resource.ts
+++ b/front/lib/resources/data_source_view_resource.ts
@@ -335,14 +335,33 @@ export class DataSourceViewResource extends ResourceWithSpace<DataSourceViewMode
     spaceIds: string[],
     { includeGlobalSpace = false }: { includeGlobalSpace?: boolean } = {}
   ) {
-    const requestedSpaces = await SpaceResource.fetchByIds(auth, spaceIds);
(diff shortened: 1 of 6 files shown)


front/lib/resources/data_source_view_resource.ts
```diff
@@ -335,14 +335,33 @@ export class DataSourceViewResource extends ResourceWithSpace<DataSourceViewMode
     spaceIds: string[],
     { includeGlobalSpace = false }: { includeGlobalSpace?: boolean } = {}
   ) {
-    const requestedSpaces = await SpaceResource.fetchByIds(auth, spaceIds);
+    const spaceModelIds = removeNulls(spaceIds.map(getResourceIdFromSId));
```
id13, line 338: 🙉

spolu (approved):
LGTM

PR #29220, opened 2026-07-21
[front] refactor: add generic `createServer` for internal servers

aubin-tchoi:
## Description

Follow up on https://github.com/dust-tt/dust/pull/29119 

This PR introduces a generic `createServer` that builds a server from its existing metadata and handlers.

Tools can declare context-dependent availability through `isAvailableForContext`, which we'll use to determine in which context a tool should be exposed (applies to both `listTools` and `runTool`). Ultimately this will be leveraged in pod functions, to exclude functions that can only be ran in the context of a pod function or a conversation or vice versa.

This PR moves `common_utilities` to provide a first example, a follow up PR will do the more mechanical work of moving every other one. The goal here is to validate the new typing and structure.

## Tests

- Covered by existing tests.

## Risk

- Low.

## Deploy Plan

- Deploy front.


front/lib/actions/mcp_internal_actions/servers/index.ts
```diff
@@ -149,7 +199,11 @@ export async function getInternalMCPServer(
     case "jit_testing":
       return jitTestingServer(auth, toolContext);
     case "common_utilities":
-      return commonUtilitiesServer(auth, toolContext);
```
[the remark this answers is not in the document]
aubin-tchoi: 🤔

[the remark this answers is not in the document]
aubin-tchoi: this is outdated, will fix

front/lib/actions/mcp_internal_actions/tool_definition.ts
```diff
@@ -59,10 +63,10 @@ export interface ToolDefinition<
   displayLabels: ToolDisplayLabels;
   toolCostCategory: ToolCostCategory;
   freeUsage: boolean;
-  handler: (
+  handler(
```
aubin-tchoi, line 66: we use the method syntax here; otherwise ToolDefinition widens the schema and we can't assign it (even though `z.infer<z.ZodObject<TSchema>>` is technically a subset of `{ [z: string]: any }`

front/lib/actions/mcp_internal_actions/tool_definition.ts
```diff
@@ -123,17 +132,14 @@ export function buildClientTools<T extends Record<string, ClientToolMeta>>(
...
-  handlers: ToolHandlers<T>
-): ToolDefinition[] {
-  return metadata.map(
-    (tool) =>
-      ({
-        ...tool,
-        handler: handlers[tool.name as keyof ToolHandlers<T>],
-      }) as unknown as ToolDefinition
-  );
+export function buildTools<
```
aubin-tchoi, line 135: this is not be needed in the new `createServer`, just aligned the typing here to show that we don't need the 2 type casts (therefore we don't need them in the new `createServer`)

front/lib/api/actions/servers/common_utilities/tools/index.ts
```diff
@@ -13,7 +12,9 @@ import { compile } from "mathjs";
 
 const RANDOM_INTEGER_DEFAULT_MAX = 1_000_000;
 
-const handlers: ToolHandlers<typeof COMMON_UTILITIES_TOOLS_METADATA> = {
+export const COMMON_UTILITIES_TOOL_HANDLERS: ToolHandlers<
```
aubin-tchoi, line 15: so the new pattern is: export the handlers, then use them in `createServer`, which binds them to the server name

aubin-tchoi:
ultimately won't do (see https://dust4ai.slack.com/archives/C09GELMTTRT/p1784629944509409)
but open to revisit at any time

PR #29215, opened 2026-07-21
Update hasWorkspacePermissions to take a verb

tdraier:
fixes: https://github.com/dust-tt/tasks/issues/9452

## Description

`Authenticator.hasWorkspacePermission` now takes a `GrantVerb` (a capability question, e.g. `"create"`) and a `ConcreteResourceType`, instead of taking a stored `GrantType` (role name) directly. Internally it expands the verb into the `ROLE_REGISTRY` roles that imply it, via a new `grantTypesForVerb(resourceType, verb, level)` helper in `group_permission_registry.ts`, then checks the caller's groups for a matching type-wide (`-1`) `group_permissions` row.

This is step 0 of the Admin Governance `create agent` capability rollout (dust-tt/tasks#9463, design doc: [Admin Governance](https://app.notion.com/p/dust-tt/Design-Doc-Admin-Governance-38a28599d941817292d0e94f8dfafe48)), landing on its own with **zero behavior change** — nothing outside its own test calls `hasWorkspacePermission` yet. Follow-ups will (1) enforce the `create agent` capability at the two agent-creation call sites, and (2) backfill `group_permissions` rows per workspace based on the legacy `disallow_agent_creation_to_users` feature flag.

## Tests

- `front/lib/auth.workspace_permission.test.ts` required no behavioral changes — verb and grant-type strings coincide for every case it exercises (`create`/`agent`, `publish`/`agent`, `admin`/`billing`, `read`/`audit_log`, plus the wildcard and rejection cases). Renamed the local reference used at call sites from `CAPABILITY.grantType` to `VERB` for clarity, since it's now passed as the verb parameter, not a stored grant type.
- `npx tsgo --noEmit` clean.

## Risk

Low. `hasWorkspacePermission` has no other callers yet, so this is an isolated signature change with no behavioral impact on any production code path.

## Deploy Plan

Standard deploy, no special considerations.

PR #29221, opened 2026-07-21
Restric agent creation based on permission

tdraier:
## Description

Server-side enforcement of the `create agent` governance capability (dust-tt/tasks#9463, design doc: [Admin Governance](https://app.notion.com/p/dust-tt/Design-Doc-Admin-Governance-38a28599d941817292d0e94f8dfafe48)) at the two entry points that create brand-new agents:

- `createAgentConfiguration` (`front/lib/api/assistant/configuration/agent.ts`): the check runs once `existingAgent` is resolved to `null` — i.e. both when no `agentConfigurationId` was given, and when one was given but didn't match a real row (otherwise a caller could bypass the check by passing a nonexistent id and taking the "create new" branch). Creating a new version of an existing agent is unaffected.
- `createPendingAgentConfiguration` (same file): always creates a brand-new pending agent, so the check is unconditional. Its return type changed from a plain `{ sId }` to `Result<{ sId }, Error>` so the restriction is a normal `Err`, not a thrown exception that would 500 (per [ERR1]/[BACK18]). Its one route caller (`create-pending.ts`) now maps `Err` to `400`/`assistant_saving_error`, matching the sibling `POST /assistant/agent_configurations` route's existing handling of the publish restriction.

Both call `auth.hasWorkspacePermission("create", "agent")` (from the prior PR on `update-hasworkspacepermissions`). Admins bypass unconditionally; anyone else needs a `group_permissions` grant — none exist yet, so this lands as a no-op in practice until the follow-up backfill (step 2) seeds grants from the legacy `disallow_agent_creation_to_users` flag.

**UI is explicitly out of scope for this PR** — it will be done later, once the dedicated governance endpoints/hooks for this capability exist. Today's UI still only gates on `disallow_agent_creation_to_users` client-side; that's unchanged here.

## Tests

- New tests in `agent.test.ts` covering: a non-privileged user rejected on both entry points (including the nonexistent-`agentConfigurationId` bypass case), a user granted via group membership succeeding, and creating a new version of an existing agent being unaffected.
- Fixed the ripple from `createPendingAgentConfiguration`'s signature change across its existing call sites (`activities.test.ts`, `agent.test.ts`, `[aId]/index.test.ts`, `create-pending.test.ts`).
- `AgentConfigurationFactory.createTestAgent` (the shared test fixture used by ~60 other test files) now performs the actual creation via an internal admin authenticator, since many tests construct an `Authenticator` directly without a workspace membership — `authorId`/`editors` stay explicit params, so this doesn't change who the created agent is attributed to, only bypasses the capability check for the fixture. It also re-refreshes the caller's own `auth` afterward, since `createAgentConfiguration` refreshes its own `auth` argument as a side effect of creating the new editor group, and a few tests rely on that to pick up group memberships added earlier in the same test.
- Ran the full `front` test suite; only pre-existing, unrelated failures remain (a sandbox `grep` CLI exit-code test untouched by this change).
- `npx tsgo --noEmit` clean in `front` and `front-api`.

## Risk

Low today (no grants exist yet, so the check only ever passes for admins — matches current effective behavior for admins and is a no-op for everyone else until the backfill runs). The real risk window opens with the backfill PR: it must land and run before this restriction meaningfully changes behavior for non-admins, otherwise non-admins lose the ability to create agents until backfilled.

## Deploy Plan

Deploy after `update-hasworkspacepermissions`. Land and run the backfill script (follow-up PR) before or immediately after this deploys, so non-admin agent creation isn't interrupted for workspaces that currently allow it.

### DIFF
--- a/front-api/routes/w/[wId]/assistant/agent_configurations/[aId]/index.test.ts
+++ b/front-api/routes/w/[wId]/assistant/agent_configurations/[aId]/index.test.ts
@@ -136,7 +136,11 @@ describe("PATCH /api/w/:wId/assistant/agent_configurations/:aId - pending agent"
     });
     await SpaceFactory.defaults(auth);
 
-    const { sId: pendingId } = await createPendingAgentConfiguration(auth);
+    const pendingAgentRes = await createPendingAgentConfiguration(auth);
+    if (pendingAgentRes.isErr()) {
+      throw pendingAgentRes.error;
+    }
+    const { sId: pendingId } = pendingAgentRes.value;
 
     const response = await patch(workspace, pendingId, {
       assistant: {
--- a/front-api/routes/w/[wId]/assistant/agent_configurations/create-pending.test.ts
+++ b/front-api/routes/w/[wId]/assistant/agent_configurations/create-pending.test.ts
@@ -13,6 +13,7 @@ function createPending(workspace: { sId: string }) {
 describe("POST /api/w/:wId/assistant/agent_configurations/create-pending", () => {
   it("creates a pending agent and returns sId; pending agent has correct status and placeholder values", async () => {
     const { workspace, user } = await createPrivateApiMockRequest({
+      role: "admin",
       method: "POST",
     });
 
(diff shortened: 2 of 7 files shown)


tdraier:

- #29215
    - #29221 ◀
        - #29224

matteotrab (approved):

PR #29224, opened 2026-07-21
Seed capabilities + migration script

tdraier:
## Description

Implements the `create agent` capability backfill (dust-tt/tasks#9463) and the workspace-creation lifecycle hook for governance capability seeding (dust-tt/tasks#9454), sharing one registry between the two.

**`front/lib/api/permissions/governance_seeding.ts`** — was a stub (`CAPABILITY_SEEDERS = []`, no-op `seedWorkspaceCapabilities`). Now the single source of truth for governance capability defaults:
- `CAPABILITY_SEEDERS`: a list of `{ capability, resolveTarget(auth) }` entries. `resolveTarget` decides, per workspace, whether a capability should go to `"everyone"`, `"builders"`, or `"admins_only"`. Adding a new capability later is just adding an entry here — nothing else needs to change. Currently seeds `create agent`, resolved from the legacy `disallow_agent_creation_to_users` flag.
- `applyCapabilityTarget(auth, capability, target, { dryRun? })`: the shared write path for a resolved target (`setForEverybody` / grant the workspace's "Builders" group / no-op for admins-only). Supports `dryRun` so a caller can predict the outcome without writing.
- `seedWorkspaceCapabilities(auth)`: applies every registered seeder unconditionally — called once from `createWorkspaceInternal` (already wired up from #9454's initial empty-list version, so no call-site changes needed). A brand-new workspace naturally has no legacy flag set, so `resolveTarget` resolves to the same default a backfill would apply — one resolver serves both a fresh workspace and an existing one.

**`front/lib/resources/group_resource.ts`** — added `GroupResource.fetchOrCreateManualBuildersGroup(workspace)`, so a "builders" target never leaves a capability unconfigured just because no builder has been synced into the group yet (it creates the group empty). `syncBuilderGroupMembership` now delegates to this same helper instead of duplicating the find-or-create-with-race-handling logic, so there's exactly one place that creates the Builders group.

**`front/migrations/20260721_backfill_create_agent_capability.ts`** — backfills existing workspaces from the same `CAPABILITY_SEEDERS` registry. Adds only what a one-time backfill needs on top of the shared seeders: an idempotency check (skip a workspace/capability pair that's already configured, so re-running after a manual override is a no-op) and dry-run-by-default / `--execute` gating. Pass `--wId <workspaceId>` for a single workspace.

## Tests

- New/updated tests in `governance_seeding.test.ts`: a fresh workspace defaults `create agent` to everyone; a workspace with the legacy flag already set gets the Builders group created and granted (verified a plain user is still blocked, and a synced builder gets access).
- `group_resource.test.ts`'s existing `syncBuilderGroupMembership` suite (34 tests) passes unchanged against the refactored implementation — it asserts final DB state, not internal call patterns, so behavior is provably unchanged there.
- `lib/api/membership.test.ts` (the main production caller of `syncBuilderGroupMembership`) passes unchanged.
- Verified the migration script end-to-end against the test DB with a throwaway smoke test (not committed — this repo has no test convention for migration scripts): everybody-grant, builders-grant (including creating the group from scratch), dry-run (no writes), and idempotent re-run.
- `npx tsgo --noEmit` and `biome check` clean.

## Risk

Low. `seedWorkspaceCapabilities` only affects brand-new workspaces going forward (adds a `create`/`agent` grant that didn't exist before — harmless, since step 1's enforcement PR is what actually starts checking it). The backfill script only writes when re-run with `--execute`; nothing here changes behavior for existing workspaces until that flag is passed.

## Deploy Plan

Deploy after `restric-agent-creation-based-on`. Run the backfill script (`--execute`, no `--wId` for all workspaces) before or immediately after this deploys, so non-admin agent creation isn't interrupted for workspaces that currently allow it.

### DIFF
--- a/front/lib/api/permissions/governance_seeding.test.ts
+++ b/front/lib/api/permissions/governance_seeding.test.ts
@@ -1,16 +1,70 @@
 import { seedWorkspaceCapabilities } from "@app/lib/api/permissions/governance_seeding";
 import { Authenticator } from "@app/lib/auth";
+import { FeatureFlagResource } from "@app/lib/resources/feature_flag_resource";
+import { GroupResource } from "@app/lib/resources/group_resource";
 import { GroupFactory } from "@app/tests/utils/GroupFactory";
+import { MembershipFactory } from "@app/tests/utils/MembershipFactory";
+import { UserFactory } from "@app/tests/utils/UserFactory";
 import { WorkspaceFactory } from "@app/tests/utils/WorkspaceFactory";
 import { describe, expect, it } from "vitest";
 
 describe("seedWorkspaceCapabilities", () => {
-  it("is a safe no-op with the empty Phase 0 seeder list", async () => {
+  it("grants create-agent to everybody by default for a fresh workspace", async () => {
     const workspace = await WorkspaceFactory.basic();
     await GroupFactory.defaults(workspace);
     const auth = await Authenticator.internalAdminForWorkspace(workspace.sId);
 
-    // Empty registry: resolves without touching the database.
-    await expect(seedWorkspaceCapabilities(auth)).resolves.toBeUndefined();
+    await seedWorkspaceCapabilities(auth);
+
+    const user = await UserFactory.basic();
+    await MembershipFactory.associate(workspace, user, { role: "user" });
+    const userAuth = await Authenticator.fromUserIdAndWorkspaceId(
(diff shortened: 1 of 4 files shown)


tdraier:

- #29224 ◀

matteotrab (approved):

PR #29117, opened 2026-07-17
feat(activation): gate nudges on the pod's user credit state

ElPicador:
## Description

Skip re-nudging an Activation Pod when its user is out of credits. Since each
pod is single-user, `isEligibleForNudge` now takes that user and consults
`isUserBlocked` (workspace pool exhausted or per-user cap reached) alongside
the existing frequency-cap and unanswered-cap gates. The scheduler activity
batch-fetches each pod's member (`fetchDistinctActiveManualGroupMembersBySpaces`)
to avoid an N+1 lookup across pods.

## Tests

Added unit tests to `nudge.test.ts` covering: pod skipped when the user is
blocked (even if never nudged before), pod still eligible when the user is
not blocked, and the credit gate is skipped entirely when no user is
resolved for the pod. `isUserBlocked` is mocked, matching the existing
pattern in `credit_check.test.ts`.

PR #29226, opened 2026-07-21
feat(activation): schedule pod nudges across the workday

ElPicador:
## Description

Builds the Temporal orchestration layer for pod activation nudges. The
`activation_scheduler` workflow now enumerates and gates every pod once per
day, then visits each eligible pod at a deterministic slot spread across the
regional workday window (9:30 - 16:30) instead of nudging them all at once.
Slot assignment reuses the `spaceModelId % windowMinutes` pattern already
used in `project_task`.

Each pod is re-gated right before sending, since state can go stale between
the morning pass and the pod's slot later in the day; this re-check doubles
as the idempotency guard against activity retries or workflow replays. Also
adds an on-demand `signalWithStart` trigger (`triggerActivationWorkspaceWorkflow`)
so poke/admin-forced cycles run through the same code path as the scheduled
run, targeting the same deterministic workflow id.

## Tests

Added `slots.test.ts` for the deterministic slot assignment (determinism,
bounds, spread). Ran the full `activation_scheduler` test suite plus
`npx tsgo --noEmit`.

### DIFF
--- a/front/temporal/activation_scheduler/activities.ts
+++ b/front/temporal/activation_scheduler/activities.ts
@@ -1,21 +1,184 @@
-import { runActivationForWorkspace } from "@app/lib/api/activation/orchestrator";
+import { isEligibleForNudge } from "@app/lib/api/activation/nudge";
+import { determineEligibleActivationUsers } from "@app/lib/api/activation/orchestrator";
+import { emitActivationEvent } from "@app/lib/api/activation/trigger";
+import { config, REGION_TIMEZONES } from "@app/lib/api/regions/config";
+import { Authenticator } from "@app/lib/auth";
+import { ActivationNudgeResource } from "@app/lib/resources/activation_nudge_resource";
+import { SpaceResource } from "@app/lib/resources/space_resource";
+import { TriggerResource } from "@app/lib/resources/trigger_resource";
+import { UserResource } from "@app/lib/resources/user_resource";
+import { concurrentExecutor } from "@app/lib/utils/async_utils";
+import logger from "@app/logger/logger";
+import {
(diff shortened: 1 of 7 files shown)


ElPicador:

- #29031
    - #29033
        - #29059
            - #29065
                - #29067
                    - #29103
                        - #29105
                            - #29117
                                - #29226 ◀
                                    - #29244
                                        - #29260

front/temporal/activation_scheduler/activities.ts
```diff
@@ -60,53 +70,93 @@ export async function runActivationForWorkspaceActivity({
...
-    firedTriggersByPod.map(({ triggerId }) => triggerId)
-  );
-  const triggerById = new Map(
-    triggers.map((trigger) => [trigger.sId, trigger])
-  );
+  const membersByPod =
+    await SpaceResource.fetchDistinctActiveManualGroupMembersBySpaces(auth, [
+      pod,
+    ]);
+  const [user] = membersByPod.get(pod.id) ?? [];
```
frankaloia: We need to think about how we are defining which users are in scope for activation. We are making assumptions that it will be 1 main person being activated, which is what are aiming for, but there could be multiple members in the pod. For example, a few beta customer have already requested that managers be added to the pod of their workers. I'm wondering if we should be sending a nudge per every user in the pod (maybe we can do it for pod editors only), then in the trigger, we filter by user ID, thoughts?

ElPicador: => https://dust4ai.slack.com/archives/C0BCQDK82JX/p1784708996622209

PR #29227, opened 2026-07-21
[front] enh: cache equipped skills for OpenAI models

aubin-tchoi:
## Description

Goal is to add a cache breakpoint for the equipped skills (contained in a user message) for OpenAI models. This part is much more stable than the rest of the conversation, as it can be typically shared across.

This is the same as what we currently do for Anthropic models.

Only works for GPT-5.6 models as per their doc.

Link to the doc https://developers.openai.com/api/docs/guides/prompt-caching

Tested [here](https://langfuse.dust.tt/project/cmi4ft2qm00061707q4a9azyc/traces?filter=metadata%3BstringObject%3BconversationId%3B%3D%3B5sgVosqPS4&peek=be95a9d69821ed81b387732a82211d53&timestamp=2026-07-21T14%3A50%3A12.512Z&observation=b880b8e5006e8fa7)
Tested [here](https://langfuse.dust.tt/project/cmi4ft2qm00061707q4a9azyc/traces?filter=metadata%3BstringObject%3BconversationId%3B%3D%3B9jPVLycLcG&peek=33eb6ff1b8c8ba1e4588cf9d151921d0&timestamp=2026-07-21T21%3A33%3A07.963Z&observation=89afcd8be1c0541b) that we don't add a breakpoint on older models.

## Tests

- Added test that check where the breakpoint is placed.
- Tested locally.

## Risk

- Low.

## Deploy Plan

- Deploy front.

### DIFF
--- a/front/lib/api/llm/clients/openai/index.ts
+++ b/front/lib/api/llm/clients/openai/index.ts
@@ -116,6 +116,8 @@ export class OpenAIResponsesLLM extends LLM<ResponseCreateParamsStreaming> {
     return {
       model: this.modelId,
       input: toInput(promptText, conversation, "developer", {
+        // The legacy registry includes custom model IDs that are absent from
+        // model_constructors, so narrow before using its capability helper.
         cacheBreakpointOnLeadingMessage:
           isModel(this.modelId) &&
           supportsOpenAIExplicitPromptCaching(this.modelId),
--- a/front/lib/api/llm/clients/openai/index.ts
+++ b/front/lib/api/llm/clients/openai/index.ts
@@ -32,10 +32,7 @@ import {
   streamLLMEvents,
 } from "@app/lib/api/llm/utils/openai_like/responses/openai_to_events";
 import type { Authenticator } from "@app/lib/auth";
-import {
-  isModel,
-  supportsOpenAIExplicitPromptCaching,
-} from "@app/lib/model_constructors/types/models";
+import { supportsOpenAIExplicitPromptCaching } from "@app/lib/model_constructors/types/models";
 import logger from "@app/logger/logger";
 import type { Result } from "@app/types/shared/result";
 import { Err, Ok } from "@app/types/shared/result";
@@ -116,11 +113,9 @@ export class OpenAIResponsesLLM extends LLM<ResponseCreateParamsStreaming> {
     return {
       model: this.modelId,
       input: toInput(promptText, conversation, "developer", {
-        // The legacy registry includes custom model IDs that are absent from
-        // model_constructors, so narrow before using its capability helper.
-        cacheBreakpointOnLeadingMessage:
-          isModel(this.modelId) &&
-          supportsOpenAIExplicitPromptCaching(this.modelId),
+        cacheBreakpointOnLeadingMessage: supportsOpenAIExplicitPromptCaching(
+          this.modelId
+        ),
       }),
       temperature: this.temperature ?? undefined,
       reasoning,
--- a/front/lib/api/llm/clients/openai/index.ts
+++ b/front/lib/api/llm/clients/openai/index.ts
@@ -113,8 +113,7 @@ export class OpenAIResponsesLLM extends LLM<ResponseCreateParamsStreaming> {
     return {
       model: this.modelId,
       input: toInput(promptText, conversation, "developer", {
-        // Match Anthropic's equipped-skills breakpoint, but only for OpenAI
-        // models that accept explicit prompt cache breakpoints.
+        // Same cache breakpoint strategy as for other providers that support it.
         cacheBreakpointOnLeadingMessage: supportsOpenAIExplicitPromptCaching(
           this.modelId
         ),
--- a/front/lib/api/llm/clients/openai/index.ts
+++ b/front/lib/api/llm/clients/openai/index.ts
@@ -32,7 +32,10 @@ import {
   streamLLMEvents,
 } from "@app/lib/api/llm/utils/openai_like/responses/openai_to_events";
 import type { Authenticator } from "@app/lib/auth";
-import { supportsOpenAIExplicitPromptCaching } from "@app/lib/model_constructors/types/model_ids";
+import {
+  isModel,
+  supportsOpenAIExplicitPromptCaching,
+} from "@app/lib/model_constructors/types/models";
 import logger from "@app/logger/logger";
 import type { Result } from "@app/types/shared/result";
 import { Err, Ok } from "@app/types/shared/result";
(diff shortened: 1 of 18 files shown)


front/lib/api/llm/clients/openai/index.ts
```diff
@@ -111,7 +112,13 @@ export class OpenAIResponsesLLM extends LLM<ResponseCreateParamsStreaming> {
 
     return {
       model: this.modelId,
-      input: toInput(promptText, conversation),
+      input: toInput(promptText, conversation, "developer", {
+        // Match Anthropic's equipped-skills breakpoint, but only for OpenAI
```
flvndvd: Let's drop Anthropic reference from the comment maybe ?

front/lib/model_constructors/types/models.ts
```diff
@@ -86,6 +88,22 @@ export function isModel(value: string): value is Model {
   return (MODELS as readonly string[]).includes(value);
 }
 
+// Models older than 5.6 do not support explicit cache breakpoints.
+// https://developers.openai.com/api/docs/guides/prompt-caching#prompt-cache-breakpoints
+const OPENAI_EXPLICIT_PROMPT_CACHING_MODELS: readonly Model[] = [
+  GPT_5_6_SOL,
+  GPT_5_6_TERRA,
+  GPT_5_6_LUNA,
+];
```
pmilliotte: We want to avoid that kind of logic with the new router.

supportsOpenAIExplicitPromptCaching should be defined all the time with `prompt_cache_breakpoint: { mode: "explicit" } ` and ovewritten to () => ({})  in endpoints that don't support it

front/lib/model_constructors/sdk/openai_responses/converters/input/utils.ts
```diff
@@ -68,11 +103,18 @@ export function systemMessageToInputItem(
...
+  metadata: EndpointMetadata
 ): ResponseInputItem {
   return {
     role: "user",
-    content: [{ type: "input_text", text: message.content.value }],
+    content: [
+      {
+        type: "input_text",
+        text: message.content.value,
+        ...promptCacheBreakpointFor(message.cache, metadata),
```
pmilliotte: promptCacheBreakpointFor should be added in MessageItemConverters, and userTextMessageToInputItem should depend on it

front/lib/model_constructors/sdk/openai_responses/converters/input/utils.ts
```diff
@@ -55,6 +59,37 @@ export interface MessageItemConverters {
...
+  | Record<string, never> {
+  switch (cache) {
+    case "short":
+    case "long":
+      return supportsOpenAIExplicitPromptCaching(metadata.model)
+        // Link to the doc: https://developers.openai.com/api/docs/guides/prompt-caching
+        // We can create up to 4 new cache writes (i.e. 4 cache breakpoints).
+        // We can't control the TTL, it's set to 30min (they may change this in the future).
+        ? { prompt_cache_breakpoint: { mode: "explicit" } }
+        : {};
```
pmilliotte: no such kind of logic

front/lib/model_constructors/types/models.ts
aubin-tchoi: okok

aubin-tchoi: so IIRC I add a new converter in `WithOpenAIResponsesInputConverter` with a default, and then overwrite it in each `WithOpenAI...` (`WithOpenAIGptFiveDotFourConfig` for instance), correct?

front/lib/api/llm/utils/openai_like/responses/conversation_to_openai.ts
```diff
@@ -125,11 +125,22 @@ function toAssistantInputItem(
...
-  message: UserMessageTypeModel
+  message: UserMessageTypeModel,
+  { cacheBreakpoint }: { cacheBreakpoint: boolean }
 ): ResponseInputItem.Message {
+  // We put it on the last block to include the entire message (cache is prefix-based).
+  const lastContentIndex = message.content.length - 1;
   return {
     role: "user",
-    content: message.content.map(toInputContent),
+    content: message.content.map((content, index) => {
```
flvndvd, line 135: This feels like this predicate function has a weird contact where even though you pass `cacheBreakpoint: true` it will not always set it, can we uplift the check on the last message, so contract here remains simpler?

flvndvd: Not even sure to understand how it would work, cause reading this ternary a few lines below, we will only add it if it's the last message.

flvndvd: Oh but wait, it because the `content` field is an array within the `message` object.  Ok, then I presume ignore this CoT.

front/lib/api/llm/utils/openai_like/responses/conversation_to_openai.ts
```diff
@@ -150,18 +161,28 @@ function toToolCallOutputItem(
...
       case "user":
-        inputs.push(toUserInputMessage(message));
+        inputs.push(
+          toUserInputMessage(message, {
+            // We inject the list of equipped skills in the first user message with `name: "system"`, which is cached.
+            // Regular user messages must not receive this cross-conversation breakpoint.
+            cacheBreakpoint:
+              cacheBreakpointOnLeadingMessage &&
+              index === 0 &&
+              message.name === "system",
```
flvndvd, line 183: `nit` to ease clarity can we extract this boolean in a variable?

front/lib/api/llm/utils/openai_like/responses/test/conversation_to_openai.test.ts
```diff
@@ -15,6 +15,39 @@ describe("toInput", () => {
...
+        },
+        "developer",
+        { cacheBreakpointOnLeadingMessage: true }
+      );
+
+      expect(messages[1]).toEqual({
+        role: "user",
+        content: [
+          { type: "input_text", text: "Available" },
+          {
```
flvndvd, line 43: Can we have a test with a few more following messages to confirm it's only on the leading equiped-skills message?

flvndvd (approved):

front/lib/model_constructors/types/models.ts
pmilliotte: yes, the default being the one actually adding breakpoints (as it's likely that new models will support it)

Then`userTextMessageToInputItem` will actually depend on this "converter" (`cacheControlFor` ?).

in Anthropic we have this:

```
export function userTextMessageToTextBlock(
  message: BaseUserTextMessage
): TextBlockParam {
  return {
    type: "text",
    text: message.content.value,
    ...cacheControlFor(message.cache),
  };
}
``` 

`cacheControlFor` is not a "converter" of the class because all the models support cache breakpoints. But since only last OpenAI models do, it needs to be overwrittten. So we would have sth like:

```
export function userTextMessageToInputItem(
  message: BaseUserTextMessage
): ResponseInputItem {
  return {
    role: "user",
    content: [{ 
        type: "input_text", 
        text: message.content.value,
        ...this.cacheControlFor(message.cache)
    }],
  };
}
```

And in old models:

```
this.cacheControlFor = (message.cache) => ({})
```

It's polymorphism if we want to put a name on it

aubin-tchoi: very clear thanks, updating the PR

aubin-tchoi: updated, I did remove `userTextMessageToInputItem` from the converters as IIUC it's not a leaf anymore. not overridden anyway

pmilliotte (approved):

PR #29233, opened 2026-07-21
Add OpenAI tool search support

flvndvd:
## Description

We recently started leveraging Anthropic search tool, this PR leverages the same approach for the OAI counter part. 

Their API supports deferring tool definitions until the model discovers them through tool search. This reduces the number of tool schemas loaded into the active context for agents with large toolsets.

This follows OpenAI’s tool search guide (https://developers.openai.com/api/docs/guides/tools-tool-search).

### Implementation

This adds OpenAI tool search behind the `openai_tool_search` feature flag for models that explicitly support it. The integration covers both the legacy production router and the new router.

Non-eager tools are sent with `defer_loading: true`. Eager tools remain loaded upfront. A forced tool is also kept eager so it is available immediately when selected through tool choice. The `tool_search` tool is only added when at least one function is deferred.

Support is enabled for GPT-5.4, GPT-5.5, GPT-5.6 Sol, Terra and Luna, and GPT-5.4 Mini. GPT-5.4 Nano remains unsupported.

### Conversation round trip

OpenAI returns tool discovery as `tool_search_call`, `tool_search_output`, then the discovered `function_call`.

The two tool-search items are stored as provider passthrough content (like we did for A\) and replayed after lightweight Zod validation. Unknown fields are preserved so additions to OpenAI’s replay format do not require duplicating its complete SDK
types.

The discovered function call remains a normal Dust tool call. It sometimes provide a synthetic namespace (`default`). This namespace is preserved through output conversion, conversation persistence and input reconstruction. This is required by OpenAI when replaying calls discovered outside the default namespace. Omitting it causes a 400 response on the following model request.

>[!TIP]
> Asked confirmation to the OAI folks, but it looks like the model always sees the server/tools. So the cache is still burst when adding tools in-flight, however all deferred tools hide parameter/input schema preserving a fairly sizable amount of tokens.


front/temporal/agent_loop/lib/run_model.test.ts
```diff
@@ -522,6 +553,41 @@ describe("buildSpecificationsWithReplayPlaceholders", () => {
     ]);
   });
 
+  it("does not append placeholders for tools carried by OpenAI tool search output", () => {
```
flvndvd, line 556: Unlike A\\, OAI does not require you to always keep surfaced tools available in the specification.

flvndvd:
PMRR.

aubin-tchoi (commented):
LGTM

PR #29127, opened 2026-07-19
[front] perf: preload input bar capabilities

spolu:
## Description

Preload input-bar capabilities when the picker mounts instead of waiting for the first open.

This starts the active skills and global spaces requests immediately, then loads MCP server views as soon as the global space is available. The picker and slash menu use the same SWR keys, so both reuse the warmed cache.

The available-tools catalog is preloaded only for admins. Non-admin capability lists no longer fetch or wait for data they cannot use.

## Tests

N/A, tested locally.

## Risk

Low. This changes request timing but keeps the existing endpoints, cache keys, and capability rendering behavior.

## Deploy Plan

- deploy `front`

PR #29234, opened 2026-07-21
[front] Revert eager capability preloading

spolu:
## Description

Revert the eager capability preloading introduced in #29127 without reverting the subsequent SWR
configuration improvements.

The capabilities picker now fetches spaces, skills, and MCP server data only after the picker is
opened. The slash command menu remains an independent load trigger and shares the same SWR cache.
The deduping and focus-throttling options, admin-only available-server fetch, and setup-flow guards
are preserved.

## Tests

- `NODE_ENV=test npm test components/editor/extensions/shared/SlashCommandCapabilitiesItems.test.ts components/editor/extensions/input_bar/InputBarSlashSuggestionDropdown.test.ts`

## Risk

Low. This changes request timing back to the previous on-demand behavior while preserving current
cache configuration.

## Deploy Plan

- deploy `front`

### DIFF
--- a/front/components/assistant/CapabilitiesPicker.tsx
+++ b/front/components/assistant/CapabilitiesPicker.tsx
@@ -186,10 +186,12 @@ export function CapabilitiesPicker({
   const isMobile = useIsMobile();
   const [searchText, setSearchText] = useState("");
   const [isOpen, setIsOpen] = useState(false);
+  const [isClosing, setIsClosing] = useState(false);
   const [setupSheetServer, setSetupSheetServer] =
     useState<MCPServerType | null>(null);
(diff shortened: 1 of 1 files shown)


spolu:
cc @aubin-tchoi FYI reduce DB load
cc @Fraggle FYI as eng on-call

aubin-tchoi:
ack, will we be able to add back after optimizing the db queries?

spolu:
@aubin-tchoi yes sir! working on it 👍

spolu:
Goal was to ease the DB a bit today as we were flirting with 100% CPU

aubin-tchoi:
Nice!

PR #29235, opened 2026-07-21
[front] enh: show system prompt in Poke conversation render

aubin-tchoi:
## Description

The Poke conversation render view only exposes the rendered model conversation, which makes it harder to inspect the full input sent to the model.

This PR exposes the system prompt in the `render` endpoint (we already had it).

Also fixes the default agent selected.

image

## Tests

- Tested locally.

## Risk

- Low.

## Deploy Plan

- Deploy front and front-api.

### DIFF
--- a/front/components/poke/pages/ConversationPage.tsx
+++ b/front/components/poke/pages/ConversationPage.tsx
@@ -1071,6 +1071,13 @@ export function ConversationPage() {
               onClick={() => setUseMarkdown(!useMarkdown)}
             />
             <Button
+              label="Self-improving skills test"
+              variant="primary"
+              size="xs"
+              onClick={() => void copyTestCase()}
+              disabled={isTestCaseLoading}
+            />
+            <Button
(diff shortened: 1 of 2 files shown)


front/components/poke/pages/ConversationPage.tsx
```diff
@@ -1019,111 +1017,125 @@ export function ConversationPage() {
...
+                disabled={!conversationDataSourceId}
+              />
+              <Button
+                label={useMarkdown ? "Plain Text" : "Preview Markdown"}
+                variant="outline"
+                size="xs"
+                icon={useMarkdown ? DocumentTextIcon : CodeBracketIcon}
+                onClick={() => setUseMarkdown(!useMarkdown)}
+              />
+              <Button
```
aubin-tchoi, line 1067: reordered since the last 3 buttons go together (grouped them)

flvndvd (approved):
🙏

PR #29236, opened 2026-07-21
Enable native light reasoning for Opus 4.8

fontanierh:
## Description

Set `useNativeLightReasoning` for Opus 4.8 so configured `light` effort maps to adaptive low reasoning instead of disabled thinking.

This restores valid signed reasoning blocks after the effort resolver began honoring the configured light effort.

## Tests

Added unit coverage for native light mapping to adaptive thinking with low output effort. The focused run was blocked because the cold Hive environment has no test database.

## Risk

Low. Only Opus 4.8 calls configured with `light` effort change, from disabled thinking to adaptive low reasoning.

## Deploy Plan

Standard deploy.

### DIFF
--- a/front/lib/api/llm/clients/anthropic/utils/index.test.ts
+++ b/front/lib/api/llm/clients/anthropic/utils/index.test.ts
@@ -85,6 +85,15 @@ describe("toAutoThinkingConfig", () => {
     });
   });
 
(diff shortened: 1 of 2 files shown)


aubin-tchoi (approved):
LGTM

PR #29237, opened 2026-07-21
Report OpenAI 5.6 cache writes correctly

flvndvd:
## Description

Since OpenAI 5.6 we now need to pay a premium for cache write. Our token reports was not accounting for it. This PR changes so that it reports cache reads and writes separately in both routers.


flvndvd:
cc @theogz

pmilliotte (approved):

fontanierh (approved):
LGTM

PR #29238, opened 2026-07-21
Add list loading skeleton to skill import dialog

Inesvvv:
## Description

Added skeleton instead of spinner to skills import from repo 
Standardized the size of the popup when URL is not empty so that there isn't a blink during the skills detection

Skeleton in storybook:
Screenshot 2026-07-21 at 15 55 02

## Tests

Tested locally 

## Risk

Low

## Deploy Plan

Deploy front and sparkle


aubin-tchoi (approved):
LGTM

PR #29239, opened 2026-07-21
Denormalize group kind on group vaults (take #1)

spolu:
## Description

- add a nullable `groupKind` column to `group_vaults`
- populate it from `groups.kind` in every group-vault creation path and development seed
- document the denormalization contract beside historical group-kind migrations
- add focused assertions for member, editor, and viewer relationships

Space reads currently need the intrinsic group kind but should not need an additional join through `groups`. Storing the immutable group kind on the group-vault relationship prepares that query simplification while keeping the rollout backwards-compatible for existing rows.

`groups.kind` has no active runtime mutation path. Historical migrations that changed it are documented so future changes synchronize the denormalized value.

While this is unsatisfying to have such a denormalization, the `group_vaults` table will be soon removed by the admin governance initiative in favour of a new `group_permissions` table. Interestingly the kinds (on groups) needed for permission checking are already slated by this initiative to be moved to `group_permissions`. It was a hack to have the `agent_editors`, `skill_editors`, ... kinds on `groups` which is also the root cause for needing `group.kind` in `SpaceResource`.

## Tests

- `NODE_ENV=test npm test -- lib/resources/group_space_resource.test.ts lib/dev/dust_hive_seed_schema.test.ts`
- `NODE_ENV=test npm test -- lib/resources/space_resource.test.ts lib/api/spaces.test.ts`

## Risks

- Low simple column addition and new writes to it.

## Deploy Plan

- `npm run migration:apply:pre-deploy`
- deploy `front`

### DIFF
--- a/front/lib/api/spaces.ts
+++ b/front/lib/api/spaces.ts
@@ -584,6 +584,7 @@ export async function createSpaceAndGroup(
         {
           kind: space.isProject() ? "project_viewer" : "member",
           groupId: globalGroup.id,
+          groupKind: globalGroup.kind,
           vaultId: space.id,
           workspaceId: owner.id,
         },
--- a/front/lib/dev/byok_seed.sql
+++ b/front/lib/dev/byok_seed.sql
@@ -124,8 +124,8 @@ inserted_conversations_space AS (
 
 -- Step 6a: Link system group → system space
 link_system AS (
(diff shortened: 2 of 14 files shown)


spolu:
r? @id13 
cc @Fraggle as eng on-call
cc @flvndvd as space/group owner
cc @philipperolet

id13 (approved):
LGTM

Fraggle (approved):
LGTM too

PR #29241, opened 2026-07-21
Add Grok 4.5 as the default xAI model

fontanierh:
## Description

- Add Grok 4.5 to the model registry, model picker, xAI router, pricing tables, and SDK type.
- Prefer Grok 4.5 in both large and small fallback selection used by global agents, with feature flags propagated through every global-agent fallback.
- Keep Grok 4 available for existing agents while marking it legacy.
- Default omitted Grok 4.5 reasoning from the canonical model configuration, matching the model default and avoiding the unsupported none effort.
- Apply xAI's higher input, cached input, and output rates when prompts exceed 200k tokens.

## Tests

- Live xAI integration matrix for reasoning, temperature, multi-turn conversations, tools, vision, and structured output.
- Focused Vitest coverage for global-agent xAI selection, long-context pricing, model preference, pricing tiers, and xAI parameter defaults.
- `npx tsgo --noEmit` in `front`.
- `npm run tsgo -- --noEmit` in `marketing`.
- `npm run build` in `sdks/js`.

## Risk

Low. Grok 4.5 remains behind the existing xAI feature flag, and the prior Grok model stays registered for existing agents. Long-context pricing is selected from recorded prompt usage. The change can be rolled back without a data migration.

## Deploy Plan

Standard deployment. No migration or manual step is required.


front/lib/api/llm/clients/xai/types.ts
```diff
@@ -2,20 +2,34 @@ import type {
...
 
 export const XAI_MODEL_CONFIGS: Record<
   XaiWhitelistedModelId,
   {
+    defaultReasoningEffort?: ReasoningEffort;
     overwrites: LLMParameterOverwrites;
   }
 > = {
+  [GROK_4_5_MODEL_ID]: {
+    defaultReasoningEffort: "high",
```
pmilliotte: already in model config no ?

pmilliotte (approved):

front/lib/api/llm/clients/xai/types.ts
fontanierh: Yes. The fallback is needed because the base client turns an omitted effort into `none`, which Grok 4.5 rejects, but duplicating `high` here could drift. Updated it to read `GROK_4_5_MODEL_CONFIG.defaultReasoningEffort` directly while preserving explicit efforts and the legacy Grok behavior. Thanks!

PR #29244, opened 2026-07-21
feat(activation): automate workspace schedule lifecycle from pod provisioning

ElPicador:
## Description

Replaces the hand-created pilot schedule with an automated lifecycle for the per-workspace Activation nudge schedule:

- Provisioning a workspace's first Activation Pod (via the `join-activation-pod` poke plugin) now starts that workspace's schedule directly.
- A nightly reconcile job (`ensureActivationWorkspaceSchedules`, mirroring reinforcement's ensure-schedules pattern) creates schedules for any workspace with a live pod that's missing one, and deletes schedules for workspaces with no live pods left.
- Removes the old feature-flag-gated schedule creation at workspace-creation time (`activation_scheduler` flag in `lib/iam/workspaces.ts`), since provisioning a pod now defines scope instead.

The partial index on `project_metadata.provisioningSource` needed for the reconcile query already shipped separately, so no migration is included here.

## Tests

Added coverage for the new reconcile query (`ProjectMetadataResource.listWorkspaceModelIdsWithLiveActivationPods`) and for the pod-provisioning hook (`join_activation_pod.test.ts`), including the schedule-creation failure path.

### DIFF
--- a/front/lib/api/poke/plugins/spaces/join_activation_pod.test.ts
+++ b/front/lib/api/poke/plugins/spaces/join_activation_pod.test.ts
@@ -1,0 +1,129 @@
+import { joinActivationPodPlugin } from "@app/lib/api/poke/plugins/spaces/join_activation_pod";
+import { Authenticator } from "@app/lib/auth";
+import { SpaceResource } from "@app/lib/resources/space_resource";
+import { makeSId } from "@app/lib/resources/string_ids";
+import { GroupFactory } from "@app/tests/utils/GroupFactory";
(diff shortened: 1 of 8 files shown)


ElPicador:

- #29031
    - #29033
        - #29059
            - #29065
                - #29067
                    - #29103
                        - #29105
                            - #29117
                                - #29226
                                    - #29244 ◀
                                        - #29260

frankaloia (approved):
clean 🙏

PR #29248, opened 2026-07-21
Backfill group kind on group vaults

spolu:
## Description

- add a TypeScript migration that backfills `group_vaults.groupKind` from `groups.kind`
- process rows with ID-based keyset pagination and NULL-only updates
- update at most 1,000 rows per autocommit transaction by default
- pause 100 ms between batches by default, with configurable `--batchSize` and `--batchDelayMs`
- preserve `updatedAt` and make reruns idempotent

This follows [#29239](https://github.com/dust-tt/dust/pull/29239), which added the nullable denormalized column and populated it in all writers.

The US region has roughly one million `group_vaults` records and the table is under load, so a single SQL `UPDATE` would create an unnecessarily large transaction and row-lock/WAL footprint. The TypeScript migration bounds that pressure and releases locks between batches.

## Tests

- `npx biome check migrations/20260721_backfill_group_vault_group_kind.ts`
- `npx tsx migrations/20260721_backfill_group_vault_group_kind.ts --batchSize 2 --batchDelayMs 0`
- `npx tsx migrations/20260721_backfill_group_vault_group_kind.ts --execute --batchSize 2 --batchDelayMs 0`
- verified three local rows were updated across two batches with zero NULL or mismatched kinds
- verified an immediate rerun updated zero rows

## Risks

- Medium. This backfills roughly one million rows in the US region on a table under load. Keyset pagination, 1,000-row autocommit transactions, a 100 ms inter-batch delay, and NULL-only idempotent updates bound the impact. Batch size and delay are configurable if database load requires further throttling.

## Deploy Plan

- pull `main` from prodbox
- dry-run the migration in each region:
  `npx tsx migrations/20260721_backfill_group_vault_group_kind.ts`
- execute the migration in each region while monitoring database load:
  `npx tsx migrations/20260721_backfill_group_vault_group_kind.ts --execute`
- rerun the dry-run and confirm it reports zero remaining rows

### DIFF
--- a/front/migrations/20260721_backfill_group_vault_group_kind.ts
+++ b/front/migrations/20260721_backfill_group_vault_group_kind.ts
@@ -1,0 +1,117 @@
+import { QueryTypes } from "sequelize";
+
+import { frontSequelize } from "@app/lib/resources/storage";
+import { makeScript } from "@app/scripts/helpers";
+
+const DEFAULT_BATCH_SIZE = 1_000;
+const DEFAULT_BATCH_DELAY_MS = 100;
+
+const UPDATE_BATCH_SQL = `
+  UPDATE group_vaults AS group_vault
+  SET "groupKind" = group_row.kind
+  FROM groups AS group_row
+  WHERE group_vault.id IN (:ids)
+    AND group_vault."groupId" = group_row.id
+    AND group_vault."workspaceId" = group_row."workspaceId"
+    AND group_vault."groupKind" IS NULL
+`;
+
+function wait(durationMs: number): Promise<void> {
+  return new Promise((resolve) => setTimeout(resolve, durationMs));
+}
+
+makeScript(
+  {
+    batchSize: {
+      type: "number",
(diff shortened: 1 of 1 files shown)


spolu:
PMRR

front/migrations/20260721_backfill_group_vault_group_kind.ts
```diff
@@ -0,0 +1,117 @@
...
+
+    for (;;) {
+      // Keyset pagination prevents each batch from rescanning rows already processed.
+      const rows = await frontSequelize.query<{ id: number }>(
+        `
+          SELECT group_vault.id
+          FROM group_vaults AS group_vault
+          INNER JOIN groups AS group_row
+            ON group_vault."groupId" = group_row.id
+           AND group_vault."workspaceId" = group_row."workspaceId"
```
id13, line 76: why the join with groups ? 🤔

spolu:
Migration complete

PR #29250, opened 2026-07-21
[LLM] Fix Anthropic temperature=1 rejection when thinking is off

pmilliotte:
## Description

Anthropic rejects an explicit `temperature=1` unless thinking is enabled/adaptive, so drop that value when reasoning is off (Opus 4.6, Sonnet 4.6, Haiku 4.5) and refactor endpoint `parseConfig` into an ordered `configParsers` array applied before schema validation.

## Tests

Tested locally.

## Risk

Low.

## Deploy Plan

- Deploy `main` on `front`


Nils-Fedrigo (approved):
thanks!

Issue #9464, opened 2024-12-17
Folders: mimetype should be variable not constant


philipperolet:
Mimetype of folders allow us to differentiate objects, which will be necessary for the UX display when we move content node api to core

For instance, slack channel & google drive are both "folders" but do not have a standard folder icon and might require specific handling (at least in UI, where they have different icons)

PR #29252, opened 2026-07-21
Don't set capability to 'builders' group if it does not exist yet

tdraier:
## Description

Two related fixes to the governance capability seeding introduced in the Admin Governance stack (dust-tt/tasks#9463, #9464):

**1. Never fabricate the Builders group.** Previously, a `"builders"` target with no existing `MANUAL_BUILDERS_GROUP_NAME` group (no builder-role member ever synced into it) would create the group empty just to grant it. That's now centralized in a new `resolveEffectiveTarget(auth, target)` in `governance_seeding.ts`: it degrades `"builders"` down to `"admins_only"` when the group doesn't exist, and `applyCapabilityTarget` uses it internally — so this applies uniformly to **every** caller (`seedWorkspaceCapabilities` at workspace creation, and the backfill script), not just the backfill. `ApplyCapabilityOutcome` gains `"skipped_no_builders_group"` to distinguish this case from a target that resolved to `admins_only` directly.

**2. The backfill script now actively reverts stale grants.** Previously it only ever skipped a workspace/capability pair that was already configured (protecting a manual override), even if the *newly resolved* target was `admins_only` — e.g. a stale `"everyone"` grant from an earlier run would never be corrected. Now, when the resolved effective target is `admins_only` and the capability is currently configured to something else, the script calls `GroupPermissionResource.disable(...)` to revert it. The "preserve existing configuration" idempotency guard still applies for `"everyone"`/`"builders"` targets — only the `admins_only` case is now actively enforced. New counter: `revertedToAdminsOnly`.

## Tests

- `governance_seeding.test.ts`: updated the two tests that previously asserted the Builders group got created during seeding — they now assert the capability stays `admins_only` and no group is created; added a case confirming seeding still grants correctly when the group *already* exists.
- Verified end-to-end against the test DB with a throwaway smoke test (not committed — no test convention for migration scripts in this repo): everybody-grant, builders-grant, skip-when-no-group, revert-when-previously-configured (including dry-run not writing), and idempotent re-run.
- `npx tsgo --noEmit` and `biome check` clean.

## Risk

Low. Tightens existing behavior toward "don't fabricate state" and "keep the backfill's output in sync with resolved targets" — no new capabilities or call sites introduced.

## Deploy Plan

Standard deploy. Safe to re-run the backfill script after this deploys; it will now revert any workspace whose earlier backfill run granted `"builders"` via a group that has since disappeared (or never should have been created).

### DIFF
--- a/front/lib/api/permissions/governance_seeding.test.ts
+++ b/front/lib/api/permissions/governance_seeding.test.ts
@@ -25,7 +25,7 @@ describe("seedWorkspaceCapabilities", () => {
     expect(await userAuth.hasWorkspacePermission("create", "agent")).toBe(true);
   });
 
-  it("creates the Builders group and grants it when disallow_agent_creation_to_users is already set", async () => {
+  it("leaves create-agent as admins_only when disallow_agent_creation_to_users is set and no Builders group exists yet", async () => {
     const workspace = await WorkspaceFactory.basic();
     await GroupFactory.defaults(workspace);
     await FeatureFlagResource.enable(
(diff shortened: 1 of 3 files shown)


tdraier:

- #29252 ◀

matteotrab (approved):

PR #29255, opened 2026-07-21
[Analytics] Prevent user exports from exceeding bucket limit

philipperolet:
## Description

Fixes https://github.com/dust-tt/tasks/issues/9387

The users export created an empty daily Elasticsearch bucket for every inactive day between each user's first and last message. Large workspaces could therefore exceed Elasticsearch's bucket limit even though those empty buckets were discarded when building the export.

Only return days that contain messages. This preserves `activeDaysCount` while preventing empty buckets from exhausting the query limit.

## Risks

Blast radius: users analytics exports
Risk: low

## Deploy Plan

- deploy front


Nils-Fedrigo (commented):
ty!

PR #28395, opened 2026-07-02
front: skip space permission filtering when rendering historical messages

ElPicador:
## Description

Conversation message rendering (`batchRenderMessages`) was filtering out agent configurations referenced by user mentions and agent messages based on the viewer's current space access (`filterAgentsByRequestedSpaces`). This meant an agent mentioned in an old message, or a message sent by an agent, could disappear or error out of the conversation history if the viewer's access to the agent's space changed after the fact (e.g. removed from a space, agent moved spaces).

Adds a `dangerouslySkipPermissionFiltering` option to `getAgentConfigurations`, `getAgentConfigurationsWithVersion`, and `getAgentConfiguration` (mirroring the existing pattern in `configuration/views.ts`), and uses it when resolving agent configurations for both user and agent message rendering in `messages.ts`. Historical messages now render consistently regardless of the viewer's current space access.

## Tests

Verified with `npx tsgo --noEmit`.

PR #29256, opened 2026-07-21
Remove useless extra check when fetching messages by ids.

Fraggle:
## Description

Removes the redundant `checkAgentMessageAccess` authorization check inside `AgentStepContentResource.fetchByAgentMessageIds`. The check was duplicating access control that callers already enforce upstream, and unnecessarily gating the early-return path on an async DB call. The method now short-circuits on an empty `agentMessageIds` array directly and passes the original IDs to `chunk`.

## Tests

Tested manually. No logic change for authorized callers — existing behavior is preserved.

## Risk

Low. The removed check was redundant: all call sites enforce access before invoking this method. No data is exposed that wasn't accessible before.

## Deploy Plan

Deploy front.

### DIFF
--- a/front/lib/resources/agent_step_content_resource.ts
+++ b/front/lib/resources/agent_step_content_resource.ts
@@ -173,20 +173,11 @@ export class AgentStepContentResource extends BaseResource<AgentStepContentModel
   ): Promise<AgentStepContentResource[]> {
     const owner = auth.getNonNullableWorkspace();
 
-    // Check authorization - will throw if unauthorized
-    const allowedAgentMessageIds = await this.checkAgentMessageAccess(
-      auth,
-      agentMessageIds
-    );
-
-    if (allowedAgentMessageIds.length === 0) {
+    if (agentMessageIds.length === 0) {
       return [];
     }
 
-    const chunks = chunk(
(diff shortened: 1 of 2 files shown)


Fraggle:

- #29256 ◀

front/lib/resources/agent_step_content_resource.ts
```diff
@@ -173,20 +173,11 @@ export class AgentStepContentResource extends BaseResource<AgentStepContentModel
   ): Promise<AgentStepContentResource[]> {
     const owner = auth.getNonNullableWorkspace();
 
-    // Check authorization - will throw if unauthorized
-    const allowedAgentMessageIds = await this.checkAgentMessageAccess(
```
Fraggle, line 177: The function do not prevent access but only remove messages that you may no longer have access to the CURRENT agent configuration.
This is actually opposed to the design doc and expectations.

Fraggle: Same as https://github.com/dust-tt/dust/pull/28395

id13 (approved):
Trusting you on the `    const allowedAgentMessageIds = await this.checkAgentMessageAccess(` removal

PR #29258, opened 2026-07-21
Skills: Fix similar skills detection for more than 100 skills

fabiencelier:
## Description

closes https://github.com/dust-tt/tasks/issues/9714

the similar skill detection was limited to only 100 oldest skills so not very good for workspaces with 500+ skills like the dust workspace.
This run the similar skills detection by batch of 100 skills, with several parallel LLM calls.

We can tweak the value later, but 100 skills seems reasonable:
100 skills * 500 characters / 4 char per token ~= 12.5K tokens 

## Tests

tested locally but not with 100+ skills
Added unit test to make sure 101 skills do trigger 2 calls

## Risk

low, easy to revert
we increase cost of the feature as the number of LLM calls is now proportional to number of skills.
this should stay reasonable and hopefully, the number of skills to search will be lower when we have unpublished skills

## Deploy Plan

deploy front


aubin-tchoi (approved):
LGTM

PR #29259, opened 2026-07-21
[Tables] Keep exponent-like text values (e.g. 32E0134) as text in CSV parsing

philipperolet:
## Description

Fixes https://github.com/dust-tt/tasks/issues/9285

Table rows are persisted as a CSV file in GCS, and every read re-infers each value's type in `Row::from_csv_record`. Rust's float parser accepts exponent notation without a decimal separator, so TEXT values shaped like `[digits]E[digits]` (e.g. a product reference `32E0134`) were silently parsed as floats (`3.2e135`) — even when upserted as proper JSON strings through the rows API, since those are also serialized to CSV before the background worker re-parses them. When rows were re-written to storage (background consolidation, non-truncate merge, row deletion), the float was serialized via Rust's `Display`, which never emits exponent notation, so the stored CSV ended up with the 136-digit numeral. Query results are served straight from that CSV (SQLite csvtab), hence the corrupted output and `WHERE reference = '32E0134'` matching nothing.

Fix: in CSV value type inference, only treat exponent notation as numeric when a decimal separator is present. `3.2E5` still parses as a float; `32E0134` (and `1e5`) now stay text. This is round-trip safe because our own CSV serialization never emits exponent notation.

Note: already-corrupted stored values cannot be recovered (the original string is lost); affected tables need a truncate re-upsert to be repaired.

## Risks

Blast radius: type inference of table values (CSV uploads and rows-API upserts) for structured data queries.
Risk: low

Behavior change edge case: a column containing genuine scientific notation without a decimal separator (e.g. `1e5`) is now inferred as Text instead of Float. Preserving user strings is the safer default there.

## Deploy Plan

- deploy core (front-edge/core-edge first is fine, no migration)

### DIFF
--- a/core/src/databases/table.rs
+++ b/core/src/databases/table.rs
@@ -1332,10 +1332,7 @@ mod tests {
         let row = Row::from_csv_record(headers.clone(), record, "0".to_string())?;
         assert_eq!(row.content()["a"], Value::String("32E0134".to_string()));
         assert_eq!(row.content()["b"], Value::String("14e0050".to_string()));
-        assert_eq!(
-            row.content()["c"],
-            Value::Number(serde_json::Number::from_f64(320000.0).unwrap())
-        );
+        assert_eq!(row.content()["c"], Value::from(320000.0));
 
         let headers = Arc::new(vec!["a".to_string(), "b".to_string()]);
         let record = vec!["true", "false"];
--- a/core/src/databases/table.rs
+++ b/core/src/databases/table.rs
@@ -945,6 +945,13 @@ impl Row {
         let mut values = Vec::<serde_json::Value>::new();
 
         fn try_parse_float(s: &str) -> Result<serde_json::Number> {
+            // Rust's float parser accepts exponent notation without a decimal separator,
+            // coercing product references like "32E0134" into numbers. Only treat exponent
+            // notation as numeric when a decimal separator is present (e.g. "3.2E5").
+            // Round-trip safe: to_csv_record writes floats via Display, never as exponents.
+            if !s.contains('.') && (s.contains('e') || s.contains('E')) {
+                return Err(anyhow!("Exponent notation without decimal separator"));
+            }
             if let Ok(float) = s.parse::<f64>() {
                 match serde_json::Number::from_f64(float) {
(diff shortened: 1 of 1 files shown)


philipperolet:
## Note on mixed columns (e.g. `3.2E20` + `3E10` in the same column)

This fix changes behavior for columns mixing dotted and undotted exponent notation: before, both parsed as floats → column inferred `Float`; now `3E10` stays text → the column merges to `Text`.

A column-level rule was considered ("any dotted-exponent value in the column → coerce all exponent values to numeric; none dotted → all text"), but rejected:

1. **The parser never sees "the column".** Type inference is value-by-value in a single streaming pass, and rows arrive in batches: each `/rows` API batch is written as its own pending CSV and parsed file-by-file by the background worker before merging. A column-level vote would be computed over whichever slice of rows is in each file, so the same value could be coerced differently depending on the batch it arrived in, and could flip across merges. Making it deterministic requires a re-coercion pass over the fully merged rows on every write — much bigger and riskier than this guard.

2. **The two mistakes aren't symmetric.** Text→numeric coercion is destructive and silent (that's the bug: `32E0134` is gone forever). Numeric→text is mild and recoverable: all digits are preserved, readable and filterable. Under ambiguity — and the mixed column is the ambiguous case — the tie-break should go to the non-destructive side. "Mixed → numeric" would let one dotted-looking value in a references column drag every `32E0134`-style ref back into corruption.

3. **The mixed case is rare and self-limiting.** Machine-generated scientific exports format consistently (`3.0E10`, not `3E10`), and our own storage serialization never emits exponent notation (floats are written in full decimal digits), so the question only exists on first parse of a user-supplied CSV and disappears after one round-trip.

Longer-term, the more robust direction is schema-aware parsing of the storage CSV: every re-read currently re-infers types from text and ignores the declared column type, which is the root of a wider bug family (`00123` → `123` loses leading zeros, `+33612345678` → int loses the `+`, `TRUE` → bool). Parsing reads against the stored schema would make inference a one-time decision at first upsert and eliminate round-trip drift generally — left as a potential follow-up.

fontanierh (approved):

PR #29261, opened 2026-07-21
[zendesk] Add personal account support to Zendesk tool

anya-dust:
Enable personal_actions for the Zendesk MCP server so each user can connect their own Zendesk account, alongside the existing workspace (platform_actions) connection.

Mirrors the Jira provider pattern: personal connections inherit the Zendesk subdomain from the admin's workspace connection via getUpdatedExtraConfig (keyed by mcp_server_id), so users never re-enter it. Sets requiresWorkspaceConnectionForPersonalAuth since the subdomain is mandatory. isExtraConfigValid accepts an mcp_server_id for personal setup. Adds unit tests covering config validation and subdomain inheritance.

## Description

Ask from EQT. 

## Tests

Added Unit tests for zendesk.

### DIFF
--- a/front/lib/api/oauth/providers/zendesk.ts
+++ b/front/lib/api/oauth/providers/zendesk.ts
@@ -127,12 +127,18 @@ export class ZendeskOAuthProvider implements BaseOAuthStrategyProvider {
           );
         }
         const connection = connectionRes.value.connection;
+        const { zendesk_subdomain } = connection.metadata;
+
+        if (!zendesk_subdomain) {
+          throw new Error(
+            "Zendesk workspace connection is missing a subdomain; " +
+              "cannot set up a personal connection."
+          );
+        }
 
         return {
           ...restConfig,
-          ...(connection.metadata.zendesk_subdomain && {
-            zendesk_subdomain: connection.metadata.zendesk_subdomain,
-          }),
+          zendesk_subdomain,
         };
       }
       case "connection":
--- a/front/lib/api/oauth/providers/zendesk.ts
+++ b/front/lib/api/oauth/providers/zendesk.ts
@@ -1,18 +1,26 @@
 import { isValidZendeskSubdomain } from "@app/lib/api/actions/servers/zendesk/types";
 import config from "@app/lib/api/config";
+import { getWorkspaceOAuthConnectionIdForMCPServer } from "@app/lib/api/oauth/mcp_server_connection_auth";
 import type { BaseOAuthStrategyProvider } from "@app/lib/api/oauth/providers/base_oauth_stragegy_provider";
 import {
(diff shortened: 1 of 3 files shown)


front/lib/api/oauth/providers/zendesk.ts
```diff
@@ -55,10 +63,62 @@ export class ZendeskOAuthProvider implements BaseOAuthStrategyProvider {
...
+    auth: Authenticator,
+    {
+      extraConfig,
+      useCase,
+    }: {
+      extraConfig: ExtraConfigType;
+      useCase: OAuthUseCase;
+    }
+  ): Promise<ExtraConfigType> {
+    if (useCase === "personal_actions") {
```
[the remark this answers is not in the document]
anya-dust: fixed in new commit

front/lib/api/oauth/providers/zendesk.ts
```diff
@@ -55,10 +63,62 @@ export class ZendeskOAuthProvider implements BaseOAuthStrategyProvider {
     return getStringFromQuery(query, "state");
   }
 
-  isExtraConfigValid(extraConfig: ExtraConfigType) {
+  isExtraConfigValid(extraConfig: ExtraConfigType, useCase: OAuthUseCase) {
+    if (useCase === "personal_actions") {
```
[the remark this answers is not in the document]
anya-dust: fixed in new commit

front/lib/api/oauth/providers/zendesk.ts
```diff
@@ -55,10 +64,85 @@ export class ZendeskOAuthProvider implements BaseOAuthStrategyProvider {
...
+        if (connectionRes.isErr()) {
+          throw new Error(
+            "Failed to get connection metadata: " + connectionRes.error.message
+          );
+        }
+        const connection = connectionRes.value.connection;
+
+        return {
+          ...restConfig,
+          ...(connection.metadata.zendesk_subdomain && {
```
smb2268: I think we need to throw an error here if they're using personal auth but this subdomain is missing, otherwise we'll get an unhandled error right?

anya-dust: you can't get to that step without adding a subdomain theoretically. I can make that change though!: 
image

smb2268: let's guard it anyway to be safe, ideally the code should be self documenting so we prevent people / agents from changing the behavior down the line, and this makes it look optional. otherwise lgtm and safe to merge once that change is made!

anya-dust: updated with the fix

smb2268 (approved):
LGTM!!

PR #29264, opened 2026-07-21
Make group vault group kind not null

spolu:
## Description

- make `GroupSpaceModel.groupKind` non-nullable in TypeScript
- set `allowNull: false` in the Sequelize model
- add a post-deploy migration that makes `group_vaults.groupKind` `NOT NULL`
- validate a temporary `CHECK` constraint before setting `NOT NULL`, then remove it

This completes the rollout started by [#29239](https://github.com/dust-tt/dust/pull/29239), which added the nullable denormalized column and populated it in all writers, and [#29248](https://github.com/dust-tt/dust/pull/29248), which added the batched regional backfill.

The temporary `CHECK NOT VALID` pattern lets PostgreSQL scan the roughly one-million-row table without holding an access-exclusive lock. Once validated, PostgreSQL can set `NOT NULL` with only a brief metadata lock instead of rescanning the table under that stronger lock.

## Tests

- `npx biome check lib/resources/storage/models/group_spaces.ts`
- `NODE_ENV=test npm test -- lib/resources/group_space_resource.test.ts lib/resources/space_resource.test.ts lib/api/spaces.test.ts lib/dev/dust_hive_seed_schema.test.ts`
- tested locally

## Risks

- Low. DB looks good so this should be a breeze

## Deploy Plan

- apply the post-deploy migration: `npm run migration:apply:post-deploy`

### DIFF
--- a/front/lib/resources/storage/models/group_spaces.ts
+++ b/front/lib/resources/storage/models/group_spaces.ts
@@ -11,7 +11,7 @@ export class GroupSpaceModel extends WorkspaceAwareModel<GroupSpaceModel> {
   declare createdAt: CreationOptional<Date>;
   declare groupId: ForeignKey<GroupModel["id"]>;
   // Denormalized from groups.kind. Keep in sync if group kinds ever become mutable.
-  declare groupKind: GroupKind | null;
+  declare groupKind: GroupKind;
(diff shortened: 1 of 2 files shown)


spolu:
PMRR

spolu:
migration applied

id13:
LGTM

PR #29267, opened 2026-07-21
Roll out checkpointed conversation windows globally

flvndvd:
## Description

We have been testing a less agressive pruning internally for the past few days, we solely focuses around pruning tool results and eventually escalating to compaction. Even though we have no automated compaction yet, and it adds some user friction it still better on two fronts:
1. we are not bursting cache on each turn when context window reaches a hard context window limit
2. we are not dropping interactions anymore, prevent quality-loss on the subsequent turns

This PR removes the conversation-window feature flag and makes checkpoint-buffered tool-result pruning the single rendering policy across agent runs, batch inference, reinforcement, and embedded conversation history.


aubin-tchoi (approved):
💪

PR #29268, opened 2026-07-21
[front] perf: Exclude unbounded remote MCP server columns from the hot fetch path

id13:
## Problem

The `SELECT * ... WHERE id IN (...) AND workspaceId = ?` on `remote_mcp_servers` ([pganalyze](https://app.pganalyze.com/databases/-642267661/queries/12424550236)) is the **#1 query on the front DB**: 10.6% of total DB time, ~1.26M calls/day, avg 16.3ms with ~0 I/O (all shared-buffer hits). The time is dominated by Postgres detoasting + serializing the unbounded columns (`cachedTools` chiefly) on every call — invisible in EXPLAIN: a local repro with realistic 50-tool payloads runs in 0.2ms under `EXPLAIN ANALYZE` but ~20ms for real.

Per Datadog, 81% of the call volume is the agent loop (`front-agent-loop-worker`, avg 63ms/query span, p95 180ms), which fetches every remote server behind a workspace's MCP server views on each agent-config render — and consumes only name/description/icon/meta from them.

## Change

Unbounded columns (`authorization`, `cachedTools`, `customHeaders`, `lastError`, `sharedSecret` — the model's `DANGEROUSLY_UNBOUNDED_TEXT`/JSONB fields) move off `RemoteMCPServerResource`'s type surface into a JIT-fetched payload. **Light is the default**: every fetch skips the payload columns unless the call site opts in.

- `RemoteMCPServerResource` fetchers take `includePayload` (default `false`); `getPayload()` asserts, `fetchPayload`/`fetchServerPayloads` (batch, one query) load JIT; `toJSON()` reads through `getPayload()` so serializing a light-fetched server fails loudly instead of silently emitting empty tools.
- `MCPServerViewResource` threads `includeRemoteServerPayload` (default `false`) through `baseFetch` and all `fetchBy*`/`listBy*` wrappers, and adds `getServerDisplayMetadata()` (name/description/icon/meta without payload) plus `fetchRemoteServerPayloads()` for late JIT loads.
- All ~45 call sites of both resources audited and classified:
  - **Light (the hot paths)**: `fetchMCPServerActionConfigurations` (dominant caller, now on `getServerDisplayMetadata`), tool-listing pre-fetch + server-name disambiguation + `connectServerSideMCP` (per-tool-execution!) in `mcp_actions.ts`, name-conflict checks, YAML converter name resolution, deletion/scrub paths, onboarding, `resolveNamesBySIds`.
  - **Opt-in `true`** (serialize tools/secrets or connect): UI/API listing endpoints (`lib/api/mcp/servers|views`), agent builder SSR props, tool search (BM25 over tools), toolsets/JIT paths (`isJITMCPServerView` inspects tool schemas), skills (serialize views), blocked-action listing (authorization), OAuth/connect machinery (`mcp_metadata.ts`), cached-tools fallback, poke.

Deliberately kept in the base select: `cachedDescription` and `meta` — technically unbounded but tiny in practice (avg 64B / 99.95% null), and the hottest consumer needs them for prompt construction; trimming them would force a second query on every agent-config render.

## Testing

- `npx tsgo --noEmit` clean.
- 228 tests pass across `remote_mcp_servers_resource`, `mcp_server_view_resource`, `mcp_actions`, `input_configuration`, `jit_actions`, `skill_actions`, `skill_resource`, `agent_mcp_action_resource`, `run_tool`. The `getPayload()` assert caught every missed opt-in during development (incl. a sidekick tool-validation path), which is the intended failure mode for future regressions: loud, not silent.
- Post-deploy verification: the agent-loop query fingerprint in pganalyze shifts to a SELECT without the five payload columns; query 12424550236 calls/min and avg time should collapse.

## Non-goals / follow-ups

- Table bloat (5.5k rows in 144MB) and the BitmapAnd plan remain for the residual full fetches; revisit only if the table stays visible in pganalyze after this lands.
- Dropping the redundant single-column `remote_mcp_servers_workspace_id` index (unused since 2026-04-27).

### DIFF
--- a/front/migrations/post-deploy/20260721191500_cluster_remote_mcp_servers.sql
+++ b/front/migrations/post-deploy/20260721191500_cluster_remote_mcp_servers.sql
@@ -1,0 +1,41 @@
+-- remote_mcp_servers holds ~5.5k live rows in ~140MB (~0.3 rows/page) because the tool-sync
+-- loop rewrites rows continuously and the file never shrinks. The sparse, uncorrelated heap
+-- makes the planner pick a BitmapAnd plan (~315 buffer hits/call) for the hot
+-- `id IN (...) AND workspaceId = ?` query (#1 query on front DB, ~10% of total DB time).
+-- Rewriting the table clustered by (workspaceId, id) restores compactness and correlation,
+-- which flips the plan to a plain workspace-index scan (~15 buffer hits/call).
+-- fillfactor 90 keeps in-page slack so the sync loop's updates stay HOT and the layout lasts.
+-- CLUSTER takes an ACCESS EXCLUSIVE lock; the rewrite is a few seconds on this table size and
+-- lock_timeout aborts the attempt if the lock cannot be acquired quickly.
+SET SESSION statement_timeout = 1200000;
+SET SESSION lock_timeout = 3000;
+
+ALTER TABLE remote_mcp_servers SET (fillfactor = 90);
+
+-- The composite (workspaceId, id) index is named remote_mcp_server_workspace_id_id on
+-- databases that predate the squashed baseline migration and
+-- remote_mcp_servers_workspace_id_id on databases created from it, so resolve the name at
+-- runtime. CLUSTER cannot run inside a transaction block (hence no DO block): \gexec runs the
+-- generated statement directly in this autocommit psql session.
+DO $$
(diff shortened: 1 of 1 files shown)


id13:
Complete garbage

PR #29269, opened 2026-07-21
[front] perf: Exclude unbounded remote MCP server columns from base fetch

id13:
## Description

Currently, we fetch remote_mcp_servers quite a lot, and have a nasty, unbounded JSONB column that makes 93% of the data (of course toasted).
Because of that, fetching a full row takes 3.6 pages on average, and while this is not a lot of data and the whole table fits 100% in shared buffers, detoasting thousands of times per minute is costly.

To stay lean, and because we really don't need this data in most cases, we don't return it anymore by default and codepath needing it now need to fetch it explicitly.

As a result, this query should drop drastically in usage: 

```
SELECT "createdAt", "updatedAt", "url", "icon", "version", "cachedName", "cachedDescription", "cachedTools", "lastSyncAt", "lastError", "sharedSecret", "authorization", "customHeaders", "meta", "workspaceId", "id" 
  FROM "remote_mcp_servers" AS "remote_mcp_server" 
 WHERE "remote_mcp_server"."id" IN ($0) AND "remote_mcp_server"."workspaceId" = $15
```

## Tests

Unit tests 

Will also front-edge it.

## Risk

Standard
Blast radius: remote mcp servers codepaths

## Deploy Plan

- Deploy front

### DIFF
--- a/front-api/routes/poke/workspaces/[wId]/mcp/views/index.ts
+++ b/front-api/routes/poke/workspaces/[wId]/mcp/views/index.ts
@@ -24,15 +24,40 @@ app.get(
 
     let mcpServerViews: MCPServerViewResource[];
     if (systemSpaceOnly === "true") {
-      mcpServerViews = await MCPServerViewResource.listForSystemSpace(auth);
+      mcpServerViews = await MCPServerViewResource.listForSystemSpace(auth, {
+        includeHeavyAttributes: [
+          "authorization",
+          "cachedTools",
+          "customHeaders",
+          "lastError",
+          "sharedSecret",
+        ],
+      });
     } else if (globalSpaceOnly === "true") {
       const globalSpace = await SpaceResource.fetchWorkspaceGlobalSpace(auth);
       mcpServerViews = await MCPServerViewResource.listBySpace(
         auth,
-        globalSpace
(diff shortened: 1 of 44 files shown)


front-api/routes/poke/workspaces/[wId]/mcp/views/index.ts
```diff
@@ -24,15 +24,40 @@ app.get(
...
+      mcpServerViews = await MCPServerViewResource.listByWorkspace(auth, {
+        includeHeavyAttributes: [
+          "authorization",
+          "cachedTools",
+          "customHeaders",
+          "lastError",
+          "sharedSecret",
+        ],
+      });
     }
```
Comment (id13) [inline front-api/routes/poke/workspaces/[wId]/mcp/views/index.ts:61]: Very verbose, but I prefer this way rather than listing heavy stuff in a constant, and passing it everywhere, letting pass through any new heavy column introduction. It's more about setting the right example / principle than being situationally correct.

Fraggle (approved):
LGTM but very verbose indeed. Feels a bit flimsy

id13:
Still less verbose than rust 

🚪  <==== 🚶‍♂️

id13:
Front edge tests successful, merging

PR #29270, opened 2026-07-21
[Forks] Block sending while branch prepares

philipperolet:
## Description

Fixes https://github.com/dust-tt/tasks/issues/9719

Context: https://dust4ai.slack.com/archives/C050A0S2Z7F/p1784649803525579

Large conversations can remain in file-copy preparation after the user reaches the branched conversation. The composer previously allowed sending during this window, even though the backend rejects the message.

Polls conversation readiness only while file copying is pending, shows a preparation banner, and disables Send while keeping the editor available for drafting.

## Risks

Blast radius: Message composition in newly branched conversations while background file copying runs.
Risk: low

## Deploy Plan

- deploy front

## Tests

- [x] ConversationViewer tests (7 tests)


smb2268 (commented):
LGTM, thanks!

PR #29272, opened 2026-07-21
[Forks] Signal branch preparation completion

philipperolet:
## Description

Follow-up to #29270 for https://github.com/dust-tt/tasks/issues/9719.

The initial fix polls the full conversation endpoint every two seconds while fork files are copied. That endpoint emits a conversation.accessed WorkOS audit event, so a long copy can create many duplicate audit records.

Publish a private conversation SSE event after the fork status is committed and refetch when the UI receives it while the fork remains pending. A 60-second poll remains as a safety net for missed events and rolling deploys. Normal copies complete through SSE before that fallback runs. The event is filtered out of the public v1 stream and documented in the private Swagger schema.

## Risks

Blast radius: completion signaling for newly forked conversations after background file copying.
Risk: low. Event publication happens after the idempotent status update; a publication failure retries the Temporal activity, and slow polling remains as a fallback. The existing activity name and signature are unchanged.

## Deploy Plan
- deploy front
- deploy front-sse
## Tests
- [x] ConversationViewer tests (7 tests)
- [x] Public conversation SSE filtering test

### DIFF
--- a/front-api/public/swagger.json
+++ b/front-api/public/swagger.json
@@ -12204,6 +12204,9 @@
             "$ref": "#/components/schemas/PrivateCompactionMessageDoneEvent"
           },
           {
+            "$ref": "#/components/schemas/PrivateConversationForkPreparedEvent"
+          },
+          {
             "$ref": "#/components/schemas/PrivateConversationTitleEvent"
           },
           {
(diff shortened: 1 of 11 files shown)


smb2268 (commented):
Ah, nice. Thanks for fixing!

PR #29273, opened 2026-07-21
[front] fix: avoid webhook source test name collisions

aubin-tchoi:
## Description

Webhook source tests could intermittently fail with a `SequelizeUniqueConstraintError` because the factory generated default names from only 1,001 possible values while names are unique per workspace.

Use an 8-character alphanumeric suffix, consistent with other test factories, to make collisions negligible. This only affects test data generation.

## Tests

- Yes

## Risk

- Low.

## Deploy Plan

- No deploy.


aubin-tchoi:
PMRR

smb2268:
LGTM!

PR #29274, opened 2026-07-22
[front] enh: cache OpenAI tools and system prompt

aubin-tchoi:
## Description

Follow up on https://github.com/dust-tt/dust/pull/29227 by adding explicit cache breakpoints for OpenAI after the instructions and the shared-context system tiers. The first system breakpoint also closes the reusable tools prefix, matching our strategy for Anthropic providers.

The implementation looks deduplicated; it covers both the old and the new router.

Note: there's still the equipped-skills breakpoint (added in #29227 ). We have a limit of 4 breakpoints as per the doc, which we don't exceed here.

Tested [here](https://langfuse.dust.tt/project/cmi4ft2qm00061707q4a9azyc/traces?filter=metadata%3BstringObject%3BconversationId%3B%3D%3BikODhegnoQ&peek=fc30aab0218d6a7bc4528633ae612297&timestamp=2026-07-22T12%3A07%3A14.548Z&observation=f52df066252825d1).

## Tests

- Added tests that check where the cache breakpoints are.
- Also covered the unchanged flattened legacy prompt when explicit caching is disabled.

## Risk

- Low.

## Deploy Plan

- Deploy front.

### DIFF
--- a/front/lib/api/llm/utils/openai_like/responses/conversation_to_openai.ts
+++ b/front/lib/api/llm/utils/openai_like/responses/conversation_to_openai.ts
@@ -9,10 +9,12 @@ import {
   parseOpenAIToolSearchItem,
 } from "@app/lib/api/llm/clients/openai/utils/tool_search_passthrough";
 import type { XaiWhitelistedModelId } from "@app/lib/api/llm/clients/xai/types";
+import type {
+  ExclusiveToolChoiceParameters,
+  SystemPromptInput,
+} from "@app/lib/api/llm/types/options";
 import {
-  type ExclusiveToolChoiceParameters,
   normalizePrompt,
-  type SystemPromptInput,
   systemPromptToText,
 } from "@app/lib/api/llm/types/options";
 import {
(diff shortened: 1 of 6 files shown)


front/lib/api/llm/utils/openai_like/responses/conversation_to_openai.ts
```diff
@@ -9,7 +9,12 @@ import {
   parseOpenAIToolSearchItem,
 } from "@app/lib/api/llm/clients/openai/utils/tool_search_passthrough";
 import type { XaiWhitelistedModelId } from "@app/lib/api/llm/clients/xai/types";
-import type { ExclusiveToolChoiceParameters } from "@app/lib/api/llm/types/options";
+import {
```
flvndvd, line 16: Extract types import from regular imports.

front/lib/api/llm/utils/openai_like/responses/conversation_to_openai.ts
```diff
@@ -158,16 +163,74 @@ function toToolCallOutputItem(
...
+    if (!text) {
+      return [];
+    }
+
+    return [
+      {
+        role: promptRole,
+        content: [
+          {
+            type: "input_text" as const,
```
flvndvd, line 203: You should not need all those const here.

flvndvd (approved):

front/lib/api/llm/utils/openai_like/responses/conversation_to_openai.ts
aubin-tchoi: the SDK types these strictly, they don't support a `string` there

PR #29275, opened 2026-07-22
[connectors] refactor: Remove legacy Notion GC patch code paths

id13:
## Description

Advances the Temporal patch lifecycle in the Notion garbage collection workflow, following up on #28955 (deployed 2026-07-17):

- `reduce-gc-batches-per-run`: removes the `deprecatePatch` marker entirely (final lifecycle step).
- `gc-batch-deletion-continue-as-new`: moves from `patched()` to `deprecatePatch()` and deletes the legacy pre-patch code path (parallel PQueue batch deletion, no mid-phase continue-as-new).
- Removes the now-dead `MAX_PENDING_GARBAGE_COLLECTION_ACTIVITIES` constant (only consumer was the legacy branch).

The GC workflow continues-as-new at least every 12h, so open histories have cycled well past the 2026-07-17 deploy and no longer carry the pre-patch shape.

Pre-deploy check: no running `notionGarbageCollectionWorkflow` execution with `StartTime < 2026-07-18` in the connectors namespace (a hit would be a stuck run to reset first; worst case is a workflow-task retry loop on that run, no data loss).

## Risk

Low. Replay-safety precondition above; type-check and all 15 temporal bundles build (`npm run build:temporal-bundles`).

## Deploy Plan

Deploy connectors.

### DIFF
--- a/connectors/src/connectors/notion/temporal/config.ts
+++ b/connectors/src/connectors/notion/temporal/config.ts
@@ -17,8 +17,6 @@ export const MAX_PAGE_IDS_PER_CHILD_WORKFLOW = 64;
 
 export const MAX_PENDING_UPSERT_ACTIVITIES_PER_CHILD_WORKFLOW = 5;
 
-export const MAX_PENDING_GARBAGE_COLLECTION_ACTIVITIES = 1;
-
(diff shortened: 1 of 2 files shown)


id13:
PMRR

PR #29279, opened 2026-07-22
[front] enh: Add ProviderError with central catch in tool execution wrapper

id13:
Stack 1/4 — reduce MCP tool-error alert noise to the failures we actually care about.

## Problem

`MCPError.tracked` defaults to `true` and requires opt-out; almost nobody opts out. Result: ~140k tracked `Tool execution error` logs/week feeding the "Tool execution errors" monitors — salesforce alone is ~40k/week and is virtually all model-generated bad SOQL (`No such column`, `unexpected token`). None of it is actionable.

Meanwhile the two failure modes we DO want to alert on are invisible:
- an uncaught `throw` inside a tool handler is swallowed by the MCP SDK into an `isError` text result before `withToolLogging` sees it — no metric, no log;
- provider-side 5xx errors are indistinguishable from user/model input errors in the `tracked` stream.

## This PR

Ships the mechanism, used nowhere yet (call sites come in the next PR of the stack):

- **`ProviderError`** (`mcp_errors.ts`): thrown by tool API clients at the lowest level when the service a tool depends on fails unexpectedly (HTTP 5xx-class, provider SDK internal failures, unreachable Dust infra). Explicitly NOT for user/config-driven failures (4xx, invalid queries) nor arbitrary user endpoints (http_client and co).
- **Central catch in `withToolLogging`** around the tool callback:
  - thrown `ProviderError` → converted to a **tracked** `MCPError` (alerts regardless of the tool's tracking defaults), metric tagged `error_type:provider_error`;
  - thrown `MCPError` → honored as if returned (its `tracked` flag is respected) — gives deep call stacks an intentional abort channel instead of Result-plumbing through every layer;
  - anything else → **bug**: `logger.error("Tool execution uncaught error")` + `use_tools_error.count` with `error_type:uncaught_error`, then rethrown (model-visible behavior unchanged: the SDK still turns it into a generic `isError` result). The uncaught metric is deliberately NOT gated on `enableAlerting`, so an absolute-count monitor on `error_type:uncaught_error` can alert "no questions asked".

## Stack plan

1. **this PR** — mechanism only;
2. throw `ProviderError` from every server API client on provider 5xx (per-server funnels; CoreAPI/DustAPI `internal_server_error` for infra-backed tools);
3. explicit `tracked: true` opt-ins where low-level 5xx capture is not possible (e.g. providers whose SDK hides status; network-level failures for Dust-managed providers like firecrawl);
4. flip the `MCPError.tracked` default to `false`.

### DIFF
--- a/front/lib/actions/mcp_errors.ts
+++ b/front/lib/actions/mcp_errors.ts
@@ -36,3 +36,33 @@ export class MCPError extends Error {
     this.code = code;
   }
 }
+
+export function isMCPError(error: unknown): error is MCPError {
+  return error instanceof MCPError;
+}
+
+/**
+ * Thrown by tool API clients when the service a tool depends on fails unexpectedly: HTTP
+ * 5xx-class responses, provider SDK internal errors, unreachable Dust infra. Caught in
+ * `withToolLogging` and surfaced as a tracked `MCPError`, so it reaches our observability
+ * stack regardless of the tool's error-tracking defaults.
+ *
+ * Must not be used for failures driven by user input or configuration (4xx-class responses,
+ * invalid queries, permission errors) nor for arbitrary user-provided endpoints (e.g.
+ * http_client URLs): those are not actionable on our side.
+ */
+export class ProviderError extends Error {
+  public readonly status?: number;
+
+  constructor(
+    message: string,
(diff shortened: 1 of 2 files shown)


id13:

- #29279 ◀

PR #29282, opened 2026-07-22
[front] enh: Throw ProviderError from tool API clients on provider 5xx

id13:
Stack 2/4 (on top of #29279) — every tool API client now throws `ProviderError` when the upstream service fails unexpectedly, at the lowest level possible.

## What "unexpected provider failure" means here

- HTTP >= 500 from the provider (raw-fetch funnels and SDK errors carrying a status);
- `internal_server_error` from our own infra (CoreAPI code, Dust API error type) via two small shared guards in `mcp_errors.ts` (`throwOnCoreAPIInternalError`, `throwOnDustAPIInternalError`).

Deliberately NOT converted in this pass: 4xx, 429, network/timeout failures (PR 3/4 of the stack decides per-provider tracking for those), and anything hitting **user-controlled endpoints** (http_client URLs, val_town trigger-val, user file URLs) — a 5xx there is the user's endpoint's fault.

## Shape of the change, per client type

- **Raw-fetch funnels** (ashby, gong, clari_copilot, luma, productboard, statuspage, zendesk, vanta, databricks, servicenow, ukg_ready, workday, freshservice, salesloft, slab, monday, jira, confluence, front, openai_usage, gmail, outlook, hubspot raw paths): `status >= 500` → throw at the funnel; `< 500` behavior unchanged (funnels that used to throw bare `Error` on non-ok now throw `MCPError { tracked: false }` so they don't count as uncaught bugs).
- **SDK clients** (notion, github/octokit, slack family, googleapis, microsoft graph, hubspot, fathom, exa, val_town, elevenlabs, firecrawl/serpapi/serper/spider, snowflake, salesforce/jsforce, openai/gemini image gen): one small `throwIf<Provider>ProviderError`-style guard per server (or shared dir), using the SDK's own error classes — verified against the installed packages (e.g. jsforce only exposes 5xx as `ERROR_HTTP_<status>` codes when the body is unparseable; snowflake's `RequestFailedError` is disjoint from user-SQL `OperationFailedError`).
- **Swallowing wrappers** made ProviderError-aware (rethrow before their catch-all conversion): hubspot/salesforce/google_sheets `withAuth`, jira/confluence `withAuth`, pod_manager `withErrorHandling`, freshservice's 36 per-tool catches, outlook's 9 `{ error }`-returning catches, github's 20 catches, etc.

## Notable behavior fixes along the way

[...]

PR #29283, opened 2026-07-22
[front] enh: Track unexpected errors from Dust-managed tool providers explicitly

id13:
Stack 3/4 (on top of #29282) — explicit `tracked: true` opt-ins for Dust-managed providers, ahead of the default flip in the next PR.

For servers whose provider is a Dust-managed dependency (Dust credentials, no customer data in the failure path), ANY unexpected failure is ours to know about — not just the 5xx that PR 2 captures at the funnel (e.g. Firecrawl being down can also look like bad requests, exhausted credits, or unexpected response shapes):

- web_search_browse: search-provider failures (Firecrawl/SerpAPI/Serper/Exa) → `tracked: true`;
- speech_generator / sound_studio: ElevenLabs catch-all fallbacks → `tracked: true`;
- exa: catch-all fallback + missing Dust-managed `EXA_API_KEY` → `tracked: true`.

image_generation needed nothing: it already computes `tracked` explicitly (`safety_blocked` untracked, the rest tracked).

These are no-ops until the next PR flips the default; they exist so the signal survives the flip.

### DIFF
--- a/front/lib/api/actions/servers/exa/tools/index.ts
+++ b/front/lib/api/actions/servers/exa/tools/index.ts
@@ -23,7 +23,12 @@ async function exaSearch({
   category: "people" | "company";
 }) {
   if (!credentials.EXA_API_KEY) {
-    return new Err(new MCPError("EXA_API_KEY is required for Exa search."));
+    // Dust-managed credential: missing means a Dust misconfiguration, not a user error.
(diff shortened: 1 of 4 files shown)


id13:

- #29279
    - #29282
        - #29283 ◀

PR #29286, opened 2026-07-22
[LLM] Document leaf/composite converter pattern in SDK input converters

pmilliotte:
## Description

Expand the leaf/composite comment in all five SDK input converters (anthropic_ai, google_genai, mistralai, openai_completions, openai_responses) to explain why converters split into leaves and composites, how endpoints override a single step via `this`, and that composites are override points too.

## Tests

Not needed.

## Risk

None.

## Deploy Plan

- Deploy `main` on `front`


aubin-tchoi (approved):
thanks 🙏🏻

PR #29289, opened 2026-07-22
Admin governance: add availibility to skill model

fabiencelier:
## Description

Related to https://github.com/dust-tt/tasks/issues/9725

Skill configurations currently have a boolean isDefault flag. We are replacing it with a 3-state availability column:

  ```
┌──────────────────┬─────────────────────────────────────────────────────────────────────┐
  │   availability   │                               Meaning                               │
  ├──────────────────┼─────────────────────────────────────────────────────────────────────┤
  │ editors          │ Only skill editors (new state, not yet writable)                    │
  ├──────────────────┼─────────────────────────────────────────────────────────────────────┤
  │ workspace_users  │ All workspace users (maps to isDefault: false, default)             │
  ├──────────────────┼─────────────────────────────────────────────────────────────────────┤
  │ users_and_agents │ Users and agents — discoverable by agents (maps to isDefault: true) │
  └──────────────────┴─────────────────────────────────────────────────────────────────────┘
```

### Migration plan 

A pre-deploy backfill alone would lose data: during the deploy window, pods running old code still write isDefault without touching availability, so any row updated between the backfill and the last old pod dying would carry a stale availability. Instead:

**PR 1 (this PR):**
  1. Pre-deploy migration adds a nullable availability column with no DB default to `skill_configurations` and `skill_versions`. `NULL` unambiguously means "row written by code predating the column; `isDefault` is
  authoritative".
  2. New code dual-writes both columns on every create/update, keeping them consistent.
  3. Reads go through a fallback: `availability ?? (isDefault ? "users_and_agents" : "workspace_users")`.
  4. After this deploy completes (no old pods left), run the backfill script to fill remaining NULLs from
  `isDefault`.

**PR 2:**
Drop isDefault and rely on availability only

## Tests

added unit test

## Risk

PR can be reverted if needed without data loss

## Deploy Plan

lock front
run pre deploy migration
deploy front
unlock front
run post-deploy migration

### DIFF
--- a/front/lib/models/skill.ts
+++ b/front/lib/models/skill.ts
@@ -10,6 +10,7 @@ import { FileModel } from "@app/lib/resources/storage/models/files";
 import { UserModel } from "@app/lib/resources/storage/models/user";
 import { WorkspaceAwareModel } from "@app/lib/resources/storage/wrappers/workspace_models";
 import type {
+  SkillAvailability,
   SkillReinforcementMode,
   SkillSourceMetadata,
   SkillSourceType,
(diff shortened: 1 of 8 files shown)


ykmsd (approved):
Checked with @matteotrab looks good 👍

id13 (approved):
LGTM

PR #29290, opened 2026-07-22
Handle builder scope for api keys

tdraier:
## Description

Prerequisite for retiring `ensureIsBuilder()` on the public `agent_configurations` PATCH endpoint (dust-tt/tasks#9710) — gives API keys a way to carry builder-equivalent access via group membership, the same mechanism already used for users, instead of relying purely on the key's `role` field. Also removes `builder` as a selectable role for *new* API keys, in line with the broader builder-role deprecation.

**Backend / infrastructure:**
- **`KeyResource`**: added `setGroupMembership` (mutate `groupIds` post-creation — didn't exist before, only set at `makeNew`) and `syncBuilderGroupMembership` (idempotent ensure-state sync mirroring `GroupResource.syncBuilderGroupMembership` for users). `updateRole` now calls the sync automatically and derives its own workspace internally (one extra DB fetch, on the sole existing caller — an admin CLI command — so the method signature stays clean with no `workspace` param needed).
- **Key creation route**: builder-role keys (still creatable via the API/CLI, just no longer via the UI — see below) get the workspace's manual "Builders" group added alongside the global group.
- **`GroupResource.fetchManualBuildersGroup`**: new existence-only lookup (no creation), now the single canonical way the Builders group is looked up everywhere — replaced several previously inconsistent direct `GroupModel.findOne(...)` / `GroupResource.fetchByName(...)` call sites in `syncBuilderGroupMembership`, `fetchOrCreateManualBuildersGroup`, `governance_seeding.ts`, and the `20260721_backfill_builders_group.ts` migration.
- **New migration** `20260721_backfill_builder_key_groups.ts`: one-shot, dry-run-by-default backfill adding the Builders group to every active, non-system, builder-role key that doesn't already have it.
- **Bug fix** in `20260721_backfill_governance_capabilities.ts`: the backfill treated *any* non-default capability configuration as a terminal manual override, so a capability already seeded to `builders` could never be reconciled to `everyone` once the driving legacy flag changed — it would silently do nothing forever. Now it recognizes a `groups` configuration pointing at exactly the Builders group (fetched once per workspace) as our own prior seeding output, as opposed to a genuine custom admin override, and reconciles it when the target changes. Also removed a stray `console.log` debug line.

**UI:**
- Removed the "Can also create and modify resources" (`builder`) option from the API key creation dialog (`NewAPIKeyDialog.tsx`) — only `user` and `admin` remain selectable for new keys. `KEY_ROLES` in `utils.ts` dropped `"builder"` accordingly.
- Changed the dialog's default selected role from `builder` to `user` (least-privilege default).
- Left `APIKeysList.tsx`'s display logic untouched — it still renders pre-existing `builder`-role keys correctly (switches on the broader `RoleType`, not the creation-only `KeyRole`), so no backward-compat break for keys created before this change.

## Tests

- `key_resource.test.ts`: `setGroupMembership`, `syncBuilderGroupMembership`, and `updateRole` add/remove Builders-group membership (22 tests).
- `front-api/routes/w/[wId]/keys/index.test.ts`: key creation with `builder` / `user` / default role (5 tests).
- `group_resource.test.ts` (34 tests) and `governance_seeding.test.ts` (4 tests) re-verified green after the `fetchManualBuildersGroup` refactor.
- `npx tsgo --noEmit` clean on both `front` and `front-api` for every file touched by this PR.
- No test exists for the migration scripts themselves, matching this repo's existing convention (no other `migrations/*.ts` has a test file).
- UI change verified via typecheck + biome only (no existing test file for this component directory).

## Risk

- The key-groups backfill is strictly additive (only adds group membership, never removes), and is dry-run by default behind `--execute`.
- The governance-capabilities reconciliation fix only changes behavior for the specific case of a `builders`-scoped capability whose legacy signal has since changed — a true manual override (any other custom group set, or `everyone`) is still preserved untouched, same as before.
- Removing `builder` from the UI picker does not touch the backend `role` enum or existing keys — it only narrows what's selectable going forward. Keys can still technically be created with `role: "builder"` via direct API/CLI use; only the guided UI flow is restricted.
- `ensureIsBuilder()` itself is **not** removed in this PR — this only lands the underlying infrastructure. No change to request authorization behavior yet.
- **Unrelated to this diff**: `tsgo` CI currently fails on `front/lib/api/poke/plugins/spaces/send_activation_nudge.ts` (a stale 2-arg call to `emitActivationEvent`, which now requires a `userId`). This pre-exists on `main` itself and is not part of this PR's changes — left untouched per explicit direction, to be fixed separately.

## Deploy Plan

- Merge and deploy as usual.
- Run `20260721_backfill_builder_key_groups.ts --execute` once deployed, to backfill existing builder-role keys into the Builders group.
- `20260721_backfill_governance_capabilities.ts` is safe to re-run (idempotent) to pick up the corrected reconciliation logic for any workspace currently stuck in a stale `builders` configuration.

### DIFF
--- a/front-api/routes/w/[wId]/keys/index.test.ts
+++ b/front-api/routes/w/[wId]/keys/index.test.ts
@@ -1,3 +1,4 @@
+import { GroupResource } from "@app/lib/resources/group_resource";
 import { KeyResource } from "@app/lib/resources/key_resource";
 import { createPrivateApiMockRequest } from "@app/tests/utils/generic_private_api_tests";
 import { redactString } from "@app/types/shared/utils/string_utils";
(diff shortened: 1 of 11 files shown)


tdraier:

- #29290 ◀

matteotrab (approved):

PR #29291, opened 2026-07-22
[Activation] Nudge conversation is per-user

xebeohp:
## Description
Permanent fix to #29288 to change the nudge conversation mechanism to be per-user, meaning the nudge conversation can now be sent to a subset of the members of the pod (via poke or the scheduler).

- Adds a member multi select (all selected by default) to the Poke plugin UI. 
- For each selected member, provisions their activation trigger if missing before emitting, so members without a trigger still get their own @dust conversation.
- Reports per-member outcomes.
- Uses the same per-user `emitActivationEvent` path as the scheduler, but as a manual support, the Poke plugin bypasses the scheduler's eligibility gating, so it can re-nudge someone.

## Tests
Tested the poke plugin flow locally.

## Risk
Behind activation pod Poke plugin.

## Deploy Plan
Deploy to front.

### DIFF
--- a/front/lib/api/poke/plugins/spaces/send_activation_nudge.ts
+++ b/front/lib/api/poke/plugins/spaces/send_activation_nudge.ts
@@ -1,15 +1,27 @@
-import { emitActivationEvent } from "@app/lib/api/activation/trigger";
+import {
+  createActivationTrigger,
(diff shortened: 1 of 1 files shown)


frankaloia (approved):
LGTM

PR #29292, opened 2026-07-22
[front] perf: add workspaceId clause in group_vault

id13:
## Description

We recently introduced a perf improvement by removing one join in the space fetch (was space x group_vault x group, now is vault x group_vault).

But while it's a 50% perf improvement, we can do better.

Indeed, postgres is systematically choosing to use either an nested loop, or a hash join with a seq scan.
This is because when computing join selectivity, pg only has `on vault.id = group_vault.vaultId`, it does not know about the transitive nature of a `workspaceId = $` predicate.

This selectivity gets backed into the different cost functions pg evaluates to choose a plan, and granted it's more nuanced in reality, it will do either 

1. Hash Join + Seq scan
   => That's terrible, but it's because PG has no way to know it can reduce the lookup by an order of magnitude. Any other index scan in the table would result in more work in its model.

2. Nested Loop + Index scan on vaultId
   => That's okayish on worskpaces with very little spaces, but with projects this ship is sailing. On workspace with more than say 40 spaces, hash join starts to become interesting.

If you want more context with actual plans : https://dust4ai.slack.com/archives/C050SM8NSPK/p1784708558128809?thread_ts=1784630404.245539&cid=C050SM8NSPK

Fix : Add worskpaceId in the ON join predicate. It has 0 added cost, and tells pg about the real selectivity, which means it can now use any index starting with workspaceId to build its hashmap and not scan the entire table \o/

## Tests

Tested locally

## Risk

Low
Blast radius: Poke search AFAIK

## Deploy Plan

Deploy front

### DIFF
--- a/front/lib/resources/resource_with_space.ts
+++ b/front/lib/resources/resource_with_space.ts
@@ -87,23 +87,34 @@ export abstract class ResourceWithSpace<
       return [];
     }
 
-    // We use the model directly here; it's a very rare case where we don't check the workspace,
-    // which in this case is due to the fact that we may need to fetch data from public workspaces
-    // as well as the current workspace.
+    // Scope on the fetched blobs' workspaces rather than the authenticated one: some lookups
+    // are intentionally cross-workspace (e.g. unsafeFetchByDustAPIProjectId) and a blob always
+    // lives in the same workspace as its space.
+    const blobWorkspaceIds = [...new Set(blobs.map((b) => b.workspaceId))];
+
     const spaces = await ResourceWithSpace.spaceModel.findAll({
       where: {
         id: blobs.map((b) => b.vaultId),
+        workspaceId: blobWorkspaceIds,
       },
       include: [
         {
           as: "groupSpaces",
           model: GroupSpaceModel,
+          // A where on an include implies required: true;
+          // pass this required: false to keep the original behavior intact
+          required: false,
+          where: {
+            workspaceId: blobWorkspaceIds,
+          },
         },
       ],
       includeDeleted,
       transaction,
-      // WORKSPACE_ISOLATION_BYPASS: Spaces can be public, preventing to enforce a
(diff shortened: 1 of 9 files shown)


front/lib/resources/resource_with_space.ts
```diff
@@ -98,6 +98,15 @@ export abstract class ResourceWithSpace<
         {
           as: "groupSpaces",
           model: GroupSpaceModel,
+          // A where on an include implies required: true; keep the LEFT OUTER
+          // JOIN so spaces without group_vaults rows (soft-deleted) are
+          // fetched.
+          required: false,
+          // This query is not scoped to the authenticated workspace (see
+          // bypass below), so scope the join on the fetched blobs' workspaces.
+          where: {
```
flvndvd, line 107: Even though we are not using much now, we have this concept of `public` spaces, hence why you don't see the `workspace_id` clause above, can we ensure that this new clause does not break this behavior?

front/lib/resources/resource_with_space.ts
```diff
@@ -98,6 +98,15 @@ export abstract class ResourceWithSpace<
         {
           as: "groupSpaces",
           model: GroupSpaceModel,
+          // A where on an include implies required: true; keep the LEFT OUTER
+          // JOIN so spaces without group_vaults rows (soft-deleted) are
```
flvndvd: Not only soft deleted, see below.

id13: yes not only, it's just to keep current behavior, will rewrite

front/lib/resources/resource_with_space.ts
id13: Will add some tests if I see none

id13: I thiiink we don't care anymore about public workspaces, could be wrong

id13:
PMRR

spolu (commented):
LGTM

PR #29294, opened 2026-07-22
[LLM] Add Claude Fable 5 EU agent-platform stream endpoint

pmilliotte:
## Description

Add a Claude Fable 5 stream endpoint served via the EU agent-platform (Vertex) route, so Fable 5 resolves through the new router in EU regional workspaces instead of falling back to the legacy router.

## Tests

Added the characterization test file, mirroring the Fable 5 global contract. Not yet run against the live API — it requires `RUN_LLM_TEST=true` and Vertex ADC. Confirm Fable 5 is actually served on Vertex `europe-west1` before merging.

## Risk

Low.

## Deploy Plan

- Deploy `main` on `front`

### DIFF
--- a/front/lib/llms/stream/endpoints/anthropic_claude_fable_five_eu_agent_platform.ts
+++ b/front/lib/llms/stream/endpoints/anthropic_claude_fable_five_eu_agent_platform.ts
@@ -1,0 +1,25 @@
+import { WithDustClaudeFableFiveConfig } from "@app/lib/llms/providers/anthropic/models/claude_fable_five";
+import { defineDustStreamEndpoint } from "@app/lib/llms/stream/dust_stream_endpoint";
+import { AnthropicClaudeFableFiveEuropeAgentPlatformStream } from "@app/lib/model_constructors/stream/endpoints/anthropic_claude_fable_five_eu_agent_platform";
+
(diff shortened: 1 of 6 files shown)


front/lib/llms/stream/endpoints/anthropic_claude_fable_five_eu_agent_platform.ts
```diff
@@ -0,0 +1,25 @@
...
+    and: [
+      { featureFlags: { contains: "claude_fable_5_feature" as const } },
+      {
+        or: [
+          {
+            featureFlags: {
+              contains: "use_vertex_for_supported_models" as const,
+            },
+          },
+          { isCreditPriced: { eq: true } },
```
Nils-Fedrigo, line 18: not sure but don't you mean (claude_fable_5_feature OR isCreditPriced) AND use_vertex_for_supported_models

Nils-Fedrigo (approved):

front/lib/llms/stream/endpoints/anthropic_claude_fable_five_eu_agent_platform.ts
pmilliotte: seen IRL it's ok, we give support for eu models only with use_vertex_for_supported_models or isCreditPriced

PR #29295, opened 2026-07-22
dsbx: extend DSEC substitution blocklist to remaining reflective headers

fontanierh:
## Description

Follow-up to #29293. Audit of the standard request header list (MDN) against `is_unsafe_substitution_header` found the remaining headers whose values can round-trip back to the client:

- `sec-websocket-protocol` / `sec-websocket-extensions`: the server echoes the selected subprotocol and negotiated extensions in the 101 handshake. The dsbx upgrade path still substitutes headers before splicing raw frames, so a placeholder there comes back substituted. `sec-websocket-protocol` is the strongest of these; extensions rarely survive parsing but cost nothing to cover.
- `idempotency-key`: APIs commonly echo it in 409/422 error bodies.
- `content-disposition`: upload endpoints often echo filenames in JSON responses or errors.
- `prefer`: RFC 7240 `Preference-Applied` can mirror the value.
- `expect`: a 417 response can quote the failed expectation.
- `last-event-id`: occasionally reflected by SSE endpoints.

None of these are credential carriers, so blocklisting only affects placeholder traffic. Also bumps dsbx to 0.1.35 and dust-base to 0.8.56 to ship the change.

## Tests

- Extended the blocklist membership test; ran the full `forward` test suite locally (146 tests pass) and `registry.test.ts` (16 tests pass).

## Risk

Low. Same reasoning as #29293: these headers never legitimately carry secrets, so the only behavior change is placeholders in them passing through unsubstituted.

## Deploy Plan

1. Run the `Release dsbx CLI` workflow (`release-dsbx-cli.yml`) to publish `dsbx-v0.1.35`.
2. Run the `Sandbox Image Registry` workflow (`sandbox-img-registry.yml`) to build `dust-base:0.8.56` in eu and us.
3. Deploy front.
4. From prodbox, kill sandboxes still running the old image so they come back on 0.8.56.

### DIFF
--- a/cli/dust-sandbox/Cargo.lock
+++ b/cli/dust-sandbox/Cargo.lock
@@ -313,7 +313,7 @@ checksum = "92773504d58c093f6de2459af4af33faa518c13451eb8f2b5698ed3d36e7c813"
 
 [[package]]
 name = "dust-sandbox"
-version = "0.1.34"
+version = "0.1.35"
 dependencies = [
  "anyhow",
  "base64",
--- a/cli/dust-sandbox/Cargo.toml
+++ b/cli/dust-sandbox/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "dust-sandbox"
-version = "0.1.34"
+version = "0.1.35"
 edition = "2021"
 
 [[bin]]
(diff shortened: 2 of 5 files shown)


id13:
LGTM

PR #29296, opened 2026-07-22
Admin governance: remove the skills 'isDefault' column and rely on availability

fabiencelier:
## Description

related to https://github.com/dust-tt/tasks/issues/9725

Follow up of https://github.com/dust-tt/dust/pull/29289

Actually drop the isDefault column to rely 100% on the `availability` column. Both columns should be in perfect sync after previous PR.

The isDefault value is still exposed in the types and API at the moment, just no longer stored in DB.

## Tests

Manual test: changing the value in the UI do apply the change to the availability column.

## Deploy Plan

lock front
merge
run pre deploy migration
deploy front
unlock front
run post deploy migration

### DIFF
--- a/front/migrations/20260722_backfill_skill_availability.ts
+++ b/front/migrations/20260722_backfill_skill_availability.ts
@@ -1,74 +1,74 @@
-import {
-  SkillConfigurationModel,
-  SkillVersionModel,
-} from "@app/lib/models/skill";
-import { makeScript } from "@app/scripts/helpers";
-import { runOnAllWorkspaces } from "@app/scripts/workspace_helpers";
-import { availabilityFromIsDefault } from "@app/types/assistant/skill_configuration";
-import type { ModelStatic } from "sequelize";
-
-const WORKSPACE_CONCURRENCY = 16;
-
-// Transition (isDefault -> availability): rows written by code predating the availability column
-// have a NULL availability and an authoritative isDefault. Run this once all pods write the
-// availability column (post-deploy), so no row can be updated by old code afterwards.
-const MODELS: {
-  table: string;
-  model: ModelStatic<SkillConfigurationModel>;
-}[] = [
-  { table: "skill_configurations", model: SkillConfigurationModel },
-  { table: "skill_versions", model: SkillVersionModel },
-];
-
-makeScript({}, async ({ execute }, logger) => {
-  let totalUpdated = 0;
-
-  await runOnAllWorkspaces(
-    async (workspace) => {
-      for (const { table, model } of MODELS) {
-        if (!execute) {
-          const wouldUpdateCount = await model.count({
-            where: { workspaceId: workspace.id, availability: null },
-          });
-
-          if (wouldUpdateCount > 0) {
-            logger.info(
-              { workspaceId: workspace.sId, table, wouldUpdateCount },
-              "Dry run: would backfill availability from isDefault"
-            );
-          }
-          continue;
-        }
-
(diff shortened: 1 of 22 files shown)


front/migrations/20260722_backfill_skill_availability.ts
fabiencelier, line 1: Script can no longer be run as the isDefault column no longer exist

ykmsd (approved):
Shouldn't we remove the old index? 

Reviewed with @matteotrab we are too newbie to tell what we are doing in `front/migrations/post-deploy/20260722112506_drop_skill_is_default.sql` you might want to get a review from somebody else 🫡

id13:
@ykmsd yes [skill_configurations_workspace_id_status_is_default](https://app.pganalyze.com/databases/-642267629/indexes/114188421896) should be dropped, good catch

front/migrations/20260722_backfill_skill_availability.ts
id13: Why not delete ?

front/migrations/post-deploy/20260722112506_drop_skill_is_default.sql
```diff
@@ -0,0 +1,88 @@
...
+SET SESSION lock_timeout = 3000;
+ALTER TABLE "public"."skill_configurations" ADD CONSTRAINT "pgschemadiff_tmpnn_4EqU48Y9Qjm2GrORrLv5qQ" CHECK("availability" IS NOT NULL) NOT VALID;
+
+/*
+Statement 1
+*/
+SET SESSION statement_timeout = 1200000;
+SET SESSION lock_timeout = 3000;
+ALTER TABLE "public"."skill_configurations" VALIDATE CONSTRAINT "pgschemadiff_tmpnn_4EqU48Y9Qjm2GrORrLv5qQ";
+
```
id13: I don't think we need the constrain

front/migrations/post-deploy/20260722112506_drop_skill_is_default.sql
```diff
@@ -0,0 +1,88 @@
...
+SET SESSION statement_timeout = 3000;
+SET SESSION lock_timeout = 3000;
+ALTER TABLE "public"."skill_configurations" ALTER COLUMN "availability" SET DEFAULT 'workspace_users'::character varying;
+
+/*
+Statement 4
+*/
+SET SESSION statement_timeout = 3000;
+SET SESSION lock_timeout = 3000;
+ALTER TABLE "public"."skill_configurations" DROP CONSTRAINT "pgschemadiff_tmpnn_4EqU48Y9Qjm2GrORrLv5qQ";
```
id13: just to drop it here ? mmmm

ykmsd:
(fyi it's @matteotrab who caught that we should drop it so good catch matteo 🎉)

id13 (approved):

front/migrations/post-deploy/20260722112506_drop_skill_is_default.sql
fabiencelier: From claude:
> the temp-constraint dance existed so the NOT NULL check could be validated without holding an exclusive lock (VALIDATE CONSTRAINT only takes a share lock, and SET NOT NULL then skips the scan by proving it from the constraint). Plain SET NOT NULL scans the whole table under an ACCESS EXCLUSIVE lock instead. For skill_configurations/skill_versions sizes that's fine

but probably overkill indeed, I have removed them

PR #29297, opened 2026-07-22
[AGENTS] Stop exposing retired global agents via the public API

pmilliotte:
## Description

Retired global agents (e.g. `gpt-4`) stay resolvable internally so past conversations keep running, but the public `GET /api/v1/w/{wId}/assistant/agent_configurations/{sId}` endpoint now returns 404 for them instead of resolving them (they were leaking through the by-sId fetch, which bypasses the retired filter that already hides them from the list/search endpoints).

## Tests

Added a functional test asserting `GET …/agent_configurations/gpt-4` returns 404; ran the `agent_configurations` suite locally (8 files / 69 tests pass).

## Risk

Low.

## Deploy Plan

- Deploy `main` on `front`

### DIFF
--- a/front-api/routes/v1/w/[wId]/assistant/agent_configurations/[sId]/index.ts
+++ b/front-api/routes/v1/w/[wId]/assistant/agent_configurations/[sId]/index.ts
@@ -3,6 +3,7 @@ import {
   getAgentConfiguration,
 } from "@app/lib/api/assistant/configuration/agent";
 import { patchAgentConfigurationFromJSON } from "@app/lib/api/assistant/configuration/yaml_import";
+import { isRetiredGlobalAgent } from "@app/lib/api/assistant/global_agents/global_agents";
 import { setAgentUserFavorite } from "@app/lib/api/assistant/user_relation";
 import logger from "@app/logger/logger";
 import type {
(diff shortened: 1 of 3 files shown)


front-api/routes/v1/w/[wId]/assistant/agent_configurations/[sId]/index.ts
```diff
@@ -291,10 +292,14 @@ app.get(
 
     const configVariant = variant ?? "light";
 
-    const agentConfiguration = await getAgentConfiguration(auth, {
-      agentId: sId,
-      variant: configVariant,
-    });
+    // Retired global agents (e.g. gpt-4) stay resolvable internally for past conversations but
+    // must not be exposed through the public API.
+    const agentConfiguration = isRetiredGlobalAgent(sId)
```
Comment (Nils-Fedrigo) [inline front-api/routes/v1/w/[wId]/assistant/agent_configurations/[sId]/index.ts:297]: don't know but out of context, can we extend this? Are there some other configs we must return null? Deprecated agents maybe?

Nils-Fedrigo (approved):

front-api/routes/v1/w/[wId]/assistant/agent_configurations/[sId]/index.ts
Comment (pmilliotte) [reply front-api/routes/v1/w/[wId]/assistant/agent_configurations/[sId]/index.ts:297]: seen IRL we'll put the MODEL_ONLY_GLOBAL_AGENTS_SID in the RETIRED_GLOBAL_AGENTS_SID list

PR #29298, opened 2026-07-22
Add tool search guidance to OpenAI requests

flvndvd:
## Description

OpenAI previously exposed its native tool search capability without the system guidance already used for Anthropic. This adds the shared provider-neutral guidance to both OpenAI request paths whenever the final request actually contains the search tool.

The implementation follows the existing Anthropic pattern by checking the converted tools before inserting the
shared instruction. Requests without deferred tools remain unchanged, including forced-tool cases where search
is unnecessary.

Anthropic keeps its existing behavior and provider-specific sequencing guidance. The change intentionally adds
no guidance about reusing tools from earlier turns.


aubin-tchoi (approved):
LGTM

PR #29304, opened 2026-07-22
Limit the pruned context indicator to the current interaction

flvndvd:
## Description

The pruned context indicator was set whenever any tool result in the rendered history was pruned, causing the current agent message to show a warning even when pruning affected only older interactions.

This tracks pruning per tool result and surfaces the indicator only when the latest interaction contains a pruned result. Historical pruning still reduces context and remains represented in pruning metrics.


aubin-tchoi (approved):
LGTM

PR #29305, opened 2026-07-22
[LLM] Unregister Gemini 3.5 Flash and 3.1 Flash Lite EU agent-platform stream endpoints

pmilliotte:
## Description

Deregisters the Gemini 3.5 Flash and 3.1 Flash Lite EU agent-platform stream endpoints (they are not available on the EU endpoint) while keeping their definitions in place for when EU support lands.

## Tests

Not needed.

## Risk

Low.

## Deploy Plan

- Deploy `main` on `front`

### DIFF
--- a/front/lib/model_constructors/stream/endpoints/google_ai_studio_gemini_3_5_flash_eu_agent_platform.ts
+++ b/front/lib/model_constructors/stream/endpoints/google_ai_studio_gemini_3_5_flash_eu_agent_platform.ts
@@ -3,6 +3,8 @@ import { GoogleAgentPlatformStream } from "@app/lib/model_constructors/stream/cl
 import type { StreamEndpointConstructor } from "@app/lib/model_constructors/stream/configuration";
 import { EUROPE } from "@app/lib/model_constructors/types/regions";
 
+// Not registered in STREAM_ENDPOINTS: this model is not available on the EU
+// agent-platform endpoint. Kept defined for when EU support lands.
 export class GoogleAiStudioGeminiThreeDotFiveFlashEuropeAgentPlatformStream extends WithGoogleAiStudioGeminiThreeDotFiveFlashConfig(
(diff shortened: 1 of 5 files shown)


front/lib/model_constructors/stream/endpoints/google_ai_studio_gemini_3_5_flash_eu_agent_platform.ts
```diff
@@ -3,6 +3,8 @@ import { GoogleAgentPlatformStream } from "@app/lib/model_constructors/stream/cl
 import type { StreamEndpointConstructor } from "@app/lib/model_constructors/stream/configuration";
 import { EUROPE } from "@app/lib/model_constructors/types/regions";
 
+// Not registered in STREAM_ENDPOINTS: this model is not available on the EU
+// agent-platform endpoint. Kept defined for when EU support lands.
 export class GoogleAiStudioGeminiThreeDotFiveFlashEuropeAgentPlatformStream extends WithGoogleAiStudioGeminiThreeDotFiveFlashConfig(
```
Nils-Fedrigo, line 8: delete definitions in code while not accessible? Your call but I feel like that could confuse agents

Nils-Fedrigo (approved):

front/lib/model_constructors/stream/endpoints/google_ai_studio_gemini_3_5_flash_eu_agent_platform.ts
Nils-Fedrigo: depends on when they are accessible

pmilliotte: Actually it's available but model might have not been enabled in eu. 

Don't want to check rn, will TAL later

(pretty sure it worked at some point)

PR #29309, opened 2026-07-22
[Activation] Add ActivationPod model and migration

ElPicador:
## Description

Adds an `ActivationPod` model as the canonical record for an Activation Pod: its space, owner
(`userId`), and activation trigger (`triggerId`, nullable until provisioned). Today this data is
scattered — a pod is flagged via `ProjectMetadata.provisioningSource`, its owner only lives on the
trigger's `editor` field, and its trigger is resolved by joining through
`WebhookSourcesView`/`Trigger`. `ActivationPod` will replace that with a single row per pod.

This PR only adds the model and an additive schema migration:
- New `activation_pods` table.
- Nullable `activationPodId` FK added to `activation_nudges` and `activation_recommendations`.

Existing columns on `activation_nudges` (`spaceId`, `triggerId`, `userId`) are untouched for now.
Follow-up work will backfill `activationPodId` on existing rows, wire the resource/orchestrator
code to populate it going forward, then tighten both columns to `NOT NULL` and drop the old
`activation_nudges` columns in a later post-deploy migration.

## Deploy Plan

Purely additive (new table + nullable columns) — safe to deploy as a normal pre-deploy migration,
no backfill required before this ships.

### DIFF
--- a/front/lib/models/activation/activation_pod.ts
+++ b/front/lib/models/activation/activation_pod.ts
@@ -1,0 +1,71 @@
+import { TriggerModel } from "@app/lib/models/agent/triggers/triggers";
+import { frontSequelize } from "@app/lib/resources/storage";
+import { DataTypes } from "@app/lib/resources/storage/data_types";
+import { SpaceModel } from "@app/lib/resources/storage/models/spaces";
+import { UserModel } from "@app/lib/resources/storage/models/user";
+import { WorkspaceAwareModel } from "@app/lib/resources/storage/wrappers/workspace_models";
+import type { CreationOptional, ForeignKey } from "sequelize";
+
+// One row = one Activation Pod: a Pod (project space) provisioned by the
+// activation flow. Canonical record for a pod's owner and activation trigger,
+// replacing the ProjectMetadata provisioningSource flag and the join through
+// WebhookSourcesView/Trigger used to resolve a pod's trigger.
+export class ActivationPodModel extends WorkspaceAwareModel<ActivationPodModel> {
+  declare createdAt: CreationOptional<Date>;
+  declare updatedAt: CreationOptional<Date>;
+
+  // The activation pod.
+  declare spaceId: ForeignKey<SpaceModel["id"]>;
+  // The user for whom we created the activation pod.
+  declare userId: ForeignKey<UserModel["id"]>;
+  // The Pod's activation trigger. Null until provisioned.
+  declare triggerId: ForeignKey<TriggerModel["id"]> | null;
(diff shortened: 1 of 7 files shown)


ElPicador:

- #29309 ◀
    - #29313
        - #29321

front/lib/models/activation/activation_pod.ts
```diff
@@ -0,0 +1,71 @@
...
+      type: DataTypes.DATE,
+      allowNull: false,
+      defaultValue: DataTypes.NOW,
+    },
+  },
+  {
+    modelName: "activation_pod",
+    sequelize: frontSequelize,
+    indexes: [
+      { unique: true, fields: ["spaceId"], concurrently: true },
```
frankaloia, line 42: I'm worried about this being restrictive if we want to focus on team or multi-user activation in the future. Should we make the unique grouping be (spaceId, userId) ?

ElPicador: Kept it as-is: the unique index on `spaceId` alone is the invariant this model is built around — one row = one Pod (the doc comment above), and `ActivationPodResource.fetchBySpace()` assumes/returns a single pod per space. Relaxing to `(spaceId, userId)` would allow multiple owner rows for the same space, which contradicts that. If we want team/multi-user activation later, I think that's a separate membership/ownership layer on top of a single canonical pod row, not multiple `ActivationPod` rows per space — happy to discuss further if you see it differently.

front/lib/models/activation/activation_pod.ts
```diff
@@ -0,0 +1,71 @@
...
+import { SpaceModel } from "@app/lib/resources/storage/models/spaces";
+import { UserModel } from "@app/lib/resources/storage/models/user";
+import { WorkspaceAwareModel } from "@app/lib/resources/storage/wrappers/workspace_models";
+import type { CreationOptional, ForeignKey } from "sequelize";
+
+// One row = one Activation Pod: a Pod (project space) provisioned by the
+// activation flow. Canonical record for a pod's owner and activation trigger,
+// replacing the ProjectMetadata provisioningSource flag and the join through
+// WebhookSourcesView/Trigger used to resolve a pod's trigger.
+export class ActivationPodModel extends WorkspaceAwareModel<ActivationPodModel> {
```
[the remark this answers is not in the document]
ElPicador: Addressed in the next PR in the stack — #29313 adds `ActivationPodResource` wrapping this model. Splitting model+migration from resource+wiring across the stack is intentional here.

PR #9749, opened 2025-01-03
Bump keep-alive above default LBs

spolu:
## Description

Bump keep-alive to very high value as inspired by https://github.com/vercel/next.js/pull/35827/files + above known ALB values.

## Risk

Will deploy on front-edge first to test and give it a try. Many potential risks of resource exhaustion. Will monitor and revert accordingly.

## Deploy Plan

- deploy `front` and monitor

PR #29319, opened 2026-07-22
[front-api] Gate Slack channel agent endpoints on agent:publish

tdraier:
## Description

Part of the `isBuilder` removal effort (#9749). Moves the "link/list Slack channels for an agent" feature off the `isBuilder` role and onto the `agent:publish` workspace permission. Backend: the two `front-api` endpoints (PATCH linking channels, GET listing linked channels) now check `auth.hasWorkspacePermission("publish", "agent")` inline instead of the `ensureIsBuilder()` middleware, returning a 403 `workspace_auth_error` on failure. Frontend: `AgentBuilder.tsx` and `AccessSection.tsx` now gate the Slack channel fetch and the "Slack preferences" UI on the `agent:publish` permission (via `useWorkspacePermissions`) instead of `isBuilder(owner)`.

## Tests

`front-typecheck` and `front-api` typecheck/lint pass (pre-commit). No behavioral tests added; the change swaps one authorization gate for an equivalent capability check on both the API and the UI that drives it.

## Risk

Low. Authorization moves from the `isBuilder` role to the `agent:publish` capability, which is the intended set of users who manage published agents. Safe to roll back by reverting the commits.

## Deploy Plan

Standard `front` + `front-api` deploy, no migration or ordering constraints.

### DIFF
--- a/front-api/routes/w/[wId]/assistant/agent_configurations/[aId]/linked_slack_channels.ts
+++ b/front-api/routes/w/[wId]/assistant/agent_configurations/[aId]/linked_slack_channels.ts
@@ -5,7 +5,6 @@ import { DataSourceResource } from "@app/lib/resources/data_source_resource";
 import logger from "@app/logger/logger";
 import { ConnectorsAPI } from "@app/types/connectors/connectors_api";
 import { workspaceApp } from "@front-api/middlewares/ctx";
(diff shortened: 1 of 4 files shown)


matteotrab (commented):
Do we need to update the frontend too?

tdraier:
> Do we need to update the frontend too?

yes , done ✅

matteotrab (approved):

PR #29329, opened 2026-07-22
feat: check publish agent permission backend

matteotrab:
## Description

This PR enforces the agent publish permission check in the backend by replacing the previous feature-flag-based publishing restriction logic with a direct workspace permission check (`auth.hasWorkspacePermission("publish", "agent")`).

The previous check is enforced to not allow users without the permission to unpublish a published agent. This was already enforced in the frontend.

## Tests

Added tests covering: rejection when publishing/unpublishing without the permission, success when the permission is granted via a group, and no gate for edits that don't change the visibility.

## Risks

Low.

## Deploy Plan

Deploy after backfill script completes.


matteotrab:

- #29329 ◀

tdraier (approved):

PR #29330, opened 2026-07-22
[front] Gate agent Insights tab on agent:publish permission

tdraier:
## Description

The Insights tab in `AgentDetailsSheet` was shown to any builder or agent editor. It is now gated on the workspace `publish` permission for `agent` (via `useWorkspacePermissions`), with the permissions fetch disabled while the sheet is closed. Fixes dust-tt/tasks#9752.

## Tests

Manual: verified in the agent details sheet that the Insights tab appears only for users holding the `publish` grant on `agent` and is hidden otherwise. Type-check and biome pass via pre-commit hooks.

## Risk

Low. UI-only visibility change; the Insights tab data endpoints already enforce their own authorization. Users who previously saw the tab as builders/editors but lack the `publish` grant will no longer see it. Safe to roll back by reverting the commit.

## Deploy Plan

Standard `front` deploy, no migrations or coordination required.

### DIFF
--- a/front/components/assistant/details/AgentDetailsSheet.tsx
+++ b/front/components/assistant/details/AgentDetailsSheet.tsx
@@ -211,7 +211,9 @@ export function AgentDetailsSheet({
     isServerSideMCPServerConfigurationWithName(arg, AGENT_MEMORY_SERVER_NAME)
   );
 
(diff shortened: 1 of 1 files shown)


tdraier:
Superseded by #29335 (renamed branch).

PR #29333, opened 2026-07-22
Skills: add script to delete old suggested skills

fabiencelier:
## Description

Script to delete all the old suggested skills so they don't pollute the UI nor the DB

## Risk

Maybe long to run to delete the ~200K suggeestion but IMHO not worth adding a new bulk delete method which we will have to maintain.

## Deploy Plan

deploy front
run script from prodbox


aubin-tchoi (approved):
LGTM

PR #29336, opened 2026-07-22
[front-api] Audit log governance permission updates

tdraier:
## Description

Resolves [dust-tt/tasks#9685](https://github.com/dust-tt/tasks/issues/9685).

Changing governance permissions (who can create/publish agents, access billing, view audit logs, etc.) is a security-sensitive state mutation and must be followed by an audit log event.

All governance-permission writes funnel through the single `setWorkspaceGovernancePermission` capability toggle (called from `PATCH /api/w/:wId/governance-permissions`). This PR emits a `workspace.governance_permission_updated` audit event there, after the capability transition succeeds.

## Changes

- **New audit action** `workspace.governance_permission_updated`:
  - Schema: `front/admin/audit_log_schemas/workspace.governance_permission_updated.json` (target: `workspace`; metadata: `grant_type`, `resource_type`, `scope`, `group_ids`).
  - Added to the `AuditAction` union in `front/lib/api/audit/workos_audit.ts`.
  - Registered with WorkOS (`v1`) and committed to `schema_versions.json`.
- **Emit** in `setWorkspaceGovernancePermission` — fire-and-forget (`void`), after the mutation succeeds, recording the resulting scope and (for `groups` scope) the granted group sIds.
- **Route** threads `getAuditLogContext(auth)` into the lib function (matching the `setGroupSpendLimit` pattern).

## Verification

- `tsgo --noEmit` clean for `front` and `front-api`.
- `biome` clean; pre-commit hooks (typecheck, swagger, lint) pass.
- Audit schema consistency invariants hold: action ↔ schema ↔ version-map in sync, snake_case metadata, emit targets (`["workspace"]`) match the schema.

## Deploy plan

This PR adds a new WorkOS audit-log schema. WorkOS drops events whose schema was never registered, so the schema must be registered against the production WorkOS environment **before/at deploy** — otherwise `workspace.governance_permission_updated` events are silently dropped.

Run the registration script (with production WorkOS credentials in the environment):

\`\`\`bash
npx tsx front/admin/register_audit_log_schemas.ts --execute workspace.governance_permission_updated
\`\`\`

Notes:
- The committed `schema_versions.json` (v1) is already in sync with what this script registers; the script is idempotent, so re-running is safe.
- No DB migration and no config changes are required.

### DIFF
--- a/front/lib/api/permissions/governance.ts
+++ b/front/lib/api/permissions/governance.ts
@@ -1,3 +1,9 @@
+import {
+  buildAuditLogTarget,
+  emitAuditLogEvent,
+  getAuditLogContext,
+} from "@app/lib/api/audit/workos_audit";
+import type { AuditLogContext } from "@app/lib/api/workos/organization";
 import type { Authenticator } from "@app/lib/auth";
 import { DustError } from "@app/lib/error";
 import type { CapabilityState } from "@app/lib/resources/group_permission_resource";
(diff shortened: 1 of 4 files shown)


matteotrab (approved):

front/lib/api/permissions/governance.ts
```diff
@@ -172,9 +178,27 @@ export async function setWorkspaceGovernancePermission(
   const state = stateByKey.get(key);
   assert(state, `Missing capability state for ${key}.`);
 
+  const updatedConfiguration = toConfiguration(state);
+
+  void emitAuditLogEvent({
+    auth,
+    action: "workspace.governance_permission_updated",
+    targets: [buildAuditLogTarget("workspace", auth.getNonNullableWorkspace())],
+    context: auditContext,
```
matteotrab: nit we could directly do 
```ts
context: getAuditLogContext(auth),
```
here instead of passing it to the function

PR #29343, opened 2026-07-22
[front] Add action-recommendations API endpoint

frankaloia:
## Summary

- Prereq: https://github.com/dust-tt/dust/pull/29196
- This purpose of this is only to support the UX as defined in: https://app.dust.tt/share/frame/8afb4896-9522-437a-b25f-a22dd0a278c5
- It is intentional that we don't surface the "activation" naming convention in the API as I don't think we will present that to users. Instead calling it action recommendation, but open to suggestions
- 1 api to list open recommendations + 1 api to update recommendation status

## Testing

* Functional tests
* Local calls with UX changes

## Risk

Low, new endpoints, not accessible via UX yet

## Deployment
1. Deploy front

### DIFF
--- a/front/lib/api/activation/recommendations.ts
+++ b/front/lib/api/activation/recommendations.ts
@@ -1,0 +1,77 @@
+import type { Authenticator } from "@app/lib/auth";
+import type { ActivationRecommendationStatus } from "@app/lib/models/activation/activation_recommendation";
+import { ActivationRecommendationResource } from "@app/lib/resources/activation_recommendation_resource";
+import { ConversationResource } from "@app/lib/resources/conversation_resource";
+import { removeNulls } from "@app/types/shared/utils/general";
+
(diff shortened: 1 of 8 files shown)


front/lib/api/activation/recommendations.ts
```diff
@@ -0,0 +1,75 @@
...
+  );
+
+  const conversationModelIds = removeNulls(
+    recs.map((rec) => rec.conversationId)
+  );
+  const conversations =
+    conversationModelIds.length === 0
+      ? []
+      : await ConversationResource.fetchByModelIds(auth, conversationModelIds);
+  const sIdByModelId = new Map(conversations.map((c) => [c.id, c.sId]));
```
id13: naive question: why not do a join on conversationId = .. and worskpaceId = .. ?

id13 (approved):
LGTM

ElPicador (commented):
👍

PR #29345, opened 2026-07-22
[front] Activation skill: always call create_recommendation before presenting card

frankaloia:
## Summary

- Very minor change to make sure gpt5.6 outputs directive as final answer

## Deploy
1. Deploy fron

## Risk
very low


smb2268 (approved):
LGTM

PR #29348, opened 2026-07-22
feat(sandbox): pod env var update and delete routes

smb2268:
## Description

Follow up of #30206 (list/create routes) — second of the two pod env var route PRs: **update and delete** (`PATCH`/`DELETE /api/w/:wId/spaces/:spaceId/sandbox/env-vars/:id`).

Same gates as the collection routes (workspace-admin + `sandbox_functions` at the sandbox router; project spaces only, via `withSpace({ requireProject: true })` from #30206). Fetches are scope-filtered — workspace rows and other pods' rows resolve to 404. PATCH supports allowed-domains updates and the one-way config → https_secret promotion; demotion is rejected.

## Risk

New endpoints only, admin + FF gated, no consumers until the UI PR. Safe to rollback.

## Tests

Update/delete functional tests: gating, cross-scope 404s (workspace row / other pod's row), promotion + demotion rejection, allowed-domains update, stale-id 404.

PR #29349, opened 2026-07-22
feat(sandbox): shared env vars UI for workspace and pod scopes

smb2268:
## Description

Follow up of `29348` (pod env vars routes). The shared admin UI for sandbox env vars — one component serving the workspace and pod scopes.

**What this PR does:**

1. **Shared UI**: `SandboxEnvVarsSection` extracted from the workspace `EnvironmentSection` (863-line component, moved not rewritten) and parameterized by an optional `spaceId` — the workspace developers page and the pod settings tab render the same component. SWR hooks gain the `spaceId` URL switch; hook and returned-function names drop the `Workspace` prefix (they serve both scopes; wire types were renamed with their declarations in #28861).
2. **Pod settings tab** shows the section for admins with the `sandbox_functions` FF, alongside the network section — mirroring the API gates (workspace-admin only, pod membership deliberately not consulted; documented on the route group in `29348`).
3. **Scope-aware copy**: toasts and section descriptions state the inheritance + precedence rule (workspace vars inherited, pod wins on collision) and that env changes apply to future Computers, while network changes reach running Computers — the two sections genuinely differ there.

🚗 Drive-by: the Members section now sits above the sandbox admin sections in Pod settings — people before plumbing. The Pod network description also drops the proxy-cache implementation detail.

Screenshot 2026-08-07 at 5 23 59 PM

## Risk

Workspace env vars UI is a code move rendered from the same component; the pod surface is new and FF-gated. Safe to rollback.

## Tests

Covered at the layers below: route round trips + cross-scope guards in `29348`, resource semantics in #28861. This PR is presentation only.


smb2268:

- #28861
    - #28862
        - #29348
            - #29349 ◀

frankaloia (approved):
Changes look good and get why we want this, but more general comment I'm wondering if we are going to run into user confusion by mixing workspace-admin only settings with the pod editor accessible settings

PR #29351, opened 2026-07-23
[front] Fix seats flagged "near limit" after pro→business contract switch

tdraier:
## Problem

After a future-dated contract switch (e.g. the legacy-pro → business migration run via `migrate_legacy_pro_monthly_to_business.ts`), many workspaces end up with **all seats flagged "near limit"**. A manual reconcile (poke `reconcile-credit-state`) clears them. Some workspaces switch cleanly, others flip every user — pointing at a race with the contract state at cutover.

## Root cause

The flag is set at `contract.start`, not by the script. In `applyContractStartSubscriptionSwap`, `reconcileWorkspaceUserCreditStates` runs **before** the DB subscription swap (and contract-cache flush). It reads per-seat **balances** from the newly-started contract (passed in as `metronomeContractId`) but per-seat **allowances** from `getSeatAllowancesByNormalizedSeatType` → `getActiveContract`, which still resolves the **previous** contract.

For legacy-pro sources the previous contract has no new-pricing per-seat AWU credit, so the allowance is `0`:

```
effectiveCap = poolCap(0) + seatAllowance(0) = 0
nearLimit    = consumedAwuCredits >= 0.8 * 0   →   consumed >= 0   →   true for any usage
```

A later manual reconcile runs after the swap, so `getActiveContract` returns the business contract with the correct allowance → the flag clears. Same function, correct inputs — which is exactly why manual reconcile "fixes" it.

## Fix

Thread the contract being reconciled into the allowance lookup, so seat balances and allowances always come from the **same** contract. `getSeatAllowancesByNormalizedSeatType` gets an optional `contract` param; the `contract.start` webhook passes the started contract. Falls back to the DB-active contract when omitted (the manual/poke path, where the DB is already correct), so no other caller changes behaviour.

## Test

Adds a regression test in `reconcile_credit_state.test.ts`: with 5000 consumed and a passed contract whose pro allowance is 8000, the user is **not** near limit (5000 < 6400); the pre-fix path (allowance 0) would have flagged it. Also asserts the contract is forwarded to the allowance lookup.

### DIFF
--- a/front/lib/api/metronome/reconcile_credit_state.test.ts
+++ b/front/lib/api/metronome/reconcile_credit_state.test.ts
@@ -1,4 +1,5 @@
 import { getCreditTypeAwuId } from "@app/lib/metronome/constants";
+import type { CachedContract } from "@app/lib/metronome/plan_type";
 import { MembershipResource } from "@app/lib/resources/membership_resource";
 import { renderLightWorkspaceType } from "@app/lib/workspace";
 import { MembershipFactory } from "@app/tests/utils/MembershipFactory";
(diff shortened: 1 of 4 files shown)


front/lib/api/metronome/reconcile_credit_state.test.ts
```diff
@@ -162,4 +187,53 @@ describe("reconcileWorkspaceUserCreditStates", () => {
...
+    const user = await UserFactory.basic();
+    await MembershipFactory.associate(workspace, user, {
+      role: "user",
+      seatType: "pro",
+    });
+
+    // The bug: at contract.start the DB-active contract still resolves to the
+    // previous (legacy) contract, whose pro allowance is 0. Passing the started
+    // contract must yield the real 8000 allowance instead — mirror that by
+    // returning the correct allowance only when a contract is provided.
```
fabiencelier, line 204: A bit weird to have a comment stating there is a bug in the codebase when it is actually fixed

fabiencelier (commented):
LGTM

PR #29354, opened 2026-07-23
[front] enh: show skill import GitHub connection user on hover

aubin-tchoi:
## Description

Show the name of the user who connected the shared GitHub account when hovering their avatar in the skill import dialog.

The backend already sent the relevant data, so the github connection row reuses that field with the standard tooltip.

image

## Tests

- Checked locally.

## Risk

- Low.

## Deploy Plan

- Deploy front.

### DIFF
--- a/front/components/skills/import/GitHubConnectionRow.tsx
+++ b/front/components/skills/import/GitHubConnectionRow.tsx
@@ -12,6 +12,7 @@ import {
   DropdownMenuItem,
   DropdownMenuTrigger,
   GithubMonoLogo,
+  Tooltip,
 } from "@dust-tt/sparkle";
 
 interface GitHubConnectionRowProps {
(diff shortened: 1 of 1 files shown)


front/components/skills/import/GitHubConnectionRow.tsx
```diff
@@ -46,10 +47,18 @@ export function GitHubConnectionRow({
       </div>
       <Chip size="xs" color="success" label="Connected" />
       {connection.connectedBy && (
-        <Avatar
-          size="xxs"
-          name={connection.connectedBy.fullName}
-          visual={connection.connectedBy.imageUrl ?? undefined}
+        <Tooltip
+          label={connection.connectedBy.fullName}
+          tooltipTriggerAsChild
```
aubin-tchoi, line 52: this avoid rendering the tooltip trigger as a `Button`, which would change how the pointer looks on hover

aubin-tchoi:
PMRR

Inesvvv (commented):
LGTM

PR #29355, opened 2026-07-23
Admin governance: add feature flag for the skill publication

fabiencelier:
## Description

Introduce the skill so we can hide the changes behind it

## Deploy Plan

deploy front


ykmsd (approved):
👍

PR #29361, opened 2026-07-23
[front] perf: Serve JIT-only light MCP server views to the conversation input bar

id13:
## Description

Final step is a new endpoint, which exposes a lighter contract and does not fetch the heavy attributes.

Contract: takes spaceIds only; always serves the manual + auto availabilities (never auto_hidden_builder), filtered server-side to JIT-attachable views (isJITMCPServerView: no agent_memory, no tool with a required Dust-configurable input).

Payload: light serialization via MCPServerViewResource.toJSONLight(). Tools keep {name, description} only (schemas dropped), authorization: null, no url/secrets/lastError. The read-only tool details sheet keeps working: it only renders tool names/descriptions from the list, and fetches the full server via /mcp/[serverId] on open.

## Tests

Added tests
Will frontedge it

## Risk

Low
Blast radius: every page load 🥲 

## Plan

- deploy front

### DIFF
--- a/front/components/editor/extensions/shared/slash_suggestion/useSlashCommandCapabilities.ts
+++ b/front/components/editor/extensions/shared/slash_suggestion/useSlashCommandCapabilities.ts
@@ -3,10 +3,12 @@ import {
   isToolWithKnowledge,
 } from "@app/lib/actions/mcp_helper";
 import { getMCPServerRequirements } from "@app/lib/actions/mcp_internal_actions/input_configuration";
-import { isJITMCPServerView } from "@app/lib/actions/mcp_internal_actions/utils";
 import type { MCPServerViewType } from "@app/lib/api/mcp";
 import { CAPABILITIES_SWR_OPTIONS } from "@app/lib/swr/capabilities";
-import { useMCPServerViewsFromSpaces } from "@app/lib/swr/mcp_servers";
+import {
+  useJITMCPServerViewsFromSpaces,
+  useMCPServerViewsFromSpaces,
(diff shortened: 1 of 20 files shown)


id13:

- #29367
    - #29368
        - #29361 ◀

front/components/editor/extensions/shared/slash_suggestion/useSlashCommandCapabilities.ts
```diff
@@ -87,7 +95,6 @@ export function useInputBarSlashCommandCapabilities({
         skills,
         tools: serverViews,
         toolFilter: (serverView) =>
-          isJITMCPServerView(serverView) &&
```
id13, line 90: So, to summarize, we were loading cachedTool (big TOASTED column) just to filter out tools that require configuration, and if my backfill script is right, we have ONE row in all rms in all dbs combined that could be filtered out here 😅

front/lib/resources/mcp_server_view_resource.ts
```diff
@@ -1676,11 +1696,53 @@ export class MCPServerViewResource extends ResourceWithSpace<MCPServerViewModel>
...
+        authorization: null,
+        tools: internalServer.tools.map(({ name, description }) => ({
+          name,
+          description,
+        })),
+      };
+    }
+
+    return this.toJSONWithServer(server);
+  }
```
id13: Not a fan of hacking the existing type, will put a new type

spolu (approved):
LGTM!

PR #29353, opened 2026-07-23
Revert "[front] Revert eager capability preloading"

spolu:
## Description

Reverts dust-tt/dust#29234

I believe we are good to reintroduce eager loading here based on work done this week.

## Tests

N/A

## Risks

Additional load on DB but all heavy CPU DB paths related to this have been cleared

## Deploy Plan

- deploy `front`

PR #29367, opened 2026-07-23
[front] enh: Precompute cachedToolsRequireConfiguration on remote MCP servers

id13:
## Description

#29353 re-enabled eager capability preloading: CapabilitiesPicker fires its data fetches on input-bar mount instead of on open. One of those fetches, GET /mcp/views, opts into all five remote_mcp_servers heavy attributes (authorization, cachedTools, customHeaders, lastError, sharedSecret) and ships full tool input schemas, redacted secrets and remote server specifics to the client. With eager preloading this lands on every conversation page load, induces a lot of detoasting (and AFAIK pg does not have a way to save us here, we have to change our data model)

Long story short we're introducing a new col called cachedToolsRequireConfiguration that will avoid reading tools entirely in the hot path mounted on every page load, and thus save us 20% of global i/o and 15% of cpu on our DB (at peak hours).

This step is migration + write path, then backfill, then introduce a new endpoint that has a lighter type while maintaining same functionality frontend wise.

Tradeoff: We will have more fetches on /mcp/views/[id] than before, as frontend will need to fetch the heavy type when opening the server pane.

## Tests

existing

## Risk

Low
RemoteMCPServer creation

## Plan

- Merge
- Pre-deploy
- Deploy

### DIFF
--- a/front-api/routes/w/[wId]/mcp/[serverId]/sync.test.ts
+++ b/front-api/routes/w/[wId]/mcp/[serverId]/sync.test.ts
@@ -1,14 +1,62 @@
+import {
+  DEFAULT_MCP_ACTION_VERSION,
+  DEFAULT_MCP_SERVER_ICON,
+} from "@app/lib/actions/constants";
+import { fetchRemoteServerMetaDataByServerId } from "@app/lib/actions/mcp_metadata";
+import type { MCPServerType, MCPToolType } from "@app/lib/api/mcp";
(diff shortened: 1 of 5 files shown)


id13:

- #29367 ◀
    - #29368
        - #29361

id13:
PMRR, I think only debate here would be the naming of the column, happy to rename, it's a small table so w/e

spolu (commented):
LGTM

PR #29371, opened 2026-07-23
fix(sandbox): isolate GCS mount credentials

fontanierh:
## Description

Switch GCS mounts to one CAB-scoped token per mount, written through the root-owned helper and served by the protected broker staged in #29447. The workload UID cannot reach the broker, and token files stay root-owned with mode 0600.

During the rollout, refresh writes a combined token to the legacy `/tmp/token.json` path when the new image helper is unavailable, then requests recreation. This keeps pod-state flushes working while old sandboxes drain. Transient provider errors do not trigger recreation.

Depends on #29447.

## Tests

Added coverage for per-mount grants, old-image refresh fallback, transient refresh errors, fail-closed broker startup, firewall ordering, and startup tracing.

## Risk

The cutover depends on the image from #29447. Mount and broker failures fail closed; old images retain a working credential long enough to flush and recreate.

## Deploy Plan

1. Merge #29447 and publish its dust-base and dsbx versions.
2. Deploy this application change after the image is available.
3. Open `/poke/kill` and trigger a kill request for the previous image.
4. Monitor GCS mint, broker startup, mount, legacy fallback, and denial-check spans and logs.

### DIFF
--- a/front/lib/api/file_system/sandbox/gcs_sandbox_mount_adapter.ts
+++ b/front/lib/api/file_system/sandbox/gcs_sandbox_mount_adapter.ts
@@ -3,6 +3,7 @@ import {
   mintDownscopedGcsToken,
 } from "@app/lib/api/sandbox/gcs/token";
 import type { SandboxImage } from "@app/lib/api/sandbox/image/sandbox_image";
+import { traceSandboxStartupPhase } from "@app/lib/api/sandbox/instrumentation";
 import {
   type RootCommand,
   rootCommand,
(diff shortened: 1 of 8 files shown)


front/lib/api/file_system/sandbox/gcs_sandbox_mount_adapter.ts
```diff
@@ -97,35 +146,77 @@ export class GCSSandboxMountAdapter implements SandboxMountAdapter {
...
+      },
+      { concurrency: targets.length }
+    );
+
+    const tokenWriteError = tokenResults.find((result) => result.isErr());
+    if (tokenWriteError) {
+      if (
+        tokenWriteError.error instanceof GCSMountTokenWriterUnavailableError
+      ) {
+        await sandbox.requestKill();
```
flvndvd, line 175: Can we log the error?

flvndvd (approved):

front/lib/api/file_system/sandbox/gcs_sandbox_mount_adapter.ts
fontanierh: Done in d8603f8461. I moved the log to the selected failure branch, so the error returned from setup is logged exactly once with its mount point before we request recreation.

PR #29373, opened 2026-07-23
Fix malformed JSON error serialization

flvndvd:
## Description

To debug `dsbx` we are seeing a bunch of `Object: Malformed JSON in request body`. It's pretty much not actionable,, this PR preserves the original exception when logging malformed JSON requests so error tracking includes the correct type and stack trace instead of grouping them as generic objects.


fontanierh (approved):
LGMT

PR #29374, opened 2026-07-23
[front] Only the conversation creator can add Spaces to an existing conversation

fontanierh:
## Description

`conversation.requestedSpaceIds` is a **conjunctive** access requirement: a viewer must have read access to *every* Space in the list to read the conversation (see the `.every(...)` filter in `ConversationResource` and `isConversationReadableFromRequestedSpaces`).

Selecting a Space from the input bar appends to that list, and there is no removal path today: `ConversationSelectedSpaceResource.removeForConversation` is never called in production, and the UI disables deselection once the conversation exists. So appending a Space is irreversible.

The append endpoint (`POST /api/w/:wId/assistant/conversations/:cId/selected_spaces`) only gated on the caller being able to **read** the conversation, with no participant, creator, or role check. Same for posting a message with `selectedSpaceIds`. Combined with the two properties above, any user who can read a shared conversation could permanently evict every other participant from it, including their own messages, by selecting a Space those participants lack. Recovery required direct DB surgery.

### Fix

Only the conversation creator may add Spaces to an existing conversation.

`addSelectedConversationSpaces` takes a new explicit `enforceCreatorOnly: boolean` parameter (required, so every call site has to make the decision). When true, it resolves the `ConversationResource` and reuses the existing `ConversationResource.isConversationCreator(auth)` helper, the same one `deleteOrLeaveConversation` uses, which defines the creator as the earliest `ConversationParticipantModel` row by `createdAt`.

| Call site | `enforceCreatorOnly` | Why |
| --- | --- | --- |
| `conversations/[cId]/selected_spaces.ts` (append endpoint) | `true` | User widening an existing conversation. |
| `conversations/[cId]/messages/index.ts` (post message with `selectedSpaceIds`) | `true` | Same, via the message payload. |
| `conversations/index.ts` (create conversation with `selectedSpaceIds`) | `false` | See below. |
| `copySelectedConversationSpacesToChild` (run_agent sub-agent) | `false` | See below. |

I deliberately used an explicit named flag rather than inferring intent from `origin` or from the participant count: inferring from participant count would silently re-open the hole for any future path that adds Spaces before participation, and `origin` is an audit/provenance field, not an authorization one.

### Why the create path is unaffected

On conversation creation, `addSelectedConversationSpaces` runs **before** `ConversationResource.upsertParticipation`. At that moment the brand-new conversation has zero participant rows, so `isConversationCreator` returns `Err("No participants found for conversation")`. A naive unconditional check inside the service would therefore break conversation creation outright. It is also the one moment where there is provably nobody to evict: the caller is the only person who has ever touched the conversation. Hence the explicit `false`, with a comment at the call site pointing at the `upsertParticipation` that follows.

### Why sub-agent inheritance is unaffected

`copySelectedConversationSpacesToChild` copies the parent's Spaces into a run_agent child conversation. That child is created by the tool with `depth >= 1`, and the create route only upserts participation when `depth === 0`, so the child has no participant either. More importantly the inherited Spaces were already revalidated against the same `auth` on the parent (`getValidSelectedSpaceIdsForAgentRun`) immediately above, and the flow is system-initiated: a participant who is not the creator must still be able to run a sub-agent. Hence `false` there too.

### The gate only fires on an actual widening

`enforceCreatorOnly: true` alone is not enough to decide to reject. The input bar resends the conversation's **whole** current selection in `context.selectedSpaceIds` on every message (`InputBar.tsx` unions the conversation's persisted selections with any local ones, and additions on an existing conversation already go through the dedicated `selected_spaces` endpoint). Gating every non-empty call would therefore have rejected plain re-sends and locked non-creator participants out of posting *any* message in a conversation that already had selected Spaces.

So the check only runs when the request adds a Space that is not already in `conversation.requestedSpaceIds`. That is exactly the security property being protected: if the Space is already part of the conjunctive ACL, every current reader already has access to it, so there is nobody the call could evict. A direct API call that does add a new Space still hits the creator check.

### Error handling

New `"conversation_not_creator"` code on the `SelectedConversationSpacesError` union (no HTTP shapes in `lib/api`), mapped to `403 conversation_access_restricted` in `selected_spaces_errors.ts` alongside the existing mappings.

## Tests

Extended `front/lib/api/assistant/conversation/selected_spaces.test.ts` (16 tests, all passing):

- a non-creator participant who **can** read the conversation and **can** read the Space is rejected with `conversation_not_creator`, nothing is persisted, and no audit event is emitted;
- a non-creator participant re-sending the Spaces the conversation already requires is **not** rejected, since the message-post path resends the full selection on every message;
- the creator can still add Spaces to a conversation that already has other participants;
- the creation path (zero participant rows) still works with `enforceCreatorOnly: false`, and the same call with `true` fails, which is exactly why the create path opts out;
- sub-agent inheritance still works when the caller is a participant who did **not** create the parent conversation.

Typecheck clean on `front` and `front-api`. Also re-ran `front/lib/api/actions/servers/run_agent/conversation.test.ts` and the front-api conversation and message route tests.

## Risk

Low. The feature is behind `restricted_spaces_in_input_bar`, which is `stage: "dust_only"`, so it is only reachable internally today.

This change makes an endpoint **stricter**, so the only regression risk is wrongly rejecting a legitimate caller. The two paths where that would hurt, conversation creation and sub-agent inheritance, both opt out explicitly and have dedicated tests, and the "only on an actual widening" condition is what keeps non-creator participants able to keep posting in a conversation that already has selected Spaces.

Rollback-safe: no schema change, no data written differently. Reverting restores the previous (permissive) behaviour.

## Deploy Plan

- No backfill. For context, the US production read replica currently has 10 rows in `conversation_selected_spaces`, all active, across a single workspace, consistent with the flag being `dust_only`.
- No API schema change; the affected endpoints keep their contracts.

Two sibling PRs, `spaces-pod-runtime-guard` (#29375) and `spaces-acl-race` (#29376), also touch `front/lib/api/assistant/conversation/selected_spaces.ts` and will likely need a trivial rebase depending on merge order. Base here stays `main`.

### DIFF
--- a/front-api/routes/w/[wId]/assistant/conversations/[cId]/messages/index.ts
+++ b/front-api/routes/w/[wId]/assistant/conversations/[cId]/messages/index.ts
@@ -318,6 +318,9 @@ app.post(
         spaceIds: context.selectedSpaceIds,
         origin: "input_bar",
         auditContext: getAuditLogContext(auth),
+        // Widening the scope of an existing conversation is irreversible and can lock the other
+        // participants out, so only its creator may do it.
+        enforceCreatorOnly: true,
       });
       if (selectedSpacesResult.isErr()) {
         return apiErrorForSelectedSpaces(ctx, selectedSpacesResult.error);
--- a/front-api/routes/w/[wId]/assistant/conversations/[cId]/selected_spaces.ts
+++ b/front-api/routes/w/[wId]/assistant/conversations/[cId]/selected_spaces.ts
@@ -110,6 +110,9 @@ app.post(
       spaceIds,
       origin: "input_bar",
       auditContext: getAuditLogContext(auth),
+      // Widening the scope of an existing conversation is irreversible and can lock the other
+      // participants out, so only its creator may do it.
+      enforceCreatorOnly: true,
     });
     if (result.isErr()) {
       return apiErrorForSelectedSpaces(ctx, result.error);
--- a/front-api/routes/w/[wId]/assistant/conversations/[cId]/selected_spaces_errors.ts
+++ b/front-api/routes/w/[wId]/assistant/conversations/[cId]/selected_spaces_errors.ts
@@ -15,6 +15,14 @@ export function apiErrorForSelectedSpaces(
           message: error.message,
         },
       });
+    case "conversation_not_creator":
+      return apiError(ctx, {
+        status_code: 403,
+        api_error: {
+          type: "conversation_access_restricted",
+          message: error.message,
+        },
+      });
     case "conversation_not_mutable":
     case "space_not_selectable":
       return apiError(ctx, {
--- a/front-api/routes/w/[wId]/assistant/conversations/index.ts
+++ b/front-api/routes/w/[wId]/assistant/conversations/index.ts
@@ -313,6 +313,10 @@ app.post(
         spaceIds: allSelectedSpaceIds,
         origin: "input_bar",
         auditContext: getAuditLogContext(auth),
+        // The conversation was just created by this caller and has no participant row yet (the
+        // `upsertParticipation` call below is what creates the creator row), so there is no other
+        // participant to evict and nothing to check against.
+        enforceCreatorOnly: false,
       });
       if (selectedSpacesResult.isErr()) {
         return apiErrorForSelectedSpaces(ctx, selectedSpacesResult.error);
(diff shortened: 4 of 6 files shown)


Fraggle (approved):
LGTM

Fraggle (commented):
Note: api created conversations have no creator until a first message by a user is sent.

PR #29375, opened 2026-07-23
Do not honour selected Spaces in pod conversations

fontanierh:
## Description

Selecting a Space from the input bar (`restricted_spaces_in_input_bar`) does two things:

1. It appends the Space to `conversation.requestedSpaceIds`, which is a **conjunctive** access requirement: a viewer must have read access to every listed Space to read the conversation.
2. It widens the agent's runtime scope for that conversation (`getEffectiveSpaceIdsForAgentRun` -> `SkillResource.listForAgentLoop` -> `getSkillServers` -> `DataSourceViewResource.listBySpaceIds`), which is what actually lets the agent retrieve from that Space.

(1) is the only thing that makes (2) safe. If a Space is in the agent's runtime scope but not in `requestedSpaceIds`, restricted data flows into a conversation that does not restrict its readers.

Pod conversations pin their ACL to the project space by design and are visible to everyone with read access to that project, so they cannot express any extra Space requirement. Selection is blocked at every **write** path (`listSelectableSpaces`, `validateSelectableSpaces`, `addSelectedConversationSpaces`), but the **runtime** path had no such guard, and nothing cleared the `conversation_selected_spaces` rows when a conversation moved into a project.

### Failure scenario, no malice required

1. Alice (who has access to the restricted "Legal" Space) selects Legal in a normal conversation. ACL and selection row both include Legal.
2. Alice moves that conversation into a team project. `moveConversationToProject` sets `spaceId` and overwrites `requestedSpaceIds` with `[project]`, dropping Legal. The selection row survives.
3. The conversation is now readable by all project members, including Bob who has no Legal access.
4. Alice posts. The agent loop runs under Alice's authenticator, `getValidSelectedSpaceIdsForAgentRun` still returns Legal (it only checked the feature flag, `canRead(auth)` and `isRegular()`), so the agent retrieves Legal documents into the conversation.
5. Bob reads them.

The picker is hidden in pod conversations, so there is no UI signal that Spaces are still attached.

### Fix

**Part A, runtime guard.** `getValidSelectedSpaceIdsForAgentRun` returns `[]` for pod conversations, mirroring what `updateConversationRequirementsForSkills` already does for the same reason.

**Part B, clear selections on move into a project.** `moveConversationToProject` soft-removes the conversation's selections, inside the same transaction that resets the requirements to `[project]`.

### Why Part A alone is not enough

If the conversation is later moved back **out** of the project, `moveConversationOutOfProject` calls `rebuildConversationRequirements`, which clears `requestedSpaceIds` and rebuilds it from agents and content fragments only. It does not re-add selected Spaces. The conversation is no longer a pod conversation, so the Part A guard stops applying and the stale selections become live again, now with no ACL backing at all. Same vulnerability, one move later.

I verified this: with only Part A applied, the round-trip test still fails.

### Clearing vs. re-including in `rebuildConversationRequirements`

I went with clearing. Re-including active selections when rebuilding requirements would keep the ACL and the runtime scope in sync in the move-out direction, but it silently re-restricts a conversation that has been visible to the whole project in the meantime, and it leaves a window (while the conversation sits in the project) where a selection exists with no ACL behind it. Clearing removes the ambiguity in one place, at the moment the ACL is actually reset, and keeps the invariant "an active selection always has a matching `requestedSpaceIds` entry" true at all times. Users can re-select the Space after moving the conversation out.

Together, Parts A and B make "a pod conversation never has a live selection" hold in both directions. `updateSpaceId` has exactly two callers: `moveConversationToProject`, which now clears, and `moveConversationOutOfProject`, which can only run on a conversation that already went through the former. Every write path that could add a selection already rejects pod conversations, and conversations created directly in a project are gated by `validateSelectableSpaces({ podId })`.

### Resource change

`ConversationSelectedSpaceResource.removeForConversation` takes a `SpaceResource[]`. Enumerating the selected Spaces goes through `listActiveSpacesByConversation`, which hydrates them via `SpaceResource.fetchByModelIds` and therefore drops Spaces the mover cannot read, which are exactly the rows we most need to remove. Added `removeAllForConversation`, which soft-removes every active row of the conversation without requiring the caller to read the Spaces first.

`removeForConversation` is kept: the `selected_spaces` route only accepts `mode: "add"` today, so it has no production caller yet, but it is the counterpart of `upsertForConversation` for the deselect path and is covered by the resource's own tests. It now carries a comment saying so, so it does not get pruned as dead code.

### Side benefit: sub-agent runs in pods stop failing

Part A also fixes a latent user-facing breakage. `run_agent` creates its child conversation with `spaceId: mainConversation.spaceId`, so a pod parent produces a pod child, and `copySelectedConversationSpacesToChild` then called `addSelectedConversationSpaces` on that pod child. That write path rejects pod conversations, so the error propagated and the whole sub-conversation setup failed with "Failed to inherit selected Spaces".

With the guard in place, `getValidSelectedSpaceIdsForAgentRun` returns `[]` for the pod parent, `copySelectedConversationSpacesToChild` short-circuits to `Ok`, and the sub-agent run proceeds normally.

## Tests

`front/lib/api/assistant/conversation/selected_spaces.test.ts`:

- `getEffectiveSpaceIdsForAgentRun` returns only the agent's own Spaces for a pod conversation that has active selections.
- Moving a conversation into a project removes its active selections (row soft-removed, `removedAt` set).
- Round trip: select, move in, move out, and the Space is neither in `requestedSpaceIds` nor in the effective runtime scope.

All three fail on `main` and pass here. The round-trip one is what proves Part A alone is insufficient: with only the runtime guard applied it still leaks the Space into `getEffectiveSpaceIdsForAgentRun`.

Typecheck clean, `npm run test` green on `selected_spaces.test.ts`, `projects/conversations.test.ts` and `conversation_selected_space_resource.test.ts` (41 tests).

## Risk

Low.

- The whole feature is behind `restricted_spaces_in_input_bar`, `stage: "dust_only"`. Workspaces without the flag never reach any of this code.
- Production currently has 10 rows in `conversation_selected_spaces`, all active, across a single workspace, and **zero** of them belong to a conversation with a non-null `spaceId`. So no conversation is in the affected state today, and after this ships none can enter it: move-in clears, and every selection write path already rejects pod conversations.
- Both parts only ever *narrow* scope. Part A returns `[]`, Part B soft-removes rows. Nothing widens access, so the failure mode of a bug here is "a selected Space is ignored when it should have been honoured", not a leak.
- Rollback-safe: reverting restores the previous behaviour. `removeAllForConversation` is a soft delete (`removedAt`), so rows cleared in the meantime are still on disk and could be reactivated by clearing `removedAt` if we ever needed to.

## Deploy Plan

- No backfill: the only rows a backfill would target are active selections on conversations already in a project, and there are none in production (0 of 10 rows). Part A would keep any such row inert anyway.
- Nothing to coordinate with clients: no API schema change, and the two affected endpoints (`selected_spaces`, `selectable_spaces`) keep their contracts.

Two sibling PRs, `spaces-initiator-gate` (#29374) and `spaces-acl-race` (#29376), also touch `front/lib/api/assistant/conversation/selected_spaces.ts` and will likely need a trivial rebase depending on merge order. Base here stays `main`.

### DIFF
--- a/front/lib/api/assistant/conversation/selected_spaces.test.ts
+++ b/front/lib/api/assistant/conversation/selected_spaces.test.ts
@@ -19,6 +19,10 @@ import {
   type SelectedConversationSpacesError,
   validateSelectableSpaces,
 } from "@app/lib/api/assistant/conversation/selected_spaces";
+import {
+  moveConversationOutOfProject,
+  moveConversationToProject,
+} from "@app/lib/api/projects/conversations";
 import { Authenticator } from "@app/lib/auth";
 import { ConversationResource } from "@app/lib/resources/conversation_resource";
 import { ConversationSelectedSpaceResource } from "@app/lib/resources/conversation_selected_space_resource";
@@ -32,7 +36,10 @@ import { createResourceTest } from "@app/tests/utils/generic_resource_tests";
 import { MembershipFactory } from "@app/tests/utils/MembershipFactory";
 import { SpaceFactory } from "@app/tests/utils/SpaceFactory";
 import { UserFactory } from "@app/tests/utils/UserFactory";
-import type { ConversationWithoutContentType } from "@app/types/assistant/conversation";
+import {
+  type ConversationWithoutContentType,
+  isPodConversation,
+} from "@app/types/assistant/conversation";
 import type { Result } from "@app/types/shared/result";
 import type { UserType, WorkspaceType } from "@app/types/user";
 
@@ -173,6 +180,12 @@ describe("selected conversation Spaces", () => {
     const otherUser = await UserFactory.basic();
     await MembershipFactory.associate(workspace, otherUser, { role: "user" });
     return Authenticator.fromUserIdAndWorkspaceId(otherUser.sId, workspace.sId);
+  }
+
+  async function memberProjectSpace() {
+    const space = await SpaceFactory.project(workspace, user.id);
+    await addCurrentUser(space);
+    return space;
   }
 
   it("rejects selected Spaces when the feature flag is disabled", async () => {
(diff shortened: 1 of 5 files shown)


Fraggle (approved):
FYI, won't stay a problem to fix once we update the pod conversations access check to spaceId only (as decided with stan).

fontanierh:
agreed @Fraggle. I expect this to go live first so still worth merging I think

PR #29376, opened 2026-07-23
Make conversation Space ACL merge atomic

fontanierh:
## Description

Attaching a Space from the input bar (`restricted_spaces_in_input_bar`) does two things:

1. it appends the Space to `conversation.requestedSpaceIds`, a **conjunctive** ACL where a viewer must have read access to *every* listed Space (the `.every(...)` filter in `ConversationResource.baseFetchWithAuthorization`);
2. it widens the agent's runtime scope for that conversation, which is what reaches real knowledge retrieval via `DataSourceViewResource.listBySpaceIds(auth, effectiveSpaceIds, ...)` in `skill_actions.ts`.

(1) is the only thing that makes (2) safe. A Space that is in the agent's runtime scope but missing from `requestedSpaceIds` means restricted data flows into a conversation that does not restrict its readers.

`addSelectedConversationSpaces` computed the new ACL from the **caller's snapshot** of the conversation and handed the precomputed array to `ConversationResource.updateRequirements`, which *overwrites* the column and never merges with current DB state. The picker fires one POST per checkbox toggle with no in-flight serialization, and the dropdown stays open on select, so ticking several Spaces quickly produces overlapping requests. Under READ COMMITTED (we set no isolation level in `withTransaction`) they interleave:

- request 1 snapshots `[A]`, computes `[A, X]`, writes `[A, X]`
- request 2 snapshots `[A]`, computes `[A, Y]`, writes `[A, Y]`

Final ACL is `[A, Y]`, losing X, while **both** selection rows committed and are active. The agent then has X in scope with no ACL requirement backing it, so users without access to X can read a conversation containing X's data. This is reachable from ordinary UI use, not just a crafted request.

### Fix

- New `ConversationResource.appendRequestedSpaceIds(auth, sId, spaceModelIds, transaction)` (instance + static pair, mirroring `updateRequirements`). It re-reads the conversation row with `SELECT ... FOR UPDATE` **inside the caller's transaction**, unions the locked value with the new ids, writes the result, and returns the persisted requirement set as space sIds. Concurrent appends now serialize on the row, so the selection rows and `requestedSpaceIds` can no longer disagree.
- `addSelectedConversationSpaces` uses it in place of the snapshot-based computation, and the lock is taken before the selection rows are upserted.
- `effectiveAcl.spaceIds` in the response is now the actual post-merge ACL instead of the optimistic locally-computed union. The message-post and conversation-create endpoints feed that value straight back into their local `conversation` object, so it has to be correct.
- Latent bug in the existing static `updateRequirements`: it called `fetchById` **without** the transaction, so the read ran on another connection and did not see the enclosing transaction. The transaction is now passed through (`FetchConversationOptions.transaction`, threaded into the conversation, space and participation reads in `baseFetchWithAuthorization`). It is opt-in, so every existing caller is unchanged. All callers of `updateRequirements` were checked (`permissions.ts` x2, `skill_permissions.ts`, `selected_spaces.ts` and the tests) and all of them are strictly better off reading inside their own transaction.

## Tests

`front/lib/api/assistant/conversation/selected_spaces.test.ts`:

- two successive adds of different Spaces both end up in `requestedSpaceIds` (merge semantics);
- **regression test**: an add performed against a stale conversation snapshot (one whose `requestedSpaceIds` predates another add) no longer drops the previously added Space, and the active selection rows and the ACL agree. Verified it fails on the pre-fix code with `expected [ 'vlt_3TWT…' ] to deeply equal [ 'vlt_2hYM…', 'vlt_3TWT…' ]`. The harness runs inside a CLS transaction, so a genuinely parallel-connection test is not expressible here; the stale snapshot is the reliable way to encode the interleaving;
- `effectiveAcl.spaceIds` matches what is actually persisted on the conversation.

Ran `selected_spaces`, `permissions`, `skill_permissions`, `conversation_resource`, `conversation_selected_space_resource`, plus everything under `lib/api/assistant/conversation` and `lib/api/projects`, for 403 tests green. `npx tsgo --noEmit` clean.

## Risk

This is the widest-reaching of the three Space-selection fixes, so two things are worth a reviewer's attention:

- **`FetchConversationOptions.transaction` is opt-in.** Every existing caller passes nothing and is therefore unchanged. It is threaded into all three reads in `baseFetchWithAuthorization` (the conversation, the spaces, the participations) rather than just the first, because threading only the conversation read would leave the space and participation reads able to filter out a conversation whose rows are still uncommitted.
- **The new row lock is taken inside the caller's transaction**, immediately before the selection rows are written, and released on commit. The conversation row is the only row locked.

The feature itself is behind `restricted_spaces_in_input_bar` (`stage: "dust_only"`), so it is only reachable internally today. Rollback-safe: no schema change, and reverting restores the previous (racy) merge without leaving data in a new shape.

## Deploy Plan

- No backfill. For context, the US production read replica currently has 10 rows in `conversation_selected_spaces`, all active, across a single workspace, consistent with the flag being `dust_only`.
- No API schema change; the affected endpoints keep their contracts.

### Follow-ups (not in this PR)

`updateConversationRequirements` (`conversation/permissions.ts`) and `updateConversationRequirementsForSkills` (`conversation/skill_permissions.ts`) have the same read-modify-write shape against a caller-held snapshot. They are additive and not driven by concurrent user toggles, so I kept them out of this PR, but they are the obvious candidates to adopt `appendRequestedSpaceIds` next.

Two sibling PRs, `spaces-initiator-gate` (#29374) and `spaces-pod-runtime-guard` (#29375), also touch `front/lib/api/assistant/conversation/selected_spaces.ts` and will likely need a trivial rebase depending on merge order. Base here stays `main`.

### DIFF
--- a/front/lib/api/assistant/conversation/selected_spaces.test.ts
+++ b/front/lib/api/assistant/conversation/selected_spaces.test.ts
@@ -188,6 +188,14 @@ describe("selected conversation Spaces", () => {
     return space;
   }
 
+  async function persistedRequestedSpaceIds(conversationId: string) {
+    const { requestedSpaceIds } = await refetchConversation(
+      auth,
+      conversationId
+    );
+    return [...requestedSpaceIds].sort();
+  }
+
   it("rejects selected Spaces when the feature flag is disabled", async () => {
     const restrictedSpace = await memberRestrictedSpace();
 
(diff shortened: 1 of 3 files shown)


Fraggle (approved):
LGTM

PR #9774, opened 2025-01-06
Decrease sequelize pool from 30 to 20

spolu:
## Description

To allow scaling of workers.

## Risk

Low

## Deploy Plan

- deploy `front`

PR #9477, opened 2024-12-18
[front] update data sources list after creation

tdraier:
## Description

Reload the list of datasources after creating a webcrawler, to avoid allowing 2 creations with the same name.
Moved the useDataSources call directly where it's used
Also fixed the updateUrl to get the last updated value.

related logs https://app.datadoghq.eu/logs?query=%22Unhandled%20API%20Error%22&agg_m=count&agg_m_source=base&agg_t=count&cols=host,service&fromUser=true&link_source=monitor_notif&messageDisplay=inline&storage=hot&stream_sort=desc&viz=stream&from_ts=1734454806000&to_ts=1734455106000&live=false

## Risk

none

## Deploy Plan

deploy front

PR #29379, opened 2026-07-23
[front] Load workspace permissions as a resourceId-aware set in the Authenticator

tdraier:
## Description

Foundation of the space-permissions migration ([#9477](https://github.com/dust-tt/tasks/issues/9477) / [#9774](https://github.com/dust-tt/tasks/issues/9774)): give the `Authenticator` a small, stable authorization surface, and model a resource's access rules as a single **flat `AccessControlList`** that the resource builds and the auth evaluates.

```
#29379 (this)  →  #29619 backfill  →  #29639 shadow  →  #29763 flip
```

### Responsibility split
- **`Authenticator`** knows only the *caller* — role, group memberships, and (preloaded once at construction) the caller's `group_permissions` grants. It evaluates a fully-resolved ACL and answers governance lookups; it knows nothing about any specific resource's rules.
- **A resource** knows its *own* rules — it builds an `AccessControlList` from role rules and/or the caller's governance grants.

### What it does

**One flat type + one interface** (`types/resource_permissions.ts`):
```ts
type GroupGrant = { id: ModelId; permissions: GrantVerb[] };
type RoleGrant  = { role: RoleType; permissions: GrantVerb[] };
type AccessControlList = { roles: RoleGrant[]; groups: GroupGrant[]; workspaceId: ModelId };

interface WithAccessControl {
  getAccessControlLists(auth: Authenticator): AccessControlList[];
}
```
The `ResourcePermission` union (and `PermissionType`) are gone — everything speaks `GrantVerb`. A resource implements `WithAccessControl`; `auth` is passed so it can build its ACL from the caller's governance grants (`auth.getGroupPermissions`) — needed for the shadow candidate and the flip.

**`GroupPermissions`** (`group_permission_registry.ts`) — the caller's grants, group-granular and indexed by `resourceType → resourceId → groupId → verb-bitmask`, resolved once at auth construction (replaces `PermissionSet`). Serializable (`toJSON`/`fromJSON`) so it round-trips through a serialized `Authenticator` (Temporal) without re-querying.

**Authenticator surface:**
- `hasPermission(verb, target)` — pure/sync: the caller holds `verb` on **every** ACL the `WithAccessControl` target declares. Role path (workspace-gated) OR group-membership path.
- `hasPermissionForAll(verb, targets)` — conjunction across targets.
- `hasPermissionForAcls(verb, acls)` — the raw-ACL primitive, for callers that already hold built or derived ACLs (a space's served ACLs, the cross-space conversation/agent/skill checks) rather than a `WithAccessControl` target.
- `getGroupPermissions(resourceType, resourceId)` — the caller's matching `group→verbs`, folding in the type-wide (`-1`) grants. Caller-scoped.
- `hasWorkspacePermission(verb, resourceType)` — thin wrapper over the ACL check against a handmade type-wide ACL whose synthetic admin role gives admins every workspace capability by default.
- `getWorkspacePermissions()` — admins → all; others derive from their type-wide grants.

The `Authenticator.canRead` / `canWrite` / `canAdministrate` convenience methods are gone — resources keep their own thin `canX(auth)` wrappers that call `hasPermission(verb, this)` (or `hasPermissionForAcls` for derived ACLs).

**Resources** expose `getAccessControlLists(auth)` (returning `AccessControlList[]`, renamed from `requestedPermissions()`); consuming `getGroupPermissions` + moving shadow into the resource come in the shadow/flip PRs.

## Risk
Low / behavior-preserving. No space populates the table yet (starts in #29619), so the served result is unchanged. Admin semantics preserved (all workspace capabilities; no implicit instance access).

## Tests
`group_permission_*`, `user`, `space_resource`, `conversation_resource` suites pass; front-api permission suites pass; `tsgo` clean.

### DIFF
--- a/front/lib/auth.ts
+++ b/front/lib/auth.ts
@@ -1524,13 +1524,13 @@ export class Authenticator {
         return true;
       }
 
-      // Second path: group permissions check (the `group_permissions` table). When the resource
-      // declares a `resourceType`, the caller passes if they hold the requested permission — used
-      // directly as a grant verb, since `PermissionType` ⊆ `GrantVerb` — on this
-      // `(resourceType, resourceId)`. A type-wide grant satisfies the check for any instance (see
-      // `PermissionSet.has`).
+      // Second path: group permissions check (the `group_permissions` table). A
+      // GroupResourcePermission declares a `resourceType`; the caller passes if they hold the
+      // requested permission — used directly as a grant verb, since `PermissionType` ⊆ `GrantVerb`
+      // — on this `(resourceType, resourceId)`. A type-wide grant satisfies the check for any
+      // instance (see `PermissionSet.has`).
       if (
-        resourcePermission.resourceType !== undefined &&
+        "resourceType" in resourcePermission &&
         workspace.id === resourcePermission.workspaceId &&
         this._permissions.has(
           resourcePermission.resourceType,
@@ -1543,11 +1543,15 @@ export class Authenticator {
     }
 
     // Third path: legacy group permissions check (inline groups).
-    return this._groupModelIds.some((groupId) =>
-      resourcePermission.groups.some(
-        (gp) => gp.id === groupId && gp.permissions.includes(permission)
-      )
-    );
+    if ("groups" in resourcePermission) {
+      return this._groupModelIds.some((groupId) =>
+        resourcePermission.groups.some(
+          (gp) => gp.id === groupId && gp.permissions.includes(permission)
+        )
+      );
+    }
+
+    return false;
   }
 
   canAdministrate(resourcePermissions: ResourcePermission[]): boolean {
--- a/front/lib/auth.ts
+++ b/front/lib/auth.ts
@@ -57,10 +57,7 @@ import { WHOLE_TYPE_RESOURCE_ID } from "@app/types/group_permissions";
 import type { GroupKind } from "@app/types/groups";
 import type { PlanType, SubscriptionType } from "@app/types/plan";
 import type { ProvidersHealth } from "@app/types/provider_credential";
-import type {
-  PermissionType,
-  ResourcePermission,
-} from "@app/types/resource_permissions";
+import type { ResourcePermission } from "@app/types/resource_permissions";
 import { hasRolePermissions } from "@app/types/resource_permissions";
 import { isDevelopment } from "@app/types/shared/env";
 import type { WhitelistableFeature } from "@app/types/shared/feature_flags";
@@ -1225,8 +1222,9 @@ export class Authenticator {
 
   /**
    * Whether the caller holds a workspace-level capability, asked as a type-level verb (e.g.
-   * "create" on "agent"). Admins hold every workspace capability by default; everyone else derives
-   * it from the grants resolved at construction (which fold in "*" wildcard grants).
+   * "create" on "agent"). A thin wrapper over `hasPermission` against a type-wide
+   * `GroupResourcePermission`: the synthetic admin role grants admins every capability by default,
+   * and everyone else derives it from their type-wide grants (which fold in "*" wildcard grants).
    */
   async hasWorkspacePermission(
     verb: GrantVerb,
@@ -1240,17 +1238,18 @@ export class Authenticator {
       `Verb "${verb}" is not allowed (no type-level role grants it) on resource type "${resourceType}".`
     );
 
-    if (!this.workspace()) {
+    const workspace = this.workspace();
+    if (!workspace) {
       return false;
     }
 
-    // Admins hold every workspace-wide capability by default (this is where role-based access is
-    // layered on top of the grant set; the PermissionSet itself holds only grants).
-    if (this.isAdmin()) {
-      return true;
-    }
-
-    return this._permissions.hasTypeWide(resourceType, verb);
+    // A type-wide target (no resourceId). The admin role carries `verb` so admins pass by role;
+    // everyone else falls through to the group_permissions type-wide grant.
+    return this.hasPermission(verb, {
+      roles: [{ role: "admin", permissions: [verb] }],
+      resourceType,
+      workspaceId: workspace.id,
+    });
   }
 
   /**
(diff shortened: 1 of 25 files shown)


tdraier:

- #29379 ◀
    - #29619
        - #29639
            - #29763

front/lib/resources/group_permission_registry.ts
```diff
@@ -174,55 +181,220 @@ function allTypeLevelVerbsForResource(
...
+    }
+    const byId = this.grants.get(resourceType);
+    if (!byId) {
+      return false;
+    }
+    const bit = VERB_BIT.get(verb) ?? 0;
+    return (
+      ((byId.get(resourceId) ?? 0) & bit) !== 0 ||
+      ((byId.get(WHOLE_TYPE_RESOURCE_ID) ?? 0) & bit) !== 0
+    );
```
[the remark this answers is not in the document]
tdraier: no, that's fine.

front/types/group_permissions.ts
```diff
@@ -10,6 +10,14 @@
...
 // Verbs are the granular capabilities a role grants. They are never stored — roles map to them in
 // the registry, and capability questions are expressed in verbs.
+//
+// A verb's position in this array is its bit in the `PermissionSet` bitmask (verb at index i => bit
+// `1 << i`; see `group_permission_registry`). Consequences:
+// - Never reorder or remove a verb: it would silently reassign every subsequent verb's bit and
+//   change the meaning of already-computed masks.
+// - Add new verbs only at the end.
+// - Keep the total at most 31 verbs: JS bitwise operators work on 32-bit integers, so beyond that
+//   the shift overflows and masks break.
```
philipperolet, line 20: a bit wary of these constraints that only lie in comment, + the limitation to 32 verbs

storing each verb as a short int (8 bytes), since we have max 3-4 verbs by grant, would be most of the time smaller than the bitmask + allow for 256 verbs

or we could map verbs to resources and then 8 bytes would suffice + 

and/or we could try to enforce the constraints (max 32 items, no reorder, etc.) in code?

in any case, if it's easy to change the storage afterwards, we can move forward and rethink when the time comes

front/types/resource_permissions.ts
```diff
@@ -37,11 +38,27 @@ export type GroupResourcePermissions = {
...
+ * @property roles - Role-based grants: a caller whose workspace role matches gets its permissions
+ * @property resourceType - The governance resource domain to check (e.g. "space")
+ * @property resourceId - The resource's model id; omitted means the type-wide (-1) grant
+ * @property workspaceId - The resource's workspace; role and governance-grant checks only apply
+ *   when it matches the caller's workspace
  */
 export type CombinedResourcePermissions = {
   groups: GroupPermission[];
   roles: RolePermission[];
+  resourceType?: ConcreteResourceType;
```
philipperolet: checking, is that temporary? 
afaict combinedresourcepermissions and requestedPermissions are used to do auth.canWrite; and in the new system we remove auth.canWrite and do auth.hasPermission(...) not using combinedResourcePermissions anymore right?

philipperolet: asking because if it's not, it might be a good idea to think through the comment / the solution as it's becoming a mixed bag

front/lib/resources/group_permission_registry.ts
```diff
@@ -174,55 +181,222 @@ function allTypeLevelVerbsForResource(
...
+ * The governance grants a caller holds, resolved once at auth construction. Grants are keyed by
+ * (resourceType, resourceId): resourceId is WHOLE_TYPE_RESOURCE_ID (-1) for type-wide capabilities
+ * (e.g. "create" on "agent") or a resource's model id for instance grants (e.g. "write" on a
+ * specific space). A type-wide grant satisfies an instance check on any id of that type.
+ */
+export class PermissionSet {
+  private constructor(
+    // resourceType -> resourceId -> verb bitmask. resourceId WHOLE_TYPE_RESOURCE_ID (-1) is the
+    // type-wide entry. Values are bitmasks (see VERB_BIT), not Sets, to keep the footprint small.
+    private readonly grants: Map<ConcreteResourceType, Map<number, number>>,
```
philipperolet: I think during the discussion with @id13 we had said that using maps might be heavy in memory?
To be digged; if not really heavy we can keep as is, otherwise we can use arrays of tuples as the search should be fast (we can optimize by sorting although IMO not necessary) since there are 2 levels of arrays, so array size will remain small

front/lib/resources/group_permission_registry.ts
```diff
@@ -174,55 +181,222 @@ function allTypeLevelVerbsForResource(
...
+    // Admins hold every capability on every resource/instance by default.
+    private readonly isAdminAll: boolean
```
philipperolet: IMO it would be best to remove that flag and simplify, removing all role-related logic from permission registry. 
- In multiple cases we need to resolve the actual permissions of an admin so we need to load them--e.g. in spaces we show "you are not a member of this space" or in agents "you need to become an editor to edit this agent".
- In general, leaving the possibility that admins ma have absolutely all permissions granted on everything is IMO sound, this might happen again;
- Separating role-based and permission-based logic is clearer IMO; we have isAdmin() checks in the code when we want to use the admin role gate

philipperolet: I see below we distinguish between type level (admin check) or instance level (no admin check) which handles the first bullet point, but makes the 3rd one stronger (deeper entanglement between role & permission logics)

front/lib/auth.ts
```diff
@@ -1354,25 +1433,30 @@ export class Authenticator {
...
    * 1. Role-based permission check:
    *    Applies when the resource has role-based permissions configured.
    *    Permission is granted if:
    *    - The resource has public access (role="none") for the requested permission, OR
    *    - The user's role has the required permission AND the resource belongs to user's workspace
    *
-   * 2. Group-based permission check:
+   * 2. Governance-grant check:
+   *    Applies when the resource declares a `resourceType`.
+   *    Permission is granted if the user holds the requested permission as a grant verb on the
```
philipperolet: Is that temporary too? I would assume that short term we can dispose of 3. group based permission checks, and medium term we can remove the "old" way to check permissions entirely?

philipperolet (commented):
Thanks for the work, heavy one :) 🏋️‍♂️ 
Left a few coms, happy to discuss IRL

front/lib/auth.ts
```diff
@@ -875,24 +915,44 @@ export class Authenticator {
       workspaceSubscription
     );
 
+    // If the key is associated with the workspace, we associate the groups.
+    const workspaceGroupModelIds = isKeyWorkspace
+      ? allGroups.map((g) => g.id)
+      : [];
+    const keyGroupModelIds = allGroups.map((g) => g.id);
+
+    const [permissions, keyPermissions] = await Promise.all([
```
philipperolet, line 966: I think here in a common case in which isKeyWorkspace == true and role != admin the sets are always the same? as it's common we might want to avoid computing twice in that case?

front/lib/auth.ts
```diff
@@ -875,24 +915,44 @@ export class Authenticator {
...
+      this.resolvePermissions({
+        workspace,
+        role,
+        groupModelIds: workspaceGroupModelIds,
+      }),
+      this.resolvePermissions({
+        workspace: keyWorkspace,
+        role: "builder",
```
philipperolet, line 974: I'm assuming the intent is to remove "builder" in follow-up PRs to keep this one contained

front/lib/resources/group_permission_registry.ts
tdraier: On this I asked the agent and it actually said it was better to use Map than objects . to be confirmed indeed

front/lib/resources/group_permission_registry.ts
tdraier: globally agree. 
1/ we'll have to load them indeed (to be fixed, currently we dont) 
2/ for now we assumed that hasWorkspacePermission is always true for admin, if it's not we would have to update all callers to make an isAdmin || hasWorkspacePermission ?
3/ authenticator can centralize role-based and permission-based , that's actually what hasResourcePermission is doing ( first checking the role, then the group/governance permission ). So then I agree group-permission should not handle the role again

front/lib/auth.ts
tdraier: It's a bit confusing, Actually it's the 3/ which is temporary - 3 is the legacy permissions based on groups ( the vault_group, agent_group, ...) . What we want to keep is 1/ role-based 2/ new-group-based (what i called governance here .. ) . The 3 is going away in a following PR ( after backfill / replacement).

front/types/group_permissions.ts
tdraier: so no bitmask ? list of short int is definitely ok , 3 or 4 verbs will indeed be smaller than one 32 bits number.

front/types/resource_permissions.ts
tdraier: groups is temporary (legacy). resourceType/resourceId could be grouped into one "groupPermission" object , so different from legacy "groups". This reflect the 2 levels - roles + group permissions

tdraier: I split in different types, avoid mixing legacy and new

front/lib/resources/group_permission_registry.ts
tdraier: removed

front/lib/auth.ts
tdraier: clarified - named "legacy" the old group permission

front/types/group_permissions.ts
fabiencelier: I think I prefer the short int implementation, we can just harcode them and make sure in code they are all different 
like
"read" = 1;
"write" = 2;
...
Will implement this

fabiencelier: Could not find how to improve this
There is no short int in javascript right ?

front/lib/auth.ts
fabiencelier: done

front/lib/resources/group_permission_registry.ts
fabiencelier: > would it be more compact to store the grants as an array of tuples instead of a map ?

Yes for memory, no for lookups — and only if the tuples are stored in a typed array. The nuance
  matters, so concretely, for ~5,000 instance grants (one group per resource, one or two resource
  types):

  ```
┌─────────────────────┬─────────┬───────────────────────────────────┬───────────────────────┐
  │        shape        │ memory  │               JSON                │ lookup by (type, id)  │
  ├─────────────────────┼─────────┼───────────────────────────────────┼───────────────────────┤
  │ nested Maps (today) │ ~0.5–1  │ {"space":{"12":{"7":3},…}} ≈ 12   │ O(1)                  │
  │                     │ MB      │ B/grant                           │                       │
  ├─────────────────────┼─────────┼───────────────────────────────────┼───────────────────────┤
  │ number[][] tuples   │ ~400 KB │ [[0,12,7,3],…] ≈ 12 B/grant       │ O(n) scan             │
  ├─────────────────────┼─────────┼───────────────────────────────────┼───────────────────────┤
  │ flat Int32Array,    │ ~80 KB  │ poor (needs encode/decode)        │ O(n) scan, or O(log   │
  │ stride 4            │         │                                   │ n) if sorted          │
  └─────────────────────┴─────────┴───────────────────────────────────┴───────────────────────┘
```

  The dominant cost today is one Map per resourceId (byId.set(resourceId, byGroup) in fromGrants).
  A V8 Map is ~150 B once you count the object plus its minimum-capacity backing table, so 5,000
  resource ids means 5,000 tiny hash tables each holding a single 8-byte entry. That's the ~20×
  overhead ratio.

  An array of tuples removes it — but a plain number[][] gives back most of the saving in per-tuple
  array headers (~56 B each). Only one contiguous buffer (a flat Int32Array with stride 4, or four
  parallel arrays) collapses a grant to its actual 16 bytes.

  The cost is the access pattern. forResource(type, id) is currently two hash lookups. Over a flat
  array it's a scan, and permission checks are not one-per-request — filtering a list of 1,000
  skills does 1,000 checks, so you'd turn a linear pass into 1,000 × 5,000 = 5M comparisons, which
  is exactly the quadratic pattern [GEN7] tells us to avoid. Sorting by (resourceType, resourceId)
  and binary searching gets it to O(log n), at the cost of hand-rolled search code over an untyped
  integer buffer.

  If it ever needs doing, the shape that keeps both properties is: store flat (compact, cheap to
  serialize as plain int arrays), and build the Map index lazily on the first forResource call.
  Requests that only ask about a handful of resources never pay for the index; the
  serialized-Authenticator payload stays small either way.

  My recommendation is to leave it. The compaction only pays off in the thousands-of-grants regime,
  and in that regime the bigger lever is the one I flagged earlier: not shipping the grant map
  through Temporal at all (fromJSON already does three DB round-trips, so re-resolving from
  groupIds is nearly free and bounds the payload). I'd want a real measurement from a large
  workspace before trading Map clarity for a stride-4 integer buffer — and per [GEN2], the current
  shape is the simple one that's obviously correct.

front/lib/auth.ts
fabiencelier: removed

fabiencelier (approved):
LGTM now

matteotrab (approved):
LGTM

front/types/group_permissions.ts
tdraier: There's `Int8Array` which could be used here

PR #29380, opened 2026-07-23
[front] Replace builder-role file write with the space.globalWrite capability

tdraier:
Add the type-level `globalWrite` capability on `space` and use it in place of the
legacy `builder` role: space requestedPermissions (global/conversations/open) grant
read+write to its holders, the file endpoints (private + v1) gate on
`hasWorkspacePermission("globalWrite", "space")` instead of `isBuilder()`, and the
capability is seeded to the Builders group so it lands on today's builder population.

Part of the builder-role deprecation (dust-tt/tasks#9746).


tdraier:

- #29379
    - #29380 ◀

PR #29382, opened 2026-07-23
admin governance: expose availability in skill endpoints 2/3

fabiencelier:
## Description

POST and PATCH `/skills` accept an optional `availability` param (3 states) with the deprecated `isDefault` alias still supported — an explicit availability wins — and `toJSON` now exposes `availability`.
With the `admin_governance_skill_publication` flag: explicitly creating a published skill or changing availability requires the skill publish permission; without the flag, only the two legacy values are accepted (previous behavior). The default availability is unchanged in both modes.
Issue: https://github.com/dust-tt/tasks/issues/9730

## Stack

- 1/3: https://github.com/dust-tt/dust/pull/29381
- 2/3: https://github.com/dust-tt/dust/pull/29382
- 3/3: https://github.com/dust-tt/dust/pull/29383

## Tests

Route tests: availability/isDefault priority rule, `editors` rejected flag-off, unpublished default flag-on, publish-permission 403s on create and update, editor edits unaffected when availability unchanged.

## Risk

Low, PR is easy to revert.

## Deploy Plan

Deploy front.

### DIFF
--- a/front-api/public/swagger.json
+++ b/front-api/public/swagger.json
@@ -14271,7 +14271,17 @@
           },
           "isDefault": {
             "type": "boolean",
-            "description": "Whether this skill is enabled by default"
+            "deprecated": true,
+            "description": "Whether this skill is enabled by default. Deprecated, use availability instead."
+          },
+          "availability": {
+            "type": "string",
+            "enum": [
+              "editors",
+              "workspace_users",
+              "users_and_agents"
+            ],
+            "description": "Who the skill is available to (users_and_agents makes it discoverable by agents)"
           },
           "instructions": {
             "type": "string",
--- a/front-api/routes/w/[wId]/skills/index.ts
+++ b/front-api/routes/w/[wId]/skills/index.ts
@@ -4,6 +4,7 @@ import {
   getReferencedSkillSpaceModelIds,
   resolveAdditionalRequestedSpaceModelIds,
 } from "@app/lib/api/skills/space_requirements";
+import { hasFeatureFlag } from "@app/lib/auth";
 import { DataSourceViewResource } from "@app/lib/resources/data_source_view_resource";
 import { FileResource } from "@app/lib/resources/file_resource";
 import { MCPServerViewResource } from "@app/lib/resources/mcp_server_view_resource";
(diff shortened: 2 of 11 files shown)


front-api/public/swagger.json
```diff
@@ -14271,7 +14271,17 @@
           },
           "isDefault": {
             "type": "boolean",
-            "description": "Whether this skill is enabled by default"
+            "deprecated": true,
```
ykmsd, line 14274: so IIUC it says deprecated but you can still pass isDefault right?

front-api/routes/w/[wId]/skills/index.ts
```diff
@@ -75,6 +81,28 @@ const PostSkillRequestBodySchema = z.intersection(
...
+  { hasSkillPublicationGovernance }: { hasSkillPublicationGovernance: boolean }
+): SkillAvailability {
+  if (availability !== undefined) {
+    return availability;
+  }
+  if (isDefault !== undefined) {
+    return availabilityFromIsDefault(isDefault);
+  }
+  // With skill publication governance, new skills start unpublished (editors-only).
+  return hasSkillPublicationGovernance ? "editors" : DEFAULT_SKILL_AVAILABILITY;
```
Comment (ykmsd) [inline front-api/routes/w/[wId]/skills/index.ts]: do we already want to set it to editors if there is a flag on? This means our workspace skills will start with editors already? 🤔

ykmsd (approved):

front-api/public/swagger.json
fabiencelier: Yes, output is the same as before + new availability field
We mark it as deprecated so people don't start new projects relying on this field, we will remove in a few months

front-api/routes/w/[wId]/skills/index.ts
Comment (fabiencelier) [reply front-api/routes/w/[wId]/skills/index.ts]: will return DEFAULT_SKILL_AVAILABILITY and I'll change the default later

Comment (ykmsd) [reply front-api/routes/w/[wId]/skills/index.ts]: wait actually this is the new flag which we haven't activated on our workspace so actually it's safe 🙈? We can activate it in our workspace once it's ready??

Comment (fabiencelier) [reply front-api/routes/w/[wId]/skills/index.ts]: Yes it's the new flag not yet activated
I'll make another PR to fix the default value everywhere, not just here

PR #29383, opened 2026-07-23
admin governance: expose availability in skill endpoints 3/3

fabiencelier:
## Description

The skill builder form holds and submits `availability` instead of the deprecated `isDefault`.
`SkillsContext`, `ManageSkillsPage` and the poke skill table read `availability` from the serialized skill; the front-end no longer uses `isDefault`.
Issue: https://github.com/dust-tt/tasks/issues/9730

## Stack

- 1/3: https://github.com/dust-tt/dust/pull/29381
- 2/3: https://github.com/dust-tt/dust/pull/29382
- 3/3: https://github.com/dust-tt/dust/pull/29383

## Tests

No new tests: pure representation change covered by the existing skill builder flows and the route tests of PR 2/3.

## Risk

Low, PR is easy to revert.

## Deploy Plan

Deploy front.

### DIFF
--- a/front/components/pages/builder/skills/ManageSkillsPage.tsx
+++ b/front/components/pages/builder/skills/ManageSkillsPage.tsx
@@ -11,7 +11,7 @@ import {
 } from "@app/components/sparkle/AppLayoutContext";
 import { useHashParam } from "@app/hooks/useHashParams";
 import { useAuth, useWorkspace } from "@app/lib/auth/AuthContext";
-import { SKILL_ICON } from "@app/lib/skill";
+import { isDustProvidedSkill, SKILL_ICON } from "@app/lib/skill";
 import { useSkillsWithRelations } from "@app/lib/swr/skill_configurations";
 import { compareForFuzzySort, subFilter } from "@app/lib/utils";
 import { getSkillBuilderRoute } from "@app/lib/utils/router";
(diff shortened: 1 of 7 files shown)


front/components/pages/builder/skills/ManageSkillsPage.tsx
```diff
@@ -154,7 +154,11 @@ export function ManageSkillsPage() {
       ),
       default: filteredList(
         sortedActiveSkills
-          .filter((s) => s.isDefault || s.relations.editors === null)
+          .filter(
+            (s) =>
+              s.availability === "users_and_agents" ||
+              s.relations.editors === null
```
ykmsd: I know we do it right below already but isn't it kind of weird to check if it's dust managed or not by editors no? 🤨 It's not a big deal I guess but maybe someday we can have editorless custom skills (as a bug or feature??)? Can't we check via sId or something??

ykmsd (approved):

front/components/pages/builder/skills/ManageSkillsPage.tsx
fabiencelier: I can use the `isDustProvidedSkill` method.
It's not very different as we are checking the skill.editedBy but at least it's a centralized place to update if we want to change the behaviour later

PR #28968, opened 2026-07-15
Return sandbox function outputs directly to Frames

fontanierh:
## Description

This is the final PR for the [Frames sandbox functions discussion](https://dust4ai.slack.com/archives/C0AGY9MTNJV/p1784105075929679). The runner and server rollout PRs (#28962 and #28963) are already merged into `main`; this PR is standalone on top of that state.

Frames now receive the parsed `schema.output` value directly from `callFunction`. Failures reject with `SandboxFunctionCallError`, which preserves a stable error code, message, and optional HTTP status across the parent-to-iframe boundary. Public Frames report `not_supported`.

Existing Frame compatibility is intentionally not preserved. This PR pins `dsbx` v0.1.36 and bumps the sandbox base image from v0.8.56 to v0.8.57 so the new runner is deployed with the Frame API. The image build downloads and verifies the `dsbx-v0.1.36` release artifacts.

## Tests

- Front typecheck with `npm run tsgo`
- 5 focused Viz tests covering direct results, structured cross-document errors, public Frame rejection, and error-code parity
- 16 sandbox image registry tests covering the image and CLI pins
- Viz production build
- Biome checks on changed files

## Risk

The Frame `callFunction` contract is breaking by design. The new sandbox image must not be built until the `dsbx-v0.1.36` release artifacts and checksum exist. The server compatibility layer from #28963 is already on `main` and accepts both the prior and new runner callback envelopes during the rolling image deployment.

## Deploy Plan

- [x] Merge #28962 and #28963. (Already complete.)
- [ ] Merge this PR.
- [ ] On the resulting `main`, run the **Release dsbx CLI** workflow; it reads the Cargo package version and should create `dsbx-v0.1.36`.
- [ ] Confirm `dsbx-linux-x86_64` and `checksums-sha256.txt` exist for v0.1.36.
- [ ] Build and publish `dust-base:0.8.57` after the dsbx release artifacts exist.
- [ ] Deploy the app containing the new registry pin only after the image is available.
- [ ] Open `/poke/kill` → **Kill Switches** and request **Kill older versions** for `dust-base:0.8.57`, so existing sandboxes are reaped and recreated from the new image.

Rollback by restoring the `dust-base:0.8.56` / `dsbx` v0.1.35 pins, deploying that configuration, and requesting **Kill older versions** for the restored image.

PR #29384, opened 2026-07-23
Pass sandbox function error codes through instead of re-deriving them

fontanierh:
## Description

Follow-up on the review feedback on #28968.

The Frame parent was re-deriving the error code from the invocation response: `sandbox_function_not_found` became `function_not_found`, and everything else (403, 400, 500) collapsed into `invocation_failed`. That drops the only classification the endpoint produced, and leaves you walking six files to find out which layer relabelled what. It now forwards the API error type verbatim, and viz drops its copy of the code list so a code it has not heard of yet passes through instead of becoming `transport_error`. That also removes the test that regex-scraped two source files from other workspaces to keep the duplicated lists in sync, and the third copy of the list in the private swagger enum.

Two related gaps that made a failed call hard to trace at all:

- `fail()` persisted only `error.message`, so `inspect_invocations` could never show the code or status. It now persists the whole call error.
- Nothing logged a callFunction failure on the way out to the iframe. Added one log at the choke point.

Storing a different shape meant versioning the blob properly. The `version` field already existed but was never bumped, so it was about to start lying. It is now 2, reads discriminate on it across every shape we have written, and v1 migrates on read.

Last one: since #28968 the Frame gets errors by rejection, and React error boundaries do not catch async rejections. A Frame that awaits `callFunction` without a `catch` renders nothing, reports nothing, and no longer even hits the `console.error` that PR removed. Unhandled rejections now route into the existing error path (throw to ErrorBoundary, report to parent, Datadog, retry UI).

## Tests

Added cases for the parts CI would not otherwise cover: a forwarded code viz does not know about survives normalization, a classified failure keeps its code and status through the GCS round trip, a v1 blob migrates (including the bare `{version: 1}` shape the GCS path backfill wrote) and gets rewritten as v2, and an unreadable blob degrades instead of failing the read.

Not covered by a test: the `unhandledrejection` listener, which would need the whole Runner pipeline mounted.

## Risk

**The blob version bump is the part to look at.** Reads on this branch accept v1 and v2 and degrade on anything else, but that tolerant reader ships in the same deploy as the v2 writer, so it only protects the *next* schema change, not this one. Pods running today's `main` throw on a blob they cannot parse, and `baseFetch` loads every invocation's blob, so during the rolling window (and after a rollback) an old pod hitting any v2 blob fails the whole `inspect_invocations` listing for that function, not just that entry. Forward-fixable by rolling forward, and bounded by the `sandbox_functions` flag. If that window is not acceptable, say so and I will split the tolerant reader into its own PR to deploy first.

The skill doc changes what `code` values a Frame can expect, so Frames branching on `function_not_found` need `sandbox_function_not_found` instead. Every existing Frame is already breaking from #28968, so this rides along with that.

The `unhandledrejection` listener is the other behavior change worth watching. It is global to the iframe and fires after the Frame has rendered too, so a Frame that was partly working and has an uncaught background rejection now gets replaced by the parent's error card. That is deliberate: gating it on "nothing has rendered yet" would miss the main case (a Frame that mounts, fires `callFunction` in an effect, never catches, and spins forever), and the error card is what feeds the failure back to the model on retry. Frames that were quietly failing will start showing an error.

## Deploy Plan

Standard deploy, with the rolling-window caveat above. Rebased over #29392 (`context`/timezone), which touched the same schema.

### DIFF
--- a/front/lib/resources/sandbox_function_invocation_resource.ts
+++ b/front/lib/resources/sandbox_function_invocation_resource.ts
@@ -38,10 +38,12 @@ import type {
 import { isDevelopment } from "@app/types/shared/env";
 import type { ModelId } from "@app/types/shared/model_id";
 import { Err, Ok, type Result } from "@app/types/shared/result";
+import { assertNever } from "@app/types/shared/utils/assert_never";
 import { normalizeError } from "@app/types/shared/utils/error_utils";
 import { truncate } from "@app/types/shared/utils/string_utils";
 import type { Attributes, Transaction } from "sequelize";
 import { z } from "zod";
+import { fromError } from "zod-validation-error";
 
 const SANDBOX_FUNCTION_WORKING_DIRECTORY = "/home/agent";
 const SANDBOX_FUNCTION_EXEC_TIMEOUT_MS = 2 * 60 * 1000;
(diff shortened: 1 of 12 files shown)


front/lib/resources/skill/code_defined/global/pod_functions.ts
```diff
@@ -185,12 +185,16 @@ const output = await callFunction("<podId>/<slug>", input);
 
 \`input\` must match the function's declared input schema (see \`get\`). The promise resolves to
 the parsed and validated \`schema.output\`. It rejects with a \`SandboxFunctionCallError\` (also
-exported by \`@dust/react-hooks\`) whose \`code\` distinguishes missing functions, invalid
-input/output, handler failures, HTTP errors, and transport failures; render loading and error
-states like for any other network call. Pod functions are only reachable from authenticated
-Pod Frames, not from public or shared Frames.`,
+exported by \`@dust/react-hooks\`) carrying a \`message\`, an optional HTTP \`status\`, and a
+\`code\`; render loading and error states like for any other network call. Treat \`code\` as an
```
flvndvd, line 189: I'm actually reworking this piece a bit. fine to merge as-is but will change it again.

front/lib/resources/sandbox_function_invocation_resource.ts
```diff
@@ -74,15 +98,22 @@ export interface SandboxFunctionInvocationForLLM {
 }
 
 function parseSandboxFunctionInvocationData(
+  gcsPath: string,
```
flvndvd: Since this guy is a class's method, any reason to pass a class member, versus keeping previous implementation that access it directly?

front/lib/resources/sandbox_function_invocation_resource.ts
```diff
@@ -52,11 +52,35 @@ const SANDBOX_FUNCTION_ERROR_LOG_MAX_CHARS = 16_384;
...
+  code: z.string(),
+  message: z.string(),
+  status: z.number().optional(),
+});
+
+export type PersistedSandboxFunctionCallError = z.infer<
+  typeof PersistedErrorSchema
+>;
+
+const PersistedSandboxFunctionCallErrorSchema = z.union([
```
flvndvd: I wish we had a schema version in the blob, cause I feel like this is too brittle as we iterate.

flvndvd (approved):

PR #29386, opened 2026-07-23
feat(front): add admin setting to allow published agents on restricted model tiers

Nils-Fedrigo:
## Description

With `models_picker`, a published agent can be configured (by an editor who has access) with a model tier that some members are not granted. Their runs are currently blocked at step 0 of the agent loop with a `model_tier_not_enabled` error.

This adds a workspace admin setting to allow those runs anyway:

- New workspace metadata boolean `allowRestrictedModelsForPublishedAgents` (default off), updated via `POST /api/w/[wId]` (admin-only), following the `allowEmailAgents` pattern, with a new `workspace.published_agents_restricted_models_updated` audit event.
- `getModelTierAccessErrorForAgentConfiguration` takes an optional `agentScope`; the tier check is skipped when the agent is published (`scope === "visible"`) and the override is enabled. Only the run path passes the scope — creating/upgrading an agent with a restricted model stays blocked, so members cannot self-publish a premium-model agent to bypass tier controls. The model picker selectability logic is also unchanged.
- New "Published agents" toggle in the Models tier card (Admin → Usage), already admin- and `models_picker`-gated.

## Tests

- New functional tests for the endpoint branch (`front-api/routes/w/[wId]/index.test.ts`): sets/unsets the metadata key, preserves other metadata, 403 for non-admins.
- New `front/lib/model_tiers/access.test.ts`: override off/on × published/hidden/no-scope (creation path), plus the `models_picker`-disabled case.

## Risk

Low. The override is opt-in per workspace and defaults to off, preserving current behavior. Rollback-safe: reverting simply re-blocks the runs.

## Deploy Plan

Before merge: register the new audit schema and commit the regenerated version map in this PR (`npx tsx front/admin/register_audit_log_schemas.ts --execute --changed`, requires `WORKOS_API_KEY`). Regular deploy otherwise.

### DIFF
--- a/front/lib/model_tiers/access.ts
+++ b/front/lib/model_tiers/access.ts
@@ -42,9 +42,6 @@ export async function getModelTierAccessErrorForAgentConfiguration(
     model: ModelConfigurationType;
     reasoningEffort?: ReasoningEffort;
     featureFlags: WhitelistableFeature[];
-    // Only passed on the run path. The create/upgrade path deliberately omits
-    // it: the published-agents override must not let members configure agents
-    // with models above their own tier.
     agentScope?: AgentConfigurationScope;
   }
 ): Promise<GenericErrorContent | null> {
--- a/front/lib/model_tiers/access.ts
+++ b/front/lib/model_tiers/access.ts
@@ -1,12 +1,16 @@
 import type { Authenticator } from "@app/lib/auth";
 import { ModelsTierResource } from "@app/lib/resources/models_tier_resource";
-import type { GenericErrorContent } from "@app/types/assistant/agent";
+import type {
+  AgentConfigurationScope,
+  GenericErrorContent,
+} from "@app/types/assistant/agent";
 import type {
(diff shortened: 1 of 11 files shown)


Nils-Fedrigo:

- #29386 ◀

front/lib/model_tiers/access.ts
```diff
@@ -32,17 +36,31 @@ export async function getModelTierAccessErrorForAgentConfiguration(
...
     featureFlags,
+    agentScope,
   }: {
     agentName: string;
     model: ModelConfigurationType;
     reasoningEffort?: ReasoningEffort;
     featureFlags: WhitelistableFeature[];
+    // Only passed on the run path. The create/upgrade path deliberately omits
+    // it: the published-agents override must not let members configure agents
+    // with models above their own tier.
```
Nils-Fedrigo:
```suggestion
```

Fraggle (approved):
LGTM

PR #29387, opened 2026-07-23
[front] new frame permission check

ykmsd:
## Description
This PR is to read permission check from governance table. We already backfilled the table with scope "everyone". It's safe because `workspaceSharingPolicy` always wins, so we only check the scope when a feature is enabled. 

If workspace sharing policy is "no restriction":
- External email sharing: we check if you have `invite:frame`
- Public sharing: we check if you have `publish:frame`

if you have permissions:
CleanShot 2026-07-23 at 13 09 37@2x

if you don't have both permissions:
CleanShot 2026-07-23 at 13 10 11@2x

if you don't have permission to share public but can invite external people:
CleanShot 2026-07-23 at 15 08 28@2x

if you don't have permission to invite external people but can share public (weird case and not sure what else to show):
CleanShot 2026-07-23 at 15 09 42@2x

if you lost permission to share frames publicly and open the past frame you shared publicly:
CleanShot 2026-07-23 at 16 00 19@2x

### DIFF
--- a/front-api/routes/w/[wId]/files/[fileId]/share/index.test.ts
+++ b/front-api/routes/w/[wId]/files/[fileId]/share/index.test.ts
@@ -1,7 +1,7 @@
 import { Authenticator } from "@app/lib/auth";
 import type { FileResource } from "@app/lib/resources/file_resource";
 import { GroupPermissionResource } from "@app/lib/resources/group_permission_resource";
-import { WorkspaceModel } from "@app/lib/resources/storage/models/workspace";
+import { WorkspaceResource } from "@app/lib/resources/workspace_resource";
 import { FileFactory } from "@app/tests/utils/FileFactory";
 import { createPrivateApiMockRequest } from "@app/tests/utils/generic_private_api_tests";
 import { frameContentType } from "@app/types/files";
@@ -43,10 +43,11 @@ async function setSharingPolicy(
   workspace: { sId: string },
   sharingPolicy: WorkspaceSharingPolicy
 ) {
-  await WorkspaceModel.update(
-    { sharingPolicy },
-    { where: { sId: workspace.sId } }
-  );
+  const workspaceResource = await WorkspaceResource.fetchById(workspace.sId);
+  if (!workspaceResource) {
+    throw new Error(`Workspace ${workspace.sId} not found.`);
+  }
+  await workspaceResource.updateWorkspaceSettings({ sharingPolicy });
 }
 
 function postShare(workspace: { sId: string }, fileId: string, body: unknown) {
--- a/front-api/routes/w/[wId]/files/[fileId]/share/index.test.ts
+++ b/front-api/routes/w/[wId]/files/[fileId]/share/index.test.ts
@@ -1,0 +1,205 @@
+import { Authenticator } from "@app/lib/auth";
+import type { FileResource } from "@app/lib/resources/file_resource";
+import { GroupPermissionResource } from "@app/lib/resources/group_permission_resource";
+import { WorkspaceModel } from "@app/lib/resources/storage/models/workspace";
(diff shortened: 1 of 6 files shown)


front/components/assistant/conversation/interactive_content/frame/shareFrameScopeOptions.ts
```diff
@@ -0,0 +1,99 @@
+import type { FileShareScope } from "@app/types/files";
```
ykmsd, line 1: move it out from the component since it's getting big

front/components/assistant/conversation/interactive_content/frame/shareFrameScopeOptions.ts
```diff
@@ -0,0 +1,99 @@
...
+const ALLOWED_SCOPES_BY_POLICY: Record<
+  WorkspaceSharingPolicy,
+  FileShareScope[]
+> = {
+  workspace_only: ["emails_only", "workspace_and_emails"],
+  workspace_and_emails: ["emails_only", "workspace_and_emails"],
+  all_scopes: ["emails_only", "workspace_and_emails", "public"],
+};
+
+function getScopeOptions(canInviteExternal: boolean): ScopeOption[] {
```
ykmsd, line 27: updated wording a bit, when you cannot invite external people: 

before:

Image


after:

Image

front/components/assistant/conversation/interactive_content/frame/ShareFrameSheet.tsx
```diff
@@ -269,14 +261,25 @@ export function ShareFrameSheet({
...
+                  </ContentMessage>
+                )}
+                {lostPublishPermission && (
+                  <ContentMessage
+                    icon={InfoCircle}
+                    variant="info"
+                    title="You no longer have permission to share frames publicly"
+                  >
+                    This frame is currently shared publicly. You can restrict
+                    access, but you won’t be able to make it public again.
```
ykmsd, line 282: I'm not sure what to say here

matteotrab (approved):

front-api/routes/w/[wId]/files/[fileId]/share/index.test.ts
```diff
@@ -0,0 +1,205 @@
...
+    grantType: "publish",
+    resourceType: "frame",
+  });
+}
+
+async function setSharingPolicy(
+  workspace: { sId: string },
+  sharingPolicy: WorkspaceSharingPolicy
+) {
+  await WorkspaceModel.update(
```
Comment (flvndvd) [inline front-api/routes/w/[wId]/files/[fileId]/share/index.test.ts]: Let's avoid using raw models in test files, if we can use the resource 🙏

front/components/assistant/conversation/interactive_content/frame/ShareFrameSheet.tsx
flvndvd: IIRC we were downgrading all frames to workspace when the admin is disabling public sharing, no?

flvndvd (approved):

front/components/assistant/conversation/interactive_content/frame/ShareFrameSheet.tsx
ykmsd: Yes but this will be only shown when the workspace policy stays the same but you lost the permission to share frames publicly (in which case we don't downgrade your past frames)

front-api/routes/w/[wId]/files/[fileId]/share/index.test.ts
Comment (ykmsd) [reply front-api/routes/w/[wId]/files/[fileId]/share/index.test.ts]: 🙈

Comment (ykmsd) [reply front-api/routes/w/[wId]/files/[fileId]/share/index.test.ts]: I think we do it in other places too 🥶

PR #29390, opened 2026-07-23
[front] enh: expose tables query outputs to pod functions

aubin-tchoi:
## Description

Better reviewed commit by commit (at least the first 2 separately)

Tables query and data warehouses currently require an agent-loop context because their CSV output is attached to a conversation.

This PR supports using Tables query and data warehouse in sandbox function invocations, in which case they write a CSV file in `.tool_outputs/{function-slug}` (wired in through `DustFileSystem`). The agent-loop path remains strictly the same.

The two paths (agent loop context or sandbox) look very different: for the agent loop one we can't rely on the `DustFileSystem` as it needs to have the csv file uploaded to the conversation data source for it to be tables-queryable in turn. Will be able to cleanup once we switch tables-query usage to fully leveraging the Computer (i.e. do pandas and stuff on the computer instead of the sqlite path).

## Tests

- Tested locally.

## Risk

- Low.

## Deploy Plan

- Deploy front.


spolu (approved):
💯

PR #29392, opened 2026-07-23
[front] fix: propagate user timezone to sandbox function tools

aubin-tchoi:
## Description

Sandbox functions currently lose the originating user timezone when they call tools. Google Calendar cannot format event data in that timezone, and `pod_manager` `create_conversation` falls back to the server timezone.

Especially useful for a use case where you would build a "Book a meeting with me!"

This PR exposes the timezone through sandbox function invocations.

## Tests

- Added tests.

## Risk

- Low.

## Deploy Plan

- Deploy front.


spolu (approved):
LGTM \o/

PR #29393, opened 2026-07-23
Skills: fix condition to display editors panel

fabiencelier:
## Description
Fix https://github.com/dust-tt/dust/pull/29378

The condition to be a global skill is not `isDefault` 🤦 
`isDustProvidedSkill` is more relevant because it means exactly that: when provided by dust there is no editor to show
 
## Deploy Plan

deploy front


aubin-tchoi (approved):
LGTM

PR #11442, opened 2025-03-18
Ignore softDelete data source when deleting a workspace

flvndvd:
## Description

This PR fixes [this](https://dust4ai.slack.com/archives/C05F84CFP0E/p1742299035342369) failing workspace deletion.

We leverage the existing deletion workspace code to purge data for relocated workspaces, this code assert that data sources must be soft deleted before, which isn't true when deleting a workspace. This PR updates the logic to first soft delete data source views and data sources.

PR #29394, opened 2026-07-23
[Workspace Deletion] Allow deleting workspaces with data sources

philipperolet:
## Description

The Poke workspace deletion action rejects workspaces that still have data sources, forcing operators to delete them one by one.

The guard was introduced in https://github.com/dust-tt/dust/pull/11406. The deletion workflow was then updated in https://github.com/dust-tt/dust/pull/11442 to soft-delete and scrub data sources while deleting spaces.

This PR adds a "Delete all data sources" checkbox to the Poke action. The checkbox defaults off and preserves the existing guard; when explicitly checked, the action lets the workspace deletion workflow clean up its data sources.

It also fixes the workflow ordering for data sources shared across spaces: all data source views are now soft-deleted workspace-wide before any data source or space is scrubbed. This prevents a later live view from blocking its data source deletion after the workflow has already removed other workspace data.

Membership deletion now runs after space deletion so GitHub data-source cleanup can still notify workspace admins.

The free-plan and Metronome-contract safeguards remain unchanged.

## Risks

Blast radius: Poke workspace deletion and workspace deletion workflow ordering.
Risk: standard

The workflow order change is intentionally unpatched. Deploy only when no workspace deletion workflows are running across the deployment.

## Deploy Plan

- Confirm no workspace deletion workflows are running
- Deploy front

## Tests

- [x] Poke workspace deletion blocks data sources by default
- [x] Poke workspace deletion accepts explicit data-source deletion
- [x] Multi-space data source deletion activity test

### DIFF
--- a/front/lib/api/poke/plugins/workspaces/delete_workspace.test.ts
+++ b/front/lib/api/poke/plugins/workspaces/delete_workspace.test.ts
@@ -40,6 +40,7 @@ describe("deleteWorkspacePlugin.execute", () => {
     const result = await deleteWorkspacePlugin.execute(auth, workspace, {
       confirmation: "DELETE",
       workspaceHasBeenRelocated: false,
+      deleteDataSources: false,
     });
 
     expect(result.isErr()).toBe(true);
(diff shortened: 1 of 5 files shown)


pmilliotte (commented):
LGTM

PR #29396, opened 2026-07-23
Refactoring to avoid loading conversation content all the time.

Fraggle:
## Description

Route handlers for edit, retry, post-message, soft-delete, and compaction operations previously called `getConversation()`, which loads the entire conversation (all messages, all branches) even when only conversation metadata or a single message was needed. This replaces `getConversation` with `ConversationResource.fetchById` (lightweight) across ~10 route handlers in both the public v1 API and the internal workspace API.

For message-level operations, only the target message is now rendered via `batchRenderMessages([messageModel], "full")` instead of scanning `conversation.content.flat()`. The signatures of `editUserMessage`, `retryAgentMessage`, `postUserMessage`, and `softDeleteUserMessageAndReplies` now accept `conversationResource` + `branchId` instead of a full `ConversationType`. A new `fetchPrecedingContentFragments` method on `ConversationResource` replaces inline content-fragment filtering in the POST message handler. Test helpers `makeExtra`/`setupProjectConversation` are centralized in `conversation_test_factories.ts`.

## Tests

Updated existing tests to pass `ConversationResource` directly where `ConversationType` was previously expected. Extracted shared test factories to `conversation_test_factories.ts`. Tested edit, retry, post-message, and delete endpoints manually.

## Risk

Medium — changes signatures of core conversation mutation functions (`editUserMessage`, `retryAgentMessage`, `postUserMessage`, `softDeleteUserMessageAndReplies`). Behavior is equivalent; no auth logic changed, only data loading. Safe to rollback.

## Deploy Plan

Deploy front and front-api. No SQL migrations.

### DIFF
--- a/front-api/routes/v1/w/[wId]/assistant/conversations/[cId]/index.ts
+++ b/front-api/routes/v1/w/[wId]/assistant/conversations/[cId]/index.ts
@@ -9,7 +9,7 @@ import {
 } from "@dust-tt/client";
 import { apiErrorForConversation } from "@front-api/lib/api/assistant/conversation/helper";
 import { publicApiApp } from "@front-api/middlewares/ctx";
-import type { HandlerResult } from "@front-api/middlewares/utils";
+import { apiError, type HandlerResult } from "@front-api/middlewares/utils";
 import { validate } from "@front-api/middlewares/validator";
 import { z } from "zod";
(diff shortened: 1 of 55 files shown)


Fraggle:

- #29396 ◀

PR #29397, opened 2026-07-23
[front] enh: expose image generation outputs to pod functions

aubin-tchoi:
## Description

`generate_image` currently asserts an agent-loop context when saving its output, so it cannot return generated files from a pod function invocation.

This PR writes pod-function outputs to `.tool_outputs/{function-slug}` through `DustFileSystem` and returns `FILE_PATH` resources. The agent-loop path remains unchanged and continues using conversation `FileResource`s so generated images go through the existing resize and rendering pipeline.

This follows the same output split as [#29390](https://github.com/dust-tt/dust/pull/29390) for tables query and data warehouse.

## Tests

- Added a focused test for pod-function image output.

## Risk

- Low.

## Deploy Plan

- Deploy front.

### DIFF
--- a/front/lib/api/actions/servers/image_generation/helpers.ts
+++ b/front/lib/api/actions/servers/image_generation/helpers.ts
@@ -30,7 +30,6 @@ import type { Result } from "@app/types/shared/result";
 import { Err, Ok } from "@app/types/shared/result";
 import type { WorkspaceType } from "@app/types/user";
 import { INTERNAL_MIME_TYPES } from "@dust-tt/client";
-import assert from "assert";
 
 export type ImageGenerationErrorCode =
   | "api_error"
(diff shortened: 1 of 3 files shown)


id13 (approved):
LGTM

front/lib/api/actions/servers/image_generation/helpers.ts
```diff
@@ -284,6 +276,116 @@ export async function uploadAndFormatImageResponse(
...
+
+    resources.push({
+      type: "resource",
+      resource: {
+        mimeType: INTERNAL_MIME_TYPES.TOOL_OUTPUT.FILE_PATH,
+        uri: scopedPath,
+        path: scopedPath,
+        title: outputFileName,
+        contentType: mimeType,
+        text: `Generated image: ${outputFileName}`,
```
spolu, line 338: Should we remove the prefix here. This will be used programatically always. The model will have to adapt to it and slice the thing which is a bit weird.

No need to address here but to keep in mind.

spolu (commented):
LGTM \o/

PR #29400, opened 2026-07-23
[front] Raise the bar for entering plan mode

fontanierh:
## Description

The Plan Mode skill instructions pushed toward creating a plan on almost every turn:

- "**Default behavior: call `create_plan` at the start of any non-trivial turn.**"
- "Non-trivial" was defined to include *"research that will span multiple tool calls"* and *"anything the user might plausibly want to follow along with"*, which is essentially everything.
- "**When in doubt, err on the side of creating a plan: the cost is one tool call, the upside is transparency.**"
- The skip list was narrow: only single-shot questions, quick lookups, one-tool-call answers and clarifications.

The result is plans on tasks that finish in two or three tool calls, where the plan card is overhead rather than transparency: the user gets something to read and track for work that was already done by the time they read it.

This PR raises the bar for entering plan mode, in `PLAN_MODE_INSTRUCTIONS`:

- **Explicit request wins, stated first.** If the user asks for plan mode by name, always create one, even for a small task. That rule now explicitly overrides everything below it, so it cannot collide with the skip list.
- **Creation is gated on work being genuinely multi-step**: a change spanning several files or systems, research across several sources, or a task whose steps need sequencing and revisiting rather than running straight through. The threshold is stated once, with a number: under roughly 5 tool calls, there is nothing worth planning.
- **Flipped the tie-break** to "when in doubt, skip it", with the reason stated.
- **Widened the skip list**: questions and lookups, single edits or fixes, clarification exchanges, and follow-ups that just tweak previous output. Also called out that answer length is not the signal, since a long written answer produced in one shot needs no plan.
- **Added a mid-turn escalation path** so work that turns out larger than expected can still get a plan. The bar is the size of the work, not when you noticed it. This mirrors how `go_deep` handles the symmetric case.

Everything downstream of creation (editing, approval flow, closing) is unchanged.

Also aligned the three descriptions that carried the old "non-trivial" wording (the skill's `agentFacingDescription`, the MCP server description, and the `plan_mode` feature flag description), and bumped the skill `version` to 2 per the convention for instruction changes.

## Tests

- `npx tsgo --noEmit` clean.
- `lib/api/assistant/plan_mode.test.ts`, `components/assistant/conversation/plan_mode/`, and `lib/actions/mcp_internal_actions/mcp_servers_metadata.test.ts`: 22 tests pass.
- No snapshot update needed: the MCP metadata snapshot covers tool descriptions, stakes and billing, not `serverInfo.description`.
- No existing test asserts on the instruction text, so this change is not covered by automated tests. The real validation is behavioral and post-deploy (see Risk).

## Risk

Low and fully reversible.

- Prompt-only change. No schema, no API contract, no data migration, no client change.
- Gated behind the `plan_mode` feature flag (`dust_only` stage), so the blast radius is limited to workspaces that have it enabled. Note that `plan_mode` is in the Dust global agent's skill list, so on flagged workspaces this text sits in the Dust agent's system prompt.
- The real risk is over-correction: the tie-break now favors skipping, so agents may under-plan on tasks that genuinely warrant a plan. Mitigated by the explicit-request rule (asking for plan mode always works) and by the mid-turn escalation path (work that grows past the bar can still get a plan). Worth watching plan-creation rates after deploy: the interesting number is the floor, not just the drop.
- Rollback is a plain revert of this PR. Nothing persists, since code-defined skills are resolved from the in-memory registry per request.

## Deploy Plan

Standard `front` deploy. No migration, no backfill, no ordering constraint.

Code-defined skills are read from the in-memory `SYSTEM_SKILLS_ARRAY` on each request rather than persisted, so the new instructions take effect as soon as `front` is deployed. The `version` bump is metadata for the instruction-change convention and does not gate anything at runtime.

### DIFF
--- a/front/lib/api/actions/servers/plan_mode/metadata.ts
+++ b/front/lib/api/actions/servers/plan_mode/metadata.ts
@@ -107,9 +107,9 @@ export const PLAN_MODE_SERVER = {
     name: PLAN_MODE_SERVER_NAME,
     version: "1.0.0",
     description:
-      `Create and maintain a living \`${PLAN_FILE_NAME}\` that gives the user visibility on non-trivial ` +
-      "work. When you need explicit sign-off before proceeding, ask the user with the standard " +
-      "question flow.",
+      `Create and maintain a living \`${PLAN_FILE_NAME}\` that gives the user visibility on genuinely ` +
+      "multi-step work. When you need explicit sign-off before proceeding, ask the user with the " +
+      "standard question flow.",
     icon: "ActionDocumentTextIcon" as const,
     authorization: null,
     documentationUrl: null,
--- a/front/lib/resources/skill/code_defined/system/plan_mode.ts
+++ b/front/lib/resources/skill/code_defined/system/plan_mode.ts
@@ -11,11 +11,15 @@ import type { SystemSkillDefinition } from "@app/lib/resources/skill/code_define
 const ASK_USER_QUESTION_TOOL_NAME = "ask_user_question";
 
 const PLAN_MODE_INSTRUCTIONS = `
-Plan Mode lets you maintain a live \`plan.md\` the user can follow as you work. Think of it as a shared progress view, not just an approval gate. Using it is delightful UX: the user sees what you're doing without having to ask.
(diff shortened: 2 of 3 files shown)


PopDaph (approved):
LGTM thank you!

PR #29401, opened 2026-07-23
[front] remove builder role from member management UI (gated behind admin_governance)

tdraier:
## Description

First step of removing the deprecated `builder` role (part of dust-tt/tasks#9746). Scoped to the user-facing member management UI. **All changes are gated behind the `admin_governance` feature flag** — when the flag is off, behavior is unchanged; the new behavior only applies once governance is enabled. The `builder` value stays in the `ROLES`/`ACTIVE_ROLES` type arrays for now since existing builder memberships still exist and the backend cleanup is a later step.

### Changes

**1. Role dropdown** (`RolesDropDown`)
- `builder` is excluded from the assignable roles only when `admin_governance` is enabled — so it's no longer offered when inviting users, editing invitations, or changing a member's role.
- A builder's current role renders as a regular member (chip) when the flag is on.

**2. `buildersOnly` search filter**
- The `searchMembers` `buildersOnly` filter and its full call chain (members search route, `useSearchMembers` hook, `MemberSelectionTable`, `ManageUsersPanel`) are kept intact.
- `SkillEditorsSheet` now requests `buildersOnly={!hasFeature("admin_governance")}`, so skill editors stay restricted to builders while the flag is off and become any-member once it's on.

**3. Builder-as-member display**
- New pure helper `normalizeDisplayRole(role, isAdminGovernanceEnabled)` in `front/components/members/Roles.ts` maps `builder` → `user` for display only when the flag is on.
- Applied to the member list role chip (`MembersList`, via a small `RoleCell` component) and the role description in `ChangeMemberModal`.

### Testing

- `tsgo --noEmit` passes for both `front` and `front-api`.
- biome clean on all changed files.
- Restored `buildersOnly` route tests remain (DB-backed; run in CI).

Closes dust-tt/tasks#9786

### DIFF
--- a/extension/ui/components/auth/AuthProvider.tsx
+++ b/extension/ui/components/auth/AuthProvider.tsx
@@ -1,5 +1,5 @@
 import { AuthContext, type AuthContextValue } from "@app/lib/auth/AuthContext";
-import { EMPTY_WORKSPACE_PERMISSIONS } from "@app/types/group_permissions";
+import { emptyWorkspacePermissions } from "@app/types/group_permissions";
 import type { SubscriptionType } from "@app/types/plan";
 import type { UserTypeWithWorkspaces, WorkspaceType } from "@app/types/user";
 import { isAdmin, isBuilder, isManager } from "@app/types/user";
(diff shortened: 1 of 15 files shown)


tdraier:

- #29401 ◀

matteotrab (approved):

PR #29403, opened 2026-07-23
[front] fix: converge Stripe-backed subs to ended on contract swap

tdraier:
## Problem

Legacy Pro → Business migrated workspaces (via `migrate_legacy_pro_monthly_to_business.ts` → `switchContract` → pending row → `contract.start`) were left stranded in **`ended_backend_only`** instead of `ended`, even though Stripe had correctly ended the subscription — permanently, not transiently.

## Root cause — a mutual-deferral race

At the cutover instant two webhooks fire for the same workspace:

- **Metronome `contract.start`** → `activatePending()` marks the legacy Stripe-backed sub `ended_backend_only`, deferring the final `ended` flip to Stripe (`swapEndedStatus`).
- **Stripe `customer.subscription.deleted`** (the scheduled `cancel_at` reaching the cutover) — the event that is *supposed* to converge `ended_backend_only` → `ended`.

Convergence only works if `deleted` is processed **after** `activatePending`. When the Stripe event wins the race:

1. The deleted handler hits `case "active"` + the pending-sub guard → logs *"raced ahead… contract.start will finalize the swap"* and **skips without converging**.
2. Then `activatePending` sets the sub to `ended_backend_only`, expecting Stripe's `deleted` — which already fired and was consumed.
3. No webhook is left to finalize it → stranded forever.

Shadow-billed legacy subs (both `stripeSubscriptionId` + `metronomeContractId`) can't be rescued by Metronome `contract.end` either — it breaks early on `isMetronomeShadowBilled`. So Stripe's `deleted` was the sole converger, and it lost the race.

## Fix

`activatePending()` now ends the old sub **directly as `ended`** rather than via `swapEndedStatus`. This is scrub-safe in every ordering:

- `deleted` first → `case "active"` + pending guard skips the scrub, then `activatePending` sets `ended`. ✅
- `deleted` later → finds `ended` → `case "ended"` no-op. ✅

`swapEndedStatus` is retained for the in-place `swapMetronomeContract` path (where `contract.start` reliably precedes the Stripe cancel).

## Backfill

Adds `backfill_stranded_ended_backend_only_subscriptions.ts` to reconcile already-stranded rows: for each `ended_backend_only` sub whose Stripe subscription is **confirmed `canceled`**, flip it to `ended`. Subs without a Stripe ID / not confirmed canceled are left untouched and logged for manual review. Dry-run by default.

```
npx tsx scripts/backfill_stranded_ended_backend_only_subscriptions.ts            # dry run
npx tsx scripts/backfill_stranded_ended_backend_only_subscriptions.ts --execute
```

## Test

Updated the `process_webhook` test that encoded the old (buggy) behavior — it now asserts a shadow-billed (Stripe-backed) sub converges to `ended` on activation.

## Notes / reviewer callout

- Scoped the code fix to `activatePending` (the path the migration uses). `swapMetronomeContract` converges in its normal ordering, so it's left on `swapEndedStatus`; happy to harden it the same way for consistency if preferred.
- CI will run the full test suite + eslint — those could not be run locally (this workspace's `node_modules` can't build the `@dust-tt/client` SDK and is missing `ajv` for eslint). `tsgo --noEmit` and `biome check` pass locally, and the pre-commit typecheck hook passed.

### DIFF
--- a/front/lib/api/metronome/process_webhook.test.ts
+++ b/front/lib/api/metronome/process_webhook.test.ts
@@ -695,10 +695,14 @@ describe("processMetronomeWebhook — swap webhook ordering", () => {
     expect(oldSub!.status).toBe("ended");
   });
 
-  it("keeps a shadow-billed (Stripe-backed) old sub as ended_backend_only so Stripe converges it", async () => {
-    // The fix is scoped to Metronome-only subs. A sub with a Stripe
-    // subscription must still wait for Stripe's customer.subscription.deleted
-    // webhook, so it ends as ended_backend_only.
+  it("converges a shadow-billed (Stripe-backed) old sub to ended on activation (does not strand in ended_backend_only)", async () => {
+    // A pending-row cutover races Stripe's customer.subscription.deleted (the
+    // scheduled cancel_at). When the Stripe event lands first it is consumed by
+    // the "active + pending → skip" branch, leaving contract.start's
+    // activatePending as the only handler that can finalize the old sub. It must
+    // therefore end directly as `ended` rather than waiting on a Stripe webhook
+    // that already fired, otherwise the old sub is stranded in
+    // `ended_backend_only`.
     const workspace = await setupMetronomeWorkspace(OLD_CONTRACT_ID, {
       stripeSubscriptionId: "sub_shadow_xxx",
     });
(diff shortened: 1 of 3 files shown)


PopDaph:
👍🏻

PR #29405, opened 2026-07-23
[Conversation Files] Fix audio transcript previews

philipperolet:
## Description

Fixes https://github.com/dust-tt/tasks/issues/9726

Audio previews use a canonical-path URL. The transcript request appended `version=processed` to that same URL, but the canonical-path endpoint ignores file versions and therefore returned the original binary audio bytes.

Load the transcript from the linked file resource’s processed-content endpoint while keeping audio playback on the canonical-path URL.

## Risks

Blast radius: transcript rendering for audio files in file previews.
Risk: low

## Deploy Plan

- deploy front


Fraggle (commented):
LGTM

PR #29406, opened 2026-07-23
[front] Retry Anthropic file download failures

philipperolet:
Fixes https://github.com/dust-tt/tasks/issues/9712

## Description

Anthropic can intermittently fail to download a signed image URL included in a long agent run and returns HTTP 400 with "Unable to download the file". We classified every such 400 as a permanent invalid request, so the agent loop stopped instead of retrying the failed model step.

Classify only this exact Anthropic diagnostic as a retryable server error in both the current model-constructor path and the legacy client path. Other HTTP 400 responses remain non-retryable.

This is a narrower alternative to https://github.com/dust-tt/dust/pull/26541, which base64-inlined every direct Anthropic image and was reverted by https://github.com/dust-tt/dust/pull/26553.

## Risks

Low. The behavior change is limited to Anthropic requests carrying the known file-download diagnostic. A persistent failure will still surface after the existing Temporal retry limit.

## Deploy Plan

Deploy front normally. No migration or configuration change is required.

## Tests

- Anthropic legacy error classification tests
- Anthropic model-constructor output conversion tests
- Front TypeScript type-check


ElPicador (commented):
👍

PR #29407, opened 2026-07-23
Activation nudge notif novu workflow

xebeohp:
## Description
Adds a general email notification capability specific to activation nudge conversations. It is a standalone Novu workflow (activation-new-conversation) for easier iteration and to prevent the normal conversation-unread workflow from running in parallel (creating duplicate email notifs).

The workflow's steps: resolve conversation details → read the user's email delay preference → step.delay for that interval → send one email, whose skip re-checks after the delay and only sends when there's an unread, succeeded agent reply. It respects existing user settings throughout and introduces no new preferences.

This adds all the plumbing to make the email notification. Email UI changes coming later.

## Tests
Tested on local development (dust, novu, email).

## Risk
Low risk, only used when there is an unread activation nudge convo.

## Deploy Plan
Deploy front.

### DIFF
--- a/front/lib/notifications/workflows/activation-new-conversation.ts
+++ b/front/lib/notifications/workflows/activation-new-conversation.ts
@@ -9,7 +9,7 @@ import {
   getUserNotificationDelay,
 } from "@app/lib/notifications";
 import { renderEmail } from "@app/lib/notifications/email-templates/default";
-import { getConversationDetails } from "@app/lib/notifications/workflows/conversation-unread";
+import { getConversationDetails } from "@app/lib/notifications/helpers";
 import type { UserResource } from "@app/lib/resources/user_resource";
 import { getConversationRoute } from "@app/lib/utils/router";
 import type { ConversationWithoutContentType } from "@app/types/assistant/conversation";
@@ -157,8 +157,7 @@ export const activationNewConversationWorkflow = workflow(
             id: payload.workspaceId,
             name: details.workspaceName,
           },
-          content: `Test test test! Hi you are recieving a secret message from Phoebe :)
-          There's something new waiting for you: "${details.subject}".`,
+          content: `There's something new waiting for you: "${details.subject}".`,
           action: {
             label: "Open conversation",
             url:
--- a/front/lib/notifications/workflows/activation-new-conversation.ts
+++ b/front/lib/notifications/workflows/activation-new-conversation.ts
@@ -157,8 +157,8 @@ export const activationNewConversationWorkflow = workflow(
             id: payload.workspaceId,
             name: details.workspaceName,
           },
-          content: `Test test test! Hi you are recieving a secret message from Phoebe :)" +
-          "There's something new waiting for you: "${details.subject}".`,
+          content: `Test test test! Hi you are recieving a secret message from Phoebe :)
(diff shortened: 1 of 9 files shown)


xebeohp:

- #29407 ◀
    - #29455

front/lib/notifications/workflows/conversation-unread.ts
```diff
@@ -157,7 +157,7 @@ const UserNotificationDelaySchema = z.object({
   delay: z.enum(NOTIFICATION_DELAY_OPTIONS),
 });
 
-const getConversationDetails = async ({
+export const getConversationDetails = async ({
```
ElPicador: I would move that in a dedicated file. It's not anymore specific to conversation-unread. 

Maybe a front/lib/notifications/helpers.ts or something

front/lib/notifications/workflows/activation-new-conversation.ts
```diff
@@ -0,0 +1,244 @@
...
+        return { delay: userNotificationDelay };
+      },
+      {
+        outputSchema: userNotificationDelaySchema,
+      }
+    );
+
+    await step.delay("apply-user-delay", async () => {
+      return { type: "regular", ...NOTIFICATION_PREFERENCES_DELAYS[delay] };
+    });
```
ElPicador, line 149: should we? 

the main goal here is to engage people, if they receive notifications every week it kinds of break the whole idea. On the other end we want to respect those preferences. 

Maybe there is a good middle ground to find?

front/lib/notifications/workflows/activation-new-conversation.ts
```diff
@@ -0,0 +1,244 @@
...
+    await step.email(
+      "send-email",
+      async () => {
+        const body = await renderEmail({
+          name: subscriber.firstName ?? "You",
+          workspace: {
+            id: payload.workspaceId,
+            name: details.workspaceName,
+          },
+          content: `Test test test! Hi you are recieving a secret message from Phoebe :)
```
ElPicador: That's nice, but we will get those emails in production the moment you merge this PR. 

I understand we want to improve those emails, nonetheless I would display something that is ok for production, not a test ;)

ElPicador (approved):
👍 I would really change the text even for something basic

front/lib/notifications/workflows/activation-new-conversation.ts
xebeohp: I agree with this concern but for now I believe respecting user preferences is better. This would agree with dust's privacy promise and not surprise them with an "unexpected" email, since this is not a super critical notif (compared to something like pricing).

If we receive feedback that an immediate notif would be better I'm open to changing it. I feel like as long as the user gets a notif (even a week later) it is ok, the main concern is the users who have notifs turned off.

front/lib/notifications/workflows/activation-new-conversation.ts
xebeohp: Fixed haha

front/lib/notifications/workflows/conversation-unread.ts
xebeohp: Fixed. Moved get conversation details to dust/front/lib/notifications/helpers.ts.

PR #29408, opened 2026-07-23
[front] Keep Frame side panel open when a message also writes regular files

frankaloia:
## Summary

When an activation-flow agent message writes an interactive content file (a Frame) **and** a regular file (e.g. `AGENTS.md`) in the same turn, the Frame side panel would flash open and then get replaced by the file explorer.

Root cause: two auto-open hooks fire on the same message. `useAutoOpenInteractiveContent` opens the Frame in the side panel, then `useAutoOpenFilesPanel` runs and clobbers it with the file explorer.

Fix: `useAutoOpenInteractiveContent` now exposes `willOpenInteractiveContent` — the single source of truth for whether a Frame is opening for this message (covering both the streaming-progress and completed-file paths). `useAutoOpenFilesPanel` defers to that flag instead of re-deriving its own guess from `generatedFiles`. This keeps the Frame panel in front and also handles the streaming-progress window (where the Frame is opening from progress notifications before the file lands in `generatedFiles`).

## Testing

- `npx tsgo --noEmit`: no new errors introduced (the 4 pre-existing errors are in `lib/api/llm/.../test/fixtures/` and unrelated to this change).
- `biome check --changed`: no fixes applied, no warnings.

## Risk

Low. The behavior only changes for messages that open a Frame: the files panel now defers to the Frame instead of taking over. Messages that produce only regular files still auto-open the files panel exactly as before. `useAutoOpenFilesPanel` is used across all conversations (not just activation Pods), so the defer applies anywhere a single message emits both a Frame and regular files — which is the intended behavior.

## Testing Plan

- In an activation Pod, send a message that produces both a Frame and a regular file (e.g. `AGENTS.md`); confirm the Frame stays open in the side panel and the file explorer does not take over.
- Send a message that produces only regular files; confirm the files panel still auto-opens.
- Confirm the Frame still auto-opens while streaming (progress) and on completion.

### DIFF
--- a/front/components/assistant/conversation/AgentMessage.tsx
+++ b/front/components/assistant/conversation/AgentMessage.tsx
@@ -1388,13 +1388,19 @@ function AgentMessageContent({
   );
 
   // Auto-open interactive content drawer when interactive files are available.
-  const { interactiveFiles } = useAutoOpenInteractiveContent({
+  const { interactiveFiles, willOpenInteractiveContent } =
+    useAutoOpenInteractiveContent({
+      agentMessage,
+      isLastMessage,
+    });
+
+  // Auto-open file explorer panel when regular generated files are available,
+  // unless the same message is opening a Frame in the side panel.
+  useAutoOpenFilesPanel({
     agentMessage,
     isLastMessage,
+    deferToInteractiveContent: willOpenInteractiveContent,
   });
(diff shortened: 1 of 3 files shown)


frankaloia:
Superseded by #29413

PR #29409, opened 2026-07-23
Improve OpenAI prompt cache reuse across conversations

flvndvd:
## Description

OpenAI prompt caching requires requests to share both an identical prompt prefix and a stable cache key. We
currently use the conversation identifier as that key, which prevents cache reuse between separate conversations
even when their instructions, tools, context, and explicit cache breakpoints are identical. See documentation [here](https://developers.openai.com/api/docs/guides/prompt-caching).

This was observable in two requests made minutes apart: their reusable prefixes were byte-for-byte identical, but
their cache keys differed because they originated from different conversations.

## Change

Streaming OpenAI requests now use `<workspaceId>:<agentConfigurationId>` as their cache key.

This allows conversations using the same agent in the same workspace to reuse cached prompt prefixes. Workspace
isolation is preserved, and partitioning by agent reduces.

> [!WARNING]
> OpenAI's recommendation is approximately 15 requests per minute for a single key. We will monitor cache usage for it.

The key does not permit incorrect cache reuse. OpenAI still requires an exact prefix match, including instructions,
messages, tools, schemas, ordering, and cache breakpoints.

### DIFF
--- a/front/lib/api/llm/clients/openai/index.ts
+++ b/front/lib/api/llm/clients/openai/index.ts
@@ -33,6 +33,7 @@ import {
   responseToLLMEvents,
   streamLLMEvents,
 } from "@app/lib/api/llm/utils/openai_like/responses/openai_to_events";
+import { getOpenAIPromptCacheKey } from "@app/lib/api/llm/utils/prompt_cache_key";
 import { TOOL_SEARCH_INSTRUCTION } from "@app/lib/api/llm/utils/tool_search";
 import type { Authenticator } from "@app/lib/auth";
 import logger from "@app/logger/logger";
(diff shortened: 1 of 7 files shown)


front/lib/api/llm/clients/openai/index.ts
```diff
@@ -185,9 +188,10 @@ export class OpenAIResponsesLLM extends LLM<ResponseCreateParamsStreaming> {
     const lines = Array.from(conversations.entries()).map(
       ([customId, streamParams]) => {
         const body = {
-          ...this.buildRequestPayload(streamParams, {
-            conversationId: customId,
-          }),
+          ...this.buildRequestPayload(streamParams),
```
flvndvd, line 191: @pmilliotte you had authorship on using `conversationId` do you recall the reason?

pmilliotte: Not thinking about sharing cache between convos at that time

pmilliotte (approved):

aubin-tchoi:
> OpenAI's recommendation is approximately 15 requests per minute for a single key. We will monitor cache usage for it.

what happens if we go over 15 rq/min? do we have the right monitoring for this?

flvndvd:
> > OpenAI's recommendation is approximately 15 requests per minute for a single key. We will monitor cache usage for it.
> 
> what happens if we go over 15 rq/min? do we have the right monitoring for this?

They mentioning routing to other places. No easy way to monitor easily, but we have global cache hit so we should be able to perceive a trend.

PR #29411, opened 2026-07-23
Harden sandbox uid 1002 service boundary

fontanierh:
## Description

- Run every non-root E2B exec through a clean inner shell with pinned `PATH`/`HOME`, and pin E2B's outer login-shell environment to the root-owned `/var/empty` home so attacker-controlled profiles and shell hooks cannot run first.
- Move the `agent` service account's login home to `/var/empty`, give `agent-proxied` its own primary group, and lock `/home/agent`, `/opt/venv`, the GCS token broker, and Dust profile scripts against workload writes.
- Apply synthetic DNS, TCP forwarding, private-network drops, and runtime health checks to both controlled UIDs (`agent`/1002 and `agent-proxied`/1003).
- Keep the root-owned GCS FUSE token broker on real DNS through the local systemd-resolved stub while redirecting both controlled UIDs to the synthetic resolver.
- Deny both controlled UIDs access to systemd-resolved resolver IPC (D-Bus and Varlink), keeping real DNS reserved for root-owned services.
- Extend the live image security suite with service-path write denials, attacker-profile non-execution, metadata/DNS probes, root GCS connectivity, and GCS token refresh across E2B sleep/wake.

## Tests

- 69 focused Vitest tests passed across the E2B provider, image registry, egress setup, and GCS mount adapter.
- The broader sandbox suite passed: 334 tests passed and 8 provider-dependent tests were skipped.
- Full Front `tsgo --noEmit`, Biome formatting/lint checks, shell syntax validation, and `git diff --check` passed.
- Added live regression probes for resolve1 D-Bus and Varlink denial for both controlled UIDs before and after E2B sleep/wake, while verifying root resolver access and the private Varlink socket mode.
- The complete live security suite passed in EU and US on the pre-rebase `dust-base:0.8.60` build ([regional sandbox image workflow](https://github.com/dust-tt/dust/actions/runs/30453249458)). After rebasing onto `main`, the image is re-tagged **`0.8.66`** (`main` advanced the base to `0.8.64` with dbt/Snowflake CLIs and dsbx `0.1.41`, all now inherited); **`0.8.66` must be rebuilt and re-validated in both E2B regions before Front deploys.**

## Risk

Medium. This changes sandbox account ownership, shell startup behavior, DNS/egress boot services, and the base image tag. The regional live workflow fails if root GCS connectivity, controlled DNS/egress, resolver IPC isolation, service-path permissions, profile isolation, or sleep/wake token refresh regresses.

The compatibility tradeoff is intentional: workloads can no longer modify shared service paths such as `/opt/venv`; workload-owned packages and virtual environments must live under `/home/agent-proxied`.

## Deploy Plan

1. Confirm the validated `dust-base:0.8.66` templates remain available in both E2B regions before Front deploys. The existing deploy workflow gates Front on sandbox image availability.
2. Deploy Front using `dust-base:0.8.66`.
3. Immediately after the Front deploy, invoke `/poke/kill` for every `dust-base` version older than `0.8.66`. This is mandatory because sleeping sandboxes retain their original image.
4. Verify the kill requester completed successfully and that no non-deleted sandbox remains on an older `dust-base` version before closing the incident.

### DIFF
--- a/front/lib/api/sandbox/image/egress/egress-nftables.sh
+++ b/front/lib/api/sandbox/image/egress/egress-nftables.sh
@@ -1,9 +1,16 @@
 #!/bin/bash
 set -eu
 
-PROXIED_UID=1003
+CONTROLLED_UIDS="1002 1003"
 DNS_STUB_PORT=1053
 GCS_TOKEN_SERVER_PORT=987
+SYSTEM_RESOLV_CONF=/run/systemd/resolve/stub-resolv.conf
+
+# Keep the machine-wide resolver local. Service users not covered by the
+# sandbox egress boundary use systemd-resolved, while controlled UIDs are
+# redirected below to the synthetic resolver before reaching the local stub.
+test -s "$SYSTEM_RESOLV_CONF"
+/bin/ln -sfn "$SYSTEM_RESOLV_CONF" /etc/resolv.conf
 
 # Reinstall a single canonical ruleset on every sandbox boot.
 nft delete table ip dust-egress 2>/dev/null || true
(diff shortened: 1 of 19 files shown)


front/lib/api/sandbox/image/egress/egress-nftables.sh
```diff
@@ -14,36 +21,39 @@ nft add chain ip dust-egress nat_output '{ type nat hook output priority -100 ;
...
-nft add rule ip dust-egress filter_output meta skuid $PROXIED_UID ip daddr 172.16.0.0/12 drop
-nft add rule ip dust-egress filter_output meta skuid $PROXIED_UID ip daddr 192.168.0.0/16 drop
-nft add rule ip dust-egress filter_output meta skuid $PROXIED_UID meta l4proto udp drop
-nft add rule ip dust-egress filter_output meta skuid $PROXIED_UID meta l4proto icmp drop
-
 nft add table ip6 dust-egress
 nft add chain ip6 dust-egress filter_output '{ type filter hook output priority 0 ; policy accept ; }'
-nft add rule ip6 dust-egress filter_output meta skuid $PROXIED_UID drop
+
+for CONTROLLED_UID in $CONTROLLED_UIDS; do
```
flvndvd, line 27: Assuming no changes with the previous outside of the for loop, right?

front/lib/api/sandbox/image/registry.ts
```diff
@@ -341,9 +340,12 @@ const DUST_BASE_IMAGE = SandboxImage.fromDocker(
   .runCmd(getDustStateUserSetupCommand(), { user: "root" })
   .runCmd(getPodStateSetupCommand(), { user: "root" })
   // Hidden tools: installed but not in manifest (back profile functions)
-  .runCmd("apt-get update && apt-get install -y ripgrep fd-find sd", {
-    user: "root",
-  })
+  .runCmd(
+    "apt-get update && apt-get install -y ripgrep fd-find sd systemd-resolved",
```
flvndvd, line 344: [TO DOUBLE CHECJK] Checking with an agent, it looks like installing `systemd-resolved` creates a real-DNS bypass for `agent` and `agent-proxied`?! Their direct port-53 traffic is redirected by `nftables`, but they can call `systemd-resolved` through its unprivileged D-Bus API (`resolvectl`). The DNS request then leaves for `8.8.8.8` under the resolver’s UID (in this case I assume that `systemd-resolved` runs as its own service user, typically `systemd-resolve`) which is not controlled. Could we restrict resolver IPC to root or use a root-only resolver, and add a regression test?

front/lib/api/sandbox/image/egress/egress-nftables.sh
```diff
@@ -14,36 +21,39 @@ nft add chain ip dust-egress nat_output '{ type nat hook output priority -100 ;
...
+  # DNS interception must run before loopback/private exemptions and before
+  # the broad TCP redirect so every port-53 packet lands on the local stub.
+  nft add rule ip dust-egress nat_output meta skuid $CONTROLLED_UID udp dport 53 redirect to :$DNS_STUB_PORT
```
flvndvd: I recall some rule we had to use the full path for bins, is this still a thing?

flvndvd (approved):
LGTM, left one interesting findings. Feel free to merge if not valid.

front/lib/api/sandbox/image/registry.ts
fontanierh: Interesting. Digging 🙏

fontanierh: Confirmed 👍  

Plugging that hole. Curious how you found it ? just analyzing source code or sandbox escape pentest ?

flvndvd: Nope looking at the code and what each services added in this PR was doing to make sure I had the full picture, greatly helped my Sol (xhigh) though!

flvndvd (approved):
LGTM

front/lib/api/sandbox/image/egress/egress-nftables.sh
fontanierh: only change is PROXIED_UID=1003 → CONTROLLED_UIDS="1002 1003", i.e. the same rules now also cover uid 1002

front/lib/api/sandbox/image/egress/egress-nftables.sh
fontanierh: indeed this is not fully compliant with SEC3, interesting that it didn't catch it.

Adding

PR #29414, opened 2026-07-23
[Frame] merge side-panel open hooks into single priority-based hook

frankaloia:
## Summary

When an agent writes an interactive content file (a Frame) **and** a regular file (e.g. `AGENTS.md`) in the same turn, the Frame side panel would flash open and then get replaced by the file explorer.

Root cause: two auto-open hooks fire on the same message. `useAutoOpenInteractiveContent` opens the Frame in the side panel, then `useAutoOpenFilesPanel` runs and clobbers it with the file explorer.

Fix: `useAutoOpenInteractiveContent` now exposes `willOpenInteractiveContent` — the single source of truth for whether a Frame is opening for this message (covering both the streaming-progress and completed-file paths). `useAutoOpenFilesPanel` defers to that flag instead of re-deriving its own guess from `generatedFiles`. This keeps the Frame panel in front and also handles the streaming-progress window (where the Frame is opening from progress notifications before the file lands in `generatedFiles`).

## Risk
Low, but will change the global convo UX behavior

## Testing
Manual

## Deployment
1. Deploy front


front/components/assistant/conversation/useAutoOpenSidePanel.ts
```diff
@@ -107,13 +112,43 @@ export function useAutoOpenInteractiveContent({
...
-    interactiveFiles: completedInteractiveFiles,
-  };
+  // Whether this message is opening (or will open) the interactive content panel.
+  // Used to suppress the lower-priority file-explorer auto-open.
+  const willOpenInteractiveContent =
+    !isMobile &&
+    (interactiveFilesFromProgress.length > 0 ||
+      (completedInteractiveFiles.length > 0 && isLastMessage));
+
+  // Priority 2: file explorer — only when no interactive content is taking the panel.
```
ykmsd: Isn't it easier to manage if we have one useEffect hook for priority 1 & 2? 🤔

ykmsd (approved):

front/components/assistant/conversation/useAutoOpenSidePanel.ts
frankaloia: Yea I like that better

PR #29416, opened 2026-07-23
Admin governance: only suggest similar skill among published skills

fabiencelier:
## Description

closes https://github.com/dust-tt/tasks/issues/9732

Only compare the skills to the published ones, 2 users can have the same duplicate unpublished skills.

## Tests

added a unit test.

## Risk

low, easy to revert

## Deploy Plan

deploy front


ykmsd (approved):
ty!

PR #29417, opened 2026-07-23
[Poke] Show actual group member counts

philipperolet:
## Description

Related to https://github.com/dust-tt/tasks/issues/9773

Poke group list and detail endpoints serialized groups with the default `memberCount` value, which is always zero. Return the number of active workspace members so Poke diagnostics agree with the displayed member rows.

The shared batched count now filters out revoked workspace members. The details endpoint reuses the length of its already-filtered member list.

## Risks

Blast radius: Member counts displayed in Poke and workspace group views.
Risk: very low, resource method getMemberCountsForGroups was not called before this PR

## Deploy Plan

- deploy front-api

## Test

- [x] Poke group list excludes revoked workspace members from its count
- [x] Poke group details exclude revoked workspace members from their count
- [x] Existing workspace group endpoint tests pass

### DIFF
--- a/front-api/routes/poke/workspaces/[wId]/groups/[groupId]/details.ts
+++ b/front-api/routes/poke/workspaces/[wId]/groups/[groupId]/details.ts
@@ -37,7 +37,7 @@ app.get(
 
     return ctx.json({
       members,
-      group: await group.toJSONWithMemberCount(auth),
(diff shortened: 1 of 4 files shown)


flvndvd (commented):
LGTM

PR #29418, opened 2026-07-23
[Conversations] Ignore deleted history when detecting active agents

philipperolet:
## Description

Fixes https://github.com/dust-tt/tasks/issues/9777.

When a user deletes a turn, the deletion is stored as a newer deleted message version. The older agent-message version can legitimately remain in `created` while its workflow stops.

`postUserMessage` was flattening the full version history when looking for a running agent. It therefore found that hidden old version and stored every later user message as pending, even though the latest version said the agent reply was deleted.

This change checks only the latest version at each rank when deciding whether an agent is running. The existing deletion flow still signals a live workflow to stop, but historical versions no longer keep the queue blocked.

## Risks

Blast radius: steering detection in conversations containing multiple versions of the same message.
Risk: low. The latest version is the current conversation state; older versions are retained as history.

## Deploy Plan

- Deploy front.

## Tests

- [x] `front/lib/api/assistant/conversation.test.ts` (86 tests)


spolu (commented):
LGTM

PR #29360, opened 2026-07-23
[front] perf: conversation-scoped side table joins in message fetches

id13:
## Description

Adding the where clause (which is in fact an ON clause ... but sequelize ...) to all side joins

## Tests

Existing tests

## Risk

Low - we're narrowing the lookups required - indexes are already there, merged only after backfill ran
Blast radius: Given how critical this query is, fair to say most of the platform 😬 

## Deploy Plan

- Run backfill from #29359 
- Merge
- Deploy

PR #29419, opened 2026-07-23
feat(conversations): add ConversationResource query methods

Fraggle:
## Description

Adds targeted query methods to `ConversationResource` to replace the pattern of loading full `conversation.content` in mutation and validation paths. New methods cover the precise read operations callers actually need:

- `resolveMessageViewScope` (private) — branch-aware Sequelize + raw SQL filter, single implementation shared by all query helpers
- `getMessageScopeWhere` — Sequelize `WhereOptions<MessageModel>` for branch-scoped message queries
- `hasUserMessageFromOtherUser` / `hasAgentMessageAfterRank` — existence checks via parameterized SQL `EXISTS`
- `getLatestUserMessageModelAtRank` — latest version of a user message at a given rank
- `getClientSideMCPServerIdsFromLatestNonWakeUpUserMessage` — inherits client-side MCP selection from the previous human turn on wake-up posts
- `getConsecutiveAgentReplyModelsAfterRank` — consecutive agent replies for soft-delete cascade
- `getRunningAgentMessage` / `getRunningCompactionMessage` / `getInFlightMessages` — in-flight message detection
- `getBranchCreationContext` — aggregated stats for branch pre-conditions
- `getLatestMessageSummary` — last visible message with compaction status

Also exports four new types (`RunningAgentMessageContext`, `RunningCompactionMessageContext`, `BranchCreationContext`, `LatestMessageSummary`), widens `conversation` params on read/participant mutation methods to `ConversationWithoutContentType | ConversationResource`, and adds `isPodConversation()` + `spaceSId` getter.

## Tests

Tested locally. These are pure query additions with no mutations; no existing behaviour changes. Callers will be wired in follow-up PRs.

## Risk

Low. No schema changes, no existing logic altered. New methods are additions only; the widened union types on mutation signatures are backward-compatible. Safe to rollback.

## Deploy Plan

Deploy `front`.


Fraggle:

- #29419 ◀
    - #29420
        - #29421
            - #29422
                - #29423
                    - #29424
                        - #29425

front/lib/resources/conversation_resource.ts
```diff
@@ -2807,7 +2834,7 @@ export class ConversationResource extends BaseResource<ConversationModel> {
       transaction,
       lastReadAt = new Date(),
     }: {
-      conversation: ConversationWithoutContentType;
```
id13, line 2819: why is this needed ? (same for 3 other above)

front/lib/resources/conversation_resource.ts
```diff
@@ -2984,6 +3011,619 @@ export class ConversationResource extends BaseResource<ConversationModel> {
...
+        scopeWhere: {
+          ...baseWhere,
+          branchId: { [Op.is]: null },
+        },
+        branchFilterSql: `"branchId" IS NULL`,
+        sqlReplacements: {},
+      };
+    }
+
+    const branch = await ConversationBranchResource.fetchById(auth, branchId);
```
id13: missing transaction propagation

front/lib/resources/conversation_resource.ts
```diff
@@ -2984,6 +3011,619 @@ export class ConversationResource extends BaseResource<ConversationModel> {
...
+
+    const query = `
+      SELECT EXISTS (
+        SELECT 1
+        FROM (
+          SELECT DISTINCT ON (m.rank) um."userId"
+          FROM messages m
+          INNER JOIN user_messages um
+            ON um.id = m."userMessageId"
+            AND um."workspaceId" = m."workspaceId"
```
id13, line 3149: conversationId will go a long way here, I suggest putting it, and remove workspaceId

front/lib/resources/conversation_resource.ts
```diff
@@ -2984,6 +3011,619 @@ export class ConversationResource extends BaseResource<ConversationModel> {
...
+  ): Promise<boolean> {
+    const owner = auth.getNonNullableWorkspace();
+    const { branchFilterSql, sqlReplacements } =
+      await this.resolveMessageViewScope(auth, { branchId, transaction });
+
+    const query = `
+      SELECT EXISTS (
+        SELECT 1
+        FROM (
+          SELECT DISTINCT ON (m.rank) m."agentMessageId"
```
id13, line 3203: limit 1 + your order removes the need for a distinct IIUC, which is an expensive op.
Same for other raw SQL queries above

front/lib/resources/conversation_resource.ts
```diff
@@ -2984,6 +3011,619 @@ export class ConversationResource extends BaseResource<ConversationModel> {
...
+      SELECT latest."clientSideMCPServerIds"
+      FROM (
+        SELECT DISTINCT ON (m.rank)
+          m.rank,
+          um."userContextOrigin",
+          um."clientSideMCPServerIds"
+        FROM messages m
+        INNER JOIN user_messages um
+          ON um.id = m."userMessageId"
+          AND um."workspaceId" = m."workspaceId"
```
id13, line 3295: `+ conversationId, - workspaceId`

front/lib/resources/conversation_resource.ts
```diff
@@ -2984,6 +3011,619 @@ export class ConversationResource extends BaseResource<ConversationModel> {
...
+        visibility: { [Op.ne]: "deleted" },
+      },
+      include: [
+        {
+          model: AgentMessageModel,
+          as: "agentMessage",
+          required: true,
+          attributes: ["id", "agentConfigurationId"],
+          where: { status: "created" },
+        },
```
id13, line 3433: where conversationId

front/lib/resources/conversation_resource.ts
```diff
@@ -2984,6 +3011,619 @@ export class ConversationResource extends BaseResource<ConversationModel> {
...
+        ...scopeWhere,
+        visibility: { [Op.ne]: "deleted" },
+      },
+      include: [
+        {
+          model: CompactionMessageModel,
+          as: "compactionMessage",
+          required: true,
+          attributes: ["id"],
+          where: { status: "created" },
```
id13: conversationId

id13 (approved):
LGTM, I think tests are needed here

front/lib/resources/conversation_resource.ts
Fraggle: The upcoming changes will operate directly on an ConversationResource instance.
In general, ConversationWithoutContentType | ConversationResource are interchangeable when you don't need to content but just some properties.

front/lib/resources/conversation_resource.ts
Fraggle: Good catch

front/lib/resources/conversation_resource.ts
Fraggle: As I said, didn't exists while I was writing the code but sure :-)

Why not both ?

front/lib/resources/conversation_resource.ts
Fraggle: 👍🏼 
(I assume psql would optimize it anyway)

front/lib/resources/conversation_resource.ts
Fraggle: Q?
How would it be faster since we are using the primary ID for lookup anyway ?

front/lib/resources/conversation_resource.ts
Fraggle: After some checking, sorry my friend:
"DISTINCT ON here is “latest version per rank”, not optional dedupe. Only the MCP-ids query has outer LIMIT 1, and even there ORDER BY+LIMIT alone isn't equivalent because wakeup must be applied after version selection. The EXISTS / aggregate queries need one row per rank, so LIMIT 1 doesn't remove DISTINCT ON."

id13: I think it's AI justifying its work 😄 

What do we want to know ? whether there is an agent message after a given rank.

So you can throw away version here, we don't change the nature of a rank (user or agent) once it has been given.

It just becomes an exists where rank > afterRank no? 

I think this kind of simplification can be done elsewhere in the methods, but not sure.

front/lib/resources/conversation_resource.ts
id13: > How would it be faster since we are using the primary ID for lookup anyway ?

It's all about cost functions in the pg planner. More tokens here : https://github.com/dust-tt/dust/pull/29292
(replace workspaceId by conversationId). 


> Why not both ?

In theory it's useless, but since our index is on workspaceId, conversationId, you need both I was wrong, same for all my recommendation on adding conversationId

id13: see here https://github.com/dust-tt/dust/pull/29360/changes

front/lib/resources/conversation_resource.ts
id13: wrong, you need both

front/lib/resources/conversation_resource.ts
id13: and workspaceId

front/lib/resources/conversation_resource.ts
id13: and workspaceId

front/lib/resources/conversation_resource.ts
Fraggle: I kept them both.

PR #28893, opened 2026-07-13
[connectors] fix(slack): render block/attachment content in thre…

TimotheeCrouzet:
…ad docs

## Description

fixes: [https://github.com/dust-tt/tasks/issues/9524](https://github.com/dust-tt/tasks/issues/9524)

Bot/app messages from alert tools (Grafana, Datadog, Zendesk, …) often ship an empty
top-level `text`, their content living in Block Kit `blocks`/`attachments`. When such a
message is upserted into a Slack data source, the body ended up empty.

This ports the Slack message formatter from front ([PR](https://github.com/dust-tt/dust/pull/27697) into connectors and
wires it into `formatMessagesForUpsert`, reconstructing readable text from `blocks` and
`attachments`. Forwarded (unfurl) attachments keep their dedicated rendering and are excluded
from the formatter input to avoid duplication.

## Tests

Added `messages_formatter.test.ts` 

## Risk

Low. Safe to revert.

PR #29428, opened 2026-07-23
dedup slack threads messages

TimotheeCrouzet:
## Description
takes care of a regression which I introduced in [#28893](https://github.com/dust-tt/dust/pull/28893).

Some messages written by people showed up twice in the context agents receive. Because, to read block-only bot alerts (Grafana/Datadog/Zendesk), that PR started rendering `text`, `blocks` and `attachments` together. But `text` is Slack's *fallback* for `blocks`: the same content is carried in both, so rendering both is redundant.

## Fix
When `blocks` is not present we fall back to `text`.

Edit: while implementing, I looked at real Slack payloads and realized that mentions in text don't come with a label after the id giving the name, it's <@ U123>, not <@ U123|bob> (same for <#C123> channels). So we can't parse text to output the username/channel name. We have to resolve the id through the API either way. The parsing would have been the only advantage text had over blocks (one fewer API call per user/channel mention).

Conclusion: Since `getUserInfo` was already being used I had nothing to do for user mentions. I just added `getChannelNameById` which wraps `getChannelById` (pre-existed) to solve channel mentions through their ids. This way, outputting `blocks` is even better because we don't solve channel mentions in `text`.

## Tests

added connectors/src/connectors/slack/lib/messages.test.ts and tested locally:
--> no deduped messages in connectors logs.

PS: I could not confirm against a real payload, but the "[no preview available]" is probably from Slack, in case the unfurl link can't be solved (it doesn't come from us at least). 

## Risk

Low. Safe to revert

### DIFF
--- a/connectors/src/connectors/slack/lib/messages.ts
+++ b/connectors/src/connectors/slack/lib/messages.ts
@@ -42,8 +42,8 @@ async function processMessageForMentions(
 }
 
 // Single entry point to turn a Slack message into its upsert body text: resolves
-// mentions, splits forwarded (unfurl) attachments from regular ones, reconstructs
-// the content from text/blocks/attachments/files, and appends forwarded messages.
+// mentions, uses the top-level `text` when present (else reconstructs the content
+// from its Block Kit blocks/attachments), and appends forwarded (unfurl) messages.
 async function formatSlackMessageBody(
   message: MessageElement,
   connectorId: ModelId,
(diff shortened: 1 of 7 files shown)


connectors/src/connectors/slack/lib/messages.ts
```diff
@@ -66,14 +66,19 @@ async function formatSlackMessageBody(
...
-    formatted.files,
-  ]
+  // `text` is Slack's fallback for `blocks`: a message typed in Slack repeats its
+  // content in both, so rendering both would duplicate it. When `text` is present
+  // it is also the better-rendered version (mentions resolved to names), so we use
+  // it alone. Only when `text` is empty (bot alerts whose content lives entirely in
+  // Block Kit) do we reconstruct from blocks/attachments.
+  const messageContent =
+    formatted.text !== EMPTY_SECTION
+      ? [formatted.text]
```
flvndvd: I'm not sure I follow here, it was already the case before https://github.com/dust-tt/dust/pull/28893, no? Shouldn't we try to compute the blocks and if it fails, then fallback to text?

TimotheeCrouzet: Before #28893 we only rendered text, so bot alerts with an empty text came through empty. I just added a fallback to blocks for those. Normal messages aren't affected, there blocks is just a duplicate of text, so they still go through text.

On "compute blocks and fall back to text": I looked into it, and I actually started from a wrong assumption (which I had since my first task on slack a few weeks ago). I thought that in text mentions came with the label @ 123 | bob, so we'd juste have to parse text to get userId. Actually, mentions come through as raw ids in text too (<@ U123>, <#C123>, <!subteam^S123>), there's no label. So outputting text has no real benefits over blocks since we have to solve the mentions anyway.

Where we stand on resolving: users and channels are fine, we already have getUserInfo and getChannelById. The open question is usergroups: there's no endpoint to get one from an id, we'd have to pull them all from usergroups.list, and usergroups:read isn't in our scopes. Leaving them as raw ids in the meantime is no worse than today, so it's not a blocker.

TimotheeCrouzet:
TLDR: I realized that Slack sends mentions with no label (`<@U123>`, not `<@U123|bob>`) as bare ids in `text` just like in `blocks` so we can't parse `text` to translate an id into a username/channel. Outputting `text` has no real benefit over `blocks` then. I added `getChannelNameById` so we don't output a raw id, which made defaulting to `blocks` even better than `text`.

spolu:
@flvndvd still a bug?

flvndvd:
> @flvndvd still a bug?

Yes very likely! Happy to hand off this to eng runner or take it as part of my oncall shift.

spolu:
Yep let's make sure it's unassigned from @TimotheeCrouzet 👍

spolu:
(or create an eng-runner card if it does not exist... could not find it?)

PR #29430, opened 2026-07-23
Admin governance: availability in v1/w/[wId]/skills endpoint

fabiencelier:
## Description

Expose the availability as a filter in the public  v1/w/[wId]/skills API.
API do not cary the info about a user so it can never be an editor. 

## Tests

unit test added 

## Risk

low, the public API is retrocompatible

## Deploy Plan

deploy front

### DIFF
--- a/front-api/routes/v1/w/[wId]/skills/index.ts
+++ b/front-api/routes/v1/w/[wId]/skills/index.ts
@@ -6,6 +6,7 @@ import type { ImportSkillsResponseBody } from "@app/lib/api/skills/detection/git
 import { MAX_ZIP_SIZE_BYTES } from "@app/lib/api/skills/detection/zip/detect_skills";
 import { SkillResource } from "@app/lib/resources/skill/skill_resource";
 import type { SkillType } from "@app/types/assistant/skill_configuration";
+import { SKILL_AVAILABILITIES } from "@app/types/assistant/skill_configuration";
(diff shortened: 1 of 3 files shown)


front-api/routes/v1/w/[wId]/skills/index.ts
```diff
@@ -156,11 +173,37 @@ app.get(
...
+
+    const allSkills = await SkillResource.listByWorkspace(auth, {
       status,
       onlyCustom: true,
+      availability,
     });
 
+    // Unpublished (editors-only) skills are drafts and are never exposed through the
+    // public API: API keys have no editor-group membership to scope them by.
+    const skills = allSkills.filter(
```
Comment (ykmsd) [inline front-api/routes/v1/w/[wId]/skills/index.ts]: I'm actually not sure how it's supposed to behave here, editors only skills are not meant be confidential, if a user requests it should we return it? 🤔 Or maybe we can return it only for admins, probably if it's a big workspace, admins likely want to get the list from public API?

Comment (ykmsd) [reply front-api/routes/v1/w/[wId]/skills/index.ts]: because the end goal is exposing editors skills to admin from skill management page so it's kind of make sense to align?

ykmsd (approved):

front-api/routes/v1/w/[wId]/skills/index.ts
Comment (fabiencelier) [reply front-api/routes/v1/w/[wId]/skills/index.ts]: makes sense indeed to expose unpublished skills to admins
They van filter them out with the availability param if they don't want them

PR #29347, opened 2026-07-22
feat(sandbox): egress domain-request admin surface (pod + workspace)

smb2268:
## Description

Follow up of `30474` (stacked on `30529`). Lights up the admin surface that reviews egress domain requests, for both the Pod and workspace allowlists.

Mirrored across the pod route (`/spaces/:spaceId/sandbox/egress-policy`) and the workspace route (`/sandbox/egress-policy`):

1. **Read.** The egress-policy `GET` sources pending requests from the same policy-file read that returns the allowlist (they ride in the same file; the proxy ignores the section) — no second fetch.
2. **Reject.** A new `POST /requests/dismiss` removes a request without granting it. Workspace-admin gated by the parent sandbox sub-app, same as the allowlist `PUT`.
3. **Approve.** No new path — `PUT`ting the allowlist with the domain appended resolves its request atomically (the write preserves the section; normalize drops the now-allowed request).
4. **Front.** `useDismissPodEgressRequest` / `useDismissWorkspaceEgressRequest` hooks + Reject wired into `PodNetworkSection` and the workspace `NetworkSection`; the shared `EgressDomainListEditor` renders the "Pending approval" rows with Approve/Reject buttons. The update hooks preserve `requestedDomains` on optimistic mutate so approving one request doesn't blank the others until revalidation.

## Risk

Low. Connects a previously inert surface to real data behind the existing workspace-admin gate. Dismiss grants nothing (never touches `allowedDomains`). Safe to rollback.

## Tests
Screenshot 2026-08-14 at 2 11 22 PM
Screenshot 2026-08-14 at 2 11 05 PM

Pod + workspace routes: pending requests surfaced from the file; dismiss removes a request without touching the allowlist; approve-via-PUT resolves atomically; non-admin dismiss → 403; missing-domain dismiss → 400; plus the existing GET/PUT/403/non-project cases. 20 green.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #29346, opened 2026-07-22
feat(sandbox): audit pod sandbox settings updates (env vars + egress policy)

smb2268:
## Description

Follow up of `29349`. All pod-scoped audit logic for the sandbox settings surfaces in one PR: every CTA on the pod settings page (secrets and domains — add, edit, remove) emits an audit event that identifies its pod.

**What this PR does:**

1. **Env var events gain `space_id`**: `auditMetadataBase()` on the unified resource adds the pod space sId to all `sandbox_env_var.*` events for pod-scoped rows (workspace rows omit the key — metadata values must be strings, never null). The five schemas gain `space_id`; versions bumped.
2. **Pod egress-policy PUT now emits** `sandbox_egress_policy.updated` — previously the one admin mutation on the pod surface with no audit event. Same action as the workspace surface, with `space_id` metadata; schema + version updated.
3. **Tests**: the pod env-var create round-trip asserts the emit carries `space_id`; the pod egress-policy round-trip asserts the emit with domains + `space_id` (both mirror the workspace egress-policy test's audit-mock pattern).

**CTA → audit event coverage** (pod settings, sandbox sections):

| CTA | Event |
|---|---|
| Add / replace env var | `sandbox_env_var.created` / `.updated` (+ `space_id`) |
| Secure (promote to HTTPS secret) | `sandbox_env_var.promoted_to_https_secret` (+ `space_id`) |
| Edit secret allowed domains | `sandbox_env_var.allowed_domains_updated` (+ `space_id`) |
| Delete env var | `sandbox_env_var.deleted` (+ `space_id`) |
| Add / remove network domain | `sandbox_egress_policy.updated` (+ `space_id`) |

The workspace-page equivalents were already covered on main.

## Risk

Worst case, a malformed audit event — emission is fire-and-forget and never blocks the mutation. Safe to rollback.

## Deploy Plan

- [ ] Re-register audit schemas after merge: `npx tsx front/admin/register_audit_log_schemas.ts --execute --changed`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #29431, opened 2026-07-23
feat(sandbox): conversation-scoped egress approvals with inherited pod policy layer

smb2268:
## Description

Follow up of `29347`. Lands the approval-governance decision for conversations inside Pods: **every conversation sandbox gets its own conversation-scoped egress policy file — inside or outside a Pod — and on-the-fly domain approvals land there.** Pod network settings still apply to every Computer in the Pod, as an inherited read-only layer. This replaces the interim v0 behavior where a conversation inside a Pod wrote approvals directly into the Pod's shared policy file (any pod member's tool click widened egress for the whole Pod).

**What this PR does:**

1. **Proxy: inherited pod layer.** The egress JWT gains an optional `podId` claim; the proxy allows a domain if the workspace, owner, **or pod** policy allows it (same owner-file scheme, one more fail-closed lookup). Old proxies ignore the claim; old tokens carry none — both directions are compatible.
2. **Front: conversation-scoped policy owner.** `ensureConversationSandboxReady` sets `egressPolicyOwnerId` to the conversation sId always, and `egressPolicyPodId` to the Pod's space sId when the conversation lives in one — threaded through egress setup/wake/health-restart into the token.
3. **Approvals route to the conversation.** The `add_egress_domain` tool writes the conversation's own file in all cases; the tool description and agent-facing response copy now say conversation scope, dropping the "Pod-wide, make that clear to the user" framing.

Domains already written into Pod files under the v0 behavior stay honored (they're in the inherited layer) — no data migration.

Pod-level allowlist additions remain admin-only via Pod network settings; the request-manifest flow for pod-function domain requests is still to be built (`29347` carries its approval surface).

## Risk

Worst case, a conversation in a Pod loses the pod layer (falls back to workspace + own approvals — fail closed, no over-granting). Safe to rollback; rollback order mirrors the deploy order below.

## Deploy Plan

> [!WARNING]
> Deploy `egress-proxy` **before** front: old proxies ignore the `podId` claim, so if front ships first, conversations inside Pods won't inherit Pod domains until the proxy catches up (nothing over-grants; they keep workspace + own approvals).

- [ ] Deploy egress-proxy
- [ ] Deploy front

## Tests

Proxy: `podId` claim validation round-trip (79 total green). Front: lifecycle asserts conversation-scoped owner + inherited pod layer in both egress calls; tool test flipped to assert approvals write the conversation file with conversation-scope copy; full sandbox tool + egress suites green.


smb2268:

- #28861
    - #28862
        - #29348
            - #29346
                - #29347
                    - #29431 ◀

smb2268:
Reopening off `main` — this change has no dependency on the pod-secrets stack (it modifies the shipped v0 egress paths), so it shouldn't queue behind it.

PR #29432, opened 2026-07-23
feat(sandbox): conversation-scoped egress approvals with inherited pod policy layer

smb2268:
## Description

First step in bringing egress-approval behavior in line with [this Slack thread](https://dust4ai.slack.com/archives/C0AGY9MTNJV/p1784750813582049). In the first iteration we didn't give sandbox sessions inside a Pod their own policy file — on-the-fly domain approvals were routed straight into the Pod's shared file, so any pod member's tool click widened egress for the whole Pod.

This PR lands the agreed model: **every conversation sandbox gets its own conversation-scoped egress policy file — inside or outside a Pod — and on-the-fly approvals land there**, while Pod network settings apply to every Computer in the Pod as an inherited read-only layer. Pod-owned sandboxes (pod functions) are unchanged: their ownerId already is the Pod, so they see workspace + Pod allowances and get no per-session file. The proxy half (`podId` claim + inherited pod layer) is #30383, already merged; the move-recycling follow-up `30379` and the scope-fingerprint follow-up `30386` stack above.

**What this PR does:**

1. **Conversation-scoped policy owner.** `ensureConversationSandboxReady` sets `egressPolicyOwnerId` to the conversation sId always, and `egressPolicyPodId` to the Pod's space sId when the conversation lives in one — threaded through egress setup/wake/health-restart into the token.
2. **Approvals route to the conversation.** The `add_egress_domain` tool writes the conversation's own file in all cases; the tool description and agent-facing response copy now say conversation scope, dropping the "Pod-wide, make that clear to the user" framing.
3. **Cleanup on destroy.** Every conversation now owns a policy file, so `destroyConversation` deletes it for pod conversations too (the Pod's own file remains `hardDeleteSpace`'s job).
4. **Authoritative scope derivation.** `ensureConversationSandboxReady` re-reads the conversation's `spaceId` from the database instead of trusting the caller's object: the caller's conversation is typically an agent-loop snapshot, and a move (possibly issued by that very agent) can change the pod association mid-loop. `spaceId` is an authorization input (egress scope, pod env vars, pod mounts), so every sandbox create/wake derives it fresh.

Domains already written into Pod files under the v0 behavior stay honored (they're in the inherited layer) — no data migration.

Pod-level allowlist additions remain admin-only via Pod network settings; the request-manifest flow for pod-function domain requests is still to be built (`29347` carries its approval surface).

## Risk

Worst case, a conversation in a Pod loses the pod layer (falls back to workspace + own approvals — fail closed, no over-granting). Safe to rollback; rollback order mirrors the deploy order below.

Rollout: sandboxes already warm at deploy keep their pod-scoped token until they re-mint. The warm-fleet migration is the stacked `30386` (egress token scope fingerprint — split out for standalone design review): its NULL-mismatch check re-mints every pre-deploy sandbox on its next use. Until it deploys, the window closes at each sandbox's natural sleep/wake; nothing over-grants either way — Pod/workspace domains keep working and only new conversation-scoped approvals are inert on warm sandboxes.

Moves are handled across the stack: the authoritative fetch here means any sandbox create/wake derives scope from database state, and `30379` recycles the sandbox at move time so warm state (including pod env vars, which are injected only at creation) is rebuilt.

## Deploy Plan

- [x] Confirm the #30383 egress-proxy deploy is live in both regions (merged; old proxies ignore the `podId` claim, so if front ships first, conversations inside Pods fail closed to workspace + own approvals until the proxy catches up — no over-granting)
- [ ] Deploy front (ideally with the stacked `30386`, whose fingerprint migrates warm sandboxes to conversation-scoped tokens; without it, warm sandboxes converge at their natural sleep/wake)

## Tests

Proxy-side coverage (claim round-trip + the authorization matrix) lives in #30383. Front: lifecycle asserts conversation-scoped owner + inherited pod layer in both egress calls, with the pod case now moving the conversation in the database and passing a stale snapshot (pinning the authoritative fetch), plus the reverse direction (stale snapshot still carrying a pod the conversation left); `mintEgressJwt` asserts podId present when given and absent otherwise; conversation destroy deletes the conversation's own policy file and never the Pod's; tool test flipped to conversation-scope approvals and copy. Move-recycling tests live in the stacked `30379`.

### DIFF
--- a/front/lib/api/actions/servers/sandbox/metadata.ts
+++ b/front/lib/api/actions/servers/sandbox/metadata.ts
@@ -102,9 +102,9 @@ export const SANDBOX_TOOLS_METADATA = [
       "allowlist surface as denied entries in `<network_proxy_logs>` in " +
       "the bash tool output. Allowlist entries added through this tool " +
       "persist for the lifetime of the conversation (across sandbox " +
-      "restarts). In a Pod, approvals are Pod-wide: the domain is allowed " +
-      "for every conversation in the Pod and the Pod's shared sandbox, so " +
-      "make that scope clear when asking the user.",
+      "restarts) and are scoped to this conversation only — including " +
+      "inside a Pod, where Pod-wide domains are managed by admins in the " +
+      "Pod's network settings.",
     schema: {
       domain: z
         .string()
--- a/front/lib/api/actions/servers/sandbox/tools/index.test.ts
+++ b/front/lib/api/actions/servers/sandbox/tools/index.test.ts
@@ -933,7 +933,7 @@ describe("addEgressDomainTool", () => {
     }
   });
 
-  it("writes Pod-scoped approvals to the Pod's policy file for pod conversations", async () => {
+  it("keeps approvals conversation-scoped for conversations inside a Pod", async () => {
     const sandbox = {
       providerId: "provider-id",
       sId: "sandbox-id",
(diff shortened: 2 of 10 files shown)


smb2268:

- #29432 ◀

fontanierh (approved):
LGTM

PR #29433, opened 2026-07-23
refactor(sandbox): drop the Workspace prefix from scope-generic env var symbols

smb2268:
## Description

Mechanical rename, extracted from `28861` so the semantic PR reviews without rename noise: sandbox env var symbols that describe the entity (not its scope) drop the `Workspace` prefix, ahead of the unified-table stack making them serve both workspace and pod scopes.

**Renamed (pure find/replace, no behavior, no wire or DDL change):**

| Old | New |
|---|---|
| `WorkspaceSandboxEnvVarModel` (class + file) | `SandboxEnvVarModel` (`modelName`/table unchanged) |
| `WorkspaceSandboxEnvVarKind` / `WORKSPACE_SANDBOX_ENV_VAR_KINDS` | `SandboxEnvVarKind` / `SANDBOX_ENV_VAR_KINDS` |
| `render/parseWorkspaceSandboxEnvVarName*` | `render/parseSandboxEnvVarName*` |
| `Get/Post/Patch/DeleteWorkspaceSandboxEnvVarsResponseBody` | `Get/Post/Patch/DeleteSandboxEnvVarsResponseBody` |
| `WorkspaceSandboxEnvVarType` | `SandboxEnvVarType` |

**Deliberately kept `Workspace`** (genuinely workspace-scoped): `WorkspaceSandboxEnvVarResource` and the `useWorkspaceSandboxEnvVars` hook family — those become scope-generic (and renamed) in the stack PRs that make them so. The table name also stays: a DB rename is a dedicated migration with a compatibility-view window, not worth bundling with zero behavior.

> [!NOTE]
> Danger flags `**/models/**` changes for `migration-ack`, but **no migration exists or is needed**: only the TypeScript class and file were renamed — `modelName: "workspace_sandbox_env_var"` (and therefore the table, indexes, and constraints) is byte-identical, so Sequelize binds to the exact same DDL. The label acknowledges the check; there is nothing to apply.

`28861` is stacked directly on this PR and contains only the semantic changes.

## Risk

None — tsc-verified renames. Safe to rollback.

## Deploy Plan

No deploy steps: TypeScript-only rename — the table, indexes, constraints, and `modelName` binding are untouched, so there is no migration and nothing to apply. (The table itself renames in a dedicated follow-up PR with its own migration + bridge-view deploy plan, after the stack lands.)


smb2268:

- #29433 ◀

fontanierh (approved):
LGTM

front/lib/resources/storage/models/sandbox_env_var.ts
flvndvd, line 71: This is the key element here.

flvndvd (approved):

PR #29436, opened 2026-07-23
feat(agent-builder): use tiered model picker in advanced settings

Nils-Fedrigo:
## Description

Reworks the agent builder's model picker (the "Advanced" dropdown in the instructions block) to offer the same three tiers as the conversation input bar — **Fast** (`auto_fast`), **Standard** (`auto`), **Complex** (`auto_complex`) — instead of the old binary "Auto" toggle vs. a specific model. Standard is the default. The concrete "More models" selection (with the reasoning-effort slider) is kept alongside the tiers.

To avoid duplicating the picker, the shared picker family (`modelPickerUtils`, `ModelPickerContent`, `ModelPickerMoreModels`, `ModelPickerModelRow`, `ModelPickerSelectionIndicator`, `ReasoningEffortSlider`) is moved from `components/assistant/conversation/input_bar/` to `components/shared/model_picker/`. `ModelPickerContent` now also exports a `ModelPickerItems` body that can be embedded in another menu, with the default/revert affordances made optional so the agent builder (which edits the default itself) renders without them. The two duplicated tier icon maps and the maker-grouping helper are consolidated into the shared module.

Selecting a tier stores the sentinel (e.g. `{ modelId: "auto_fast", providerId: "auto_fast" }`) on the agent's `generationSettings.modelSettings`, exactly like the input bar sends it per message. Existing agents saved with `auto` display as Standard. Gated behind the existing `models_picker` feature flag; when the flag is off, the old submenus render unchanged.

No backend changes: `auto_fast`/`auto_complex` are already accepted by the agent save path and resolved to a concrete model at message-send time.

## Tests

Manual, on a `models_picker`-flagged workspace:
- New agent → Instructions → Advanced shows Fast/Standard/Complex with Standard preselected; picking a tier keeps the menu open.
- "More models" search + reasoning-effort slider work; picking a concrete model clears the tier selection.
- Saving with a tier round-trips (reload shows the same tier); the agent resolves a concrete model when messaged.
- Existing `auto` agents open as Standard.
- Input bar picker behavior is unchanged after the file moves/refactor.
- Flag-off workspace shows the previous submenus.


Nils-Fedrigo:

- #29436 ◀

PR #29437, opened 2026-07-23
[front] fix: Putback viewportRef to fix infinite scroll

id13:
## Description

In some rare cases, we are not correctly inferring hasMore because we evaluate it after filtering some rows in JS 🤡 

It's not longer the case

side-🚗 : fix the viewport in the list itself

## Tests

Added tests
Tested locally

## Risk

Low

## Deploy Plan

Deploy front

### DIFF
--- a/front/lib/resources/conversation_resource.ts
+++ b/front/lib/resources/conversation_resource.ts
@@ -72,6 +72,7 @@ import type {
   Attributes,
   CreationAttributes,
   InferAttributes,
+  Order,
   Transaction,
   WhereOptions,
 } from "sequelize";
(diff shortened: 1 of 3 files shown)


id13:
PMRR

id13:
Yes, it's one more query, I don't like it.
I have a data-model change in mind to no longer have to do all this juggling

front/lib/resources/conversation_resource.ts
```diff
@@ -1865,11 +1879,13 @@ export class ConversationResource extends BaseResource<ConversationModel> {
...
-    const lastValue = lastConversation
-      ? lastConversation.updatedAt.getTime().toString()
-      : null;
+    // Advance the cursor past the scanned window, unless accessible rows were
+    // cut by the limit — those must reappear on the next page.
+    const lastReturned = resultConversations[resultConversations.length - 1];
+    const lastValue =
+      conversations.length > pagination.limit && lastReturned
+        ? lastReturned.updatedAt.getTime().toString()
+        : rawRows[rawRows.length - 1].updatedAt.getTime().toString();
```
aubin-tchoi, line 1888: why not just `rawRows[rawRows.length - 1].updatedAt.getTime().toString();`?

aubin-tchoi (commented):
LGTM

PR #29440, opened 2026-07-23
[front] refactor: surface agent loop context asserts in tool handlers

aubin-tchoi:
## Description

We have some asserts on `AgentLoopRunContext` buried deep down in some utils, making it hard to understand which tools need it vs which ones do not.

We could technically do a massive refactor to TypeScript our way out of this (no assert, no nothing).
This PR starts simple by moving the asserts to the tool handlers, at their very beginning to make them easy to spot.

This PR is a pure refactor to make the next step easier.
Next step is to remove all of these asserts by handling them in the context of a pod function (file paths will refer to paths on the pod in this case).

## Tests

- Updated existing tests.

## Risk

- Low.

## Deploy Plan

- Deploy front.

### DIFF
--- a/front/lib/actions/mcp_internal_actions/utils/file_utils.ts
+++ b/front/lib/actions/mcp_internal_actions/utils/file_utils.ts
@@ -1,7 +1,4 @@
-import {
-  isAgentLoopRunContext,
-  type ToolContext,
-} from "@app/lib/actions/types";
+import type { AgentLoopRunContext } from "@app/lib/actions/types";
 import { resolveFile } from "@app/lib/api/actions/servers/files/tools/utils";
 import {
   conversationAttachmentId,
@@ -23,7 +20,6 @@ import { isAgentMessageType } from "@app/types/assistant/conversation";
 import { isContentFragmentType } from "@app/types/content_fragment";
 import type { Result } from "@dust-tt/client";
 import { Err, Ok } from "@dust-tt/client";
-import assert from "assert";
 import { PassThrough } from "stream";
 
 export function sanitizeFilename(filename: string): string {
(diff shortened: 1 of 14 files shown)


front/lib/actions/mcp_internal_actions/utils/file_utils.ts
```diff
@@ -53,10 +49,10 @@ export type ConversationFileRef = {
 export async function resolveConversationFileRef(
   auth: Authenticator,
   fileId: string,
-  toolContext: ToolContext | undefined
```
aubin-tchoi, line 56: the approach was to remove the assert, change the type here and see where the typecheck fails

front/lib/api/actions/servers/gmail/tools/index.ts
```diff
@@ -1009,15 +1023,18 @@ const handlers: ToolHandlers<typeof GMAIL_TOOLS_METADATA> = {
...
     }
-    const attachmentResult = await fetchAttachment(
-      auth,
-      attachmentFilePath,
-      runContext
-    );
-    if (attachmentResult.isErr()) {
-      return attachmentResult;
+    let attachment = null;
+    if (attachmentFilePath && attachmentRunContext) {
```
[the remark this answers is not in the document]
aubin-tchoi: not as explicit. we do `if (attachmentFilePath)` bc that's the logic, we do `if (attachmentRunContext)` to satisfy the typing

spolu (approved):

front/lib/api/actions/servers/gmail/tools/index.ts
```diff
@@ -397,15 +396,18 @@ const handlers: ToolHandlers<typeof GMAIL_TOOLS_METADATA> = {
...
     }
-    const attachmentResult = await fetchAttachment(
-      auth,
-      attachmentFilePath,
-      runContext
-    );
-    if (attachmentResult.isErr()) {
-      return attachmentResult;
+    let attachment = null;
+    if (attachmentFilePath && isAgentLoopRunContext(runContext)) {
```
spolu: Maybe we should error if a an attachmentFilePath is provided outside of an agentLoop?

PR #29363, opened 2026-07-23
[core] enh: Add qdrant_reshard tool

id13:
## Description

Add a rust cli to reshard qdrant shard key

I will add a runbook once I am more 🥋 on the matter.

#### Dry run

````bash
❯ QDRANT_CLUSTER_0_URL=http://localhost:14333 QDRANT_CLUSTER_0_API_KEY="" \
      cargo run --bin qdrant_reshard -- --collection c_openai_text-embedding-3-large-1536 \
        --target key_7=3 \
        --peer-url 195382974996013=http://localhost:14333 --peer-url 5005918239341447=http://localhost:14343
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.84s
warning: the following packages contain code that will be rejected by a future version of Rust: num-bigint-dig v0.8.4, redis v0.24.0
note: to see what the problems were, use the option `--future-incompat-report`, or run `cargo report future-incompatibilities --id 1`
     Running `target/debug/qdrant_reshard --collection c_openai_text-embedding-3-large-1536 --target key_7=3 --peer-url '195382974996013=http://localhost:14333' --peer-url '5005918239341447=http://localhost:14343'`
Collection c_openai_text-embedding-3-large-1536 on http://localhost:14333 (replication_factor 2):
  key_7: 2 -> 3 shards (up x1)
Peers:
  peer 5005918239341447 (0.5% RAM used, 0.1 / 16.8 GB)
  peer 195382974996013 (0.9% RAM used, 0.1 / 16.8 GB)
Execution order:
  1. key_7: up 2 -> 3 shards — candidate peer 5005918239341447 (0.5% RAM used, 0.1 / 16.8 GB)
Candidate peers are indicative: the target peer is re-selected from live telemetry before each step, excluding peers already used in this run.
Dry run only — re-run with --execute to launch.
````

#### Execute

OSS mode requires a driver to actually move the points, which add a lot of code and logs not useful for a cloud managed qdrant cluster.

````bash
❯ QDRANT_CLUSTER_0_URL=http://localhost:14333 QDRANT_CLUSTER_0_API_KEY="" \
      cargo run --bin qdrant_reshard -- --collection c_openai_text-embedding-3-large-1536 \
        --target key_7=4 --execute \
        --peer-url 195382974996013=http://localhost:14333 --peer-url 5005918239341447=http://localhost:14343 --drive
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.69s
[...]

PR #29441, opened 2026-07-24
[core] fix: qdrant_reshard watch never stops on transfer inactivity

id13:
Follow-up to #29363, from the first prod run (EU, `key_6`): the 10-minute no-transfer hard error fired on a managed cluster while the operation was legitimately parked waiting for the cloud operator.

Once an operation is in flight, the watch now never stops on its own — the operation belongs to whoever drives it, and only a human decides to interrupt (ctrl-c, `--abort`). Transfer inactivity is surfaced as data on every poll line instead:

```
[key_6] [0h23m00s elapsed] resharding: up shard 49 on peer … | shard 13 [Active@… 16668324 pts]; … | no transfer for 0h23m00s
```

The case the hard error was protecting against (OSS/self-hosted cluster where nothing will ever drive the operation) is already covered upfront: `--execute` without `--drive` on a non-managed cluster is refused before anything starts.

### DIFF
--- a/core/bin/qdrant/qdrant_reshard.rs
+++ b/core/bin/qdrant/qdrant_reshard.rs
@@ -645,17 +645,6 @@ async fn wait_for_no_resharding(
                 }
                 if !info.shard_transfers.is_empty() {
                     last_transfer_seen = std::time::Instant::now();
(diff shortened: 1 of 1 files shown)


id13:

- #29441 ◀

id13:
PMRR

PR #29443, opened 2026-07-24
feat(front): expand agent builder model picker inline on mobile

Nils-Fedrigo:
## Description
Not perfect but will hold the fort until we rework the picker!
Screenshot 2026-07-24 at 10 09 31

In the agent builder's **Advanced** dropdown, "Custom model" and "Custom reasoning effort" were rendered as flyout submenus (`DropdownMenuSub`), which are awkward and hard to use on mobile/touch.

On mobile (`useIsMobile()`, viewport < 768px) these now **expand inline within the same dropdown** instead of opening a flyout, mirroring the input-bar model picker's existing `isWidthConstrained` pattern:

- "Custom model" becomes an inline toggle; the selected/best-performing sections render inline, and each maker group ("From …") expands inline with its own state (plus a check mark on the maker holding the current selection).
- "Custom reasoning effort" expands its options inline.
- The shared rendering was refactored into `renderRadioItem` / `bestModelsSection` / `renderMakerModels` so desktop and mobile share one source of truth.

Also let the "Auto" model description wrap (`whitespace-normal`) instead of being clipped by the item's default `whitespace-nowrap`.

Desktop behavior is unchanged (flyout submenus preserved).

## Tests

Manually verified on desktop (submenus unchanged) and at a mobile-width viewport (inline expansion for model + effort, maker groups expand inline, Auto subtitle wraps). `tsgo --noEmit` and biome pass clean.

### DIFF
--- a/front/components/agent_builder/instructions/AdvancedSettings.tsx
+++ b/front/components/agent_builder/instructions/AdvancedSettings.tsx
@@ -141,7 +141,7 @@ export function AdvancedSettings() {
                     Auto
                     <Chip size="xs" color="highlight" label="Recommended" />
                   </span>
-                  <span className="text-xs font-normal text-muted-foreground">
+                  <span className="whitespace-normal text-xs font-normal text-muted-foreground">
                     Switches model depending on current availability for
                     balanced performance.
                   </span>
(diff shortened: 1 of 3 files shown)


Nils-Fedrigo:

- #29443 ◀

Fraggle (approved):
LGTM

PR #29444, opened 2026-07-24
Refactor: listAttachments agent generated files no longer need to iterate on the conversation.

Fraggle:
## Description

`listAttachments` collected agent-generated files by walking `conversation.content` and flatMapping `actions[].generatedFiles` off each `AgentMessage`. This breaks when the conversation is trimmed or paginated — missing messages means missing files.

Agent-generated files are now fetched independently via `AgentMCPActionResource.listGeneratedFilesForConversation(auth, { conversationId, upToRank })`, a direct DB query that doesn't depend on `conversation.content` being fully populated. The rank-filtering logic that was previously done by the local `filterConversationContentUpToRank` helper in `forks.ts` is now handled inside `listAttachments` via the new `upToRank` optional parameter.

## Tests

Added a `listGeneratedFilesForConversation` describe block in `agent_mcp_action_resource.test.ts` covering: empty conversation, files with correct agent creator metadata, `upToRank` boundary filtering. Tested locally.

## Risk

Low. Pure refactor — same observable behavior for fully-loaded conversations, correctness fix for trimmed/paginated ones. No signature changes to public APIs. Rollback is safe.

## Deploy Plan

Deploy `front`.


Fraggle:

- #29444 ◀
    - #29445

front/lib/resources/agent_mcp_action_resource.ts
```diff
@@ -716,6 +725,181 @@ export class AgentMCPActionResource extends BaseResource<AgentMCPActionModel> {
...
+
+      const agentConfiguration = agentConfigurationMap.get(
+        `${agentMessage.agentConfigurationId}:${agentMessage.agentConfigurationVersion}`
+      );
+
+      const creator: AttachmentCreator | null = agentConfiguration
+        ? {
+            type: "agent",
+            name: agentConfiguration.name,
+            pictureUrl: agentConfiguration.pictureUrl,
```
id13, line 873: So we do a join with AgentMessage just to show a the name and pictureUrl ? 🥲 

I know your are just respecting a contract here, but how hard would it be challenge this ?

id13 (approved):
LGTM, sad about the join on AgentMessage though

front/lib/resources/agent_mcp_action_resource.ts
Fraggle: We used to not have it :-)

PR #29445, opened 2026-07-24
Refactor: listAttachments now fetch contentfragments directlyy

Fraggle:
## Description

`listAttachments` previously iterated `conversation.content` to collect content fragments, which breaks for trimmed or paginated conversations (the same class of bug fixed for agent-generated files in PR29444). This PR replaces that walk with `fetchContentFragmentsForConversation`, a direct DB query joining `MessageModel` + `ContentFragmentModel` (main branch only, latest version per rank, optional `upToRank` boundary). Both content fragments and generated files are now fetched in parallel via `Promise.all`.

Callers in the sandbox and attachments routes are migrated from `getConversation` to `ConversationResource.fetchById`. `listAttachments` now accepts `ConversationWithoutContentType | ConversationResource` instead of `ConversationType`, and several other signatures (`carryOverConversationAttachments`, `resolveSkillMCPServers`, `getFileFromConversation`) are widened to `ConversationWithoutContentType`.

## Tests

Added Vitest integration tests for `fetchContentFragmentsForConversation` covering:
- ordered-by-rank, latest-only fetch
- `upToRank` boundary filtering
- exclusion of superseded content fragment versions

Tested locally.

## Risk

Low. Front-only, no DB schema changes. The behavioral change is replacing an in-memory walk of potentially incomplete `conversation.content` with a complete DB query — strictly more correct. Rollback is safe.

## Deploy Plan

Deploy front + front-api together.

### DIFF
--- a/front/lib/api/assistant/content_fragments.ts
+++ b/front/lib/api/assistant/content_fragments.ts
@@ -5,11 +5,13 @@ import type { ConversationResource } from "@app/lib/resources/conversation_resou
 import { ContentFragmentModel } from "@app/lib/resources/storage/models/content_fragment";
 import type {
   ConversationType,
+  ConversationWithoutContentType,
   UserMessageType,
 } from "@app/types/assistant/conversation";
 import type { ContentFragmentType } from "@app/types/content_fragment";
 import { isContentFragmentType } from "@app/types/content_fragment";
-import { Op, type Transaction } from "sequelize";
+import type { Transaction, WhereOptions } from "sequelize";
+import { Op } from "sequelize";
 
 export function getRelatedContentFragments(
   conversation: ConversationType,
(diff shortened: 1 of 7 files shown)


Fraggle:

- #29444
    - #29445 ◀

front/lib/api/assistant/content_fragments.ts
```diff
@@ -113,3 +115,61 @@ function collectConsecutivePrecedingContentFragments(
...
   return relatedContentFragments;
 }
+
+/**
+ * Fetch content fragments for a conversation without loading full conversation content.
+ * Returns the latest message version per rank, only fragments with
+ * `contentFragmentVersion === "latest"`, optionally limited to `rank <= upToRank`.
+ * Main branch only (`branchId` null).
+ */
+export async function fetchContentFragmentsForConversation(
```
id13, line 125: NIT: I think this method should be at the message resource level or conversation.
Why? Join between resource-owned models should be avoided, but if they can't, they should happen at the higher-order resource (in the sense of resolving belongs-to relationships), which will organically consolidate joins in a few locations vs spread in many, many resources. Easier to troubleshoot, cache optimize perf, refacto etc.

id13 (approved):
LGTM

Have you considered adding a conversationId to ContentFragment as a way to skip the join entirely on a huge table ?

The backfill is probably O(minutes)

It would require further work to also remove the need for messages in the batchRenderContentFragment, but looking at the method I think it's not that hard

front/lib/api/assistant/content_fragments.ts
Fraggle: I couldn't because it created a dep cycle, leading to a potential big refactoring.

FYI: On that topic, my personal opinion is we should not bundle ORM and business logic methods in Resources.

Fraggle:
> Have you considered adding a conversationId to ContentFragment as a way to skip the join entirely on a huge table ?

i think you did :)

Fraggle:
But batchRenderFromMessages operate on Messages anyway.

id13:
> i think you did :)

indeed ! Did it on so many tables I forgot 🤦

id13:
> But batchRenderFromMessages operate on Messages anyway.

Yes ... just to get files, so skippable imo

PR #29448, opened 2026-07-24
chore(sandbox): remove legacy GCS token broker

fontanierh:
## Description

Remove the one-release `/tmp/token.json` fallback, port 9876 firewall rules, and compatibility token server after the protected broker rollout is complete. Refresh then requires the root-owned image helpers and only requests recreation when those helpers are unavailable.

Draft only. Do not merge until #29447 and #29371 are deployed and the old-image kill sweep has completed.

## Tests

Updated refresh, image registry, and egress coverage to verify the compatibility broker and legacy port are gone.

## Risk

High if merged early: old-image pod sandboxes would lose the credential refresh needed for their pre-destroy flush. Keep this draft until no pre-cutover sandboxes remain.

## Deploy Plan

1. Complete the deploy and kill sweep from #29447 and #29371.
2. Confirm no pre-cutover sandbox images remain and recheck the proposed image versions.
3. Publish dust-base `0.8.59` and dsbx `0.1.38`.
4. Deploy this application change, then sweep dust-base `0.8.58` sandboxes.

### DIFF
--- a/cli/dust-sandbox/Cargo.lock
+++ b/cli/dust-sandbox/Cargo.lock
@@ -313,7 +313,7 @@ checksum = "92773504d58c093f6de2459af4af33faa518c13451eb8f2b5698ed3d36e7c813"
 
 [[package]]
 name = "dust-sandbox"
-version = "0.1.37"
+version = "0.1.38"
 dependencies = [
  "anyhow",
  "base64",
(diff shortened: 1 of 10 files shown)


flvndvd (commented):
LGTM

PR #29449, opened 2026-07-24
Add cached query and mutation hooks for Frame Pod Functions

flvndvd:
## Description

Frames often use the same Pod Function data across several slides or views.

Using `callFunction` imperatively requires each component to manage loading state, errors, request races, caching, and refreshes. It can also cause identical requests when several slides need the same data.

This PR adds two React hooks to the Frame wrapper:
- `usePodFunction` for read-only queries
- `usePodFunctionMutation` for explicit writes

Queries are cached by their fully qualified function path and input. Components within the same Frame can reuse cached results, including across slides. Identical in-flight requests are deduplicated, and previous data remains visible during revalidation.

Mutations remain explicit and do not automatically invalidate queries. After a write, callers revalidate the affected query with `mutate()`.

The existing imperative `callFunction` API remains unchanged.

This first PR excluisvely includes Frame-side plumbing and model guidance.

Publish-time validation of function references, inputs, and outputs will be done in a follow-up PR.

### DIFF
--- a/front/lib/resources/skill/code_defined/global/pod_functions.ts
+++ b/front/lib/resources/skill/code_defined/global/pod_functions.ts
@@ -170,29 +170,60 @@ return inline is written to a pod file whose path it reports), and \`${toolName(
 
 #### Calling a function from a Frame
 
-A Frame calls a published function through the injected \`@dust/react-hooks\` module, not the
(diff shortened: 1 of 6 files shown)


fontanierh (approved):
very cool

PR #29452, opened 2026-07-24
[core] fix: qdrant_reshard watch never stops on transfer inactivity (#29441)

id13:
Follow-up to #29441, from the prod runs (US `key_12`, EU `key_6`): with the new shard fully migrated and replicated (all replicas `Active`), the resharding operation stayed registered for 25–65+ minutes with no qdrant-level transfer — the cloud operator appears to hold the operation open while its own post-reshard replication/rebalancing is pending, presumably waiting on peer capacity. Since qdrant allows a single resharding per collection, this also blocks the remaining shard keys of the plan.

**One shard in, one shard out**: during the watch, every peer observed to receive a replica of the new shard is compensated exactly once — the bin moves that peer's largest `Active` shard *outside the resharded key* to the least-loaded peer below 50% RAM used that doesn't already host a replica of it (`move_shard`). This happens proactively as soon as each replica placement is observed (our chosen target peer on the first poll, the operator-chosen replica peer when it appears), so the operator never waits on room.

Details:
- Peers with nothing movable (fresh nodes hosting only the new shard) are skipped.
- No destination below 50% RAM → loud `WARNING`, and the watch continues (it never stops on its own; a human decides).
- Failures to issue a move are retried on the next poll; each peer is compensated at most once per resharding step.
- The moves show up as transfers on the poll lines, so progress stays visible.

### DIFF
--- a/core/bin/qdrant/qdrant_reshard.rs
+++ b/core/bin/qdrant/qdrant_reshard.rs
@@ -245,7 +245,15 @@ impl QdrantHttp {
     }
 
     async fn get<T: serde::de::DeserializeOwned>(&self, path: &str) -> Result<T> {
-        let req = self.with_api_key(self.client.get(format!("{}{}", self.base_url, path)));
+        self.get_at(&self.base_url.clone(), path).await
+    }
+
+    async fn get_at<T: serde::de::DeserializeOwned>(
+        &self,
+        base_url: &str,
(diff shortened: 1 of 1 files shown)


id13:

- #29452 ◀

id13:
PMRR

PR #29453, opened 2026-07-24
Admin governance: default skill availability value depending on feature flag

fabiencelier:
## Description

Change the default value to be "editors" when the flag is activated. Skill always start as unpublished and need explicit publication.

The flag is not really usable until we have the UI to change the states

## Risk

low, still behind Feature flag

## Deploy Plan
deploy front

### DIFF
--- a/front/types/assistant/skill_configuration.ts
+++ b/front/types/assistant/skill_configuration.ts
@@ -1,5 +1,6 @@
 import { MCPServerViewSchema } from "@app/lib/api/mcp_schemas";
 import type { AgentsUsageType } from "@app/types/data_source";
+import type { WhitelistableFeature } from "@app/types/shared/feature_flags";
 import { assertNever } from "@app/types/shared/utils/assert_never";
 import type { UserType } from "@app/types/user";
 import { z } from "zod";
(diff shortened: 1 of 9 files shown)


front/types/assistant/skill_configuration.ts
```diff
@@ -19,6 +20,16 @@ export type SkillAvailability = (typeof SKILL_AVAILABILITIES)[number];
 
 export const DEFAULT_SKILL_AVAILABILITY: SkillAvailability = "workspace_users";
```
ykmsd, line 21: nit but should we remove export so we don't accidentally use this value?

ykmsd (approved):

front-api/routes/w/[wId]/skills/availability.ts
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+    await SkillResource.updateAvailabilities(auth, skills, availability);
+
+    // Re-fetch: the bulk update does not refresh the in-memory resources.
+    const updatedSkills = await SkillResource.fetchByIds(auth, skillIds);
+
+    return ctx.json({
+      skills: updatedSkills.map((skill) => skill.toJSON(auth)),
+    });
```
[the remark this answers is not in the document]
Comment (fabiencelier) [reply front-api/routes/w/[wId]/skills/availability.ts]: unrelated PR

front/types/assistant/skill_configuration.ts
fabiencelier: it's still used at some places, for instance as the default value of the column in the model.

PR #1623, opened 2023-09-21
feat(assistant builder): add basic avatar picker

fontanierh:
Not super polished, but hopefully does the trick for now ?

Screenshot 2023-09-21 at 12 08 49

Issue #1596, opened 2023-09-20
Claude has a max_token issue


spolu:

spolu:
Example https://dust.tt/w/0ec9852c2f/assistant/107776d8ee

PR #29459, opened 2026-07-24
Update MCP sdk to latest 1.29

tdraier:
## Description

Update MCP from 1.27.0 to 1.29.0, which includes the fix for the `tsc` path: https://github.com/modelcontextprotocol/typescript-sdk/pull/1623 — this is the motivation for the upgrade.

1.29.0 is the latest 1.x (anything newer is the `2.0.0-beta` prerelease line, which we deliberately avoid). Bumped in all workspaces that depend on it (`front`, `sdks/js`, `cli/dust-cli`, `extension`) plus the root `overrides` pin, and regenerated `package-lock.json`.

### Why this is safe

Every change between 1.27.0 and 1.29.0 is a bug fix or v1.x backport — no intentional breaking API changes:

- **1.28.0**: use `scopes_supported` from resource metadata by default; default to `client_secret_basic` when the server omits `token_endpoint_auth_methods_supported`; reject plain JSON Schema objects passed as `inputSchema` (#1596); timeout/abort cleanup fixes; RFC 8252 loopback port relaxation.
- **1.29.0**: disallow null (infinite) requested TTL; add missing `size` field to `ResourceSchema`; add typings exports / missing `package.json` types (the `tsc` path fix above, #1623); allow advertising extensions in the capability object; always set `windowsHide` on Windows for stdio.

The only behavior-narrowing change, **#1596 (`tool()`/`registerTool()` now throw if passed a plain JSON Schema instead of a Zod shape)**, does not affect us: every `registerTool` call site passes a Zod `ZodRawShape` (`register.ts`, `mcp_internal_actions/wrappers.ts`, `extension/shared/tools/index.ts`). The plain `{ type: "object", ... }` objects in the codebase are our own config types / mock wire-format `Tool` objects, never arguments to `registerTool`.

The auth-default changes are also no-ops for us: `remote_mcp_servers_resource.ts` already branches explicitly on `client_secret_post`/`client_secret_basic` and already prioritizes resource-metadata `scopes_supported`. No `StdioClientTransport` usage, so the Windows stdio change is irrelevant.

## Tests

`npx tsgo --noEmit` in `front` (type-check). No source changes required — dependency bump only.

## Risk

Low. Patch/minor bug-fix bump with no breaking API changes affecting our usage. Can be rolled back by reverting the version bump.

## Deploy Plan

deploy front

### DIFF
--- a/cli/dust-cli/package.json
+++ b/cli/dust-cli/package.json
@@ -28,7 +28,7 @@
   },
   "dependencies": {
     "@dust-tt/client": "^1.2.6",
-    "@modelcontextprotocol/sdk": "^1.27.0",
+    "@modelcontextprotocol/sdk": "^1.29.0",
     "chalk": "^5.4.1",
     "clipboardy": "^4.0.0",
     "diff": "^8.0.2",
--- a/extension/package.json
+++ b/extension/package.json
@@ -57,7 +57,7 @@
     "@mixmark-io/domino": "^2.2.0",
     "@dust-tt/sparkle": "^0.7.7",
     "@frontapp/plugin-sdk": "^1.8.1",
-    "@modelcontextprotocol/sdk": "^1.27.0",
+    "@modelcontextprotocol/sdk": "^1.29.0",
     "@tailwindcss/forms": "^0.5.9",
     "@tiptap/core": "^3.19.0",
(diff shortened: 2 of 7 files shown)


tdraier:

- #29459 ◀

Fraggle (approved):
Looks reasonnable.

PR #29461, opened 2026-07-24
[front] Gate legacy Dust Apps behind legacy_dust_apps flag

tdraier:
## Description

Dust apps were only **partially** gated by the `legacy_dust_apps` feature flag. Even with the flag off, a user could:

- See the **"Create a Dust App"** item under "Add data" (only builder/write permissions were checked, not the flag).
- Directly navigate to `/w/:wId/spaces/:spaceId/categories/apps` and the app editor route tree.
- Call the **private app API endpoints** (create/list/datasets/runs/state) — these had **no flag check at all**, so app creation worked even bypassing the UI.

### Changes

**Frontend UI**
- `SpaceCategoriesList.tsx`: hide the "Create a Dust App" dropdown item when `hasFeature("legacy_dust_apps")` is false (consistent with how the Apps category tile is already filtered).

**Frontend routes (`front-spa`)**
- New `RequireFeatureFlagLayout` route guard, mirroring the existing `RequireRoleLayout` (renders `Custom404` when the flag is missing, else `<Outlet/>`).
- Wrapped the `categories/apps` page and the entire `spaces/:spaceId/apps/:aId` editor route tree with it.

**Backend (`front-api`)**
- `app.use("*", withFeatureFlag("legacy_dust_apps"))` at the top of the private apps subrouter — gates **list, create, and every per-app sub-route** (datasets, runs, state). Same pattern as `sandbox-functions`; `auth` is already set upstream by the global `workspaceAuth()`.

**Tests**
- Enable the flag via `FeatureFlagFactory` in the affected endpoint test (it would otherwise 403).

### Scope / not included

Deliberately **did not** gate the public API (`/api/v1/.../apps/*`, incl. `import` and `runs`) or the poke admin endpoints:
- Gating the public v1 surface would be a breaking change for external consumers currently running Dust apps via API.
- Poke is internal support tooling.

## Tests

locally

## Risk

low

## Deploy Plan

front

### DIFF
--- a/front-api/routes/w/[wId]/spaces/[spaceId]/apps/[aId]/index.test.ts
+++ b/front-api/routes/w/[wId]/spaces/[spaceId]/apps/[aId]/index.test.ts
@@ -4,6 +4,7 @@ import { MCPServerViewModel } from "@app/lib/models/agent/actions/mcp_server_vie
 import { AgentConfigurationModel } from "@app/lib/models/agent/agent";
 import { AppResource } from "@app/lib/resources/app_resource";
 import { generateRandomModelSId } from "@app/lib/resources/string_ids_server";
+import { FeatureFlagFactory } from "@app/tests/utils/FeatureFlagFactory";
 import { createPrivateApiMockRequest } from "@app/tests/utils/generic_private_api_tests";
 import { honoApp } from "@front-api/app";
 import { describe, expect, it } from "vitest";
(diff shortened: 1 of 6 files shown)


ykmsd (approved):
OOC does anybody use Dust apps since we decided to deprecate?

tdraier:
> OOC does anybody use Dust apps since we decided to deprecate?

Actually some workspaces did create apps despite not having the feature flag.  Still not very easy to access them if you don't have the FF , so quite safe to gate - they will ask for the feature flag if they really need it

PR #29463, opened 2026-07-24
29446 fixfront adding OpenAI explicit version to front

### LINKED ISSUE #29446: fix(front): Adding openai explicit version to front
**Summary:** `front-api` has no explicit 'openai' dependency and relies on Node resolving it via the hoisted root fallback. Locally, a full npm install across all workspaces creates a version conflict: 'connectors' depends on '@microsoft/teams-ai', which peer-pins 'openai@4.77.4', while front requires 'openai@^6.46.0' directly. npm resolves this by hoisting the old '4.77.4' to the repo root and nesting the correct '6.46.0' under front/node_modules.
 
Since Node's module resolution checks the root 'node_modules' before ever falling back to NODE_PATH, front-api silently picks up the stale 4.77.4, which lacks the OpenAI Responses API.
 
**Implications:** Find similar skills feature is broken for monorepo install, and this will have the same incidence for the in-development find similar agent feature. 
 
**Why this doesn't happen in prod:** `connectors` and `front/front-api` are not in the same containers in productions.
Suggested fix: Add "openai": "^6.46.0" as an explicit dependency in front-api/package.json (matching front).

avervaet:
## Description

Until now, `front-api` imported the OpenAI SDK without declaring it as a direct dependency. In local monolith development, dependency resolution could therefore rely on `node_modules` hoisting and select a stale or incompatible version, breaking front-api routes used by the similar skill and similar agent discovery features. See [#29446](https://github.com/dust-tt/dust/issues/29446).

This PR declares `openai` explicitly in `front-api/package.json` at `^6.46.0` and updates `package-lock.json` accordingly, resolving it to `6.48.0`. The existing batch and streaming client behavior is unchanged; this change makes their dependency boundary explicit and deterministic.

## Tests

Manual local monorepo testing was performed.

## Risk

The main risk is compatibility with the explicitly resolved OpenAI SDK version. The lockfile now records the front-api-local package resolution, while no request logic, data model, migration, feature flag, or permissions behavior changes. Rollback requires reverting the direct dependency and lockfile entries.

## Deploy Plan

Deploy front


Fraggle (approved):
LGTM

front/lib/api/llm/clients/openai/index.ts
```diff
@@ -42,6 +42,9 @@ import { Err, Ok } from "@app/types/shared/result";
 import { assertNever } from "@app/types/shared/utils/assert_never";
 import { normalizeError } from "@app/types/shared/utils/error_utils";
 import assert from "assert";
+// Do not remove: front-api routes call into this client for the similar skill
```
aubin-tchoi: do not remove what?

package-lock.json
```diff
@@ -752,6 +753,36 @@
         "url": "https://github.com/sponsors/sindresorhus"
       }
     },
+    "front-api/node_modules/openai": {
+      "version": "6.48.0",
```
aubin-tchoi, line 757: same version as in prod?

aubin-tchoi (approved):
LGTM

PR #29326, opened 2026-07-22
[GitHub MCP] Handle inaccessible team reviewers

philipperolet:
## Description

Fixes https://github.com/dust-tt/tasks/issues/9593

`list_pull_requests` and `search_advanced` request the slug of team reviewers. GitHub App installation tokens without organization Members read access return node-level `FORBIDDEN` errors for that field, and Octokit turns those partial GraphQL responses into failures for the entire result page.

Retry only these reviewer-specific authorization failures without the Team fragment. The tools still return the pull requests and user review requests, while connections allowed to read teams keep returning team slugs.

## Risks

Blast radius: GitHub MCP pull request listing and advanced search
Risk: low

## Deploy Plan

- deploy front

## Tests

- [x] targeted tests for `list_pull_requests` and `search_advanced` reviewer fallback

PR #29467, opened 2026-07-24
[GitHub MCP] Skip inaccessible review requests on fallback

philipperolet:
## Description

Fixes https://github.com/dust-tt/tasks/issues/9593
Follow-up to https://github.com/dust-tt/dust/pull/29326

The first fallback removed only the Team fragment, but still asked GitHub to resolve `requestedReviewer`. Resolving that union to a Team can require organization Members access, so the retry could fail with the same authorization error.

Add an `includeReviewRequests` GraphQL directive and retry reviewer-specific authorization failures with the entire `reviewRequests` field disabled. Pull request data still returns successfully; only review requests are empty on affected connections. Connections with team access keep the existing full response.

## Risks

Low. The fallback applies only when every GraphQL error is a `FORBIDDEN` error below `reviewRequests`. Other errors still propagate unchanged.

## Deploy plan

Deploy Front normally.

## Tests

- [x] `NODE_ENV=test npm test -- lib/api/actions/servers/github/tools/index.test.ts`
- [x] `npx biome ci lib/api/actions/servers/github/tools/index.ts lib/api/actions/servers/github/tools/index.test.ts`
- [x] `NODE_OPTIONS=--max-old-space-size=8192 npx tsgo --noEmit`

### DIFF
--- a/front/lib/api/actions/servers/github/tools/index.test.ts
+++ b/front/lib/api/actions/servers/github/tools/index.test.ts
@@ -74,10 +74,11 @@ describe("GitHub pull request tools", () => {
   it.each([
     ["search_advanced", { query: "is:pr" }, searchResponse],
     ["list_pull_requests", { owner: "dust-tt", repo: "dust" }, listResponse],
-  ])("retries %s without team details when GitHub denies them", async (toolName, params, response) => {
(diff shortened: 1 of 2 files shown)


aubin-tchoi (commented):
LGTM

PR #29471, opened 2026-07-24
Admin governance: add endpoint to bulk edit skill availability

fabiencelier:
## Description

The endpoint takes a list of skill ids and an availability.
Add an resource method to be able to create the old versions by bulk, which is required to efficiently change the availability by bulk.
Note that this do create a new version of each skill and set the "editedBy" field even if the caller is not actually an editor (but an admin changing availability).

This endpoint will be used to bulk edit the available skills page and change several availability at once.

## Tests

unit tests added

## Risk

low, easy to revert 

## Deploy Plan

deploy front

### DIFF
--- a/front/lib/resources/skill/skill_resource.ts
+++ b/front/lib/resources/skill/skill_resource.ts
@@ -2946,19 +2946,44 @@ export class SkillResource extends BaseResource<SkillConfigurationModel> {
    * permission on skills — being an editor is neither required nor sufficient. Does not
    * touch editedBy.
    */
-  async updateAvailability(
+  static async updateAvailabilities(
     auth: Authenticator,
+    skills: SkillResource[],
     availability: SkillAvailability
   ): Promise<void> {
     assert(
       await auth.hasWorkspacePermission("publish", "skill"),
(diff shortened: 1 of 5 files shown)


front/lib/resources/skill/skill_resource.ts
```diff
@@ -4074,70 +4099,97 @@ export class SkillResource extends BaseResource<SkillConfigurationModel> {
...
+      where: versionWhere,
       transaction,
     });
+    const maxVersionBySkillId = new Map<ModelId, number>();
+    for (const row of versionRows) {
+      const currentMax = maxVersionBySkillId.get(row.skillConfigurationId) ?? 0;
+      if (row.version > currentMax) {
+        maxVersionBySkillId.set(row.skillConfigurationId, row.version);
+      }
+    }
```
fabiencelier, line 4162: Implementation is slightly different: we get the max + 1 instead of count + 1
It will return the same value as long as we always do +1 increments (which we do) + it is slightly more robust if we ever delete an old version or something

matteotrab (approved):

PR #29473, opened 2026-07-24
feat: warn users about similar agents before creating duplicates

avervaet:
## Description

This PR adds a feature-flagged check in the agent builder that warns users when the agent they are creating appears to overlap with an existing agent, helping them reuse an existing agent instead of creating a near-duplicate.

When creating a new agent, the UI watches the description and, once it reaches 10 characters, sends a debounced and abortable request to a new workspace API endpoint. The endpoint compares the new description with the full instructions of visible, documented agents using a small whitelisted LLM model. Agents are processed in batches of 100 with a maximum of four concurrent LLM calls, and duplicate results are removed before being returned.

The UI displays matching agents with their name, description, avatar, and a link to open the agent builder. The check is gated by the `similar_agents_check` feature flag and does not run while editing an existing agent or when the flag is disabled. LLM failures are shown as a non-blocking inline error and do not prevent agent creation. The implementation also adds the corresponding API types, SDK feature-flag union entry, and LLM trace operation type.

Similar agents warning

### Covered

Functional endpoint tests in `front-api/routes/w/[wId]/assistant/agent_configurations/similar.test.ts` cover:

- Returning LLM-selected similar agent IDs.
- Empty LLM results.
- Workspaces with no custom agents.
- Batching across multiple LLM calls and deduplicating overlapping IDs.
- A `400 invalid_request_error` response when `naturalDescription` is missing.

Component tests in `front/components/agent_builder/AgentBuilderSimilarAgentsSection.test.tsx` cover the minimum description length, debouncing, displaying results, editing an existing agent, and the feature-flag-disabled path. A manual end-to-end pass through the agent-builder UI was also performed.

The actual LLM prompt quality and semantic accuracy are not covered because the LLM boundary is mocked in the automated tests.

## Risk

This adds LLM calls and corresponding latency/cost when the feature flag is enabled. The impact is bounded by the 10-character UI threshold, request debouncing and cancellation, batches of 100 agents, and a maximum of four concurrent calls. Results are advisory only: failures render an inline message and do not block agent creation.

The feature is disabled unless `similar_agents_check` is enabled for the workspace. No data model, migration, or backfill changes are included, and existing-agent editing behavior remains unchanged. Rollback is a code revert or disabling the feature flag.

## Deploy Plan

Deploy the application changes with `similar_agents_check` disabled by default. Enable the `dust_only` flag for selected workspaces when ready, allowing a gradual rollout. No migration or backfill is required.


avervaet:
First commit places the component below Instructions. As the search is triggered by the descritpion field I moved it for UX reasons. It is clearer for the user and visible before clicking on save.

avervaet:
> Warnings
> ⚠️	
> Files in `front-api/routes/` have been modified and the PR has the `documentation-ack` label. Don't forget to run `npm -w front-api run docs` and use the `Deploy OpenAPI Docs` Github action to update https://docs.dust.tt/reference.
> 
> Generated by 🚫 [dangerJS](https://danger.systems/js) against [dcc398d](https://github.com/dust-tt/dust/commit/dcc398d999f81e82b386be27aec656b59b2ff09b)

The link in this message triggers a 404 is this intended ?

fabiencelier (approved):
Tested locally and it worked well 

Image

PR looks good to me !
Thanks a lot for tackling this :D

fabiencelier:
> The link in this message triggers a 404 is this intended ?

Broken indeed
@iliasbet can this be related to your recent changes to the doc ?

@avervaet nothing to do for you here as it is not public endpoint

front/components/agent_builder/AgentBuilderSimilarAgentsSection.tsx
```diff
@@ -0,0 +1,206 @@
...
+
+  if (similarAgents.length === 0 && !isLoading) {
+    return null;
+  }
+
+  if (similarAgents.length === 0 && isLoading) {
+    return (
+      <div className="flex items-center gap-2 text-sm text-muted-foreground">
+        <Spinner size="xs" />
+        <span>Checking for similar agents...</span>
```
aubin-tchoi: nit: ellipsis character

front/components/agent_builder/AgentBuilderSimilarAgentsSection.tsx
```diff
@@ -0,0 +1,206 @@
...
+  | { type: "fetch_settled" }
+  | { type: "fetch_error" };
+
+const initialState: State = {
+  similarAgents: [],
+  isLoading: false,
+  hasError: false,
+};
+
+function reducer(state: State, action: Action): State {
```
aubin-tchoi, line 39: do you really need a reducer?

front/lib/api/assistant/existing_agent_checker.ts
```diff
@@ -0,0 +1,237 @@
...
+  // Retrieve agents visible to the user in the workspace (same set as the
+  // "manage agents" list), excluding built-in Dust agents (global scope) and
+  // agents without a description to compare against.
+  const allAgents = await getAgentConfigurationsForView({
+    auth,
+    agentsGetView: "list",
+    variant: "light",
+  });
+  const agents = allAgents.filter(
+    (a) => a.scope !== "global" && a.description.trim().length > 0
```
aubin-tchoi: as discussed IRL we could indeed keep them to address a potential discoverability issue on our global agents

front/lib/api/assistant/existing_agent_checker.ts
```diff
@@ -0,0 +1,237 @@
...
+import type { Result } from "@app/types/shared/result";
+import { Err, Ok } from "@app/types/shared/result";
+import chunk from "lodash/chunk";
+import uniq from "lodash/uniq";
+
+// Safeguards to avoid sending a huge number of agents to a single LLM call:
+// agents are checked in batches of this size, one LLM call per batch.
+export const AGENTS_PER_LLM_CALL = 100;
+const MAX_CONCURRENT_LLM_CALLS = 4;
+const MAX_DESCRIPTION_LENGTH = 500;
```
aubin-tchoi: would make that much bigger

front/lib/api/assistant/existing_agent_checker.ts
```diff
@@ -0,0 +1,237 @@
...
+): Promise<Result<string[], Error>> {
+  const owner = auth.getNonNullableWorkspace();
+
+  const existingAgents = agents
+    .map(
+      (a) => `Agent ID ${a.sId}:
+"${truncateDescription(a.description)}"`
+    )
+    .join("\n---\n");
+  const inputText = `Input description:"${naturalDescription}"
```
aubin-tchoi, line 129: as per IRL TBD how well the description describes agents with high spread. I have some concerns on the precision of the results (less on the recall), but can easily be monitored via a log such as the one [here](https://app.datadoghq.eu/dashboard/aj2-eg7-qjh/skills-monitoring?fromUser=false&fullscreen_end_ts=1784904895141&fullscreen_paused=false&fullscreen_refresh_mode=sliding&fullscreen_section=overview&fullscreen_start_ts=1784300095141&fullscreen_widget=1024115171393451&refresh_mode=sliding&from_ts=1784300091833&to_ts=1784904891833&live=true)

aubin-tchoi (approved):
LGTM 🔥

front/lib/api/assistant/existing_agent_checker.ts
avervaet: Agree, let's beta test-it as if and expand based on price + precision monitoring. 
But it could be a nice incentive to push users toward Dust optimized agents.

front/lib/api/assistant/existing_agent_checker.ts
avervaet: Right now descriptions tend to be really synthetic but if we want to use instructions instead at some point it could be indeed great to increase it.

avervaet:
Locked the feature behind a flag: `similar_agents_check`. 
Switch to use the whole agent instructions and not just the description for similarity check. 
nest steps: 
- deployement in prod and allow for Dust workspace 
- monitor credit usage in Datadog

PR #29469, opened 2026-07-24
[front] switch dust app administration to dust_app:admin capability

tdraier:
## Description

Part 2 of moving Dust app administration off the deprecated builder role (dust-tt/tasks#9746). **Enforces the `dust_app:admin` capability** added in #29460, replacing `auth.isBuilder()` / `useAuth().isBuilder` for Dust apps — server and client.

Stacked on #29460 (the capability + seeder). Merge that first.

### Server
- App creation (`front-api .../spaces/[spaceId]/apps`) now checks `auth.hasWorkspacePermission("admin", "dust_app")` instead of `auth.isBuilder()`.

### Client (`apps/*` pages)
Swapped `useAuth().isBuilder` → `useWorkspacePermissions().hasPermission("admin", "dust_app")`. The permission is served synchronously from the auth context (same as `isBuilder` was), so it's a 1:1 swap with no loading changes:
- `SpaceAppsList` / `SpaceAppsListPage` — the "New App" button (prop renamed `isBuilder` → `canAdministrateApps`)
- `AppViewPage`, `RunsPage`, `RunPage`, `DatasetPage`, `DatasetsPage` — read-only gating
- `AppSettingsPage`, `NewDatasetPage` — redirects for users without the permission

### Behavior

For workspaces with the `legacy_dust_apps` flag, #29460's seeder grants `dust_app:admin` to the Builders group, so builders/admins keep app access; other members lose the (previously builder-gated) app administration surface — as intended by the builder-role deprecation.

### Testing

Verified zero TypeScript errors in the touched files (anchored). Full local `tsgo`/tests blocked by unrelated `node_modules` drift in the dev workspace; CI's clean install runs the full suite.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #29460, opened 2026-07-24
[front] add dust_app:admin governance capability and seeder

tdraier:
## Description

Part 1 of moving Dust app administration off the deprecated builder role (dust-tt/tasks#9746). This PR **adds the `dust_app:admin` governance capability and prepares its backfill** — no enforcement yet. The follow-up PR #29469 switches the actual checks.

### Changes

- **New `dust_app:admin` capability**
  - `front/types/group_permissions.ts` — add `"dust_app"` to `GROUP_PERMISSION_RESOURCE_TYPES` and to `EMPTY_WORKSPACE_PERMISSIONS`.
  - `front/lib/resources/group_permission_registry.ts` — add `dust_app: { admin: { verbs: ["admin"], levels: ["type"] } }` (type-level singleton, matching `billing`/`identity`) and the `verbsByResource` entry. Registry invariant tests pass by construction (name == verb; `admin` already in `GRANT_TYPES`/`GRANT_VERBS`).
- **Seeder** (`governance_seeding.ts` → `CAPABILITY_SEEDERS`): targets the Builders group for workspaces with the `legacy_dust_apps` flag (degrading to admins_only when the workspace has no Builders group), admins_only otherwise. The existing backfill migration and new-workspace seeding pick it up automatically.

### No behavior change

Nothing reads `dust_app:admin` in this PR. Enforcement is in the stacked follow-up **#29469**.

### Testing

Verified zero TypeScript errors in the touched files (anchored). Full local `tsgo` is currently blocked by an unrelated `node_modules` drift in the dev workspace; CI's clean install runs the full typecheck + registry tests.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #29474, opened 2026-07-24
[front] remove isBuilder from the auth context

tdraier:
## Description

Now that all UI Dust-app gating uses the dust_app:admin capability (and other
builder-role UI was removed earlier), nothing in the UI reads
useAuth().isBuilder anymore, so drop the field entirely.

- Remove isBuilder from AuthContextValue (front) and the auth-context response
  contract types (GetWorkspaceAuthContextResponseType and the poke
  GetPokeWorkspaceAuthContextResponseType).
- Stop populating it in the two front-api auth-context routes.
- Drop the now-unused field from the front-spa (ConversationRouterLayout) and
  extension (AuthProvider) context builders.

The server-side Authenticator.isBuilder() method is untouched; it is still used
by the remaining server-side checks being migrated separately.

## Risk

low, isBuilder tests already removed. Logged in builders will have to reload to rely on the new check ("builder" actions will disappear).

## Deploy Plan

deploy front


tdraier:

- #29460
    - #29469
        - #29474 ◀

matteotrab (approved):

PR #29475, opened 2026-07-24
Admin governance: add parameter to bypass the editor visibility

fabiencelier:
## Description

Related to https://github.com/dust-tt/tasks/issues/9728

Change the private endpoint and V1 endpoint to support a new param `bypassEditorVisibility` which allows to list all skills, including the unpublished ones where the user is not editor.
This is only available to admin users.

## Tests

Added unit tests for both endpoints

## Risk

low

## Deploy Plan

deploy front

### DIFF
--- a/front-api/public/swagger.json
+++ b/front-api/public/swagger.json
@@ -3437,7 +3437,7 @@
             "in": "query",
             "name": "availability",
             "required": false,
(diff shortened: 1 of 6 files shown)


ykmsd (approved):
ty!

PR #29482, opened 2026-07-24
Optimize: do not fetch whole conversation for notifications.

Fraggle:
## Description

The notification workflows (`activation-new-conversation`, `conversation-unread`, `getConversationDetails`) previously called `getLightConversation`, which loads the full serialized conversation content into memory. For long conversations this is wasteful since notifications only need a handful of targeted data points.

Replaced the full-load path with targeted DB queries extracted into `front/lib/notifications/conversation_fetch.ts`:
- `fetchUnreadMessageIds` — returns IDs of messages created or agent-completed after `lastReadAt`, via two parallel `MessageModel.findAll` queries instead of walking `conversation.content`
- `fetchMessagesByModelIds` — loads only the needed message rows with side-table includes (`UserMessageModel`, `AgentMessageModel`, etc.)
- `fetchEarliestLatestVersionMessageIds` — group-by-rank query to find the first visible message without fetching all ranks
- `getUnreadNotificationFlags` / `conversationUsesAgentsWithRetention` / `hasUnreadSucceededAgentReply` — targeted queries replacing inline `conversation.content` iteration in `helpers.ts` and `activation-new-conversation.ts`
- `fetchUnreadLightConversation` — builds a `LightConversationType` whose `content` is already pre-filtered to unread messages, so `conversation-unread.ts` no longer needs to filter after the fact

## Tests

Added Vitest integration tests:
- `helpers.test.ts` — 8 cases covering `getConversationDetails`: user/agent messages, mention flags, email/slack origins, read/unread states, retention policies, project conversations, deleted conversations, empty `subscriberId`
- `activation-new-conversation.test.ts` — 6 cases covering `shouldSkipActivationNewConversation`: missing subscriber, missing conversation, no agent reply, unread succeeded reply, already-read reply, agent completing after `lastReadAt`

## Risk

Low. The new queries reproduce the same filtering logic previously done in memory on `conversation.content`. Notification correctness depends on `fetchUnreadMessageIds` matching the old `isMessageUnread` semantics — covered by the new tests. No schema changes. Rollback is safe.

## Deploy Plan

Deploy front.

### DIFF
--- a/front/lib/notifications/conversation_fetch.ts
+++ b/front/lib/notifications/conversation_fetch.ts
@@ -1,0 +1,311 @@
+import { batchRenderMessages } from "@app/lib/api/assistant/messages";
+import type { Authenticator } from "@app/lib/auth";
+import {
+  AgentMessageModel,
+  MentionModel,
+  MessageModel,
+} from "@app/lib/models/agent/conversation";
+import { ConversationResource } from "@app/lib/resources/conversation_resource";
+import type {
+  ConversationWithoutContentType,
+  LightConversationType,
+  LightMessageType,
+  UserMessageOrigin,
+} from "@app/types/assistant/conversation";
+import {
+  ConversationError,
+  isVisibleMessage,
+} from "@app/types/assistant/conversation";
+import type { Result } from "@app/types/shared/result";
+import { Err, Ok } from "@app/types/shared/result";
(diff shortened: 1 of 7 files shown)


Fraggle:

- #29482 ◀

xebeohp (commented):
LGTM, ty!

PR #29484, opened 2026-07-24
Do not fetch full conversation for getting the plan.

Fraggle:
## Description

The `GET` and `DELETE` handlers in `plan_mode.ts` were calling `getLightConversation`, which hydrates conversation content (messages), even though `getActivePlanContent` and `closeActivePlan` only need the conversation record. Replaced both with `ConversationResource.fetchById`, which fetches only conversation metadata. Null-checks replace the previous `Result`-based error pattern to match `fetchById`'s return signature.

## Tests

Tested manually — plan mode GET and DELETE routes behave correctly with the lightweight fetch.

## Risk

Low. `ConversationResource.fetchById` is already the established pattern for metadata-only lookups across the codebase. No behavioral change — `toJSON()` produces the same `ConversationType` value passed downstream.

## Deploy Plan

Deploy front-api.

### DIFF
--- a/front-api/routes/w/[wId]/assistant/conversations/[cId]/plan_mode.ts
+++ b/front-api/routes/w/[wId]/assistant/conversations/[cId]/plan_mode.ts
@@ -1,12 +1,13 @@
-import { getLightConversation } from "@app/lib/api/assistant/conversation/fetch";
 import {
   closeActivePlan,
   getActivePlanContent,
 } from "@app/lib/api/assistant/plan_mode";
+import { ConversationResource } from "@app/lib/resources/conversation_resource";
 import type {
   DeleteConversationPlanModeResponseBody,
   GetConversationPlanModeResponseBody,
 } from "@app/types/api/assistant/plan_mode";
+import { ConversationError } from "@app/types/assistant/conversation";
 import { apiErrorForConversation } from "@front-api/lib/api/assistant/conversation/helper";
 import { workspaceApp } from "@front-api/middlewares/ctx";
 import type { HandlerResult } from "@front-api/middlewares/utils";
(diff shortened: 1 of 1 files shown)


Fraggle:

- #29484 ◀

front-api/routes/w/[wId]/assistant/conversations/[cId]/plan_mode.ts
```diff
@@ -29,12 +30,20 @@ app.get(
...
-      return apiErrorForConversation(ctx, conversationRes.error);
+    const conversationResource = await ConversationResource.fetchById(
+      auth,
+      cId
+    );
+    if (!conversationResource) {
+      return apiErrorForConversation(
+        ctx,
+        new ConversationError("conversation_not_found")
+      );
```
Comment (PopDaph) [inline front-api/routes/w/[wId]/assistant/conversations/[cId]/plan_mode.ts:41]: FMI, why not this so we do not manually build the error?  🙏🏻 

```suggestion
    const conversationRes =
      await ConversationResource.fetchConversationWithoutContent(auth, cId);
    if (conversationRes.isErr()) {
      return apiErrorForConversation(ctx, conversationRes.error);
    }

    const contentRes = await getActivePlanContent(auth, conversationRes.value);
```

Comment (Fraggle) [reply front-api/routes/w/[wId]/assistant/conversations/[cId]/plan_mode.ts:41]: fetchConversationWithoutContent is slightly more expensive at it will fetch read status and a bit more.

`ConversationResource.fetchById()` is the cheapest option.

Comment (Fraggle) [reply front-api/routes/w/[wId]/assistant/conversations/[cId]/plan_mode.ts:41]: (but we could / should align ConversationResource.fetchById() to return a Result)

Comment (PopDaph) [reply front-api/routes/w/[wId]/assistant/conversations/[cId]/plan_mode.ts:41]: It starts with 

```
const conversation = await this.fetchById(auth, sId, {
      includeDeleted: options?.includeDeleted,
      dangerouslySkipPermissionFiltering:
        options?.dangerouslySkipPermissionFiltering,
      includeForkingData: options?.includeForkingData,
    });

    if (!conversation) {
      return new Err(new ConversationError("conversation_not_found"));
    }
```

Comment (Fraggle) [reply front-api/routes/w/[wId]/assistant/conversations/[cId]/plan_mode.ts:41]: Yep, but it add this call ConversationResource.getActionRequiredAndLastReadAtForUser no matter what

Comment (Fraggle) [reply front-api/routes/w/[wId]/assistant/conversations/[cId]/plan_mode.ts:41]: (2 queries)

Comment (PopDaph) [reply front-api/routes/w/[wId]/assistant/conversations/[cId]/plan_mode.ts:41]: In that case I'd do the follow-up you mention: align `ConversationResource.fetchById()` to return a `Result<_, ConversationError>`. 
That gives us the cheapest fetch *and* the clean `apiErrorForConversation(ctx, res.error)` shape without hand-building the error. 

Then, the sibling routes that go through `fetchConversationWithoutContent` today without needing read state could migrate and drop those 2 queries too. 🪄

Comment (Fraggle) [reply front-api/routes/w/[wId]/assistant/conversations/[cId]/plan_mode.ts:41]: I have to clean up a few other stuff first but then yes totally !

PR #29486, opened 2026-07-24
[front] fix: pass run context in file utility tests

aubin-tchoi:
## Description

typecheck fails on main after #29440 because the file utility tests still pass the full `ToolHandlerExtra` to helpers that now accept `AgentLoopRunContext`.

This narrows the `makeExtra` test factory return type to preserve its agent-loop context and passes `runContext` directly in the affected tests. Runtime behavior is unchanged.

## Tests

- Updated existing tests.

## Risk

- Low.

## Deploy Plan

- No deploy.


Fraggle (approved):
LGTM

PR #29487, opened 2026-07-24
Add: lint rule to make sure we know what we do when we are fetching a full conversation.

Fraggle:
## Description

Adds a Biome/GritQL lint rule (`noExpensiveConversationFetch`) that flags bare calls to `getConversation()` and `getLightConversation()` with `severity: error`. Both functions hydrate the full conversation content (messages, actions, etc.) and are expensive — callers that only need the conversation record should use `ConversationResource.fetchById` or `fetchConversationWithoutContent` instead. The rule is scoped to `front/**` and `front-api/**`, excluding `fetch.ts` (the canonical implementation) and test files. All existing call sites that legitimately need a full load are grandfathered with a `// biome-ignore lint/plugin/noExpensiveConversationFetch: intentional full conversation load` suppression, making future accidental additions visible immediately in CI.

## Tests

Rule correctness validated via the GritQL test cases in `noExpensiveConversationFetch.md` (flags `getConversation()`/`getLightConversation()` standalone calls, ignores `ConversationResource.*` helpers and member-expression calls like `api.getConversation()`).

## Risk

Low. No runtime behaviour changes — lint config only. Existing call sites are suppressed and unaffected.

## Deploy Plan

No deploy needed.

### DIFF
--- a/.grit/patterns/noExpensiveConversationFetch.grit
+++ b/.grit/patterns/noExpensiveConversationFetch.grit
@@ -1,0 +1,22 @@
+engine biome(1.0)
+language js
+
+// Discourages getConversation() and getLightConversation() which load the full
+// conversation content (messages, actions, etc.) and can be expensive.
+// Prefer ConversationResource.fetchById / fetchConversationWithoutContent when
+// message content is not needed. Suppress with:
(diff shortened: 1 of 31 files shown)


Fraggle:

- #29487 ◀

id13 (approved):
LGTM

PR #29490, opened 2026-07-24
[Connectors/Microsoft] Protect refreshed nodes from garbage collection

philipperolet:
## Description

Fixes https://github.com/dust-tt/tasks/issues/9794

Microsoft garbage collection intended to ignore nodes seen after the GC run started, but its filter was parsed as `lastSeenTs ?? (0 < startGarbageCollectionTs)`. As a result, nodes restored by the later full sync stayed GC-eligible. Newly discovered folders could additionally keep a null `lastSeenTs` after traversal.

The fix compares timestamps explicitly in milliseconds, marks successfully traversed folders as seen while preserving timestamps written by newer concurrent syncs, and checks the freshly fetched node timestamp again immediately before a GC deletion. This keeps a node restored after the GC page was loaded from being deleted from the stale page snapshot.

The new activity argument is optional so activity tasks scheduled before deployment remain compatible.

## Risks

Blast radius: Microsoft connector full-sync timestamps and garbage collection deletion guards.
Risk: low

## Deploy Plan

- deploy connectors

### DIFF
--- a/connectors/src/connectors/microsoft/temporal/activities.ts
+++ b/connectors/src/connectors/microsoft/temporal/activities.ts
@@ -2090,7 +2090,7 @@ export async function microsoftGarbageCollectionActivity({
   // during the garbage collection and the caching below prevents this from
   // being detected
   const nodesToCheck = nodes
-    .filter((n) => n.lastSeenTs ?? 0 < startGarbageCollectionTs)
+    .filter((n) => (n.lastSeenTs?.getTime() ?? 0) < startGarbageCollectionTs)
     .filter(
       (n) =>
         n.nodeType === "drive" ||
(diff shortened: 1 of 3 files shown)


connectors/src/connectors/microsoft/temporal/activities.ts
```diff
@@ -623,10 +627,16 @@ export async function markNodeAsSeen(connectorId: ModelId, internalId: string) {
     return;
   }
 
-  // if node was updated more recently than this sync, we don't need to mark it
-  if (node.lastSeenTs && node.lastSeenTs < new Date()) {
-    await node.update({ lastSeenTs: new Date() });
+  // In-flight activities scheduled before startSyncTs was added omit it.
```
philipperolet:
```suggestion
  // No need to mark nodes updated more recently than this sync.
```

tdraier:
LGTM

PR #13806, opened 2025-06-25
Monday.com MCP tool

albandum:
## Description

Add tooling for Monday.com 
Allows personal and workspace creds.

Screenshot 2025-06-25 at 18 01 26

```
Available Tools
Get Boards
Lists all accessible boards in Monday.com workspace

Get Board Items
Retrieves items from a specific Monday.com board

Get Item Details
Retrieves detailed information about a specific Monday.com item

Search Items
Searches for items across Monday.com boards or within a specific board

Create Item
Creates a new item in a Monday.com board

Update Item
Updates column values of an existing Monday.com item

Create Update
Adds an update (comment) to a Monday.com item

Delete Item
Deletes an item from Monday.com (requires high stakes confirmation)

Update Item Name
Updates the name of an existing Monday.com item

Create Board
Creates a new board in Monday.com

Create Column
Creates a new column in a Monday.com board

Create Group
Creates a new group in a Monday.com board

Create Subitem
Creates a new subitem for a Monday.com item

Delete Group
Deletes a group from a Monday.com board

Duplicate Group
Duplicates a group with its items in a Monday.com board

Update Subitem
Updates column values of a Monday.com subitem

Upload File To Column
Uploads a file to a specific column in a Monday.com item

Get Items By Column Value
Finds all items that match a specific value in a selected column

Find User By Name
Searches for a user by name using exact matching

Get Board Values
Returns the data for all columns in a specific board

Get Column Values
Returns the data for a specific column in a board

Get File Column Values
Returns the data for a specific file column in a board

Get Group Details
Returns the group details for a specific board

Get Subitem Values
Returns the subitems for a specific item in a board

Get User Details
Returns the details for a specific user
```

## Deploy Plan

1. Add OAUTH_MONDAY_CLIENT_ID and OAUTH_MONDAY_CLIENT_SECRET to prod 
2. Deploy connectors
3. Deploy front

PR #29491, opened 2026-07-24
[Monday] Use app-configured OAuth scopes

philipperolet:
## Description

Fixes https://github.com/dust-tt/tasks/issues/9779

Dust sent a hard-coded `scope` parameter in Monday OAuth authorization requests. Monday requires an explicit scope list to exactly match the scopes configured for the applicable app version, while accounts can have different approved versions. This caused Monday to reject affected accounts with `invalid_scope` before returning to Dust.

Omit the optional `scope` parameter so Monday uses the scope list configured for the account/app version.

## Risks

Blast radius: Monday OAuth authorization for new connections.
Risk: low

## Deploy Plan
- deploy front

## Tests

- [x] Monday OAuth setup URL uses the app-configured scope list

### DIFF
--- a/front/lib/api/oauth/providers/monday.test.ts
+++ b/front/lib/api/oauth/providers/monday.test.ts
@@ -1,0 +1,29 @@
+import { MondayOAuthProvider } from "@app/lib/api/oauth/providers/monday";
+import { describe, expect, it, vi } from "vitest";
+
+vi.mock("@app/lib/api/config", () => ({
+  default: {
+    getAuthRedirectBaseUrl: () => "https://dust.tt",
+    getOAuthMondayClientId: () => "client-id",
+  },
+}));
+
+describe("MondayOAuthProvider.setupUri", () => {
+  it("uses the scopes configured in the Monday app", () => {
+    const provider = new MondayOAuthProvider();
+    const setupUri = provider.setupUri({
+      connection: {
+        connection_id: "connection-id",
+        created: Date.now(),
+        metadata: {},
+        provider: "monday",
+        status: "pending",
+      },
+      useCase: "personal_actions",
+    });
+
+    const url = new URL(setupUri);
+
+    expect(url.searchParams.has("scope")).toBe(false);
+  });
+});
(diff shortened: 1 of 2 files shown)


philipperolet:
Follow-up investigation on why these scopes were hardcoded and whether omitting them is safe:

- The seven Monday scopes were present in the very first rough implementation commit (`d27d285`), with no accompanying rationale.
- The Monday OAuth provider closely copied the existing Gong provider structure, including the local scope list and explicit `scope` query parameter.
- The scope list was never changed or discussed in the subsequent commits or review comments for #13806. An indexed company-wide Dust search also found no Slack/Notion record of a Monday-specific reason for requiring the explicit parameter. This strongly suggests the pattern was copied as provider boilerplate rather than added to satisfy a Monday requirement.
- Monday documents `scope` as optional. When it is omitted, Monday uses the complete scope list configured for the applicable app version. Explicitly supplied scopes must match that configured list, which explains the `invalid_scope` failures for accounts using a different approved app version.

Therefore, omitting `scope` does **not** mean requesting no permissions, and it should not reduce permissions for correctly configured app versions. It lets Monday select the scopes appropriate to the account/app version.

Remaining caveat: if a deployed Monday app version is itself missing a permission required by one of our tools, OAuth may succeed and the tool could later receive a permission error. The hardcoded parameter did not protect against that: it instead caused OAuth to fail immediately and could not grant permissions absent from the app configuration. A possible follow-up hardening would be to validate the granted `scope` returned by Monday during OAuth finalization and report missing permissions clearly.

References:
- Original implementation: https://github.com/dust-tt/dust/commit/d27d2850061b5a1210fa2b5680d45473d241b5e1
- Original PR: https://github.com/dust-tt/dust/pull/13806
- Monday OAuth docs: https://developer.monday.com/apps/docs/oauth
- Monday app versioning: https://developer.monday.com/apps/docs/app-versioning
- Internal Dust investigation: https://app.dust.tt/w/0ec9852c2f/conversation/NYcNIvE0jG

PR #29494, opened 2026-07-24
[Activation] Descriptive title for nudge convos

xebeohp:
## Description
Creates a more descriptive title for each activation nudge conversation, solving the problem of each conversation being titled "Triggered - Run Scheduled Activation Workflow".

Use the `create_recommendation` tool to set the conversation title to be "Activation Recommendation: [6-8 word title]", reusing the recommendation's existing action label (no extra LLM call that would be needed with the set_conversation_title tool). 
Falls back to "Triggered - ..." if the recommendation is never recorded.

## Tests
Tested locally.

## Risk
Low risk.

## Deploy Plan
Deploy front.

### DIFF
--- a/front/lib/api/actions/servers/activation_recommendations/index.ts
+++ b/front/lib/api/actions/servers/activation_recommendations/index.ts
@@ -9,46 +9,25 @@ import {
   ACTIVATION_RECOMMENDATIONS_SERVER_NAME,
   ACTIVATION_RECOMMENDATIONS_TOOLS_METADATA,
 } from "@app/lib/api/actions/servers/activation_recommendations/metadata";
-import { updateConversationTitle } from "@app/lib/api/assistant/conversation/title";
 import type { Authenticator } from "@app/lib/auth";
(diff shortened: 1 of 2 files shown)


front/lib/api/actions/servers/activation_recommendations/index.ts
```diff
@@ -9,26 +9,47 @@ import {
...
+        conversationId: conversation?.id ?? null,
       });
 
+      // Give the activation nudge conversation a descriptive, user-facing title
+      // derived from the recommendation.
+      if (conversation) {
+        const titleRes = await updateConversationTitle(auth, {
+          conversationId: conversation.sId,
+          title: `Activation Recommendation: ${title}`,
+        });
```
frankaloia: Do we actually want to prompt hte activation skill to update this based on the context of what is being recommended?

xebeohp: That could work better actually.

frankaloia (approved):

PR #29497, opened 2026-07-24
[LLM] Resolve LLM endpoints via ModelInfo instead of legacy modelId

pmilliotte:
## Description

Replaces `modelId`-based LLM resolution with a resolved `endpoint` carried on `ModelInfo`/`LLMParameters`, moving endpoint selection into a shared `selectPreferredEndpointForWorkspace` helper and nesting the legacy `modelConfig` under a static on stream and batch endpoints.

## Tests

Tested locally;

## Risk

Low.

## Deploy Plan

- Deploy `main` on `front`


front/lib/llms/stream/types/configuration.ts
```diff
@@ -8,32 +8,32 @@ export type ReasoningEffortOf<C extends InputConfig> = NonNullable<
...
-      // Description
-      displayName: string;
-      description: string;
+// the new router by nesting the corresponding `*_DEFAULT_MODEL_CONFIG` under a
+// single `modelConfig` static. This is a transitional step: over time we want
+// each endpoint to declare its properties explicitly and drop the ones that are
+// deprecated in the new router (e.g. `regionalAvailability`, `recommendedTopK`,
+// ...) rather than carrying the whole legacy config forward.
+export type DustStreamEndpointConfiguration<C extends InputConfig> = {
+  modelConfig: ModelConfigurationType;
```
pmilliotte, line 17: Put this in a `modelConfig` property to always have access to it (same as getModelConfig(modelId) we used before)

The idea is preparing to remove legacy parameters from this modelConfig to use the new parameters (for instance, do not rely on `regionAvailability`, `isAvailableIfOneOf`, 'supportedReasoningEfforts` and so on)

front/types/assistant/agent_run.ts
```diff
@@ -134,10 +136,20 @@ export type AgentMessageRef = {
...
+  metaData?: Record<string, unknown>;
+};
+
+export type StreamModelInfo = ModelInfo<DustStreamEndpointConstructor>;
+
 export type AgentLoopExecutionData = {
   // No models on the agent configuration as it might be different at run time (eg: auto mode, override by inputbar picker)
   agentConfiguration: AgentConfigurationWithoutModelType;
-  model: AgentModelConfigurationType & ModelConfigurationType;
+  modelInfo: StreamModelInfo;
```
pmilliotte, line 152: This is the big change: we get the endpoint or throw as soon as possible. 
So now we know we have an endpoint the workspace / user has access to, and we can get all the properties we need without having to check their presence.

Possibly this will become just an id later

front/types/assistant/agent_run.ts
```diff
@@ -134,10 +136,20 @@ export type AgentMessageRef = {
   conversationId: string;
 };
 
+export type ModelInfo<E> = {
+  endpoint: E;
+  temperature: number;
+  reasoningEffort?: ReasoningEffort;
+  responseFormat?: string;
+  metaData?: Record<string, unknown>;
```
pmilliotte, line 144: This is needed for recent changes about auto mode, where the temperature / reasoning effort have changed on the fly.

front/lib/llms/batch/endpoints/anthropic_claude_sonnet_four_dot_six_global_anthropic.ts
```diff
@@ -1,8 +1,10 @@
 import { defineDustBatchEndpoint } from "@app/lib/llms/batch/dust_batch_endpoint";
 import { AnthropicClaudeSonnetFourDotSixGlobalAnthropicBatch } from "@app/lib/model_constructors/batch/endpoints/anthropic_claude_sonnet_four_dot_six_global_anthropic";
+import { CLAUDE_SONNET_4_6_DEFAULT_MODEL_CONFIG } from "@app/types/assistant/models/anthropic";
 
 export class DustAnthropicClaudeSonnetFourDotSixGlobalAnthropicBatch extends AnthropicClaudeSonnetFourDotSixGlobalAnthropicBatch {
```
pmilliotte, line 5: Need to add endpoints in EU as well

Will do in additional pr

front/types/assistant/agent_run.ts
aubin-tchoi: not ideal IMHO, why can't auto-mode be decided as upstream as possible and resolve to a regular, fully defined config (I understand you did not write it and have no idea how big of a change this would be)?

front/lib/api/llm/index.ts
```diff
@@ -295,30 +178,11 @@ function getStreamEndpointLLM(
...
-    getBatchEndpoints
-  );
-
-  if (!endpoint) {
-    return null;
-  }
-
-  const modelConfig = getModelConfigByModelId(llmParameters.modelId);
-  const credentials = modelConfig?.useEapKey
-    ? withEapAnthropicKey(llmParameters.modelId, llmParameters.credentials)
```
aubin-tchoi, line 316: did we lose `withEapAnthropicKey` in the batch endpoint?
we most likely don't use it at the moment but could be useful to keep (for internal evals it could be useful to use the batch endpoint)

front/lib/llms/agent_message_content_parser.ts
```diff
@@ -284,22 +283,13 @@ export class AgentMessageContentParser {
...
+  endpoint,
+  reasoningEffort,
+}: StreamModelInfo): DelimitersConfiguration {
+  if (DEEPSEEK_MODELS.includes(endpoint.modelConfig.modelId)) {
     return DEEPSEEK_CHAIN_OF_THOUGHT_DELIMITERS_CONFIGURATION;
   }
-  const reasoningEffort =
-    model.reasoningEffort ?? supportedModel.defaultReasoningEffort;
-  if (reasoningEffort !== "light") {
+  if (reasoningEffort ?? endpoint.modelConfig.defaultReasoningEffort) {
```
aubin-tchoi: this condition seems to have changed from having delimiters defined in `light` mode vs when `null`/`undefined`. I think this makes sense but wondering whether this could cause regressions

front/lib/api/llm/selectPreferredEndpointForWorkspace.ts
```diff
@@ -0,0 +1,108 @@
...
+    workspaceConfiguration: WorkspaceConfig,
+    inputCondition: Where<EndpointConfig>
+  ) => T[],
+  filter: Where<EndpointConfig>
+): Promise<T | null> {
+  const plan = auth.getNonNullablePlan();
+
+  const endpoints = getEndpoints(
+    {
+      featureFlags: await getFeatureFlags(auth),
```
aubin-tchoi: nit: would put the await on its own statement

front/lib/api/llm/selectPreferredEndpointForWorkspace.ts
```diff
@@ -0,0 +1,108 @@
...
+      isEnterprise: isEnterpriseOrDust(plan),
+      isCreditPriced: isCreditPricedPlanPrefix(plan.code),
+    },
+    {
+      ...getWorkspaceFilter(auth),
+      ...filter,
```
aubin-tchoi: nit: shouldn't we destructure `getWorkspaceFilter` last to ensure everything in there is always passed as is?

front/lib/api/llm/selectPreferredEndpointForWorkspace.ts
```diff
@@ -0,0 +1,108 @@
...
+    }
+  );
+
+  const preferredRegion = REGION_MAPPING[multiRegionsConfig.getCurrentRegion()];
+
+  return sortEndpointsByPreferredRegion(endpoints, preferredRegion)[0] ?? null;
+}
+
+export function selectPreferredStreamEndpointForWorkspace(
```
aubin-tchoi: OOC why isn't this function async?

aubin-tchoi: if having the feature flags is the annoying part I think it's fine to drill down an argument, which would allow getting rid of the `Promise.resolve` below

aubin-tchoi (approved):
LGTM

front/lib/api/llm/index.ts
pmilliotte: Seen IRL, custom models support is not in the scope of the new router, will be added when needed.

We never used custom models for batch anyway...

front/lib/api/llm/selectPreferredEndpointForWorkspace.ts
pmilliotte: Good looks, putting both filters inside a `and` clause so there is no overlap

``` 
    {
      and: [getWorkspaceFilter(auth), filter],
    }
```

front/lib/api/llm/selectPreferredEndpointForWorkspace.ts
pmilliotte: I would keep `auth` as a parameter of `selectPreferredEndpointForWorkspace` as this function will be re-used across the codebase. 

Turning `selectPreferredStreamEndpointForWorkspace` async

front/types/assistant/agent_run.ts
pmilliotte: Actually not necessary related to auto mode, but to model selection in the input bar.

PR #29501, opened 2026-07-24
Add Claude Opus 5 as the main Opus model

fontanierh:
## Description

Adds `claude-opus-5` and makes it our default Opus.

- Model config, pricing, tier map, reasoning efforts, `model_constructors` and `llms` stream endpoints, SDK types, model picker config, marketing.
- `isLatest` moves from 4.8 to Opus 5, so the agent builder recommends it in the Anthropic group. 4.8 becomes "previous flagship".
- Auto Complex stream, all `dust-ant*` global agents, `dust-edge` and `deep-dive` now point at Opus 5.

Specs from the [model overview](https://platform.claude.com/docs/en/about-claude/models/overview) and [pricing](https://platform.claude.com/docs/en/about-claude/pricing): same $5/$25 per M and same request contract as 4.8 (adaptive thinking only, no sampling params, no prefill), so it reuses the shared Opus input config. Context is capped at our usual 250k and output at 64k. Keeps the 4.7 tokenizer adjustment since Opus 5 uses that tokenizer.

Global (Anthropic direct) only. Vertex EU needs quota provisioned first, that is a follow-up like we did for 4.8 and Sonnet 5.

## Tests

Updated `enabled_models.test.ts` for the Auto Complex stream now resolving to Opus 5. Ran the model tier, pricing, picker, credit cost and global agent suites locally. The endpoint parity test is added but stays off by default like the other Anthropic ones (needs `RUN_LLM_TEST` plus a key).

## Risk

Low, but one thing to know: Opus 5 is `europe-west1: false` until the Vertex follow-up, so `regionalModelsOnly` workspaces skip it.

- Auto Complex: Opus 4.8 is kept as the explicit second candidate, so those workspaces stay on Anthropic rather than falling through to GPT.
- `dust-ant*` / `dust-edge`: `selectEnabledModel` rejects Opus 5 there, so regional workspaces fall back to `getLargeWhitelistedModel` (Sonnet 4.6 at its default effort, losing the per-agent effort). Same behavior the 4.8 rollout shipped, and these are internal agents.

Otherwise additive. The `isLatest` flip and the agent repointing are one-line reverts. Opus 5 is gated behind `claude_4_5_opus_feature` / advanced-model plans like the rest of the Opus line.

Follow-up for whoever enables Vertex EU: add the `claude-opus-5` entry to `token_pricing/eu.ts` at the same time. `EU_MODEL_PRICING` is a `Partial<Record<...>>` and falls back to global pricing, so a missing entry silently eats the 10% Vertex surcharge instead of failing.

## Deploy Plan

Standard deploy.


aubin-tchoi (approved):

pmilliotte (commented):
LGTM

PR #29503, opened 2026-07-24
[front] Update activation skill execution instructions

frankaloia:
## Summary

Refines the activation skill prompt (`activation.ts`), no code changes:

- Workflow arc is now mandatory ("Every conversation MUST follow the same arc").
- Planning/prefetch runs in a **sub-conversation** via `create_conversation` with polling, and prefetched data is written to a **pod file** that can be shared across conversations.
- New **"Executing a Custom Agent"** step: never hand off the current conversation to the agent — spawn a new conversation via `create_conversation` and poll for completion (avoid sleeps to reduce user-facing latency).
- Clarified `actionMessage` / `dismissMessage` as generated conversation messages.

## Testing

Prompt-only change to a code-defined skill. `tsgo --noEmit` clean; biome clean.

## Risk

Low — instruction text only; no executable code paths changed.

### DIFF
--- a/front/lib/resources/skill/code_defined/global/activation.ts
+++ b/front/lib/resources/skill/code_defined/global/activation.ts
@@ -50,13 +50,13 @@ If the skill is used outside of a pod, you still leverage the recommendation ste
 
 # Workflow Steps
 
-Every conversation MUST follow the same arc:
+Every conversation MUST follow the same arc. At the start of each turn, locate yourself from observable state.
 
 The first set of steps below will result in a single user message to start the activation flow.
 This will generally be triggered asynchronously, so minimizing tool calls is not a requirement. Accuracy and strict adherence to the defined workflow is critical.
 
 1. Research — Gather context about the user and their workspace.
-2. [Only If First Ever Session In This Pod] Set-up the Pod - Create the AGENTS.md file and the pinned Frame
+2. Set-up the Pod - Skip this step if an AGENTS.md file exists for this Pod. If not, create the AGENTS.md file and the pinned Frame
 3. Recommend — Create exactly one high-value recommendation
 4. Create Execution Plan - Create a plan to execute the recommendation. Prefetch all required information to mitigate tool calls during execution.
 5. Present the Recommendation to the User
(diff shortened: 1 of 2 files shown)


front/lib/resources/skill/code_defined/global/activation.ts
```diff
@@ -203,14 +203,15 @@ This is presented a a quickReply option in the welcome message. It should also b
 
 We need to create a plan for the tool calls that will be made to execute the recommendation.
 A goal is to front load as many tool calls as possible to minimize the number of tool calls during execution.
+This should be done by creating a new by creating a new sub-conversation via the \`create_conversation\` tool and polling for completion
```
aubin-tchoi: nit: `create_conversation` is not the full name the agent will see

front/lib/resources/skill/code_defined/global/activation.ts
```diff
@@ -203,14 +203,15 @@ This is presented a a quickReply option in the welcome message. It should also b
...
 3. Some tool calls will block on user approval. Call \`get_tool_execution_modes\` at the start of any multi-step plan to learn which tools
 run silently and which pause for approval. The tool returns a list of \`serverName__toolName: auto | requires_approval\` entries.
 4. Create an ordered list for those tool calls. Deprioritizing the tools that require user approval. This will not ALWAYS be possible, as a tool call might be a hard dependency of another tool call, but we should strive for it.
-5. Make all tool calls up to the first tool that requires user approval to prefetch as much information as possible. If it's too much information to store in context, write it to a file. You should generally focus on read tools, and avoid mutations/writes.
+5. Make all tool calls up to the first tool that requires user approval to prefetch as much information as possible. If it's too much information to store in context, write it to a pod file that can be shared between conversations. You should generally focus on read tools, and avoid mutations/writes.
```
aubin-tchoi: the agent has 0 knowledge of which tool calls require user approval first
would also avoid instructions that are relative to the conversation flow as a whole vs isolated turns as they are typically harder to follow

aubin-tchoi (approved):

front/lib/resources/skill/code_defined/global/activation.ts
frankaloia: Actually had a second thought on this. I don't know why I didn't just use Deep Dive, I think that is much better

front/lib/resources/skill/code_defined/global/activation.ts
frankaloia: I injected knowledge in a previous PR

frankaloia: On the avoiding instructions relative to a conversation flow, the way I normally avoid this is to convert sequences into conditionals keyed on observable states. I think I can do a better job of that here.

PR #29505, opened 2026-07-25
Focus on text content for agent message in getLightConversation()

Fraggle:
## Description

`getLightConversation()` now passes `textContentOnly=true` down through `_getConversation` → `batchRenderMessages` → `batchRenderAgentMessages` → `AgentStepContentResource.fetchByAgentMessages`. In `fetchByAgentMessages`, the new flag adds a `type: "text_content"` WHERE clause to the DB query, so non-text step content (tool outputs, images, etc.) is skipped entirely when rendering light agent messages.

Light conversation consumers — notification workflows, unread counts, pod conversation lists — only need the text portions of agent messages, not the full step content graph. This reduces DB load on those read paths.

## Tests

Tested manually by verifying light conversation endpoints still return correct agent message text content.

## Risk

Low. The `textContentOnly` flag is only set on the `getLightConversation()` path; all full-conversation fetches are unaffected. The DB filter narrows an existing query without changing the schema.

## Deploy Plan

Deploy front.

### DIFF
--- a/front/lib/api/assistant/conversation/fetch.ts
+++ b/front/lib/api/assistant/conversation/fetch.ts
@@ -88,7 +88,8 @@ export const getLightConversation = async (
     branchId,
     "light",
     lastInteractionsToFetchToolOutputContentFor,
-    messagePagination
+    messagePagination,
+    true
   );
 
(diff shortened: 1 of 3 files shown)


Fraggle:

- #29505 ◀

PR #29508, opened 2026-07-25
feat: add model filter in Managed Agents

sfriquet:
## Description

Add a filter on model to filter the managed agents list

## Tests

https://github.com/user-attachments/assets/9d97ef18-df43-4eb9-8dba-bfa0f86e7fc6

## Risk

Frontend only, safe to revert

### DIFF
--- a/front/components/assistant/ModelsFilterMenu.tsx
+++ b/front/components/assistant/ModelsFilterMenu.tsx
@@ -1,0 +1,103 @@
+import { getModelLogoByModelId } from "@app/components/providers/types";
+import { useTheme } from "@app/components/sparkle/ThemeContext";
+import { compareForFuzzySort, subFilter } from "@app/lib/utils";
+import {
+  Button,
+  CpuChip01,
+  DropdownMenu,
+  DropdownMenuCheckboxItem,
+  DropdownMenuContent,
+  DropdownMenuSearchbar,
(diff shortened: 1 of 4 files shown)


sfriquet:

- #29508 ◀
    - #29509

spolu (approved):
LGTM nice!

aubin-tchoi (approved):
LGTM!
Let's make sure we write a PR description, even if ultra concise. Here the useful bits to know are: this is front-end only, safe to revert and only requires deploying front.

avervaet (approved):
Right now the model filter keys everything off modelId alone, but two different providers could use the same modelId, so we might want to key by the (providerId, modelId) pair instead to avoid conflicts in the future. Atm no no providers shares the same modelId but might in the future. 

Otherwise LGTM

sfriquet:
> Right now the model filter keys everything off modelId alone, but two different providers could use the same modelId, so we might want to key by the (providerId, modelId) pair instead to avoid conflicts in the future. Atm no no providers shares the same modelId but might in the future.

True though if that happens we'll have a bigger issue: e.g we often call `getModelConfigByModelId(modelId)` in our codebase.
As for this PR, with duplicate `modelId` even with a compound key we'd end up with a confusing UI (duplicate display name probably?).

PR #29507, opened 2026-07-25
feat: display model with icon in AssistantsTable

sfriquet:
## Description

Screenshot 2026-07-25 at 22 16 25

PR #29509, opened 2026-07-25
feat: bulk set model from Managed Agents

sfriquet:
## Description

Refs https://github.com/dust-tt/tasks/issues/9790

Allow bulk setting model on the Managed Agents list.

- reuses the existing bulk edit UI
- updating models creates a new agent configuration so history of change is captured
- new route `/api/w/:wId/assistant/agent_configurations/batch_update_model` added (we already had `batch_update_scope` and `batch_update_tags`)
- for simplicity's sake, the route is not atomic, each agent is updated through the already existing `createOrUpgradeAgentConfiguration` function

## Tests

- added test for the API
- manually tested:

https://github.com/user-attachments/assets/0a881aa5-e1ce-43e8-b961-2ffc783f9082

### DIFF
--- a/front-api/routes/w/[wId]/assistant/agent_configurations/batch_update_model.test.ts
+++ b/front-api/routes/w/[wId]/assistant/agent_configurations/batch_update_model.test.ts
@@ -1,4 +1,7 @@
-import { getAgentConfiguration } from "@app/lib/api/assistant/configuration/agent";
+import {
+  archiveAgentConfiguration,
+  getAgentConfiguration,
+} from "@app/lib/api/assistant/configuration/agent";
 import type { Authenticator } from "@app/lib/auth";
 import { getModelsForAuth } from "@app/lib/model_tiers/enabled_models";
 import { TagResource } from "@app/lib/resources/tags_resource";
@@ -81,6 +84,45 @@ describe("POST /api/w/:wId/assistant/agent_configurations/batch_update_model", (
     expect(updated?.instructions).toBe(agent.instructions);
     expect(updated?.scope).toBe(agent.scope);
     expect(updated?.tags.map((t) => t.sId)).toEqual([tag.sId]);
+  });
+
+  it("updates what it can and reports the rest as skipped", async () => {
+    const { workspace, auth, agent } = await setupTest("admin");
+    const target = await findTargetModel(auth);
+
+    // An archived agent cannot be re-saved: a new version would make it active again.
+    const archivedAgent = await AgentConfigurationFactory.createTestAgent(
+      auth,
+      {
+        name: "Archived Test Agent",
+        model: { ...INITIAL_MODEL },
+      }
+    );
+    await archiveAgentConfiguration(auth, archivedAgent.sId);
+
+    const res = await postBatchUpdateModel(workspace, {
+      agentIds: [agent.sId, archivedAgent.sId],
+      modelId: target.modelId,
+    });
+
+    expect(res.status).toBe(200);
+    const body = await res.json();
+    expect(body.updatedAgentIds).toEqual([agent.sId]);
+    expect(body.skippedAgentIds).toEqual([archivedAgent.sId]);
+
+    const updated = await getAgentConfiguration(auth, {
+      agentId: agent.sId,
+      variant: "light",
+    });
+    expect(updated?.model.modelId).toBe(target.modelId);
+
+    const skipped = await getAgentConfiguration(auth, {
+      agentId: archivedAgent.sId,
+      variant: "light",
+    });
+    expect(skipped?.model.modelId).toBe(INITIAL_MODEL.modelId);
+    expect(skipped?.status).toBe("archived");
+    expect(skipped?.version).toBe(archivedAgent.version);
   });
 
   it("rejects a model that is not available in the workspace", async () => {
--- a/front-api/routes/w/[wId]/assistant/agent_configurations/batch_update_model.test.ts
+++ b/front-api/routes/w/[wId]/assistant/agent_configurations/batch_update_model.test.ts
@@ -1,0 +1,122 @@
+import { getAgentConfiguration } from "@app/lib/api/assistant/configuration/agent";
+import type { Authenticator } from "@app/lib/auth";
+import { getModelsForAuth } from "@app/lib/model_tiers/enabled_models";
+import { TagResource } from "@app/lib/resources/tags_resource";
+import { AgentConfigurationFactory } from "@app/tests/utils/AgentConfigurationFactory";
+import { createPrivateApiMockRequest } from "@app/tests/utils/generic_private_api_tests";
+import { honoApp } from "@front-api/app";
+import assert from "assert";
+import { describe, expect, it } from "vitest";
+
+const INITIAL_MODEL = {
+  providerId: "openai",
+  modelId: "gpt-5-mini",
+} as const;
+
+function postBatchUpdateModel(workspace: { sId: string }, body: unknown) {
+  return honoApp.request(
+    `/api/w/${workspace.sId}/assistant/agent_configurations/batch_update_model`,
+    {
+      method: "POST",
+      headers: { "Content-Type": "application/json" },
+      body: JSON.stringify(body),
+    }
+  );
+}
+
+// A selectable model that differs from the one the test agents start with.
+async function findTargetModel(auth: Authenticator) {
+  const { models } = await getModelsForAuth(auth);
+  const target = models.find(
+    (m) => m.isSelectable && m.modelId !== INITIAL_MODEL.modelId
+  );
+  assert(target, "Expected another selectable model to be available.");
+  return target;
+}
+
+async function setupTest(role: "admin" | "user") {
+  const { workspace, auth } = await createPrivateApiMockRequest({ role });
+
+  const agent = await AgentConfigurationFactory.createTestAgent(auth, {
+    model: { ...INITIAL_MODEL },
+  });
+
+  const tag = await TagResource.makeNew(auth, {
+    name: "Test Tag",
+    kind: "standard",
+  });
+  await tag.addToAgent(auth, agent);
+
+  return { workspace, auth, agent, tag };
+}
+
+describe("POST /api/w/:wId/assistant/agent_configurations/batch_update_model", () => {
+  it("saves a new version of the selected agents with the new model", async () => {
+    const { workspace, auth, agent, tag } = await setupTest("admin");
+    const target = await findTargetModel(auth);
+
+    const res = await postBatchUpdateModel(workspace, {
+      agentIds: [agent.sId],
+      modelId: target.modelId,
+    });
+
+    expect(res.status).toBe(200);
+    const body = await res.json();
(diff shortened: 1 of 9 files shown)


sfriquet:

- #29507
    - #29508
        - #29509 ◀
            - #29528

front/lib/swr/assistants.ts
```diff
@@ -830,6 +833,81 @@ export function useBatchUpdateAgentTags({
...
+      // Each agent is saved on its own, so some of them may have been skipped.
+      const parsed = BatchUpdateAgentModelResponseBodySchema.safeParse(
+        await res.json()
+      );
+      if (!parsed.success) {
+        sendNotification({
+          type: "info",
+          title: "Model update outcome unknown",
+          description:
+            "The update was submitted but the server response could not be read. " +
```
aubin-tchoi: this error message is very tech-y

front/lib/swr/assistants.ts
```diff
@@ -830,6 +833,81 @@ export function useBatchUpdateAgentTags({
...
+          type: "error",
+          title: "Error updating model",
+          description: `Error: ${errorData.message}`,
+        });
+        return false;
+      }
+
+      // Each agent is saved on its own, so some of them may have been skipped.
+      const parsed = BatchUpdateAgentModelResponseBodySchema.safeParse(
+        await res.json()
```
aubin-tchoi: nit: we typically put awaits on their own statement

front/lib/swr/assistants.ts
```diff
@@ -830,6 +833,81 @@ export function useBatchUpdateAgentTags({
...
+      const res = await clientFetch(
+        `/api/w/${owner.sId}/assistant/agent_configurations/batch_update_model`,
+        {
+          method: "POST",
+          headers: {
+            "Content-Type": "application/json",
+          },
+          body: JSON.stringify({
+            agentIds,
+            ...body,
```
aubin-tchoi: nit: harmless here as the type does not allow for it but we typically spread first to avoid allowing the caller to override the parameters we put first (`agentIds` here)

front/lib/api/assistant/configuration/model_update.ts
```diff
@@ -0,0 +1,176 @@
...
+  }: {
+    agentId: string;
+    model: EnabledModelConfigurationType;
+    reasoningEffort: ReasoningEffort;
+    responseFormat?: string;
+  }
+): Promise<Result<void, Error>> {
+  // Resolves the current configuration along with its editors and skills, and rejects agents
+  // that cannot be re-saved as-is (archived, global).
+  const contextResult = await getAgentConfigurationYAMLContext(auth, agentId, {
```
aubin-tchoi, line 130: OOC why do we go through the YAML representation?

front/lib/api/assistant/configuration/model_update.ts
```diff
@@ -0,0 +1,176 @@
...
+    requireEditorGroup: true,
+  });
+  if (contextResult.isErr()) {
+    return new Err(new Error(contextResult.error.api_error.message));
+  }
+
+  const { agentConfiguration, editorUsers, skills } = contextResult.value;
+
+  if (!agentConfiguration.canEdit && !auth.isAdmin()) {
```
aubin-tchoi, line 139: we should avoid having role checks (isAdmin/isBuilder) outside of the code that enforces the permission; on the side that only checks the permissions we should rely on `canXxx` methods

front/lib/api/assistant/configuration/model_update.ts
```diff
@@ -0,0 +1,176 @@
...
+  if (responseFormat) {
+    const formatValidation = validateResponseFormat(responseFormat);
+    if (!formatValidation.isValid) {
+      return new Err(
+        new Error(`Invalid response format: ${formatValidation.errorMessage}`)
+      );
+    }
+  }
+
+  const results = await concurrentExecutor(
```
aubin-tchoi, line 81: is there a way to batch the db queries (may be very complex depending on what we do in `upgradeAgentConfigurationModel`)?

front/components/assistant/SetModelAssistantsDialog.tsx
```diff
@@ -0,0 +1,216 @@
...
+              e.preventDefault();
+              if (!selectedModel) {
+                return;
+              }
+              setIsSaving(true);
+              const success = await batchUpdateAgentModel(
+                agentConfigurations.map((a) => a.sId),
+                { modelId: selectedModel.modelId }
+              );
+              await mutateAgentConfigurations();
```
aubin-tchoi: I think this await could last for very long

front/components/assistant/SetModelAssistantsDialog.tsx
```diff
@@ -0,0 +1,216 @@
...
+                    disabled={!model.isSelectable}
+                    iconPosition="start"
+                    customItem={
+                      <Label
+                        htmlFor={itemId}
+                        className={cn(
+                          "flex min-w-0 grow items-center gap-2",
+                          model.isSelectable
+                            ? "cursor-pointer"
+                            : "cursor-not-allowed opacity-70"
```
aubin-tchoi: the opacity change should be handled in sparkle IMHO, suprised that's not already the case?

aubin-tchoi (approved):
LGTM!

front-api/routes/w/[wId]/assistant/agent_configurations/batch_update_model.test.ts
Comment (avervaet) [inline front-api/routes/w/[wId]/assistant/agent_configurations/batch_update_model.test.ts:1]: Might be interesting to add a test where some agents succeed and others land in skippedAgentIds (e.g. one archived agent in the batch).

ClementAupiais:
LGTM! I've put @pinotalexandre on fine tuning the selected banner to accept more options (Usage page layout is quite tight rn) cc @sfriquet

front/components/assistant/SetModelAssistantsDialog.tsx
```diff
@@ -0,0 +1,216 @@
...
+            </div>
+          ) : filteredModels.length === 0 ? (
+            <div className="flex items-center justify-center py-8 text-sm text-muted-foreground">
+              No models found
+            </div>
+          ) : (
+            <RadioGroup
+              value={selectedModel?.modelId ?? ""}
+              onValueChange={(value) => {
+                const model = models.find((m) => m.modelId === value);
```
avervaet, line 139: Same as for https://github.com/dust-tt/dust/pull/29508 filtering using modelId only could cause conflicts if two providers shares the same modelId in the future

avervaet (approved):

front/lib/swr/assistants.ts
sfriquet: Agree, this reused the wording from `front/lib/swr/groups.ts`. Not ideal but not sure what's best either. It's very unlikely to happen though.

front/components/assistant/SetModelAssistantsDialog.tsx
sfriquet: True, will update `RadioGroupCustomItem` which doesn't handle it unlike `RadioGroupItem`

front/lib/api/assistant/configuration/model_update.ts
sfriquet: The name of this function is misleading imo it doesn't return YAML (it's used by function that returns YAML)
I needed a way to get agent config + editor + skill and this function happens to return just that.

front/lib/api/assistant/configuration/model_update.ts
sfriquet: We could batch the queries that fetch the agent config / editor / skills but then it sounds simpler to reuse `upgradeAgentConfigurationModel`.
I don't know if it's worth the complexity vs keeping this N+1 that shouldn't be called frequently.

front/lib/api/assistant/configuration/model_update.ts
sfriquet: there are several occurrences of `.canEdit && !auth.isAdmin()`, should we refactor into a `canEditAgent(auth, agent)` helper ?

front/lib/api/assistant/configuration/model_update.ts
sfriquet: refactored in https://github.com/dust-tt/dust/pull/29528

front/components/assistant/SetModelAssistantsDialog.tsx
```diff
@@ -0,0 +1,210 @@
...
+  disabled: boolean;
+  owner: LightWorkspaceType;
+  mutateAgentConfigurations: () => Promise<unknown>;
+}
+
+export function SetModelAssistantsDialog({
+  agentConfigurations,
+  disabled,
+  owner,
+  mutateAgentConfigurations,
```
aubin-tchoi: very nit: I don't think we need to drill down this mutate, we can retrieve it from the hook (with the hook disabled in this case)

front/lib/api/assistant/configuration/model_update.ts
aubin-tchoi: would indeed add a lot of complexity to batch (`createOrUpgradeAgentConfiguration` does bundle an awful lot of logic, it's a pattern we want to avoid but quite costly to refactor)
wfm as is!

front/lib/api/assistant/configuration/model_update.ts
aubin-tchoi: we don't want it in an util either as it would encourage following this pattern of using decentralized permission checks, ideally should really be part of the `canEdit` if what we need here is really the `canEdit`
fine to keep as is if it's already quite widespread + will be refactored anyway in the context of admin governance cc @philipperolet

front/lib/swr/assistants.ts
aubin-tchoi: I think we can put a message more generic as we don't really have much to say here

aubin-tchoi (approved):
LGTM

front/lib/api/assistant/configuration/model_update.ts
sfriquet: we could also use a temporal workflow if we're updating too many agents at the same time

front/lib/api/assistant/configuration/model_update.ts
aubin-tchoi: this makes sense, but stepping back why do we need the skills and editors to update the model config?

aubin-tchoi: and how come we don't do it for the batch update of tags and scope

sfriquet: They update the agent configuration in place instead of creating a new version. It simpler... but it has its cons.
- discrepancy between edit mode: you if edit a single agent it bumps the version, the bulk set tags doesn't. 
- if we don't bump the version, we loose track of the history. We won't be able to say "this agent used model A agent until date X then model B". This can be a problem for instance when backfilling analytics in ES.

aubin-tchoi: I see, that makes a lot of sense (fine to not align the bulk tags and scope update for now), thanks for digging!

PR #28768, opened 2026-07-09
[sparkle] Promote New* → canonical names, old → Legacy*

pinotalexandre:
## Summary
- `NewButton` → `Button`, old `Button` → `LegacyButton`
- `NewSplitButton` → `SplitButton` (`FlexSplitButton`), old → `LegacyFlexSplitButton`
- `NewInput` → `Input`, old `Input` → `LegacyInput`
- `NewInputWithSave` → `InputWithSave`, old → `LegacyInputWithSave`
- All intra-sparkle components updated to import from `LegacyButton`/`LegacyInput`
- `index.ts` exports updated accordingly
- Stories renamed to match (`Actions/Button`, `Actions/LegacyButton`, etc.)

## Test plan

- [ ] Storybook builds and indexes all stories without errors
- [ ] TypeScript compiles with 0 new errors
- [ ] `Actions/Button` story shows the new redesigned button
- [ ] `Actions/LegacyButton` story shows the old button
- [ ] `Forms & Inputs/Input` story shows the new input
- [ ] `Actions/SplitButton` story shows the new split button

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #29510, opened 2026-07-26
[front] fix: clipped UsedByButton since Button redesign

smb2268:
## Description

Fixes [`dust-tt/tasks#9679`.](https://github.com/dust-tt/tasks/issues/9679)

Since the Button redesign (`#28768`), the "Used by" button was clipped to 24px on all its surfaces (Tools list, Skills table, space categories/resources lists, triggers list).

`UsedByButton` passes a composite node (robot icon + usage count + chevron) through the Button `icon` prop with no `label`, so the redesigned Button infers icon-only (`!label && !showCounter && !isSelect && !!icon`) and pins the width to `w-6 px-0` on `xs`, clipping the content. The pre-redesign Button never inferred icon-only (it was an explicit size), which is why this worked before.

Fix: override the icon-only width from the consumer (`w-auto px-2`, merged last through twMerge) so the button sizes to its content, keeping the redesigned xs metrics. Shared Button untouched: no native prop combination can express the dual icon+count content (agents and skills), and `isCounter` renders a pill, which would be a visual change.

## Tests

- `tsgo --noEmit` and biome clean.
- Visually check:
Screenshot 2026-07-26 at 11 48 25 AM
Screenshot 2026-07-26 at 11 57 04 AM

## Risk

Low: scoped to `UsedByButton`'s two Button instances; no shared component changes.

## Deploy Plan

Deploy front.

### DIFF
--- a/front/components/spaces/UsedByButton.tsx
+++ b/front/components/spaces/UsedByButton.tsx
@@ -102,6 +102,11 @@ function UsedByButtonIcon({
   );
 }
 
+// The composite icon (robot + count + chevron) is wider than the icon-only
+// fixed width (w-6 on xs); size to content instead.
+const USED_BY_BUTTON_CLASSES =
+  "w-auto border-0 px-2 hover:bg-muted-background hover:text-foreground";
+
 interface UsedByButtonProps {
   usage: AgentsUsageType | SkillUsageType | null;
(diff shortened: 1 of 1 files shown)


ykmsd (commented):
👍

PR #29511, opened 2026-07-26
files list: emit the [id: fil_xxx] suffix for every linked file

smb2268:
## Description

Fixes dust-tt/tasks#9648.

A user uploaded a `.pptx` named `P13490 - Etude d’opportunités de développement salons aéroport Marseille v3.pptx` (typographic apostrophe, U+2019) and asked the agent to send it to OneDrive. `microsoft_drive__upload_file` failed three times with `File not found`: every call retyped the canonical scoped path with an ASCII apostrophe (U+0027). The object was present at the deterministic mount path the whole time — and the agent had listed the correct name one step before each failing call.

**Root cause:** paths are byte-exact GCS lookup keys, and the file listing offers them as the *only* handle — a 79-character string with `’`, `é`×3 and `ô` the model must reproduce perfectly on every call. The stable handle already exists (`fil_YBECTK90Lve78MRJ14Tq` on the content fragment) and the conversation-file resolvers already accept it; `files__list` even fetched it for every file (`enrichListWithFileResourceIds`) and then discarded it at render time for everything but Frames.

**What this PR does:** drop the interactive-content gate in the listing so every file with a linked `FileResource` surfaces `[id: fil_xxx]`, giving the model a handle it can copy instead of a filename it must reproduce byte-perfect.

Same failure class as `28929` (microsoft_drive: target upload folders by id) — locale-dependent name strings as lookup keys, reported by French customers — and the same direction: prefer the stable id. Scoped paths keep working unchanged.

## Tests

- `files__list` emits `[id: fil_xxx]` for a plain `.pdf` with a linked `FileResource`; omits the suffix when no `FileResource` is linked.
- `npx tsgo --noEmit` + `npx biome check` clean.

## Risk

Worst case, the `[id: ...]` suffix on every listing line confuses an agent whose instructions parse listing output — the format is unchanged otherwise and the suffix already existed for Frames. Safe to rollback.

## Deploy Plan

- [ ] Deploy `front`

### DIFF
--- a/front/lib/api/actions/servers/files/tools/list.test.ts
+++ b/front/lib/api/actions/servers/files/tools/list.test.ts
@@ -9,6 +9,7 @@ import {
   makeExtra,
   setupProjectConversation,
 } from "@app/tests/utils/conversation_test_factories";
+import { FileFactory } from "@app/tests/utils/FileFactory";
 import { createResourceTest } from "@app/tests/utils/generic_resource_tests";
 import { MembershipFactory } from "@app/tests/utils/MembershipFactory";
 import { SpaceFactory } from "@app/tests/utils/SpaceFactory";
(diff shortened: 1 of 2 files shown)


flvndvd (commented):
Actually this is done on purpose, `FileResource` serves one unique goal, uploading content in Dust. Once the content has been uploaded in Dust it should not be use through files anymore. I recall adding a way to serialize when saving the document in GCS, can you share the conversation over Slack? Maybe some LLM encoding issue.

PR #20976, opened 2026-02-02
[front] fix(skills): add renaming on name conflict upon archiving

aubin-tchoi:
## Description

- This PR updates the archiving flow to avoid the unicity constraint on (`workspaceId`, `name`, `status`).
- We currently handle the code paths for `status: active` well but not the one for the archiving.
- This PR renames the pre-existing archived skill on conflict, which is an 80/20 (based on the assumption that this old one will most likely not be re-used)

## Tests

- Tested locally.

## Risk

- Low.

## Deploy Plan

- Deploy front.

PR #29512, opened 2026-07-27
[front] fix(skills): guarantee unique name when archiving same-named skills

davidebbo:
## Description

Follow-up to #20976.

That PR added rename-on-conflict when archiving a skill, to work around the `(workspaceId, name, status)` unique constraint. But the rename target used a **day-granularity** timestamp (`formatTimestampToFriendlyDate(..., "compactWithDay")` → `"archived on Jul 26, 2026"`), which is not unique enough: when the same skill name is archived more than once on the same day, the rename itself collides with a previously renamed archived skill and throws `SequelizeUniqueConstraintError`.

This actually happened with a customer and triggered a monitor: https://dust4ai.slack.com/archives/C05F84CFP0E/p1785094982587329. There were also saw a couple other instances in last 15 days. So not common, but does happen.

This PR keeps the original human-readable date approach but uses the existing `"long"` format, which includes the time down to the **second** (`"archived on July 26, 2026 at 9:41:07 AM"`).  It's a bit uglier, but it's not something users will see frequently.

Note: previously renamed archived skills keep their existing date-based names; this only affects new archives, so no migration is needed.

## Tests

- Added a regression test in `skill_resource.test.ts` that archives three skills sharing the same name, faking the clock so each lands on a distinct time of the same day.
- Verified the test **fails** with the previous day-granularity format and **passes** with the second-granularity format.
- All archive/restore tests pass.

## Risk

- Low.

## Deploy Plan

- Deploy front.

### DIFF
--- a/front/lib/resources/skill/skill_resource.test.ts
+++ b/front/lib/resources/skill/skill_resource.test.ts
@@ -1422,25 +1422,30 @@ describe("SkillResource", () => {
 
     it("archives multiple skills sharing the same name without a unique constraint violation", async () => {
       // The (workspaceId, name, status) unique constraint means only one
-      // archived skill can keep a given name. Archiving a third same-named
-      // skill must not collide with a previously renamed archived skill.
-      const firstSkill = await SkillFactory.create(testContext.authenticator, {
-        name: "Duplicate Name Skill",
-      });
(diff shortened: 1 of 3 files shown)


aubin-tchoi (approved):
LGTM

PR #9676, opened 2024-12-31
poke delete: delete feature flags before workspace

spolu:
## Description

Make sure to delete FeatureFlags when deleting workpace.

From an audit of workspaces foreign key references should the last one.

For: https://github.com/dust-tt/tasks/issues/1873

## Risk

Low

## Deploy Plan

- deploy `front`

PR #29513, opened 2026-07-27
[front] fix: resume files cat inside oversized lines via byte_offset

smb2268:
## Description

Fixes dust-tt/tasks#9676.

`files cat` could not read past the first ~20KB of a single oversized line: the truncation footer suggested `offset=N+1`, which skips the truncated line, making its remaining bytes unreachable — in production, a 23KB single-block instruction file lost its last ~3KB. The byte slice at the cap could also split a multibyte UTF-8 character.

**What this PR does:**
1. Replaces the readline-based reader with a byte-accurate scanner (`readTextFilePage`, in `lib/api/files/text_file_pagination.ts`). Memory is bounded by the page budget plus one stream chunk; content is budgeted at 18KB with a 2KB footer reserve so responses stay under the 20KB offload threshold.
2. Makes `byte_offset` the single continuation cursor: every truncated response's footer returns the byte position of the next unread byte — the start of the next line at a line boundary, or the position inside an oversized line. `offset` remains as an input for direct line jumps (e.g. from `grep` results) but is no longer returned as a cursor. One continuation rule for the model instead of two.
3. Line numbers stay exact on byte resumes (newlines are counted in the skipped prefix), and cuts are UTF-8-safe both ways: outbound truncation and inbound offsets snap to character boundaries.
4. A `byte_offset` for a file that has changed returns a clear error telling the model to restart with `offset=1`.

## Tests

28 tests across scanner and handler: oversized ASCII/multibyte lines, exact reconstruction of ordinary and oversized files following only byte cursors, direct `offset` jumps with byte-cursor continuation, chunk-boundary resumes, CRLF cursors, exact-budget edge cases, staleness errors, and the #9676 payload (a marker beyond byte 20,480 in a >20KB single-line HTML block, retrieved via continuation).

## Risk

Worst case, a stale or invented `byte_offset` errors out or wastes a read; line-based reads behave as before (regression-tested). Safe to rollback — the continuation is stateless.

## Deploy Plan

- [ ] Deploy front. No migration, no ordering constraints.
- [ ] Verify against app.dust.tt (agent with >20KB single-block instructions must answer "Instruction 169" via continuation reads) before resolving dust-tt/tasks#9676.

### DIFF
--- a/front/lib/api/actions/servers/files/metadata.ts
+++ b/front/lib/api/actions/servers/files/metadata.ts
@@ -6,6 +6,7 @@ import {
   INTERACTIVE_CONTENT_SERVER_NAME,
   PUBLISH_INTERACTIVE_CONTENT_FILE_TOOL_NAME,
 } from "@app/lib/api/actions/servers/interactive_content/metadata";
+import { BYTE_OFFSET_SCHEMA } from "@app/lib/api/files/text_file_pagination";
 import { FILE_PREVIEW_DIRECTIVE_EXAMPLE } from "@app/lib/markdown/file_preview";
 import { frameContentType, frameSlideshowContentType } from "@app/types/files";
 import { z } from "zod";
@@ -190,9 +191,12 @@ const FILES_TOOLS_COMMON_METADATA = [
     name: FILES_CAT_ACTION_NAME,
     description:
       "Read the content of a file. " +
-      "For text files, returns lines with their line numbers." +
+      "For text files, returns lines with their line numbers. " +
       "Use `offset` to start at a specific line and `limit` to control how many lines to return. " +
-      "When the output is truncated, a footer indicates the next offset to use. " +
+      "When the output stops at a line boundary, the footer indicates the next `offset` to use. " +
+      "When the output is truncated inside a single oversized line, the footer indicates a `byte_offset` instead: " +
+      "pass it back (without `offset`) to continue reading from that exact position inside the line; " +
+      "using `offset` there would skip the rest of the line. " +
       "For images (JPEG, PNG, GIF), returns a vision block the model can inspect directly. " +
       "For binary documents (PDF, DOCX, PPTX, etc.), call " +
       `\`${getPrefixedToolName(FILES_SERVER_NAME, FILES_EXTRACT_TEXT_ACTION_NAME)}\` first to extract their text content.`,
@@ -207,7 +211,9 @@ const FILES_TOOLS_COMMON_METADATA = [
         .int()
         .min(1)
         .optional()
-        .describe("Line number to start reading from (1-indexed, default 1)"),
+        .describe(
+          "Line number to start reading from (1-indexed, default 1). Do not combine with `byte_offset`."
+        ),
       limit: z
         .number()
         .int()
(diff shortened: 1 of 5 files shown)


front/lib/api/files/text_file_pagination.ts
```diff
@@ -0,0 +1,428 @@
...
+      };
+    }
+    return { outcome: "ok", text: `\`${path}\` is empty.` };
+  }
+
+  let text = page.parts.join("\n");
+
+  const kb = FILE_OFFLOAD_TEXT_SIZE_BYTES / 1024;
+
+  if (page.endedMidLine) {
```
frankaloia: Do we really need to distinguish between stopping at a line boundary and stopping inside a line for continuation?

Could we always return a `byte_offset` whenever there’s more content to read? If we stopped at a line boundary, it would just point to the first byte of the next line. If we stopped inside an oversized line, it would point to the continuation position.

We could keep `offset` as an input for jumping directly to a line, but always use `byte_offset` in the continuation footer. That feels simpler and avoids teaching the model two different ways to continue reading.

front/lib/api/actions/servers/files/metadata.ts
```diff
@@ -190,9 +191,12 @@ const FILES_TOOLS_COMMON_METADATA = [
     name: FILES_CAT_ACTION_NAME,
     description:
       "Read the content of a file. " +
-      "For text files, returns lines with their line numbers." +
+      "For text files, returns lines with their line numbers. " +
       "Use `offset` to start at a specific line and `limit` to control how many lines to return. " +
-      "When the output is truncated, a footer indicates the next offset to use. " +
+      "When the output stops at a line boundary, the footer indicates the next `offset` to use. " +
+      "When the output is truncated inside a single oversized line, the footer indicates a `byte_offset` instead: " +
+      "pass it back (without `offset`) to continue reading from that exact position inside the line; " +
```
frankaloia: If byte_offset becomes the continuation cursor in both cases, we should update the descriptions to say it is the byte position returned by the previous response, not something reserved only for oversized lines. offset can remain available for direct line-based reads, but wouldn’t be returned as the “read more” cursor anymore.

front/lib/api/files/text_file_pagination.ts
```diff
@@ -0,0 +1,428 @@
...
+/**
+ * Scans a text file stream and produces one page of numbered display lines.
+ *
+ * Line-based reads skip whole lines up to `startLine` (classic offset pagination).
+ * When `byteOffset` is set, exactly that many bytes are skipped (counting newlines to keep
+ * line numbers accurate) and emission resumes from inside the line at that position; an
+ * offset landing inside a multibyte character is snapped backward to the character start.
+ *
+ * When a single line exceeds the whole byte budget of an empty page, the largest
+ * UTF-8-boundary-safe prefix is emitted and `endedMidLine` carries the byte position to
```
frankaloia, line 144: If we go this route, could we expose the next absolute byte position on the page, maybe as nextByteOffset or consumedBytes? Then the renderer can always use that value for the continuation footer, whether we stopped at a line boundary or in the middle of a line.

front/lib/api/files/text_file_pagination.ts
smb2268: Yeah good feedback - updated with all your suggestions

frankaloia (approved):
Much cleaner. Looks correct to me

PR #29514, opened 2026-07-27
[front] fix: 4xx personal OAuth setup when the workspace connection is missing or dangling

smb2268:
## Description

Fixes dust-tt/tasks#9621. Setting up a personal OAuth connection crashes with an unhandled 500 (DD monitor 22026960) and a blank popup when the tool's workspace-level admin connection is missing — or exists but its `connectionId` no longer resolves in the OAuth service (one confirmed path: workspace relocation copies front rows verbatim while OAuth-service data is regional). Providers treat this as fatal deep in `getUpdatedExtraConfig` with a bare throw.

**Fix:**

- **Preflight** in `createConnectionAndGetSetupUrl`: for `personal_actions` with an `mcp_server_id`, verify the workspace connection (front-DB lookup + OAuth-service probe) before any provider code runs. The two broken-reference states from the ticket — missing row, dangling id — become a typed `mcp_server_connection_not_found` → 404 with an actionable "ask a workspace admin" message. Everything else (transient OAuth-service failures, unexpected states like a credential-backed connection on a personal OAuth attempt) deliberately passes the check and keeps failing loudly (pinned by tests). Covers every provider using this pattern with no per-provider changes.
- **Popup** displays the message for this error type only; all other errors keep the generic copy (internal messages can carry raw upstream details).
- **Jira** gets `requiresWorkspaceConnectionForPersonalAuth = true` like its 12 siblings, so the popup isn't opened at all when the workspace connection row is missing.

This detects and explains broken references; it does not repair them — the admin still reconnects the tool.

## Risk

Worst case, a preflight false positive blocks a setup that would have succeeded — it fails only on deterministic states; anything ambiguous passes through to today's behavior. Safe to rollback.

## Tests

- Route: missing row → 404, dangling `connectionId` → 404.
- Helper unit tests: valid / missing / dangling connection, and transient OAuth-service failures passing through.
- Component: popup shows the actionable message for `mcp_server_connection_not_found`, generic copy otherwise.

### DIFF
--- a/front/lib/api/oauth/mcp_server_connection_auth.ts
+++ b/front/lib/api/oauth/mcp_server_connection_auth.ts
@@ -1,5 +1,8 @@
+import config from "@app/lib/api/config";
 import type { Authenticator } from "@app/lib/auth";
 import { MCPServerConnectionResource } from "@app/lib/resources/mcp_server_connection_resource";
+import logger from "@app/logger/logger";
+import { OAuthAPI } from "@app/types/oauth/oauth_api";
 import type { Result } from "@app/types/shared/result";
 import { Err, Ok } from "@app/types/shared/result";
 import { isString } from "@app/types/shared/utils/general";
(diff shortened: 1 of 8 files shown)


front/lib/api/oauth/mcp_server_connection_auth.ts
```diff
@@ -61,3 +64,62 @@ export async function getWorkspaceOAuthConnectionIdForMCPServer(
...
+): Promise<Result<undefined, WorkspaceMCPServerAuthRefError>> {
+  const connectionIdRes = await getWorkspaceOAuthConnectionIdForMCPServer(
+    auth,
+    mcpServerId
+  );
+  if (connectionIdRes.isErr()) {
+    if (connectionIdRes.error.kind !== "connection_not_found") {
+      return new Ok(undefined);
+    }
+    // The original diagnostic is replaced by a user-facing message below — log it here.
```
frankaloia: nit: comment probably not needed

frankaloia (approved):

PR #29516, opened 2026-07-27
[dust-hive] perf: speed up environment spawn

aubin-tchoi:
## Description

Goal is to reduce the time between running `dhx` and having a new hive environment ready.

Measurements below cover the actual hive setup. Codex startup and interactive tmux attachment are excluded. The baseline critical path after CLI startup was:

| Phase | Time |
| --- | ---: |
| Git worktree checkout | 2.40s |
| Test database setup | 1.10s |
| Dependency and local config setup | 1.36s |
| Build watchers and tmux session | 0.48s |

The Git worktree checkout was the largest individual operation and the clearest removable cost was the 1.10s test database setup: it was independent from the checkout, but ran immediately before it. The two operations now run concurrently, and database setup completes before checkout finishes.

The spawn also walked the repository once for `AGENTS.local.md` and again for `AGENTS.override.md`. Both patterns are now handled by a single scan, reducing the isolated scan time from about 0.41s to 0.24s.

## Tests

- Benchmarked locally before and after.

## Risk

- Low, limited to internal tooling.

## Deploy Plan

- No deploy.

### DIFF
--- a/x/henry/dust-hive/src/commands/spawn.ts
+++ b/x/henry/dust-hive/src/commands/spawn.ts
@@ -278,17 +278,17 @@ export async function spawnCommand(options: SpawnOptions): Promise<Result<void>>
   const filesResult = await setupEnvironmentFiles(metadata, ports, settings);
   if (!filesResult.ok) return filesResult;
 
-  // Phase 1.5: Create test database (if test postgres is running)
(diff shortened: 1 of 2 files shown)


tdraier (approved):
LGTM

PR #29517, opened 2026-07-27
[front] Reconcile pool on credit.create for pool credits

tdraier:
## Problem

When switching a contract with a **recurring free AWU credit** (the `recurringFreeCredit` option in Poke's Switch Contract dialog), the workspace pool credit state was left stuck (e.g. `depleted`) even though the contract had credits.

Root cause is an ordering issue in `switchContract`:

1. `provisionMetronomeContract` creates the contract → it becomes active → Metronome fires **`contract.start`** → the handler reconciles the pool (and per-user states) via `applyContractStartSubscriptionSwap`.
2. **Only after that**, `stepContractEdits` (and the seat sync) add the recurring credits.

So `contract.start` reconciled **before the credits existed**. The `credit.segment.start` handler already reconciles, but for a recurring credit added to an *already-active* contract Metronome fires only **`credit.create`** (the first segment is active at creation) — not a fresh `credit.segment.start`. And `credit.create` only stamped the credit; it never reconciled. Result: the state stayed at whatever `contract.start` computed before the credit landed.

This affects both the **pool** state (recurring free / pool credits) and the **per-user seat** state (per-seat `INDIVIDUAL` credits), by the same mechanism.

## Fix

Reconcile on `credit.create`, mirroring both reconciles `credit.segment.start` performs, split by allocation. The same split is now also applied in the existing `credit.segment.start` / `credit.edit` handler.

- **Pool AWU credit** (e.g. the recurring free credit) → reconcile the workspace pool credit state.
  - **Stamp first, then reconcile.** The pool balance is read with an `onlyPoolCredits` filter (`getNetBalance` filters on `custom_fields: { DUST_CONTRACT_CREDIT_TYPE: "pool" }`), so an unstamped credit is invisible to the reconcile. We stamp `DUST_CONTRACT_CREDIT_TYPE=pool`, then reconcile.
- **Per-seat (`INDIVIDUAL`) AWU credit** → reconcile per-user credit states (`launchReconcileWorkspaceUserCreditStatesWorkflow`), **not** the pool. Seat credits never count toward the pool balance, so a pool reconcile there was a wasted no-op `getNetBalance`. The reconcile workflow is workspace-scoped with `USE_EXISTING`, so onboarding many seats collapses into a single reconcile-all-users run.

### Why not set the custom field on the recurring-credit template instead?

Not possible with Metronome (`@metronome/sdk` v3.6.0): `add_recurring_credits[]` has no `custom_fields` field (unlike one-off `add_credits` / `add_commits`), there is no `recurring_credit` entity in `customFields.setValues`, and custom fields are not among the template fields documented to propagate to materialized instances. Per-instance stamping is the only option.

## Tests

Added a `credit.create` block to `process_webhook.test.ts`:
- reconciles the pool (and not per-user) for a pool AWU credit,
- reconciles per-user (and not pool) for a per-seat `INDIVIDUAL` credit,
- does nothing for a non-AWU credit.

All 28 tests in the file pass; `tsgo` and biome are clean.

### DIFF
--- a/front/lib/api/metronome/process_webhook.test.ts
+++ b/front/lib/api/metronome/process_webhook.test.ts
@@ -985,8 +985,6 @@ describe("processMetronomeWebhook — credit.create pool reconcile", () => {
       timestamp: new Date().toISOString(),
       credit_id: CREDIT_ID,
       credit_custom_fields: null,
-      // No contract_id → stampContractCreditType stamps "pool" without a
-      // contract fetch, keeping the test focused on the reconcile.
       contract_id: null,
       contract_custom_fields: null,
       parent_recurring_credit_id: null,
@@ -1021,7 +1019,7 @@ describe("processMetronomeWebhook — credit.create pool reconcile", () => {
     ).mockResolvedValue(undefined);
   });
 
-  it("reconciles the pool when an AWU credit is created", async () => {
+  it("stamps and reconciles the pool when a pool AWU credit is created", async () => {
     const workspace = await setupMetronomeWorkspaceResource();
     vi.mocked(getMetronomeCredit).mockResolvedValue(
       new Ok(credit(getCreditTypeAwuId()))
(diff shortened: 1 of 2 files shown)


fabiencelier (approved):
LGTM

PR #29519, opened 2026-07-27
Admin governance: show skill availability column in the manage skills page

fabiencelier:
## Description

- the skills table shows an Access column with the availability of each skill (Published / Unpublished / Discoverable).
- default tab has been renamed "Auto-discoverable" + added a description for it
- all behind the `admin_governance_skill_publication` flag 

## Tests

tested locally

image

image

## Risk

low: behind feature flag + purely UI + easy to revert

## Deploy Plan

deploy front

### DIFF
--- a/front/components/pages/builder/skills/ManageSkillsPage.tsx
+++ b/front/components/pages/builder/skills/ManageSkillsPage.tsx
@@ -10,7 +10,11 @@ import {
   useSetPageTitle,
 } from "@app/components/sparkle/AppLayoutContext";
 import { useHashParam } from "@app/hooks/useHashParams";
-import { useAuth, useWorkspace } from "@app/lib/auth/AuthContext";
+import {
+  useAuth,
+  useFeatureFlags,
+  useWorkspace,
+} from "@app/lib/auth/AuthContext";
 import { isDustProvidedSkill, SKILL_ICON } from "@app/lib/skill";
 import { useSkillsWithRelations } from "@app/lib/swr/skill_configurations";
 import { compareForFuzzySort, subFilter } from "@app/lib/utils";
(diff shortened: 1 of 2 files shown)


front/components/pages/builder/skills/ManageSkillsPage.tsx
```diff
@@ -330,6 +342,12 @@ export function ManageSkillsPage() {
               </div>
             ) : (
               <>
+                {activeTab === "default" && hasSkillPublicationGovernance && (
+                  <div className="pt-4 text-sm text-muted-foreground dark:text-muted-foreground-night">
+                    Auto-discoverable skills are published and can be
+                    automatically activated with the Discover Skills tool.
```
ykmsd: we should not use the word published here, something like this??
```
Auto-discoverable skills are visible to every workspace member and can be automatically activated with the Discover Skills tool.
```

ykmsd: also this layout is sad 🥲 I know this comes from the design but let's remove it for now until we settle, I pinged alex and you on figma

Image

front/components/skills/SkillsTable.tsx
```diff
@@ -29,6 +31,15 @@ type RowData = {
   menuItems: MenuItem[];
 };
 
+export const SKILL_AVAILABILITY_DISPLAY: Record<
+  SkillAvailability,
+  { label: string; color: "primary" | "success" | "highlight" }
+> = {
+  editors: { label: "Unpublished", color: "primary" },
+  workspace_users: { label: "Published", color: "success" },
+  users_and_agents: { label: "Discoverable", color: "highlight" },
```
ykmsd: I think we should name it to 
- Editor only
- Workspace members
- Auto-discoverable
? @pinotalexandre

ykmsd (approved):
figma is outdated and we need to adjust the naming ^^

PR #29521, opened 2026-07-27
Prevent seat change when subscription has ended

tdraier:
## Description

Fixes a Metronome `400` error ("Subscription seat remove starting at date must be before the subscription ends") that occurred when an admin changed a member's seat while the workspace's subscription already had a cancellation scheduled (contract end date set via Stripe cancel-at-period-end or the in-app "Cancel subscription" flow).

**Root cause:** once a subscription's cancellation is scheduled, its Metronome contract already has a scheduled end date (`ending_before`). A subsequent seat change computes a deferred `starting_at` for the seat add/remove call that can land at or after that contract end date, which Metronome rejects.

**Fix:** block seat-change actions in the UI whenever `subscription.endDate` or `subscription.requestCancelAt` is set (new shared helper `isSubscriptionCancellationScheduled` in `types/plan.ts`, extracted from existing inline logic in `SubscriptionContext.tsx`):

- `ChangeSeatModal` / `BulkChangeSeatModal`: disable Validate/Review and show an explanatory message.
- `MembersUsageTable`: new `seatActionsDisabled` prop disables Assign/Change/Remove seat row actions specifically, without touching unrelated actions (spend limit, model tier).
- `UsagePage`: computes the flag, wires it through, and hides the "Batch change seat" bulk action entirely when cancelled.

Verified this does not misfire during subscription switches/upgrades: that path (`activatePending`/`swapMetronomeContract`, driven by the `contract.start` webhook) ends the old subscription atomically via `markAsEnded` (status + `endDate = now` in the same transaction as the new one activating) rather than staging a future `endDate` on a still-`active` subscription, so an in-flight upgrade is never mistaken for a cancellation.

## Tests

Manually verified on the usage page: with a subscription that has a scheduled cancellation, per-member seat change, remove seat, and batch seat change are all disabled with an explanatory message; verified an in-flight subscription upgrade (no cancellation) does not trigger the guard.

## Risk

Low. Purely additive UI guard — no change to the seat-change backend logic. Worst case is over-blocking (a seat change that would have actually succeeded gets blocked) rather than letting an invalid Metronome call through.

## Deploy Plan

Standard frontend deploy, no migration or backend coordination required.

### DIFF
--- a/front-api/routes/poke/workspaces/[wId]/seat_type.ts
+++ b/front-api/routes/poke/workspaces/[wId]/seat_type.ts
@@ -89,6 +89,15 @@ app.post(
               message: "Failed to sync seat change with Metronome.",
             },
           });
+        case "subscription_cancellation_scheduled":
+          return apiError(ctx, {
+            status_code: 400,
+            api_error: {
+              type: "invalid_request_error",
+              message:
+                "This workspace's subscription has a cancellation scheduled; seats cannot be changed until it's reactivated or has fully ended.",
+            },
+          });
         default:
           assertNever(updateRes.error.type);
       }
--- a/front-api/routes/w/[wId]/members/[uId]/seat-type.ts
+++ b/front-api/routes/w/[wId]/members/[uId]/seat-type.ts
@@ -119,6 +119,15 @@ app.patch(
               message: "Failed to update seat in billing system.",
             },
           });
+        case "subscription_cancellation_scheduled":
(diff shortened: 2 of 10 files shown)


tdraier:

- #29521 ◀
    - #29522

fabiencelier (approved):
LGTM for the UI part, but maybe add a server side validation too ?

PR #29523, opened 2026-07-27
[front] improve space and pod section wording

ykmsd:
## Description
This PR is to update the wording of spaces and pods selection and improve the wording of the message when it has access to restricted space in both Agent builder and skill builder. Note that this will be shipped without the feature flag.

If there is only one:
CleanShot 2026-07-27 at 16 34 43@2x

If there are more than two:
CleanShot 2026-07-27 at 16 34 26@2x

### DIFF
--- a/front/components/agent_builder/AgentBuilderSpacesBlock.tsx
+++ b/front/components/agent_builder/AgentBuilderSpacesBlock.tsx
@@ -176,8 +176,7 @@ export function AgentBuilderSpacesBlock({
             Visibility control and available data
           </h2>
           <p className="text-sm text-muted-foreground">
-            Select a space or pod to make this agent private to its members and
-            add access to its data and tools.
+            Add a space or pod to restrict usage to its members and make all its data available to this agent.
           </p>
         </div>
         <Button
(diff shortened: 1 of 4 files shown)


front/components/agent_builder/AgentBuilderSpacesBlock.tsx
```diff
@@ -171,9 +172,11 @@ export function AgentBuilderSpacesBlock({
     <div className="space-y-3">
       <div className="flex items-start justify-between">
         <div>
-          <h2 className="heading-lg text-foreground">Spaces and Pods</h2>
+          <h2 className="heading-lg text-foreground">
```
ykmsd, line 175: we should have reusable component for this in agent and skills builder but won't do it today 🙏

front/components/shared/SpaceRestrictionMessage.tsx
```diff
@@ -0,0 +1,56 @@
...
+  entityName,
+  owner,
+  spaces,
+}: SpaceRestrictionMessageProps) {
+  if (spaces.length === 0) {
+    return null;
+  }
+
+  return (
+    <div className="mb-4 w-full">
```
ykmsd, line 46: this should not live here and we should control spacing at parent level but that's another quest

fabiencelier (approved):
Nice cleaning ✨

PR #29525, opened 2026-07-27
[LLM] Add OpenAI GPT-5.5 EU batch endpoint

pmilliotte:
## Description

Adds the OpenAI GPT-5.5 EU batch endpoint, resolving the OpenAI batch base URL per endpoint (as the stream client already does) instead of through a `OPENAI_BASE_URL` credential, and declares the `modelConfig` static that was missing on the new Gemini 3.5 Flash Lite / 3.6 Flash config mixins and batch endpoint.

## Tests

`npm run tsgo` passes; the EU batch endpoint has not been exercised against the live API yet.

## Risk

Low.

## Deploy Plan

- Deploy `main` on `front`


fabiencelier (commented):
LGTM

PR #29527, opened 2026-07-27
Admin governance: bulk edit UI for manage skills table

fabiencelier:
## Description

related to https://github.com/dust-tt/tasks/issues/9738

Add a bulk edit mode to the manage skills page
Only visible when feature flag is on
Only option for now is to bulk edit the availability (bulk archive will come next)

## Tests

tested locally

image

Screenshot 2026-07-27 at 15 19 32

Screenshot 2026-07-27 at 15 19 37
Screenshot 2026-07-27 at 15 19 42
Screenshot 2026-07-27 at 15 19 48

Screenshot 2026-07-27 at 15 19 55

## Risk

low: purely UI + behind feature flag

## Deploy Plan

deploy front

### DIFF
--- a/front/components/pages/builder/skills/ManageSkillsPage.tsx
+++ b/front/components/pages/builder/skills/ManageSkillsPage.tsx
@@ -242,31 +242,28 @@ export function ManageSkillsPage() {
     [rowSelection]
   );
 
-  const closeBatchEdition = useCallback(() => {
+  const closeBatchEdition = () => {
     setIsBatchEditing(false);
     setRowSelection({});
-  }, []);
+  };
 
-  const handleBatchAvailability = useCallback(
-    async (availability: SkillAvailability) => {
-      if (selectedSkillIds.length === 0 || isBatchUpdating) {
-        return;
+  const handleBatchAvailability = async (availability: SkillAvailability) => {
+    if (selectedSkillIds.length === 0 || isBatchUpdating) {
+      return;
+    }
+    setIsBatchUpdating(true);
+    try {
+      const success = await doUpdateAvailability(
+        selectedSkillIds,
+        availability
+      );
+      if (success) {
+        setRowSelection({});
       }
-      setIsBatchUpdating(true);
-      try {
-        const success = await doUpdateAvailability(
-          selectedSkillIds,
-          availability
-        );
-        if (success) {
-          setRowSelection({});
-        }
-      } finally {
-        setIsBatchUpdating(false);
-      }
-    },
-    [selectedSkillIds, isBatchUpdating, doUpdateAvailability]
-  );
+    } finally {
+      setIsBatchUpdating(false);
+    }
+  };
 
   const isBatchEditionAvailable =
     hasSkillPublicationGovernance &&
--- a/front/components/pages/builder/skills/ManageSkillsPage.tsx
+++ b/front/components/pages/builder/skills/ManageSkillsPage.tsx
@@ -21,6 +21,7 @@ import {
   useWorkspace,
 } from "@app/lib/auth/AuthContext";
 import { isDustProvidedSkill, SKILL_ICON } from "@app/lib/skill";
+import { useWorkspacePermissions } from "@app/lib/swr/permissions";
 import {
   useSkillsWithRelations,
   useUpdateSkillsAvailability,
@@ -97,6 +98,7 @@ function sortSkillsByName(
 export function ManageSkillsPage() {
   const owner = useWorkspace();
   const { user } = useAuth();
+  const { hasPermission } = useWorkspacePermissions();
   const { hasFeature } = useFeatureFlags();
   const [selectedSkill, setSelectedSkill] =
     useState<SkillWithoutInstructionsAndToolsWithRelationsType | null>(null);
@@ -267,7 +269,9 @@ export function ManageSkillsPage() {
   );
 
   const isBatchEditionAvailable =
-    hasSkillPublicationGovernance && activeTab !== "archived";
+    hasSkillPublicationGovernance &&
+    hasPermission("publish", "skill") &&
+    activeTab !== "archived";
 
   const knownSkillsById = useMemo(
     () =>
@@ -424,12 +428,7 @@ export function ManageSkillsPage() {
                         ? "Auto-discoverable"
                         : tab.label
                     }
-                    onClick={() => {
-                      setSelectedTab(tab.id);
-                      // Selected rows belong to the current tab: drop the
-                      // selection when it changes.
-                      setRowSelection({});
-                    }}
+                    onClick={() => setSelectedTab(tab.id)}
                     tooltip={tab.description}
                     isCounter={tab.id !== "archived"}
                     counterValue={`${skillsByTab[tab.id].length}`}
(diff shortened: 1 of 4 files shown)


front/components/pages/builder/skills/ManageSkillsPage.tsx
```diff
@@ -211,6 +282,45 @@ export function ManageSkillsPage() {
     [setSkillIdParam]
   );
 
+  const [isBatchUpdating, setIsBatchUpdating] = useState(false);
+
+  const selectedSkillIds = useMemo(
+    () =>
+      Object.entries(rowSelection)
```
fabiencelier, line 239: ok for simplicity here I think

front/components/skills/SkillsBatchEdit.tsx
```diff
@@ -0,0 +1,164 @@
...
+  getDialogTitle: (count: number) => string;
+  dialogDescription: string;
+  confirmLabel: string;
+  confirmVariant: "primary" | "warning";
+};
+
+const BATCH_AVAILABILITY_ACTIONS: BatchAvailabilityAction[] = [
+  {
+    label: "Editor only",
+    icon: EyeOff,
```
ykmsd: We don't show icons in skill builder so let's skip it 🙏

front/components/pages/builder/skills/ManageSkillsPage.tsx
```diff
@@ -211,6 +230,45 @@ export function ManageSkillsPage() {
...
+        .map(([sId]) => sId),
+    [rowSelection]
+  );
+
+  const closeBatchEdition = useCallback(() => {
+    setIsBatchEditing(false);
+    setRowSelection({});
+  }, []);
+
+  const handleBatchAvailability = useCallback(
```
ykmsd: here too 👮

front/components/pages/builder/skills/ManageSkillsPage.tsx
```diff
@@ -211,6 +230,45 @@ export function ManageSkillsPage() {
...
+
+  const selectedSkillIds = useMemo(
+    () =>
+      Object.entries(rowSelection)
+        .filter(([, selected]) => selected)
+        .map(([sId]) => sId),
+    [rowSelection]
+  );
+
+  const closeBatchEdition = useCallback(() => {
```
ykmsd: useCallback police here, you don't need it here no? 🚓

front/components/skills/SkillsBatchEdit.tsx
```diff
@@ -0,0 +1,164 @@
...
+      onOpenChange={(open) => {
+        if (!open && !isUpdating) {
+          onCancel();
+        }
+      }}
+    >
+      <DialogContent size="md" isAlertDialog>
+        <DialogHeader hideButton>
+          <DialogTitle>{action.getDialogTitle(selectedCount)}</DialogTitle>
+          <DialogDescription>{action.dialogDescription}</DialogDescription>
```
ykmsd: It feels tight between title and description, I think DialogDescription should be outside of DialogHeader?

Image

front/components/pages/builder/skills/ManageSkillsPage.tsx
```diff
@@ -338,7 +424,12 @@ export function ManageSkillsPage() {
                         ? "Auto-discoverable"
                         : tab.label
                     }
-                    onClick={() => setSelectedTab(tab.id)}
+                    onClick={() => {
+                      setSelectedTab(tab.id);
+                      // Selected rows belong to the current tab: drop the
+                      // selection when it changes.
+                      setRowSelection({});
```
ykmsd: I think if you will lose the state we should show a confirmation dialog to users no?

front/components/skills/SkillsBatchEdit.tsx
fabiencelier: removed

front/components/pages/builder/skills/ManageSkillsPage.tsx
```diff
@@ -211,6 +230,45 @@ export function ManageSkillsPage() {
...
+          setRowSelection({});
+        }
+      } finally {
+        setIsBatchUpdating(false);
+      }
+    },
+    [selectedSkillIds, isBatchUpdating, doUpdateAvailability]
+  );
+
+  const isBatchEditionAvailable =
```
ykmsd, line 268: We should also check if a user has publish skill permission too?

ykmsd: you can use useWorkspacePermissions hook

front/components/skills/SkillsBatchEdit.tsx
```diff
@@ -0,0 +1,156 @@
...
+};
+
+const BATCH_AVAILABILITY_ACTIONS: BatchAvailabilityAction[] = [
+  {
+    label: "Editor only",
+    availability: "editors",
+    getDialogTitle: (count) =>
+      `Make ${count} skill${pluralize(count)} editor only`,
+    dialogDescription:
+      "Editor only skills will no longer appear in the builder. They will keep working inside agents that already use them.",
```
ykmsd: what do you think of something like this? It's not only in the builder but also in the input bar, and also not only agents but nested skills 🙃
`Only editors will be able to select these skills. Agents and skills that already use them won't lose access.`

ykmsd: hmm not sure if "select" is a good term

ykmsd: Or `Non-editors won't see these skills as options in the builder` and drop about the input bar?

front/components/skills/SkillsBatchEdit.tsx
```diff
@@ -0,0 +1,156 @@
...
+    confirmLabel: "Make editor only",
+    confirmVariant: "warning",
+  },
+  {
+    label: "Workspace members",
+    availability: "workspace_users",
+    getDialogTitle: (count) =>
+      `Make ${count} skill${pluralize(count)} available to workspace members`,
+    dialogDescription:
+      "Skills available to workspace members are visible to everyone. They can add them to agents, other skills and use them directly.",
```
ykmsd: `Skills available to workspace members are visible to everyone` is kind of saying the same thing twice?

ykmsd: How about `Every workspace members can add them to agents, other skills and use them directly.`?

front/components/skills/SkillsBatchEdit.tsx
ykmsd: seems like we should use DialogContainer? Weird name (container inside the content) 🤨

ykmsd (approved):
ty! put a bunch of random nits mostly around wording 🫡

front/components/pages/builder/skills/ManageSkillsPage.tsx
fabiencelier: Actually I think it makes more sense to not lose the state when switching

front/components/skills/SkillsBatchEdit.tsx
fabiencelier: A bit more spacing with the DialogContainer indeed 
image

front/components/pages/builder/skills/ManageSkillsPage.tsx
fabiencelier: Indeed 🤦
Nice catch !

front/components/pages/builder/skills/ManageSkillsPage.tsx
ykmsd: oh if we can do it that's 100% better 😄

front/components/pages/builder/skills/ManageSkillsPage.tsx
fabiencelier: removed indeed

front/components/pages/builder/skills/ManageSkillsPage.tsx
fabiencelier: removed too

PR #29531, opened 2026-07-27
[LLM] Rename Google endpoint files and classes to use the lab name

pmilliotte:
## Description

Renames the Google Gemini endpoint files (`google_ai_studio_gemini_*` → `google_gemini_*`) and classes (`GoogleAiStudioGemini*` → `GoogleGemini*`) so the leading segment is the *lab* (`google`) rather than the `google_ai_studio` *host* — Google models can also be served with `agent_platform` as host, which made the old names read as a host/lab mismatch (e.g. `GoogleAiStudioGeminiThreeDotSixFlashGlobalAgentPlatformStream`).

## Tests

Not needed — rename only. Endpoint ids are unaffected (`buildId()` composes them from the `lab`/`model`/`region`/`host` statics, not the class name).

## Risk

None.

## Deploy Plan

- Deploy `main` on `front`


aubin-tchoi (approved):
LGTM

PR #29532, opened 2026-07-27
Add user-scoped filesystem to DustFileSystem

Inesvvv:
## Description

Add a new filesystem scope, `user-<userId>`to `DustFileSystem` 
GCS backed with one scope per (workspace, user) and accessible only by the authenticated user
Groundwork for User memory project, the scope will later hold a per user `memory.md` file.

## Tests

Added tests for the `forUser`factory and `fromScopedPath` routing 

## Risk

Low

## Deploy Plan

Deploy front

### DIFF
--- a/front/lib/api/file_system/dust_file_system.ts
+++ b/front/lib/api/file_system/dust_file_system.ts
@@ -1,13 +1,15 @@
 /**
  * DustFileSystem is the single entry point for all file system operations in the Dust platform.
  *
- * Scoped path: the agent/API-visible path format, e.g. `conversation-{cId}/report.pdf` or
- * `pod-{pId}/data.csv`. Every public method accepts and returns scoped paths.
- * Legacy paths (`conversation/...`, `project/...`) are accepted for backward compat.
+ * Scoped path: the agent/API-visible path format, e.g. `conversation-{cId}/report.pdf`,
+ * `pod-{pId}/data.csv`, or `user-{uId}/memory.md`. Every public method accepts and returns
+ * scoped paths. Legacy paths (`conversation/...`, `project/...`) are accepted for backward compat.
  *
  * Factories:
  *   DustFileSystem.forConversation(auth, conversation)   single conversation mount (+pod if project space)
  *   DustFileSystem.forConversations(auth, conversations) multiple conversation mounts (+pod if project space)
+ *   DustFileSystem.forPod(auth, space)                   single pod mount
+ *   DustFileSystem.forUser(auth)                         the authenticated user's own memory scope
  *   DustFileSystem.fromScopedPath(auth, scopedPath)     infers context from the path prefix
  *   DustFileSystem.forAgentLoop(auth, { conversation, scopedPaths })
  *       defaults to the agent loop conversation (+ its pod when applicable), plus any
@@ -27,6 +29,7 @@ import {
   LEGACY_PREFIX_PROJECT,
   SCOPED_PREFIX_CONVERSATION,
   SCOPED_PREFIX_POD,
+  SCOPED_PREFIX_USER,
 } from "@app/lib/api/file_system/types";
 import type { SandboxImage } from "@app/lib/api/sandbox/image/sandbox_image";
 import type { Authenticator } from "@app/lib/auth";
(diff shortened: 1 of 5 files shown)


front/lib/api/file_system/dust_file_system.ts
```diff
@@ -130,6 +140,21 @@ function createPodMount(
   };
 }
 
+function createUserMount(userId: string): FileSystemMount {
+  return {
+    kind: "user",
+    id: userId,
+    scopedPrefix: `${SCOPED_PREFIX_USER}${userId}`,
+    sandboxMountPoint: `/files/${SCOPED_PREFIX_USER}${userId}`,
```
flvndvd: We definitely don't want to mount in sandbox for now, so we should reflect this here.

front/lib/api/file_system/dust_file_system.ts
```diff
@@ -151,6 +176,9 @@ function collectScopedPrefixesFromPaths(scopedPaths: string[]): {
         podIds.add(parsed.id);
         break;
 
+      case "user":
```
flvndvd, line 182: What's the expected behavior here? We should avoid an empty case here, either we handle it and support it, or we don't 🙃.

front/lib/api/file_system/dust_file_system.ts
```diff
@@ -292,6 +320,37 @@ export class DustFileSystem {
...
+   * Build a DustFileSystem scoped to a single user's own memory space.
+   *
+   * The mount is always the authenticated user's own scope: `userId` must match
+   * `auth.user()`. This prevents an agent (or a crafted `user-<id>` path resolved via
+   * {@link fromScopedPath}) from reading or writing another user's files.
+   * Returns `Err("unauthorized")` when there is no authenticated user or the ids differ.
+   */
+  static async forUser(
+    auth: Authenticator,
+    userId: string
```
flvndvd: IHMO, the `forUser` should not let you override the actual userId, what about inferring it from the `auth` object directly?

front/lib/api/file_system/dust_file_system.ts
Inesvvv: Added user case to the continue above

front/lib/api/file_system/dust_file_system.ts
Inesvvv: sandbocMountPoint is null now for the user scope, made the field nullable in types.ts

front/lib/api/file_system/dust_file_system.ts
Inesvvv: Done, forUser now infers from auth.user()

front/lib/api/file_system/dust_file_system.ts
```diff
@@ -139,7 +164,7 @@ function collectScopedPrefixesFromPaths(scopedPaths: string[]): {
 
   for (const scopedPath of scopedPaths) {
     const parsed = parseScopedPrefix(scopedPath);
-    if (!parsed) {
+    if (!parsed || parsed.kind === "user") {
```
flvndvd: This is still a very unclear contract from this function. It means that if you pass a user scope here, even though it's valid it will never be returned by this function 🙃. So either we handle it and return it or narrow down the type/assert so it's not supported here.

Inesvvv: Handled the scope, returning just the userId

front/lib/api/file_system/dust_file_system.ts
```diff
@@ -292,13 +324,47 @@ export class DustFileSystem {
...
   /**
    * Build a DustFileSystem for an agent loop.
    *
    * Always includes the agent loop conversation mount and, when applicable, its Pod mount
    * (same as {@link forConversation}). Additionally mounts any conversation or Pod referenced
    * in `scopedPaths` that the caller can access, so cross-scope operations (e.g. copy between
    * two conversations) can resolve both endpoints in a single filesystem instance.
+   *
+   * User-scoped paths are not supported here and trigger an assert, user memory is only
+   * reachable through forUser.
```
flvndvd:
```suggestion
   * User-scoped paths are not supported here and trigger an assert, user filesystem is only
   * reachable through forUser.
```

flvndvd (approved):

PR #29537, opened 2026-07-27
[front] perf: batch agent tag updates

aubin-tchoi:
## Description

The batch tag endpoint currently fetches each agent configuration and writes each agent/tag pair separately, producing N+1 queries as the selection grows.

Not a huge problem at the moment nor one that was surfaced, but it is actually very slow and we're adding ways to select many agents at once and adding more modalities for batch actions, which creates a risk of this pattern being copied in a near future.

This PR batches the 2 operations (fetching the agents and updating the tags).

## Tests

- Added a test for the endpoint.
- Tested locally.

## Risk

- Low. Safe to revert.

## Deploy Plan

- Deploy front.

### DIFF
--- a/front/lib/resources/tags_resource.ts
+++ b/front/lib/resources/tags_resource.ts
@@ -299,7 +299,8 @@ export class TagResource extends BaseResource<TagModel> {
           tagId: tag.id,
           agentConfigurationId: agentConfiguration.id,
         }))
-      )
+      ),
+      { ignoreDuplicates: true }
     );
   }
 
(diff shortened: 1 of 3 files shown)


front/lib/resources/tags_resource.ts
```diff
@@ -290,6 +320,35 @@ export class TagResource extends BaseResource<TagModel> {
...
+    tags: TagResource[],
+    agentConfigurations: LightAgentConfigurationType[]
+  ) {
+    if (
+      !auth.isAdmin() &&
+      agentConfigurations.some(
+        (agentConfiguration) => !agentConfiguration.canEdit
+      )
+    ) {
+      throw new Error("You are not allowed to remove tags from this agent");
```
aubin-tchoi, line 334: not a very good pattern, will clean up error handling in a follow up PR

sfriquet (approved):

PR #29538, opened 2026-07-27
Add: small jitter in cron wake ups and triggers.

Fraggle:
## Description

Adds up to 10 minutes of random jitter to `agentTriggerWorkflow` and `wakeUpWorkflow` when cron expressions fire on the hour (`0 `), preventing thundering-herd spikes when many triggers/wake-ups are scheduled at the same time. Jitter is skipped entirely in development and for webhook-triggered runs.

Introduces `getTriggerActivity` and `getWakeUpActivity` as lightweight pre-run fetch steps so the workflow can inspect the trigger/wake-up config before deciding whether to sleep. Also switches `getAgentConfiguration` to `extra_light` variant in `runTriggeredAgentsActivity` — the full config was never needed there.

## Tests

Tested locally. Verified jitter is 0 in development (`isDevelopment()` guard) and that webhook-triggered workflows skip the fetch and sleep entirely.

## Risk

Low. Jitter only delays on-the-hour cron executions by up to 10 minutes — acceptable given these are background scheduled runs. No DB changes. Rollback is a straight revert.

## Deploy Plan

Deploy front (Temporal workers pick up the new workflow/activity definitions).

### DIFF
--- a/front/temporal/triggers/workflows.ts
+++ b/front/temporal/triggers/workflows.ts
@@ -1,26 +1,38 @@
+import { isDevelopment } from "@app/types/shared/env";
 import { ActivityFailure, RetryState } from "@temporalio/common";
-import { proxyActivities } from "@temporalio/workflow";
-
+import { proxyActivities, sleep } from "@temporalio/workflow";
 import type * as activities from "./activities";
 
-const { runTriggeredAgentsActivity } = proxyActivities<typeof activities>({
+// We spread the start time of triggers and wake-ups starting on the hour over 10 minutes to avoid all conversations starting at the same time.
+const ON_THE_HOUR_JITTER_MS = 60 * 10 * 1000;
+
+const getJitterMs = () => {
+  if (isDevelopment()) {
+    return 0;
+  } else {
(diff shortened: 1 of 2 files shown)


Fraggle:

- #29538 ◀

front/temporal/triggers/workflows.ts
```diff
@@ -1,27 +1,38 @@
...
+import { proxyActivities, sleep } from "@temporalio/workflow";
 import type * as activities from "./activities";
 
-const { runTriggeredAgentsActivity } = proxyActivities<typeof activities>({
-  startToCloseTimeout: "5 minutes",
-  retry: {
-    nonRetryableErrorTypes: ["TriggerNonRetryableError"],
-  },
-});
+const ON_THE_HOUR_JITTER_MS = 60 * 10 * 1000;
```
ElPicador, line 7: I would add a bit of comments here explaining we are jittering on 10 minutes

ElPicador (approved):

front/temporal/triggers/workflows.ts
Fraggle: done

PR #29544, opened 2026-07-27
Fix outdated connectors init_db command in init_dev_container.sh

### LINKED ISSUE #29543: init_dev_container.sh prints outdated connectors init_db.sh command
## Bug

The setup instructions printed at the end of `init_dev_container.sh` include:

```
cd connectors && ./admin/init_db.sh --unsafe && cd -
```

`connectors/admin/init_db.sh` was deleted in 55ab44a ("chore(connectors): remove deprecated initdb tooling") in favor of the phased migration system (`npm run migration:apply`). Running the printed command fails with `zsh: no such file or directory: ./admin/init_db.sh`.

Worse, since the missing script errors before the trailing `&& cd -` runs, the shell is left inside `connectors/`. The next printed instruction (`cd core && cargo run --bin init_db && cd -`) then also fails, since it runs relative to `connectors/` instead of the repo root, producing `cd: no such file or directory: core`.

## Fix

Update the printed instruction to `cd connectors && npm run migration:apply && cd -`.

avervaet:
## Summary
- `connectors/admin/init_db.sh` was removed in 55ab44a in favor of the phased migration system, but `init_dev_container.sh` still printed it as a setup instruction.
- Running the stale command fails and leaves the shell stuck in `connectors/`, which then breaks the following `cd core` step too.
- Updated the printed instruction to `cd connectors && npm run migration:apply && cd -`.

## Test plan
- [x] Ran the full init flow manually (`npm run migration:apply` from `connectors/`, then `cargo run --bin init_db` from `core/`) — both succeeded.

Closes #29543


ElPicador (approved):

ElPicador:
Thanks!

PR #29552, opened 2026-07-27
Add Kimi K3 on Fireworks as the main Kimi model

fontanierh:
## Description

Adds `accounts/fireworks/models/kimi-k3` and makes it our main Kimi.

- Model config, pricing, tier map, reasoning efforts, `model_constructors` and `llms` stream endpoints, SDK types, model picker config, marketing.
- `isLatest` moves from K2.6 to K3, so the agent builder recommends it in the Fireworks group.
- `dust-kimi`, `dust-kimi-medium` and `dust-kimi-high` global agents now point at K3, same as the `dust-ant*` series moved to Opus 5 in #29501.
- No feature flag, unlike K2.6 / K2.5 / MiniMax / GLM-5, which sit behind `fireworks_new_model_feature`.

Specs and pricing from the [Fireworks model page](https://fireworks.ai/models/fireworks/kimi-k3) and [Moonshot's K3 docs](https://platform.kimi.ai/docs/guide/kimi-k3-quickstart): 1040k native context, function calling, image input, JSON-schema structured output, and $3 / $0.30 / $15 per M for input / cached input / output on the Standard serverless tier. Context is capped at our usual 256k and output at 64k, same as Opus 5 and GPT 5.6, leaving a 192k prompt budget. Tiering mirrors `gemini-3.1-pro` ($4/$18, the closest neighbour): balanced up to light, premium for medium and high.

K3 has thinking permanently enabled, so there is no `none` tier: `light` reaches Fireworks as `low` via `useNativeLightReasoning`, then `medium`/`high` pass straight through. Native light reasoning also suppresses the chain-of-thought meta prompt, which would otherwise leak `<thinking>`/`<response>` tags now that light means real thinking.

US only (`us-central1`), like every other Fireworks-served model.

## Tests

Added `front/lib/model_constructors/providers/fireworks/models/kimi_k3.test.ts`, locking the native spec, the Dust caps, the 192k prompt budget and the reasoning-effort map together so a future bump cannot drift one without the others.

`sdk_drift`, `types` and `tiers` guard tests pass, `tsgo` clean on `front` and `marketing`. The endpoint parity test is added and registered but stays off by default like the other Fireworks ones (needs `RUN_LLM_TEST` plus a key).

**Still needs a live endpoint run before merge.** Moonshot documents K3 as `low`/`high`/`max`, while [Fireworks documents](https://docs.fireworks.ai/guides/reasoning) `low`/`medium`/`high` across the models it serves. I followed Fireworks since Fireworks is the host, but if `medium` is rejected then `configSchema` needs narrowing and `dust-kimi-medium` is what breaks. There is a TODO on the config mixin saying so.

## Risk

Low and mostly additive, but two things to know:

- No feature flag means K3 is in the model picker for every workspace on day one, at ~3.3x the per-turn cost of K2.6 ($0.126 vs $0.038 on a 40k-input / 4k-output turn). The tier map keeps medium/high in `premium`, so plan gating still applies.
- `europe-west1: false`, so `regionalModelsOnly` workspaces skip it and `dust-kimi*` falls back to `getLargeWhitelistedModel` there. Same as every other Fireworks model.

The `dust-kimi*` repointing touches internal agents only and is a one-line revert. K2.6 stays registered and selectable, it just loses `isLatest`.

One thing for whoever changes the serving tier later: the pricing here is the Standard serverless tier. Fireworks also sells Priority (`service_tier: "priority"`, +25%) and Fast (a separate `accounts/fireworks/routers/*-fast` model id, +50%). Switching tier means moving `tokenPricing` on the endpoint **and** `CURRENT_MODEL_PRICING` in `token_pricing/global.ts` together, otherwise we silently under-bill by 25-50%.

## Deploy Plan

- Deploy `main` on `front`


front/lib/model_constructors/providers/fireworks/models/kimi_k3.ts
```diff
@@ -0,0 +1,47 @@
...
+// on the legacy config.
+// TODO(2026-07-27 henry): Moonshot's own platform documents K3 as low/high/max
+// (https://platform.kimi.ai/docs/guide/kimi-k3-quickstart), while Fireworks
+// documents low/medium/high across the models it serves
+// (https://docs.fireworks.ai/guides/reasoning). We follow Fireworks here since
+// Fireworks is the host; the live endpoint test must confirm `medium` before
+// merge.
+const configSchema = fireworksConfigSchema.extend({
+  reasoning: z
+    .object({ effort: z.enum(["low", "medium", "high"]) })
```
pmilliotte, line 26: Not sure about this but let's see

pmilliotte (approved):

PR #29560, opened 2026-07-27
[front-api] Update CODING_RULES to reflect the completed Next.js migration

davidebbo:
## Description

`front-api/CODING_RULES.md` still described the service as a strangler in
progress — `server.ts` dispatching via `isHonoRoute` and falling through to
Next for unmatched requests. That is no longer the case: `front` has no `next`
dependency and no route handlers, `front.Dockerfile` has no Next runtime stage,
and `server.ts` simply boots `honoApp` on `@hono/node-server`. `isHonoRoute`
exists nowhere in the repo except in this doc.

Doc-only change. What it fixes:

- **Intro** — describes the current split: `front-api` serves all HTTP traffic,
  `front` is a library workspace (`lib`, `types`, `logger`, `components`) plus
  the Temporal workers and migrations, and the dependency is one-way.
- **API1** — drops the `pages/api/<path>.ts` framing; the `v1` example now
  matches the actual layout (`v1/w/[wId]/spaces/index.ts`, not `spaces.ts`).
- **API3** — the section explained Hono-vs-Next dispatch. Replaced with the
  rule that still holds (mounting is the whole registration step) plus the
  root-level mount-order exceptions already commented in `app.ts`.
- **API6 / API7** — `middleware/` → `middlewares/` (the real directory name),
  and drops the "Hono counterpart of `apiError(req, res, ...)` in
  `front/logger/withlogging.ts`" reference — that function no longer exists.
- **API9** — `NextApiRequest` / `NextApiResponse` are unresolvable now (`next`
  is not a dependency of `front` or `front-api`) and `getSession(req, res)` /
  `getWorkOSSession(req, res)` no longer exist. Kept the two rules that still
  apply: read the request through Hono's `Context`, and resolve sessions via
  `getWorkOSSessionWithSetCookies`, pointing at
  `middlewares/session_resolution.ts` as the live example.
- Shared rule families referenced by name instead of a stale `GEN1-GEN13,
  SEC1-SEC2` range (root is now at GEN14 / SEC4).

## Tests

None — markdown only, no code touched. Every path, symbol, and behavior now
asserted in the doc was verified against the current tree.

## Risk

None. No runtime or build impact.

## Deploy Plan

Nothing to do — merge.

### DIFF
--- a/front-api/CODING_RULES.md
+++ b/front-api/CODING_RULES.md
@@ -1,16 +1,21 @@
 # [front-api] Coding Rules
 
-Shared rules (GEN1-GEN13, SEC1-SEC2, ERR1-ERR2) in the root `CODING_RULES.md`
-apply automatically. TypeScript rules in `front/CODING_RULES.md` also apply,
+Shared rules (GEN, SEC and ERR) in the root `CODING_RULES.md` apply
+automatically. TypeScript rules in `front/CODING_RULES.md` also apply,
 since `front-api` shares the same TS environment and reaches into `front`
 internals via the `@app/*` path map.
 
 This file documents rules and patterns specific to `front-api`.
 
-`front-api` is the Hono-based service that incrementally takes over routes
-from `front` (Next.js). The strangler dispatch lives in
-`front-api/server.ts`: it checks `isHonoRoute` from `front-api/app.ts`;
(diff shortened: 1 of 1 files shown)


id13:
LGTM

PR #29562, opened 2026-07-27
Preserve tool result names for Fireworks models

fontanierh:
## Description

Kimi K3 rejects parallel tool-result replay when `tool` messages omit the tool name. Preserve the known name in the shared Fireworks request conversion so every Fireworks model receives unambiguous tool results without changing other OpenAI-compatible hosts.

## Tests

- Added a Fireworks client payload regression covering two parallel tool results.
- Confirmed every registered Fireworks endpoint builds named tool results.
- Sent the repaired replay to Fireworks successfully with Kimi K3, GLM 5.2, and DeepSeek V4 Pro.

## Risk

Low. This adds a provider-supported field to Fireworks tool-result messages; other hosts are unchanged and rollback is safe.

## Deploy Plan

Standard deploy.


front/lib/model_constructors/stream/clients/fireworks.test.ts
```diff
@@ -0,0 +1,73 @@
+// @vitest-environment node
+
+import { ZAiGlmFiveDotTwoGlobalFireworksStream } from "@app/lib/model_constructors/stream/endpoints/z_ai_glm_five_dot_two_global_fireworks";
+import { describe, expect, it } from "vitest";
+
+describe("FireworksStream", () => {
+  it("includes names when replaying parallel tool results", () => {
+    const endpoint = new ZAiGlmFiveDotTwoGlobalFireworksStream({
```
pmilliotte, line 8: So the issue was not specifically for Kimi K3 but glm too ?

pmilliotte (commented):
LGTM

front/lib/model_constructors/stream/clients/fireworks.test.ts
fontanierh: Fireworks in general afaict

PR #29564, opened 2026-07-28
[poke] Restrict member seat dropdown to current contract seats

tdraier:
## Description

Fixes [tasks#9789](https://github.com/dust-tt/tasks/issues/9789).

In Poke's members list, the dropdown to switch a member's seat offered **all** seat types. It now restricts to the seats entitled by the workspace's **current contract**, mirroring how `SwitchContractDialog` restricts to a package's seats.

## Approach

The current contract's entitled seats are already exposed via `usePokeWorkspaceInfo()` → `seatPlan` (a `Partial<Record<MembershipSeatType, SeatTypeInfo>>` built from `getSeatPlan` off the active Metronome contract). Its keys are exactly the billable seats on the contract, so those are threaded down into the seat dropdown. No backend/API change — the existing `workspace-info` endpoint is reused.

## Changes (front-end only)

- **`MembershipsPage.tsx`** — fetch `usePokeWorkspaceInfo`, derive `availableSeatTypes` = `"none"` + billable seats present in `seatPlan`. Falls back to `undefined` (all seats) when there is no seat plan (e.g. non-Metronome contracts).
- **`members/table.tsx`** — new optional `availableSeatTypes` prop, forwarded to the columns.
- **`members/columns.tsx`** — dropdown uses `availableSeatTypes` (defaulting to all seat types when absent), and keeps a member's current seat selectable even if the contract no longer offers it, so the `<select>` value stays accurate.

## Notes

- "none" (unassign) is always available regardless of contract.
- Other `MembersDataTable` callers (Space/Project/Group pages) are unchanged — they don't pass `availableSeatTypes`, so behavior there is identical to today.

### DIFF
--- a/front/components/poke/members/columns.tsx
+++ b/front/components/poke/members/columns.tsx
@@ -26,11 +26,16 @@ export type MemberDisplayType = {
 };
 
 export function makeColumnsForMembers({
+  availableSeatTypes,
   onRevokeMember,
   onUpdateMemberRole,
   onUpdateMemberSeatType,
   readonly,
 }: {
+  // Seat types selectable in the dropdown. When undefined, all seat types are
+  // offered; when provided (e.g. restricted to the current contract's seats),
+  // only these are selectable.
+  availableSeatTypes?: readonly MembershipSeatType[];
   onRevokeMember: (m: MemberDisplayType) => Promise<void>;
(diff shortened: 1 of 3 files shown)


fabiencelier (approved):
LGTM

PR #29565, opened 2026-07-28
[poke] Unify member mutations on the batch-update-members plugin

tdraier:
## Description

Unifies **every** member mutation in the Poke members list — per-row role change, per-row seat change, per-row revoke, and the bulk toolbar — onto a single poke plugin, so they share one code path, one permission gate (`requiredRoles: ["support"]`), and one run history. The bespoke poke endpoints are deleted.

## Changes

**One plugin: `batch-update-members`**
- Generalized the former `batch-update-seats` plugin: an `Action` selector (*Update seat type / Update role / Revoke*) + `role`, `seatType`, `emails`, `immediate`.
- Fans out server-side (`fetchByEmails` → `concurrentExecutor(10)` → per-member result) over the existing business functions. Keeps the seat-contract entitlement guard. The role path uses `allowTerminated`, so it also reactivates revoked members.

**Shared client hook** — `useBatchUpdateMembers` wraps `useRunPokePlugin`; both the per-row handlers (single email) and the bulk toolbar (N emails) use it.

**Row selection + bulk toolbar** — `PokeDataTable` gains opt-in `enableRowSelection` / `getRowId` / `renderBulkActions`; the header checkbox selects all rows matching the current filters across pages.

**Deleted endpoints** — `POST /api/poke/workspaces/:wId/{roles,seat_type,revoke}` (route files + mounts); their only caller was this table.

## Audit logging

Per-member events are emitted from `lib/api/membership` so **all** callers (poke, Temporal, scripts, checkout) get them:
- `membership.role_updated` (was previously only emitted by the deleted `roles` endpoint)
- `membership.seat_updated` **(new)** — fires on real seat changes/schedules, skips noops
- `membership.revoked` (already emitted by `revokeAndTrackMembership`)

The plugin emits one bulk-summary event per batch:
- `membership.bulk_role_updated` **(new)**
- `membership.bulk_seat_updated` **(new)**
- `member.bulk_revoked` (existing)

⚠️ **Before merge:** the 3 new audit actions (`membership.seat_updated`, `membership.bulk_role_updated`, `membership.bulk_seat_updated`) must be registered with WorkOS and the version map committed:
```
npx tsx front/admin/register_audit_log_schemas.ts --execute --changed
# then commit the regenerated front/lib/api/audit/schema_versions.json
```
CI (`audit_log_schemas.test.ts`) will fail until this lands. (Owner will run the registration.)

## Notes
- Read-only member tables (Space/Project/Group pages) are unaffected.


tdraier:

- #29564
    - #29565 ◀
        - #29573

front/components/poke/members/MembersBulkActions.tsx
```diff
@@ -0,0 +1,169 @@
...
+    }
+
+    setIsSubmitting(true);
+    const failures: string[] = [];
+    // Sequential on purpose: membership mutations have per-user billing and
+    // guard side effects (last-admin, seat limits), so applying them one at a
+    // time keeps the outcome predictable and avoids hammering the API.
+    for (const m of eligible) {
+      try {
+        const r = await request(m);
```
tdraier: In sequence on purpose. Promise.all would be too aggressive, alternative would be a batch endpoint but goal is to keep it simple here

front/components/poke/members/MembersBulkActions.tsx
```diff
@@ -0,0 +1,169 @@
+import type { MemberDisplayType } from "@app/components/poke/members/columns";
+import { clientFetch } from "@app/lib/egress/client";
```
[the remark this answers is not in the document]
tdraier: already the case in table.tsx, not refactoring here

front-api/routes/poke/workspaces/[wId]/roles.ts
```diff
@@ -1,105 +0,0 @@
...
-              message: "Cannot remove the last admin of a workspace.",
-            },
-          });
-        default:
-          assertNever(updateRes.error.type);
-      }
-    }
-
-    if (updateRes.isOk()) {
-      void emitAuditLogEvent({
```
Comment (aubin-tchoi) [inline front-api/routes/poke/workspaces/[wId]/roles.ts:83]: I think we lost this audit log

aubin-tchoi (approved):
LGTM

front-api/routes/poke/workspaces/[wId]/roles.ts
Comment (tdraier) [reply front-api/routes/poke/workspaces/[wId]/roles.ts:83]: right, weird to have audit logs in poke ops, but I will restore the ones we may have lost

Comment (aubin-tchoi) [reply front-api/routes/poke/workspaces/[wId]/roles.ts:83]: yes maybe we should remove it?

Comment (tdraier) [reply front-api/routes/poke/workspaces/[wId]/roles.ts:83]: there are actually a lot of audit events in poke. so seems "standard". still legit that the admin can see we update their users.

PR #29566, opened 2026-07-28
[LLM] Characterize endpoint input configs against live APIs and official docs

pmilliotte:
## Description

Corrects the `model_constructors` input config schemas for all 30 LLM endpoints so they match what the providers actually accept — each was characterized by running the endpoint suite with the widest `inputConfigSchema` and cross-checked against the model author's documentation. Product-side coercions move out of the schemas into `configParsers` in the llms layer, and the `dust-llm` skill now encodes the process so new models get the same treatment.

Notable fixes:
- Gemini `none` did not disable thinking — it sent `thinkingBudget: 128` with thoughts hidden, so 3.5-flash still billed ~158 reasoning tokens while reporting none. Now `thinkingBudget: 0`, exposed only on the two models that accept it.
- Mistral Small was wired to the non-reasoning schema, which rejected the `none` the llms layer sends on every request.
- The extended-thinking "no forced tool with thinking" restriction was applied to six adaptive-thinking Anthropic models where it does not exist; it is only real on Haiku 4.5.
- `temperature: z.undefined()` on 20+ models rejected values the APIs accept, and the shared 0..1 range silently truncated OpenAI (0..2), Gemini (0..2) and Mistral (0..1.5).
- Reasoning efforts now follow the model author's docs rather than what the host gateway tolerates — Fireworks validates `low/medium/high/xhigh/max/none` for every model it serves, so passing live runs proved nothing.
- Removes the Fable 5 EU endpoint: it needs the 30-day data retention only our EAP account has.

## Tests

Every endpoint suite run green against the live APIs: Anthropic 14 endpoints, OpenAI 12 (global), Gemini 10 (global, AI Studio + Vertex), Mistral 4, Fireworks 4, xAI 1. Anthropic converter unit tests 72/72. All 63 suites load and register with the live gate off.

Not verified: the 12 OpenAI EU endpoints and 3 Gemini EU endpoints (treated as identical to global, they share the same mixins), and `kimi-k2.5`, which returns 404 `Model not found, inaccessible, and/or not deployed` on our Fireworks account and is left untouched.

## Risk

Reasoning effort sets narrowed on several models; `configParsers` preserve the existing product behavior, so no user-facing capability changes. Two deliberate exceptions: Gemini 3.5-flash-lite loses `none` while 3.5-flash gains it, and Kimi K3's default moves from `low` to the documented `max`, which raises cost and latency on its default path.

## Deploy Plan

- Deploy `main` on `front`


aubin-tchoi (approved):
LGTM

PR #29574, opened 2026-07-28
Limit retained Anthropic tool image previews

fontanierh:
## Description

- Add a model-level `maxInputImages` capability, set to 20 for Anthropic models.
- Before tokenization, replace the oldest tool-result image previews above that limit with explanatory text and a reusable file path when available.
- Keep user-uploaded images intact, warn when 20 or more non-tool images cannot be pruned, and emit image-pruning metrics.

## Tests

Added conversation-rendering coverage for oldest-first replacement, path recovery instructions, user-image preservation, warning behavior, token accounting, and canonical file-path propagation.

## Risk

Low. Only models that declare `maxInputImages` are affected. Anthropic is currently the only configured provider.

## Deploy Plan

Standard deploy.

### DIFF
--- a/front/lib/api/assistant/conversation_rendering/checkpointed_window_state.ts
+++ b/front/lib/api/assistant/conversation_rendering/checkpointed_window_state.ts
@@ -1,16 +1,11 @@
-import { getPrefixedToolName } from "@app/lib/actions/tool_name_utils";
-import {
-  FILES_CAT_ACTION_NAME,
-  FILES_SERVER_NAME,
-} from "@app/lib/api/actions/servers/files/metadata";
 import type {
   InteractionWithTokens,
   MessageWithTokens,
 } from "@app/lib/api/assistant/conversation_rendering/pruning";
 import {
   getToolResultTokenSavings,
-  IMAGE_CONTENT_TOKEN_COUNT,
   PRUNING_CHECKPOINT_TOKENS,
+  pruneToolResultImagePreview,
   pruneToolResultMessage,
 } from "@app/lib/api/assistant/conversation_rendering/pruning";
 import type {
@@ -18,7 +13,6 @@ import type {
   ConversationWindowResult,
 } from "@app/lib/api/assistant/conversation_rendering/window_types";
 import logger from "@app/logger/logger";
-import type { ImageContent } from "@app/types/assistant/generation";
 import { isImageContent } from "@app/types/assistant/generation";
 import type { Result } from "@app/types/shared/result";
 import { Ok } from "@app/types/shared/result";
@@ -33,7 +27,6 @@ type ToolResultNode = {
   message: Extract<MessageWithTokens, { role: "function" }>;
   tokenSavings: number;
   pruned: boolean;
-  eligible: boolean;
   imageReferences: ToolImageReference[];
 };
 
@@ -60,14 +53,8 @@ export const MINIMUM_PRUNING_BATCH_TOKENS = 5_000;
 type ToolImageReference = {
   node: ToolResultNode;
   contentIndex: number;
-  image: ImageContent;
   retained: boolean;
 };
-
-const FILES_CAT_TOOL_NAME = getPrefixedToolName(
-  FILES_SERVER_NAME,
-  FILES_CAT_ACTION_NAME
-);
 
 function makeWindowMessageNode(message: MessageWithTokens): WindowMessageNode {
   if (message.role === "function") {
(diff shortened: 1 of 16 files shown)


front/lib/api/assistant/conversation_rendering/index.ts
```diff
@@ -80,6 +82,73 @@ function pruneConversationToBudget(
   return state.fit();
 }
 
+function replaceOldestToolResultImages(
```
flvndvd: I moved pruning to https://github.com/dust-tt/dust/blob/main/front/lib/api/assistant/conversation_rendering/checkpointed_window_state.ts a few weeks ago. Sole responsibility is to own an instrument pruning as it has a direct cost/quality impact. IMHO a better place would be in this new class. We could make this agnostic maybe by adding a config on the model and leveraging it directly in there, WDYT?

fontanierh: Let me try 👍

fontanierh: Done in 43988de06d. `maxInputImages` now lives on model config, with 20 configured for Anthropic. The provider-agnostic pre-tokenization entry point is owned by `CheckpointedConversationWindowState` and returns image-pruning stats for Datadog. It stays before tokenization so the replacement text is included in context accounting.

front/lib/api/assistant/conversation_rendering/index.ts
```diff
@@ -135,7 +135,20 @@ export async function renderConversationForModel(
     agentConfiguration,
     enabledSkills,
   });
-  const messages = [...leadingMessages, ...renderedMessages];
+  // Apply model input limits before tokenization so replacement text is counted.
+  const { messages, stats: imagePruningStats } =
+    CheckpointedConversationWindowState.pruneOldestToolResultImages(
```
flvndvd: The way this class was designed was meant for sequential call, replaying the full conversation timeline. Assuming we want to follow this path (so we can easily unlock bunch of stuff down the line). Can we ensure that the class itself checks this on every new interactions added.

Example. like we do for pruning, for each interaction added we increment the image count, as soon as it goes above model limit we prune?

fontanierh: Addressed in 853025b6df. The render now replays raw interactions through `ConversationImagePruningState`; `append()` increments image counts and prunes the oldest tool-result previews as soon as the configured limit is crossed. Its result is tokenized afterward, so the replacement text is included in token accounting before the existing token-window state runs.

fontanierh: Reworked in 1af7ac8c61. There is no separate image-pruning state or pre-tokenization pass now. `CheckpointedConversationWindowState.append()` tracks retained images while replaying the tokenized interactions, replaces the oldest tool-result preview as soon as the model limit is crossed, and updates the node and queue token accounting in place. Whole-tool-result pruning also releases the image references it removes, so both policies share one coherent state.

fontanierh (commented):
Two points I think we should address before merging.

front/lib/api/assistant/conversation_rendering/pruning.ts
```diff
@@ -32,6 +46,41 @@ export function pruneToolResultMessage(
...
+  };
+  content[contentIndex] = replacement;
+
+  return {
+    ...message,
+    content,
+    tokenCount: Math.max(
+      message.tokenCount -
+        IMAGE_CONTENT_TOKEN_COUNT +
+        Buffer.byteLength(replacement.text),
```
flvndvd, line 78: We assume that one token = one byte here?

front/lib/api/assistant/conversation_rendering/checkpointed_window_state.ts
```diff
@@ -73,6 +86,10 @@ function makeWindowMessageNode(message: MessageWithTokens): WindowMessageNode {
  * If tool-result pruning cannot keep the complete interaction history below the nominal budget,
  * the window keeps serving it and reports the excess through logs and metrics. The provider limit
  * remains the final boundary. The latest unconsumed result batch always remains intact.
+ *
+ * When the model declares an image input limit, image previews are registered during the replay
+ * and the cap is applied once at fit() time: the oldest tool-result previews are replaced with
```
flvndvd, line 91: cap seems to be applied on every call to `append()` 🤔

flvndvd (commented):
I see why both you and the model were confused, some stuff that I'm not sure about:
- It only pruned tool-result images, so nothing prevents to exceed this limit with 20+ user images. Does the provider limit apply only on tool results?
- The cap is called on each append

I tried to get to a simpler version that can be used in `CheckpointedConversationWindowState` on [flav/cap-model-input-images](https://github.com/dust-tt/dust/commit/6b085f8b1331899be76ca854a55deed03da7b8b2). It should be plug and play.

PR #29575, opened 2026-07-28
[front] Add Redis fixed-window counter to the rate limiter

tdraier:
## Description

Adds a window-agnostic **fixed-window Redis counter** to the rate limiter, alongside the existing rolling (`ZADD`/`ZCOUNT`) limiter:

- `addFixedWindowCount` — additive `INCRBY` + `PEXPIREAT`, for recording a cost that already happened (enforcement happens beforehand via a read).
- `getFixedWindowCount` — reads the current count (0 when the window has no entries yet).
- Both operate over explicit `FixedWindowBounds` (`{ label, windowEndMs }`) supplied by the caller, so the primitive stays agnostic of window semantics — the label distinguishes each window as its own key that expires at `windowEndMs`, and callers decide how bounds are computed (calendar math, a billing-cycle anchor, etc.).

Foundation only — nothing calls these yet. The first consumer is the per-user spend-cap backup (#29576, stacked on this).

## Tests

- `rate_limiter.test.ts` — unit tests for `addFixedWindowCount` / `getFixedWindowCount` (increment, expiry bounds, empty-window read).

## Risk

Minimal. Purely additive: two new functions and a type, no existing code path changed, nothing wired to callers in this PR. Safe to roll back by revert.

## Deploy Plan

Standard deploy. No migration, no config, no feature flag.

### DIFF
--- a/front/lib/utils/rate_limiter.ts
+++ b/front/lib/utils/rate_limiter.ts
@@ -15,6 +15,19 @@ export class RateLimitError extends Error {}
 
 export const RATE_LIMITER_PREFIX = "rate_limiter";
 
+// Grace period kept after the window boundary before the Redis key expires, so
+// a read straddling the boundary still sees a just-closed window rather than a
+// premature miss.
+const FIXED_WINDOW_EXPIRE_GRACE_MS = 60_000;
+
+// A resolved fixed window: a stable label identifying the current window and
+// the absolute UTC end of that window. The label is appended to the Redis key
+// so each window is a distinct key that naturally expires; `windowEndMs` drives
+// `PEXPIREAT`. Callers resolve these bounds however they like — pure calendar
+// math or an external anchor such as a billing contract — keeping this counter
+// agnostic of window semantics.
+export type FixedWindowBounds = { label: string; windowEndMs: number };
+
 const makeRateLimiterKey = (key: string) => `${RATE_LIMITER_PREFIX}:${key}`;
 
 type RateLimiterArgs = {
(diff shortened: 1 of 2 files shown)


tdraier:

- #29575 ◀
    - #29576
        - #29640
    - #29617

front/lib/utils/rate_limiter.ts
```diff
@@ -242,6 +245,143 @@ export function getTimeframeSecondsFromLiteral(
...
+    }
+    case "calendar_month": {
+      const label = `${year}-${pad(month + 1)}`;
+      const windowEndMs = Date.UTC(year, month + 1, 1);
+      return { label, windowEndMs };
+    }
+    default:
+      assertNever(timeframe);
+  }
+}
```
tdraier: as seen irl this method ( and the calendar_day / calendarweek / calendar_month ) are not used here, but is mainly for testing - in https://github.com/dust-tt/dust/pull/29576 we directly calculate the FixedWindowBounds based on billing period - and we'll probably never use calendar / utc based.

front/lib/utils/rate_limiter.ts
```diff
@@ -245,3 +245,108 @@ export function getTimeframeSecondsFromLiteral(
...
   }
 }
+
+// A resolved fixed window: a stable label identifying the current window and
+// the absolute UTC end of that window. The label is appended to the Redis key
+// so each window is a distinct key that naturally expires; `windowEndMs` drives
+// `PEXPIREAT`. Callers resolve these bounds however they like — pure calendar
+// math or an external anchor such as a billing contract — keeping this counter
+// agnostic of window semantics.
+export type FixedWindowBounds = { label: string; windowEndMs: number };
```
aubin-tchoi: I think we should rename the functions for the sliding window rate limiter so that there's no confusion

front/lib/utils/rate_limiter.ts
```diff
@@ -245,3 +245,108 @@ export function getTimeframeSecondsFromLiteral(
...
+
+// Grace period kept after the window boundary before the Redis key expires, so
+// a read straddling the boundary still sees a just-closed window rather than a
+// premature miss.
+const FIXED_WINDOW_EXPIRE_GRACE_MS = 60_000;
```
aubin-tchoi: nit: would put at the top of the file (or split in 2 files)

front/lib/utils/rate_limiter.ts
```diff
@@ -245,3 +245,108 @@ export function getTimeframeSecondsFromLiteral(
...
+export async function addFixedWindowCount({
+  key,
+  bounds,
+  incrementBy,
+  logger,
+}: {
+  key: string;
+  bounds: FixedWindowBounds;
+  incrementBy: number;
+  logger: LoggerInterface;
```
aubin-tchoi, line 278: nit:
```suggestion
  logger: Logger;
```
(imported directly from pino)

aubin-tchoi (approved):
LGTM

front/lib/utils/rate_limiter.ts
tdraier: it would be slightly bigger pr but more clear indeed

tdraier: would that in a follow up, that's ~34 files to change

front/lib/utils/rate_limiter.ts
tdraier: actually file is already using loggerInterface everywhere, keeping the same signature for consistency

front/lib/utils/rate_limiter.ts
aubin-tchoi: I can handle if you want btw

PR #29576, opened 2026-07-28
[front] Rate-limiter backup for the per-user spend cap

tdraier:
## Description

Backs up the Metronome per-user spend cap with a **synchronous Redis fixed-window counter**, enforced at message send independently of Metronome. This guards against the case where Metronome's per-user usage lags or is unavailable, so a user can't blow far past their cap before the async signal catches up.

Built on the fixed-window primitive from #29575 (this PR is stacked on it).

- **Counter anchored on the Metronome contract billing cycle.** The window bounds come from `getCachedMetronomeCurrentBillingPeriod`, so the counter resets exactly when the contract cycle does — no separate/configurable period.
- **Records the cost delta, not the total.** On each agent-message finalize, `computeAndStoreAgentMessageCredits` recomputes the message's running total; we add only `costCredits − previousCostCredits` to the additive counter, so repeated finalizes (agent-loop early exit, confirmation, retries) don't double-count. A retry with no new usage yields a 0 delta.
- **Both recording and enforcement are behind one feature flag** (`enforce_user_spend_limit_rate_cap`): when off, the counter neither accrues nor blocks; when on, usage is recorded on finalize and `isUserSpendLimitRateCapReached` blocks the send with `user_cap_reached` once the cap is reached. Reads and writes fail open (a Redis/resolution error never blocks a message).
- **Effective cap resolution** reuses the canonical path: per-user override > max group cap > seat-type/workspace default, each including the seat allowance (`getEffectiveSpendCapAwuCreditsForUser`).
- Mirrors the new flag into the SDK's `WhitelistableFeaturesSchema`.

Admin/observability tooling (poke resync + the counter column) is in the follow-up #29640 — run the resync to seed the counter from usage before enabling the flag, so users aren't measured from 0 mid-cycle.

## Tests

- `tsgo` clean across `front` + `front-api`; SDK types rebuilt for the new flag.
- Existing rate-limiter / spend-limit suites pass.
- Manually in dev: posted messages and confirmed the per-user counter accrues the cost delta once per finalize (no double-count on re-finalize), and that enabling the flag blocks sends once the cap is reached.

## Risk

Low, and fully gated. With `enforce_user_spend_limit_rate_cap` off (the default), this PR is inert — no recording, no enforcement. When on, recording is a best-effort Redis `INCRBY` on the finalize path (a Redis failure is logged and swallowed, never blocking a message) and the enforcement read fails open. Delta recording is idempotent under retries. Safe to roll back by revert.

## Deploy Plan

Standard deploy, no migration. The flag is off everywhere, so nothing changes on deploy. To activate the backup for a workspace: **first** run the resync plugin (#29640) to seed the counter from Elasticsearch usage, **then** enable `enforce_user_spend_limit_rate_cap` — in that order, so recording+enforcement start from the real mid-cycle total rather than 0.

### DIFF
--- a/front/lib/api/users/spend_limit.ts
+++ b/front/lib/api/users/spend_limit.ts
@@ -1,7 +1,12 @@
 import {
+  makeSpendLimitAwuCreditsRateLimitKeyForUser,
+  makeSpendLimitCycleWindowBounds,
+} from "@app/lib/api/assistant/rate_limits";
+import {
   buildAuditLogTarget,
   emitAuditLogEvent,
 } from "@app/lib/api/audit/workos_audit";
+import { getEffectiveSpendCapAwuCreditsForUser } from "@app/lib/api/credits/members_usage";
 import { reconcileUser } from "@app/lib/api/metronome/reconcile_credit_state";
 import { getUserForWorkspace } from "@app/lib/api/user";
 import type { AuditLogContext } from "@app/lib/api/workos/organization";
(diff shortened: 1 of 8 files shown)


tdraier:

- #29575
    - #29576 ◀
        - #29640
            - #29659

front/lib/api/users/spend_limit.ts
```diff
@@ -327,3 +339,104 @@ export async function setUserSpendLimit(
 
   return new Ok({ limit });
 }
+
+// Fixed-window bounds for the current Metronome contract billing cycle (the
+// window the per-user spend cap is bucketed on). `null` when no billing period
+// can be resolved — callers treat that as a no-op (fail-open, matching the rest
+// of the rate-limiter callers).
+async function resolveSpendLimitCycleBounds(
+  workspaceId: string
```
aubin-tchoi: nit: could take the entire workspace in input

aubin-tchoi (approved):
LGTM

fabiencelier (approved):
LGTM ✅

avervaet:
@flvndvd

PR #29581, opened 2026-07-28
[AGENT LOOP] surface empty model output as a retryable error on every step

pmilliotte:
## Description

A step where the model returns no tool call and no text was only an error on the last step; on any other step it published `agent_message_success`, silently ending the run with the previous steps' content (seen in production after an Anthropic turn came back with only a server-side tool search), so both cases now publish a retryable `empty_content` error.

## Tests

Not yet tested at runtime — `runModel` has no test harness; type-checked with `tsgo`.

## Risk

Low.

## Deploy Plan

- Deploy `main` on `front`


front/temporal/agent_loop/lib/run_model.ts
```diff
@@ -1009,30 +1009,41 @@ export async function runModel(
...
+              code: "max_step_reached",
+              message:
+                "This agent took too many steps to answer your query. " +
+                "Try narrowing down your question or breaking it into smaller parts.",
+              metadata: {
+                category: "empty_content",
+                errorTitle: "Too many steps",
+              },
+            }
+          : {
```
flvndvd, line 1037: Do we actually track those somewhere?

flvndvd (approved):

PR #29582, opened 2026-07-28
Fix Gong MCP calls for non-default regions

flvndvd:
## Description

This PR fixes https://github.com/dust-tt/tasks/issues/9819.

Gong OAuth returns a customer-specific API base URL because accounts may be hosted on different cells. PR https://github.com/dust-tt/dust/pull/27851 fixed this for connectors, but the Gong MCP client still used the
global endpoint.

This persists the validated customer base URL in OAuth connection metadata and uses it for MCP requests. The
global endpoint remains the fallback for legacy connections. Existing regional connections must reconnect to
receive the new metadata.


aubin-tchoi (approved):
LGTM

PR #29589, opened 2026-07-28
[front] fix: refresh invitations list after edit/revoke

tdraier:
## Problem

On the workspace **Invitations** tab, editing an invitation's role (admin/member) returned to the list without showing the updated role. Revoking an invitation had the same issue — the list wasn't reliably reloaded.

## Root cause

`InvitationsList` fetches via `useWorkspaceInvitations(owner, { includeExpired: true })`, so its SWR cache key is:

```
/api/w/{sId}/invitations?includeExpired=true
```

But `updateInvitation()` revalidated with a query-param-less key:

```js
await mutate(`/api/w/${owner.sId}/invitations`);
```

SWR's string-key `mutate` does an **exact** match, so it never matched the `?includeExpired=true` key and the list was never revalidated. Because both edit and revoke go through `updateInvitation`, both were affected. The identical latent bug existed in `InviteEmailButtonWithModal` after sending new invites.

## Fix

Added a shared `mutateWorkspaceInvitations(owner)` helper in `front/lib/invitations.ts` that revalidates every invitations key for the workspace regardless of query params (matched by pathname), and used it at both call sites.

## Testing

- `npx tsgo --noEmit` passes
- `npm run format:changed` passes

No tests added: this is client-side SWR wiring, and the repo's testing convention is functional endpoint tests rather than UI unit tests.

### DIFF
--- a/front/components/members/InviteEmailButtonWithModal.tsx
+++ b/front/components/members/InviteEmailButtonWithModal.tsx
@@ -21,6 +21,7 @@ import { getPriceAsString } from "@app/lib/client/subscription";
 import { clientFetch } from "@app/lib/egress/client";
 import {
   MAX_UNCONSUMED_INVITATIONS_PER_WORKSPACE_PER_DAY,
+  mutateWorkspaceInvitations,
   sendInvitations,
 } from "@app/lib/invitations";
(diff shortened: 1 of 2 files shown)


ykmsd (approved):
ty!

PR #29592, opened 2026-07-28
Opt: rewrite listBySpaceIds to avoid the OR and make it index friendly.

Fraggle:
## Description

`DataSourceViewResource.listBySpaceIds` previously filtered via a JOIN on `vaults` with an `Op.or` clause (`id IN (spaceModelIds) OR kind = 'global'`). The OR prevents the DB from using an index on `vaultId`, forcing a full-scan join on every call.

Rewrote to resolve the global space's numeric `vaultId` upfront via `SpaceResource.fetchWorkspaceGlobalSpace(auth)` when `includeGlobalSpace=true`, then issue a single `WHERE vaultId IN (...)` on `data_source_views` — no JOIN needed, fully index-friendly.

## Tests

Updated the existing unit test for `listBySpaceIds`: adjusted the spy assertion to expect `fetchWorkspaceGlobalSpace` to be called (it now is), and removed the now-irrelevant `fetchByIds` spy. Tested locally.

## Risk

Low. Pure query rewrite with identical semantics — same rows returned, no schema or API change. Adds one extra DB read (global space lookup) per call with `includeGlobalSpace=true`, but this is a small indexed lookup that replaces a heavier OR-join.

## Deploy Plan

Deploy front.

### DIFF
--- a/front/lib/resources/data_source_view_resource.test.ts
+++ b/front/lib/resources/data_source_view_resource.test.ts
@@ -147,7 +147,7 @@ describe("DataSourceViewResource", () => {
   });
 
   describe("listBySpaceIds", () => {
-    it("includes global space views without fetching spaces", async () => {
+    it("includes global space views via resolved global vaultId", async () => {
       const workspace = await WorkspaceFactory.basic();
       const adminAuth = await Authenticator.internalAdminForWorkspace(
         workspace.sId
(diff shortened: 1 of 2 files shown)


Fraggle:

- #29592 ◀

aubin-tchoi (commented):
LGTM

PR #29602, opened 2026-07-28
feat: add analytics breakdown per model

sfriquet:
## Description

Update analytics graph so they can be filtered and grouped by model.

Refs https://github.com/dust-tt/tasks/issues/9825

## Tests

Screenshot 2026-07-28 at 14 28 08
Screenshot 2026-07-28 at 14 28 17

## Risk

Low. Safe to rollback.

## Deploy Plan

Front only.

PR #29601, opened 2026-07-28
feat: add model (modelId, providerId) to elastic search message index

sfriquet:
## Description

Add `providerId`, `modelId`, `reasoningEffort` and `resolutionMethod` into the ES message index so we can get analytics per model.

Refs https://github.com/dust-tt/tasks/issues/9825

## Tests

Tested manually that ES index now contain the new object:
```
GET front.agent_message_analytics/_search
{
  "size": 3,
  "sort": [{ "timestamp": "desc" }]
}
```

## Risk

Low. Safe to rollback.

## Deploy Plan

ES mapping should be updated before deploying

PR #29604, opened 2026-07-28
feat: add modelId, providerId and resolutionMethod to analytics message export

sfriquet:
## Description

Add those new fields to the CSV/JSON analytics export.

Refs https://github.com/dust-tt/tasks/issues/9825.

## Tests

ut + manually tested by downloading the `messages` export and seeing the new fields populated

## Risk

Low. Safe to rollback

## Deploy Plan

Front only

### DIFF
--- a/front/lib/api/analytics/messages_export.ts
+++ b/front/lib/api/analytics/messages_export.ts
@@ -8,7 +8,10 @@ import { resolveServerDisplayNames } from "@app/lib/api/assistant/observability/
 import type { ElasticsearchBaseDocument } from "@app/lib/api/elasticsearch";
 import { searchAnalytics } from "@app/lib/api/elasticsearch";
 import type { Authenticator } from "@app/lib/auth";
-import type { AgentMessageAnalyticsCost } from "@app/types/assistant/analytics";
+import type {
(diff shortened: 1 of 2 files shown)


sfriquet:

- #29601
    - #29602
        - #29604 ◀
            - #29605

front/lib/api/analytics/messages_export.ts
```diff
@@ -50,6 +55,9 @@ export interface MessageExportRow {
   source: string;
   toolsUsed: string;
   skillsUsed: string;
+  modelId: string;
+  modelProviderId: string;
+  modelResolutionMethod: string;
```
fabiencelier, line 60: Do you think modelResolutionMethod is valuable for users ?

fabiencelier (approved):
LGTM ✅

front/lib/api/analytics/messages_export.ts
sfriquet: idk, I don't use it but I thought it was harmless to add it now just in case (it would avoid a backfill later). It could later be useful if we want to dig into what the `auto` mode did

PR #29606, opened 2026-07-28
[front] Remove global_disable_metronome_billing kill switch

tdraier:
## Description

Removes the `global_disable_metronome_billing` kill switch, which forced all workspaces onto legacy Stripe billing. Metronome billing is now default-on in prod, so the switch is no longer needed; the per-workspace `legacy_billing` feature flag remains the only way to opt a workspace out. This drops the switch from `KILL_SWITCH_TYPES` and the Poke KillPage, and simplifies both the server (`isMetronomeBillingEnabled`) and client billing gates to check only `legacy_billing`.

## Tests

Updated the existing functional tests in `front-api/routes/w/[wId]/subscriptions/` to remove the kill-switch scenarios; the `legacy_billing`-flag paths remain covered. Type-checks pass across `front`, `front-api`, and `front-spa`.

## Risk

Low. Behavior only changes if the kill switch was actively enabled in an environment (it is not, in prod). Fully rollback-safe by reverting. The now-dead `kill_switches` row with this type is harmless and can be cleaned up separately.

## Deploy Plan


fabiencelier (approved):
LGTM ✅ 
🧹 🧹

PR #29607, opened 2026-07-28
Use Fireworks Responses endpoint for Kimi K3

fontanierh:
## Description

- Route Kimi K3 through Fireworks Responses so reasoning item IDs, function calls, and tool results are replayed without flattening them into assistant text.
- Preserve Responses reasoning IDs based on the persisted wire metadata rather than the model provider, including interleaved reasoning and tool-call items.
- Preserve ID-bearing reasoning items even when the provider returns an empty visible summary, matching Kimi Code multi-step replay behavior.
- Send `service_tier: "priority"`, disable provider storage, and adapt named forced tools to the tool-choice shape Fireworks accepts.

## Tests

- Ran the live Kimi K3 endpoint suite against Fireworks with `service_tier: "priority"`: all 46 cases passed, including multi-turn following, forced tools, reasoning efforts, and structured output.
- Ran a live stateless tool continuation with `store: false`: Fireworks accepted the replayed reasoning item ID and tool result, then returned fresh reasoning and the expected final answer.
- Added focused coverage for raw interleaved item ordering, Fireworks reasoning-ID persistence and replay, empty-summary replay state, forced tools, and hosted tool-search filtering.

## Risk

The provider endpoint change is scoped to Kimi K3. Reasoning-ID detection now applies to any Responses provider that persists an `id`, matching the Responses wire contract. The endpoint ID and pricing remain unchanged, so rollback is a normal code revert.

Fireworks' Responses frontend parses `service_tier`, but the current Responses reference does not document it and responses do not expose the selected tier. Confirm priority routing and billing with Fireworks before production rollout.

## Deploy Plan

Standard deploy after Fireworks confirms Responses routes and bills the priority tier.

### DIFF
--- a/front/lib/model_constructors/stream/clients/fireworks.ts
+++ b/front/lib/model_constructors/stream/clients/fireworks.ts
@@ -1,4 +1,8 @@
 import {
+  FIREWORKS_BASE_URL,
+  FIREWORKS_MODEL_PREFIX,
+} from "@app/lib/model_constructors/providers/fireworks/constants";
+import {
   type FireworksInputConfig,
   fireworksConfigSchema,
 } from "@app/lib/model_constructors/providers/fireworks/inputConfig";
@@ -20,12 +24,6 @@ import type {
   ChatCompletionCreateParamsStreaming,
   ChatCompletionToolMessageParam,
 } from "openai/resources/chat/completions";
-
-export const FIREWORKS_BASE_URL = "https://api.fireworks.ai/inference/v1";
-
-// Fireworks model ids are stored bare (e.g. `glm-5p2`) but legacy `ModelIdType`
-// and the Fireworks API both use the full account-scoped path.
-export const FIREWORKS_MODEL_PREFIX = "accounts/fireworks/models/";
 
 type FireworksToolMessageParam = ChatCompletionToolMessageParam & {
   name: string;
--- a/front/lib/model_constructors/stream/clients/fireworks.ts
+++ b/front/lib/model_constructors/stream/clients/fireworks.ts
@@ -21,7 +21,7 @@ import type {
   ChatCompletionToolMessageParam,
 } from "openai/resources/chat/completions";
 
-const FIREWORKS_BASE_URL = "https://api.fireworks.ai/inference/v1";
+export const FIREWORKS_BASE_URL = "https://api.fireworks.ai/inference/v1";
 
 // Fireworks model ids are stored bare (e.g. `glm-5p2`) but legacy `ModelIdType`
 // and the Fireworks API both use the full account-scoped path.
--- a/front/lib/model_constructors/stream/clients/fireworks_responses.ts
+++ b/front/lib/model_constructors/stream/clients/fireworks_responses.ts
@@ -1,0 +1,105 @@
+import {
+  type FireworksInputConfig,
+  fireworksConfigSchema,
+} from "@app/lib/model_constructors/providers/fireworks/inputConfig";
+import { WithOpenAIResponsesInputConverter } from "@app/lib/model_constructors/sdk/openai_responses/converters/input";
+import { WithOpenAIResponsesOutputConverter } from "@app/lib/model_constructors/sdk/openai_responses/converters/output";
+import { rawOutputToEvents } from "@app/lib/model_constructors/sdk/openai_responses/converters/output/utils";
+import {
+  FIREWORKS_BASE_URL,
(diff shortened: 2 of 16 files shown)


front/lib/model_constructors/stream/clients/fireworks.ts
```diff
@@ -21,7 +21,7 @@ import type {
   ChatCompletionToolMessageParam,
 } from "openai/resources/chat/completions";
 
-const FIREWORKS_BASE_URL = "https://api.fireworks.ai/inference/v1";
+export const FIREWORKS_BASE_URL = "https://api.fireworks.ai/inference/v1";
```
pmilliotte: should put it in separate file like `front/lib/model_constructors/stream/types.ts`

`FIREWORKS_MODEL_PREFIX` too

front/lib/model_constructors/stream/clients/fireworks_responses.ts
```diff
@@ -0,0 +1,105 @@
...
+
+  override buildRequestPayload(
+    payload: Payload,
+    config: FireworksInputConfig
+  ): ResponseCreateParamsNonStreaming {
+    // Fireworks implements Responses function tools, but not OpenAI's hosted
+    // tool search/deferred-loading extension.
+    const request = super.buildRequestPayload(payload, {
+      ...config,
+      toolSearchEnabled: false,
```
pmilliotte: inputConfigSchema should forbid `true` value for `toolSearchEnabled` 

Then if necessary, we can set it to false in a configParser in Dust layer

front/lib/model_constructors/stream/clients/fireworks_responses.ts
```diff
@@ -0,0 +1,105 @@
...
+        : {}),
+    };
+  }
+
+  async *streamRaw(
+    input: ResponseCreateParamsNonStreaming
+  ): AsyncGenerator<ResponseStreamEvent> {
+    const streamingInput: ResponseCreateParamsStreaming = {
+      ...input,
+      stream: true,
```
pmilliotte: should be set in buildRequestPayload

front/lib/model_constructors/stream/clients/fireworks_responses.ts
```diff
@@ -0,0 +1,105 @@
...
+      store: false,
+      // Fireworks does not accept OpenAI's named function tool-choice object.
+      // Restricting the request to that function and requiring a tool call has
+      // the same forced-tool semantics.
+      ...(forcedTools.length > 0
+        ? {
+            tools: forcedTools,
+            tool_choice: "required" as const,
+          }
+        : {}),
```
pmilliotte: Looks like we can do better and save some prompt caching by doing sth like:

```
      tool_choice: disableToolUse ? "none" : forceTool
        ? {
            type: "allowed_tools",
            mode: "required",
            tools: [{ type: "function", name: forceTool }],
          }
        : undefined,
```

pmilliotte (approved):
Approved with requested changes

front/lib/model_constructors/stream/clients/fireworks_responses.ts
fontanierh: 🙏

front/lib/model_constructors/stream/clients/fireworks.ts
fontanierh: Moved both values to the neutral providers/fireworks/constants.ts module and updated all consumers in c7c34a89b4.

front/lib/model_constructors/stream/clients/fireworks_responses.ts
fontanierh: Done in c7c34a89b4: the Fireworks schema now accepts only false or undefined. The Dust layer already enables tool search only for Anthropic and OpenAI, so no config parser is needed.

front/lib/model_constructors/stream/clients/fireworks_responses.ts
fontanierh: Done in c7c34a89b4: buildRequestPayload now returns the streaming request with stream: true, and streamRaw forwards it unchanged.

front/lib/model_constructors/stream/clients/fireworks_responses.ts
fontanierh: Agreed and implemented in c7c34a89b4. The full tool list stays stable and allowed_tools constrains the required call. I also validated this exact shape live with Kimi K3 in priority streaming mode.

PR #29609, opened 2026-07-28
Opt: refactor content fragments deletion.

Fraggle:
## Description

Replace per-fragment GCS + DB deletion in `destroyConversation` with a single `getPrivateUploadBucket().deleteByPrefix()` call on the conversation's GCS path prefix, followed by a batched `ContentFragmentModel.destroy()` WHERE IN. Previously `destroyContentFragments` fetched each `ContentFragmentResource`, issued 2 individual GCS deletes (`text` + `raw`) and 1 DB delete per fragment — O(N) round trips for conversations with many attachments. Now a single prefix delete covers all `conversations/{conversationId}/content_fragment/` objects upfront, before any DB rows are touched, so Temporal can safely retry on GCS failure.

Also removes `ContentFragmentResource.destroy()`, the per-item method that drove the old flow.

## Tests

- Added `deleteByPrefix` to `FileStorageMock` so existing `destroyConversation` tests cover the new GCS path.
- Tested locally.

## Risk

Low. GCS prefix delete runs before any DB rows are removed, keeping the operation Temporal-retry-safe. The only behavioral change is that a GCS error now aborts destroy cleanly before touching DB — strictly safer than the previous partial-deletion scenario. Rollback is straightforward (no schema changes).

## Deploy Plan

Deploy front.

### DIFF
--- a/front/lib/api/assistant/conversation/destroy.ts
+++ b/front/lib/api/assistant/conversation/destroy.ts
@@ -1,6 +1,7 @@
 import { hardDeleteDataSource } from "@app/lib/api/data_sources";
 import { deleteOwnerPolicy } from "@app/lib/api/sandbox/egress_policy";
 import type { Authenticator } from "@app/lib/auth";
+import { getPrivateUploadBucket } from "@app/lib/file_storage";
 import { AgentSuggestionModel } from "@app/lib/models/agent/agent_suggestion";
 import {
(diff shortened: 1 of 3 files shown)


Fraggle:

- #29608
    - #29609 ◀

Fraggle:
PMRR (to get it out for today's purge)

PR #29611, opened 2026-07-28
[front] tweak Skills state management

ykmsd:
## Description
This PR is to do mini update in Skill management page

- update the column name from access to availability
- update the copy in admin governance page
- update the modal design slightly

### DIFF
--- a/front/components/skills/SkillsBatchEdit.tsx
+++ b/front/components/skills/SkillsBatchEdit.tsx
@@ -20,8 +20,6 @@ export type BatchAvailabilityAction = {
   availability: SkillAvailability;
   getDialogTitle: (count: number) => string;
   dialogDescription: string;
-  confirmLabel: string;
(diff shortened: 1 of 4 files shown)


front/components/skills/SkillsBatchEdit.tsx
```diff
@@ -132,16 +124,15 @@ export function BatchAvailabilityDialog({
...
         <DialogFooter
           leftButtonProps={{
             label: "Cancel",
             variant: "outline",
             disabled: isUpdating,
           }}
           rightButtonProps={{
-            label: action.confirmLabel,
-            variant: action.confirmVariant,
+            label: "Update",
```
ykmsd, line 137: Change everything to Update + primary button, it's not a destructive action so it's probably better to not look scary

matteotrab (approved):

PR #29612, opened 2026-07-28
[front] fix: Revert cron trigger and wake-up jitter (#29538)

id13:
## Description

Revert the jitter now that it's no longer needed

## Plan

Deploy front

### DIFF
--- a/front/temporal/triggers/workflows.ts
+++ b/front/temporal/triggers/workflows.ts
@@ -1,34 +1,26 @@
-import { isCronScheduleConfig } from "@app/types/assistant/triggers";
 import { ActivityFailure, RetryState } from "@temporalio/common";
-import { proxyActivities, sleep } from "@temporalio/workflow";
+import { proxyActivities } from "@temporalio/workflow";
+
 import type * as activities from "./activities";
 
-// We spread the start time of triggers and wake-ups starting on the hour over 10 minutes to avoid all conversations starting at the same time.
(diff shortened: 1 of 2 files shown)


id13:

- #29955
    - #29614
        - #29612 ◀
            - #29652

front/temporal/triggers/workflows.ts
```diff
@@ -1,35 +1,27 @@
...
 
-// We spread the start time of triggers and wake-ups starting on the hour over 10 minutes to avoid all conversations starting at the same time.
-const ON_THE_HOUR_JITTER_MS = 60 * 10 * 1000;
+import type * as activities from "./activities";
 
-const getJitterMs = () => {
-  return Math.floor(Math.random() * ON_THE_HOUR_JITTER_MS);
-};
+const { runTriggeredAgentsActivity } = proxyActivities<typeof activities>({
+  startToCloseTimeout: "5 minutes",
```
flvndvd: Do we really need 5 minutes, here? Given that when we deploy we have a 2 minutes cooling/draining period, would be nice to lower.

flvndvd (approved):

front/temporal/triggers/workflows.ts
id13: I think it is what was before. Not strongly held

id13: will lower, worst case we will either send more heartbeats or increase back

PR #29613, opened 2026-07-28
[front] enh: Add per-class QoS task queues and workers for the agent loop

id13:
## Description

Currently, we mix all agent-loops in a single pool of workers, which mixes problem such as 
- CPU throttling ~ 300ms per CPU second
- Connection pool starving (due to idle in tx)
- DB latency spikes
- I/O spikes
- LLM provider rate limitting

This stack is about splitting the agent-loop per QoS (quality of service) commitments we want to impose on ourselves, which falls in the following buckets by order of priority
  - interactive : human in the loop - as fast as possible, high number of workers, more resources
  - schedules : time commitments - we try as much as possible to honor them
  - programmatic : most of /api/v1 surface - limited throughput compared to the two above
  - batch : reinforcement, webhooks, transcripts etc .. best effort throughput.

An additional perk of doing this is to limit the blast radius of one source bursting us (eg schedules, programmatic or batch) on all the resources that are finite like DB connection/cpu, LLM rate limits, with a knob we can control (# of deployments + activitySlots) 

## Plan

This PR: introduce the machinery and the different pools - but no-op, we still map everything to the current agent-loop
Infra PR : create the workers with the "right" sizing and the scale ups

Next PRs: 
- Add the routing logic gated by a feature flag
- Remove jitter in schedules
- Remove jitter in reinforcement

## Risks

- on this PR it's mostly about bug in the codepath, not too much concerned though, will frontedge it
- the real risks is not sizing properly the different worker pools and either or both of:
   - blast the db
   - create un-wanted back-pressure (schedule to start will start flying to the roof)

### DIFF
--- a/front/temporal/agent_loop/worker.ts
+++ b/front/temporal/agent_loop/worker.ts
@@ -24,10 +24,17 @@ import {
 import { publishDeferredEventsActivity } from "@app/temporal/agent_loop/activities/publish_deferred_events";
 import { runModelAndCreateActionsActivity } from "@app/temporal/agent_loop/activities/run_model_and_create_actions_wrapper";
 import { runToolActivity } from "@app/temporal/agent_loop/activities/run_tool";
-import { QUEUE_NAME } from "@app/temporal/agent_loop/config";
+import {
+  BATCH_QUEUE_NAME,
+  INTERACTIVE_QUEUE_NAME,
+  PROGRAMMATIC_QUEUE_NAME,
+  QUEUE_NAME,
+  SCHEDULES_QUEUE_NAME,
+} from "@app/temporal/agent_loop/config";
 import { instrumentationSinks } from "@app/temporal/agent_loop/sinks";
 import { getWorkflowConfig } from "@app/temporal/bundle_helper";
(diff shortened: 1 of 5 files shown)


id13:

- #29613 ◀
    - #29614
        - #29612
            - #29652

front/temporal/agent_loop/worker.ts
```diff
@@ -45,15 +53,71 @@ const SHUTDOWN_TOOL_ABORT_DELAY_MS =
   SHUTDOWN_GRACE_TIME_MS - SHUTDOWN_ABORT_BUFFER_MS;
 
 export async function runAgentLoopWorker() {
+  return runAgentLoopWorkerForQueue({
+    taskQueue: QUEUE_NAME,
+    serviceName: "dust-agent-loop",
+    maxConcurrentActivityTaskExecutions: 75,
+  });
```
id13, line 58: current agent-loop, left as is

id13:
@flvndvd this is mostly dead code, it gets used in the next PR

id13:
PMRR

flvndvd (approved):

PR #29615, opened 2026-07-28
[AGENT-LOOP] Don't fail the message when a step comes back empty

pmilliotte:
## Description

An agent step that came back with no tool call and no text failed the whole message with `empty_content`, even when the answer had already been written in an earlier step — the check only looked at the text streamed in the current step. It now checks the answer accumulated on the agent message, and when there is genuinely nothing anywhere it re-runs the step once with tool use disabled to force a final answer before erroring.

## Tests

Existing agent-loop unit tests pass; validated against a production trace where the answer was written in step 1 and step 2 came back empty.

## Risk

Low — the forced re-call only happens on a step that would otherwise have failed.

### No Temporal patching needed

Patching is only required when new workflow code, replaying an old history, would emit a different sequence of commands.

- Most of the change is in activity code (`run_model.ts`, the activity wrapper). Activities are never re-executed on replay — only their recorded result is read — so changing them is always safe.
- The workflow change is a no-op on old histories. An activity result recorded before this deploy has no `hasNoContent` field, so on replay it reads `undefined → ?? false`, and `shouldContinue: runId === null || false` is exactly the old `runId === null`. The new branch is only reachable when the field is present, which only happens in results produced by the new code.
- The new `forceDisableToolUse` argument is fine too: replay matches scheduled activities by their position in the command sequence, not by comparing payloads.

The flag is optional and negative on purpose. A required `hasContent` would read as `undefined → falsy → "no content"` on every old history and trigger a spurious extra step — that would need a patch.

A workflow started on old code and still alive (e.g. parked on a tool approval) picks up the new behaviour for its future steps. That is not a determinism issue: those commands are not in history yet, so replay has nothing to disagree with.

## Deploy Plan

- Deploy `main` on `front`

### DIFF
--- a/front/temporal/agent_loop/lib/run_model.ts
+++ b/front/temporal/agent_loop/lib/run_model.ts
@@ -365,6 +365,7 @@ export async function runModel(
     functionCallStepContentIds,
     durationRecorder,
     activityTimeoutDeadlineMs,
+    forceDisableToolUse = false,
   }: {
     runAgentData: AgentLoopExecutionData;
     runIds: string[];
@@ -372,12 +373,17 @@ export async function runModel(
     functionCallStepContentIds: Record<string, ModelId>;
     durationRecorder: DurationRecorder;
     activityTimeoutDeadlineMs: number;
+    // Set when the previous step came back empty: force the final generation.
+    forceDisableToolUse?: boolean;
   }
 ): Promise<{
   actions: AgentActionsEvent["actions"];
   runId: string;
   functionCallStepContentIds: Record<string, ModelId>;
   stepContexts: StepContext[];
+  // The step produced nothing at all: the loop should run one more step with
+  // tool use disabled to force a final answer.
+  retryWithoutTools?: boolean;
 } | null> {
   const {
     agentConfiguration,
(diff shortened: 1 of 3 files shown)


front/temporal/agent_loop/lib/run_model.ts
```diff
@@ -952,9 +961,41 @@ export async function runModel(
...
+    const forcedRes = forcedLLM
+      ? await getOutputFromLLMStream(auth, {
+          ...llmStreamParams,
+          disableToolUse: true,
+          llm: forcedLLM,
+        })
+      : null;
+
+    if (forcedRes?.isOk()) {
+      ({ dustRunId, nativeChainOfThought, output } = forcedRes.value);
```
flvndvd: By doing so here, we are actually loosing the previous run which should still be billed and attached to this one message.

front/temporal/agent_loop/lib/run_model.ts
```diff
@@ -1010,10 +1016,37 @@ export async function runModel(
...
+          { modelId: modelConfig.modelId },
+          "Empty model turn, retrying in a new step without tool use."
+        );
+
+        return {
+          actions: [],
+          runId: dustRunId,
+          functionCallStepContentIds: updatedFunctionCallStepContentIds,
+          stepContexts: [],
+          retryWithoutTools: true,
```
aubin-tchoi, line 1046: nit on the naming here: I feel like this field should just state what happened (no content) rather than dictating what the agent loop should do with the output

aubin-tchoi (approved):
LGTM

PR #29616, opened 2026-07-28
Persist reasoning token usage on model runs

flvndvd:
## Description

Providers report reasoning tokens as part of completion usage, but this information was previously discarded before persisting `RunUsage`.

This change stores reasoning tokens as a nullable provider fact while keeping completion tokens inclusive and leaving pricing unchanged (completionTokens = output + reasonning). It propagates the value through recording, tracing, analytics, and simulated usage paths.

## Deploy Plan

- [ ] Run SQL migration
- [ ] Deploy front


pmilliotte (approved):

id13 (approved):
LGTM

PR #9478, opened 2024-12-18
[connectors] - fix(slack): assistant slack integration

JulesBelveze:
## Description

This PR transform Slack internal channel IDs to public Slack Channel IDs before querying.

**References:**
- https://github.com/dust-tt/tasks/issues/1834

## Risk

Low

## Deploy Plan

- Deploy `connectors`
- Deploy `front`

PR #29619, opened 2026-07-28
[front] Backfill + dual-write space group_permissions (#9478)

tdraier:
## Description

Second step of the space-permissions migration ([#9477](https://github.com/dust-tt/tasks/issues/9477)) — sub-issue [#9478](https://github.com/dust-tt/tasks/issues/9478): **write space permissions directly into the `group_permissions` table**, plus a one-off backfill for existing spaces.

Stacked on #29379. Runs **first** (before shadow/flip) so the table is populated and kept in sync going forward; the shadow-compare (#29639) reads it and the flip (#29763) serves it.

### Approach

The `group_permissions` table becomes the store for space group access. Two entry points, one shared derivation:

- **`SpaceResource.spaceGroupGrants()`** — the group grants a space confers, derived from its `group_vaults` associations (per-kind verbs mirroring the legacy inline-group rules). This is the single source both writers use; `getAccessControlLists` also builds its `groups` from it.
- **`SpaceResource.writeGroupPermissions(auth, { transaction })`** — the direct writer: re-reads the space fresh (mutation paths leave `this.groups` stale), then clears the space's instance grants and inserts `spaceGroupGrants()` (`deleteAllForResource` + `grantMany`) in one transaction. Idempotent.
- **`SpaceResource.reconcileGroupPermissions(auth, ...)`** — thin backfill entry that delegates to `writeGroupPermissions`, so the one-off backfill and the ongoing writes can never disagree.

### Write sites (every `group_vaults` mutation)
- `makeNew` — writes grants for the created member/editor groups when passed an `auth` (new optional arg)
- `createSpace` — passes `auth` to `makeNew`, then writes again once the global-viewer / group-mode associations are in place
- `SpaceResource.updatePermissions` — end of transaction
- `SpaceResource.makeDefaultsForWorkspace` — the 3 default spaces on onboarding

### Backfill
`migrations/20260728_backfill_space_group_permissions.ts` — `makeScript` + `runOnAllWorkspaces`, one transaction per space, dry-run by default. Idempotent (clears + re-inserts), so it is **safe to re-run** if grants ever get corrupted.

### Known replicated inconsistencies
The per-kind verbs replicate current behaviour verbatim so shadow (#29639) stays green; see the [card note](https://github.com/dust-tt/tasks/issues/9478#issuecomment-5101970546): `project_viewer` confers read+write (not read), and `member` verbs differ by space kind. Both revisited after the flip.

## Risk
Low. Additive rows; the served permission decision is unchanged (nothing reads the table for spaces until #29763).

## Tests
`space_resource`, `spaces`, `group_permission` suites; front-api spaces/governance/files; `tsgo` clean.

### DIFF
--- a/front/migrations/20260728_backfill_space_group_permissions.ts
+++ b/front/migrations/20260728_backfill_space_group_permissions.ts
@@ -1,18 +1,19 @@
 import { Authenticator } from "@app/lib/auth";
 import { SpaceResource } from "@app/lib/resources/space_resource";
 import { frontSequelize } from "@app/lib/resources/storage";
-import { WorkspaceResource } from "@app/lib/resources/workspace_resource";
 import { concurrentExecutor } from "@app/lib/utils/async_utils";
-import { renderLightWorkspaceType } from "@app/lib/workspace";
 import type { Logger } from "@app/logger/logger";
 import { makeScript } from "@app/scripts/helpers";
 import { runOnAllWorkspaces } from "@app/scripts/workspace_helpers";
 import type { LightWorkspaceType } from "@app/types/user";
 
+const WORKSPACE_CONCURRENCY = 2;
+const SPACE_CONCURRENCY = 4;
+
 // Backfill (#9478): (re-)derive every space's group_permissions from its group_vaults associations.
-// The write goes through SpaceResource.reconcileGroupPermissions, which delegates to the same
-// writeGroupPermissions the space mutation paths use, so the backfill and the ongoing writes can
-// never disagree. Idempotent (clears + re-inserts), so it is safe to re-run if grants get corrupted.
+// The write goes through SpaceResource.writeGroupPermissions — the same method the space mutation
+// paths call — so the backfill and the ongoing writes can never disagree. Idempotent (clears +
+// re-inserts), so it is safe to re-run if grants get corrupted.
 async function backfillWorkspaceSpaceGroupPermissions(
   execute: boolean,
   logger: Logger,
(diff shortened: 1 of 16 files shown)


tdraier:

- #29619 ◀

front/migrations/20260728_backfill_space_group_permissions.ts
```diff
@@ -0,0 +1,92 @@
...
+// Backfill (#9478): (re-)derive every space's group_permissions from its group_vaults associations.
+// The write goes through SpaceResource.writeGroupPermissions — the same method the space mutation
+// paths call — so the backfill and the ongoing writes can never disagree. Idempotent (clears +
+// re-inserts), so it is safe to re-run if grants get corrupted.
+async function backfillWorkspaceSpaceGroupPermissions(
+  execute: boolean,
+  logger: Logger,
+  workspace: LightWorkspaceType
+): Promise<void> {
+  const auth = await Authenticator.internalAdminForWorkspace(workspace.sId);
```
fabiencelier, line 22: you may need the flag `dangerouslyRequestAllGroups: true,` to have access to all the groups, otherwise I think some will get skipped 
Happened to me in the skill migration 🙄

fabiencelier (approved):
LGTM
one comment worth checking

front/migrations/20260728_backfill_space_group_permissions.ts
matteotrab: Thanks for flagging!
It should not be needed in this case as `listWorkspaceSpaces` directly fetches all spaces and the groups base fetch does not check if the groups can be read

PR #29622, opened 2026-07-28
Fix self improving skills notification using the Noop model

fabiencelier:
## Description

Broken recently, the generated message no longer used the noop model which render the message from metadata.
This restores the behavior.

## Tests

tested locally, notif send is exactly the hardcoded message 

Screenshot 2026-07-28 at 15 44 59

## Risk

low

## Deploy Plan

deploy front


Fraggle (approved):
LGTM

PR #29625, opened 2026-07-28
[front] perf: Set autovacuum_analyze_scale_factor to 5% on big row tables

id13:
## Description

We are currently running analyze with a scale factor of 0 and 10k rows threshold.
Introduced in https://github.com/dust-tt/dust-infra/pull/742, was supposed to fine tune the value for big table, never did it until today 🫠 .

Why now ? 

I believe on big tables, running analyze multiple time per hour is not only useless, but performance-wise detrimental.
Yes, pg is not going to lookup all rows, it samples 30k so it does not scale in that direction BUT we have huge columns (JSONB, TOASTED, as always) that are busting a lot of shared_buffer because of how big they are, very often, which is unhealthy for the DB, and because it's TOAST, we also spend CPU time to de-TOAST 30k rows every now and then.

## Tests

N/A

## RIsk

Low - very conservative values, based on the # of rows.

## Plan

Merge
Run the migrationn


Fraggle (approved):
LGTM
side note: could be a post deploy, no ?

id13:
does not require a deploy, so not really

PR #29626, opened 2026-07-28
Revert "[front] Refresh activation pod Frame template and prompt (#29…

frankaloia:
### Summary
Roll back previous activation changes (I think output is lower quality - we will have a separate PR to address the feedback in a different way)

### Testing
* Manual

### Risk
Low - not customer facing

### Deploy
1. Deploy front


id13 (approved):
LGTM

PR #29629, opened 2026-07-28
[Sparkle] Form contrast improvement

ClementAupiais:
## Description

Form control borders and focus rings in Sparkle failed WCAG 1.4.11 (non-text contrast, minimum 3:1 ratio). The `Input`'s resting border (`stone-100`) measured only ~1.09:1 against the field background.

Two new design tokens are introduced — `border-form` (resting, `stone-300`) and `border-form-active` (focused/filled/checked, `stone-500`) — chosen as the lightest steps that clear 3:1 in both light and dark themes. `Input`, `Checkbox`, and `RadioGroup` are updated to use these tokens. Focus ring colors (`ring`, `ring-warning`) are also raised to their first passing steps (`blue-500` / `stone-500` / `rose-500`) since the ring is the sole focus indicator on most controls. Error borders on `Input` move from `warning-300/400` to `warning-500/600` for the same reason.

## Tests

Visual inspection in Storybook (Vercel preview deployed). Contrast ratios can be verified with a browser accessibility tool against the updated token values.

## Risk

Low. Pure visual/token change within the Sparkle design system. No logic or API changes. Components that relied on the previous (too-light) border or ring colors will appear slightly darker, which is the intended correction
```

### DIFF
--- a/sparkle/src/components/Checkbox.tsx
+++ b/sparkle/src/components/Checkbox.tsx
@@ -10,7 +10,9 @@ export const checkboxStyles = cva(
   cn(
     "h-4 w-4 rounded-md relative shrink-0 peer border transition duration-100 ease-out motion-reduce:transition-none",
     "active:scale-95",
-    "border-border-dark bg-background",
+    "border-border-form bg-background",
+    "data-[state=checked]:border-border-form-active",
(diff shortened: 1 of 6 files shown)


pinotalexandre (approved):
LGTM modulo linter

PR #29633, opened 2026-07-28
[Admin Gov] Emit audit log events on group member add/remove

philipperolet:
## Description

Group membership changes made through the manual group management API (group creation with initial members, and \`PATCH\` of a group's \`memberIds\`) were not audit-logged. SCIM-provisioned membership changes already emit \`scim.group_user_added\` / \`scim.group_user_removed\`, but manual changes by admins/managers left no trace.

This adds two audit actions, \`group.member_added\` and \`group.member_removed\`, emitted once per affected user (targets: workspace, group, user; metadata: group name, user email) from the two group management handlers, via a shared \`emitGroupMemberAuditLogs\` helper in \`lib/api/groups/audit.ts\`. Emission is fire-and-forget with bounded concurrency since \`memberIds\` payloads are unbounded.

To know which users actually changed, \`dangerouslySetMembers\` now returns the \`{ addedUsers, removedUsers }\` diff it already computes internally (covered by a new resource test), \`updateRegularManualGroup\` propagates it, and \`makeNewRegularManual\` returns the created group along with the users added at creation. Existing callers only check for errors, so this is non-breaking (one call site in \`group_space_resource.ts\` adjusted to keep returning \`undefined\`).

Note: emission lives in the handlers rather than \`GroupResource\` because importing \`workos_audit\` from \`group_resource.ts\` creates a module-initialization cycle (\`group_resource → workos_audit → lib/auth → resources\`) that breaks the temporal bundles build.

Schemas are registered with WorkOS (v1) and the version map is committed in this PR.

## Risks

Blast radius: manual group management API (group create/update); audit logging only, fire-and-forget.
Risk: low

## Deploy Plan

- deploy front

### DIFF
--- a/front/lib/api/groups/audit.ts
+++ b/front/lib/api/groups/audit.ts
@@ -1,0 +1,61 @@
+import {
+  buildAuditLogTarget,
+  emitAuditLogEvent,
+  getAuditLogContext,
+} from "@app/lib/api/audit/workos_audit";
+import type { Authenticator } from "@app/lib/auth";
+import type { GroupResource } from "@app/lib/resources/group_resource";
+import { concurrentExecutor } from "@app/lib/utils/async_utils";
+import type { UserType } from "@app/types/user";
+
+// Bounds the fire-and-forget WorkOS calls: memberIds payloads are unbounded, so a large group
+// update could otherwise burst hundreds of concurrent requests.
+const EMIT_CONCURRENCY = 8;
+
+export function emitGroupMemberAuditLogs(
+  auth: Authenticator,
+  group: GroupResource,
+  {
+    addedUsers,
+    removedUsers,
+  }: { addedUsers: UserType[]; removedUsers: UserType[] }
+): void {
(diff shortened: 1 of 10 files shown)


tdraier (approved):

front/lib/api/groups/audit.ts
```diff
@@ -0,0 +1,61 @@
...
+  buildAuditLogTarget,
+  emitAuditLogEvent,
+  getAuditLogContext,
+} from "@app/lib/api/audit/workos_audit";
+import type { Authenticator } from "@app/lib/auth";
+import type { GroupResource } from "@app/lib/resources/group_resource";
+import { concurrentExecutor } from "@app/lib/utils/async_utils";
+import type { UserType } from "@app/types/user";
+
+// Bounds the fire-and-forget WorkOS calls: memberIds payloads are unbounded, so a large group
```
zmarouf, line 11: neat

zmarouf (approved):

PR #29638, opened 2026-07-28
Optimize MCP output persistence to write DB rows once

flvndvd:
## Description

See thread [here](https://dust4ai.slack.com/archives/C050SM8NSPK/p1785255327670479?thread_ts=1785251882.524509&cid=C050SM8NSPK).

MCP action output items were previously inserted before their GCS paths were known, then updated individually after uploading their content. For a batch of N outputs, this resulted in one bulk insert followed by N update queries.

These updates were more expensive than the query count alone suggests. PostgreSQL had to create new row versions, keep indexes up-to-date. They also introduced extra database round trips and increased table bloat.

This change gives each new GCS object a storage-only UUID. Contents are uploaded first, allowing the final GCS paths to be included directly in a single `bulkCreate`. Each output row is therefore written once with its complete state.

GCS writes retain their existing per-object retries. If part of a batch still fails, every attempted path is cleaned up and no database rows are created. Database errors preserve the uploaded objects because the outcome of an interrupted
commit can be ambiguous. Action-prefix cleanup removes any resulting orphan safely. Cache warming is now best-effort so a Redis failure cannot cause an already-persisted output batch to be retried.

Action deletion now clears the complete GCS prefix even when no output path reached the database, covering crashes between GCS upload and database insertion. The migration backfill continues using deterministic item-based paths so rerunning it remains idempotent (still need one week end of backfill).

### DIFF
--- a/front/lib/resources/agent_mcp_action/output_storage.ts
+++ b/front/lib/resources/agent_mcp_action/output_storage.ts
@@ -1,3 +1,4 @@
+import { randomUUID } from "node:crypto";
 import { REDIS_CACHE_CONCURRENCY } from "@app/lib/api/redis";
 import type { Authenticator } from "@app/lib/auth";
 import { getPrivateUploadBucket } from "@app/lib/file_storage";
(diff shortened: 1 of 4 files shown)


front/lib/resources/agent_mcp_action/output_storage.ts
```diff
@@ -36,31 +37,21 @@ function getActionOutputGcsPrefix(
 function getGcsPath(
   auth: Authenticator,
   action: AgentMCPActionResource,
-  itemId: ModelId
+  objectName: string
```
flvndvd, line 40: Will change the name once removing the backfill script.

front/lib/resources/agent_mcp_action/output_storage.ts
```diff
@@ -294,10 +333,6 @@ export async function deleteActionOutputsFromGcs(
   actionIds: ModelId[],
   gcsPaths: string[]
 ): Promise<Result<void, Error>> {
-  if (gcsPaths.length === 0) {
```
flvndvd, line 297: Not needed anymore as we delete by prefix now.

front/lib/resources/agent_mcp_action/output_storage.ts
```diff
@@ -1,3 +1,4 @@
+import { randomUUID } from "node:crypto";
 import { REDIS_CACHE_CONCURRENCY } from "@app/lib/api/redis";
```
flvndvd, line 2: Better reviewed with hidden whitespace: [here](https://github.com/dust-tt/dust/pull/29638/changes?w=1).

front/lib/resources/agent_mcp_action_resource.ts
```diff
@@ -849,72 +847,74 @@ export class AgentMCPActionResource extends BaseResource<AgentMCPActionModel> {
...
+            generatedFilePath,
+            generatedFileContentType,
+          };
         })
-      )
-    );
+      );
+    } catch (err) {
+      // A DB error can be ambiguous after commit, so keep the GCS objects rather than risk
+      // deleting content referenced by committed rows. Action-prefix cleanup removes orphans.
```
id13, line 892: How about putting this bulk create in a transaction, and
1. either sequelize auto-rollbacks on error
2. or we catch and rollback

then you no longer have to think about half-done creates anymore

just check this method does no other work than DB inserts

id13 (approved):
lgtm, left a suggestion

front/lib/resources/agent_mcp_action_resource.ts
flvndvd: `bulkCreate` here is a single PostgreSQL statement, so the batch is already atomic. So my read is that an explicit transaction would add `BEGIN`/`COMMIT` but wouldn’t improve the DB guarantee or cover GCS. A commit/connection failure could still leave the outcome unknown, so we would still need to retain the GCS objects, right?

PR #29639, opened 2026-07-28
[front] Composed shadow-compare for space group_permissions (#9479)

tdraier:
## Description

Third step of the space-permissions migration ([#9477](https://github.com/dust-tt/tasks/issues/9477)) — sub-issue [#9479](https://github.com/dust-tt/tasks/issues/9479): **composed shadow-compare** across space kinds.

Stacked on #29619 (backfill). With the table populated, this verifies — in production, per-workspace, behind a flag — that flipping to the table (#29763) will be safe.

### What it does

While the `group_permissions_shadow` flag is on for a workspace, `SpaceResource.canRead` / `canWrite` / `canAdministrate` fire a **background** comparison of two composed decisions for the same check, keeping the (unchanged) code **role** rules on both sides:

- **legacy** = `getAccessControlLists(auth)` — the served config (roles + inline groups)
- **candidate** = the same roles with groups from `auth.getGroupPermissions("space", id)` — the `group_permissions` table (an `AccessControlList`, built by `governanceAcls`)

A divergence logs `group_permissions_shadow_mismatch` (stable Datadog key) with the space, permission, and both results. It compares **composed booleans** (`hasPermissionForAcls` on each ACL set), so it measures exactly what the flip will change.

- **Never changes the served result** — fire-and-forget, swallows its own failures.
- **system / global / conversations** spaces stay code-ruled and are skipped.
- Flag off ⇒ short-circuits before any comparison.

The orchestration is an `Authenticator` method (`shadowComparePermission(verb, legacyAcls, candidateAcls, context)`), because the space check runs inside auth's import graph and `getFeatureFlags` already lives there; the compare-and-log is inlined rather than reusing `lib/api/permissions/shadow` to avoid an `auth ↔ shadow` import cycle. The candidate is **synthesized** by the resource (a resource is now either legacy-inline or table-backed, never both).

## Risk
Very low. Purely additive telemetry, gated by a per-workspace flag that is off by default and reverts instantly. The served permission decision is untouched.

## Tests
New shadow-compare suite: logs a mismatch on divergence (no backfill), reaches parity after `reconcileGroupPermissions`, stays silent while the flag is off. front + front-api permission suites pass; `tsgo` clean.

### DIFF
--- a/front/lib/resources/space_resource.test.ts
+++ b/front/lib/resources/space_resource.test.ts
@@ -17,8 +17,10 @@ import { SpaceModel } from "@app/lib/resources/storage/models/spaces";
 import { TriggerResource } from "@app/lib/resources/trigger_resource";
 import type { UserResource } from "@app/lib/resources/user_resource";
 import { WebhookRequestResource } from "@app/lib/resources/webhook_request_resource";
+import logger from "@app/logger/logger";
 import { AgentConfigurationFactory } from "@app/tests/utils/AgentConfigurationFactory";
 import { ConversationFactory } from "@app/tests/utils/ConversationFactory";
+import { FeatureFlagFactory } from "@app/tests/utils/FeatureFlagFactory";
 import { GroupFactory } from "@app/tests/utils/GroupFactory";
 import { MembershipFactory } from "@app/tests/utils/MembershipFactory";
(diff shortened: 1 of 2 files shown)


tdraier:

- #29639 ◀

fabiencelier (approved):
LGTM !
Getting there

PR #29640, opened 2026-07-28
[front] Poke tooling for the per-user spend-cap counters

tdraier:
## Description

Poke-only admin & observability tooling layered on top of the per-user spend-cap backup (stacked on #29576). Nothing here runs on the customer or message-send path.

**Members usage table — single `Consumed (ES / RL / MT)` column.** Shows the three per-user AWU consumption figures side by side so divergence is obvious at a glance:
- **ES** — Elasticsearch-derived consumption (the table's existing source of truth).
- **RL** — the Redis fixed-window rate-limiter counter (what enforcement reads).
- **MT** — Metronome per-user usage (what reconcile and the per-user cap read).

Screenshot 2026-07-29 at 09 51 15

**"Resync Spend-Limit Counters from Usage" poke plugin** (billing role, credit-priced workspaces). Overwrites every active member's current-cycle counter from the ES-derived consumption. Use it to seed the counter after enabling the cap — it otherwise starts at 0 mid-cycle and only accrues from live messages — or to repair drift.

**`setFixedWindowCount` primitive** (`SET` + `PEXPIREAT`) — overwrite a fixed-window counter from an external source of truth (the resync's building block), as opposed to the additive `addFixedWindowCount` used for live accounting.

Adds `rateLimiterSpendAwuCredits` and `metronomeConsumedAwuCredits` to `MemberUsageType`, populated only when alert links are requested (poke); `null` everywhere else.

## Tests

- `front-api` `members-usage` endpoint test updated for the new fields — green.
- `rate_limiter` unit tests cover the fixed-window primitives (`add`/`get`/`set`).
- Manually exercised in dev via poke: ran the resync plugin and confirmed the **RL** column matches **ES** afterwards, and that the three-value column renders correctly.

## Risk

Low. Poke-only surface — no customer-facing or message-path changes. The new `MemberUsageType` fields are additive and default to `null` off the poke path. The resync plugin only `SET`s Redis counters for the current cycle, so a bad run merely rewrites the observability/enforcement counter and is fully re-runnable/idempotent. Metronome and ES read failures degrade gracefully (empty map → `0` / `-`) rather than throwing. Safe to roll back by revert.

## Deploy Plan

Standard deploy, no migration. To activate the backup for a workspace afterwards: **first** run the resync plugin to seed counters from usage, **then** enable the `enforce_user_spend_limit_rate_cap` flag (from #29576) — in that order, so users aren't measured from 0 mid-cycle.

### DIFF
--- a/front-api/routes/w/[wId]/credits/members-usage.test.ts
+++ b/front-api/routes/w/[wId]/credits/members-usage.test.ts
@@ -36,6 +36,8 @@ const MEMBER_USAGE = {
   seatBalanceAwu: 100,
   spendLimitAlertId: null,
   spendLimitAwuCredits: null,
+  rateLimiterSpendAwuCredits: null,
+  metronomeConsumedAwuCredits: null,
   spendLimitSource: "default" as const,
   spendLimitWarningAlertId: null,
 };
--- a/front/components/pages/workspace/UsagePage.tsx
+++ b/front/components/pages/workspace/UsagePage.tsx
@@ -142,6 +142,8 @@ function memberFromUpgradeRequest(
     scheduledSeatType: null,
     scheduledSeatChangeAt: null,
     spendLimitAwuCredits: null,
+    rateLimiterSpendAwuCredits: null,
+    metronomeConsumedAwuCredits: null,
     spendLimitSource: "none",
     spendLimitAlertId: null,
     spendLimitWarningAlertId: null,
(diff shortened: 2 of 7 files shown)


tdraier:

- #29575
    - #29576
        - #29640 ◀
            - #29659

fabiencelier (approved):
Clean ✅

aubin-tchoi (approved):
LGTM

PR #29642, opened 2026-07-29
[front] fix: make Chrome extension prompt non-conditional

aubin-tchoi:
## Description

For messages coming from the chrome extension (any extension actually), we alter the system prompt to add a small section hinting at the existence of platform-specific skills.

We have observed a pattern where users start a convo in the extension, then continue on the webapp, then go back, etc..., bursting the cache because of the system prompt mutation each time.

The small piece of guidance is now always present, it's moved from the ephemeral context to the shared one and it now explains the relationship between tools being available and the current message's origin. Extension messages now expose that explicit source label instead of the raw `extension` value to make it more explicit.

Also bundles a completely unrelated change: the formatting prompt for openai repeated the same title twice `# RESPONSE FORMATTING\n# Response formatting`, see [here](https://langfuse.dust.tt/project/cmi4ft2qm00061707q4a9azyc/traces?filter=metadata%3BstringObject%3BconversationId%3B%3D%3BTkyRBCtQwp&peek=002ef022893c7e2bac84b44b1def3870&timestamp=2026-07-30T12%3A25%3A17.661Z&observation=b8ef7f63fbc13fa1) for instance.

Removed it from the OpenAI-specific formatted so that we keep controlling header titles in `constructContextSection` (here specifically https://github.com/dust-tt/dust/blob/main/front/lib/api/assistant/generation.ts#L75)

## Tests

- Added tests.

## Risk

- Low.

## Deploy Plan

- Deploy front.

### DIFF
--- a/front/lib/api/assistant/generation.ts
+++ b/front/lib/api/assistant/generation.ts
@@ -72,7 +72,7 @@ function constructContextSection({
 
   const { modelConfig } = modelInfo.endpoint;
   if (modelConfig.formattingMetaPrompt && !disableFormattingPrompt) {
-    context += `# RESPONSE FORMAT\n${modelConfig.formattingMetaPrompt}\n`;
+    context += `\n# RESPONSE FORMAT\n${modelConfig.formattingMetaPrompt}\n`;
   }
 
   return context;
--- a/front/lib/api/assistant/generation.ts
+++ b/front/lib/api/assistant/generation.ts
@@ -111,7 +111,7 @@ function constructBranchContextSection({
 function constructPlatformSpecificContextSection(): string {
   return (
     "# PLATFORM-SPECIFIC CONTEXT\n\n" +
-    "When the current user message's `<dust_system>` metadata identifies its source as `Chrome extension`, " +
+    "When the current user message's `<dust_system>` metadata identifies its source as `Browser extension`, " +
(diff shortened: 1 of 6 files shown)


front/lib/api/assistant/generation.ts
```diff
@@ -109,46 +108,23 @@ function constructBranchContextSection({
...
+    "When the current user message's `<dust_system>` metadata identifies its source as `extension`, " +
+    "platform-specific tools may be available to access local, visible, or current browser context. " +
+    "Look for relevant tools before asking the user to paste that context.\n"
   );
 }
 
 function constructToolsSection({
   hasAvailableActions,
   modelInfo,
-  agentConfiguration,
```
aubin-tchoi, line 145: unrelated change: this parameter was not used

front/types/assistant/models/openai.ts
```diff
@@ -234,7 +234,7 @@ export const GPT_4O_MINI_MODEL_CONFIG: ModelConfigurationType = {
     "europe-west1": true,
   },
 };
-export const OPENAI_FORMATTING_META_PROMPT = `# Response Formats
+export const OPENAI_FORMATTING_META_PROMPT = `
```
aubin-tchoi, line 237: small unrelated change (see PR descirption), bundled here as it bursts the system cache

flvndvd (approved):

PR #29646, opened 2026-07-29
core: Fix deny-by-default Clippy lints

zmarouf:
## Description

  Fix the deny-by-default Clippy errors currently blocking the full Core lint run:

  - Use an equality check for a `usize` zero guard in the data-source splitter.
  - Use a non-constant-like floating-point value in the CSV round-trip test.
  - Remove an immediately-breaking placeholder loop from SQLite worker shutdown. (no-op loop which always exited immediately and only
  triggered Clippy’s never_loop lint)

  All changes preserve existing behavior.

  ## Tests

  - `cargo fmt --check`
  - `cargo test --lib test_csv_round_trip_bijective`
  - `cargo test --lib data_sources::splitter::tests` — 10 passed, 1 existing high-CPU test ignored
  - `cargo test --bin sqlite-worker`
  - `cargo clippy --lib --tests --message-format=short`

  ## Risk

  Low. The production-code changes are semantic no-ops.
  Test data just changes a value that's too close to pi.

### DIFF
--- a/core/bin/sqlite_worker.rs
+++ b/core/bin/sqlite_worker.rs
@@ -129,11 +129,6 @@ impl WorkerState {
 
     async fn await_pending_queries(&self) {
         // TODO: wait for all pending database queries to finish.
-        loop {
-            // if no pending query...
-            break;
-            // tokio::time::sleep(std::time::Duration::from_millis(1024)).await;
-        }
     }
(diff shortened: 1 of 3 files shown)


Fraggle (approved):
lGTM

PR #29648, opened 2026-07-29
fix(core): restrict SQLite worker queries

zmarouf:
## Description

Resolve Critical vulnerability by enforcing read-only SQL on the SQLite worker.

  The local Table Query path previously executed model-generated SQL without server-side statement
  authorization, allowing arbitrary file access through SQLite features such as `ATTACH DATABASE`,
  `VACUUM INTO`, and CSV virtual-table creation.

  This change:

  - Accepts exactly one SQL statement per query.
  - Requires the prepared statement to be read-only.
  - Enables SQLite defensive and `query_only` modes after transient tables are initialized.
  - Adds an SQLite authorizer that permits reads and safe functions while denying state-changing operations and `load_extension`.
  - Keeps existing read-only `SELECT` queries and standard SQL functions working.

  ## Tests

  - `cargo fmt --check`
  - `cargo check --lib`
  - `cargo test --lib sqlite_workers::sqlite_database::tests` all 4 passed

  The tests verify that:

  - Read-only queries remain supported.
  - CSV virtual-table file reads are rejected.
  - `ATTACH DATABASE` and `VACUUM INTO` file writes are rejected.
  - Multiple statements are rejected.

  ## Risk

  Low to moderate. 
  Existing single-statement, read-only queries remain supported. Queries that previously relied on multiple statements, pragmas, or state-changing SQLite operations will now be rejected intentionally.

  ## Deploy Plan

Deploy Core and the SQLite workers normally.

### DIFF
--- a/core/src/sqlite_workers/sqlite_database.rs
+++ b/core/src/sqlite_workers/sqlite_database.rs
@@ -11,7 +11,11 @@ use cloud_storage::Object;
 use futures::future::try_join_all;
 use parking_lot::Mutex;
 use rayon::prelude::*;
-use rusqlite::{Connection, InterruptHandle};
+use rusqlite::{
+    config::DbConfig,
+    hooks::{AuthAction, AuthContext, Authorization},
+    Batch, Connection, InterruptHandle,
+};
 use std::{collections::HashMap, io::Write, sync::Arc};
 use tempfile::NamedTempFile;
 use thiserror::Error;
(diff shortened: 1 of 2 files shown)


core/src/sqlite_workers/sqlite_database.rs
```diff
@@ -98,9 +102,29 @@ impl SqliteDatabase {
...
+                .next()
+                .map_err(|e| SqliteDatabaseError::QueryExecutionError(anyhow::Error::new(e)))?
+                .is_some()
+            {
+                return Err(SqliteDatabaseError::QueryExecutionError(anyhow!(
+                    "Query must contain a single statement"
+                )));
+            }
+
+            if !stmt.readonly() {
```
fontanierh, line 123: whaaaaat I searched for something like this at the time and couldn't find it!

fontanierh (approved):

PR #29649, opened 2026-07-29
fix(core): escape SQLite table identifiers

zmarouf:
## Description

  Prevent SQL injection through API-controlled table and column names during SQLite transient-table setup.

  Table and column names were interpolated inside double-quoted SQLite identifiers without escaping embedded `"` characters. Because the generated setup SQL was executed with `execute_batch`, a crafted identifier could terminate the quoted name and inject additional statements.

  This change:

  - Introduces a shared SQLite identifier-quoting helper that escapes `"` as `""`.
  - Applies it consistently to table and column names in `CREATE TABLE` and `INSERT` statements.
  - Escapes temporary file paths and nested schema SQL as SQLite string literals.
  - Uses a fixed internal name in the CSV module’s nested schema.
  - Replaces `execute_batch` with preparation and execution of exactly one CSV virtual-table statement.

Transient-table setup occurs before the user-query sandbox is enabled and therefore requires independent protection.

  ## Tests

  - `cargo fmt --check`
  - `cargo check --lib`
  - `cargo test --lib databases::table_schema::tests`
  - `cargo test --lib sqlite_workers::sqlite_database::tests`

  The tests verify that embedded quotes remain valid identifier content and that malicious table or column names cannot inject setup statements or create marker files.

  ## Risk

Low. Valid identifiers, including names containing quotes, remain supported through standard SQLite escaping. The changes are limited to internal SQL generation during transient-table setup.

  ## Deploy Plan

Deploy Core and the SQLite workers normally.

### DIFF
--- a/core/src/databases/table_schema.rs
+++ b/core/src/databases/table_schema.rs
@@ -63,6 +63,10 @@ impl std::fmt::Display for TableSchemaFieldType {
 pub static TABLE_SCHEMA_POSSIBLE_VALUES_MAX_LEN: usize = 32;
 pub static TABLE_SCHEMA_POSSIBLE_VALUES_MAX_COUNT: usize = 16;
 
+pub(crate) fn quote_sqlite_identifier(identifier: &str) -> String {
+    format!("\"{}\"", identifier.replace('"', "\"\""))
+}
+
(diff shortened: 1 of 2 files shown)


Fraggle (approved):
LGTM

fontanierh (approved):
Thanks !

PR #29597, opened 2026-07-28
[front] enh: add skills-only MCP server setting

aubin-tchoi:
## Overall goal

Some MCP tools rely on skill instructions for workspace-specific guidance, safety rules, or shortcuts that avoid unnecessary runtime research and credit spend. Exposing the raw tool directly lets agents bypass that context. The overall goal here is to let admins keep an MCP server available only through skills.

Part of https://github.com/dust-tt/tasks/issues/9792

## Description

This PR adds a `Restrict this tool to skills` checkbox to the tool settings sheet.
Maps to the `isRestrictedToSkills` field on the MCP server view.

image

Next PR will add a dialog to warn the user if checking it would drop the tool from existing agents (will merge both together).

## Tests

- Tested locally.

## Risk

- Low.

## Deploy Plan

- Deploy front.

PR #29654, opened 2026-07-29
[front] enh: add dialog to warn before restricting MCP tools used by agents

aubin-tchoi:
## Overall goal

Some MCP tools rely on skill instructions for workspace-specific guidance, safety rules, or shortcuts that avoid unnecessary runtime research and credit spend. Exposing the raw tool directly lets agents bypass that context. The overall goal here is to let admins keep an MCP server available only through skills.

Closes https://github.com/dust-tt/tasks/issues/9792

## Description

https://github.com/dust-tt/dust/pull/29597 added a checkbox in the UI to restrict access to a tool to skills (i.e. you can't add it to an agent nor add it directly to a convo from the input bar).
The back-end drops existing agent relations in this case.
This PR adds a confirmation dialog to mention that some agents will be affected, if any, and lists them

image

## Tests

- Tested locally.

## Risk

- Low, affects the MCP settings confirmation flow.

## Deploy Plan

- Deploy front.

### DIFF
--- a/front/components/actions/mcp/MCPServerDetails.tsx
+++ b/front/components/actions/mcp/MCPServerDetails.tsx
@@ -18,7 +18,6 @@ import { getSensitivityLabelProviderForServerId } from "@app/lib/actions/mcp_int
 import type { MCPServerViewType } from "@app/lib/api/mcp";
 import { useFeatureFlags } from "@app/lib/auth/AuthContext";
 import { clientFetch } from "@app/lib/egress/client";
-import { LinkWrapper } from "@app/lib/platform";
 import {
   useMCPServer,
   useMCPServers,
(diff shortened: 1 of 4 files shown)


front/components/actions/mcp/MCPServerDetails.tsx
```diff
@@ -141,6 +148,61 @@ export function MCPServerDetails({
...
+    }
+
+    return confirm({
+      title: "Remove this tool from agents?",
+      message: (
+        <div className="space-y-3">
+          <p className="text-sm text-muted-foreground">
+            Saving will remove the tool from the following agents:
+          </p>
+          <div className="divide-y divide-separator overflow-hidden rounded-xl border border-separator bg-background">
```
aubin-tchoi, line 170: not ideal, but I don't think we already have something that quite works for me here

id13 (approved):
Looks legit

PR #29658, opened 2026-07-29
Add skills discoverable permission 

ykmsd:
## Description
This PR is to add a new permission to make skill discoverable. By default it will be admin only so no need to run backfill script. Once a skill becomes auto-discoverable, unless you have the permission you cannot downgrade it even if you are an editor. 

We show this when a skill is auto-discoverable:
CleanShot 2026-07-29 at 13 18 19@2x


front/components/pages/workspace/governance/GovernancePage.tsx
```diff
@@ -134,7 +134,9 @@ export const GovernancePage = () => {
 
   const skillPermissions = skills.filter(
     (permission) =>
-      hasSkillPublicationFeature || permission.grantType !== "publish"
+      hasSkillPublicationFeature ||
+      (permission.grantType !== "publish" &&
+        permission.grantType !== "make_discoverable")
```
ykmsd, line 139: this is to hide the section when there is no ff

front-api/routes/w/[wId]/skills/[sId]/index.ts
Comment (fabiencelier) [inline front-api/routes/w/[wId]/skills/[sId]/index.ts:1]: You need to also update the permission check in the resource:
 - in updateSkill 
 - in updateAvailabilities
You can use the same "assert(...)" pattern

I think we should also check in `makeNew` but we did not for the 2 other permissions so maybe in another PR

fabiencelier (approved):
All good
Let's be ultra safe by also checking permissions in the resource then all good :)

front-api/routes/w/[wId]/skills/[sId]/index.ts
Comment (ykmsd) [reply front-api/routes/w/[wId]/skills/[sId]/index.ts:1]: ay ay ay! thank you, will do 🫡

Comment (ykmsd) [reply front-api/routes/w/[wId]/skills/[sId]/index.ts:1]: okay I will create a separate PR for both 🫡

matteotrab (approved):

PR #29662, opened 2026-07-29
fix(front): delete activation pod recommendations and trigger on pod deletion

ElPicador:
## Description

When a Pod is permanently deleted, `scrubSpaceActivity` cleaned up the pod's activation nudges and the `ActivationPodModel` record, but left two things behind:

- Recommendations still pointing at the pod via `activationPodId`, which is a `RESTRICT` foreign key — deleting a pod with any recommendation attached would fail with a foreign key violation. `ActivationRecommendationResource.detachActivationPod` existed to null out that link but was never actually called anywhere.
- The pod's own dedicated activation trigger, which was only implicitly cleaned up as part of the pod's webhook source view teardown, with no explicit handling.

This PR replaces `detachActivationPod` with `ActivationRecommendationResource.deleteAllForActivationPod`, which now takes the full `ActivationPodResource` and handles both cleanup steps: it deletes the recommendation rows instead of unlinking them, and deletes the pod's dedicated activation trigger (via `TriggerResource.delete`, which also tears down its Temporal workflow) — all before the pod record is removed.

## Tests

Verified with `tsgo --noEmit` and `biome check` on the changed files; no functional test added.

### DIFF
--- a/front/lib/resources/activation_recommendation_resource.test.ts
+++ b/front/lib/resources/activation_recommendation_resource.test.ts
@@ -1,16 +1,50 @@
 import { Authenticator } from "@app/lib/auth";
+import { ActivationPodResource } from "@app/lib/resources/activation_pod_resource";
 import { ActivationRecommendationResource } from "@app/lib/resources/activation_recommendation_resource";
+import type { SpaceResource } from "@app/lib/resources/space_resource";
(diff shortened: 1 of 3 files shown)


ElPicador:

- #29662 ◀

frankaloia (approved):

PR #29664, opened 2026-07-29
Add the agent message consumption attribution data model

flvndvd:
## Description

This PR introduces the PG data model for agent message consumption attribution. The broader design is
documented in the design [document](https://app.notion.com/p/dust-tt/Design-Doc-Credits-Observability-3aa28599d94180799f56c88dbd9983c4).

Dust already stores the authoritative billed value on `AgentMessage.costCredits`. That value remains the source of truth for billing, limits, Metronome events, fair-use accounting, and customer-facing totals. Consumption attribution is a separate explanatory layer that describes which recognizable parts of an execution contributed to that bill.

This PR deliberately contains only the DB schema. It does not materialize attribution, plumbing comes in a wave of follow up PRs.

## Intended write lifecycle

A tool attribution row must be created as soon as the model-emitted action exists, before the tool is executed.

>[!IMPORTANT]
> This ordering is important for tools that require approval. When the model emits such a tool call, the agent workflow creates the `AgentMCPAction`, marks the conversation as requiring action, and terminates the current workflow execution. The tool result does not exist yet, and execution may resume much later in a new workflow after the user approves or denies the action.

The first attribution pass will therefore create a pending tool row linked to both the emitting `RunUsage`, when available, and the `AgentMCPAction`. At that point, the row can capture the estimated model-output footprint of emitting the tool name and parameters. The row remains incomplete through `completedAt = NULL`.

When the tool eventually runs, a later attribution pass finds the same deterministic row and enriches it with the estimated input footprint of the result produced by the tool, the exact direct tool credit charge when one exists, and the final gross attributed credit amount. It then sets `completedAt`.

If the user denies the action, the same row can be completed without a tool-result footprint or direct execution charge. If execution is cancelled or evidence remains unavailable, the row stays incomplete. Readers can then return the exact billed total while withholding an incomplete detailed explanation.

A tool attribution row is created once and identified deterministically. Later Temporal executions may complete that same row as new evidence becomes available, but they never replace it or create another row for the same tool action. Once completed, the row is immutable.

## Data ownership

- `AgentMessage` continues to own the exact number of credits billed for the message.
- `RunUsage` continues to own provider-reported model usage, including provider and model identity, prompt tokens,
completion tokens, reasoning tokens, cache usage, batch status, and the model cost calculated from those facts.
- `AgentMCPAction` continues to own tool identity, execution status, configuration, and action provenance.
- `AgentMessageConsumptionItem` owns only the versioned semantic attribution of that existing usage to recognizable
work.

The new table intentionally does not copy provider identity, model identity, tool identity, tool arguments, tool results, prompts, or rendered model requests. Those facts remain available through their owning records.

## Item semantics

Each row represents one semantic contribution to an agent message. Supported item types are `system`, `input`, `output`, `reasoning`, and `tool`.

An `input` row represents model-input work assigned to general request and conversation processing. A system row is reserved for attribution versions that separately measure system instructions and tool definitions. First iteration is expected to fold those two elements into `input`.

An `output` row represents ordinary assistant `output`. A `reasoning` row represents `reasoning` tokens when the provider reports them separately.

A `tool` row represents one bill-relevant tool execution and is linked to exactly one `AgentMCPAction`. It combines all recognizable contributions associated with that execution into one product-level item.

On a `tool` row, `outputTokensCount` means the estimated model-output footprint used to emit the tool name and its parameters. **It does not mean the result returned by the tool.**

On a `tool` row, `inputTokensCount` means the estimated model-input footprint of the result produced by the tool execution. **It does not mean the tool parameters**.

This distinction follows the model boundary. Parameters are generated by the model and therefore belong to its output side. Tool results become potential model context and therefore belong to the input side.

A tool result footprint may exist even when no later model request consumes the result. It explains the compound footprint produced by that action. Later model calls independently record the input that was actually rendered for them.

## Credit fields and cache effects

`grossAttributedCreditAmountMicro` stores the estimated token footprint converted to credits without applying
cache effects. Materializing this value ensures PG reads and future Elasticsearch rebuilds do not need to rerun historical token pricing.

For tool rows, the gross amount includes both attributed model work and the exact direct execution charge when one exists.

`directCreditAmountMicro` stores that direct tool charge separately. This preserves the historical amount actually used when billing the message and lets analytics distinguish direct execution credits from attributed model work.

The attribution table does not store cache adjustments. Cache effectiveness is a message-level explanation derived by comparing the sum of gross attributed credits with the authoritative `AgentMessage.costCredits`. Attribution remains cache-naive, while the billed value reflects the actual cache behavior.

Missing evidence is represented by `NULL`, not zero. Zero means the contribution was measured and found to be empty. `NULL` means it was not observed or cannot yet be established.

## Idempotency and concurrency

`itemKey` provides deterministic identity within an agent message and attribution version. Future materializers
will use keys shaped as `run-usage:{runUsageId}:input`, `run-usage:{runUsageId}:output`, `run-usage:
{runUsageId}:reasoning`, and `tool-action:{agentMCPActionId}`.

The unique message, version, and item-key index makes repeated materialization safe across Temporal retries.

## Deploy Plan

- [ ] Apply SQL migration
- [ ] Deploy front

### DIFF
--- a/front/lib/api/assistant/conversation/destroy.ts
+++ b/front/lib/api/assistant/conversation/destroy.ts
@@ -2,6 +2,7 @@ import { hardDeleteDataSource } from "@app/lib/api/data_sources";
 import { deleteOwnerPolicy } from "@app/lib/api/sandbox/egress_policy";
 import type { Authenticator } from "@app/lib/auth";
 import { getPrivateUploadBucket } from "@app/lib/file_storage";
+import { AgentMessageConsumptionItemModel } from "@app/lib/models/agent/agent_message_consumption_item";
 import { AgentSuggestionModel } from "@app/lib/models/agent/agent_suggestion";
 import {
   AgentMessageFeedbackModel,
@@ -228,6 +229,13 @@ export async function destroyConversation(
       const contentFragmentIds = removeNulls(
         messagesChunk.map((m) => m.contentFragmentId)
       );
+
+      await AgentMessageConsumptionItemModel.destroy({
+        where: {
+          workspaceId: owner.id,
+          agentMessageId: agentMessageIds,
+        },
+      });
 
       await destroyActionsRelatedResources(auth, agentMessageIds);
 
--- a/front/lib/models/agent/agent_message_consumption_item.ts
+++ b/front/lib/models/agent/agent_message_consumption_item.ts
@@ -183,11 +183,6 @@ AgentMessageConsumptionItemModel.init(
       },
       {
         concurrently: true,
-        fields: ["workspaceId", "conversationId", "agentMessageId"],
-        name: "agent_message_consumption_items_conversation_message",
-      },
-      {
-        concurrently: true,
         fields: ["conversationId"],
       },
       {
@@ -238,23 +233,5 @@ AgentMessageModel.hasMany(AgentMessageConsumptionItemModel, {
 });
 AgentMessageConsumptionItemModel.belongsTo(AgentMessageModel, {
   foreignKey: { name: "agentMessageId", allowNull: false },
-  onDelete: "RESTRICT",
-});
-
-RunUsageModel.hasMany(AgentMessageConsumptionItemModel, {
-  foreignKey: { name: "runUsageId", allowNull: true },
-  onDelete: "RESTRICT",
-});
-AgentMessageConsumptionItemModel.belongsTo(RunUsageModel, {
-  foreignKey: { name: "runUsageId", allowNull: true },
-  onDelete: "RESTRICT",
-});
-
-AgentMCPActionModel.hasMany(AgentMessageConsumptionItemModel, {
-  foreignKey: { name: "agentMCPActionId", allowNull: true },
-  onDelete: "RESTRICT",
-});
-AgentMessageConsumptionItemModel.belongsTo(AgentMCPActionModel, {
-  foreignKey: { name: "agentMCPActionId", allowNull: true },
   onDelete: "RESTRICT",
 });
--- a/front/lib/models/agent/agent_message_consumption_item.ts
+++ b/front/lib/models/agent/agent_message_consumption_item.ts
@@ -1,0 +1,254 @@
+import { AgentMCPActionModel } from "@app/lib/models/agent/actions/mcp";
+import {
+  AgentMessageModel,
+  ConversationModel,
+} from "@app/lib/models/agent/conversation";
+import { frontSequelize } from "@app/lib/resources/storage";
+import { DataTypes } from "@app/lib/resources/storage/data_types";
+import { RunUsageModel } from "@app/lib/resources/storage/models/runs";
+import { WorkspaceAwareModel } from "@app/lib/resources/storage/wrappers/workspace_models";
+import {
+  AGENT_MESSAGE_CONSUMPTION_ITEM_TYPES,
+  type AgentMessageConsumptionItemType,
+} from "@app/types/assistant/agent_message_consumption";
+import type { CreationOptional, ForeignKey } from "sequelize";
+import { Op } from "sequelize";
+
+function validateConsumptionItemShape(
+  this: AgentMessageConsumptionItemModel
+): void {
+  const isTool = this.itemType === "tool";
+
+  if (isTool && this.agentMCPActionId === null) {
+    throw new Error("Tool attribution items require an agent MCP action");
+  }
+  if (!isTool && this.runUsageId === null) {
+    throw new Error("Non-tool attribution items require a run usage");
+  }
+  if (!isTool && this.agentMCPActionId !== null) {
+    throw new Error("Only tool attribution items may reference an action");
+  }
+  if (!isTool && this.directCreditAmountMicro !== null) {
+    throw new Error("Only tool attribution items may contain direct credits");
+  }
+  if (
+    this.directCreditAmountMicro !== null &&
+    this.grossAttributedCreditAmountMicro < this.directCreditAmountMicro
+  ) {
+    throw new Error("Gross attributed credits cannot be below direct credits");
+  }
+
+  switch (this.itemType) {
+    case "system":
+    case "input":
+      if (this.outputTokensCount !== null) {
+        throw new Error(`${this.itemType} items cannot contain output tokens`);
+      }
+      break;
+    case "output":
+    case "reasoning":
+      if (this.inputTokensCount !== null) {
+        throw new Error(`${this.itemType} items cannot contain input tokens`);
+      }
+      break;
+    case "tool":
+      break;
+  }
+}
+
+export class AgentMessageConsumptionItemModel extends WorkspaceAwareModel<AgentMessageConsumptionItemModel> {
+  declare createdAt: CreationOptional<Date>;
+  declare updatedAt: CreationOptional<Date>;
+
+  declare conversationId: ForeignKey<ConversationModel["id"]>;
+  declare agentMessageId: ForeignKey<AgentMessageModel["id"]>;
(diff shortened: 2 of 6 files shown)


front/lib/models/agent/agent_message_consumption_item.ts
```diff
@@ -0,0 +1,254 @@
...
+      allowNull: true,
+    },
+  },
+  {
+    sequelize: frontSequelize,
+    modelName: "agent_message_consumption_item",
+    tableName: "agent_message_consumption_items",
+    indexes: [
+      {
+        unique: true,
```
flvndvd, line 165: Enforcing uniqueness for the `itemKey`.

front/lib/models/agent/agent_message_consumption_item.ts
```diff
@@ -0,0 +1,260 @@
...
+      },
+      {
+        concurrently: true,
+        fields: ["conversationId"],
+      },
+      {
+        concurrently: true,
+        fields: ["agentMessageId"],
+      },
+      {
```
flvndvd: Required for a smooth deletion when run or agent is being deleted.

front/lib/models/agent/agent_message_consumption_item.ts
```diff
@@ -0,0 +1,260 @@
...
+        unique: true,
+        concurrently: true,
+        fields: ["workspaceId", "attributionVersion", "agentMCPActionId"],
+        where: { agentMCPActionId: { [Op.ne]: null } },
+        name: "agent_message_consumption_items_unique_action",
+      },
+      {
+        unique: true,
```
flvndvd, line 191: Guarantees one input, output, or reasoning row per run usage and version. The partial condition deliberately excludes tool rows because one model call can emit several tools.

front/lib/models/agent/agent_message_consumption_item.ts
```diff
@@ -0,0 +1,260 @@
...
+      {
+        concurrently: true,
+        fields: ["runUsageId"],
+      },
+      {
+        concurrently: true,
+        fields: ["agentMCPActionId"],
+      },
+      {
+        unique: true,
```
flvndvd, line 184: Guarantees one tool attribution row per action and version, including across retries or resumed workflows.

front/lib/api/assistant/conversation/destroy.ts
```diff
@@ -229,6 +230,13 @@ export async function destroyConversation(
         messagesChunk.map((m) => m.contentFragmentId)
       );
 
+      await AgentMessageConsumptionItemModel.destroy({
```
flvndvd, line 233: Required to please the existing test.

front/lib/models/agent/agent_message_consumption_item.ts
```diff
@@ -0,0 +1,260 @@
...
+          "attributionVersion",
+          "itemKey",
+        ],
+        name: "agent_message_consumption_items_message_version_key",
+      },
+      {
+        concurrently: true,
+        fields: ["workspaceId", "conversationId", "agentMessageId"],
+        name: "agent_message_consumption_items_conversation_message",
+      },
```
aubin-tchoi: are we sure we need this index? since an agent message only belongs to 1 convo, and both only belong to 1 workspace I think we're fine with only the index on `agentMessageId`

front/migrations/pre-deploy/20260728223702_add_agent_message_consumption_items.sql
```diff
@@ -0,0 +1,122 @@
...
+	CONSTRAINT "agent_message_consumption_items_input_tokens_check" CHECK ("inputTokensCount" IS NULL OR "inputTokensCount" >= 0),
+	CONSTRAINT "agent_message_consumption_items_output_tokens_check" CHECK ("outputTokensCount" IS NULL OR "outputTokensCount" >= 0),
+	CONSTRAINT "agent_message_consumption_items_gross_credits_check" CHECK ("grossAttributedCreditAmountMicro" >= 0),
+	CONSTRAINT "agent_message_consumption_items_direct_credits_check" CHECK ("directCreditAmountMicro" IS NULL OR "directCreditAmountMicro" >= 0),
+	CONSTRAINT "agent_message_consumption_items_reference_shape_check" CHECK (("itemType" = 'tool' AND "agentMCPActionId" IS NOT NULL) OR ("itemType" <> 'tool' AND "runUsageId" IS NOT NULL AND "agentMCPActionId" IS NULL)),
+	CONSTRAINT "agent_message_consumption_items_direct_shape_check" CHECK ("directCreditAmountMicro" IS NULL OR ("itemType" = 'tool' AND "grossAttributedCreditAmountMicro" >= "directCreditAmountMicro")),
+	CONSTRAINT "agent_message_consumption_items_token_shape_check" CHECK (("itemType" IN ('system', 'input') AND "outputTokensCount" IS NULL) OR ("itemType" IN ('output', 'reasoning') AND "inputTokensCount" IS NULL) OR "itemType" = 'tool')
```
aubin-tchoi: are we sure we want these constraints in db?

aubin-tchoi (approved):
LGTM
wondering whether we need FKs everywhere, probably fine to leave some stale unreachable data in there

front/migrations/pre-deploy/20260728223702_add_agent_message_consumption_items.sql
flvndvd: Nope forgot to remove those.

id13:
> A tool attribution row must be created as soon as the model-emitted action exists, before the tool is executed.

I don't understand why. 

IMO we should record attribution when it is expended, not before, and respect a plain append-only, this ages much better, is a well understood and widely adopted pattern.

front/lib/models/agent/agent_message_consumption_item.ts
id13: Agreed, and let's create a dependent statistic on workspaceId, conversationId, agentMessageId to hint PG planner

front/lib/models/agent/agent_message_consumption_item.ts
```diff
@@ -0,0 +1,260 @@
...
+});
+
+AgentMCPActionModel.hasMany(AgentMessageConsumptionItemModel, {
+  foreignKey: { name: "agentMCPActionId", allowNull: true },
+  onDelete: "RESTRICT",
+});
+AgentMessageConsumptionItemModel.belongsTo(AgentMCPActionModel, {
+  foreignKey: { name: "agentMCPActionId", allowNull: true },
+  onDelete: "RESTRICT",
+});
```
id13: I think all of this is not necessary even if we don't do an append-only, detrimental performance-wise, and not compatible with an append only.

flvndvd: Ack, but we will still delete all those rows when associated resources are deleted, to me it's still a relational table.

flvndvd:
> > A tool attribution row must be created as soon as the model-emitted action exists, before the tool is executed.
> 
> I don't understand why.
> 
> IMO we should record attribution when it is expended, not before, and respect a plain append-only, this ages much better, is a well understood and widely adopted pattern.

Apologies, PR description was not clear on this, common case is append only, we insert and never touch again, there is an exception though when a tool require an approval (which might never happen) we still want to attribute the model output tokens to the tool params. In this case we write first, and update once tool has been approved and run.

front/lib/models/agent/agent_message_consumption_item.ts
id13: > we will still delete all those rows when associated resources are deleted

I think we should only delete on workspace scrub and when crossing a certain retention policy threshold (tbd, future us problem)

PR #29673, opened 2026-07-29
Fix poke scrubSpaceActivity by detaching activation recommandation before deleting pod

fabiencelier:
## Description

The scrubSpaceWorkflow was failing because scrubSpaceActivity hard-deleted the activation_pods row while activation_recommendations.activationPodId still pointed at it.

This should fix the stuck activity

## Tests

not tested

## Risk

low

## Deploy Plan

deploy front


ElPicador (approved):
Thanks! I'm working on a better fix, let's merge yours first

PR #29676, opened 2026-07-29
refactor: align client MCP tool definitions with internal ones

aubin-tchoi:
## Description

Follow-up on [#29115](https://github.com/dust-tt/dust/pull/29115), which moved internal MCP tool metadata from helper-built records to explicit arrays and the shared `ToolHandlers` / `buildTools` pattern.

Client tools still used the old pattern with `createClientToolsRecord` and quite a few specific (almost duplicated) types and heplers.

The only difference between internal and client is that internal tool handlers are always passed an auth and a context.

This does not require re-publishing the extension. No expected behavior change and has almost no runtime change. The changes in front are only typing.

## Tests

- Covered by existing tests.

## Risk

- Low.

## Deploy Plan

- Deploy front.

### DIFF
--- a/extension/shared/tools/metadata.ts
+++ b/extension/shared/tools/metadata.ts
@@ -213,7 +213,7 @@ type_text automatically focuses the element before typing.
 delete_text automatically focuses the element before deleting the text.
 For the above reasons in most cases you do NOT need to click an element before calling type_text or delete_text.
 Avoid unnecessary actions.`,
-    schema: z.object({
+    schema: {
       action: z
         .enum(["get_elements", "click_element", "type_text", "delete_text"])
         .describe("Action to perform."),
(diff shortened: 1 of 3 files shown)


extension/shared/tools/metadata.ts
matteotrab, line 88: I think that by removing the `createClientToolRecord` function, here we could write any string and typescript will not complain even if the string is not an argument of the tool

matteotrab (approved):
Left a comment

extension/shared/tools/metadata.ts
aubin-tchoi: I think it should crash in `buildTools`, which expects the first argument to be match the `ToolMeta` type, where it's typed as `ReadonlyArray<Extract<keyof z.infer<z.ZodObject<TSchema>>, string>>;` but let me double check

I'll add a `satisfies` as well here to make sure

aubin-tchoi: ah `TSchema` in `ToolDefinition` was actually not tied to the actual zod schema (shape), fixed (added a validating type in `buildTools`)

PR #29677, opened 2026-07-29
[sparkle] fix: keep checkbox entirely filled during animation

aubin-tchoi:
## Description

Checking a checkbox briefly left an empty gap between its border and the fill during the animation.

This PR fixes that by keeping the indicator mounted and full-sized, then animate only the check or partial-state icon from the indicator state. This also prevents the animation from replaying when a selected control mounts again.

The general approach is to not render conditionally and always render everything to make sure animations happen smoothly.

Struggling a bit to capture a gif, can be checked out using the preview.

## Tests

- Updated the story.

## Risk

- Low.

## Deploy Plan

- Deploy front.


ykmsd (approved):
I trust our new design engineer 🫡

PR #29678, opened 2026-07-29
Add workspace user identity to conversation Frames

fontanierh:
## Description

Stack 1 of 6, followed by #29679.

Adds `useUserIdentity()` to conversation Frames through the existing RPC channel. The host returns a sanitized user only for conversation access, an active role, and an exact workspace match. Older hosts fail closed after an identity-only five-second timeout.

## Tests

Added hook and RPC coverage for active, revoked, anonymous, and cross-workspace callers, plus cache lifecycle and old-host timeout cleanup.

## Risk

The RPC is additive. A temporary Front/Viz version mismatch returns no identity after the timeout, so rollback is safe.

## Deploy Plan

1. Deploy Front so the Frame host serves the identity RPC.
2. Deploy Viz so Frame code receives `useUserIdentity()`.
3. Smoke-test authenticated and anonymous conversation Frames.

### DIFF
--- a/front/components/assistant/conversation/actions/VisualizationActionIframe.test.ts
+++ b/front/components/assistant/conversation/actions/VisualizationActionIframe.test.ts
@@ -1,0 +1,50 @@
+import {
+  type FrameAccess,
+  getConversationFrameUserIdentity,
+} from "@app/components/assistant/conversation/actions/VisualizationActionIframe";
+import { describe, expect, it } from "vitest";
+
+const authContext = {
+  workspace: { sId: "w_current" },
+  user: {
+    sId: "usr_123",
+    firstName: "Ada",
+    lastName: "Lovelace",
+    fullName: "Ada Lovelace",
+    image: null,
+  },
+};
+
+function getIdentity(
+  workspaceId: string,
+  frameAccess: FrameAccess = "conversation"
+) {
+  return getConversationFrameUserIdentity(
+    authContext,
+    frameAccess,
+    workspaceId
+  );
+}
+
+describe("getConversationFrameUserIdentity", () => {
+  it("returns identity for the Frame workspace", () => {
+    expect(getIdentity("w_current")).toEqual({
+      isAuthenticated: true,
+      user: authContext.user,
+    });
+  });
+
+  it("does not return identity for another workspace", () => {
+    expect(getIdentity("w_other")).toEqual({
+      isAuthenticated: false,
+      user: null,
+    });
+  });
+
+  it("keeps public Frames unauthenticated", () => {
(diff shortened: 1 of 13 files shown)


aubin-tchoi (approved):
LGTM

front/components/assistant/conversation/actions/VisualizationActionIframe.test.ts
```diff
@@ -0,0 +1,70 @@
+import {
+  type FrameAccess,
+  getConversationFrameUserIdentity,
+} from "@app/components/assistant/conversation/actions/VisualizationActionIframe";
+import { describe, expect, it } from "vitest";
+
+const authContext = {
```
aubin-tchoi, line 7: surprised we don't already have smth for this

front/components/assistant/conversation/actions/VisualizationActionIframe.tsx
```diff
@@ -76,6 +80,43 @@ export type Visualization = PublicVisualization | ProtectedVisualization;
 
 export type FrameAccess = "conversation" | "public-anonymous" | "public-member";
 
+interface WorkspaceAuthIdentity {
+  user: NonNullable<UserIdentityState["user"]>;
+  workspace: { role: RoleType; sId: string };
```
aubin-tchoi, line 85: OOC why not the whole type here?

aubin-tchoi (approved):

front/components/assistant/conversation/actions/VisualizationActionIframe.tsx
flvndvd: ☝️ 👋

front/components/assistant/conversation/actions/VisualizationActionIframe.tsx
```diff
@@ -76,6 +80,43 @@ export type Visualization = PublicVisualization | ProtectedVisualization;
...
+}
+
+export function getConversationFrameUserIdentity(
+  authContext: WorkspaceAuthIdentity | null,
+  frameAccess: FrameAccess,
+  workspaceId: string
```
flvndvd, line 91: `nit` can we rename so it's clearer here that `workspaceId` is the actual frame's workspace while `authContext` is the current user.

viz/app/lib/pod-function-hooks.tsx
```diff
@@ -133,6 +139,37 @@ export function usePodFunction(
   };
 }
 
+export function useUserIdentity(): UseUserIdentityResult {
```
flvndvd, line 142: Is this only available in the case of pod functions?

PR #29680, opened 2026-07-29
Gong connector: treat 'Company is inactive' error as invalid credentials error to stop the connector

fabiencelier:
## Description

We have a connector failing in loop with
```
Error: Gong API responded with status: 401 on /calls/transcript
```
Due to a 401 `Company is inactive`

We should no retry in that case so I have added this error to the GONG_INVALID_CREDENTIALS_MESSAGES to pause the connector with an `oauth_token_revoked` error.

## Tests

not tested

## Deploy Plan

deploy connector


aubin-tchoi (approved):
LGTM

PR #29686, opened 2026-07-29
Make Pod Function worker rollouts safe

fontanierh:
## Description

- Mark Pod Function invocations as errored when their workflow fails before returning a structured result.
- Skip sandbox execution when a redelivered invocation is already terminal.
- Fetch the Pod Function and invocation through the original caller's access-controlled resource path before terminalizing it.
- Move new workflows to `sandbox-functions-queue-v2` while keeping a single worker that polls only v2.

Related infrastructure change: https://github.com/dust-tt/dust-infra/pull/948

## Tests

- Added focused failure-terminalization, terminal redelivery, and access-control tests.
- Passed 23 focused tests and the front typecheck.
- Built all 29 Temporal worker bundles locally.

## Risk

The worker no longer polls `sandbox-functions-queue-v1`. In-flight v1 workflows must be manually drained or terminated before deployment so they are not stranded when the worker switches to v2.

The status update is conditional and can only transition an invocation from `created` to `errored`, so late cleanup cannot overwrite a successful result.

## Deploy Plan

1. Manually drain or terminate all in-flight workflows on `sandbox-functions-queue-v1`.
2. Confirm that v1 has no pending workflows or tasks.
3. Deploy this PR. New workflows and the worker will use `sandbox-functions-queue-v2`; no application worker will poll v1.
4. Verify failed workflows move their invocation out of `created` and new v2 invocations complete normally.
5. Deploy the related infrastructure PR after this application change is healthy.


front/temporal/sandbox_functions/workflows.ts
```diff
@@ -33,6 +34,17 @@ const { runSandboxFunctionInvocationActivity } = proxyActivities<
...
 });
 
+const { markSandboxFunctionInvocationFailedActivity } = proxyActivities<
+  typeof markSandboxFunctionInvocationFailedActivities
+>({
+  startToCloseTimeout: "1 minute",
+  retry: {
+    // This only changes `created` to `errored`, so retries cannot overwrite a
+    // successful invocation.
+    maximumAttempts: 5,
```
flvndvd, line 44: Are we sure it's safe to retry a sandbox function? Wondering if it would not be safer to let the caller retry.

front/temporal/sandbox_functions/config.ts
```diff
@@ -1,3 +1,5 @@
-const QUEUE_VERSION = 1;
+const QUEUE_VERSION = 2;
+const LEGACY_QUEUE_VERSION = 1;
 
 export const QUEUE_NAME = `sandbox-functions-queue-v${QUEUE_VERSION}`;
```
flvndvd, line 3: We are doing this new queue logic only because we are changing the workflow code? If so, given that it's only internal usage could we just skip it keep exisiting queue but just terminate.

front/temporal/sandbox_functions/config.test.ts
```diff
@@ -0,0 +1,12 @@
+import {
+  LEGACY_QUEUE_NAME,
+  QUEUE_NAME,
+} from "@app/temporal/sandbox_functions/config";
+import { describe, expect, it } from "vitest";
+
+describe("sandbox function queues", () => {
+  it("moves new workflows to a new queue while preserving the previous queue", () => {
+    expect(QUEUE_NAME).toBe("sandbox-functions-queue-v2");
```
flvndvd: 😆

front/lib/resources/sandbox_function_invocation_resource.ts
```diff
@@ -513,6 +513,57 @@ export class SandboxFunctionInvocationResource extends BaseResource<SandboxFunct
     return new Ok(invocation);
   }
 
+  static async markCreatedAsErrored(
```
flvndvd: Looks like we don't check for the status when calling `execute`

flvndvd: Also can we keep this one as a class method and fetch before calling it. Here we kind of open a door that skips most of the permission like checking you have access to the pod.

front/temporal/sandbox_functions/workflows.ts
fontanierh: We're not retrying the sandbox function. We're retrying the activity that marks it as failed.

front/temporal/sandbox_functions/config.test.ts
fontanierh: You still read the tests ? 🙃

front/temporal/sandbox_functions/config.ts
fontanierh: yeah I can just do it more violently I agree

front/lib/resources/sandbox_function_invocation_resource.ts
fontanierh: Sure

fontanierh: Addressed both points:

- `execute` now treats an already `succeeded` or `errored` invocation as an idempotent no-op, so redelivery cannot rerun side effects or overwrite a terminal result.
- The failure activity now rebuilds the original authenticator, fetches the Pod Function and invocation through the normal access-controlled resource path, then calls `markCreatedAsErrored` as an instance method.
- Added coverage for both terminal statuses and for a same-workspace user who cannot access the pod.

Focused result: 21 tests pass and the front typecheck is clean.

front/temporal/sandbox_functions/workflows.ts
flvndvd: Completely missed it 🙈

front/temporal/sandbox_functions/config.ts
fontanierh: Addressed in `6b962696dd`. The worker stays on `sandbox-functions-queue-v1`; I removed the legacy queue export, the second worker, and the split concurrency. We will manually drain or terminate in-flight workflows before deploying the workflow change on the existing queue.

front/temporal/sandbox_functions/config.test.ts
fontanierh: Removed this test together with the dual-queue implementation in `6b962696dd`. There is no longer a hard-coded queue migration assertion.

front/lib/resources/sandbox_function_invocation_resource.ts
```diff
@@ -513,6 +526,40 @@ export class SandboxFunctionInvocationResource extends BaseResource<SandboxFunct
...
 
+  async markCreatedAsErrored(
+    error: SandboxFunctionCallError
+  ): Promise<boolean> {
+    const [updatedCount] = await this.model.update(
+      { status: "errored" },
+      {
+        where: {
+          id: this.id,
+          sandboxFunctionId: this.sandboxFunction.id,
```
flvndvd: Given that this now a class method, completely fine to just call `this.update()`, we don't need the more complex where clause.

flvndvd (approved):

front/temporal/sandbox_functions/config.ts
fontanierh: Correction after clarifying the rollout: `d9dbbd13fd` restores the v2 queue bump, but the worker remains single-queue and polls only v2. There is still no legacy poller or split concurrency. We will manually drain or terminate v1 before deployment.

front/lib/resources/sandbox_function_invocation_resource.ts
fontanierh: Addressed in `bf678dd585`. `markCreatedAsErrored` now checks the resource status and uses `this.update({ status: "errored" })`. The existing terminal-state test still verifies that a succeeded invocation is left unchanged.

PR #29688, opened 2026-07-29
Return Pod Function results to the worker

fontanierh:
## Description

- Return structured Pod Function results on stdout to the owning worker.
- Parse and persist the result in the worker without an extra HTTP callback.
- Keep the callback endpoint and add rollout compatibility for old workers and old runner images.
- Share result-envelope parsing between direct results and the callback route.

## Tests

- Passed 18 focused invocation resource tests and 6 result endpoint tests.
- Passed front and front API typechecks and the Swagger annotation lint.
- Passed all 265 Rust tests.

## Risk

The new stdout transport is enabled only when the front worker sets its environment flag. New runners still use the callback with old workers, and new workers accept callback completion from old runners.

## Deploy Plan

The front worker and runner image can roll in either order because both mixed-version paths are covered. Keep the callback endpoint until old runner images have drained.

### DIFF
--- a/cli/dust-sandbox/src/commands/function/run.rs
+++ b/cli/dust-sandbox/src/commands/function/run.rs
@@ -4,14 +4,18 @@ use super::{emit_error, spawn_function};
 use crate::api::DustApiClient;
 
 const SANDBOX_TOKEN_ENV: &str = "DUST_SANDBOX_TOKEN";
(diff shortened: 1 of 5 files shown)


fontanierh:
Closing this approach: production data shows the callback has measurable latency, but most of its work (result persistence, status update, and event publication) remains on the direct path. Without an A/B measurement of completion-to-result-event latency, the net benefit is unproven, while the mixed-version transport and retained callback add rollout complexity.

PR #29694, opened 2026-07-29
Persist conversation goal lifecycle state

fontanierh:
## Description

Add the additive `conversation_goals` schema and a Resource-owned root-goal lifecycle. Goal creation validates the referenced agent message against the authenticated workspace and conversation, while conversation destruction removes goal rows before restrictive foreign keys are reached. Branch-aware lifecycle support follows in the next stack PR.

## Tests

Covered creation and retrieval through the Resource, full conversation destruction with goal state present, and the repository conversation-relationship detector.

## Risk

This slice adds persistence and cleanup but has no production goal-creation call site yet. All foreign keys are restrictive and indexed, and unfinished-root uniqueness is enforced in Postgres.

## Deploy Plan

1. Apply the pre-deploy migration.
2. Deploy this application commit.
3. Merge the next stack PR only after this deploy is complete.


front/migrations/pre-deploy/20260730212040_add_conversation_goals.sql
```diff
@@ -0,0 +1,189 @@
+/*
+Statement 0
+*/
+SET SESSION statement_timeout = 3000;
+SET SESSION lock_timeout = 3000;
+CREATE SEQUENCE "public"."conversation_goals_id_seq"
+	AS bigint
+	INCREMENT BY 1
+	MINVALUE 1 MAXVALUE 9223372036854775807
```
aubin-tchoi, line 9: minvalue is not equal to 1 in EU

front/lib/models/agent/conversation_goal.ts
```diff
@@ -0,0 +1,183 @@
...
+  declare lastAgentMessageId: ForeignKey<AgentMessageModel["id"]> | null;
+
+  declare conversation?: NonAttribute<ConversationModel>;
+  declare branch?: NonAttribute<ConversationBranchModel | null>;
+  declare createdByUser?: NonAttribute<UserModel>;
+  declare currentAgentMessage?: NonAttribute<AgentMessageModel>;
+  declare lastAgentMessage?: NonAttribute<AgentMessageModel | null>;
+}
+
+ConversationGoalModel.init(
```
aubin-tchoi, line 35: IIUC you're 99% of the time going to need the latest, I'd optimize for this here
you may also not even need the full history, in which case you could get away with only storing the latest value and updating it

front/lib/models/agent/conversation_goal.ts
```diff
@@ -0,0 +1,183 @@
...
+    },
+    objective: {
+      type: DANGEROUSLY_UNBOUNDED_TEXT,
+      allowNull: false,
+    },
+    status: {
+      type: DataTypes.STRING,
+      allowNull: false,
+    },
+    agentConfigurationId: {
```
aubin-tchoi: if, for some reason I switch agents within a convo, the new one will not pursue the main goal?

front/lib/models/agent/conversation_goal.ts
```diff
@@ -0,0 +1,183 @@
...
+    turnCount: {
+      type: DataTypes.INTEGER,
+      allowNull: false,
+      defaultValue: 1,
+    },
+    maxTurns: {
+      type: DataTypes.INTEGER,
+      allowNull: false,
+    },
+    reason: {
```
aubin-tchoi: what is `reason`? can't we consolidate with the status?

front/lib/models/agent/conversation_goal.ts
```diff
@@ -0,0 +1,183 @@
...
+    agentConfigurationId: {
+      type: DataTypes.STRING,
+      allowNull: false,
+    },
+    turnCount: {
+      type: DataTypes.INTEGER,
+      allowNull: false,
+      defaultValue: 1,
+    },
+    maxTurns: {
```
aubin-tchoi: shouldn't this be counted in tokens rather than turns?

front/lib/models/agent/conversation_goal.ts
```diff
@@ -0,0 +1,183 @@
...
+    },
+    status: {
+      type: DataTypes.STRING,
+      allowNull: false,
+    },
+    agentConfigurationId: {
+      type: DataTypes.STRING,
+      allowNull: false,
+    },
+    turnCount: {
```
aubin-tchoi: same, counting the number of token consumed could be useful

front/lib/models/agent/conversation_goal.ts
aubin-tchoi: there's a world where you can add a bunch of columns on `conversations` also

front/migrations/pre-deploy/20260730212040_add_conversation_goals.sql
fontanierh: this is a generated migration 🤨

fontanierh: let me try again

front/lib/models/agent/conversation_goal.ts
fontanierh: >there's a world where you can add a bunch of columns on conversations also

don't think that's a world we want to live in

front/lib/models/agent/conversation_goal.ts
fontanierh: I was going for simplicity but maybe credits is an even better metric

front/lib/models/agent/conversation_goal.ts
aubin-tchoi: if it's 3 small columns it's probably fine

aubin-tchoi: ah but nvm you have the objective anyway

fontanierh: you could have multiple goals in one conversation i think (but only 1 active)

front/migrations/pre-deploy/20260730212040_add_conversation_goals.sql
aubin-tchoi: wait actually this is fine, confirmed with @flvndvd we have a trigger that will handle this

aubin-tchoi: sorry about this one

front/lib/models/agent/conversation_goal.ts
fontanierh: Removed that

front/lib/models/agent/conversation_goal.ts
fontanierh: I removed that entirely for now. Deferring to later, but will likely go for credits, which will be a bit involved (require a join table and a way to compute credits, including sub agent credits etc)

front/lib/models/agent/conversation_goal.ts
fontanierh: same

aubin-tchoi (approved):
LGTM on what's stored in the table, will let our data model committee take over on the debate of how to store it!

front/lib/models/agent/conversation_goal.ts
fontanierh: reason provides context for why the goal entered that state (user_interrupted, credits exhausted, agent-provided reason etc..)

let me rename it to statusReason

id13:
Why can't we store this as a .md file ala plan mode ?

fontanierh:
> Why can't we store this as a .md file ala plan mode ?

It's not pure display, the idea is to go for a full /goal mode that keeps relaunching the loop as long as goal is active

id13:
why can't it be done with wake-ups + files ? :D

fontanierh:
@id13 Wake-ups are CRONs, they don't wait for the agent loop to complete. They also don't include a tool that lets you complete the goal. 
Technically you could achieve a broken version of this with wake ups, but every other harness offers this as a first class feature

fontanierh:
very selfishly I need it for sandbox security work

id13:
> They also don't include a tool that lets you complete the goal.

You mean a tool to edit a file ? 

> Technically you could achieve a broken version of this with wake ups, but every other harness offers this as a first class feature

What is first-class, and what a cron + a skill + a file can't do that a first class does ?
I'm not challenging UI changes, I'm challenging the data-model and everything that goes with it

fontanierh:
>You mean a tool to edit a file ?

 Not really no, completing the goal stops the thing that re-triggers the agent loop

>What is first-class, and what a cron + a skill + a file can't do that a first class does ?

We're going to add ability to set a credit budget, and you'll need to associate agent messages with a specific goal object, which you can only achieve with a proper data model

id13:
> Not really no, completing the goal stops the thing that re-triggers the agent loop

nothing here that can't be done with prompting

> We're going to add ability to set a credit budget, and you'll need to associate agent messages with a specific goal object, which you can only achieve with a proper data model

Pretty sure we can do it with what's out and what's being worked on in credits, or by tweaking it slightly, not certain a new data-model with so many foreign keys is the right answer

fontanierh:
>nothing here that can't be done with prompting

The wake-up is a cron that fires at a certain interval. It doesn't immediately fire once the loop completes.

> Pretty sure we can do it with what's out and what's being worked on in credits, or by tweaking it slightly, not certain a new data-model with so many foreign keys is the right answer

I don't see how you could count specific agent messages towards a unique "goal" entity without materializing it in DB. If you have a solution for this then I am happy to reconsider of course.

id13:
> The wake-up is a cron that fires at a certain interval. It doesn't immediately fire once the loop completes.

yes it's a trade-off, 80/20, still think it will work just fine

> I don't see how you could count specific agent messages towards a unique "goal" entity without materializing it in DB. If you have a solution for this then I am happy to reconsider of course.

Will think about smth, coming back tomorrow

fontanierh:
>yes it's a trade-off, 80/20, still think it will work just fine

But I don't really get what we're getting out of the trade-off ? Adding a table to our DB isn't really something I'm worried about

aubin-tchoi:
We need some very well typed metadata like the status, the user ID, at some point the budget
A new table seems like a natural choice, why not a table @id13?

PR #29697, opened 2026-07-29
Replace private-API isBuilder() checks with isManager()

tdraier:
## Description

The `builder` role is deprecated and can no longer be assigned, so the private (non-v1) front-api endpoints that gated global-space writes on `auth.isBuilder()` (= builder + manager + admin) now gate on `auth.isManager()` (= manager + admin) — exactly what the inline `TODO(governance)` markers prescribed. The only practical effect is that legacy `builder`-role users lose these write grants; managers/admins are unaffected.

Direct-check swaps:
- **Spaces** — `spaces/[spaceId]/data_sources/index.ts` (update on a global space).
- **Private files** — `files/[fileId]/index.ts` (delete + upload paths) and `files/[fileId]/rename.ts`.

Middleware swap:
- **Private** `spaces/[spaceId]/data_sources/[dsId]/configuration.ts` PATCH switches from the shared `ensureIsBuilder()` middleware to the existing `ensureIsManager()`. (The `ensureIsBuilder` middleware itself is left as-is because **v1** endpoints still use it — see below.)

Also:
- Error messages updated `` `builders` `` → `` `managers` ``.
- The `noDirectRoleCheck` lint plugin flags `isBuilder`/`isAdmin`/`isUser`/`isDustSuperUser` but **not** `isManager`, so the now-stale `biome-ignore` in `rename.ts` is removed.
- Tests updated: authorized-path tests use `manager` instead of `builder`; message expectations updated.

## Out of scope (intentionally unchanged)

- **v1 / public API**: `v1/.../files/[fileId].ts` direct `isBuilder()`, and the `ensureIsBuilder()` middleware used by `v1` agent-configurations, `v1` tables, and `v1` analytics export. Per request, v1 is not touched — which is also why the shared `ensureIsBuilder` middleware stays.

## Risks

Risk: low. Private-API authorization narrows from {builder, manager, admin} to {manager, admin}; builder can no longer be assigned, and legacy builders are being phased out.

## Verification

- `tsgo --noEmit` clean in `front-api`; `biome check` clean (stale suppression removed).
- `files/[fileId]/index.test.ts` (16) and `rename.test.ts` (8) pass.

### DIFF
--- a/front-api/routes/w/[wId]/files/[fileId]/index.test.ts
+++ b/front-api/routes/w/[wId]/files/[fileId]/index.test.ts
@@ -227,11 +227,11 @@ describe("DELETE /api/w/:wId/files/:fileId", () => {
     vi.clearAllMocks();
   });
 
-  it("should allow builder to delete any file", async () => {
+  it("should allow manager to delete any file", async () => {
     const { auth, user, workspace, globalSpace } =
       await createPrivateApiMockRequest({
         method: "DELETE",
-        role: "builder",
+        role: "manager",
       });
 
     const file = await FileFactory.create(auth, user, {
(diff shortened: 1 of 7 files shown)


tdraier:

- #29697 ◀

matteotrab (approved):

PR #29703, opened 2026-07-29
Fix newly published pod function visibility

fontanierh:
## Description

New pod functions could be invisible to an already-running sandbox for up to 60 seconds because their GCSFuse mount cached directory listings. Give the function bundle mount its own no-list-cache profile while preserving the existing cache for regular pod files.

Also preserve the command result when a short-lived E2B process exits before the stdin RPC arrives, instead of replacing its output with a misleading process-not-found error. Production logs showed 21 occurrences in seven days, all 2 to 43 seconds after function creation.

## Tests

Added focused regression coverage for the real pod mount wiring and the early E2B process-exit race.

- `NODE_ENV=test npm test -- lib/api/file_system/sandbox/gcs_sandbox_mount_adapter.test.ts lib/api/sandbox/providers/e2b.test.ts`
- `npx biome check` on the five changed files

## Risk

Low. Only the read-only pod function mount stops caching directory listings. This may add GCS list calls when functions are resolved; regular pod file mounts keep the existing 60-second cache. The E2B fallback is limited to process-not-found responses and waits on the already-started command handle.

## Deploy Plan

Standard front deployment. Monitor pod function invocation failures for `process with pid ... not found` and newly created function lookup errors.

### DIFF
--- a/front/lib/api/file_system/backends/gcs_file_system_backend.ts
+++ b/front/lib/api/file_system/backends/gcs_file_system_backend.ts
@@ -674,7 +674,7 @@ export class GCSFileSystemBackend implements FileSystemBackend {
   ): GCSMountTarget["mountProfile"] {
     switch (mount.kind) {
       case "pod_sandbox_functions":
-        return "workload";
+        return "pod_sandbox_functions";
 
       case "pod_state":
         return "pod_state_replica";
--- a/front/lib/api/file_system/sandbox/gcs_sandbox_mount_adapter.test.ts
+++ b/front/lib/api/file_system/sandbox/gcs_sandbox_mount_adapter.test.ts
@@ -11,11 +11,8 @@ import {
   type RootCommand,
   renderRootCommand,
 } from "@app/lib/api/sandbox/root_command";
-import { Authenticator } from "@app/lib/auth";
-import { SandboxResource } from "@app/lib/resources/sandbox_resource";
-import { SandboxModel } from "@app/lib/resources/storage/models/sandbox";
-import { WorkspaceModel } from "@app/lib/resources/storage/models/workspace";
-import { WorkspaceResource } from "@app/lib/resources/workspace_resource";
+import { setupPlainConversation } from "@app/tests/utils/conversation_test_factories";
+import { SandboxFactory } from "@app/tests/utils/SandboxFactory";
 import { Err, Ok } from "@app/types/shared/result";
 import { beforeAll, beforeEach, describe, expect, test, vi } from "vitest";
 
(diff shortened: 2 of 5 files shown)


front/lib/api/file_system/sandbox/gcs_sandbox_mount_adapter.test.ts
```diff
@@ -165,6 +173,45 @@ describe("pod sandbox mount wiring", () => {
...
+    );
+    const adapter = createPodSandboxAdapter();
+    const sandbox = {
+      sId: "sandbox-id",
+      execRoot: vi.fn().mockResolvedValue(successfulExec()),
+      requestKill: vi.fn(),
+    };
+    const auth = {
+      getNonNullableWorkspace: () => ({ sId: "workspace-id" }),
+    } as never;
```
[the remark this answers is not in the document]
fontanierh: Addressed in cde8471e44 by replacing the casts with typed test resources.

front/lib/api/file_system/sandbox/gcs_sandbox_mount_adapter.test.ts
```diff
@@ -165,6 +173,45 @@ describe("pod sandbox mount wiring", () => {
...
+    const image = {
+      hasCapability: (capability: string) => capability === "gcsfuse",
+    } as never;
```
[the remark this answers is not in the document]
fontanierh: Addressed in cde8471e44 by replacing the casts with typed test resources.

front/lib/api/file_system/sandbox/gcs_sandbox_mount_adapter.test.ts
```diff
@@ -165,6 +173,45 @@ describe("pod sandbox mount wiring", () => {
...
+
+    const result = await adapter.setup(auth, sandbox as never, image);
```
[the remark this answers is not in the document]
fontanierh: Addressed in cde8471e44 by replacing the casts with typed test resources.

front/lib/api/file_system/sandbox/gcs_sandbox_mount_adapter.test.ts
```diff
@@ -42,13 +48,99 @@ function successfulExec() {
...
+    role: "user",
+    groupModelIds: [],
+    authMethod: "internal",
+    subscription: null,
+  });
+}
+
+function createTestSandbox() {
+  const timestamp = new Date();
+  const sandbox = new SandboxResource(SandboxModel, {
```
flvndvd: IIRC we already have a `SandboxFactory` we should be able to use it.

front/lib/api/file_system/sandbox/gcs_sandbox_mount_adapter.test.ts
```diff
@@ -42,13 +48,99 @@ function successfulExec() {
   return new Ok({ exitCode: 0, stdout: "", stderr: "" });
 }
 
+function createTestAuth(): Authenticator {
+  const timestamp = new Date();
+  const workspace = new WorkspaceResource(WorkspaceModel, {
```
flvndvd: Can we use the existing setupTests helper and factory?

front/lib/api/file_system/backends/gcs_file_system_backend.ts
```diff
@@ -674,7 +674,7 @@ export class GCSFileSystemBackend implements FileSystemBackend {
   ): GCSMountTarget["mountProfile"] {
     switch (mount.kind) {
       case "pod_sandbox_functions":
-        return "workload";
```
flvndvd, line 677: Is this profile still used somewhere?

flvndvd (approved):

front/lib/api/file_system/backends/gcs_file_system_backend.ts
fontanierh: yeah eg conversation files

front/lib/api/file_system/sandbox/gcs_sandbox_mount_adapter.test.ts
fontanierh: Addressed in e4e4a9deb7 by reusing setupPlainConversation and SandboxFactory.

front/lib/api/file_system/sandbox/gcs_sandbox_mount_adapter.test.ts
fontanierh: Addressed in e4e4a9deb7 by reusing setupPlainConversation and SandboxFactory.

front/lib/api/file_system/backends/gcs_file_system_backend.ts
fontanierh: Yes. All agent-facing FileSystemMounts with a sandbox mount point still use workload, including conversation files and Pod/project files (also additional cross-scope mounts). They keep the 60-second list cache; only the read-only function bundle uses the new freshness-specific profile.

PR #29705, opened 2026-07-29
Add Memory tab to user settings

Inesvvv:
## Description

Add a feature-flag-gated Memory tab in the user settings popover with a toggle to enable personal memory and an editable `MEMORY.md` with a character counter and Save button. Toggling is a pending change confirmed via Save. Moved MAX_USER_MEMORY_CHARS to a client-safe module.

Toggle off:
Screenshot 2026-07-29 at 16 00 19

Toggle on: 
Screenshot 2026-07-29 at 16 01 04

## Tests

Tested locally 

## Risk

Low gated behind ff 

## Deploy Plan

Deploy front

### DIFF
--- a/front/components/UserSettingsPopover.tsx
+++ b/front/components/UserSettingsPopover.tsx
@@ -1,4 +1,5 @@
 import { UsageUpgradeButton } from "@app/components/credits/UsageUpgradeButton";
+import { MarkdownEditor } from "@app/components/editor/MarkdownEditor";
 import {
   NotificationPreferences,
   useNotificationPreferencesForm,
@@ -17,7 +18,7 @@ import { AwuUsageBar } from "@app/components/workspace/MembersUsageTable";
 import { useFileUploaderService } from "@app/hooks/useFileUploaderService";
 import { useIsMac } from "@app/hooks/useKeyboardShortcutLabel";
 import { useSendNotification } from "@app/hooks/useNotification";
-import { useAuth } from "@app/lib/auth/AuthContext";
+import { useAuth, useFeatureFlags } from "@app/lib/auth/AuthContext";
 import { isSubmitMessageKey } from "@app/lib/keymaps";
 import { useAppRouter } from "@app/lib/platform";
 import {
@@ -30,9 +31,14 @@ import {
   usePatchUser,
   usePendingInvitations,
   useUser,
+  useUserMemory,
   useWorkspaceUsageStatus,
 } from "@app/lib/swr/user";
 import { getConversationRoute } from "@app/lib/utils/router";
+import {
+  MAX_USER_MEMORY_CHARS,
+  MAX_USER_MEMORY_CONTENT_LENGTH,
+} from "@app/types/api/me/memory";
 import type { PendingInvitationOption } from "@app/types/membership_invitation";
 import { isCreditPricedPlan } from "@app/types/plan";
 import type { WorkspaceType } from "@app/types/user";
@@ -41,6 +47,7 @@ import {
   Avatar,
   BarChart01,
   Bell01,
+  Brain,
   Button,
   ContentMessageInline,
   Dialog,
@@ -63,6 +70,7 @@ import {
   Separator,
   Settings01,
   ShapesPlus,
+  SliderToggle,
   Spinner,
   Stars02,
   Sun,
@@ -87,6 +95,7 @@ type SettingsSection =
   | "usage"
   | "customization"
   | "notifications"
+  | "memory"
   | "tools"
   | "invitations";
 
@@ -427,7 +436,7 @@ function PersonalInfoSection({ owner }: { owner: WorkspaceType }) {
 
   if (isUserLoading) {
     return (
-      <SectionContent title="Personal Informations">
+      <SectionContent title="Personal Information">
         <div className="flex justify-center p-6">
           <Spinner />
         </div>
@@ -437,7 +446,7 @@ function PersonalInfoSection({ owner }: { owner: WorkspaceType }) {
 
   return (
     <SectionContent
-      title="Personal Informations"
+      title="Personal Information"
       footer={
         <Button
           label="Save"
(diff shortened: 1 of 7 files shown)


front/components/UserSettingsPopover.tsx
```diff
@@ -792,6 +799,112 @@ function InvitationsSection({
...
+                    isOverLimit
+                      ? "copy-sm text-warning"
+                      : "copy-sm text-muted-foreground"
+                  }
+                >
+                  {value.length.toLocaleString()} /{" "}
+                  {MAX_USER_MEMORY_CHARS.toLocaleString()} characters
+                </span>
+              </div>
+              <TextArea
```
aubin-tchoi: having a text area is kinda sad no? why not have an editor here OOC? I'd expect ppl to have sections, text in bold, etc.. here

aubin-tchoi: we shouldn't be able to edit while it's saving, otherwise the changes would be discarded and disappear on the mutate

front/components/UserSettingsPopover.tsx
```diff
@@ -792,6 +799,112 @@ function InvitationsSection({
...
+              <div className="flex items-center justify-between">
+                <span className="heading-base text-foreground">Memory</span>
+                <span
+                  className={
+                    isOverLimit
+                      ? "copy-sm text-warning"
+                      : "copy-sm text-muted-foreground"
+                  }
+                >
+                  {value.length.toLocaleString()} /{" "}
```
aubin-tchoi: nit: 
```suggestion
                  {value.length.toLocaleString()} /&nbsp;
```

front/lib/swr/user.ts
```diff
@@ -237,6 +238,61 @@ export function usePendingInvitations({
...
+  return {
+    content: data?.content ?? "",
+    isMemoryEnabled: data?.enabled ?? false,
+    isMemoryLoading: !error && !data && !disabled,
+    isMemoryError: error,
+    mutateMemory: mutate,
+  };
+}
+
+export function useSetUserMemory({ owner }: { owner: LightWorkspaceType }) {
```
aubin-tchoi: I think I'd group the 2 hooks

front/components/UserSettingsPopover.tsx
```diff
@@ -792,6 +799,112 @@ function InvitationsSection({
...
+    >
+      {isMemoryLoading ? (
+        <div className="flex justify-center py-8">
+          <Spinner />
+        </div>
+      ) : (
+        <>
+          <div className="flex items-start justify-between gap-4 rounded-2xl border border-border dark:border-border-dark p-4">
+            <div className="flex flex-col gap-1">
+              <span className="heading-base text-foreground">My Memory</span>
```
aubin-tchoi: quite redundant to have on the same page "Memory" (main title) then a section "My Memory" (toggle) and then "Memory" (actual content) + does not help explaining what each section does

front/lib/swr/user.ts
```diff
@@ -237,6 +238,61 @@ export function usePendingInvitations({
...
+  }): Promise<boolean> => {
+    const res = await clientFetch(`/api/w/${owner.sId}/me/memory`, {
+      method: "PATCH",
+      headers: { "Content-Type": "application/json" },
+      body: JSON.stringify(update),
+    });
+
+    if (res.ok) {
+      sendNotification({ type: "success", title: "Memory saved." });
+      return true;
```
aubin-tchoi: let's still return the data in this case

front/lib/swr/user.ts
```diff
@@ -237,6 +238,61 @@ export function usePendingInvitations({
...
+    isMemoryLoading: !error && !data && !disabled,
+    isMemoryError: error,
+    mutateMemory: mutate,
+  };
+}
+
+export function useSetUserMemory({ owner }: { owner: LightWorkspaceType }) {
+  const sendNotification = useSendNotification();
+
+  const setMemory = async (update: {
```
aubin-tchoi: let's wrap this in a `useCallback` in case it's used as a hook's dependency

aubin-tchoi (commented):
Looks great so far! left a few suggestions for polish

front/components/UserSettingsPopover.tsx
Inesvvv: Screenshot 2026-07-30 at 15 26 13

front/components/UserSettingsPopover.tsx
Inesvvv: Replaced it by the MarkdownEditor

Inesvvv: Added readOnly={isSaving} to the MarkdownEditor

front/lib/swr/user.ts
Inesvvv: Grouped them 👍

front/lib/swr/user.ts
```diff
@@ -237,6 +242,60 @@ export function usePendingInvitations({
...
+    }): Promise<PatchUserMemoryResponseBody | null> => {
+      const res = await clientFetch(`/api/w/${owner.sId}/me/memory`, {
+        method: "PATCH",
+        headers: { "Content-Type": "application/json" },
+        body: JSON.stringify(update),
+      });
+
+      if (res.ok) {
+        sendNotification({ type: "success", title: "Memory saved." });
+        return PatchUserMemoryResponseBodySchema.parse(await res.json());
```
aubin-tchoi: let's mutate here rather than having the caller do it (sorry missed it in the first review)

aubin-tchoi: let's not use `parse` here as this will throw if the validation does not pass (which can happen, for instance if we change the type returned by the backend and you're still running an old client)

front/components/UserSettingsPopover.tsx
```diff
@@ -814,22 +945,29 @@ export function UserSettingsPopover({
...
-    [hasPendingInvitations]
-  );
+  // "Memory" is gated on the user_memory feature flag; "Invitations" only
+  // appears when the user has pending invitations.
+  const navItems = useMemo<typeof NAV_ITEMS>(() => {
+    const base = hasUserMemory
+      ? NAV_ITEMS.flatMap((item) =>
+          item.section === "notifications" ? [item, MEMORY_NAV_ITEM] : [item]
+        )
+      : NAV_ITEMS;
```
aubin-tchoi: nit:
```suggestion
    const base = [...NAV_ITEMS, ...(hasUserMemory ? [MEMORY_NAV_ITEM] : [])];
```

front/components/UserSettingsPopover.tsx
```diff
@@ -792,6 +801,122 @@ function InvitationsSection({
...
+        <div className="flex justify-center py-8">
+          <Spinner />
+        </div>
+      ) : (
+        <>
+          <div className="flex items-start justify-between gap-4 rounded-2xl border border-border dark:border-border-dark p-4">
+            <div className="flex flex-col gap-1">
+              <span className="heading-base text-foreground">
+                {" "}
+                Enable Memory{" "}
```
aubin-tchoi: nit: OOC do we need those whitespaces?

front/components/UserSettingsPopover.tsx
```diff
@@ -792,6 +801,122 @@ function InvitationsSection({
...
+  const [enabledDraft, setEnabledDraft] = useState<boolean | null>(null);
+  const [isSaving, setIsSaving] = useState(false);
+
+  const value = draft ?? content;
+  const enabledValue = enabledDraft ?? isMemoryEnabled;
+
+  // Count visible characters: strip common markdown formatting so bolding a
+  // letter counts as one character, not the surrounding syntax.
+  const characterCount = useMemo(
+    () => value.replace(/[#*_`~>[\]]/g, "").length,
```
aubin-tchoi: this means the UI can tell you you're under the limit, but the backend will reject

front/components/UserSettingsPopover.tsx
```diff
@@ -814,22 +945,29 @@ export function UserSettingsPopover({
...
+  const navItems = useMemo<typeof NAV_ITEMS>(() => {
+    const base = hasUserMemory
+      ? NAV_ITEMS.flatMap((item) =>
+          item.section === "notifications" ? [item, MEMORY_NAV_ITEM] : [item]
+        )
+      : NAV_ITEMS;
+    return hasPendingInvitations
+      ? [
+          ...base,
+          { section: "invitations", icon: Mail01, label: "Invitations" },
```
aubin-tchoi: nit: could be defined just below `MEMORY_NAV_ITEM` as `INVITATION_NAV_ITEM`

front/components/UserSettingsPopover.tsx
aubin-tchoi: as per IRL let's re-use the behavior already implemented in `MarkdownEditor`

aubin-tchoi (approved):
LGTM once comments are addressed

front/lib/swr/user.ts
Inesvvv: Now awaits mutate() and doesnt use parse Since the hook now revalidates via mutate(), we don't need to read the response body at all, setMemory just returns true

front/components/UserSettingsPopover.tsx
Inesvvv: Dropped the custom regex counter and reused MarkdownEditor's built-in character count. The Save button gates on the raw markdown length against the server cap, so we never submit a body the backend would reject.

front/components/UserSettingsPopover.tsx
Inesvvv: There are no _NAV_ITEM constants anymore since inlined both memory and invitations directly into NAV_ITEMS

front/lib/swr/user.ts
aubin-tchoi: hmmm ok but why? 😬

aubin-tchoi (approved):

front/lib/swr/user.ts
Inesvvv: Before we were retruning the data with parse but the aim of this part is just to say if the save succeeded

PR #29707, opened 2026-07-29
feat(marketing): add USD/credits toggle to API pricing page

ClementAupiais:
## Description

Adds a USD / credits toggle to the `/home/api-pricing` marketing page so visitors can view model prices either in USD per million tokens (existing behaviour) or in intelligence credits. Credits are derived from raw model prices using the same `$0.0085 per credit, rounded up` formula that mirrors `awuFromMicroUsd` in `front/lib/metronome/events.ts`. Cache read/write columns show dashes in credits mode, and a short disclaimer is appended below the table when credits are selected.

## Tests

Manual verification on the marketing page: switch the toggle between USD and credits and confirm values update correctly, cache columns show dashes in credits mode, and the disclaimer appears/disappears as expected.

## Risk

Low — marketing-only change, no backend or data model involved.

## Deploy Plan

Deploy marketing.

### DIFF
--- a/marketing/pages/home/api-pricing.tsx
+++ b/marketing/pages/home/api-pricing.tsx
@@ -32,6 +32,8 @@ const modelConfigMap = new Map<
   ])
 );
 
+type PriceUnit = "usd" | "credits";
+
 interface PricingRow {
   modelId: string;
   displayName: string;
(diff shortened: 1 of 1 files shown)


pinotalexandre (approved):
LGTM modulo dust comments

PR #29710, opened 2026-07-29
[FILES] Point audio file fragments at their transcript sibling

pmilliotte:
## Description

Audio `<file>` tags rendered for the new file explorer now carry a `processedPath` attribute pointing at the mounted `.processed.txt` transcript sibling, since no tool can read the original binary and `files_list` filters `.processed.*` entries out.

## Tests

Added a unit test in `content_fragment_resource.test.ts`; full file suite passes (20 tests).

## Risk

Low.

## Deploy Plan

- Deploy `main` on `front`

### DIFF
--- a/front/lib/api/assistant/conversation/attachments.ts
+++ b/front/lib/api/assistant/conversation/attachments.ts
@@ -218,6 +218,7 @@ export function getAttachmentFromFileContentFragment(
     ...baseAttachment,
     fileId,
     path: cf.path ?? null,
+    processedPath: cf.processedPath ?? null,
     source: "user",
     createdAt: cf.created,
   };
--- a/front/lib/resources/content_fragment_resource.ts
+++ b/front/lib/resources/content_fragment_resource.ts
@@ -767,6 +767,7 @@ export class ContentFragmentResource extends BaseResource<ContentFragmentModel> 
           contentFragmentType: "file",
           expiredReason: fr.expiredReason,
           path: null,
+          processedPath: null,
           skipDataSourceIndexing: false,
           skipFileProcessing: false,
           fileId: null,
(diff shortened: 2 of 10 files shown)


front/lib/resources/content_fragment_resource.ts
```diff
@@ -107,6 +114,36 @@ function getConversationFilePath({
...
+ */
+export function getTextProcessedMountPath(
+  path: string,
+  contentType: SupportedContentFragmentType
+): string | null {
+  if (!isSupportedFileContentType(contentType)) {
+    return null;
+  }
+
+  const processedContentType = getProcessedContentType(
```
flvndvd: Current logic is to actually rely on `getContentVersion()` and check that it returns `processed` this provides strong indicator that the file should have a processed version.

front/lib/api/assistant/conversation/attachments.ts
```diff
@@ -218,6 +218,7 @@ export function getAttachmentFromFileContentFragment(
     ...baseAttachment,
     fileId,
     path: cf.path ?? null,
+    processedPath: cf.processedPath ?? null,
```
flvndvd, line 221: We really want to avoid exposing the file once it has been moved to a conversation / pod file system.

front/lib/resources/file_resource.ts
```diff
@@ -670,6 +670,37 @@ export class FileResource extends BaseResource<FileModel> {
     return "original";
   }
 
+  /**
+   * Mount path of the processed sibling `copyMountFiles` writes next to the original (e.g.
+   * `voice.processed.txt`), or null when this file has no processed version.
+   */
+  getProcessedMountFilePath(): string | null {
```
flvndvd, line 677: Really file, is just a way to upload something in Dust and host frames, it should not exposed this abstraction.

PR #29711, opened 2026-07-29
[front] perf: Add workspace /exists probe endpoint and kinds filter on public spaces list

id13:
## Description

Currently, we fetch all spaces from a workspace in not-debounced path from slack connector's webhooks handler, just to check whether the workspace exists and whether we should do a signal-with-start.

As a result, slack connector is currently responsible for front's DB plateauing at 99.9% CPU.

This PR lays the ground work to optimize this by 

- introducing a /exists to workspace scope, does what it says
- introducing a kind for /spaces, because most of the time we do not care about pods (which is the variable that made this explode recently)

## Tests

Unit tests

## Risk

Low

## Deploy Plan

Deploy front


id13:

- #29711 ◀
    - #29712

id13:
PMRR (merge when green)

sdks/js/src/index.ts
```diff
@@ -1914,10 +1915,13 @@ export class DustAPI {
     return new Ok(r.value.apps);
   }
 
-  async getSpaces() {
+  async getSpaces(options?: { kinds?: SpaceType["kind"][] }) {
```
Fraggle, line 1942: you are adding .exists() method in another PR?

Fraggle (approved):
LGTM

sdks/js/src/index.ts
id13: yes, but makes sense to add it now in fact

id13: done

PR #29714, opened 2026-07-29
Avoid fetching all spaces in multiple places. Optimize when we fetch Pods the uer is a member of.

Fraggle:
## Description

Two related query optimizations:

- `SpaceResource.listWorkspaceSpacesAsMember` has an optional `kinds` filter, pushing kind filtering into the DB query rather than post-fetching all spaces. Callers in `selected_spaces.ts` (regular only), `workspace_capabilities.ts` (global + regular), and the public spaces API (system + global + regular) pass the exact kinds they need.
- New `listWorkspacePodsAsMember` fast-path for Pod-only lookups: queries `GroupSpaceModel` for `member`/`project_editor` vault IDs matching the auth's `groupModelIds()`, then fetches only `kind: "project"` spaces from those IDs. Used by `listNonArchivedMemberSpacesWithMetadata`. Includes a safety `isMember` guard with error logging in case of a mismatch.

## Tests

Tested locally. No behavioral change — callers were already filtering to the same kinds in application code.

## Risk

Low. Pure query optimization; no data model or API contract changes. The Pod fast path adds a safety filter with error logging, making any regression observable before it causes incorrect results. Fully rollback-safe.

## Deploy Plan

Deploy front + front-api.

### DIFF
--- a/front/lib/api/assistant/conversation/selected_spaces.ts
+++ b/front/lib/api/assistant/conversation/selected_spaces.ts
@@ -67,7 +67,7 @@ export async function listSelectableSpaces(
   }
 
   const [spaces, selectedSpaces] = await Promise.all([
-    SpaceResource.listWorkspaceSpacesAsMember(auth),
+    SpaceResource.listWorkspaceSpacesAsMember(auth, { kinds: ["regular"] }),
     ConversationSelectedSpaceResource.listActiveSpacesByConversation(auth, {
       conversation,
     }),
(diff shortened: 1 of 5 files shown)


Fraggle:

- #29714 ◀

front/lib/api/assistant/conversation/selected_spaces.ts
```diff
@@ -67,7 +67,7 @@ export async function listSelectableSpaces(
   }
 
   const [spaces, selectedSpaces] = await Promise.all([
-    SpaceResource.listWorkspaceSpacesAsMember(auth),
+    SpaceResource.listWorkspaceSpacesAsMember(auth, { kinds: ["regular"] }),
```
Fraggle, line 70: we are filtering just below anyway

PR #29715, opened 2026-07-29
Notion connector: unstuck notionDrainDocumentUpsertQueueWorkflow with continueAsNew

fabiencelier:
## Description

Drain could not finish in 1h because the upsert queue was disable.
The workflows are now stuck on the last step, which keep throwing.
We should continue as new the workflow so we can handle situations where the queue is disabled.

## Tests

not tested

## Risk

may not see other issues caused by not being able to drain in 1h
Maybe not deterministic error, but as far as I understand no 

## Deploy Plan

deploy connector


davidebbo (approved):
LGTM

PR #29719, opened 2026-07-29
[front] Fix analytics llm_awu single-ceiling multi-execution messages

tdraier:
## Description

Fixes `cost.llm_awu` in the `agent_message_analytics` ES doc **under-counting multi-execution messages** (ES-side accuracy; Metronome billing is unaffected).

The analytics indexer computes:

```ts
const llmAwu = intelligenceAwuFromRunUsagesGroupedByRunKey(runUsages, contextOrigin);
```

That function ceils intelligence **per runKey (execution)** — but the analytics activity reads runs via `fetchRunUsagesForMessage` and **never tags their runKeys**; it relied on the credit path (`computeAndStoreAgentMessageCredits` → `setRunKeyForDustRunIds`) having already tagged them. When the analytics index runs before that tagging, every execution collapses into the single `LEGACY_RUN_KEY` group and the message is **single-ceiled** (`ceil(Σ micro)`) instead of ceiled per execution (`Σ ceil(micro)`). Since `ceil(a) + ceil(b) ≥ ceil(a+b)`, the stored `llm_awu` under-counts multi-execution (interrupt/resume) messages relative to the billed amount.

**Fix:** tag this execution's runKey (`computeRunKey(dustRunIds)`) before reading usages — mirroring the credit path, and idempotent (same `dustRunIds` → same runKey the credit/emit paths use). The analytics ceiling then matches the billed per-execution value.

## Evidence (prod)

Message `zXWcbNElAO` — 2 executions, `43164 + 9108` micro-USD. Stored `cost.llm_awu = 7` (`= ceil(52272/8500)`, a single ceil), while the correct per-execution value is `ceil(43164/8500) + ceil(9108/8500) = 6 + 2 = 8` (which Metronome billed). Drill-down (`computeRunKey` over the persisted runs) confirms 2 distinct runKeys, recomputing to 8. This produced the `−1` (ES 13 vs Metronome 14) on that day.

## Tests

- `tsgo` clean (front).
- No automated test added (same reason as the sibling PR — Temporal indexing timing); verified by the prod recompute above.
- The other caller of `storeAgentAnalytics` (the `20251103_backfill_agent_analytics` migration) is unaffected — `dustRunIds` is optional and omitted there.
- ⚠️ Committed with `--no-verify` (misconfigured local pre-commit hook) — **rely on CI for format/lint**.

## Risk

Low. Adds one idempotent `setRunKeyForDustRunIds` write in the analytics activity (the exact same write the credit path already makes, so no new value/conflict). Only affects the computed `cost.llm_awu`/`full_awu` on the analytics doc for multi-execution messages; single-execution messages are unchanged (`ceil(x) == ceil(x)`). Safe to roll back by revert.

## Deploy Plan

Standard deploy, no migration. Corrects new analytics docs going forward. Existing docs remain single-ceiled until re-indexed — a re-backfill would be needed to correct historical `llm_awu` (separate decision; the drift is ≤ ~1 AWU per multi-execution message).

### DIFF
--- a/front/temporal/analytics_queue/activities.ts
+++ b/front/temporal/analytics_queue/activities.ts
@@ -20,6 +20,7 @@ import { isLLMTraceId } from "@app/lib/api/llm/traces/buffer";
 import type { AuthenticatorType } from "@app/lib/auth";
 import { Authenticator } from "@app/lib/auth";
 import {
+  computeRunKey,
   intelligenceAwuFromRunUsagesGroupedByRunKey,
   toolAwuFromAction,
 } from "@app/lib/metronome/events";
@@ -151,6 +152,7 @@ export async function storeAgentAnalyticsActivity(
     userMessageModel: userUserMessageRow,
     conversationRow,
     contextOrigin: userUserMessageRow.userContextOrigin,
+    dustRunIds: agentLoopArgs.dustRunIds,
   });
 }
 
@@ -166,6 +168,7 @@ export async function storeAgentAnalytics(
     userMessageModel: UserMessageModel;
     conversationRow: ConversationModel;
     contextOrigin: UserMessageOrigin | null;
+    dustRunIds?: string[];
   }
 ): Promise<void> {
   const {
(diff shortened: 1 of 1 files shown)


tdraier:

- #29719 ◀

fabiencelier (approved):

PR #29720, opened 2026-07-29
[SCRUB] Cap the free-plan downgrade cron so it stops wedging on large backlogs

pmilliotte:
## Description

The downgrade cron starts one scrub child per expired workspace, so it hit Temporal's 1000-pending-children cap on Apr 27, wedged, and was terminated on May 2 — 40,701 expired free plans are still active. Now capped at 500 workspaces per run (oldest expiry first), 3 runs per weekday at 15:00-17:00 UTC (scrub children email on start, so that doubles as the send window).

## Tests

No.

## Risk

Relaunching drains the 40,701 backlog at ~1,500/weekday, each workspace getting a data-deletion email and a scrub 30 days later. Rate is tunable via `MAX_WORKSPACES_TO_DOWNGRADE_PER_RUN`.

## Deploy Plan

- Deploy `main` on `front`
- The cron is terminated and won't restart on its own — relaunch from Prodbox (US only):
  ```
  cd /dust/front
  npx tsx temporal/scrub_workspace/admin/cli.ts launch-workflow-to-downgrade-free-ended-workspaces
  ```
  It fires a run immediately, so the first 500 emails go out on the spot.

### DIFF
--- a/front/lib/resources/subscription_resource.ts
+++ b/front/lib/resources/subscription_resource.ts
@@ -698,7 +698,11 @@ export class SubscriptionResource extends BaseResource<SubscriptionModel> {
     return res !== null;
   }
 
-  static async internalFetchWorkspacesWithFreeEndedSubscriptions(): Promise<{
+  static async internalFetchWorkspacesWithFreeEndedSubscriptions({
+    limit,
+  }: {
+    limit?: number;
+  } = {}): Promise<{
     workspaces: LightWorkspaceType[];
   }> {
     const freeEndedSubscriptions = await SubscriptionModel.findAll({
@@ -710,6 +714,9 @@ export class SubscriptionResource extends BaseResource<SubscriptionModel> {
         },
       },
       include: [WorkspaceModel],
+      // Oldest expiry first, so a capped run drains the backlog deterministically.
+      order: [["endDate", "ASC"]],
+      limit,
     });
 
     const workspaces = freeEndedSubscriptions.map((s) =>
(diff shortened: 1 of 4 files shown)


front/temporal/scrub_workspace/config.ts
```diff
@@ -4,6 +4,10 @@ export const QUEUE_NAME = `scrub-workspace-queue-v${QUEUE_VERSION}`;
 export const DOWNGRADE_FREE_ENDED_WORKSPACES_WORKFLOW_ID =
   "downgrade-and-scrub-free-ended-workspaces";
 
+// One scrub child per workspace, and Temporal caps pending children at 1000 per
+// run — going over wedges the cron. Backlog drains over successive runs.
+export const MAX_WORKSPACES_TO_DOWNGRADE_PER_RUN = 500;
```
aubin-tchoi, line 9: won't that take > 80 days to drain the queue then?

front/lib/resources/subscription_resource.ts
```diff
@@ -710,6 +714,9 @@ export class SubscriptionResource extends BaseResource<SubscriptionModel> {
         },
       },
       include: [WorkspaceModel],
+      // Oldest expiry first, so a capped run drains the backlog deterministically.
+      order: [["endDate", "ASC"]],
```
aubin-tchoi, line 718: may end up doing a sequential scan, to monitor when rolled out

aubin-tchoi (approved):
LGTM

front/temporal/scrub_workspace/config.ts
pmilliotte: I increased x3 the frequency of the cron so it should take a month, will be able to increase again if it's not fast enough

PR #29726, opened 2026-07-29
feat(marketing): add standalone credits pricing page

ClementAupiais:
## Description

Adds a new standalone `/credits` marketing page (`/home/credits`) listing intelligence credit consumption per model, grouped in collapsible provider sections with logos (OpenAI, Anthropic, Google, Mistral, xAI, DeepSeek, Z.ai, Moonshot AI). A permanent redirect from `/credits` → `/home/credits` is added in `next.config.js`. The page is not yet linked in the navigation.

## Tests

Manual: navigate to `/credits` and verify the redirect works, sections are collapsible, all provider logos render, and the disclaimer is visible.

## Risk

Low — new page only, no existing functionality changed.

## Deploy Plan

1. Deploy front 
2. Deploy marketing.

### DIFF
--- a/front-api/routes/marketing/model_credits.test.ts
+++ b/front-api/routes/marketing/model_credits.test.ts
@@ -1,0 +1,21 @@
+import { honoApp } from "@front-api/app";
+import { describe, expect, it } from "vitest";
+
+describe("GET /api/marketing/model-credits", () => {
+  it("returns the public model credits table", async () => {
+    const response = await honoApp.request("/api/marketing/model-credits");
+
+    expect(response.status).toBe(200);
+    const body = await response.json();
+    expect(Array.isArray(body.models)).toBe(true);
+    expect(body.models.length).toBeGreaterThan(0);
+
+    const [model] = body.models;
+    expect(typeof model.modelId).toBe("string");
+    expect(typeof model.displayName).toBe("string");
+    expect(typeof model.modelMaker).toBe("string");
+    expect(typeof model.modelMakerDisplayName).toBe("string");
+    expect(typeof model.inputCreditsPerMTokens).toBe("number");
+    expect(typeof model.outputCreditsPerMTokens).toBe("number");
+  });
+});
--- a/front/lib/api/marketing/model_credits.ts
+++ b/front/lib/api/marketing/model_credits.ts
@@ -1,0 +1,80 @@
+import {
+  computeTokensCostForUsageInMicroUsd,
+  MODEL_PRICING,
+} from "@app/lib/api/assistant/token_pricing";
+import { awuFromMicroUsd } from "@app/lib/metronome/events";
+import { SUPPORTED_MODEL_CONFIGS } from "@app/types/assistant/models/models";
+import {
+  getModelMaker,
+  getModelMakerDisplayName,
+} from "@app/types/assistant/models/providers";
+import type { ModelMakerIdType } from "@app/types/assistant/models/types";
+
+export interface PublicModelCredit {
+  modelId: string;
+  displayName: string;
+  modelMaker: ModelMakerIdType;
+  modelMakerDisplayName: string;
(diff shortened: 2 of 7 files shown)


avervaet:
**Where the model list and pricing come from**

The credits table is no longer hardcoded — it's now computed from the same data the app itself uses to bill customers:
- Model list: `SUPPORTED_MODEL_CONFIGS` in `front/types/assistant/models`, the canonical list of currently offered models.
- Pricing: `MODEL_PRICING` in `front/lib/api/assistant/token_pricing`, read through the real `computeTokensCostForUsageInMicroUsd` billing function and converted to credits via `awuFromMicroUsd` (same conversion Metronome billing uses).

This is served by a new public endpoint (`/api/marketing/model-credits`) and fetched by the marketing site at build time (ISR, revalidated hourly), so the page can't drift from real pricing and new models show up automatically once priced.

avervaet:
**Side note on tiered pricing (e.g. Grok 4.5)**

Some models have a higher-priced long-context tier past a prompt-token threshold (Grok 4.5 switches to a pricier rate above 200k prompt tokens). This page shows the base-tier rate only and ignores that tiering for now, for the sake of a simple, readable rate card. Flagging in case we want to surface it later.

front-api/routes/marketing/model_credits.test.ts
```diff
@@ -0,0 +1,21 @@
...
+  it("returns the public model credits table", async () => {
+    const response = await honoApp.request("/api/marketing/model-credits");
+
+    expect(response.status).toBe(200);
+    const body = await response.json();
+    expect(Array.isArray(body.models)).toBe(true);
+    expect(body.models.length).toBeGreaterThan(0);
+
+    const [model] = body.models;
+    expect(typeof model.modelId).toBe("string");
```
flvndvd, line 14: Ahah not sure it's very relevant 🙃.

front/lib/api/marketing/model_credits.ts
```diff
@@ -0,0 +1,80 @@
...
+
+// Some models (e.g. Grok 4.5) have a higher-priced long-context tier past a
+// prompt token threshold. Pricing a single token keeps every model on its
+// base tier, so the displayed rate is the flat per-million-token price
+// rather than the blended rate of one huge long-context request.
+const SINGLE_TOKEN = 1;
+
+// Provider ids that are internal routing/no-op concepts rather than actual,
+// separately priced models — not relevant on a public per-model credits page.
+const EXCLUDED_PROVIDER_IDS = new Set([
```
flvndvd, line 32: We sometimes uses custom model injected at built time, we don't want those to be surfaced. Rather than name, maybe it we should exclude all models behind a FF?

front-api/routes/marketing/model_credits.test.ts
flvndvd, line 14: I'd expect those tests to be more about not returning excluded models.

marketing/lib/api/model_credits.ts
```diff
@@ -0,0 +1,42 @@
...
+import { z } from "zod";
+import { fromError } from "zod-validation-error";
+
+const modelCreditSchema = z.object({
+  modelId: z.string(),
+  displayName: z.string(),
+  // Plain string at the wire boundary — front may introduce a new model
+  // maker before this app's logo mapping is updated. Unknown makers fall
+  // back to a generic rendering instead of crashing the page.
+  modelMaker: z.string(),
```
flvndvd, line 11: `nit` IIRC @pmilliotte has been updating the codebase to use `lab`.

flvndvd (approved):

marketing/lib/api/model_credits.ts
pmilliotte: This is a different property added by @Nils-Fedrigo to display the right "provider" when selecting a model

PR #29727, opened 2026-07-29
Fix: unread conversations in pod summary are enriched with the next wake up (so they show up in the correct section)

Fraggle:
## Description

`listSpaceUnreadConversationsAndActivityForUser` was never calling `enrichWithNextWakeupAt`, so `nextWakeupAt` was always `null` on conversations returned in the pod sidebar summary. As a result, conversations with an active scheduled wake-up were missing from the "Triggered" section and landing in the wrong inbox bucket.

Fix: call `enrichWithNextWakeupAt` on `unreadConversations` + `nonParticipantUnreadConversations` before building the summary. Also refactors `fetchNextWakeupAtByConversationId` → `enrichWithNextWakeupAt` to mutate instances directly (storing `nextWakeupAt` as an instance field), so `toListItem()` can read it instead of hardcoding `null`.

## Tests

Added a unit test in `conversation_resource.test.ts` that creates both a cancelled and a scheduled `WakeUpResource` for an unread conversation, then asserts `listSpaceUnreadConversationsAndActivityForUser` returns the correct `nextWakeupAt` timestamp (and ignores the cancelled one).

## Risk

Low. Pure bug fix — adds one `WakeUpModel.findAll` query in a path that was already missing it (the same query already ran in the private conversations path). No schema changes, no API contract changes. Safe to rollback.

## Deploy Plan

Deploy front.

### DIFF
--- a/front/lib/resources/conversation_resource.test.ts
+++ b/front/lib/resources/conversation_resource.test.ts
@@ -3431,6 +3431,65 @@ describe("listSpaceUnreadConversationsForUser", () => {
     expect(allConversationIds).toContain(conversationIds[0]);
     expect(allConversationIds).not.toContain(subConversation.sId);
   });
+
+  it("hydrates nextWakeupAt on unread space conversations", async () => {
+    vi.spyOn(
+      wakeUpTemporalClient,
+      "launchOrScheduleWakeUpTemporalWorkflow"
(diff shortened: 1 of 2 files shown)


Fraggle:

- #29727 ◀

ElPicador (commented):
👍

PR #29731, opened 2026-07-29
Fix sandbox tool auth for agent spaces

fontanierh:
## Description

Sandbox exec auth only retained groups from the conversation requested spaces. Agents using tools backed by other authorized spaces were filtered out by `/sandbox/actions` and returned `agent_configuration_not_found`.

Include requested spaces from the exact agent version that issued the sandbox token while keeping the result intersected with the current user group memberships. New tokens sign the agent version; pre-deploy tokens recover it from their signed conversation, message, and agent identifiers. Both tool listing and execution use that exact version and fail closed if it is unavailable.

## Tests

- Restricted agent space with a separate pod conversation
- Agent version drift for tool listing and child execution
- Legacy token compatibility and missing-version rejection
- Focused token, child-action, endpoint, and conversation-auth suites

## Risk

Moderate. This changes a security-sensitive authorization path, but access remains workspace-bound, limited to the token user current memberships, pinned to the issuing agent version, and covered by positive and negative regressions.

## Deploy Plan

Deploy front and front-sse together.


front/tests/utils/SandboxTokenFactory.ts
```diff
@@ -24,8 +24,10 @@ process.env.DUST_SANDBOX_JWT_SECRET ??= "test-sandbox-jwt-secret";
 
 export async function createSandboxTokenTestContext({
   disableComputerFeature = false,
+  useProjectSpaceForConversation = false,
 }: {
   disableComputerFeature?: boolean;
+  useProjectSpaceForConversation?: boolean;
```
flvndvd: Can we avoid using `project` name here?

flvndvd (approved):

front/tests/utils/SandboxTokenFactory.ts
fontanierh: project is really the name in the code tbh, but sure can use pod

flvndvd: Ok, cc @matteotrab hard to say today which name we should use around pod/project.

fontanierh: (fixing though because directionally I suppose the word project will ~disappear)

fontanierh: Addressed in 95b6435f22. The fixture now uses a generic regular space with explicit membership, and the option is named `useDifferentSpaceForConversation` to describe the condition under test.

fontanierh: Follow-up in 1e739abc7b: renamed this to `usePodSpaceForConversation` and kept the fixture pod-scoped, matching the production case. The underlying test factory still uses the legacy `SpaceFactory.project` method.

PR #29736, opened 2026-07-30
Add threshold to hide pod conversations in inbox.

Fraggle:
## Description

When a user has ≥ 16 unread pod conversations, `AgentSidebarMenu` now excludes them from the merged `allConversations` list to avoid flooding the inbox sidebar. The `spaces` summary is still computed in both branches so pod section headers remain visible even when the conversations are suppressed.

## Tests

Tested manually — verified inbox rendering with both few and many (≥ 16) unread pod conversations.

## Risk

Low. Front-only change to a single `useMemo` in `AgentSidebarMenu`. The capped path (`conversations` only) is effectively the pre-pod-inbox baseline. Fully safe to rollback.

## Deploy Plan

Deploy front.

### DIFF
--- a/front/components/assistant/conversation/SidebarMenu.tsx
+++ b/front/components/assistant/conversation/SidebarMenu.tsx
@@ -119,6 +119,9 @@ import {
   useState,
 } from "react";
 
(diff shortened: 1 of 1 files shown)


Fraggle:

- #29736 ◀

PR #29737, opened 2026-07-30
[front] chore: move MCP server view helper out of `lib/models`

aubin-tchoi:
## Description

In our repo, the folder `lib/models` is reserved for Sequelize models.
This PR moves some helpers (`mcp_server_view_helper.ts`) outside of there and into `lib/resources`. It's not ideal but if we really think about it, these functions should not exist: they should be methods on resources but some of them belong on the non-existing `AgentResource`.

Having them in `lib/models` makes them hard to find and messes up with our code owner scopes.

## Tests

- Tested locally.

## Risk

- Low.

## Deploy Plan

- Deploy front.


Fraggle (approved):
LGTM

id13 (approved):

PR #29722, opened 2026-07-29
[connectors] feat: defer slack webhook events to workflow instead of processing inline

id13:
## Description

Defer webhooks processing to a temporal workflow with controllable back-pressure instead of blasting our databases.

## Tests

Tested locally

## Risk

Standard : moving a lot of logic, and potentially putting some load on the slack workers (sending a lot of signals) - will need monitoring
Blast radius : slack bot and slack connector

## Deploy Plan

- [ ] Deploy connectors

PR #29738, opened 2026-07-30
[connectors] enh: Add the Slack webhook event workflow and activity

id13:
## Description

First PR of two: 

1. (this PR) Put all the content of slack webhook handler into a temporal workflow resisting the urge to refacto and just make the switch case cleaner
2. remove the "simple calls" in the webhook handler and defer all the heavy DB duty into the brand new temporal workflow to have a controllable back-pressure.

## Tests

Added some tests
Tested locally (on the top branch)

## Risks

Standard : moving a lot of logic, and potentially putting some load on the slack workers (sending a lot of signals) - will need monitoring
Blast radius : slack bot and slack connector

## Plan

- Deploy connectors
Note this branch is a no-op, incident risk is triggered by the next one

### DIFF
--- a/connectors/src/api/webhooks/webhook_slack.ts
+++ b/connectors/src/api/webhooks/webhook_slack.ts
@@ -4,6 +4,7 @@ import {
 } from "@connectors/api/webhooks/slack/created_channel";
 import { handleDeprecatedChatBot } from "@connectors/api/webhooks/slack/deprecated_bot";
 import type {
+  SlackWebhookEvent,
   SlackWebhookReqBody,
   SlackWebhookResBody,
 } from "@connectors/api/webhooks/slack/utils";
@@ -29,6 +30,7 @@ import { concurrentExecutor } from "@connectors/lib/async_utils";
 import { upsertDataSourceFolder } from "@connectors/lib/data_sources";
 import { ExternalOAuthTokenError } from "@connectors/lib/error";
 import { SlackChannelModel } from "@connectors/lib/models/slack";
+import type { Logger } from "@connectors/logger/logger";
 import mainLogger from "@connectors/logger/logger";
 import { apiError, withLogging } from "@connectors/logger/withlogging";
 import { ConnectorResource } from "@connectors/resources/connector_resource";
@@ -37,6 +39,29 @@ import { INTERNAL_MIME_TYPES } from "@connectors/types";
 import { DustAPI, removeNulls } from "@dust-tt/client";
 import { JSON } from "@jsonjoy.com/util/lib/json-brand";
 import type { Request, Response } from "express";
+
+// Answers Slack before doing the Slack API work, otherwise they retry the HTTP
+// request. `handleDeprecatedChatBot` used to own that response.
+async function notifyDeprecatedBot(
+  res: Response<SlackWebhookResBody>,
+  event: SlackWebhookEvent,
+  teamId: string,
+  logger: Logger
+) {
+  res.status(200).send();
+
+  if (!event.channel || !event.ts) {
+    logger.info("Ignoring event without a channel or a timestamp");
+    return;
+  }
+
+  await handleDeprecatedChatBot({
+    logger,
+    slackChannel: event.channel,
+    slackMessageTs: event.ts,
+    slackTeamId: teamId,
+  });
+}
 
 const _webhookSlackAPIHandler = async (
   req: Request<
@@ -108,7 +133,7 @@ const _webhookSlackAPIHandler = async (
     try {
       switch (event.type) {
         case "app_mention": {
-          await handleDeprecatedChatBot(req, res, logger);
+          await notifyDeprecatedBot(res, event, teamId, logger);
           break;
         }
         /**
@@ -177,7 +202,7 @@ const _webhookSlackAPIHandler = async (
               return res.status(200).send();
             }
             // Message from an actual user (a human)
-            await handleDeprecatedChatBot(req, res, logger);
+            await notifyDeprecatedBot(res, event, teamId, logger);
             break;
           } else if (
             event.channel_type === "channel" ||
@@ -499,7 +524,8 @@ const _webhookSlackAPIHandler = async (
         case "channel_created": {
           if (isChannelCreatedEvent(event)) {
             const onChannelCreationRes = await onChannelCreation({
-              event,
+              channelId: event.channel.id,
+              contextTeamId: event.channel.context_team_id,
               logger,
               provider: "slack",
             });
--- a/connectors/src/connectors/slack/temporal/client.ts
+++ b/connectors/src/connectors/slack/temporal/client.ts
@@ -8,11 +8,15 @@ import { ConnectorResource } from "@connectors/resources/connector_resource";
 import type { ModelId } from "@connectors/types";
 import { normalizeError } from "@connectors/types";
 import { Err, Ok, removeNulls } from "@dust-tt/client";
+import { WorkflowExecutionAlreadyStartedError } from "@temporalio/common";
 
 import { getWeekStart } from "../lib/utils";
 import { QUEUE_NAME } from "./config";
 import { newWebhookSignal, syncChannelSignal } from "./signals";
-import type { JoinChannelUseCaseType } from "./workflows";
+import type {
+  JoinChannelUseCaseType,
+  SlackWebhookEventPayload,
+} from "./workflows";
 import {
   joinChannelWorkflow,
   joinChannelWorkflowId,
@@ -20,6 +24,8 @@ import {
   migrateChannelsFromLegacyBotToNewBotWorkflowId,
   slackGarbageCollectorWorkflow,
   slackGarbageCollectorWorkflowId,
+  slackWebhookEventWorkflow,
+  slackWebhookEventWorkflowId,
   syncOneMessageDebounced,
   syncOneMessageDebouncedWorkflowId,
   syncOneThreadDebounced,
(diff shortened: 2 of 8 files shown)


id13:

- #29738 ◀
    - #29722

connectors/src/connectors/slack/temporal/workflows.ts
```diff
@@ -368,6 +432,10 @@ export function syncOneChanneWorkflowlId(
   return `slack-syncOneChannel-${connectorId}-${channelId}`;
 }
 
+export function slackWebhookEventsWorkflowId(teamId: string) {
```
flvndvd: OOC what drove the design towards one workflow per team vs one workflow per event?

id13: There is only one slack team AFAIK (it's our app id ?), slack sends us events in batches, so we process them in batches.

One workflow per event is a temporal anti-pattern I think, given the numbers we are talking about.

flvndvd: IIRC `teamId` maps to the customer Slack's team. And I don't recall getting batch of them. Let me take a look.

flvndvd: Looking [here](https://github.com/dust-tt/dust/blob/efad307cd7a5410dc75d6a7f5c05bd6e45c8622e/connectors/src/api/webhooks/webhook_slack.ts#L66), it seems that webhook is for an individual event, no?

id13: True, my bad, that being said I still don't want to launch one workflow per event 😅 and would rather send signals.

But if we receive very little events it will end up with one workflow per event anyway

id13: What's your FUD about this design (other than me being incorrect on teamId) ?

id13: seen IRL, signals don't scale well, doing per event workflow

connectors/src/connectors/slack/temporal/webhook_activities.ts
```diff
@@ -0,0 +1,505 @@
...
+ * and dispatching the sync.
+ *
+ * Slack has already been answered by the time this runs, so a failing event is
+ * logged and dropped rather than retried: replaying would re-post the Slack
+ * messages the handlers already sent.
+ */
+export async function processSlackWebhookEventActivity({
+  event,
+  teamId,
+}: {
```
id13, line 520: one workflow per event now

connectors/src/api/webhooks/webhook_slack.ts
```diff
@@ -38,6 +40,29 @@ import { DustAPI, removeNulls } from "@dust-tt/client";
...
 
+// Answers Slack before doing the Slack API work, otherwise they retry the HTTP
+// request. `handleDeprecatedChatBot` used to own that response.
+async function notifyDeprecatedBot(
+  res: Response<SlackWebhookResBody>,
+  event: SlackWebhookEvent,
+  teamId: string,
+  logger: Logger
+) {
+  res.status(200).send();
```
flvndvd, line 51: Assuming that `handleDeprecatedChatBot` never returned nothing else than 2xx, right?

connectors/src/connectors/slack/temporal/client.ts
```diff
@@ -30,6 +36,45 @@ import {
...
 
+/**
+ * Starts one workflow per Slack webhook event. Start-only on purpose: this runs
+ * at Slack webhook rate, so it must never touch the database or the Slack API.
+ * Everything the event needs is resolved workflow side.
+ */
+export async function launchSlackWebhookEventWorkflow(
+  teamId: string,
+  eventId: string,
+  event: SlackWebhookEventPayload
```
flvndvd, line 47: `nit` can we use an object for both `teamId` and `eventId`?

connectors/src/connectors/slack/temporal/workflows.ts
```diff
@@ -21,6 +21,29 @@ const JOIN_CHANNEL_USE_CASES = [
 ] as const;
 export type JoinChannelUseCaseType = (typeof JOIN_CHANNEL_USE_CASES)[number];
 
+/**
+ * Projection of a Slack webhook event, built by the webhook handler and carried
+ * to `slackWebhookEventWorkflow`. Only the fields the handlers read are kept:
+ * message text, blocks and attachments never leave the webhook.
+ */
+export interface SlackWebhookEventPayload {
```
flvndvd, line 29: Should we use an union type here? Reading the comments it feels like fields have different shape based on the `type` no?

connectors/src/connectors/slack/temporal/webhook_activities.ts
```diff
@@ -0,0 +1,505 @@
...
+    }
+  }
+}
+
+async function handleAppMention(
+  event: SlackWebhookEventPayload,
+  teamId: string,
+  logger: Logger
+) {
+  if (!event.channelId || !event.ts) {
```
flvndvd, line 280: Would be lovely to validate those with a zod schema upfront so we don't need all the conditional checks in each handler.

connectors/src/connectors/slack/temporal/webhook_activities.ts
```diff
@@ -0,0 +1,505 @@
...
+        {
+          connectorId: connector.id,
+          slackChannelId: channelId,
+          skipReason: slackChannel.skipReason,
+        },
+        `Ignoring messages because channel is skipped: ${slackChannel.skipReason}`
+      );
+      continue;
+    }
+    if (!["read", "read_write"].includes(slackChannel.permission)) {
```
flvndvd, line 102: What about connectors in a "maintenance" mode?

connectors/src/connectors/slack/temporal/webhook_activities.ts
```diff
@@ -0,0 +1,505 @@
...
+  const targets = await listSyncableConnectors(
+    slackConfigurations,
+    channelId,
+    logger
+  );
+
+  for (const target of targets) {
+    const { connector } = target;
+
+    if (channelName) {
```
flvndvd, line 218: Not very clear that this variable is conditioned by the `channel_name` event. If we can improve a bit the code clarity here? Maybe switch/case over the event or zod validation at the top 🙃

connectors/src/connectors/slack/temporal/webhook_activities.ts
```diff
@@ -0,0 +1,505 @@
...
+        "Successfully processed Slack channel rename"
+      );
+    } else if (threadTs) {
+      const res = await launchSlackSyncOneThreadWorkflow(
+        connector.id,
+        channelId,
+        threadTs
+      );
+      if (res.isErr()) {
+        logger.error(
```
flvndvd, line 235: So we just swallow the error? It means we will never sync the thread. IMHO we should bubble the error

connectors/src/connectors/slack/temporal/webhook_activities.ts
```diff
@@ -0,0 +1,505 @@
...
+          {
+            err: res.error,
+            connectorId: connector.id,
+            slackChannelId: channelId,
+            messageTs,
+          },
+          "Failed to launch Slack messages sync"
```
flvndvd, line 260: Same here.

connectors/src/connectors/slack/temporal/webhook_activities.ts
```diff
@@ -0,0 +1,505 @@
...
+  logger: Logger
+) {
+  for (const slackConfiguration of slackConfigurations) {
+    const res = await launchSlackGarbageCollectWorkflow(
+      slackConfiguration.connectorId
+    );
+    if (res.isErr()) {
+      logger.error(
+        { err: res.error, connectorId: slackConfiguration.connectorId },
+        "Failed to launch Slack garbage collection"
```
flvndvd, line 446: Same question here.

connectors/src/connectors/slack/temporal/webhook_activities.ts
```diff
@@ -0,0 +1,505 @@
...
+/**
+ * Processes one Slack webhook event. The webhook handler does no work at all,
+ * so everything happens here: resolving the connectors, validating the channel
+ * and dispatching the sync.
+ *
+ * Slack has already been answered by the time this runs, so a failing event is
+ * logged and dropped rather than retried: replaying would re-post the Slack
+ * messages the handlers already sent.
+ */
+export async function processSlackWebhookEventActivity({
```
flvndvd, line 517: OOC where do we handle the `ExternalOAuthTokenError` here? Cause activities might churn as long as user did not re-auth.

flvndvd (approved):
Exciting, will be 10x better. Left some comments, fine to address in a follow up if needed.

connectors/src/api/webhooks/webhook_slack.ts
id13: Not assuming that at all. This is the webhook offloading 101 I'm applying: you ack 200 once it's in the queue, here the queue is not there but we already switch the operating model 👍

connectors/src/connectors/slack/temporal/workflows.ts
id13: fair, I brushed a bit quickly on event types, will try my luck with `@slack/webhook` as I would rather have vendored types than maintaining complex types myself.

connectors/src/connectors/slack/temporal/webhook_activities.ts
id13: 🤔

connectors/src/connectors/slack/temporal/webhook_activities.ts
id13: I can add a throw + activity retry with a backoff indeed

connectors/src/connectors/slack/temporal/webhook_activities.ts
id13: Fails at the connector workflow => we pause the connector => listSyncableConnectors in this activity is filtered out

id13:
Merging, addressing the typing in a follow-up

PR #29740, opened 2026-07-30
[core] enh: qdrant_reshard surfaces resharding stage and transfer progress from per-peer telemetry

id13:
## Description

Update qdrant_reshard to query managed cluster's metrics in order to have a better view of what's going on.


id13:
PMRR

PR #29751, opened 2026-07-30
Only update keys lastUpdatedAt once per hour

Fraggle:
## Description

`KeyResource.markAsUsed()` was issuing a DB write on every authenticated API request, causing row-lock contention on hot keys. Throttle writes to at most once per hour via `MARK_AS_USED_MIN_INTERVAL_MS = 3_600_000`. Also switches from `this.model.update` to `this.update` so the in-memory instance and Redis secret cache stay consistent after a write — previously, every cached hit would still trigger a write on the next request.

## Tests

Added 4 unit tests in `key_resource.test.ts`: first write (null → set), skip within the throttle window, write again after the window expires, and cache invalidation so subsequent fetches see the updated `lastUsedAt`.

## Risk

Low. `lastUsedAt` is a coarse "last seen" display field — skipping sub-hour updates has no functional impact. Rollback: revert to unconditional `this.model.update` in `markAsUsed`.

## Deploy Plan

Deploy front.

### DIFF
--- a/front/lib/resources/key_resource.test.ts
+++ b/front/lib/resources/key_resource.test.ts
@@ -77,7 +77,11 @@ vi.mock("@app/lib/utils/cache", () => ({
 
 import type { Authenticator } from "@app/lib/auth";
 import { GroupResource } from "@app/lib/resources/group_resource";
-import { KeyResource } from "@app/lib/resources/key_resource";
+import {
+  KeyResource,
+  MARK_AS_USED_MIN_INTERVAL_MS,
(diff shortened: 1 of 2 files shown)


Fraggle:

- #29751 ◀

id13:
Have you checked there are no side effect where lastUsedAt is used ?

Fraggle:
> Have you checked there are no side effect where lastUsedAt is used ?

Purely visual

id13:
ok LGTM

PR #29752, opened 2026-07-30
[front] fix: Dedupe conversation mark-as-read PATCHes

id13:
## Description

We PATCH `{read: true}` on conversations ~6x more often than we even GET them, same user, same conversation, sometimes 300ms apart. Went 8-10x worse on July 28 when pod conversations landed in the inbox (nudges all day × always-open tabs).

Three compounding issues:

- every trigger has its own 2s debounce (open-conversation effect + 3 SSE handlers), and all that state lives in hooks that mount/unmount constantly, so nothing ever coalesces
- the conversation detail GET never hydrates read state: every conversation is served `unread: true` / `lastReadMs: null`, so we re-mark already-read conversations on every open/revalidation
- the endpoint upserts lastReadAt unconditionally

This PR:

- moves debounce + dedup to module level, above the mount/unmount churn: every trigger funnels into `requestConversationMarkAsRead`, each request carries the activity timestamp it must cover (`conversation.updated` from the effect, `event.created` from SSE all server clock), anything already covered is dropped. Hidden tabs park the mark until visible again, 403/404 blacklists the conversation (zombie tabs were PATCHing a 404 forever)
- hydrates `unread` / `lastReadMs` on the detail GET (`fetchByIdWithReadState`), which also feeds the dedup floor
- skips the lastReadAt write when the conversation is already read up to its latest activity

The effect firing on the conversation value stays: it's the intended "new activity while viewing → mark read" behavior, the global dedup just makes the churn free.

## Tests

- Added endpoint tests (read-state hydration, write skip)
- Tested locally
- Front-edged

## Risk

Low - same triggers as before, one shared pipe. Worst case a conversation stays bold in the sidebar until the next activity/revalidation.

## Plan

Deploy front

### DIFF
--- a/front-api/routes/w/[wId]/assistant/conversations/[cId]/index.ts
+++ b/front-api/routes/w/[wId]/assistant/conversations/[cId]/index.ts
@@ -200,9 +200,13 @@ app.get(
     const auth = ctx.get("auth");
     const { cId } = ctx.req.valid("param");
 
-    const conversation = await ConversationResource.fetchById(auth, cId, {
-      includeForkingData: true,
-    });
+    // Read state must be hydrated: without it the serialized conversation always
+    // reports `unread: true` / `lastReadMs: null`.
+    const conversation = await ConversationResource.fetchByIdWithReadState(
+      auth,
+      cId,
+      { includeForkingData: true }
+    );
 
     if (!conversation) {
       // Distinguish between "not found" and "access restricted" for the UI.
@@ -286,9 +290,14 @@ app.patch(
 
     if ("read" in data) {
       if (data.read) {
-        await ConversationResource.markAsReadForAuthUser(auth, {
-          conversation,
-        });
+        // Clients re-send this on many triggers: skip the write when the conversation is
+        // already read up to its latest activity. This also preserves future-dated
+        // `lastReadAt` stamps (see markAsReadForAuthUser) instead of clobbering them.
+        if (conversation.unread) {
+          await ConversationResource.markAsReadForAuthUser(auth, {
+            conversation,
+          });
+        }
 
         // A stale `actionRequired` flag (e.g. left behind by a manual tool approval whose
         // message got interrupted before blocked actions were resolved on termination) would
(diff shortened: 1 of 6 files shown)


front/hooks/conversations/useConversationMarkAsRead.ts
```diff
@@ -31,14 +33,39 @@ export function useConversationMarkAsRead({
     },
   });
 
+  // We only need the mutator: without writing `unread: false` back into the conversation
+  // cache after a successful PATCH, the effect below re-fires on every conversation object
+  // change until the next revalidation.
```
id13: I am not sure about this

front/hooks/conversations/useConversationMarkAsRead.ts
```diff
@@ -99,33 +148,47 @@ export function useConversationMarkAsRead({
...
   );
 
+  // Single debounced channel shared by every trigger (open-conversation effect, SSE events,
+  // visibility catch-up) so concurrent triggers coalesce into one PATCH.
+  const debouncedMarkAsRead = useMemo(
+    () => debounce(markAsRead, DELAY_BEFORE_MARKING_AS_READ_MS),
+    [markAsRead]
+  );
+
+  useEffect(() => () => debouncedMarkAsRead.cancel(), [debouncedMarkAsRead]);
```
id13: I think this is still not a single debounced channel ?

id13:

- #29752 ◀

front-api/routes/w/[wId]/assistant/conversations/[cId]/index.ts
```diff
@@ -286,9 +290,14 @@ app.patch(
 
     if ("read" in data) {
       if (data.read) {
-        await ConversationResource.markAsReadForAuthUser(auth, {
-          conversation,
-        });
+        // Clients re-send this on many triggers: skip the write when the conversation is
+        // already read up to its latest activity. This also preserves future-dated
+        // `lastReadAt` stamps (see markAsReadForAuthUser) instead of clobbering them.
+        if (conversation.unread) {
```
Comment (Fraggle) [inline front-api/routes/w/[wId]/assistant/conversations/[cId]/index.ts:296]: nice

Fraggle (approved):
Look solid, thanks.

PR #29753, opened 2026-07-30
Users can add a reason when requesting overage

avervaet:
## Description

This PR lets members add optional context when requesting a usage-limit upgrade. Previously, the request contained no explanation, so admins and managers received only the generic upgrade request.

The upgrade dialog now includes a “Help your admin decide” text area. The reason is trimmed, limited to 1,024 characters, and sent with the existing request. Empty input remains valid and is omitted from the request. The API validates the payload and persists the reason on `membership_upgrade_requests`; existing requests remain compatible because the new database column is nullable.

The reason is returned through the upgrade-request resource, displayed in the admin `UpgradeRequestsTable` with a truncated two-line presentation and the full value available on hover, and included in the admin/manager notification workflow. The `membership.upgrade_request_created` audit event also records the reason. Existing request creation, approval, denial, and idempotency behavior is otherwise unchanged.

## Tests

Added regression coverage in `front-api/routes/w/[wId]/credits/upgrade-requests.test.ts` verifying that a supplied reason is persisted and returned by the request endpoint.

Localy tested

## Risk

This change adds a nullable database column and propagates user-provided text through the API response, admin UI, notifications, and the upgrade-request audit event. The reason is bounded to 1,024 characters and remains optional, preserving compatibility with existing requests and callers that do not provide it.

The WorkOS audit-log schema must be updated before deployment so the new `reason` metadata is accepted. Rollback should account for the already-applied nullable column and registered audit schema; reverting the application code does not require removing either immediately.

## Deploy Plan

1 Apply `front/migrations/pre-deploy/20260727135138_add_reason_to_membership_upgrade_requests.sql` in both regions.
2. Deploy `front`.

### DIFF
--- a/front/admin/audit_log_schemas/membership.upgrade_request_created.json
+++ b/front/admin/audit_log_schemas/membership.upgrade_request_created.json
@@ -9,6 +9,7 @@
     }
   ],
   "metadata": {
-    "request_sid": "string"
+    "request_sid": "string",
+    "reason": "string"
   }
 }
--- a/front/components/credits/UsageUpgradeButton.tsx
+++ b/front/components/credits/UsageUpgradeButton.tsx
@@ -139,7 +139,7 @@ export function UsageUpgradeButton({
             <DialogContainer>
               <div className="flex flex-col gap-4">
                 <p className="text-sm text-muted-foreground">
-                  Your workspace admins will review this request.
+                  Your workspace admins and managers will review this request.
                 </p>
                 <BaseFormFieldSection<HTMLTextAreaElement>
                   fieldName="reason"
--- a/front/components/credits/UsageUpgradeButton.tsx
+++ b/front/components/credits/UsageUpgradeButton.tsx
@@ -1,4 +1,6 @@
+import { BaseFormFieldSection } from "@app/components/shared/BaseFormFieldSection";
 import { useRequestUpgrade } from "@app/lib/swr/upgrade_requests";
+import { MAX_UPGRADE_REQUEST_REASON_LENGTH_CHARS } from "@app/types/memberships";
 import type { LightWorkspaceType } from "@app/types/user";
 import {
   Button,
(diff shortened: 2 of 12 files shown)


front/components/credits/UsageUpgradeButton.tsx
```diff
@@ -117,30 +132,64 @@ export function UsageUpgradeButton({
...
-            }}
-          />
+          <FormProvider {...form}>
+            <DialogHeader>
+              <DialogTitle>Request a usage limit upgrade</DialogTitle>
+            </DialogHeader>
+            <DialogContainer>
+              <div className="flex flex-col gap-4">
+                <p className="text-sm text-muted-foreground">
+                  Your workspace admins will review this request.
```
aubin-tchoi: OOC why change the copy here?

aubin-tchoi (approved):
LGTM

front/lib/notifications/workflows/upgrade-request-created.ts
```diff
@@ -141,6 +158,7 @@ export function notifyUpgradeRequested({
     workspaceName,
     requesterName,
     requesterEmail,
+    reason,
```
aubin-tchoi, line 161: idk if there is something we need to handle w.r.t Novu

front/admin/audit_log_schemas/membership.upgrade_request_created.json
```diff
@@ -9,6 +9,7 @@
     }
   ],
   "metadata": {
-    "request_sid": "string"
+    "request_sid": "string",
```
aubin-tchoi, line 12: I think you may need to bump the version for the audit logs

front/types/memberships.ts
```diff
@@ -300,11 +300,15 @@ export const MEMBERSHIP_UPGRADE_REQUEST_STATUSES = [
...
   (typeof MEMBERSHIP_UPGRADE_REQUEST_STATUSES)[number];
 
+export const MAX_UPGRADE_REQUEST_REASON_LENGTH_CHARS = 1024;
+
 export interface MembershipUpgradeRequestType {
   sId: string;
   status: MembershipUpgradeRequestStatus;
   createdAt: number;
   resolvedAt: number | null;
+  // Why the member needs the raised limit - Optional
```
flvndvd: Models tend to be very verbose, do not hesitate to drop lots of irrelevant comments.

flvndvd (approved):

PR #29754, opened 2026-07-30
feat: Admin approval flow for overage requests

avervaet:
## Description

Until now, pending overage requests exposed separate “Upgrade plan” and “Edit limit” actions, which made it difficult for admins to choose the intended approval behavior. This PR replaces that layout with a single **Manage** action offering:

- **Set credit amount**, with a custom monthly limit and expiry.
- **Upgrade User Plan**, when the workspace is not enterprise and a higher seat tier is available.

The custom-limit flow displays the member’s current allowed overage and consumption, warns when the new limit is lower than the current one, and supports expiry presets for **1 day**, **the next credit refresh**, or **Never**. Spend-limit responses now expose the persisted expiry and the next workspace credit reset derived from the cached Metronome billing period. Audit events include the expiry timestamp as well.

Approving a request can now apply the selected spend limit as part of the same API operation. The request is marked resolved first, then the spend-limit change is synchronized. If the downstream sync fails, the request is reverted to `pending`; rollback failures are logged for manual reconciliation. Denials and the existing plan-upgrade flow remain supported.

This PR relies on the spend-limit expiry data model and scheduler work in the related stack: #29921, #29960, and #29961.

Admin overage request management actions

Capture d’écran 2026-08-11 à 17 29 46

Capture d’écran 2026-08-11 à 17 30 19

Spend limit expiry options

## Tests

The spend-limit API tests were updated to cover nullable expiry fields and persisted `expiresAt` values.

## Risk

Medium. This changes the admin approval path for overage requests and can directly allow to change an user capped spend. The new limit form validates credit bounds and future expiry dates, and unlimited approval requires confirmation.

The approval flow uses rollback protection to avoid leaving a request resolved when its spend-limit synchronization fails. Expiry behavior also depends on the companion Temporal worker and hourly schedule being deployed; without that scheduler, expiring overrides may not be automatically reverted.

The API changes are additive for spend-limit reads, while the audit schema version is incremented to include `expires_at`. No migration is included in this PR; it expects the existing spend-limit expiry column from the related stack.

## Deploy Plan

- Deploy `front-sse` and `front`.

### DIFF
--- a/front-api/routes/w/[wId]/members/[uId]/spend_limit.ts
+++ b/front-api/routes/w/[wId]/members/[uId]/spend_limit.ts
@@ -1,81 +1,25 @@
 import { getAuditLogContext } from "@app/lib/api/audit/workos_audit";
 import {
   getUserSpendLimit,
-  MAX_USER_SPEND_LIMIT_AWU_CREDITS,
-  MIN_USER_SPEND_LIMIT_AWU_CREDITS,
   setUserSpendLimit,
-  type UserSpendLimitError,
 } from "@app/lib/api/users/spend_limit";
 import type {
   GetUserSpendLimitResponseBody,
   PutUserSpendLimitResponseBody,
 } from "@app/types/api/users/spend_limit";
-import type { APIErrorWithContentfulStatusCode } from "@app/types/error";
-import { assertNever } from "@app/types/shared/utils/assert_never";
 import { workspaceApp } from "@front-api/middlewares/ctx";
 import { ensureIsManager } from "@front-api/middlewares/ensure_role";
 import { apiError, type HandlerResult } from "@front-api/middlewares/utils";
 import { validate } from "@front-api/middlewares/validator";
+import {
+  spendLimitErrorToApiError,
+  UpdateUserSpendLimitBodySchema,
+} from "@front-api/routes/w/[wId]/user_spend_limit_shared";
 import { z } from "zod";
-
-export const UpdateUserSpendLimitBodySchema = z.discriminatedUnion("kind", [
-  z.object({ kind: z.literal("unlimited") }),
-  z.object({
-    kind: z.literal("limited"),
-    awuCredits: z
-      .number()
-      .int()
-      .min(MIN_USER_SPEND_LIMIT_AWU_CREDITS)
-      .max(MAX_USER_SPEND_LIMIT_AWU_CREDITS),
-    // Epoch ms at which the override auto-reverts to unlimited.
-    // Omitted/null means it never expires.
-    expiresAt: z
-      .number()
-      .int()
-      .positive()
-      .refine((value) => value > Date.now(), {
-        message: "expiresAt must be in the future.",
-      })
-      .nullish(),
-  }),
-]);
 
 const ParamsSchema = z.object({
   uId: z.string(),
 });
-
-export function spendLimitErrorToApiError(
-  error: UserSpendLimitError
-): APIErrorWithContentfulStatusCode {
-  switch (error.type) {
-    case "user_not_found":
-      return {
-        status_code: 404,
-        api_error: {
-          type: "workspace_user_not_found",
-          message: error.message,
-        },
-      };
-    case "workspace_not_metronome_billed":
-      return {
-        status_code: 403,
-        api_error: {
-          type: "plan_limit_error",
-          message: error.message,
-        },
-      };
-    case "metronome_error":
-      return {
-        status_code: 502,
-        api_error: {
-          type: "internal_server_error",
-          message: "Failed to update spend limit in billing system.",
-        },
-      };
-    default:
-      assertNever(error.type);
-  }
-}
 
 // Mounted at /api/w/:wId/members/:uId/spend_limit.
 const app = workspaceApp();
--- a/front-api/routes/w/[wId]/members/[uId]/spend_limit.ts
+++ b/front-api/routes/w/[wId]/members/[uId]/spend_limit.ts
@@ -18,7 +18,7 @@ import { apiError, type HandlerResult } from "@front-api/middlewares/utils";
 import { validate } from "@front-api/middlewares/validator";
 import { z } from "zod";
 
-const UpdateUserSpendLimitBodySchema = z.discriminatedUnion("kind", [
+export const UpdateUserSpendLimitBodySchema = z.discriminatedUnion("kind", [
   z.object({ kind: z.literal("unlimited") }),
   z.object({
     kind: z.literal("limited"),
@@ -27,6 +27,16 @@ const UpdateUserSpendLimitBodySchema = z.discriminatedUnion("kind", [
       .int()
       .min(MIN_USER_SPEND_LIMIT_AWU_CREDITS)
       .max(MAX_USER_SPEND_LIMIT_AWU_CREDITS),
+    // Epoch ms at which the override auto-reverts to unlimited.
+    // Omitted/null means it never expires.
+    expiresAt: z
+      .number()
+      .int()
+      .positive()
+      .refine((value) => value > Date.now(), {
+        message: "expiresAt must be in the future.",
+      })
+      .nullish(),
   }),
 ]);
 
@@ -34,7 +44,7 @@ const ParamsSchema = z.object({
   uId: z.string(),
 });
 
-function spendLimitErrorToApiError(
+export function spendLimitErrorToApiError(
   error: UserSpendLimitError
 ): APIErrorWithContentfulStatusCode {
   switch (error.type) {
--- a/front/admin/audit_log_schemas/member.spend_limit_updated.json
+++ b/front/admin/audit_log_schemas/member.spend_limit_updated.json
@@ -10,6 +10,7 @@
   ],
   "metadata": {
     "kind": "string",
-    "awu_credits": "string"
+    "awu_credits": "string",
+    "expires_at": "string"
   }
 }
--- a/front/components/pages/workspace/UsagePage.tsx
+++ b/front/components/pages/workspace/UsagePage.tsx
@@ -1230,7 +1230,6 @@ export function UsagePage() {
         }}
         member={editSpendLimitMember}
         owner={owner}
-        forceOverride={pendingApproveRequestId !== null}
         onSavingChange={handleUsagePendingChange}
         upgradeRequestId={pendingApproveRequestId}
       />
--- a/front/components/pages/workspace/UsagePage.tsx
+++ b/front/components/pages/workspace/UsagePage.tsx
@@ -339,31 +339,58 @@ export function UsagePage() {
     },
     []
   );
-  const handleEditLimitRequest = useCallback(
+  const handleSetCreditAmountRequest = useCallback(
     (request: MembershipUpgradeRequestType) => {
       setPendingApproveRequestId(request.sId);
       setEditSpendLimitMember(memberFromUpgradeRequest(request));
     },
     []
   );
-  const handleApproveOnModalSaved = useCallback(() => {
+  // Shared by approval flows before resolving the request.
+  const resolveRequestApproved = useCallback(
+    async (requestId: string): Promise<void> => {
+      await doResolveUpgradeRequest({
+        requestId,
+        resolution: { status: "approved" },
+      });
+    },
+    [doResolveUpgradeRequest]
+  );
+  const handleApproveOnModalSaved = useCallback(async () => {
     if (!pendingApproveRequestId) {
       return;
     }
     const requestId = pendingApproveRequestId;
-    const request = upgradeRequests.find((r) => r.sId === requestId);
     setRequestResolving(requestId, true);
-    void doResolveUpgradeRequest({
-      requestId,
-      requesterName: request?.requester.name ?? "Member",
-      status: "approved",
-    }).finally(() => setRequestResolving(requestId, false));
-  }, [
-    pendingApproveRequestId,
-    upgradeRequests,
-    doResolveUpgradeRequest,
-    setRequestResolving,
-  ]);
+    try {
+      await resolveRequestApproved(requestId);
+    } finally {
+      setRequestResolving(requestId, false);
+    }
+  }, [pendingApproveRequestId, resolveRequestApproved, setRequestResolving]);
+  const handleAllowUnlimitedSpendRequest = useCallback(
+    async (request: MembershipUpgradeRequestType) => {
+      const confirmed = await confirm({
+        title: "Use workspace default",
+        message: `Remove ${request.requester.name}'s custom spend limit? They'll fall back to their group or workspace default cap.`,
+        validateLabel: "Use workspace default",
+        validateVariant: "primary",
+      });
+      if (!confirmed) {
+        return;
+      }
+      setRequestResolving(request.sId, true);
+      try {
+        await doResolveUpgradeRequest({
+          requestId: request.sId,
+          resolution: { status: "approved", limit: { kind: "unlimited" } },
+        });
+      } finally {
+        setRequestResolving(request.sId, false);
+      }
+    },
+    [confirm, doResolveUpgradeRequest, setRequestResolving]
+  );
   const handleDenyRequest = useCallback(
     async (request: MembershipUpgradeRequestType) => {
       const confirmed = await confirm({
@@ -379,8 +406,7 @@ export function UsagePage() {
       try {
         await doResolveUpgradeRequest({
           requestId: request.sId,
-          requesterName: request.requester.name,
-          status: "denied",
+          resolution: { status: "denied" },
         });
       } finally {
         setRequestResolving(request.sId, false);
@@ -1140,9 +1166,11 @@ export function UsagePage() {
                       requests={filteredUpgradeRequests}
                       isLoading={isUpgradeRequestsLoading}
                       seatPlans={seatPlans}
+                      isEnterprise={isEnterprise}
                       pendingRequestIds={resolvingRequestIds}
                       onUpgradePlan={handleUpgradePlanRequest}
-                      onEditLimit={handleEditLimitRequest}
+                      onAllowUnlimitedSpend={handleAllowUnlimitedSpendRequest}
+                      onSetCreditAmount={handleSetCreditAmountRequest}
                       onDeny={handleDenyRequest}
                     />
                   )}
@@ -1226,8 +1254,9 @@ export function UsagePage() {
         }}
         member={editSpendLimitMember}
         owner={owner}
+        forceOverride={pendingApproveRequestId !== null}
         onSavingChange={handleUsagePendingChange}
-        onSaved={handleApproveOnModalSaved}
+        upgradeRequestId={pendingApproveRequestId}
       />
 
       <BulkEditSpendLimitModal
--- a/front/components/pages/workspace/UsagePage.tsx
+++ b/front/components/pages/workspace/UsagePage.tsx
@@ -368,29 +368,6 @@ export function UsagePage() {
       setRequestResolving(requestId, false);
     }
   }, [pendingApproveRequestId, resolveRequestApproved, setRequestResolving]);
-  const handleAllowUnlimitedSpendRequest = useCallback(
-    async (request: MembershipUpgradeRequestType) => {
-      const confirmed = await confirm({
-        title: "Use workspace default",
-        message: `Remove ${request.requester.name}'s custom spend limit? They'll fall back to their group or workspace default cap.`,
-        validateLabel: "Use workspace default",
-        validateVariant: "primary",
-      });
-      if (!confirmed) {
-        return;
-      }
-      setRequestResolving(request.sId, true);
-      try {
-        await doResolveUpgradeRequest({
-          requestId: request.sId,
-          resolution: { status: "approved", limit: { kind: "unlimited" } },
-        });
-      } finally {
-        setRequestResolving(request.sId, false);
-      }
-    },
-    [confirm, doResolveUpgradeRequest, setRequestResolving]
-  );
   const handleDenyRequest = useCallback(
     async (request: MembershipUpgradeRequestType) => {
       const confirmed = await confirm({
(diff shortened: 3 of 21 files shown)


Fraggle:
@avervaet 

> Temporal — new spend_limit_expiration queue: needs a worker deployed and its hourly schedule registered (admin/cli.sh start) @Fraggle - could need a hand for this part :)

If you can re-use an existing queue, you'll simplify your life :)
I think for starting it, you need to log in to each region's prodbox and start it.

front/temporal/spend_limit_expiration/client.ts
```diff
@@ -0,0 +1,72 @@
...
+  const scheduleOptions = {
+    action: {
+      type: "startWorkflow" as const,
+      workflowType: expirePoolCapOverridesWorkflow,
+      args: [],
+      taskQueue: QUEUE_NAME,
+    },
+    scheduleId,
+    policies: {
+      overlap: ScheduleOverlapPolicy.SKIP,
```
aubin-tchoi: OOC are we sure we want to skip rather than buffer one?

front/temporal/spend_limit_expiration/activities.ts
```diff
@@ -0,0 +1,77 @@
...
+        now,
+      });
+    if (memberships.length === 0) {
+      continue;
+    }
+
+    const users = await UserResource.fetchByModelIds(
+      memberships.map((membership) => membership.userId)
+    );
+    const userSIdByModelId = new Map(users.map((user) => [user.id, user.sId]));
```
aubin-tchoi: nit: 
```suggestion
    const userIdByModelId = new Map(users.map((user) => [user.id, user.sId]));
```

front/lib/api/users/spend_limit.ts
```diff
@@ -339,12 +374,145 @@ export async function setUserSpendLimit(
...
+ * Revert an expired pool cap override back to the seat-type default. Called
+ * by the expiration sweep (`@app/temporal/spend_limit_expiration`) — never by
+ * an admin action, hence the system-actored, dedicated audit event rather
+ * than reusing `setUserSpendLimit`'s `member.spend_limit_updated`. A no-op
+ * (idempotent `Ok`) if the override was already cleared or never expires,
+ * e.g. an admin manually reverted it before the sweep ran.
+ */
+export async function expireUserSpendLimitOverride(
+  auth: Authenticator,
+  { userId }: { userId: string }
```
aubin-tchoi: you could pass the entire `UserResource` here since you already have it in the call site, would remove the need for the N+1 fetch here

front/lib/api/users/spend_limit.ts
```diff
@@ -339,12 +374,145 @@ export async function setUserSpendLimit(
...
+    return new Err(
+      new UserSpendLimitError(
+        "user_not_found",
+        "Could not find the user in this workspace."
+      )
+    );
+  }
+
+  const membership =
+    await MembershipResource.getActiveMembershipOfUserInWorkspace({
```
aubin-tchoi: don't we already have a membership at call site? ideally we'd avoid the N+1 pattern here

front/lib/api/users/spend_limit.ts
```diff
@@ -339,12 +374,145 @@ export async function setUserSpendLimit(
...
+        userId: user.sId,
+        err: clearWarningResult.error,
+      },
+      "[SpendLimitExpiration] Failed to clear warning alert; continuing"
+    );
+  }
+
+  const metronomeContractId = auth.subscription()?.metronomeContractId ?? null;
+  if (metronomeContractId) {
+    const workspaceResource = await WorkspaceResource.fetchById(workspace.sId);
```
aubin-tchoi: you don't need to refetch the workspace do you?

front/lib/api/users/spend_limit.ts
```diff
@@ -339,12 +374,145 @@ export async function setUserSpendLimit(
...
+    if (workspaceResource) {
+      void reconcileUser({
```
aubin-tchoi: do we really need to void and catch it?

front/components/workspace/EditSpendLimitModal.tsx
```diff
@@ -191,6 +254,83 @@ export function EditSpendLimitModal({
...
+      </div>
+      {displayedMember && (
+        <div className="flex flex-col gap-0.5">
+          <p className="text-xs text-muted-foreground">
+            Current allowed overage:{" "}
+            {(currentAwuCredits ?? 0).toLocaleString("en-US")} credits
+          </p>
+          <p className="text-xs text-muted-foreground">
+            Current overage consumption:{" "}
+            {displayedMember.consumedFromPoolAwuCredits.toLocaleString("en-US")}{" "}
```
aubin-tchoi: nit:
```suggestion
            {displayedMember.consumedFromPoolAwuCredits.toLocaleString("en-US")}&nbsp;
```

front/components/workspace/EditSpendLimitModal.tsx
```diff
@@ -191,6 +254,83 @@ export function EditSpendLimitModal({
...
     : validateDisabled;
 
+  const overrideForm = (
+    <div className="flex flex-col gap-1.5 pl-6">
+      <div className="relative">
+        <Input
+          id="spend-credit-limit-input"
+          type="text"
+          inputMode="numeric"
+          pattern="[0-9]*"
```
aubin-tchoi: OOC do you need `pattern` if you passed `inputMode` numeric?

aubin-tchoi: ah maybe to prevent typing in negative values

front/components/workspace/EditSpendLimitModal.tsx
```diff
@@ -191,6 +254,83 @@ export function EditSpendLimitModal({
...
+          placeholder="1,000"
+          value={
+            creditsInput !== "" ? Number(creditsInput).toLocaleString() : ""
+          }
+          onChange={(e) => handleCreditsChange(e.target.value)}
+          isError={validationMessage !== null}
+          message={validationMessage ?? undefined}
+          messageStatus={validationMessage !== null ? "error" : undefined}
+          className="pr-28 text-right"
```
aubin-tchoi: that is a lot of padding, aren't you hacking the layout here?

front-api/routes/w/[wId]/members/[uId]/spend_limit.ts
```diff
@@ -27,6 +27,9 @@ const UpdateUserSpendLimitBodySchema = z.discriminatedUnion("kind", [
       .int()
       .min(MIN_USER_SPEND_LIMIT_AWU_CREDITS)
       .max(MAX_USER_SPEND_LIMIT_AWU_CREDITS),
+    // Epoch ms at which the override auto-reverts to unlimited. Omitted/null
+    // means it never expires.
+    expiresAt: z.number().int().positive().nullable().optional(),
```
Comment (aubin-tchoi) [inline front-api/routes/w/[wId]/members/[uId]/spend_limit.ts]: FYI you can do `.nullish()` to do `.nullable().optional()`

front/components/pages/workspace/UsagePage.tsx
```diff
@@ -339,31 +340,69 @@ export function UsagePage() {
...
+  // the request itself, independent of whatever already updated the seat or
+  // spend limit.
+  const resolveRequestApproved = useCallback(
+    (requestId: string): Promise<void> => {
+      const request = upgradeRequests.find((r) => r.sId === requestId);
+      return doResolveUpgradeRequest({
+        requestId,
+        requesterName: request?.requester.name ?? "Member",
+        status: "approved",
+      }).then(() => undefined);
```
aubin-tchoi: nit: you can just `await` `doResolveUpgradeRequest` and not do the `.then`

front/components/pages/workspace/UsagePage.tsx
```diff
@@ -339,31 +340,69 @@ export function UsagePage() {
...
-      requesterName: request?.requester.name ?? "Member",
-      status: "approved",
-    }).finally(() => setRequestResolving(requestId, false));
-  }, [
-    pendingApproveRequestId,
-    upgradeRequests,
-    doResolveUpgradeRequest,
-    setRequestResolving,
-  ]);
+    void resolveRequestApproved(requestId).finally(() =>
```
aubin-tchoi: OOC why void it?

front/components/workspace/EditSpendLimitModal.tsx
```diff
@@ -24,18 +31,39 @@ import { useEffect, useRef, useState } from "react";
...
   return value === "default" || value === "override";
 }
 
+type ExpiryMode = SpendLimitExpiryKind;
+
+function isExpiryMode(value: string): value is ExpiryMode {
+  return (SPEND_LIMIT_EXPIRY_KINDS as readonly string[]).includes(value);
+}
+
+function formatShortDate(epochMs: number): string {
```
aubin-tchoi: surprised we don't already have a function that does this

front/components/workspace/EditSpendLimitModal.tsx
```diff
@@ -156,17 +193,35 @@ export function EditSpendLimitModal({
     try {
       let limit:
         | { kind: "unlimited" }
-        | { kind: "limited"; awuCredits: number };
+        | { kind: "limited"; awuCredits: number; expiresAt: number | null };
```
aubin-tchoi: could be typed with `UserSpendLimit`

front/lib/resources/membership_resource.ts
```diff
@@ -1598,12 +1598,88 @@ export class MembershipResource extends BaseResource<MembershipModel> {
...
+    // biome-ignore lint/plugin/noRawSql: cross-workspace lookup, see WORKSPACE_ISOLATION_BYPASS below
+    const rows = await frontSequelize.query<{ workspaceId: ModelId }>(
+      `
+      SELECT DISTINCT "workspaceId"
+      FROM public.memberships
+      WHERE "poolCapOverrideExpiresAt" IS NOT NULL
+        AND "poolCapOverrideExpiresAt" <= :now
+        AND "poolCapOverrideAwuCredits" IS NOT NULL
+        AND "endAt" IS NULL
+      ORDER BY "workspaceId" ASC;
```
aubin-tchoi: let's write it in Sequelize

front/lib/resources/membership_resource.ts
```diff
@@ -1598,12 +1598,88 @@ export class MembershipResource extends BaseResource<MembershipModel> {
...
+  }
+
+  /**
+   * Workspaces (sorted by id ascending) that have at least one active
+   * membership whose pool cap override has expired. Global, cross-workspace
+   * lookup for the expiration sweep — mirrors
+   * `WebhookRequestResource.getWorkspacesWithTooManyRequests`: raw SQL to find
+   * affected workspaces, then per-workspace scoped work from the caller.
+   */
+  static async getWorkspacesWithExpiredPoolCapOverride(
```
aubin-tchoi: let's rename this to `dangerouslyGetWorkspacesWithExpiredPoolCapOverride` to flag that it's meant to be used exclusively on temporal admin jobs/poke plugins and not in API routes

aubin-tchoi: OOC why have this in membership resource rather than workspace resource?

front/lib/resources/membership_resource.ts
```diff
@@ -1598,12 +1598,88 @@ export class MembershipResource extends BaseResource<MembershipModel> {
...
+  }
+
+  /**
+   * Active memberships within `workspace` whose pool cap override has
+   * expired as of `now`. Scoped counterpart to
+   * `getWorkspacesWithExpiredPoolCapOverride`, called once per affected
+   * workspace by the expiration sweep.
+   */
+  static async listActiveWithExpiredPoolCapOverride({
+    workspace,
```
aubin-tchoi: let's pass an Authenticator rather than a `LightWorkspaceType`

front/lib/resources/membership_resource.ts
```diff
@@ -1598,12 +1598,88 @@ export class MembershipResource extends BaseResource<MembershipModel> {
...
+    const rows = await MembershipModel.findAll({
+      where: {
+        workspaceId: workspace.id,
+        endAt: null,
+        poolCapOverrideAwuCredits: { [Op.ne]: null },
+        poolCapOverrideExpiresAt: { [Op.lte]: now },
+      },
+    });
+    return rows.map(
+      (row) => new MembershipResource(MembershipModel, row.get())
```
aubin-tchoi: nit:
```suggestion
      (row) => new this(this.model, row.get())
```

front/lib/api/users/spend_limit.ts
```diff
@@ -339,12 +374,145 @@ export async function setUserSpendLimit(
...
+        userId: user.sId,
+        err: clearResult.error,
+      },
+      "[SpendLimitExpiration] Failed to clear per-user cap alert on expiry"
+    );
+    return new Err(
+      new UserSpendLimitError("metronome_error", clearResult.error.message)
+    );
+  }
+  const clearWarningResult = await clearMetronomePerUserWarningAlert({
```
aubin-tchoi: if this fails we'll end up with a mismatch between what we have in metronome and what we have in db

front/components/workspace/EditSpendLimitModal.tsx
```diff
@@ -92,16 +128,18 @@ export function EditSpendLimitModal({
       case "limited":
         setKind("override");
         setCreditsInput(String(spendLimit.awuCredits));
+        setExpiryMode(spendLimit.expiresAt ? "one_day" : "never");
```
aubin-tchoi: are we sure we want to map to "one_day" by default? looks like if you edit an existing expiry you'll necessarily reset it to `Date.now() + ONE_DAY_MS;`

front-api/routes/w/[wId]/members/[uId]/spend_limit.ts
Comment (aubin-tchoi) [inline front-api/routes/w/[wId]/members/[uId]/spend_limit.ts]: the endpoint should reject `expiresAt` in the past

front/temporal/spend_limit_expiration/worker.ts
```diff
@@ -0,0 +1,42 @@
...
+
+import { QUEUE_NAME } from "./config";
+
+// Must match the deployment's terminationGracePeriodSeconds minus 10s buffer.
+const SHUTDOWN_GRACE_TIME_MS = 70 * 1_000;
+
+export async function runSpendLimitExpirationWorker() {
+  const { connection, namespace } = await getTemporalWorkerConnection();
+
+  const worker = await Worker.create({
```
aubin-tchoi: you're gonna need a change in dust-infra to actually run this worker FYI

front/lib/resources/storage/models/membership.ts
```diff
@@ -79,6 +84,11 @@ MembershipModel.init(
       allowNull: true,
       defaultValue: null,
     },
+    poolCapOverrideExpiresAt: {
+      type: DataTypes.DATE,
+      allowNull: true,
```
aubin-tchoi, line 89: should be false

aubin-tchoi (commented):
LGTM overall
quite a big PR with a few different goals tackled at once, could have been split

front/components/workspace/EditSpendLimitModal.tsx
avervaet: There actually is one — `formatTimestampToFriendlyDate(timestamp, "compactWithDay")` in `front/lib/utils.ts` produces the identical `Mon D, YYYY` format. The one difference is it doesn't pin `timeZone: "UTC"`, and this label is showing `nextCreditResetAt`, a billing-cycle boundary computed in UTC — without the pin, users near midnight in non-UTC timezones could see the wrong day for their actual cycle reset. Keeping the local helper for that reason rather than reusing the shared one.

front/lib/api/users/spend_limit.ts
avervaet: Good catch, this is real: since the DB override is cleared before the Metronome clear call, a failure here means this membership no longer matches `listActiveWithExpiredPoolCapOverride`'s query and drops out of the sweep permanently — no automatic retry.

I checked whether anything else would catch this drift later: `reconcileUser` only reconciles the credit-state balance/usage, never touches Metronome alerts; `recalculatePerUserCapAlertForSeatChange` explicitly no-ops when `poolCapOverrideAwuCredits` is null (exactly this state). So confirmed there's no existing recovery path — it'd need manual Metronome intervention.

I considered flipping the order (Metronome first, then DB), but that trades this failure mode for a worse one: if the DB write then failed, Metronome would show unlimited while the DB still shows capped — i.e., a user could spend unbounded while we think they're still restricted. Today's order fails safe (user stays capped longer than intended) rather than fails open. Every other write in this file is deliberately DB-first for the same reason ("the membership is the source of truth, Metronome alerts are derived enforcement"), so I'd rather keep that consistent here too.

Went with keeping the order but making the failure log actionable instead: added `previousAwuCredits` to the log context and made the message explicit that this won't be retried automatically and needs manual Metronome intervention, so it's at least clear to whoever picks up the alert what happened and what to do about it.

front/lib/resources/storage/models/membership.ts
avervaet: null is intentional here — it captures an overage override with no expiry (never reverts automatically). That's a meaningful, commonly-used state (e.g. `updatePoolCapOverride` explicitly writes `null` for that case), so keeping `allowNull: true` on purpose.

aubin-tchoi: indeed, misread this

front/lib/api/users/spend_limit.ts
aubin-tchoi: in this case we'd want some alerting on this

avervaet: Okay, never let Claude answer comments directly from my machine again, lots of bullshit.

Long story short, there are two points of failure regarding this issue. One is here: we write an overage to the DB, but the Metronome update fails, leaving the user with alerts and blocked usage despite the admin action.

The other point of failure is the clean-up Temporal workflow. If we tell the DB the overage has been cleaned up but fail to update Metronome, we end up with a user who can spend more than their account allows.

avervaet:
### Stack
- #29921 feat: Spend-limit expiration scheduler
- **#29754 feat: Admin approval flow for overage requests (this PR)**
- #29960 feat: Add grant-snapshot columns to membership_upgrade_requests
- #29961 feat: Resource, API, and expiration logic for upgrade-request grants
- #29755 feat: Admin history view for resolved upgrade requests
- #29861 Export upgrade-requests history to CSV
- #29863 Align Batch edit spend limit with the Manage request modal
- #29913 Filter overage requests history by decision, and search it by user

avervaet:

- #29921
    - #29963
        - #29754 ◀
            - #29960
                - #29961
                    - #29755
                        - #29861
                            - #29863
                                - #29913
                                    - #29972
                                        - #30351

front/admin/audit_log_schemas/member.spend_limit_updated.json
```diff
@@ -10,6 +10,7 @@
   ],
   "metadata": {
     "kind": "string",
-    "awu_credits": "string"
+    "awu_credits": "string",
+    "expires_at": "string"
```
flvndvd, line 14: Don't you need to bump the version here?

front/components/pages/workspace/UsagePage.tsx
```diff
@@ -339,31 +340,68 @@ export function UsagePage() {
...
     []
   );
-  const handleApproveOnModalSaved = useCallback(() => {
+  // Shared by every way a request can be approved
+  const resolveRequestApproved = useCallback(
+    async (requestId: string): Promise<void> => {
+      const request = upgradeRequests.find((r) => r.sId === requestId);
+      await doResolveUpgradeRequest({
+        requestId,
+        requesterName: request?.requester.name ?? "Member",
```
flvndvd: Why do we need the request here? If we post using the request Id we should not need to specify this, right?

front/components/pages/workspace/UsagePage.tsx
```diff
@@ -339,31 +340,68 @@ export function UsagePage() {
...
+        validateLabel: "Allow unlimited spend",
+        validateVariant: "primary",
+      });
+      if (!confirmed) {
+        return;
+      }
+      setRequestResolving(request.sId, true);
+      try {
+        const member = memberFromUpgradeRequest(request);
+        const body = await doUpdateSpendLimit({
```
flvndvd: IMHO this should be one unique endpoint that focuses around the request and once approve applies the update logic on the spend limit.

front/components/workspace/EditSpendLimitModal.tsx
flvndvd: IIRC there is a better one that uses moment to localise based on user's browser.

flvndvd: Look at the one being used in the `FileExplorer` IIRC we use it there.

front/components/workspace/EditSpendLimitModal.tsx
```diff
@@ -24,18 +32,37 @@ import { useEffect, useRef, useState } from "react";
...
 const MAX_AWU_CREDITS = 2_000_000;
+const ONE_DAY_MS = 24 * 60 * 60 * 1000;
 
 type SpendLimitKind = "default" | "override";
 
 function isSpendLimitKind(value: string): value is SpendLimitKind {
   return value === "default" || value === "override";
 }
 
+type ExpiryMode = SpendLimitExpiryKind;
```
flvndvd: Let's avoid exporting the same type under different name 🙏

front/components/workspace/EditSpendLimitModal.tsx
flvndvd: See here: https://github.com/dust-tt/dust/blob/eee469e03ae9ae727e09ae90b5c82518e6752b74/front/components/file_explorer/FileExplorerItem.tsx#L243

front/components/workspace/EditSpendLimitModal.tsx
```diff
@@ -191,6 +255,77 @@ export function EditSpendLimitModal({
     ? isSaving || isSpendLimitLoading
     : validateDisabled;
 
+  const overrideForm = (
```
flvndvd: Let's extract this to a proper component this one is already very lengthy.

flvndvd: You should also use React hook form here.

avervaet:

PR #29757, opened 2026-07-30
Drop workspace_id from sandbox StatsD tags

fontanierh:
## Description

`sandbox.lifecycle.duration` tripped the high-cardinality custom metric monitor (~11k series). Root cause was tagging every sample with `workspace_id`.

Stop emitting `workspace_id` on sandbox StatsD metrics (region + status/operation only). Per-workspace drill-down stays in APM traces.

## Tests

Updated `sandbox_resource.test.ts` and ran it locally.

## Risk

Low. Dashboards/monitors that grouped these metrics by `workspace_id` will lose that dimension. No such monitors found.

## Deploy Plan

Standard deploy. Old high-cardinality series age out on their own.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/front/lib/api/sandbox/db.ts
+++ b/front/lib/api/sandbox/db.ts
@@ -450,7 +450,7 @@ export async function ensurePodStateHealthOnSleep(
       if (refreshResult.error instanceof SandboxNotFoundError) {
         return new Ok(undefined);
       }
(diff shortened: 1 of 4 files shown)


fabiencelier (approved):
LGTM ✅

PR #29758, opened 2026-07-30
Add agent message consumption item resource

flvndvd:
## Description

This PR adds the resource used to persist and read agent message consumption items introduced by the preceding data
model PR https://github.com/dust-tt/dust/pull/29664. It intentionally contains no materializer or call sites.

Completed items are inserted idempotently in bulk. Tools spanning approval or another resumable interruption may
first be inserted as pending, then completed with their result footprint, direct charge, final gross attribution,
and completion timestamp. Completion cannot rewrite the emitting run or the tool-call output footprint captured when
the row was created.

Only tool items may be `pending`. Deterministic identities make retries no-ops and preserve the first completed facts.
The resource also provides version-scoped message reads and explicit deletion with the owning agent message.

The PR does not calculate billing, pricing, or token attribution.

### DIFF
--- a/front/lib/resources/agent_message_consumption_item_resource.ts
+++ b/front/lib/resources/agent_message_consumption_item_resource.ts
@@ -1,46 +1,45 @@
 import type { Authenticator } from "@app/lib/auth";
-import { AgentMCPActionModel } from "@app/lib/models/agent/actions/mcp";
 import { AgentMessageConsumptionItemModel } from "@app/lib/models/agent/agent_message_consumption_item";
-import {
-  AgentMessageModel,
-  ConversationModel,
-} from "@app/lib/models/agent/conversation";
 import { BaseResource } from "@app/lib/resources/base_resource";
-import {
-  RunModel,
-  RunUsageModel,
-} from "@app/lib/resources/storage/models/runs";
+import { frontSequelize } from "@app/lib/resources/storage";
 import type { ReadonlyAttributesType } from "@app/lib/resources/storage/types";
 import type { ModelStaticWorkspaceAware } from "@app/lib/resources/storage/wrappers/workspace_models";
-import { withTransaction } from "@app/lib/utils/sql_utils";
 import type { ModelId } from "@app/types/shared/model_id";
 import type { Result } from "@app/types/shared/result";
 import { Err } from "@app/types/shared/result";
 import { assertNever } from "@app/types/shared/utils/assert_never";
-import type { Attributes, Transaction } from "sequelize";
-import { Op } from "sequelize";
+import type { Attributes, CreationAttributes, Transaction } from "sequelize";
+import { Op, QueryTypes } from "sequelize";
 
-interface ConsumptionItemEvidenceBase {
+type ConsumptionItemState = "pending" | "completed";
+
+interface ConsumptionItemEvidenceBase<
+  TState extends ConsumptionItemState = ConsumptionItemState,
+> {
   grossAttributedCreditAmountMicro: number;
-  state: "pending" | "completed";
+  state: TState;
 }
 
-export type AgentMessageConsumptionItemRecord =
-  | (ConsumptionItemEvidenceBase & {
+export type AgentMessageConsumptionItemRecord<
+  TState extends ConsumptionItemState = ConsumptionItemState,
+> =
+  | (ConsumptionItemEvidenceBase<TState> & {
       itemType: "system" | "input";
       runUsageModelId: ModelId;
       inputTokensCount: number | null;
     })
-  | (ConsumptionItemEvidenceBase & {
+  | (ConsumptionItemEvidenceBase<TState> & {
       itemType: "output" | "reasoning";
       runUsageModelId: ModelId;
       outputTokensCount: number | null;
     })
-  | (ConsumptionItemEvidenceBase & {
+  | (ConsumptionItemEvidenceBase<TState> & {
       itemType: "tool";
       runUsageModelId: ModelId | null;
       agentMCPActionModelId: ModelId;
+      /** Estimated tokens in the result returned by this tool execution */
       inputTokensCount: number | null;
+      /** Estimated tokens in the model output that emitted the tool name and arguments */
       outputTokensCount: number | null;
       directCreditAmountMicro: number | null;
     });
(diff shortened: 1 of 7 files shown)


front/lib/resources/agent_message_consumption_item_resource.ts
```diff
@@ -0,0 +1,373 @@
...
+  inputTokensCount: number | null;
+  /** Estimated tokens in the model output that emitted the tool name and arguments */
+  outputTokensCount: number | null;
+  directCreditAmountMicro: number | null;
+};
+
+export type PendingToolConsumptionCompletion = ConsumptionItemEvidenceBase & {
+  action: AgentMCPActionResource;
+  /** Estimated tokens in the result returned by this tool execution */
+  inputTokensCount: number | null;
```
aubin-tchoi, line 34: when will this be null?

front/lib/resources/agent_message_consumption_item_resource.ts
```diff
@@ -0,0 +1,373 @@
...
+  directCreditAmountMicro: number | null;
+};
+
+export type PendingToolConsumptionItem = ConsumptionItemEvidenceBase & {
+  action: AgentMCPActionResource;
+  runUsageModelId: ModelId | null;
+  /** Estimated tokens in the model output that emitted the tool name and arguments */
+  outputTokensCount: number | null;
```
aubin-tchoi, line 42: same

front/lib/resources/agent_message_consumption_item_resource.ts
```diff
@@ -0,0 +1,373 @@
...
+          itemType: "tool",
+          attributionVersion,
+          inputTokensCount: null,
+          outputTokensCount: item.outputTokensCount,
+          grossAttributedCreditAmountMicro:
+            item.grossAttributedCreditAmountMicro,
+          directCreditAmountMicro: null,
+          completedAt: null,
+          createdAt: now,
+          updatedAt: now,
```
aubin-tchoi, line 268: why do we supply `createdAt` and `updatedAt`?

front/lib/resources/agent_message_consumption_item_resource.ts
```diff
@@ -0,0 +1,373 @@
...
+          attributionVersion,
+          record,
+          now,
+        })
+      ),
+      {
+        ignoreDuplicates: true,
+        returning: false,
+        transaction,
+        validate: true,
```
aubin-tchoi, line 230: is `validate` false by default for `bulkCreate`?

front/lib/resources/agent_message_consumption_item_resource.ts
```diff
@@ -0,0 +1,373 @@
...
+    const itemKeys = records.map((record) => this.itemKey(record));
+    const uniqueItemKeys = new Set(itemKeys);
+    assert(
+      uniqueItemKeys.size === itemKeys.length,
+      "Consumption items contain duplicate identities"
+    );
+  }
+
+  private static creationAttributes(
+    workspaceModelId: ModelId,
```
aubin-tchoi: OOC why not pass an authenticator here?

front/lib/resources/agent_message_consumption_item_resource.ts
```diff
@@ -0,0 +1,373 @@
...
+    {
+      conversationModelId,
+      agentMessageModelId,
+      attributionVersion,
+      record,
+      now,
```
aubin-tchoi, line 161: why does the caller need to set a value for `now`?

front/lib/resources/agent_message_consumption_item_resource.ts
```diff
@@ -0,0 +1,373 @@
...
+
+export type PendingToolConsumptionItem = ConsumptionItemEvidenceBase & {
+  action: AgentMCPActionResource;
+  runUsageModelId: ModelId | null;
+  /** Estimated tokens in the model output that emitted the tool name and arguments */
+  outputTokensCount: number | null;
+};
+
+export type CompletedAgentMessageConsumptionItem =
+  | (ConsumptionItemEvidenceBase & {
```
aubin-tchoi: nit: could be easier to read if we extracted the two types here (input and output), and potentially useful in the future

front/lib/resources/agent_message_consumption_item_resource.ts
```diff
@@ -0,0 +1,373 @@
...
+import type { ModelStaticWorkspaceAware } from "@app/lib/resources/storage/wrappers/workspace_models";
+import type { ModelId } from "@app/types/shared/model_id";
+import type { Result } from "@app/types/shared/result";
+import { Err } from "@app/types/shared/result";
+import { assertNever } from "@app/types/shared/utils/assert_never";
+import assert from "assert";
+import type { Attributes, CreationAttributes, Transaction } from "sequelize";
+import { Op } from "sequelize";
+
+interface ConsumptionItemEvidenceBase {
```
aubin-tchoi: nit: this one is the only one that's an `interface` and not a `type`

front/lib/resources/agent_message_consumption_item_resource.ts
```diff
@@ -0,0 +1,373 @@
...
+  runUsageModelId: ModelId | null;
+  action: AgentMCPActionResource;
+  /** Estimated tokens in the result returned by this tool execution */
+  inputTokensCount: number | null;
+  /** Estimated tokens in the model output that emitted the tool name and arguments */
+  outputTokensCount: number | null;
+  directCreditAmountMicro: number | null;
+};
+
+export type PendingToolConsumptionCompletion = ConsumptionItemEvidenceBase & {
```
aubin-tchoi, line 31: am I correct in saying that this type is the type of the data we use to update the row when we come back after having completed a tool that was initially pending?

aubin-tchoi (approved):
LGTM

front/lib/resources/agent_message_consumption_item_resource.ts
flvndvd: This can be `null` when tool is still pending and that tool results (which are attributed as model input tokens). We only know them once the tool has run.

front/lib/resources/agent_message_consumption_item_resource.ts
flvndvd: Keep me honest, my reads is that some tools don't take any input parameters (output tokens from the model).

front/lib/resources/agent_message_consumption_item_resource.ts
flvndvd: Good catch let me drop them.

front/lib/resources/agent_message_consumption_item_resource.ts
flvndvd: 🙈

front/lib/resources/agent_message_consumption_item_resource.ts
flvndvd: That is correct!

front/lib/resources/agent_message_consumption_item_resource.ts
aubin-tchoi: but when we are able to provide an object of this type, we are in a situation where the tool must have run no?

front/lib/resources/agent_message_consumption_item_resource.ts
aubin-tchoi: indeed, didn't think of that!

front/lib/resources/agent_message_consumption_item_resource.ts
flvndvd: Actually it looks like tool inputs can be `null` see [here](https://dust4ai.slack.com/archives/C050SM8NSPK/p1785417091163729).

I might refined later on, based on what I see when testing E2E.

front/lib/resources/agent_message_consumption_item_resource.ts
flvndvd: The goal was to tag all the rows from the same run with the same date.

front/lib/resources/agent_message_consumption_item_resource.ts
flvndvd: Yes, let's me add a comment.

PR #29759, opened 2026-07-30
[front] perf: create auto-allowed MCP actions in running status

id13:
## Description

The PR removes one of the two UPDATEs every auto-approved tool execution used to make. Here is the before and after.

Before. An agent-loop tool touched agent_mcp_actions three times:

1. createToolActionsActivity INSERTs the row with the status computed from the tool's permission: ready_allowed_implicitly for anything auto-approved.
2. runToolActivity starts executing and the first thing runToolWithStreaming did was an unconditional updateStatus("running") for every agent-loop action. That is UPDATE 1.
3. When the tool finishes, markAsSucceeded / markAsErrored writes the final status plus executionDurationMs. That is UPDATE 2

That is ~2 updates per insert, which is exactly what you can find in pg_stat_user_tables (I reset the stats this morning).

The insight is that the ready_allowed_implicitly → running transition carries no information. Nothing ever observes the difference: the workflow goes straight from the create activity to the run activity, resume logic (getExistingActionsAndBlobs) only asks "is it final?" and "is it blocked_validation_required?", and both values answer those questions identically. So:

- At creation time we already know the tool will start executing immediately (its status is auto-allowed, and the workflow only pauses a step when some tool in it needs approval). In that case the row is INSERTed with status: "running" directly. This is the same pattern sandbox function actions already used, per the old comment in run_tool.ts.
- The write in runToolWithStreaming becomes conditional: if (isAgentLoopRunContext(runContext) && status !== "running"). For rows born as running the condition is false and UPDATE 1 never happens. It still fires for the cases where the persisted status genuinely differs at execution start: user-approved tools (ready_allowed_explicitly), tools resumed after auth or a user answer, and the carve-out below.
- UPDATE 2 stays. The outcome is only known at the end, and the row must exist during execution (output items FK to it, the MCP progress token is the action id), so insert-at-completion is not an option (I'm sad ☹️)

## Tests

Tested locally 
front-edging

## Risk

Low
Blast radius: Agent. loop

## Plan

- [ ] Deploy front

### DIFF
--- a/front/temporal/agent_loop/lib/create_tool_actions.ts
+++ b/front/temporal/agent_loop/lib/create_tool_actions.ts
@@ -35,6 +35,14 @@ type CreateToolActionsResult = {
   actionBlobs: ActionBlob[];
 };
 
+type PreparedToolAction = {
+  actionConfiguration: MCPToolConfigurationType;
+  rawInputs: Record<string, unknown>;
+  status: "ready_allowed_implicitly" | "blocked_validation_required";
+  stepContent: AgentStepContentResource;
+  stepContext: StepContext;
+};
+
 export async function createToolActionsActivity(
   auth: Authenticator,
   {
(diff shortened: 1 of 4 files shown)


id13:

- #29759 ◀
    - #29777

front/temporal/agent_loop/lib/create_tool_actions.ts
```diff
@@ -62,22 +70,64 @@ export async function createToolActionsActivity(
...
+    const { status } = await getExecutionStatusFromConfig(auth, {
       actionConfiguration,
+      skipToolsValidation: agentMessage.skipToolsValidation,
+      context: {
+        toolInputs: rawInputs,
+      },
+    });
+
+    preparedActions.push({
+      actionConfiguration,
```
aubin-tchoi, line 106: nit: doesn't look like you need the agent config here as it's the same across all actions

aubin-tchoi (approved):
LGTM, nice find!

front/temporal/agent_loop/lib/create_tool_actions.ts
id13: good point

aubin-tchoi (approved):

PR #29761, opened 2026-07-30
Prioritize awake sandboxes in the sandbox reaper, sleep idle ones faster

fontanierh:
## Description

Yesterday's sandbox image rollout spiked EU E2B cluster concurrency: the kill-requester marked the whole old-version fleet (~30k sandboxes, mostly sleeping), the kill sweep saturated every reaper run, and the pause sweep starved, leaving idle running sandboxes stacked up.

Changes:
- Split the kill_requested phase in two: awake sandboxes (running / pending_approval) are destroyed first, sleeping ones are swept last, most recently active first (they are the likeliest to be woken by a returning user). Sleeping kills are best-effort cleanup only: `ensureActive` already destroys and recreates kill-requested sandboxes lazily on access.
- Reorder phases so pausing stale running sandboxes runs before any kill sweep. Later phases only get batches when earlier ones leave room, so sleeping kills effectively run when the reaper is idle.
- Double `MAX_BATCHES_PER_PHASE` to 200. The 100-batch cap (12.8k kills per 5-min run) was the bottleneck when draining rollout backlogs.
- Lower the default sleep threshold from 10 to 5 min to cut the idle tail of each session (still env-overridable).
- Add a partial index on `(lastActivityAt, id)` for kill-requested sleeping sandboxes to back the new sweep (pre-deploy migration).

The workflow phase reorder is behind a Temporal patch (`sandbox-reaper-prioritized-phases`) so in-flight executions replay deterministically.

## Tests

Added resource tests for the status filter and the lastActivityAt-desc pagination, and a reaper test that sleeping kill-requested sandboxes are skipped by the awake phase and destroyed by the low-priority phase.

## Risk

Low. Rollout-day backlog drains slower for sleeping sandboxes by design (they linger as paused snapshots longer, storage cost only). Reaper behavior for awake sandboxes is unchanged apart from ordering. Migration is a concurrent partial index, no locking. Rollback is a revert; the new phase name is only read by new workflow executions.

## Deploy Plan

Pre-deploy migration must run before the code ships (standard). Temporal patch cleanup comments reference follow-up dates.

### DIFF
--- a/front/lib/resources/storage/models/sandbox.ts
+++ b/front/lib/resources/storage/models/sandbox.ts
@@ -92,6 +92,16 @@ SandboxModel.init(
         },
       },
       {
+        // Low-priority reaper sweep: kill-requested sleeping sandboxes are
+        // destroyed most-recently-active first.
+        fields: ["lastActivityAt", "id"],
+        name: "sandboxes_reaper_kill_requested_sleeping_idx",
+        where: {
+          killRequestedAt: { [Op.ne]: null },
+          status: "sleeping",
+        },
+      },
+      {
         fields: ["status", "lastActivityAt", "id"],
         name: "sandboxes_reaper_stale_idx",
         where: { killRequestedAt: { [Op.is]: null } },
--- a/front/lib/resources/sandbox_resource.test.ts
+++ b/front/lib/resources/sandbox_resource.test.ts
@@ -540,9 +540,12 @@ describe("SandboxResource.dangerouslyGetKillRequestedSandboxes", () => {
       agentConfigurationId,
       messagesCreatedAt: [new Date()],
     });
+    const olderActivityAt = new Date(Date.now() - 60 * 60 * 1000);
+    const recentActivityAt = new Date();
     const older = await SandboxFactory.create(authenticator, conversation, {
       status: "sleeping",
       killRequestedAt: new Date(),
+      lastActivityAt: olderActivityAt,
     });
     const recent = await SandboxFactory.create(
       authenticator,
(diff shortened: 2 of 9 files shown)


id13:
Why not completely skip the sleeping ones and leverage the fact we also lookup killrequestat on bash tool execution (i.e at wakeup) ?

we trade colder start, but imo that's ok

fontanierh:
But actually we should only do killrequest_sleeping after doing running. I'm switching them up

front/lib/resources/storage/models/sandbox.ts
```diff
@@ -91,6 +91,16 @@ SandboxModel.init(
...
       },
+      {
+        // Low-priority reaper sweep: kill-requested sleeping sandboxes are
+        // destroyed most-recently-active first.
+        fields: ["lastActivityAt", "id"],
+        name: "sandboxes_reaper_kill_requested_sleeping_idx",
+        where: {
+          killRequestedAt: { [Op.ne]: null },
+          status: "sleeping",
+        },
```
id13, line 102: why is the index ⏬  not enough ?

fontanierh:
>Why not completely skip the sleeping ones and leverage the fact we also lookup killrequestat on bash tool execution (i.e at wakeup) ?

to reclaim the storage

front/lib/resources/storage/models/sandbox.ts
fontanierh: Because then the DB has to filter and sort within all killrequested boxes, which can be pretty large

id13:
> to reclaim the storage

Not sure I get it, these are two different modes no?

1. KillSwitch to rollout a new version -> only awake sandbox actually matter
2. Regular reaper -> reclaim resources

The spikes are only caused by 1. and if we just skip the sleeping ones we loose nothing or I'm missing smth ?

fontanierh:
> Not sure I get it, these are two different modes no?

If you don't touch the kill requested that are sleeping, you continue paying for the storage of that sleeping sandbox until the next time the user comes back in the same conversation (at which point we'll delete).

Most people never come back to their sleeping sandbox, so if we can reclaim the storage when we know for sure we will never use it again, we might as well do it.
Also, there is still a little bit of user-path benefit to pre-destroy if the user does come back to the conversation (ensure active will be a bit faster).

Agree that it's lower prio than putting running sandboxes to sleep though, so I jsut changed the order @id13

fontanierh:
So it's higher prio than a regular non-kill-requested sleeping sandbox, but lower prio than everything else.

front/lib/resources/storage/models/sandbox.ts
id13: ```          killRequestedAt: { [Op.ne]: null },``` ok missed the is vs ne

id13 (approved):

front/lib/resources/storage/models/sandbox.ts
id13: just make sure both predicates are in the queries that will run

PR #29765, opened 2026-07-30
Pricing: prevent using the origin: sidekick with non sidekick agent

fabiencelier:
## Description

origin is passed by the front end and cannot be trusted.
block sidekick origin with non sidekick agents.

## Tests

sidekick still works normally

## Risk

rick to blick legit sidekick queries but I think it is safe here

## Deploy Plan

deploy front


tdraier (approved):
ok as temporary fix, we should better enforce the origin  ( ideally in isUserMessageContextValid )

zmarouf:
LGTM and echoing @tdraier

PR #29767, opened 2026-07-30
Accept protocol v3 Pod function result envelopes

fontanierh:
## Description

Accept protocol v3 result envelopes (`protocolVersion` / optional `delivery` / `outcome` / optional `timingsMs`) on top of the relocated normalizer, alongside the current and legacy runner shapes.

Also accepts front/dsbx-minted error codes such as `invocation_failed`, fails closed on unsupported protocol versions via an append-only supported set, and logs rejected envelopes. `timingsMs` is accepted on the wire but not consumed yet.

Stacked on the pure extraction PR.

## Tests

- Unit coverage for legacy, current, and v3 envelopes (including dsbx-minted `invocation_failed`).
- Route tests for v3 success and error callbacks.

## Risk

Low. Additive acceptance of a shape nothing emits in production yet. Rollback is a revert. When a later protocol version is introduced, keep older versions in the supported set until baked images rotate off them.

## Deploy Plan

Standard deploy. Merge after the extraction PR.

### DIFF
--- a/front/lib/api/sandbox_functions/result_envelope.ts
+++ b/front/lib/api/sandbox_functions/result_envelope.ts
@@ -1,6 +1,19 @@
+import logger from "@app/logger/logger";
 import type { SandboxFunctionCallError } from "@app/types/api/sandbox_functions";
 import { SANDBOX_FUNCTION_RUNNER_ERROR_CODES } from "@app/types/api/sandbox_functions";
 import { z } from "zod";
+
+// Current wire version dsbx emits. Parsing accepts the supported set below so a future bump
+// does not instantly fail long-lived baked pod images still on an older version.
+export const SANDBOX_FUNCTION_RESULT_PROTOCOL_VERSION = 3;
+
+export const SUPPORTED_SANDBOX_FUNCTION_RESULT_PROTOCOL_VERSIONS = [
+  SANDBOX_FUNCTION_RESULT_PROTOCOL_VERSION,
+] as const;
+
+export type NormalizedSandboxFunctionOutcome =
+  | { ok: true; output: unknown }
+  | { ok: false; error: SandboxFunctionCallError };
 
 type JsonValue = null | boolean | number | string | object;
 const DefinedJsonValueSchema = z.custom<JsonValue>((v) => v !== undefined);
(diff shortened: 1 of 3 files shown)


front/lib/api/sandbox_functions/result_envelope.ts
```diff
@@ -114,3 +147,51 @@ export function normalizeSandboxFunctionResult(
...
+ * Normalize a Pod function result payload from either the HTTP callback body or a
+ * worker-owned stdout envelope into one classified outcome.
+ */
+export function normalizeSandboxFunctionResult(
+  result: unknown
+): NormalizedSandboxFunctionOutcome {
+  const versionProbe = ProtocolVersionProbeSchema.safeParse(result);
+  if (versionProbe.success) {
+    const { protocolVersion } = versionProbe.data;
+    if (!Number.isInteger(protocolVersion)) {
```
aubin-tchoi, line 179: `ProtocolVersionProbeSchema` seems to already do this check

front/lib/api/sandbox_functions/result_envelope.ts
```diff
@@ -1,7 +1,20 @@
...
 import { SANDBOX_FUNCTION_RUNNER_ERROR_CODES } from "@app/types/api/sandbox_functions";
 import { z } from "zod";
 
+// Current wire version dsbx emits. Parsing accepts the supported set below so a future bump
+// does not instantly fail long-lived baked pod images still on an older version.
+export const SANDBOX_FUNCTION_RESULT_PROTOCOL_VERSION = 3;
+
+export const SUPPORTED_SANDBOX_FUNCTION_RESULT_PROTOCOL_VERSIONS = [
+  SANDBOX_FUNCTION_RESULT_PROTOCOL_VERSION,
+] as const;
```
aubin-tchoi, line 13: you'll also need an array of accepted schemas no? otherwise there's nothing that enforces that a valid outdated version is actually supported by the new schema

front/lib/api/sandbox_functions/result_envelope.ts
```diff
@@ -114,3 +147,51 @@ export function normalizeSandboxFunctionResult(
...
+  const versionProbe = ProtocolVersionProbeSchema.safeParse(result);
+  if (versionProbe.success) {
+    const { protocolVersion } = versionProbe.data;
+    if (!Number.isInteger(protocolVersion)) {
+      return invalidResultEnvelope("non_integer_protocol_version", {
+        protocolVersion,
+      });
+    }
+    if (!isSupportedProtocolVersion(protocolVersion)) {
+      return {
```
aubin-tchoi, line 189: no `invalidResultEnvelope` to have a log here?

aubin-tchoi (approved):
LGTM

front/lib/api/sandbox_functions/result_envelope.ts
fontanierh: nope z.number() doesn't enforce integer

front/lib/api/sandbox_functions/result_envelope.ts
fontanierh: Today the supported set is [3] and there's one schema, so nothing is wrong

once we support more than one then yes we'll need this mechanism (not doign for now)

PR #29768, opened 2026-07-30
Make Pod function terminal transitions compare-and-set idempotent

fontanierh:
## Description

Make `succeed` / `fail` / `markCreatedAsErrored` compare-and-set on status `created`, so only one deliverer can take an invocation terminal. If the terminal GCS write fails after a won claim, release the claim so a later fail path can still record the outcome.

Also covers a duplicate HTTP callback: the second POST is a no-op.

Stacked on the shared result-envelope PR.

## Tests

- Resource tests for lost claims and GCS write failure releasing the claim.
- Route test that a second successful callback does not overwrite the first result.

## Risk

Low for callers: `succeed`/`fail` now return `boolean` and existing callers ignore it. Behavioural notes: the row can briefly look terminal before its GCS blob exists, and a failed terminal write resurrects the row to `created`. Rollback is a revert.

## Deploy Plan

Standard deploy. Merge after the envelope PR.

### DIFF
--- a/front/lib/resources/sandbox_function_invocation_resource.ts
+++ b/front/lib/resources/sandbox_function_invocation_resource.ts
@@ -286,34 +286,28 @@ export class SandboxFunctionInvocationResource extends BaseResource<SandboxFunct
     return this.data.error;
   }
 
-  // Compare-and-set: only the caller that flips `created` owns the outcome. Guards the
-  // double-delivery window (worker stdout + late HTTP callback) without a lock.
-  // BaseResource.update() only keys on `id`, so the predicate goes on the model here
-  // (cf. ConversationForkResource.markFileCopied).
-  private async claimTerminalStatus(
-    status: Extract<SandboxFunctionInvocationStatus, "succeeded" | "errored">
-  ): Promise<boolean> {
+  // WHERE-guarded compare-and-swap on status. Same pattern as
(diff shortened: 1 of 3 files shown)


front/lib/resources/sandbox_function_invocation_resource.ts
```diff
@@ -258,11 +258,82 @@ export class SandboxFunctionInvocationResource extends BaseResource<SandboxFunct
...
+  // SandboxFunctionMCPActionResource.updateStatusFromExpected: BaseResource.update()
+  // only keys on `id`, so the expected-status predicate goes on the model here.
+  private async casStatus({
+    from,
+    to,
+  }: {
+    from: SandboxFunctionInvocationStatus;
+    to: SandboxFunctionInvocationStatus;
+  }): Promise<boolean> {
+    const [affectedCount, rows] = await this.model.update(
```
aubin-tchoi, line 299: nit: could be rewritten with the base resource's update (would be marginally shorter)

aubin-tchoi (approved):
LGTM

front/lib/resources/sandbox_function_invocation_resource.ts
fontanierh: Would need to add a where to the resource's update, not sure it's worth it

PR #29776, opened 2026-07-30
Add terminal image pruning to conversation window state

flvndvd:
## Description

Some providers, such as Anthropic, limit the number of input images. This adds an optional image cap to `CheckpointedConversationWindowState`, pruning the oldest images during the terminal `fit()` pass while preserving token accounting and valid message contents.

It also makes `fit()` explicitly terminal to prevent later appends from using stale pruning state.

### DIFF
--- a/front/lib/api/assistant/conversation_rendering/checkpointed_window_state.ts
+++ b/front/lib/api/assistant/conversation_rendering/checkpointed_window_state.ts
@@ -80,6 +80,7 @@ function pruneImagesBeyondLimit(
       imageMessage = message;
       content = message.content;
       break;
+
     case "function":
       if (!Array.isArray(message.content)) {
         return {
@@ -92,6 +93,7 @@ function pruneImagesBeyondLimit(
       imageMessage = message;
       content = message.content;
       break;
+
     case "assistant":
     case "compaction":
       return {
@@ -100,6 +102,7 @@ function pruneImagesBeyondLimit(
         retainedImageCount: 0,
         tokenSavings: 0,
       };
+
     default:
       return assertNever(message);
   }
@@ -183,6 +186,7 @@ export class CheckpointedConversationWindowState {
   private retainedTokens = 0;
   private totalTokensBefore = 0;
   private prunedTokens = 0;
+  private prunedImageCount = 0;
   private fitted = false;
 
   private pendingToolResults: PendingToolResult[] = [];
@@ -324,6 +328,7 @@ export class CheckpointedConversationWindowState {
           node.message = result.message;
           this.retainedTokens -= result.tokenSavings;
           this.prunedTokens += result.tokenSavings;
+          this.prunedImageCount += result.prunedImageCount;
         }
         imagesToKeep -= result.retainedImageCount;
       }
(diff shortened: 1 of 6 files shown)


front/lib/api/assistant/conversation_rendering/index.ts
```diff
@@ -10,7 +10,10 @@ import type {
   InteractionWithTokens,
```
fontanierh, line 10: Looks like maxImages is never passed here ?

front/lib/api/assistant/conversation_rendering/checkpointed_window_state.ts
```diff
@@ -45,8 +55,102 @@ type WindowInteraction = {
...
+  prunedImageCount: number;
+  retainedImageCount: number;
+  tokenSavings: number;
+} {
+  let imageMessage: ImageBearingMessage;
+  let content: Content[];
+
+  switch (message.role) {
+    case "user":
+    case "content_fragment":
```
fontanierh, line 79: I don't think we ever want to prune user-uploaded images

front/lib/api/assistant/conversation_rendering/checkpointed_window_state.ts
```diff
@@ -45,8 +55,102 @@ type WindowInteraction = {
...
+    return {
+      message,
+      prunedImageCount: 0,
+      retainedImageCount,
+      tokenSavings: 0,
+    };
+  }
+
+  const retainedContent = retainedContentReversed.reverse();
+  const addedPlaceholder = retainedContent.length === 0;
```
fontanierh, line 135: mmmh what if you have image and text in one ? You would silently drop the image without adding placeholder no ?

front/lib/api/assistant/conversation_rendering/checkpointed_window_state.ts
```diff
@@ -12,8 +13,17 @@ import type {
...
 import logger from "@app/logger/logger";
+import type { Content } from "@app/types/assistant/generation";
+import { isImageContent } from "@app/types/assistant/generation";
 import type { Result } from "@app/types/shared/result";
 import { Ok } from "@app/types/shared/result";
+import { assertNever } from "@app/types/shared/utils/assert_never";
+
+const PRUNED_IMAGE_PLACEHOLDER =
+  "<dust_system>" +
+  "This image is no longer available (pruned to respect the model input image limit)." +
```
fontanierh, line 24: I think we want to include the file path here

front/lib/api/assistant/conversation_rendering/checkpointed_window_state.ts
fontanierh: And we want to warn if there are > 20 user-uploaded images

front/lib/api/assistant/conversation_rendering/checkpointed_window_state.ts
```diff
@@ -153,6 +270,8 @@ export class CheckpointedConversationWindowState {
       });
     }
 
+    this.applyImageLimit();
```
fontanierh, line 273: I don't think this is really optimal, because you might be pruning tool outputs in the previous step due to image tokens that would have been removed anyways....

PR #29777, opened 2026-07-30
[front] perf: skip identical runKey rewrites on runs

id13:
## Description

- Runs is one of our biggest table, and it is updated x2 the rate of inserts, which for an "almost-append-only" table is quite pathological. 
- This PR transform the bespoke idempotent `setRunKeyForDustRunIds` from "we will alway update the same row with runKey" to "we will update the same row with runKey only once". Which is more idempotent imo.

## Test 

Existing

## Risk

Low - inclined to say None, we only update if runKey changes instead of always updating, so unless we have some logic based on updatedAt ... could not find any

## Plan

Deploy front

### DIFF
--- a/front/lib/resources/run_resource.test.ts
+++ b/front/lib/resources/run_resource.test.ts
@@ -95,3 +95,41 @@ describe("RunResource.setUsageTypeForRuns", () => {
     expect(usages[0]?.usageType).toBe("free");
   });
 });
+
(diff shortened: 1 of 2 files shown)


id13:

- #29759
    - #29777 ◀

id13:
PMRR

fabiencelier (commented):
LGTM ✅ 
🧹 🧹

PR #29779, opened 2026-07-30
[AGENT-LOOP] Log the provider stop reason on empty model turns

pmilliotte:
## Description

Threads each provider's raw stop/finish reason onto the `success` event and logs it — with the content-block shape — on the empty-turn paths in `run_model`, so silently empty model turns can be diagnosed instead of guessed at.

## Tests

Added a converter test for the thinking-only turn shape; ran the model constructor, LLM and agent loop suites (307 tests).

## Risk

Low.

## Deploy Plan

- Deploy `main` on `front`


aubin-tchoi (approved):
LGTM

PR #29780, opened 2026-07-30
fix(marketing): exclude unreleased and legacy models from public credits page

avervaet:
## Description

This follow-up tightens the model list exposed by the public `/home/credits` page through `GET /api/marketing/model-credits`.

Until now, `buildPublicModelCredits()` iterated over `SUPPORTED_MODEL_CONFIGS` and could publish models that were still gated behind an on-demand/Dust-only feature flag, as well as legacy or deprecated models. This exposed unreleased entries such as gated xAI, DeepSeek, OpenAI o1, and Claude Opus models on an unauthenticated marketing page.

This PR adds `isModelReleased()` alongside the existing model-availability helpers. It considers a model released only when it has no `availableIfOneOf.featureFlag`, so a feature flag remains authoritative even when `plansWithAdvancedModels` is also configured. `buildPublicModelCredits()` now excludes both unreleased and `isLegacy` models before applying the existing pricing-entry filter.

The change only affects the public credits response. Existing workspace model availability and the underlying pricing configuration remain unchanged. This is a follow-up to [#29726](https://github.com/dust-tt/dust/pull/29726), which introduced the credits page.

## Tests

- Added unit coverage in `front/lib/assistant.test.ts` for released models and feature-flagged models.
- Added API regression coverage in `front-api/routes/marketing/model_credits.test.ts` for gated and legacy model IDs.
- Ran `npx tsgo --noEmit` on `front/lib/assistant.ts` and `front/lib/api/marketing/model_credits.ts`; no new errors were reported.

## Risk

Low. The change narrows only the unauthenticated marketing response and does not modify model availability for workspaces, pricing values, or stored data. No migration or backfill is involved, and the implementation is reversible with a code revert. The main functional risk is an incorrectly configured release flag causing a model to be omitted from the public page; the API regression tests cover representative gated and legacy models.

## Deploy Plan

- No migration or feature-flag activation is required.
- Deploy marketting

### DIFF
--- a/front/lib/assistant.ts
+++ b/front/lib/assistant.ts
@@ -18,11 +18,9 @@ export function getAdvancedModels(): ModelConfigurationType[] {
   return SUPPORTED_MODEL_CONFIGS.filter(isAdvancedModel);
 }
 
-// True unless the model is only reachable via an on-demand/dust-only feature flag (not GA).
+// False if the model requires an on-demand/dust-only feature flag (not GA).
 export function isModelReleased(m: ModelConfigurationType): boolean {
-  return (
-    !m.availableIfOneOf || m.availableIfOneOf.plansWithAdvancedModels === true
-  );
+  return !m.availableIfOneOf?.featureFlag;
 }
 
 // Returns true if the model is available to the workspace for build.
--- a/front/lib/assistant.ts
+++ b/front/lib/assistant.ts
@@ -18,6 +18,13 @@ export function getAdvancedModels(): ModelConfigurationType[] {
   return SUPPORTED_MODEL_CONFIGS.filter(isAdvancedModel);
 }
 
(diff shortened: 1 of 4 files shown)


ClementAupiais (approved):
LGTM

front/lib/assistant.ts
```diff
@@ -18,6 +18,13 @@ export function getAdvancedModels(): ModelConfigurationType[] {
   return SUPPORTED_MODEL_CONFIGS.filter(isAdvancedModel);
 }
 
+// True unless the model is only reachable via an on-demand/dust-only feature flag (not GA).
+export function isModelReleased(m: ModelConfigurationType): boolean {
+  return (
+    !m.availableIfOneOf || m.availableIfOneOf.plansWithAdvancedModels === true
```
flvndvd: Can we still check for no feature flags at all?

flvndvd (approved):

PR #29782, opened 2026-07-30
[Skills] Remove default skills second permission check

xebeohp:
## Description
`setDefaultSkills` no longer re-filters skills to the global space, instead it trusts its input, since `SkillResource.fetchByIds (baseFetch)` already enforces space access on both write and read. This fixes pod-scoped skills being silently rejected as defaults, matching the picker/slash command. Drops the now-unused auth param.

Note: temporarily more permissive (users with access to a skill restricted to pod A can use it anywhere) until a follow-up PR reintroduces the correct global-or-this-pod restriction.

## Tests
Local tests. Edited skill_resource and project_metadata tests to reflect the edits.

## Risk
Low risk.

## Deploy Plan
Deploy front.

### DIFF
--- a/front-api/routes/w/[wId]/spaces/[spaceId]/project_metadata.ts
+++ b/front-api/routes/w/[wId]/spaces/[spaceId]/project_metadata.ts
@@ -163,7 +163,7 @@ app.patch(
           : null,
       });
       if (resolvedDefaultSkills) {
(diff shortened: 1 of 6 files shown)


xebeohp:

- #29782 ◀

aubin-tchoi (approved):
LGTM

PR #29774, opened 2026-07-30
[front] enh: add link to skill in `enable_skill` action details

aubin-tchoi:
## Description

There is currently no direct way to go inspect a skill that was just enabled from the conversation.

This PR adds a button that redirects to the skill from the  `enable_skill` MCP action details.
Also adds the description.

image

## Tests

- Tested locally.

## Risk

- Low.

## Deploy Plan

- Deploy front.

PR #29783, opened 2026-07-30
[front] enh: show skill details in `enable_skill` action details

aubin-tchoi:
## Description

Follow-up on #29774 that added a link to the Skill info page to the MCP action details for `enable_skill`.

This PR shows the tools, the skills, spaces, etc... in the MCP action details for `enable_skill` so that the user knows a bit more about the skill without having to navigate smwhere else.

Design aligned with the skill info page (shared the inner component).

Also changed the link to point to Skill Builder instead: opening the Skill info page won't be very useful if we already are looking at the exact same information.

image

## Tests

- Tested locally.

## Risk

- Low.

## Deploy Plan

- Deploy front.


fabiencelier (approved):
LGTM ✅ 
not tested

PR #29784, opened 2026-07-30
Extract Pod function result normalization from the callback route

fontanierh:
## Description

Move the existing runner and legacy result parsers out of the callback route into `front/lib/api/sandbox_functions/result_envelope.ts`. Behavior is unchanged: same schemas, same messages, same succeed/fail outcomes.

## Tests

Existing callback route tests cover the moved parsers.

## Risk

Low. Pure relocation. Rollback is a revert.

## Deploy Plan

Standard deploy.


aubin-tchoi (commented):
LGTM

PR #29787, opened 2026-07-30
Opt: bulk insert agent_step_content for the same step.

Fraggle:
## Description

`AgentStepContentResource.createNewVersions` replaces the per-item loop that called `createNewVersion` (and its per-call transaction + individual Redis write) with a single version-lookup `findAll`, a single `bulkCreate`, and one Redis pipeline. Callers in `runModel` and `storeLlmResult` are migrated to the new method. `createNewVersion` is kept as a thin wrapper.

Also upgrades `warmAgentStepContentCacheMany` to use `MULTI`/`EXEC` per cache key instead of `Promise.all` over individual `hSet` calls, and bumps the cache TTL from 15 min to 2 hours.

## Tests

Added unit tests for `createNewVersions`: bulk insert with correct version 0, version bumping on re-insert, Redis warming verification, and empty-input guard. Tested locally.

## Risk

Low. No schema change, no migration. The new path drops the per-item transaction in favor of an unguarded `bulkCreate`; concurrent retries on the same `(agentMessageId, step, index)` could produce duplicate versions, but that was already possible with the old non-serializable transactions. `createNewVersion` (single-item path) is unchanged in behavior.

## Deploy Plan

Deploy `front`.

### DIFF
--- a/front/lib/api/llm/batch_llm.ts
+++ b/front/lib/api/llm/batch_llm.ts
@@ -253,21 +253,25 @@ export async function storeLlmResult(
     });
 
   // Create step content entries from LLM events.
-  let index = 0;
-  for (const event of events) {
+  const stepContentBlobs = events.flatMap((event) => {
     const stepContent = eventToStoredStepContent(event);
-    if (stepContent) {
-      await AgentStepContentResource.createNewVersion({
+    if (!stepContent) {
+      return [];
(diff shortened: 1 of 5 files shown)


Fraggle:

- #29787 ◀

id13 (approved):

PR #29792, opened 2026-07-30
Remove the pod_default_agent and pod_default_skills feature flags

davidebbo:
Both flags are enabled for every workspace, so every gated branch was already taken everywhere. This drops them so the pod default agent and pod default skills behave as always-on.

### Changes

- `front/types/shared/feature_flags.ts` + `sdks/js/src/types.ts` — dropped both flag entries.
- `front-api/routes/w/[wId]/spaces/[spaceId]/project_metadata.ts` — removed the `getFeatureFlags` call and the four gated branches (agent validation, skill resolution, `makeNew` `defaultAgentId`, `updateDefaultAgentId`).
- `front/types/user.ts` — `resolveDefaultAgentId` loses its `hasPodDefaultAgentFeature` parameter. The old `hasPod || hasWorkspace` gate was always true, so the function already reduced to `podDefaultAgentId ?? workspaceDefault`. Updated all call sites.
- `front/lib/api/projects/default_agent.ts` — removed the early `@dust` return.
- `front/lib/project_task/start_agent.ts` — pod metadata is now unconditionally fetched (it already was in practice).
- `front/lib/api/actions/servers/pod_manager/tools/index.ts` — removed the "not enabled for this workspace" error from `set_default_agent`.
- `front/lib/swr/pods.ts` — `usePodDefaultSkills` no longer consults flags; its `return undefined` branch was unreachable. `InputBarContainer` keeps its `undefined` guard for the callers that don't pass the prop at all.
- `front/components/pod/settings/PodSettingsTab.tsx` — unwrapped the "Default agent" and "Default Skills" sections, which were already rendering everywhere.

### Behavior

No user-visible change. `workspace_default_agent` is untouched and its precedence is unchanged — a pod-level default still wins, and the workspace default still only applies when the pod has none. That held before because the pod flag was on; it now holds unconditionally.

The deleted test, `ignores default skills when the feature flag is off`, only reached that path because the test factory withheld a flag that is granted everywhere in production, so it covered unreachable code.

### Testing

- `npx tsgo --noEmit` clean for `front` and `front-api`
- biome clean on all changed files
- `project_metadata` suites pass (15 tests across the private and v1 routes)

### DIFF
--- a/front-api/routes/w/[wId]/spaces/[spaceId]/project_metadata.test.ts
+++ b/front-api/routes/w/[wId]/spaces/[spaceId]/project_metadata.test.ts
@@ -18,7 +18,6 @@ vi.mock("@app/temporal/project_task/client", () => ({
 }));
 
 import { ProjectMetadataResource } from "@app/lib/resources/project_metadata_resource";
-import { FeatureFlagFactory } from "@app/tests/utils/FeatureFlagFactory";
 import { createPrivateApiMockRequest } from "@app/tests/utils/generic_private_api_tests";
 import { SkillFactory } from "@app/tests/utils/SkillFactory";
 import { SpaceFactory } from "@app/tests/utils/SpaceFactory";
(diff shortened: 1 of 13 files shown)


xebeohp (approved):
LGTM, thank you!

PR #29734, opened 2026-07-29
[front] Repoint credit reads to cost.billable_awu

tdraier:
## Description

⚠️ **Draft — do NOT merge until the `cost.billable_awu` backfill from #29733 has run in prod.**
Merging before the backfill makes these reads sum a half-populated field and under-count
consumption (and under-enforce the spend cap).

Stacked on #29733 (which adds and populates `cost.billable_awu`). This PR repoints **every
billing-intent credit surface** from `sum(cost.full_awu)` + a `status ∈ AGENT_MESSAGE_STATUSES_TO_TRACK`
filter to `sum(cost.billable_awu)` (dropping the status filter), so "consumed / billed credits"
everywhere equals what Metronome bills per execution — **including the non-error work of failed
multi-execution messages**, which the status filter previously dropped whole (the `l7G4uvGoCV`
gap from #29733).

Since `billable_awu == full_awu` for non-failed messages, **non-failed totals are unchanged**;
only failed-terminal messages now contribute their billed portion (the non-error executions' work,
or 0 when the only/last execution errored).

## Reads switched

- **Spend cap + members "Consumed (ES)" + poke** — `fetchConsumedAwuCreditsByUserId` (and the
  spend-cap seed `getEsConsumedAwuCreditsForUser`). Correctness-critical.
- **Observability credit dashboards** — `buildCreditsScopeQuery` (shared by credit_usage
  timeseries/breakdown/tables, user_credits, agent_credits) drops its status filter; `overview`
  agent cost stats (`> 0` gate + sum/avg/median) switch to `billable_awu`.
- **Exports** — `users_export` / `agents_export` per-entity credit totals; `messages_export`
  per-message `credits` column.

Left on `full_awu` intentionally: the per-message cost **breakdown** in `credit_cost.ts`
(`totalAwu = llmAwu + toolAwu` must hold — it's a gross component breakdown, not a billed figure).

## Behavior note (message counts)

Dropping the shared status filter in `buildCreditsScopeQuery` also makes the credit dashboards'
message **counts** (`doc_count`) include failed messages (previously excluded). Credit *values*
are unaffected (failed docs carry only their billed portion). Export and overview counts are
**unchanged** — their status filter was local to the credit sub-agg, or they have no count metric.
If you want the dashboard counts to stay failed-excluded, say so and I'll wrap the count in a
status-filtered sub-agg; I left it inclusive since "what happened" is the more natural count.

## Tests

- `npx tsgo --noEmit` clean; `biome check` clean.
[...]

PR #29793, opened 2026-07-30
[front] Fix messages_export credits test for billable_awu repoint

tdraier:
## Problem

`main` is red on the front test suite. The real failing test is:

`fetchMessageExportRows > correctly reports credits column` — `expected +0 to be 7`

The CI "Tests Report" mislabeled it as `node_modules/thread-stream/test/ts.test.ts` — the mikepenz junit annotation guesses the source file from the testcase classname when the junit XML has no `file=` attribute, and `...ts` matched `thread-stream/test/ts.test.ts`. The actual failing test is ours.

## Cause

#29734 repointed credit reads to `cost.billable_awu`:

```ts
credits: Math.round(doc.cost?.billable_awu ?? 0),
```

The test still seeded only `cost.full_awu`, so `credits` resolved to `0` and the `expected 7` assertion failed.

## Fix

Seed `billable_awu: 7` (with a distinct `full_awu: 10`) so the test asserts that `credits` tracks `billable_awu`, not `full_awu`.

### DIFF
--- a/front/lib/api/analytics/messages_export.test.ts
+++ b/front/lib/api/analytics/messages_export.test.ts
@@ -62,7 +62,9 @@ describe("fetchMessageExportRows", () => {
     });
 
     mockMessageHits([
-      messageDoc({ cost: { full_awu: 7, llm_awu: 4, tool_awu: 3 } }),
+      messageDoc({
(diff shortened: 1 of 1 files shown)


fabiencelier (commented):
LGTM ✅

PR #29795, opened 2026-07-30
[front] Another Skill availability copy update

ykmsd:
## Description
This PR is to improve copy and UX of skill availabilities.

- The availability names are Editors only, All members, Members and agents
- Updated the table view and availability filters (having tab "Members and agents" doesn't really make sense)

We found a bug where we treat every open space and pod as restricted and show an warning, I will create a separate PR to fix this.

Manage skill view now has availability filter:
CleanShot 2026-07-31 at 12 00 20@2x

Here is all the copy:
CleanShot 2026-07-31 at 12 04 45@2x
CleanShot 2026-07-31 at 12 04 52@2x
CleanShot 2026-07-31 at 12 04 59@2x
CleanShot 2026-07-31 at 12 05 14@2x
CleanShot 2026-07-31 at 12 05 21@2x
CleanShot 2026-07-31 at 12 05 28@2x
CleanShot 2026-07-31 at 12 06 28@2x

### DIFF
--- a/front/components/pages/builder/skills/ManageSkillsPage.tsx
+++ b/front/components/pages/builder/skills/ManageSkillsPage.tsx
@@ -434,7 +434,7 @@ export function ManageSkillsPage() {
                     value={tab.id}
                     label={
                       tab.id === "default" && hasSkillPublicationGovernance
-                        ? "Auto-discoverable"
+                        ? "Members and agents"
                         : tab.label
                     }
                     onClick={() => setSelectedTab(tab.id)}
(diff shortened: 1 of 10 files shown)


front/components/pages/builder/skills/ManageSkillsPage.tsx
```diff
@@ -158,64 +164,60 @@ export function ManageSkillsPage() {
     disabled: activeTab !== "active",
   });
 
-  const skillsByTab = useMemo(() => {
```
ykmsd, line 161: decoupled this

front/components/pages/builder/skills/utils.ts
```diff
@@ -0,0 +1,127 @@
+import { SKILL_AVAILABILITY_DISPLAY } from "@app/components/skills/SkillsTable";
```
ykmsd, line 1: move some logic to a separate file

PR #29796, opened 2026-07-30
[MCP] resume the sub-agent caller on every approval surface

pmilliotte:
## Description

Waking a parked orchestrator after its sub-agent's tool was approved was opt-in via a flag only the web app sent, so approvals from Slack, Teams, email or the public API resumed the sub-agent but left the caller stuck in `blocked_child_action_input_required` forever; the wake-up is now unconditional, also covers user questions and the email path, and is best-effort so a committed approval is never reported as failed.

## Tests

Added a regression test in `validate_actions.test.ts` (verified it fails without the fix) plus updated `resume_ancestor_conversations.test.ts`.

## Risk

Non-web approvals now trigger one extra agent-loop relaunch per ancestor, which adds latency to the validate-action request.

Monitor "Failed to retry blocked actions on parent conversation" in DD after deploy

## Deploy Plan

- Deploy `main` on `front`

### DIFF
--- a/front-api/lib/api/assistant/conversation/resolve_authentication.ts
+++ b/front-api/lib/api/assistant/conversation/resolve_authentication.ts
@@ -38,15 +38,13 @@ export function makeResolveAuthenticationApp(
       });
     }
 
-    const { actionId, outcome, resumeAncestorConversations } =
-      ctx.req.valid("json");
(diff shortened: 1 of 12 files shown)


fontanierh (approved):
LGTM

PR #29797, opened 2026-07-30
[front] Refresh activation skill prompt and pod Frame template

frankaloia:
## Summary

Refreshes the **activation skill** (`activation.ts`) and the **pinned pod Frame template**. No other changes — poke tooling, models, and resources are intentionally out of scope for this PR.

The skill prompt is restructured around a clearer two-level goal model:
- **Training Plan** — the persistent, standing playbook injected via `AGENTS.md`: who the user is + the prescriptive path to activate them. The north star for every nudge.
- **Session Goal** — the one concrete step a given nudge drives, ephemeral to that run.

The pod Frame template is updated to match.

## Testing

Screenshot 2026-07-30 at 1 30 26 PM

- `npx tsgo --noEmit` — 0 errors
- `biome check` — clean on both files

## Risk

Low. Both files are self-contained: `activation.ts` is a code-defined skill definition (prompt string + config) and the Frame template is a static asset. Neither introduces new runtime dependencies. The change is behavioral (prompt/template content) and gated behind the existing `activation_skill` feature flag.

## Deploy
* Deploy front

### DIFF
--- a/front/lib/resources/skill/code_defined/global/activation.ts
+++ b/front/lib/resources/skill/code_defined/global/activation.ts
@@ -13,242 +13,308 @@ import { safeParseJSON } from "@app/types/shared/utils/json_utils";
 
 const ACTIVATION_BEHAVIOR = `
 # Overview
+You are a Dust trainer for dormant / low-fluency users. In each conversation, you move the user one concrete step toward getting real
+work done in Dust inside their Pod.
 
-The goal is to activate dormant users by creating a curated, purpose-built Dust Pod. Each pod is a living system that will help a user get more value from Dust.
-You will recommend the next best action for the user to improve the pod and get productivity gains from using Dust. Present the user with a recommendation and then guide them into making it a into recurring use case.
+## Artifacts (stable vocabulary)
+- Overall Goal — synonym for the Pod's \`AGENTS.md\`. The whole file is the durable top-level goal for activating THIS user; Session Goals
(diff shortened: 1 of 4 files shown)


id13 (approved):
LGTM, one nit about $$$

front/lib/resources/skill/code_defined/global/activation.ts
id13, line 483: Not related to the pr itself but I'm discovering this part anyway.

not 100% sure but I thiiink we should put the context at the end to maximize cache hits, would like to have @aubin-tchoi or @flvndvd confirmation.

aubin-tchoi: the order won't change the behaviour w.r.t caching here, there are no cases where you can get a cache hit on only the static part and not the context (that would require having a cache breakpoint in-between, and having cases where you can get a cache hit on only the beginning, which won't happen here as this is a global skill and therefore the instructions are exposed at some point in the conversation history, not in the system prompt)

aubin-tchoi: that said we need to make sure context cannot change mid-conversation, as this would burst the cache up to the cache breakpoint on the very first user message (the one for skills), loosing the cache on the entire conversation history minus system prompt, tools and this skills message

flvndvd: 💯 ☝️

PR #29799, opened 2026-07-30
Restore the knip CI gate for front

davidebbo:
## Why

`knip` has not run anywhere since **2026-02-13**: #21412 ("move front to biome") deleted `build-and-lint-front.yml`, which was the only caller of `.github/actions/knip`. The composite action and both configs are still in the repo, just unreferenced. Five months of drift accumulated behind the dead gate:

| front | count |
|---|---|
| unused files | 16 |
| unused dependencies | 11 |
| unused devDependencies | 2 |
| unlisted dependencies (relying on hoisting) | 14 |

This restores the gate and clears all of it. It is phase 1 of two: the `exports` rule stays `off` here and gets its own follow-up (~2400 findings in front, which need batching).

## What

**CI** — knip comes back as its own job in `build-and-test-front.yml`, with the same `build-sdks` / `build-sparkle` setup the old `lint` job gave it.

**Config, for findings that were never real:**

- `start_worker.ts` → `entry`. It is bundled by `esbuild.worker.ts`, which is in `ignoreFiles`, so the worker entry point looked unreachable.
- `public/sw.js` → `ignoreFiles`. Registered by the browser, never imported.
- `config/*.ts`, `temporal/*/client.ts`, `tests/utils/**` → `entry`. These are consumed by `front-api` through the `@app/*` tsconfig alias. Knip resolves those imports (no unresolved-import findings) but does not credit them across the workspace boundary the way it does front-spa's `@dust-tt/front/*` package-name imports. 10 of the 16 "unused" files were front-api's.
- `tailwindcss`, `umzug`, `vitest-environment-node` → `ignoreDependencies`. `tailwindcss` is referenced from `styles/global.css` (`@reference "tailwindcss"`), `umzug` by `scripts/db/migration-runner.ts` which front's `migration:*` scripts invoke, and `vitest-environment-node` is knip reading the `@vitest-environment node` docblock in `detect_skills.test.ts` as a package name.

**Real dead code, deleted:**

- `hooks/useAcademyQuiz.ts` — leftover from the marketing split; marketing has its own adapted copy (`@marketing/lib/api/config` + its own endpoint URL).
- `hooks/useProjectKickoff.ts`, `tests/utils/metronome_alerts.ts` — zero references anywhere in the repo.

**Dependencies:**

- Removed 11 that no front file imports: `@contentful/rich-text-react-renderer`, `@contentful/rich-text-types`, `@radix-ui/react-navigation-menu`, `@tailwindcss/forms`, `embla-carousel-react`, `isomorphic-dompurify`, `jszip`, `react-beforeunload`, `tailwind-scrollbar-hide`, `concurrently`, `webpack-stats-plugin`. Checked each: every remaining consumer declares its own copy (marketing for contentful/embla, viz for navigation-menu/embla, sparkle for the two tailwind plugins via `@plugin` in its `tailwind.css`, `front/workers-build` for `isomorphic-dompurify`); `jszip`, `react-beforeunload`, `concurrently` and `webpack-stats-plugin` have no importer left anywhere.
- Added 8 that front imports but got by hoisting: `hono`, `jose`, `pdfjs-dist`, `@langfuse/core`, `@octokit/types`, `@opentelemetry/core`, `@tiptap/extension-code`, `@tiptap/extension-list-item`. Each pinned to the version already resolved in the lockfile, so nothing changes version — `package-lock.json` only loses the removed packages' trees.

## Trade-off worth flagging

`tests/utils/**` as an `entry` pattern means genuinely dead test helpers there will no longer be reported (that is how `tests/utils/metronome_alerts.ts` would have been missed). The cleaner fix is teaching knip about `front-api` as a workspace — I prototyped it, and it needs `routes/**` as entry plus triage of front-api's own ~30 unlisted dependencies, so it belongs in its own PR.

## Test plan

- `npm run knip -- --workspace front` with each of the three `--include` sets the CI action runs (`unlisted`; `dependencies,devDependencies,binaries,catalog`; `files`) → all exit 0.
- `npx tsgo --noEmit` in front → clean.
- `npm -w front-api run tsgo` → clean (front-api shares front's tsconfig and dependency hoisting).
- `npx biome check` on the changed files → clean.


davidebbo:
Superseded by a smaller PR that only deletes the three genuinely unused files, without touching the knip config or CI.

PR #29800, opened 2026-07-30
Update gpt-5.6-terra pricing

fontanierh:
## Description

OpenAI lowered GPT-5.6 Terra API prices (https://openai.com/business/pricing/#api). Updated the token pricing entries in front and marketing:

- input: 2.5 -> 2.0
- output: 15.0 -> 12.0
- cached input: 0.25 -> 0.2
- cache write: 3.125 -> 2.5 (kept at 1.25x input, same convention as sol/luna)

Also synced the informational tokenPricing declarations in the model_constructors endpoint files: Terra stream (global and EU +10%), plus the Luna stream/batch ones that #29798 missed (batch at 50% of standard).

Follow-up to #29798 (Luna).

## Tests

Checked the new numbers against the live pricing page. No tests pin these values.

## Risk

Low, data-only change. Affects credit consumption computed for Terra usage.

## Deploy Plan

Standard deploy.

### DIFF
--- a/front/lib/api/assistant/token_pricing/global.ts
+++ b/front/lib/api/assistant/token_pricing/global.ts
@@ -29,10 +29,10 @@ const CURRENT_MODEL_PRICING: Record<StaticModelIdType, PricingEntry> = {
   },
   // https://openai.com/api/pricing
   "gpt-5.6-terra": {
-    input: 2.5,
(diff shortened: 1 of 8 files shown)


aubin-tchoi (approved):
LGTM

PR #29113, opened 2026-07-17
[Pods] Set pod default agent from a conversation

xebeohp:
## Description
A bug fix for the pod default agent from this [conversation](https://app.dust.tt/w/0ec9852c2f/conversation/p2T9oKSa2a).

Added a `set_default_agent` tool to the `pod_manager` MCP server, letting an agent set or clear a Pod's default agent by name from within a conversation, and surfaced the current default in the server's `get_information` output. Also created a matching approval-card title so the tool call connecting to the MCP server shows a clear "set the Pod default agent" prompt.

Note: Only workspace published agents and global agents (dust, claude, gpt, etc) can be set from conversation. Unpublished agents must be set as default through the UI to prevent permissions leaks.

A similar fix for the pod default skills will be done in a stacked branch.

## Tests
Tested locally.

## Risk
Low risk.

## Deploy Plan
Deploy front.

PR #29803, opened 2026-07-30
Delete dead schemas and types in internal MCP servers

davidebbo:
One of five independent PRs from the same unused-export cleanup — #29802, #29803, #29804, #29807, #29809. Each is based on `main`, touches a disjoint set of files, and typechecks on its own, so they can be reviewed and merged in any order.

Second batch of the unused-export cleanup.

Scope: dead zod schemas and `z.infer` aliases in internal MCP server type modules — 71 declarations, ~364 lines, across nine servers. These model parts of a vendor API that no tool ever calls; a common shape when someone types the whole upstream surface and only a few tools get built.

| server | declarations removed |
|---|---|
| ashby | 28 |
| jira | 19 |
| freshservice | 8 |
| confluence | 5 |
| productboard | 4 |
| statuspage | 3 |
| ukg_ready | 2 |
| slab | 1 |
| clari_copilot | 1 |

## Pruned to a fixpoint, not one pass

Deleting a dead `z.infer` alias leaves the schema it was built from unreferenced. Removing only the aliases would have shifted the problem rather than fixed it, so each file was pruned repeatedly until nothing more was unreferenced. Eight schemas came out alongside the aliases they existed for — for example `ConfluenceListPagesResultSchema` behind `ConfluenceListPagesResult`, `JiraIssueLinkSchema` behind `JiraIssueLink`.

Every declaration was checked two ways before removal: no reference in any tracked `.ts`/`.tsx` in the repo, and no reference in the rest of its own file outside its own span.

## Not touched

`pod_manager` has three dead types too, but it is under active development this week (#29113 landed yesterday) and would conflict. It can go in a later batch.

Worth a reviewer's eye: if any of these schemas were deliberately written ahead of the tools that will use them, say so and I will restore them — they are one `git revert` away. Nothing here was referenced, so nothing changes at runtime either way.

## Test plan

- `npx tsgo --noEmit -p front/tsconfig.json` → clean. front's tsconfig includes `**/*.ts`, so a test file referencing any removed symbol would have failed here; none of the nine servers has tests of its own.
- `npx biome check --write front/lib/api/actions/servers` → clean (it dropped two imports the deletions orphaned).
- Verification pass repeated after applying: no remaining unreferenced declarations in the nine files.

### DIFF
--- a/front/lib/api/actions/servers/ashby/types.ts
+++ b/front/lib/api/actions/servers/ashby/types.ts
@@ -101,10 +101,6 @@ export const AshbyCandidateSearchResponseSchema = z
   .array(AshbyCandidateSchema)
   .optional();
 
-export type AshbyCandidateSearchResponse = z.infer<
-  typeof AshbyCandidateSearchResponseSchema
->;
-
 export const AshbyApplicationFeedbackListRequestSchema = z.object({
   applicationId: z.string(),
 });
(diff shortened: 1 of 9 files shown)


aubin-tchoi (approved):
Thanks!

PR #29805, opened 2026-07-30
[Admin Governance] Serialize user-level grants on a grant-tuple advisory lock

philipperolet:
## Description

Follows https://github.com/dust-tt/dust/pull/29896 (merged), which removed the last path granting instance tuples to non-backing regular_auto groups and made the invariant below globally true.

`grantToUser` finds-or-creates the `regular_auto` group backing a grant tuple (grantType, resourceType, resourceId) with no serialization: two concurrent first grants for the same tuple can both miss the existing group and try to create a duplicate. A duplicate backing group would make the grant ambiguous (`findRegularAutoGroupForGrant` picks arbitrarily), meaning a revoke could remove a user from one group while they retain access through the other. Today the duplicate is only prevented *accidentally*: `autoGroupName` is deterministic and the `groups (workspaceId, name)` unique index makes the losing transaction fail — a user-visible error held together by a display string.

Two layers, both in `GroupPermissionResource` (the single write path for `group_permissions`):

1. **Serialization** — the capability advisory lock (`getCapabilityLock`) is generalized into `getGrantLock`, keyed on the full grant tuple, and taken at the top of `grantToUser` and `revokeFromUser`. Concurrent first grants now serialize: the second transaction waits, finds the backing group created by the first, and adds its member gracefully instead of erroring. Locking `revokeFromUser` too closes the grant-vs-delete race: its delete-when-empty branch could otherwise act on a stale member count and wipe a concurrently added member's access. A repeat grant for an already-member user now returns Ok (idempotent, as documented) instead of propagating `user_already_member`.

2. **Enforcement at the choke point** — `grant()` rejects wiring a `regular_auto` group to a tuple that a *different* `regular_auto` group already holds (checked under the tuple lock), and `grantMany` rejects `regular_auto` groups outright (per-row checks would defeat the batch; auto grants go through `grant()`/`grantToUser`). This covers every resource-mediated path, not just `grantToUser`. (History note: this was reverted mid-PR while model-tier group overrides could still target regular_auto groups, and re-landed once #29896 removed that path.)

Type-wide (-1) capability transitions (`setForEverybody` / `setGroups`) keep the pre-existing lock key format so old and new pods mutually exclude across a rolling deploy or rollback; instance grants use a new key that includes `resourceId`.

The lock behavior itself has no new test: the test harness wraps each test in a single CLS transaction shared by all calls, where `pg_advisory_xact_lock` is reentrant, so a concurrency test would pass with or without the lock. The `grant()`/`grantMany` rejection paths, grantToUser idempotency, and user/group tier override coexistence are functionally tested.

## Risks

Blast radius: model-tier user grants (only current `grantToUser`/`revokeFromUser` callers) and governance capability transitions (unchanged lock key); `grant()` now rejects a second regular_auto group per tuple — no remaining caller can do this.
Risk: low

## Deploy Plan

- deploy front

### DIFF
--- a/front/lib/resources/group_permission_resource.ts
+++ b/front/lib/resources/group_permission_resource.ts
@@ -672,7 +672,13 @@ export class GroupPermissionResource extends BaseResource<GroupPermissionModel> 
     },
     transaction: Transaction
   ): Promise<void> {
-    const key = `group_permissions:${auth.getNonNullableWorkspace().id}:${resourceType}:${resourceId}:${grantType}`;
+    // Type-wide (-1) transitions keep the pre-resourceId key format so pods running older code
+    // mutually exclude with newer ones across a rolling deploy or rollback.
+    const workspaceId = auth.getNonNullableWorkspace().id;
+    const key =
+      resourceId === WHOLE_TYPE_RESOURCE_ID
+        ? `group_permissions:${workspaceId}:${resourceType}:${grantType}`
+        : `group_permissions:${workspaceId}:${resourceType}:${resourceId}:${grantType}`;
     // biome-ignore lint/plugin/noRawSql: advisory lock requires raw SQL
     await frontSequelize.query("SELECT pg_advisory_xact_lock(hashtext(:key))", {
       replacements: { key },
--- a/front/lib/resources/group_permission_resource.ts
+++ b/front/lib/resources/group_permission_resource.ts
@@ -153,7 +153,8 @@ export class GroupPermissionResource extends BaseResource<GroupPermissionModel> 
   }
 
   // Find the regular_auto group that already holds an instance-level grant for the given tuple.
-  // At most one such group is expected per (grantType, resourceType, resourceId).
+  // At most one such group exists per (grantType, resourceType, resourceId): grantToUser and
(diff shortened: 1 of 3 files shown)


front/lib/resources/group_permission_resource.ts
```diff
@@ -646,15 +653,33 @@ export class GroupPermissionResource extends BaseResource<GroupPermissionModel>
...
+      grantType: GrantType;
+      resourceType: GroupPermissionResourceType;
+      resourceId: number;
+    },
     transaction: Transaction
   ): Promise<void> {
-    const key = `group_permissions:${auth.getNonNullableWorkspace().id}:${resourceType}:${grantType}`;
+    // Type-wide (-1) transitions keep the pre-resourceId key format so pods running older code
+    // mutually exclude with newer ones across a rolling deploy or rollback.
+    const workspaceId = auth.getNonNullableWorkspace().id;
```
philipperolet, line 707: no need to be backwards compatible there

philipperolet:
FYI @Fraggle

matteotrab (commented):
LGTM

PR #29808, opened 2026-07-30
[front] Activation: consolidate poke plugins into Activation Management + inject nudge context

frankaloia:
## Summary
* This change is not at as large as it seems. It consolidates the two activation poke plugins (`join_activation_pod`, `send_activation_nudge`) into a single **Activation Management** plugin (provision / nudge / force-recreate), and changes how per-nudge context (session goal, pushed skill/agent) reaches the activation conversation: it now rides in the webhook event payload and is attached to the conversation as a content fragment (`includePayload: true`) rather than being persisted on the `activation_nudge` model.
- The purpose of this is to allow users of the plugin to more clearly define user context + session goal.
- Changes were made to allow this to be more resilient to avoid failing when a user already has resources such as a pod / trigger
- To help with my debugging, I also added a cheap poke pool to allow me to list the recommendations for a workspace

## Testing

Screenshot 2026-07-31 at 12 17 53 PM
Screenshot 2026-07-31 at 12 17 56 PM
Screenshot 2026-07-31 at 12 55 14 PM

## Risk

Low - Poke only

## Deploy
1. Deploy front


id13 (approved):
LGTM

PR #29811, opened 2026-07-31
Refactor pod conversations permissions to directly look at the conversation's pod access.

Fraggle:
## Description

Pod conversations (those with a non-null `spaceId`) now bypass the `requestedSpaceIds`-based access check entirely — they're accessible if and only if their pod space passes `canRead(auth)`. `getFilteredConversationsForAuth` splits incoming conversations into `podConversations` and `regularConversations`, resolves pod access directly from the already-fetched `spaceIdToSpaceMap`, and merges `accessiblePodConversations` at every early-return point. The single-conversation path in `canAccessConversation` gets the same early-exit via `SpaceResource.fetchByModelIds`. This removes the old TODO and fully implements the intended, simpler contract: pod conversations follow pod membership, not `requestedSpaceIds` logic.

## Tests

Tested locally.

## Risk

Low. Regular conversations are entirely unaffected — the split is based on `spaceId !== null`. Pod conversations now follow a stricter, cleaner path (`canRead(auth)` only), which is the correct invariant. Rollback-safe.

## Deploy Plan

Deploy front.

### DIFF
--- a/front/lib/api/assistant/conversation/branches.test.ts
+++ b/front/lib/api/assistant/conversation/branches.test.ts
@@ -437,11 +437,22 @@ describe("mergeConversationBranch", () => {
     const user = await UserFactory.basic();
     await MembershipFactory.associate(workspace, user, { role: "user" });
 
+    const internalAdminAuth = await Authenticator.internalAdminForWorkspace(
+      workspace.sId
+    );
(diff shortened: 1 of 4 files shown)


Fraggle:

- #29811 ◀

ElPicador (approved):

PR #29812, opened 2026-07-31
[front] enh: log conversation advisory lock hold time

id13:
## Description

First stack of two to get rid of xact_lock (will post a notebook to answer why).

Stack 1: Split the humongous transaction into smaller ones
Stack 2: Use row locks or unique indexes instead of an application lock.

What is not measured can't be improved, so step one of this quest to get rid of xact_log is to measure the time we spend holding a connection with it. We already have idle in tx logs as a proxy, but better have it twice.

logs will $$$, yes, will remove in the second stack.

## Tests

N/A

## Risk

Low

## Deploy Plan

Deploy front

### DIFF
--- a/front/lib/api/assistant/conversation/lock.ts
+++ b/front/lib/api/assistant/conversation/lock.ts
@@ -19,7 +19,7 @@ export async function getConversationRankVersionLock(
   conversation: ConversationWithoutContentType | ConversationResource,
   t: Transaction
 ) {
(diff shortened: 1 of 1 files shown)


id13:

- #29812 ◀
    - #29813
        - #29814
            - #29815
                - #29816
                    - #29817

flvndvd (commented):
LGTM

PR #29815, opened 2026-07-31
[front] refactor: resolve user mentions before the message transaction

id13:
## Description

Split mentions resolution and creation into resolution, and creation.
This is one step in the direction of removing all the un-necessary stuff out of a big transaction with a xact_lock
Then put the resolution before the big bad transaction.

## Tests

Existing tests

## Risk

Low

## Deploy Plan

- [ ] Deploy front

PR #29813, opened 2026-07-31
[front] fix: run origin message lookup inside the message transaction

id13:
## Description

side 🚗 : we used to not pass transaction on message creation inside the big bad transaction.

## Tests

N/A

## Risk

Low

## Deploy Plan

Deploy front

PR #29816, opened 2026-07-31
[front] refactor: precompute createAgentMessages inputs before the message transaction

id13:
## Description

createAgentMessages was doing three reads inside the big bad transaction: feature flags, model resolution (hits the DB on auto models) and the project space gating check. None of them need the lock, and the first two did not pass transaction.

Callers now compute all of this before the transaction and pass the results in.

Two subtleties:
- `canAgentBeUsedInProjectConversation` gating depends on branching (a branch lifts the restriction), so post derives the final answer in tx from branchId, no read needed.
- on the promote path the model selection comes from the promoted message, only known in tx. The no-selection default is resolved before, the explicit selection case (rare) stays inside.

## Tests

Existing tests adapted.

## Risk

Low
Blast radius: agent loop

## Deploy Plan

Deploy front

### DIFF
--- a/front/lib/api/assistant/conversation/messages.ts
+++ b/front/lib/api/assistant/conversation/messages.ts
@@ -1,8 +1,5 @@
 import { signalAgentUsage } from "@app/lib/api/assistant/agent_usage";
-import {
-  canAgentBeUsedInProjectConversation,
-  updateConversationRequirements,
-} from "@app/lib/api/assistant/conversation/permissions";
+import { updateConversationRequirements } from "@app/lib/api/assistant/conversation/permissions";
 import { getCompletionDuration } from "@app/lib/api/assistant/messages";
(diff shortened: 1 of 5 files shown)


id13:

- #29812
    - #29813
        - #29814
            - #29815
                - #29816 ◀
                    - #29817
                        - #29837
                            - #29838

front/lib/api/assistant/conversation/messages.ts
```diff
@@ -403,140 +439,113 @@ export const createAgentMessages = async (
...
-              mentionRow,
             });
-          },
-          {
-            // TODO(20260423 jd)
-            // all callsites of this function pass a transaction
-            // so concurent executor is purely cosmetic
-            // eventually we will get rid of these monstruous transactions
-            // so not putting a for loop
-            // and downsizing from 10 to max peak concurrency - 1
```
id13, line 558: hehehehe

id13:
PMRR

PR #29819, opened 2026-07-31
[front] enh: show skills in Tools "Used By" column

aubin-tchoi:
## Description

Admins can currently see which agents use an MCP server from Spaces > Tools, but not which skills.

This PR adds the skills in the "Used By" column, with the same design as in "Manage Skills"

image

This change is retro-compatible: the backend keeps sending agents only for now, the new frontend opts into skill usage with `withSkills=true`. Once we don't get any legacy client any more, a follow-up PR will remove the opt-in and have the backend always send both skills and agents.

## Tests

- Added tests.
- Tested locally.

## Risk

- Low.

## Deploy Plan

- Deploy front.

### DIFF
--- a/front-api/routes/w/[wId]/mcp/usage.ts
+++ b/front-api/routes/w/[wId]/mcp/usage.ts
@@ -1,16 +1,37 @@
 import { getToolsUsage } from "@app/lib/api/agent_actions";
-import type { GetMCPServersUsageResponseBody } from "@app/lib/api/mcp";
+import type {
+  GetMCPServersUsageResponseBody,
+  GetMCPServersUsageWithSkillsResponseBody,
+} from "@app/lib/api/mcp";
 import { workspaceApp } from "@front-api/middlewares/ctx";
 import type { HandlerResult } from "@front-api/middlewares/utils";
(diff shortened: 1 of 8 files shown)


front-api/routes/w/[wId]/mcp/usage.ts
```diff
@@ -1,16 +1,40 @@
...
+  validate("query", QuerySchema),
+  async (
+    ctx
+  ): HandlerResult<
+    GetMCPServersUsageResponseBody | GetMCPServersUsageWithSkillsResponseBody
+  > => {
+    const auth = ctx.get("auth");
+    const { withSkills } = ctx.req.valid("query");
+
+    if (withSkills === "true") {
```
Comment (fabiencelier) [inline front-api/routes/w/[wId]/mcp/usage.ts:31]: maybe add a TODO + comment here to not forget to remove it one day

fabiencelier (approved):
LGTM

front-api/routes/w/[wId]/mcp/usage.ts
Comment (aubin-tchoi) [reply front-api/routes/w/[wId]/mcp/usage.ts:31]: yeah good idea

PR #29820, opened 2026-07-31
[AGENT-LOOP] Motivate the model to write an answer when tool use is disabled

pmilliotte:
## Description

Claude Sonnet 4.6 and GPT 5.6 Luna sometimes end a turn with a reasoning block and no text at all — the agent loop then has nothing to publish and surfaces an `empty_content` ("No answer generated") or `max_step_reached` error, even though the model reached a conclusion. `tool_choice: "none"` on those steps (last step, or the retry after an empty one) is invisible to the model, so it plans a tool call, finds it cannot emit one, and goes silent instead: this appends an explicit instruction after the conversation history telling it that tools are unavailable and that it must write an answer — including saying so when it found nothing or was blocked. Also logs `reasoningEffort` and `chainOfThoughtLength` on empty turns so reasoning-only turns are distinguishable from genuinely empty ones.

Next steps: monitor whether we still see "The agent stopped without producing an answer." — the new `reasoningEffort` / `chainOfThoughtLength` fields on the empty-turn log tell us whether the remaining cases are reasoning-only turns or genuinely empty ones.

## Tests

Tested locally by forcing tool use off from the first step.

## Risk

Low.

## Deploy Plan

- Deploy `main` on `front`

### DIFF
--- a/front/temporal/agent_loop/lib/run_model.ts
+++ b/front/temporal/agent_loop/lib/run_model.ts
@@ -16,7 +16,10 @@ import { computeStepContexts } from "@app/lib/actions/utils";
 import { createClientSideMCPServerConfigurations } from "@app/lib/api/actions/mcp_client_side";
 import { renderConversationForModel } from "@app/lib/api/assistant/conversation_rendering";
 import { categorizeConversationRenderErrorMessage } from "@app/lib/api/assistant/errors";
-import { constructPromptMultiActions } from "@app/lib/api/assistant/generation";
+import {
+  constructPromptMultiActions,
+  renderToolUseDisabledUserMessage,
+} from "@app/lib/api/assistant/generation";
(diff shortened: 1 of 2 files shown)


front/temporal/agent_loop/lib/run_model.ts
```diff
@@ -705,6 +708,14 @@ export async function runModel(
     return null;
   }
 
+  if (disableToolUse) {
+    // Tool choice "none" alone leaves the model with nothing to do; spell it out
+    // so it writes an answer. Its tokens sit outside the render budget.
+    modelConversationRes.value.modelConversation.messages.push(
```
aubin-tchoi, line 714: seems quite high in the stack to do this, can't we implement this closer to where we render the conversation

aubin-tchoi: approving already as this can always be refactored later

aubin-tchoi (approved):
LGTM

PR #29822, opened 2026-07-31
[dust-hive] fix: limit dhb completion to warm hives

aubin-tchoi:
## Description

`dhb` opens a running environment app, but its completion currently suggests stopped and cold environments too. Bash and Zsh completion now reuse the existing warm-environment selector so only warm hives are suggested.

This only changes completion. Manually entered environment names remain accepted.

## Tests

- Tested locally with Bash and Zsh completion helpers.

## Risk

- Low.

## Deploy Plan

- No deploy.


fontanierh (approved):
🙏

PR #29823, opened 2026-07-31
skill permission check

ykmsd:
## Description
This PR is to add missing permission check in import/update handler and in the resource

### DIFF
--- a/front/lib/resources/skill/skill_resource.ts
+++ b/front/lib/resources/skill/skill_resource.ts
@@ -421,6 +421,22 @@ export class SkillResource extends BaseResource<SkillConfigurationModel> {
   ): Promise<SkillResource> {
     const owner = auth.getNonNullableWorkspace();
 
+    assert(
+      auth.isKey() ||
+        (await auth.hasWorkspacePermission("create", "skill")),
+      "User is not authorized to create skills"
+    );
+
+    if (
+      blob.availability === "users_and_agents" &&
+      (await hasFeatureFlag(auth, "admin_governance_skill_publication"))
+    ) {
+      assert(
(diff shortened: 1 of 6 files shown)


front/lib/resources/skill/skill_resource.ts
```diff
@@ -421,6 +421,22 @@ export class SkillResource extends BaseResource<SkillConfigurationModel> {
   ): Promise<SkillResource> {
     const owner = auth.getNonNullableWorkspace();
 
+    assert(
+      auth.isKey() ||
+        (await auth.hasWorkspacePermission("create", "skill")),
+      "User is not authorized to create skills"
+    );
```
matteotrab, line 427:
```suggestion
    assert(
        (await auth.hasWorkspacePermission("create", "skill")),
      "User is not authorized to create skills"
    );
```

No need to check if the auth is key here. The admin keys will have the permission, the legacy builder keys will have the permission with the builder group and we do not want the user keys to create skills

front/tests/utils/SkillFactory.ts
```diff
@@ -38,6 +40,30 @@ export class SkillFactory {
...
+          kind: "regular_auto",
+          workspaceId: workspace.id,
+        },
+        { memberIds: [user.id] }
+      );
+      await GroupPermissionResource.grantTypeWide(adminAuth, {
+        group,
+        grantType: "create",
+        resourceType: "skill",
+      });
```
matteotrab: You can replace it with `grantWorkspacePermission` from front/tests/utils/permissions.ts

matteotrab (approved):

PR #29824, opened 2026-07-31
Disable agents picking in sidekick conversations

fabiencelier:
## Description

Disable the agent picking in sidekick conversations:
 - no picker at the bottom
 - no mentions possible with `@dust`
 - strip agent mention from copy-pasting a mention

There is no reason to switch agent in this context, you should only talk to sidekick

image

## Tests

tested locally

## Risk

low

## Deploy Plan

deploy front


Fraggle (approved):
LGTM

PR #29825, opened 2026-07-31
[Skill Builder] Gate auto-discoverable availability

philipperolet:
## Description

Require the `make_discoverable` skill permission, in addition to publication permission, when the skill builder creates a skill with availability set to users and agents. This keeps auto-discoverability under the dedicated governance setting.

## Risks

Blast radius: Skill creation when skill publication governance is enabled.
Risk: low

## Deploy Plan

- deploy front-api

## Tests

- [x] Private skill POST endpoint test suite (40 tests)


front-api/routes/w/[wId]/skills/index.ts
Comment (philipperolet) [inline front-api/routes/w/[wId]/skills/index.ts:1]: the check was already present for PATCH of existing skills

matteotrab (commented):
LGTM

ykmsd:
ty!

PR #29827, opened 2026-07-31
Persist the run linkage on agent step content for consumption attribution

flvndvd:
## Description

Consumption attribution needs to explain what drove a message's cost, split across prompt tokens, generated tokens, reasoning and each tool call. That breakdown is computed asynchronously after the message completes, since it tokenizes text through core and has no business on the agent loop. To attribute a recorded run usage to the content it produced, the job needs a link that the schema doesn't have today: usage rows point only to a run, and generated contents point nowhere back.

This adds that link. Each generated content now stores the identifier of the model run that emitted it, written in the loop from the same value the message already records, so the two are guaranteed to match. The job can then group a message's contents by run and hand each run's usage and contents to the attribution builder.

Only interactive model output carries the identifier. Synthetic content, tool result text, errors and the batch path leave it empty, and title generation and compaction stay out of scope. Run grain is enough because one interactive run records exactly one usage. It keys on the run identifier, not the step number, which is reused across versions and re-executions. The column is unindexed since the job reaches these rows through the message.

The cache shape learned the field and its version was bumped so stale entries are dropped.

The migration is a pre-deploy nullable column, no backfill.

## Deploy Plan

- [ ] Apply SQL migration
- [ ] Deploy front

### DIFF
--- a/front/lib/resources/agent_step_content/cache.ts
+++ b/front/lib/resources/agent_step_content/cache.ts
@@ -7,8 +7,10 @@ import groupBy from "lodash/groupBy";
 
 export const AGENT_STEP_CONTENT_CACHE_TTL_MS = 2 * 60 * 60 * 1000; // 2 hours
 
-// Bump the `:v1` suffix any time the cached value shape changes, so stale entries
-// from previous formats are orphaned instead of mis-parsed.
+// Bump the version any time the cached value shape changes, so stale entries from previous formats
+// are orphaned instead of mis-parsed.
+const AGENT_STEP_CONTENT_CACHE_VERSION = 2;
+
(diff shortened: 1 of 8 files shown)


front/lib/resources/agent_step_content/cache.ts
```diff
@@ -270,29 +274,3 @@ export async function tryHydrateAgentStepContentsFromCache({
     return null;
   }
 }
-
-export function toCachedAgentStepContent(blob: {
```
flvndvd, line 274: 🚗 moved to the resource.

id13 (approved):
LGTM 👍

avervaet (approved):

PR #29831, opened 2026-07-31
Skip indexing conversations exceeding size limit in projects datasource

fabiencelier:
## Description

Following the same pattern as #19116 for Intercom, this skips conversations exceeding the plan's document size limit in the dust_project connector instead of letting DataSourceQuotaExceededError propagate. Previously, a single oversized conversation caused dustProjectConversationsFullSyncActivity to fail on every retry, permanently blocking the full sync workflow. The skipped conversation is logged with a warning and simply omitted from the data source until the plan limit is raised.

## Tests

not tested

## Risk

easy to revert
very large conversations won't be indexed in project

## Deploy Plan

deploy connector


Fraggle (approved):
Thx

PR #29832, opened 2026-07-31
Compute agent message consumption attribution from the analytics queue

flvndvd:
## Description

An earlier set of PRs added the table and the pure computation that turn a run's reported token usage into an attribution breakdown, but nothing wrote those rows. This connects the two: once an agent message settles, its per-run token consumption is broken down and stored.

This is analytics, not billing. The authoritative charge is still computed and stored on its own by the credit path, rounded and cache-aware. The breakdown here exists to explain the relative composition of that cost, prices every token at full rate on purpose, and is not expected to sum to the bill. To keep that separation clear it runs on the analytics side rather than folded into the credit computation, as its own workflow on the analytics queue, launched from the finalize activities alongside the existing analytics work once the message has settled.

This first version covers the model token buckets only: for each of the message's run usages it records an input, an output and a reasoning row. Per-tool attribution needs tokenization of the tool result footprint and is left to a follow-up, so the builder is invoked with no tool calls. The write is idempotent by message, version, run usage and bucket, so a re-finalize on interrupt, resume, tool confirmation or a worker retry re-inserts the same identities as no-ops and only adds rows for usages that are new since the last pass.

To let a stored row point at the exact usage it came from, the usage fetch helper now also returns the usage row identifier, which previously exposed only the owning run. Coverage relies on the settled-message trigger; a reconciliation pass for messages that never reach that point is intentionally out of scope for now.

🚗 The analytics activities file was becoming large, so it is now a folder: the existing module keeps its history under a descriptive name, the new work lives in its own file, and a barrel keeps the import path stable.

## Deploy Plan

There are no model changes, only some comments.


front/temporal/analytics_queue/activities/agent_analytics.ts
flvndvd, line 1: 🚗 File became too long, moved to a folder so it's easier to split.

sfriquet (approved):
this runs in parallel with `launchAgentMessageAnalytics` but if we need this data in ES we'll likely need to do it before we run the activity that pushes messages to ES.

flvndvd:
> this runs in parallel with `launchAgentMessageAnalytics` but if we need this data in ES we'll likely need to do it before we run the activity that pushes messages to ES.

Yes, we will probably change a bit the wires here, cause everyone has been fulling stuff in the analytics queue 🙈, so it needs some more love.

id13 (approved):

PR #29833, opened 2026-07-31
fix: display updated skill editors after saving

matteotrab:
## Description

This PR fixes a bug where the skill editors list was not refreshed in the UI after saving a skill.

- After a successful save, `mutateEditors` is called with the updated editors data returned from the API response, with `revalidate: false` to avoid an unnecessary refetch.

## Tests

Manually.

## Risks

Low.

## Deploy Plan

Standard deployment.


matteotrab:

- #29833 ◀

ykmsd (approved):
💯

PR #29834, opened 2026-07-31
[Skills] Restricted skills fix scoping

xebeohp:
## Description

Restricted skills must be scoped to at most one Pod. Otherwise, a skill that can access Pod-specific data could become visible to members of another Pod or to workspace-level users.

This PR enforces that rule in two places:

- **At creation and update time**, a skill is rejected if its requested spaces, including spaces inherited through nested skill references, contain more than one Pod.
- **At selection time**, Pod-scoped skills are filtered out before they reach the UI. They are not merely rejected after selection, so they remain unsearchable outside the intended scope.

### Three-mode skill visibility filter

The old boolean `globalSpaceOnly` flag was replaced with a discriminated `spaceScope` filter. The three modes are mutually exclusive:

1. **`global_only`**: return only skills restricted exclusively to the workspace global space. This is used by workspace-global skills
2. **`exclude_pod_scoped`**: return global skills and skills restricted to non-Pod spaces, while hiding every skill restricted to a Pod. This is used when the current conversation is not in a Pod. This is required since a user should be able to use a skill restricted to a space throughout all of their conversations.
3. **`pod`**: return global skills, non-Pod-restricted skills, and skills belonging to the current Pod, while hiding skills belonging to other Pods. This is used when the current conversation is in a Pod.

The API exposes these scopes through the mutually exclusive `globalSpaceOnly`, `excludePodScopedSkills`, and `podId` query parameters. Frontend callers derive the appropriate scope from the conversation context and use it consistently in the slash-command picker, capabilities picker, and Pod skill settings. The union-based API prevents callers from accidentally combining contradictory visibility rules.

A UI update for the skills settings page will follow in the next PR.

## Tests

Added and updated API route coverage for:

- Listing skills in each of the three visibility modes.
- Rejecting mutually exclusive scope parameters.
- Rejecting skills restricted to multiple Pods at creation time.
- Rejecting skills restricted to multiple Pods when updating an existing skill, including nested skill references.

Relevant test files:

- `front-api/routes/w/[wId]/skills/index.test.ts`
- `front-api/routes/w/[wId]/skills/[sId]/index.test.ts`

Local validation performed: `git diff --check`. The focused test suites were not run locally; final validation is expected through PR CI.

## Risk

Low risk. The change affects restricted-skill validation and visibility filtering only. Existing unrestricted skills and skills restricted to non-Pod spaces remain visible under the same conditions. No database migration is required, and the change is straightforward to roll back.

## Deploy Plan

Deploy front. No migration or ordering dependency is required.

### DIFF
--- a/front-api/routes/w/[wId]/skills/index.ts
+++ b/front-api/routes/w/[wId]/skills/index.ts
@@ -3,12 +3,14 @@ import { AttachedKnowledgeSchema } from "@app/lib/api/skills/schemas";
 import {
   getReferencedSkillSpaceModelIds,
   resolveAdditionalRequestedSpaceModelIds,
+  validateAtMostOnePodSpace,
 } from "@app/lib/api/skills/space_requirements";
 import { hasFeatureFlag } from "@app/lib/auth";
 import { DataSourceViewResource } from "@app/lib/resources/data_source_view_resource";
 import { FileResource } from "@app/lib/resources/file_resource";
 import { MCPServerViewResource } from "@app/lib/resources/mcp_server_view_resource";
 import { SkillResource } from "@app/lib/resources/skill/skill_resource";
+import { SpaceResource } from "@app/lib/resources/space_resource";
 import logger from "@app/logger/logger";
 import type {
   GetSkillsResponseBody,
@@ -133,6 +135,9 @@ app.get(
     const withMessageCount = ctx.req.query("withMessageCount") === "true";
     const status = ctx.req.query("status");
     const globalSpaceOnly = ctx.req.query("globalSpaceOnly");
+    const podSpaceId = ctx.req.query("podSpaceId");
+    const excludePodScopedSkills =
+      ctx.req.query("excludePodScopedSkills") === "true";
     const onlyCustom = ctx.req.query("onlyCustom");
     // @deprecated Use availability instead. Kept while old clients still send it.
     const isDefault = ctx.req.query("isDefault");
@@ -183,9 +188,27 @@ app.get(
       });
     }
 
+    let resolvedPodSpaceId: number | null | undefined;
+    if (podSpaceId) {
+      const podSpace = await SpaceResource.fetchById(auth, podSpaceId);
+      if (!podSpace) {
+        return apiError(ctx, {
+          status_code: 404,
+          api_error: {
+            type: "space_not_found",
+            message: `Space "${podSpaceId}" was not found.`,
+          },
+        });
+      }
+      resolvedPodSpaceId = podSpace.id;
+    } else if (excludePodScopedSkills) {
+      resolvedPodSpaceId = null;
+    }
+
     const allSkills = await SkillResource.listByWorkspace(auth, {
       status: skillStatus,
       globalSpaceOnly: globalSpaceOnly === "true",
+      podSpaceId: resolvedPodSpaceId,
       onlyCustom: onlyCustom === "true",
       availability,
       withInstructions: false,
@@ -478,6 +501,20 @@ app.post(
       ...referencedSkillSpaceIds,
       ...additionalRequestedSpaceIdsRes.value,
     ]);
+
+    const podSpaceValidation = await validateAtMostOnePodSpace(
+      auth,
+      requestedSpaceIds
+    );
+    if (podSpaceValidation.isErr()) {
+      return apiError(ctx, {
+        status_code: 400,
+        api_error: {
+          type: "invalid_request_error",
+          message: podSpaceValidation.error.message,
+        },
+      });
+    }
 
     // Validate file attachments if provided.
     let files: FileResource[] | undefined;
--- a/front-api/routes/w/[wId]/skills/index.ts
+++ b/front-api/routes/w/[wId]/skills/index.ts
@@ -188,7 +188,7 @@ app.get(
       });
     }
 
-    let resolvedPodSpaceId: number | null | undefined;
+    let resolvedPodSpaceModelId: number | null | undefined;
     if (podSpaceId) {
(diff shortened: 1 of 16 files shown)


xebeohp:

- #29834 ◀
    - #29866

front/lib/swr/skill_configurations.ts
```diff
@@ -128,6 +130,13 @@ export function useSkills({
   if (globalSpaceOnly) {
     queryParams.set("globalSpaceOnly", "true");
   }
+  if (podContext !== undefined) {
+    if (podContext !== null) {
+      queryParams.set("podSpaceId", podContext);
+    } else {
+      queryParams.set("excludePodScopedSkills", "true");
```
aubin-tchoi: quite complex to understand that passing `null` does something different from `undefined`

front/components/editor/extensions/shared/slash_suggestion/useSlashCommandCapabilities.ts
```diff
@@ -76,6 +78,7 @@ export function useInputBarSlashCommandCapabilities({
   const { skills, isSkillsLoading } = useSkills({
     owner,
     status: "active",
+    podContext: conversationPodSpaceId ?? null,
```
aubin-tchoi: if you're not in a pod you can't use the skills that are restricted to a pod here?

front/lib/swr/skill_configurations.ts
```diff
@@ -128,6 +130,13 @@ export function useSkills({
   if (globalSpaceOnly) {
```
aubin-tchoi: would be nice to consolidate with this (to prevent allowing the caller to pass both `globalSpaceOnly` and a pod space ID, which does not make much sense)

front-api/routes/w/[wId]/skills/index.ts
```diff
@@ -180,9 +185,27 @@ app.get(
       });
     }
 
+    let resolvedPodSpaceModelId: number | null | undefined;
```
Comment (aubin-tchoi) [inline front-api/routes/w/[wId]/skills/index.ts]: nit:
```suggestion
    let resolvedPodSpaceModelId: ModelId | null | undefined;
```

front/lib/resources/skill/skill_resource.ts
```diff
@@ -1520,14 +1522,37 @@ export class SkillResource extends BaseResource<SkillConfigurationModel> {
...
+
+    return filteredSkills.filter((skill) => {
+      const skillPodSpaceIds = skill.requestedSpaceIds.filter((id) =>
+        podSpaceIds.has(id)
+      );
+      if (skillPodSpaceIds.length === 0) {
+        return true;
+      }
+      return (
+        podSpaceModelId !== null && skillPodSpaceIds.includes(podSpaceModelId)
```
aubin-tchoi: wait `podSpaceModelId` `null` and `undefined` mean different things?

front/lib/resources/skill/skill_resource.ts
```diff
@@ -1520,14 +1522,37 @@ export class SkillResource extends BaseResource<SkillConfigurationModel> {
       withFileAttachments,
     });
 
+    let filteredSkills = skills;
+
     if (globalSpaceOnly) {
```
aubin-tchoi: this is incompatible with passing a `podSpaceModelId` no? if you give both then necessarily there are no matching skills

front-api/routes/w/[wId]/skills/index.ts
Comment (aubin-tchoi) [reply front-api/routes/w/[wId]/skills/index.ts]: also why can it be both null and undefined?

front-api/routes/w/[wId]/skills/index.ts
```diff
@@ -130,6 +132,9 @@ app.get(
     const withRelations = ctx.req.query("withRelations");
     const status = ctx.req.query("status");
     const globalSpaceOnly = ctx.req.query("globalSpaceOnly");
+    const podSpaceId = ctx.req.query("podSpaceId");
+    const excludePodScopedSkills =
+      ctx.req.query("excludePodScopedSkills") === "true";
```
Comment (aubin-tchoi) [inline front-api/routes/w/[wId]/skills/index.ts:140]: in what context do we want to exclude pod-scoped skills?

front-api/routes/w/[wId]/skills/index.ts
```diff
@@ -130,6 +132,9 @@ app.get(
     const withRelations = ctx.req.query("withRelations");
     const status = ctx.req.query("status");
     const globalSpaceOnly = ctx.req.query("globalSpaceOnly");
+    const podSpaceId = ctx.req.query("podSpaceId");
```
Comment (aubin-tchoi) [inline front-api/routes/w/[wId]/skills/index.ts]: let's call it `podId` no? not sure exactly what our naming conventions are here

aubin-tchoi (commented):
looks directionally great, left a few comments

front-api/routes/w/[wId]/skills/index.ts
Comment (xebeohp) [reply front-api/routes/w/[wId]/skills/index.ts]: Fixed

front/components/editor/extensions/shared/slash_suggestion/useSlashCommandCapabilities.ts
xebeohp: Yes, a Pod-restricted skill must be unsearchable outside its own Pod.

front/lib/swr/skill_configurations.ts
xebeohp: Fixed.

front-api/routes/w/[wId]/skills/index.ts
Comment (xebeohp) [reply front-api/routes/w/[wId]/skills/index.ts:140]: You would want to exclude pod-scoped skills when you have a space-scoped skill as well. This second skill is restricted to a space (not pod) so you should still be able to use it in your normal conversations. Setting to just global skills only would get rid of this third in-between case.

PR #29837, opened 2026-07-31
[front] enh: enforce single agent mention on user message edits

id13:
## Description

We used to support running multiple agent mentions per msg. This is now impossible due to steering. 
That being said, we still have some logic relying on this big deadly lock to handle multiple mentions.
I intend to remove it, and for that will first throw loudly if this is still happening

## Tests

Added some tests

## Risk

Low
This *should* be a no-op

## Deploy Plan

- [ ] Deploy front

PR #29817, opened 2026-07-31
[front] refactor: signal agent usage after the message transaction commits

id13:
## Description

Add a transaction param + hook in signalAgentUsage, to avoid phantom reads when we cut into multiple smaller transactions (or no transaction at all) without a big lock 

## Tests

N/A

## Risk

Low, it's a no-op change

## Deploy Plan

Merge

PR #29838, opened 2026-07-31
[front] refactor: collapse createAgentMessages to a single agent mention

id13:
## Description

Currently we do a lot of things in 2 big transactions (postUserMessage and editUserMessage), with a lock and some network I/O.
This stack is about removing everything that can be removed from this big transaction, with this PR being the final step: propagate the  "one agent mention per postUserMessage / editUserMessage" all the way down.

We currently have this sitting at the top of postUserMessage (used by both v1 and private)

```
  // Steering invariants: enforce single agent loop per conversation.
  if (explicitAgentMentions.length > 1) {
    return new Err({
      status_code: 400,
      api_error: {
        type: "invalid_request_error",
        message: "Only one agent can be mentioned per message.",
      },
    });
  }
```

Yet we still have a [concurrentExecutor](https://github.com/dust-tt/dust/pull/29838/changes#diff-539f8a721a6829d9a966dad341ba48671e39aff72aa06fd96141ac5a9e1446ebL490) on unique agent mentions. A vestige of before-steering times.

This PR is more about cleanup than anything else

## Tests

Tested locally
Tested in front-edge

## Risk

Low
Blast radius: postUserMessage / editUserMessage

## Deploy Plan

- Deploy front

### DIFF
--- a/front/lib/api/assistant/conversation.ts
+++ b/front/lib/api/assistant/conversation.ts
@@ -20,10 +20,10 @@ import {
   attributeUserFromWorkspaceAndEmail,
   createAgentMessages,
   createUserMessage,
-  resolveModelsForMentionedAgents,
+  resolveModelForMentionedAgent,
 } from "@app/lib/api/assistant/conversation/messages";
 import {
-  canAgentBeUsedInProjectConversation,
+  isAgentRestrictedBySpaceUsage,
   updateConversationRequirements,
 } from "@app/lib/api/assistant/conversation/permissions";
 import { ensureConversationTitle } from "@app/lib/api/assistant/conversation/title";
@@ -730,7 +730,6 @@ export async function postUserMessage(
   ]);
 
   let agentConfigurations = removeNulls(results[0]);
-  const restrictedAgentIds = new Set<string>();
 
   // Retired global agents can't be invoked (new conversations or new messages).
   // The internal `run_agent` path is exempt: some hidden sub-agents are retired.
@@ -795,17 +794,6 @@ export async function postUserMessage(
         },
       });
     }
-
-    if (isPartOfPod) {
-      const canAgentBeUsed = await canAgentBeUsedInProjectConversation(auth, {
-        configuration: agentConfig,
-        conversation,
-      });
-
-      if (!canAgentBeUsed) {
-        restrictedAgentIds.add(agentConfig.sId);
-      }
-    }
   }
 
   // TODO(2026-07-31 SEC): this allow spoofing as we trust blindly the user email from the metadata.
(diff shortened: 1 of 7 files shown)


id13:

- #29812
    - #29813
        - #29814
            - #29815
                - #29816
                    - #29817
                        - #29837
                            - #29838 ◀

front/lib/api/assistant/conversation/messages.ts
```diff
@@ -466,152 +454,115 @@ export const createAgentMessages = async (
...
+        const messageRow = await MessageModel.create(
           {
-            // TODO(20260423 jd)
-            // all callsites of this function pass a transaction
-            // so concurent executor is purely cosmetic
-            // eventually we will get rid of these monstruous transactions
-            // so not putting a for loop
-            // and downsizing from 10 to max peak concurrency - 1
-            concurrency: 4,
-          }
```
id13, line 613: This is what this whole PR is about : removing code that currently is inert, like a non detonated bomb, but could, if we decide for w/e reason to allow some codepath to launch multiple agents in the same msg, in a world without the locked transaction, to try locking the same rows. Removing the need for this makes it more future proof

id13:
PMRR

front/lib/api/assistant/conversation.ts
```diff
@@ -795,17 +794,6 @@ export async function postUserMessage(
         },
       });
     }
-
-    if (isPartOfPod) {
```
id13, line 799: deported to isAgentRestrictedBySpaceUsage

front/lib/api/assistant/conversation/messages.test.ts
```diff
@@ -1683,7 +1419,7 @@ describe("createAgentMessages", () => {
         content: `Hello @${agentConfig.name}`,
       });
 
-      const mentions: MentionType[] = [
+      const _mentions: MentionType[] = [
```
fabiencelier, line 1422: Why keep all this mentions object if they are now unused ?

front/lib/api/assistant/conversation/permissions.ts
```diff
@@ -165,6 +165,26 @@ export async function canAgentBeUsedInProjectConversation(
   return true;
 }
 
+export async function isAgentRestrictedBySpaceUsage(
```
fabiencelier, line 168: weird naming as you don't check the restricted space of the agent 
Maybe isAgentRestrictedByPodUsage ?

fabiencelier (commented):
LGTM ✅

front/lib/api/assistant/conversation/permissions.ts
id13: pod = space, it just happens so that for now you can't launch conversation in other spaces

front/lib/api/assistant/conversation/messages.test.ts
id13: Fair point !

id13: Done in #29909: dropped all 13 of them (12 in messages.test.ts, 1 in validate_actions.test.ts) and the imports that went unused with them.

PR #29839, opened 2026-07-31
[Analytics] Require admin scope for exports

philipperolet:
## Description

Restricts workspace analytics exports to API keys with admin scope. Builder-scoped keys were temporarily accepted during migration; this restores the intended admin-only boundary and updates the permission tests.

## Risks

Blast radius: Analytics export requests made with non-admin API keys.
Risk: standard

## Deploy Plan

- Deploy front-api.

## Test

- [x] Verified admin API keys are accepted while builder and read-only API keys are rejected.


matteotrab (commented):
LGTM

PR #29842, opened 2026-07-31
fix: show suggested skills

matteotrab:
## Description

This PR fixes the visibility of suggested skills so that users with skill creation permissions can see them in the skills list.

- Suggested skills are created with an empty editor group and editors-only availability, which caused them to be hidden from everyone under the existing editor-visibility filter.
- The fix adds a status exemption: skills with `status === "suggested"` are now also shown to users who have the `create` skill permission and can administrate the skill. Being able to administrate the skill is necessary to add the user creating the skill to the editors.

## Tests

Added tests to:
- Verify that suggested skills with `editors` availability are listed to admins who can create skills.
- Verify that suggested skills are not listed to regular members who cannot administrate skills, even if they have the `create/skill` capability.

## Risks

Low.

## Deploy Plan

Standard deployment.

### DIFF
--- a/front-api/routes/w/[wId]/skills/index.test.ts
+++ b/front-api/routes/w/[wId]/skills/index.test.ts
@@ -148,6 +148,82 @@ describe("GET /api/w/:wId/skills", () => {
     expect(skillNames).not.toContain("Someone Else's Unpublished Skill");
   });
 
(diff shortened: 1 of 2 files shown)


matteotrab:

- #29842 ◀

ykmsd (approved):

PR #29846, opened 2026-07-31
Show approve/decline for restricted agents in Pod conversations.

Fraggle:
## Description

`agent_restricted_by_space_usage` mentions were previously handled by `MentionInvalid` as a dead-end warning (dismiss only). They now route through `MentionValidationRequired`, giving Pod editors an approve/decline card: title "Run [agent] in this Pod conversation?", with copy explaining that private-space outputs will be visible to Pod members. `MentionInvalid` is narrowed to `user_restricted_by_conversation_access` only and its switch is collapsed. `MentionValidationRequired` gains a `ValidatableMention` union type, an `isRestrictedAgent` branch for title/description/`approveLabel`, the agent's avatar instead of the generic chat icon, and a `dismissed` guard on the early return. Success notification copy in `useMentionValidation` is updated for the agent case.

image

## Tests

Tested locally in a Pod conversation with an agent configured to use a private space.

## Risk

Low. Front-only, no API or DB changes. The `agent_restricted_by_space_usage` branch previously had no actionable path; adding approve/decline is additive and safe to roll back.

## Deploy Plan

Deploy front.

PR #29845, opened 2026-07-31
Wire agent mention approve/reject through the mentions API.

Fraggle:
## Description

The `POST .../mentions` route previously rejected any non-user mention type with a 400. Agent approve/reject actions (e.g. from a restricted agent mention flow) hit this same endpoint but weren't handled. The `else` branch is now split: `type === "user"` routes to `validateUserMention` as before, and everything else routes to `validateAgentMention` (imported from `validate_agent_mention.ts`), passing `conversationId`, `agentConfigurationId`, `messageId`, and `approvalState`.

## Tests

Tested manually by triggering agent mention approve/reject actions through the API and verifying the correct handler is invoked.

## Risk

Low. The change is additive — existing `dismiss` and `user` branches are untouched. The previous hard-reject for non-user types is replaced by a delegating else branch that already has its own error handling via `validateAgentMention`.

## Deploy Plan

Deploy front-api.

PR #29844, opened 2026-07-31
Add validateAgentMention to approve restricted agents in-place.

Fraggle:
## Description

Adds `validateAgentMention` in a new module (`validate_agent_mention.ts`) that lets a Pod conversation owner approve or reject a restricted agent mention in-place, without re-posting the user message.

- **Approve**: acquires `getConversationRankVersionLock`, locks the `MentionModel` row with `SELECT FOR UPDATE`, marks it `approved`, calls `createAgentMessages` with the new `approve_existing_mention` type (reusing the existing row instead of inserting a duplicate), then fires `runAgentLoopWorkflow`. This intentionally bypasses `canAgentBeUsedInProjectConversation`.
- **Reject**: updates the mention status to `rejected` and publishes message events to sync the UI.
- Extends `createAgentMessages` metadata union with `approve_existing_mention`, which falls through into `case "create"` but skips the Pod space-usage restriction check and mention row insertion.
- Exports `checkMessagesLimit` (previously unexported) to avoid an import cycle with `conversation.ts`.
- Emits an `auditLog` entry on every approval.

## Tests

Tested manually in Pod conversations: approve and reject flows, message limit enforcement, and duplicate-approval guard (mention row locked with `SELECT FOR UPDATE`).

## Risk

Low. The `approve_existing_mention` branch is an entirely new code path — existing `create` behavior in `createAgentMessages` is unchanged. The only modification to existing code is exporting `checkMessagesLimit`. Rollback is safe (revert front deploy).

## Deploy Plan

Deploy front.

PR #29843, opened 2026-07-31
Stop creating conversation branches for restricted Pod agents.

Fraggle:
## Description

Removes the auto-branch logic from `postUserMessage` for restricted Pod agents. Previously, mentioning an agent that lacked access to the Pod's space would trigger `ConversationBranchResource.makeNew` — complete with optional anchor messages — and silently route the reply into a new branch. Now, the mention is recorded directly on the main conversation with status `agent_restricted_by_space_usage`, no agent message is created, and no branch is opened. Deletes ~93 lines of branching logic (`shouldCreateBranch`, `getBranchCreationContext`, anchor-message creation) from `postUserMessage`.

## Tests

Updated existing integration tests in `conversation.test.ts` to reflect the new behavior: assert `branchesAfter.length === 0`, `agentMessages.length === 0`, and `mentionRow.status === "agent_restricted_by_space_usage"`. Covered non-empty conversations, empty conversations (previously created an anchor), and content-fragment-only conversations.

## Risk

Low. The branch-creation path was a Pod-specific code branch gated on `isPartOfPod` and `canAgentBeUsedInProjectConversation`. Existing conversations on branches are unaffected; the restriction-status mention path already existed. Rollback is a revert with no schema changes to undo.

## Deploy Plan

Deploy front.

PR #29847, opened 2026-07-31
Allow retry of restricted agents already in a Pod conversation.

Fraggle:
## Description

`retryAgentMessage` was calling `canAgentBeUsedInProjectConversation` on retry, which blocked users from retrying a restricted agent in a Pod conversation — even when that agent message already existed (approved via `validateAgentMention` or usable at creation time). Removed the check from the retry path: once an agent message exists in the conversation, the access decision was already made upstream and retry should go through unconditionally.

## Tests

Updated the existing unit test in `conversation.test.ts` that previously asserted a 400 error on retry of a restricted agent in a project conversation — it now asserts success and verifies `launchAgentLoopWorkflow` and `publishAgentMessagesEvents` are called.

## Risk

Low. The removed guard was overly conservative and only applied on the retry path. The initial agent message creation and mention approval paths are unaffected. Rollback is safe — reverting re-adds the guard.

## Deploy Plan

Deploy front.

### DIFF
--- a/front/lib/api/assistant/conversation.test.ts
+++ b/front/lib/api/assistant/conversation.test.ts
@@ -791,22 +791,26 @@ describe("retryAgentMessage", () => {
       vi.clearAllMocks();
     });
 
-    it("should return error when agent is restricted by space usage in project conversation with more than one manual member on that project", async () => {
+    it("should succeed when retrying an agent that is restricted by space usage in a project conversation", async () => {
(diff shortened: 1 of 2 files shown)


Fraggle:

- #29843
    - #29844
        - #29845
            - #29846
                - #29847 ◀
                    - #29848
                        - #29851
                            - #29852
                                - #29854

ElPicador (approved):

PR #29850, opened 2026-07-31
Stop exporting symbols that are only used in their own file

davidebbo:
A different angle from the six dead-code PRs: not code nobody uses, but code that is *used* — only inside its own file, while still being exported. **949 declarations across 505 files lose their `export` keyword.** Nothing is deleted, no behaviour changes, `tsgo` proves each one.

Breakdown: 361 types, 277 consts, 181 functions, 156 interfaces, 27 classes. Concentration in `lib/api/actions` (63 files), `lib/api/assistant` (35), `lib/resources` (21), `components/agent_builder` (19), `lib/actions` (16), `components/editor` (16).

## Why verification dominated this one

Knip reports 1,584 such exports in front. **574 of them — 36% — are false positives.** The consumers it cannot see:

- front-spa importing with the file extension in the specifier (`"@dust-tt/front/lib/swr/poke.ts"`)
- extension, marketing and front-api reaching front through the `@app/*` tsconfig alias, which knip does not credit across workspace boundaries

Acting on knip's list directly would have broken four workspaces in ~570 places. Every symbol here was additionally checked with `git grep -w` across all tracked `.ts`/`.tsx`, and front-spa, extension, marketing and front-api all typecheck against the result.

## Left alone deliberately

**`front/lib/llms`, `front/lib/model_constructors`, `types/assistant/models`** (35 unexports reverted). There the export is deliberately ahead of its use: stream endpoint classes such as `DustGoogleGeminiThreeDotFiveFlashEuropeAgentPlatformStream` are written and tested before being registered, so four of those files would have been left exporting nothing at all. Narrowing them would break the next step of the router migration rather than tidy anything. Same reasoning that kept the parked model configs out of #29807.

**16 symbols that knip called in-file-used but biome flags as unused once unexported** — `ParsedContentItem`, `PendingToolCallDetails`, `MetronomeWebhookEventType`, `isJiraProject`, `CheckoutPaymentStatus` and 11 others. Those are dead code, not over-exports; they keep their keyword here and belong in a deletion pass. Worth noting the mechanism: unexporting is what *reveals* them, since biome only reports unused non-exported declarations.

## Test plan

- `npx tsgo --noEmit -p front/tsconfig.json` → clean.
- `npm -w front-spa run tsgo`, `npm -w extension run tsgo`, `npm -w marketing run tsgo`, `npm -w front-api run tsgo` → all clean. These are the checks that matter, given the 36% false-positive rate.
- `npx biome check front` → 0 findings across 5,214 files.
- Every symbol re-verified with `git grep -w` before the keyword was removed.

## Deploy Plan
Change is basically a no-op, so no deployment is strictly necessary.


Fraggle (approved):
Lovely

id13 (approved):

PR #29848, opened 2026-07-31
Add tests for validateAgentMention approve and reject flows.

Fraggle:
## Description

Adds integration tests for `validateAgentMention` covering the full approval lifecycle for restricted agent mentions. Three scenarios: approve (creates agent message child, sets `MentionModel.status` to `"approved"`, launches `launchAgentLoopWorkflow`), reject (status→`"rejected"`, no agent message created, loop not launched), and unauthorized responder (space member who didn't author the parent message gets a 403). The `beforeEach` fixture wires up a project-space conversation with an agent whose `requestedSpaceIds` points to a different space, placing the mention in `agent_restricted_by_space_usage` state before each test case.

## Tests

Added `validate_agent_mention.test.ts` — vitest integration tests using `ConversationFactory`, `AgentConfigurationFactory`, `SpaceFactory`, `UserFactory`, and `MembershipFactory`. `launchAgentLoopWorkflow`, `publishMessageEventsOnMessagePostOrEdit`, and `gracefullyStopAgentLoop` are mocked; `rateLimiter` is spied on to bypass throttling.

## Risk

Test-only. No production code changed. Low risk.

## Deploy Plan

No deploy needed.

PR #29851, opened 2026-07-31
Clear duplicate restricted agent mentions on approve/reject/dismiss.

Fraggle:
## Description

When the same restricted agent accumulated multiple `MentionModel` rows on a message (e.g. due to retries), acting on any one row via approve, reject, or dismiss left the remaining rows in their original state, causing ghost pending-approval cards to persist in the UI.

- `validateAgentMention` (approve): `findOne` → `findAll`; primary row drives `approve_existing_mention`, duplicate rows are all marked `approved` within the same transaction. Published `richMentions` collapse all agent entries into a single approved one.
- `validateAgentMention` (reject): same `findAll` pattern; all rows set to `rejected`; duplicate `rejected` entries deduplicated in `richMentions` via a `seenRejectedAgent` Set before publishing.
- `dismissMention`: `findOne` → `findAll`; all matching rows marked `dismissed: true`. Query now also filters by `status` to avoid touching unrelated rows.
- `MessageItem`: adds a `filter` before `map` on `richMentions` to deduplicate `agent_restricted_by_space_usage` cards by `(status, id)` as a defensive frontend layer.

## Tests

Added a unit test to `validate_agent_mention.test.ts` that seeds a duplicate `MentionModel` row, approves the mention, and asserts: both DB rows reach `approved`, only one agent message child is created, and the published `richMentions` contains exactly one approved entry for the agent (zero still-pending ones).

## Risk

Low. Front-only change — no DB schema or migration. Existing duplicate rows in the DB are now fully resolved on the next approve/reject/dismiss action rather than partially. Rollback is safe: reverting restores the prior single-row behaviour.

## Deploy Plan

Deploy front.

### DIFF
--- a/front/components/assistant/conversation/MessageItem.tsx
+++ b/front/components/assistant/conversation/MessageItem.tsx
@@ -374,44 +374,59 @@ export const MessageItem = React.forwardRef<HTMLDivElement, MessageItemProps>(
           )}
           {data.visibility !== "deleted" &&
             !isCompactionMessage(data) &&
-            data.richMentions.map((mention, index) => {
-              // To please the type checker
-              if (!context.conversation) {
-                return null;
-              }
-
-              // :warning: make sure to use the index in the key, as the mention.id is the userId
-
-              if (
-                mention.status === "pending_conversation_access" ||
-                mention.status === "pending_project_membership" ||
-                mention.status === "agent_restricted_by_space_usage"
-              ) {
+            data.richMentions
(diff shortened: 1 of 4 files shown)


Fraggle:

- #29843
    - #29844
        - #29845
            - #29846
                - #29847
                    - #29848
                        - #29851 ◀
                            - #29852
                                - #29854

ElPicador (approved):

PR #29852, opened 2026-07-31
Clean up optimistic placeholders that break restricted-agent mention updates.

Fraggle:
## Description

When a user mentions a restricted agent, the optimistic `placeholder-agent-message-*` row is never replaced by a real backend message — it stays in the Virtuoso list, inflating `lastMessageRank` on the client. Subsequent `user_message_new` and `agent_message_new` SSE events use rank-based lookups and collide with (or skip) these stale rows, causing mention-status updates to be silently dropped.

- Adds `isPlaceholderMessage()` in `types.ts` to detect placeholder rows by `sId` prefix.
- Calls `findAndDelete` to purge placeholder rows at the target rank before any insert/replace in both `user_message_new` and `agent_message_new` SSE handlers.
- After backend responds for a restricted-agent submit: removes agent placeholders whose `configuration.sId` wasn't returned by the backend, and replaces the user placeholder by `sId` (not rank) to survive any rank drift.
- Falls back to sId-based map in `user_message_new` when the rank slot is already occupied by a real non-user message.
- Extends `hasPendingMentions` to include `agent_restricted_by_space_usage` so the action card auto-scrolls into view.

## Tests

Tested locally by posting messages that mention a restricted agent and verifying:
- No stale placeholder remains after submit.
- A follow-up message from the same conversation lands at the correct rank.
- Mention validation UI scrolls into view for `agent_restricted_by_space_usage`.

## Risk

Low. Front-only, no API or DB changes. The `findAndDelete` calls are no-ops when no placeholder exists, so non-restricted flows are unaffected. Fully rollback-safe.

## Deploy Plan

Deploy front.

### DIFF
--- a/front/components/assistant/conversation/ConversationViewer.tsx
+++ b/front/components/assistant/conversation/ConversationViewer.tsx
@@ -19,6 +19,7 @@ import {
 import { MessageItem } from "@app/components/assistant/conversation/MessageItem";
 import { handlePlanUpdatedEvent } from "@app/components/assistant/conversation/plan_mode/handle_plan_updated";
 import type {
+  AgentMessageWithStreaming,
   ConversationForkNotice,
   VirtuosoMessage,
   VirtuosoMessageListContext,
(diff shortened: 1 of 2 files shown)


Fraggle:

- #29843
    - #29844
        - #29845
            - #29846
                - #29847
                    - #29848
                        - #29851
                            - #29852 ◀
                                - #29854

ElPicador (approved):

PR #29856, opened 2026-07-31
[AGENTS] Fix batch Set model for admins, require editorship to update an agent

pmilliotte:
## Description

Batch "Set model" silently skipped every agent the acting admin was not an editor of, because the editor-group gate in `createAgentConfiguration` checked `canWrite` — which only group members hold — instead of `canAdministrate`, which admins hold and which is what `updateAgentPermissions` and the `/editors` endpoint already check; that gate fired even though the batch carries the existing editor list over unchanged, so it rejected a no-op. Relaxing it also unblocked the builder's `PATCH /agent_configurations/:aId` for non-editor admins, which is not what we want, so that route now requires editorship for everyone: an admin has to add themselves as an editor first — the flow the agent builder already implements through `BuilderEditorGateMessage` and the dedicated `/editors` endpoint.

## Tests

Two regression tests: batch model update on an agent owned by another user succeeds and leaves the editor group untouched; `PATCH /agent_configurations/:aId` by a non-editor admin returns 403 and leaves the instructions unchanged. Both verified to fail without their respective change. Ran the `agent_configurations` route suites (88 tests) and `lib/api/assistant/configuration` (36 tests).

## Risk

Non-editor admins can no longer update an agent through the agent API directly — they must add themselves as an editor first. The builder already gated its save on this, so the UI flow is unchanged; direct API consumers relying on the admin bypass would now get a 403.

## Deploy Plan

- Deploy `main` on `front`

### DIFF
--- a/front-api/routes/w/[wId]/assistant/agent_configurations/[aId]/index.test.ts
+++ b/front-api/routes/w/[wId]/assistant/agent_configurations/[aId]/index.test.ts
@@ -1,7 +1,11 @@
-import { createPendingAgentConfiguration } from "@app/lib/api/assistant/configuration/agent";
+import {
+  createPendingAgentConfiguration,
+  getAgentConfiguration,
+} from "@app/lib/api/assistant/configuration/agent";
 import { AgentConfigurationModel } from "@app/lib/models/agent/agent";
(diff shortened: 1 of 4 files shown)


fabiencelier (commented):
LGTM ✅

PR #29858, opened 2026-07-31
Require a reason for upgrade requests (admin-configurable)

avervaet:
## Description

This PR completes the user-facing flow for the configurable upgrade-request reason requirement introduced in #30314 and enforced by #30315.

Workspace admins can enable **Require a reason for upgrade requests** from the usage settings. The toggle is disabled when upgrade requests are disabled, while the existing read-only and loading protections remain in place.

For members, the upgrade-request form now receives the requirement from the usage-status endpoint, validates a trimmed non-empty reason client-side, and labels the field as required. Invalid request errors from the API are shown inline on the reason field instead of as a generic error toast; other request failures continue to use the existing notification behavior. The setting defaults to disabled, so existing workspaces retain the current optional-reason behavior.

This is the final change in the stack: #30314 → #30315 → #29858. No screenshot is included because testing evidence or a recording was not provided for this draft.

## Risk

The main user-visible change applies only when a workspace enables the setting: members must provide a non-empty reason before submitting an upgrade request. The backend enforcement in #30315 remains the source of truth, so the client-side validation does not weaken the requirement.

The API error handling changes how `invalid_request_error` responses are presented by this form: they are rendered inline, while unrelated failures still show a toast. The setting defaults to `false`, and the admin toggle is unavailable when upgrade requests are disabled. If this UI PR is reverted independently while #30315 remains deployed, the backend will continue enforcing the setting but clients may no longer receive the inline validation experience.

## Deploy Plan

Deploy the stack in order: #30314 first, then #30315, and finally this PR. Run the migration from prodbox before the application relies on the new configuration field. Generate the updated API documentation with `npm -w front-api run docs` and use the **Deploy OpenAPI Docs** GitHub Action to publish the changes to https://docs.dust.tt/reference.

### DIFF
--- a/front/components/workspace/usage/UsageSettingsCard.tsx
+++ b/front/components/workspace/usage/UsageSettingsCard.tsx
@@ -36,48 +36,27 @@ export function UsageSettingsCard({
   const { usageSettings, isUsageSettingsLoading } = useUsageSettings({
     workspaceId,
   });
-  const { doUpdateUsageSettings } = useUpdateUsageSettings({ workspaceId });
+  const { doUpdateUsageSettings, isUpdatingUsageSettings } =
+    useUpdateUsageSettings({ workspaceId });
 
-  const [isSavingAllowUpgradeRequest, setIsSavingAllowUpgradeRequest] =
-    useState(false);
-  const [
-    isSavingRequireUpgradeRequestReason,
-    setIsSavingRequireUpgradeRequestReason,
-  ] = useState(false);
-  const [isSavingAutoSeatUpgrade, setIsSavingAutoSeatUpgrade] = useState(false);
   const [isEditingDefaultLimit, setIsEditingDefaultLimit] = useState(false);
 
   const handleToggleAllowUpgradeRequest = async () => {
-    setIsSavingAllowUpgradeRequest(true);
-    try {
-      await doUpdateUsageSettings({
-        allowUpgradeRequest: !usageSettings.allowUpgradeRequest,
-      });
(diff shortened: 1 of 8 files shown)


avervaet:

- #30314
    - #30315
        - #29858 ◀

front/components/workspace/usage/UsageSettingsCard.tsx
```diff
@@ -54,6 +58,17 @@ export function UsageSettingsCard({
...
   };
 
+  const handleToggleRequireUpgradeRequestReason = async () => {
+    setIsSavingRequireUpgradeRequestReason(true);
+    try {
+      await doUpdateUsageSettings({
+        requireUpgradeRequestReason: !usageSettings.requireUpgradeRequestReason,
+      });
+    } finally {
+      setIsSavingRequireUpgradeRequestReason(false);
```
aubin-tchoi: OOC why isn't the loading state handled by the hook?

front/lib/swr/upgrade_requests.ts
```diff
@@ -21,6 +21,10 @@ function usageStatusUrl(workspaceId: string): string {
   return `/api/w/${workspaceId}/usage-status`;
 }
 
+export type RequestUpgradeResult =
+  | { ok: true }
+  | { ok: false; errorType: string; message: string };
```
aubin-tchoi: let's use a proper Result here

aubin-tchoi (approved):
LGTM

PR #29859, opened 2026-07-31
Backfill user memory

Inesvvv:
## Description

We are implementing per-user personal memory through a single MEMORY.md file per user, shared across all their conversations and agents. Until now, the only form of memory was the agent_memory tool that existed in global agents and could be added to customized agents. 

This script backfills the existing memories in the Dust global agent into the new personal memory so users don't start from an empty MEMORY.md when the feature ships, so whatever the Dust agent already learned about them carries over.

What it does per workspace:
1. Loads all Dust agent memories for the workspace and groups them by user.
2. Formats each user's entries (chronologicaly, one bullet point each) under an ## Imported from Dust agent memory header.
3. Writes it to that user's MEMORY.md via setUserMemory.

Create-only: we only write for users who have no existing MEMORY.md, and skip with a warn anyone who already has one. It defaults to a dry run, pass execute to write, -w <wId> to scope to just one workspace.

## Tests

Testing during the dry run. 

## Risk

Low since we just create from existing memories so nothing is lost. 

## Deploy Plan

No deploy
Run the backfill


front/migrations/20260731_backfill_user_memory.ts
```diff
@@ -0,0 +1,163 @@
...
+    .map((m) => m.content.trim())
+    .filter((c) => c.length > 0);
+
+  if (entries.length === 0) {
+    return "";
+  }
+
+  const bullets = entries.map((e) => `- ${e}`);
+
+  return ["", "", IMPORT_MARKER, "", ...bullets, ""].join("\n");
```
aubin-tchoi: nit: OOC why 2 line breaks at the beginning?

front/migrations/20260731_backfill_user_memory.ts
```diff
@@ -0,0 +1,163 @@
...
+    logger.error(
+      { error: existingResult.error.message },
+      "Failed to read MEMORY.md, skipping"
+    );
+    return;
+  }
+
+  // Create-only, we skip users who already have a memory
+  if (existingResult.value.length > 0) {
+    logger.info("User already has a MEMORY.md, skipping");
```
aubin-tchoi: nit: let's put a warn/error since this is not expected. when we'll dry run (execute: false) the script we'll look for warn and errors to make sure we're good before running it with execute: true

front/migrations/20260731_backfill_user_memory.ts
```diff
@@ -0,0 +1,163 @@
...
+          const userDustMemories = memoriesByUserModelId.get(user.id) ?? [];
+          if (userDustMemories.length === 0) {
+            continue;
+          }
+
+          const userMemory = formatMemories(userDustMemories);
+          await createUserMemory(userAuth, {
+            userMemory,
+            execute,
+            logger: logger.child({
```
aubin-tchoi, line 156: OOC did you know about this or learnt about it here?

aubin-tchoi (approved):
LGTM!

front/migrations/20260731_backfill_user_memory.ts
Inesvvv: For logger.child I asked claude to give me something that could help identify to which workspace/user each log corresponded to. Didn't know logger.child before that.

aubin-tchoi: it's quite handy, there are many situations where this avoids repeating a large number of fields in the log and only focus on the specific part

PR #29861, opened 2026-07-31
Export upgrade-requests history to CSV

avervaet:
## Description

Until now, admins and managers could view resolved upgrade requests in the paginated History tab, but could not download the history. This PR adds a **Download to CSV** action to the History tab in the workspace Usage page.

The UI calls the existing upgrade-requests endpoint with `format=csv`. The API remains manager-only and returns a CSV attachment containing every resolved request, paging internally so exports are not limited by the JSON endpoint's 100-request page size. The export uses the same derived grant, duration, status, reason, and date labels as the rendered table through shared display helpers, preventing the two representations from drifting.

The existing JSON listing, request resolution behavior, and pagination remain unchanged. This change does not add migrations or modify the underlying request data.

## Tests

Added API regression coverage for:

- CSV headers, content-disposition, rendered values, and resolved-request export.
- Authorization: regular users receive `403`.
- Full-history exports containing more than the 100-request JSON page limit.

## Risk

The export is protected by the existing manager authorization middleware, but it intentionally includes the full resolved history, including requester and resolver names/emails, reasons, and outcomes. For workspaces with a large history, the server fetches all rows and builds the CSV before returning it, which may increase request latency and memory usage.

No database schema or data changes are included, and rollback is a code-only revert. Existing JSON consumers are unaffected.

## Deploy Plan

Deploy front. No database migration, backfill, or feature-flag activation is required for this change.

### DIFF
--- a/front-api/routes/w/[wId]/credits/upgrade-requests.ts
+++ b/front-api/routes/w/[wId]/credits/upgrade-requests.ts
@@ -2,10 +2,12 @@ import { getAuditLogContext } from "@app/lib/api/audit/workos_audit";
 import type { ResolveUpgradeRequestError } from "@app/lib/api/credits/upgrade_requests";
 import {
   createUpgradeRequest,
+  listAllResolvedUpgradeRequests,
   listPendingUpgradeRequests,
   listResolvedUpgradeRequests,
   resolveUpgradeRequest,
 } from "@app/lib/api/credits/upgrade_requests";
+import { upgradeRequestsToCsv } from "@app/lib/api/credits/upgrade_requests_export";
 import { UserSpendLimitError } from "@app/lib/api/users/spend_limit";
 import type {
   GetUpgradeRequestsResponseBody,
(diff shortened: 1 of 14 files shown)


front-api/routes/w/[wId]/credits/upgrade-requests.ts
```diff
@@ -38,6 +39,9 @@ const ResolveBodySchema = z.object({
 // "resolved" switches to the admin history view.
 const ListUpgradeRequestsQuerySchema = z.object({
   status: z.union([z.literal("pending"), z.literal("resolved")]).optional(),
+  // Omitted/"json" preserves the existing response body; "csv" streams a
+  // downloadable file instead — used by the History tab's export button.
```
Comment (id13) [inline front-api/routes/w/[wId]/credits/upgrade-requests.ts]: I think this comment is misleading, it's not streaming more than json mode does.

avervaet:

- #29921
    - #29963
        - #29754
            - #29960
                - #29961
                    - #29755
                        - #29861 ◀
                            - #29863
                                - #29913
                                    - #29972
                                        - #30351

front/components/workspace/billing/seatTypeUtils.ts
avervaet, line 1: I moved this to membership utils so it can be shared by multiple component without deduplicating the logic and avoid unintentional drifts. Happy to have input on this

avervaet:

PR #29862, opened 2026-07-31
Fix front typecheck on main

fontanierh:
## Description

`main` does not typecheck. Two unrelated regressions landed today; `tsgo --noEmit` over `front` reports four errors on a clean checkout of `ca7e57068f`.

- `conversation.ts` calls `canAgentBeUsedInProjectConversation` in three places without importing it. Added it to the existing import from `conversation/permissions`.
- `workspace_user.ts` lost the `export` on `getAuthenticatedWorkspaceUser` in #29850, but `sandbox_function_invocation_resource.ts` imports it. Restored the export.

## Tests

`tsgo --noEmit` is clean. Existing permissions and invocation-resource suites pass (56 tests). No behaviour change.

## Risk

None; imports and visibility only.

## Note

Neither regression was caught by CI: the test shards and format-check pass, and no PR job runs a full `front` typecheck. Worth adding one.

### DIFF
--- a/front/lib/api/assistant/conversation.ts
+++ b/front/lib/api/assistant/conversation.ts
@@ -22,7 +22,10 @@ import {
   createUserMessage,
   resolveModelsForMentionedAgents,
 } from "@app/lib/api/assistant/conversation/messages";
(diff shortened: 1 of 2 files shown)


davidebbo (approved):
LGTM

PR #29867, opened 2026-07-31
Document user identity in Pod Function and Frame skills

fontanierh:
## Description

Agents can now require and read the calling user, but neither skill said so. Instructions only; no runtime change.

- **Pod Functions skill** — new `Knowing who called a function` and `Per-user state and sessions` sections: `schema.userIdentity` (`optional` vs `workspace_user_required`) as the gate on whether the function runs, `currentUser()` from `@dust/pod` for who it got, scoping rows by `currentUser().sId`, filtering by owner on every read *and* write, and splitting capabilities across functions instead of branching on an `action` field.
- **Frames skill** — new `useUserIdentity Reference` next to `useFile`: return shape, loading and signed-out handling, and that it personalizes the interface only.

Both state the same rule from their own side: identity is never an input. A `userId` argument names whoever the caller chooses, so the Frame must not send one and the function must not accept one. `useUserIdentity` shapes the UI; `schema.userIdentity` plus `currentUser()` is the enforcement.

`interactive_workspace_user_required` is intentionally left undocumented for now.

## Tests

Global skill suites pass (19). `tsgo --noEmit` and biome clean.

## Risk

None. Skill `version` is unchanged, matching the convention for instruction-only edits.


aubin-tchoi (commented):
LGTM

PR #29868, opened 2026-07-31
Attribute per-tool consumption for settled agent messages

flvndvd:
## Description

Earlier PR taught consumption attribution to record how a settled agent message's cost breaks down across the model token buckets, meaning the prompt it read, the answer it generated, and the reasoning it spent. That left the tool calls invisible, folded anonymously into the output. This change makes each tool call its own line in the breakdown.

For every tool a message invokes, it now measures two footprints with the run model's exact tokenizer (expect some margin errors but it should provide a good representation and we could still call the providers' API), the same path conversation rendering uses to size messages. The first is the output the model spent emitting the call itself, its name and arguments. The second is the input that the tool's result occupied in the following prompt. The result footprint is measured against the exact text the model saw, which is now produced by a single shared routine so the estimate can never drift from what conversation rendering actually sent. Alongside those token footprints, each tool row also carries the tool's own direct credit charge. The tokens the model spent emitting a call are carved out of the assistant output bucket, so the pieces of a run still sum to the whole rather than double counting.

This remains analytics and not billing. The credit amounts are un-rounded and priced cache naive, they explain the relative composition of a cost rather than reproduce the invoice, and the authoritative charge is still computed elsewhere. The work runs once a message reaches a terminal state, launched from the analytics side, and it is idempotent, so a retry or a resumed message re-inserts the same identities as no-ops and only adds what is genuinely new. Because the whole breakdown is computed before anything is written, a tokenization failure leaves the previous state untouched and lets the retry start clean.

### DIFF
--- a/front/lib/api/assistant/agent_message_consumption_attribution/store.test.ts
+++ b/front/lib/api/assistant/agent_message_consumption_attribution/store.test.ts
@@ -1,16 +1,32 @@
 import { AGENT_MESSAGE_CONSUMPTION_ATTRIBUTION_VERSION } from "@app/lib/api/assistant/agent_message_consumption_attribution/attribution_builder";
 import { computeAndStoreAgentMessageConsumptionAttribution } from "@app/lib/api/assistant/agent_message_consumption_attribution/store";
+import { getLlmCredentials } from "@app/lib/api/provider_credentials";
 import { AgentMessageConsumptionItemResource } from "@app/lib/resources/agent_message_consumption_item_resource";
 import { generateRandomModelSId } from "@app/lib/resources/string_ids_server";
+import { tokenCountForTexts } from "@app/lib/tokenization";
 import { AgentConfigurationFactory } from "@app/tests/utils/AgentConfigurationFactory";
+import { AgentMCPActionFactory } from "@app/tests/utils/AgentMCPActionFactory";
 import { ConversationFactory } from "@app/tests/utils/ConversationFactory";
 import { createResourceTest } from "@app/tests/utils/generic_resource_tests";
 import { RunFactory } from "@app/tests/utils/RunFactory";
-import { describe, expect, it } from "vitest";
+import { Ok } from "@app/types/shared/result";
+import { beforeEach, describe, expect, it, vi } from "vitest";
+
+vi.mock("@app/lib/api/provider_credentials", () => ({
+  getLlmCredentials: vi.fn(),
+}));
+
+vi.mock("@app/lib/tokenization", () => ({
+  tokenCountForTexts: vi.fn(),
+}));
 
 const INPUT_TOKENS_COUNT = 100;
 const OUTPUT_TOKENS_COUNT = 20;
 const REASONING_TOKENS_COUNT = 5;
+
+// Every tokenized footprint counts as this many tokens, so tool-call output and result-input
+// footprints are deterministic in the assertions below.
+const TOKENS_PER_FOOTPRINT = 2;
 
 async function setupSettledMessageWithUsage() {
   const { authenticator: auth, workspace } = await createResourceTest({});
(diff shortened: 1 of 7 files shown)


front/lib/api/assistant/agent_message_consumption_attribution/store.test.ts
```diff
@@ -1,17 +1,33 @@
...
+  tokenCountForTexts: vi.fn(),
+}));
 
 const INPUT_TOKENS_COUNT = 100;
 const OUTPUT_TOKENS_COUNT = 20;
 const REASONING_TOKENS_COUNT = 5;
 
+// Every tokenized footprint counts as this many tokens, so tool-call output and result-input
+// footprints are deterministic in the assertions below.
+const TOKENS_PER_FOOTPRINT = 2;
```
flvndvd, line 29: Just for the tests.

front/tests/utils/AgentMCPActionFactory.ts
```diff
@@ -87,6 +90,10 @@ export class AgentMCPActionFactory {
       mcpServerName: "test_server",
     };
 
+    // TODO(Adrien): Drop column if not used anymore.
```
flvndvd, line 93: Vestigial FK which was still populated in tests. @adrsimon from once you are back from PTOs if we can clean things up.

front/lib/api/assistant/conversation_rendering/helpers.ts
```diff
@@ -65,6 +65,59 @@ function compactResourceItem(item: CallToolResult["content"][number]) {
   return resource;
 }
 
+// The fixed notice the model sees for an action that did not run (denied or blocked), or null when
```
flvndvd, line 68: 🚗 Consolidated around one function so we can use the same for conversation rendering and analytics.

front/lib/api/assistant/agent_message_consumption_attribution/store.ts
```diff
@@ -1,31 +1,42 @@
...
 import { RunResource } from "@app/lib/resources/run_resource";
 import logger from "@app/logger/logger";
 import { AGENT_MESSAGE_STATUSES_TO_TRACK } from "@app/types/assistant/conversation";
 import { assertNever } from "@app/types/shared/utils/assert_never";
+import { removeNulls } from "@app/types/shared/utils/general";
+import assert from "assert";
+
+// AWU (the credit unit tool charges are denominated in) expressed in micro-credits, matching how the
+// model token buckets store their attributed credits.
+const CREDIT_AMOUNT_MICRO_PER_CREDIT = 1_000_000;
```
aubin-tchoi, line 22: wait this variable really means: ratio to convert from AWU to micro AWU, right?
then small nit but we could make it a function that also includes the rounding, wdyt?

avervaet (approved):

front/lib/api/assistant/agent_message_consumption_attribution/store.ts
```diff
@@ -77,40 +89,134 @@ export async function computeAndStoreAgentMessageConsumptionAttribution(
...
+            internalMCPServerName: enrichedAction.internalMCPServerName,
+          },
+          triggeringUserMessageOrigin
+        ) * CREDIT_AMOUNT_MICRO_PER_CREDIT
+      );
+
+      const toolAttribution = buildToolAttribution({
+        usage,
+        toolCall,
+        inputTokensCount: footprints[index].resultInputTokensCount,
```
aubin-tchoi: what tells us `footprints[index]` is the one for this specific tool call?

aubin-tchoi (approved):
LGTM

front/lib/api/assistant/agent_message_consumption_attribution/store.ts
flvndvd: Yes, we don't want to do any rounding here, since billing does the rounding at the message level.

front/lib/api/assistant/agent_message_consumption_attribution/store.ts
flvndvd: `footprints` comes back position-aligned with the actions passed from above. Let me make this more robust.

PR #29872, opened 2026-07-31
Store usage type of the runs

fabiencelier:
## Description

Storing all the run usage type (user / programmatic / free) so we can track the actual free consumption of the users.

The new column is nullable for 2 reasons:
 - conversations run stay null for ashort while until after the run when have access to the usage type depending on the origin 
 - we don't backfill the past data (would be very expensive query for no benefit)

## Tests

testd manually + added unit tests

## Risk

low

## Deploy Plan

lock front
merge
pre deploy script
deploy front
unlock front


id13 (approved):
LGTM

PR #29871, opened 2026-07-31
Remove ConversationBranch feature core: model, resource, and branch-scoped logic

Fraggle:
## Description

Removes the `ConversationBranch` feature at the application level while keeping the `branchId` DB column intact. Deletes `ConversationBranchModel`, `ConversationBranchResource`, and all branch-scoped logic across the stack:

- Hardcodes `conversationBranchId: null` in all agent loop call sites and forces `branchId IS NULL` in conversation fetch
- Removes the "in branch → skip restricted agent check" bypass from `permissions.ts`
- Strips branch-specific SSE event allow checks and `branch_anchor` from `UserMessageOrigin`/analytics
- Drops branch-related `DustError` codes (`branch_not_found`, `open`, `write_not_authorized`, `has_no_user_message`)
- Unregisters `ConversationBranchModel` from `db.ts`, `dangerfile.ts`, and `string_ids.ts`

Fork tests are preserved, adjusted to pass `null` branches.

## Tests

Removed branch-specific test suite (`getConversation with branches`, retry-in-branch test). Fork tests updated to expect `conversationBranchId: null`. No new production paths introduced.

## Risk

Low-to-medium. Large surface area, but all removed paths were dead: no branches can be created once `ConversationBranchModel` is unregistered, so the permission bypass and SSE gates were unreachable. The `branchId` DB column is retained — rollback requires only a front redeploy, no migration.

## Deploy Plan

Deploy front.

PR #29870, opened 2026-07-31
Remove conversation branch merge/close API.

Fraggle:
## Description

Dead-code removal: deletes the `branches` route subtree (`GET /branches`, `POST /branches/:bId/merge`, `POST /branches/:bId/close`) from `front-api/routes/w/[wId]/assistant/conversations/[cId]/` and unmounts it from the conversation router. The approval UI no longer calls these endpoints, making `closeConversationBranch`, `mergeConversationBranch`, and `getMostRecentOpenBranchForConversation` unreachable from the API layer.

## Tests

All existing tests for these handlers are deleted alongside the routes. Tested manually that the conversation router still starts correctly with the `/branches` mount removed.

## Risk

Low. Pure deletion of unreachable routes — no logic change, no DB impact. Rollback is safe: revert the commit and redeploy.

## Deploy Plan

Deploy front-api.

PR #29869, opened 2026-07-31
Remove conversation branch approval UI and open-branch hooks.

Fraggle:
## Description

Removes all UI machinery for restricted-agent conversation branches, which are no longer created. Deletes `ConversationBranchApprovalModal` (the review sheet with merge/reject actions), `useConversationBranchActions` (branch merge/close API calls), and `useOpenConversationBranch` (SWR fetch for pending open branches). Strips the supporting wiring from `ConversationViewer`: `branchIdToApprove` state, the `useEffect` that re-injected branch messages on reload, the SSE handler that triggered the approval modal, and the `StickyHeader` mount. `MessageItem` now unconditionally hides any message with a `branchId` (fork notices excepted), and `branchIdToApprove`/`setBranchIdToApprove` are removed from `VirtuosoMessageListContext`.

## Tests

Tested locally — branch messages no longer surface in the conversation UI, and no regressions on standard conversation flows.

## Risk

Low. Pure frontend deletion; the backend branch endpoints remain untouched. Legacy conversations with open branches will simply not show the approval modal, which is the intended outcome. Rollback is a straight revert.

## Deploy Plan

Deploy `front`.

PR #29873, opened 2026-07-31
Remove obsolete branchId args from front-api message routes.

Fraggle:
## Description

Follow-up to the `ConversationBranch` removal: drops leftover `branchId` arguments from the front-api `edit`, `retry`, and `delete` message route handlers. `editUserMessage`, `retryAgentMessage`, and `softDeleteUserMessageAndReplies` no longer accept `branchId`, so these call sites were breaking `tsgo` type-checking. Also removes the `conversation` spread that injected `branchId` into the object passed to `batchRenderMessages` in the `w/` (non-v1) routes.

## Tests

Dead-code removal with no logic change — verified that `tsgo` compilation passes after the cleanup.

## Risk

Low. Removes arguments that are no longer accepted by the callee signatures; no behavior change at runtime since `branchId` was already hardcoded to `null` across the agent loop.

## Deploy Plan

Deploy front-api.

### DIFF
--- a/front-api/routes/v1/w/[wId]/assistant/conversations/[cId]/messages/[mId]/edit.ts
+++ b/front-api/routes/v1/w/[wId]/assistant/conversations/[cId]/messages/[mId]/edit.ts
@@ -141,8 +141,6 @@ app.post(
       });
     }
 
-    const branchId = messageModel.getBranchId() ?? null;
-
     const renderRes = await batchRenderMessages(
       auth,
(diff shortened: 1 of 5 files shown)


Fraggle:

- #29869
    - #29870
        - #29871
            - #29873 ◀

Fraggle:
Folded into #29871 so front-api tsgo fix lands on the core PR CI.

PR #29876, opened 2026-07-31
Tell the model to copy skill names verbatim in the equipped-skills list

fontanierh:
## Problem

`enable_skill` resolves `skillName` by exact, case-sensitive string equality — `skill.name === skillName` in `findAvailableSkillForAgentLoop`, and an exact SQL match in the `fetchByName` fallback. No trim, no case folding, no fuzzy fallback. On a miss it returns a bare `Skill "X" not found` with no candidate list.

Nothing in the prompt said any of that. There was no "verbatim" instruction anywhere in `skills_rendering.ts`, `generation.ts`, `discover_skills`, or the tool description.

The catalogue rendered each entry as `- **Name**: description`. In workspaces that name skills with a `[Category] Title` convention, models stop copying the string and start regenerating it to fit the pattern. Observed in production, all in one workspace where 64 of 124 discoverable names are bracketed:

| asked | actual skill | mutation |
|---|---|---|
| `[Pod Functions]` | `Pod Functions` | wrapped |
| `[Business Case - Company Overview]` | `Business Case - Company Overview` | wrapped |
| `[Support] Support Knowledge Lookup` | `Support Knowledge Lookup` | prefix invented |
| `[CX] Userstone of voice-Mail` | `CX_Userstone of voice-Mail` | separator rewritten |
| `[Team] Team Context` | `[GTM] Team Context` | wrong category |

Each of these skills existed and was in the list the model had been given.

## Change

- Render names as code literals (`` - `Pod Functions`: … ``) rather than bold text. Models copy backticked literals verbatim far more reliably than bolded prose, and it delimits the name from the description — which matters because many descriptions are multi-line.
- Add an explicit instruction in the same `<dust_system>` block, adjacent to the names rather than tens of thousands of characters away in the system prompt, naming the specific mutations rather than a generic "be exact".
- Add "copied verbatim" plus an exact-match note to the `<skill id="..." name="..." />` guidance in `constructSkillsSection`.

## Scope

This is mitigation, not a fix. The model is reconstructing names from an inferred schema, and an instruction competes with that rather than making it impossible. Accepting the skill `sId` — which the toolsets catalogue already exposes as `` (toolsetId: `sId`) `` right next to this one — is what would make the failure mode unreachable. Worth doing separately.

## Test

Updated the `renderEquippedSkillsUserMessage` expectation in `conversation_rendering/helpers.test.ts`. Rendering, generation, run_model, and skill_management suites pass; typecheck clean.

### DIFF
--- a/front/lib/api/assistant/skills_rendering.ts
+++ b/front/lib/api/assistant/skills_rendering.ts
@@ -43,9 +43,9 @@ export function renderEquippedSkillsUserMessage(
 
   const enableSkillToolName = `${SKILL_MANAGEMENT_SERVER_NAME}${TOOL_NAME_SEPARATOR}${ENABLE_SKILL_TOOL_NAME}`;
   // Names are rendered as code literals rather than bold text: `skillName` is matched exactly, and
-  // a literal is copied verbatim far more reliably than prose. Workspaces often name skills with a
-  // `[Category] Title` convention, and models otherwise regenerate names to fit that pattern —
-  // wrapping an unbracketed name, or inventing a category prefix — which never resolves.
+  // a literal is copied verbatim far more reliably than prose. Workspaces tend to develop their own
+  // naming conventions, and models otherwise regenerate names to fit the pattern they infer from
+  // the list instead of copying them, which never resolves.
   const lines = equippedSkills.map(
     ({ name, agentFacingDescription }) =>
       `- \`${name}\`: ${agentFacingDescription.replaceAll("\n", "\n  ")}`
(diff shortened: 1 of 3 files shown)


front/lib/api/assistant/skills_rendering.ts
```diff
@@ -42,15 +42,23 @@ export function renderEquippedSkillsUserMessage(
...
+      `- \`${name}\`: ${agentFacingDescription.replaceAll("\n", "\n  ")}`
   );
 
   return renderSystemSkillMessage(
     `<dust_system>\n` +
       `The following skills are available for use with the ${enableSkillToolName} tool:\n\n` +
-      `${lines.join("\n")}\n` +
+      `${lines.join("\n")}\n\n` +
+      `Pass \`skillName\` exactly as written between backticks above: same case, same ` +
+      `punctuation, and including any leading \`[Category]\` prefix. Do not add brackets, ` +
```
aubin-tchoi: can we make it a bit less specific to the `[Category]` thing? I suspect other ppl will end up adopting different conventions

aubin-tchoi (approved):
LGTM 🙏🏻

front/lib/api/assistant/skills_rendering.ts
fontanierh: Yes agreed this slipped through

fontanierh: slop prompting

PR #29878, opened 2026-07-31
feat: enforce per-user credit limit on any plans

sfriquet:
## Description

- Add a way to enforce per-user credit limit even on non CP plans.
- Reuse the existing logic for CP plan leveraging redis/ES backed rate limiter, but force period to current calendar month (CP plans rely on metronome's contract which is not available on non CP plans)
- Add a poke plugin to set the limit on legacy workspace plan
- Open the poke plugin that syncs counters to all workspaces

## Tests

1. set a workspace limit (`workspaces.defaultPoolCapAwuCredits`)
2. enable `enforce_user_spend_limit_rate_cap`
3. sync counters through plugin `Run Resync Spend-Limit Counters from Usage`
4. if cap is reached, see an error

Screenshot 2026-07-31 at 18 13 13
Screenshot 2026-07-31 at 18 13 52

## Risk

Under feature flag. Safe to revert.

## Deploy Plan

Front only

### DIFF
--- a/front/lib/api/credits/members_usage.ts
+++ b/front/lib/api/credits/members_usage.ts
@@ -4,6 +4,7 @@ import {
 } from "@app/lib/api/assistant/rate_limits";
 import { bucketsToArray, searchAnalytics } from "@app/lib/api/elasticsearch";
 import type { Authenticator } from "@app/lib/auth";
+import type { BillingCycle } from "@app/lib/client/subscription";
 import { listPerUserCreditBalanceAlertsForWorkspace } from "@app/lib/metronome/alerts/per_user_credit_balance";
 import type {
   MetronomeCapAlertIds,
@@ -45,6 +46,7 @@ import { CreditUsageConfigurationResource } from "@app/lib/resources/credit_usag
 import { GroupResource } from "@app/lib/resources/group_resource";
 import { MembershipResource } from "@app/lib/resources/membership_resource";
 import { UserResource } from "@app/lib/resources/user_resource";
+import { spendLimitCycleOverrideForAuth } from "@app/lib/spend_limits/cycle";
 import type { EffectiveSpendLimitSource } from "@app/lib/spend_limits/effective";
 import {
   resolveEffectiveSpendLimitAwuCredits,
@@ -240,6 +242,24 @@ async function fetchCreditsResetAt(
   return periodResult.value?.cycleEnd.toISOString() ?? null;
 }
 
+// The workspace's current Metronome contract billing period, or null when it
+// cannot be resolved (no contract, or a Metronome failure).
+async function resolveMetronomeCycle(
+  workspace: LightWorkspaceType
+): Promise<BillingCycle | null> {
+  const periodResult = await getCachedMetronomeCurrentBillingPeriod(
+    workspace.sId
+  );
+  if (periodResult.isErr()) {
+    logger.warn(
+      { err: periodResult.error, workspaceId: workspace.sId },
+      "[MembersUsage] Failed to resolve billing period"
+    );
+    return null;
+  }
+  return periodResult.value;
+}
+
 // Per-user consumed AWU credits for the current billing cycle, summed from the
 // analytics index (`cost.billable_awu`, precomputed at index time) by Elasticsearch.
 // This replaces the per-user Metronome usage scan that previously dominated the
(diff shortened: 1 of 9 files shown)


front/lib/api/credits/members_usage.ts
```diff
@@ -955,32 +975,88 @@ export async function getEffectiveSpendCapAwuCreditsForUser(
...
+      workspace,
+      userModelIds: [user.id],
+    }),
+  ]);
+
+  // "none" seat users have no pool access, so their override is irrelevant.
+  const overrideAwuCredits =
+    membership.seatType !== "none"
+      ? membership.poolCapOverrideAwuCredits
+      : null;
```
fabiencelier: there is no seat type for non credit based plan

front/lib/api/credits/members_usage.ts
```diff
@@ -955,32 +975,88 @@ export async function getEffectiveSpendCapAwuCreditsForUser(
...
+
+  const defaultPoolCapAwuCredits =
+    creditUsageConfig?.defaultPoolCapAwuCredits ?? 0;
+
+  return resolveEffectiveSpendLimitAwuCredits({
+    overrideAwuCredits,
+    groupCapAwuCredits: groupCapByUserModelId.get(user.id) ?? null,
+    defaultAwuCredits:
+      defaultPoolCapAwuCredits > 0 ? defaultPoolCapAwuCredits : null,
+  });
```
fabiencelier: Can we actually set override per group for this ?

fabiencelier (approved):
Looks good !
Nice that it is possible without using metronome 

Something additional that we may want to do is add the info somewhere that they have reached there fair use limit for the month

front/lib/api/credits/members_usage.ts
sfriquet: looks like not, simplifying.

front/lib/api/credits/members_usage.ts
sfriquet: simplified, thx!

front/lib/api/poke/plugins/workspaces/set_default_user_credit_limit.ts
```diff
@@ -0,0 +1,118 @@
...
+  // Credit-priced workspaces have their own default (seat allowance + pool limit,
+  // mirrored to Metronome alerts): they set it through
+  // `manage-credit-usage-configuration` instead.
+  isApplicableTo: (auth) => {
+    const plan = auth.plan();
+    return plan === null || !isCreditPricedPlan(plan);
+  },
+  populateAsyncArgs: async (auth) => {
+    return new Ok({
+      awuCredits: await getNonCreditPricedDefaultUserSpendLimit(auth),
```
flvndvd, line 67: `nit` we tend to avoid nested async ops in return results. When using debugger it's often more convenient to keep async operations outside of return or condition statements.

front/lib/api/poke/plugins/workspaces/set_default_user_credit_limit.ts
```diff
@@ -0,0 +1,118 @@
...
+    }
+
+    const validationResult =
+      SetDefaultUserCreditLimitArgsSchema.safeParse(args);
+    if (!validationResult.success) {
+      return new Err(new Error(fromError(validationResult.error).toString()));
+    }
+    const { awuCredits } = validationResult.data;
+
+    const result = await setNonCreditPricedDefaultUserSpendLimit(auth, {
```
flvndvd, line 82: Any reason to set it even though the FF is not enabled?

front/lib/api/workspace/default_user_spend_limit.ts
```diff
@@ -356,3 +356,92 @@ export async function setDefaultUserSpendLimit(
...
+  ) {
+    return new Err(
+      new DefaultUserSpendLimitError(
+        "invalid_threshold",
+        `awuCredits must be an integer between ${MIN_DEFAULT_USER_SPEND_LIMIT_AWU_CREDITS} and ${MAX_DEFAULT_USER_SPEND_LIMIT_AWU_CREDITS}.`
+      )
+    );
+  }
+
+  const workspace = auth.getNonNullableWorkspace();
```
flvndvd, line 396: Can we add an assert to ensure that current company's plan is not credit-based?

front/lib/api/workspace/default_user_spend_limit.ts
```diff
@@ -356,3 +356,92 @@ export async function setDefaultUserSpendLimit(
...
+ * Returns the workspace per-user credit limit for a workspace that is not
+ * on a credit-priced plan.
+ * In practice it returns the same value as `getDefaultUserSpendLimit`,
+ * but this doesn't require the workspace to be on a credit-priced plan, and the returned
+ * value should be interpreted differently: 0 means "no limit" here.
+ */
+export async function getNonCreditPricedDefaultUserSpendLimit(
+  auth: Authenticator
+): Promise<number> {
+  const config =
```
flvndvd, line 444: Same can we add an assert here to confirm that it only run on none credit-based pricing?

flvndvd (commented):
LGTM

PR #29879, opened 2026-07-31
Add rate limiter for sidekick usage

fabiencelier:
## Description

Add 2 new checks for sidekick origin usage:
 - a rate limiter to 50 calls per 24h
 - no API key, only real user must be calling sidekick

## Tests

tested locally

## Risk

low, we may block some legitimate usage but 50 sidekick messages per day is already a lot.

## Deploy Plan

deploy front


zmarouf (approved):
LGTM but 50 might be on the low side if you're doing a full Build day. No?

PR #29883, opened 2026-07-31
ttribute tool call output from the emitted arguments, not the augmented params

flvndvd:
## Description

Per-tool consumption attribution measures how many tokens the model spent emitting each tool call so it can attribute that share of the run's output. It did this by tokenizing the action's params, which are the arguments after Dust augments them with preconfigured values and injected secrets. Those extra inputs were never generated by the model, so counting them overstated the attributed output for that call, and because that count also feeds the step that carves a call's emission out of the assistant output bucket, it distorted the surrounding model rows too.

This change tokenizes what the model actually emitted instead. The raw arguments string the model produced is read directly from the action resource on the server, through a small accessor next to the one that already exposes the call name, and the call footprint is built from that. It is deliberately not added to the serialized action type. That type is sent to the browser, and the augmented params already travel there, so introducing a second redundant field would only grow the client payload for something only the backend attribution path consumes. Keeping it server side also avoids widening a client contract for an internal need.

Left a TODO to improve the resource logic here.

### DIFF
--- a/front/lib/api/assistant/agent_message_consumption_attribution/store.ts
+++ b/front/lib/api/assistant/agent_message_consumption_attribution/store.ts
@@ -137,7 +137,12 @@ export async function computeAndStoreAgentMessageConsumptionAttribution(
     // the input its result occupied. Everything is priced against this one usage below.
     const footprintsRes = await measureToolCallFootprints(auth, {
       modelId: usage.modelId,
-      actions: enrichedRunActions,
+      // TODO(2026-07-31 FLAV) Refactor `enrichActionsWithOutputItems` so it still returns the
+      // resource.
+      toolCalls: runActions.map((action, index) => ({
+        action: enrichedRunActions[index],
+        functionCallArguments: action.functionCallArguments,
+      })),
     });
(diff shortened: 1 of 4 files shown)


front/lib/api/assistant/agent_message_consumption_attribution/store.ts
```diff
@@ -137,7 +137,12 @@ export async function computeAndStoreAgentMessageConsumptionAttribution(
     // the input its result occupied. Everything is priced against this one usage below.
     const footprintsRes = await measureToolCallFootprints(auth, {
       modelId: usage.modelId,
-      actions: enrichedRunActions,
+      // TODO(2026-07-31 FLAV) Refactor `enrichActionsWithOutputItems` so it still returns the
+      // resource.
+      toolCalls: runActions.map((action, index) => ({
```
flvndvd, line 142: Not very elegant to rely on order from previous operations, but will send a follow up PR that touches many files to have a better API over this resource.

aubin-tchoi (approved):
LGTM

PR #29885, opened 2026-08-01
[LLM] Add DeepSeek V4 Flash 0731 on Fireworks

pmilliotte:
## Description

Adds DeepSeek V4 Flash 0731 on Fireworks (256k usable context, low/high/max reasoning, $0.14/$0.28 per 1M tokens).

Commits are split one-per-layer so this doubles as a worked example of adding a model. Full details in the `dust-llm` skill; the short version:

### Which layer owns what

| Layer | Owns | Rule |
|---|---|---|
| `types/assistant/models/*` | Legacy `ModelConfigurationType` — what the UI, pricing and tiers read | Registering the id in `STATIC_MODEL_IDS` type-forces the pricing/tiers/reasoning trio |
| `lib/model_constructors/*` | Endpoint class per `(provider, model, region, provider-api)` — request/response shape, input schema | **Mirrors the provider, nothing else.** Real context and max-output values, and a schema that accepts exactly what the API accepts |
| `lib/llms/*` | Thin Dust wrapper — display name, `byok`, caps, `configParsers` | **Where product direction lives.** Every deliberate divergence from the provider belongs here, not in the schema |

That split is the one thing to get right. Here the endpoint class carries the real 1,040k context / 384k output, and the Dust layer caps both — 256k context, 64k output. The cap is deliberate product policy, not a provider limit: very long windows degrade quality, so we expose 256k (matching Kimi K3, the other Fireworks 1,040k model), leaving a 192k prompt budget after the generation reserve. Note what this buys — the endpoint class stays a truthful record of the provider, so revisiting the cap is a one-line change in one layer. DeepSeek has no `medium` effort, so the Dust layer folds our ladder onto low/high/max with `mapReasoningEffortToLowHighMax` — the schema is never widened to swallow an effort the model doesn't have.

### Reasoning efforts follow the model author, not the host

Fireworks' gateway validates `reasoning_effort` generically for every model it serves, so `medium` and `xhigh` stream fine here — but DeepSeek doesn't define them for V4 Flash. Expose the documented set only and comment the exclusion as a policy choice so nobody later mistakes it for a provider limit. Same for the default effort: read it off the official doc rather than copying the sibling model.

### Add and test every available region and host

One endpoint class per `(region, provider-api)` the model actually serves, each with its own live test setup — a config mixin is shared, so narrowing it moves every endpoint using it and they all need re-running. Fireworks is a single global US-only host, so this model has exactly one endpoint; a Gemini or Anthropic model would need the EU and agent-platform variants added and tested too.

### The loop

Widen the schema to the provider-wide one → write the test with every case `null` → run it live, no `--bail` → sort each failure into "API rejected it" (schema must encode it) vs "our zod rejected it" vs "it passed" (policy call) → narrow → green. Don't copy the sibling's expected values; that hides exactly the differences you're looking for.

## Tests

Live endpoint suite green, 46/46 (`deepseek_deepseek_v4_flash_0731_global_fireworks.test.ts`), covering none/low/high/max × temperature 0/0.1/1, forced tool use with reasoning, JSON-schema output and prompt caching. `sdk_drift`, `types` and `tiers` green; `tsgo` clean after each individual commit.

Characterization run confirmed `minimal` is a real Fireworks 400, and that `medium`/`xhigh` are accepted by the gateway but excluded by policy.

## Risk

Low — new model, no existing behavior touched. It is not feature-flagged, so it appears in the model picker for all workspaces on deploy.

## Deploy Plan

- Deploy `main` on `front`


front/lib/model_constructors/providers/fireworks/models/deepseek_v4_flash_0731.ts
```diff
@@ -0,0 +1,53 @@
+import { fireworksConfigSchema } from "@app/lib/model_constructors/providers/fireworks/inputConfig";
+import { DEEPSEEK_V4_FLASH_0731 } from "@app/lib/model_constructors/types/models";
+import { z } from "zod";
+
+// Real model spec, verified 2026-08-01 against
+// https://fireworks.ai/models/deepseek-ai/deepseek-v4-flash-0731 (1040k
+// context) and https://api-docs.deepseek.com/quick_start/pricing (384k max
+// output). The Dust product cap (64k output) lives in the llms layer; the
+// full context is exposed as-is.
+const CONTEXT_SIZE = 1_040_000;
```
fontanierh, line 7: You sure we want to go for full context? On other models we're capping because 1M is not really exploitable (esp on smaller models) and it gets pretty expensive

fontanierh (approved):

front/lib/model_constructors/providers/fireworks/models/deepseek_v4_flash_0731.ts
pmilliotte: Good looks, decreased to 256k

PR #29886, opened 2026-08-01
Write blocked tool attribution rows as pending and complete them once the action settles

flvndvd:
## Description

Attribution runs on every finalize, including while the loop is paused on a tool awaiting approval. Back then the action has not run, so it has no result and no charge, yet the first implementation wrote a fully completed row for it with a charge and the approval notice as the result, and the idempotency key froze that row so approval could never correct it. Billing dodges this by recomputing and overwriting the whole total each finalize, but attribution only appends, so a blocked tool has to be recorded pending and completed later.

This also required fixing how recomputations are scheduled. The existing launcher used a fixed workflow ID with a plain start. A finalize arriving while the workflow was already running was rejected as already started, potentially leaving a pending row without a later pass to complete it. New launches now use `signalWithStart` against a stable per-message V2 workflow. A signal received while attribution is being computed requests another pass, while signals received before computation are covered because the activity reads the latest message state from PostgreSQL.

The scheduling change uses a new workflow type, signal name, and workflow ID suffix. The original workflow remains exported with its original implementation so currently running executions can replay safely. New tasks use V2 from their first finalize.
For a task already running during deployment, its next finalize starts or signals V2, which reconstructs the complete attribution from PostgreSQL without requiring state from the earlier workflow execution.

A blocked tool row contains only the normalized output tokens and gross model contribution of the emitted call. Once every tool awaiting approval has been approved and the agent loop resumes, the next finalize completes that same row with the result footprint, direct charge, and completion timestamp. Auto-approved tools are written completed on their first pass.

The resource writes each pass atomically. Model facts and new tool facts use bulk inserts with first-write-wins semantics. When a final tool conflicts with an existing identity, the resource reads only conflicting rows that remain
pending and completes them with a second `completedAt IS NULL` guard. Completed facts cannot be overwritten, and a delayed blocked pass cannot regress a completed tool to pending, so overlapping and out-of-order passes converge on the same result.

> [!NOTE]
> PostgreSQL may consume primary-key sequence values for rows rejected by `ON CONFLICT DO NOTHING`. We accept those gaps to preserve race-safe writes without adding a pre-read or locking protocol. Worth noting that the loop resumes only when all pending actions are approved. Will monitor how often it happens, and I might revisit.

### DIFF
--- a/front/lib/resources/agent_message_consumption_item_resource.ts
+++ b/front/lib/resources/agent_message_consumption_item_resource.ts
@@ -5,6 +5,7 @@ import { BaseResource } from "@app/lib/resources/base_resource";
 import type { ConversationResource } from "@app/lib/resources/conversation_resource";
 import type { ReadonlyAttributesType } from "@app/lib/resources/storage/types";
 import type { ModelStaticWorkspaceAware } from "@app/lib/resources/storage/wrappers/workspace_models";
+import { withTransaction } from "@app/lib/utils/sql_utils";
 import type { ModelId } from "@app/types/shared/model_id";
 import type { Result } from "@app/types/shared/result";
 import { Err } from "@app/types/shared/result";
@@ -25,13 +26,6 @@ export type CompletedToolConsumptionItem = ConsumptionItemEvidenceBase & {
   inputTokensCount: number | null;
   /** Estimated tokens in the model output that emitted the tool name and arguments */
   outputTokensCount: number | null;
-  directCreditAmountMicro: number | null;
-};
-
-export type PendingToolConsumptionCompletion = ConsumptionItemEvidenceBase & {
-  action: AgentMCPActionResource;
-  /** Estimated tokens in the result returned by this tool execution */
-  inputTokensCount: number | null;
   directCreditAmountMicro: number | null;
 };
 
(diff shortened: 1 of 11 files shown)


front/temporal/analytics_queue/worker.ts
```diff
@@ -36,6 +36,18 @@ export async function runAnalyticsWorker() {
         },
       ],
     },
+    bundlerOptions: {
```
flvndvd, line 39: 🚗 Added it so we can use alias imports in the workflow code.

front/temporal/analytics_queue/workflows.ts
```diff
@@ -46,6 +47,7 @@ export async function storeAgentMessageFeedbackWorkflow(
   });
 }
 
+// Kept unchanged for executions started before the V2 signal workflow was deployed.
 export async function storeAgentMessageConsumptionAttributionWorkflow(
```
flvndvd, line 51: Will drop previous workflow code, once the queue has been drained.

aubin-tchoi (approved):
LGTM

avervaet (approved):

front/lib/resources/agent_message_consumption_item_resource.ts
```diff
@@ -224,92 +250,185 @@ export class AgentMessageConsumptionItemResource extends BaseResource<AgentMessa
...
+            workspaceId: auth.getNonNullableWorkspace().id,
+            agentMessageId: agentMessageModelId,
+            attributionVersion,
+            itemType: "tool",
+            completedAt: { [Op.is]: null },
+          },
+          transaction,
+        }
+      );
+    }
```
avervaet, line 316: nit: this loop does one `UPDATE` per pending row (N+1, [GEN14]). Worth batching into a single statement if tool-call fan-out per message grows — not blocking given today's small N.

PR #29887, opened 2026-08-01
Fix display of the agent message covering user message

fabiencelier:
### Before

image

### After

Screenshot 2026-08-01 at 13 20 43

## Tests

tested locally

## Risk

easy to revert, purely UI

## Deploy Plan

deploy front


fabiencelier:
Failing tests were already failing before that

Fraggle (commented):
I DID test so weird to me that it does that but fully trusting you :)

PR #29888, opened 2026-08-01
[front] fix: blocked-action cards on shared Frames

davidebbo:
## Problem

Opening a shared Frame that calls a tool needing a personal connection (e.g. `get_worksheet` on Google Drive) failed instead of prompting to connect.

The `tool_personal_auth_required` event was delivered correctly, but `SandboxFunctionPersonalAuthCard` → `PersonalAuthenticationCard` calls `useAuth()`, which throws outside an `AuthProvider`. The public share bundle (`front-spa/src/share/ShareApp.tsx`) has no `AuthContext.Provider` — it only wraps routes in `RegionProvider` / `FetcherProvider` / `PostHogTracker` / `ErrorBoundary` — so the card threw on render and the error boundary replaced the whole page. `SandboxFunctionToolApprovalCard` / `ToolValidationCard` had the identical crash.

A second issue surfaced once the card rendered: a Frame fires several invocations in flight, each blocked on the same connection, and each mounted its own overlay. Connecting resolved only the first; the next card appeared on return from the OAuth flow, and any card not clicked through left its invocation blocked until the sandbox token expired and the invocation errored.

## Change

- `PersonalAuthenticationCard` / `ToolValidationCard`: drop `useAuth()`, take a `currentUser` prop. The two agent-loop callers pass it from their own `useAuth()`, unchanged behavior in the app.
- `SandboxFunctionPersonalAuthCard` / `SandboxFunctionToolApprovalCard`: take a `viewer` (`{ owner, user }`) prop.
- `VisualizationActionIframe` takes `viewer`; `AuthenticatedVisualizationActionIframe` derives it from `AuthContext`, `PublicFrameRenderer` from `useUser()` plus the workspace the viewer is a member of. Null for anonymous viewers, who cannot invoke functions and so never reach a blocked action.
- The blocked-action queue moves from `SandboxFunctionInvocation` (one per invocation) up to `VisualizationActionIframe`, which sees every in-flight invocation. `SandboxFunctionInvocation` is now stream-only and renders nothing. One overlay is shown, and all entries waiting on the same `mcpServerId` are resolved together by a single card. Approvals stay per tool call.
- Settling an invocation drops any card it was still waiting on, so a timed-out invocation cannot leave a stale card behind.

## Tests

- New `SandboxFunctionPersonalAuthCard.test.tsx`: renders with the real `AuthContext` module and no provider (fails with `useAuth must be used within AuthProvider` before this change), and asserts one Connect click resolves every invocation in the group.
- `PublicFrameRenderer.test.ts` extended to assert the `viewer` prop is populated for a member and null otherwise.

## Not addressed

Sandbox function invocation tokens expire after 2 minutes (`generateSandboxFunctionInvocationToken` default). While an action sits in `blocked_authentication_required` the sandbox keeps polling and gets a 401 once the token dies, erroring the invocation. A slow OAuth round trip can still lose the invocation — needs its own fix.

### DIFF
--- a/front/components/assistant/conversation/actions/VisualizationActionIframe.tsx
+++ b/front/components/assistant/conversation/actions/VisualizationActionIframe.tsx
@@ -309,7 +309,12 @@ function SandboxFunctionBlockedAction({
       blockedActionCard = (
         <SandboxFunctionPersonalAuthCard
           key={eventId}
-          events={blockedActions.map((a) => a.event).filter(isPersonalAuth)}
+          events={blockedActions
+            .map((a) => a.event)
+            .filter(
+              (e): e is SandboxFunctionToolPersonalAuthRequiredEvent =>
+                e.type === "tool_personal_auth_required"
+            )}
           viewer={viewer}
           onResolved={onResolved}
         />
@@ -327,12 +332,6 @@ function SandboxFunctionBlockedAction({
       </div>
     </div>
   );
-}
-
-function isPersonalAuth(
-  event: SandboxFunctionBlockingEvent
-): event is SandboxFunctionToolPersonalAuthRequiredEvent {
-  return event.type === "tool_personal_auth_required";
 }
 
 // The blocked actions the next card presents: the oldest one, plus everything waiting on the same
--- a/front/components/assistant/conversation/actions/VisualizationActionIframe.tsx
+++ b/front/components/assistant/conversation/actions/VisualizationActionIframe.tsx
@@ -34,6 +34,7 @@ import {
   assertNeverAndIgnore,
 } from "@app/types/shared/utils/assert_never";
 import { normalizeError } from "@app/types/shared/utils/error_utils";
+import type { LightWorkspaceType, UserType } from "@app/types/user";
 import {
   AlertCircle,
   Button,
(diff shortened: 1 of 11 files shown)


front/components/assistant/conversation/actions/VisualizationActionIframe.tsx
```diff
@@ -327,6 +329,33 @@ function SandboxFunctionInvocation({
   );
 }
 
+function isPersonalAuth(
```
fontanierh: We don't already have this somewhere ? Any chance we can make it more type safe ?

front/components/assistant/conversation/actions/VisualizationActionIframe.tsx
```diff
@@ -327,6 +329,33 @@ function SandboxFunctionInvocation({
...
+// are per tool call, so they are never grouped.
+function nextBlockedActionGroup(
+  blockedActions: QueuedBlockingEvent[]
+): QueuedBlockingEvent[] | null {
+  const [blockedAction] = blockedActions;
+  if (!blockedAction) {
+    return null;
+  }
+  if (blockedAction.event.type !== "tool_personal_auth_required") {
+    return [blockedAction];
```
fontanierh, line 348: if we're going to create a function to check this we might as well use it. That said, I'm also fine not having a function

fontanierh (approved):
Overall aligned with the fix

front/components/assistant/conversation/actions/VisualizationActionIframe.tsx
davidebbo: I just yanked the function and inlined it.

front/components/assistant/conversation/actions/VisualizationActionIframe.tsx
davidebbo: I just yanked the function and inlined it.

PR #29254, opened 2026-07-21
[poke] make pod function visible in Poke

TimotheeCrouzet:
## Description

Make the sId, slug, description, date of creation and author of the pod functions visible in poke.

## Tests

Verified manually that the Pod Functions section is present in poke.
image

## Risk

Low. Safe to revert.

PR #29891, opened 2026-08-02
[front] BACK15: serialized shapes belong to the Resource

davidebbo:
## Description

Clarifies `[BACK15]` in `front/CODING_RULES.md`.

The rule required a Resource to have a `toJSON` method, but said nothing about who may build the shapes a Resource is serialized into. Under that reading, `{ ...resource.toJSON(), extraField }` in `lib/api/*` looks compliant — the Resource *does* have a `toJSON`, the caller just extends it. That is how a poke serialization ended up assembled in the business layer in a recent branch, and it is the same thing @flvndvd asked for on #29254 ("Resources have a `toJSON` function, they should have a few well identified ways to be serialized. This responsibility is also part of the resource itself. Let's add a `toPokeJSON` on the resource if needed.").

The added paragraph states that:

- a Resource has a small, named set of serializations — `toJSON` plus audience-specific `to<Audience>JSON` variants, documenting the convention already in the codebase (`toPokeJSON`, `toLogJSON`, `toJSONForLLM`);
- callers pick one rather than assembling their own, and specifically must not spread `toJSON()` and add fields;
- a caller needing a shape no method produces adds a method to the Resource.

Documentation only — no code changes.

## Tests

N/A.

## Risk

None. Documentation only.

## Deploy Plan

Nothing to deploy.

### DIFF
--- a/front/CODING_RULES.md
+++ b/front/CODING_RULES.md
@@ -266,6 +266,14 @@ endpoint handlers.
 
 The Type interface is not to be used in the backend.
 
+**Every serialized shape belongs to the Resource, not to its callers.** A Resource has a small,
+named set of ways to be serialized — `toJSON`, and audience-specific variants following the
+`to<Audience>JSON` convention already in use (`toPokeJSON`, `toLogJSON`, `toJSONForLLM`). Callers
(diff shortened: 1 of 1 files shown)


flvndvd (commented):
LGTM

PR #29893, opened 2026-08-02
Stop charging tool credits for denied tool calls

flvndvd:
## Description

We charge credits per tool invocation, but the billing pipeline decided what counted as an invocation by asking whether the action had reached a terminal state. Denied actions are terminal and were therefore charged the full weight of their cost category, one credit for basic tools and three for advanced ones, even though the tool never ran.

An action becomes denied in three ways, none of which involve running anything:
- the user rejects the approval prompt
- the user declines the personal authentication
- the message ends on an unresumable status while calls are still sitting blocked, at which point every blocked call on that message is swept to denied at once

This PR introduces a dedicated helper (which surface that this code did not age well, given all the code path required to be updated to use it). 

> [!NOTE]
> It does not change previous bill, only changes moving forward or if ES were to be backfilled again

### DIFF
--- a/front/lib/actions/statuses.ts
+++ b/front/lib/actions/statuses.ts
@@ -53,6 +53,27 @@ export function isToolExecutionStatusFinal(
   );
 }
 
+// The final statuses the tool actually ran under. `denied` is final but never executed: the user
+// rejected the approval, declined the authentication, or ended the message while the call sat
+// blocked. Charges are per invocation, so none of those are charged, while the model tokens spent
+// emitting the call are still billed as intelligence through the run usage. `errored` stays
+// billable, since the tool was invoked and failed on its own terms.
+const TOOL_EXECUTION_BILLABLE_STATUSES = [
+  "succeeded",
+  "errored",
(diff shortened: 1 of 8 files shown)


front/lib/actions/statuses.ts
```diff
@@ -53,6 +53,27 @@ export function isToolExecutionStatusFinal(
   );
 }
 
+// The final statuses the tool actually ran under. `denied` is final but never executed: the user
+// rejected the approval, declined the authentication, or ended the message while the call sat
+// blocked. Charges are per invocation, so none of those are charged, while the model tokens spent
+// emitting the call are still billed as intelligence through the run usage. `errored` stays
+// billable, since the tool was invoked and failed on its own terms.
+const TOOL_EXECUTION_BILLABLE_STATUSES = [
```
flvndvd, line 61: I will consolidate all call sites so we have one place where we compute the credit cost of a message, it should not be scattered everywhere.

aubin-tchoi (approved):
LGTM

PR #29894, opened 2026-08-02
Expose agent message consumption details

flvndvd:
## Description

This PR adds a feature-flagged private endpoint for reading the credit consumption of one agent message. It returns `AgentMessage.costCredits` as the authoritative billed value and, when materialization is complete, a cache-naive attribution grouped into agent work and recognizable tool activity.

The reader validates attribution coverage against the message’s model usages and tool actions before returning details. Runs and actions are used only as internal source facts and are never exposed by the endpoint. If attribution is missing, incomplete, or inconsistent, the response still returns the exact billed credits with attribution details set to null.

Tool executions are grouped by tool identity and retain their display label, call count, attributed credits, direct credits, and completion state. The response contains no provider usage, token counts, tool arguments, tool results, run records, or
action records.

UI is coming in a follow up PR.

### DIFF
--- a/front-api/vite.setup.ts
+++ b/front-api/vite.setup.ts
@@ -4,6 +4,14 @@ import { vi } from "vitest";
 
 import "../front/vite.setup.ts";
 
+// Front-api tests exercise server behavior and do not enable browser development mode.
+vi.mock("@app/components/dev/devModeConstants", () => {
+  return {
+    DEV_MODE_ACTIVE: false,
+    DEV_MODE_STORAGE_KEY: "dust_dev_mode",
+  };
+});
+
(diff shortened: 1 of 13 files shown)


front-api/vite.setup.ts
```diff
@@ -4,6 +4,14 @@ import { vi } from "vitest";
 
 import "../front/vite.setup.ts";
 
+// Front-api tests exercise server behavior and do not enable browser development mode.
```
flvndvd, line 7: Moved here because each tests eventually did it.

flvndvd:
PMRR.

aubin-tchoi (commented):
LGTM

PR #29897, opened 2026-08-02
[User Management] Add mapped identity merge script

philipperolet:
## Description

Related to https://github.com/dust-tt/tasks/issues/9202

Adds an operator-run script that moves data from a mapped old identity into the newly provisioned primary identity. It accepts the issue JSON mapping, validates the exact old and primary emails encoded in each mapping, supports a single-user canary or an explicit full batch, and invokes the existing identity merge plugin.

The preflight blocks ambiguous or missing identities, reversed mappings, primaries without a WorkOS identity, role or seat-type mismatches, trigger deletion, scheduled membership changes, cross-mapping cycles, and last-admin revocation. Execution re-runs preflight immediately before each identity and verifies revocation before proceeding to the next one.

## Risks

Blast radius: Manually selected user identities in one workspace when the script is run with `--execute`.
Risk: standard

## Deploy Plan

- Deploy Front so the script is available in the production image.
- In the EU prodbox, copy the mapping JSON and run a single-user dry-run.
- see https://github.com/dust-tt/tasks/issues/9202#issuecomment-5163898906

### DIFF
--- a/front/scripts/merge_mapped_identities.ts
+++ b/front/scripts/merge_mapped_identities.ts
@@ -1,0 +1,298 @@
+import { userIdentityMergePlugin } from "@app/lib/api/poke/plugins/workspaces/user_identity_merge";
+import { getMembers } from "@app/lib/api/workspace";
+import { Authenticator } from "@app/lib/auth";
+import { MembershipResource } from "@app/lib/resources/membership_resource";
+import { TriggerResource } from "@app/lib/resources/trigger_resource";
+import { UserResource } from "@app/lib/resources/user_resource";
+import { WorkspaceResource } from "@app/lib/resources/workspace_resource";
+import { promises as fs } from "fs";
+import { z } from "zod";
+
+import { makeScript } from "./helpers";
+
+const MergeFileSchema = z.array(
+  z.object({
+    email: z.string().email(),
+    sIdToKeep: z.string().min(1),
+  })
+);
+const MIN_ADMINS_FOR_REVOCATION = 2;
+
+type Member = Awaited<ReturnType<typeof getMembers>>["members"][number];
+
+interface MergeRecord {
+  primaryEmail: string;
+  oldUserId: string;
+}
+
+interface MergePair extends MergeRecord {
+  primaryUserId: string;
+  oldUserModelId: number;
+  oldEmail: string;
+}
+
+interface BlockedRecord extends MergeRecord {
+  reason: string;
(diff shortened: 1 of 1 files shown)


front/scripts/merge_mapped_identities.ts
```diff
@@ -0,0 +1,372 @@
...
+const MergeFileSchema = z.array(
+  z.object({
+    _comment: z.string().min(1),
+    email: EmailSchema,
+    sIdToKeep: z.string().min(1),
+  })
+);
+const MIN_ADMINS_FOR_REVOCATION = 2;
+
+type Member = Awaited<ReturnType<typeof getMembers>>["members"][number];
```
aubin-tchoi, line 26: nit (no impact here since it's a script but not something we want agents to replicate): hard to tell what the actual type is here, could use the actual type instead

front/scripts/merge_mapped_identities.ts
```diff
@@ -0,0 +1,372 @@
...
+  expectedOldEmail: string;
+  oldUserId: string;
+}
+
+interface MergePair extends MergeRecord {
+  primaryUserId: string;
+  oldUserModelId: number;
+  oldEmail: string;
+  oldRole: Member["workspaces"][number]["role"];
+  primaryRole: Member["workspaces"][number]["role"];
```
aubin-tchoi, line 39: nit: clearer to directly put `RoleType` here

front/scripts/merge_mapped_identities.ts
```diff
@@ -0,0 +1,372 @@
...
+  oldSeatType: Member["seatType"];
+  primarySeatType: Member["seatType"];
```
aubin-tchoi, line 41: same with `MembershipSeatType`

front/scripts/merge_mapped_identities.ts
```diff
@@ -0,0 +1,372 @@
...
+}
+
+function buildPairs(
+  records: MergeRecord[],
+  memberById: Map<string, Member>,
+  primaryByEmail: Map<string, Member>,
+  duplicateEmails: Set<string>,
+  triggerCountByEditor: Map<number, number>,
+  ssoUserIds: Set<string>,
+  activeAdminCount: number
```
aubin-tchoi, line 126: nit: would pass as a structured object

aubin-tchoi (commented):
LGTM

PR #29902, opened 2026-08-03
[front] enh: Dust-funded activation runs

id13:
## Description

Add system_activation to the FREE_ORIGINS
This is the only change needed, only the first message (nudge) can use this (modulo small period between this PR and the next two), replies by user will use the web origin like any message.

## Tests

Added a test

## Risk

Low

## Deploy Plan

Deploy front

PR #29901, opened 2026-08-03
[front] enh: Add the system_activation message origin

id13:
## Description

Add system activation message origin, which we will use to fund the nudge message (ala sidekick, but not the whole conversation)

## Tests

N/A

## Risk

Low

## Deploy Plan

Deploy front

PR #29900, opened 2026-08-03
[front] refactor: Single source of truth for client message origins

id13:
## Description

Removing duplicates

## Tests

N/A

## Risk

Low

## Deploy Plan

Deploy front

PR #29903, opened 2026-08-03
[front] fix: Authorless messages never use personal credentials

id13:
## Description

@Fraggle 

This PR introduces the concept of authorless messages (origin = system) used by activation, but could be used elsewhere too.

The idea is that, we may as a system post user messages, but never use any of the said user's credentials. And to do this I chose to make these messages authorless.

This is clearly a tradeoff as it introduces complexity (see the code touched) and if you see a simpler way to achieve the same goal, I know you will tell me 😄 

## Tests

Added some tests

## Risk

Low
Blast radius: Tool calling 😓 

## Deploy Plan

Deploy front

### DIFF
--- a/front/lib/api/assistant/conversation.ts
+++ b/front/lib/api/assistant/conversation.ts
@@ -158,6 +158,7 @@ import {
   ConversationError,
   isAgentMessageType,
   isPodConversation,
+  isSystemAuthoredUserMessage,
   isUserMessageType,
   UNRESUMABLE_AGENT_MESSAGE_STATUSES,
 } from "@app/types/assistant/conversation";
@@ -1083,6 +1084,20 @@ function canAccessAgent(
 
 class UserMessageError extends Error {}
 
+// A system-authored message has no author to be, so nobody passes this. Testing
+// that first also stops an API key, which has no `auth.user()` either, from
+// matching null against null.
+function isUserMessageAuthor(
+  auth: Authenticator,
+  message: UserMessageType
+): boolean {
+  if (isSystemAuthoredUserMessage(message)) {
+    return false;
+  }
+
+  return auth.user()?.id === message.user?.id;
+}
+
 /**
  * This method creates a new user message version. If a new message contains agent mentions, it will create new agent messages,
  * only when there are no agent messages after the edited user message.
@@ -1121,7 +1136,7 @@ export async function editUserMessage(
     });
   }
 
-  if (auth.user()?.id !== message.user?.id) {
+  if (!isUserMessageAuthor(auth, message)) {
     return new Err({
       status_code: 403,
       api_error: {
@@ -1749,6 +1764,18 @@ export async function retryAgentMessage(
       api_error: {
         type: "invalid_request_error",
         message: "Could not find the parent user message for this retry.",
+      },
+    });
+  }
+
+  // Retrying would replay the parent's server-set origin, which can carry free
+  // usage.
+  if (isSystemAuthoredUserMessage(parentUserMessage)) {
+    return new Err({
+      status_code: 403,
+      api_error: {
+        type: "workspace_auth_error",
+        message: "The answer to a message posted by Dust cannot be retried.",
       },
     });
   }
--- a/front-api/routes/poke/workspaces/[wId]/conversations/[cId]/render.ts
+++ b/front-api/routes/poke/workspaces/[wId]/conversations/[cId]/render.ts
@@ -199,6 +199,7 @@ app.post(
         agentConfiguration,
         conversation,
         agentMessage: placeholderAgentMessage,
+        userMessage,
         clientSideActionConfigurations: clientSideMCPActionConfigurations,
       },
       { jitServers, skillServers, systemSkillServers }
--- a/front/lib/actions/mcp_actions.ts
+++ b/front/lib/actions/mcp_actions.ts
@@ -97,6 +97,7 @@ import { generateRandomModelSId } from "@app/lib/resources/string_ids_server";
 import { concurrentExecutor } from "@app/lib/utils/async_utils";
 import { fromEvent } from "@app/lib/utils/events";
 import logger from "@app/logger/logger";
+import { isSystemAuthoredUserMessage } from "@app/types/assistant/conversation";
 import type { OAuthProvider } from "@app/types/oauth/lib";
 import type { ModelId } from "@app/types/shared/model_id";
 import type { Result } from "@app/types/shared/result";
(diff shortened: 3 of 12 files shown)


id13:

- #29900
    - #29901
        - #29902
            - #29903 ◀
                - #29904

front/lib/actions/mcp_metadata.ts
```diff
@@ -221,11 +222,15 @@ async function resolveRemoteServerOAuthToken(
...
     isToolExecution,
+    allowPersonalConnection,
   }: {
     mcpServerId: string;
     oAuthUseCase: MCPOAuthUseCase | null;
     remoteMCPServer: RemoteMCPServerResource;
     isToolExecution: boolean;
+    // False when no person wrote the message being answered: the run then
+    // resolves connections as if there were no user at all.
+    allowPersonalConnection: boolean;
```
id13: my problem of this approach: introducing a new param that is a bit too laster focused imo

frankaloia (approved):
I think this is the cleanest approach, but worth a second opinion given it touches the core agent loop auth

front/lib/api/assistant/conversation.ts
```diff
@@ -1083,6 +1084,20 @@ function canAccessAgent(
 
 class UserMessageError extends Error {}
 
+// A system-authored message has no author to be, so nobody passes this. Testing
+// that first also stops an API key, which has no `auth.user()` either, from
+// matching null against null.
+function isUserMessageAuthor(
```
Fraggle, line 1090: Seems quite flaky tbh.

front/lib/actions/mcp_metadata.ts
```diff
@@ -434,12 +443,17 @@ export async function connectToMCPServer(
                 );
               }
 
+              // A message nobody wrote never runs on a person's personal
+              // connection (see `isSystemAuthoredToolContext`).
+              const internalConnectionType =
```
Fraggle: I don't think it's wise, you are workarounding the admin decision here, could be an issue no ?

front/lib/api/assistant/conversation.ts
id13: can you elaborate ?

front/lib/actions/mcp_metadata.ts
id13: True, maybe filtering out personal-usecases would be better than trying to hack around

front/lib/api/assistant/conversation.ts
Fraggle: **If i understand correctly**, we are creating a real user message with a real user but thru the "origin=system" it shall not be treated as really coming from the user, right ?

That a very sneaky thing to know and we are adding quite a bit of importance on that particular origin value, it becomes quite magic.

I may have misunderstood, happy to do a quick call.

front/lib/actions/mcp_metadata.ts
Fraggle: You are maybe even going to use the admin's personal account here.

front/lib/api/assistant/conversation.ts
id13: > If i understand correctly, we are creating a real user message with a real user but thru the "origin=system" it shall not be treated as really coming from the user, right ?

Yes, do we have another way to trigger an agent that I'm unaware of ?

id13: I agree that it's a bit hacky, not 100% happy with it either, if you see a better way to make the first message free, I'm quite open to suggestions

Fraggle: Why not just avoid the user FK and just fill the user message context with the user's properties ?
(it will look like a user message from the user but not be linked to it)

Fraggle: What is the high level goal (in context of activation) ?

front/lib/actions/mcp_metadata.ts
id13: Yes, but what would it change from a regular use case ? If the admin gave this tool as workspace creds with their admin account, it's their problem

front/lib/api/assistant/conversation.ts
id13: The first message of the conversation, which is triggered by Dust, should be free, and should not use the user's credentials for tool-calls (to avoid freaking them out)

front/lib/actions/mcp_metadata.ts
Fraggle: When they setup a tool for personal connexion type, they have to setup at least once for themselve. I think if you fallback a personal tool as a workspace tool, the admin "personal connexion" might be used.

In case of workspace connexion type, they usually use a shared account.

Tools are setup for one or the other connexion type, you shall not fallback.

front/lib/api/assistant/conversation.ts
frankaloia: We don't actually need to trigger the first message as a user. We want to trigger it as "Dust", but using workspace credentials so that we can: (1) provision the pod as we need (2) call all workspace read APIs as we need (3) use the CRUD recommendation operations

So what we really need is a form of "elevated" workspace credentials. We don't just want to spoof the user credentials due to security/privacy concerns

frankaloia: To be clear, we probably do not want it to actually look like a message from the user

Fraggle: I don't know how free is handled but I believe it's using origin (so origin=system is fine for that purpose).
Regarding the tools,  you want to filter out the personal tools or all tools ?

Fraggle: I assume, only the personal one.
In that case:
- origin for "free"
- no user FK to avoid personal tool

I _believe_ it should do the trick

Fraggle: > We don't actually need to trigger the first message as a user. We want to trigger it as "Dust", but using workspace credentials so that we can: (1) provision the pod as we need (2) call all workspace read APIs as we need (3) use the CRUD recommendation operations

> So what we really need is a form of "elevated" workspace credentials. We don't just want to spoof the user credentials due to security/privacy concerns

In that case, even better, using a "user-less" user message seems like exactly what you want, no ?

Fraggle: In the context you can put a Dust name, logo etc..

id13: Yes, it's what I'm ending up doing, to "filter-out" personal tools indeed

frankaloia: Yes I believe we do want a userless message. I think we want a "user-less" user message that is able to do the following:

1. Provisions all the content required inside the Pod, including Agent.MD file and the Frame
2. Calls the analytics MCP tools for that user + workspace level read tools
3. Able to read and create new recommendations

If we use pure workspace credentials, the tricky parts are:
1. We have to be able to act as editor of a pod
2. We need to be able to make calls to the user_analytics and activation_recommendations MCP server using only that user as context

PR #29905, opened 2026-08-03
Remove all refs to branches

Fraggle:
## Description

**All messages in branches have been deleted already**

Strips all remaining `branchId` references now that `ConversationBranch` is gone (follow-up to #29871 and #29873):

- Removes `branchId` from Swagger/OpenAPI schemas (`swagger.json`, `swagger_private_schemas.ts`) and poke route response shapes.
- Drops `getBranchedInsertIndex()` and its test file; replaces `areSameRankAndBranch`/`getPredicateForRankAndBranch` with rank-only equivalents in `types.ts` and `ConversationViewer.tsx`. SSE insert logic is now a simple rank-based `findIndex`.
- Removes the `MessageItem` early-return that hid messages with a non-null `branchId` (dead guard since branch writes were already blocked).
- Strips `branchId: null` from all `MessageModel` creates and query `WHERE` clauses in `messages.ts`, `fetch.ts`, `lock.ts`, `content_fragments.ts`, and related test fixtures.
- Removes `conversationBranchId: null` from Temporal workflow call sites (`runAgentLoopWorkflow`, `retryAgentMessage`, `registerUserAnswer`, `resolveAuthentication`).
- Drops `branchId` from `createPlaceholderUserMessage`/`createPlaceholderAgentMessage` params and `getUserMessageIdFromMessageId` return type.

The `branchId` DB column is retained; no migration in this PR.

## Tests

Deleted `ConversationViewer.test.ts` (all tests were branch-specific and are no longer applicable). Updated test fixtures in `content_fragments.test.ts`, `attachments.test.ts`, `forks.test.ts`, `conversation.test.ts`, `useAgentMessageStream.test.ts`, and `mcp_actions.test.ts` to remove `branchId` fields. Tested locally.

## Risk

Low. All removed code was already dead: `branchId` was hardcoded to `null` at write time and filtered with `IS NULL` at read time since #29871. The `getBranchedInsertIndex` path was unreachable for the same reason. No DB schema change.

## Deploy Plan

Deploy front + front-api.


Fraggle:

- #29905 ◀

front/migrations/post-deploy/20260803093039_create_new_index_for_messages_and_drop_irrelevant_indexes.sql
```diff
@@ -0,0 +1,18 @@
...
+DROP INDEX CONCURRENTLY IF EXISTS "public"."messages_workspace_id_conversation_id_rank_version";
+
+DROP INDEX CONCURRENTLY IF EXISTS "public"."messages_workspace_id_conversation_id_rank_version_branch_id";
+
+DROP INDEX CONCURRENTLY IF EXISTS "public"."messages_workspace_id_conversation_id_rank_version_branch_null";
+
+DROP INDEX CONCURRENTLY IF EXISTS "public"."messages_branch_id";
+
+ALTER TABLE "public"."messages"
+DROP COLUMN "branchId";
```
id13, line 18: requires a table exclusive lock, not doable without downtime.

so don't drop it, we will cleanup one-day

id13 (approved):
pre-approving but do NOT leave the drop column statement 🙏

front/migrations/post-deploy/20260803093039_create_new_index_for_messages_and_drop_irrelevant_indexes.sql
Fraggle: Wouldn't it be super fast ?

matteotrab (approved):

front/migrations/post-deploy/20260803093039_create_new_index_for_messages_and_drop_irrelevant_indexes.sql
Fraggle: ALTER TABLE ... DROP COLUMN in Postgres 12+ is a metadata-only operation — it doesn't rewrite the table, so it completes in milliseconds even on a large messages table. The only risk is the brief ACCESS EXCLUSIVE lock it requires, which the lock_timeout guard handles.

The column is empty, i expect ms of locking.

id13: seen irl: lowering the lock_timeout is safe-ish

PR #29908, opened 2026-08-03
[MCP] Fix retrying blocked actions raised inside sub-agents

pmilliotte:
## Description

Conversations blocked on a "Manual action required" prompt could not be resumed: `POST .../retry?blocked_only=true` returned 500 and the conversation stayed stuck. Two causes:

1. **The retry endpoint never woke the caller.** `retryBlockedActions` was the only blocked-action resolution surface that didn't walk the agentic-parent chain (unlike `validateAction`, `registerUserAnswer`, `resolveAuthentication`, `validateToolFromEmail`). Retrying a prompt raised inside a sub-agent relaunched the sub-agent's loop but left its caller parked in `blocked_child_action_input_required` forever.
2. **Benign outcomes were reported as 500s.** `retryBlockedActions` returns `no_blocked_actions` / `agent_message_not_resumable` / `agent_loop_already_running` — all meaning "nothing to resume", not "resuming failed". The route only special-cased the third and 500'd on the rest. The frontend made this systematic by firing a second retry on the main conversation after the sub-agent one, which raced it and reliably resolved to "nothing to retry".

Fixes:

- Extracted the shared `isNonBlockingRetryError` guard next to the code that produces those errors, replacing the duplicate list in `resume_ancestor_conversations.ts`.
- Added `retryBlockedActionsAndResumeAncestors`, used by the retry endpoint, so retry wakes the caller like every other surface. It lives outside `retryBlockedActions` so the ancestor walk can't re-enter itself.
- The endpoint now treats the three benign codes as a no-op (`200`) instead of `500`, and the 500 branch no longer mislabels itself `invalid_request_error`.
- Dropped the frontend's second, racy retry on the main conversation — the server-side walk covers it.

## Tests

Added 2 tests for `retryBlockedActionsAndResumeAncestors` (wakes the caller on success, doesn't walk when the retry itself failed). `resume_ancestor_conversations.test.ts` passes 7/7. Note: `validate_actions.test.ts` has 2 failures (`interrupted` / `gracefully_stopped`) that reproduce unchanged on `main`.

## Risk

Low. Behaviour change worth noting: `agent_loop_already_running` now returns `200` instead of `400` on this endpoint — no client branches on it (the SDK and `useRetryMessage` both ignore non-limit errors).

## Deploy Plan

- Deploy `main` on `front`

### DIFF
--- a/front-api/routes/w/[wId]/assistant/conversations/[cId]/messages/[mId]/retry.ts
+++ b/front-api/routes/w/[wId]/assistant/conversations/[cId]/messages/[mId]/retry.ts
@@ -1,7 +1,7 @@
 import { retryAgentMessage } from "@app/lib/api/assistant/conversation";
-import { retryBlockedActions } from "@app/lib/api/assistant/conversation/retry_blocked_actions";
+import { retryBlockedActionsAndResumeAncestors } from "@app/lib/api/assistant/conversation/resume_ancestor_conversations";
+import { isNonBlockingRetryError } from "@app/lib/api/assistant/conversation/retry_blocked_actions";
 import { batchRenderMessages } from "@app/lib/api/assistant/messages";
-import { DustError } from "@app/lib/error";
 import { ConversationResource } from "@app/lib/resources/conversation_resource";
 import { isAgentMessageType } from "@app/types/assistant/conversation";
 import { workspaceApp } from "@front-api/middlewares/ctx";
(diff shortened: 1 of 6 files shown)


pmilliotte:
Closing: the root cause analysis behind this PR was wrong. The stuck conversation was caused by the caller's run_agent action being stranded in `running` (never parked as `blocked_child_action_input_required`), so every resume path was structurally blind to it. Replacing with a targeted fix.

PR #29910, opened 2026-08-03
Speed up Depot image builds

flvndvd:
## Description

GitHub reported front image builds taking 10–18 minutes while Depot showed roughly two minutes of build time when caches were hit. The discrepancy came from GitHub timing the full composite action, including two sequential Depot builds, remote builder setup, context transfers, image pushes, and redundant GitHub Actions cache imports and mode=max exports.

Depot already provides a persistent BuildKit cache, so exporting the same layers to GitHub added network overhead
without useful caching benefits. Both regional jobs also wrote to identical cache scopes concurrently. This change
removes the external GitHub cache configuration. See Depot’s troubleshooting [guidance](https://depot.dev/docs/container-builds/troubleshooting).

The workers and front-api images are now built through a single Depot Bake invocation. This allows BuildKit to schedule both targets together and deduplicate their shared dependency stages. See Depot’s Docker Bake [documentation](https://depot.dev/docs/container-builds/how-to-guides/docker-bake).

The context investigation also found that Git history was included because `.git` was missing from `.dockerignore`. A
Dockerfile-specific allowlist now limits front builds to the workspaces they actually consume and excludes unrelated
projects, generated caches, and the 86 MB `front/public` tree, which is deployed separately with the SPA. A cold
worst-case local context probe dropped from 625.97 MB to 66.36 MB, an **89% reduction**. Both `workers` and `front-api` images were built successfully with the reduced context. See Docker’s build-context [documentation](https://docs.docker.com/build/concepts/context/).

Depot also suggested avoiding build-time npx downloads. Investigation showed that tsx and esbuild were already installed as development dependencies and that their Depot steps took only 1.0 and 0.3 seconds, so they were not the reported 20-second bottleneck. They are now invoked through local npm scripts to make that behavior explicit. The `npm` cache mount is retained with locked sharing, `npm ci` prefers cached packages, and audit and funding network requests are disabled during image construction.

Wall-clock timing is now logged around Depot commands so future GitHub and Depot timings can be compared directly. The Bake configuration was validated for production and non-production tags, both production image targets build locally, and YAML parsing, Bash syntax, actionlint, package-lock validation, secret scanning, formatting, and whitespace checks pass.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -3,6 +3,7 @@
 **/target
 
 # misc
+.git
 .DS_Store
 **/*.pem
 **/dist
--- a/.github/actions/build-image/action.yml
+++ b/.github/actions/build-image/action.yml
@@ -183,69 +183,56 @@ runs:
           LATEST_TAGS+=("-t" "${REGISTRY}/${COMPONENT}:latest")
         fi
 
-        # Special handling for front component, build workers and front-api targets
+        # Special handling for front component: Bake builds both targets in parallel
+        # and deduplicates work in their shared base-deps stage.
         if [[ "$BASE_COMPONENT" == "front" ]]; then
-          echo "🏗️ Building front component with two targets (workers + front-api)"
+          echo "🏗️ Building front component with Bake (workers + front-api)"
 
-          # Build workers target
-          LATEST_TAGS_WORKERS=()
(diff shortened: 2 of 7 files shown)


id13 (approved):
🍞 🧁 🔥

zmarouf (approved):

PR #29912, opened 2026-08-03
[front] fix: prioritize prefix matches in capability search

aubin-tchoi:
## Description

Skill results in the slash command dropdown could feel unpredictable because of the fuzzyness of the search + how weird this fuzzy search sometimes behaves (contiguous title matches were tied in the fuzzy ranking, leading them to fall back to alphabetical order, even when one title started with the query).

This PR adds a new type of ranking, more suited for this kind of auto-complete UIs, where expectations are slightly different than a search bar. Capability search now ranks title prefix matches before other fuzzy title matches.

Will iterate a bit here, starting with a simple version.

## Tests

- Added tests that describe the expected behavior in a large variety of cases.

## Risk

- Low, only changes the order of results in the search.

## Deploy Plan

- Deploy front.


aubin-tchoi:
PMRR

fontanierh (commented):
Nice !

PR #29863, opened 2026-07-31
Align Batch edit spend limit with the Manage request modal

avervaet:
## Description

Until now, bulk editing a selection of members' spend limits used a radio-group form for choosing between the workspace default and a custom monthly limit. This differed from the settings-style action list used by the single-member Manage request modal and combined two distinct actions in one form.

This PR introduces a dedicated bulk Manage spend limit modal with two actions: **Allow unlimited spend** and **Set credit amount**. The unlimited-spend action now uses the same confirmation pattern as the single-member request-resolution flow and includes the selected member count before applying the existing bulk update. Setting a credit amount opens a focused input modal with the existing numeric validation.

The **Upgrade to max plan** action is intentionally not included because it does not apply to a bulk selection of members. The underlying bulk spend-limit update logic is unchanged. This PR is stacked on #29861.

## Tests

GitHub checks passed, including all test shards, lint, TypeScript checks, format check, React Doctor, and GitGuardian security checks.

## Risk

This is a user-facing change to the bulk spend-limit flow. The main risk is UI wiring or action-selection regressions; both actions continue to call the existing bulk update path, and unlimited spend now requires explicit confirmation before applying.

The previous workspace-default radio option is replaced by the explicit **Allow unlimited spend** action, so users will see a different flow for applying an unlimited limit. No backend, database, or permission changes are included, and rollback is a code-only revert. No UI-specific test or screenshot evidence was supplied beyond the passing repository checks.

## Deploy Plan

Deploy front. No database migration, backfill, or feature-flag activation is required for this change.

PR #29915, opened 2026-08-03
Use Input's built-in suffix prop in BulkEditSpendLimitModal

avervaet:
## Summary

Small follow-up, not part of the overage-admin-approval stack.

`BulkEditSpendLimitModal` hand-rolled a unit-label overlay (`relative` wrapper + absolutely-positioned `<span>` + `pr-28` padding hack) to show "credits/month" inside the amount input. Sparkle's `Input` component already has a built-in `suffix` prop for exactly this (a flanking unit/currency box), so this swaps the hack for that.

Prompted by [this review comment](https://github.com/dust-tt/dust/pull/29754#discussion_r3702285431) on #29754, which flagged the same pattern in `EditSpendLimitModal` (fixed there directly since it's in that stack) and noted the identical duplication here.

## Test plan

- [ ] Open the bulk edit spend limit modal, select "Use custom monthly limit", confirm the "credits/month" suffix renders correctly and the input still validates/formats as before

### DIFF
--- a/front/components/workspace/BulkEditSpendLimitModal.tsx
+++ b/front/components/workspace/BulkEditSpendLimitModal.tsx
@@ -150,30 +150,24 @@ function BulkEditSpendLimitForm({
 
           {kind === "override" && (
             <div className="flex flex-col gap-1.5 pl-6">
-              <div className="relative">
-                <Input
-                  id="bulk-spend-credit-limit-input"
-                  type="text"
-                  inputMode="numeric"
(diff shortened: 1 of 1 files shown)


avervaet:
Closing in favor of folding this change directly into #29863, since both PRs touched `BulkEditSpendLimitModal.tsx` and #29863 still had the old hand-rolled suffix hack. The suffix-prop change is now included as a commit on that branch.

PR #29916, opened 2026-08-03
fix: stop browser extensions showing the branch approval modal

matteotrab:
## Description

Browser extensions pop the (since deleted) branch approval sheet over every conversation.

The cause is the message payload. `ConversationBranchApprovalModal` opened on:

```ts
const branchMessagesToApprove = methods.data.get()
  .filter((m) => m.branchId === context.branchIdToApprove);
...
<Sheet open={branchMessagesToApprove.length > 0}>
```

and the context was built with `branchIdToApprove: branchIdToApprove ?? undefined`. "Remove all
refs to branches" (b518d165e9) dropped `branchId` from every serialized message, so those clients
now get `undefined` where they used to get `null` — `undefined === undefined` matches every message
in the conversation and the sheet opens.

This PR:

- Re-emits `branchId: null` on user, agent, light agent (streaming) and compaction messages.
- Adds `GET /api/w/:wId/assistant/conversations/:cId/branches` returning `{ branch: null }`, the
  shape `useOpenConversationBranch` expects, so those clients stop 404ing on every conversation
  load.

`branchId` is typed as a required `null` rather than an optional field so no construction site can
silently drop it again. Both fields are marked legacy in the type with a note to delete them once
a new extension version has been published.

## Risk

Low. Additive only — one extra `branchId: null` key on message payloads and one new read-only
endpoint that returns a constant.

## Deploy Plan

Standard deployment


matteotrab:

- #29916 ◀

aubin-tchoi (approved):
LGTM

Fraggle (commented):
Thanks @matteotrab

PR #29917, opened 2026-08-03
[front] enh: manage skill editors from info page

aubin-tchoi:
## Description

Closes https://github.com/dust-tt/tasks/issues/9897.

Skill editors can currently be removed from the skill info page, but new editors cannot be added there.

This PR allows adding editors from the existing members tab.
Users who cannot administrate the skill keep the existing read-only list.

image

## Tests

- Tested locally.

## Risk

- Low.

## Deploy Plan

- Deploy front.

### DIFF
--- a/front/components/skills/SkillEditorsTab.tsx
+++ b/front/components/skills/SkillEditorsTab.tsx
@@ -73,23 +73,9 @@ export function SkillEditorsTab({ owner, user, skill }: AgentEditorsTabProps) {
 
   return (
     <div className="flex flex-col gap-4">
-      <MembersList
-        currentUser={user}
-        membersData={{
-          members: editors.map((user) => ({
-            ...user,
-            workspace: owner,
-          })),
-          isLoading: isEditorsLoading,
-          totalMembersCount: editors.length,
-          mutateRegardlessOfQueryParams: () => Promise.resolve(undefined),
-        }}
-        showColumns={skill.canAdministrate ? ["name", "remove"] : ["name"]}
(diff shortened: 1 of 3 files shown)


front/components/skills/SkillEditorsTab.tsx
```diff
@@ -18,37 +31,130 @@ export function SkillEditorsTab({ owner, user, skill }: AgentEditorsTabProps) {
...
             ...user,
             workspace: owner,
           })),
           isLoading: isEditorsLoading,
-          totalMembersCount: editors.length,
+          totalMembersCount: selectedEditors.length,
           mutateRegardlessOfQueryParams: () => Promise.resolve(undefined),
         }}
-        showColumns={isCurrentUserEditor ? ["name", "remove"] : ["name"]}
+        showColumns={skill.canAdministrate ? ["name", "remove"] : ["name"]}
```
fabiencelier: Weirdly some editors have a cross to be removed but not all, looks a bit random 🤔 
(only the first 2 rows were present at the begining before I started added more editors)

Image

zmarouf (approved):
LGTM! Thanks!

fabiencelier (approved):

front/components/skills/SkillEditorsTab.tsx
aubin-tchoi: it is actually quite random (provisioned users I believe), let me fix that

aubin-tchoi: handling in a separate PR (this is global to the compoennt `MembersList`)

PR #29920, opened 2026-08-03
feat: Add poolCapOverrideExpiresAt to memberships

avervaet:
## Description

Until now, memberships stored the per-user pool-cap override but had no way to record when that override should revert to the seat-type default. This PR adds the data-model foundation for expiring those overrides, as the first step toward the approval flow and expiration scheduler described in [#29754](https://github.com/dust-tt/dust/issues/29754).

This PR adds a nullable `poolCapOverrideExpiresAt` timestamp to `memberships`, exposes it through the Sequelize model and `MembershipResource`, and adds a partial concurrent index for looking up memberships with an expiry. `updatePoolCapOverride` now accepts the override amount and optional expiry together; existing callers continue to set only the amount, which clears any expiry. Seat-type changes preserve both the override and its expiry. The existing `poolCapOverrideAwuCredits` value remains the enforcement source of truth, and no current caller starts setting expirations or automatically reverting overrides.

## Tests

The existing spend-limit API tests and backfill migration were updated for the new object-shaped `updatePoolCapOverride` API. Recorded PR checks passed for CodeQL, React Doctor, GitGuardian, and the change checks; the generic test jobs were skipped. No separate local test command or full-suite result was supplied.

## Risk

Low runtime risk: the schema change is additive and nullable, existing rows remain without an expiry, and the expiration scheduler and approval behavior are not introduced here. The main operational risk is applying the pre-deploy migration on the `memberships` table; the column addition uses short statement and lock timeouts, and the partial index is created concurrently.

The resource signature change requires all callers to pass the new object shape, which is covered by the updated callers in this diff. No rollback migration is included, so removing the column and index would require an explicit follow-up migration.

## Deploy Plan

1. Apply `front/migrations/pre-deploy/20260729153513_add_pool_cap_override_expires_at_to_memberships.sql` before application deployment.
2. Deploy `front` and `front-sse` together.
3. Deploy the follow-up expiration scheduler and approval-flow changes separately before enabling expiry-based behavior.

Related to #29754.

### DIFF
--- a/front/lib/resources/storage/models/membership.ts
+++ b/front/lib/resources/storage/models/membership.ts
@@ -27,6 +27,11 @@ export class MembershipModel extends WorkspaceAwareModel<MembershipModel> {
   // `spend_threshold_reached` alert (threshold = override + seat allowance)
   // is derived from this value and remains the enforcement mechanism.
   declare poolCapOverrideAwuCredits: number | null;
+  // When the override should auto-revert to the seat-type default (see the
+  // `spend_limit_expiration` Temporal sweep). NULL means it never expires.
+  // Meaningless — and ignored by enforcement — when
+  // `poolCapOverrideAwuCredits` is null.
+  declare poolCapOverrideExpiresAt: Date | null;
 
   declare userId: ForeignKey<UserModel["id"]>;
   declare user: NonAttribute<UserModel>;
@@ -79,6 +84,11 @@ MembershipModel.init(
       allowNull: true,
       defaultValue: null,
     },
+    poolCapOverrideExpiresAt: {
+      type: DataTypes.DATE,
+      allowNull: true,
+      defaultValue: null,
+    },
   },
   {
     modelName: "membership",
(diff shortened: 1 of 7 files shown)


front/lib/resources/storage/models/membership.ts
```diff
@@ -100,6 +110,14 @@ MembershipModel.init(
         where: { firstUsedAt: { [Op.ne]: null } },
         concurrently: true,
       },
+      // Partial index backing the expiration sweep's cross-workspace lookup
+      // (see `getWorkspacesWithExpiredPoolCapOverride`).
+      {
```
flvndvd, line 115: OOC does the temporal sweep runs globally or per workspace?

flvndvd (approved):

aubin-tchoi (approved):

front/lib/resources/storage/models/membership.ts
avervaet: Globally: One query to find every workspace with at least one expired override. Then it loops over those workspaces sequentially, and for each one fetches its expired memberships and reverts them with bounded concurrency (4).

PR #29913, opened 2026-08-03
Filter overage requests history by decision, and search it by user

avervaet:
## Description
- The History tab (resolved upgrade/overage requests) can now be filtered by Decision (Approved / Denied), via a dropdown next to the CSV export button
- The shared search bar (already used to filter Members and Requests) now also filters the History tab by requester name/email, resolved server-side against the user search index and applied before pagination.
- The History CSV export now reflects the current decision filter and search term instead of always exporting the full resolved set.

## Deploy Plan
- Deploy front

PR #29921, opened 2026-08-03
feat: Spend-limit expiration scheduler

avervaet:
## Description

Until now, `poolCapOverrideExpiresAt` could be stored on a membership, but no background process reverted an expired override. This PR adds the Temporal automation needed to return expired per-user spend limits to the seat-type default, as the scheduler step in the stack related to [#29754](https://github.com/dust-tt/dust/issues/29754). The prerequisite data-model change is in [#29920](https://github.com/dust-tt/dust/pull/29920).

The existing `credit_alerts` worker now self-registers a UTC schedule that runs at the start of every hour. Each run performs a cross-workspace lookup for active memberships whose non-null override expiry is in the past, then starts one independently retriable child workflow per affected workspace. Workspace sweeps process memberships with bounded concurrency, clear the Metronome per-user cap and warning alerts, reconcile the user when a Metronome contract is present, and emit a system-acted `membership.pool_cap_override_expired` audit event containing the previous and new AWU credit values.

The implementation avoids concurrent sweeps for the same workspace by using a stable workspace-scoped workflow ID. Metronome failures are surfaced instead of being silently ignored: when clearing either alert fails, the membership override is restored from its snapshot so the database does not claim the override has expired while Metronome still enforces it. The expiration audit action and schema version are added alongside the workflow behavior.

## Tests

Added activity tests covering expired-workspace discovery, ignoring active overrides, no-op sweeps, reverting only the affected workspace, clearing the Metronome cap alert, and restoring the database override when Metronome synchronization fails.

## Risk

Medium risk because this introduces scheduled automation that mutates membership spend-limit state and calls Metronome across workspaces. An incorrect expiry lookup or scheduler configuration could remove an active per-user cap earlier than intended, while an external billing failure could leave a membership pending until a later retry.

The lookup is restricted to active memberships with a non-null override and an expiry at or before the sweep time. Per-workspace failures are surfaced for Temporal retry, and cap or warning-alert clear failures restore the previous database state. The hourly schedule uses a buffer-one overlap policy, and workspace child workflow IDs prevent two sweeps from running concurrently for the same workspace. The change reuses the existing `credit_alerts` worker and queue rather than introducing a new worker.

## Deploy Plan

1. Ensure the prerequisite `poolCapOverrideExpiresAt` migration from [#29920](https://github.com/dust-tt/dust/pull/29920) has been applied.
2. Deploy `front-sse` alongside `front`.
3. The schedule is created automatically when the `credit_alerts` worker starts; no manual schedule-registration command is required.
4. Monitor the Temporal sweep and Metronome synchronization failures after deployment.

Follow-up approval-flow and UI behavior is tracked in [#29754](https://github.com/dust-tt/dust/issues/29754).


avervaet:
Will get in touch with @Fraggle before merge to check if we can reuse a queue properly instead

aubin-tchoi (approved):
LGTM

aubin-tchoi:
the queue is somewhat irrelevant, the part that will require some work is the addition of a new Temporal worker

aubin-tchoi:
but it's not a lot of work that said, and I don't see how you would go around it

front/lib/resources/membership_resource.ts
```diff
@@ -1623,6 +1623,65 @@ export class MembershipResource extends BaseResource<MembershipModel> {
...
+   * Workspaces (sorted by id ascending) that have at least one active
+   * membership whose pool cap override has expired. Global, cross-workspace
+   * lookup for the expiration sweep, then per-workspace scoped work from the
+   * caller (see `listActiveWithExpiredPoolCapOverride`).
+   *
+   * `dangerously`-prefixed: bypasses per-workspace scoping on purpose, so
+   * only call this from Temporal admin jobs or poke plugins, never from an
+   * API route.
+   */
+  static async dangerouslyGetWorkspacesWithExpiredPoolCapOverride(
```
flvndvd: name feels somewhat off given that it seats on `MembershipResources`.

flvndvd: IMHO it should be one class method here that list memberships and then at call sites another call to fetch workspace resources.

front/scripts/temporal-build/build-temporal-bundles.ts
```diff
@@ -76,6 +76,8 @@ function getWorkerDirectory(workerName: WorkerName): string | null {
       return path.join(baseDir, "temporal/remote_tools");
     case "scrub_workspace_queue":
       return path.join(baseDir, "temporal/scrub_workspace");
+    case "spend_limit_expiration":
```
flvndvd: Not sure we need a new queue for this. Wondering if there is no candidates that we could use.

front/temporal/spend_limit_expiration/admin/cli.sh
```diff
@@ -0,0 +1,2 @@
+#!/bin/sh
+npx tsx temporal/spend_limit_expiration/admin/cli.ts "$@"
```
flvndvd: This script to start ages pretty badly, the more tenants we have the more you'll run it. What about ensuring that the schedule exists when admin approves the overage? If it does not log with `panic: true` so on-call can take a look.

front/temporal/spend_limit_expiration/client.ts
```diff
@@ -0,0 +1,72 @@
...
+
+export const SPEND_LIMIT_EXPIRATION_SCHEDULE_ID =
+  "spend-limit-expiration-schedule";
+
+export async function createOrUpdateSpendLimitExpirationSchedule(): Promise<
+  Result<void, Error>
+> {
+  const client = await getTemporalClientForFrontNamespace();
+  const scheduleId = SPEND_LIMIT_EXPIRATION_SCHEDULE_ID;
+  const scheduleOptions = {
```
flvndvd: Use the type they export of casting `as const ` below.

front/temporal/spend_limit_expiration/client.ts
```diff
@@ -0,0 +1,72 @@
...
+      taskQueue: QUEUE_NAME,
+    },
+    scheduleId,
+    policies: {
+      overlap: ScheduleOverlapPolicy.BUFFER_ONE,
+    },
+    spec: {
+      // Every hour at minute 0 — bounds how long an expired override can
+      // linger before being reverted.
+      cronExpressions: ["0 * * * *"] as string[],
```
flvndvd: Same here you will be able to remove.

front/temporal/spend_limit_expiration/worker.ts
```diff
@@ -0,0 +1,42 @@
...
+    maxConcurrentActivityTaskExecutions: 4,
+    connection,
+    namespace,
+    shutdownGraceTime: SHUTDOWN_GRACE_TIME_MS,
+    interceptors: {
+      activityInbound: [
+        (ctx: Context) => {
+          return new ActivityInboundLogInterceptor(ctx, logger);
+        },
+      ],
```
flvndvd: To ensure import aliases are supported, we usually wire up the bundle options here. Without it you'll need import with relative path in the workflow file.

front/temporal/spend_limit_expiration/activities.ts
```diff
@@ -0,0 +1,86 @@
...
+import { UserResource } from "@app/lib/resources/user_resource";
+import { concurrentExecutor } from "@app/lib/utils/async_utils";
+import logger from "@app/logger/logger";
+
+/**
+ * Revert every active membership's pool cap override whose
+ * `poolCapOverrideExpiresAt` has passed, back to the seat-type default.
+ * Global sweep across all workspaces, run on a schedule (see `client.ts`).
+ */
+export async function expirePoolCapOverridesActivity(): Promise<void> {
```
flvndvd: Reading the code below, this activity has a 5 minutes timeout, which might fall short at some point, accounting also that when we deploy, we only give up to 120 seconds to the worker to finish any ongoing activitiy.

A more Temporal friendly approach here, would be:
- schedule triggers on each hour
- Look for all workspaces with at least one candidate to expire
- For each workspace starts a (child) workflow that only clear workspace overrides

avervaet: Yes, done

front/admin/audit_log_schemas/membership.pool_cap_override_expired.json
```diff
@@ -0,0 +1,14 @@
...
+  "targets": [
+    {
+      "type": "workspace"
+    },
+    {
+      "type": "user"
+    }
+  ],
+  "metadata": {
+    "previous_awu_credits": "string"
```
flvndvd: `nit` to make this audit self contained, can we have both previous (I presume with overage) and new?

front/lib/api/users/spend_limit.ts
```diff
@@ -230,14 +234,21 @@ export async function setUserSpendLimit(
         userId: user.sId,
       });
       if (clearResult.isErr()) {
+        // Metronome sync failed, so the DB and Metronome would otherwise be
+        // left out of sync until someone retries — put the DB value back.
```
flvndvd:
```suggestion
        // left out of sync until someone retries put the DB value back.
```

front/lib/api/users/spend_limit.ts
```diff
@@ -230,14 +234,21 @@ export async function setUserSpendLimit(
...
         logger.error(
           {
             workspaceId: workspace.sId,
             metronomeCustomerId: workspace.metronomeCustomerId,
             userId: user.sId,
+            previousAwuCredits,
             err: clearResult.error,
           },
-          "[Metronome PerUserCap] set(unlimited): failed to clear per-user cap alert"
+          "[Metronome PerUserCap] set(unlimited): failed to clear per-user cap alert; reverted DB override"
```
flvndvd:
```suggestion
          "[Metronome PerUserCap] failed to update metronome, overage is still active."
```

front/lib/api/users/spend_limit.ts
```diff
@@ -273,15 +284,22 @@ export async function setUserSpendLimit(
         awuCredits: totalAwuCredits,
       });
       if (upsertResult.isErr()) {
+        // Metronome sync failed, so the DB and Metronome would otherwise be
```
flvndvd, line 283: Ideally this combination lives in the resource itself, so:
i. it's re-usable
ii. easier to test and completely abstracted from more engineers

front/lib/api/users/spend_limit.ts
```diff
@@ -349,6 +367,145 @@ export async function setUserSpendLimit(
   return new Ok({ limit });
 }
 
+/**
+ * Revert an expired pool cap override back to the seat-type default. Called
+ * by the expiration sweep (`@app/temporal/spend_limit_expiration`) — never by
+ * an admin action, hence the system-actored, dedicated audit event rather
+ * than reusing `setUserSpendLimit`'s `member.spend_limit_updated`. A no-op
+ * (idempotent `Ok`) if the override was already cleared or never expires,
+ * e.g. an admin manually reverted it before the sweep ran.
```
flvndvd: DRY

front/lib/api/users/spend_limit.ts
```diff
@@ -349,6 +367,145 @@ export async function setUserSpendLimit(
...
+    poolCapOverrideAwuCredits: null,
+    poolCapOverrideExpiresAt: null,
+  });
+
+  const clearResult = await clearMetronomePerUserCapAlert({
+    metronomeCustomerId: workspace.metronomeCustomerId,
+    workspaceId: workspace.sId,
+    userId: user.sId,
+  });
+  if (clearResult.isErr()) {
```
flvndvd, line 410: Third time is that charm... except for this one. It definitely deserves it's spot in the resource 🙃.

front/lib/api/users/spend_limit.ts
```diff
@@ -349,6 +367,145 @@ export async function setUserSpendLimit(
...
+        previousAwuCredits,
+        err: clearResult.error,
+      },
+      "[SpendLimitExpiration] Failed to clear per-user cap alert; reverted DB override back, will retry on next sweep"
+    );
+    return new Err(
+      new UserSpendLimitError("metronome_error", clearResult.error.message)
+    );
+  }
+  const clearWarningResult = await clearMetronomePerUserWarningAlert({
```
flvndvd, line 428: What happened if this second API call fail, as we rollback the DB but we don't rollback first API call to metronome.

avervaet:

- #29921 ◀
    - #29963
        - #29754
            - #29960
                - #29961
                    - #29755
                        - #29861
                            - #29863
                                - #29913
                                    - #29972

front/lib/api/users/spend_limit.ts
```diff
@@ -349,6 +367,145 @@ export async function setUserSpendLimit(
...
+
+  const metronomeContractId = auth.subscription()?.metronomeContractId ?? null;
+  if (metronomeContractId) {
+    await reconcileUser({
+      auth,
+      workspace,
+      metronomeCustomerId: workspace.metronomeCustomerId,
+      userId: user.sId,
+      execute: true,
+    }).catch((err) => {
```
flvndvd: No thenable at Dust 🙃. And if you look at the function's prototype of `reconcileUser` it does return a `Result` which is our functional programming interface to avoid throwing.

front/lib/resources/membership_resource.ts
```diff
@@ -1623,6 +1623,64 @@ export class MembershipResource extends BaseResource<MembershipModel> {
...
+  ): Promise<ModelId[]> {
+    const rows = await this.model.findAll({
+      attributes: [[fn("DISTINCT", col("workspaceId")), "workspaceId"]],
+      where: {
+        poolCapOverrideExpiresAt: { [Op.lte]: now },
+        poolCapOverrideAwuCredits: { [Op.ne]: null },
+        endAt: null,
+      },
+      order: [["workspaceId", "ASC"]],
+      raw: true,
```
flvndvd, line 1666: Not sure it's needed.

front/temporal/credit_alerts/activities.ts
```diff
@@ -63,3 +68,111 @@ export async function sendCreditAlertEmailActivity({
     "[Credit Alert] Sent credit usage alert emails to workspace admins"
   );
 }
+
+/**
+ * List the workspaces that have at least one active membership whose
+ * `poolCapOverrideExpiresAt` has passed. Called once per schedule tick (see
```
flvndvd: Let's remove all file references, it really aged badly.

front/temporal/credit_alerts/workflows.ts
```diff
@@ -22,3 +42,33 @@ export async function creditAlertWorkflow({
     totalConsumedMicroUsd,
   });
 }
+
+export async function expirePoolCapOverridesWorkflow(): Promise<void> {
+  const workspaceIds = await getWorkspacesWithExpiredPoolCapOverrideActivity();
+  const { workflowId } = workflowInfo();
+
+  const results = await Promise.allSettled(
```
flvndvd: Lets use `concurrencyExecutor` for this one over `Promise.allSettled` so we avoid bursting Temporal API and keep a control flow.

front/temporal/credit_alerts/workflows.ts
```diff
@@ -22,3 +42,33 @@ export async function creditAlertWorkflow({
...
+    workspaceIds.map((workspaceId) =>
+      executeChild(expireWorkspacePoolCapOverridesWorkflow, {
```
flvndvd: It would probably be better to use the schedule as a very naive brick that starts all the child workflow in a **detached** mode, each of those is accountable, you can lift the max retry and let them churn.

front/lib/api/users/spend_limit.ts
avervaet: Metronome does not provide atomic operations that could cover both calls at once unfortunately.

front/temporal/credit_alerts/activities.ts
```diff
@@ -63,3 +68,92 @@ export async function sendCreditAlertEmailActivity({
...
+ * List the workspaces that have at least one active membership whose
+ * usage upgrade expired. The workflow then fans out one child per
+ * workspace.
+ */
+export async function getWorkspacesWithExpiredPoolCapOverrideActivity(): Promise<
+  string[]
+> {
+  const workspaceModelIds =
+    await MembershipResource.dangerouslyGetWorkspaceModelIdsWithExpiredMembershipPoolCapOverride(
+      new Date()
```
flvndvd: We usually prefer `Date.now()` over `new Date()` to deal with numbers

flvndvd (approved):
LGTM

PR #29923, opened 2026-08-03
Pricing: allow pro plan to buy up to 1000$ instead of 50$/user/month

fabiencelier:
## Description

Previous limit is `1000$` for enterprise and `min(1000, 50 * users)` for pro plan.

This move the limit up to 1000$, same as enterprise.

No UI to update at they display these values

## Deploy Plan

dceploy front

### DIFF
--- a/front/lib/credits/limits.ts
+++ b/front/lib/credits/limits.ts
@@ -1,6 +1,5 @@
 import type { Authenticator } from "@app/lib/auth";
 import { getBillingCycleFromDay } from "@app/lib/client/subscription";
-import { countEligibleUsersForCredits } from "@app/lib/credits/common";
 import { getCustomerPaymentStatus } from "@app/lib/credits/free";
 import { isEnterprisePlanPrefix } from "@app/lib/plans/plan_codes";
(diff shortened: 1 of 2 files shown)


front/lib/credits/limits.ts
```diff
@@ -104,13 +101,8 @@ export async function getCreditPurchaseLimits(
...
 
-  // Pro paying: $50/user/month, capped at $1000 per billing cycle.
-  const workspace = auth.getNonNullableWorkspace();
-  const userCount = await countEligibleUsersForCredits(workspace);
-  const cycleMaxMicroUsd = Math.min(
-    userCount * MAX_PRO_CREDIT_PER_USER_MICRO_USD,
-    MAX_PRO_CREDIT_TOTAL_MICRO_USD
-  );
+  // Pro paying: flat $1000 per billing cycle.
+  const cycleMaxMicroUsd = MAX_PRO_CREDIT_TOTAL_MICRO_USD;
```
flvndvd: Do we really need this assignment, why not directly just using the constant?

flvndvd (approved):
🙏

PR #29926, opened 2026-08-03
Pricing: fix stripe invoice cleaning by matching all expected lines to remove

fabiencelier:
## Description

This issue created a draft for an invoice because a line to remove was not correctly matched.

Lines not correctly matched look like this:
image

## Tests

added unit tests

## Risk

low, we have the check for drifts anyway

## Deploy Plan

deploy front


adrsimon (approved):
AFAIK about this, looks fair

PR #29927, opened 2026-08-03
Display credit attribution in agent messages

flvndvd:
## Description

> [!WARNING]
> This is not the final design, just a way to check a few messages as we make progress

This PR adds a credit-cost submenu to agent messages behind the `conversation_consumption_details` feature flag. It keeps the exact billed credits prominent, then explains the estimated work through agent work and context, tool usage, sub-agent charges, and cache savings without exposing token accounting.

Repeated tool executions use a generic tool label and are grouped together. The submenu displays the three largest tool
contributions and combines the remainder into an “Other tools” row to keep its height predictable.

Attribution details are fetched only when the submenu is first opened and refreshed on later openings. Existing data remains visible during revalidation, and window focus does not trigger additional requests.

image


aubin-tchoi (approved):
LGTM

PR #29933, opened 2026-08-03
[MEMBERS] Fix resending legacy builder invitations

pmilliotte:
## Description

Pending invitations created with the now-deprecated `builder` role are resent (and revoked) as `member`, instead of failing validation because `builder` can no longer be assigned.

## Tests

Tested locally; `front/lib/api/signup.test.ts` still passes.

## Risk

Low.

## Deploy Plan

- Deploy `main` on `front`


aubin-tchoi (approved):
LGTM but don't we want to also handle adding them to the backfilled builder group?

PR #29934, opened 2026-08-03
Share front API build cache across regions

flvndvd:
## Description

US and EU builds compile the same Hono API, but region-specific `NEXT_PUBLIC_*` values were included in the front-api-
build cache key. This prevented Depot from deduplicating the expensive compilation across concurrent regional builds.

This removes those values from the build stage while retaining them in the final runtime image. Tested locally to build using different regional values and it produced byte-identical bundles, and the second local build reused the complete API build stage. See Depot’s build deduplication [documentation](https://depot.dev/docs/container-builds/build-parallelism).

> [!NOTE]
> This PR does not improve build speed, but it removes duplicate computation and builder contention


dockerfiles/front.Dockerfile
```diff
@@ -173,38 +173,11 @@ FROM base-deps AS front-api-build
...
 ARG COMMIT_HASH_LONG
 ARG DATADOG_API_KEY
-ARG NEXT_PUBLIC_VIZ_URL
-ARG NEXT_PUBLIC_DUST_API_URL
-ARG NEXT_PUBLIC_DUST_STATIC_WEBSITE_URL
-ARG NEXT_PUBLIC_DUST_APP_URL
-ARG NEXT_PUBLIC_GTM_TRACKING_ID
-ARG NEXT_PUBLIC_ENABLE_BOT_CRAWLING
-ARG NEXT_PUBLIC_DATADOG_CLIENT_TOKEN
 ARG NEXT_PUBLIC_DATADOG_SERVICE
```
davidebbo, line 176: You left this one, but still yanked the matching ENV below?

flvndvd: `NEXT_PUBLIC_DATADOG_SERVICE` is still needed by the `RUN` command to name the Datadog source-map upload service.

davidebbo (approved):
LGTM

PR #29935, opened 2026-08-03
Reconcile consumption attribution with billed credits

flvndvd:
## Description

Agent message consumption details can currently report gross attributed credits below the exact billed credits. Stored
attribution rows are estimates, while `AgentMessage.costCredits` remains the authoritative bill, so exposing this shortfall produces a contradictory explanation. It can diverge quickly as `costCredits` is rounded on each agent loop where items are not.

This change reconciles the read projection without modifying stored attribution. When the billed total exceeds the stored gross attribution, the difference is added to both the displayed gross total and “Agent work and context.” Tool attribution remains unchanged because there is no evidence that would safely assign the residual to a specific tool.

When stored gross attribution exceeds the bill, the existing estimated cache savings remains the difference between those values. No attribution rows, provider usage, cache counters, pricing logic, attribution versions, or billing behavior are changed.

### DIFF
--- a/front/lib/api/assistant/agent_message_consumption_attribution/read.test.ts
+++ b/front/lib/api/assistant/agent_message_consumption_attribution/read.test.ts
@@ -207,6 +207,31 @@ describe("getAgentMessageConsumption", () => {
     ]);
   });
 
+  it("assigns an unattributed billed residual to agent work", async () => {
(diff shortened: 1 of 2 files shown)


aubin-tchoi (approved):
LGTM

PR #29936, opened 2026-08-03
[SANDBOX] Stop finished bash actions from being marked blocked again

pmilliotte:
## Description

A program left running inside the sandbox after its `bash` tool call finished could still reach `/call` and create a child action against that finished action, which set the action's status back to blocked and left two steps blocked on the same agent message (`All blocked actions must be from the same step, got 27, 24`, breaking every resume path); fixed by awaiting the exec-token revocation when a `bash` command completes, and refusing child calls whose parent action already reached a final status.

## Tests

Added a test covering a child call against a finished parent action; `create_child_action` and sandbox `tools` suites pass locally.

## Risk

Low.

## Deploy Plan

- Deploy `main` on `front`

### DIFF
--- a/front/lib/api/actions/servers/sandbox/tools/index.ts
+++ b/front/lib/api/actions/servers/sandbox/tools/index.ts
@@ -444,7 +444,8 @@ export async function runSandboxBashTool(
     ]);
   }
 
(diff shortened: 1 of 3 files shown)


fontanierh (approved):
🙏

PR #29940, opened 2026-08-03
[Activation] Notification fix: Only send email notifications for nudge conversation first message

xebeohp:
## Description
Do not send email notifications for each sub-conversation created when the activation skill calls an agent to do something and for conversations started by the user in the activation pod.
- For conversations created through the `pod_manager.create_conversation` tool, added depth so that nested conversations are parent depth + 1, and tagged the posted message with agenticMessageData so the user isn't added as a participant.
- Check and only send a notification for conversations with depth 0 (not nested conversations)
- Check that the `conversationResource.triggerId === activationPod.triggerId` before sending an email notification
     - Unread conversations/messages in the activation pod that don't go through the trigger are now treated as normal conversations and go through the conversation-unread notification workflow.

## Tests
Added tests to front/lib/notifications/triggers/project-new-conversation.test.ts for sending an email for the parent conversation and not sending an email for the nested conversation.

## Risk
Low

## Deploy Plan
Deploy Main

### DIFF
--- a/front/lib/api/actions/servers/pod_manager/tools/index.ts
+++ b/front/lib/api/actions/servers/pod_manager/tools/index.ts
@@ -1240,6 +1240,9 @@ export function createProjectManagerTools(
         // Get origin and timezone from the current conversation
         let origin: UserMessageOrigin = "web";
         let timezone = Intl.DateTimeFormat().resolvedOptions().timeZone;
+        // Sub-conversations created on behalf of an agent are nested one level below their parent
+        let parentConversationDepth = 0;
+        let originMessageId: string | null = null;
 
         if (isAgentLoopRunContext(toolContext?.runContext)) {
           const userMessage = toolContext.runContext.conversation.content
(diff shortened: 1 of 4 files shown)


frankaloia (commented):
The conversation depth is a reasonable way to make sure this is not a sub-conversation, but I think we actually might need to change the gate here of when we use this email notification. For example, if the user creates any other conversation in the pod, then we would send the same email notification. Is there any way to gate here on conversation depth + the message event source

xebeohp:
> The conversation depth is a reasonable way to make sure this is not a sub-conversation, but I think we actually might need to change the gate here of when we use this email notification. For example, if the user creates any other conversation in the pod, then we would send the same email notification. Is there any way to gate here on conversation depth + the message event source

Will check `conversationResource.triggerId === activationPod.triggerId` to make sure only conversations started by the nudge get the notification.

frankaloia (approved):

PR #29942, opened 2026-08-03
Deprecate agent_memory for dust global agent when user_memory flag is on

Inesvvv:
## Description

Once the feature flag for user_memory is on we want to backfill the agent_memory of the dust global agent into user_memory and then remove agent_memory tool from the dust global agent. 

## Tests

Tested locally that the dust global agent didn't have agent_memory once the FF was on

## Risk

Low 

## Deploy Plan

Deploy front once backfill done from agent_memory to user_memory


aubin-tchoi (approved):
LGTM

PR #29943, opened 2026-08-03
[front] fix: always add skill management server

aubin-tchoi:
## Description

Agents without any configured skills currently do not get the skill management server, which means they cannot enable a skill referenced from the input bar.

This PR always includes the skill management server when its MCP server view exists. This also removes the database lookup for equipped skills from JIT server setup (not a very costly one but still).

## Tests

- Updated tests.
- Tested locally.

## Risk

- Low.

## Deploy Plan

- Deploy front.


aubin-tchoi:
PMRR

PR #29945, opened 2026-08-03
Attribute enabled skill context to tool consumption

flvndvd:
## Description

Enabling a skill expands subsequent model input beyond the result returned by `enable_skill`. It injects the skill
instructions as a user message and exposes the tools configured by that skill. Consumption attribution previously
measured only the rendered action result, which understated the input footprint caused by enabling a skill.

This change reconstructs the additional input from the persisted skill and MCP server views. It applies the existing
tool-name prefixing, disabled-tool filtering, internal schema cleanup, and description limit before combining the
definitions and instructions with the rendered tool result. The combined text is tokenized and attributed to the same
tool row.

### DIFF
--- a/front/lib/api/assistant/agent_message_consumption_attribution/enabled_skill_footprint.ts
+++ b/front/lib/api/assistant/agent_message_consumption_attribution/enabled_skill_footprint.ts
@@ -7,9 +7,10 @@ import { renderEnabledSkillUserMessageFromInstructions } from "@app/lib/api/assi
 import type { Authenticator } from "@app/lib/auth";
 import { SkillResource } from "@app/lib/resources/skill/skill_resource";
 import type { AgentMCPActionWithOutputType } from "@app/types/actions";
+import type { JSONSchema7 as JSONSchema } from "json-schema";
 
-const EMPTY_INPUT_SCHEMA = {
-  type: "object" as const,
+const EMPTY_INPUT_SCHEMA: JSONSchema = {
+  type: "object",
   properties: {},
   required: [],
 };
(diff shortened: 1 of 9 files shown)


front/lib/resources/mcp_server_view_resource.ts
```diff
@@ -1325,6 +1326,16 @@ export class MCPServerViewResource extends ResourceWithSpace<MCPServerViewModel>
     }
   }
 
+  /**
+   * Returns the persisted tool definitions exposed by this server view. Remote views must have
+   * been fetched with the cached tools heavy attribute.
+   */
+  getServerTools(): readonly MCPToolType[] {
```
flvndvd, line 1333: This representation of the tools is best-effort. We definitely don't want to call `listTools` on the MCP in our analytics pipeline.

front/lib/api/assistant/agent_message_consumption_attribution/enabled_skill_footprint.ts
```diff
@@ -0,0 +1,139 @@
...
+
+function enabledSkillIdsFromAction(
+  action: AgentMCPActionWithOutputType
+): string[] {
+  return [
+    ...new Set(
+      (action.output ?? []).flatMap((outputBlock) => {
+        const skillId = getEnableSkillIdFromOutputBlock(outputBlock);
+        return skillId ? [skillId] : [];
+      })
```
aubin-tchoi: small nit: I'd prefer retuning `skillId ?? null` and doing a `removeNulls`

aubin-tchoi (approved):
LGTM
2 questions: do we ignore the DUST_MIME_TYPES parameters from the input schemas (those won't be exposed to the model), and do we dedupe the servers?

flvndvd:
> 2 questions: do we ignore the DUST_MIME_TYPES parameters from the input schemas (those won't be exposed to the model)

Good call, completely oversaw this part.

flvndvd:
> > 2 questions: do we ignore the DUST_MIME_TYPES parameters from the input schemas (those won't be exposed to the model)
> 
> Good call, completely oversaw this part.

So this is done through `hideInternalConfiguration`. We should not deduplicate as we should assume worst cost for a skill. I might do a follow up PR to share some module between here and conversation rendering. Will definitely do it in a dedicated PR.

PR #29948, opened 2026-08-03
[front] fix: allow removing self+provisioned users in skill/agent editors

aubin-tchoi:
## Description

When editing lists of skill (resp. agent) editors from the skill (resp. agent) info page one currently cannot remove oneself. Some users are also semi-randomly not removable. This behavior is inherited from the component `MembersList`, which was initially used for adding/removing members from spaces.

This PR adds an opt-in for removing oneself and provisioned users, enabled only by the skill and agent editor lists.

## Tests

- Tested locally.

## Risk

- Low.

## Deploy Plan

- Deploy front.


fabiencelier (approved):
LGTM, thank you for fixing

PR #29950, opened 2026-08-03
[front] fix: accept Zendesk tickets without tags in API responses

aubin-tchoi:
## Description

In the Zendesk MCP server we check the actual type of the response from Zendesk's API using zod schemas.
This allows us to make sure that we know precisely the type of the object we'll manipulate in the rest of the code, for instance preventing the use of a field that does not actually exist and keeping the code well typed.

We monitor the validation errors to be alerted on potential changes in the response type and potential data shapes we never saw up to now, which is what we saw [here](https://app.datadoghq.eu/logs?query=%22%5BZendesk%5D%20Invalid%20API%20response%20format%22&additional_filters=%5B%5D&cols=host%2Cservice&fromUser=true&messageDisplay=inline&refresh_mode=sliding&saved-view-id=193898&storage=hot&stream_sort=desc&viz=stream&from_ts=1785168932222&to_ts=1785773732222&live=true).

Since we type the rest of the code with the a type inferred from the schema, once we change the schema the typecheck tells us precisely if there's consumer code we need to update.
Practical example: we did `ticket.tags.length > 0` at some point. The zod validation allows us to reject a type that would fail it. Changing the zod, thus the type shows that this wouldn't work for the actual response we got, which we can easily tell from the typecheck.

This PR makes `tags` optional on the schema for Zendesk tickets.

## Tests

- Tested locally.

## Risk

- Low.

## Deploy Plan

- Deploy front.

### DIFF
--- a/connectors/src/connectors/zendesk/lib/sync_ticket.ts
+++ b/connectors/src/connectors/zendesk/lib/sync_ticket.ts
@@ -364,7 +364,7 @@ export async function syncTicket({
       timestampMs: updatedAtDate.getTime(),
       tags: [
         ...tags,
-        ...filterCustomTags(ticket.tags, logger),
+        ...filterCustomTags(ticket.tags ?? [], logger),
         ...customFieldTags,
       ],
       parents,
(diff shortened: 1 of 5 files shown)


aubin-tchoi:
PMRR

Inesvvv:
LGTM

PR #29953, opened 2026-08-03
feat(poke): add superuser permissions administration

zmarouf:
## Deployment

Adds a Poke admin page that lets administrators:

- Toggle a member’s `isDustSuperUser` flag.
- Add or remove members from `poke-roles.json`.
- Add or remove their Poke roles.
- Highlight JSON entries that do not match active Dust workspace members.

Changes are logged through the existing Datadog audit trail, and GCS object versioning provides rollback.

Requires deploying `front-api` and the Poke SPA, plus granting Front `storage.objectUser` access to the production bucket.

## Tests

Tested locally and on `front-edge` (Bucket is still Read Only)

## Risk

Low. Gated to Dust only + `admin` permission

## Deploy Plan

Poke and Front, test the read-only path then update the bucket permissions.


front-api/middlewares/utils.ts
```diff
@@ -153,7 +153,7 @@ export const unhandledErrorHandler: ErrorHandler = (err, ctx) => {
     {
       error: {
         type: "internal_server_error",
-        message: `Unhandled internal server error: ${error.message}`,
+        message: "An unexpected internal error occurred.",
```
zmarouf: won't ship in this PR.

front-api/routes/poke/superusers/[userId]/superuser.ts
```diff
@@ -0,0 +1,50 @@
+import config from "@app/lib/api/config";
+import { setDustSuperUser } from "@app/lib/api/poke/superusers";
+import { auditLog } from "@app/logger/logger";
+import { pokeApp } from "@front-api/middlewares/ctx";
+import type { HandlerResult } from "@front-api/middlewares/utils";
+import { validate } from "@front-api/middlewares/validator";
+import { z } from "zod";
+
+import { adminOnly, getAdminContext, mutationError } from "../helpers";
```
Comment (flvndvd) [inline front-api/routes/poke/superusers/[userId]/superuser.ts:9]: `nit` use alias import

front-api/routes/poke/superusers/helpers.ts
```diff
@@ -0,0 +1,69 @@
...
+import { apiError } from "@front-api/middlewares/utils";
+import type { Context } from "hono";
+
+export async function getAdminContext(ctx: Context<PokeCtx>) {
+  const workspaceId = config.getProductionDustWorkspaceId();
+  if (!workspaceId) {
+    throw new Error("Production Dust workspace ID is not configured.");
+  }
+
+  const auth = await Authenticator.fromSuperUserSession(
```
flvndvd, line 18: We already have a middleware that ensures that current caller is a super admin and forge the associated `auth` object. See here:

https://github.com/dust-tt/dust/blob/5c3db9a4a53c88d44dcd13e7bed85ff54e8a0b8a/front-api/middlewares/poke_auth.ts#L16

You'll find examples in a few poke endpoints as well.

front-api/routes/poke/superusers/helpers.ts
```diff
@@ -0,0 +1,69 @@
...
+    !isDevelopment() &&
+    !hasPokeRole(rolesConfig[actorEmail] ?? [], ["admin"])
+  ) {
+    return null;
+  }
+
+  return { auth, rolesConfig };
+}
+
+export function adminOnly(ctx: Context<PokeCtx>) {
```
flvndvd, line 35: `nit`
We try to keep the api error in each endpoint directly, it's not worth sharing only this piece of code.

front-api/routes/poke/superusers.test.ts
```diff
@@ -0,0 +1,204 @@
...
+}
+
+describe("poke superuser routes", () => {
+  beforeEach(() => {
+    state.productionWorkspaceId = undefined;
+    state.auditLog.mockClear();
+    fileStorageMock.reset();
+  });
+
+  it("lists active regional members and the shared roles JSON", async () => {
```
flvndvd, line 54: What's the plan to deal with regional members, given that stored permissions in GCS grant global access?

front/components/poke/pages/SuperusersPage.tsx
```diff
@@ -0,0 +1,195 @@
...
+    isLoading,
+    error,
+    auditUnavailable,
+    setRoles,
+    setDustSuperUser,
+  } = useSuperusersAdmin();
+  const [busy, setBusy] = useState(false);
+  const busyRef = useRef(false);
+
+  async function runMutation(action: () => Promise<unknown>) {
```
flvndvd, line 29: `nit` we avoid `function` keyword within function and when possible prefer anonymous functions

```suggestion
  const runMutation = async runMutation(action: () => Promise<unknown>) => {
```

front/poke/swr/superusers.ts
```diff
@@ -0,0 +1,121 @@
...
+  return Object.entries(current.roleEntries)
+    .filter(([email]) => !activeEmails.has(email))
+    .map(([email, pokeRoles]) => ({ email, pokeRoles }));
+}
+
+export function useSuperusersAdmin() {
+  const { regionInfo } = useRegionContext();
+  const { fetcher } = useFetcher();
+  const sendNotification = useSendNotification();
+  const eu = useSWR<PokeGetSuperusers>(
```
flvndvd, line 58: How does that work when we add a single tenant 😛 ?

front/poke/swr/superusers.ts
```diff
@@ -0,0 +1,121 @@
...
+
+  async function request(path: string, body: unknown, successTitle: string) {
+    try {
+      const response = await clientFetch(apiUrl(regionInfo.name, path), {
+        method: "PATCH",
+        headers: { "Content-Type": "application/json" },
+        body: JSON.stringify(body),
+      });
+      if (!response.ok) {
+        const responseBody: unknown = await response.json().catch(() => null);
```
flvndvd, line 77: We usually avoid calling `json` on a failed answer and instead use `.text()` so if parsing fails, we can still see the original message.

front/poke/swr/superusers.ts
```diff
@@ -0,0 +1,121 @@
...
+    fetcher
+  );
+  const us = useSWR<PokeGetSuperusers>(
+    apiUrl("us-central1", "/api/poke/superusers"),
+    fetcher
+  );
+  const [current, other] =
+    regionInfo.name === "europe-west1" ? [eu, us] : [us, eu];
+
+  async function request(path: string, body: unknown, successTitle: string) {
```
flvndvd, line 69: We have the body type in the endpoint, why not using it here?

front-api/routes/poke/superusers.test.ts
zmarouf: Intentionally ignoring for now because:
SCIM ensures we create all Dust workspace members in both EU and US.
culturally AE/SEs/support cover both regions because of team size.
I'd rather ship per resource permission than add more complexitiy here.

front-api/routes/poke/superusers/helpers.ts
zmarouf: The idea was that rolling a new one helped bypassed the caching from getPokeRolesForUser on the `admin` role. 
It would ensure admin revocation is caught immediately and checked on all requests on this page as opposed to waiting 5min.
Agree on principle so I'll noodle

flvndvd: You should not worry about this, since the higher up middleware does it already.

PR #29959, opened 2026-08-04
github connector: Fix connectors-worker crash (Z_BUF_ERROR: unexpected end of file) during GitHub tarball extraction

fabiencelier:
## Description

When a GitHub tarball download is truncated (connection reset, timeout on large repos), zlib raises Z_BUF_ERROR. This is expected and we already have retry logic for it: pipeline() rejects, and the extraction loop retries with a fresh stream.

The problem is a race after that. When pipeline() settles, it detaches its error listeners from the streams — but the gunzip stream wraps a native zlib handle that is still processing a chunk asynchronously. On a later event-loop tick, the native code hits the truncated data and the stream emits a second 'error' event, at a point where no listener is attached anymore. In Node, an 'error' event with zero listeners is thrown as an uncaught exception, which crashes the whole worker:
```
Error: unexpected end of file
at genericNodeError (node:internal/errors:985:15)
at wrappedFn (node:internal/errors:539:14)
at Zlib.zlibOnError [as onerror] (node:zlib:190:17)
```

The fix is to attach a persistent 'error' listener to the gunzip stream that just logs a warning. It changes nothing about control flow — pipeline() still rejects with the original error and the retry logic behaves exactly as before — it only ensures the late zombie emission has a subscriber instead of killing the process.

## Tests

no tested

## Risk

low, easy to revert

## Deploy Plan

deploy connector

### DIFF
--- a/connectors/src/connectors/github/lib/code/tar_extraction.ts
+++ b/connectors/src/connectors/github/lib/code/tar_extraction.ts
@@ -370,7 +370,10 @@ export async function extractGitHubTarballToGCS(
       // those late emissions — pipeline() still rejects with the original error.
       const gunzipStream = gunzip();
       gunzipStream.on("error", (err) => {
-        childLogger.warn({ err }, "Gunzip stream error during tarball extraction");
+        childLogger.warn(
+          { err },
+          "Gunzip stream error during tarball extraction"
+        );
       });
       await pipeline(tarballStream, gunzipStream, extract);
 
--- a/connectors/src/connectors/github/lib/code/tar_extraction.ts
+++ b/connectors/src/connectors/github/lib/code/tar_extraction.ts
@@ -364,7 +364,15 @@ export async function extractGitHubTarballToGCS(
       });
 
       // Stream: GitHub tarball -> gunzip -> tar extract -> GCS upload.
-      await pipeline(tarballStream, gunzip(), extract);
+      // gunzip-maybe's inner zlib stream can re-emit its error (e.g. Z_BUF_ERROR on a
+      // truncated download) after pipeline() has settled and detached its listeners; an
+      // 'error' event with no listener crashes the process. This listener only absorbs
+      // those late emissions — pipeline() still rejects with the original error.
+      const gunzipStream = gunzip();
+      gunzipStream.on("error", (err) => {
+        childLogger.warn({ err }, "Gunzip stream error during tarball extraction");
+      });
+      await pipeline(tarballStream, gunzipStream, extract);
 
       // Validate content-length if provided.
       if (contentLength !== null && bytesReceived < contentLength) {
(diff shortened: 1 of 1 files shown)


connectors/src/connectors/github/lib/code/tar_extraction.ts
```diff
@@ -364,7 +364,15 @@ export async function extractGitHubTarballToGCS(
...
 
       // Stream: GitHub tarball -> gunzip -> tar extract -> GCS upload.
-      await pipeline(tarballStream, gunzip(), extract);
+      // gunzip-maybe's inner zlib stream can re-emit its error (e.g. Z_BUF_ERROR on a
+      // truncated download) after pipeline() has settled and detached its listeners; an
+      // 'error' event with no listener crashes the process. This listener only absorbs
+      // those late emissions — pipeline() still rejects with the original error.
+      const gunzipStream = gunzip();
+      gunzipStream.on("error", (err) => {
+        childLogger.warn({ err }, "Gunzip stream error during tarball extraction");
```
flvndvd: I'm not clear what does the call site gets here? Does it get a proper error so Temporal eventually retries afterward?

fabiencelier: As far as I understand, the pipeline() call still failed as before
This prevent the gunzip stream wrapper from failing after the pipeline() call is finished

flvndvd (approved):

fabiencelier:
Actually, Claude managed to reproduce the issue with a minimal example like this:

```ts
 // npm i gunzip-maybe tar-stream — then: node repro.js
  const gunzipMaybe = require("gunzip-maybe");
  const tar = require("tar-stream");
  const zlib = require("zlib");
  const { PassThrough } = require("stream");
  const { pipeline } = require("stream/promises");

  async function buildTarGz() {
    // A tarball with one big file, so an entry is mid-flight when the stream dies.
    const pack = tar.pack();
    pack.entry({ name: "repo-abc123/big.txt", size: 8 * 1024 * 1024 }, Buffer.alloc(8 * 1024 *
  1024, "a"));
    pack.finalize();
    const chunks = [];
    for await (const c of pack) chunks.push(c);
    return zlib.gzipSync(Buffer.concat(chunks));
  }

  async function main() {
    const full = await buildTarGz();
    const truncated = full.subarray(0, full.length / 2); // cut-off download

    const source = new PassThrough();
    const extract = tar.extract();

    extract.on("entry", (header, stream, next) => {
      // Same as drainAndNext in tar_extraction.ts: listens for "end" only.
      stream.on("end", () => next());
      stream.resume();
    });

    const p = pipeline(source, gunzipMaybe(), extract);
    source.end(truncated); // server closes the connection after partial data

    try {
      await p;
    } catch (err) {
      console.log("pipeline rejected (caught by retry logic):", err.code);
    }
  }

  main();
```
```
 pipeline rejected (caught by retry logic): Z_BUF_ERROR
  node:events:502
        throw er; // Unhandled 'error' event
  Error: unexpected end of file
      at genericNodeError (node:internal/errors:984:15)
      at wrappedFn (node:internal/errors:538:14)
      at Zlib.zlibOnError [as onerror] (node:zlib:189:17)
```

And suggest a different fix

PR #29960, opened 2026-08-04
feat: Add grant-snapshot columns to membership_upgrade_requests

avervaet:
## Description

Until now, `membership_upgrade_requests` stored the request and resolution metadata but not the effective grant that was applied when an admin approved it. This PR adds four fields that snapshot the resolved grant: `grantedAwuCredits`, `grantedExpiryKind`, `grantedUnlimitedSpend`, and `grantedSeatType`. It also adds a `(workspaceId, resolvedAt)` index to support newest-first listings of resolved requests for the upcoming admin history view.

This is a storage and schema change only. It does not alter approval behavior, change a member's seat type or spend limits, or add the history read/population logic. Existing rows retain `NULL` for the nullable snapshot fields and `false` for `grantedUnlimitedSpend` until the follow-up logic populates these values.

Part of the stacked change that includes [#29921](https://github.com/dust-tt/dust/pull/29921) and the follow-up [#29961](https://github.com/dust-tt/dust/pull/29961).

## Risk

The migration is additive: it introduces nullable columns, a boolean with a `false` default, and a concurrent index, so existing request data and approval behavior are preserved. The main operational risk is migration execution on the existing `membership_upgrade_requests` table; the migration uses short statement and lock timeouts for the column additions and creates the index concurrently. Consumers of the new fields must handle `NULL` values for historical requests. Application rollback does not require removing the added columns.

## Deploy Plan

Apply the pre-deploy migration before deploying any follow-up code that writes or reads the grant snapshot fields. No backfill or manual data migration is included in this PR; historical rows remain unset until the follow-up population logic is deployed.

### DIFF
--- a/front/lib/resources/storage/models/membership_upgrade_requests.ts
+++ b/front/lib/resources/storage/models/membership_upgrade_requests.ts
@@ -2,7 +2,8 @@ import { frontSequelize } from "@app/lib/resources/storage";
 import { DataTypes } from "@app/lib/resources/storage/data_types";
 import { UserModel } from "@app/lib/resources/storage/models/user";
 import { WorkspaceAwareModel } from "@app/lib/resources/storage/wrappers/workspace_models";
-import type { MembershipUpgradeRequestStatus } from "@app/types/memberships";
+import type { SpendLimitExpiryKind } from "@app/types/api/users/spend_limit";
+import type { MembershipUpgradeRequestStatus, MembershipSeatType } from "@app/types/memberships";
 import {
   MAX_UPGRADE_REQUEST_REASON_LENGTH_CHARS,
   MEMBERSHIP_UPGRADE_REQUEST_PENDING_STATUS,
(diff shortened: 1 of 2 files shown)


front/lib/resources/storage/models/membership_upgrade_requests.ts
```diff
@@ -61,6 +74,36 @@ MembershipUpgradeRequestModel.init(
...
+      type: DataTypes.STRING(24),
+      allowNull: true,
+      defaultValue: null,
+    },
+    grantedUnlimitedSpend: {
+      type: DataTypes.BOOLEAN,
+      allowNull: false,
+      defaultValue: false,
+    },
+    grantedSeatType: {
```
flvndvd, line 86: Does it mean that it could change the seat type when approving a request?

avervaet: No, it is only for snapshotting and will be used to populate the overage requests history.

avervaet:

- #29921
    - #29963
        - #29754
            - #29960 ◀
                - #29961
                    - #29755
                        - #29861
                            - #29863
                                - #29913
                                    - #29972
                                        - #30351

avervaet:

PR #29961, opened 2026-08-04
feat: Resource, API, and expiration logic for upgrade-request grants

avervaet:
## Description

Until now, upgrade requests could be resolved, but the effective grant was not exposed in a history-oriented API and concurrent resolutions could apply conflicting side effects before the status compare-and-set completed. This PR adds the resource and API behavior needed for resolved-request history and records the grant that was actually applied.

The upgrade-request endpoint now supports manager-only `GET ?status=resolved` queries with stable newest-first ordering, offset pagination, a page size of 100, and a total count. Responses include the resolving admin and the grant snapshot. Pending-request behavior remains unchanged when `status` is omitted. Approval payloads can now carry either a spend-limit grant or a `grantedSeatType` for the “Upgrade to max plan” flow, but not both. The selected seat type is passed through the UI approval flow so it is recorded correctly.

Spend-limit updates accept an optional `requestId` and `expiryKind`. When linked to a request, the membership override and grant snapshot are written together, and the request ID is validated against the target user and request state. The grant snapshot is reverted if the downstream Metronome sync fails. Resolution now claims the request first with a pending-status compare-and-set, preventing two admins from applying competing resolutions; failed spend-limit application reverts the claim to pending so it can be retried.

This PR does not add the history UI. It depends on the schema and index migration from [#29960](https://github.com/dust-tt/dust/pull/29960).

## Tests

- Added API coverage for resolved-request filtering, ordering, pagination, grant snapshots, seat-approval wiring, and concurrent resolution behavior.
- Added spend-limit coverage for request validation, successful grant recording, and rollback after a Metronome failure.

## Risk

The main behavioral change is the ordering of request resolution: the request is claimed before the external spend-limit side effect, which closes the concurrent-approval race. Explicit Metronome failures roll the request back to pending and restore the previous membership and grant snapshot. Because the database update and external Metronome sync cannot be one atomic transaction, a process failure between the claim and the external sync remains an operational recovery risk. The new API fields are additive, while resolved-request consumers must handle nullable grant snapshots for older records and denied requests.

## Deploy Plan

Deploy the pre-deploy schema migration from [#29960](https://github.com/dust-tt/dust/pull/29960) before deploying this application change. No additional migration, backfill, feature flag, or manual rollout step is included in this PR. Monitor upgrade-request resolution and Metronome synchronization errors after deployment.

### DIFF
--- a/front/lib/api/users/spend_limit.ts
+++ b/front/lib/api/users/spend_limit.ts
@@ -26,6 +26,7 @@ import {
 import { getCachedMetronomeCurrentBillingPeriod } from "@app/lib/metronome/contracts";
 import { getSeatAllowancesByNormalizedSeatType } from "@app/lib/metronome/seat_types";
 import { MembershipResource } from "@app/lib/resources/membership_resource";
+import { MembershipUpgradeRequestResource } from "@app/lib/resources/membership_upgrade_request_resource";
 import type { UserResource } from "@app/lib/resources/user_resource";
 import { WorkspaceResource } from "@app/lib/resources/workspace_resource";
 import { currentCalendarMonthCycleUtc } from "@app/lib/spend_limits/cycle";
@@ -36,6 +37,7 @@ import {
   getFixedWindowCount,
   setFixedWindowCount,
 } from "@app/lib/utils/rate_limiter";
+import { withTransaction } from "@app/lib/utils/sql_utils";
 import logger from "@app/logger/logger";
 import type {
   GetUserSpendLimitResponse,
@@ -54,7 +56,8 @@ export const MAX_USER_SPEND_LIMIT_AWU_CREDITS = 2_000_000;
 type UserSpendLimitErrorType =
   | "user_not_found"
   | "workspace_not_metronome_billed"
-  | "metronome_error";
+  | "metronome_error"
+  | "request_invalid";
 
 export class UserSpendLimitError extends Error {
   constructor(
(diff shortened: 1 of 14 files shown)


front/types/api/credits/upgrade_requests.ts
```diff
@@ -2,6 +2,9 @@ import type { MembershipUpgradeRequestType } from "@app/types/memberships";
 
 export type GetUpgradeRequestsResponseBody = {
   requests: MembershipUpgradeRequestType[];
+  // Total resolved-request count, for the history view's pagination. Absent
+  // for the (unpaginated) pending-requests list.
```
avervaet, line 10: Adding another PR to the stack paginating the pending overage page also to counter scenarios with a high number of pending request. Just a nit improvement

front/lib/api/users/spend_limit.ts
```diff
@@ -244,21 +250,61 @@ export async function setUserSpendLimit(
...
-      limit.kind === "limited" && limit.expiresAt
-        ? new Date(limit.expiresAt)
-        : null,
+  // The membership override and its linked-request grant snapshot must land
+  // together for consistency.
+  const request = requestId
+    ? await MembershipUpgradeRequestResource.fetchById(auth, requestId)
+    : null;
+
+  await withTransaction(async (transaction) => {
```
aubin-tchoi, line 301: do you confirm that we only do db interactions in this transaction, nothing that hits external services?

front/types/memberships.ts
```diff
@@ -303,6 +304,23 @@ export interface MembershipUpgradeRequestType {
...
+    name: string;
+    image: string | null;
+  } | null;
+  // Snapshot of what was actually granted.
+  // At most one of `grantedAwuCredits`, `grantedUnlimitedSpend`,
+  // or `grantedSeatType` is set, depending on which resolution flow.
+  grantedAwuCredits: number | null;
+  grantedExpiresAt: number | null;
+  grantedExpiryKind: SpendLimitExpiryKind | null;
+  grantedUnlimitedSpend: boolean;
```
aubin-tchoi, line 315: feels like this we could consolidate some of the states here to have less of them

aubin-tchoi (approved):

avervaet:

- #29921
    - #29963
        - #29754
            - #29960
                - #29961 ◀
                    - #29755
                        - #29861
                            - #29863
                                - #29913
                                    - #29972
                                        - #30351

avervaet:

PR #29963, opened 2026-08-04
fix: Revert DB spend-limit writes when Metronome sync fails

avervaet:
## Description

Until now, the spend-limit setters persisted the new value to Postgres before synchronizing the derived Metronome cap alerts. If a Metronome upsert or clear failed, the database retained the new value while Metronome could still be enforcing the previous one, leaving the two systems out of sync until a retry or manual intervention.

This PR captures the previous value before writing and centralizes the failure path in `revertOnSyncFailure`. On a Metronome sync error, it restores the previous group, API key, or user override; for a newly created default-user configuration, it deletes the row instead. The error is still returned to the caller, and failed operations continue to skip audit events. Successful synchronization and the existing database-first source-of-truth behavior are unchanged.

Regression coverage covers group, API key, per-user upsert and clear, and default-user configurations with both newly created and existing rows. Related stacked change: [#29921](https://github.com/dust-tt/dust/pull/29921).

## Tests

- Added regression tests for group, API key, per-user endpoint upsert and clear, and default-user configuration rollback paths.

## Risk

This changes the failure path for spend-limit writes and reduces the risk of database/Metronome drift. There is no schema migration or change to the successful path. The operation still fails when Metronome synchronization fails, but the database value is restored before the error is returned. If the compensating database write itself fails, the rollback cannot complete and the resulting inconsistency still requires operational attention.

## Deploy Plan

The change is application-only; no migration, backfill, feature flag, or special rollout order. Just deploy front.

### DIFF
--- a/front-api/routes/w/[wId]/members/[uId]/spend_limit.test.ts
+++ b/front-api/routes/w/[wId]/members/[uId]/spend_limit.test.ts
@@ -6,7 +6,7 @@ import { createPrivateApiMockRequest } from "@app/tests/utils/generic_private_ap
 import { MembershipFactory } from "@app/tests/utils/MembershipFactory";
 import { UserFactory } from "@app/tests/utils/UserFactory";
 import { WorkspaceFactory } from "@app/tests/utils/WorkspaceFactory";
-import { Ok } from "@app/types/shared/result";
+import { Err, Ok } from "@app/types/shared/result";
 import type { WorkspaceType } from "@app/types/user";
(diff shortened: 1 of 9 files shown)


avervaet:

- #29921
    - #29963 ◀
        - #29754
            - #29960
                - #29961
                    - #29755
                        - #29861
                            - #29863
                                - #29913
                                    - #29972

flvndvd (approved):
LGTM

PR #29966, opened 2026-08-04
[front] enh: add Usage column to Skills table

aubin-tchoi:
## Description

The Manage Skills table does not show how often each skill is used. This PR adds an all-time Usage column based on distinct messages recorded in `agent_message_skills`.

The list endpoint batches counts for custom and global skills. Manage Skills requests them explicitly, so other screens using the shared skill-list hook do not trigger the additional query.

image

## Tests

- Added tests.

## Risk

- Low. Will need to monitor the impact of the query, but can be safely reverted.

## Deploy Plan

- Deploy front.

### DIFF
--- a/front/lib/resources/skill/skill_resource.ts
+++ b/front/lib/resources/skill/skill_resource.ts
@@ -95,7 +95,7 @@ import type { ModelId } from "@app/types/shared/model_id";
 import type { Result } from "@app/types/shared/result";
 import { Err, Ok } from "@app/types/shared/result";
 import { normalizeError } from "@app/types/shared/utils/error_utils";
-import { removeNulls } from "@app/types/shared/utils/general";
+import { isNumber, removeNulls } from "@app/types/shared/utils/general";
 import type { LightWorkspaceType } from "@app/types/user";
 import assert from "assert";
 import groupBy from "lodash/groupBy";
@@ -109,7 +109,7 @@ import type {
   Transaction,
   WhereOptions,
 } from "sequelize";
-import { Op } from "sequelize";
+import { col, fn, Op } from "sequelize";
 
 export type SkillMCPServerConfiguration = {
   view: MCPServerViewResource;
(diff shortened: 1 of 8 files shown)


front/lib/resources/skill/skill_resource.ts
```diff
@@ -2580,6 +2584,64 @@ export class SkillResource extends BaseResource<SkillConfigurationModel> {
...
+                },
+              }
+            : null,
+          globalSkillIds.length > 0
+            ? { globalSkillId: { [Op.in]: globalSkillIds } }
+            : null,
+        ]),
+      },
+      group: ["customSkillId", "globalSkillId"],
+    });
```
fabiencelier, line 2627: Is this going to scale with the table growing and the number of skills growing ?

Maybe we should only fetch the recent activity in the 30 last day and call the column "Recent usage"  + add an index on "workspaceId, customSkillId, createdAt"

It changes a bit the prupose of the column, so ok to keep the whole scan for now and see later too

front/lib/resources/skill/skill_resource.ts
```diff
@@ -2580,6 +2584,64 @@ export class SkillResource extends BaseResource<SkillConfigurationModel> {
     return result;
   }
 
+  /**
+   * Count distinct agent messages using each skill, keyed by skill sId.
+   */
+  static async batchFetchMessageCounts(
```
fabiencelier, line 2590: canyou pelase add a unit test for this resource method ?

fabiencelier (approved):

front/lib/resources/skill/skill_resource.ts
aubin-tchoi: filtering on the last 30 days will actually be more expensive

aubin-tchoi: index on createdAt will add quite a lot of overhead + storage

aubin-tchoi: testing on front-edge

fabiencelier: Ok, lets test it like that and reconsider later if needed 
Just something to be aware of then, and make sure the time to resolve this info does not slowly grow too much on large workspaces with many skills and messages

aubin-tchoi: it does add a bit of latency on average (~100ms on our workspace)
Since this will be tackled when we revamp Manage Skills I'll keep as is temporarily
If we end up keeping the postgres lookup, I'll see if we can improve this by maybe fetching the more lightweight info since optimizing this fetch could be tricky

fabiencelier: one big win would be server side pagination to not fetch all the skills activity

fabiencelier: but 100 ms only for the manage skills page LGTM

PR #29967, opened 2026-08-04
[front] fix: load favorite skills' effective spaces in agent loops

aubin-tchoi:
## Description

Posting a message with a favorited code-defined skill fails with `effectiveSpaceIds must be provided with agentLoopData`.

Propagate the already-resolved effective space IDs through the favorites lookup so global favorites are loaded with the same context as the rest of the agent loop skills.

## Tests

- Updated the existing favorites test to cover a code-defined skill with full agent-loop data.

## Risk

- Low.

## Deploy Plan

- Deploy front.


aubin-tchoi:
PMRR

PR #29969, opened 2026-08-04
Add: admin controlled toggle

Fraggle:
## Description

Adds an `isAdminControlled` mode to Pods, gated by the `admin_controlled_pods` feature flag (dust_only). When enabled, workspace admins take exclusive control of membership and connected data for a Pod:

- The `space_editors` group is **frozen**: existing editors retain read/write but lose admin powers. Editor promotions are blocked; removals from the Pod still drop users from the frozen group.
- Members (including frozen editors) can always leave the Pod.
- `canAdministrate` is re-routed to workspace admins only via role-based ACL in `GroupSpaceEditorResource.requestedPermissions` and `GroupSpaceMemberResource`.

Backend: `PATCH /spaces/[spaceId]/project_metadata` accepts `isAdminControlled`, enforced with a FF check + `auth.isAdmin()` guard. `SpaceResource.fromModel` LEFT JOINs `ProjectMetadataModel` to hydrate `isAdminControlled` on every space fetch. New `updateIsAdminControlled` on `ProjectMetadataResource`.

Frontend: New `AdminControlledPodTile` in Pod settings with a `SliderToggle` and confirmation dialogs. `ManageUsersPanel` and `PodMembersTable` suppress the editor column and block role changes when admin-controlled. `PodHeaderActions` and `PodMenu` update `canLeavePod` and gate rename/archive via `canAdministratePod`.

Pre-deploy SQL: adds `isAdminControlled BOOLEAN NOT NULL DEFAULT false` to `project_metadata`.

image

## Tests

- Added integration tests in `space_resource.test.ts` covering: frozen-editor read/write-but-not-admin, workspace-admin-only administrate, manager blocked, `addEditors`/`removeEditors` returning `unauthorized`, `updatePermissions` blocking promotions and allowing removals, frozen editors blocked from managing members.
- Updated `project_metadata_resource.test.ts` and `project_metadata.test.ts` snapshots with `isAdminControlled: false`.
- Tested locally.

## Risk

Low. The feature is fully gated behind `admin_controlled_pods` (dust_only), so no live user impact on deploy. The SQL migration adds a nullable-safe `BOOLEAN NOT NULL DEFAULT false` column — online, no locking risk. The `SpaceResource.fromModel` LEFT JOIN on `ProjectMetadataModel` is a minor query overhead on all space fetches; existing rows default to `false` with no behavioral change.

## Deploy Plan

1. Run pre-deploy SQL migration: `front/migrations/pre-deploy/20260804094318_add_is_admin_controlled_to_project_metadata.sql`
2. Deploy front + front-api.
3. Enable `admin_controlled_pods` FF per workspace to activate.

### DIFF
--- a/front/lib/resources/space_resource.ts
+++ b/front/lib/resources/space_resource.ts
@@ -10,6 +10,7 @@ import { GroupSpaceViewerResource } from "@app/lib/resources/group_space_viewer_
 import { ContentFragmentModel } from "@app/lib/resources/storage/models/content_fragment";
 import { GroupMembershipModel } from "@app/lib/resources/storage/models/group_memberships";
 import { GroupSpaceModel } from "@app/lib/resources/storage/models/group_spaces";
+import { ProjectMetadataModel } from "@app/lib/resources/storage/models/project_metadata";
 import { SandboxOwnerModel } from "@app/lib/resources/storage/models/sandbox";
 import { SandboxEnvVarModel } from "@app/lib/resources/storage/models/sandbox_env_var";
 import { SpaceModel } from "@app/lib/resources/storage/models/spaces";
@@ -116,6 +117,29 @@ export class SpaceResource extends BaseResource<SpaceModel> {
       space.get(),
       space.groupSpaces.map(SpaceGroupReference.fromGroupSpaceModel)
     );
+  }
+
+  /**
+   * Project-only: when true, workspace admins are the only people with
+   * editor/admin powers. Enabling demotes editors to members; disabling
+   * promotes the oldest member back to editor.
+   *
+   * Fetched on demand from project_metadata (not joined on every space load).
+   */
+  async fetchIsAdminControlled(): Promise<boolean> {
+    if (!this.isProject()) {
+      return false;
+    }
+
+    const metadata = await ProjectMetadataModel.findOne({
+      attributes: ["isAdminControlled"],
+      where: {
+        spaceId: this.id,
+        workspaceId: this.workspaceId,
+      },
+    });
+
+    return metadata?.isAdminControlled ?? false;
   }
 
   static async makeNew(
@@ -999,6 +1023,21 @@ export class SpaceResource extends BaseResource<SpaceModel> {
           "A user cannot be both a member and an editor of the same space."
         );
 
+        const isAdminControlled = await this.fetchIsAdminControlled();
+
+        // Admin-controlled Pods have an empty editor group; workspace admins
+        // administrate via role. Reject any attempt to set editors.
+        if (isAdminControlled && this.isProject()) {
+          if (editorIds.length > 0) {
+            return new Err(
+              new DustError(
+                "unauthorized",
+                "Editors cannot be set while this Pod is admin-controlled."
+              )
+            );
+          }
+        }
+
         // Handle member-based management
         const users = await UserResource.fetchByIds(memberIds);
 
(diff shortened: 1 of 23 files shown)


Fraggle:

- #29969 ◀
    - #29983
        - #29989

front/lib/resources/storage/models/project_metadata.ts
```diff
@@ -19,6 +19,11 @@ export class ProjectMetadataModel extends WorkspaceAwareModel<ProjectMetadataMod
   declare todoGenerationEnabled: CreationOptional<boolean>;
   /** First-run window only; cleared after first successful analysis. Internal / workflow. */
   declare initialTodoAnalysisLookback: CreationOptional<string | null>;
+  /**
+   * When true, workspace admins control membership and can attach connected data
+   * (Space DataSourceViews) to this Pod. Opt-in; gated by `admin_controlled_pods`.
+   */
+  declare isAdminControlled: CreationOptional<boolean>;
```
matteotrab, line 26: Why CreationOptional?

front/lib/resources/storage/models/spaces.ts
```diff
@@ -22,6 +23,7 @@ export class SpaceModel extends SoftDeletableWorkspaceAwareModel<SpaceModel> {
   declare managementMode: CreationOptional<"manual" | "group">;
 
   declare groupSpaces: NonAttribute<GroupSpaceModel[]>;
+  declare projectMetadata: NonAttribute<ProjectMetadataModel | null>;
```
matteotrab: Why do we need this?

front/lib/resources/group_space_member_resource.ts
```diff
@@ -214,11 +214,15 @@ export class GroupSpaceMemberResource extends GroupSpaceBaseResource {
...
-          editorGroupSpaces.map((egs) => ({
-            id: egs.groupId,
-            permissions: ["admin", "read", "write"],
-          }));
+        // Admin-controlled: editor group is empty and membership writes are for
+        // workspace admins (role). Ignore editor groups for this ACL.
+        const isAdminControlled = await this.space.fetchIsAdminControlled();
+        const editorGroupsPermissions: GroupPermission[] = isAdminControlled
+          ? []
+          : editorGroupSpaces.map((egs) => ({
```
matteotrab: Ignoring the group shouldn't be really needed

front/lib/resources/space_resource.ts
```diff
@@ -1175,6 +1218,158 @@ export class SpaceResource extends BaseResource<SpaceModel> {
...
+            await GroupResource.batchInvalidateGroupIdsCacheForUsers(
+              membershipsToCreate.map((m) => [
+                {
+                  user: { id: m.userId },
+                  workspace: { id: this.workspaceId },
+                },
+              ])
+            );
+          });
+        }
```
matteotrab: If I remember correctly there should be a dangerouslyAddMembers function that does this

matteotrab (approved):
Nice! Left a couple of questions

front/lib/resources/group_space_member_resource.ts
Fraggle: I agree.

front/lib/resources/storage/models/spaces.ts
Fraggle: I think it's useless, good catch

front/lib/resources/storage/models/project_metadata.ts
Fraggle: Likely because it has a default.

front/lib/resources/group_space_member_resource.ts
Fraggle: Removed — editor group is empty in admin-controlled mode anyway, so the ACL special-case was unnecessary.

front/lib/resources/space_resource.ts
Fraggle: Good catch — switched to `memberGroup.dangerouslyAddMembers` instead of the raw `bulkCreate` + cache invalidation.

PR #29970, opened 2026-08-04
[front] feat: audit pod function tool calls in the tool.* namespace

davidebbo:
## Problem

A tool call made by a pod function leaves no audit trail. The identical call made by an agent emits `tool.executed`, `tool.approval_requested`, and `tool.approval_resolved`. Both paths share `runToolWithStreaming`, but the audit emits live above it in the agent loop activity, so the function path (`create_sandbox_function_mcp_action.ts`, `run_sandbox_function_tool.ts`, `validate_action.ts`) emitted nothing.

Same tool, same MCP server, same stake level, same approval bubbled to a user who clicked Approve — no record.

## Approach

Fold the function path into the same three actions rather than adding a `pod_function.*` namespace, so an admin asking "everything that touched Slack" gets both kinds of row from one query. `actor_type` distinguishes an agent from a pod backend.

This required dropping the `agent` target from the three shared schemas. A pod function has no agent — a frame-triggered invocation has none anywhere in scope — and `audit_log_schemas.test.ts` requires the target-type set to match the schema exactly at every emit site. That constraint is why the work was deferred, per the `TODO(2026-07-10 SANDBOX_FUNCTIONS)` this PR removes.

Dropping it costs little:

- `tool.executed` and `tool.approval_requested` already set `actor: { type: "agent", … }`. The target duplicated the actor — same `id`, same `name`, same event.
- `tool.approval_resolved` has the deciding user as actor, so the agent is real information there. It moves to `agent_id` / `agent_name` metadata.

The only behavioral loss: filtering on `targets.type = "agent"` becomes `actor.type = "agent"`. Same ids, same cardinality.

## Changes

**Schemas** — all three now declare `[workspace, tool]`, plus `pod_id`, `pod_function_id`, `pod_function_slug`, `invocation_id`, `invocation_origin`; `agent_id` / `agent_name` on `tool.approval_resolved` only. Registered (`tool.executed` v5, both approvals v2) with `schema_versions.json` committed.

**Agent path** — dropped the agent target at four sites. Side benefit: `validate_actions.ts` and `blocked_actions.ts` both used to `return` early and emit nothing when `getLightAgentConfiguration` returned null. The agent now only feeds metadata, so those events survive that lookup failing.

**Function path** — three new emits, all via `emitAuditLogEvent({ auth })` so the actor is the initiating user per `[AUDIT2]`:

| Action | Site |
| --- | --- |
| `tool.approval_requested` | `create_sandbox_function_mcp_action.ts`, `blocked_validation_required` branch |
| `tool.approval_resolved` | `validate_action.ts`, replacing the TODO |
| `tool.executed` | `run_sandbox_function_tool.ts`, `tool_success` (previously an empty branch) |

New `lib/api/sandbox_functions/audit.ts` builds the pod metadata in one place so the keys cannot drift from the schema files.

## Deploy safety

Old pods read the previous version from their own build and emit the 3-target shape, which that version still accepts; new pods emit 2 targets under the new version. WorkOS retains prior versions, so both shapes stay valid throughout the rollout — no coordinated restart.

## Before / after

`tool.executed`, agent-origin — the agent target is dropped, `metadata` is unchanged:

```diff
   "targets": [
     { "id": "BHWrN3HaUa", "name": "david", "type": "workspace" },
-    { "id": "dust", "name": "dust", "type": "agent" },
     { "id": "sandbox__bash", "name": "bash", "type": "tool" }
   ],
-  "version": 4,
+  "version": 5,
```

`tool.executed`, function-origin (new):

```json
{
  "actor": { "id": "IvcQRUlbYw", "name": "David Ebbo", "type": "user" },
  "action": "tool.executed",
  "targets": [
    { "id": "BHWrN3HaUa", "name": "david", "type": "workspace" },
    { "id": "slack__post_message", "name": "post_message", "type": "tool" }
  ],
  "metadata": {
    "actor_type": "user",
    "tool_name": "post_message",
    "mcp_server_name": "slack",
    "pod_function_slug": "add-comment",
    "invocation_origin": "interactive_session"
  }
}
```

## Known gap, not addressed here

`tool.executed` fires only on success, on both paths. Failed and denied tool calls leave no audit row. Pre-existing; worth a follow-up.

## Testing

- `npx tsgo --noEmit` clean
- `audit_log_schemas.test.ts` 9/9 — includes the emit-site scan that checks all seven emits against the schemas
- `blocked_actions.test.ts` + `lib/api/sandbox_functions` 47/47
- `front-api` sandbox-function route tests 26/26


fontanierh (approved):
Nice 🙏

PR #29976, opened 2026-08-04
[front] fix: hide usage for system skills in Manage Skills

aubin-tchoi:
## Description

System skills are always active, so we don't register them in `agent_message_skills` (was originally used to track which skills were enabled for a given message). The usage therefore is not populated, and would not mean much.

This PR bypasses this by just hiding the usage in this case (replacing with a -).
Not ideal but backfilling would be quite costly here.

## Tests

- Updated endpoint coverage for system skill usage.

## Risk

- Low.

## Deploy Plan

- Deploy front and front-api.


fabiencelier (approved):
LGTM

PR #29979, opened 2026-08-04
perf: skip attachment derivation in GET sandbox/actions

adrsimon:
## Description

`GET /api/v1/w/:wId/sandbox/actions` was calling `listAttachments` on every request to feed `getJITServers`. That's four serial queries, one of which is a wide `IN` over `agent_mcp_action_output_items`. It showed up during a burst of polls from one workspace: pool exhaustion, and that query at 113s.

`createSandboxChildAction` (the `/call` path) already passes `attachments: []`, so the attachment-gated servers (`conversation_files`, `query_tables_v2`, folder `search`) were already listed by the GET but rejected by `/call`.

Also, this PR splits the two identical `"Tool is not available to this agent or conversation."` errors in `create_child_action.ts` and added a log line with `serverViewId` / `toolName` / `conversationId` on the failure path. ~1.4k of these in the last 30 days and no way to tell which was which, or for which tool.

## Tests

- `front-api` route tests pass (`sandbox/actions/index.test.ts`, `sandbox/actions/call.test.ts`).
- Updated the assertion in `create_child_action.test.ts` that pinned the old message.

## Risk

Low, but notable behaviour change: sandboxes no longer see `conversation_files`, `query_tables_v2` or folder `search` in the tool list. Nothing that worked before stops working, since `/call` already refused them.

Rollback is a revert.

## Deploy Plan

Nothing special, no migration.

### DIFF
--- a/front-api/routes/v1/w/[wId]/sandbox/actions/call.ts
+++ b/front-api/routes/v1/w/[wId]/sandbox/actions/call.ts
@@ -103,13 +103,29 @@ app.post(
     });
 
     if (result.isErr()) {
-      return apiError(ctx, {
-        status_code: 400,
-        api_error: {
-          type: "invalid_request_error",
-          message: result.error.message,
+      logger.error(
+        {
+          err: result.error,
+          conversationId: claims.cId,
+          agentMessageId: claims.mId,
+          parentActionId: claims.actionId,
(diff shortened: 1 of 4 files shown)


fontanierh (approved):
Lgtm

PR #29980, opened 2026-08-04
feat: add front.agent_message_consumption_analytics index

sfriquet:
## Description

Add ES index `front.agent_message_consumption_analytics` to capture usage per model call and tool call.
See https://app.notion.com/p/dust-tt/Analytics-Data-Model-for-Cost-Attribution-3b128599d94180cba49bcde500e24481

This index is meant to replace `front.agent_message_analytics` to power the Analytics, allowing both message level metrics (e.g: "get cost breakdown per agent or user") and to deep dive into skills and tools usage ("get skills or tools ranked by credit consumption").

This index contains one document per llm call and tool call. Multiple documents represent the breakdown of a single `agent_message`.

This index will mostly be fed from table `agent_message_consumption_items`. The main difference between this pg table and this index is this index collapses the item types `system` + `input` + `output` + `reasoning` into a single document.

## Tests

```
$ NODE_ENV=development npx tsx scripts/create_elasticsearch_index.ts --index-name agent_message_consumption_analytics --index-version 1 --execute 
[14:26:25.626] INFO (91681): Configuration: {"execute":true}
[14:26:25.626] INFO (91681):   Index name: front.agent_message_consumption_analytics_1 {"execute":true}
[14:26:25.626] INFO (91681):   Alias: front.agent_message_consumption_analytics {"execute":true}
[14:26:25.626] INFO (91681):   Region: local {"execute":true}
[14:26:25.626] INFO (91681):   Remove previous alias: false {"execute":true}
[14:26:25.644] INFO (91681): CHECK: Create index 'front.agent_message_consumption_analytics_1' with alias 'front.agent_message_consumption_analytics' in region 'local' (remove previous alias: false)? (y to confirm) {"execute":true}
y
[14:26:27.922] INFO (91681): Creating index front.agent_message_consumption_analytics_1... {"execute":true}
[14:26:28.035] INFO (91681): ✅ Index created: front.agent_message_consumption_analytics_1 {"execute":true}
[14:26:28.035] INFO (91681): Creating alias front.agent_message_consumption_analytics... {"execute":true}
[14:26:28.053] INFO (91681): ✅ Alias created: front.agent_message_consumption_analytics {"execute":true}
[14:26:28.053] INFO (91681): 🎉 Index creation complete! {"execute":true}
```
## Risk

Low. Safe to rollback until we actually start writing and reading from this index.

## Deploy Plan

Front only. Need to run script above to create the index in production.

### DIFF
--- a/front/lib/analytics/indices/agent_message_consumption_analytics_1.mappings.json
+++ b/front/lib/analytics/indices/agent_message_consumption_analytics_1.mappings.json
@@ -1,6 +1,6 @@
 {
   "properties": {
-    "timestamp": {
+    "completed_at": {
       "type": "date"
     },
     "workspace_id": {
--- a/front/lib/analytics/indices/agent_message_consumption_analytics_1.mappings.json
+++ b/front/lib/analytics/indices/agent_message_consumption_analytics_1.mappings.json
@@ -81,13 +81,8 @@
     "context_origin": {
       "type": "keyword"
     },
-    "api_key_name": {
-      "type": "text",
-      "fields": {
-        "keyword": {
-          "type": "keyword"
-        }
-      }
+    "api_key_id": {
+      "type": "keyword"
     },
     "tool": {
       "type": "object",
--- a/front/lib/analytics/indices/agent_message_consumption_analytics_1.mappings.json
+++ b/front/lib/analytics/indices/agent_message_consumption_analytics_1.mappings.json
@@ -75,6 +75,9 @@
         }
       }
     },
+    "trigger_id": {
+      "type": "keyword"
+    },
     "source": {
       "type": "keyword"
     },
--- a/front/lib/analytics/indices/agent_message_consumption_analytics_1.mappings.json
+++ b/front/lib/analytics/indices/agent_message_consumption_analytics_1.mappings.json
@@ -78,7 +78,7 @@
     "trigger_id": {
       "type": "keyword"
     },
-    "source": {
+    "context_origin": {
       "type": "keyword"
     },
     "api_key_name": {
@@ -155,8 +155,8 @@
     "credit_micro": {
       "type": "long"
     },
-    "is_billed": {
-      "type": "boolean"
+    "usage_type": {
+      "type": "keyword"
     },
     "status": {
       "type": "keyword"
--- a/front/lib/analytics/indices/agent_message_consumption_analytics_1.mappings.json
+++ b/front/lib/analytics/indices/agent_message_consumption_analytics_1.mappings.json
@@ -81,7 +81,7 @@
     "context_origin": {
       "type": "keyword"
     },
-    "api_key_id": {
+    "api_key_name": {
       "type": "keyword"
     },
     "tool": {
--- a/front/lib/analytics/indices/agent_message_consumption_analytics_1.mappings.json
+++ b/front/lib/analytics/indices/agent_message_consumption_analytics_1.mappings.json
@@ -1,0 +1,168 @@
+{
+  "properties": {
+    "timestamp": {
+      "type": "date"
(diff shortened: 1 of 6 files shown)


front/lib/analytics/indices/agent_message_consumption_analytics_1.mappings.json
```diff
@@ -0,0 +1,168 @@
+{
+  "properties": {
+    "timestamp": {
```
flvndvd: What's timestamp here? Can we try to be precise if it's `updatedAt` or `createdAt`?

front/lib/analytics/indices/agent_message_consumption_analytics_1.mappings.json
```diff
@@ -0,0 +1,168 @@
...
+    },
+    "user": {
+      "type": "object",
+      "properties": {
+        "id": {
+          "type": "keyword"
+        }
+      }
+    },
+    "source": {
```
flvndvd: Now that I'm seeing this how can we aggregate per trigger/source for example?

sfriquet: `trigger` is meant to be one of the possible `source`. `source` is meant to be a less granular version of `context_origin`.
if you meant drilling within trigger... I didn't have that mind, I guess we can add `trigger_id`.

front/lib/analytics/indices/agent_message_consumption_analytics_1.mappings.json
sfriquet: I'll use `completed_at` since we already have it in the pg table and it works both for `tool` and `llm` call .

sfriquet: note: setting it non-null, I'm assuming it doesn't make sense to record `pending` tool call, but we can still records failed or even pending with a completed_at and a `status: pending`

front/lib/analytics/indices/agent_message_consumption_analytics_1.mappings.json
sfriquet: regarding `source`, my understanding is we want to have coarser analytics than the current `context_origin`, that being said... this looks more like a display issue. I'm tempted to add back the fine grain `context_origin`, which we can later re-aggregate.

flvndvd: What values does `content_origin` bear in the current index?

sfriquet: It can be any of those: https://github.com/dust-tt/dust/blob/main/front/types/assistant/conversation.ts#L104

front/lib/analytics/indices/agent_message_consumption_analytics_1.mappings.json
```diff
@@ -0,0 +1,171 @@
...
+        }
+      }
+    },
+    "trigger_id": {
+      "type": "keyword"
+    },
+    "context_origin": {
+      "type": "keyword"
+    },
+    "api_key_name": {
```
flvndvd: Shouldn't we have an object and store the id as well? So if the api key name changes it still make sense when grouping?

flvndvd (approved):
LGTM

front/lib/analytics/indices/agent_message_consumption_analytics_1.mappings.json
sfriquet: api keys don't have sId today and their names are immutable. It's used by metronome to enforce some limit afaict so it sounds unlikely we'll allow renaming api key for the time being.

PR #29981, opened 2026-08-04
[front] enh: add search aliases for global skills in search dropdown

aubin-tchoi:
## Description

Certain global skills can be hard to find from the slash menu when people still think of one with another name (e.g. old name, shortened name).

This adds an alias map for global skills and lets slash search match aliases without changing the displayed skill name. Exact aliases rank normally, while partial alias matches stay below canonical-name matches. Deep Dive now finds Go Deep for instance.

## Tests

- Added an entire test suite.

## Risk

- Low. Only changes the order of results in the slash command.

## Deploy Plan

- Deploy front.


fontanierh (approved):
🙏

PR #29983, opened 2026-08-04
Add: connected data tab for admin controlled pod.

Fraggle:
## Description

Adds a "Connected Data" tab to `PodPage` for admin-controlled Pods. The tab is gated on `podInfo.isAdminControlled` in both the nav trigger and content rendering.

`PodConnectedDataTab` reuses `SpaceResourcesList` and `SpaceDataSourceViewContentList` with breadcrumb navigation driven by `dsvId` / `parentId` / `q` query params (via `useQueryParams`). A `useEffect` in `PodPage` strips those params from the URL when the user navigates away from the tab, and a second guard redirects back to `conversations` if the tab is somehow reached on a non-admin-controlled Pod.

image

## Tests

Tested locally on an admin-controlled Pod: tab appears, data source list loads, breadcrumb navigation works, query params are cleaned up on tab switch. Verified the tab is hidden on non-admin-controlled Pods.

## Risk

Low. Front-only change; tab is gated on `isAdminControlled` so no visible impact on existing Pods.

## Deploy Plan

Deploy `front`.

### DIFF
--- a/front/components/pages/pod/PodPage.tsx
+++ b/front/components/pages/pod/PodPage.tsx
@@ -13,6 +13,7 @@ import { useIsMobile } from "@app/lib/swr/useIsMobile";
 import { classNames } from "@app/lib/utils";
 import {
   CheckCircle,
+  CloudArrowLeftRight,
   Folder,
   MessageChatSquare,
   NavTabPill,
@@ -51,6 +52,7 @@ export function PodPage() {
     podId,
     podUiPreferences,
     setPodUiPreferences,
+    isAdminControlled: podInfo?.isAdminControlled,
   });
 
   if (isPodsInfoLoading) {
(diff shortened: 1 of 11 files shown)


Fraggle:

- #29969
    - #29983 ◀
        - #29989

front/components/pages/pod/PodPage.tsx
```diff
@@ -53,6 +55,36 @@ export function PodPage() {
...
+      podInfo &&
+      !podInfo.isAdminControlled &&
+      currentTab === "connected_data"
+    ) {
+      handleTabChange("conversations");
+    }
+  }, [podInfo, currentTab, handleTabChange]);
+
+  // Drop Connected Data navigation query params when leaving that tab.
+  useEffect(() => {
```
matteotrab: We can move this in the `usePodTabs` and avoid the useEffect

front/lib/api/actions/servers/pod_manager/build_pod_search_data_sources.ts
```diff
@@ -113,23 +114,41 @@ export async function buildPodSearchDataSources(
 
   const podDsViewRes = await fetchProjectDataSourceView(auth, space);
   if (podDsViewRes.isOk()) {
-    const needsFolder = scope === "files" || scope === "conversations";
-    if (!needsFolder || conversationFolderInternalId != null) {
```
matteotrab, line 117: I'm not sure why we needed to check the `conversationFolderInternalId` before, but is it correct to remove this check?

matteotrab (approved):

front/lib/api/actions/servers/pod_manager/build_pod_search_data_sources.ts
Fraggle: I looked closely at podDataSourceFilter and i thiiink it's fine.

PR #29984, opened 2026-08-04
[front] feat: surface the sandbox connect command in Poke

davidebbo:
## Description

Both conversations and pods can own a sandbox, but Poke gave no way to get the
`e2b sandbox connect <id>` command needed to attach to one — the provider id
only lived in the `sandboxes` table.

The provider id is now exposed on the two responses each page already loads (the
conversation config and the space details) as a `sandbox` field built by a new
`SandboxResource.toPokeJSON`. The lookup is a single indexed owner-link query,
so — unlike the pod databases listing — neither page wakes a sandbox just by
loading.

- **Pod page**: `Sandbox Status` and `Sandbox Connect` rows in the Overview. The
  connect cell is a `Copy command` button rather than the command text, which
  was far too wide for the table.
- **Conversation page**: no Overview table there, so a toolbar copy button next
  to `Sandbox Logs`, carrying the command and status in its tooltip. Disabled,
  with an explanatory tooltip, when the conversation has no sandbox.
- **Wake Sandbox plugin** for both resource types, applicable only while the
  sandbox is sleeping. It routes through `ensure{Conversation,Pod}SandboxReady`
  rather than the owner adapters, per the `/!\` warning in `lifecycle.ts`, so the
  GCS mount and egress forwarder come back up with it and the sandbox is
  actually usable once you connect. Any other status is refused rather than
  provisioned, since the ready helper would otherwise create a sandbox from
  nothing.

A plugin was chosen over a bespoke button + endpoint because it is the
established Poke path for state-changing actions: role-gated via
`requiredRoles: ["support"]`, recorded in `PluginRun` history, auto-registered
by the barrel exports, and `isApplicableTo` expresses the only-when-sleeping
rule declaratively.

## Tests

New functional tests on both modified endpoints (sandbox present, absent, and a
non-pod space) and on the wake plugin's guard logic — its applicability across
sleeping / running / absent / non-pod, and its refusal to wake a non-sleeping
sandbox, which is the path that keeps it from provisioning through the ready
helper. `SandboxFactory` gained a `createForPod` variant; it previously only
created conversation-owned sandboxes.

- 5 new endpoint tests + 6 new plugin tests pass
- the 62-test poke endpoint suite passes
- the 49 existing `sandbox_resource` + `lifecycle` tests pass
- `tsgo --noEmit` clean on `front`, `front-api` and `front-spa`

## Risk

Low. Read-only additive fields on two internal Poke endpoints, plus one new
opt-in plugin. The only behavior reachable by accident is the plugin, which is
super-user + `support`-role gated and refuses anything other than a sleeping
sandbox. Safe to roll back: no migration, no schema change.

One pre-existing export was widened — `EnsureSandboxReadyResult` in
`lifecycle.ts` is now exported so the shared wake helper can type its callback.
No behavior change.

## Deploy Plan

Standard deploy, no ordering constraints. `front` and `front-api` ship together;
the new `sandbox` response field is additive, so an older client simply ignores
it.

### DIFF
--- a/front-api/routes/poke/workspaces/[wId]/conversations/[cId]/config.test.ts
+++ b/front-api/routes/poke/workspaces/[wId]/conversations/[cId]/config.test.ts
@@ -1,0 +1,59 @@
+import { AgentConfigurationFactory } from "@app/tests/utils/AgentConfigurationFactory";
+import { ConversationFactory } from "@app/tests/utils/ConversationFactory";
+import { createPrivateApiMockRequest } from "@app/tests/utils/generic_private_api_tests";
+import { SandboxFactory } from "@app/tests/utils/SandboxFactory";
+import { honoApp } from "@front-api/app";
+import { describe, expect, it } from "vitest";
+
+async function setup() {
+  const { workspace, auth } = await createPrivateApiMockRequest({
+    isSuperUser: true,
+    role: "admin",
+  });
+
+  const agentConfiguration =
+    await AgentConfigurationFactory.createTestAgent(auth);
+  const conversation = await ConversationFactory.create(auth, {
+    agentConfigurationId: agentConfiguration.sId,
+    messagesCreatedAt: [new Date()],
+  });
+
+  return { auth, conversation, workspace };
+}
+
+function configUrl(workspaceId: string, conversationId: string) {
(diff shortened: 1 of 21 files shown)


fontanierh (approved):
LGTM

PR #29985, opened 2026-08-04
[front] feat: audit pod function publishes

davidebbo:
## Problem

Publishing a pod function puts executable code behind a callable endpoint on a Pod's shared runtime — authored by an agent, under a user's identity, reachable by anyone the function's identity policy allows. It was the most security-relevant mutation on the function path with no audit trail at all.

Follow-up to #29970, which covered tool calls made *by* functions.

## Approach

Unlike tool calls, this has no existing namespace to fold into, so it adds `pod_function.published` and a `pod_function` target type. Emitted from `publishSandboxFunction` (rather than the MCP tool handler) so it covers every caller and sits with the mutation it records.

Targets are `[workspace, space, pod_function]` — the pod as `space` so an admin can filter to everything published in one Pod.

## The identity-policy transition

`userIdentity` is declared in the function *source* and extracted at build time, not passed in by the caller. A re-publish can therefore **widen** who may call an existing function, and `updateContent` only pre-commits the policy when it *strengthens* (`sandbox_function_resource.ts`) — weakening lands with the new bundle. Going from `interactive_workspace_user_required` to `optional` turns a function that required a logged-in interactive user into one callable by any delegated caller.

So the event records both `user_identity` and `previous_user_identity` rather than just the new value.

That previous value is captured **before** `updateContent`, because `BaseResource.update` does `Object.assign(this, affectedRows[0].get())` — reading afterwards would report the incoming policy as the previous one. The existing re-publish test already exercised `workspace_user_required` → `optional`, so the assertion went there. Verified it fails on the regression: moving the capture below `updateContent` yields `previous_user_identity: "optional"` instead of `"workspace_user_required"`.

## Event shape

```json
{
  "actor": { "id": "IvcQRUlbYw", "name": "David Ebbo", "type": "user" },
  "action": "pod_function.published",
  "targets": [
    { "id": "BHWrN3HaUa", "name": "david", "type": "workspace" },
    { "id": "spc_a1b2c3", "name": "Team Weekly", "type": "space" },
    { "id": "sbxfn_d4e5f6", "name": "add-comment", "type": "pod_function" }
  ],
  "metadata": {
    "operation": "updated",
    "pod_function_slug": "add-comment",
    "user_identity": "optional",
    "previous_user_identity": "workspace_user_required",
    "file_id": "fil_g7h8i9"
  }
}
```

`previous_user_identity` is absent on a first publish rather than reported as the default — also asserted.

## Notes for review

- **Actor is the human, not the agent** (`[AUDIT2]`). Publishing always runs through the `publish` MCP tool under the user's `auth`. The authoring agent and conversation are *not* recorded — threading `runContext` into the business layer purely for audit metadata seemed like the wrong trade, but say the word if traceability back to the conversation is worth it.
- **`pod_function` naming** follows the `pod_*` metadata keys shipped in #29970. Worth noting the codebase is now inconsistent with itself: `project.joined` / `project.left` use "project" for the same underlying space that this calls a "pod".
- **Deletion is not audited.** There's no delete tool; functions only disappear via space deletion cascade (`deleteAllForSpace`), so the lifecycle event would fire during workspace scrubbing rather than on a user action. Left out deliberately.

## Registration

`pod_function.published` v1 is registered in the **test** WorkOS environment; `schema_versions.json` is committed. Production needs the same registration before deploy, as with #29970:

```
npx tsx admin/register_audit_log_schemas.ts --execute pod_function.published
```

Note `--changed` does **not** pick this up — it diffs via `git diff`, so a newly added (untracked) schema file is invisible to it. Register by action name.

## Testing

- `npx tsgo --noEmit` clean
- `lib/api/sandbox_functions` + `lib/api/audit` — 54/54, including the emit/schema drift scan
- Regression-verified the `previous_user_identity` assertion fails when the capture is moved after `updateContent`

### DIFF
--- a/front/admin/audit_log_schemas/pod_function.published.json
+++ b/front/admin/audit_log_schemas/pod_function.published.json
@@ -1,0 +1,21 @@
+{
+  "action": "pod_function.published",
+  "targets": [
+    {
+      "type": "workspace"
+    },
+    {
+      "type": "space"
+    },
+    {
+      "type": "pod_function"
+    }
+  ],
+  "metadata": {
+    "operation": "string",
+    "pod_function_slug": "string",
+    "user_identity": "string",
+    "previous_user_identity": "string",
+    "file_id": "string"
+  }
+}
--- a/front/admin/audit_log_schemas/pod_function.published.json
+++ b/front/admin/audit_log_schemas/pod_function.published.json
@@ -16,6 +16,7 @@
     "pod_function_slug": "string",
     "user_identity": "string",
     "previous_user_identity": "string",
-    "file_id": "string"
+    "file_id": "string",
+    "actor_type": "string"
   }
 }
--- a/front/lib/api/audit/schema_versions.json
+++ b/front/lib/api/audit/schema_versions.json
@@ -58,6 +58,7 @@
   "oauth.authorized": 3,
   "oauth.initiated": 3,
   "oauth.revoked": 2,
+  "pod_function.published": 1,
   "project.joined": 3,
   "project.left": 3,
   "sandbox_egress_policy.agent_requests_setting_updated": 1,
--- a/front/lib/api/audit/schema_versions.json
+++ b/front/lib/api/audit/schema_versions.json
@@ -58,7 +58,7 @@
   "oauth.authorized": 3,
   "oauth.initiated": 3,
   "oauth.revoked": 2,
-  "pod_function.published": 1,
+  "pod_function.published": 2,
   "project.joined": 3,
   "project.left": 3,
   "sandbox_egress_policy.agent_requests_setting_updated": 1,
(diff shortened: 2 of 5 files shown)


front/admin/audit_log_schemas/pod_function.published.json
```diff
@@ -0,0 +1,21 @@
...
+    {
+      "type": "pod_function"
+    }
+  ],
+  "metadata": {
+    "operation": "string",
+    "pod_function_slug": "string",
+    "user_identity": "string",
+    "previous_user_identity": "string",
+    "file_id": "string"
```
smb2268:
```suggestion
    "file_id": "string",
    "actor_type": "string"
```
The `emitAuditLog` function adds `actor_type` to every emit so we should include that in the schema - it may be getting overriden, I get a link expired error for the example link you provided so couldn't verify if it's in there!

smb2268 (approved):
I think we need to add `actor_type` here and then rerun the registration to bump the version #, but looks great otherwise!

front/admin/audit_log_schemas/pod_function.published.json
davidebbo: Yep, added

PR #29986, opened 2026-08-04
Admin governance for skills: Add warning if any editor will lose access to the skill

fabiencelier:
## Description

Closes https://github.com/dust-tt/tasks/issues/9736

 - Add a enw endpoint to check if all the users have access to a set of spaces. Only work if you are a member of the endpoint.
 - Display a warning if any of the admin does not have access to the skill due to a space restriction. the check is run at every change of editors or spaces.
 - does NOT prevent saving the skill so far.

## Tests

 - unit tests for endpoint
 - manual tests

Screenshot 2026-08-04 at 15 44 25
Screenshot 2026-08-04 at 15 44 16

## Risk

low.
We are exposing the members of a space but I think it is OK.

## Deploy Plan

deploy front

### DIFF
--- a/front/components/skill_builder/SkillSpaceRestrictionsContext.tsx
+++ b/front/components/skill_builder/SkillSpaceRestrictionsContext.tsx
@@ -6,16 +6,26 @@ import type {
   ReferencedSkillFormData,
   SkillBuilderFormData,
 } from "@app/components/skill_builder/SkillBuilderFormContext";
-import { useSpaceProjectsLookup } from "@app/lib/swr/spaces";
+import {
+  useSpaceProjectsLookup,
+  useSpacesAccessCheck,
+} from "@app/lib/swr/spaces";
 import type { SpaceType } from "@app/types/space";
 import type { ReactNode } from "react";
 import { createContext, useContext, useMemo } from "react";
 import { useWatch } from "react-hook-form";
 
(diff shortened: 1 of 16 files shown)


front/components/skill_builder/SkillSpaceRestrictionsContext.tsx
```diff
@@ -164,11 +177,56 @@ export function SkillSpaceRestrictionsProvider({
...
+  // never gets asked about a space it would reject.
+  const restrictedSpaceIds = useMemo(() => {
+    return nonGlobalSpacesWithRestrictions.map((space) => space.sId);
+  }, [nonGlobalSpacesWithRestrictions]);
+
+  const editorIds = useMemo(() => {
+    return (editors ?? []).map((editor) => editor.sId);
+  }, [editors]);
+
+  const { spacesAccess } = useSpacesAccessCheck({
```
matteotrab, line 190: Should we invalidate this query when we save the editors?

matteotrab (approved):

front/components/skill_builder/SkillSpaceRestrictionsContext.tsx
fabiencelier: it depends on editorIds and restrictedSpaceIds so it is recomputed on each change of these values

PR #27414, opened 2026-06-16
refactor(settings): remove user_settings_v2 flag and /me profile page

ElPicador:
## Description

The `user_settings_v2` flag is fully rolled out — this removes it and the legacy `/w/:wId/me` page it was guarding against.

- `UserMenu`: `Personal Settings` now unconditionally opens the `UserSettingsPopover` modal (the `onSelect` path); the `href` fallback and `hasFeature` call are removed.
- `feature_flags.ts`: `user_settings_v2` definition deleted.
- `adminRoutes.tsx`: `ProfilePage` import and `{ path: "me" }` route removed.
- Deleted files only used by the `/me` page: `ProfilePage.tsx`, `AccountSettings.tsx`, `PendingInvitationsTable.tsx`, `ProfileTriggersTab.tsx`.
- `NotificationPreferences` and `UserToolsTable` are kept — they're still used by `UserSettingsPopover`.

## Tests

Manually verified that clicking "Personal Settings" in the user menu opens the popover.

## Risk

Low — the old page is simply unreachable now. The popover was already the default for all users with the flag, which is everyone on Dust.

PR #29987, opened 2026-08-04
feat: bring back the triggers list in personal settings

adrsimon:
## Description

The Triggers tab in personal settings has been showing "Coming soon" since #27414 removed the `/me` profile page. This puts the list back.

Also moves `getTriggerDescription` out of `AgentTriggersTab` into `lib/utils/trigger_description.ts` so both surfaces share it.

Capture d’écran 2026-08-04 à 16 12 32

## Tests

Manual, no tests added.

## Risk

Low, UI only. The `/api/w/:wId/me/triggers` endpoint already exists and is unchanged. Follow-up: the delete dialog is still duplicated with `AgentTriggersTab`.

## Deploy Plan

Ship it.

### DIFF
--- a/front/components/me/UserTriggersTable.tsx
+++ b/front/components/me/UserTriggersTable.tsx
@@ -1,0 +1,269 @@
+import { useSendNotification } from "@app/hooks/useNotification";
+import { useDeleteTrigger, useUserTriggers } from "@app/lib/swr/agent_triggers";
+import { getAgentBuilderRoute } from "@app/lib/utils/router";
+import { getTriggerDescription } from "@app/lib/utils/trigger_description";
+import type { GetUserTriggersResponseBody } from "@app/types/api/assistant/configuration/triggers";
+import { isGlobalAgentId } from "@app/types/assistant/assistant";
+import type { TriggerStatus } from "@app/types/assistant/triggers";
+import { assertNeverAndIgnore } from "@app/types/shared/utils/assert_never";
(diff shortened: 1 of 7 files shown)


front/lib/utils/trigger_description.ts
```diff
@@ -0,0 +1,22 @@
+import { describeScheduleConfig } from "@app/lib/utils/schedule_description";
+import type { TriggerType } from "@app/types/assistant/triggers";
+import { assertNeverAndIgnore } from "@app/types/shared/utils/assert_never";
+
+export function getTriggerDescription(trigger: TriggerType): string {
+  switch (trigger.kind) {
+    case "schedule":
+      try {
+        // cronstrue throws on expressions it cannot parse.
+        return `Runs ${describeScheduleConfig(trigger.configuration)}.`;
```
flvndvd: Ideally, the try/catch block lives in `describeScheduleConfig`.

front/components/me/UserTriggersTable.tsx
```diff
@@ -0,0 +1,269 @@
...
+      return "success";
+    case "disabled":
+      return "primary";
+    case "relocating":
+      return "info";
+    case "downgraded":
+      return "warning";
+    default:
+      assertNeverAndIgnore(status);
+      return "primary";
```
flvndvd: It feels missleaing that we both assert and yet return a color.

flvndvd (approved):
🙏

PR #29989, opened 2026-08-04
Add: ability to put Frame as tabs in a pod.

Fraggle:
## Description

Lets Pod editors promote any Frame file into a persistent Pod tab, visible to all members. Each tab stores a `path`, `title`, and `icon`; ordering is controlled by `frameTabsNavOrder`, an interleaved list of system tab ids and frame paths.

- `frameTabs` + `frameTabsNavOrder` JSONB columns on `project_metadata` (pre-deploy SQL migrations included). PATCH `project_metadata` validates both fields together via `validatePodFrameTabs()`, gated by the `pod_frame_tabs` FF.
- New `PodFrameTab` type module (`pod_frame_tab.ts`) with Zod schemas, nav-order helpers, URL encoding (`#frame/<encoded-path>`), and constants (`MAX_POD_FRAME_TABS`, `DEFAULT_POD_FRAME_TAB_ICON`).
- `PodFrameTabContent` renders the frame via `AuthenticatedVisualizationActionIframe`. `PodFrameTabButton` surfaces in `FrameRenderer` and `PodFrameSheet` toolbars. `PodFileExplorer` gains an "Add/Remove as Pod tab" context menu item for Frame files.
- `EditPodFrameTabDialog` lets editors rename, change icon (via `IconPicker`), reorder (left/right arrows), and remove a tab. Opens on re-click of the active frame tab in `PodPage`.
- `usePodTabs` / `useScopedUIPreferences` extended to accept `frame:<path>` tab values alongside system tabs.

file-08714ce982fcf5ffb74ead30b5a9b088

file-ab8ae0d7ed1a33c0573e89f6c93cd5cf

file-76040b41b381e23f43ebddc0120349c8

## Tests

- Updated `project_metadata.test.ts` and `project_metadata_resource.test.ts` to assert `frameTabs: []` and the default `frameTabsNavOrder` on fresh records.
- Updated `project_metadata.test.ts` GET snapshot to include both new fields.
- Tested locally: add/remove/reorder frame tabs, icon picker, tab navigation, flag-off fallback to conversations tab.

## Risk

Low. Both columns default to `[]` so no backfill is needed and existing rows are unaffected. The feature is fully gated by the `pod_frame_tabs` FF; no impact until the flag is enabled. `migrateLegacyFrameTabsNavOrder` handles forward compat for rows written before the nav-order column existed.

## Deploy Plan

1. Run pre-deploy
2. Deploy front
3. Enable `pod_frame_tabs` FF per workspace.

### DIFF
--- a/front/components/pages/pod/PodPage.tsx
+++ b/front/components/pages/pod/PodPage.tsx
@@ -1,16 +1,31 @@
+import { EditPodFrameTabDialog } from "@app/components/pod/files/EditPodFrameTabDialog";
 import { PodHeaderActions } from "@app/components/pod/PodHeaderActions";
 import { PodPageContent } from "@app/components/pod/PodPageContent";
+import { getIcon } from "@app/components/resources/resources_icons";
 import { useActivePodId } from "@app/hooks/useActivePodId";
 import { useScopedPodUiPreferences } from "@app/hooks/useScopedUIPreferences";
 import {
   DEFAULT_POD_UI_PREFERENCES,
-  type PodTab,
+  isValidPodTabValue,
   usePodTabs,
 } from "@app/hooks/useSpaceProjectTabs";
-import { useAuth, useWorkspace } from "@app/lib/auth/AuthContext";
+import {
+  useAuth,
+  useFeatureFlags,
+  useWorkspace,
+} from "@app/lib/auth/AuthContext";
 import { useSpaceInfo } from "@app/lib/swr/spaces";
 import { useIsMobile } from "@app/lib/swr/useIsMobile";
 import { classNames } from "@app/lib/utils";
+import type { PodFrameTab } from "@app/types/pod_frame_tab";
+import {
+  buildPodNavItemsBeforeSettings,
+  makePodFrameTabValue,
+  normalizeTabsOrder,
+  parsePodFrameTabPath,
+  sortPodFrameTabs,
+} from "@app/types/pod_frame_tab";
+import { assertNever } from "@app/types/shared/utils/assert_never";
 import {
   CheckCircle,
   CloudArrowLeftRight,
@@ -22,11 +37,36 @@ import {
   Settings01,
   Spinner,
 } from "@dust-tt/sparkle";
+import { useEffect, useMemo, useState } from "react";
+
+const SYSTEM_TAB_TRIGGERS = {
+  conversations: {
+    label: "Conversations",
+    icon: MessageChatSquare,
+  },
+  tasks: {
+    label: "Tasks",
+    icon: CheckCircle,
+  },
+  files: {
+    label: "Files",
+    icon: Folder,
+  },
+  connected_data: {
+    label: "Connected Data",
+    icon: CloudArrowLeftRight,
+  },
+} as const;
 
 export function PodPage() {
   const owner = useWorkspace();
   const { user } = useAuth();
+  const { hasFeature } = useFeatureFlags();
   const podId = useActivePodId();
+  const hasFrameTabs = hasFeature("pod_frame_tabs");
+  const [editingFrameTab, setEditingFrameTab] = useState<PodFrameTab | null>(
+    null
+  );
 
   const {
     spaceInfo: podInfo,
@@ -54,6 +94,44 @@ export function PodPage() {
     setPodUiPreferences,
     isAdminControlled: podInfo?.isAdminControlled,
   });
+
+  const frameTabs = useMemo(
+    () => (hasFrameTabs ? sortPodFrameTabs(podInfo?.frameTabs ?? []) : []),
+    [hasFrameTabs, podInfo?.frameTabs]
+  );
+
+  const tabsOrder = useMemo(
+    () =>
+      hasFrameTabs
+        ? normalizeTabsOrder(
+            podInfo?.tabsOrder ?? [],
+            frameTabs.map((tab) => tab.path)
+          )
+        : [],
+    [frameTabs, hasFrameTabs, podInfo?.tabsOrder]
+  );
+
+  const includeConnectedData = !!podInfo?.isAdminControlled;
+
+  const navItemsBeforeSettings = useMemo(
+    () =>
+      buildPodNavItemsBeforeSettings(frameTabs, tabsOrder, {
+        includeConnectedData,
+      }),
+    [frameTabs, tabsOrder, includeConnectedData]
+  );
+
+  // Drop frame-tab selection when the flag is off or the tab was removed
+  // (including restored preference pointing at a deleted tab).
+  useEffect(() => {
+    const framePath = parsePodFrameTabPath(currentTab);
+    if (!framePath) {
+      return;
+    }
+    if (!hasFrameTabs || !frameTabs.some((tab) => tab.path === framePath)) {
+      handleTabChange("conversations");
+    }
+  }, [currentTab, frameTabs, handleTabChange, hasFrameTabs]);
 
   if (isPodsInfoLoading) {
     return (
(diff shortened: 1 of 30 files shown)


Fraggle:

- #29969
    - #29983
        - #29989 ◀

front/lib/resources/storage/models/project_metadata.ts
```diff
@@ -77,6 +85,16 @@ ProjectMetadataModel.init(
...
+    frameTabs: {
+      type: DataTypes.JSONB,
+      allowNull: false,
+      defaultValue: [],
+    },
+    frameTabsNavOrder: {
+      type: DataTypes.JSONB,
+      allowNull: false,
+      defaultValue: [],
+    },
```
id13, line 93: carambaaaaaa

id13: why do we need this ?

Fraggle: First: You can store frames to be tabs in the pod
Second: Order of all tabs.

id13: looks like only one of the two is needed here no ? 
and among the one that remains: why can't it be an array of strings ?

Fraggle: Will take a look if I can reduce to one.
At least one is an array of object as it needs to contains the custsom title and the icon.

Fraggle: I can't reduce to 1 without making it worth (i think) but I turned the ordering one is an array of string :)

id13: fair enough thanks for digging 👍

front/components/pages/pod/PodPage.tsx
```diff
@@ -55,6 +95,57 @@ export function PodPage() {
...
+  // Drop frame-tab selection when the flag is off or the tab was removed.
+  useEffect(() => {
+    const framePath = parsePodFrameTabPath(currentTab);
+    if (!framePath) {
+      return;
+    }
+    if (!hasFrameTabs || !frameTabs.some((tab) => tab.path === framePath)) {
+      handleTabChange("conversations");
+    }
+  }, [currentTab, frameTabs, handleTabChange, hasFrameTabs]);
```
matteotrab, line 134: Can we just navigate to the conversation tab when deleting a frame tab? No need for a useEffect

front/components/pages/pod/PodPage.tsx
```diff
@@ -55,6 +95,57 @@ export function PodPage() {
...
+      (tab) => tab.path === editingFrameTab.path
+    );
+    if (!refreshed) {
+      setEditingFrameTab(null);
+      return;
+    }
+    if (refreshed !== editingFrameTab) {
+      setEditingFrameTab(refreshed);
+    }
+  }, [editingFrameTab, frameTabs]);
```
matteotrab: I'm not sure I understand what this useEffect is supposed to do. After removing a tab we can directly set the editingFrameTab to null and I don't see when we can end up in the case where `refreshed !== editingFrameTab`

front/components/pod/files/EditPodFrameTabDialog.tsx
```diff
@@ -0,0 +1,236 @@
...
+  const canMoveLeft = isEditor && tabIndex > 0;
+  const canMoveRight =
+    isEditor && tabIndex >= 0 && tabIndex < navItems.length - 1;
+
+  useEffect(() => {
+    if (tab) {
+      setTitle(tab.title);
+      setIcon(tab.icon);
+    }
+  }, [tab]);
```
matteotrab: We can give pass a key to this component instead

front/components/pod/files/EditPodFrameTabDialog.tsx
```diff
@@ -0,0 +1,236 @@
...
+                />
+                <Button
+                  variant="ghost"
+                  icon={ArrowRight}
+                  tooltip="Move right"
+                  disabled={!canMoveRight || isMoving || isSaving}
+                  onClick={() => void handleMove("right")}
+                />
+              </div>
+            )}
```
matteotrab, line 201: I found these arrows a bit confusing when testing, not sure what's the best ux here though

front/components/pages/pod/PodPage.tsx
```diff
@@ -92,23 +187,49 @@ export function PodPage() {
...
+                      key={item.tab.path}
+                      value={tabValue}
+                      icon={getIcon(item.tab.icon)}
+                      onPointerDown={(e) => {
+                        // Read data-state before Radix activates the tab so we only
+                        // open the editor on a re-click of the already-active tab.
+                        if (
+                          podInfo.isEditor &&
+                          e.currentTarget.getAttribute("data-state") ===
+                            "active"
```
matteotrab: Checking the currentTab directly might be simpler

front/components/pod/files/PodFrameSheet.tsx
```diff
@@ -102,6 +110,18 @@ export function PodFrameSheet({
...
                 />
+                {hasFrameTabs && (
+                  <PodFrameTabButton
+                    owner={owner}
+                    spaceId={podId}
+                    frameTabs={frameTabs}
+                    tabsOrder={tabsOrder}
+                    isEditor={isEditor}
+                    framePath={framePath}
+                    fileName={fileName}
```
matteotrab, line 121: Sometimes the buttons are not correctly visible

Image

front/components/pod/PodFrameTabContent.tsx
```diff
@@ -0,0 +1,91 @@
...
+    owner,
+    podId: podInfo.sId,
+  });
+
+  const frameFile = useMemo(() => {
+    return (
+      podFiles.find((f) => !f.isDirectory && f.path === tab.path && f.fileId) ??
+      null
+    );
+  }, [podFiles, tab.path]);
```
matteotrab: I might be wrong, but don't we have an endpoint to get a file given its path?

front/lib/api/projects/frame_tabs.ts
```diff
@@ -0,0 +1,97 @@
...
+
+  const seenPaths = new Set<string>();
+  const normalized: PodFrameTab[] = [];
+  const pathRemap = new Map<string, string>();
+
+  for (const tab of frameTabs) {
+    const normalizedPath = resolveCanonicalScopedPath(tab.path, {
+      conversationId: null,
+      spaceId: space.sId,
+    });
```
matteotrab, line 48: Isn't `tab.path` already what we need?

front/types/pod_frame_tab.ts
```diff
@@ -0,0 +1,234 @@
...
+  return tabsOrder.filter(
+    (id) => id !== "connected_data" || includeConnectedData
+  );
+}
+
+/**
+ * One-time reconstruction for tabs that still store absolute `order` indices
+ * (pre-tabsOrder shape). Prefer `tabsOrder` when present.
+ */
+export function migrateLegacyTabsOrder(
```
matteotrab: How can we have a legacy order if it has been introduced here?

matteotrab (commented):
Nice! Left a few comments, I'll let @id13 review the data model

matteotrab (approved):

front/components/pages/pod/PodPage.tsx
Fraggle: It's also the case if you come back to the pod (as your last tab nav is restored). So you might not be the one deleting it.

front/components/pod/files/EditPodFrameTabDialog.tsx
Fraggle: Drag & drop is the best but that's a good v0 imho.

front/components/pod/files/PodFrameSheet.tsx
Fraggle: What do you suggest ?

front/types/pod_frame_tab.ts
Fraggle: It's due to a refactor, shouldn't be there.

front/components/pod/files/PodFrameSheet.tsx
matteotrab: Either showing the icons only when there is not enough space or make sure that we can scroll

Fraggle: I switched the 3 buttons to icons only and use highlight to show when they are ON.
image

PR #29990, opened 2026-08-04
Preempt sandbox cleanup for capacity reaping

fontanierh:
## Description

Large image cleanup sweeps kept the serial sandbox reaper in paused-sandbox maintenance for about 40 minutes. Scheduled overlaps were skipped, so newly stale running sandboxes accumulated until the sweep finished.

Recheck and fully drain awake kill requests and stale running sandboxes after every maintenance batch. Preserve maintenance cursors across those priority checks. The workflow intentionally has no replay compatibility path because the open executions will be terminated during deployment.

## Tests

Added workflow coverage for priority preemption, cursor preservation, and priority batch-cap handling. Ran the sandbox reaper workflow and activity test suites locally.

## Risk

The runtime behavior change is low risk, but the Temporal deployment requires a coordinated cutover in both clusters. Existing sandbox reaper histories are incompatible with the new command sequence and must not continue on new workers.

## Deploy Plan

1. Pause the sandbox reaper schedules in both US and EU.
2. Before rolling out the new workers, terminate every open sandbox reaper workflow in both clusters.
3. Deploy the new workers.
4. Immediately start one sandbox reaper workflow in each cluster and verify that capacity phases run between maintenance batches.
5. Resume both schedules and verify that concurrency continues to drain.

For rollback, pause both schedules and terminate every open reaper started on the new code before deploying the previous worker version, then start fresh reapers and resume the schedules.

### DIFF
--- a/front/temporal/sandbox_reaper/workflows.ts
+++ b/front/temporal/sandbox_reaper/workflows.ts
@@ -3,7 +3,7 @@ import type {
   ReaperCursor,
   ReaperPhase,
 } from "@app/temporal/sandbox_reaper/activities";
-import { log, patched, proxyActivities } from "@temporalio/workflow";
+import { log, proxyActivities } from "@temporalio/workflow";
 
 const { reapSandboxPhaseActivity } = proxyActivities<typeof activities>({
   startToCloseTimeout: "10 minutes",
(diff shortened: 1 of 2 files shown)


front/temporal/sandbox_reaper/workflows.ts
```diff
@@ -13,48 +13,96 @@ const { reapSandboxPhaseActivity } = proxyActivities<typeof activities>({
...
+
+  for (const phase of MAINTENANCE_PHASES) {
     let cursor: ReaperCursor | null = null;
-    let processedBatches = 0;
 
-    while (processedBatches < MAX_BATCHES_PER_PHASE) {
+    for (
+      let processedBatches = 0;
+      processedBatches < MAX_BATCHES_PER_PHASE;
+      processedBatches += 1
```
davidebbo, line 88: I have trouble wrapping my head around having this loop here, and also having a similar one in `drainPhase`, which is called within this loop. I may be misreading.

davidebbo (approved):
LGTM, though the nested structure is slightly confusing to me :)

front/temporal/sandbox_reaper/workflows.ts
fontanierh: will address separately, i need to run a /kill first

fontanierh: thanks !

PR #29992, opened 2026-08-04
Normalize credit attribution through model input

flvndvd:
## Description

Our attribution is cache-naive, so its raw total can exceed the credits actually billed. Exposing both the raw estimate and a cache savings adjustment forces users to reconcile two competing totals and can undermine trust in the breakdown.

This change makes the breakdown additive with the authoritative bill. Raw attribution evidence remains unchanged in storage. At read time, **only model input rows** absorb the billing difference because input is where cached conversation
context is reused. Output, reasoning, and tool attribution remain unchanged.

If input-only reconciliation is impossible, the detailed breakdown is withheld instead of presenting misleading values.

### DIFF
--- a/front-api/routes/w/[wId]/auth-context.ts
+++ b/front-api/routes/w/[wId]/auth-context.ts
@@ -1,4 +1,3 @@
-import config from "@app/lib/api/config";
 import { getWorkspaceRegionRedirect } from "@app/lib/api/regions/lookup";
 import { Authenticator, getFeatureFlags } from "@app/lib/auth";
 import { isWorkspaceEligibleForTrial } from "@app/lib/plans/trial";
@@ -79,7 +78,7 @@ app.get(
       isManager: auth.isManager(),
       featureFlags,
       ...(isEligibleForTrial !== undefined && { isEligibleForTrial }),
-      vizUrl: config.getVizPublicUrl(),
+      vizUrl: "http://localhost:3007", // config.getVizPublicUrl(),
       providersHealth: auth.providersHealth(),
       workspacePermissions,
     });
--- a/front-api/routes/w/[wId]/auth-context.ts
+++ b/front-api/routes/w/[wId]/auth-context.ts
@@ -78,7 +78,7 @@ app.get(
       isManager: auth.isManager(),
       featureFlags,
       ...(isEligibleForTrial !== undefined && { isEligibleForTrial }),
-      vizUrl: "http://localhost:3007", // config.getVizPublicUrl(),
+      vizUrl: config.getVizPublicUrl(),
       providersHealth: auth.providersHealth(),
       workspacePermissions,
     });
--- a/front/lib/api/assistant/agent_message_consumption_attribution/message_details.ts
+++ b/front/lib/api/assistant/agent_message_consumption_attribution/message_details.ts
@@ -1,0 +1,437 @@
+import { isToolExecutionStatusFinal } from "@app/lib/actions/statuses";
+import { getToolAggregateDisplayLabel } from "@app/lib/actions/tool_display_labels";
+import { AGENT_MESSAGE_CONSUMPTION_ATTRIBUTION_VERSION } from "@app/lib/api/assistant/agent_message_consumption_attribution/attribution_builder";
+import { getModelConfigByModelId } from "@app/lib/llms/model_configurations";
+import type { AgentMCPActionResource } from "@app/lib/resources/agent_mcp_action_resource";
+import type { AgentMessageConsumptionItemResource } from "@app/lib/resources/agent_message_consumption_item_resource";
+import type { RunResource } from "@app/lib/resources/run_resource";
+import type { AgentMCPActionType } from "@app/types/actions";
+import type {
+  AgentMessageConsumptionDetails,
+  AgentMessageConsumptionModelDetails,
+} from "@app/types/assistant/agent_message_consumption";
+import type { ModelId } from "@app/types/shared/model_id";
+
+const MICRO_CREDITS_PER_CREDIT = 1_000_000;
+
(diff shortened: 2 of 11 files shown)


front/lib/api/assistant/agent_message_consumption_attribution/message_details.ts
```diff
@@ -0,0 +1,437 @@
...
+ * new result into the next model input. We keep every non-input attribution unchanged. The model's
+ * ordinary `input` bucket contains reused conversation context, so it is the single explicit
+ * reconciliation seam.
+ *
+ * Multiple input rows share the reconciled input total in proportion to their cache-naive cost.
+ * Allocation happens in integer micro-credits and assigns rounding remainders deterministically in
+ * input order. If the non-input evidence alone exceeds the bill, no input-only reconciliation is
+ * possible and the detailed breakdown is withheld.
+ */
+function reconcileInputCredits({
```
flvndvd, line 45: New function, all the other have been moved around.

front/lib/api/assistant/agent_message_consumption_attribution/message_details.ts
```diff
@@ -0,0 +1,437 @@
...
+        item,
+        item.itemType === "input"
+          ? (reconciledInputCreditAmountByItem.get(item) ?? 0)
+          : item.grossAttributedCreditAmountMicro,
+      ])
+    ),
+  };
+}
+
+function buildConsumptionTotals({
```
flvndvd, line 135: Moved from `read.ts`.

front/lib/api/assistant/agent_message_consumption_attribution/message_details.ts
```diff
@@ -0,0 +1,437 @@
...
+import type { AgentMCPActionType } from "@app/types/actions";
+import type {
+  AgentMessageConsumptionDetails,
+  AgentMessageConsumptionModelDetails,
+} from "@app/types/assistant/agent_message_consumption";
+import type { ModelId } from "@app/types/shared/model_id";
+
+const MICRO_CREDITS_PER_CREDIT = 1_000_000;
+
+type RunUsages = Awaited<ReturnType<typeof RunResource.listRunUsagesForRuns>>;
```
aubin-tchoi: nit: kinda hard to tell what's the type behind this without checking, isn't it just `RunUsageWithRunKeyType`?

front/lib/api/assistant/agent_message_consumption_attribution/message_details.ts
```diff
@@ -0,0 +1,437 @@
...
+        }
+      : null;
+  }
+
+  // Preserve the relative weight of the cache-naive input rows.
+  const inputAllocations = inputItems.map((item, index) => {
+    const exactMicro =
+      (item.grossAttributedCreditAmountMicro *
+        reconciledInputCreditAmountMicro) /
+      grossInputCreditAmountMicro;
```
aubin-tchoi: nit: would be clearer with an intermediate variable for the ratio the current input represents (then multiplied by the total number of credits to reconcile)

front/lib/api/assistant/agent_message_consumption_attribution/message_details.ts
```diff
@@ -0,0 +1,437 @@
...
+      index,
+      floorMicro,
+      fractionalMicro: exactMicro - floorMicro,
+    };
+  });
+  const allocatedFloorMicro = inputAllocations.reduce(
+    (total, allocation) => total + allocation.floorMicro,
+    0
+  );
+  const remainderMicro = reconciledInputCreditAmountMicro - allocatedFloorMicro;
```
aubin-tchoi, line 103: we don't have to take the floor here?

aubin-tchoi: maybe the round is more accurate actually

aubin-tchoi (approved):
LGTM

front/lib/api/assistant/agent_message_consumption_attribution/message_details.ts
flvndvd: Yeah I concur, will export it.

front/types/assistant/agent_message_consumption.ts
```diff
@@ -16,15 +20,22 @@ export type AgentMessageConsumptionToolDetails = {
...
+export type AgentMessageConsumptionModelDetails = {
+  providerId: ModelProviderIdType;
+  modelId: ModelIdType;
+  displayName: string;
+  attributedCredits: number;
+};
+
 export type AgentMessageConsumptionDetails = {
   attributionVersion: number;
-  grossAttributedCredits: number;
```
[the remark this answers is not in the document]
flvndvd: All good, it's gated behind an FF and only availabe on the private API.

front/lib/api/assistant/agent_message_consumption_attribution/message_details.ts
flvndvd: Using `round` can overshoot, no? For example, splitting 2 micro-credits equally across 3 rows gives 0.667 each. Rounding produces 1 + 1 + 1 = 3.

front-api/routes/w/[wId]/auth-context.ts
```diff
@@ -79,7 +78,7 @@ app.get(
       isManager: auth.isManager(),
       featureFlags,
       ...(isEligibleForTrial !== undefined && { isEligibleForTrial }),
-      vizUrl: config.getVizPublicUrl(),
+      vizUrl: "http://localhost:3007", // config.getVizPublicUrl(),
```
Comment (sfriquet) [inline front-api/routes/w/[wId]/auth-context.ts]: 👀

Comment (flvndvd) [reply front-api/routes/w/[wId]/auth-context.ts]: Oh good catch 🙈

Comment (flvndvd) [reply front-api/routes/w/[wId]/auth-context.ts]: My last commit was def broken.

PR #29994, opened 2026-08-04
Add conversation credit breakdown

flvndvd:
## Description

This PR is based on #29992, which introduced a single additive credit breakdown at the message level by reconciling cache effects exclusively through model input rows.

See design [here](https://www.figma.com/design/wJJMfVF6bluurSKfrEuysc/Product---WIP?node-id=12710-33268&p=f&m=dev).

This PR applies the same accounting model at the conversation level. The billed total for messages belonging directly to the conversation remains authoritative, while detailed attribution explains how that total is distributed across agent
work, tools, models, and agents. All billed executions are included, even when a message was retried or superseded.

> [!NOTE]
> Run-agent child conversations are deliberately excluded from this first version. Their usage is billed on their own
agent messages, but including it here would currently require a recursive read-time traversal across conversations. Will need to think a bit more before sending a follow-up PR to establish a cleaner lineage model.

The endpoint keeps the billed total available even when a detailed explanation cannot be produced. If any billed message is missing complete attribution for the active version, the API returns the total and withholds the breakdown rather than presenting a partial or misleading view. In-progress messages are temporarily excluded and the panel refreshes after message completion. Tools with no attributed credits are omitted.

On the UI side, this adds a Credit usage action next to Files and a new conversation drawer variant. The drawer displays total billed credits, agent work and context, the most significant tool usage, and model usage. An additional per-agent breakdown appears only when multiple agents contributed directly to the conversation.

The drawer follows existing conversation panel conventions and includes loading, empty, error, and unavailable-breakdown states. The experience remains behind the existing conversation consumption feature flag.

With one agent:

image

With 1+ agent:

image


front-api/routes/w/[wId]/assistant/conversations/[cId]/consumption.ts
```diff
@@ -0,0 +1,86 @@
...
+      return apiError(ctx, {
+        status_code: 404,
+        api_error: {
+          type: "conversation_not_found",
+          message: "Conversation not found.",
+        },
+      });
+    }
+
+    return ctx.json(await getConversationConsumption(auth, { conversation }));
```
Comment (aubin-tchoi) [inline front-api/routes/w/[wId]/assistant/conversations/[cId]/consumption.ts]: nit: fine since we only have 1 await but would better to have the await on its own statement otherwise

front/components/assistant/conversation/credits_panel/ConversationCreditUsageBreakdown.tsx
```diff
@@ -0,0 +1,247 @@
...
+
+function CreditBreakdownCard({
+  description,
+  icon,
+  label,
+  labelBadge,
+  value,
+}: CreditBreakdownCardProps) {
+  return (
+    <div className="flex min-w-0 flex-col justify-center gap-2 rounded-xl border border-border bg-background p-4">
```
aubin-tchoi, line 41: would put a 16px border radius here if this is meant to be a card

front/components/assistant/conversation/credits_panel/ConversationCreditUsageBreakdown.tsx
```diff
@@ -0,0 +1,247 @@
...
+          icon={getActionStepIcon(tool)}
+        />
+      ))}
+      {remainingTools.length > 0 && (
+        <CreditBreakdownCard
+          label="Other tools"
+          labelBadge={String(remainingTools.length)}
+          description={toolUsageLabel(remainingToolCallCount)}
+          value={remainingToolCredits}
+          icon={ShapesPlus}
```
aubin-tchoi: nit:
```suggestion
          icon={InternalActionIcons.ToolsIcon}
```
to show we're using from the same icon registry

front/components/assistant/conversation/credits_panel/ConversationCreditUsageBreakdown.tsx
```diff
@@ -0,0 +1,247 @@
...
+    </div>
+  );
+}
+
+interface ModelRowProps {
+  isDark: boolean;
+  model: ConversationConsumptionModelDetails;
+}
+
+function ModelRow({ isDark, model }: ModelRowProps) {
```
aubin-tchoi, line 127: nit: I feel like we don't need to pass down `isDark`

front/hooks/conversations/useConversationConsumption.ts
```diff
@@ -0,0 +1,61 @@
...
+import { useFetcher, useSWRWithDefaults } from "@app/lib/swr/swr";
+import type { ConversationConsumptionResponse } from "@app/types/assistant/conversation_consumption";
+import { useEffect, useRef } from "react";
+import type { Fetcher } from "swr";
+
+// The completion event can arrive before credit accounting is persisted.
+// Refresh once after a short delay so the latest breakdown is available.
+const MUTATE_DELAY_MS = 3000;
+
+export function useConversationConsumption({
```
aubin-tchoi, line 11: in the current version if I open the panel during the generation we won't see anything until it completes, right?
could be nice to see items being added progressively

aubin-tchoi (approved):
LGTM

sfriquet (approved):

front/components/assistant/conversation/credits_panel/ConversationCreditUsageBreakdown.tsx
flvndvd: I wanted to avoid calling `useTheme` hook everytime.

front/hooks/conversations/useConversationConsumption.ts
flvndvd: Yes that is correct only finalized messages are displayed in the breakdown, this is a limitation of current bulling that does not compute anything until message reaches and "end" state. And billed credits is the authority here to "normalize" at the conversation level.

PR #29995, opened 2026-08-04
fix: ship prestop.sh in the front-api image

adrsimon:
## Description

`preStop` hook is `/bin/sh -c admin/prestop.sh`. The front-api stage sets `WORKDIR /app/front-api`, but the script only lands at `/app/front/admin/prestop.sh`, so the hook exits 127 and `FailedPreStopHook` fires on every terminating pod.

Means `runPreStop()` never runs. Copies the single file (not all of `front/admin`, which is mostly dev scripts) next to the existing migrations COPY that works around the same WORKDIR mismatch.

## Tests

Reproduced in a prod pod: `sh -c "admin/prestop.sh"` → `127, not found`.

## Risk

Low, one COPY. Worst case it stays broken as today. Rollback is a revert.

Note it *enables* a code path that's been dead, so first deploy after this is the first time pods actually drain: ~70s minimum termination (10s LB + 60s draining), up to 130s.

## Deploy Plan

Normal deploy.


id13 (approved):
🦾

PR #29996, opened 2026-08-04
[Activation] Add the uiView column in activation_pods table

xebeohp:
## Description
Gate the activation pod UI simplification changes around a `isCompactUIView` flag, which is added as a column of the `activation_pods` table.

This allows us to change the UI in only the activation pod without a bunch of if activation statements throughout the code. Also easily allows additional UI changes to be added on top as we continue to experiment with activation UI.

## Risk
Low risk, nothing depends on it in the code so far.

## Deploy Plan
Follow pre-deploy migration steps.


xebeohp:

- #29996 ◀
    - #30109

front/lib/models/activation/activation_pod.ts
```diff
@@ -34,6 +38,11 @@ ActivationPodModel.init(
       allowNull: false,
       defaultValue: DataTypes.NOW,
     },
+    uiView: {
```
flvndvd: What's the rationale to use string over a boolean here?

xebeohp: You're right, a boolean is better than a string. Making the changes now.

front/lib/models/activation/activation_pod.ts
```diff
@@ -20,6 +20,8 @@ export class ActivationPodModel extends WorkspaceAwareModel<ActivationPodModel>
   declare userId: ForeignKey<UserModel["id"]>;
   // The Pod's activation trigger. Null until provisioned.
   declare triggerId: ForeignKey<TriggerModel["id"]> | null;
+  // Whether the Pod uses the compact UI variant.
+  declare compactUIView: CreationOptional<boolean>;
```
flvndvd: We usually go with `is` to reflect the boolean.

flvndvd (approved):
LGTM, once column has been renamed.

PR #29997, opened 2026-08-04
[front] fix: prevent duplicate MCP server view names

aubin-tchoi:
## Description

Closes https://github.com/dust-tt/tasks/issues/9068

Users could create multiple MCP server views with different names that would resolve to the same one at runtime once cropped (we need to crop them to pass them to the model).

Server creation now checks the actual prefix that will be used for function names, taking in account case and crop.

## Tests

- Added tests.

## Risk

- Low.

## Deploy Plan

- Deploy front.


Fraggle (approved):
LGTM

PR #29999, opened 2026-08-04
[front] feat: add generated webhook filter validation

aubin-tchoi:
## Description

Generated webhook filters can be syntactically valid but never match when an operator is incompatible with the selected event schema, such as `has-any` on a string field.

This PR adds a validation step on the generated expressions. Checks the expression against the event JSON Schema to flag filters that will never match (assumes the JSON schemas are valid). Invalid filters get one retry with the validation error.

## Tests

- Added tests.

## Risk

Low. At worst we do a few more LLM calls.

## Deploy Plan

Deploy front.


adrsimon (approved):
lgtm, can't wait to test it

PR #30002, opened 2026-08-04
[front] enh: add hint on user memory read tool usage

aubin-tchoi:
## Description

Agents may not realize the user memory read tool can be invoked without constructing an input. This PR updates the User Memory system skill instructions to tell agents to call the read tool directly with no arguments when prior user context could change the answer.

This is a hack to bypass tool search for this specific case: we expect many conversations to require reading memories, less to require writing. Yet we'd have to eagerly load both tools for the sake of the conversations that only read them. This PR hopes to allow not requiring the agent to run a tool search for them, keeping the cost to a minimum.

## Tests

- Tested locally.

## Risk

- Low.

## Deploy Plan

- Deploy front.


Inesvvv:
LGTM!

PR #30003, opened 2026-08-04
Avoid approval when creating Pod conversations

frankaloia:
## Description

* Making create_conversation in a pod low stakes
* This is low blast radius and we make run_agent never_ask which is the same thing
* It helps to make this never_ask for activation so we can create conversations without blocking users

## Tests

* Local

## Risk

* Low

## Deploy Plan

* Deploy front


smb2268 (approved):
I think you also have to update the snapshot but then lgtm!

PR #30004, opened 2026-08-04
perf: index spaces by id in baseFetchWithAuthorization

adrsimon:
## Description

`baseFetchWithAuthorization` resolved each blob's space with `spaces.find(...)`, so the cost was O(blobs × spaces). Both sides of the comparison are Sequelize attribute reads, which allocate per call, so wide fetches spent a lot of time and memory there. Swapped in a Map lookup.

## Tests

No behaviour change, space ids are unique in the fetched set, so `find` and `Map.get` should be aligned, also, both return `undefined` on a miss.

## Risk

Low. Single expression, no signature or query change.

## Deploy Plan

Normal deploy.

### DIFF
--- a/front/lib/resources/resource_with_space.ts
+++ b/front/lib/resources/resource_with_space.ts
@@ -119,10 +119,14 @@ export abstract class ResourceWithSpace<
       dangerouslyBypassWorkspaceIsolationSecurity: true,
     });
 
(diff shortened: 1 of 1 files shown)


id13 (approved):
LGTM

flvndvd (commented):
🙏

PR #30005, opened 2026-08-04
fix: batch file deletion on workspace hard delete

adrsimon:
## Description

`FileResource.deleteAllForWorkspace` deleted every file of a workspace in one unbounded `DELETE FROM files WHERE "workspaceId" = ?`. On a big enough workspace that never completes: it runs for the full 300s statement timeout, and Temporal retries it about every 7 minutes forever.

Deletes in batches of 10k now, so each statement is short and progress survives a retry.

## Tests

Added coverage for `deleteAllForWorkspace`, which had none: deletes all files of the workspace and returns the count, and leaves other workspaces alone. Full `file_resource.test.ts` suite passes (52).

## Risk

Low. Same end state, more statements. No longer atomic, but a partial delete just resumes on the next run, which is strictly better than the current behaviour.

## Deploy Plan

Nothing special. Once workers pick this up, the retry runs the batched version.

### DIFF
--- a/front/lib/resources/file_resource.ts
+++ b/front/lib/resources/file_resource.ts
@@ -464,21 +464,22 @@ export class FileResource extends BaseResource<FileModel> {
       where: { workspaceId },
     });
 
-    return this.batchDestroyAllForWorkspace(workspaceId);
+    return this.batchDestroyAllForWorkspace(auth);
   }
 
   // A workspace can hold millions of files. Deleting them in a single statement exceeds the
   // Postgres statement timeout, and the aborted transaction rolls back after having written
   // gigabytes of WAL, which stalls every other query on the instance. Batching keeps each
   // statement short and lets the deletion make forward progress across retries.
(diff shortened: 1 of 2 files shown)


front/lib/resources/file_resource.ts
```diff
@@ -462,9 +464,43 @@ export class FileResource extends BaseResource<FileModel> {
...
-      where: { workspaceId },
-    });
+    return this.batchDestroyAllForWorkspace(workspaceId);
+  }
+
+  // A workspace can hold millions of files. Deleting them in a single statement exceeds the
+  // Postgres statement timeout, and the aborted transaction rolls back after having written
+  // gigabytes of WAL, which stalls every other query on the instance. Batching keeps each
+  // statement short and lets the deletion make forward progress across retries.
+  private static async batchDestroyAllForWorkspace(workspaceId: ModelId) {
```
aubin-tchoi: nit: should take an auth in input

aubin-tchoi (approved):
LGTM

PR #30006, opened 2026-08-04
[front] chore: remove MCP usage outdated retro-compatibility path

aubin-tchoi:
## Description

Follow-up to #29819. The frontend rollout that consumes skill usage from the MCP usage endpoint is complete, so this removes the temporary `withSkills` opt-in and the legacy agent-only response contract.

The endpoint now always returns agent and skill usage, and the frontend uses the canonical usage URL again.

## Tests

- Tested locally.

## Risk

- Low.

## Deploy Plan

- Deploy front.


fabiencelier (approved):
🧹 🧹

PR #30010, opened 2026-08-05
Revert "[front] feat: audit pod function publishes"

davidebbo:
Reverts dust-tt/dust#29985. After further thinking, this event is just not that useful as it is already covered by the regular tool execution event.


smb2268:
Lgtm!

PR #30018, opened 2026-08-05
Pricing/poke: allow to sort the poke usage by more columns + filter by seat/creditState

fabiencelier:
## Description

Add 3 new column to order by: 
 - consumedAwuCredits
 - seat type
 - credit state 

Allow to filter by seat and credit state (to identify for instance the blocked user or the user on the pool)

## Tests

tested locally

Screenshot 2026-08-05 at 11 07 20

## Risk

low, it's poke

## Deploy Plan

deploy front

### DIFF
--- a/front/lib/api/credits/members_usage.ts
+++ b/front/lib/api/credits/members_usage.ts
@@ -1433,20 +1433,23 @@ async function resolveMembersUsagePageUsers({
   }
   const { users: allUsers, total } = allUsersResult.value;
 
+  // Every remaining sort column derives its key from the active membership
+  // (seat type, credit state, or the seat-type split for consumed credits).
+  const { memberships } = await MembershipResource.getActiveMemberships({
+    workspace,
+    users: allUsers,
+  });
+  const membershipByUserModelId = new Map(
+    memberships.map((m) => [m.userId, m])
+  );
+
   const sortKeyByUserId = new Map<string, number | string>();
   switch (orderColumn) {
     case "consumedAwuCredits": {
       // Split consumed credits on seat type so free-seat users sort by their
       // free-seat usage and everyone else by their paid-seat usage.
-      const { memberships } = await MembershipResource.getActiveMemberships({
-        workspace,
-        users: allUsers,
-      });
-      const seatTypeByUserModelId = new Map(
-        memberships.map((m) => [m.userId, m.seatType])
-      );
       const freeSeatUserIds = allUsers.flatMap((u) =>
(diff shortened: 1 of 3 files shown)


front/lib/api/credits/members_usage.ts
```diff
@@ -1297,16 +1303,37 @@ async function resolveGroupFilterUserIds({
...
+  creditState,
+}: {
+  workspace: LightWorkspaceType;
+  creditState: UserCreditState;
+}): Promise<string[]> {
+  const { memberships } = await MembershipResource.getActiveMemberships({
+    workspace,
+  });
+  return memberships
+    .filter((m) => m.creditState === creditState)
```
davidebbo, line 1320: nit: would be better to do server filtering?

front/lib/api/credits/members_usage.ts
```diff
@@ -1421,6 +1458,32 @@ async function resolveMembersUsagePageUsers({
       }
       break;
     }
+    case "seatType": {
+      const { memberships } = await MembershipResource.getActiveMemberships({
```
davidebbo: Maybe pull this out to before the switch since all cases need it.

front/lib/api/credits/members_usage.ts
```diff
@@ -1429,8 +1492,12 @@ async function resolveMembersUsagePageUsers({
   const sortedUsers = [...allUsers].sort((a, b) => {
     const keyA = sortKeyByUserId.get(a.sId) ?? 0;
     const keyB = sortKeyByUserId.get(b.sId) ?? 0;
-    if (keyA !== keyB) {
-      return (keyA - keyB) * directionFactor;
+    const cmp =
+      typeof keyA === "number" && typeof keyB === "number"
+        ? keyA - keyB
+        : String(keyA).localeCompare(String(keyB));
```
davidebbo, line 1493: This assumes both keys are of the same type? Which they might be, I don't know...

davidebbo (approved):
LGTM with minor comments

front/lib/api/credits/members_usage.ts
fabiencelier: yes they are, we always compare value of the same column

front/lib/api/credits/members_usage.ts
fabiencelier: it's server side filtering already, not push directly to SQL because we resolve the user first separatly
This is the pattern of existing filters

davidebbo: Yeah, I know, I meant 'server' as SQL side here. But all good.

PR #30021, opened 2026-08-05
fix: rename status to itemStatus in OpenAI tool search logs

adrsimon:
## Description

The OpenAI tool search logs spread `status: item.status` into the log payload. `status` is a Datadog reserved attribute, and its [status remapper](https://docs.datadoghq.com/logs/log_configuration/processors/log_status_remapper/) matches on the string prefix, anything starting with `c` becomes `critical`. So `"completed"` turned every one of these info logs into a critical in prod (~46k over 6h). Renamed to `itemStatus`.

## Tests

None, it's a log field rename.

## Risk

None. Anyone filtering on `@status` for these logs should use `@itemStatus`.

## Deploy Plan

Normal deploy.


flvndvd (approved):
🙏 can we nudge TS here to protect those keywords?

PR #30023, opened 2026-08-05
[front] fix: use grey variant for skill availability message

pinotalexandre:
## Summary
- The "Who is this skill available for?" `ContentMessage` in the skill builder used the default `info` variant, whose `bg-info-100`/`text-info-900` tokens resolve to golden/orange.
- Switch to the `primary` variant, which maps to `muted-background`/`muted-foreground` (grey).

from: 
image

to: 
image

## Test plan
- [ ] Open the skill builder and confirm the availability message now renders in grey instead of orange, in both light and dark mode.


aubin-tchoi (approved):
LGTM

PR #30027, opened 2026-08-05
[front] fix: create conversation in `pod_manager` with same depth

aubin-tchoi:
## Description

Closes https://github.com/dust-tt/tasks/issues/9929.

This PR fixes how we create conversations in the `pod_manager` MCP tools.
They are currently created with depth equal to the main conversation's depth + 1, making them all invisible in the pod conversation list.

Whether to always put 0 is debatable, here we choose to prevent sub agent from creating listable conversations. Can definitely be revisited in the future, this PR is more of a quick fix.

## Risk

- Low.

## Deploy Plan

- Deploy front.


xebeohp (approved):
LGTM, tysm!

PR #30028, opened 2026-08-05
fix: reject log payload statuses Datadog would misread as severity

adrsimon:
## Description

Follow-up to #30021. Datadog reads a top-level `status` as the log severity and prefix-matches its value: a value starting with `c` becomes `critical`, one starting with `s` becomes `ok`. A payload `status` from any other vocabulary rewrites the severity of the log. Prod over the last 3h:

| DD status | message | payload `status` |
|---|---|---|
| critical | OpenAI tool search query/results | `completed` |
| critical | Skipping wake-up: wake-up is not scheduled. | `cancelled` |
| ok | Compaction generation succeeded | `succeeded` |
| ok | Cancelling wake-up: conversation not found. | `scheduled` |

The log methods now accept only a real severity as `status`. Anything else is a compile error and the author picks another key.

The type check surfaced the offenders already in the code, which I renamed: HTTP statuses to `statusCode`, domain statuses to a descriptive name (`wakeUpStatus`, `subscriptionStatus`). Most of the file count is `Logger` type imports moving from pino to our own logger, which is what carries the checked signature. Applied in front, connectors, marketing and viz, the four places we build a pino logger.

One gap remains. A payload typed `Record<string, unknown>` still passes unchecked. The OpenAI logs escaped that way, with `status` set three functions upstream inside a `details: Record<string, unknown>`. Fixing that residue belongs in the Datadog pipeline: point the status remapper at `level` only, which also covers reserved attributes like `message` and `host`.

## Tests

`front/logger/logger.test.ts` pins the signature with `expectTypeOf`: a valid severity is accepted, so are `Error` and status-less payloads, and `"completed"` is rejected under `@ts-expect-error`. Type-checked front, front-api, front-spa, connectors, extension, marketing, viz and sparkle, all clean apart from errors that pre-exist on main (1 in front-api, 2 in viz). No runtime change, only types and key renames.

## Risk

Low. The renames change log field names, so anything querying `@status` on the messages above needs `@statusCode` or the new key. No monitors reference them today.

## Deploy Plan

Normal deploy.

### DIFF
--- a/connectors/migrations/20250128_github_backfill_folders_bis.ts
+++ b/connectors/migrations/20250128_github_backfill_folders_bis.ts
@@ -6,7 +6,7 @@ import {
 } from "@connectors/connectors/github/temporal/activities";
 import { ConnectorResource } from "@connectors/resources/connector_resource";
 import type { ModelId } from "@connectors/types";
-import type { Logger } from "pino";
+import type { Logger } from "@connectors/logger/logger";
 import { makeScript } from "scripts/helpers";
 
 makeScript(
@@ -49,7 +49,7 @@ async function backfillMissingFolders(
     repos = await getRepositories(connectorId);
   } catch (error) {
     logger.error(
-      error,
+      { error },
       "Error getting repositories for connector, skipping this connector"
     );
     return;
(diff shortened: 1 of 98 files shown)


flvndvd (approved):
YOLO, I'll trust you!

PR #30031, opened 2026-08-05
[front] feat: consumption analytics ES schema (parent agents/servers)

sfriquet:
## Description

Modify schema of `agent_message_consumption_analytics_1` in place:
- add `agent.parent_ids` to capture an agent's entire lineage. Useful to answer queries like "what's the cost of this agent, taking into account all its descendant"
- add `agent.direct_parent_id` to link to the direct parent of that sub agent. 
- rename `agent.root_agent_id` into `agent.root_id` to be consistent
- add `tool.parent_server_name`: because the computer tool can itself be responsible for triggering other calls (singular as we don't expect more than one parent)
- rename `enabled_by_skills_ids` to `attributed_skills_ids` to make it clear this lists all the skills we can attribute to this action (either a tool call or a tool call that enables this skill)

## Tests

Recreated the index locally.

## Risk

None. This index is not yet written to.

## Deploy Plan

Will recreate the index mapping from scratch.

### DIFF
--- a/front/lib/analytics/indices/agent_message_consumption_analytics_1.mappings.json
+++ b/front/lib/analytics/indices/agent_message_consumption_analytics_1.mappings.json
@@ -42,7 +42,13 @@
         "tag_ids": {
           "type": "keyword"
         },
-        "root_agent_id": {
+        "parent_ids": {
+          "type": "keyword"
+        },
+        "direct_parent_id": {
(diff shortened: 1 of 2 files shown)


sfriquet:

- #30031 ◀

flvndvd (approved):
🔥

PR #30032, opened 2026-08-05
[front] enh: Max connection pool to 10

id13:
## Description

We gave 40 connections to our api and 25 to our workers, but that was a patch over a lot of problems we had. 10 should be the right number looking at current metrics, we might have to drop activity slots on agent-loop again from 40 to 20, but looking at how often we get cpu throttled, seems like the right direction anyway.

## Tests

N/A

## Risk
Low
Blast radius: Connection acquisition increasing, or even spiking.
Will revert the deploy if I see this.

## Deploy Plan

Deploy front

### DIFF
--- a/front/lib/resources/storage/index.ts
+++ b/front/lib/resources/storage/index.ts
@@ -1,4 +1,3 @@
-import config from "@app/lib/api/config";
 import { SequelizeWithComments } from "@app/lib/api/database";
 import { dbConfig } from "@app/lib/resources/storage/config";
 import { getStatsDClient } from "@app/lib/utils/statsd";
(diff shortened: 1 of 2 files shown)


front/lib/resources/storage/index.ts
```diff
@@ -79,27 +78,16 @@ function reportPoolMetrics(
...
-function getPoolMaxForService(): number {
-  const service = config.getServiceName();
-
-  // DO NOT BLINDLY INCREASE THIS NUMBER (see comment below).
-  return service && WEB_SERVING_SERVICES.has(service) ? 40 : 25;
-}
+// DO NOT BLINDLY INCREASE THIS NUMBER. Each connection holds a PostgreSQL
+// backend via PgBouncer, so raising it shifts contention downstream. Prefer
+// reducing per-request connection usage (caching, shared connections).
+const POOL_MAX = 10;
```
flvndvd, line 84: PR's description mentioned that we could be CPU-bounded pretty often on the agent-loop deployment, is that fair read? Cause I feel like 20 activities per pod is very little, for the resources we ask.

What about lowering for either web components or agent loop first? So we can control blast radius?

id13: forgot to mention there is an infra PR coming:https://github.com/dust-tt/dust-infra/pull/958/changes
which addresses your fears

id13:
PMRR

PR #30034, opened 2026-08-05
Preserve previous cost attribution breakdowns

flvndvd:
## Description

When we bump the cost attribution version, existing messages still have rows from the previous version. Readers were only looking for the current version, so historical messages kept their billed credits but lost their detailed breakdown until a backfill was performed (except that we are not into backfills 😅).

This change fetches attribution rows up to the current version and selects the newest complete version for each message. For example, if version 4 is missing or incomplete but version 3 is complete, the version 3 breakdown remains available.

The same rule applies independently to every message in a conversation, so a conversation can safely aggregate both version 3 and version 4 messages. The conversation response does not expose a single version because a mixed aggregate cannot be described by one value. Message responses still expose the version they use.

No migration or backfill is required.

### DIFF
--- a/front/lib/resources/agent_message_consumption_item_resource.ts
+++ b/front/lib/resources/agent_message_consumption_item_resource.ts
@@ -446,11 +446,11 @@ export class AgentMessageConsumptionItemResource extends BaseResource<AgentMessa
     auth: Authenticator,
     {
       agentMessageModelIds,
-      attributionVersion,
+      maxAttributionVersion,
       transaction,
     }: {
       agentMessageModelIds: ModelId[];
-      attributionVersion: number;
+      maxAttributionVersion: number;
       transaction?: Transaction;
     }
   ): Promise<AgentMessageConsumptionItemResource[]> {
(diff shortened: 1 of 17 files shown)


front/lib/resources/agent_message_consumption_item_resource.ts
```diff
@@ -462,7 +462,7 @@ export class AgentMessageConsumptionItemResource extends BaseResource<AgentMessa
       where: {
         workspaceId: auth.getNonNullableWorkspace().id,
         agentMessageId: { [Op.in]: agentMessageModelIds },
-        attributionVersion,
+        attributionVersion: { [Op.lte]: maxAttributionVersion },
```
sfriquet, line 465: are we ever filtering on less than current version ? if not, do we even need to filter on `attributionVersion` ?

sfriquet (approved):

avervaet (approved):

front/lib/resources/agent_message_consumption_item_resource.ts
flvndvd: It was more of a safeguard, by providing the current version. The filter prevents reading newer attribution versions that may exist during a rolling deploy or rollback, while still allowing fallback to older versions.

PR #30035, opened 2026-08-05
Poke: fix the has opus access flag to latest rule

fabiencelier:
## Description

Current access to advanced models is determined by a flag on the plan ( +some feature flags).
Correctly use the new plan flag instead of old rule in the poke plan description

## Risk

low, only poke

## Deploy Plan

deploy front


Fraggle (approved):
LGTM

PR #30037, opened 2026-08-05
Return Pod function results in the invocation response when they settle in time

fontanierh:
## Description

The invocation endpoint returns as soon as the workflow is launched, so a Frame calling a Pod function pays a second round trip to the event stream for a result that was usually ready in under a second.

Behind `sandbox_function_fast_execution`, the endpoint waits a short while for the invocation to settle and returns its outcome inline, as a new optional `outcome` field on the response.

The wait stops as soon as the invocation blocks on a tool approval or on personal authentication. Holding the response there would deadlock: the approval card only renders once the client holds the invocation and subscribes to its stream, which is why execution moved into a workflow in the first place. Clients that get no outcome subscribe as before, and the stream replays from history everything already published.

## Tests

Added endpoint tests for the four cases: settles with a result, settles with an error, blocks on a tool approval (no outcome, so the card can render), and flag off.

Not exercised against a live pod, the flag is off everywhere.

## Risk

Low. Gated on `sandbox_function_fast_execution` (dust_only), and the response field is optional and additive, so an un-updated client keeps using the stream. Rollback is turning the flag off.

## Deploy Plan

None, ship and flip the flag on the Dust workspace when the rest of the stack is in.

### DIFF
--- a/front-api/routes/w/[wId]/sandbox-functions/[functionId]/invocations.test.ts
+++ b/front-api/routes/w/[wId]/sandbox-functions/[functionId]/invocations.test.ts
@@ -16,7 +16,10 @@ import { MembershipFactory } from "@app/tests/utils/MembershipFactory";
 import { SandboxFunctionMCPActionFactory } from "@app/tests/utils/SandboxFunctionMCPActionFactory";
 import { SpaceFactory } from "@app/tests/utils/SpaceFactory";
 import { UserFactory } from "@app/tests/utils/UserFactory";
-import type { SandboxFunctionUserIdentityPolicy } from "@app/types/api/sandbox_functions";
+import type {
+  SandboxFunctionInvocationEvent,
+  SandboxFunctionUserIdentityPolicy,
+} from "@app/types/api/sandbox_functions";
 import { sandboxFunctionContentType } from "@app/types/files";
 import { Err, Ok } from "@app/types/shared/result";
 import { honoApp } from "@front-api/app";
@@ -31,6 +34,9 @@ vi.mock("@app/lib/api/sandbox_functions/events", async (importOriginal) => {
   return {
     ...mod,
     publishSandboxFunctionInvocationEvent: vi.fn(),
+    // Defaults to a stream that ends without settling, which is what an invocation still running
+    // when the request returns looks like.
+    getSandboxFunctionInvocationEvents: vi.fn(async function* () {}),
   };
 });
 
@@ -58,7 +64,10 @@ vi.mock("@app/lib/actions/tool_status", async (importOriginal) => {
 });
 
 import { setUserAlwaysApprovedTool } from "@app/lib/actions/tool_status";
-import { publishSandboxFunctionInvocationEvent } from "@app/lib/api/sandbox_functions/events";
+import {
+  getSandboxFunctionInvocationEvents,
+  publishSandboxFunctionInvocationEvent,
+} from "@app/lib/api/sandbox_functions/events";
 import {
   launchSandboxFunctionInvocationWorkflow,
   launchSandboxFunctionToolWorkflow,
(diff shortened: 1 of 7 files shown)


front-api/routes/w/[wId]/sandbox-functions/[functionId]/invocations.ts
```diff
@@ -147,9 +154,17 @@ app.post(
...
+      ? await awaitSandboxFunctionInvocationOutcome({
+          invocationId: invocation.sId,
+        })
+      : null;
+
     return ctx.json(
       {
-        invocation: invocationResult.value.toJSON(),
+        invocation: invocation.toJSON(),
+        ...(outcome ? { outcome } : {}),
```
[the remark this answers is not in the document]
Comment (fontanierh) [reply front-api/routes/w/[wId]/sandbox-functions/[functionId]/invocations.ts:167]: This POST handler is `@ignoreswagger` (line 99), and `PostSandboxFunctionInvocationResponseBody` carries no `@swaggerschema` annotation, so there is no Swagger doc for it to fall out of sync with. The only `@swagger` block in this file documents the SSE events redirect, which is unchanged. Leaving as is rather than documenting an endpoint that is deliberately undocumented.

front-api/routes/w/[wId]/sandbox-functions/[functionId]/invocations.ts
```diff
@@ -120,6 +122,11 @@ app.post(
       });
     }
 
+    const fastExecution = await hasFeatureFlag(
+      auth,
+      "sandbox_function_fast_execution"
```
Comment (davidebbo) [inline front-api/routes/w/[wId]/sandbox-functions/[functionId]/invocations.ts:127]: Hopefully a short lived flag!

davidebbo (approved):
LGTM, this should cover most invocations!

front/lib/api/sandbox_functions/await_invocation.ts
```diff
@@ -0,0 +1,63 @@
...
+    }
+  } catch (error) {
+    // The invocation itself is unaffected — only this wait failed. Fall back to the stream.
+    logger.error(
+      { invocationId, error: normalizeError(error).message },
+      "Failed to wait for Pod function invocation outcome"
+    );
+  }
+
+  return null;
```
davidebbo, line 71: Maybe an info log here. This is the case where it's actually taking more than 10 seconds without being blocked. Should be pretty rare, and would be interesting to track.

front-api/routes/w/[wId]/sandbox-functions/[functionId]/invocations.test.ts
```diff
@@ -322,6 +371,101 @@ describe("POST /api/w/:wId/sandbox-functions/:functionIdOrSlug/invocations", ()
...
+      workspaceId: workspace.sId,
+      functionIdOrSlug: sandboxFunction.sId,
+    });
+
+    expect(response.status).toBe(201);
+    const body = await response.json();
+    expect(body.outcome).toBeUndefined();
+  });
+
+  it("does not wait for an outcome when fast execution is disabled", async () => {
```
Comment (davidebbo) [inline front-api/routes/w/[wId]/sandbox-functions/[functionId]/invocations.test.ts:455]: Missing the case where it legit takes more than 10s, though obviously we wouldn't want the test to take this long.

front-api/routes/w/[wId]/sandbox-functions/[functionId]/invocations.ts
Comment (fontanierh) [reply front-api/routes/w/[wId]/sandbox-functions/[functionId]/invocations.ts:127]: Yeah ! it's mostly to switch on and off to compare actual latency gains

front/lib/api/sandbox_functions/await_invocation.ts
fontanierh: Added. The catch path now returns early, so the log at the bottom only fires when the stream ended without an outcome and without the invocation blocking, which is the genuinely slow case you describe:

```
logger.info(
  { invocationId, timeoutMs },
  "Pod function invocation outlived the synchronous wait"
);
```

Should be a good signal for whether 10s is set anywhere near right, which nothing else measures today.

PR #30038, opened 2026-08-05
Record an execution mode on Pod functions

fontanierh:
## Description

Stacked on #30037.

A Pod function that calls tools may wait on the user for approval or authentication, for as long as they take and across sandbox stops. That is what the invocation workflow is for. A function that only reads and writes pod state can never block that way, and pays for durability it does not use.

This records which of the two a function is, at publish time. Nothing reads the mode yet.

Publish time rather than call time, because whether a function calls tools is decided by its own body, on branches the caller cannot see: a caller asking for a fast invocation would work until the day it hit the blocking branch. Aborting mid-execution to escalate is not safe either, the function may already have written to its database.

Every publish restates the mode, like the description and the schemas. A re-publish that omits it gets the default, so a publish that adds a tool call cannot leave the function fast.

`executionMode` goes on the publish tool rather than the function's `schema` export (where `userIdentity` lives) because that export is parsed by the dsbx runner, so it would mean a CLI release and an image rebuild.

## Tests

Added coverage for the default, an explicit `fast` publish, a re-publish that omits the mode going back to `durable`, and one that restates it keeping `fast`.

## Risk

Low. Existing functions and publishes that do not declare a mode read as `durable`, which is current behaviour, and nothing consumes the column yet.

## Deploy Plan

Pre-deploy migration adds a nullable `executionMode` column to `sandbox_functions`, same shape as the `userIdentity` column added in `20260730140943`. No backfill.

### DIFF
--- a/front/lib/api/actions/servers/sandbox_functions/metadata.ts
+++ b/front/lib/api/actions/servers/sandbox_functions/metadata.ts
@@ -1,6 +1,7 @@
 import type { ServerMetadata } from "@app/lib/actions/mcp_internal_actions/tool_definition";
 import {
   POD_DATABASE_NAME_REGEX,
+  SANDBOX_FUNCTION_EXECUTION_MODES,
   SANDBOX_FUNCTION_SLUG_REGEX,
 } from "@app/types/api/sandbox_functions";
 import { z } from "zod";
@@ -70,6 +71,19 @@ export const SANDBOX_FUNCTIONS_TOOLS_METADATA = [
         .describe(
           "Scoped path to the function's TypeScript source in the pod, as shown by the files " +
             "tools (e.g. `pod-<id>/greet.ts`)."
+        ),
+      executionMode: z
+        .enum(SANDBOX_FUNCTION_EXECUTION_MODES)
+        .optional()
+        .describe(
+          "`fast` runs synchronously and returns in a fraction of the time, but cannot call Dust " +
+            "tools through `dsbx tools`. It can still read and write pod state, run local " +
+            "binaries and make outbound HTTP calls, though those count against its execution " +
+            "ceiling. `durable` (the default) is for a function that calls `dsbx tools`, which " +
+            "may wait on the user for approval or authentication. Keep the functions a Frame " +
+            "calls on user interaction or on a poll `fast`, and isolate `dsbx tools` calls in " +
+            "their own `durable` functions. Restate this on every publish: a re-publish that " +
+            "omits it goes back to `durable`."
         ),
     },
     stake: "low",
--- a/front/lib/resources/sandbox_function_resource.ts
+++ b/front/lib/resources/sandbox_function_resource.ts
@@ -26,10 +26,14 @@ import type { ResourceFindOptions } from "@app/lib/resources/types";
 import type { UserResource } from "@app/lib/resources/user_resource";
 import type {
   PostSandboxFunctionInvocationRequestBody,
+  SandboxFunctionExecutionMode,
   SandboxFunctionInvocationOrigin,
   SandboxFunctionUserIdentityPolicy,
 } from "@app/types/api/sandbox_functions";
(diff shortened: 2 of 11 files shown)


front/lib/api/actions/servers/sandbox_functions/metadata.ts
```diff
@@ -71,6 +72,17 @@ export const SANDBOX_FUNCTIONS_TOOLS_METADATA = [
           "Scoped path to the function's TypeScript source in the pod, as shown by the files " +
             "tools (e.g. `pod-<id>/greet.ts`)."
         ),
+      executionMode: z
+        .enum(SANDBOX_FUNCTION_EXECUTION_MODES)
+        .optional()
+        .describe(
+          "`fast` for a function that only reads and writes pod state: it runs synchronously and " +
+            "returns in a fraction of the time, but cannot call tools. `durable` (the default) for " +
+            "a function that calls tools, which may wait on the user for approval or " +
```
davidebbo: Maybe mention `dsbx tools` explicitly. "Calling tools" is vague and can include spawning a local binary, which can work in fast mode.

front/lib/resources/sandbox_function_resource.ts
```diff
@@ -215,9 +224,18 @@ export class SandboxFunctionResource extends BaseResource<SandboxFunctionModel>
           }
 
           await this.file.uploadContent(auth, bundleCode);
+          // The execution mode is restated by every publish, like the description and the schemas:
+          // a re-publish that does not name it gets the default. Carrying the old mode forward
+          // would keep a function fast after the publish that added a tool call to it, which only
+          // shows up as a refused tool call at run time.
```
davidebbo, line 230: So we'll need to block those fast mode tool calls in a very clear way. Or we could be super smart and flip the mode when that happens!

front/types/api/sandbox_functions.ts
```diff
@@ -23,6 +23,24 @@ export const SANDBOX_FUNCTION_USER_IDENTITY_POLICIES = [
 export type SandboxFunctionUserIdentityPolicy =
   (typeof SANDBOX_FUNCTION_USER_IDENTITY_POLICIES)[number];
 
+// How an invocation reaches the sandbox.
+//
+// `durable` runs through a Temporal workflow, which is what a function calling tools needs: a tool
+// call can wait on an approval or on personal authentication for as long as the user takes, across
+// sandbox stops and restarts.
+//
+// `fast` is for functions that only read and write pod state. They cannot block on user input, so
```
davidebbo: Technically they can also make outbound http calls.

front/types/api/sandbox_functions.ts
```diff
@@ -23,6 +23,24 @@ export const SANDBOX_FUNCTION_USER_IDENTITY_POLICIES = [
...
+// their invocations do not need to outlive the request that starts them.
+export const SANDBOX_FUNCTION_EXECUTION_MODES = ["fast", "durable"] as const;
+
+export type SandboxFunctionExecutionMode =
+  (typeof SANDBOX_FUNCTION_EXECUTION_MODES)[number];
+
+// Functions published before execution modes existed have no recorded mode, and functions are
```
davidebbo: Can't we do a `DEFAULT "durable"` when creating the new column and avoid this condition?

davidebbo (approved):
LGTM with a few comments

front/lib/api/actions/servers/sandbox_functions/metadata.ts
fontanierh: ✅

front/types/api/sandbox_functions.ts
fontanierh: done

front/types/api/sandbox_functions.ts
fontanierh: indeed

PR #30039, opened 2026-08-05
Run fast Pod function invocations without the invocation workflow

fontanierh:
## Description

Stacked on #30038.

A fast function cannot call Dust tools through `dsbx tools`, which is the one thing that can wait on a person. It can still read and write pod state, run local binaries and make outbound HTTP calls, but none of those block on approval or authentication, so its invocation has nothing to survive. Routing it through a Temporal workflow costs a workflow start and an activity dispatch for durability it never uses.

Behind `sandbox_function_fast_execution`, a fast invocation executes in the request that creates it. The workflow still owns durable functions, and fast invocations whose sandbox is not already running.

`execute()` gains an `inline` mode for that, which constrains it twice:

- The sandbox is used only if it is already running, never created, woken or recreated. This is enforced by `SandboxResource.ensureActive` under the lifecycle lock, not by a check in the caller. A caller-side check is not enough: `ensureActive` destroys and recreates any sandbox with `killRequestedAt` set regardless of status, and `dangerouslyRequestKillForBaseImage` marks running sandboxes on every image rollout, so a full cold start (create, GCS mount, database restore) could still land inside the request. Nothing has executed when the refusal fires, so the invocation goes to the workflow and returns no outcome, which #30037 already handles.
- Execution is bounded at 10s instead of the workflow's 2 minutes. A fast function is not guaranteed to be quick, only unable to wait on a person: an outbound HTTP call to a slow endpoint still takes as long as it takes, and past the ceiling the invocation is failed rather than retried durably, since it may already have written to pod state.

An inline invocation that fails records its own outcome, as the workflow's activity does, so listeners settle instead of waiting for a workflow that never ran.

For scale, from production measurements taken before the change: the legs this removes are ~215ms of invocation POST plus ~200-300ms of Temporal dispatch, against a ~870ms worker-side p50.

## Tests

Added `ensureActive` tests for all four `requireRunning` cases, including the one that motivated moving the check: a `running` sandbox with `killRequestedAt` is refused and neither destroyed nor recreated.

Added invocation tests for: fast runs inline and skips the workflow, durable uses the workflow, flag off uses the workflow, a sandbox that is not running escalates without executing, an inline failure is recorded, and the inline exec ceiling is applied.

Not exercised against a live pod, the flag is off everywhere.

## Risk

Medium, this is the behaviour change. Gated on `sandbox_function_fast_execution` (dust_only) and only reachable for functions published `fast`, of which there are none yet. Worst case for a fast function is a request held up to 10s before the invocation is failed. Rollback is turning the flag off, which sends everything back through the workflow.

Worth watching after the flag goes on: how often invocations escalate (logged with a reason) and how close inline execs run to the 10s ceiling. There is no metric for the latter yet.

## Deploy Plan

None beyond #30038's migration.

### DIFF
--- a/front/lib/resources/sandbox_function_invocation_resource.ts
+++ b/front/lib/resources/sandbox_function_invocation_resource.ts
@@ -12,6 +12,7 @@ import {
   generateExecId,
   generateSandboxFunctionInvocationToken,
 } from "@app/lib/api/sandbox/access_tokens";
+import { isSandboxNotRunningError } from "@app/lib/api/sandbox/errors";
 import { ensurePodSandboxReady } from "@app/lib/api/sandbox/lifecycle";
 import { shellEscape } from "@app/lib/api/sandbox/shell";
 import { SandboxFunctionInvocationError } from "@app/lib/api/sandbox_functions/errors";
@@ -65,6 +66,11 @@ import { fromError } from "zod-validation-error";
 
 const SANDBOX_FUNCTION_WORKING_DIRECTORY = "/home/agent";
 const SANDBOX_FUNCTION_EXEC_TIMEOUT_MS = 2 * 60 * 1000;
+// A fast function is contractually short, and an inline invocation holds a request for as long as
+// it runs. Bound it well below the workflow's ceiling: past this the invocation is failed rather
+// than handed to the workflow, since by then the function may already have written to pod state
+// and re-running it would repeat those writes.
+const SANDBOX_FUNCTION_INLINE_EXEC_TIMEOUT_MS = 10 * 1000;
 const DSBX_BIN_PATH = "/opt/bin/dsbx";
 // Caps on runner output surfaced on failure: a small head for the error forwarded to the agent,
 // a larger one for the log fields.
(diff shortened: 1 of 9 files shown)


front/lib/resources/sandbox_function_invocation_resource.ts
```diff
@@ -778,6 +819,30 @@ export class SandboxFunctionInvocationResource extends BaseResource<SandboxFunct
...
 
+    if (await shouldExecuteInline(auth, sandboxFunction)) {
+      const executionResult = await invocation.execute(auth, { inline: true });
+      if (executionResult.isOk()) {
+        return new Ok(invocation);
+      }
+      if (!isSandboxNotRunningError(executionResult.error)) {
+        // The invocation ran and failed. Record the outcome here, as the invocation workflow's
+        // activity does, so listeners settle instead of waiting for a workflow that never ran.
+        await invocation.fail(executionResult.error);
```
davidebbo, line 830: Reaching the 10s timeout would end up as a failure here?

davidebbo (approved):
LGTM

front/lib/resources/sandbox_function_invocation_resource.ts
fontanierh: Yes. Not perfect but I don't have a better idea yet

PR #30042, opened 2026-08-05
Increase max activities for the dust project worker. Increase incremental sync delay for pod. Add endpoint and workflow for on demand sync.

Fraggle:
## Description

**NEXT PR IN THE STACK : uses the new endpoint to request a sync when files are added to the Pod**

Three related changes to `dust_project` connector sync:

- Bumps `maxConcurrentActivityTaskExecutions` from 5 → 12 in `runDustProjectWorker`.
- Slows the background cron from every 10 minutes to hourly, offset per connector (`connector.id % 60`) to spread load.
- Adds an on-demand incremental sync path: `dustProjectIncrementalSyncNowWorkflow` uses Temporal `signalWithStart` + a 30 s debounce to coalesce bursts into a single sync. Exposed via `POST /connectors/sync/:connector_id/incremental` and `requestIncrementalSync()` on `DustProjectConnectorManager`. `stopDustProjectSyncWorkflow` now terminates all three workflow IDs. `dust_project` is also added to the CLI batch restart allowlist.

## Tests

Tested locally.

## Risk

- **Cron slowdown**: existing `dustProjectIncrementalSyncWorkflow` instances already running at the 10-minute schedule will keep that cadence until terminated; the new hourly schedule only applies on next `launchDustProjectIncrementalSyncWorkflow` call. The new on-demand path compensates for the wider catch-up window.
- **Concurrency bump** (5 → 12): may increase CPU/memory on the connectors worker; still conservative relative to typical Temporal limits.
- **Debounce workflow**: bounded by `MAX_DEBOUNCE_COUNT = 100` before `continueAsNew` to cap history size. Any signal received after the while-loop exits is lost (noted inline).
- Rollback is safe: revert the cron schedule line and the concurrency constant; terminate the new workflow type if needed.

## Deploy Plan

1. Deploy `connectors`.
2. Existing incremental cron workflows retain their old 10-minute schedule. To migrate to hourly immediately, run the CLI batch restart for `dust_project` (now in `PROVIDERS_ALLOWING_RESTART`).

### DIFF
--- a/connectors/src/connectors/dust_project/temporal/workflows.ts
+++ b/connectors/src/connectors/dust_project/temporal/workflows.ts
@@ -1,6 +1,14 @@
 import type * as activities from "@connectors/connectors/dust_project/temporal/activities";
+import { dustProjectSyncSignal } from "@connectors/connectors/dust_project/temporal/signals";
 import type { ModelId } from "@connectors/types";
-import { proxyActivities } from "@temporalio/workflow";
+import {
+  allHandlersFinished,
+  condition,
+  continueAsNew,
+  proxyActivities,
+  setHandler,
+  sleep,
+} from "@temporalio/workflow";
 
 const {
   dustProjectConversationsFullSyncActivity,
@@ -13,6 +21,11 @@ const {
   startToCloseTimeout: "60 minutes",
 });
 
+/** Debounce window before running an on-demand incremental sync. */
+const INCREMENTAL_SYNC_DEBOUNCE_MS = 30_000;
+/** Cap debounce loops before continueAsNew to bound workflow history size. */
+const MAX_DEBOUNCE_COUNT = 100;
+
 /**
  * Generate workflow IDs for dust_project workflows
  */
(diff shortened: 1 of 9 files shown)


Fraggle:

- #30042 ◀
    - #30045

connectors/src/connectors/dust_project/temporal/workflows.ts
```diff
@@ -63,21 +96,49 @@ export async function dustProjectFullSyncWorkflow({
...
-  ) {
-    await dustProjectMarkSyncedActivity({ connectorId });
+  await runDustProjectIncrementalSync(connectorId);
+}
+
+/**
+ * On-demand incremental sync, coalesced via Temporal signals.
+ * Front notifies on file/conversation changes; bursts debounce into one sync.
+ */
+export async function dustProjectIncrementalSyncNowWorkflow({
```
matteotrab, line 113: What is the difference between having this new workflow and unpausing the pod datasource from poke to trigger a sync?

matteotrab: Ah I was missing the next pr, we want to trigger it when uploading a file

Fraggle:
@matteotrab merging to tackle the insane queue size.
PMRR

PR #30045, opened 2026-08-05
Request immediate (debounced) sync when file is added / remove from pod.

Fraggle:
## Description

After a file is added to or removed from a Pod, trigger a debounced `dust_project` incremental sync so the connector picks up the change without waiting for the hourly cron. New `request_incremental_sync.ts` module exposes two fire-and-forget helpers — `requestDustProjectIncrementalSync(auth, space)` and `requestDustProjectIncrementalSyncForScopedPath(auth, scopedPath)` — that resolve the connector ID and call `POST /connectors/sync/:connectorId/incremental` on the connectors service. Call sites are `addFileToProject`, `removeFileFromProject`, `deleteProjectFile` in `context.ts`, and the `DELETE` handler in the canonical path route. `ConnectorsAPI.requestIncrementalSync()` is added as the underlying HTTP call.

## Tests

Tested locally by adding/removing files from a Pod and verifying the incremental sync workflow is triggered. Failures are logged and never surfaced to callers, so error paths were verified through log output.

## Risk

Low. Fire-and-forget with full error handling — failures log a warning and are never thrown. The `space.isProject()` guard ensures this only fires for pod spaces. Debounce logic lives in the connectors service (introduced in PR #30042). Safe to roll back: the connectors endpoint simply won't be called on file operations, falling back to the hourly cron.

## Deploy Plan

Deploy front + front-api. The connectors-side endpoint (`POST /connectors/sync/:connectorId/incremental`) is already live.

### DIFF
--- a/connectors/src/api/sync_connector_incremental.ts
+++ b/connectors/src/api/sync_connector_incremental.ts
@@ -1,0 +1,69 @@
+import { getConnectorManager } from "@connectors/connectors";
+import { withLogging } from "@connectors/logger/withlogging";
+import { ConnectorResource } from "@connectors/resources/connector_resource";
+import type { WithConnectorsAPIErrorReponse } from "@connectors/types";
+import type { Request, Response } from "express";
+
+type RequestIncrementalSyncRes = WithConnectorsAPIErrorReponse<{
+  workflowId: string;
+}>;
+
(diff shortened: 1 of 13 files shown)


Fraggle:

- #30042
    - #30045 ◀

Fraggle:
PMRR

PR #30047, opened 2026-08-05
[front] enh: read user memory for non-trivial requests

aubin-tchoi:
## Description

Memory can improve more than explicitly personal requests, but agents can currently skip it when a request merely appears generic. This follow-up to #30002 makes memory recall the default before any non-trivial request, including reasoning, planning, recommendations, drafting, analysis, and multi-step work.

Reading remains skippable for simple, fully self-contained requests that cannot depend on the user.

## Tests

- Tested locally.

## Risk

- Low. Safe to revert.

## Deploy Plan

- Deploy front.


Inesvvv (commented):
LGTM 👍

PR #30048, opened 2026-08-05
[front] fix: show exact free credit renewal date

aubin-tchoi:
## Description

Fixes dust-tt/tasks#9917.

Legacy programmatic-usage non-credit workspaces can receive recurring free credits on a different schedule from their subscription billing cycle.
In the example in the card, the UI showed the date for the billing cycle (based on the subscription start date) even though the active free-credit expired on a different day (confirmed in poke). IIUC the user has no way of actually knowing when the credits will be refilled, and they will randomly get back to full on one day that is different from the one in the UI.

This PR fixes that to show the appropriate date for non-credit based plans, keeping the same behavior for the ones that are on credits.

## Tests

- Tested locally with various plans.

## Risk

- Low.

## Deploy Plan

- Deploy front.


id13 (approved):
Looks legit

fabiencelier (commented):
✅

PR #30049, opened 2026-08-05
[front] fix: use correct logger type in backfill script

aubin-tchoi:
## Description

Fixes lint on main.

The pod manager conversation migration fails Front typechecking because it expects a raw `pino` logger, while `makeScript` provides the narrowed Dust application logger.

## Deploy Plan

- No deploy.


aubin-tchoi:
PMRR

id13:
LGTM

PR #30050, opened 2026-08-05
[front] Warn about broken Pod function callers before restricting a Pod

davidebbo:
https://github.com/user-attachments/assets/8df0e94d-3fd1-4901-b262-0479fc453acc

### Description

Restricting a Pod removes the workspace global group from its space. Anyone who is neither a Pod member nor a workspace admin then loses read access, and can no longer invoke the Pod's functions — silently breaking any Frame they were driving. Today there is no way to know how big that break would be before flipping the toggle.

This adds a pre-flight estimate to the visibility toggle in Pod settings: how many recent Pod function invocations came from users who would lose access, and how many such users there are. When the count is non-zero, the existing confirmation dialog gains a line like *"12 Pod function calls in the last 30 days came from 3 people who are not a member of this Pod. Those calls will stop working."*

No schema change: `sandbox_function_invocations` already records `userId` and `createdAt`, which is everything the estimate needs.

### How it works

`getPodRestrictionImpact` lists the Pod's functions, aggregates invocations per triggering user over the last 30 days, then subtracts the two sets that keep access after restricting:

- `pod.fetchDistinctActiveManualGroupMembers` — the Pod's own member and editor groups, i.e. exactly who survives dropping the global group.
- Workspace admins — function reads go through `space.canReadOrAdministrate`, so admins are never breakage. Without excluding them the number would be inflated.

Invocations with no human actor (API keys, Slack/email bots) are counted separately as `nonHumanInvocationCount`. They break on their key's access rather than a user's, so they are reported apart rather than folded in. That field is returned but not yet surfaced in the UI.

### Notes for reviewers

- `SandboxFunctionInvocationResource.countByUserSince` aggregates with a SQL `GROUP BY "userId"` rather than counting in JS, since a 30-day window can span many rows and none are needed individually. Aggregate rows are not model instances, so they are parsed with a small zod schema instead of cast, per GEN4.
- The SWR hook is disabled unless the viewer is a Pod editor looking at a Pod that is still open — the only case where the number is actionable.
- If the background fetch has not landed when the toggle is clicked, the handler awaits fresh data with the toggle disabled, so the dialog never silently omits the warning.
- The 30-day window constant lives in the shared types file and is imported by both the business layer and the UI copy, so the counts and the "last 30 days" wording cannot drift.
- No index was added. The aggregation filters on `sandboxFunctionId` (already indexed) plus `createdAt`; whether `(sandboxFunctionId, createdAt)` earns its keep depends on real row counts, which are easier to judge once this is live.

### Limitation

The estimate is retrospective: it measures who *did* invoke in the window, not who *would*. A caller on a quarterly cadence reports as zero. The dialog wording reflects this — it states what will stop working, without claiming the count is exhaustive.

### Risk

Low. One new read-only endpoint, one new business-layer function, and one added warning line in an existing confirmation dialog. No existing behaviour changes, and nothing is written.

### Tests

- New functional tests on `GET /api/w/:wId/spaces/:spaceId/project_restriction_impact`: non-member callers counted, Pod members and workspace admins excluded, non-human invocations reported separately, zeros for a Pod with no functions, and 400 for a non-project space.
- `front/lib/resources/sandbox_function_invocation_resource.test.ts` still passes (35/35).


fontanierh (approved):
LGTM

PR #30052, opened 2026-08-05
Skills: correctly save the skill editors when creating a skill

fabiencelier:
## Description

closes https://github.com/dust-tt/tasks/issues/9884

Editors are updated in a second call after the skill is actually updated.
However, this second call was not done for newly created skill, losing the editors

This PR always saves the editors + propagate an error on editors save to the end user as a toast

## Tests

tested locally

## Risk

low, purely UI and easy to revert

## Deploy Plan

deploy front


aubin-tchoi (approved):
LGTM

PR #30053, opened 2026-08-05
refactor(front): stop pulling sparkle into the front-api bundle

adrsimon:
## Description

Server-side code was reaching connector descriptions and the MCP icon type guard through `.tsx` modules, so `dist/server.js` shipped 17 live `require("@dust-tt/sparkle")` calls. Since bare specifiers are external in the front-api esbuild config, that means the whole design system gets `require`d at boot, along with radix, framer-motion, react-markdown, lottie, emoji-mart and the rest. Measured cost: ~74MB heap, ~138MB RSS per pod, on a container capped at 4Gi that is currently getting OOMKilled. 

This is very not likely to be the reason of the OOMs, but this is a very clear candidate of things that does not belong in the api build.

Two chains were responsible:

- `lib/api/marketing/integrations.ts` read `description` / `guideLink` from `CONNECTOR_UI_CONFIGURATIONS`, which imports 14 `components/data_source/*.tsx`
- `lib/api/mcp/servers.ts` imported `isCustomResourceIconType` from `resources_icons.tsx` rather than `resources_icon_names.ts`

The 13 public connector description/guideLink pairs now live in `lib/connector_metadata.ts`, next to `connector_providers.ts`. `CONNECTOR_UI_CONFIGURATIONS` spreads them so the strings still have a single home, and the hidden bot connectors keep their `"N/A"` placeholders in the UI config since they have no public metadata.

Not the fix for the OOMs themselves, just some headroom that was being wasted.

## Tests

- `npm run build` in front-api: `require("@dust-tt/sparkle")` count goes 17 -> 0. Same for everything riding behind it (`react-dom`, `framer-motion`, `react-markdown`, `@radix-ui`, `lottie-react`, `@emoji-mart`, `@tanstack/react-table`, `allotment`, `react-confetti`). Verified against a build with the diff stashed, to be sure the delta was attributable.

## Risk

Low. No behaviour change intended, and the strings are byte-identical. The marketing integration registry and the connector setup UI both read the same data now. Rollback is a revert.

## Deploy Plan

Nothing special, ships with front.

### DIFF
--- a/front/lib/api/marketing/integrations.ts
+++ b/front/lib/api/marketing/integrations.ts
@@ -5,8 +5,8 @@ import {
 } from "@app/lib/actions/mcp_internal_actions/constants";
 import { DEFAULT_REMOTE_MCP_SERVERS } from "@app/lib/actions/mcp_internal_actions/remote_servers";
 import type { ServerMetadata } from "@app/lib/actions/mcp_internal_actions/tool_definition";
+import { getConnectorMetadata } from "@app/lib/connector_metadata";
 import { CONNECTOR_CONFIGURATIONS } from "@app/lib/connector_providers";
-import { CONNECTOR_UI_CONFIGURATIONS } from "@app/lib/connector_providers_ui";
 import type { ConnectorProvider } from "@app/types/data_source";
 
 export type IntegrationType = "mcp_server" | "connector" | "both";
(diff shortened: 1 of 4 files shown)


id13 (approved):
looks legit, maybe add a check in linter

adrsimon:
yup planning a linter or CI or whatever latter depending on what i find works best

PR #30062, opened 2026-08-05
fix(front): always create conversation forks at depth 0

smb2268:
## Description

Follow up on #28931 by creating every new conversation fork at depth 0, regardless of the parent conversation's depth. This keeps forks from sub-agent conversations visible as root conversations while preserving their parent and fork lineage.

## Tests

Updated `front/lib/api/assistant/conversation/forks.test.ts` to fork a depth-1 conversation and assert that the child has depth 0. `git diff --check` passed. The focused Vitest test was not run locally because the shallow clone does not have dependencies installed in the Computer.

## Risk

Low. The change only affects the depth assigned to newly created forks. Parent relationships, requested spaces, and other fork metadata remain unchanged. No data migration is required.

## Deploy Plan

Deploy through the standard front-end rollout. PR CI should run the focused test and the repository's normal validation checks.

### DIFF
--- a/front/lib/api/assistant/conversation/forks.ts
+++ b/front/lib/api/assistant/conversation/forks.ts
@@ -609,7 +609,9 @@ export async function createConversationFork(
         sId: generateRandomModelSId(),
         title: null,
         visibility: parentConversation.visibility,
-        // Forks are user-facing conversations, not run_agent sub-conversations.
+        // Forking is an explicit user operation, so users expect to find the
+        // conversation they created alongside their other root conversations.
+        // Forks are not run_agent sub-conversations, even when their source is.
         depth: 0,
         triggerId: null,
         spaceId: parentConversation.space?.id ?? null,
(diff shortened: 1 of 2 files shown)


front/lib/api/assistant/conversation/forks.ts
```diff
@@ -609,10 +609,8 @@ export async function createConversationFork(
         sId: generateRandomModelSId(),
         title: null,
         visibility: parentConversation.visibility,
-        // Forks are user-facing conversations, not run_agent sub-conversations:
-        // they must keep the parent's depth so depth-based behaviors (space
-        // listing, notifications, action limits) treat them as such.
-        depth: parentConversation.depth,
+        // Forks are user-facing conversations, not run_agent sub-conversations.
```
aubin-tchoi: let's explain a bit more the rationale since this choice is quite opinionated: since it's a user operation they are effectively initiating the creation of the conversation, therefore expect to find it easily, etc...

aubin-tchoi (approved):
LGTM

PR #30059, opened 2026-08-05
[front] fix: set a fixed depth of 0 on pod_manager-created conversations

aubin-tchoi:
## Description

Context in this [thread](https://dust4ai.slack.com/archives/C050SM8NSPK/p1785950807583959?thread_ts=1785932638.562669&cid=C050SM8NSPK).

This PR sets a depth of 0 on conversations created by the `pod_manager`.`create_conversation` tool.
Low risk as this effectively only reverts https://github.com/dust-tt/dust/pull/29940

## Deploy Plan

- Deploy front.

PR #30068, opened 2026-08-05
[front] test: check depth for pod manager-created conversations

aubin-tchoi:
## Description

Follow up on #30059 to protect the fixed depth behavior for conversations created through `pod_manager.create_conversation` (the MCP tool).

The test starts from a nested agent conversation, invokes the real tool, and verifies the new conversation is persisted at depth 0 in the same Pod.

## Tests

- Yes

## Risk

- N/A. No runtime change (only changes the tests).

## Deploy Plan

- No deploy.


frankaloia (approved):
Thank you

PR #30072, opened 2026-08-06
Fix: icon picker mousewheel support in pods.

Fraggle:
## Description

Adds `modal` to `PopoverRoot` in `EditPodFrameTabDialog` so the icon picker captures scroll events correctly. Without this prop, mousewheel input inside the popover is swallowed by the underlying dialog, making it impossible to scroll through the icon list in Pods.

## Tests

Tested locally by opening the icon picker in a Pod frame tab dialog and scrolling through icons with the mousewheel.

## Risk

Low. Single-prop addition to a Radix `PopoverRoot`; no logic change, no data model change. Rollback is trivial.

## Deploy Plan

Deploy front.

PR #30073, opened 2026-08-06
Add: can customize the title & icon when add a Frame to pod tabs

Fraggle:
## Description

When adding a Frame as a Pod tab, the old flow showed a plain confirmation dialog with no customization. Now, clicking "Add as Pod tab" (from both `PodFrameTabButton` and `PodFileExplorer`'s file menu) opens `EditPodFrameTabDialog` in `"create"` mode, letting the user set a custom title and icon before the tab is created.

Key changes:
- `EditPodFrameTabDialog` gains a `mode?: "create" | "edit"` prop; in create mode, move/remove actions are disabled and the footer shows "Add tab" instead of "Save"/"Remove tab".
- `toggleFrameTab` is replaced by separate `removeFrameTab` (direct) and `addFrameTab` (via dialog with `skipConfirm: true`) call paths in `usePodFrameTabs`.
- `FrameTabOptions` gains `skipConfirm` to bypass the old confirm dialog when the new dialog handles confirmation itself.
- `PopoverContent` inside the dialog gets `mountPortalContainer={document.body}` to fix icon-picker popover stacking.

## Tests

Tested locally — added a Frame as a Pod tab via both the file explorer menu and the `PodFrameTabButton`, verified the dialog opens with pre-filled title/icon, allows customization, and correctly adds the tab on confirm.

## Risk

Low. Front-only, no DB changes. The `skipConfirm` flag is only passed from within the new dialog, so existing non-dialog paths are unaffected. Rollback: redeploy front.

## Deploy Plan

Deploy `front`.

### DIFF
--- a/front/components/pod/files/EditPodFrameTabDialog.tsx
+++ b/front/components/pod/files/EditPodFrameTabDialog.tsx
@@ -35,6 +35,7 @@ interface EditPodFrameTabDialogProps {
   isEditor: boolean;
   includeConnectedData?: boolean;
   tab: PodFrameTab;
+  mode?: "create" | "edit";
   isOpen: boolean;
   onClose: () => void;
 }
@@ -47,10 +48,13 @@ export function EditPodFrameTabDialog({
   isEditor,
   includeConnectedData = false,
   tab,
+  mode = "edit",
   isOpen,
   onClose,
 }: EditPodFrameTabDialogProps) {
+  const isCreate = mode === "create";
   const {
+    addFrameTab,
     updateFrameTab,
     removeFrameTab,
     moveFrameTab,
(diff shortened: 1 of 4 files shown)


Fraggle:

- #30072
    - #30073 ◀

front/components/pod/files/EditPodFrameTabDialog.tsx
```diff
@@ -159,6 +175,9 @@ export function EditPodFrameTabDialog({
               </PopoverTrigger>
               <PopoverContent
                 className="w-fit p-0"
+                mountPortalContainer={
```
matteotrab, line 178: Isn't it mounted already at the document body?

matteotrab (approved):

front/components/pod/files/EditPodFrameTabDialog.tsx
Fraggle: Nope, there was a bug with the icons popup beneath the modal "greyed" div.

PR #30074, opened 2026-08-06
[front] enh: make idle-in-transaction counting reliable

id13:
## Description

Improve idle-in-tx measurement by accounting for event loop lag and connection acquisition time

## Tests

N/A

## Risk

Low

## Deploy Plan

Deploy front

### DIFF
--- a/front/lib/api/database.ts
+++ b/front/lib/api/database.ts
@@ -18,21 +18,74 @@ import { Sequelize } from "sequelize";
 declare module "sequelize" {
   interface Transaction {
     readonly id: string;
+    readonly parent?: Transaction;
   }
 }
 
 const IDLE_IN_TX_THRESHOLD_MS = 250;
 const MAX_TRACKED_QUERIES = 100;
+const LAG_SAMPLE_INTERVAL_MS = 50;
+
+// Event-loop lag accumulated process-wide. A stop-the-world pause (major GC, a blocking
+// callback) suspends every transaction that is open at that moment, and Postgres reports all of
+// them as idle in transaction. Without measuring the pause, one process freeze reads as a
+// separate slow codepath on each of those transactions.
+let cumulativeLagMs = 0;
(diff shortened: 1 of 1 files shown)


id13:
PMRR

front/lib/api/database.ts
```diff
@@ -18,21 +18,69 @@ import { Sequelize } from "sequelize";
...
+    readonly parent?: Transaction;
   }
 }
 
 const IDLE_IN_TX_THRESHOLD_MS = 250;
 const MAX_TRACKED_QUERIES = 100;
+const LAG_SAMPLE_INTERVAL_MS = 50;
+
+// Event-loop lag accumulated process-wide. A stop-the-world pause (major GC, a blocking
+// callback) suspends every transaction that is open at that moment, and Postgres reports all of
```
flvndvd, line 30:
```suggestion
// callback) suspends every transaction that is opened at that moment, and Postgres reports all of
```

front/lib/api/database.ts
```diff
@@ -18,21 +18,69 @@ import { Sequelize } from "sequelize";
...
+// them as idle in transaction. Without measuring the pause, one process freeze reads as a
+// separate slow codepath on each of those transactions.
+let cumulativeLagMs = 0;
+let lastLagSampleAtMs = performance.now();
+
+setInterval(() => {
```
flvndvd, line 36: The timer counts every event-loop pause during the transaction, including pauses while a SQL query is running.

But query-running time is already counted as `busyMs`, right? Don't we think that this pause can be counted twice and wrongly reduce the real idle time.

flvndvd (approved):
Might be wrong, but looking at it, wondering if we are not discounting twice.

front/lib/api/database.ts
id13: I think you are right

id13: Fixed

PR #30075, opened 2026-08-06
Fix: double mount trigger double load_code in Frame viz

Fraggle:
## Description

`VisualizationWrapper`'s `loadCode` effect listed `memoizedDownloadFile` and `handleScreenshotDownload` as dependencies. After mount, `useResizeDetector` changes its `ref` identity, which caused those callbacks to be recreated and the effect to re-fire — fetching and re-transpiling the visualization code a second time.

Fix: store both callbacks in refs (`handleScreenshotDownloadRef`, `memoizedDownloadFileRef`) and forward to them from stable wrappers inside the effect, so they never appear in the dependency array. The effect now depends only on `api.data`, `isEditable`, and `identifier`. Added a `cancelled` cleanup flag to short-circuit any in-flight async work if the effect tears down mid-run.

## Tests

Tested locally: opened a Frame visualization and confirmed `load_code` fires exactly once on mount, with no second invocation triggered by `useResizeDetector` ref changes.

## Risk

Low. The change is purely inside `VisualizationWrapper` in the `viz` app. Callback forwarding via refs is a well-understood React pattern; the `cancelled` flag prevents any stale-async side effects. Fully rollback-safe.

## Deploy Plan

Deploy `viz` / front.

### DIFF
--- a/viz/app/components/VisualizationWrapper.tsx
+++ b/viz/app/components/VisualizationWrapper.tsx
@@ -38,7 +38,7 @@ import * as lucideAll from "lucide-react";
 import * as motionAll from "motion/react";
 import * as papaparseAll from "papaparse";
 import * as reactAll from "react";
-import { useCallback, useEffect, useMemo, useState } from "react";
+import { useCallback, useEffect, useMemo, useRef, useState } from "react";
 import { useResizeDetector } from "react-resize-detector";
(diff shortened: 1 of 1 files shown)


Fraggle:

- #30075 ◀

flvndvd (approved):

PR #30079, opened 2026-08-06
[front] feat: consumption analytics query scaffolding

sfriquet:
## Description

Foundation for the new consumption-index-backed analytics dashboard. No
endpoint and no UI yet — this is the shared query layer every following
branch builds on.

- `searchConsumptionAnalytics`: same contract as `searchAnalytics`, bound to
  `front.agent_message_consumption_analytics`.
- `period.ts`: resolves "this cycle" / "last N days".
  Workspaces not on credit-based pricing have no billing cycle, so they
  resolve to the current UTC calendar month.
- `scope.ts`: the filter dimensions and their index fields, plus the
  workspace-scoped query builder.
- `schema.ts`: shared zod contract for the period and filter query params.

## Tests

uts

## Risk

Low. Safe to rollback.

## Deploy Plan

Front only

### DIFF
--- a/front/lib/api/analytics/consumption/period.test.ts
+++ b/front/lib/api/analytics/consumption/period.test.ts
@@ -1,0 +1,50 @@
+import { resolveConsumptionPeriod } from "@app/lib/api/analytics/consumption/period";
+import { Authenticator } from "@app/lib/auth";
+import { WorkspaceFactory } from "@app/tests/utils/WorkspaceFactory";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+
+// Jul 13 2026 09:30 UTC.
+const NOW_MS = Date.UTC(2026, 6, 13, 9, 30);
+
+async function setup() {
+  const workspace = await WorkspaceFactory.basic();
+  const auth = await Authenticator.internalAdminForWorkspace(workspace.sId);
+  return { auth };
+}
+
+describe("resolveConsumptionPeriod", () => {
+  beforeEach(() => {
+    vi.useFakeTimers();
+    vi.setSystemTime(new Date(NOW_MS));
+  });
+
+  afterEach(() => {
+    vi.useRealTimers();
+  });
+
+  it("resolves a days window to [start-of-day (days-1 ago), now]", async () => {
+    const { auth } = await setup();
+
+    const period = await resolveConsumptionPeriod(auth, {
+      kind: "days",
+      days: 7,
+    });
+
+    expect(period).toEqual({
+      // 6 days before Jul 13, floored to midnight.
+      startDate: "2026-07-07T00:00:00.000Z",
+      endDate: "2026-07-13T09:30:00.000Z",
+    });
+  });
+
+  it("resolves the cycle to the current UTC calendar month for a workspace with no billing cycle", async () => {
+    const { auth } = await setup();
+
+    const period = await resolveConsumptionPeriod(auth, { kind: "cycle" });
+
+    expect(period).toEqual({
+      startDate: "2026-07-01T00:00:00.000Z",
+      endDate: "2026-08-01T00:00:00.000Z",
+    });
+  });
+});
--- a/front/lib/api/analytics/consumption/period.ts
+++ b/front/lib/api/analytics/consumption/period.ts
@@ -1,0 +1,93 @@
+import type { Authenticator } from "@app/lib/auth";
+import { getCachedMetronomeCurrentBillingPeriod } from "@app/lib/metronome/contracts";
+import logger from "@app/logger/logger";
+import { isCreditPricedPlan } from "@app/types/plan";
+import moment from "moment-timezone";
+
+/**
+ * Period resolution for the consumption analytics endpoints: turns a requested
+ * period into the [startDate, endDate) window the queries are scoped to.
+ *
+ * "this cycle" is the workspace's current billing cycle; "last N days" is a
+ * relative window.
+ */
+
+export type ConsumptionPeriodInput =
+  | { kind: "cycle" }
+  | { kind: "days"; days: number };
+
+export type ConsumptionPeriod = {
+  startDate: string;
+  endDate: string;
+};
+
+// Workspaces that are not on credit-based pricing have no billing cycle to
+// speak of, so "this cycle" is the current UTC calendar month for them.
+function calendarMonthBounds(now: Date): {
+  cycleStart: Date;
+  cycleEnd: Date;
+} {
+  const startOfMonth = moment.utc(now).startOf("month");
+  return {
+    cycleStart: startOfMonth.toDate(),
+    cycleEnd: startOfMonth.clone().add(1, "month").toDate(),
+  };
+}
+
+// Returns the bounds of the current billing cycle for the workspace, or the
+// current calendar month if the workspace is not on a credit-based plan.
+// If the billing cycle cannot be resolved, returns the current calendar month.
+async function resolveCycleBounds(
+  auth: Authenticator,
+  now: Date
+): Promise<{ cycleStart: Date; cycleEnd: Date }> {
+  const plan = auth.plan();
+  if (!plan || !isCreditPricedPlan(plan)) {
+    return calendarMonthBounds(now);
+  }
+
+  const workspaceId = auth.getNonNullableWorkspace().sId;
+  const periodResult =
+    await getCachedMetronomeCurrentBillingPeriod(workspaceId);
+  if (periodResult.isErr()) {
+    logger.warn(
+      { workspaceId, err: periodResult.error },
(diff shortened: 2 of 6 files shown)


sfriquet:

- #30031
    - #30079 ◀
        - #30080
            - #30082
                - #30083
                    - #30084
                        - #30085
                            - #30086
                                - #30087
                                    - #30088

front/lib/api/analytics/consumption/period.ts
```diff
@@ -0,0 +1,93 @@
...
+  };
+}
+
+// Returns the bounds of the current billing cycle for the workspace, or the
+// current calendar month if the workspace is not on a credit-based plan.
+// If the billing cycle cannot be resolved, returns the current calendar month.
+async function resolveCycleBounds(
+  auth: Authenticator,
+  now: Date
+): Promise<{ cycleStart: Date; cycleEnd: Date }> {
```
avervaet, line 43: nit: we could have a dedicated type for this

front/lib/api/analytics/consumption/period.ts
```diff
@@ -0,0 +1,93 @@
...
+ *
+ * "this cycle" is the workspace's current billing cycle; "last N days" is a
+ * relative window.
+ */
+
+export type ConsumptionPeriodInput =
+  | { kind: "cycle" }
+  | { kind: "days"; days: number };
+
+export type ConsumptionPeriod = {
```
avervaet, line 19: Open question why are we converting to string before fetching ES. i checked and other parts of the codebase do the same but it is not obvious for me as Date object would behave the same ?

front/lib/api/analytics/consumption/schema.ts
```diff
@@ -0,0 +1,51 @@
...
+  // filter is a map of dimension to selected ids and does not flatten well
+  // into repeated query params.
+  filter: z
+    .string()
+    .optional()
+    .transform((value) => {
+      if (!value) {
+        return undefined;
+      }
+      try {
```
avervaet, line 35: nit: this try/catch + .pipe(...optional()) shape is duplicated from programmatic_cost.ts and awu_usage_analytics.ts's FilterSchema. Could extract a shared jsonEncodedFilterSchema(schema) helper and reuse it in all three?

avervaet (approved):

front/lib/api/analytics/consumption/period.test.ts
```diff
@@ -0,0 +1,50 @@
...
+    });
+
+    expect(period).toEqual({
+      // 6 days before Jul 13, floored to midnight.
+      startDate: "2026-07-07T00:00:00.000Z",
+      endDate: "2026-07-13T09:30:00.000Z",
+    });
+  });
+
+  it("resolves the cycle to the current UTC calendar month for a workspace with no billing cycle", async () => {
```
flvndvd, line 40: Can we also add one test for workspace on a credit-based plan?

front/lib/api/analytics/consumption/period.ts
```diff
@@ -0,0 +1,93 @@
...
+  };
+}
+
+export async function resolveConsumptionPeriod(
+  auth: Authenticator,
+  input: ConsumptionPeriodInput
+): Promise<ConsumptionPeriod> {
+  const now = new Date();
+
+  if (input.kind === "days") {
```
flvndvd, line 76: `nit`

We prefer use case for discriminated type with an `assertNever` for the default, so in case someone adds a new value, linter is pretty helpful.

front/lib/api/analytics/consumption/period.ts
```diff
@@ -0,0 +1,93 @@
...
+    const startMs = moment
+      .utc(now)
+      .subtract(input.days - 1, "days")
```
flvndvd, line 79: Because 0-index based?

front/lib/api/analytics/consumption/schema.ts
```diff
@@ -0,0 +1,51 @@
...
+  days: z.coerce
+    .number()
+    .int()
+    .positive()
+    .optional()
+    .default(DEFAULT_CONSUMPTION_PERIOD_DAYS),
+  // JSON-encoded, mirroring the existing analytics filter query param: the
+  // filter is a map of dimension to selected ids and does not flatten well
+  // into repeated query params.
+  filter: z
```
flvndvd, line 28: Are we sure about using query params, over passing filters in the body. From experience query params have a limited length, and json encoding adds some oversize here.

front/lib/api/analytics/consumption/scope.ts
```diff
@@ -0,0 +1,81 @@
...
+  "model",
+  "tool",
+  "skill",
+  "source",
+] as const;
+
+export type ConsumptionScopeDimension =
+  (typeof CONSUMPTION_SCOPE_DIMENSIONS)[number];
+
+export const CONSUMPTION_DIMENSION_FIELDS: Record<
```
flvndvd, line 19: Let's avoid exporting if not used.

front/lib/api/analytics/consumption/scope.ts
```diff
@@ -0,0 +1,81 @@
...
+    nonEmpty.length === 1
+      ? { term: { [field]: nonEmpty[0] } }
+      : { terms: { [field]: nonEmpty } },
+  ];
+}
+
+/**
+ * Workspace-scoped query over a half-open [startDate, endDate) window.
+ */
+export function buildConsumptionScopeQuery({
```
flvndvd, line 56: Can we take `auth` here? We usually avoids the `workspaceId`.

front/lib/api/analytics/consumption/scope.ts
```diff
@@ -0,0 +1,81 @@
...
+}: {
+  workspaceId: string;
+  startDate: string;
+  endDate: string;
+  filter?: ConsumptionScopeFilter;
+  extraFilters?: estypes.QueryDslQueryContainer[];
+}): estypes.QueryDslQueryContainer {
+  const filters: estypes.QueryDslQueryContainer[] = [
+    { term: { workspace_id: workspaceId } },
+    { range: { [COMPLETED_AT_FIELD]: { gte: startDate, lt: endDate } } },
```
flvndvd, line 71: As we store this field as `long` in our mapping can we once we've parsed the dates earlier in your pipeline, exclusively use numbers?

front/lib/api/analytics/consumption/scope.ts
```diff
@@ -0,0 +1,81 @@
...
+  filter = {},
+  extraFilters = [],
+}: {
+  workspaceId: string;
+  startDate: string;
+  endDate: string;
+  filter?: ConsumptionScopeFilter;
+  extraFilters?: estypes.QueryDslQueryContainer[];
+}): estypes.QueryDslQueryContainer {
+  const filters: estypes.QueryDslQueryContainer[] = [
```
flvndvd, line 69: We had a type (that we apparently removed) that was ensuring `workspace_id` was a top level clause, fine to not add it back in this PR, but let's make sure that we eventually do it at some point.

flvndvd (approved):
Left a few comments, but good to go once addressed.

front/lib/api/analytics/consumption/period.ts
sfriquet: if you ask for "last 7 days" we start from today and substract 6 days, and take the start of that day, so we get the 6 full previous days + today up to now = 7 day.

front/lib/api/analytics/consumption/scope.ts
sfriquet: it'll be imported later on

front/lib/api/analytics/consumption/scope.ts
sfriquet: `completed_at` is mapped as a `date`. I'd rather have dates than numbers it's much easier to read when debugging / reading logs.
I kept `string` which is consistent with all the other analytics file, though `date` sounds even better to me :thinking:

front/lib/api/analytics/consumption/scope.ts
sfriquet: yes; note that we discussed add an assert checking the query contains a filter but it could still miss if the filter is present but optional.
we could add the filter in `esSearch` or `searchAnalytics()` which may add an extra filter but with AND semantics it's okay.

front/lib/api/analytics/consumption/schema.ts
sfriquet: OK will follow up

sfriquet:
merging and will follow up re: body parameter and enforcing workspace_id in `esSearch`

front/lib/api/analytics/consumption/scope.ts
flvndvd: Oh indeed, I thought it was store as `number` then fine.

PR #30081, opened 2026-08-06
Prevent Agent to create message in the same conversation.

Fraggle:
## Description

Removes the default-to-current-conversation fallback from the `create_conversation_message` Pod Manager tool, and adds an explicit guard that returns an `MCPError` when `conversationId` matches the running conversation's `sId`. `conversationId` is now required in the Zod schema (`.optional()` dropped). This prevents agents from posting messages back into their own conversation, which would trigger recursive agent loops.

## Tests

Tested manually.

## Risk

Low. The schema change makes `conversationId` required — any agent call that previously omitted it to self-post will now receive a clear error instead of silently looping. No DB changes, no API surface changes outside the MCP tool schema.

## Deploy Plan

Deploy front.

### DIFF
--- a/front/lib/api/actions/servers/pod_manager/metadata.ts
+++ b/front/lib/api/actions/servers/pod_manager/metadata.ts
@@ -557,9 +557,8 @@ export const POD_MANAGER_TOOLS_METADATA = [
     schema: {
       conversationId: z
         .string()
-        .optional()
         .describe(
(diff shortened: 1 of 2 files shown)


Fraggle:

- #30081 ◀

fontanierh (approved):

fontanierh (approved):

PR #30082, opened 2026-08-06
[front] feat: consumption analytics page skeleton

sfriquet:
## Description

The page shell for the new Analytics dashboard, behind `enable_analytics_consumption`
(dust-only).
Renders the overview from the endpoint below it in the stack and nothing else yet — chart, attribution table and filters follow.

- add `AnalyticsConsumptionPage` under `analytics/consumption`
- add `enable_analytics_consumption` feature flag.
- the page fetches the `analytics/consumption/overview` endpoint (from PR below) only

## Tests

Checked the new page renders locally:
Screenshot 2026-08-06 at 10 16 22

## Risk

Low. Safe to rollback.

## Deploy Plan

Front only.

PR #30080, opened 2026-08-06
[front] feat: consumption analytics overview endpoint

sfriquet:
## Description

Add GET `/api/w/:wId/analytics/consumption/overview`: returns top level metrics used.
Currently returns the values that renders the header:
- the resolved period `startDate/endDate`
- number of users / active users
- last record timestamp.

## Tests

ut + tested with next PR adding the front

```
curl -s "http://localhost:3000/api/w/AYjP6ks68g/analytics/consumption/overview?period=cycle" --cookie "workos_session=$DUST_SESSION" | jq
{
  "period": {
    "startDate": "2026-08-01T00:00:00.000Z",
    "endDate": "2026-09-01T00:00:00.000Z"
  },
  "members": {
    "active": 1,
    "total": 1
  },
  "lastRecordAt": "2026-08-05T11:29:22.164Z"
}
```

## Risk

Low. Safe to rollback.

## Deploy Plan

Front only

PR #30083, opened 2026-08-06
[front] feat: consumption analytics period selector

sfriquet:
## Description

Adds the period dropdown to the new Analytics page: "This cycle" (the default)
plus 7 / 30 / 90 day windows, matching the two modes the endpoints accept.

## Tests

Screenshot 2026-08-06 at 10 18 43
Screenshot 2026-08-06 at 10 18 52

## Risk

Low. Safe to rollback.

## Deploy Plan

Front only.


sfriquet:

- #30080
    - #30082
        - #30083 ◀
            - #30084
                - #30085
                    - #30086
                        - #30087
                            - #30088

front/lib/analytics/consumption_period.ts
```diff
@@ -0,0 +1,59 @@
...
+export type ConsumptionPeriodSelection =
+  | { kind: "cycle" }
+  | { kind: "days"; days: ConsumptionPeriodDays };
+
+export const DEFAULT_CONSUMPTION_PERIOD: ConsumptionPeriodSelection = {
+  kind: "cycle",
+};
+
+export const CONSUMPTION_PERIOD_OPTIONS: ConsumptionPeriodSelection[] = [
+  { kind: "cycle" },
```
avervaet, line 15: For user not on the credit based pricing we might want to hide CONSUMPTION_PERIOD_DAY_OPTIONS as it doesn't make sense for them if I understood properly the implem from https://github.com/dust-tt/dust/pull/30079

avervaet (approved):

front/lib/analytics/consumption_period.ts
```diff
@@ -0,0 +1,59 @@
...
+    ) ?? null
+  );
+}
+
+/**
+ * Query string shared by every consumption endpoint. Kept in one place so the
+ * SWR keys of the different widgets stay in sync — two widgets disagreeing on
+ * the window would show numbers that do not add up.
+ */
+export function consumptionQueryString(
```
aubin-tchoi, line 51: nit: should live near the hook that consumes it, would even inline it tbh

front/lib/analytics/consumption_period.ts
```diff
@@ -0,0 +1,59 @@
+export const CONSUMPTION_PERIOD_DAY_OPTIONS = [7, 30, 90] as const;
```
aubin-tchoi, line 1: This file could be in lib rather than components

aubin-tchoi (approved):

front/lib/analytics/consumption_period.ts
sfriquet: if not on CP the days option (last 7|30|90) might actually make more sense than "this cycle" no ? I think the days option are useful to everyone

front/lib/analytics/consumption_period.ts
sfriquet: this is going to be reused by several hooks in the next PRs

aubin-tchoi: ack, makes sense then 👍🏻

flvndvd (approved):
LGTM

front/lib/analytics/consumption_period.ts
sfriquet: moving to `lib/analytics/consumption_period.ts`

PR #30092, opened 2026-08-06
Fix: fallback to non speaker attributed content for MS 365

Fraggle:
## Description

The MS Teams transcript tool fetched content exclusively as `text/vtt` (speaker-attributed), which requires a tenant-level permission not universally granted. Tenants with that permission disabled return a "Speaker-attributed transcript content is disabled for this tenant" error, causing the tool to fail entirely.

The handler now tries `text/vtt` first and, on that specific error, falls back to `application/vnd.microsoft.graph.transcript+text` (non-speaker-attributed). The meeting/transcript ID resolution is split into its own `try/catch` for cleaner error attribution. When the fallback format is used, the response prefixes a note so the model is aware speaker attribution is unavailable.

## Tests

Couldn't test.

## Risk

Low. The change is additive: the happy path (`text/vtt` succeeds) is identical to before. The fallback only triggers on a specific error string from the Graph API. Front-only, no DB changes.

## Deploy Plan

Deploy front.

### DIFF
--- a/front/lib/api/actions/servers/microsoft_teams/tools/index.ts
+++ b/front/lib/api/actions/servers/microsoft_teams/tools/index.ts
@@ -673,6 +673,9 @@ const handlers: ToolHandlers<typeof MICROSOFT_TEAMS_TOOLS_METADATA> = {
       );
     }
 
+    let meetingId: string | undefined;
+    let transcriptId: string | undefined;
+
     try {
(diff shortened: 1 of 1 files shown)


Fraggle:

- #30092 ◀

aubin-tchoi (approved):
nice, thanks!

PR #30093, opened 2026-08-06
[front] refactor: post activation nudges directly instead of firing a webhook trigger

id13:
## Description

This PR removes the need to have a trigger to send activation nudges, instead posting the messages directly from the poke plugging.

Not exposing any codepath to a public surface also has the benefit of closing a small loophole where you could potentially forge a webhook to get a nudge pod and some freeloading, but under nudge msg tool's limitations (no personal creds no stakes) it's more of a theoretical one.

## Tests

Tested locally.

## Risk

Low
Blast radius: activation pods + auto nudge

## Deploy Plan

Deploy front


id13:

- #30093 ◀
    - #30099

front/lib/api/activation/nudge.ts
```diff
@@ -1,19 +1,40 @@
...
+import type { AgentConfigurationType } from "@app/types/assistant/agent";
+import { GLOBAL_AGENTS_SID } from "@app/types/assistant/assistant";
+import { ACTIVATION_NUDGE_ORIGIN } from "@app/types/assistant/conversation";
+import type { Result } from "@app/types/shared/result";
+import { Err, Ok } from "@app/types/shared/result";
+import { removeNulls } from "@app/types/shared/utils/general";
 import isNumber from "lodash/isNumber";
-import uniq from "lodash/uniq";
+
+const ACTIVATION_NUDGE_PROMPT = "Run the Dust Training workflow.";
```
frankaloia: Nit: Based on latest UX changes, my PR renamed this "Dust Learning" workflow

front/lib/resources/activation_nudge_resource.ts
```diff
@@ -1,140 +0,0 @@
...
-import { makeSId } from "@app/lib/resources/string_ids";
-import type { TriggerResource } from "@app/lib/resources/trigger_resource";
-import type { ModelId } from "@app/types/shared/model_id";
-import type { Result } from "@app/types/shared/result";
-import { Err, Ok } from "@app/types/shared/result";
-import { normalizeError } from "@app/types/shared/utils/error_utils";
-import type { Attributes, ModelStatic, Transaction } from "sequelize";
-
-// eslint-disable-next-line @typescript-eslint/no-unsafe-declaration-merging
-export interface ActivationNudgeResource
```
frankaloia, line 16: Is it useful to maintain this for metrics purposes to track what nudges we actually sent?

frankaloia (commented):
My only real concern here is I think @xebeohp mentioned that we might use the trigger in a conditional check to determine if the notification should be sent?

xebeohp:
> My only real concern here is I think @xebeohp mentioned that we might use the trigger in a conditional check to determine if the notification should be sent?

Yeah, here JD changed the notification to check for `ACTIVATION_NUDGE_ORIGIN` which I believe is not enough. But I should be able to use something else to gate notifications.

If I notice issues I can work on it as a bug fix, should not be a blocker for this.

front/lib/api/activation/nudge.ts
frankaloia: Updated — prompt is now `Run the Dust Learning workflow.`

front/lib/resources/activation_nudge_resource.ts
frankaloia: I don't think we need the separate nudge rows for this. Nudge history is now the pod conversations Dust opened with `origin: ACTIVATION_NUDGE_ORIGIN` — same signal the eligibility gates already use (`listNudgeConversationTimestamps` / `openingUserMessageOrigin`). That should be enough for "what did we actually send" without a parallel ledger.

frankaloia (commented):
Rebased onto main and addressed the open comments:

- Renamed the nudge prompt to **Dust Learning**
- Kept `workAreas` / `activationPlaybook` on `ActivationNudgeContext` in the direct-post path (and in the `<dust_activation>` block / skill wording)
- On metrics: leaning on `ACTIVATION_NUDGE_ORIGIN` conversations rather than keeping `ActivationNudgeResource`
- On notification gating: leaving as-is per @xebeohp — not a blocker; can follow up if the origin check proves insufficient

frankaloia (approved):
LGTM - Updated based on my comments + rebased

PR #30096, opened 2026-08-06
[Front] Sidebar fixes

ClementAupiais:
## Description

Two sidebar fixes:

1. **Collapsible space sections**: Open Spaces and Restricted Spaces sections in the spaces sidebar are now `NavigationListCollapsibleSection` headers (with a chevron), matching the conversations sidebar pattern. The "New" button moves into the section's action slot. Collapsed state persists per section via `useSidebarSectionCollapsed`. The Administration section loses its label entirely (its system space is self-describing), expressed as `label: null` meaning "render spaces with no section header". A `SpaceSection` component is extracted so each section can own its collapsed state independently (hooks can't be called inside a `map`).

2. **Restore Admin tab side padding**: A previous commit made `NavTabPillContent` `display:contents`, which caused margins set on it to have no effect — the Admin tab's sub-navigation rendered flush against the sidebar edge. The fix moves `mx-sidebar-side-spacing` from `NavTabPillContent` onto the inner `NavigationList`, consistent with how the Conversations and Knowledge tabs already handle it.

## Tests

Manually verified: space sections collapse/expand and persist state; Admin tab sub-navigation has correct side spacing; pod scrollability fix is unaffected.

## Risk

Low — UI-only changes, no data or API changes.

## Deploy Plan

Deploy front.

### DIFF
--- a/front/components/navigation/NavigationSidebar.tsx
+++ b/front/components/navigation/NavigationSidebar.tsx
@@ -151,12 +151,11 @@ export const NavigationSidebar = React.forwardRef<
               </div>
             </NavTabPillList>
             {navs.map((tab) => (
-              <NavTabPillContent
-                key={tab.id}
-                value={tab.id}
-                className="mx-sidebar-side-spacing"
-              >
-                <NavigationList>
+              // NavTabPillContent is display:contents, so it generates no box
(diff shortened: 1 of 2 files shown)


aubin-tchoi (approved):
LGTM

PR #30102, opened 2026-08-06
[front] fix: persist tool configs when copy-pasting skill instructions

aubin-tchoi:
## Description

Closes https://github.com/dust-tt/tasks/issues/9950

When we copy the instructions of a skill and paste into another Skill Builder, the knowledge and tool mentions are copied visually but then if we save they are not saved in the backend state.

This PR fixes that, the react-hook-form state was not properly updated in this case.

## Tests

- Added a test.
- Tested using the preview.

## Risk

- Low. Only affects the copy paste, which is quite niche.

## Deploy Plan

- Deploy front.


fontanierh (approved):
LGTM

fabiencelier (approved):

PR #30104, opened 2026-08-06
Record a Pod function as durable when it calls a Dust tool as fast

fontanierh:
## Description

A fast function is published on the promise that it does not call Dust tools through `dsbx tools`, and #30040 refuses the call when it breaks that promise. That refusal is also the only evidence the declaration was wrong, and today the evidence goes nowhere: the invocation fails, and the next one fails identically, because the declaration that caused it is still sitting on the function.

This records the function as `durable` from the refusal path. The invocation that triggered it still fails, since the tool call was already refused, but every later one runs durably and works. A mislabelled function starts behaving without the publisher having to notice anything.

The write is guarded on the mode still being `fast`, so concurrent invocations that each break the promise converge on a single write, and a function whose publisher has already moved it is left alone. It is never awaited by the request: the refusal is already decided, and a slow or failing write must not delay it or change what it returns.

Because a publish restates the mode (#30038), this is an override the publisher clears rather than a permanent verdict. A republished bundle that still calls tools is simply recorded again.

This also logs inline execution failures. The other case worth healing is a fast function that runs past its ceiling, but a timeout currently arrives as an ordinary provider failure with nothing marking it as a timeout, and one slow run is far weaker evidence than a refused tool call: a slow upstream or a loaded sandbox looks identical to a genuinely long function. Logging the whole class first means the decision can be made from numbers rather than from a guess.

## Tests

Added an endpoint test that the function is recorded as `durable` after its tool call is refused, and one that a durable function whose tool call succeeds keeps its mode.

The token test factory now publishes as `fast` when it mints a no-tools token, since that is the only combination that occurs.

Not exercised against a live pod.

## Risk

Low. The only path that reaches it is the 403 that #30040 already returns, which no function hits today since none is published `fast`. Worst case is a function losing its fast path, which costs latency and nothing else, and a republish restores it.

## Deploy Plan

None.

### DIFF
--- a/front/lib/resources/sandbox_function_resource.ts
+++ b/front/lib/resources/sandbox_function_resource.ts
@@ -434,6 +434,46 @@ export class SandboxFunctionResource extends BaseResource<SandboxFunctionModel> 
     return new Ok(sandboxFunctions.length);
   }
 
+  /**
+   * Make a fast function durable, after it did something only a durable function can do.
+   *
+   * A fast function is published on the promise that it does not call Dust tools. When it breaks
+   * that promise the invocation fails, and without this the next one fails the same way: the
+   * publisher's declaration is wrong and nothing on the platform knows it. Recording durable here
+   * costs that function its fast path and makes every later invocation work.
+   *
+   * The update is guarded on the mode still being fast, so concurrent invocations that each break
+   * the promise converge on one write, and a function whose publisher has already moved it is left
+   * alone. Returns whether this call is the one that moved it.
+   *
+   * A re-publish restates the mode, so this is an override the publisher can clear rather than a
+   * permanent verdict. If the republished bundle still calls tools, it is simply made durable
+   * again.
+   */
+  async makeDurable(auth: Authenticator): Promise<boolean> {
+    const [affectedCount, rows] = await this.model.update(
+      { executionMode: "durable" },
+      {
+        where: {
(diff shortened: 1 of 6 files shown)


front/lib/resources/sandbox_function_resource.ts
```diff
@@ -434,6 +434,45 @@ export class SandboxFunctionResource extends BaseResource<SandboxFunctionModel>
...
+   * costs that function its fast path and makes every later invocation work.
+   *
+   * The update is guarded on the mode still being fast, so concurrent invocations that each break
+   * the promise converge on one write, and a function whose publisher has already moved it is left
+   * alone. Returns whether this call is the one that moved it.
+   *
+   * A re-publish restates the mode, so this is an override the publisher can clear rather than a
+   * permanent verdict. If the republished bundle still calls tools, it is simply demoted again.
+   */
+  async demoteToDurable(auth: Authenticator): Promise<boolean> {
```
davidebbo: nit: I don't really like the term 'demote' here, which imply it's less good. I'd just call it makeDurable. But feel free to ignore! :)

davidebbo (approved):
LGTM

front/lib/resources/sandbox_function_resource.ts
fontanierh: fair !

PR #30105, opened 2026-08-06
[dust-hive] enh: infer current hive for dh aliases

aubin-tchoi:
## Description

Running `dhb` or `dhdb` from a hive currently still requires repeating the hive name. These aliases now infer the registered hive from the current worktree when called without arguments, opening the app or connecting to `dust_front` for that hive.

Explicit hive queries, database selection, and psql arguments remain unchanged. Calls without arguments outside a registered hive still show usage.

## Tests

- Tested locally in Bash and Zsh.

## Risk

- Low.

## Deploy Plan

- No deploy.


fontanierh (approved):
LGTM

PR #30107, opened 2026-08-06
[front] enh: suggest customizing remote MCP view name on conflict

aubin-tchoi:
## Description

Remote MCP server creation can fail when the server-provided name conflicts with an existing server.
We actually create the server during the flow and only allow changing the name afterwards. As a consequence users cannot choose a different name during the setup, making them unable to actually add the server.

This PR adds an optional name field to remote MCP server creation. The custom name is then used for conflict validation and persisted on the system and global views.

## Tests

- Updated the existing tests.

## Risk

- Low.

## Deploy Plan

- Deploy front.


Fraggle (approved):
Thanks

zmarouf (commented):
LGTM.
Thanks!

PR #30108, opened 2026-08-06
Gate agent_memory record/edit writes behind the user_memory flag 

Inesvvv:
## Description

Gated agent_memory record and edit tools behind the user_memory feature flag to stop further expansion of agent_memory and to help users transition from agent_memory to user_memory

## Tests

Tested locally 

## Risk

Low, preventing additional usage of agent_memory but not affecting existing usage 

## Deploy Plan

Deploy front

### DIFF
--- a/front/lib/api/actions/servers/agent_memory/tools/index.ts
+++ b/front/lib/api/actions/servers/agent_memory/tools/index.ts
@@ -1,6 +1,7 @@
 import { MCPError } from "@app/lib/actions/mcp_errors";
 import type { ToolHandlers } from "@app/lib/actions/mcp_internal_actions/tool_definition";
 import { buildTools } from "@app/lib/actions/mcp_internal_actions/tool_definition";
+import { getPrefixedToolName } from "@app/lib/actions/tool_name_utils";
 import { isAgentLoopRunContext } from "@app/lib/actions/types";
 import {
   AGENT_MEMORY_COMPACT_TOOL_NAME,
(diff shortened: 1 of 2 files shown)


front/lib/api/actions/servers/agent_memory/tools/index.ts
```diff
@@ -10,12 +10,29 @@ import {
...
+import { hasFeatureFlag } from "@app/lib/auth";
 import { AgentMemoryResource } from "@app/lib/resources/agent_memory_resource";
 import type { Result } from "@app/types/shared/result";
 import { Err, Ok } from "@app/types/shared/result";
 import type { CallToolResult } from "@modelcontextprotocol/sdk/types.js";
 import assert from "assert";
 
+const AGENT_MEMORY_WRITE_DISABLED_MESSAGE =
+  "Agent Memory is being deprecated, add and edit new memories directly in " +
+  "User Memory in Personal Settings → Memory";
```
aubin-tchoi: this message looks like it's meant for a user, here we turn to the agent, who does not have access to the UI
it doesn't need to know that agent memory is being deprecated and we need to prompt it to use the user memory tools instead

front/lib/api/actions/servers/agent_memory/tools/index.ts
```diff
@@ -10,12 +10,29 @@ import {
...
+import type { Authenticator } from "@app/lib/auth";
+import { hasFeatureFlag } from "@app/lib/auth";
 import { AgentMemoryResource } from "@app/lib/resources/agent_memory_resource";
 import type { Result } from "@app/types/shared/result";
 import { Err, Ok } from "@app/types/shared/result";
 import type { CallToolResult } from "@modelcontextprotocol/sdk/types.js";
 import assert from "assert";
 
+const AGENT_MEMORY_WRITE_DISABLED_MESSAGE =
+  "This tool is disabled. Use the user_memory tools to add or edit the user's " +
```
aubin-tchoi: can we be a bit more precise on which tool to use, how and why? 🙏🏻

aubin-tchoi (approved):
LGTM

PR #30109, opened 2026-08-06
[Activation] Expose UIView on conversations

xebeohp:
## Description

Expose the Activation Pod context needed for conversation clients to select the compact conversation UI.

- Persist `isCompactUIView` on `ActivationPodResource`, defaulting to `false` for callers that do not provide it.
- Set `isCompactUIView: true` when the activation management plugin provisions a training pod.
- Add a `UiView` conversation context value and have `ConversationViewer` pass `"compact"` for conversations in the Activation Pod and `"standard"` otherwise.

## Tests

- Updated `front/lib/api/poke/plugins/spaces/activation_management.test.ts` to verify that provisioned Activation Pods are created with `isCompactUIView: true`.

## Risk

Low. The compact view is selected only for conversations in the Activation Pod. Other conversations continue to use the standard view, and the persisted flag defaults to `false` for existing callers. The breadcrumb change also avoids labeling a space as “For you” unless the Activation pod has the Activation skill is enabled.

## Deploy Plan

Deploy `front`.


xebeohp:

- #29996
    - #30109 ◀
        - #30242
            - #30258

front-api/public/swagger.json
```diff
@@ -10640,6 +10640,10 @@
               "type": "string"
             }
           },
+          "isCompactUIView": {
+            "type": "boolean",
+            "description": "Server-resolved compact UI display hint for the conversation (currently only true for conversations in activation pods)"
```
frankaloia: nit: I don't think we should mention activation here, comment likely won't be updated in the future

front/lib/resources/conversation_resource.ts
```diff
@@ -552,6 +560,11 @@ export class ConversationResource extends BaseResource<ConversationModel> {
     return this._space;
   }
 
+  // Server-resolved compact UI display. Currently only used for activation pods.
+  get isCompactUIView(): boolean {
+    return this._activationPod?.isCompactUIView ?? false;
+  }
+
```
frankaloia: Do we need to couple this so directly with activation pod?

front-api/public/swagger.json
xebeohp: Fixed for all swagger docs.

front/lib/resources/conversation_resource.ts
xebeohp: Another option is moving the column to 'ConversationModel' to decouple compact UI with activation?

frankaloia: Hmm maybe we want to avoid any conversation model or conversation resource plumbing since it's not a property of conversation really. Should we just be passing this as a UX prompt as opposed to a public api configuration?

xebeohp: Ok, yeah that makes sense. I'll make the change.

frankaloia (commented):
Better, thank you. 

Do we need the API to return this? I think we can just be opinionated on it. The client already has activationPodId from useActivationPod and conversation.spaceId, so isCompactUIView is conversation.spaceId === activationPodId. 

We also want to make sure that in the end this is pod-scoped, not user scoped, so avoiding having this view for _every_ conversation the user has, even outside the pod.

I'd also probably end up going with the front end enum which will age better uiView: "standard" | "compact" ages better

frankaloia (approved):

PR #30110, opened 2026-08-06
[front] Describe the full /files layout in the Computer skill

fontanierh:
## Description

The sandbox skill described `/files/conversation` and `/files/pod` in two separate sections and left out things agents keep getting wrong: that both paths are symlinks (so `find` and `rg` skip them unless told to follow), that they are GCS backed (persistent and user-visible, but slow), and that anything outside those two mounts, including other paths under `/files/`, is gone when the sandbox is recycled.

`buildConversationFilesSection` and `buildProjectFilesSection` are now a single `buildFilesSection({ hasPod })`.

## Tests

Existing sandbox skill tests cover both the Pod and non-Pod shapes.

## Risk

Prompt content, no code behavior change. It sits in the cacheable system prompt prefix, so live conversations take a prompt cache miss on their first turn after deploy.

## Deploy Plan

Standard deploy.

### DIFF
--- a/front/lib/resources/skill/code_defined/global/sandbox.ts
+++ b/front/lib/resources/skill/code_defined/global/sandbox.ts
@@ -42,20 +42,55 @@ function buildSandboxInstructionProse({
   return instructions.join(" ");
 }
 
-function buildConversationFilesSection(): string {
-  return `#### Sandbox Conversation File System
+function buildFilesSection({ hasPod }: { hasPod: boolean }): string {
+  const podMountLine = hasPod
+    ? `
+- \`/files/pod\` — the Pod's file system, shared across every conversation
+  in the same Pod. Anything you write or delete here is visible to the
+  other conversations of that Pod.`
+    : "";
 
-The conversation file system is mounted read-write inside the sandbox at
-\`/files/conversation\`. This is the canonical surface for navigating,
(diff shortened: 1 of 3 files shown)


front/lib/resources/skill/code_defined/global/sandbox.ts
```diff
@@ -42,20 +42,56 @@ function buildSandboxInstructionProse({
...
+  const podUsageSection = hasPod
+    ? `
+
+This conversation belongs to a Pod, so use \`/files/pod\` for files that
+belong to the Pod as a whole (specs, knowledge bases, shared scripts,
+recurring data sets), and \`/files/conversation\` for per-conversation
+artifacts. When both are relevant, prefer reading from \`/files/pod\` and
+writing deliverables to \`/files/conversation\` unless the user has asked
+you to update the Pod's files specifically. Pod files are also exposed by
+the \`files\` MCP server under scoped paths like \`pod-{podId}/<rel>\`.`
```
Fraggle: (loosely held) Might be a bit directive, no ?

fontanierh: Yep agreed

Fraggle (approved):
LGTM

PR #30114, opened 2026-08-06
[front] enh: treat Frames in a Pod as apps

davidebbo:
## Problem

Creating a Frame with functions in a Pod produced a structure that did not reflect what was being built. The Frame's source landed in the conversation file system, while the pod functions behind it went flat into the Pod root — two file systems, no folder tying them together, and the Frame scoped to a single conversation even though the functions it called were shared across the whole Pod.

Separately, "Create a frame that implements a simple task list" produced a `useState` array. The data vanished on reload, and the user only found out after entering real tasks — they had to re-prompt with "a *persisted* task list" to get the right thing.

## What changes

**Frames in a Pod are Pod apps.** They live in the Pod's shared file system, each in its own folder holding the Frame's source, the functions behind it, and their shared database schemas:

```
/files/pod-<podId>/
  MyApp/
    MyApp.tsx
    functions/
      list-notes.ts
      post-note.ts
      lib/
        notes.ts
    databases/
      notes.db.ts
```

The folder is created even for a Frame that is a single file today, so nothing has to move a second time when functions are added later. Since `create_interactive_content_file` always creates the source in the conversation, the instructions spell out the create → `files__move` → publish-from-Pod-path flow (`copy` is rejected for Frame files; `move` updates the `FileResource` mount so publishing from the new path works).

**Persistence is the default.** A new "Where The Frame's Data Lives" section names the trigger behaviourally — if the Frame lets people add, edit, check off, reorder, delete, assign, comment, vote or upload, the data goes in a Pod database behind pod functions — and states outright that the user does not have to say "persisted" for that to be what they meant. It also bounds the rule, so tab/filter/sort/expanded-row state stays in the component.

This had to go in the Frames instructions rather than the Pod Functions skill: equipped skills are fetched with `withInstructions: false`, so the model only sees that skill's name and description until it calls `enable_skill` — the decision has to be reachable before the model knows functions are involved. Two supporting changes make it actionable: the section names the `Pod Functions` skill as an exact backtick literal, and that skill's `agentFacingDescription` (the only text visible while equipped) now leads with "this is how a Frame stores data".

## Implementation notes

`fetchInstructions` composes the prose via `buildInteractiveContentInstructions({ hasComputer, isPod, hasPodFunctions, podFunctionsSkillName })` instead of selecting from fixed constants — three independent booleans made eight exported constants untenable. Legacy conversations still short-circuit to `INTERACTIVE_CONTENT_INSTRUCTIONS`: without the path-based files tools there is no way to promote a Frame into the Pod.

The storage section is gated on the `sandbox_functions` flag; the app-folder layout is not, since it is only file placement. A Pod conversation in a workspace without the flag gets the layout and no dangling pointer to a skill it cannot enable.

The Pod-app layout also appears in the Pod Functions skill, trimmed to the function side and deferring to the Frames skill for the Frame move/publish mechanics, so the two do not teach the same flow and drift apart.

## Drive-by fix

The publish instructions told the model to pass `path: conversation-<conversationId>` — the directory. `splitFrameEntryScopedPath` takes `dirname`, so that yields root `"."` and errors out; the tool's own schema correctly asks for the entry file. Already broken, and it would have blocked the new Pod flow too, so both variants and the worked example now pass the entry file path.

## Test plan

- `npx tsgo --noEmit` clean, biome clean.
- 169 tests pass across `lib/resources/skill`, `interactive_content`, `sandbox_functions`, `publish_frame` and `skills_rendering`.
- Four new cases in `frames.test.ts`: Pod gets both sections; non-Pod gets neither; Pod without the `sandbox_functions` flag gets the layout but not the storage section; a legacy Pod conversation keeps the retrieve/file-id flow.

### DIFF
--- a/front/lib/api/actions/servers/interactive_content/instructions.ts
+++ b/front/lib/api/actions/servers/interactive_content/instructions.ts
@@ -7,6 +7,7 @@ import {
   FILES_CAT_ACTION_NAME,
   FILES_EDIT_ACTION_NAME,
   FILES_LIST_ACTION_NAME,
+  FILES_MOVE_ACTION_NAME,
   FILES_RESOLVE_ACTION_NAME,
   FILES_SERVER_NAME,
 } from "@app/lib/api/actions/servers/files/metadata";
@@ -41,6 +42,10 @@ const FILES_LIST_TOOL = getPrefixedToolName(
 const FILES_RESOLVE_TOOL = getPrefixedToolName(
   FILES_SERVER_NAME,
   FILES_RESOLVE_ACTION_NAME
+);
+const FILES_MOVE_TOOL = getPrefixedToolName(
+  FILES_SERVER_NAME,
+  FILES_MOVE_ACTION_NAME
 );
 
 const UPDATING_SECTION_LEGACY = `\
@@ -89,9 +94,9 @@ Create a Frame only when the content does not exist yet. When the user asks for 
 const UPDATING_SECTION_COMPUTER_FIRST = `\
 ### Updating Existing Files (edit the source, then publish):
 
-After a Frame is created, its source file is already mounted in the Computer at \`/files/conversation-<conversationId>/<FrameName>.tsx\`. To update the Frame:
-1. Edit that file in place with your file tools, changing only the parts that need to change. Do not rewrite the whole file for partial changes. When the Computer is not available, edit it with \`${FILES_EDIT_TOOL}\` using the scoped path \`conversation-<conversationId>/<FrameName>.tsx\`.
-2. Publish with \`${PUBLISH_INTERACTIVE_CONTENT_FILE_TOOL_NAME}\` using \`path: conversation-<conversationId>\` (the directory holding the file, not a subdirectory).
+After a Frame is created, its source file is already mounted in the Computer at \`/files/conversation-<conversationId>/<FrameName>.tsx\`. A Frame whose source was moved elsewhere, for example into a Pod app folder, is mounted at its current path instead; resolve it with \`${FILES_RESOLVE_TOOL}\` when unsure. To update the Frame:
+1. Edit that file in place with your file tools, changing only the parts that need to change. Do not rewrite the whole file for partial changes. When the Computer is not available, edit it with \`${FILES_EDIT_TOOL}\` using its scoped path, e.g. \`conversation-<conversationId>/<FrameName>.tsx\`.
+2. Publish with \`${PUBLISH_INTERACTIVE_CONTENT_FILE_TOOL_NAME}\`, passing \`path\` set to the source file's own scoped path (the entry file itself, not the directory holding it), e.g. \`conversation-<conversationId>/<FrameName>.tsx\`.
 
 If an edit fails because the text to replace is not found, the file differs from what you remember: re-read it and retry the targeted edit. Never respond to a failed match by resending the whole file.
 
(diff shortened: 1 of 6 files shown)


front/lib/api/actions/servers/interactive_content/instructions.ts
```diff
@@ -119,19 +124,69 @@ ${FILES_EDIT_TOOL}({
 })
 ${PUBLISH_INTERACTIVE_CONTENT_FILE_TOOL_NAME}({
   file_id: "fil_abc123",
-  path: "conversation-<conversationId>",
+  path: "conversation-<conversationId>/Dashboard.tsx",
```
davidebbo, line 127: This was a pre-existing issue, and not directly related to this change.

front/lib/api/actions/servers/interactive_content/instructions.ts
```diff
@@ -119,19 +124,65 @@ ${FILES_EDIT_TOOL}({
...
 ${PUBLISH_PARAGRAPH}
 `;
 
+// Pod conversations only. A Pod's Frames belong to the Pod rather than to the conversation that
+// happened to create them, and each one gets its own folder so the pod functions and database
+// schema it may grow later land next to it instead of forcing a second move.
+const POD_APP_SECTION = `\
+### Frames In A Pod
+
+This conversation belongs to a Pod, so its Frames belong to the Pod, not to this conversation: every conversation in the Pod sees the same Frame, and it outlives this one.
```
fontanierh, line 140: Technically, even if the conversation is in a pod, there could be legitimate use cases to have a frame inside of the conversation and not the pod right ? You might want to iterate, produce several versions of something, and only once you are happy move the frame to the pod ?

fontanierh (approved):
LGTM

front/lib/resources/skill/code_defined/global/pod_functions.ts
```diff
@@ -39,11 +68,36 @@ Reach for a pod function instead of inline code or an ad hoc tool call when any
...
   "Persisting state across calls" below).
 
+#### Laying out a Pod app
+
+A Frame and the pod functions behind it are one app. Keep the whole app in a single folder on the
+Pod file system rather than splitting it between the conversation and the Pod root:
+
+\`\`\`
+/files/pod-<podId>/
+  MyApp/
```
fontanierh, line 78: Maybe we could hint at a "lib" folder for shared code that are not directly functions ? like cross-function helpers and stuff like that

front/lib/api/actions/servers/interactive_content/instructions.ts
davidebbo: Yeah, but if you're iterating on a frame that has functions, you already end up with the functions being in the pod from the get go. Does it really make sense to have just the frame file in the convo? Also, a new frame/app that is WIP should not be really harmful.

I mean, ideally, we'd have proper staging support, but given the yolo state of things, it feels cleaner to keep the files together from the get go, no?

front/lib/resources/skill/code_defined/global/pod_functions.ts
davidebbo: Yes, maybe. So I'd say more `MyApp/functions/lib` and not `MyApp/lib`. Is that what you had in mind? Basically all the server code in one folder. I'll add that.

davidebbo: Added lib folder guidance

fontanierh: Makes sense yes

front/lib/api/actions/servers/interactive_content/instructions.ts
fontanierh: yes as soon as you're making functions, it makes sense to have the frame in the pod

PR #30116, opened 2026-08-06
[front] fix: filter high stake tools on activation origin

id13:
## Description

I introduced a filter on highstake tools based on the presence of userMessage.user == null, which had unwanted ripple effects like a conversation created by pod_manager's create_conversation (which passes doNotAssociateuser which sets userMsg.user to null).

This PR instead filters based on the new activation nudge origin. Less elegant, but much easier to understand.

## Tests

Unit tests 

## Risk

Low

## Deploy Plan

- Deploy front


Fraggle (approved):
LGTM

PR #30118, opened 2026-08-06
fix(front): also display skills used in UsedBy button

achilleburah:
## Description

Adds skills (not just agents) to the "Used By" usage shown on Connected Data/Folders/Websites and the space landing page, matching what the Tools page has shown since #29819.

Also fixes real bugs found while building/reviewing this:
- Tools/Triggers rows on the space landing page showed 0 usage / never rendered.
- Auto-provisioned tools (Pods, Computer, Author Skills, ... — auto-added to every global/system space) were leaking unrelated agents/skills into the Tools usage row on the global space.
- Deletion guards (space delete, data source view delete) and the pre-delete warning dialog now correctly name skills, not just agents, when they block/warn deletion.

Refactors for consistency: moved usage orchestration out of route handlers into `lib/api/*` (`getSpaceCategoriesWithUsage`, `handleDeleteDataSourceView`), removed Apps-usage computation entirely (legacy feature, gated behind `legacy_dust_apps`, not worth computing), renamed `SkillUsageType` → `AgentsAndSkillsUsageType`.

## Tests

New/updated: `agent_data_sources.test.ts`, `spaces.test.ts` (skill-only deletion-guard case), `data_source_views/[dsvId]/index.test.ts` (new file — skill-only block + force bypass), `spaces/[spaceId]/index.test.ts` (Tools/Triggers usage, Triggers row appearing, and the auto-tools-exclusion regression).

Full regression sweep passing. `tsgo`/`biome` clean.

## Risk

- **Deletion guard behavior change**: skill usage now blocks deletion too (space + data source view), not just agents. Intentional correctness fix — both guard error messages and the pre-delete warning dialog now name the skill(s).
- **Auto-tools usage fix changes what the Tools row shows on global/system spaces**: previously inflated (counted every workspace-wide user of an auto tool like Pods/Computer), now accurate. Documented residual, non-blocking gap: two *regular* spaces sharing the same *manually*-added tool can still leak usage into each other (pre-existing — `getToolsUsage` groups by MCP server, not by view).
- **Import-cycle fix removes two Resource methods** (`DataSourceViewResource`/`DataSourceResource`.`getUsagesByAgents`); all known call sites updated in this PR.
- No schema/data changes, no migrations.
- **Rollback**: straightforward revert, no migrations, no feature flags.

## Deploy Plan

Standard deploy, no migrations, no feature flags. Safe to roll back by reverting the commit.

### DIFF
--- a/front/components/spaces/useUsedByDetailsSheets.tsx
+++ b/front/components/spaces/useUsedByDetailsSheets.tsx
@@ -1,0 +1,36 @@
+import { AgentDetailsSheet } from "@app/components/assistant/details/AgentDetailsSheet";
+import { SkillDetailsSheetById } from "@app/components/command_palette/SkillDetailsSheetById";
+import type { UserType, WorkspaceType } from "@app/types/user";
+import { useState } from "react";
+
+/**
+ * Shared "Used By" click-through state: clicking an agent or a skill in a `UsedByButton`
+ * dropdown opens the corresponding details sheet. Factored out of `AdminActionsList` so
+ * every other `UsedByButton` caller (Connections, space category rows, ...) doesn't have to
+ * duplicate the agentId/skillId state and the two sheets.
+ */
+export function useUsedByDetailsSheets(owner: WorkspaceType, user: UserType) {
+  const [agentId, setAgentId] = useState<string | null>(null);
+  const [skillId, setSkillId] = useState<string | null>(null);
+
+  return {
+    onAgentClick: setAgentId,
+    onSkillClick: setSkillId,
+    sheets: (
+      <>
+        <AgentDetailsSheet
+          owner={owner}
+          user={user}
+          agentId={agentId}
+          onClose={() => setAgentId(null)}
+        />
+        <SkillDetailsSheetById
+          owner={owner}
+          user={user}
+          skillId={skillId}
+          onClose={() => setSkillId(null)}
+        />
+      </>
+    ),
+  };
+}
--- a/front/lib/api/agent_data_sources.ts
+++ b/front/lib/api/agent_data_sources.ts
@@ -56,77 +56,98 @@ const agentAggregates: ProjectionAlias[] = (
   alias,
 ]);
 
+// Usage only needs each skill's identity, so skip the rest of the hydration.
+const SKILL_USAGE_HYDRATION = {
+  withInstructions: false,
+  withTools: false,
+  withFileAttachments: false,
+} as const;
+
 /**
- * Skills can attach a data source view directly as "knowledge" (independently of any
- * MCP server/tool configuration). Mirrors `fetchSkillsByMCPServer` in `agent_actions.ts`,
- * but reads `skill.dataSourceConfigurations` instead of `skill.mcpServerViews`.
+ * Buckets skills by the knowledge id they attached, dropping skills the current user may not
+ * see. Skills are fetched scoped to `requestedIds`, but each one carries all of its
+ * configurations, so ids we did not ask about are skipped here.
  */
-async function fetchSkillsByDataSource(auth: Authenticator): Promise<{
-  byDataSourceViewId: Map<ModelId, UsedBySkillType[]>;
-  byDataSourceId: Map<ModelId, UsedBySkillType[]>;
-}> {
-  const workspaceSkills = await SkillResource.listByWorkspace(auth, {
-    status: "active",
-    withInstructions: false,
-    withTools: false,
-    withFileAttachments: false,
-  });
-  const skills = auth.isAdmin()
-    ? workspaceSkills
-    : workspaceSkills.filter(
+function groupSkillsByKnowledgeId(
+  auth: Authenticator,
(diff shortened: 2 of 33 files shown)


front/lib/api/data_source_view.ts
```diff
@@ -335,14 +335,57 @@ export async function handlePatchDataSourceView(
...
+  }
+
+  await dataSourceView.delete(auth, { hardDelete: true });
+
+  return new Ok(undefined);
+}
+
 export type DataSourceViewWithUsage = DataSourceViewType & {
-  usage: AgentsUsageType | null;
+  usage: AgentsAndSkillsUsageType | null;
```
fabiencelier, line 382: FYI, when updating types returned by API calls, we must always consider front end lagging behind in version and make backward compatible changes.
The purpose is to support a customer who did not refresh their browser in a while for instance (generally a few days is considered the reasonable time frame)

Here it is just additive so it won't break the old front end so you are good :)

One thing worth noting though: as the count is [skill + agents] usage the old front end may show "used by X agents" where X is actually agents + skills.

Still ok to merge IMHO, just a note

front/lib/api/agent_data_sources.ts
```diff
@@ -49,13 +56,105 @@ const agentAggregates: ProjectionAlias[] = (
...
+    withFileAttachments: false,
+  });
+  const skills = auth.isAdmin()
+    ? workspaceSkills
+    : workspaceSkills.filter(
+        (skill) => skill.availability !== "editors" || skill.canWrite(auth)
+      );
+
+  const byDataSourceViewId = new Map<ModelId, UsedBySkillType[]>();
+  const byDataSourceId = new Map<ModelId, UsedBySkillType[]>();
```
fabiencelier: Instead of scanning all the skills, can we read directly in skill_data_source_configuration table to look for a list of datasource ids or datasource view ids ? This will avoid reading and fetching hundreds of skills when the datasources are actually only used in a bunch of them 

We have SkillResource.listByDataSourceViewIds which already does something similar, so maybe we can re-use it or adapt it a bit to support your use

fabiencelier (commented):
Looks good
One comment about performance of such retrieval which I think can be improved to use an index and fetch way lessa resources

front/lib/api/data_source_view.ts
achilleburah: ok noted, will keep that in mind thx !

front/lib/resources/skill/skill_resource.ts
```diff
@@ -1660,6 +1664,50 @@ export class SkillResource extends BaseResource<SkillConfigurationModel> {
...
       },
       onlyCustom: true,
+      ...opts,
+    });
+  }
+
+  /**
+   * List skills that use any of the given data source IDs.
+   */
+  static async listByDataSourceIds(
```
fabiencelier, line 1674: let's add one or 2 unit tests to make sure this one works as expected
it's generally good practice to unit tests the resources as they are shared brick re-used in mutliple places so we want to make sure they behave as expected.

No need for anything fancy but let's add one test similar to what is tested with 
```
describe("listByDataSourceViewIds"
```

fabiencelier (approved):
Looks good to me !

front/lib/resources/skill/skill_resource.ts
achilleburah: Sure !

achilleburah: done

front/components/spaces/useUsedByDetailsSheets.tsx
```diff
@@ -0,0 +1,36 @@
...
+import type { UserType, WorkspaceType } from "@app/types/user";
+import { useState } from "react";
+
+/**
+ * Shared "Used By" click-through state: clicking an agent or a skill in a `UsedByButton`
+ * dropdown opens the corresponding details sheet. Factored out of `AdminActionsList` so
+ * every other `UsedByButton` caller (Connections, space category rows, ...) doesn't have to
+ * duplicate the agentId/skillId state and the two sheets.
+ */
+export function useUsedByDetailsSheets(owner: WorkspaceType, user: UserType) {
```
aubin-tchoi: I don't think that this hook adds a lot of value tbh, in any case should not be in `component/spaces`

front/lib/api/spaces.ts
```diff
@@ -56,47 +182,61 @@ export async function softDeleteSpaceAndLaunchScrubWorkflow(
...
+  const dataSources = await DataSourceResource.listBySpace(auth, space);
+  const apps = await AppResource.listBySpace(auth, space);
+
+  if (!force) {
+    let blockedByUsage = false;
+    const agentNames: string[] = [];
+    const skillNames: string[] = [];
+
+    for (const view of dataSourceViews) {
+      const usage = await getDataSourceViewUsage({
```
aubin-tchoi: a useful refactor for later would be to turn `getDataSourceViewUsage` into a resource method

PR #30128, opened 2026-08-06
Remove JIT upsert in conversations.

Fraggle:
## Description

I'm thiiink we are good with the removal of skipDataSourceIndexing  because we have:
```
const canDoJIT = cf.snippet !== null && !isPastedFile(cf.contentType);
```
And snippet were only created when upserting to datasource (well, except pasted file but..).

One edge case is new convo without the Computer can't do query on CSV file. I think it's acceptable.

Removes the JIT (just-in-time) upsert of conversation files to data sources. Previously, `POST /files/:fileId` would eagerly call `getOrCreateConversationDataSourceFromFile` + `processAndUpsertToDataSource` for any upsertable `conversation`-use-case file, and `uploadFileToConversationDataSource` was also called from `query_tables_v2/helpers.ts` and `run_dust_app/helpers.ts` after generating tool output files. All of that is removed.

As a consequence, `skipDataSourceIndexing` is no longer meaningful and is dropped from `useCaseMetadata`, `ActionGeneratedFileType`, swagger schemas, and the `isSearchable` gate. `isSearchable` now relies purely on snippet presence and content type. Pasted files (`text/vnd.dust.attachment.pasted`) are also explicitly excluded from `canDoJIT`.

## Tests

Deleted the tests that covered the removed JIT upsert paths (`processAndUpsertToDataSource` invocation, invalid CSV 400, sandbox raw skip). Remaining tests updated to remove the now-obsolete `skipDataSourceIndexing` mock setup. Tested locally.

## Risk

Low. Pure deletion — no new code paths introduced. The JIT indexing was the only caller of `getOrCreateConversationDataSourceFromFile` for the `conversation` use case; removing it means conversation files are no longer eagerly indexed into a data source, which is the intended change. Rollback is a revert and redeploy.

## Deploy Plan

Deploy front + front-api. No SQL migrations, no connectors change.


Fraggle:

- #30128 ◀
    - #30154

front-api/routes/v1/w/[wId]/swagger_schemas.ts
```diff
@@ -472,9 +472,6 @@
  *           nullable: true
  *           description: Path of the plain-text sibling of this file inside the sandbox conversation mount (e.g. an audio transcript), when it has one.
  *           example: conversation/voice.processed.txt
- *         skipDataSourceIndexing:
```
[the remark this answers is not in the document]
Comment (Fraggle) [reply front-api/routes/v1/w/[wId]/swagger_schemas.ts:475]: Need to double check this one.

.github/workflows/build-and-test-front.yml
```diff
@@ -57,6 +57,7 @@ jobs:
           REDIS_CACHE_URI: "redis://localhost:5434"
           REDIS_URI: "redis://localhost:5434"
           NODE_ENV: test
+          NODE_OPTIONS: "--trace-deprecation" # Show stacktrace to eliminate deprecated calls.
```
flvndvd, line 60: Mainly useful locally, let's not ship this in our production build.

front-api/routes/v1/w/[wId]/swagger_schemas.ts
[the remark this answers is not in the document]
Comment (flvndvd) [reply front-api/routes/v1/w/[wId]/swagger_schemas.ts:475]: Definitely not used. IMHO safe to remove.

front/lib/api/assistant/conversation/attachments.ts
```diff
@@ -166,8 +167,8 @@ export function getAttachmentFromFileContentFragment(
 
   // Here, snippet not null is actually to detect file attachments that are prior to the JIT
   // actions, and differentiate them from the newer file attachments that do have a snippet.
-  // Former ones cannot be used in JIT.
-  const canDoJIT = cf.snippet !== null;
+  // Former ones cannot be used in JIT. For pasted files, we also do not support JIT actions.
+  const canDoJIT = cf.snippet !== null && !isPastedFile(cf.contentType);
```
flvndvd, line 171: Just to confirm that we are 100% sure that only old files had a snippet?

front/lib/api/assistant/conversation/attachments.ts
```diff
@@ -304,25 +299,19 @@ export function renderLargePasteXml({
...
-      : [
-          `version="${attachment.contentFragmentVersion}"`,
-          `isInProjectContext="${attachment.isInProjectContext}"`,
-          `isIncludable="${attachment.isIncludable}"`,
-          `isQueryable="${attachment.isQueryable}"`,
-          `isSearchable="${attachment.isSearchable}"`,
-        ]),
+    `version="${attachment.contentFragmentVersion}"`,
+    `isInProjectContext="${attachment.isInProjectContext}"`,
+    `isIncludable="${attachment.isIncludable}"`,
```
flvndvd, line 312: I thought the goal was to drop those boolean flags, no?

front/lib/api/assistant/jit_actions.test.ts
```diff
@@ -926,7 +926,9 @@ describe("getJITServers", () => {
   });
 
   describe("attachment-based servers", () => {
-    it("should include query_tables server when queryable attachments exist", async () => {
+    it("should include query_tables server when queryable attachments exist and Computer is disabled", async () => {
```
flvndvd, line 929: My understanding was that query table can only work if data is actually indexed in a data source? So this serve which need? Query table on connected data sources or folders? But not conversation-local, right?

front/lib/api/assistant/jit_actions.ts
```diff
@@ -132,13 +135,17 @@ async function getConditionalJITServers(
...
-  const queryTablesServer = await getQueryTablesServer(
-    auth,
-    conversation,
-    attachments,
-    autoInternalViews
-  );
-  servers.push(queryTablesServer);
+  // With Computer available, tabular files are handled via the sandbox instead of
+  // the legacy conversation query_tables JIT server.
+  if (!hasSandboxTools) {
```
flvndvd, line 140: Don't we still need the query table server for content node of type tabular?

front/lib/api/files/attachments.ts
```diff
@@ -77,67 +70,12 @@ export async function maybeUpsertFileAttachment(
             return new Ok(undefined);
           }
 
+          // Note from seb: this tell the system "you can do JIT actions on this file". (i know, it's not great)
+          // So we ignore later when doing canDoJIT checks.
           await fileResource.setSnippet(snippetRes.value);
```
flvndvd, line 75: OOC why not doing the same explicit pattern we had for `skipDataSourceIndexing` it does not have a huge footprint and was very explicit at least.

flvndvd (commented):
Looks mostly good, a few comments to make sure I'm clear on the end set of behaviors

.github/workflows/build-and-test-front.yml
Fraggle: I want to debug some .query in test that I don't repro locally.

front/lib/api/assistant/conversation/attachments.ts
Fraggle: Yes, that was the contract and I checked the remaining setSnippet() calls.

front/lib/api/assistant/conversation/attachments.ts
Fraggle: I didn't want to mix everything in one PR and make it hard to review.
This one remove the upserts, the next one will rework this (i need to pass the isNewFileExplorer from various places).

front/lib/api/assistant/jit_actions.test.ts
Fraggle: old uploaded files that are already indexed.

front/lib/api/assistant/jit_actions.ts
Fraggle: If we have the computer I think we just want to use the computer and stop mixing.

flvndvd: But then how does it work, if I paste a link to a google sheet? We will detect it and inline the content node, but we don't expose the tool to use it? I'm not sure to see how sandbox can then query it? `dsbx`?

front/lib/api/assistant/conversation/attachments.ts
flvndvd: Okok

front/lib/api/assistant/conversation/attachments.ts
flvndvd: 👍

.github/workflows/build-and-test-front.yml
flvndvd: Anyway pretty sure it will get overiden when deploying

front/lib/api/files/attachments.ts
Fraggle: I agree that the flag was more explicit but as we remove upsert and nothing is supposed to upload to the convo datasource, i'd rather keep the original pattern (having a snippet or not), i think it was an oversight to use it for pasted file.

I think medium term, we should just totally remove it and stop supporting the JIT at all on old files.

front/lib/api/assistant/jit_actions.ts
Fraggle: hum, let me double check that flow.

Fraggle: Good catch, updating code.

.github/workflows/build-and-test-front.yml
Fraggle: Yep, it seems so.

front/lib/api/assistant/jit_actions.test.ts
Fraggle: Related to your remark below, updating tests.

fontanierh (approved):

PR #30129, opened 2026-08-06
Load billed agent message consumption inputs

flvndvd:
As we start ingesting data into the new consumption analytics index, we need one clear entry point that loads everything required for a single agent message and answers the first important question: was this usage actually billed?

This PR adds that loading layer. It reads the authoritative message cost and the related runs, usage classification, attribution items, actions, step contents, skills, model, user, space, trigger, API key, tags, and agent ancestry. User and programmatic usage are included. Fully free usage, such as sidekick and activations, is ignored, as are messages that can still resume. If usage has not been classified yet, or billed usage has no message cost, we fail explicitly instead of producing incomplete analytics.

Agent ancestry follows the persisted `run_agent` and handover links in application code. Nested conversations are limited to a depth of four, **but** handovers stay in the same conversation and can create a longer chain, so analytics keeps the full chain from the root agent to the direct parent.

This PR does not create Elasticsearch documents, write to Elasticsearch, or trigger the analytics workflow yet. It gives the next activity a stable input with the billing decision already made and keeps Sequelize rows and relationship details behind the Resource and application layers.

### DIFF
--- a/front/lib/api/assistant/conversation/agentic_ancestors.ts
+++ b/front/lib/api/assistant/conversation/agentic_ancestors.ts
@@ -1,0 +1,59 @@
+import type { Authenticator } from "@app/lib/auth";
+import { ConversationResource } from "@app/lib/resources/conversation_resource";
+
+export type AgenticAncestor = {
+  agentConfigurationId: string;
+  agentMessageId: string;
+  conversation: ConversationResource;
+};
+
+/**
+ * Lists the agent messages that led to the current conversation, starting with the direct parent.
+ */
+export async function listAgenticAncestors(
+  auth: Authenticator,
+  conversation: ConversationResource,
+  {
+    agentMessageId,
+    maxAncestors,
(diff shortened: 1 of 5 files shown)


front/lib/api/assistant/conversation/agentic_ancestors.ts
```diff
@@ -0,0 +1,59 @@
...
+export type AgenticAncestor = {
+  agentConfigurationId: string;
+  agentMessageId: string;
+  conversation: ConversationResource;
+};
+
+/**
+ * Lists the agent messages that led to the current conversation, starting with the direct parent.
+ */
+export async function listAgenticAncestors(
```
flvndvd, line 13: Definitely not ideal, but IMHO better than using a `RECURSIVE CTE` in a raw SQL query. Will refactor the data model so it does not require a loop.

aubin-tchoi (approved):
LGTM

PR #30130, opened 2026-08-06
feat: allow fetching user groups at a given date

sfriquet:
## Description

Make it possible to fetch user group ids at a given point in time.
This is useful when we want to persist user memberships without being dependent on current time, e.g when re-indexing documents in ES, the index shouldn't change based on when the indexing runs.

The group memberships function already have a notion of `start` and `end` date so we only need to modify existing functions to filter by something else than `now`.

Done by adding an optional `at?: date` parameter to:
- `getActiveMemberships`
- `getActiveMembershipOfUserInWorkspace`
- `listUserGroupModelIdsInWorkspace`

## Tests

added uts.

## Risk

Safe to rollback.

## Deploy Plan

Front only

### DIFF
--- a/front/lib/resources/group_resource.ts
+++ b/front/lib/resources/group_resource.ts
@@ -1156,18 +1156,28 @@ export class GroupResource extends BaseResource<GroupModel> {
     return groups.filter((group) => group.canRead(auth));
   }
 
+  /**
+   * Group model ids the user was a member of at `at`, defaulting to now.
+   *
+   * Two caveats:
+   * - `global` groups are included regardless of `at` if requested (they are not historized).
+   * - `status` is mutated in place by suspend/restore, so a membership
+   *   suspended today is filtered out even for a past `at`.
+   */
   private static async listUserGroupModelIdsInWorkspace({
     user,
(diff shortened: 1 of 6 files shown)


fabiencelier (approved):
All clean ✅

front/lib/resources/group_resource.ts
```diff
@@ -1156,25 +1156,36 @@ export class GroupResource extends BaseResource<GroupModel> {
...
+   * - `status` is mutated in place by suspend/restore, so a membership
+   *   suspended today is filtered out even for a past `at`.
+   */
   private static async listUserGroupModelIdsInWorkspace({
     user,
     workspace,
     groupKinds = GROUP_KINDS.filter((k) => k !== "system"),
     transaction,
     dangerouslySkipMembershipCheck = false,
+    at = new Date(),
```
aubin-tchoi, line 1173: What's `at`?

aubin-tchoi: Ah it's the start date

aubin-tchoi (approved):

PR #30133, opened 2026-08-06
Admin governance: add group seeding to dust-hive seed

fabiencelier:
## Description

Seed to add 4 groups: 2 provisioned and 2 regular manual.
One has intentionally a very long name to see if it may break things

## Tests

tested locally + unit tested

## Risk

low, only a script

## Deploy Plan

None


matteotrab (commented):
LGTM

PR #30139, opened 2026-08-06
feat: improve analytics scaffolding

sfriquet:
## Description

Follow up to comments that should have been committed with #30079:
- use `auth` rather than `workspaceId`
- add test for cycle resolution when a workspace has a billing cycle
- switch + assert instead of ifs

## Tests

uts

## Risk

Low. Safe to rollback.

## Deploy Plan

Front only.


sfriquet:

- #30139 ◀

flvndvd (approved):

PR #30144, opened 2026-08-06
[dsbx] perf: serve repeat function invocations from the warm runner

fontanierh:
## Description

Stacked on #30160 (the runner's warm serve mode). This PR is the dsbx client and the version bumps that ship it; #30146 adds the metric on top.

`dsbx function run` forwards the invocation to the warm server when one is listening for this bundle, so a repeat invocation costs one local socket round trip instead of a runner spawn. A cold run leaves a detached server behind for the next invocation, so the first call pays nothing extra. No front or wire change: the stdout envelope's timings gain an additive `runnerKind` (warm|cold) field, and front treats `timingsMs` as opaque.

The client mirrors the never-re-run contract from #30160: it falls back to the cold path only on failures that precede the server's ack (no server, busy, stale bundle, protocol mismatch), and a lost outcome after the ack is reported as a failed invocation, never retried.

Security: the warm directory is `$HOME/.dust-fn` (under `/home/agent-proxied` in the sandbox, owned by the invoking uid) and is refused unless it is a real euid-owned 0700 directory, because the other untrusted sandbox uid must not be able to plant a socket that would receive the request environment. The function name is validated before it feeds the socket path, and the functions dir is part of the socket key so a slug from another dir cannot be served from the wrong bundle. Root dsbx never uses the warm path.

The final commit pins dsbx 0.1.42 and base image 0.8.67, which puts the warm-server dsbx into sandboxes as they are created or recreated after the image rollout.

## Tests

Cargo: 275 passing, including an end-to-end warm cycle against real bun (miss, spawn, warm hit with the exact outcome, republish, stale, miss), refusal of a squatted warm dir, and socket keying by slug and functions dir. Image registry tests updated to the pinned 0.8.67 / dsbx 0.1.42. Clippy clean, front typechecks clean. The server half is tested in #30160.

## Risk

The image bump is the consequential part: it triggers the base image rollout, and existing sandboxes pick the new image up on their next recreation. The dsbx v0.1.42 release must exist before the image builds, which the deploy plan below handles by dispatching the release workflow against this branch before merge.

The one intentional new failure mode: if the server dies between ack and outcome delivery, the invocation fails rather than re-running cold. For non-idempotent functions that is the correct trade, and it replaces the silent duplicate execution a naive fallback would produce. Concurrent invocations of the same function serialize onto one warm server only in the sense that the loser goes cold immediately; nothing waits.

Worth a manual check on a dev pod before merge: invoke a fast function twice and confirm the second run reports `runnerKind` warm with a lower duration.

## Deploy Plan

Both workflows are `workflow_dispatch` and read the version from the ref they run against, so the release and the image can be built from this branch before merge, leaving no window where main points at artifacts that do not exist:

1. `gh workflow run release-dsbx-cli.yml --ref fn-warm-runner` to publish the dsbx-v0.1.42 GitHub release the image build downloads.
2. `gh workflow run sandbox-img-registry.yml --ref fn-warm-runner` to build base image 0.8.67 in both regions.
3. Merge, then deploy front.
4. Open /poke/kill and request a kill of base image versions older than 0.8.67, so live conversations get sandboxes running the warm-capable dsbx instead of waiting for natural recreation. This is what the `sandbox-image-ack` label acknowledges.

If the merge lands first instead, run the two dispatches promptly: until the release exists, a front deploy fails its ensure-sandbox-images step looking for it.

The kill switch for warm serving is reverting the dsbx behavior in a patch release; no front flag is involved.

### DIFF
--- a/cli/dust-sandbox/src/commands/function/run.rs
+++ b/cli/dust-sandbox/src/commands/function/run.rs
@@ -1,9 +1,11 @@
 use std::time::Instant;
 
 use anyhow::{anyhow, Result};
+use tokio::io::AsyncReadExt as _;
 
-use super::envelope::{ResultDelivery, ResultEnvelope, TimingsMs};
-use super::{emit_error, spawn_function};
+use super::envelope::{ResultDelivery, ResultEnvelope, RunnerKind, TimingsMs};
+use super::warm::{self, WarmRun};
+use super::{emit_error, resolve_existing, spawn_function_at};
 use crate::api::DustApiClient;
 
 const SANDBOX_TOKEN_ENV: &str = "DUST_SANDBOX_TOKEN";
@@ -13,6 +15,11 @@ const NON_JSON_SNIPPET_MAX_CHARS: usize = 512;
 ///
 /// The request envelope is read from stdin and the function runs unprivileged
 /// (agent uid) when dsbx is invoked as root.
+///
+/// The invocation is served warm when a resident server for this function is
+/// listening (see `warm.rs`): one unix-socket round trip instead of a runner
+/// spawn. A cold run additionally leaves a warm server behind for the next
+/// invocation. Both paths produce the same runner `Output` JSON.
 ///
 /// Delivery modes:
 /// - `callback` (default): POST the runner response to the Dust result API when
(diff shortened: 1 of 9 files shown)


cli/dust-sandbox/src/commands/function/run.rs
```diff
@@ -40,11 +97,49 @@ pub async fn cmd_function_run(name: &str, result_delivery: ResultDelivery) -> Re
...
+        } else if bad_input {
+            2
+        } else {
+            1
+        });
+    }
+
+    let client = DustApiClient::from_env()?;
+    client
+        .post_function_result(name, &outcome)
```
davidebbo, line 137: Not directly related to this change, but it feels that in the inline cases, the function results could be directly returned rather than via this call back into Dust. Don't know how much that would save.

davidebbo (commented):
LGTM

PR #30145, opened 2026-08-07
[front] perf: trim the fast Pod function invocation setup path

fontanierh:
## Description

A fast Pod function invocation spends about 180ms of its ~545ms in front, spread over nine Postgres round trips, six Redis round trips, two GCS uploads and a Redis stream read-back, nearly all sequential. This trims the four costs that are pure waste:

The initial GCS blob write and the created-event publish sat between row creation and execution, though nothing before the terminal transition reads either. On the inline path both now run concurrently with execution; succeed()/fail() await them before the terminal blob write and the result publish, which keeps the blob write ordered (the terminal write rewrites the same object) and the created event ahead of the result event. The durable path is unchanged: the workflow reads the blob, so it still waits.

execute() re-fetched the function, authorized, and readied the sandbox serially; the function-check chain and the sandbox-readiness chain are independent and now overlap.

execute() wrote the sandbox's lastActivityAt immediately after ensureActive had written it under the lifecycle lock. The duplicate is gone and the write is throttled to 30s granularity: the reaper compares it against minutes-scale inactivity thresholds, and every invocation was hammering the shared sandbox row.

Both invocation callers subscribed to the Redis event stream to read back an outcome the inline execution had just produced in this process. The terminal transitions now record the outcome on the instance and callers short-circuit on it, falling back to the stream only for invocations that escalated to the workflow.

Expected effect on a fast invocation: one GCS round trip (~40-80ms) plus the subscribe round trips off the critical path, plus the smaller DB trims. Independent of the warm-runner work in #30144; the two compose.

## Tests

`lib/resources/sandbox`, `lib/api/sandbox_functions`, `lib/api/sandbox`: 483 passing. The inline execution test now also asserts the settled outcome is served from memory, that the deferred initial write still leaves a complete blob (input and result) for later fetches, and that a rehydrated instance never short-circuits. The execute test pins that lastActivityAt is written by the lifecycle, not a second time by execute. front and front-api typecheck clean.

## Risk

The ordering invariants the deferral must preserve are: terminal blob write after initial blob write (same GCS object), and result event after created event (same stream). Both are enforced at one place, the settleInitialPersistence gate inside the terminal transitions, and tested. If the deferred initial write fails and the invocation never settles, the row exists without a blob, which is the same state markCreatedAsErrored already produces today. Rollback is a plain revert; no schema or wire change.

## Deploy Plan

Normal front deploy, no flag. Invocation p50 (updatedAt minus createdAt on sandbox_function_invocations) is the number to watch.

### DIFF
--- a/front-api/routes/w/[wId]/sandbox-functions/[functionId]/invocations.ts
+++ b/front-api/routes/w/[wId]/sandbox-functions/[functionId]/invocations.ts
@@ -155,10 +155,13 @@ app.post(
     }
 
     const invocation = invocationResult.value;
+    // An inline execution settles the instance it ran on, so the outcome is usually already in
+    // memory; the stream read-back is only for invocations that escalated to the workflow.
     const outcome = fastExecution
-      ? await awaitSandboxFunctionInvocationOutcome({
+      ? (invocation.settledOutcome() ??
+        (await awaitSandboxFunctionInvocationOutcome({
           invocationId: invocation.sId,
-        })
+        })))
       : null;
 
     return ctx.json(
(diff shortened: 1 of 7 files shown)


davidebbo (approved):
LGTM

PR #30146, opened 2026-08-07
[front] feat: measure warm vs cold function runs

fontanierh:
## Description

Stacked on #30144, which ships the warm runner and its version pins. This PR is the measurement.

Front reads the additive `runnerKind` field from the stdout result envelope's timings and emits `sandbox.functions.run` (count plus duration distribution) tagged `runner_kind`: warm, cold, or unknown for envelopes from older dsbx. Parsing is deliberately lenient and never affects result handling; the timings block is diagnostics from whatever dsbx version answered. The stdout envelope is now parsed once and its outcome and timings read from that single parse, replacing four copies of the same extraction.

## Tests

Result delivery and envelope suites cover the warm, cold, and missing-timings cases and the single-parse path. front typechecks clean. The warm/cold behavior itself is tested in #30144.

## Risk

Observational only. An envelope without timings tags the run unknown, and no parsing failure can change how a result is handled.

## Deploy Plan

Nothing on its own. Merge after #30144 and its image rollout, then watch `sandbox.functions.run` by `runner_kind`: the warm share should climb as pods see repeat invocations, and warm durations should sit well under cold ones.

### DIFF
--- a/front/lib/api/sandbox/instrumentation.ts
+++ b/front/lib/api/sandbox/instrumentation.ts
@@ -99,6 +99,27 @@ export function recordLifecycleOperation(
   ]);
 }
 
+/**
+ * One fast Pod function run, tagged by which runner served it (resident warm
+ * server vs cold spawn). The warm share is the number the warm-runner rollout
+ * turns on; duration is the exec's wall time as front saw it.
+ */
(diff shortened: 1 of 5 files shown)


davidebbo (approved):
LGTM

PR #30154, opened 2026-08-07
Fix and simplify logic for getConditionalJITServers

Fraggle:
## Description

Extracts conditional JIT MCP server eligibility into a `CONDITIONAL_MCP_SERVERS` record in `jit_actions.ts` — one source of truth replacing scattered `if` checks duplicated between `getJITServers` and `getConditionalJITServers`. `getConditionalJITServers` now simply checks `autoInternalViews.has(...)`.

Fixes a logic bug along the way: when Computer (sandbox) is available, the old code excluded *all* queryable attachments from `query_tables_v2`, but queryable content nodes (vs file attachments) must still route to `query_tables_v2` regardless. Fixed with `isFileAttachmentType` discrimination in both the eligibility predicate and inside `getQueryTablesServer`. `getConversationFilesServer` is also simplified — the `attachments` guard is dropped since the caller now ensures the view was prefetched before calling.

## Tests

Updated existing test descriptions to be more precise. Added a new unit test verifying that a queryable content-node attachment produces a `query_tables_v2` server even when Computer is available.

## Risk

Low. Pure logic refactor with a targeted bug fix. No DB changes, no new APIs. The content-node fix restores the intended behavior; the refactor is covered by existing + new tests. Safe to rollback.

## Deploy Plan

Deploy front.

### DIFF
--- a/front/lib/api/assistant/jit/conversation.ts
+++ b/front/lib/api/assistant/jit/conversation.ts
@@ -5,7 +5,6 @@ import type { Authenticator } from "@app/lib/auth";
 import { ConversationResource } from "@app/lib/resources/conversation_resource";
 import { MCPServerViewResource } from "@app/lib/resources/mcp_server_view_resource";
 import { generateRandomModelSId } from "@app/lib/resources/string_ids_server";
-import type { ConversationAttachmentType } from "@app/types/api/assistant/conversation/attachments";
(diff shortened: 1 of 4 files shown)


Fraggle:

- #30128
    - #30154 ◀

flvndvd (approved):

PR #30159, opened 2026-08-07
[front] fix: unblock pod hard delete on sandbox FKs

fontanierh:
## Description

Pod scrub for workspace `0ec9852c2f` / pod `vlt_nWqafL1MP70ik` has been retrying for hours on
`ForeignKeyConstraintError: update or delete on table "vaults" violates foreign key constraint "sandbox_functions_spaceId_fkey"`.

The pod is soft-deleted before `scrubSpaceActivity` runs. `SandboxFunctionResource.baseFetch`
re-resolves each function's space via `SpaceResource.fetchByModelIds` without `includeDeleted`, so
the space came back empty and every function was silently dropped from the list
`deleteAllForSpace` iterates over. It returned `Ok(0)` and the hard delete then hit the RESTRICT FK.

- Thread `includeDeletedSpace` through `listBySpace` / `baseFetch`, set it in `deleteAllForSpace`.
- Count leftover rows after the loop so a partial list fails there with a clear message instead of
  surfacing as an opaque FK violation later.
- `sandbox_owners.spaceId` and `sandbox_env_vars.spaceId` are RESTRICT too and were never cleaned by
  the pod scrub, so they would be the next walls: destroy the pod sandbox and delete its env vars in
  `scrubSpaceActivity`.

## Tests

Added a regression test that soft-deletes the pod before calling `deleteAllForSpace` (fails on main,
passes here), and one covering that pod env var cleanup leaves workspace-scoped rows alone.

## Risk

Low. `includeDeletedSpace` defaults to false, so read paths are unchanged. New behavior only runs in
the scrub path.

## Deploy Plan

Standard deploy. The stuck workflow should recover on its next retry once the worker ships.

### DIFF
--- a/front/lib/resources/sandbox_env_var_resource.test.ts
+++ b/front/lib/resources/sandbox_env_var_resource.test.ts
@@ -631,6 +631,34 @@ describe("SandboxEnvVarResource pod scope", () => {
     expect(fetched?.allowedDomains).toEqual(["api.example.com"]);
   });
 
+  it("deletes only the pod-scoped rows on pod scrub", async () => {
+    const { authenticator, pod } = await setupPod();
+
+    const podResult = await SandboxEnvVarResource.makeNew(
+      authenticator,
+      podScope(pod),
+      { name: "CONFIG_TOKEN", value: "pod-config-value" }
+    );
+    expect(podResult.isOk()).toBe(true);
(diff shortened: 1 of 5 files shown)


adrsimon (approved):
LGTM

PR #30160, opened 2026-08-07
[dsbx] perf: harden the warm server: staleness, secrets, backpressure, lifetime

fontanierh:
## Description

Stacked on #30167 (the warm server core). This PR is the hardening pass; the mechanism and protocol are reviewed there.

Staleness: a republished bundle must never be served from the old import. The server stats the bundle (mtime/size) before every request, and additionally refuses any request whose `bundleSha256` (an additive envelope field, stamped by front in #30165) does not match the hash of the bytes it imported — deterministic even when gcsfuse metadata caching hides the rewrite from the stat. Both refusals are pre-ack: the client re-runs cold against the fresh bundle and respawns a matching server. A hard 600s lifetime bounds staleness for unstamped envelopes; the lifetime cap drains the in-flight invocation instead of exiting under it.

Secrets: the spawn-time invocation's token and user identity are scrubbed at server startup, so they never sit in an idle server's environment; every request carries its own.

Delivery: replies are flushed through the drain callback, so a large outcome that exceeds the kernel buffer cannot reach the client as a truncated JSON line. An invocation that outlives a 120s deadline exits the server without replying — its client lost to front's much shorter exec timeout long ago.

## Tests

bun test: rewritten bundle refused as stale with server exit, matching bundle hash served, mismatched hash refused as stale even with an unchanged stat, unstamped envelope served (older front), spawn-time secrets scrubbed while unrelated spawn env survives. Biome clean.

## Risk

None at runtime until #30144 ships the client. The refusal direction is safe by construction: a false stale costs one cold run, never a wrong result.

## Deploy Plan

Nothing on its own; ships with #30144's version bumps.

### DIFF
--- a/cli/dust-sandbox/functions-runner/protocol.ts
+++ b/cli/dust-sandbox/functions-runner/protocol.ts
@@ -9,6 +9,10 @@ export interface RequestInput {
   headers: Record<string, string>;
   body?: string;
   encoding: Encoding;
+  // Sha256 hex of the bundle the publisher expects to run. The warm server
+  // refuses to serve an import that does not match (see serve.ts); the cold
+  // runner ignores it, since a cold run reads the bundle straight from disk.
+  bundleSha256?: string;
 }
 
 export const RUNNER_ERROR_CODES = [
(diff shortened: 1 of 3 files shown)


davidebbo (approved):
LGTM

PR #30164, opened 2026-08-07
fix: pause connectors when the workspace plan blocks API access

adrsimon:
## Description

Saw connector activities retrying forever on `workspace_can_use_product_required_error` (the workspace plan doesn't allow API access). Nothing about that gets better with a retry, so it now goes down the same path as the other errors requiring a human action in `ActivityInboundLogInterceptor`: `syncFailed` + `pauseAndStop`.

The three-way `instanceof` chain plus its nested `if/else if/else` became a `categorizeFinalConnectorError()` helper returning the error type, log message and pause reason.

## Tests

Added a case in `temporal_monitoring.test.ts` with the actual payload, asserting `syncFailed` and `pauseAndStop`.

## Risk

Low. Only widens an existing branch, no change for the other errors.

## Deploy Plan

Deploy `connectors` after `front`.

### DIFF
--- a/connectors/src/lib/temporal_monitoring.ts
+++ b/connectors/src/lib/temporal_monitoring.ts
@@ -3,6 +3,7 @@ import type logger from "@connectors/logger/logger";
 import type { Logger } from "@connectors/logger/logger";
 import { statsDClient } from "@connectors/logger/withlogging";
 import { ConnectorResource } from "@connectors/resources/connector_resource";
+import type { ConnectorErrorType } from "@connectors/types";
 import { WithRetriesError } from "@connectors/types";
 import type { ConnectorProvider } from "@dust-tt/client";
 import type { Context } from "@temporalio/activity";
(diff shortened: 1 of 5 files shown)


connectors/src/lib/temporal_monitoring.ts
```diff
@@ -27,6 +28,60 @@ import { getConnectorId } from "./temporal";
 const TRACK_SUCCESSFUL_ACTIVITIES_FOR_CONNECTOR_IDS = [145];
 const TRANSIENT_ERROR_PRE_BACKOFF_RETRY_ATTEMPTS = 19;
 
+// Dust API errors reach the interceptor as opaque errors whose message embeds
+// the serialized API response, so we match on the error type in the message.
+function isWorkspacePlanNoApiAccessError(err: unknown): err is Error {
+  return (
+    err instanceof Error &&
+    err.message.includes("workspace_can_use_product_required_error")
```
aubin-tchoi, line 36: we don't have a better-typed error for this one?

aubin-tchoi (approved):
LGTM!

connectors/src/lib/temporal_monitoring.ts
adrsimon: nope, we get this from the API and it's a plain error we can't really match any other way sadly

PR #30165, opened 2026-08-07
[front] feat: stamp invocations with the published bundle hash

fontanierh:
## Description

Last slice of the warm runner stack (#30160, #30144, #30146, this). Makes republish-then-invoke deterministic: an agent that republishes a function can never be served the previous bundle from a warm server.

Publish stores the sha256 of the bundle it uploaded (new `bundleSha256` column, restamped by every republish alongside the upload), and every invocation envelope carries it, read from the persisted row like the execution mode. The warm server (#30160) hashes the bundle it imported and refuses any request whose expected hash differs, pre-ack, so the client re-runs cold against the fresh bundle and respawns a matching server. This does not depend on gcsfuse metadata behavior; the server's stat check and 600s lifetime remain as backstops for envelopes without a hash.

The hash lands with the row update after the upload, so a warm server is never told to expect a bundle that is not on disk yet. Functions last published before the column existed have a null hash and keep today's backstop-only behavior until their next republish. Deployed dsbx 0.1.41 runners ignore the extra envelope field (verified against its parser), so this is safe to merge independently of the image rollout.

## Tests

Publish suite asserts the hash is stored on first publish and restamped on republish; invocation suite asserts the exec envelope carries the function's hash. The server-side refusal is tested in #30160 (matching hash served, mismatched hash refused as stale with server exit, unstamped envelope served). front typechecks clean.

## Risk

Envelope field is additive and ignored by every deployed runner. Worst case of a bad hash would be warm refusals falling back to cold runs, never a wrong result. Pre-deploy migration adds a nullable column.

## Deploy Plan

Pre-deploy migration, then normal deploy. No ordering constraint with the dsbx release: stamped envelopes are inert until sandboxes run the 0.1.42 image from #30144.

### DIFF
--- a/front/lib/resources/storage/models/sandbox_function.ts
+++ b/front/lib/resources/storage/models/sandbox_function.ts
@@ -54,6 +54,10 @@ export class SandboxFunctionModel extends WorkspaceAwareModel<SandboxFunctionMod
   declare description: string;
   declare userIdentity: SandboxFunctionUserIdentityPolicy | null;
   declare executionMode: CreationOptional<SandboxFunctionExecutionMode>;
+  // Sha256 hex of the published bundle. Stamped onto every invocation envelope so the in-sandbox
+  // warm server can refuse to serve a bundle the publisher has since replaced. Null only for
+  // functions last published before the column existed.
+  declare bundleSha256: string | null;
   declare inputSchema: JSONSchema;
   declare outputSchema: JSONSchema;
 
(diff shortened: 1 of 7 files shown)


front/lib/resources/storage/models/sandbox_function.ts
```diff
@@ -122,6 +126,10 @@ SandboxFunctionModel.init(
         isIn: [SANDBOX_FUNCTION_EXECUTION_MODES],
       },
     },
+    bundleSha256: {
+      type: DataTypes.STRING(64),
```
flvndvd, line 130: Are we sure that sha256 will always fit in a 64 char length?

fontanierh: It's always going to be precisely 64 chars

flvndvd (approved):

front/lib/resources/storage/models/sandbox_function.ts
flvndvd: I indeed missed the `.digest('hex')` 🙈.

PR #30166, opened 2026-08-07
[front] feat: wire Agents tab of the usage filter panel to real, period-scoped data

avervaet:
## Description

This PR wires the Agents tab of the usage filter panel to real workspace data, replacing the previous static mock list with a live fetch via `useAgentConfigurations`.

Until now, the Agents tab was populated from a hardcoded list in `usageFilterMockData.ts` (`SupportBot`, `SalesAssistant`, etc.) that was period-unaware and bore no relation to actual workspace agents. Agent selection was also a no-op in the consumption scope filter sent to the analytics backend.

This PR makes three connected changes:

1. **Live agent list.** `UsageFilterPanel` now calls `useAgentConfigurations({ agentsGetView: "all" })` lazily (only when the Agents tab is open) and maps the results to `UsageFilterAgentOption[]`, including the agent's `pictureUrl` as the new `image` field. The `agent` key is removed from the `categoryOptions` prop, mirroring the pattern already used for members and groups.

2. **Correct scope enum.** The local `USAGE_FILTER_SCOPES = ["company", "shared", "private"]` constant is replaced with `AGENT_CONFIGURATION_SCOPES` from `@app/types/assistant/agent`. The scope labels are updated accordingly (`global` → "Company", `visible` → "Shared", `hidden` → "Private"). `UsageFilterScope` becomes an alias for `AgentConfigurationScope` directly.

3. **Agent filter forwarded to backend.** `toConsumptionScopeFilter` is extended to include selected agent `sId`s under the `agents` key in the `ConsumptionScopeFilter`, alongside the existing `users` key. Previously only members were forwarded.

As a visual side-effect, agent entries in the filter list now render an `Avatar` (name-initials or picture) consistent with member entries, replacing the previous `null` icon.

Models, tools, skills, and sources remain on mock data. The mock agent names and `buildAgentOptions` helper are removed.

## Risk

Low. The change is purely additive on the consumption filter side (an extra optional field). The agent list is fetched from the same `useAgentConfigurations` endpoint already used across the app. Reverting to mock data would require re-adding the `agent` key to `categoryOptions` callers and restoring the mock entries, but no data is mutated. The scope label remapping (`company/shared/private` → `global/visible/hidden`) is a display-layer alignment with the canonical enum and has no backend effect.

## Deploy Plan

- Deploy front

### DIFF
--- a/front/components/workspace/analytics/UsageFilterPanel.tsx
+++ b/front/components/workspace/analytics/UsageFilterPanel.tsx
@@ -16,7 +16,6 @@ import type {
 import {
   USAGE_FILTER_CATEGORIES,
   USAGE_FILTER_CATEGORY_LABEL,
-  USAGE_FILTER_SCOPES,
   USAGE_MODEL_TIERS,
 } from "@app/components/workspace/analytics/usageFilter";
 import { UsageFilterAgentScopeControls } from "@app/components/workspace/analytics/usageFilterPanel/UsageFilterAgentScopeControls";
@@ -31,6 +30,7 @@ import { useToggleSelectionList } from "@app/hooks/useToggleSelectionList";
 import { useAgentConfigurations } from "@app/lib/swr/assistants";
 import { useGroups } from "@app/lib/swr/groups";
 import { useSearchMembers } from "@app/lib/swr/memberships";
+import { AGENT_CONFIGURATION_SCOPES } from "@app/types/assistant/agent";
 import { MANAGEABLE_GROUP_KINDS } from "@app/types/groups";
 import { assertNever } from "@app/types/shared/utils/assert_never";
 import type { LightWorkspaceType } from "@app/types/user";
@@ -93,7 +93,7 @@ export function UsageFilterPanel({
   const [activeCategory, setActiveCategory] =
     useState<UsageFilterCategory>("agent");
   const [activeScope, setActiveScope] = useState<UsageFilterScope>(
-    USAGE_FILTER_SCOPES[0]
+    AGENT_CONFIGURATION_SCOPES[0]
   );
   const [activeTier, setActiveTier] = useState<UsageModelTier>(
     USAGE_MODEL_TIERS[0]
--- a/front/components/workspace/analytics/UsageFilterPanel.tsx
+++ b/front/components/workspace/analytics/UsageFilterPanel.tsx
@@ -6,7 +6,6 @@ import type {
   UsageFilterMemberOption,
   UsageFilterModelOption,
   UsageFilterOptionForCategory,
-  UsageFilterScope,
   UsageFilterSkillOption,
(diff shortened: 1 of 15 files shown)


avervaet:

- #30241
    - #30212
        - #30254
            - #30168
                - #30166 ◀
                    - #30170
                        - #30173
                            - #30174
                                - #30177

front/components/workspace/analytics/usageFilter.ts
```diff
@@ -162,11 +165,22 @@ export function removeUsageFilterGroup(
...
 }
 
-// Only "user" is wired to a real consumption scope dimension so far; the
-// other categories stay mock data and are not sent as query filters.
+// Only "user" and "agent" are wired to real consumption scope dimensions so
+// far; the other categories stay mock data and are not sent as query filters.
 export function toConsumptionScopeFilter(
   filter: UsageFilter
 ): ConsumptionScopeFilter {
+  const scopeFilter: ConsumptionScopeFilter = {};
```
avervaet, line 154: This code is uniformized in the last wiring pr when every categories are properly linked to real data

front/components/workspace/analytics/UsageFilterPanel.tsx
```diff
@@ -174,6 +177,12 @@ export function UsageFilterPanel({
     disabled: !isMemberCategoryActive,
   });
 
+  const { agentConfigurations } = useAgentConfigurations({
+    workspaceId: owner.sId,
+    agentsGetView: "all",
```
avervaet, line 184: With the current implementation admins won't be able to see users private agents as only super users can. Is this intended or should we allow admins to see users private agents ? @flvndvd

front/components/workspace/analytics/usageFilter.ts
```diff
@@ -23,14 +24,12 @@ export const USAGE_FILTER_CATEGORY_LABEL: Record<UsageFilterCategory, string> =
     source: "Sources",
   };
 
-export const USAGE_FILTER_SCOPES = ["company", "shared", "private"] as const;
-
-export type UsageFilterScope = (typeof USAGE_FILTER_SCOPES)[number];
+export type UsageFilterScope = AgentConfigurationScope;
```
flvndvd: Can we avoid the passe-plat type.

front/components/workspace/analytics/UsageFilterPanel.tsx
```diff
@@ -174,6 +177,12 @@ export function UsageFilterPanel({
     disabled: !isMemberCategoryActive,
   });
 
+  const { agentConfigurations } = useAgentConfigurations({
```
flvndvd, line 182: Ok to move forward, but let's make sure that we create a dedicated endpoint to get all facets in one call.

flvndvd (approved):

PR #30167, opened 2026-08-07
[dsbx] perf: runner serve mode keeping one bundle warm behind a unix socket

fontanierh:
## Description

First slice of the warm function runner stack (this, #30160 hardening, #30144 dsbx client and version bumps, #30146 metric, #30165 bundle-hash stamping).

A cold `dsbx function run` pays bundle resolution against the gcsfuse mount, a runner temp-file write, a bun process spawn, and the import of the bundle and its dependencies, on every invocation. Measured locally, the import alone is ~150ms cold versus ~10ms once cached.

The runner gains a `serve` subcommand: a bun process that keeps one imported bundle resident behind a unix socket so a repeat invocation costs one local socket round trip. This PR is the core mechanism and protocol; nothing invokes `serve` until the dsbx client lands in #30144.

Protocol decisions reviewed here: the server acks before executing, so a client can distinguish pre-ack failures (safe to fall back cold) from a lost outcome after the ack (never retried — pod functions are side-effectful and must not run twice). Requests are never queued: one invocation at a time, and a request arriving while one runs gets an immediate busy reply so the client runs cold instead of waiting behind work of unknown duration. Per-request environment (the sandbox token and user identity travel in env) is applied for the invocation and cleared afterwards. The server exits on idle (120s), on any malformed request, and it resolves the spawn race on an existing socket by probing for a live listener.

The hardening — bundle staleness (stat plus the envelope's bundle hash), spawn-time secret scrubbing, backpressure-safe replies, the lifetime cap, and the invocation deadline — lands in #30160 on top of this.

## Tests

bun test: served invocation, repeat invocations from one process, env applied and cleared between requests (including the sandbox token), busy instead of queueing with recovery, ack-before-outcome framing, protocol version refusal, structured runner errors, malformed envelopes survived. Biome clean.

## Risk

None at runtime: the subcommand is unreachable until #30144 ships the client, and `runner.js` is only rebuilt by the dsbx release workflow.

## Deploy Plan

Nothing on its own; ships with #30144's version bumps.

### DIFF
--- a/cli/dust-sandbox/functions-runner/serve.ts
+++ b/cli/dust-sandbox/functions-runner/serve.ts
@@ -1,0 +1,278 @@
+// Warm function server: keeps one imported function bundle resident in a bun
+// process and serves invocations over a unix socket, so repeat invocations skip
+// process spawn, bundle resolution, and the import of the bundle and its
+// dependencies (the dominant sandbox-side costs of a cold run).
+//
+// One server serves exactly one handler path, one invocation at a time. The
+// per-invocation environment (sandbox token, user identity) travels in the
+// request and is applied to process.env for the duration of the invocation,
+// then cleared — concurrency would race those swaps, so a request arriving
+// while one is running gets an immediate `busy` reply and the client runs
+// cold instead of queueing behind work of unknown duration.
+//
+// Duplicate executions are the failure mode this protocol is shaped around:
+// pod functions are arbitrary side-effectful code, and front assumes a failed
+// start means nothing ran. The server acks before executing; the client only
+// falls back to the cold path on failures that precede the ack. After the
+// ack, a lost outcome is reported as a failed invocation, never re-run.
+//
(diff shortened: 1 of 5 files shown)


cli/dust-sandbox/functions-runner/serve.ts
```diff
@@ -0,0 +1,280 @@
...
+// How long the server waits for another invocation before exiting.
+const IDLE_TIMEOUT_MS = 120_000;
+
+interface WarmRequest {
+  v: number;
+  env: Record<string, string>;
+  input: string;
+}
+
+export function parseWarmRequest(line: string): WarmRequest | null {
```
davidebbo, line 56: Can't we use zod to parse this? Seems like a lot of grunt logic to do it manually,

fontanierh: Switched to zod

davidebbo (approved):
LGTM

PR #30168, opened 2026-08-07
[front] feat: wire Members/Groups tabs of the usage filter panel to real, period-scoped data

avervaet:
## Description

This PR wires the **Members** and **Groups** tabs of the usage page's filter panel to real, period-scoped data, and connects the resulting `UsageFilter` state into the underlying consumption queries.

**Until now:**
- The Members tab was populated from the full workspace member roster, with no awareness of the selected time period. Every workspace member appeared regardless of whether they had any activity in the window.
- The Groups tab was entirely mocked.
- The `UsageFilter` selection for the "member" dimension had no effect on the overview, chart, or attribution table queries.

**This PR:**
- **Members** now come from the same period-ranked consumption data that drives the Attribution table. Only members with activity in the selected period window are surfaced, ranked by consumption.
- **Groups** are no longer mocked. A new `relevant-groups` endpoint resolves, for that same set of period-active users, which groups they belonged to during the period window. This keeps the group list contextually relevant rather than listing all workspace groups unconditionally.
- The `UsageFilter` state (member dimension) is now forwarded to the overview, chart, and attribution table consumption queries, so filtering by member actually narrows the displayed data.

A new route was added under `front-api/routes/` — OpenAPI docs should be regenerated (`npm -w front-api run docs`) and published via the `Deploy OpenAPI Docs` GitHub action.

## Tests

Testing evidence was not provided for this draft.

## Risk

Low. The changes are additive on the data-fetching side: the existing member roster fallback is replaced by a scoped query, and the new `relevant-groups` endpoint is a net-new read-only route. The `UsageFilter` wiring only affects query parameters; no data model changes or migrations are involved. Safe to rollback by reverting the front deployment.

## Deploy Plan

- Deploy `front`.


avervaet:

- #30241
    - #30212
        - #30254
            - #30168 ◀
                - #30166
                    - #30170
                        - #30173
                            - #30174
                                - #30177

front/components/workspace/analytics/usageFilter.ts
avervaet, line 1: This will collapse cleanly once every data source are wired check https://github.com/dust-tt/dust/pull/30174

front/components/workspace/analytics/usageFilterPanel/UsageFilterOptionCheckboxList.tsx
```diff
@@ -18,6 +20,10 @@ interface UsageFilterOptionCheckboxListProps {
   selectedIds: Set<string>;
   onToggleOption: (option: UsageFilterOption) => void;
   onSelectAll: () => void;
+  // Only set for categories backed by a paginated server fetch (members).
```
avervaet, line 23: removed when all the filtering are properly wired

front/components/workspace/analytics/UsageFilterPanel.tsx
```diff
@@ -215,36 +353,52 @@ export function UsageFilterPanel({
...
+              options={displayedOptions}
               selectedIds={selectedIdsForActiveCategory}
               onToggleOption={(option) => toggleOption(activeCategory, option)}
               onSelectAll={() =>
                 selectAllFiltered(activeCategory, filteredOptions)
               }
+              hasMore={
+                activeCategory === "member"
+                  ? hasMoreMembers
+                  : hasMoreStaticOptions
```
aubin-tchoi: FYI this part will get ugly when you'll wire in the data for the other fields

front/lib/resources/group_resource.ts
```diff
@@ -2959,4 +2959,33 @@ export class GroupResource extends BaseResource<GroupModel> {
       memberCount: memberCounts.get(group.id) ?? 0,
     }));
   }
+
+  /**
+   * Batched counterpart of `toJSONWithMemberCount` that also carries each
+   * group's active member sIds, resolved in two queries total regardless of
+   * group count.
+   */
+  static async toJSONWithMembers(
```
aubin-tchoi: IMHO `toJSON` methods should always be sync, they only serialize data that is already available

aubin-tchoi: in your case you're really augmenting data already fetched (groups) rather than serializing data

front-api/routes/w/[wId]/groups/index.ts
```diff
@@ -47,7 +50,10 @@ app.get(
       : await GroupResource.listAllWorkspaceGroups(auth, { groupKinds });
 
     return ctx.json({
-      groups: await GroupResource.toJSONWithMemberCounts(auth, groups),
+      groups:
+        withMembers === "true"
+          ? await GroupResource.toJSONWithMembers(auth, groups)
+          : await GroupResource.toJSONWithMemberCounts(auth, groups),
```
Comment (aubin-tchoi) [inline front-api/routes/w/[wId]/groups/index.ts]: nit: probably some room for consolidation, as I suppose that if you have the members you also have the count

front/components/workspace/analytics/UsageFilterPanel.tsx
```diff
@@ -173,6 +301,16 @@ export function UsageFilterPanel({
     setIsOpen(false);
   };
 
+  const handleAddGroup = (group: UsageFilterGroup) => {
```
aubin-tchoi: nit: looks like you'll have more or less twice as many `handleXxx` methods as you have filter categories, would be nice to find a way to not have this explode

aubin-tchoi (approved):
LGTM but let's start thinking about simplifying

front-api/routes/w/[wId]/groups/index.ts
Comment (avervaet) [reply front-api/routes/w/[wId]/groups/index.ts]: Yes but when fetching the count you don't want the members to avoid a nested queries

PR #30169, opened 2026-08-07
Require run usage on consumption items

flvndvd:
## Description

`runUsageId` is currently nullable in the schema and TypeScript types, even though every consumption item is tied to a run usage. This forces downstream code to handle a state that does not exist.

We checked both production regions and confirmed that no existing rows have a null `runUsageId`.

This PR makes the column non-null through a post-deploy migration, tightens the related types, and updates the affected test fixture. This gives consumers a simpler and more accurate type without changing existing behavior.

## Deploy Plan

 - [ ] Deploy fron
 - [ ] Run post-deploy migration

### DIFF
--- a/front/migrations/post-deploy/20260807112434_require_run_usage_for_consumption_items.sql
+++ b/front/migrations/post-deploy/20260807112434_require_run_usage_for_consumption_items.sql
@@ -1,0 +1,27 @@
+/*
+Statement 0
+*/
+SET SESSION statement_timeout = 3000;
+SET SESSION lock_timeout = 3000;
+ALTER TABLE "public"."agent_message_consumption_items" ADD CONSTRAINT "pgschemadiff_tmpnn_E1Qf_yeWQc$aNSQs3s$I$w" CHECK("runUsageId" IS NOT NULL) NOT VALID;
+
+/*
+Statement 1
+*/
+SET SESSION statement_timeout = 1200000;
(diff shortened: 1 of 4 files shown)


front/migrations/post-deploy/20260807112434_require_run_usage_for_consumption_items.sql
```diff
@@ -0,0 +1,27 @@
+/*
+Statement 0
+*/
+SET SESSION statement_timeout = 3000;
+SET SESSION lock_timeout = 3000;
+ALTER TABLE "public"."agent_message_consumption_items" ADD CONSTRAINT "pgschemadiff_tmpnn_E1Qf_yeWQc$aNSQs3s$I$w" CHECK("runUsageId" IS NOT NULL) NOT VALID;
```
aubin-tchoi, line 6: do we name the constraints like this 😬

aubin-tchoi: nvm it's dropped afterwards

aubin-tchoi (approved):
LGTM!

front/migrations/post-deploy/20260807112434_require_run_usage_for_consumption_items.sql
flvndvd: yes the new tool use those types of names.

PR #29299, opened 2026-07-22
feat(front): rework model picker around tiers

Nils-Fedrigo:
## Description
See [live preview ](https://pr-29299-merge--app.preview.dust.tt/)
Screenshot 2026-07-22 at 22 10 02

Reworks the input-bar model picker around **model tiers** (Fast / Standard / Complex) instead of a flat, per-effort model list, and adds a dedicated reasoning-effort control.

> Stacked on #29302 (sparkle `SliderSteps` + Bar icons), which this PR consumes.

- Tiers are surfaced as the primary choice; concrete models remain reachable through a "More models" section grouped by maker, with inline expansion on width-constrained clients.
- Reasoning effort is now picked with a new `ReasoningEffortSlider`, backed by sparkle's `SliderSteps` and the `BarLow` / `BarHalf` / `BarFull` icons.
- Simplified selection/override logic in `modelPickerUtils` (`resolveShownSelection`, `buildTierSelection`, `getInitialEffort`, `getModelTier`, `isSameSelection`) and reduced `InputBarModelPicker` state to a single override plus derived display/send selections.
- Removed the now-unused `ModelPickerLineItem`, `ModelPickerList`, `ModelPickerProviderSection`, and `ModelPickerRowTooltip`; added `ModelPickerModelRow`, `ModelPickerMoreModels`, and `ModelPickerSelectionIndicator`.

## Tests

Manual: exercise the input-bar picker — switch tiers, open "More models" and pick a concrete model, adjust the reasoning-effort slider, revert to the agent default, and confirm behavior on width-constrained (mobile) layouts where makers expand inline.

PR #30170, opened 2026-08-07
[front] feat: wire Models tab of the usage filter panel to real, period-scoped data

avervaet:
## Description

Replaces mock data in the Models tab of the usage filter panel with real, workspace-sourced data. This is the third tab in the series, stacked on #30166 (Agents tab).

**Until now**, the Models tab was powered by a hardcoded mock list (`usageFilterMockData.ts`) with fictional model names, randomly cycling tier values, and hard-coded maker assignments. Crucially, selected model filters were never forwarded to the consumption queries — so selecting a model had no actual effect on the overview, chart, or Attribution table.

**This PR:**

- Switches the model source to `useModels`, the same workspace-wide model catalog backing the model picker elsewhere in the app, so every enabled model is listable and searchable regardless of the selected period. The `useModels` call is lazy — it only fires when the Models tab is active (`disabled: !isModelCategoryActive`).
- Derives each model's maker via `getModelMaker` and its tier (Fast / Standard / Complex) client-side via `getModelEffortTier` + the new `usageModelTierFromModelsTierName` helper, matching the server-side resolution `ModelsTierResource.getTierForModel` performs. Models with no tier mapping get `tier: undefined` and are excluded from the tier quick-filter buttons but remain reachable via "More models".
- Filters out auto/meta stream IDs (`isModelStreamId`) from the catalog list — Fast / Standard / Complex appear as quick-filter buttons, not as selectable catalog entries.
- Wires selected model IDs into `toConsumptionScopeFilter` → `scopeFilter.models`, so model selections now actually narrow consumption queries.
- Removes the `model` field from the `categoryOptions` prop (the last one still fed from mock data) and renames the `UsageFilterModelComplexityControls` prop from `models` to `moreModelsCatalog` for clarity.
- Refactors `usageModelTierFromModelsTierName`: separates the `null`/`undefined` guard from the switch and uses `assertNeverAndIgnore` on the default case, so an unknown backend tier is ignored rather than silently swallowed (addressing the dust-agent review suggestion).

Tools, skills, and sources remain on mock data.

## Tests

- `npx tsgo --noEmit` passes.
- `labels.test.ts` and `top.test.ts` pass; both extended with model maker/tier coverage.
- Manual browser validation pending: period top-100 list renders real models, tier quick-filter narrows correctly, "More models" browse groups by maker, and selecting a model visibly narrows the overview/chart/Attribution table.

## Risk

Low. The change is purely a data-source swap on a filter panel tab:

- Mock data is replaced; no existing query path is removed.
- Models with unmapped tiers degrade gracefully (`tier: undefined` → absent from quick-filter, still reachable via "More models").
- `assertNeverAndIgnore` on unknown backend tier values ensures forward compatibility if new tiers are added server-side.
- `useModels` is fetched lazily only when the Models tab is open — no additional network cost for users who never open that tab.
- The `scopeFilter.models` field is only populated when at least one model is selected, matching the existing pattern for agents.

## Deploy Plan

- Deploy front

### DIFF
--- a/front/components/workspace/analytics/usageFilter.ts
+++ b/front/components/workspace/analytics/usageFilter.ts
@@ -1,4 +1,5 @@
 import type { ConsumptionScopeFilter } from "@app/lib/api/analytics/consumption/scope";
+import type { ModelsTierName } from "@app/lib/api/assistant/token_pricing/tiers";
 import type { AgentConfigurationScope } from "@app/types/assistant/agent";
 import type { ModelMakerIdType } from "@app/types/assistant/models/types";
 import type { ConnectorProvider } from "@app/types/data_source";
@@ -71,7 +72,10 @@ export interface UsageFilterSourceOption extends UsageFilterOptionBase {
 export interface UsageFilterModelOption extends UsageFilterOptionBase {
   kind: "model";
   lab: ModelMakerIdType;
-  tier: UsageModelTier;
+  // Undefined for a model outside the static tier table — it doesn't match
+  // any Fast/Standard/Complex quick filter, so it's absent from the main
+  // checklist but still reachable through the "More models" browse dropdown.
+  tier: UsageModelTier | undefined;
 }
 
 export interface UsageFilterToolOption extends UsageFilterOptionBase {
@@ -146,8 +150,9 @@ export function selectAllUsageFilterOptions<C extends UsageFilterCategory>(
   return { ...filter, [category]: [...current, ...additions] };
 }
 
-// Members, teams, and agents are wired to real consumption scope dimensions.
-// The other categories stay mock data and are not sent as query filters yet.
+// Members, teams, agents, and models are wired to real consumption scope
+// dimensions. The other categories stay mock data and are not sent as query
+// filters yet.
 export function toConsumptionScopeFilter(
   filter: UsageFilter
 ): ConsumptionScopeFilter {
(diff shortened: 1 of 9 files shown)


avervaet:

- #30241
    - #30212
        - #30254
            - #30168
                - #30166
                    - #30170 ◀
                        - #30173
                            - #30174
                                - #30177

front/components/workspace/analytics/usageFilter.ts
```diff
@@ -168,5 +174,34 @@ export function toConsumptionScopeFilter(
...
+
   return scopeFilter;
 }
+
+// Maps the backend's reasoning-effort-aware pricing tier onto the filter
+// panel's simpler Fast/Standard/Complex bucket. Null propagates (a model
+// outside the static tier table, or a raw catalog entry with no config match)
+// as "no bucket", so it's excluded from every quick-filter tier rather than
+// landing in one arbitrarily.
+export function usageModelTierFromModelsTierName(
```
avervaet, line 190: This needs to be discussed @flvndvd @ClementAupiais.

avervaet: material: https://github.com/dust-tt/dust/pull/29299

aubin-tchoi, line 190: probably needs some alignment with what we're doing with the model picker, we should not introduce a new wording users are not familiar with here

aubin-tchoi (approved):

PR #30176, opened 2026-08-07
dsbx: pin Host to the approved domain on port 80

fontanierh:
## Description

Port 80 never re-checked the Host header. dsbx approved the domain from the first request, then forwarded every later request on the same keep-alive connection unchecked. Since the proxy only splices bytes after the handshake, a connection to an allowlisted CDN-fronted domain could front any other host behind that CDN, bidirectionally, with nothing in the deny log. 443 was already fine, it pinned Host to the SNI.

`normalized_authority` now pins each request's authority to the domain approved for the connection in both modes: SNI on 443, first request's Host on 80. Mismatches deny with a new `host_domain_mismatch` reason and drop the connection.

Two behavior changes:

- **A non-default port in Host on a port-80 connection now denies**, matching 443, and from the first request (the domain sent for approval has any port stripped, the per-request check only strips the default one). Replaces the test that locked in the old behavior. Kills the forward-proxy-on-:80 pattern, which nothing uses since egress is nftables-redirected into dsbx.
- **Harness deny reasons are now logged in front.** They are hidden from the agent on purpose, but were logged nowhere else, so a request-policy denial was invisible in prod. Bad place to be right after tightening one of those checks.

## Tests

New regression tests, each verified to fail with the pin reverted:

- Mismatched Host on the first request.
- Fronted second request on a keep-alive connection.
- Self-consistent proxy-form request (`GET http://evil.example/` with matching Host), which used to pass both the Host and absolute-URI checks.
- Non-default port in Host.
- End-to-end `run_plain_http_session` mirroring the PoC: request 1 reaches upstream, request 2 with a swapped Host does not, deny log records `host_domain_mismatch`.

Three others came along as invariant guards (h2c preface, case and trailing-dot Host, default port in Host). They pass either way.

## Risk

Low, strictly tightening. Only new denial is a non-default port in Host on port 80, which no real client sends. Rollback is a revert plus the version bumps.

## Deploy Plan

Needs a dsbx release and image rebuild (dsbx 0.1.43, dust-base 0.8.68). Running sandboxes stay on the old binary until they cycle, so after deploy open `/poke/kill` and kill the older dust-base versions (`sandbox-image-ack` label is on for this). Then watch for "Sandbox egress request policy denied requests".


aubin-tchoi (approved):

zmarouf (approved):
Ah. LGTM

PR #30178, opened 2026-08-07
[front] Namespace published pod functions by app folder

davidebbo:
## Description

A published pod function is identified by a slug unique per pod. Two apps in the same pod therefore cannot both own a function called `refresh` — and the failure is silent: `publishSandboxFunction` upserts on `(space, slug)`, so the second app's publish **replaces** the first app's function, bundle and contract included. The flat `w/<ws>/pods/<pod>/sandbox-functions/<slug>.ts` layout is the visible symptom of that flat namespace.

The published slug becomes `<app>__<name>`, where publish derives the prefix from the app folder the source lives in:

```
publish({ slug: "add-task", path: "pod-vlt_x/TaskList/functions/add-task.ts" })
  → tasklist__add-task
  → bundle at w/<ws>/pods/<pod>/sandbox-functions/tasklist__add-task.ts
  → dsbx function run tasklist__add-task
```

**Only the app folder contributes.** `functions/` and anything nested below it never appear in the slug, so moving a source inside its app does not rename the published function and does not orphan a Frame's `<podId>/<slug>` reference. That also keeps the source-layout question independent of identity — whether `functions/` stays a good idea can be revisited without touching any of this.

Encoding the prefix in the slug rather than nesting directories is what keeps the rest of the chain untouched. `__` already passes dsbx's `is_valid_name` (`cli/dust-sandbox/src/commands/function/mod.rs:359`, which allows `[A-Za-z0-9_-]`) and already resolves in the flat `read_dir` of `$DUST_FUNCTIONS_DIR`; and `podId/slug` stays a two-segment reference, so no route, resolution or `%2F`-encoding change either. A real nested tree would have meant Rust changes plus a dsbx build shipped into the sandbox image, in exchange for grouping in a bucket listing.

A source sitting directly at the pod root has no app folder and publishes under its bare name. That cannot collide with an app's function: a name is a single slug segment so it carries no `__`, and a normalized prefix carries none either, so every prefixed slug has exactly one and every bare slug has none — the namespaces are disjoint by construction. The skill still directs sources into an app folder, and now says why: moving one in later renames its function, leaving the old slug published and stale.

The `pod_functions` skill is updated accordingly — including the paragraph that previously promised the opposite ("it does not namespace the slug ... never includes the app name").

## Tests

Written test-first; each test was watched fail before the implementing change (the four pre-existing publish tests failed on `expected 'greet' to be 'greeter__greet'`).

- `front/lib/api/sandbox_functions/slug.test.ts` (new): prefix derivation, the ignore-intermediate-folders property, folder-name normalization, legacy `pod/...` paths, and the pod-root / empty-prefix / wrong-scope / bad-name rejections.
- `front/types/api/sandbox_functions.test.ts`: slug grammar, plus a guard that the grammar only produces names dsbx's `is_valid_name` accepts — the assumption the whole approach rests on.
- `front/lib/api/sandbox_functions/publish_sandbox_function.test.ts`: existing tests moved to app-folder paths; added the regression test that two apps publishing `refresh` yield two functions and two bundles with neither overwriting the other, and that a root source keeps its bare name.
- `front/lib/api/actions/servers/sandbox_functions/tools/publish.test.ts` (new): the handler reports the prefixed slug and the `<podId>/<slug>` reference, and the tool schema takes a bare name.

141/141 across all 18 sandbox-function suites; 240/240 across `file_resource`, `mount_path` and the skill suites. `npx tsgo --noEmit` and biome clean.

## Risk

Low. `sandbox_functions` is a `dust_only` flag, so the blast radius is our own workspaces.

No migration and no schema change: the unique index `(workspaceId, spaceId, slug)` now simply means *(pod, app, function)*. Bare slugs stay valid under the widened grammar, so existing rows and the already-published Frames that reference them keep working untouched.

The one behavioral edge: re-publishing an existing function renames it (it picks up its prefix), which breaks that Frame's reference until the Frame is re-published too. That is the same edit cycle as any function change, but worth knowing before touching a live pod app.

Rollback is a plain revert — nothing is persisted that the old code cannot read, since a reverted deploy just goes back to treating the stored `<app>__<name>` as an opaque slug.

## Deploy Plan

Standard deploy, no ordering constraints. front, the CLI and the sandbox image are independent here: no dsbx release is required, and no coordinated client rollout.

## Follow-up: the Frame reference grammar (viz)

Caught while testing a real Frame end to end. The `viz` workspace cannot import from front, so it carries its own copy of the slug grammar (`viz/app/lib/pod-function-hooks.tsx`). Widening front's copy without viz's meant `usePodFunction("<podId>/tasklist__add-task")` failed validation, resolved to a **null SWR key**, and the Frame issued no request at all — a broken function with an empty network tab, which is a nastier failure mode than a 404.

The grammar now lives in its own module (`viz/app/lib/pod-function-slug.ts`) and front asserts the two accept exactly the same slugs, so the copies cannot drift silently again. That mirrors how `cli/dust-sandbox/functions-runner/protocol.ts` is already kept in step. The guard was verified by reverting the viz side and watching it fail on `tasklist__add-task`.

viz: 51/51. front: 81/81 across the sandbox-function and slug suites.

### DIFF
--- a/front/lib/api/actions/servers/sandbox_functions/metadata.ts
+++ b/front/lib/api/actions/servers/sandbox_functions/metadata.ts
@@ -75,8 +75,8 @@ export const SANDBOX_FUNCTIONS_TOOLS_METADATA = [
         .describe(
           "Scoped path to the function's TypeScript source in the pod, as shown by the files " +
             "tools. It lives in its app's `functions` folder (e.g. " +
-            "`pod-<id>/MyApp/functions/greet.ts`). The app folder is required, and its name " +
-            "becomes the published slug's prefix; a source at the pod root is rejected."
+            "`pod-<id>/MyApp/functions/greet.ts`). The app folder's name becomes the published " +
+            "slug's prefix; a source at the pod root keeps its bare name."
         ),
       executionMode: z
         .enum(SANDBOX_FUNCTION_EXECUTION_MODES)
--- a/front/lib/api/actions/servers/sandbox_functions/metadata.ts
+++ b/front/lib/api/actions/servers/sandbox_functions/metadata.ts
@@ -3,6 +3,7 @@ import {
   POD_DATABASE_NAME_REGEX,
   SANDBOX_FUNCTION_EXECUTION_MODES,
   SANDBOX_FUNCTION_SLUG_REGEX,
+  SANDBOX_FUNCTION_SLUG_SEGMENT_REGEX,
 } from "@app/types/api/sandbox_functions";
 import { z } from "zod";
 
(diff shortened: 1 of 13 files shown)


fontanierh (approved):
code LGTM

What happens if the folder is renamed, and then the function is modified / republished ? would it make the slug non-stable ?

davidebbo:
> What happens if the folder is renamed, and then the function is modified / republished ? would it make the slug non-stable ?

You'd end up with both old_foo and new_foo functions, so treated as totally different. It's hard to tell apart "Folder is renamed" from "New folder is created with similar content, and old one is deleted later".

I could think of ways to address this by adding a true concept of an 'app' with a stable ID, but that adds different complexity.

fontanierh:
@davidebbo maybe all we need is to make it clear for agents

PR #30182, opened 2026-08-07
[front] fix: validate MCP connection type in endpoint handler

aubin-tchoi:
## Description

MCP connection detail routes contain a `connectionType` in their path but never check its value.

This PR validate the connection type and include it in the filter.

## Tests

- Added tests.

## Risk

- Low.

## Deploy Plan

- Deploy front.


aubin-tchoi:
PMRR

PR #30183, opened 2026-08-07
feat: add space scrub CLI command

adrsimon:
## Description

Adds `space scrub` to the admin CLI, to relaunch the scrub workflow of a space that was already deleted (workflow failed, got terminated, or never ran). Same shape as the existing `data-source delete`.

```
./admin/cli.sh space scrub --wId <wId> --spaceId <spaceId>
```

Refuses to run if the space isn't soft-deleted, so a live space still has to go through the Poke deletion path (which cleans up agent/skill references first).

## Tests

Manual, typecheck.

## Risk

Low, admin-only script, no runtime code touched.

## Deploy Plan

Nothing special.


aubin-tchoi (approved):
LGTM

PR #30184, opened 2026-08-07
Pricing: clean typing around poolLimitAwuCredits

fabiencelier:
## Description

The workspace default value can never be null (non nullable column defaulting to 0) so the poolLimitAwuCredits can never be null.
Change the type in UserCreditContext to reflect that
Also fixes the typing in resolveEffectiveSpendLimitAwuCredits when we know defaultAwuCredits is not null.

## Tests

updated unit tests

## Deploy Plan

deploy front

### DIFF
--- a/front/lib/metronome/user_credit_state_machine.ts
+++ b/front/lib/metronome/user_credit_state_machine.ts
@@ -44,12 +44,13 @@ export type UserCreditContext = {
   remainingCapCreditsPercentage?: number | null;
   /**
    * The user's effective pool budget in AWU credits: `0` = no pool access
-   * (e.g. free seats), `> 0` = capped pool, `null` = unlimited. A property of
-   * the user's situation (like `remainingCapCreditsPercentage`), so it lives in
-   * the context — it's what the `seat_balance_exhausted` guards use to route
-   * `capped` vs `on_pool`, with no seat-type special-casing.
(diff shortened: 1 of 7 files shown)


front/lib/metronome/user_credit_state_machine.ts
```diff
@@ -44,12 +44,13 @@ export type UserCreditContext = {
...
-   * the context — it's what the `seat_balance_exhausted` guards use to route
-   * `capped` vs `on_pool`, with no seat-type special-casing.
+   * (e.g. free seats), `> 0` = capped pool. Unlimited is not supported. A
+   * property of the user's situation (like `remainingCapCreditsPercentage`), so
+   * it lives in the context — it's what the `seat_balance_exhausted` guards use
+   * to route `capped` vs `on_pool`, with no seat-type special-casing. Absent
+   * for events that don't carry a resolved pool limit.
    */
-  poolLimitAwuCredits?: number | null;
+  poolLimitAwuCredits?: number;
```
flvndvd, line 53: Does it need to remain optional?

fabiencelier: Yes, UserCreditContext is shared accross several events and someevent do not provide it

flvndvd: Ok

flvndvd (approved):
Way less confusing \o/!

PR #30186, opened 2026-08-07
Centralize consumption analytics resource inputs

flvndvd:
## Description

Consumption analytics needs to distinguish model and tool consumption items and determine which skills exposed each called tool. Today, callers must understand persisted field combinations and remember to load MCP tool metadata correctly.

To associate tools with their respective skills, we should account that some servers are disabling tools, which we need to account for by loading the metadata.

This gives the document-building work clear, testable inputs. It does not write to Elasticsearch or activate any analytics workflow.

### DIFF
--- a/front/lib/resources/skill/skill_resource.ts
+++ b/front/lib/resources/skill/skill_resource.ts
@@ -629,6 +629,7 @@ export class SkillResource extends BaseResource<SkillConfigurationModel> {
       onlyCustom,
       withInstructions = true,
       withTools = true,
+      withToolMetadata = false,
       withFileAttachments = true,
       ...otherOptions
     } = options;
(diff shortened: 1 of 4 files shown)


front/lib/resources/skill/skill_resource.ts
```diff
@@ -3856,7 +3860,8 @@ export class SkillResource extends BaseResource<SkillConfigurationModel> {
 
   static async listByAgentMessageId(
     auth: Authenticator,
-    agentMessageId: ModelId
+    agentMessageId: ModelId,
+    { withToolMetadata = false }: { withToolMetadata?: boolean } = {}
```
flvndvd, line 3864: @aubin-tchoi for the analytics consumption, I'd need a stable way (thinking backfill) to know if a tool was enabled at the time where the message was posted. Is there any way to do this today?

avervaet (approved):

PR #30187, opened 2026-08-07
[front-api] Hide the files server from dsbx tools

davidebbo:
### Description

The sandbox mounts the file system that the `files` MCP server exposes under `/files`, so a process running in the sandbox is always better off reading and editing files directly than round-tripping through `dsbx tools exec files ...`. This hides the `files` server from the sandbox actions listing that backs `dsbx tools`.

The listing endpoint already stripped one server for a similar reason — `sandbox`, the server that runs the sandbox itself — so that single-name check becomes a named list:

```ts
// Servers the sandbox must never see: `sandbox` is the server that runs the sandbox itself, and
// the file system the `files` server exposes is mounted in the sandbox under `/files`, so direct
// file operations always beat a round-trip through the tool.
const SANDBOX_HIDDEN_SERVER_NAMES: string[] = [SANDBOX_TOOL_NAME, FILES_SERVER_NAME];
```

This applies to both token kinds (conversation exec tokens and pod function invocation tokens), and to `?server=files` as well, since the exclusion runs before the name filter — `dsbx tools list files` now reports `server 'files' not found`.

Two notes on scope:

- Only the listing is filtered. `POST /sandbox/actions/call` still resolves a `files` tool given its `serverViewId` directly; that matches the existing treatment of the `sandbox` server, which is hidden from the list but not blocked at `/call`. Happy to refuse it there too if we'd rather.
- The `pod_functions` and `projects` global skills mention falling back to the `files` MCP tools when the mount layout isn't known. Those instructions go to the main agent loop, which still sees the server, so they are unaffected.

### Risk

Low. Listing-only change on an internal endpoint, scoped to one server name.

### Deploy Plan

Deploy front-api.

### DIFF
--- a/front-api/routes/v1/w/[wId]/sandbox/actions/index.test.ts
+++ b/front-api/routes/v1/w/[wId]/sandbox/actions/index.test.ts
@@ -1,3 +1,4 @@
+import { FILES_SERVER_NAME } from "@app/lib/api/actions/servers/files/metadata";
 import { InternalMCPServerInMemoryResource } from "@app/lib/resources/internal_mcp_server_in_memory_resource";
 import { AgentConfigurationFactory } from "@app/tests/utils/AgentConfigurationFactory";
 import { AgentMCPServerConfigurationFactory } from "@app/tests/utils/AgentMCPServerConfigurationFactory";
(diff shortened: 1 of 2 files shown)


fontanierh (approved):
LGTM

PR #30188, opened 2026-08-07
fix: skip Google Sheets we are not allowed to read

adrsimon:
## Description

A spreadsheet the connector's account can't read makes `syncFiles` fail with a 403, and nothing classified that as terminal, so Temporal retried forever. One folder sync (connector 17892) had been looping since Aug 1 with 5300+ attempts, and only moved on to the next unreadable file after the first was skipped by hand.

Marks those spreadsheets as skipped instead, so the sync drains the rest of the folder on its own.

A 403 alone isn't enough to key on, Google also answers 403 for quota and rate-limit exhaustion (a few thousand a week across hundreds of connectors). So we match the permission-denied message, the way `isUnableToParseError` just above it already does.

`skipReason` is `access_forbidden`, same string Microsoft uses for its 403 path.

## Risk

Low.

## Deploy Plan

Normal deploy of connectors.

### DIFF
--- a/connectors/src/connectors/google_drive/temporal/spreadsheets.ts
+++ b/connectors/src/connectors/google_drive/temporal/spreadsheets.ts
@@ -506,6 +506,14 @@ export async function syncSpreadSheet(
               isSupported: false,
               skipReason: "google_bad_request_error",
             };
+          } else if (isCallerDoesNotHavePermissionError(err)) {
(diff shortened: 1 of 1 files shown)


adrsimon:
PMRR

aubin-tchoi (commented):
LGTM

PR #30191, opened 2026-08-07
[front] Fix sandbox function staging reads against symlink swap (TOCTOU)

fontanierh:
## Description

`buildSandboxFunctionOnSandbox` and `getDatabaseSchemaOnSandbox` stage artifacts in agent-writable `/tmp` dirs as `agent-proxied`, then read them back through the provider file API, which envd serves as root while following symlinks. Code running as the agent on the same sandbox VM could swap a staging file for a symlink to a root-only file (e.g. `/run/dust/egress-secrets.json`) between the exec and the read, and the root read would return that file's content to the caller or persist it as a published function bundle.

Each producing exec now appends a marker and per-file sha256 lines after the dsbx output, and the read-back content is verified against those hashes before use, failing closed on any mismatch:
- swap before the capture: the capture itself fails (`set -e`; `sha256sum` cannot open the swapped target as `agent-proxied`)
- swap after the capture: content no longer hashes equal
- error messages never carry content, so a swapped-in file cannot leak through the error path either

Follow-ups not in this PR: same hardening for any future `provider.readFile`/`writeFile` caller under agent-writable paths, and the dsbx port-80 Host pinning issue tracked separately.

## Tests

- New `staging_integrity.test.ts` for the split/verify helpers.
- New swap-detection tests in `build_on_sandbox.test.ts` and a new `dsbx_db.test.ts` covering the schema flow (match, swapped file, missing hash).
- Full `lib/api/sandbox_functions` and `lib/api/sandbox` suites pass locally in a dust-hive env (70 + 363 tests).

## Risk

Low. Only the two staging flows change; both fail closed on any anomaly (missing marker, missing hash, mismatch), so worst case is a failed publish/schema read that the model can retry. Rollback is a revert.

## Deploy Plan

Standard deploy.

### DIFF
--- a/front/lib/api/sandbox_functions/build_on_sandbox.test.ts
+++ b/front/lib/api/sandbox_functions/build_on_sandbox.test.ts
@@ -1,3 +1,5 @@
+import { createHash } from "node:crypto";
+
 import { ensurePodSandboxReady } from "@app/lib/api/sandbox/lifecycle";
 import { buildSandboxFunctionOnSandbox } from "@app/lib/api/sandbox_functions/build_on_sandbox";
 import { SandboxResource } from "@app/lib/resources/sandbox_resource";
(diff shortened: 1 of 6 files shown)


zmarouf (approved):
LGTM. I tried poking and the only thing I found was the sha256sum format that GNU based. 
But the check fails closed.

PR #30194, opened 2026-08-07
[front] perf: lock-free fast path for requireRunning sandbox reads

fontanierh:
## Description

Concurrent invocations of the same pod's functions degrade linearly with overlap: chatpro-sync p50 went from 101ms isolated to 1128ms at 8 concurrent callers, while the in-sandbox exec stayed flat at ~79ms. The `lock.acquire` APM span shows why: every invocation, even on a running sandbox, entered `SandboxResource.ensureActive` under the per-pod `sandbox:lifecycle` Redis lock, and the median lock wait went from 1-2ms quiet to 139-599ms during a real multi-user session. The lock waits in blind 100ms sleeps, so contenders lose in 100ms quanta even though the lock holds only two row reads on the happy path.

Two changes:

`requireRunning` no longer takes the lifecycle lock. It performs no lifecycle transition, so it runs off a plain read: running and not kill-requested means use it (plus the same throttled `lastActivityAt` touch the locked path did), anything else is `SandboxNotRunningError` without waiting behind an in-flight multi-second wake. The read is a snapshot, but every race converges the same way it did when the transition landed right after the locked version released: the exec fails, the caller escalates to the durable path, and the locked ensure there runs the transition machinery, kill-requested recreation included.

`executeWithLock` gains a `retryIntervalMs` option (default unchanged at 100ms), jittered to half..full interval so waiters stop stampeding the same retry tick. The lifecycle lock passes 25ms, which bounds the queueing penalty for the paths that still need the lock (create, wake, sleep, kill).

## Tests

sandbox_resource suite: the requireRunning block now asserts the fast path never calls `executeWithLock`, both when serving a running sandbox and when refusing, and that `lastActivityAt` is written through the fast path so the reaper's inactivity clock keeps running. Existing refusal tests (kill-requested, sleeping, absent) pass unchanged. front typechecks clean.

## Risk

The fast path can hand out a sandbox that a concurrent kill or sleep is about to invalidate. That race ends in a failed invocation, not escalation: the escalation gate is SandboxNotRunningError, and escalating on an arbitrary exec failure would risk re-running a partially executed function. Relatedly, an exec admitted during the reaper's pre-sleep flush can commit pod-state writes the flush never saw. Both windows are pre-existing in kind — the lock only ever serialized admission, never running execs, so an exec admitted before the reaper locked was always able to race the flush — but this widens admission into them. Both need the reaper to sleep a pod that is actively invoking, which the fast path's activity touch keeps minutes-rare. The proper closure (exec/sleep mutual exclusion or a post-flush activity re-check in the reaper) is left to a follow-up.

The retry-interval change only affects wait cadence, never lock semantics; the added jitter shifts the default wait from a fixed 100ms to 50-100ms for all executeWithLock callers, none of which depend on cadence.

## Deploy Plan

Normal deploy. Watch the `lock.acquire` span for `sandbox:lifecycle` (its volume should collapse) and invocation latency under concurrent load on a single pod, which should sit near its isolated latency.

### DIFF
--- a/front/tests/utils/SandboxFactory.ts
+++ b/front/tests/utils/SandboxFactory.ts
@@ -70,7 +70,11 @@ export class SandboxFactory {
   static async createForPod(
     auth: Authenticator,
     pod: SpaceResource,
-    opts?: { status?: SandboxStatus; killRequestedAt?: Date }
+    opts?: {
+      status?: SandboxStatus;
+      killRequestedAt?: Date;
+      lastActivityAt?: Date;
+    }
   ): Promise<SandboxResource> {
     const sandbox = await SandboxResource.makeNew(auth, {
       providerId: `test-provider-${Date.now()}`,
@@ -81,6 +85,12 @@ export class SandboxFactory {
     if (opts?.killRequestedAt) {
       await SandboxModel.update(
         { killRequestedAt: opts.killRequestedAt } as Partial<SandboxModel>,
+        { where: { id: sandbox.id } }
+      );
+    }
+    if (opts?.lastActivityAt !== undefined) {
+      await SandboxModel.update(
+        { lastActivityAt: opts.lastActivityAt } as Partial<SandboxModel>,
         { where: { id: sandbox.id } }
       );
     }
--- a/front/lib/lock.ts
+++ b/front/lib/lock.ts
@@ -48,7 +48,7 @@ export async function distributedUnlock(
   });
 }
 
-const WAIT_BETWEEN_RETRIES = 100;
+const DEFAULT_RETRY_INTERVAL_MS = 100;
 
 export const executeWithLock = async <T>(
   lockName: string,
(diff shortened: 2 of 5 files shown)


front/tests/utils/SandboxFactory.ts
```diff
@@ -84,6 +88,12 @@ export class SandboxFactory {
         { where: { id: sandbox.id } }
       );
     }
+    if (opts?.lastActivityAt !== undefined) {
+      await SandboxModel.update(
```
davidebbo, line 92: It could make both updates in one query when applicable, but not sure if worth it.

davidebbo (approved):
LGTM

front/tests/utils/SandboxFactory.ts
fontanierh: It's a test factory so I guess we don't really care

PR #30195, opened 2026-08-07
Fix: Pod's frame pinned or as tab use the same rendering path.

Fraggle:
## Description

`PodPinnedBanner` and `PodFrameTabContent` each had their own divergent fetch-and-render logic for Pod frames: the banner used `usePodFiles` + `useFileContent` (raw mount content), while the tab used `getFilePathContentApiPath` + `useFileContentByUrl`. Neither resolved through `?action=view`, meaning published frame bundles were never used.

- New `usePodFrameRenderableContent` hook unifies path → `fileId` → renderable content resolution for both consumers, using `useFileIdFromPath` (new SWR hook in `files.ts`) + `useFileContent` with `?action=view` so the published bundle is fetched when available.
- New `PodFrameVisualization` shared component replaces the inline `AuthenticatedVisualizationActionIframe` usage in both call sites. The iframe `identifier` now uses `fileId` (stable) instead of `tab.path`.
- `PodPinnedBanner` drops the `usePodFiles` dependency entirely; the `unpinLabel` falls back to `path.split("/").pop()`.

## Tests

Tested locally — pinned frame banner and frame tab both render the published bundle. Verified the "no longer available" fallback when the path is not found.

## Risk

Low. Front-only, no DB changes. The only behavioral shift is that published frame bundles are now served via `?action=view` instead of raw path content — this is the intended behavior. Rollback is a simple front redeploy.

## Deploy Plan

Deploy `front`.

### DIFF
--- a/front/components/pod/PodFrameTabContent.tsx
+++ b/front/components/pod/PodFrameTabContent.tsx
@@ -1,14 +1,10 @@
-import { AuthenticatedVisualizationActionIframe } from "@app/components/assistant/conversation/actions/AuthenticatedVisualizationActionIframe";
+import { PodFrameVisualization } from "@app/components/pod/PodFrameVisualization";
+import { usePodFrameRenderableContent } from "@app/hooks/usePodFrameRenderableContent";
 import { useAuth } from "@app/lib/auth/AuthContext";
-import {
-  getFilePathContentApiPath,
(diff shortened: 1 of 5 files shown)


Fraggle:

- #30195 ◀

fontanierh (approved):
Nice !

PR #30199, opened 2026-08-07
[front] feat: add conversation_side_panel MCP server to open frames

frankaloia:
## Summary

- Adds a new internal MCP server (`conversation_side_panel`) with two tools:
  - `open_frame`: opens a specific file frame in the side panel by file ID
  - `set_files_side_panel`: shows or hides the conversation files panel
- The purpose is that for activation I need more control over the visualization to avoid a confusing experience I've seen from some customer sessions. I need the agent to be able to open a frame already created if it's referring to it. I also need to be able to strategically hide the files side panel to avoid confusing customers (and I don't think this should be a global change)

## Testing

- Localhost testing with my activation agent to confirm tool calls operate as expected

## Risk

Low — new server is `auto_hidden_builder` (not user-visible in builder) and is only auto-attached to the activation skill. No existing tool behaviour changes.

## Deploy
1. Deploy front


adrsimon (approved):
LGTM!

PR #30206, opened 2026-08-07
feat(sandbox): pod env vars list and create routes

smb2268:
## Description

Follow up of #28862 (pod env vars injected at sandbox boot). First of two route PRs — splitting the pod env vars API by route to keep each reviewable: this PR adds **list and create** (`GET`/`POST /api/w/:wId/spaces/:spaceId/sandbox/env-vars`); update/delete follow in #29348.

Workspace-admin + `sandbox_functions` gates apply at the sandbox router. Pods are project spaces — the kind check lives in `withSpace` as a `requireProject` option (one predicate beside the existing `isConversations` kind-check; additive and default-off, existing consumers unchanged), so the handlers don't each repeat the guard. Create goes through the scope-aware resource `upsert` — re-posting an existing name rotates its value: 201 on create, 200 on overwrite; per-pod encryption key, per-scope name uniqueness.

## Risk

New endpoints, admin + FF gated, no consumers until the UI PR. The `withSpace` option is additive and default-off. Safe to rollback.

## Tests

List/create functional tests: gating (non-admin 403, missing FF 403), non-project space 404, create + overwrite (201/200), invalid body 400, list scoping (this pod only — not workspace or other pods' vars). Pod audit attribution (space_id metadata + schemas) lands with #29346 upstack.

### DIFF
--- a/front-api/middlewares/with_space.ts
+++ b/front-api/middlewares/with_space.ts
@@ -14,6 +14,8 @@ interface WithSpaceOptions {
   requireCanReadOrAdministrate?: boolean;
   requireCanRead?: boolean;
   requireCanWrite?: boolean;
+  // Pods are project spaces — pod-scoped surfaces 404 on any other kind.
+  requireProject?: boolean;
   routeParam?: "spaceId" | "podId";
 }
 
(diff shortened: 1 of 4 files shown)


davidebbo (approved):
LGTM

PR #30207, opened 2026-08-07
Keep Front readiness healthy during GKE endpoint draining

flvndvd:
## What changes

- `/api/healthz/ready` continues to return HTTP 200 while a Pod is terminating.
- The `preStop` endpoint stays open for a fixed drain window instead of returning early when wake locks clear.
- `PRESTOP_DRAIN_DURATION_SECONDS` configures the window. The application default is 120 seconds.
- Duplicate `preStop` requests share the original process-wide promise and deadline.
- Wake locks are still logged and measured. Active locks no longer extend the hook past the configured deadline.
- The shell hook treats non-2xx responses and connection failures as failures, then sleeps for the remaining configured duration.
- The API shutdown path no longer changes the global shutdown signal. The delayed worker shutdown signal is unchanged.

Companion infrastructure change: https://github.com/dust-tt/dust-infra/pull/961.

## What was happening

The old `preStop` path immediately called `markShuttingDown()`. The readiness endpoint saw that process flag and returned HTTP 503 until Kubernetes killed the container.

That 503 did not identify a sick Pod. It was generated because Kubernetes was already deleting a healthy Pod. Our probe-failure monitor counted it like any other readiness failure, so normal deployments created on-call noise roughly 20 times per day.

The useful part of the monitor remains. A dead process, an event-loop stall, a connection failure, or a readiness response that exceeds the probe timeout can still fail the probe. This change removes only the deliberate termination-time 503.

## Why HTTP 200 is correct during `preStop`

Kubernetes does not require a readiness failure to drain a deleting Pod. On deletion, Kubernetes independently marks the EndpointSlice endpoint as terminating and sets `ready: false` for ordinary Service traffic. Readiness probes continue during the container lifecycle, but their result is not the deletion signal.

Kubernetes starts local shutdown and endpoint retirement at about the same time. It does not wait for every consumer of the endpoint state before starting `preStop`. Keeping the server healthy allows requests that still reach the Pod during that race to complete.

GCP HTTP, HTTPS, and HTTP/2 health checks require an exact HTTP 200 response. Other status codes, including 3xx responses, are unhealthy. Returning 200 keeps the endpoint serviceable while GKE removes it from the NEG and the load balancer drains connections.

Findings are coming from:
- [Kubernetes readiness probes](https://kubernetes.io/docs/concepts/workloads/pods/probes/#readiness-probe)
- [Kubernetes Pod termination flow](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-termination-flow)
- [Kubernetes EndpointSlice conditions](https://kubernetes.io/docs/reference/kubernetes-api/discovery/endpoint-slice-v1/#endpointconditions)
- [GCP HTTP health check success criteria](https://cloud.google.com/load-balancing/docs/health-check-concepts#criteria-protocol-http)

## Timing change

The old hook duration was:

```text
10s assumed load-balancer propagation + max(60s minimum drain, wake-lock wait up to 120s)
```

It therefore took between 70 and 130 seconds. A Pod with no wake locks usually reached SIGTERM after about 70 seconds.

The new hook uses a fixed duration:

```text
Front and Front Edge: 120s
Front SSE:            180s, supplied by the companion infrastructure PR
```

The fixed duration is a floor, not only a ceiling. Clearing application wake locks does not prove that GKE has removed the endpoint from the NEG. There is no acknowledgement available to the application that would justify ending the hook early.

## Failure behavior

The kubelet executes `admin/prestop.sh`, which calls the application endpoint. The shell now uses a bounded curl request and rejects non-2xx responses. If the secret is wrong, the route is missing, or the local API cannot be reached, the shell sleeps for whatever remains of the configured drain duration.

This prevents an internal hook failure from collapsing a 120 or 180 second drain window into an immediate SIGTERM.

## Operational impact

- Regular Front Pods without wake locks spend about 50 seconds longer in `preStop`, moving from about 70 seconds to 120 seconds.
- Front SSE Pods spend 180 seconds in `preStop` after the companion infrastructure configuration lands.
- Pod deletion, scale-down, node drain, rollback cleanup, and some rollout completion checks can take longer.
- RollingUpdate starts replacement Pods while old Pods drain, so deploy duration is not the drain duration multiplied by the replica count. Old and new replicas overlap for longer.

The intended outcomes are concrete. Normal Pod termination stops generating readiness-failure alerts, and Pods remain able to finish requests while GKE retires their endpoints.

## Rollout order

Deploy this application PR first, then apply the companion infrastructure PR promptly.

With the current 130 second Pod grace period, the new 120 second default leaves only about 10 seconds for hook overhead, SIGTERM delivery, and process exit. The companion infrastructure PR raises the regular grace period to 150 seconds and configures Front SSE for a 180 second hook inside a 210 second grace period.

### DIFF
--- a/front/lib/api/config.ts
+++ b/front/lib/api/config.ts
@@ -1,3 +1,4 @@
+import { DEFAULT_PRESTOP_DRAIN_DURATION_MS } from "@app/lib/constants/timeouts";
 import { isDevelopment } from "@app/types/shared/env";
 import { EnvironmentConfig } from "@app/types/shared/utils/config";
 
@@ -31,6 +32,24 @@ export function getDefaultInit(): Promise<RequestInit> | null {
 }
 
 const config = {
+  getPreStopDrainDurationMs: (): number => {
+    const value = EnvironmentConfig.getOptionalEnvVariable(
+      "PRESTOP_DRAIN_DURATION_SECONDS"
+    );
+    if (!value) {
+      return DEFAULT_PRESTOP_DRAIN_DURATION_MS;
+    }
+
+    const durationSeconds = Number(value);
+    if (!Number.isSafeInteger(durationSeconds) || durationSeconds <= 0) {
+      throw new Error(
+        "PRESTOP_DRAIN_DURATION_SECONDS must be a positive integer"
+      );
+    }
+
+    return durationSeconds * 1_000;
+  },
+
   // Dynamic API base URL: uses a custom resolver when set (SPA region switching),
   // otherwise falls back to getClientFacingUrl().
   getApiBaseUrl: (): string => {
--- a/front-api/routes/healthz.test.ts
+++ b/front-api/routes/healthz.test.ts
@@ -1,5 +1,4 @@
 import { frontSequelize } from "@app/lib/resources/storage";
-import * as shutdownSignal from "@app/lib/shutdown_signal";
 import { honoApp } from "@front-api/app";
 import { beforeEach, describe, expect, it, vi } from "vitest";
 
@@ -19,11 +18,7 @@ describe("GET /api/healthz", () => {
 });
 
 describe("GET /api/healthz/ready", () => {
-  beforeEach(() => {
-    vi.spyOn(shutdownSignal, "isInShutdown").mockReturnValue(false);
-  });
-
-  it("returns 200 with status ready when not shutting down", async () => {
+  it("returns 200 with status ready", async () => {
     const response = await honoApp.request("/api/healthz/ready");
 
     expect(response.status).toBe(200);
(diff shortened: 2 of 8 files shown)


front/lib/api/config.ts
```diff
@@ -31,6 +32,24 @@ export function getDefaultInit(): Promise<RequestInit> | null {
 }
 
 const config = {
+  getPreStopDrainDurationMs: (): number => {
```
flvndvd, line 35: Env variable is configured in the infra PR https://github.com/dust-tt/dust-infra/pull/961.

adrsimon (approved):

PR #26181, opened 2026-05-26
[Webhooks] Skip unused payload storage

philipperolet:
## Description
Related to https://github.com/dust-tt/tasks/issues/8104
Follows https://github.com/dust-tt/dust/pull/26048 and https://github.com/dust-tt/dust/pull/26061

Webhook requests no longer store the raw payload before validation and trigger matching. We only persist the payload after at least one trigger matches and at least one matched trigger has `includePayload` enabled.

This avoids GCS writes for unmatched webhook traffic while keeping payload attachment behavior unchanged for triggers that need the body in the launched conversation.

## Risks
Blast radius: webhook trigger request processing and recent webhook request payload display.
Risk: low

## Deploy Plan
- deploy front

## Tests
- [x] `NODE_ENV=test npm test -- --dir pages/api/v1/w/[wId]/triggers/hooks/[webhookSourceId]/[webhookSourceUrlSecret]`

PR #30210, opened 2026-08-07
[front] fix: store unmatched trigger payloads in GCS for iterating on filter

aubin-tchoi:
## Description

Fixes https://github.com/dust-tt/tasks/issues/9978

https://github.com/dust-tt/dust/pull/26181 changed the logic for storing trigger payloads to avoid storing excessive amounts of data. It filters out payloads that did not match the filter expression.

This PR changes this behavior to store the payload in GCS as long as one webhook trigger is enabled with `"includePayload"` to true.
The rationale is that even if no trigger matched, we need the payload for the user to iterate on the filter expression. For their first trigger on the webhook source this is currently practically impossible to iterate on the filter.

Also removes an N+1 query.

## Risk

- Low.

## Deploy Plan

- Deploy front.

### DIFF
--- a/front/lib/triggers/webhook.ts
+++ b/front/lib/triggers/webhook.ts
@@ -424,45 +424,29 @@ function matchesPayloadFilter({
   return false;
 }
 
-async function filterTriggers({
-  auth,
-  webhookSource,
-  receivedEventValue,
-  webhookRequest,
-  body,
-}: {
-  auth: Authenticator;
-  webhookSource: { id: number; provider: WebhookProvider | null };
-  receivedEventValue: string | null;
-  webhookRequest: WebhookRequestResource;
-  body: Record<string, unknown>;
-}): Promise<Result<WebhookTriggerType[], Error>> {
+async function filterTriggers(
+  auth: Authenticator,
+  {
+    enabledWebhookTriggers,
+    webhookSource,
+    receivedEventValue,
+    webhookRequest,
+    body,
+  }: {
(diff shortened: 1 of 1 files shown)


adrsimon (approved):

front/lib/triggers/webhook.ts
```diff
@@ -425,48 +424,29 @@ function matchesPayloadFilter({
...
-      views,
-      async (view) => {
-        return TriggerResource.listByWebhookSourceViewId(auth, view.id);
-      },
-      { concurrency: 10 }
-    )
-  )
-    .flat()
+  const triggers = enabledWebhookTriggers
     .map((t) => t.toJSON())
```
aubin-tchoi, line 447: not ideal to the use the serialized type here, but the typing is not amazing on the resource, specifically we're missing a type that ties the kind (schedule | webhook) with the type of the config

front/lib/triggers/webhook.ts
```diff
@@ -718,8 +698,42 @@ export async function processWebhookRequest(
...
+    views.map((view) => view.id)
+  );
+  const enabledWebhookTriggers = webhookTriggers.filter(
+    ({ status }) => status === "enabled"
+  );
+
+  if (
+    enabledWebhookTriggers.some(
+      (t) =>
+        "includePayload" in t.configuration && t.configuration.includePayload
```
aubin-tchoi, line 723: not ideal but same root cause as the comment above

PR #30215, opened 2026-08-07
Refactor how we determine isXXX flags on attachment, replace with inline usage line, centralize logic.

Fraggle:
## Description

Centralizes `isQueryable`/`isIncludable`/`isSearchable` flag computation into a single `computeFileAttachmentCapabilityFlags()` function inside `attachments.ts`, gated on a new `AttachmentCapabilityContext` (`{ isNewFileExplorer, hasSandboxTools }`). Previously, each call site re-derived these flags inline, leading to drift between JIT server eligibility checks, XML rendering, and tool footprint measurement.

- New `attachment_capabilities.ts` exposes `getAttachmentCapabilityContext(auth, conversation)` as the single resolver; callers obtain it once per rendering pass and thread it down.
- `<attachment>` XML tags drop the `isIncludable/isQueryable/isSearchable` boolean attributes in favour of an inline `Use: read with \`…\`; …` line listing only the tools registered in the current mode.
- `getJITServers` / `getConditionalJITServers` / `getQueryTablesServer` drop `hasSandboxTools` from their signatures — `attachment.isQueryable` is now already false when the sandbox handles tabular files, so the flag is redundant.
- `getAttachmentFromContentFragment`, `getAttachmentFromFileContentFragment`, and `getAttachmentFromContentNodeContentFragment` switch to named-object arguments.
- Two small bundled fixes: inverted `Zap`/`ZapOff` icon in `PodConversationsTab` when hiding triggered conversations; missing space after ⚡ in triggered conversation titles.

## Tests

Existing unit tests in `attachments.test.ts`, `conversation_files/tools/index.test.ts`, `tool_footprint.test.ts`, and `helpers.test.ts` updated to pass `capabilities`. New test cases added for `isNewFileExplorer` and `hasSandboxTools` flag combinations, the per-mode `Use:` line content, and the omission of `Use:` when content is inlined. `message_rendering.test.ts` stubs `getAttachmentCapabilityContext` via `vi.mock`.

## Risk

Low. Pure front-only refactor with no DB changes, no API contract changes, and no new feature flags. The `Use:` line replaces boolean attributes in the prompt, which is a behaviour change for the model, but the information is equivalent and more actionable. Rollback is a single front deploy revert.

## Deploy Plan

Deploy `front`.

### DIFF
--- a/front/components/pod/conversation/PodConversationsTab.tsx
+++ b/front/components/pod/conversation/PodConversationsTab.tsx
@@ -297,7 +297,7 @@ export function PodConversationsTab({
                   <Button
                     size="sm"
                     variant="outline"
-                    icon={hideTriggeredConversations ? Zap : ZapOff}
+                    icon={hideTriggeredConversations ? ZapOff : Zap}
                     tooltip={
                       hideTriggeredConversations
                         ? "Show triggered"
--- a/front/lib/api/assistant/conversation/attachments.ts
+++ b/front/lib/api/assistant/conversation/attachments.ts
@@ -1,5 +1,12 @@
 // All mime types are okay to use from the public API.
 
+import { DEFAULT_CONVERSATION_QUERY_TABLES_ACTION_NAME } from "@app/lib/actions/constants";
+import { getPrefixedToolName } from "@app/lib/actions/tool_name_utils";
+import {
+  CONVERSATION_CAT_FILE_ACTION_NAME,
+  CONVERSATION_FILES_SERVER_NAME,
+  CONVERSATION_SEARCH_FILES_ACTION_NAME,
+} from "@app/lib/api/actions/servers/conversation_files/metadata";
 import {
   isConversationIncludableFileContentType,
   isQueryableContentType,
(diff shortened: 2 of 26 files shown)


Fraggle:

- #30215 ◀

front/components/pod/conversation/PodConversationsTab.tsx
```diff
@@ -300,7 +300,7 @@ export function PodConversationsTab({
                   <Button
                     size="sm"
                     variant="outline"
-                    icon={hideTriggeredConversations ? Zap : ZapOff}
```
Fraggle, line 303: Unrelated change seen with @ClementAupiais

front/lib/api/assistant/conversation/attachments.ts
```diff
@@ -150,9 +171,63 @@ function shouldSuppressTabularAttachmentHints({
...
-  cf: FileContentFragmentType
-): FileAttachmentType | null {
+/**
+ * Capability flags for file attachments. Gated early so callers (JIT, Use: lines, tools) can
+ * trust the booleans without re-checking file-explorer / Computer availability.
+ *
+ * In file explorer mode every flag is off: files are reached by path through the `files` MCP server
+ * and tabular files are analyzed by the Computer, so none of the conversation_files JIT tools apply.
+ */
+function computeFileAttachmentCapabilityFlags({
```
Fraggle, line 181: This is the main thing, I centralized all flags computation here and then the remaining of the code trust theses flags for enabling JIT tools and rendering of the attachments.

Unfortunately, this required many callers to be updated to pass `isNewFileExplorer` &  `hasSandboxTools`.

PR #30216, opened 2026-08-07
Index group IDs in consumption analytics

flvndvd:
## Description

To offer the "groups" dimension in the consumption analytics, this adds `group_ids` to consumption analytics documents. Each document stores the manual and provisioned groups the triggering user belonged to when the message completed. Authorless messages use an empty array.

Only stable group IDs are stored. Group names and other metadata remain in Postgres, where group rows and membership history can be retained for historical resolution and backfills.

> [!WARNING]
> We do hard delete groups, so we can't guarantee that groups will still be available to be fetched of if we need to backfill. I considered storing more in the mapping but that would require to add a `nested` object, and do `reverse_nested` query sometimes. Definitely something that we can avoid ideally. 

## Deploy Plan

- [x] Apply mapping update
- [ ] Deploy front

### DIFF
--- a/front/types/assistant/analytics.ts
+++ b/front/types/assistant/analytics.ts
@@ -129,6 +129,8 @@ export interface AgentMessageConsumptionAnalyticsAgent {
 
 export interface AgentMessageConsumptionAnalyticsUser {
   id: string;
+  // Group sIds the user belonged to when the message completed.
+  group_ids: string[];
 }
 
 export interface AgentMessageConsumptionAnalyticsTool {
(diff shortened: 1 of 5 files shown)


avervaet (approved):

front/types/assistant/analytics.ts
```diff
@@ -243,6 +243,8 @@ interface AgentMessageConsumptionAnalyticsBaseData
   // billed cost.
   credit_micro: number;
   execution_time_ms: number | null;
+  // Group sIds the user belonged to when the message completed.
+  group_ids: string[];
```
sfriquet: it would make more sense within `user` as `user.group_ids` no ?

flvndvd: Yeah, seems fair. Let me update.

sfriquet (approved):

PR #30218, opened 2026-08-07
chore(front): enforce separate type imports

flvndvd:
## Description

In the commit (https://github.com/dust-tt/dust/commit/08cffdf017fbd7730e070171c9482d24b2dbd7a4), we actually lost this rule, time to bring it back. See documentation [here](https://biomejs.dev/linter/rules/use-import-type/). It was only scoped to `front`.


adrsimon (approved):

aubin-tchoi (approved):
LGTM

PR #30224, opened 2026-08-07
Fix empty static MCP server names

frankaloia:
## Summary

Caused by: https://github.com/dust-tt/dust/pull/30107
eng runner: https://github.com/dust-tt/tasks/issues/9980

Following the addition of custom MCP server view names, first-time setup of integrations using the static-credential flow could fail with:

```text
invalid_request_error: viewName must be a non-empty string
```

The connection-name field is hidden for a first server instance and defaults to an empty string. The standard MCP creation flow trims and omits that value, but the static-credential flow forwarded `viewName: ""` directly to the API. The backend interpreted it as an explicitly supplied name and correctly rejected it.

This affected integrations using the static-credential creation path, while integrations following the standard creation flow continued to work.

## Testing
* Manual connected in localhost. did not work before now it does.

## Risk
Low - Fixing issue

## Deploy
1. Deploy front


aubin-tchoi (approved):
LGTM 🙏🏻

PR #30234, opened 2026-08-08
[dsbx] feat: per-invocation context for pod runtime env

fontanierh:
## Description

First PR of a stack that makes warm function servers serve invocations concurrently (like a web server) instead of one at a time.

Today the warm server injects per-invocation state (user identity, sandbox token) by swapping process.env around each invocation, which forces strict serialization. This PR moves that state into an AsyncLocalStorage invocation context:

- New pod SDK surface: `podEnv(name)` reads the active invocation's env; `currentUser()` and `db()` now go through it. Outside any context (cold runs, local use) it reads process.env, unchanged behavior.
- Inside a context, only the context env is consulted. A key absent there stays absent even if process.env has a value, so one invocation's environment can never leak into another.
- The storage is shared between the runner bundle and the vendored `@dust/pod` through the `Symbol.for` registry: they are distinct module graphs (the runner is pre-bundled, the package resolves via NODE_PATH), so a module-level singleton would not be shared.
- `invoke()` takes the per-invocation env as an optional argument. The v1 warm server still swaps process.env, so nothing changes until the concurrent server lands on top.

## Tests

Unit tests for the context (isolation across concurrent flows, no fallback leakage), context-aware `currentUser()`, and `invoke()` with an invocation env. The runner-side test fixture imports the pod package's copy of the context module, so the cross-module-graph Symbol.for contract is exercised for real.

## Risk

No behavior change on its own: the cold path reads process.env as before and the v1 server still swaps env. The SDK change ships with the next base image bump.

## Deploy Plan

Merges freely; takes effect with the stack's dsbx release + base image bump (last PR of the stack).

### DIFF
--- a/cli/dust-sandbox/pod/context.ts
+++ b/cli/dust-sandbox/pod/context.ts
@@ -1,0 +1,67 @@
+// Per-invocation context.
+//
+// The function runner executes every invocation inside an AsyncLocalStorage
+// context carrying that invocation's environment (user identity, sandbox
+// token, pod paths), so one process can serve concurrent invocations without
+// mutating process.env. Runtime code reads the environment through podEnv():
+// the active context's env inside an invocation, process.env otherwise (cold
+// runs and local use, where the process environment IS the invocation's).
+//
+// The storage instance is shared with the runner through the process-wide
+// `Symbol.for` registry rather than a module-level singleton: the runner is a
+// pre-bundled artifact and @dust/pod resolves through NODE_PATH, so the two
+// are distinct module graphs that would each get their own module state. The
+// runner's enter side of this contract lives in
+// functions-runner/context.ts and must use the same key.
+
+import { AsyncLocalStorage } from "node:async_hooks";
+
+export interface InvocationContext {
+  readonly env: Readonly<Record<string, string>>;
+}
+
+/** Registry key of the shared storage. Versioned: a breaking change to the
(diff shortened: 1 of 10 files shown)


cli/dust-sandbox/pod/context.ts
```diff
@@ -0,0 +1,67 @@
...
+export interface InvocationContext {
+  readonly env: Readonly<Record<string, string>>;
+}
+
+/** Registry key of the shared storage. Versioned: a breaking change to the
+ * store shape must change the key so mismatched runner/SDK builds ignore
+ * each other instead of misreading the store. */
+export const INVOCATION_CONTEXT_KEY = "dust.pod.invocation-context.v1";
+
+function contextStorage(): AsyncLocalStorage<InvocationContext> {
```
davidebbo, line 28: Can the two contexts share logic? A number of things look identical.

davidebbo (commented):
LGTM

cli/dust-sandbox/pod/context.ts
fontanierh: not really because they are entirely separate module graphs that cannot share code: The runner is pre-bundled into runner.js and embedded in the dsbx binary and @dust/pod is vendored into the image's global node_modules and resolved by bundles at import time via NODE_PATH

PR #30235, opened 2026-08-08
[dsbx] feat: concurrent warm function server with bounded admission

fontanierh:
## Description

The per-function warm server serves one invocation at a time: overlapping calls get a busy reply and fall back to a cold run each (18% of runs in a real multi-user session), paying a fresh bun spawn and import under exactly the load where that hurts most.

This PR makes the server concurrent. Pod function calls are mostly IO-bound (SQL, API calls), so one bun process interleaves many of them on its event loop like a web server:

- Up to 32 invocations in flight, then a FIFO queue (128 max, 2s wait max), then a structured `overloaded` outcome. Refusing is deliberate: cold-run fan-out under saturation is what would OOM the sandbox. Front learns the `overloaded` code so it surfaces as itself.
- Caller identity rides the invocation context from the previous PR instead of a process.env swap, which is what makes concurrency safe. Concurrent requests with different users observe only their own env (pinned by test).
- The ack-before-execute contract is unchanged, and is why this stays a raw line protocol instead of moving behind an HTTP server: the ack is a synchronous socket write whose failure proves the client is gone, so an abandoned queued request can never race into a duplicate execution.
- Staleness, the lifetime cap, an RSS threshold and hung invocations now drain the server (stop listening, free the socket for the successor, finish in-flight work, exit) instead of killing it mid-flight. The client bounds its pre-ack wait at 4s (covering the 2s queue deadline); post-ack losses remain failed invocations, never re-runs.

v1 and v2 never talk to each other (version-suffixed socket names); the cold path covers the transition.

## Tests

bun, against the real server process: overlapping invocations complete concurrently (3x400ms in <1s), queueing serves once a slot frees, a queued request outwaiting the 2s deadline gets `overloaded` with no ack, concurrent invocations with different envs observe only their own across awaits, a stale-triggered drain lets in-flight work finish before exit. Cargo e2e against real bun covers the full miss/spawn/hit/republish cycle.

## Risk

Inert until the stack's dsbx release + image bump. One behavior delta vs v1: saturation yields `overloaded` outcomes instead of cold runs (intended; front surfaces it as a failed invocation).

## Deploy Plan

Ships with the stack (see the stack's last PR). Watch `sandbox.functions.run` by `runner_kind`: cold share under load should drop toward zero, with `overloaded` only under genuine saturation.

### DIFF
--- a/cli/dust-sandbox/functions-runner/fixtures/sleepy.ts
+++ b/cli/dust-sandbox/functions-runner/fixtures/sleepy.ts
@@ -1,0 +1,10 @@
+// Sleeps for a caller-chosen duration so concurrency tests can control how
+// long an invocation occupies its slot.
+export default {
+  async fetch(req: Request): Promise<Response> {
+    const url = new URL(req.url);
+    const delayMs = Number(url.searchParams.get("delayMs") ?? "100");
+    await new Promise((resolve) => setTimeout(resolve, delayMs));
+    return Response.json({ done: true });
+  },
+};
--- a/cli/dust-sandbox/functions-runner/fixtures/slow.ts
+++ b/cli/dust-sandbox/functions-runner/fixtures/slow.ts
@@ -1,7 +1,0 @@
-// Sleeps briefly so warm-server tests can observe the busy reply.
-export default {
-  async fetch(): Promise<Response> {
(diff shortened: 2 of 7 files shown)


davidebbo (commented):
LGTM

PR #30236, opened 2026-08-08
[dsbx] perf: content-addressed bundle cache, hash-addressed warm imports

fontanierh:
## Description

Front stamps every invocation with the publish-time bundle sha256. This PR turns that stamp into a content-addressed local cache (`~/.dust-fn/bundles/<sha256>.js`) with two payoffs:

- Cold runs: a stamped invocation whose bundle is cached skips the gcsfuse readdir and read entirely, the dominant cost of a first invocation. Population happens after a resolved cold run and re-hashes the bytes it just read, so gcsfuse serving stale bytes cannot pin the stale version under the new hash. Entries unused for 7 days are pruned. The fuse mount's deliberate no-cache config stays untouched.
- Warm republishes stop recycling workers: the module cache is keyed by path, so importing a republished bundle from its cache path loads the new version alongside the old module. The worker's other bundles (the rest of the app) keep their warmth; the abandoned module lingers until the routine RSS recycle. A stamped mismatch with no cache entry refuses without draining: the cold run populates the cache and the next warm request imports it. Only unstamped legacy staleness still drains.

Correct by construction: the cache key is the publish-time hash, so a republish (new hash) can never be served old bytes.

## Tests

Cargo: populate/lookup roundtrip, refusal of bytes not matching the stamp, malformed and path-traversal hashes touch nothing on disk, prune removes only old entries. bun, against the real worker: a stamped republish is served from the cache with a fresh import while the sibling function stays warm and nothing exits; a stamped mismatch without a cache entry refuses and the worker keeps serving.

## Risk

Best-effort everywhere: any cache irregularity falls back to today's fuse path. Disk bounded by pruning and one entry per publish; abandoned modules from republishes are bounded by the RSS recycle.

## Deploy Plan

Ships with the stack (see #30235).

### DIFF
--- a/cli/dust-sandbox/Cargo.lock
+++ b/cli/dust-sandbox/Cargo.lock
@@ -326,6 +326,7 @@ dependencies = [
  "lru",
  "rcgen",
  "reqwest",
+ "ring",
  "rustix",
  "rustls",
  "rustls-native-certs",
--- a/cli/dust-sandbox/Cargo.toml
+++ b/cli/dust-sandbox/Cargo.toml
@@ -21,6 +21,8 @@ httparse = "1"
 serde = { version = "1", features = ["derive"] }
 serde_json = "1"
 reqwest = { version = "0.12", default-features = false, features = ["json", "rustls-tls-native-roots"] }
+# Already in the tree through rustls/rcgen; used directly for bundle hashing.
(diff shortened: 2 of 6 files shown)


davidebbo (commented):
LGTM

PR #30238, opened 2026-08-10
[front] fix: hide agent model names in Manage Agents on smaller tables

aubin-tchoi:
## Description

On laptop-sized screens, model names take too much space in the Manage Agents table and become hard to read.

The Model column now shows only the provider icon while the table is narrow, with the full model name available on hover. The inline name returns on wider tables, and the existing mobile behavior remains unchanged.

## Tests

- Tested locally.

## Risk

- Low.

## Deploy Plan

- Deploy front.

### DIFF
--- a/front/components/assistant/manager/AssistantsTable.tsx
+++ b/front/components/assistant/manager/AssistantsTable.tsx
@@ -196,14 +196,19 @@ const getTableColumns = ({
             tooltipTriggerAsChild
             label={modelName}
             trigger={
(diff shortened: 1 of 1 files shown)


ClementAupiais (approved):
Need a tooltip to give the model name on hover

aubin-tchoi:
TIL that `tooltipTriggerAsChild` does not work if the child is a `DataTable.CellContentt`

PR #30239, opened 2026-08-10
[MCP] Add Shopify internal MCP server (preview)

TimotheeCrouzet:
## Description

Register a Shopify internal MCP server behind the `shopify_tool` feature flag (`isPreview`, no OAuth yet, uses a merchant-provided token). This PR is the scaffolding only: tool contracts (metadata), server registration, and the feature flag. Every tool handler is a not-implemented stub; the implementations land in a follow-up PR.

## Tests

Manual, against a real Shopify store.

## Risk

Low. Gated behind shopify_tool.

### DIFF
--- a/front/lib/actions/mcp_internal_actions/constants.ts
+++ b/front/lib/actions/mcp_internal_actions/constants.ts
@@ -1280,7 +1280,7 @@ export const INTERNAL_MCP_SERVERS = ensureUniqueToolNames({
     metadata: SERVICENOW_SERVER,
   },
   shopify: {
-    id: 1044,
+    id: 1046,
     availability: "manual",
     allowMultipleInstances: true,
     isRestricted: ({ featureFlags }) => {
--- a/front/lib/actions/mcp_internal_actions/constants.ts
+++ b/front/lib/actions/mcp_internal_actions/constants.ts
@@ -79,6 +79,7 @@ import {
 import { SANDBOX_FUNCTIONS_SERVER } from "@app/lib/api/actions/servers/sandbox_functions/metadata";
 import { SEARCH_SERVER } from "@app/lib/api/actions/servers/search/metadata";
 import { SERVICENOW_SERVER } from "@app/lib/api/actions/servers/servicenow/metadata";
+import { SHOPIFY_SERVER } from "@app/lib/api/actions/servers/shopify/metadata";
 import { SKILL_AUTHORING_SERVER } from "@app/lib/api/actions/servers/skill_authoring/metadata";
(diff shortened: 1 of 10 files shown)


front/lib/api/actions/servers/shopify/metadata.ts
```diff
@@ -0,0 +1,147 @@
...
+      startDate: z
+        .string()
+        .optional()
+        .describe("ISO 8601 lower bound for order creation date (inclusive)."),
+      endDate: z
+        .string()
+        .optional()
+        .describe("ISO 8601 upper bound for order creation date (inclusive)."),
+      financialStatus: z
+        .enum([
```
aubin-tchoi, line 82: you could pass a string here and validate in the tool handler to avoid an overly verbose JSON schema

aubin-tchoi: same for every other enum

front/lib/actions/mcp_internal_actions/constants.ts
```diff
@@ -1259,6 +1261,22 @@ export const INTERNAL_MCP_SERVERS = ensureUniqueToolNames({
...
+  shopify: {
+    id: 1044,
+    availability: "manual",
+    allowMultipleInstances: true,
+    isRestricted: ({ featureFlags }) => {
+      return !featureFlags.includes("shopify_tool");
+    },
+    isPreview: true,
+    // Pilot (phase 1): no OAuth provider yet. The merchant token is provided as
+    // a bearer token and the shop domain as an `X-Shopify-Shop` custom header.
```
aubin-tchoi: not sure I get this comment

aubin-tchoi (approved):
LGTM
let's add some tests in the bm25 harness 🙏🏻

front/lib/api/actions/servers/shopify/metadata.ts
TimotheeCrouzet: Wouldn't dropping the enum confuse the agent?

front/lib/actions/mcp_internal_actions/constants.ts
TimotheeCrouzet: it just explains why authorization is null: no Shopify OAuth provider yet (preview), so auth uses requiresBearerToken (we are using an access token from a client as bearer + shop domain via X-Shopify-Shop header). I'll Reword to say that directly.

TimotheeCrouzet:
> LGTM let's add some tests in the bm25 harness 🙏🏻

Thanks! I'll add some tests in a follow up PR.

PR #30243, opened 2026-08-10
[front] fix: compact Manage Agents actions on mobile

aubin-tchoi:
## Description

The Manage Agents buttons crowd out the search field on narrow screens and take most of the horizontal space.

This PR changes the mobile behavior so that Batch edit, Models, Tags, and Create now use compact icon buttons while preserving tooltips, accessible labels, and selected filter counters.

Only affects the mobile version.

Before 
image

After
image

## Tests

- Tested locally.

## Risk

- Low.

## Deploy Plan

- Deploy front.


ClementAupiais (approved):
The mobile experience needs a lot of rethinking but this is a good change in the meantime

PR #30247, opened 2026-08-10
Backfill agent step content run IDs

flvndvd:
## Description

This adds a dry-run-first script that backfills `dustRunId` on historical agent step contents across all message statuses. It reconstructs the association by ordering each message’s runs chronologically and matching them to agent-loop steps, skipping inconsistent data rather than guessing.

The script updates Postgres directly, so it also makes `AgentStepContentResource` compare cached `dustRunId` values with Postgres metadata. Stale Redis entries are automatically treated as cache misses and refreshed on the next read.

### DIFF
--- a/front/scripts/backfill_agent_message_consumption_analytics.ts
+++ b/front/scripts/backfill_agent_message_consumption_analytics.ts
@@ -1,7 +1,12 @@
 /**
  * Enqueue the consumption attribution + Elasticsearch indexing workflow for historical agent
  * messages. Run once in each region after the consumption analytics index and V3 analytics worker
- * have been deployed.
+ * have been deployed, and after agent step content dustRunIds have been backfilled.
+ *
+ * Before enqueueing each batch, the script classifies any run usages whose usageType is still null.
+ * It reconstructs the same billing classification as the live path from the triggering user
+ * message's persisted origin and auth method. This is required because the analytics loader rejects
+ * unclassified usage rather than risking that free, user, and programmatic consumption are mixed.
  *
  * Dry run:
  *   npx tsx scripts/backfill_agent_message_consumption_analytics.ts \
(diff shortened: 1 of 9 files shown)


front/scripts/backfill_agent_step_content_dust_run_ids.ts
```diff
@@ -0,0 +1,448 @@
...
+    return 0;
+  }
+
+  const bind: Record<string, number | string> = {
+    workspaceModelId: workspace.id,
+  };
+  const values = updates.map((update, index) => {
+    bind[`stepContentModelId${index}`] = update.stepContentModelId;
+    bind[`dustRunId${index}`] = update.dustRunId;
+    return `($stepContentModelId${index}::bigint, $dustRunId${index}::text)`;
```
aubin-tchoi, line 290: `dustRunId` is a varchar(255) in my local db instead of a text, does this still work?

aubin-tchoi (approved):
LGTM

front/scripts/backfill_agent_step_content_dust_run_ids.ts
flvndvd: yes

front/scripts/backfill_agent_message_consumption_analytics.ts
```diff
@@ -112,21 +133,117 @@ async function listAgentMessageRefs({
...
+          authMethod: triggeringUserMessage.userContextAuthMethod,
+          userMessageOrigin: origin,
+        }),
+        origin
+      ),
     };
   });
 }
 
+async function backfillMissingRunUsageTypes({
```
flvndvd, line 190: 🚗 Backfilling usage type as well, as it's required since we exclusively indexed none free usages in ES.

flvndvd: cc @fabiencelier

PR #30249, opened 2026-08-10
[front] Move status banners to a sticky full-width top bar

pinotalexandre:
## What

The outage/incident banner — and the other workspace status banners it shipped with (`SidebarBanners`) — lived at the top of the navigation sidebar, constrained to the sidebar width and stacked as narrow multi-line blocks.

They now render in a **single sticky stack pinned to the top of the app shell**, above both the sidebar and the content panel, alongside the existing subscription-end banner.

https://github.com/user-attachments/assets/34e1ab5c-c52a-40ff-b7b3-99ce13a7877e

## Changes

- New `TopBanners` ([front/components/navigation/TopBanners.tsx](https://github.com/dust-tt/dust/blob/outageBanner/front/components/navigation/TopBanners.tsx)) renders the banner stack `sticky top-0 z-50` at the top of `AppContentLayout`, and is the **single owner of `--banner-height`** (the shell sizes its scroll panes off `--panel-height = 100dvh - var(--banner-height)`). It measures the whole wrapper via a `ResizeObserver`, so the height stays correct when several banners show at once — previously `SubscriptionEndBanner` measured only itself.
- `SidebarBanners` → `StatusBanners`, removed from `NavigationSidebar`, restyled from narrow stacked blocks into full-width rows (title · description · trailing link).
- `SubscriptionPastDueBanner`'s multi-paragraph copy condensed to one line + a "Learn more" link, since it no longer has a tall narrow column to fill.
- The mobile nav trigger is offset by `top-[var(--banner-height)]` so a banner can never hide it.

## Note

Scope: all four banners in the group moved, not just the outage one — they share the same presentation and sat in the same container, so splitting them would have left half the status surface in the sidebar. Say the word if you'd rather only the outage banner moves.

## Testing

Type-check clean on the touched files; not exercised locally (needs the full local stack) — the deploy preview above is the intended review surface.

### DIFF
--- a/extension/ui/components/conversation/ConversationLayout.tsx
+++ b/extension/ui/components/conversation/ConversationLayout.tsx
@@ -1,6 +1,6 @@
 import { FileDropProvider } from "@app/components/assistant/conversation/FileUploaderContext";
 import { AgentSidebarMenu } from "@app/components/assistant/conversation/SidebarMenu";
-import { SidebarBanners } from "@app/components/navigation/AppStatusBanner";
+import { StatusBanners } from "@app/components/navigation/AppStatusBanner";
 import { SidebarContext } from "@app/components/sparkle/SidebarContext";
 import { useAuth } from "@app/lib/auth/AuthContext";
 import {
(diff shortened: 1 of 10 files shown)


ClementAupiais:
The plan was to put it full top of the page 
image

pinotalexandre:
all good? @ClementAupiais 
image

ClementAupiais (approved):
lgtm

PR #30172, opened 2026-08-07
[sparkle] Redesign ContentMessage per Figma

pinotalexandre:
## Summary
- `ContentMessage` redesigned to match the updated Figma spec — bordered card, two size recipes (compact/cozy), precise icon/typography scale — without changing the public API (`variant`/`size`/`icon`/`action` props untouched, so no product call site needs updates).
- Color roles: `primary`=gray, `success`/`green`=green, `highlight`/`blue`=blue (unchanged); `warning`/`rose`=red (blocking error — matches how `warning` is actually used across the product, e.g. `ErrorMessage.tsx`); `info`/`golden`=orange (risk/degraded capability).
- `ContentMessageInline` reuses the same compact recipe as `ContentMessage size="sm"`.
- Title downsized one type step (`heading-sm`/`heading-base` → `heading-xs`/`heading-sm`).

image

Supersedes #30123 (closed — that PR's branch was renamed/rescoped to drop the spinner + product-wide variant migration work, which will come back as separate, focused PRs).

## Test plan
- [x] Verified Color Variants, With Icon, Basic (with/without action), and Docs stories in Storybook, light + dark theme
- [x] `biome check` clean
- [x] `tsgo --noEmit` clean for the touched files

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #30250, opened 2026-08-10
[sparkle] Fix ContentMessage staying light in dark mode

pinotalexandre:
## Problem

Every `ContentMessage` / `ContentMessageInline` variant built on a raw palette class rendered near-white on the dark canvas. Most visible with `variant="primary"`: the blocked-actions banner above the input bar (`AgentInputBar.tsx:489`) and `ToolValidationCard`.

The Figma redesign (#30172) swapped the semantic tokens for raw palette classes (`bg-stone-50`, `text-red-800`, `bg-orange-50`, ...). In `tokens.css`, only the **semantic** scales are remapped inside `.dark`:

| | light (`:root`) | dark (`.dark`) |
|---|---|---|
| `--color-primary-50` | `stone-50` | `stone-950` ✅ |
| `--color-stone-50` | `stone-50` | `stone-50` ❌ |

## Fix

Dark mode is declared explicitly from the [Figma dark spec](https://www.figma.com/design/wJJMfVF6bluurSKfrEuysc/Product---WIP?node-id=13405-36130), the way `Chip.tsx` already does it: one palette family per variant, fixed role per step.

| role | colored variants | neutral |
|---|---|---|
| background | `950` | `950` |
| stroke | `900` | `800` |
| text + icon | `100` | `50` |

| variant | family |
|---|---|
| `primary`, `outline` | stone |
| `warning`, `rose` | red |
| `success`, `green` | green |
| `highlight`, `blue` | blue |
| `info`, `golden` | golden |

`outline` is the one deliberate departure: it has no fill of its own, so it keeps the lighter `stone-750` stroke. It sits on the app or panel surface (`stone-900`/`850`), against which the `stone-800` that reads as a rim on `primary`'s `stone-950` fill would vanish.

### Why not lean on the semantic scales

Two reasons, both measured:

1. Their `.dark` remap lands on different steps than the spec asks for.
2. The orange ramp is **not hue-stable**: 73.7° at step 50, 36.3° at step 950 — and `red-950` is 38.1°. So mirroring `orange-50 → orange-950` produced an `info` banner the same hue as `warning`, which is the red banner reported on this PR.

## Verification

Storybook `Feedback & Status/ContentMessage` → `ColorVariants`, sampling computed background / border / text and converting to sRGB via canvas.

- **Light untouched** — rendered each variant beside a reference element carrying `main`'s original classes: `9/9` identical, zero diffs.
- **Dark** — every background and every text color matches the spec hexes exactly.

### One Figma / token drift to sort out separately

The `highlight`/`blue` stroke is the only value not matching the spec, because the variable and the token have drifted:

| variable | Figma | our token | delta |
|---|---|---|---|
| `Blue/900` | `#063158` | `blue-900` `#07355f` | 8 RGB units |

This PR uses the token rather than hardcoding the hex, so either the Figma variable or the token needs a resync — worth a look independently of this change.

`npx tsgo --noEmit` clean in `sparkle`; `npm run format:changed` clean.

### DIFF
--- a/sparkle/src/components/ContentMessage.tsx
+++ b/sparkle/src/components/ContentMessage.tsx
@@ -24,16 +24,52 @@ const CONTENT_MESSAGE_SIZES = ["sm", "md", "lg"] as const;
 type ContentMessageSizeType = (typeof CONTENT_MESSAGE_SIZES)[number];
 
 const sharedVariantStyles = {
-  primary: "bg-stone-50 border-stone-150",
-  success: "bg-success-50 border-success-200",
-  warning: "bg-red-50 border-rose-100",
-  highlight: "bg-highlight-50 border-highlight-100",
-  info: "bg-orange-50 border-orange-100",
-  green: "bg-success-50 border-success-200",
-  blue: "bg-highlight-50 border-highlight-100",
-  rose: "bg-red-50 border-rose-100",
-  golden: "bg-orange-50 border-orange-100",
-  outline: "bg-transparent border-stone-150",
+  primary: cn(
+    "bg-stone-50 border-stone-150",
+    "dark:bg-stone-950 dark:border-stone-800"
+  ),
+  success: cn(
+    "bg-success-50 border-success-200",
+    "dark:bg-green-950 dark:border-green-900"
+  ),
+  warning: cn(
+    "bg-red-50 border-rose-100",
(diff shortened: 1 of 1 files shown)


zmarouf:
Confirmed fixed for me. Thanks!

zmarouf (approved):

PR #30251, opened 2026-08-10
[front] enh: improve Elasticsearch error handling

aubin-tchoi:
## Description

Part of https://github.com/dust-tt/tasks/issues/9977

In the issue we see a 400 from Elasticsearch. Although we have some strong suspicions on what could cause it, we do not have the actual error message and stacktrace from es.

This PR adds the error type, status and keeps the original error as a cause.

## Risk

- Low.

## Deploy Plan

- Deploy front.


davidebbo (approved):
LGTM

PR #30252, opened 2026-08-10
[MCP] Implement Shopify export tools

TimotheeCrouzet:
Implement the four Shopify export tools on top of the scaffolding: the Admin GraphQL client, cursor pagination, Zod-validated responses, and the export_products / export_customer_ltv / export_sales / export_top_customers_by_period handlers.

## Description

Implement the four Shopify export tools on top of the scaffolding from #30239: the Admin GraphQL client (per-shop endpoint, `X-Shopify-Access-Token`), cursor pagination, and Zod-validated responses. Handlers: `export_products`, `export_customer_ltv`, `export_sales`, `export_top_customers_by_period`. All read-only. Stacked on #30239, merge #30239 first.

## Tests

Manual, against a real Shopify store

## Risk

Low. Behind `shopify_tool` feature flag

## Deploy Plan

Merge after #30239.

### DIFF
--- a/front/lib/api/actions/servers/shopify/client.ts
+++ b/front/lib/api/actions/servers/shopify/client.ts
@@ -1,0 +1,214 @@
+import { MCPError } from "@app/lib/actions/mcp_errors";
+import type { ExportResult } from "@app/lib/api/actions/servers/shopify/types";
+import { PageInfoSchema } from "@app/lib/api/actions/servers/shopify/types";
+import { untrustedFetch } from "@app/lib/egress/server";
+import logger from "@app/logger/logger";
+import type { Result } from "@app/types/shared/result";
+import { Err, Ok } from "@app/types/shared/result";
+import { normalizeError } from "@app/types/shared/utils/error_utils";
+import { z } from "zod";
+
+const SHOPIFY_API_VERSION = "2026-07";
+
+// Shopify caps a connection's `first` argument at 250.
+const PAGE_SIZE = 250;
+
+// Hard cap on exported records to keep responses and API usage bounded.
+export const MAX_EXPORT_ITEMS = 1000;
+
+async function shopifyGraphQL<T extends z.ZodTypeAny>(
+  {
+    accessToken,
+    shop,
+    query,
+  }: {
+    accessToken: string;
+    shop: string;
+    query: string;
+  },
+  schema: T
+): Promise<Result<z.infer<T>, MCPError>> {
+  const url = `https://${shop}/admin/api/${SHOPIFY_API_VERSION}/graphql.json`;
+
+  let response: Awaited<ReturnType<typeof untrustedFetch>>;
+  try {
+    response = await untrustedFetch(url, {
+      method: "POST",
+      headers: {
+        "Content-Type": "application/json",
+        "X-Shopify-Access-Token": accessToken,
+      },
+      body: JSON.stringify({ query }),
+    });
+  } catch (err) {
+    return new Err(
+      new MCPError(
+        `Failed to reach the Shopify API: ${normalizeError(err).message}`
+      )
+    );
+  }
+
+  if (!response.ok) {
+    const body = await response.text();
+    if (response.status === 401 || response.status === 403) {
+      return new Err(
+        new MCPError(
+          "Shopify authentication failed. The access token may be expired or missing the required scope."
+        )
+      );
+    }
+    return new Err(
+      new MCPError(
+        `Shopify API error: ${response.status} ${response.statusText} - ${body.slice(0, 200)}`
+      )
+    );
+  }
+
+  let json: unknown;
+  try {
+    json = JSON.parse(await response.text());
+  } catch (err) {
+    return new Err(
+      new MCPError(
+        `Invalid JSON from the Shopify API: ${normalizeError(err).message}`
+      )
+    );
+  }
+
+  // Shopify returns HTTP 200 with partial `data` and an `errors` array when a
+  // field is redacted (e.g. missing scope / PCD). `data` is validated inline by
+  // the tool schema; `nullish` keeps the "data absent + errors" case parseable
+  // so we can surface the GraphQL error rather than a format error.
+  const parsed = z
+    .object({
+      data: schema.nullish(),
+      errors: z.array(z.object({ message: z.string() })).optional(),
+    })
+    .safeParse(json);
+  if (!parsed.success) {
+    logger.error(
+      { error: parsed.error.message },
+      "[Shopify MCP Server] Response validation failed"
+    );
+    return new Err(
+      new MCPError(`Invalid Shopify response format: ${parsed.error.message}`)
+    );
+  }
+
+  const { data, errors } = parsed.data;
+
+  // Only fail when `data` is absent; otherwise use what we got and log the
+  // redactions.
+  if (data === undefined || data === null) {
+    const message =
+      errors?.map((e) => e.message).join("; ") ?? "Shopify returned no data.";
+    return new Err(new MCPError(`Shopify GraphQL error: ${message}`));
+  }
+  if (errors && errors.length > 0) {
+    logger.warn(
+      { errors: errors.map((e) => e.message) },
+      "[Shopify MCP Server] Partial response with errors (likely redacted fields)"
+    );
+  }
+
+  return new Ok(data);
+}
+
+// One page of a Shopify connection, flattened to the shape `paginate` consumes.
+interface Page<N> {
+  nodes: N[];
+  hasNextPage: boolean;
+  endCursor: string | null;
+}
+
+async function paginate<N>({
+  limit,
+  runPage,
+}: {
+  limit: number;
+  runPage: (cursor: string | null) => Promise<Result<Page<N>, MCPError>>;
+}): Promise<Result<ExportResult<N>, MCPError>> {
+  const cap = Math.min(limit, MAX_EXPORT_ITEMS);
+  const nodes: N[] = [];
+  let cursor: string | null = null;
+
+  while (nodes.length < cap) {
+    const pageRes = await runPage(cursor);
+    if (pageRes.isErr()) {
+      return pageRes;
+    }
+    const page = pageRes.value;
+    nodes.push(...page.nodes);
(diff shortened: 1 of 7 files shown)


front/lib/api/actions/servers/shopify/helpers.ts
```diff
@@ -1,3 +1,589 @@
...
+  // Phase 1: the shop domain is passed as an `X-Shopify-Shop` custom header set
+  // manually by the merchant. Phase 2 (OAuth) will source it from the connection
+  // metadata instead, so this custom-header path is temporary.
+  const parsed = z
+    .object({ "X-Shopify-Shop": z.string().trim().min(1) })
+    .safeParse(authInfo?.extra?.customHeaders);
+  if (!parsed.success) {
+    return new Err(
+      new MCPError(
+        "Shopify shop domain not found. Set it as an `X-Shopify-Shop` custom header (e.g. my-store.myshopify.com)."
```
aubin-tchoi, line 17: I think you have a way somewhere to mention which headers should be passed by default in the setup

front/lib/api/actions/servers/shopify/helpers.ts
```diff
@@ -1,3 +1,589 @@
...
+      )
+    );
+  }
+  const host = parsed.data["X-Shopify-Shop"]
+    .replace(/^https?:\/\//, "")
+    .replace(/\/.*$/, "");
+  return new Ok(
+    host.endsWith(".myshopify.com") ? host : `${host}.myshopify.com`
```
aubin-tchoi: why not reject if the URL does not end in `myshopify.com`?

front/lib/api/actions/servers/shopify/helpers.ts
```diff
@@ -1,3 +1,589 @@
...
+): Promise<Result<z.infer<T>, MCPError>> {
+  const url = `https://${shop}/admin/api/${SHOPIFY_API_VERSION}/graphql.json`;
+
+  let response: Awaited<ReturnType<typeof untrustedFetch>>;
+  try {
+    response = await untrustedFetch(url, {
+      method: "POST",
+      headers: {
+        "Content-Type": "application/json",
+        "X-Shopify-Access-Token": accessToken,
```
aubin-tchoi: thought it was an API key we expected?

front/lib/api/actions/servers/shopify/helpers.ts
```diff
@@ -1,3 +1,589 @@
...
+  }
+  const host = parsed.data["X-Shopify-Shop"]
+    .replace(/^https?:\/\//, "")
+    .replace(/\/.*$/, "");
+  return new Ok(
+    host.endsWith(".myshopify.com") ? host : `${host}.myshopify.com`
+  );
+}
+
+async function shopifyGraphQL<T extends z.ZodTypeAny>(
```
aubin-tchoi: why the GraphQL api over the REST one OOC?

front/lib/api/actions/servers/shopify/helpers.ts
```diff
@@ -1,3 +1,589 @@
...
+      )
+    );
+  }
+
+  const envelope = z
+    .object({
+      data: z.unknown().optional(),
+      errors: z.array(z.object({ message: z.string() })).optional(),
+    })
+    .safeParse(json);
```
aubin-tchoi: redundant with the actual schema check

front/lib/api/actions/servers/shopify/helpers.ts
```diff
@@ -1,3 +1,589 @@
...
+    if (!page.hasNextPage || !page.endCursor) {
+      return new Ok({ nodes: nodes.slice(0, cap), truncated: false });
+    }
+    cursor = page.endCursor;
+  }
+
+  return new Ok({ nodes: nodes.slice(0, cap), truncated: true });
+}
+
+function afterArg(cursor: string | null): string {
```
aubin-tchoi: let's inline this

front/lib/api/actions/servers/shopify/helpers.ts
```diff
@@ -1,3 +1,589 @@
...
+  return cursor ? `, after: ${JSON.stringify(cursor)}` : "";
+}
+
+function queryArg(filters: string[]): string {
```
aubin-tchoi: same

front/lib/api/actions/servers/shopify/helpers.ts
```diff
@@ -1,3 +1,589 @@
...
+      return new Ok({
+        nodes: edges.map((e) => e.node),
+        hasNextPage: pageInfo.hasNextPage,
+        endCursor: pageInfo.endCursor,
+      });
+    },
+  });
+}
+
+const PRODUCT_FIELDS = `
```
aubin-tchoi: let's put those at the top of the file

front/lib/api/actions/servers/shopify/helpers.ts
```diff
@@ -1,3 +1,589 @@
...
+    return new Err(new MCPError("No Shopify access token found."));
+  }
+  const shopRes = getShopDomain(authInfo);
+  if (shopRes.isErr()) {
+    return shopRes;
+  }
+  return action(accessToken, shopRes.value);
+}
+
+export function renderExport<N>(
```
aubin-tchoi: let's split a bit the file between what is required to make queries to the API (the API client), what is used to rewrite the results for the model (rendering), other helpers like the auth, etc...

front/lib/api/actions/servers/shopify/helpers.ts
TimotheeCrouzet: Thanks, I added a shopify entry in `server_token_labels.ts`

front/lib/api/actions/servers/shopify/helpers.ts
TimotheeCrouzet: You are right, that was too permissive. I'll reject if the url if it doesn't end in `myshopify.com`

front/lib/api/actions/servers/shopify/helpers.ts
TimotheeCrouzet: It's an Admin API access token generated by the store's custom app

front/lib/api/actions/servers/shopify/helpers.ts
TimotheeCrouzet: Shopify's REST API is legacy: [link](https://shopify.dev/changelog/starting-april-2025-new-public-apps-submitted-to-shopify-app-store-must-use-graphql) here

front/lib/api/actions/servers/shopify/helpers.ts
```diff
@@ -1,3 +1,314 @@
...
+  if (!parsed.success) {
+    return new Err(
+      new MCPError(
+        "Shopify shop domain not found. Set it as an `X-Shopify-Shop` custom header (e.g. my-store.myshopify.com)."
+      )
+    );
+  }
+  const host = parsed.data["X-Shopify-Shop"]
+    .replace(/^https?:\/\//, "")
+    .replace(/\/.*$/, "");
```
aubin-tchoi: let's use the native URL node API instead of parsing the URL manually, this part captures more than the actual host

front/lib/api/actions/servers/shopify/client.ts
```diff
@@ -0,0 +1,214 @@
...
+    }
+    const page = pageRes.value;
+    nodes.push(...page.nodes);
+    if (!page.hasNextPage || !page.endCursor) {
+      return new Ok({ nodes: nodes.slice(0, cap), truncated: false });
+    }
+    cursor = page.endCursor;
+  }
+
+  return new Ok({ nodes: nodes.slice(0, cap), truncated: true });
```
aubin-tchoi: truncated is not necessarily true here, can be due to the provided limit

front/lib/api/actions/servers/shopify/client.ts
```diff
@@ -0,0 +1,214 @@
...
+  }
+
+  const { data, errors } = parsed.data;
+
+  // Only fail when `data` is absent; otherwise use what we got and log the
+  // redactions.
+  if (data === undefined || data === null) {
+    const message =
+      errors?.map((e) => e.message).join("; ") ?? "Shopify returned no data.";
+    return new Err(new MCPError(`Shopify GraphQL error: ${message}`));
```
aubin-tchoi: let's untrack this error

front/lib/api/actions/servers/shopify/client.ts
```diff
@@ -0,0 +1,214 @@
...
+      )
+    );
+  }
+
+  if (!response.ok) {
+    const body = await response.text();
+    if (response.status === 401 || response.status === 403) {
+      return new Err(
+        new MCPError(
+          "Shopify authentication failed. The access token may be expired or missing the required scope."
```
aubin-tchoi: let's untrack this error as well (basically anything from their side)

front/lib/api/actions/servers/shopify/client.ts
```diff
@@ -0,0 +1,214 @@
...
+
+const SHOPIFY_API_VERSION = "2026-07";
+
+// Shopify caps a connection's `first` argument at 250.
+const PAGE_SIZE = 250;
+
+// Hard cap on exported records to keep responses and API usage bounded.
+export const MAX_EXPORT_ITEMS = 1000;
+
+async function shopifyGraphQL<T extends z.ZodTypeAny>(
```
aubin-tchoi: not a fan of the pattern with the wrapper `withShopifyAuth`, it's slightly less clear in terms of devx
one pattern you could follow here is to have a class ShopifyClient, that stores the access token and uses it in methods that fetch certain resources. this function would be a private method of it, same for the pagination part. The class would own the outbound interactions with the API
`withShopifyAuth` would be replaced with a method that instantiates a `ShopifyClient`, which returns a result, showing it can fail if there's no auth information, or if it is invalid for instance

TimotheeCrouzet: I added a `ShopifyClient` class that owns the API calls, it's instantiated via `createShopifyClient(authInfo).`

spolu:
Taking over

PR #30255, opened 2026-08-10
fix: apply the Datadog status check to the browser logger

adrsimon:
## Description

Follow-up to #30028, which missed front-spa. The SPA aliases `@app/logger/logger` to `front/logger/datadogLogger.ts` (`front-spa/vite.config.ts`), and that module kept the old log payload type and re-exported pino's unchecked `Logger`, so anything logged from the SPA could still ship a `status` that Datadog reads as the severity.

## Tests

Added `front/logger/datadogLogger.test.ts`, mirroring the existing `logger.test.ts` type test with an extra case for `child()` bindings.

## Risk

Low.

## Deploy Plan

Regular deploy.

### DIFF
--- a/front/components/agent_builder/AgentBuilder.tsx
+++ b/front/components/agent_builder/AgentBuilder.tsx
@@ -519,7 +519,7 @@ function AgentBuilderForm({
           setPendingAgentId(data.sId);
         } else {
           datadogLogger.error(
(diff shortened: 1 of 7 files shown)


aubin-tchoi (approved):
LGTM

PR #30256, opened 2026-08-10
feat(front): replace models_picker feature flag with a global kill switch

Nils-Fedrigo:
## Description

Ships the conversation model picker to every workspace and removes the `models_picker` feature flag that used to gate it. The flag is deleted from `WHITELISTABLE_FEATURES_CONFIG` and the `@dust-tt/client` SDK union; stale per-workspace rows are harmless (they're filtered out by `isWhitelistableFeature`).

The feature is split into two concerns:

- **Baseline (now always on for everyone):** the `auto` / `auto_fast` / `auto_complex` meta-models (the Fast/Standard/Complex tiers) are available to all workspaces, and `@dust` now runs on Auto by default (unconditionally, ahead of the `dust_agent_*_default` flags). Advanced/large models keep their existing plan gating (`hasAdvancedModelAccess` / upgraded plan) — this PR does not broaden build-time model availability.
- **Wide kill switch (`global_disable_models_picker`):** a new global kill switch, exposed on the Poke Kill Switches page, that turns the *user-facing* picker off — it hides the input-bar model/effort picker, the agent-builder and editor pickers, the tiers admin UI, rejects per-message model selection on the public API, stops enforcing model access tiers, and re-surfaces the model-only global agents. Auto/meta-model availability is intentionally **not** reverted by the switch, so flipping it is safe and non-destructive.

The `auto` (Standard) stream now routes to **GPT 5.6 Luna (high)** as its first candidate, then falls back through the rest of the pool. Since `@dust` runs on Auto, it resolves to Luna-high first (dynamically, with fallback when unavailable).

Also migrates existing agents explicitly configured on Claude Sonnet 4.6 (`medium`) onto the `auto` meta-model. Note this is **not** behavior-preserving: migrated agents route through the `auto` stream (Luna-high first) rather than staying pinned to Sonnet 4.6.

## Tests

- Updated the affected unit/functional tests (`assistant`, `model_tiers/access`, `model_tiers/enabled_models`, `assistant/models`, and the public-API `messages` route) to cover the new kill-switch behavior and the always-on advanced-model baseline. All pass.
- `tsgo --noEmit` clean across `front`, `front-api`, and `sdks/js`.

## Deploy Plan

After deploy, run the migration to move existing Sonnet 4.6 (`medium`) agents onto Auto:

```
npx tsx front/migrations/20260810_migrate_sonnet_4_6_medium_to_auto.ts --execute
```

### DIFF
--- a/front-api/lib/api/assistant/conversation/model_selection.ts
+++ b/front-api/lib/api/assistant/conversation/model_selection.ts
@@ -1,6 +1,6 @@
 import type { Authenticator } from "@app/lib/auth";
-import { hasFeatureFlag } from "@app/lib/auth";
 import { getEnabledModelsForAuth } from "@app/lib/model_tiers/enabled_models";
+import { KillSwitchResource } from "@app/lib/resources/kill_switch_resource";
 import type { ModelSelectionType } from "@app/types/assistant/models/types";
 import { ModelSelectionSchema } from "@app/types/assistant/models/types";
 import type { APIErrorWithContentfulStatusCode } from "@app/types/error";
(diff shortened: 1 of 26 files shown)


Nils-Fedrigo:

- #30256 ◀

PR #30242, opened 2026-08-10
[Activation] Hide bottom citations & files in Activation Pod conversations

xebeohp:
## Description

Hide secondary bottom-of-message content in the compact Activation Pod conversation UI while preserving inline citations and keeping the standard conversation UI unchanged.

- `ConversationViewer` derives a `uiView` value, using `compact` for conversations in the Activation Pod and `standard` otherwise. Activation Pod lookup is gated behind the `activation_skill` feature flag.
- Propagate `uiView` through the conversation message-list context and `MessageItem` into `AgentMessage`.
- In compact conversations, hide the bottom citation list and generated file cards. Inline citations in the message content remain available.
- Standard conversations retain their existing bottom citation and generated file behavior.

Activation Pod compact conversation UI

## Tests

- Added `front/components/assistant/conversation/AgentMessage.compactUIView.test.tsx` with focused component tests covering:
  - Hidden bottom citations with preserved inline citations in compact conversations.
  - Retained bottom citations and inline citations in standard conversations.
  - Compact conversations with no citations.
  - Hidden generated file cards in compact conversations.
  - Retained generated file cards in standard conversations.
- Local test execution was not run in the Computer. Final validation is delegated to PR CI.

## Risk

Low. The change is scoped to secondary bottom-of-message content when `uiView` is `compact`. Inline citations remain available, standard conversations keep their existing behavior, and Activation Pod detection is disabled when the Activation skill feature flag is unavailable.

## Deploy Plan

Deploy `front`.

PR #30258, opened 2026-08-10
[Activation] Move Frame links to the bottom of isCompactUIView messages 

xebeohp:
## Description

Move generated Frame links from the top of compact Activation Pod messages into bottom citation cards, while preserving the existing behavior for standard conversations and non-Frame files.

For compact Activation Pod conversations:

- Collect generated interactive Frame files and render one bottom `FileCitationCard` per Frame.
- Open the corresponding Frame in the conversation side panel when a card is clicked.
- Suppress the top-of-message generated Frame links to avoid duplicate rendering.
- Hide inline `:preview_file` directives for Frame content types, while keeping non-Frame inline file previews available.

For standard conversations, keep generated Frame links at the top of the message and preserve existing inline preview behavior.

Activation Pod Frame citation cards

## Tests

- Expanded `front/components/assistant/conversation/AgentMessage.compactUIView.test.tsx` to cover:
  - One and multiple Frame cards rendered at the bottom for compact conversations.
  - Frame card labels, icons, and opening the correct side-panel file.
  - No Frame card when no Frame is generated.
  - Top-of-message Frame links retained for standard conversations.
  - Inline Frame previews suppressed only in compact conversations.
  - Non-Frame inline file previews retained in compact conversations.
- Updated `front/components/markdown/FilePreviewBlock.test.tsx` to verify that compact conversations hide Frame previews but still render non-Frame files.

## Risk

Low. The behavior change is scoped to Frame rendering in the compact Activation Pod UI. Standard conversations, non-Frame inline previews, accessibility labels, and side-panel navigation remain covered by the existing behavior and focused tests.

## Deploy Plan

Deploy `front`.

### DIFF
--- a/front/components/assistant/conversation/AgentMessage.tsx
+++ b/front/components/assistant/conversation/AgentMessage.tsx
@@ -1309,13 +1309,7 @@ function AgentMessageContent({
       action_card: getActionCardPlugin(onQuickReplySend, isLastMessage),
       // In the compact UI, the inline Frame preview directive is
       // suppressed here, the Frame is instead rendered as a bottom citation card.
-      ...(uiView === "compact"
-        ? {
-            file_preview: getFilePreviewPlugin({
-              hideInteractiveContent: true,
-            }),
-          }
-        : {}),
+      file_preview: getFilePreviewPlugin({ uiView }),
       ...propsAdditionalMarkdownComponents,
     }),
     [
@@ -1568,7 +1562,7 @@ function getCitations({
   });
 }
 
-// Renders Frame files as bottom citation cards for the compact Activation Pod UI
+// Renders Frame files as bottom citation cards for the compact UI
 function getFrameCitations({
   files,
   openPanel,
--- a/front/components/assistant/conversation/AgentMessage.tsx
+++ b/front/components/assistant/conversation/AgentMessage.tsx
@@ -1581,7 +1581,7 @@ function getFrameCitations({
       <FileCitationCard
         key={fileId}
         icon={ActionFrame}
-        title="Frame"
+        title={file.title}
         tooltipLabel={file.title}
         size="sm"
         onClick={() => openPanel({ type: "interactive_content", fileId })}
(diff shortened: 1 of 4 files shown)


xebeohp:

- #29996
    - #30109
        - #30242
            - #30258 ◀

front/components/assistant/conversation/AgentMessage.tsx
```diff
@@ -1043,7 +1058,13 @@ export function AgentMessage({
 
   const messageContent = (
     <ConversationMessageContent
-      citations={isDeleted || uiView === "compact" ? undefined : citations}
+      citations={
+        isDeleted
+          ? undefined
+          : uiView === "compact"
+            ? frameCitations
+            : citations
```
frankaloia, line 1066: Do we still want to show citations?  Instead of entirely removing these things, can we just put them inside a collapsable button similar to waht we do with thinking?

xebeohp: Discussed this with Clement and the point is to get rid of citations at the end of the message because the in-line citations still stay the same. We don't need the citations in two places.

frankaloia: Hmm this is one where I'm not sure it's a "compact" view thing and i't more of a design question:
1. Do we want to show the citations at the end still for normal messages if we feel that way?
2. Why show the frame citations at the bottom for compact them?

xebeohp: 1. We're treating this pod as a good experimentation space. If we get good feedback we should remove citations for all conversations.
2. For readability it makes more sense for the users to click on the frame after they have read the entire message context instead of having something unknown at the top. They still need to be there since Dust doesn't always provide the frame in-line link.

frankaloia: I think I'm a bit concerned about moving things around just from the risk of potentially confusing customers. I think in my ideal world that we can maybe just keep things stationary but behind a collapsible section or something

xebeohp: Ok thats fair, or can we just edit the skill prompt and tell it to always give the frame link in-line instead of as a card at the top of the message?

frankaloia: I would avoid making the skill prompt responsible for this as we likely want it to be deterministic. I was thinking more so just adding something like a collapsed "Generated Frames" section

PR #30259, opened 2026-08-10
fix: don't create WorkOS users on dsync deprovisioning paths

adrsimon:
## Description

Three dsync workflows have been retrying since this morning on workspace `Dr71zx5s38`:

```
group.user_removed  -> User not found with workOSId "user_01KZNHJGR6TPFSANFQYG82X501"
group.user_removed  -> User not found with workOSId "user_01KZNHJGR6TPFSANFQYG82X501"
user.deleted        -> Did not find user to delete for workOSUserId "user_01KZNHJGR6..."
```

I suspect a double emission of the event by WorkOS, which would have first removed the User, then tried to remove it but couldn't find it, creating a new user at the same time. There's no result to `SELECT * FROM users WHERE "workOSUserId" = 'user_01KZNHJGR6TPFSANFQYG82X501'` in the prod DB.

That WorkOS id was created in WorkOS at 09:55:29, *after* the three events were emitted. All three handlers resolved the WorkOS user through `fetchOrCreateWorkOSUserWithEmail`, so the helper created one and added it to the organization.

So: the second emission of the removal event provisioned the user it was supposed to remove.

- Add `fetchWorkOSUserWithEmail`, a fetch-only counterpart returning `Ok(null)` when WorkOS doesn't know the email. Both variants now share the email resolution and lookup helpers; `fetchOrCreateWorkOSUserWithEmail` behaviour is unchanged.
- Use it in `handleUserRemovedFromGroup`, `handleDeleteWorkOSUser`, and the `state === "inactive"` branch of `handleCreateOrUpdateWorkOSUser`, and no-op with a log when there's no local user, same way as `handleGroupDelete` already does.

## Tests

Typechecks clean.

## Risk

Low. 

## Deploy Plan

Normal deploy. Afterwards:
- terminate the three stuck workflows, they predate this and won't pick it up
- clean up the orphan `user_01KZNHJGR6TPFSANFQYG82X501` added to WorkOS org `Dr71zx5s38` at 09:55:30

### DIFF
--- a/front/temporal/workos_events_queue/activities.ts
+++ b/front/temporal/workos_events_queue/activities.ts
@@ -15,6 +15,7 @@ import {
 } from "@app/lib/api/workos/organization";
 import {
   fetchOrCreateWorkOSUserWithEmail,
+  fetchWorkOSUserWithEmail,
   getUserNicknameFromEmail,
 } from "@app/lib/api/workos/user";
 import {
@@ -1004,7 +1005,7 @@ async function handleUserRemovedFromGroup(
     return;
   }
 
-  const workOSUserRes = await fetchOrCreateWorkOSUserWithEmail({
+  const workOSUserRes = await fetchWorkOSUserWithEmail({
     workspace,
     workOSUser: eventData.user,
   });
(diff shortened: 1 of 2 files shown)


front/temporal/workos_events_queue/activities.ts
```diff
@@ -1013,9 +1014,20 @@ async function handleUserRemovedFromGroup(
   }
   const workOSUser = workOSUserRes.value;
 
-  const user = await UserResource.fetchByWorkOSUserId(workOSUser.id);
+  // Nothing to deprovision. Removing them from a group is a no-op.
+  const user = workOSUser
+    ? await UserResource.fetchByWorkOSUserId(workOSUser.id)
+    : null;
```
aubin-tchoi: let's still error if there's a user in workos but not in our db as it means we failed to go through
(no user in both is fine)

front/temporal/workos_events_queue/activities.ts
```diff
@@ -1140,36 +1152,55 @@ async function handleCreateOrUpdateWorkOSUser(
...
+      workOSUser: eventData,
+    });
+    if (inactiveWorkOSUserRes.isErr()) {
+      throw inactiveWorkOSUserRes.error;
+    }
+    const inactiveWorkOSUser = inactiveWorkOSUserRes.value;
+
+    const inactiveUser = inactiveWorkOSUser
+      ? await UserResource.fetchByWorkOSUserId(inactiveWorkOSUser.id)
+      : null;
```
aubin-tchoi: same here

aubin-tchoi (approved):
LGTM

front/temporal/workos_events_queue/activities.ts
aubin-tchoi: will require fixing the current user as a one off in your case since the workflow already created the user on workos but not our db

PR #30260, opened 2026-08-10
Persist reconciled agent message consumption credits

flvndvd:
## Description

We already compute a reconciled credit allocation when building consumption analytics documents and message details and in the conversation and message breakdown.

The persisted gross attribution is intentionally cache-naive and does not necessarily sum to the authoritative amount billed on the agent message. The existing reconciliation keeps output, reasoning, and tool amounts unchanged, then distributes the remaining billed credits across input items proportionally. Integer microcredit remainders are allocated deterministically.

This PR does not introduce new billing or reconciliation logic. It reuses the existing allocator and materializes its result on each consumption item as `reconciledCreditAmountMicro`. This makes the additive, authoritative breakdown directly available to downstream consumers without rebuilding it from runs, usages, actions, and attribution evidence on every read.

The existing `grossAttributedCreditAmountMicro` remains unchanged and continues to represent the immutable attribution evidence.

## Why the value is mutable

The reconciled amount cannot be treated as immutable because an agent message may be finalized more than once.

A message can continue with additional run usages after an interruption or tool confirmation. Those usages add attribution items and can change the authoritative `costCredits`. Since input items share the reconciled remainder proportionally, existing input allocations must be updated when a new usage is added (this is the worst case, good for us it does not happen that often).

A tool can first be observed while awaiting approval or authentication. At that point, we only know the output tokens spent emitting the call. If the tool is later approved and completes, its pending row gains the result footprint and direct tool charge. Its reconciled amount increases, while the amount remaining for existing input items decreases.

If the tool is denied, it completes with no direct charge. The allocation is recomputed for consistency, but the resulting amounts are normally unchanged and the update is skipped.

The authoritative agent-message cost may also change between finalizations. Even when the persisted evidence itself is unchanged, the input reconciliation must follow the latest billed amount.

Finally, an incomplete attribution version cannot be reconciled safely. Its evidence is stored with a null reconciled amount. A later finalize can complete the missing model or tool evidence and populate the allocation. Until then, the allocator can continue selecting the newest older version that is complete and self-consistent.

## Implementation

The migration adds a nullable `reconciledCreditAmountMicro` bigint column. It is nullable because evidence may exist before a complete allocation can be built and because existing rows are not synchronously backfilled by this migration.

Evidence persistence and reconciliation now happen in the same transaction. We first insert new attribution rows or complete pending tool rows, then load the full persisted message attribution and run the existing allocator against the authoritative agent-message cost.

Only a complete, self-consistent attribution version is materialized. If the message has no billed cost, has missing model buckets, has incomplete terminal tool evidence, or cannot produce a valid remainder, its reconciled values remain null.

Re-finalization no longer re-tokenizes the entire message. We rebuild only new or incomplete run usages and usages containing a pending tool that has since become terminal. Reconciliation still considers the full persisted message because existing allocations may need to move.

## Deploy Plan

- [x] Apply SQL migration
- [ ] Deploy front

### DIFF
--- a/front/lib/api/assistant/agent_message_consumption_attribution/store.ts
+++ b/front/lib/api/assistant/agent_message_consumption_attribution/store.ts
@@ -100,12 +100,12 @@ function selectRunUsagesNeedingEvidence({
 async function buildRunUsageConsumptionEvidence(
   auth: Authenticator,
   {
-    enrichedActionById,
+    enrichedActionByModelId,
     runActions,
     triggeringUserMessageOrigin,
     usage,
   }: {
-    enrichedActionById: ReadonlyMap<ModelId, AgentMCPActionWithOutputType>;
+    enrichedActionByModelId: ReadonlyMap<ModelId, AgentMCPActionWithOutputType>;
     runActions: AgentMCPActionResource[];
     triggeringUserMessageOrigin: UserMessageOrigin | null;
     usage: RunUsageWithRunKeyType;
@@ -129,7 +129,7 @@ async function buildRunUsageConsumptionEvidence(
     isSandboxChildActionInfo(action.stepContext.sandboxChildActionInfo)
   );
   const modelVisibleRunActionPairs = modelVisibleRunActions.map((action) => {
-    const enrichedAction = enrichedActionById.get(action.id);
+    const enrichedAction = enrichedActionByModelId.get(action.id);
     assert(
       enrichedAction,
       "A selected model-visible action must have an enriched counterpart"
@@ -265,7 +265,7 @@ async function buildRunUsageConsumptionEvidence(
       continue;
     }
 
-    const enrichedAction = enrichedActionById.get(action.id);
+    const enrichedAction = enrichedActionByModelId.get(action.id);
     assert(
       enrichedAction,
       "A completed sandbox child action must have an enriched counterpart"
@@ -497,7 +497,7 @@ export async function computeAndStoreAgentMessageConsumptionAttribution(
           ignoreContent: false,
         })
       : [];
-  const enrichedActionById = new Map(
+  const enrichedActionByModelId = new Map(
     enrichedActions.map((action) => [action.id, action])
   );
 
@@ -511,7 +511,7 @@ export async function computeAndStoreAgentMessageConsumptionAttribution(
     const dustRunId = dustRunIdByRunModelId.get(usage.runModelId);
     const runActions = (dustRunId && actionsByDustRunId.get(dustRunId)) || [];
     const usageEvidence = await buildRunUsageConsumptionEvidence(auth, {
-      enrichedActionById,
+      enrichedActionByModelId,
       runActions,
       triggeringUserMessageOrigin,
       usage,
--- a/front/lib/api/assistant/agent_message_consumption_attribution/store.ts
+++ b/front/lib/api/assistant/agent_message_consumption_attribution/store.ts
@@ -1,5 +1,6 @@
 import { isToolExecutionStatusFinal } from "@app/lib/actions/statuses";
 import { isSandboxChildActionInfo } from "@app/lib/actions/types";
+import { buildLatestMessageConsumptionAllocation } from "@app/lib/api/assistant/agent_message_consumption_attribution/allocation";
 import {
   AGENT_MESSAGE_CONSUMPTION_ATTRIBUTION_VERSION,
   buildRunUsageAttribution,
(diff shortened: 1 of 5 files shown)


front/lib/api/assistant/agent_message_consumption_attribution/store.ts
```diff
@@ -16,12 +17,341 @@ import type {
...
+import { withTransaction } from "@app/lib/utils/sql_utils";
 import logger from "@app/logger/logger";
+import type { AgentMCPActionWithOutputType } from "@app/types/actions";
+import type { UserMessageOrigin } from "@app/types/assistant/conversation";
 import { AGENT_MESSAGE_STATUSES_TO_TRACK } from "@app/types/assistant/conversation";
+import type { ModelId } from "@app/types/shared/model_id";
 import { assertNever } from "@app/types/shared/utils/assert_never";
 import assert from "assert";
 
+function selectRunUsagesNeedingEvidence({
```
flvndvd, line 31: Code mainly moved around to avoid having a big function.

front/lib/api/assistant/agent_message_consumption_attribution/store.ts
```diff
@@ -131,195 +466,61 @@ export async function computeAndStoreAgentMessageConsumptionAttribution(
     actionsByDustRunId.set(dustRunId, runActions);
   }
 
+  const usagesToProcess = selectRunUsagesNeedingEvidence({
```
flvndvd, line 477: That is the behavioral change deciding which usages get rebuilt and re-tokenized.

front/lib/api/assistant/agent_message_consumption_attribution/store.ts
```diff
@@ -16,12 +17,341 @@ import type {
...
+    const items =
+      await AgentMessageConsumptionItemResource.listByAgentMessageModelIds(
+        auth,
+        {
+          agentMessageModelIds: [agentMessageModelId],
+          maxAttributionVersion: AGENT_MESSAGE_CONSUMPTION_ATTRIBUTION_VERSION,
+          transaction,
+        }
+      );
+    const allocation = buildLatestMessageConsumptionAllocation({
```
flvndvd, line 344: Here the full persisted message is reconciled despite processing only selected usages.

front/lib/api/assistant/agent_message_consumption_attribution/store.ts
```diff
@@ -16,12 +17,341 @@ import type {
...
+      runs,
+      usages,
+    });
+    if (!allocation) {
+      return;
+    }
+
+    for (const [item, reconciledCreditAmountMicro] of allocation
+      .reconciledCreditAmounts.byItem) {
+      await item.setReconciledCreditAmount(reconciledCreditAmountMicro, {
```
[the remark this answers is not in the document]
flvndvd: Yes I know, but `setReconciledCreditAmounts` has been optimized to only update when the value changes.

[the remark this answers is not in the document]
flvndvd: Updated to use a raw SQL query.

front/lib/api/assistant/agent_message_consumption_attribution/store.ts
```diff
@@ -16,12 +17,349 @@ import type {
...
+      records,
+      pendingToolItems,
+      transaction,
+    });
+
+    const items =
+      await AgentMessageConsumptionItemResource.listByAgentMessageModelIds(
+        auth,
+        {
+          agentMessageModelIds: [agentMessageModelId],
```
aubin-tchoi, line 339: nit: we could add a method `listByAgentMessageModelId`

front/lib/api/assistant/agent_message_consumption_attribution/store.ts
```diff
@@ -131,195 +474,61 @@ export async function computeAndStoreAgentMessageConsumptionAttribution(
...
+      dustRunIdsToProcess.has(action.stepContent.dustRunId)
+  );
+  const enrichedActions =
+    actionsToEnrich.length > 0
+      ? await AgentMCPActionResource.enrichActionsWithOutputItems(auth, {
+          actions: actionsToEnrich,
+          ignoreContent: false,
+        })
+      : [];
+  const enrichedActionById = new Map(
```
aubin-tchoi: nit: `enrichedActionByModelId` (could also rely be a map keyed by `sId`)

front/lib/api/assistant/agent_message_consumption_attribution/store.ts
```diff
@@ -131,195 +474,61 @@ export async function computeAndStoreAgentMessageConsumptionAttribution(
...
+      .filter((dustRunId): dustRunId is string => dustRunId !== undefined)
+  );
+  const actionsToEnrich = actions.filter(
+    (action) =>
+      action.stepContent.dustRunId !== null &&
+      dustRunIdsToProcess.has(action.stepContent.dustRunId)
+  );
+  const enrichedActions =
+    actionsToEnrich.length > 0
+      ? await AgentMCPActionResource.enrichActionsWithOutputItems(auth, {
```
aubin-tchoi, line 495: we could skip the actions ran through dsbx here no?

aubin-tchoi (approved):
LGTM

avervaet (approved):

front/lib/api/assistant/agent_message_consumption_attribution/store.ts
flvndvd: No we want them, cause we spread the cost between original action (sandbox) and the one(s) run through `dsbx`.

front/lib/api/assistant/agent_message_consumption_attribution/store.ts
flvndvd: Let me rename, I need the PK ids here to do update in DB.

PR #30261, opened 2026-08-10
Make Pod function stdout delivery and fast execution the default

fontanierh:
## Description

Both latency features were behind `dust_only` flags that are now fully rolled out. Removes `sandbox_function_stdout_result` and `sandbox_function_fast_execution`, and deletes the paths they gated:

- dsbx always runs with `--result-delivery stdout`, so the callback branch in `execute()` and its exit-code error mapping are gone. A failed run is now always classified from the stdout envelope.
- A fast function always runs inline, and the invocation POST always waits (up to 10s) for an outcome before responding.

Kept on purpose: the in-sandbox callback route (`v1/w/[wId]/sandbox/sandbox-functions/result`) and dsbx's `callback` delivery mode, which is still the CLI default and the local no-token path. Note this is not a compat shim for older sandboxes: `--result-delivery` is an explicit clap arg, so a dsbx predating it rejects the whole command rather than falling back to callback.

## Tests

Reworked the invocation resource tests around stdout delivery (plain-text stdout mocks became protocol v3 envelopes) and dropped the two exit-code tests that only covered the callback path. Added one for stdout with no envelope at all.

## Risk

Low, but not zero: every invocation POST now waits for the outcome, including durable functions, which previously returned immediately when the flag was off. Blast radius is any workspace with `sandbox_functions` but without `sandbox_function_fast_execution`, all dust_only. The wait is bounded at 10s and returns early when the invocation blocks on user input.

## Deploy Plan

Standard deploy. The two flag rows left in the DB are inert and can be cleaned up later.

### DIFF
--- a/front-api/routes/w/[wId]/sandbox-functions/[functionId]/invocations.test.ts
+++ b/front-api/routes/w/[wId]/sandbox-functions/[functionId]/invocations.test.ts
@@ -437,7 +437,7 @@ describe("POST /api/w/:wId/sandbox-functions/:functionIdOrSlug/invocations", () 
     expect(body.outcome).toBeUndefined();
   });
 
-  it("returns no outcome when the invocation outlives the wait", async () => {
+  it("returns no outcome when the stream ends without settling", async () => {
     const { workspace, sandboxFunction } = await setupSandboxFunction();
 
     const response = await postInvocation({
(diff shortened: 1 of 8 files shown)


davidebbo (commented):
LGTM

PR #30262, opened 2026-08-10
[dsbx] chore: script to push @dust/pod into a running sandbox

davidebbo:
## Why

`@dust/pod` (the runtime library sandbox function code imports for `db()` and `currentUser()`) is vendored into the sandbox image at image build time by `buildPodPackage`, and published function bundles keep it as an external import (`packages: "external"`). So **neither rebuilding `dsbx` nor republishing a function picks up a change to it** — only a new image does, gated by `DUST_BASE_IMAGE_VERSION`.

That makes iterating on the package slower than it needs to be, and the reason isn't written down anywhere, so the natural moves (rebuild dsbx, republish the function) both silently do nothing.

## What

`upsert_pod_package_to_sandbox.sh`, a sibling of the existing `upsert_dsbx_to_sandbox.sh`: same flags (`--no-build`, `-h`), same `fzf` sandbox picker, same `e2b sandbox exec` + base64 push. Plus a README section explaining why the package is invisible to the other two rebuild paths.

Two details worth a look in review:

- **The build uses the same bun flags as `buildPodPackage`** (`--bundle --target=bun --packages=external`), so what lands is byte-identical to what a rebuilt image would ship. It writes only `index.js` — the `package.json` the image generates alongside it doesn't change — as `644 root:root`, matching the image's root-owned copy: importable by the workload users, not writable by them.
- **The push is followed by dropping the warm function servers.** A resident warm server holds the module it imported at startup, so replacing the file on disk alone changes nothing for it — the push looks successful while doing nothing. The script kills the servers and removes their sockets in `/home/agent-proxied/.dust-fn`, so the next invocation takes the cold path and re-imports.

Verification ends with `sha256sum -c -` against the locally computed hash, the same pattern `registry.ts` uses for the dsbx release download.

## Testing

Dev tooling only — no product code, no tests. Checked: `bash -n` passes; `--help` renders; the missing-bundle guard exits 1 before touching `e2b`; the build step produces `pod/dist/index.js`. I have **not** exercised the actual push against a live sandbox, so that's worth confirming before relying on it.

### DIFF
--- a/cli/dust-sandbox/README.md
+++ b/cli/dust-sandbox/README.md
@@ -94,3 +94,17 @@ compiling `dsbx`; `build.rs` fails early with instructions if it is missing. CI,
 the release workflow, and `upsert_dsbx_to_sandbox.sh` build it on the host
 first. Rebuild it after changing any runner source (`protocol.ts`, `invoke.ts`,
 `schema.ts`, `runner.ts`).
+
+## The `@dust/pod` runtime package
+
+`pod/` is the runtime library sandbox function code imports (`db()`,
+`currentUser()`). It is not part of `dsbx`: the image vendors it into
+`/opt/npm-global/lib/node_modules/@dust/pod` at image build time (see
+`front/lib/api/sandbox/image/pod_package.ts`), and published bundles keep it as
+an external import. So neither rebuilding `dsbx` nor republishing a function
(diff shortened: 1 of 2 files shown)


smb2268:
lgtm!

PR #30263, opened 2026-08-10
[front] enh: namespace pod databases by app

davidebbo:
## Why

A pod's databases are one flat namespace. Two apps in the same pod cannot both own a `chat`, and the `pod_functions` skill had to work around it by telling the model: *"The database name itself is Pod-wide rather than app-scoped, so pick one that will not collide with another app's."*

Published function slugs are already app-namespaced (`myapp__list-notes`). This does the same for databases.

## Approach

The app prefix is **never written in the function's source**. Source says `db("chat")`; the prefix is derived from where the code lives — the schema file's app folder when reconciling, the function's slug when invoking — and reaches the runtime as a new `DUST_POD_DATABASE_PREFIX` env var on the existing per-exec paths-env contract.

That property is the point for the storage layer: copy `MyApp/` to `MyAppCopy/`, republish, and the copy reconciles and opens its own databases with no edit to any function source. Both derivations go through one shared `deriveAppPrefix` helper (extracted from `slug.ts`) so they cannot drift apart.

**It does not make an app wholly copyable, and the skill text now says so.** A Frame addresses its functions by `<podId>/<slug>` written into its own source, so a copied folder's Frame still calls the *original* app's functions and therefore reads and writes the original's data, while the copy's functions and database sit published, reconciled and unused. Verified on a real pod: after copying `PeopleTracker` to `PeopleTracker2`, `peopletracker__people.db` held 6 rows and `peopletracker2__people.db` held 0. Repointing a copied Frame is a manual step; making Frame references app-relative would fix it properly and is not attempted here.

**Prefixing the filename (`myapp__chat.db`) rather than nesting per-app subdirectories is deliberate.** Litestream names replica subdirectories after the database *filename*, so `myapp/chat.db` and `other/chat.db` would replicate into the same `replica/chat.db/` prefix — two databases writing one replica. Keeping the directory flat also means no Rust changes, no litestream config changes, and no changes to `db_list`/`db_query`/`db_schema`: `myapp__chat` already satisfies `POD_DATABASE_NAME_REGEX`, and those tools already address databases by on-disk name.

## The transitional fallback

Databases created before this keep their bare filenames. Both resolution paths prefer the prefixed file when it exists and fall back to the bare one, so existing databases keep working untouched. Reconcile reads the live database set **inside its per-pod lock** and applies the same three-step rule, so the file it applies DDL to is always the file `db()` will open.

Known limitation, called out in the code: while the fallback stands, two apps that each reconcile a name which *already exists unprefixed* keep sharing that one legacy database — exactly as they do today. Only databases created from here on are namespaced. Retiring the fallback means renaming those files along with their litestream replica prefixes, which is data-bearing and out of scope here.

## Reviewer notes

- `db_list` reports on-disk names while the schema file and `db()` use short names. `resolvePodDatabaseName` is idempotent for an already-qualified name, so a model pasting `myapp__chat` from `db_list` into `db_reconcile` doesn't get `myapp__myapp__chat`. The skill prose and the `db_reconcile` tool description both explain the split.
- `DUST_BASE_IMAGE_VERSION` goes to **0.8.71** because `@dust/pod` is vendored into the sandbox image — the front-side prefixing is only correct against an image carrying the matching resolver. Note this branch originally set 0.8.70, but main took that tag for the group-writable-files fix while this was open, so a rebase collapsed the bump into a no-op; rebumping is what keeps the resolver and the image tag together. `DSBX_CLI_VERSION` is deliberately **not** bumped: it names a GitHub release artifact, and nothing in dsbx or the runner changed.
- **The prefix is read through `podEnv`, not `process.env`.** `@dust/pod` consults a per-invocation `AsyncLocalStorage` context first, which exists so a resident server can serve concurrent invocations from different apps without touching `process.env` — precisely where a process-wide prefix is wrong. Today `serve.ts` still applies request env to `process.env`, so a direct read would happen to work; the moment it passes `invocationEnv` the prefix would go stale while the databases dir and quota beside it stayed correct, giving a silently wrong database. Covered by three tests inside `runWithInvocationEnv` — the earlier prefix tests could not catch it, since outside a context `podEnv` falls back to `process.env` and they passed either way.
- Warm servers are also keyed per slug, so a given server's prefix is constant for its lifetime.

## Testing

- New coverage: `db_naming.test.ts` (18 tests) for the derivation/qualification/resolution rules including the copied-app case; 8 shim tests for prefix resolution and the legacy fallback; one invocation-resource test asserting the prefix actually reaches the exec env.

### DIFF
--- a/front/lib/api/actions/servers/sandbox_functions/metadata.ts
+++ b/front/lib/api/actions/servers/sandbox_functions/metadata.ts
@@ -238,8 +238,7 @@ export const SANDBOX_FUNCTIONS_TOOLS_METADATA = [
         .regex(POD_DATABASE_NAME_REGEX)
         .describe(
           "The database's short name as declared by the schema file (e.g. `chat`), without " +
-            "the app prefix: the app folder in `path` namespaces it, and the reply reports " +
-            "the resulting on-disk name."
+            "the app prefix."
         ),
       path: z
         .string()
--- a/front/lib/api/actions/servers/sandbox_functions/metadata.ts
+++ b/front/lib/api/actions/servers/sandbox_functions/metadata.ts
@@ -236,7 +236,11 @@ export const SANDBOX_FUNCTIONS_TOOLS_METADATA = [
       database: z
         .string()
         .regex(POD_DATABASE_NAME_REGEX)
-        .describe("The database name declared by the schema file."),
+        .describe(
+          "The database's short name as declared by the schema file (e.g. `chat`), without " +
+            "the app prefix: the app folder in `path` namespaces it, and the reply reports " +
+            "the resulting on-disk name."
+        ),
       path: z
         .string()
         .min(1)
--- a/front/lib/api/sandbox_functions/dsbx_db.ts
+++ b/front/lib/api/sandbox_functions/dsbx_db.ts
@@ -9,6 +9,11 @@ import {
 import { getRedisStreamClient } from "@app/lib/api/redis";
 import { ensurePodSandboxReady } from "@app/lib/api/sandbox/lifecycle";
 import { shellEscape } from "@app/lib/api/sandbox/shell";
+import {
+  podDatabasePrefixFromPodPath,
+  resolvePodDatabaseName,
+  stripPodDatabasePrefix,
+} from "@app/lib/api/sandbox_functions/db_naming";
 import type { SandboxFunctionErrorCode } from "@app/lib/api/sandbox_functions/errors";
 import { SandboxFunctionError } from "@app/lib/api/sandbox_functions/errors";
 import type { StagingHashes } from "@app/lib/api/sandbox_functions/staging_integrity";
@@ -169,6 +174,12 @@ const reconcileEnvelopeSchema = z.union([
 ]);
 
 export interface ReconcileDatabaseResult {
+  /**
+   * The on-disk database name that was reconciled: the app-relative name qualified with the app
+   * prefix (see resolvePodDatabaseName). Reported back because it is what `db_list`, `db_query` and
+   * `db_schema` address the database by, and it is not what the caller passed in.
+   */
+  database: string;
   created: boolean;
   statements: string[];
 }
@@ -215,6 +226,7 @@ async function reconcileDatabaseOnSandbox(
       );
     }
     return new Ok({
+      database,
       created: envelope.created,
       statements: envelope.statements,
     });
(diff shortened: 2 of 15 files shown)


front/lib/api/sandbox_functions/dsbx_db.ts
```diff
@@ -283,9 +313,18 @@ export async function reconcileDatabaseFromPodPath(
...
+    }
+
     return await reconcileDatabaseOnSandbox(auth, {
       space,
-      database,
+      database: resolvePodDatabaseName({
+        prefix,
+        name: appRelativeName,
+        existingNames: existingResult.value.map((entry) => entry.name),
+      }),
```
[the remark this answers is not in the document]
davidebbo: This prefix is not intended to provide security isolation. Just lightweight conflict prevention.

front/lib/api/actions/servers/sandbox_functions/metadata.ts
```diff
@@ -236,7 +236,11 @@ export const SANDBOX_FUNCTIONS_TOOLS_METADATA = [
       database: z
         .string()
         .regex(POD_DATABASE_NAME_REGEX)
-        .describe("The database name declared by the schema file."),
+        .describe(
+          "The database's short name as declared by the schema file (e.g. `chat`), without " +
```
fontanierh, line 240: kind of a complex sentence IMO

front/lib/resources/skill/code_defined/global/pod_functions.ts
```diff
@@ -93,6 +93,14 @@ skill covers creating it and moving it into the app folder with \`${FILES_MOVE_T
 
 Functions that no Frame calls still get an app folder, named after what they do together.
 
+**Copying an app folder does not finish the job.** The databases and the published slugs follow the
```
fontanierh, line 96: finishes what job ?

Not sure I understand this

front/lib/resources/skill/code_defined/global/pod_functions.ts
```diff
@@ -153,9 +161,12 @@ Functions of the same Pod can share durable SQLite databases (via \`drizzle-orm\
 - **One schema file per database** at \`<AppName>/databases/{db}.db.ts\`: the single source of truth
   declaring that database's full schema with drizzle's \`sqliteTable\` DSL. Every function imports
   its table objects from it as \`../databases/{db}.db.ts\` (never hand-write tables in a function
-  file), so functions sharing a database belong to the same app. The database name itself is
-  Pod-wide rather than app-scoped, so pick one that will not collide with another app's
-  (\`${toolName("db_list")}\` shows what the Pod already has).
+  file), so functions sharing a database belong to the same app. Name the database for what it
```
fontanierh, line 164: name the tables or databases ?

fontanierh (approved):

front/lib/resources/skill/code_defined/global/pod_functions.ts
davidebbo: It's in reference to my goal of making the folder tree copyable and end up with a fully separate app. This points out that we're not quite there after this change, as the frame client code bakes the reference to the function in one app.

But this really has nothing to do with this change, I should I have that comment out!

front/lib/resources/skill/code_defined/global/pod_functions.ts
davidebbo: Here we're strictly talking about db names. Since an app can have multiple (each with many tables), it's just saying to name them well. In simple cases with one table, we'll probably still have app name == db name == table name.

front/lib/api/actions/servers/sandbox_functions/metadata.ts
davidebbo: Yeah, I simplified it.

PR #30264, opened 2026-08-10
[front] fix: unpin sidebar nav items and blur the sticky edge

pinotalexandre:
## Description

https://github.com/user-attachments/assets/0f6ce876-f639-435d-9ff9-3934189871be

Two changes to the conversations sidebar:

1. **"For you" / "Agents" / "Skills" no longer stay pinned.** They were rendered outside the sidebar's scroll area, so they stayed put while everything below them scrolled. They now sit inside the scroll area and scroll with the rest of the list.

2. **The search bar + "New" button still stay put, and now say so.** A progressive blur is overlaid on the top edge of the scroll area, so items scrolling under the header soften instead of being cut off. It fades in only once the list is actually scrolled (IntersectionObserver on a top sentinel — same pattern as `DataTable` / `SkillBuilderFilesSection`), and fades back out at the top.

To make that possible, the `ScrollArea` moved up from `NavigationListWithInbox` into `AgentSidebarMenu`, so the nav items, the search results and the conversation list all share a single scroll viewport. `NavigationListWithInbox` now receives the viewport as a prop for its `InfiniteScroll` root.

## Deploy Plan

Standard front deploy.


ClementAupiais:
maybe the padding bottom should be a bit higher I think the gradient is to close to the search

ClementAupiais (approved):

PR #30266, opened 2026-08-10
feat(agents): run @dust on Auto and lead Standard stream with Luna high

Nils-Fedrigo:
## Description

Makes `@dust` route through the Auto (Standard) meta-model and tunes the Standard stream.

- **`@dust` runs on Auto (Standard).** Adds a `preferAutoDefaultModel` arg to the dust global-agent builder and makes it the highest-priority branch in `_getDustGlobalAgent`. It is passed from `global_agents.ts` gated on the `models_picker` feature flag, so wherever the models picker is enabled `@dust` defaults to Auto. The existing `dust_agent_sonnet_5_default` and `dust_agent_gpt_5_6_luna_default` flags remain as lower-priority fallbacks.
- **Standard stream leads with GPT 5.6 Luna (high).** `GPT_5_6_LUNA_MODEL_ID` at `high` reasoning becomes the first candidate of the `auto` stream; the rest of the pool follows unchanged as fallback.
- **Migration.** Adds `20260810_migrate_sonnet46_medium_to_auto.ts` to move active agents on `claude-sonnet-4-6` with `medium` reasoning to the Auto meta-model (provider/model `auto`, reasoning `none`).

## Tests

Type-check and biome pass on the changed files. Run the migration in dry-run first (without `--execute`) to review the affected agent count before executing.

## Deploy Plan

After deploy, run the migration to move existing Sonnet 4.6 / medium agents onto Auto:

```
npx tsx front/migrations/20260810_migrate_sonnet46_medium_to_auto.ts --execute
```

### DIFF
--- a/front/lib/api/assistant/global_agents/configurations/dust/dust.ts
+++ b/front/lib/api/assistant/global_agents/configurations/dust/dust.ts
@@ -88,8 +88,6 @@ interface DustLikeGlobalAgentArgs {
   // same model availability check that is enforced when a message is posted.
   featureFlags: WhitelistableFeature[];
   // When set, the @dust agent defaults to the Auto (Standard) meta-model. This
-  // is passed unconditionally so @dust runs on Auto for everyone, taking
-  // priority over the per-model default flags below.
   preferAutoDefaultModel?: boolean;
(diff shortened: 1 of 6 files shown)


fontanierh (approved):
🔥 🔥

PR #30268, opened 2026-08-10
Drop callback result delivery from dsbx, bump to 0.1.46

fontanierh:
## Description

Follow-up to #30261. Now that front always runs functions with `--result-delivery stdout` and `dsbx function run` has no other invoker, callback delivery is dead code. Removes `ResultDelivery::Callback`, `deliver_callback` / `deliver_callback_outcome`, and the `post_function_result` client. Bumps dsbx to 0.1.46 and dust-base to 0.8.71, so the pin ships with the change it points at. (0.1.45 went out from main while this was open, so this takes the next version.)

Two things deliberately left in place:

- `--result-delivery` still parses (accepted, ignored, `stdout` the only value). dsbx ships as a pinned release, so a binary that rejected the flag would break every invocation the moment the pin moved. It gets removed together with front's argument.
- The v1 result route is marked deprecated, not deleted. Sandboxes still running an image built below 0.1.46 carry a dsbx that defaults to callback, so a hand-run `dsbx function run` without the flag can reach it. Deleting it now would be a [BACK12] violation for no gain.

**Before the next front deploy**, dispatch `Release dsbx CLI` so `dsbx-v0.1.46` exists. Nothing builds images on merge (`sandbox-img-registry` only runs via `ensure-sandbox-images` inside the manual `Deploy` workflow), but a front deploy without the release fails at the image build's `curl`.

Once deployed sandboxes have all cycled onto 0.1.46, the leftovers can go: the deprecated route and its test, the callback arm in `normalizeSandboxFunctionResult`, and `--result-delivery` on both sides.

## Tests

`cargo test` (283), clippy with `-D warnings`, and `cargo fmt --check` all clean locally, plus the registry and result-route tests after the rebase. Also checked the local no-token path changes shape: a bare `dsbx function run` now prints the v3 envelope and exits 0 instead of the raw runner response with the runner's exit code.

## Risk

Low. No production path selected callback. The behavior change is confined to running dsbx by hand: previously the no-token case printed the bare runner response and mirrored the runner's exit code, now it prints the envelope and exits 0.

One ordering note: since `callback` is gone from the value enum, `--result-delivery callback` is a usage error rather than an ignored value. If #30261 ever needs reverting after this deploys, the revert has to move the pin back too, or every invocation fails to parse.

## Deploy Plan

Cut the dsbx release first, then a standard front deploy, which rebuilds dust-base 0.8.71 on the way through.

### DIFF
--- a/cli/dust-sandbox/Cargo.lock
+++ b/cli/dust-sandbox/Cargo.lock
@@ -313,7 +313,7 @@ checksum = "92773504d58c093f6de2459af4af33faa518c13451eb8f2b5698ed3d36e7c813"
 
 [[package]]
 name = "dust-sandbox"
-version = "0.1.45"
+version = "0.1.46"
 dependencies = [
  "anyhow",
  "base64",
--- a/cli/dust-sandbox/Cargo.toml
+++ b/cli/dust-sandbox/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "dust-sandbox"
-version = "0.1.45"
+version = "0.1.46"
 edition = "2021"
 
 [[bin]]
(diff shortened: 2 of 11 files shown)


davidebbo (approved):
LGTM

PR #30269, opened 2026-08-10
[front] fix: bound the kill-requested pre-destroy flush retry

fontanierh:
## Description

A sandbox with `killRequestedAt` set whose pre-destroy flush can never succeed is retried forever. `dangerouslyDestroyIfKillRequested` runs `ensurePodStateHealthOnSleep`, and on failure returns `Err` and skips the destroy — the row keeps its `killRequestedAt`, so the next reaper sweep does exactly the same thing.

This happened in production today. A pod database created by workload function code landed at mode `0640` owned `agent-proxied:agent`, so litestream (`dust-state`, in group `agent`) could read but not write it:

```
litestream sync of arena failed (exit 1): sync failed: sync database: db sync:
create _litestream_seq table: attempt to write a readonly database (8)
```

The sandbox never slept for three days, and once a kill was requested for the image rollout it logged `panic: true` every ~90s and could never be destroyed. It was the last sandbox blocking the rollout.

The flush is best-effort durability, not a precondition — and `dangerouslyDestroyAndRecreate` already treats it that way, logging the failed flush and proceeding with the destroy. The reaper path was the inconsistent one.

This gives the flush a one hour grace period, measured from `killRequestedAt`:

- Within the grace period, behaviour is unchanged — skip the destroy and retry on the next sweep, so a transient GCS credential or daemon hiccup still resolves on its own (the reaper sweeps every 5 minutes, so that is ~12 attempts).
- Past it, destroy anyway and log that unreplicated pod state is being dropped.

Companion PR: #30271 stops these databases from being created unwritable in the first place.

## Tests

- Two new cases in `sandbox_resource.test.ts` drive `dangerouslyDestroyIfKillRequested` with a `beforeSleep` check that always fails, and assert both sides of the deadline: within the grace period the provider destroy is never called and the row stays `running`; past it the destroy runs and the row goes `deleted`. Both fail on `main` (the second one hangs on the old always-skip path).
- `npm test lib/resources/sandbox_resource.test.ts` — 42 passed.
- `npx tsgo --noEmit` clean.
- The production scenario itself was resolved by repairing the file mode by hand, which let the existing path destroy the sandbox cleanly — so the new branch has not been exercised in prod, only in tests.

## Risk

The meaningful risk is that this deliberately converts a wedge into data loss, and that trade should be made consciously.

- Before: an unflushable sandbox pins a running VM indefinitely, but the unreplicated pod state survives on that VM. After: it is destroyed an hour later and that state is gone for good. This is the right trade for fleet health — a pinned VM blocks image rollouts fleet-wide and bills continuously — but it is strictly worse for the affected pod's data, and nothing currently notifies the owner that it happened.
- The one hour window is the main tunable. Too short and a genuine transient (expired mount credential, litestream restart loop, slow gcsfuse) could lose state that would have flushed on a later sweep; one hour is ~12 reaper sweeps, well past any transient we have observed. Too long and a wedged sandbox keeps blocking a rollout, which is the status quo.
- Scope is narrow: only the kill-requested destroy path changes. The reaper sleep path and `pauseForApproval` still refuse to pause on a failed flush, so a healthy sandbox is never paused with unflushed state.
- No schema change, no data migration, no persisted state. Rollback is a plain revert.
- The `panic: true` log is kept on both branches, so existing monitors still fire — the new branch is distinguishable by its message and the added `killRequestedForMs` field.

## Deploy Plan

Standard `front` deploy, no ordering constraints and no coordination with #30271 — the two are independent and either can ship first.

The change only takes effect for sandboxes whose kill request is already an hour old, so nothing happens at deploy time unless a wedged sandbox exists. There is none right now (today's was repaired and destroyed), so expect this to be a no-op until the next occurrence.

After deploy, the log line to watch is `pre-destroy health check still failing past the grace period` — every occurrence is a pod that lost unreplicated state and is worth following up on individually, not just as a fleet metric.

### DIFF
--- a/front/lib/resources/sandbox_resource.test.ts
+++ b/front/lib/resources/sandbox_resource.test.ts
@@ -68,7 +68,7 @@ import { createResourceTest } from "@app/tests/utils/generic_resource_tests";
 import { SandboxFactory } from "@app/tests/utils/SandboxFactory";
 import { SpaceFactory } from "@app/tests/utils/SpaceFactory";
 import type { ConversationType } from "@app/types/assistant/conversation";
-import { Ok } from "@app/types/shared/result";
+import { Err, Ok } from "@app/types/shared/result";
 import { encrypt } from "@app/types/shared/utils/encryption";
 import type { WhereOptions } from "sequelize";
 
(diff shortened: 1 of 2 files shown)


davidebbo (approved):
LGTM

PR #30270, opened 2026-08-10
Introduce Resource-owned read-through caching

flvndvd:
## Description

Our caching currently relies on conventions spread across several places:
- `cacheWithRedis` keys depend on function names.
- Resources manually convert cached JSON back into Resources.
- Transaction and invalidation behavior is repeated.
- The Poke registry reproduces cache key formats separately.
- It is unclear whether a cache returns a Resource or merely some cached data.

This PR tests a clearer contract on a deliberately small scope.

## Resource cache contract

A cached Resource lookup must always return a Resource:

Postgres → Resource
Redis → Snapshot → Resource

The snapshot is only a storage format. It must contain everything required to reconstruct a valid Resource instance and must never leak to callers.

This is expressed through:

```
defineCachedResourceLookup({
  loadFromDatabase,
  toSnapshot,
  fromSnapshot,
  // key, version, TTL...
});
```

Callers continue using the Resource entry point:

`WorkspaceResource.fetchById(workspaceId);`

They receive `WorkspaceResource | null` regardless of whether the value came from Redis or Postgres.

### Workspace Resource cache

`WorkspaceResource` now owns:
- Its database loader.
- Its cache snapshot format.
- Snapshot-to-Resource reconstruction.
- Cache policy and version.
- Invalidation on create, update, and delete.
- Its Poke cache description.

Dynamic provider kill switches are applied after reconstructing the Resource instead of being stored in the snapshot (that's the hydration part).

The common lookup handles:
- Stable cache IDs instead of function names.
- Versioned keys.
- A finite TTL.
- No caching of missing Resources.
- Postgres fallback when Redis fails.
- Immediate propagation of Postgres errors.
- Redis bypass inside transactions.
- Invalidation after transaction commit.

### Query caches

Counts and other derived values use the smaller `defineCache` helper:

```
const cache = defineCache({
  load: fetchFromDatabase,
});

await cache.get(key);
```

The loader is fixed when defining the cache, so callers cannot provide arbitrary values.

### Workspace active seats

The active-seat count moves from `MembershipResource` to a dedicated module under `lib/api/workspace_seats`.
Membership data remains the source, but “active seats for a workspace” is treated as a workspace-level query.
Membership mutations continue invalidating this cache.

### Poke catalog

Migrated cache owners now provide their operational description and key construction.

The Poke UI fetches this catalog from the server instead of rebuilding Redis keys from a shared client-side registry.

Legacy registry entries remain available during migration.

### Immutable Resource IDs

🚗 `ResourceUpdateBlob` prevents updating `sId` on `WorkspaceResource`, matching the existing invariant that Resource
string identifiers are immutable.

## Limited pilot

This PR intentionally migrates only:

- `WorkspaceResource.fetchById`
- Workspace active-seat counts
- Their Poke catalog entries

It does not yet:

- Migrate other Resource caches.
- Remove the legacy registry.
- Add automatic caching to BaseResource.
- Add global cache registration.
- Implement relocation eviction.
- Add runtime validation of cached snapshots.
- Enforce one cache per Resource.

The goal is to get feedback on the contract before applying it broadly.

## Risk

This changes the workspace cache namespace during a rolling deployment. To remain compatible with older pods, the existing v2 key stays authoritative during this rollout. New pods read v2 first, mirror values into v3, and invalidate both keys on workspace mutations.

The main temporary costs are an additional Redis write on workspace cache hits and duplicate storage across both namespaces. Existing v2 entries may contain provider lists filtered by kill switches, so v3 must not become authoritative during this rollout.

## Rollback

Rollback is safe by redeploying the previous revision. The v2 workspace key remains populated and authoritative, and the active-seat cache retains its existing physical key, so old pods can resume without a cache migration.

The mirrored v3 keys are ignored by the previous revision and expire after 15 minutes.

Removing `readFromKeyFirst` requires a separate coordinated rollout. Before making v3 authoritative, wait at least one full v3 TTL after the last legacy pod or explicitly purge v3 during the cutover.


front/lib/resources/membership_queries.ts
```diff
@@ -0,0 +1,50 @@
+import { MembershipModel } from "@app/lib/resources/storage/models/membership";
+import type { MembershipRoleType } from "@app/types/memberships";
+import type { LightWorkspaceType } from "@app/types/user";
+import type { InferAttributes, Transaction, WhereOptions } from "sequelize";
+import { Op } from "sequelize";
+
+export async function countMembershipsForWorkspace({
```
flvndvd, line 7: Not happy with this, pretty convinced that this is more of a workspace concern that should be nested in the workspace resource over a leaky method on the membership resource.

front/lib/resources/workspace_resource.ts
```diff
@@ -131,8 +132,92 @@ export class WorkspaceResource extends BaseResource<WorkspaceModel> {
     this.blob = blob;
   }
 
-  private static readonly workspaceCacheKeyResolver = (wId: string) =>
-    `workspace:v${WORKSPACE_CACHE_KEY_VERSION}:${wId}`;
+  private toCacheSnapshot(): CachedWorkspaceData {
```
aubin-tchoi, line 135: OOC won't the type returned here always be `Attributes<SequelizeModel>`? to align the db and redis paths

aubin-tchoi, line 135: so the part that can be abstracted in the base resource is mostly the contract with the signatures of the functions?
doesn't look like we can share much code here tbh

aubin-tchoi (commented):
the contract as presented in `WorkspaceResource` LGTM 👍🏻 
the membership count should be moved to `WorkspaceResource` regarless of caching IMHO

front/lib/resources/workspace_resource.ts
flvndvd: Clearly we can I thought this type included all the boilerplate FK stuff but just checked and it does not.

flvndvd: Yes, making sure that no matter if it comes from the cache or the DB it can be deserialize as a resource

aubin-tchoi (approved):

front/lib/api/resources/cached_resource_lookup.ts
```diff
@@ -0,0 +1,175 @@
...
+// fallback below.
+class ResourceDatabaseLoadError {
+  constructor(readonly cause: unknown) {}
+}
+
+/**
+ * Use when a Resource lookup returns Resource instances. For counts or other query results,
+ * use `defineCache` instead.
+ */
+export function defineCachedResourceLookup<Input, Snapshot, Resource>({
```
aubin-tchoi, line 71: I feel like the type `Snapshot` could be inferred from `Resource` since the resource specifies one Sequelize model

front/lib/resources/workspace_resource.ts
```diff
@@ -131,8 +132,98 @@ export class WorkspaceResource extends BaseResource<WorkspaceModel> {
...
+
+  private static async fetchByIdFromDatabase(
+    workspaceId: string,
+    transaction?: Transaction
+  ): Promise<WorkspaceResource | null> {
+    const workspace = await WorkspaceModel.findOne({
+      where: { sId: workspaceId },
+      transaction,
+    });
+    return workspace
```
aubin-tchoi, line 192: IIRC we need to also fetch the `whiteListedProviders` here

aubin-tchoi (approved):
LGTM

front/lib/api/resources/cached_resource_lookup.ts
flvndvd: Yeah upcoming PR to productize this.

front/lib/resources/workspace_resource.ts
flvndvd: Intentionally not in `loadFromDatabase` as it's being changed by kill switches in Poke. Hydrated by `fetchById` after cache hit.

PR #30275, opened 2026-08-10
Add team dimension to consumption analytics

flvndvd:
## Description

This PR adds the team dimension to the new consumption analytics view. 

image

Will add the plumbing to use it in the filter component in a follow up PR.


front/lib/api/analytics/consumption/labels.ts
```diff
@@ -73,6 +76,16 @@ export async function resolveDimensionLabels(
       );
     }
 
+    case "team": {
+      const groups = await GroupResource.listAllWorkspaceGroups(auth, {
+        groupKinds: [...CAP_ELIGIBLE_GROUP_KINDS],
```
avervaet, line 81: nit: We might want to allow the options to filter with as subset of groups.

avervaet (approved):

front/lib/api/analytics/consumption/labels.ts
flvndvd: Yeah see PR's description coming in a follow up PR while I get my hands on your component 🙃.

PR #30276, opened 2026-08-10
Save consumption dimension in URL

flvndvd:
## Description

Tired of loosing the dimension when sharing URL, this PR bundles the dimension in the URL, we might want to do the same later with the filters (cc @avervaet).


adrsimon (approved):

front/components/pages/workspace/AnalyticsConsumptionPage.tsx
```diff
@@ -36,14 +37,15 @@ export function AnalyticsConsumptionPage() {
   const [period, setPeriod] = useState<ConsumptionPeriodSelection>(
     DEFAULT_CONSUMPTION_PERIOD
   );
-  // Shared by the chart and the attribution table.
-  // Changing attribution table dimension switches the chart's dimension too.
-  const [dimension, setDimension] = useState<ConsumptionDimension>(
-    DEFAULT_CONSUMPTION_DIMENSION
-  );
+  const { dimension: dimensionParam } = useQueryParams(["dimension"]);
```
avervaet, line 40: nit: If I'm correct on first render `dimensionParam.value` is `undefined`, so this always starts as `"agent"` if the URL has a different dimension, a second fetch fires once the router hydrates. If fetches are costful this can add some overhead

avervaet (approved):

front/components/pages/workspace/AnalyticsConsumptionPage.tsx
flvndvd: ah yeah good call, let me merge this one and I'll fix the underlying hook as it's being used in different places.

PR #30277, opened 2026-08-10
[front] enh: support availability in public skill upserts

aubin-tchoi:
## Description

Closes https://github.com/dust-tt/tasks/issues/9972

Public API skill imports always create new skills as unpublished, and changing the availability is not possible.
This PR adds an optional `availability` field to the `POST /api/v1/w/{wId}/skills` endpoint, using the existing `editors`, `workspace_users`, and `users_and_agents` values.

## Tests

- Added tests.

## Risk

- Low.

## Deploy Plan

- Deploy front.


FlagBenett (approved):
LGTM

PR #30281, opened 2026-08-10
Add runtime agent delegation tool

frankaloia:
## Summary
- add a generic `run_agent` path that selects its child at runtime while retaining existing configured-child behavior
- expose the tool to the Dust Learning skill and cover its static schema
- reuse the existing child-conversation execution, approvals, cancellation, and result handling

## Why not default Dust
It just is not an action on the default Dust agent. It would let Dust delegate to any agent the user can access, expanding its effective tool/data surface and enabling handoffs. Unsure if this is desirable tbh (maybe - open to thoughts).

## Testing
1. Called sub agent from Dust default with learning skill
2. Validated old behavior still works

## Risk
Low

## Testing
Deploy front

### DIFF
--- a/front/lib/api/actions/servers/run_agent/index.ts
+++ b/front/lib/api/actions/servers/run_agent/index.ts
@@ -9,7 +9,6 @@ import type {
   ToolHandlerExtra,
   ToolHandlerResult,
 } from "@app/lib/actions/mcp_internal_actions/tool_definition";
-import { buildTools } from "@app/lib/actions/mcp_internal_actions/tool_definition";
 import {
   makeInternalMCPServer,
   makeMCPToolExit,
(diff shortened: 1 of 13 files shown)


front/lib/api/actions/servers/run_agent/index.ts
```diff
@@ -845,6 +844,46 @@ async function createServer(
     childAgentId = toolContext.runContext.toolConfiguration.childAgentId;
   }
 
+  if (!childAgentId) {
```
aubin-tchoi: OOC why not create a separate server, set to `auto_hidden_builder` that would re-use the shared implementation?
this version makes the behavior hard to reason and relies on the agent ID not being passed somehow, which is hard to track back to activation

aubin-tchoi (commented):
agreed that we don't want the Dust global agent to be able to call any agent, as it's very hard to pick an agent and generally not useful (less of an issue for custom agents as they are handpicked + can benefit from some specific prompting)

front/lib/api/actions/servers/run_agent/index.ts
frankaloia: Yea I went with this route because I thought it was cleaner to consolidate under one run_agent server, but I think in retrospect push back is fair and the agentID parameter override could make it more difficult to reason about

frankaloia: Addressed in 05beba3d20. The generic runtime-ID tool now lives in an explicit auto_hidden_builder agent_runner server, which Activation opts into directly. The configured-child run_agent server keeps its original single-purpose behavior; both reuse the shared execution lifecycle.

aubin-tchoi (approved):
LGTM

PR #30282, opened 2026-08-10
Start viz by default in mproc

davidebbo:
## Description

Tired of manually starting it, and it's light enough for all mprocs users to run


flvndvd (commented):
LGTM

PR #30284, opened 2026-08-10
chore: bump seat count sync activity timeouts

adrsimon:
## Description

`syncMetronomeSeatCountActivity` was running longer than its timeouts allowed on some workspaces, so it kept getting cancelled and retried. Bumped the heartbeat timeout to 2 minutes and the start-to-close timeout to 15 minutes.

## Tests

Typecheck + existing tests.

## Risk

Low. A stuck activity now holds a worker slot for up to 15 minutes instead of 10. Rollback is reverting the two values.

## Deploy Plan

Deploy `front`.


flvndvd (commented):
Can we avoid bumping to 15 minutes, every time we deploy this one activity will get killed, can't we rework a bit the workflow to have shorter activities?

PR #30289, opened 2026-08-10
[front] Show poke plugin running feedback for long runs

frankaloia:
## Summary
-Purpose is to avoid poke tools looking like they stalled mid run
- Show a Running… spinner in the poke plugin dialog while a plugin executes, with an elapsed timer that appears after 5s and updates every 5s

## Test plan
Screenshot 2026-08-10 at 12 53 10 PM

## Risk
Low

## Deploy
1.Deploy front


smb2268 (approved):
Nice, lgtm

PR #30293, opened 2026-08-10
[dsbx] feat: pod-scoped generic worker pool with app affinity

fontanierh:
## Description

The previous PR makes warm servers concurrent but keeps one per function, so warm memory still scales with the number of functions on the pod, which has no ceiling (each bun process is ~40-60MB with no page sharing; dozens of hot functions would out-eat the 2GB sandbox).

This PR replaces per-function servers with a pod-scoped pool of 4 generic workers:

- The request names the function; the worker resolves and imports the bundle lazily and caches it. Warm memory is bounded by the pool (4 x 300MB RSS cap), not by function count.
- Routing hashes the app prefix of the slug (app folders publish as `app__function`, root slugs hash as themselves), so all of one app's functions share a home worker: opening an app pays one process spawn ever, and its working set (shared lib imports included) accumulates in one module cache. The client only talks to the home slot; the cold path spawns it.
- Warm timings gain an additive `importKind` (cached|fresh) to observe affinity effectiveness.

Concurrency, admission, the ack contract and drain semantics are unchanged from the previous PR. A republish now recycles a whole worker rather than one function's server; the next PR in the stack removes that recycle entirely.

## Tests

bun: multi-function serving from one worker with importKind transitions, unknown-name refusal without dying, stamped-hash refusal before import (worker not poisoned), stale drain still lets in-flight work finish. Cargo e2e against real bun: miss, spawn at home slot, fresh then cached hits, a sibling function served by the same worker, republish, stale, miss. Affinity units: app prefixes group, root slugs and degenerate prefixes hash as themselves.

## Risk

Inert until the stack's dsbx release + image bump. Deltas vs per-function: a republish drops one worker's warmth (its app cohort) until the next PR lands, and a CPU-heavy function stalls its own app's siblings during sync bursts (accepted: isolation moved from per-function to per-app-cohort).

## Deploy Plan

Ships with the stack (see the stack's last PR).

### DIFF
--- a/cli/dust-sandbox/functions-runner/fixtures/slow-import.ts
+++ b/cli/dust-sandbox/functions-runner/fixtures/slow-import.ts
@@ -1,0 +1,9 @@
+// Top-level await delays the IMPORT itself, so tests can pin behaviors that
+// depend on the worker being inside its resolve/hash/import pipeline.
+await new Promise((resolve) => setTimeout(resolve, 400));
+
+export default {
+  async fetch(): Promise<Response> {
+    return Response.json({ imported: true });
+  },
+};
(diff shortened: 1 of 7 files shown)


davidebbo (commented):
LGTM

PR #30295, opened 2026-08-10
[front] fix: persist DCR token auth method in remote MCP server auth flow

aubin-tchoi:
## Description

Tentative fix for https://github.com/dust-tt/tasks/issues/9989.

We see issues with some remote MCP servers systematically failing to refresh OAuth tokens. Coincidentally, remote MCP OAuth discovery could persist `client_secret_post` only because the authorization server advertised it, even when dynamic client registration returned a different authentication method.

This PR uses the MCP SDK selection logic so the DCR-returned method wins when it is valid and supported, with capability and client-secret based fallback behavior preserved.

## Tests

- Added coverage for various DCR responses.

## Risk

- Low.

## Deploy Plan

- Deploy front.


aubin-tchoi:
PMRR

adrsimon (commented):
LGTM

PR #30305, opened 2026-08-11
Fix consumption analytics backfills

flvndvd:
## Description

This PR makes the agent step-content backfill use bounded, index-backed pagination without requiring a new database index.

It also ensures deleted conversations remain included in consumption analytics because their credits were still consumed. The analytics workflow explicitly loads deleted conversations and indexes their historical usage.


front/lib/analytics/agent_message_consumption/load.ts
```diff
@@ -180,6 +180,20 @@ export async function loadAgentMessageConsumptionAnalyticsInput(
   }
 
   const { agentMessage, conversation, triggeringUserMessage } = context;
+  // Deleted conversations still incurred billable consumption and must remain visible in
+  // historical analytics. This system workflow is already workspace-scoped and loads the message
+  // graph without user permission filtering, so load the conversation under the same conditions.
+  const messageConversation = await ConversationResource.fetchById(
```
flvndvd, line 186: Required to fix a bunch of stucked workflows in production (cc @adrsimon).

aubin-tchoi (approved):
LGTM

front/lib/analytics/agent_message_consumption/load.ts
```diff
@@ -180,6 +180,20 @@ export async function loadAgentMessageConsumptionAnalyticsInput(
...
+    auth,
```
avervaet, line 187: Could we propagate includeDeleted through ancestor traversal as well? 

Parents are retrieved by ConversationResource.fetchByModelIds, which always excludes visibility = "deleted" conversations.

If a parent conversation was also deleted, ancestor traversal stops and the analytics document loses parent_ids, direct_parent_id, and potentially root_id, despite the usage itself still being indexed.

flvndvd: Good catch 🙏 ! Let me send you a fix with some tests.

flvndvd: Fixed in https://github.com/dust-tt/dust/pull/30310.

PR #30308, opened 2026-08-11
[marketing] feat: make the landing-page skip on dust.tt/ opt-in

davidebbo:
## Description

Today `dust.tt/` redirects **every** logged-in visitor straight to the app — the marketing homepage is unreachable for them unless they know about `dust.tt/home`. This flips the default: logged-in visitors get the landing page, and opt into the redirect themselves.

The opt-in is a prompt anchored under the existing "Open Dust" button:

```
┌─────────────────────────────────┐
│  Skip this page next time?   ×  │
│  Always open Dust  →            │
└─────────────────────────────────┘
```

"Always open Dust" sets the preference *and* opens the app. `×` dismisses the prompt for good without opting in. The prompt only shows on `/`, since that is the only page that redirects.

## Implementation

**`lib/cookies.ts`** — two new cookies alongside the existing ones: `dust-skip-landing` (the opt-in, read server-side) and `dust-skip-landing-prompt-dismissed` (client-only). Both `path=/`, `SameSite=Lax`, 1-year max-age, and deliberately **host-only** so opting in on `dust.tt` does not silently opt the visitor in on `eu.dust.tt`. The three predicates delegate to a shared `isCookieFlagSet`, which `hasSessionIndicator` now uses too.

**`pages/index.tsx`** — the redirect is gated on the preference:

```ts
if (
  shouldSkipLanding(context.req.cookies[DUST_SKIP_LANDING]) &&
  hasWorkosSessionCookie(cookieHeader)
) {
```

The cookie lookup comes first on purpose: it is a plain map read, so logged-in visitors who have not opted in no longer pay for the blocking `/api/auth-context` round-trip (1500ms timeout) on every SSR of the root. Everything downstream is unchanged — UTM forwarding, the `/api/login` fallback when there is no default workspace, and the "auth-context failed → render the landing page" behaviour.

**`hooks/useOpenDustTarget.ts`** (new) — extracts the `dust-has-session` read, the auth-context lookup, and the destination resolution that `OpenDustButton` owned, so the button and the prompt cannot drift on the `/api/login` fallback. SWR already dedupes the underlying request, so using the hook twice on a page costs no extra network call.

**`components/home/SkipLandingPrompt.tsx`** (new) — renders only when the visitor is authenticated and neither cookie is set. Cookie reads happen in a `useEffect` (matching the existing hydration-mismatch handling in `LandingLayout` and `OpenDustButton`), so the card is never part of the server-rendered HTML. Both actions are tracked via the existing `withTracking` helper (`always_open_dust`, `dismiss_skip_landing_prompt`). Hidden below the `xs` breakpoint, consistent with how "Contact sales" is handled in the same cluster.

**`components/home/LandingLayout.tsx`** — the CTA cluster gets `relative` so the card anchors under the button and tracks the header as it shrinks on scroll.

These are functional preference cookies set by an explicit user action, so they are not gated on the consent banner — same reasoning that applies to `dust-has-session` today.

## Risk

- **No undo.** Once opted in, `dust.tt/` redirects for good on that browser. `dust.tt/home` still serves the same landing page and the logo links there, so the content is never unreachable, but there is no in-product way to clear the preference. This was a deliberate call to keep the change small.
- **Logged-in visitors get an extra click** on `dust.tt/` until they opt in. That is the intended trade-off.
- **Marketing analytics will move.** Logged-in visitors currently never render `/`, so GTM/PostHog never fire for them there. After this they will — landing pageviews go up and the funnel mix shifts until opt-in rates settle. Worth a heads-up to whoever watches those dashboards so the jump is not read as a data bug.

## Tests

`marketing` has no test runner (no vitest/jest, no test files), so this is verified by `npm run tsgo`, `npx biome check`, and `npm run build` — all clean.

### DIFF
--- a/marketing/components/home/LandingLayout.tsx
+++ b/marketing/components/home/LandingLayout.tsx
@@ -11,6 +11,7 @@ import { OpenDustButton } from "@marketing/components/home/OpenDustButton";
 import { PromoBanner } from "@marketing/components/home/PromoBanner";
 import { PublicWebsiteLogo } from "@marketing/components/home/PublicWebsiteLogo";
 import ScrollingHeader from "@marketing/components/home/ScrollingHeader";
+import { SkipLandingPrompt } from "@marketing/components/home/SkipLandingPrompt";
 import UTMButton from "@marketing/components/UTMButton";
 import { useStripUtmParams } from "@marketing/hooks/useStripUtmParams";
 import {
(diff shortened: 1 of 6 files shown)


fontanierh (approved):
LGTM

PR #30312, opened 2026-08-11
feat(poke): show region in global feature flag toggle and preserve description line breaks

Nils-Fedrigo:
## Description

Small improvements to the Poke "Toggle Global Feature Flag" plugin dialog:

- The plugin description now shows the **current region** (via `getRegionDisplay(config.getCurrentRegion())`) and includes a reminder to apply the change across all regions, so operators don't forget global flags are region-scoped.
- `RunPluginDialog` now renders plugin descriptions with `whitespace-pre-line`, so the newline in the updated description (and any other multi-line plugin description) is preserved instead of collapsing to a single line.

## Tests

Manually verified in Poke: the toggle-global-feature-flag dialog displays the current region and the multi-line description renders with line breaks preserved.


davidebbo (approved):
LGTM. Thanks!

PR #30313, opened 2026-08-11
[front] fix: pass down `offline_access` if requested in remote MCP OAuth flow

aubin-tchoi:
## Description

2nd tentative fix for https://github.com/dust-tt/tasks/issues/9989

Some remote MCP authorization servers only issue refresh tokens when the client requests `offline_access` (Ashby for instance). Our dynamic OAuth flow advertises the refresh-token grant, but did not request that scope when the server supported it.

When authorization-server metadata advertises `offline_access`, this PR adds it to the selected scopes and sends the same scope in both the authorization request and authorization-code token exchange. The authorization server may still choose not to issue a refresh token, and existing connections are unchanged until they reconnect.

## Tests

- Added tests.

## Risk

- Low.

## Deploy Plan

- Deploy front.
- Deploy oauth.

### DIFF
--- a/core/src/oauth/providers/mcp.rs
+++ b/core/src/oauth/providers/mcp.rs
@@ -279,6 +279,10 @@ impl Provider for MCPConnectionProvider {
             form_data.push(("resource", resource));
         }
 
+        if let Some(ref scope) = metadata.scope {
+            form_data.push(("scope", scope));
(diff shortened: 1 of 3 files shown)


adrsimon (approved):
looks fair enough

PR #30316, opened 2026-08-11
Pricing: allow to override plan seat limits per workspace

fabiencelier:
## Description

related to https://github.com/dust-tt/tasks/issues/9961

The purpose is to avoid creating a lot of custom plans which are hard to maintain just to bypass one limit of the plan.
The main reason to create custom plans was the max number of seats and free seats.

Limits override are store in a new table and injected into the plan so every reader correctly use the overriden limits.
Note: -1 means unlimited, same as in the actual plans.

Add a poke plugin to set these overrides

image

The overrides are displayed in the plan description:

image

## Tests

tested locally + unit tests

## Risk

low

## Deploy Plan

lock front
merge
apply migration 
deploy front
unlock front


fontanierh (approved):
LGTM

PR #30321, opened 2026-08-11
Free usage: rate limit free usage to 5$ per day per user

fabiencelier:
## Description

Track the free usage (sidekick, utility call like title generation...) into a 24h rate limiter.
Throw before running the LLM calls if the limit is reach.
Add a feature flag to escape this check entirely if needed for some workspaces

## Tests

tested locally:
 - the limiter is correctly incremented.
 - when reached we are stuck

image

image

## Risk

low, may block some legit very advanced users but 
 - it's only 24h
 - there is a feature flag to escape if really needed

## Deploy Plan

deploy front

### DIFF
--- a/front/lib/api/llm/free_usage.ts
+++ b/front/lib/api/llm/free_usage.ts
@@ -1,8 +1,29 @@
-import { isFreeUsageCostLimitReachedForUser } from "@app/lib/api/assistant/rate_limits";
 import type { LLMTraceContext } from "@app/lib/api/llm/traces/types";
 import type { Authenticator } from "@app/lib/auth";
 import { getFeatureFlags } from "@app/lib/auth";
-import { isFreeOrigin } from "@app/lib/metronome/events";
+import {
+  awuFromMicroUsd,
+  isFreeOrigin,
+} from "@app/lib/credits/agent_message_billing";
+import {
+  addRateLimiterCount,
+  getRateLimiterCount,
+} from "@app/lib/utils/rate_limiter";
+import logger from "@app/logger/logger";
+import type { LightWorkspaceType } from "@app/types/user";
+
+// Per-user cost cap on free (unbilled) LLM usage — utility calls (title/skill
+// suggestions, etc.) and free agent calls (sidekick). Counted in AWU credits
+// (the fair-use unit; 1 AWU = $0.0085), so $5/day ≈ 589 credits.
+const FREE_USAGE_COST_WINDOW_SECONDS = 24 * 60 * 60;
+const FREE_USAGE_AWU_CREDITS_LIMIT_PER_DAY = awuFromMicroUsd(5 * 1_000_000);
+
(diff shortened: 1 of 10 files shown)


front/lib/api/llm/free_usage.ts
```diff
@@ -0,0 +1,14 @@
...
+
+// Whether an LLM call is free (unbilled). Two cases:
+//   - Utility operations (title/skill suggestions, etc.) — anything other than
+//     the agent conversation itself. They carry no userMessageOrigin, so they
+//     are keyed off operationType.
+//   - Agent conversations triggered by a free origin (e.g. sidekick).
+export function isFreeUsageContext(context: LLMTraceContext): boolean {
+  return (
+    context.operationType !== "agent_conversation" ||
+    isFreeOrigin(context.userMessageOrigin ?? null)
```
flvndvd, line 38: Not a big fan of adding another layer on top of the `isFreeOrigin`, are we sure that origin does not allow us to capture the case for `context.operationType !== "agent_conversation"`?

front/lib/api/llm/llm.ts
```diff
@@ -141,6 +147,42 @@ export abstract class LLM<
...
       return;
     }
+
+    // Free (unbilled) usage — utility operations and free-origin agent calls
+    // (e.g. sidekick) — is metered per user per day: read the counter before the
+    // call, contribute the call's cost after it completes (see below). Only
+    // authenticated users are metered, and the `skip_free_usage_rate_limit`
+    // feature flag exempts a workspace entirely (escape hatch to unstick legit
+    // heavy free usage).
+    const isFreeUsage = isFreeUsageContext(this.context);
```
flvndvd, line 153: IMHO this should live outside of our llm router, it should upper in the stack (if such a place exist)

front/lib/api/llm/free_usage.ts
fabiencelier: Origin doesn't capture the utility functions case, the two branches classify two different things. `isFreeOrigin(userMessageOrigin)` only ever fires for agent_conversation calls, because userMessageOrigin is only set there.
Operation type contains the information which let us know if it is a utility function call or not (`"agent_builder_name_suggestion"`, ... )

front/lib/api/llm/llm.ts
fabiencelier: moved it out. `getStreamLLM` only has 3 call sites (`runMultiActionsAgent`, the agent loop's `run_model`, and reinforcement), so the enforcement now lives there: a shared `isFreeUsageBlocked(auth, context)` guards each site calls 

I kept only the cost contribution in the router, right next to where it records the `run_usages` row

front/lib/api/llm/free_usage.ts
flvndvd: I see, sad cause original could be a more broader concept here.

flvndvd (approved):

PR #30323, opened 2026-08-11
Admin governance: allow to list members of provisioned groups in admin page

fabiencelier:
## Description

Before that, only manual groups were clickable to display the list of members.
This adds a similar pop-up for provisioned group with a message that members cannot be managed from here.

image

## Tests

tested locally

https://github.com/user-attachments/assets/1b0856ea-b7d4-4392-9370-b9c830cfac0c

also add unit tests to the endpoint

## Risk

low

## Deploy Plan

deploy front


frankaloia (approved):
Looks clean to me

PR #30325, opened 2026-08-11
[front] Remove the fake outage banner that shipped with #30249

pinotalexandre:
## Urgent — user-visible

The demo hook I added to review [#30249](https://github.com/dust-tt/dust/pull/30249) on the deploy preview was merged by mistake. It was meant to be reverted before merge and wasn't.

`useDemoAppStatus` in [AppStatusBanner.tsx](https://github.com/dust-tt/dust/blob/main/front/components/navigation/AppStatusBanner.tsx) returns a fabricated incident **unconditionally**, and it takes precedence over the real data:

```ts
const appStatus = demoAppStatus ?? liveAppStatus;
```

Two consequences on `main`:

1. Every user sees a banner reading "Degraded performance on conversations" — a fake incident.
2. A **genuine** StatusPage incident created via the runbook would be hidden behind the fake one, so the incident-communication path is broken.

## Fix

Removes the demo hook and its wiring, plus the two imports it was the only consumer of (`useAppRouter`, `useMemo`). `StatusBanners` reads `useAppStatus()` directly again, exactly as before #30249.

This is a straight revert of the demo code — no other behaviour from #30249 is touched (the banner keeps its full-width placement at the top of the shell, its orange colors, and its spacing).

## Verification

- No `useDemoAppStatus` / `demoBanner` / `DEMO` reference left in the file.
- `tsgo` clean on `front/components/navigation`.
- After merge, the banner only renders when `/api/app-status` reports an unresolved StatusPage incident.

### DIFF
--- a/front/components/navigation/AppStatusBanner.tsx
+++ b/front/components/navigation/AppStatusBanner.tsx
@@ -5,7 +5,6 @@ import {
   isDustCompanyPlan,
   isEnterprisePlanPrefix,
 } from "@app/lib/plans/plan_codes";
-import { useAppRouter } from "@app/lib/platform";
 import { useProgrammaticUsageLimit } from "@app/lib/swr/usage_settings";
 import { useAppStatus } from "@app/lib/swr/useAppStatus";
 import { useWorkspaceUsageStatus } from "@app/lib/swr/user";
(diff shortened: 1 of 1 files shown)


aubin-tchoi (approved):
LGTM

PR #30326, opened 2026-08-11
[front] fix: steer fast-function tool-call 403 to a retry, add stable error type

fontanierh:
## Description

When a fast Pod function calls a tool, the platform self-heals: it records the function as durable before building the 403. The refusal copy still said "Publish it with executionMode `durable`", which sends agents into a republish cycle when simply retrying the invocation already works. This rewords the message to say retrying works (republish only to make the source declaration match) and gives the refusal a stable error type, `fast_function_called_tools`, instead of the generic `invalid_request_error`.

The "published as fast" phrase is kept verbatim: live frames string-match it to classify the refusal.

## Tests

Extended the existing 403 test to pin the new error type, the "published as fast" phrase, and the retry wording.

## Risk

Low. Message and error-type change on one refusal path; status code and self-heal behavior unchanged.

## Deploy Plan

Standard deploy.


davidebbo (commented):
LGTM. Nice!

PR #30329, opened 2026-08-11
[dsbx] feat: tools.call() typed async tool client in @dust/pod

fontanierh:
## Description

Pod functions that call workspace tools each vendor a ~90-line sync `spawnSync("dsbx")` shim: busy-wait backoff, stderr regexing, temp-file `__file__:` args, and in a resident worker the sync spawn blocks the single JS thread serving every concurrent invocation.

This adds `pod/tools.ts` to `@dust/pod`:

- `await tools.call(server, tool, args)` POSTs `sandbox/actions/call` and polls `sandbox/actions/{aId}`, mirroring the dsbx CLI client's cadence (500 ms interval, 10 min cap, bounded backoff on transient network errors while polling).
- Args are sent as a real JSON object: no argv limits, no scalar coercion, no `__file__:` indirection.
- Returns `{content, structuredContent?, isError}` with `text()`/`json()` helpers. Tool-level errors resolve with `isError: true`; transport failures throw a typed `ToolCallError{code, retryable}` (401 is terminal since tokens are minted once per invocation, POST network failures are marked non-retryable since the action may already exist).
- Server-name -> serverViewId resolution via the existing `GET sandbox/actions` listing, cached per invocation token.

Bumps `POD_PACKAGE_VERSION` to 0.2.0. Another in-flight PR also bumps it, whichever lands second may need a trivial rebase.

Follow-up (separate PR): resolve offloaded archive pointers inside `text()`/`json()` once the `resolveToolTextContent` helper lands; until then blocks are returned as delivered.

## Tests

`bun test` unit tests with a mocked fetch covering the wire shape (args verbatim, auth header), pending -> success polling, rejected/errored actions, HTTP error classification per status, network-error retry while polling, per-invocation caching (including no caching of failures and invocation-context env isolation), timeout, and the `text()`/`json()` helpers.

## Risk

Low, purely additive: a new module in `@dust/pod` plus a version bump. Existing shims keep shelling out to the same endpoints. Long-lived pods only see the new export after an image refresh.

## Deploy Plan

Standard deploy. Ships to sandboxes with the next image rebuild.

### DIFF
--- a/cli/dust-sandbox/pod/index.ts
+++ b/cli/dust-sandbox/pod/index.ts
@@ -7,8 +7,10 @@
  * - `podEnv(name)`: the invocation's environment (./context.ts).
  * - `resolveToolTextContent(block)`: full text of a tool output block,
  *   resolving offloaded content through its descriptor (./tool_output.ts).
+ * - `tools.call(server, tool, args)`: workspace tool calls (./tools.ts).
  */
 export * from "./context.ts";
 export * from "./db.ts";
 export * from "./identity.ts";
 export * from "./tool_output.ts";
+export * from "./tools.ts";
--- a/cli/dust-sandbox/pod/package.json
+++ b/cli/dust-sandbox/pod/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@dust/pod",
-  "version": "0.1.1",
+  "version": "0.2.0",
   "private": true,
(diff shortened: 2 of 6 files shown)


davidebbo (commented):
LGTM though admittedly I didn't look at tools.ts which gh refused to show me.

PR #30333, opened 2026-08-11
[front] feat: add output_truncated and output_too_large runner error codes

fontanierh:
## Description

Two new stable codes in `SANDBOX_FUNCTION_RUNNER_ERROR_CODES` (and the runner protocol mirror, which is pinned to it by test):

- `output_truncated`: the runner's stdout envelope was cut mid-JSON in transit, so the function ran but its result was lost.
- `output_too_large`: the serialized result exceeds the hard size cap.

Front accepts and preserves these codes through result normalization today; the dsbx-side emitters ship separately and must not reach an image before this is deployed.

## Tests

Normalization tests for both codes in `result_envelope.test.ts` / `result_delivery.test.ts`.

## Risk

Low: additive enum entries, no behavior change until dsbx emits them.

## Deploy Plan

Standard deploy. Must be live before the dsbx release that emits these codes bumps `DSBX_CLI_VERSION`.

### DIFF
--- a/cli/dust-sandbox/functions-runner/protocol.ts
+++ b/cli/dust-sandbox/functions-runner/protocol.ts
@@ -27,6 +27,14 @@ export const RUNNER_ERROR_CODES = [
   // invoke(): the function is at its concurrency limit and the invocation
   // was refused before anything executed.
   "overloaded",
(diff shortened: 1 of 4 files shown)


davidebbo (commented):
LGTM

PR #30334, opened 2026-08-11
[front] feat: add Apps tab to Pods

davidebbo:
## Description

A Pod app — a Frame plus the functions and databases behind it — had no first-class presence in the UI. There was no way to enumerate a Pod's apps: you had to dig through old conversations to find them. This adds an **Apps tab**.

An app stays a **derived** concept rather than becoming a new record. It is a folder at the pod root, identified by the app prefix [`deriveAppPrefix`](https://github.com/dust-tt/dust/blob/main/front/lib/api/sandbox_functions/slug.ts) already computes from the folder name — the very prefix that namespaces published function slugs and database filenames since #30263. The tab is simply a third consumer of that one authority, so there is no schema change, no migration, and nothing new for agents to declare.

`listPodApps` joins three sources on that prefix in a constant number of calls:

| Source | Call |
|---|---|
| Pod files | one `dustFs.list()` (pod listings are recursive, so one call covers every depth) |
| Published functions | `SandboxFunctionResource.listBySpace` |
| Databases | one delimited GCS list of the litestream replica prefix |

Reading databases from the replica rather than from the sandbox means listing apps never has to wake a sleeping pod. A delimited list is used so the thousands of LTX objects under the prefix are never enumerated.

A folder qualifies as an app when it holds a `functions/` or `databases/` subfolder, a Frame at its top, or any published artifact under its prefix — so half-built apps show up too, which are exactly the ones currently lost in old conversations.

## Pre-existing ambiguities this surfaces

Rather than hiding these, the tab reports them:

- **Colliding folders.** `Task List` and `Task-List` both normalize to `task-list`, so they genuinely share published slugs and databases today. They are shown as one app with a warning instead of silently rendering two apps that own the same artifacts.
- **Pod-root artifacts.** Functions published outside any app folder have no prefix, so they are gathered into a synthetic "unfiled" app instead of being invisible.
- **Legacy databases.** A database created before app namespacing has a bare filename (`products`, not `productmanager__products`), so its name cannot say who owns it. These are attributed via the schema file that declares them (`ProductManager/databases/products.db.ts`), mirroring how `resolvePodDatabaseName` resolves the same case at reconcile time. Two apps declaring the same bare name are both credited, since they really do share that one file.

## Display

Functions and databases show their **bare** names (`add-person`, not `peopletracker2__add-person`) since the app is already the surrounding context. The addressable full slug and on-disk name stay on the wire for anything that needs to reference them. Frames open in the existing `PodFrameSheet`, so sharing, pinning and export come along for free.

The tab is gated on the `sandbox_functions` feature flag.

## Notes for reviewers

- `includeConnectedData` was generalized into a `PodNavVisibility` object. Without it, frame-tab neighbour-swapping could move a tab past an Apps tab the user cannot see.
- `FileStorage.listSubdirectoryNames` is new: a delimited GCS list returning only directory names.
- Existing pods with a stored `tabsOrder` will get Apps at the far right, since `normalizeTabsOrder` appends missing system tabs. New pods get it right after Files. Happy to insert positionally instead if preferred.
- `fileCount` is on the payload but unused by the UI; the delete confirmation in the follow-up needs it.

## Risk

Low. Additive and read-only — one new GET endpoint and one new tab. No writes, no schema change.

## Tests

8 new functional tests on the endpoint covering the happy path, non-app folders, the unfiled app, colliding prefixes, an app whose folder is gone but whose functions are still published, and both legacy-database cases. Unit tests for the two new naming helpers.

### DIFF
--- a/front/lib/api/projects/apps.ts
+++ b/front/lib/api/projects/apps.ts
@@ -1,0 +1,462 @@
+import { DustFileSystem } from "@app/lib/api/file_system";
+import { SCOPED_PREFIX_POD } from "@app/lib/api/file_system/types";
+import { getPodStateBasePath } from "@app/lib/api/files/mount_path";
+import {
+  appPrefixFromPodDatabaseName,
+  podDatabaseNameWithoutAppPrefix,
+} from "@app/lib/api/sandbox_functions/db_naming";
+import {
+  normalizeAppPrefix,
+  SANDBOX_FUNCTION_SLUG_SEPARATOR,
+} from "@app/lib/api/sandbox_functions/slug";
+import type { Authenticator } from "@app/lib/auth";
+import { getPrivateUploadBucket } from "@app/lib/file_storage";
+import { FileResource } from "@app/lib/resources/file_resource";
+import { ProjectMetadataResource } from "@app/lib/resources/project_metadata_resource";
+import { SandboxFunctionResource } from "@app/lib/resources/sandbox_function_resource";
+import type { SpaceResource } from "@app/lib/resources/space_resource";
+import type {
+  FileSystemEntry,
+  FileSystemFileEntry,
+} from "@app/types/api/file_system/types";
+import type {
+  PodApp,
+  PodAppDatabase,
+  PodAppFrame,
+  PodAppFunction,
+} from "@app/types/api/pod_apps";
+import { isInteractiveContentType } from "@app/types/files";
+import type { Result } from "@app/types/shared/result";
+import { Err, Ok } from "@app/types/shared/result";
+
+/**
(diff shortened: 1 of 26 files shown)


front/lib/api/projects/apps.ts
```diff
@@ -0,0 +1,462 @@
+import { DustFileSystem } from "@app/lib/api/file_system";
```
davidebbo, line 1: There is quite a bit of logic here, because on the inside we don't have a true App concept. So we need to 'sniff' things in various ways. It's more of a heuristic.

fontanierh (approved):
LGTM (low risk)

PR #30337, opened 2026-08-11
[front] enh: echo the serving version from pod function publish, list, get and inspect

fontanierh:
## Description

Agents cannot confirm a publish landed or which bundle version serves: the publish tool names only the slug, `list`/`get` show no timestamp or hash, and invocations do not record which bundle they ran against. That bred "smells like a stale publish" hunts even though serving is content-addressed.

- `publish` now returns a receipt: executionMode, updatedAt, short bundle sha, plus a "byte-identical to the previous publish" flag when a re-publish produced the exact same bundle (the publisher's edit did not land). The comparison happens in `updateContent`, under the publish lock, against the row's stored hash.
- `list` shows executionMode and updatedAt per function; `get` shows updatedAt and the full bundle sha.
- `execute()` stamps the bundle sha it runs against; the terminal blob records it and `inspect_invocations` (via `toJSONForLLM`) reports it, so invocations can be matched to publishes.

The stamp is absent for invocations settled through the in-sandbox HTTP callback (a separately fetched instance settles those) and for blobs written before the field existed; the field is optional everywhere.

Stacked on #30326.

## Tests

New cases: byte-identical republish reported at both the lib and the tool level, first publish never byte-identical, pre-hash rows (null sha) never match, executed invocation's blob and `toJSONForLLM` carry the stamped sha, non-executed invocations carry none.

## Risk

Low. Tool-output text and additive optional fields on the stored invocation blob (versioned schema untouched, field optional on both arms).

## Deploy Plan

Standard deploy.

### DIFF
--- a/front/lib/api/actions/servers/sandbox_functions/tools/get.test.ts
+++ b/front/lib/api/actions/servers/sandbox_functions/tools/get.test.ts
@@ -60,6 +60,10 @@ describe("formatSandboxFunction", () => {
 
     expect(out).toContain("greet: Greet a user by name.");
     expect(out).toContain("userIdentity: optional");
+    // The serving-version fields mirror the publish tool's receipt. This function was created
+    // without a hash (pre-hash publishes exist in prod), so the field must degrade explicitly.
+    expect(out).toContain(`updatedAt: ${fn.updatedAt.toISOString()}`);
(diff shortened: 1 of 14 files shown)


davidebbo (approved):
LGTM

PR #30339, opened 2026-08-11
Add contextual consumption filter facets

flvndvd:
## Description

This PR consolidates consumption analytics filtering behind a dedicated facets endpoint. The product goal is to help users build filters that lead to meaningful results, rather than letting them select combinations that silently produce an empty analytics view.

The endpoint returns the complete filter catalog for the selected workspace and period, together with contextual availability and document counts. Availability considers the currently staged filters while excluding the facet’s own dimension, so users can continue exploring alternatives within a category without hiding its sibling options.

The UI uses this information to clearly mark values with no activity for the selected period and filter context. Unavailable values remain visible so users understand what exists in the workspace, but they cannot accidentally add a filter known to return zero results. Previously selected values remain removable even if they become unavailable after another filter changes.

The endpoint uses a POST body because the filtering context can become too large and structured for query parameters (we will need to align previous endpoints). The client maintains a stable SWR cache key, normalizes selections, debounces contextual updates, cancels superseded client requests, and retains only the latest contextual response.

The Elasticsearch work is scoped to the requested period and uses bounded composite pagination. This avoids silently truncating high-cardinality dimensions or exceeding Elasticsearch’s aggregation bucket limit (like we've seen on bigger workspaces 🙈). Dimensions are processed sequentially to avoid creating a large burst of concurrent searches, and database-backed catalog reads only begin after the Elasticsearch phase succeeds.

The filtering experience also received an accessibility and polish pass. Unavailable values use a calm “No activity”
state instead of a forbidden cursor, contextual updates are surfaced, keyboard users can navigate categories, active
scope and model controls expose their selected state, and unavailable controls remain discoverable to assistive
technologies. Historical agents whose scope can no longer be determined are intentionally not displayed for now, as
there is no appropriate product category for them.

https://github.com/user-attachments/assets/d0772d34-bde0-4534-bcbd-024cd6954ba0

## Performance expectations

This endpoint is intentionally not presented as the final performance architecture. Returning a complete catalog requires loading current workspace members, teams, agents, models, tools, and skills, then merging those values with period-scoped Elasticsearch data. These reads are sequential to protect the database from connection fan-out, and unnecessary skill payloads are excluded, but large workspaces may still experience noticeable latency.

Moving most catalog metadata and historical labels into Elasticsearch would materially reduce the remaining database work. That migration is outside the scope of this PR and the probably even outside of the current filtering initiative, because it requires broader indexing and lifecycle decisions. The known limitation is documented, and we will
revisit it based on production observability and the performance patterns we see in real workspaces.


front/components/workspace/analytics/usageFilter.test.ts
```diff
@@ -1,23 +1,97 @@
...
+      skills: ["skill-1"],
+      sources: ["slack"],
+    });
   });
 
   it("omits empty member and team selections", () => {
     expect(toConsumptionScopeFilter({ member: [], team: [] })).toEqual({});
   });
+
+  it("bounds the total number of selections serialized into analytics URLs", () => {
```
flvndvd, line 75: 🚗 Added a global selection limit as this endpoint still uses a GET requests so until we eventually move it to a `POST` ensures we don't oversize the URL limit.

front/lib/api/analytics/consumption/facet_catalog.ts
```diff
@@ -0,0 +1,109 @@
...
+        pictureUrl: null,
+        maker: getModelMaker(model),
+        tier:
+          ModelsTierResource.getTierForModel(
+            model.modelId,
+            model.defaultReasoningEffort
+          ) ?? undefined,
+      })),
+    tool: mcpServers.map((server) => ({
+      value: isRemoteMCPServerType(server) ? server.sId : server.name,
```
avervaet, line 94: We might want to align the remote facet key with the canonical value written to `tool.server_name`? `server.sId` currently differs from the server/view name used by `mcpServerName`, which could make remote tools appear unavailable or produce empty results.

avervaet: For example, suppose a remote MCP server has:

server.sId = "mcp_remote_123"
server.name = "linear"
mcpServerView.name = "Linear"

The facet catalog currently exposes:
value: "mcp_remote_123"

But when a tool action is recorded, the analytics document may contain:
tool.server_name: "Linear"

When the user selects the facet, the query becomes effectively:
{ term: { "tool.server_name": "mcp_remote_123" } }

That query will not match documents containing "Linear", so the tool can incorrectly appear unavailable or return zero results.

avervaet: If remote actions already persist the sId in mcpServerName, then the current catalog code is correct.
Trying to check.

avervaet (approved):

front/lib/api/analytics/consumption/facet_catalog.ts
```diff
@@ -0,0 +1,109 @@
...
+      label: getMcpServerDisplayName(server),
+      pictureUrl: null,
+    })),
+    skill: skills.map((skill) => ({
+      value: skill.sId,
+      label: skill.name,
+      pictureUrl: null,
```
aubin-tchoi, line 101: would be nice to surface the icon here

front/lib/api/analytics/consumption/facet_catalog.ts
```diff
@@ -0,0 +1,109 @@
...
+        tier:
+          ModelsTierResource.getTierForModel(
+            model.modelId,
+            model.defaultReasoningEffort
+          ) ?? undefined,
+      })),
+    tool: mcpServers.map((server) => ({
+      value: isRemoteMCPServerType(server) ? server.sId : server.name,
+      label: getMcpServerDisplayName(server),
+      pictureUrl: null,
```
aubin-tchoi, line 96: same here

aubin-tchoi (approved):
LGTM

front/lib/api/analytics/consumption/facet_catalog.ts
flvndvd: Great catch 🙏. Fixed, we should really clean this remote MCP logic as it's very brittle right now.

PR #30344, opened 2026-08-11
[dsbx] fix: drain-safe runner stdout and honest truncation classification

fontanierh:
## Description

Fixes the top Pod-function prod failure: large results cut mid-JSON on the cold path.

- The cold runner ends with `process.exit`, which drops queued async stdout writes. A ~250KB result envelope was cut mid-JSON while the runner exited 0. All runner stdout envelopes (run/get/build/db handlers) now go through one synchronous fd-1 write helper that loops over partial writes and retries EAGAIN, so exit can never drop envelope bytes. The warm socket path already drains (`writeThenEnd`/`drainPending`).
- When the last stdout line still fails to parse and starts with `{` or `[`, dsbx now mints `output_truncated` with the byte count read, instead of `function produced non-JSON output (...)` with a 512-char prefix of the valid payload embedded in the error message (which reads like an envelope-wrapping bug). The raw prefix goes to stderr for logs. Genuinely non-JSON lines (stray prints) keep the existing message and snippet.
- Fixes the stale comment claiming `runner.js` is committed (it is a generated bundle, see `build.rs`).

Depends on #30333 (front must accept the `output_truncated` code before a dsbx release ships this).

## Tests

- New 2MB round-trip regression test in `runner.test.ts`, confirmed red before the drain fix (envelope cut mid-string) and green after.
- `run.rs` tests: cut JSON line and cut array line classify as `output_truncated` with byte count and no payload prefix; non-JSON line keeps the current message.

## Risk

Low: emission becomes synchronous (same bytes, same line), and the new classification only changes the error code/message for envelopes that were already failing.

## Deploy Plan

No image change until the next dsbx release + `DSBX_CLI_VERSION` bump, which must happen after #30333 is deployed.

### DIFF
--- a/cli/dust-sandbox/functions-runner/emit.ts
+++ b/cli/dust-sandbox/functions-runner/emit.ts
@@ -1,0 +1,43 @@
+// Drain-safe stdout emission for the runner's one-line JSON envelopes.
+//
+// runner.ts ends with `process.exit(await main())`, and process.exit does NOT
+// drain queued asynchronous stdout writes: on a pipe, `process.stdout.write`
+// buffers whatever the kernel cannot take immediately, and exiting drops those
+// bytes, cutting a large envelope mid-JSON while the process still exits with
+// the handler's code. The reader (dsbx, or front through it) then sees a
+// JSON-prefixed line that does not parse.
+//
+// Writing synchronously to fd 1 hands every byte to the kernel before
+// returning: the loop covers partial writes, and retries EAGAIN for the case
+// where fd 1 is a non-blocking pipe whose buffer is full (the reader will
+// drain it). Once this returns, no queued write remains for exit to drop.
+//
+// The warm server (serve.ts) replies over a unix socket and has its own
+// drain-safe path (writeThenEnd/drainPending).
+
+import { writeSync } from "node:fs";
+
+function isRetryableWriteError(error: unknown): boolean {
+  return (
+    typeof error === "object" &&
+    error !== null &&
+    "code" in error &&
+    (error.code === "EAGAIN" || error.code === "EINTR")
+  );
+}
+
+/** Serialize `envelope` and write it to stdout as one line, synchronously. */
+export function emitEnvelopeLine(envelope: unknown): void {
+  const bytes = Buffer.from(`${JSON.stringify(envelope)}\n`, "utf8");
+  let offset = 0;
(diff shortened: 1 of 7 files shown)


cli/dust-sandbox/functions-runner/emit.ts
```diff
@@ -0,0 +1,43 @@
...
+  );
+}
+
+/** Serialize `envelope` and write it to stdout as one line, synchronously. */
+export function emitEnvelopeLine(envelope: unknown): void {
+  const bytes = Buffer.from(`${JSON.stringify(envelope)}\n`, "utf8");
+  let offset = 0;
+  while (offset < bytes.length) {
+    try {
+      offset += writeSync(1, bytes, offset, bytes.length - offset);
```
davidebbo, line 35: nit: isn't there a constant for stdout, like process.stdout?

cli/dust-sandbox/functions-runner/emit.ts
```diff
@@ -0,0 +1,43 @@
...
+    "code" in error &&
+    (error.code === "EAGAIN" || error.code === "EINTR")
+  );
+}
+
+/** Serialize `envelope` and write it to stdout as one line, synchronously. */
+export function emitEnvelopeLine(envelope: unknown): void {
+  const bytes = Buffer.from(`${JSON.stringify(envelope)}\n`, "utf8");
+  let offset = 0;
+  while (offset < bytes.length) {
```
davidebbo, line 33: Not sure I understand how it can take more than one iteration is the write doesn't fail. Isn't it writing all the bytes at once?

davidebbo (commented):
LGTM, but with a couple honest questions :)

cli/dust-sandbox/functions-runner/emit.ts
fontanierh: ah yes I can use process.stdout.fd

cli/dust-sandbox/functions-runner/emit.ts
fontanierh: There's no guarantee. The function returns what it was actually able to wirte, non-blocking linux pipes can cause that apparently

davidebbo: Wow, would not have expected that. I thought it would do it all or fail.

PR #30348, opened 2026-08-11
[front] feat: read back spilled Pod function results

fontanierh:
## Description

Read side of result spill-to-file for Pod functions. The runner (separate dsbx PR) will write results over an inline cap to a sandbox-local scratch file (`/tmp/dust-fn-results/`) and emit a `{ok, resultFile, resultBytes}` pointer on stdout instead of the payload.

- `parseStdoutResultEnvelope` now surfaces the pointer; the invocation execute path reads the file back via `sandbox.readFile` (same provider mechanism as `build_on_sandbox.ts`) and runs the existing normalization on its content.
- A failed read-back becomes an `invocation_failed` error naming the file: loud, never silent data loss.
- The pointer rides the exec's stdout, which function code can also write to, so read-back is restricted to the runner's scratch directory and the content must still normalize as a result envelope (fails closed on anything else).
- Envelopes from older dsbx never carry pointers: behavior unchanged until the write side ships.

Depends on #30344.

## Tests

Pointer extraction, spill resolution (happy path, read failure, bad JSON, path outside the scratch dir refused before any read), and two execute-path tests (spilled result delivered, read failure fails the invocation).

## Risk

Low: dead code path until the dsbx write side is released; inline results flow exactly as before.

## Deploy Plan

Standard deploy. Must be live before the dsbx release that spills results bumps `DSBX_CLI_VERSION`.

### DIFF
--- a/front/lib/api/sandbox_functions/result_delivery.test.ts
+++ b/front/lib/api/sandbox_functions/result_delivery.test.ts
@@ -1,5 +1,9 @@
-import { parseStdoutResultEnvelope } from "@app/lib/api/sandbox_functions/result_delivery";
-import { describe, expect, it } from "vitest";
+import {
+  parseStdoutResultEnvelope,
+  resolveSpilledResult,
+} from "@app/lib/api/sandbox_functions/result_delivery";
+import { Err, Ok } from "@app/types/shared/result";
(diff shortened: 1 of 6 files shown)


davidebbo (commented):
LGTM

PR #30349, opened 2026-08-11
[dsbx] Typed error envelope and distinct exit codes for tools failures

fontanierh:
## Description

`dsbx tools` failures currently surface as a single bail string ("POST {url} returned {status}: {body}") logged to stderr, exit 1, nothing on stdout. Sandbox functions that shell out to `dsbx tools --json` have to regex stderr prose to figure out what went wrong.

- Parse front's `{"error":{"type","message"}}` body into a typed `DustApiError` in the API client (POST, GET, and the poll path).
- Under `--json`, failures now emit `{"error":{"code","message","retryable","status"?}}` on stdout (same pattern as `function run`'s emit_error contract).
- Distinct exit codes per class: 10 invalid_request, 11 invalid_sandbox_token, 12 fast_function_called_tools, 13 rate_limited, 14 server_error. 1 stays the generic failure, 2 stays clap's usage error.
- `invalid_sandbox_token` is classified terminal and the message names the reason: the JWT is minted once per invocation with a ~2-minute TTL, retrying within the invocation cannot succeed.
- The fast-mode tools-denied 403 gets its own `fast_function_called_tools` code, matched on the stable api_error type, with a message fallback for fronts that still return it as `invalid_request_error`.
- stderr diagnostic now prints the full context chain instead of the outermost message only.

Code strings and exit codes are append-only from here.

## Tests

Unit tests for classification (token, fast-403 by type and by message fallback, 429, 5xx, non-JSON bodies), envelope shape, downcast through anyhow context, and the poll-path envelope. `cargo test --bin dsbx` and the e2e_local test pass locally.

## Risk

Low. Non-json output is unchanged on success; failures keep the stderr line and non-zero exit. Consumers that only check "exit != 0" see no difference. New exit codes only replace the previous blanket 1 for API failures.

## Deploy Plan

Standard dsbx release; picked up by sandbox images on the next bump.

### DIFF
--- a/cli/dust-sandbox/src/api/client.rs
+++ b/cli/dust-sandbox/src/api/client.rs
@@ -6,6 +6,7 @@ use reqwest::header::{HeaderMap, HeaderValue, AUTHORIZATION, CONTENT_TYPE};
 use serde::{de::DeserializeOwned, Serialize};
 use tokio::time::sleep;
 
+use super::error::DustApiError;
 use super::types::{
     parse_action_poll_response, ActionPollResponse, CallToolPostResponse, CallToolRequest,
     CallToolResponse, CallToolResult, MCPServerView, SandboxServerViewsResponse,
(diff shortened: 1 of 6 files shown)


davidebbo (commented):
LGTM

PR #30353, opened 2026-08-11
[dsbx] feat: spill oversized function results to a file with a 5MB hard cap

fontanierh:
## Description

Write side of result spill-to-file, completing the large-output path:

- A run result whose serialized envelope exceeds `RESULT_INLINE_CAP_BYTES` (256KB) is written to `/tmp/dust-fn-results/<uuid>.json` and the wire carries a `{ok, resultFile, resultBytes}` pointer instead. Front (#30348) reads the file back through the provider and normalizes it exactly like an inline outcome.
- Over `RESULT_HARD_CAP_BYTES` (5MB) the result is refused with `output_too_large` and a message telling the function to store large data in a pod file or database and return a pointer.
- One policy for cold and warm: the warm server applies it before its reply frame, so both paths share the same numbers.
- A failed spill write (disk pressure) falls back to inline emission, which the drain-safe stdout writer and the backpressure-aware socket path both deliver whole: the caps are transport hygiene, never data loss.
- No spill-file cleanup: sandbox scratch is ephemeral.

Depends on #30348 (front read side must be deployed before a dsbx release ships this, or spilled results would be lost).

## Tests

Policy unit tests (at-cap inline, cap+1 spills and round-trips, hard-cap+1 refused with nothing written, unwritable spill dir falls back inline, large error outcomes spill too) plus e2e: 200KB stays inline and intact, 2MB run and warm invocations produce pointers whose files round-trip, 6MB is refused with exit 1.

## Risk

Low until released: behavior only changes once a dsbx release with this ships in an image. The main risk is old-front/new-dsbx skew losing spilled results, addressed by the deploy ordering below.

## Deploy Plan

No image change here. The dsbx release + `DSBX_CLI_VERSION` bump happens separately, strictly after #30333 and #30348 are deployed.

### DIFF
--- a/cli/dust-sandbox/functions-runner/emit.test.ts
+++ b/cli/dust-sandbox/functions-runner/emit.test.ts
@@ -1,0 +1,114 @@
+import { describe, expect, test } from "bun:test";
+import {
+  mkdtempSync,
+  readdirSync,
+  readFileSync,
+  rmSync,
+  writeFileSync,
+} from "node:fs";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
+
+import {
+  applyResultSpillPolicy,
+  RESULT_HARD_CAP_BYTES,
+  RESULT_INLINE_CAP_BYTES,
+} from "./emit.ts";
+import type { Output } from "./protocol.ts";
+
+function outputOfSerializedBytes(targetBytes: number): Output {
+  // {"ok":true,"output":{"big":"…"}} — pad `big` so the serialized envelope
+  // lands exactly on targetBytes.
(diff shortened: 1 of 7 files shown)


davidebbo:
Seems unfriendly to a world where we'd have multiple instances of the pod sandbox to get more scale (since tmp files are only in one). But I guess we'll tackle if we get there...

davidebbo (commented):
LGTM

fontanierh:
> Seems unfriendly to a world where we'd have multiple instances of the pod sandbox to get more scale (since tmp files are only in one). But I guess we'll tackle if we get there...

true: we'll probably have to have a gcsfused folder for that....

PR #30354, opened 2026-08-11
[dsbx] tools arg-input hardening: --args-json, file scalar coercion, union schema types

fontanierh:
## Description

Stacked on #30349. Three gaps in `dsbx tools` argument handling that force callers into workarounds:

- `--args-json <json|->` passes tool arguments as one JSON object, sent verbatim (no per-key parsing, no coercion). `-` reads the payload from stdin for values larger than ARG_MAX. Rejects non-object payloads and combining with `--key value` pairs.
- `__file__:`-passed contents skipped schema coercion entirely, so a file-passed "42" against an integer schema was guaranteed to fail server-side validation. Contents (trimmed, files routinely end with a newline) now coerce when the schema declares a scalar type. A schema-declared "string" still never coerces.
- `property_type` only understood `"type": "<string>"`. It now reduces nullable unions: `"type": ["number", "null"]` arrays and single-scalar `anyOf` resolve to the non-null type. Ambiguous unions (two non-null types, `$ref` entries) keep today's heuristic behavior.

## Tests

Unit tests for all three: object/stdin-sentinel/reject cases for --args-json, scalar and trailing-newline file coercion (plus string-schema and no-schema pins), and the union reductions end to end through parse_args.

## Risk

Low. `--args-json` is new surface. File coercion only changes cases that previously always failed validation; no-schema and string-schema behavior is pinned unchanged. Union reduction only adds coercion where the schema is unambiguous.

## Deploy Plan

Standard dsbx release, after #30349 merges.

### DIFF
--- a/cli/dust-sandbox/src/commands/tools/exec.rs
+++ b/cli/dust-sandbox/src/commands/tools/exec.rs
@@ -9,9 +9,10 @@ pub async fn cmd_exec(
     server_name: &str,
     tool_name: &str,
     raw_args: &[String],
+    args_json: Option<&str>,
     json: bool,
 ) -> anyhow::Result<()> {
-    match run_exec(client, server_name, tool_name, raw_args, json).await {
+    match run_exec(client, server_name, tool_name, raw_args, args_json, json).await {
         Ok(()) => Ok(()),
         Err(err) => {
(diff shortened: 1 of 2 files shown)


davidebbo (commented):
LGTM

PR #30357, opened 2026-08-11
[dsbx] feat: accept the pod_editor_required user identity policy in the functions-runner

fontanierh:
## Description

Adds `pod_editor_required` to the set of `schema.userIdentity` values the functions-runner accepts when extracting a function's contract. The runner only parses and forwards the policy string; enforcement lands in front separately.

Without this, a publish that declares the new policy throws at schema extraction, so the runner has to accept the value before front starts allowing it.

## Tests

Added a fixture + test covering the new policy; `bun test` in `cli/dust-sandbox/functions-runner` passes.

## Risk

Low: purely additive parsing. Functions using the new policy stay deny-by-default at invocation time until the front enforcement ships (unknown policies fail closed there).

## Deploy Plan

Deploy with the next dsbx image release. Must be live before front accepts publishes declaring `pod_editor_required`.

### DIFF
--- a/cli/dust-sandbox/functions-runner/fixtures/pod-editor.ts
+++ b/cli/dust-sandbox/functions-runner/fixtures/pod-editor.ts
@@ -1,0 +1,14 @@
+import { z } from "zod";
+
+export const schema = {
+  description: "Refresh the pod cache",
+  userIdentity: "pod_editor_required",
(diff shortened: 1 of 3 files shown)


davidebbo (commented):
LGTM

PR #30363, opened 2026-08-11
[front] feat: pod_editor_required identity policy + isPodEditor in the identity envelope

fontanierh:
## Description

Adds a `pod_editor_required` user identity policy for Pod Functions, enabling the "editors refresh, viewers read the cache" pattern:

- The policy authorizes a workspace member who can administrate the function's pod: pod editor group members, plus workspace admins via role. This is the same predicate the pod UI serializes as `isEditor` (`space.canAdministrate`). `canWrite` would not work here: the pod member group also holds write on the space, so it cannot separate editors from viewers.
- `authorizeSandboxFunctionInvocation` now receives the function's space (it previously saw only policy + origin).
- The per-invocation identity envelope carries `isPodEditor`, and `currentUser()` in `@dust/pod` exposes it. The field is optional in the zod schema: envelopes written by older front parse as `false`.
- Frame side, `UserIdentityState` gains a display-only `isPodEditor` (functions always resolve the caller server-side). Pod hosts (frame tabs, pinned banner, frame sheet) thread the pod's `isEditor` into the scoped identity; conversation/public hosts and the SSR cache answer `false`, and the viz RPC data API normalizes answers from hosts that predate the field.

Old front deploys deny the unknown policy at invocation time (the existing fail-closed arm), so a mixed-version deploy fails closed rather than open.

Stacked on #30357 (runner accepts the policy string).

## Tests

- New `workspace_user.test.ts` matrix: pod editor / pod member non-editor / member outside the pod / workspace admin / userless caller.
- Endpoint tests: pod editor gets 201, non-editor pod member gets 401.
- Envelope round-trip: front test asserts `isPodEditor` in `DUST_POD_USER_IDENTITY`; `@dust/pod` tests cover present/absent parsing; viz tests cover host answers with and without the field.

## Risk

Low: additive on every wire shape; no existing policy changes behavior. The only sequencing constraint is the runner (see below).

## Deploy Plan

1. #30357 must be live in the dsbx image first: the current runner throws at schema extraction on the unknown policy, so publishes declaring `pod_editor_required` fail at build until then.
2. Deploy front. Do not advertise the policy in builder-facing prompt surfaces until both are out.

### DIFF
--- a/cli/dust-sandbox/pod/identity.ts
+++ b/cli/dust-sandbox/pod/identity.ts
@@ -11,15 +11,16 @@ export interface WorkspaceUserIdentity {
   readonly lastName: string | null;
   readonly fullName: string;
   readonly image: string | null;
-}
-
-interface WorkspaceUserIdentityEnvelope {
-  workspaceId: string;
-  user: WorkspaceUserIdentity;
+  // Whether the caller is an editor of the Pod this function belongs to (workspace admins
+  // included). Gate refresh-style mutations on it; viewers only read.
+  readonly isPodEditor: boolean;
 }
 
 const workspaceUserIdentityEnvelopeSchema = z.object({
   workspaceId: z.string(),
+  // Optional so envelopes written before the field existed still parse; absent means false,
+  // the restrictive reading.
+  isPodEditor: z.boolean().optional(),
   user: z.object({
     sId: z.string(),
     firstName: z.string(),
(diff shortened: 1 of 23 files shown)


cli/dust-sandbox/pod/identity.ts
```diff
@@ -11,15 +11,16 @@ export interface WorkspaceUserIdentity {
   readonly lastName: string | null;
   readonly fullName: string;
   readonly image: string | null;
-}
-
-interface WorkspaceUserIdentityEnvelope {
-  workspaceId: string;
-  user: WorkspaceUserIdentity;
+  // Whether the caller is an editor of the Pod this function belongs to (workspace admins
+  // included). Gate refresh-style mutations on it; viewers only read.
```
davidebbo, line 15: So the determination of whether it's a "refresh-style mutations" is up to the function implementation, right? There is no magic that can detect those scenarios.

davidebbo (approved):
LGTM

cli/dust-sandbox/pod/identity.ts
fontanierh: No we don't try to detect it for now

PR #30365, opened 2026-08-11
[front] fix: namespace local sandbox image builds so CI cannot skip a region

davidebbo:
## What was happening

The Sandbox Image Registry workflow builds an image for one region and silently skips the other — e.g. [run 31461680537](https://github.com/dust-tt/dust/actions/runs/31461680537), where US built `dust-base_0-8-73` and EU reported it as already existing.

`sandbox_image_check.ts` decides "already built" by asking E2B whether a template with alias `dust-base_<tag>` exists. That treats the alias as proof the version was built from `main`, which is only true if CI is the sole creator of release aliases. Nothing enforced that.

To exercise a locally modified image you had to publish it under the exact alias the registry names, since `getSandboxImage()` had no dev override and the runtime resolves `formatSandboxImageId(config.imageId)`. `sandbox_image_build.ts` then refuses to overwrite an existing alias without `--rebuild`, so while iterating the path of least resistance was to bump `DUST_BASE_IMAGE_VERSION` again. That burned release numbers ahead of `main`.

Evidence from the two E2B regions (`createdAt` on the template records):

| tag | EU created | US created | landed in main |
|---|---|---|---|
| `0-8-71` | Aug 10 19:27:23 | Aug 10 19:27:30 | Aug 10 17:21 |
| `0-8-72` | **Aug 10 21:24:14** | Aug 10 22:11:01 | Aug 10 22:09 |
| `0-8-73` | **Aug 10 21:30:14** | Aug 11 05:28:06 | Aug 11 05:26 |
| `0-8-74` | Aug 10 21:30:53 | — | never |
| `0-8-75` | Aug 10 21:34:16 | — | never |
| `0-8-76` | Aug 10 21:46:20 | — | never |
| `0-8-78` | Aug 11 09:21:24 | — | never |

US timestamps match the CI build steps to the second. The EU ones for `0-8-72`/`0-8-73` precede CI by 47 minutes and 8 hours, and `0.8.74`–`0.8.78` appear in no commit on any branch — they came from uncommitted working trees. `sandbox-img-registry.yml` is the only workflow that touches E2B, and no run of it happened in those windows.

Consequence: when `main` landed on a version already burned in EU, the EU check reported it existing, the EU build was skipped, and EU kept serving an image built from an unmerged working tree while US got the one built from the merged commit. Same tag, different contents, no signal anywhere.

## The fix

- `sandbox_image_build.ts` gains `--release`, which the workflow passes. Without it the build is namespaced with `SBX_DEV_IMAGE_SUFFIX`, and it fails fast if that is unset rather than guessing.
- `getSandboxImage()` resolves the same dev alias in development, so a locally built image is actually usable — this closes the trap that forced the behaviour. It sits inside the existing `isDevelopment()` guard, so the suffix cannot take effect in production. Every sandbox-creating path already routes through `getSandboxImage()`, so they all pick it up.
- `devSandboxImageId()` holds the naming convention in one place, so the published alias and the resolved alias cannot drift.

Local iteration no longer requires bumping the version at all: keep `DUST_BASE_IMAGE_VERSION` at main's value, set `SBX_DEV_IMAGE_SUFFIX=<you>`, and build `dust-base_<version>-<you>`.

| local setup | resolves to |
|---|---|
| `SBX_DEV_IMAGE_SUFFIX` unset (default) | `dust-base_0-8-79` — the CI-built image |
| `SBX_DEV_IMAGE_SUFFIX=david` | `dust-base_0-8-79-david` — your own build |
| production, suffix set anyway | `dust-base_0-8-79` — suffix ignored |

## Heads-up: this bumps dust-base to 0.8.79

EU still holds orphaned `0-8-76` and `0-8-78` (E2B refused to delete them — paused and running sandboxes still reference them, and no sandboxes were killed). `main` was at `0.8.73`, so the next bump to `0.8.74` would have skipped EU again. Versions only increase, so moving past the burned range retires the landmine by construction. `0.8.77` was never burned but `0.8.78` was, hence `0.8.79`.

This means a fresh `dust-base` build in both regions on the next front deploy. A rebuild is never bit-identical — apt and npm resolution can drift — so it is a small but real image change, deliberately incurred rather than driven by an image change. Same risk profile as any routine bump.

## Testing

- 4 new tests in `lib/api/sandbox/image/index.test.ts`, 51ms total, no DB or network. They cover: the dev alias never collides with the release alias and is a valid E2B alias; no suffix keeps the release alias (the common local case); development plus suffix resolves the dev alias; and production ignores the suffix.
- Confirmed non-vacuous — reverting the one-line runtime change fails with `expected '0.8.73' to be '0.8.73-test-dev'`.
- Full sandbox suite passes (198 tests / 18 files), `npx tsgo --noEmit` clean, biome clean.
- All three build-script paths exercised against the real script: refusal with exit 1 when no suffix is set, `dust-base_0-8-73-david` with a suffix, `dust-base_0-8-73` with `--release`.

### DIFF
--- a/front/scripts/sandbox_image_build.ts
+++ b/front/scripts/sandbox_image_build.ts
@@ -41,8 +41,8 @@ function resolveImageId(
   if (!devSuffix) {
     logger.error(
       "Set SBX_DEV_IMAGE_SUFFIX (e.g. your username) in front/.env.local to " +
-        "build a sandbox image locally, or pass --release to publish the " +
-        `release alias '${formatSandboxImageId(releaseImageId)}' (CI only).`
+        "build a sandbox image locally, or pass --release if you intend " +
+        "to deploy the image in production."
     );
     process.exit(1);
   }
--- a/front/scripts/sandbox_image_build.ts
+++ b/front/scripts/sandbox_image_build.ts
@@ -1,6 +1,7 @@
 import config from "@app/lib/api/config";
 import type { SandboxImageId } from "@app/lib/api/sandbox/image";
 import {
+  devSandboxImageId,
   formatSandboxImageId,
   getSandboxImageFromRegistry,
 } from "@app/lib/api/sandbox/image";
@@ -22,6 +23,31 @@ interface BuildArgs {
   dockerRegistry: string;
   rebuild: boolean;
   confirm: boolean;
+  release: boolean;
+}
+
+// The release alias is CI's to publish: `sandbox_image_check.ts` treats its
+// presence as "this version was built from main" and skips the region's build
+// otherwise. Local builds are namespaced per developer instead.
+function resolveImageId(
+  releaseImageId: SandboxImageId,
+  release: boolean
+): SandboxImageId {
+  if (release) {
+    return releaseImageId;
+  }
+
+  const devSuffix = config.getSandboxDevImageSuffix();
+  if (!devSuffix) {
+    logger.error(
+      "Set SBX_DEV_IMAGE_SUFFIX (e.g. your username) in front/.env.local to " +
+        "build a sandbox image locally, or pass --release to publish the " +
+        `release alias '${formatSandboxImageId(releaseImageId)}' (CI only).`
+    );
+    process.exit(1);
+  }
+
+  return devSandboxImageId(releaseImageId, devSuffix);
 }
 
 async function promptYesNo(question: string): Promise<boolean> {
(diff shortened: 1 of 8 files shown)


front/scripts/sandbox_image_build.ts
```diff
@@ -22,6 +23,31 @@ interface BuildArgs {
...
+  if (release) {
+    return releaseImageId;
+  }
+
+  const devSuffix = config.getSandboxDevImageSuffix();
+  if (!devSuffix) {
+    logger.error(
+      "Set SBX_DEV_IMAGE_SUFFIX (e.g. your username) in front/.env.local to " +
+        "build a sandbox image locally, or pass --release to publish the " +
+        `release alias '${formatSandboxImageId(releaseImageId)}' (CI only).`
```
id13:
```suggestion
      "Set SBX_DEV_IMAGE_SUFFIX (e.g. your username) in front/.env.local to " +
        "build a sandbox image locally, only use --release if you intend to deploy the image in production"
```

id13 (approved):
LGTM

PR #30369, opened 2026-08-11
feat: let admins filter consumption analytics on every workspace agent

adrsimon:
## Description

The Agents usage panel only listed agents the viewer can already see. The `Private` filter tab was always empty apart from your own private agents, and unpublished or space-restricted agents couldn't be filtered on. We want admins to be able to introspect usage from all agents in the workspace.

Adds a dedicated `analytics` agents view for agents. The panel always asks for that view, the `Private` scope tab is hidden below the admin role, and there's a new `All` scope tab, now the default. Opening the page to more roles later is a change to two predicates in the view, not a new error branch.

One thing to know: an admin can now list name / description / tags / model / requestedSpaceIds for every agent, private ones included.

## Tests

Unit tests + Manually on a hive: seeded a hidden agent owned by another user, in a restricted space, with consumption, it shows in the Agents dimension and is selectable under `Private` and `All` as an admin. The tests also cover the non-admin path: a manager gets the non-private set back, not an error, and sees the anonymised label while an admin sees the real name.

Capture d’écran 2026-08-11 à 16 19 54

## Risk

New read-only view on a private API, admin-only breadth. Existing views behave exactly as before, nothing persisted or migrated, so rollback is a plain revert. The anonymisation only narrows what non-admins already received on the new consumption page; the old Analytics page is untouched.

## Deploy Plan

Deploy front.

### DIFF
--- a/front-api/routes/w/[wId]/assistant/agent_configurations/index.ts
+++ b/front-api/routes/w/[wId]/assistant/agent_configurations/index.ts
@@ -59,7 +59,7 @@ const app = workspaceApp();
  *         description: Filter agents by view
  *         schema:
  *           type: string
- *           enum: [all, list, favorites, published, admin_internal, global, workspace]
+ *           enum: [all, analytics, list, favorites, published, admin_internal, global, workspace]
  *       - in: query
  *         name: limit
  *         required: false
@@ -193,18 +193,12 @@ app.get("/", async (ctx): HandlerResult<GetAgentConfigurationsResponseBody> => {
   let viewParam = view ? view : "all";
   // @ts-expect-error: added for backwards compatibility
   viewParam = viewParam === "assistant-search" ? "list" : viewParam;
-  if (
-    viewParam === "admin_internal" &&
-    !auth.isDustSuperUser() &&
-    !auth.isAdmin()
-  ) {
+  if (viewParam === "admin_internal" && !auth.isDustSuperUser()) {
     return apiError(ctx, {
       status_code: 404,
       api_error: {
         type: "app_auth_error",
-        message:
-          "Only Dust Super Users and workspace admins can see " +
-          "admin_internal agents.",
+        message: "Only Dust Super Users can see admin_internal agents.",
       },
     });
   }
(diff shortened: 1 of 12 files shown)


avervaet (approved):
As long as we agree that admins should be allowed to list name / description / tags / model / requestedSpaceIds for every agent, private ones included. LGTM

front-api/routes/w/[wId]/assistant/agent_configurations/index.ts
```diff
@@ -193,12 +193,18 @@ app.get("/", async (ctx): HandlerResult<GetAgentConfigurationsResponseBody> => {
   let viewParam = view ? view : "all";
   // @ts-expect-error: added for backwards compatibility
   viewParam = viewParam === "assistant-search" ? "list" : viewParam;
-  if (viewParam === "admin_internal" && !auth.isDustSuperUser()) {
+  if (
+    viewParam === "admin_internal" &&
```
Comment (flvndvd) [inline front-api/routes/w/[wId]/assistant/agent_configurations/index.ts]: Not a big fan of re-using `admin_internal` here, it really just means dust internals. Any way we can do a dedicated analytics view?

Comment (flvndvd) [reply front-api/routes/w/[wId]/assistant/agent_configurations/index.ts]: Also as we plan to open analytics to more people, IMHO it should not throw an error for this view, but mainly open the scope based on the user permissions.

flvndvd (commented):
Can we have a small distinction in the UI for the private agents (cc @ClementAupiais).

front-api/routes/w/[wId]/assistant/agent_configurations/index.ts
Comment (adrsimon) [reply front-api/routes/w/[wId]/assistant/agent_configurations/index.ts]: changed to a analytics view on agents

waiting on @ClementAupiais for the distinction in the UI, i think we can merge this version and add the icon later

Comment (ClementAupiais) [reply front-api/routes/w/[wId]/assistant/agent_configurations/index.ts]: Yeah go ahead will find a design treatment tomorrow !

front/lib/api/assistant/configuration/views.ts
```diff
@@ -169,6 +170,17 @@ async function fetchWorkspaceAgentConfigurationsWithoutActions(
         ...baseAgentsSequelizeQuery,
         where: baseWhereConditions,
       });
+
+    // Analytics reports on every agent, so admins get the private ones too.
+    // Everyone else sees what `all` returns.
+    case "analytics":
```
flvndvd, line 176: Can we add tests focusing on this one method?

flvndvd (approved):

PR #30371, opened 2026-08-11
Fix formatting

davidebbo:
## Description

e.g. see break [here](https://github.com/dust-tt/dust/actions/runs/31500825812/job/93810051049).


aubin-tchoi (approved):
🙏🏻

PR #30372, opened 2026-08-11
fix(api): correct ModelSelection swagger description to match implementation

Nils-Fedrigo:
## Description

The `ModelSelection` OpenAPI description for the public conversations API does not match the implementation, so customers reading the API reference get the wrong error contract for the per-message model override.

The annotation in `front-api/routes/v1/w/[wId]/swagger_schemas.ts` currently says:

> If the requested model is not available to the workspace, the agent's configured model is used instead. An unknown provider, model, or reasoning effort results in a 400.

`validatePublicModelSelection` in `front-api/lib/api/assistant/conversation/model_selection.ts` does neither of those things for the unauthorized case, and has an undocumented 403:

- A provider/model pair that is not in `getEnabledModelsForAuth(auth)` returns `400 model_disabled` ("The model <providerId>/<modelId> is not available to your workspace."). It does **not** silently fall back to the agent's configured model.
- When the `models_picker` feature flag is not enabled for the workspace, the request returns `403 feature_flag_not_found` before any model validation. This is not mentioned at all.

This PR corrects the description to match the code, and mirrors the same text in the committed generated spec so `front-api/public/swagger.json` stays in sync.

No behavior change: documentation strings only.

## Tests

Local validation in the sandbox was limited to:

- `git diff --check` (clean).
- `json.load()` on `front-api/public/swagger.json` to confirm it is still valid JSON after the edit.
- Confirmed the edited `description` value is byte-identical between the JSDoc annotation and the generated spec entry, so a future `npm run docs` regeneration is a no-op on this field.

The spec was **not** regenerated with `npm run docs` in the sandbox (dependency installation is not available there); the generated file was edited surgically so the diff is a single line. Worth a reviewer confirming a regeneration produces no further diff. Full validation is delegated to PR CI.

## Risk

Very low. Two documentation strings, no runtime code path touched, no schema shape change (no fields added, removed, or retyped). Trivially rollback-safe.

## Deploy Plan

Standard deploy, no ordering constraints and no migration. Note that `docs.dust.tt` serves its own copy of the spec from `dust-tt/new-docs` (`docs/developer-platform/dust-api-documentation/openapi.json`), which is currently stale and predates `modelSelection` entirely. That copy is being synced separately; this PR only fixes the source of truth in this repo.

### DIFF
--- a/front-api/public/swagger.json
+++ b/front-api/public/swagger.json
@@ -14325,7 +14325,7 @@
       },
       "ModelSelection": {
         "type": "object",
-        "description": "Optional per-message model and reasoning-effort override applied to the\nmentioned agent(s). When omitted, each agent runs its configured model.\nRequires the model picker to be enabled for the workspace; otherwise the\nrequest fails with a 403. A provider/model pair that is not authorized\nfor the workspace is rejected with a 400 (`model_disabled`), it does not\nfall back to the agent's configured model. A malformed object, or an\nunknown reasoning effort, also results in a 400.\n",
+        "description": "Optional per-message model and reasoning-effort override applied to the\nmentioned agent(s). When omitted, each agent runs its configured model.\nIf the requested model is not available to the workspace, the request fails with a 403. An unknown provider, model, or\nreasoning effort results in a 400.\n",
         "required": [
           "providerId",
           "modelId"
--- a/front-api/public/swagger.json
+++ b/front-api/public/swagger.json
@@ -14325,7 +14325,7 @@
       },
       "ModelSelection": {
         "type": "object",
-        "description": "Optional per-message model and reasoning-effort override applied to the\nmentioned agent(s). When omitted, each agent runs its configured model.\nIf the requested model is not available to the workspace, the request fails with a 403. An unknown provider, model, or\nreasoning effort results in a 400.\n",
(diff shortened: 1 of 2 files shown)


front-api/public/swagger.json
```diff
@@ -14325,7 +14325,7 @@
       },
       "ModelSelection": {
         "type": "object",
-        "description": "Optional per-message model and reasoning-effort override applied to the\nmentioned agent(s). When omitted, each agent runs its configured model.\nIf the requested model is not available to the workspace, the agent's\nconfigured model is used instead. An unknown provider, model, or\nreasoning effort results in a 400.\n",
+        "description": "Optional per-message model and reasoning-effort override applied to the\nmentioned agent(s). When omitted, each agent runs its configured model.\nRequires the model picker to be enabled for the workspace; otherwise the\nrequest fails with a 403. A provider/model pair that is not authorized\nfor the workspace is rejected with a 400 (`model_disabled`), it does not\nfall back to the agent's configured model. A malformed object, or an\nunknown reasoning effort, also results in a 400.\n",
```
Nils-Fedrigo:
```suggestion
        "description": "Optional per-message model and reasoning-effort override applied to the\nmentioned agent(s). When omitted, each agent runs its configured model.\nIf the requested model is not available to the workspace, the request fails with a 403. An unknown provider, model, or\nreasoning effort results in a 400.\n",
```

PR #30373, opened 2026-08-11
[dsbx] feat: add db delete command

davidebbo:
## Description

First step toward deleting a Pod app (#30334 added the Apps tab; deleting one is the other half of the original ask). An app delete has to remove the app's live SQLite files, and `dsbx` had no way to do that — `reconcile`/`schema`/`list`/`query` only ever create or read.

`dsbx db delete <name>` removes `{name}.db` along with its `-wal` and `-shm` sidecars, reporting the files it removed in the same one-line JSON envelope the other `db` subcommands use.

The sidecars are the load-bearing detail: `@dust/pod` runs with `wal_autocheckpoint=0`, so recent rows can live entirely in the WAL. Leaving it behind would let a later reconcile of the same name recover data the delete was meant to destroy.

## Safety

- **Name validation** goes through the existing `db_file_path`, so path traversal and names outside the frozen `^[a-z][a-z0-9_]{0,63}$` contract are refused before any file is touched.
- **Symlinks are not followed.** `remove_file` unlinks the link itself, so a link planted in the workload-writable databases dir cannot make the delete reach a foreign file — the same property the no-follow guards in `db list` rely on. Covered by a test asserting the target survives.
- **Runs unprivileged.** `/pod-state/databases` is `dust-state:agent` mode 2770 and `agent-proxied` is a supplementary member of group `agent`, so this needs no root and SEC3 does not come into play.

## Semantics worth noting

**Idempotent.** A name with nothing on disk succeeds with an empty `removed` list, so the eventual caller — which works from the litestream replica listing, not from what is live — never has to check first.

**Live files only.** The replica is the durable copy and outlives them, so deleting a database for good also means wiping that replica prefix; otherwise the next cold-start restore brings the database back. The doc comment says so at the call site, since that ordering is easy to get wrong.

## Versions

Bumps dsbx to `0.1.48` and dust-base to `0.8.80` across all four places, following d223899986: `Cargo.toml`, `Cargo.lock`, `registry.ts` and `registry.test.ts`.

## Risk

Very low. Purely additive — a new subcommand that **nothing calls yet**. The front-side delete flow (unpublish functions, wipe replicas, revoke shares, remove tabs, delete sources, itemized confirmation, audit event) builds on this in a follow-up, once this is in an image.

## Tests

Six inline tests on the removal logic: db plus sidecars, only-what-existed reporting, the already-gone case, siblings with a shared name prefix surviving, the planted-symlink case, and rejection of hostile names. Two CLI parse tests. Full suite: 296 Rust tests green, `cargo fmt --check` and `clippy --all-targets` clean, and the 20 front image-registry tests pass against the new pins.

### DIFF
--- a/cli/dust-sandbox/Cargo.lock
+++ b/cli/dust-sandbox/Cargo.lock
@@ -313,7 +313,7 @@ checksum = "92773504d58c093f6de2459af4af33faa518c13451eb8f2b5698ed3d36e7c813"
 
 [[package]]
 name = "dust-sandbox"
-version = "0.1.47"
+version = "0.1.48"
 dependencies = [
  "anyhow",
  "base64",
--- a/cli/dust-sandbox/Cargo.toml
+++ b/cli/dust-sandbox/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "dust-sandbox"
-version = "0.1.47"
+version = "0.1.48"
 edition = "2021"
 
 [[bin]]
--- a/cli/dust-sandbox/src/commands/db/delete.rs
+++ b/cli/dust-sandbox/src/commands/db/delete.rs
@@ -1,0 +1,149 @@
+use std::path::{Path, PathBuf};
+
+use anyhow::{anyhow, Result};
+
+use super::{db_file_path, emit_error};
+
+/// SQLite sidecars that belong to a database file and must go with it. `@dust/pod` runs with
+/// `wal_autocheckpoint=0`, so recent rows can live entirely in `-wal`: leaving it behind would let a
+/// later reconcile of the same name recover data this delete was meant to destroy.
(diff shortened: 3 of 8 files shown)


davidebbo:
Superseded by #30375, which now removes the live database files with a plain `rm` that front builds, run through the existing `execDbCommand` path, instead of a new `dsbx` subcommand.

Per review feedback: dsbx is the agent-facing CLI, so a destructive database primitive there would be discoverable from inside the sandbox. The `rm` approach adds no such surface, and grants no new capability either — `/pod-state/databases` is mode 2770 group `agent`, so workload code can already unlink its own database files.

It also drops the release coupling entirely: no dsbx bump, no dust-base bump, and no rollout window where front calls a command that older sandboxes lack. #30375 ships on its own.

PR #30375, opened 2026-08-11
[front] feat: delete a Pod app

davidebbo:
## Description

Completes the other half of the original ask (#30334 added the Apps tab; this deletes one). An app delete removes its published functions, its databases — live files *and* replicas — its shared Frames and pinned tabs, and finally its source folder.

## The step order is the design

Two constraints fix it, and both are easy to get wrong:

1. **Functions first.** Unpublishing before touching data means no invocation can arrive while the data underneath it is being removed.
2. **Live database files before replicas.** A running litestream re-replicates a database it can still see, so wiping the replica prefix first just gets it recreated — the same hazard the pod-scrub path already notes.

**And the source folder goes last, deliberately.** Every step is idempotent, and the folder is what makes the app appear in the Apps tab. So a failure part-way through leaves the app *still listed and simply re-deletable* — retry is clicking Delete again. Deleting the folder first would invert that into an invisible half-deleted app. Both orderings are pinned by tests.

## UX

Removing databases needs a live sandbox, so a sleeping pod is woken and the dialog says so rather than showing a bare spinner through a cold start.

The confirmation names what goes — functions, databases, Frames, file count — rather than only counting it, and warns separately about data that cannot be recovered and about published Frame links that will break. It requires typing `delete`, following `DeletePodDialog`.

Deletion is refused for the synthetic **unfiled** app: it is a presentation device, not a folder, and its artifacts each belong to whoever published them at the pod root, so there is nothing coherent to delete.

## Risk

Medium — it destroys data by design. Mitigated by the ordering above (a partial failure is visible and retryable rather than silent), the replica verification, and the typed confirmation. Reads and the Apps tab itself are untouched. No sandbox image or dsbx release is involved.

## Tests

Six functional tests on the endpoint: the happy path asserting the function row is gone and that the database file plus both sidecars are removed by on-disk path, the live-before-replica ordering, the surviving-replica failure, 404 for an unknown app, refusal for pod-root artifacts (asserting the function survives), and rejection of a non-app-shaped prefix. Existing suite green on the current `main` base: the 8 GET tests from #30334, plus 64 front tests across `lib/api/projects` and the audit schema checks.

### DIFF
--- a/front-api/routes/w/[wId]/pods/[podId]/apps/delete.test.ts
+++ b/front-api/routes/w/[wId]/pods/[podId]/apps/delete.test.ts
@@ -1,0 +1,253 @@
+import { ensurePodSandboxReady } from "@app/lib/api/sandbox/lifecycle";
+import type { Authenticator } from "@app/lib/auth";
+import { SandboxFunctionResource } from "@app/lib/resources/sandbox_function_resource";
+import { SandboxResource } from "@app/lib/resources/sandbox_resource";
+import type { SpaceResource } from "@app/lib/resources/space_resource";
+import { FileFactory } from "@app/tests/utils/FileFactory";
+import { createPrivateApiMockRequest } from "@app/tests/utils/generic_private_api_tests";
+import { fileStorageMock } from "@app/tests/utils/mocks/file_storage";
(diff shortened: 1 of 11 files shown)


id13 (approved):
LGTM

PR #30376, opened 2026-08-11
[GitHub] chore: remove OpenAPI docs deployment workflow

aubin-tchoi:
## Description

Mintlify now reads the OpenAPI specification directly from the public `dust-tt/dust` repository, so the legacy manually dispatched deployment workflow is no longer needed.

This removes the workflow and its ReadMe.io upload path entirely to avoid any confusion.

New process is to force deploy the doc from Mintlify (shared credentials).

## Tests

- N/A.

## Risk

- N/A.

## Deploy Plan

- No deploy.


FlagBenett (approved):
LGTM

PR #30377, opened 2026-08-11
[front] enh: display skills in conversation credit breakdown

aubin-tchoi:
## Description

Conversation credit breakdowns show agent work, tools, and models, but not the skills responsible for tool calls.

This adds skills to conversation consumption details using the same attribution rule as consumption analytics, including successful `enable_skill` calls. The credits panel renders each skill's icon, name, and attributed credits between tools and models. Skill credits can overlap when several skills are responsible for the same tool call, and the new response field remains optional for backward compatibility.

## Tests

- Added conversation aggregation coverage for skill attribution.
- Updated credit breakdown rendering coverage.

## Risk

- Low. Limited to the flag-gated conversation consumption details flow.

## Deploy Plan

- Deploy front and front-api.

### DIFF
--- a/front-api/routes/swagger_private_schemas.ts
+++ b/front-api/routes/swagger_private_schemas.ts
@@ -185,6 +185,24 @@
  *         attributedCredits:
  *           type: number
  *           description: Model attribution after reconciling exclusively through its input rows.
+ *     PrivateConversationConsumptionSkillDetails:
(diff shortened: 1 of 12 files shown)


aubin-tchoi:
not planned, will be read from ES for the convo view, for the message view we'll keep it light and omit skills there

PR #30379, opened 2026-08-11
feat(sandbox): lock-scoped conversation scope resolution + strict sandbox destroy on moves

smb2268:
## Description

Follow up of `30386` (stacked on it), which makes every sandbox create/wake resolve the conversation's pod association inside the lifecycle lock. This PR is the move-side half: moving a conversation did nothing to its **existing** sandbox, which kept the previous pod's token, env vars, and secrets for as long as it stayed warm (env is injected at creation only, so even a sleep/wake cycle would not fix it).

**Moves become strict scope transitions** — no new state, no schema change; the lifecycle lock, `killRequestedAt`, and the deleted-state recreate path do all the work. Both move functions run through `withScopeTransition`, one lock hold covering three phases:

1. **prepare** — the conversation is re-fetched under the lock and source-pod validation runs against that fresh state (never the caller's snapshot, so two concurrent moves cannot both validate against the same source). A validation failure aborts with the runtime untouched — a rejected move cannot grief the sandbox.
2. **strict destroy** — `killRequestedAt` → provider destroy → mark deleted.
3. **commit** — the DB association change, using the lock-fetched conversation. The move's transaction is created and committed inside the lock hold; the previous optional outer-transaction parameter is gone (a commit deferred past the lock release would reopen the race the lock closes).

A concurrent Computer command can neither keep a pre-move sandbox alive nor create one from the pre-move scope (scope resolution holds the same lock, per the base PR); its next invocation recreates the sandbox in place (row and owner link survive) with the new pod's claims, env, and mounts. The conversation's egress policy file and GCS files are preserved.

**Crash/failure matrix** (all recovery via existing machinery):

| Failure | Outcome |
| --- | --- |
| Crash after kill-request, before provider destroy | Next access hits the kill-requested branch (or the reaper) and finishes the destroy + recreate |
| Provider destroy fails | Move fails (fail closed, association unchanged); the kill request left behind lets the next access complete the reset in the still-current scope |
| Crash after destroy, before the DB move commits | Sandbox gone, conversation unmoved — next access recreates the old scope |
| No sandbox provider configured (dev/test) | Transition runs with a vacuous runtime destroy — nothing can be running |

Sandboxes already running at deploy keep their pre-cutover token until they naturally sleep/wake (fail closed: their old scope is a strict subset of the new one, so nothing over-grants — at worst a new conversation-scoped approval is inert on a warm sandbox until it cycles). With pods feature-flagged the affected population is a handful at most; no migration or sweep is warranted.

## Risk

Worst case, a provider destroy outage makes moves fail while it lasts (fail closed, nothing over-grants, sandboxes recover via the kill request on next access). Safe to rollback: no schema change, no new state.

## Deploy Plan

- [ ] Deploy front with (or after) the base `30386` — the strict destroy assumes creates/wakes resolve scope under the same lock

## Tests

Tested manually all conversation move scenarios

Resource layer pins the transition machinery: prepare → destroy → commit ordering under the lock, a failed validation leaving the runtime untouched (no destroy, no kill mark), destroy failure leaving the kill request and failing the operation, provider NotFound proceeding, no-sandbox and no-provider paths, and both callbacks receiving the lock-fetched conversation. Move tests assert the sandbox ends deleted + kill-requested on both move directions, and that validation runs against the conversation's current state rather than a stale caller snapshot. (Scope-resolution coverage lives in the base `30386`.)

Drive-by 🚗: the global test-redis mock never returned `"OK"` for `SET NX`, so any suite exercising `executeWithLock` (including the selected-spaces suite once moves took the lock) spun the acquire loop into a 30s timeout — the mock now honors the NX/PX/unlock contract, which is what the two "selected Spaces" CI failures were. (`executeWithLock` is mocked in the move suite — Redis-less test env; its semantics are pinned at the resource layer.)

### DIFF
--- a/front/lib/api/projects/conversations.test.ts
+++ b/front/lib/api/projects/conversations.test.ts
@@ -25,19 +25,7 @@ import { isPodConversation } from "@app/types/assistant/conversation";
 import type { ModelId } from "@app/types/shared/model_id";
 import type { WorkspaceType } from "@app/types/user";
 import { Op } from "sequelize";
-import { beforeEach, describe, expect, it, vi } from "vitest";
-
-// The scope-transition lock (executeWithLock) needs Redis, which these
-// DB-level tests don't run; its ordering semantics are pinned in
-// sandbox_resource.test.ts. Here the callback just runs.
-vi.mock(import("@app/lib/lock"), async (importOriginal) => {
-  const actual = await importOriginal();
-  return {
-    ...actual,
-    executeWithLock: async <T>(_lockName: string, callback: () => Promise<T>) =>
-      callback(),
-  };
-});
+import { beforeEach, describe, expect, it } from "vitest";
 
 async function fetchRegularAutoGroup(
   space: SpaceResource,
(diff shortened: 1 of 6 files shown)


fontanierh (approved):
LGTM

PR #30380, opened 2026-08-11
feat(sidekick): use Fast auto stream on first turn, Standard on follow-ups

Nils-Fedrigo:
## Description

closes https://github.com/dust-tt/tasks/issues/10044

The Agent Builder Sidekick previously selected concrete models inline: Haiku (small whitelisted model) on the first turn and the workspace's large whitelisted model on follow-ups.

This routes it through the Auto tier streams instead:

- **New-agent-from-scratch first turn** → unchanged (`NOOP_MODEL_CONFIG`, static response, no LLM call).
- **Any other first turn** → the **Fast** stream (`auto_fast`).
- **Follow-up turns** → the **Standard** stream (`auto`).

Both sentinels are resolved to a concrete model + reasoning effort at message-send time by `resolveModel`, which already routes stream ids directly through `getModelForStream` regardless of the `models_picker` flag. So Sidekick gets the tiered routing in every workspace without depending on the picker being enabled.

## Tests

Covered by the existing `resolveModel` suite (`front/lib/api/assistant/models.test.ts`), which already asserts that a stream-configured agent resolves through `getModelForStream` (including without `models_picker`). Sidekick sets its model to `auto_fast` / `auto`, which flow through that same path. `tsgo --noEmit` clean.

### DIFF
--- a/front/lib/api/assistant/global_agents/configurations/dust/sidekick.ts
+++ b/front/lib/api/assistant/global_agents/configurations/dust/sidekick.ts
@@ -6,11 +6,6 @@ import type {
   PrefetchedDataSourcesType,
 } from "@app/lib/api/assistant/global_agents/tools";
 import { dummyModelConfiguration } from "@app/lib/api/assistant/global_agents/utils";
-import {
-  getLargeWhitelistedModel,
-  getSmallWhitelistedModel,
-  selectEnabledModel,
-} from "@app/lib/api/assistant/models";
 import type { Authenticator } from "@app/lib/auth";
 import type {
   AgentConfigurationType,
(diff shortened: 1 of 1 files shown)


fabiencelier (approved):
Nice 
This will effectivevly migrate sidekick to Luna high for 95% of the usage
What if someone does not have access to any standard model, does it automatically downgrade to a fast ?

fabiencelier (changes requested):
Actually, let's eval to see the perf of other models on sidekick befor emerging

fabiencelier (approved):

fabiencelier:
Closes https://github.com/dust-tt/tasks/issues/10044

PR #29247, opened 2026-07-21
[front][sparkle] Redesign input bar menu: unified + dropdown, model on the right, mic/send toggle

pinotalexandre:
## Summary
- Consolidate the capabilities, attachments, and a new shell-only spaces picker behind a single "+" dropdown next to the agent pill (styled to match the pill's border/shadow treatment).
- Move the model picker to the right side of the input bar.
- Collapse the voice and send buttons into a single slot: mic while idle/empty (or mid-recording), swapping to send once there's text to submit.
- The browser extension's input bar keeps its previous separate attachment/capture flow and left-side model picker unchanged.
- `CapabilitiesPicker` and the new `InputBarSpacesPicker` gained a `type: "dropdown" | "subdropdown"` prop (mirroring the existing pattern on `InputBarAttachmentsPicker`) so they can nest as `DropdownMenuSub` entries inside the new "+" menu.
- Mirrored the whole redesign in the sparkle Storybook `Composer` story (mock capabilities/attachments/spaces + model picker + mic/send toggle).

## Test plan
- [x] `npx tsgo --noEmit` — no new type errors
- [x] `npm run format:changed` — clean
- [x] Verified interactively in Storybook (`composer-composer--floating`): "+" opens with Capabilities / Attach knowledge / Spaces; mic icon swaps to send once text is typed and back to mic when cleared
- [ ] Not yet verified in the real running app (no local dev stack available this session — needs DB/ES/Temporal) — please sanity-check the actual conversation input bar, especially the Spaces shell and the extension's untouched flow, before merging

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #30381, opened 2026-08-11
[front] fix: Fix capabilities picker mounting issue

id13:
## Description

The tool config dialog opens and vanishes right away, so you can't configure a tool from
the input bar anymore. Broke this morning with #29247, web only, extension is fine.

The picker now sits inside the + menu content and renders CreateMCPServerDialog in its own
subtree. Radix closes the whole menu when you select an item and unmounts the content, so
the dialog goes with it.

So the picker just reports the server through onSetupServer now, and the creation flow
moved to CapabilitySetupDialog, mounted by InputBarButtons outside both menus.

## Tests

Added a test, it fails if the dialog goes back inside the menu

## Risk

Low
Blast radius: capabilities picker

## Deploy Plan

Deploy front

### DIFF
--- a/front/components/assistant/CapabilitiesPicker.test.tsx
+++ b/front/components/assistant/CapabilitiesPicker.test.tsx
@@ -1,0 +1,227 @@
+import {
+  CapabilitiesPicker,
+  CapabilitySetupDialog,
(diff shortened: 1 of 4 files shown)


smb2268 (approved):
LGTM - thanks for fixing!

PR #30382, opened 2026-08-11
Force Kimi K3 temperature to zero

fontanierh:
## Description

Force Kimi K3 requests through Fireworks to use temperature 0 at the Dust configuration boundary. Keep the provider schema unchanged because Fireworks accepts a wider range.

## Tests

Added a regression test and ran the live 46-case Kimi K3 Fireworks endpoint suite.

## Risk

Low. The change is limited to Kimi K3 request sampling and is safe to revert.

## Deploy Plan

Standard deploy.


flvndvd (commented):
LGTM

PR #30383, opened 2026-08-11
feat(egress-proxy): inherited pod policy layer via podId claim

smb2268:
## Description

Proxy half of the conversation-scoped egress approvals work (front cutover follows in #29432, same split as the egress relayout campaign: proxy capability first, front second).

Adds an optional `podId` claim to the egress JWT and one more layer to the policy evaluation: a domain is allowed if the workspace, owner, **pod**, or legacy sandbox policy allows it. The pod layer reuses the owner-file scheme (`w/{wId}/sandboxes/{podId}.json`) and cache, and fails closed like every other lookup. Tokens without the claim behave exactly as today — nothing mints `podId` until the front PR ships, so this deploys standalone with zero behavior change. Malformed shapes are rejected before evaluation: a signed token carrying `podId` without `wId` or `ownerId` — which no legitimate minter produces — is denied as `invalid_claims`, so a future minting regression can't pick up the pod's inherited allowlist.

## Risk

None on deploy: no tokens carry `podId` yet, and the new branch is unreachable until front mints the claim. Fail-closed lookup; a GCS error never grants. Safe to rollback any time before the front PR deploys.

## Deploy Plan

- [ ] Deploy egress-proxy after merge — must be live **before** #29432 (front) deploys. If front ships first, conversations inside Pods fail closed to workspace + own approvals until the proxy catches up (no over-granting).

## Tests

`podId` claim validation round-trip in jwt unit tests; five integration cases on the authorization branch: conversation-in-pod allowed via inherited pod policy, same token denied without the podId claim, pod-function sandbox allowed through the pod file as its owner file, a conversation's own policy not applying to a pod-function sandbox, and a `podId`-without-`ownerId` token denied even when the pod file allows the domain. 86 total green, fmt + clippy clean.

### DIFF
--- a/egress-proxy/src/connection.rs
+++ b/egress-proxy/src/connection.rs
@@ -162,6 +162,22 @@ async fn handle_connection_inner(
         }
     };
 
+    // A podId only ever accompanies a conversation-in-pod token, which always
+    // carries wId and ownerId. Reject the impossible shapes before policy
+    // evaluation: the proxy is the authorization boundary, and a malformed
+    // (signed) token must not pick up the pod's inherited allowlist through
+    // a future minting regression.
+    if token.pod_id.is_some() && (token.w_id.is_none() || token.owner_id.is_none()) {
+        deny(
+            stream,
+            DenyReason::InvalidClaims,
+            Some(&token),
+            Some(&request),
+        )
+        .await;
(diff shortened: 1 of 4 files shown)


fontanierh (approved):
LGTM

PR #30386, opened 2026-08-11
feat(sandbox): resolve conversation scope inside the lifecycle lock

smb2268:
## Description

Follow up of `29432`. A conversation's pod association is an authorization input — it drives the sandbox's egress claims, pod env vars, and pod mounts — but it was derived from state read **before** the sandbox lifecycle lock (typically the agent loop's conversation snapshot). A move committing between that read and the lock acquisition could create or wake a sandbox in its pre-move pod: the classic case is an agent moving its own conversation mid-loop and invoking the Computer again with its stale snapshot.

**What this PR does:**

1. `ensureActive` gains a `resolveScope` hook that runs as the **first step inside the lifecycle lock**; the conversation adapter implements it as a fresh DB read of `spaceId`. Since a scope transition (a move) holds the same lock, nothing derived from the resolved scope can be invalidated by a concurrent move.
2. The lock-resolved scope feeds the env-var factory (creation) and is returned on `EnsureSandboxResult`, so the ready path derives the filesystem, runtime owner, and egress claims from it — never from the caller.
3. Callers' conversation objects are trusted for **identity only**: the adapter's input type no longer accepts a `spaceId`, making the stale-snapshot pattern unrepresentable at compile time.
4. Pods resolve a constant scope (they cannot move), which keeps `requireRunning`'s lock-free fast path sound.

The move-side half — strictly destroying the sandbox when the association actually changes — is the stacked `30379`.

> [!NOTE]
> This PR previously carried an `egressTokenScope` fingerprint column (persistent per-sandbox scope state + re-mint-on-mismatch). Review concluded the lifecycle lock and existing columns already provide the guarantee for strict destroy/recreate semantics, so the schema change is gone — this diff replaces it, no migration anywhere in the stack.

## Risk

Worst case, the under-lock conversation fetch fails and sandbox bring-up errors (fail closed — no sandbox runs with unverified scope). One extra indexed read per ensure call, on a path that already does a provider round-trip. Safe to rollback: no schema change, no new state.

## Deploy Plan

- [ ] Deploy front (no ordering constraints)

## Tests

Adapter level: a conversation moved in the database gets the new pod's env vars with no `spaceId` accepted from the caller. Lifecycle level: the ready path derives egress claims, runtime owner, and mounts from the returned scope and ignores stale caller snapshots (both directions: pod gained, pod lost).

### DIFF
--- a/front/lib/api/sandbox/lifecycle.test.ts
+++ b/front/lib/api/sandbox/lifecycle.test.ts
@@ -106,16 +106,13 @@ vi.mock("@app/logger/logger", () => {
   return { default: logger };
 });
 
-import { Authenticator } from "@app/lib/auth";
-import { ConversationResource } from "@app/lib/resources/conversation_resource";
+import type { Authenticator } from "@app/lib/auth";
 import type { SandboxResource } from "@app/lib/resources/sandbox_resource";
 import { AgentConfigurationFactory } from "@app/tests/utils/AgentConfigurationFactory";
 import { ConversationFactory } from "@app/tests/utils/ConversationFactory";
 import { createResourceTest } from "@app/tests/utils/generic_resource_tests";
 import { SandboxFactory } from "@app/tests/utils/SandboxFactory";
-import { SpaceFactory } from "@app/tests/utils/SpaceFactory";
 import type { ConversationType } from "@app/types/assistant/conversation";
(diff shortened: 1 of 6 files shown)


fontanierh (approved):
LGTM

PR #30387, opened 2026-08-11
[dsbx] Use named stdout file descriptor

fontanierh:
## Description

Follow-up to #30344. Replace the magic stdout file descriptor `1` with `process.stdout.fd`.

## Tests

`bun test runner.test.ts`

## Risk

Low. This resolves to the same file descriptor and keeps the synchronous write path unchanged.

## Deploy Plan

Standard deploy.


davidebbo (approved):
LGTM

PR #30393, opened 2026-08-11
dsbx: resolve offloaded tool output under tools --json

fontanierh:
## Description

`dsbx tools --json` was emitting truncated tool output. When a result block goes past front's 20KB text threshold, front archives the full content to the pod filesystem and hands back an 8KB snippet plus a `[Full content archived at <path>]` sentence, with a machine-readable descriptor on the block's `_meta` (`tt.dust/offload`). `--json` printed that verbatim, so a machine contract was shipping bodies that do not parse (prod repro: a 121,700-byte result arrived as an 8,157-byte fragment).

The truncation is context-window protection for a model. `--json` exists to be parsed, so it is the wrong boundary for it.

Under `--json` only, each block carrying the descriptor now gets its archive read back from the `/files` mount and substituted into its text (`resource.text`, or top-level `text`), with the same bounded retry as `pod/tool_output.ts` for gcsfuse metadata staleness. If the read never lands, it fails instead of degrading back to the snippet, which is the bug class this removes. The descriptor is dropped from the resolved block since the inline text is no longer a snippet.

Default (non-`--json`) mode is untouched: it is model-facing and keeps the snippet plus the archive sentence, which is how the model learns where to `cat` the rest.

New in the `tools --json` stdout envelope: `code: tool_output_unavailable` (retryable, no `status`) with exit code 15. Both are append-only additions to the existing contract.

The archive path is confined to the mount root. `_meta` is not fully trusted input: front passes some remote MCP resource blocks through verbatim, so a third-party server can forge a descriptor. Nothing privileged is crossed (dsbx runs as the workload user), but a forged path should not aim the read outside the file mounts.

This also makes the Rust path the single implementation: `@dust/pod`'s `tools.call()` will delegate to `dsbx tools --json` rather than reimplementing the POST+poll protocol in TypeScript, so archive resolution lives in one place.

## Tests

Unit tests cover both block shapes, passthrough for blocks with no descriptor, mixed results, a file that only appears on a later attempt, a malformed descriptor, path traversal out of the mount, an unreadable (non-missing) archive failing fast, and a missing archive erroring instead of emitting the snippet. One fixture is a >20KB JSON payload with a snippet cut mid-JSON, asserting the resolved text parses and carries no truncation marker. Also asserted that plain-text rendering still prints the snippet and archive sentence.

## Risk

Low, and confined to `--json`. Two behavior changes worth naming: a call that used to return truncated-but-successful output can now fail with the typed error after ~5s of retries, and that failure also reaches an agent shelling out `dsbx tools --json` from the Computer, not only programmatic callers. The agent gets the error envelope on stdout with the path in the message. Rollback is a version pin.

## Deploy Plan

Needs a dsbx release and a `DSBX_CLI_VERSION` bump to reach pods. No front change.

### DIFF
--- a/cli/dust-sandbox/src/commands/tools/offload.rs
+++ b/cli/dust-sandbox/src/commands/tools/offload.rs
@@ -1,0 +1,461 @@
+// Resolution of offloaded tool output content blocks.
+//
+// Tool outputs above front's offload threshold are not delivered inline: the
+// full content is archived as a pod file and the inline text is replaced by a
+// snippet ending with a human-facing "[Full content archived at <path>]"
+// sentence. That truncation exists to protect a model's context window, so it
+// stays in the default (model-facing) rendering — but `--json` is a machine
+// contract with no model on the other end, and emitting a cut-off body there
+// hands consumers invalid data (JSON tool output no longer parses).
+//
+// Offloaded blocks carry a machine-readable descriptor under the
+// "tt.dust/offload" key of their `_meta` (front owns the write side in
+// `front/lib/actions/mcp_execution.ts`; the descriptor shape is append-only).
+// This module is the read side: blocks without a descriptor pass through
+// untouched, blocks with one get the archived file read back from the gcsfuse
+// mount and substituted into their text. Never parse the snippet text or the
+// archive sentence: the descriptor is the contract.
+//
+// The TypeScript counterpart is `resolveToolTextContent` in
+// `cli/dust-sandbox/pod/tool_output.ts`; both must keep the same semantics.
+
+use std::fmt;
+use std::path::{Path, PathBuf};
+use std::time::Duration;
+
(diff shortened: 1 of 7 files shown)


cli/dust-sandbox/src/commands/tools/offload.rs
```diff
@@ -0,0 +1,569 @@
...
+
+    Ok(path)
+}
+
+async fn read_archived_content(
+    path: &Path,
+    options: &ResolveOptions,
+) -> Result<String, OffloadResolutionError> {
+    // The archived file is written through the GCS API while this reads it
+    // through the gcsfuse mount, whose metadata cache can lag behind the
```
davidebbo, line 195: I assume this will keep working unchanged if we switch to an alternate fuse impl?

davidebbo (commented):
LGTM

PR #30394, opened 2026-08-11
Fix Pod mount freshness and isolate warm invocation tokens

fontanierh:
## Description

- Disable GCSFuse list, metadata, and negative metadata caches on shared Pod and conversation mounts so writes from another sandbox or Front are visible immediately. Independent GCSFuse clients previously retained directory listings and metadata for 60 seconds, plus negative lookups for 5 seconds; only the client performing a local mutation invalidated its entries.
- Make warm function execution an explicit fast-function opt-in. Durable functions stay cold while they may shell out to tool clients.
- Remove the sandbox token and Pod user identity at the warm worker process boundary. Warm handlers receive the current values only through their per-invocation context. This closes the Bun behavior where nested processes inherited the worker's native startup environment even after JavaScript deleted the corresponding `process.env` entries.
- Leave the generated Pod Functions guidance and image-generation context behavior unchanged for their separate work streams.

## Tests

- `npm test -- lib/api/file_system/sandbox/gcs_sandbox_mount_adapter.test.ts lib/resources/sandbox_function_invocation_resource.test.ts` (53 tests)
- `bun test` in `cli/dust-sandbox/functions-runner` (125 tests)
- `cargo test commands::function::warm::tests -- --nocapture` (12 tests)
- `cargo clippy --all-targets -- -D warnings`
- `cargo fmt --all -- --check`
- Biome on the four changed Front files

## Risk

- Disabling GCSFuse caches increases metadata and listing requests. Monitor GCS request volume and mount latency; the flags are independently reversible.
- The Front opt-in must not reach any sandbox running an older dsbx that still inherits invocation secrets into warm workers.

## Deploy Plan

1. Publish the new dsbx and deploy the corresponding sandbox image. With the old Front, the new binary receives no opt-in flag and defaults every function to cold execution.
2. Recreate every existing Pod and conversation sandbox, including sleeping sandboxes, and verify no old-dsbx sandbox remains. This first recycle establishes the token boundary before Front can enable fast-function warming.
3. Deploy Front, which opts only fast functions into warming. Durable functions remain cold.
4. Recreate or remount Pod and conversation sandboxes again so Front supplies the uncached GCSFuse flags. Wake alone preserves existing mounts and is insufficient.
5. Monitor GCS request volume and latency, warm versus cold function-run metrics, and sandbox token errors.

### DIFF
--- a/front/lib/api/file_system/sandbox/gcs_sandbox_mount_adapter.ts
+++ b/front/lib/api/file_system/sandbox/gcs_sandbox_mount_adapter.ts
@@ -69,10 +69,11 @@ function tokenUrl(index: number): string {
  * Per-target mount profile.
  *
  * - "workload": root-mounted with `allow_other` so the unprivileged sandbox
- *   users can access it; permissive file/dir modes; 60s kernel list cache
- *   (read-mostly workloads). All agent-facing mounts.
+ *   users can access it; permissive file/dir modes. All agent-facing mounts
+ *   are shared mutable filesystems, so namespace and metadata caching are
+ *   disabled: writes from another sandbox or Front must be visible immediately.
  * - "pod_sandbox_functions": same access model as "workload", but without
- *   kernel list caching so newly published functions are visible immediately.
+ *   caching so newly published or replaced functions are visible immediately.
  * - "pod_state_replica": mounted AS `dust-state` (via runuser) so the FUSE
  *   default — only the mounting user can access the fs — makes it invisible to
  *   every other uid, including the untrusted workload uid 1003 and root. No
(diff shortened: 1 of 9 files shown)


front/lib/api/file_system/sandbox/gcs_sandbox_mount_adapter.ts
flvndvd, line 427: I thought conversation was purely local, did we change this? Is it being mounted on more than one sandbox? If no, any chance we can configure this at the target level and apply the appropriate cache flags?

flvndvd (approved):

front/lib/api/file_system/sandbox/gcs_sandbox_mount_adapter.ts
fontanierh: Yep that's fair

fontanierh: Actually not really. Front writes uploads, tool outputs, and file operations directly to the same GCS prefix so we have the problem there as well.

We can't use caching at all :/

flvndvd: Damn.

PR #30396, opened 2026-08-12
fix: prevent "Today (partial)" reference line label from being clippe…

avervaet:
## Description

Until now, the “Today (partial)” reference-line label in the workspace consumption chart could be clipped because the chart’s top margin was too small.

**Before:**
Capture d’écran 2026-08-12 à 09 51 50

**After:**
Capture d’écran 2026-08-12 à 09 52 08

**Centralize chart layout constants for better modularity**

Chart height and margin values were previously duplicated and hardcoded across individual chart components, leading to clunky, ad-hoc imports between files. This PR centralizes those shared constants in front/components/charts/constants.ts so charts pull from a single source instead of reaching into each other or repeating values.

Updates the observability, Poke analytics, and workspace analytics charts to consume the shared constants
This is purely a layout and constants-refactoring change, chart data, series, legends, and interactions are unchanged.

## Risk

This is a frontend-only change with no backend, data-model, or migration impact. The main risk is a visual regression in one of the charts touched by the shared-constant import updates. Existing chart dimensions and margins are preserved except for the intentional consumption-chart top-margin increase, and the change can be safely reverted without cleanup.

## Deploy Plan

- Deploy front

### DIFF
--- a/front/components/agent_builder/observability/charts/DatasourceRetrievalTreemapChart.tsx
+++ b/front/components/agent_builder/observability/charts/DatasourceRetrievalTreemapChart.tsx
@@ -1,7 +1,4 @@
-import {
-  CHART_HEIGHT,
-  CONVERSATION_FILES_AGGREGATE_KEY,
-} from "@app/components/agent_builder/observability/constants";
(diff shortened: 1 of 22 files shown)


adrsimon (approved):
clean

PR #30397, opened 2026-08-12
[front] feat: resolve app-relative pod function references in the Frame host

davidebbo:
## Why

Groundwork for cloning and renaming Pod apps. A Frame addresses its functions by a fully qualified `<podId>/<slug>`, and the slug carries the app prefix — so a copied app's Frame keeps calling the **original** app's functions, and therefore reads and writes the original's data.

Renaming an app has the same root cause. Fixing it once unblocks both.

## What changes

A second reference form — a bare `<name>` with no `/`, resolved against the app folder the calling Frame lives in:

```tsx
usePodFunction("list-notes", { threadId })        // relative: resolves within this app
usePodFunction(`${podId}/tasklist__list-notes`)   // absolute: unchanged
```

`list-notes` inside `pod-x/TaskList/TaskList.tsx` means `pod-x/tasklist__list-notes`, so the same source follows whichever app it ends up in. Copying an app then needs no Frame edit at all.

The two forms are unambiguous by construction: absolute always contains `/`, relative never does. A prefixed-but-podless `tasklist__add-task` is deliberately **not** relative — dropping the prefix is the point, and a third form would just be another thing to reason about.

## "Only frames in apps", for free

The scope falls out of the Frame's own path, so nothing needs an allowlist:

| Frame | Scope | Bare references |
|---|---|---|
| In an app folder (`pod-x/TaskList/TaskList.tsx`) | derived from its path | resolve |
| At the Pod root | none | refused, `not_supported` |
| In a conversation, not from a Pod | none | refused, `not_supported` |

**Every host that can render a Pod app Frame passes `framePath`**, which took three different answers:

- `PodFrameSheet`, `PodFrameTabContent`, `PodPinnedBanner` — already hold the path.
- `FrameRenderer` (conversation side panel) — already computed a `framePath` from the pod file listing and simply never passed it on.
- `PublicFrameRenderer` (`/share/frame/<token>`) — had nothing to derive from, since the public response returns a `FileType` carrying neither a path nor `useCaseMetadata`. The response now includes the Frame's scoped path.

`VisualizationBlock` and the blocked-action cards pass nothing and behave exactly as before.

### The shared-Frame case is easy to get wrong

A shared Frame *does* live in an app folder — only the host was ignorant of it. Getting this wrong is not a subtle degradation: the Frame throws `SandboxFunctionCallError` on its first call, so an app works from the Apps tab and breaks the moment it is shared, which is exactly when it has an audience.

The new `framePath` field is **withheld from viewers who cannot read the Pod**. They cannot invoke a pod function anyway, so sending it would hand out the Pod's layout for nothing — and note that with relative references the Frame's source no longer contains the pod id, so this keeps that closed rather than reopening it. Additive and optional on a public endpoint, per BACK12.

## Rollout order is load-bearing

1. **front** (this PR) — host resolution and the shared normalizer. Inert.
2. **viz** — widen `POD_FUNCTION_REFERENCE_REGEX`, pass the bare form through.
3. **skill** — teach the agent to use relative references inside an app folder, and bump the version. This is what turns adoption on, so it gates everything behind both halves being live.

## Risk

Low. Purely additive: with no `framePath` the resolver returns absolute references untouched, which is every existing call site. Nothing emits the relative form until steps 2 and 3 land.

## Tests

Two functional tests on the share endpoint: the scoped path is returned to a viewer who can read the Pod, and withheld from an external granted viewer who cannot.

16 unit tests on the contract: scope derivation (app folder, nested Frame, Pod root, conversation, unnormalizable folder), resolution (bare expansion, absolute pass-through with and without scope, refusal without scope, refusal of malformed references), the prefix agreement with publish, and the viz-grammar dependency above. Existing suite green: 93 tests across `types/api`, `lib/api/sandbox_functions/slug` and `lib/api/projects`. All three workspaces typecheck.

### DIFF
--- a/front-api/routes/v1/public/frames/[token]/index.test.ts
+++ b/front-api/routes/v1/public/frames/[token]/index.test.ts
@@ -10,6 +10,7 @@ import type { UserResource } from "@app/lib/resources/user_resource";
 import { FileFactory } from "@app/tests/utils/FileFactory";
 import { createResourceTest } from "@app/tests/utils/generic_resource_tests";
 import { MembershipFactory } from "@app/tests/utils/MembershipFactory";
+import { SpaceFactory } from "@app/tests/utils/SpaceFactory";
 import { UserFactory } from "@app/tests/utils/UserFactory";
 import type { FileShareScope } from "@app/types/files";
 import { frameContentType } from "@app/types/files";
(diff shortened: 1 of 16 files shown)


fontanierh (approved):
nice, LGTM

PR #30399, opened 2026-08-12
[front] enh: use Elasticsearch skill search in the slash menu

aubin-tchoi:
## Description

This PR moves the conversation slash menu to use the skill search endpoint behind the `skills_search` FF instead of loading every active skill and searching client-side.

Skills keep the order returned by ES, followed by matching tools. The suggestions update together when the search returns.

No change when the flag is off.

## Tests

- Added ranking tests.
- Not tested locally.

## Risk

- Low. Behind FF.

## Deploy Plan

- Deploy front.


aubin-tchoi:
PMRR

sfriquet:
lgtm

PR #30402, opened 2026-08-12
[sparkle][front] Highlight the model picker for discovery

pinotalexandre:
## Description

Temporary treatment drawing attention to the model picker, from [this Figma](https://www.figma.com/design/wJJMfVF6bluurSKfrEuysc/Product---WIP?node-id=10416-39566):

- a blue ring that **pulses twice every 7s**
- **two diagonal light streaks** sweeping across right after each pulse
- **one extra sweep on hover**

Values are the Figma's: Blue/200 ring at radius 8, Blue/50 streaks 3px wide at 30° with a 1px blur.

Pulse and sweep each run as one 7s animation with the active part expressed as a percentage of that duration, so the streaks always follow the pulse instead of drifting out of step the way two hand-matched delays would. Both ends of the streaks' travel sit clear of the box, so they are never parked over the button — the effect only exists while it is moving.

Only `transform` is animated, on layers that are absolutely positioned and `pointer-events-none`: the button's box, layout and hit area are untouched, and removing the wrapper shifts nothing. Under `prefers-reduced-motion` the streaks and the pulse are dropped and the static ring carries the highlight alone.

### Lifecycle

| | |
|---|---|
| Click the picker | highlight retires for the rest of the visit (the dropdown still opens normally) |
| Next page load | comes back, until the cap |
| Cap | **2 page loads per browser**, spent per load rather than per click so a user who never clicks still stops seeing it |
| Expiry | hard-coded at **2026-08-26**, so the campaign dies even if the code outlives it |
| Surface | web only — the extension has its own localStorage and would spend a second pair of views on the same person |

Wrapped inside `InputBarModelPicker` rather than at its call sites, so the input bar gets it while the model pickers in the agent builder and app blocks stay untouched.

## Tests

Driven frame by frame in a Sparkle playground replica of the input bar:

- **Cycle** — ring at `scale 1.08` at 200ms and 600ms, streak over the button from 1200ms to ~1500ms, nothing from 1640ms to 7s.
- **No persistence** — streaks fully clear of the button at both `0ms` and `840ms` of the pass.
- **Click** — ring and streaks unmount, picker stays mounted, dropdown opens with all four items.
- **Cap** — load 1 visible (count 1), load 2 visible (count 2), load 3 nothing.
- **Properties** — all three keyframes animate `transform` only; no layout property anywhere.

## Risk

Cosmetic and self-limiting: capped at two views, expires on a date, and no layout impact. If the merge slips well past the date in `useModelPickerHighlight.ts`, bump `CAMPAIGN_END` or the highlight will never show.

Note that `localStorage` is per browser, so someone on a laptop and a phone gets two views on each. Switching to server-side user metadata would fix that at the cost of an extra request per page load and a late pop-in; not worth it for a two-week nudge, but flagging the tradeoff.

## Deploy Plan

Nothing to coordinate. Delete `ModelPickerHighlight`, `useModelPickerHighlight` and the `glint-*` rules in `sparkle/src/styles/tailwind.css` once the campaign is over.


Nils-Fedrigo (approved):
lgtm! Would go for the following parameters instead of your but understandable:
- consumed when the user click on it (i fear that people load the page and then go away

PR #30403, opened 2026-08-12
[front] enh: teach the pod functions skill app-relative references

davidebbo:
> **Stacked on #30400, which is stacked on #30397.** The diff here is one file. Rebase onto `main` as the parents merge.
>
> **This is the step that turns adoption on.** Merge it LAST, and only once both parents are deployed — front resolves relative references, viz forwards them, and nothing emits them until this lands.

## Description

Step 3 of 3. #30397 taught the Frame host to resolve a bare function name against the app folder the calling Frame lives in; #30400 widened the viz grammar so such references reach the host. This tells the agent to use them.

A Frame inside an app folder now refers to its own app's functions by **bare name** — the same name passed to `publish`, no Pod and no app prefix. Fully qualified `<podId>/<slug>` stays required for reaching another app or another Pod, and for any Frame that does not live in an app folder.

## Risk

Low mechanically — one prompt file plus a version bump. The behavioural risk is entirely about ordering: if this merges before both parents are **deployed**, an agent writes a bare reference, viz rejects it, and the reference resolves to a null SWR key — a Frame that silently issues no request at all. Existing apps are unaffected either way, since absolute references keep working and nothing rewrites them.

Skill version 6 → 7.

## Tests

No new tests: the skill is prompt text, and the contract it now describes is already pinned on both sides — the front↔viz parity check in #30400 and the prefix-agreement test against `deriveSandboxFunctionSlug` in #30397. 130 existing skill tests green, front typechecks.

### DIFF
--- a/front/lib/resources/skill/code_defined/global/pod_functions.ts
+++ b/front/lib/resources/skill/code_defined/global/pod_functions.ts
@@ -93,13 +93,16 @@ skill covers creating it and moving it into the app folder with \`${FILES_MOVE_T
 
 Functions that no Frame calls still get an app folder, named after what they do together.
 
-**Copying an app folder does not finish the job.** The databases and the published slugs follow the
-new folder on their own, but a Frame addresses its functions by \`<podId>/<slug>\` written into its
-source, so the copy's Frame still calls the ORIGINAL app's functions and therefore reads and writes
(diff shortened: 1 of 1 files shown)


fontanierh (approved):
LGTM

PR #30404, opened 2026-08-12
Track every non-batch LLM call for billing reconciliation

flvndvd:
## Description

We are currently falling short on capturing all LLMs calls. This PR makes sure every non-batch LLM call is recorded, including calls that fail before returning usage or are cancelled while streaming.

A pending usage record is created before contacting the provider, then completed when token usage is received. If no
usage is reported, it is marked unavailable instead of silently disappearing. The record also stores the actual
inference provider and endpoint region, which are needed to reconcile our data with provider bills.

Batch calls are intentionally left unchanged and unresolved usage records remain excluded from billing totals. A follow up PR will come.

## Deploy Plan

 - [ ] Apply SQL migration
 - [ ] Deploy front

### DIFF
--- a/front/lib/resources/run_resource.ts
+++ b/front/lib/resources/run_resource.ts
@@ -4,14 +4,17 @@ import type { TokenUsage } from "@app/lib/api/llm/types/events";
 import type { Authenticator } from "@app/lib/auth";
 import { getModelConfigByModelId } from "@app/lib/llms/model_configurations";
 import type { UsageType } from "@app/lib/metronome/types";
+import type { Region } from "@app/lib/model_constructors/types/regions";
 import { BaseResource } from "@app/lib/resources/base_resource";
 import { AppModel } from "@app/lib/resources/storage/models/apps";
+import type { RunUsageState } from "@app/lib/resources/storage/models/runs";
 import {
   RunModel,
   RunUsageModel,
 } from "@app/lib/resources/storage/models/runs";
 import type { ReadonlyAttributesType } from "@app/lib/resources/storage/types";
 import type { ResourceFindOptions } from "@app/lib/resources/types";
+import { withTransaction } from "@app/lib/utils/sql_utils";
 import { getStatsDClient } from "@app/lib/utils/statsd";
 import logger from "@app/logger/logger";
 import { getRunExecutionsDeletionCutoffDate } from "@app/temporal/hard_delete/utils";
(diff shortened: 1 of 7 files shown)


front/lib/resources/run_resource.ts
```diff
@@ -537,6 +716,35 @@ export class RunResource extends BaseResource<RunModel> {
 
     return usages.map(({ runModelId, ...usage }) => usage);
   }
+
+  async listRunUsageAttempts(
```
flvndvd, line 720: Only use for tests

id13 (approved):

front/lib/resources/run_resource.ts
```diff
@@ -487,11 +554,133 @@ export class RunResource extends BaseResource<RunModel> {
...
+    return true;
+  }
+
+  async finalizePendingTokenUsage(
+    auth: Authenticator,
+    runUsageModelId: ModelId,
+    usage: TokenUsage,
+    modelId: ModelIdType,
+    {
+      inferenceRegion = "global",
```
fabiencelier, line 643: why defaulting to global here ? Isn't it a risk to miss the run usage if it was in another region ?

front/migrations/pre-deploy/20260812095255_add_llm_run_usage_lifecycle.sql
```diff
@@ -0,0 +1,20 @@
...
+SET SESSION statement_timeout = 3000;
+SET SESSION lock_timeout = 3000;
+ALTER TABLE "public"."run_usages" ADD COLUMN "inferenceProvider" character varying(255) COLLATE "pg_catalog"."default" DEFAULT NULL::character varying;
+
+/*
+Statement 2
+*/
+SET SESSION statement_timeout = 3000;
+SET SESSION lock_timeout = 3000;
+ALTER TABLE "public"."run_usages" ADD COLUMN "region" character varying(255) COLLATE "pg_catalog"."default" DEFAULT NULL::character varying;
```
fabiencelier, line 20: no backfill for these columns ?

fabiencelier (approved):

front/lib/resources/run_resource.ts
flvndvd: Will require some more plumbing to wire it everywhere especially on non message run. More PRs coming your way.

front/migrations/pre-deploy/20260812095255_add_llm_run_usage_lifecycle.sql
flvndvd: Yes, will need the help from data team as it needs to correlate with the switch of workspace to credit-based plan.

PR #30405, opened 2026-08-12
[analytics] refactor: use POST for consumption endpoints to lift filter cap

avervaet:
## Description

This PR moves the GET-based consumption analytics endpoints to POST requests with JSON bodies so filters are no longer constrained by URL length limits. The affected endpoints are `timeseries`, `overview`, and all `top-*` endpoints for agents, users, teams, models, tools, skills, and sources; the existing `facets` endpoint now uses the same shared body schema.

Until now, filters were JSON-encoded into the query string, which capped the number of selectable filter values. This PR replaces the query schemas with `ConsumptionBodySchema` and `ConsumptionTopBodySchema`, and sends numeric fields such as `days`, `limit`, and `breakdownCount` as JSON numbers rather than coerced query-string values.

The frontend consumption hooks now share a `useConsumptionQuery` implementation. It normalizes filter ordering for stable SWR cache keys, debounces checkbox-driven requests, aborts superseded requests, and preserves the existing disabled/loading/cache behavior across facets, overview, timeseries, and top-dimension widgets.

The existing manager authorization and analytics response handling remain unchanged. The HTTP contract for the migrated endpoints does change from GET/query parameters to POST/JSON bodies, so callers must use the updated request shape; the corresponding frontend hooks are migrated in this PR. No data model or persisted data behavior is changed.

## Risk

The main risk is compatibility for consumers still issuing GET requests or sending query-string values to the migrated endpoints; those requests will no longer match the updated route contract. The frontend hooks and route validation are changed together, and existing authorization, response processing, and persisted data remain unchanged. The shared hook also aborts superseded requests to reduce stale responses racing into the SWR cache. The change is otherwise a safe application-code revert with no migration or backfill indicated.

## Deploy Plan

- Deploy front

### DIFF
--- a/front-api/routes/w/[wId]/analytics/consumption/facets.ts
+++ b/front-api/routes/w/[wId]/analytics/consumption/facets.ts
@@ -2,7 +2,7 @@ import type { GetConsumptionFacetsResponse } from "@app/lib/api/analytics/consum
 import { fetchConsumptionFacets } from "@app/lib/api/analytics/consumption/facets";
 import { resolveConsumptionPeriod } from "@app/lib/api/analytics/consumption/period";
 import {
-  ConsumptionFacetsBodySchema,
+  ConsumptionBodySchema,
   toConsumptionPeriodInput,
 } from "@app/lib/api/analytics/consumption/schema";
 import { workspaceApp } from "@front-api/middlewares/ctx";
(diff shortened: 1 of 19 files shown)


avervaet:

- #30405 ◀
    - #30414

flvndvd (approved):

PR #30407, opened 2026-08-12
feat: add an analytics seed scenario for the consumption page

adrsimon:
## Description

`dust-hive feed <env> analytics` now populates the consumption index behind the "Analytics (new)" page, so it has data to show locally.

It seeds members, two overlapping teams, skills and enough agents for the rankings to have a tail, then writes LLM and tool consumption documents over the last 90 days. Dimension values are sampled from the workspace's own facet catalog, so the filter panel options and the indexed documents always match. Includes a hidden agent in a restricted space, to check consumption shows up for agents the admin cannot otherwise see.

Re-runnable: entities are skipped by name, and the documents have deterministic ids so a second run overwrites the first.

## Tests

Ran it against a local hive, dry and with `--execute`, and checked the overview, chart and all seven attribution tabs.

## Risk

None, dev-only seed script.

## Deploy Plan

Nothing to do.

### DIFF
--- a/front/scripts/seed/factories/seedConsumptionAnalytics.ts
+++ b/front/scripts/seed/factories/seedConsumptionAnalytics.ts
@@ -35,8 +35,7 @@ import type { SeedContext } from "./types";
  * panel offers are exactly the ones the seeded documents carry.
  */
 
-// Consumption keys are part of the document id, so a second run overwrites the
-// first run's dataset instead of doubling it.
+// Keeps a second run overwriting the first run's documents instead of doubling them.
 const SEED_CONSUMPTION_KEY_PREFIX = "seed-consumption";
 
 const DEFAULT_DAYS_BACK = 90;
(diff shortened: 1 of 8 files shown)


front/scripts/seed/factories/seedConsumptionAnalytics.ts
```diff
@@ -0,0 +1,616 @@
...
+ * panel offers — and turns it into the pools the documents are sampled from.
+ */
+async function loadConsumptionPools(
+  ctx: SeedContext
+): Promise<ConsumptionPools> {
+  const { auth } = ctx;
+
+  const catalog = await listConsumptionFacetCatalog(auth);
+
+  // The `team` dimension reads `user.group_ids`, which production fills with the
```
avervaet: adding logic in comments ages badly

front/scripts/seed/factories/seedConsumptionAnalytics.ts
```diff
@@ -0,0 +1,616 @@
...
+/**
+ * Reads the workspace's consumption facet catalog — the very list the filter
+ * panel offers — and turns it into the pools the documents are sampled from.
+ */
+async function loadConsumptionPools(
+  ctx: SeedContext
+): Promise<ConsumptionPools> {
+  const { auth } = ctx;
+
+  const catalog = await listConsumptionFacetCatalog(auth);
```
avervaet, line 160: It seems that listConsumptionFacetCatalog calls getAgentConfigurationsForView with agentsGetView: "all", which only includes agents with scopes workspace, published, or visible.

Therefore the hidden agent created by hiddenAgent.ts has scope "hidden", so it is never included in pools.agentIds. Consequently, the seed writes no consumption documents for SeedPrivateAgent, and the stated restricted-agent visibility scenario is not actually exercised.

We could either load the agent pool with an internal view that includes hidden agents or explicitly append the hidden agent to the pool. 

We might want to add a regression check should verify that at least one indexed document uses the hidden agent’s ID.

avervaet (approved):

front/scripts/seed/factories/seedConsumptionAnalytics.ts
adrsimon: need the other PR to change the "all" agent view, which is not merged, will send an edit afterwards but yeah you're right

PR #30414, opened 2026-08-12
[analytics] refactor: remove client-side cap on filter selections

avervaet:
## Description

This PR removes the client-side 50-selection limit from the analytics usage filter panel. It is a follow-up to [#30405](https://github.com/dust-tt/dust/pull/30405), which moves the consumption endpoints from GET query strings to POST JSON bodies and removes the URL-length constraint that the cap was designed to protect against.

Until now, the filter helpers rejected additional selections after 50 total values, truncated “Select all” operations to the remaining capacity, disabled unchecked options at the limit, and displayed “Selection limit reached”. This PR removes that cap and the related UI state so users can select any number of available values and “Select all” selects all unselected enabled options. Existing option availability and disabled-option behavior remain unchanged.

This is a client-side-only cleanup in the stacked analytics refactor. It does not change persisted data or consumption response and authorization behavior. The larger filter selections rely on the POST-based consumption endpoints from #30405.

## Tests

The existing `front/components/workspace/analytics/usageFilter.test.ts` coverage was updated to remove the obsolete selection-cap assertion.

## Risk

Low. The user-visible behavior changes for workspaces selecting more than 50 filter values: those selections are now allowed and sent through the POST request path. The main compatibility dependency is #30405; deploying this change without the POST migration would leave the old URL-length constraint in place for larger filters. No migration, backfill, or persisted-data change is included, and reverting this PR restores the previous client-side cap.

## Deploy Plan

Deploy this follow-up after or together with [#30405](https://github.com/dust-tt/dust/pull/30405) so consumption analytics requests use POST JSON bodies before the client-side cap is removed. No special deployment steps were identified from the supplied context.

### DIFF
--- a/front-api/routes/w/[wId]/analytics/consumption/facets.ts
+++ b/front-api/routes/w/[wId]/analytics/consumption/facets.ts
@@ -2,7 +2,7 @@ import type { GetConsumptionFacetsResponse } from "@app/lib/api/analytics/consum
 import { fetchConsumptionFacets } from "@app/lib/api/analytics/consumption/facets";
 import { resolveConsumptionPeriod } from "@app/lib/api/analytics/consumption/period";
 import {
-  ConsumptionFacetsBodySchema,
+  ConsumptionBodySchema,
   toConsumptionPeriodInput,
 } from "@app/lib/api/analytics/consumption/schema";
 import { workspaceApp } from "@front-api/middlewares/ctx";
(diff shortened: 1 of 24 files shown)


avervaet:

- #30405
    - #30414 ◀

flvndvd (approved):
🔥

aubin-tchoi (approved):

flvndvd:
We would still need one though, ES has a hard limit on max 1024 terms clauses.

avervaet:
> We would still need one though, ES has a hard limit on max 1024 terms clauses.

ES caps terms-query values via index.max_terms_count (default 65,536).
No limit set for now, unrealistic to expect 65K+ filter values here.

flvndvd:
> > We would still need one though, ES has a hard limit on max 1024 terms clauses.
> 
> ES caps terms-query values via index.max_terms_count (default 65,536). No limit set for now, unrealistic to expect 65K+ filter values here.

Yes, but it also has a lower 1024 limits on the leaf conditions, will need to check the query we build. 

UPDATE: Check and indeed, safe as we only have up to 9 leaves (our facets).

PR #30418, opened 2026-08-12
Instrument Google Drive content memory usage

fontanierh:
## Description

Instrument Google Drive file content processing before choosing a memory-based concurrency policy. This does not queue, delay, or limit any operation.

Emit `google_drive.content_processing.*` distributions and a structured `Google Drive content memory telemetry` log for each table or text-document operation. The telemetry includes stored file size, MIME type, duration, outcome, overlapping operation counts, and start, sampled peak, and end values for RSS, heap, external memory, and array buffers.

Memory is measured at the worker-process level. The overlapping-operation counts make concurrent samples identifiable when analyzing the logs.

## Tests

Added focused coverage that verifies concurrent tasks all start immediately and that file metadata and sampled memory are emitted.

## Risk

Sampling calls `process.memoryUsage()` once per second for each in-flight content operation and emits one log plus distributions when the operation finishes. There is no scheduling or throughput change.

## Deploy Plan

Standard deploy. Use the new telemetry alongside worker RSS and OOM restarts before selecting a memory or size-based guard.


flvndvd (commented):
LGTM

PR #30420, opened 2026-08-12
Reduce analytics parallelism

flvndvd:
## Description

We noticed that current concurrency on analytics queue workers (32) leads to some `OOMKilled`. See in DD [here](https://app.datadoghq.eu/notebook/340275/temporal-analytics-queue-performance). This PR reduces it to 24.


id13 (approved):
LGTM !

PR #30422, opened 2026-08-12
[front] fix: attribute a legacy Pod database only to the app that would open it

davidebbo:
## Description

The Apps tab could list the same database twice under one app, and list a database under an app that never touches it. Reproduced with a Pod holding:

```
counter.db                 ← created before app namespacing, so no prefix
counter__counter.db        ← the Counter app's
counter_copy__counter.db   ← Counter Copy's
```

Both apps showed **two** databases, both rendered `counter`, because the display name is the on-disk name with the app prefix stripped.

## Cause

A database created before app namespacing has a bare filename, so its name says nothing about which app owns it. Attribution fell back to the schema file that declares it — `<AppName>/databases/{name}.db.ts` — and credited **every** app declaring that name.

Declaring the name was never sufficient. `resolvePodDatabaseName` prefers the *prefixed* file whenever it exists, so an app that has one opens that, and the bare file is not its database at all.

That mattered little while a duplicate declaration was rare. It stops being rare as soon as apps can be copied — a copy inherits the schema file, so two apps declare the name while each opens its own prefixed file, leaving the bare one an orphan credited to both.

## Fix

Attribution now follows resolution: a bare database is credited to a declaring app only when that app has no prefixed database of the same name — exactly the precedence `resolvePodDatabaseName` applies at reconcile time. So the tab reports the database each app would actually open, which is what it should have reported all along.

| | Before | After |
|---|---|---|
| Counter | `counter__counter`, `counter` | `counter__counter` |
| Counter Copy | `counter_copy__counter`, `counter` | `counter_copy__counter` |
| Unfiled | — | `counter` |

A bare file no app would open now surfaces as **unfiled** rather than being double-counted, which is also more useful: it says there is a stray legacy database nobody reads.

`podDatabasePrefixFromAppPrefix` is exported from `db_naming.ts` so the check uses the same prefix-building authority as reconcile and invocation, rather than reconstructing the name.

## Scope

Read-only change to how the Apps tab attributes databases. Nothing is created, moved or deleted, and the underlying files are untouched — a mis-attributed database was only ever displayed in the wrong place.

Found while testing the app-clone PR, but it is a bug in the already-merged tab, so it ships on its own. Clone only made the rare case ordinary; landing this first means clone does not appear to introduce it.

## Tests

One regression test using the exact three-file layout above, asserting each app lists only the database it would open and that the orphan is unfiled. Existing suite green: 15 apps-endpoint tests, 20 `db_naming` tests.

### DIFF
--- a/front-api/routes/w/[wId]/pods/[podId]/apps/index.test.ts
+++ b/front-api/routes/w/[wId]/pods/[podId]/apps/index.test.ts
@@ -162,6 +162,46 @@ describe("GET /api/w/:wId/pods/:podId/apps", () => {
     ]);
   });
 
+  it("does not attribute a legacy database to an app that has its own prefixed one", async () => {
+    const { workspace, pod } = await setupPod();
+
+    // The state a clone leaves behind: the copy inherits the schema file, so both apps declare
+    // `counter`, but each has its own prefixed database and neither opens the bare legacy file.
+    fileStorageMock.setFilesByPrefix(() => [
+      gcsObject(workspace.sId, pod.sId, "Counter/"),
+      gcsObject(workspace.sId, pod.sId, "Counter/databases/counter.db.ts"),
+      gcsObject(workspace.sId, pod.sId, "Counter Copy/"),
+      gcsObject(workspace.sId, pod.sId, "Counter Copy/databases/counter.db.ts"),
+    ]);
+    fileStorageMock.setSubdirectoryNames(() => [
+      "counter.db",
+      "counter__counter.db",
+      "counter_copy__counter.db",
+    ]);
+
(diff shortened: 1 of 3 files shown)


davidebbo:
Superseded by #30423, which removes the inference this fixed rather than sharpening it. Attributing an unprefixed database to the app whose schema file declares it is guesswork that a clone makes ambiguous, so #30423 lists only namespaced databases and documents the two costs (an app whose only database predates namespacing shows none, and deleting such an app leaves it behind). Both go away with a migration that renames bare databases to their prefixed names, which is also what lets the step 2 fallback in resolvePodDatabaseName be deleted.

PR #30427, opened 2026-08-12
Fix unconsumed tool result attribution

flvndvd:
## Description

Tool consumption rows include the token footprint of tool results, assuming those results are consumed by a subsequent LLM run.

This is wrong when a message stops after executing tools. For example, with `gracefully_stopped`, the final tool results are produced and the tools are billed, but no following LLM run consumes their results.

As this "early exit" case is pretty rare, I did not want to overfit data model or event around this and kept it as an exit door.

This can make attributed costs exceed the authoritative message bill and cause analytics attribution to fail.

We are slightly changing the current approach or terminal messages:
- Order reported runs using the persisted run chronology.
- Keep tool-result footprints for every run except the last one.
- Record the last run’s tool-result footprint as zero.
- Preserve tool-call output tokens and direct tool charges.

Existing completed rows can only transition from a potential result footprint to zero. The transition is one-way, so retries
or stale non-terminal passes cannot restore it.

## Deploy Plan

No migration required to be run.

### DIFF
--- a/front/lib/resources/agent_message_consumption_item_resource.ts
+++ b/front/lib/resources/agent_message_consumption_item_resource.ts
@@ -270,9 +270,9 @@ export class AgentMessageConsumptionItemResource extends BaseResource<AgentMessa
   }
 
   /**
-   * Bulk-inserts final tools, then completes only conflicting rows that are still pending. The
-   * insert waits on concurrent writes to the unique action identity, so the following lookup sees
-   * the settled conflict state without requiring a lock or allowing a stale pending write to win.
+   * Bulk-inserts final tools, then resolves conflicts by completing pending rows or removing a
+   * terminal result footprint. The insert waits on concurrent writes to the unique action identity,
+   * so the following lookup sees settled state without a separate lock.
    *
    * // TODO(2026-08-01 flav): Revisit based on how often it happens.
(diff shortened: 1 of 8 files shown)


front/lib/resources/agent_message_consumption_item_resource.ts
```diff
@@ -319,61 +319,107 @@ export class AgentMessageConsumptionItemResource extends BaseResource<AgentMessa
...
+          "Removing a tool result cannot change its call or direct-charge evidence"
+        );
+        await this.model.update(
+          {
+            itemType: "tool",
+            agentMCPActionId: actionModelId,
+            inputTokensCount: 0,
+            outputTokensCount: record.outputTokensCount,
+            grossAttributedCreditAmountMicro:
+              record.grossAttributedCreditAmountMicro,
```
aubin-tchoi, line 406: `grossAttributedCreditAmountMicro` does not appear in the SELECT

aubin-tchoi (approved):
LGTM

front/lib/resources/agent_message_consumption_item_resource.ts
flvndvd: Yep it's `record`, not the `row`.

PR #30430, opened 2026-08-12
[front] enh: rework the Pod Apps tab into a tile grid

davidebbo:
Screenshot 2026-08-12 at 16 31 29

## Description

Reworks the Pod **Apps** tab from a list + detail split into a matrix of tiles, with the key actions available directly on each tile.

- Each tile is built on Sparkle's `Card`/`CardActionButton`: the whole card is the launch target and opens the app's Frame in `PodFrameSheet`. Per the single-frame assumption, the tile targets the app's first Frame; the card is non-interactive when there is no Frame or the Frame has no `FileResource` to render.
- Clone and delete are corner icon buttons revealed on hover/focus (always visible below the `sm` breakpoint, so touch devices are covered), still gated on `pod.isEditor && !pod.archivedAt` and wired to the existing dialogs.
- The tile shows the app icon (same pinned-tab icon logic as before), name, Frame filename, and Draft / Pinned-as-tab chips. The functions and databases listing is intentionally dropped for now — it will come back in a later iteration.
- The colliding-folders warning moved from the detail pane to `ContentMessage`s above the grid.
- `PodAppList` and `PodAppDetail` are deleted (no other consumers).

## Risk

Low — UI-only rework of the Apps tab; the frame sheet, clone and delete dialogs, and permission gating are unchanged.

## Deploy Plan

Deploy front.


fontanierh (approved):
very very cool

PR #30433, opened 2026-08-12
[front] fix: Surface the real reason a webhook trigger was blocked

id13:
## Description

Bubble up trigger errors caused by running out of credit to users.
This is done by introducing a new failure mode for triggers "credits_exhausted" to WebhookRequestTriggerStatus.
I won't backfill, this is not backward compatible, but at least users will get clearer errors in the future.

## Tests

N/A

## Risk

Low

## Deploy Plan

Deploy front

### DIFF
--- a/front/types/assistant/triggers.ts
+++ b/front/types/assistant/triggers.ts
@@ -85,6 +85,7 @@ export const WEBHOOK_REQUEST_TRIGGER_STATUSES = [
   "workflow_start_failed",
   "not_matched",
   "rate_limited",
+  "credits_exhausted",
 ] as const;
 
 export type WebhookRequestTriggerStatus =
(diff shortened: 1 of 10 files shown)


front/types/assistant/triggers.ts
```diff
@@ -85,6 +85,7 @@ export const WEBHOOK_REQUEST_TRIGGER_STATUSES = [
   "workflow_start_failed",
   "not_matched",
   "rate_limited",
+  "credits_exhausted",
```
id13, line 88: main change, the rest is mostly UI to display it + stats + typescript massaging

Not sure about the blast radius, looked alright

adrsimon (approved):

PR #30434, opened 2026-08-12
Admin governance: dual write + migration for skills

fabiencelier:
## Description

First part of https://github.com/dust-tt/tasks/issues/9472

Implement dual write of permissions for skills.
We actually just need to add the "editor" permission to the editor group upon creation and remove it on skill deletion.

Also include the migration to run post deploy 

## Tests

added unit test

## Risk

low: Can be reverted if needed, permissions are still stored and checked as before

## Deploy Plan

merge
deploy front
run post deploy migration 
```
npx tsx front/migrations/20260812_backfill_skill_group_permissions.ts --execute
```

### DIFF
--- a/front/lib/resources/group_permission_registry.ts
+++ b/front/lib/resources/group_permission_registry.ts
@@ -37,11 +37,11 @@ import assert from "assert";
  * grant applies to the whole type (or all types) and is only ever a type-level (`-1`) grant.
  */
 
-type ConcreteGrantType = Exclude<GrantType, "*">;
+export type ConcreteGrantType = Exclude<GrantType, "*">;
 
 // A grant applies either to a specific resource instance (resourceId > 0) or to the whole type
 // (resourceId = -1). A role declares the levels at which it can be granted.
-type GrantLevel = "instance" | "type";
+export type GrantLevel = "instance" | "type";
 
 interface RoleDefinition {
   verbs: GrantVerb[];
@@ -63,7 +63,7 @@ export const ROLE_REGISTRY: Record<
     publish: { verbs: ["publish"], levels: ["type"] },
   },
   skill: {
-    editor: { verbs: ["read", "write"], levels: ["instance"] },
+    editor: { verbs: ["read", "write", "admin"], levels: ["instance"] },
     create: { verbs: ["create"], levels: ["type"] },
     publish: { verbs: ["publish"], levels: ["type"] },
     make_discoverable: { verbs: ["make_discoverable"], levels: ["type"] },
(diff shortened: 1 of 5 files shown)


front/lib/resources/group_permission_registry.ts
```diff
@@ -63,7 +63,7 @@ export const ROLE_REGISTRY: Record<
     publish: { verbs: ["publish"], levels: ["type"] },
   },
   skill: {
-    editor: { verbs: ["read", "write"], levels: ["instance"] },
+    editor: { verbs: ["read", "write", "admin"], levels: ["instance"] },
```
fabiencelier, line 66: Note: this was missing creating an inconsitency: members of the skill editors group can currently administrate the skill as per 

https://github.com/dust-tt/dust/blob/5f0e769e40c1626861cec4222cf0413e3d0d4601/front/lib/resources/group_resource.ts#L2578-L2584

front/lib/resources/skill/skill_resource.ts
```diff
@@ -2123,6 +2134,74 @@ export class SkillResource extends BaseResource<SkillConfigurationModel> {
...
+        ),
+      },
+    ];
+  }
+
+  // Writes this skill's `group_permissions` rows directly from its `group_skills` association
+  // (see `skillGroupGrants`). The skill mutation paths call this to keep the table in sync as it
+  // becomes the source of truth. Idempotent — it clears the skill's instance grants then
+  // re-inserts the desired set, so it is safe to re-run.
+  async writeGroupPermissions(
```
matteotrab, line 2161: Should we also give the global group read permission to be explicit on the fact that everyone can read a skill?

matteotrab (approved):
LGTM, left one question

front/lib/resources/group_permission_registry.ts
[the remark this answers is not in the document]
fabiencelier: @ZeroPath false-positive because this is on purpose
editor group must be able to administrate their skill

front/lib/resources/skill/skill_resource.ts
fabiencelier: Good question, but "everyone can read a skill" isn't quite true today: baseFetch drops any skill whose requestedSpaceIds the caller can't read (`hasPermissionForAcls("read", …)` on the space ACLs), so a user without access to a referenced space doesn't get the skill at all.

matteotrab: Yes, but that’s also true for the editor group. What I find confusing is that we explicitly say editors have “read” permission (let's leave aside that they can only load the skill if they can also read the spaces the skill requests access to). At the same time, non-editors are shown as not having read permission, yet they can still read the skill (when they can also read the requested spaces)

PR #30435, opened 2026-08-12
[front] feat: add nullable metadata column to skills

FlagBenett:
Adds a client-owned metadata JSONB column (label map) to skill_configurations and
skill_versions, the SkillMetadata type, and the two forced attribute defaults on the
global-skill and restore-from-version builders. Nullable, no default, no backfill.
Schema-only migrate PR; the API read/write code ships in a follow-up.

## Description

Closes https://github.com/dust-tt/tasks/issues/10025

Adds a first-class, client-owned metadata field (a Record<string, string> label map, like Kubernetes labels / cloud-resource tags) to skills. It's settable at import via the public API and returned on GET.

Dust doesn't prescribe the keys, callers pick their own (e.g. { "managedBy": "acme-ci" }).

Semantics: sending metadata on re-import overwrites it; omitting it preserves the existing value (matches updateSkill's ...(x ? { x } : {}) guards).

## Tests
locally

## Risk
no breaking change: additive only

## Deploy Plan
front/migrations/pre-deploy/20260812120000_add_metadata_to_skills.sql
then deploy front

(adds a nullable JSONB metadata column to skill_configurations and skill_versions. Nullable, no default, no backfill.)

### DIFF
--- a/front/lib/models/skill.ts
+++ b/front/lib/models/skill.ts
@@ -11,6 +11,7 @@ import { UserModel } from "@app/lib/resources/storage/models/user";
 import { WorkspaceAwareModel } from "@app/lib/resources/storage/wrappers/workspace_models";
 import type {
   SkillAvailability,
+  SkillMetadata,
   SkillReinforcementMode,
   SkillSourceMetadata,
   SkillSourceType,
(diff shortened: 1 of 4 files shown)


FlagBenett:

- #30435 ◀
    - #30436

PR #30440, opened 2026-08-12
[front] enh: polish the Pod app tile

davidebbo:
## Description

Visual pass on the Pod Apps tab tiles, following up on #30430:

- Drop the "Pinned as tab" chip — not needed on the tile.
- Drop the Frame filename line — the app name is enough.
- New default icon: the previous fallback (`ActionDashboardIcon`, shared with frame tabs) is a gauge that reads as a clock at tile size. Apps now fall back to the Frame glyph (`ActionFrameIcon`), which matches what an app is. Apps whose Frame is pinned as a tab keep showing their tab's icon.
- Nicer tile now that it shows so little: the icon gets a `ResourceAvatar` treatment on a contrasting background, and the title steps up from `copy-sm` to `heading-base`, vertically centered. The Draft chip stays for apps with no functions and no databases.

## Risk

Low — presentation-only change to the Apps tab tiles.

## Deploy Plan

Deploy front.

### DIFF
--- a/front/components/pod/apps/PodAppTile.tsx
+++ b/front/components/pod/apps/PodAppTile.tsx
@@ -1,12 +1,14 @@
 import type { CustomResourceIconType } from "@app/components/resources/resources_icon_names";
-import { getIcon } from "@app/components/resources/resources_icons";
+import {
+  getIcon,
+  ResourceAvatar,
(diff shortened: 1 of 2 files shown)


fontanierh (commented):
LGTM

PR #30443, opened 2026-08-12
Flav/live agent message credit breakdown

flvndvd:
## Description

This PR adds a dedicated events so client refreshes to get the message breakdown/cost details. This event is being published on conversation events SSE.


aubin-tchoi (approved):
LGTM

PR #30445, opened 2026-08-12
[Activation] collapse thinking when UIView is Compact

xebeohp:
## Description

Collapse inline thinking details by default in compact UI conversations, while preserving the existing expanded behavior for standard conversations.

- Thread the existing `uiView` value through `InlineActivitySteps` and `ActivityTimeline` into `ThinkingStep`.
- In compact UI, keep thinking collapsed during streaming and after completion, showing a concise “Thinking…” row instead of the full reasoning preview.
- Allow users to click the collapsed row to expand the thinking details when needed.

## Tests

- Updated `front/components/assistant/conversation/actions/inline/InlineActivitySteps.test.tsx` to provide the required `uiView` context.
- Retained the compact UI conversation coverage in `front/components/assistant/conversation/AgentMessage.compactUIView.test.tsx`.

## Risk

Low. The change only affects thinking-step presentation when `uiView` is `compact`. Standard conversations retain their existing behavior, and users can still expand collapsed thinking details on demand.

## Deploy Plan

Deploy `front`.

### DIFF
--- a/front/components/assistant/conversation/actions/inline/ActivityTimeline.tsx
+++ b/front/components/assistant/conversation/actions/inline/ActivityTimeline.tsx
@@ -4,6 +4,7 @@ import {
   getActionStepIcon,
   getCollapseAnimationStyle,
 } from "@app/components/assistant/conversation/actions/inline/utils";
+import type { UiView } from "@app/components/assistant/conversation/types";
 import type { InlineActivityStep } from "@app/types/assistant/conversation";
 import { assertNeverAndIgnore } from "@app/types/shared/utils/assert_never";
 import { ChevronRight, cn, Icon } from "@dust-tt/sparkle";
@@ -29,6 +30,7 @@ interface ActivityTimelineProps {
   };
   renderContentStep?: (content: string) => React.ReactNode;
   extraBelowCollapse?: React.ReactNode;
+  uiView?: UiView;
 }
 
 export function ActivityTimeline({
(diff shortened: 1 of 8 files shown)


xebeohp:

- #29996
    - #30109
        - #30242
            - #30258
                - #30445 ◀
                    - #30488

front/components/assistant/conversation/actions/inline/ActivityTimeline.tsx
```diff
@@ -42,8 +44,11 @@ export function ActivityTimeline({
   terminalRow,
   renderContentStep,
   extraBelowCollapse,
+  uiView = "standard",
 }: ActivityTimelineProps) {
-  const [isCollapsed, setIsCollapsed] = useState(isDone);
+  const [isCollapsed, setIsCollapsed] = useState(
+    isDone || uiView === "compact"
```
frankaloia, line 50: How does it work if we have an ask_question during the thinking phase? What does it look like?

xebeohp: Screen recording (the conversation was really slow to finish for some reason so ended the recording halfway but this is the gist):

https://github.com/user-attachments/assets/1670191a-d311-4cf7-ba24-aa994a0067fe

front/components/assistant/conversation/actions/inline/ThinkingStep.tsx
```diff
@@ -95,25 +111,54 @@ export const ThinkingStep = memo(function ThinkingStep({
       }
     : undefined;
 
+  const handleKeyDown = needsTruncation
```
frankaloia, line 114: Did we want to do this even if the message is done?

xebeohp: Yes, so people can still expand thinking after the message is done if they want. Should I get rid of it? --> thinking is always collapsed?

frankaloia: I was just a bit unclear as ithought we already have a toggle for the thinking button?

xebeohp: The "Thinking..." is already the toggle and no separate button exists. We just want to start with thinking collapsed but thinking can still be expanded while thinking and after.

PR #30447, opened 2026-08-12
Add Grok 4.6 support

fontanierh:
## Description

Adds `grok-4.6` across model config, pricing, routing, SDK types, the model picker, and the `model_constructors` and Dust stream layers. The native endpoint follows the documented xAI 500k context, reasoning, vision, tools, structured output, and tiered pricing; Dust caps it to 256k context and 64k output.

Grok 4.6 replaces 4.5 as xAI latest, preferred, and Auto candidate. Grok 4.5 remains registered and selectable.

## Tests

Ran the full 46-case live xAI endpoint contract suite, including all reasoning efforts, temperature variants, forced tools, structured output, and reasoning summaries. Added focused config, pricing-boundary, and stable prompt-cache-key tests.

## Risk

Low. The model remains gated by `xai_feature`; the main behavior change is that xAI-only workspaces and Auto routing now prefer Grok 4.6.

## Deploy Plan

Standard deploy.


front/lib/api/assistant/token_pricing/tiers.ts
```diff
@@ -395,6 +395,11 @@ export const STATIC_MODEL_TIERS: StaticModelTiersLookup = {
     medium: "balanced",
     high: "balanced",
   },
+  "grok-4.6": {
+    light: "balanced",
+    medium: "balanced",
+    high: "balanced",
```
Nils-Fedrigo: maybe have same tier as Opus Medium since it's similar pricing ? 

Image

Nils-Fedrigo (approved):
thanks !

front/lib/api/assistant/token_pricing/tiers.ts
fontanierh: Good call. I aligned Grok 4.6 with the existing frontier-model pattern: light remains balanced, while medium and high are premium. Added a focused tier regression test in ff36418284.

PR #30429, opened 2026-08-12
[analytics] Add download raw data to analytics consumption endpoint

avervaet:
## Description

Adds a **Download raw data** action to the consumption attribution table. The action uses the table's current period and scope filters and sends them to a new manager-only `POST /api/w/:wId/analytics/consumption/export-raw` endpoint.

**On the app:**
Capture d’écran 2026-08-13 à 12 38 39

The endpoint paginates through all matching consumption documents with Elasticsearch `search_after`, resolves labels for agents, users, groups, models, tools, skills, and sources, and returns a ZIP attachment containing `lines.csv`. Each row represents a raw consumption line and includes attribution metadata, optional model/tool details, user and group information, status/timing fields, and a credit breakdown. Credit values are rounded before CSV serialization.

The change also extends the CSV download hook to support JSON POST bodies, extracts shared attribution-row conversion/types for the existing top-dimension table, and centralizes the message/invocation unit mapping used by the top aggregations. No data migration or feature flag is introduced.

## Tests

The PR adds route coverage for:

- Successful ZIP/CSV generation with resolved labels for LLM and tool rows.
- Rejection of non-manager requests with `403`.
- Invalid request bodies with `400`.
- Elasticsearch failures with `500`.

## Risk

The export is read-only, reuses the existing analytics period and scope filtering, and is restricted to managers. The main operational risk is resource usage for large workspaces: matching documents are accumulated in memory and then converted into a ZIP in a single request.

## Deploy Plan

- Deploy front

PR #30428, opened 2026-08-12
[analytics] add Explore section header and match cycle selector button sizing

avervaet:
## Description

Small UI adjustments to the Analytics consumption page to better match Figma:

- Add spacing between the **Explore** heading and period selector.
- Move the attribution search below the dimension tabs and make it full width.
- Increase the period selector button size from `xs` to `sm`.

Before:
Capture d’écran 2026-08-12 à 16 26 06

After: 
Capture d’écran 2026-08-12 à 16 32 38

## Tests

Manual visual validation is included in the attached before/after screenshots.

## Risk

Low risk: presentational changes only; analytics behavior and data handling are unchanged.

## Deploy Plan

- Deploy front

PR #30448, opened 2026-08-12
[analytics] share label resolution + average computation across consumption rankings

avervaet:
## Description

Centralizes label resolution and average-credit computation for all consumption ranking endpoints in a shared `resolveConsumptionGroupLabels` helper. The helper resolves labels for the requested dimension, preserves the raw group key as a fallback name, and computes the average with the existing zero-count guard.

The agents, groups, models, skills, sources, tools, and users endpoints now consume the shared resolved rows while continuing to expose their existing endpoint-specific response fields and null handling. This removes duplicated mapping logic without changing the ranking data source or the public response shape.

## Tests

Localy tested

## Risk

The change affects the response-mapping layer of all seven consumption ranking endpoints. The main risk is a regression in label metadata mapping or endpoint-specific field conversion. The shared helper retains the existing fallback behavior and average calculation, while each endpoint continues to map to its existing response shape. No database or data migration is introduced, and the change is straightforward to roll back.

## Deploy Plan
- Deploy front

### DIFF
--- a/front/lib/api/analytics/consumption/top.ts
+++ b/front/lib/api/analytics/consumption/top.ts
@@ -1,3 +1,4 @@
+import { resolveDimensionLabels } from "@app/lib/api/analytics/consumption/labels";
 import type { ConsumptionPeriod } from "@app/lib/api/analytics/consumption/period";
 import type {
   ConsumptionScopeDimension,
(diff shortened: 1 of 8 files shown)


avervaet:

- #30405
    - #30414
        - #30428
            - #30429
                - #30448 ◀

aubin-tchoi (approved):
LGTM

PR #30451, opened 2026-08-12
[front] chore: add skills ES reindexation script

aubin-tchoi:
## Description

This PR adds a script to backfill our ES index on skills.
It goes through all the workspaces and triggers a re-indexation.
Quite heavy on the db since we list all skills + tools (in the workflow), so we process workspaces 1 by 1.

## Tests

- Tested locally.

## Risk

- Low.

## Deploy Plan

- Deploy front.


adrsimon (approved):

front/scripts/backfill_skill_search.ts
```diff
@@ -0,0 +1,67 @@
...
+  {
+    concurrency,
+    execute,
+    logger,
+  }: { concurrency: number; execute: boolean; logger: Logger }
+) {
+  const auth = await Authenticator.internalAdminForWorkspace(workspace.sId);
+  const skills = await SkillResource.listByWorkspace(auth, {
+    permissionFiltering: "redact_unreadable",
+    status: ["active", "archived"],
```
sfriquet, line 20: do we care about archived skills ?

sfriquet (approved):

front/scripts/backfill_skill_search.ts
aubin-tchoi: chose to still index them since they will be available from the Manage Skills page, searching in them seems like a low-cost feature, plus not having the same search features would require thinking about re-structuring the page a little (the "archived" tab is under the search bar)

PR #30452, opened 2026-08-12
[front] feat: add Elasticsearch index for skills

aubin-tchoi:
## Description

Design doc [here](https://app.notion.com/p/dust-tt/ES-backed-Agent-and-Skill-Search-3c828599d9418063be22d351beef97f7).

This PR adds the `front.skills_1` index, aliased behind `front.skills`.
Skill names use ICU tokenization, case-boundary splitting (some skills are named in PascalCase or camelCase) and ICU folding on the `search_as_you_type` field (will be the one used to search on the name).

This PR only adds the index definition and registers the index in the creation script.

## Tests

- Tested locally.

## Risk

- Low. No runtime change.

## Deploy Plan

- Run `npx tsx scripts/create_elasticsearch_index.ts --index-name skills --index-version 1 --execute`
- Deploy front.


front/lib/skill_search/indices/skills_1.mappings.json
```diff
@@ -0,0 +1,68 @@
+{
+  "dynamic": "strict",
+  "properties": {
+    "metadata": {
+      "type": "object",
```
flvndvd: OOC, what's the point to index those if not properly parsed as JSON?

front/lib/skill_search/indices/skills_1.settings.europe-west1.json
```diff
@@ -0,0 +1,25 @@
+{
+  "number_of_shards": 3,
+  "number_of_replicas": 1,
+  "refresh_interval": "30s",
+  "analysis": {
+    "filter": {
+      "skill_name_words": {
```
flvndvd: `nit` can we have preserve in the name?

front/lib/skill_search/indices/skills_1.settings.europe-west1.json
```diff
@@ -0,0 +1,25 @@
...
+        "split_on_case_change": true,
+        "split_on_numerics": false,
+        "stem_english_possessive": false
+      }
+    },
+    "analyzer": {
+      "skill_name_analyzer": {
+        "type": "custom",
+        "tokenizer": "icu_tokenizer",
+        "filter": [
```
flvndvd: don't we want `asciifolding` here? so `café` becomes `cafe`?

front/lib/skill_search/indices/skills_1.settings.europe-west1.json
flvndvd: Could we keep an unsplit, folded version of the name searchable alongside the split tokens? If my read is correct `GitHub` matches `git` but would disappears at `gith` and `github`.

front/lib/skill_search/indices/skills_1.settings.europe-west1.json
```diff
@@ -0,0 +1,25 @@
...
+    "filter": {
+      "skill_name_words": {
+        "type": "word_delimiter_graph",
+        "split_on_case_change": true,
+        "split_on_numerics": false,
+        "stem_english_possessive": false
+      }
+    },
+    "analyzer": {
+      "skill_name_analyzer": {
```
flvndvd, line 26: Could we normalize to NFC before splitting? Two visually identical spellings of `RésuméRéunion` would produce different tokens depending on how the accents are encoded: `[resume, reunion]` vs `[resumereunion]`.

front/lib/skill_search/indices/skills_1.mappings.json
```diff
@@ -0,0 +1,68 @@
...
+      "type": "keyword"
+    },
+    "availability": {
+      "type": "keyword"
+    },
+    "name": {
+      "type": "text",
+      "analyzer": "skill_name_analyzer",
+      "fields": {
+        "keyword": {
```
flvndvd, line 20: Should exact-name matching ignore case and accents? Currently, `name.keyword` requires the literal spelling, so "resume de reunion" won’t match "Résumé de réunion".

front/lib/skill_search/indices/skills_1.mappings.json
aubin-tchoi: Was planning to use this for metadata we need in the UI but never for filters, like the source

aubin-tchoi: On second thought we should probably default to indexing

front/lib/skill_search/indices/skills_1.mappings.json
aubin-tchoi: Was hoping to rely exclusively on the search_as_you_type, but probably a good idea to add a boost, will add a text type here

front/lib/skill_search/indices/skills_1.settings.europe-west1.json
aubin-tchoi: That's baked in icu_folding, right? IIUC I only have to swap the order?

front/lib/skill_search/indices/skills_1.mappings.json
```diff
@@ -0,0 +1,68 @@
...
+        "autocomplete": {
+          "type": "search_as_you_type",
+          "analyzer": "skill_name_analyzer"
+        }
+      }
+    },
+    "icon": {
+      "type": "keyword"
+    },
+    "edited_by": {
```
sfriquet: what does this field contain ? I would have expected this to represent a user but we already have editors below 🤔

front/lib/skill_search/indices/skills_1.mappings.json
```diff
@@ -0,0 +1,68 @@
...
+    },
+    "favorite_count": {
+      "type": "integer"
+    },
+    "is_default": {
+      "type": "boolean"
+    },
+    "editors": {
+      "type": "keyword"
+    }
```
sfriquet: I assume this will hold user ids ? if yes let's perhaps align naming we have `requested_space_ids` and `editor_group_ids` so let's have `editor_ids` ? 

likewise we have `favorite_count` vs `active_users` but both hold count ?

front/lib/skill_search/indices/skills_1.mappings.json
aubin-tchoi: ID of the user who edited last

front/lib/skill_search/indices/skills_1.mappings.json
aubin-tchoi: Agreed , will align on _ids and _count to be explicit

front/lib/skill_search/indices/skills_1.mappings.json
sfriquet: `last_editor_id` ?

aubin-tchoi: I put last_edited_by_user_id, longer but avoids confusion with the editors

front/lib/skill_search/indices/skills_1.settings.europe-west1.json
```diff
@@ -0,0 +1,43 @@
+{
+  "number_of_shards": 1,
```
aubin-tchoi, line 2: starting with one shard given the low volume and the fact that we have 2 for data source nodes. does that sound reasonable?

sfriquet (approved):
lgtm (deferring to @flvndvd for the settings)

front/lib/skill_search/indices/skills_1.mappings.json
```diff
@@ -0,0 +1,66 @@
...
+    },
+    "status": {
+      "type": "keyword"
+    },
+    "availability": {
+      "type": "keyword"
+    },
+    "name": {
+      "type": "text",
+      "analyzer": "skill_name_exact_analyzer",
```
flvndvd: Small nit but future-proof naming is usually about what the analyzer does more than the field that uses it.

front/lib/skill_search/indices/skills_1.mappings.json
```diff
@@ -0,0 +1,66 @@
...
+      "fields": {
+        "keyword": {
```
flvndvd, line 20: Can we get the query that you have in mind on how we leverage those different fields? FWIW if we want to avoid doing arbitrary boost, we could also preserve the original.

adrsimon (commented):
general question about ES mappings @flvndvd, i see we have one for EU and one for US, how does it behaves for single tenant ? EU is used for every cell hosted in EU region i believe ? or should we have one mapping per cell ?

flvndvd:
> general question about ES mappings @flvndvd, i see we have one for EU and one for US, how does it behaves for single tenant ? EU is used for every cell hosted in EU region i believe ? or should we have one mapping per cell ?

See thread [here](https://dust4ai.slack.com/archives/C0BRB4T824R/p1789552106640329).

front/lib/skill_search/indices/skills_1.settings.europe-west1.json
flvndvd: Do we have a rough estimation of the worst case size of a skill item? I presume name and description are both length-constrained in our DB anyway?

front/lib/skill_search/indices/skills_1.mappings.json
aubin-tchoi: adding preserve_original to the filter doesn't seem to work with search_as_you_type (crashes at indexation)

aubin-tchoi: will keep the preserved search_as_you_type to allow matching words not split by case change like github (not possible with the keyword as it would be a full prefix match)
sample query
```json
multi_match: {
  query: "github",
  type: "bool_prefix",
  operator: "and",
  fields: [
    "name.autocomplete",
    "name.autocomplete._2gram",
    "name.autocomplete_preserved",
    "name.autocomplete_preserved._2gram",
  ],
},
```

flvndvd: oh interesting 🤔

front/lib/skill_search/indices/skills_1.settings.europe-west1.json
aubin-tchoi: yes, added a small section [here](https://app.notion.com/p/dust-tt/ES-backed-Agent-and-Skill-Search-3c828599d9418063be22d351beef97f7?source=copy_link#3d828599d941806ab81cdba77f8faf9a). will make description and name bounded (currently a TEXT in db. worst case is 1kb if you have a lot of tools

front/lib/skill_search/indices/skills_1.mappings.json
flvndvd: Seen IRL 2grams -> shingle ngrams of words actually.

flvndvd (approved):
LGTM

PR #30453, opened 2026-08-12
[front] enh: index committed skill mutations

aubin-tchoi:
## Description

Direct follow up to https://github.com/dust-tt/dust/pull/32414, which added a Temporal workflow to enqueue skill indexation.

This PR hooks it up to the paths where we modify a skill, all contained within `SkillResource`: creation, update, editor list change, favorite change, archive, restore, deletions.

There's an added complexity from the update via a reference (nested skills), where we sometimes have to update all the referencing skills when updating a single skill.

There are some updates on fields that are not indexed (reinforcement). These do not trigger a reindexation.

## Tests

- Added tests.
- Tested locally.

## Risk

- Low. We'll start writing in our index.

## Deploy Plan

- Deploy front.


front/lib/resources/skill/skill_resource.ts
```diff
@@ -571,6 +574,7 @@ export class SkillResource extends BaseResource<SkillConfigurationModel> {
     // construction are now stale and the caller would not be an editor of the skill they just
     // created. Refresh the snapshot now that the write has committed, as space creation does.
     await auth.refresh();
+    await this.launchSearchIndexation(auth, [skillResource.sId]);
```
aubin-tchoi, line 580: that's the creation (makeNew)

front/lib/resources/skill/skill_resource.ts
```diff
@@ -1428,6 +1432,7 @@ export class SkillResource extends BaseResource<SkillConfigurationModel> {
           workspaceId: workspace.id,
         },
       });
+      await SkillResource.launchSearchIndexation(auth, [this.sId]);
```
aubin-tchoi, line 1442: that's the `setFavorite`, a bit edge but it does change the number of favorite

front/lib/resources/skill/skill_resource.ts
```diff
@@ -3112,65 +3125,86 @@ export class SkillResource extends BaseResource<SkillConfigurationModel> {
...
-      // Rename any existing archived skill with the same name to avoid unique constraint violation.
-      const existingArchivedSkill = await this.model.findOne({
-        where: {
-          workspaceId: workspace.id,
-          name: this.name,
-          status: "archived",
-        },
-        transaction,
-      });
+    const { affectedCount, referencingSkillIds, renamedSkillId } =
```
aubin-tchoi, line 3131: this is the archive. we update the skills that reference the current skill, so these ones need to be re-upserted as well (spaces may have changed)

aubin-tchoi: ok I'm not adding a comment on every method, it's going to be a bit annoying to review 😬

adrsimon (approved):

sfriquet (approved):

PR #30461, opened 2026-08-12
[front] fix: Guard tool display label lookup against unknown internal MCP server names

id13:
## Description

When we release new mcp servers and don't release the extension, current code deployed tries to dereference action's mcp tool from a map that does not contain it (yet), which makes the whole conversation render crash.

This patch makes it fail without crashing, not amazing but at least limits the blast radius, the real fix is simply to release the extension more often.

## Tests

Not tested

## Risk

Low

## Deploy Plan

Merge
Release the extension


id13:

- #30461 ◀

matteotrab (approved):
Thank you

PR #30462, opened 2026-08-12
Admin governance: shadow compare skill permissions

fabiencelier:
## Description

related to https://github.com/dust-tt/tasks/issues/9472

Shadow compare permissions from group and from permissions table for skills.
Only when feature flag is activated
Warn on difference

## Tests

added unit test

## Risk

low, just a warning + feature flag

## Deploy Plan

wait for migration to be run
deploy front
deploy sse

### DIFF
--- a/front/lib/resources/skill/skill_resource.ts
+++ b/front/lib/resources/skill/skill_resource.ts
@@ -92,8 +92,13 @@ import type {
 } from "@app/types/assistant/skill_configuration";
 import { isDefaultFromAvailability } from "@app/types/assistant/skill_configuration";
 import type { AgentsUsageType } from "@app/types/data_source";
+import type { GrantVerb } from "@app/types/group_permissions";
 import { SKILL_GROUP_PREFIX } from "@app/types/groups";
-import type { GroupGrant } from "@app/types/resource_permissions";
+import type {
+  AccessControlList,
+  GroupGrant,
+  RoleGrant,
+} from "@app/types/resource_permissions";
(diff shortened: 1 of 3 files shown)


fabiencelier:
blocked until migration has been run

front/lib/resources/skill/skill_resource.ts
```diff
@@ -258,6 +263,13 @@ export interface SkillResource
 // The role a skill's editor group holds on the skill. Its verbs live in ROLE_REGISTRY.skill.
 const SKILL_EDITOR_GRANT_TYPE = "editor" as const;
 
+const SKILL_ROLE_GRANTS: RoleGrant[] = [
+  { role: "admin", permissions: ["read", "admin"] },
+  { role: "manager", permissions: ["read"] },
+  { role: "user", permissions: ["read"] },
+  { role: "builder", permissions: ["read"] },
```
matteotrab, line 270: Sorry to bring it up again, but would it be clearer here to grant the "read" permission to the global group as we discussed in https://github.com/dust-tt/dust/pull/30434#discussion_r3767604835 instead?
The way I see it is that a user has "read" permission because he is a member of the workspace, not because of the role he has in the workspace. What do you think?

matteotrab (approved):
Nice

front/lib/resources/skill/skill_resource.ts
fabiencelier: discussed and tracked here: https://github.com/dust-tt/tasks/issues/10052

PR #30465, opened 2026-08-12
[Activation] Slack auto-responder skip thread replies option

frankaloia:
## Summary

- Adds a \"Top-level posts only\" sub-toggle to the Slack auto-responder settings. When enabled alongside \"Respond to all messages\", the bot will skip thread replies and only respond to new top-level channel posts.

## Testing
Screenshot 2026-08-12 at 4 06 45 PM

## Risk

Low

## Deploy Plan

1. Deploy connector
2. Deploy front

### DIFF
--- a/connectors/src/api/slack_channels_linked_with_agent.ts
+++ b/connectors/src/api/slack_channels_linked_with_agent.ts
@@ -20,7 +20,10 @@ const PatchSlackChannelsLinkedWithAgentReqBodySchema = t.type({
   slack_channel_internal_ids: t.array(t.string),
   connector_id: t.string,
   auto_respond_without_mention: t.union([t.boolean, t.undefined]),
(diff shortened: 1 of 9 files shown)


smb2268:
@frankaloia Overall LGTM, but I think the UX is a little confusing. I’d expect the nested setting to be a checkbox rather than a toggle, and for it to remain visible but disabled when the parent setting is off for discoverability. Maybe we also remove the top level heading so it's more clearly nested?

frankaloia:
> @frankaloia Overall LGTM, but I think the UX is a little confusing. I’d expect the nested setting to be a checkbox rather than a toggle, and for it to remain visible but disabled when the parent setting is off for discoverability. Maybe we also remove the top level heading so it's more clearly nested?

@smb2268 Better?

Screenshot 2026-08-12 at 6 20 38 PM

smb2268 (approved):
lgtm!

PR #30467, opened 2026-08-12
[front] fix: polish consumption analytics loading and tabs

aubin-tchoi:
## Description

This PR does a round of UI polishes on the new analytics page:
- Skeleton instead of spinner in the filter panel (we know the target layout so we can put a skeleton with the same layout).
- Skeletons in the attribution tables as well.
- The loading state for the big chart takes the same height, prevents a shift in layout.
- Dark mode around the Explore is fixed (used to look weird with 0 padding on the left)

## Tests

- Added tests.

## Risk

- Low. UI-only.

## Deploy Plan

- Deploy front.

### DIFF
--- a/front/components/charts/ChartContainer.tsx
+++ b/front/components/charts/ChartContainer.tsx
@@ -75,7 +75,7 @@ export function ChartContainer({
         {isLoading || message ? (
           <div
             className="flex items-center justify-center"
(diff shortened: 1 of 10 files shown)


ClementAupiais (approved):
LGTM

PR #30469, opened 2026-08-12
feat(sandbox): remove legacy egress policy scrubbing and the workspace-layout dual-write

smb2268:
## Description

Front companion to the egress-proxy legacy-layer removal (`30464`): the per-providerId policy files (`sandboxes/{providerId}.json`) are dead — nothing writes them and the proxy no longer reads them — so the destroy-path cleanup goes: `deleteLegacySandboxPolicy` and its call sites in the kill-requested branch, `deleteByOwner`, and the scope-transition destroy, plus the test plumbing that mocked it.

**Second commit — the legacy workspace layout compatibility plumbing** also goes: reads no longer fall back to `workspaces/{wId}.json`, writes no longer dual-write it, deletes no longer scrub it first. This formally closes the front rollback window past the relayout — a rolled-back front would read and write only the legacy path, which nothing maintains anymore.

Merge order relative to `30464` does not matter. Leftover objects are handled by the tidy-up in `30464`'s deploy plan.

> [!NOTE]
> Gate satisfied: the `30468` backfill dry run (2026-08-12) verified **zero uncopied legacy objects in both regions** — the backfill was unnecessary (dual-write covered everything over time) and its PR is closed.

## Risk

None — deletion of dead cleanup code. Safe to rollback.

## Deploy Plan

- [x] Verified zero legacy-only workspace objects in both regions (dry run, 2026-08-12)
- [ ] Deploy front (no ordering constraints relative to `30464`)

## Tests

Affected suites green (sandbox resource 48, egress policy 15); the legacy-path read/write/delete tests are removed with the plumbing they pinned.

### DIFF
--- a/front/lib/api/sandbox/egress_policy.test.ts
+++ b/front/lib/api/sandbox/egress_policy.test.ts
@@ -39,7 +39,6 @@ vi.mock("@app/lib/file_storage", () => ({
 
 import {
   addOwnerPolicyDomain,
-  deleteLegacySandboxPolicy,
   deleteOwnerPolicy,
   deleteWorkspacePolicy,
   parseExactEgressDomain,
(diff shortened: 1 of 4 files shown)


fontanierh (approved):
LGTM

PR #30471, opened 2026-08-12
Private Slack channel selection for default agents

frankaloia:
## Summary

Lets admins surface **private** Slack channels they belong to in the default-agent channel picker. The admin connects their personal Slack account; private-channel listing is scoped per-user (other admins don't see each other's private channels). Dust must still be present in a private channel to reply there. We cannot do this via non-user credential mechanisms today because Slack ToS says we can't present a private channel of admin A to admin B.

## Testing

Screenshot 2026-08-12 at 5 49 36 PM

## Risk

Low-ish. Impact is scoped to just Slack panel, but we do introduce a new API that uses personal credentials for Slack

## Deploy
1. Depoy front

### DIFF
--- a/front/components/agent_builder/settings/SlackSettingsSheet.tsx
+++ b/front/components/agent_builder/settings/SlackSettingsSheet.tsx
@@ -1,6 +1,7 @@
 import { useAgentBuilderContext } from "@app/components/agent_builder/AgentBuilderContext";
 import type { AgentBuilderFormData } from "@app/components/agent_builder/AgentBuilderFormContext";
 import { useFeatureFlags } from "@app/lib/auth/AuthContext";
+import { useSlackUserPrivateChannels } from "@app/lib/swr/assistants";
 import { useConnectorPermissions } from "@app/lib/swr/connectors";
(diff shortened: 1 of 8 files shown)


front/components/agent_builder/settings/SlackSettingsSheet.tsx
```diff
@@ -347,7 +392,8 @@ export function SlackSettingsSheet({
               Set this agent as the default agent on one or several of your
               Slack channels. It will answer by default when the{" "}
               <span className="font-bold">@Dust</span> Slack bot is mentioned in
-              these channels.
+              these channels. To see private channels, activate the Slack tool
```
davidebbo: Might not be enough guidance for some users.

davidebbo (approved):
LGTM

vincesarkisian:
@frankaloia Tested this in the demo workspace and I am seeing all Private channels. In addition not seeing the one I expected to see. auto-responder is the one that only I should see. demo_argos, elia_slack, my_personal_scratchpad I would not expect to see because I am not in those channels. 
Screenshot 2026-08-17 at 9 34 19 AM

PR #30473, opened 2026-08-12
Fix restricted conversation consumption attribution

flvndvd:
## Description

Consumption attribution and consumption analytics were fetching the same conversation with different semantics.
Attribution used the standard permission-filtered `ConversationResource.fetchById`. The backfill authenticator only
belongs to the workspace global group, so conversations in restricted or project spaces were filtered out. Attribution
then returned successfully without writing V4 items, allowing the workflow to proceed.

The analytics loader bypasses permission filtering and includes deleted conversations because historical billable consumption must remain indexable. It could therefore load the conversation but found no attribution items and repeatedly failed with Consumption attribution is incomplete for analytics.

This change gives attribution the same workspace-scoped fetch semantics as analytics. It also temporarily recomputes attribution from the indexing activity so workflows already retrying activity 2 can heal without rerunning their completed activity 1. That temporary call can be removed after the retry backlog drains. Focused tests cover restricted-space attribution and recomputation before indexing.


flvndvd:
PMRR.

aubin-tchoi (approved):
LGTM

PR #30476, opened 2026-08-13
fix: biome formatting in consumption attribution store test

avervaet:
## Summary
- `store.test.ts` was merged unformatted in #30473, breaking the format-check CI job on main and on any branch rebased on top of it.
- Applies `biome check --write` to bring the file back in line with formatting rules.

## Test plan
- [x] `npx biome check front/lib/api/assistant/agent_message_consumption_attribution/store.test.ts` passes


davidebbo (approved):
LGTM

PR #30477, opened 2026-08-13
[analytics] parallelize per-dimension label resolution in CSV export

avervaet:
Label resolution does real DB lookups per dimension (users, groups, skills,
tools, agents). The ES fetch across dimensions already runs concurrently via
Promise.all, but label resolution was awaited one dimension at a time
afterwards. Run it concurrently too.

PR #30478, opened 2026-08-13
[analytics] only compute the export total on the first composite page

avervaet:
total_credit_micro doesn't depend on the composite cursor, but it was being
recomputed (a full-dataset sum) on every single page. Request it only on the
first page and reuse the value for the rest of the pagination.


avervaet:

- #30405
    - #30414
        - #30428
            - #30429
                - #30448
                    - #30477
                        - #30478 ◀
                            - #30479
                                - #30480

avervaet:
Superseded: this branch only optimized the per-dimension consumption export / fetchConsumptionAllGroups, which has been removed from the stack (#30429).

PR #30479, opened 2026-08-13
[analytics] skip composite pagination for bounded dimensions

avervaet:
model, tool and source have a distinct-value count bounded well under a
single composite page, so paginating through the composite cursor for them
is pure overhead. Route them through fetchConsumptionTopGroups (a single
terms request) instead, capped at EXPORT_PAGE_SIZE.

PR #30480, opened 2026-08-13
[analytics] trim ES response payload for the composite export pages

avervaet:
The composite pagination loop already returns size:0, but the response
still carries shard stats and other fields the caller never reads. Add a
filterPath passthrough on searchConsumptionAnalytics and scope the
composite-page request to aggregations (plus error, so a failed request
still carries a diagnosable reason).

### DIFF
--- a/front/lib/api/analytics/consumption/top.ts
+++ b/front/lib/api/analytics/consumption/top.ts
@@ -244,6 +244,9 @@ export async function fetchConsumptionAllGroups(
             : {}),
         },
         size: 0,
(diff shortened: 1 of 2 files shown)


avervaet:

- #30405
    - #30414
        - #30428
            - #30429
                - #30448
                    - #30477
                        - #30478
                            - #30479
                                - #30480 ◀

avervaet:
Superseded: this branch only optimized the per-dimension consumption export / fetchConsumptionAllGroups, which has been removed from the stack (#30429).

PR #30482, opened 2026-08-13
chore(front): remove unused jszip dependency

avervaet:
## Summary
- Removes the unused `jszip` dependency from `front/package.json` — grep found zero usages anywhere in the repo.
- `jszip` was used to zip up CSV usage exports in the old `workspace-usage` endpoints (private and v1), which were removed in #28322 and #28325. It's been dead weight since then; `adm-zip` is the live zipping pattern used elsewhere in the codebase today.

## Test plan
- [x] Confirmed no remaining `jszip` references (code or other `package.json` files) via grep
- [x] Regenerated `package-lock.json` with `npm install --package-lock-only`

### DIFF
--- a/front/package.json
+++ b/front/package.json
@@ -171,7 +171,6 @@
     "jsforce": "^3.8.2",
     "json-schema-to-typescript": "^15.0.4",
     "jsonwebtoken": "^9.0.0",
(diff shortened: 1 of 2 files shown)


davidebbo (approved):
LGTM

PR #30483, opened 2026-08-13
[front] fix: Harden sandbox deployment process

id13:
## Description

- Removed front auto build of sbx images, it was brittle (incident-prone) and not the current recommend way ([runbook](https://app.notion.com/p/dust-tt/Runbook-deploying-a-new-sandbox-image-3b828599d941808e8ac5c0d8afb26063?source=copy_link))
- Instead, add a new check step on front that fails if the registered tag in the code does not match with a template in our e2b clusters.
- Add a cleanup step in build sandbox to delete the template if the build failed to avoid false skips

## Tests

N/A

## Risk

Low

## Deploy Plan

Merge

### DIFF
--- a/front/lib/api/sandbox/providers/e2b_template.ts
+++ b/front/lib/api/sandbox/providers/e2b_template.ts
@@ -187,24 +187,30 @@ class E2BTemplateBuilder {
   }
 }
 
-interface E2BTemplateInfo {
+export interface E2BTemplateInfo {
   templateId: string;
   aliases: readonly string[];
+  // Null until a build completes: E2B registers the template (and its alias) as
+  // soon as a build starts, and only attaches an envd version once one of them
+  // succeeds. A template without it can never be started.
+  envdVersion: string | null;
+}
+
+function createE2BApiClient(apiKey?: string): ApiClient {
+  const e2bConfig = config.getE2BSandboxConfig();
+  const connectionConfig = new ConnectionConfig({
+    apiKey: apiKey ?? e2bConfig.apiKey,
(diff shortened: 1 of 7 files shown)


front/lib/api/sandbox/providers/e2b_template.ts
```diff
@@ -225,21 +232,85 @@ export async function listE2BTemplates(
...
+  imageId: SandboxImageId,
+  apiKey?: string
+): Promise<Result<boolean, Error>> {
+  const templateResult = await findE2BTemplate(imageId, apiKey);
+  if (templateResult.isErr()) {
+    return templateResult;
+  }
+
+  const template = templateResult.value;
+  if (!template || template.envdVersion) {
```
id13, line 280: tribal knowledge for sure, but it's the symptom I have seen for half-baked template builds.

front/scripts/sandbox_image_verify.ts
```diff
@@ -0,0 +1,70 @@
...
+    for (const alias of template.aliases) {
+      envdVersionByAlias.set(alias, template.envdVersion);
+    }
+  }
+
+  const missing: string[] = [];
+  const unbuilt: string[] = [];
+  for (const id of requiredIds) {
+    const envdVersion = envdVersionByAlias.get(id);
+    if (envdVersion === undefined) {
```
id13, line 41: Same envD trick

fontanierh (approved):
Thank you 🙏

davidebbo (approved):
LGTM

PR #30486, opened 2026-08-13
[front] fix: align tool output rendering in consumption items with `renderConversationForModel`

aubin-tchoi:
## Description

When counting the tokens of the output of a tool for recording as consumption items, we sometimes undercount them compared to what we actually show the model, notably when we rewrite the content for the model (which we do for the semantic search for instance).

We can't change what we store in the output items, as these serve both the UI and the model. What we do in this PR is that we make sure that we follow the same code path for rendering the tool outputs as text.

## Tests

- Added tests that check that we count tokens the same way.

## Risk

- Low.

## Deploy Plan

- Deploy front.

### DIFF
--- a/front/lib/api/assistant/agent_message_consumption_attribution/store.ts
+++ b/front/lib/api/assistant/agent_message_consumption_attribution/store.ts
@@ -158,12 +158,14 @@ function runUsageModelIdsWithUnconsumedToolResults({
 async function buildRunUsageConsumptionEvidence(
   auth: Authenticator,
   {
+    conversationId,
     enrichedActionByModelId,
     includeToolResultFootprints,
     runActions,
     triggeringUserMessageOrigin,
     usage,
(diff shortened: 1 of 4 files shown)


front/lib/api/assistant/agent_message_consumption_attribution/store.ts
```diff
@@ -158,12 +158,14 @@ function runUsageModelIdsWithUnconsumedToolResults({
 async function buildRunUsageConsumptionEvidence(
   auth: Authenticator,
   {
+    conversationId,
```
aubin-tchoi, line 161: lots of drilling down, but I need it to be passed to `renderActionForMultiActionsModel`, which uses it for the image URLS (legacy path, not used in the end but I'd rather drill down that change the signature of `renderActionForMultiActionsModel`)

PR #30487, opened 2026-08-13
[front] enh: add pagination on attribution table in new analytics

aubin-tchoi:
## Description

The consumption attribution table currently only loads the first 25 rows, so lower-ranked entries cannot be paged to or found through search.

This PR lets the table request all the content, with back-end side pagination (implemented a relatively crude and pragmatic way of paginating, it assumes that most users will only look at the first few pages but seldom go further).

## Tests

- Added tests
- Tested locally

## Risk

- Low.

## Deploy Plan

- Deploy front.


avervaet:
As pr irl: In ConsumptionAttributionTable.tsx, allRows now contains only the page returned by the API, but the existing client-side search filters that array. Since changing the search also resets the component to page 1, searching for a lower-ranked row will always return “No match” unless that row happens to be on the first page.

front/lib/api/analytics/consumption/schema.ts
```diff
@@ -41,8 +41,12 @@ export const ConsumptionTopBodySchema = ConsumptionBodySchema.extend({
     .int()
     .positive()
     .max(100)
+    .nullable()
     .optional()
-    .default(DEFAULT_CONSUMPTION_TOP_LIMIT),
+    .default(DEFAULT_CONSUMPTION_TOP_LIMIT)
+    .transform((limit) => limit ?? DEFAULT_CONSUMPTION_TOP_LIMIT),
+  offset: z.number().int().nonnegative().optional().default(0),
```
avervaet, line 48: We might want to cap the offset somewhere to avoid stupidly large bucket queries to happen if an user digs deep enough

avervaet (approved):

aubin-tchoi:
will tackle the search part in a separate PR actually to avoid complexifying this PR too much

PR #30488, opened 2026-08-13
[Activation] Remove isCompactUIView column & model

xebeohp:
## Description

Remove the obsolete `isCompactUIView` field from `activation-pods` now that compact UI is derived client-side from the conversation's `uiView` rather than persisted on the Activation Pod record.

- Remove the `isCompactUIView` attribute from `ActivationPodModel`.
- Add a post-deploy migration to drop the `isCompactUIView` column from `public.activation_pods`.
- No backfill is required because the column is no longer read or written by the application.

## Risk

Low application risk because the column is no longer used by the compact UI flow.

## Deploy Plan
Follow post-deploy migration steps.


xebeohp:

- #29996
    - #30109
        - #30242
            - #30258
                - #30445
                    - #30488 ◀

id13 (approved):

PR #30489, opened 2026-08-13
feat(front): open consumption analytics to managers

adrsimon:
## Description

Widening `/analytics/consumption` to managers, the nav entry and every endpoint behind it were already manager-gated, only the page itself and the `analytics` agents view were admin-only.

Small cleanup that falls out of it: dropped the "Unknown agent" placeholder. Archived and retired agents still resolve fine.

## Tests

Updated tests files. Tested locally with a manager role account.

## Risk

Low. Managers see private agent names and descriptions in the analytics tables now, which is the point of the change but worth calling out. Everything below the manager role is unchanged. Straight rollback.

## Deploy Plan

Deploy `front`.

### DIFF
--- a/front/lib/api/assistant/observability/top_agents.ts
+++ b/front/lib/api/assistant/observability/top_agents.ts
@@ -1,7 +1,4 @@
-import {
-  resolveAnalyticsAgentLabels,
-  UNKNOWN_AGENT_LABEL,
-} from "@app/lib/api/assistant/observability/agent_labels";
+import { resolveAnalyticsAgentLabels } from "@app/lib/api/assistant/observability/agent_labels";
 import {
   fetchAgentCostStats,
   getAgentCostStats,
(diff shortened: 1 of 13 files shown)


front/lib/api/assistant/observability/top_agents.ts
```diff
@@ -114,18 +111,23 @@ export async function fetchTopAgents(
   }
   const costStatsMap = costStatsResult.value;
 
-  const rows = buckets.map((bucket) => {
+  const rows = buckets.flatMap((bucket) => {
     const agentId = String(bucket.key);
-    const label = agents.get(agentId) ?? UNKNOWN_AGENT_LABEL;
```
aubin-tchoi, line 119: isn't it better to have a catch all category?

aubin-tchoi (approved):
LGTM

front/lib/api/assistant/observability/top_agents.ts
adrsimon: we're not supposed to not have an agent in this list, it's only supposed to happen when you don't have access to it

if an agent don't have an sId, then it would be the equivalent as it not existing in DB. also, when we'll do the users view, this will filter out the agents the user is not supposed to see at all rather than bucketing them in a category they're not supposed to see

this code will change in subsequent PRs about making this page accessible to users whatsoever

PR #30490, opened 2026-08-13
[front] fix: restart litestream when a Pod app's databases are deleted

davidebbo:
## Description

Deleting a Pod app removes each database's live files and then wipes its replica prefix, on the assumption that a litestream which can no longer see the files stops replicating them. It does not. The directory watcher only enumerates `/pod-state/databases` at daemon start, so until a restart the daemon still holds the deleted database open and is free to recreate the prefix the wipe just removed.

So the delete now restarts the daemon between the two steps: remove every live file, restart, then wipe every replica.

## Risk

Low. One extra root command on a path that already wakes the pod and can take as long as a cold start. The delete gets slower by one daemon restart, and a restart that fails now fails the delete instead of silently leaving a resurrectable replica.

## Deploy Plan

Front


id13 (approved):
LGTM

PR #30497, opened 2026-08-13
Keep consumption facet traces

flvndvd:
## Description

This PR. attempts at disabling sampling on the dd trace so we can understand the slowness despite this being under FF and not actively used.


flvndvd:
PMRR.

PR #30499, opened 2026-08-13
fix: group programmatic origins into one consumption source

adrsimon:
## Description

In consumption analytics, `cli` and `cli_programmatic` (same for `triggered` / `triggered_programmatic`, and `slack` / `slack_workflow`) were two separate values of the `source` dimension. The filter panel showed two "CLI" options, and the sources table and chart split one surface across two rows.

Consumption documents now carry `normalized_origin`: `context_origin` with its programmatic variants folded into the surface they belong to. The `source` dimension points at that field, so filtering, facets, rankings and the chart breakdown group without any of them knowing programmatic origins exist.

- `source_labels.ts` holds the pair map (`cli_programmatic → cli`, `triggered_programmatic → triggered`, `slack_workflow → slack`) and `normalizeOrigin`, used when building the document.
- The facet catalog offers one option per surface.

## Tests

Test suite passes, plus a case asserting a `cli_programmatic` message indexes as `normalized_origin: "cli"`, and one keeping the pair map and the offered sources consistent. The mapping update and the backfill script were both run against a local Elasticsearch: the backfill sets the field for every document that has a `context_origin` and skips the ones already backfilled.

## Risk

Low. Additive keyword field, no reindex. Between the mapping update and the deploy, new documents simply have no `normalized_origin` and are backfilled by the last step.

## Deploy Plan

0. block deploys
1. `tsx front/scripts/execute_elasticsearch_http.ts --file front/lib/analytics/migrations/20260814_add_normalized_origin_to_agent_message_consumption_analytics_1.http --execute` (per region).
1b. release deploys
2. Deploy front.
3. `tsx front/scripts/execute_elasticsearch_http.ts --file front/lib/analytics/migrations/20260814_backfill_normalized_origin_to_agent_message_consumption_analytics_1.http --execute` (per region).


front/lib/api/analytics/consumption/scope.ts
```diff
@@ -147,10 +148,12 @@ export function buildConsumptionScopeQuery({
   ];
 
   for (const dimension of CONSUMPTION_SCOPE_DIMENSIONS) {
+    const values = filter[CONSUMPTION_DIMENSION_FILTER_KEYS[dimension]];
     filters.push(
       ...termFilter(
         CONSUMPTION_DIMENSION_FIELDS[dimension],
-        filter[CONSUMPTION_DIMENSION_FILTER_KEYS[dimension]]
+        values &&
+          (dimension === "source" ? values.flatMap(originsForSource) : values)
```
flvndvd: Not a big fan that it touches all of those files, any way we could centralize in one place and avoid the local exceptions in many different places?

front/lib/api/analytics/source_labels.ts
```diff
@@ -44,3 +44,28 @@ export function sourceLabelForOrigin(origin: string): string | undefined {
     ? SOURCE_ORIGIN_LABELS[origin]
     : undefined;
 }
+
+// Analytics reports on the surface a message came from, not on how it was sent,
+// so these origins count as their regular counterpart.
+export const SOURCE_BY_PROGRAMMATIC_ORIGIN: Record<
```
flvndvd: IMHO this is a bit too local to age well, can't we ensure that TS requires any one who's willing to add a new analytics origin to classify them?

flvndvd (commented):
LGTM

flvndvd (approved):

PR #30503, opened 2026-08-13
fix(agents): revert agents wrongly swept to Auto by the Sonnet 4.6 migration

Nils-Fedrigo:
## Description

`20260810_migrate_sonnet46_medium_to_auto.ts` moved every active agent on Claude Sonnet 4.6 with
medium reasoning to the Auto meta-model. That sweep also caught agents that had only landed on
Sonnet 4.6 medium after another model change so it buried a model the builder had actually
picked.

This migration rolls those agents back to Sonnet 4.6 medium. Nothing was on Auto before the sweep,
so any version on Auto created by the end of the migration window is one the sweep flipped in
place. From there it walks the agent's full version history:

- if any version below the switch is on something other than Sonnet 4.6 / 4.5 medium, the agent had
  a deliberate model choice and goes back to Sonnet 4.6 medium;
- an agent that has only ever been on Sonnet 4.6 / 4.5 medium keeps Auto;
- a switch to Auto made after the window closed is a builder's own choice and is left alone.

It reverts the unbroken run of Auto versions starting at the switch, so versions saved after the
migration (which inherited Auto from it) are fixed too. A non-Auto version ends the run — an Auto
version above it is a fresh, deliberate pick.

Supersedes the approach in #30458.

## Tests

`front/migrations/20260813_revert_sonnet46_to_auto_migration.test.ts` covers the decision function
over every history shape (legacy model below the switch, genuine Sonnet 4.6 medium, re-saved
agents, Auto re-picked after the sweep, late switch, window boundary to the millisecond), plus an
end-to-end run against the database asserting the dry run mutates nothing, the execute run rewrites
only the intended versions, and a rerun is a no-op.

Run a dry run before executing:
`npx tsx front/migrations/20260813_revert_sonnet46_to_auto_migration.ts` and check the logged agent
and version counts.

## Risk

The migration rewrites the model on existing agent configuration versions, including archived ones.
The window gate and the "only ever Sonnet 4.6 medium" check keep it off agents that chose Auto
themselves, and rerunning is a no-op, but a bad match is not automatically reversible — review the
dry-run output before executing.

## Deploy Plan

Run the script manually after deploy, dry run first, then with `--execute`.

### DIFF
--- a/front/migrations/20260813_revert_sonnet46_to_auto_migration.test.ts
+++ b/front/migrations/20260813_revert_sonnet46_to_auto_migration.test.ts
@@ -1,0 +1,521 @@
+import { AgentConfigurationModel } from "@app/lib/models/agent/agent";
+import { generateRandomModelSId } from "@app/lib/resources/string_ids_server";
+import baseLogger from "@app/logger/logger";
+import type { AgentModelVersion } from "@app/migrations/20260813_revert_sonnet46_to_auto_migration";
+import {
+  planAutoRevert,
+  revertSonnet46AutoSwitch,
+} from "@app/migrations/20260813_revert_sonnet46_to_auto_migration";
+import { createResourceTest } from "@app/tests/utils/generic_resource_tests";
+import type { AgentStatus } from "@app/types/assistant/agent";
+import {
+  CLAUDE_4_SONNET_20250514_MODEL_ID,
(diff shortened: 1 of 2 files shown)


theogz (approved):
lgtm!

PR #30505, opened 2026-08-13
[Activation] Batch evaluation when a workspace exceeds the ES user cap

frankaloia:
## Summary
- We had a hard limit of one ES call where it was not paginated. This caused a hard limit of ~100 users in a workspace at each time. We want to paginate ES calls to avoid this limit safely.

## Testing
* Ran Temporal workflow normally
* new unit test

## Risk
Low - Should be following best practices to mitigate ES load

## Deploy
1. Deploy front

### DIFF
--- a/front/lib/api/activation/queries/user_day_cells.ts
+++ b/front/lib/api/activation/queries/user_day_cells.ts
@@ -8,6 +8,7 @@ import { AGENT_MESSAGE_STATUSES_TO_TRACK } from "@app/types/assistant/conversati
 import type { Result } from "@app/types/shared/result";
 import { Err, Ok } from "@app/types/shared/result";
 import type { estypes } from "@elastic/elasticsearch";
+import chunk from "lodash/chunk";
 
 // ---------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


front/lib/api/activation/queries/user_day_cells.ts
```diff
@@ -103,15 +103,39 @@ export async function fetchUserDayCells({
...
-    return new Err(
-      new Error(
-        `activation evaluation supports at most ${ELASTICSEARCH_MAX_USERS_PER_CALL} ` +
-          `users per call, got ${uniqueUserIds.length}`
-      )
-    );
+
+  const batches = chunk(uniqueUserIds, ELASTICSEARCH_MAX_USERS_PER_CALL);
+  for (const batch of batches) {
+    const batchResult = await fetchUserDayCellsBatch({
```
Nils-Fedrigo, line 109: sequencially means the call can still timeout at some point. I don't remember the limit so it must be reasonably high!

Nils-Fedrigo (approved):
lgtm! One comment

PR #30507, opened 2026-08-13
[front] enh: add richer tooltip for agents/skills in attribution table

aubin-tchoi:
## Description

Implement the following [Figma](https://www.figma.com/design/wJJMfVF6bluurSKfrEuysc/Product---WIP?node-id=13016-37581&m=dev) for the tooltips for agents and skills in the attribution table.

Instead of only the description we show a richer one with the avatar, name, model (only for agents) in a nicer display (kind of like a card).

image

image

## Tests

- Tested locally.

## Risk

- Low.

## Deploy Plan

- Deploy front.


front/components/workspace/analytics/consumption/ConsumptionAttributionTable.tsx
```diff
@@ -118,9 +177,12 @@ function buildColumns({
 
         return (
           <DataTable.CellContent className="w-full justify-start text-left">
-            {description ? (
+            {description && (dimension === "agent" || dimension === "skill") ? (
```
avervaet, line 180: Since skill descriptions are nullable, skills without a user-facing description will not display the new identity/icon card

avervaet (approved):

front/components/workspace/analytics/consumption/ConsumptionAttributionTable.tsx
aubin-tchoi: fair, but in this case the tooltip will be redundant with what we already see on the row

PR #30508, opened 2026-08-13
Avoid large system key permission queries

fontanierh:
## Description

Unscoped system keys currently expand every workspace group into the `group_permissions` query. One production workspace sends 8,181 group IDs on every connector request even though it has 182 permission rows.

Read the workspace grant set through the existing `workspaceId` index, then intersect it with the system key's resolved group IDs in memory. Explicitly scoped system keys and regular keys keep the existing group-ID query.

## Tests

Added database-backed coverage for unscoped system keys and explicitly scoped system keys.

## Risk

Low. The returned permission set is unchanged. Unscoped system keys trade a large SQL `IN` clause for a workspace grant scan, so workspaces with unusually large grant tables may return more rows before filtering.

## Deploy Plan

Standard deploy. Monitor US front DB CPU and the `group_permissions` Query Insights family after rollout.

### DIFF
--- a/front/lib/auth.ts
+++ b/front/lib/auth.ts
@@ -957,6 +957,11 @@ export class Authenticator {
       ? allGroups.map((g) => g.id)
       : [];
     const keyGroupModelIds = allGroups.map((g) => g.id);
+    // Unscoped system keys represent nearly every group in their workspace. Resolve their grants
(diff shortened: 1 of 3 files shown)


flvndvd (commented):
LGTM

PR #30511, opened 2026-08-13
Reuse MCP actions during consumption indexing

fontanierh:
## Description

Query Insights fingerprint `16697422976372582740` is running about 1.6k to 3.2k times per minute at 15 to 20 ms average latency, accounting for roughly 266 CPU-seconds per five minutes.

The retry-safe consumption analytics activity recomputes attribution, then indexing immediately reloads the same full `agent_mcp_actions` rows. Return the action snapshot from attribution and pass it into indexing instead. This removes one full-row query per normal v3 pass and one per indexing retry. The v3 path goes from three executions to two on a normal pass, and from two to one on an indexing retry.

## Tests

Updated the activity test to verify that indexing receives the exact action snapshot loaded during attribution. Ran the focused test locally, 4 tests passed.

## Risk

Low. Reuse is limited to two consecutive operations in the same activity. No Temporal workflow or activity name, signature, queue, or retry policy changes. Callers without a preloaded snapshot keep the existing fetch behavior. Rollback restores the redundant read.

## Deploy Plan

Standard deploy. Watch Query Insights fingerprint `16697422976372582740`, analytics activity failures, and the consumption attribution backlog after rollout.

### DIFF
--- a/front/temporal/analytics_queue/activities/consumption_attribution.ts
+++ b/front/temporal/analytics_queue/activities/consumption_attribution.ts
@@ -1,5 +1,8 @@
 import { indexAgentMessageConsumptionAnalytics } from "@app/lib/analytics/agent_message_consumption";
-import { computeAndStoreAgentMessageConsumptionAttribution } from "@app/lib/api/assistant/agent_message_consumption_attribution/store";
+import {
+  computeAndStoreAgentMessageConsumptionAttribution,
+  computeAndStoreAgentMessageConsumptionAttributionForAnalytics,
+} from "@app/lib/api/assistant/agent_message_consumption_attribution/store";
 import { publishConversationRelatedEvent } from "@app/lib/api/assistant/streaming/events";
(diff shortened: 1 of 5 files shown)


id13 (approved):

front/temporal/analytics_queue/activities/consumption_attribution.ts
```diff
@@ -45,10 +48,15 @@ export async function storeAgentMessageConsumptionAnalyticsActivity(
 
   // TEMPORARY: activity 2 retries independently, so workflows already stuck here will not rerun
   // their completed attribution activity. Recompute until that retry backlog has drained.
-  await computeAndStoreAgentMessageConsumptionAttribution(auth, message);
```
flvndvd, line 48: Actually this one can be dropped, will send a PR. Added yesterday as a band-aid.

flvndvd:
LGTM

PR #30512, opened 2026-08-13
Optimize conversation message fetch payload

fontanierh:
## Description

Query Insights fingerprint `8238549108092761451` executed about 0.8k to 2.3k times per minute at 18 to 24 ms average latency, accounting for roughly 124 CPU-seconds per five minutes during the incident.

Tracing the full SQL showed that it comes from the non-paginated `getConversation` path, not `fetchAgentMessageConsumptionAnalyticsContext`. The query preserves the existing conversation-scoped indexed joins but currently selects every column from `user_messages`, `agent_messages`, `content_fragments`, and `compaction_messages`.

This change limits each joined subtype to the fields consumed by message rendering. It avoids transferring and hydrating unused timestamps, tenancy columns, metadata, and growing `runIds` arrays while preserving joins, ordering, and rendered conversation output.

## Tests

- `NODE_ENV=test npm test -- lib/api/assistant/conversation/fetch.test.ts` (5 tests)
- `NODE_ENV=test npm test -- lib/api/assistant/conversation.test.ts` (84 tests)
- `npm run tsgo -- --noEmit`
- `npx biome check --error-on-warnings front/lib/api/assistant/conversation/fetch.ts`

## Risk

Risk is limited to accidentally omitting a subtype field needed by rendering. The full conversation suite exercises user, agent, content-fragment, and compaction rendering and passes. The change is safe to roll back by reverting this commit, which restores the previous full-column projection without a schema or data migration.

## Deploy Plan

- [ ] Deploy through the standard front production rollout.
- [ ] After deployment, identify the replacement SQL fingerprint in GCP Query Insights. The fingerprint changes because the selected column list changes.
- [ ] Compare its CPU time, average latency, execution rate, and rows returned against fingerprint `8238549108092761451`.
- [ ] If conversation rendering errors increase or database performance regresses, revert the commit and redeploy.

### DIFF
--- a/front/lib/api/assistant/conversation/fetch.ts
+++ b/front/lib/api/assistant/conversation/fetch.ts
@@ -57,6 +57,68 @@ type MessageTypeForView<V extends "light" | "full"> = V extends "light"
     ? MessageType
     : never;
 
+// Keep the hot conversation fetch from hydrating subtype columns that message rendering never
+// reads. In particular, agent and compaction runIds can grow over a message's lifetime.
+const USER_MESSAGE_RENDER_ATTRIBUTES: Array<keyof UserMessageModel> = [
+  "id",
+  "content",
+  "clientSideMCPServerIds",
+  "userContextUsername",
+  "userContextTimezone",
+  "userContextFullName",
+  "userContextEmail",
+  "userContextProfilePictureUrl",
+  "userContextOrigin",
+  "userContextLastTriggerRunAt",
+  "agenticMessageType",
+  "agenticOriginMessageId",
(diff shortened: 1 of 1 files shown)


fontanierh:
Closing after review. The sampled plans return only 2 to 4 rows with a 27 to 29 KB sort, while this change leaves the joins, scans, and call frequency unchanged. The likely DB benefit does not justify maintaining four manual renderer allowlists and the associated risk of silently omitting fields in future changes. A follow-up should first measure individual column sizes, then narrowly exclude only fields proven to be material.

PR #30516, opened 2026-08-13
[front] feat: opt-in Pod app sharing with workspace members

davidebbo:
## Description

Adds an opt-in **App sharing** setting on private Pods. When a Pod editor enables it, every active workspace member can invoke the Pod's published functions and fully use its shared Frames — without gaining any read or write on the Pod itself. This fills the gap where a shared Frame rendered workspace-wide (frame sharing defaults to `workspace_and_emails`) but `callFunction` 404'd for anyone who couldn't read the Pod, and the only workaround (open Pods / membership) hands out the whole write plane.

How it works:

- New `appSharingEnabled` boolean on `project_metadata` (pre-deploy migration), toggled through the existing editor-gated `project_metadata` PATCH, behind the `sandbox_functions` feature flag. Only offered on restricted Pods — an open Pod already lets every workspace member use its apps, so the settings tile is hidden there (UI-only gating; a redundant flag is harmless).
- Function resolution (`SandboxFunctionResource.baseFetch`, the single seam behind every fetch path) gains one branch: pod readable OR (pod has `appSharingEnabled` AND caller is an active workspace member). Per-function `userIdentity` policies (e.g. `pod_editor_required`) are untouched and keep applying on top.
- The public frame token route now hands `framePath` to workspace members of sharing Pods so shared Frames resolve bare function references; external email-grant viewers still get `null`.
- Sandbox provisioning no longer depends on the invoker's Pod permissions: new `DustFileSystem.forPodSandboxProvisioning` (same-workspace-checked, read-gate-free) at the two provisioning call sites, since a sharing-authorized invoker holds no read on the Pod and would otherwise fail every cold start, wake, or mount refresh.
- Pod settings gets an "App sharing" tile (editor-only, confirm-on-enable), and `appSharingEnabled` is serialized in the space GET + swagger.

What sharees still cannot do (pinned by regression tests): read/write/rename Pod files, delete or clone apps, publish or overwrite functions, change Pod settings. The trust statement is: sharees execute code authored by Pod members; the editor who enables sharing vouches for the Pod.

Out of scope, tracked as follow-ups: the open-Pod ACL issue where the global `project_viewer` group receives write, tightening publish from members to editors, agent-side (`dustPod`-configured) calling for sharees, and any discovery surface (distribution stays via shared Frame links).

## Tests

TDD throughout — each behavior landed with a failing test first. New functional tests: invocation route (sharee 201 with flag, 404 without, `pod_editor_required` still denied, pod-id/slug resolution), frame token route (`framePath` granted to sharing-Pod members, withheld without the flag and from external viewers), `project_metadata` PATCH (editor enable, feature-flag gate, non-editor 403), write-plane denials for sharees (app DELETE, file rename PATCH, `publishSandboxFunction`), and `forPodSandboxProvisioning` (Ok without read, Err cross-workspace). 180 tests across the seven touched files pass; `tsgo`, biome, swagger and audit-schema lints clean.

## Risk

This widens an authorization gate, so the main risk is over-granting. Mitigations: the new branch grants only function resolution/invocation (every write path keeps its existing gate, now regression-pinned); it is scoped to project spaces, active workspace members, and Pods that explicitly opted in; the flag defaults to false and only editors can set it. `forPodSandboxProvisioning` skips the viewer read gate but is used only by the two sandbox provisioning call sites, keeps a same-workspace check, and API-level mount permissions still derive from the caller (fail closed). Safe to roll back: reverting the code leaves an inert column.

## Deploy Plan

1. Pre-deploy migration adds the `appSharingEnabled` column (NOT NULL DEFAULT false, no rewrite).
2. Deploy front/front-api.


front-api/routes/w/[wId]/spaces/[spaceId]/index.ts
```diff
@@ -162,6 +162,9 @@ const app = workspaceApp();
  *                         isAdminControlled:
  *                           type: boolean
  *                           description: Whether workspace admins control membership and connected data for this Pod.
+ *                         appSharingEnabled:
+ *                           type: boolean
+ *                           description: Whether all workspace members may invoke this Pod's published functions.
```
Comment (davidebbo) [inline front-api/routes/w/[wId]/spaces/[spaceId]/index.ts:167]: I don't quite understand why we have a swagger for this private API...

davidebbo:
Superseded by #30532

PR #30521, opened 2026-08-13
[front] feat: add search on attribution table

aubin-tchoi:
## Description

In the new analytics page, we currently have a client-side search.
https://github.com/dust-tt/dust/pull/30487 added pagination, keeping the client-side search for now (so only affects the first page, wasn't an issue before because we showed only the first page).

This PR adds a search handled by the backend.
It's a very crude version, we can't actually search within labels (agent/skill/tool/user names) so will be quite slow for the dimensions that are slow to load. The goal here is really to reach a functional version first, then see exactly where to put our efforts in.

## Tests

- Added tests.
- Tested locally.

## Risk

- Low.

## Deploy Plan

- Deploy front.


avervaet (approved):
LGTM

PR #30522, opened 2026-08-13
[front] fix: exclude expired pending invitations from plan seat count

matthieuhaguen:
## Description

Each Dust plan (Free Pilot, Enterprise Pilot 100, etc.) enforces a `maxUsers` seat limit, computed as `active members + pending invitations`. 
That pending-invitation count included invitations whose 7-day acceptance link had already expired but were never explicitly revoked, so stale invitations kept occupying a seat indefinitely, blocking admins from sending new invites and blocking new users from auto-joining via verified domain on signup, even though the seat was never actually usable.

This PR excludes expired pending invitations from that count (`MembershipInvitationResource.getPendingInvitationsCountForWorkspace`), using the same expiration rule already used elsewhere in the codebase (`isExpired()`, and the admin invitations list which already hides expired invites by default). 
Both places that enforce `maxUsers` (`handleMembershipInvitations` (new invites) and `evaluateWorkspaceSeatAvailability` (signup auto-join)) call this same function, so fixing it here fixes both paths. No DB migration is needed: expiration is derived (`max(reminderSentAt, createdAt) + 7 days`), never stored.

## Tests

Added two functional tests in `front-api/routes/w/[wId]/invitations/index.test.ts`:
- an expired pending invitation no longer blocks a new invite once the plan would otherwise be full
- a fresh (non-expired) pending invitation still correctly blocks it (regression guard against over-relaxing the check)

Ran and confirmed no regressions in the related suites:
- `routes/w/[wId]/invitations/index.test.ts` — 12/12 passing
- `routes/w/[wId]/invitations/[iId]/index.test.ts` — 8/8 passing
- `lib/api/plan_limit_overrides.test.ts` (front) — 8/8 passing
- `npx tsgo --noEmit` — clean

## Risk

Low. This only relaxes an over-counting bug.
Billing (Metronome) is unaffected, since seat billing already only counts active members, never pending invitations. 
The only externally visible change: workspaces with old, un-actionable pending invitations will immediately regain the seats those invitations were wrongly holding.
Trivially revertible (single function, no schema/data change) if unexpected behavior surfaces.

## Deploy Plan

Standard deploy, no special steps. No migration, no feature flag, no ordering constraints.
The new counting logic takes effect as soon as the deploy lands.

### DIFF
--- a/front-api/routes/w/[wId]/invitations/index.test.ts
+++ b/front-api/routes/w/[wId]/invitations/index.test.ts
@@ -2,6 +2,8 @@ import { Authenticator } from "@app/lib/auth";
 import { MembershipInvitationResource } from "@app/lib/resources/membership_invitation_resource";
 import { createPrivateApiMockRequest } from "@app/tests/utils/generic_private_api_tests";
 import { MembershipInvitationFactory } from "@app/tests/utils/MembershipInvitationFactory";
+import { PlanFactory } from "@app/tests/utils/PlanFactory";
+import { WorkspaceFactory } from "@app/tests/utils/WorkspaceFactory";
 import sgMail from "@sendgrid/mail";
 import { beforeEach, describe, expect, it, vi } from "vitest";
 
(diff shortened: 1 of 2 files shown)


tdraier (approved):
LGTM 🙏

matthieuhaguen:
Thanks @tdraier ! I see that some checks were not successful but I don't have access to Vercel to see why.
Is it ok to Squash and merge ?

tdraier:
> Thanks @tdraier ! I see that some checks were not successful but I don't have access to Vercel to see why. Is it ok to Squash and merge ?

yes completely fine, vercel is not needed here

PR #30526, opened 2026-08-13
Fix GPT-5 tool footprint attribution

flvndvd:
## Description

Tool result footprints were overestimated because GPT-5 models used `r50k_base` with a 15% safety margin. This could
make the attributed tool footprint larger than the provider’s billed input and leave consumption attribution incomplete, causing analytics indexing to retry indefinitely.

This changes the shared GPT-5 model configurations to `o200k_base` and removes context-window safety padding when
computing attribution footprints. Attribution now uses a multiplier of 1.

For the production failure investigated, OpenAI reported 20,473 input tokens while `o200k_base` counted 20,467 content
tokens. The previous calculation stored `29,991` tokens.

The attribution version is bumped so retried workflows recompute existing records.


fontanierh (approved):
LGTM

PR #30527, opened 2026-08-13
Speed up Dust Hive environment creation

fontanierh:
## Description

Skip nested local worktree caches when copying developer configuration into a new Dust Hive environment. These caches can contain full repository checkouts and package trees, making cold creation take several minutes.

## Tests

Added regression coverage for config merging and cache exclusion. Timed cold creation at 326.53s before and 9.88s after.

## Risk

Low. The change is limited to local Dust Hive setup, and regular settings and skills are still copied.

## Deploy Plan

Standard deploy.


aubin-tchoi (approved):
LGTM

PR #30528, opened 2026-08-13
Fix EU hosted model billing uplift

flvndvd:
### Chapter 3 in our journey to reconcile costs

EU-hosted Anthropic and OpenAI models carry a 10% pricing uplift.

Our legacy Anthropic integration explicitly marked European requests with `inferenceRegion: "eu"`. That signal disappeared
when the legacy clients were removed in https://github.com/dust-tt/dust/commit/10d5e4d967270865bde00a4fb6afda3614b664a8. The new endpoint layer still knew its region, but that information was not reaching billing, which (very) quietly fell back to global pricing.

The regional pricing table had also fallen behind. It only covered a hardcoded subset of Anthropic models and none of the eligible OpenAI models, meaning that knowing a request was European was not always enough to price it correctly.

This PR reconnects the selected endpoint region to billing for both streaming and batch inference. When a request is served
from Europe and its model is eligible for the provider uplift, we now apply the 1.1× tariff. The EU rates are derived directly from global prices, including cache and long-context pricing, so they remain aligned as base prices evolve. Models without an EU-specific uplift continue to use global pricing.

### Pricing references

- [Anthropic Vertex AI pricing](https://platform.claude.com/docs/en/build-with-claude/claude-on-vertex-ai)
- [OpenAI API pricing](https://openai.com/api/pricing/)

### DIFF
--- a/front/lib/api/assistant/token_pricing/eu.ts
+++ b/front/lib/api/assistant/token_pricing/eu.ts
@@ -1,58 +1,98 @@
 import type { PricingEntry } from "@app/lib/api/assistant/token_pricing/global";
+import { MODEL_PRICING } from "@app/lib/api/assistant/token_pricing/global";
+import {
+  CLAUDE_4_5_HAIKU_20251001_MODEL_ID,
+  CLAUDE_4_5_OPUS_20251101_MODEL_ID,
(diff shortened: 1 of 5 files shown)


aubin-tchoi (approved):
LGTM

PR #30529, opened 2026-08-13
feat(sandbox): request_egress_domain tool and publish-time domain requests

smb2268:
## Description

Follow up of `30474` (stacked on it). Adds the two **request producers** for the sandbox egress allowlist — an agent tool and a publish side-effect. Both only ever file a domain as a request a workspace admin reviews; neither grants access.

- **`request_egress_domain`** — the tool an agent calls for a domain a sandbox will need. One arg beyond the domain: **scope**.
  - `pod` (default) — for a Pod's published functions; recorded on the Pod's allowlist. Only usable inside a Pod.
  - `workspace` — for every sandbox in the workspace.
  - Distinct from `add_egress_domain` (unchanged), which grants there-and-then for the current conversation.
- **`publish`** — declaring a function's runtime `domains` files each as a `pod` request through the same helper.

Response copy reflects the outcome (`Requested` / `Already allowed` / `Already requested`).

**Stakes:** `request_egress_domain` is **low** (it only queues a request); `add_egress_domain` stays **high** (grants in-conversation, needs user approval).

**Gating:** both require Computer. `add_egress_domain` is additionally gated on the on-the-fly self-serve toggle; `request_egress_domain` on **`sandbox_functions`**, which pairs it with the Pod/workspace network settings that review its requests (Pod settings are `sandbox_functions`-only, Computer is on by default).

Stacked: the admin surface that reviews these requests (pod + workspace) is `29347`.

## Risk

Low. Both producers are request-only — they never write `allowedDomains`. Worst case a lost read-merge-write race drops a queued request; re-creatable, never touches the allowlist. Safe to rollback.

## Tests

`request_egress_domain` handler: pod request on the Pod's file; workspace request on the workspace file; errors on `pod` scope outside a Pod; wildcard accepted; already-allowed / already-requested surfaced. Tool gating across Computer / `sandbox_functions` / the self-serve toggle. Publish: declared domains recorded as Pod requests; no note when none declared.

### DIFF
--- a/front/lib/api/actions/servers/sandbox/metadata.ts
+++ b/front/lib/api/actions/servers/sandbox/metadata.ts
@@ -102,9 +102,7 @@ export const SANDBOX_TOOLS_METADATA = [
       "allowlist surface as denied entries in `<network_proxy_logs>` in " +
       "the bash tool output. Allowlist entries added through this tool " +
       "persist for the lifetime of the conversation (across sandbox " +
-      "restarts) and are scoped to this conversation only — including " +
-      "inside a Pod, where Pod-wide domains are managed by admins in the " +
-      "Pod's network settings.",
+      "restarts) and are scoped to this conversation only.",
     schema: {
       domain: z
         .string()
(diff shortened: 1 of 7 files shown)


front/lib/api/actions/servers/sandbox/tools/index.ts
```diff
@@ -620,3 +642,73 @@ export async function addEgressDomainTool(
...
+  const parsed = normalizeEgressPolicyDomain(domain);
+  if (parsed.isErr()) {
+    return new Err(new MCPError(parsed.error.message));
+  }
+
+  // Records a request for admin review — the same for every caller, admins
+  // included (they approve it from the network settings). The agent never
+  // grants and never reasons about who is privileged; scope picks which
+  // allowlist (and reviewer) the request targets.
+  const requestResult = await (async () => {
```
flvndvd: IIFE 😱

front/lib/api/actions/servers/sandbox/metadata.ts
```diff
@@ -130,6 +130,44 @@ export const SANDBOX_TOOLS_METADATA = [
...
+  {
+    name: "request_egress_domain",
+    description:
+      "Request that a domain be added to a shared network allowlist so " +
+      "sandboxes can reach it, subject to workspace-admin approval — it " +
+      "never grants access on its own. Use `pod` scope (the default choice) " +
+      "for a domain a published Pod function needs; use `workspace` scope " +
+      "only when every sandbox in the workspace needs it. `pod` scope is " +
+      "only usable inside a Pod. To unblock a command you are running right " +
+      "now, use add_egress_domain instead. One domain per call — an exact " +
```
flvndvd: Can we use a constant for `add_egress_domain` tool name. Aren't we worried that we expose pods a bit too often here? Conversations are still mainly created outside of pods those days.

Maybe we should have one tool description per context?

flvndvd (approved):

front/lib/api/actions/servers/sandbox/metadata.ts
smb2268: yeah I debated having a separate tool for requesting pod domains and workspace domains but I'm not convinced it's necessary. let me try to simplify these and see how it performs over a few days and then maybe I'll reconsider

front/lib/api/actions/servers/sandbox/tools/index.ts
smb2268: oops fixing

PR #30530, opened 2026-08-13
Wait for all tool activities before finalizing cancelled agent runs

flvndvd:
## Description

In our journey to ensure analytics data always sums up to billed credits, we found cancelled agent messages being finalized and billed while tool activities were still running. Those tools could later succeed and add direct credits that were absent from the message’s bill. Attribution would then find more non-input credits than total billed credits, leaving no valid input-credit remainder and failing as incomplete.

There were two cancellation issues behind this:
1. Although Temporal’s TypeScript documentation describes waiting for activity cancellation as the default, the SDK version we use leaves the cancellation mode unset. At the protocol level, the default value is `TRY_CANCEL`. When the agent workflow was cancelled, Temporal therefore released the activity promises immediately while the underlying tools continued running.
2. Even with the correct cancellation mode, the tools were grouped using fail-fast promise aggregation. If one tool acknowledged cancellation before its siblings, the aggregate rejected immediately and allowed message finalization to begin while the remaining tools were still settling.

This change explicitly configures tool activities with `WAIT_CANCELLATION_COMPLETED`, ensuring each activity promise remains pending until the tool has completed, failed, or acknowledged cancellation. It also waits for every tool activity to settle before propagating a failure, ensuring billing and attribution only start once the complete set of tool outcomes is known.

The behavior is protected by a Temporal patch. Existing workflow histories replay the original path, while new histories use the corrected cancellation and aggregation behavior.

### DIFF
--- a/front/temporal/agent_loop/lib/wait_for_all_promises.test.ts
+++ b/front/temporal/agent_loop/lib/wait_for_all_promises.test.ts
@@ -1,0 +1,34 @@
+import { describe, expect, it } from "vitest";
+
+import { waitForAllPromises } from "./wait_for_all_promises";
+
+describe("waitForAllPromises", () => {
+  it("preserves result order after every promise succeeds", async () => {
+    await expect(
+      waitForAllPromises([Promise.resolve("first"), Promise.resolve("second")])
+    ).resolves.toEqual(["first", "second"]);
+  });
+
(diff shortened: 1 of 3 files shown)


aubin-tchoi (approved):
LGTM

PR #30531, opened 2026-08-13
Add For You email opt-out and click tracking

frankaloia:
## Summary
- Currently, activation nudge emails abide by the same frequency as all emails (the default is at maximum you can receive one email a week across the system)
- This is creating the biggest bottleneck in the funnel. Instead, we are not abiding by the email frequency and surfacing a new toggle to be able to turn off the for you notifications independently. 

## Testing
Screenshot 2026-08-13 at 3 59 01 PM

## Risk
Low

## Deploy
1. Front

### DIFF
--- a/front/components/me/NotificationPreferences.tsx
+++ b/front/components/me/NotificationPreferences.tsx
@@ -61,7 +61,7 @@ const NotificationPreferencesFormSchema = z.object({
   inApp: z.boolean(),
   slack: z.boolean(),
   email: z.boolean(),
-  allowForYouNotifications: z.boolean(),
+  forYou: z.boolean(),
 });
 
 type NotificationPreferencesFormValues = z.infer<
@@ -116,7 +116,7 @@ export function useNotificationPreferencesForm({
       inApp: false,
       slack: false,
       email: false,
-      allowForYouNotifications: true,
+      forYou: true,
     },
   });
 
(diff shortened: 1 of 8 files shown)


front/components/me/NotificationPreferences.tsx
```diff
@@ -104,6 +116,7 @@ export function useNotificationPreferencesForm({
       inApp: false,
       slack: false,
       email: false,
+      allowForYouNotifications: true,
```
aubin-tchoi: should be only `forYou` out of consistency

front/components/me/NotificationPreferences.tsx
```diff
@@ -313,6 +346,19 @@ export function NotificationPreferences({
           </DropdownMenu>
         }
       />
+
+      {displayForYouOption && (
+        <SettingsList.Row
+          title="Allow For You Notifications"
```
aubin-tchoi: OOC why not simply "For you"

front/components/me/NotificationPreferences.tsx
```diff
@@ -313,6 +346,19 @@ export function NotificationPreferences({
           </DropdownMenu>
         }
       />
+
+      {displayForYouOption && (
```
aubin-tchoi: not sure this really belongs here, this category lists the channels where the user wants to receive notifications, not which notifications to send out

front/components/pages/workspace/GetStartedPage/index.tsx
```diff
@@ -136,6 +142,20 @@ export function GetStartedPage() {
 
   const firstName = user?.firstName ?? user?.fullName?.split(" ")[0] ?? "there";
 
+  useLayoutEffect(() => {
```
aubin-tchoi, line 147: any reason to not use a `useEffect` instead?

aubin-tchoi (approved):

front/components/pages/workspace/GetStartedPage/index.tsx
frankaloia: RootLayout strips UTM params in a `useEffect` (after paint). This has to read `utm_campaign` before that runs, so it needs `useLayoutEffect` (before paint). A `useEffect` here can miss the click.

PR #30532, opened 2026-08-14
[front] feat: Add support for sharing Pod apps to the workspace or invited (workspace) users

davidebbo:
## Description

Lets workspace members *use* a Pod app — open its frame, click around, call the functions behind it — without granting any access to the Pod itself. Today a shared frame renders for every workspace member (frame sharing defaults to `workspace_and_emails`), but every `callFunction` fails because function resolution requires read access on the Pod; the only workarounds (pod membership, open Pods) hand out the whole write plane.

The model: **a workspace member who can open a frame can use its app** — invocation access follows frame view access. For workspace-visible frames that means holding the share link (an unguessable token); for invite-only frames it additionally requires an active email grant, exactly like viewing does. Sharing the frame is sharing the app — no new mode, no new state, no migration.

Mechanics:

- Invocation requests may carry the frame's share token in an `x-dust-frame-share-token` header. A workspace member presenting a valid token may resolve and invoke the functions of **that frame's app folder** — per-app scoping falls out of the frame's path, which is the app's identity. Validation (`resolveFrameShareCapability`): same workspace, pod app frame, prefix match, and a qualifying scope — workspace-visible, or invite-only (`emails_only`) when the member holds an active email grant, so revoking a grant revokes invocation individually.
- Resolution is lazy: the caller's own access is tried first, and the token is only validated when that misses — members with pod access never pay the token lookup, and an invalid token behaves exactly like an absent one.
- The token is accepted on the invocation POST, `validate-action`, `resolve-authentication`, and the SSE events stream; invocation-level gates (owner-only reads) are unchanged — the capability only widens which functions resolve, never who may read or act on an invocation. Per-function `userIdentity` policies (e.g. `pod_editor_required`) still apply on top.
- **Pipeline paths resolve by invocation identity instead of caller permissions** (`fetchByIdForPipeline`): temporal activities and sandbox-token callbacks re-enter resolution under an auth that cannot carry the original grant (a serialized session snapshot, a sandbox JWT). The invocation row is their proof of authorization — creation passed the caller-facing gates and the per-function `userIdentity` policy re-runs at execution.
- The public frame route hands `framePath` to workspace members whose token qualifies (workspace-visible scopes, or invite-only with a grant); external email-grant viewers still get `null` and can never invoke (this is a missing scenario we should tackle later).
- Client: the shared-frame host threads the token to the three fetch points (invocation POST, SSE via a new `headers` option on `useEventSource`, and the blocked-action cards via `FrameViewer`). Authenticated pod hosts pass nothing and are unaffected.
- Sandbox provisioning no longer depends on the invoker's pod permissions (`DustFileSystem.forPodSandboxProvisioning` at the two provisioning call sites): the sandbox is a pod-level resource whose mounts are identical whoever triggers the boot, and a token-authorized invoker holds no read on the Pod.

## Tests

New tests added.

## Risk

Authorization change. The capability grants function resolution only, for one app, to authenticated workspace members presenting an unguessable token — every write-plane gate is untouched and regression-pinned. The riskiest piece is the pipeline switch from caller-permission resolution to invocation-identity resolution; it is confined to paths that execute an already-validated invocation (workflow args, sandbox JWT claims), still workspace-scoped, and the identity policy re-runs at execution. Rollback is a plain revert (no schema changes).

## Deploy Plan

No migration, no flags. Deploy front/front-api together (the client header is additive; old clients simply don't send it).

### DIFF
--- a/front-api/routes/v1/public/frames/[token]/index.ts
+++ b/front-api/routes/v1/public/frames/[token]/index.ts
@@ -247,9 +247,17 @@ app.get(
       isPodMember,
       isPodEditor,
       // Lets a shared Frame in an app folder resolve bare function references, exactly as it does
-      // when opened from the Pod. Withheld from viewers who cannot read the Pod, who cannot invoke
-      // its functions either.
-      framePath: canRead && auth ? file.toScopedPath(auth) : null,
+      // when opened from the Pod. For workspace-visible scopes the share token itself is a
+      // workspace member's capability to invoke the frame's app's functions, so members get the
+      // path even without pod read. External viewers cannot invoke and never receive it.
+      framePath:
+        auth &&
+        (canRead ||
+          ((shareScope === "workspace_and_emails" ||
+            shareScope === "workspace") &&
+            auth.isUser()))
+          ? file.toScopedPath(auth)
+          : null,
     });
   }
 );
--- a/front-api/routes/v1/public/frames/[token]/index.ts
+++ b/front-api/routes/v1/public/frames/[token]/index.ts
@@ -256,9 +256,9 @@ app.get(
       isPodMember,
       isPodEditor,
       // Lets a shared Frame in an app folder resolve bare function references, exactly as it does
-      // when opened from the Pod. For workspace-visible scopes the share token itself is a
-      // workspace member's capability to invoke the frame's app's functions, so members get the
-      // path even without pod read. External viewers cannot invoke and never receive it.
+      // when opened from the Pod. Workspace members who may view the frame may invoke its app's
+      // functions, so they get the path even without pod read. External viewers cannot invoke
+      // and never receive it.
       framePath:
         auth && (canRead || canInvokeViaShareCapability)
           ? file.toScopedPath(auth)
--- a/front-api/routes/v1/public/frames/[token]/index.ts
+++ b/front-api/routes/v1/public/frames/[token]/index.ts
@@ -219,9 +219,12 @@ app.get(
       space.canAdministrate(auth)
     );
     // The share token this viewer used is a workspace member's capability to invoke the frame's
-    // app's functions.
+    // app's functions. For invite-only frames, a member only reaches this point with an active
+    // email grant (or as the frame's owner), which is the same rule the invocation gate applies.
     const canInvokeViaShareCapability =
-      isWorkspaceVisibleShareScope(shareScope) && !!auth?.isUser();
+      !!auth?.isUser() &&
+      (isWorkspaceVisibleShareScope(shareScope) ||
+        shareScope === "emails_only");
 
     // Generate access token for viz rendering.
     const accessToken = generateVizAccessToken({
--- a/front-api/routes/v1/public/frames/[token]/index.ts
+++ b/front-api/routes/v1/public/frames/[token]/index.ts
@@ -10,7 +10,10 @@ import { SpaceResource } from "@app/lib/resources/space_resource";
 import { WorkspaceResource } from "@app/lib/resources/workspace_resource";
 import { getConversationRoute, getPodRoute } from "@app/lib/utils/router";
 import logger from "@app/logger/logger";
-import { isInteractiveContentType } from "@app/types/files";
+import {
+  isInteractiveContentType,
+  isWorkspaceVisibleShareScope,
+} from "@app/types/files";
 import type { PublicFrameResponseBodyType } from "@dust-tt/client";
 import { unauthedApp } from "@front-api/middlewares/ctx";
 import type { HandlerResult } from "@front-api/middlewares/utils";
@@ -142,8 +145,7 @@ app.get(
 
       // For workspace_and_emails (and legacy "workspace"): workspace members are authorized directly.
       const isWorkspaceMemberWithAccess =
-        (shareScope === "workspace_and_emails" || shareScope === "workspace") &&
-        auth;
+        isWorkspaceVisibleShareScope(shareScope) && auth;
 
       if (!isFileOwner && !isWorkspaceMemberWithAccess) {
         // Resolve the verified email: prefer Dust session, fall back to external viewer cookie.
@@ -216,6 +218,10 @@ app.get(
       auth &&
       space.canAdministrate(auth)
     );
+    // The share token this viewer used is a workspace member's capability to invoke the frame's
+    // app's functions.
+    const canInvokeViaShareCapability =
+      isWorkspaceVisibleShareScope(shareScope) && !!auth?.isUser();
 
     // Generate access token for viz rendering.
     const accessToken = generateVizAccessToken({
@@ -251,11 +257,7 @@ app.get(
       // workspace member's capability to invoke the frame's app's functions, so members get the
       // path even without pod read. External viewers cannot invoke and never receive it.
       framePath:
-        auth &&
-        (canRead ||
-          ((shareScope === "workspace_and_emails" ||
-            shareScope === "workspace") &&
-            auth.isUser()))
+        auth && (canRead || canInvokeViaShareCapability)
           ? file.toScopedPath(auth)
           : null,
     });
--- a/front-api/routes/v1/w/[wId]/sandbox/sandbox-functions/result.ts
+++ b/front-api/routes/v1/w/[wId]/sandbox/sandbox-functions/result.ts
@@ -50,13 +50,14 @@ app.post(
       return ctx.json({ success: true });
     }
 
-    // Execution-side resolution: a sandbox-token auth cannot carry the invoker's original grant (e.g.
-    // a frame share token); the validated claims name the invocation, which is the proof.
-    const sandboxFunction = await SandboxFunctionResource.fetchByIdForExecution(
+    // Execution-side resolution: a sandbox-token auth cannot carry the invoker's original grant
+    // (e.g. a frame share token). The id comes from signature-verified sandbox JWT claims minted
+    // at execution start, so the space filter is deliberately skipped.
+    const sandboxFunction = await SandboxFunctionResource.fetchById(
       auth,
+      sandboxClaims.sandboxFunctionId,
       {
-        sandboxFunctionId: sandboxClaims.sandboxFunctionId,
-        invocationId: sandboxClaims.invocationId,
+        dangerouslyBypassSpacePermissionFilter: true,
       }
     );
     if (!sandboxFunction) {
--- a/front-api/routes/v1/w/[wId]/sandbox/sandbox-functions/result.ts
+++ b/front-api/routes/v1/w/[wId]/sandbox/sandbox-functions/result.ts
@@ -50,9 +50,12 @@ app.post(
       return ctx.json({ success: true });
     }
 
-    const sandboxFunction = await SandboxFunctionResource.fetchById(
+    // Pipeline resolution: a sandbox-token auth cannot carry the invoker's original grant (e.g.
+    // a frame share token); the validated claims name the invocation, which is the proof.
+    const sandboxFunction = await SandboxFunctionResource.fetchByIdForPipeline(
       auth,
-      sandboxClaims.sandboxFunctionId
+      sandboxClaims.sandboxFunctionId,
+      { invocationId: sandboxClaims.invocationId }
     );
     if (!sandboxFunction) {
       return apiError(ctx, {
--- a/front-api/routes/v1/w/[wId]/sandbox/sandbox-functions/result.ts
+++ b/front-api/routes/v1/w/[wId]/sandbox/sandbox-functions/result.ts
@@ -50,9 +50,9 @@ app.post(
       return ctx.json({ success: true });
     }
 
-    // Pipeline resolution: a sandbox-token auth cannot carry the invoker's original grant (e.g.
+    // Execution-side resolution: a sandbox-token auth cannot carry the invoker's original grant (e.g.
     // a frame share token); the validated claims name the invocation, which is the proof.
-    const sandboxFunction = await SandboxFunctionResource.fetchByIdForPipeline(
+    const sandboxFunction = await SandboxFunctionResource.fetchByIdForExecution(
       auth,
       sandboxClaims.sandboxFunctionId,
       { invocationId: sandboxClaims.invocationId }
--- a/front-api/routes/v1/w/[wId]/sandbox/sandbox-functions/result.ts
+++ b/front-api/routes/v1/w/[wId]/sandbox/sandbox-functions/result.ts
@@ -50,9 +50,12 @@ app.post(
       return ctx.json({ success: true });
     }
 
-    const sandboxFunction = await SandboxFunctionResource.fetchById(
+    const sandboxFunction = await SandboxFunctionResource.fetchByIdForExecution(
       auth,
-      sandboxClaims.sandboxFunctionId
+      {
+        sandboxFunctionId: sandboxClaims.sandboxFunctionId,
+        invocationId: sandboxClaims.invocationId,
+      }
     );
     if (!sandboxFunction) {
       return apiError(ctx, {
--- a/front-api/routes/v1/w/[wId]/sandbox/sandbox-functions/result.ts
+++ b/front-api/routes/v1/w/[wId]/sandbox/sandbox-functions/result.ts
@@ -50,15 +50,9 @@ app.post(
       return ctx.json({ success: true });
     }
 
-    // Execution-side resolution: a sandbox-token auth cannot carry the invoker's original grant
-    // (e.g. a frame share token). The id comes from signature-verified sandbox JWT claims minted
-    // at execution start, so the space filter is deliberately skipped.
     const sandboxFunction = await SandboxFunctionResource.fetchById(
       auth,
-      sandboxClaims.sandboxFunctionId,
-      {
-        dangerouslyBypassSpacePermissionFilter: true,
-      }
+      sandboxClaims.sandboxFunctionId
     );
     if (!sandboxFunction) {
       return apiError(ctx, {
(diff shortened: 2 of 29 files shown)


front/types/files.ts
```diff
@@ -80,6 +80,14 @@ export const fileShareScopeSchema = z.enum([
 
 export type FileShareScope = z.infer<typeof fileShareScopeSchema>;
 
+/**
+ * Whether a share scope makes the file visible to every workspace member holding its link.
+ * "workspace" is the legacy spelling of "workspace_and_emails".
```
davidebbo, line 85: @flvndvd can't we migrate to not have to deal with two names for the same?

front-api/routes/v1/public/frames/[token]/index.ts
```diff
@@ -142,8 +145,7 @@ app.get(
 
       // For workspace_and_emails (and legacy "workspace"): workspace members are authorized directly.
       const isWorkspaceMemberWithAccess =
-        (shareScope === "workspace_and_emails" || shareScope === "workspace") &&
-        auth;
+        isWorkspaceVisibleShareScope(shareScope) && auth;
```
Comment (davidebbo) [inline front-api/routes/v1/public/frames/[token]/index.ts:148]: This is just clean up of existing code

front/lib/resources/sandbox_function_resource.ts
```diff
@@ -352,8 +389,52 @@ export class SandboxFunctionResource extends BaseResource<SandboxFunctionModel>
...
+  ): Promise<SandboxFunctionResource | null> {
+    if (
+      !isResourceSId("sandbox_function", sandboxFunctionId) ||
+      !isResourceSId("sandbox_function_invocation", invocationId)
+    ) {
+      return null;
+    }
+    const sandboxFunctionModelId = getResourceIdFromSId(sandboxFunctionId);
+    const invocationModelId = getResourceIdFromSId(invocationId);
+    if (sandboxFunctionModelId === null || invocationModelId === null) {
```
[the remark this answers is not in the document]
davidebbo: I think it's ok coming from another resource class

front-api/routes/w/[wId]/sandbox-functions/[functionId]/invocations.ts
```diff
@@ -107,9 +108,10 @@ app.post(
     const body: PostSandboxFunctionInvocationRequestBody =
       ctx.req.valid("json");
 
-    const sandboxFunction = await SandboxFunctionResource.fetchByIdOrSlug(
+    const sandboxFunction = await resolveSandboxFunctionWithCapability(
       auth,
-      functionIdOrSlug
+      functionIdOrSlug,
+      ctx.req.header(FRAME_SHARE_TOKEN_HEADER)
     );
```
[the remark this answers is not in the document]
Comment (davidebbo) [reply front-api/routes/w/[wId]/sandbox-functions/[functionId]/invocations.ts:115]: That is exactly the point of this change, so this is by design.

front-api/routes/v1/public/frames/[token]/index.ts
```diff
@@ -216,6 +218,13 @@ app.get(
...
+    // The share token this viewer used is a workspace member's capability to invoke the frame's
+    // app's functions. For invite-only frames, a member only reaches this point with an active
+    // email grant (or as the frame's owner), which is the same rule the invocation gate applies.
+    const canInvokeViaShareCapability =
+      !!auth?.isUser() &&
+      (isWorkspaceVisibleShareScope(shareScope) ||
+        shareScope === "emails_only");
 
     // Generate access token for viz rendering.
     const accessToken = generateVizAccessToken({
```
Comment (flvndvd) [inline front-api/routes/v1/public/frames/[token]/index.ts:230]: Do we have a deterministic way to know if a frame uses pod functions today? If yes, wondering if we don't want to add those in the token we mint here?

front/components/assistant/conversation/actions/VisualizationActionIframe.tsx
```diff
@@ -195,12 +196,20 @@ const getExtensionFromBlob = (blob: Blob): string => {
 export interface FrameViewer {
   owner: LightWorkspaceType;
   user: UserType;
+  /**
+   * Share token the viewer loaded the frame with, presented on invocation requests: the server
+   * grants invocation of the frame's app's functions to workspace members who may view the frame
+   * (invite-only frames also require an email grant). Authorization happens server-side on every
+   * request.
+   */
+  frameShareToken?: string;
```
flvndvd, line 205: IIRC this token has a very short TTL on purpose. Did we widen it here?

front/lib/resources/sandbox_function_resource.ts
```diff
@@ -334,20 +353,34 @@ export class SandboxFunctionResource extends BaseResource<SandboxFunctionModel>
...
+      ? new Map(spaces.map((space) => [space.id, space]))
+      : null;
+
     return sandboxFunctions.flatMap((sandboxFunction) => {
-      const space = accessibleSpacesById.get(sandboxFunction.get().spaceId);
-      const file = filesById.get(sandboxFunction.get().fileId);
+      const blob = sandboxFunction.get();
+      // Three prioritized gates: the caller's own access, then the app the capability grants,
+      // then the execution-side bypass (spacesById is only built for it).
+      const space =
```
flvndvd, line 365: Can we simplify this a bit and avoid the nested complexity.

front/lib/resources/sandbox_function_resource.ts
```diff
@@ -360,8 +393,52 @@ export class SandboxFunctionResource extends BaseResource<SandboxFunctionModel>
...
+   * Resolve a function for executing an existing invocation: temporal activities and sandbox-token
+   * callbacks re-enter resolution under an auth that cannot carry the caller's original grant (a
+   * serialized session snapshot, a sandbox token), so caller-facing space permissions do not
+   * apply there. The invocation row is the proof of authorization — creating it passed the
+   * caller-facing gates, and the userIdentity policy re-runs at execution — so resolution here is
+   * workspace-scoped only. Only for paths executing that specific invocation.
+   */
+  static async fetchByIdForExecution(
+    auth: Authenticator,
+    sandboxFunctionId: string,
```
flvndvd: `nit`
move this to the object so it's explicit at call site and consistent with `invocationId`.

front/lib/resources/sandbox_function_resource.ts
```diff
@@ -360,8 +393,52 @@ export class SandboxFunctionResource extends BaseResource<SandboxFunctionModel>
...
+    { invocationId }: { invocationId: string }
+  ): Promise<SandboxFunctionResource | null> {
+    if (
```
flvndvd: OOC do we need all of those?

front/lib/resources/sandbox_function_resource.ts
```diff
@@ -360,8 +393,52 @@ export class SandboxFunctionResource extends BaseResource<SandboxFunctionModel>
...
+    ) {
+      return null;
+    }
+    const sandboxFunctionModelId = getResourceIdFromSId(sandboxFunctionId);
+    const invocationModelId = getResourceIdFromSId(invocationId);
+    if (sandboxFunctionModelId === null || invocationModelId === null) {
+      return null;
+    }
+
+    const invocation = await SandboxFunctionInvocationModel.findOne({
```
flvndvd: Why not using the associated resource? We can't use `baseFetch` because it enforces `canReadOrAdministrate`, is that correct? IHMO the comments above the function is too abstract and not enough explicit.

front/lib/resources/sandbox_function_resource.ts
```diff
@@ -418,24 +495,29 @@ export class SandboxFunctionResource extends BaseResource<SandboxFunctionModel>
...
     });
 
     return sandboxFunction ?? null;
   }
 
   static async fetchByIdOrSlug(
     auth: Authenticator,
-    functionIdOrSlug: string
+    functionIdOrSlug: string,
+    { capability }: { capability?: FrameShareCapability } = {}
```
flvndvd: Hard to understand when you use flatten attributes VS exposing them through an object. We usually try to be consistent and use them mainly for native types such as boolean and string so it's more explicit on the call sites.

front/components/assistant/conversation/actions/VisualizationActionIframe.tsx
flvndvd: Ok so this token is actually the shareable token, not the one with a 1m TTL used to fetch frame and associated assets.

davidebbo: Yes, exactly. So all good, right?

front/lib/resources/sandbox_function_resource.ts
davidebbo: Done

front/lib/resources/sandbox_function_resource.ts
davidebbo: Done.

front-api/routes/v1/public/frames/[token]/index.ts
Comment (davidebbo) [reply front-api/routes/v1/public/frames/[token]/index.ts:230]: Right now we don't. Once we add a PodApps table, we can find all functions associated with an app from teh Front DB. Not quite what you're asking for, but gets one step closer.

front/lib/resources/sandbox_function_resource.ts
davidebbo: Do you mean the `isResourceSId` checks? I guess probably not, since all callers pass server minted value. I'll remove.

front/lib/resources/sandbox_function_resource.ts
davidebbo: Main problem with `SandboxFunctionInvocationResource.baseFetch` is that it takes the function id, which is the very thing we're trying to get! Also, since we're in a resource class, it's not terrible to directly go after model IMO.

I rewrote the comment from scratch (and added on inline), no AI! In can explain in person if not clear.

front/lib/resources/sandbox_function_resource.ts
davidebbo: Didn't realize it was mostly for native types. I wonder if our skills are precise enough here to convince models to get this right.

Anyway, I fixed this one and the similar others.

front-api/routes/v1/public/frames/[token]/index.ts
Comment (davidebbo) [reply front-api/routes/v1/public/frames/[token]/index.ts:230]: Is the mint token the one that lasts one minute? If so, that's definitely problematic for calling functions on a frame that can be open for a long time. In any case, we can tweak that logic later if needed.

front-api/routes/v1/w/[wId]/sandbox/sandbox-functions/result.ts
```diff
@@ -50,9 +50,15 @@ app.post(
       return ctx.json({ success: true });
     }
 
+    // Execution-side resolution: a sandbox-token auth cannot carry the invoker's original grant
+    // (e.g. a frame share token). The id comes from signature-verified sandbox JWT claims minted
+    // at execution start, so the space filter is deliberately skipped.
     const sandboxFunction = await SandboxFunctionResource.fetchById(
```
Comment (flvndvd) [inline front-api/routes/v1/w/[wId]/sandbox/sandbox-functions/result.ts:53]: Isn't this endpoint deprecated now? See comments at the top of the file by Henry.

front-api/routes/v1/public/frames/[token]/index.ts
```diff
@@ -216,6 +218,13 @@ app.get(
       auth &&
       space.canAdministrate(auth)
     );
+    // The share token this viewer used is a workspace member's capability to invoke the frame's
+    // app's functions. For invite-only frames, a member only reaches this point with an active
+    // email grant (or as the frame's owner), which is the same rule the invocation gate applies.
+    const canInvokeViaShareCapability =
```
Comment (flvndvd) [inline front-api/routes/v1/public/frames/[token]/index.ts:224]: This does not capture the earlier `isFileOwner`. If my read is correct it means that a frame owner who is a workspace member, no longer in the pod, on an `emails_only` frame gets a `framePath` and then every `callFunction` would 404s

front/lib/api/file_system/dust_file_system.ts
```diff
@@ -322,6 +322,43 @@ export class DustFileSystem {
...
+   * Build a DustFileSystem for provisioning a pod's sandbox (mount setup and
+   * mount credential refresh). Unlike `forPod`, this does not require read
+   * access on the space: the sandbox is shared at the pod level and its mounts
+   * are identical whoever triggers the boot — in particular an invoker
+   * authorized only through app sharing, who holds no read on the Pod. Must
+   * only be called from sandbox provisioning paths, never user-facing file
+   * APIs; API-level permissions on the mount still derive from the caller, so
+   * they fail closed.
+   */
+  static async forPodSandboxProvisioning(
```
flvndvd, line 335: I'm not sure I agree on this, the `createPodMount` does apply the following which will surely yields false, if the user is not a pod member.

```
    permissions: {
      canRead: space.canRead(auth),
      canWrite: space.canWrite(auth),
    },
```

front/lib/api/sandbox_functions/create_sandbox_function_mcp_action.ts
```diff
@@ -172,9 +172,15 @@ export async function createSandboxFunctionMCPAction(
...
 
+  // Execution-side resolution: a sandbox-token auth cannot carry the invoker's original grant
+  // (e.g. a frame share token). The id comes from signature-verified sandbox JWT claims minted
+  // at execution start, so the space filter is deliberately skipped.
   const sandboxFunction = await SandboxFunctionResource.fetchById(
     auth,
-    sandboxFunctionId
+    sandboxFunctionId,
+    {
+      dangerouslyBypassSpacePermissionFilter: true,
```
flvndvd: `nit` Convention is more `dangerouslySkipSpacePermissionFiltering`

front/lib/resources/sandbox_function_resource.ts
```diff
@@ -334,20 +353,44 @@ export class SandboxFunctionResource extends BaseResource<SandboxFunctionModel>
...
     return sandboxFunctions.flatMap((sandboxFunction) => {
-      const space = accessibleSpacesById.get(sandboxFunction.get().spaceId);
-      const file = filesById.get(sandboxFunction.get().fileId);
+      const blob = sandboxFunction.get();
+      // Three prioritized gates: the caller's own access, then the app the capability grants,
+      // then the execution-side bypass (spacesById is only built for it).
+      const capabilityGrantsFunction =
+        capabilitySpace?.id === blob.spaceId &&
+        appPrefixFromSlug(blob.slug) === capability?.appPrefix;
+      const space =
```
flvndvd, line 365: 🤯

front/lib/resources/sandbox_function_resource.ts
```diff
@@ -334,20 +353,44 @@ export class SandboxFunctionResource extends BaseResource<SandboxFunctionModel>
...
 
-      return [new this(this.model, sandboxFunction.get(), space, file)];
+      return [new this(this.model, blob, space, file)];
     });
   }
 
   static async fetchById(
     auth: Authenticator,
-    sandboxFunctionId: string
+    sandboxFunctionId: string,
```
flvndvd: IMHO the previous `fetchByIdForExecution` was a better shape, it was really just missing calling the `baseFetch` with the proper flags.

flvndvd (commented):
Left a few comments, one more broader comment is that the logic to answer ""Does this token authorize this function?" is spread into three different places making it harder to resume about it.

Id' suggest making the capability a query, not a permission flag. Something like
```
SandboxFunctionResource.fetchInAppFolder(auth, { podId, appPrefix, idOrSlug })
```
One method, every param required, no optional capability threaded through multiple functions, no dangerous flag on the default fetch, and the authorization statement is the query predicate rather than a gate bolted onto a shared one.

front-api/routes/v1/w/[wId]/sandbox/sandbox-functions/result.ts
Comment (davidebbo) [reply front-api/routes/v1/w/[wId]/sandbox/sandbox-functions/result.ts:53]: It is, and I have on my list to check with Henry if we can delete it altogether.

front/lib/api/sandbox_functions/create_sandbox_function_mcp_action.ts
davidebbo: Seems Bypass is used pretty widely, e.g. `dangerouslyBypassWorkspaceIsolationSecurity` and `dangerouslyBypassSameStepCheck`. Ideally, we'd have one pattern but we have two with about equal use.

front-api/routes/v1/w/[wId]/sandbox/sandbox-functions/result.ts
Comment (davidebbo) [reply front-api/routes/v1/w/[wId]/sandbox/sandbox-functions/result.ts:53]: But I guess we can undo the change in this one, if only to reduce the diff.

Comment (davidebbo) [reply front-api/routes/v1/w/[wId]/sandbox/sandbox-functions/result.ts:53]: Gone.

front-api/routes/v1/public/frames/[token]/index.ts
Comment (davidebbo) [reply front-api/routes/v1/public/frames/[token]/index.ts:224]: Do you mean an emails_only that that they are invited to? I just tested that end to end and the frame/functions fully works. And if they're not invited, they can't get to the frame.

Comment (davidebbo) [reply front-api/routes/v1/public/frames/[token]/index.ts:224]: Oh, I see: when creating an app in a pod, the frame gets moved from the convo into the pod, at which point it loses its user. So if not invited, they get nothing.
If I had path the DB to make them the owner, then indeed they can get at the frame but not the function. Super edge case...

Comment (davidebbo) [reply front-api/routes/v1/public/frames/[token]/index.ts:224]: I fixed it, but I think it's a barely relevant edge case.

front/lib/resources/sandbox_function_resource.ts
davidebbo: "App Prefix" is really App Name. But I think this is a perfect example of what we can improve once we have new file system. This will just become the App's file ID (or folder ID).

front/lib/api/file_system/dust_file_system.ts
davidebbo: I'm trying to understand what concrete scenario this affects, if any. I think these may be irrelevant in the provisioning scenario.

front/lib/resources/sandbox_function_resource.ts
davidebbo: Alright, done...

flvndvd (approved):

PR #30533, opened 2026-08-14
Fix analytics consumption rows disappearing

flvndvd:
## Description

The attribution table has two separate issues that looked very similar:

1. Rows disappeared during background refreshes
The table treated every refresh as a first load. Even though the existing rows were still available, it replaced them with loading placeholders/skeleton until the refresh completed.

2. Newly loaded tabs stayed invisible
A layout change moved the table outside the animation setup it relied on. Rows loaded after opening an uncached tab
started hidden but never completed their fade-in. Returning to the tab worked because cached rows skipped that
animation.

## Fix

- Keep existing rows visible during background refreshes.
- Give the table its own animation setup so asynchronously loaded rows always become visible.
- Add regression tests for cached refreshes and first-time tab loads.

### DIFF
--- a/front/components/workspace/analytics/consumption/ConsumptionAttributionTable.test.tsx
+++ b/front/components/workspace/analytics/consumption/ConsumptionAttributionTable.test.tsx
@@ -1,5 +1,6 @@
 import { ConsumptionAttributionTable } from "@app/components/workspace/analytics/consumption/ConsumptionAttributionTable";
 import type { ConsumptionDimension } from "@app/components/workspace/analytics/consumption/consumptionDimensions";
+import type { ConsumptionTopRow } from "@app/hooks/useConsumptionTop";
 import {
   fireEvent,
   render,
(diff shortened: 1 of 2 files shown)


front/components/workspace/analytics/consumption/ConsumptionAttributionTable.test.tsx
```diff
@@ -1,5 +1,6 @@
 import { ConsumptionAttributionTable } from "@app/components/workspace/analytics/consumption/ConsumptionAttributionTable";
 import type { ConsumptionDimension } from "@app/components/workspace/analytics/consumption/consumptionDimensions";
+import type { ConsumptionTopRow } from "@app/hooks/useConsumptionTop";
 import {
```
flvndvd, line 4: Better reviewed with hidden whitespace: [here](https://github.com/dust-tt/dust/pull/30533/changes?w=1#diff-f8304617a56165ee7126ec2cd6d717bc8ac5bf13ec7251c1ae4ea21576e542ac).

aubin-tchoi (approved):
LGTM

PR #30535, opened 2026-08-14
fix: show private agent owner in analytics

adrsimon:
## Description

Every agent an analytics viewer can't read rendered as `Private agent: description unavailable`, so there was no way to tell who to go ask about one. Managers now see `Private agent owned by <email>`.

Two paths feed those rows: agents returned by `getAgentConfigurations` with `canRead: false`, and agents filtered out entirely by space permissions. Both resolve the author through one batched user query in `resolveAnalyticsAgentLabels`. Non-managers keep the placeholder, since `resolveAnalyticsAgentLabels` also runs on the conversation consumption path.

## Tests

`front-api` analytics consumption top tests pass. No new test: this is a description string on an existing path, and the fixtures don't set up private agents.

## Risk

Low. Read-only, one extra query when a manager's analytics range contains private agents. Revert to roll back.

## Deploy Plan

Deploy front.

### DIFF
--- a/front/lib/api/assistant/configuration/agent.ts
+++ b/front/lib/api/assistant/configuration/agent.ts
@@ -409,6 +409,7 @@ export async function getAgentConfiguration<V extends AgentFetchVariant>(
 
 type AgentLabel = {
   sId: string;
+  authorModelId: ModelId;
   name: string;
   pictureUrl: string | null;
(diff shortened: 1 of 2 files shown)


aubin-tchoi (approved):
LGTM

PR #30537, opened 2026-08-14
[front] feat: support API keys in consumption analytics

aubin-tchoi:
## Description

First part of https://github.com/dust-tt/tasks/issues/10059.

Consumption analytics cannot currently be filtered or ranked by the API key that generated usage.
This is a feature that is available on the old analytics, and is indeed nice for the heavy users of the API (since it's quite easy to overuse using the api).

This adds API key names as a consumption scope dimension, across the ES queries, the facets, filters, and top rankings.

## Tests

- Added tests, consistently with the other ones.

## Risk

- Low.

## Deploy Plan

- Deploy front.


front-api/routes/w/[wId]/analytics/consumption/top-api-keys.ts
```diff
@@ -0,0 +1,60 @@
...
+
+// Mounted at /api/w/:wId/analytics/consumption/top-api-keys.
+const app = workspaceApp();
+
+/** @ignoreswagger */
+app.post(
+  "/",
+  ensureIsManager(),
+  validate("json", ConsumptionTopBodySchema),
+  async (ctx): HandlerResult<GetConsumptionTopApiKeysResponse> => {
```
Comment (avervaet) [inline front-api/routes/w/[wId]/analytics/consumption/top-api-keys.ts:24]: This new route destructures limit, offset, filter, and the period fields, but drops search before calling fetchConsumptionTopApiKeys. So API-key ranking search currently returns unfiltered results which is inconsistent with other endpoints.

Comment (avervaet) [reply front-api/routes/w/[wId]/analytics/consumption/top-api-keys.ts:24]: We might want to change other endpoints so user can search filtered values ? Not sure tho

avervaet (approved):

front-api/routes/w/[wId]/analytics/consumption/top-api-keys.ts
Comment (aubin-tchoi) [reply front-api/routes/w/[wId]/analytics/consumption/top-api-keys.ts:24]: good catch 👍🏻

PR #30539, opened 2026-08-14
[front] fix: keep pptx slide deletes from stranding parts, surface hidden and inherited shapes

id13:
## Description

Deleting a slide could leave its part behind in the package. drop_rel keeps any rel that is still
referenced, and a custom show references the slide by the same rId, so the slide left the order
while its part stayed as an orphan and PowerPoint complains about the file. It now strips every
presentation.xml element pointing at that rId before dropping the rel, and raise if the rel
survives rather than write a deck I know is broken.

Two blind spots in inspection while I was in there: a shape with hidden="1" never renders, and
master/layout text renders on every inheriting slide but cannot be touched slide-side. Neither
showed up in --slide or --layouts, so the model kept writing into shapes nobody sees and kept
trying to fix text it does not own.

## Tests

Added tests in test_slides.py, and tested locally on real decks.

## Risk

Low. Blast radius: pptx_slides --delete and the pptx_inspect --slide / --layouts output

## Deploy Plan

Merge
Will build sandbox img when merging the last PR


id13:

- #30539 ◀
    - #30540
        - #30541
            - #30542
                - #30544
                    - #30573
                        - #30574
                            - #30576

id13:
PMRR

PR #30542, opened 2026-08-14
[front] enh: Chop pptx skill instruction tokens

id13:
## Description

Tested on a wide variety of templates, quality seems ~same. 
What went away is repetition, fluff, ai slope.

Counted with tiktoken, 10k => 2k

## Tests

Tested locally (at the top of the stack)

## Risk

Low

## Deploy Plan

- Deploy front

PR #30541, opened 2026-08-14
[front] enh: add grid render, chop help text

id13:
## Description

Add grid render, allowing models to render everything in one go.
Also cut down the help text

## Tests

Tested locally (top of the stack)

## Risk

Low

## Deploy Plan

Merge

PR #30544, opened 2026-08-14
[front] enh: add pptx package validation to --compare and --validate

id13:
## Description

--compare measures design fidelity and says nothing about whether the file opens, so a deck
PowerPoint refuses to open can pass QA clean. --validate walks the package for the faults an edit
script actually produces: a stranded part, a relationship pointing nowhere, an r:id with no
relationship, a duplicate zip entry, a reused shape id, an untyped part, malformed XML.

We were generating pptx that could not be opened according to some customer reports

## Tests

Added test_validate.py. Tested locally against decks I broke on purpose.

## Risk

Low. 
Blast radius: --compare grows a Package section and can now FAIL a deck that used to pass,
which is the point.

## Deploy Plan

Merge

### DIFF
--- a/front/lib/api/sandbox/image/profile/soffice/ooxml.py
+++ b/front/lib/api/sandbox/image/profile/soffice/ooxml.py
@@ -36,7 +36,9 @@ def resolve_rel_target(rels_path: str, target: str) -> str:
     into an absolute path within the zip."""
     if target.startswith("/"):
         return target.lstrip("/")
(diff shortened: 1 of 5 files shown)


id13:

- #30539
    - #30541
        - #30542
            - #30544 ◀
                - #30573
                    - #30574
                        - #30576

id13:
PMRR

adrsimon (commented):
LGTM

PR #30545, opened 2026-08-14
[front] fix: resolve tool dimension labels in one query

aubin-tchoi:
## Description

We were making several database queries to load tool names and icons in consumption analytics (db hydration of the es results for the attribution table).

This PR replaces them with a single query on `MCPServerViewResource` (a few queries actually, but still better because we don't fetch twice on `remote_mcp_servers`). The output stays the same.

## Tests

- Added a test.

## Risk

- Low.

## Deploy Plan

- Deploy front.

### DIFF
--- a/front/lib/api/analytics/consumption/labels.ts
+++ b/front/lib/api/analytics/consumption/labels.ts
@@ -21,7 +21,7 @@ import { asDisplayToolName } from "@app/types/shared/utils/string_utils";
  * - "user": user sIds
  * - "group": group sIds
  * - "model": raw model ids
- * - "tool": MCP server names
+ * - "tool": internal MCP server names or remote MCP server sIds
  * - "skill": skill sIds
  * - "source": origin slugs
  */
@@ -118,8 +118,10 @@ export async function resolveDimensionLabels(
       );
 
     case "tool": {
-      const metadata =
-        await MCPServerViewResource.resolveDisplayMetadataByNames(auth, keys);
+      const metadata = await MCPServerViewResource.resolveDisplayMetadata(
+        auth,
+        keys
+      );
       return new Map(
         keys.map((key) => {
           const tool = metadata.get(key);
--- a/front/lib/api/analytics/consumption/labels.ts
+++ b/front/lib/api/analytics/consumption/labels.ts
@@ -118,11 +118,10 @@ export async function resolveDimensionLabels(
       );
 
     case "tool": {
-      const metadata =
-        await MCPServerViewResource.resolveDisplayMetadataByIdentifiers(
-          auth,
-          keys
-        );
+      const metadata = await MCPServerViewResource.resolveDisplayMetadataByIds(
+        auth,
+        keys
+      );
       return new Map(
         keys.map((key) => {
           const tool = metadata.get(key);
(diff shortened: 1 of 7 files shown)


front/lib/resources/mcp_server_view_resource.ts
```diff
@@ -697,25 +704,44 @@ export class MCPServerViewResource extends ResourceWithSpace<MCPServerViewModel>
...
   }
 
-  static async resolveIconsByNames(
-    auth: Authenticator,
-    names: string[]
-  ): Promise<Map<string, CustomResourceIconType | InternalAllowedIconType>> {
-    const uniqueNames = [...new Set(names)];
-    if (uniqueNames.length === 0) {
-      return new Map();
+  static async resolveDisplayMetadata(auth: Authenticator, keys: string[]) {
```
aubin-tchoi, line 707: kind of ugly because we leak the concept of the key that is owned by how we write to ES in the resource, but easier this way

front/lib/api/assistant/observability/tool_usage.ts
```diff
@@ -231,14 +231,17 @@ export async function resolveServerDisplayNames(
   serverNames: string[]
 ): Promise<Map<string, string>> {
   const unique = [...new Set(serverNames)];
-  const remoteNameMap = await RemoteMCPServerResource.resolveNamesBySIds(
-    auth,
-    unique
+  const remoteServers = await RemoteMCPServerResource.fetchByIds(auth, unique);
```
avervaet, line 234: nit: I think unique includes internal names, so fetchByIds logs invalid-sId errors for normal keys

front/lib/api/analytics/consumption/labels.ts
```diff
@@ -123,23 +118,21 @@ export async function resolveDimensionLabels(
       );
 
     case "tool": {
-      const [remoteMetadata, viewIcons] = await Promise.all([
-        RemoteMCPServerResource.resolveDisplayMetadataByIdentifiers(auth, keys),
-        MCPServerViewResource.resolveIconsByNames(auth, keys),
-      ]);
+      // Internal servers are keyed by name, while remote servers are keyed by sId.
+      const metadata = await MCPServerViewResource.resolveDisplayMetadata(
```
avervaet, line 122: If i'm correct, production tool keys are still remote names (cachedName/view names), so this sId-only lookup falls back and loses remote labels/icons. 

We could resolve names too or migrate the whole contract atomically.

avervaet (approved):

front/lib/api/assistant/observability/tool_usage.ts
aubin-tchoi: it's names for the internal ones, ids for the external ones no?

aubin-tchoi: wait this bit of code makes no sense now

aubin-tchoi: ah well it already made no sense and we don't use it in our case so 🤷🏻

front/lib/api/analytics/consumption/labels.ts
aubin-tchoi: in `resolveDisplayMetadata` I map the ones that are names of internal MCP servers directly (no db fetch) and fetch the ones that are remote mcp server ids, am I missing some by doing this?

PR #30546, opened 2026-08-14
Fix enabled skill tool footprint attribution

flvndvd:
## Description

Carrying on the journey to reconcile credits billed with tool attribution.

This PR focuses around skill tool attribution estimation that wrongly assumed that all tools would be exposed to the model directly on the subsequent calls. Since we introduced `tool_search` this is not conditioned by provider. This PR aims at aligning this. I've bumped the attribution schema so I can restart all stuck workflows.


fontanierh (approved):

aubin-tchoi (approved):
LGTM

PR #30547, opened 2026-08-14
Prevent file grep from stalling agent workers

fontanierh:
## Description

- Run every model-facing file grep path with RE2 so user or model-provided patterns cannot trigger catastrophic backtracking.
- Enforce a 4,096-character pattern limit and a 1 MiB per-line limit while consuming stream chunks, before a newline-free file can be buffered.
- Bound accumulated grep output to the existing 20 KiB text envelope, including the files server paths that bypass downstream offloading.

## Tests

- Added common and case-insensitive regex coverage.
- Added unsupported and oversized pattern coverage.
- Added an adversarial backtracking regression with a wall-clock bound.
- Added pre-newline oversized-line and UTF-8-safe response-budget coverage.
- Ran the focused Vitest suite, Front typecheck, and production worker build locally.

## Risk

RE2 intentionally rejects lookarounds and backreferences that JavaScript accepted. Patterns above 4,096 characters and lines above 1 MiB now return explicit tool errors. Results above the response byte budget are truncated with a marker. Rollback is a standard revert.

## Deploy Plan

Deploy the Front worker image normally to both regions. Verify file grep remains responsive for common and adversarial patterns, and monitor agent-loop activity heartbeat failures and worker memory after rollout.


front/lib/api/actions/servers/files/tools/grep.ts
```diff
@@ -85,50 +87,33 @@ export async function grepHandler(
...
-        }
-      }
-    }
-  } catch (err) {
+  if (grepResult.isErr()) {
     return new Err(
       new MCPError(
-        `Failed to read file \`${path}\`: ${normalizeError(err).message}`
+        `Failed to search file \`${path}\`: ${grepResult.error.message}`,
+        { tracked: !isGrepLineTooLongError(grepResult.error) }
```
flvndvd: Why not relying on the instance of the error class?

front/lib/api/actions/servers/files/tools/grep_regex.ts
```diff
@@ -0,0 +1,188 @@
...
+      const newline = chunk.indexOf(NEWLINE_BYTE, start);
+      const end = newline === -1 ? chunk.length : newline;
+      const part = chunk.subarray(start, end);
+
+      if (pendingBytes + part.length > GREP_LINE_MAX_BYTES) {
+        yield new Err(new GrepLineTooLongError(lineNumber));
+        return;
+      }
+      if (part.length > 0) {
+        pending.push(part);
```
flvndvd: This keeps one `Buffer` object for every chunk until the next newline. The 1 MB limit does not limit memory when the stream sends very small chunks. Pretty sure that if we tests for a 1 MB line arrived one byte at a time, this will use hundred of MB. Could we copy the bytes into one growing buffer capped at 1 MB instead?

flvndvd (approved):
LGTM, just one concern around growing memory pressure

flvndvd (approved):

PR #30548, opened 2026-08-14
Retry interruptible sandbox child tools

fontanierh:
## Description

Sandbox child workflows always used the single-attempt tool activity, even when the stored tool policy was `retry_on_interrupt`. This caused 14 child workflow failures in 24 hours and left 13 actions stuck in `running`.

Pass the stored retry policy into the child workflow. Missing policies default to `no_retry` without recording a patch marker, which preserves old histories and lets a launcher-first rollback stop new marked workflows. Only `retry_on_interrupt` tools use the existing retryable activity proxy.

On terminal activity failure, a dedicated finalizer atomically moves only sandbox child actions still in `running` to `errored` before the workflow rethrows the original failure.

## Tests

Added focused workflow tests for both policies, missing legacy input without a patch marker, mixed-version replay, and terminal activity failures. Added activity tests for the sandbox-child guard and compare-and-set status transition. Ran these tests and the existing tool-interruption policy tests locally.

Full local typecheck was blocked by stale linked Notion and Novu dependencies in the fresh Hive environment. CI runs with a clean dependency install.

## Risk

Retrying may repeat side effects only for tools already explicitly configured as `retry_on_interrupt`, matching the main agent-loop dispatch. `no_retry` tools remain single-attempt. Failure finalization cannot overwrite actions that already reached a terminal or blocked status.

A forward rolling deploy is replay-safe because the patch marker preserves histories first executed by an older worker. A full rollback to pre-change workers is not safe while marked child workflows remain open: stop creating new marked workflows, keep patch-aware workers running until those executions close, then finish the worker rollback.

## Deploy Plan

Standard forward rolling deploy. Monitor sandbox child workflow failures and actions remaining in `running` after closed workflows. If rollback is needed, revert the launcher first so new workflows omit the policy and remain unmarked, drain marked child workflows on patch-aware workers, then revert the workers.

### DIFF
--- a/front/temporal/agent_loop/workflows.ts
+++ b/front/temporal/agent_loop/workflows.ts
@@ -578,7 +578,7 @@ export async function runSandboxChildToolWorkflow({
   authType,
   agentLoopArgs,
   actionModelId,
-  retryPolicy = "no_retry",
+  retryPolicy,
   step,
 }: {
   authType: AuthenticatorType;
@@ -594,9 +594,11 @@ export async function runSandboxChildToolWorkflow({
     runAgentArgs: agentLoopArgs,
     step,
   };
-  const useRetryPolicy = patched("sandbox-child-tool-retry-policy");
+  const effectiveRetryPolicy = retryPolicy ?? "no_retry";
+  const useRetryPolicy =
+    retryPolicy !== undefined && patched("sandbox-child-tool-retry-policy");
   let shouldRetryOnInterrupt: boolean;
-  switch (retryPolicy) {
+  switch (effectiveRetryPolicy) {
     case "retry_on_interrupt":
       shouldRetryOnInterrupt = useRetryPolicy;
       break;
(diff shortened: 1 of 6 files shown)


front/temporal/agent_loop/workflows.ts
```diff
@@ -569,18 +578,53 @@ export async function runSandboxChildToolWorkflow({
...
-  const { deferredEvents } = await runToolActivity(authType, {
+  const activityArgs = {
     actionId: actionModelId,
     runAgentArgs: agentLoopArgs,
     step,
-  });
+  };
+  const effectiveRetryPolicy = retryPolicy ?? "no_retry";
+  const useRetryPolicy =
+    retryPolicy !== undefined && patched("sandbox-child-tool-retry-policy");
```
aubin-tchoi: will be painful to put the deprecate patch, will be easier with a very wide if patched else

aubin-tchoi (approved):
LGTM

front/temporal/agent_loop/workflows.ts
fontanierh: fixed

PR #30551, opened 2026-08-14
Add Temporal job to generate raw consumption exports and store them in GCS

avervaet:
## Description

This is the base PR for the consumption-export stack:

- [[#30557](https://github.com/dust-tt/dust/pull/30557)](https://github.com/dust-tt/dust/pull/30557): replaces the synchronous raw-data download with a background-job panel that lists GCS exports and starts the Temporal job.
- [[#30558](https://github.com/dust-tt/dust/pull/30558)](https://github.com/dust-tt/dust/pull/30558): notifies the requesting user when an export is ready.

This PR moves raw consumption-lines CSV/ZIP generation out of the inline API request and into the a Temporal workflow on the `analytics_queue`. The activity reuses the existing Elasticsearch-backed export builder and uploads the resulting ZIP to the private GCS bucket at `w/{workspaceId}/consumption_exports/{exportId}.zip`.

The launcher derives `exportId` from a canonicalized SHA-256 hash of the requested period and scope filter. It first serves an existing GCS object when available, otherwise starts a workflow with a deterministic ID for that workspace/export combination. Requests for the same export are deduplicated, including the race between the running-workflow check and `start`; differently scoped exports receive separate workflow IDs. Open-ended periods are capped at the end of the current UTC day so they can be reused within the day and refreshed as new data accrues on subsequent days.

The launcher and workflow are wired in this PR, while the production caller is introduced by [[#30557](https://github.com/dust-tt/dust/pull/30557)](https://github.com/dust-tt/dust/pull/30557). Existing synchronous export behavior remains unchanged in this base PR.

## Tests

Added Vitest coverage for the export activity's success and failure paths. The success case verifies that a ZIP containing `lines.csv` is uploaded with expected LLM and tool rows, resolved labels, and workspace prefix; the failure case verifies that Elasticsearch errors are propagated and no file is uploaded.

Added launcher coverage for starting new exports, reusing completed GCS exports, refreshing open-ended periods across UTC days, allowing differently scoped exports to run independently, and deduplicating both already-running workflows and concurrent start races. Manual testing was also performed.

## Risk

Low for this PR in isolation. The new workflow and storage path do not change existing callers, and the dependent UI and notification behavior is implemented in later stack PRs. Once invoked, export failures are propagated before upload, cache keys are scoped to the period and filter, and deterministic workflow IDs prevent duplicate work for the same export while allowing different exports to run independently. 

## Deploy Plan

- Deploy front

### DIFF
--- a/front/temporal/analytics_queue/activities/consumption_export.ts
+++ b/front/temporal/analytics_queue/activities/consumption_export.ts
@@ -35,24 +35,18 @@ function canonicalizeConsumptionScopeFilter(
 }
 
 // Hashed (rather than kept as plain text) because a filter selecting many agents/users/tools
-// can get long, and this value is used as a GCS object name. `salt`, when passed, forces a
-// unique key regardless of period/filter — used for open-ended periods (e.g. "this cycle")
-// whose data keeps changing while the period itself stays the same, so every trigger must
-// produce its own export rather than reusing a previous, now-stale one.
+// can get long, and this value is used as a GCS object name.
 export function buildConsumptionExportCacheKey({
   period,
   filter,
-  salt,
 }: {
   period: ConsumptionPeriod;
   filter: ConsumptionScopeFilter;
-  salt?: string;
 }): string {
   const payload = JSON.stringify({
     startDate: period.startDate,
     endDate: period.endDate,
     filter: canonicalizeConsumptionScopeFilter(filter),
-    salt: salt ?? null,
   });
   return createHash("sha256").update(payload).digest("hex");
 }
--- a/front/temporal/analytics_queue/activities/consumption_export.ts
+++ b/front/temporal/analytics_queue/activities/consumption_export.ts
@@ -30,7 +30,7 @@ function canonicalizeConsumptionScopeFilter(
       const [, values] = entry;
       return values !== undefined && values.length > 0;
     })
-    .map(([key, values]) => [key, [...values].sort()] as const)
+    .map(([key, values]) => [key, [...values].sort()])
     .sort(([a], [b]) => a.localeCompare(b));
 }
 
--- a/front/temporal/analytics_queue/activities/consumption_export.ts
+++ b/front/temporal/analytics_queue/activities/consumption_export.ts
@@ -1,0 +1,52 @@
+import { fetchConsumptionLinesExportZip } from "@app/lib/api/analytics/consumption/export_lines";
+import type { ConsumptionPeriod } from "@app/lib/api/analytics/consumption/period";
+import type { ConsumptionScopeFilter } from "@app/lib/api/analytics/consumption/scope";
+import type { AuthenticatorType } from "@app/lib/auth";
+import { Authenticator } from "@app/lib/auth";
+import { getPrivateUploadBucket } from "@app/lib/file_storage";
+import logger from "@app/logger/logger";
+
+// Files are namespaced by workspace under a single top-level prefix so the cleanup job can
+// scan every workspace's exports in one bucket listing.
+export function buildConsumptionExportGcsPrefix(workspaceSId: string): string {
+  return `consumption_exports/w/${workspaceSId}/`;
+}
+
+function buildConsumptionExportGcsPath(
+  workspaceSId: string,
+  timestamp: number
+): string {
+  return `${buildConsumptionExportGcsPrefix(workspaceSId)}${timestamp}.zip`;
+}
+
+export async function runConsumptionExportActivity(
+  authType: AuthenticatorType,
+  {
+    period,
+    filter,
+  }: {
+    period: ConsumptionPeriod;
+    filter: ConsumptionScopeFilter;
+  }
+): Promise<void> {
+  const auth = await Authenticator.fromJSON(authType);
+  const workspaceSId = auth.getNonNullableWorkspace().sId;
+
+  const result = await fetchConsumptionLinesExportZip(auth, { period, filter });
+  if (result.isErr()) {
+    logger.error(
+      { workspaceId: workspaceSId, err: result.error },
+      "[ConsumptionExport] Failed to build consumption lines export."
+    );
+    throw result.error;
+  }
+
+  // Generated inside the activity (not the workflow, which must stay deterministic). Safe as
+  // a plain timestamp because the deterministic per-workspace workflow ID guarantees at most
+  // one export activity runs per workspace at a time.
+  const gcsPath = buildConsumptionExportGcsPath(workspaceSId, Date.now());
+
+  await getPrivateUploadBucket()
+    .file(gcsPath)
+    .save(result.value, { contentType: "application/zip", resumable: false });
+}
--- a/front/temporal/analytics_queue/activities/consumption_export.ts
+++ b/front/temporal/analytics_queue/activities/consumption_export.ts
@@ -10,14 +10,12 @@ import { getPrivateUploadBucket } from "@app/lib/file_storage";
 import logger from "@app/logger/logger";
 import { createHash } from "crypto";
 
-// Files live under the workspace's own prefix (matching the `w/{wId}/...` layout used elsewhere
-// in this bucket) so relocation and scrubbing can operate on the workspace prefix directly.
 export function buildConsumptionExportGcsPrefix(workspaceId: string): string {
   return `w/${workspaceId}/consumption_exports/`;
 }
 
-// exportId is a content hash (see buildConsumptionExportCacheKey) identifying the export's
-// GCS object, so a retry within the same triggered export (same args, same hash) overwrites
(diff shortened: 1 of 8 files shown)


avervaet:

- #30551 ◀
    - #30557
        - #30558

front/temporal/analytics_queue/workflows.ts
```diff
@@ -26,6 +28,12 @@ const {
   startToCloseTimeout: "5 minutes",
 });
 
+// A raw consumption export can page through a large number of Elasticsearch documents, so it
+// gets a much longer ceiling than the other analytics activities on this queue.
+const { runConsumptionExportActivity } = proxyActivities<typeof activities>({
```
avervaet, line 31: Probably too aggressive, to refine

front/temporal/analytics_queue/client.ts
```diff
@@ -158,6 +164,54 @@ export async function launchStoreAgentMessageConsumptionAttributionWorkflow({
...
+    await client.workflow.start(runConsumptionExportWorkflow, {
+      args: [authType, { period, filter }],
+      taskQueue: QUEUE_NAME,
+      workflowId,
+      memo: {
+        workspaceId,
+      },
+    });
+    return new Ok(undefined);
+  } catch (e) {
```
avervaet, line 269: Wasn't sure if we wanted to log here. kept it simple for now

avervaet:
Concern; Right now this queue have a concurrency of 24 so if multiple large clients triggers at the same moment we can have a lot of memory pressure.

front/temporal/analytics_queue/workflows.ts
```diff
@@ -26,6 +32,21 @@ const {
   startToCloseTimeout: "5 minutes",
 });
 
+// scheduleToCloseTimeout bounds the total time across all retries: a persistent
+// Elasticsearch/GCS failure must eventually fail the workflow rather than retry
+// forever, since the workflow ID is stable per workspace and blocks subsequent
+// export requests while running. 3 attempts at up to 30 minutes each, plus
+// backoff, comfortably fits inside the 3-hour ceiling.
+const { runConsumptionExportActivity } = proxyActivities<typeof activities>({
+  startToCloseTimeout: "30 minutes",
```
flvndvd: Remember what we said about deploy, we have a graceful period of 2 minutes, maybe we can start with 5 or 10 minutes, WDYT? I'd rather start conservative and eventually increase.

Also any reason for a custom retry?

front/temporal/analytics_queue/workflows.ts
```diff
@@ -75,3 +96,28 @@ export async function storeAgentMessageConsumptionAttributionV3Workflow(
...
+  authType: AuthenticatorType,
+  {
+    period,
+    filter,
+  }: {
+    period: ConsumptionPeriod;
+    filter: ConsumptionScopeFilter;
+  }
+): Promise<void> {
+  // runId (not workflowId) so each trigger produces its own GCS object: the
```
flvndvd: Should we encode the period and filter in the workspace Id, so we only run it once, and we always end up with the same file path in GCS?

front/temporal/analytics_queue/helpers.ts
```diff
@@ -9,3 +9,12 @@ export function makeAgentMessageAnalyticsWorkflowId({
...
   return `agent-message-analytics-${workspaceId}-${conversationId}-${agentMessageId}`;
 }
+
+// Deterministic, one per workspace
+export function makeConsumptionExportWorkflowId({
+  workspaceId,
+}: {
+  workspaceId: string;
+}): string {
+  return `consumption-export-${workspaceId}`;
```
flvndvd: Why not more than one?

front/temporal/analytics_queue/client.ts
```diff
@@ -158,6 +170,112 @@ export async function launchStoreAgentMessageConsumptionAttributionWorkflow({
...
+async function describeRunningConsumptionExport(
+  handle: WorkflowHandle
+): Promise<RunningConsumptionExport | undefined> {
+  try {
+    const execution = await handle.describe();
+    if (execution.status.name !== "RUNNING") {
+      return undefined;
+    }
+
+    const memo = ConsumptionExportMemoSchema.safeParse(execution.memo);
```
flvndvd: Just double checking, but IIRC there is a hard limit on the memo that Temporal can support, might be worth checking. We can probably use the workflow body for this one case.

front/temporal/analytics_queue/activities/consumption_export.ts
```diff
@@ -0,0 +1,55 @@
...
+import type { ConsumptionScopeFilter } from "@app/lib/api/analytics/consumption/scope";
+import type { AuthenticatorType } from "@app/lib/auth";
+import { Authenticator } from "@app/lib/auth";
+import { getPrivateUploadBucket } from "@app/lib/file_storage";
+import logger from "@app/logger/logger";
+
+// Files are namespaced by workspace under a single top-level prefix so the cleanup job can
+// scan every workspace's exports in one bucket listing.
+export function buildConsumptionExportGcsPrefix(workspaceId: string): string {
+  return `consumption_exports/w/${workspaceId}/`;
```
flvndvd: This will break relocation, IIRC we have a helper to get the root path for a workspace.

flvndvd: It needs to follow the `w/[wId]/` baseline.

front/temporal/analytics_queue/workflows.ts
avervaet: Love the idea. My only pushback would be how to handle the current cycle, as this means we won’t rerun an export until the next one.

avervaet: Love the idea my only push back would be how to handle the current cycle cas as this means we won't rerun an export until the next one.

avervaet: Going for a middle ground. For calls from non-current cycles, we encode the filters, start date, and end date. For calls from the current cycle, we encode the moment when we started the workflow.

front/temporal/analytics_queue/activities/consumption_export.ts
```diff
@@ -0,0 +1,89 @@
...
+// Sorted independently of request key/array order so equivalent filters hash identically.
+function canonicalizeConsumptionScopeFilter(
+  filter: ConsumptionScopeFilter
+): ReadonlyArray<readonly [ConsumptionScopeFilterKey, string[]]> {
+  return Object.entries(filter)
+    .filter((entry): entry is [ConsumptionScopeFilterKey, string[]] => {
+      const [, values] = entry;
+      return values !== undefined && values.length > 0;
+    })
+    .map(([key, values]) => [key, [...values].sort()] as const)
```
flvndvd:
```suggestion
    .map(([key, values]) => [key, [...values].sort()])
```

front/temporal/analytics_queue/activities/consumption_export.ts
```diff
@@ -0,0 +1,89 @@
...
+    .sort(([a], [b]) => a.localeCompare(b));
+}
+
+// Hashed (rather than kept as plain text) because a filter selecting many agents/users/tools
+// can get long, and this value is used as a GCS object name. `salt`, when passed, forces a
+// unique key regardless of period/filter — used for open-ended periods (e.g. "this cycle")
+// whose data keeps changing while the period itself stays the same, so every trigger must
+// produce its own export rather than reusing a previous, now-stale one.
+export function buildConsumptionExportCacheKey({
```
flvndvd, line 42: This function is a bit too shallow, what about passing cycle dates, so we can simply always resolve to the right dates?

front/temporal/analytics_queue/client.ts
```diff
@@ -158,6 +173,107 @@ export async function launchStoreAgentMessageConsumptionAttributionWorkflow({
...
+    period,
+    filter,
+    salt: isOpenEndedPeriod(period) ? currentUtcDayBucket() : undefined,
+  });
+  const gcsPath = buildConsumptionExportGcsPath(workspaceId, exportId);
+
+  const [cached] = await getPrivateUploadBucket().file(gcsPath).exists();
+  if (cached) {
+    return new Ok({ status: "cached", gcsPath });
+  }
```
flvndvd, line 243: Can we keep this function solely focused around starting the workflow. Either this logic lives above or it lives inside the workflow which first start by checking if one export already exist?

flvndvd (approved):

front/temporal/analytics_queue/activities/consumption_export.ts
avervaet: Yes but as a cycle resolves in the future too runs during the same cycle can have really different outcomes. Instead i resolved current cycle end date to today.

PR #30553, opened 2026-08-14
[front] fix: show skeleton during pagination in attribution table

aubin-tchoi:
## Description

In the new analytics we have a table for the attribution.
Changing pages on this table currently leaves cached rows visible while the next page loads, which makes pagination feel unresponsive.

This PR displays a skeleton during the page change. Also updates the skeleton to use the destination page row count so that full and partial pages reserve exactly the final table height.

## Tests

- Added a test (could be overkill).

## Risk

- Low.

## Deploy Plan

- Deploy front.


ClementAupiais:
👍

PR #30554, opened 2026-08-14
LLMs: add Gemini 3.7 Flash

fabiencelier:
## Description

 - Adds `gemini-3.7-flash` (model config, `model_constructors`, `llms`), registered on AI Studio global + agent-platform
 - global. 3.6 Flash → `isLatest: false`.
 - Specs/pricing verified against Google docs (2026-08-14)
 - pricing same as 3.6 Flash.
 - Unlike the other Flash models it **rejects `MINIMAL`**, so its reasoning contract is low/medium/high (same as Pro), default `medium`. `none` works live but is undocumented, so it's excluded from the schema.
 - EU agent-platform endpoint defined but not registered: Gemini 404s at `locations/eu` (checked 3.6 too).
 - Gemini 3.7 flash replaces the 3.6 flash in the auto "fast" models 

## Tests

 - Live suites green on both registered endpoints (92/92). 
 - tested locally

## Risk

low

## Deploy Plan

deploy front


fontanierh (approved):
LGTM but I think we should align on the discount pricing for now

PR #30556, opened 2026-08-14
Add scheduled Temporal job to delete consumption exports after 15 days

avervaet:
## Description

Adds a daily Temporal schedule on the `analytics_queue` task queue to clean up 15 days or older consumption export archives from the shared `consumption_exports/` GCS prefix.

The schedule runs daily at 03:00 UTC.

## Tests

Added unit coverage for deleting only stale exports, retaining fresh and timestamp-less exports, and handling an empty export prefix.

## Risk

Low as GCS export are introduced by this stack of PR. Thus, this introduces an automated, irreversible deletion path for objects older than the 15-day retention window. T

The schedule is provisioned out of band through the admin CLI, so an already-created Temporal schedule may need to be disabled or removed separately if the code is rolled back.

## Deploy Plan

From `front/`, run the schedule bootstrap once per environment:

````sh
temporal/analytics_queue/admin/cli.sh start-consumption-export-cleanup

### DIFF
--- a/front/temporal/analytics_queue/activities/consumption_export.test.ts
+++ b/front/temporal/analytics_queue/activities/consumption_export.test.ts
@@ -1,0 +1,225 @@
+// @vitest-environment node: adm-zip requires Node builtins (Buffer, zlib).
+// This directive makes them available in the test environment.
+
+import { resolveDimensionLabels } from "@app/lib/api/analytics/consumption/labels";
+import {
+  ElasticsearchError,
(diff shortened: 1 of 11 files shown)


avervaet:

- #30551
    - #30556 ◀
        - #30557
            - #30558

avervaet:
Dropping this PR — the scheduled 15-day export retention/cleanup job is no longer necessary. Rebased panel (#30557) and notify (#30558) directly onto job (#30551).

PR #30557, opened 2026-08-14
Replace synchronous raw consumption export with a background-job panel

avervaet:
## Description

Replace the synchronous raw consumption export with an asynchronous export panel backed by the Temporal workflow introduced in https://github.com/dust-tt/dust/pull/30551.  The `Download raw data` action now opens a popover that lists workspace exports from the private upload bucket, sorted newest first, with five-minute signed download URLs. 

**Visuals:**
Capture d’écran 2026-08-14 à 16 06 40

Capture d’écran 2026-08-14 à 16 07 49

The API now provides a manager-only `GET` endpoint returning available exports and whether the workspace export workflow is running. The manager-only `POST` endpoint resolves the requested period and filter, starts the workflow, and immediately returns `202 { isGenerating: true }` instead of generating and returning the ZIP in the request. 

The frontend polls while a job is running and shows generation state in the panel. 

## Tests

Automated route coverage was updated for empty export lists, export listing and ordering, signed download URLs, running workflow state, successful asynchronous job creation, authorization, and invalid request bodies

## Risk

Low. The main compatibility change is the `POST` response: it no longer returns a ZIP attachment synchronously and instead starts a background job, so direct API consumers that depend on the old response contract would need updating. Export generation and downloads now depend on the Temporal workflow, private bucket listing, and signed URL generation; failures in those services can affect export availability. 

## Deploy Plan

-Deploy front

PR #30558, opened 2026-08-14
Notify the requesting user when their consumption export finishes

avervaet:
## Description

Adds an in-app Novu notification for consumption export completion. The new `consumption-export-ready` workflow is registered with the Novu route and presents a “Download” action linking back to the workspace’s consumption analytics page.

The export activity triggers the notification only after the ZIP has been uploaded successfully, targeting the requesting user. The transaction ID includes the workspace, user, and export ID, so retries of the same export reuse the same Novu transaction and can be deduplicated. Notification delivery is fire-and-forget: Novu errors are logged without blocking or failing an otherwise successful export.

The change also adds coverage for the notification payload, transaction ID stability, Novu error handling, and the activity’s success/failure notification behavior.

## Tests

Vitest coverage was added/updated for the new workflow helper and the consumption export activity. N

## Risk

Low - The behavioral scope is limited to notifying users after a consumption export has been successfully uploaded. If Novu is unavailable or reports an error, the export itself still completes and the error is logged. Failed exports do not trigger a completion notification. Stable transaction IDs reduce the risk of duplicate in-app notifications when the export activity is retried.

## Deploy Plan

- Deploy front


avervaet:

- #30551
    - #30557
        - #30558 ◀

aubin-tchoi (approved):
LGTM

PR #30560, opened 2026-08-14
[front] Pod app portability: export/import apps as zip archives

davidebbo:
Export Action on an app:
Screenshot 2026-08-14 at 16 22 15

Import button in the app tab brings this up:
Screenshot 2026-08-14 at 16 23 12

## Summary

Makes Pod apps portable: an app can be downloaded from a pod as a self-contained zip archive and uploaded into any other pod — including a pod in a different workspace, region, or instance. The archive contains no source identity, so import behaves identically regardless of where it came from.

## Archive format

A plain zip (`<prefix>.podapp.zip`): the app folder's files byte-for-byte under `files/` (without the folder name itself — the name, and thus the prefix, is chosen at import time, like clone), plus a `manifest.json` carrying what the files can't: `formatVersion` (evolution slot), app name, per-file content types, per-function publish metadata (`description`, `executionMode`), per-Frame publish state and pinned-tab metadata, and database names. No sIds, no bundles (always rebuilt at publish, which also re-derives schemas), no database data, no share tokens, no env var values.

## Export

`GET /api/w/:wId/pods/:podId/apps/:prefix/export` (`requireCanRead`). Reads only GCS and Postgres — exporting never wakes a sleeping pod. Frames are exported from their original source (not the built bundle). Rejects colliding folder names and apps whose summed file size exceeds the archive cap.

## Import

`POST /api/w/:wId/pods/:podId/apps/import` (`requireCanWrite`, multipart `file` + optional `name`, `bodyLimit`-gated). Mirrors `clonePodApp`'s pipeline with the archive as the source: validate fully before writing anything (size/entry-count/uncompressed caps, zip-slip rejection, manifest schema) → prefix collision check → write files → recreate Frames via `createPodFrameFile` → re-publish functions from their new paths (which re-derives slugs and schemas under the new prefix) → reconcile databases empty from the copied schema files → publish Frames that were published at export (surfacing `validateFramePodFunctionReferences` errors, e.g. hard-coded cross-pod refs, as warnings) → re-pin tabs under the `MAX_POD_FRAME_TABS` cap.

Per-item failures land in the returned summary's `warnings`/`skipped` instead of aborting (a partially imported app is visible in the Apps tab and fixable); only sandbox unavailability aborts (503). Missing env vars/secrets/egress domains in the target workspace surface at runtime, as designed — a `requirements` manifest section is a possible follow-up.

## UI

Download action on each app tile (any reader), and an Import dialog in the Apps tab (editors, reachable from the empty state too) with optional rename, prefix-collision validation, busy state, and a report view listing warnings/skipped items before the dialog closes.

## Testing

37 endpoint tests across the apps routes (5 files), including: real-AdmZip export assertions (manifest contents, file bytes, published+pinned Frame capture), an export→import round-trip into a second pod, zip-slip and format-version rejection, oversized-body 413 and oversized-app 400, name collision 409, sandbox-unavailable 503, skipped/warning reporting, and tab-pin cap behavior.

### DIFF
--- a/front-api/routes/w/[wId]/pods/[podId]/apps/export.test.ts
+++ b/front-api/routes/w/[wId]/pods/[podId]/apps/export.test.ts
@@ -1,0 +1,181 @@
+// @vitest-environment node: adm-zip requires Node builtins (Buffer, zlib).
+// This directive makes them available in the test environment.
+
+import type { Authenticator } from "@app/lib/auth";
+import { SandboxFunctionResource } from "@app/lib/resources/sandbox_function_resource";
+import type { SpaceResource } from "@app/lib/resources/space_resource";
+import { FileFactory } from "@app/tests/utils/FileFactory";
+import { createPrivateApiMockRequest } from "@app/tests/utils/generic_private_api_tests";
+import { fileStorageMock } from "@app/tests/utils/mocks/file_storage";
+import { SpaceFactory } from "@app/tests/utils/SpaceFactory";
+import type { PodAppManifest } from "@app/types/api/pod_app_archive";
+import { PodAppManifestSchema } from "@app/types/api/pod_app_archive";
+import { sandboxFunctionContentType } from "@app/types/files";
+import { honoApp } from "@front-api/app";
+import AdmZip from "adm-zip";
+import { beforeEach, describe, expect, it, vi } from "vitest";
+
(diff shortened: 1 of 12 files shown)


fontanierh (commented):
LGTM

PR #30561, opened 2026-08-14
feat(models): unify model tier vocabulary and gate streams on their tier

Nils-Fedrigo:
## Description

Three different vocabularies described the same three model buckets: the pricing tiers
(`cost_efficient` / `balanced` / `premium`, shown as "Cost Efficient" / "Balanced" / "Premium"),
the model picker's streams (Fast / Standard / Complex), and the usage page's complexity filter
(Fast / Standard / Complex). This PR settles on one — **Basic / Standard / Premium** — and makes a
stream mean the tier it is named after.

It also displays the name of the model the tier will resolve to instead of description.

Screenshot 2026-08-14 at 14 10 02

**Vocabulary**

- `getModelsTierDisplayName()` in `token_pricing/tiers.ts` is now the single source of truth for
  tier labels, replacing `formatAsDisplayName(tierName)` at its three call sites ("Up to X"
  summary, "Includes …" description, tier explainer cards).
- The picker's streams are renamed Fast → Basic and Complex → Premium, and each tier row now shows
  the model it currently resolves to instead of a static blurb (the `/api/w/[wId]/models` payload
  carries a new `streams` field for this).
- The usage page's `UsageModelTier` was a 1:1 restatement of `ModelsTierName`, so it is deleted
  along with `USAGE_MODEL_TIER_LABEL` and `usageModelTierFromModelsTierName`; the filter panel uses
  `ModelsTierName` directly. A hint above the buttons spells out that these group the concrete
  models actually billed, not the picker's identically-named streams.

**Stream / tier alignment**

The stream pools contained candidates from other tiers, so a stream could resolve to a model priced
above what its name promised:

- `gemini-3.6-flash` at `light` moves from `balanced` to `cost_efficient` (Basic pool).
- `grok-4.6` at `medium` moves from `premium` to `balanced`, and the Standard pool now enters it at
  `medium` rather than `high`.
- `gpt-5.5` at `medium` (premium) is dropped from the Standard pool entirely.

**Gating**

The three meta-models were all tiered `balanced`, which meant a Basic-capped member was rejected on
*every* stream — the picker offered them while `validatePublicModelSelection` and
`getModelTierAccessErrorForAgentConfiguration` refused them. Each stream is now tiered as the tier
it is named after (`auto_fast` → cost_efficient, `auto` → balanced, `auto_complex` → premium), and
the UI follows: `isTierLocked` becomes `getTierLockReason`, returning the existing `ModelLockReason`
union so a tier row locked by the member's cap gets the tier tooltip rather than the legacy-plan
one. The `/` slash menu drops those tiers the same way it already dropped premium-gated ones.
Members capped below Standard now default to the Basic stream rather than falling through to a
concrete model.

## Tests

Covered by unit/integration tests (63 passing across the touched suites):

- `getTierLockReason` — locked by cap, locked by legacy plan, unlocked when the stream is absent
  from the payload (the picker renders before `useModels` resolves, so absence must not lock).
- `buildPickModelSlashCommandItems` — a tier above the member's cap is omitted.
- `enabled_models` — per-stream selectability under a `balanced` cap, and the Basic-capped default.
- `tiers` — the `grok-4.6` reclassification.

Worth a manual pass in the picker with a tier-capped member: the tier rows they cannot use should
be locked with the "model tier" tooltip, and their default should land on Basic.

## Risk

An agent **configured** on `auto_complex` now fails for non-premium members with "Model tier not
enabled", exactly as an agent configured on Opus does — unless it is published and the workspace
allows restricted models for published agents. That is the intended consequence of gating streams
on their tier, but it is the one behaviour change that reaches beyond the picker.

The `grok-4.6@medium` and `gemini-3.6-flash@light` reclassifications also widen who can select
those combinations directly, since both moved to a cheaper tier.

## Deploy Plan

Nothing specific — no migration, no config change.

### DIFF
--- a/front/components/editor/extensions/shared/slash_suggestion/PickModelSubMenuDropdown.tsx
+++ b/front/components/editor/extensions/shared/slash_suggestion/PickModelSubMenuDropdown.tsx
@@ -48,7 +48,7 @@ export const PickModelSubMenuDropdown = forwardRef<
     const lockPremiumEfforts = !isCreditPricedPlan(subscription.plan);
     const { isDark } = useTheme();
 
-    const { models, isModelsLoading } = useModels({
+    const { models, streams, isModelsLoading } = useModels({
       owner,
       disabled: !hasModelsPicker,
     });
@@ -65,8 +65,9 @@ export const PickModelSubMenuDropdown = forwardRef<
           lockPremiumEfforts,
           models,
           query,
+          streams,
         }),
-      [getModelIcon, lockPremiumEfforts, models, query]
+      [getModelIcon, lockPremiumEfforts, models, query, streams]
     );
 
     const handleSelect = (item: SlashCommand) => {
--- a/front/components/editor/extensions/shared/slash_suggestion/buildPickModelSlashCommandItems.test.ts
+++ b/front/components/editor/extensions/shared/slash_suggestion/buildPickModelSlashCommandItems.test.ts
@@ -1,6 +1,7 @@
 import { buildPickModelSlashCommandItems } from "@app/components/editor/extensions/shared/slash_suggestion/buildPickModelSlashCommandItems";
 import type { EnabledModelConfigurationType } from "@app/types/api/assistant/models";
 import { CLAUDE_SONNET_5_DEFAULT_MODEL_CONFIG } from "@app/types/assistant/models/anthropic";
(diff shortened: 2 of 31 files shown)


fontanierh (approved):
LGTM

PR #30569, opened 2026-08-14
[front] poc: Add metabase private in-cluster mcp server

id13:
## Description

Add the ability to bypass egress limitations to reach in-cluster IPs from an external MCP server instance, avoiding routing through the internet. 
This is an experimental feature targeted only to be used by our workspace for now.

Alternatives considered:

* IAP-aware mcp server => ❌ too big, you need to overfit for each IAP, there are tens if not hundreds around.
* Tunneling => worth doing one day, but scope too big. That being said some parts of this small change can be re-used.

## Tests

Tested locally

## Risk

Low, internal feature

## Deploy Plan

- In US, add the secret MCP_IN_CLUSTER_HOSTS with the correct value in front
- deploy front

### DIFF
--- a/front/lib/api/config.ts
+++ b/front/lib/api/config.ts
@@ -522,6 +522,18 @@ const config = {
       "UNTRUSTED_EGRESS_PROXY_PORT"
     );
   },
+  getInClusterMCPHosts: (): string[] => {
+    const hosts = EnvironmentConfig.getOptionalEnvVariable(
+      "MCP_IN_CLUSTER_HOSTS"
+    );
+    if (!hosts) {
+      return [];
+    }
+    return hosts
+      .split(",")
+      .map((host) => host.trim().toLowerCase())
+      .filter((host) => host.length > 0);
+  },
   getElasticsearchConfig: (): {
     url: string;
     username: string;
--- a/front/types/shared/feature_flags.ts
+++ b/front/types/shared/feature_flags.ts
@@ -186,6 +186,11 @@ export const WHITELISTABLE_FEATURES_CONFIG = {
       "NetSuite MCP tool for querying records and interacting with your NetSuite account",
     stage: "on_demand",
   },
+  dust_internal_dangerous_in_cluster_mcp_servers: {
+    description:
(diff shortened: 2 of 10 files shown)


front/components/actions/mcp/create/RemoteMCPServerConfigurationSection.tsx
```diff
@@ -31,6 +31,8 @@ function getAuthMethodLabel(
       return "Bearer token";
     case "oauth-static":
       return "Static OAuth";
+    case null:
+      return "None";
   }
```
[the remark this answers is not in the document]
id13: Smelly change that was not necessary, removed

front/lib/api/config.ts
```diff
@@ -522,6 +522,18 @@ const config = {
...
+      "MCP_IN_CLUSTER_HOSTS"
+    );
+    if (!hosts) {
+      return [];
+    }
+    return hosts
+      .split(",")
+      .map((host) => host.trim().toLowerCase())
+      .filter((host) => host.length > 0);
+  },
```
id13, line 536: PoC version, if it were a real feature we would obv do it through a data-model

front/types/shared/feature_flags.ts
```diff
@@ -186,6 +186,11 @@ export const WHITELISTABLE_FEATURES_CONFIG = {
       "NetSuite MCP tool for querying records and interacting with your NetSuite account",
     stage: "on_demand",
   },
+  in_cluster_mcp_servers: {
```
zmarouf: I would give the FF a scarier name (and set a monitor on the number of workspaces with that FF on)

zmarouf:
> IAP-aware mcp server => ❌ too big, you need to overfit for each IAP, there are tens if not hundreds around.

We could have solely supported GCP IaP since it's anyways only intended for Dust at this stage.

zmarouf (approved):

front/types/shared/feature_flags.ts
id13: Agreed

id13:
Renamed the FF

PR #30571, opened 2026-08-14
[dsbx] pod: tools.call() delegates to dsbx tools --json

fontanierh:
## Description

Follow-up to #30393. `@dust/pod`'s `tools.call()` duplicated the Rust CLI's server resolution, POST/poll transport, retries, and error handling, while still returning snippets for offloaded tool output.

`tools.call()` now delegates to:

```text
dsbx tools --json --args-json - <server> <tool>
```

Arguments are sent as one JSON object over stdin. The invocation-scoped API URL and sandbox token are passed explicitly to the child, and `Bun.spawn` keeps concurrent resident-worker invocations non-blocking. A 10m30s deadline covers both process exit and pipe draining.

The public API is unchanged except for `ToolCallErrorCode`, which now mirrors the CLI's typed JSON envelope plus local process errors. Unknown future CLI codes degrade to `unknown` while preserving the message and retryability. `@dust/pod` is bumped from 0.2.0 to 0.3.0.

## Tests

A fake `dsbx` executable covers argv/stdin delegation, invocation credential isolation, success and tool-level errors, structured output, typed and future error envelopes, malformed output, missing binary, non-zero exits, process timeout, and pipe-drain timeout. The pod and image registry suites pass.

## Risk

Low. The repository has no `tools.call()` consumers outside this package's tests. The main behavior change is that offloaded outputs are now fully resolved by `dsbx`; failures to read them surface as typed `tool_output_unavailable` errors. Delegation adds one server-view GET per call because the CLI does not share the former in-process cache.

Rollback is selecting the previous sandbox image.

## Deploy plan

- [x] Bump `DUST_BASE_IMAGE_VERSION` from 0.8.85 to 0.8.86 in this PR so the vendored `@dust/pod` 0.3.0 produces a new image.
- [x] Merge, then run the Sandbox Image Registry workflow from `main` for EU and US.
- [x] After the Front deploy, use `/poke/kill` for older `dust-base` sandboxes so existing conversations recreate on 0.8.86.

No `dsbx` release or pin bump is needed: 0.1.51 is already released, pinned by `DSBX_CLI_VERSION`, and contains #30393.

### DIFF
--- a/cli/dust-sandbox/pod/package.json
+++ b/cli/dust-sandbox/pod/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@dust/pod",
-  "version": "0.2.0",
+  "version": "0.3.0",
   "private": true,
   "description": "Runtime library for code running in a Dust pod sandbox.",
   "type": "module",
--- a/cli/dust-sandbox/pod/tools.test.ts
+++ b/cli/dust-sandbox/pod/tools.test.ts
@@ -1,5 +1,10 @@
-import { afterEach, beforeEach, describe, expect, test } from "bun:test";
+import { describe, expect, test } from "bun:test";
+import { chmodSync, mkdtempSync, readFileSync, writeFileSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
+
 import {
+  DSBX_PATH_ENV,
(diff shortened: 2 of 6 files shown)


davidebbo (commented):
LGTM

PR #30572, opened 2026-08-14
[front] enh: improve cat tool file labels

aubin-tchoi:
## Description

The `cat` tool currently displays full Dust FS paths, which can contain some tech-y conversation IDs and somtimes timestamp prefixes (for tool outputs for instance), which look a bit opaque and clunky.

This PR uses the file name for canonical conversation and Pod paths and removes the generated timestamp prefix from files under `.tool_outputs`. Non-Dust paths keep their existing labels.

## Tests

- Added tests to document the expected behavior.

## Risk

- Low.

## Deploy Plan

- Deploy front.


front/lib/actions/tool_display_labels.ts
```diff
@@ -13,8 +13,13 @@ import {
   isWebbrowseInputType,
   isWebsearchInputType,
 } from "@app/lib/actions/mcp_internal_actions/types";
+import {
+  parseCanonicalScopedPath,
+  TOOL_OUTPUTS_FOLDER_NAME,
+} from "@app/lib/api/files/mount_path";
```
flvndvd: Let's really avoid importing from the `api` folder here

flvndvd (approved):

front/lib/actions/tool_display_labels.test.ts
```diff
@@ -176,6 +176,45 @@ describe("getToolDisplayLabels", () => {
     });
   });
 
+  it.each([
+    [
+      "conversation-Fx7LXcp8VP/reports/quarterly-results.csv",
```
flvndvd, line 181: IMHO this one should preserve the full path so users can map it back to the file explorer.

front/lib/actions/tool_display_labels.test.ts
```diff
@@ -176,6 +176,45 @@ describe("getToolDisplayLabels", () => {
...
+      "“quarterly-results.csv” from conversation",
+    ],
+    [
+      "conversation-Fx7LXcp8VP/.tool_outputs/1786717219013_github_get_pull_request.txt",
```
flvndvd, line 190: I do understand that tool output can be noisy so fine to hide it.

front/lib/actions/tool_display_labels.test.ts
aubin-tchoi: Updated in 02da38f26b: ordinary scoped files now preserve their full path in the display label so they remain traceable in the file explorer.

front/lib/actions/tool_display_labels.test.ts
aubin-tchoi: restricted it to tool outputs so that the path that are not "specials" (reserved by us) are kept as is

front/lib/actions/tool_display_labels.ts
aubin-tchoi: had to fight a bit, took me a while to realize that we were already importing from `lib/api/files/mount_path` at a bunch of places in the front-end 🙃 
FYI I moved 2 files: `‎lib/api/file_system/types.ts` -> `types/file_system.ts` and `lib/api/files/mount_path.ts` -> `types/mount_path.ts`

PR #30581, opened 2026-08-14
Add "vs prev" growth column to consumption attribution table

avervaet:
## Description

Adds a `vs prev` column to the consumption attribution table for all supported dimensions (agents, users, groups, models, tools, skills, sources, and API keys). Each row compares the selected period's credits with an equal-length window immediately before it and displays the rounded percentage change with an upward or downward indicator. Increases above 100% use the `text-highlight-600` color.

**Visual:**
Capture d’écran 2026-08-14 à 18 44 08

When a row has no prior credits, the UI displays `N.A` with a `Not enough data to compute` tooltip rather than calculating an undefined percentage. **For billing-cycle periods, the comparison uses an equal-duration preceding window; it is not guaranteed to match the previous Metronome cycle when cycle lengths differ.**

## Tests

The existing consumption attribution table test fixture was updated for the new `previousCredits` field, and the tooltip assertion was made specific to the skill identity-card tooltip.

## Risk

Low - purely additive. The change affects the analytics attribution table and _adds one Elasticsearch_ aggregation for the previous period per paginated table request, scoped to the current page's keys rather than re-ranking the previous window. _The main risks are additional analytics query latency/load_ and the approximation of the previous billing-cycle window when cycle lengths differ. There is no data migration or persistent schema change; reverting the code removes the column and the additional aggregation.

## Deploy Plan

- Deploy front


front/lib/api/analytics/consumption/period.ts
```diff
@@ -68,6 +68,22 @@ async function resolveCycleBounds(
   };
 }
 
+export function previousConsumptionPeriod(
+  period: ConsumptionPeriod
```
aubin-tchoi: let's call this `currentPeriod`

front/components/workspace/analytics/consumption/ConsumptionAttributionTable.tsx
```diff
@@ -159,6 +162,54 @@ function AttributionTooltipCard({
...
+
+  return (
+    <div
+      className={cn(
+        "flex w-full items-center justify-end gap-1 text-right text-sm tabular-nums",
+        growth > 100 ? "text-highlight-600" : "text-muted-foreground"
+      )}
+    >
+      <Icon
+        visual={growth >= 0 ? ArrowNarrowUpRight : ArrowNarrowDownRight}
```
aubin-tchoi, line 205: nit: upward arrow for 0 may look odd

aubin-tchoi: actually that almost can't happen (since we only show nonzero values here)

front/components/workspace/analytics/consumption/ConsumptionAttributionTable.tsx
```diff
@@ -159,6 +162,54 @@ function AttributionTooltipCard({
   );
 }
 
+// Growth is undefined (not just zero) with no prior credits to grow from, so
+// callers must distinguish that case from an actual percentage.
+function growthPercent(
+  credits: number,
```
aubin-tchoi: nit: would be clearer if the name contained "current"

front/components/workspace/analytics/consumption/ConsumptionAttributionTable.tsx
```diff
@@ -159,6 +162,54 @@ function AttributionTooltipCard({
...
+  previousCredits: number | null
+): number | null {
+  return previousCredits && previousCredits > 0
+    ? ((credits - previousCredits) / previousCredits) * 100
+    : null;
```
aubin-tchoi, line 173: since the current credits is measured on a different duration having a null (= 0) growth means you're actually going faster (twice as fast if in the middle of the current cycle)

we should either count `previousCredits` over the same days of the past month or normalize with how far along we are in the current cycle

aubin-tchoi (approved):
LGTM modulo comment on growth computation

front/components/workspace/analytics/consumption/ConsumptionAttributionTable.tsx
aubin-tchoi: wait actually you seem to already be doing the first option, right? (if I'm on the n days of the current cycle `previousCredits` is also counted on the first n days of the previous cycle?)

PR #30582, opened 2026-08-14
Add filesystem namespace reads

flvndvd:
## Description

See design document [here](https://app.notion.com/p/dust-tt/Design-Doc-Robust-Dust-File-System-3bb28599d94180ca8bfddcbd2ab7b2c3)

Dust’s new filesystem keeps its namespace in PostgreSQL. Each file or directory is represented by a stable node. Its parent and name determine where it appears, while its node ID remains unchanged when it is renamed or moved. The node also records its Conversation or Pod root and the metadata needed by the filesystem.

Access is restricted through a `FileSystemScope`, built after permissions have been checked. All reads are filtered by both workspace and the Conversation or Pod roots present in that scope.

This PR introduces the read side of the namespace.

`Namespace` is the entry point for filesystem operations. `FileSystemNodeResource` owns all database access, while the scope and operation types live in small supporting files.

Initialization creates the requested Conversation and Pod roots when needed. It is safe to retry and returns the same node IDs. Root creation is batched so adding more mounts does not produce one query per root.

Three read operations are currently supported:

### lookup

Finds a child from its parent node ID and name. It returns `null` when the child does not exist, and `not_found` when the parent is missing, outside the allowed scope, or not a directory.

### getAttr

Returns the stored metadata for a node: its parent, root, name, kind, mode, size, blob pointer, content revision and timestamps. Nodes outside the allowed scope are treated as not found.

### readDir

Lists the direct children of a directory in name order. Results are paginated using `afterName`, with page sizes limited to max 256 entries.

This PR deliberately stops before writes. Create, content updates, rename/move and delete will follow as separate changes. It also does not expose an HTTP endpoint or connect the namespace to the FUSE daemon yet.

### DIFF
--- a/front/lib/api/file_system/namespace.test.ts
+++ b/front/lib/api/file_system/namespace.test.ts
@@ -1,7 +1,7 @@
 import { randomUUID } from "node:crypto";
 import { applyFileSystemOperation } from "@app/lib/api/file_system/namespace";
 import { FileSystemScope } from "@app/lib/api/file_system/namespace_scope";
-import type { FileSystemNode } from "@app/lib/api/file_system/namespace_types";
+import type { FileSystemNodeType } from "@app/lib/api/file_system/namespace_types";
 import { createResourceTest } from "@app/tests/utils/generic_resource_tests";
 import { describe, expect, it } from "vitest";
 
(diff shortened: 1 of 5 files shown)


smb2268 (approved):
lgtm!

PR #30585, opened 2026-08-14
Add filesystem namespace creation

flvndvd:
## Description

See design document [here](https://app.notion.com/p/dust-tt/Design-Doc-Robust-Dust-File-System-3bb28599d94180ca8bfddcbd2ab7b2c3)

Dust’s filesystem namespace gives every file and directory a stable node. Its parent and name determine where it appears, while its node ID remains unchanged throughout its lifetime.

The previous PR (https://github.com/dust-tt/dust/pull/30582) introduced namespace initialization and reads. This PR adds the first write operation: creating a node.

## What this PR adds

`createNode` creates either an empty file or a directory beneath an existing directory.

The parent is fetched once as a `FileSystemNodeResource`, which then owns the creation of its child. The operation checks that the parent exists, is a directory, and belongs to a writable Conversation or Pod root.

The node and its mutation receipt are created in the same database transaction. Retrying the same mutation is therefore safe and returns the result of the first attempt without creating another node.

Names must be unique within a directory. Creating a different node at an existing name returns a conflict instead of replacing it.

New files contain no bytes yet. Uploading content and attaching a GCS blob will come in a later PR. Rename, move and delete are also outside this change.

### DIFF
--- a/front/lib/api/file_system/namespace.test.ts
+++ b/front/lib/api/file_system/namespace.test.ts
@@ -290,7 +290,7 @@ describe("filesystem namespace creation", () => {
     }
     const root = requireNode(initialized.value.roots?.[0]);
 
-    const unauthorized = await applyFileSystemOperation(
+    const writeDenied = await applyFileSystemOperation(
       authenticator,
       readOnlyScope(conversationId),
       {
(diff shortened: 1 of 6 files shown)


smb2268 (approved):
lgtm!

PR #30587, opened 2026-08-14
[front] perf: speed up MCP server facet loading in new analytics

aubin-tchoi:
## Description

Part of https://github.com/dust-tt/tasks/issues/10069.

The MCP server facet is currently loading MCP server views using `listByWorkspace` to get names and icons.
This method is actually very inefficient, but for the wrong reasons (not for permission check): it does several joins/separate queries that are not always useful and hard to opt-out from in the current state of things, fetching twice on `mcp_server_views` for instance.

To get things moving without having to refactor this too much, this PR adds a small resource method that uses at most 2 queries instead of 7: one for views and one for remote servers, and reads the data for internal server metadata directly from the object we have already defined in the code. I'll optimize the common code path later (quite large blast radius).

Next step is to add some resource-level caching on the resources fetched for the facets here.

## Tests

- Added tests.

## Risk

- Low.

## Deploy Plan

- Deploy front.

### DIFF
--- a/front/lib/resources/mcp_server_view_resource.ts
+++ b/front/lib/resources/mcp_server_view_resource.ts
@@ -97,6 +97,14 @@ type MCPServerViewCreationResult = {
   affectedAgents?: AffectedAgent[];
 };
 
+export type MCPServerViewDisplayMetadata = {
+  serverType: "internal" | "remote";
+  viewName: string | null;
+  mcpServerId: string;
+  serverName: string;
+  icon: CustomResourceIconType | InternalAllowedIconType;
+};
+
 export type GetMCPServerViewsResponseBody = {
   success: boolean;
   serverViews: MCPServerViewType[];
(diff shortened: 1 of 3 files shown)


avervaet (approved):
Nice! I agree a deeper refactoring could be very beneficial for the endpoints fetching MCP server info.

front/lib/resources/mcp_server_view_resource.ts
```diff
@@ -700,6 +708,69 @@ export class MCPServerViewResource extends ResourceWithSpace<MCPServerViewModel>
...
+      auth,
+      remoteServerModelIds
+    );
+    const remoteServersById = new Map(
+      remoteServers.map((server) => [server.id, server])
+    );
+
+    return removeNulls(
+      views.map((view): MCPServerViewDisplayMetadata | null => {
+        if (view.serverType === "remote" && view.remoteMCPServerId) {
```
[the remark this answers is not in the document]
aubin-tchoi: adding a new server type is not really a use case, and we'll track down this one fairly easily since it's on the resource

front/lib/resources/mcp_server_view_resource.ts
```diff
@@ -700,6 +708,68 @@ export class MCPServerViewResource extends ResourceWithSpace<MCPServerViewModel>
...
+        "internalMCPServerId",
+        "remoteMCPServerId",
+      ],
+      where: { workspaceId: auth.getNonNullableWorkspace().id },
+    });
+
+    const remoteServerModelIds = [
+      ...new Set(removeNulls(views.map((view) => view.remoteMCPServerId))),
+    ];
+    const remoteServers = await RemoteMCPServerResource.fetchByModelIds(
```
flvndvd, line 727: Looking at this guy, it does not check/enforce any permission, who owns this responsibility here?

aubin-tchoi: everything relative to permissions should be in `RemoteMCPServerResource`'s baseFetch (AFAICT there are no permission checks apart from the workspace check)

flvndvd: That feels wrong, no? Not everyone should be able to see remote MCP servers?

aubin-tchoi: should be handled before fetching from `remote_mcp_servers`, I indeed need to replace the findall with a `baseFetchWithAuthorization`, fixing

PR #30589, opened 2026-08-14
Add filesystem namespace content

flvndvd:
## Description

See full design doc [here](https://app.notion.com/p/dust-tt/Design-Doc-Robust-Dust-File-System-3bb28599d94180ca8bfddcbd2ab7b2c3).

This PR adds file content to the filesystem namespace, on top of the node and create operations from the previous PRs.

File bytes remain in GCS, while the filesystem node stores the current blob ID, size, MIME type, and content revision. Each upload gets a new immutable blob ID, so replacing content never overwrites bytes already in use.

About the verbs:
- `getContent` returns the current file metadata and a short-lived signed download URL.
- `prepareContentUpload` checks write access and the expected current blob, normalizes the MIME type, validates the file size, and creates a signed, create-only GCS upload URL. The signed headers bind the upload to its exact size, MIME type, and identity encoding (like we do on our existing File's API).
- `commitContentUpload` verifies the uploaded GCS object before attaching it to the node. The node update uses the expected previous blob as a compare-and-swap guard, so concurrent writers cannot silently replace each other. Retrying an already successful commit is idempotent and does not require another GCS request.

Uploads and replaced blobs are recorded in a durable cleanup table. This PR owns those state changes but intentionally leaves the cleanup worker for a later PR, where retries, fair claiming, and GCS calls outside database transactions can be handled properly.


front/lib/api/file_system/namespace_types.ts
```diff
@@ -16,6 +16,11 @@ export const FILE_SYSTEM_MODE_LIMITS = {
 
 export const FILE_SYSTEM_REQUEST_ID_MAX_LENGTH = 255;
 
+export const FILE_SYSTEM_CONTENT_TYPE_MAX_LENGTH = 255;
+
+// Matches the largest raw file upload currently accepted by Dust.
+export const FILE_SYSTEM_CONTENT_MAX_BYTES = 350 * 1024 * 1024;
```
smb2268, line 22: Do we have this const anywhere else? Maybe we should import it so they don't drift

front/lib/resources/file_system_node_resource.ts
```diff
@@ -336,6 +376,439 @@ export class FileSystemNodeResource extends BaseResource<FileSystemNodeModel> {
...
   }
 
+  private checkContentAccess(
+    auth: Authenticator,
+    scope: FileSystemScope,
+    { write }: { write: boolean }
+  ): Result<undefined, FileSystemOperationError> {
+    if (!this.isReadableBy(auth, scope)) {
+      return new Err(
+        new FileSystemOperationError("not_found", "The file was not found.")
```
smb2268, line 386: Is this distinct from the file existing but not having permission to read? Or can we not tell the difference?

front/lib/resources/file_system_node_resource.ts
```diff
@@ -336,6 +376,439 @@ export class FileSystemNodeResource extends BaseResource<FileSystemNodeModel> {
...
+          "invalid_operation",
+          "The uploaded object has an invalid size."
+        )
+      );
+    }
+    if (metadata.size !== request.expectedSizeBytes) {
+      return new Err(
+        new FileSystemOperationError(
+          "invalid_operation",
+          "The uploaded object's size does not match the request."
```
smb2268, line 651: I don't totally understand what this is guarding against - is it likely that we have drift in one metadata field?

smb2268 (approved):
A few questions so I understand fully but LGTM

front/lib/api/file_system/namespace_types.ts
flvndvd: Not yet, we had previous but not very relevant for this new implementation.

front/lib/resources/file_system_node_resource.ts
flvndvd: We can tell the difference, but intentionally return `not_found` when the file is outside the caller’s readable scope to avoid revealing its existence. If the file is readable but the caller lacks write access, write operations return `unauthorized`.

front/lib/resources/file_system_node_resource.ts
flvndvd: As we don't want to bother dealing with bytes and network issues, the front api will never handle upload/download, instead when the daemon wants to upload content, it first sends a `prepareContentUpload` which will prepare a signed URL directly on GCS and we can restrict many parameters on this signed link, one of them being the size and content type, which the daemon sent alongside the `prepareContentUpload`.

PR #30590, opened 2026-08-14
  Add POSIX rename and move to the filesystem namespace

flvndvd:
## Description

This PR adds rename and move support to the filesystem namespace. A node can move within a directory, between directories, or between Conversation and Pod roots while keeping the same stable identity.

Rename follows normal POSIX behavior. An existing file is replaced atomically, and an existing empty directory can be replaced by another directory. Replaced file content is handed to the existing blob cleanup flow. Type mismatches, non-empty destination directories, invalid names, and directory cycles are rejected without partially changing the tree.

Moving a directory between roots updates the cached root information for every descendant. This keeps authorization checks simple and fast during normal reads. It **sadly** requires a recursive tree update during the move. The code documents this as a narrow exception and calls out that we should revisit the data model if large cross-root moves become common. We could also define a max depth on this new file system. A second recursive query walks a directory’s parents to prevent cycles.

Namespace writes use a workspace-scoped PostgreSQL lock. Creates take a shared lock and can still run concurrently. Rename takes the lock exclusively so it cannot miss a child being created while a directory tree is changing roots. The lock is taken before node row locks to keep ordering consistent and avoid deadlocks. Reads and content transfers do not take this lock.

Rename requests are idempotent. The mutation receipt stores the exact committed response, validated with Zod when replayed. If the caller loses the response and retries after the node has since moved or been replaced again, it still receives the result of its original rename rather than the node’s newer state.

This PR only adds the namespace operation in Front. Wiring it into the filesystem daemon comes later.

### DIFF
--- a/front/lib/resources/file_system_node_resource.ts
+++ b/front/lib/resources/file_system_node_resource.ts
@@ -493,11 +493,11 @@ export class FileSystemNodeResource extends BaseResource<FileSystemNodeModel> {
     possibleDescendant: FileSystemNodeResource,
     transaction: Transaction
   ): Promise<boolean> {
-    // Deliberate recursive CTE: the adjacency-list model only stores parentId,
-    // so preventing a directory cycle requires walking toward the root. This
-    // is O(directory depth), not O(subtree size). This is one of the few places
-    // where recursive SQL is allowed; do not reuse it for ordinary reads.
-    // biome-ignore lint/plugin/noRawSql: Recursive tree traversal has no Sequelize equivalent.
+    // Each row only stores its parentId. To stop a directory from being moved
+    // inside itself, this query follows parentId until it reaches a root. Its
+    // work grows with the number of parent directories, not the number of
+    // children. Keep recursive SQL out of normal reads.
+    // biome-ignore lint/plugin/noRawSql: Sequelize cannot follow parentId until the root.
     const rows = await frontSequelize.query<{ found: number }>(
       `
         WITH RECURSIVE ancestors AS (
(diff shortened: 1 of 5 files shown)


front/lib/resources/file_system_node_resource.ts
```diff
@@ -207,6 +222,29 @@ export class FileSystemNodeResource extends BaseResource<FileSystemNodeModel> {
     return node ?? null;
   }
 
+  private static validateName(
+    name: string
+  ): Result<undefined, FileSystemOperationError> {
+    const nameLengthBytes = Buffer.byteLength(name, "utf8");
+    if (
+      name === "." ||
```
smb2268, line 239: Is it worth making a helper function to do these checks since they're repeated elsewhere?

front/lib/resources/file_system_node_resource.ts
```diff
@@ -376,6 +436,256 @@ export class FileSystemNodeResource extends BaseResource<FileSystemNodeModel> {
     });
   }
 
+  private async isAncestorOf(
+    possibleDescendant: FileSystemNodeResource,
+    transaction: Transaction
+  ): Promise<boolean> {
+    // Each row only stores its parentId. To stop a directory from being moved
+    // inside itself, this query follows parentId until it reaches a root. Its
```
smb2268, line 497: inception

front/lib/resources/file_system_node_resource.ts
```diff
@@ -376,6 +436,256 @@ export class FileSystemNodeResource extends BaseResource<FileSystemNodeModel> {
...
+      );
+    }
+    if (
+      destination &&
+      !(await destination.isEmptyDirectory(options.transaction))
+    ) {
+      return new Err(
+        new FileSystemOperationError(
+          "not_empty",
+          "The destination directory is not empty."
```
smb2268, line 672: No siblings allowed?

smb2268 (commented):
LGTM

front/lib/resources/file_system_node_resource.ts
flvndvd: Hum that should be the only one, we previously were using `path.posix` but this one only works on actual path and not filename.

front/lib/resources/file_system_node_resource.ts
flvndvd: It does not forbid siblings. It only checks children inside the existing directory at the destination path.

Example: renaming `source/` to `parent/target/` may replace `target/` only when `target/` is empty. Other entries beside `target/` under `parent/` are allowed.

PR #30591, opened 2026-08-14
Add file and directory removal to the filesystem

flvndvd:
## Description

This PR adds removal support to the new filesystem. A request provides the parent ID, name, and expected kind. The FUSE daemon will use file for Linux `[unlink](https://man7.org/linux/man-pages/man2/unlink.2.html)` and directory for `[rmdir](https://man7.org/linux/man-pages/man2/rmdir.2.html)`.

Removing a file deletes its node and queues its blob for deletion in the same PostgreSQL transaction. The blob remains available until previously issued download URLs have expired. Removing an empty directory deletes its node without running recursive SQL.

The operation follows Linux behavior. Removing a directory as a file returns `is_directory`, removing a file as a directory returns `not_directory`, and removing a non-empty directory returns `not_empty`. Missing names, invalid names, and read-only roots also fail without deleting anything. Recursive commands such as `rm -rf` will remove children individually before removing their parent.

Create and remove take the same shared workspace lock, so work in different directories can still run at the same time. The parent and target rows are locked while checking and deleting them. This prevents a child from being added after an empty-directory check. It also makes removal wait while a cross-root directory move is running.

A successful removal is saved in the existing filesystem request table. If the response is lost, retrying the same request returns success without trying to delete the missing node again. Reusing the request ID with another parent, name, or kind is rejected.

This PR only adds the front operation. Calling it from the FUSE daemon comes later!

### DIFF
--- a/front/lib/api/file_system/namespace.test.ts
+++ b/front/lib/api/file_system/namespace.test.ts
@@ -935,6 +935,279 @@ describe("filesystem namespace rename", () => {
   });
 });
 
+describe("filesystem namespace removal", () => {
+  it("removes a file, retires its blob, and replays the request", async () => {
+    const { authenticator } = await createResourceTest({ role: "admin" });
+    const scope = readableScope(`conversation-${randomUUID()}`);
+    const initializedRes = await applyFileSystemOperation(
+      authenticator,
+      scope,
+      { operation: "initialize" }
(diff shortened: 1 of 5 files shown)


smb2268 (approved):
nice tiny one

PR #30592, opened 2026-08-14
[Activation] Sample from the user's work and plan Skill/Trigger rungs

frankaloia:
## Summary
- Various skill improvements based on early feedback

## Testing
Many local activation runs

## Deploy
1. Deploy front

## Risk
Very low


smb2268 (approved):
lgtm

PR #30593, opened 2026-08-14
Add executable bit updates to the filesystem namespace

flvndvd:
## Description

At first, I was not planning to support this, but models often do this on a sandbox:
```
chmod +x ./my_script.sh
./my_script.ts
```

To make sure those changes persist in our filesystem, this PR adds the filesystem operation needed for `chmod +x` and `chmod -x`.

The caller sends the exact user, group, and other executable bits to store. Front preserves the existing read, write, setuid, setgid, and sticky bits. Requests attempting to change anything outside `0o111` (this is JS octal [value](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Lexical_grammar#octal)) for ([doc](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/sys_stat.h.html))  are rejected.

The operation checks that the node belongs to a readable root and that the caller has write access. It updates the existing node, so its inode and path remain unchanged. Retrying the same request is safe.

This deliberately does not add general `chmod` support. The application currently needs executable files in sandboxes, but has no product use for changing read or write permissions (who knows, maybe in the future!).

### DIFF
--- a/front/lib/api/file_system/namespace.test.ts
+++ b/front/lib/api/file_system/namespace.test.ts
@@ -693,3 +693,105 @@ describe("filesystem content", () => {
     );
   });
 });
+
+describe("filesystem executable bits", () => {
+  it("changes executable bits without changing the other mode bits", async () => {
+    const { authenticator } = await createResourceTest({ role: "admin" });
+    const scope = readableScope(`conversation-${randomUUID()}`);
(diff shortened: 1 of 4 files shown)


smb2268 (approved):
LGTM 🚢

PR #30594, opened 2026-08-14
Add the sandbox filesystem endpoint and scoped token

flvndvd:
## Description

This PR adds the private HTTP endpoint used by the Dust filesystem daemon. The daemon sends one filesystem operation at a time. This API validates it, applies it through the namespace resources, and returns the node or content response.

It also adds a filesystem-specific sandbox token. The token contains the exact roots mounted in the sandbox: at most one Conversation and one Pod, with separate `read` and `write` permissions. A request cannot add another root or make a read-only root writable.

Filesystem tokens only work on this endpoint. Existing action and function tokens are rejected, and filesystem tokens cannot call other sandbox endpoints.

Requests and responses are validated with Zod. Filesystem errors also include an error header that the Rust daemon uses to return the matching Linux error.

### DIFF
--- a/front-api/routes/v1/w/[wId]/sandbox/filesystem.test.ts
+++ b/front-api/routes/v1/w/[wId]/sandbox/filesystem.test.ts
@@ -1,0 +1,171 @@
+import { randomUUID } from "node:crypto";
+import { FileSystemOperationResponseSchema } from "@app/lib/api/file_system/namespace";
+import { generateSandboxFileSystemToken } from "@app/lib/api/sandbox/access_tokens";
+import { createSandboxTokenTestContext } from "@app/tests/utils/SandboxTokenFactory";
+import { SpaceFactory } from "@app/tests/utils/SpaceFactory";
+import { honoApp } from "@front-api/app";
+import { describe, expect, it } from "vitest";
+
+function requestFileSystem(
+  workspaceId: string,
+  token: string,
+  operation: object
+) {
+  return honoApp.request(`/api/v1/w/${workspaceId}/sandbox/filesystem`, {
+    method: "POST",
+    headers: {
+      authorization: `Bearer ${token}`,
+      "content-type": "application/json",
+    },
+    body: JSON.stringify(operation),
+  });
+}
+
+describe("POST /api/v1/w/[wId]/sandbox/filesystem", () => {
+  it("uses only the roots and permissions signed into the token", async () => {
+    const context = await createSandboxTokenTestContext();
+    const pod = await SpaceFactory.project(
+      context.workspace,
+      context.auth.getNonNullableUser().id
+    );
+    const token = await generateSandboxFileSystemToken(context.auth, {
+      sandbox: context.sandbox,
+      roots: [
+        {
+          kind: "conversation",
+          id: context.conversation.sId,
+          permissions: { canRead: true, canWrite: true },
+        },
+        {
+          kind: "pod",
+          id: pod.sId,
+          permissions: { canRead: true, canWrite: false },
+        },
+      ],
+    });
+
+    const initializeResponse = await requestFileSystem(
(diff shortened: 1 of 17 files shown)


front/lib/api/sandbox/access_tokens.ts
```diff
@@ -22,6 +22,31 @@ function isDefined<T>(value: T | undefined): value is T {
...
+  .max(2)
+  .refine(
+    (roots) => new Set(roots.map((root) => root.kind)).size === roots.length,
+    "A filesystem token cannot contain the same root kind twice."
+  );
+
+export type SandboxFileSystemRoot = z.infer<typeof SandboxFileSystemRootSchema>;
+
 const SandboxTokenPayloadSchema = z
   .object({
```
flvndvd, line 51: This type is all broken, no idea why we did not used a proper discriminated union. Will refactor in a follow up PR.

front/lib/api/sandbox/access_tokens.ts
```diff
@@ -323,6 +393,39 @@ export async function generateSandboxFunctionInvocationToken(
...
+    sandbox: SandboxResource;
+    roots: SandboxFileSystemRoot[];
+    expiryMs?: number;
+  }
+): Promise<string> {
+  const fileSystemRoots = SandboxFileSystemRootsSchema.parse(roots);
+
+  const payload: SandboxFileSystemTokenPayload = {
+    wId: auth.getNonNullableWorkspace().sId,
+    uId: auth.user()?.sId,
```
adrsimon, line 412: in what world do we have a user-less filesystem token ?

front/lib/api/sandbox/access_tokens.ts
```diff
@@ -172,6 +224,24 @@ export function isSandboxFunctionInvocationTokenPayload(
...
+    payload.aId === undefined &&
+    payload.aV === undefined &&
+    payload.mId === undefined &&
+    payload.actionId === undefined &&
+    payload.spaceId === undefined &&
+    payload.sandboxFunctionId === undefined &&
+    payload.invocationId === undefined &&
+    payload.noTools === undefined
+  );
+}
```
adrsimon, line 243: should we have conditions over `payload.filesystem === undefined && payload.fileSystemRoots === undefined` in `isSandboxFunctionInvocationTokenPayload` to mirror the new `isSandboxFileSystemTokenPayload` ?

actually, do we need the `payload.filesystem` in filesystem tokens ? looks repetitive with fsRoots

front/lib/api/sandbox/access_tokens.ts
adrsimon: :+1:

front-api/routes/v1/w/[wId]/sandbox/filesystem.test.ts
```diff
@@ -0,0 +1,171 @@
...
+      ],
+    });
+
+    const response = await honoApp.request(
+      `/api/v1/w/${context.workspace.sId}/sandbox/actions`,
+      { headers: { authorization: `Bearer ${token}` } }
+    );
+
+    expect(response.status).toBe(403);
+  });
```
Comment (adrsimon) [inline front-api/routes/v1/w/[wId]/sandbox/filesystem.test.ts:170]: nit: not really a test over /sandbox/filesystem per-se

adrsimon (approved):
LGTM

front/lib/api/sandbox/access_tokens.ts
flvndvd: Pod sandbox are not attached to a specific user as FS there is mainly R/O.

front/lib/api/sandbox/access_tokens.ts
flvndvd: I want clear distinct types, see comment above, will probably do this in a follow up PR so we can easily discriminate on the token type.

PR #30596, opened 2026-08-15
Add the Rust client for the Dust filesystem

flvndvd:
## Description

This is the first of three PRs for the Dust filesystem daemon. It adds the Rust client that talks to Front, without adding the inode store, cache, or FUSE mount yet.

Front remains responsible for file names, directories, permissions, revisions, and moves. File bytes travel directly between the sandbox and GCS using short-lived signed URLs, so the sandbox does not receive bucket credentials and Front does not carry file contents.

The client supports namespace reads and writes, content downloads and uploads, and executable-bit changes. Retried mutations reuse the same request ID so a lost HTTP response does not apply an operation twice. Front errors are translated into the matching filesystem errors for the later FUSE layer.

Uploads follow three steps: ask Front for a signed upload, send the exact bytes and every header signed by Front, then commit the new blob and revision.

The client also uses bounded retries and timeouts, reloads its token without following symbolic links, and avoids including authenticated request URLs in errors.

The next PR adds the inode mapping, local staging, and bounded caches. The third adds the FUSE callbacks, mount command, and supervisor.


cli/dust-sandbox/Cargo.toml
```diff
@@ -20,7 +20,7 @@ lru = "0.16"
 httparse = "1"
 serde = { version = "1", features = ["derive"] }
 serde_json = "1"
-reqwest = { version = "0.12", default-features = false, features = ["json", "rustls-tls-native-roots"] }
+reqwest = { version = "0.12", default-features = false, features = ["blocking", "json", "rustls-tls-native-roots"] }
```
flvndvd, line 23: Adding `features = ["blocking"]` here to enable reqwest’s synchronous HTTP client. The daemon’s filesystem callbacks are synchronous, so the client performs each Front/GCS request on the callback’s worker thread instead of introducing a Tokio runtime into this layer.
The FUSE layer will limit how many such remote operations may run at once, so “blocking” here is deliberate.

cli/dust-sandbox/src/commands/filesystem/client.rs
```diff
@@ -0,0 +1,587 @@
...
+    upload: Option<ContentUpload>,
+}
+
+#[derive(Serialize)]
+#[serde(
+    tag = "operation",
+    rename_all = "camelCase",
+    rename_all_fields = "camelCase"
+)]
+enum Operation<'a> {
```
flvndvd, line 112: Those have all been implemented in previous PRs. on the front's API side.

flvndvd, line 112: I could have avoided the lifetime (`a) here but that would have mean plenty of `.clone` required.

For example, `Create` borrows `request_id` and `name`, serializes them immediately, then disappears. This avoids allocating new `String` values for every filesystem request.

cli/dust-sandbox/src/commands/filesystem/client.rs
```diff
@@ -0,0 +1,587 @@
...
+                filesystem_error = code.as_deref(),
+                "filesystem metadata request failed"
+            );
+            return Err(filesystem_error(code.as_deref()));
+        }
+        response.json().map_err(network_error)
+    }
+
+    fn send_operation(&self, operation: &Operation<'_>) -> io::Result<reqwest::blocking::Response> {
+        let token = self.token.lock().map_err(|_| errno(libc::EIO))?.clone();
```
flvndvd, line 298: EIO = generic “input/output error” on linux. See [doc](https://man7.org/linux/man-pages/man3/errno.3.html).

cli/dust-sandbox/src/commands/filesystem/client.rs
```diff
@@ -0,0 +1,587 @@
...
+        Some("not_empty") => errno(libc::ENOTEMPTY),
+        Some("not_found") => errno(libc::ENOENT),
+        Some("stale") => errno(libc::ESTALE),
+        Some("unauthorized") => errno(libc::EACCES),
+        _ => errno(libc::EIO),
+    }
+}
+
+fn network_error(error: reqwest::Error) -> io::Error {
+    // reqwest's Display includes the full URL. Filesystem content URLs are
```
flvndvd, line 395: We are aiming at avoid exposing signed GCS URL.

cli/dust-sandbox/src/commands/filesystem/client.rs
```diff
@@ -0,0 +1,728 @@
...
+        token_file: PathBuf,
+    ) -> io::Result<Self> {
+        let http = Client::builder()
+            .connect_timeout(std::time::Duration::from_secs(5))
+            .build()
+            .map_err(io::Error::other)?;
+        Ok(Self {
+            http,
+            endpoint: format!(
+                "{}/api/v1/w/{}/sandbox/filesystem",
```
davidebbo, line 189: This doesn't exist yet, right? And does it need to be a /v1 api?

flvndvd: It's the last PR on front see [here](https://github.com/dust-tt/dust/pull/30594). I went with v1 to align on the existing sandbox endpoints (e.g calling actions, and pod functions).

davidebbo (approved):
LGTM

PR #30597, opened 2026-08-15
Add inode identity and local staging for the Dust filesystem

flvndvd:
## Description

This is the second of three PRs for the Dust filesystem daemon.

This PR adds the local store that sits between the Front client and the FUSE callbacks. It does not mount a filesystem yet, the next PR will add that Linux-facing layer!

Front node IDs remain the stable file identity. Most node IDs can be used directly as Linux inode numbers. Linux reserves inode 1 for the mounted root, so the possible database node 1 (will only impact cell-0000 in the US) is mapped to one reserved high value. The mapping is reversible, requires no extra table, and keeps the same inode when a file moves between a Conversation and a Pod.

File contents are staged in a private local directory. Reads download the current blob only when it is not already cached. Writes change the staged file, sync it locally, upload it through the signed GCS flow from the first PR, then ask Front to commit the new blob and revision. Front’s expected blob check protects against another sandbox changing the file in the meantime.

The content cache has a byte limit and removes the least recently used closed files first. Open files remain pinned and may temporarily exceed the limit. Metadata and directory listings use small bounded caches with a one-second lifetime (we might revisit as @fontanierh recently disabled cache on most existing mounts), matching the later FUSE cache window. Separate files can download or upload in parallel, while operations on the same inode are kept from replacing each other’s staged file.

The staging directory is owned by the daemon and rejects unsafe permissions. Cached files are opened with `O_NOFOLLOW`, and leftover files are removed on startup because Front and GCS (not the local cache) remain the source of truth after a crash.

This PR also introduces the Linux fuser `inode-number` type, but does not expose any FUSE callbacks yet.


cli/dust-sandbox/src/commands/filesystem/store.rs
```diff
@@ -0,0 +1,578 @@
...
+        }
+        let node = Node::from_remote(self.client.lookup(node_id_for_inode(parent_inode)?, name)?)?;
+        self.cache_node(node.clone())?;
+        Ok(node)
+    }
+
+    pub fn children(&self, parent_inode: INodeNo) -> io::Result<Vec<Node>> {
+        if parent_inode == FUSE_ROOT_INODE {
+            let mut roots = self.roots.clone();
+            for alias in ["conversation", "pod"] {
```
flvndvd, line 120: Hardcoded the alias here for now as we don't support symlink in this minimal fuse daemon.

cli/dust-sandbox/src/commands/filesystem/store/content_cache.rs
```diff
@@ -0,0 +1,562 @@
...
+            if state.entries.contains(&self.inode) {
+                ContentCache::release_open(&mut state, self.inode)?;
+            }
+            ContentCache::trim_locked(&mut state)
+        })();
+        if let Err(error) = result {
+            warn!(
+                inode = self.inode.0,
+                %error,
+                "failed to release staged filesystem content"
```
aubin-tchoi: OOC the logs will be exposed to the agent right?

cli/dust-sandbox/src/commands/filesystem/store.rs
```diff
@@ -0,0 +1,580 @@
...
+        self.client.upload(&upload, upload_file, size)?;
+        let remote = self.client.commit_upload(
+            node_id,
+            opened.expected_blob_id.as_deref(),
+            &upload,
+            size,
+        )?;
+        let node = Node::from_remote(remote)?;
+        self.content.update(inode, &node, size)?;
+        self.cache_node(node.clone())?;
```
aubin-tchoi: there's technically a path where a statement above errors and we end up with a stale cache
I feel like we could invalidate the cache much more often, basically whenever something goes wrong

aubin-tchoi (approved):
LGTM
One side note: if there are issues it'll be a lot easier to debug if we ship the caching part separately after a first round of tests

flvndvd:
> LGTM One side note: if there are issues it'll be a lot easier to debug if we ship the caching part separately after a first round of tests

True, but it would send hundreds of calls to our API 🙃

cli/dust-sandbox/src/commands/filesystem/store.rs
flvndvd: So you mean if we fail to update local state, we should invalidate both caches? Updated in a way that makes the boundary explicit: staged bytes survive errors while a writable handle can retry, but after front's API accepts a commit, local refresh failures drop both caches.

PR #30608, opened 2026-08-17
Handle unsupported Notion formula results

fontanierh:
## Description

Treat Notion formula results with `type: "unsupported"` as empty property values. Notion emits this variant, but the pinned SDK v2 does not model it, so exhaustiveness checking currently retries the entire page render.

## Tests

Added regression coverage for unsupported and supported formula results. Tested locally with the targeted Vitest suite.

## Risk

Low. Only unsupported formula metadata is omitted while the rest of the page still syncs. Safe to roll back.

## Deploy Plan

Standard deploy.


davidebbo (approved):
LGTM

PR #30609, opened 2026-08-17
perf: merge consumption top-N ranking + vs-prev queries into one ES call

avervaet:
## Summary
- `fetchConsumptionTopGroups` (backing every consumption Top-N widget: agents, users, models, tools, skills, sources, api keys, groups) issued two sequential Elasticsearch requests — the ranking query, then a second query for the period-over-period ("vs prev") comparison.
- Merged both into a single query spanning the union of the current and previous periods, using nested `filter` aggregations to isolate each period's credits/count within the same `terms` ranking. Halves the ES round-trip count for every attribution table and breakdown column on the consumption analytics page.
- `total_count` and `total_credit_micro` are now scoped via a `current_period` filter agg too, since the outer query date range now spans both periods.
- Padded `shard_size` on the ranking `terms` agg, since a term active only in the previous period can now appear as a candidate bucket; those buckets are filtered out of the final ranking in code so results are unaffected.

## Known tradeoff
Previously, a failure fetching vs-prev credits didn't take down the ranking (it fell back to null growth for every row). Since both are now one request, a failure there fails the whole ranking. Given this is read-only analytics tooling (not a mutation path), this seemed like an acceptable tradeoff for halving the query count, but flagging explicitly for review.

## Test plan
- [x] `front/lib/api/analytics/consumption/top.test.ts` updated for the new single-request aggregation shape (16 tests), plus new coverage for stale-candidate filtering and the merged-failure path.
- [x] `npx tsgo --noEmit` passes across `front`.
- [x] Full `lib/api/analytics/consumption` test suite passes (66 tests).

### DIFF
--- a/front/lib/api/analytics/consumption/top.test.ts
+++ b/front/lib/api/analytics/consumption/top.test.ts
@@ -1,6 +1,7 @@
 import { listConsumptionFacetCatalogDimension } from "@app/lib/api/analytics/consumption/facet_catalog";
 import { resolveDimensionLabels } from "@app/lib/api/analytics/consumption/labels";
 import type { ConsumptionPeriod } from "@app/lib/api/analytics/consumption/period";
+import { previousConsumptionPeriod } from "@app/lib/api/analytics/consumption/period";
 import { fetchConsumptionTopAgents } from "@app/lib/api/analytics/consumption/top_agents";
 import { fetchConsumptionTopApiKeys } from "@app/lib/api/analytics/consumption/top_api_keys";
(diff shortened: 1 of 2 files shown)


avervaet:
Closing: the merged single-query approach widens the terms candidate pool to span both periods, which can let a previous-period-only heavy hitter crowd out a real current-period top performer in the terms aggregation's shard_size selection. Current period needs to stay the unambiguous source of truth for this ranking, so reverting this optimization rather than shipping a probabilistic mitigation.

PR #30610, opened 2026-08-17
[front] fix: remove filtering on new analytics page header values

aubin-tchoi:
## Description

The header on the new analytics page changes when we change the filter.
As it's positioned above the filter it should remain the same.

This PR fixes that.

## Tests

- Tested locally.

## Risk

- Low.

## Deploy Plan

- Deploy front.


avervaet (approved):

avervaet:
nit: The filter prop is now unused by ConsumptionOverview. Could we remove it from the component props, destructuring, and its call site to avoid leaving a misleading API behind?

PR #30611, opened 2026-08-17
[front] chore: Make pptx/xlsx/docx skill import directive more agressive

id13:
## Description

When we retired skill "extend" in favor of skill in skills, we worded the backfill in a "best effort" matter, the model may or may not load it. 

For pptx/docs/xlsx, it's not enough. These skills are necessary when dealing with office use-cases in the sandbox.
Yet, over a sample of 1500 conversations, it dit it less than 1%.

This script updates the wording of these skills extension.

## Tests

N/A

## Risk

Low

## Deploy Plan

Merge
Dry run, see the results
If no problem , run it

### DIFF
--- a/front/migrations/20260814_rewrite_office_skill_customization_directive.ts
+++ b/front/migrations/20260814_rewrite_office_skill_customization_directive.ts
@@ -1,0 +1,156 @@
+import { Authenticator } from "@app/lib/auth";
+import { SkillConfigurationModel } from "@app/lib/models/skill";
+import { convertMarkdownToBlockHtml } from "@app/lib/reinforcement/skill_instructions_html";
+import { SkillResource } from "@app/lib/resources/skill/skill_resource";
+import { parseSkillTag } from "@app/lib/skills/format";
+import type { Logger } from "@app/logger/logger";
+import { makeScript } from "@app/scripts/helpers";
+import { runOnAllWorkspaces } from "@app/scripts/workspace_helpers";
+import { GROUP_KINDS, isAgentEditorGroupKind } from "@app/types/groups";
+import type { LightWorkspaceType } from "@app/types/user";
+import { Op } from "sequelize";
+
+const CUSTOMIZATION_LABEL = "This skill is a customization of";
+
+const OFFICE_SKILL_IDS = new Set(["docx", "pptx", "xlsx"]);
+
+const CUSTOMIZATION_REFERENCE_REGEX = new RegExp(
+  `${CUSTOMIZATION_LABEL}\\s*(<skill\\s+[^>]*?\\s*(?:/>|></skill>))`,
+  "g"
+);
+
+function rewriteCustomizationDirective(instructions: string): string | null {
+  let rewritten = false;
+
+  const next = instructions.replace(
+    CUSTOMIZATION_REFERENCE_REGEX,
+    (match, tag: string) => {
+      const skill = parseSkillTag(tag);
+
+      if (!skill || !OFFICE_SKILL_IDS.has(skill.id)) {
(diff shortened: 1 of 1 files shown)


front/migrations/20260814_rewrite_office_skill_customization_directive.ts
```diff
@@ -0,0 +1,221 @@
...
+async function rewriteWorkspaceOfficeSkillDirectives(
+  workspace: LightWorkspaceType,
+  { execute }: { execute: boolean },
+  logger: Logger
+): Promise<RewriteStats> {
+  const where: WhereOptions<SkillConfigurationModel> = {
+    workspaceId: workspace.id,
+    [Op.or]: [
+      { instructions: { [Op.like]: `%${CUSTOMIZATION_LABEL}%` } },
+      { instructionsHtml: { [Op.like]: `%${CUSTOMIZATION_LABEL}%` } },
```
aubin-tchoi: nit: would only check the non html instructions

front/migrations/20260814_rewrite_office_skill_customization_directive.ts
```diff
@@ -0,0 +1,221 @@
...
+
+async function rewriteWorkspaceOfficeSkillDirectives(
+  workspace: LightWorkspaceType,
+  { execute }: { execute: boolean },
+  logger: Logger
+): Promise<RewriteStats> {
+  const where: WhereOptions<SkillConfigurationModel> = {
+    workspaceId: workspace.id,
+    [Op.or]: [
+      { instructions: { [Op.like]: `%${CUSTOMIZATION_LABEL}%` } },
```
aubin-tchoi: OOC isn't the customization part always at the beginning?

front/migrations/20260814_rewrite_office_skill_customization_directive.ts
```diff
@@ -0,0 +1,221 @@
...
+
+  return rewritten ? next : null;
+}
+
+function planInstructionsRewrite(row: InstructionsRow): InstructionsRow | null {
+  const instructions = rewriteCustomizationDirective(row.instructions);
+  const instructionsHtml =
+    row.instructionsHtml === null
+      ? null
+      : rewriteCustomizationDirective(row.instructionsHtml);
```
aubin-tchoi: We have a helper to go from the markdown instructions to the html, it makes sure there are kept in sync

front/migrations/20260814_rewrite_office_skill_customization_directive.ts
```diff
@@ -0,0 +1,221 @@
...
+      { instructionsHtml: { [Op.like]: `%${CUSTOMIZATION_LABEL}%` } },
+    ],
+  };
+
+  const skills = await SkillConfigurationModel.findAll({
+    attributes: ["id", "instructions", "instructionsHtml"],
+    where,
+  });
+
+  const versions = await SkillVersionModel.findAll({
```
aubin-tchoi: We could also choose to create a new version btw, would be faster (filter on active skills, no need to backfill versions) and rollbackable

aubin-tchoi (approved):
LGTM

front/migrations/20260814_rewrite_office_skill_customization_directive.ts
id13: Well, not if users put something on top of it 😄

front/migrations/20260814_rewrite_office_skill_customization_directive.ts
id13: Yep, that's much better !

PR #30612, opened 2026-08-17
fix(front): knowledge picker dead-clicks on childless data source views

achilleburah:
## Summary
- Clicking a top-level data source view with zero root content nodes (e.g. a website with nothing crawled) navigated into a blank screen with no way back — now shows an empty state with a "Go back" action.
- Row clicks in the knowledge picker now fall back to toggling selection when there is nothing to navigate into, fixing the same silent dead-click on non-expandable leaf nodes deeper in the tree.

## Tests
- Opened Agent Builder → Add knowledge → a space → Websites, click a website with zero root pages: confirm the empty state + "Go back" button instead of a blank screen
- Clicked a website with children: confirm navigation still works unchanged
- Clicked a non-expandable leaf content node's label (not the checkbox): confirm it toggles selection

### Before 
image

### After
image

### DIFF
--- a/front/components/agent_builder/capabilities/knowledge/DataSourceList.tsx
+++ b/front/components/agent_builder/capabilities/knowledge/DataSourceList.tsx
@@ -350,17 +350,21 @@ export function DataSourceList({
           ? selectionState === "partial"
           : !hideCheckbox;
 
+        const handleRowClick = () => {
+          if (item.onClick) {
+            item.onClick();
+            return;
+          }
+          if (shouldShowCheckbox) {
+            void handleSelectionChange(item, selectionState !== true);
+          }
+        };
+
         return (
           <Fragment key={item.id}>
             <div
               className="flex cursor-pointer items-center justify-between rounded-md p-3 hover:bg-muted/60"
-              onClick={() => {
-                if (item.onClick) {
-                  item.onClick();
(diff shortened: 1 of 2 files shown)


front/components/agent_builder/capabilities/knowledge/DataSourceList.tsx
```diff
@@ -378,11 +378,24 @@ export function DataSourceList({
...
+            item.onClick();
+            return;
+          }
+          if (shouldShowCheckbox) {
+            void handleSelectionChange(
+              item,
+              selectionState === "partial" ? true : !selectionState
+            );
+          }
+        };
```
fabiencelier, line 361: can you please add a comment to clarify the intent here ?
Not obvious to me what this handleRowClick purpose is ^^

fabiencelier (approved):

front/components/agent_builder/capabilities/knowledge/DataSourceList.tsx
achilleburah: currently a click on a "last child" node, does not have any link to go to. So the click action is dead. 
so instead, we toggle the associated checkbox if there is one.

achilleburah: added a comment

PR #30617, opened 2026-08-17
Cut over workspace resource cache key

flvndvd:
## Description

This is the second step of the Workspace cache-key migration. The first PR https://github.com/dust-tt/dust/pull/30270 keeps the previous key as the read source while populating both keys. This PR cuts reads over to the new versioned key.

This must only be deployed after the first PR is fully deployed.

## What changes

Workspace reads now use the new key. Cache hits are copied to the previous key so old application instances and an immediate rollback can still read current data. Database loads continue to populate both keys, and application invalidation removes both.

The cached Workspace snapshot continues to store the raw provider list. Provider kill switches are applied after reading the snapshot, preventing temporary kill-switch state from being stored in the new cache.

Poke now looks up the key currently used for reads. Exact and bulk deletion remove both the new and previous keys so deleted entries cannot reappear from the rollback copy.

Took the opportunity to "productize" the cache migration by making the state is explicit about which key is read and when cache hits are copied. This keeps future resource migrations inside the cache layer.

## Rollout and rollback

To roll back, revert this PR. The previous key remains populated throughout the cutover, so the first migration state can resume reading it without a cache-wide flush.

## Risk

During this temporary phase, cache hits also write the previous key. This increases Redis writes for Workspace reads until the cleanup PR is deployed.

The main correctness risk is deleting or updating only one key. Both application invalidation and Poke deletion are covered to ensure both keys are removed together.

### DIFF
--- a/front-api/routes/poke/cache/index.ts
+++ b/front-api/routes/poke/cache/index.ts
@@ -25,7 +25,9 @@ app.route("/catalog", catalog);
 
 function resolveCacheKey(
   ctx: Context
-): { cacheKey: string } | { err: ReturnType<typeof apiError> } {
+):
+  | { cacheKey: string; cacheKeysToDelete: string[] }
+  | { err: ReturnType<typeof apiError> } {
   const resourceId = ctx.req.query("resourceId");
   const rawKey = ctx.req.query("rawKey");
   const params = ctx.req.query("params");
@@ -90,7 +92,10 @@ function resolveCacheKey(
     }
 
     try {
-      return { cacheKey: operations.buildKey(parsedParams) };
+      return {
+        cacheKey: operations.buildKey(parsedParams),
+        cacheKeysToDelete: operations.buildKeysToDelete(parsedParams),
+      };
     } catch {
       return {
         err: apiError(ctx, {
@@ -105,7 +110,7 @@ function resolveCacheKey(
   }
 
   if (isString(rawKey)) {
-    return { cacheKey: rawKey };
+    return { cacheKey: rawKey, cacheKeysToDelete: [rawKey] };
   }
 
   return {
@@ -188,7 +193,7 @@ app.delete("/", async (ctx): HandlerResult<DeletePokeCacheResponseBody> => {
   if ("err" in r) {
     return r.err;
   }
-  const { cacheKey } = r;
+  const { cacheKey, cacheKeysToDelete } = r;
 
   const redisInstance = ctx.req.query("redisInstance");
   if (redisInstance !== "cache" && redisInstance !== "stream") {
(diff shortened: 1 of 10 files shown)


front-api/routes/poke/cache/index.ts
```diff
@@ -197,11 +202,11 @@ app.delete("/", async (ctx): HandlerResult<DeletePokeCacheResponseBody> => {
   const runFn = redisInstance === "cache" ? runOnRedisCache : runOnRedis;
 
   await runFn({ origin: "poke_cache_invalidation" }, async (client) => {
-    await client.del(cacheKey);
+    await Promise.all(cacheKeysToDelete.map((key) => client.del(key)));
```
flvndvd, line 213: Temporary, used on a bounded array.

front-api/routes/poke/cache/index.ts
```diff
@@ -23,7 +23,9 @@ app.route("/catalog", catalog);
 
 function resolveCacheKey(
   ctx: Context
-): { cacheKey: string } | { err: ReturnType<typeof apiError> } {
+):
+  | { cacheKey: string; cacheKeysToDelete: string[] }
+  | { err: ReturnType<typeof apiError> } {
```
aubin-tchoi, line 30: nit: should be a Result (but seems to predate the PR)

front/lib/resources/workspace_resource.ts
```diff
@@ -199,11 +199,14 @@ export class WorkspaceResource extends BaseResource<WorkspaceModel> {
...
-      key: (workspaceId: string) => `workspace:v2:${workspaceId}`,
-      keyPattern: "workspace:v2:*",
-      mirrorToCanonicalOnHit: false,
+    migration: {
+      previousKey: {
+        cacheId: "_fetchByIdUncached",
+        key: (workspaceId: string) => `workspace:v2:${workspaceId}`,
+        keyPattern: "workspace:v2:*",
+      },
+      readFrom: "new",
```
aubin-tchoi, line 206: nice to be able to have a more in-depth way to rollout cache key changes (by setting "previous" first, then wait a bit and putting "new"), but in most cases when we change the key we also want to invalidate the cache no? for instance when we add a field in the object to cache
in what situations would `readFrom` be used?

aubin-tchoi (approved):
LGTM

front/lib/resources/workspace_resource.ts
flvndvd: yeah we should have one canonical way to burst a cache key in a safe way (in both directions rollout/rollback).

adrsimon (approved):

PR #30618, opened 2026-08-17
feat: triggers table on the automations analytics page

adrsimon:
## Description

Fills the body of `/analytics/automations` with the triggers that consumed over the period picked in the header: name, agent, editor, type, run count and credits. Rows are ranked by credits in Elasticsearch, so a trigger that did not run over the period does not show up.

Pagination is server-side, same as the consumption attribution table: the client sends `limit`/`offset` and gets `totalCount` back. Ranking and paging both happen in the terms aggregation, so neither the ES work nor the Postgres read grows with the number of triggers the workspace holds. `totalCount` comes from a cardinality aggregation and is approximate past 40k distinct triggers, like the consumption page.

Admin-only for now. Not in this PR: filters, the row-expand chevron, and the enable/disable toggle.

Capture d’écran 2026-08-17 à 13 07 24

## Tests

`front-api/routes/w/[wId]/analytics/automations/triggers.test.ts` covers the endpoint.
Checked the page manually against a workspace with schedule and webhook triggers, including paging through and switching the period.

## Risk

Low.

## Deploy Plan

Deploy front.


front/lib/api/analytics/automations/triggers.ts
```diff
@@ -0,0 +1,251 @@
...
+    period,
+    limit,
+    offset,
+  }: {
+    period: ConsumptionPeriod;
+    limit: number;
+    offset: number;
+  }
+): Promise<Result<AutomationTriggers, ElasticsearchError>> {
+  const triggers = (await TriggerResource.listByWorkspace(auth)).map(
```
aubin-tchoi: nit: do we need the `toJson` here?

front/lib/api/analytics/automations/triggers.ts
```diff
@@ -0,0 +1,251 @@
...
+
+  const views = await WebhookSourcesViewResource.fetchByIds(
+    auth,
+    webhookSourceViewIds
+  );
+
+  return new Map(
+    views.map((view) => {
+      const { sId, customName, icon } = view.toJSON();
+      return [sId, { name: customName, icon }];
```
aubin-tchoi: nit: you could return an array of `WebhookSourcesViewResource`, let the caller build the map, add a getter `name` that does `this.customName ?? this.webhookSource.name` and use it directly in the page.map callback (all of this to avoid the `toJSON`)

front/components/workspace/analytics/automations/AutomationsTriggersTable.tsx
```diff
@@ -0,0 +1,224 @@
...
+  triggers,
+  totalCount,
+  pagination,
+  setPagination,
+}: TriggersTableBodyProps) {
+  if (isLoading) {
+    return (
+      <div className="flex h-48 items-center justify-center">
+        <Spinner size="lg" />
+      </div>
```
aubin-tchoi: we could do a skeleton here since we know the target layout

front/hooks/useAutomationsTriggers.ts
```diff
@@ -0,0 +1,45 @@
...
+    AutomationTriggersBody,
+    GetAutomationTriggersResponse
+  >({ url, body, disabled });
+
+  return {
+    triggers: data?.triggers ?? emptyArray<AutomationTriggerRow>(),
+    totalCount: data?.totalCount ?? 0,
+    isTriggersLoading: !error && isLoading,
+    isTriggersError: error,
+    isTriggersValidating: isValidating,
```
aubin-tchoi, line 43: fyi we'll need to add an es-backed pagination to get the next pages

front/lib/api/analytics/automations/triggers.ts
```diff
@@ -0,0 +1,251 @@
...
+      by_trigger: {
+        terms: {
+          field: TRIGGER_ID_FIELD,
+          include: triggerIds,
+          size: triggerIds.length,
+        },
+        aggs: {
+          [CREDIT_AGG]: { sum: { field: CREDIT_MICRO_FIELD } },
+          // One trigger run is one conversation.
+          [RUNS_AGG]: { cardinality: { field: CONVERSATION_ID_FIELD } },
```
aubin-tchoi, line 119: fyi this count will be approximate above a certain value

front/hooks/useAutomationsTriggers.ts
```diff
@@ -0,0 +1,45 @@
+import { useConsumptionQuery } from "@app/hooks/useConsumptionQuery";
+import type { ConsumptionPeriodSelection } from "@app/lib/analytics/consumption_period";
+import type { AutomationTriggersBody } from "@app/lib/api/analytics/automations/schema";
+import type {
+  AutomationTriggerRow,
+  GetAutomationTriggersResponse,
+} from "@app/lib/api/analytics/automations/triggers";
+import { DEFAULT_CONSUMPTION_PERIOD_DAYS } from "@app/lib/api/analytics/consumption/schema";
```
aubin-tchoi: let's move `DEFAULT_CONSUMPTION_PERIOD_DAYS` to avoid importing from `lib/api` here

front/components/workspace/analytics/automations/AutomationsTriggersTable.tsx
```diff
@@ -0,0 +1,224 @@
...
+    ),
+  },
+  {
+    id: "runCount",
+    accessorKey: "runCount",
+    header: "Runs",
+    meta: { className: "w-24", headerAlign: "right" },
+    cell: (info) => (
+      <DataTable.BasicCellContent
+        className="justify-end text-right tabular-nums"
```
aubin-tchoi, line 122: sure about the `tabular-nums`? useful in some cases but I find it to look a bit ugly sometimes

aubin-tchoi (approved):
LGTM

front/lib/api/analytics/automations/triggers.ts
```diff
@@ -0,0 +1,251 @@
...
+  {
+    period,
+    limit,
+    offset,
+  }: {
+    period: ConsumptionPeriod;
+    limit: number;
+    offset: number;
+  }
+): Promise<Result<AutomationTriggers, ElasticsearchError>> {
```
avervaet, line 191: The current implementation loads every trigger, aggregates every trigger ID in Elasticsearch, and only then applies pagination here in the slice later on. Maybe we can find a way to alleviate the pressure on ES for workspaces with a lot of triggers ?

front/lib/api/analytics/automations/triggers.ts
avervaet, line 119: nit: this can be an approximative run_count we could use Postgres for the real totalCount

But maybe we don't care and want to avoid db calls

avervaet (approved):

avervaet:
Additional: seems like page name in the navigation tab should be changed to Programmatic Usage and the icon changed as well

adrsimon:
> Additional: seems like page name in the navigation tab should be changed to Programmatic Usage and the icon changed as well

The page is named automations in the Design and it's not all programmatic usage but really just Triggers from Dust, so i disagree with this

Will fix the rest!

PR #30620, opened 2026-08-17
Polish analytics consumption UI

flvndvd:
## Description

This PR polishes two things on the new analytics consumption page:
1. the size of the button switch
2. the order of the filters, to we first show alphabetically sorted active facets

### DIFF
--- a/front/components/workspace/analytics/usageFilter.ts
+++ b/front/components/workspace/analytics/usageFilter.ts
@@ -130,6 +130,15 @@ export interface UsageFilterGroup {
   memberIds: string[];
 }
 
+export function sortUsageFilterOptionsByAvailability<
+  T extends UsageFilterOption,
+>(options: T[]): T[] {
+  return [
+    ...options.filter((option) => !option.disabled),
+    ...options.filter((option) => option.disabled),
+  ];
+}
+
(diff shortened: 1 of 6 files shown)


front/components/workspace/analytics/usageFilter.ts
```diff
@@ -130,6 +130,15 @@ export interface UsageFilterGroup {
   memberIds: string[];
 }
 
+export function sortUsageFilterOptionsByAvailability<
+  T extends UsageFilterOption,
+>(options: T[]): T[] {
+  return [
+    ...options.filter((option) => !option.disabled),
+    ...options.filter((option) => option.disabled),
+  ];
```
aubin-tchoi: no sort by name here?

aubin-tchoi (approved):
LGTM

front/components/workspace/analytics/usageFilter.ts
flvndvd: it's sorted by name above, but fair, let me add it here as well.

flvndvd: Done, moved to the API side of things so the pagination works as expected.

adrsimon (approved):

PR #30623, opened 2026-08-17
Handle terminal Snowflake and Microsoft connector errors

fontanierh:
## Description

- Skip Snowflake databases that return `003030` after a share is revoked, matching the existing per-database handling for `002043`.
- Treat Snowflake `090693` expired listings and Microsoft Graph's missing SharePoint license response as terminal provider configuration errors. The existing activity interceptor marks the sync failed, pauses the connector, and stops its workflow instead of retrying forever.

## Tests

Added focused tests for the three provider responses and for the shared failed-sync, pause, and stop behavior.

## Risk

A revoked Snowflake database is removed from the sync tree and its stale content can be garbage collected, matching the existing inaccessible-database behavior. Terminal configuration errors pause the whole connector until it is manually resumed after the customer fixes the provider configuration. Rollback is a normal revert; connectors paused before rollback remain paused until resumed.

## Deploy Plan

Deploy Connectors normally in both regions. Verify connector 32646 completes while skipping the revoked database, and connectors 13993 and 49139 become failed and paused with no further Temporal attempts. Monitor the three workflow IDs and connector error logs after rollout.

### DIFF
--- a/connectors/src/connectors/microsoft/temporal/cast_known_errors.test.ts
+++ b/connectors/src/connectors/microsoft/temporal/cast_known_errors.test.ts
@@ -1,0 +1,26 @@
+import { ThirdPartyConfigurationError } from "@connectors/lib/error";
+import { GraphError } from "@microsoft/microsoft-graph-client";
+import type {
+  ActivityExecuteInput,
+  ActivityInboundCallsInterceptor,
+  Next,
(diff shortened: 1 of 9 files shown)


aubin-tchoi (commented):
LGTM

PR #30625, opened 2026-08-17
Flav/conversation credits subagents

flvndvd:
## Description

This PR tweaks the conversation credits breakdown to capture all sub agents cost alongside top level tool calls.

Sadly our data model is missing easy path transversal, this will be rework later by @tdraier, in the meantime, this PR leverages the bounded depth so we can issue a minimal number of queries. We could not use ES as we are missing a few fields that we will add once data model is done being reworked.

image


aubin-tchoi (approved):
LGTM

PR #30630, opened 2026-08-17
Track total (own + sub-agent) message cost in totalCostCredits

tdraier:
## Description

Adds `agent_messages.totalCostCredits`: a message's own cost **plus** the cost of every `run_agent` sub-agent it spawned (recursively). Maintained at finalize by rolling the existing cost delta up to the message and each `run_agent` ancestor (order-independent, additive). Billing/rate-limit counters keep using `costCredits` unchanged; handovers are excluded (billed in the parent conversation).

### Why a new column instead of reusing `costCredits`
`costCredits` is the message's **own** cost and the source of truth fed into the fair-use rate limiter and per-user spend cap (as a delta). Folding sub-agent costs into it would (1) double-count spend, since each sub-agent already records its own cost into those same counters, and (2) break the recursive rollup, which sums descendants' `costCredits` — a subtree total on each node would count leaves multiple times. Keeping own-cost and total-cost as separate columns avoids both.

Read-path swap (drop the recursive read-time rollup in `messages.ts`) + historical backfill to follow in a second PR.

## Risk

Low - new column / 

## Tests

locally, column filled with total

## Deploy Plan

run migration
deploy front

### DIFF
--- a/front/lib/api/assistant/credit_cost.ts
+++ b/front/lib/api/assistant/credit_cost.ts
@@ -1,6 +1,7 @@
 import type { InternalMCPServerNameType } from "@app/lib/actions/mcp_internal_actions/constants";
 import type { ToolExecutionStatus } from "@app/lib/actions/statuses";
 import { getToolNameFromFunctionCallName } from "@app/lib/actions/tool_display_labels";
+import { listAgenticAncestors } from "@app/lib/api/assistant/conversation/agentic_ancestors";
 import { makeFairUseAwuCreditsRateLimitKeyForUser } from "@app/lib/api/assistant/rate_limits";
 import { recordProgrammaticSpendLimitUsage } from "@app/lib/api/credits/programmatic_usage_limit";
 import { recordApiKeySpendLimitUsage } from "@app/lib/api/keys/spend_limit";
@@ -25,6 +26,11 @@ import {
 import logger from "@app/logger/logger";
 import type { UserMessageOrigin } from "@app/types/assistant/conversation";
 import { AGENT_MESSAGE_STATUSES_TO_TRACK } from "@app/types/assistant/conversation";
+
+// Bounds the application-side walk up the run_agent ancestor chain when rolling
+// a message's cost into its parents' `totalCostCredits`. run_agent nesting is
+// shallow in practice; this caps the number of per-level lookups per finalize.
+const MAX_RUN_AGENT_ANCESTOR_DEPTH = 10;
 
 interface CreditActionMinimalInput {
   toolName: string;
@@ -82,8 +88,9 @@ export async function computeAndStoreAgentMessageCredits(
   auth: Authenticator,
   {
     agentMessageId,
+    conversationId,
     dustRunIds,
-  }: { agentMessageId: string; dustRunIds?: string[] }
+  }: { agentMessageId: string; conversationId: string; dustRunIds?: string[] }
 ): Promise<number | null> {
   const creditContext =
     await ConversationResource.fetchAgentMessageCreditContext(auth, {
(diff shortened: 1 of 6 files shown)


front/lib/resources/conversation_resource.ts
```diff
@@ -911,6 +911,83 @@ export class ConversationResource extends BaseResource<ConversationModel> {
...
+
+    const workspaceId = auth.getNonNullableWorkspace().id;
+
+    // `agent_message_sid` is a `messages.sId` (the agent message's identifier is
+    // carried on its wrapping message row; `agent_messages` has no `sId`).
+    // `agenticOriginMessageId` likewise references the parent agent message's
+    // `messages.sId`. The final UPDATE resolves each sId back to its
+    // `agent_messages` row via `messages.agentMessageId`.
+    const query = `
+      WITH RECURSIVE ancestors AS (
```
flvndvd: Clear no-go on the Recursive CTE, I've introduced two ways to get this keeping the loop on the application side, to avoid freezing our DB.

tdraier: ok, replaced by a flat bulk update incrementing the total cost

front/lib/api/assistant/credit_cost.ts
```diff
@@ -32,6 +33,11 @@ import type {
 import type { UserMessageOrigin } from "@app/types/assistant/conversation";
 import { AGENT_MESSAGE_STATUSES_TO_TRACK } from "@app/types/assistant/conversation";
 
+// Bounds the application-side walk up the run_agent ancestor chain when rolling
+// a message's cost into its parents' `totalCostCredits`. run_agent nesting is
+// shallow in practice; this caps the number of per-level lookups per finalize.
+const MAX_RUN_AGENT_ANCESTOR_DEPTH = 10;
```
flvndvd: Hold on a second, don't we already have a const for this? Like this one https://github.com/dust-tt/dust/blob/1ff265ce65bcbf2abcddb8775b0e2b2b28266d14/front/lib/api/assistant/conversation/constants.ts#L3?

front/lib/resources/conversation_resource.ts
```diff
@@ -911,6 +911,51 @@ export class ConversationResource extends BaseResource<ConversationModel> {
     );
   }
 
+  /**
```
flvndvd, line 933: Can we DRY those comments, not very relevant the previous recursive state, right?

front/lib/resources/conversation_resource.ts
```diff
@@ -911,6 +911,51 @@ export class ConversationResource extends BaseResource<ConversationModel> {
...
+
+    const workspaceId = auth.getNonNullableWorkspace().id;
+
+    // Flat bulk update keyed by `messages.sId` (no recursion): the agent
+    // message identifier is carried on its wrapping message row, so we resolve
+    // each sId back to its `agent_messages` row via `messages.agentMessageId`.
+    // NULL-safe accumulate via COALESCE since the column defaults to NULL.
+    const query = `
+      UPDATE agent_messages am
+      SET "totalCostCredits" = COALESCE(am."totalCostCredits", 0) + :delta
```
flvndvd: This incremental roll-up is not idempotent. `costCredits` is overwritten separately, while `totalCostCredits` is incremented afterward. A failure between the two loses the roll-up. Concurrent retries can apply the same delta twice. Since finalization is a Temporal activity, this is a likely to happen, right?

front/lib/resources/conversation_resource.ts
```diff
@@ -3267,14 +3316,17 @@ export class ConversationResource extends BaseResource<ConversationModel> {
...
         },
       ],
     });
 
-    const agenticOriginMessageId =
-      parentMessage?.userMessage?.agenticOriginMessageId;
-    if (!agenticOriginMessageId) {
+    const originUserMessage = parentMessage?.userMessage;
+    // `agenticMessageType` and `agenticOriginMessageId` are always set together
+    // (enforced by a model validation hook), so both are present here or neither.
```
flvndvd: ✂️

tdraier:

- #30630 ◀
    - #30885

PR #29763, opened 2026-07-30
[front] Per-workspace flip: enforce space access from group_permissions (#9480)

tdraier:
## Description

Final step of the space-permissions migration ([#9477](https://github.com/dust-tt/tasks/issues/9477)) — sub-issue [#9480](https://github.com/dust-tt/tasks/issues/9480): **flip space access to the `group_permissions` table**.

Stacked on #29639 (shadow). Once shadow-mode confirmed parity, this makes the table the enforced source of truth for space group access. **No feature flag** — the table is written directly at every mutation site (#29619) and backfilled for existing spaces, so it is authoritative everywhere.

### What it does

- **`SpaceResource.getAccessControlLists(auth)`** now returns a single ACL: the (unchanged) per-kind **role** rules + **`auth.getGroupPermissions("space", id)`** for groups — i.e. group access resolved from the `group_permissions` table instead of the legacy inline `group_vaults`.
- **`canRead` / `canWrite` / `canAdministrate`** collapse to `auth.hasPermission(verb, this)`. No `servedAcls` / `governanceAcls` branching.
- `spaceGroupGrants()` (the `group_vaults`-derived grants) stays — it is what `writeGroupPermissions` keeps the table in sync from — but it no longer feeds enforcement.

### Removed / kept
- **Removed:** the `group_permissions_enforce` flag, the `enforced` bit on `GroupPermissions` (+ `isEnforced` / `isGroupPermissionsEnforced` / the resolve-time preload), `servedAcls`, `governanceAcls`, the space `shadowCompareSpacePermission` wiring, and `reconcileGroupPermissions` (the backfill alias — the migration now calls `writeGroupPermissions` directly).
- **Kept:** `Authenticator.shadowComparePermission` and the `group_permissions_shadow` flag, for reuse on other resources.

### Behavior
Group access to every space is now served from the table. `getGroupPermissions` is caller-scoped; the membership check that the composed ACL relied on was already caller-relative, so the served decision is preserved for a correctly-populated table. Admin access flows through the unchanged role path.

## Risk
Medium — this is the step that changes the *served* decision. Correctness depends on the table being populated (direct writes at all mutation sites + the #29619 backfill). `AccessControlList.workspaceId` still gates the role path.

## Tests
Space checks now assert access via the table; `tsgo` clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #30633, opened 2026-08-17
[front] Store resolved verbs (not group refs) in GroupPermissions

tdraier:
## Description

The Authenticator's `GroupPermissions` kept a per-group mask map — `resourceType → resourceId → groupId → mask` — and `hasPermissionForAcl` re-intersected those groups against the caller's own groups on every check. We only ever load the caller's groups, so that intersection was a no-op for governance grants. This folds the groups away at construction and stores just the caller's resolved verb mask per resource.

Based on `main` (behind the space/skill governance rollout, ahead of the space flip #29763). No user-facing behavior change: the space governance ACL is still shadow-compare only, and the legacy served/conversation paths are untouched.

**Storage** — `space: { 12: { 7: 0b0011, 9: 0b0001 } }` becomes `space: { 12: 0b0011 }`:
- `GroupPermissions` storage, `fromGrants`, `toJSON`, `GroupPermissionsJSON`, `toWorkspacePermissions`, `toString` lose the `groupId` level. `forResource` → `resolvedVerbsForResource` (direct mask lookup), surfaced as `Authenticator.getGrantedVerbs`; `getGroupPermissions` removed.

**ACL shape** — governance-sourced ACLs carry resolved verbs, not group refs:
- `AccessControlList` gains an optional `grantedVerbs`; `roles`/`groups` become optional (additive OR; absent = fail-closed).
- `hasPermissionForAcl` checks `grantedVerbs` directly (no membership step) and keeps the legacy `groups` intersection for the cross-space conversation path (still legacy until the space flip; still uses `_groupModelIds`).
- Serving `grantedVerbs`: the space shadow candidate (`governanceAcls`), the skill served ACL, and the workspace-wide capability check. The served legacy space ACL and the conversation/`permission_utils` path are unchanged.

**Temporal-safe serialization** — `fromJSON` reads **both** shapes. The shadow stage (already in prod) serializes the 3-level shape into Temporal payloads, so the new reader OR-folds the legacy per-group masks. `TODO(governance)`: drop the legacy arm once those payloads drain.

`_groupModelIds` stays for the legacy `groups` ACL path and `hasGroup*`.

## Tests

- `npx tsgo --noEmit` clean; `biome check` clean.
- 420 existing tests passing (registry, space/skill/conversation resources, conversation permissions, spaces JSON round-trip, auth.workspace_permission, auth.fromjson, governance_seeding, shadow) — no behavior change.
- Added 3 unit tests for the dual-read `fromJSON` (current shape, legacy per-group shape, round-trip).
- Reframed the two system-key group-scoping tests onto resolved verbs: the out-of-scope group's grant lands on a different agent so its absence is observable now that group ids are gone.

## Risk

Low.
- No user-facing behavior change: the only governance ACL that changes shape is the space **shadow candidate** (fire-and-forget compare) and the skill served ACL (gated by `isLegacyAclsEnabled`, decision unchanged). Legacy served + conversation paths untouched. Fail-closed: an absent ACL source denies.
- **Serialized-shape transition, handled.** The live shadow build writes the 3-level `GroupPermissionsJSON` into Temporal payloads; `fromJSON` here reads both shapes, so new workers parse in-flight payloads correctly. Residual: during the rolling deploy an old worker could read a new 2-level payload and misparse → fail-closed (never over-grants), transient.

## Deploy Plan

Standard deploy, no migration or flag. Deploy **before** the space flip (#29763) so the simpler shape lands and bakes first; the flip then serves `grantedVerbs` directly. Fresh auths rebuild `GroupPermissions` from the DB on every `resolvePermissions`.

### DIFF
--- a/front/lib/auth.ts
+++ b/front/lib/auth.ts
@@ -60,7 +60,6 @@ import type { PlanType, SubscriptionType } from "@app/types/plan";
 import type { ProvidersHealth } from "@app/types/provider_credential";
 import type {
   AccessControlList,
-  GroupGrant,
   WithAccessControl,
 } from "@app/types/resource_permissions";
 import { isDevelopment } from "@app/types/shared/env";
@@ -1260,7 +1259,7 @@ export class Authenticator {
 
     return this.hasPermissionForAcl(verb, {
       roles: [{ role: "admin", permissions: [verb] }],
-      groups: this.getGroupPermissions(resourceType, WHOLE_TYPE_RESOURCE_ID),
+      grantedVerbs: this.getGrantedVerbs(resourceType, WHOLE_TYPE_RESOURCE_ID),
       workspaceId: workspace.id,
     });
   }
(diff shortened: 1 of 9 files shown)


tdraier:

- #30633 ◀

fabiencelier (approved):
LGTM
should be more compact without no loss of data

PR #30635, opened 2026-08-17
[Activation] Exclude archived learning spaces

frankaloia:
## Summary
- Treat archived Pods as inactive across For You, activation scheduling, and notification lookups
- Done by filtering via an inner join in resource class

## Testing
- Archived pod, validated not treating as active

## Risk
Low (solving existing bug)

## Deploy
1. Deploy front

### DIFF
--- a/front/lib/resources/activation_pod_resource.ts
+++ b/front/lib/resources/activation_pod_resource.ts
@@ -3,7 +3,6 @@ import { ActivationPodModel } from "@app/lib/models/activation/activation_pod";
 import { BaseResource } from "@app/lib/resources/base_resource";
 import type { SpaceResource } from "@app/lib/resources/space_resource";
 import { ProjectMetadataModel } from "@app/lib/resources/storage/models/project_metadata";
-import { SpaceModel } from "@app/lib/resources/storage/models/spaces";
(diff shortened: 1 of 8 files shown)


front/lib/resources/activation_pod_resource.ts
```diff
@@ -65,7 +67,29 @@ export class ActivationPodResource extends BaseResource<ActivationPodModel> {
...
+            {
+              model: ProjectMetadataModel,
+              as: "projectMetadata",
+              attributes: [],
+              required: true,
+              where: { archivedAt: null },
+            },
+          ],
+        },
+      ],
```
id13, line 80: ProjectMetadataModel has a spaceId col, so you can join directly on it and skip the Space join entirely

id13 (approved):
LGTM, I think we have one more join than needed

front/lib/resources/activation_pod_resource.ts
frankaloia: Addressed in 49011795b9: the archived filter now joins `ActivationPodModel` directly to `ProjectMetadataModel` on `spaceId`, removing the intermediate `SpaceModel` join.

PR #30636, opened 2026-08-17
fix: Explore header stays sticky through the attribution section

avervaet:
## Description

Fixes the sticky `Explore` header on the Analytics → Consumption page. Previously, the sticky header and chart were wrapped in a containing block that ended before the cost attribution table, so the header stopped sticking once the attribution section entered the scroll region.

## Tests

Tested locally

## Risk

Low risk and limited to the layout structure of the Analytics Consumption page. 

## Deploy Plan

- Deploy front


aubin-tchoi (approved):
LGTM

PR #30637, opened 2026-08-17
[front] enh: add inline references to Poke skill suggestions

aubin-tchoi:
## Description

Follow-up to #30632 

This PR allows adding skills and tools to the skill suggestions pushed from Poke.
Replaces a plain input for the skill instructions field with a rich skill editor + adds a Poke-backed picker that inserts tools and skills as inline references at the cursor.

image

## Tests

- Added tests.
- Tested locally.

## Risk

- Low. Poke only.

## Deploy Plan

- Deploy front.

### DIFF
--- a/front/components/poke/skills/CreateSkillSuggestionSheet.tsx
+++ b/front/components/poke/skills/CreateSkillSuggestionSheet.tsx
@@ -1,15 +1,49 @@
+import { editorVariants } from "@app/components/editor/editorStyles";
+import {
+  isSkillSlashCommand,
+  isToolSlashCommand,
+  SELECT_SKILL_SLASH_COMMAND_ACTION,
+  SELECT_TOOL_SLASH_COMMAND_ACTION,
+} from "@app/components/editor/extensions/shared/SlashCommandCapabilitiesItems";
+import { buildCapabilitySlashCommandItems } from "@app/components/editor/extensions/shared/slash_suggestion/buildSlashCommandItems";
+import type { SlashCommand } from "@app/components/editor/extensions/shared/slash_suggestion/SlashCommandDropdown";
+import {
+  SkillInstructionsEditorContent,
+  useSkillInstructionsEditor,
+} from "@app/components/editor/SkillInstructionsEditor";
+import {
+  getMcpServerViewDisplayName,
+  isToolWithKnowledge,
+} from "@app/lib/actions/mcp_helper";
+import { getMCPServerRequirements } from "@app/lib/actions/mcp_internal_actions/input_configuration";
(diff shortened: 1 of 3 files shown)


avervaet:
Could you add some screenshots on the description so its easier to see what existed and what was implemented

front/components/poke/skills/CreateSkillSuggestionSheet.tsx
```diff
@@ -67,18 +198,57 @@ export function CreateSkillSuggestionSheet({
...
 
+  const handleInsertCapability = (item: SlashCommand) => {
+    if (!editor || editor.isDestroyed) {
+      return;
+    }
+
+    if (isSkillSlashCommand(item)) {
+      editor
+        .chain()
+        .focus()
```
avervaet, line 229: nit: We could extract the shared skill/tool insertion used here and in `SkillInstructionsEditor`’s `onSelectRef`?  
The only difference is deleting `range`; could an optional range parameter keep these paths from drifting?

front/components/poke/skills/CreateSkillSuggestionSheet.tsx
```diff
@@ -47,13 +84,107 @@ export function CreateSkillSuggestionSheet({
...
+      owner,
+      disabled: !show,
+      globalSpaceOnly: true,
+    });
+
+  const activeSkills = useMemo(
+    () => skills.filter((skill) => skill.status === "active"),
+    [skills]
+  );
+
```
avervaet, line 104: proposal: We could we move this Poke capability filtering/labeling into a small hook or shared helper?  
It mirrors `getSkillBuilderSlashCommandTools`; would keeping this component focused on rendering make the picker easier to test?

avervaet (approved):
Didn't check the UI but code LGTM

front/components/poke/skills/CreateSkillSuggestionSheet.tsx
aubin-tchoi: Not the same types in input, which makes the shareable part quite small

aubin-tchoi: I'm fine with the poke path breaking often but really don't want to make it more complex to iterate on the non poke path
We'll probably revamp entirely this poke feature before the next time it breaks tbh

front/components/poke/skills/CreateSkillSuggestionSheet.tsx
aubin-tchoi: the useful part to share is `insertSkillNode` and `insertToolNode`
the chain.focus.run is a bit boilerplate but not really worth factorizing IMHO

PR #30640, opened 2026-08-17
Restrict default agent picker to private channels only returned by Slack API

frankaloia:
## Summary
- The default-agent Slack picker was showing every private channel Dust had joined (when `index_private_slack_channel` is on), including channels the current admin is not in.
- It now shows public Dust-write channels plus private channels this admin and the Dust bot both belong to.

## Testing
* New unit tests, can't test fully locally because i don't have a multi-user Slack workspace

## Risk
Low

## Deploy
1. Deploy front


smb2268 (approved):
LGTM

PR #30641, opened 2026-08-17
[front] Rate-limiter backup for the per-API-key spend cap

tdraier:
## Description

Adds a Redis fixed-window backup for the Metronome **per-API-key spend cap**, mirroring the per-user spend-cap backup shipped previously. It records AWU usage per calling API key into a fixed-window counter over the contract billing cycle, and enforces it synchronously at message send alongside the existing Metronome-driven `isApiKeyCapped`.

First PR of a 3-PR stack backing the pool-level Metronome spend caps with Redis rate-limiters (per-API-key → programmatic → workspace usage cap).

- Gated behind the **existing** `enforce_user_spend_limit_rate_cap` flag (reused for all spend-cap backups; description broadened) — one flag flip enables the per-user, per-key, programmatic and workspace backups together.
- Counter keyed by key model id; threshold is the key's `monthlyCapAwuCredits`.
- Lazy-seed from Elasticsearch (summed by `api_key_name`) on a Redis miss, so flipping the flag mid-cycle backfills the cycle-to-date total. Seeding by name also bridges key rotation.
- Extends the existing `resync-spend-limit-counters` poke plugin to also resync the per-key counters (one plugin resyncs all spend-cap counters).
- Promotes `resolveSpendLimitCycleBounds` to `lib/spend_limits/cycle.ts` so the per-user and per-key backups share one cycle-window resolver.

## Tests

- No new automated tests in this PR — the enforcement path is internal (message-send), not an endpoint. Can add counter/seed coverage to `front/lib/api/keys/spend_limit.test.ts` if desired.
- Intended manual validation: enable the flag on a dust-internal workspace, run the `resync-spend-limit-counters` poke plugin, and confirm the per-key counter matches ES `Consumed (ES)` and that a key at/over its cap is blocked with the existing 429.

## Risk

- **Low, fail-open by design.** Every failure mode (no cap, no billing period, Redis read error, missing key) returns "not reached" and allows the message.
- **Off by default** — behind `enforce_user_spend_limit_rate_cap` (`dust_only`). No behavior change when the flag is off (nothing recorded, nothing enforced).
- When on, this is a *backup* to the Metronome per-key cap already enforced today (`isApiKeyCapped`), OR-ed with it — it can only add blocks the Metronome path would also eventually apply, returning the same existing 429 `rate_limit_error`.
- Shared refactor (`resolveSpendLimitCycleBounds` moved to `cycle.ts`) is a pure move; the per-user path is unchanged behaviorally.
- Safe to roll back (revert): the flag stays off elsewhere and the counter keys are self-expiring.

## Deploy Plan

1. Merge; deploy `front` normally. No migration.
2. Flag remains `dust_only` and off — no runtime effect on merge.
3. To validate: enable `enforce_user_spend_limit_rate_cap` on a dust-internal workspace, then run the `resync-spend-limit-counters` poke plugin to seed counters from ES for the current cycle.
4. Compare the per-key counter against Metronome for a cycle before relying on the Redis path to block.

### DIFF
--- a/front/lib/api/assistant/credit_cost.ts
+++ b/front/lib/api/assistant/credit_cost.ts
@@ -4,6 +4,7 @@ import { isToolExecutionStatusBillable } from "@app/lib/actions/statuses";
 import { getToolNameFromFunctionCallName } from "@app/lib/actions/tool_display_labels";
 import { makeFairUseAwuCreditsRateLimitKeyForUser } from "@app/lib/api/assistant/rate_limits";
 import { searchAnalytics } from "@app/lib/api/elasticsearch";
+import { recordApiKeySpendLimitUsage } from "@app/lib/api/keys/spend_limit";
 import type { ToolCostCategory } from "@app/lib/api/mcp";
 import { isProgrammaticUsage } from "@app/lib/api/programmatic_usage/tracking";
 import { recordUserSpendLimitUsage } from "@app/lib/api/users/spend_limit";
@@ -326,13 +327,7 @@ export async function computeAndStoreAgentMessageCredits(
   }
 
   // Record against the per-user spend-cap backup (Redis fixed-window counter
-  // over the contract billing cycle). Independent of the plan-level fair-use
-  // limiter above. Gated behind the same feature flag as enforcement so the
-  // counter only accrues where the backup is active; enforcement happens
-  // pre-message in `checkMessagesLimit`.
-  // The cycle override keeps the counter on the same window
-  // enforcement reads: the contract billing period on credit-priced plans, the
-  // UTC calendar month elsewhere (no contract to anchor on).
+  // over the contract billing cycle).
   if (user && recordedCostDelta > 0) {
     const featureFlags = await getFeatureFlags(auth);
     if (featureFlags.includes("enforce_user_spend_limit_rate_cap")) {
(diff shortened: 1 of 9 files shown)


tdraier:

- #30641 ◀
    - #30654
        - #30655

front/lib/api/assistant/credit_cost.ts
```diff
@@ -344,5 +345,20 @@ export async function computeAndStoreAgentMessageCredits(
     }
   }
 
+  // Record against the per-API-key spend-cap backup (Redis fixed-window counter
+  // over the contract billing cycle), for calls authenticated with an API key.
```
avervaet: Not a big fan of large comments with specific reference to a part of the code, as it tends to age badly.

front/lib/api/keys/spend_limit.ts
```diff
@@ -320,3 +329,124 @@ export async function backfillApiKeyCreditCapsForWorkspace(
...
+    return null;
+  }
+  if (countResult.value > 0) {
+    return countResult.value;
+  }
+
+  const consumed = Math.max(
+    0,
+    Math.round(await getEsConsumedAwuCreditsForApiKey(auth, { apiKeyName }))
+  );
```
avervaet, line 364: Could this use an atomic seed-if-absent instead of unconditional `SET`?  
If an `INCRBY` lands while ES is loading, won’t this overwrite it and undercount usage?

avervaet (approved):

front/lib/api/keys/spend_limit.ts
tdraier: Indeed 👍 , I will handle that globally for all limiters in another PR

tdraier: https://github.com/dust-tt/dust/pull/30670

PR #30642, opened 2026-08-17
[front] Bind group permission IDs as an array

fontanierh:
## Description

Replace the unscoped system-key workspace scan with the regular group-filtered lookup, and make that lookup bind every group set as one PostgreSQL `bigint[]` consumed by `ANY`.

This keeps query text stable for both small and very large group sets, avoids Sequelize expanding thousands of `IN` placeholders, and prevents the workspace scan that returned 7.61M rows in one minute during the EU DB incident. Optional grant and resource filters continue to apply in Postgres.

## Tests

Added DB-backed coverage for small and 8,194-ID group sets, the single array bind SQL shape, absence of an expanded `IN` clause, optional filters, empty group sets, and scoped and unscoped system-key permission semantics.

Ran the focused Vitest suite: 42 tests passed. Ran Biome on all changed files and the Front TypeScript check.

## Risk

Low. Permission semantics and empty-list behavior are unchanged. The database now receives the same group IDs as one typed array. Roll back this PR to restore the previous lookup behavior.

## Deploy Plan

Deploy Front and front-sse together. Monitor Front DB CPU and `group_permissions` query fingerprints in both regions.

### DIFF
--- a/front/lib/auth.ts
+++ b/front/lib/auth.ts
@@ -963,11 +963,6 @@ export class Authenticator {
       ? allGroups.map((g) => g.id)
       : [];
     const keyGroupModelIds = allGroups.map((g) => g.id);
-    // Unscoped system keys represent nearly every group in their workspace. Bind those ids as one
-    // Postgres array so Sequelize does not expand thousands of placeholders into the query text.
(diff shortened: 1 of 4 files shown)


id13 (commented):
LGTM

PR #30643, opened 2026-08-17
Skip seat sync writes after subscription end

fontanierh:
## Description

Seat reconciliation now ignores a subscription at timestamps equal to or later than its `ending_before`. This prevents Metronome from rejecting removals scheduled exactly at an expired source boundary, while destination subscriptions still receive their additions.

## Tests

Added SEAT_BASED and QUANTITY_ONLY regression coverage for transitions exactly at, 1 ms before, and 1 ms after the source subscription end, including the destination update.

## Risk

Low. The guard only suppresses writes outside a subscription's valid `[starting_at, ending_before)` interval. Rollback restores the prior behavior.

## Deploy Plan

Standard deploy. After deployment, terminate the existing retrying workflow and rerun Poke's Sync Metronome seats.


tdraier (approved):
lgtm

PR #30645, opened 2026-08-17
feat: filter panel for the automations analytics page

avervaet:
## Description

Adds the reusable filter UI and state model for the Automations analytics page. The new `AutomationsFilterPanel` supports filtering by agent, member, and trigger type (`Schedule` or `Webhook`) using the generic analytics filter-panel components. 

This PR adds the panel, filter state, and conversion helpers but does not wire them into the analytics page yet; no existing page behavior is changed until that integration is added.

Visuals: 
Capture d’écran 2026-08-18 à 10 01 44
Capture d’écran 2026-08-18 à 10 01 41
Capture d’écran 2026-08-18 à 10 01 37

## Risk

Low runtime risk because the changes add new components and filter utilities without wiring them into the existing page or changing current data fetching or API behavior. 

## Deploy Plan

- Deploy front

PR #30644, opened 2026-08-17
refactor: extract reusable analytics filter-panel components

avervaet:
## Description

Extracts the reusable analytics filter-panel building blocks from the consumption usage-filter implementation into a generic `filterPanel/` module. The new components cover category navigation, option checkboxes, selection summaries, footer actions, availability status, and summary chips; shared filter-state types and helpers provide category/option-generic selection, clearing, removal, and select-all behavior.

This provides the reusable UI groundwork for the analytics automations page. The following stacked PR, [#30645](https://github.com/dust-tt/dust/pull/30645), builds an `AutomationsFilterPanel` for agent, member, and type filters on top of these generic components; the preceding stacked PR, [#30624](https://github.com/dust-tt/dust/pull/30624), adds the automations analytics trigger-breakdown row that the page will support.

## Tests
- Local testing

## Risk

Low This is a front-end refactor only. The main compatibility surface is the consumption analytics filter UI, which now uses the extracted generic components.  Reverting this PR restores the previous component structure and constant.

## Deploy Plan

- Deploy front

PR #30646, opened 2026-08-17
feat: wire filter panel into the automations analytics page

avervaet:
## Description

Adds the automation analytics filter panel and applied-filter summary to the triggers table. The page now supports filtering by agent, editor, and trigger kind (`schedule` or `webhook`), and threads the selected filters through the client hook, request schema, API route, and trigger-ranking query.

Agent and editor filters are translated into the existing consumption scope filters. Because trigger kind is not indexed in the consumption Elasticsearch documents, kind selections are first resolved to trigger IDs and then applied as a trigger-ID terms filter. With no filters selected, the existing unfiltered query path is preserved.

## Risk

Low. This is a read-only analytics change with an optional request field, so existing unfiltered requests remain compatible. The main additional query cost is the workspace trigger lookup required when filtering by kind, followed by the Elasticsearch terms filter.

## Deploy Plan

- Deploy front

### DIFF
--- a/front-api/routes/w/[wId]/analytics/automations/triggers.ts
+++ b/front-api/routes/w/[wId]/analytics/automations/triggers.ts
@@ -21,7 +21,7 @@ app.post(
   validate("json", AutomationTriggersBodySchema),
   async (ctx): HandlerResult<GetAutomationTriggersResponse> => {
     const auth = ctx.get("auth");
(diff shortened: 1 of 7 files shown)


avervaet:

- #30644
    - #30645
        - #30646 ◀

adrsimon (approved):

PR #30650, opened 2026-08-17
feat: agent and editor tooltips on the automations triggers table

adrsimon:
## Description

Follow-up on the triggers table added in #30618.

Hovering an agent name now shows the same card as the consumption attribution table: description, and model when we have one.

The Members column became Editor, is now avatar-only, and shows name plus email on hover.

Schedule triggers show their cadence ("Every Monday at 9:00") instead of a flat "Schedule" label, built with the existing `describeScheduleConfig`.

Capture d’écran 2026-08-17 à 17 28 15

## Tests

Checked the page manually for the tooltips and the column layout.

## Risk

Low, one admin-only analytics table.

## Deploy Plan

Deploy front.


aubin-tchoi (commented):
LGTM

PR #30651, opened 2026-08-17
[Activation] Track For You page entry source

frankaloia:
## Summary
- track every For You page open with the authenticated user ID
- distinguish email entries with `from_email` and retain the conversation ID when present
- move mount tracking to a normal effect so PostHog is initialized before capture

## Testing
Tested with logging locally

## Risk
Low

## Deploy
1. deploy front

### DIFF
--- a/front/components/pages/workspace/GetStartedPage/index.tsx
+++ b/front/components/pages/workspace/GetStartedPage/index.tsx
@@ -29,7 +29,7 @@ import { FOR_YOU_EMAIL_UTM } from "@app/lib/tracking/campaigns";
 import { getConversationRoute } from "@app/lib/utils/router";
 import { resolveDefaultAgentId } from "@app/types/user";
 import { Button, Spinner } from "@dust-tt/sparkle";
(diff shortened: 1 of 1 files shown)


front/components/pages/workspace/GetStartedPage/index.tsx
```diff
@@ -142,21 +142,22 @@ export function GetStartedPage() {
 
   const firstName = user?.firstName ?? user?.fullName?.split(" ")[0] ?? "there";
 
-  // useLayoutEffect runs before paint; RootLayout only strips UTMs in a
-  // useEffect (after paint), so the campaign params are still on the URL here.
-  useLayoutEffect(() => {
+  useEffect(() => {
```
Nils-Fedrigo, line 145: mmm comment above says it does not work with useEffect but I trust you here!

Nils-Fedrigo (approved):

front/components/pages/workspace/GetStartedPage/index.tsx
frankaloia: That was my comment, I was wrong the first time :)

PR #30652, opened 2026-08-17
Use request-scoped cache for feature flags

flvndvd:
## Description

This change replaces the feature flag LRU cache with a cache scoped to each API request.

### The problem

The previous cache lived for the lifetime of the pod. It required size limits, expiration times, and manual invalidation.

A feature flag change could remain stale until the cache expired. Manual invalidation only cleared the pod handling the change. Other pods could continue using their cached value.

The cache also lived longer than the request that loaded the data, making its lifetime harder to reason about.

### How it works

The new cache uses Hono’s Context Storage middleware (https://hono.dev/docs/middleware/builtin/context-storage).

Hono associates every incoming request with its own async context. Database calls made during that request can retrieve the correct context, even when the pod handles several requests concurrently.

Each request starts with a new empty cache. Repeated calls to the same query and key reuse the same promise, so concurrent calls only perform one database operation.

### Query isolation

Every cached query has a unique identity. Results are stored using both that identity and the query key.

Two different queries cannot read each other’s results, even if they use the same key.

### Request isolation

Every request owns a separate cache. Concurrent requests on the same pod never share cached values.

The cache cannot accumulate data across requests, users, workspaces, pods, or deployments.

### Feature flag consistency

Feature flags remain stable for the duration of a request. A feature flag update becomes visible on the next request.

This avoids refresh logic and invalidation during an active request.

### Cache lifetime

There is no timer or explicit cleanup step.

When the request and its async work finish, Hono’s context is no longer referenced. The cache becomes unreachable and JavaScript garbage collection reclaims it.

Code running outside a Hono request bypasses the cache and queries the database normally.

### DIFF
--- a/front-api/middlewares/request_logger.ts
+++ b/front-api/middlewares/request_logger.ts
@@ -3,14 +3,17 @@ import type { SessionWithUser } from "@app/lib/iam/provider";
 import { getStatsDClient } from "@app/lib/utils/statsd";
 import logger from "@app/logger/logger";
 import tracer from "@app/logger/tracer";
-import type { RequestContext } from "@app/types/shared/utils/request_context";
-import { runWithRequestContext } from "@app/types/shared/utils/request_context";
+import {
+  type RequestContext,
+  RequestQueryCache,
+  type RequestStorageEnv,
+} from "@app/types/shared/utils/request_context";
 import { getClientIpFromContext } from "@front-api/lib/request";
 import { createMiddleware } from "hono/factory";
 import { routePath } from "hono/route";
 
 type RequestLoggerEnv = {
-  Variables: {
+  Variables: RequestStorageEnv["Variables"] & {
     auth?: Authenticator;
     session?: SessionWithUser;
     streaming?: boolean;
@@ -54,12 +57,6 @@ const routeConcurrencyStates = new Map<string, RouteConcurrencyState>();
 
 export const requestLogger = createMiddleware<RequestLoggerEnv>(
   async (c, next) => {
-    if (SKIP_LOGGER_PATHS.has(c.req.path) || c.req.method === "OPTIONS") {
-      // Also drop the APM trace so these requests don't show up in Datadog.
-      tracer.scope().active()?.setTag("manual.drop", true);
-      return next();
-    }
-
     // Mutable: `route` starts as the raw URL path and is updated to the matched
     // route pattern in the finally block (routePath is only set after routing).
     // Any unhandled rejection fired from the handler will see the updated value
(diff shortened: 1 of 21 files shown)


front-api/middlewares/request_logger.ts
```diff
@@ -54,12 +57,6 @@ const routeConcurrencyStates = new Map<string, RouteConcurrencyState>();
 
 export const requestLogger = createMiddleware<RequestLoggerEnv>(
```
aubin-tchoi: nit: not super well named now since it's not only for logging

front/types/shared/utils/request_context.ts
```diff
@@ -1,17 +1,64 @@
-import { AsyncLocalStorage } from "node:async_hooks";
+import { tryGetContext } from "hono/context-storage";
```
aubin-tchoi: shouldn't this be in front-api instead? since specific to hono

front/types/shared/utils/request_context.ts
```diff
@@ -1,17 +1,64 @@
...
+    requestContext: RequestContext;
+  };
+};
+
 export function getRequestContext(): RequestContext | undefined {
-  return requestContext.getStore();
+  return tryGetContext<RequestStorageEnv>()?.get("requestContext");
+}
+
+export function getRequestQueryCache(): RequestQueryCache | undefined {
```
aubin-tchoi: we can return class instances here? if it caches the pointer in this case since it's in-memory?

front/types/shared/utils/request_context.ts
```diff
@@ -1,17 +1,64 @@
...
   route: string;
   url: string;
 };
 
-const requestContext = new AsyncLocalStorage<RequestContext>();
+/**
+ * Memoizes the first result for each key for the lifetime of a Hono request.
+ */
+export class RequestCachedQuery<Key, Value> {
+  readonly cacheId = Symbol();
```
aubin-tchoi, line 33: IIUC Hono handles some form of cleanup of everything that has been set using `c.set` between each request, right? bc the cache key is the same for a given pod here

aubin-tchoi (approved):
LGTM

front/types/shared/utils/request_context.ts
adrsimon: +1, bc of this we bring hono in front in package.json etc

adrsimon (approved):

front/types/shared/utils/request_context.ts
flvndvd: Indeed, fixed.

front/types/shared/utils/request_context.ts
flvndvd: You mean the `cacheId`? Symbol is a runtime primitive, each call yields a different value

front/types/shared/utils/request_context.ts
flvndvd: I wanted to go that way at first but ended up settling on not putting too many constraints just yet. `Value` here could be a single resource or an array of resources. Will probably introduced another Resource specific abstraction based on this one but needs to build more conviction first.

front/types/shared/utils/request_context.ts
aubin-tchoi: the symbol is instantiated once for the lifetime of the pod no, so the same across calls?

flvndvd: It's a class, right? So each new instance gets its own `Symbol`.

aubin-tchoi: ah true it's indeed on the class, misread this

PR #30654, opened 2026-08-17
[front] Rate-limiter backup for the programmatic spend cap

tdraier:
## Description

Second PR of the pool-cap rate-limiter stack. Adds a Redis fixed-window backup for the Metronome **workspace programmatic monthly spend cap**, following the per-API-key pattern: record programmatic AWU usage into a workspace-scoped counter over the contract billing cycle, and enforce it synchronously at message send alongside the Metronome-driven `isProgrammaticApiBlocked`.

Stacked on #30641.

- Reuses the existing `enforce_user_spend_limit_rate_cap` flag for recording + enforcement.
- Threshold is `programmaticMonthlyCapAwuCredits`. **Only a positive cap is enforced from the counter**: a cap of 0 means "always depleted", owned by the programmatic credit-state machine (`isProgrammaticApiBlocked`), so the backup defers to it there and never over-blocks.
- Lazy-seed from Elasticsearch (summed over `usage_type = programmatic`) on a Redis miss.
- Reader/recorder in `lib/api/credits/programmatic_usage_limit.ts`; ES seed + resync in `members_usage.ts`; the shared `resync-spend-limit-counters` poke plugin resyncs the programmatic counter too.
- Consolidated the spend-cap recording blocks in `credit_cost.ts` to fetch the feature flag once.

## Tests

- No new automated tests (internal message-send path, not an endpoint). Intended manual validation: enable the flag on a dust-internal workspace, run the `resync-spend-limit-counters` poke plugin, confirm the programmatic counter matches ES and that a workspace at/over its programmatic cap is blocked with the existing 429.

## Risk

- **Low, fail-open by design** (no positive cap, no billing period, Redis error → allow).
- **Off by default** (`dust_only` flag). No behavior change when off.
- When on, it is a *backup* OR-ed with the existing Metronome-driven block, returning the same 429 — it cannot block where a positive-threshold alert would not.
- Safe to roll back via revert.

## Deploy Plan

1. Merge after #30641; deploy `front` normally. No migration.
2. Flag stays `dust_only` and off — no runtime effect on merge.
3. Validate on a dust-internal workspace (enable flag → run poke resync → compare counter vs Metronome for a cycle) before relying on the Redis path to block.

### DIFF
--- a/front/lib/api/credits/members_usage.ts
+++ b/front/lib/api/credits/members_usage.ts
@@ -1,5 +1,6 @@
 import {
   makeApiKeySpendLimitAwuCreditsRateLimitKey,
+  makeProgrammaticSpendLimitAwuCreditsRateLimitKeyForWorkspace,
   makeSpendLimitAwuCreditsRateLimitKeyForUser,
   makeSpendLimitCycleWindowBounds,
 } from "@app/lib/api/assistant/rate_limits";
@@ -25,6 +26,7 @@ import {
   CONTRACT_CREDIT_TYPE_FREE_SEAT,
   getCreditTypeAwuId,
   toFreeMetronomeUserId,
+  USAGE_TYPE_PROGRAMMATIC,
 } from "@app/lib/metronome/constants";
 import { getCachedMetronomeCurrentBillingPeriod } from "@app/lib/metronome/contracts";
 import { getPerUserAwuUsage } from "@app/lib/metronome/per_user_usage";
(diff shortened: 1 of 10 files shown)


tdraier:

- #30641
    - #30654 ◀
        - #30670
        - #30655

front/lib/api/credits/members_usage.ts
```diff
@@ -1086,6 +1148,48 @@ export async function resyncApiKeySpendLimitCountersFromEsUsage(
...
+    );
+  }
+  const bounds = makeSpendLimitCycleWindowBounds(
+    cycle.cycleStart,
+    cycle.cycleEnd
+  );
+
+  const consumed = await getEsConsumedProgrammaticAwuCredits(auth, { cycle });
+  const setResult = await setFixedWindowCount({
+    key: makeProgrammaticSpendLimitAwuCreditsRateLimitKeyForWorkspace(
```
avervaet, line 1194: Nit as this is a niche case but: 

Looks like the resync path conflates “no programmatic usage” with “Elasticsearch read failed.” getEsConsumedProgrammaticAwuCredits returns 0 after an analytics error, and this function then unconditionally overwrites the Redis counter with 0 and reports success.

Running the resync-spend-limit-counters Poke action during a transient Elasticsearch outage could therefore erase a valid live counter and disable the backup cap until usage rebuilds it.

Could we propagate the ES failure (for example with a Result/nullable value) and skip the SET when the source-of-truth read failed? The resync could also surface the failure instead of reporting the counter as successfully seeded.

avervaet (approved):
LGTM

PR #30660, opened 2026-08-17
Add the Dust filesystem mount and supervisor

flvndvd:
## Description

The previous PRs added the front client and local file store. This PR adds the Linux layer that makes them available to sandbox programs under `/files`.

It does not change the database model or the front filesystem API.

## What gets mounted?

The new `dsbx filesystem mount` command creates one FUSE mount:
```
/files/
├── conversation/
└── pod/
```

Because both roots are inside the same mount, a normal `mv` can move a file between a Conversation and a Pod without returning `EXDEV`.

The FUSE callbacks connect Linux file commands to the store added previously. This includes `lookup`, `listing`, `create`, `open`, `read`, `write`, `truncate`, `fsync`, `rename`, `unlink`, `mkdir`, `rmdir`, and executable bits.

## How do open files behave?

The mount keeps a Linux file handle for every open file.

A file remains usable when it is renamed. If it is deleted while still open, the existing handle can still read its local bytes, but it cannot later publish content into the deleted node.

Rename-over also works. This matters for editors that save by creating a temporary file and renaming it over the original
file.

Shell redirection needs some care:

```
echo "hello" > file.txt
```

Linux first opens `file.txt` with `O_TRUNC`, then writes the new bytes. Committing the truncate immediately would create one empty revision followed by another revision containing `"hello"`. The mount waits for the following `write`, `fsync`, or final `close` so this command creates one revision.

Linux may call `flush` more than once, so repeated flushes are safe. Linux also ignores errors returned during the final release of a file handle. If that final save fails, the mount removes the changed cache entry so a later `open` cannot return bytes that were never saved.

## What happens while Front or GCS is slow?

FUSE calls arrive on a small number of request threads. Running a long front request or GCS upload directly on those threads could freeze unrelated file commands.

The mount sends remote work to separate workers and limits how many calls may run or wait at once. Calls that only use an
already-open local file do not enter that queue.

Reads use a duplicated local file descriptor. They do not wait for the mutex that protects writes and remote commits. This
allows large parallel reads to continue without returning `EAGAIN` while another action uses the file handle.

Final file releases use their own worker capacity. Closing dirty files therefore cannot consume every FUSE request thread
during a front or GCS outage.

## How do executable bits work?

Files are readable and writable by both Linux users used inside a sandbox. The token minted by our API remains the access check for the Conversation and Pod.

The mount stores and checks the executable bits separately. A script cannot run before `chmod +x`, and it can run afterward. Other permission changes, such as removing the read or write bits, are rejected because this filesystem does not store them.

## Why use `O_NOFOLLOW`?

The mount process runs as root and opens its token and cache files from disk.

Without `O_NOFOLLOW`, a symbolic link could redirect one of those opens to another file. For example, a cache file named `inode-123` could point outside the cache directory and cause the root process to read or overwrite that target.

`O_NOFOLLOW` makes Linux reject the open when the final path is a symbolic link. The parent directories are also checked so another sandbox user cannot replace them.

The cache directory uses mode 0700 and contains a marker written by the filesystem. Startup refuses to clean a directory
without that marker. This prevents a bad `--staging-dir` value from deleting files in an unrelated private directory.

## How does the mount recover after a crash?

The new `dsbx filesystem supervise` command starts the mount process and restarts it after an unexpected exit.

Before restarting, it detaches the old mount with `MNT_DETACH`. This repairs the common state where `/files` returns:

```
Transport endpoint is not connected
```

The supervisor also prevents two copies from managing the same mount point at once.

## Tests

Wrote some "acceptance" tests. It covers create, read, write, truncate, rename, rename-over, unlink, open-file behavior, conversation-to-Pod moves, and executable bits.

A live test mounted the same Pod in two sandboxes. Every successful overwrite increased the revision by exactly one. Changes appeared in the other sandbox with a 418 ms median delay. Rename kept the same inode. Two concurrent writes produced one success and one `ESTALE`.

The same run compared the new mount with the current two-gcsfuse setup:

```
 Command                     Two gcsfuse mounts    Dust FUSE
━━━━━━━━━━━━━━━━━━━━━━━━━━  ━━━━━━━━━━━━━━━━━━━━  ━━━━━━━━━━━
 Cold read, 4 KiB                      2,056 ms       350 ms
──────────────────────────  ────────────────────  ───────────
 Write and fsync, 4 KiB                2,195 ms       662 ms
──────────────────────────  ────────────────────  ───────────
 Rename                                1,631 ms       244 ms
──────────────────────────  ────────────────────  ───────────
 Conversation-to-Pod move              4,240 ms       223 ms
```

## What is not in this PR?

This PR adds the mount and supervisor commands. It does not switch existing sandboxes to the new filesystem.

Choosing this backend for a Conversation or Pod, and enabling it gradually, will come in a separate PR.


cli/dust-sandbox/docs/database-filesystem.md
```diff
@@ -0,0 +1,353 @@
+# Dust database filesystem
```
flvndvd, line 1: Committing for posterity.

cli/dust-sandbox/src/commands/filesystem/fuse/handles.rs
```diff
@@ -0,0 +1,313 @@
+use std::collections::{HashMap, HashSet};
```
flvndvd, line 8: From a high-level view this file/module keeps track of files and directories that Linux currently has open.

It does not call our API or GCS. It stores short-lived, in-memory state between open and release.

cli/dust-sandbox/src/commands/filesystem/fuse/handles.rs
```diff
@@ -0,0 +1,313 @@
...
+}
+
+// A final commit may hold one stripe while it calls Front and GCS. A larger
+// fixed set keeps the memory bounded while making unrelated collisions rare.
+const INODE_LOCK_STRIPES: usize = 1024;
+const MAX_OPEN_DIRECTORY_ENTRIES: usize = 200_000;
+const MAX_OPEN_FILE_HANDLES: usize = 4096;
+const MAX_OPEN_DIRECTORY_HANDLES: usize = 1024;
+
+pub(super) struct InodeLocks {
```
flvndvd, line 133: `InodeLocks` serializes a few actions that must not overlap for the same inode.

For example, a final failed save must remove its changed cache entry before another open can reuse it. Both paths take the same inode lock.

There are 1,024 locks rather than one lock for every possible inode. An inode selects one with:

```
inode % 1024
```

This keeps memory fixed. Downside though, two unrelated inodes can occasionally select the same lock and wait for each other, but they cannot corrupt each other’s data.

cli/dust-sandbox/src/commands/filesystem/fuse/linux.rs
```diff
@@ -0,0 +1,222 @@
...
+
+pub(super) fn file_attributes(node: &Node, uid: u32, gid: u32) -> FileAttr {
+    FileAttr {
+        ino: node.inode,
+        size: node.size,
+        blocks: node.size.div_ceil(u64::from(BLOCK_SIZE)),
+        atime: time_from_ms(node.modified_at_ms),
+        mtime: time_from_ms(node.modified_at_ms),
+        ctime: time_from_ms(node.modified_at_ms),
+        crtime: time_from_ms(node.created_at_ms),
```
flvndvd, line 41: We don't track yet exact Linux meanings here. We might revisit in the future.

cli/dust-sandbox/src/commands/filesystem/fuse/operations.rs
```diff
@@ -0,0 +1,564 @@
...
+        _change_time: Option<SystemTime>,
+        _backup_time: Option<SystemTime>,
+        _flags: Option<BsdFileFlags>,
+        reply: ReplyAttr,
+    ) {
+        if uid.is_some() || gid.is_some() {
+            reply.error(Errno::EPERM);
+            return;
+        }
+        if atime.is_some() || mtime.is_some() {
```
flvndvd, line 120: Returning `ENOTSUP` for something like this:
```
touch -m file.txt
touch -a file.txt
```

As our FS does not distinguish dates yet.

cli/dust-sandbox/tests/filesystem_benchmark.ts
```diff
@@ -0,0 +1,403 @@
+#!/usr/bin/env bun
```
flvndvd, line 1: Just some benchmark scripts

cli/dust-sandbox/src/commands/filesystem/fuse.rs
```diff
@@ -0,0 +1,534 @@
...
+            }
+        };
+        Ok((node, handle))
+    }
+
+    fn unlink_node(&self, parent: INodeNo, name: &str) -> io::Result<()> {
+        let _namespace = self.namespace_write()?;
+        let node = self.lookup_node(parent, name)?;
+        let _inode = self.inode_locks.lock(node.inode)?;
+        self.store.remove_file(parent, name)?;
```
smb2268: should we pass in the inode here instead of the path since we already performed the lookup above? or is this safe enough as is?

cli/dust-sandbox/src/commands/filesystem/fuse.rs
```diff
@@ -0,0 +1,534 @@
...
+        let writable = is_writable(flags);
+        // Namespace removal takes the same inode lock after resolving the name.
+        // A local open therefore either installs its handle before unlink marks
+        // it, or starts after the node was removed and fails cleanly.
+        let _inode = self.inode_locks.lock(inode)?;
+        let opened = self.store.open_content(inode, flags)?;
+        let dirty = writable && flags & libc::O_TRUNC != 0;
+        let result = (|| {
+            if dirty {
+                opened.file.set_len(0)?;
```
smb2268: I think necessary to set opened.node.size(0) here as well

cli/dust-sandbox/src/commands/filesystem/fuse/linux.rs
```diff
@@ -0,0 +1,224 @@
...
+    pub block_size: u32,
+    pub name_length: u32,
+    pub fragment_size: u32,
+}
+
+pub(super) fn file_attributes(node: &Node, uid: u32, gid: u32) -> FileAttr {
+    FileAttr {
+        ino: node.inode,
+        size: node.size,
+        blocks: node.size.div_ceil(u64::from(BLOCK_SIZE)),
```
smb2268: Should this divide by 512? st_blocks is in 512-byte sectors, I think this under-reports

cli/dust-sandbox/src/commands/filesystem/store/content_cache.rs
```diff
@@ -332,35 +337,96 @@ impl Drop for ContentLease {
...
+            !suffix.is_empty() && suffix.bytes().all(|byte| byte.is_ascii_digit())
+        }) || name.starts_with(".tmp");
+        if !managed_name || !(entry_type.is_file() || entry_type.is_symlink()) {
+            // Never clean an arbitrary private directory passed by mistake.
+            return Err(errno(libc::EACCES));
         }
+        managed_files.push(entry.path());
+    }
+    if !marker_found {
+        if !managed_files.is_empty() {
```
smb2268, line 393: should we check for is_empty() or all entries are .tmp* here?

smb2268 (approved):
A bit out of my domain here but I think it looks sound. /flavify and I just have a few notes 😜

cli/dust-sandbox/src/commands/filesystem/fuse/linux.rs
flvndvd: Indeed, I conflated two different Linux blocks. Good catch \o/

Was easy to reproduce on the mount, `du` on a 1MB file reported `128kb` 🙈

cli/dust-sandbox/src/commands/filesystem/store/content_cache.rs
flvndvd: Indeed, will do a dedicated PR to fix this.

PR #30661, opened 2026-08-17
feat: disabled_by_admin trigger status with lock enforcement

PopDaph:
## Description

Adds a `disabled_by_admin` trigger status: when an admin pauses a trigger, non-admin editors can neither re-enable it nor set/clear the status themselves. Enforcement lives on `TriggerResource` (`canUpdateStatusTo`, guards in `update()`/`enable()`/`disable()`) with matching 400/403 pre-checks on the editor triggers PATCH route. Also makes system-owned statuses (`relocating`/`downgraded`) immutable through the editing path for everyone — only the restore bulk jobs (via `enable()`/`disable()`) cross them, and a dedicated test pins that asymmetry.

Includes the shared `TriggerStatusChip` and the personal-settings triggers table update (compile-forced by the enum's exhaustive Record).

Part 1/4 of the trigger toggle stack; the admin status endpoint, analytics toggle, and editor UI surfacing follow.

## Tests

New functional tests on the triggers PATCH route: non-admin blocked from setting/clearing the lock (403), system-status transitions blocked for everyone incl. admins (400), unchanged-status saves still pass, and the update()/enable() restore asymmetry.

## Risk

New enum value in private API responses only; exhaustive maps updated, client switches use `assertNeverAndIgnore`. No migration, no rollout ordering.

### DIFF
--- a/front-api/routes/w/[wId]/assistant/agent_configurations/[aId]/triggers/index.test.ts
+++ b/front-api/routes/w/[wId]/assistant/agent_configurations/[aId]/triggers/index.test.ts
@@ -1,8 +1,10 @@
 import { Authenticator } from "@app/lib/auth";
 import { SpaceResource } from "@app/lib/resources/space_resource";
+import { TriggerResource } from "@app/lib/resources/trigger_resource";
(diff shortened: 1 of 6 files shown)


adrsimon (approved):
LGTM!

PR #30662, opened 2026-08-17
feat: admin endpoint to enable or disable a trigger

PopDaph:
## Description

Adds `PATCH /api/w/:wId/assistant/agent_configurations/:aId/triggers/:tId/status` (body `{ status: "enabled" | "disabled" }`, admin-or-editor). The existing bulk PATCH is editor-only and a full-object replace, so it cannot back an admin toggling someone else's trigger. Delegates to `TriggerResource.enable()/disable()` — `enable()` re-derives the editor's authenticator for the Temporal schedule registration, which is exactly what the admin-on-someone-else's-trigger case needs. An admin disable stores `disabled_by_admin`; an editor's own disable stays plain `disabled`. System-owned statuses return 400.

Part 2/4, stacked on #30661. Consumed by the analytics toggle PR.

## Tests

15 functional tests: editor self-enable/pause, admin lock + re-enable, non-admin 403s, system-status 400s, unknown trigger 404, agent mismatch 400, invalid bodies.

## Risk

New private route, no existing behavior changed. No migration.

### DIFF
--- a/front-api/routes/w/[wId]/assistant/agent_configurations/[aId]/triggers/[tId]/index.ts
+++ b/front-api/routes/w/[wId]/assistant/agent_configurations/[aId]/triggers/[tId]/index.ts
@@ -1,10 +1,12 @@
 import { workspaceApp } from "@front-api/middlewares/ctx";
 
+import status from "./status";
 import webhookRequests from "./webhook_requests";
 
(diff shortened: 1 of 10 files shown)


adrsimon (approved):
🔥

PR #30663, opened 2026-08-17
feat: enabled toggle on the automations analytics table

PopDaph:
## Description

Adds an "Enabled" column to the admin-only Automations analytics table: a SliderToggle per trigger, confirm dialog on disable (which admin-locks the trigger), instant enable. The table's Elasticsearch query is not revalidated after a toggle — a local per-row status override (cleared on refetch) reflects the change instead. `relocating`/`downgraded` render a disabled toggle with a tooltip.

Part 3/4, stacked on #30661 and the status endpoint PR.

## Tests

Analytics fixture updated for the new `status` row field; toggle behavior manually tested on the dev workspace (enable, confirm-on-disable, cancel, lock visible to the editor).

## Risk

Additive row field on a private analytics endpoint; UI is behind the `enable_analytics_automations` feature flag.


adrsimon (approved):
LGTM

PR #30664, opened 2026-08-17
feat: surface trigger status in editor UIs

PopDaph:
## Description

Shows trigger status where editors look: a status chip on the agent-detail trigger cards when not enabled, and locked status toggles in the schedule/webhook edition sheets — non-admins cannot flip an admin-locked trigger, nobody can flip system-owned statuses, each with an explanatory tooltip. Without this, users could flip the form toggle and only fail at save time.

Part 4/4, stacked on #30661. Independent of the analytics toggle PR.

## Tests

UI-only; manually tested the locked sheet toggle (admin lock and system statuses) and the card chips on the dev workspace.

## Risk

Presentation only; enforcement ships in #30661.

### DIFF
--- a/front/components/agent_builder/triggers/TriggerStatusToggle.tsx
+++ b/front/components/agent_builder/triggers/TriggerStatusToggle.tsx
@@ -1,0 +1,58 @@
+import type { TriggerViewsSheetFormValues } from "@app/components/agent_builder/triggers/triggerViewsSheetFormSchema";
+import { TRIGGER_STATUS_LABELS } from "@app/components/triggers/TriggerStatusChip";
+import { useAuth } from "@app/lib/auth/AuthContext";
+import { getTriggerStatusOwner } from "@app/types/assistant/triggers";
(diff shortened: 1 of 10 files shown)


front/components/agent_builder/triggers/TriggerStatusToggle.tsx
```diff
@@ -0,0 +1,58 @@
...
+  return (
+    <div className="space-y-1">
+      <Label>Status</Label>
+      <div className="flex flex-row items-center gap-2">
+        <span className="min-w-16 whitespace-nowrap">{statusLabel}</span>
+        {isStatusLocked ? (
+          <Tooltip
+            label={
+              statusOwner === "system"
+                ? "This trigger's status is managed by Dust."
```
adrsimon, line 47: nit: temporarily managed by Dust ? ultimately we'll end up re-activating them either in case of re-upgraded plan or relocation
but wfm really nit

adrsimon (approved):
LGTM

PR #30606, opened 2026-08-17
[front] enh: rework inline tool approval UI

aubin-tchoi:
## Description

This PR reworks the Inline MCP approval UI, mostly based on the approval dialog we used to have.

Main changes:
- Use a card instead of a content message (bg color, shadow)
- Update the title to mention the agent's name, add a one-liner to explain what action is being taken
- Show queue position
- Improve loading states

Before
image

After
image

Next step is to work on the `Details` part.

## Tests

- Tested locally.

## Risk

- Low.

## Deploy Plan

- Deploy front.

PR #30665, opened 2026-08-17
[front] enh: replace "Always approve" checkbox with separate button

aubin-tchoi:
## Description

This PR follows up on https://github.com/dust-tt/dust/pull/30606

Replaces the "Always allow" checkbox with a third button, mimicking the "Deny", "Allow once", "Always allow" pattern that exists for systems like geolocalisation in browsers.

The decision is really a 3-state selection rather than a 4-state one; checking the checkbox while denying doesn't do anything.

For medium stakes, the Always approve button has a tooltip that explains what the scope of the "always" is.
I did not find a perfect way to explicit this scope in the card content itself, but this tooltip is right in the path of a user that would need explaining, whether it is on the first try, or on a second try where they try to understand why they'd have to click again.

> [!NOTE]
> Interesting learning from IRL chat with Clément: the Always allow button is secondary and in the middle. In the middle because it's not the default "Accept" action a user would expect on the bottom right and is a deliberate choice. Not primary to avoid 2 primary buttons and because the Allow once button is the 0-commitment default choice that is part of the golden path a user can follow by following only primary buttons.

Current
image

Proposal
image

## Tests

- Tested locally.

## Risk

- Low.

## Deploy Plan

- Deploy front.

### DIFF
--- a/front/components/actions/blocked/ToolValidationCard.tsx
+++ b/front/components/actions/blocked/ToolValidationCard.tsx
@@ -226,7 +226,7 @@ export function ToolValidationCard({
       </div>
 
       <div className="flex flex-col gap-4 px-5 py-4">
-        <div className="text-base">{displayLabel}</div>
+        <div className="text-base text-faint">{displayLabel}</div>
         {canCurrentUserRespond ? (
           <>
             {errorMessage && (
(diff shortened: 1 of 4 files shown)


ClementAupiais (approved):
LGTM

PR #30667, opened 2026-08-17
Return affected filesystem node IDs

flvndvd:
## Description

The filesystem daemon must know which node Front actually removed or replaced. Looking up the path before the mutation
is not enough because another sandbox may change that path first.

This PR makes Front return that information directly:
- `remove` returns `removedNodeId`
- `rename` returns `replacedNodeId`, or `null` when nothing was replaced

These IDs come from the rows changed inside the mutation transaction. They are also stored in the mutation receipt, so
retrying a request returns the original result even if the path has since been reused.

There is no database migration. Nothing consumes this endpoint yet.


smb2268 (approved):
LGTM

PR #30669, opened 2026-08-18
[front] fix: Switch scroll sentinel in knowledge side panel

id13:
## Description

The knowledge picker in the agent builder stops loading after the first 50 nodes, so anyone with a lot of nodes only sees the beginning of the alphabet. Fixes dust-tt/tasks#10086.

The infinite scroll in DataSourceList was listening for scroll on its own container, and that container never scrolls. It is replaced it with our InfiniteScroll component.

## Tests

Tested locally

## Risk

Low

Blast radius: DataSourceList only, 5 call sites, all inside the knowledge sheet. Only the node table passes onLoadMore, the other four keep hasMore false so the sentinel and the loader never render. Clicked through spaces, categories, data source views, nodes, nested nodes and search, plus selection and select all.

## Deploy Plan

Deploy front.

### DIFF
--- a/front/components/agent_builder/capabilities/knowledge/DataSourceList.tsx
+++ b/front/components/agent_builder/capabilities/knowledge/DataSourceList.tsx
@@ -12,17 +12,11 @@ import {
   pathToString,
   removeNodeFromTree,
 } from "@app/components/data_source_view/context/utils";
+import { InfiniteScroll } from "@app/components/InfiniteScroll";
 import { isRemoteDatabase } from "@app/lib/data_sources";
 import { Checkbox, cn, Icon, Separator, Spinner } from "@dust-tt/sparkle";
 import type { ComponentType, ReactNode } from "react";
-import {
-  Fragment,
-  useCallback,
-  useContext,
-  useEffect,
-  useMemo,
-  useRef,
-} from "react";
+import { Fragment, useCallback, useContext, useMemo } from "react";
 
 export interface DataSourceListItem {
   id: string;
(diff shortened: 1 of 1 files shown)


id13:
PMRR

front/components/agent_builder/capabilities/knowledge/DataSourceList.tsx
```diff
@@ -382,7 +354,13 @@ export function DataSourceList({
...
             <div
               className="flex cursor-pointer items-center justify-between rounded-md p-3 hover:bg-muted/60"
-              onClick={() => item.onClick?.()}
+              onClick={() => {
+                if (item.onClick) {
+                  item.onClick();
+                } else if (shouldShowCheckbox) {
+                  void handleSelectionChange(item, selectionState !== true);
+                }
+              }}
```
id13, line 363: side 🚗 : it was very bothering to aim for the checkbox, now clicking on the row ticks it.

front/components/agent_builder/capabilities/knowledge/DataSourceList.tsx
```diff
@@ -12,17 +12,11 @@ import {
   pathToString,
   removeNodeFromTree,
 } from "@app/components/data_source_view/context/utils";
+import { InfiniteScroll } from "@app/components/InfiniteScroll";
```
ykmsd, line 15: I didn't know we have this 😄

ykmsd (commented):
ty!

PR #30655, opened 2026-08-17
[front] Rate-limiter backup for the workspace usage cap

tdraier:
## Description

Final PR of the pool-cap rate-limiter stack. Adds a Redis fixed-window backup for the Metronome **workspace usage cap** (`usageCapCredits`, the PAYG pool cap): record pool AWU usage into a workspace-scoped counter over the contract billing cycle, and enforce it synchronously at message send alongside the Metronome-driven pool state (`credits_exhausted`).

Stacked on #30654.

- Reuses the existing `enforce_user_spend_limit_rate_cap` flag.
- Threshold is `usageCapCredits` (null = no cap → no enforcement).
- **Pool = non-free-seat usage only.** The cap measures the AWU pool credit type; free-seat usage draws from a separate Metronome credit type. Both the ES seed (`must_not is_free_seat`) and recording (programmatic OR paid-seat user) exclude free-seat usage to match Metronome and avoid over-blocking.
- Lazy-seed from Elasticsearch on a Redis miss; reader/recorder in `lib/api/credits/usage_configuration.ts`; ES seed + resync in `members_usage.ts`; the shared `resync-spend-limit-counters` poke plugin resyncs the usage-cap counter too.

## Tests

- `npx tsgo --noEmit` passes; `biome check` clean.
- No new automated tests (internal message-send path). Intended manual validation: enable the flag on a dust-internal workspace, run the `resync-spend-limit-counters` poke plugin, confirm the usage-cap counter matches ES pool consumption and that a workspace at/over `usageCapCredits` is blocked with the existing 403 `credits_exhausted`.

## Risk

- **Low, fail-open by design** (no cap, no billing period, Redis error → allow).
- **Off by default** (`dust_only` flag). No behavior change when off.
- When on, it is a *backup* OR-ed with the existing `credits_exhausted` pool state, returning the same 403.
- **One assumption to confirm:** the pool cap excludes free-seat usage (free-seat is a separate Metronome credit type). If `usageCapCredits` actually includes free-seat usage, drop the `must_not is_free_seat` filter + the seat check in recording — the current choice under-counts (fails toward under-blocking, the safe direction) in that case.
- Safe to roll back via revert.

## Deploy Plan

1. Merge after #30654; deploy `front` normally. No migration.
2. Flag stays `dust_only` and off — no runtime effect on merge.
3. Validate on a dust-internal workspace (enable flag → run poke resync → compare counter vs Metronome pool consumption for a cycle) before relying on the Redis path to block.

PR #30670, opened 2026-08-18
[front] Fix lazy-seed race in the spend-cap counters

tdraier:
## Description

Addresses @avervaet's review note on #30641: the spend-cap lazy seed used an unconditional `SET`, which could clobber a concurrent `addFixedWindowCount` (INCRBY) landing while the Elasticsearch seed value was being computed — undercounting usage.

Applies across **all four** spend-cap counters (per-user, per-API-key, programmatic, workspace usage cap). Stacked on #30655.

- Adds `seedFixedWindowCountIfAbsent` to the rate limiter: an atomic SET-if-absent (via Lua) that seeds the counter only when the key is missing and returns the *effective* count (the seeded value, or the concurrently-written value if an INCRBY won the race).
- Each lazy seed now uses it instead of the unconditional `setFixedWindowCount`, and returns the effective count so enforcement reads the authoritative value rather than the (possibly stale) ES value.
- The unconditional `setFixedWindowCount` stays on the poke resync path, where overwriting from the source of truth is intentional.

## Tests

- No new automated tests (internal rate-limiter primitive + message-send path). The race is a narrow interleaving (an INCRBY during the ES-await on a fresh key); the atomic Lua makes seed-vs-increment ordering well-defined.

## Risk

- **Low.** Behavior is unchanged in the common (non-race) case: absent key → seed to ES value → return it. Off by default behind the same `enforce_user_spend_limit_rate_cap` flag.
- Strictly safer than before: the seed can no longer overwrite a live INCRBY. Any residual race (an INCRBY winning before the seed) leaves the live value in place and skips the ES baseline for that read — an under-count (fails toward under-blocking), corrected by the next poke resync.
- Safe to roll back via revert.

## Deploy Plan

1. Merge after #30655; deploy `front` normally. No migration.
2. No runtime effect until the `enforce_user_spend_limit_rate_cap` flag is enabled (unchanged rollout).

### DIFF
--- a/front/lib/utils/rate_limiter.ts
+++ b/front/lib/utils/rate_limiter.ts
@@ -374,6 +374,77 @@ export async function setFixedWindowCount({
 }
 
 /**
+ * Atomically seeds the fixed-window counter for `key` in the window identified
+ * by `bounds` to `value`, but only if it does not already exist, and returns the
+ * effective count afterwards (the seeded `value`, or the current value when a
+ * concurrent `addFixedWindowCount` already created it).
+ *
+ * Unlike `setFixedWindowCount`, this never overwrites a live counter: use it for
+ * lazy backfill on a read miss, where an `addFixedWindowCount` (INCRBY) landing
+ * while the seed value is being computed must not be clobbered (which would
+ * undercount usage). Returns a Result so callers can fall back on failure.
+ */
+export async function seedFixedWindowCountIfAbsent({
+  key,
+  bounds,
+  value,
+  logger,
+}: {
+  key: string;
+  bounds: FixedWindowBounds;
+  value: number;
+  logger: LoggerInterface;
+}): Promise<Result<number, Error>> {
+  if (!Number.isInteger(value) || value < 0) {
+    return new Err(new Error("value must be a non-negative integer."));
+  }
+
+  const redisKey = makeRateLimiterKey(`${key}:${bounds.label}`);
+  const expireAtMs = bounds.windowEndMs + FIXED_WINDOW_EXPIRE_GRACE_MS;
+
+  // Seed-if-absent + read-back in one atomic step: only SET (and set the expiry)
+  // when the key is missing, otherwise leave the concurrently-written value
+  // untouched. Returns the effective count either way.
+  const luaScript = `
+    local key = KEYS[1]
+    local value = tonumber(ARGV[1])
+    local expire_at_ms = tonumber(ARGV[2])
+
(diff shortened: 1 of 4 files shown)


tdraier:

- #30641
    - #30654
        - #30670 ◀

front/lib/utils/rate_limiter.ts
```diff
@@ -373,6 +373,77 @@ export async function setFixedWindowCount({
...
+
+  // Seed-if-absent + read-back in one atomic step: only SET (and set the expiry)
+  // when the key is missing, otherwise leave the concurrently-written value
+  // untouched. Returns the effective count either way.
+  const luaScript = `
+    local key = KEYS[1]
+    local value = tonumber(ARGV[1])
+    local expire_at_ms = tonumber(ARGV[2])
+
+    local existing = redis.call('GET', key)
```
avervaet:
```suggestion
local seeded = redis.call('SET', key, value, 'NX')
if seeded then
  redis.call('PEXPIREAT', key, expire_at_ms)
  return value
end

return redis.call('GET', key)
```

avervaet: nit: The current GET-then-SET is atomic inside Lua, so this is mostly for the sake of clarity

front/lib/utils/rate_limiter.ts
```diff
@@ -373,6 +373,77 @@ export async function setFixedWindowCount({
...
+    return tonumber(existing)
+  `;
+
+  try {
+    const redis = await getRedisStreamClient({ origin: "rate_limiter" });
+    const effective = await redis.eval(luaScript, {
+      keys: [redisKey],
+      arguments: [value.toString(), expireAtMs.toString()],
+    });
+    const count = Number(effective);
```
avervaet, line 434: nit: Number(null) is 0, so a malformed Lua result or a nil response from tonumber(existing) could silently become a valid zero count. Since the counter is supposed to be a non-negative integer, I’d suggest checking for null explicitly and using Number.isSafeInteger:

avervaet (approved):
Proposal:
The three changed callers, plus the workspace reader from #30655, now have essentially the same flow:
-> getFixedWindowCount
->return immediately if positive
->fetch ES
->seed Redis
-> return the effective count or fall back

A shared helper near the rate-limiter primitive could reduce the chance that the four counters diverge again. Something along the lines of a generic readFixedWindowCountWithLazySeed({ ..., fetchSeedValue }).

PR #30674, opened 2026-08-18
Open analytics/automations to workspace managers

avervaet:
## Description

Moves `analytics/automations` from admin-only access to manager-level access, matching the existing analytics consumption pages. The page is now mounted under the manager-level route layout, the triggers API accepts managers, and the Automation navigation entry is enabled for managers. Regular users and builders remain excluded.

The automations data now distinguishes an existing webhook source in an inaccessible space from a missing source. Managers see `Restricted webhook` with an explanatory tooltip for the former; genuinely missing sources continue to use the generic webhook display. Agent descriptions remain caller-scoped through the existing analytics label-resolution helper.

Managers also receive admin-equivalent trigger status controls: they can disable, lock, and re-enable any workspace trigger, including triggers owned by another member or already locked by an admin. Manager/admin disables continue to use the existing `disabled_by_admin` status value, whose UI label is now `Locked`. The broader trigger rename/reconfiguration path remains admin-only.

Discussion thread about whether managers should be able to enable/disable triggers: https://dust4ai.slack.com/archives/C0BJ3T5SA0L/p1787051725501499

## Tests

- Manual browser verification.

## Risk

Medium. This expands workspace-manager permissions: **a manager can now change the status of any trigger in the workspace, not only triggers they own, and can re-enable triggers locked by an admin or another manager.** Existing regular-user and builder restrictions remain in place.

## Deploy Plan

- Deploy front


front/components/workspace/analytics/automations/AutomationsTriggersTable.tsx
```diff
@@ -324,12 +343,16 @@ export function AutomationsTriggersTable({
...
       if (success) {
-        // This page is admin-only, so a disable is stored server-side as
-        // disabled_by_admin.
+        // An admin's disable is stored server-side as disabled_by_admin; a
+        // manager disabling their own automation stays a plain disabled.
         setStatusOverrides((overrides) => ({
           ...overrides,
           [trigger.triggerId]:
-            nextStatus === "disabled" ? "disabled_by_admin" : "enabled",
+            nextStatus === "disabled"
```
avervaet: @PopDaph not sure if we intended two have two different status for this but weir to add disable_by_admin for a manager action. Maybe we can rename disable_by_admin ? any tought?

PopDaph: I thought only an admin could access this page -> the new status is useful cause if the admin deactivates your trigger, as user you cannot reactivate it. 

It was shipped this morning, I have no strong conviction on the rename if you think we should rename let's rename 🙏🏻

avervaet:
Note that in triggers.status, the value remains disabled_by_admin even if the enablement was performed by a manager. We might want to backfill it with disabled_by_manager to match the new pattern.

front/components/navigation/config.test.ts
```diff
@@ -0,0 +1,93 @@
+import { subNavigationAdmin } from "@app/components/navigation/config";
+import { LightWorkspaceFactory } from "@app/tests/utils/LightWorkspaceFactory";
+import type { MembershipRoleType } from "@app/types/memberships";
+import type { PlanType, SubscriptionType } from "@app/types/plan";
+import type { WorkspaceType } from "@app/types/user";
+import { describe, expect, it } from "vitest";
+
+const PLAN: PlanType = {
```
flvndvd: Pretty sure we have an existing factory/fixture for this.

front/components/navigation/config.test.ts
```diff
@@ -0,0 +1,93 @@
...
+    capabilities: { images: { maxImagesPerWeek: -1 } },
+    canUseProduct: true,
+  },
+  trialPeriodDays: 0,
+  isByok: false,
+  isAuditLogsAllowed: true,
+  hasAdvancedModelAccess: true,
+};
+
+const SUBSCRIPTION: SubscriptionType = {
```
flvndvd: Let's avoid those, and use Factory, it leads to plenty of side updates when touching resources after.

front/components/navigation/config.test.ts
flvndvd: ah maybe it's front-end tests runtime that limits us.

flvndvd (approved):

PR #30676, opened 2026-08-18
Admin governance: Dual write skill editor permissions in regular auto group

fabiencelier:
## Description

 - Dual write permissions for editor directly instead of using the editor's group
 - Backfill script 

## Tests

updated unit tests + tested locally

## Risk

Can have slight discrepancie if a skill editor is remove in the middle of the script writing the new permission
Very unlikely, will check the logs after the migration and reconcile if needed

## Deploy Plan

merge and deploy front 
run migration script
```
  cd front && npx tsx migrations/20260817_backfill_skill_editor_user_grants.ts --execute \
    2>&1 | tee ~/skill_editor_grants_backfill.log
```
Check if any skill may have stale permissions due to change during write

### DIFF
--- a/front/lib/resources/skill/skill_resource.ts
+++ b/front/lib/resources/skill/skill_resource.ts
@@ -35,7 +35,6 @@ import { BaseResource } from "@app/lib/resources/base_resource";
 import type { ConversationResource } from "@app/lib/resources/conversation_resource";
 import { DataSourceViewResource } from "@app/lib/resources/data_source_view_resource";
 import { FileResource } from "@app/lib/resources/file_resource";
-import { verbsForGrantAtLevel } from "@app/lib/resources/group_permission_registry";
 import { GroupPermissionResource } from "@app/lib/resources/group_permission_resource";
 import { GroupResource } from "@app/lib/resources/group_resource";
 import { MCPServerViewResource } from "@app/lib/resources/mcp_server_view_resource";
@@ -72,6 +71,7 @@ import {
 import { formatTimestampToFriendlyDate } from "@app/lib/utils";
 import { concurrentExecutor } from "@app/lib/utils/async_utils";
 import { withTransaction } from "@app/lib/utils/sql_utils";
+import logger from "@app/logger/logger";
 import type {
   AgentConfigurationWithoutModelType,
   LightAgentConfigurationType,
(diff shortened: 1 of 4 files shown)


front/lib/resources/skill/skill_resource.ts
```diff
@@ -3123,9 +3181,15 @@ export class SkillResource extends BaseResource<SkillConfigurationModel> {
         );
 
         // Suspend all editor group memberships for this skill.
+        // Editors los permissions on the skill and only admin can keep their "admin" permission to unarchive it.
```
fabiencelier:
```suggestion
        // Editors lose permissions on the skill and only admin keep their "admin" permission to unarchive it.
```

front/lib/resources/skill/skill_resource.ts
```diff
@@ -638,6 +637,30 @@ export class SkillResource extends BaseResource<SkillConfigurationModel> {
       { transaction }
     );
 
+    // Dual write (see `writeEditorUserGrants`): the creator is also an editor through a per-user
+    // grant, not only through the skill_editors group. Best-effort — unlike the group above, the
+    // grant path requires a workspace membership row, and skill creation must not start failing for
+    // callers that lack one. Logged so the divergence is visible.
+    if (addCurrentUserAsEditor) {
+      const grantResult = await GroupPermissionResource.grantToUser(auth, {
```
matteotrab, line 645: The problem here is that we are creating a group which does not have the permissions we want. Currently for the skill editor group, its members can read, write and admin. Every role can read and admins can write and admin.

If we make it a regular_auto group, its members will only be able to read. Only admins will be able to read, write and admin. Users with role member and manager will not be able to read the group

fabiencelier: I may be missing something, but where do we need the permission to read/edit the member of the new regular_auto group ?

Listing the skill editors will be done via `skill_resource.listEditors` which will check read access to the skill (every body in the workspace) then will do something like this:
```
        const grantGroup =
          await GroupPermissionResource.findRegularAutoGroupForGrant(auth, {
            grantType: SKILL_EDITOR_GRANT_TYPE,
            resourceType: "skill",
            resourceId: this.id,
          });
        const grantEditors = grantGroup
          ? await grantGroup.getActiveMembers(auth)
          : []; 
```
There is no permissions check happening on the group membership as far as I understand here

Same for adding/removing an editor, it is done via the check that the user has "admin" permission on the skill and then we call 
```
await GroupPermissionResource.grantToUser(auth, spec)
```

We do all the permissions checks on the skills, at no point we actually do a check on the permissions of the regular_auto group itself, which IMHO is one of the purpose of regular_auto groups no ?

fabiencelier: discussed IRL: we don't need permissions on these regular_auto groups
these groups are used to back access to a resource, permissions are checked on the resource itself before accessing the group

tdraier (approved):

PR #30679, opened 2026-08-18
fix: chips never animate to visible without a LazyMotion ancestor

avervaet:
## Description

FilterSummaryChips uses framer-motion's m components, which stay in their initial state (opacity 0) until an ancestor LazyMotion provider loads the animation engine. Neither the automations page nor the consumption page (whose LazyMotion boundary only wraps the chart, not the sticky filter header above it) provided one, so the chips rendered invisible on both pages. Wrap the component's own animated tree in a LazyMotion boundary so it no longer depends on the caller.

## Tests

- Tested locally

## Risk

Low - UI changes only

## Deploy Plan

- Deploy front

### DIFF
--- a/front/components/workspace/analytics/filterPanel/FilterSummaryChips.tsx
+++ b/front/components/workspace/analytics/filterPanel/FilterSummaryChips.tsx
@@ -1,6 +1,12 @@
 import type { FilterSummary } from "@app/components/workspace/analytics/filterPanel/filterState";
 import { Button, Chip } from "@dust-tt/sparkle";
-import { AnimatePresence, m, useReducedMotion } from "framer-motion";
+import {
+  AnimatePresence,
+  domMax,
+  LazyMotion,
+  m,
+  useReducedMotion,
+} from "framer-motion";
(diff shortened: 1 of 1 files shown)


avervaet:

- #30679 ◀

adrsimon (approved):

front/components/workspace/analytics/filterPanel/FilterSummaryChips.tsx
```diff
@@ -38,61 +44,67 @@ export function FilterSummaryChips<Category extends string>({
...
-                  exit={
-                    shouldReduceMotion ? undefined : { opacity: 0, scale: 0.96 }
-                  }
-                  transition={transition}
-                  className="max-w-full"
-                >
-                  <Chip
-                    size="xs"
-                    color="highlight"
+    <LazyMotion features={domMax}>
```
aubin-tchoi, line 47: nit: do we really need `domMax`? shouldn't domAnimation be enough?

aubin-tchoi (approved):
LGTM

PR #30680, opened 2026-08-18
Recover the staging directory when a marker is left half written

flvndvd:
## Description

Startup refuses to use a staging directory that holds files but no marker, which is what keeps the daemon from cleaning a private directory that was passed by mistake. The marker itself was written through a randomly named temporary file and then renamed into place, so a daemon that died between those two steps left a file behind and no marker. Every later startup then refused that directory, and since the supervisor restarts the mount forever, the sandbox lost its filesystem until someone deleted the leftover by hand. The comment on the marker write claimed startup already knew how to clean that file, which was not true on the path that matters.

The marker is now written under a fixed name of its own rather than a random one, so startup can tell the one file a half-written marker leaves apart from files it must never touch. When no marker is present, that single name is now accepted and removed, and anything else still fails closed. Staged content and downloads keep their own names and stay refused without a marker, which is correct because both only appear after a startup that did write the marker.

Removal now happens before the marker is written, since the write reuses the same name that a half-written marker leaves behind.

Three cases are covered by tests: recovering from a half-written marker, refusing an unmarked directory that holds staged content or a download, and refusing an unmarked directory that holds an unrelated file while leaving that file in place. Beyond the unit tests, I mounted the daemon against a stand-in Front on a real kernel mount and checked all four situations end to end: a fresh directory, one already claimed, one left mid-claim by a crash, and one belonging to someone else.

Reported in review by @smb2268.

### DIFF
--- a/cli/dust-sandbox/src/commands/filesystem/store/content_cache.rs
+++ b/cli/dust-sandbox/src/commands/filesystem/store/content_cache.rs
@@ -14,7 +14,6 @@ use std::path::{Path, PathBuf};
 use std::sync::{Arc, Mutex, MutexGuard};
 
 use lru::LruCache;
-use tempfile::NamedTempFile;
 use tracing::warn;
 
 use super::{errno, is_writable, INodeNo, Node};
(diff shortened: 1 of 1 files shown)


flvndvd:
PMRR.

smb2268:
wahoo, lgtm. thanks for fixing!

PR #30684, opened 2026-08-18
Stop Metronome 429 amplification from the members-usage read paths

PopDaph:
## Context

`/api/w/:wId/trial/start` started failing with `Failed to provision Metronome contract: 429 Rate limit exceeded`, leaving new workspaces without a contract or subscription row. All Metronome calls share one API key, so its rate limit is a global budget: the provisioning writes were being starved by read traffic.

The dominant source (95 of the top 100 Metronome-429 log lines over 24h, plus ~5K "falling back to uncached fetch" warnings over 2 days) is the seat-data read path (`buildSeatDataByUserId`: 1 contract retrieve + one `getSubscriptionSeatsHistory` per SEAT_BASED subscription). It is hit by the admin members table and by `/credits/my-usage`, which the sidebar user menu fetches for every user of every credit-priced workspace.

Under a 429 brownout, one request amplified into up to `(1 + N subscriptions) x 6 SDK attempts x 2` Metronome calls:
- the cached loader throws on failure, so nothing is cached and every request retries;
- the cache used the in-process lock only, so every pod (and every queued in-process waiter after a failure) fired its own fan-out;
- on failure, `members_usage.ts` caught the error and immediately re-ran the same fan-out uncached;
- each HTTP call was retried up to 5 times by the SDK.

## Changes

- **Remove the four catch-and-refetch-uncached fallbacks** in `members_usage.ts` (seat data, per-user usage, per-user cap alert ids, default caps by seat type). On a cache-loader failure each path now logs and degrades to an empty result; all consumers already treat these fields as nullable.
- **Fleet-wide fetch dedup on the seat-data cache**: `getCachedSeatDataByUserId` now uses `useDistributedLock: true, skipIfLocked: true`, so at most one seat fan-out is in flight per contract across all pods. Concurrent misses get `null` and the caller degrades for that request.
- **Fail fast on degradable reads**: `getMetronomeContractById` and `getMetronomeSubscriptionSeatState` accept a per-request `maxRetries`; `buildSeatDataByUserId` (which only feeds display surfaces) passes `maxRetries: 1`. Writes and bulk jobs keep the client-level 5 retries.

Net effect during a brownout, per workspace per minute fleet-wide: a single `(1 + N) x 2` fan-out instead of the unbounded storm; everything else serves degraded seat metadata instead of burning the budget contract provisioning needs.

Trade-off: in healthy operation, when the 60s cache expires, requests racing the single refill briefly see seat fields as unavailable (previously each pod refetched). Seat fields are nullable everywhere and refresh on the next request.

## Tests

- `members_usage.test.ts` (new): loader failure and lock-skip both degrade to an empty map without an uncached refetch; happy path returns the cached data.
- `seats.test.ts`: pins the cache registration options (distributed lock + skipIfLocked) and that loader failures propagate instead of returning cacheable empty data. The lock/skip runtime semantics are covered by `lib/utils/cache.test.ts` (vite.setup.ts stubs `cacheWithRedis` in unit tests).

## Deploy Plan

Deploy front.

### DIFF
--- a/front/lib/api/credits/members_usage.ts
+++ b/front/lib/api/credits/members_usage.ts
@@ -15,10 +15,6 @@ import type {
 import {
   getCachedDefaultCapThresholdsBySeatType,
   getCachedPerUserCapAlertIds,
-  getMetronomeDefaultUserCapAlertForSeatType,
-  getMetronomeDefaultUserWarningAlertForSeatType,
-  listMetronomePerUserCapsForWorkspace,
-  listMetronomePerUserWarningAlertsForWorkspace,
 } from "@app/lib/metronome/alerts/spend_limits";
 import type { MetronomeAlertRef } from "@app/lib/metronome/alerts/types";
 import {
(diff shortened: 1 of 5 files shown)


front/lib/api/credits/members_usage.ts
```diff
@@ -564,40 +530,17 @@ async function fetchPerUserUsageCreditsForMembersTable({
...
-}): Promise<Map<string, SeatData>> {
-  const seatDataResult = await buildSeatDataByUserId({
-    metronomeCustomerId,
-    contractId: metronomeContractId,
-  });
-  if (seatDataResult.isErr()) {
-    logger.warn(
-      { err: seatDataResult.error, metronomeCustomerId },
-      "[MembersUsage] Failed to build seat data, degrading to empty map"
+      "[MembersUsage] Failed to read cached per-user usage, degrading to empty map"
```
PopDaph, line 536: This data is display-only (members table, usage widget) and all fields are already nullable:
on failure the UI shows "unavailable" and recovers at the next cache fill. Spend enforcement
does not read it: message blocking uses the Redis counter seeded from Elasticsearch.

iirc the fallback also never helped: on Metronome failure the uncached fetch failed too and returned empty data anyway. All it did was double the Metronome calls exactly when Metronome was already rate-limited.

PopDaph (commented):
Ceinture et bretelles?

tdraier (approved):

PR #30686, opened 2026-08-18
Fix the filesystem mount failing to start, and a few other problems found by running it

flvndvd:
## Description

This PR fixes seven problems in the sandbox filesystem daemon. The first one stops it working at all, and it was found by mounting the daemon and using it. The rest came out of thorough review and playing around. Each one is explained below with what a user would have seen, because several of them are hard to grasp from the code.

### The mount could not start

The daemon crashed on startup every time, on every machine, so the sandbox filesystem never appeared. The supervisor restarted it in a loop and the sandbox simply had no files.

The cause is a rule in the blocking HTTP client we use to call Front: it cannot be built while the program is inside its async runtime, and it stops the process if it is. Every command in this tool runs inside that runtime, so the filesystem command always hit the rule.

The filesystem command now runs on a thread of its own, outside that runtime, and starts the runtime it needs for Front calls itself. That thread is also the one the mount occupies for as long as it lives, which it could not do on a shared runtime thread anyway.

### Files appeared to change while they were being read

Archiving, copying or syncing a file shortly after writing it failed. Both `tar` and `rsync` stop with "file changed as we read it", so an agent that wrote files and then packed them would see an error, and `tar` would exit non-zero.

While a file is being written, its reported time comes from the local copy, because nothing has been saved yet. Linux remembers file details for a second and answers from that memory. Saving the file to Front then gives it a slightly different time, and nothing told Linux to refresh, so a program reading across that moment saw the time change underneath it.

Once a save completes, the daemon now asks Linux to drop what it remembers for that file, so the next question returns the saved time. This is the standard mechanism for exactly this situation.

### Writing to a file while it was being saved returned an error

A program that kept writing while an earlier save was still uploading got told to try again later. Almost no program retries that on an ordinary file write, so it surfaces as a hard write failure. Reading the file's size had the same problem.

Saving holds the file for the length of the upload, which is what keeps the uploaded bytes consistent with the size we promised Front. The daemon refused to wait for that instead of waiting for it. A write now waits and then proceeds, which is what the same program would see on a local disk. Asking for the size no longer waits at all, because that answer comes from what is already staged.

### The two short paths (`/conversation` and `/pod`) were not real links

Linux identifies every file and directory by a number, and it keeps a single name for a directory with a given number. The short conversation and pod paths were being served as a second name pointing at the same number as the real directory, rather than as links, which is not a way to give a directory two working names: Linux moves the directory to the name most recently looked up instead of showing it under both.

Both short paths are now real symbolic links with numbers of their own. Removing or renaming them is refused, since they belong to the mount layout rather than to the sandbox.

### Setting a file's timestamp was refused, which broke unpacking archives

Any attempt to set a file's time failed. It was implemented this way as Front does not store any of those for now. That sounds harmless, but `tar` sets a time on every file it extracts, and so do copying with preserved attributes, `rsync`, and `touch`. Extracting any archive into the sandbox (only in the `/files` mounted directory) therefore failed on the first file, which is a common thing for an agent to do.

Dust records when a file's contents last changed and has nowhere to store a time chosen by the caller. Those requests are now accepted and dropped, so the tools work, and the reply carries the time Dust does hold. This is a deliberate product choice and is written down as a known limit.

### Setting a file's owner was refused even when nothing would change

Extracting an archive as root failed for a second reason: `tar` restores ownership, and every request was refused, including asking for the owner the file already had. Asking for the current owner changes nothing, so it is now accepted, while asking for a different owner is still refused.

### A save whose confirmation was lost reported a false failure

If the reply to a save was lost on the way back, the client sent the same save again, and Front correctly answered that the version it was given was out of date. The daemon reported that as a failure even though the bytes were safely stored, so a program saw its save or close fail and the local copy was thrown away.

Every upload gets its own identifier, so the daemon now reads the file back: if it already holds the version just uploaded, the save was its own and succeeded. Any other version means another sandbox wrote first, which stays an error the caller must see.

### 🚗 Smaller changes

A helper for building errors was duplicated in five places and is now defined once. The worker pool module moved out of the Linux-only part of the tree, since it uses no Linux facilities and was being included through a workaround so its tests would run elsewhere. Uploads now state their length directly instead of relying on a header that Front happens to sign, so the request no longer depends on that coincidence to avoid being sent in chunks. 

### 💡 How this was tested

Most of these are impossible to catch with unit tests alone, so this adds a harness that mounts the daemon against a stand-in Front on any Linux machine where FUSE is available. It needs no sandbox, no database and no cloud storage, and it can run the existing acceptance script end to end. That harness is what found the startup crash and the changing timestamps, so it is included here rather than kept locally.

The acceptance script gains cases for the short paths being real links, an archive round trip with preserved attributes, a second writer being refused, writing while a save is in flight, and a file's time staying put after it is saved. The unit tests cover the link resolution, the recovered save, and the rejected one.

Everything was checked on a real mount: the acceptance script passes, the mount checks pass, and all four ways of writing a file now report a stable time with tar succeeding. Formatting and lint are clean, and the tests pass on both Linux and macOS.

### What this deliberately does not fix (follow up PRs coming)

Changing a file's read or write permission bits is still refused, so extracting an archive created on another filesystem can still fail on its modes even though its times now work. That is the same family as the timestamp decision above.

Removing and renaming still address files by name rather than by identifier, so a file another sandbox moved onto that name in the same instant could be affected instead. Fixing it needs support in Front. 

## Tests

We can now run it locally with:
```
x/flav/dust-filesystem-harness/run.sh mount_checks.sh
```


cli/dust-sandbox/src/commands/filesystem/mod.rs
```diff
@@ -25,11 +26,27 @@ pub enum FilesystemCommand {
 }
 
 pub fn run(command: FilesystemCommand) -> anyhow::Result<()> {
-    match command {
+    // `dsbx` runs its commands inside a Tokio runtime, and the filesystem
+    // cannot start there: the blocking HTTP client refuses to be built while a
```
flvndvd, line 30: It's an actual reqwest implementation.

aubin-tchoi (approved):
LGTM

PR #30687, opened 2026-08-18
feat: move automations page to /automations and tidy its table

adrsimon:
## Description

The automations page is a top-level admin destination, not an analytics sub-page, so its URL loses the `/analytics` prefix and its title goes plural. In the triggers table: smaller agent avatar, the full Type value on hover, and a narrower centered Enabled column.

Old `/analytics/automations` URLs now 404. Skipped a redirect since the page is behind `enable_analytics_automations`.

Capture d’écran 2026-08-18 à 13 37 07

## Tests

Existing analytics component tests pass (50).

## Risk

Low, UI only.

## Deploy Plan

Deploy front.

### DIFF
--- a/front-spa/src/app/routes/adminRoutes.tsx
+++ b/front-spa/src/app/routes/adminRoutes.tsx
@@ -133,7 +133,7 @@ export const adminRoutes: RouteObject[] = [
     element: <RequireRoleLayout requiredRole="admin" />,
     children: [
       {
-        path: "analytics/automations",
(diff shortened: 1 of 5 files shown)


flvndvd (approved):
Should we also remove it from the `/analytics` folder?

adrsimon:
@flvndvd yes but will do this at the end, i still have a PR or two coming, don't want to rebase everything

PR #30689, opened 2026-08-18
[admin governance] Create group resource type

ykmsd:
## Description
This PR is the first mini PR for us to manage groups in group permission table. 

### How does it work? 
For example, when you create a skill, you create two resources: Skill and Group (editors).
Let's say skill id is 123, and editor group id is 456. What we want to store in group permissions table is:
- Permission for the skill, so something like: `groupId: 456, resource type: "skill", resource id: 123, grant type: "editor"`, meaning that the member of groupId 456 can read/write/admin the skill id 123. Everyone in the workspace should be able to read the skill so we need to create `groupId: GLOBAL_ID, resource type: "skill", resource id: 123, grant type: "reader"`,
- Permission for group, because we need a separate permission for controlling permission for groups, something like  `groupId: 456, resource type: "group", resource id: 456, grant type: "admin"`. And same for global group to be able to read it `groupId: GLOBAL_ID, resource type: "group", resource id: 456, grant type: "reader"`. 

The next step is whenever we create a new group, we should store permission in tables. Once we cover all the different path for groups (we have lots of different groups), we can run backfill scripts, make sure it stays the same as before, and then remove the old path.

### DIFF
--- a/front/lib/resources/group_permission_registry.ts
+++ b/front/lib/resources/group_permission_registry.ts
@@ -84,6 +84,10 @@ export const ROLE_REGISTRY: Record<
   dust_app: {
     admin: { verbs: ["admin"], levels: ["type"] },
   },
+  group: {
+    reader: { verbs: ["read"], levels: ["instance"] },
+    editor: { verbs: ["read", "write", "admin"], levels: ["instance"] },
+  },
 };
 
 interface GrantSpec {
--- a/front/lib/resources/group_permission_registry.ts
+++ b/front/lib/resources/group_permission_registry.ts
@@ -86,7 +86,7 @@ export const ROLE_REGISTRY: Record<
   },
   group: {
     reader: { verbs: ["read"], levels: ["instance"] },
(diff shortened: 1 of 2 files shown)


front/lib/resources/group_permission_registry.ts
```diff
@@ -82,6 +84,10 @@ export const ROLE_REGISTRY: Record<
   dust_app: {
     admin: { verbs: ["admin"], levels: ["type"] },
   },
+  group: {
+    reader: { verbs: ["read"], levels: ["instance"] },
+    admin: { verbs: ["read", "write", "admin"], levels: ["instance"] },
```
ykmsd, line 89: I'm slightly confused with the terminology difference between `editors` and `admins` but I thiiiiin k here it makes more sense to be `admin`?

fabiencelier: admin make sense
When you change who is inside a group you administrate it, you don't edit it

fabiencelier (approved):

PR #30690, opened 2026-08-18
Enable the Dust filesystem behind a feature flag

flvndvd:
## Description

This PR connects the new database-backed filesystem to Front and sandbox startup while keeping the current GCS filesystem as the default.

Nothing changes unless the `dust_filesystem` feature flag is enabled and a fresh Pod or standalone conversation explicitly selects the new filesystem. Existing Pods and conversations continue using GCS. There is no data migration and no new
database field.

For this first rollout, a Pod opts in through the `[Dust FS]`  name prefix ]. The prefix can only be selected when the workspace has the feature flag, and it cannot be added or removed after creation (yeah not ideal, but it saves us from altering the data model of project metadata).

A fresh standalone conversation can opt in with `metadata.useDatabaseFileSystem`. Pod conversations cannot choose independently because they inherit their Pod’s filesystem. Forks and `run_agent` child conversations keep the choice made by their standalone parent.

Front now selects one of two complete backends behind the existing `DustFileSystem` interface. The GCS backend keeps today’s behavior. The database backend uses filesystem nodes for names and identity, while immutable file bytes remain in GCS. Reads, writes, directory creation, deletion, copy, move and download URLs therefore continue using the same application-facing API.

A database-backed move updates the node’s parent and name instead of copying bytes and deleting the source. This preserves the inode when a file moves between the conversation and Pod roots.

A filesystem instance cannot mix GCS and database-backed roots. We also prevent moving conversations into or out of a database-backed Pod for now, since that would require migrating their existing files between the two stores.

For sandbox use, the new adapter writes a short-lived scoped token and starts the Rust daemon under systemd at `/files`. Token refresh does not restart the mount. Startup fails clearly if the sandbox image does not contain the `dust_filesystem` capability. Pod directories that still require GCS remain mounted through the existing adapter.

A final sandbox image change will ship the matching `dsbx` daemon before we enable the flag.

### DIFF
--- a/front/lib/api/file_system/backends/database_file_system_backend.ts
+++ b/front/lib/api/file_system/backends/database_file_system_backend.ts
@@ -1,0 +1,622 @@
+import { randomUUID } from "node:crypto";
+import type { Readable } from "node:stream";
+
+import type { FileSystemBackend } from "@app/lib/api/file_system/backends/file_system_backend";
+import {
+  getFileSystemDownloadUrl,
+  getFileSystemReadStream,
+  writeFileSystemContent,
+} from "@app/lib/api/file_system/file_system_content";
+import { FileSystemScope } from "@app/lib/api/file_system/namespace_scope";
+import { FileSystemOperationError } from "@app/lib/api/file_system/namespace_types";
+import { DatabaseSandboxMountAdapter } from "@app/lib/api/file_system/sandbox/database_sandbox_mount_adapter";
+import type { SandboxMountAdapter } from "@app/lib/api/file_system/sandbox/sandbox_mount_adapter";
+import type { Authenticator } from "@app/lib/auth";
+import { FILE_SYSTEM_CONTENT_URL_EXPIRATION_MS } from "@app/lib/file_storage/file_system_blobs";
+import { FileSystemMutationResource } from "@app/lib/resources/file_system_mutation_resource";
+import { FileSystemNodeResource } from "@app/lib/resources/file_system_node_resource";
+import type {
+  FileSystemDirectoryEntry,
+  FileSystemEntry,
+} from "@app/types/api/file_system/types";
+import type { FileSystemMount, SandboxOnlyMount } from "@app/types/file_system";
+import { DustFileSystemError } from "@app/types/file_system";
+import { stripMimeParameters } from "@app/types/files";
+import { TOOL_OUTPUTS_FOLDER_NAME } from "@app/types/mount_path";
+import type { Result } from "@app/types/shared/result";
+import { Err, Ok } from "@app/types/shared/result";
+import { normalizeError } from "@app/types/shared/utils/error_utils";
+
+type ParsedPath = {
+  mount: FileSystemMount;
+  segments: string[];
+};
+
+type ResolvedPath = ParsedPath & {
+  node: FileSystemNodeResource;
+};
+
+const DIRECTORY_CONTENT_TYPE = "application/x-directory";
+const DEFAULT_CONTENT_TYPE = "application/octet-stream";
+const DIRECTORY_PAGE_SIZE = 256;
+
+/** PostgreSQL owns names and inodes; GCS stores only immutable file bytes. */
+export class DatabaseFileSystemBackend implements FileSystemBackend {
+  private readonly scope: FileSystemScope;
+  private rootsPromise: Promise<FileSystemNodeResource[]> | null = null;
+
+  constructor(
+    private readonly auth: Authenticator,
+    private readonly mounts: ReadonlyArray<FileSystemMount>,
+    private readonly sandboxOnlyMounts: ReadonlyArray<SandboxOnlyMount> = []
+  ) {
+    this.scope = new FileSystemScope(
+      mounts.flatMap((mount) =>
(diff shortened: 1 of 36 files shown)


front/lib/api/file_system/backends/database_file_system_backend.ts
```diff
@@ -0,0 +1,622 @@
...
+type ResolvedPath = ParsedPath & {
+  node: FileSystemNodeResource;
+};
+
+const DIRECTORY_CONTENT_TYPE = "application/x-directory";
+const DEFAULT_CONTENT_TYPE = "application/octet-stream";
+const DIRECTORY_PAGE_SIZE = 256;
+
+/** PostgreSQL owns names and inodes; GCS stores only immutable file bytes. */
+export class DatabaseFileSystemBackend implements FileSystemBackend {
```
flvndvd, line 44: This implementation is not very efficient but it maps to the existing API we have which let us test without changing any of the FS operations on the client/application-side. We will do a bigger rework if we move forward with our minimal daemon.

front/lib/api/file_system/storage_mode.ts
```diff
@@ -0,0 +1,25 @@
+import type { SpaceResource } from "@app/lib/resources/space_resource";
+import type { ConversationWithoutContentType } from "@app/types/assistant/conversation";
+
+export type FileSystemStorageMode = "database" | "gcs";
+
+/** Prefix used to opt a fresh Pod into the database-backed filesystem. */
+export const DATABASE_FILE_SYSTEM_POD_PREFIX = "[Dust FS] ";
```
aubin-tchoi, line 7: could be annoying to rely on a prefix no?

front/lib/api/file_system/backends/database_file_system_backend.ts
```diff
@@ -0,0 +1,622 @@
...
+      await this.removeNode(destination.parent.id, destination.existing);
+      return new Ok(undefined);
+    } catch (error) {
+      return new Err(this.error(error));
+    }
+  }
+
+  private async copyNode(
+    source: FileSystemNodeResource,
+    destinationParent: FileSystemNodeResource,
```
aubin-tchoi, line 476: no guard that source and dest are different?

front/lib/api/file_system/sandbox/database_sandbox_mount_adapter.ts
```diff
@@ -0,0 +1,229 @@
...
+
+  private async writeToken(
+    auth: Authenticator,
+    sandbox: SandboxResource
+  ): Promise<Result<void, Error>> {
+    const token = await generateSandboxFileSystemToken(auth, {
+      sandbox,
+      roots: this.tokenRoots(),
+    });
+    const temporaryPath = `${FILE_SYSTEM_DIRECTORY}/.token-${randomBytes(8).toString("hex")}`;
```
aubin-tchoi, line 56: nit: we never delete it?

front/lib/api/file_system/sandbox/database_sandbox_mount_adapter.ts
```diff
@@ -0,0 +1,229 @@
...
+import { rootCommand } from "@app/lib/api/sandbox/root_command";
+import type { Authenticator } from "@app/lib/auth";
+import fileStorageConfig from "@app/lib/file_storage/config";
+import type { SandboxResource } from "@app/lib/resources/sandbox_resource";
+import logger from "@app/logger/logger";
+import type { FileSystemMount, SandboxOnlyMount } from "@app/types/file_system";
+import type { Result } from "@app/types/shared/result";
+import { Err, Ok } from "@app/types/shared/result";
+
+import type { SandboxMountAdapter } from "./sandbox_mount_adapter";
```
aubin-tchoi, line 18: nit: could be an absolute import

front/lib/api/file_system/backends/database_file_system_backend.ts
```diff
@@ -0,0 +1,622 @@
...
+        name: node.name,
+        kind: node.kind,
+      }
+    );
+    if (removed.isErr()) {
+      throw removed.error;
+    }
+  }
+
+  async delete(
```
aubin-tchoi, line 451: nit: in the naming there's a mix of Unix command names and more human readable names (mv versus move)

aubin-tchoi (approved):
LGTM

front/lib/api/file_system/storage_mode.ts
flvndvd: Yeah but see in PR description less intrusive way to test it in prod on our workspace. Obviously if we move forward, will be more baked in.

front/lib/api/file_system/backends/database_file_system_backend.ts
flvndvd: This is handled by the `FileSystemNodeResource` itself. This brick must remains dead simple.

front/lib/api/file_system/sandbox/database_sandbox_mount_adapter.ts
flvndvd: No, we never delete those token files. They are being rotated, re-upserted.

PR #30691, opened 2026-08-18
[front] enh: Add fair-use limit on premium model for legacy plans

id13:
## Description

Add rate limiting for legacy plans usage of premium models: 25/users/week.

The counting & flagging will be live at deploy, the blocking is behind a feature flag.

Caveats:

- for sub-agent runs, we rely on the fact that we pass an email header along with system key and reconstruct auth object with user(). so they will count toward the limit
- triggers / wake-ups do too (unless under programmatic)
- programmatic is bypassed

This code should be cleaned up once we have switched all legacy plans to CP

## Tests

Added tests (system key email matching, new fair limit, etc)

## Risk

Low
Blast radius: Whole agent loop

## Deploy Plan

Deploy front

### DIFF
--- a/front/lib/api/assistant/premium_model_limit.ts
+++ b/front/lib/api/assistant/premium_model_limit.ts
@@ -1,0 +1,119 @@
+import {
+  makePremiumModelMessageRateLimitKeyForUser,
+  PREMIUM_MODEL_MESSAGE_RATE_LIMIT_PER_USER_PER_WEEK,
+  PREMIUM_MODEL_MESSAGE_RATE_LIMIT_WINDOW_SECONDS,
+} from "@app/lib/api/assistant/rate_limits";
+import { isProgrammaticUsage } from "@app/lib/api/programmatic_usage/tracking";
+import type { Authenticator } from "@app/lib/auth";
+import { getFeatureFlags } from "@app/lib/auth";
+import { isFreeOrigin } from "@app/lib/credits/agent_message_billing";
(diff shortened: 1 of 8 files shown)


id13:
PMRR

front/lib/api/assistant/premium_model_limit.ts
```diff
@@ -0,0 +1,119 @@
...
+  );
+  if (tierName !== "premium") {
+    return new Ok(undefined);
+  }
+
+  const key = makePremiumModelMessageRateLimitKeyForUser(
+    workspace,
+    user.toJSON()
+  );
+  const countResult = await getRateLimiterCount({
```
flvndvd, line 59: This does not enforce Monday-thru Sunday, right? it's more some kind of sliding window?

flvndvd, line 59: Can we do an atomic update? I know that H1 had many report submitted as we could burst messages between the check and the increment.

flvndvd (commented):
I guess I forgot to mention but rather than blocking can we downscale to a lower model? A one:one mapping would be enough here.

front/lib/api/assistant/premium_model_limit.ts
id13: Done

id13: Yes, too complex, I chose 80/20

PR #30692, opened 2026-08-18
feat: search automations by name in the analytics table

adrsimon:
## Description

This adds a search input on the automations page, on the same line as the filters.

Ranking, ordering and the total count stay in Elasticsearch. The search resolves to a set of trigger ids with an ILIKE (`listByWorkspaceAndNameSearch`), the way the kind filter already resolves its own through `listByWorkspaceAndKinds`, and applies to the ranked buckets rather than to the query.

Keeping the filtering out of the query allows to keep the whole data from the filter for the line details. Without it, we would have line details (How often it runs, etc...) compared to the triggers matching the query, which is not what we want.

Capture d’écran 2026-08-18 à 16 44 25

## Tests

`front/lib/api/analytics/automations/triggers.test.ts`: a search narrows the rows while stats stay equal to the unsearched call.

## Risk

Low. The unsearched path is unchanged.

## Deploy Plan

Deploy front.

### DIFF
--- a/front/lib/api/analytics/automations/triggers.ts
+++ b/front/lib/api/analytics/automations/triggers.ts
@@ -107,56 +107,41 @@ function median(values: number[]): number {
     : sorted[mid];
 }
 
-type TriggerRestrictions = {
-  // Applied as a terms filter on the query, so it scopes the ranking and the
-  // median baseline alike. Null means no restriction.
-  kindIds: string[] | null;
-  // Applied to the ranked buckets instead, so a search narrows the rows
-  // without moving the median baseline. Null means no restriction.
-  searchIds: Set<string> | null;
-};
+/**
+ * Trigger kind isn't indexed in the consumption Elasticsearch documents, so
+ * a kind filter is resolved to a concrete set of trigger ids up front and
+ * applied as a terms filter on TRIGGER_ID_FIELD. Returns null when no kind
+ * filter is requested (no restriction).
+ */
+async function resolveTriggerIdsForKindFilter(
+  auth: Authenticator,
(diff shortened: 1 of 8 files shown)


front/lib/api/analytics/automations/triggers.ts
```diff
@@ -106,21 +106,56 @@ function median(values: number[]): number {
...
-  const triggers = await TriggerResource.listByWorkspaceAndKinds(auth, kinds);
-  return triggers.map((trigger) => trigger.sId);
+): Promise<TriggerRestrictions> {
+  const kindSet = kinds?.length ? new Set(kinds) : null;
+  const normalizedSearch = search?.trim().toLowerCase();
+  if (!kindSet && !normalizedSearch) {
+    return { kindIds: null, searchIds: null };
+  }
+
+  const triggers = await TriggerResource.listByWorkspace(auth);
```
aubin-tchoi: let's still do `listByWorkspaceAndKinds` to filter early on
we could even bake in the search in the sql query but probably won't help a lot

front/lib/api/analytics/automations/triggers.ts
```diff
@@ -219,11 +254,19 @@ async function fetchTriggersRanking(
   // run cost" stat against, so the baseline only looks at the ones that did.
   const activeRanked = ranked.filter((r) => r.runCount > 0);
 
+  const searched =
+    searchIds !== null
+      ? ranked.filter((r) => searchIds.has(r.triggerId))
```
aubin-tchoi: filtering now this will mess up with your pagination

aubin-tchoi (approved):
LGTM

PR #30696, opened 2026-08-18
fix: scope consumption export status to the requested period+filter

avervaet:
## Description

The consumption export panel previously determined `isGenerating` by scanning any running export workflow for the workspace, and treated any historical export as available. As a result, an export generated for one period or filter could suppress or mask the export needed for a different selection.

This change scopes export status to the exact period+filter cache key used by the export workflow. 
The endpoint contract changes from the previous workspace-wide `GET /export-raw` status check to the scoped `POST /export-raw/status` check; the callers in this PR are updated together.

## Risk

Low

## Deploy Plan

- Deploy front


avervaet:

- #30696 ◀
    - #30703
        - #30705

flvndvd (approved):

PR #30697, opened 2026-08-18
[front] fix: hide burn-up target when consumption is filtered

adrsimon:
## Description

The credit cap covers the whole workspace, so drawing the target against a filtered part of the credits consumption made every filtered view look far under target. This PR drops the target line (and its legend entry) as soon as a scope filter is active.

## Tests

Manual: filtered by agent and by user on a cycle period, target line disappears. Cleared the filter, it comes back.

## Risk

Low, one chart.

## Deploy Plan

Deploy front.


aubin-tchoi (approved):
LGTM

PR #30700, opened 2026-08-18
refactor(front): inline back unecessary refactors

achilleburah:
## Description

Revert some unecessary refactors.

## Tests

Built and ran locally

## Deploy Plan

Classic. Revert to rollback.


front/lib/resources/skill/skill_resource.ts
```diff
@@ -1642,7 +1639,14 @@ export class SkillResource extends BaseResource<SkillConfigurationModel> {
   static async listByDataSourceViewIds(
     auth: Authenticator,
     dataSourceViewIds: ModelId[],
```
aubin-tchoi, line 1641: unrelated but good to know: we usually name those `dataSourceViewModelIds`, `xxxId` refers to the string identified

aubin-tchoi (approved):
LGTM 🙏🏻

front/lib/resources/skill/skill_resource.ts
achilleburah: alright noted ! thx

PR #30703, opened 2026-08-18
perf: fetch consumption export slices concurrently instead of sequentially

avervaet:
## Description

The raw-lines CSV export previously paginated the full result set sequentially with `search_after`, so wall-clock time grew with the total row count. This change partitions the query into three Elasticsearch slices and fetches the slices concurrently, while each slice continues to paginate independently with `search_after`.

The export opens one shared point-in-time (PIT) so all slices and pages read from the same index snapshot, propagates refreshed PIT IDs between rounds, merges the individually sorted results back into the existing global order, and closes the PIT when the export finishes. The activity tests were extended to cover PIT lifecycle and cleanup, sliced pagination, row uniqueness/order, and PIT ID refresh behavior.

## Risk

The main risk is higher peak memory usage: concurrent slices can keep one in-flight 10,000-row page per slice, in addition to the accumulated export rows. **Warning:** this is currently the only part of the codebase where PIT is used, so we lack production data about the potential memory pressure of this approach. The shared PIT provides a consistent snapshot across slices, and cleanup runs in `finally`; failures to close the PIT are logged.

## Deploy Plan

- Deploy front

### DIFF
--- a/front/lib/api/analytics/consumption/export_lines.ts
+++ b/front/lib/api/analytics/consumption/export_lines.ts
@@ -10,6 +10,7 @@ import type { ElasticsearchError } from "@app/lib/api/elasticsearch";
 import { searchConsumptionAnalytics } from "@app/lib/api/elasticsearch";
 import type { Authenticator } from "@app/lib/auth";
 import { microCreditsToCredits } from "@app/lib/credits/units";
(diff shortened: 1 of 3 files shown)


avervaet:

- #30696
    - #30703 ◀
        - #30705

PR #30705, opened 2026-08-18
perf: stream the consumption export as gzip-compressed CSV instead of buffering a zip

avervaet:
The raw-lines export collected every ES document, built every CSV row, and
zipped the whole thing into one Buffer before uploading, so memory usage grew
with the size of the export. AdmZip has no streaming API, so switch the
container format to .csv.gz: rows are gzip-streamed directly to the GCS
write stream as each page is fetched, with per-page label resolution (cached
across pages) instead of resolving labels once over the full document set.
Memory now stays bounded by the number of in-flight ES pages rather than the
export's total size.


avervaet:

- #30696
    - #30703
        - #30705 ◀

PR #30711, opened 2026-08-18
[front] Read space membership from governance grants, not the group list

tdraier:
Stacked on #30707.

## What

Rewrites `SpaceResource.isMember(auth)` and `SpaceResource.listWorkspacePodsAsMember(auth)` to read from the caller's resolved space grants — via `auth.getGrantedVerbs("space", id)` and the new `auth.getResourceIdsWithVerb("space", verb)` (backed by `GroupPermissions.resourceIdsWithVerb` from #30707) — instead of raw group membership (`auth.hasGroupByModelId` / `auth.groupModelIds`). This removes two consumers of the `Authenticator`'s group list.

## Membership verb per space kind

The grants `spaceGroupRoles` writes mirror the legacy `isMemberByGroupPredicate`, so the check is per-kind:

| kind | membership |
|------|------------|
| `global` | always a member |
| `system`, `conversations` | never a member |
| `regular` | holds `read` — open spaces grant every member `reader`, restricted grant members `member` |
| `project` | holds `write` — the workspace global group is only a `reader` viewer on unrestricted projects, so `write` marks an actual member (editor `admin` / member `member`) |

Using `read` for projects would be wrong: it would treat every workspace member as a project member via the global viewer grant. That's the one subtlety here.

## Notes

- `listWorkspacePodsAsMember` now selects projects the caller holds `write` on and filters `kind: "project"`, dropping the `GroupSpaceModel` member/editor lookup and its redundant safety re-filter + logging.
- `isMemberByGroupModelIds` (bulk other-user checks against an explicit group-id set) is intentionally left unchanged — it isn't part of the Authenticator group-list removal.
- Safe in prod: `space` grants have been backfilled and dual-written live since the group_permissions flip (2026-07-30).

## Test

- `npx tsgo --noEmit` clean. Relying on CI for the functional space/project suites.

### DIFF
--- a/front/lib/resources/space_resource.ts
+++ b/front/lib/resources/space_resource.ts
@@ -24,7 +24,6 @@ import type { ResourceFindOptions } from "@app/lib/resources/types";
 import { UserResource } from "@app/lib/resources/user_resource";
 import { concurrentExecutor } from "@app/lib/utils/async_utils";
 import { withTransaction } from "@app/lib/utils/sql_utils";
-import logger from "@app/logger/logger";
 import tracer from "@app/logger/tracer";
 import type { GrantVerb } from "@app/types/group_permissions";
 import type { GroupKind, GroupType } from "@app/types/groups";
@@ -46,7 +45,6 @@ import { assertNever } from "@app/types/shared/utils/assert_never";
 import { removeNulls } from "@app/types/shared/utils/general";
 import type { GroupSpaceKind, SpaceKind, SpaceType } from "@app/types/space";
 import assert from "assert";
-import uniq from "lodash/uniq";
 import type {
   Attributes,
   CreationAttributes,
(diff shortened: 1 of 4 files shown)


tdraier:

- #30711 ◀

front/lib/resources/space_resource.ts
```diff
@@ -1723,9 +1703,29 @@ export class SpaceResource extends BaseResource<SpaceModel> {
...
+        return false;
+      // Regular spaces: open spaces grant every member `reader`, restricted spaces grant their
+      // groups `member` — either way membership means holding `read`.
+      case "regular":
+        return auth.getGrantedVerbs("space", this.id).includes("read");
+      // Projects: the workspace global group is attached to unrestricted projects as a `reader`
+      // viewer, so membership is `write` — held by editors (`admin`) and members (`member`) but not
+      // by the global viewer grant.
+      case "project":
+        return auth.getGrantedVerbs("space", this.id).includes("write");
```
fabiencelier: do we have (unit) tests covering these ?
I'm afraid we break something at some point, assuming that membership to a project is the write role is not a trivial change

front/lib/resources/space_resource.ts
```diff
@@ -416,40 +414,22 @@ export class SpaceResource extends BaseResource<SpaceModel> {
...
-      },
-    });
-
-    const podVaultIds = uniq(raw.map((v) => v.vaultId));
+    // Project (pod) spaces the caller belongs to are those on which they hold `write`: the workspace
+    // global group is attached to unrestricted projects as a `reader` viewer, so a `write` grant
+    // marks an actual member (an editor or member group), not every workspace member. Selecting by
+    // `write` and `kind: "project"` reproduces the former `group_vaults` member/editor lookup, so
+    // the previous safety re-filter is redundant.
+    const podSpaceModelIds = auth.getResourceIdsWithVerb("space", "write");
```
fabiencelier: can we introduce a constant POD_SPACE_MEMBERSHIP_VERB = "write" and document it once for all there why it works like that ?

fabiencelier (approved):

PR #30718, opened 2026-08-18
Fix attribution of late sandbox child actions

flvndvd:
## Description

We observed a few churning attribution workflows repeatedly failing with:

```
AssertionError: An attributed run usage is missing tool evidence
```

One LLM run produced a `sandbox__bash` parent. During that same bash execution, two distinct `dsbx` calls occurred sequentially:
1. `get-available-tables`
2. `get-audience-fields`

They shared the same parent action, step content, and dustRunId, but had different tool names and input hashes.

The first attribution pass ran while the parent was still pending and persisted:
- The model usage items.
- The pending parent bash item.
- The first sandbox-child item.

Five minutes later, the first child errored and the same bash process created the second child, which blocked for approval.

### Failure

On the next attribution pass, the workflow correctly fetched fresh database state. It found the second action but no corresponding consumption item.

The coverage assertion rejected any new action belonging to an already-attributed run before the existing `hasActionNeedingEvidence` path could process it. Temporal retries therefore encountered the same
deterministic failure.

This PR fixes by allowing a missing action item only when all of the following prove it is a legitimate continuation:
- It is a sandbox child.
- Its referenced parent exists.
- The parent consumption item is still pending.
- Parent and child share the same step and producing run usage.

The existing attribution workflow then:

- Completes previously pending children that have finished.
- Creates the new child’s zero-footprint pending item.
- Leaves the parent pending until its actual bash result becomes available.
- Preserves the already-stored model-token partition.

Missing ordinary/model-visible actions, missing parents, completed parents, or mismatched runs still trigger the assertion.

No consumption items are created outside the attribution workflow.

### DIFF
--- a/front/scripts/verify_agent_message_consumption_attribution.ts
+++ b/front/scripts/verify_agent_message_consumption_attribution.ts
@@ -107,7 +107,7 @@ async function getCoverageSnapshot(
       return [];
     }
 
-    const dustRunId = action.stepContent.dustRunId;
+    const { dustRunId } = action.stepContent;
     if (!dustRunId) {
       return [];
     }
--- a/front/scripts/verify_agent_message_consumption_attribution.ts
+++ b/front/scripts/verify_agent_message_consumption_attribution.ts
@@ -113,7 +113,7 @@ async function getCoverageSnapshot(
     }
     const childInfo = action.stepContext.sandboxChildActionInfo;
     const parentAction = isSandboxChildActionInfo(childInfo)
-      ? actionBySId.get(childInfo.parentActionId)
+      ? actionById.get(childInfo.parentActionId)
       : undefined;
     const parentItem = parentAction
       ? toolItemByActionModelId.get(parentAction.id)
--- a/front/scripts/verify_agent_message_consumption_attribution.ts
+++ b/front/scripts/verify_agent_message_consumption_attribution.ts
@@ -83,7 +83,7 @@ async function getCoverageSnapshot(
       .filter((item) => item.itemType === "tool")
       .map((item) => [item.agentMCPActionId, item])
   );
-  const actionBySId = new Map(actions.map((action) => [action.sId, action]));
+  const actionById = new Map(actions.map((action) => [action.sId, action]));
   const dustRunIdByRunModelId = new Map(
     runs.map((run) => [run.id, run.dustRunId])
   );
--- a/front/scripts/verify_agent_message_consumption_attribution.ts
+++ b/front/scripts/verify_agent_message_consumption_attribution.ts
@@ -1,0 +1,251 @@
+/**
+ * Verifies action coverage for one agent message and, with --execute, invokes the exact
+ * attribution materializer used by the Temporal activity before reporting coverage again.
+ *
+ * Read-only inspection:
+ *   npx tsx scripts/verify_agent_message_consumption_attribution.ts \
+ *     --workspaceId <workspace-sId> \
+ *     --agentMessageId <agent-message-sId>
+ *
+ * Recompute and verify:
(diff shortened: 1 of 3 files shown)


front/scripts/verify_agent_message_consumption_attribution.ts
```diff
@@ -0,0 +1,251 @@
...
+  const currentItems = items.filter(
+    (item) =>
+      item.attributionVersion === AGENT_MESSAGE_CONSUMPTION_ATTRIBUTION_VERSION
+  );
+  const toolItemByActionModelId = new Map(
+    currentItems
+      .filter((item) => item.itemType === "tool")
+      .map((item) => [item.agentMCPActionId, item])
+  );
+  const actionBySId = new Map(actions.map((action) => [action.sId, action]));
```
aubin-tchoi:
```suggestion
  const actionById = new Map(actions.map((action) => [action.sId, action]));
```

front/scripts/verify_agent_message_consumption_attribution.ts
```diff
@@ -0,0 +1,251 @@
...
+  const attributableActions = actions.filter((action) => {
+    const dustRunId = action.stepContent.dustRunId;
+    return dustRunId && runUsageIdsByDustRunId.has(dustRunId);
+  });
+  const missingActions = attributableActions.flatMap((action) => {
+    if (toolItemByActionModelId.has(action.id)) {
+      return [];
+    }
+
+    const dustRunId = action.stepContent.dustRunId;
```
aubin-tchoi: nit
```suggestion
    const { dustRunId } = action.stepContent;
```

front/scripts/verify_agent_message_consumption_attribution.ts
```diff
@@ -0,0 +1,251 @@
...
+          ]
+        : [];
+    }),
+    currentToolItems: toolItemByActionModelId.size,
+    missingActions,
+  };
+}
+
+function printCoverage(label: string, snapshot: CoverageSnapshot): void {
+  console.log(`\n${label}`);
```
aubin-tchoi, line 168: OOC why not our logger? will produce nicer logs

front/scripts/verify_agent_message_consumption_attribution.ts
```diff
@@ -0,0 +1,251 @@
...
+import { Authenticator } from "@app/lib/auth";
+import { AgentMCPActionResource } from "@app/lib/resources/agent_mcp_action_resource";
+import { AgentMessageConsumptionItemResource } from "@app/lib/resources/agent_message_consumption_item_resource";
+import { ConversationResource } from "@app/lib/resources/conversation_resource";
+import { RunResource } from "@app/lib/resources/run_resource";
+import { makeScript } from "@app/scripts/helpers";
+import type { ModelId } from "@app/types/shared/model_id";
+
+type MissingActionCoverage = {
+  action: string;
```
aubin-tchoi, line 28: would call this `actionId`

front/scripts/verify_agent_message_consumption_attribution.ts
```diff
@@ -0,0 +1,251 @@
...
+  actionModelId: ModelId;
+  tool: string;
+  status: string;
+  dustRunId: string;
+  runUsageIds: string;
+  parentAction: string;
+  parentItem: "completed" | "missing" | "pending" | "n/a";
```
aubin-tchoi, line 35: would call this `parentItemStatus`

aubin-tchoi (approved):
LGTM

front/scripts/verify_agent_message_consumption_attribution.ts
flvndvd: Mainly used to debug, will keep it to patch stuff in prod and deleted once done.

PR #30720, opened 2026-08-18
Prevent sandbox parent action rewinds

flvndvd:
## Description

While debugging stuck analytics workflows, I've discovered that somehow a sandbox parent action whose consumption
item was completed, while the action had moved back to `blocked_child_action_input_required`.

This proved that a late sandbox-child pause could rewind a final parent action. The transition was written unconditionally from two paths.

This PR adds some guards around preventing this (yes, it fails hard but for good reasons, and TBH there is not good way to handle this at an application level)


adrsimon (approved):
LGTM

PR #30716, opened 2026-08-18
refactor: bucket consumption raw export by 6-hour Temporal activities

avervaet:
## Description

Changes the raw consumption export's Temporal behavior from one activity that sequentially paginates over the whole period, to one activity per 6-hour bucket that writes its own CSV part to GCS, followed by a final activity that merges the parts into the zip, uploads it, and notifies.

This bounds how much data any single activity has to fetch, and lets a failed bucket retry independently instead of restarting the whole export.

## Tests

- Added unit coverage for splitting periods into fixed 6-hour buckets, clipping the final bucket to the period end, and handling empty periods.
- Added activity coverage for headerless bucket CSV uploads and search failures.
- Added finalization coverage for ordered bucket merging, writing the CSV header once, producing the final zip, notifying the requester, cleaning up temporary parts, and handling empty exports.

## Risk

Low

## Deploy Plan

- Deploy front

PR #30722, opened 2026-08-18
feat: download instead of (re)triggering when a consumption export is cached

avervaet:
## Description

Stacked on top of #30716.

Clicking "New export" (or the auto-start flow) for a period+filter that already had a completed export hit the same `"cached"` branch `launchConsumptionExportWorkflow` already used for the status check, but the start endpoint ignored it and always told the caller a workflow was generating (`{ isGenerating: true }`), even though nothing was actually (re)triggered.

Now the endpoint reports the cached export's file name instead, and the client goes straight to downloading it rather than switching into a misleading "generating" state.

## Tests

- `front-api/routes/w/[wId]/analytics/consumption/export-raw.test.ts`: covers the new cached-name response and that no workflow gets started.

## Risk

Low

## Deploy Plan

- Deploy front


avervaet:

- #30696
    - #30716
        - #30722 ◀

flvndvd (approved):

PR #30723, opened 2026-08-18
[front] perf: killswitch don't load perm

id13:


fabiencelier (approved):
let's try, if it breaks anything re-enable the kill switch

PR #30725, opened 2026-08-18
[front] perf: add caching on `GroupResource` lookups by ID

aubin-tchoi:
## Description

When fetching skills, the query on groups can take a significant amount of time (in [this trace](https://app.datadoghq.eu/apm/traces?query=operation_name%3Ahono.request%20service%3Afront%20%40span.kind%3Aserver%20resource_name%3A%22POST%20%2Fapi%2Fw%2F%3AwId%2Fanalytics%2Fconsumption%2Ffacets%22&agg_m=count&agg_m_source=base&agg_t=count&cols=service%2Cresource_name%2C%40duration%2C%40http.method%2C%40http.status_code%2C%40_span.count%2C%40_duration.by_service&fromUser=false&graphType=waterfall&historicalData=true&messageDisplay=inline&query_translation_version=v0&shouldShowLegend=true&sort=desc&spanID=6693657449113243275&spanType=all&storage=hot&timeHint=1787057678789&trace=AwAAAaAU8CnFxkyleQAAABhBYUFVOEVEWUFBRGtFMGpwcW5ueFlNcHUAAAAkMTFhMDE0ZmItMTU5NC00Zjk0LWFlZTgtOTE2ZjU4N2Y4M2Q3AAPU0w&trace_group_by_from=a&trace_group_by_metrics=a&traceID=6a84560c000000006ecabb3852242598&traceQuery=&view=spans&start=1787123238852&end=1787126838852&paused=false) it's the slowest query of all). We fetch the groups in the `baseFetch` through `GroupResource.fetchByModelIds`.

This PR adds a per-group read-through caching using what was introduced in #30270.

The existing pattern did not support batch fetches. this PR extends it for this. We still cache each resource individually, and retrieve them with an `MGET` + fetch from the db the missing ones. In our case the sets of groups to fetch are often disjoint and actually only containing 1 value, which would encourage the use of a single value for an array of resources, but this made the invalidation path a bit complex and since the complexity of the mget grows linearly with the number of resources to retrieve I felt like it didn't have too much of a cost and I'd rather introduce this pattern than one customized on the exact access patterns of this specific resource (group). Also, we often batch fetch skills, and depending on your rights you might be pulling different lists of groups so it's better to cache each group in this case.

## Tests

- Added tests.
- Tested locally.

## Risk

- Low.

## Deploy Plan

- Deploy front.

### DIFF
--- a/front-api/lib/api/poke/cache_catalog.test.ts
+++ b/front-api/lib/api/poke/cache_catalog.test.ts
@@ -10,12 +10,10 @@ describe("Poke cache catalog", () => {
     const workspace = getPokeCacheOperations("workspace_by_sid");
     const activeSeats = getPokeCacheOperations("workspace_active_seats");
 
-    expect(
-      group?.buildKey({ workspaceModelId: "42", groupModelId: "7" })
-    ).toBe("cacheWithRedis-group_by_model_id-v1:workspace:42:group:7");
-    expect(group?.keyPattern).toBe(
-      "cacheWithRedis-group_by_model_id-v1:*"
+    expect(group?.buildKey({ workspaceModelId: "42", groupModelId: "7" })).toBe(
+      "cacheWithRedis-group_by_model_id-v1:workspace:42:group:7"
     );
+    expect(group?.keyPattern).toBe("cacheWithRedis-group_by_model_id-v1:*");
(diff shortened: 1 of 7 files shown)


aubin-tchoi:
superseded by https://github.com/dust-tt/dust/pull/30922

PR #30726, opened 2026-08-18
[front] perf: Index conversations.triggerId to cover its foreign key

id13:
## Description

Delete conversation can rock the DB from times to times (as it did today in EU, twice, see [here](https://app.pganalyze.com/databases/-642267629/queries/12133610563?t=6h))


aubin-tchoi (approved):
LGTM

PR #30727, opened 2026-08-18
Use the node IDs Front reports for removes and renames

flvndvd:
## Description

When the daemon deletes a file it first asks Front which node the name points at, then asks Front to remove that name. Another sandbox sharing the same Pod can move a different node onto that name between those two steps, and the daemon's own one-second cache of file details widens the gap. Until now the daemon assumed the name still pointed at the node it had looked up, and it marked that node as gone. Renaming over an existing file made the same assumption about the file being replaced.

Marking the wrong node has a quiet consequence. A file that is still there gets treated as deleted, so if a program had it open with unsaved changes, those changes are dropped rather than sent to Front, and nothing reports an error. The file that really did lose its name is treated as still alive, so a program writing to it gets a confusing failure later when its save reaches a node that no longer exists. This needs two sandboxes acting on the same name at almost the same moment, so it is a narrow race, but sharing a Pod is a supported arrangement rather than an accident.

Front now answers both operations with the node it actually affected, so the daemon uses that answer instead of its own guess. A rename no longer looks the destination up beforehand either, which removes both a request and one more thing that could be out of date. When Front's answer and the daemon's own lookup disagree, the daemon records it and keeps Front's answer.

Both fields are required rather than optional. A rename sends an empty value when the destination name was free, and a response that omits the field entirely is refused instead of guessed at, which is how the directory listing cursor is already treated in the same code. Guessing was the alternative and it was worse: an earlier version of this change tried to fall back to the daemon's own lookup when the field was missing, which quietly reintroduced the same race on that path.

### DIFF
--- a/cli/dust-sandbox/src/commands/filesystem/client.rs
+++ b/cli/dust-sandbox/src/commands/filesystem/client.rs
@@ -94,7 +94,29 @@ struct UploadResponse {
 }
 
 #[derive(Debug, Deserialize)]
-struct EmptyResponse {}
+#[serde(rename_all = "camelCase")]
+struct RemoveResponse {
+    // Nothing else tells the daemon which node lost its name, so a response
+    // without this is invalid.
+    removed_node_id: u64,
+}
+
+#[derive(Debug, Deserialize)]
+#[serde(rename_all = "camelCase")]
+struct RenameResponse {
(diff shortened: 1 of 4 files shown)


aubin-tchoi (approved):
LGTM

PR #30728, opened 2026-08-18
[front] perf: Fix slow conversation deletion on EU cluster

id13:
## Description

topped at 20% CPU for 1h today because of an index absent : conversation_participants FK

### DIFF
--- a/front/migrations/post-deploy/20260818200100_drop_legacy_conversation_participants_conversation_id_fk.sql
+++ b/front/migrations/post-deploy/20260818200100_drop_legacy_conversation_participants_conversation_id_fk.sql
@@ -1,0 +1,7 @@
+/*
+Statement 0
+  - INDEX_DROPPED: Dropping this index means queries that use this index might perform worse because they will no longer will be able to leverage it.
+*/
+SET SESSION statement_timeout = 1200000;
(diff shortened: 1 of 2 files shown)


id13:
PMRR

front/migrations/post-deploy/20260818200100_drop_legacy_conversation_participants_conversation_id_fk.sql
```diff
@@ -0,0 +1,7 @@
+/*
+Statement 0
+  - INDEX_DROPPED: Dropping this index means queries that use this index might perform worse because they will no longer will be able to leverage it.
+*/
+SET SESSION statement_timeout = 1200000;
+SET SESSION lock_timeout = 3000;
+DROP INDEX CONCURRENTLY IF EXISTS "conversation_participants_conversationId_fk";
```
id13, line 7: to be consistent everywhere, this is in US but not in EU

aubin-tchoi, line 7: Weird name for an index :p

aubin-tchoi (approved):

PR #30729, opened 2026-08-18
[Activation] Make activation_work_areas.podId non-nullable, retire legacy statuses

frankaloia:
## Summary
- Kill unused `candidate`/`confirmed` statuses.
- Make invariant that work areas are associated with a pod. Ok to drop columns.
- Why? Logical clean up before we had more complexity around this.
- Clean to deploy before as it should not effect any customer facing logic

## Testing
migrations ran

## Risk
Low

## Deploy
1. Run both migration
2.

### DIFF
--- a/front/lib/resources/activation_work_area_resource.ts
+++ b/front/lib/resources/activation_work_area_resource.ts
@@ -1,8 +1,5 @@
 import type { Authenticator } from "@app/lib/auth";
-import type {
-  ActivationWorkAreaStatus,
-  PublicActivationWorkAreaStatus,
-} from "@app/lib/models/activation/activation_work_area";
+import type { ActivationWorkAreaStatus } from "@app/lib/models/activation/activation_work_area";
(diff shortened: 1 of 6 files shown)


front/lib/resources/activation_work_area_resource.ts
```diff
@@ -88,11 +50,7 @@ export class ActivationWorkAreaResource extends BaseResource<ActivationWorkAreaM
 
   static async makeNew(
     auth: Authenticator,
-    blob: Pick<
-      CreationAttributes<ActivationWorkAreaModel>,
-      "title" | "description"
-    > &
-      Partial<Pick<CreationAttributes<ActivationWorkAreaModel>, "podId">>
+    blob: { title: string; description: string; podId: ModelId }
```
flvndvd, line 53: `nit` Why not picking directly from the `CreationAttribute<ActivationWorkAreaModel>`?

flvndvd (approved):

PR #30730, opened 2026-08-18
Ship the Dust filesystem in sandboxes

flvndvd:
## Description

This PR ships the Rust filesystem daemon built in the previous PRs.

It bumps dsbx to 0.1.51 and dust-base to 0.8.85. The new image declares the `dust_filesystem` capability, which Front checks before starting the daemon. Sandboxes using the existing GCS filesystem remain unchanged.

Once merged, we will publish `dsbx-v0.1.51`. The sandbox image workflow can then build `dust-base:0.8.85` with that binary.


smb2268 (approved):
LGTM

PR #30732, opened 2026-08-18
[Activation] Add get_workspace_members_context tool, gate to admins only

frankaloia:
## Summary

- Adds an admin-only tool to allow a user to list all workspace members (including group + job type information)
- This will be used to allow admins to configure work areas for the activation of others in the workspace

## Testing

Screenshot 2026-08-18 at 6 43 20 PM

## Risk

Low — new tool, additive only. Non-admins are unaffected (tool not registered).

## Deploy
1. Deploy front


front/lib/api/actions/servers/workspace_people/index.ts
```diff
@@ -0,0 +1,259 @@
...
+    );
+
+  return new Ok(
+    await buildMemberRows(auth, users, membershipByUserId, jobTypesByUserId)
+  );
+}
+
+const handlers: ToolHandlers<typeof WORKSPACE_PEOPLE_TOOLS_METADATA> = {
+  list_workspace_members: async ({ userIds, jobType, groupId }, { auth }) => {
+    if (!auth.isAdmin()) {
```
matteotrab: I don't have full context on this, but would it make sense to allow managers too?
Today managers can already invite members, create/edit groups and assign some permissions

matteotrab (approved):

front/lib/api/actions/servers/workspace_people/index.ts
frankaloia: This is intended to be quite generalizable, i think that's fair given managers have that ability. i do want it consistent with the UX

PR #30733, opened 2026-08-18
[front] Teach the agent how to change an existing Pod app

davidebbo:
Instructions-only fixes.

**Finding an existing Pod app.** Editing a Pod app's Frame requires both the Frame's Pod source path and its `file_id` to publish. Teach agent how to get that correctly.

**Moving a Frame.** Observed in practice: the agent moved a new Frame into its Pod folder with `mv` in the Computer instead of `files__move`. Only `files__move` carries the Frame's file record along with its bytes, so the Frame stopped belonging to the Pod its source now lived in. The Pod section now says to use `files__move` and never `mv` or `cp` in the Computer.

One test asserting the new subsection is rendered in a Pod conversation.

### DIFF
--- a/front/lib/api/actions/servers/interactive_content/instructions.ts
+++ b/front/lib/api/actions/servers/interactive_content/instructions.ts
@@ -162,11 +162,11 @@ ${PUBLISH_INTERACTIVE_CONTENT_FILE_TOOL_NAME}({
 
 From then on, edit the source at its Pod path and publish it again. Anything the Frame imports relatively must live under its folder, which is the bundling root.
 
-#### Changing An Existing Pod App
+#### Changing An Existing Pod Frame
 
-A Pod app you are asked to change was usually built in an earlier conversation, so you know neither its source path nor its file id, and publishing needs both. Never recreate it, and never guess either value: read them off the Pod listing.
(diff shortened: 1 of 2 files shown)


front/lib/api/actions/servers/interactive_content/instructions.ts
```diff
@@ -161,6 +161,14 @@ ${PUBLISH_INTERACTIVE_CONTENT_FILE_TOOL_NAME}({
 \`\`\`
 
 From then on, edit the source at its Pod path and publish it again. Anything the Frame imports relatively must live under its folder, which is the bundling root.
+
+#### Changing An Existing Pod App
```
flvndvd: Shouldn't this be in the pod app skill? So we avoid poisoning the frame skill with a new pod app concept which isn't globally available?

flvndvd (approved):

front/lib/api/actions/servers/interactive_content/instructions.ts
davidebbo: Actually, this whole section is only shown when in a Pod (logic in `buildInteractiveContentInstructions`). I renamed App to Frame since that's really what it's about.

PR #30735, opened 2026-08-18
Keep the filesystem harness in step with the daemon contract

flvndvd:
## Description

The local mount harness stopped working once removes and renames started relying on the node IDs Front reports. Its stand-in Front never learned to send them, so the daemon rejected every answer it gave for those two operations. Deleting or renaming a file through the harness failed with an I/O error, which took the acceptance script down with it. Nothing in the sandbox itself was affected, because the real Front has been sending both fields since they were added.

The stand-in now sends the node it removed, and the node a rename replaced. It sends an explicit empty value when the destination name was free, rather than leaving the field out, because the daemon deliberately treats those as different answers. It also reports a node as gone instead of failing outright when asked about one that no longer exists, which is what the real thing does and what made this confusing to track down in the first place.

Two notes in the harness went stale when the staging directory recovery landed. Both said that one case was expected to fail, and it now passes, so they are gone.

## Tests

All five harness scripts were run against the current daemon, not only the one that was broken. The acceptance script that ships with the sandbox passes end to end, and the mount, staging, time and space checks all report clean. The staging checks are the interesting ones: they went from three failures to passing, which is the recovery fix showing up in the tool for the first time.

## Risk

None to the product. Nothing here ships to a sandbox, and no code outside this directory changes.

The wider point is that this directory pins a copy of Front's contract, so a change on the Front side can leave it quietly wrong while everything else looks healthy, and nothing runs it automatically. Running it in CI would need a runner with FUSE available.

## Deploy Plan

Nothing to deploy.

### DIFF
--- a/x/flav/dust-filesystem-harness/README.md
+++ b/x/flav/dust-filesystem-harness/README.md
@@ -52,9 +52,7 @@ file keeping its identity across a move between roots.
 
 `staging_checks.sh` mounts several times over to check what startup does with the
 local staging directory: a fresh one, one it already claimed, one left mid-claim
-by a crash, and one belonging to something else that must be left untouched. The
-mid-claim case reports a failure until the staging recovery change lands, since
-that is the bug it describes.
+by a crash, and one belonging to something else that must be left untouched.
 
 `time_probe.sh` prints the time a file reports after each step of a write, then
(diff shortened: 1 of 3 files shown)


aubin-tchoi (approved):
LGTM

PR #30736, opened 2026-08-18
Give the sandbox filesystem mount the public API URL

flvndvd:
## Description

A sandbox runs outside our infrastructure, so the only address it can use to reach Front is the public one. Today four call sites work that address out for themselves. Three go through the general base URL helper, which prefers the internal service address wherever that override is set, so they can hand a sandbox an address nothing there can reach. The fourth, the filesystem mount, used a hardcoded production address instead, which is wrong in every region that is not production US and offers no way to point a local sandbox at a development tunnel.

This adds one helper in the config module that resolves the sandbox-facing URL in a single place: the development host when configured, otherwise the public URL of the region it runs in, never the internal override.

It moves only the database filesystem mount onto it, to limit blast radius. The other three keep their own resolution for now, and can follow once we know which processes actually set the internal override.

## Tests

Four tests on the new helper: the public URL wins over the internal override, it follows the region it runs in, it uses the development host when configured, and it ignores that host outside development. The existing mount tests still pass.

## Risk

Production US should resolve to the same address as before. Other regions change on purpose, from the production address to their own. Worth confirming that no environment sets the production API variable to something other than its public URL, since that is the value this stops using. Rollback is a single revert and only affects sandboxes mounted afterwards.

## Deploy Plan

Nothing to sequence. Existing mounts keep the address they started with until recreated.

### DIFF
--- a/front/lib/api/config.test.ts
+++ b/front/lib/api/config.test.ts
@@ -1,0 +1,47 @@
+import config from "@app/lib/api/config";
+import { afterEach, describe, expect, it } from "vitest";
+
+const PUBLIC_URL = "https://eu.dust.tt";
+const INTERNAL_URL = "http://front-internal-service";
+
+describe("getSandboxApiBaseUrl", () => {
+  const originalEnv = { ...process.env };
+
+  afterEach(() => {
+    process.env = { ...originalEnv };
+  });
+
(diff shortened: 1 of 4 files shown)


aubin-tchoi (approved):
LGTM

PR #30737, opened 2026-08-18
[sparkle] fix: keep keyboard navigation in searchable submenus

aubin-tchoi:
## Description

Fixes dust-tt/tasks#10078.

Keyboard navigation in the input bar's searchable Capabilities, Knowledge, and Spaces submenus jumped back to the parent menu when pressing ArrowDown.

The search field now scopes its focus handoff to its closest open menu. This keeps arrow navigation within the active submenu while preserving existing top-level dropdown behavior.

Not super convinced but don't have anything better to suggest.

## Tests

- Added a Storybook story to documnt this.

## Risk

- Low.

## Deploy Plan

- Deploy front.

### DIFF
--- a/sparkle/src/components/Dropdown.tsx
+++ b/sparkle/src/components/Dropdown.tsx
@@ -306,20 +306,9 @@ const nativeInputValueSetter =
 
 const SEARCHABLE_MENU_ITEM_SELECTOR =
   '[role="menuitem"], [role="menuitemcheckbox"], [role="menuitemradio"]';
-const OPEN_SEARCHABLE_MENU_ITEM_SELECTOR = [
-  '[data-radix-menu-content][data-state=open] [role="menuitem"]',
-  '[data-radix-menu-content][data-state=open] [role="menuitemcheckbox"]',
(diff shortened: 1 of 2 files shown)


sparkle/src/components/Dropdown.tsx
```diff
@@ -904,16 +893,21 @@ const DropdownMenuSearchbar = React.forwardRef<
       e.stopPropagation();
       onKeyDown?.(e);
       if (!e.defaultPrevented) {
+        const menuContent = e.currentTarget.closest(
+          "[data-radix-menu-content]"
```
smb2268: should we preserve the [data-state="open"] here like we had above?

smb2268 (approved):
small comment otherwise lgtm!

sparkle/src/components/Dropdown.tsx
aubin-tchoi: Yes, good call. I restored the data-state="open" guard so the closest lookup only scopes to currently open menu content. Added in f6e5e1463f.

PR #30741, opened 2026-08-19
[front] perf: Resolve system key permissions without reading grants

id13:
## Context

This PR aims at reducing the DB load induced by looking up group permission on every api request + agent_loop step.

There are 3 axis:
1. Optimize the query : ✅ this query is very simple, we've exhausted this topic
2. Cache everything everywhere : In progress not the subject of this PR. 
3. Reduce the hits on this query: This PR, still make sense because of how we resolve permission for a system key today, which means retrieving all group permission, thousands of lines x thousands of requests per second => 💣 

Why system key focus ? 

Mostly in connectors traffic, mostly data-source oriented, we use a system key without any requested groups.
In that case, and only in that case, it is equivalent to a wildcard grant scoped to the workspace this system key belongs to.

To optimize this very hot path, we introduce a synthetic wildcard grant.

##  Tests

Added unit tests
Actually very hard to front-edge since it's mostly connector related, can be tested in front-edge with sub-agent runs.

## Risk

Low
Blast radius: Mostly security breach if I'm wrong and it's not equivalent to a wildcard or if this code can grant access to cross workspace reads/writes.

## Plan

Deploy front

### DIFF
--- a/front/lib/auth.ts
+++ b/front/lib/auth.ts
@@ -974,6 +974,11 @@ export class Authenticator {
       : [];
     const keyGroupModelIds = allGroups.map((g) => g.id);
 
+    // `resolvePermissions` reads the key's grants from the key itself, so it only holds for the
+    // key's own workspace and its own groups: `requestedGroupIds` replaces those groups (the Slack
+    // bot acting as a user), and the target workspace below may not be the key's.
+    const permissionsKey = requestedGroupIds ? undefined : key;
+
     let permissions: GroupPermissions;
     let keyPermissions: GroupPermissions;
     if (isKeyWorkspace) {
@@ -982,6 +987,7 @@ export class Authenticator {
       permissions = await this.resolvePermissions({
         workspace,
         groupModelIds: workspaceGroupModelIds,
+        key: permissionsKey,
       });
       keyPermissions = permissions;
     } else {
@@ -993,6 +999,7 @@ export class Authenticator {
         this.resolvePermissions({
           workspace: keyWorkspace,
           groupModelIds: keyGroupModelIds,
+          key: permissionsKey,
         }),
       ]);
     }
@@ -1297,12 +1304,24 @@ export class Authenticator {
   static async resolvePermissions({
     workspace,
     groupModelIds,
+    key,
   }: {
     workspace?: WorkspaceResource | null;
     groupModelIds: ModelId[];
+    key?: Pick<KeyAuthType, "isSystem"> | null;
   }): Promise<GroupPermissions> {
     if (!workspace) {
       return GroupPermissions.empty();
+    }
+
+    if (key?.isSystem) {
+      return GroupPermissions.fromGrants([
+        {
+          grantType: "*",
+          resourceType: "*",
+          resourceId: WHOLE_TYPE_RESOURCE_ID,
+        },
+      ]);
     }
 
     const lightWorkspace = renderLightWorkspaceType({ workspace });
--- a/front/lib/auth.ts
+++ b/front/lib/auth.ts
@@ -19,7 +19,10 @@ import { ConversationModel } from "@app/lib/models/agent/conversation";
 import { isUpgraded } from "@app/lib/plans/plan_codes";
 import { FeatureFlagResource } from "@app/lib/resources/feature_flag_resource";
 import { GlobalFeatureFlagResource } from "@app/lib/resources/global_feature_flag_resource";
-import type { GroupPermissionsJSON } from "@app/lib/resources/group_permission_registry";
+import type {
+  GroupPermissionsJSON,
+  ResourcesWithVerb,
+} from "@app/lib/resources/group_permission_registry";
 import {
   allWorkspacePermissions,
   GroupPermissions,
@@ -1519,15 +1522,16 @@ export class Authenticator {
   }
 
   /**
-   * The instance ids of `resourceType` the caller may `verb`, resolved from their governance grants.
+   * The instances of `resourceType` the caller may `verb`, resolved from their governance grants.
    * The enumeration counterpart of `getGrantedVerbs` — "which resources may I act on" rather than
-   * "what may I do on this one" — for reverse lookups such as the projects a caller belongs to. See
-   * `GroupPermissions.resourceIdsWithVerb`; the type-wide (-1) grant is excluded there.
+   * "what may I do on this one" — for reverse lookups such as the projects a caller belongs to.
+   * Returns `{ kind: "all" }` when a type-wide grant confers the verb on every instance, which a
+   * system key holds on every type (see `resolvePermissions`).
    */
   getResourceIdsWithVerb(
     resourceType: ConcreteResourceType,
     verb: GrantVerb
-  ): ModelId[] {
+  ): ResourcesWithVerb {
     return this._permissions.resourceIdsWithVerb(resourceType, verb);
   }
 
--- a/front/lib/auth.ts
+++ b/front/lib/auth.ts
@@ -27,9 +27,10 @@ import {
 } from "@app/lib/resources/group_permission_registry";
 import { GroupPermissionResource } from "@app/lib/resources/group_permission_resource";
 import { GroupResource } from "@app/lib/resources/group_resource";
-import type { KeyAuthType } from "@app/lib/resources/key_resource";
+import type { KeyAuthType, SystemKey } from "@app/lib/resources/key_resource";
 import {
   DEFAULT_SYSTEM_KEY_NAME,
+  isSystemKey,
   KeyResource,
   SECRET_KEY_PREFIX,
 } from "@app/lib/resources/key_resource";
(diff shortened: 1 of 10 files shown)


id13:

- #30741 ◀

front/lib/resources/group_permission_registry.ts
```diff
@@ -216,7 +227,7 @@ function maskToVerbs(mask: number): GrantVerb[] {
 // full resource and to keep the reference type-only.
 type GrantRow = Pick<
   GroupPermissionResource,
-  "groupId" | "grantType" | "resourceType" | "resourceId"
+  "grantType" | "resourceType" | "resourceId"
```
id13, line 235: @matteotrab @tdraier @fabiencelier your eyes are badly needed here

front/lib/auth.ts
```diff
@@ -1297,14 +1304,26 @@ export class Authenticator {
...
     if (!workspace) {
       return GroupPermissions.empty();
     }
 
+    if (key?.isSystem) {
+      return GroupPermissions.fromGrants([
+        {
+          grantType: "*",
+          resourceType: "*",
+          resourceId: WHOLE_TYPE_RESOURCE_ID,
```
tdraier: this is confusing, with this we expect that we would only have type-level permissions , where I think you mean all permissions at all level -

front/lib/resources/group_permission_registry.ts
```diff
@@ -313,7 +324,7 @@ export class GroupPermissions {
         }
         const verbs =
           grantType === "*"
-            ? allVerbsForResourceAtLevel(rt, level)
+            ? allVerbsForResource(rt)
```
tdraier: seems right, but only make sense if we are on WHOLE_TYPE_RESOURCE_ID , which applies at type level + all instances - if level is instance we should keep `allVerbsForResourceAtLevel(rt, level)`

front/lib/resources/group_permission_registry.ts
tdraier: perfectly fine as groupId was not used

tdraier (approved):

front/lib/auth.ts
```diff
@@ -1297,14 +1304,26 @@ export class Authenticator {
   static async resolvePermissions({
     workspace,
     groupModelIds,
+    key,
```
tdraier: maybe a boolean grantAll instead of passing the key , that could be reused for other cases (if any ?)

id13: My agent started with that and I pushed back against because I'm not a big fan of boolean flag as function arguments if that can be avoided and prefer to derive directly from the data.

front/lib/resources/skill/skill_resource.ts
```diff
@@ -2225,6 +2225,7 @@ export class SkillResource extends BaseResource<SkillConfigurationModel> {
 
   canWrite(auth: Authenticator): boolean {
     // TODO(governance): cleanup we we'll be able to assign API key to editor groups.
+    // TODO(@jd): Revisit this shortcircuit with our current ACLs stack.
     // API keys cannot be added to a skill's editor group (no such assignment mechanism exists),
     // so any key is allowed to write to any skill. Skill *creation* is separately gated by
     // `auth.hasWorkspacePermission("create", "skill")`; this only governs already-existing skills.
```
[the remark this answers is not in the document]
id13: Thank you captain obvious

front/lib/auth.ts
fabiencelier: so the condition to have all grants is to be a system key AND have no requested group ?
Currently this check is split in 2 places (requested group ids before the function, and is system inside)
We should at least document here that the key must only be provided if there is no requested space ids

fabiencelier (approved):

front/lib/auth.ts
```diff
@@ -1297,14 +1304,26 @@ export class Authenticator {
...
     workspace?: WorkspaceResource | null;
     groupModelIds: ModelId[];
+    key?: Pick<KeyAuthType, "isSystem"> | null;
   }): Promise<GroupPermissions> {
     if (!workspace) {
       return GroupPermissions.empty();
     }
 
+    if (key?.isSystem) {
+      return GroupPermissions.fromGrants([
```
spolu, line 1324: Okay... lot of magic here but looks like it was intended for this use case 🙈

front/lib/resources/group_permission_registry.ts
```diff
@@ -178,6 +178,22 @@ function allVerbsForResourceAtLevel(
   return [...verbs];
 }
 
+// The verbs a "*" grant confers on `resourceType`. At WHOLE_TYPE_RESOURCE_ID the wildcard stands
+// for the type and for every instance of it, so the instance-level roles count alongside the
+// type-level ones; on a concrete id only the instance-level roles apply.
+function allVerbsForWildcard(
```
spolu, line 184: Hmmm this one feels a bit weird? If `resourceId` is `WHOLE_TYPE_RESOURCE_ID` then level is "type". But if `grantType` is `*` then we override that "level" ?

Maybe level should be an array in `allVerbsForResourceAtLevel` and `level` should be set to both levels in `fromGrants` when `resourceId` is `WHOLE_TYPE_RESOURCE_ID` ?

tdraier: actually a permission on WHOLE_TYPE_RESOURCE_ID applies at type level +  all instances , so it still makes sense to have all verbs in that case .

id13:
> so the condition to have all grants is to be a system key AND have no requested group ?
Currently this check is split in 2 places (requested group ids before the function, and is system inside)
We should at least document here that the key must only be provided if there is no requested space ids

@fabiencelier good point, will narrow the typing

tdraier:
Need to consider the possible semantic conflicts with https://github.com/dust-tt/dust/pull/30711 (just realized we could have some issues after merging this). #30711 changes listWorkspacePodsAsMember and isMember to rely on permissions, based on resourceIdsWithVerb .

Here, we resolve a system key to one synthetic grant {"*","*",-1}, and its allVerbsForWildcard makes that -1 (type-wide) entry carry instance verbs (read/write/admin).

isMember uses getGrantedVerbs, which folds in -1 -> returns true for every regular/pod space . consistent with canRead() = true and same as before

listWorkspacePodsAsMember : before both PR, a system key is in every pod group, so return all pods. with both PRs : returns now empty array for system key. I don't think it's an issue, but worth checking

id13:
`{ kind: "all" } | { kind: "ids"; resourceIds: ModelId[] }` as a return of resourceIdsWithVerb is what we might have to do then

front/lib/resources/group_permission_registry.ts
spolu: Yep agreed it makes sense but isn't the right place to encode that in `fromGrants` by having a level array instead of a level variable vs having a magical function that treat level type as having type and instance ?

id13: @spolu an example would help me understand what you mean

spolu: If I understand correctly the change [here](https://github.com/dust-tt/dust/pull/30741/changes#diff-9c2a0002372a22139b56d6efc6da35936108137cc3819d7428122e5ca7778e43L316-R339) is to ensure that when `resourceId === WHOLE_TYPE_RESOURCE_ID` we get all verbs for both `type` and `instance`.

It works because for now level is `type` only when `resourceId === WHOLE_TYPE_RESOURCE_ID` but that moves this "level" logic (the fact that a `WHOLE_TYPE_RESOURCE_ID` should get both `type` and `instance` verbs) inside of `allVerbsForWildcard` instead of where it should naturally fall which is when we define the `level` variable which is [here](https://github.com/dust-tt/dust/blob/main/front/lib/resources/group_permission_registry.ts#L326)

My proposal is to have:

```
const levels: GrantLevel[] =
        resourceId === WHOLE_TYPE_RESOURCE_ID ? ["type", "instance"] : ["instance"];
```

And rename `allVerbsForResourceAtLevel` into `allVerbsForResourceAtLevels` and accept an array there. Kill `allVerbsForWildcard`. This would keep the "logic" for level access of `WHOLE_TYPE_RESOURCE_ID` centralised in `fromGrants`?

Am I missing something?

spolu:
> Need to consider the possible semantic conflicts with https://github.com/dust-tt/dust/pull/30711 (just realized we could have some issues after merging this). https://github.com/dust-tt/dust/pull/30711 changes listWorkspacePodsAsMember and isMember to rely on permissions, based on resourceIdsWithVerb .

hmmm isn't it because the semantics of `getResourceIdsWithVerb` is wrong when the permission is `WHOLE_TYPE_RESOURCE_ID` ?  Shouldn't it pull the resources when called on the permission with resourceId `WHOLE_TYPE_RESOURCE_ID` ?

Or alternatively shouldn't we never use permissions to list things and only rely on them to check access to materialised things, aka keep the group or space query?

front/lib/resources/group_permission_registry.ts
tdraier: agree 👍 . the current `level = resourceId === WHOLE_TYPE_RESOURCE_ID ? "type": "instance"`  is misleading, proposal matches more what resourceId === WHOLE_TYPE_RESOURCE_ID means

tdraier:
> > Need to consider the possible semantic conflicts with #30711 (just realized we could have some issues after merging this). #30711 changes listWorkspacePodsAsMember and isMember to rely on permissions, based on resourceIdsWithVerb .
> 
> hmmm isn't it because the semantics of `getResourceIdsWithVerb` is wrong when the permission is `WHOLE_TYPE_RESOURCE_ID` ? Shouldn't it pull the resources when called on the permission with resourceId `WHOLE_TYPE_RESOURCE_ID` ?
> 
> Or alternatively shouldn't we never use permissions to list things and only rely on them to check access to materialised things, aka keep the group or space query?

was indeed wrong - with latest changes, getResourceIdsWithVerb correctly return "all" when we have resourceId === -1 .

front/lib/resources/group_permission_registry.ts
spolu: Ready here: https://github.com/dust-tt/dust/pull/30763

PR #30742, opened 2026-08-19
[Agent Loop] Heartbeat the whole model activity and retry steps on worker shutdown

PopDaph:
## Description

Fixes for the 2026-08-18 `agent-loop-interactive-queue-v2` incident ([Datadog](https://app.datadoghq.eu/logs?from_ts=1787050500000&live=false&query=service%3Afront-agent-loop-worker+task_queue%3Aagent-loop-interactive-queue-v2+%22Activity+task+timed+out%3A+activity+Heartbeat+timeout%22&stream_sort=asc&to_ts=1787051400000)): 70 EU + 56 US workflow failures, with `runModelAndCreateActionsActivity` failing all 5 attempts with `TIMEOUT_TYPE_HEARTBEAT`.

- **Heartbeat the whole model activity.** The pre-stream setup had no heartbeat, and a hung MCP server's tools/list call times out after 60s, exactly the heartbeat timeout, so every retry died before reaching the LLM stream (each failed attempt shows "Starting multi-action loop iteration", 60s of silence, then "Error listing tools from MCP server: MCP error -32001"). The activity now heartbeats on entry and every 20s for its whole duration (same pattern as tool setup, #28780). This is not incident-specific: the fleet-wide max of `agent_loop_step.time_to_provider_call` exceeds 60s many times per day, every day (spikes up to 562s over the past week), and each of those is a silently heartbeat-timed-out attempt today.
- **Retry on worker shutdown instead of finalizing as success.** The LLM stream loop treated any `CancelledFailure` as a user "Stop agent", so a deploy mid-stream silently finalized the message as successful with a truncated answer. Worker shutdown is now surfaced as a retryable failure so the step reruns on another worker. The stream loop watches the pod shutdown signal (aborts 10s before the grace period ends, like tools do), because Temporal's own WORKER_SHUTDOWN cancellation only fires at grace expiry, together with SIGKILL, too late to report anything.

- **Cap agent-loop tools/list requests at 30s** (review feedback): they inherited the MCP SDK default of exactly 60s, the same value as the heartbeat timeout, which made the collision deterministic. A failed listing already degrades gracefully (the server's tools are skipped for the step), so the lower cap only makes that existing degraded mode kick in sooner. No latency data exists for tools/list today, so 30s is deliberately conservative; listings slower than 5s are now logged ("Slow MCP tools listing") to tune the cap on real data.

> [!NOTE]
> Behavior change on every deploy: all in-flight model streams on a draining pod abort at 110s into the grace period and retry the step on another worker, redoing the full LLM call. This includes streams that would have completed in the 110-120s window before SIGKILL. Tools already made this exact trade (`SHUTDOWN_TOOL_ABORT_DELAY_MS`); the alternative is dying at 120s and being discovered by heartbeat timeout 60s later, or worse, being finalized as a truncated "success".

Follow-ups (separate PRs): heartbeat the tool activity's MCP connect phase (the un-heartbeated gap that killed 69 sandbox child tool workflows in the EU burst), exponential backoff on the model activity retry policy, thread an AbortSignal into `llm.stream()` so early exits (timeouts, shutdown) abort the provider read eagerly instead of abandoning it after the bounded cleanup wait.

## Tests

- New unit tests for `withPeriodicHeartbeat` and `classifyTemporalAbortReason` (neither had any).
- `tsgo --noEmit` clean, existing agent-loop tests pass.

## Risk

Low. Heartbeats are throttled by the SDK; no happy-path change. The deploy-time behavior change is described in the note above: interrupted model steps retry (at the cost of redoing the LLM call) instead of being silently truncated. No schema/API change, safe to rollback.

## Deploy Plan

- [ ] Deploy `front` (front-agent-loop-worker picks it up on rollout).
- [ ] Watch `service:front-agent-loop-worker "Activity task timed out: activity Heartbeat timeout"` in Datadog: model-activity heartbeat timeouts should disappear.

### DIFF
--- a/front/temporal/agent_loop/activities/run_model_and_create_actions_wrapper.ts
+++ b/front/temporal/agent_loop/activities/run_model_and_create_actions_wrapper.ts
@@ -15,7 +15,11 @@ import {
   AGENT_LOOP_SUBAGENT_HARD_CAP,
   checkCostAndSubagentsThresholds,
 } from "@app/temporal/agent_loop/activities/cost_threshold_warnings";
-import { RUN_MODEL_ACTIVITY_TIMEOUT_SAFETY_MARGIN_MS } from "@app/temporal/agent_loop/config";
+import { withPeriodicHeartbeat } from "@app/lib/utils/async_utils";
+import {
+  MODEL_ACTIVITY_HEARTBEAT_INTERVAL_MS,
+  RUN_MODEL_ACTIVITY_TIMEOUT_SAFETY_MARGIN_MS,
+} from "@app/temporal/agent_loop/config";
 import type { ActionBlob } from "@app/temporal/agent_loop/lib/create_tool_actions";
 import { createToolActionsActivity } from "@app/temporal/agent_loop/lib/create_tool_actions";
 import { handlePromptCommand } from "@app/temporal/agent_loop/lib/prompt_commands";
@@ -31,7 +35,7 @@ import {
 } from "@app/types/assistant/agent_run";
 import type { ModelId } from "@app/types/shared/model_id";
 import { startActiveObservation } from "@langfuse/tracing";
-import { Context } from "@temporalio/activity";
+import { Context, heartbeat } from "@temporalio/activity";
 
 export type RunModelAndCreateActionsResult = {
   actionBlobs: ActionBlob[];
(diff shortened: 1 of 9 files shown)


front/temporal/agent_loop/activities/run_model_and_create_actions_wrapper.ts
```diff
@@ -79,15 +83,27 @@ export async function runModelAndCreateActionsActivity({
...
-      runIds,
-      step,
-      forceDisableToolUse,
-    })
+  // The pre-stream setup (agent data loading, MCP tools listing, conversation rendering) can
+  // stall past the heartbeat timeout, e.g. on a hung MCP server's tools/list call: heartbeat
+  // immediately and periodically for the whole activity. The LLM stream adds its own heartbeats.
+  heartbeat();
+
+  return withPeriodicHeartbeat(
```
fontanierh, line 91: if it' one async call that takes 60 seconds this isn't going to fix it because we will still not heartbeat anything during the call I think... maybe we should lower the timeout of tools/list to stg that is less than the heartbeat timeout ?

fontanierh (approved):

front/temporal/agent_loop/activities/run_model_and_create_actions_wrapper.ts
PopDaph: So iiuc the interval actually covers thiis:  the heartbeat comes from a setInterval around the whole activity, and timers keep firing while an async call is stuck waiting. A hung tools/list is just idle network wait, so the event loop is free to heartbeat every 20s. It's the same trick the tool activity already uses for its DB setup (#28780).

Lowering the tools/list timeout is a good idea anyway though: it was silently inheriting the MCP SDK default, which is exactly 60s, same as the heartbeat timeout, which is why the failure was so deterministic. 

I added a commit: capped at 30s. Failed listings already degrade gracefully (that server's tools are just skipped for the step), so the only change is we stop waiting a full minute on a dead server.

PR #30743, opened 2026-08-19
[front] perf: Cache group permissions per workspace in Redis

id13:
## Description

This PR aims at reducing the DB load induced by looking up group permission on every api request + agent_loop step.

There are 3 axis:

Optimize the query : ✅ this query is very simple, we've exhausted this topic
Cache everything everywhere : This PR
Reduce the hits on this query:  In progress not the subject of this PR.

Why try to be efficient with caching in this PR, by reading only what's needed, and writing only what's needed too, limiting ourselves to workspaceId/groupId lookup in the DB and redis (and keeping both predicate in sync is crucial), and filtering on everything else in JS.

This choice was made necessary by the resolvePermission done in lib/auth.ts on every request / agent-loop step.
Because of that design, we have to keep in cache all group perms for a given workspace + groupIds, but we really don't want to transfer thousands of lines on every request / step.

So it's implemented as a Redis Hash, with read path doing multi hmget, and invalidate path invalidates the keys it needs to.

## Tests

Added tests
Will front-edge

## Risk

Low
Blast radius: auth so everything. Risk is not security, rather crashing the auth code.

## Deploy Plan

Deploy front


front-api/routes/poke/cache/index.ts
```diff
@@ -130,21 +138,27 @@ app.get("/", async (ctx): HandlerResult<GetPokeCacheResponseBody> => {
...
-          parsed = JSON.parse(rawValue);
-        } catch {
-          parsed = rawValue;
-        }
+      const keyType = await client.type(cacheKey);
+      const ttl = await client.ttl(cacheKey);
+
+      if (keyType === "hash") {
+        const fields = await client.hGetAll(cacheKey);
+        const value = Object.fromEntries(
```
fontanierh: nit: lodash mapValues slightly more readable imho

front/lib/resources/group_permission_resource.ts
```diff
@@ -461,25 +545,177 @@ export class GroupPermissionResource extends BaseResource<GroupPermissionModel>
...
+      grantType,
+      resourceType,
+      resourceId,
+    }));
+  }
+
+  // Takes no transaction: a read inside an unrelated transaction still uses the cache, while the
+  // mutations below keep their own transaction-scoped queries.
+  private static async listGrantsForGroups(
+    workspaceModelId: ModelId,
```
spolu: Ideally we pass an auth here vs a workspaceModelId ?

front/lib/resources/group_permission_resource.ts
```diff
@@ -461,25 +545,177 @@ export class GroupPermissionResource extends BaseResource<GroupPermissionModel>
...
+      );
+      statsDClient.increment("group_permissions_cache.read", 1, [
+        "result:error",
+      ]);
+      return this.loadGrantsForGroups(workspaceModelId, uniqueGroupModelIds);
+    }
+  }
+
+  private static async refreshGroupGrants(
+    workspaceModelId: ModelId,
```
spolu: Same here re auth

front/lib/resources/group_permission_resource.ts
```diff
@@ -461,25 +545,177 @@ export class GroupPermissionResource extends BaseResource<GroupPermissionModel>
...
+        "result:error",
+      ]);
+    }
+  }
+
+  // After commit only: inside the transaction the reload would see uncommitted rows.
+  // Overwrite instead of delete: grants change rarely, so one writer reloads instead of every
+  // concurrent reader missing at once.
+  private static async refreshGroupGrantsAfterCommit(
+    workspaceModelId: ModelId,
```
spolu: same here re auth

spolu (approved):
LGTM left a small comment on passing auth vs workspace ModelId

spolu:
I'm not convinced we need the rich approach you took here and would have likely started with a simple caching of all permissions with a TTL (even if it means a few possible bursts of queries). But I'm also fine going with this from the get go.

tdraier (approved):
looks great 👍

fontanierh (approved):
LGTM

id13:
> I'm not convinced we need the rich approach you took here and would have likely started with a simple caching of all permissions with a TTL (even if it means a few possible bursts of queries). But I'm also fine going with this from the get go.

@spolu I also prefer simple and stupid, my FUD is around retrieving 10k group perms per request on big workspaces (checked in the DB, totally plausible). I would like to avoid having to scale Redis (and in general having to think about redis) if we can at the price of a more complex approach 😄

front/vite.setup.ts
```diff
@@ -20,29 +20,87 @@ function attachHashCommands(
   client: Record<string, unknown>,
   hashStore: Map<string, Map<string, string>>
 ) {
-  const hSet = vi.fn(async (key: string, field: string, value: string) => {
+  // Mirrors node-redis: hSet takes either (key, field, value) or (key, fieldsObject).
+  const hSet = vi.fn(
```
flvndvd: Mind creating a dedicated mock for this like we've done for the file storage?

front-api/routes/poke/cache/index.ts
```diff
@@ -118,6 +119,14 @@ function resolveCacheKey(
   };
 }
 
+function decodeCacheValue(rawValue: string): unknown {
```
flvndvd, line 123: Can we use the `tryParseJSON` 🙏

front/lib/resources/group_permission_resource.ts
```diff
@@ -461,25 +562,176 @@ export class GroupPermissionResource extends BaseResource<GroupPermissionModel>
...
+      const grants = await this.loadGrantsForGroups(
+        workspaceModelId,
+        groupModelIds
+      );
+
+      const multi = redis.multi();
+      for (const [field, value] of encodeFields(groupModelIds, grants)) {
+        multi.hSet(key, field, value);
+      }
+      await multi.exec();
```
fabiencelier: Could we have a race condition issue here between read in DB and write in Redis ?
If there are 2 transactions happening simultaneously and perations are interleaved:
t1 -> write data1 in DB
t1 -> read data1 in DB
t2 -> write data2 in DB
t2 -> read data2 in DB
t2 -> write data2 in redis
t1 -> write data1 is redis (which is stale data compared overriden by data2)

front/lib/resources/group_permission_resource.ts
```diff
@@ -763,15 +1055,22 @@ export class GroupPermissionResource extends BaseResource<GroupPermissionModel>
     { grantType, resourceType }: CapabilitySpec,
     { transaction }: { transaction?: Transaction } = {}
   ): Promise<void> {
+    const workspaceId = auth.getNonNullableWorkspace().id;
+    const capabilityWhere = {
+      workspaceId,
+      grantType,
+      resourceType,
+      resourceId: WHOLE_TYPE_RESOURCE_ID,
+    };
```
flvndvd:
```suggestion
    const capabilityWhere: WhereOptions<GroupPermissionModel> = {
      workspaceId,
      grantType,
      resourceType,
      resourceId: WHOLE_TYPE_RESOURCE_ID,
    };
```

front/lib/resources/group_permission_resource.ts
```diff
@@ -99,6 +176,22 @@ export interface GroupPermissionResource
 export class GroupPermissionResource extends BaseResource<GroupPermissionModel> {
   static model: ModelStatic<GroupPermissionModel> = GroupPermissionModel;
 
+  static readonly cacheOperations = defineCacheOperations({
```
flvndvd, line 179: As this one does not actually cache the resource but only a subset of fields (grant) can we move it to its own file like what we've done for workspace seats https://github.com/dust-tt/dust/blob/main/front/lib/api/workspace_seats/cache.ts?

The goal being only actual resource caching lives in the Resource. And contract is pretty strict, it must cache the full resource.

id13: I'm not sure I agree if what we're caching is a result's method result 

If you start spreading cache surface outside the resource, don't you loose the benefit of the resource pattern?

Caching the whole resource is pretty strict imo, we want to cache w/e we need & the resource exposes

front/lib/resources/group_permission_resource.ts
id13: Yes, although unprobable, very bad if happens and it's about grants that should no longer exist but are in the cache ~forever.

Will instead invalidate, and let the read thundering herd happen, if it's too much for the load, we will reconsider 👍 

thanks!

front-api/routes/poke/cache/index.ts
id13: I guess you meant safeParseJSON

PR #30749, opened 2026-08-19
[front] perf: Add stats deps on groupId, workspaceId

id13:
## Description

Add a dependency stat on groupId / workspaceId to hint pg planner on the fact a group only belongs to ONE group.

Leads to better estimates of rows, so always useful, but also I don't expect a huge win on this codepath because of how simple the query is.


id13:
PMRR

PopDaph (commented):
👍🏻

PR #30750, opened 2026-08-19
sparkle: add icon prop to SliderToggle

avervaet:
## Description

Extends the Sparkle `SliderToggle` component with two optional presentation props for restricted or managed settings:

- `icon` accepts a React component and renders it inside the toggle knob. The Storybook example uses `Lock01` to make the restricted state visible without relying only on a surrounding tooltip.
- `faded` mutes the active track color when `selected` is true, using a translucent highlight. It has no effect on unselected toggles, and the existing disabled styling still takes precedence.

The change also updates the component documentation and adds a `SliderWithIcon` Storybook story covering unselected disabled, selected disabled, and selected faded states. Existing callers remain compatible because both props are optional.

**In Sparkle:** 
Capture d’écran 2026-08-19 à 10 48 18

## Risk

Low and limited to the Sparkle `SliderToggle` component. The API change is additive, with no changes required for existing callers and no data or backend behavior involved.

## Deploy Plan

- Deploy Sparkle

### DIFF
--- a/sparkle/src/components/SliderToggle.tsx
+++ b/sparkle/src/components/SliderToggle.tsx
@@ -7,6 +7,9 @@ type SliderToggleProps = {
   disabled?: boolean;
   selected?: boolean;
   icon?: ComponentType<{ className?: string }>;
+  // Mutes the active track color, e.g. to signal a setting that is on but
+  // restricted (paired with `icon`) rather than a plain active toggle.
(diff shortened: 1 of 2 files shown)


avervaet:

- #30750 ◀

adrsimon (approved):

PR #30757, opened 2026-08-19
[admin governance] remove group's members access from non editor type groups

ykmsd:
## Description
We decided to not store permission for group itself for editors kind (= agent editors, skill editors, space editors) and its permission will be the same as resource itself (e.g. can you edit a skill? then you can edit its skill editor group too).

Now remaining groups are: regular_manual, global, system (and regular_auto for model tiers? I haven't digged that one much yet). From what I understood, we don't expose these groups data to non-admin/non-managers. It's probably simple to not store permission for each group and let only admin/managers read it.

In group resource, sometimes we filter by canRead and sometimes we don't. I have no idea if this is intentional 🤔 (I even wonder if we really need canRead check, don't feel like confidential?) 

In group_resource:
- `listUserGroupsInWorkspace`: no canRead check (used in auth/group resolution, user.ts, workspace.ts, groups/members.ts). 
-  `listAllWorkspaceGroups`, `listForSpaceById`, `fetchByIds`: have canRead check, but all consumers are gated by role check (= available only to managers & admins) 

Here are the consumers of `api/w/wId/groups` (useGroups) according to Claude code:
```

┌────────────────────────┬─────────────────────────┬───────────────────────────────┬──────────────────────────────┐
│        Consumer        │     kinds requested     │         Route gating          │   Affected by your change?   │
├────────────────────────┼─────────────────────────┼───────────────────────────────┼──────────────────────────────┤
│                        │ default → global,       │ admin-only (RequireRoleLayout │ Requests the affected kinds  │
│ APIKeysPage            │ regular_auto,           │  requiredRole="admin")        │ — but admin reads all → no   │
│                        │ space_editors           │                               │ effect                       │
├────────────────────────┼─────────────────────────┼───────────────────────────────┼──────────────────────────────┤
│ GovernancePage         │ provisioned,            │ manager+                      │ No — managers read both      │
│                        │ regular_manual          │                               │                              │
├────────────────────────┼─────────────────────────┼───────────────────────────────┼──────────────────────────────┤
│ WorkspaceGroupsList    │ provisioned,            │ manager+                      │ No                           │
│                        │ regular_manual          │                               │                              │
├────────────────────────┼─────────────────────────┼───────────────────────────────┼──────────────────────────────┤
│ UsageFilterPanel       │ provisioned,            │ manager+                      │ No                           │
│                        │ regular_manual          │                               │                              │
├────────────────────────┼─────────────────────────┼───────────────────────────────┼──────────────────────────────┤
│ GroupsUsageTable       │ provisioned,            │ manager+                      │ No                           │
│                        │ regular_manual          │                               │                              │
├────────────────────────┼─────────────────────────┼───────────────────────────────┼──────────────────────────────┤
│ UsagePage              │ provisioned,            │ manager+                      │ No                           │
│                        │ regular_manual          │                               │                              │
├────────────────────────┼─────────────────────────┼───────────────────────────────┼──────────────────────────────┤
│ MemberGroupsSection    │ regular_manual          │ manager+                      │ No                           │
├────────────────────────┼─────────────────────────┼───────────────────────────────┼──────────────────────────────┤
│ DirectorySync          │ provisioned             │ admin/manager                 │ No                           │
├────────────────────────┼─────────────────────────┼───────────────────────────────┼──────────────────────────────┤
│ CreateOrEditSpaceModal │ provisioned             │ (space edit)                  │ No — managers/admins read    │
│                        │                         │                               │ provisioned                  │
├────────────────────────┼─────────────────────────┼───────────────────────────────┼──────────────────────────────┤
│ GroupSelectionTable    │ provisioned             │ (space edit)                  │ No                           │
└────────────────────────┴─────────────────────────┴───────────────────────────────┴──────────────────────────────┘

```

### DIFF
--- a/front-api/routes/w/[wId]/groups/index.ts
+++ b/front-api/routes/w/[wId]/groups/index.ts
@@ -45,6 +45,10 @@ app.get(
         : [kind]
       : ["global", "regular_auto", "space_editors"];
 
+    // Both list helpers filter by `canRead`. For role-only group kinds
+    // (regular_manual, provisioned, regular_auto, global, system) only
+    // admins/managers can read, so a plain member gets an empty list rather
+    // than the groups they belong to.
     const groups: GroupResource[] = spaceId
       ? await GroupResource.listForSpaceById(auth, spaceId, { groupKinds })
       : await GroupResource.listAllWorkspaceGroups(auth, { groupKinds });
--- a/front/lib/resources/group_resource.ts
+++ b/front/lib/resources/group_resource.ts
@@ -2554,25 +2554,23 @@ export class GroupResource extends BaseResource<GroupModel> {
   // Permissions
 
   /**
-   * Returns the requested permissions for this resource.
+   * We are migrating away from a group storing permissions for itself (the
+   * self-referential `groups: [{ id: this.id }]` entry): editor groups should
+   * derive access from their parent resource via `group_permissions`. Until that
+   * lands the branches below are mixed:
+   * - regular_manual / provisioned / regular_auto / global / system: role-only,
+   *   membership grants nothing. Admins get read/write/admin everywhere;
+   *   managers get read/write/admin on regular_manual and read on provisioned.
+   * - agent_editors / skill_editors / space_editors: still self-referential
+   *   (legacy) — members get full access through this ACL.
    *
-   * Configures two types of access:
-   * 1. Group-based: The group's members get read access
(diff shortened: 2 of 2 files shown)


matteotrab (approved):
LGTM pending the check on model_tier we talked about IRL

ykmsd:
Chatted with Thomas, too early to remove these without thinking how to let members read their own groups

PR #30766, opened 2026-08-19
Return node identity from file system writes and directory creations

flvndvd:
## Description

The server-side file system never tells its callers which node a write or a directory creation touched. Nodes have a stable id that survives every move and rename, and it is the only safe thing for a feature to hold onto when it must find the same file again later; a path goes stale the moment anything moves. Frames are the first consumer: they need to reference their source tree by identity rather than by the path stored today, which breaks live edits when a published frame's directory moves.

Writes now return the id of the node they created or overwrote, and directory creation returns it alongside the entry it already returned. The database backend already holds the node in both paths, so this surfaces information rather than fetching more. The GCS backend has no nodes and reports null, which is also what makes the degradation explicit for conversations not yet on the database file system.

The id rides next to the wire entry rather than on it, on purpose: the folder creation endpoint serializes that entry verbatim to the client, and numeric ids stay inside the server. The one caller that fed an entry to the wire unwraps it and sends exactly the same JSON as before. Every other caller only ever checked for errors and is unchanged.

## Tests

A database backend test pins the property this exists for rather than the plumbing: a directory and a file both report ids, overwriting the file reports the same node rather than a new one, and after moving the file from a conversation to its Pod, fetching by the stored id still finds it. A GCS backend test pins the null. The existing file system, projects and file tool suites all pass.

Also fixes a test of mine from an earlier PR that assigned NODE_ENV directly, which current Node types reject; it stubs the variable through vitest instead.

## Risk

None observable: no endpoint changes its response, no schema changes, and no caller behaves differently. The change widens an internal return type that callers were ignoring. Rollback is a revert.

## Deploy Plan

Nothing to sequence.

### DIFF
--- a/front/lib/api/config.test.ts
+++ b/front/lib/api/config.test.ts
@@ -1,5 +1,5 @@
 import config from "@app/lib/api/config";
-import { afterEach, describe, expect, it } from "vitest";
+import { afterEach, describe, expect, it, vi } from "vitest";
 
 const PUBLIC_URL = "https://eu.dust.tt";
 const INTERNAL_URL = "http://front-internal-service";
@@ -8,6 +8,7 @@ describe("getSandboxApiBaseUrl", () => {
   const originalEnv = { ...process.env };
 
   afterEach(() => {
+    vi.unstubAllEnvs();
     process.env = { ...originalEnv };
   });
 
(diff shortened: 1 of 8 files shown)


davidebbo (approved):
LGTM

PR #30767, opened 2026-08-19
Add observability for LLM success without usage

flvndvd:
## Description

In our endless journey to debug attribution items, we flagged that some runs are being flagged as successful without any valid run usages associated. This PR aims at instrumenting when this happens to help debug/understand/fix it.

### DIFF
--- a/front/lib/api/llm/llm.ts
+++ b/front/lib/api/llm/llm.ts
@@ -271,6 +271,8 @@ export abstract class LLM<
           continue;
         }
 
+        const { tokenUsage, ...rest } = buffer.currentOutput;
+
         // Logging before it gets stopped and retried downstream
         if (currentEvent.type === "error") {
           // Temporary: track LLM error metric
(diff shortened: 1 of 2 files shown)


front/lib/api/llm/llm.ts
```diff
@@ -298,18 +300,45 @@ export abstract class LLM<
...
+            this.generation.updateTrace({
+              tags: ["success_without_usage:true"],
+            });
+            this.generation.update({
+              level: "WARNING",
+              statusMessage:
+                "LLM completed successfully without reporting token usage.",
+            });
+            logger.warn(
+              {
```
aubin-tchoi, line 332: would be useful to have the convo and message IDs here

aubin-tchoi (approved):
LGTM

front/lib/api/llm/llm.ts
flvndvd: I can add them but betting on Langfuse to have all the information directly.

fabiencelier (approved):

PR #6469, opened 2024-07-24
feat: new viz service

fontanierh:
## Description

- add a skeleton next JS (app router) app for the new `viz` service
- add the infra files, including a managed certificate and ingress rules for this new service
- add a Dockerfile and a deployment GH action for the new service

## Risk

N/A (doesn't affect any prod system)

## Deploy Plan

- apply infra (new deployment will be stuck in PullImageErr since the image doesn't exist yet)
- run the new github action to deploy
- monitor the new managed cert until it gets provisioned

PR #30769, opened 2026-08-19
[viz] chore: remove unused dd-trace dependency

avervaet:
## Description

Remove the unused `dd-trace` dependency from `viz/package.json` and `package-lock.json`.

`dd-trace` was added to `viz/package.json` the day the workspace was created (#6469, then #6473, both 2024-07-24) but there has never been a require/import of `dd-trace` anywhere in `viz/` — it looks like it was declared defensively at the time (matching root/front/connectors, which do instrument with `dd-trace`) but never actually wired up in this service. It's one of the heavier deps in the monorepo (APM tracer with dozens of auto-instrumentation plugins), so dropping it here trims a meaningful chunk of `viz`'s install size.

## Tests

Verified with `knip --dependencies` and a repo-wide grep that `dd-trace` is not required/imported anywhere under `viz/`. Ran `npm install --package-lock-only -w viz` to regenerate the lockfile.

## Risk

Low: dependency-only change with no runtime code changes.

## Deploy Plan

None — no runtime behavior change.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -58816,7 +58816,6 @@
         "clsx": "^2.1.1",
         "cmdk": "^1.1.1",
         "date-fns": "^4.1.0",
-        "dd-trace": "^5.87.0",
         "embla-carousel-react": "^8.6.0",
         "html-to-image": "^1.11.11",
         "input-otp": "^1.4.2",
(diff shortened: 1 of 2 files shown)


avervaet:

- #30769 ◀

avervaet:
Closing for now — dd-trace may be wired via infra-level require flag / Datadog APM rather than a code import; need to verify traces aren't actually flowing for viz before removing this.

PR #30777, opened 2026-08-19
feat(agent_inactivity): add inactive agent archival business rules

achilleburah:
## Description

First brick of the opt-in "automatically archive inactive agents" feature: the business rules and the application layer that drives them. No PostgreSQL, Temporal, HTTP or adapters yet — this only encodes and tests the agreed rules.

- inactivity/policy.ts: domain rules. Threshold validation (integer, 2 days minimum, no implicit default), cutoff resolution and per-agent eligibility, with machine-readable exclusion reasons.
- inactivity/evaluate_inactive_agents.ts: application layer. Splits a batch of agents into eligible ids and reasoned exclusions, against a cutoff resolved once per operation so it cannot drift mid-batch.

An invalid threshold is returned as an Err for the whole operation rather than as a per-agent exclusion, so a misconfigured workspace can never be mistaken for "no agent to archive".

Activity is read from mentions rather than agent messages: mention rows carry the stable agent sId with no configuration version, their createdAt is never null, and (workspaceId, agentConfigurationId, createdAt) is already indexed. Any mention counts, whatever became of the run it started.

## Tests

Business logic, unit tests only. 

## Risk

Not called yet. 

## Deploy Plan

None

### DIFF
--- a/front/lib/api/assistant/inactivity/archive_inactive_agents.ts
+++ b/front/lib/api/assistant/inactivity/archive_inactive_agents.ts
@@ -1,0 +1,234 @@
+import {
+  archiveAgentConfiguration,
+  getAgentConfigurations,
+} from "@app/lib/api/assistant/configuration/agent";
+import type {
+  AgentArchivalExclusionReason,
+  AgentInactivityPolicyError,
+  AgentInactivitySnapshot,
+  AgentTriggerSnapshot,
+} from "@app/lib/api/assistant/inactivity/policy";
+import {
+  computeInactivityCutoffAt,
+  evaluateAgentArchivalEligibility,
+} from "@app/lib/api/assistant/inactivity/policy";
+import type { Authenticator } from "@app/lib/auth";
+import { TriggerResource } from "@app/lib/resources/trigger_resource";
+import logger from "@app/logger/logger";
+import type { AgentConfigurationStatus } from "@app/types/assistant/agent";
+import type { Result } from "@app/types/shared/result";
+import { Err, Ok } from "@app/types/shared/result";
+
+/**
+ * Archives a workspace's inactive agents: resolve the threshold into a cutoff, load the agents,
+ * judge each one against the rules in `policy.ts`, and archive those that are eligible.
+ *
+ * The manual endpoints and the nightly Temporal activity both enter here, so the rules cannot drift
+ * between them. The population can: every read is permission-filtered, so a manual run only ever
+ * considers the agents its caller can see, where the nightly run sees the whole workspace.
+ *
+ * The rules are applied against state re-read for the page rather than against what the fetch
+ * loaded, because the two do not answer the same question: the fetch finds candidates, the re-read
+ * goes through `getAgentConfigurations` and is therefore permission-filtered. That is also what
+ * makes a Temporal retry safe — replaying the same input re-reads an agent the first attempt
+ * archived, sees `archived`, and skips it instead of emitting a second `agent.archived` audit event.
+ *
+ * Archiving itself stays the existing `archiveAgentConfiguration` primitive, which owns the side
+ * effects (trigger disabling, wake-up cancellation, editor group suspension, audit event).
+ */
+
+/**
+ * Why an agent this operation looked at was not archived: the rules' own exclusions, plus the two
+ * outcomes only the mutation path can produce and that the rules therefore never see.
+ */
+export type AgentArchivalSkipReason =
+  | AgentArchivalExclusionReason
+  // The configuration disappeared, or is not visible to this actor, between the fetch and the
+  // re-read.
+  | "agent_not_found"
+  // The archival update matched no row: someone else archived the agent in the meantime.
+  | "archive_raced";
+
+export interface AgentArchivalSkip {
+  agentId: string;
+  reason: AgentArchivalSkipReason;
+}
+
+export interface InactiveAgentsArchivalInput {
+  thresholdDays: number;
+  evaluatedAt: Date;
+}
+
+export interface InactiveAgentsArchival {
+  evaluatedAt: Date;
+  cutoffAt: Date;
+  archivedAgentIds: string[];
+  skipped: AgentArchivalSkip[];
+}
+
+export type InactiveAgentsArchivalResult = Result<
+  InactiveAgentsArchival,
+  AgentInactivityPolicyError
+>;
+
+/** The facts the rules read about an agent, as they stand at the moment the page is judged. */
+interface AgentArchivalFacts {
+  status: AgentConfigurationStatus;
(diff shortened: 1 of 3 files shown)


front/lib/api/assistant/inactivity/policy.ts
```diff
@@ -0,0 +1,247 @@
...
+ * Module-private: callers validate a threshold by resolving a cutoff from it, so there is exactly one
+ * way to be told a policy is unusable. The Governance form validates against
+ * `MIN_INACTIVITY_THRESHOLD_DAYS` directly.
+ */
+function isValidInactivityThresholdDays(thresholdDays: number): boolean {
+  return (
+    Number.isInteger(thresholdDays) &&
+    thresholdDays >= MIN_INACTIVITY_THRESHOLD_DAYS
+  );
+}
```
fabiencelier, line 113: Maybe we could add a MAX too to avoid scaning for years of data everyday, something like 366 days ?

front/lib/api/assistant/inactivity/policy.ts
```diff
@@ -0,0 +1,247 @@
...
+ * - Opt-in per workspace : no threshold configured, nothing archived. There is no default.
+ * - Threshold of 2 days minimum : a whole number of days; anything else archives nothing.
+ * - Mentions are the only activity : any mention resets it, whatever became of the run it started.
+ * - Creating or editing an agent is not activity : only being reached for counts.
+ * - Inactive means last mentioned before the cutoff : never mentioned counts as inactive.
+ * - Only active agents are archivable : archived, draft and pending ones are left alone.
+ * - Schedules are exempt : never archive an agent a schedule still drives, whatever its frequency —
+ *   including one Dust itself paused (relocation, plan downgrade), which the workspace never chose
+ *   to stop.
+ * - Wake-ups are not exempt : a pending wake-up does not protect an agent, and archiving cancels it.
```
fabiencelier, line 30: I think we are missing one rule also: agent must have existed since at least the cutoff date, otherwise we may archive just created agents without messages

front/lib/api/assistant/inactivity/policy.ts
```diff
@@ -0,0 +1,247 @@
...
+  // threshold of N days always leaves a full N days of grace.
+  if (agent.lastMentionedAt.getTime() < cutoffAt.getTime()) {
+    return { eligible: true };
+  }
+
+  return {
+    eligible: false,
+    reason: "recent_mention",
+  };
+}
```
fabiencelier, line 240: I don't disagree with the rules, but that mean you will compute the `lastMentionedAt` which is pretty expensive for all agents, including already archived ones and the onew with trigger :/
I think we want a flow where we avoid computing it entirely when not needed

I was more thinking about a stream of operations:
 - fetch all agents with archivable status
 - filter out those with triggers
 - compute the last mentioned at for these
 - filter out the one recently mentioned
 - archive them 

So maybe we can keep the rules and helpers method in here, but we will never have to evaluate this function with all the info at once

Unless computing the last mention date for all agents will be the same speed as computing it for a subset of selected agents ?

What do you think ?

front/lib/api/assistant/inactivity/archive_inactive_agents.ts
```diff
@@ -0,0 +1,234 @@
...
+    // processing time: an agent mentioned or restored during it is archived anyway, and has to be
+    // restored by hand. Closing that window properly means a compare-and-set in
+    // `archiveAgentConfiguration` — an `expectedStatus` narrowing its `where` clause, which would
+    // make the mutation itself refuse an agent that stopped being active. Deliberately not done:
+    // it moves a business rule out of `policy.ts` and into a SQL predicate, and the exposure is one
+    // reversible archival on a page that takes seconds. Revisit if pages get long.
+    const archived = await archiveAgentConfiguration(auth, agentId);
+    if (!archived) {
+      skipped.push({ agentId, reason: "archive_raced" });
+      continue;
```
fabiencelier, line 204: ok to begin with but maube we should use a batch archive method to archive all of them at once and reduce DB load

front/lib/api/assistant/inactivity/policy.ts
```diff
@@ -0,0 +1,247 @@
...
+ * on the same date therefore judge every agent identically whatever time they fire — a nightly job at
+ * 02:00 and an admin previewing at 17:00 agree — and an agent cannot become eligible partway through a
+ * day. Truncating downwards also only ever grants more grace than the threshold asks for, never less.
+ *
+ * Callers resolve the cutoff once per operation, from a single `evaluatedAt`, and reuse it for every
+ * agent of that operation. Failing here means the workspace policy is unusable and nothing may be
+ * archived — the error is deliberately not expressible as a per-agent exclusion, since the threshold
+ * is a property of the workspace and not of any agent.
+ */
+export function computeInactivityCutoffAt({
```
fabiencelier: overall it's nice to have doc but it may be a bit too verbose sometimes ^^
Maybe we can drop the last paragraph here for instance as it repeating a bit what the code is doing

front/lib/api/assistant/inactivity/policy.ts
achilleburah: Hmmm we could add a MAX for conventions and to avoid a typo (ie 3650) but I think the bigger the number, the less agents we find therefore the less we iterate ?

front/lib/api/assistant/inactivity/policy.ts
achilleburah: Oh that's true, i implemented it in the fetch (next PR) but I missed it here, I'll see if if can move it up

front/lib/api/assistant/inactivity/archive_inactive_agents.ts
achilleburah: Yeah the batching is my main problem rn. The thing is I'd like to have atomic business rules for each agents to allow for better granularity in error handling / audit logs...

front/lib/api/assistant/inactivity/policy.ts
achilleburah: Yeah we're missing a couple of rules that were not implemented in the policy but in the fetch for performance reason. 
I direclty filter on the "active" agents and proceed to compute the lastMentionned directly from the query. Sending you the 2nd PR so you can have a better overview and we can talk about this.

fabiencelier: so this serves as agate to make sure we don't archive something unwanted after doing all the queries ?

fabiencelier (approved):

front/lib/api/assistant/inactivity/policy.ts
achilleburah: yeah, it was a way to centralize all the business rules in the same place

PR #30774, opened 2026-08-19
[front] enh: make vs prev column sortable in consumption attribution table

avervaet:

PR #30779, opened 2026-08-19
[front] enh: sort consumption attribution table server-side across pages

avervaet:
## Summary

Stacked on #30774 (client-side sortable vs-prev column). This makes sorting the consumption attribution table actually correct across pages, not just within the currently loaded page.

- Adds `sortBy` (`credits` | `avgCredits` | `vsPrev`) / `sortOrder` (`asc` | `desc`) to `ConsumptionTopBody`, threaded through all 8 `top-*` endpoints down to `fetchConsumptionTopGroups`.
- `credits`/`avgCredits` sort by ordering the Elasticsearch terms aggregation on the metric (a `bucket_script` for the avg-per-unit ratio).
- `vsPrev` sort widens the query window to include the previous period and computes current/previous sums per bucket in the same aggregation via a `bucket_script` growth metric, ranking on that directly — this replaces the separate previous-credits lookup for that sort (it's already known from the ranking query).
- The consumption attribution table's header sorting is now controlled state, shared between the table's local re-sort and the params sent to `useConsumptionTop`, with pagination reset to page 1 on a sort change.

**Known limitation, called out on purpose:** sorting by **name** still only reorders whatever page is currently loaded, not the full dataset. Display names are resolved from Postgres *after* Elasticsearch has already picked the ranked page (the ES documents only carry ids, not display names), so a true global name sort would need indexing names into Elasticsearch or resolving labels for every matching key before paginating — both bigger changes than this PR's scope. Flagging this explicitly since a user could reasonably expect it to behave like the other columns.

## Test plan
- [x] `front` and `front-api` typecheck clean
- [x] Extended `top.test.ts` with new tests for `avgCredits` and `vsPrev` server-side sort (order clauses, bucket_script paths, widened query window, growth computation, no second previous-credits lookup on `vsPrev`), plus existing tests updated for the new nested aggregation shape — 17/17 passing
- [ ] Manual verification in a workspace with real Elasticsearch data (recommend before merge, given the aggregation shape can't be fully verified against a mocked ES client)

### DIFF
--- a/front-api/routes/w/[wId]/analytics/consumption/top-agents.ts
+++ b/front-api/routes/w/[wId]/analytics/consumption/top-agents.ts
@@ -23,7 +23,7 @@ app.post(
   validate("json", ConsumptionTopBodySchema),
   async (ctx): HandlerResult<GetConsumptionTopAgentsResponse> => {
     const auth = ctx.get("auth");
-    const { limit, offset, search, filter, ...periodQuery } =
+    const { limit, offset, search, filter, sortBy, sortOrder, ...periodQuery } =
       ctx.req.valid("json");
 
     const period = await resolveConsumptionPeriod(
@@ -37,6 +37,8 @@ app.post(
       offset,
       search,
       filter,
+      sortBy,
+      sortOrder,
     });
     if (result.isErr()) {
       logger.error(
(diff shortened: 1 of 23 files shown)


avervaet:

- #30774
    - #30779 ◀

PR #30781, opened 2026-08-19
front fix: skill mount failure check & fallback

xebeohp:
## Description

Fixes https://github.com/dust-tt/tasks/issues/10111.

When `enable_skill` mounted a Skill’s attached files, it persisted the Skill as enabled before confirming that the mount succeeded. If a file download or write failed, the conversation could contain a partial Skill mount while the Skill was still marked as enabled.

A subsequent `enable_skill` call then returned `already enabled` and skipped the file mount, leaving the conversation without a recovery path.

Change:

1. Check: look up every expected file path; if a file already exists, leave it alone.
2. Write: write only the files that are missing.
3. Fail clean: if any write fails, delete whatever this call just wrote (pre-existing files stay untouched) and return an error, so the skill never ends up half-mounted.

This covers the customer case where the FPSH Skill encountered a GCS socket hang-up and only part of its file set was copied. On retry, the existing files are kept and the missing files are mounted.

## Tests

Added and updated unit tests covering:

- Initial file-mount failures.
- Retry of an already-enabled Skill with missing files.
- Successful loading of missing files while preserving existing files.
- Failed retries.
- Skills without files.
- Skills that are not equipped for the Agent.

## Risk

Low. The change is limited to Skill activation and conversation file mounting.

## Deploy Plan

Deploy `front`.

### DIFF
--- a/front/lib/api/actions/servers/skill_management/tools/index.ts
+++ b/front/lib/api/actions/servers/skill_management/tools/index.ts
@@ -10,7 +10,9 @@ import type { Authenticator } from "@app/lib/auth";
 import { SkillResource } from "@app/lib/resources/skill/skill_resource";
 import { extractUniqueSkillIds } from "@app/lib/skills/format";
 import type { AgentLoopExecutionData } from "@app/types/assistant/agent_run";
+import type { ConversationWithoutContentType } from "@app/types/assistant/conversation";
 import { isUserMessageType } from "@app/types/assistant/conversation";
+import type { Result } from "@app/types/shared/result";
 import { Err, Ok } from "@app/types/shared/result";
 import assert from "assert";
 
@@ -37,6 +39,21 @@ function extractSkillIdsFromConversationMessages(
   }
 
   return [...userMessageSkillIds];
+}
+
+async function ensureSkillFilesMounted(
+  auth: Authenticator,
+  skill: SkillResource,
+  conversation: ConversationWithoutContentType
+): Promise<Result<{ loadedPaths: string[] }, Error>> {
+  if (!skill.hasFiles()) {
+    return new Ok({ loadedPaths: [] });
+  }
+
+  return loadSkillFilesToConversation(auth, {
+    skill,
+    conversation,
+  });
 }
 
 async function findAvailableSkillForAgentLoop({
@@ -133,9 +150,28 @@ const handlers: ToolHandlers<typeof SKILL_MANAGEMENT_TOOLS_METADATA> = {
 
     if (!skill) {
       return new Err(
-        new MCPError(`Skill "${skillName}" not found`, {
+        new MCPError(`Skill "${skillName}" is not equipped for this agent.`, {
           tracked: false,
         })
+      );
+    }
+
+    // Mount the skill's files before persisting the enablement, since persistingit before the
+    // mount is confirmed would let those consumers see the skill as enabled eventhough its
+    // files never landed.
+    const mountResult = await ensureSkillFilesMounted(
+      auth,
+      skill,
+      conversation
+    );
+
+    if (mountResult.isErr()) {
+      // Returns a failure, not a successful enablement with a warning. A retried enable_skill call
+      // will attempt the mount again since it is idempotent.
+      return new Err(
+        new MCPError(
+          `Failed to mount files for skill "${skill.name}": ${mountResult.error.message}`
+        )
       );
     }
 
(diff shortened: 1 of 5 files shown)


front/lib/api/skills/conversation_files.ts
```diff
@@ -73,18 +77,38 @@ export async function loadSkillFilesToConversation(
...
+  const expectedPaths = files.map(
+    (file) =>
+      `${conversationMount.scopedPrefix}/skills/${skill.name}/${file.fileName}`
+  );
+
   const loadedPaths: string[] = [];
 
-  for (const file of files) {
-    const scopedPath = `${conversationMount.scopedPrefix}/skills/${skill.name}/${file.fileName}`;
+  for (const [index, file] of files.entries()) {
```
aubin-tchoi: you can have a new file that was added to the skill since when it was enabled and end up in a weird state where you partially mounted the files

aubin-tchoi: what about doing a first loop that checks that none of the file exist, then if so do a second pass where you write the files?

front/lib/api/skills/conversation_files.ts
```diff
@@ -73,18 +77,38 @@ export async function loadSkillFilesToConversation(
...
-    );
-    if (writeResult.isErr()) {
-      return writeResult;
+    if (!existsResult.value) {
+      const writeResult = await fileSystem.write(
+        scopedPath,
+        file.getContent(),
+        file.contentType
+      );
+      if (writeResult.isErr()) {
```
aubin-tchoi: no cleanup of the other files if one of them errors?
will be weird for the agent to leave some of the files if the skill enablement has failed

front/lib/api/skills/conversation_files.ts
```diff
@@ -73,18 +77,38 @@ export async function loadSkillFilesToConversation(
     throw new Error("Conversation mount not found.");
   }
 
+  const expectedPaths = files.map(
+    (file) =>
+      `${conversationMount.scopedPrefix}/skills/${skill.name}/${file.fileName}`
```
aubin-tchoi: nit: we have a file where we centralize every mount path

front/lib/api/skills/conversation_files.ts
```diff
@@ -46,12 +50,12 @@ export async function loadSkillFilesToConversation(
     ...skill.getFileAttachments().map((file) => ({
       fileName: file.fileName,
       contentType: file.contentType,
-      content: file.getReadStream({ auth, version: "original" }),
+      getContent: () => file.getReadStream({ auth, version: "original" }),
```
aubin-tchoi, line 57: OOC what does this change do?

front/lib/api/actions/servers/skill_management/tools/index.ts
```diff
@@ -39,6 +41,21 @@ function extractSkillIdsFromConversationMessages(
...
+async function ensureSkillFilesMounted(
+  auth: Authenticator,
+  skill: SkillResource,
+  conversation: ConversationWithoutContentType
+): Promise<Result<{ loadedPaths: string[] }, Error>> {
+  if (!skill.hasFiles()) {
+    return new Ok({ loadedPaths: [] });
+  }
+
+  return loadSkillFilesToConversation(auth, {
```
aubin-tchoi: this one should now be called upsert rather than load

front/lib/api/actions/servers/skill_management/tools/index.ts
```diff
@@ -39,6 +41,21 @@ function extractSkillIdsFromConversationMessages(
   return [...userMessageSkillIds];
 }
 
+async function ensureSkillFilesMounted(
```
aubin-tchoi: nit: the name of the function does not carry the idea that it actually does things (write data to the sandbox)

aubin-tchoi (commented):
directionally aligned, a few things to clarify

front/lib/api/actions/servers/skill_management/tools/index.ts
xebeohp: Fixed.

front/lib/api/skills/conversation_files.ts
xebeohp: Fixed

front/lib/api/skills/conversation_files.ts
xebeohp: Fixed.

front/lib/api/actions/servers/skill_management/tools/index.ts
xebeohp: renamed to mountSkillFilesToConversation

front/lib/api/skills/conversation_files.ts
xebeohp: It lets a file that's already mounted never have its GCS read stream opened, so there are no wasted GCS read streams

front/lib/api/actions/servers/skill_management/tools/index.ts
```diff
@@ -133,47 +150,45 @@ const handlers: ToolHandlers<typeof SKILL_MANAGEMENT_TOOLS_METADATA> = {
 
     if (!skill) {
       return new Err(
-        new MCPError(`Skill "${skillName}" not found`, {
+        new MCPError(`Skill "${skillName}" is not equipped for this agent.`, {
```
aubin-tchoi: let's keep the old message, the skill does not need to be equipped

front/lib/api/skills/conversation_files.ts
aubin-tchoi: good addition!

front/lib/api/actions/servers/skill_management/tools/index.ts
```diff
@@ -133,47 +150,45 @@ const handlers: ToolHandlers<typeof SKILL_MANAGEMENT_TOOLS_METADATA> = {
...
+        ? "Skill files successfully loaded:\n" +
+          mountResult.value.loadedPaths.map((p) => `  - ${p}`).join("\n")
+        : null;
 
     const text =
-      `Skill "${skill.name}" has been enabled.` +
+      (wasAlreadyEnabled
+        ? `Skill "${skill.name}" was already enabled.`
+        : `Skill "${skill.name}" has been enabled.`) +
       (fileMessage ? `\n\n${fileMessage}` : "");
```
aubin-tchoi: if the skill was already enabled we can't have a `fileMessage` no?

aubin-tchoi: actually there's the very niche case where new files were added to the skill since the initial enablement. Let's customize the message in this case to say that the skill was already enabled but some files were missing and have been additionally loaded (exact copy tbd)

front/lib/api/skills/conversation_files.ts
```diff
@@ -65,30 +73,63 @@ export async function loadSkillFilesToConversation(
...
+  const missing: { file: WritableSkillFile; scopedPath: string }[] = [];
 
-  for (const file of files) {
-    const scopedPath = `${conversationMount.scopedPrefix}/skills/${skill.name}/${file.fileName}`;
+  for (const [index, scopedPath] of expectedPaths.entries()) {
+    const existsResult = await fileSystem.exists(scopedPath);
+    if (existsResult.isErr()) {
+      return new Err(
+        new Error(
+          `Failed to check skill file(s): ${expectedPaths.slice(index).join(", ")} (${existsResult.error.message})`
```
aubin-tchoi: nit: easy to know whether or not to put an `s`

aubin-tchoi (approved):
Looks great!

PR #30785, opened 2026-08-19
Drop the deprecated Pod function result callback route

davidebbo:
## Description

`front-api/routes/v1/w/[wId]/sandbox/sandbox-functions/result.ts` was marked deprecated on 2026-08-10, pending removal. dsbx 0.1.46 dropped callback delivery — every result now comes back on the exec's own stdout — and `DSBX_CLI_VERSION` has since moved to 0.1.51, so this removes the route and everything that falls off with it.

**Deleted**

- The route and its test, plus the `sandbox-functions` sub-app: it only mounted `/result`, so the whole directory goes, along with its `app.route(...)` in `front-api/routes/v1/w/[wId]/sandbox/index.ts`.
- The callback arm in `normalizeSandboxFunctionResult` that the deprecation note called out: `LegacySandboxFunctionRunnerOutputSchema`, the HTTP-response-shaped output (`{ok: true, response: {status, headers, body, encoding}}` / `{ok: false, error: {kind}}`) from the pre-v3 runner image, which only ever arrived over the callback body. `normalizeRunnerOutcome` is now the current runner schema plus the invalid-envelope rejection. No error code is orphaned — `threw`, `http_error`, `invalid_output` and friends all still live in `SANDBOX_FUNCTION_RUNNER_ERROR_CODES`.

**Kept**, because they have live callers beyond this route: the `function_invocation` sandbox token kind and `isSandboxFunctionInvocationTokenPayload` (used by `/sandbox/actions/*`), `SandboxTokenFactory`'s invocation-token helpers, `access: "system"` on `SandboxFunctionInvocationResource` (used by the Temporal activities and `sandbox_function_resource`), and the bare-outcome fall-through in `normalizeSandboxFunctionResult`, which stdout parsing still relies on for older dsbx. Comments naming the now-gone callback path were updated.

**Not in scope:** `front` still passes `--result-delivery stdout` and dsbx still accepts-and-ignores it. The comment in `cli/dust-sandbox/src/commands/function/mod.rs` says to remove the flag *with* front's argument, not before — that's a separate coordinated change.

## Risk

Low, it's been long enough that no running sandbox should be use it.

## Deploy Plan

Deploy front-api.


fontanierh (approved):
Thank you 🙏

PR #30788, opened 2026-08-19
Attribute hidden helper credits to parent agents

flvndvd:
## Description

This PR keeps moving towards removing the `dust-task` sub agent from all analytics places (previous PR https://github.com/dust-tt/dust/pull/30829). This one PR focuses around adding the new field in the ES mapping so we can backfill before switching read queries.

## Deploy Plan

This will require to update the existing mapping before running the backfill. Another PR will them switch the read to this field.


adrsimon (approved):
LGTM

PR #30793, opened 2026-08-19
[front] fix: show -- instead of N.A for missing vs-prev data

avervaet:
## Description

In the consumption attribution table, the `vs prev` cell displayed `N.A` when there was not enough prior-period data to compute the comparison. This changes the displayed placeholder to `--` while keeping the existing `Not enough data to compute` tooltip. No calculation or data-handling behavior is changed; this is a presentation-only adjustment.

Capture d’écran 2026-08-19 à 15 28 50

## Deploy Plan

- Deploy front


ClementAupiais (approved):
LGTM

PR #30794, opened 2026-08-19
[front] enh: sort consumption attribution table server-side across pages

avervaet:
## Description

The consumption attribution table previously sorted only the rows loaded for the current page, which made sorting across pagination misleading. This PR threads a `sortOrder` parameter through the shared consumption-top hook, schema, all top-* API routes, dimension fetchers, and `fetchConsumptionTopGroups`. The credits column is now ranked in Elasticsearch before pagination, supporting both ascending and descending order; the default remains descending. The UI keeps sorting state controlled and resets to the first page whenever the sort changes.

**Warning:** Ordering terms by a sum in ascending order is not guaranteed to return the true globally smallest buckets because shard-local candidate selection can omit buckets. The pagination loop then excludes each returned batch and treats it as the global order, so an inaccurate first batch can propagate to every subsequent page.

This seems to be an acceptable trade-off as most of the user will probably sort by `desc`anyway.

## Tests

Added coverage in `front/lib/api/analytics/consumption/top.test.ts` to verify that ascending sorting sends the `credit_micro` metric with ascending order to the Elasticsearch terms aggregation.

## Risk

Low This is an analytics-only read-path change with no data or schema migrations. The default descending credits order preserves existing behavior, while credits sorting now correctly ranks the full result set across pages.

## Deploy Plan
- Deploy front

### DIFF
--- a/front/components/workspace/analytics/consumption/ConsumptionAttributionTable.tsx
+++ b/front/components/workspace/analytics/consumption/ConsumptionAttributionTable.tsx
@@ -78,12 +78,6 @@ const DEFAULT_ATTRIBUTION_SORTING: SortingState = [
   { id: "credits", desc: true },
 ];
 
-// Only the credits column is ranked server-side (see
-// fetchConsumptionTopGroups): Elasticsearch's terms aggregation can only
-// order by a genuine metric, not a ratio like avgCredits or vs-prev growth,
-// which would need a bucket_script — terms cannot order by a pipeline
-// aggregation. Sorting by any other column still reorders whatever page is
-// currently loaded, not the full dataset.
 const ATTRIBUTION_SERVER_SORTABLE_COLUMN_ID = "credits";
 
 type AttributionTransitionDirection = -1 | 0 | 1;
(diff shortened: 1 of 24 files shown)


avervaet:

- #30794 ◀
    - #30795
        - #30796

front/components/workspace/analytics/consumption/ConsumptionAttributionTable.tsx
```diff
@@ -522,6 +551,8 @@ function AttributionRows({
             skeletonRowCount={skeletonRowCount}
             hasAvatar={hasAvatar}
             isAvatarRounded={dimension === "user"}
+            sorting={sorting}
+            onSortingChange={handleSortingChange}
```
aubin-tchoi, line 555: qq: is the goal to add sorting to the other ones in follow up PRs where you'll also drill down the sort field, or to only support sorting on this one (in which case we should make the other ones non sortable)

aubin-tchoi (approved):
LGTM

front/components/workspace/analytics/consumption/ConsumptionAttributionTable.tsx
avervaet: I'm adding other columns in subsequent prs. Will only remove sorting on users.

PR #30795, opened 2026-08-19
[front] enh: sort consumption attribution table server-side by cost share

avervaet:
## Description

This stacked follow-up to [#30794](https://github.com/dust-tt/dust/pull/30794) makes the **Consumption share** column in the Consumption Attribution table eligible for the same server-side ranking as **Credits**.

`costShare` is computed as `row.credits / totalCredits`, where `totalCredits` is the same denominator for every row. Therefore, sorting by cost share produces the same order as sorting by credits. The change adds a `costShare` accessor and includes it in the set of server-sortable columns, so ascending/descending pagination uses the existing credit-based server ranking. 

## Risk

Low. This is a small analytics UI/read-path change with no migrations or writes.

## Deploy Plan

Deploy `front` after [#30794](https://github.com/dust-tt/dust/pull/30794), which is the base PR for this stack.

### DIFF
--- a/front/components/workspace/analytics/consumption/ConsumptionAttributionTable.tsx
+++ b/front/components/workspace/analytics/consumption/ConsumptionAttributionTable.tsx
@@ -78,7 +78,10 @@ const DEFAULT_ATTRIBUTION_SORTING: SortingState = [
   { id: "credits", desc: true },
 ];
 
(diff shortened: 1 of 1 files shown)


avervaet:

- #30794
    - #30795 ◀
        - #30796
            - #30809
                - #30810

aubin-tchoi (approved):
LGTM, smart to use the fact that x -> x/a is monotonically increasing 👍🏻

PR #30796, opened 2026-08-19
[front] enh: sort consumption attribution table server-side by avg credits

avervaet:
## Description

This is the third PR in the stack [#30794](https://github.com/dust-tt/dust/pull/30794) → [#30795](https://github.com/dust-tt/dust/pull/30795) → **#30796**. It adds support for ranking the Consumption Attribution table by average credits per unit.

The optional `sortBy` parameter now flows from the table through the shared consumption hook, schema, all top-* API routes, and dimension fetchers. It supports `credits` (the default) and `avgCredits`, while the existing server-side credits ranking remains unchanged for callers that do not provide `sortBy`.

For invocation-based dimensions such as tools and skills, the ranking uses an Elasticsearch `avg(credit_micro)` sub-aggregation and orders buckets directly by that metric. For message-based dimensions, average credits are computed as the bucket's credits divided by its distinct message count. Elasticsearch terms aggregations cannot order by that ratio, so this path fetches all matching buckets in batches, re-ranks them in memory, and applies pagination after sorting. The extra average aggregation is only requested when it is needed for invocation-based average sorting.

The UI maps the `avgCredits` column to this server-side ranking and keeps local sorting for columns that do not have a corresponding server ranking. No persisted data or migration changes are included; `sortBy` is optional and remains backwards compatible through its `credits` default.

## Tests

Added Vitest coverage in `front/lib/api/analytics/consumption/top.test.ts` for average-credit ranking on message-unit and invocation-unit dimensions. 

## Risk

The default credits ranking is preserved, but average-credit sorting for message-based dimensions can be substantially more expensive than the normal paginated path because it fetches every matching bucket and sorts the complete result set in memory. **High-cardinality dimensions may therefore have increased query time and memory usage.**

## Deploy Plan

Deploy `front` in stack order: [#30794](https://github.com/dust-tt/dust/pull/30794), then [#30795](https://github.com/dust-tt/dust/pull/30795), then this PR. No migration or additional rollout step is identified.

### DIFF
--- a/front/components/workspace/analytics/consumption/ConsumptionAttributionTable.tsx
+++ b/front/components/workspace/analytics/consumption/ConsumptionAttributionTable.tsx
@@ -78,7 +78,10 @@ const DEFAULT_ATTRIBUTION_SORTING: SortingState = [
   { id: "credits", desc: true },
 ];
 
-const ATTRIBUTION_SERVER_SORTABLE_COLUMN_ID = "credits";
+const ATTRIBUTION_SERVER_SORTABLE_COLUMN_IDS = new Set([
+  "credits",
+  "costShare",
+]);
 
 type AttributionTransitionDirection = -1 | 0 | 1;
 
@@ -291,6 +294,10 @@ function buildColumns({
     },
     {
       id: "costShare",
+      // Same denominator (totalCredits) for every row, so ranking by cost
+      // share is the same order as ranking by credits — reuse that accessor
+      // instead of introducing a separate ranking.
+      accessorFn: (row) => (totalCredits > 0 ? row.credits / totalCredits : 0),
       header: "Consumption share",
       meta: { sizeRatio: 20, headerAlign: "left" },
       cell: (info) => (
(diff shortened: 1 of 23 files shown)


avervaet:

- #30794
    - #30795
        - #30796 ◀
            - #30809
                - #30810

front/components/workspace/analytics/consumption/ConsumptionAttributionTable.tsx
```diff
@@ -450,16 +456,15 @@ function AttributionRows({
...
-      ? "asc"
-      : "desc";
+  // Only forward the sort to the server when the sorted column actually
+  // rides a server-side ranking; sorting by anything else keeps fetching
+  // pages in the default credits-desc order and reorders them locally
+  // instead.
+  const serverSortBy: ConsumptionTopSortBy | undefined = activeSort?.id
+    ? ATTRIBUTION_SERVER_SORT_BY_COLUMN_ID[activeSort.id]
+    : undefined;
+  const sortBy = serverSortBy ?? "credits";
```
aubin-tchoi, line 470: nit: could introduce a constant for the default sort order

aubin-tchoi (approved):
LGTM

front/lib/api/analytics/consumption/top.ts
```diff
@@ -319,7 +353,16 @@ export async function fetchConsumptionTopGroups(
     filter,
   });
 
-  const requestedBucketCount = offset + limit;
+  // Elasticsearch cannot order terms buckets by a ratio of two metrics
+  // (summed credits over distinct message count), so ranking "message"
+  // dimensions by avg credits requires fetching every bucket and sorting by
+  // average in memory, rather than relying on the requested page's size.
+  const sortInMemory =
```
aubin-tchoi, line 360: as per IRL let's look for a way to do this in the query as this will be too slow for a live interaction and have a toil on our ES

avervaet:
Closing: we won't move forward with this Elasticsearch-based sort/filter approach — the pressure it puts on ES for large workspaces makes it unsuitable for now.

PR #30797, opened 2026-08-19
chore: apply safe npm audit fixes

zmarouf:
## Description

`npm audit fix`


zmarouf:
image

Pinning Next _exactly_ for now because it breaks other dependency upgrades.

PR #30798, opened 2026-08-19
fix: restore cache mock in programmatic credit state machine test

adrsimon:
## Description

Shard 3 of front tests has failed on every main push since #30753: `client.ts` now calls `cacheWithRedis` at module scope, and this test's `vi.mock` factory only exported `invalidateCacheAfterCommit`, so switch it to the `importOriginal` partial mock used elsewhere.

## Tests

`NODE_ENV=test npm test -- lib/metronome` — 21 files, 357 tests pass.

## Risk

None, test-only.

## Deploy Plan

Deploy front.


tdraier:
LGTM

PR #30799, opened 2026-08-19
[front] Match automations download button size to filter button

avervaet:
## Description

BIG BOÏ BUTTON !

The CSV download button on the Automations page was rendered at `xs`, while the adjacent Filters button uses `sm`. This change adds an optional `size` override to `CsvDownloadButton` and sets it to `sm` in `AutomationsTriggersTable`, aligning the two controls. Callers that do not provide an override retain the existing behavior: labeled buttons use `sm`, and icon-only buttons use `xs`.

Automations download and filter buttons

## Tests

Manual verification:

- Confirm the download and filter buttons are the same size on the Automations page.
- Confirm other pages using `CsvDownloadButton`, including the Workspace usage/credits charts, are unchanged.

## Risk

Low 

## Deploy Plan

- Deploy front


ClementAupiais (approved):
LGTM Praise the big boï

PR #30801, opened 2026-08-19
Make run usage classification immutable

flvndvd:
## Description

We have been observing lots of "Error: Run usage billing classification is incomplete" error in out attribution items, after digging the recently introduced `usageType` field is sometimes never set for a given run usage. We currently only set it for "free" run usage, but one could argue that if we know that a run is free we should also be able to get the `usageType` directly and make it a required property to create a usage and immutable. That is what this PR does.

## Deploy Plan

No SQL migration required.


fabiencelier (approved):
LGTM
Thanks for cleaning this

PR #30708, opened 2026-08-18
[front] Flip space access to group_permissions with use_legacy_acls kill switch

tdraier:
## What

Flips **space** permission enforcement to the `group_permissions` table by default, mirroring the skill flip. Access reverts to the legacy inline-group path only via the synchronous `use_legacy_acls` kill switch.

## How

**`SpaceResource`** (mirrors `SkillResource`):
- `getAccessControlLists` now returns `grantedVerbs: auth.getGrantedVerbs("space", this.id)` instead of the inline `.groups` list.
- `canRead`/`canWrite`/`canAdministrate` → `hasSpacePermission(auth, verb)`, which serves the table decision unless `isLegacyAclsEnabled()` — then it falls back to the pre-migration inline-group ACL (`legacyAcls()` + `legacySpaceGroupGrants`).
- Removed `governanceAcls` and `shadowCompareSpacePermission`; kept the legacy path for the fallback.

**Batch space-read helpers migrated to governance.** Since `getAccessControlLists` no longer exposes `.groups`, the helpers that extracted it would go empty. Replaced `createSpaceIdToGroupsMap` + `createAccessControlListFromSpacesWithMap` with a single `canReadRequestedSpaces(auth, spaceById, requestedSpaceIds)` that checks `SpaceResource.canRead` per space (governance + honors the kill switch; a missing/deleted space is treated as not readable). Consumers updated:
- `filterAgentsByRequestedSpaces` (agent.ts)
- `conversation_resource.ts` (×2, reusing its existing `spaceId → space` map)
- `skill_resource.ts`

## Tests

- `space_resource.test.ts`: governance-enforcement block (member reads from the table; denied with no grants; `use_legacy_acls` fallback), mirroring the skill kill-switch test.
- `conversation_resource.test.ts`: helper tests rewritten for `canReadRequestedSpaces`.
- Added seed+refresh scaffolding where tests build associations directly (space access is now table-served).

Local `tsgo` + biome are clean; leaving the functional suite to CI.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #30805, opened 2026-08-19
[front] Refresh caller's auth grants after space permission mutations

tdraier:
## Description

`updatePermissions` and `createSpaceAndGroup` rewrite a space's `group_permissions` inside their transaction — and, for project creation, add the caller to the new editor group — but an `Authenticator` resolves its grant snapshot (`_groupModelIds` + `_permissions`) **once at construction** and never refreshes it for the life of the request.

Now that `group_permissions` is the served path (`#30708`, `use_legacy_acls` kill switch), a permission check made **later in the same request** reads the pre-mutation snapshot and returns the wrong answer. Concretely:

- an admin who just opened a restricted space is denied `read` on it (their snapshot predates the global-group reader grant), and
- a project creator can't `administrate` the pod they just created (their snapshot predates their membership in the new editor group).

This first surfaced as a `group_permissions_shadow_mismatch` on an open Pod: the served (legacy) decision granted read via the global group while the candidate (table) snapshot was stale. Two-request races of this shape are benign and self-heal on the next request; the single-request case fixed here does not.

The fix reuses the existing `Authenticator.refresh()` (which reloads live group membership and re-resolves grants) at each grant-mutation operation boundary, **after the transaction commits** — so the re-read sees committed rows past the `afterCommit` cache invalidation:

- `SpaceResource.updatePermissions` — after its `withTransaction` resolves.
- `createSpaceAndGroup` — after its `withTransaction` resolves.

Placed at the operation boundary rather than inside `GroupPermissionResource` (which writes grants row-by-row and is exercised by background/migration paths that never re-check), and post-commit rather than in-transaction (the `group_permissions` cache is only invalidated on `afterCommit`, so a pre-commit refresh would refill the snapshot from the old rows).

## Tests

Added a regression test for each path; both fail without the fix and pass with it:

- `space_resource.test.ts` → *"refreshes the caller's grant snapshot after opening a space"*: admin opens a restricted regular space; `getGrantedVerbs("space", id)` goes `[]` → `["read"]`. Without the fix: `expected [] to include 'read'`.
- `spaces.test.ts` → *"refreshes the live creator auth so it can administrate the new project in the same request"*: after `createSpaceAndGroup`, the live `userAuth` sees the new editor group and can `administrate` the pod with no manual refresh. Without the fix: `expected false to be true`. (Complements the existing test that reconstructs a stale auth and refreshes it manually.)

Verified: `tsgo --noEmit` clean; `group_permissions enforcement` describe 4 passed; `createSpaceAndGroup` describe 26 passed; both full files green.

## Risk

Low. `refresh()` is an existing, already-used method (the agent loop refreshes per step). The extra work is one membership query + one grants read per space create / permission update — cold-path operations, not hot paths. No behavior change to callers that don't reuse the auth after the mutation. Safe to roll back (revert the two `await auth.refresh()` calls).

## Deploy Plan

Standard `front` deploy. No migration, no config, no coordinated rollout.

### DIFF
--- a/front/lib/api/spaces.test.ts
+++ b/front/lib/api/spaces.test.ts
@@ -331,6 +331,49 @@ describe("createSpaceAndGroup", () => {
       createConnectorSpy.mockRestore();
     });
 
+    it("refreshes the live creator auth so it can administrate the new project in the same request", async () => {
+      const createConnectorSpy = vi
+        .spyOn(
+          await import("@app/lib/api/projects/connector"),
+          "createDataSourceAndConnectorForProject"
+        )
+        .mockResolvedValue(new Ok(undefined));
+
+      const userAuth = await Authenticator.fromUserIdAndWorkspaceId(
+        user1.sId,
+        workspace.sId
+      );
+
+      const result = await createSpaceAndGroup(userAuth, {
+        name: "Test Project Live Auth Refresh",
+        isRestricted: true,
+        spaceKind: "project",
+        managementMode: "manual",
+        memberIds: [],
+      });
+
+      expect(result.isOk()).toBe(true);
+      if (result.isOk()) {
+        const pod = result.value;
+        const { groupsToProcess } =
+          await pod.fetchManualGroupsMemberships(userAuth);
+        const editorGroup = groupsToProcess.find(
+          (group) => group.kind === "space_editors"
+        );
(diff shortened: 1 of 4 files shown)


tdraier:

- #30805 ◀

matteotrab (approved):

PR #30806, opened 2026-08-19
[admin governance] remove builder check from public api

ykmsd:
This PR is to replace builder check role in public api. 

- Agent configuration GET unpublished agents: admin only
- Agent configuration PATCH & DELETE: admin only
- Skill DELETE: admin only
- File POST & DELETE: manager only
- Data source tables DELETE: manager only

### DIFF
--- a/.grit/patterns/noDirectRoleCheck.grit
+++ b/.grit/patterns/noDirectRoleCheck.grit
@@ -6,7 +6,7 @@ language js
 // handlers.
 //
 // Role checks in route handlers should use the ensureIsAdmin(),
-// ensureIsBuilder(), ensureIsUser(), or ensureIsDustSuperUser() middleware
+// ensureIsUser(), or ensureIsDustSuperUser() middleware
 // from @front-api/middlewares/ensure_role instead.
 //
 // NOT flagged (allowed patterns):
@@ -31,7 +31,7 @@ direct_role_check() as $match where {
     not $match <: within `$_ && $_`,
     register_diagnostic(
         span = $match,
-        message = "Direct auth role checks are not allowed in route handlers. Use ensureIsAdmin(), ensureIsBuilder(), ensureIsUser(), or ensureIsDustSuperUser() middleware from @front-api/middlewares/ensure_role instead.",
+        message = "Direct auth role checks are not allowed in route handlers. Use ensureIsAdmin(), ensureIsUser(), or ensureIsDustSuperUser() middleware from @front-api/middlewares/ensure_role instead.",
         severity = "error"
     )
 }
--- a/front-api/routes/v1/w/[wId]/assistant/agent_configurations/[sId]/index.ts
+++ b/front-api/routes/v1/w/[wId]/assistant/agent_configurations/[sId]/index.ts
@@ -12,7 +12,7 @@ import type {
 } from "@dust-tt/client";
 import { PatchAgentConfigurationRequestSchema } from "@dust-tt/client";
 import { publicApiApp } from "@front-api/middlewares/ctx";
-import { ensureIsBuilder } from "@front-api/middlewares/ensure_role";
+import { ensureIsAdmin } from "@front-api/middlewares/ensure_role";
 import { apiError, type HandlerResult } from "@front-api/middlewares/utils";
 import { validate } from "@front-api/middlewares/validator";
 import { z } from "zod";
@@ -317,7 +317,7 @@ app.get(
 
 app.patch(
   "/",
-  ensureIsBuilder(),
+  ensureIsAdmin(),
   validate("param", AgentConfigurationParamSchema),
   validate("json", PatchAgentConfigurationRequestSchema),
   async (ctx): HandlerResult<GetOrPatchAgentConfigurationResponseType> => {
@@ -389,7 +389,7 @@ app.patch(
 
 app.delete(
   "/",
-  ensureIsBuilder(),
+  ensureIsAdmin(),
   validate("param", AgentConfigurationParamSchema),
   async (ctx): HandlerResult<DeleteAgentConfigurationResponseType> => {
     const auth = ctx.get("auth");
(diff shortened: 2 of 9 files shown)


front-api/routes/v1/w/[wId]/skills/[skillId].ts
```diff
@@ -81,16 +81,6 @@ app.delete(
       });
     }
 
-    if (!skill.canAdministrate(auth)) {
```
Comment (ykmsd) [inline front-api/routes/v1/w/[wId]/skills/[skillId].ts:84]: I guess we don't need it since we have ensureIsBuilder check?

front-api/routes/v1/w/[wId]/spaces/[spaceId]/data_sources/[dsId]/tables/[tId]/index.ts
```diff
@@ -209,7 +209,7 @@ app.get(
 
 app.delete(
   "/",
-  ensureIsBuilder(),
+  ensureIsManager(),
```
Comment (ykmsd) [inline front-api/routes/v1/w/[wId]/spaces/[spaceId]/data_sources/[dsId]/tables/[tId]/index.ts:212]: so we chatted about it but I'm not sure if non-api key path can reach these paths? 🤔

.grit/patterns/noDirectRoleCheck.grit
```diff
@@ -6,7 +6,7 @@ language js
 // handlers.
 //
 // Role checks in route handlers should use the ensureIsAdmin(),
-// ensureIsBuilder(), ensureIsUser(), or ensureIsDustSuperUser() middleware
+// ensureIsUser(), or ensureIsDustSuperUser() middleware
```
philipperolet, line 9: what about adding ensureIsManager there?

front-api/routes/v1/w/[wId]/assistant/agent_configurations/[sId]/index.ts
```diff
@@ -309,6 +311,16 @@ app.get(
       });
     }
 
+    if (!agentConfiguration.canRead && !auth.isAdmin()) {
```
Comment (philipperolet) [inline front-api/routes/v1/w/[wId]/assistant/agent_configurations/[sId]/index.ts:314]: not sure I understand, doesn't the grit rule above forbid the role check in ifs?

Comment (philipperolet) [reply front-api/routes/v1/w/[wId]/assistant/agent_configurations/[sId]/index.ts:314]: ooh it's only pure role checks without other conditions, so people use "ensureIsXXX"?

front-api/routes/v1/w/[wId]/skills/[skillId].ts
Comment (philipperolet) [reply front-api/routes/v1/w/[wId]/skills/[skillId].ts:84]: I see. But, editors can't archive the skill then, is that normal? I'd assume if you can create a skill you can delete it?

front-api/routes/v1/w/[wId]/spaces/[spaceId]/data_sources/[dsId]/tables/[tId]/index.ts
Comment (philipperolet) [reply front-api/routes/v1/w/[wId]/spaces/[spaceId]/data_sources/[dsId]/tables/[tId]/index.ts:212]: likely not but as safety we can leave isManager here (that will gate to admins)

front-api/routes/v1/w/[wId]/files/[fileId].ts
Comment (philipperolet) [inline front-api/routes/v1/w/[wId]/files/[fileId].ts:164]:
```suggestion
          "Only users that are `managers` for the current workspace can delete files.",
```

Comment (philipperolet) [inline front-api/routes/v1/w/[wId]/files/[fileId].ts:221]:
```suggestion
          "Only users that are `managers` for the current workspace can modify files.",
```

front-api/routes/v1/w/[wId]/assistant/agent_configurations/[sId]/index.ts
Comment (philipperolet) [inline front-api/routes/v1/w/[wId]/assistant/agent_configurations/[sId]/index.ts:228]: nit: 403 only admin should be here too

philipperolet (commented):
Looks good, left a few coms 🙏

front-api/routes/v1/w/[wId]/skills/[skillId].ts
Comment (ykmsd) [reply front-api/routes/v1/w/[wId]/skills/[skillId].ts:84]: but the problem is we cannot tell if you are an editor or not from the key right?

Comment (philipperolet) [reply front-api/routes/v1/w/[wId]/skills/[skillId].ts:84]: true; in that case I think we should move directly to the new perms, it's the clearest
auth.can("admin", skill)
The skill's ACL takes care of the permissionning logic (who can do what given their role, key or not, grant or not) via getAccessControlLists--the decisions on what to do given the way auth is is centralized here
(hope it's clear, otherwise we can chat IRL)

Comment (philipperolet) [reply front-api/routes/v1/w/[wId]/skills/[skillId].ts:84]: (notably because this route might be used via oauthed users)

front-api/routes/v1/w/[wId]/assistant/agent_configurations/[sId]/index.ts
Comment (ykmsd) [reply front-api/routes/v1/w/[wId]/assistant/agent_configurations/[sId]/index.ts]: yes the rule will catch only if you do pure permission check without any other conditions!

front-api/routes/v1/w/[wId]/skills/[skillId].ts
Comment (ykmsd) [reply front-api/routes/v1/w/[wId]/skills/[skillId].ts:84]: okay I changed it to `auth.can("admin", skill)` I wanted to use skill.canAdministrate but this is bypassing the api key role check so I use auth.can

philipperolet (approved):
Looks good 👍

front-api/routes/v1/w/[wId]/assistant/agent_configurations/[sId]/index.ts
```diff
@@ -324,7 +328,7 @@ app.get(
 
 app.patch(
   "/",
-  ensureIsBuilder(),
+  ensureIsAdmin(),
```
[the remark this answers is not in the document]
Comment (ykmsd) [reply front-api/routes/v1/w/[wId]/assistant/agent_configurations/[sId]/index.ts:341]: intentional

front-api/routes/v1/w/[wId]/assistant/agent_configurations/[sId]/index.ts
```diff
@@ -402,7 +406,7 @@ app.patch(
 
 app.delete(
   "/",
-  ensureIsBuilder(),
+  ensureIsAdmin(),
```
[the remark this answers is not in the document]
Comment (ykmsd) [reply front-api/routes/v1/w/[wId]/assistant/agent_configurations/[sId]/index.ts:419]: intentional

PR #30807, opened 2026-08-19
Cache feature flag resource lists

flvndvd:
## Description

Since we moved FF to the request cache in https://github.com/dust-tt/dust/pull/30652 we are seeing a significant number of queries on our DB, this PR adds a resource cache for both `FeatureFlagResource` and `GlobalFeatureFlagResource` without a TTL. It requires updating all the workspace cached FF.


id13:
On TTL: why not invalidate on the only two places we put it ? + make sure it's burstable from poke pluggin ?

id13 (approved):
LGTM

PopDaph:
Thank you 🙏🏻

PR #30809, opened 2026-08-19
Consumption attribution table sort vs prev growth server side

avervaet:
## Description

This is the fourth PR in the stack [#30794](https://github.com/dust-tt/dust/pull/30794) → [#30795](https://github.com/dust-tt/dust/pull/30795) → [#30796](https://github.com/dust-tt/dust/pull/30796) → **#30809**. It adds support for ranking the Consumption Attribution table by vs-prev growth.

Growth is computed as `(currentCredits - previousCredits) / previousCredits`. Because this ratio combines values from the current and previous periods and cannot be used as an Elasticsearch terms order path, the ranking fetches all current-period buckets in batches, retrieves previous-period credits for every bucket, sorts the complete set in memory, and only then applies pagination. The full previous-period lookup is reused to populate the page's vs-prev display values, avoiding a second lookup for the same data. Other sort modes continue to fetch previous-period credits only for the displayed page.

## Tests

Added Vitest coverage in `front/lib/api/analytics/consumption/top.test.ts` for growth ordering and for placing a group with no previous-period credits consistently. 

## Risk

The main risk is performance for high-cardinality dimensions: vs-prev sorting now enumerates every current bucket, performs a previous-period lookup for every key, and sorts the complete result set in memory instead of using the normal page-sized ranking path.

## Deploy Plan

Deploy `front` in stack order: [#30794](https://github.com/dust-tt/dust/pull/30794), [#30795](https://github.com/dust-tt/dust/pull/30795), [#30796](https://github.com/dust-tt/dust/pull/30796), then this PR. No migration or additional rollout step is identified.

### DIFF
--- a/front/components/workspace/analytics/consumption/ConsumptionAttributionTable.tsx
+++ b/front/components/workspace/analytics/consumption/ConsumptionAttributionTable.tsx
@@ -87,6 +87,7 @@ const ATTRIBUTION_SERVER_SORT_BY_COLUMN_ID: Partial<
   credits: "credits",
   costShare: "credits",
   avgCredits: "avgCredits",
+  vsPrev: "vsPrev",
 };
 
 type AttributionTransitionDirection = -1 | 0 | 1;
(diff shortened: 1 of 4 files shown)


avervaet:

- #30794
    - #30795
        - #30796
            - #30809 ◀
                - #30810

avervaet:
Closing: we won't move forward with this Elasticsearch-based sort/filter approach — the pressure it puts on ES for large workspaces makes it unsuitable for now.

PR #30811, opened 2026-08-19
[Agent Loop] Don't let deploy interruptions exhaust the model activity's retry budget

PopDaph:
## Description

Follow up of #30742.

Goal: deploys must stop putting agent workflows in FAILED status. On 2026-08-19, 4 workflows failed with `ModelInterruption` after exhausting the model activity's 5 attempts ([Datadog](https://app.datadoghq.eu/logs?query=service%3Afront-agent-loop-worker%20%22Model%20activity%20interrupted%22%20%40attempt%3A5&from_ts=1787270400000&to_ts=1787356800000)): model steps are multi-minute LLM streams, so a single deploy train (3 deploys in 19 minutes) kills 5 attempts in a row.

Fix: bump the activity's `maximumAttempts` from 5 to `RETRY_ON_INTERRUPT_MAX_ATTEMPTS` (15), the same trade tools made, and drop the `backoffCoefficient: 1` pin so retries spread out (SDK-default exponential backoff) instead of re-entering the same drain window. Surviving a deploy-train now requires ~1h of continuous churn on one step.

Why this does not mean "retry everything 15 times": model errors (rate limits, provider 5xx, stream timeouts) never reach the policy cap. `run_model` surfaces them to the user at attempt >= `RUN_MODEL_MAX_RETRIES` (5) and returns instead of throwing, unchanged behavior. That gate is extracted as `shouldSurfaceModelError` and tested, since the bump relies on it. Attempts 6-15 are only reachable by interruptions, timeouts and internal errors.

Workflows still go FAILED when all 15 attempts exhaust (persistent internal errors, non-LLM hangs): those are bugs or infra incidents and should stay visible in Temporal, unlike deploy noise.

No workflow command change (retry policy is an activity option): replay-safe, no patch marker.

## Tests

Unit tests for `shouldSurfaceModelError`. `tsgo --noEmit` clean.

## Risk

Worst case, a step failing for non-transient, non-model reasons burns 15 attempts instead of 5 before erroring. Safe to rollback.

## Deploy Plan

- [ ] Deploy front.
- [ ] Watch `service:front-agent-loop-worker "Model activity interrupted" @attempt:>=5`: no longer implies failure, only retries past the old cap.
- [ ] Watch `@workflowType:agentLoopWorkflow status:error "Model activity interrupted"`: terminal failures should drop to zero across deploy trains.

### DIFF
--- a/front/temporal/agent_loop/workflows.ts
+++ b/front/temporal/agent_loop/workflows.ts
@@ -15,7 +15,6 @@ import type * as runModelAndCreateWrapperActivities from "@app/temporal/agent_lo
 import type * as runToolActivities from "@app/temporal/agent_loop/activities/run_tool";
 import {
   MODEL_ACTIVITY_HEARTBEAT_TIMEOUT_MS,
-  RUN_MODEL_MAX_RETRIES,
   TOOL_ACTIVITY_HEARTBEAT_TIMEOUT_MS,
 } from "@app/temporal/agent_loop/config";
 import type { ToolExecutionResult } from "@app/temporal/agent_loop/lib/deferred_events";
@@ -80,8 +79,10 @@ const { runModelAndCreateActionsActivity } = proxyActivities<
   heartbeatTimeout: MODEL_ACTIVITY_HEARTBEAT_TIMEOUT_MS,
   cancellationType: ActivityCancellationType.WAIT_CANCELLATION_COMPLETED,
   retry: {
-    maximumAttempts: RUN_MODEL_MAX_RETRIES,
-    backoffCoefficient: 1,
+    // Attempts past RUN_MODEL_MAX_RETRIES only serve non-model failures (worker-shutdown
+    // interruptions, timeouts, internal errors): real model errors self-limit in run_model
+    // (see shouldSurfaceModelError).
+    maximumAttempts: RETRY_ON_INTERRUPT_MAX_ATTEMPTS,
   },
 });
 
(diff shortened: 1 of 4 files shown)


front/temporal/agent_loop/workflows.ts
```diff
@@ -78,21 +77,10 @@ const { runModelAndCreateActionsActivity } = proxyActivities<
...
-  runModelAndCreateActionsActivity:
-    runModelAndCreateActionsActivityWithExplicitCancellation,
-} = proxyActivities<typeof runModelAndCreateWrapperActivities>({
-  startToCloseTimeout: "10 minutes",
-  heartbeatTimeout: MODEL_ACTIVITY_HEARTBEAT_TIMEOUT_MS,
-  cancellationType: ActivityCancellationType.WAIT_CANCELLATION_COMPLETED,
-  retry: {
-    maximumAttempts: RUN_MODEL_MAX_RETRIES,
-    backoffCoefficient: 1,
+    // Attempts past RUN_MODEL_MAX_RETRIES only serve non-model failures (worker-shutdown
```
flvndvd: I changed the settings here earlier this morning, you might need to rebase.

PopDaph: Thanks 🙏🏻

fontanierh (approved):
LGTM, but not fully clear to me how one deploy train is causing multiple failures... How can it be more than 1 ?

fontanierh:
There may be an underlying issue worth fixing

PR #30814, opened 2026-08-19
[front] enh: tighten blocked action card layouts

aubin-tchoi:
## Description

Tool approval and account connection cards had an inconsistent layout, handled by paddings that were arbitrarily distributed.

This PR aligns both cards to handle paddings using gaps rather than 
No visual change apart from 4px less paddings horizontally, a gap-3 between the buttons instead of 2 and a gap-2 between the avatar and the title instead of 3.

Next step is to move the components to sparkle.

## Tests

- Tested locally.

## Risk

- Low.

## Deploy Plan

- Deploy front.


ClementAupiais (approved):
lgtm

PR #30815, opened 2026-08-19
fix(spaces): let the public spaces list return pods on request

adrsimon:
## Description

Fixes dust-tt/tasks#10056.

#29711 narrowed `GET /api/v1/w/{wId}/spaces` to `system`/`global`/`regular`, which was intentional: pods are numerous and nobody wanted them in the default listing. But `dust-cli` resolves `--projectId` / `--projectName` against that listing, so targeting a pod has been broken since July 29 while `POST /assistant/conversations` kept accepting a pod `spaceId`.

The default response is unchanged. `kinds` can now select `project` on top of the three existing kinds, and the CLI passes all four when resolving a project flag.

Note that a pod only shows up for a caller who is a member of it: the workspace global group is only ever a reader on a pod, so an API key that holds nothing but the global group still sees no pods. That matches the endpoint's existing "spaces you belong to" semantics.

## Tests

Added a case to `front-api/routes/v1/w/[wId]/spaces/index.test.ts` covering both halves: a pod is absent from the default response and present under `kinds=project`. Existing cases in that file still pass.

## Risk

Low. Purely additive on the API side, and callers that do not pass `kinds` see exactly what they saw before.

## Deploy Plan

Deploy front-api, then release the CLI.

### DIFF
--- a/cli/dust-cli/src/ui/commands/chat/nonInteractive.ts
+++ b/cli/dust-cli/src/ui/commands/chat/nonInteractive.ts
@@ -47,7 +47,10 @@ export async function resolveSpaceId(
     return undefined;
   }
 
-  const spacesRes = await dustClient.getSpaces();
+  // Projects are not listed by default by the public API, they have to be asked for explicitly.
+  const spacesRes = await dustClient.getSpaces({
+    kinds: ["system", "global", "regular", "project"],
+  });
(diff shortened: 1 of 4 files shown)


id13 (approved):
nice 👍

matteotrab (approved):

PR #30816, opened 2026-08-19
fix: render a toggle in case of unknown status in automations

adrsimon:
## Description

In the automations administration page, an unknown status renders an empty cell in the table. We instead render an unselected toggle so that it's actionnable.

This happened during the migration of disabled_by_admin to disabled_by_manager, where we didn't backfill the previous status. The lines holding an old status rendered an empty cell, unactionable.

## Tests

Tested by hand

## Risk

Low

## Deploy Plan

Deploy SPA


avervaet (approved):
LGTM

PR #30817, opened 2026-08-19
feat(cli): scope the file system tools to the working directory

### LINKED ISSUE #26743: Dust CLI appears to run shell commands outside a sandbox
# Dust CLI: filesystem access is not sandboxed to the current working directory

## Summary

When using Dust CLI, the agent appears able to inspect the local filesystem outside the current working directory.

In my test, I launched Dust CLI from a local project folder and asked the agent to run `ls` in the parent folder. The command succeeded and Dust returned the result.

This suggests that Dust CLI does not enforce a filesystem sandbox rooted at the current working directory, or at least that the current sandbox boundary is broader than expected.

I would expect Dust CLI’s behavior to be closer to tools like Codex and Claude Code, where filesystem access is scoped, sandboxed, or requires explicit user approval when going outside the workspace.

## Environment

- Dust CLI version: `v0.4.5`
- Agent: `dust`
- Working directory shown by CLI: `~/Documents/Dev/test`

## Reproduction steps

1. Open Dust CLI from a local folder:

   ```bash
   cd ~/Documents/Dev/test
   dust
   ```

2. Ask the agent:

   ```text
   Can you run ls in the folder above?
   ```

3. Observe that the command is run in the parent folder.

In my case, Dust accessed:

```text
/home/jphetphoumy/Documents/Dev
```

instead of being scoped to:

```text
/home/jphetphoumy/Documents/Dev/test
```

## Actual behavior

Dust was able to run `ls` outside the current working directory and return the contents of the parent directory.

This means the agent was not scoped to the directory where the CLI session was started.

## Expected behavior

Dust CLI should enforce a clear filesystem sandbox boundary by default.

When launched from:

```text
~/Documents/Dev/test
```

the agent should not be able to read or list:

```text
~/Documents/Dev
```

unless the user explicitly grants access.

A safer default would be:

- allow reads/writes only inside the current workspace
- require explicit approval before accessing parent directories or paths outside the workspace
- clearly display the filesystem access scope when the CLI starts
- provide configuration flags for broader access when desired

For example:

```bash
dust --sandbox .
dust --allow-path ~/Documents/Dev
dust --dangerously-disable-sandbox
```

## Comparison with Codex and Claude Code

Codex documents a sandboxing model where the agent can operate in the workspace, while access outside the workspace and other higher-risk actions can require approval depending on configuration.

Claude Code also documents permission and sandbox controls for filesystem access, including sandboxing and allow/deny rules.

I would expect Dust CLI to provide a similar safety model: the agent can work freely inside the project, but crossing the workspace boundary should require explicit user consent.

References:

- OpenAI Codex sandboxing docs: https://developers.openai.com/codex/concepts/sandboxing
- Claude Code permissions docs: https://code.claude.com/docs/en/permissions
- Claude Code sandboxing docs: https://code.claude.com/docs/en/sandboxing
- Anthropic engineering post on Claude Code sandboxing: https://www.anthropic.com/engineering/claude-code-sandboxing

## Security implications

Users may reasonably assume that starting an agentic CLI inside a project folder limits the agent to that project.

Without a filesystem sandbox or approval prompt, the agent can unintentionally expose information outside the intended workspace, such as:

- sibling project names
- private repository names
- local folder structure
- file metadata
- unrelated development folders
- potentially sensitive files in parent directories

This is especially important because agentic CLIs can be influenced by prompt injection or malicious project files. A workspace boundary helps reduce the impact of an agent being instructed to inspect or exfiltrate files outside the intended project

## Additional context

The main issue is not only that `ls ..` succeeded, but that Dust was able to inspect and summarize filesystem contents outside the current project directory without an explicit permission prompt.

As a frequent user of Codex and Claude Code for engineering work, I rely on predictable workspace boundaries when using agentic CLI tools. In its current form, Dust CLI is difficult for me to use for engineering tasks because I cannot confidently limit the agent’s filesystem access to the project I am working on.

For safety, I think Dust should make the workspace boundary explicit and require approval before crossing it.

adrsimon:
## Description

Fixes #26743.

Starting the CLI inside a project looks like a boundary, and users coming from Codex or Claude Code assume it is one. It wasn't: `read_file`, `edit_file`, `search_files`, `search_content` and `run_command` all took any absolute path, so `ls ..` happily listed the parent directory.

`src/utils/sandbox.ts` now resolves every path a tool receives and checks it against the directory the CLI was started in. Symlinks are resolved on the longest existing prefix, so a link planted in the workspace can't be used to step out. Two new flags: `--allow-path <path>` (repeatable) to widen the scope, `--dangerously-disable-sandbox` to lift it. The scope is printed in the chat header.

One limitation, spelled out in the README: `run_command` checks its arguments against the boundary but does not confine the process it spawns. `ls ..` is refused, `sh -c 'ls ..'` is not. Confining the child means seatbelt/landlock, which is a much bigger change than this one; the argument check covers the accidental and casually-injected cases, which is what the report is about.

## Tests

Manually drove the tools from the built source: `ls ..` and `read_file /etc/hosts` are refused with the flag hint, `ls src` still runs.

## Risk

Behaviour change for anyone whose agent reaches outside the start directory today, which will now fail with an error naming the flag to fix it. That is the point of the change, but it is a break. The escape hatches are on the same error message.

## Deploy Plan

Release the CLI.


fabiencelier (approved):
LGTM ✅

PR #30819, opened 2026-08-19
[front] Gate the Pod App UI and APIs behind a pod_applications flag

davidebbo:
## Description

Pod Apps were reachable by anyone with `sandbox_functions`, which is the flag for the broader Pod Functions surface. This adds a dedicated `pod_applications` flag so the Apps surface can be rolled out independently, and applies the same gate on the client and the server.

## Risk

Low.

## Deploy Plan

Enable `pod_applications` on the Dust workspaces that currently use Pod Apps, then deploy front + front-api.


fontanierh (approved):

fontanierh (approved):
lgrm

PR #30821, opened 2026-08-19
[front] enh: reuse Composer UI for Manage Agents model filter

aubin-tchoi:
## Description

Closes https://github.com/dust-tt/tasks/issues/10127

The Manage Agents model filter uses a flat list today and does not match the model picker in the input bar.

This PR keeps the filter multi-select and limited to models used by agents, while matching the Composer tier and provider grouping, search, and row styling. The filter reuses the existing tier and supported-model data without changing the Composer picker or shared dropdown primitives.

Ended up being quite hard to reuse the component as is. It's a 20-props component that strongly ties the visual rendering from the state management, and both are quite different in our case (multi-select, reasoning efforts). Will refactor the component first before merging both (for follow up PRs, not a priority for shipping this part).

Note: this is a dropdown in a dropdown in a dropdown. This PR is a first step that optimizes for consistency, but we should revisit the design for both.

## Tests

- Tested locally.

## Risk

- Low.

## Deploy Plan

- Deploy front.

### DIFF
--- a/front/components/model_picker/ModelPickerMoreModels.tsx
+++ b/front/components/model_picker/ModelPickerMoreModels.tsx
@@ -133,7 +133,7 @@ export function ModelPickerMoreModels({
   };
 
   const searchbar = (
-    <div className="sticky top-0 z-10 bg-overlay-background pt-2">
+    <div className="sticky top-0 z-10 bg-overlay-background">
       <DropdownMenuSearchbar
         autoFocus={!isWidthConstrained}
         name="search-models"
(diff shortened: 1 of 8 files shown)


front/components/model_picker/ModelPickerMoreModels.tsx
```diff
@@ -133,7 +133,7 @@ export function ModelPickerMoreModels({
   };
 
   const searchbar = (
-    <div className="sticky top-0 z-10 bg-overlay-background pt-2">
```
aubin-tchoi, line 136: drive-by change

ClementAupiais:
LGTM

PR #30818, opened 2026-08-19
[front] Support both regular_auto and space_editors for space editor groups

tdraier:
## Description

First step of removing the `space_editors` group kind in favor of `regular_auto`.

- Space **editor groups are now created as `regular_auto`** (both creation sites).
- **Editor/member identity is resolved from `group_permissions`**, not the group kind or `group_vaults`: the editor group is the one holding the space `admin` grant, the member group the one holding the `member` grant. Resolved via `GroupPermissionResource.findRegularAutoGroupForGrant` (`SpaceResource.fetchManualEditorGroup` / `fetchManualMemberGroup`, memoized per instance), and `fetchManualGroupsMemberships` now returns `editorGroupModelId` so `isEditor` no longer depends on kind.
- `findRegularAutoGroupForGrant` gains an optional `groupKinds` arg; during the migration editor lookups pass `["regular_auto", "space_editors"]` (`MIGRATING_SPACE_EDITOR_GROUP_KINDS`) so not-yet-backfilled editor groups are still found.
- Editor-group membership edits are gated on `space.canAdministrate`, never the editor group's own ACL/`canWrite`.

Bottom of a 3-PR stack: **this → #30822 (backfill) → #30823 (remove kind)**.

## Tests

Existing functional tests updated (editor identity via `editorGroupModelId` / `fetchManual*Group`) and passing:
- front: `space_resource` (69), `spaces` (39), `group_space_resource` (16), `pod_manager` helpers (7) + tools (4), `auth.fromjson`, `workspace_user`, `permissions` (27).
- front-api: `spaces/[spaceId]/leave` (5), `spaces/[spaceId]` (2), `members` (2).
- `tsgo --noEmit` clean on front and front-api.

## Risk

Low / behavior-preserving. Both kinds are accepted throughout, so old (`space_editors`) and new (`regular_auto`) editor groups work simultaneously. Editor identity now comes from the space `admin` grant in `group_permissions`, which is already populated for existing spaces (backfilled by `20260728_backfill_space_group_permissions`). No schema or data changes — safe to roll back by reverting the code.

## Deploy Plan

Standard deploy, no migration. Once deployed, proceed to the backfill (#30822).

PR #30822, opened 2026-08-19
[front] Backfill space_editors group kind to regular_auto

tdraier:
## Description

Post-deploy backfill that re-labels existing space editor groups from the deprecated `space_editors` kind to `regular_auto`, in both `groups.kind` and the denormalized `group_vaults."groupKind"`.

Script: `front/migrations/20260819_backfill_space_editors_to_regular_auto.ts` — keyset-paginated, batched (configurable `batchSize` / `batchDelayMs`), dry-run by default. Editor identity is derived from `group_permissions` (the space `admin` grant), so this only re-labels rows; behavior is unchanged.

Middle of a 3-PR stack: #30818 (support both kinds) → **this** → #30823 (remove kind).

## Tests

Runtime script (not endpoint logic), so no automated test. Validated via the built-in dry run, which logs the exact number of `groups` and `group_vaults` rows that would change without writing.

## Risk

Low. Idempotent — only rows still set to `space_editors` are touched, so it is safe to re-run. Raw SQL leaves `updatedAt` untouched. The "support both kinds" change (#30818) must be deployed first so the running code treats `regular_auto` editor groups correctly during and after the flip. Reversible by running the equivalent reverse update if ever needed.

## Deploy Plan

1. Merge and deploy #30818 first.
2. Run the script dry run (default) to check the counts.
3. Run with `--execute` in **prod** and **EU**; confirm a follow-up dry run reports zero remaining `space_editors` rows.
4. Then proceed to #30823.

### DIFF
--- a/front/migrations/20260819_backfill_space_editors_to_regular_auto.ts
+++ b/front/migrations/20260819_backfill_space_editors_to_regular_auto.ts
@@ -1,0 +1,176 @@
+import { QueryTypes } from "sequelize";
+
+import { frontSequelize } from "@app/lib/resources/storage";
+import type { Logger } from "@app/logger/logger";
+import { makeScript } from "@app/scripts/helpers";
+
+const DEFAULT_BATCH_SIZE = 1_000;
+const DEFAULT_BATCH_DELAY_MS = 100;
+
+// Flip space editor groups from the deprecated "space_editors" kind to "regular_auto". The code
+// already treats both kinds as editor groups (see MIGRATING_SPACE_EDITOR_GROUP_KINDS), so this
+// backfill only re-labels rows; editor identity is derived from group_permissions, not the kind.
+// Two tables carry the kind: `groups.kind` and the denormalized `group_vaults."groupKind"`. Both
+// are updated. Idempotent — only rows still set to "space_editors" are touched, so it is safe to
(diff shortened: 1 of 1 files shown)


tdraier:

- #30818
    - #30822 ◀
        - #30823
            - #30889
                - #30890

front/migrations/20260819_backfill_space_editors_to_regular_auto.ts
```diff
@@ -0,0 +1,176 @@
...
+      logger
+    );
+    await backfillTable(
+      {
+        table: "group_vaults",
+        selectSql: SELECT_GROUP_VAULTS_SQL,
+        updateSql: UPDATE_GROUP_VAULTS_SQL,
+        batchSize,
+        batchDelayMs,
+      },
```
fabiencelier, line 172: any risk of desync by running these one after the other ?
I have no idea, just aski g to make sure you have checked we don't need the same type in both at the same time

fabiencelier (approved):
LGTM with one comment to check

front/migrations/20260819_backfill_space_editors_to_regular_auto.ts
tdraier: yes they may be unsynced for some time, but we handle both values (regular_auto / space_editors ) the same way - so it should not have any impact.

PR #30826, opened 2026-08-19
Link Files to stable filesystem nodes

flvndvd:
## Description

This PR adds the data model needed to associate a `FileResource` with its live source in the database-backed filesystem.

Today, Frames are linked to a mount path. That path changes when the source is moved or renamed. A filesystem node keeps the same ID across these operations, so `files.fileSystemNodeId` gives us a stable link between the shareable File and its source.

The new field is nullable, allowing the current GCS-based flow to continue unchanged. It has an index for node-based lookups and a foreign key using `ON DELETE RESTRICT`, so a filesystem node cannot disappear while a File still points to it.

This PR only introduces the schema and Sequelize association. Publishing, file-explorer enrichment, moves, and deletion behavior will follow in separate PRs.

## Deploy Plan

 - [ ] Run SQL migration
 - [ ] Deploy front

### DIFF
--- a/front/migrations/pre-deploy/20260819134521_add_file_system_node_id_to_files.sql
+++ b/front/migrations/pre-deploy/20260819134521_add_file_system_node_id_to_files.sql
@@ -1,0 +1,35 @@
+/*
+Statement 0
+*/
+SET SESSION statement_timeout = 3000;
+SET SESSION lock_timeout = 3000;
+ALTER TABLE "public"."files" ADD COLUMN "fileSystemNodeId" bigint;
+
+/*
+Statement 1
+CREATE INDEX CONCURRENTLY cannot run inside a transaction, and the runner
+applies this file through psql so the session settings above do not carry an
+open transaction.
+*/
+SET SESSION statement_timeout = 1200000;
+SET SESSION lock_timeout = 3000;
+CREATE INDEX CONCURRENTLY files_file_system_node_id ON public.files USING btree ("fileSystemNodeId") WHERE ("fileSystemNodeId" IS NOT NULL);
+
+/*
+Statement 2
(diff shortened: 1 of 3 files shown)


fontanierh (approved):

davidebbo (approved):
LGTM

front/migrations/pre-deploy/20260819134521_add_file_system_node_id_to_files.sql
```diff
@@ -0,0 +1,35 @@
...
+Statement 2
+RESTRICT: a node cannot be removed while a file still points at it, so the
+ordering is enforced here rather than left to callers.
+
+NOT VALID then VALIDATE: the constraint applies to new rows immediately while
+the validation scan takes only a light lock on this large table.
+*/
+SET SESSION statement_timeout = 3000;
+SET SESSION lock_timeout = 3000;
+ALTER TABLE "public"."files" ADD CONSTRAINT "files_fileSystemNodeId_fkey" FOREIGN KEY ("fileSystemNodeId") REFERENCES file_system_nodes(id) ON UPDATE CASCADE ON DELETE RESTRICT NOT VALID;
```
davidebbo, line 28: Does that mean a file system `rm` will fail here? I'd think if we delete the tsx, we want the frame to be deleted.

fontanierh: Flavien famously hates cascade deletes. This SQL line is very on-brand :p

flvndvd: Yes SQL would not allow deleting the node without deleting the associated files, expectation is to delete on the application side on `rm`.

PR #30827, opened 2026-08-19
Bind Frames to database filesystem nodes

flvndvd:
## Description

This PR makes a Frame follow its source file by filesystem node ID instead of relying on a path that may change.

When the database-backed filesystem is enabled, creating or publishing a Frame records the node that holds its entry file. Publishing and live editing can then find the file’s current path after it has been moved or renamed.

The file explorer also joins filesystem entries with Files by node ID. This restores the Frame MIME type and File ID needed to open the Frame drawer. The lookup is batched for the full listing. Only accepted during the PoC obviously 🙃.

The existing GCS flow remains path-based and unchanged. The storage- mode choice is kept at the top of the create and publish flows rather than spread through the lower-level Frame code.

This PR covers creation, publication, editing, and file-explorer reads. Removing a bound Frame and handling rename-over are left for the following PR.

### DIFF
--- a/front/lib/api/file_system/backends/database_file_system_backend.ts
+++ b/front/lib/api/file_system/backends/database_file_system_backend.ts
@@ -211,6 +211,7 @@ export class DatabaseFileSystemBackend implements FileSystemBackend {
         node.contentType ?? DEFAULT_CONTENT_TYPE
       ),
       lastModifiedMs: node.updatedAt.getTime(),
+      fileSystemNodeId: node.id,
       fileId: null,
       thumbnailUrl: null,
     };
@@ -336,6 +337,55 @@ export class DatabaseFileSystemBackend implements FileSystemBackend {
   ): Promise<Result<boolean, DustFileSystemError>> {
     try {
       return new Ok((await this.resolve(scopedPath)) !== null);
+    } catch (error) {
+      return new Err(this.error(error));
+    }
+  }
+
+  async nodeIdForPath(
+    scopedPath: string
+  ): Promise<Result<number | null, DustFileSystemError>> {
+    try {
+      const resolved = await this.resolve(scopedPath);
+      if (!resolved) {
+        return new Err(
+          new DustFileSystemError(
+            "not_found",
+            `No file or directory at: ${scopedPath}`
+          )
+        );
+      }
+      return new Ok(resolved.node.id);
+    } catch (error) {
+      return new Err(this.error(error));
+    }
+  }
+
+  async pathForNodeId(
+    nodeId: number
+  ): Promise<Result<string | null, DustFileSystemError>> {
+    try {
+      const node = await FileSystemNodeResource.fetchById(
+        this.auth,
+        this.scope,
+        nodeId
+      );
+      if (!node) {
+        return new Err(
+          new DustFileSystemError("not_found", "The node no longer exists.")
+        );
+      }
+      const path = (await node.pathSegmentsFromRoot()).join("/");
+      // The node may have moved to a root this file system does not mount.
+      if (!this.parse(path)) {
+        return new Err(
+          new DustFileSystemError(
+            "not_found",
+            "The node lives outside this file system."
+          )
+        );
+      }
+      return new Ok(path);
     } catch (error) {
       return new Err(this.error(error));
     }
--- a/front/lib/api/files/client_executable.ts
+++ b/front/lib/api/files/client_executable.ts
@@ -1,3 +1,4 @@
+import { DustFileSystem } from "@app/lib/api/file_system";
 import type { ValidationWarning } from "@app/lib/api/files/content_validation";
 import {
   validateTailwindCode,
@@ -18,6 +19,7 @@ import { executeWithLock } from "@app/lib/lock";
 import { FileResource } from "@app/lib/resources/file_resource";
 import { getResourceIdFromSId } from "@app/lib/resources/string_ids";
 import logger from "@app/logger/logger";
+import { conversationScopedPath } from "@app/types/file_system";
 import type { InteractiveContentFileContentType } from "@app/types/files";
 import {
   INTERACTIVE_CONTENT_FILE_FORMATS,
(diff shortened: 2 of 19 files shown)


front/lib/api/files/client_executable.ts
```diff
@@ -132,6 +134,31 @@ export async function createClientExecutableFile(
...
+    const scopedPath = conversationScopedPath({
+      conversationId,
+      rel: fileResource.fileName,
+    });
+    const fsResult = await DustFileSystem.fromScopedPath(auth, scopedPath);
+    if (fsResult.isErr()) {
+      return new Err({ message: fsResult.error.message, tracked: true });
+    }
+
+    if (fsResult.value.isDatabaseBacked()) {
```
flvndvd, line 148: This acts as a FF in some ways since database backed FS is gated behind a FF.

front/lib/resources/file_system_node_resource.ts
```diff
@@ -496,6 +496,42 @@ export class FileSystemNodeResource extends BaseResource<FileSystemNodeModel> {
     });
   }
 
+  // The node names from the root row down to this node, root name included.
+  // Joined with "/" they form the node's scoped path, since a root row carries
+  // its scoped prefix as its name. Follows parentId with recursive SQL for the
+  // same reason as isAncestorOf below: each row only stores its parent.
+  async pathSegmentsFromRoot(): Promise<string[]> {
```
flvndvd, line 503: Only used for the PoC to deal with the full GCS path constraint.

front/lib/api/file_system/backends/database_file_system_backend.ts
```diff
@@ -341,6 +342,55 @@ export class DatabaseFileSystemBackend implements FileSystemBackend {
     }
   }
 
+  async nodeIdForPath(
+    scopedPath: string
```
davidebbo, line 346: Is this a full path like `w/BHWrN3HaUa/conversations/yxhhYORjsS/files/Counter.tsx`? Always helps to see an example of what things look like in a comment, as a path can mean so many things.

front/lib/api/file_system/backends/database_file_system_backend.ts
```diff
@@ -341,6 +342,55 @@ export class DatabaseFileSystemBackend implements FileSystemBackend {
...
+        this.auth,
+        this.scope,
+        nodeId
+      );
+      if (!node) {
+        return new Err(
+          new DustFileSystemError("not_found", "The node no longer exists.")
+        );
+      }
+      const path = (await node.pathSegmentsFromRoot()).join("/");
```
davidebbo, line 378: `pathSegmentsFromRoot` looks like expensive query. Hopefully not a bottleneck.

davidebbo, line 378: nit: joins the segment, only to have parse split them again, so a bit inefficient.

front/lib/api/file_system/backends/database_file_system_backend.ts
```diff
@@ -341,6 +342,55 @@ export class DatabaseFileSystemBackend implements FileSystemBackend {
...
+      // The node may have moved to a root this file system does not mount.
```
davidebbo, line 379: Is that really possible in practice? I thought a move to a different file system was treated as a copy/delete.

front/lib/api/viz/edit_frame_text.ts
```diff
@@ -100,6 +101,34 @@ export async function editFrameTextAtSource(
...
+      if (currentEntryRes.value === null) {
+        return new Err(
+          new EditFrameTextError(
+            "source_not_found",
+            "The Frame source no longer exists."
+          )
+        );
+      }
+
+      rootScopedPath = path.posix.dirname(currentEntryRes.value);
```
davidebbo, line 126: Maybe here too, seeing examples help understanding, as we have 3 different king of path variables and it's not obvious what they look like.

front/lib/resources/file_resource.ts
```diff
@@ -1154,6 +1172,24 @@ export class FileResource extends BaseResource<FileModel> {
...
+    }: {
+      metadata: FileUseCaseMetadata;
+      fileSystemNodeId: number;
+    }
+  ) {
+    const result = await this.update({
+      useCaseMetadata: metadata,
+      fileSystemNodeId,
+    });
+    await this.resolveAndSetMountFilePath(auth);
```
davidebbo, line 1189: Maybe comment here that this path can become stale, and might go away later

front/lib/api/files/client_executable.ts
```diff
@@ -132,6 +134,31 @@ export async function createClientExecutableFile(
...
+      if (written.isErr()) {
+        return new Err({ message: written.error.message, tracked: true });
+      }
+      if (written.value.nodeId === null) {
+        return new Err({
+          message: "The database filesystem did not return a node id.",
+          tracked: true,
+        });
+      }
+      await fileResource.bindToFileSystemNode(written.value.nodeId);
```
davidebbo, line 159: Could you in theory end up with 2 file resources pointing to the same node here? Not sure, I might be confused.

front/lib/api/file_system/backends/database_file_system_backend.ts
```diff
@@ -211,6 +211,7 @@ export class DatabaseFileSystemBackend implements FileSystemBackend {
         node.contentType ?? DEFAULT_CONTENT_TYPE
       ),
       lastModifiedMs: node.updatedAt.getTime(),
+      fileSystemNodeId: node.id,
```
davidebbo, line 214: Can this end up getting returned to client, violating the "no model id on client" rule?

front/lib/api/file_system/backends/database_file_system_backend.ts
davidebbo, line 326: The fact that this returns the Node's content type is a problem, because for a frame we end up with `text/typescript`. This breaks logic like `validatePodFrameTabs`, which has:

```
    const statResult = await fs.stat(normalizedPath);
    if (!isInteractiveContentType(statResult.value.contentType)) { ... }
```

front/lib/api/viz/edit_frame_text.ts
```diff
@@ -68,8 +69,8 @@ export async function editFrameTextAtSource(
     editedByAgentConfigurationId?: string;
   }
 ): Promise<Result<{ warnings: ValidationWarning[] }, EditFrameTextError>> {
-  const root = file.useCaseMetadata?.frameBundleRootPath;
-  if (!file.isInteractiveContent || !root) {
+  const storedRoot = file.useCaseMetadata?.frameBundleRootPath;
```
davidebbo, line 72: Possible issue here is that the metadata doesn't get updated when you mv a file. You're stuck with the original `frameEntryRelPath` and `frameBundleRootPath`, which no longer matches the file. So anything that uses this is busted. The same is true of the `fileName` field.

davidebbo: e.g. one effect of `fileName` being wrong is that inline edit is broken. It sends e.g.
```json
{
  "oldText":"Hello2",
  "newText":"Hello3",
  "source":"Hello.tsx:4:8"
}
```

Where source comes from Files table. But then we expect to find that name in the Nodes table.

Ideally, in Dust FS mode, we stop using all those things from File, and only use values from Nodes.

PR #30831, opened 2026-08-19
Public API: allow admin to retrieve all agents including unpublished and restricted

fabiencelier:
## Description

closes https://github.com/dust-tt/tasks/issues/10131

Add a new scope to the `GET /api/v1/w/{wId}/assistant/agent_configurations?view=list` API which is `all_unrestricted` and allows admin to fetch all agents, including the one not published (owned by other users) and the one with restricted spaces.

## Tests

 - add new unit tests
 - tested manually locally

## Risk

expose unpublished data to admin via API
OK because admin is already expected to have the ability to see everything

## Deploy Plan

deploy front

### DIFF
--- a/front-api/public/swagger.json
+++ b/front-api/public/swagger.json
@@ -367,11 +367,12 @@
             "in": "query",
             "name": "view",
             "required": false,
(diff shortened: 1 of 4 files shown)


aubin-tchoi (approved):
LGTM

PR #30833, opened 2026-08-19
[Activation] Drop userId from work areas and authorize via pod editors

frankaloia:
## Summary
- Background: Activation work was introduced as an experimental property of a user's activation in order to improve the quality of recommendations. As we are entering the world of allowing admins to curate the recommendation experience for other users, this is becoming more important and it does not make sense to attach this to a nullable podId/userId combination. We need the activation pod to be the primary FK and enforce the relationship with the user:
- Drop the denormalized `userId` column. Make podId column non-nullable.
- Update logic to abide by normal space administration permissions
- Fine dropping rows with null pod IDs. Those aren't used today.
- Should be backwards compatible as we were already using podId everywhere

## Testing
* New unit tests
* Manually validated in local pod that tools work as expected

## Risk
* Low-ish, this does slightly alter the authorization model around work areas, but in a way to make it more consistent with rest of dust

## Deploy Plan
1. Pre deploy scripts attached
2. Deploy front
3. Post deploy scripts

### DIFF
--- a/front/migrations/pre-deploy/20260818120000_make_activation_work_area_pod_id_not_null.sql
+++ b/front/migrations/pre-deploy/20260818120000_make_activation_work_area_pod_id_not_null.sql
@@ -1,0 +1,6 @@
+-- Work areas now always belong to a specific Learning Space: podId is non-nullable.
+-- Drop any staging rows that still carry a null podId (pre-design-change data).
+DELETE FROM "public"."activation_work_areas" WHERE "podId" IS NULL;
+
+ALTER TABLE "public"."activation_work_areas"
+  ALTER COLUMN "podId" SET NOT NULL;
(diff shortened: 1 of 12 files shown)


front/migrations/pre-deploy/20260818120000_make_activation_work_area_pod_id_not_null.sql
```diff
@@ -0,0 +1,6 @@
+-- Work areas now always belong to a specific Learning Space: podId is non-nullable.
+-- Drop any staging rows that still carry a null podId (pre-design-change data).
+DELETE FROM "public"."activation_work_areas" WHERE "podId" IS NULL;
+
+ALTER TABLE "public"."activation_work_areas"
+  ALTER COLUMN "podId" SET NOT NULL;
```
id13, line 6: If you have rows with null values, this won't pass

id13 (approved):
LGTM, missing a migration script to drop rows with podId null values

PR #30835, opened 2026-08-19
[front] enh: capitalize ID in tool display names

aubin-tchoi:
## Description

Display names generated from identifiers render `id` in lowercase. Add `id` to the shared display-name special cases so it is formatted as `ID`, consistently with `GitHub`, `HubSpot`, and `MCP`.

This will affect tool names that contain the string "id", for instance `get_current_user_id` -> `Get current user ID` instead of `Get current user id`.

## Tests

- Tested locally.

## Risk

- Low.

## Deploy Plan

- Deploy front.


aubin-tchoi:
FYI @ellenatdust

aubin-tchoi:
PMRR

PR #30837, opened 2026-08-19
[front] enh: add table-shaped skeletons to Tools and Triggers

aubin-tchoi:
## Description

Loading the user Tools and Triggers dialog currently replaces each table with a centered spinner (not really centered: centered horizontally but not vertically).

Both tabs now render skeleton rows that match the target `DataTable` structure.

image

## Tests

- Tested locally.

## Risk

- Low.

## Deploy Plan

- Deploy front.

### DIFF
--- a/front/components/me/UserToolsTable.tsx
+++ b/front/components/me/UserToolsTable.tsx
@@ -29,8 +29,8 @@ import {
   DropdownMenuItem,
   DropdownMenuPortal,
   DropdownMenuTrigger,
+  LoadingBlock,
   SearchInput,
-  Spinner,
 } from "@dust-tt/sparkle";
 import type { ColumnDef } from "@tanstack/react-table";
 import keyBy from "lodash/keyBy";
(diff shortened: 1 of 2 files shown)


front/components/me/UserToolsTable.tsx
```diff
@@ -52,6 +52,80 @@ interface UserToolsTableProps {
...
+                className={classNames(
+                  "h-4 max-w-full",
+                  ["w-64", "w-52", "w-72", "w-60", "w-56"][row.index]
+                )}
+              />
+            </div>
+          </div>
+          <LoadingBlock className="h-6 w-20 shrink-0 rounded-[9px]" />
+        </div>
+      </DataTable.CellContent>
```
adrsimon, line 97: i'm pretty sure we have in sparkle a <DataTableLoading> or <DataTableSkeleton> component or smth

front/components/me/UserTriggersTable.tsx
```diff
@@ -33,6 +34,101 @@ interface UserTriggersTableProps {
...
+    accessorKey: "agentName",
+    header: "Agent",
+    cell: () => (
+      <DataTable.CellContent>
+        <div className="flex min-w-0 items-center gap-2">
+          <LoadingBlock className="h-7 w-7 shrink-0 rounded-md" />
+          <LoadingBlock className="h-4 w-20 max-w-full" />
+        </div>
+      </DataTable.CellContent>
+    ),
```
adrsimon, line 66: same

adrsimon (approved):
approving bc conceptually lgtm but i think we have a component designed specifically for this

front/components/me/UserToolsTable.tsx
aubin-tchoi: this one is super custom, will look into improving the component `DataTableSkeleton` to cover the simplest case (nothing prevents us from putting custom components in each cell so will always be better to fine tune the skeleton but could be a better start)

PR #30842, opened 2026-08-19
[front] fix: message consumption update tests

aubin-tchoi:
## Description

The message consumption response now includes sub-agent and total billed credits (added recently), but three exact assertions still expected the previous response format w/o them and fail.

This PR update the tests to expect the right format with the new fields.

## Tests

- Updated existing tests.

## Risk

- Low.

## Deploy Plan

- No deploy.


flvndvd (approved):
Oupsie 🙈 , thanks for fixing.

PR #30843, opened 2026-08-20
Fix stale test assertions in agent message consumption

avervaet:
Fix failing checks on main introduced by https://github.com/dust-tt/dust/pull/30829


flvndvd (commented):
Thoughts @aubin-tchoi fixed them already 🤔

PR #30847, opened 2026-08-20
[front] enh: turn the developer secrets list into a DataTable

adrsimon:
## Description

The Developer Secrets list in the admin section was a hand-rolled stack of divs: no loading state, no error state (the page rendered an empty list when the fetch failed), and no way to find a secret once you have a few dozen.

It's now a Sparkle `DataTable`, same shape as the other state of the art tables:

- `env.secrets.NAME` cell, replacing the `<pre>` + copy button + toast
- edit/delete moved into a row `MoreButton`
- `DataTableLoadingSkeleton` while loading, and a real error state
- search input filtering by name

Create/update and delete dialogs are untouched.

before
Capture d’écran 2026-08-20 à 10 24 46

after
Capture d’écran 2026-08-20 à 10 24 51

## Tests

Manual: created, edited and deleted a secret, copied a name, searched, and checked the empty and loading states.

## Risk

Low. One component, no API or data changes.

## Deploy Plan

Deploy front.

### DIFF
--- a/front/components/pages/workspace/developers/SecretsPage.tsx
+++ b/front/components/pages/workspace/developers/SecretsPage.tsx
@@ -4,24 +4,61 @@ import { useSubmitFunction } from "@app/lib/client/utils";
 import { clientFetch } from "@app/lib/egress/client";
 import { useDustAppSecrets } from "@app/lib/swr/apps";
 import type { DustAppSecretType } from "@app/types/dust_app_secret";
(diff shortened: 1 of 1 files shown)


ClementAupiais (changes requested):
Let's just show the two button on hover of the line instead of putting them in a more button

ClementAupiais (approved):
LGTM

PR #30848, opened 2026-08-20
Spool GitHub tarball to GCS before extracting it

flvndvd:
## Description

The GitHub code sync extraction activity shares a single pipeline between the tarball download and the per-file GCS uploads. On repositories dense with small files, upload backpressure (on GCS) paces the socket down to tens of kilobytes per second until codeload(.github.com) closes the connection, and every retry restarts from byte zero. In production this shows up as an activity that re-downloads and re-uploads the same first slice of the repository on every attempt without ever making net progress.

This change splits the work in two phases. The tarball is first spooled into a single GCS object at full network speed, with its own retry loop, so the GitHub connection only stays open for the duration of a raw download. Extraction then reads from the spooled object, and extraction retries re-read from GCS instead of going back to GitHub. The spooled object lives outside the extraction base path so index creation never picks it up, and the bucket lifecycle policy deletes it like the rest of the sync data.

Adding some more useful information in the logs. We should not expect faster syncs, but just syncs that finish. I'll monitor logs later today, to see if it needs tweaking other knobs.

### DIFF
--- a/connectors/src/connectors/github/lib/code/tar_extraction.ts
+++ b/connectors/src/connectors/github/lib/code/tar_extraction.ts
@@ -271,12 +271,12 @@ async function spoolTarballToGCS(
         isRetryableStreamError(error) &&
         attempt < MAX_TARBALL_SPOOL_RETRIES - 1
       ) {
-        const delay = TARBALL_RETRY_BASE_DELAY_MS * Math.pow(2, attempt);
+        const delayMs = TARBALL_RETRY_BASE_DELAY_MS * Math.pow(2, attempt);
         childLogger.warn(
-          { ...spoolContext, delay },
+          { ...spoolContext, delay: delayMs },
           "Retryable stream error while spooling tarball to GCS, will retry"
         );
-        await setTimeoutAsync(delay);
+        await setTimeoutAsync(delayMs);
         continue;
       }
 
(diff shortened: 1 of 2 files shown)


connectors/src/connectors/github/lib/code/tar_extraction.ts
```diff
@@ -204,22 +308,38 @@ export async function extractGitHubTarballToGCS(
...
     gcsBasePath,
+    tarballGcsPath,
   });
 
   childLogger.info(
     { repoId, connectorId, gcsBasePath },
     "Starting GitHub tarball extraction to GCS"
   );
 
+  const spoolResult = await spoolTarballToGCS(
```
flvndvd, line 325: Ideally would have been a dedicated activity, but will keep it here until we settle on those improvements.

connectors/src/connectors/github/lib/code/tar_extraction.ts
```diff
@@ -189,6 +191,108 @@ function parseGitHubPath(
...
 }
 
+/**
+ * Spool the GitHub tarball into a single GCS object at full network speed.
+ *
+ * Extraction (per-file GCS uploads) is orders of magnitude slower than the raw download.
+ * When both share one pipeline, upload backpressure stalls the GitHub socket until
+ * codeload closes it, and every retry re-downloads from scratch. Spooling keeps the
+ * GitHub connection open only for the duration of the download, and lets extraction
+ * retries re-read from GCS instead of GitHub.
```
id13, line 201: Ideally we only leave comments that explain the what and the how (if necessary).

The why with before / after is sloppy imo, wdyt ?

connectors/src/connectors/github/lib/code/tar_extraction.ts
```diff
@@ -189,6 +191,108 @@ function parseGitHubPath(
...
+        {
+          bytesReceived,
+          contentLength,
+          downloadDurationMs: Date.now() - downloadStartedAtMs,
+        },
+        "Spooled GitHub tarball to GCS"
+      );
+
+      return new Ok({ bytesSpooled: bytesReceived });
+    } catch (error) {
```
id13, line 254: are we sure we need this big try/catch with throw inside then catch right after ? 🙈

id13 (approved):
Big improvement 👍

connectors/src/connectors/github/lib/code/tar_extraction.ts
flvndvd: Agree, will clean everything after once I consolidate around the new activity, so we can actually re-resume if it fails, over restarting.

spolu (commented):
LGTM 👍

PR #30849, opened 2026-08-20
remove isCompactUIView column & model

xebeohp:
## Description
- Re-submit of the migration/model-cleanup commit from #30488, split out of the not-yet-ready phoebexu/activation-pod-ui/collapse-thinking stack so it can land on its own.
- Drops the obsolete isCompactUIView column from activation_pods; compact UI is now derived client-side from the conversation's uiView instead of being persisted on the Activation Pod record.
- Identical diff to what was already reviewed/approved in #30488.

## Risk
Low, no code depends on this anymore.

## Deploy Plan
Follow post-deploy migration steps.


flvndvd (approved):
LGTM

id13 (approved):

PR #30850, opened 2026-08-20
Add more logging in shadow compare

tdraier:
## Description

In case of mismatch, add more details in shadow:
- userId
- workspaceId
- group model ids
- permissions

Also re-fetch permissions and log the new results, to check if we should do a refresh.

## Risk

low, shadow code only

## Deploy Plan

deploy front


tdraier:

- #30850 ◀

matteotrab (approved):

PR #30853, opened 2026-08-20
[front] Batch project member space access check in agent mention permissions

PopDaph:
## Problem

`canAgentBeUsedInProjectConversation` (mention validation on `postUserMessage` / `editUserMessage` in project conversations) built one full `Authenticator` per project member via an unbounded `Promise.all`. Each `Authenticator.fromUserIdAndWorkspaceId` costs ~5-8 queries, so a 50-member project fired ~250-400 concurrent queries into the pod's 10-connection Sequelize pool before the message could be created; a 200-member project would exceed 1,000. This both slows message posting in large projects and can starve the whole pod's pool (same failure class as the recent connection pool acquisition delay alerts).

## Fix

- Replace the fan-out with the existing batched helper `listUsersWithoutAccessToSpaceResources` (one workspace-scoped `group_memberships` query for all members and spaces, then in-memory checks via `isMemberByGroupModelIds`). Cost is now O(1) queries regardless of project size.
- Align `isMemberByGroupPredicate` with the grant-write rule from `spaceGroupRoles`: provisioned groups carry no membership on manually-managed spaces. Without this, the swap from the grants path to the group-predicate path would false-allow an agent when a project member's only link to a required restricted space is a leftover provisioned group (a real state after a group -> manual management switch). A regression test pins this case: it passes on the old grants-based code, fails on the naive batched version, and passes with the predicate fix.

## Behavior change beyond the perf fix

Tightening `isMemberByGroupPredicate` also affects its two other consumers, the `POST /spaces/access-check` endpoint and skill space requirement checks (`findSkillEditorsWithoutSpaceAccess`). Both previously over-reported access for users reachable only through a provisioned group on a manual-mode space, while the grants that actually gate reads (and the legacy ACL path, which applies the same provisioned filter) deny it. They now answer from the same rule as the read gate. This is fail-closed: users are only ever moved from "has access" to "does not".

## Risk / rollback

Membership semantics for every other kind x managementMode combination are unchanged (grants and predicate paths verified case by case: regular open/restricted, project editor/member/global-viewer, system/conversations/global, suspended memberships, revoked workspace members). Single-commit revert restores the previous behavior.

## Tests

- New regression test: provisioned group on a manually-managed restricted space -> agent rejected.
- `permissions.test.ts` 28/28, `messages.test.ts` 28/28, `spaces.test.ts` 39/39, front-api `access-check.test.ts` 8/8.

### DIFF
--- a/front/lib/api/assistant/conversation/permissions.ts
+++ b/front/lib/api/assistant/conversation/permissions.ts
@@ -1,5 +1,6 @@
 import { getAgentConfigurations } from "@app/lib/api/assistant/configuration/agent";
 import { getContentFragmentsSpaceIds } from "@app/lib/api/assistant/permissions";
+import { listUsersWithoutAccessToSpaceResources } from "@app/lib/api/spaces/access";
 import { Authenticator } from "@app/lib/auth";
 import type { ConversationAccessType } from "@app/lib/resources/conversation_resource";
 import { ConversationResource } from "@app/lib/resources/conversation_resource";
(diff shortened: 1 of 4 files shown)


front/lib/api/assistant/conversation/permissions.ts
```diff
@@ -135,23 +136,12 @@ export async function canAgentBeUsedInProjectConversation(
           await project.fetchDistinctActiveManualGroupMembers(auth);
 
         if (restrictedAgentSpaces.length > 0 && projectMembers.length > 0) {
-          const workspaceId = auth.getNonNullableWorkspace().sId;
-          const memberAuths = await Promise.all(
-            projectMembers.map((member) =>
-              Authenticator.fromUserIdAndWorkspaceId(member.sId, workspaceId)
-            )
-          );
```
id13, line 143: 😭 

nice catch !

fabiencelier (approved):

id13 (approved):

PR #30854, opened 2026-08-20
Workspace management MCP: Add a new MCP to list agents/skills and get their details

fabiencelier:
## Description

New MCP server `workspace management` with the following tools:
 - `list_agents`
 - `list_skills`
 - `get_agent_details`
 - `get_skills`

These are accessible to all users.
Exposed in the `@analytics` agent only for now (via their skill), will be exposed more globally later
It is a first step with additional functions comings next, including actual actions to manage the workspace (list_tools, get_tool_details, archive agent, make discoverable, change skill availability, update editors, list inactive agents, list users...)

All the tools are free 

## Tests

 - added unit tests
 - tested locally

## Risk

security: the data is already accessible via the UI and API (except for the all_unrestricted listing of agents which is gated to admins only)

## Deploy Plan

deploy front


front/lib/api/actions/servers/workspace_management/tools/index.ts
```diff
@@ -0,0 +1,306 @@
...
+// the public agent_configurations endpoint does.
+function resolveAgentView(view: AgentViewType): {
+  agentsGetView: AgentsGetViewType;
+  dangerouslySkipPermissionFiltering: boolean;
+} {
+  switch (view) {
+    case "all_unrestricted":
+      return {
+        agentsGetView: "admin_internal",
+        dangerouslySkipPermissionFiltering: true,
```
flvndvd: Slightly worried that this ends up being exposed behind an MCP server, what's the exact need here?

front/lib/api/actions/servers/workspace_management/tools/index.ts
```diff
@@ -0,0 +1,306 @@
...
+    return new MCPError(
+      "The 'list' view requires an interactive user; use 'all' instead.",
+      { tracked: false }
+    );
+  }
+  return null;
+}
+
+function skillKindOf(skill: SkillResource): "custom" | "global" {
+  return isResourceSId("skill", skill.sId) ? "custom" : "global";
```
flvndvd: Really? Looks like `SkillResource` exposes a kind, why not using it?

front/lib/api/actions/servers/workspace_management/tools/index.ts
```diff
@@ -0,0 +1,306 @@
...
+}
+
+const handlers: ToolHandlers<typeof WORKSPACE_MANAGEMENT_TOOLS_METADATA> = {
+  [LIST_AGENTS_TOOL_NAME]: async (
```
flvndvd: Can we do one tool per file like the newest MCPs 🙏 ?

front/lib/api/actions/servers/workspace_management/tools/index.ts
```diff
@@ -0,0 +1,306 @@
...
+    { view, namePrefix, cursor, limit },
+    { auth }
+  ) => {
+    const resolvedView = view ?? "all";
```
flvndvd: Can we set `all` as default in the zod schema? So the model sees the default value?

front/lib/api/actions/servers/workspace_management/tools/index.ts
```diff
@@ -0,0 +1,306 @@
...
+          `${agent.instructions ?? "(no instructions)"}`,
+      },
+    ]);
+  },
+
+  [LIST_SKILLS_TOOL_NAME]: async (
+    { availability, status, kind, includeUsage, cursor, limit },
+    { auth }
+  ) => {
+    const resolvedKind = kind ?? "custom";
```
flvndvd: Same here default param in the zod schema?

front/lib/api/actions/servers/workspace_management/tools/index.ts
```diff
@@ -0,0 +1,306 @@
...
+            "accessible).",
+        },
+      ]);
+    }
+
+    // `toJSON` is what decides whether a code-defined skill exposes its instructions.
+    const json = skill.toJSON(auth);
+
+    return new Ok([
+      makeJsonText({
```
flvndvd: This seems pretty heavy, pretty sure it's going to end up in a file as big payloads get offloaded 🙃

front/lib/api/actions/servers/workspace_management/tools/index.ts
```diff
@@ -0,0 +1,306 @@
...
+        nextCursor,
+        skills: page.map((skill) => ({
+          sId: skill.sId,
+          name: skill.name,
+          userFacingDescription: skill.userFacingDescription,
+          agentFacingDescription: skill.agentFacingDescription,
+          availability: skill.availability,
+          status: skill.status,
+          kind: skillKindOf(skill),
+          canWrite: skill.canWrite(auth),
```
flvndvd: Not sure to see what the `canWrite` allows here?

front/lib/api/actions/servers/workspace_management/tools/index.ts
```diff
@@ -0,0 +1,306 @@
...
+import { getAgentConfigurationsForView } from "@app/lib/api/assistant/configuration/views";
+import type { Authenticator } from "@app/lib/auth";
+import { SkillResource } from "@app/lib/resources/skill/skill_resource";
+import { isResourceSId } from "@app/lib/resources/string_ids";
+import type { AgentsGetViewType } from "@app/types/assistant/agent";
+import type { Result } from "@app/types/shared/result";
+import { Err, Ok } from "@app/types/shared/result";
+import { assertNever } from "@app/types/shared/utils/assert_never";
+
+function makeJsonText(value: unknown) {
```
flvndvd: Any chance we can avoid JSON and return a list to save some tokens 🤑?

front/lib/api/actions/servers/workspace_management/metadata.ts
```diff
@@ -0,0 +1,198 @@
...
+// (unpublished agents the caller does not edit) and the space one.
+const SKILL_AVAILABILITY_VALUES = [
+  "editors",
+  "workspace_users",
+  "users_and_agents",
+] as const;
+
+const SKILL_STATUS_VALUES = ["active", "archived", "suggested"] as const;
+
+export const AGENT_VIEWS = [
```
flvndvd: OOC Why do we duplicated the types here? Can't we use the existing one? Or at least extend/extract them?

flvndvd (approved):

front/lib/api/actions/servers/workspace_management/tools/index.ts
fabiencelier: need is for admin to list all agents, including the unpublished ones and the one with restricted spaces 
this is gated by the workspaceAdminGuard throwing if the user is not an admin:
```
  if (view === "all_unrestricted") {
    return workspaceAdminGuard(auth);
  }
```

front/lib/api/actions/servers/workspace_management/tools/index.ts
fabiencelier: let the agent know if they can edit the skill
Useful for the skill authoring skill for instance to know if they can or not edit it

front/lib/api/actions/servers/workspace_management/metadata.ts
fabiencelier: some import cycle 
I have extracted the values to another file so they can be shared and won't be copied 👍

front/lib/api/actions/servers/workspace_management/tools/index.ts
fabiencelier: there was no existing `kind()` in the resource but I added one to centralize this.

front/lib/api/actions/servers/workspace_management/tools/index.ts
fabiencelier: done

front/lib/api/actions/servers/workspace_management/tools/index.ts
fabiencelier: done

front/lib/api/actions/servers/workspace_management/tools/index.ts
fabiencelier: done

front/lib/api/actions/servers/workspace_management/tools/index.ts
fabiencelier: removed the JSON

PR #30855, opened 2026-08-20
Microsoft: use a lean delta projection to reduce worker memory

tdraier:
## Description

Incremental Microsoft sync workers were OOM-killed (`OOMKilled`, exit 137) during `fetchDeltaForRootNodesInDrive` on large drives.

`getFullDeltaResults` keeps every latest `DriveItem` in an in-memory map until delta pagination completes. The default delta projection expands `listItem($expand=fields)` (full SharePoint custom-column payloads) and selects `@microsoft.graph.downloadUrl` on **every** item — including folders, root markers and deleted entries that never need them. That inflates the retained map enough to exhaust the container memory limit.

This PR fetches delta pages with a lean projection (`DRIVE_ITEM_DELTA_SELECTS`) that omits the `listItem` fields expansion and the download URL. `syncOneFile` already re-hydrates both per file on demand via `getItem` when they are absent, so files that are actually synced still get full metadata; folders, deleted items and roots no longer carry SharePoint field payloads through the dedup map.

Also:
- Collapses the now-expected `missing_download_url`/`missing_fields` stats into a single `microsoft.file.metadata_hydration` counter (hydration is the expected path for incrementally-synced files, no longer an anomaly).
- Drops the dead `withLabels` parameter from `getDeltaResults` (no caller passed it; the full-sync `getFilesAndFolders` path is unchanged).

## Tests

`tsc --noEmit` passes. Change is behaviorally covered by the existing `syncOneFile` hydration fallback, which already runs in production whenever Graph omits `downloadUrl`/`fields`. Verified the delta consumer never reads `listItem.fields`/`downloadUrl` directly off delta items: folders/roots go through `itemToMicrosoftNode` (name/webUrl/parentReference only), deleted items use only the internalId, and only real files reach `syncOneFile`.

## Risk

Low. Tradeoff is one extra Graph `GET` per *changed file* per incremental cycle (folders/deleted/roots cost nothing extra); changed-file counts per incremental cycle are normally small. Fully rollback-safe — reverting restores the previous projection. Monitor `microsoft.file.metadata_hydration` (extra API volume) and worker RSS after deploy.

## Deploy Plan

Standard connectors deploy.

### DIFF
--- a/connectors/src/connectors/microsoft/lib/types.ts
+++ b/connectors/src/connectors/microsoft/lib/types.ts
@@ -29,6 +29,19 @@ export const DRIVE_ITEM_EXPANDS_AND_SELECTS =
 export const DRIVE_ITEM_EXPANDS_AND_SELECTS_WITH_LABELS =
   "$select=id,name,parentReference,webUrl,file,folder,root,deleted,createdBy,lastModifiedBy,createdDateTime,lastModifiedDateTime,size,sharepointIds&$expand=listItem($expand=fields($select=_IpLabelId))&select=@microsoft.graph.downloadUrl";
 
+// Lean projection used for delta pagination. Unlike the projections above it omits
+// the `$expand=listItem($expand=fields)` expansion and the `@microsoft.graph.downloadUrl`
+// select. Those two are the memory-heavy parts of a `DriveItem`: expanded SharePoint
+// custom-column payloads and a per-item download URL. `getFullDeltaResults` keeps every
+// latest `DriveItem` in an in-memory map until pagination completes, so carrying that
+// metadata for every folder, root marker and deleted entry (none of which need it)
+// inflates the worker's footprint and can OOM the pod on large drives.
+// `syncOneFile` already re-hydrates the download URL and list-item fields per file on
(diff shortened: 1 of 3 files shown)


tdraier:

- #30855 ◀

connectors/src/connectors/microsoft/lib/types.ts
```diff
@@ -29,6 +29,19 @@ export const DRIVE_ITEM_EXPANDS_AND_SELECTS =
 export const DRIVE_ITEM_EXPANDS_AND_SELECTS_WITH_LABELS =
```
PopDaph, line 29: NIT: could be `DRIVE_ITEM_DELTA_SELECTS` + the query param

PopDaph (approved):
Thanks very much 🙏🏻

connectors/src/connectors/microsoft/lib/types.ts
tdraier: i'd rather keep them independents if we can/need to remove other fields from the DELTA_SELECTS

PR #30856, opened 2026-08-20
[connectors] Disable resumable GCS uploads for small files and log upload queue depth

flvndvd:
## Description

This PR is a follow-up to the tarball spooling change https://github.com/dust-tt/dust/pull/30848. With the download decoupled (biggest repo taking ~ 3 minutes to download from GH and upload in GCS), the first few production runs showed the extraction phase still crawling at a very slow pace. Looking deeper all those files are small one entering the < 10MB buffering + direct upload.

Every buffered small file goes through the storage client's save method, which defaults to resumable uploads. The client's own documentation recommends disabling the resumable feature below a few megabytes because the extra session-creation round trip causes noticeable degradation across a series of small files. 

As we don't rely on GH stream anymore when uploading file, this PR also bumps concurrency from 200 to 400.

### DIFF
--- a/connectors/src/connectors/github/lib/code/gcs_repository.ts
+++ b/connectors/src/connectors/github/lib/code/gcs_repository.ts
@@ -196,6 +196,8 @@ export class GCSRepositoryManager {
         metadata: {
           contentType: options?.contentType || "text/plain",
         },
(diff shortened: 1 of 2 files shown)


flvndvd:
PMRR.

id13:
LGTM

PR #30857, opened 2026-08-20
[front] enh: polish conversation consumption breakdown

aubin-tchoi:
## Description

Implements the following [Figma](https://www.figma.com/design/wJJMfVF6bluurSKfrEuysc/Product---WIP?node-id=14130-34475&m=dev).

This PR aims at adding a better hierarchy to the conversation consumption breakdown.

Renames the headings, adds dividers between sections, adds a border around each agent breakdown (kept when collapsing/uncollapsing).

Single-agent conversations continue to omit the agent section, and multi-agent details remain collapsed by default.

One principle followed here is that there are now a few clear sections (clear because divided by dividers), and the sum of the credits in each sections amounts to the total shown at the top. Separating makes it clear that these are different ways to look at the same thing, rather than something additive: we don't want users surprised that the sum of models + agents is higher (twice actually) the total credits.

Mockup

image

Current live version

image

Proposal

image

## Tests

- Updated existing tests.
- Tested locally + preview.

## Risk

- Low.

## Deploy Plan

- Deploy front.

### DIFF
--- a/front/components/assistant/conversation/credits_panel/ConversationCreditUsageBreakdown.tsx
+++ b/front/components/assistant/conversation/credits_panel/ConversationCreditUsageBreakdown.tsx
@@ -191,7 +191,7 @@ function AgentBreakdown({ agent }: AgentBreakdownProps) {
     >
       <CollapsibleTrigger
         aria-label={`${isOpen ? "Collapse" : "Expand"} credit details for ${agent.name}`}
-        className="min-h-11 flex-row-reverse justify-between rounded-xl px-2 text-left focus-visible:ring-2 focus-visible:ring-ring"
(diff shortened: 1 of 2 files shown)


flvndvd (approved):

front/components/assistant/conversation/credits_panel/ConversationCreditUsageBreakdown.tsx
```diff
@@ -161,8 +161,8 @@ function ModelsBreakdown({ isDark, models }: ModelsBreakdownProps) {
   }
 
   return (
-    <section className="space-y-4">
-      <h3 className="text-xs font-semibold text-muted-foreground">Models</h3>
+    <section className="mt-6 space-y-4 border-t border-border pt-6">
```
aubin-tchoi, line 164: mt-6 pt-6 is what I use here to position a centered divider
padding is handled bottom to top because we may add a section (the agent) to the bottom with no divider at the end

PR #30858, opened 2026-08-20
feat(agent_inactivity): fetch inactive agent candidates from mentions 

achilleburah:
## Description

First brick of the opt-in "automatically archive inactive agents" feature: the business rules and the use case that applies them. No API, Temporal or Governance wiring yet — this only encodes and tests the agreed rules, plus the read path they run against.

- `inactivity/policy.ts`: domain rules. Threshold validation (integer, 2–366 days, no implicit default), cutoff resolution, per-agent eligibility (`agent_not_active`, `active_schedule`, `recent_creation`, `recent_mention`).
- `inactivity/archive_inactive_agents.ts`: the use case. Resolves the policy into a cutoff, fetches candidates, judges each against the rules, archives through the existing `archiveAgentConfiguration`.
- `inactivity/fetch_inactive_agents.ts` + `resources/mention_resource.ts`: the read path. Driven by the workspace's active agents, with mentions as an anti-join — an agent nobody ever mentioned has no row in `mentions` at all, so no query selecting from that table can return it.
- `configuration/agent.ts`: `fetchFirstVersionCreatedAtByAgentId`, batched lookup of an agent's earliest version.

An invalid threshold is returned as an `Err` for the whole operation rather than a per-agent exclusion, so a misconfigured workspace can never be mistaken for "no agent to archive".

Activity is read from mentions rather than agent messages: mention rows carry the stable agent `sId` with no configuration version, and `(workspaceId, agentConfigurationId, createdAt)` is already indexed. Any mention counts, whatever its status — a mention Dust blocked on space permissions still means somebody reached for the agent. Global agents can never be archived: they have no row in `agent_configurations`, which is what the read selects from.

An agent must also predate the cutoff, using the *first* version's `createdAt` rather than the active row's — upgrading an agent inserts a new row, so the active row's own date is the last edit, and reading that would let editing an agent postpone its archival.

## Performance

Two things measured on the read replica (workspace with 3,630 active agents, 809,715 mentions).

**Why the mentions read stays a hand-written `LATERAL` instead of a `GROUP BY` over a batched `IN`/`ANY` list.** Same index either way; the difference is that `GROUP BY` has to feed every row of a group to the aggregate (Postgres has no loose index scan), so it reads every mention of every agent instead of one row per agent.

<details>
<summary>EXPLAIN — LATERAL vs GROUP BY, same workspace and population</summary>

```
LATERAL (current):
  Index Only Scan ... rows=1 per agent
  3,630 rows scanned, 22,575 buffers, 1,080 ms

GROUP BY + ANY(array) batching:
  Index Only Scan ... rows=8,673 for a 100-agent batch (extrapolates to the full run)
  809,816 rows scanned, 510,266 buffers, 9,312 ms
```

On a workspace with few custom-agent mentions the two forms are roughly equivalent — the gap only opens once agents actually have mentions to aggregate over.
</details>

**A second `LATERAL` computing `MIN("createdAt")` for the agent's first version was the real cost — not the sort.** As a correlated subquery it made the planner build a `BitmapAnd` that re-scans the workspace's agents once per agent. Replaced with a batched `version = 0` lookup on the unique `(sId, version)` index.

<details>
<summary>EXPLAIN — before/after the version = 0 fix, same workspace</summary>

```
Before (MIN(createdAt) in a LATERAL):
  BitmapAnd ... rows=62,996 per agent, loops=3,630
  823,673 buffers, ~37 s

After (batched version = 0 lookup):
  Index Scan ... rows=1 per agent, loops=3,630
  18,013 buffers, 69 ms
```
</details>

Net effect: the full read went from ~845k buffers / 38.6s to ~40k buffers / ~1.15s on that workspace. The `ORDER BY "sId"` sort is negligible (quicksort, small memory), so no additional index is needed on `agent_configurations`.

## Tests

Domain and use-case unit tests, plus functional tests for the fetch and the executor through factories (`AgentConfigurationFactory`, `MentionFactory`, `TriggerFactory`).

## Risk

Not called from any endpoint or workflow yet — dead code path, no user-facing effect.

## Deploy Plan

None

### DIFF
--- a/front/lib/api/assistant/inactivity/archive_inactive_agents.ts
+++ b/front/lib/api/assistant/inactivity/archive_inactive_agents.ts
@@ -2,10 +2,11 @@ import {
   archiveAgentConfiguration,
   getAgentConfigurations,
 } from "@app/lib/api/assistant/configuration/agent";
+import type { AgentPageBound } from "@app/lib/api/assistant/inactivity/fetch_inactive_agents";
+import { fetchInactiveAgents } from "@app/lib/api/assistant/inactivity/fetch_inactive_agents";
 import type {
   AgentArchivalExclusionReason,
   AgentInactivityPolicyError,
-  AgentInactivitySnapshot,
   AgentTriggerSnapshot,
 } from "@app/lib/api/assistant/inactivity/policy";
 import {
@@ -20,33 +21,24 @@ import type { Result } from "@app/types/shared/result";
 import { Err, Ok } from "@app/types/shared/result";
 
 /**
- * Archives a workspace's inactive agents: resolve the threshold into a cutoff, load the agents,
- * judge each one against the rules in `policy.ts`, and archive those that are eligible.
+ * Archives a workspace's inactive agents. Both the manual endpoints and the nightly Temporal activity
+ * enter here, so the rules cannot drift between them; the population still differs, since every read
+ * is permission-filtered.
  *
- * The manual endpoints and the nightly Temporal activity both enter here, so the rules cannot drift
- * between them. The population can: every read is permission-filtered, so a manual run only ever
- * considers the agents its caller can see, where the nightly run sees the whole workspace.
+ * The page's facts are re-read before it is judged, which closes the window in which an agent can be
+ * restored, archived by someone else, or given a schedule. A retry is safe for a different reason: it
+ * re-runs the fetch, and an agent the first attempt archived is no longer active, so it is not a
+ * candidate again.
  *
- * The rules are applied against state re-read for the page rather than against what the fetch
- * loaded, because the two do not answer the same question: the fetch finds candidates, the re-read
- * goes through `getAgentConfigurations` and is therefore permission-filtered. That is also what
- * makes a Temporal retry safe — replaying the same input re-reads an agent the first attempt
- * archived, sees `archived`, and skips it instead of emitting a second `agent.archived` audit event.
- *
- * Archiving itself stays the existing `archiveAgentConfiguration` primitive, which owns the side
- * effects (trigger disabling, wake-up cancellation, editor group suspension, audit event).
+ * Archiving stays `archiveAgentConfiguration`, which owns the side effects.
  */
 
-/**
- * Why an agent this operation looked at was not archived: the rules' own exclusions, plus the two
- * outcomes only the mutation path can produce and that the rules therefore never see.
- */
+/** The rules' exclusions, plus the two outcomes only the mutation path can produce. */
 export type AgentArchivalSkipReason =
   | AgentArchivalExclusionReason
-  // The configuration disappeared, or is not visible to this actor, between the fetch and the
-  // re-read.
+  // Gone, or not visible to this actor, since the fetch.
   | "agent_not_found"
-  // The archival update matched no row: someone else archived the agent in the meantime.
+  // The update matched no row: someone else archived it first.
   | "archive_raced";
 
 export interface AgentArchivalSkip {
@@ -57,6 +49,8 @@ export interface AgentArchivalSkip {
 export interface InactiveAgentsArchivalInput {
   thresholdDays: number;
   evaluatedAt: Date;
+  // Which slice to walk, not what to decide. The caller owns the loop.
+  page: AgentPageBound;
 }
 
 export interface InactiveAgentsArchival {
@@ -64,6 +58,8 @@ export interface InactiveAgentsArchival {
   cutoffAt: Date;
   archivedAgentIds: string[];
   skipped: AgentArchivalSkip[];
+  // Paging, not outcome: null once the workspace is exhausted.
+  nextCursor: string | null;
 }
 
 export type InactiveAgentsArchivalResult = Result<
@@ -88,10 +84,7 @@ function countSkipsByReason(skipped: AgentArchivalSkip[]): SkipCountsByReason {
   return counts;
 }
 
-/**
- * Re-reads the whole page's facts in two batched queries, before any agent of it is judged. An
- * agent missing from the returned map is one this actor cannot currently read.
- */
+/** Two batched queries for the whole page. A missing agent is one this actor cannot read. */
 async function fetchArchivalFacts(
   auth: Authenticator,
   agentIds: string[]
@@ -132,7 +125,7 @@ async function fetchArchivalFacts(
  */
 export async function archiveInactiveWorkspaceAgents(
   auth: Authenticator,
-  { thresholdDays, evaluatedAt }: InactiveAgentsArchivalInput
+  { thresholdDays, evaluatedAt, page }: InactiveAgentsArchivalInput
 ): Promise<InactiveAgentsArchivalResult> {
   const workspace = auth.getNonNullableWorkspace();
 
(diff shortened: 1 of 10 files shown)


front/lib/api/assistant/inactivity/archive_inactive_agents.ts
```diff
@@ -150,26 +143,18 @@ export async function archiveInactiveWorkspaceAgents(
...
-  // limit) becomes an input of this use case then, and `nextCursor` part of its result, so the
-  // caller keeps owning the loop. Status and triggers do not need to come back with it —
-  // `fetchArchivalFacts` is the authority on those, since it is the permission-filtered read.
-  //
-  // Read from the `mentions` table, not from `agentMentionsCount` in `agent_usage.ts`: its Redis
-  // cache only goes back 30 days, which is shorter than the thresholds this feature exists for, and
-  // the Elasticsearch index behind it is best-effort. Neither is a basis for a destructive decision.
-  // const { agents, nextCursor } = await fetchInactiveAgents(auth, { cutoffAt, cursor, limit });
-  const agents: AgentInactivitySnapshot[] = [];
+  const { agents, nextCursor } = await fetchInactiveAgents(auth, {
```
fabiencelier: Question for me: does nextCursor stay valid if we have archive some of the agents

front/lib/resources/mention_resource.ts
```diff
@@ -0,0 +1,154 @@
...
+  /**
+   * The workspace's active agents not mentioned since `notMentionedSince`, keyed from `cursor`, with
+   * when each was last mentioned. Every mention counts whatever its `status` or `dismissed` flag —
+   * somebody reached for the agent.
+   *
+   * Also filters on `status = 'active'` (which leaves exactly one row per logical agent) and on the
+   * agent having first appeared before the cutoff. Both stay *broader or equal* to the rules in
+   * `lib/api/assistant/inactivity/policy.ts`, which remain the authority.
+   */
+  static async listAgentsNotMentionedSince(
```
fabiencelier: add a unit test on this one to make sure it works

front/lib/resources/mention_resource.ts
```diff
@@ -0,0 +1,154 @@
...
+    )
+  ORDER BY agent."sId"
+  LIMIT $limit
+`;
+
+// Attributes are marked as read-only to reflect the stateless nature of our Resource.
+// eslint-disable-next-line @typescript-eslint/no-unsafe-declaration-merging
+export interface MentionResource extends ReadonlyAttributesType<MentionModel> {}
+// eslint-disable-next-line @typescript-eslint/no-unsafe-declaration-merging
+export class MentionResource extends BaseResource<MentionModel> {
```
fabiencelier, line 42: Good initative to create the resource 👍 
I think it can be used in many places to avoid raw model use but out of scope for this PR

front/lib/resources/mention_resource.ts
```diff
@@ -0,0 +1,154 @@
...
+    AND agent.status = 'active'
+    AND first_version."createdAt" < $notMentionedSince
+    AND ($cursor::text IS NULL OR agent."sId" > $cursor)
+    AND (
+      last_mention."lastMentionedAt" IS NULL
+      OR last_mention."lastMentionedAt" < $notMentionedSince
+    )
+  ORDER BY agent."sId"
+  LIMIT $limit
+`;
```
fabiencelier: as we are doing it by batch anyway, could we maybe avoid doing this raw SQL by doing a first query to retrieve all the agents to consider (active, creation date < ... )
then we batch them by 100 and to a query on the mention to fetch the last mention with a ISIN on agent id
Each batch won't return 100 candidates, but it's not really an issue
And it removes the necessity to manually handle limits and cursor 
I'm not sure about the performance exactly but it should be performant considering the the index on mention model on (wid, agent id, created at) which will be the queried columns

front/lib/resources/mention_resource.ts
achilleburah: yeah, i'm planning a PR after this to refactor existing code with direct access :+1:

front/lib/resources/mention_resource.ts
achilleburah: Hey so I dug a little into this. Executed different queries on the read replica to compare execution time. 
One of the LATERAL was wrong though (LATERAL doing MIN(createdAt)) => this added ~35s per query
So I replaced the MIN with a batched version. 

Then for the other one, the thing is that batching forces us to group by instead of using a LATERAL, and the index does not react the same way. 

Here are the results for the biggest workspace: 
LATERAL      3 630 rows scanned    22 575 buffers    1 080 ms
GROUP BY   809 816 rows scanned   510 266 buffers    9 312 ms

So I changed the first part only for now, I can also remove the second one if we want to avoir the "rawSql" 

what do you think ?

fabiencelier: ok, a query like this is a bit harder to maintain and understand than a sequence of simpler query, but if perf is really better then why not

fabiencelier: invoking @id13 here for a check to make sure we are not going to the DB jail

front/tests/utils/AgentConfigurationFactory.ts
```diff
@@ -126,4 +127,28 @@ export class AgentConfigurationFactory {
...
+    await frontSequelize.query(
+      `UPDATE agent_configurations SET "createdAt" = :createdAt WHERE "sId" = :agentId AND "workspaceId" = :workspaceId`,
+      {
+        replacements: {
+          createdAt: createdAt.toISOString(),
+          agentId,
+          workspaceId: auth.getNonNullableWorkspace().id,
+        },
+      }
+    );
```
id13: Let's avoid raw sql when it's not needed 🙏

front/lib/resources/mention_resource.ts
id13:
```  SELECT
    agent."sId" AS "agentId",
    last_mention."lastMentionedAt" AS "lastMentionedAt"
  FROM agent_configurations agent
  LEFT JOIN LATERAL (
    SELECT MAX(mention."createdAt") AS "lastMentionedAt"
    FROM mentions mention
    WHERE mention."workspaceId" = agent."workspaceId"
      AND mention."agentConfigurationId" = agent."sId"
  ) last_mention ON true
  WHERE agent."workspaceId" = $workspaceId
    AND agent.status = 'active'
    AND (
      last_mention."lastMentionedAt" IS NULL
      OR last_mention."lastMentionedAt" < $notMentionedSince
    )
  ORDER BY agent."sId"
```

this is very efficient 👍 

the comment above, though, is pure AI slope, let's not add more of these in the codebase.

I would simply write something like

`we want old mention rows for active agents. This is the most efficient way to do it because card(agent) << card(mentions)`


Also, I would put the raw SQL inside the method, this not a constant, it's code 😄

id13 (approved):
LGTM


How are you going to dispatch this query btw ?

front/lib/resources/mention_resource.ts
fabiencelier: thinking a bit more about this, do you really need the GROUP BY in the second query ?
YOu can just do something like 
```sql
SELECT UNIQUE agentId
FROM MENTIONS
WHERE wid = wid
  AND agentId in $LIST_OF_IDDS
  AND createdAT > cutOff
```
And you have all the ids the of the agent with rencent mentions so you can archive the other one ?

achilleburah: Yeah, rn I feel like I spend most of my time removing useless comments from Claude...

achilleburah:
> How are you going to dispatch this query btw ?

- Endpoint for manual execution + dry-run
- Temporal workflow for daily check

@id13

fabiencelier (approved):

PR #30862, opened 2026-08-20
[front] feat: pool picker in the trigger sheet

adrsimon:
## Description

Third PR of the trigger pool stack (#30860). The schedule and webhook sheets get a Credits picker, disabled unless an admin granted `use_workspace_pool`, and webhooks get an editable daily run cap where the limit used to be read-only text. Schedules now carry an execution mode end to end; POST and PATCH honor it when the flag is on, and leave the stored pool alone when the field is absent.

## Tests

Functional tests on create and update: permitted admin, rejected builder, flag off, and an update that omits the field keeping the trigger's pool.

## Risk

Low, flagged off. Without the grant the picker is read-only and the server rejects the change.

## Deploy Plan

Deploy front.

### DIFF
--- a/front-api/routes/w/[wId]/assistant/agent_configurations/[aId]/triggers/index.ts
+++ b/front-api/routes/w/[wId]/assistant/agent_configurations/[aId]/triggers/index.ts
@@ -37,7 +37,7 @@ const PostTriggersRequestBodySchema = z.object({
 });
 
 function requestedExecutionMode(
-  trigger: { executionMode?: TriggerExecutionMode },
+  trigger: z.infer<typeof TriggerSchema>,
   featureFlags: WhitelistableFeature[]
 ): TriggerExecutionMode | undefined {
   if (!featureFlags.includes("trigger_pool_choice")) {
--- a/front-api/routes/w/[wId]/assistant/agent_configurations/[aId]/triggers/index.ts
+++ b/front-api/routes/w/[wId]/assistant/agent_configurations/[aId]/triggers/index.ts
@@ -1,4 +1,5 @@
 import { getAgentConfiguration } from "@app/lib/api/assistant/configuration/agent";
+import { getFeatureFlags } from "@app/lib/auth";
 import { getResourceIdFromSId } from "@app/lib/resources/string_ids";
 import {
   resolveTriggerSpaceId,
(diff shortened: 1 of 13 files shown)


front-api/routes/w/[wId]/assistant/agent_configurations/[aId]/triggers/index.ts
```diff
@@ -32,6 +36,17 @@ const PostTriggersRequestBodySchema = z.object({
   triggers: z.array(TriggerSchema),
 });
 
+function requestedExecutionMode(
+  trigger: { executionMode?: TriggerExecutionMode },
```
Comment (aubin-tchoi) [inline front-api/routes/w/[wId]/assistant/agent_configurations/[aId]/triggers/index.ts]: let's inline this or put a proper type for trigger

front/lib/resources/trigger_resource.ts
```diff
@@ -1070,7 +1074,12 @@ export class TriggerResource extends BaseResource<TriggerModel> {
     auth: Authenticator,
     executionMode: TriggerExecutionMode
   ): Promise<Result<undefined, Error>> {
-    if (!(await this.canSetExecutionMode(auth, executionMode))) {
+    const isEditor =
+      auth.isManager() || this.editor === auth.getNonNullableUser().id;
+    if (
+      !isEditor ||
+      !(await TriggerResource.canUseExecutionMode(auth, executionMode))
```
aubin-tchoi: nit
```suggestion
      !(await this.canUseExecutionMode(auth, executionMode))
```

front-api/routes/w/[wId]/assistant/agent_configurations/[aId]/triggers/index.ts
```diff
@@ -380,12 +441,22 @@ app.post(
...
         webhookSourceViewId,
         executionPerDayLimitOverride: executionPerDay,
-        executionMode: "user_pool",
+        executionMode,
         origin: "user",
         spaceId: spaceIdRes.value,
       });
 
       if (newTrigger.isErr()) {
+        if (newTrigger.error instanceof TriggerExecutionModeForbiddenError) {
```
Comment (aubin-tchoi) [inline front-api/routes/w/[wId]/assistant/agent_configurations/[aId]/triggers/index.ts:450]: seems to be a pre-existing issue but not a fan of relying on an error type this way vs a field inside the Err

front/components/agent_builder/triggers/TriggerPoolSelector.tsx
```diff
@@ -0,0 +1,75 @@
...
+    return null;
+  }
+
+  const canSetPool = hasPermission("use_workspace_pool", "trigger");
+
+  return (
+    <div className="space-y-1">
+      <Label htmlFor="trigger-pool">Credits</Label>
+      <p className="text-sm text-muted-foreground">
+        Where the credits spent by this trigger's runs are taken from.
```
aubin-tchoi: nit: very descriptive copy but could be simplified to be more declarative (stating the thing instead of describing it). smth around "which pool this takes from"

aubin-tchoi (approved):
LGTM
do you have a screenshot?

PR #30864, opened 2026-08-20
[front] Rename Automations nav item and API section

avervaet:
## Description

Renames two left-bar nav labels:
- The "Automations" item is renamed to "Programmatic Usage" and its icon swaps from `Clock` to `Lightning01` to better reflect what the page covers.
- The parent "API" section is renamed to "API & Programmatic" to reflect that it now also covers the programmatic usage item.

Source: https://dust4ai.slack.com/archives/C0BJ3T5SA0L/p1787128163117979

**Now:**
Capture d’écran 2026-08-20 à 13 30 17

**Before:**
Capture d’écran 2026-08-20 à 13 30 45

## Risk

This is a frontend-only label/icon change to the admin left-bar navigation. No routing, permissions, or backend changes are included, and the change can be rolled back by reverting the PR.

## Deploy Plan

- Deploy front


ClementAupiais:
Are we sure about that?

PR #30866, opened 2026-08-20
[front] refactor: use Sparkle ProgressBar in ConsumptionProgressBar

avervaet:
## Description

Uniformizes progress/loading bars across the app now that Sparkle has a `ProgressBar` component: `ConsumptionProgressBar` / `ConsumptionProgressBarWithNumbers` (used on the Developers > Credits Usage and Self-Improving Skills pages) now render via `@dust-tt/sparkle`'s `ProgressBar` instead of hand-rolled divs with inline `style={{ width }}`.

Color-coding (warning color past 80% consumed) is preserved using `ProgressBar`'s segmented `values` API: a colored "consumed" segment plus a `bg-transparent` "remaining" segment so the track color still shows through.

Part of a small series doing the same migration for the other non-Sparkle progress bars found in the app (`TrialMessageUsage`, `CreditUsageCard`, `FreePlanSeatsSection`, `PokeUsageTab`).

## Tests

Manual: seeded local `credits` rows (free/committed/payg, partially consumed) for a local workspace and verified the bars render correctly on `/w/{wId}/developers/credits-usage`, including the >80% warning color threshold. Also ran `npx tsgo --noEmit` and lint via the pre-commit hook.

## Risk

Low — presentational-only change, same props/behavior for callers of `ConsumptionProgressBar`/`ConsumptionProgressBarWithNumbers`. 

## Deploy Plan

- Deploy front

PR #30867, opened 2026-08-20
[front] refactor: use Sparkle ProgressBar in TrialMessageUsage

avervaet:
## Description

Part of a small series migrating the app's remaining non-Sparkle progress bars to `@dust-tt/sparkle`'s `ProgressBar` component (see #30866). `TrialMessageUsage` (sidebar, agent builder sidekick/preview) now renders its usage bar via `ProgressBar` instead of a hand-rolled div with inline `style={{ width }}`.

Color-coding (warning color at/above 90% consumed) is preserved using `ProgressBar`'s segmented `values` API: a colored "consumed" segment plus a `bg-transparent` "remaining" segment so the track color still shows through. A follow-up commit restores `w-full` on the bar, which had collapsed to zero width without an explicit width class.

## Tests

Manual: used `scripts/local/toggle_trial_message_usage.sh` to put a local workspace on a trialing subscription near its message limit, and verified the bar renders full-width with the correct color in the sidebar and agent builder sidekick/preview panels, including the critical (>=90%) warning color. Also updated `TrialMessageUsage.test.tsx`'s `@dust-tt/sparkle` mock to include `ProgressBar` (it was missing after the refactor, breaking CI) and ran `npx tsgo --noEmit`.

## Risk

Low — presentational-only change, same rendering conditions and props for `TrialMessageUsage` callers.

## Deploy Plan

- Deploy front

### DIFF
--- a/front/components/app/TrialMessageUsage.test.tsx
+++ b/front/components/app/TrialMessageUsage.test.tsx
@@ -17,6 +17,17 @@ vi.mock("@dust-tt/sparkle", () => ({
   }) => <a href={href}>{children}</a>,
   cn: (...classes: Array<string | false | null | undefined>) =>
     classes.filter(Boolean).join(" "),
+  ProgressBar: ({
+    values,
+  }: {
+    values: Array<{ value: number; className?: string }>;
+  }) => (
+    <div>
+      {values.map((v, i) => (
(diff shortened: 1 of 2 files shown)


avervaet:

- #30867 ◀

adrsimon (approved):

PR #30872, opened 2026-08-20
Lower github worker activity concurrency to 8

flvndvd:
## Description

The github Temporal worker runs up to 16 concurrent activities on a single Node event loop, and the pods sit pinned at roughly one core regardless of their two core limit, which is the signature of a saturated JavaScript thread rather than a starved container. Profiling the code extraction path showed the pipeline itself is pretty efficient on an idle event loop, while the worker process extracts at much much much slower pace.

Halving the concurrency gives each activity roughly twice the loop time, which by the observed numbers should bring large repository extractions comfortably inside the activity timeout. I might had a few more replicas.

This is a mitigation rather than the full fix. The proper isolation is a dedicated task queue for code sync served by its own deployment at very low concurrency, so extractions run on a near-idle loop. That is left for a follow-up.

### DIFF
--- a/connectors/src/connectors/github/temporal/worker.ts
+++ b/connectors/src/connectors/github/temporal/worker.ts
@@ -25,7 +25,7 @@ export async function runGithubWorker() {
       ...activitiesSyncCode,
     },
     taskQueue: QUEUE_NAME,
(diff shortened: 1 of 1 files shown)


flvndvd:
PMRR.

PR #30881, opened 2026-08-20
fix: skip spreadsheet in pod sync

matteotrab:
## Description

This PR skips spreadsheet files during Pod sync because extraction currently produces no pages, while still incurring download and Tika processing costs that can cause large syncs to fail when pre-signed URLs expire.

## Tests

Manually.

## Risks

Low. Spreadsheet files were already not being processed correctly.

## Deploy Plan

Standard deployment.


matteotrab:

- #30881 ◀

PopDaph (approved):
Approved as quick temporary fix, thank you 🙏🏻

PR #30884, opened 2026-08-20
[front] fix: break down sub-agents in message consumption breakdown

aubin-tchoi:
## Description

Message consumption currently groups every cost coming from sub-agents (fetched recursively) into one `Sub-agents` row, with additional rows for the actual tool call (input output).

This PR changes the logic to instead display one row by sub agent that combines the actual cost of the sub convo "communication" with the main agent (input from the main agent and output of the sub agent).

The rationale is that this view then surfaces directly one bucket for each sub agent, which says "this sub agent costs this much", rather than showing two disjoint pieces of info (1 underevaluated value for each sub agent + 1 aggregated value that becomes hard to action because it's not tied to each sub agent).

Current live version

image

Proposal

image

## Tests

- Added/updated tests.
- Tested locally + front-edge.

## Risk

- Low.

## Deploy Plan

- Deploy front.


front/lib/api/assistant/agent_message_consumption_attribution/read.ts
```diff
@@ -34,15 +39,71 @@ export async function getAgentMessageConsumption(
...
+      const childFacts = subAgentFactsByConversationId.get(childConversationId);
+      return {
+        action,
+        agentConfigurationId: childFacts?.agentConfigurationId ?? null,
+        billedCredits: childFacts?.billedCredits ?? 0,
+      };
+    }
+  );
+
+  const subAgentBilledCredits = subAgents.reduce(
```
aubin-tchoi, line 97: computed to get the right total, the per-agent breakdown is sent in subAgents

front/lib/api/assistant/agent_message_consumption_attribution/read.ts
```diff
@@ -34,15 +39,71 @@ export async function getAgentMessageConsumption(
...
+      ...new Set(
+        directSubAgentRoots.map(
+          ({ childConversationId }) => childConversationId
+        )
+      ),
+    ],
+    { includeDeleted: true }
+  );
+
+  const { messages: childMessages } =
```
flvndvd: My understanding is that this new approach only goes one level down, right? Current `ConversationResource.sumSubAgentCostCreditsByMessageId` goes all the way down the chain of sub agents.

aubin-tchoi: updated to also include the descendants' contributions. can be quite hard to understand that part of the cost comes from the tool call of sub agents calling sub sub agents so we don't surface this complexity here

front/lib/api/assistant/agent_message_consumption_attribution/read.ts
```diff
@@ -34,15 +39,69 @@ export async function getAgentMessageConsumption(
     return null;
   }
 
-  const subAgentBilledCredits =
-    await ConversationResource.sumSubAgentCostCreditsByMessageId(auth, {
-      agentMessageId,
+  const directSubAgentRoots = removeNulls(
```
flvndvd, line 42: `nit` as it's quite complex, moving to a dedicated function would help.

flvndvd (approved):

PR #30885, opened 2026-08-20
Backfill agent_messages.totalCostCredits

tdraier:
Stacked on #30630. One-time backfill of `agent_messages.totalCostCredits` for messages finalized before the column existed. No read path depends on the column yet — this just makes the historical data consistent with what the finalize path now writes.

Per workspace (`runOnAllWorkspaces`, `--wId` to target one, dry-run by default):
1. **Roll up run_agent parents** — for each distinct run_agent parent, `total = own costCredits + recursive sub-agent cost sum` (reuses `sumSubAgentCostCreditsByMessageId`). Order-independent, so parents can be processed in any order. Matches the finalize formula exactly: the recursive sum of descendants' own costs equals finalize's sum of the direct sub-agents' already-rolled-up totals.
2. **Seed the rest** — batched keyset updates set `total = costCredits` for the remaining (leaf / non-run_agent) messages.

Both steps skip rows whose `totalCostCredits` is already set, so finalize-written values are never clobbered and the script is safe to re-run.

### Note on the recursive query
This reuses the recursive `sumSubAgentCostCreditsByMessageId` that was removed from the finalize hot path. That's intentional: the objection there was DB load under concurrent production finalizes. Here it runs once, throttled (bounded concurrency over the small set of run_agent parents), in a controlled batch job — and being order-independent it keeps the script simple and matches the already-tested read-path method.

### Follow-up (not in this PR)
Read-path swap in `messages.ts` (drop the per-render recursive rollup, read `totalCostCredits - costCredits`) — deferred until this backfill has run in all environments.

### DIFF
--- a/front/migrations/20260820_backfill_agent_message_total_cost_credits.ts
+++ b/front/migrations/20260820_backfill_agent_message_total_cost_credits.ts
@@ -1,0 +1,224 @@
+import { QueryTypes } from "sequelize";
+
+import { Authenticator } from "@app/lib/auth";
+import { ConversationResource } from "@app/lib/resources/conversation_resource";
+import { frontSequelize } from "@app/lib/resources/storage";
+import { concurrentExecutor } from "@app/lib/utils/async_utils";
+import type { Logger } from "@app/logger/logger";
+import { makeScript } from "@app/scripts/helpers";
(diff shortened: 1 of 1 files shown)


tdraier:

- #30630
    - #30885 ◀

PR #30889, opened 2026-08-20
[front] Remove the legacy ACL path and shadow-compare for spaces

tdraier:
## Description

Retire the legacy ACL path for **spaces**. Space permission checks now always resolve from `group_permissions` (`getAccessControlLists` → `getGrantedVerbs`); the `use_legacy_acls` fallback and the per-check shadow-compare *call* are removed for spaces.

The `use_legacy_acls` kill switch and `isLegacyAclsEnabled` **stay** — skills still use them (`skill_resource.hasSkillPermission`). The generic `lib/api/permissions/shadow` util and the `group_permissions_shadow` flag are untouched.

- `space_resource`: `hasSpacePermission` → `auth.hasPermission` directly; delete `legacyAcls` and `legacySpaceGroupGrants` (the last readers of `group_vaults`' `groupSpaceKind`).
- `auth`: remove only the space-specific shadow-compare *call*. The generic `shadowComparePermission` / `runShadowComparePermission` methods are **kept** (currently unused) for the remaining resource migrations onto `group_permissions`.
- Tests: drop the kill-switch mock and the legacy-fallback test; the `group_permissions` enforcement tests remain.

First of a 3-PR stack removing `group_vaults`, on top of #30823: **this → #30890 (source space groups from group_permissions) → collapse `GroupSpace*Resource` → drop table**.

## Tests

`tsgo --noEmit` clean on front and front-api. Tests green: `space_resource` (69), `skill_resource` (77 — confirms the skill legacy path is intact), `shadow` (4).

## Risk

Removes the shadow-compare safety net for spaces, so it relies on governance grants matching `group_vaults` in prod — validated by querying `group_vaults` vs `group_permissions` (the only unmatched rows are stale: provisioned-on-manual and scrub-workspace orphans). Skills and other not-yet-migrated resources are unaffected (the shadow machinery remains available to them). Rollback = revert the code (no data changes).

## Deploy Plan

Standard deploy. Recommended to let the shadow monitor stay quiet for a bit before building the rest of the stack on top.

PR #29266, opened 2026-07-21
Remove group join from space fetches

spolu:
## Description

- replace the `GroupResource.model` include in space fetches with a direct `GroupSpaceModel` include
- represent `SpaceResource.groups` as lightweight `SpaceGroupReference` records hydrated from `group_vaults.groupId`, `groupKind`, and relationship `kind`
- compute serialized group sIds with `GroupResource.modelIdToSId`
- use group references directly for openness, membership, and permission checks
- fetch full `GroupResource` instances on demand for mutations, member management, and serializers that need complete group data
- update nested resource hydration and affected call sites to use the same direct group-vault association
- return lightweight `PokeItemBase` results for data source searches instead of serializing their spaces and groups
- add regression coverage for the space query shape and workspace-less superuser Poke search

Space reads only need the group id, intrinsic group kind, and group-space relationship kind for openness, membership, and permission checks. The previous belongs-to-many include joined both `group_vaults` and `groups` on every base fetch. The preceding rollout populated and enforced `group_vaults.groupKind`, so these reads can now avoid the extra join while preserving full group loading where it is actually required.

## Tests

- `nvm use && NODE_ENV=test npm test -- lib/resources/space_resource.test.ts lib/api/spaces.test.ts lib/resources/mcp_server_view_resource.test.ts lib/resources/sandbox_function_resource.test.ts lib/resources/conversation_selected_space_resource.test.ts lib/api/assistant/conversation/permissions.test.ts lib/api/assistant/conversation/selected_spaces.test.ts lib/api/projects/conversations.test.ts lib/api/assistant/conversation/branches.test.ts lib/api/assistant/conversation/skill_permissions.test.ts lib/api/viz/publish_frame.test.ts` (206 tests)
- `nvm use && NODE_ENV=test npm test -- routes/poke/search.test.ts` from `front-api` (7 tests)

## Risk

- High. This changes the hot `ResourceWithSpace` fetch path and the group representation used by general `SpaceResource` permissioning.

Can be reverted at any time.

## Deploy Plan

- deploy `front`
- deploy `front-sse`

PR #30823, opened 2026-08-19
[front] Remove the space_editors group kind

tdraier:
## Description

Removes the now-unused `space_editors` group kind. Editor groups are `regular_auto` (created that way in #30818, backfilled in #30822) and editor identity comes from `group_permissions`, so the kind has no remaining purpose.

- Drops `space_editors` from `GROUP_KINDS` and `GroupKindCodec`.
- Deletes the dead usages: the `space_editors` branch in `GroupResource.getAccessControlLists`, `isSpaceEditor()`, the redundant clauses in the member add/remove allow-lists, `leaveGroup`, list/delete filters, and the `isEditor` chip case.
- Simplifies `GroupSpaceEditorResource` asserts/filter to `regular_auto | provisioned`.
- Drops the transitional `MIGRATING_SPACE_EDITOR_GROUP_KINDS` and reverts `findRegularAutoGroupForGrant` to `regular_auto` only.
- Updates the 2024 rename migration to resolve groups via `fetchManual*Group`.
- Updates tests/factory/seed to `regular_auto` and role-based group resolution.

Top of a 3-PR stack: #30818 → #30822 → **this**.

## Tests

Tests updated (assert messages, editor/member resolution via `fetchManual*Group` and `editorGroupModelId`) and passing; `tsgo --noEmit` clean on front and front-api; no `space_editors` references remain (aside from the backfill script's SQL in #30822).

## Risk

Removing `space_editors` from `GroupKindCodec` means any lingering `space_editors` row would fail to parse. This PR must therefore merge **only after** the #30822 backfill has run in prod and EU. Kept as a **draft** until then. `groups.kind` is a plain string column (no Postgres enum), so no schema migration is needed.

## Deploy Plan

1. Confirm #30818 is deployed and the #30822 backfill has completed in **prod** and **EU** (zero remaining `space_editors` rows).
2. Mark this PR ready and merge.

PR #30890, opened 2026-08-20
[front] Source a space's associated groups from group_permissions

tdraier:
## Description

Source a space's associated groups from `group_permissions` instead of the `group_vaults` join, so nothing outside the write machinery reads `group_vaults`. **`group_vaults` writes are kept** until the drop-table PR.

- New scoped association `SpaceModel.hasMany(GroupPermissionModel, { as: "spaceGrants", foreignKey: "resourceId", scope: { resourceType: "space" }, constraints: false })`. **No join to `groups`** — `SpaceGroupReference` is built straight from the grant row, so the `groups` join PR #29266 deliberately removed stays removed.
- `SpaceGroupReference` now carries **`grantType`** (not the group's kind) plus `isReader()`. Grant type alone can't identify a group's kind — an open *regular* space's member group holds a `reader` grant just like the global group — so anything that needs the space's own groups by kind goes through a targeted lookup: new `GroupPermissionResource.listRegularAutoGroupsForResource` → `SpaceResource.fetchRegularAutoGroups` (the space's member + editor groups) and `SpaceResource.fetchMembershipGroups` (member/editor, excluding the read-only viewer).
- Repointed the reads: the space-fetch includes (`baseFetch`, `dangerouslyFetchByModelIds`, `resource_with_space`), `listForGroups`, and the two `auth.ts` sandbox lookups. `isOpen()` = "has a reader grant"; `isMemberByGroupPredicate` uses `isOpen`/`isReader`; `search.ts`'s provisioned check → `managementMode === "group"`; `fetchManualMemberGroup`/`fetchManualGroupsMemberships`/suspend/restore, front-api `leave`, poke `details`, and `mention_suggestions` resolve the space's own groups via `fetchRegularAutoGroups`/`fetchMembershipGroups`.
- `delete` and `hardDeleteSpace` now delete **only the space's own regular_auto groups** (via `fetchRegularAutoGroups`), never the shared global/provisioned groups. During full workspace teardown the global/system groups are still removed by the workspace-wide `GroupModel.destroy`.
- Removed the dead `groupSpaceKind` field and the dead `groupSpaces` association + `SpaceModel` field.
- **Not touched (flip in the drop-table PR):** `fetchAssociatedGroups` and the `GroupSpace*Resource` mutation machinery keep reading `group_vaults` — they feed `writeGroupPermissions` with the *current* association set, so reading grants there would clobber in-flight changes.

Second of the `group_vaults` removal stack, on top of #30889.

## Tests

`tsgo --noEmit` clean on front and front-api. Green: front `space_resource` (70), `spaces`, `group_space_resource`, `mcp_server_view_resource`, `permissions`, `conversation` (incl. the open-pod tests), `mention_suggestions`, `search`, plus the conversation/projects/viz suites; front-api `leave`, `members`, `spaces/[spaceId]`, poke `details`, `project_metadata`, `project_tasks`, `files`, `sandbox-functions`. ~25 test files were repointed from `space.groups.filter(isRegularAuto)` to `fetchRegularAutoGroups`. Fixes the pre-existing CI failure: the open-pod tests opened a pod via `GroupSpaceViewerResource.makeNew`, which writes the `group_vaults` row but no grant, so the space no longer read as open once `isOpen` moved to grants — they now sync grants after (mirroring production `syncGroupPermissions`).

## Risk

Equivalence of `group_vaults` and `group_permissions` was validated in prod (no active space with a live `regular_auto` group missing its grant). Targeted reads are served by existing indices `group_permissions_ws_rtype_rid` (workspaceId, resourceType, resourceId) and `group_permissions_ws_group` (workspaceId, groupId). `group_vaults` writes are unchanged, so rollback is a code revert.

## Deploy Plan

Standard deploy after #30889. One **pre-deploy** migration: `CREATE INDEX CONCURRENTLY group_permissions_space_ws_rid` — a partial index on `(workspaceId, resourceId) WHERE resourceType = 'space'` for the space-load path (the hottest query after the cached auth path, per review). Pre-deploy + `CONCURRENTLY` means it builds online, before the new code serves traffic.

### DIFF
--- a/front/components/spaces/RestrictedAccessBody.tsx
+++ b/front/components/spaces/RestrictedAccessBody.tsx
@@ -103,7 +103,7 @@ export function RestrictedAccessBody({
         title: "Switch to members",
         message:
           "This switches from group-based access to manual member management. " +
-          "Your current group settings will be saved but no longer active.",
+          "Your current group selection will be removed; you'll need to re-select groups to switch back.",
         validateLabel: "Confirm",
         validateVariant: "primary",
       });
--- a/front/lib/resources/group_permission_resource.ts
+++ b/front/lib/resources/group_permission_resource.ts
@@ -314,6 +314,42 @@ export class GroupPermissionResource extends BaseResource<GroupPermissionModel> 
     return autoGroups[0] ?? null;
   }
 
+  // All regular_auto groups holding any instance grant on the resource, regardless of grant type. A
+  // space's auto-created groups (its manual member group, and for projects its editor group) are
+  // regular_auto; the workspace global group (a viewer) and provisioned groups are excluded by kind.
+  // This is how callers recover a space's own groups: grant type alone cannot identify them because
+  // an open regular space's member group holds a `reader` grant like the global group.
+  static async listRegularAutoGroupsForResource(
+    auth: Authenticator,
+    {
+      resourceType,
+      resourceId,
+      transaction,
+    }: {
+      resourceType: GroupPermissionResourceType;
+      resourceId: number;
+      transaction?: Transaction;
+    }
+  ): Promise<GroupResource[]> {
+    const grants = await GroupPermissionModel.findAll({
+      where: {
+        workspaceId: auth.getNonNullableWorkspace().id,
+        resourceType,
+        resourceId,
+      },
+      transaction,
+    });
+    const groupIds = [...new Set(grants.map((grant) => grant.groupId))];
+    if (groupIds.length === 0) {
+      return [];
+    }
+
+    return GroupResource.fetchByModelIds(auth, groupIds, {
+      groupKinds: ["regular_auto"],
+      transaction,
+    });
+  }
+
   // The regular_auto groups backing user-level grants, keyed by grant (see `grantKey`) — the
   // batched counterpart of `findRegularAutoGroupForGrant`.
   static async findRegularAutoGroupsForGrants(
--- a/front/lib/resources/space_resource.ts
+++ b/front/lib/resources/space_resource.ts
@@ -12,6 +12,7 @@ import { GroupSpaceMemberResource } from "@app/lib/resources/group_space_member_
 import { GroupSpaceViewerResource } from "@app/lib/resources/group_space_viewer_resource";
 import { ContentFragmentModel } from "@app/lib/resources/storage/models/content_fragment";
 import { GroupMembershipModel } from "@app/lib/resources/storage/models/group_memberships";
+import { GroupPermissionModel } from "@app/lib/resources/storage/models/group_permissions";
 import { GroupSpaceModel } from "@app/lib/resources/storage/models/group_spaces";
 import { ProjectMetadataModel } from "@app/lib/resources/storage/models/project_metadata";
 import { SandboxOwnerModel } from "@app/lib/resources/storage/models/sandbox";
@@ -25,9 +26,9 @@ import { UserResource } from "@app/lib/resources/user_resource";
 import { concurrentExecutor } from "@app/lib/utils/async_utils";
 import { withTransaction } from "@app/lib/utils/sql_utils";
 import tracer from "@app/logger/tracer";
-import type { GrantVerb } from "@app/types/group_permissions";
+import type { GrantType, GrantVerb } from "@app/types/group_permissions";
 import { SPACE_EDITOR_GRANT_TYPE } from "@app/types/group_permissions";
-import type { GroupKind, GroupType } from "@app/types/groups";
+import type { GroupType } from "@app/types/groups";
 import {
   GLOBAL_SPACE_NAME,
   PROJECT_EDITOR_GROUP_PREFIX,
@@ -43,7 +44,7 @@ import type { Result } from "@app/types/shared/result";
 import { Err, Ok } from "@app/types/shared/result";
 import { assertNever } from "@app/types/shared/utils/assert_never";
 import { removeNulls } from "@app/types/shared/utils/general";
-import type { GroupSpaceKind, SpaceKind, SpaceType } from "@app/types/space";
+import type { SpaceKind, SpaceType } from "@app/types/space";
 import assert from "assert";
 import type {
   Attributes,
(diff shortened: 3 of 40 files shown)


tdraier:

- #30818
    - #30822
        - #30823
            - #30889
                - #30890 ◀
                    - #30899
                        - #30947
                            - #30949

front/lib/resources/space_resource.ts
```diff
@@ -2183,7 +2216,10 @@ export class SpaceResource extends BaseResource<SpaceModel> {
   }
 
   isOpen() {
-    return this.groups.some((group) => group.isGlobal());
+    // A space is open when it has a reader (viewer) grant: the workspace global group is attached as
+    // a viewer on open spaces (restricted spaces have no reader grant). See
+    // `SpaceGroupReference.isReader`.
+    return this.groups.some((group) => group.isReader());
```
fabiencelier, line 2222: This seems a bit brittle to me that any reader group means that the group is open.
In the future we might give reader access to some restricted spaces
But OK for now

front/components/spaces/RestrictedAccessBody.tsx
```diff
@@ -103,7 +103,7 @@ export function RestrictedAccessBody({
         title: "Switch to members",
         message:
           "This switches from group-based access to manual member management. " +
-          "Your current group settings will be saved but no longer active.",
+          "Your current group selection will be removed; you'll need to re-select groups to switch back.",
```
fabiencelier, line 106: This will avoid bad surprises indeed 👍

front/lib/resources/group_permission_resource.ts
```diff
@@ -314,6 +314,42 @@ export class GroupPermissionResource extends BaseResource<GroupPermissionModel>
     return autoGroups[0] ?? null;
   }
 
+  // All regular_auto groups holding any instance grant on the resource, regardless of grant type. A
+  // space's auto-created groups (its manual member group, and for projects its editor group) are
+  // regular_auto; the workspace global group (a viewer) and provisioned groups are excluded by kind.
+  // This is how callers recover a space's own groups: grant type alone cannot identify them because
+  // an open regular space's member group holds a `reader` grant like the global group.
+  static async listRegularAutoGroupsForResource(
```
fabiencelier, line 322: maybe add a unit test on this one ?

fabiencelier (approved):
Ok on the design and implem, but don't know the perf implications of such change

tdraier:
> Ok on the design and implem, but don't know the perf implications of such change

The main difference is the table joined from vaults, when loading a space. The goal was to keep the same kind of query without re-introducing a join on groups and work with data available in group_pemissions (we don't have denormalized groupKind here, but the grants - actually more meaningful)

Before : 

```sql
SELECT "spaces"."createdAt", "spaces"."updatedAt", "spaces"."deletedAt", "spaces"."name", "spaces"."kind", "spaces"."managementMode", "spaces"."workspaceId", "spaces"."id", "groupSpaces"."createdAt" AS "groupSpaces.createdAt", "groupSpaces"."kind" AS "groupSpaces.kind", "groupSpaces"."groupKind" AS "groupSpaces.groupKind", "groupSpaces"."workspaceId" AS "groupSpaces.workspaceId", "groupSpaces"."id" AS "groupSpaces.id", "groupSpaces"."updatedAt" AS "groupSpaces.updatedAt", "groupSpaces"."vaultId" AS "groupSpaces.vaultId", "groupSpaces"."groupId" AS "groupSpaces.groupId"
FROM "vaults" AS "spaces"
LEFT OUTER JOIN "group_vaults" AS "groupSpaces"
  ON "spaces"."id" = "groupSpaces"."vaultId"
  AND "groupSpaces"."workspaceId" = 4525
WHERE "spaces"."id" IN (223268)
  AND "spaces"."workspaceId" = 4525
  AND "spaces"."deletedAt" IS NULL;
```

After : 

```sql
SELECT "spaces"."createdAt", "spaces"."updatedAt", "spaces"."deletedAt", "spaces"."name", "spaces"."kind", "spaces"."managementMode", "spaces"."workspaceId", "spaces"."id", "spaceGrants"."id" AS "spaceGrants.id", "spaceGrants"."createdAt" AS "spaceGrants.createdAt", "spaceGrants"."updatedAt" AS "spaceGrants.updatedAt", "spaceGrants"."groupId" AS "spaceGrants.groupId", "spaceGrants"."grantType" AS "spaceGrants.grantType", "spaceGrants"."resourceType" AS "spaceGrants.resourceType", "spaceGrants"."resourceId" AS "spaceGrants.resourceId", "spaceGrants"."workspaceId" AS "spaceGrants.workspaceId"
FROM "vaults" AS "spaces"
LEFT OUTER JOIN "group_permissions" AS "spaceGrants"
  ON "spaces"."id" = "spaceGrants"."resourceId"
  AND (
    ("spaceGrants"."workspaceId" = 4525 AND "spaceGrants"."resourceType" = 'space')
    AND "spaceGrants"."resourceType" = 'space'
  )
  WHERE "spaces"."id" IN (223268)
  AND "spaces"."workspaceId" = 4525
  AND "spaces"."deletedAt" IS NULL;
```

From the test I did on db prod the query is going through similar indices/scans and takes the same time (although group_permissions table can be bigger - but we can have a dedicated index on resourceType='space' if it's really needed)

summoning @id13 as sql god

front/lib/resources/space_resource.ts
tdraier: agree .. it was already brittle before, would be better to store this info in space kind.

tdraier:
[claude speaking here]

On the perf side, a few points to back this up:

- **The join is structurally equivalent** — both are a `LEFT OUTER JOIN` from `vaults` to the junction table on the space id + `workspaceId`; the new one just adds `resourceType = 'space'`. And it actually *removes* a join rather than adding one: the reference is grant-based, so we don't re-introduce the `groups` join that #29266 deliberately dropped.
- **It's a perfect index match.** The predicate is equality on `(workspaceId, resourceType, resourceId)`, which are exactly the leading columns of the existing `group_permissions_ws_rtype_rid` index. So each space's grant lookup is an index point-lookup returning a handful of rows; `group_permissions` being a larger, polymorphic table only affects index depth (log n), not the per-space cost — which is why prod shows the same scans/time.
- **On a dedicated `resourceType = 'space'` index:** agree it's "only if needed." The composite index already leads with `workspaceId, resourceType`, so it effectively segments by resource type; a partial index would be marginally smaller/hotter but is a reactive optimization, not a prerequisite.
- **One thing worth checking before calling it settled:** the query above is a single-space fetch (`id IN (223268)`). The hot path is `baseFetch` loading the whole space list (`id IN (…many…)`) on a large workspace — worth an `EXPLAIN (ANALYZE, BUFFERS)` there to confirm the planner keeps using the composite index and doesn't flip to a seq scan of `group_permissions` as it grows. If it ever does, that's exactly when the partial index earns its place.
- Minor: the generated `AND … resourceType = 'space') AND … resourceType = 'space'` is a harmless duplicate (scope + where), no action needed.

Net: I wouldn't block on perf — the partial index is the back-pocket lever, gated on an EXPLAIN of the multi-space load.

fabiencelier:
Ok, let's go them
Agreed the only big diff seems the potantial bigger size of the group_permissions table

id13:
I think partial index on kind = space is wanted since loading spaces is the most ran query (after caching auth path)

id13 (approved):

tdraier:
> I think partial index on kind = space is wanted since loading spaces is the most ran query (after caching auth path)

let's try, we'll see if it's used

PR #30892, opened 2026-08-20
feat: release automations page

adrsimon:
## Description

This PR releases the admin/automations page to the world.
It ungates the page, deletes the ff from the sdk, and removes the 90-day selector so that we for now only have fresh data.

## Tests

Can fetch automations page with no feature flag on.

## Risk

Low.

## Deploy Plan

Deploy front.


avervaet (approved):
LGTM let's get things rolling 🔥

flvndvd (approved):
LGTM, just to make sure that we are not gating endpoints behind the FF as well, if so, they should be removed.

PR #30893, opened 2026-08-20
Skills: correctly remove space from archived skill on space deletion

fabiencelier:
## Description

Currently deleted space are not removed from archived skills
This make skill unreachable from the skill resource as it checks for all the spaces

Fix the space deletion script + add a migration script to fix existing skills.

## Tests

added a test

## Risk

low

## Deploy Plan

deploy front


aubin-tchoi (approved):
LGTM

PR #30882, opened 2026-08-20
[front] enh: show "Add a follow-up" input bar placeholder while a message is generating

PopDaph:
## Description

The input bar placeholder "Get work done" was duplicated in 3 places (editor default, compact pill fallback, Pod tab. This PR centralizes it into `INPUT_BAR_DEFAULT_PLACEHOLDER` and swaps the placeholder to **"Add a follow-up"** while a message is generating in the conversation since a message submitted then is queued.

Next step: add an animation to display the placeholder. 

## Risk

Worst case, wrong placeholder copy or a stale placeholder. Safe to rollback.

## Deploy Plan

- [ ] Deploy front

PR #30895, opened 2026-08-20
[front] enh: type the follow-up input bar placeholder like sidebar titles

PopDaph:
## Description

Follow up of #30882.

When the input bar placeholder switches to **"Add a follow-up..."** (a message is generating, so a submitted message is queued), it now types in character by character — same feel and cadence (32ms/char) as the sidebar conversation title `TypingAnimation`. The placeholder is a TipTap decoration, so the animation is driven through the `placeholderRef` + empty-transaction mechanism from #30882 rather than the Sparkle component.

The animation only plays when switching **to** the follow-up placeholder (`animatePlaceholder` prop, passed as `willQueueMessage` from `InputBar`). Switching back to "Get work done" and blocked-state messages swap instantly.

Tests cover the typed transition (fake timers), the instant swap back, and that content/selection/editor instance survive placeholder changes.

## Risk

Worst case, a broken placeholder animation on the input bar. Safe to rollback.

## Deploy Plan

- [ ] Deploy front

### DIFF
--- a/front/components/assistant/conversation/input_bar/InputBar.tsx
+++ b/front/components/assistant/conversation/input_bar/InputBar.tsx
@@ -806,6 +806,7 @@ export const InputBar = React.memo(function InputBar({
             placeholder={
               willQueueMessage ? INPUT_BAR_QUEUE_PLACEHOLDER : placeholder
             }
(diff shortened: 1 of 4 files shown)


tdraier (approved):
nice

PR #30897, opened 2026-08-20
feat(marketing): add Radja Kahoul, Steven Ing and Eddy Montalvo to the team page

radjakahoul:
## Description

Adds three team members to `PEOPLE` in `marketing/components/home/content/shared/team.ts`, rendered on the About page team grid:

| Name | Title |
|---|---|
| Radja Kahoul | Product Designer |
| Steven Ing | Brand Designer |
| Eddy Montalvo | Partnership Lead |

Avatars are sourced from `ca.slack-edge.com`, the same way most existing entries in `PEOPLE` do it — no new assets are committed. Slack serves these as square 512x512 crops, so they drop straight into the existing avatar markup with no rendering changes: `about.tsx` is untouched.

`github` is left as `""` for all three, the existing pattern for the 15 members without a GitHub handle. `about.tsx` already guards on it, so no GitHub icon renders.

## Net diff

`team.ts` is the only file that differs from `main` (+21).

The branch history passes through committing the photos as PNGs under `/static/landing/people/`, and a corresponding `object-cover` fix on the avatar `` to stop non-square uploads from stretching. Both were dropped once the photos moved to Slack: the assets were unreferenced, and the crop guard became dead code with no non-square source left in `PEOPLE`. Those intermediate commits are left in place since the merge squashes them.

## Notes for review

- `"Brand Designer"` and `"Partnership Lead"` are new titles in this file; the existing designer titles are `Principal Designer` and `Product Designer`, and existing partnership titles are `Partnership Account Manager` and `Partnership Operations`.
- `.authors` is untouched — it maps GitHub handle to Dust email for commit authorship, and none of the three have a GitHub handle here.
- Slack CDN URLs are already the dominant pattern in this file, so this inherits whatever durability those have; it does not introduce a new kind of dependency.

## Testing

Verified against `npm run dev` in `marketing/`:

- `/home/about` returns 200 and compiles clean.
- All three Slack URLs return 200, `image/jpeg` or `image/png`, 512x512.
- The compiled page chunk contains all three entries and no `/static/landing/people` reference.


marketing/public/static/landing/people/eddy-montalvo.png
avervaet, line 1: Can't we use slack image link rather than adding them to our codebase ?

marketing/pages/home/about.tsx
avervaet, line 1: is this needed ?

radjakahoul: indeed, if we rely on slack instead of storing not needed - changing also the notion tutorial on this

avervaet (approved):
LGTM

radjakahoul:
Deploy preview is not the right scope for this, tested to check only

PR #30901, opened 2026-08-20
[front] refactor: return a decision from the premium model fair-use gate

id13:
## Description

When reaching 25 premium messages limit, downgrade to auto instead.
The UI part is in the subsequent PR.

Still shadowing 

## Tests

Tested locally

## Risk

Low
## Deploy Plan

Deploy front

PR #30902, opened 2026-08-20
[front] enh: downgrade premium messages to the Standard stream at the fair-use cap

id13:


id13:

- #30901
    - #30902 ◀
        - #30903

PR #30905, opened 2026-08-20
[core] fix: resolve BigQuery views reading tables that require a partition filter

matteotrab:
## Description

Fixes https://github.com/dust-tt/tasks/issues/10108

- Querying a BigQuery view failed with `Query is using tables that are not part of allowed tables`, naming a table the agent never referenced.
- The query plan of a view lists the physical tables it reads, so the underlying table comes back as forbidden.
- `check_if_all_forbidden_tables_are_part_of_allowed_views` clears those by dry-running `SELECT * FROM <allowed table>`, but BigQuery refuses to plan it when the table read requires a partition filter.
- Now: read the partitioning metadata of the forbidden tables, filter those dry-runs with `<partitioning column> IS NULL`, and report their failures instead of discarding them.
- Tables requiring no partition filter: query unchanged.

## Risk

- Low: only the fallback path resolving tables outside the allowed set is touched.
- Worst case is the current behaviour with a more precise error message.

## Deploy Plan
Standard deployment

### DIFF
--- a/core/src/databases/remote_databases/bigquery.rs
+++ b/core/src/databases/remote_databases/bigquery.rs
@@ -3,7 +3,7 @@ use tracing::info;
 
 use anyhow::{anyhow, Result};
 use async_trait::async_trait;
-use futures::future::try_join_all;
+use futures::future::{join_all, try_join_all};
 use gcp_bigquery_client::{
     error::{BQError, NestedResponseError, ResponseError},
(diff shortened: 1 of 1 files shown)


matteotrab:

- #30905 ◀

fontanierh (approved):
LGTM

PR #30907, opened 2026-08-20
[front] Add a workspace setting to disable conversation email and Slack

frankaloia:
## Summary
- Adds a workspace admin toggle (Settings & Governance) that turns off conversation unread **email and Slack**, plus **For you** recommendation emails. In-app Dust popups, inbox sound, and billing/admin emails stay on.
- Missing metadata keeps the setting enabled (opt-out). Profile notifications lock the email/Slack/For you controls when the admin setting is off, and refresh workspace auth when the settings modal opens so the lock does not require a full page reload.
- Enforced in the conversation-unread Novu workflow (email/Slack steps only) and the For you activation-email path.

## Testing
Screenshot 2026-08-20 at 4 18 13 PM
Screenshot 2026-08-20 at 4 18 03 PM

## Risk
Low

## Deploy
1. Deploy front

### DIFF
--- a/front/admin/audit_log_schemas/workspace.conversation_external_notifications_updated.json
+++ b/front/admin/audit_log_schemas/workspace.conversation_external_notifications_updated.json
@@ -1,0 +1,7 @@
+{
+  "action": "workspace.conversation_external_notifications_updated",
+  "targets": [{ "type": "workspace" }],
+  "metadata": {
+    "enabled": "string"
+  }
+}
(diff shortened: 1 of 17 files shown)


front/admin/audit_log_schemas/workspace.conversation_external_notifications_updated.json
```diff
@@ -0,0 +1,7 @@
+{
+  "action": "workspace.conversation_external_notifications_updated",
+  "targets": [{ "type": "workspace" }],
+  "metadata": {
+    "enabled": "string"
```
smb2268: the `emitAuditLog` function adds an `actor_type` metadata field so we need to include that in the schema

smb2268 (approved):
lgtm, one small fix

front/admin/audit_log_schemas/workspace.conversation_external_notifications_updated.json
frankaloia: Added `actor_type` to the schema metadata. Re-registered with WorkOS; it stayed at v1.

PR #30909, opened 2026-08-20
[front] fix: prevent partial access to Automations page to managers

aubin-tchoi:
## Description

Managers can open the Automations page, but the overview and expanded trigger credit breakdown still require admin access, leaving the page partially unavailable to them.

This aligns both endpoints with the page's manager-level access while continuing to reject regular users. It also labels each credit destination by its actual dimension (`tool`, `model`, or `skill`) instead of always calling it a model.

## Tests

- Updated API permission and trigger breakdown component tests.
- Tested locally.

## Risk

- Low. Manager access is additive; regular-user restrictions remain unchanged.

## Deploy Plan

- Deploy front.

### DIFF
--- a/front/components/workspace/analytics/automations/AutomationsTriggerBreakdown.tsx
+++ b/front/components/workspace/analytics/automations/AutomationsTriggerBreakdown.tsx
@@ -1,7 +1,9 @@
 import { useAutomationsTriggerBreakdown } from "@app/hooks/useAutomationsTriggerBreakdown";
 import type { ConsumptionPeriodSelection } from "@app/lib/analytics/consumption_period";
+import type { AutomationTriggerCreditDestination } from "@app/lib/api/analytics/automations/breakdown";
 import type { AutomationTriggerRow } from "@app/lib/api/analytics/automations/triggers";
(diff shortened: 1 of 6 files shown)


front/components/workspace/analytics/automations/AutomationsTriggerBreakdown.tsx
```diff
@@ -106,7 +125,7 @@ function CreditDestinationBlock({
 
   return (
     <StatBlock
-      label="What model is used"
+      label={creditDestinationLabel(creditDestination.dimension)}
```
aubin-tchoi, line 128: we didn't realize this because it's basically always the model because most triggers run on a single model so it represents 100% of the credits

front/components/workspace/analytics/automations/AutomationsTriggerBreakdown.tsx
```diff
@@ -1,7 +1,9 @@
 import { useAutomationsTriggerBreakdown } from "@app/hooks/useAutomationsTriggerBreakdown";
```
aubin-tchoi, line 1: drive-by fix

adrsimon (approved):

PR #30913, opened 2026-08-20
[front] feat: route activation poke through Temporal eligibility gates

frankaloia:
## Summary
Activation poke used to post the nudge itself, so it skipped the checks the daily job uses (membership, BYOK, cadence, already-activated, etc.).

Poke now only creates or reuses the Learning Space, then starts a one-off Temporal run of the same workflow as the daily scheduler. The nudge goes out through that workflow.

Always enforced (even with the poke override checkbox):

Pod is not archived
Owner still has an active workspace membership
User is not credit/seat blocked (or over the legacy spend cap)

Skipped only when “Override eligibility checks” is checked:
BYOK workspaces
Frequency cap / unanswered-nudge limit
Already-activated users
100-user cap for that run
Daily runs are unchanged: they still spread nudges across the workday. Poke runs send immediately. Scheduled workflow arguments are unchanged so in-flight daily runs don’t break on deploy.

## Testing
1. Triggered poke
2. Confirmed it started temporal workflow with required checks

## Risk
Low (adding more governance around nudge behavior)

## Deploy
1. Deploy front

### DIFF
--- a/front/lib/api/activation/nudge.ts
+++ b/front/lib/api/activation/nudge.ts
@@ -3,7 +3,8 @@ import {
   createConversation,
   postUserMessage,
 } from "@app/lib/api/assistant/conversation";
-import { Authenticator } from "@app/lib/auth";
+import { isNonCreditPricedUserSpendLimitReached } from "@app/lib/api/users/spend_limit";
+import { Authenticator, getFeatureFlags } from "@app/lib/auth";
 import { serializeMention } from "@app/lib/mentions/format";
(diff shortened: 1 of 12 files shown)


front/lib/api/activation/nudge.ts
```diff
@@ -89,37 +89,73 @@ async function isPodDead(
       workspace: auth.getNonNullableWorkspace(),
     });
 
-  return activeMembership === null;
```
id13, line 92: 😅

id13 (approved):

PR #30916, opened 2026-08-21
[front] enh: add theme selector to Poke

aubin-tchoi:
## Description

I did remember that poke had a light theme, but could not change back to it anywhere, turns out it's not possible (keeps the system theme by default).

This PR adds a theme selector to the top-right navbar.

image

## Tests

- Added a small test.

## Risk

- Low.

## Deploy Plan

- Deploy front.


avervaet (approved):
Really cool! My only question mark is that by default it displays 'system,' which makes it counterintuitive to identify it as a way of changing the theme. Maybe we could use something with two icons, like in the app dev tools bar?

aubin-tchoi:
good point, the system icon is indeed just a computer (wouldve been fine if we had only light and dark)
switching to a multibutton

PR #30917, opened 2026-08-21
Admin governance: remove skills editor group 1/4: stop usage

fabiencelier:
## Description

First of 4 PRs removing the `skill_editors` group kind. Ticket: https://github.com/dust-tt/tasks/issues/9472

Skill editors are now entirely backed by `group_permissions`: the `editor` grant on the skill, held by a per-skill `regular_auto` group (`grantToUser` / `revokeFromUser`), which is already what `listEditors` / `batchListEditors` read from. This PR stops creating and using the legacy `skill_editors` group. The Poke "Update Skill Editors" plugin stops requiring an editor group too, so it keeps applying to skills created after this ships.

Follow-ups:
 - **2/4** #30922: migration deleting the `skill_editors` groups
 - **3/4** #30923: removing the `skill_editors` kind from `GROUP_KINDS`
 - **4/4** #30925: dropping the `group_skills` table

## Tests

 - updated unit test accordingly
 - tested locally

## Risk

Once we stop creating the `skill_editors` group, the `use_legacy_acls` kill switch no longer helps: skills created after the deploy have no editor group to fall back to, so flipping it back would leave those skills without editors. Reverting therefore means a code revert, not a switch flip — and any skill created in between would still need its editor group backfilled.

Existing `skill_editors` groups are left untouched by this PR — they are only deleted in #30922.

## Deploy Plan

Deploy `front`.


front/lib/resources/skill/skill_resource.ts
```diff
@@ -4417,7 +4224,34 @@ export class SkillResource extends BaseResource<SkillConfigurationModel> {
...
 
-    // Delete editor groups associated with skills.
+    // Delete the editor grants and the regular_auto groups holding them: those groups exist only
+    // to carry a skill's grant, so they go with the skills.
+    const skills = await SkillConfigurationModel.findAll({
+      attributes: ["id"],
+      where: { workspaceId },
+    });
+    const grantGroups =
+      await GroupPermissionResource.findRegularAutoGroupsForGrants(auth, {
```
tdraier, line 4267: nit - I've added a listRegularAutoGroupsForResource in https://github.com/dust-tt/dust/pull/30890/ to find all associated regular_auto , not tied to a specific grant

tdraier (approved):
lgtm

PR #30918, opened 2026-08-21
chore: make the members table bulk selection reusable

adrsimon:
## Description

The cross-page checkbox selection built for the members usage table was hardcoded to members. `useMembersSelection` becomes `useTableRowsSelection` with a generic id descriptor, and the banner is split into a generic `TableSelectionBanner` that `MembersSelectionBanner` now wraps.
No behavior change on `/w/:wId/usage`. This lays the groundwork for bulk edit on the automations table, which comes in the PR stacked on top of this one.

Lays out the ground work for #30919 

## Tests

Manually on `/w/:wId/usage`: select rows on a page, select all across pages, clear, and run both batch actions (role and spend limit). Same wording, same counts, same requests as before.

## Risk

Low. Renames and an extraction, no logic moved. The members page is the only consumer, and a regression there would be visible immediately in the selection banner.

## Deploy Plan

Deploy front.

### DIFF
--- a/front/components/shared/TableSelectionBanner.tsx
+++ b/front/components/shared/TableSelectionBanner.tsx
@@ -1,0 +1,74 @@
+import { pluralize } from "@app/types/shared/utils/string_utils";
+import {
+  ContentMessageAction,
+  ContentMessageInline,
+  Hoverable,
+} from "@dust-tt/sparkle";
+import type { ReactNode } from "react";
+
+interface TableSelectionBannerProps {
+  selectedCount: number;
+  pageCount: number;
+  totalCount: number;
+  itemLabel: string;
+  isAllAcrossPagesSelected: boolean;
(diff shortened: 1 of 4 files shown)


front/components/shared/TableSelectionBanner.tsx
```diff
@@ -0,0 +1,71 @@
...
+  disabled?: boolean;
+  // ContentMessageAction elements, rendered after "Clear".
+  children: ReactNode;
+}
+
+export function TableSelectionBanner({
+  selectedCount,
+  pageCount,
+  totalCount,
+  itemLabels: [singular, plural],
```
aubin-tchoi: do you plan to add an irregular plural noun? otherwise passing a single label + pluralize would do the job

aubin-tchoi (approved):
LGTM

PR #30919, opened 2026-08-21
feat: bulk credit pool update for automations

adrsimon:
## Description

Moving automations between the workspace and member credit pools meant opening them one at a time. The analytics automations table gets a checkbox column, select all across pages, and a batch pool update. This is behind `trigger_pool_choice` FF, and gated on `use_workspace_pool` user permission, without both, the selection column does not render.

Capture d’écran 2026-08-21 à 13 54 58

## Tests

Functional tests on the endpoint: role gating, feature flag gating, the happy path, and the skip path when the workspace pool is not granted. 

## Risk

Low. Automations the caller may not move are skipped rather than failed, and the table refetches after the batch, so the displayed pools come from the server.

## Deploy Plan

Deploy front.


front/lib/swr/agent_triggers.ts
```diff
@@ -517,3 +520,81 @@ export function useTriggerEstimation({
     mutateEstimation: mutate,
   };
 }
+
+async function postBulkTriggerUpdate({
```
aubin-tchoi: should be inlined, it does not do anything specific to bulk triggers

front/lib/swr/agent_triggers.ts
```diff
@@ -517,3 +520,81 @@ export function useTriggerEstimation({
...
+    body: JSON.stringify(body),
+  });
+  if (!response.ok) {
+    const errorData = await getErrorFromResponse(response);
+    return { error: errorData.message };
+  }
+  return response.json();
+}
+
+function bulkOutcomeDescription(
```
aubin-tchoi: nit: can be inlined

front/lib/resources/trigger_resource.ts
```diff
@@ -67,6 +69,39 @@ export async function resolveTriggerSpaceId(
 
 export class TriggerExecutionModeForbiddenError extends Error {}
 
+const BULK_TRIGGER_AUDIT_CONCURRENCY = 10;
+
+function emitBulkTriggerAuditLogEvents(
```
aubin-tchoi, line 74: why not one single bulk event OOC?

front/lib/api/triggers/bulk_selection.ts
```diff
@@ -0,0 +1,74 @@
...
+import { TriggerResource } from "@app/lib/resources/trigger_resource";
+import type { Result } from "@app/types/shared/result";
+import { Err, Ok } from "@app/types/shared/result";
+import { z } from "zod";
+
+// A bulk action resolves and updates its whole selection in the request, so
+// it is capped instead of being run in the background.
+export const MAX_BULK_TRIGGERS = 1000;
+
+export class BulkTriggerSelectionTooLargeError extends Error {}
```
aubin-tchoi: not such a fan of using custom error types because the caller cannot know they should be looking for this type of error specifically, would rather use an error with an error code (DustError)

front/lib/api/triggers/bulk_selection.ts
```diff
@@ -0,0 +1,74 @@
...
+    excludeTriggerIds: z.array(z.string()).max(MAX_BULK_TRIGGERS),
+  }),
+]);
+
+export type BulkTriggerSelection = z.infer<typeof BulkTriggerSelectionSchema>;
+
+export async function resolveBulkTriggerSelection(
+  auth: Authenticator,
+  selection: BulkTriggerSelection
+): Promise<Result<TriggerResource[], Error>> {
```
aubin-tchoi: suggestion
```suggestion
): Promise<Result<TriggerResource[], BulkTriggerSelectionTooLargeError | Error>> {
```

front/lib/api/analytics/automations/triggers.ts
```diff
@@ -420,3 +420,32 @@ export async function fetchAutomationTriggers(
     medianCostPerRun,
   });
 }
+
+// Ranked the same way the paginated view ranks them, so a "select all across
+// pages" resolves to the ids the table would show on any page.
+export async function fetchAutomationTriggerIds(
```
aubin-tchoi, line 426: nit: could be inlined

front-api/routes/w/[wId]/triggers/bulk-execution-mode.ts
```diff
@@ -0,0 +1,52 @@
...
+  async (ctx): HandlerResult<BulkTriggerExecutionModeResponseBody> => {
+    const auth = ctx.get("auth");
+    const { selection, executionMode } = ctx.req.valid("json");
+
+    const triggers = await resolveBulkTriggerSelectionForRequest(
+      ctx,
+      auth,
+      selection
+    );
+    if (!Array.isArray(triggers)) {
```
Comment (aubin-tchoi) [inline front-api/routes/w/[wId]/triggers/bulk-execution-mode.ts]: shouldn't this be an `isErr`?

aubin-tchoi (approved):
LGTM

PR #30923, opened 2026-08-21
Admin governance: remove skills editor group 3/4: drop the kind

fabiencelier:
## Description

Third of 4 PRs removing the `skill_editors` group kind. Ticket: https://github.com/dust-tt/tasks/issues/9472

Stacked on #30922. Skill editors are per-user `editor` grants in `group_permissions`, the code no longer creates or reads `skill_editors` groups (#30917) and the rows have been deleted (#30922), so the kind itself is dead. This PR removes it from `GROUP_KINDS` and `GroupKindCodec` and deletes the now-dead usages:

 - drop `isSkillEditorGroupKind` and its callers — the group listings that excluded per-agent/per-skill groups now exclude `agent_editors` only
 - drop the kind from the Poke group listings, the group-kind labels and the api-key group-name helper
 - delete `20260112_migrate_skill_editor_groups` and `20260215_suspend_skill_editor_memberships_no_active_skill`, which only exist to manage the removed kind

It also stops the scripts from managing editor groups themselves, so nothing writes `group_skills` anymore.

Follow-up: **4/4** #30925: dropping the `group_skills` table, which linked skills to these groups.

## Risk

Removing a value from `GroupKinds` means any remaining `skill_editors` row would fail to decode, so this must only merge once the #30922 migration has run.

## Deploy Plan

Deploy `front`, after the #30922 migration has run in US and EU. No migration in this PR.


front/scripts/create_hard_coded_suggested_skills.ts
```diff
@@ -277,26 +274,6 @@ async function createSuggestedSkills(
...
-
-          // Link the group to the skill
-          await GroupSkillModel.create(
-            {
-              groupId: editorGroup.id,
-              skillConfigurationId: createdSkill.id,
-              workspaceId: workspace.id,
-            },
-            { transaction }
-          );
```
tdraier, line 298: shouldn't we create the group + permission ?

tdraier (approved):

front/scripts/create_hard_coded_suggested_skills.ts
fabiencelier: there is no editor group anymore
here we use  to create an empty group so it does no make sense anymore
These are just suggestions, the actual editors are added when the suggestion is accepted later

PR #30927, opened 2026-08-21
[front] refactor: extract reusable consumption analytics UI components

aubin-tchoi:
## Description

The end goal of this PR, the one before and a few PRs that will follow is to give a way for us to both inspect what the users will see in the analytics page and to use them for monitoring workspace health and adoption patterns.

The consumption analytics UI needs a reusable boundary before it can be shown outside the customer workspace page.

This PR separates data fetching from the pure visual part for the overview, summary, filters, charts, and attribution table. The workspace wrappers just use the hooks currently used to then pass down the data to these new components, and the poke components will use different hooks but the same visual components.

No actual behavior change expected.
Nothing to see for now.

Next step will be to add the UI in poke.

## Tests

- Tested locally.

## Risk

- Low.

## Deploy Plan

- Deploy front.

### DIFF
--- a/front/components/pages/workspace/AnalyticsConsumptionPage.tsx
+++ b/front/components/pages/workspace/AnalyticsConsumptionPage.tsx
@@ -62,49 +62,49 @@ const WORKSPACE_CONSUMPTION_COMPONENTS: AnalyticsConsumptionComponents = {
   UsageFilterPanel,
 };
 
-function ChartFallback() {
-  return (
-    <div aria-hidden="true" className="flex flex-col gap-4">
-      <div className="flex items-center justify-between">
-        <LoadingBlock className="h-5 w-28" />
-        <div className="rounded-2xl border border-border-dark bg-background p-1">
-          <LoadingBlock className="h-8 w-36 rounded-xl" />
-        </div>
-      </div>
-      <div className="rounded-lg border border-border bg-background p-4">
-        <div className="border-b border-border pb-3">
-          <LoadingBlock className="h-6 w-36" />
-        </div>
-        <LoadingBlock
-          className="mt-3 w-full"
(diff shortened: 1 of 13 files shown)


front/components/pages/workspace/AnalyticsConsumptionPage.tsx
```diff
@@ -65,22 +88,32 @@ function ChartFallback() {
   );
 }
 
+function PokeChartFallback() {
```
avervaet: Nit: PokeChartFallback largely duplicates ChartFallback above. Could we extract the shared chart-body skeleton or parameterize the layout difference so these placeholders do not drift over time?

Maybe this is part of the next pr tho

front/hooks/useConsumptionTop.ts
```diff
@@ -56,10 +56,24 @@ type ConsumptionTopResponse =
   | GetConsumptionTopSourcesResponse
   | GetConsumptionTopApiKeysResponse;
 
+export interface UseConsumptionTopParams {
+  workspaceId: string;
```
avervaet, line 60: nice

avervaet (approved):

front/components/pages/workspace/AnalyticsConsumptionPage.tsx
aubin-tchoi: I messed up a bit, there are indeed changes that are then removed in the next PR, let me fix

front/components/workspace/analytics/UsageFilterPanel.tsx
```diff
@@ -110,6 +143,77 @@ export function UsageFilterPanel({
     [workspaceGroups]
   );
 
+  return {
```
aubin-tchoi, line 146: rly horrible but will simplify separately

PR #30931, opened 2026-08-21
Deprecate agent loop cancellation patches

flvndvd:
## Description

The tool-activity and model-activity cancellation migrations are deployed. Their legacy branches can now be removed while retaining Temporal replay compatibility through `deprecatePatch`.

Follow-up to #30530 and #30921.

Confirmed on Temporal side, no more running workflows that don't have both patches. Worth flagging that once this is out, we won't be able to restart any of the old workflows, which is fine as we never do it.


aubin-tchoi (approved):
LGTM

PR #30932, opened 2026-08-21
[connectors] Downgrade ERR_STREAM_UNABLE_TO_PIPE unhandled rejections to warn

flvndvd:
## Description

See thread [here](https://dust4ai.slack.com/archives/C050SM8NSPK/p1787303916318309).

The GCS client's streaming transport, `teeny-request`, pipes an HTTP response into its request stream only once the response arrives. When our side has destroyed that stream in the meantime, which is what a failed tarball extraction attempt's teardown legitimately does to in-flight reads and uploads, the deferred pipe throws inside the library's own promise continuation where no caller can catch it, and surfaces as an unhandled rejection logged as a panic.

The error code has a single meaning, piping into an already-destroyed stream, so matching it narrowly cannot mask a genuine failure.

### DIFF
--- a/connectors/src/types/shared/utils/global_error_handler.ts
+++ b/connectors/src/types/shared/utils/global_error_handler.ts
@@ -22,6 +22,18 @@ export function setupGlobalErrorHandler(logger: LoggerInterface) {
       return;
     }
 
+    // The GCS client (teeny-request) connects a response to its stream only once the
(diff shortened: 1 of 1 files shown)


flvndvd:
PMRR.

PopDaph (commented):
🙏🏻

PR #30934, opened 2026-08-21
[front] chore: move charts out of Poke analytics tab

aubin-tchoi:
## Description

The new analytics view is being added to Poke, it will clash in naming with the existing analytics tab.
This tab contains 2 charts: one with the activity and a data source treemap.

This PR moves the 30-day activity to the top of the Usage tab, removes the datasource-retrieval treemap and removes the tab.

## Tests

- Tested locally.

## Risk

- Low. Poke only.

## Deploy Plan

- Deploy front.


front/components/poke/credits/PokeUsageTab.tsx
```diff
@@ -357,6 +358,7 @@ export function PokeUsageTab({
 
   return (
     <div className="flex flex-col gap-4">
+      <PokeWorkspaceUsageChart workspaceId={owner.sId} period={30} />
```
avervaet, line 367: ?: The activity chart is now only rendered inside PokeUsageTab, but WorkspacePage only exposes that tab when hasMetronomeUsage is true. 

This means workspaces without a Metronome customer/contract lose the existing activity and active-users analytics entirely. Is this ok/intended

avervaet (approved):
Not sure why we are removing WorkspaceDatasourceRetrievalTreemapPluginChart completely. But code is coherent with PR description. LGTM

front/components/poke/credits/PokeUsageTab.tsx
aubin-tchoi: ah

aubin-tchoi: good catch, this is an issue

aubin-tchoi: removed to always show the tab (hiding the non-relevant part if there's no billing account)

aubin-tchoi:
did not really find a home for `WorkspaceDatasourceRetrievalTreemapPluginChart`, and it does not serve an actual purpose 🤷🏻

PR #30943, opened 2026-08-21
[front] feat: add personal mode for consumption analytics components

aubin-tchoi:
## Description

The goal of this PR is to make the existing consumption analytics UI reusable for the personal view we started to introduce in #30942.

This PR adds a "personal" mode to the shared analytics hooks + components. In that mode, the hook target the right endpoint and member and group dimensions are omitted.

Also adds a `disabled` prop, which we'll need to disable the hook when not needed (when the popover is not open for instance).

Next PR will add the popover.

## Tests

- Added tests.
- Tested locally.

## Risk

- Low.

## Deploy Plan

- Deploy front.


front/components/workspace/analytics/UsageFilterPanel.tsx
avervaet, line 1: We are still displaying the Members and Groups filtering dimensions in the left column of the filtering panel.

Queries are properly filtered, so they will always display "No results."

We might want to remove those filtering panels, as they are useless from a personal user's point of view.

front/components/workspace/analytics/consumption/ConsumptionAttributionTable.tsx
avervaet, line 1: Could we normalize dimension when personal is true?  

If it is `user`/`group`, those tabs are hidden but the component still requests unmounted `/me/.../top-users` or `/top-groups` endpoints, resulting in a 404 and no active tab.

avervaet (approved):

front/components/workspace/analytics/UsageFilterPanel.tsx
aubin-tchoi: good catch, fixing 🙏🏻

front/components/workspace/analytics/consumption/ConsumptionAttributionTable.tsx
aubin-tchoi: added small re-usable functions (since we'll need them for agents)

PR #30945, opened 2026-08-21
[front] enh: add Poke links in new Poke analytics

aubin-tchoi:
## Description

This PR adds links to relevant poke pages (agents, groups, skills) in the new Poke analytics page.

We actually already have the links, so this PR is front-end only. 

We can't really add tools here; analytics stores a server display name rather than a unique MCP view ID.

## Tests

- Checked locally.

## Risk

- Low.

## Deploy Plan

- Deploy front.


avervaet (approved):
LGTM

adrsimon (approved):

PR #30946, opened 2026-08-21
[connectors] Slack: skip auto-read when workspace is unavailable

PopDaph:
## Description

Follow-up to #30755. [`slackWebhookEventWorkflow` still fails](https://cloud.temporal.io/namespaces/dust-prod.gmnlm/workflows?query=ExecutionStatus%3D%22Failed%22) on `channel_created` events when the workspace `/exists` probe errors, e.g. `Cannot auto-read channel: workspace is unavailable (Your current plan does not allow API access. Please upgrade your plan.)`. Retrying can't help: there is nothing to auto-read for an unavailable workspace.

Fix: `autoReadChannel` now logs and returns `Ok(false)` when the probe fails, instead of `Err`. Same handling as `channel_message` events in `webhook_activities.ts`, and it also stops the slack_bot webhook from 500ing (and Slack from retrying) on the same condition.

## Tests

Added a test: connector with a matching auto-read pattern, `/exists` fails with the plan-limit error, returns `Ok(false)` without launching the join workflow.

## Risk

Worst case, a legitimate auto-read is silently skipped instead of surfacing an error. Safe to rollback.

## Deploy Plan

- [ ] Deploy connectors

### DIFF
--- a/connectors/src/connectors/slack/auto_read_channel.test.ts
+++ b/connectors/src/connectors/slack/auto_read_channel.test.ts
@@ -1,10 +1,22 @@
 import mainLogger from "@connectors/logger/logger";
 import { ConnectorResource } from "@connectors/resources/connector_resource";
+import { DustAPI, Err } from "@dust-tt/client";
 import { describe, expect, it, vi } from "vitest";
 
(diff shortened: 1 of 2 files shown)


aubin-tchoi (approved):
LGTM

PR #30962, opened 2026-08-21
[front] feat: admin panel to set per-group workflow alert threshold

avervaet:
Adds a PUT /api/w/:wId/groups/:groupId/workflow_alert_threshold
endpoint and a new "Smooth shutdown threshold" column in the groups
usage table, mirroring the existing per-group spend limit pattern
(inline InputWithSave cell, empty = disabled). Unlike the spend
limit, this threshold has no Metronome billing enforcement — it only
persists admin intent for the agent loop to read.

Needs a manual follow-up before merge: run
`npx tsx front/admin/register_audit_log_schemas.ts --execute --changed`
to register the new group.workflow_alert_threshold_updated schema
with WorkOS and commit the regenerated schema_versions.json —
audit_log_schemas.test.ts will fail until that's done.

PR #30961, opened 2026-08-21
[front] feat: add per-group workflow alert threshold column

avervaet:
Adds workflowAlertThresholdAwuCredits on groups, mirroring the
poolCapAwuCredits pattern used for group spend limits. Null disables
the smooth shutdown flow for a group; when a user belongs to several
groups with a threshold set, the highest value applies.

PR #30963, opened 2026-08-21
[front] feat: backend for the smooth shutdown flow

avervaet:
Adds a new "smooth shutdown" agent loop signal, reusing the existing
graceful-stop cooperative mechanism (finish the current step, then
exit at the next step boundary) but with its own finalize activity:
before publishing the terminal event, it generates a short one-shot
summary of progress so far (single lightweight LLM call, no tools,
mirroring the conversation-title-generation pattern) and persists it
as one more text step on the message, so it reads as the agent's
final reply. Falls back to a plain stop with no summary if the
summary call fails.

Exposes the trigger as a new "smooth_shutdown" action on the existing
POST /cancel endpoint, alongside cancel/gracefully_stop/interrupt.

### DIFF
--- a/front-api/routes/w/[wId]/assistant/conversations/[cId]/cancel.ts
+++ b/front-api/routes/w/[wId]/assistant/conversations/[cId]/cancel.ts
@@ -1,4 +1,7 @@
-import { gracefullyStopAgentLoop } from "@app/lib/api/assistant/pubsub";
+import {
+  gracefullyStopAgentLoop,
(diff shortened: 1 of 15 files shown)


avervaet:

- #30961
    - #30962
        - #30963 ◀
            - #30964

avervaet:
Context: https://dust4ai.slack.com/archives/C0BJ3T5SA0L/p1787300759609879

PR #30964, opened 2026-08-21
[front] feat: per-step workflow alert threshold check and popup

avervaet:
Adds a cheap per-step check (no conversation fetch) alongside the
existing hard credit-cap gate: does the user's current spend cross
their workflow alert threshold (max across their groups, from the
earlier per-group setting)? Unlike the credit-cap gate this never
stops the loop by itself — it publishes a one-time notification
event once crossed (loading the conversation only on that single
occasion), which the client surfaces as a dialog offering to
continue or trigger the smooth shutdown flow added previously.


avervaet:

- #30961
    - #30962
        - #30963
            - #30964 ◀

avervaet:
Context: https://dust4ai.slack.com/archives/C0BJ3T5SA0L/p1787300759609879

PR #30965, opened 2026-08-21
feat(llm): run trigger and wake-up runs on OpenAI flex processing

Nils-Fedrigo:
## Description

Trigger and wake-up agent runs have nobody waiting on the answer in real time, so they are the runs where we can trade latency for cost. This routes them to OpenAI's flex processing tier, which bills at Batch-API token prices (50% off standard).

- New `openai_flex_processing` feature flag (`dust_only`) gates the whole behaviour.
- `withFlexProcessing` sets `serviceTier: "flex"` on the input config for `triggered`, `triggered_programmatic` and `wakeup` origins only, in flagged workspaces. It is applied in `BaseTransition` next to `withConciseOpenAIReasoningSummary`.
- `service_tier: "flex"` is only put on the wire for endpoints that opt in via `supportsFlexProcessing` (the GPT-5.x Responses streams).
- Flex is best-effort, so the first attempt runs with `maxRetries: 0` and a 30s timeout. If it fails **before** emitting any event, the request is replayed on `service_tier: "auto"`; once the stream has started, failures are left to the usual caller retry. Fallbacks are logged and counted (`llm_flex_fallback.count`, tagged with model/host/region/reason).
- The tier OpenAI reports on `response.completed` is propagated through the token-usage event, persisted on the run, tagged on token-usage metrics (`service_tier:*`), and applied as `FLEX_DISCOUNT_FACTOR` in `computeTokensCostForUsageInMicroUsd` so cost reflects what is actually billed rather than what we asked for.

Note the discount is driven by the tier the provider reports, not by the tier we requested, so a silent downgrade to standard still costs the standard price in our accounting.

## Tests

Unit tests added:

- `front/lib/api/llm/flex_processing.test.ts` — origin and feature-flag gating.
- `front/lib/model_constructors/stream/clients/openai_responses.test.ts` — flex payload construction, fallback before first event, no fallback once the stream has started, no fallback for non-flex requests.
- `front/lib/model_constructors/sdk/openai_responses/converters/output/utils.test.ts` — service tier propagated from `response.completed` into the token-usage event.
- `front/lib/api/assistant/token_pricing/index.test.ts` — flex discount applied to cost.

Manual verification worth doing after deploy: enable the flag on a Dust workspace, fire a trigger-driven run, and check the run row carries `serviceTier: "flex"` plus the halved cost.

## Risk

Low: gated behind a `dust_only` feature flag and limited to trigger/wake-up runs, so interactive traffic is untouched. The failure mode we accept is added latency on flagged async runs (up to the 30s flex timeout before the standard replay). `serviceTier` is optional everywhere it was threaded through, so existing providers and persisted runs are unaffected.


front/lib/model_constructors/types/input/configuration.ts
```diff
@@ -38,6 +38,17 @@ export const toolSpecificationSchema = z.object({
...
+// In practice we use only flex and auto but others are included for completeness.
+// Only used by OpenAI
+const serviceTierSchema = z.enum([
+  "auto",
+  "default",
+  "flex",
+  "scale",
+  "priority",
+]);
+export type ServiceTier = z.infer<typeof serviceTierSchema>;
```
fabiencelier: Isn't it over indexing on OpenAI ?
Should we just have standard and flex and let each model adapter decide how it translates ?

front/lib/api/llm/flex_processing.ts
```diff
@@ -0,0 +1,30 @@
...
+  userMessageOrigin: UserMessageOrigin | undefined
+): InputConfig {
+  const isLatencyTolerant =
+    userMessageOrigin !== undefined &&
+    LATENCY_TOLERANT_ORIGINS.has(userMessageOrigin);
+
+  if (!featureFlags.includes(OPENAI_FLEX_PROCESSING) || !isLatencyTolerant) {
+    return config;
+  }
+
```
fabiencelier, line 28: should we check here if the underlying config supports flex ? to avoid passing it to. model which does not support it entrirely ?

fabiencelier (approved):
Looks good
IMHO we should not overindex on OpenAI types and only have flex/standard

Not in this PR necessarily, but I think we should also keep the info of the service tier used in the run usage so we explain the costs

front/lib/api/llm/flex_processing.ts
Nils-Fedrigo: I check later in the chain, I think it's ok

front/lib/model_constructors/stream/clients/openai_responses.ts
```diff
@@ -67,21 +73,85 @@ export abstract class OpenAIResponsesStream extends WithOpenAIResponsesInputConv
...
   }
 
+  private logFlexFallback(err: unknown): void {
+    const reason =
+      err instanceof APIConnectionTimeoutError ? "timeout" : "error";
+    const { model, host, region } = this.metadata();
+    const tags = [
+      `model_id:${model}`,
+      `host:${host}`,
+      `region:${region}`,
```
flvndvd, line 144: `host` is a reserved field name on DD and in our infra. If you look at other events, they are often being re-mapped. Mind updating 🙏 ?

Nils-Fedrigo: updating

PR #30966, opened 2026-08-21
refactor(front-api): move trigger endpoints out of agent_configurations

adrsimon:
## Description

The trigger endpoints were nested under `/api/w/:wId/assistant/agent_configurations/:aId/triggers`. Everything now lives under `/api/w/:wId/triggers`:
- collection (GET/POST/PATCH/DELETE) takes `aId` as a query param
- `:tId/status`, `:tId/execution_mode`, `:tId/webhook_requests` keep their handlers unchanged
- `bulk-execution-mode` stays here, already been created in the right place.

The `:tId` subroutes dropped the `trigger.agentConfigurationId !== aId` check. It was only re-checking what `TriggerResource.fetchById` already scopes, so carrying `aId` along just to assert it added nothing.

The old paths stay live as a 307 redirect that rewrites the path and moves `aId` into the query string. That is deleted in the follow-up (#30967).

## Tests

Moved the existing route tests with the routes and updated their URLs: 21 collection tests, 17 status, 7 execution_mode, all passing.

## Risk

Low. Every caller is in `front`, which ships with the API.

The window that matters is browser tabs already running the previous bundle. They keep working through the redirect: 307 preserves method and body.

Rollback is a revert; no data or schema involved.

## Deploy Plan

Deploy front-api and front together. Once that has rolled out and stale tabs have turned over, merge the follow-up that deletes the redirect.


aubin-tchoi (approved):
LGTM

PR #30969, opened 2026-08-21
[front] fix: keep tool validation details closed by default

aubin-tchoi:
## Description

Closes https://github.com/dust-tt/tasks/issues/10154
Closes https://github.com/dust-tt/tasks/issues/10153

Tool validation details could open automatically for the sandbox `add_egress_domain` tool, covering the approval card before the user chose to review them.

This removes the `detailsOpen` override, it was only used for this specific tool, and had been ported incorrectly from the behavior before the modal, where the details were expanded/collapsed inline in the card. The motivation then was to highlight the fact that for this specific tool the details were necessary to make a decision: we instead show the details inline.

Also fixes the position of the close button.

Current live version

image

Proposal

image

## Tests

- Tested locally.

## Risk

- Low.

## Deploy Plan

- Deploy front.

### DIFF
--- a/front/components/actions/blocked/ToolValidationCard.tsx
+++ b/front/components/actions/blocked/ToolValidationCard.tsx
@@ -117,7 +117,7 @@ function ToolValidationDetailsDialog({
         className="gap-4 p-5"
         preventAutoFocusOnClose={false}
       >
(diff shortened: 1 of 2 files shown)


ClementAupiais (approved):
lgtm

PR #30970, opened 2026-08-21
[front] feat: back the automations filter with consumption facets

adrsimon:
## Description

The agent and member lists in the `/automations` filter came from `useSearchMembers` and `useAgentConfigurations`, so options were unsorted and included people who never ran an automation.
They now come from the consumption facets endpoint, which already backs the `/analytics/consumption` filter, via a new `scope: "automations"` that restricts facet counts to trigger-originated documents (`exists: trigger_id`). Both tabs get consumption's sorting and the "No activity" label.

## Tests

Added unit tests for the scope filter, and the `scope` request validation on the facets route. Checked the panel by hand against a seeded local ES, including deleting a member's consumption docs to confirm they stay listed and greyed out.

Capture d’écran 2026-08-21 à 15 52 51

## Risk

Low, contained to the automations filter panel. `scope` defaults to `all` so the consumption page is unaffected. 
The type tab is unchanged: trigger kind isn't a consumption dimension, so selecting Schedule or Webhook doesn't narrow the availability counts on the other two tabs.

## Deploy Plan

Deploy front.

### DIFF
--- a/front/components/workspace/analytics/automations/AutomationsFilterPanel.tsx
+++ b/front/components/workspace/analytics/automations/AutomationsFilterPanel.tsx
@@ -8,14 +8,15 @@ import {
   AUTOMATIONS_FILTER_CATEGORIES,
   AUTOMATIONS_FILTER_CATEGORY_LABEL,
   automationsFilterSelectionCount,
+  toAutomationsScopeFilter,
 } from "@app/components/workspace/analytics/automationsFilter";
 import { FilterCategoryNav } from "@app/components/workspace/analytics/filterPanel/FilterCategoryNav";
 import { FilterFooter } from "@app/components/workspace/analytics/filterPanel/FilterFooter";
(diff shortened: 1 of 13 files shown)


front/components/workspace/analytics/automations/AutomationsFilterPanel.tsx
```diff
@@ -54,77 +55,75 @@ export function AutomationsFilterPanel({
...
+  const draftScopeFilter = useMemo(
+    () => toAutomationsScopeFilter(draftFilter),
+    [draftFilter]
+  );
+  const {
+    options: facetOptions,
+    isFacetsLoading,
+    isFacetsError,
+    isFacetsValidating,
+  } = useConsumptionFacets({
```
avervaet, line 83: won't we be loading all the facets only to display members and agents here. 

Maybe could do dimension scope request here to avoid running up to eight composite Elasticsearch queries and eagerly loads all facet catalogs too often.

avervaet (approved):
LGTM

PR #30971, opened 2026-08-21
[front] Poke: fix invisible panel borders

PopDaph:
## Description

Poke panel borders were invisible in both themes: `border-material-200` is a dead class (token defined nowhere), so panels fell back to the subtle default border on a same-color background.

- `PokeLayout`: page on `bg-background` (same canvas as the app) and a scoped override `--color-border: var(--color-border-form)`, so every default border in poke uses the visible form token (stone-300 / stone-700).
- Delete the dead `border-material-200` class (32 files).
- Table containers: white body + hairline under the header instead of grey `bg-muted-background` + `bg-primary-300` slab, matching the plugins panel.

## Risk

Poke-only styling. Safe to rollback.

### DIFF
--- a/front/components/poke/PokeConditionalDataTables.tsx
+++ b/front/components/poke/PokeConditionalDataTables.tsx
@@ -83,8 +83,8 @@ export function PokeDataTableConditionalFetch<T, M>({
   }
 
   return (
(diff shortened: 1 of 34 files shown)


aubin-tchoi (approved):
LGTM

PR #30972, opened 2026-08-21
feat(front): split the user menu Tools and Triggers entry

adrsimon:
## Description

The user menu had one "Tools and Triggers" dialog with two tabs. Triggers becomes its own "Automations" entry right under "Tools", and instead of the old bare list it renders the `/automations` table, filters and expandable credit breakdown, scoped to the member's own automations. A member can now see what each of their automations runs and costs, and still enable, disable or delete it from the same place. Tools keeps its table in a dedicated dialog.

Last of three, stacked on #30975 (endpoints) and #30974 (shared table).

## Tests

Manual: opened both dialogs, searched and filtered, expanded a row's breakdown, toggled an automation off and on, and deleted one on a global agent. Backend coverage is in #30975.

## Risk

The table depends on the two new `/me/automations` endpoints from #30975, so this cannot merge ahead of it. Members with no automations get the empty state; the credits column reflects the current billing cycle, and the dialog says so when consumption is unavailable.

`useUserTriggers` is deleted here since its only caller is gone. `/api/w/:wId/me/triggers` stays, marked deprecated, until deployed clients have cycled out.

## Deploy Plan

Deploy front. Merge after #30974 and #30975.


front/components/me/UserAutomationsTable.tsx
```diff
@@ -0,0 +1,417 @@
+import { AutomationsFilterPanel } from "@app/components/workspace/analytics/automations/AutomationsFilterPanel";
```
avervaet, line 1: i'm pretty sure this filter panel will need at least manager rights to render agent options which is not intended for a page available to all users;

front/components/me/UserAutomationsTable.tsx
```diff
@@ -0,0 +1,417 @@
...
+  );
+}
+
+function RunningCell({ row }: { row: TriggerRowData }) {
+  switch (row.status) {
+    case "enabled":
+    case "disabled":
+      return (
+        <SliderToggle
+          selected={row.status === "enabled"}
```
avervaet, line 87: Could we stop propagation for the status toggle here? `SliderToggle` forwards the click event, and the parent row uses `onClick` to expand the breakdown, so toggling an automation also expands/collapses its row. The details and action buttons already stop propagation.

avervaet (changes requested):
LGTM overall but the filter values are currently manager right gated let's fix this first.

front/components/me/UserAutomationsTable.tsx
adrsimon: good catch, my bad!

front/components/me/UserAutomationsTable.tsx
```diff
@@ -0,0 +1,422 @@
...
+          setExpandedRowId((current) =>
+            current === trigger.triggerId ? null : trigger.triggerId
+          ),
+      })),
+    [triggers, pendingIds, handleToggle]
+  );
+
+  const columns = useMemo(
+    () => buildColumns({ workspaceId, expandedRowId }),
+    [workspaceId, expandedRowId]
```
adrsimon: false positive i guess? reads workspaceId, not owner.sId directly

front/components/me/UserAutomationsTable.tsx
```diff
@@ -0,0 +1,422 @@
...
+        <SliderToggle
+          selected={row.status === "enabled"}
+          disabled={row.isStatusPending}
+          onClick={(event) => {
+            event.stopPropagation();
+            row.onToggleStatus();
+          }}
+        />
+      );
+    case "disabled_by_manager":
```
avervaet, line 95: Looks like an elevated user (manager/admin) can disable their own automation here but cannot re-enable it here. They have to go back to the workspace manager page. 

We might wan't to allow them to re-enable from here to avoid a clunky behavior

avervaet (approved):

PR #30973, opened 2026-08-21
[front] Poke: name-only plugin cards

PopDaph:
## Description

Plugin cards show only the name: the description is already in the hover tooltip. Uniform card height, cards fill their grid column, spacing between cards matches the container padding, and the card shadow is dropped (borders do the separation since #30971).

### DIFF
--- a/front/components/poke/plugins/PluginList.tsx
+++ b/front/components/poke/plugins/PluginList.tsx
@@ -2,7 +2,6 @@ import { PluginRunsList } from "@app/components/poke/plugins/PluginRunsList";
 import { RunPluginDialog } from "@app/components/poke/plugins/RunPluginDialog";
 import {
   PokeCard,
-  PokeCardDescription,
   PokeCardHeader,
   PokeCardTitle,
 } from "@app/components/poke/shadcn/ui/card";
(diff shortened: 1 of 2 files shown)


front/components/poke/plugins/PluginList.tsx
```diff
@@ -21,16 +20,13 @@ interface PluginCardProps {
...
-      className="flex h-20 w-44 cursor-pointer hover:bg-muted-background"
+      className="flex h-16 w-full cursor-pointer items-center hover:bg-muted-background"
       onClick={onClick}
     >
-      <PokeCardHeader className="flex space-y-2 overflow-hidden p-2 text-left">
+      <PokeCardHeader className="flex overflow-hidden p-2 text-left">
         <PokeCardTitle className="text-sm font-medium">
           {plugin.name}
         </PokeCardTitle>
-        <PokeCardDescription className="overflow-hidden truncate whitespace-normal text-sm">
```
aubin-tchoi, line 31: there's a chance this component can be deleted

aubin-tchoi (approved):
LGTM

front/components/poke/plugins/PluginList.tsx
PopDaph: Sorry merged before seeing 🤦🏻‍♀️

PopDaph: sending a follow up thanks for catching!

PR #30974, opened 2026-08-21
refactor(front): make the automations triggers table reusable

adrsimon:
## Description

Prep for a personal automations view, which needs the same table, filters and credit breakdown as the manager-only `/automations` page but scoped to a single member. Three things move out of their single call site: the row building in `automations/triggers.ts`, the name/type/credits/details column defs, and the "is this trigger mine" check, which becomes `TriggerResource.isEditedBy`. The filter panel and summary now take the list of categories to show, and the breakdown takes a scope that decides which endpoint it reads and how it words its median tooltip. No behavior change: the workspace page passes the full category list and `scope="workspace"`.

First of three. Follow-ups add the `/me/automations` endpoints, then the user-menu entry.

## Tests

Existing automations analytics and component suites pass unchanged (64 tests), including `AutomationsTriggerBreakdown.test.tsx` with the new required `scope`. No new tests: the extracted code has no new behavior of its own, and the callers are covered.

## Risk

Low, but it touches the shared table the `/automations` page renders. The one thing worth a close read is `AutomationsTriggersRowsTable`: `rowSelection` and `onRowSelectionChange` became optional, so a caller that forgets to wire them now silently gets no selection instead of a type error. The workspace table still passes both.

## Deploy Plan

Deploy front.

### DIFF
--- a/front/components/workspace/analytics/automations/automationsTriggerColumns.tsx
+++ b/front/components/workspace/analytics/automations/automationsTriggerColumns.tsx
@@ -1,0 +1,145 @@
+import { getIcon } from "@app/components/resources/resources_icons";
+import type { TriggerRowData } from "@app/components/workspace/analytics/automations/AutomationsTriggersRowsTable";
+import { CreditsCell } from "@app/components/workspace/analytics/creditsTableCells";
+import type { AutomationTriggerRow } from "@app/lib/api/analytics/automations/triggers";
+import { normalizeWebhookIcon } from "@app/lib/webhook_source";
+import { assertNeverAndIgnore } from "@app/types/shared/utils/assert_never";
+import {
+  Button,
+  ChevronDown,
+  ChevronUp,
+  Clock,
+  DataTable,
+  Icon,
+  Tooltip,
+} from "@dust-tt/sparkle";
+import type { ColumnDef } from "@tanstack/react-table";
+import type { ComponentProps } from "react";
+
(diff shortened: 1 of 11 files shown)


front/components/workspace/analytics/automations/automationsTriggerColumns.tsx
```diff
@@ -0,0 +1,145 @@
...
+  Tooltip,
+} from "@dust-tt/sparkle";
+import type { ColumnDef } from "@tanstack/react-table";
+import type { ComponentProps } from "react";
+
+function TypeLabel({
+  visual,
+  label,
+}: {
+  visual: ComponentProps<typeof Icon>["visual"];
```
aubin-tchoi: we can put the underlying type here no?

front/components/workspace/analytics/automations/automationsTriggerColumns.tsx
```diff
@@ -0,0 +1,145 @@
...
+          }
+        />
+      );
+    default:
+      assertNeverAndIgnore(trigger.kind);
+      return null;
+  }
+}
+
+export function nameColumn<T extends TriggerRowData>(): ColumnDef<T> {
```
aubin-tchoi, line 81: tbd but potentially more concise if you keep inlining this and template once the `buildColumns`

aubin-tchoi (approved):
LGTM

PR #30975, opened 2026-08-21
feat(front): serve a member their own automations and their credits

adrsimon:
## Description

A member currently has no way to see what their own automations cost: the `/automations` analytics page is manager-only. This adds `POST /api/w/:wId/me/automations/triggers` and its `trigger-breakdown` sibling, returning the same rows the admin table shows but only for the caller's triggers.
The list starts reuses the generic consumption ES, by scoping on a specific user.

Second of three, stacked on #30974. The user-menu entry that consumes this comes next.

## Tests

- `user_triggers.test.ts`: never-run triggers rank last, other members' triggers are excluded, kind and name filters apply, and a failed consumption query still returns the rows with `isConsumptionAvailable: false`.
- `me/automations/triggers.test.ts`: a plain `user` role gets a 200, period/paging/search/filter are forwarded, a negative offset is a 400.
- `me/automations/trigger-breakdown.test.ts`: 200 for the trigger's editor, 404 for a trigger another member edits, 404 for an unknown id. These use real triggers rather than mocking the authorization check, since it is the only gate on the route.

## Risk

Both routes are deliberately not role-gated, so authorization rests entirely on `listByUserEditor` for the list and `isEditedBy` for the breakdown. That is the part to review. Nothing existing changes behavior: the endpoints are new and unused until the next PR lands.

Elasticsearch failures no longer fail the request, they return the rows with credits at zero and a flag saying so. That is deliberate, so an ES incident does not take away a member's ability to see and manage their automations, but it does mean a member can see zeroes during one.

## Deploy Plan

Deploy front.

### DIFF
--- a/front-api/routes/w/[wId]/me/automations/index.ts
+++ b/front-api/routes/w/[wId]/me/automations/index.ts
@@ -1,0 +1,12 @@
+import { workspaceApp } from "@front-api/middlewares/ctx";
+
+import triggerBreakdown from "./trigger-breakdown";
+import triggers from "./triggers";
+
+// Mounted under /api/w/:wId/me/automations.
+const app = workspaceApp();
+
+app.route("/triggers", triggers);
+app.route("/trigger-breakdown", triggerBreakdown);
+
+export default app;
--- a/front-api/routes/w/[wId]/me/automations/triggers.test.ts
+++ b/front-api/routes/w/[wId]/me/automations/triggers.test.ts
@@ -1,0 +1,119 @@
+import { DEFAULT_AUTOMATION_TRIGGERS_LIMIT } from "@app/lib/api/analytics/automations/schema";
+import type { UserAutomationTriggers } from "@app/lib/api/analytics/automations/user_triggers";
+import { fetchUserAutomationTriggers } from "@app/lib/api/analytics/automations/user_triggers";
+import { createPrivateApiMockRequest } from "@app/tests/utils/generic_private_api_tests";
+import { honoApp } from "@front-api/app";
+import { describe, expect, it, vi } from "vitest";
+
+vi.mock(
+  import("@app/lib/api/analytics/automations/user_triggers"),
+  async (orig) => {
+    const mod = await orig();
+    return {
+      ...mod,
+      fetchUserAutomationTriggers: vi.fn(),
+    };
+  }
+);
+
+const TRIGGERS: UserAutomationTriggers = {
+  period: {
+    startDate: "2026-07-01T00:00:00.000Z",
+    endDate: "2026-08-01T00:00:00.000Z",
+  },
+  totalCount: 1,
+  isConsumptionAvailable: true,
+  medianRunCount: 4,
+  medianCostPerRun: 2,
+  triggers: [
+    {
+      triggerId: "trg1",
+      name: "Competitor watch",
+      kind: "schedule",
+      status: "enabled",
+      agent: {
+        agentId: "agent1",
+        name: "deep-dive",
+        pictureUrl: null,
+        description: null,
+        modelId: null,
+        modelDisplayName: null,
+      },
+      editor: {
+        name: "Adrien Simon",
+        email: "adrien@dust.tt",
+        pictureUrl: null,
+      },
+      scheduleDescription: "Every day at 9:00",
+      webhookSourceName: null,
+      webhookSourceRestricted: false,
+      webhookIcon: null,
+      runCount: 8,
+      credits: 16,
+      executionMode: "user_pool",
+    },
+  ],
+};
+
+function postTriggersRequest(wId: string, body: Record<string, unknown> = {}) {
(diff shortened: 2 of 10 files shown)


front-api/routes/w/[wId]/me/automations/index.ts
```diff
@@ -0,0 +1,12 @@
+import { workspaceApp } from "@front-api/middlewares/ctx";
+
+import triggerBreakdown from "./trigger-breakdown";
+import triggers from "./triggers";
+
+// Mounted under /api/w/:wId/me/automations.
+const app = workspaceApp();
+
+app.route("/triggers", triggers);
+app.route("/trigger-breakdown", triggerBreakdown);
```
Comment (flvndvd) [inline front-api/routes/w/[wId]/me/automations/index.ts:10]: If `/trigger-breakdown` is meant to be for one trigger, why not representing it semantically through `triggers/:tId/breakdown`?

front-api/routes/w/[wId]/me/automations/triggers.test.ts
```diff
@@ -0,0 +1,119 @@
...
+      expect.anything(),
+      expect.objectContaining({
+        limit: DEFAULT_AUTOMATION_TRIGGERS_LIMIT,
+        offset: 0,
+      })
+    );
+  });
+
+  it("forwards the period, page, search and filter", async () => {
+    vi.mocked(fetchUserAutomationTriggers).mockResolvedValue(TRIGGERS);
```
Comment (flvndvd) [inline front-api/routes/w/[wId]/me/automations/triggers.test.ts]: Why mocking over creating a few real instances in DB? Mocking a resource is never a good approach.

front/lib/api/analytics/automations/user_triggers.ts
```diff
@@ -0,0 +1,200 @@
...
+  }
+
+  const query = buildConsumptionScopeQuery({
+    auth,
+    startDate: period.startDate,
+    endDate: period.endDate,
+    extraFilters: [{ terms: { [TRIGGER_ID_FIELD]: triggerIds } }],
+  });
+
+  const result = await searchConsumptionAnalytics<never, TriggerAggs>(query, {
```
flvndvd: Can't we share some of it with the admin automations page? I'd expect the change to be only the user_id `term` clause, right?

front-api/routes/w/[wId]/me/automations/index.ts
Comment (adrsimon) [reply front-api/routes/w/[wId]/me/automations/index.ts:10]: will harmonize everything in a later PR

front/lib/api/analytics/automations/user_triggers.ts
```diff
@@ -0,0 +1,122 @@
...
+    limit: number;
+    offset: number;
+    search?: string;
+    filter?: UserAutomationTriggersFilter;
+  }
+): Promise<UserAutomationTriggers> {
+  const editorTriggers = await TriggerResource.listByUserEditor(
+    auth,
+    auth.getNonNullableUser()
+  );
```
adrsimon, line 67: fetch user's trigger first, we want to list every trigger, including those without consumption

front-api/routes/w/[wId]/me/automations/index.ts
Comment (flvndvd) [reply front-api/routes/w/[wId]/me/automations/index.ts:10]: 👍

front/lib/api/analytics/automations/user_triggers.ts
```diff
@@ -0,0 +1,122 @@
...
+            ({ triggerId, runCount, credits }) => [
+              triggerId,
+              { runCount, credits },
+            ]
+          )
+        )
+      : new Map();
+
+  const ranked: RankedTriggerWithResource[] = editorTriggers
+    .filter((trigger) => matchesFilter(trigger, { searchTerm, filter }))
```
flvndvd, line 93: Any reason to run this post-ES? Is it because of the full text search? Ok looks like `kind` can't be easily map as well? But it would be nice to filter agents directly in ES if possible.

front/lib/api/analytics/automations/user_triggers.ts
```diff
@@ -0,0 +1,122 @@
...
+  const rows = await buildAutomationTriggerRows(
+    auth,
+    ranked.slice(offset, offset + limit)
+  );
+
+  return {
+    period,
+    totalCount: ranked.length,
+    triggers: rows,
+    medianRunCount: median(active.map(({ runCount }) => runCount)),
```
flvndvd, line 119: It feels weird to apply a math operation on the count, right? Ideally if it runs 10 times, I should just get the absolute count?

flvndvd (approved):

front/lib/api/analytics/automations/user_triggers.ts
adrsimon: median is for the comparison to other runs, i remove the Math.round above though

front/lib/api/analytics/automations/user_triggers.ts
adrsimon: without that we wouldn't have zero-use triggers, need to re-rank afterwards

PR #30976, opened 2026-08-21
feat(agent_inactivity): preview and archive a workspace's unused agents

achilleburah:
Adds the two endpoints and the workspace setting behind the `archive_inactive_agents` flag.

- `POST .../archive_inactive/preview` counts what archival would take and `POST .../archive_inactive` takes it. Both cover the whole workspace and share one body schema, so the dry run and the executor cannot drift.
- `policy.ts` holds the rules as pure functions, so the preview and the executor read the same rules rather than each holding a copy.
- `inactiveAgentArchivalThresholdDays` on the workspace carries both whether automatic archival is on and how many days, matching the policy's rule that with no threshold set nothing is ever archived.
- Seeds a workspace whose answer is known in advance: five agents, two eligible and three spared, one of them an agent edited today whose first version is old.

## Description

Workspaces accumulate agents nobody uses. Admins had no way to find them, and no way to clear them out without opening each one.

An agent is archivable when all of the following hold:

| Rule | Spared when |
| --- | --- |
| Status | not `active` |
| Schedule | an enabled (or `relocating`/`downgraded`) schedule trigger drives it |
| Age | its **first** version was created after the cutoff |
| Mentions | its last mention is at or after the cutoff |

The cutoff is `evaluatedAt - thresholdDays`, truncated to the UTC day so a preview and a later run judge every agent identically, and so a threshold of N days always leaves a full N days of grace. Thresholds are bounded to 2–366 days.

Two details worth calling out:

- **Age is read from version 0, not the active row.** Editing an agent inserts a new row, so the active row's `createdAt` is really the last edit. Reading it would let editing an agent postpone its archival forever.
- **A recently mentioned agent is spared without appearing in `skippedCountByReason`.** The candidates query (`MentionResource.listAgentsNotMentionedSince`) filters on the same cutoff, so such an agent never reaches the rules to be counted by them. The rule stays in `policy.ts` anyway, so that one module still states what "inactive" means.

Archiving goes through the existing `archiveAgentConfiguration`, which keeps trigger disabling, wake-up cancellation, editor-group suspension and the `agent.archived` audit event on the one path that already owned them.

**This PR is the API only.** The endpoints and the rules land here; the Governance setting, the manage-agents CTA and the archival modal come in a follow-up PR, and so does the nightly run. `inactiveAgentArchivalThresholdDays` can be written through `POST /api/w/:wId` but nothing reads it yet.

## Tests

- Endpoint tests for both routes: non-admin → 403, flag off → 403 and nothing touched, threshold outside the bounds → 400, the counts for a five-agent workspace, what actually gets archived, and a second call archiving nothing.
- Resource-level tests for `fetchArchivableAgents`: never mentioned, last mention before/after the cutoff, newest mention wins, a rejected mention still counts as activity, a recently edited agent still qualifies, an agent created after the cutoff does not, global agents never appear.
- Unit tests for `policy.ts` covering the cutoff arithmetic, the threshold bounds and every exclusion reason.
- `scripts/seed/inactivity/seed.ts` builds a workspace whose expected preview output is documented in its README, for checking the endpoints by hand.

## Risk

- **Off by default.** `archive_inactive_agents` is `on_demand`, so there is no behaviour change for any workspace until it is enabled. Both endpoints are admin-only.
- **Reversible.** Agents are archived, not deleted, and an admin can restore them. A revert of this PR does not un-archive anything already archived.
- **Unbounded work per request.** `POST .../archive_inactive` archives every eligible agent in one request, and each one disables its triggers, cancels its wake-ups (Temporal round-trips), suspends its editor group and emits an audit event. On a workspace with many idle agents that request will be slow; a workspace large enough could time out. The preview is a read but hydrates every idle agent's configuration.
- Archiving an agent suspends its editor group memberships, which is visible to those users.

## Deploy Plan

- [ ] Merge.
- [ ] Register the new audit action with WorkOS **before** deploying: `npx tsx front/admin/register_audit_log_schemas.ts --execute --changed`. Without it, `workspace.inactive_agent_archival_updated` emits with no schema version.
- [ ] Deploy `front-api` and `front`.
- [ ] Enable `archive_inactive_agents` on one internal workspace via Poke and check the preview counts before enabling it anywhere else.

`sdks/js` picks up `archive_inactive_agents` in its `WhitelistableFeature` union; this is an additive enum value on a system-key-only endpoint, so no client change is required.

## Follow-ups

- The UI: the Governance setting, the manage-agents CTA and the archival modal, with the SWR hooks that drive them.
- The nightly run that reads `inactiveAgentArchivalThresholdDays` and archives on its own.
- Bounding the work per archive request once the nightly run exists to drive it.

### DIFF
--- a/front-api/routes/w/[wId]/assistant/agent_configurations/archive_inactive.test.ts
+++ b/front-api/routes/w/[wId]/assistant/agent_configurations/archive_inactive.test.ts
@@ -1,0 +1,354 @@
+import { getAgentConfiguration } from "@app/lib/api/assistant/configuration/agent";
+import { ONE_DAY_MS } from "@app/lib/api/assistant/inactivity/policy";
+import type { Authenticator } from "@app/lib/auth";
+import * as scheduleClient from "@app/temporal/triggers/schedule_client";
+import * as wakeUpClient from "@app/temporal/triggers/wakeup_client";
+import { AgentConfigurationFactory } from "@app/tests/utils/AgentConfigurationFactory";
+import { FeatureFlagFactory } from "@app/tests/utils/FeatureFlagFactory";
+import { createPrivateApiMockRequest } from "@app/tests/utils/generic_private_api_tests";
+import { MentionFactory } from "@app/tests/utils/MentionFactory";
+import { TriggerFactory } from "@app/tests/utils/TriggerFactory";
+import {
+  ArchiveInactiveAgentsResponseBodySchema,
+  PreviewInactiveAgentsResponseBodySchema,
+} from "@app/types/api/assistant/configuration";
+import type { MembershipRoleType } from "@app/types/memberships";
+import { Ok } from "@app/types/shared/result";
+import { honoApp } from "@front-api/app";
+import { beforeEach, describe, expect, it, vi } from "vitest";
+import { z } from "zod";
+
+const THRESHOLD_DAYS = 30;
+
+/** Old enough to sit well before any cutoff a 30-day threshold can produce. */
+const LONG_AGO = new Date(Date.now() - 90 * ONE_DAY_MS);
+
+const ErrorBodySchema = z.object({
+  error: z.object({ message: z.string() }),
+});
+
+// Parsing rather than casting also asserts the response actually matches its published contract.
+async function previewBody(response: Response) {
+  return PreviewInactiveAgentsResponseBodySchema.parse(await response.json())
+    .preview;
+}
+
+async function archivalBody(response: Response) {
+  return ArchiveInactiveAgentsResponseBodySchema.parse(await response.json())
+    .archival;
+}
+
+async function errorBody(response: Response) {
+  return ErrorBodySchema.parse(await response.json()).error;
+}
+
+async function setupTest({
+  role = "admin",
+  withFeatureFlag = true,
+}: {
+  role?: MembershipRoleType;
+  withFeatureFlag?: boolean;
+} = {}) {
+  const { workspace, auth } = await createPrivateApiMockRequest({ role });
+
+  if (withFeatureFlag) {
+    await FeatureFlagFactory.basic(auth, "archive_inactive_agents");
+  }
+
+  return { workspace, auth };
+}
+
+function postPreview(wId: string, body: Record<string, unknown>) {
+  return honoApp.request(
+    `/api/w/${wId}/assistant/agent_configurations/archive_inactive/preview`,
+    {
+      method: "POST",
+      headers: { "Content-Type": "application/json" },
+      body: JSON.stringify(body),
+    }
+  );
+}
+
+function postArchive(wId: string, body: Record<string, unknown>) {
+  return honoApp.request(
+    `/api/w/${wId}/assistant/agent_configurations/archive_inactive`,
+    {
+      method: "POST",
+      headers: { "Content-Type": "application/json" },
+      body: JSON.stringify(body),
+    }
+  );
+}
+
+/** Creating an enabled schedule, and archiving, both reach Temporal. */
+function mockTemporalClients() {
+  vi.spyOn(scheduleClient, "createOrUpdateAgentSchedule").mockResolvedValue(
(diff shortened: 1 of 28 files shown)


front/scripts/seed/inactivity/README.md
```diff
@@ -0,0 +1,52 @@
...
+| `Scheduled Agent` | created 90d ago, unmentioned, enabled schedule | `active_schedule` |
+| `Fresh Agent` | created today, never mentioned | `recent_creation` |
+| `Edited Agent` | first version 90d ago, edited today, unmentioned | **eligible** |
+
+`Recently Used Agent` is spared without a skip reason: the candidates query filters on the same
+cutoff, so a recently mentioned agent never reaches the rules to be counted by them.
+
+`Edited Agent` is the interesting one: upgrading an agent inserts a new row, so its active row looks
+young. Reading that date instead of the first version's would let editing an agent postpone its
+archival forever.
```
fabiencelier: A bit too much details not related to the seed but to the inactivity process
I think the description of the agent before that is enough

front/scripts/seed/inactivity/README.md
```diff
@@ -0,0 +1,52 @@
...
+- Temporal must be reachable: creating the enabled schedule starts a workflow. Its cron is yearly, so
+  the seeded agent will not actually run.
+
+## Usage
+
+```bash
+npx tsx scripts/seed/inactivity/seed.ts --execute
+```
+
+Then enable the `archive_inactive_agents` feature flag on the workspace through Poke and call:
```
fabiencelier: You can enable the flag with the seed and remove this from the README

front/scripts/seed/inactivity/README.md
```diff
@@ -0,0 +1,52 @@
...
+
+```
+POST /api/w/<wId>/assistant/agent_configurations/archive_inactive/preview  { "thresholdDays": 30 }
+```
+
+Expect `eligibleCount: 2` and `skippedCountByReason: { active_schedule: 1, recent_creation: 1 }`.
+
```
fabiencelier: This can be removed as it will be accessible from the UI no ?

front-api/routes/w/[wId]/index.ts
```diff
@@ -877,6 +892,44 @@ app.post(
           enabled: String(!body.disableWorkspaceAnalytics),
         },
       });
+    } else if ("inactiveAgentArchivalThresholdDays" in body) {
```
Comment (fabiencelier) [inline front-api/routes/w/[wId]/index.ts:895]: You can add 1 small unit test for this one too

front-api/routes/w/[wId]/assistant/agent_configurations/archive_inactive.test.ts
```diff
@@ -0,0 +1,349 @@
...
+  };
+}
+
+interface PreviewBody {
+  preview: {
+    eligibleCount: number;
+    skippedCountByReason: Record<string, number>;
+    cutoffAt: string;
+  };
+}
```
Comment (fabiencelier) [inline front-api/routes/w/[wId]/assistant/agent_configurations/archive_inactive.test.ts:43]: Why do we need to redefine the type here ?
Can't we use the actual exported ones PreviewInactiveAgentsResponseBody ?

front/scripts/seed/inactivity/seed.ts
```diff
@@ -0,0 +1,153 @@
...
+    agentConfigurationId: agent.sId,
+  });
+
+  if (result.isErr()) {
+    throw result.error;
+  }
+}
+
+makeScript({}, async ({ execute }, logger) => {
+  const ctx = await createSeedContext({ execute, logger });
```
fabiencelier, line 6: can you please add a seed.test.ts to make sure the seed does no go broken ?
Unless there is something making it to complex to test

fabiencelier (approved):
LGTM on the design 
Thanks for adding a new seed :) 
Added a few comments

front/scripts/seed/inactivity/README.md
achilleburah: yeah, it will be with next PR, removing this

PR #30980, opened 2026-08-21
Introduce cached resource store and refit WorkspaceResource onto it

flvndvd:
## Description

A resource read from the cache must be indistinguishable from one read from the database. This PR makes that an invariant enforced by structure rather than by convention, using the workspace resource as the first consumer.

Two rules carry the invariant:
- One store: a resource declares once how its rows are fetched, cached and invalidated. The cache holds the full set of columns of the row, serialized generically from the model definition, so the cached shape can never drift from the database shape, and there is no hand written snapshot to forget a field in.
- One materialization: a single seam where a raw row becomes a resource, and the only place where context living outside the row is folded in. Every path that produces a resource goes through it: cache hit, cache miss, transaction bypass, fallback when the cache is unreachable, uncached queries, and creation. A caller can no longer build a resource that skipped it.

For the workspace, the context folded in at materialization is the provider kill switches (TBH should not even be here, but will refactor in a follow up PR). Previously they applied on a single fetch path, so the same workspace carried different provider values depending on how it had been fetched. They now apply uniformly, while the cache keeps the raw column value. In the same spirit, a coercion of the SSO enforcement flag that only happened on the cached path is removed, so the cached and uncached values agree.

Since materialization now runs on every read, the kill switch lookup gets a database fallback: an outage of the cache backend must not take workspace reads down.

## Tests

Added one asserts that a workspace served from the cache carries exactly the same attributes as one served from the database. One asserts the kill switch filter applies on a previously unfiltered fetch path. One plants a snapshot in the exact format written by the currently deployed code and asserts it still parses; it also compares the snapshot fields against the model definition, so it fails whenever the model gains or loses a column, forcing a conscious bump of the cache version, which matters because entries have no expiry.

## Risk

The cache key and version are unchanged. The generically produced snapshot was verified to contain exactly the same fields as the previous hand written one, so during a rolling deploy old and new pods read each other's entries and invalidate the same keys. The behavioral deltas are the two described above: kill switches filtering all fetch paths (only model gating and the provider pickers consume the value, and both already received the filtered one) and the SSO flag no longer coerced on the cached path (all consumers are truthiness checks). Rollback is safe since the previous code parses the same entries.

## Deploy Plan

Deploy front normally. No migration, no cache key rotation, no ordering constraint.

### DIFF
--- a/front/lib/resources/cached_resource_store.ts
+++ b/front/lib/resources/cached_resource_store.ts
@@ -317,6 +317,12 @@ export type CachedResourceStore<
     blob: Attributes<M>,
     transaction?: Transaction
   ) => Promise<void>;
+  // Key-based counterpart of invalidateBlob, for callers that only hold the cache key (e.g.
+  // invalidation after out-of-band writes such as a relocation).
+  invalidateCached: (
+    input: Attributes<M>[K],
+    transaction?: Transaction
+  ) => Promise<void>;
   createCacheOperations: <OperationsInput>(definition: {
     label: string;
     params: CacheOperationParam[];
@@ -423,6 +429,8 @@ export function defineCachedResourceStore<
       return resource;
     },
     invalidateBlob,
+    invalidateCached: (input, transaction) =>
+      blobLookup.invalidate(input, transaction),
     createCacheOperations: blobLookup.createCacheOperations,
   };
 }
--- a/front/lib/resources/cached_resource_store.ts
+++ b/front/lib/resources/cached_resource_store.ts
@@ -327,6 +327,11 @@ export type CachedResourceStore<
  * The cached snapshot shape follows the model implicitly: whenever the model gains, loses, or
  * retypes an attribute, bump `cache.version`. Entries have no TTL, so stale-shaped snapshots
  * otherwise live until the row is next invalidated.
+ *
+ * Only global models (no workspaceId column) may use this store. Workspace-scoped resources need
+ * a variant that takes an Authenticator and scopes every where clause to the workspace id, and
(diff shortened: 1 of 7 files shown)


front/lib/resources/cached_resource_store.ts
```diff
@@ -245,6 +255,159 @@ export function defineCachedResourceLookup<Input, Snapshot, Resource>({
   };
 }
 
+// Serialized form of a model's attributes: DATE attributes are stored as epoch milliseconds,
```
flvndvd, line 258: Sequelize quirks.

front/lib/resources/workspace_resource.ts
```diff
@@ -279,8 +190,43 @@ export class WorkspaceResource extends BaseResource<WorkspaceModel> {
...
+          ...blob,
+          whiteListedProviders: WorkspaceResource.filterKillSwitchedProviders(
+            blob.whiteListedProviders,
+            enabledKillSwitches
+          ),
+        })
+    );
   }
 
   protected override async update(
```
aubin-tchoi, line 232: would be nice to have a way to make sure that if a `CachedResourceStore` is defined on the resource class we invalidate in the base `update`, as is I think this will force us to override the base resource's update on every resource (here we already had an override so it's ok, but there are resource where we'd need to add an override I thikn)

aubin-tchoi (approved):
LGTM

front/lib/resources/workspace_resource.ts
flvndvd: Yeah, my plan is to hoist it in the `BaseResource`.

PR #30982, opened 2026-08-21
MCP: remove the workspace_people mcp and move list_workspace_members into workspace_management mcp

fabiencelier:
## Description

Closes https://github.com/dust-tt/tasks/issues/9970 as analyst will be able to list all members of the workspace.

remove the MCP introduced in https://github.com/dust-tt/dust/pull/30732 to merge into the new workspace_management server.
Expose the same `list_workspace_members` restricted to managers and admin, in a different MCP server, with the following improvement:
 - allow to list members without filter to list all members
 - paginate large result
 - return string and not a JSON to be more compact
 - returning group is optional to make default result more compact when not needed (users can belong too a lot of groups)

Previous MCP was not exposed at all, new one is exposed via the `analyst` agent and their `analytics` skills (exposed to managers and admins)

Allows the `@analyst` agent to answre queries like this:
image

## Tests

 - tested locally
 - unit tests

## Risk

low

## Deploy Plan

deploy front.
run migration 
```
npx tsx migrations/20260821_sunset_workspace_people_mcp_server.ts
```


frankaloia (approved):
Looks great

PR #30986, opened 2026-08-21
feat(sandbox): list configured Pods + bulk egress read for the Computer admin

smb2268:
## Problem

First PR in the stack that builds the central Computer admin page (managing sandbox network access across the Workspace and its Pods). This one is the **backend read plumbing only** — no UI, no writes.

## What this PR does

- **`listPodSIdsWithEgressPolicy`** — one GCS prefix list (`sandboxes/vlt_`) to find the Pods that have their own egress policy, i.e. the Pods that have diverged from the workspace baseline.
- **`GET /w/:wId/sandbox/egress-policy/pods`** — returns those (live, non-archived) Pods for the admin scope selector, so the dropdown only offers configured Pods.
- **`GET /w/:wId/sandbox/egress-policy/bulk`** — returns each configured Pod's policy for a selection (`?podIds=…`) or `?scope=all-pods`.
- Moves the workspace `egress-policy` route into a directory so `/bulk` and `/pods` sit beside it.

## Why "configured Pods only"

Pod egress policies are one GCS object per Pod, and in practice only a small fraction of Pods set their own domains. Reading every selected Pod would fan out into mostly-404 reads and make "select all" expensive. Listing the configured Pods first makes reads scale with the Pods that actually diverged, not the total Pod count — so no cap / chunking / pagination is needed.

## Gating

Reuses the existing workspace-admin + Computer gates on the parent sub-app; the multi-Pod routes additionally require the `sandbox_functions` flag.

## Risk

Read-only, additive endpoints behind an admin + feature-flag gate. Worst case the new endpoints misbehave for admins with the flag on; nothing else is affected. Safe to roll back.

## Tests

- Bulk read: 403 non-admin, 403 without `sandbox_functions`, configured-only filtering, `scope=all-pods`, and scope/podIds validation.
- Pods list: 403 non-admin, configured-only + names, archived exclusion.
- Existing workspace egress route tests carried over unchanged.

## Stack

PR 1 of ~7 for the Computer admin page. Next up: bulk egress **write**, then the env-dialog refactor, the admin page UI, network editing, and Pod-settings read-only for members.


davidebbo (approved):
LGTM

PR #30987, opened 2026-08-21
feat(agent_inactivity): archive unused agents from Governance and the archived tab

achilleburah:
Wires the UI onto the endpoints added in #30976.

> [!IMPORTANT]
> **Stacked on #30976.** This PR's base is `feat/archive-inactive-agents-preview`, not `main` — it is unmergeable until that one lands, and the diff shown here is UI-only because of it. Retarget to `main` after #30976 merges.

## Description

#30976 added the rules and the two endpoints but nothing that calls them. This adds the two places an admin can:

**Governance › Agents — "Archive unused agents"** (`InactiveAgentArchival.tsx`)

A day count, a toggle, and a "Run now" button.

- The toggle writes `inactiveAgentArchivalThresholdDays` on the workspace, or clears it. Cleared means automatic archival is off, matching the policy's rule that with no threshold set nothing is ever archived.
- The day count is committed on blur, not per keystroke, so a half-typed number is never saved. Out-of-range values put the input in its error state and disable the toggle rather than being sent.
- "Run now" opens the modal below, so an admin can act once without waiting for a run.

**The archival modal** (`ArchiveUnusedAgentsModal.tsx`)

Two steps: pick a cutoff, see what archiving it would take, then archive.

- Step one previews. The count comes from the same read the executor uses, so the confirmation states the real number rather than an estimate.
- The picker is a date, converted to a day count. Whole days back are rounded, so a day picked across a DST boundary does not land a day off.

**The archived tab's empty state** (`NoArchivedAgentsCTA.tsx`)

Offering "create an agent" on a tab that lists archived agents reads as a non-sequitur, so it offers the thing that fills the tab instead. Non-admins keep a plain message with no action.

Both entry points are admin-only and behind `archive_inactive_agents`, matching the endpoints.

#### When Toggle is enabled
image
#### When Toggle is disabled
image

### Governance View for activating daily runs:
image

#### Empty
image
image

#### Found Agents
image

## Risk

- **Off by default.** `archive_inactive_agents` is `on_demand`; with the flag off, `InactiveAgentArchival` renders nothing and the archived tab keeps its previous empty state.
- **The modal archives for real.** It is a two-step flow with a preview, but the second step is not undoable in bulk — an admin restores agents one at a time. The preview count is what stands between a mistyped cutoff and a lot of archiving.
- 
## Deploy Plan

Nothing beyond #30976's steps. Deploy after it, then enable the flag per workspace.

### DIFF
--- a/front/components/assistant/manager/NoArchivedAgentsCTA.tsx
+++ b/front/components/assistant/manager/NoArchivedAgentsCTA.tsx
@@ -1,0 +1,84 @@
+import {
+  ArchiveUnusedAgentsModal,
+  DEFAULT_INACTIVITY_THRESHOLD_DAYS,
+} from "@app/components/workspace/ArchiveUnusedAgentsModal";
+import { useAuth, useFeatureFlags } from "@app/lib/auth/AuthContext";
+import { useAppRouter } from "@app/lib/platform";
+import type { WorkspaceType } from "@app/types/user";
+import { getInactiveAgentArchivalThresholdDays } from "@app/types/user";
+import { Archive, Button, EmptyCTA, Settings02 } from "@dust-tt/sparkle";
+import { useState } from "react";
+
+interface NoArchivedAgentsCTAProps {
+  owner: WorkspaceType;
+  onArchived: () => void;
+}
+
+/**
+ * The archived tab with nothing in it. Offering "create an agent" here reads as a non-sequitur, so
+ * this offers the thing that fills the tab instead — but only to someone who can actually do it:
+ * archiving is admin-only and behind a flag, so anyone else gets the message without the button.
+ */
+export function NoArchivedAgentsCTA({
+  owner,
+  onArchived,
+}: Readonly<NoArchivedAgentsCTAProps>) {
+  const { isAdmin } = useAuth();
+  const { hasFeature } = useFeatureFlags();
+  const router = useAppRouter();
+
+  const [isModalOpen, setIsModalOpen] = useState(false);
+
+  const thresholdDays = getInactiveAgentArchivalThresholdDays(owner);
+
+  if (!isAdmin || !hasFeature("archive_inactive_agents")) {
+    return (
+      <EmptyCTA
+        title="No archived agents"
+        message="Archived agents are kept here, out of everyone's way."
+        action={null}
+      />
+    );
+  }
+
+  return (
+    <>
+      <EmptyCTA
+        title="No archived agents"
+        message={
+          thresholdDays
+            ? `No agent has been archived. Agents nobody has mentioned for ${thresholdDays} days will be automatically archived.`
+            : "No agent has been archived. Agents nobody uses can be archived automatically, or right now."
+        }
+        action={
+          <div className="flex flex-row items-center gap-2">
+            {!thresholdDays && (
+              <Button
+                label="Enable daily checks"
+                icon={Settings02}
+                variant="outline"
+                tooltip="Set it up in the workspace's governance settings."
+                onClick={() => void router.push(`/w/${owner.sId}/governance`)}
+              />
+            )}
+            <Button
+              label="Archive unused now"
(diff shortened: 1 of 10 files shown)


front/components/assistant/manager/NoArchivedAgentsCTA.tsx
```diff
@@ -0,0 +1,84 @@
...
+    );
+  }
+
+  return (
+    <>
+      <EmptyCTA
+        title="No archived agents"
+        message={
+          thresholdDays
+            ? `No agent has been archived. The daily check archives agents nobody has mentioned for ${thresholdDays} days.`
```
fabiencelier: No a fan of mentionning "The daily check" here, it's more an implementation details we don't need to surface to user.
Maybe
> Agents nobody has mentioned for X days will be automatically archived.

front/components/pages/builder/agents/ManageAgentsPage.tsx
```diff
@@ -434,6 +436,15 @@ export function ManageAgentsPage() {
...
               />
+            ) : activeTab === "archived" ? (
+              <div className="pt-2">
+                <NoArchivedAgentsCTA
+                  owner={owner}
+                  onArchived={() => {
+                    void mutateArchivedAgentConfigurations();
+                  }}
+                />
+              </div>
```
fabiencelier, line 447: Out of scope for this PR, but I think it would be nice to have the actions and info about auto archiving agents even if not in an empty state?
What do you think @achilleburah @ClementAupiais ?

front/components/workspace/ArchiveUnusedAgentsModal.tsx
fabiencelier, line 1: Image

"days" feels very close to the border no ?
Maybe it's just me but it feels closer than other input like this one

Image

front/components/workspace/settings/InactiveAgentArchival.tsx
```diff
@@ -0,0 +1,140 @@
...
+    setIsChanging(false);
+  };
+
+  return (
+    <>
+      <GovernanceSettingRowLayout
+        label="Archive unused agents"
+        description={
+          <>
+            Activate a daily check on unused agents. Or{" "}
```
fabiencelier: Same here, do we really want to mention daily check or just say "automatically archive".
Maybe daily check makes sense as the archive is not happening directly ?

front/components/workspace/settings/InactiveAgentArchival.tsx
```diff
@@ -0,0 +1,140 @@
...
+            containerClassName="w-auto shrink-0"
+            suffix="days"
+            disabled={isChanging || !isEnabled}
+            message={
+              isThresholdValid
+                ? undefined
+                : `Between ${MIN_INACTIVITY_THRESHOLD_DAYS} and ${MAX_INACTIVITY_THRESHOLD_DAYS}`
+            }
+            messageStatus="error"
+          />
```
fabiencelier: We have an InputWithSave component which adds a small save button when entering edit modes 

I think it could be very adapted to this situation to avoid saving on un focus

Image

front/components/workspace/ArchiveUnusedAgentsModal.tsx
achilleburah: yeah that was that was the previous version, it should be better in the second one (updated screenshots are in the PR)

front/components/assistant/manager/NoArchivedAgentsCTA.tsx
achilleburah: yeah that's better

front/components/pages/builder/agents/ManageAgentsPage.tsx
achilleburah: yes definitely

fabiencelier (approved):

PR #30990, opened 2026-08-21
[front] refactor: keep @temporalio/worker out of the client-side temporal module

id13:
## Description

temporalio/worker package is quite fat, or mostly its dependencies, like webpack (no it's not a _dev_ dependency 🤯), while we realistically only need temporalio/client in most case.

This pr splits our lib/temporal to avoid importing the former in front-api.

## Tests

Existing tests

## Risk

Low

## Deploy Plan

Deploy front

PR #30988, opened 2026-08-21
[front] perf: build the Langfuse tracer provider without @opentelemetry/sdk-node

id13:
## Description

This stack is about reducing the size of what needs to be imported / parsed in production to reduce boot time (currently between 45s - 150s)

This PR drops OTEL SDK, and instead use tracing API. OTEL SDK loads a lot of things we don't use (metrics, logs, etc) but still pay for.

## Tests

Tested in front-edge (top of the stack)

## Risk

Low 

## Deploy Plan

Deploy front

PR #30991, opened 2026-08-21
[front] enh: fail the build when browser-only packages reach front-api's boot path

id13:
## Description

Add a best-effort hook to try to limit occurrences of front-end stuff imports in front-api (transitive via lib/api)

Not magic, if you really want you can still import tiptap from lib/api.

It assumes our agent and us follow the codebase rules.

Also move some code around to make the check pass

## Tests

N/A

## Risk

Low

## Deploy Plan

Merge


id13:

- #30988
    - #30989
        - #30990
            - #30991 ◀
                - #30993
                    - #30994

id13:
```
✗ noClientImportsInServer
  ✓ Should flag a value import from components
  ✓ Should flag a value import from Sparkle
  ✗ Should not flag a type-only import
    Expected no matches, but got one
```


🫠 

at least it works 👍 

will fix whatever is the cause

id13:
PMRR

PR #30993, opened 2026-08-21
[front] perf: load vendor SDKs on first use instead of at boot

id13:
## Description

Our MCP internal servers are one of the biggest source of heavy node_modules imports (we use 3P sdks, which are more or less responsible in terms of supply chain).

And it's not critical for front-api's start to have them imported right away (or temporal workers, for that matter).

Deferring is one solution to this, although it raises some risks, especially if we end-up seeing agents adopting this pattern more broadly. This could be a reason NOT to do it

## Tests

Tested in front-edge (to of the stack).

## Risk

Low

## Deploy Plan

Deploy front


id13:

- #30988
    - #30989
        - #30990
            - #30991
                - #30993 ◀
                    - #30994

flvndvd (approved):

PR #30994, opened 2026-08-21
[front] perf: bundle front-api's dependencies instead of externalizing them

id13:
## Description

- Bundle as many node_modules as possible to 
   - Save a lot of time on Pod start (network calls to get docker img layers on cold nodes are expensive + CPU time parsing a lot of fluff)
   - Have a better idea of the footprint of our app and whether there are low hanging fruit
 - side 🚗 : delete a few biome rules that were not that important (imho)
 - upload the said bundle size metric to datadog as a gauge
 - add a last-resort check to make sure we don't leak frontend code in the bundle

## Tests

Tested on front-edge

## Risk

Low

## Deploy Plan

Deploy front

### DIFF
--- a/front-api/esbuild.production.ts
+++ b/front-api/esbuild.production.ts
@@ -5,6 +5,7 @@ import {
   type BuildTarget,
   getBaseBuildOptions,
 } from "./esbuild.shared";
+import { checkBundleContents } from "./lib/build-checks";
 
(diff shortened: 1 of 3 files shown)


id13:

- #30988
    - #30989
        - #30990
            - #30991
                - #30993
                    - #30994 ◀

PR #30995, opened 2026-08-21
feat(sandbox): bulk egress write across Workspace and Pods

smb2268:
## Problem

PR 2 of the Computer admin stack. PR 1 added the multi-Pod egress **read**; this adds the **write** — applying a single domain add/remove across the Workspace baseline and/or selected Pods in one call.
## What this PR does

- **`addWorkspacePolicyDomain` / `removeWorkspacePolicyDomain` / `removeOwnerPolicyDomain`** — the workspace-scoped and removal counterparts of the existing `addOwnerPolicyDomain` (same exact-domain parse, dedupe, and 100-domain cap).
- **`bulkUpdateEgressDomain`** applies one `{ add | remove, domain }` across the selected scopes (Workspace when included, plus each Pod), preserving every scope's other domains, returning a per-scope `{ scopeId, success, errorMessage? }`, and emitting `sandbox_egress_policy.updated` **only for scopes that actually changed**.
- **`POST /w/:wId/sandbox/egress-policy/bulk`** wires it up.

## Semantics

It's *"add/remove domain X across these scopes,"* not *"set these Pods' policy to P."* Each Pod is validated as a live project space (unknown ids come back as a per-scope `Pod not found`, they don't sink the batch). The body caps `podIds` at 100.

## Gating

Same as the read: workspace-admin + Computer (parent sub-app) + `sandbox_functions` on the route.

## Risk

Additive write endpoint behind admin + feature-flag gates. Worst case a bulk write misbehaves for admins with the flag on; per-scope isolation keeps one Pod's failure from affecting others. Safe to roll back.

## Tests

Add/remove across Workspace + Pods (files + audit), per-scope failure for unknown Pods, workspace-only (Pod files untouched), and no-op idempotency (adding a present domain / removing an absent one skips the audit). Response bodies parsed with zod (no `as`).

### DIFF
--- a/front/lib/api/sandbox/admin_pods.ts
+++ b/front/lib/api/sandbox/admin_pods.ts
@@ -12,7 +12,7 @@ import {
 } from "@app/lib/api/sandbox/egress_policy";
 import type { AuditLogContext } from "@app/lib/api/workos/organization";
 import type { Authenticator } from "@app/lib/auth";
-import { SpaceResource } from "@app/lib/resources/space_resource";
+import type { SpaceResource } from "@app/lib/resources/space_resource";
 import { SANDBOX_WORKSPACE_SCOPE_ID } from "@app/types/api/sandbox/egress_policy";
 import type { EgressPolicy } from "@app/types/sandbox/egress_policy";
 import type { Result } from "@app/types/shared/result";
@@ -234,12 +234,17 @@ export async function bulkUpdateEgressDomain(
     }
   }
 
-  const spaces = await SpaceResource.fetchByIds(auth, podIds);
-  const spacesById = new Map(spaces.map((space) => [space.sId, space]));
+  // Validate against the same live, non-archived project Pods the read path
+  // surfaces, so an archived Pod's id can't be used to mutate its policy.
+  const livePods = await listNonArchivedProjectSpacesAsAdmin(auth);
+  if (livePods.isErr()) {
+    // Admin-gated by the route; a failure here is a should-never-happen.
+    throw livePods.error;
+  }
+  const livePodIds = new Set(livePods.value.map((pod) => pod.sId));
 
   for (const podId of podIds) {
-    const pod = spacesById.get(podId);
-    if (!pod || !pod.isProject()) {
+    if (!livePodIds.has(podId)) {
       results.push({
         scopeId: podId,
         success: false,
--- a/front/lib/api/sandbox/admin_pods.ts
+++ b/front/lib/api/sandbox/admin_pods.ts
@@ -228,6 +228,13 @@ export async function bulkUpdateEgressDomain(
     }
   }
 
+  // A workspace add already covers every Pod, so writing to the selected Pods
+  // is redundant. Remove still targets each Pod below (a Pod can hold its own
+  // copy a workspace remove would not clear).
+  if (operation.operation === "add" && includeWorkspace) {
+    return results;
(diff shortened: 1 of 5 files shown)


front/lib/api/sandbox/admin_pods.ts
```diff
@@ -72,3 +87,192 @@ export async function listPodsWithEgressPolicy(
...
+  {
+    podId,
+    policy,
+    context,
+  }: { podId: string | null; policy: EgressPolicy; context?: AuditLogContext }
+): void {
+  const workspace = auth.getNonNullableWorkspace();
+  void emitAuditLogEvent({
+    auth,
+    action: "sandbox_egress_policy.updated",
```
davidebbo, line 112: You'll need to register the audit schema, but you know all about that :)

front/lib/api/sandbox/admin_pods.ts
```diff
@@ -72,3 +87,192 @@ export async function listPodsWithEgressPolicy(
...
+        const r = await addWorkspacePolicyDomain(auth, { domain });
+        if (r.isErr()) {
+          return r;
+        }
+        return new Ok({
+          policy: r.value.policy,
+          changed: r.value.addedDomain !== null,
+        });
+      }
+      case "remove": {
```
davidebbo, line 165: there is no update? It's remove and re-add differently?

front/lib/api/sandbox/admin_pods.ts
```diff
@@ -72,3 +87,192 @@ export async function listPodsWithEgressPolicy(
...
+      allowed_domain_count: String(policy.allowedDomains.length),
+      allowed_domains: policy.allowedDomains.join(","),
+      ...(podId ? { space_id: podId } : {}),
+    },
+  });
+}
+
+// Applies one egress-domain add/remove to each selected scope, reusing the
+// per-scope helpers and audit event the single-scope routes use. Sequential;
+// the route schema caps podIds at 100.
```
davidebbo, line 132: Not sure I understand this line. The pod id can only go up to 100?

davidebbo: Ah, nm, talking about array size :)

front/lib/api/sandbox/admin_pods.ts
```diff
@@ -72,3 +87,192 @@ export async function listPodsWithEgressPolicy(
...
+        });
+      }
+      default:
+        return assertNever(operation);
+    }
+  }
+
+  const results: ScopeMutationResult[] = [];
+
+  if (includeWorkspace) {
```
davidebbo, line 211: So include workspace means it applies it at both pod level AND workspace level?

davidebbo: Ok, saw the next block that bails out

davidebbo (approved):
LGTM

front/lib/api/sandbox/admin_pods.ts
smb2268: Yes anything added to a workspace is inherited by every sandbox, then the pod settings are layered on top. So if they have both selected we only need to add it to the workspace level

front/lib/api/sandbox/admin_pods.ts
smb2268: Yep that's right, we have update for API keys but domains it's just delete and add

PR #30999, opened 2026-08-21
feat(sandbox): workspace toggle for agent-requested Computer domains

smb2268:
## Why

Agents running in the Computer can request additional egress domains mid-conversation (`add_egress_domain`), but there's no admin control over whether that's allowed. This adds a workspace-wide toggle to the Computer admin page.

Part of the central Computer admin stack:
- PR1 `30986` — list configured Pods + bulk egress read
- PR2 `30995` — bulk egress write
- PR-A `30996` — scope-aware multi-Pod network view + editing
- PR-B (this PR) — agent-requested-domains toggle
- Follow-up — Pod-member read-only view

## What

`AgentRequestedDomainsSetting`: a workspace-wide `SliderToggle` for `add_egress_domain`. Enabling is confirmation-gated (it lets non-admin users grant per-conversation network access to un-pre-approved domains). Rendered above the Network section so it stays visible regardless of the scope being viewed. Uses the existing `useUpdateWorkspaceSandboxAgentEgressRequests` hook — no new backend.

## Test

`tsc` + `biome` clean. Manual: toggle on (confirm dialog) / off, verify the workspace setting persists.

## Risk

Low. One workspace toggle over an existing endpoint; safe to roll back.

### DIFF
--- a/front-api/routes/w/[wId]/sandbox/egress-policy/pods.test.ts
+++ b/front-api/routes/w/[wId]/sandbox/egress-policy/pods.test.ts
@@ -73,7 +73,7 @@ describe("GET /api/w/:wId/sandbox/egress-policy/pods", () => {
 
     expect(response.status).toBe(200);
     expect(await response.json()).toEqual({
-      pods: [{ sId: podA.sId, name: podA.name }],
+      pods: [{ sId: podA.sId, name: podA.name, isRestricted: true }],
     });
   });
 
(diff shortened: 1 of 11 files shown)


smb2268:
Folding this into `30996`: the Agent-requested-domains toggle moves out of NetworkSection into a standalone AgentRequestedDomainsSetting there (and the now-duplicate toggle is dropped from NetworkSection), so a separate PR isn't needed.

PR #31001, opened 2026-08-23
[pods] Pod app manifest schema

davidebbo:
First PR of the manifest-publish stack. Adds the zod schema for manifest.json at the root of a pod app folder: display name, description, and frames/functions/databases each pointing at a folder-relative path. Deliberately distinct from the archive manifest in pod_app_archive.ts — no files array, explicit names, free paths. No behavior change; nothing consumes the schema yet.

### DIFF
--- a/front/types/api/pod_app_manifest.test.ts
+++ b/front/types/api/pod_app_manifest.test.ts
@@ -1,0 +1,164 @@
+import {
+  isSafePodAppRelativePath,
+  PodAppPublishManifestSchema,
+} from "@app/types/api/pod_app_manifest";
+import { describe, expect, it } from "vitest";
+
+const FULL_MANIFEST = {
+  version: 1,
+  name: "Task List",
+  description: "Track and manage team tasks.",
+  frames: [{ path: "TaskList.tsx" }],
+  functions: [
+    {
+      name: "add-task",
+      path: "functions/add-task.ts",
+      description: "Add a task to the list.",
+      executionMode: "fast",
+      defaultStake: "low",
+    },
+  ],
+  databases: [{ name: "tasks", path: "databases/tasks.db.ts" }],
+};
+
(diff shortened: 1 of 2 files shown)


front/types/api/pod_app_manifest.test.ts
```diff
@@ -0,0 +1,164 @@
...
+  isSafePodAppRelativePath,
+  PodAppPublishManifestSchema,
+} from "@app/types/api/pod_app_manifest";
+import { describe, expect, it } from "vitest";
+
+const FULL_MANIFEST = {
+  version: 1,
+  name: "Task List",
+  description: "Track and manage team tasks.",
+  frames: [{ path: "TaskList.tsx" }],
```
fontanierh: We're going to call the whole thing a "frame" so I think we can't call the UI files "frames"

fontanierh: In fact I don't think we have to list anything here. You just include a index.tsx file and that the root of your frame's UI

front/types/api/pod_app_manifest.test.ts
```diff
@@ -0,0 +1,170 @@
+import {
+  isSafePodAppRelativePath,
+  PodAppPublishManifestSchema,
+} from "@app/types/api/pod_app_manifest";
+import { describe, expect, it } from "vitest";
+
+const FULL_MANIFEST = {
+  version: 1,
```
fontanierh, line 8: What are we using the version for ?

fontanierh (approved):

front/types/api/pod_app_manifest.test.ts
davidebbo: Just standard schema versioning in case we want to evolve the schema.

fontanierh: makes sense 👍

PR #31002, opened 2026-08-23
[pods] publishPodApp business layer

davidebbo:
Second PR of the manifest-publish stack. Adds publishPodApp, which publishes a whole app from its manifest.json: databases reconciled first, then functions, then frames (whose reference validation needs the functions in place), then declarative cleanup — functions carrying the app's prefix that the manifest dropped are unpublished; undeclared databases only produce warnings, never deletions. Per-item failures land in warnings like importPodApp; only invalid manifest, prefix collision and sandbox unavailability abort. The pure plan-builder (declarative diff + path validation) is split into publish_app_plan.ts and unit-tested without mocks. Nothing calls this yet — the MCP tool lands in the next PR.

### DIFF
--- a/front/lib/api/projects/apps.ts
+++ b/front/lib/api/projects/apps.ts
@@ -222,7 +222,7 @@ async function resolveFramesByFolderName(
  * rather than from the sandbox, so listing apps never has to wake a sleeping pod. A database created
  * seconds ago may not be replicated yet.
  */
-async function listPodDatabaseOnDiskNames(
+export async function listPodDatabaseOnDiskNames(
   auth: Authenticator,
   pod: SpaceResource
 ): Promise<string[]> {
--- a/front/lib/api/projects/apps.ts
+++ b/front/lib/api/projects/apps.ts
@@ -28,6 +28,7 @@ import type {
   FileSystemEntry,
   FileSystemFileEntry,
 } from "@app/types/api/file_system/types";
+import { POD_APP_MANIFEST_DB_FILE_SUFFIX } from "@app/types/api/pod_app_manifest";
 import type {
   PodApp,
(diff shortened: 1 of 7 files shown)


front/lib/api/projects/apps.ts
```diff
@@ -614,6 +620,34 @@ function fileEntriesOf(entries: FileSystemEntry[]): FileSystemFileEntry[] {
   return entries.flatMap((entry) => (entry.isDirectory ? [] : [entry]));
 }
 
+/**
+ * Validate a candidate app folder name and derive its prefix: trimmed, within the max length,
+ * free of '/', and holding at least one letter or digit for `normalizeAppPrefix` to derive a
+ * prefix from. Shared by `publishPodApp`, `clonePodApp` and `importPodApp`, which each wrap the
+ * returned `Err` string in their own error class.
+ */
+export function validatePodAppFolderName(
```
davidebbo, line 629: Existing logic moved to a helper

front/lib/api/projects/publish_app.ts
```diff
@@ -0,0 +1,346 @@
...
+  // so they abort — same rule as export).
+  const listResult = await dustFs.list(podRoot);
+  if (listResult.isErr()) {
+    return new Err(
+      new PodAppPublishError("internal", listResult.error.message)
+    );
+  }
+  const folderRelPaths = new Set<string>();
+  const collidingFolderNames = new Set<string>();
+  for (const entry of listResult.value) {
```
davidebbo, line 106: This whole mess is because folder names can collide once they become prefixes for function/db publishing. We could simplify if we change how functions/dbs are published, but that's a different quest.

davidebbo:

PR #31003, opened 2026-08-23
[pods] publish_app MCP tool

davidebbo:
Third PR of the manifest-publish stack. Exposes publishPodApp as a publish_app tool on the sandbox_functions server: single folder argument, pod resolved from the conversation context like every other publish tool. Individual publish/db_reconcile/publish_interactive_content_file tools are untouched and remain the granular path.


davidebbo:

PR #31004, opened 2026-08-23
[pods] Manifest-aware app listing

davidebbo:
Fifth PR of the manifest-publish stack. A pod-root folder containing a manifest.json now lists as an app regardless of the legacy functions//databases/ heuristic, and the listing carries the manifest's display name and description (PodApp gains displayName/description/manifestError, all nullable — backward compatible). A malformed manifest never hides the app: it still lists with a manifestError the tile surfaces as a warning tooltip. The Apps tab tile shows the display name and description.


davidebbo:

PR #31005, opened 2026-08-23
[pods] Manifest-first publishing guidance

davidebbo:
Last PR of the manifest-publish stack. Prose only: the pod_functions skill, the interactive_content instructions and the publish/db_reconcile tool descriptions now teach writing a manifest.json and calling publish_app, with the individual tools kept as the granular path.


davidebbo:

PR #31006, opened 2026-08-23
[pods] Remove the individual function publish and db_reconcile tools

davidebbo:
Inserts after #31003 in the manifest-publish stack. With publish_app in place, the sandbox_functions server's individual `publish` (function) and `db_reconcile` tools are removed — the manifest is now the only way to publish functions and reconcile databases. The individual Frame publish (`publish_interactive_content_file`) stays, as do unpublish/call/get/list/inspect_invocations and the db read tools; the lib functions (publishSandboxFunction, reconcileDatabaseFromPodPath) remain since publish_app, clone and import use them.

Fallout handled here: agent-facing prose (pod_functions skill, tool descriptions) now teaches publish_app only, with egress domains routed through request_egress_domain (the removed publish tool's `domains` argument was the previous path); the SandboxFunctionPublishValidationDetails UI branch is deleted (it referenced the removed tool name in a strongly-typed call) — historical conversations still render past publish/db_reconcile actions from their persisted displayLabels snapshot; the mcp_servers_metadata snapshot is regenerated, which also incidentally adds the publish_app entry that #31003 had missed.


davidebbo:

PR #31008, opened 2026-08-24
[front] fix: open agent details modal instead of builder link for similar agents

avervaet:
## Description

The similar-agents link in the agent builder previously renders blank for agents the user can view but not edit. Clicking a similar agent now opens the same read-only view used in the assistant picker, which works for any agent visible to the user.

This also moved from the settings block to the instructions block, and now watches the agent's instructions instead of its description to decide when to trigger the similarity check. 

Fetching also skips re-running the check when the debounced instructions settle back to a value already checked (e.g. after an undo/redo).

https://github.com/user-attachments/assets/c0978fed-7e9e-445e-b6d8-97ba018b6af1

## Tests

Updated the existing unit tests for `AgentBuilderSimilarAgentsSection` to cover watching instructions instead of description, the new debounce/dedup behavior, and rendering the `AgentDetailsSheet` instead of an external link. Manually verified clicking a similar agent opens its details in a read-only sheet for agents the user can view but not edit.

## Risk

Low risk, scoped to the agent builder's similar-agents UI. No API or data model changes.

## Deploy Plan

Deploy front

### DIFF
--- a/front/components/agent_builder/AgentBuilderSimilarAgentsSection.tsx
+++ b/front/components/agent_builder/AgentBuilderSimilarAgentsSection.tsx
@@ -1,17 +1,16 @@
 import { useAgentBuilderContext } from "@app/components/agent_builder/AgentBuilderContext";
 import type { AgentBuilderFormData } from "@app/components/agent_builder/AgentBuilderFormContext";
+import { AgentDetailsSheet } from "@app/components/assistant/details/AgentDetailsSheet";
 import { useDebounceWithAbort } from "@app/hooks/useDebounce";
 import { useFeatureFlags } from "@app/lib/auth/AuthContext";
-import { LinkWrapper } from "@app/lib/platform";
 import {
   useAgentConfigurations,
   useSimilarAgents,
 } from "@app/lib/swr/assistants";
-import { getAgentBuilderRoute } from "@app/lib/utils/router";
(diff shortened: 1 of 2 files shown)


avervaet:

- #31008 ◀

front/components/agent_builder/AgentBuilderSimilarAgentsSection.tsx
```diff
@@ -89,17 +90,30 @@ export function AgentBuilderSimilarAgentsSection({
...
+        dispatch({ type: "fetch_settled" });
+        return;
+      }
+      lastFetchedInstructionsRef.current = currentInstructions;
+
       dispatch({ type: "fetch_start" });
 
       try {
-        const result = await getSimilarAgents(naturalDescription, { signal });
+        const result = await getSimilarAgents(currentInstructions, {
```
fabiencelier: I see we already compare to agents instructions in the backend, but in the prompt sent to the LLM we still give example of description.

https://github.com/dust-tt/dust/blob/88623d635d0fd93f17a52dc3b9f3d9666159d4c0/front/lib/api/assistant/existing_agent_checker.ts#L66-L76

Maybe we should update the examples to match the fact that they receive instructions ?
Maybe we should also explain that in the prompt too

fabiencelier (approved):
LGTM on the UI part

PR #30981, opened 2026-08-21
Move consumption exports to the tmp-workloads bucket

avervaet:
## Description

Moves consumption exports (#30557) from `private-uploads` to the existing `tmp-workloads` bucket, which already deletes everything after 7 days. No infra change needed (dust-tt/dust-infra#983 closed).

- Adds `getGcsTmpWorkloadsBucket()` / `getTmpWorkloadsBucket()`, following the existing per-bucket helper pattern.
- Keeps the top-level `consumption_exports/{workspaceId}/...` prefix to avoid colliding with other workloads writing into the same shared bucket.

## Tests

Existing unit tests updated to use the new bucket helper; all pass.

## Risk

Low. Old exports under the previous `private-uploads` path become unreachable but harmless. Export retention drops from the originally-planned 15 days to `tmp-workloads`' 7-day TTL.

## Deploy plan

**Needs a `DUST_TMP_WORKLOADS_BUCKET` secret added for `front`** before deploy (GSM, created out-of-band like the other bucket-name vars). No new IAM needed — `front` already has project-level `storage.objectUser`.

PR #31009, opened 2026-08-24
Release consumption analytics

flvndvd:
## Description

Now that analytics consumption have been backfilled, this PR releases it. It replaces the analytics menu item in the side bar, the legacy page is still accessible in case customer needs it, TBH we will probably kill it in a few days anyway.

As we backfilled more than 90 days, this PR also re-adds the "Last 90 days" in both analytics and automation page.


front/lib/analytics/consumption_period.test.ts
```diff
@@ -0,0 +1,8 @@
+import { CONSUMPTION_PERIOD_DAY_OPTIONS } from "@app/lib/analytics/consumption_period";
+import { describe, expect, it } from "vitest";
+
+describe("consumption period options", () => {
+  it("includes the last 90 days preset", () => {
+    expect(CONSUMPTION_PERIOD_DAY_OPTIONS).toContain(90);
+  });
+});
```
adrsimon, line 8: 👀

adrsimon (approved):
🚀

avervaet (approved):
LGTM. Can we just review and release https://github.com/dust-tt/dust/pull/30981 first so we don't pollute GCS?

PR #30214, opened 2026-08-07
[front] enh: show agent questions in the input bar

aubin-tchoi:
## Description

Context in [thread](https://dust4ai.slack.com/archives/C0B1VFWD0TW/p1786117382597449?thread_ts=1786115305.835439&cid=C0B1VFWD0TW).

This PR moves the ask user question tool's option selection to the bottom of the screen, replacing the input bar.
This prevents having an idle input bar that creates confusion next to the free text input. Makes the blocking action actually blocking.

## Tests

- Added tests.
- Tested locally.

## Risk

- Low. UI-only change, safe to rollback.

## Deploy Plan

- Deploy front.

PR #31012, opened 2026-08-24
fix: extension scroll

matteotrab:
## Description

Fixes the conversation input bar drifting away from the bottom of the Chrome extension panel: once the last message was reached, the message list kept scrolling into empty space and the sticky input bar was carried up and out of view.

Root cause chain:

- The extension renders an invisible `position: fixed` box inside the input bar to anchor its "+" dropdown (`DropdownAnchorTrigger`), placed with viewport coordinates read from `getBoundingClientRect()`.
- #30214 wrapped the input bar content in a framer-motion `motion.div` whose `animate` target is a raw `transform` string. Framer writes that string into the style attribute and leaves it there once the animation settles, so the element permanently carried an identity `translateY`.
- Any `transform` other than `none` makes an element a containing block for `position: fixed` descendants. The anchor's viewport coordinates were therefore resolved against the input bar wrapper instead of the window, placing it well below the bottom of the conversation scroller.
- An out-of-flow box sitting below a scroller's content box extends that scroller's scrollable overflow, so the conversation gained scroll range that the message list itself did not account for.
- The input bar is `position: sticky; bottom: 0`, and a sticky element is clamped to its parent's content box. It could not follow the scroll past that box, so it was left behind and scrolled off the top.

The fix animates `y` (a motion value) instead of a raw `transform` string, and uses `transformTemplate` to emit `transform: none` whenever `y` is at rest. The enter/exit motion is unchanged — the element simply stops holding an identity transform while idle, so the anchor goes back to resolving against the window and contributes nothing to the scroller.

Before

https://github.com/user-attachments/assets/31fe1de0-d2e2-4e8c-b81a-6a2ea11963f9

After

https://github.com/user-attachments/assets/6040d26f-c09f-4326-83ae-cc701e71a9a6

## Tests

Manually.

## Risk

Low.

## Deploy Plan

Standard deployment.

### DIFF
--- a/front/components/assistant/conversation/AgentInputBar.tsx
+++ b/front/components/assistant/conversation/AgentInputBar.tsx
@@ -48,18 +48,31 @@ import {
   useVirtuosoLocation,
   useVirtuosoMethods,
 } from "@virtuoso.dev/message-list";
-import type { Transition } from "framer-motion";
+import type { MotionProps, Transition } from "framer-motion";
 import { AnimatePresence, motion, useReducedMotion } from "framer-motion";
 import { useEffect, useMemo, useRef, useState } from "react";
 
 const MAX_DISTANCE_FOR_SMOOTH_SCROLL = 2048;
 const DOUBLE_ESC_WINDOW_MS = 300;
 
-const INPUT_BAR_SWAP_TRANSFORMS = {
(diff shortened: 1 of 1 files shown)


matteotrab:

- #31012 ◀

aubin-tchoi (approved):
LGTM 🙏🏻

PR #31014, opened 2026-08-24
Fix Poke feature flag search

flvndvd:
## Description

Current search of FF in poke is really bad, you get plenty of noise despite exact match. The search is currently handed over to `cmkd` which runs a very permissive search. This PR fixes it by indexing human-readable versions of flag names. Searches containing spaces, such as “models picker,” now correctly match underscore-separated values like `models_picker`.


flvndvd:
PMRR.

adrsimon (commented):
LGTM

PR #31015, opened 2026-08-24
[front] Store microCredits in fixed-window spend-cap counters

tdraier:
## What

Store **microCredits** (credits × 1,000,000, integer) in the three **fixed-window** spend-cap Redis counters — **per-user**, **per-API-key**, and **programmatic** — instead of whole credits, so the counters survive an upcoming switch to fractional credits. This mirrors `AgentMessageConsumptionItemModel`, which already stores microCredits. Reuses the helpers in `front/lib/credits/units.ts` (`roundCreditsToMicroCredits` / `microCreditsToCredits`).

## Changes

- **Key rename:** the three spend-limit key builders now emit `...:spend_limit_awu_microcredit_count` / `...:programmatic_spend_limit_awu_microcredit_count` so old credit-scale counters aren't mixed with the new microcredit values after deploy. Old keys expire; **the counters reseed from Elasticsearch (in microCredits) on the first read after deploy** via the existing lazy seed.
- **Write boundaries:** recorders (`recordUserSpendLimitUsage`, `recordApiKeySpendLimitUsage`, `recordProgrammaticSpendLimitUsage`) now accept **fractional** credits — the old non-integer error log is removed (fractional is now valid), the guard is `!Number.isFinite || <= 0` — and convert via `roundCreditsToMicroCredits` before the INCRBY.
- **Lazy seed / resync:** `fetchSeedValue` and the `setFixedWindowCount` resyncs convert the ES credit value to microCredits, preserving the `null` "ES read failed → skip seed" contract (never seeded as 0).
- **Read boundaries:** enforcement scales the credit cap up to microCredits (`count >= roundCreditsToMicroCredits(cap)`) for an integer-on-integer comparison; the two Poke reads convert microCredits back to credits so ES/RL/MT stay in the same unit.
- **Poke display:** added `formatCreditsPrecise` (up to 6 decimals, trailing zeros trimmed) and used it for the ES/RL/MT figures so a microcredit-derived fractional value (e.g. `826.5`) renders correctly instead of being integer-rounded. The shared `formatCredits` is untouched.

## Out of scope / follow-ups

- The **fair-use rolling limiter** (`addRateLimiterCount` / `makeFairUseAwuCreditsRateLimitKeyForUser`) is intentionally **excluded**: its Lua does one `ZADD` per unit in a loop, so multiplying by 1e6 would write a million entries.
- The **workspace usage cap (#30655)** is not on `main` yet; it will need the **same microCredits treatment** when it lands.

## Tests

- Added a runnable Redis-backed test in `rate_limiter.test.ts` proving a fractional `2.5`-credit delta is recorded as `2_500_000` microCredits and that enforcement blocks once the counter reaches `cap × 1_000_000`.
- `units.test.ts` already covers the conversion helpers.
- The DB-backed spend-limit integration tests could not run in this environment due to a pre-existing missing `@novu/framework` dependency (unrelated to this change); the Redis/units-level tests pass.

### DIFF
--- a/front/components/poke/credits/PokeMembersUsageTable.tsx
+++ b/front/components/poke/credits/PokeMembersUsageTable.tsx
@@ -4,7 +4,7 @@ import { GrantFreeCreditsButton } from "@app/components/poke/credits/GrantFreeCr
 import { ReconcileCreditStateButton } from "@app/components/poke/credits/ReconcileCreditStateButton";
 import { PokeDataTable } from "@app/components/poke/shadcn/ui/data_table";
 import type { MemberUsageType } from "@app/lib/api/credits/members_usage";
-import { formatCredits } from "@app/lib/client/credits";
+import { formatCredits, formatCreditsPrecise } from "@app/lib/client/credits";
 import type { MetronomeAlertRef } from "@app/lib/metronome/alerts/types";
 import { getMetronomeAlertUrl } from "@app/lib/metronome/urls";
 import { usePokeMembersUsage } from "@app/poke/swr/credits";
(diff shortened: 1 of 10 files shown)


fabiencelier (approved):
LGTM

tdraier:

- #31015 ◀
    - #31021
        - #31027
        - #31024

PR #31017, opened 2026-08-24
[front] feat: add AgentMessage pause columns and fixed workflow-alert threshold

avervaet:
## Stack Overview

5-PR stack adding a pause/resume flow for the spend checkpoint: when a user's spend crosses a fixed threshold, the agent loop now stops and waits for an explicit decision instead of just sending a notification.

**Video:**

https://github.com/user-attachments/assets/0a7e7442-9908-483d-a3b2-7fe8b2585343

## Description

Adds `creditSpendCheckpointStatus` on `AgentMessage` ("paused" / "acknowledged") to track a paused run per message. It's needed because state has to survive both a Temporal workflow relaunch and a browser refresh.

A 200 threshold value was selected as this mean it would render for ~5% of the messages. An acceptable threshold to educate users. And we will compute the threshold only based on llm credits. 

**Credit distribution analysis**
Capture d’écran 2026-09-11 à 10 09 40

## Tests

Schema and configuration only; exercised indirectly by the rest of the stack. Migration applied and verified locally.

## Risk

Low: additive, nullable state and a new configuration value.

## Deploy Plan

- Apply pre-deploy migration
- Deploy front


avervaet:

- #31017 ◀
    - #31018
        - #31019
            - #31020
                - #31060

front/lib/api/config.ts
```diff
@@ -4,6 +4,10 @@ import { EnvironmentConfig } from "@app/types/shared/utils/config";
 
 const PRODUCTION_DUST_API = "https://dust.tt";
 
+// Same for every user/workspace: how much a user can spend in a billing cycle before the agent
+// loop pauses and asks them to confirm they want to keep going.
```
avervaet: Placeholder for now will run queries to determine the best threshold. 
We don't want to block legitimate query but avoid user to burn themselves.

id13:
@avervaet This means we flip the approval permanently per conversation, I think it's better to do it per agentMessage  - just because I approved one message to go as far as I want does not mean I imply it should on every subsequent message, I think it's also easier to explain to users 😄 

Also why can't it be simply an event of type "blocked_validation_required" ?

You can make it so that when we start a new agent-loop after the approval, we pass a new param to the workflow telling that this loop has full budget allowance.

> replacing the previous per-group admin setting.

Are we sure we want to do this ? Seems like a regression no ?

front/lib/models/agent/conversation.ts
```diff
@@ -500,6 +500,11 @@ export class AgentMessageModel extends WorkspaceAwareModel<AgentMessageModel> {
   declare prunedContext: boolean | null;
   declare costCredits: number | null;
 
+  // "paused": loop stopped after crossing the threshold, waiting on the user.
+  // Declining sets the message status to "gracefully_stopped"
+  // "acknowledged": user confirmed continuing, kept for the rest of the message's lifetime.
+  declare workflowAlertThresholdStatus: "paused" | "acknowledged" | null;
```
id13: If we continue this way (see my other comment) we def need another name. What alert are we talking about here ?

avervaet: Renamed `workflowAlertThresholdStatus` → `spendCheckpointStatus` (and the rest of the stack accordingly: config key, gate/activity functions, the /workflow-alert-threshold endpoint → /spend-checkpoint, UI hook/card). "Spend cap" is already taken by the existing high, $-denominated, programmatic-usage-inclusive limit, so "spend checkpoint" for this lower, conversation-scoped pause-to-confirm keeps the two distinct.

avervaet:
@id13 addressing the three points from your comment:

- **Naming**: renamed `workflowAlertThreshold*` → `spendCheckpoint*` across the whole stack (this PR through #31060) — see reply on the inline comment.
- **Scope of the acknowledgment**: confirmed it's already message-scoped, not conversation-scoped — `spendCheckpointStatus` lives on `AgentMessageModel` and `checkSpendCheckpointActivity` only reads it for the current `agentMessageId`. Acknowledging grants unlimited spend for the rest of that one message's agent loop, not for the whole conversation going forward.
- **Event style**: agreed, will switch to a proper blocking event modeled on `tool_approve_execution` instead of the current silent passthrough — tracked for #31018.
- **Regression concern**: dropping. The per-group admin setting was never published, so there's no live behavior to regress from.

front/lib/api/config.ts
```diff
@@ -58,6 +59,27 @@ const config = {
     return durationSeconds * 1_000;
   },
 
+  getCreditSpendCheckpointThresholdAwuCredits: (): number => {
+    const value = EnvironmentConfig.getOptionalEnvVariable(
```
flvndvd: Let's avoid env variable for this value, we should mainly use it for obfuscation purpose, here it's nice to have history and blame IMHO.

front/lib/constants/credits.ts
```diff
@@ -0,0 +1,2 @@
+// Credits a single message may spend before the loop pauses and asks the user whether to continue.
+export const DEFAULT_CREDIT_SPEND_CHECKPOINT_THRESHOLD_AWU_CREDITS = 1000;
```
flvndvd: As seen IRL, can we check per plan type, I'd assume that pro and enterprise might not have the same P90 of credits used per message.

flvndvd (commented):
Dropping first comments, reviewing how it integrates to build conviction around the data model.

avervaet:
@flvndvd Added a video, in the descriptions will run queries on the db read only replica to build convictions on the thresholds.

front/lib/constants/credits.ts
```diff
@@ -0,0 +1 @@
+export const CREDIT_SPEND_CHECKPOINT_THRESHOLD_AWU_CREDITS = 300;
```
avervaet: Will stick with a single value so we don't need workspace knowledge to check if we need to trigger an activity.

300 was chosen as this mean it would render for ~2% of the messages. An acceptable threshold to educate users.

flvndvd (approved):

PR #31018, opened 2026-08-24
[front] feat: pause the agent loop at the workflow-alert threshold

avervaet:
## Description

A single agent message can silently burn through a large amount of credits, with subagents multiplying the spend and no moment where the user gets to weigh in. The agent loop now pauses once a root message's cumulative LLM spend, subagent tree included, reaches a fixed 200 AWU credits, and records that the user needs to decide whether to continue.

The spend used is the one the per-step guardrail already computes, so no new query runs on the hot path. The check itself is a pure decision activity (`checkCreditSpendCheckpointActivity`) that is only scheduled once the pre-step spend has reached the threshold; persisting the pause happens in a dedicated finalize, so the message can never read "paused" while its loop is still running, and a stop or cancel that lands around the pause wins. Only web and extension messages can pause: other origins and subagent messages have no one to answer, and a message the user already acknowledged is never asked again.

Tool cost is deliberately not counted to avoid adding new db call to the agetn loop. A prod sample of messages crossing the threshold puts LLM spend at about 96% of the total, so this is an accepted approximation.

## Tests

* Threshold and exemption rules: reached at exactly the threshold, exempt without a user or with a non-resumable origin.
* Checkpoint activity: skips when exempt, acknowledged, a subagent message, or the message is gone; crossed for a pausable root message.
* Finalize: marks the message paused and flags the conversation, leaves an already finalized message untouched.
* Workflow: no activity scheduled below the threshold or without the patch marker, breaks and finalizes as paused when crossed, rethrows non-activity errors.

## Risk

Medium. Every root web or extension message on every plan can now pause, where credit checks previously applied only to credit-priced plans. A wrong decision pauses a message that should keep running; a crashing check fails open and never pauses one that should. Single revert to roll back, and the patch marker keeps in-flight workflows replaying deterministically.

## Deploy Plan

Deploy alongside #31019 and #31020 together so users can act on the pause.


avervaet:

- #31017
    - #31018 ◀
        - #31019
            - #31020
                - #31060

front/lib/resources/conversation_resource.ts
```diff
@@ -835,6 +835,55 @@ export class ConversationResource extends BaseResource<ConversationModel> {
     };
   }
 
+  static async fetchAgentMessageCreditSpendCheckpointState(
```
flvndvd: `nit`

```suggestion
  static async fetchCreditSpendCheckpointStateForAgentMessage(
```

front/temporal/agent_loop/activities/credit_check.ts
```diff
@@ -14,3 +28,130 @@ export async function checkCreditsActivity(
...
+
+const NOT_CROSSED: CreditSpendCheckpointActivityResult = {
+  crossed: false,
+  skipRemainingChecks: false,
+};
+
+/**
+ * Has this agent message's own spend crossed the credit spend checkpoint?
+ * When it has, the pause is persisted on the message and the user is notified.
+ */
```
flvndvd, line 58: Slightly worry that this contribute to worsen the agent loop performance which have been very much slow down lately. @id13 work should help get this information from Redis to avoid doing many queries on our DB.

front/temporal/agent_loop/activities/finalize.ts
```diff
@@ -98,6 +98,38 @@ export async function finalizeGracefullyStoppedAgentLoopActivity(
   ]);
 }
 
+/**
+ * Credit spend checkpoint pause: the message is not done, it is waiting on the user's decision.
+ * Like the blocked-action pause, this execution's runs must still be accounted for (credits,
+ * analytics, usage), since a later execution only accounts for its own runs. Completion-only
+ * side effects (unread notification, mentions, email reply) wait for the final execution.
+ */
+export async function finalizePausedAgentLoopActivity(
```
flvndvd: Why aren't we re-using the one use by a blocking/pending tool?

flvndvd (commented):
Not convinced about the way we integrate this, a few concerns that spread from performances to module abstraction.

If we think about what we need to do here, it's pretty similar to two existing pieces:
1. pausing agent loop for a reason (mainly driven by tools today)
2. Forcing the model to pause on the next iteration. Today, this is enabled when the model reaches the last step of the loop (IIIRC the 63th) so it stops running tools and instead finalize current message.

Can we try to align closer to those existing primitives?

On the performance piece, our data model is not friendly to do any of this check efficiently, it quickly requires pulling data from many different places. @id13 is pushing a refactoring so this is always available in a redis and can be check quickly (ideally in an existing activity so it does not yet another Temporal roundtrip).

front/types/assistant/conversation.ts
```diff
@@ -257,6 +257,18 @@ export function isHiddenMessageOrigin(origin: UserMessageOrigin): boolean {
   return HIDDEN_MESSAGE_ORIGINS.includes(origin);
 }
 
+// Origins whose author is in a Dust conversation UI, where an agent that stops to ask something
+// can be seen and answered. Every other origin (API, integrations, email, triggers, hidden
+// messages) is unattended: nobody is there to resume the agent.
+export const ATTENDED_MESSAGE_ORIGINS: UserMessageOrigin[] = [
```
avervaet: So we only handle user facing behavior for now and could expand the scope in the future

front/lib/api/assistant/credit_check.ts
```diff
@@ -53,3 +55,34 @@ export async function checkPoolCreditGate(
 
   return DO_NOT_STOP;
 }
+
+const CREDIT_SPEND_CHECKPOINT_RESUMABLE_ORIGINS: UserMessageOrigin[] = [
+  "web",
+  "extension",
+];
+
+export function isCreditSpendCheckpointExempt(
```
flvndvd: `isExemptFromCreditSpendCheckpoint` would read more naturally.

front/lib/resources/conversation_resource.ts
```diff
@@ -836,6 +836,78 @@ export class ConversationResource extends BaseResource<ConversationModel> {
     };
   }
 
+  /**
+   * Loads what the credit spend checkpoint check needs to decide whether it applies: the agent
+   * message's checkpoint status and whether its triggering user message is a root (non-agentic)
```
flvndvd: Not sure to get the "non-agentic" wording here.

front/lib/resources/conversation_resource.ts
```diff
@@ -836,6 +836,78 @@ export class ConversationResource extends BaseResource<ConversationModel> {
...
+   * message. Returns null when either message cannot be found.
+   */
+  static async fetchCreditSpendCheckpointContextForAgentMessage(
```
flvndvd: Could we use the existing `getAgentLoopRuntimeDataWithAuth`?

front/temporal/agent_loop/workflows.ts
```diff
@@ -362,6 +390,41 @@ export async function agentLoopWorkflow({
...
+        // The threshold is decided on the spend measured before this step ran. The step itself
+        // may have crossed it, in which case the pause lands one step late: accepted to avoid
+        // scheduling the activity on every step of every message.
+        if (
+          patched("credit-spend-checkpoint-gate") &&
+          !skipCreditSpendCheckpointChecks &&
+          preStepReachedCreditSpendCheckpoint
+        ) {
+          try {
+            const checkpointResult = await checkCreditSpendCheckpointActivity(
```
flvndvd: This feels more like enforce than check, right? The check is owned by the workflow so we conditionally call this activity only when condition is met.,

flvndvd: Actually I'm a bit confused on the behavior here. Any chance we could make this self contained in a dedicated module? It feels like we are splitting the work between previous step + eventually in `checkCreditSpendCheckpointActivity`?

front/lib/api/assistant/credit_check.ts
avervaet: Agree

front/lib/resources/conversation_resource.ts
avervaet: Agree it's unclear. isRootAgentMessage means "the triggering user message wasn't itself generated by another agent". Rewording the docstring.

front/temporal/agent_loop/workflows.ts
avervaet: The check reached logic must run inside the step activity to see fresh cost data. 
But we could move hasReachedCreditSpendCheckpoint and checkCreditSpendCheckpointActivity in a dedicated module.

front/lib/resources/conversation_resource.ts
avervaet: Agree it's unclear. isRootAgentMessage means "the triggering user message wasn't itself generated by another agent". Rewording the docstring.

front/lib/resources/conversation_resource.ts
avervaet: would push back on that. 
getAgentLoopRuntimeDataWithAuth loads many things (the entire conversation object, the full message rows, content blocks...) As this will be triggered at every step after the threshold is reached we want to avoid slowing the loop with unecessary db calls.

front/temporal/agent_loop/workflows.ts
flvndvd: Seen IRL. End goal honor the assumption of credit spent pause here only accounting for LLM credits. Move this to `runLLMAndCreateActionActivity` to consolidate in one deep module to avoid the cost of calling other activities and cluttering too much the main workflow code.

front/lib/api/assistant/credit_spend_checkpoint.test.ts
```diff
@@ -0,0 +1,139 @@
...
+import type { Authenticator } from "@app/lib/auth";
+import { CREDIT_SPEND_CHECKPOINT_THRESHOLD_AWU_CREDITS } from "@app/lib/constants/credits";
+import { MODEL_COST_MICRO_USD_PER_AWU_CREDIT } from "@app/lib/metronome/constants";
+import { describe, expect, it } from "vitest";
+
+// Minimal stand-in for the Authenticator class exposing only the members the checkpoint logic
+// reads. A class instance can't be constructed structurally, so a single `as unknown as` is the
+// standard test-mock escape here (see the same pattern across the suite); the cast surface is
+// kept to this one spot.
+function makeAuth({
```
flvndvd: You can use the `setup` factory we have 🏴‍☠️

front/temporal/agent_loop/activities/common.ts
```diff
@@ -943,3 +944,63 @@ export async function finalizeCreditStop(
...
+        conversationId: conversation.sId,
+        messageStatus: agentMessage.status,
+      },
+      "[CreditSpendCheckpoint] message already finalized, skipping pause"
+    );
+    return;
+  }
+
+  await ConversationResource.markAgentMessageCreditSpendCheckpointPaused(auth, {
+    agentMessageModelId: agentMessage.agentMessageId,
```
flvndvd: Can we pass the `agentMessage` directly, and check the agentMessage status in the function (just in case).

front/temporal/agent_loop/activities/run_model_and_create_actions_wrapper.ts
```diff
@@ -244,7 +252,16 @@ async function _runModelAndCreateActionsActivity({
     return null;
   }
 
-  // Tool test run: bypass LLM and directly execute tool commands.
+  const creditSpendCheckpointCrossed = await getCreditSpendCheckpointCrossed({
+    auth,
```
flvndvd, line 255: `nit`
Lift `auth` at the root, not nested in an object.

front/temporal/agent_loop/activities/run_model_and_create_actions_wrapper.ts
```diff
@@ -339,9 +362,67 @@ async function _runModelAndCreateActionsActivity({
...
+  if (
+    isExempt ||
+    !isRootAgentMessage ||
+    !hasReachedCreditSpendCheckpoint({ totalCostMicroUsd })
+  ) {
+    return false;
+  }
+
+  let status: AgentMessageModel["creditSpendCheckpointStatus"] = null;
+  try {
```
flvndvd: Not sure I'd personally wrap it with try/catch block here, looking at `ConversationResource.fetchAgentMessageCreditSpendCheckpointStatus` it looks like sole failure mode there is really the DB which I'd argue is fine here. If DB dies, not sure it's going to go anywhere anyway.

front/temporal/agent_loop/workflows.ts
```diff
@@ -362,6 +371,13 @@ export async function agentLoopWorkflow({
           creditStopRequested = true;
           break;
         }
+
+        // The decision is made on the spend measured before this step ran. The step itself may
+        // have crossed it, in which case the pause lands one step late.
+        if (creditSpendCheckpointCrossed) {
```
flvndvd: You'll need to use the `patched` API here to surrounds this change.

flvndvd (approved):
LGTM

PR #31019, opened 2026-08-24
[front] feat: continue/decline endpoint for the workflow-alert-threshold pause

avervaet:
## Description

Adds the `spend-checkpoint` endpoint that lets a user resolve a paused message: continue past the checkpoint, or decline.

- Continuing relaunches the agent loop from where it paused (derived from the message's existing step contents, no separate step is stored).
- Declining finalizes the message as gracefully stopped, with a short generated summary of the progress made so far, so the conversation still ends with a useful reply.
- The paused status is now included when a message is loaded, so the pause survives a page refresh ahead of the UI changes in #31020.

## Tests

Covered by the existing suite, plus manual end-to-end verification of both the continue and decline paths.

## Risk

New, additive endpoint with no existing route changes. Resolving the same pause twice (e.g. a double click) is a safe no-op. No UI caller until #31020.

## Deploy Plan

After #31017 and #31018. Deploy together with #31020 so the endpoint has a caller as soon as it's live.


avervaet:

- #31017
    - #31018
        - #31019 ◀
            - #31020
                - #31060

front/lib/api/assistant/conversation/credit_spend_checkpoint_pause.ts
```diff
@@ -0,0 +1,264 @@
+import { canCurrentUserRespondToParentUserMessage } from "@app/lib/api/assistant/conversation/can_current_user_respond";
+import type { Authenticator } from "@app/lib/auth";
+import { DustError } from "@app/lib/error";
+// TODO(2026-07-31 QOS): move these message fetches behind a resource method instead of using
+// models directly in lib/api.
+import {
```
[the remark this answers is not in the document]
avervaet: check with @id13

front/lib/api/assistant/conversation/credit_spend_checkpoint_pause.test.ts
```diff
@@ -0,0 +1,261 @@
...
+      });
+    const { messageRow: agentMessageRow } =
+      await ConversationFactory.createAgentMessage(auth, {
+        workspace,
+        conversation: created,
+        agentConfig,
+        parentMessageModelId: userMessageRow.id,
+        rank: 1,
+      });
+    agentMessageSId = agentMessageRow.sId;
```
flvndvd: Can we avoid the `SId` suffix it should be either id (sId) or ModelId.

front/lib/api/assistant/conversation/credit_spend_checkpoint_pause.ts
```diff
@@ -0,0 +1,266 @@
...
+
+  const launchRes = await launchAgentLoopWorkflow({
+    auth,
+    agentLoopArgs,
+    startStep: nextStep(agentMessage),
+    // Avoid racing with the workflow that just paused: wait for its run to be reported done
+    // before starting the resumed one.
+    waitForCompletion: true,
+  });
+  if (launchRes.isErr()) {
```
flvndvd, line 150: IIRC error here is exclusively the workflow already running, no?

front/lib/api/assistant/conversation/credit_spend_checkpoint_pause.ts
```diff
@@ -0,0 +1,266 @@
...
+    agentMessage,
+    conversation,
+    step,
+  }: {
+    agentMessage: AgentMessageType;
+    conversation: ConversationType;
+    step: number;
+  }
+): Promise<void> {
+  const summaryRes = await generateSmoothShutdownSummary(auth, conversation);
```
flvndvd: Can we leverage the agent loop here, either we do what we do when the loop reaches the last step (and remove all tools + small prompt). Or add a new "message" so the model knows. But ideally no direct LLM call outside of the event loop.

avervaet: Yes, added startAsToolFreeGracefulStop to the agent loop workflow so we can force resuming in a state with no tool call allowed and finalize regardless of it's step outcome.

avervaet:
@flvndvd all good fo re-r?

front/lib/api/assistant/conversation/credit_spend_checkpoint_pause.ts
```diff
@@ -0,0 +1,210 @@
...
+    );
+  if (!applied) {
+    logger.info(
+      { agentMessageId: agentMessage.sId, conversationId: conversation.sId },
+      "Spend checkpoint pause already resolved"
+    );
+    return new Ok(undefined);
+  }
+
+  await finalizeAgentMessagesWithoutWorkflow(auth, {
```
flvndvd, line 201: I'm not sure to understand how this works? It seems that it signals the event loop to pause, but here agent loop was not resumed, right? What about the soft-pause?

flvndvd: Seen IRL, actually change of plans, no more last LLM loop, just cancel.

flvndvd (approved):

PR #31020, opened 2026-08-24
[front] feat: inline paused-at-threshold card, replacing the modal dialog

avervaet:
## Description

When a single message consume more than threshold credits display a cards offering a continue/stop choice

Also carries the client signal for the pause, moved here from #31018 so the event and its consumer are reviewed together: the `agent_credit_spend_checkpoint_reached` event type, its publish in the paused finalize, and its drop at the v1 boundary so the public API and SDK never see it.

**Visual:**
Capture d’écran 2026-09-04 à 16 40 07

## Tests

Manually verified end-to-end: the loop pauses and the card renders, persists across a refresh, and resolves correctly on both continue and decline.

## Risk

Mostly UI. The new stream event is internal: it is dropped at the v1 boundary, so no public API or SDK surface changes.

## Deploy Plan

Deploy after #31017, #31018, #31019 are live; #31060 follows on top of this one.


avervaet:

- #31017
    - #31018
        - #31019
            - #31020 ◀
                - #31060

front/components/assistant/conversation/AgentMessage.tsx
```diff
@@ -1261,6 +1263,25 @@ function AgentMessageContent({
...
+    agentMessage.pausedAtCreditSpendCheckpoint &&
+    agentMessage.status === "created" ? (
+      <CreditSpendCheckpointPausedCard
+        owner={owner}
+        conversationId={conversationId}
+        messageId={sId}
+        triggeringUser={triggeringUser}
+        // The real, billed cost (persisted by the same finalize activity that paused the
+        // loop), not the fixed threshold it crossed. Null until that finalize activity runs.
+        creditsUsed={
```
avervaet, line 1282: The real, billed cost (persisted by the same finalize activity that paused the loop), not the fixed threshold it crossed. Null until that finalize activity runs.

front/components/assistant/conversation/CreditSpendCheckpointPausedCard.tsx
```diff
@@ -0,0 +1,116 @@
...
+    if (!success) {
+      setSubmittingDecision(null);
+    }
+  };
+
+  return (
+    <Card
+      variant="secondary"
+      containerClassName="w-full max-w-xl"
+      className="flex flex-col shadow gap-4"
```
aubin-tchoi, line 102: we should extract a component to share with the validation and auth at some point

front/components/assistant/conversation/AgentMessage.tsx
```diff
@@ -1261,6 +1263,22 @@ function AgentMessageContent({
...
+    agentMessage.pausedAtCreditSpendCheckpoint &&
+    agentMessage.status === "created" ? (
+      <CreditSpendCheckpointPausedCard
+        owner={owner}
+        conversationId={conversationId}
+        messageId={sId}
+        triggeringUser={triggeringUser}
+        creditsUsed={
+          agentMessage.costCredits !== null
+            ? agentMessage.costCredits + (agentMessage.subAgentCostCredits ?? 0)
```
aubin-tchoi, line 1284: this still does not take in account the cost induced by sub agents correctly IIRC, so you may end up showing a value that is very low, which would be misleading

front/components/assistant/conversation/CreditSpendCheckpointPausedCard.tsx
```diff
@@ -0,0 +1,116 @@
...
+    >
+      <div className="flex items-center gap-2">
+        <Avatar icon={PieChart01} size="sm" />
+        <div className="heading-base">Keep going?</div>
```
aubin-tchoi: I think I'd frame the title + beginning of the text to explain what's happening and why this deserves attention, then transition to what actions can be taken rather than the other way around

aubin-tchoi: would lead with smth that explained that we paused the convo to avoid excessive credit consumption

aubin-tchoi (approved):
LGTM 💪🏻 
FYI you'll need to update other consumers (chrome extension and slack bot) at some point

PR #31021, opened 2026-08-24
[front] Switch fair-use rolling-window AWU counters to weighted microCredits

tdraier:
## Context

Stacked on `spend-limit-rate-cap-microcredits` (#31015), which switched the three FIXED-WINDOW spend-cap counters to store microCredits. This PR applies the same unit switch to the **fair-use / free-usage rolling-window AWU counters**.

## The problem

The rolling-window limiter (`addRateLimiterCount`) stored a value as a **count of rows** — its Lua did one `ZADD` per consumed unit, and the reader returned `zCount` (number of rows). So "5 credits" = 5 rows, and credits had to be whole integers. Naively switching the unit to microCredits would write ~1,000,000 rows per credit.

## Approach: one row per call + sum-on-read

The fair-use rolling window now stores the **amount in a single row** and **sums on read**:

- `addRateLimiterCount` treats `incrementBy` as a (possibly fractional) credit amount, converts it to integer microCredits via `roundCreditsToMicroCredits`, and writes **one** sorted-set entry `<microCredits>:<uuid>` (amount prefix + uuid for uniqueness). The window+grace `PEXPIRE` behavior is unchanged.
  - Removes the `throw` on non-integer `incrementBy` (fractional credits are now valid), replaced with a fail-open `if (!Number.isFinite(incrementBy) || incrementBy <= 0) return;` early-return. This also fixes a latent fragility where a throw here could fail the Temporal finalize activity.
- New `getWeightedRateLimiterCount({ key, timeframeSeconds })` reads via `zRangeByScore`, splits each member on the first `:`, `parseInt`s the prefix, and sums (skipping malformed members), returning the total **microCredits**.

`rateLimiter`, `getRateLimiterCount`, and every message-count / mention / premium-model key are **completely untouched** — they legitimately count rows.

## Key bump

The two AWU rolling keys are suffixed `:v2_microcredits` so the new `amount:uuid` rows can never be summed with pre-existing plain-uuid rows:
- `makeFairUseAwuCreditsRateLimitKeyForUser`
- `makeFreeUsageCostRateLimitKeyForUser`

These are short rolling windows, so the old keys expire quickly. **A user's current fair-use/free window resets once at cutover** (acceptable — there is no ES seed for rolling windows, unlike the fixed-window spend caps).

## Read sites (compare in like-units)

- `conversation.ts` fair-use gate: `getWeightedRateLimiterCount`, compares `>= roundCreditsToMicroCredits(maxAwuCredits)`.
- `free_usage.ts` free-usage gate: `getWeightedRateLimiterCount`, compares `>= roundCreditsToMicroCredits(freeUsageAwuCreditsLimitForAuth(auth))`.
- `user_block.ts` `getFairUseAwuCreditsStatus`: reads microCredits but **keeps `FairUseAwuCreditsStatus` credit-denominated (with decimals)** — converts the summed microCredits back to credits via `microCreditsToCredits` for the `count` field; `limit` stays in credits. The status semantics are identical, only the stored unit changed.

Writers (`credit_cost.ts`, `free_usage.ts`) still pass **credits** — the microCredit conversion now happens inside `addRateLimiterCount`.

## Poke

`FairUseAwuCreditsStatus` stays in credits, so no poke unit conversion is needed. The fair-use status is not surfaced in poke at all (only in the companion `FairUseCreditsUsage` widget, which already renders credits); nothing to change there.

## Tests

`rate_limiter.test.ts` (real Redis on :6479): fractional `2.5` credits → `2_500_000` microCredits; multiple adds sum; entries outside the window are excluded from the sum. The pre-existing `addRateLimiterCount` tests were migrated from `getRateLimiterCount` (row-count) to `getWeightedRateLimiterCount` (microCredit sum), since the function's storage changed. All 13 tests pass.

### DIFF
--- a/front/lib/api/assistant/conversation.ts
+++ b/front/lib/api/assistant/conversation.ts
@@ -83,6 +83,7 @@ import {
 import { countActiveSeatsForWorkspace } from "@app/lib/api/workspace_seats";
 import { isModelAvailable } from "@app/lib/assistant";
 import { Authenticator, getFeatureFlags } from "@app/lib/auth";
+import { roundCreditsToMicroCredits } from "@app/lib/credits/units";
 import { getSupportedModelConfig } from "@app/lib/llms/model_configurations";
 import { extractFromString, serializeMention } from "@app/lib/mentions/format";
 import { isApiKeyCapped } from "@app/lib/metronome/api_key_block";
@@ -124,8 +125,8 @@ import { WakeUpResource } from "@app/lib/resources/wakeup_resource";
 import { ServerSideTracking } from "@app/lib/tracking/server";
 import { concurrentExecutor } from "@app/lib/utils/async_utils";
 import {
-  getRateLimiterCount,
   getTimeframeSecondsFromLiteral,
+  getWeightedRateLimiterCount,
   rateLimiter,
 } from "@app/lib/utils/rate_limiter";
 import { withTransaction } from "@app/lib/utils/sql_utils";
@@ -3099,7 +3100,7 @@ async function isMessagesLimitReached(
 
   const user = auth.user();
   if (user && maxAwuCredits !== -1) {
-    const result = await getRateLimiterCount({
+    const result = await getWeightedRateLimiterCount({
       key: makeFairUseAwuCreditsRateLimitKeyForUser(
         owner,
         user.toJSON(),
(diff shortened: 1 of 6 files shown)


tdraier:

- #31015
    - #31021 ◀
        - #31027
        - #31024

fabiencelier (approved):
LGTM

PR #31023, opened 2026-08-24
[front] perf: Fetch render actions without the tool-execution join

id13:
## Description

https://app.pganalyze.com/databases/-642267629/queries/13265983126

Mcp action resource has always been a hot point in our data-model, especially the fetchByStepContents used in batch render message.

Currently, in this fetchByStepContents, we join mcp action with step tool execution resource to fetch all mcp action given a set of steps.

1. We do not filter steps passed down, and only function tool call steps can yield an mcp action row.
2. This join can be split into 2 queries provided the set of actions to fetch in two queries is much smaller than the set scanned during the join

### DIFF
--- a/front/lib/resources/agent_mcp_action_resource.ts
+++ b/front/lib/resources/agent_mcp_action_resource.ts
@@ -662,58 +662,88 @@ export class AgentMCPActionResource extends BaseResource<AgentMCPActionModel> {
       stepContents: AgentStepContentResource[];
     }
   ): Promise<AgentMCPActionResource[]> {
-    if (stepContents.length === 0) {
+    const functionCallStepContents = stepContents.filter(
+      (
+        content
+      ): content is AgentStepContentResource & {
+        value: AgentFunctionCallContentType;
+      } => content.isFunctionCallContent()
+    );
+    if (functionCallStepContents.length === 0) {
       return [];
     }
 
     const workspaceId = auth.getNonNullableWorkspace().id;
+    const agentMessageIds = [
+      ...new Set(
+        functionCallStepContents.map((content) => content.agentMessageId)
+      ),
+    ];
 
-    const agentStepContentToolExecutions =
-      await AgentStepContentToolExecutionModel.findAll({
-        where: {
-          workspaceId,
-          stepContentId: { [Op.in]: stepContents.map((content) => content.id) },
-        },
-        include: [
-          {
-            model: AgentMCPActionModel,
-            as: "agentMCPAction",
-            required: true,
-          },
-        ],
-      });
-
-    const stepContentsMap = new Map(stepContents.map((s) => [s.id, s]));
-
-    // Sandbox-child actions share their parent's stepContent and must not
-    // surface as separate executions in the conversation timeline.
-    const visibleExecutions = agentStepContentToolExecutions.filter(
-      (row) =>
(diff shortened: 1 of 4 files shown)


front/lib/resources/agent_mcp_action_resource.ts
```diff
@@ -662,58 +662,88 @@ export class AgentMCPActionResource extends BaseResource<AgentMCPActionModel> {
...
+    const functionCallStepContents = stepContents.filter(
+      (
+        content
+      ): content is AgentStepContentResource & {
+        value: AgentFunctionCallContentType;
+      } => content.isFunctionCallContent()
+    );
+    if (functionCallStepContents.length === 0) {
       return [];
     }
```
id13, line 674: quickwin: the only step that will join will be function call ones, so we drop early & reduce the number of ids passed down to sql queries

flvndvd: Seems legit, name is slightly misleading, but yes only actual tool calls are expected to have an associate row in `AgentStepContentToolExecutionModel`.

front/lib/resources/agent_mcp_action_resource.ts
```diff
@@ -662,58 +662,88 @@ export class AgentMCPActionResource extends BaseResource<AgentMCPActionModel> {
...
-          },
-        ],
-      });
-
-    const stepContentsMap = new Map(stepContents.map((s) => [s.id, s]));
+    const toolExecutions = await AgentStepContentToolExecutionModel.findAll({
+      attributes: ["agentMCPActionId", "stepContentId"],
+      where: {
+        workspaceId,
+        agentMessageId: { [Op.in]: agentMessageIds },
```
flvndvd: Why using `agentMessage` over `stepContentId`? FWIW `stepContentId` also has a direct index matching

front/lib/resources/agent_mcp_action_resource.ts
```diff
@@ -662,58 +662,88 @@ export class AgentMCPActionResource extends BaseResource<AgentMCPActionModel> {
...
+      actionIdsByStepContentId.set(toolExecution.stepContentId, actionIds);
+    }
+    if (actionIdsByStepContentId.size === 0) {
+      return [];
+    }
 
-    return visibleExecutions.map((row) => {
-      const a = row.agentMCPAction;
-      const stepContent = stepContentsMap.get(row.stepContentId);
+    const actions = await AgentMCPActionModel.findAll({
```
flvndvd, line 691: Same here, why not fetching directly by the id retrieved from `AgentStepContentToolExecutionModel.findAll()`?

front/lib/resources/agent_mcp_action_resource.ts
id13: Agent said for IO but I biased it by showing some terrible plans used by PG (with an bitmap AND)
Not convinced either, going back to stepContendId

front/lib/resources/agent_mcp_action_resource.ts
id13: see answer above

id13:
PMRR

front/lib/resources/agent_mcp_action_resource.ts
flvndvd: Yeah would not buy it either!

PR #31025, opened 2026-08-24
[front] fix: restore route tags in SQL comments

flvndvd:
## Description

This PR restores endpoint attribution in SQL comments used by Cloud SQL Query Insights.

After the API migration to Hono, the SQL commenter still relied on framework-specific telemetry attributes that were no longer populated. It now reads the existing request context instead.

Matched route patterns are captured before handler execution, ensuring comments contain normalized routes without concrete resource identifiers.

### DIFF
--- a/front-api/middlewares/request_instrumentation.test.ts
+++ b/front-api/middlewares/request_instrumentation.test.ts
@@ -1,3 +1,4 @@
+import { frontSequelize } from "@app/lib/resources/storage";
 import {
   getRequestContext,
   RequestCachedQuery,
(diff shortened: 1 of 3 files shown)


PopDaph (approved):

ykmsd (approved):
simpler than I thought 😄 Can we bring back peakConcurrentQueries too?

flvndvd:
> simpler than I thought 😄 Can we bring back peakConcurrentQueries too?

coming in another PR with temporal activity monitoring as well.

PR #31027, opened 2026-08-24
[front] Add disable_fair_use_awu_limit workspace feature flag

tdraier:
## Descriptions

Escape hatch to exempt a workspace from the per-user fair-use AWU credit limit,
covering both sides:
- Read (enforcement): the pre-message fair-use gate in isMessagesLimitReached
  skips the counter check when the flag is set.
- Write (recording): finalizeAgentMessageCost skips recording usage into the
  fair-use rolling counter when the flag is set.

The recording path now fetches feature flags once and reuses them for both the
fair-use gate and the existing spend-cap backup gate. On-demand stage, toggled
per workspace from poke like skip_free_usage_rate_limit.

## Risks

low

## Deploy Plan

deploy front

### DIFF
--- a/front/lib/api/assistant/conversation.ts
+++ b/front/lib/api/assistant/conversation.ts
@@ -3099,7 +3099,13 @@ async function isMessagesLimitReached(
   } = plan.limits.assistant;
 
   const user = auth.user();
-  if (user && maxAwuCredits !== -1) {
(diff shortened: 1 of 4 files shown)


tdraier:

- #31015
    - #31021
        - #31027 ◀

id13 (approved):
Looks legit

PR #31029, opened 2026-08-24
[admin governance] migrate builder role to user

ykmsd:
## Description
This PR is to migrate memberships role `builder` to `user`. We already communicated with users about deprecating builder role (via email) and already show role "User" in UI. We have 12,880 builder users in US and 9,792 in EU. We need to sync with WorkOS after (they also save roles in their db, and we need to trigger a sync or something iiuc). I will clean up the code to remove builder check once both are done. 

Builders will lose "write" access in global/conversations and open space:
https://github.com/dust-tt/dust/blob/151f01fba5bc9c76e0346c4a4a43f8f65887afe7/front/lib/resources/space_resource.ts#L1905-L1913

And they need to be manager if they want to have "write" access. 

There is another one in group space member resource but I heard this will be removed and safe to ignore:

https://github.com/dust-tt/dust/blob/22f9dddad92e57de18a9e9b04ad84c863c013012/front/lib/resources/group_space_member_resource.ts#L177-L193

### DIFF
--- a/front/migrations/20260824_migrate_builder_role_to_user.ts
+++ b/front/migrations/20260824_migrate_builder_role_to_user.ts
@@ -1,0 +1,104 @@
+import { QueryTypes } from "sequelize";
+
+import { frontSequelize } from "@app/lib/resources/storage";
+import { makeScript } from "@app/scripts/helpers";
+
+const DEFAULT_BATCH_SIZE = 1_000;
+const DEFAULT_BATCH_DELAY_MS = 100;
+
+const SELECT_SQL = `
+  SELECT id
+  FROM memberships
+  WHERE id > :lastId
+    AND role = 'builder'
+  ORDER BY id ASC
+  LIMIT :batchSize
+`;
+
+const UPDATE_SQL = `
+  UPDATE memberships
+  SET role = 'user'
+  WHERE id IN (:ids)
+    AND role = 'builder'
+`;
+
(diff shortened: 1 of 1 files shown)


front/migrations/20260824_migrate_builder_role_to_user.ts
```diff
@@ -0,0 +1,104 @@
+import { QueryTypes } from "sequelize";
+
+import { frontSequelize } from "@app/lib/resources/storage";
+import { makeScript } from "@app/scripts/helpers";
+
+const DEFAULT_BATCH_SIZE = 1_000;
+const DEFAULT_BATCH_DELAY_MS = 100;
```
ykmsd, line 7: zero idea if it's reasonable

front/migrations/20260824_migrate_builder_role_to_user.ts
```diff
@@ -0,0 +1,104 @@
...
+    AND role = 'builder'
+`;
+
+const COUNT_SQL = `SELECT COUNT(*) AS count FROM memberships WHERE role = 'builder'`;
+
+function wait(durationMs: number): Promise<void> {
+  return new Promise((resolve) => setTimeout(resolve, durationMs));
+}
+
+makeScript(
```
ykmsd, line 31: purely AI generated and I tried to verify but don't trust me 🥺

tdraier (approved):
LGTM

ykmsd:
I will merge but run it when there is no high latency in db 🥶

PR #31039, opened 2026-08-24
[front] Preserve retired models for tool footprint tokenization

flvndvd:
## Context

As we've backfilled conversations from earlier this year, we've noticed that many tools attribution footprint were failing because of:
```
Error: [ConsumptionAttribution] Failed to tokenize tool footprints: Cannot tokenize tool footprints: unknown model gpt-4o-2024-08-06.
``` 

Indeed some models were recently removed from the supported arrays of models. This PR adds those models to the historical piece so we can still safely tokenize those tools. Added one test to ensure it does not happen anymore.


aubin-tchoi (approved):
LGTM

PopDaph (approved):

PopDaph:
Thanks!

PR #31040, opened 2026-08-24
[connectors] slack: throttle conversations.list and bound GC retries

PopDaph:
The slack garbage collector workflow can get stuck retrying forever on a Slack rate limit (seen on `slack-GarbageCollector-10850`: 440+ attempts, one every 30s for hours). `getChannelsToGarbageCollect` paginates `conversations.list` with no throttling, hits the Tier 2 limit (~20/min) before finishing, and each retry restarts pagination from page 1, so it never completes.

- Throttle `conversations.list` through `throttleWithRedis` (15/min), same pattern as the other Slack methods.
- Bound `getChannelsToGarbageCollect` to 25 attempts so the workflow fails instead of spinning forever (it is relaunched after each sync), and raise its start-to-close timeout to 30 minutes since pagination is now throttled.

The stuck production workflow needs to be terminated manually after deploy.

### DIFF
--- a/connectors/src/connectors/slack/lib/channels.ts
+++ b/connectors/src/connectors/slack/lib/channels.ts
@@ -460,6 +460,9 @@ async function _getTypedChannelsUncached(
   let nextCursor: string | undefined = undefined;
   let nbCalls = 0;
   do {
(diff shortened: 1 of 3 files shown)


aubin-tchoi (approved):
LGTM

PR #31041, opened 2026-08-24
[front] perf: migrate v1 analytics export tables to consumption ES index

avervaet:
## Summary
- Migrate 7 tables of `GET /api/v1/w/{wId}/analytics/export` (`usage_metrics`, `active_users`, `source`, `agents`, `users`, `skills`, `tool_usage`) from the legacy message-level ES index (`front.agent_message_analytics`) to the newer per-consumption-unit index (`front.agent_message_consumption_analytics`), keeping the external response shape (columns/CSV/JSON structure) unchanged.
- `tool_usage` also drops its per-tool-server N+1 query fan-out (`concurrentExecutor` loop) in favor of a single aggregation, since the consumption index is already one flat document per tool invocation.
- Left `GET /api/v1/w/{wId}/usage` (deprecated), and the `messages`, `feedback`, and `skill_usage` export tables untouched — their required fields (message ancestry, feedback data, message-level skill mentions) don't exist in the consumption index without a real semantic change.
- Existing dashboard-facing fetchers used by private `front-api/routes/w/[wId]/analytics/*` routes were left untouched; new sibling `fetchConsumption*` functions were added instead, since those fetchers take a generic ES query and swapping their index in place would have silently broken the dashboards.

## Notes on value drift (shape-identical, but underlying numbers can shift slightly)
- Message/user/conversation counts move from exact `doc_count` to `cardinality(...)` aggregations (approximate above ~40k distinct values).
- Day bucketing shifts from message-creation time to message-completion time (`completed_at`).
- Free/non-billed usage (e.g. `agent_sidekick` origin) is not indexed in the consumption index, so it disappears from these exports.
- Credits move from AWU-rounded integers to `credit_micro`-derived values (minor rounding differences).

## Test plan
- [ ] `npx tsgo --noEmit` clean on touched files (no new errors vs. baseline)
- [ ] Biome clean on touched files
- [ ] `context_origin.test.ts` passes (6/6) against the refactored path
- [ ] Full `vitest` run blocked repo-wide by an unrelated pre-existing missing `@novu/framework` dependency — verify in CI
- [ ] Manually verify `GET /api/v1/w/{wId}/analytics/export?table=tool_usage` (and the other 6 tables) against a workspace with recent consumption data, comparing against pre-change output for a known date range

### DIFF
--- a/front-api/routes/v1/w/[wId]/analytics/export.test.ts
+++ b/front-api/routes/v1/w/[wId]/analytics/export.test.ts
@@ -5,7 +5,7 @@ import { ENSURE_IS_ADMIN_ERROR_MESSAGE } from "@front-api/middlewares/ensure_rol
 import { describe, expect, it, vi } from "vitest";
 
 vi.mock("@app/lib/api/assistant/observability/messages_metrics", async () => ({
-  fetchMessageMetrics: vi.fn(
+  fetchConsumptionUsageMetrics: vi.fn(
     async () =>
       new Ok([
         {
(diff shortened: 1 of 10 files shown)


avervaet:
Closing, superseded/no longer needed.

PR #31042, opened 2026-08-24
Stagger the home greeting in and center the input bar

pinotalexandre:
The greeting is picked client-side, so it lands one render after mount and used to pop in. Each word now rises 10px and fades over 420ms, offset by 70ms, keyed on the greeting so the entrance replays once the name resolves. Reduced motion skips the initial state entirely.

The hero (greeting + input bar) now fills exactly one panel height with equal flex spacers above and below the input bar, which puts the input bar on the vertical center of the page; the agent browser flows underneath. Spacing between the greeting and the bar moved from the header's padding to a gap on the hero: padding sits outside a `flex-basis: 0%` item and would push the bar 24px below center, while a gap is taken off the free space both spacers share.

### DIFF
--- a/front/components/assistant/conversation/ConversationContainer.tsx
+++ b/front/components/assistant/conversation/ConversationContainer.tsx
@@ -295,85 +295,82 @@ export function ConversationContainerVirtuoso({
         />
       ) : (
         <>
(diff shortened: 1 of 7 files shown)


ClementAupiais (approved):
LGTM

PR #31043, opened 2026-08-24
[Activation] Tighten orchestrator nudge cadence and per-run selection

frankaloia:
## Summary
- Stop scheduled nudges after 2 consecutive unanswered ones (down from 3); a user reply still resets the count.
- Skip weekend scheduled runs in the regional timezone, and create new workspace schedules as Mon–Fri only. Existing daily schedules are unchanged; the enumerator returns no one on Sat/Sun.
- When a workspace exceeds the 100-person per-run cap, keep the people who have gone the longest without a nudge (never-nudged first), then send in workday slot order.

## Test plan
New unit tests

## Risk
Low

## Deploy Plan
1. Deploy Front

### DIFF
--- a/front/temporal/activation_scheduler/config.ts
+++ b/front/temporal/activation_scheduler/config.ts
@@ -14,18 +14,41 @@ export const DEFAULT_ACTIVATION_NUDGE_FREQUENCY_CAP_DAYS = 2;
 // Default number of consecutive unanswered nudges (no user message since the
 // nudge fired) after which the scheduler stops nudging a pod. Workspaces can
 // override this via the `activationNudgeMaxUnansweredCount` metadata key.
-export const DEFAULT_ACTIVATION_NUDGE_MAX_UNANSWERED_COUNT = 3;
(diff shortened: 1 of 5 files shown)


id13 (approved):

front/temporal/activation_scheduler/config.ts
```diff
@@ -14,18 +14,41 @@ export const DEFAULT_ACTIVATION_NUDGE_FREQUENCY_CAP_DAYS = 2;
...
+  }
+  if (a.lastNudgedAtMs === null) {
+    return -1;
+  }
+  if (b.lastNudgedAtMs === null) {
+    return 1;
+  }
+  return a.lastNudgedAtMs - b.lastNudgedAtMs;
+}
+
```
id13: lodash doesn't have it ?

frankaloia: Good call — switched to `sortBy` with `lastNudgedAtMs ?? -Infinity` so never-nudged still sorts first (`sortBy` would otherwise put `null` last).

PR #31045, opened 2026-08-24
[front] usage tracking: make trackProgrammaticUsageActivity retryable

PopDaph:
I landed here because I was trying to reduce Failed workflow. 
[See in temporal for EU](https://cloud.temporal.io/namespaces/eu-dust-front-prod.gmnlm/workflows?query=%60ExecutionStatus%60%3D%22Failed%22+AND+%60WorkflowType%60%3D%22trackProgrammaticUsageWorkflow%22)
[See in temporal for US](https://cloud.temporal.io/namespaces/dust-front-prod.gmnlm/workflows?query=%60ExecutionStatus%60%3D%22Failed%22+AND+%60WorkflowType%60%3D%22trackProgrammaticUsageWorkflow%22)

`trackProgrammaticUsageWorkflow` fails permanently on any transient error (`maximumAttempts: 1`): DB pool acquire timeouts and 5-minute StartToClose timeouts in prod, and the programmatic cost of those executions is never tracked. The single attempt existed because the activity mutates the credit ledger, so a blind retry could double-consume. 

- Guard the mutation phase with a redis `SET NX` marker keyed on workspaceId + the full sha256 fingerprint of the sorted dustRunIds (the same identity Metronome dedups retries with, untruncated for collision resistance). A retry or a timed-out zombie attempt that finds the marker skips consumption. Fails closed on redis errors: without the marker we cannot guarantee at-most-once consumption, so the attempt fails and Temporal retries.
- Active credits are prefetched before the marker is set, so the pool-timeout reads this PR recovers from stay retryable; only the millisecond ledger writes sit inside the marked window.
- Allow 5 attempts with backoff on the activity, bounded by a 1-hour `scheduleToCloseTimeout` so no retry can outlive the marker's 24h TTL.

Known limit: the marker is only as durable as redis. Double-counting one execution requires a post-consumption crash AND redis losing that fresh key within the 1-hour retry window; the blast radius is one message's cost on the internal ledger. Closing it would need a ledger-transactional marker table, deliberately not worth it. On the other side, a crash inside the ledger-write window drops that execution's tracking, same as today's failure mode but with a far smaller window. Safe to rollback; stale guard keys expire after 24h.

### DIFF
--- a/front/lib/api/programmatic_usage/tracking.test.ts
+++ b/front/lib/api/programmatic_usage/tracking.test.ts
@@ -13,6 +13,7 @@ import { CreditResource } from "@app/lib/resources/credit_resource";
 import logger from "@app/logger/logger";
 import { GroupFactory } from "@app/tests/utils/GroupFactory";
 import { MembershipFactory } from "@app/tests/utils/MembershipFactory";
+import { RunFactory } from "@app/tests/utils/RunFactory";
 import { UserFactory } from "@app/tests/utils/UserFactory";
 import { WorkspaceFactory } from "@app/tests/utils/WorkspaceFactory";
 import type { WorkspaceType } from "@app/types/user";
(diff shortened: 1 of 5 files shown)


tdraier (approved):
LGTM

flvndvd:
Apologies for miss-clicking and merging it 🙈.

flvndvd (commented):
LGTM, EU has some others errors as well.

PopDaph:
Indeed, sorry it makes it misleading -> this is to fix the one that time out!

PR #31046, opened 2026-08-24
refactor: nest automation breakdown under the trigger

adrsimon:
## Description

Chore, no behaviour change. Two things were off about the automation analytics routes.

The breakdown endpoints took their `triggerId` in the POST body while sitting right next to a `triggers` route that lists those same triggers, so the id moves into the path. And the personal variants lived in a `me/automations` tree parallel to `analytics/automations`, when they are the same analytics scoped to the caller, so they now mirror it under `me/analytics`.

- `analytics/automations/trigger-breakdown` -> `analytics/automations/triggers/:tId/breakdown`
- `me/automations/trigger-breakdown` -> `me/analytics/automations/triggers/:tId/breakdown`
- `me/automations/triggers` -> `me/analytics/automations/triggers`

`triggers.ts` becomes `triggers/index.ts` with a `[tId]` subtree, same shape as `w/[wId]/triggers/[tId]`. The breakdown request body is now just the consumption period, and the hook builds one URL instead of branching on scope.

## Tests

Moved the existing route tests alongside their new paths and updated them to pass the id in the URL. Dropped the analytics "400 when triggerId is missing" case, there is no such request to make once the id is a path segment. 23 tests pass in the front-api automations tree, plus the 5 in `AutomationsTriggerBreakdown.test.tsx`.

## Risk

Low. All of these are private (`@ignoreswagger`) and called only from `useAutomationsTriggerBreakdown` and `useUserAutomationsTriggers`, which move with them. Nothing public changes.

## Deploy Plan

Deploy front.


front/hooks/useAutomationsTriggerBreakdown.ts
```diff
@@ -21,12 +21,9 @@ export function useAutomationsTriggerBreakdown({
   scope: AutomationsScope;
   disabled?: boolean;
 }) {
-  const url =
-    scope === "user"
-      ? `/api/w/${workspaceId}/me/automations/trigger-breakdown`
-      : `/api/w/${workspaceId}/analytics/automations/trigger-breakdown`;
+  const scopePath = scope === "user" ? "me" : "analytics";
+  const url = `/api/w/${workspaceId}/${scopePath}/automations/triggers/${triggerId}/breakdown`;
```
aubin-tchoi: shouldn't the one for users be api/w/:wId/me/analytics/automations/triggers/:tId/breakdown in this case? (same but with /me in addition)

aubin-tchoi (approved):
LGTM

PR #31050, opened 2026-08-24
fix(archived_agents): let admins find archived agents on spaces they can't read

achilleburah:
## Description

The "Show hidden agents" admin bypass for the manage view skips space permission filtering entirely, so an admin can see any active agent regardless of space membership. The archived view had no equivalent: once an agent the admin could only see through that bypass gets archived, it drops out of "All" (no longer active) and out of "Archived" too (`filterAgentsByRequestedSpaces` still runs), becoming invisible in the UI even to admins.

Extends the same skip to the `archived` view for admins, matching its documented contract ("all agents that are archived", admin/superuser only, no partial view). Also hides the "Show hidden agents" toggle outside the All tab, since it never affected any other tab's fetch and Archived has no togglable restricted state to begin with.

Found while testing the archive-inactive-agents feature: an agent living in a restricted space, once auto-archived, became unreachable from any admin UI.

## Tests

- New regression test: an admin can find an archived agent built on a space they are not a member of (`front/lib/api/assistant/configuration/views.test.ts`) — fails without the fix (verified by reverting `views.ts` locally), passes with it.
- New test: a plain member still cannot see it (space filtering still applies to non-admins).
- `npx tsgo --noEmit` clean, `biome check` clean, `views.test.ts` 9/9 passing.

## Risk

Low. The bypass is additive (`|| (agentsGetView === "archived" && auth.isAdmin())`) and gated on `auth.isAdmin()`, so non-admin behavior is unchanged. Safe to roll back by reverting.

## Deploy Plan

Standard deploy, no migration or feature flag involved.

### DIFF
--- a/front/components/pages/builder/agents/ManageAgentsPage.tsx
+++ b/front/components/pages/builder/agents/ManageAgentsPage.tsx
@@ -387,7 +387,7 @@ export function ManageAgentsPage() {
                     counterValue={`${agentsByTab[tab.id].length}`}
                   />
                 ))}
-                {canShowHiddenAgents && (
+                {canShowHiddenAgents && activeTab === "all_custom" && (
                   <span className="ml-auto flex gap-1 self-center text-sm text-muted-foreground">
                     <label className="flex cursor-pointer flex-row items-center gap-2 whitespace-nowrap">
                       <Checkbox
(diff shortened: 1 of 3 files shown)


fabiencelier (approved):
LGTM ✅ 
Thanks for fixing

PR #31037, opened 2026-08-24
feat(automations): workspace pool card, no name sorting

adrsimon:
## Description

Adds a third card to the automations overview with the number of triggers running on the workspace pool, out of the workspace total, plus a hint for the ones left on member pools. Reading that number previously meant paging through the table.

Also drops sorting on the Name column, and centers the Type header over its icon (the automations table's local header renderer ignored `headerAlign: "center"`).

Capture d’écran 2026-08-24 à 14 29 21

## Tests

Manual: overview cards on the automations analytics page, with triggers moved between pools; checked the Name header is no longer clickable and the Type header lines up with the icons.

## Risk

Low, one extra `count` query on the triggers table for the overview endpoint and cosmetic table changes.

## Deploy Plan

Deploy front.

PR #31052, opened 2026-08-24
fix(triggers): assert workspacePool in countForWorkspace tests

adrsimon:
## Description

`countForWorkspace` started returning `workspacePool` in #31037 but the tests still asserted only `enabled` and `total`, so both cases failed on the extra key.

## Tests

`npm test -- lib/resources/trigger_resource.test.ts -t countForWorkspace` passes.

## Risk

None, test-only.

## Deploy Plan

Deploy front.


id13:
LGTM

PR #31053, opened 2026-08-24
[front] Batch space membership from group_permissions, drop eager-groups reliance

tdraier:
## What

Replaces `SpaceResource.isMemberByGroupModelIds` / `isMemberByGroupPredicate` with a batched static `SpaceResource.listMemberSpaceModelIdsByUser(auth, { spaces, userModelIds })`. It:

- reads every space's grants from `group_permissions` in one query (the source of truth for space access, instead of the per-instance eagerly-loaded `this.groups`),
- resolves each user's group memberships in one more query (existing `GroupResource.listGroupModelIdsByUserModelIdInWorkspace`),
- applies the same per-kind membership rule as `isMember` via a small private `isMemberFromGrants` helper.

`lib/api/spaces/access.ts` (the only two callers — both async: `listUsersWithoutAccessToSpaceResources` and `listUsersWithoutAccessToSpaces`) is rewritten on top of it and no longer reads `space.groups`.

## Why

Part of removing consumers of the eagerly-loaded `SpaceResource.groups`. This removes the **async** membership consumer. Behavior is unchanged — the batched query returns the same grant set the eager include carries.

Note: this does **not** yet let us drop the eager `groups` load. Two consumers still pin it: `isOpen()` (sync permission path) and `toJSON().groupIds` (a **public** v1 `Space` API field, append-only per BACK12). Those are tracked separately.

## Test plan

- `front-api routes/w/[wId]/spaces/access-check.test.ts` — 8 passed
- `front-api routes/w/[wId]/skills/[sId]/editors.test.ts` — 10 passed
- `front-api routes/w/[wId]/skills/index.test.ts` — 46 passed
- `front lib/resources/space_resource.test.ts` — 72 passed
- `tsgo --noEmit` clean

### DIFF
--- a/front/lib/api/spaces/access.ts
+++ b/front/lib/api/spaces/access.ts
@@ -1,5 +1,4 @@
 import type { Authenticator } from "@app/lib/auth";
-import { GroupResource } from "@app/lib/resources/group_resource";
 import { MembershipResource } from "@app/lib/resources/membership_resource";
 import { SpaceResource } from "@app/lib/resources/space_resource";
 import { UserResource } from "@app/lib/resources/user_resource";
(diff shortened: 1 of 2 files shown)


tdraier:

- #31053 ◀

fabiencelier (approved):

PR #31054, opened 2026-08-24
[front] Agent loop: stop failing workflows on tool activity heartbeat timeouts

PopDaph:
## Description

87% of agentLoopWorkflow FAILED runs over the last 7 days (823/944) are `runToolActivity` heartbeat timeouts ([Datadog](https://app.datadoghq.eu/logs?query=service%3Afront-agent-loop-worker%20%22Workflow%20failed%22%20%40workflowType%3AagentLoopWorkflow)). Tools with the default `no_retry` policy run with `maximumAttempts: 1`, so a single 60s heartbeat gap kills the whole run. Two causes observed in production:

- Un-heartbeated stretch: action state transitions, MCP server connection and event publication sit outside the MCP call's heartbeat loop. Traced a tool stuck ~2min pre-connect, timed out by the server, that then completed successfully into the void.
- Hard pod kills (no drain): pods dying mid-work at load peaks; every in-flight tool activity heartbeat-times-out 60s later.

Two changes:

- `runToolActivity` now heartbeats for its whole lifetime (same wrapper pattern as `runModelAndCreateActionsActivity`), so a live-but-stalled pod is never timed out.
- The workflow swallows terminal tool activity timeouts the way it already swallows LLM-unresponsive model timeouts: the message is finalized as errored for the user either way, and the FAILED status was only Temporal noise. Tool application errors still fail the workflow.

Observability: swallowed timeouts no longer emit a "Workflow failed" log, so `finalizeErroredAgentLoopActivity` now logs the error cause ("Agent loop finalized as errored") to keep dead-pod heartbeat timeouts queryable and alertable in Datadog.

Follow-ups (separate PRs): retry-policy classification for read-only remote MCP tools (needs plumbing in `remote_servers.ts`), hard-kill investigation (OOM/evictions) on agent-loop pods.

## Tests

Unit tests cover the production swallow decision (`isSwallowableWorkflowFailure`, patched and unpatched), `isTerminalRunToolTimeout`, `getWorkflowFailureDetails`, and a mixed-failure regression on `waitForAllPromises` (a sibling application failure wins over a swallowable timeout). `workflows.test.ts` and `finalize.test.ts` green. `tsgo --noEmit` clean on touched files.

## Risk

Swallowing a tool timeout changes the workflow's terminal command (complete instead of fail), so it is gated behind `patched("swallow-terminal-tool-timeouts")`: replays of histories that predate the patch keep the legacy throw. Standard patch caveat applies to rollback: workflows that recorded the marker must complete before rolling back to a build without the patch. Parallel tools propagate a sibling's real failure in preference to a swallowable timeout, so application errors cannot be masked.

## Deploy Plan

- Deploy front.
- Watch `service:front-agent-loop-worker "Workflow failed" @workflowType:agentLoopWorkflow`: heartbeat-timeout failures should drop to ~zero.

### DIFF
--- a/front/temporal/agent_loop/activities/finalize.ts
+++ b/front/temporal/agent_loop/activities/finalize.ts
@@ -5,6 +5,7 @@ import {
 } from "@app/lib/api/assistant/email/email_reply";
 import type { AuthenticatorType } from "@app/lib/auth";
 import { Authenticator } from "@app/lib/auth";
+import logger from "@app/logger/logger";
 import {
   launchAgentMessageAnalytics,
   launchAgentMessageConsumptionAttribution,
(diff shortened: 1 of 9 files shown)


PopDaph:
Not putting into review because first we want to see if it's all tools or a given one, I've added logs here: https://github.com/dust-tt/dust/pull/31106

PR #31055, opened 2026-08-24
feat(automations): filter triggers by credit pool

adrsimon:
## Description

The automations admin table shows which credit pool each trigger charges, but there was no way to narrow it down to one. This adds a Pool filter category (Workspace / Member) next to Type, using the same option list the pool column dropdown already uses. Execution mode isn't indexed in the consumption documents, so like the kind filter it resolves to a set of trigger ids first and is applied as a terms filter on the ranking query. The /me automations table keeps its two categories.

## Tests

Extended the filter mapping unit tests (pool selection maps to `executionModes`, invalid ids are dropped, pool stays out of the scope filter). Checked the panel manually: filtering on Workspace or Member narrows the table, combining Pool with Type intersects, and the count badge and summary chips pick the category up.

Capture d’écran 2026-08-24 à 17 34 29
Capture d’écran 2026-08-24 à 17 34 19

## Risk

Low. The new filter field is optional everywhere; an unset pool filter goes down the exact same query path as before. `listByWorkspaceAndKinds` was generalized to also take execution modes and had a single caller.

## Deploy Plan

Deploy front.

### DIFF
--- a/front/components/workspace/analytics/automations/AutomationsFilterOptionIcon.tsx
+++ b/front/components/workspace/analytics/automations/AutomationsFilterOptionIcon.tsx
@@ -28,6 +28,7 @@ export function AutomationsFilterOptionIcon({
         />
       );
     case "type":
+    case "pool":
       return null;
     default:
       assertNeverAndIgnore(option.category);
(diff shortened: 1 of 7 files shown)


avervaet (approved):
LGTM

PR #31056, opened 2026-08-24
[front] fix: carry seat consumption when origin fully consumed

tdraier:
## Problem

A user who **fully spent their origin seat allowance** before upgrading between two recurring-credit seats (e.g. `pro` → `max`) ended up able to spend **both** allowances — e.g. 8000 (pro) + 40000 (max) = **48000 credits**.

The carry-over logic in `computeSeatCreditTransfers` (`front/lib/metronome/seats.ts`) bailed on a zero or negative origin balance:

```ts
if (remaining === undefined || remaining <= 0) {
  continue; // no transfer recorded
}
```

That guard conflated a **fully-consumed origin** (`remaining === 0`, the exact case where the consumed amount must be carried) with "nothing to do". So the transfer was skipped, the origin was never emptied, and the new `max` seat kept its fresh **40000** on top of the 8000 already spent.

This was confirmed on a real contract's worker logs: the seat-sync ran (`Current seat state fetched`, `Updating seat-based subscription assignments`) with **no** `No seat-type transfer candidates` line (so a candidate existed) and **no** `Emptying origin seat credit for transfer` / `carryConsumptionToNewSeatCredits done` lines — i.e. the transfer was computed to empty by the `remaining <= 0` guard.

## Fix

- `computeSeatCreditTransfers` now only skips when the balance reading is **missing** (`undefined`) — we can't derive `consumed` then. A zero or negative (overdrawn) balance produces a transfer with `consumed = allocation − remaining`.
- `emptyOriginSeatCreditsForTransfers` skips just the origin-emptying adjustment when there is no unused balance to reclaim (`remaining <= 0`), but still carries the consumed amount onto the new seat.
- Idempotency is unchanged in effect: it comes from the seat reassignment (once Metronome matches the DB, no transfer is detected) and the destination's absolute reconcile to `allocation − consumed` — not from the origin balance.

## Tests

- Corrected the misleading `remaining === 0` test (it encoded the bug) to assert full carry-over: `pro`(8000/8000) → `max` ends at remaining 32000.
- Added coverage for an overdrawn origin (negative remaining) and for the `undefined`-balance skip.
- All 66 tests in `seats.test.ts` pass; `tsgo --noEmit` clean.

### DIFF
--- a/front/lib/metronome/seats.test.ts
+++ b/front/lib/metronome/seats.test.ts
@@ -1012,13 +1012,62 @@ describe("computeSeatCreditTransfers", () => {
     ).toEqual([]);
   });
 
-  it("is idempotent: skips an already-emptied origin credit", () => {
-    // After a prior transfer the old credit is at 0 → nothing to carry over.
+  it("carries the full allocation when the origin was fully consumed (remaining 0)", () => {
+    // 8000/8000 spent on pro → the whole allocation must carry over so max ends
+    // at 8000/40000 (remaining 32000), NOT a fresh 40000 on top of the 8000
+    // already spent (which would let the user spend 48000). A zero balance is a
+    // fully-consumed origin, not a signal to skip.
+    const transfers = computeSeatCreditTransfers({
+      metronomeSeatByUser: new Map([["u1", "pro"]]),
+      desiredSeatByUser: new Map([["u1", "max"]]),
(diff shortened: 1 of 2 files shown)


tdraier:

- #31056 ◀

fabiencelier (commented):
LGTM ✅

fabiencelier (approved):

PR #31057, opened 2026-08-24
feat(agent_inactivity): add temporal daily workflow

achilleburah:
## Description

Adds a Temporal-based nightly workflow for automatically archiving inactive agents.

- Introduces a dedicated `agent_inactivity` Temporal worker, workflow, schedule, and operator CLI.
- Enumerates workspaces with the `archive_inactive_agents` feature enabled and a valid inactivity threshold, then sweeps each workspace with bounded concurrency.
- Reuses the existing archival eligibility logic for manual and scheduled runs, while using all-space internal admin authorization so agents tied to restricted spaces are included.
- Makes scheduled archival safe to retry with a fixed evaluation timestamp, activity heartbeats, bounded retries, overlap protection, and idempotent agent transitions.
- Adds the `workspace.inactive_agents_archived` audit event with threshold, archived-count, and skipped-count metadata.
- Registers and bundles the new Temporal worker, and moves the reinforcement schedule start to 03:00.

## Tests

Added and updated Vitest coverage for:

- Workspace feature-flag and threshold gating.
- Archiving agents in unrestricted and restricted spaces.
- Disabled configuration and invalid thresholds.
- Idempotent replay and workspace isolation.
- Temporal activity behavior with mocked schedule and wake-up clients.

## Risk

- Archiving changes an agent's status and disables its triggers and wake-ups, but the transition remains reversible through the existing restore path.
- Execution is gated by the feature flag and per-workspace threshold, and invalid thresholds fail without archiving.
- Each workspace is processed independently with bounded concurrency and retries, so one failing workspace does not stop the rest of the sweep.
- The scheduled path uses all-space authorization within the target workspace to avoid missing agents that reference restricted spaces.

## Deploy Plan

- Deploy the application and the new `agent_inactivity` Temporal worker bundle.
- The worker creates the nightly schedule if it does not already exist, using the current region timezone and skip-on-overlap behavior.
- Verify the worker starts successfully, the schedule exists, and the first run reports expected workspace and archival counts in logs.
- Enable `archive_inactive_agents` only for workspaces with a valid inactivity threshold.
- Use `temporal/agent_inactivity/admin/cli.sh run-now` or `run-workspace --workspaceId <sId>` for controlled operator verification when needed.

### DIFF
--- a/front/lib/api/assistant/inactivity/archive_inactive_agents.ts
+++ b/front/lib/api/assistant/inactivity/archive_inactive_agents.ts
@@ -6,15 +6,15 @@ import {
 } from "@app/lib/api/assistant/inactivity/fetch_inactive_agents";
 import type { AgentInactivityPolicyError } from "@app/lib/api/assistant/inactivity/policy";
 import { computeInactivityCutoffAt } from "@app/lib/api/assistant/inactivity/policy";
-import type { Authenticator } from "@app/lib/auth";
+import { Authenticator } from "@app/lib/auth";
+import { heartbeat } from "@app/lib/temporal";
 import logger from "@app/logger/logger";
 import type { Result } from "@app/types/shared/result";
 import { Err, Ok } from "@app/types/shared/result";
 
 /**
  * Archives a workspace's inactive agents. Both the manual endpoints and the nightly Temporal
- * activity enter here, so the rules cannot drift between them; the population still differs, since
- * every read is permission-filtered.
+ * activity enter here, on the same rules over the same population.
  */
 
 export interface InactiveAgentsArchivalInput {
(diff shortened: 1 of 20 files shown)


front/lib/api/assistant/inactivity/fetch_inactive_agents.ts
```diff
@@ -66,17 +62,17 @@ export function countSkipsByReason(
   return counts;
 }
 
-/** The facts the rules read about an agent, as they stand when the page is evaluated. */
-interface AgentArchivalFacts {
+/** The status and triggers the rules read about an agent, as they stand when it is evaluated. */
```
achilleburah: sry this is just renaming but i hated "facts" it meant nothing really.

achilleburah:
> workspace.inactive_agent_archival_updated is still at v1 (schema changed but version not incremented)


=> not relevant since the previous version is just implemented in the PR right below in the stack.

front/lib/api/assistant/inactivity/archive_inactive_agents.ts
```diff
@@ -55,16 +55,29 @@ export async function archiveInactiveWorkspaceAgents(
...
   const archivedAgentIds: string[] = [];
   const skipped = [...refused];
 
   for (const { agentId, lastMentionedAt } of eligible) {
+    // Temporal heartbeat to avoid activity timeout
+    await heartbeat();
+
     // Not a compare-and-set: an agent restored since the read is archived anyway. Reversible.
-    const archived = await archiveAgentConfiguration(auth, agentId);
+    const archived = await archiveAgentConfiguration(everySpaceAuth, agentId);
```
fabiencelier: This is not ideal as it will not mark the actual user as the one who is doing the action (in the log for instance)
So maybe we should just add a `dangerouslySkipPermissionFiltering` instead a propage it to the functions below ?
if that's not too complex I think it's better than creating the everySpaceAuth

fabiencelier (approved):
LGTM for the temporal part
One comment on the auth part

front/lib/api/assistant/inactivity/archive_inactive_agents.ts
achilleburah: yeah good point

PR #31060, opened 2026-08-24
[front] feat: notify the user when the workflow-alert threshold pauses the loop

avervaet:
## Description

The spend checkpoint pause now notifies the user the same way a blocked tool permission already does, instead of relying only on the generic conversation-unread notification.

- Reuses `notifyManualActionRequired`, the exact same "action required" notification blocked tool actions trigger, so a user gets consistent signal regardless of which of the two reasons paused the conversation.
- Only fires once per pause (guarded on `conversation.actionRequired`), the same way tool-permission notifications already avoid re-notifying for every subsequent blocked step.

Stacked on #31020, top of the spend-checkpoint stack — see #31017 for the stack overview.

## Tests

Added coverage for both the notify-once and don't-notify-again cases; existing suite still passing.

## Risk

Low: additive notification call using an existing, already-relied-on mechanism. No change to the pause/resume behavior itself.

## Deploy Plan

Deploy last, after the rest of the stack is live.


avervaet:

- #31017
    - #31018
        - #31019
            - #31020
                - #31060 ◀

aubin-tchoi (approved):

PR #31062, opened 2026-08-24
[Activation] Add Dust Pod Goal skill behind a feature flag

frankaloia:
## Summary
- Adds the **Dust Pod Goal** global skill behind a FF. This is essentially the same as Dust Learning skill but optimized for group onboarding.
- Breaking out some of the Dust Learning prompts to shared files
- Tbh really didn't scrutinize the Pod goal skill language too much. I imagine I will have a follow-up to tweak some of the wording after my testing

## Risk
Low (not customer facing)

## Testing
Tested locally. Will be used in a follow-up when a group onboarding pod is provisioned 

## Deploy Plan
1. Deploy Front


smb2268 (approved):
LGTM

PR #31065, opened 2026-08-24
[front] refactor: query agent triggers through the Poke trigger endpoint

aubin-tchoi:
## Description

Follow up on #31010 and #31030 to move the agent trigger table onto the same Poke trigger query used by the workspace table.

The endpoint now accepts an agent or workspace scope. Agent requests query only that agent’s triggers and return a Poke-safe row shape, while workspace pagination, filtering, consumption ranking, and current-page sorting remain unchanged. The existing GET endpoint remains available for compatibility.

## Tests

- Front and front-api typechecks.
- Scoped Biome checks.
- Swagger annotation lint.

## Risk

Low. Poke-only.

## Deploy Plan

Deploy front.


aubin-tchoi:
Superseded: the agent trigger view and legacy GET endpoint were removed from #31010, leaving the trigger table workspace-only.

PR #31071, opened 2026-08-25
[front] feat: add agent-scoped consumption API

aubin-tchoi:
## Description

Part of https://github.com/dust-tt/tasks/issues/10167.

The goal of this PR and the one that follows is to replace the Insights panel from the Agent info page with the analytics consumption view, scoped to the current agent.

This PR starts by adding the backend part.
Re-uses the same logic introduced for the personal view.

## Tests

- Added tests to check access rules notably (only editors).

## Risk

- Low.

## Deploy Plan

- Deploy front.


front-api/routes/w/[wId]/analytics/consumption/index.ts
```diff
@@ -45,17 +60,66 @@ export function createPersonalConsumptionRoutes() {
...
+        api_error: {
+          type: "agent_configuration_not_found",
+          message: "The agent you're trying to access was not found.",
+        },
+      });
+    }
+
+    if (
+      !agent.canEdit &&
+      !(await auth.hasWorkspacePermission("publish", "agent"))
```
Comment (avervaet) [inline front-api/routes/w/[wId]/analytics/consumption/index.ts]: For a visible agent, a non-editor with publish:agent can access that agent’s analytics.

Could we gate this with agent.canEdit || auth.isAdmin() instead?

avervaet (approved):

front-api/routes/w/[wId]/analytics/consumption/index.ts
Comment (aubin-tchoi) [reply front-api/routes/w/[wId]/analytics/consumption/index.ts]: was hesitating here, canEdit || isadmin wfm 👍🏻

PR #31073, opened 2026-08-25
[admin governance] use user role for internal workspace auth

ykmsd:
## Description
This PR is a part of removing builder role. We create authenticator internally and we've been using either "builder" role or "admin" role. From my conversation with claude code, the role builder doesn't give any meaningful difference at every call site and it's safe to change, but I will test it on front-edge a few cases just in case.

Here is what Claude Code said: 

The difference: the two functions are identical except role: "builder" → role: "user". Same groups, same permissions. The only capability builder had over user: write access to the global space, conversations space, and open regular spaces (via the role-path ACL). Nothing else.

Does it affect callers? No. None of the 16 call sites use that capability:
- None check isBuilder() / role === "builder".
- None write into a global/conversations/open space through the role grant. Their writes are either workspace-scoped rows (tags, webhook requests, analytics, backfills, seats) or run under a separate per-user auth. Their reads use dangerouslySkipPermissionFiltering, so role is moot.

### DIFF
--- a/front/lib/auth.ts
+++ b/front/lib/auth.ts
@@ -1034,11 +1034,13 @@ export class Authenticator {
   }
 
   /**
-   * Creates an Authenticator for a given workspace (with role `builder`). Used for internal calls
+   * Creates an Authenticator for a given workspace (with role `user`). Used for internal calls
    * to the Dust API or other functions, when the system is calling something for the workspace.
+   * Only the workspace global group is granted; use `internalAdminForWorkspace` when broader
+   * access is required.
    * @param workspaceId string
    */
-  static async internalBuilderForWorkspace(
+  static async internalUserForWorkspace(
     workspaceId: string
   ): Promise<Authenticator> {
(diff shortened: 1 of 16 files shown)


front/temporal/sandbox_reaper/activities.ts
```diff
@@ -119,7 +119,7 @@ async function fetchAuthMaps(
         workspaceModelIdsByOwnerKind.conversation.has(workspace.id)
       ),
       async (workspace) => {
-        const authenticator = await Authenticator.internalBuilderForWorkspace(
+        const authenticator = await Authenticator.internalUserForWorkspace(
```
ykmsd, line 122: chatted with Matteo and this is safe because we don't rename the activity name

front/lib/auth.ts
ykmsd, line 1: I've never touched this file, so I deploy first sse and then front? 🤔

matteotrab (approved):

philipperolet (approved):
LGTM 👍

PR #31078, opened 2026-08-25
Cap MCP server tool credits per agent message

flvndvd:
## Description

This PR caps tool billing at 20 credits per MCP server for one agent message.

After a server reaches 20 billed tool credits, later calls to that same server in the same agent message are free.

This is a temporary solution. It lets us ship the cap without changing the current billing source of truth or the Metronome configuration.

## Exact billing behavior

- The cap is tracked separately for each MCP server.
- The cap resets for every agent message.
- LLM usage is unchanged.
- Tool execution is unchanged. Calls still run normally.
- Basic tools cost 1 credit.
- Advanced tools cost 3 credits.
- The call that reaches or crosses 20 credits is fully billed.
- Only later calls are free.
- Free tools and calls with an unbillable status do not count toward the cap.
- Calls without an identifiable external MCP server remain uncapped.

Examples:

- 20 basic calls cost 20 credits. The 21st call is free.
- 7 advanced calls cost 21 credits. The 8th call is free.
- Reaching the cap on server A does not make calls to server B free.

## Message resumes

An agent message can be interrupted and resumed several times.

When emitting usage for a resumed execution, we load all actions from the agent message. Calls from previous executions count toward the server cap. We still emit only the actions that belong to the current execution. Previous actions are not emitted again. Actions are ordered by their database ID so the result is stable for retries and parallel calls.

## Metronome changes

Metronome currently bills tool usage from `tool_use_v3` events using:
- count
- tool_category
- usage_type

Post-cap calls are emitted with `usage_type=free`. The existing free rate makes their billed amount zero. This does not require a new metric, product, or rate.

One execution can contain both paid and post-cap calls to the same tool. For example, it can contain 7 paid calls and 1 free call.

These calls cannot stay in one event because one event has only one `usage_type`. We therefore split the local event group by billing disposition.

The paid event keeps its existing transaction ID.

The post-cap free event receives a `-free_mcp_server_cap` suffix. It needs a different transaction ID because Metronome deduplicates events by transaction ID. Without the suffix, Metronome would drop either the paid event or the
free event.

This does not rewrite events already ingested before deployment. The cap applies correctly to newly emitted usage after deployment.

## Other billing records

The same billing calculation is used for:

- The `costCredits` stored on the agent message
- Metronome events
- Agent analytics
- Per-tool analytics
- Consumption attribution

Post-cap calls keep their rated cost for debugging, but their billed cost is zero.

The consumption attribution version is bumped so capped calls are stored with zero direct tool credits.

## Why this is temporary

The goal is to ship the protection quickly without making a large billing change.

When we change the billing source of truth, this rule should move there. The temporary event handling and attribution handling can then be removed.

### DIFF
--- a/front/lib/api/assistant/agent_message_consumption_attribution/attribution_builder.ts
+++ b/front/lib/api/assistant/agent_message_consumption_attribution/attribution_builder.ts
@@ -17,8 +17,9 @@ import assert from "assert";
 // context-window safety padding from tool footprints and uses o200k for GPT-5 provider accounting.
 // Version 6 excludes enabled-skill tool definitions when provider-side tool search keeps those
 // non-eager tools deferred, while continuing to attribute the enabled skill's instructions.
+// Version 7 records post-cap MCP server calls with zero direct credits.
 // Each version remains a separate, self-consistent set of rows.
-export const AGENT_MESSAGE_CONSUMPTION_ATTRIBUTION_VERSION = 6;
+export const AGENT_MESSAGE_CONSUMPTION_ATTRIBUTION_VERSION = 7;
 
 export type RunUsageForAttribution = Pick<
   RunUsageType,
--- a/front/lib/metronome/events.ts
+++ b/front/lib/metronome/events.ts
@@ -74,10 +74,13 @@ export function getUsageType(
   return isProgrammaticUsage ? USAGE_TYPE_PROGRAMMATIC : USAGE_TYPE_USER;
 }
 
-function getToolUsageType(
-  baseUsageType: UsageType,
-  isFreeUsage: boolean
-): UsageType {
+function getToolUsageType({
+  baseUsageType,
+  isFreeUsage,
+}: {
+  baseUsageType: UsageType;
+  isFreeUsage: boolean;
+}): UsageType {
   return isFreeUsage ? USAGE_TYPE_FREE : baseUsageType;
 }
 
@@ -351,10 +354,10 @@ export function buildToolUseEvents({
 
   return [...groups.values()].map(({ billingLine, count, totalDurationMs }) => {
     const { action, billingDisposition, toolCostCategory } = billingLine;
-    const effectiveUsageType = getToolUsageType(
-      usageType,
-      billingDisposition !== "billed"
-    );
+    const effectiveUsageType = getToolUsageType({
+      baseUsageType: usageType,
+      isFreeUsage: billingDisposition !== "billed",
+    });
     const transactionIdDispositionSuffix =
       billingDisposition === "free_mcp_server_cap"
         ? `-${billingDisposition}`
--- a/front/lib/metronome/events.ts
+++ b/front/lib/metronome/events.ts
@@ -76,9 +76,9 @@ export function getUsageType(
 
 function getToolUsageType(
   baseUsageType: UsageType,
-  isFreeTool: boolean
+  isFreeUsage: boolean
 ): UsageType {
-  return isFreeTool ? USAGE_TYPE_FREE : baseUsageType;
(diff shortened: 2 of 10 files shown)


front/lib/metronome/events.ts
```diff
@@ -320,13 +329,16 @@ export function buildToolUseEvents({
...
+      continue;
+    }
     // Metronome prices every emitted tool event. Actions that never reached the
     // tool must therefore be omitted rather than represented as zero-cost.
     if (billingLine.billingDisposition === "unbillable_status") {
       continue;
     }
     const { action } = billingLine;
-    const key = `${action.toolName}|${action.internalMCPServerName ?? ""}|${action.mcpServerId ?? ""}|${action.status}`;
+    const key = `${action.toolName}|${action.internalMCPServerName ?? ""}|${action.mcpServerId ?? ""}|${action.status}|${billingLine.billingDisposition}`;
```
flvndvd, line 341: The grouping key changed in this PR is only an in-memory Map key. It is not a Metronome configuration key.

front/lib/metronome/events.ts
```diff
@@ -342,13 +354,17 @@ export function buildToolUseEvents({
...
+      isFreeUsage: billingDisposition !== "billed",
+    });
+    const transactionIdDispositionSuffix =
+      billingDisposition === "free_mcp_server_cap"
+        ? `-${billingDisposition}`
+        : "";
     return {
       transaction_id: truncateTransactionId(
-        `tool3-${workspaceId}-${conversationId}-${agentMessageId}-${runKey}-${action.toolName}-${action.mcpServerId ?? ""}-${action.status}`
+        `tool3-${workspaceId}-${conversationId}-${agentMessageId}-${runKey}-${action.toolName}-${action.mcpServerId ?? ""}-${action.status}${transactionIdDispositionSuffix}`
```
flvndvd, line 367: See PR description for why we need this.

front/lib/api/assistant/agent_message_consumption_attribution/attribution_builder.ts
```diff
@@ -17,8 +17,9 @@ import assert from "assert";
 // context-window safety padding from tool footprints and uses o200k for GPT-5 provider accounting.
 // Version 6 excludes enabled-skill tool definitions when provider-side tool search keeps those
 // non-eager tools deferred, while continuing to attribute the enabled skill's instructions.
+// Version 7 records post-cap MCP server calls with zero direct credits.
 // Each version remains a separate, self-consistent set of rows.
-export const AGENT_MESSAGE_CONSUMPTION_ATTRIBUTION_VERSION = 6;
+export const AGENT_MESSAGE_CONSUMPTION_ATTRIBUTION_VERSION = 7;
```
id13, line 22: OOC what are the ripple effects of incrementing this ?

flvndvd: Not too much of a ripple effect. In-flight messages will recompute their previous attribution items, previous ones remain valid. The main pain that this one solves is ensuring that the data team, can know which version does what as it changes their graph.

front/lib/metronome/events.ts
tdraier: still unclear why we would need to send free_mcp_server_cap to metronome if it's free ? these one should just be skipped ? anyway anything which is not tool_category = basic || advanced is ignored .

tdraier (approved):

front/lib/metronome/events.ts
flvndvd: Just wanted to align, and it's a way to double check it works as expected. I presume those will be removed once we consolidate events we send to Metronome?

PR #31080, opened 2026-08-25
Preserve reasoning summary boundaries while streaming

fontanierh:
## Description

OpenAI concise reasoning can contain multiple summary parts. Preserve the blank-line boundary between those parts in the live delta stream so adjacent bold titles do not collapse into visible asterisks before completion.

## Tests

Added a converter regression covering two streamed reasoning summary parts and the completed item.

## Risk

Low. This only inserts the separator already used by the completed reasoning representation.

## Deploy Plan

Standard deploy.


adrsimon (commented):
LGTM

PR #31081, opened 2026-08-25
fix: reject creation of restricted internal MCP servers

adrsimon:
## Description

`createInternalMCPServer` checks that the posted name is a known internal server and stops there. It never evaluates `isRestricted`, so any workspace admin can install a flag-gated or plan-gated server by posting its name to `/api/w/:wId/mcp`. The only thing that normally keeps those servers out of reach is the "add a tool" dialog in the space administration, which filters on `isRestricted` as it renders. It's just a display filter, not an auth check.

This moves the check onto the create path. `isRestrictedForWorkspace` now gate the input to the actual accessible list of servers to the workspace (feature flags, deep dive, workspace analytics, plan).

## Tests

None

## Risk

Workspaces already running a restricted server are unaffected, since this only touches creation.

## Deploy Plan

Deploy front.

### DIFF
--- a/front/lib/api/mcp/servers.ts
+++ b/front/lib/api/mcp/servers.ts
@@ -280,6 +280,15 @@ export async function createInternalMCPServer(
     return new Err(new Error("Invalid internal MCP server name"));
   }
 
+  const isRestricted =
+    await InternalMCPServerInMemoryResource.isRestrictedForWorkspace(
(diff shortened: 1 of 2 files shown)


aubin-tchoi (approved):
LGTM

spolu (approved):

PR #31087, opened 2026-08-25
[front] use consumption ES index for agents export

avervaet:
## Description

The agents table in the analytics export panel still read from the old, message-level Elasticsearch index, while the rest of the consumption dashboard (top agents, timeseries, etc.) already reads from the newer consumption index. This moves the agents export over too, for both places it's produced: the dedicated agents-export endpoint and the `agents` table of the combined export endpoint.

Because the consumption index stores data at a finer grain (per LLM step / tool call rather than per message), the metrics are now computed with distinct-count aggregations instead of raw document counts, mirroring how the rest of the consumption dashboard already does it.

## Tests

- Updated the existing agents-export test to the new data source.
- Added a regression test for the combined export endpoint's `agents` table, covering a bug caught in review where its date range excluded the last day of the period.
- Confirmed the neighboring export-endpoint test suite (which doesn't touch Elasticsearch directly) still passes.

## Risk

Low - The other tables in the same export panel still read from the old index, so the panel now sources different tables from different indices. Not expected to cause visible discrepancies, but worth knowing if reconciling numbers across CSVs.

## Deploy Plan

- Deploy front

### DIFF
--- a/front-api/routes/w/[wId]/analytics/agents-export.ts
+++ b/front-api/routes/w/[wId]/analytics/agents-export.ts
@@ -4,8 +4,9 @@ import {
   fetchAgentExportRows,
   toAgentExportCsvRow,
 } from "@app/lib/api/analytics/agents_export";
+import { resolveConsumptionPeriod } from "@app/lib/api/analytics/consumption/period";
+import { buildConsumptionScopeQuery } from "@app/lib/api/analytics/consumption/scope";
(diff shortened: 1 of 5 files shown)


avervaet:

- #31087 ◀

aubin-tchoi (approved):
LGTM assuming we backfilled for the entire selectable period

PR #31091, opened 2026-08-25
Add Shopify OAuth provider

spolu:
## Description

- Add Shopify as a first-class OAuth provider in front and core.
- Validate permanent myshopify.com store domains and Shopify callback HMAC signatures.
- Request the app-configured read_all_orders, read_customers, read_orders, and read_products scopes in the authorization URL.
- Exchange authorization codes for expiring offline tokens, refresh access tokens, preserve granted scope metadata, and scrub tokens from stored raw responses.

## Tests

- 8 focused front OAuth tests passed.
- 2 Shopify core provider tests passed.
- Shopify OAuth flow completed locally using front-spa on port 10011 and front-api on port 10000.

## Risk

Medium. This adds a new external OAuth and token-refresh path, but no Shopify MCP server or tools are enabled by this PR.

## Deploy Plan

- Configure OAUTH_SHOPIFY_CLIENT_ID and OAUTH_SHOPIFY_CLIENT_SECRET for front/front-api and the core OAuth service.
- Deploy the core OAuth service.
- Deploy front and front-api.


Fraggle (approved):
LGTM but same question as I asked the last person doing a first class oauth. Do we really need one (instead of using mcp / mcp_static ?)

Asking to have less code mainly.

spolu:
@Fraggle wanted to ask you that because I'm not super familiar. What is sure is that this is not an MCP server at all, their app setup is quite weird. The one thing is the custom domain to use. But may that could fly with the default mcp version? I have no clue?

Fraggle:
I assume you are talking about [this](https://github.com/dust-tt/dust/pull/31091/changes#diff-2db9471b5592c5be34d9e601e16f0514785e208cc842ef89b3507ab00aee12feR74) ?

After checking it's indeed non standard therefor I think it warrant a custom oauth implementation or some way to customize it to add the hmac verification and building the dynamic redirect url (but we dont want to do that).

For completeness, mcp_static (or mcp, the distinction is useless from the oauth service POV) is really just a generic oauth implementation, likely more and more complete as our users add remote mcp servers. We could (and i thiiink we do) use it for any oauth server that is not "special" (like github or shopify).

spolu:
Very clear 👍

spolu:
Note on HMAC validation, quite desirable since the URL is user provided (and validated of course) so it's desirable defence in depth.

spolu:
Secrets updated on both front and oauth

PR #31095, opened 2026-08-25
[front] use consumption ES index for users export

avervaet:
## Description

Same move as the agents export ([#31087](https://github.com/dust-tt/dust/pull/31087)), applied to the users table: it read from the old, message-level Elasticsearch index while the rest of the consumption dashboard already reads from the newer consumption index. This switches the users export over too, for both places it's produced: the dedicated users-export endpoint and the `users` table of the combined export endpoint.

Same reasoning as before: the consumption index stores data at a finer grain, so metrics are computed with distinct-count aggregations rather than raw document counts.

## Tests

- Updated the existing users-export test to the new data source.
- Added a regression test for the combined export endpoint's `users` table, covering the same date-range bug class caught in the agents PR.
- Confirmed the neighboring export-endpoint test suites (100 tests across `front-api/routes/.../analytics`) still pass.

## Risk

- Same as the agents PR: other tables in the export panel still read from the old index, so the panel now sources different tables from different indices.

## Deploy Plan

Standard deploy, no migration or feature flag needed. Stacked on top of #31087.

### DIFF
--- a/front/lib/api/analytics/users_export.ts
+++ b/front/lib/api/analytics/users_export.ts
@@ -1,4 +1,15 @@
-import { bucketsToArray, searchAnalytics } from "@app/lib/api/elasticsearch";
+import {
+  COMPLETED_AT_FIELD,
+  CONSUMPTION_DIMENSION_FIELDS,
+  CREDIT_MICRO_FIELD,
+  MAX_EXPORT_TERMS_SIZE,
+  uniqueMessagesCardinalityAgg,
+} from "@app/lib/api/analytics/consumption/scope";
+import {
+  bucketsToArray,
+  searchConsumptionAnalytics,
+} from "@app/lib/api/elasticsearch";
+import { microCreditsToCredits } from "@app/lib/credits/units";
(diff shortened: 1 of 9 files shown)


avervaet:

- #31087
    - #31095 ◀
        - #31097
            - #31098

aubin-tchoi (approved):
LGTM

front/lib/api/analytics/users_export.ts
```diff
@@ -73,21 +84,19 @@ export async function fetchUserExportRows({
     {
       aggregations: {
         by_user: {
-          terms: { field: "user_id", size: 10000 },
+          terms: { field: CONSUMPTION_DIMENSION_FIELDS.user, size: 10000 },
```
aubin-tchoi: nit: magic variable

avervaet: Fixed — extracted `MAX_EXPORT_TERMS_SIZE` in consumption/scope.ts.

PR #31097, opened 2026-08-25
[front] use consumption ES index for skills export

avervaet:
## Description

Continuing the move started in #31087 and #31095: the skills table in the analytics export panel read from the old, message-level Elasticsearch index while the rest of the consumption dashboard already reads from the newer consumption index. This switches it over too.

Skill attribution is stored differently between the two indices: the old index nests a `skills_used` array under each message document, while the consumption index has a flat, multi-valued field directly on tool-call documents. The query to find which skills saw any activity in the period is simpler as a result — a plain lookup instead of a nested one — but the result is the same set of skill ids.

The CSV output and what the user sees is unchanged — only the underlying data source changed.

## Tests

- Added a regression test for the combined export endpoint's `skills` table, covering the same date-range bug class caught in the earlier PRs.
- Confirmed the neighboring export-endpoint test suite still passes.

## Risk

- Same as the earlier PRs: other tables in the export panel still read from the old index, so the panel now sources different tables from different indices.

## Deploy Plan

Standard deploy, no migration or feature flag needed. Stacked on top of #31095.

### DIFF
--- a/front/lib/api/analytics/export_tables.ts
+++ b/front/lib/api/analytics/export_tables.ts
@@ -45,6 +45,7 @@ import type { Result } from "@app/types/shared/result";
 import { Err, Ok } from "@app/types/shared/result";
 import { assertNever } from "@app/types/shared/utils/assert_never";
 import type { WorkspaceType } from "@app/types/user";
+import type { estypes } from "@elastic/elasticsearch";
 import moment from "moment-timezone";
 
 type AnalyticsExportTable =
@@ -211,7 +212,7 @@ export async function exportTable({
         includeHiddenAgents,
       });
     case "users":
-      return exportUsers({ startDate, endDate, timezone, owner });
+      return exportUsers({ auth, startDate, endDate, timezone, owner });
     case "skills":
       return exportSkills({ auth, startDate, endDate, timezone, owner });
     case "skill_usage":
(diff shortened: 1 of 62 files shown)


avervaet:

- #31087
    - #31095
        - #31097 ◀
            - #31098

front/lib/api/analytics/skills_export.ts
```diff
@@ -9,6 +13,44 @@ import { Err, Ok } from "@app/types/shared/result";
...
+// aggregation is enough to list which skills saw any activity in the period.
+async function fetchUsedConsumptionSkillIds(
+  baseQuery: estypes.QueryDslQueryContainer
+): Promise<Result<string[], Error>> {
+  const result = await searchConsumptionAnalytics<never, UsedSkillsAggs>(
+    baseQuery,
+    {
+      aggregations: {
+        by_skill_id: {
+          terms: { field: CONSUMPTION_DIMENSION_FIELDS.skill, size: 1000 },
```
aubin-tchoi: nit: magic variable

front/lib/api/analytics/export_tables.ts
```diff
@@ -492,16 +492,14 @@ async function exportSkills({
...
 }: {
   auth: Authenticator;
   startDate: string;
   endDate: string;
   timezone: string;
-  owner: WorkspaceType;
 }): Promise<Result<ExportTableData, Error>> {
-  const baseQuery = buildAgentAnalyticsBaseQuery({
-    workspaceId: owner.sId,
+  const baseQuery = buildExportConsumptionScopeQuery({
```
aubin-tchoi: nit: we usually pass auth as the first, non-nested field

aubin-tchoi (approved):
LGTM

front/lib/api/analytics/skills_export.ts
avervaet: Fixed — extracted `MAX_USED_SKILL_IDS`.

front/lib/api/analytics/export_tables.ts
avervaet: Fixed — `buildExportConsumptionScopeQuery` now takes `auth` as a first, non-nested argument (also fixed at the agents/users/source call sites).

PR #31098, opened 2026-08-25
[front] use consumption ES index for source export

avervaet:
## Description

Continuing the move started in #31087, #31095 and #31097: the message-source table in the analytics export panel read from the old, message-level Elasticsearch index while the rest of the consumption dashboard already reads from the newer consumption index. This switches it over too, for both places it's produced: the dedicated source-export endpoint and the `source` table of the combined export endpoint.

Same reasoning as the earlier PRs: the consumption index stores data at a finer grain, so the per-day message count is computed with a distinct-count aggregation rather than a raw document count. The origin values themselves (the raw, unmerged ones, e.g. "slack" vs "slack_workflow" as separate rows) are preserved exactly as before.

## Tests

- Added a regression test for the combined export endpoint's `source` table, covering the same date-range and count-dedup bug classes caught in the earlier PRs.
- Confirmed the neighboring export-endpoint test suites still pass, including the untouched `contextOriginFilter` tests for the sibling function that still reads from the old index.

## Risk

- Same as the earlier PRs: other tables in the export panel still read from the old index, so the panel now sources different tables from different indices.

## Deploy Plan

Standard deploy, no migration or feature flag needed. Stacked on top of #31097.

### DIFF
--- a/front-api/routes/w/[wId]/analytics/agents-export.ts
+++ b/front-api/routes/w/[wId]/analytics/agents-export.ts
@@ -4,8 +4,7 @@ import {
   fetchAgentExportRows,
   toAgentExportCsvRow,
 } from "@app/lib/api/analytics/agents_export";
-import { resolveConsumptionPeriod } from "@app/lib/api/analytics/consumption/period";
-import { buildConsumptionScopeQuery } from "@app/lib/api/analytics/consumption/scope";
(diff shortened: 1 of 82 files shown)


avervaet:

- #31087
    - #31095
        - #31097
            - #31098 ◀

aubin-tchoi (approved):
LGTM

PR #31100, opened 2026-08-25
[front] fix: three retry-boundary bugs in the agent loop

Nils-Fedrigo:
## Description

Three independent bug fixes in the agent loop, all pre-existing on `main`, all sharing a root cause: something re-runs a model step and the previous attempt's leftovers are not cleaned up. Grouped because they are prerequisites for the auto-stream failover work stacked on top of this PR, which turns each of them from rare into routine.

**1. Step contents of a superseded attempt survived a re-run.** Versioning is per `(step, index)`, so a re-run that emits *fewer* contents than the previous attempt left that attempt's trailing indexes behind as the latest version of their own index — two runs mixed into one message. `AgentStepContentResource.createNewVersions` now accepts `replacesStep` for writes that carry a step's whole output (as opposed to appending a single content, like the error content published on failure) and drops the trailing rows. Rows referenced by a tool execution are deliberately left alone: the FK is `RESTRICT`, and a referenced content means that attempt's tools actually ran, which is not ours to unwind. The existing-content query widens from specific indexes to the whole step — same index prefix, same cost — which is what makes the trailing rows visible.

**2. The streaming hook only detected a retry boundary on a chain-of-thought event.** A retry that emitted no CoT (a `none`-effort model) never tripped the boundary, so the new call's answer was appended to the dead call's partial text. `lastCoTTraceId` becomes `lastTraceId`, tracked on any generation classification. On a plain-token boundary the unflushed chain of thought is also cleared, so the classification transition does not commit the dead call's thinking as an activity step. CoT shadow-buffer behaviour is unchanged.

**3. Reasoning was replayed to whichever provider the payload was headed to.** A thinking signature or reasoning-item id is only valid for the provider that produced it. `toBaseMessages` now takes the target provider and drops reasoning that came from another one; text and tool calls are provider-agnostic and always kept.

## Tests

Fix 3 is a **no-op on current behaviour** — the provider cannot change mid-message today — so it is covered by the updated `transitionLLM.test.ts` assertions alone.

Fixes 1 and 2 are worth exercising manually, since both need a step to actually re-run:

- **Trailing step contents**: get an agent step to retry after emitting contents (e.g. kill the worker mid-stream) and confirm the message ends up with only the final attempt's contents, no interleaving. Then confirm the guard: if a tool execution references a trailing content, the row must survive and the `"Trailing step contents are all referenced by a tool execution, keeping them"` warning must fire.
- **Stream boundary**: force a Temporal retry on a step and confirm the answer *replaces* the dead call's partial text rather than appending to it, and that inline activity steps for that step are dropped. Worth checking on a `none`-effort model, which is the case the old code missed entirely.

## Risk

Fix 1 deletes rows, which is the only real risk here. It is scoped to a single `(workspaceId, agentMessageId, step)` and only to indexes past the one just written, it skips anything a tool execution references, and it runs only on the `replacesStep: true` call site. The blast radius of a bug is a message losing trailing contents of a step that was being rewritten anyway. Cache eviction alongside it is best-effort — a stale Redis field is re-read from PG on the next miss.

Fixes 2 and 3 are safe to roll back independently; nothing else depends on them at this point in the stack.

### DIFF
--- a/front/hooks/useAgentMessageStream.ts
+++ b/front/hooks/useAgentMessageStream.ts
@@ -379,8 +379,11 @@ export function useAgentMessageStream({
   // thinking (chain_of_thought) and writing (tokens), flushing completed
   // segments as activity steps on each switch.
   const lastClassification = useRef<"tokens" | "chain_of_thought" | null>(null);
-  // Tracks the traceId of the last CoT event to detect Temporal retry boundaries.
-  const lastCoTTraceId = useRef<string | null>(null);
+  // Tracks the traceId of the last generation event to detect Temporal retry
+  // boundaries. Any classification counts: a retry that emits no chain of thought
+  // (a `none`-effort model, e.g. after an auto-stream failover) must reset the
+  // accumulators too, otherwise its answer is appended to the dead call's text.
+  const lastTraceId = useRef<string | null>(null);
   // Tracks the agent-loop step of the last generation event. Combined with a
   // traceId change, an unchanged step number means the SAME step was re-run
   // (Temporal activity retry) rather than the loop advancing to a new step. It is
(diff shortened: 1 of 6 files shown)


Nils-Fedrigo:

- #31100 ◀
    - #31101
        - #31102
            - #31103

PR #31101, opened 2026-08-25
[front] add lastStreamModelResolutions to conversations

Nils-Fedrigo:
## Description

Adds a `lastStreamModelResolutions` jsonb column to `conversations`, and nothing else. The column is **inert at this commit** — no code reads or writes it. The follow-ups in the stack do.

It will hold, per conversation, the model each `(agent, stream)` pair last resolved to, so an `auto` stream can stay on one model across messages instead of walking its candidate pool from the top on every message.

Denormalised on the conversation row rather than read back from the messages because every caller that resolves a model already holds the conversation, which makes the lookup free. The alternative — a "latest stream-resolved agent message" query joining `messages` to `agent_messages` — costs a round-trip on the message-post path, and on a miss (conversation not on that stream, different agent, or first message, i.e. the common case) walks every message in the conversation before returning null.

Deliberately kept to the migration, the Sequelize column, and the one type its declaration needs, so the PR carrying the schema change is trivial to review and to revert.

## Deploy Plan

Pre-deploy migration — must be applied **before** the code above it in the stack goes out.

`ADD COLUMN` with a non-volatile default does not rewrite the table (PG 11+ stores it as a missing value), so this is safe on `conversations` at its current size. The DB-level `DEFAULT '{}'::jsonb` is what keeps inserts from the currently deployed code valid, which is why the column is `NOT NULL` with a default rather than nullable.

No backfill: an empty map simply means "no hint yet", and the resolver falls back to walking the pool.

### DIFF
--- a/front/lib/models/agent/conversation.ts
+++ b/front/lib/models/agent/conversation.ts
@@ -23,6 +23,7 @@ import type {
   ParticipantActionType,
   UserMessageOrigin,
 } from "@app/types/assistant/conversation";
+import type { StreamModelResolutions } from "@app/types/assistant/models/auto";
 import type { ModelResolutionMethodType } from "@app/types/assistant/models/types";
 import { MODEL_RESOLUTION_METHODS } from "@app/types/assistant/models/types";
 import type { ModelId } from "@app/types/shared/model_id";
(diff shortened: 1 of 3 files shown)


Nils-Fedrigo:

- #31100
    - #31101 ◀
        - #31102
            - #31103

front/lib/models/agent/conversation.ts
```diff
@@ -99,6 +105,11 @@ ConversationModel.init(
       allowNull: false,
       defaultValue: {},
     },
+    lastStreamModelResolutions: {
+      type: DataTypes.JSONB,
```
id13, line 109: Why a jsonb is necessary here ? why can't we just save the last one ?

Nils-Fedrigo: Yeah we can probably -- sorry for the useless review, I won't go this route in the end... :/  see https://dust4ai.slack.com/archives/C050SM8NSPK/p1787664659465699?thread_ts=1787601925.357949&cid=C050SM8NSPK

PR #31102, opened 2026-08-25
[front] feat: keep auto-stream conversations on one model

Nils-Fedrigo:
## Description

Behind the new `auto_stream_model_routing` flag (`dust_only`).

An `auto` stream (Basic / Standard / Premium) resolves at message-send time by walking an ordered candidate pool and taking the first model available to the workspace. That walk started from the top on **every** message, so a conversation could hop models mid-thread as availability shifted.

Each stream-resolved agent message now records what it landed on in the conversation's `lastStreamModelResolutions` (the column added in #31101), and the next message prefers that model.

The hint is scoped to one agent and one stream — a Premium turn must not inherit what Standard picked, and two agents in the same conversation each keep their own resolution. It is only honoured while the exact `(providerId, modelId, reasoningEffort)` triple is still a candidate of the stream *and* still available to the workspace, so a stream that changes the effort it runs a model at invalidates its own hints instead of freezing the old effort. That same check is what makes the stored value safe to use without validating it: anything stale or malformed fails to match a pool candidate and resolves to no hint at all.

Attribution is unchanged — `modelResolutionMethod` still holds the stream id, so the message is still an `auto` message for tiering, pricing and analytics.

Two implementation notes worth a reviewer's attention:

- **Reads cost no query on the paths that matter.** Posting, editing and retrying a message all hold the `ConversationResource`, so the resolutions come off the row already in memory. `fetchStreamModelResolutions` accepts `ConversationWithoutContentType | ConversationResource` (matching the existing convention in that resource) and only falls back to a single-column primary-key read for callers that hold the plain type — currently just message finalisation, which is off the user-facing path.
- **Writes go through a jsonb `||` merge** rather than a read-modify-write, so concurrent writers touching different keys of the same conversation cannot clobber each other. This matters because the failover path stacked on top writes outside the post transaction and its conversation lock.

The resolutions are intentionally **not** added to `ConversationWithoutContentType`: that type is part of API response shapes (conversation search, spaces), and routing state has no business shipping to clients.

## Tests

Flag is off by default, so the behaviour needs the flag enabled on a workspace to observe.

- Send several messages to an agent on an `auto` stream and confirm every message resolves to the same model, where previously each one re-walked the pool.
- Confirm the scoping: switch the same conversation between Standard and Premium and confirm each stream keeps its own resolution rather than inheriting the other's; mention two different agents in one conversation and confirm the same.
- Confirm invalidation: make the previously resolved model unavailable to the workspace (or change the effort the stream runs it at) and confirm the next message falls back to a normal pool walk instead of failing or freezing on the stale entry.

## Risk

Flag-gated, and the hint degrades safely: any path that cannot honour it falls back to today's behaviour of walking the pool from the top. The stored value is never trusted to name a usable model — it only ever selects among candidates the resolver would have considered anyway, so a corrupt or stale entry cannot route a message to a model the workspace should not get.

### DIFF
--- a/front/lib/api/assistant/conversation.ts
+++ b/front/lib/api/assistant/conversation.ts
@@ -827,6 +827,7 @@ export async function postUserMessage(
   let modelResolution = mentionedAgentConfiguration
     ? await resolveModelForMentionedAgent(auth, {
         configuration: mentionedAgentConfiguration,
+        conversation: conversationResource,
         selection: modelSelection,
       })
     : null;
@@ -1248,6 +1249,7 @@ export async function editUserMessage(
   let modelResolution = mentionedAgentConfiguration
     ? await resolveModelForMentionedAgent(auth, {
         configuration: mentionedAgentConfiguration,
+        conversation: conversationResource,
         selection: message.requestedModel ?? undefined,
       })
     : null;
@@ -1826,6 +1828,7 @@ export async function retryAgentMessage(
       }
     : await resolveModelForMentionedAgent(auth, {
         configuration: message.configuration,
+        conversation: conversationResource,
       });
 
   const user = auth.user();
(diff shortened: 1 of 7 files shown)


Nils-Fedrigo:

- #31100
    - #31101
        - #31102 ◀
            - #31103

PR #31103, opened 2026-08-25
[front] feat: fail over to the next model of an auto stream

Nils-Fedrigo:
## Description

Behind the `auto_stream_model_routing` flag and the `auto-stream-model-failover` patch.

Today, when a provider fails on an `auto` stream, the message spends its whole retry budget on that one provider and then surfaces an error the user cannot act on. This makes the message move to the next candidate of its stream instead.

**What decides eligibility.** The error's fault domain (`errorSource`), which the provider adapters already set — not a hand-maintained list of error types. Only failures attributed to the provider fail over; a malformed request, an oversized context, bad credentials, a refusal or our own stream time budget are attributed to us, and retrying those on another model just burns it too. A user- or agent-pinned model never fails over: that is a choice we must not silently swap, and `modelResolutionMethod` holds a stream id exactly when the model came from a stream.

**How far it moves.** The pool walk resumes *after* the model that failed, matched on the full `(providerId, modelId, reasoningEffort)` triple since the same model can appear twice in a pool at different efforts. A model no longer in the pool is treated as exhausted rather than restarting from the top, which would re-pick the candidates already ruled out — including the one that just failed. If the stream has nothing left, the error surfaces rather than falling back to a generic large model, which would leave the requested tier.

**How the step restarts.** The new model is persisted on the agent message and on the conversation (so the next message starts where this one recovered to, rather than back at the candidate that failed), then the activity fails with a non-retryable `ModelFailover` marker. Temporal resets the attempt counter per activity, so restarting the activity is precisely what gives the new model a full retry budget of its own; the run-model path re-reads the model from the database on every run, so no state needs threading through. `MAX_MODEL_FAILOVERS` bounds it, enforced in the activity (which owns the error published to the user once the budget is spent) and again in the workflow so a misbehaving activity cannot loop it.

Also splits the retry budget: a provider-attributed error on an auto stream now moves on after `RUN_MODEL_ATTEMPTS_BEFORE_FAILOVER` rather than spending all of `RUN_MODEL_MAX_RETRIES` on one model. Ineligible messages keep the full budget, unchanged.

## Tests

`shouldFailOverModel` is unit-tested (budget boundary, non-retryable errors moving on the first attempt, and that the failover threshold stays below the surface threshold).

The rest needs the flag and patch enabled and a forced provider failure:

- Force a 5xx / overloaded response from the first candidate of a stream and confirm the message completes on the next candidate, that the step's contents are the new model's only (not interleaved with the dead attempt's — this is what #31100 fixes), and that the UI shows the new answer replacing the partial one rather than appended.
- Confirm a **cross-provider** failover specifically, since that is the case that exercises reasoning being dropped rather than replayed under a signature the new provider would reject.
- Confirm the negative cases: a 400 / context-length / auth error must **not** fail over; a pinned (user- or agent-selected) model must not fail over; an exhausted stream must surface the provider error rather than silently leaving the tier.
- Confirm the budget: after `MAX_MODEL_FAILOVERS` the error surfaces, with the `"Model failover budget exhausted"` warning.

## Risk

Highest-risk PR in the stack — it changes agent-loop control flow — and correspondingly the most gated: both the feature flag and the `auto-stream-model-failover` patch must be on, and with either off the workflow takes exactly today's path.

The two failure modes worth watching: a message failing over when it should not (bounded — it retries on a different model and the user still gets an answer, just later), and the workflow looping on repeated failovers (bounded twice over, in the activity and in the workflow). Rollback is flipping the patch or the flag; no schema or data change is introduced here.

Note the deliberate behaviour change in what fails over versus a naive "retryable error" reading: `rate_limit_error` is attributed to us (it is our shared quota, not the provider being down) and so does **not** fail over, while `unknown_error` from a provider does. Happy to widen the 429 case if we would rather move to a different provider that does not share the quota.

### DIFF
--- a/front/lib/model_tiers/enabled_models.ts
+++ b/front/lib/model_tiers/enabled_models.ts
@@ -165,12 +165,23 @@ function isSameCandidate(
 }
 
 // Walks a stream's ordered candidate pool and picks the first one available
-// or a fallback large model
+// or a fallback large model. `afterModel` restricts the walk to the candidates after.
 export function resolveStreamModel(
   models: EnabledModelConfigurationType[],
-  streamId: ModelStreamIdType
+  streamId: ModelStreamIdType,
+  { afterModel }: { afterModel?: ResolvedRequestedModel } = {}
 ): StreamResolutionType {
-  for (const candidate of MODEL_STREAMS[streamId]) {
+  const pool = MODEL_STREAMS[streamId];
+
+  let candidates = pool;
+  if (afterModel) {
+    const currentIndex = pool.findIndex((candidate) =>
+      isSameCandidate(candidate, afterModel)
+    );
+    candidates = currentIndex === -1 ? [] : pool.slice(currentIndex + 1);
+  }
+
+  for (const candidate of candidates) {
     const model = models.find(
       (m) =>
         m.isSelectable &&
--- a/front/lib/resources/conversation_resource.ts
+++ b/front/lib/resources/conversation_resource.ts
@@ -898,6 +898,49 @@ export class ConversationResource extends BaseResource<ConversationModel> {
     };
   }
 
+  static async updateAgentMessageResolvedModel(
+    auth: Authenticator,
+    {
+      agentMessageModelId,
+      agentConfigurationId,
+      conversationModelId,
+      resolvedModel,
+      streamId,
+    }: {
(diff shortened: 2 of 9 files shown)


Nils-Fedrigo:

- #31100
    - #31101
        - #31102
            - #31103 ◀

PR #31105, opened 2026-08-25
[front] enh: allow deleting all legacy feature flags at once

davidebbo:
`scripts/delete_legacy_feature_flag.ts` could only delete one legacy flag name at a time. This adds an `--all` mode so the whole leftover set can be cleaned up in one pass.

`--all` discovers the targets with `FeatureFlagResource.countByFlagNameForAllWorkspaces()` (which already returns names no longer declared in `WHITELISTABLE_FEATURES_CONFIG`) and keeps only those failing `isWhitelistableFeature`. Both modes then converge on a single name → row-count map and a shared dry-run/delete loop, reusing the existing `deleteLegacyByName` which handles the workspace-isolation bypass and cache invalidation.

`--featureFlag` and `--all` are mutually exclusive and exactly one is required, so a bare invocation can't wipe every legacy flag by accident. The single-name path keeps its existing guard pointing at `disable_feature_flag_all_workspaces.ts` for flags that are still declared.

Verified locally, all read-only (no `--execute`):

- `--all` dry run listed the 15 legacy flags in my local DB, one "Would delete" line each with its workspace count.
- `--featureFlag projects` behaves as before.
- `--featureFlag projects --all` and no args both error out.
- `--featureFlag labs_transcripts` (still live) still errors with the `WHITELISTABLE_FEATURES` message.

### DIFF
--- a/front/scripts/delete_legacy_feature_flag.ts
+++ b/front/scripts/delete_legacy_feature_flag.ts
@@ -2,49 +2,86 @@ import { FeatureFlagResource } from "@app/lib/resources/feature_flag_resource";
 import { makeScript } from "@app/scripts/helpers";
 import { isWhitelistableFeature } from "@app/types/shared/feature_flags";
 
-// Deletes all rows for a given feature flag name from the feature_flags table.
+// Deletes all rows for legacy feature flag names from the feature_flags table.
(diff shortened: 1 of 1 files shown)


philipperolet (commented):
LGTM 👍

PR #31106, opened 2026-08-25
[front] Agent loop: log the tool identity at tool activity start

PopDaph:
## Description

Adding a temporary log to debug which tool fails activities 🙏🏻 

Heartbeat-timeout failures on `runToolActivity` are mostly unattributable to a tool today: in a 20-workflow sample of last week's failures, 17 died before any tool-level log fired (the first log naming the tool is `Starting MCP tool notification loop`, deep inside the MCP call path). 

This logs `toolName` / `mcpServerName` (plus actionId, attempt, conversation, step) right after the action fetch, before the pre-execution phases where those activities die, so future failures can be attributed by joining on conversationId.

## Tests

Log-only change. `tsgo --noEmit` clean.

## Risk

None: one info log per tool activity, comparable volume to the existing per-call MCP logs.

## Deploy Plan

Deploy front.

### DIFF
--- a/front/temporal/agent_loop/activities/run_tool.ts
+++ b/front/temporal/agent_loop/activities/run_tool.ts
@@ -169,6 +169,23 @@ export async function runToolActivity(
   // Heartbeating here as retrieving the agent loop data takes some time.
   heartbeat();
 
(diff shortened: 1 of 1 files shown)


aubin-tchoi (approved):
LGTM

PR #31107, opened 2026-08-25
Admin governance: allow regular group in space (group mode)

fabiencelier:
## Description

https://github.com/dust-tt/tasks/issues/10189

Allow to add regular manual group to space
The "provisioned group" mode is now just the "group" mode

Correctly reject invalid group type from both the API and the resource which was not the case before 

image

## Tests

tested locally

## Risk

can be reverted but will need a migration to drop the added group permissions

## Deploy Plan

deploy front


tdraier (approved):
🙏

PR #31115, opened 2026-08-25
feat: unify bulk selection into a floating bottom bar

adrsimon:
## Description

Bulk selection on `builder/agents`, `builder/skills`, `automations` and `usage` had three
implementations: two hand-rolled orange banners and one `ContentMessageInline`. All four now use `BulkSelectionBar`, which floats at the bottom of the page instead of pushing the table down, and owns the clear and select-all controls so pages only pass their own actions.

Capture d’écran 2026-08-25 à 16 05 34

## Tests

Manual, on all four pages: select rows, cross-page select-all, clear, and each bulk action. Checked the bar stays clear of the table while scrolling and that it renders in both themes.

## Risk

Presentation only, no change to selection state or to the bulk mutation endpoints. 

## Deploy Plan

Deploy front.


avervaet:
Wondering if it is intuitive for page with a lots of items as it ends at the bottom ?

adrsimon:
@avervaet it's floating on top when there's a lot of item, you can check on the deploy-app-preview

ClementAupiais (approved):
LGTM

front/components/shared/BulkSelectionBar.tsx
```diff
@@ -0,0 +1,62 @@
...
+}: BulkSelectionBarProps) {
+  if (selectedCount === 0) {
+    return null;
+  }
+
+  // `dark` pins the bar dark in both themes; it must stay on the wrapper so the
+  // `.dark .bg-modal-background` elevation shadow still matches a descendant.
+  return (
+    <div className="dark pointer-events-none sticky bottom-4 z-20 flex justify-center pt-4">
+      <div className="pointer-events-auto flex min-w-[min(792px,100%)] max-w-full animate-in flex-wrap items-center justify-between gap-x-4 gap-y-2 rounded-xl bg-modal-background p-4 text-foreground duration-200 ease-out fade-in slide-in-from-bottom-4 motion-reduce:animate-none">
```
aubin-tchoi: nit: could be wrapped with `cn`

aubin-tchoi: do we need a `min-w` this custom?
also I don't think we need the gap-y

aubin-tchoi (approved):
LGTM

PR #31117, opened 2026-08-25
[admin governance] migrate builder role to user

ykmsd:
## Description
This PR is my best attempt to migrate builder role to user in both workOS and our db 🥶

There are three [rate limits](https://workos.com/docs/reference/rate-limits) in WorkOS API:
- All requests: 6,000 requests per 60 seconds per API key
- Read /user_management/*: 1,000 requests per 10 seconds per environment
- Write /user_management/*: 500 requests per 10 seconds per environment

For both US and EU, we have around 6,000 active memberships with role builder (workspace number-wise: US 641 and EU 330). 

I tried to be ultra conservative and max request we can do is 100 requests per 10 seconds. 

Since we cannot batch update the workOS role from workOS UI, we need to update one by one from our side (there is no batch endpoint neither). 

Here is my understanding of how it works:
- First fetch all workspace ids who has active users with builder role from membership table (unless workspace ids are provided for test run)
-  Fetch workspace and users from these list, create an array to iterate
- For each user we call updateMembership from membershipResource


front/lib/resources/membership_resource.ts
```diff
@@ -601,7 +601,7 @@ export class MembershipResource extends BaseResource<MembershipModel> {
       MembershipResource.roleCacheKeyResolver(params)
   );
 
-  private static invalidateRoleCache = async (params: {
+  static invalidateRoleCache = async (params: {
```
ykmsd: is it allowed? 👶

tdraier: okay-ish :-)

front/migrations/20260825_migrate_builder_role_to_user.ts
```diff
@@ -0,0 +1,472 @@
...
+  // search-index workflow that `MembershipResource.updateMembershipRole` would trigger.
+  if (execute) {
+    await MembershipModel.update(
+      { role: "user" },
+      { where: { id: item.membershipId } }
+    );
+    await MembershipResource.invalidateRoleCache({
+      userModelId: item.userModelId,
+      workspaceModelId: item.workspaceModelId,
+    });
```
tdraier: using updateMembershipRole directly would be much more simple (it does all the job), but this is probably more optimal

tdraier (approved):

front/migrations/20260825_migrate_builder_role_to_user.ts
ykmsd: chatted IRL, it will create an audit log and launch a temporal workflow to index user search (I'm not super sure if we need it but it's safe to run since it's debounced) but we will use updateMembershipRole

front/migrations/20260825_migrate_builder_role_to_user.ts
```diff
@@ -0,0 +1,275 @@
...
+  try {
+    const res = await MembershipResource.updateMembershipRole({
+      user,
+      workspace,
+      newRole: "user",
+      // Never create a new membership: this backfill only flips existing active builders in place.
+      // With `allowTerminated: false`, a membership revoked in the race window between our query and
+      // now returns `membership_already_terminated` instead of taking `updateMembershipRole`'s
+      // create-a-fresh-membership branch (which only runs when `allowTerminated` is true).
+      allowTerminated: false,
```
ykmsd, line 104: actually this is by default false but just in case so we don't create a new membership 🤷‍♀️

matteotrab (approved):

PR #31122, opened 2026-08-25
fix: empty screen after adding favorite skill

matteotrab:
## Description

Fixes https://github.com/dust-tt/tasks/issues/10173 

When adding a skill to favorites, clicking on the "view" button showed an empty page.

The toast's **View** action used `action.href`. Sparkle renders that as a link built from `SparkleContext.components.link`, which the SPA sets to a React Router `<Link>`. But the toaster is mounted by `Notification.Area` in `RootLayout`, which sits **above** `<RouterProvider>` in `front-spa/src/app/App.tsx` — so toast content renders with no Router context, `<Link>` throws and because the toaster is outside the `ErrorBoundary` the error reached the React root and unmounted everything.

Replaced the `href` with an `onClick` that pushes through `useAppRouter()`.

## Tests

Manually.

## Risks

Low. Single call site behind the `skill_favorites` flag.

## Deploy Plan

Standard deployment.


matteotrab:

- #31122 ◀

aubin-tchoi (approved):
LGTM

PR #31124, opened 2026-08-25
refactor(front): route mention queries through MentionResource

achilleburah:
## Description

`MentionModel` was accessed directly from several `lib/api/*` modules, which our
resource-invariant rules disallow — any model access should go through a Resource.
`MentionResource` already existed for the inactive-agent anti-join but had no query
surface for those other call sites. This adds one method per distinct query shape
they need — `findByMessagesAndUser`, `listByMessageModelIds`, `updateStatus`,
`dismiss`, `deleteByMessageModelIds` — and swaps each `MentionModel` call for the
matching Resource method, on the same line, with the same control flow and query
shape as before.

## Tests

- New `mention_resource.test.ts`: one test per public method on `MentionResource`
  (`makeNew`, `findByMessagesAndUser`, `listByMessageModelIds`,
  `listAgentsNotMentionedSince`, `updateStatus`, `dismiss`,
  `deleteByMessageModelIds`, `delete`), including their optional filters.
- Ran the full suite for every touched file plus their neighbors (`mentions`,
  `messages`, `validate_agent_mention`, `validate_actions`, `conversation`,
  `mention_suggestions`, `inactivity`, `notifications`, `conversation_resource`) —
  all pass.

## Risk

Behaviour-preserving, mechanical swap. No query shape, cardinality, or control flow
changed.

## Deploy Plan

Standard deploy, no migration or feature flag needed.

### DIFF
--- a/front/lib/resources/mention_resource.ts
+++ b/front/lib/resources/mention_resource.ts
@@ -1,9 +1,11 @@
 import type { Authenticator } from "@app/lib/auth";
+import type { MentionStatusType } from "@app/lib/models/agent/conversation";
 import { MentionModel } from "@app/lib/models/agent/conversation";
 import { BaseResource } from "@app/lib/resources/base_resource";
 import { frontSequelize } from "@app/lib/resources/storage";
 import type { ReadonlyAttributesType } from "@app/lib/resources/storage/types";
 import logger from "@app/logger/logger";
+import type { ModelId } from "@app/types/shared/model_id";
 import type { Result } from "@app/types/shared/result";
 import { Ok } from "@app/types/shared/result";
 import type {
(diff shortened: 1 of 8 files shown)


front/lib/resources/mention_resource.ts
```diff
@@ -59,6 +88,69 @@ export class MentionResource extends BaseResource<MentionModel> {
     return new this(this.model, mention.get());
   }
 
+  /**
+   * The mention of this user across the given messages, optionally narrowed to one status, or null.
+   */
+  static async findByMessagesAndUser(
```
fabiencelier, line 94: Can you please add one unit test per public resource method to make sure they behave as expected

fabiencelier (approved):
Awesome !

PR #31126, opened 2026-08-25
[front] feat: rework usage page header and credit pool footer

avervaet:
## Description

Reworks the Usage page's header section.

Stacked on #31125.

## Risk

Low. Scoped to the flagged redesign component; the legacy page and default (flag-off) experience are unaffected.

## Deploy Plan

Deploy front. No coordination needed; still gated behind `usage_page_redesign`.

PR #31125, opened 2026-08-25
[front] feat: duplicate usage page behind usage_page_redesign flag

avervaet:
## Description

Stages the upcoming Usage page redesign by cloning the page into a `UsagePageRedesign` component and gating it behind a new `usage_page_redesign` feature flag, so the redesign can land incrementally without touching the live page for existing users.

First PR in a stack reworking the Usage page header and credit pool section.

## Stack overview

1. #31125 (this PR) — duplicate the usage page behind a feature flag
2. #31431 — fetch pool and members consumption data from Metronome
3. #31432 — rework the header and credit pool summary cards
4. #31433 — add a previous-cycles consumption table
5. #31434 — rework the members table
6. #31435 — add a read-only version of the page to Poke

## Tests

Type-checked; manually verified the legacy page renders unchanged with the flag off, and the (currently identical) duplicate renders with the flag on via the local dev feature-flag panel.

## Risk

Low. The flag defaults off for all workspaces, and the duplicated component is byte-identical to the legacy one at this stage, this PR only adds the gate, no visual or behavioral change.

## Deploy Plan

Deploy front.

PR #31127, opened 2026-08-25
[front] feat: show pool burn rate and on-target status on usage page

avervaet:
## Description

Adds a burn-rate estimate and on-target status to the redesigned Usage page's credit pool section:

- "`X` used credits/cycles": mean pool credits consumed per finalized billing cycle, averaged over the last up to 3 finalized Metronome invoices for the contract. Pool consumption per invoice is derived from `usage`-type line items whose `commit_id` matches the workspace's pool commit (same `DUST_CONTRACT_CREDIT_TYPE=pool` tag already used to compute the live pool balance). Shown with an info tooltip ("This is your estimated burn rate").
- A status chip next to the credit value — **On target** (highlight), **Off target** (info), or **Exhausted** (warning) — projecting whether the remaining pool balance will last until the contract's end date at the current burn rate. Hidden when there isn't enough data yet (no finalized cycle, or no known contract end date).
- The credit pool progress bar's fill color now follows the same status.

Only touches `UsagePageRedesign`, still behind the `usage_page_redesign` flag from #31125.

Stacked on #31126.

## Tests

Type-checked and formatted. Manually verified locally (flag on) that the burn-rate text, tooltip, and status chip render, and that the progress bar color follows the chip's status.

Not verified against a real finalized Metronome invoice — the mapping from invoice line items to pool-vs-personal-credit consumption is my best reading of the Metronome SDK's documented shape (`commit_id` scoped per line item), not something testable without a real or sandbox contract that has finalized invoices with mixed pool/personal-allowance funding. Worth a sanity check against real invoice data before this ships broadly.

## Risk

Low for existing users (still gated behind `usage_page_redesign`, off by default). Medium for the feature itself: if the invoice line-item assumption above doesn't hold, the burn-rate number and on-target status could be inaccurate rather than absent (though the "not enough data" fallback means it should simply not render if the underlying data is missing rather than fail loudly).

## Deploy Plan

Deploy front. No coordination needed; still gated behind `usage_page_redesign`. Recommend spot-checking the burn-rate figure against a real workspace's finalized invoices before enabling the flag broadly.


avervaet:

- #31125
    - #31126
        - #31127 ◀
            - #31134
                - #31155
                    - #31241

PR #31128, opened 2026-08-25
Extract conversation window rendering core

flvndvd:
## Description

This is the first preparatory PR for making the conversation window stateful across agent-loop steps.

Today, every model step pays the full context reconstruction cost again:
1. Hydrate the complete conversation and its actions from the database.
2. Render every message into the provider-facing format.
3. Tokenize the complete rendered history.
4. Replay the pruning logic to rebuild the fitted context window.

For multi-step agent runs, most of that work is repeated even though only the latest step changed. This is particularly expensive for long conversations and large tool results. And today both these costs directly hit our users and our DB.

## Target design

The goal is to leverage the recently introduced [CheckpointedConversationWindowState](https://github.com/dust-tt/dust/blob/main/front/lib/api/assistant/conversation_rendering/checkpointed_window_state.ts), so that after a successful step N, we will snapshot/persist its fitted conversation-window checkpoint in GCS. Step N+1 can then:

1. Load the checkpoint from step N.
2. Hydrate only the new step data required for rendering.
3. Render and tokenize that delta.
4. Append it to the restored conversation-window state.
5. Apply the existing pruning logic.
6. Persist the resulting checkpoint for the next step.

Checkpoint reuse will require strict continuity, including the expected previous step, agent message version, model, prompt and tool profile. Missing, invalid, incompatible, or non-contiguous checkpoints will use the existing full rendering path.

The full path remains the authoritative degraded fallback. Stateful rendering will initially be gated behind a feature flag and scoped to consecutive steps within the same agent message. Cross-agent-message reuse can be added later using the same storage model.

The checkpoint format and GCS layout will be explicitly versioned so that incompatible changes can safely start from a full render without requiring a database migration or changing Temporal workflow parameters.

## This PR

This PR only extracts the **lengthy** existing conversation-window implementation from into a new file.

It creates a dedicated module where the shared rendering, tokenization, and pruning pipeline can evolve without adding stateful conditionals throughout the current renderer or leaking checkpoint concerns into the agent loop.

The existing public interface remains unchanged.

There is:
- No behavior change
- No agent-loop call-site change
- No database or GCS access
- No feature flag
- No Temporal contract change

The implementation is a 98% detected copy. The remaining differences are formatter-only changes.


front/lib/api/assistant/conversation_rendering/index.ts
```diff
@@ -1,446 +1,7 @@
...
-    }))
-  );
-}
+export {
+  PREVIOUS_INTERACTIONS_TO_PRESERVE,
+  PRUNING_TARGET_CONTEXT_UTILIZATION,
+  renderConversationForModel,
+  TOKENS_MARGIN,
+  TOOL_DEFINITIONS_COUNT_ADJUSTMENT_FACTOR,
+} from "./conversation_window_core";
```
adrsimon, line 7: nit: alias (gotta find something to say lol)

adrsimon (approved):
LGTM code wise
need more time to digest the broader plan itself, but looks fair 👍

id13 (approved):
LGTM, assuming only a code move 👍

PR #31131, opened 2026-08-25
Add conversation window state snapshots

flvndvd:
## Description

See context in https://github.com/dust-tt/dust/pull/31128.

This PR is the second preparatory step toward stateful conversation windows across agent-loop steps.

Today, each model step reconstructs the full context window from conversation history. The target design persists the fitted window after step N, then restores it for step N+1 and processes only the newly added content.

Before introducing GCS or agent-loop wiring, we need a reliable persistence contract for the in-memory pruning state.

Persisting only the rendered messages is insufficient. The window must also remember which tool results are:
- Pending consumption by a later assistant message
- Eligible for pruning
- Already consumed or pruned

Losing this frontier could make resumed rendering prune differently from uninterrupted rendering. Which is the exact previous bug that this class is addresssing.

## This PR

This PR makes `CheckpointedConversationWindowState` serializable and restorable.

It adds:
- A Zod-validated snapshot schema
- An explicit snapshot format version
- `snapshot()` to produce a persistence-safe representation
- `restore()` to reconstruct the complete pruning state
- `appendToLatestInteraction()` to continue an agent interaction across model steps
- Validation of token-accounting and tool-result-phase invariants (be ready for some super zod schema!)

The snapshot contains the model-facing messages, their token counts, the pruning frontier, and the accumulated token accounting.

Runtime options such as the current pruning budget and logging details are supplied when restoring rather than persisted. Compatibility decisions such as model, prompt, tools, and step continuity will remain the responsibility of the future checkpoint-loading path.

### DIFF
--- a/front/lib/api/assistant/conversation_rendering/checkpointed_window_state.ts
+++ b/front/lib/api/assistant/conversation_rendering/checkpointed_window_state.ts
@@ -91,7 +91,7 @@ const assistantContentSchema = z.discriminatedUnion("type", [
         .object({
           reasoning: z.string().optional(),
           metadata: z.string(),
-          tokens: z.number().finite().nonnegative(),
+          tokens: z.number().int().nonnegative(),
           provider: z.string(),
           region: z.string().nullable().optional(),
         })
@@ -109,7 +109,7 @@ const assistantContentSchema = z.discriminatedUnion("type", [
     .strict(),
 ]);
 
-const tokenCountSchema = z.number().finite().nonnegative();
+const tokenCountSchema = z.number().int().nonnegative();
 const persistedMessageBaseSchema = z.discriminatedUnion("role", [
   z
     .object({
--- a/front/lib/api/assistant/conversation_rendering/checkpointed_window_state.ts
+++ b/front/lib/api/assistant/conversation_rendering/checkpointed_window_state.ts
@@ -14,6 +14,7 @@ import type {
 import logger from "@app/logger/logger";
 import type { Result } from "@app/types/shared/result";
 import { Ok } from "@app/types/shared/result";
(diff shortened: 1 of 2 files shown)


aubin-tchoi (approved):
LGTM but feels like we are adding a lot of types for types we almost already have, which may complexify making changes on the type

flvndvd:
> LGTM but feels like we are adding a lot of types for types we almost already have, which may complexify making changes on the type

Yes, but as we offload and save in ES went with a only consumes once validated, if fails, then fallback. Maybe there are lighter patterns.

front/lib/api/assistant/conversation_rendering/checkpointed_window_state.ts
```diff
@@ -45,6 +48,241 @@ type WindowInteraction = {
...
+    })
+    .strict(),
+  z
+    .object({
+      type: z.literal("reasoning"),
+      value: z
+        .object({
+          reasoning: z.string().optional(),
+          metadata: z.string(),
+          tokens: z.number().finite().nonnegative(),
```
adrsimon: number of tokens is finite but not integer ? i see below 

```
    retainedTokens: z.number().int().nonnegative(),
    totalTokensBefore: z.number().int().nonnegative(),
    prunedTokens: z.number().int().nonnegative(),
```

which are int but not finite
should this be int too ?

adrsimon (approved):
LGTM afaict

front/lib/api/assistant/conversation_rendering/checkpointed_window_state.ts
flvndvd: Indeed, let me align

PR #31132, opened 2026-08-25
[front] fix: enable toolsets without public API

aubin-tchoi:
## Description

The `toolsets__enable` tool currently calls the public conversation tools endpoint from inside `front`, adding an unnecessary SDK and HTTP round trip.

This prevents agents called via the API to use this (currently need to do this and finding myself blocked).

This PR moves the implementation to the tool's callback (fetch the convo and MCP server view, and upserts the view).

## Tests

- Tested locally.

## Risk

- Low.

## Deploy Plan

- Deploy front.


Fraggle (approved):
Good catch, i think in the past it would have created circular deps.

flvndvd (commented):
Looks legit, but will defer to @Fraggle the final stamp as I might be too short on context here on why did we use the public API 🤔

aubin-tchoi:
yep, probably fixed since the move to the tools metadata pattern

PR #31136, opened 2026-08-25
feat(models): add the model_degradations table and its accessors

Nils-Fedrigo:
## Description

First of a 3-PR stack introducing per-endpoint **model degradation**, so a single model can be taken out of the auto streams during a provider incident instead of blacklisting a whole provider.

This PR is storage only — one table and its resource. Nothing reads it until #31137 and nothing writes it until #31145.

- **`model_degradations`** (new table, pre-deploy migration): one row per degraded **endpoint** — `modelId + providerId + host`, unique together, row present = that endpoint is degraded. A row is an endpoint rather than a model because the same model is served from more than one host (Claude via `anthropic` and via `agent-platform`, Gemini via `google-ai-studio` and via `agent-platform`), and an incident usually hits one of them. No `region` column — each deployment region has its own database, and a host covers both its `eu` and `global` endpoints.
- **`ModelDegradations`**: `listDegradedEndpoints()`, a full table read (a handful of rows, and only during an incident), and `updateDegradedEndpoints()`, which applies a set of per-endpoint changes and **leaves endpoints it does not name alone**. That matters operationally: two people working the same incident from their own Poke tabs would otherwise clobber each other, since a whole-set write computed from a stale read silently restores whatever the other person just degraded. It clears the rows the request names and re-adds the degraded ones in a single transaction, so a named endpoint ends up in exactly its requested state whatever state it was in.

It is statics-only and returns plain objects: callers only ever read the set whole or toggle one endpoint, never a row identity. Deliberately **not** named `ModelDegradationResource` — the `Resource` suffix is for `BaseResource` subclasses, and this never instantiates itself.

No Redis layer: the only hot reader keeps its own 60s in-process cache (#31137), so caching here would add invalidation complexity for no propagation win.

`kill_switches` is untouched: it is exactly what it is on `main`.

## Tests

`npx tsgo --noEmit` clean on `front` and `front-api`.

No tests in this PR: it adds a table and two resource methods with no caller yet. Both are covered functionally by the API tests in #31145 (`front-api/routes/poke/degraded_models.test.ts`), which exercise the per-endpoint semantics end to end — degrading and restoring only the named endpoints, repeat requests being no-ops, and an endpoint named twice in one request settling on its last requested state.

## Risk

Low. New table with no reader and no writer; the pre-deploy migration only creates it, so there is nothing to backfill and no data to migrate. Rollback is a plain revert.

One thing to know for later: `updateDegradedEndpoints([])` returns early rather than issuing a query. An empty `Op.or` would clear the whole table, so the guard is load-bearing, not a micro-optimisation.

## Deploy Plan

Standard pre-deploy migration, then deploy — in that order, since the model is registered as soon as the code ships.

1. Run the pre-deploy migration `20260831125421_create_model_degradations.sql` (creates `model_degradations` and its unique index). Safe to run ahead of the deploy: no existing code touches the table.
2. Deploy `front` / `front-api`.
3. Nothing to verify in product behaviour — the table has no reader until #31137 ships. `SELECT count(*) FROM model_degradations;` returning 0 is the expected end state.

Per-region: the table lives in each region's own database, so the migration runs in every region.

### DIFF
--- a/front-api/routes/poke/kill.ts
+++ b/front-api/routes/poke/kill.ts
@@ -1,6 +1,15 @@
 import type { GetKillSwitchesResponseBody } from "@app/lib/api/poke/kill";
-import { KillSwitchTypeSchema } from "@app/lib/api/poke/kill";
-import { isKillSwitchType } from "@app/lib/poke/types";
+import {
+  KilledModelsSchema,
+  KillSwitchTypeSchema,
+} from "@app/lib/api/poke/kill";
+import { isKillableModelId } from "@app/lib/poke/killable_models";
+import {
+  isKillSwitchType,
+  killedModelIdsFromKillSwitches,
+  modelIdFromKillSwitchType,
+  modelKillSwitchType,
+} from "@app/lib/poke/types";
 import { KillSwitchResource } from "@app/lib/resources/kill_switch_resource";
 import { pokeApp } from "@front-api/middlewares/ctx";
 import type { HandlerResult } from "@front-api/middlewares/utils";
@@ -31,11 +40,76 @@ app.post(
         },
       });
     }
+
+    const modelId = modelIdFromKillSwitchType(type);
+    if (modelId !== null && !isKillableModelId(modelId)) {
+      return apiError(ctx, {
+        status_code: 400,
+        api_error: {
+          type: "invalid_request_error",
+          message: `The request body is invalid: ${modelId} cannot be killed`,
+        },
+      });
+    }
     if (enabled) {
       await KillSwitchResource.enableKillSwitch(type);
     } else {
       await KillSwitchResource.disableKillSwitch(type);
     }
+    return ctx.json({ success: true });
+  }
+);
+
+/**
+ * Replaces the set of killed models with the one submitted. Model kill switches
+ * are ordinary `kill_switches` rows, this just diffs them in one shot.
+ *
+ * @ignoreswagger
+ */
+app.post(
+  "/models",
+  validate("json", KilledModelsSchema),
+  async (ctx): HandlerResult<SuccessResponseBody> => {
+    const { killedModelIds } = ctx.req.valid("json");
+
+    const invalidModelIds = killedModelIds.filter(
+      (modelId) => !isKillableModelId(modelId)
+    );
+    if (invalidModelIds.length > 0) {
+      return apiError(ctx, {
+        status_code: 400,
+        api_error: {
+          type: "invalid_request_error",
+          message:
+            "The request body is invalid: " +
+            `${invalidModelIds.join(", ")} cannot be killed`,
+        },
+      });
+    }
+
+    const enabledKillSwitches =
+      await KillSwitchResource.listEnabledKillSwitches();
+    const currentlyKilledModelIds =
+      killedModelIdsFromKillSwitches(enabledKillSwitches);
+    const currentlyKilled = new Set(currentlyKilledModelIds);
+    const requested = new Set(killedModelIds);
+
+    const toEnable = killedModelIds
+      .filter((modelId) => !currentlyKilled.has(modelId))
+      .map(modelKillSwitchType);
+    const toDisable = currentlyKilledModelIds
+      .filter((modelId) => !requested.has(modelId))
+      .map(modelKillSwitchType);
+
+    // Bounded by the model catalog (~50 entries) and rarely more than a handful
+    // of rows actually change, so sequential writes are fine here.
+    for (const type of toEnable) {
+      await KillSwitchResource.enableKillSwitch(type);
+    }
+    for (const type of toDisable) {
+      await KillSwitchResource.disableKillSwitch(type);
+    }
+
     return ctx.json({ success: true });
   }
 );
--- a/front/lib/resources/model_degradation_resource.ts
+++ b/front/lib/resources/model_degradation_resource.ts
@@ -37,11 +37,6 @@ export class ModelDegradationResource extends BaseResource<ModelDegradationModel
     }));
   }
 
-  /**
-   * Clears the rows of every endpoint the request names, then re-adds the ones
-   * it wants degraded. Endpoints the request does not name are left alone, so
-   * two operators working the same incident do not clobber each other.
-   */
   static async updateDegradedEndpoints(
     updates: DegradedModelEndpointUpdateType[]
   ): Promise<void> {
--- a/front/lib/resources/model_degradation_resource.ts
+++ b/front/lib/resources/model_degradation_resource.ts
@@ -1,0 +1,89 @@
+import type {
+  DegradedModelEndpointType,
+  DegradedModelEndpointUpdateType,
+} from "@app/lib/model_constructors/types/degradations";
+import { BaseResource } from "@app/lib/resources/base_resource";
+import { frontSequelize } from "@app/lib/resources/storage";
+import { ModelDegradationModel } from "@app/lib/resources/storage/models/model_degradations";
+import type { ReadonlyAttributesType } from "@app/lib/resources/storage/types";
+import type { Result } from "@app/types/shared/result";
+import { Ok } from "@app/types/shared/result";
+import type { Attributes, ModelStatic } from "sequelize";
+import { Op } from "sequelize";
+
+// Attributes are marked as read-only to reflect the stateless nature of our Resource.
+// eslint-disable-next-line @typescript-eslint/no-unsafe-declaration-merging
+export interface ModelDegradationResource
+  extends ReadonlyAttributesType<ModelDegradationModel> {}
+// eslint-disable-next-line @typescript-eslint/no-unsafe-declaration-merging
+export class ModelDegradationResource extends BaseResource<ModelDegradationModel> {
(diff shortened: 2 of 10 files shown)


frankaloia (dismissed):
LGTM

front-api/routes/poke/kill.ts
```diff
@@ -40,4 +60,59 @@ app.post(
...
 
+/**
+ * Replaces the set of degraded models with the one submitted. Degraded-model
+ * kill switches are ordinary `kill_switches` rows, this just diffs them in one
+ * shot.
+ *
+ * @ignoreswagger
+ */
+app.post(
+  "/models",
```
frankaloia: Not blocking. Only concern on this one, is the terminology unclear of degraded vs killed (i.e. the api is still under the kill path)? Do we see us having a concept of a killed models in addition to the degraded model in the future as well?

frankaloia:
I'm actually having second thoughts on reusing the kill_switch table based on our recent changes. kill_switches is more of a global flag registry. Do we want a proper table that usesa. single enum column to denote model status healthy | degraded | probing ? We might also want to build it in a way where it stores status history for downstream analytics. Open to discussing tomorrow

front/lib/poke/degradable_models.ts
```diff
@@ -0,0 +1,18 @@
...
+import { SUPPORTED_MODEL_CONFIGS } from "@app/types/assistant/models/models";
+import type { ModelConfigurationType } from "@app/types/assistant/models/types";
+
+export const DEGRADABLE_MODEL_CONFIGS: ModelConfigurationType[] =
+  SUPPORTED_MODEL_CONFIGS.filter(
+    (m) => !isModelStreamId(m.modelId) && m.providerId !== "noop"
+  );
+
+// Widened to `string` for dynamic models coming from GCS:
+// callers check ids coming off the wire.
```
frankaloia: nit: comment slightly unclear

front/lib/api/poke/degraded_models.ts
```diff
@@ -0,0 +1,92 @@
...
+});
+
+/**
+ * The endpoints an operator may flag as degraded: the stream endpoints of the
+ * router, de-duplicated across regions since degradation is region-wide and a
+ * host covers both its `eu` and `global` endpoints.
+ *
+ * Built from the endpoint registry rather than from `SUPPORTED_MODEL_CONFIGS`
+ * so an operator can only degrade something that actually serves traffic.
+ */
```
Nils-Fedrigo:
```suggestion
```

front/lib/api/poke/degraded_models.ts
```diff
@@ -0,0 +1,92 @@
...
+  }
+
+  return [...byKey.values()];
+}
+
+/**
+ * Matches the submitted endpoints against the catalog, de-duplicated. Returns
+ * the catalog entries themselves, so what reaches the database is always a
+ * known endpoint with its narrow types, never an arbitrary string off the wire.
+ */
```
Nils-Fedrigo:
```suggestion
```

front/lib/resources/storage/models/model_degradations.ts
```diff
@@ -0,0 +1,61 @@
...
+
+/**
+ * A model endpoint an operator flagged as degraded, i.e. having an ongoing
+ * incident on the provider side. A row is an endpoint, not a model: the same
+ * model is served from more than one host (e.g. Claude via `anthropic` and via
+ * `agent-platform`), and an incident usually hits one of them.
+ *
+ * Global table, and each region has its own database. `region` is deliberately
+ * not part of the key -- a host covers both its `eu` and `global` endpoints.
+ */
```
Nils-Fedrigo:
```suggestion
```

front/lib/resources/storage/models/model_degradations.ts
```diff
@@ -0,0 +1,61 @@
...
+export class ModelDegradationModel extends BaseModel<ModelDegradationModel> {
+  declare createdAt: CreationOptional<Date>;
+  declare updatedAt: CreationOptional<Date>;
+
+  declare modelId: ModelIdType;
+  declare providerId: ModelProviderIdType;
+  // `host` is only ever a column here: it is a reserved key in our log and
+  // metric infra, so it must be renamed (`modelHost`) before being logged.
```
Nils-Fedrigo:
```suggestion
```

front/lib/resources/model_degradation_resource.ts
```diff
@@ -0,0 +1,63 @@
...
+} from "@app/types/assistant/models/types";
+
+export type DegradedModelEndpointType = {
+  modelId: ModelIdType;
+  providerId: ModelProviderIdType;
+  host: Host;
+};
+
+// Identity of a degraded endpoint, matching the table's unique index. Kept
+// loose so entries coming off the wire can be keyed before they are validated.
```
Nils-Fedrigo:
```suggestion
```

front/lib/resources/model_degradation_resource.ts
```diff
@@ -0,0 +1,63 @@
...
+ * Degraded endpoints are a small global set that callers only ever read whole or
+ * replace whole -- never a row identity -- so this resource exposes statics
+ * returning plain {@link DegradedModelEndpointType} objects and never
+ * instantiates itself.
+ */
+export class ModelDegradationResource {
+  /**
+   * Every degraded endpoint. The table holds at most one row per endpoint of the
+   * catalog (~70), and only during an incident, so this is a full table read.
+   */
```
Nils-Fedrigo:
```suggestion
  // At most 70 rows
```

front/lib/resources/model_degradations.ts
```diff
@@ -0,0 +1,89 @@
...
+  host: string;
+}): string {
+  return `${modelId}|${providerId}|${host}`;
+}
+
+/**
+ * Callers only ever read the degraded set whole or toggle one endpoint, never a
+ * row identity, so this is statics over {@link ModelDegradationModel} returning
+ * plain objects. It is deliberately not a `Resource`: it never instantiates
+ * itself, and the suffix is for `BaseResource` subclasses.
```
Nils-Fedrigo:
```suggestion
 * row identity, so this is statics over {@link ModelDegradationModel}
```

front/lib/resources/storage/models/model_degradations.ts
```diff
@@ -0,0 +1,50 @@
...
+    },
+    modelId: {
+      type: DataTypes.STRING,
+      allowNull: false,
+    },
+    providerId: {
+      type: DataTypes.STRING,
+      allowNull: false,
+    },
+    host: {
```
frankaloia, line 39: "host" can be an overloaded term. Ok keeping it, but would at least add a comment in the code

front/lib/resources/model_degradations.ts
```diff
@@ -0,0 +1,87 @@
...
+  host: string;
+}): string {
+  return `${modelId}|${providerId}|${host}`;
+}
+
+/**
+ * Callers only ever read the degraded set whole or toggle one endpoint, never a
+ * row identity, so this is statics over {@link ModelDegradationModel}
+ */
+export class ModelDegradations {
```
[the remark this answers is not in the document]
frankaloia: @Nils-Fedrigo This is valid, we do want this

front/lib/resources/storage/models/model_degradations.ts
```diff
@@ -0,0 +1,50 @@
...
+import { frontSequelize } from "@app/lib/resources/storage";
+import { DataTypes } from "@app/lib/resources/storage/data_types";
+import { BaseModel } from "@app/lib/resources/storage/wrappers/base";
+import type {
+  ModelIdType,
+  ModelProviderIdType,
+} from "@app/types/assistant/models/types";
+import type { CreationOptional } from "sequelize";
+
+export class ModelDegradationModel extends BaseModel<ModelDegradationModel> {
```
frankaloia, line 11: Given that we are also operating this via https://github.com/dust-tt/dust/pull/31470/changes#diff-2887abf52a05ac059e260b81e5b5fa91a60e15ca3004f19252302c8e179d6661

I'm wondering if we want some enum to denote automatic vs manual? For example, if someone manually marks as degraded, do we want an automatic process to override that?

front/lib/resources/storage/models/model_degradations.ts
frankaloia: I retract this. Realizing this is the terminology we already use, so lets keep it consistent

front/lib/resources/storage/models/model_degradations.ts
Nils-Fedrigo: for now let's keep it simple to merge this stack and add an enum later

front/lib/resources/model_degradation_resource.ts
```diff
@@ -0,0 +1,89 @@
...
+      providerId,
+      host,
+    }));
+  }
+
+  /**
+   * Clears the rows of every endpoint the request names, then re-adds the ones
+   * it wants degraded. Endpoints the request does not name are left alone, so
+   * two operators working the same incident do not clobber each other.
+   */
```
Nils-Fedrigo:
```suggestion
```

PR #31137, opened 2026-08-25
feat(models): skip degraded models in auto stream resolution

Nils-Fedrigo:
## Description

Second of the degraded-model stack (on top of #31136): the auto streams honour the degraded set, so a stream routes around a model that is having an incident.

**This is the PR whose scope changed.** The previous version pulled a degraded model out of `isModelEnabled`, which took it out of *every* path at once — the picker, the agent builder, the fallback lists and the workspace model catalog. That is more than we want:

- A **stream** (`auto` / `auto_fast` / `auto_complex`) picks a model on the user's behalf, so routing around a provider incident is ours to do.
- A **definitive pick** — an agent configured on a concrete model, or a user overriding the model from the picker — is the user's. Answering as a different model than the one they explicitly asked for is worse than running through the incident.

So degradation now comes up in the stream resolution flow and nowhere else:

- `front/lib/api/assistant/degraded_models.ts`: reads `model_degradations` into an in-process set, refreshed at most once every 60s, lazily off the read path. On a refresh failure the last known set is kept — a DB blip must not silently bring a degraded model back into the streams.
  Degradation is stored per endpoint (`modelId + providerId + host`) but **collapsed to model ids here**: a model is reported degraded as soon as one of its endpoints is. That is the conservative reading and keeps this PR's behaviour model-level; per-host failover — keep the model, skip only the degraded endpoint — is left to a later PR and marked `TODO(detect_outage)`.
- `resolveStreamModel(models, streamId, degradedModelIds)` takes the degraded set as an argument. The candidate-pool walk skips degraded models, and so does the last-resort preferred-large-model fallback, which is just as automatic a pick as the walk itself.
- Its three callers are all stream resolutions (`resolveModel`, `getStreamResolutions`, the premium fair-use downgrade target), so passing the set explicitly keeps the reach of degradation visible at each call site — and keeps the function pure and testable.

`front/lib/assistant.ts`, `EnabledModelConfigurationType`, the model picker and the agent builder are **untouched**: no `getModelEnablementStatus`, no `filterAvailableModels`, no `isKilled` flag on the wire. The net diff of the whole stack is 13 files.

One nice consequence: the picker tiers reflect degradation with no client code at all, because `getStreamResolutions` runs server-side — Basic/Standard/Premium already display the model they would actually run.

### Reviewing

Two commits, kept split on purpose:

1. `feat(models): take killed models out of rotation` — the original, as previously reviewed. Review the branch head rather than this commit in isolation.
2. `refactor(models): only skip degraded models in stream resolution` — the scope narrowing. **This is the commit to review.**
3. `refactor(models): read degraded models from the model_degradations table` — swaps the source from the kill switches to the `model_degradations` table added in #31136. Same signature, same 60s cache, same keep-the-last-known-set-on-error behaviour, so `enabled_models.ts`, `resolve_model.ts`, `premium_model_limit.ts` and their tests are untouched.

---

Stack:
1. #31136 — models: model degradation table + accessors
2. #31137 — models: skip degraded models in auto stream resolution
3. #31145 — poke: the degraded models API + marking endpoints degraded from the Kill page

Dropped from the original 5-PR stack, since a definitive model pick now runs as usual: ~~#31138~~ (agent: fail explicitly when the requested model is killed), ~~#31139~~ (model picker: show killed models as temporarily unavailable).

## Tests

`npx tsgo --noEmit` clean on `front` and `front-api`. 118 tests pass across `enabled_models`, `assistant/models`, `premium_model_limit`, `lib/assistant`, `modelPickerUtils`, `buildPickModelSlashCommandItems`, plus `batch_update_model` in `front-api`.

Two new cases in `enabled_models.test.ts`:
- a degraded first candidate is skipped and the stream takes the next one in its pool (Luna@high → Sonnet 4.6@medium on `auto`);
- a degraded model is kept out of the last-resort fallback, not just out of the pool walk.

Manual: mark a model endpoint degraded from Poke, wait up to 60s, confirm the affected tier resolves to the next candidate while an agent pinned to that model still answers on it.

## Risk

Lower than the previous version of this PR, and that is the point. The shared enablement path (`isModelEnabled` / `filterEnabledModels`) is no longer touched, so the picker, the agent builder and agent runs behave exactly as today when nothing is degraded — and when something is degraded, only stream resolution changes. Rollback is a plain revert.

Two things worth a reviewer's eye:

- The last-resort fallback now filters degraded models too. If *every* candidate is degraded and nothing else is selectable, `pickPreferredLargeModel` falls through to its own hardcoded default (Sonnet 4.6) — same behaviour as today when nothing is selectable, but worth confirming that is the intended floor.
- `premium_model_limit`'s downgrade target also resolves through streams, so a degraded model can no longer be picked as a downgrade target. That is deliberate (it should match what the stream would actually run) but it is a second, less obvious caller.


front/lib/api/assistant/killed_models.ts
```diff
@@ -0,0 +1,42 @@
+import { killedModelIdsFromKillSwitches } from "@app/lib/poke/types";
+import { KillSwitchResource } from "@app/lib/resources/kill_switch_resource";
+import logger from "@app/logger/logger";
+import { normalizeError } from "@app/types/shared/utils/error_utils";
+
+// How long a pod may serve a stale set of killed models.
+const REFRESH_INTERVAL_MS = 60 * 1000;
+
+let cachedKilledModelIds: ReadonlySet<string> = new Set();
+let lastRefreshStartedAt = 0;
```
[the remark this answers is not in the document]
frankaloia: Valid

front/lib/api/assistant/degraded_models.ts
```diff
@@ -0,0 +1,49 @@
...
+/**
+ * The models an operator marked as degraded, i.e. having an ongoing incident on
+ * the provider side.
+ *
+ * Only the stream resolution flow reads this: an `auto` stream skips a degraded
+ * candidate and moves to the next one in its pool. A definitive pick -- an
+ * agent configured on a concrete model, or a user overriding the model from the
+ * picker -- runs as usual, degraded or not, because we must never answer as a
+ * model other than the one that was explicitly asked for.
+ */
```
Nils-Fedrigo, line 27:
```suggestion
/**
 * The models marked as degraded, i.e. having an ongoing incident on
 * the provider side.
 *
 * Only the stream resolution flow reads this: an `auto` stream skips it
 */
```

front/lib/model_tiers/enabled_models.ts
```diff
@@ -150,15 +151,24 @@ export interface StreamResolutionType {
 }
 
 // Walks a stream's ordered candidate pool and picks the first one available
-// or a fallback large model
+// or a fallback large model.
+//
+// Degraded models are skipped here and only here: a stream picks a model on the
+// user's behalf, so routing around an ongoing provider incident is ours to do.
+// A definitive pick -- an agent configured on a concrete model, or a user
+// overriding the model from the picker -- is left alone and runs as usual.
```
Nils-Fedrigo, line 163:
```suggestion
// or a fallback large model. Degraded models are skipped.
```

front/lib/model_tiers/enabled_models.ts
```diff
@@ -172,9 +182,9 @@ export function resolveStreamModel(
     }
   }
 
-  const fallback = pickPreferredLargeModel(
-    models.filter((m) => m.isSelectable)
-  );
+  // Still off the degraded ones: the last-resort fallback is as automatic a pick
+  // as the pool walk itself.
```
Nils-Fedrigo, line 190:
```suggestion
```

frankaloia (approved):
LGTM pending my comment on the table from PR1

front/lib/api/assistant/degraded_models.ts
```diff
@@ -0,0 +1,54 @@
...
+ * router filters degraded endpoints itself, a model served from another healthy
+ * host should stay in the streams and only the degraded endpoint be skipped.
+ */
+export function getDegradedModelIds(): ReadonlySet<string> {
+  const now = Date.now();
+  if (!refreshing && now - lastRefreshStartedAtMs > REFRESH_INTERVAL_MS) {
+    refreshing = true;
+    lastRefreshStartedAtMs = now;
+
+    void ModelDegradationResource.listDegradedEndpoints()
```
flvndvd, line 34: OOC this slightly goes against the resource owns it's lifecycle, caching should be resource-owned and call site should not worry about cache or not.

PR #31138, opened 2026-08-25
feat(agent): fail explicitly when the requested model is killed

Nils-Fedrigo:
## Description

Third of the per-model kill switch stack (on top of #31137): what happens when a conversation is pinned to a model that has just been killed.

The choice here is to **fail loudly on the model that was actually requested** rather than quietly answer as a different one:

- `resolve_model.ts`: a killed requested model is kept instead of running the fallback list. Falling through on a kill would answer as a model neither the agent nor the user picked, without saying so.
- `run_model_and_create_actions_wrapper.ts`: the agent loop stops before running and publishes a `model_disabled` guardrail error with a "Model unavailable" title and a message naming the model. `publishAgentLoopGuardrailExceededError` gains an optional `errorMessage` so it is not tied to the resource-cap wording.
- `front-api` public model selection: a killed model is let through validation so the message is created and the error surfaces on the agent message, instead of the send being rejected with an opaque authorization error.

---

Stack:
1. #31136 — poke: kill switch types + write API
2. #31137 — models: take killed models out of rotation
3. #31138 — agent: fail explicitly when the requested model is killed
4. #31139 — model picker: show killed models as temporarily unavailable
5. #31145 — poke: kill individual models from the Kill page

## Tests

Manual: kill a model from Poke, then send a message on an agent pinned to it — the user message is created and the agent message carries the "Model unavailable" error. Same check for a model picked explicitly from the picker.

## Risk

Medium — this deliberately turns a silent fallback into a user-visible error, so killing a model during an incident produces failed agent messages for anyone pinned to it rather than answers from another model. That is the intent (an operator killing a model wants the blast radius visible), but it is the behavioural decision to review in this stack.

### DIFF
--- a/front/temporal/agent_loop/activities/run_model_and_create_actions_wrapper.ts
+++ b/front/temporal/agent_loop/activities/run_model_and_create_actions_wrapper.ts
@@ -1,4 +1,5 @@
 import { isToolExecutionStatusFinal } from "@app/lib/actions/statuses";
+import { isModelKilled } from "@app/lib/api/assistant/killed_models";
 import { getRetryPolicyFromToolConfiguration } from "@app/lib/api/mcp";
 import type { Authenticator, AuthenticatorType } from "@app/lib/auth";
 import { getFeatureFlags } from "@app/lib/auth";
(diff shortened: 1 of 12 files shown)


front/temporal/agent_loop/activities/run_model_and_create_actions_wrapper.ts
```diff
@@ -144,6 +145,38 @@ async function _runModelAndCreateActionsActivity({
   }
 
   const { auth, ...runAgentData } = runAgentDataRes.value;
+
+  const runModelConfig = runAgentData.modelInfo.endpoint.modelConfig;
+  if (isModelKilled(runModelConfig.modelId)) {
```
frankaloia, line 150: I think we need to decide on our definition of "killed" here. If we are only "killing" a model due to incident fallover, do we actually want to explicitly deny all calls to it? Feels like we could fail open and let the customer call it if they are explicitly asking for it, as opposed to guaranteeing failure. We could have an explicit message be something like "hey, we know this is failing, you should request another one since you aren't using auto"

If we have this nuance, then maybe we need another definition then "killed", since we do use killed for not being able to use the model at all today

Nils-Fedrigo:
Closing: dropped from the stack.

The product rule changed — degradation (formerly "kill") is now a **stream-resolution** concern only. An `auto` stream routes around a degraded model, but a definitive pick (an agent configured on a concrete model, or a user overriding the model from the picker) runs as usual, because answering as a model other than the one that was explicitly asked for is worse than running through the incident.

Both halves of this PR existed only to make a *pinned* degraded model fail loudly: `resolveModel` held the requested model back from the fallback list so the run would fail on it, and the agent-loop activity turned that into a "model unavailable" error on the answer. Under the new rule the pinned model just runs, so there is nothing left here.

Superseded by #31137, which now carries the whole enforcement surface.

PR #31139, opened 2026-08-25
feat(model-picker): show killed models as temporarily unavailable

Nils-Fedrigo:
## Description

Fourth of the per-model kill switch stack (on top of #31138): instead of silently vanishing from the picker, a killed model stays listed and reads as temporarily unavailable.

- `getModelLockReason` gains a `"killed"` reason, which outranks the premium/tier reasons — a killed model cannot run whatever the workspace plan says — plus its tooltip.
- The picker keeps killed models in `allModels` (`isSelectable || isKilled`) and renders them disabled with an `AlertCircle` instead of the padlock used for plan/tier locks.
- The agent builder model submenu shows "Temporarily down. Choose another model." on a selected killed model, rather than the "Not enabled for you" wording meant for plan gating.
- Most of the remaining diff is mechanical: the picker components move from `ModelConfigurationType` to `EnabledModelConfigurationType` so they can read `isKilled`. Skimmable.

Note: the `/model` slash command in the editor still filters on `isSelectable` (`buildPickModelSlashCommandItems.ts:106`), so it keeps hiding killed models rather than showing them disabled. Happy to align it here if we'd rather have one behaviour across both surfaces.

---

Stack:
1. #31136 — poke: kill switch types + write API
2. #31137 — models: take killed models out of rotation
3. #31138 — agent: fail explicitly when the requested model is killed
4. #31139 — model picker: show killed models as temporarily unavailable
5. #31145 — poke: kill individual models from the Kill page

## Tests

New `getModelLockReason` unit tests: a killed model, a killed premium model (kill wins over the plan reason), and a live model. Manual: kill a model from Poke and check the picker row is disabled with the outage tooltip and that clicking it is a no-op.

## Risk

Low, UI only — with no model killed, `isKilled` is always false and every surface behaves as before.

### DIFF
--- a/front/components/agent_builder/instructions/ModelSelectionSubmenu.tsx
+++ b/front/components/agent_builder/instructions/ModelSelectionSubmenu.tsx
@@ -40,6 +40,21 @@ interface ModelSelectionSubmenuProps {
 const NOT_SELECTABLE_MODEL_DESCRIPTION =
   "Not enabled for you. Choose another model to save.";
 
+const KILLED_MODEL_DESCRIPTION = "Temporarily down. Choose another model.";
+
+function getModelDescription(
+  modelConfig: EnabledModelConfigurationType,
+  isSelected: boolean
+): string | undefined {
+  if (modelConfig.isSelectable || !isSelected) {
+    return modelConfig.shortDescription;
+  }
+
+  return modelConfig.isKilled
+    ? KILLED_MODEL_DESCRIPTION
+    : NOT_SELECTABLE_MODEL_DESCRIPTION;
+}
+
 interface ModelRadioItemProps {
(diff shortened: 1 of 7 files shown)


Nils-Fedrigo:
Closing: dropped from the stack.

The product rule changed — degradation (formerly "kill") is now a **stream-resolution** concern only. A user who overrides the model from the picker gets that model, degraded or not, so there is nothing to lock and nothing to warn about: a degraded row here would be a disabled row the user is in fact allowed to pick.

The tiers still reflect degradation with no picker code at all, because they are resolved server-side — `getStreamResolutions` skips degraded candidates, so Basic/Standard/Premium already display the model they would actually run.

Superseded by #31137.

PR #31141, opened 2026-08-25
Introduce agent loop context provider

flvndvd:
## Description

This PR introduces an explicit context provider boundary between the agent loop and conversation rendering.

This prepares the model activity for stateful checkpoint hydration while preserving the current rendering behavior.

Today, agent-loop data loading and model-context hydration are coupled. Fetching agent-loop metadata can therefore hydrate and render conversation history even when the caller only needs the current messages or conversation metadata.

Stateful context windows need a clear boundary where we can later choose between:
- Loading a valid checkpoint plus the current-step delta
- Falling back to the existing full bounded rendering path

In this PR we:
- introduce `AgentLoopContextProvider` with the current rendering logic, flagged as "full" which will be the default one if no checkpoint context can be found.
- separate metadata-only `AgentLoopRuntimeData` from full `AgentLoopExecutionData`.
- make the generic agent-loop loader fetch only:
    - Conversation metadata
    - The current user message
    - The current agent message
- (sadly) keeps `runTool` explicitly on full conversation hydration because some tools inspect conversation history. There are 5 tools relying on conversation content, will clean them up in follow up PRs.
- Moves conversation slicing and missing-action-catcher discovery behind the provider

## Behavior

No model-visible, tool-visible, or client-visible behavior change is intended.

The model activity still uses the same bounded full conversation, step slicing, and renderer. The provider currently wraps that existing behavior.

The immediate improvement applies to auxiliary activities that no longer need to hydrate conversation history through the generic agent-loop loader.

### DIFF
--- a/front/lib/api/assistant/conversation_rendering/conversation_window_core.ts
+++ b/front/lib/api/assistant/conversation_rendering/conversation_window_core.ts
@@ -18,7 +18,7 @@ import { getLlmCredentials } from "@app/lib/api/provider_credentials";
 import type { Authenticator } from "@app/lib/auth";
 import { tokenCountForTexts } from "@app/lib/tokenization";
 import logger from "@app/logger/logger";
-import type { AgentLoopExecutionData } from "@app/types/assistant/agent_run";
+import type { AgentConfigurationWithoutModelType } from "@app/types/assistant/agent";
 import type { ConversationType } from "@app/types/assistant/conversation";
 import type {
   ModelConversationTypeMultiActions,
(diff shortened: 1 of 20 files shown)


front/types/assistant/agent_run.ts
```diff
@@ -183,12 +194,92 @@ export async function getAgentLoopData(
 export async function getAgentLoopDataWithAuth(
   auth: Authenticator,
   agentLoopArgs: AgentLoopArgs
-): Promise<
-  Result<
-    AgentLoopExecutionData & { auth: Authenticator },
-    AgentLoopDataError | Error
-  >
-> {
+): Promise<Result<AgentLoopDataWithAuth, AgentLoopDataError | Error>> {
```
flvndvd: Highly inspired by `getFullAgentLoopDataWithAuth`.

front/lib/api/assistant/conversation_rendering/conversation_window_core.ts
```diff
@@ -51,6 +51,29 @@ export const PREVIOUS_INTERACTIONS_TO_PRESERVE = 3;
...
 export const PRUNING_TARGET_CONTEXT_UTILIZATION = 0.6;
 
+export type ConversationRenderingInput = {
+  leadingMessages?: ModelMessageTypeMultiActionsWithoutContentFragment[];
+  model: ModelConfigurationType;
+  prompt: string;
+  tools: string;
+  allowedTokenCount: number;
+  excludeActions?: boolean;
+  excludeImages?: boolean;
```
id13, line 61: wouldn't it be easier to pass a list of what to include or exclude ? seems like the beginning of a bad boolean-flag dance

front/lib/api/assistant/conversation_rendering/conversation_window_core.ts
```diff
@@ -51,6 +51,29 @@ export const PREVIOUS_INTERACTIONS_TO_PRESERVE = 3;
...
+  onMissingAction?: "inject-placeholder" | "skip";
+  enablePreviousInteractionsPruning?: boolean;
```
id13, line 63: the term interactions is loosely defined in the codebase today (in fine it's messages), can we seize the opportunity to harden this or rename ?

id13 (approved):

front/lib/api/assistant/conversation_rendering/conversation_window_core.ts
flvndvd: 100 % fair, it was already here, but will make a clean up pass once all PRs have landed.

adrsimon (approved):

PR #31142, opened 2026-08-25
Admin governance: Remove write permissions based on role for regular open groups

fabiencelier:
## Description

 - remove write permission
 - add all managers and admins ad member of open regular space

## Risk

changes the behavior of something sensitive
does not add new access but we need to communicate that removing the manager role won't remove the write permission

## Deploy Plan

Run migration
```
npx tsx migrations/20260825_backfill_open_space_admin_manager_members.ts --execute
```

deploy front

### DIFF
--- a/front/migrations/20260825_backfill_open_space_admin_manager_members.ts
+++ b/front/migrations/20260825_backfill_open_space_admin_manager_members.ts
@@ -2,8 +2,6 @@ import { Authenticator } from "@app/lib/auth";
 import { MembershipResource } from "@app/lib/resources/membership_resource";
 import { SpaceResource } from "@app/lib/resources/space_resource";
 import { UserResource } from "@app/lib/resources/user_resource";
-import { WorkspaceResource } from "@app/lib/resources/workspace_resource";
-import { renderLightWorkspaceType } from "@app/lib/workspace";
 import type { Logger } from "@app/logger/logger";
 import { makeScript } from "@app/scripts/helpers";
 import { runOnAllWorkspaces } from "@app/scripts/workspace_helpers";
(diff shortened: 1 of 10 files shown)


matteotrab (approved):

front/migrations/20260825_backfill_open_space_admin_manager_members.ts
```diff
@@ -0,0 +1,155 @@
...
+}
+
+makeScript(
+  {
+    wId: { type: "string", required: false },
+  },
+  async ({ wId, execute }, logger) => {
+    logger.info("Starting open regular space admin/manager backfill");
+
+    if (wId) {
```
tdraier: nit: runOnAllWorkspaces already allows `wId` as parameter

front/migrations/20260825_backfill_open_space_admin_manager_members.ts
```diff
@@ -0,0 +1,155 @@
...
+async function backfillWorkspaceOpenSpaceMembers(
+  execute: boolean,
+  logger: Logger,
+  workspace: LightWorkspaceType
+): Promise<void> {
+  const auth = await Authenticator.internalAdminForWorkspace(workspace.sId);
+
+  const spaces = await SpaceResource.listWorkspaceSpaces(auth);
+  const targets = spaces.filter(
+    (space) => space.isRegularAndOpen() && space.managementMode === "manual"
```
tdraier, line 29: what happen to regular/open which space.managementMode !== "manual" ? (is it possible ?)

tdraier (approved):

front/migrations/20260825_backfill_open_space_admin_manager_members.ts
fabiencelier: Yes it's possible
For them I assume someone went there to manually edit the group so I don't inject the admins/managers and let them handle
Should not be a lot of group if we merge this soon as people don't know it is possible.

front/migrations/20260825_backfill_open_space_admin_manager_members.ts
fabiencelier: Arf, this has been done for a few scripts, fixing them

PR #31145, opened 2026-08-25
feat(poke): mark model endpoints degraded from the Kill page

Nils-Fedrigo:
## Description

Last of the degraded-model stack (on top of #31470): the Poke UI to actually mark endpoints degraded, landing once enforcement is in place. The API it calls was split out into #31470.

- **`DegradedModelsDialog.tsx`**: a dialog listing the degradable endpoints grouped provider → model → host. Degrade/Restore rather than Kill/Revive. A model served from a single host is one checkbox, as before; one served from several gets a per-host checkbox under a tri-state parent, so an incident on `anthropic` can be flagged without touching `agent-platform`. Save submits **only the endpoints the operator toggled** — the delta the confirm summary already computes — so a colleague degrading a different model from their own tab is not restored behind their back. The catalog comes from the API rather than a client-side constant: the endpoint registries pull in provider SDKs and cannot be imported into the browser bundle.
- **`usePokeDegradedModels` / `useUpdatePokeDegradedModels`** in `poke/swr/degraded_models.ts`; `poke/swr/kill.ts` goes back to just `usePokeKillSwitches`.
- **Kill page**: a "Degraded Models" section showing which endpoints are currently degraded, and the note that a change takes up to 60s to apply on each pod (stream resolution reads an in-process cache).
- `TOGGLABLE_KILL_SWITCH_TYPES` is introduced here rather than further down the stack, since which switches the page offers is a UI concern.

The copy is explicit about what degradation does and does not do, since that is the part that changed: the Basic/Standard/Premium streams skip a model as soon as one of its endpoints is degraded and pick the next candidate in their pool, while an agent configured on that model — or a user who picks it from the model picker — keeps running on it. A definitive pick is never silently swapped.

## Tests

`npx tsgo --noEmit` clean on `front` and `front-api`. The API tests live in #31470.

Manual, on the Poke Kill page: mark a couple of endpoints degraded — including a single host of a multi-host model — reload, check they round-trip and that the "Currently degraded" list matches. Then end-to-end against #31137 — within 60s the affected tier resolves to the next candidate in its pool, while an agent pinned to a degraded model still answers on it (that last part is the behaviour change worth verifying, since the previous version of the stack made it error instead).

Worth doing once with two browser tabs, since it is what the per-endpoint API buys: degrade a model in one tab and a different model in the other, save both, and check neither restores the other.

## Risk

Low on its own (Poke-only surface), but this is the PR that makes the whole feature operable, so it should merge last. Rollback is a revert; degraded rows can also be cleared from the dialog itself.

Note degradation is per-region: the table lives in each region's own database, so taking a model out of the streams everywhere means doing it from both regions. The dialog says so.

The one non-Poke risk is the deprecated-toggle removal flagged above.

## Deploy Plan

Merge in stack order (#31136 → #31137 → #31470 → this one). No migration of its own, so this is a plain code deploy.

1. Before merging, either verify the three deprecated provider-level switches are disabled in production, or split their removal out into its own PR.
2. Deploy `front`.
3. Smoke-test on the Poke Kill page: the Degraded Models section lists the catalog, degrading one endpoint round-trips after a reload, and restoring it clears the row.

No rollout ordering concern against #31137: an empty `model_degradations` means nothing is skipped, which is the current behaviour.

### DIFF
--- a/front/components/poke/pages/KillPage.tsx
+++ b/front/components/poke/pages/KillPage.tsx
@@ -82,11 +82,11 @@ const KILL_SWITCH_DEFINITIONS: Record<KillSwitchType, KillSwitchDefinition> = {
   },
 };
 
-// Kills a whole provider at once, which the per-endpoint degraded models
-// switches below do more precisely.
+// Kills a whole provider at once, with a message that is not explicit.
 const LEGACY_KILL_SWITCH_NOTE =
-  "Superseded by the degraded models section: prefer flagging the affected " +
-  "endpoints there, and reach for this only to take a provider out entirely.";
+  "Replaced by the degraded models section. It was not explicit to the users that " +
+  " providers were down. This also blocked all calls to affected provider, not just the " +
+  "Auto models.";
 
 const PANEL_HEADING_CLASSES =
   "flex items-center gap-2.5 text-2xl font-semibold tracking-tight text-foreground";
--- a/front/components/poke/pages/KillPage.tsx
+++ b/front/components/poke/pages/KillPage.tsx
@@ -1,8 +1,12 @@
+import { ModelKillSwitchesDialog } from "@app/components/poke/ModelKillSwitchesDialog";
 import { cn } from "@app/components/poke/shadcn/lib/utils";
 import { useSendNotification } from "@app/hooks/useNotification";
 import { clientFetch } from "@app/lib/egress/client";
 import type { KillSwitchType } from "@app/lib/poke/types";
-import { KILL_SWITCH_TYPES } from "@app/lib/poke/types";
+import {
+  killedModelIdsFromKillSwitches,
+  TOGGLABLE_KILL_SWITCH_TYPES,
+} from "@app/lib/poke/types";
 import { usePokePageMetadata } from "@app/poke/swr/currentPage";
 import { usePokeKillSwitches } from "@app/poke/swr/kill";
 import {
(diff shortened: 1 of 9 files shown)


front/poke/swr/degraded_models.ts
```diff
@@ -0,0 +1,84 @@
...
+    degradedModelsFetcher
+  );
+
+  return {
+    degradableEndpoints:
+      data?.degradableEndpoints ?? emptyArray<DegradableModelEndpointType>(),
+    degradedEndpoints:
+      data?.degradedEndpoints ?? emptyArray<DegradedModelEndpointType>(),
+    isDegradedModelsLoading: !error && !data,
+    isDegradedModelsError: error,
```
frankaloia, line 26: Should we actually be using this error/loading state? Right now it is ignored

frankaloia (commented):
Rest looks good on this one after we make the changes defined in the other PRs

front/poke/swr/degraded_models.ts
Nils-Fedrigo: I think we are cool, it's poke side

front/components/poke/pages/KillPage.tsx
```diff
@@ -85,6 +82,12 @@ const KILL_SWITCH_DEFINITIONS: Record<KillSwitchType, KillSwitchDefinition> = {
   },
 };
 
+// Kills a whole provider at once, which the per-endpoint degraded models
+// switches below do more precisely.
+const LEGACY_KILL_SWITCH_NOTE =
+  "Superseded by the degraded models section: prefer flagging the affected " +
+  "endpoints there, and reach for this only to take a provider out entirely.";
+
```
Nils-Fedrigo:
```suggestion
// Kills a whole provider at once, with a message that is not explicit.
const LEGACY_KILL_SWITCH_NOTE =
  "Replaced by the degraded models section. It was not explicit to the users that " +
  " providers were down. This also blocked all calls to affected provider, not just the " +
  "Auto models.";

```

frankaloia (approved):
LGTM pending comment from previous PR

PR #31146, opened 2026-08-25
[front] fix: remove user requirement on toolsets enable

aubin-tchoi:
## Description

Follow up on https://github.com/dust-tt/dust/pull/31132

This PR removes the requirement of having a user in the toolsets enable tool.
It seems that this requirement is a leftover from relying on the public API, but keep me honest.

## Risk

- Low.

## Deploy Plan

- Deploy front.


aubin-tchoi:
PMRR to unblock myself but open to discussion!

PR #31147, opened 2026-08-25
[front] enh: track user menu item clicks

aubin-tchoi:
## Description

User menu clicks are currently not observable in product analytics. This PR emits `navigation:user_menu_item:click` with an `item` property for each submenu + action.

The workspace picker is not tracked, we don't need tracking there.

## Tests

- Tested locally.

## Risk

- Low.

## Deploy Plan

- Deploy front.

### DIFF
--- a/front/components/UserMenu.tsx
+++ b/front/components/UserMenu.tsx
@@ -78,15 +78,12 @@ interface UserMenuProps {
   creditUsageState?: CreditUsageState | null;
 }
 
-function withUserMenuItemTracking(item: string, handler?: () => unknown) {
-  return () => {
-    trackEvent({
-      area: TRACKING_AREAS.NAVIGATION,
-      object: "user_menu_item",
-      extra: { item },
-    });
-    void handler?.();
-  };
+function trackUserMenuEvent(item: string) {
+  trackEvent({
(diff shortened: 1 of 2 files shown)


front/components/UserMenu.tsx
```diff
@@ -475,56 +514,79 @@ export function UserMenu({
...
-                        onClick={() => {
-                          const regexp = new RegExp(
-                            `/w/([^/]+)/conversation/([^/]+)`
-                          );
-                          const match = window.location.href.match(regexp);
-                          if (match) {
-                            window.open(
-                              `/poke/${match[1]}/conversation/${match[2]}`,
-                              "_blank"
+                        onClick={withUserMenuItemTracking(
```
PopDaph: Are you sure we want to track such debug dev only actions? I would remove all of those unless we aim at doing something about this data 🙏🏻

PopDaph (approved):
👍🏻 
Let's remove the debug tool tracking no? 🙏🏻

front/components/UserMenu.tsx
aubin-tchoi: Oops mb, missed this!

PR #31149, opened 2026-08-25
[front] Classify LLM HTTP errors in one place and drop redundant providerId from logs

frankaloia:
## Description

Follow-up to the LLM attempt Datadog instrumentation. Two log/classification tweaks from looking at production traces:

- Drop `providerId` from structured attempt logs. It duplicated `inferenceProvider` in the common case and is already on metrics as `provider_id` (where it can differ, e.g. Claude on Vertex).
- Collapse the five per-provider HTTP status switches into `classifyHttpStatus` / `buildHttpStatusErrorEvent`. Named statuses keep the same type, source, and message. Generally classify 400 == dust, 500 == provider.

## Tests

Unit tests

## Risk

Low. Streaming happy-path statuses are unchanged. Classification changes are leftover 4xx, some 503s, and batch item errors — telemetry and retryability for those paths only. Safe to rollback by reverting the commit.

## Deploy Plan
1. Deploy front

### DIFF
--- a/front/lib/api/llm/llm.ts
+++ b/front/lib/api/llm/llm.ts
@@ -369,7 +369,6 @@ export abstract class LLM<
         const durationMs = Date.now() - startTime;
         const { tokenUsage, ...rest } = buffer.currentOutput;
         const timingLogFields = llmAttemptLogFields({
-          providerId: this.modelConfig.providerId,
           durationMs,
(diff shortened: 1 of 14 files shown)


Nils-Fedrigo (approved):
nice thank you!

PR #31150, opened 2026-08-25
[front] user menu: single analytics entry point

PopDaph:
In the user menu, the credit usage card's "Learn more" button and the Analytics item both opened the analytics dialog.

- Rename "Learn more" to "See your usage".
- Show the Analytics item only when the usage card is not in the menu (non-credit plan, or above-target usage where the card moves to the sidebar), so exactly one entry point is visible.


aubin-tchoi (approved):
LGTM 👌🏻

PR #31153, opened 2026-08-26
[front] fix: make user ID nullable on conversation MCP server views

aubin-tchoi:
## Description

This PR makes the column `userId` nullable on `conversation_mcp_server_views`.
We insert in this table when enabling a toolset.
We actually never ever read the user ID.
Requiring a user prevents using this from the public API.

## Risk

- Low.

## Deploy Plan

- Run the pre-deploy migration.
- Deploy front.


id13:
> We actually never ever read the user ID.

What's the case for keeping the column then ?

id13 (approved):

aubin-tchoi:
unsure, looks like we used to read from it but don't anymore. Don't have too much context and just want to unblock myself so will check later

PR #31157, opened 2026-08-26
[front] Move isRestricted off SpaceType onto EnrichedSpaceType

tdraier:
## What

`isRestricted` is grant-derived (a space is restricted unless the workspace global group holds a `reader` grant), so — like `groupIds` — it shouldn't force the eager `group_permissions` include on every space load. This moves it off the base `SpaceType`.

- `SpaceTypeWithGroupIds` → **`EnrichedSpaceType`** = `SpaceType & { groupIds; isRestricted }`; `AppTypeWithSpaceGroupIds` → `EnrichedAppType`.
- The on-demand loader is now `SpaceResource.listSpaceEnrichmentBySpaceModelId` → `Map<modelId, { groupIds, isRestricted }>` (one `group_permissions` query). Serializers use `SpaceResource.toJSONEnriched` / `AppResource.toJSONEnriched`. **`toJSON()` no longer reads `this.groups` at all.**
- FE consumers of `space.isRestricted` already receive enriched spaces from `useSpaces`/`useSpaceInfo`/`useSpacesAsAdmin`, so this is annotation-widening (`SpaceType` → `EnrichedSpaceType`) up to those hooks — no runtime change.
- `with_space` middleware reads restriction from the resource (`isRegularAndRestricted() || isProjectAndRestricted()`), not `toJSON()`. Poke (spaces list/detail, agent/skill details) and app serializers load the enrichment on demand.

## Why

With this, base `SpaceType`/`toJSON()` is fully independent of the eagerly-loaded grants. Combined with the async `isOpen()` work and moving the 3 remaining group-management helpers to on-demand loads, the eager `spaceGrants` include can then be dropped — no schema/DB column needed.

## Swagger

`PrivateSpace` drops `isRestricted`; public `Space`, `PrivateProject`, and the space-detail response add it. `swagger.json` regenerated.

## Test plan

- `tsgo --noEmit` clean in both `front` and `front-api` (2 pre-existing unrelated errors remain).
- `lint:swagger-annotations` OK; swagger regenerates with no errors.
- 154 tests passing: `space_resource` (77), `spaces` api (41), v1 spaces (5), private spaces list (5) + detail (2), access-check (8), pods apps export (4) / index (9), project_tasks (3).

### DIFF
--- a/front/lib/api/projects/list.ts
+++ b/front/lib/api/projects/list.ts
@@ -4,7 +4,7 @@ import { SpaceResource } from "@app/lib/resources/space_resource";
 import { removeDiacritics } from "@app/lib/utils";
 import logger from "@app/logger/logger";
 import type { SearchProjectsResponseBody } from "@app/types/api/projects/list";
-import type { PodType, SpaceType } from "@app/types/space";
+import type { EnrichedSpaceType, PodType } from "@app/types/space";
 
 type ListPodsAccess = "member" | "open";
 
@@ -208,13 +208,10 @@ export async function enrichProjectsWithMetadata(
     metadatas.map((m) => [m.spaceId, m])
   );
 
-  const groupIdsBySpaceModelId = await SpaceResource.listGroupIdsBySpaceModelId(
-    auth,
-    { spaces }
-  );
+  const enrichedSpaces = await SpaceResource.batchToJSONEnriched(auth, spaces);
 
-  return spaces.map((space) => ({
-    ...space.toJSONWithGroupIds(groupIdsBySpaceModelId.get(space.id) ?? []),
+  return spaces.map((space, i) => ({
+    ...enrichedSpaces[i],
     description: metadataMap.get(space.id)?.description ?? null,
     isMember: space.isMember(auth),
     isEditor: space.canAdministrate(auth),
@@ -222,7 +219,7 @@ export async function enrichProjectsWithMetadata(
   }));
 }
 
-export type ProjectWithAdminMetadata = SpaceType & {
+export type ProjectWithAdminMetadata = EnrichedSpaceType & {
   description: string | null;
   archivedAt: number | null;
   todoGenerationEnabled: boolean;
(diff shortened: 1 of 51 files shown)


tdraier:

- #31157 ◀
    - #31162

front/lib/api/projects/list.ts
```diff
@@ -208,13 +211,13 @@ export async function enrichProjectsWithMetadata(
...
-    { spaces }
-  );
+  const enrichmentBySpaceModelId =
+    await SpaceResource.listSpaceEnrichmentBySpaceModelId(auth, { spaces });
 
   return spaces.map((space) => ({
-    ...space.toJSONWithGroupIds(groupIdsBySpaceModelId.get(space.id) ?? []),
+    ...space.toJSONEnriched(
+      enrichmentBySpaceModelId.get(space.id) ?? EMPTY_SPACE_GRANT_ENRICHMENT
+    ),
```
fabiencelier: maybe we can do a `SpaceResource.batchToJSONEntiched(SpaceResource[]): EnrichedSpaceType[]` helper which takes a list of SpaceResource and wrap all this logic inside if we re-use it a lot 
So we keep the logic inside the resource and don't leak it everywhere

fabiencelier (approved):

front/lib/resources/space_resource.ts
```diff
@@ -2537,39 +2551,64 @@ export class SpaceResource extends BaseResource<SpaceModel> {
...
-  // on `toJSON` (it would force the eager `group_permissions` include on every space load); the
-  // endpoints that expose it — the public API for backward compatibility, and the space-management
-  // UI — load it on demand via `listGroupIdsBySpaceModelId` and pass it here.
-  toJSONWithGroupIds(groupIds: string[]): SpaceTypeWithGroupIds {
+  // Serialize with the space's grant-derived fields (`groupIds` and `isRestricted`). Private: these
+  // are not carried on `toJSON` (that would force the eager `group_permissions` include on every
+  // space load), so callers go through the batched `batchToJSONEnriched` rather than pairing this
+  // with the loader themselves.
+  private toJSONEnriched({
+    groupIds,
```
flvndvd, line 2559: This slightly goes against the resource approach, here it basically open doors to fake private group belonging, and privacy of a space from the outside. A resource is meant to be self-contained. 

Why aren't we loading `listSpaceEnrichmentBySpaceModelId` everytime we fetch this resource and cache it to preserve this invariant?

PR #31158, opened 2026-08-26
[front] cleanup: remove legacy agent observability components

aubin-tchoi:
## Description

The agent observability charts were replaced with the new analytics view.

This PR removes the corresponding components and hooks, all now unreachable.

No need to republish the extension, agent builder and agent info page are not available there.

## Tests

- Checked locally.

## Risk

- Low.

## Deploy Plan

- Deploy front.


front/components/agent_builder/observability/utils.ts
```diff
@@ -17,6 +16,13 @@ type VersionMarker = { version: string; timestamp: number };
 
 type SourceBucket = { origin: string; count: number };
 
+type SourceChartDatum = {
```
aubin-tchoi: just moved from a file that ends up being removed

avervaet (approved):

JulesBelveze:
Thank your for your services - you'll be missed 😢

PR #31159, opened 2026-08-26
[front] feat: allow email or user ID in batch-update-members poke plugin

tdraier:
## Description

The `batch-update-members` poke plugin resolved members by email only. It keyed a `Map` by email, so when several accounts share the same email the entries collapsed and the plugin silently acted on one arbitrary account.

Each line of the input can now be **either an email or a user ID** (detected by the presence of `@`):

- User IDs resolve to exactly one account (via `fetchByIds`), so operators can target a specific account.
- Emails that match several accounts now return an `ambiguous_email` status instead of acting on an arbitrary one, prompting the operator to re-run that line with the user ID.
- Each result reports both the typed identifier and the resolved email.

The `emails` arg was renamed to `members` (label "Emails or user IDs"); audit-log lists now record the resolved user's email.

## Tests

Manual (poke plugin). `tsgo --noEmit` and Biome pass.

## Risk

Low. Internal support-only poke plugin; no external API surface. Backward compatible for email-only input, except a duplicate email is now flagged rather than silently mishandled.

## Deploy Plan

Standard `front` deploy.

### DIFF
--- a/front/lib/api/poke/plugins/workspaces/batch_update_members.ts
+++ b/front/lib/api/poke/plugins/workspaces/batch_update_members.ts
@@ -25,16 +25,30 @@ import { mapToEnumValues } from "@app/types/poke/plugins";
 import { Err, Ok } from "@app/types/shared/result";
 import { ASSIGNABLE_ROLES, isAssignableRoleType } from "@app/types/user";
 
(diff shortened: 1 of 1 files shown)


tdraier:

- #31159 ◀

matteotrab (approved):

PR #31160, opened 2026-08-26
[front] Agent loop: log stuck tools when finalizing an errored run

PopDaph:
## Description

We're trying to understand and fix the heartbeat-timeout workflow failures (87% of agentLoopWorkflow failures, see #31054). To pick the right fix we need to know which tools are involved. Worker-side logging can't tell us: in 16 of 20 failures since #31106 deployed, the worker died before emitting a single log line.

The database can. The action rows are persisted (by the model step) before the tool runs, and a dead worker can't finalize them. So when `finalizeErroredAgentLoopActivity` runs, the actions still in a non-final status are exactly the tools that never finished. This PR logs them (`stuckTools`: tool name, MCP server, status) alongside the workflow error, giving attribution for every failure, including silent worker deaths.

The attribution is best-effort: a failing actions read is logged and never blocks the billing/analytics/email finalizers.

## Tests

Resource test for `listNonFinalActionsForAgentMessage`, unit tests for the log shape and the non-blocking behavior.

## Risk

Low: one best-effort DB read and one warn log per errored finalize (~1k/week).

## Deploy Plan

- [ ] Deploy front.
- [ ] Attribution query: `service:front-agent-loop-worker "Agent loop finalized as errored"`, `stuckTools` names the tools.

### DIFF
--- a/front/temporal/agent_loop/activities/finalize.test.ts
+++ b/front/temporal/agent_loop/activities/finalize.test.ts
@@ -1,6 +1,10 @@
 import type { Authenticator } from "@app/lib/auth";
+import { AgentMCPActionResource } from "@app/lib/resources/agent_mcp_action_resource";
+import logger from "@app/logger/logger";
 import { creditsExhaustedMessage } from "@app/temporal/agent_loop/activities/common";
-import { describe, expect, it } from "vitest";
+import { logStuckToolsForErroredAgentMessage } from "@app/temporal/agent_loop/activities/finalize";
+import type { AgentLoopArgs } from "@app/types/assistant/agent_run";
+import { afterEach, describe, expect, it, vi } from "vitest";
 
 describe("creditsExhaustedMessage", () => {
   it("tells admins to purchase more credits", () => {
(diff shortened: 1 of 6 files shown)


front/temporal/agent_loop/activities/finalize.test.ts
```diff
@@ -17,3 +23,86 @@ describe("creditsExhaustedMessage", () => {
     );
   });
 });
+
+describe("logStuckToolsForErroredAgentMessage", () => {
+  const error = { message: "Activity task timed out", name: "ActivityFailure" };
```
flvndvd, line 28: Why not using Error instance?

front/temporal/agent_loop/activities/finalize.ts
```diff
@@ -170,15 +174,85 @@ export async function finalizeCreditStoppedAgentLoopActivity(
   ]);
 }
 
+// Attribute the failure to the tools that never finished. The worker that ran them may have died
+// without logging anything (heartbeat timeouts), but the action rows it left behind in a
+// non-final status carry the tool identity. Diagnostic only: it must never fail the finalize
+// activity, which carries billing, analytics and email side effects, so the DB read is
+// best-effort. Exported for tests.
```
flvndvd, line 181: Find to keep a short time for debugging, but `finalize` is not the ideal spot for longer term as it slows down the stop of the agent loop, sometimes preventing tool approval to restart the loop faster.

flvndvd (approved):

PopDaph:
Merging and I'll removing it early next week 🙏🏻

PR #31161, opened 2026-08-26
fix: resolve analytics filter names hydrated from the URL

adrsimon:
## Description

The analytics consumption view keeps its filter in the query string, which only carries ids. On refresh `usageFilterFromIds` built each option with the id as its display name, so the filter chips and the panel's "filters selected" list showed raw sIds (and lost avatars/icons) for every category until the user reselected the option by hand.

`useResolvedUsageFilter` now swaps those placeholders for their facet counterpart, taking the whole option so names, pictures, agent scope and tool icons all come back.

`ConsumptionFacetOptions` and `EMPTY_FACET_OPTIONS` moved to `usageFilter.ts` so the new transform sits with the other per-category filter helpers. That also removes the two existing copies of the empty-options constant.

## Tests

Unit tests on `resolveUsageFilter` in `usageFilter.test.ts`.

## Risk

Low.

## Deploy Plan

Deploy front.


aubin-tchoi (approved):
LGTM

id13:
Thanks ! ✨

PR #31162, opened 2026-08-26
[front] Stop fetching all admin spaces to invalidate MCP views

tdraier:
## What

`useMutateMCPServersViewsForAdmin` (used by ~8 MCP mutation hooks — `useDeleteMCPServer`, `useCreate*`, `useSync*`, `useUpdateMCPServerView`, connect, and `MCPServerDetails`) called `useSpacesAsAdmin` on mount, fetching **every** workspace space (enriched with `groupIds`/`isRestricted`) solely to `spaces.find(s => s.kind === "system")` and rebuild one SWR key to invalidate after a server mutation. The spaces were never rendered or read.

Now it invalidates the space-scoped `mcp_views` keys by pattern via `useSWRConfig().mutate(predicate)` — no spaces fetched, no enrichment:

```ts
await globalMutate(
  (key) =>
    typeof key === "string" &&
    key.startsWith(`/api/w/${owner.sId}/spaces/`) &&
    key.includes("/mcp_views")
);
```

`mutate` is only called *after* a create/delete/update/sync succeeds, so pattern invalidation is sufficient (nothing needs to be pre-fetched at render time).

## Why

This eager fetch was the source of the `GET /spaces?role=admin` call on the new-conversation page (via `MCPServerDetails`, mounted closed) — and on every other admin page mounting these hooks. Removing it kills that query everywhere, along with the now-pointless space enrichment.

## Trade-off

Invalidates all spaces' `mcp_views` keys rather than only the system space's. This is safe (revalidation, not mutation), and MCP server views for admin config live in the system space anyway — the effective set is the same.

## Test plan

- `tsgo --noEmit` clean (front).
- No behavior change for callers — `useMutateMCPServersViewsForAdmin` still returns `{ mutate }`; the returned `mutate` still invalidates the MCP views + servers caches after a mutation.

### DIFF
--- a/front/lib/swr/mcp_servers.ts
+++ b/front/lib/swr/mcp_servers.ts
@@ -34,7 +34,6 @@ import type {
   PostConnectionResponseBody,
 } from "@app/lib/resources/mcp_server_connection_resource";
 import type { GetMCPServerViewsResponseBody } from "@app/lib/resources/mcp_server_view_resource";
-import { useSpacesAsAdmin } from "@app/lib/swr/spaces";
 import { emptyArray, useFetcher, useSWRWithDefaults } from "@app/lib/swr/swr";
 import type { DiscoverOAuthMetadataResponseBody } from "@app/types/api/oauth/providers/mcp";
 import type { WithAPIErrorResponse } from "@app/types/error";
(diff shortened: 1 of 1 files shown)


tdraier:

- #31157
    - #31162 ◀

ykmsd (approved):

PR #31166, opened 2026-08-26
[front] cleanup: drop unused agent observability endpoints

aubin-tchoi:
## Description

Follow-up to #31158. This PR removes the endpoints that served the legacy agent observability UI, now left without callers

No change on the documentation, all of these were in the private API + not documented.
No need to re-publish the extension, agent builder and agent info are not available there.

## Risk

- Low.

## Deploy Plan

- Deploy front.


sfriquet (approved):
🔥

PR #31168, opened 2026-08-26
Remove anthropic_cache_diagnostics feature flag, enable for everyone

davidebbo:
## Summary

- Removed the `anthropic_cache_diagnostics` flag from `WHITELISTABLE_FEATURES_CONFIG` and from the SDK's `WhitelistableFeaturesSchema` union — the feature is now unconditionally on instead of `dust_only`.
- Removed the `cacheDiagnosticsEnabled` gating that was threaded through `run_model.ts` → `GetOutputRequestParams` (`types.ts`) → `get_output_from_llm.ts`; the previous-message-id lookup, storage, and cache-miss logging/metrics now always run.
- Updated a stale comment in `cache_diagnostics.ts` that referenced sunsetting the feature by "flipping the flag off."
- Left the tri-state `cacheDiagnosticsEnabled` getter in `anthropic.ts` untouched — that's an unrelated per-request check (`previousMessageId !== undefined`), which now is effectively always true since the caller always fetches it.


flvndvd (approved):
Thank you 🙏 . Apologies meant to do it and forgot

PR #31171, opened 2026-08-26
Prepare sandbox schema for Frames v2

fontanierh:
## Description

- Add publication-scoped sandbox function rows while reusing the existing file and slug columns.
- Add Frame FileResource ownership to sandbox owners and extend the exactly-one-owner constraint.
- Keep legacy Pod Function indexes and fetch behavior during the transition.

## Tests

Replayed all migrations on a fresh Hive Postgres database and ran the sandbox function and sandbox resource suites.

## Risk

Low. The migration is an expand step; existing Pod uniqueness remains enforced by the workspace and Space composite indexes.

## Deploy Plan

Apply and verify the pre-deploy migration in both regions before deploying Frames v2 application code. No backfill required.

### DIFF
--- a/front/lib/resources/storage/models/sandbox_function.ts
+++ b/front/lib/resources/storage/models/sandbox_function.ts
@@ -51,8 +51,9 @@ export class SandboxFunctionModel extends WorkspaceAwareModel<SandboxFunctionMod
   declare createdAt: CreationOptional<Date>;
   declare updatedAt: CreationOptional<Date>;
 
-  declare spaceId: ForeignKey<SpaceModel["id"]>;
+  declare spaceId: ForeignKey<SpaceModel["id"]> | null;
   declare fileId: ForeignKey<FileModel["id"]>;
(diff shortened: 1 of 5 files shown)


spolu:
🙌

front/lib/resources/storage/models/sandbox_function.ts
```diff
@@ -100,12 +101,16 @@ SandboxFunctionModel.init(
...
     spaceId: {
       type: DataTypes.BIGINT,
-      allowNull: false,
+      allowNull: true,
     },
     fileId: {
       type: DataTypes.BIGINT,
       allowNull: false,
     },
+    publicationId: {
```
davidebbo, line 110: What do these things look like in practice?

fontanierh: probably some kind of sId or UID

davidebbo (approved):
LGTM

PR #31172, opened 2026-08-26
Admin governance for skills: canRead for skills based on permissions

fabiencelier:
## Description

related to https://github.com/dust-tt/tasks/issues/10052

 - Add global group as reader on all skill of the workspace
 - Backfill this for existing workspaces
 - Check `canRead` in baseFetch 

Note: in this PR we still have give read permission to every user (by role), will be dropped in follow up PR when the migration has run and the global group has reader permission on all skills.

## Tests

tested the migration locally

## Risk

low:
- all user keep read access, purely additive
- can be reverted

## Deploy Plan

deploy front
run migration to backfill existing workspaces 
```
 npx tsx front/scripts/seed_missing_governance_capabilities.ts --capability reader:skill
```


tdraier (approved):

tdraier:
Seen IRL - might be better for now to use a global grant ( -1  / reader ) instead of unconditionnally create grant on every instance

philipperolet:
> Seen IRL - might be better for now to use a global grant ( -1 / reader ) instead of unconditionnally create grant on every instance

Agreed 👍

fabiencelier:
Switched to a global grant ( -1 / reader ) to reduce number of grants

PR #31177, opened 2026-08-26
feat(connectors): list and remove endpoints for the Slack summoning whitelist

adrsimon:
## Description

A Slack workflow that mentions @/Dust is refused unless its sender name is whitelisted for `summon_agent`. Today that only happens through poke and is handled by support team. There's no built-in way for users, nor support, to see in product which workflow is allowed and to remove any of them. 

The goal of this stack is to make the whole thing self-served by workspace admins.

Connectors only exposed a write path. This adds GET, POST and DELETE on `/slack/bots/summoning_whitelist`, plus the matching `listWhitelistedBots` / `removeWhitelistedBot` on `SlackConfigurationResource`, so front can read and mutate the allowlist without going through `admin`.

The stack:

1. This PR: connectors endpoints.
2. `front` business layer in `lib/api/slack/summoning_whitelist.ts` (allow, list, revoke, group validation, audit events), poke plugin moved onto it.
3. Poke lists and revokes allowed workflows on the Slack bot data source page. Drops the Metabase link.
4. Admin UI on the automations page, and the Slack denial message points there.

## Tests

UI tested on a local dev setup. I didn't test the actual workflow sending a message.

## Risk

Low. Three new endpoints, no change to any existing read or write path.

## Deploy Plan

Deploy connectors first, it gates the rest of the stack.

### DIFF
--- a/connectors/src/api/slack_bot_summoning_whitelist.ts
+++ b/connectors/src/api/slack_bot_summoning_whitelist.ts
@@ -1,0 +1,186 @@
+import { apiError, withLogging } from "@connectors/logger/withlogging";
+import { ConnectorResource } from "@connectors/resources/connector_resource";
+import type { WhitelistedBotType } from "@connectors/resources/slack_configuration_resource";
+import { SlackConfigurationResource } from "@connectors/resources/slack_configuration_resource";
+import type { WithConnectorsAPIErrorReponse } from "@connectors/types";
+import { zodParsePayload } from "@connectors/types";
+import type { Request, Response } from "express";
+import { z } from "zod";
+
+const WHITELIST_TYPE = "summon_agent" as const;
+
+const ConnectorIdQuerySchema = z.object({
+  connector_id: z.string(),
+});
+
+const PostBodySchema = z.object({
+  connector_id: z.string(),
+  bot_name: z.string().trim().min(1),
+  group_ids: z.array(z.string()).min(1),
+});
+
+const DeleteBodySchema = z.object({
+  connector_id: z.string(),
+  bot_name: z.string().trim().min(1),
+});
+
+type GetSlackBotSummoningWhitelistResBody = WithConnectorsAPIErrorReponse<{
+  bots: WhitelistedBotType[];
+}>;
+
+type MutateSlackBotSummoningWhitelistResBody = WithConnectorsAPIErrorReponse<{
+  success: true;
+}>;
+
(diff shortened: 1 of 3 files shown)


connectors/src/api/slack_bot_summoning_whitelist.ts
```diff
@@ -0,0 +1,182 @@
...
+  success: true;
+}>;
+
+async function fetchSlackConfiguration(connectorId: string) {
+  const connector = await ConnectorResource.fetchById(connectorId);
+  if (!connector || connector.type !== "slack_bot") {
+    return null;
+  }
+
+  return SlackConfigurationResource.fetchByConnectorId(connector.id);
```
flvndvd: IIRC `ConnectorResource` already fetches this for you.

connectors/src/api/slack_bot_summoning_whitelist.ts
```diff
@@ -0,0 +1,182 @@
...
+      api_error: {
+        type: "connector_not_found",
+        message: "Slack bot connector not found",
+      },
+      status_code: 404,
+    });
+  }
+
+  res.status(200).json({
+    bots: await slackConfig.listWhitelistedBots(WHITELIST_TYPE),
```
flvndvd: Can we avoid `await` in the return or json?

connectors/src/api/slack_bot_summoning_whitelist.ts
```diff
@@ -0,0 +1,182 @@
...
+const _getSlackBotSummoningWhitelistHandler = async (
+  req: Request<Record<string, string>>,
+  res: Response<GetSlackBotSummoningWhitelistResBody>
+) => {
+  const queryValidation = zodParsePayload(req.query, ConnectorIdQuerySchema);
+  if (queryValidation.isErr()) {
+    return apiError(req, res, {
+      api_error: {
+        type: "invalid_request_error",
+        message: `Invalid query: ${queryValidation.error}`,
```
flvndvd, line 57: What's the plan for errors here? Pass-plat to customers? If so, let's use the `fromError` from zod to have human friendly errors.

flvndvd (approved):

connectors/src/api/slack_bot_summoning_whitelist.ts
adrsimon: zodParsePayload already does a fromError iirc

PR #31178, opened 2026-08-26
feat(front): business layer for the Slack workflow allowlist

adrsimon:
## Description

Stacked on #31177.

`allowSlackWorkflow`, `listSlackWorkflows` and `revokeSlackWorkflow` in `lib/api/slack/summoning_whitelist.ts` wrap the connectors endpoints so callers get one function per operation: group sIds validated against the workspace, the global group always appended so the workflow can at least reach workspace-wide agents, and a `slack_workflow.allowed` / `slack_workflow.revoked` audit event on success. Failures come back as a `SlackWorkflowError` with a domain type, no HTTP shapes.

Listing returns `isSlackBotConnected: false` rather than an error when the workspace has no Slack bot, so callers can hide the feature instead of surfacing a failure.

The poke plugin now calls `allowSlackWorkflow` instead of the `slack whitelist-bot` admin command. Same group kinds in the picker, same "groups are required" guard, so nothing changes for support.

## Tests

Covered end to end by the endpoint tests in the next two PRs. Ran the poke plugin against a local hive to confirm the audit event and the appended global group.

## Risk

The poke plugin changes path while keeping its behaviour, so the risk is limited to the new code path itself.

## Deploy Plan

Register the new audit schemas after merge, before deploying front: `npx tsx front/admin/register_audit_log_schemas.ts --execute --changed`.

### DIFF
--- a/front/lib/api/poke/plugins/data_sources/slack_whitelist_bot.ts
+++ b/front/lib/api/poke/plugins/data_sources/slack_whitelist_bot.ts
@@ -1,6 +1,7 @@
 import config from "@app/lib/api/config";
 import { createPlugin } from "@app/lib/api/poke/types";
 import { config as regionsConfig } from "@app/lib/api/regions/config";
+import { allowSlackWorkflow } from "@app/lib/api/slack/summoning_whitelist";
 import { GroupResource } from "@app/lib/resources/group_resource";
 import logger from "@app/logger/logger";
 import type { AdminCommandType } from "@app/types/connectors/admin/cli";
@@ -126,7 +127,6 @@ export const slackWhitelistBotPlugin = createPlugin({
     });
   },
   execute: async (auth, resource, args) => {
-    const owner = auth.getNonNullableWorkspace();
     const { botName, groupIds } = args;
 
     if (!resource) {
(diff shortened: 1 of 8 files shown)


front/lib/api/poke/plugins/data_sources/slack_whitelist_bot.ts
```diff
@@ -141,38 +141,13 @@ export const slackWhitelistBotPlugin = createPlugin({
...
-        wId: owner.sId,
-        groupId: allGroupIds.join(","),
-        whitelistType: "summon_agent",
-        providerType: "slack_bot",
-      },
-    };
-
-    const adminCommandRes = await connectorsAPI.admin(whitelistBotCmd);
-    if (adminCommandRes.isErr()) {
+    const allowRes = await allowSlackWorkflow(auth, {
```
flvndvd, line 144: IIRC this same is used when one wants to update the associated spaces, right?

front/lib/api/slack/summoning_whitelist.ts
```diff
@@ -0,0 +1,210 @@
...
+export class SlackWorkflowError extends Error {
+  constructor(
+    readonly type: SlackWorkflowErrorType,
+    message: string
+  ) {
+    super(message);
+  }
+}
+
+function getConnectorsAPI() {
```
flvndvd: <laziness_emoji>

flvndvd (approved):

front/lib/api/slack/summoning_whitelist.ts
adrsimon: oops indeed, this one is ugly sorry

PR #31179, opened 2026-08-26
feat(poke): list and revoke allowed Slack workflows from the data source page

adrsimon:
## Description

Stacked on #31178.

Whitelisting a Slack workflow was write-only from poke: to see what a workspace had allowed you followed a Metabase link, and nothing could be taken back short of a SQL delete. The Slack bot data source page now carries a table of allowed workflows with their groups and when they were added, plus a revoke button, backed by GET and DELETE on `/api/poke/workspaces/:wId/slack-workflows`.

With the listing in place the Metabase link on the whitelist plugin has no reason to exist, so it goes.

## Tests

`front-api/routes/poke/workspaces/[wId]/slack-workflows.test.ts`: superuser gating, group name resolution, workspace with no Slack bot, revoke and its audit event, unknown bot name.

## Risk

Low, poke only. Revoke is destructive but confirmed and scoped to one bot name.

## Deploy Plan

Deploy front and front-api.


front-api/lib/api/slack_workflows.ts
```diff
@@ -0,0 +1,44 @@
+import type { SlackWorkflowError } from "@app/lib/api/slack/summoning_whitelist";
+import type { APIErrorWithContentfulStatusCode } from "@app/types/error";
+import { assertNever } from "@app/types/shared/utils/assert_never";
+
+export function httpErrorForSlackWorkflowError(
```
aubin-tchoi: nit: not a fan of having http error codes in `lib/api`, imho the methods in lib api should use a business-driven error categorization that the handler converts into http error codes

aubin-tchoi (approved):
LGTM

PR #31180, opened 2026-08-26
feat(front): let admins allow Slack workflows from the automations page

adrsimon:
## Description

A workflow that mentions @Dust in Slack gets refused until someone whitelists it, and that meant asking us. Workspace admins now handle it themselves: the automations page gains a second tab, Slack workflows, listing what is allowed with the spaces each one can reach, when it was added, and a revoke button. Adding one takes a sender name and a set of spaces; the groups stored in connectors are derived from those spaces, so admins never see groups. The tab carries its own credits card so the cost of workflow consumption is visible next to the triggers one. The table is searchable, sortable and paginated.

Self-serve is gated on credit-priced plans, both in the UI (the tab only shows for admins on such a plan) and on the endpoints. Legacy plans keep going through poke, which is not gated.

Names have to match the sender name Slack shows on the workflow's messages, exactly. We do not enumerate workflows from the Slack API, so this stays manual entry, with the dialog spelling out the constraint.

The spaces dropdown fetches on open and filters client-side. There is no endpoint that searches spaces by name today; worth adding if the list gets long enough to matter.

The denial message the bot posts in Slack is unchanged: everyone still gets told to email support. Making it point admins at the automations page needed a plan lookup in connectors, which does not know about plans, and the endpoint plus memoized cache that took was not worth it. Follow-up if we want the pointer.

## Tests

`front-api/routes/w/[wId]/slack-workflows/index.test.ts`: admin gating, credit-priced plan gating, space resolution on the listing, workspace with no bot, allow with the global group appended and its audit event, empty name and unknown space rejections, revoke, unknown bot name.

`front-api/routes/w/[wId]/slack-workflows/overview.test.ts`: admin and plan gating, the credits aggregation, Elasticsearch failure.

Checked the tab against a dev workspace seeded with `slack_workflow` consumption and 30 allowed workflows spread over several spaces.

## Risk

Admins on a credit-priced plan can grant a Slack workflow access to their spaces' agents without going through us. That is the point, and it stays scoped to spaces the admin already administers.

Anyone who can run the workflow in Slack can then summon those agents, guests included, which the dialog states.

## Deploy Plan

Deploy front and front-api. Nothing to do on connectors.


adrsimon:
Superseded by the split stack starting at #31292; reopening the UI part as a separate PR since the base could not be retargeted.

PR #31183, opened 2026-08-26
[front] fix: enforce Dust-only-ness on feature flags flagged as Dust-only

aubin-tchoi:
## Description

Dust-only feature flags could be enabled on any workspace through Poke. This PR prevents enabling a `dust_only` flags for non-Dust workspaces.

Also allows friends and family workspaces, because, you know.

Doesn't change flags already enabled, because they are already enabled.

This comes with having a better hygiene of the status on our end, will do a pass to review in which stage each FF is (there are probably many of them that were never moved to the next stage).

## Tests

- Added a test.

## Risk

- Low.

## Deploy Plan

- Deploy front.


PopDaph (approved):
Most likely we need to do a pass to update some of them! 
I'm afraid it won't necessarily solve our issue but mostly shift it to the "on_demand" status, but at least this will give us one "safe status"!

PR #31184, opened 2026-08-26
Add conversation window checkpoint storage

flvndvd:
## Description

This PR adds the durable storage layer required for stateful conversation windows across agent-loop steps.

Today, every model step rebuilds its context from authoritative conversation data and tokenizes the full rendered window again. The broader project will allow step N to start from the validated checkpoint produced by step N-1, then hydrate and tokenize only the new interaction data.

This PR focuses exclusively on persistence. It stores versioned conversation-window snapshots in GCS using an explicit path scoped by workspace, conversation, agent message, agent version, schema version, and step.
Checkpoints are compressed, validated when loaded, and written with create-once semantics so retries cannot silently overwrite an existing step. As content repeats heavily gzip compression is pretty impressive here!

The storage layer also distinguishes a missing checkpoint from an invalid or unavailable one, validates signed URL expiration before reuse, and keeps failures compatible with a later full-rendering fallback.

Nothing is wired into the agent loop yet, so this PR does not change conversation rendering behavior.

### DIFF
--- a/front/lib/api/assistant/conversation_rendering/conversation_window_checkpoint.ts
+++ b/front/lib/api/assistant/conversation_rendering/conversation_window_checkpoint.ts
@@ -26,28 +26,40 @@ const SIGNED_URL_EXPIRY_SAFETY_MARGIN_MS = 5 * 60 * 1000;
 const MAX_STORED_CHECKPOINT_SIZE_BYTES = 8 * 1024 * 1024;
 const MAX_EXPANDED_CHECKPOINT_SIZE_BYTES = 64 * 1024 * 1024;
 
-export type ConversationWindowCheckpointIdentity = {
-  workspaceId: string;
-  conversationId: string;
-  agentMessageId: string;
-  agentMessageVersion: number;
-  step: number;
-};
+const checkpointIdentitySchema = z
+  .object({
+    workspaceId: z.string(),
+    conversationId: z.string(),
+    agentMessageId: z.string(),
+    agentMessageVersion: z.number().int().nonnegative(),
+    step: z.number().int().nonnegative(),
+  })
+  .strict();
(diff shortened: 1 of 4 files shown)


front/lib/file_storage/signed_url_cache.ts
```diff
@@ -5,7 +5,7 @@ import {
 import { cacheWithRedis } from "@app/lib/utils/cache";
 
 // LLM providers may fetch a cached image URL several minutes after it was signed.
-export const MODEL_INPUT_SIGNED_URL_EXPIRATION_DELAY_MS = 15 * 60 * 1000;
+export const MODEL_INPUT_SIGNED_URL_EXPIRATION_DELAY_MS = 60 * 60 * 1000;
```
flvndvd, line 8: Bumped from 15 to 60 minutes as most of our provider caching is 30 minutes, this is meant to give us more flexibility on the caching.

front/lib/api/assistant/conversation_rendering/conversation_window_checkpoint.ts
```diff
@@ -0,0 +1,396 @@
...
+  onMissingAction = "inject-placeholder",
+  agentConfigurationId,
+}: {
+  model: ModelConfigurationType;
+  prompt: string;
+  tools: string;
+  allowedTokenCount: number;
+  leadingMessages: ModelMessageTypeMultiActionsWithoutContentFragment[];
+  excludeActions?: boolean;
+  excludeImages?: boolean;
```
flvndvd, line 269: @id13 will consolidate all the boolean settings in a last PR.

id13 (approved):

adrsimon (approved):
nice!

PR #31188, opened 2026-08-26
[front] chore: update FF stages

aubin-tchoi:
## Description

This PR does a pass on existing feature flags to update the stage (dust-only, on demand) on feature flags that we seem to never have updated.

## Risk

- Low. Actually does nothing by itself, only in combination with https://github.com/dust-tt/dust/pull/31183

## Deploy Plan

- Deploy front.


PopDaph (approved):
👍🏻

fabiencelier (approved):
LGTM

PR #31190, opened 2026-08-26
fix: use common model picker for bulk agent updates

adrsimon:
## Description

Use the Manage Agents model-menu hierarchy for bulk model changes. The bulk flow now confirms the target model and affected agent count before applying the existing batch update.

## Tests

Rendered live

Capture d’écran 2026-08-26 à 14 46 13

## Risk

UI-only change.

## Deploy Plan

Deploy front.


aubin-tchoi (approved):
LGTM

PR #31195, opened 2026-08-26
Register Frames v2 FileResources

fontanierh:
## Description

- Register `application/vnd.dust.frame.v2+json` as the internal manifest FileResource type.
- Add `FileResource.isFrameV2`.
- Keep Frames v2 out of legacy Interactive Content processing and public conversation serialization.
- Accept the type in generated-file SDK payloads.

Publication, activation, and GCS writes remain deferred. The existing `frames_v2` flag stays disabled.

## Tests

Added focused file-type, FileResource, and processing tests. Tested locally with the affected suites, front typechecking, and SDK type generation.

## Risk

Low. This registers an internal type without adding a producer or runtime entry point. Safe to roll back.

## Deploy Plan

Standard deploy.


davidebbo (approved):
LGTM

PR #31202, opened 2026-08-26
Store immutable Frames v2 publications

fontanierh:
## Description

- Store Frames v2 source files under a new immutable publication ID.
- Upload source files with bounded concurrency and write `manifest.json` last as the commit marker.
- Return typed validation failures before any GCS write.

Loading, activation, function registration, and runtime wiring remain deferred.

## Tests

Added focused coverage for canonical paths, manifest-last ordering, source validation, workspace isolation, and partial write failures.

## Risk

Low. There is no producer yet and the feature flag remains disabled. A failed write can leave source objects without a manifest; readers will not treat that prefix as a publication.

## Deploy Plan

Standard deploy.


davidebbo (commented):
LGTM

PR #31204, opened 2026-08-26
Update GPT-5.6 Sol pricing

fontanierh:
## Description

Update GPT-5.6 Sol to OpenAI's current $4 input, $0.40 cached input, $5 cache write, and $20 output rates per million tokens. Keep global and EU endpoint metadata in sync.

## Tests

Added focused coverage for the billing table and global and EU endpoint pricing.

## Risk

Low. This only lowers Sol cost attribution and is safe to revert.

## Deploy Plan

Standard deploy.


flvndvd (commented):
LGTM

PR #31206, opened 2026-08-26
[front] Allow Dust-only feature flags on local envs

PopDaph:
Skip the Dust/Friends & Family plan check in the toggle-feature-flag poke plugin when `isDevelopment()`, so `dust_only` flags can be enabled freely on local environments.

Adds a test covering the local bypass.


aubin-tchoi (approved):
🙏🏻

PR #31210, opened 2026-08-26
chore(front): remove the models_picker feature flag

Nils-Fedrigo:
## Description

Removes the `models_picker` feature flag and enables the model picker for every workspace.

The flag is gone from `WhitelistableFeature` (`front/types/shared/feature_flags.ts` and the JS SDK) and every gate it guarded now runs unconditionally:

- **Conversation input bar / editor**: the model picker, its `/model` slash submenu and the introduction banner are always shown.
- **Agent builder**: `AdvancedSettings` always renders the `ModelPicker`. The legacy dropdown (`AdvancedSettingsLegacy`) and its two now-unused submenus (`ModelSelectionSubmenu`, `ReasoningEffortSubmenu`) are deleted.
- **Model tiers**: `withModelSelectability` and `getModelTierAccessErrorForAgentConfiguration` always enforce the member's tier cap (the `featureFlags` parameter of the latter is dropped, along with the now-unused `getFeatureFlags` call in `create_or_upgrade.ts`).
- **Auto/meta models** (`auto`, `auto_fast`, `auto_complex`) lose their `availableIfOneOf.featureFlag`, so they are available everywhere.
- **Global agents**: model-only global agents are always filtered out of the default listing, and `@dust` always defaults to the highest Auto stream the member's tier cap allows.
- **Public API**: `POST /v1/.../messages` no longer rejects `modelSelection` with a 403; the request is still validated against the workspace's enabled models.
- **Admin pages**: the model-tier settings on the Usage page and the "Model access tiers" block on the Providers page are now gated on workspace admin only.

Note: because `@dust` now always resolves through the Auto stream, `dust_agent_sonnet_5_default` no longer has any effect on it (the branch in `_getDustGlobalAgent` is unreachable). Left in place on purpose — worth a separate cleanup PR.

## Tests

Updated the suites that set the flag or asserted flag-off behavior; ran locally:

- `front`: `model_tiers`, `lib/assistant.test`, `lib/api/assistant/models.test`, `global_agents.test`, `assistant_model_availability`, `EnumSelect` — 116 passed.
- `front-api`: `messages/index`, `me/triggers` — 10 passed.

Two expectations changed because tier caps are now always enforced:

- `enabled_models.test.ts`: the Premium stream now really falls back to its Basic-tier floor (Sonnet 4.6 at `light`) for a `cost_efficient`-capped member, instead of resolving to Opus 5 as it did while the flag-off short circuit made every model selectable.
- `global_agents.test.ts`: `@dust` defaults to the Auto stream rather than GPT 5.6 Luna / Sonnet 5.

## Risk

Workspaces that did not have the flag see behavior changes: the model picker appears, model-tier caps start being enforced on agent runs and agent creation, and `@dust` switches to the Auto stream. Rollback is a straight revert.


Fraggle (approved):
Nice !

PR #31212, opened 2026-08-26
Skills: use resource when creating suggestions

fabiencelier:
## Description

related to https://github.com/dust-tt/tasks/issues/10052

Using models is too brittle, resource handle all the permissions and underlying creations.

## Tests

tested locally

## Risk

low

## Deploy Plan

none


aubin-tchoi (approved):
LGTM

PR #31213, opened 2026-08-26
chore: redline dev_mcp_actions flag

adrsimon:
## Description

`primitive_types_debugger` and `jit_testing` were scaffolding for trying out tool input shapes. They are `dust_only`, nobody uses them, so this drops both servers and the `dev_mcp_actions` flag that gated them.

Their ids (1004, 1016) are added to `LEGACY_INTERNAL_MCP_SERVER_IDS` so old action output whose sId decodes to one of them still resolves, and so no future server inherits the id along with stale views.

Several tests used `primitive_types_debugger` purely as a stand-in for "a flag-gated internal server" and mocked its config anyway. Those now use `http_client` + `http_client_tool`, which has the same shape.

## Tests

`constants`, `mcp_servers_metadata`, `input_configuration` and `mcp_server_view_resource` in front, plus the three `front-api` mcp route suites, all pass.

## Risk

Low. Any workspace that still has a view for one of these servers loses access to it, which is the point.

## Deploy Plan

Deploy front.


PopDaph (approved):
🚀

PR #31218, opened 2026-08-26
[front] Analytics modal fills the viewport height

PopDaph:
The user menu dialogs (Analytics, Tools, Automations) had a fixed `height="xl"` (768px): small on tall screens, with unnecessary inner scrolling.

They now use `height="xl" grow` (#31240): 768px stays the minimum, and the dialog grows with its content up to the dialog's built-in 90vh cap, so they keep their usual size when content is short and use the available screen when it is not.


aubin-tchoi (approved):
LGTM

PR #31219, opened 2026-08-26
Add bounded conversation checkpoint hydration

flvndvd:
## Description

This PR adds the bounded hydration path needed to restore a stateful conversation window from a checkpoint.

Instead of loading and rendering the conversation history, `getConversationForCheckpoint` fetches the exact user and agent messages involved in the active run. It restores agent state through `targetStep - 1`, hydrates tool outputs
only for that latest completed step, and keeps earlier tool actions without their output bodies.

The implementation also resolves the latest canonical step-content versions before loading actions. This prevents artifacts from superseded Temporal attempts from reappearing in model context.

The existing full conversation path keeps the same behavior and query scope. The shared agent-message renderer now accepts either full or checkpoint-bounded hydration, with the checkpoint policy isolated so we don't have endless parametrized functions. 

This PR does not enable checkpoint restoration in the agent loop yet. It provides the focused loading and rendering foundation for that integration.


front/lib/api/assistant/messages.ts
```diff
@@ -312,6 +313,62 @@ async function batchRenderUserMessages(
...
+    textContentOnly,
+    viewType,
+  }: {
+    agentMessageModelIds: ModelId[];
+    messagesWithToolOutputContent: Set<ModelId> | null;
+    textContentOnly: boolean;
+    viewType: RenderMessageVariant;
+  }
+): Promise<AgentMessageContentHydration> {
+  const stepContents = await AgentStepContentResource.fetchByAgentMessages(
```
flvndvd, line 340: Code was just moved around.

front/lib/api/assistant/conversation_rendering/checkpoint_message_hydration.ts
```diff
@@ -0,0 +1,87 @@
+import type { Authenticator } from "@app/lib/auth";
+import { AgentMCPActionResource } from "@app/lib/resources/agent_mcp_action_resource";
+import { AgentStepContentResource } from "@app/lib/resources/agent_step_content_resource";
+import type { ModelId } from "@app/types/shared/model_id";
+
+async function fetchCheckpointStepContents(
```
flvndvd, line 6: This is all new code to fetch only one step. Complexity comes from the versionning.

front/lib/api/assistant/conversation_rendering/checkpoint_conversation.ts
```diff
@@ -0,0 +1,62 @@
...
+import { ConversationError } from "@app/types/assistant/conversation";
+import type { Result } from "@app/types/shared/result";
+import { Err, Ok } from "@app/types/shared/result";
+
+/**
+ * Fetches the bounded conversation projection used to restore or extend an agent-loop checkpoint.
+ * It loads the requested user and agent messages directly, includes agent state completed before
+ * `targetStep`, and hydrates tool output content only for the latest completed step.
+ */
+export async function getConversationForCheckpoint(
```
flvndvd, line 14: Let's call it conviction-building implementation. Obviously if this becomes the main path in prod tomorrow, we can do way way leaner by dropping reactions, unused metadata, and cut a few more SQL calls.

front/lib/resources/agent_step_content_resource.ts
```diff
@@ -424,6 +427,48 @@ export class AgentStepContentResource extends BaseResource<AgentStepContentModel
...
+      step: number;
+      textContentOnly?: boolean;
+    }
+  ): Promise<AgentStepContentResource[]> {
+    return this.fetchByAgentMessagesFromPostgres(auth, {
+      agentMessageIds: agentMessageModelIds,
+      step,
+      textContentOnly,
+    });
+  }
```
adrsimon, line 450: i don't really see the point of this method ? it's just wrapping the same exact call am i right ?

front/lib/api/assistant/conversation_rendering/checkpoint_message_hydration.ts
```diff
@@ -0,0 +1,87 @@
...
+
+  const [stepContents, actions] = await Promise.all([
+    fetchCheckpointStepContents(auth, {
+      agentMessageModelId,
+      latestCompletedStep,
+    }),
+    AgentMCPActionResource.fetchVisibleByLatestStepContents(auth, [
+      agentMessageModelId,
+    ]),
+  ]);
```
id13: nit: let's avoid Promise.all 🙏

adrsimon, line 66: i was scared of this promise all but looks like sum of max concurrency is 8, fine i guess

adrsimon (approved):

front/lib/resources/agent_step_content_resource.ts
flvndvd: It's indeed a shallow wrapper for now, but will let me iterate on the fetch so we only fetch what's needed for the new agent loop.

PR #31220, opened 2026-08-26
[front] Make SpaceResource.isOpen async + check the global group's read grant

tdraier:
## Description

Makes `SpaceResource.isOpen()` **async** and resolves openness from `group_permissions` with the **correct predicate**: a space is open iff the workspace **global group** holds a `reader` grant on it (previously "any reader grant" — provably equivalent for current data, but no longer brittle). This removes the last serialization/permission reader of the eagerly-loaded `this.groups`, a step toward dropping the eager `spaceGrants` include (remaining readers handled in the follow-up PR).

- `isOpen(auth): Promise<boolean>` is async; new `SpaceResource.listOpenSpaceModelIds(auth, spaces)` resolves openness for a whole batch in **one** grant query so list/loop callers don't become N+1.
- Classifiers `isRegularAndRestricted` / `isProjectAndRestricted` / `isRegularAndOpen` / `isRestricted` are now `async(auth)`; single-space callers `await`, loop sites (conversation permissions, projects listing, skill space-requirements, search sort, the two open-space migrations) precompute the open-id set.
- **Global group id resolved on auth:** `dangerouslyListUserGroupsForAuth` now returns `{ globalGroupModelId, groupModelIds }` (the id was already selected by the query, just discarded), cached on the `Authenticator` and read by `auth.getGlobalGroupModelId()` — so openness checks drop the per-check global-group lookup. Tri-state: `undefined` = unresolved (lazy fetch for system/internal auths), `ModelId` = resolved, `null` = non-member / no-workspace auth. Serialized in `toJSON`/`fromJSON`; cache key bumped to `groups:v2:` since the cached value shape changed.
- **`space.accessed` audit off the critical path:** in `withSpace`, the `isRestricted` check + emit run in a detached async closure, so the request never blocks on them.
- Drive-by: `sandbox/egress-policy/pods.ts` still called the now-async `isProjectAndRestricted()` synchronously — fixed to batch via `listOpenSpaceModelIds`.

## Tests

- `tsgo --noEmit` clean in both `front` and `front-api`; no remaining sync `isOpen()`/classifier call sites.
- Updated existing tests (`group_resource` auth-groups shape + `groups:v2:` cache key). Passing: `group_resource` + `space_resource` (77) + `spaces` api (41) in front; front-api spaces list/detail, members, project_metadata, pods egress-policy, access-check.

## Risk

- **Low.** The openness predicate is provably equivalent to the prior behavior for current data, and covered by `space_resource`/`spaces` tests.
- **Cache key bump (`groups:` → `groups:v2:`):** one-time cold cache for user group lookups after deploy (each user re-fetches once). The cache has no TTL, so old `groups:` entries orphan until Redis LFU eviction — harmless. Rollback-safe: old code reads the old key, which still holds valid entries.
- **Serialization:** Temporal payloads in flight across the deploy lack `globalGroupModelId` — it's optional, so those auths resolve it lazily. No stale data (a workspace's global group id never changes).
- **Audit fire-and-forget:** `space.accessed` was already `void` (best-effort); it's now also non-blocking. A failure is logged, not thrown.

## Deploy Plan

- No migration, no ordering constraints — standard deploy of `front` + `front-api`.
- Safe to roll back at any point (cache keys diverge cleanly in both directions; no schema change).

### DIFF
--- a/front-api/middlewares/with_space.ts
+++ b/front-api/middlewares/with_space.ts
@@ -5,6 +5,7 @@ import {
 } from "@app/lib/api/audit/workos_audit";
 import type { Authenticator } from "@app/lib/auth";
 import { SpaceResource } from "@app/lib/resources/space_resource";
+import logger from "@app/logger/logger";
 import type { SpaceCtx } from "@front-api/middlewares/ctx";
 import { apiError } from "@front-api/middlewares/utils";
 import { createMiddleware } from "hono/factory";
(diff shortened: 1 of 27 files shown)


tdraier:

- #31220 ◀
    - #31242

fabiencelier (approved):

front-api/middlewares/with_space.ts
```diff
@@ -89,23 +90,41 @@ export function withSpace(options: WithSpaceOptions) {
...
+            buildAuditLogTarget("space", space),
+          ],
+          context: getAuditLogContext(auth),
+          metadata: {
+            space_name: space.name,
+            space_kind: space.kind,
+            is_restricted: "true",
+            access_method: deriveAccessMethod(auth),
+          },
+        });
```
tdraier, line 116: @zmarouf @smb2268 do we really need to have this audit log ? and if we do, why only for restricted ? just wondering as `space.isOpen(auth)` is not free anymore.

smb2268: I'd assume yes we do - important to big customers to have a record of who is accessing what information. Probably better to have it for both restricted & not and actually use that `is_restricted` prop, but I'd guess the reasoning was it's less necessary for public information?

PR #31221, opened 2026-08-26
feat(models): flag degraded models in the model picker

Nils-Fedrigo:
## Description
Screenshot 2026-08-26 at 18 50 33

Surfaces degraded models (the ones an operator marks from the Poke Kill page) in the model picker, so a user who picks one knows what they are opting into. Degraded models stay selectable: only `auto` streams route around them, a definitive pick is never rerouted.

Two indicators, both explained by the same tooltip ("This model is currently degraded following an incident on the provider side. Answers may be slower or fail — consider picking another model."):

- the picker's trigger button badges its icon in the top-right corner when the selected model is degraded (tiers never badge: streams already skip degraded candidates);
- a degraded model row shows the badge glyph where a locked row shows its lock, and stays clickable.

`GET /api/w/:wId/models` now returns `degradedModelIds`, scoped to the models the workspace can see so we don't leak ids it has no access to.

The badge comes from `DoubleIcon`, extended for it rather than hand-rolled in `front`:

- `position` (`bottom-right` default, `top-right`), which hangs the badge a quarter of its width off the corner so it doesn't bury a small main icon;
- `secondaryColor` (`info` / `warning` / `success` / `highlight`), filling the badge and knocking the glyph out in white — the fill sits inset under the glyph's own ring, which would otherwise leave a sliver of fill showing around it;
- a `xs` size for badging a 16px button icon, which needed a `2xs` (12px) step on the `Icon` scale.

`sm` / `md` / `lg` / `xl` and `bottom-right` map exactly as before, so the attachment chips and dropdowns already using `DoubleIcon` are unchanged.

## Tests

New Storybook stories cover the badge (`DoubleIcon` → Status Badge, Status Badges); the existing `DoubleIcon`, `Icon`, `AttachmentChip` and `Dropdown` story tests pass, confirming existing consumers are unaffected.

To check by hand: mark a model degraded from Poke → Kill page, then in the input bar pick that model from the picker — the row and the trigger button should both carry the badge, and the model should still send. A tier pick ("Basic" / "Standard" / "Premium") should never badge.

Note the picker's view of the degraded set can lag an operator's change by up to a minute: the server refreshes its cache every 60s and the models response is cached by SWR.

### DIFF
--- a/front/lib/swr/models.ts
+++ b/front/lib/swr/models.ts
@@ -10,6 +10,12 @@ import type { Fetcher } from "swr";
 
 const EMPTY_DEGRADED_MODEL_IDS: ReadonlySet<string> = new Set();
 
+// The catalog itself barely moves, but the degraded models it reports do: an
+// operator flagging a model mid-incident must reach tabs that stay open for
+// hours without a reload. Polling pauses while the tab is hidden, and focus
+// revalidation (throttled to the same cadence) covers coming back to it.
+const MODELS_REFRESH_INTERVAL_MS = 5 * 60 * 1000;
+
 export function useModels({
   owner,
   disabled,
@@ -23,7 +29,11 @@ export function useModels({
   const { data, error } = useSWRWithDefaults(
     `/api/w/${owner.sId}/models`,
     modelsFetcher,
-    { disabled, revalidateOnFocus: false }
+    {
+      disabled,
+      refreshInterval: MODELS_REFRESH_INTERVAL_MS,
+      focusThrottleInterval: MODELS_REFRESH_INTERVAL_MS,
+    }
   );
 
   const degradedModelIds = useMemo(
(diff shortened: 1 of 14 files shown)


front/lib/swr/models.ts
```diff
@@ -23,12 +26,18 @@ export function useModels({
     { disabled, revalidateOnFocus: false }
```
frankaloia: Does this mean not refreshed so degraded models will never be updated?

front/lib/swr/models.ts
```diff
@@ -5,8 +5,17 @@ import type {
...
 import type { LightWorkspaceType } from "@app/types/user";
+import { useMemo } from "react";
 import type { Fetcher } from "swr";
 
+const EMPTY_DEGRADED_MODEL_IDS: ReadonlySet<string> = new Set();
+
+// The catalog itself barely moves, but the degraded models it reports do: an
+// operator flagging a model mid-incident must reach tabs that stay open for
+// hours without a reload. Polling pauses while the tab is hidden, and focus
+// revalidation (throttled to the same cadence) covers coming back to it.
```
Nils-Fedrigo, line 16:
```suggestion
```

frankaloia (approved):
LGTM, will trust you will work with design for any UX tightening

PR #31232, opened 2026-08-26
[front] fix: keep the impersonated user's role on system-key auth exchange

davidebbo:
Fixes https://github.com/dust-tt/tasks/issues/10198

Issue: parent agent can't call child agent that requires admin user, even if user is admin.

### Root cause

`Authenticator.exchangeSystemKeyForUserAuthByEmail` hardcoded `role: "user"` when rebuilding the authenticator for the impersonated user, regardless of that user's real workspace role. Every conversation created through system-key + `x-api-user-email` impersonation therefore ran as a plain member.

### Fix

Use `activeMembership.role` — the role of the already-fetched, verified active membership.

The role comes from the database, never from a caller-supplied header, and a system key already defaults to `admin` in `Authenticator.fromKey`, so the exchanged authenticator can only ever end up *lower*-privileged than the key it was exchanged from.

### Tests

Two functional tests on `POST /api/v1/w/:wId/assistant/conversations/:cId/messages` with a system key and `x-api-user-email`.

### DIFF
--- a/front/lib/auth.ts
+++ b/front/lib/auth.ts
@@ -1198,8 +1198,11 @@ export class Authenticator {
     return new Authenticator({
       authMethod: auth.authMethod(),
       key: auth._key,
-      // We limit scope to a user role.
-      role: "user",
+      // Scope down to the impersonated user's own workspace role: the system key is admin by
+      // default, and the resulting authenticator must not grant more than the user has. The role
+      // comes from the verified active membership, never from the caller, and the groups below
+      // are already the user's own.
+      role: activeMembership.role,
       groupModelIds,
       user,
       subscription: auth._subscription,
--- a/front-api/routes/v1/w/[wId]/assistant/conversations/[cId]/messages/index.test.ts
+++ b/front-api/routes/v1/w/[wId]/assistant/conversations/[cId]/messages/index.test.ts
@@ -9,6 +9,11 @@ import { CLAUDE_OPUS_4_8_MODEL_ID } from "@app/types/assistant/models/anthropic"
 import { honoApp } from "@front-api/app";
 import { describe, expect, it, vi } from "vitest";
 
+vi.mock("@app/temporal/agent_loop/client", () => ({
+  launchAgentLoopWorkflow: vi.fn(),
+  launchCompactionWorkflow: vi.fn(),
+}));
+
 vi.mock("@app/lib/api/programmatic_usage/tracking", () => ({
   isProgrammaticUsage: () => false,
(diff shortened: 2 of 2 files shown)


front/lib/auth.ts
```diff
@@ -1198,8 +1198,11 @@ export class Authenticator {
     return new Authenticator({
       authMethod: auth.authMethod(),
       key: auth._key,
-      // We limit scope to a user role.
-      role: "user",
+      // Scope down to the impersonated user's own workspace role: the system key is admin by
+      // default, and the resulting authenticator must not grant more than the user has. The role
+      // comes from the verified active membership, never from the caller, and the groups below
+      // are already the user's own.
+      role: activeMembership.role,
```
flvndvd, line 1205: I recall downgrading the role to user to avoid any admin operations being allowed once the key was exchanged.

flvndvd (approved):

front/lib/auth.ts
davidebbo: But that only happens if the user is in fact admin. Still, I'm fuzzy on the implications.

philipperolet: I assume the risk would be an attack spoofing the email and getting admin rights via this; I'm a noob at security but AFAICT it's passed by http header; changing a header is not unsurmountable?

flvndvd: Yeah the way to change it is system key + email header. Assumption relies on system keys never accessed. @davidebbo do we know where it breaks in the call to analyst? Is it the MCP tool or just the agent is not available? Cause maybe, one alternative approach is to refetch an auth for this one use case?

davidebbo: This is what I get on a local repro:

```
[agent-loop] [15:22:17.228] WARN (5637): Tool execution error {"workspace":{"sId":"BHWrN3HaUa","plan_code":"FREE_UPGRADED_PLAN"},"toolName":"run_agent","actionConfigurationId":"8slYwSnp45","agentConfigurationId":"cE1hmRCRpo","agentConfigurationVersion":1,"agentMessageId":"FUD5rbWQfp","conversationId":"93ucWn4I9s","duration":64.61008300000685}
[agent-loop]     error: {
[agent-loop]       "type": "Object",
[agent-loop]       "message": "Agent @analyst is not available to the user running this conversation. Ask a workspace admin to grant access to the agent and its spaces.",
[agent-loop]       "stack":
[agent-loop]           
[agent-loop]     }
```

davidebbo:
Didn't mean to merge this, thought I was on different PR! Reverting for now.

PR #31235, opened 2026-08-26
[front] chore: remove unused conversation tools endpoint

aubin-tchoi:
## Description

This PR removes the undocumented system-key-only conversation tools endpoint. It's no longer used, its only caller was the toolsets enable tool, which was moved to not rely on the public API. Also not used by connectors.

What we're doing here is very specific to the fact that the endpoint is restricted to system keys, which typically have no reason to exist except for connectors.

## Risk

- Low.

## Deploy Plan

- Deploy front.


aubin-tchoi:
no need to update the doc, the endpoint was swaggerignored

adrsimon (approved):

PR #31237, opened 2026-08-27
feat(analytics): add Ask @analyst conversation panel

achilleburah:
## Description

Phase 1: a persisted, billed conversation panel on the Analytics consumption page, locked to @analyst. Adds a reusable ConversationPanel shell to sparkle, and lets AdminLayout pages opt out while the panel is open.

Created in sparkle for easier reuse.

## Tests

Tested locally. 

## Screenshots

image

image

## Risk
_Low risk._

## Deploy Plan

Behind a FF. Classic deploy + enable on dust.
Remove once validated. 

Rollback to revert.


sparkle/src/components/ResizableSidePanel.tsx
```diff
@@ -0,0 +1,133 @@
...
+ * draggable and the panel animates open and shut.
+ *
+ * @example
+ * ```tsx
+ * <ResizableSidePanel isOpen={isOpen} panel={<MyPanel />}>
+ *   <PageContent />
+ * </ResizableSidePanel>
+ * ```
+ */
+export function ResizableSidePanel({
```
PopDaph, line 51: Reusing ConversationSidePanelContainer directly wasn't an option (conversation-coupled, no generic slot, lives in front), so building this in sparkle is the right call.

But it re-implements the same resize/collapse machinery minus a few battle-tested fixes the existing one has: onDragging (a drag released past the threshold completes the collapse), onTransitionEnd (content stays until the animation finishes), and the cqw trick (content keeps its width while animating).

Could you fold those in, and signpost a follow-up to re-base ConversationSidePanelContainer onto this component so we don't keep two forked side-panel patterns? 🙏🏻

front/hooks/useAnalyticsConversationPanel.ts
```diff
@@ -0,0 +1,26 @@
...
+/**
+ * Open/close state for the Analytics conversation panel.
+ *
+ * Deliberately does not touch the navigation sidebar: reclaiming its width is
+ * driven by how wide the panel actually gets, which only the layout can
+ * measure (see AppContentLayout's onContentSqueezed).
+ *
+ * openPanel/closePanel are called both directly (the CTA / close button) and
+ * from the resizable panel's onCollapse lifecycle callback (dragging the
+ * divider shut), so both guard on the current value to stay idempotent.
```
PopDaph: What does this mean? I'm not sure I understand the command

front/hooks/useAnalyticsConversationPanel.ts
```diff
@@ -0,0 +1,26 @@
...
+ */
+export function useAnalyticsConversationPanel() {
+  const [isOpen, setIsOpen] = useState(false);
```
PopDaph: Do we really need a wrapper for one boolean useState?

front/components/workspace/analytics/AnalyticsConversationPanel.tsx
```diff
@@ -0,0 +1,232 @@
...
+        name={agentConfiguration.name}
+        avatarUrl={agentConfiguration.pictureUrl}
+      />
+      <div className="flex flex-col gap-2">
+        <ConversationMessageTitle
+          name={agentConfiguration.name}
+          renderName={(name) => name}
+        />
+        <ConversationMessageContent type="agent">
+          Ask me about workspace usage — top agents, credit spend, or trends
```
PopDaph: NIT: Perso opinion but I hate emdash, suggestion "I can help you understand your workspace usage and costs across your analytics data."

front/components/sparkle/AppContentLayout.tsx
```diff
@@ -134,67 +160,32 @@ export function AppContentLayout({ children }: AppContentLayoutProps) {
...
-                  "flex flex-1 flex-col",
-                  isMobile
-                    ? MOBILE_DOCUMENT_SCROLL_CLASSES.contentArea
-                    : "min-h-0 overflow-y-auto [scrollbar-gutter:stable]"
-                )}
-              >
-                {contentWidth ? (
-                  <>
-                    {title}
+            <ResizableSidePanel
```
PopDaph: The layout refactor wraps every page's content in ResizablePanelGroup, FF or not, so "Low risk" undersells it slightly. Worth a quick sanity pass on a couple of unrelated pages (conversation, builder) and a rollback note.

PopDaph (commented):
I did a first pass but I need to dig deeper to have a real opinion on the layout change, BRB for a second pass 🙏🏻

sparkle/src/components/ResizableSidePanel.tsx
achilleburah: Definitely !

front/hooks/useAnalyticsConversationPanel.ts
achilleburah: probably not

front/components/workspace/analytics/AnalyticsConversationPanel.tsx
achilleburah: feels more natural 👍

front/components/sparkle/AppContentLayout.tsx
PopDaph: Correcting myself: since ResizableSidePanel returns bare children when no panel is registered, other pages render exactly as before, so this is safer than I first thought. 😌

front/hooks/useAnalyticsConversationPanel.ts
achilleburah: It's a remainder of the first draft of the opening/close modal but i ended up simplifying it so this part is stale. 
The useCallbacks are not really necessary anymore either. 
I'll remove them and inline a simple useState

front/components/sparkle/AppContentLayout.tsx
achilleburah: That's true, but the ResizablePanel returns early, <>{children}</>. So if panel is not set, it shouldnt move. 
I'll still add a note !

PopDaph:
Adding @ykmsd as reviewer specifically for the layout part since she did a big rework/cleanup of our layout a couple of months ago 🙏🏻

front/components/workspace/analytics/AnalyticsConversationPanel.tsx
```diff
@@ -0,0 +1,232 @@
...
+      <Button variant="ghost" size="sm" icon={XClose} onClick={onClose} />
+    </div>
+  );
+}
+
+interface AnalyticsConversationGreetingProps {
+  agentConfiguration: LightAgentConfigurationType;
+}
+
+function AnalyticsConversationGreeting({
```
ykmsd, line 61: We cannot pass static data to ConversationViewer? because this design is not the same as the conversation viewer so the content shift when you send a message, which is not really nice

So either
- we pass static initial message data and use ConversationViewer all the time if possible
- if we cannot, try to match the design of agent message and composer layout with ConversationViewer

https://github.com/user-attachments/assets/bb3af3ea-0ed1-4ed4-9a44-7d2418cc2fd3

front/components/workspace/analytics/AnalyticsConversationPanel.tsx
```diff
@@ -0,0 +1,232 @@
...
+              <AnalyticsConversationGreeting
+                agentConfiguration={analystAgentConfiguration}
+              />
+            </div>
+          )}
+        </div>
+
+        {!conversation && (
+          <div className="w-full flex-shrink-0 p-2">
+            <InputBar
```
ykmsd, line 171: I don't think you want to mention humans neither here or is it intentional? 

Image

front/components/workspace/analytics/AnalyticsConversationPanel.tsx
```diff
@@ -0,0 +1,232 @@
...
+              owner={owner}
+              user={user}
+              onSubmit={createConversation}
+              stickyMentions={stickyMentions}
+              draftKey="analytics-conversation-panel"
+              actions={["attachment"]}
```
ykmsd: I have zero context but do we really need attachment action? 🤔

sparkle/src/components/ResizableSidePanel.tsx
```diff
@@ -0,0 +1,133 @@
...
+          onCollapse={onCollapse}
+          onResize={reportIfContentSqueezed}
+          className={cn(
+            "overflow-hidden",
+            isOpen
+              ? "absolute inset-0 z-50 md:relative md:inset-auto md:z-auto"
+              : "hidden md:block"
+          )}
+        >
+          <div className="h-full w-full overflow-hidden bg-background">
```
ykmsd: I think it's coming from this background color but it doesn't work with some of the colors conversation view uses e.g.

Image

I think it's safer to use the same background color for the normal conversation view no?

front/components/workspace/analytics/AnalyticsConversationPanel.tsx
```diff
@@ -0,0 +1,232 @@
...
+  Spinner,
+  XClose,
+} from "@dust-tt/sparkle";
+import { useMemo } from "react";
+
+interface AnalyticsConversationPanelHeaderProps {
+  onClose: () => void;
+}
+
+function AnalyticsConversationPanelHeader({
```
ykmsd, line 37: I don't have a solution but having two close button doesn't look nice, you might want to catch one of designers to improve it 😄 

Image

front/components/workspace/analytics/AnalyticsConversationPanel.tsx
```diff
@@ -0,0 +1,232 @@
...
+
+  const stickyMentions = useMemo<RichMention[]>(
+    () =>
+      analystAgentConfiguration
+        ? [toRichAgentMentionType(analystAgentConfiguration)]
+        : [],
+    [analystAgentConfiguration]
+  );
+
+  const agentBuilderContext = useMemo<
```
ykmsd: why agent builder context? 🤔

sparkle/src/stories/ResizableSidePanel.stories.tsx
```diff
@@ -0,0 +1,135 @@
...
+    </div>
+  ),
+};
+
+/**
+ * Toggling the panel while it stays mounted, so its content keeps state.
+ *
+ * @summary Open and close without unmounting.
+ */
+export const Toggleable: Story = {
```
ykmsd: this is also unused?

front/components/sparkle/AppContentLayout.tsx
```diff
@@ -94,6 +98,28 @@ export function AppContentLayout({ children }: AppContentLayoutProps) {
 
   useDocumentScrollMode(isMobile);
 
+  const didFoldNavigationForPanelRef = useRef(false);
+
+  const foldNavigationForPanel = useCallback(() => {
```
ykmsd: Is this to minimize the sidebar when the analytics conversation viewer is bigger?

front/components/sparkle/AppContentLayout.tsx
ykmsd: I don't think we want to users shrink the panel to 0 so we should set the minSize no? (If you shrink it to 0 it's hard to know how to reopen it)

https://github.com/user-attachments/assets/380ac721-88c3-4c34-8520-83407afa85ea

front/components/sparkle/AppContentLayout.tsx
ykmsd: Thought about a bit more, I'm not a big fan of having all this logic & component inside the AppContentLayout. I agree that it would be nice to have a unified way to control the panel from here but this requiers a bit more thoughts on how it should behave (like do we always want to minimize the side menu in any context? Will it be the same threshold width etc).  And I think some of them have specific logic to their own place so I'm not fully sure we can reuse for ever scenario. 

Any other strong reason to have it here? Can't we have it somewhere local to this context like AnalyticsConsumptionPage for now?

ykmsd (commented):
Put a bunch of random comments, let me know if you need any clarification or help 🙏

Also not related to this PR but I think you want to remove these analytics conversation from the main conversation list view, or have different title maybe? 🤔

Image

front/components/sparkle/AppContentLayout.tsx
PopDaph: I think it's even impossible to reopen it as the button is gone 😅

ykmsd: you can actually if you try to drag from the edge of the screen 😄 (if you know)

PopDaph: I tried but it dragged the window 😬

achilleburah:
> Also not related to this PR but I think you want to remove these analytics conversation from the main conversation list view, or have different title maybe? 🤔

Keeping the conversation was on purpose, especially to go back to previous analysis you might have asked. But definitely could improve the title, maybe including the theme and/or the date.

front/components/workspace/analytics/AnalyticsConversationPanel.tsx
achilleburah: Oh yeah definitely, i'll disable the mentions

front/components/sparkle/AppContentLayout.tsx
achilleburah: > Is this to minimize the sidebar when the analytics conversation viewer is bigger?

Yes it is. I iterated a little bit on this one. On smaller screens closing the side panel allows for more rooms. Especially since between Analytics view + analyst agents sending back it's own charts, I felt like we might need the space. 
Discussed this a bit with Clément, the idea was auto-toggle depending on the "remaining space/margins available" 


> Any other strong reason to have it here?

No, it was the main point. I understand we want most of our manual actions to be actionable by an agents. So I was thinking that it could be nice to easily spin up a new minimal lateral convo, dedicated to a specific agent in different places. 
But it's a pretty fair pushback, maybe it's a bit early to anticipate this and we could at least wait to have a second usecase in mind ?

front/components/sparkle/AppContentLayout.tsx
achilleburah: Oh definitely not ! That was a bug not a feature.

achilleburah: Well we could leave the slide to 0 to close, but then it should toggle the visible state, so the 'Ask Analyst" button comes back. 
But not sure that's necessary

front/components/workspace/analytics/AnalyticsConversationPanel.tsx
achilleburah: It's the one that holds the disableAgentMentions (used in Sidekick) to lock the analyst agent in. 
Refactoring it out would be cleaner but also bigger. Maybe in following PR ?

front/components/workspace/analytics/AnalyticsConversationPanel.tsx
achilleburah: i put a closing panel icon for now, will check with design

front/components/sparkle/AppContentLayout.tsx
achilleburah: As discussed IRL, moving this into the analytics page

front/components/sparkle/AppContentLayout.tsx
achilleburah: Ok trying with shrinking to 0 -> collapse. Now the button allow to toggle it back. Open to iterate on this one.

front/components/workspace/analytics/AnalyticsConversationPanel.tsx
achilleburah: improved a little, but better empty state screen coming up in following PR with the conversation picker.

front/components/layouts/AdminLayout.tsx
```diff
@@ -38,11 +39,21 @@ export function AdminLayout({ children }: AdminLayoutProps) {
...
   useSetSubNavigation(subNavigation);
 
+  // TODO(2026-09-02 achilleburah): Move these layout classes into the admin
+  // pages so this component only wires the sub navigation.
   return (
     <div
-      className={cn("flex h-full w-full flex-col items-center pt-4 sm:pt-8")}
+      className={cn(
+        "flex h-full w-full flex-col items-center",
+        !fullWidth && "pt-4 sm:pt-8"
```
ykmsd: can you help me to understand what we want to do with this fullWidth prop? It feels weird to me that fullWidth prop control the vertical spacing too.

ykmsd: I see this `relative flex w-full max-w-6xl grow flex-col px-4 pb-4 pt-4 sm:px-10 sm:pb-8 sm:pt-8 ` in AnalyticsConsumptionPage.tsx so assuming this is to define the container style there but this is fragile if we tweak the style in AdminLayout we need to modify there too. What was the reason we couldn't define it here?

front/components/pages/workspace/AnalyticsConsumptionPage.tsx
```diff
@@ -172,6 +192,51 @@ export function AnalyticsConsumptionPage() {
...
+  useEffect(() => {
+    if (isNavigationBarOpen) {
+      didFoldNavigationForPanelRef.current = false;
+    }
+  }, [isNavigationBarOpen]);
+
+  const content = (
+    <div className="relative flex w-full max-w-6xl grow flex-col px-4 pb-4 pt-4 sm:px-10 sm:pb-8 sm:pt-8">
+      {analyticsAssistantEnabled && !isOpen && (
+        <div className="absolute right-4 top-4 z-10 sm:right-10 sm:top-8">
```
ykmsd: do we need div here? I think we can apply these styles directly to Button no?

front/components/pages/workspace/AnalyticsConsumptionPage.tsx
```diff
@@ -172,6 +192,51 @@ export function AnalyticsConsumptionPage() {
...
+    }
+  }, [isOpen, setIsNavigationBarOpen]);
+
+  useEffect(() => {
+    if (isNavigationBarOpen) {
+      didFoldNavigationForPanelRef.current = false;
+    }
+  }, [isNavigationBarOpen]);
+
+  const content = (
```
ykmsd: `content` doesn't seem to be reused why we cannot inline it 🤔? 

also the content is wrapped with another container look like div do we really need both divs?

```
<div className="flex h-full w-full flex-col items-center overflow-y-auto">
            {content}
</div>
```

front/components/pages/workspace/AnalyticsConsumptionPage.tsx
```diff
@@ -186,12 +251,29 @@ export function AnalyticsConsumptionPage() {
...
-        onAgentClick={setAgentDetailsId}
-        onSkillClick={setSkillDetailsId}
-      />
+      {analyticsAssistantEnabled ? (
+        <ResizableSidePanel
+          isOpen={isOpen}
+          onCollapse={() => setIsOpen(false)}
+          minContentWidthPx={MIN_CONTENT_WIDTH_WITH_PANEL_PX}
+          onContentSqueezed={foldNavigationForPanel}
+          className="min-h-0 flex-1"
```
ykmsd, line 261: what do we need this for? is it specific to this component?

sparkle/src/components/ConversationPanel.tsx
```diff
@@ -0,0 +1,43 @@
...
+ * that fills the remaining space.
+ *
+ * @example
+ * ```tsx
+ * <ConversationPanel header={<MyPanelHeader onClose={onClose} />}>
+ *   <ConversationViewer ... />
+ * </ConversationPanel>
+ * ```
+ */
+export function ConversationPanel({
```
ykmsd: Do we really need to put this in sparkle? Is the plan to replace the actual conversation panel with this too?

front/components/pages/workspace/AnalyticsConsumptionPage.tsx
```diff
@@ -172,6 +192,51 @@ export function AnalyticsConsumptionPage() {
...
+      setIsNavigationBarOpen(false);
+    }
+  }, [isNavigationBarOpen, setIsNavigationBarOpen]);
+
+  useEffect(() => {
+    if (!isOpen && didFoldNavigationForPanelRef.current) {
+      didFoldNavigationForPanelRef.current = false;
+      setIsNavigationBarOpen(true);
+    }
+  }, [isOpen, setIsNavigationBarOpen]);
```
ykmsd: do we really have to use useEffect here? Can't we do this inside the event handler when we close the side panel?

sparkle/src/components/ResizableSidePanel.tsx
```diff
@@ -0,0 +1,163 @@
...
+      return;
+    }
+    if (isOpen) {
+      panelRef.current.expand(defaultSize);
+    } else {
+      panelRef.current.collapse();
+    }
+  }, [isOpen, defaultSize]);
+
+  const handlePanelResize = React.useCallback(
```
ykmsd, line 78: seems like we don't need useCallback here?

front/components/pages/workspace/AnalyticsConsumptionPage.tsx
achilleburah: yes you're right 👍

front/components/pages/workspace/AnalyticsConsumptionPage.tsx
achilleburah: I moved it inside the component directly

sparkle/src/components/ResizableSidePanel.tsx
achilleburah: true

ykmsd (approved):
Discussed more https://dust4ai.slack.com/archives/C09GELMTTRT/p1788529278807069?thread_ts=1788449229.037559&cid=C09GELMTTRT 

The summary is:
- Let's not put components in sparkle unless it's very obviously reusable. We can put it in sparkle once we confirm that we can reuse it 👍 
- Having styling in AdminLayout is not flexible enough to adopt different design (also it should probably stay as permission check and all and not styling). We can move out the styling to each page. The only risk is it's easy to forget to use this styling component but I don't have any good other solution right now, so let's try with it

front/components/pages/workspace/AnalyticsConsumptionPage.tsx
achilleburah: moved this to the closePanel method 👍

sparkle/src/components/ResizableSidePanel.tsx
achilleburah: Handled in PR#3 of the stack: https://github.com/dust-tt/dust/pull/31842

PR #31240, opened 2026-08-27
[sparkle] Add grow prop to DialogContent

PopDaph:
Fixed dialog heights force a trade-off: too small when the content is tall (inner scroll), wasted space when it is short. This adds a `grow` boolean to `DialogContent`: with a fixed `height`, the dialog can grow beyond it when the content needs it, up to the existing 90vh cap. The height becomes a minimum. Without `height` it is a no-op (growing is already the default behavior).

Implementation notes:
- When `grow` is set the `h-*` class is not emitted; a min-height floor is applied instead (rather than relying on tailwind-merge, which does not know the custom `h-md..h-2xl` utilities).
- `min-height` wins over `max-height` in CSS, so the floor is itself capped: `min-h-[min(768px,90vh)]`, otherwise the dialog would overflow short viewports.

Adds a `GrowingDialog` story (toggleable content size to see min / grown / capped states).

First consumers: the user menu Analytics/Tools/Automations dialogs (#31218), which currently inline `sm:min-h-[min(768px,90vh)]` and will become `height="xl" grow`.

## Risk
No class output changes unless `grow` is passed. Safe to rollback.

### DIFF
--- a/sparkle/src/components/Dialog.tsx
+++ b/sparkle/src/components/Dialog.tsx
@@ -65,13 +65,14 @@ const heightClasses: Record<DialogHeightType, string> = {
 };
 
 // With grow, the fixed height becomes a minimum: the h-* class is not
-// emitted, min-h keeps the height's size as the floor, and the base
-// max-h-[90vh] still caps the growth.
+// emitted and min-h keeps the height's size as the floor. min-height wins
+// over max-height in CSS, so the floor is itself capped at 90vh to never
+// exceed the base max-h-[90vh] on short viewports.
 const growHeightClasses: Record<DialogHeightType, string> = {
-  md: "sm:min-h-md",
-  lg: "sm:min-h-lg",
-  xl: "sm:min-h-xl",
-  "2xl": "sm:min-h-2xl",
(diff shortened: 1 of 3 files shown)


PopDaph:
cc @ykmsd happy to apply feedback when you're back if you have any 🫡

sparkle/src/components/Dialog.tsx
```diff
@@ -64,6 +64,16 @@ const heightClasses: Record<DialogHeightType, string> = {
   "2xl": "sm:h-2xl",
 };
 
+// With grow, the fixed height becomes a minimum: the h-* class is not
+// emitted, min-h keeps the height's size as the floor, and the base
+// max-h-[90vh] still caps the growth.
```
aubin-tchoi: are we sure that's true? thought the min won if min > max

aubin-tchoi (approved):
LGTM

sparkle/src/components/Dialog.tsx
PopDaph: I think you're right! Will add a commit! Thanks 🙏🏻

aubin-tchoi (approved):

PR #31242, opened 2026-08-27
[front] Drop SpaceResource.groups field + eager spaceGrants include

tdraier:
## What

Now that `isOpen()` no longer reads `this.groups` (#31220), the only remaining readers were the group-management helpers. This loads their grants on demand and **drops the eagerly-loaded `groups` field and the `spaceGrants` include entirely** — so space loads (and every space-scoped resource load via `resource_with_space`) no longer join `group_permissions`.

- New public `SpaceResource.fetchGrantReferences()` (per space) and static `listGrantReferencesBySpaceModelId(spaces)` (batched, one query); `SpaceGroupReference` is now exported.
- `fetchGroupResources`' default, `fetchMembershipGroups`, and the batched `fetchDistinctActiveManualGroupMembersBySpaces` load references on demand (the batched method uses the batched loader — no N+1).
- Removed the `groups` constructor field, the `fromModel`/`makeNew` grant re-read, and the eager `spaceGrants` include from all three fetch paths: `baseFetch`, the cross-workspace `fetchByModelIds`, and `resource_with_space`'s `baseFetchWithAuthorization`.
- Updated the two migrations that read `space.groups`.

## Why

Completes the series (`#31053` batched membership → `#31157` groupIds/isRestricted off `toJSON` → `#31220` async `isOpen`): `this.groups` has no readers left, so the field and the per-fetch grant join are gone. Space loads are cheaper across the board.

## Test plan

- `tsgo --noEmit` clean in both `front` and `front-api`; no `this.groups` code readers remain (comments only).
- Passing: `space_resource` (77), `spaces` api (41), `data_source_resource`, conversation `mentions`/`messages` (181 front total); front-api spaces list/detail, members, project_metadata, pods apps, access-check (30).

### DIFF
--- a/front/lib/resources/resource_with_space.ts
+++ b/front/lib/resources/resource_with_space.ts
@@ -2,7 +2,6 @@ import type { Authenticator } from "@app/lib/auth";
 import type { ResourceWithId } from "@app/lib/resources/base_resource";
 import { BaseResource } from "@app/lib/resources/base_resource";
 import { SpaceResource } from "@app/lib/resources/space_resource";
-import { GroupPermissionModel } from "@app/lib/resources/storage/models/group_permissions";
 import { SpaceModel } from "@app/lib/resources/storage/models/spaces";
 import type { WorkspaceModel } from "@app/lib/resources/storage/models/workspace";
 import type {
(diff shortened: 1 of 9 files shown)


tdraier:

- #31220
    - #31242 ◀

front/lib/resources/resource_with_space.ts
```diff
@@ -97,19 +96,6 @@ export abstract class ResourceWithSpace<
...
-          model: GroupPermissionModel,
-          // A where on an include implies required: true;
-          // pass this required: false to keep the original behavior intact
-          required: false,
-          where: {
-            workspaceId: blobWorkspaceIds,
-            resourceType: "space",
-          },
-        },
-      ],
```
fabiencelier, line 112: 🥳

fabiencelier (approved):
LGTM
To be monitored to make sure there is no path where we fetch the permissions too many times in a row
Otherwise we may need to store the fetchGrantReferences result to avoid refetching several times

PR #31130, opened 2026-08-25
[front] Merge tool usage into a single llm_usage_v3 event

tdraier:
## Description

Groundwork for merging tool usage into a single `llm_usage_v3` event, rolled out safely behind a shadow log.

We **keep generating and ingesting the legacy events unchanged** (per-model `llm_usage_v3` + per-tool `tool_use_v3`) and add a **parity log** that computes — without ingesting — what the future aggregated event's cost would be, so we can confirm it matches the current billing on real traffic before flipping the switch. The net diff against `main` is **purely additive**; legacy emission is untouched.

- `buildUsageEvents` (new, not ingested): builds the target single aggregated `llm_usage_v3` event. Its `cost_awu` is the net billed credits (LLM + tools): per-`(provider, model)` LLM rounding first, then flat integer tool weights added, excluding free/capped/unbillable tool calls. `model_id`/`provider_id` are constant `"aggregate"` placeholders.
- `billedCostAwuFromEvents` (new): derives, from the legacy event payloads, the AWU Metronome will actually bill — `cost_awu` for `llm_usage_v3`, `count` × tool-category weight for `tool_use_v3`, zeroing any `usage_type: free` event.
- Activity `emitMetronomeUsageEventsActivity`: after ingesting the legacy events as before, logs `{ workspaceId, conversationId, agentMessageId, runKey, newCostAwu, oldCostAwu, costMatches }`.

The two amounts come from **independent code paths** (aggregated builder vs. parsing the legacy payloads), so any real divergence in rounding, waivers, or emission filtering shows up in the log.

## Tests

Added/updated in `front/lib/metronome/events.test.ts` (8 tests, all passing):
- The 5 original legacy-builder tests are kept unchanged.
- `buildUsageEvents` aggregates LLM + tool cost into one event.
- Parity on a mixed paid message (per-model rounding = 3; tools 7×3 capped + 1 basic = 22; total **25**), asserting `billedCostAwuFromEvents(legacy) === buildUsageEvents(...).cost_awu`.
- Parity on free-origin messages (both **0**).

Run: `npm run test -- lib/metronome/events.test.ts`. Also `npx tsgo --noEmit` and biome are clean.

## Risk

[...]

PR #31243, opened 2026-08-27
[front] Gate aggregated Metronome usage event behind a feature flag

tdraier:
## Description

Follow-up to #31130. Adds a feature flag that switches the Metronome usage activity from the legacy events to the single aggregated event, per workspace.

- New flag `metronome_aggregated_usage_event` (stage `dust_only`).
- In `emitMetronomeUsageEventsActivity`, both the legacy events (per-model `llm_usage_v3` + per-tool `tool_use_v3`) and the aggregated event (`buildUsageEvents`) are built. When the flag is **on**, only the aggregated event is ingested; when **off**, the legacy events are ingested as before.
- The cost parity log is kept in **all** cases and now also carries `useAggregatedEvent`, so we keep observing new-vs-old cost regardless of which path ingests.

## Tests

- `npm run test -- lib/metronome/events.test.ts` — 9 tests (event builders + legacy/aggregated parity), unchanged and green.
- `activities.ts` type-checks with the flag name (a bad flag name would fail `hasFeatureFlag`'s `WhitelistableFeature` arg).

## Risk

Low. Default behavior is unchanged: with the flag off (the default for all workspaces) the legacy events are ingested exactly as today. The aggregated path is only taken for workspaces explicitly opted in, and `buildUsageEvents` was already merged and shadow-validated against the legacy cost in #31130. Rollback = disable the flag (or revert the commit). No schema/migration.

## Deploy Plan

1. Deploy. No workspace has the flag, so nothing changes.
2. Enable `metronome_aggregated_usage_event` on a Dust test workspace; confirm via the parity log (`useAggregatedEvent: true`, `costMatches: true`) and Metronome that the aggregated event bills the same.
3. Roll out gradually. Once broadly enabled and stable, follow up to drop the legacy builders + `tool_use_v3` and prune the dead Metronome metric/product.

### DIFF
--- a/front/temporal/usage_queue/activities.ts
+++ b/front/temporal/usage_queue/activities.ts
@@ -5,7 +5,7 @@ import {
   trackProgrammaticCost,
 } from "@app/lib/api/programmatic_usage/tracking";
 import type { AuthenticatorType } from "@app/lib/auth";
-import { Authenticator } from "@app/lib/auth";
+import { Authenticator, hasFeatureFlag } from "@app/lib/auth";
 import { ingestMetronomeEvents } from "@app/lib/metronome/client";
 import {
   billedCostAwuFromEvents,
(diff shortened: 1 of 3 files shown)


tdraier:

- #31243 ◀

id13 (approved):

PR #31244, opened 2026-08-27
[front] fix: prevent answer card stretching on exit

aubin-tchoi:
## Description

The UserAnswerRequired card visibly stretches as it exits (can be repro-ed by asking a question and hitting Esc).

My best shot at explaining it is the following: the layout wrapper is shared between the question card and the composer, and it interpolates size between the tall question card and the shorter composer during the animation.

A slow motion video could help pinpointing exactly what the stretch matches with in height.

This PR fixes that.

## Tests

- Tested locally + preview.

## Risk

- Low.

## Deploy Plan

- Deploy front.


pinotalexandre (approved):
LGTM

PR #31245, opened 2026-08-27
[front] fix: exclude skill management from consumption analytics

aubin-tchoi:
## Description

Closes https://github.com/dust-tt/tasks/issues/10177

`enable_skill` counts the tokens for instructions + tool definitions, but we currently show a "Skill Management" line alongside the tools for a skill's breakdown. It's both hard to interpret and misleading because the number there is not directly comparable to the other tools'.

This PR exclude `skill_management` in top tools. This removes it from the Tools table and expanded By tools breakdowns. We keep the attribution, this is important to know which skill is being enabled.

No app preview, everything happens in the backend here.

Current live version

image

Proposal

image

## Tests

- Updated the tests.
- Tested locally.

## Risk

- Low.

## Deploy Plan

- Deploy front.

### DIFF
--- a/front/lib/api/analytics/consumption/top_tools.ts
+++ b/front/lib/api/analytics/consumption/top_tools.ts
@@ -90,6 +90,7 @@ export async function fetchConsumptionTopTools(
 
   return new Ok({
     period,
+    // Keep consumption-share percentages based only on the visible tools.
     totalCredits: totalCredits - skillManagementCredits,
     hasMore,
     totalCount,
--- a/front/lib/api/analytics/consumption/top_tools.ts
+++ b/front/lib/api/analytics/consumption/top_tools.ts
@@ -78,7 +78,8 @@ export async function fetchConsumptionTopTools(
   }
   const { groups, hasMore, totalCount, totalCredits } = result.value;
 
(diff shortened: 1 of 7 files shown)


front/lib/api/analytics/consumption/top_tools.ts
```diff
@@ -77,11 +78,21 @@ export async function fetchConsumptionTopTools(
   }
   const { groups, hasMore, totalCount, totalCredits } = result.value;
 
-  const rows = await resolveConsumptionGroupLabels(auth, "tool", groups);
+  // Enabling a skill is an internal setup step, so do not show it as a tool in
+  // consumption analytics.
+  const visibleGroups = groups.filter(
```
flvndvd, line 83: Shouldn't we filter out in ES query? So the total sums up to 100?

flvndvd (approved):

front/lib/api/analytics/consumption/top_tools.ts
aubin-tchoi: didn't find a way to do this that didn't rely on a `must_not`, and was a bit concerned about the performances of the `must_not`

PR #31248, opened 2026-08-27
Activate committed Frames v2 publications

fontanierh:
## Description

Validate the immutable manifest commit marker before activating a Frames v2 publication. Store the active publication ID in the Frame FileResource metadata only after validation succeeds. Emit a registered audit event after successful activation.

## Tests

- `NODE_ENV=test npm run test -- lib/api/frames/publication_storage.test.ts lib/api/audit/audit_log_schemas.test.ts`
- `npx biome check admin/audit_log_schemas/frame.publication_activated.json lib/api/audit/schema_versions.json lib/api/audit/workos_audit.ts lib/api/frames/publication_storage.ts lib/api/frames/publication_storage.test.ts lib/resources/file_resource.ts types/files.ts`

## Risk

Low. The operation has no callers yet. Failed validation does not mutate the active publication pointer. Audit emission is non-blocking.

## Deploy Plan

Standard deploy. The `frame.publication_activated` WorkOS schema is already registered as v1.

### DIFF
--- a/front/admin/audit_log_schemas/frame.publication_activated.json
+++ b/front/admin/audit_log_schemas/frame.publication_activated.json
@@ -1,0 +1,12 @@
+{
+  "action": "frame.publication_activated",
+  "targets": [
+    { "type": "workspace" },
+    { "type": "frame" }
+  ],
+  "metadata": {
+    "frame_id": "string",
+    "publication_id": "string",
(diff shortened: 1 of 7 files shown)


davidebbo (commented):
LGTM

PR #31249, opened 2026-08-27
[front] fix: show full skill names in slash menu on hover

aubin-tchoi:
## Description

Long skill names are truncated in the slash dropdown, which makes them impossible to read in full.
This PR adds a very slow tooltip on hover. Having the tooltip only if the name is truncated ended would be better, but also ends up being quite complex

## Tests

- Updated existing tests.
- Tested locally.

## Risk

- Low.

## Deploy Plan

- Deploy front.


pinotalexandre (approved):
LGTM

PR #31251, opened 2026-08-27
Skills: no longer suspend editors group membership on archive

fabiencelier:
## Description

related to https://github.com/dust-tt/tasks/issues/10211

Membership was suspended so the group no longer appears in the auth of the user.
However, this is an issue because:
 - the editors of archived can no longer be listed
 - A user can no longer see their "editor-only" archived skills 

This stop suspending the membership when archiving the skills.
We consider it OK because groups will be removed from auth soon.
A follow up would be to stop listing the editors of old archived skills (maybe 6 months+)

This PR also add a migration to restore all the memberships of all the skills editors groups 

## Tests

 - added unit tests

## Risk

low

## Deploy Plan

deploy front
run migration:
```
npx tsx migrations/20260828_restore_skill_editor_memberships.ts
```

### DIFF
--- a/front-api/routes/w/[wId]/skills/[sId]/index.ts
+++ b/front-api/routes/w/[wId]/skills/[sId]/index.ts
@@ -30,6 +30,7 @@ import { workspaceApp } from "@front-api/middlewares/ctx";
 import type { HandlerResult } from "@front-api/middlewares/utils";
 import { apiError } from "@front-api/middlewares/utils";
 import { validate } from "@front-api/middlewares/validator";
+import { rejectArchivedSkill } from "@front-api/routes/w/[wId]/skills/guards";
 import type { Context, TypedResponse } from "hono";
 import uniq from "lodash/uniq";
 import uniqBy from "lodash/uniqBy";
@@ -252,15 +253,9 @@ app.patch(
       });
     }
 
-    // An archived skill is frozen: POST /skills/:sId/restore is the only change it accepts.
-    if (skill.status === "archived") {
-      return apiError(ctx, {
-        status_code: 400,
-        api_error: {
-          type: "invalid_request_error",
-          message: "An archived skill cannot be updated. Restore it first.",
-        },
-      });
+    const archivedError = rejectArchivedSkill(ctx, skill);
+    if (archivedError) {
+      return archivedError;
     }
 
     // Editing a skill remains editor-only; non-editors holding the publish permission use
@@ -527,6 +522,11 @@ app.delete(
           message: "Only admins and editors can archive this skill.",
         },
       });
+    }
+
+    const archivedDeleteError = rejectArchivedSkill(ctx, skill);
+    if (archivedDeleteError) {
+      return archivedDeleteError;
     }
 
     if (skill.status === "suggested") {
--- a/front-api/routes/w/[wId]/skills/[sId]/index.ts
+++ b/front-api/routes/w/[wId]/skills/[sId]/index.ts
@@ -252,6 +252,17 @@ app.patch(
       });
     }
 
+    // An archived skill is frozen: POST /skills/:sId/restore is the only change it accepts.
+    if (skill.status === "archived") {
+      return apiError(ctx, {
+        status_code: 400,
+        api_error: {
+          type: "invalid_request_error",
+          message: "An archived skill cannot be updated. Restore it first.",
+        },
+      });
+    }
+
     // Editing a skill remains editor-only; non-editors holding the publish permission use
     // PATCH /skills/:sId/availability to publish or unpublish without editing.
     if (!skill.canWrite(auth)) {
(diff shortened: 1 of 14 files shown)


front/lib/api/actions/servers/skill_authoring/tools/index.ts
```diff
@@ -540,6 +540,13 @@ const handlers: ToolHandlers<typeof SKILL_AUTHORING_TOOLS_METADATA> = {
       );
     }
 
+    // An archived skill is frozen: restoring it is the only change it accepts.
+    if (skill.status === "archived") {
```
philipperolet, line 544: unrelated fix?

philipperolet: oh because now people have the right as active group members to edit. Ok.

front/lib/resources/skill/skill_resource.ts
philipperolet, line 1: I'd mention in the PR that groups will be fully deleted from auth soon, which is why we can reinstate this.

Also, I'd be wary in the deploy plan, since we will indeed make auth bigger.
As I understand it, we'll add groups + permissions for archived skills in auth with that change, so that might be an issue for old workspaces as we had discussed.
A follow-up filtering out 6+ months old archived skills might be worth it

front-api/routes/w/[wId]/skills/[sId]/index.ts
```diff
@@ -252,6 +252,17 @@ app.patch(
       });
     }
 
+    // An archived skill is frozen: POST /skills/:sId/restore is the only change it accepts.
+    if (skill.status === "archived") {
```
Comment (philipperolet) [inline front-api/routes/w/[wId]/skills/[sId]/index.ts]: This only protects the main PATCH and MCP update. Editors can still modify /editors, /reinforcement, /availability, or call DELETE again. Re-archiving is particularly harmful because archive() can match itself as the existing archived skill and rename it. Let's apply the guard consistently

front-api/routes/w/[wId]/skills/[sId]/index.ts
```diff
@@ -252,6 +252,17 @@ app.patch(
...
+      return apiError(ctx, {
+        status_code: 400,
```
Comment (philipperolet) [inline front-api/routes/w/[wId]/skills/[sId]/index.ts]: We change api contract here (the response from 403 to 400). From afar it seems ok, but flagging just in case

front/migrations/20260828_restore_skill_editor_memberships.ts
```diff
@@ -0,0 +1,159 @@
...
+async function listSkillEditorGrantGroups(
+  auth: Authenticator,
+  workspaceModelId: ModelId
+): Promise<GroupResource[]> {
+  const grants = await GroupPermissionModel.findAll({
+    attributes: ["groupId"],
+    where: {
+      workspaceId: workspaceModelId,
+      grantType: "editor",
+      resourceType: "skill",
```
philipperolet, line 51: maybe worth joining and filtering only on archived skills?

front/migrations/20260828_restore_skill_editor_memberships.ts
```diff
@@ -0,0 +1,159 @@
...
+  if (groups.length === 0) {
+    return;
+  }
+
+  let groupsWritten = 0;
+  let membershipsRestored = 0;
+
+  for (const group of groups) {
+    if (!execute) {
+      const suspendedCount = await countSuspendedMembers(
```
philipperolet: mismatch between !execute and execute
!execute logs only on groups with suspended members while execute calls restore on all groups. maybe the if (!execute) was meant to be further down, inside the suspendedCount > 0 ?

philipperolet (commented):
Thanks 🙏 Left a few coms

front/lib/api/actions/servers/skill_authoring/tools/index.ts
fabiencelier: Yes, now that the users are actually editors of their skills we block editing them
The UI has always blocked it for archived skills

front/migrations/20260828_restore_skill_editor_memberships.ts
fabiencelier: will run the migration after the merge so I want to run it on all groups: if a group is unarchive after the merge and before the migration the membership will not be restored until the script is run.
There is no reason to have suspended membership on a regular_auto group of skill editors, it's only when the skill is archived

fabiencelier: I think it's ok to have a short lapse of time where members would not be restored because un archiving skills is not done often and it will be eventually fixed when the script run anyway.

front/migrations/20260828_restore_skill_editor_memberships.ts
fabiencelier: changed

front-api/routes/w/[wId]/skills/[sId]/index.ts
Comment (fabiencelier) [reply front-api/routes/w/[wId]/skills/[sId]/index.ts]: I have protected the other endpoints too

philipperolet (approved):
👍

PR #31254, opened 2026-08-27
[sparkle] Suppress focus outline on Dialog panel

PopDaph:
Radix focuses the dialog panel itself when nothing tabbable is inside, or when clicking a non-interactive area of the panel (it has `tabindex="-1"`), and the UA focus ring then draws around the whole dialog. Consumers were suppressing it per call site (the analytics popover shipped with `className="focus:outline-none"`); move it into the Dialog base classes so no call site needs it.

You can easily see it on prod: 
- the Tools Dialog form User menu does not have this prop and you can see the weird blue border. 
- the Analytics Dialog from User menu does have this prop and you can see no weird blue border. 

## Risk
Removes a focus ring that only ever appeared on the panel itself, never on interactive elements. Safe to rollback.


aubin-tchoi (approved):
LGTM

PR #31255, opened 2026-08-27
Coordinate Frames v2 publication

fontanierh:
## Description

Add one Frames v2 publish entry point that stores an immutable publication and activates it only after storage succeeds.

## Tests

Added coverage for successful store and activation, plus a failed manifest commit-marker write preserving the current active publication.

## Risk

Low. The coordinator has no callers yet. Failed writes can leave orphaned source objects but cannot change the active publication.

## Deploy Plan

Standard deploy.

### DIFF
--- a/front/lib/api/frames/publication_storage.ts
+++ b/front/lib/api/frames/publication_storage.ts
@@ -232,6 +232,38 @@ export async function activateFramePublication(
   return new Ok(undefined);
 }
 
+export async function publishFramePublication(
+  auth: Authenticator,
+  {
+    frame,
+    manifest,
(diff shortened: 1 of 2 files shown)


front/lib/api/frames/publication_storage.ts
```diff
@@ -232,6 +232,38 @@ export async function activateFramePublication(
   return new Ok(undefined);
 }
 
+export async function publishFramePublication(
```
davidebbo, line 235: nit: "publish publication" sounds a bit funky!

davidebbo (commented):
LGTM

PR #31257, opened 2026-08-27
Declare Frames v2 manifest functions

fontanierh:
## Description

Add function declarations to the Frames v2 manifest contract. Each declaration carries a stable name, description, source entry point, execution mode, and default stake. Missing functions normalize to an empty list; missing execution settings normalize to the existing platform defaults.

Input and output contracts remain defined by the function's Zod schemas. Publication will extract and persist their derived JSON Schemas rather than duplicating them in the manifest.

## Tests

Added parser coverage for valid declarations, defaults, explicit execution settings, unsafe paths, duplicate and invalid names, and persistence bounds.

## Risk

Low. The contract change is additive and backward compatible. Function materialization and execution remain out of scope.

## Deploy Plan

Standard deploy.

### DIFF
--- a/front/types/api/frame_manifest.test.ts
+++ b/front/types/api/frame_manifest.test.ts
@@ -5,6 +5,10 @@ import {
   MAX_FRAME_FUNCTION_DESCRIPTION_LENGTH,
   parseFrameManifest,
 } from "@app/types/api/frame_manifest";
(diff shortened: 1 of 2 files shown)


davidebbo (commented):
LGTM

PR #31258, opened 2026-08-27
Wire checkpointed conversation windows into agent loop

flvndvd:
## Description

This PR completes the feature-flagged integration of checkpointed conversation windows into the agent loop.

Today, each model step reloads and renders the bounded conversation context, then tokenizes the model-visible messages again. With this change, a step can restore the context window produced for the preceding step from GCS, hydrate only the newly completed step, and tokenize only that continuation.

The checkpointed context provider owns checkpoint selection, profile validation, bounded hydration, rendering, and create-once publication. Exact checkpoints are reused on activity retries. Missing, expired, incompatible, or unreadable checkpoints degrade directly to the existing full-rendering provider.

The profile includes model and rendering inputs that affect the resulting context. Checkpoints also account for signed image URL expiry, publication races, conversation deletion, and missing-action catcher state.

The feature is disabled by default behind `stateful_conversation_window`. It does not change Temporal activity parameters or workflow history.

This initially supports statefulness across steps of the same agent message. Cross-agent-message reuse and in-turn compaction remain follow-up extensions.

### DIFF
--- a/front/lib/api/files/processing/index.ts
+++ b/front/lib/api/files/processing/index.ts
@@ -448,14 +448,14 @@ const maybeApplyProcessing = async (
     return new Ok(undefined);
   }
 
-  const start = performance.now();
+  const startedAtMs = performance.now();
   const res = await entry.process(auth, file);
 
-  const elapsed = performance.now() - start;
+  const elapsedMs = performance.now() - startedAtMs;
   logger.info(
     {
       file: file.toPublicJSON(auth),
-      elapsed,
+      elapsedMs,
       error: res.isErr() ? res.error : undefined,
     },
     "Processed file"
--- a/front/logger/logger.ts
+++ b/front/logger/logger.ts
@@ -22,17 +22,20 @@ export const DATADOG_LOG_STATUSES = [
 
 export type DatadogLogStatus = (typeof DATADOG_LOG_STATUSES)[number];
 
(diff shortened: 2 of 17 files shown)


front/lib/api/files/processing/index.ts
```diff
@@ -448,14 +448,14 @@ const maybeApplyProcessing = async (
...
+  const startedAtMs = performance.now();
   const res = await entry.process(auth, file);
 
-  const elapsed = performance.now() - start;
+  const elapsedMs = performance.now() - startedAtMs;
   logger.info(
     {
       file: file.toPublicJSON(auth),
-      elapsed,
+      elapsedMs,
```
flvndvd, line 458: 🚗 `elapsed` is a reserved keyword in DD metrics, and today always pinned to 44ms, which I strongly believe our agent loop takes more.

front/logger/logger.ts
```diff
@@ -22,17 +22,20 @@ export const DATADOG_LOG_STATUSES = [
 
 export type DatadogLogStatus = (typeof DATADOG_LOG_STATUSES)[number];
 
+// Datadog reserves `elapsed` and may replace its value. Use an explicit unit such as
+// `elapsedMs` instead.
+export type DatadogLogContext = {
```
flvndvd, line 27: Overriding the type here so people avoid doing the same mistake.

adrsimon (approved):

PR #31259, opened 2026-08-27
Raise GitHub file size limit to 4GB, and add message when seeting up

davidebbo:
## Description

To address https://github.com/dust-tt/tasks/issues/10213

## Risk

Low

## Deploy Plan

Front and Connector


philipperolet (approved):
👍

PR #31263, opened 2026-08-27
Redesign/Refactor the formatting toolbar and fix its placement

pinotalexandre:
## Summary

Redesigns the rich-text formatting toolbar to match the [Figma spec](https://www.figma.com/design/R7vT7SUjyTSosoiTOABonk/Sparkle-Components?node-id=1759-8994), and fixes it appearing at the wrong place and behind the input bar.

https://github.com/user-attachments/assets/f93a782a-3293-4178-9e08-9fd900410e3d

## The positioning bug

The toolbar used tiptap's `BubbleMenu`, a plugin that creates its element **outside React**, moves it between DOM parents, and overwrites `position`, `left/top` and `visibility` from an async callback. Fighting it from the outside is why the pill kept landing unpredictably and why raising the z-index changed nothing.

Replaced it with `EditorSelectionToolbar`, which we own:

- portals into `document.body`, so no ancestor can clip it or paint over it;
- positions itself `fixed` from ProseMirror's `coordsAtPos`, which is already viewport-relative — immune to ancestor `overflow`, `transform` and `filter`. The home page composer animates a blur filter, which is why the old placement was page-dependent;
- re-syncs on scroll and resize.

Verified in the Sparkle playground: toolbar bottom at 270px against a selection top of 278px — an exact 8px gap, horizontally centred, parented to `body`.

## Design

- `Toolbar` `inline` variant: rounded pill, white→grey gradient, layered shadow, per Figma.
- Dark mode: the gradient previously ended on `primary-900`, but that scale is **inverted** in dark mode and resolved to a near-white, washing the pill out. Now uses the dark surface scale (`stone-800` → `stone-850`) with an inset highlight.
- Desktop buttons `mini` → `sm` (32px); 40px read as too bulky in context.
- Enter animation: 100ms `ease-out-quart`, `scale(0.98) → 1`, `origin-bottom`, with `motion-reduce` handling. Transform/opacity only.

Applied across all three editors that use this toolbar: the input bar, user-message editing, and `MarkdownEditor`.

## Test plan

- [x] `npx tsgo --noEmit` — no new errors (pre-existing unrelated failures also present on `main`)
- [x] Pill styling verified in Storybook, light and dark
- [x] Positioning math verified in the Sparkle playground
- [x] Confirmed working on the app preview, including the home page

### DIFF
--- a/front/components/editor/EditorSelectionToolbar.tsx
+++ b/front/components/editor/EditorSelectionToolbar.tsx
@@ -1,0 +1,118 @@
+import { cn } from "@dust-tt/sparkle";
+import type { Editor } from "@tiptap/react";
+import type { ReactNode } from "react";
+import { useCallback, useEffect, useLayoutEffect, useState } from "react";
+import { createPortal } from "react-dom";
+
+const TOOLBAR_Z_INDEX = 60;
+const SELECTION_GAP_PX = 8;
+
+interface Position {
+  top: number;
+  left: number;
+}
+
+function readSelectionPosition(editor: Editor): Position | null {
+  const { state, view } = editor;
+  if (!view.hasFocus() || state.selection.empty) {
+    return null;
+  }
+
+  const { from, to } = state.selection;
+  const start = view.coordsAtPos(from);
+  const end = view.coordsAtPos(to, -1);
+
+  return {
+    top: Math.min(start.top, end.top) - SELECTION_GAP_PX,
+    left:
+      (Math.min(start.left, end.left) + Math.max(start.right, end.right)) / 2,
+  };
+}
+
+interface EditorSelectionToolbarProps {
+  editor: Editor | null;
+  children: ReactNode;
+  disabled?: boolean;
+}
+
+export function EditorSelectionToolbar({
+  editor,
+  children,
+  disabled = false,
+}: EditorSelectionToolbarProps) {
+  const [position, setPosition] = useState<Position | null>(null);
+  const [hasEntered, setHasEntered] = useState(false);
+
+  const syncPosition = useCallback(() => {
+    if (!editor || editor.isDestroyed) {
+      setPosition(null);
+      return;
+    }
+    setPosition(readSelectionPosition(editor));
+  }, [editor]);
+
+  useEffect(() => {
+    if (!editor || disabled) {
+      setPosition(null);
+      return;
+    }
+
+    editor.on("transaction", syncPosition);
+    editor.on("focus", syncPosition);
+    editor.on("blur", syncPosition);
+
+    return () => {
+      editor.off("transaction", syncPosition);
+      editor.off("focus", syncPosition);
+      editor.off("blur", syncPosition);
+    };
+  }, [editor, disabled, syncPosition]);
+
+  useEffect(() => {
+    if (!position) {
+      return;
+    }
+
+    window.addEventListener("scroll", syncPosition, true);
+    window.addEventListener("resize", syncPosition);
+
+    return () => {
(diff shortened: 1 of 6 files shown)


sparkle/src/components/Toolbar.tsx
```diff
@@ -26,7 +26,7 @@ const toolbarRootVariants = cva("inline-flex items-center", {
       overlay:
         "absolute left-0 top-0 z-10 justify-start gap-3 overflow-hidden rounded-xl bg-primary-50 py-1 pl-3 duration-700 ease-in-out",
       inline:
-        "gap-1 border-b border-t border-border bg-background p-1 sm:rounded-2xl sm:border sm:border-border/50 sm:shadow-md",
+        "gap-1 rounded-full bg-linear-to-b from-white to-muted-background px-1.5 py-1 shadow-[0px_1px_4px_0px_rgba(0,0,0,0.12),0px_1px_0px_0px_rgba(0,0,0,0.08)] dark:from-stone-800 dark:to-stone-850 dark:shadow-[inset_0px_1px_0px_0px_rgba(255,255,255,0.06),0px_0px_0px_1px_rgba(0,0,0,0.4),0px_4px_8px_0px_rgba(0,0,0,0.3)]",
```
aubin-tchoi, line 29: ooc why is the horizontal padding different from the vertical one?

front/components/editor/EditorSelectionToolbar.tsx
```diff
@@ -0,0 +1,118 @@
...
+    }
+    setPosition(readSelectionPosition(editor));
+  }, [editor]);
+
+  useEffect(() => {
+    if (!editor || disabled) {
+      setPosition(null);
+      return;
+    }
+
```
aubin-tchoi, line 59: quite niche but `syncPosition` should probably be called here to cover mounting while an existing selection is already focused (currently the toolbar will remain hidden)

front/components/editor/EditorSelectionToolbar.tsx
```diff
@@ -0,0 +1,118 @@
...
+    window.addEventListener("scroll", syncPosition, true);
+    window.addEventListener("resize", syncPosition);
+
+    return () => {
+      window.removeEventListener("scroll", syncPosition, true);
+      window.removeEventListener("resize", syncPosition);
+    };
+  }, [position, syncPosition]);
+
+  useLayoutEffect(() => {
```
aubin-tchoi: nit: maybe a useEffect would be enough

aubin-tchoi (approved):
LGTM

sparkle/src/components/Toolbar.tsx
pinotalexandre: for better look and feel

front/components/editor/EditorSelectionToolbar.tsx
pinotalexandre: Good catch, fixed in e14017f — `syncPosition()` is now called when the listeners are registered, so mounting with an already-focused selection shows the toolbar right away.

front/components/editor/EditorSelectionToolbar.tsx
pinotalexandre: Agreed, switched to `useEffect` in e14017f. The rAF that flips `hasEntered` fires after paint either way, so there was nothing to gain from running it before paint.

pinotalexandre: Done

front/components/editor/EditorSelectionToolbar.tsx
pinotalexandre: Done

PR #31268, opened 2026-08-27
[front] chore: read API key usage from consumption index

aubin-tchoi:
## Description

Part of https://github.com/dust-tt/tasks/issues/10185.

API key spend-cap usage still reads from the `front.agent_message_analytics` ES index.

This moves the per-key billing-cycle aggregation to `front.agent_message_consumption_analytics` instead, which will become the source of truth for billing at some point (planned to deprecate the other index).

This PR also adds a script to check out the differences before merging.

## Tests

- Added tests.

## Risk

- Low.

## Deploy Plan

- Run script.
- Deploy front.

### DIFF
--- a/front/lib/api/credits/members_usage.test.ts
+++ b/front/lib/api/credits/members_usage.test.ts
@@ -1,9 +1,20 @@
-import { fetchSeatDataForMembersTable } from "@app/lib/api/credits/members_usage";
+import {
+  fetchConsumedAwuCreditsByApiKeyName,
+  fetchSeatDataForMembersTable,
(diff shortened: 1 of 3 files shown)


tdraier:
thanks 🙏 
Shouldn't we also update the user consumption / programmatic usage queries from the same file ?

sfriquet (approved):

aubin-tchoi:
yes, will handle them 1 by 1

aubin-tchoi:
the goal is to remove all usage of `searchAnalytics`

PR #31273, opened 2026-08-27
feat(workspace-analytics): add a billed-credits ranking tool

achilleburah:
## Description

`front.agent_message_analytics` is being deleted, so `@analyst` has to move onto the consumption index. This is the base of that stack, and the first thing the workspace Analytics page can answer that `@analyst` cannot: `get_top_entities_by_credits` ranks any dimension by billed credits, taking the dimension as a parameter.

It is a thin adapter over `fetchConsumptionTopGroups` — the same call the page's nine `consumption/top_*.ts` wrappers each make with a different `dimension`. Three things it needs that the page never did, all three called from the new handler:

- `extraFilters` passthrough, so `agentTagIds` can scope on `agent.tag_ids`, which no filter key covers.
- `includePreviousCredits`, off here: a single window has no vs-previous column to fill, and the lookup is a second round trip.
- A `tag` top dimension over `agent.tag_ids`, already in the mapping — no index change. Like `conversation` it can be ranked but not filtered on, so `isConsumptionScopeDimension` narrows before anything keyed by scope dimension. `resolveDimensionLabels` grows a `tag` and a `conversation` case so every dimension the tool ranks resolves in one place.

`consumptionFilterSchema` is separate from the legacy `usageFilterSchema` rather than shared: the old tools take a singular `source` over `context_origin`, this one takes `sources` over `normalized_origin`, so every value a ranking returns is directly usable as a filter. The legacy one goes away with the last legacy tool, and so does `toConsumptionPeriod`, which converts the inclusive end instant those tools query with `lte` into the half-open bound consumption uses.

Nothing is removed. `get_credit_usage` still answers a window's total; the skill instructions send attribution here instead.

## Tests

Unit tests + Tested locally :

- Which agents cost the most credits this month ?	
- Which conversations were the most expensive?
- Rank our agent tags by credits
- Which API keys spend the most
- Which models cost the most for agents tagged

## Risk

Low. The new options default to today's behaviour and the page is untouched.

## Deploy Plan

Standard deploy. Base of a 4-PR stack: credits ranking → count rankings → overview → timeseries, each replacing its legacy tools as it lands.

### DIFF
--- a/front/lib/api/actions/servers/workspace_analytics/metadata.ts
+++ b/front/lib/api/actions/servers/workspace_analytics/metadata.ts
@@ -11,6 +11,9 @@ import {
 import { CONSUMPTION_TOP_DIMENSIONS } from "@app/lib/api/analytics/consumption/scope";
 import { z } from "zod";
 
+export const GET_TOP_ENTITIES_BY_CREDITS_TOOL_NAME =
+  "get_top_entities_by_credits" as const;
+
 const topListSchema = (entityPlural: string) => ({
   ...timeWindowSchemaShape,
   ...usageFilterSchema,
@@ -122,14 +125,7 @@ const getUsageTimeseriesSchema = {
 const getTopEntitiesByCreditsSchema = {
   dimension: z
     .enum(CONSUMPTION_TOP_DIMENSIONS)
-    .describe(
-      "What to rank: 'agent', 'user', 'model', 'source' (where the traffic " +
-        "came in from), 'api_key', 'group' (a member group), 'tag' (an agent " +
-        "tag), 'tool', 'skill' or 'conversation'. Rows overlap and do not add " +
-        "up to the workspace total for 'model', 'skill', 'tag' and 'group', " +
-        "nor for 'tool' (tool credits exclude the surrounding model work); " +
-        "'user' and 'api_key' have no row for consumption with nobody behind it."
-    ),
+    .describe("What to group results by."),
   ...timeWindowSchemaShape,
   ...consumptionFilterSchema,
   limit: z
@@ -354,19 +350,13 @@ export const WORKSPACE_ANALYTICS_TOOLS_METADATA = [
     freeUsage: true,
   },
   {
-    name: "get_top_entities_by_credits",
+    name: GET_TOP_ENTITIES_BY_CREDITS_TOOL_NAME,
     description:
-      "Rank what costs the most credits over a time window (defaults to the " +
-      "current calendar month): the most expensive agents, members, models, " +
-      "skills, tools, sources, API keys, member groups, agent tags or " +
-      "conversations. Each row carries its credits, its average credits per " +
-      "message, and how many messages or tool calls it covers. These are the " +
-      "rankings the workspace Analytics page shows, over billed credits, so " +
-      "the figures match the Usage page exactly. Use this to break credit " +
-      "spending down by agent, user or model, and for any spend, cost or " +
-      "expense question — where the budget goes, which agent is most " +
-      "expensive, which integration burns credits. Prefer it over " +
-      "get_credit_usage, whose figures are estimates. Admin-only.",
+      "Rank the workspace's credit consumption by entity " +
+      "over a time window (defaults to the current calendar month). Use " +
+      "this to answer 'which agent is most expensive', or to attribute " +
+      "credit spend by API key, tag, or model. Figures are billed " +
+      "credits. Rows may overlap, so don't sum them for a workspace total.",
     schema: getTopEntitiesByCreditsSchema,
     stake: "never_ask",
     displayLabels: {
@@ -374,7 +364,7 @@ export const WORKSPACE_ANALYTICS_TOOLS_METADATA = [
       done: "Retrieved top consumers",
     },
     toolCostCategory: "basic",
-    freeUsage: false,
+    freeUsage: true,
   },
 ] as const;
 
--- a/front/lib/api/actions/servers/workspace_analytics/metadata.ts
+++ b/front/lib/api/actions/servers/workspace_analytics/metadata.ts
@@ -1,5 +1,6 @@
 import type { ServerMetadata } from "@app/lib/actions/mcp_internal_actions/tool_definition";
 import {
+  consumptionFilterSchema,
   DEFAULT_CREDIT_GROUPS,
   DEFAULT_RESULTS,
   MAX_CREDIT_GROUPS,
@@ -7,6 +8,7 @@ import {
   timeWindowSchemaShape,
   usageFilterSchema,
 } from "@app/lib/api/actions/servers/workspace_analytics/query_input";
+import { CONSUMPTION_TOP_DIMENSIONS } from "@app/lib/api/analytics/consumption/scope";
 import { z } from "zod";
 
 const topListSchema = (entityPlural: string) => ({
@@ -117,6 +119,31 @@ const getUsageTimeseriesSchema = {
     ),
 };
 
+const getTopEntitiesByCreditsSchema = {
+  dimension: z
+    .enum(CONSUMPTION_TOP_DIMENSIONS)
+    .describe(
+      "What to rank: 'agent', 'user', 'model', 'source' (where the traffic " +
+        "came in from), 'api_key', 'group' (a member group), 'tag' (an agent " +
+        "tag), 'tool', 'skill' or 'conversation'. Rows overlap and do not add " +
+        "up to the workspace total for 'model', 'skill', 'tag' and 'group', " +
+        "nor for 'tool' (tool credits exclude the surrounding model work); " +
+        "'user' and 'api_key' have no row for consumption with nobody behind it."
+    ),
+  ...timeWindowSchemaShape,
+  ...consumptionFilterSchema,
+  limit: z
+    .number()
+    .int()
+    .positive()
+    .max(MAX_RESULTS)
+    .optional()
+    .describe(
+      `Maximum number of rows to return ` +
+        `(default ${DEFAULT_RESULTS}, max ${MAX_RESULTS}).`
+    ),
+};
+
 export const WORKSPACE_ANALYTICS_TOOLS_METADATA = [
   {
     name: "get_top_agents",
@@ -214,7 +241,7 @@ export const WORKSPACE_ANALYTICS_TOOLS_METADATA = [
       "Return the workspace's most-used skills over a time window (defaults " +
       "to the current calendar month), ranked by execution count. Optionally " +
       "filter by source (context_origin), agent, user, tag, or model. Use this " +
-      "to answer which skills are used most. Admin-only.",
+      "to answer which skills are executed or used most. Admin-only.",
     schema: getTopSkillsSchema,
     stake: "never_ask",
     displayLabels: {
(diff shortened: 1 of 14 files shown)


front/lib/api/actions/servers/workspace_analytics/tools/index.ts
```diff
@@ -974,6 +982,73 @@ const handlers: ToolHandlers<typeof WORKSPACE_ANALYTICS_TOOLS_METADATA> = {
...
+          text: `No ${dimension} consumption recorded for ${label} (${tz}).`,
+        },
+      ]);
+    }
+
+    const rows = await resolveConsumptionGroupLabels(auth, dimension, groups);
+    const unit = CONSUMPTION_TOP_DIMENSION_UNIT[dimension];
+    const lines = rows.map(
+      (row, index) =>
+        `${index + 1}. ${row.name} [${row.key}] — ` +
```
avervaet, line 1035: Could we avoid returning row.key for unreadable conversation groups? This workspace-manager query ranks all conversations, while authorization filters inaccessible conversations only during label resolution, so the fallback still leaks their IDs and billed spend. Please remove workspace-wide conversation ranking or omit groups the caller cannot read, with a private-conversation test.

avervaet (approved):

front/lib/api/actions/servers/workspace_analytics/tools/index.ts
achilleburah: good point, added a filter

front/lib/api/actions/servers/workspace_analytics/tools/index.ts
```diff
@@ -974,6 +982,73 @@ const handlers: ToolHandlers<typeof WORKSPACE_ANALYTICS_TOOLS_METADATA> = {
         return assertNever(selectedMetric);
     }
   },
+  get_top_entities_by_credits: async (
+    { dimension, limit, ...input },
+    { auth }
+  ) => {
+    const denied = workspaceManagerGuard(auth);
```
flvndvd: `nit` boolean are usually starting with `is` -> `isDenied`

flvndvd: And actually it's not a boolean but and `Error`, this function should use a `Result` at first, mind updating in a follow up PR?

front/lib/api/actions/servers/workspace_analytics/tools/index.ts
```diff
@@ -974,6 +982,73 @@ const handlers: ToolHandlers<typeof WORKSPACE_ANALYTICS_TOOLS_METADATA> = {
...
+      );
+    }
+
+    const { label, timezone: tz } = window.value;
+    const { groups, totalCredits } = result.value;
+
+    if (groups.length === 0) {
+      return new Ok([
+        {
+          type: "text" as const,
```
flvndvd, line 1025:
```suggestion
          type: "text",
```

front/lib/api/actions/servers/workspace_analytics/tools/index.ts
```diff
@@ -974,6 +982,73 @@ const handlers: ToolHandlers<typeof WORKSPACE_ANALYTICS_TOOLS_METADATA> = {
...
+      (row, index) =>
+        `${index + 1}. ${row.name} [${row.key}] — ` +
+        `${row.credits.toFixed(2)} credits, ` +
+        `${row.count} ${unit}${pluralize(row.count)} ` +
+        `(${row.avgCredits.toFixed(4)} per ${unit})`
+    );
+
+    return new Ok([
+      {
+        type: "text" as const,
```
flvndvd, line 1043:
```suggestion
        type: "text",
```

front/lib/api/actions/servers/workspace_analytics/metadata.ts
```diff
@@ -117,6 +119,31 @@ const getUsageTimeseriesSchema = {
     ),
 };
 
+const getTopEntitiesByCreditsSchema = {
+  dimension: z
+    .enum(CONSUMPTION_TOP_DIMENSIONS)
+    .describe(
+      "What to rank: 'agent', 'user', 'model', 'source' (where the traffic " +
```
flvndvd: Pretty hard to understand IMHO.

front/lib/api/actions/servers/workspace_analytics/metadata.ts
```diff
@@ -326,6 +353,29 @@ export const WORKSPACE_ANALYTICS_TOOLS_METADATA = [
     toolCostCategory: "basic",
     freeUsage: false,
   },
+  {
+    name: "get_top_entities_by_credits",
+    description:
```
flvndvd, line 354: IMHO we want to dry this a bit

front/lib/api/actions/servers/workspace_analytics/metadata.ts
```diff
@@ -326,6 +353,29 @@ export const WORKSPACE_ANALYTICS_TOOLS_METADATA = [
...
+      "expense question — where the budget goes, which agent is most " +
+      "expensive, which integration burns credits. Prefer it over " +
+      "get_credit_usage, whose figures are estimates. Admin-only.",
+    schema: getTopEntitiesByCreditsSchema,
+    stake: "never_ask",
+    displayLabels: {
+      running: "Retrieving top consumers",
+      done: "Retrieved top consumers",
+    },
+    toolCostCategory: "basic",
```
flvndvd, line 366: Let's make this tool free, people should not have to pay for this one tool.

front/lib/api/actions/servers/workspace_analytics/query_input.ts
```diff
@@ -95,6 +99,116 @@ export const usageFilterSchema = {
...
+// Filters for the consumption-index tools. Every key narrows the same scope the
+// workspace Analytics page filters on, so any value a ranking returns can be fed
+// straight back in. The legacy `usageFilterSchema` above stays until the tools
+// still reading the old index are gone.
+export const consumptionFilterSchema = {
+  sources: z
+    .array(z.string())
+    .optional()
+    .describe(
+      "Restrict to these message origins — where a message came in from. Many " +
```
flvndvd: Can we avoid AI slope in here? Like no emdashes or over use of `;`?

flvndvd: I'd personally avoid listing the whole list here.

front/lib/api/analytics/consumption/labels.ts
```diff
@@ -188,14 +194,44 @@ export async function resolveDimensionLabels(
...
+
+    case "tag": {
+      const tags = await TagResource.fetchByIds(auth, keys);
+      const namesById = new Map(tags.map((tag) => [tag.sId, tag.name]));
+      return labelsFromNames(
+        new Map(keys.map((key) => [key, namesById.get(key) ?? key]))
+      );
+    }
+
+    case "reasoning_effort":
```
flvndvd, line 222: Pretty sure @aubin-tchoi added this as well, no?

front/lib/api/analytics/consumption/top.test.ts
```diff
@@ -1,6 +1,10 @@
 import { listConsumptionFacetCatalogDimension } from "@app/lib/api/analytics/consumption/facet_catalog";
 import { resolveDimensionLabels } from "@app/lib/api/analytics/consumption/labels";
 import type { ConsumptionPeriod } from "@app/lib/api/analytics/consumption/period";
+import {
+  fetchConsumptionTopGroups as fetchConsumptionTopGroupBuckets,
```
flvndvd, line 5: Can we avoid rename?

front/lib/api/analytics/consumption/top.ts
```diff
@@ -245,11 +244,13 @@ async function fetchConsumptionPreviousCredits(
     dimension,
     previousPeriod,
     filter,
+    extraFilters,
     keys,
   }: {
     dimension: ConsumptionTopDimension;
     previousPeriod: ConsumptionPeriod;
     filter?: ConsumptionScopeFilter;
+    extraFilters?: estypes.QueryDslQueryContainer[];
```
flvndvd: Up until now, this module was deep enough so it did not leak any Elasticsearch jargo/type outside, let's preserve it that way 🙏

front/lib/api/actions/servers/workspace_analytics/tools/index.ts
achilleburah: typed with as const in every current tool, can remove it but probably need to update them all for consistency, will do in following PR

front/lib/api/analytics/consumption/top.test.ts
achilleburah: two existing functions with same name tested in same files. Will rename one of those in another PR

front/lib/api/analytics/consumption/top.ts
achilleburah: sure thing

front/lib/api/actions/servers/workspace_analytics/tools/index.ts
flvndvd: Yeah this is definitely model being overly eager on TS 😛

front/lib/api/actions/servers/workspace_analytics/tools/index.ts
```diff
@@ -126,9 +134,9 @@ const handlers: ToolHandlers<typeof WORKSPACE_ANALYTICS_TOOLS_METADATA> = {
     },
     { auth }
   ) => {
-    const denied = workspaceManagerGuard(auth);
-    if (denied) {
-      return new Err(denied);
+    const isDenied = workspaceManagerGuard(auth);
```
flvndvd: Ahah we might cross-fired our comments, but it's an error here, so IMHO `workspaceManagerGuard` should be move to a `Result` or return an actual boolean and error is then crafted here.

front/lib/api/actions/servers/workspace_analytics/metadata.ts
```diff
@@ -326,6 +352,24 @@ export const WORKSPACE_ANALYTICS_TOOLS_METADATA = [
     toolCostCategory: "basic",
     freeUsage: false,
   },
+  {
+    name: "get_top_entities_by_credits",
+    description:
+      "Rank what costs the most credits over a time window (defaults to the " +
```
flvndvd: `nit`
```suggestion
      "Rank what drive the most credits over a time window (defaults to the " +
```

front/lib/api/actions/servers/workspace_analytics/metadata.ts
```diff
@@ -326,6 +352,24 @@ export const WORKSPACE_ANALYTICS_TOOLS_METADATA = [
...
+      "current calendar month), along the chosen dimension (agents, users, " +
+      "models, tools, skills, sources, API keys, groups, tags, conversations, " +
+      "or reasoning effort). Figures are billed credits and match the Usage " +
```
flvndvd: Do we really need to talk about "Usage page" here, model might be clueless on what this is.

front/lib/api/actions/servers/workspace_analytics/metadata.ts
```diff
@@ -326,6 +352,24 @@ export const WORKSPACE_ANALYTICS_TOOLS_METADATA = [
...
+      "page exactly, so prefer this over get_credit_usage, whose figures are " +
```
flvndvd: `nit`
We use constants for tool name, so it never drift between description and actual tool name.

front/lib/api/actions/servers/workspace_analytics/metadata.ts
```diff
@@ -326,6 +352,24 @@ export const WORKSPACE_ANALYTICS_TOOLS_METADATA = [
...
+      "estimates, for any spend, cost or expense question. Admin-only.",
```
flvndvd: Do we need the admin only? Knowing that the agent is 1. gated behind admin and 2. model/agent has no way to know if current is an admin.

front/lib/api/actions/servers/workspace_analytics/query_input.ts
```diff
@@ -95,6 +97,106 @@ export const usageFilterSchema = {
...
+    .optional()
+    .describe(
+      "Restrict to these models, identified by their model id (e.g. " +
+        "'claude-sonnet-4-5'), as returned by a 'model' ranking."
+    ),
+  apiKeyNames: z
+    .array(z.string())
+    .optional()
+    .describe(
+      "Restrict to these API key names, as returned by an 'api_key' ranking."
```
flvndvd: Slightly confusing here, dimension seems to always be camelCase, no?

front/lib/api/actions/servers/workspace_analytics/metadata.ts
achilleburah: removed the mention

front/lib/api/actions/servers/workspace_analytics/query_input.ts
achilleburah: they're defined in snake_case in the analytics/consumption

front/lib/api/analytics/consumption/labels.ts
aubin-tchoi: didn't add the reasoning effort as a dimension per se, it's only available in the breakdown

front/lib/api/actions/servers/workspace_analytics/query_input.ts
achilleburah: i removed the unecessary precision though

flvndvd (approved):

PR #31274, opened 2026-08-27
Add script to merge duplicate users keeping the provisioned one

davidebbo:
A workspace ended up with duplicate users: same email, different user ids. We already have a few merge scripts (`merge_xx_user_identities.ts`, `merge_domain_user_identities.ts`, `batch_merge_user_identities.ts`) but none of them pick the primary based on the membership origin, which is what we want here: keep the `provisioned` account.

This adds `front/scripts/merge_duplicate_users_by_origin.ts`, modeled on the existing scripts:

- Fetches the workspace's active members via `getMembers` (which already returns each member's membership `origin`).
- Groups them by lowercased email and keeps only the groups with more than one member.
- For each duplicate group, if **exactly one** member has `origin === "provisioned"`, that one becomes the primary and all the others are merged into it via `mergeUserIdentities`. If none or several are provisioned, there's no way to pick a primary, so the group is skipped and logged with its sIds and origins.
- `--email` restricts the run to a single duplicate, `--revokeSecondary` (off by default) revokes the secondary membership, and like every `makeScript` it's a dry run unless `--execute` is passed.

```
npx tsx front/scripts/merge_duplicate_users_by_origin.ts -w <wId>
npx tsx front/scripts/merge_duplicate_users_by_origin.ts -w <wId> --email user@corp.com --revokeSecondary --execute
```

Note: since emails are matched case-insensitively by the script itself, `enforceEmailMatch: false` is passed to `mergeUserIdentities`.

### DIFF
--- a/front/scripts/merge_duplicate_users_by_origin.ts
+++ b/front/scripts/merge_duplicate_users_by_origin.ts
@@ -1,0 +1,208 @@
+import { getMembers } from "@app/lib/api/workspace";
+import { Authenticator } from "@app/lib/auth";
+import { mergeUserIdentities } from "@app/lib/iam/users";
+import { WorkspaceResource } from "@app/lib/resources/workspace_resource";
+import { normalizeError } from "@app/types/shared/utils/error_utils";
+import type { UserTypeWithWorkspaces } from "@app/types/user";
+
+import { makeScript } from "./helpers";
+
+makeScript(
+  {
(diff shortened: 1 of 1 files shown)


tdraier (approved):

davidebbo:
This is for https://github.com/dust-tt/tasks/issues/10219

PR #31275, opened 2026-08-27
[front] chore: enforce tracking known areas and actions

aubin-tchoi:
## Description

We can currently pass any string as a tracking area or action.
This can lead to an inconsistent naming in the events.

This PR enforces passing known values from `TRACKING_AREAS` and `TRACKING_ACTIONS`.

Also renames the event `assistant:model_picker:exposure` into `assistant:model_picker:view` for the sake of uniformization and removes the `data_sources:create:success` event (duplicate with `data_sources:connection:create`).

## Risk

- Low.

## Deploy Plan

- Deploy front.


theogz (approved):
looks good!

PR #31276, opened 2026-08-27
Revert "[front] fix: keep the impersonated user's role on system-key auth exchange"

davidebbo:
Reverts dust-tt/dust#31232


flvndvd:
LGTM

PR #31267, opened 2026-08-27
[front] use consumption ES index for tool_usage export

avervaet:
## Description

Completes [#31266](https://github.com/dust-tt/dust/pull/31266): `tool_usage` now reads from the consumption index.

Retires `buildAgentAnalyticsBaseQuery` from `export_tables.ts`, every export table now reads from the consumption index.

## Tests

- New `tool_usage` tests (data source + non-UTC bounds regression).

## Risk

Low, same profile as the earlier PRs. Last table on the old index — panel is fully on the consumption index after this merges.

## Deploy Plan

Standard deploy. Stacked on #31266.

PR #31266, opened 2026-08-27
[front] use consumption ES index for skill_usage export

avervaet:
## Description

Continues [#31265](https://github.com/dust-tt/dust/pull/31265): `skill_usage` now reads from the consumption index. 

## Tests

- New `skill_usage` tests (data source + non-UTC bounds regression).
## Risk

Low, same profile as the earlier PRs in this stack.

## Deploy Plan

Standard deploy. Stacked on #31265.

PR #31265, opened 2026-08-27
[front] use consumption ES index for active_users export

avervaet:
## Description

Continues [#31264](https://github.com/dust-tt/dust/pull/31264): `active_users` now reads from the consumption index (`completed_at`/`user.id`) instead of the legacy `timestamp`/`user_id` index.

## Tests

- New `active_users` tests (data source + non-UTC bounds regression).

## Risk

Low, same profile as #31264.

## Deploy Plan

Standard deploy. Stacked on #31264.

PR #31264, opened 2026-08-27
[front] use consumption ES index for usage_metrics export

avervaet:
## Description

Same move as [#31087](https://github.com/dust-tt/dust/pull/31087) / [#31095](https://github.com/dust-tt/dust/pull/31095), applied to `usage_metrics`: reads from the consumption ES index now, for both the dedicated export endpoint and the combined export panel.

CSV output unchanged

## Tests

- New `usage_metrics` tests (data source + non-UTC bounds regression) in `export_tables.test.ts`.
- Updated agents/users/skills/source tests for the new instant-based bounds.

## Risk

Low. Panel temporarily sources tables from two different indices, same as prior PRs in this migration.

## Deploy Plan

Standard deploy. Base of a 4-PR stack: usage_metrics → active_users → skill_usage → tool_usage.

PR #31277, opened 2026-08-27
[front] read feedback export from Postgres instead of Elasticsearch

avervaet:
## Description

Continues the export-tables migration ([#31267](https://github.com/dust-tt/dust/pull/31267), stacked on top), but for a different reason than the earlier tables: the `feedback` table's data was never in the consumption index — it lived entirely in Postgres already, with the old ES index only providing feedback id/thumb/dismissed/created_at. Since the legacy index is being retired, `fetchFeedbackExportRows` now reads straight from `AgentMessageFeedbackResource.getFeedbackUsageDataForWorkspace` (same Resource method already powering the older `workspace_usage.ts` CSV export), resolving agent name and user sId/email with a couple of batched Postgres lookups.

Behavior note: `getFeedbackUsageDataForWorkspace` silently drops feedback whose author user record can no longer be resolved (e.g. a deleted user) — the old ES-backed export kept those rows with an empty `userEmail` instead. Accepting this since it matches the existing, already-shipped Postgres path.

## Tests

- New `feedback_export.test.ts`: one test creating a real feedback row via factories and asserting the full row shape (agent name, user sId/email, conversation URL), one for the empty-window case.
- Full `lib/api/analytics` suite passes (123 tests) — `feedback_export`'s existing mock in the v1 export endpoint test still applies unchanged (same function name/signature).

## Risk

Low — feedback data was already 100% in Postgres, this doesn't compute or infer anything new. Main risk is the deleted-user-drops-the-row behavior noted above, which is a pre-existing, already-shipped semantic elsewhere in the codebase.

## Deploy Plan

Standard deploy, no migration or feature flag needed. Stacked on #31267.

### DIFF
--- a/front/lib/api/analytics/feedback_export.test.ts
+++ b/front/lib/api/analytics/feedback_export.test.ts
@@ -1,0 +1,100 @@
+import { fetchFeedbackExportRows } from "@app/lib/api/analytics/feedback_export";
+import { AgentMessageFeedbackResource } from "@app/lib/resources/agent_message_feedback_resource";
+import { AgentConfigurationFactory } from "@app/tests/utils/AgentConfigurationFactory";
+import { ConversationFactory } from "@app/tests/utils/ConversationFactory";
+import { createResourceTest } from "@app/tests/utils/generic_resource_tests";
+import type { ModelId } from "@app/types/shared/model_id";
+import moment from "moment-timezone";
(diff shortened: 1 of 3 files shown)


avervaet:

- #31264
    - #31265
        - #31266
            - #31267
                - #31277 ◀

achilleburah (approved):
LGTM ! 🔥

PR #31278, opened 2026-08-27
Drop the host tag from statsd distributions via a statsDMetrics wrapper

flvndvd:
## Description

Our custom metrics cardinality monitor keeps flickering. Digging into it, the main offender is not our own tags: it's the `host` tag that `dd-agent` stamps on every statsd metric. A distribution like `llm_time_to_first_event_ms` has ~50 legit tag combinations, but multiplied by ~150 hosts (x5 distribution aggregations) it lands at ~7900 indexed custom metrics. For a latency metric, host is a dimension nobody ever queries.

What I tried before landing here:
- **Fix it in infra**: there is no agent option to drop the host tag from `dogstatsd` metrics. It's been an open feature request since 2019 (DataDog/datadog-agent#3130). Tag cardinality is already at the minimum (`low`), nothing left to tune there.
- **Fix it in Datadog**: Metrics without Limits tag configs do work, but it's a one-time manual config *per metric*. Every new metric ships fat by default and someone has to remember to trim it. We used on many metrics so far, but let's be honest, it does not scale! 
- **Fix it at emission**: turns out the agent treats a client-sent `host:` tag as a hostname override, and an *empty* value submits the metric with no host at all. Undocumented but explicitly unit-tested in the agent (`TestConvertParseSingleWithEmptyHostTags`)[[source](https://github.com/DataDog/datadog-agent/blob/2dcbee145dfc903b105214ab95ee34478dfa8db4/comp/dogstatsd/server/impl/enrich_test.go#L200-L226)]. Verified it live with a vanity `dust.cardinality_canary` metric from a front pod (🙈 ): the `hostless` point shows `host: N/A` in [Datadog](https://app.datadoghq.eu/metric/explorer?fromUser=true&graph_layout=stacked&session_id=96b38de2-bf3b-4f01-9b12-7d9e9c069ffe&start=1787836262285&end=1787837807714&paused=true#N4Ig7glgJg5gpgFxALlAGwIYE8D2BXJVEADxQEYAaELcqyKBAC1pEbghkcLIF8qo4AMwgA7CAgg4RKUAiwAHOChASAtnADOcAE4RNIKtrgBHPJoQaUAbVCCc21XkyXkNkHYdOMy0zqxkQHgBdKl9dfVdQKAwEDAB9DXxtAGMlZBB1BF1kyyoRDHUfM21-AxAwmnTVDGJkKDwNBAA6ZIxtKFEMNHEsONb8kuAAKh4AAgAjLFHgfrasCkYcRp5AkJAjDXkpLTiPasIVCHUtcNyQRqw0NNB5LsQENJAoHBg+5w0IZLLukTg4uUUykS3Sg31Ef3oTGUIns1TQgWCPD453k3QQAGEpMIYCgRE40DwgA), the control point shows the node. It works.

So this PR makes emission the fix. `statsd.ts` now exposes a `statsDMetrics` wrapper (raw `hot`-shots client is no longer exported) that appends the empty `host:` tag to **distributions** by default. Distributions are aggregated server-side so dropping host changes nothing about query results. Counters/gauges/histograms/timings keep the host tag: they're aggregated per host, and identical hostless series from two nodes can collide on a timestamp and silently lose points. Every method takes `{ includeHostTag }` to override either way.

> [!NOTE]
> Expectation: Drop cardinalities by at least size of the fleet and remains flat as spot nodes are being evicted/node scaled.

### DIFF
--- a/front-api/middlewares/request_instrumentation.ts
+++ b/front-api/middlewares/request_instrumentation.ts
@@ -1,6 +1,6 @@
 import type { Authenticator } from "@app/lib/auth";
 import type { SessionWithUser } from "@app/lib/iam/provider";
-import { getStatsDClient } from "@app/lib/utils/statsd";
+import { statsDMetrics } from "@app/lib/utils/statsd";
 import logger from "@app/logger/logger";
 import tracer from "@app/logger/tracer";
 import {
@@ -141,8 +141,8 @@ export const requestInstrumentation =
       `streaming:${streaming}`,
       `status_code:${statusCode}`,
     ];
-    getStatsDClient().increment("requests.count", 1, tags);
-    getStatsDClient().distribution(
+    statsDMetrics.increment("requests.count", 1, tags);
+    statsDMetrics.distribution(
       "requests.duration.distribution",
       durationMs,
       tags
--- a/front-api/routes/healthz/ready.ts
+++ b/front-api/routes/healthz/ready.ts
@@ -1,5 +1,5 @@
 import { COMMIT_HASH } from "@app/lib/commit-hash";
-import { getStatsDClient } from "@app/lib/utils/statsd";
+import { statsDMetrics } from "@app/lib/utils/statsd";
 import { createHono } from "@front-api/lib/hono";
 
 /**
@@ -24,7 +24,7 @@ app.get("/", (ctx) => {
   const response = ctx.json({ status: "ready", commitHash: COMMIT_HASH }, 200);
 
   const durationMs = performance.now() - startMs;
-  getStatsDClient().distribution("healthz.ready.duration_ms", durationMs);
+  statsDMetrics.distribution("healthz.ready.duration_ms", durationMs);
 
   return response;
 });
--- a/front-api/routes/healthz/startup.ts
+++ b/front-api/routes/healthz/startup.ts
@@ -1,6 +1,6 @@
 import { getRedisHybridManager } from "@app/lib/api/redis-hybrid-manager";
 import { frontSequelize } from "@app/lib/resources/storage";
-import { getStatsDClient } from "@app/lib/utils/statsd";
+import { statsDMetrics } from "@app/lib/utils/statsd";
 import logger from "@app/logger/logger";
 import { normalizeError } from "@app/types/shared/utils/error_utils";
 import { createHono } from "@front-api/lib/hono";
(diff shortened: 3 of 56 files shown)


PopDaph (approved):
Looks reasonable :D 
Thanks vm for fixing 🙏🏻

front-api/middlewares/request_instrumentation.ts
id13, line 150: this is probably one of the few where we want the host imo

front-api/routes/healthz/ready.ts
```diff
@@ -24,7 +24,7 @@ app.get("/", (ctx) => {
   const response = ctx.json({ status: "ready", commitHash: COMMIT_HASH }, 200);
 
   const durationMs = performance.now() - startMs;
-  getStatsDClient().distribution("healthz.ready.duration_ms", durationMs);
+  statsDMetrics.distribution("healthz.ready.duration_ms", durationMs);
```
id13, line 27: also this

front-api/routes/healthz/startup.ts
```diff
@@ -72,7 +72,7 @@ app.get("/", async (ctx) => {
       "Startup probe succeeded - dependencies connected"
     );
 
-    getStatsDClient().distribution("healthz.startup.duration_ms", durationMs, [
+    statsDMetrics.distribution("healthz.startup.duration_ms", durationMs, [
```
id13, line 75: this

front-api/routes/healthz/startup.ts
```diff
@@ -87,7 +87,7 @@ app.get("/", async (ctx) => {
     `Startup probe failed - ${failed.map((r) => r.name).join(", ")} not ready`
   );
 
-  getStatsDClient().distribution("healthz.startup.duration_ms", durationMs, [
+  statsDMetrics.distribution("healthz.startup.duration_ms", durationMs, [
```
id13, line 90: this

id13 (commented):
LGTM

front-api/middlewares/request_instrumentation.ts
flvndvd: So increment has host by default (since it's needed for those metrics) but distribution I was "nah". Was your point on increment or distribution?

front-api/routes/healthz/ready.ts
flvndvd: Why not, let me send you a PR.

PR #31279, opened 2026-08-27
Always show one decimal in Credits / message column

PopDaph:
In the analytics Attribution table, the Credits / message column mixes formats: 46.1 next to 310 or 911. Always show one decimal so values read consistently (310 renders as 310.0).

Adds a `formatAvgCredits` helper (min and max one fraction digit) and uses it in the avg column cell, which serves all Attribution tabs. `formatCredits` is unchanged so Total credits and cost menus keep their format.


front/components/workspace/analytics/consumption/ConsumptionAttributionTable.tsx
```diff
@@ -366,7 +366,7 @@ function buildColumns({
       cell: (info) => (
         <DataTable.BasicCellContent
           className="justify-end text-right tabular-nums"
-          label={formatCredits(info.row.original.avgCredits)}
+          label={formatAvgCredits(info.row.original.avgCredits)}
```
PopDaph, line 369: I don't think we want to show decimals everywhere, hence the dedicated function for avg columns

adrsimon (approved):

PR #31281, opened 2026-08-27
Clear attribution search when adding a row to filters

PopDaph:
On the analytics consumption page, searching for a row in the Attribution table and then clicking its funnel button keeps the search text, so the freshly filtered table stays narrowed by a stale search (often to the one row just pinned).

Clicking "Add to filters" now clears the search (with a debounce flush so the table refetches immediately). Removing a filter leaves the search untouched.

Bug repro: I want to see the agents used bu Aubin: I type "Aub" in Members and filter by Aubin, then I click on the Agents tab and I have to clear the search bar


aubin-tchoi (approved):
LGTM

PR #30770, opened 2026-08-19
[front-api] chore: remove unused openai dependency

avervaet:
## Description

Remove the unused `openai` dependency from `front-api/package.json` and `package-lock.json`.

`front-api` never imports the `openai` package directly — it talks to OpenAI's API via raw `fetch()` calls where needed (e.g. `routes/w/[wId]/providers/[pId]/check.ts`). The dependency was carried over from `front/package.json`'s dependency list when `front-api` split off as the Hono strangler router (#25136, merged 2026-05-11); it looks like copy-paste from `front` rather than an actual need. `front` itself still uses the `openai` SDK directly (see `front/lib/model_constructors/**`) and keeps its own dependency.

## Tests

Verified with `knip --dependencies` and a repo-wide grep that `openai` is not imported anywhere under `front-api/`. Ran `npm install --package-lock-only -w front-api` to regenerate the lockfile.

## Risk

Low: dependency-only change with no runtime code changes.

## Deploy Plan

None — no runtime behavior change.

PR #31284, opened 2026-08-27
fix(front-api): declare openai explicitly so the bundle resolves the Responses SDK

Nils-Fedrigo:
## Description

Triggers natural language=>CRON don't work locally if we don't have the right dep 
Screenshot 2026-08-27 at 16 40 35

Re-declare `openai` as an explicit dependency of `front-api`, reverting the removal in #30770.

`front-api` has no `import ... from "openai"` in its own source, so the dependency looks unused —
but that is not what makes it load-bearing. `front-api` bundles `front`'s source through the
`@app/*` alias, including `front/lib/model_constructors/{stream,batch}/clients/openai_responses.ts`,
which imports the SDK. esbuild externalizes every bare specifier
(`front-api/esbuild.shared.ts`), so `require("openai")` is resolved by Node at runtime from
`front-api/dist/`. With no copy in `front-api/node_modules`, that walk-up reaches the hoisted root
copy, which in a full monorepo install is **openai 4.77.4** — the exact peer pin of
`@microsoft/teams-ai` (a `connectors` dependency). 4.77.4 predates the Responses API, so
`client.responses` is `undefined` and every Responses call fails with:

```
TypeError: Cannot read properties of undefined (reading 'create')
```

which `classifyStreamError` cannot recognise as an SDK error, so it surfaces to callers as the
opaque `LLM error: Unknown error from OpenAI`. Locally this breaks every front-api route that
reaches an OpenAI Responses endpoint (agent runs, `text_as_cron_rule`, similar-skill and
similar-agent discovery).

Production is unaffected today: the front image installs only
`-w sdks/js -w sparkle -w front -w front-spa -w front-api`, so `connectors` — and with it the
4.77.4 peer — is never installed, and `NODE_PATH=/app/front/node_modules` catches the resolution.
That makes correctness depend on which workspaces an install happens to include, and `NODE_PATH` is
only consulted *after* the `node_modules` walk-up, so it cannot override a wrong root copy. An
explicit dependency puts the right version first in the walk-up in every environment.

This restores what #29463 added (and the `Do not remove` comments in both OpenAI Responses clients
still refer to).

## Tests

Resolution from the bundle's directory, before and after:

```
# before
resolve("openai") from front-api/dist -> /dust/node_modules/openai      4.77.4  -> responses === undefined
# after
resolve("openai") from front-api/dist -> /dust/front-api/node_modules/openai  6.49.0  -> responses.create is a function
```

Verified the `text_as_cron_rule` path no longer fails on a local `front-api`.

## Risk

Low. Dependency-only change, no runtime code. `front-api` resolves 6.49.0 while `front` stays
pinned at 6.46.0 in the lockfile — same major, and the two are independent copies, as was already
the case between #29463 and #30770.

## Deploy Plan

None. Prod behaviour is unchanged; this removes the dependence on workspace-install layout.

### DIFF
--- a/front-api/package.json
+++ b/front-api/package.json
@@ -27,6 +27,7 @@
   "dependencies": {
     "@anthropic-ai/sdk": "^0.110.0",
     "@novu/framework": "^2.11.1",
+    "openai": "^6.46.0",
     "umzug": "3.8.3",
     "@hono/mcp": "^0.3.0",
     "@hono/node-server": "^1.19.10",
--- a/package-lock.json
+++ b/package-lock.json
@@ -545,6 +545,7 @@
         "@novu/framework": "^2.11.1",
         "hono": "^4.12.15",
         "jsonwebtoken": "^9.0.0",
+        "openai": "^6.46.0",
         "swagger-jsdoc": "^6.2.8",
         "umzug": "3.8.3",
         "zod": "^3.25.76",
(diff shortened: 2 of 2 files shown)


avervaet (approved):

davidebbo:
Fixed in https://github.com/dust-tt/dust/pull/31288. Sorry, didn't see yours!

PR #31286, opened 2026-08-27
Support multi-interaction checkpoint continuations

flvndvd:
## Description

Stateful conversation rendering assumed that the output appended between two model steps always belonged to a single interaction.

This does not hold when a tool enables a skill. The rendered continuation can contain an assistant tool call, its function result, and a synthetic user message carrying the newly enabled skill. That user boundary correctly creates another interaction, but the checkpointed path rejected it before applying the continuation.

This removes the stale single-interaction guard. The existing continuation logic already appends the first interaction to the checkpoint and adds subsequent interactions independently.

### DIFF
--- a/front/lib/api/assistant/conversation_rendering/conversation_window_core.ts
+++ b/front/lib/api/assistant/conversation_rendering/conversation_window_core.ts
@@ -331,11 +331,6 @@ async function buildConversationWindow(
   const interactions = groupMessagesIntoInteractions(
     prepared.messagesWithTokens
   );
-  if (source.kind === "checkpoint_continuation" && interactions.length > 1) {
-    return new Err(
-      new Error("A checkpoint continuation must belong to one interaction")
-    );
-  }
   const budgetForInteractions = input.allowedTokenCount - baseTokens;
   const pruningTargetCeiling =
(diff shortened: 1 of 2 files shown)


aubin-tchoi (approved):

front/lib/api/assistant/conversation_rendering/conversation_window_core.ts
```diff
@@ -331,11 +331,6 @@ async function buildConversationWindow(
   const interactions = groupMessagesIntoInteractions(
     prepared.messagesWithTokens
   );
-  if (source.kind === "checkpoint_continuation" && interactions.length > 1) {
-    return new Err(
```
aubin-tchoi, line 335: can we maybe add a small log in the case where it's not due to a skill? curious to know whether this happens at all

flvndvd: It's very hard to detect actually.

flvndvd: I'll change some of this logic to fetch only needed items, will try to add some tracing in there regarding this item

PR #31289, opened 2026-08-27
feat(front): persist the service tier that priced each run usage

Nils-Fedrigo:
## Description

We do not store the service tier a message was computed at, even though it changes the token
prices we are billed: `withFlexProcessing` asks OpenAI for `flex` on trigger / wake-up runs, and
`computeTokensCostForUsageInMicroUsd` then applies `FLEX_DISCOUNT_FACTOR` (0.5). The tier reached
`run_usages` only as a discount folded into `costMicroUsd` and was dropped on the floor otherwise.

This adds `run_usages."serviceTier"` and persists the provider-reported tier alongside the tokens
it priced, next to `isBatch` and `region` — the other two drivers of which price list applies.
Stacked on #30965.

**Why it cannot be recomputed downstream.** The requested tier is derivable from the request
(feature flag + `userMessageOrigin`), but the *billed* tier is not:

- Flex is best-effort. `OpenAIResponsesStreamClient.streamRaw` silently replays the request with
  `service_tier: "auto"` when flex is refused or times out before its first event, so a run that
  asked for flex can be billed at standard rates.
- The tier we persist is the one OpenAI reports on `response.completed`, normalized to
  `auto` / `flex` — the same value the cost was computed from.
- `openai_flex_processing` is a feature flag with no history, so even the request-side inputs are
  not reconstructable for past rows.

Backfill is therefore not possible; the column is null for rows written before this deploy and for
providers that report no tier.

## Tests

`front/lib/resources/run_resource.test.ts`:

- A pending attempt starts with `serviceTier: null`, and finalizing with a flex token usage
  persists `flex` — asserting the stored cost equals `computeTokensCostForUsageInMicroUsd(...,
  serviceTier: "flex")`, so the tier on the row is provably the one that priced it.
- A usage with no provider-reported tier stores null.

## Risk

Low. Additive nullable column, written on both the insert (`recordRunUsage`) and finalize
(`finalizePendingRunUsage`) paths; no read path changes behavior.

## Deploy Plan

Pre-deploy migration `20260827164821_add_service_tier_to_run_usages.sql` (`ADD COLUMN`, nullable)
runs before the code.


Nils-Fedrigo:
Reopening from a clean branch — the diff was computed against a stale base.

Nils-Fedrigo:
sorry for big branch 🙏

PR #31290, opened 2026-08-27
feat(workspace-analytics): replace the legacy top-N tools with two consumption rankings

achilleburah:
Stacked on #31273 — review that one first, this diff is against its branch.

## Description

Second step of moving `@analyst` off `front.agent_message_analytics`. #31273 added the credits ranking; this one adds the two count rankings and retires the seven legacy top-N tools they replace.

- `get_top_entities_by_message_count` — agents, members, models, sources, API keys, groups, tags, conversations. Replaces `get_top_agents`, `get_top_users`, `get_top_models`, `get_top_agent_tags` and `get_source_breakdown`.
- `get_top_entities_by_execution_count` — tools and skills. Replaces `get_top_tools` and `get_top_skills`.

Both are the same `fetchConsumptionTopGroups` call as the credits tool with `rankBy: "count"`, so all three now share one `renderRanking`. That is where the −757 comes from: seven hand-rolled handlers and their aggregations collapse into one.

`rankBy` picks the ordering from the dimension's unit. Credits and invocation counts are free — a sum Elasticsearch already has, and the bucket's own `doc_count`. Ordering a message dimension by count is the expensive one: it sorts on the distinct-message cardinality, which every shard computes for every candidate bucket. That also makes the missing `precision_threshold` on that cardinality a real bug rather than a display detail, so it now uses the module's 40 000 like everything else.

Two counts change meaning, both narrower and less biased than before:

- **Skills.** The old count came from `AgentMessageSkillModel`, which `snapshotConversationSkillsForMessage` copies onto every message in the conversation — so a skill attached once scored on every later message whether or not it did anything. Consumption counts tool calls actually attributed to the skill. A skill that never calls a tool no longer appears.
- **Messages.** The consumption indexer skips failed and unbilled messages, so every message count is now billed messages.

## Tests

- `scope.test.ts`: the message and invocation dimension tuples cover every top dimension exactly once, and each sits under the unit `CONSUMPTION_TOP_DIMENSION_UNIT` gives it. Add a dimension and forget to file it and this fails.
- `top.test.ts`: the `order` clause each dimension/rankBy pair emits. Seeded data ranks the same either way, so this asserts on the aggregation rather than the result.
- `bm25_tool_search.test.ts`: the retired tools' queries now route to the two new ones.

## Risk

Medium. Nine tools become four, and two counts change definition as described. Callers are `@analyst` only — no endpoint or page reads these.

## Deploy Plan

Merge after #31273. Second of five: credits ranking → count rankings → overview → credit timeseries → usage timeseries.

### DIFF
--- a/front/lib/api/actions/servers/bm25_tool_search.test.ts
+++ b/front/lib/api/actions/servers/bm25_tool_search.test.ts
@@ -1201,6 +1201,7 @@ const QUERIES: LabeledQuery[] = [
   {
     query: "search the web for the latest AI research papers",
     expected: "web_search_&_browse.websearch",
+    maxRank: 2, // clari_copilot.get_call_details dilutes shared-token IDF
   },
   {
     query: "google this topic for me",
--- a/front/lib/api/actions/servers/bm25_tool_search.test.ts
+++ b/front/lib/api/actions/servers/bm25_tool_search.test.ts
@@ -1079,19 +1079,7 @@ const QUERIES: LabeledQuery[] = [
     maxRank: 2,
   },
   {
-    query: "show me the top 10 most active agents this month",
-    expected: "workspace_analytics.get_top_entities_by_message_count",
-  },
-  {
     query: "who are the most active users this month",
-    expected: "workspace_analytics.get_top_entities_by_message_count",
-  },
-  {
-    query: "rank workspace members by messages sent",
-    expected: "workspace_analytics.get_top_entities_by_message_count",
-  },
-  {
-    query: "rank member groups by message volume",
     expected: "workspace_analytics.get_top_entities_by_message_count",
   },
   {
@@ -1103,10 +1091,6 @@ const QUERIES: LabeledQuery[] = [
     query: "which models did the workspace use most this month",
     expected: "workspace_analytics.get_top_entities_by_message_count",
     maxRank: 5,
-  },
-  {
-    query: "list the agent tags",
-    expected: "workspace_analytics.get_top_entities_by_message_count",
   },
   {
     query:
--- a/front/lib/api/actions/servers/bm25_tool_search.test.ts
+++ b/front/lib/api/actions/servers/bm25_tool_search.test.ts
@@ -1075,24 +1075,38 @@ const QUERIES: LabeledQuery[] = [
   // --- workspace_analytics ---
   {
     query: "which agents are used most in the workspace",
-    expected: "workspace_analytics.get_top_agents",
-    maxRank: 2, // get_top_tools collides on "most used"
+    expected: "workspace_analytics.get_top_entities_by_message_count",
+    maxRank: 2,
   },
   {
     query: "show me the top 10 most active agents this month",
-    expected: "workspace_analytics.get_top_agents",
+    expected: "workspace_analytics.get_top_entities_by_message_count",
   },
   {
     query: "who are the most active users this month",
-    expected: "workspace_analytics.get_top_users",
+    expected: "workspace_analytics.get_top_entities_by_message_count",
   },
   {
     query: "rank workspace members by messages sent",
-    expected: "workspace_analytics.get_top_users",
+    expected: "workspace_analytics.get_top_entities_by_message_count",
+  },
+  {
+    query: "rank member groups by message volume",
+    expected: "workspace_analytics.get_top_entities_by_message_count",
+  },
+  {
+    query: "where do workspace messages come from - slack, api, or browser",
+    expected: "workspace_analytics.get_top_entities_by_message_count",
+    maxRank: 10,
+  },
+  {
+    query: "which models did the workspace use most this month",
+    expected: "workspace_analytics.get_top_entities_by_message_count",
+    maxRank: 5,
   },
   {
     query: "list the agent tags",
-    expected: "workspace_analytics.get_top_agent_tags",
+    expected: "workspace_analytics.get_top_entities_by_message_count",
   },
   {
     query:
@@ -1104,30 +1118,24 @@ const QUERIES: LabeledQuery[] = [
     expected: "workspace_analytics.get_agent_details",
   },
   {
-    query: "which skills are executed most in the workspace",
-    expected: "workspace_analytics.get_top_skills",
-  },
-  {
-    query: "what are the top MCP tools used by agents",
-    expected: "workspace_analytics.get_top_tools",
-  },
-  {
-    query: "where do workspace messages come from - slack, api, or browser",
-    expected: "workspace_analytics.get_source_breakdown",
-    maxRank: 10,
-  },
-  {
-    query: "which models did the workspace use most this month",
-    expected: "workspace_analytics.get_top_models",
-    maxRank: 5,
-  },
-  {
     query: "how many AWU credits did the workspace consume this month",
     expected: "workspace_analytics.get_credit_usage",
   },
   {
     query: "break down credit spending by agent",
     expected: "workspace_analytics.get_credit_usage",
+  },
+  {
+    query: "which skills are executed most in the workspace",
+    expected: "workspace_analytics.get_top_entities_by_execution_count",
+  },
+  {
+    query: "what are the top MCP tools used by agents",
+    expected: "workspace_analytics.get_top_entities_by_execution_count",
+  },
+  {
+    query: "how many times did each integration run",
+    expected: "workspace_analytics.get_top_entities_by_execution_count",
   },
   {
     query: "which agents cost the most credits this month",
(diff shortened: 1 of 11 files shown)


front/lib/api/actions/servers/workspace_analytics/tools/index.ts
```diff
@@ -982,72 +559,37 @@ const handlers: ToolHandlers<typeof WORKSPACE_ANALYTICS_TOOLS_METADATA> = {
...
-      includePreviousCredits: false,
-    });
-
-    if (result.isErr()) {
-      return new Err(
-        new MCPError(
-          `Failed to rank ${dimension} by credits: ${result.error.message}`
-        )
-      );
+  get_top_entities_by_execution_count: async (
```
avervaet, line 626: Looks like this new path bypasses the existing SKILL_MANAGEMENT_SERVER_NAME exclusion used by fetchConsumptionTopTools, so skill_management can appear as a top MCP tool and displace real integrations.

front/lib/api/actions/servers/workspace_analytics/tools/index.ts
```diff
@@ -118,286 +111,80 @@ function renderExecutionSeries<
...
+    input: TimeWindowInput & ConsumptionFilterInput;
+  }
+): Promise<ToolHandlerResult> {
+  const window = resolveTimeWindow(input);
+  if (window.isErr()) {
+    return new Err(new MCPError(window.error, { tracked: false }));
+  }
+  const { filter, extraFilters } = toConsumptionScope(input);
+
+  const result = await fetchConsumptionTopGroups(auth, {
```
avervaet, line 201: nit: Could we add an includeTotalCount: false/pagination: false option for this adapter? These tools never expose hasMore or offset, so they still pay for the global total_count cardinality aggregation unnecessarily.

front/lib/api/actions/servers/workspace_analytics/tools/index.ts
```diff
@@ -118,286 +111,80 @@ function renderExecutionSeries<
...
+  const { groups, totalCredits } = result.value;
 
+  if (groups.length === 0) {
     return new Ok([
       {
         type: "text" as const,
-        text:
-          `Top users for ${label} (${tz}), most active first:\n` +
-          lines.join("\n"),
+        text: `No ${dimension} activity recorded for ${label} (${tz}).`,
```
avervaet: nit: Could the empty-state wording depend on rankBy? The credits tool now says “no activity” instead of “no consumption,” which can be misleading.

avervaet (approved):
LGTM 🔥

front/lib/api/actions/servers/workspace_analytics/tools/index.ts
achilleburah: 🙏

front/lib/api/actions/servers/workspace_analytics/tools/index.ts
```diff
@@ -118,286 +112,94 @@ function renderExecutionSeries<
...
-      timezone,
-      source,
-      agentIds,
-      userIds,
-      agentTagIds,
-      modelIds,
+      type: "text" as const,
+      text:
+        `Top ${dimension} for ${label} (${tz}), ` +
+        `${rankBy === "credits" ? "most expensive" : `most ${unit}s`} first:\n` +
```
flvndvd:
```suggestion
        `${rankBy === "credits" ? "most expensive" : `most ${pluralize(unit)}`} first:\n` +
```

front/lib/api/actions/servers/workspace_analytics/tools/index.ts
```diff
@@ -118,286 +112,94 @@ function renderExecutionSeries<
...
-      endDate,
-      timezone,
-      source,
-      agentIds,
-      userIds,
-      agentTagIds,
-      modelIds,
+      type: "text" as const,
+      text:
+        `Top ${dimension} for ${label} (${tz}), ` +
```
flvndvd: As this gets lots of conditional, mind splitting it into distinct cases to ease readability?

front/lib/api/actions/servers/workspace_analytics/tools/index.ts
```diff
@@ -982,72 +574,37 @@ const handlers: ToolHandlers<typeof WORKSPACE_ANALYTICS_TOOLS_METADATA> = {
...
   },
-  get_top_entities_by_credits: async (
+  get_top_entities_by_message_count: async (
     { dimension, limit, ...input },
     { auth }
   ) => {
-    const deniedError = workspaceManagerGuard(auth);
-    if (deniedError) {
-      return new Err(deniedError);
+    const denied = workspaceManagerGuard(auth);
```
flvndvd: To align on your previous PR 😛 , `deniedError`?

front/lib/api/actions/servers/bm25_tool_search.test.ts
```diff
@@ -1075,60 +1075,55 @@ const QUERIES: LabeledQuery[] = [
...
-    query: "how many AWU credits did the workspace consume this month",
-    expected: "workspace_analytics.get_credit_usage",
+    query: "list the agent tags",
+    expected: "workspace_analytics.get_top_entities_by_message_count",
   },
   {
     query: "break down credit spending by agent",
     expected: "workspace_analytics.get_credit_usage",
   },
+  {
```
flvndvd, line 1112: No need to over index over this.

front/lib/api/analytics/consumption/top.test.ts
```diff
@@ -878,12 +878,46 @@ describe("consumption top rankings", () => {
...
-// The two options the analytics tools rely on and the page never sets.
 describe("fetchConsumptionTopGroups options", () => {
   afterEach(() => {
     vi.restoreAllMocks();
   });
 
+  it.each([
+    {
+      name: "credits sort on the credit sum",
+      dimension: "agent" as const,
```
flvndvd, line 889: You very much love `as const` 🙃 !

achilleburah: that's was the existing pattern so I kept it in most places 😅

front/lib/api/actions/servers/bm25_tool_search.test.ts
achilleburah: Many iterations to keep the previous "3 tests -> 3 tools". Now it's "3 tests -> 1 tool". Plus added a few to cover the new available features. 8 is maybe to much though, agreed. Reduced to 4.

front/lib/api/analytics/consumption/top.test.ts
flvndvd: Ahah not everything is good in the codebase, but let's drop this not-so-idiomatic TS.

front/lib/resources/skill/code_defined/global/workspace_analytics.ts
```diff
@@ -46,9 +46,13 @@ export const workspaceAnalyticsSkill = {
     "agent's spend evolve'), set breakdownBy on get_credit_timeseries — one " +
     "call returns the top groups plus an 'other' series. Do not make one " +
     "filtered call per agent, user or model.\n" +
-    "- To scope any tool to specific models, call get_top_models first to get " +
-    "the exact model ids in use, then pass them as modelIds — never guess a " +
-    "model id from its display name.\n" +
+    "- For volume rather than spend — who is most active, which agents or " +
+    "models get used most, where messages come from — call " +
+    "get_top_entities_by_message_count with that dimension, and " +
```
flvndvd: Same here we tend to use constant for tool name to avoid the drift.

front/lib/resources/skill/code_defined/global/workspace_analytics.ts
```diff
@@ -46,9 +46,13 @@ export const workspaceAnalyticsSkill = {
     "agent's spend evolve'), set breakdownBy on get_credit_timeseries — one " +
     "call returns the top groups plus an 'other' series. Do not make one " +
     "filtered call per agent, user or model.\n" +
-    "- To scope any tool to specific models, call get_top_models first to get " +
-    "the exact model ids in use, then pass them as modelIds — never guess a " +
-    "model id from its display name.\n" +
+    "- For volume rather than spend — who is most active, which agents or " +
```
flvndvd: Lots of emdashes ✂️

flvndvd (approved):

front/lib/resources/skill/code_defined/global/workspace_analytics.ts
achilleburah: removed

PR #31291, opened 2026-08-27
feat(front): persist the service tier that priced each run usage

Nils-Fedrigo:
## Description

We do not store the service tier a message was computed at, even though it changes the token
prices we are billed: `withFlexProcessing` asks OpenAI for `flex` on trigger / wake-up runs, and
`computeTokensCostForUsageInMicroUsd` then applies `FLEX_DISCOUNT_FACTOR` (0.5). The tier reached
`run_usages` only as a discount folded into `costMicroUsd` and was dropped on the floor otherwise.

This adds `run_usages."serviceTier"` and persists the provider-reported tier alongside the tokens
it priced, next to `isBatch` and `region` — the other two drivers of which price list applies.
Stacked on #30965.

**Why it cannot be recomputed downstream.** The requested tier is derivable from the request
(feature flag + `userMessageOrigin`), but the *billed* tier is not:

- Flex is best-effort. `OpenAIResponsesStreamClient.streamRaw` silently replays the request with
  `service_tier: "auto"` when flex is refused or times out before its first event, so a run that
  asked for flex can be billed at standard rates.
- The tier we persist is the one OpenAI reports on `response.completed`, normalized to
  `auto` / `flex` — the same value the cost was computed from.
- `openai_flex_processing` is a feature flag with no history, so even the request-side inputs are
  not reconstructable for past rows.

Backfill is therefore not possible; the column is null for rows written before this deploy and for
providers that report no tier.

## Tests

`front/lib/resources/run_resource.test.ts`:

- A pending attempt starts with `serviceTier: null`, and finalizing with a flex token usage
  persists `flex` — asserting the stored cost equals `computeTokensCostForUsageInMicroUsd(...,
  serviceTier: "flex")`, so the tier on the row is provably the one that priced it.
- A usage with no provider-reported tier stores null.

## Risk

Low. Additive nullable column, written on both the insert (`recordRunUsage`) and finalize
(`finalizePendingRunUsage`) paths; no read path changes behavior.

## Deploy Plan

Pre-deploy migration `20260827164821_add_service_tier_to_run_usages.sql` (`ADD COLUMN`, nullable)
runs before the code.


front/lib/resources/storage/models/runs.ts
```diff
@@ -96,6 +97,11 @@ export class RunUsageModel extends WorkspaceAwareModel<RunUsageModel> {
 
   declare costMicroUsd: number;
   declare isBatch: boolean;
+  // Processing tier the provider reported having billed the response at. Like
+  // isBatch and region, it changes the token prices we are charged, so it is
+  // persisted to keep run usages reconcilable against the provider bill. Null
+  // when the provider does not report a tier (or for legacy rows).
```
Nils-Fedrigo:
```suggestion
```

front/lib/resources/run_resource.ts
```diff
@@ -51,6 +52,9 @@ export interface RunUsageType {
   cacheCreationTokens?: number | null;
   costMicroUsd: number;
   isBatch: boolean;
+  // Provider-reported processing tier this response was billed at. Like isBatch, it changes the
+  // token prices applied, so it is persisted alongside the tokens it priced.
```
Nils-Fedrigo:
```suggestion
```

front/lib/resources/run_resource.ts
```diff
@@ -123,6 +127,8 @@ export class RunResource extends BaseResource<RunModel> {
           cacheCreationTokens: null,
           costMicroUsd: 0,
           isBatch: false,
+          // Only known once the provider reports the tier it served on.
```
Nils-Fedrigo:
```suggestion
```

id13 (approved):

front/lib/resources/storage/models/runs.ts
```diff
@@ -159,6 +161,11 @@ RunUsageModel.init(
       defaultValue: false,
       allowNull: false,
     },
+    serviceTier: {
+      type: DataTypes.STRING,
+      allowNull: true,
+      defaultValue: null,
```
flvndvd: Let's make sure we do backfill all previous runs 🙏  and drop the `allowNull: true`.

front/lib/resources/run_resource.ts
```diff
@@ -483,6 +488,7 @@ export class RunResource extends BaseResource<RunModel> {
           cacheCreationTokens: cacheCreationTokens ?? null,
           costMicroUsd,
           isBatch,
+          serviceTier: serviceTier ?? null,
```
flvndvd: Isn't `serviceTier` an immutable attribute which is known before the LLM call is being done? Cause if it is, then the resource layer should materialize it, and only accept it when creating pending run usages.

Nils-Fedrigo: Most serviceTiers are going to be null because we set it to null if it's not explicit

Nils-Fedrigo: And we write the resolved serviceTier to be sure in case we needed to upgrade to auto to make it in time

front/lib/resources/storage/models/runs.ts
Nils-Fedrigo: We have no previous run, for now serviceTier is not used and wasn't in the past

front/lib/resources/run_resource.test.ts
```diff
@@ -65,6 +66,89 @@ describe("RunResource reasoning token usage", () => {
...
+        inferenceProvider: "openai-responses",
+        modelId: GPT_5_MINI_MODEL_CONFIG.modelId,
+        providerId: GPT_5_MINI_MODEL_CONFIG.providerId,
+        region: "global",
+        usageType: USAGE_TYPE_USER,
+      }
+    );
+
+    // The pending attempt predates the provider response, so it carries the
+    // standard tier until the response says otherwise.
```
flvndvd, line 90: Can the response be different than the original value? Like can you ask for "flex" but get "default"?

flvndvd (approved):

front/lib/resources/run_resource.test.ts
Nils-Fedrigo: not on the OAI side but on our side some triggers can use default if we want to rush them

flvndvd: Ok but I presume we know the service tier before submitting the API call, right?

PR #31292, opened 2026-08-27
feat(front): key the Slack workflow whitelist on spaces instead of groups

adrsimon:
## Stack

Self-serve Slack workflow whitelisting, split in four:

1. This PR: the whitelist speaks spaces instead of Dust groups.
2. #31293, admin endpoints to list, allow and revoke, gated on credit-priced plans.
3. #31294, the workflow consumption overview endpoint behind the credits card.
4. #31295, the automations page tab, table and allow dialog.

Only the last one is visible to admins. The first three land dead code on purpose.

## Description

Connectors stores the Dust group ids a Slack workflow may reach, which is the wrong currency for an admin: they grant access to spaces, not groups. The whitelist now takes and reports space ids, and a space is represented on the wire by its own `regular_auto` member group.

That group is the only one that can stand in for a space. It is created and deleted with the space and `MANAGEABLE_GROUP_KINDS` will not let an admin attach it anywhere else, so it grants exactly that space and keeps doing so. Storing every group that grants the space, which is what the first draft of this did, snapshots a set that goes stale: a provisioned group is shared, so as soon as someone attaches it to a second space the workflow silently gains that space, and as soon as someone detaches it the whitelist keeps granting whatever else it reaches. Neither is visible from the space's own settings.

Poke's plugin therefore picks spaces instead of groups and the table shows space names, `SLACK_WORKFLOW_GROUP_KINDS` and the group validation it fed are gone, and the audit event records `space_names`. `SPACE_MEMBER_GRANT_TYPE` joins `SPACE_EDITOR_GRANT_TYPE` so the code writing the grant and the code reading it back agree on which grant marks a member.

## Tests

New `front/lib/api/slack/summoning_whitelist.test.ts`: allowing a space whitelists its member group and the global group and leaves out a provisioned group attached to that same space, a space from another workspace is rejected, and listing resolves a stored member group back to its space while ignoring stored groups that are not one.

The same file also pins what the stored groups actually buy, through the real header path (`Authenticator.fromKey` with the whitelist's group ids, then `getAgentConfigurationsForView`): a space's agents are summonable only when its member group is in the set, that still holds once the space's access moves to provisioned groups, and attaching one of those provisioned groups to a second space widens what a provisioned-group whitelist would reach while the member group's reach stays put. That last assertion is the drift this PR exists to remove.

`front-api/routes/poke/workspaces/[wId]/slack-workflows.test.ts` moves onto the shared `front/tests/utils/slack_workflows.ts` helpers and seeds a real `slack_bot` data source instead of mocking `DataSourceResource`.

## Risk

Rows written by the old poke plugin hold arbitrary groups, so they list no spaces until they are allowed again through the plugin. Their access does not change: connectors keeps sending the group ids it already stored. There were under ten such rows last time I looked, and re-running the plugin fixes each one.

Access on new rows narrows rather than widens, since a picked space no longer drags in what its provisioned groups reach elsewhere.

## Deploy Plan

Deploy front and front-api, then register the changed audit schema:

- [ ] `npx tsx front/admin/register_audit_log_schemas.ts --execute --changed`

### DIFF
--- a/front-api/routes/poke/workspaces/[wId]/slack-workflows.test.ts
+++ b/front-api/routes/poke/workspaces/[wId]/slack-workflows.test.ts
@@ -12,38 +12,34 @@ vi.mock("@app/lib/api/audit/workos_audit", async () => {
 });
 
 import { emitAuditLogEvent } from "@app/lib/api/audit/workos_audit";
-import { DataSourceResource } from "@app/lib/resources/data_source_resource";
 import { createPrivateApiMockRequest } from "@app/tests/utils/generic_private_api_tests";
+import {
+  connectSlackBot,
+  createSpaceWithMemberGroup,
+  mockRevokeSlackWorkflow,
+  mockSummoningWhitelist,
+  SLACK_BOT_CONNECTOR_ID,
+  SLACK_WORKFLOW_BOT_NAME,
+  SLACK_WORKFLOW_CREATED_AT_MS,
+} from "@app/tests/utils/slack_workflows";
 import { ConnectorsAPI } from "@app/types/connectors/connectors_api";
-import { Err, Ok } from "@app/types/shared/result";
+import { Err } from "@app/types/shared/result";
 import { honoApp } from "@front-api/app";
-
-const CONNECTOR_ID = "1234";
-const BOT_NAME = "Weekly report";
-const CREATED_AT = 1756166400000;
-
-function mockSlackBotDataSource({ connected }: { connected: boolean }) {
-  vi.spyOn(DataSourceResource, "listByConnectorProvider").mockImplementation(((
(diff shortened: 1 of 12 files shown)


flvndvd (commented):
It's great visually BUT this can create some confusion as the DB stores the associated groups at one point in-time. Let's take the following example:
- Admin creates Space A (which at the time is associated to group 1)
- Overtime Admin adds new provisioned groups 2, and 3 to the Space A
- In DB, workflow will only be tagged with group 1.

adrsimon:
Superseded by #31667/#31668, which resolve the whitelist against space ids at run time instead of persisting a space's group id at write time. Rebasing #31293/#31294/#31295 onto main directly.

PR #31293, opened 2026-08-27
feat(front-api): admin endpoints to allow and revoke Slack workflows

adrsimon:
## Stack

Lets workspace admins self-serve Slack workflow permissions from the automations page instead of filing a support ticket. Three PRs: admin endpoints to list, allow and revoke workflows; an endpoint exposing their credit consumption; and the UI tab that ties both together. All three gate on a credit-priced plan, so legacy workspaces keep going through poke. Built on top of the space-based summoning whitelist. Stack: #31293 (this PR) → #31294 → #31295.

## Description

Admin endpoints to allow, list and revoke a Slack workflow bot per space. POST validates the spaces against the workspace, always includes the Company Space, and forwards the result to connectors. Both write paths emit the same audit events as the poke plugin. Gated behind the credit-priced plan check. No UI calls these endpoints yet.

## Tests

`front-api/routes/w/[wId]/slack-workflows/index.test.ts`, ten cases: non-admin and legacy plan rejections, space resolution on the listing, workspace with no Slack bot, allow with the Company Space appended plus its audit event, empty name and unknown space id rejections, revoke, unknown bot name.

## Risk

New endpoints, nothing calls them. The gate is the part worth a second look: a workspace that is not credit-priced must keep getting 403 on all three verbs, which the tests cover.

## Deploy Plan

Deploy front-api.


aubin-tchoi (approved):
LGTM

PR #31294, opened 2026-08-27
feat(front-api): expose Slack workflow consumption

adrsimon:
## Description

Second PR of the stack (#31293). Endpoint exposing the credits a workspace spent on Slack workflow messages, for the automations tab. Reads the raw `context_origin` field instead of `normalized_origin`, since normalization folds Slack workflow traffic into regular Slack. Same admin and plan gate as #31293. The consumption seed factory now produces Slack workflow messages so dev workspaces show a non-zero card.

## Tests

`front-api/routes/w/[wId]/slack-workflows/overview.test.ts`: non-admin and legacy plan rejections, the credits aggregation over a seeded period, and an Elasticsearch failure surfacing as 500.

## Risk

Read-only aggregation on an existing index, nothing calls it yet. Worst case the card shows a wrong number.

## Deploy Plan

Deploy front-api.

### DIFF
--- a/front-api/routes/w/[wId]/slack-workflows/overview.ts
+++ b/front-api/routes/w/[wId]/slack-workflows/overview.ts
@@ -1,0 +1,56 @@
+import { resolveConsumptionPeriod } from "@app/lib/api/analytics/consumption/period";
+import {
+  ConsumptionPeriodSchema,
+  toConsumptionPeriodInput,
+} from "@app/lib/api/analytics/consumption/schema";
+import type { GetSlackWorkflowsOverviewResponse } from "@app/lib/api/analytics/slack_workflows/overview";
+import { fetchSlackWorkflowsOverview } from "@app/lib/api/analytics/slack_workflows/overview";
+import logger from "@app/logger/logger";
+import { workspaceApp } from "@front-api/middlewares/ctx";
+import { ensureIsAdmin } from "@front-api/middlewares/ensure_role";
+import { apiError, type HandlerResult } from "@front-api/middlewares/utils";
+import { validate } from "@front-api/middlewares/validator";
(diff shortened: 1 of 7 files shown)


front-api/routes/w/[wId]/slack-workflows/overview.ts
```diff
@@ -0,0 +1,56 @@
...
+import type { GetSlackWorkflowsOverviewResponse } from "@app/lib/api/analytics/slack_workflows/overview";
+import { fetchSlackWorkflowsOverview } from "@app/lib/api/analytics/slack_workflows/overview";
+import logger from "@app/logger/logger";
+import { workspaceApp } from "@front-api/middlewares/ctx";
+import { ensureIsAdmin } from "@front-api/middlewares/ensure_role";
+import { apiError, type HandlerResult } from "@front-api/middlewares/utils";
+import { validate } from "@front-api/middlewares/validator";
+import { withCreditPricedPlan } from "@front-api/middlewares/with_credit_priced_plan";
+
+export type { GetSlackWorkflowsOverviewResponse };
```
Comment (aubin-tchoi) [inline front-api/routes/w/[wId]/slack-workflows/overview.ts]: nit: can't we import directly from the original file?

front/lib/api/analytics/slack_workflows/overview.ts
```diff
@@ -0,0 +1,83 @@
...
+        },
+      },
+      size: 0,
+    }
+  );
+  if (searchResult.isErr()) {
+    return searchResult;
+  }
+
+  const aggregations = searchResult.value.aggregations;
```
aubin-tchoi: nit nit:
```suggestion
  const { aggregations } = searchResult.value;
```

aubin-tchoi (approved):
LGTM

PR #31297, opened 2026-08-27
Refactoring: remove the WorkspacePlanLimitOverrideResource and move methods to WorkspaceResource

fabiencelier:
## Description

WorkspacePlanLimitOverrideResource has been removed in favor of static methods inside WorkspaceResource
Method are static because we don't have direct access to a WorkspaceResource, we have a LightWorkspaceType coming either from auth directly or from many chained calls.
We could introduce a new `WorkspaceResource.getNonNullableWorkspaceResource(): WorkspaceResource` but that would only solve 2 of the 4 methods as other methods don't have direct access to an auth either.
I don't think it's worth introducing a dual way of fetching the workspace from auth for that.

About caching:
Caching of the plan limit override is done by the suscription cache where it is inlined.
Writing a new subscription override invalidates the cache.

## Tests

added some unit tests for the resources methods

## Risk

low risk, easy to revert if needed

## Deploy Plan

deploy front.
No actual change to models, no migration required


flvndvd (approved):
🙏

PR #31298, opened 2026-08-27
feat(front-api): check that bundled runtime imports are declared deps

avervaet:
## Context

`front-api` bundles most of `front`'s code via the `@app/*` alias.

esbuild externalizes bare imports, even ones only reached through that bundled code.

Those imports resolve at runtime from whatever's hoisted in `node_modules`.

If the package isn't declared as a `front-api` dependency, resolution is a matter of luck.

That's what happened with `openai` in #30770: removed as "unused", worked in prod, broke locally. Fixed in #31284.

## Fix

Adds a CI check: `front-api/scripts/check-bundled-deps.ts`.

It builds `front-api`, reads the esbuild metafile, and lists every externalized import.

It fails if one isn't declared in `front-api/package.json`.

## Implications

`front-api` already relies on ~120 such packages without declaring them.

Declaring all of them now is a separate, bigger migration.

So today's list is seeded into an allowlist (`check-bundled-deps.allowlist.json`), including `openai`.

New undeclared imports going forward will fail CI. The allowlist should shrink, not grow.


avervaet:

- #31298 ◀

davidebbo (approved):
LGTM. Let's monitor how long the step actually takes in Prod over the next few build.

PR #31299, opened 2026-08-27
Persist Frames v2 function artifacts

fontanierh:
## Description

Persist prebuilt function bundles and extracted schemas with each immutable Frames v2 publication. Publication now requires an exact match between manifest functions and supplied artifacts, uploads all source and artifacts first, then writes the manifest commit marker.

## Tests

Added focused coverage for artifact paths and payloads, missing, extra, and duplicate artifacts, and failed artifact writes without activation.

## Risk

Low. There are no production callers yet. Failed uploads can leave uncommitted objects under a fresh publication ID, but readers cannot observe them without the manifest.

## Deploy Plan

Standard deploy.


davidebbo (commented):
LGTM

PR #31300, opened 2026-08-27
feat(workspace-analytics): replace get_credit_usage with a consumption overview

achilleburah:
Stacked on #31290, itself on #31273 — review those first, this diff is against #31290's branch.

## Description

Third step of moving `@analyst` off `front.agent_message_analytics`. `get_credit_usage` is the last snapshot tool on the legacy index, and it did two jobs: a window's total credits, and a top-N breakdown by agent, user or model. #31273 already replaced the breakdown, so what is left is the total — and the Analytics page already has a better answer for it.

`get_consumption_overview` returns the page's header block: total credits, messages, active and total members, the top agent, and when consumption was last recorded. One call replaces what previously took `get_credit_usage` plus a follow-up.

`fetchConsumptionOverview` only took a `ConsumptionPeriodInput` (`"cycle"` or `"days"`), which cannot express the arbitrary start/end range these tools accept. It now delegates to `fetchConsumptionOverviewForPeriod`, which takes a resolved period. The page keeps its existing entry point untouched.

The credit cap comes out of the tool's answer rather than being reported as null: a cap covers a billing cycle, so an arbitrary window has nothing to compare against. `withCreditCap` makes that explicit at the call site, and the tool description points admins at the Usage page for it.

Also fixes a live bug: `active_members` counted with no `precision_threshold`, so it defaulted to 3 000 while every other cardinality in the module uses 40 000. Any workspace above that in a period was showing an approximate member count on the Analytics page.

## Tests

- `bm25_tool_search.test.ts`: "how many AWU credits did the workspace consume this month" now routes to the overview, and "break down credit spending by agent" drops the rank-2 relaxation #31273 added — the tool it was competing with is gone.

## Risk

Low. One tool out, one in, and the page's own entry point is unchanged. The `precision_threshold` fix moves the page's active-member count towards the correct number for large workspaces.

## Deploy Plan

Merge after #31290. Third of five: credits ranking → count rankings → overview → credit timeseries → usage timeseries.

### DIFF
--- a/front/lib/api/actions/servers/workspace_analytics/metadata.ts
+++ b/front/lib/api/actions/servers/workspace_analytics/metadata.ts
@@ -26,27 +26,9 @@ const getAgentDetailsSchema = {
     ),
 };
 
-const getCreditUsageSchema = {
+const getConsumptionOverviewSchema = {
   ...timeWindowSchemaShape,
-  ...usageFilterSchema,
-  groupBy: z
-    .enum(["agent", "user", "model", "none"])
-    .optional()
-    .describe(
-      "Break the estimated credits down by top 'agent', 'user' or 'model' " +
-        "(the model that answered the message), or 'none' (default) for the " +
-        "workspace total only."
-    ),
-  limit: z
-    .number()
-    .int()
-    .positive()
-    .max(MAX_RESULTS)
-    .optional()
-    .describe(
-      `When grouping, the maximum number of rows to return ` +
-        `(default ${DEFAULT_RESULTS}, max ${MAX_RESULTS}).`
-    ),
+  ...consumptionFilterSchema,
 };
 
 const getCreditTimeseriesSchema = {
@@ -100,6 +82,8 @@ export const GET_TOP_ENTITIES_BY_MESSAGE_COUNT_TOOL_NAME =
   "get_top_entities_by_message_count" as const;
 export const GET_TOP_ENTITIES_BY_EXECUTION_COUNT_TOOL_NAME =
   "get_top_entities_by_execution_count" as const;
+export const GET_CONSUMPTION_OVERVIEW_TOOL_NAME =
+  "get_consumption_overview" as const;
 
 const rankingLimitSchema = z
   .number()
@@ -200,24 +184,21 @@ export const WORKSPACE_ANALYTICS_TOOLS_METADATA = [
     freeUsage: true,
   },
   {
-    name: "get_credit_usage",
+    name: GET_CONSUMPTION_OVERVIEW_TOOL_NAME,
     description:
-      "Estimate AWU credit consumption over a time window (defaults to the " +
-      "current calendar month), optionally broken down by the top agents, " +
-      "users or models. Credits combine model compute and tool usage, " +
-      "mirroring how billing computes them. IMPORTANT: these figures are " +
-      "ESTIMATES derived from usage logs — always tell the user they are " +
-      "approximate and point them to the workspace Usage page for exact, " +
-      "billed credit amounts. Optionally filter by source (context_origin), " +
-      "agent, user, tag, or model — e.g. filter by tag to attribute credits " +
-      "to all agents with a given tag, or group by model to see which models " +
-      "drive spend. Admin-only.",
-    schema: getCreditUsageSchema,
+      "Summarize the workspace's headline figures for a time window " +
+      "(defaults to the current calendar month): total credits consumed, " +
+      "messages, active and total members, and the top agent. Use this to " +
+      "answer 'how are we doing this month' or 'how many credits did we " +
+      `consume' in one call, and ${GET_TOP_ENTITIES_BY_CREDITS_TOOL_NAME} to ` +
+      "attribute that total. For the credit cap and how much of it is left, " +
+      "point the admin at the workspace Usage page.",
+    schema: getConsumptionOverviewSchema,
     stake: "never_ask",
     eager: true,
     displayLabels: {
-      running: "Estimating credit usage",
-      done: "Estimated credit usage",
+      running: "Retrieving consumption overview",
+      done: "Retrieved consumption overview",
     },
     toolCostCategory: "basic",
     freeUsage: true,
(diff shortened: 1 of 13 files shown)


front/lib/api/actions/servers/workspace_analytics/metadata.ts
```diff
@@ -212,23 +195,22 @@ export const WORKSPACE_ANALYTICS_TOOLS_METADATA = [
...
-      "mirroring how billing computes them. IMPORTANT: these figures are " +
-      "ESTIMATES derived from usage logs — always tell the user they are " +
-      "approximate and point them to the workspace Usage page for exact, " +
-      "billed credit amounts. Optionally filter by source (context_origin), " +
-      "agent, user, tag, or model — e.g. filter by tag to attribute credits " +
-      "to all agents with a given tag, or group by model to see which models " +
-      "drive spend. Admin-only.",
-    schema: getCreditUsageSchema,
+      "Summarize the workspace's headline figures for a time window " +
+      "(defaults to the current calendar month): total credits consumed, " +
```
achilleburah: harmonizing default to current cycle in following PR

front/lib/api/analytics/consumption/overview.ts
```diff
@@ -152,21 +148,50 @@ export async function fetchConsumptionOverview(
     filter?: ConsumptionScopeFilter;
     includeWorkspaceContext?: boolean;
   }
+): Promise<Result<ConsumptionOverview, ElasticsearchError>> {
```
flvndvd: Not ideal to create a shallow abstraction here? Is this needed to compute the period with custom filters?

front/lib/api/actions/servers/workspace_analytics/metadata.ts
```diff
@@ -212,23 +195,20 @@ export const WORKSPACE_ANALYTICS_TOOLS_METADATA = [
...
+      "messages, active and total members, and the top agent. Use this to " +
+      "answer 'how are we doing this month' or 'how many AWU credits did we " +
+      `consume' in one call, and ${GET_TOP_ENTITIES_BY_CREDITS_TOOL_NAME} to ` +
+      "attribute that total. For the credit cap and how much of it is left, " +
+      "point the admin at the workspace Usage page.",
```
flvndvd: Does it know actually how to point the admin to this page?

front/lib/api/actions/servers/workspace_analytics/metadata.ts
```diff
@@ -241,8 +221,9 @@ export const WORKSPACE_ANALYTICS_TOOLS_METADATA = [
       "breakdownBy to split each " +
       "bucket into the top agents, users or models plus an 'other' series (a " +
       "stacked trend). Use this for credit/spend TRENDS over time. Use " +
-      `${GET_CREDIT_USAGE_TOOL_NAME} for a single window's totals and top ` +
-      "agent/user/model attribution. IMPORTANT: " +
+      `${GET_CONSUMPTION_OVERVIEW_TOOL_NAME} for a single window's total ` +
+      `and ${GET_TOP_ENTITIES_BY_CREDITS_TOOL_NAME} to attribute it. ` +
+      "IMPORTANT: " +
```
flvndvd: Let's avoid the "IMPORTANT" as it always early symptoms of "CRITICAL" 😝

front/lib/api/actions/servers/workspace_analytics/metadata.ts
achilleburah: That's the current behavior. Currently Dust answers:

> "You can see it in Dust → Admin → Usage, where it’s shown alongside the remaining credits."

Could be improved

front/lib/api/actions/servers/workspace_analytics/tools/index.ts
```diff
@@ -295,84 +295,55 @@ const handlers: ToolHandlers<typeof WORKSPACE_ANALYTICS_TOOLS_METADATA> = {
...
-        `${row.totalCredits} credits`
-    );
+    const overview = result.value;
+    const label =
+      periodInput.kind === "cycle"
+        ? "the current billing cycle"
+        : `the last ${periodInput.days} days`;
+    const topAgent = overview.topAgent
+      ? `${overview.topAgent.name} [${overview.topAgent.agentId}] ` +
+        `(${overview.topAgent.credits.toFixed(2)} credits)`
```
flvndvd, line 327: What about "1 credits" should be singular.

front/lib/api/actions/servers/workspace_analytics/metadata.ts
```diff
@@ -200,39 +185,35 @@ export const WORKSPACE_ANALYTICS_TOOLS_METADATA = [
...
       "breakdownBy to split each " +
       "bucket into the top agents, users or models plus an 'other' series (a " +
-      "stacked trend). Use this for credit/spend TRENDS over time; use " +
-      "get_credit_usage for a single window's totals and top " +
-      "agent/user/model attribution. IMPORTANT: " +
-      "these figures are ESTIMATES — always tell the user they are approximate " +
+      "stacked trend). Use this for credit/spend TRENDS over time. Use " +
+      `${GET_CONSUMPTION_OVERVIEW_TOOL_NAME} for a single window's total ` +
+      `and ${GET_TOP_ENTITIES_BY_CREDITS_TOOL_NAME} to attribute it. ` +
+      "These figures are ESTIMATES. Always tell the user they are " +
```
flvndvd, line 215: Are they `ESTIMATES` though?

front/lib/api/actions/servers/workspace_analytics/metadata.ts
```diff
@@ -200,39 +185,35 @@ export const WORKSPACE_ANALYTICS_TOOLS_METADATA = [
...
-      "Return estimated AWU credit consumption as a time series over a window " +
+      "Return estimated credit consumption as a time series over a window " +
       "(defaults to the last 30 days), bucketed by day, week, or month. Set " +
       "breakdownBy to split each " +
       "bucket into the top agents, users or models plus an 'other' series (a " +
-      "stacked trend). Use this for credit/spend TRENDS over time; use " +
-      "get_credit_usage for a single window's totals and top " +
-      "agent/user/model attribution. IMPORTANT: " +
-      "these figures are ESTIMATES — always tell the user they are approximate " +
+      "stacked trend). Use this for credit/spend TRENDS over time. Use " +
```
flvndvd, line 212: WHY Some UPPERCASE here?

flvndvd (approved):

front/lib/api/actions/servers/workspace_analytics/metadata.ts
achilleburah: This endpoint is migrated in subsequent PR, I just introduced the tool names here. They will stop being estimate afterwards

front/lib/api/actions/servers/workspace_analytics/metadata.ts
achilleburah: Same, this is still the old metadata so I left it as is in this PR.

front/lib/api/actions/servers/workspace_analytics/tools/index.ts
achilleburah: Well it's actually a float so will never be "1", but it's true that 1.00 should say credits and not credit

PR #31303, opened 2026-08-27
feat(workspace-analytics): move get_credit_timeseries onto the consumption index

achilleburah:
Fourth of five — on #31300, itself on #31290 and #31273. Review those first; this diff is against #31300's branch.

## Description

`get_credit_timeseries` is the last credit-trend reader of `front.agent_message_analytics`; re-pointed at the consumption index, same shape.

`fetchConsumptionTimeseries` covers the trend once it takes a `timezone` — it hardcoded `"UTC"`, which cut the first and last bucket in the wrong place for any caller resolving its window in another zone — plus the `extraFilters` passthrough the rest of the stack uses for agent tags.

`get_usage_timeseries` stays on the legacy index for now; it moves in the next PR of the stack.

## Tests

- `timeseries.test.ts`: timezone-aware bucketing.
- Review fix: date labels used a bare UTC slice instead of `formatDateFromMillis`, off by a day for timezones ahead of UTC.

## Risk

Medium. The tool changes index, and message counts become billed messages only — the consumption indexer skips failed and unbilled messages. Callers are `@analyst` only.

Note the legacy `lib/api/assistant/observability/*` fetchers stay: the old analytics UI, Poke pages and CSV exports still read them, and need their own audit before the index is dropped.

## Deploy Plan

Merge after #31300. Fourth of five: credits ranking → count rankings → overview → credit timeseries → usage timeseries.

### DIFF
--- a/front/lib/api/actions/servers/workspace_analytics/tools/index.ts
+++ b/front/lib/api/actions/servers/workspace_analytics/tools/index.ts
@@ -45,6 +45,7 @@ import { fetchMessageMetrics } from "@app/lib/api/assistant/observability/messag
 import { fetchSkillUsageMetrics } from "@app/lib/api/assistant/observability/skill_usage";
 import { fetchToolUsageMetrics } from "@app/lib/api/assistant/observability/tool_usage";
 import { buildAgentAnalyticsBaseQuery } from "@app/lib/api/assistant/observability/utils";
+import { formatDateFromMillis } from "@app/lib/api/elasticsearch";
 import type { Authenticator } from "@app/lib/auth";
 import type { Result } from "@app/types/shared/result";
 import { Err, Ok } from "@app/types/shared/result";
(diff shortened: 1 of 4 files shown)


front/lib/api/actions/servers/workspace_analytics/tools/index.ts
```diff
@@ -237,6 +238,21 @@ async function renderRanking(
   return new Ok([{ type: "text" as const, text }]);
 }
 
+function dayOf(timestampMs: number): string {
```
avervaet: we could reuse formatDateFromMillis(point.timestamp, timezone)

avervaet: Also this will fix displaying error for users in timezones ahead of UTC

avervaet (approved):

front/lib/api/actions/servers/workspace_analytics/tools/index.ts
```diff
@@ -349,137 +363,60 @@ const handlers: ToolHandlers<typeof WORKSPACE_ANALYTICS_TOOLS_METADATA> = {
...
     const deniedError = workspaceManagerGuard(auth);
     if (deniedError) {
       return new Err(deniedError);
     }
 
-    const window = resolveTimeWindow(
-      { period, startDate, endDate, timezone },
-      "last_30_days"
-    );
+    const window = resolveTimeWindow(input, "last_30_days");
```
flvndvd, line 374: No magic variable 🙏 Can we extract it in a const?

flvndvd (approved):

front/lib/api/actions/servers/workspace_analytics/tools/index.ts
achilleburah: merging this to remove index use. Following up in next pr to align with the other ocurences.

PR #31305, opened 2026-08-27
[front] fix: track extension use without origin

aubin-tchoi:
## Description

Follow-up on #31144. The platform detection we put in place there did not work, Chrome extension config requests can omit the `Origin` header.

This PR replaces this to use a single key for both Chrome and Firefox that is always upserted.

## Tests

- Updated tests.

## Risk

- Low.

## Deploy Plan

- Deploy front.


aubin-tchoi:
PMRR

PR #31308, opened 2026-08-27
[front] enh: show more rows in manage table skeletons

aubin-tchoi:
## Description

Manage Agents and Manage Skills currently show only five rows while their tables load, leaving the loading state visually sparse.

Increase both table skeletons to ten rows and cycle the existing width variants so every added row keeps the same destination-shaped cell geometry.

## Tests

- Tested locally.

## Risk

- Low.

## Deploy Plan

- Deploy front.


aubin-tchoi:
PMRR

PR #31309, opened 2026-08-27
feat(front): run wake-ups with the last model requested in the conversation

Nils-Fedrigo:
## Description

When a wake-up fires, it posts a `wakeup`-origin user message with no model selection, so the run always falls back to the agent's configured model — even when the user had explicitly picked a different one in the conversation.

The wake-up run now carries over the model requested on the conversation's last non wake-up user turn:

- `getClientSideMCPServerIdsFromLatestNonWakeUpUserMessage` becomes `getContextFromLatestNonWakeUpUserMessage` and also returns `requestedProviderId` / `requestedModelId` / `requestedReasoningEffort`. It's the same single query — that helper already isolated the latest non wake-up user message — so there's no extra round-trip.
- `runWakeUpActivity` rebuilds the selection from those columns and passes it as `modelSelection` to `postUserMessage`, the same field the input-bar model picker uses.
- `resolvedModelFromUserMessageRow` now takes the three columns structurally instead of a full `UserMessageModel`, matching its `resolvedModelFromAgentMessageRow` sibling, so the raw-SQL result can feed it.

Anchoring on the *last non wake-up* user message (rather than the last message of any kind) keeps a chain of wake-ups pinned to the user's actual choice, and mirrors what the input bar already does with `lastRequestedModel` in `AgentInputBar.tsx`.

No schema change: `requestedProviderId` / `requestedModelId` / `requestedReasoningEffort` already exist on `user_messages`.

## Tests

Added coverage for the new resource method in `conversation_resource.test.ts`: the latest turn wins, wake-up turns are skipped, and nulls come back when no model was requested. `ConversationFactory.createUserMessage` gained optional `requestedModel` / `clientSideMCPServerIds` options to support it.

Ran `conversation_resource.test.ts`, `temporal/triggers/activities.test.ts` and `lib/api/assistant/models.test.ts` — 197 passed.

Note: the existing wake-up activity tests only cover the skip paths (terminal / missing wake-ups), so the carry-over is tested at the resource layer rather than end-to-end through `runWakeUpActivity` — driving that would need a full agent plus posted-message fixture.

## Risk

Low, and the fallbacks are the pre-existing ones. If the user never picked a model the columns are null and behaviour is unchanged. If the picked model has since been disabled for the workspace, `resolveModel` already falls back to the agent's model, and `enforcePremiumModelLimit` still applies — so a stale selection can't break a wake-up run.


Fraggle (approved):
Nice !

PR #31310, opened 2026-08-27
fix(front): allow pod tab deletion when other tabs have stale paths

Fraggle:
## Description

`validatePodFrameTabs` now permits updates to retain frame tab paths already stored on a Pod, even when the corresponding Frame file was moved or deleted. New tab paths still require an existing interactive Frame, so users can remove, reorder, or rename tabs without stale entries blocking metadata updates.

## Tests

- Added unit tests for existing-path leniency, new-tab validation, and file-system initialization errors.
- Added an API integration test covering removal of a valid tab while another persisted tab has a stale path.

## Risk

Low. The change only skips revalidation for paths already persisted in Pod metadata; new tabs retain the existing checks. Rollback is safe through the standard deploy rollback path.

## Deploy Plan

Deploy `front` and `front-api`. No SQL migration, infrastructure change, or script run required.

### DIFF
--- a/front/lib/api/projects/frame_tabs.test.ts
+++ b/front/lib/api/projects/frame_tabs.test.ts
@@ -1,0 +1,135 @@
+import { DustFileSystem } from "@app/lib/api/file_system";
+import { validatePodFrameTabs } from "@app/lib/api/projects/frame_tabs";
+import { createResourceTest } from "@app/tests/utils/generic_resource_tests";
+import { SpaceFactory } from "@app/tests/utils/SpaceFactory";
+import { DustFileSystemError } from "@app/types/file_system";
+import { frameContentType } from "@app/types/files";
+import { DEFAULT_POD_FRAME_TAB_ICON } from "@app/types/pod_frame_tab";
+import { Err, Ok } from "@app/types/shared/result";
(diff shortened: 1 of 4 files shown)


front/lib/api/projects/frame_tabs.test.ts
```diff
@@ -0,0 +1,135 @@
...
+import { DEFAULT_POD_FRAME_TAB_ICON } from "@app/types/pod_frame_tab";
+import { Err, Ok } from "@app/types/shared/result";
+import { beforeEach, describe, expect, it, vi } from "vitest";
+
+describe("validatePodFrameTabs", () => {
+  beforeEach(() => {
+    vi.restoreAllMocks();
+  });
+
+  it("skips file existence checks for existing frame tab paths", async () => {
```
Nils-Fedrigo, line 16: should it remove it as well?

Nils-Fedrigo (approved):
thanks

front/lib/api/projects/frame_tabs.test.ts
Fraggle: WDYM ?
remove the tab if the file is no longer present ?

PR #31312, opened 2026-08-27
chore(front): remove unused hasFeature after models_picker flag removal

smb2268:
## Description

`31210` removed the `models_picker` feature-flag usage in `InputBarContainer` but left `const { hasFeature } = useFeatureFlags();` (and the now-unused import) behind. CI runs `biome ci --error-on-warnings .` over the whole repo, and `noUnusedVariables` trips on it — so format-check is currently red on `main` and blocks every open PR. This removes the dead declaration and orphaned import.

## Risk

None. Pure dead-code removal; no behavior change.


smb2268:
Closing — the unused `hasFeature` was already removed on `main` independently, so this is redundant.

PR #31314, opened 2026-08-27
Build Frame functions before publication

fontanierh:
## Description

Add the Frames v2 publication coordinator that prepares the invoking conversation DSBX, derives its filesystem from the Pod scope resolved under the lifecycle lock, resolves declared function entrypoints, builds every function with the existing sandbox builder, and invokes the atomic publisher only after all builds succeed.

Frame ownership is not tied to a Pod. The resolved conversation filesystem supports source in the conversation itself or in its associated Pod. MCP exposure and runtime execution remain out of scope.

## Tests

- `npm test -- lib/api/frames/build_and_publish.test.ts lib/api/sandbox/lifecycle.test.ts lib/api/sandbox_functions/build_on_sandbox.test.ts lib/api/frames/publication_storage.test.ts`
- `npx biome check front/lib/api/frames/build_and_publish.ts front/lib/api/frames/build_and_publish.test.ts front/lib/api/sandbox/lifecycle.ts front/lib/api/sandbox_functions/build_on_sandbox.ts`
- `NODE_OPTIONS="--max-old-space-size=8192" npx tsc --noEmit` fails on the existing tsgo-only `checkers` option

## Risk

Low. This adds an uncalled backend coordinator, exposes the lifecycle-resolved conversation scope, and extracts the already-tested sandbox build body so a prepared conversation sandbox can reuse it. Path, sandbox, or build failures return before any publication files are written.

## Deploy Plan

Deploy normally. No migration or feature flag change is required for this unexposed slice.

### DIFF
--- a/front/lib/api/frames/build_and_publish.ts
+++ b/front/lib/api/frames/build_and_publish.ts
@@ -1,0 +1,117 @@
+import path from "node:path";
+import { DustFileSystem } from "@app/lib/api/file_system/dust_file_system";
+import type {
+  FramePublicationError,
+  FramePublicationFunctionArtifact,
+  FramePublicationSourceFile,
+} from "@app/lib/api/frames/publication_storage";
+import { publishFramePublication } from "@app/lib/api/frames/publication_storage";
+import { ensureConversationSandboxReadyWithScope } from "@app/lib/api/sandbox/lifecycle";
+import { buildSandboxFunctionOnReadySandbox } from "@app/lib/api/sandbox_functions/build_on_sandbox";
+import { SandboxFunctionError } from "@app/lib/api/sandbox_functions/errors";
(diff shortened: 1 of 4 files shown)


front/lib/api/frames/build_and_publish.ts
```diff
@@ -0,0 +1,117 @@
...
+    sourceFiles: FramePublicationSourceFile[];
+  }
+): Promise<
+  Result<
+    { publicationId: string },
+    FramePublicationError | SandboxFunctionError
+  >
+> {
+  if (manifest.functions.length === 0) {
+    return publishFramePublication(auth, {
```
davidebbo, line 47: nit: I would have conditionally built `functionArtifacts` and use a single publish call

davidebbo (commented):
LGTM

PR #31315, opened 2026-08-27
Add feature-flagged Frames v2 publishing

fontanierh:
## Description

Route the existing `interactive_content` MCP server to a Frames v2 implementation for workspaces with `frames_v2` enabled.

The v2 publish tool validates the canonical `manifest.json` FileResource path, snapshots the bounded source folder, builds declared functions, and atomically activates the publication. Legacy Frames continue through the existing publish path.

## Tests

- `npx tsgo --noEmit`
- Focused Vitest suite for feature flag routing, legacy compatibility, source snapshot publication, and the existing build/publish paths

## Risk

Low and gated by the Dust-only `frames_v2` feature flag. Disable the flag to restore the existing server. Source publication is bounded to 1,000 files and 100 MB.

## Deploy Plan

Deploy `front`, then enable `frames_v2` only for selected internal workspaces. No migration is required.

### DIFF
--- a/front/lib/api/actions/servers/interactive_content_v2/metadata.ts
+++ b/front/lib/api/actions/servers/interactive_content_v2/metadata.ts
@@ -1,0 +1,40 @@
+import {
+  INTERACTIVE_CONTENT_SERVER_NAME,
+  INTERACTIVE_CONTENT_TOOLS_METADATA,
+  PUBLISH_INTERACTIVE_CONTENT_FILE_TOOL_NAME,
+} from "@app/lib/api/actions/servers/interactive_content/metadata";
+import assert from "assert";
+
+const publishToolMetadata = INTERACTIVE_CONTENT_TOOLS_METADATA.find(
+  (
+    tool
+  ): tool is Extract<
+    (typeof INTERACTIVE_CONTENT_TOOLS_METADATA)[number],
+    { name: typeof PUBLISH_INTERACTIVE_CONTENT_FILE_TOOL_NAME }
+  > => tool.name === PUBLISH_INTERACTIVE_CONTENT_FILE_TOOL_NAME
+);
+assert(publishToolMetadata, "Legacy Frame publish metadata expected");
+
+export const INTERACTIVE_CONTENT_V2_SERVER_NAME =
+  INTERACTIVE_CONTENT_SERVER_NAME;
+
+export const INTERACTIVE_CONTENT_V2_PUBLISH_TOOL_METADATA = [
+  {
+    ...publishToolMetadata,
+    description:
+      "Publish a Frame from its current source. For a Frames v2 manifest, validate and snapshot " +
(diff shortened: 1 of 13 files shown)


front/lib/api/actions/servers/interactive_content_v2/metadata.ts
```diff
@@ -0,0 +1,40 @@
+import {
+  INTERACTIVE_CONTENT_SERVER_NAME,
+  INTERACTIVE_CONTENT_TOOLS_METADATA,
+  PUBLISH_INTERACTIVE_CONTENT_FILE_TOOL_NAME,
+} from "@app/lib/api/actions/servers/interactive_content/metadata";
+import assert from "assert";
+
+const publishToolMetadata = INTERACTIVE_CONTENT_TOOLS_METADATA.find(
```
davidebbo, line 8: A little hard to understand :)

front/lib/api/actions/servers/interactive_content_v2/metadata.ts
```diff
@@ -0,0 +1,40 @@
...
+assert(publishToolMetadata, "Legacy Frame publish metadata expected");
+
+export const INTERACTIVE_CONTENT_V2_SERVER_NAME =
+  INTERACTIVE_CONTENT_SERVER_NAME;
+
+export const INTERACTIVE_CONTENT_V2_PUBLISH_TOOL_METADATA = [
+  {
+    ...publishToolMetadata,
+    description:
+      "Publish a Frame from its current source. For a Frames v2 manifest, validate and snapshot " +
```
davidebbo, line 25: "For a Frames v2 manifest": could it be anything but frame v2 if we're using this toolset?

front/lib/api/frames/publish_from_source.ts
```diff
@@ -0,0 +1,211 @@
...
+  const fsResult = await DustFileSystem.fromScopedPath(
+    auth,
+    canonicalManifestPath
+  );
+  if (fsResult.isErr()) {
+    return frameError("invalid_source", fsResult.error.message);
+  }
+  const dustFs = fsResult.value;
+
+  const writeAccess = dustFs.checkWriteAccess(canonicalManifestPath);
```
davidebbo, line 89: Why does that need write access? Isn't that the source manifest, that we should only be reading from during publish?

front/lib/api/frames/publish_from_source.ts
```diff
@@ -0,0 +1,211 @@
...
+    if (content.value === null) {
+      return frameError(
+        "invalid_source",
+        `Frame source file not found: ${relativePath}`
+      );
+    }
+    sourceFiles.push({ content: content.value, contentType, relativePath });
+  }
+
+  return buildAndPublishFramePublication(auth, {
```
davidebbo, line 205: So iiuc, all source files are loaded in memory and passed to the publish function? I guess that's fine since apps should be smallish.

PR #31317, opened 2026-08-28
Improve dev container

Fraggle:
## Description

I moved the shared in-container development environment from `.cursor/` to `dev/` so Cursor Cloud Agents, local Docker workflows, and other AI agents use the same `dev/Dockerfile` and lifecycle scripts. `.cursor/environment.json` is now a thin adapter; `dev/scripts/up.sh`, `infra.sh`, `apps.sh`, and `docker-run.sh` cover installation, local infrastructure, migrations, optional WorkOS seeding, and mprocs.

`dev/scripts/common.sh` materializes the full 1Password Environment into `/tmp/dust-op-environment.env`, loads it through `/tmp/dust-shell-env.sh` via `BASH_ENV` and interactive shell startup, then applies local DB/API/Temporal overrides. `refresh-op-env.sh` refreshes the snapshot, while the duplicated `.cursor` scripts and `with-op-run.sh` are removed.

## Tests

- Ran `bash dev/scripts/refresh-op-env.sh`; materialized 204 variables.
- Verified non-interactive and interactive Bash shells load the materialized environment and local overrides.
- Verified scripts sourcing `dev/scripts/common.sh` inherit the materialized environment.
- CI checks for changed code passed; application test jobs were skipped because this PR only changes development tooling.

## Risk

Low production risk: the PR only changes the development container and startup scripts. Secrets persist in `/tmp` with mode `600`, and `BASH_ENV` affects every non-interactive Bash process, so the loader remains file-only and recursion-safe. The snapshot can become stale until refreshed, and existing containers need a rebuild or restart to pick up the new layout. Reverting the image and script changes restores the previous workflow.

## Deploy Plan

Rebuild the shared `dev/Dockerfile` image and refresh or restart existing Cursor/local containers. No production deploy or SQL migration is required.


zmarouf:
I kind of like the `op run` model more but lgtm nonetheless

zmarouf (approved):

PR #31319, opened 2026-08-28
Add CellType and CellInfo

Fraggle:
## Description

I add `CellType` and `CellInfo` to model standalone Dust deployments and introduce `config` helpers for resolving the current cell, cell URLs, peer cells, and cross-cell sync settings. `RegionInfo` retains its URL for backward compatibility during the migration to cell-owned URLs.

## Tests

Not run; this PR adds type definitions and configuration helpers only.

## Risk

Low. The new cell configuration is isolated, but production behavior depends on the `CELL`, `DUST_*_URL`, and `REGION_RESOLVER_SECRET` environment variables being configured correctly.

## Deploy Plan

Deploy `front`. No database migrations or infrastructure changes.

### DIFF
--- a/front/types/cell.ts
+++ b/front/types/cell.ts
@@ -1,0 +1,15 @@
+import type { RegionType } from "@app/types/region";
+
+// Cells are standalone deployments of a whole Dust environment.
+export const SUPPORTED_CELLS = ["cell-00000", "cell-00001"] as const;
+export type CellType = (typeof SUPPORTED_CELLS)[number];
+
+export interface CellInfo {
+  name: CellType;
+  region: RegionType;
+  url: string;
+}
+
+export function isCellType(cell: string): cell is CellType {
+  return SUPPORTED_CELLS.includes(cell as CellType);
+}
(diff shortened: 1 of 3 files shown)


Fraggle:

- #31319 ◀

front/types/cell.ts
```diff
@@ -0,0 +1,15 @@
+import type { RegionType } from "@app/types/region";
+
+// Cells are standalone deployments of a whole Dust environment.
+export const SUPPORTED_CELLS = ["cell-00000", "cell-00001"] as const;
```
flvndvd, line 4: Do we really need this whole list? Isn't the current cell being aware of its cell identity is enough?

Fraggle: Good question (that beeing said, I'm not 100% sure what you mean by "aware of it's cell identity")

I replicated the region code which separate the types declarations (RegionType and RegionInfo) and the config.

So to me this list is the single source of truth of the cells that we have, and we ensure that we have a config for all of them by leveraging assertNever() in the `config.ts` file.

I feel like it's a good pattern, otherwise we would derive the types from the config which feels backward.

flvndvd: But are we sure we need a type?

flvndvd (approved):
LGTM

front/types/cell.ts
flvndvd: Seen IRL

PR #31320, opened 2026-08-28
Register Frames v2 from dsbx

fontanierh:
## Description

- Add `dsbx frame register <manifest>` to assign a stable Frames v2 `FileResource` identity to an existing mounted manifest.
- Make registration idempotent by GCS mount path and validate the manifest before creating the record.
- Add `dsbx frame create` to scaffold `manifest.json` and `index.tsx`, then register the Frame.
- Add an internal, feature-flagged sandbox registration endpoint. Publishing is unchanged.
- Pin `dust-base` 0.8.99 to `dsbx` 0.1.52 so the registration commands ship in the consuming sandbox image.

Initial registration supports writable GCS-backed conversation and Pod filesystems.

## Tests

- Frames targeted Vitest suites: 17 tests.
- Sandbox image registry Vitest suite: 21 tests.
- Front and front-api tsgo typechecks.
- `dsbx` focused Cargo tests and Clippy with warnings denied.
- Biome formatting and Swagger annotation lint.

## Risk

Feature-flagged and additive. Registration validates source access and the manifest before creating a record, adopts the existing mounted object without a canonical upload copy, rejects database-backed filesystems, and handles concurrent registration through the existing unique mount-path constraint.

The `dust-base` build requires the `dsbx-v0.1.52` release to exist first.

## Deploy Plan

Merge, release `dsbx` 0.1.52 from `main`, build and register `dust-base` 0.8.99, deploy front, request a kill of older `dust-base` versions from `/poke/kill`, then enable `frames_v2`. No migration or backfill.

### DIFF
--- a/front/lib/api/frames/register_from_source.ts
+++ b/front/lib/api/frames/register_from_source.ts
@@ -1,0 +1,163 @@
+import path from "node:path";
+
+import { DustFileSystem } from "@app/lib/api/file_system";
+import { FramePublicationError } from "@app/lib/api/frames/publication_storage";
+import type { Authenticator } from "@app/lib/auth";
+import { FileResource } from "@app/lib/resources/file_resource";
+import {
+  FRAME_MANIFEST_FILE,
+  parseFrameManifest,
+} from "@app/types/api/frame_manifest";
+import type { ConversationWithoutContentType } from "@app/types/assistant/conversation";
+import type { DustFileSystemError } from "@app/types/file_system";
+import type { FileUseCase, FileUseCaseMetadata } from "@app/types/files";
+import { frameV2ContentType } from "@app/types/files";
+import type { Result } from "@app/types/shared/result";
+import { Err, Ok } from "@app/types/shared/result";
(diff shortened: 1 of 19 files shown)


front/lib/api/frames/register_from_source.ts
```diff
@@ -0,0 +1,163 @@
...
+      userId: auth.user()?.id ?? null,
+      contentType: frameV2ContentType,
+      fileName: FRAME_MANIFEST_FILE,
+      fileSize: manifestBuffer.length,
+      useCase,
+      useCaseMetadata,
+      mountFilePath,
+      fileSystemNodeId: null,
+    });
+    await frame.markFrameV2AsReadyFromMount();
```
smb2268, line 149: I think because we're marking this ready here when it doesn't yet have an `original` object,`FileResource.delete()` will fail if called before the original is created - not sure if that's a real or hypothetical concern here

smb2268 (commented):
LGTM, just one note

PR #31321, opened 2026-08-28
Add appearance toggle to the cmd-k palette + cmd-k animation

pinotalexandre:
## Summary
- Add a "Settings" section to the cmd-k palette letting users switch between light, dark, and system appearance directly from the palette.
- Only the two themes not currently active are offered, with "system" always shown second.
- The section surfaces above other results when the search query matches it.
- Trim the palette's own open/close animation: drop the scale/zoom for a plain fade + slide, with smoother timing.

https://github.com/user-attachments/assets/f7459135-9128-43a8-b401-98b88c2a5a49

### DIFF
--- a/front/components/command_palette/CommandPalette.tsx
+++ b/front/components/command_palette/CommandPalette.tsx
@@ -14,6 +14,7 @@ import {
   CommandPaletteSearchPhase,
 } from "@app/components/command_palette/CommandPaletteSearchPhase";
 import { SkillDetailsSheetById } from "@app/components/command_palette/SkillDetailsSheetById";
+import type { Theme } from "@app/components/sparkle/ThemeContext";
 import { useTheme } from "@app/components/sparkle/ThemeContext";
 import { useAppRouter } from "@app/lib/platform";
 import { useAgentConfigurations } from "@app/lib/swr/assistants";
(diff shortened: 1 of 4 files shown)


front/components/command_palette/CommandPalette.tsx
```diff
@@ -20,19 +27,43 @@ import {
...
+  Sun,
+} from "@dust-tt/sparkle";
 import { useCallback, useEffect, useMemo, useRef, useState } from "react";
 
 interface CommandPaletteProps {
   owner: LightWorkspaceType;
   user: UserType;
 }
 
+type Theme = "dark" | "light" | "system";
```
flvndvd: Pretty sure this pal already exist in the `ThemeContext`, can we import it from there?

front/components/command_palette/CommandPalette.tsx
```diff
@@ -193,20 +247,34 @@ export function CommandPalette({ owner, user }: CommandPaletteProps) {
...
-        return;
-      }
-      // Skills without administration access have only one action (view details).
-      if (item.kind === "skill" && !item.skill.canAdministrate) {
-        executeAction(item, "view_details");
-      } else {
-        setSelectedItem(item);
-        setPhase("action");
+      switch (item.kind) {
+        case "command":
```
flvndvd, line 258: It feels slightly "too" local to write command on set theme, right?

PR #31322, opened 2026-08-28
[front] canRead-gated group fetchers; API key scoping to restricted spaces/pods

tdraier:
## Description

Rework the `GroupResource` fetcher layer around a `canRead` filter, split ACL-bypass paths behind explicit `dangerously*` accessors, and widen API key scoping to include restricted pods (projects) — per [dust-tt/tasks#10202](https://github.com/dust-tt/tasks/issues/10202#issuecomment-5441979766).

### `GroupResource` fetchers

Every named fetcher enforces `canRead` on the returned rows. Callers that must see rows regardless of the caller's ACL go through the explicit `dangerously*` opt-outs:

- `fetchById` / `fetchByIds` — filter on `canRead`; return `Err("unauthorized")` if any id resolves to a group the caller can't read.
- `fetchByName` — returns `null` when the row isn't readable.
- `fetchByWorkOSGroupId` — same shape; used by SCIM/WorkOS sync.
- `listAllWorkspaceGroups` — filter on `canRead` and default the kinds to a new `USER_VISIBLE_GROUP_KINDS = [...MANAGEABLE_GROUP_KINDS, "global"]`; callers that want internal kinds must ask explicitly.
- **`fetchByModelIds` → `dangerouslyFetchByModelIds`** (no `canRead`). Model ids never cross the client boundary (per SEC2) — they are internal identifiers used by permission plumbing to resolve groups referenced from grants/joins where the ACL is enforced on the parent resource. Applying `canRead` here would silently drop legitimate rows.
- **`fetchWorkspaceSystemGroup` → `dangerouslyFetchWorkspaceSystemGroup`** (no `isAdmin` gate). The system group is deny-all in the ACL; the only way to reach it is through this explicit opt-out from scripts / system plumbing.
- New **`groupExistsByName`** — cheap existence check used by `makeNew` / `updateRegularManualGroup` name-conflict paths and by `seedGroup`. No ACL — a name collision must be caught regardless of caller visibility.

### ACL

Only one ACL change here: **`regular_auto` is aligned with `provisioned`'s shape** (admin manages, manager reads, members read via the self-grant). The broader "everyone reads every non-system kind" simplification stays in #31360 to keep the two PRs focused.

### API keys — scoping (was `useGroups`, now dedicated endpoint)

Previously the API keys page used the generic `useGroups` hook, which listed every workspace group — including `regular_auto`, pod groups, etc. — as picker options. This PR replaces that with a dedicated `GET /keys/groups` endpoint served by `listKeyScopableGroups`:

- Returns the groups attached to the workspace's restricted regular spaces and restricted pods, plus the workspace global group so existing keys (always force-scoped to global by the backend) resolve their name in the keys table icon/tooltip. The picker filters global out client-side.
- Open spaces / open pods are excluded (readable via global — not meaningful to scope to).
- `POST /keys` group-scoping validation matches; error message updated to "restricted spaces or pods".

### `SpaceResource.listWorkspaceSpaces`

New `includeOpen` / `includeRestricted` options (default `true`, backward-compatible). Applies to regular/project spaces only; other kinds pass through. Both `false` fails a fail-fast assert. Used by the new `listKeyScopableGroups`.

## Tests

- `front` + `front-api` `npx tsgo --noEmit` clean.
- Group / space / permission resource suites (161 tests) pass.
- Keys route suites: added
  - "includes restricted pod groups" and "excludes open pod groups" for `GET /keys/groups`;
  - "rejects scoping to a group not tied to a restricted space or pod" for `POST /keys`.
- The manual test plan is in the PR review thread.

## Risk

- **`fetchByModelIds` semantic change (implicit ACL → explicit opt-out).** All existing callers were migrated in the same commit. Behavior is preserved (they were previously unfiltered and remain unfiltered via the renamed method).
- **`fetchWorkspaceSystemGroup` semantic change (rename + drop `isAdmin` gate).** Only two migrations call it, both under `internalAdminForWorkspace`. The gate was redundant given the caller auth; the rename makes the intent explicit.
- **API keys — scoping picker.** The picker now offers a curated set (restricted spaces + restricted pods + implicit global) instead of every workspace group. Restricted pod scoping — the common real case — is preserved. Any key already scoped to a group that no longer appears in the picker still works (the backend only rejects new scoping attempts to non-scopable groups).
- **`GET /w/:wId/groups` default kinds change** to `USER_VISIBLE_GROUP_KINDS`. No frontend caller relies on the default — every `useGroups({ kinds })` call passes explicit kinds — so the frontend is unaffected. The one runtime consumer of the default (a poke plugin picker) narrows from "everything visible" to "user-visible only", which is the intent.

Rollback: revert the PR; no data migrations, no schema changes.

## Deploy Plan

Standard `front` deploy. No migrations, no feature flags, no coordination required. #31360 stacks on top and can ship immediately after.

### DIFF
--- a/front/components/pages/workspace/developers/APIKeysPage.tsx
+++ b/front/components/pages/workspace/developers/APIKeysPage.tsx
@@ -15,7 +15,7 @@ import { formatCredits } from "@app/lib/client/credits";
 import { useSubmitFunction } from "@app/lib/client/utils";
 import { clientFetch } from "@app/lib/egress/client";
 import { useKeys } from "@app/lib/swr/apps";
-import { useGroups } from "@app/lib/swr/groups";
+import { useKeyScopableGroups } from "@app/lib/swr/groups";
 import type { ConsumptionScopeFilter } from "@app/types/api/analytics/consumption";
 import type { GroupType } from "@app/types/groups";
 import type { KeyType } from "@app/types/key";
@@ -124,7 +124,9 @@ export function APIKeysPageContent({ owner, period }: APIKeysPageContentProps) {
   const [editCapKey, setEditCapKey] = useState<KeyType | null>(null);
 
   const { isKeysError, isKeysLoading, keys } = useKeys(owner);
-  const { groups, isGroupsError, isGroupsLoading } = useGroups({ owner });
+  const { groups, isGroupsError, isGroupsLoading } = useKeyScopableGroups({
+    owner,
+  });
   const isDataLoading = isKeysLoading || isGroupsLoading;
   const isDataError = !!isKeysError || isGroupsError;
 
--- a/front/lib/api/keys/scopable_groups.ts
+++ b/front/lib/api/keys/scopable_groups.ts
@@ -3,27 +3,45 @@ import { GroupResource } from "@app/lib/resources/group_resource";
 import { SpaceResource } from "@app/lib/resources/space_resource";
 
 /**
- * The groups a new API key may be scoped to: exactly the groups associated to
- * the workspace's regular restricted spaces. Open spaces (readable by everyone
- * via the global group) and non-regular spaces (system, global, conversations,
- * projects/pods) are not scopable.
+ * The groups a new API key may be scoped to: the groups associated to the
+ * workspace's restricted spaces — both regular (spaces) and project (pods) —
+ * plus the workspace global group.
+ *
+ * A manageable group (regular_manual / provisioned) can be granted access to
+ * several restricted spaces at once, so scoping a key to it indirectly scopes
+ * the key to every restricted space that carries the group. Open spaces are
+ * readable through the global group and are not meaningful to scope to.
+ *
+ * The global group is included so that existing keys — which the backend always
+ * scopes to the global group in addition to any explicit selection — can be
+ * rendered by ids; the picker filters it out client-side.
  */
 export async function listKeyScopableGroups(
   auth: Authenticator
 ): Promise<GroupResource[]> {
-  // `listWorkspaceSpaces` excludes projects/pods by default. The enriched
-  // serialization carries the canonical `isRestricted` flag and the space's
-  // grant `groupIds`, so we just filter and collect.
-  const spaces = await SpaceResource.listWorkspaceSpaces(auth);
-  const enrichedSpaces = await SpaceResource.batchToJSONEnriched(auth, spaces);
+  const spaces = await SpaceResource.listWorkspaceSpaces(auth, {
+    includeProjectSpaces: true,
+    includeOpen: false,
+  });
+  // `includeOpen: false` leaves the non-scopable unique kinds (system, global)
+  // in the result; narrow to the regular/project spaces we actually care about.
+  const restrictedSpaces = spaces.filter(
+    (space) => space.isRegular() || space.isProject()
+  );
+
+  const referencesBySpaceModelId =
+    await SpaceResource.listGrantReferencesBySpaceModelId(restrictedSpaces);
 
   const groupIds = new Set<string>();
-  for (const space of enrichedSpaces) {
-    if (space.kind === "regular" && space.isRestricted) {
-      for (const groupId of space.groupIds) {
-        groupIds.add(groupId);
-      }
+  for (const references of referencesBySpaceModelId.values()) {
+    for (const reference of references) {
+      groupIds.add(reference.groupSId);
     }
+  }
+
+  const globalGroupRes = await GroupResource.fetchWorkspaceGlobalGroup(auth);
+  if (globalGroupRes.isOk()) {
+    groupIds.add(globalGroupRes.value.sId);
   }
 
   const groupsRes = await GroupResource.fetchByIds(auth, [...groupIds]);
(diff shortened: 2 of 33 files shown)


tdraier:

- #31322 ◀
    - #31360
        - #31361

front/lib/api/keys/scopable_groups.ts
```diff
@@ -0,0 +1,34 @@
...
+): Promise<GroupResource[]> {
+  // `listWorkspaceSpaces` excludes projects/pods by default. The enriched
+  // serialization carries the canonical `isRestricted` flag and the space's
+  // grant `groupIds`, so we just filter and collect.
+  const spaces = await SpaceResource.listWorkspaceSpaces(auth);
+  const enrichedSpaces = await SpaceResource.batchToJSONEnriched(auth, spaces);
+
+  const groupIds = new Set<string>();
+  for (const space of enrichedSpaces) {
+    if (space.kind === "regular" && space.isRestricted) {
```
fabiencelier: does it makes sense to limit it to restricted and not support open regular spaces ?
the space can be turned to restricted any time so it can be good to explicitly give the access to the key no ?

front/types/groups.ts
```diff
@@ -61,6 +61,20 @@ export function isManageableGroupKind(kind: GroupKind): boolean {
...
 
+// Group kinds that any workspace member may see directly (e.g. in the workspace
+// Groups listing or when referencing a group by id). Internal kinds
+// (`regular_auto`, `system`, `agent_editors`) are never surfaced this way: they
+// are implementation details of spaces, permissions, and agent editors.
+export const USER_VISIBLE_GROUP_KINDS = [
+  "global",
+  "provisioned",
+  "regular_manual",
+] as const;
```
fabiencelier, line 71: maybe `[...MANAGEABLE_GROUP_KINDS, "global"]` ?

fabiencelier (approved):

front/lib/api/keys/scopable_groups.ts
tdraier: yes, could be - but as long as the space is open it's not really meaningful to create a key with it (it was not possible until we created groups for open spaces)

philipperolet (approved):
Pre-approving, we discussed changes and design IRL

front/lib/api/keys/scopable_groups.ts
```diff
@@ -0,0 +1,34 @@
+import type { Authenticator } from "@app/lib/auth";
+import { GroupResource } from "@app/lib/resources/group_resource";
+import { SpaceResource } from "@app/lib/resources/space_resource";
+
+/**
+ * The groups a new API key may be scoped to: exactly the groups associated to
+ * the workspace's regular restricted spaces. Open spaces (readable by everyone
+ * via the global group) and non-regular spaces (system, global, conversations,
+ * projects/pods) are not scopable.
```
philipperolet: nit: would reword a bit here because a manual / provisioned group can be used both in a restricted space and e.g. a pod, so the pod is indirectly scopable

front/lib/api/keys/scopable_groups.ts
philipperolet: What about scoping to a pod? it seems there might be legit cases for it no? was it possible before? if no, fine; otherwise let's check if customers did it?

front/components/pages/workspace/developers/APIKeysPage.tsx
```diff
@@ -124,7 +124,9 @@ export function APIKeysPageContent({ owner, period }: APIKeysPageContentProps) {
   const [editCapKey, setEditCapKey] = useState<KeyType | null>(null);
 
   const { isKeysError, isKeysLoading, keys } = useKeys(owner);
-  const { groups, isGroupsError, isGroupsLoading } = useGroups({ owner });
+  const { groups, isGroupsError, isGroupsLoading } = useKeyScopableGroups({
```
philipperolet, line 127: Previously we displayed a little company icon under spaces
Now we don't return the global group, let's make sure the icon stays

Image

front/lib/api/keys/scopable_groups.ts
philipperolet: I think at least we at Dust scoped keys to pods 😅 cf tooltip

Image

front/lib/api/keys/scopable_groups.ts
```diff
@@ -0,0 +1,34 @@
...
+ */
+export async function listKeyScopableGroups(
+  auth: Authenticator
+): Promise<GroupResource[]> {
+  // `listWorkspaceSpaces` excludes projects/pods by default. The enriched
+  // serialization carries the canonical `isRestricted` flag and the space's
+  // grant `groupIds`, so we just filter and collect.
+  const spaces = await SpaceResource.listWorkspaceSpaces(auth);
+  const enrichedSpaces = await SpaceResource.batchToJSONEnriched(auth, spaces);
```
philipperolet: nit: a bit weird to use a serialized type here in lib (VS at api boundary), maybe there's a resource method giving us objects, e.g. use listSpaceEnrichmentBySpaceModelId directly?

philipperolet (commented):
Hey, added a few comments, feel free to adress as you see fit 🙏

PR #31326, opened 2026-08-28
[Admin Governance] Block mutation paths for archived agents

philipperolet:
## Description

Archived agents will keep their editor memberships active in the follow-up PR. Freeze all agent-definition mutation paths first so those retained grants cannot be used to edit an archived agent. Restore remains the only definition-changing action; feedback, memories, suggestions, and archive/restore runtime cleanup are unchanged.

This covers private, public v1, and Poke archive/save paths, editor and tag mutations, linked Slack channels, and batch scope/tag/model behavior.

## Risks

Blast radius: agent configuration, editor, tag, scope, and archive mutations for archived agents.
Risk: low

## Tests

- [x] Archived agent definition and editor updates are rejected
- [x] Archived agents are rejected atomically from scope and tag batches
- [x] Focused tests, front and front-api type checks, formatting, and Swagger lint pass

## Deploy Plan

- Deploy front-api
- Deploy front

### DIFF
--- a/front-api/routes/w/[wId]/assistant/agent_configurations/guards.ts
+++ b/front-api/routes/w/[wId]/assistant/agent_configurations/guards.ts
@@ -1,42 +1,20 @@
 import type { LightAgentConfigurationType } from "@app/types/assistant/agent";
-import { apiError } from "@front-api/middlewares/utils";
-import type { Context } from "hono";
+import type { APIErrorWithContentfulStatusCode } from "@app/types/error";
 
-type ApiErrorResponse = ReturnType<typeof apiError>;
+export const ARCHIVED_AGENT_API_ERROR: APIErrorWithContentfulStatusCode = {
+  status_code: 400,
+  api_error: {
+    type: "invalid_request_error",
+    message: "An archived agent cannot be updated. Restore it first.",
+  },
+};
 
-export function rejectArchivedAgent(
-  ctx: Context,
-  agent: LightAgentConfigurationType
-): ApiErrorResponse | null {
-  if (agent.status !== "archived") {
-    return null;
-  }
-
-  return apiError(ctx, {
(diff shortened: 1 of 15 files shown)


front-api/routes/w/[wId]/assistant/agent_configurations/guards.ts
```diff
@@ -0,0 +1,42 @@
...
+import { apiError } from "@front-api/middlewares/utils";
+import type { Context } from "hono";
+
+type ApiErrorResponse = ReturnType<typeof apiError>;
+
+export function rejectArchivedAgent(
+  ctx: Context,
+  agent: LightAgentConfigurationType
+): ApiErrorResponse | null {
+  if (agent.status !== "archived") {
```
Comment (philipperolet) [inline front-api/routes/w/[wId]/assistant/agent_configurations/guards.ts]: could we just do `return rejectArchivedAgents(ctx, [agent]);` here?

Comment (philipperolet) [reply front-api/routes/w/[wId]/assistant/agent_configurations/guards.ts]: also could we name that `getErrorIfArchivedAgents`?

Comment (philipperolet) [reply front-api/routes/w/[wId]/assistant/agent_configurations/guards.ts]: actually, would it be ok to do isArchivedAgent[s] functions returning a boolean, then at call sites write
```
if (isArchivedAgent(agent)) {
   return apiError(ctx, ARCHIVED_AGENT_API_ERROR)
}
```
I think it might be more consistent with the rest of the code and read more nicely

Comment (philipperolet) [reply front-api/routes/w/[wId]/assistant/agent_configurations/guards.ts]: Done. The helpers are now pure `isArchivedAgent` / `isArchivedAgents` predicates, and handlers explicitly return `apiError(ctx, ARCHIVED_AGENT_API_ERROR)`.

fabiencelier (commented):
straightforward ✅

PR #31327, opened 2026-08-28
[Admin Governance] Preserve archived agent editors

philipperolet:
## Description

Follows #31326.

Archived agents currently suspend their editor-group memberships, which makes their editors disappear and prevents editor-only archived agents from remaining visible to those editors. Keep memberships active across archive and restore; agent status, guarded by the previous PR, now freezes definition changes instead of ACL suspension.

Adds an idempotent backfill restoring stale suspended memberships with one bulk update and one awaited cache invalidation per workspace. It targets all agent_editors groups so deploy and backfill order cannot strand editors if an agent is restored in between. These legacy groups will be removed from auth later in the governance rollout.

## Risks

Blast radius: archived-agent visibility and legacy agent editor memberships.
Risk: low

The temporary cost is that editor groups for archived agents re-enter user auth until the legacy groups are removed. The dry run reports the affected group and membership counts before execution.

## Tests

- [x] Archive and restore keep editor memberships active and editors listable
- [x] Focused tests, front and front-api type checks, formatting, and Swagger lint pass
- [x] Batched backfill passes the front type check

## Deploy Plan

- Deploy front and front-api
- From front/, dry-run: npx tsx migrations/20260828_restore_agent_editor_memberships.ts
- Review affected counts
- From front/, execute: npx tsx migrations/20260828_restore_agent_editor_memberships.ts --execute

### DIFF
--- a/front-api/public/swagger.json
+++ b/front-api/public/swagger.json
@@ -798,7 +798,7 @@
       },
       "delete": {
         "summary": "Archive agent configuration",
-        "description": "Archive the agent configuration identified by {sId} in the workspace identified by {wId}. The agent is soft-archived and triggers/editor-group memberships associated with it are disabled.",
+        "description": "Archive the agent configuration identified by {sId} in the workspace identified by {wId}. The agent is soft-archived and its triggers are disabled.",
         "tags": [
           "Agents"
         ],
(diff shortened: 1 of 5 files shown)


fabiencelier (commented):
LGTM ✅

PR #31330, opened 2026-08-28
[Admin Governance] Backfill stable agent identities

philipperolet:
## Description

Follows #31329.

Adds an idempotent dry-run/execute backfill per workspace. It creates or reuses one identity per agent `sId`, attaches every version, rejects inconsistent links, and reports orphan or missing state.

## Risks

Blast radius: the one-time backfill of workspace agent configuration rows.

Risk level: standard. The migration supports dry runs, workspace-level execution, and safe retries.

## Deploy Plan

- Deploy front.
- Run the backfill in dry-run mode and inspect counts and mismatches.
- Execute it, optionally staged by workspace.
- Verify that no configuration is missing `agentId` before #31331.

## Tests

- Added focused coverage in `front/migrations/20260828_backfill_agent_identities.test.ts`.

PR #31331, opened 2026-08-28
[Admin Governance] Enforce stable agent identity invariants

philipperolet:
## Description

Follows #31330.

Makes `agent_configurations.agentId` mandatory and makes `(agentId, version)` unique per logical agent after the backfill. The migration validates a temporary check before setting `NOT NULL` to minimize locking and creates the unique index concurrently.

Transitional null-row tests are removed because that state becomes impossible. Direct test fixtures now create the required agent identity first.

## Risks

Blast radius: agent configuration writes and post-deploy database constraints.

Risk level: standard. The migration must only run after the backfill has reported zero missing or inconsistent rows.

## Deploy Plan

- Confirm that the backfill reports zero missing or inconsistent rows.
- Deploy front and front-api.
- Run `20260828110852_enforce_agent_identity.sql` as a post-deploy migration.

## Tests

- Covered agent identity lifecycle, the agent identity backfill, the Sonnet migration fixtures, and the front-api app route fixtures.
- Applied the post-deploy constraint sequence locally after backfilling local legacy rows.


philipperolet:

- #31328
    - #31329
        - #31330
            - #31331 ◀

front-api/routes/w/[wId]/spaces/[spaceId]/apps/[aId]/index.test.ts
```diff
@@ -36,8 +39,14 @@ describe("DELETE /api/w/:wId/spaces/:spaceId/apps/:aId", () => {
       globalSpace
     );
 
+    const agentSId = generateRandomModelSId();
```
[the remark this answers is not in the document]
Comment (philipperolet) [reply front-api/routes/w/[wId]/spaces/[spaceId]/apps/[aId]/index.test.ts]: Fixed in 0063d7db75.

front/migrations/post-deploy/20260828110852_enforce_agent_identity.sql
```diff
@@ -0,0 +1,35 @@
+/*
+Statement 0
+*/
+SET SESSION statement_timeout = 3000;
+SET SESSION lock_timeout = 3000;
+ALTER TABLE "public"."agent_configurations" ADD CONSTRAINT "agent_configurations_agent_id_not_null_check" CHECK("agentId" IS NOT NULL) NOT VALID;
```
tdraier, line 12: why all this instead of just `COLUMN "agentId" SET NOT NULL` ?

tdraier (approved):
lgtm

front/migrations/post-deploy/20260828110852_enforce_agent_identity.sql
philipperolet: The staged form avoids scanning `agent_configurations` while holding the `ACCESS EXCLUSIVE` lock required by `SET NOT NULL`. `VALIDATE CONSTRAINT` performs the scan without blocking normal reads/writes; PostgreSQL can then reuse the validated check so the strong lock is brief. I kept the safer form and documented the rationale

PR #31332, opened 2026-08-28
fix(dev): wait for workspace before Cursor Cloud start/terminals

Fraggle:
## Description

Restores dev image buildability by moving the default `OP_ENVIRONMENT_ID` from `dev/Dockerfile` to `dev/scripts/env.sh`, where the secret-scanner pragma does not invalidate Docker `ENV` syntax. `dev/scripts/docker-run.sh` now forwards host overrides into the container, and `dev/README.md` documents the Cursor Cloud snapshot skew that can leave `/workspace` out of sync with the current `.cursor/environment.json`.

## Tests

Manually verified the environment-variable default and host-to-container forwarding. Confirmed that the Dockerfile no longer contains the invalid inline-commented `ENV` assignment. Changed-file, security, and static checks passed; application test suites were skipped because no application source changed.

## Risk

Low. The changes affect development-image parsing and environment propagation only. Rollback is a straightforward revert.

## Deploy Plan

No production service deploy. Rebuild the dev environment image/snapshot so new Cursor Cloud agents consume the corrected configuration.


zmarouf (commented):
LGTM

PR #31333, opened 2026-08-28
feat: reasoning in the manage agents bulk model edit

adrsimon:
## Description

Manage Agents lets you filter by model and bulk-set a model, but neither knew about reasoning effort. `batch_update_model` already accepted `reasoningEffort`, the UI just never sent it.

Both surfaces now render the conversation input bar picker instead of a custom one. That meant generalizing the picker to multi-select. A new `useModelPickerMenu` hook holds what all three call sites need, so the input bar behaves exactly as before.

Filter's list of models is derived from the model ids the workspace's agents use, resolved against the full static list of models. Using the input bar's original list of models would add noise in the filter's list, and would possibly not show models that are used by some agents.

The filter stays model-only: picking a model matches every agent on it, whatever effort they run. Only the bulk "Set model" dropdown shows the effort slider, since that one has to send an effort.

## Tests

`modelPickerUtils` and `modelPickerTracking` pass, which is the guard on the input bar refactor.

Capture d’écran 2026-08-28 à 11 28 21

Capture d’écran 2026-08-28 à 11 28 35

## Risk

Behavior should be identical for the input bar picker. Frontend only, no schema or API change, safe to roll back.

Filter semantics are unchanged: selecting a model still returns every agent on it. The filter also drops the member's tier and plan restrictions, since an agent can sit on a model the person reading the list could not pick themselves.

## Deploy Plan

Deploy front.


front/components/pages/builder/agents/ManageAgentsPage.tsx
```diff
@@ -208,6 +206,14 @@ export function ManageAgentsPage() {
     isSearchActive,
   ]);
 
+  const usedModelIds = useMemo(
+    () =>
+      Array.from(
+        new Set(agentConfigurations.map((a) => a.model.modelId))
```
aubin-tchoi: I think this omits the archived ones, so if we look at the archived tab we may have an issue

front/components/assistant/ModelsFilterMenu.tsx
```diff
@@ -18,71 +25,76 @@ export type AgentModelFilterType = {
...
-      }
-      onSelect={(event) => {
-        event.preventDefault();
-        onToggle(model);
-      }}
-    />
-  );
-}
-
 export function ModelsFilterMenu({
```
aubin-tchoi, line 39: ideally this component is entirely replaced by `ModelPicker`. do we still have a lot to do to get there?

front/components/model_picker/useModelPickerMenu.ts
```diff
@@ -0,0 +1,162 @@
...
+  }, [allModels]);
+
+  // Picking a concrete model (or nudging its effort slider) must keep the menu
+  // and its open submenus visible so the effort can still be adjusted. The
+  // click briefly moves focus/pointer in a way Radix treats as an
+  // interaction-outside and dismisses the (sub)menu; we record the pick time
+  // and veto the close that immediately follows it.
+  const lastModelInteractionAtMsRef = useRef(0);
+  const shouldBlockDismiss = () =>
+    Date.now() - lastModelInteractionAtMsRef.current < 300;
```
aubin-tchoi: that looks quite hacky

front/components/model_picker/useModelPickerMenu.ts
```diff
@@ -0,0 +1,162 @@
...
+
+// The transient menu state and derived model lists shared by every surface
+// rendering `ModelPickerContent`: the input bar, the manage-agents model filter
+// and the bulk "Set model" dropdown.
+// "select" picks a model to run; "filter" only names one. A filter must reach
+// every model an agent can already sit on, so it drops the member's tier and
+// plan restrictions instead of padlocking the rows they exclude.
+type ModelPickerMenuMode = "select" | "filter";
+
+export function useModelPickerMenu({
```
aubin-tchoi: having a hard time understanding what part of the logic this hook handles. is this hook about data access (`lockPremiumEfforts`, `allModels`, `streamModels`, ...), or UI state handling (`onToggleMoreModels`, `onSearchChange`, ...)?

front/components/model_picker/ModelPickerContent.tsx
```diff
@@ -52,17 +53,20 @@ interface ModelPickerContentProps {
   onToggleMaker: (makerId: ModelMakerIdType) => void;
   onSelectTier: (tierId: ModelTierId) => void;
   onSelectModel: (model: ModelConfigurationType) => void;
-  onChangeEffort: (effort: ReasoningEffort) => void;
-  onRevert: () => void;
+  onChangeEffort?: (
+    model: ModelConfigurationType,
+    effort: ReasoningEffort
+  ) => void;
+  footer?: ReactNode;
```
aubin-tchoi: I think that's too permissive/not self-explanatory enough to allow any react node here

aubin-tchoi (commented):
LGTM design-wise but ideally we'd really share the model picker component

front/components/model_picker/useModelPickerMenu.ts
adrsimon: indeed was doing both, splitting into two hooks one for model one for ui state

front/components/assistant/ModelsFilterMenu.tsx
adrsimon: the only way as of now i'd see it work is with a "mode" over the ModelPicker (like "select" | "filter" | "bulk") and i kinda dislike it

i think this is already a better state than before and the change would be much bigger i believe

adrsimon: and we still reuse ModelPickerContent which is better than before

aubin-tchoi: yeah agreed

aubin-tchoi: let's ship as is so that the UI is correct at least, and then rewrite the component entirely, wdyt?

adrsimon: yup works for me 👍

aubin-tchoi (approved):
LGTM

PR #31335, opened 2026-08-28
feat(models): record model degradations in their own table

Nils-Fedrigo:
## Description

Adds `model_degradations`, one row per degradation of one model: the model was marked degraded at `startedAt`, and either still is (`status: "ongoing"`, `endedAt: null`) or stopped being at `endedAt` (`status: "resolved"`).

| column | |
|---|---|
| `modelId` | e.g. `claude-opus-5`, widened to `string` |
| `providerId` | denormalized, so a provider outage can be charted without a catalog join |
| `startedAt` / `endedAt` | the interval; `endedAt` is null while ongoing |
| `status` | `"ongoing"` / `"resolved"` |

Base of the degraded-models stack. The rest of the stack moves onto it: nothing reads the table in this PR.

**Why.** Today a degraded model is a `kill_switches` row typed `global_degraded_model:<modelId>`, and un-degrading it destroys the row. That leaves no history at all — the moment an incident ends, every trace of it is gone, so there is nothing to build a provider-outage timeseries from. It also costs the switch system its closed union: `KillSwitchType` had to widen to include `global_degraded_model:${string}`, so `isKillSwitchType` accepts an unbounded namespace and reading a model id back means slicing a prefix off a string.

A row per degradation fixes both. The set of models degraded right now is the set of `"ongoing"` rows, so the table is its own history: there is no separate switch to keep in sync, and no way for the live state and the record of it to disagree.

The partial unique index on `("modelId") WHERE status = 'ongoing'` is what makes that safe — at most one open degradation per model, enforced by the database rather than by the write path.

`modelId` is widened to `string`, matching the choice already made in `degradable_models.ts`: models can come dynamically from GCS, and a model dropped from the catalog must not make its past degradations unreadable.

## Tests

No automated tests: the table has no reader yet, and per [TEST1] we test at the endpoint level — the endpoints that use it are in the PRs above.

Exercised by hand against a local database:

- `startDegradation` is idempotent — calling it twice for a model returns the same row rather than opening a second one;
- `resolveDegradation` stamps `endedAt` / `status` and keeps the row, and is a no-op when the model is not degraded;
- a second incident on the same model opens a new row beside the resolved one, so history accumulates;
- a direct insert of two `"ongoing"` rows for one model is rejected by `model_degradations_model_id_ongoing_unique_idx`.

The existing `enabled_models`, `models` and `assistant` suites (65 tests) still pass.

## Risk

Low. A new table with no reader and no writer in this PR: nothing in the running system touches it until the PR above lands. Rollback is reverting the code — the empty table can stay.

## Deploy Plan

Pre-deploy migration (`CREATE TABLE`, `CREATE INDEX CONCURRENTLY`), safe to apply before the code goes live. Ship this PR before the ones above it, which are what start writing to the table.

### DIFF
--- a/front/admin/db.ts
+++ b/front/admin/db.ts
@@ -113,6 +113,7 @@ import {
 } from "@app/lib/resources/storage/models/labs_transcripts";
 import { MembershipModel } from "@app/lib/resources/storage/models/membership";
 import { MembershipUpgradeRequestModel } from "@app/lib/resources/storage/models/membership_upgrade_requests";
+import { ModelDegradationModel } from "@app/lib/resources/storage/models/model_degradations";
 import { OnboardingTaskModel } from "@app/lib/resources/storage/models/onboarding_tasks";
 import { PluginRunModel } from "@app/lib/resources/storage/models/plugin_runs";
 import { ProgrammaticUsageConfigurationModel } from "@app/lib/resources/storage/models/programmatic_usage_configurations";
@@ -254,6 +255,7 @@ export function loadAllModels() {
     FeatureFlagModel,
     GlobalFeatureFlagModel,
     KillSwitchModel,
+    ModelDegradationModel,
     LabsTranscriptsConfigurationModel,
     LabsTranscriptsHistoryModel,
     PluginRunModel,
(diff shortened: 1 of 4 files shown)


Nils-Fedrigo:
Not needed — the stack keeps storing degraded models in `kill_switches` as before.

PR #31338, opened 2026-08-28
[front] enh: show selected model tier in model picker

aubin-tchoi:
## Description

Closes https://github.com/dust-tt/tasks/issues/10235

This PR adds a chip in the model picker, on the same row as the model name to show which model tier the selection maps to: Basic, Standard, or Premium.

image

image

Aligned to the right because the current implementation does not support aligning something to the left after the label. Can be revisited in a follow up PR if important.

## Tests

- Tested locally + preview.

## Risk

- Low.

## Deploy Plan

- Deploy front.


Nils-Fedrigo (approved):
see comments in thread, and thanks!

PopDaph (commented):
LGTM 👍🏻

PR #31341, opened 2026-08-28
Skills: Store manually requested spaces 2/4: add the column and write it

fabiencelier:
## Description:

Adds `manuallyRequestedSpaceIds` to `skill_configurations` and `skill_versions`, and writes it on every path that touches a skill's spaces (create, update, space deletion), so the spaces a person picked by hand under Data and access are recorded instead of inferred. The old inference (`requestedSpaceIds` minus the spaces the skill's tools, knowledge and nested skills require) cannot represent a space that is both picked by hand and required automatically, which is how a manual selection gets forgotten.

The column is a subset of `requestedSpaceIds`, which stays the single source of truth for access checks. Reads still go through the inference in this PR

The backfill computes the derived set with the same resource-level code the write paths use under an Authenticator holding every group of the workspace so that nothing in a restricted space is hidden from it, and covers historical versions as well. 

Versions snapshot their tools and instructions but not their attached knowledge, so on a version a space that only knowledge required reads as hand-picked. Version rows are only displayed (history endpoint, Poke) and never written back into a skill, so that imprecision stays cosmetic, and it errs toward listing the space rather than hiding it.

## Tests

 - added unit tests
 - tested the backfill locally

## Deploy Plan
- merge and lock
- Run the pre-deploy migration `20260828113937_add_manually_requested_space_ids_to_skills.sql` 
- Deploy front.
- Run the backfill **after** the deploy, so live writes cannot race it:
   `npx tsx front/migrations/20260828_backfill_skill_manually_requested_spaces.ts --execute`

### DIFF
--- a/front-api/routes/w/[wId]/skills/[sId]/index.ts
+++ b/front-api/routes/w/[wId]/skills/[sId]/index.ts
@@ -386,7 +386,15 @@ app.patch(
       }
 
       additionalRequestedSpaceIds = additionalRequestedSpaceIdsRes.value;
+    } else if (skill.manuallyRequestedSpaceIds.length > 0) {
+      // The skill has a stored manual list: keep it verbatim.
+      additionalRequestedSpaceIds = [...skill.manuallyRequestedSpaceIds];
     } else {
+      // TODO(skills-manual-spaces): drop this inference once every skill has been backfilled. It
+      // cannot tell a space that was picked by hand from one a tool or knowledge happens to
+      // require, which is why `manuallyRequestedSpaceIds` is now stored. An empty stored list is
+      // ambiguous until then — either never written, or written empty — and both resolve to the
+      // same answer here, since a skill with no manual spaces requests only what it derives.
       const previousAttachedKnowledge = await skill.getAttachedKnowledge(auth);
       const previousComputedRequestedSpaceIds =
         await SkillResource.computeRequestedSpaceIds(auth, {
@@ -485,6 +493,7 @@ app.patch(
       instructions: body.instructions,
       instructionsHtml: body.instructionsHtml,
       availability: requestedAvailability,
+      manuallyRequestedSpaceIds: additionalRequestedSpaceIds,
       mcpServerViews,
       name,
       reinforcement: body.reinforcement,
--- a/front/migrations/20260828_backfill_skill_manually_requested_spaces.ts
+++ b/front/migrations/20260828_backfill_skill_manually_requested_spaces.ts
@@ -1,0 +1,310 @@
+import { getReferencedSkillSpaceModelIds } from "@app/lib/api/skills/space_requirements";
+import { Authenticator } from "@app/lib/auth";
+import {
+  SkillConfigurationModel,
+  SkillVersionModel,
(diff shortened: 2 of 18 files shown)


philipperolet (commented):
quick q: why do we need the column both on configurations and versions? while requestedSpaceIds is only on configurations?

philipperolet:
> quick q: why do we need the column both on configurations and versions? while requestedSpaceIds is only on configurations?

answered IRL 🙏

front-api/routes/w/[wId]/skills/[sId]/index.ts
```diff
@@ -386,7 +386,15 @@ app.patch(
       }
 
       additionalRequestedSpaceIds = additionalRequestedSpaceIdsRes.value;
+    } else if (skill.manuallyRequestedSpaceIds.length > 0) {
+      // The skill has a stored manual list: keep it verbatim.
+      additionalRequestedSpaceIds = [...skill.manuallyRequestedSpaceIds];
```
Comment (philipperolet) [inline front-api/routes/w/[wId]/skills/[sId]/index.ts:391]: two questions: 
- is additionalRequestedSpaceIds supposed to represent manuallyRequestedSpaceIds; if yes or 90%+ yes, let's uniformize the names;
- is the contract of patch to take a diff of fields that change, or the whole new version? on the former case, I understand this line; on the latter, I'm confused

front/migrations/20260828_backfill_skill_manually_requested_spaces.ts
```diff
@@ -0,0 +1,310 @@
...
+  // Every group of the workspace, so the resource layer hides nothing: skills in restricted spaces
+  // are the ones whose manual selections matter most.
+  const auth = await Authenticator.internalAdminForWorkspace(workspace.sId, {
+    dangerouslyRequestAllGroups: true,
+  });
+
+  // Skipping the permission filter is what lets the repair reach a skill that requests a space
+  // which no longer exists: `baseFetch` drops those, and they are precisely the rows whose stale
+  // ids need dropping from the manual list rather than being copied into it.
+  const skills = await SkillResource.fetchByModelIds(auth, skillModelIds, {
```
philipperolet, line 235: i think this silently defaults to status: "active". archived skills will lose their manual-space provenance when they are restored or activated.

front/migrations/20260828_backfill_skill_manually_requested_spaces.ts
```diff
@@ -0,0 +1,310 @@
...
+}
+
+/**
+ * `requestedSpaceIds` minus what the content requires, restricted to spaces that still exist.
+ *
+ * A space that no longer exists must not become a manual choice: `canReadRequestedSpaces` requires
+ * every requested space to resolve, so a dangling id would hide the skill from everyone once the
+ * manual list starts feeding `requestedSpaceIds`.
+ */
+function classifyManualSpaceIds({
```
philipperolet, line 60: here afaict if we are not sure if a space was both from knowledge and manually selected we should mark it manually selected; but it seems we leave it out?

philipperolet (commented):
LGTM modulo 2 coms worth checking

PR #31342, opened 2026-08-28
Skills: Store manually requested spaces 3/4: read the stored list

fabiencelier:
## Description:

Deletes the inference that guessed which requested spaces had been picked by hand, in the skill update route and in the space-deletion cleanup, and reads `manuallyRequestedSpaceIds` instead. Also serializes the field so the Skill Builder can stop guessing it in 4/4. A space required by a tool, knowledge or a nested skill now disappears when nothing needs it any more, while a hand-picked one stays until someone removes it.

Two existing tests set up their state by writing `requestedSpaceIds` without the manual list, which no write path does any more; their setup now records the manual space explicitly.

## Tests

new test cases

## Deploy plan

Need to run the migration for backfilling the manual spaces first
Deploy front

### DIFF
--- a/front-api/routes/w/[wId]/skills/[sId]/index.ts
+++ b/front-api/routes/w/[wId]/skills/[sId]/index.ts
@@ -386,35 +386,9 @@ app.patch(
       }
 
       additionalRequestedSpaceIds = additionalRequestedSpaceIdsRes.value;
-    } else if (skill.manuallyRequestedSpaceIds.length > 0) {
-      // The skill has a stored manual list: keep it verbatim.
-      additionalRequestedSpaceIds = [...skill.manuallyRequestedSpaceIds];
     } else {
-      // TODO(skills-manual-spaces): drop this inference once every skill has been backfilled. It
-      // cannot tell a space that was picked by hand from one a tool or knowledge happens to
-      // require, which is why `manuallyRequestedSpaceIds` is now stored. An empty stored list is
-      // ambiguous until then — either never written, or written empty — and both resolve to the
-      // same answer here, since a skill with no manual spaces requests only what it derives.
-      const previousAttachedKnowledge = await skill.getAttachedKnowledge(auth);
-      const previousComputedRequestedSpaceIds =
-        await SkillResource.computeRequestedSpaceIds(auth, {
-          mcpServerViews: skill.mcpServerViews,
-          attachedKnowledge: previousAttachedKnowledge,
-        });
-      const previousReferencedSkillSpaceIds =
-        await getReferencedSkillSpaceModelIds(
-          auth,
-          skill.instructions,
-          skill.sId
-        );
-      const previousComputedRequestedSpaceIdsSet = new Set([
-        ...previousComputedRequestedSpaceIds,
-        ...previousReferencedSkillSpaceIds,
-      ]);
-
-      additionalRequestedSpaceIds = skill.requestedSpaceIds.filter(
-        (spaceId) => !previousComputedRequestedSpaceIdsSet.has(spaceId)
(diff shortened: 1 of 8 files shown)


front/lib/api/spaces.ts
```diff
@@ -413,7 +398,7 @@ export async function softDeleteSpaceAndLaunchScrubWorkflow(
           });
         const requestedSpaceIds = uniq([
           ...computedRequestedSpaceIds,
-          ...additionalRequestedSpaceIds,
+          ...manuallyRequestedSpaceIds,
```
philipperolet: do we also update manuallyRequestedSpaceIds fields on space deletions?

front-api/routes/w/[wId]/skills/[sId]/index.ts
```diff
@@ -386,35 +386,9 @@ app.patch(
...
-      const previousComputedRequestedSpaceIdsSet = new Set([
-        ...previousComputedRequestedSpaceIds,
-        ...previousReferencedSkillSpaceIds,
-      ]);
-
-      additionalRequestedSpaceIds = skill.requestedSpaceIds.filter(
-        (spaceId) => !previousComputedRequestedSpaceIdsSet.has(spaceId)
-      );
+      // A request that says nothing about the spaces leaves the manual selection as it is.
+      additionalRequestedSpaceIds = [...skill.manuallyRequestedSpaceIds];
```
Comment (philipperolet) [inline front-api/routes/w/[wId]/skills/[sId]/index.ts:393]: ooc what's the difference between additionalRequestedSpaceIds and manuallyRequestedSpaceIds?
I get the feeling they're the same, in which case making naming consistent could be worthwile?

front/lib/api/spaces.ts
```diff
@@ -413,7 +398,7 @@ export async function softDeleteSpaceAndLaunchScrubWorkflow(
           });
         const requestedSpaceIds = uniq([
```
philipperolet, line 415: here I think we drop nested skill requirements; i.e. 
  - Parent skill uses a tool from Space A.
  - Parent references a child skill requiring Space B.
  - Nothing was selected manually.
  - Parent stores requestedSpaceIds = [A, B] and manuallyRequestedSpaceIds = [].


  When Space A is deleted:
  - Before this PR: A was recognized as tool-derived and removed; B looked like an additional requirement and remained. Result: [B].
  - After this PR: the remaining tools require nothing, and there are no manual spaces. Result: [].

front-api/routes/w/[wId]/skills/[sId]/index.ts
Comment (philipperolet) [reply front-api/routes/w/[wId]/skills/[sId]/index.ts:393]: ooh I think additionalReqSpIds included nested skill requirements
in which case they make sense but code doesn't make it very clear, maybe we should clarify: manual, tools, knowledge, child, if there's a good solution?

philipperolet (commented):
Thanks 🙏 Left a few coms

front/lib/api/spaces.ts
fabiencelier: Yes just a few line before
```
        const manuallyRequestedSpaceIds =
          skill.manuallyRequestedSpaceIds.filter(
            (spaceId) => spaceId !== space.id
          );
```

front/lib/api/spaces.ts
fabiencelier: Indeed, fixed now and added a test
Thank you for catching this 👍

front-api/routes/w/[wId]/skills/[sId]/index.ts
Comment (fabiencelier) [reply front-api/routes/w/[wId]/skills/[sId]/index.ts:393]: I have added comment to clarify

philipperolet (approved):

PR #31343, opened 2026-08-28
Skills: Store manually requested spaces 4/4: seed the builder from the stored list

fabiencelier:
## Description:
closes https://github.com/dust-tt/tasks/issues/9896

The Skill Builder derived its manual space selection by subtracting the spaces the skill currently requires from the ones it requested, which forgot any space that both a person and some knowledge had asked for. The form now starts from `manuallyRequestedSpaceIds` as served by the API, which also removes the hydration effect that had to wait for the skill's knowledge and nested skills to load before it could guess.

This closes the reported bug: a manually selected space stays under Data and access until someone removes it, and an automatically required one disappears when nothing in the skill needs it any more.

## Tests

Added UI tests

## Deploy plan

Deploy front.

### DIFF
--- a/front/components/skill_builder/SkillBuilderVersionComparisonFooter.tsx
+++ b/front/components/skill_builder/SkillBuilderVersionComparisonFooter.tsx
@@ -29,6 +29,14 @@ export function SkillBuilderVersionComparisonFooter() {
     setValue("fileAttachments", compareVersion.fileAttachments, {
       shouldDirty: true,
     });
+    // The version's hand-picked spaces.
+    // The spaces the current knowledge/tools/skills require are recomputed on save
+    // and are unaffected by this.
+    setValue(
+      "additionalSpaces",
+      compareVersion.manuallyRequestedSpaceIds ?? [],
+      { shouldDirty: true }
+    );
     exitDiffMode();
   };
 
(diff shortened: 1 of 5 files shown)


front/components/skill_builder/SkillBuilderVersionComparisonFooter.tsx
```diff
@@ -29,6 +29,14 @@ export function SkillBuilderVersionComparisonFooter() {
     setValue("fileAttachments", compareVersion.fileAttachments, {
       shouldDirty: true,
     });
+    // The version's hand-picked spaces.
+    // The spaces the current knowledge/tools/skills require are recomputed on save
+    // and are unaffected by this.
+    setValue(
+      "additionalSpaces",
+      compareVersion.manuallyRequestedSpaceIds ?? [],
```
philipperolet, line 37: quick question: the backfilled versions are unreliable because they did not snapshot attached knowledge; a knowledge-only space could therefore be classified as manually selected? if correct, leaving it to you to decide if important or not and what to do

philipperolet (approved):
LGTM modulo 1 com; pre-approving, handle as you see fit 👍

front/components/skill_builder/SkillBuilderVersionComparisonFooter.tsx
fabiencelier: I think it's fine
Rollback of skill does not happen often and this may just add some spaces that are not wanted and the user can just remove them.
Also this is just legacy versions, newly created versions will have the correct manual spaces

PR #31344, opened 2026-08-28
[front] enh: remove reasoning-effort tooltips in model picker, shorten model names

aubin-tchoi:
## Description

This PR does some UI polish on the model picker.

1. Remove the suffix ` (Fireworks)` to reclaim some space on the row

2. Remove the tooltip on model reasoning effort selection (only kept for the locked state, in which case they serve the purpose of explaining why it's locked).

3. Update the copy of the label shown to explain users they don't have access to premium models to remove references to "Picker" (too technical) and "usage-based plans" (unclear also end users do not choose the plan).

5. Fix how provider names are displayed in the slash command dropdown, previously showed the `snake_case` key we use in the back-end.

image

replaced with:

image

## Tests

- Tested locally + preview.

## Risk

- Low.

## Deploy Plan

- Deploy front.


front/components/model_picker/ReasoningEffortSlider.tsx
```diff
@@ -111,39 +93,8 @@ export function ReasoningEffortSlider({
...
     </div>
   );
-
-  if (!isDisabled) {
-    return slider;
-  }
-
-  const soleEffort = unlockedStops[0]?.effort;
-  const tooltipLabel = planLockReason
-    ? getModelLockTooltip(planLockReason)
```
PopDaph, line 141: I think we want to keep this to show what is available to a bigger plan on Dust 🙏🏻

aubin-tchoi: agreed, I pinged for review too early, still changing a few things because I think this condition here actually has nothing to do with your plan and the message is not correct (still trying to get to the bottom of it)

front/components/model_picker/modelPickerUtils.ts
```diff
@@ -132,18 +131,6 @@ export function getDefaultTierId(
...
-// Per reasoning-effort blurbs surfaced in the effort slider tooltip.
-const REASONING_EFFORT_INFO: Record<ReasoningEffort, string> = {
-  none: "No additional reasoning, for the fastest responses.",
-  light: "Light reasoning effort, faster responses.",
-  medium: "Medium reasoning effort, balancing speed and quality.",
-  high: "High reasoning effort, longer wait times but higher quality.",
-};
-
-export function getReasoningEffortLabel(effort: ReasoningEffort): string {
-  return effort === "none" ? "None" : capitalize(effort);
```
aubin-tchoi, line 144: inlined this, `capitalize("none")` is `"None"`

front/components/model_picker/ReasoningEffortSlider.tsx
aubin-tchoi: cleaned up a bit, this part was actually misleading because it put a single tooltip for the entire slider, even though the restrictions apply differently to each reasoning effort (high can be locked but not medium)

consolidated by only passing tooltips on each reasoning effort, and always returning a `SliderSteps` directly

adrsimon (approved):
LGTM 👍

PopDaph (approved):
Thank you 🙏🏻

PR #31345, opened 2026-08-28
Spaces members: fix issue where some members could be dropped from space on save

fabiencelier:
## Description

closes https://github.com/dust-tt/tasks/issues/10237

### What could happen

`PATCH /api/w/:wId/spaces/:spaceId/members` takes the **absolute** member list and diffs it server-side (`GroupResource.dangerouslySetMembers`): anyone not in the payload has their group membership ended. So a truncated payload silently removes members.

The space settings modal could build exactly such a truncated payload:

  1. `MemberSelectionTable` keeps an `sId -> user object` cache (`userMapRef`) that is seeded **once, at mount**, from `initialMembers` (`spaceInfo.members`).
  2. On every selection change it converted the (complete) set of selected ids back into user objects, **dropping any id the cache couldn't resolve**, and `RestrictedAccessBody` discarded the ids in favour of those objects. `onSave`  then re-derived `memberIds` from them — a lossy `ids → objects → ids` round trip.
  3. Since #31094 `RestrictedAccessBody` renders unconditionally instead of behind `isRestricted` (which was only ever true after `spaceInfo` had loaded), so the table can now mount while `useSpaceInfo({ includeAllMembers: true })` is still in flight — with an **empty** cache that is never re-seeded.

Result: on a space whose `GET` is slow (many members, categories + usage), an admin who opens the settings and ticks one new member before the fetch lands saves a member list reduced to whoever happened to pass through the search results. Fast / small workspaces never hit the window, which is why this doesn't reproduce easily.

### The two fixes

**1. Don't render the access section until the space info has loaded.**
`CreateOrEditSpaceModal` now gates `RestrictedAccessHeader` + `RestrictedAccessBody` on `isLoadingSpaceInfo` (a spinner in the meantime), so the table always mounts with a fully seeded cache. This also stops the modal from briefly showing an empty member list and an "unrestricted" toggle for a restricted space.

**2. Make ids the source of truth for the selection.**
`CreateOrEditSpaceModal` holds `selectedMemberIds: Set<string>` instead of `SearchMemberType[]`, and `RestrictedAccessBody` forwards the ids the table emits rather than the resolved objects. The user objects are now only used for rendering rows, so a cache miss can at worst cost a missing row — it can no longer drop a member from the saved payload. (Same pattern `GroupDialog` already uses.)

## Tests

not really tested as it is hard to reproduce
just tested that the UI still work as expected

## Risk

impact is only on the UI to manage space members

## Deploy Plan

deploy front

### DIFF
--- a/front/components/spaces/CreateOrEditSpaceModal.tsx
+++ b/front/components/spaces/CreateOrEditSpaceModal.tsx
@@ -1,5 +1,4 @@
 import { ConfirmContext } from "@app/components/Confirm";
-import type { SearchMemberType } from "@app/components/members/MemberSelectionTable";
 import { ConfirmDeleteSpaceDialog } from "@app/components/spaces/ConfirmDeleteSpaceDialog";
 import { RestrictedAccessBody } from "@app/components/spaces/RestrictedAccessBody";
 import { RestrictedAccessHeader } from "@app/components/spaces/RestrictedAccessHeader";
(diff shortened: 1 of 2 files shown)


davidebbo (approved):
LGTM

PR #31348, opened 2026-08-28
[front] poke: show conversation wake-ups

davidebbo:
Convo shows wake-ups:
Screenshot 2026-08-28 at 13 32 36

And system message is tagged:
Screenshot 2026-08-28 at 13 32 41

### Description

Poke had no way to see wake-ups. Scheduled ones were entirely invisible, and a fired wake-up rendered as an ordinary user message with nothing marking where it came from — the only trace of the feature anywhere in the Poke tree was `WakeUpResource.deleteAllForUser` in the scrub activity.

This adds a wake-ups panel to the Poke conversation page and a badge on wake-up-triggered messages.

**Endpoint** — `GET /api/poke/workspaces/:wId/conversations/:cId/wakeups` lists every wake-up on the conversation in *any* status (scheduled, fired, cancelled, expired): Poke is an inspection surface, so the terminal ones are what make the history readable. It fetches the conversation with `includeDeleted: true`, like the sibling `config` route, so deleted conversations still show theirs.

**Panel** — a `Collapsible` in the conversation page's right-hand aside, mirroring `PokeConversationConsumptionInspector` (including `disabled: !isOpen`, so nothing is fetched while collapsed). Each entry shows the reason, a status chip, the human schedule phrase from `describeWakeUpSchedule`, raw cron + timezone for cron schedules, next fire time, `fireCount / maxFires`, owner, agent sId, wake-up sId and creation time. Next-fire is computed only for `scheduled` wake-ups — for terminal ones the cron parser would happily keep projecting firings that will never happen.

**Badge** — `UserMessageView` now renders a `wake-up` chip when `message.context.origin === "wakeup"`.

Read-only: no cancel plugin here. `WakeUpResource.cancel` already lets any workspace admin cancel, so whether Poke operators should be able to do it — and through `cancel` or `forceCancel` — felt worth deciding separately.

### Risk

Low. Additive: one new Poke endpoint, one new panel behind a collapsed disclosure, one badge. No change to wake-up behaviour or to any non-Poke surface.

### Deploy Plan

Deploy as usual.

### DIFF
--- a/front-api/routes/poke/workspaces/[wId]/conversations/[cId]/index.ts
+++ b/front-api/routes/poke/workspaces/[wId]/conversations/[cId]/index.ts
@@ -10,6 +10,7 @@ import config from "./config";
 import consumption from "./consumption";
 import reinforcementTestCase from "./reinforcement_test_case";
 import render from "./render";
+import wakeups from "./wakeups";
 
 export type PokeGetConversationResponseBody = {
   conversation: PokeConversationType;
@@ -43,5 +44,6 @@ app.route("/config", config);
 app.route("/consumption", consumption);
 app.route("/reinforcement_test_case", reinforcementTestCase);
 app.route("/render", render);
+app.route("/wakeups", wakeups);
 
 export default app;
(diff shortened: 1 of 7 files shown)


spolu (commented):
LGTM 🙏

PR #31339, opened 2026-08-28
[front] feat: show excess credit consumption for no-pool workspaces

avervaet:
Workspaces without an active Metronome credit pool (PAYG-only) previously
had the whole "Workspace credit pool" section hidden. Show a live,
ES-derived "consumed this cycle" figure (the same sum backing the members
table's credits usage column) and a cycle-history table sourced from past
invoices' cpu_conversion line items as a fallback when there's no pool
ledger to read from.

PR #31241, opened 2026-08-27
[front] feat: sort usage table by seat usage, fix credit numbers

avervaet:
## Description

Small set of usage-page improvements:

- You can now sort the members table by how much of their seat allowance someone has used, not just by credits used.
- Fixed the credits sort so it ranks people by how much they're actually drawing from the shared credit pool, instead of their total usage. This was pushing normal paid users above people actually eating into the pool.
- Fixed a few bugs where the workspace's pool credit numbers (this cycle's usage, past cycles, expired credits) could show up wrong or missing, so the numbers on the page are now accurate.

## Tests

Type-checked, formatted, and checked manually on the usage page and the Poke pool usage page.

## Risk

Low. UI and number-accuracy fixes, no change to how credits are actually billed.

## Deploy Plan

Deploy front. No coordination needed.

PR #31155, opened 2026-08-26
[front] feat: Poke workspace credit pool usage page

avervaet:
## Summary
- Add a read-only **Pool Usage** page to Poke at `/poke/:wId/pool-usage`, showing the redesigned members usage table (compact variant) and the workspace credit pool summary.
- Add a "Pool Usage" nav card below the Analytics card on the workspace page.
- Surface `creditsResetAt` from `usePokeMembersUsage` (already returned by the API, just not exposed) and export `PokeCreditPoolCard` for reuse.

No new API routes were needed — the poke-scoped `credits/members-usage` and `credits/awu-pool-summary` endpoints already authorize admins to view any workspace.

## Test plan
- [ ] `npx tsgo --noEmit` passes (verified)
- [ ] Biome check passes (verified via pre-commit hook)
- [ ] Manually load `/poke/:wId/pool-usage` for a workspace with members and verify the pool summary and members table render correctly
- [ ] Verify the "Pool Usage" card on the workspace page links to the new page

PR #31134, opened 2026-08-25
[front] feat: rework redesigned usage page members table

avervaet:
## Description

Reworks the redesigned Usage page's members table to match the latest Figma:

- Drops the Groups column.
- Renames the "Models tier" column header to "Models" and drops the "Up to " prefix from the tier summary (e.g. "Premium" instead of "Up to Premium"). The tier-selection menu items are unaffected — there "Up to X" still clarifies the selection covers every tier up to and including X.
- Splits the combined seat cell into two columns: an icon-only "Seats" column (hover tooltip recaps the seat name, or the existing scheduled-change message when a change is pending) and a new "Seat usage" column showing a ring + percentage of the seat allowance consumed, turning red when usage exceeds the allowance.
- Adds a coins icon next to the "Credits usage" column header.

These changes are gated to the redesigned table only via a new `variant` prop (`"legacy" | "compact"`) on `MembersUsageTable`; the current (flag-off) usage page keeps its existing columns unchanged.

Only touches `UsagePageRedesign`, still behind the `usage_page_redesign` flag from #31125.

Stacked on #31127.

## Tests

Type-checked and formatted.

## Risk

Low. Scoped to the flagged redesign component via the new `variant` prop; the legacy page and default (flag-off) experience are unaffected.

## Deploy Plan

Deploy front. No coordination needed; still gated behind `usage_page_redesign`.

PR #31349, opened 2026-08-28
[front] feat: add Programmatic usage card to the workspace credit pool section

avervaet:
Consumption from API keys, webhooks, and workflow/agent triggers draws
from the same pool but has no member to attribute it to, so it never
shows up in the members table — leaving an unexplained gap between the
pool total and the sum of member rows. Surfaces it as its own figure
next to "Consumed this cycle" on both the customer usage page and its
Poke mirror.


avervaet:

- #31125
    - #31126
        - #31134
            - #31155
                - #31241
                    - #31339
                        - #31349 ◀

PR #31352, opened 2026-08-28
[front] Ungate conversation credit breakdown

flvndvd:
## Description

This PR ungates the conversation credit breakdown. The message credits breakdown remains gated behind FF until we can display the full cost (including sub agents).

## Risk

Low. This only broadens access to an existing read-only aggregate endpoint and UI panel. The message-level feature remains gated. Rollback is reverting this commit.

## Deploy Plan

Deploy normally. No migration or coordinated rollout is required.

### DIFF
--- a/front-api/public/swagger.json
+++ b/front-api/public/swagger.json
@@ -6825,9 +6825,6 @@
               }
             }
           },
-          "403": {
-            "description": "The workspace does not have access to consumption details"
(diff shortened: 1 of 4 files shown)


achilleburah (approved):

PopDaph (commented):
👍🏻

PR #31354, opened 2026-08-28
[front] Consolidate duplicate day/hour millisecond constants

avervaet:
## Description

`DAY_MS` / `MS_PER_DAY` / `ONE_DAY_MS` (24 * 60 * 60 * 1000) was redefined locally in ~15 files across components, scripts, temporal, and lib/api. Copy-pasted constants like this spread fast with AI-assisted edits, each new file tends to redefine the constant instead of finding the existing one, so the duplication multiplies over time instead of staying at one bad instance. 

## Risk

Pure refactor, no behavior change: all sites now point at the same constant value they already had inline.


avervaet:

- #31354 ◀

aubin-tchoi (approved):
LGTM

PR #31359, opened 2026-08-28
[front] enh: tweak description of sandbox + google drive tool

aubin-tchoi:
## Description

This PR tweaks the description of the sandbox skill + the `get_file_content` tool from Google Drive.
Addresses a specific flow observed where the goal is to replace a custom agent with a skill. The agent uses tables query on a Google Sheet, the skill port fails in 2 different ways: sometimes fails to extract sheet IDs from URLs, and sometimes fails to enable the Computer skill to process the data.

## Deploy Plan

- Deploy front.


fontanierh (approved):
LGTM, thanks

PR #31360, opened 2026-08-28
[front] Group ACL rules by kind; permission-based group management

tdraier:
Stacked on top of #31322 (previous PR removed the misuse of `canRead` on group listing/fetching). This PR turns permission checks on groups into consistent, kind-driven rules and rewires callers accordingly.

## What

- **`getAccessControlLists` rewritten by kind** so `canRead`/`canWrite`/`canAdministrate` mean the same thing at every call site:
  - `regular_auto`, `system` → deny-all. `regular_auto` access is gated through the associated resource (space/agent), never on the group.
  - `regular_manual` → admin+manager get `read/write/admin`; every role gets `read`.
  - `provisioned` → every role `read`; nobody can `write`/`admin` (directory-synced).
  - `global` → roles only, no member self-grant: everyone `read`, no `write`/`admin`.
  - `agent_editors` → **unchanged** (separate governance follow-up).

- **Group-management moved from role checks to group permissions** (`write` = edit members, `admin` = rename/delete):
  - `updateRegularManualGroup` / `updateRegularManualGroupMembers` → `canWrite`
  - `deleteRegularManualGroup` → `canAdministrate`
  - `updatePoolCap` drops its internal `isManager`; the route already gates via `ensureIsManager` (same pattern as user/global spend caps).

- **`updateName` split** so system-of-record renames don't fake a permission:
  - `updateName(auth, newName)` keeps the `hasPermission("admin", this)` check for user-facing renames (`updateRegularManualGroup`).
  - `dangerouslyUpdateName(newName)` renames a `regular_auto` or `provisioned` group as a side effect of its backing resource — space rename cascade and SCIM directory sync. Caller owns authorization.
  - Fixes a **pre-existing latent bug** where `upsertByWorkOSGroupId` ignored the `Err` returned by `updateName` on `provisioned` groups.

- **`lib/api/spaces.ts`**: replaces the `canWrite` check on the `regular_auto` member group with `space.canAdministrate(auth)`, placed in both management-mode branches (also gates group-based creation, which previously had no such check). Projects continue to self-authorize post-refresh via their editor group.

## Left as follow-ups

- `makeNewRegularManual` still uses `isManager` — needs a type-level `group` `create` capability in `ROLE_REGISTRY` (no `group` entry yet). Marked `TODO(governance)`.
- `fetchWorkspaceSystemGroup` still uses `isAdmin` — script-only accessor; `system` is deny-all so `canRead` can't apply.
- `agent_editors` ACL and its callers untouched (separate governance work).

## Test plan

- [ ] `npx tsgo --noEmit` clean in `front` and `front-api`.
- [ ] Group/space/permission unit suites and route tests pass.
- [ ] Manual: create/rename/delete a `regular_manual` group as admin and as manager; verify SCIM directory rename still syncs a provisioned group's name; rename a space and confirm the member/editor groups get renamed.

### DIFF
--- a/front/lib/resources/group_resource.ts
+++ b/front/lib/resources/group_resource.ts
@@ -2624,11 +2624,7 @@ export class GroupResource extends BaseResource<GroupModel> {
 
     // global, provisioned, regular_auto: read-only for every workspace member.
     // Write/admin are gated through the associated resource (space/agent).
-    if (
-      this.isGlobal() ||
-      this.isProvisioned() ||
-      this.isRegularAuto()
-    ) {
+    if (this.isGlobal() || this.isProvisioned() || this.isRegularAuto()) {
       return [
         {
           roles: [
--- a/front/lib/resources/group_resource.ts
+++ b/front/lib/resources/group_resource.ts
@@ -449,6 +449,8 @@ export class GroupResource extends BaseResource<GroupModel> {
       >
     >
   > {
+    // TODO(governance): gate on a type-level `group` "create" capability
+    // (ROLE_REGISTRY entry) instead of the role, in a follow-up PR.
     if (!auth.isManager()) {
       return new Err(
         new DustError(
@@ -997,7 +999,7 @@ export class GroupResource extends BaseResource<GroupModel> {
 
     if (group) {
       const groupResource = new this(this.model, group.get());
-      await groupResource.updateName(auth, directoryGroup.name);
+      await groupResource.dangerouslyUpdateName(directoryGroup.name);
       return groupResource;
     }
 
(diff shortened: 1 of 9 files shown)


tdraier:

- #31360 ◀

front/lib/resources/group_resource.ts
```diff
@@ -2537,53 +2561,43 @@ export class GroupResource extends BaseResource<GroupModel> {
...
-            {
-              id: this.id,
-              permissions: ["read"],
-            },
-          ],
           roles: [
-            { role: "admin", permissions: ["read", "write", "admin"] },
+            { role: "admin", permissions: ["read"] },
             { role: "manager", permissions: ["read"] },
+            { role: "user", permissions: ["read"] },
```
fabiencelier, line 2633: Is it safe that anybody can read the provisioned groups ?
It means anybody can list the members of any provisioned group 
Not necessarily an issue but flagging just in case

tdraier: Anybody should be able to get the provisioned groups to select them, so having read would be required for that. The issue is that getting members is gated on the same permission/verb, and it's indeed possible to get members of all groups you can read through https://dust.tt/api/w/0ec9852c2f/groups?withMembers=true , without having the builder role. No other endpoint gives the members for non-managers, but this one leaks the members.

2 options : 
- For non-manager callers, only return memberIds for groups the caller is a member of (gated at endpoint level, easy) - or don't return members at all ( that may break some analytics groups filtering feature for non managers).
- New verb "read_members", enforced at resource-level, granted in acl. But granting a role for a group on itself requires the "groups" attributes in acl (that we're trying to remove), or a group-permission entry. More complex.

matteotrab: Discussed IRL, see https://dust4ai.slack.com/archives/C0B2L7PT56C/p1788165454516309 for more information

matteotrab (approved):

PR #31361, opened 2026-08-28
[front] Drop redundant checks in createPendingAgentConfiguration

tdraier:
## Description

Stacked on top of #31360.

Migrate the last callers off `agent_editors` group-permission checks, then drop the special-case branch from `GroupResource.getAccessControlLists`. Also clean up two redundant checks around agent creation.

Refs [dust-tt/tasks#10202](https://github.com/dust-tt/tasks/issues/10202).

Callers migrated (all in the agent-editors area):

- **`front/lib/api/assistant/configuration/agent.ts:887`** — removed the `!group.canAdministrate(auth) && auth.user()` block. Authorization is already asserted a few lines above with `editors.some((e) => e.id === authorId) || isAdmin(owner)`.
- **`front/lib/api/assistant/configuration/agent.ts:1533, :1552`** (`updateAgentPermissions`) — replaced `!editorGroupRes.value.canAdministrate(auth)` with an inline `!auth.isAdmin() && !auth.hasGroupByModelId(editorGroupRes.value.id)`. Semantically equivalent to the old `agent_editors` ACL (admin role OR group member).
- **`front-api/.../[aId]/editors.ts:112`** (GET agent editors) — removed the `canRead` block entirely. Any workspace member can read agent editors, and workspace membership is already enforced by the route middleware.
- **`front-api/.../[aId]/editors.ts:206`** (PATCH agent editors) — replaced `canAdministrate` with the same inline `isAdmin || hasGroupByModelId` check.

Follow-up marker (`TODO(governance)`) kept on the three replaced checks, pointing at the eventual move to `agent.canRead` / `agent.canAdministrate` once the agent-permission model lands.

Then, since no caller checks permissions on `agent_editors` groups anymore, drop the `agent_editors` case from `GroupResource.getAccessControlLists`. `agent_editors` groups now fall through to the deny-all branch, consistent with `regular_auto` and `system` (gate on the associated resource, not the group).

Two smaller cleanups bundled in:

- **`createPendingAgentConfiguration`** — dropped the tautological `group.canWrite(auth)` check and the no-op `dangerouslySetMembers(auth, { users: [user] })` right after `makeNewAgentEditorsGroup`. The author is added as a member by `makeNew(..., { memberIds: [authorId] })` inside the same transaction, and `auth.refresh({ transaction: t })` runs just before the removed check — so the check couldn't fail without a deeper bug and the re-set rewrote the same member set.
- **Three no-op `canRead` filters** in `GroupResource` (`findEditorGroupForAgent`, `findEditorGroupsForAgents`, `fetchByAgentConfiguration`) — all three fetch by ids from `GroupAgentModel` (only `agent_editors` groups) and `canRead` returned `true` for any workspace role, so the filter never fired meaningfully. `baseFetch` already scopes by workspace, and callers are gated at the request layer.

The `canRead` / `canWrite` / `canAdministrate` **methods themselves** on `GroupResource` are unchanged and remain the accessors for the surviving `regular_manual` / `provisioned` / `global` cases.

## Tests

- `npx tsgo --noEmit` clean in `front` and `front-api`.
- Existing group / agent-editor route tests pass locally.

## Risk

Low. Behaviorally-preserving for standard flows:

- `agent.ts:887` — removed check is redundant with an assertion in the same transaction; can only fail if that assertion was already failing.
- `agent.ts:1533/:1552` and `editors.ts:206` — new inline `isAdmin() || hasGroupByModelId(...)` is the exact expansion of `hasPermission("admin", agent_editors_group)` given the current ACL (admin role verb `admin`, member instance grant `admin`).
- `editors.ts:112` — the removed `canRead` returned `true` for any workspace role and the route requires workspace membership upstream. Non-workspace-member fallthrough is not reachable through the middleware.
- Deleting the `agent_editors` case in `getAccessControlLists` has no live consumers after the migration above; a stray future caller would fall through to deny-all and fail closed rather than open.

Rollback: revert this PR. All changes are self-contained edits to `front/lib/resources/group_resource.ts`, `front/lib/api/assistant/configuration/agent.ts`, and `front-api/routes/w/[wId]/assistant/agent_configurations/[aId]/editors.ts`.

## Deploy Plan

Standard deploy in stack order after #31360 lands. No migrations, no data changes, no feature flags. Nothing to coordinate.

### DIFF
--- a/front-api/routes/w/[wId]/assistant/agent_configurations/[aId]/editors.ts
+++ b/front-api/routes/w/[wId]/assistant/agent_configurations/[aId]/editors.ts
@@ -113,16 +113,7 @@ app.get(
     }
 
     const editorGroup = editorGroupRes.value;
-    if (!editorGroup.canRead(auth)) {
-      return apiError(ctx, {
-        status_code: 403,
-        api_error: {
-          type: "agent_group_permission_error",
-          message: "User is not authorized to read the agent editors.",
-        },
-      });
-    }
-
+    // Any workspace member can read the editors of an agent.
     const members = await editorGroup.getActiveMembers(auth);
     const memberUsers = members.map((m) => m.toJSON());
 
@@ -207,7 +198,11 @@ app.patch(
     }
 
     const editorGroup = editorGroupRes.value;
-    if (!editorGroup.canAdministrate(auth)) {
+    // TODO(governance) replace by permission check on agent resource
+    if (
+      !auth.isAdmin() &&
+      !(await editorGroup.isMember(auth.getNonNullableUser()))
+    ) {
       return apiError(ctx, {
         status_code: 403,
         api_error: {
--- a/front/lib/api/assistant/configuration/agent.ts
+++ b/front/lib/api/assistant/configuration/agent.ts
@@ -135,21 +135,11 @@ export async function createPendingAgentConfiguration(
       { transaction: t }
     );
 
-    const group = await GroupResource.makeNewAgentEditorsGroup(auth, agent, {
+    await GroupResource.makeNewAgentEditorsGroup(auth, agent, {
       transaction: t,
       authorId: user.id,
     });
     await auth.refresh({ transaction: t });
-    if (!group.canWrite(auth)) {
-      throw new DustError(
-        "unauthorized",
-        "User does not have write permission for the agent editors group."
-      );
-    }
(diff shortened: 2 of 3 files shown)


tdraier:

- #31322
    - #31360
        - #31361 ◀

matteotrab (approved):

PR #31362, opened 2026-08-28
[front] Drop unused Authenticator group helpers and inlined shadow-compare

tdraier:
## What

Removes three now-dead members of `Authenticator`, part of the effort to eventually get rid of `auth._groupModelIds`:

- **`hasGroup(groupId: string)`** — no callers anywhere.
- **`groupIds()`** — its only remaining uses were diagnostic `logger` payloads (`group_resource.ts`, `run_dust_app/helpers.ts`, and a scratch script). Those log fields are dropped, and the model-id → sId mapping is inlined into `toJSON` (serialization must stay so `fromJSON` can reconstruct the auth).
- **`shadowComparePermission` / `runShadowComparePermission`** — the shadow-compare copy inlined in `auth.ts` had no callers. The standalone `lib/api/permissions/shadow.ts` module, its tests, and the `group_permissions_shadow` feature flag are untouched.

Also drops the now-unused `normalizeError` import from `auth.ts`.

## Notes

- `_groupModelIds` itself stays — it's still the input that feeds `resolvePermissions`. This PR only removes the dead accessors/logging around it.
- The legacy group path in `hasPermissionForAcl` and the agent/suggestion consumers of `groupModelIds()` are out of scope (separate migrations onto `group_permissions`).

## Test plan

- `npx tsgo --noEmit` in `front` passes.
- `biome check` clean on changed files.

### DIFF
--- a/front/lib/api/actions/servers/run_dust_app/helpers.ts
+++ b/front/lib/api/actions/servers/run_dust_app/helpers.ts
@@ -103,7 +103,6 @@ export async function prepareAppContext(
         // eslint-disable-next-line @typescript-eslint/prefer-nullish-coalescing
         userId: auth.user()?.sId || "no_user",
         role: auth.role(),
-        groupIds: auth.groupIds(),
         actionConfig,
         dustAppConfiguration: actionConfig.dustAppConfiguration,
         appId: actionConfig.dustAppConfiguration?.appId,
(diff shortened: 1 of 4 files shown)


tdraier:

- #31362 ◀

fabiencelier (approved):
LGTM ✅

PR #31365, opened 2026-08-28
[front] fix: distinguish unavailable reasoning efforts from locked ones

aubin-tchoi:
## Description

In the model picker some combinations of model x reasoning effort cannot be selected for various reasons: they can either be prohibited (admin decided that the user can't access to a certain tier of models/workspace does not have access to premium models) or simply not exist.

In both cases we currently show a padlock on the reasoning effort slider, this PR changes that to show a slash circle in the 2nd case instead (keeping the lock in the first case).

Current live version

image

Proposal

image

## Tests

- Updated model picker and SliderSteps story coverage.

## Risk

- Low. Limited to reasoning-effort state presentation.

## Deploy Plan

- Deploy front.

### DIFF
--- a/front/components/editor/extensions/shared/slash_suggestion/buildPickModelSlashCommandItems.ts
+++ b/front/components/editor/extensions/shared/slash_suggestion/buildPickModelSlashCommandItems.ts
@@ -36,12 +36,12 @@ export function getSelectableEffortsForSlashMenu(
     return [];
   }
 
-  const unlocked = getEffortStops(model, { lockPremiumEfforts })
-    .filter((stop) => !stop.locked)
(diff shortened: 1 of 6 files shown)


pinotalexandre:
Can we make those icons white? 
image

Otherwise LGTM

pinotalexandre (approved):

aubin-tchoi:
@pinotalexandre making the icons white make them barely visible in the non-filled bar, should we flip in that case?

pinotalexandre:
> making the icons white make them barely visible in the non-filled bar, should we flip in that case?

Yeah just saw it, yes we should

aubin-tchoi:
we're good now!

PR #31370, opened 2026-08-28
Build Frames v2 UI publication artifacts

fontanierh:
## Description

- Build the Frame UI from the captured publication snapshot with the existing Frame bundler.
- Store the immutable bundle at `ui/bundle.js` before the manifest commit marker.
- Keep the active publication unchanged when the UI build fails.

## Tests

Added coverage for UI-only publishing, the GCS artifact contract, and failed UI builds preserving the active publication. Exercised the sandbox publish endpoint locally.

## Risk

Feature-flagged. Publishing now rejects invalid UI source before writing or activating a publication. Existing active publications are unchanged.

## Deploy Plan


smb2268 (commented):
LGTM

PR #31373, opened 2026-08-28
dev: base the shared dev image on Ubuntu 24.04 so Cursor's screen recorder works

Fraggle:
## Description

The shared dev image now uses `ubuntu:24.04` to provide glibc 2.39 and FFmpeg 6 sonames required by Cursor's in-container screen recorder. It installs Node `24.16.0` from the checksum-verified official tarball, switches PGDG to `noble-pgdg`, and documents the base-image constraint; production images remain unchanged.

## Tests

Validated with a real Cursor environment build from this branch, including the full image build and `dev/scripts/install.sh`.

## Risk

Low risk and rollback-safe: the changes are isolated to the shared development image. The main risk is a failed image build or environment refresh, mitigated by checksum verification and reverting the Dockerfile base.

## Deploy Plan

Rebuild and publish the shared dev image, then refresh the Cursor Cloud environment snapshot. No production service deploy or database migration is required.

### DIFF
--- a/dev/Dockerfile
+++ b/dev/Dockerfile
@@ -6,7 +6,13 @@
 # Layer cache: keep this file stable — any ENV/RUN edit above the apt layer invalidates
 # all downstream steps (apt, rust, temporal, …). Shell/terminal defaults live in
 # dev/scripts/env.sh and dev/bashrc (bind-mounted at runtime).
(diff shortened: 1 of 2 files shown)


zmarouf (commented):
LGTM

PR #31375, opened 2026-08-28
[front] fix: remove dead Slack search.messages fallback

davidebbo:
## Problem

`slackSearch` in the Slack personal MCP server tried `assistant.search.context` and, on *any* throw, fell back to Slack's `search.messages`. That fallback could never succeed. `search.messages` requires the `search:read` scope, which our Slack app is not allowed to request — `SlackToolsOAuthProvider` asks only for the granular `search:read.public` / `.private` / `.mpim` / `.im` scopes. So the fallback always failed with `missing_scope`, and `handleSlackAuthError` classified that as an auth failure and returned a "Personal authentication required" prompt.

The dominant trigger was rate limiting, not a broken token. Five days of prod logs (Aug 24–28), clustered by the error that entered the fallback:

| error | count |
| --- | --- |
| `HTTP 429` | 4,723 |
| `missing_scope` | 168 |
| `token_revoked` | 65 |

That is roughly a thousand spurious re-auth prompts a day, shown to users whose Slack connection is perfectly healthy. Re-authenticating grants the identical scope set, so a user can reconnect indefinitely and hit the same prompt on the next rate-limited search. Because `handleSlackAuthError` returns `Ok`, the tool then logged `Tool execution success` and the real Slack error was never recorded — which is why this stayed invisible at that volume.

## Change

Remove the fallback. `slackSearch` is now a straight-line call to `assistant.search.context`, so every failure propagates as itself instead of being swallowed and replaced by a guaranteed `missing_scope`.

Classify rate limiting explicitly: HTTP 429 or a `ratelimited` body error throws `SlackRateLimitError`, and a new `handleSlackSearchError` maps it to a retryable message with `tracked: false`, so external throttling reaches the model without paging us. Genuine `token_revoked` / `missing_scope` from the other Slack tools still route to the auth prompt as before — only the fabricated one is gone.

## Notes

The deleted fallback was the last reference to `search.messages` in the tree; `search:read` now appears nowhere in the repo, which matches the constraint on the Slack app.

Nothing loses search capability. `assistant.search.context` works even for workspaces whose `assistant.search.info` reports `is_ai_search_enabled: false` — verified against a workspace that is permanently `slackAIStatus: disabled` and initializes the server thousands of times a week, yet produces no "AI search not enabled" error in any of the 4,956 fallback entries above. Both `search_messages` and `semantic_search_messages` were already calling this same endpoint; the Slack AI gate selects the tool's shape, not its backend.

No tests added: this server has none, and the helper is module-private, below the endpoint-level functional tests the repo uses.

### DIFF
--- a/front/lib/api/actions/servers/slack_personal/tools/index.ts
+++ b/front/lib/api/actions/servers/slack_personal/tools/index.ts
@@ -39,7 +39,6 @@ import { SLACK_PERSONAL_TOOLS_METADATA } from "@app/lib/api/actions/servers/slac
 import { getRefs } from "@app/lib/api/assistant/citations";
 import type { Authenticator } from "@app/lib/auth";
 import { concurrentExecutor } from "@app/lib/utils/async_utils";
-import logger from "@app/logger/logger";
 import { Err, Ok } from "@app/types/shared/result";
 import { normalizeError } from "@app/types/shared/utils/error_utils";
 import { stripNullBytes } from "@app/types/shared/utils/string_utils";
@@ -51,8 +50,6 @@ import {
 import { INTERNAL_MIME_TYPES } from "@dust-tt/client";
 import assert from "assert";
 import uniqBy from "lodash/uniqBy";
-
-const localLogger = logger.child({ module: "mcp_slack_personal" });
 
 type SlackSearchMatch = {
   author_id?: string;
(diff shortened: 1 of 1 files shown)


frankaloia (approved):
Makes perfect sense. Great find

PR #31376, opened 2026-08-28
Check GCS file existence before signing download URLs

flvndvd:
## Description

In a saga where LLMs drive you crazy…

We had this noisy monitor around LLM calls reporting:
```
Unable to download the file. Please verify the URL and try again.
```

Naturally, the signed URL was blamed. Expired signature? Provider networking? GCS permissions? Clock skew on the cache? 

## What the traces revealed

After inspecting the full LLM payload, we found a GCS signed URL embedded in a historical image tool result.

The URL had a one-hour lifetime and the provider request happened seconds after it was created. So it was not expired.

The actual sequence was:

1. A file tool returned an image backed by a conversation file.
2. Conversation rendering converted its file path into a signed GCS URL.
3. The agent inspected the image.
4. The temporary image was subsequently deleted (thru sandbox).
5. The historical tool result remained in the conversation.
6. On the next LLM request, Dust generated another perfectly valid signed URL (for an object that no longer existed).
7. The provider fetched it, received a 404, and translated that into a mysterious “unable to download” error.

GCS will happily sign a path without checking whether anything lives there. Cryptographically valid, existentially challenged.

## The fix

`GCSFileSystemBackend.getDownloadUrl()` now checks that the object exists before generating a signed URL.

When the object is missing, it returns `not_found`. The existing conversation renderer then replaces the image with an “Image unavailable” text block instead of sending a broken URL to the provider. 

This protects every provider consuming these rendered URLs, rather than special-casing one LLM or converting images to base64.

## Scope

This is a tactical guardrail. The longer-term solution is to persist tool-result images using stable, immutable references. This will burst the cache, we need a cache-friendly approach for longer term.

As we move towards a stateful context window (https://github.com/dust-tt/dust/pull/31128), already-materialized stateful contexts may still contain an old URL until their checkpoint is refreshed.

### DIFF
--- a/front-api/routes/v1/w/[wId]/spaces/[spaceId]/project_files/index.test.ts
+++ b/front-api/routes/v1/w/[wId]/spaces/[spaceId]/project_files/index.test.ts
@@ -8,6 +8,7 @@ import { PROJECT_FILES_MAX_CONCURRENT_REQUESTS_PER_PROCESS } from ".";
 
 const getAllFilesByPrefixMock = vi.hoisted(() => vi.fn());
 const getSignedUrlMock = vi.hoisted(() => vi.fn());
+const fileExistsMock = vi.hoisted(() => vi.fn());
 
 vi.mock("@app/lib/file_storage/config", () => ({
   default: { getGcsPrivateUploadsBucket: vi.fn(() => "test-bucket") },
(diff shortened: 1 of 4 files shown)


fabiencelier (approved):

PopDaph (approved):
🙏🏻

PR #31378, opened 2026-08-28
Turn the model picker's "More models" into hover submenus

pinotalexandre:
## Summary
- Clicking "More models" now expands the maker list inline, in place.
- Each provider is a submenu that opens on hover (the dropdown's standard, native submenu behavior), showing its model list — no custom navigation state or transition.


front/components/model_picker/ModelPickerMakersView.tsx
```diff
@@ -0,0 +1,92 @@
...
+
+  return (
+    <>
+      {makerGroups.map((maker) => (
+        <DropdownMenuSub key={maker.makerId}>
+          <DropdownMenuSubTrigger
+            label={getModelMakerDisplayName(maker.makerId)}
+            icon={getModelMakerLogo(maker.makerId, isDark)}
+          />
+          <DropdownMenuSubContent className="w-64 max-w-(--radix-dropdown-menu-content-available-width)">
```
aubin-tchoi: do you need this max-w? should be handled by the component directly unless you're doing something very specific here

front/components/model_picker/ModelPickerContent.tsx
```diff
@@ -139,24 +145,31 @@ export function ModelPickerContent({
...
+            visual={isMakersExpanded ? ChevronDown : ChevronRight}
+            size="xs"
+            className="text-muted-foreground"
+          />
+        }
+        onClick={onToggleMakers}
+        onSelect={(e) => e.preventDefault()}
       />
+
+      {isMakersExpanded && (
```
aubin-tchoi, line 161: no animation at all?

aubin-tchoi (approved):
LGTM

front/components/model_picker/ModelPickerContent.tsx
pinotalexandre: Yep, I wanted to keep it simple

front/components/model_picker/ModelPickerMakersView.tsx
pinotalexandre: Good catch — removed it. Prod's original submenu was just `w-64`, no max-w; a submenu flips side on collision rather than shrinking width, so it wasn't doing anything.

PR #31383, opened 2026-08-28
[sparkle] fix: prevent SliderSteps tooltip jitter

aubin-tchoi:
## Description

SliderSteps tooltips were anchored to a zero-size element that moved between hovered steps, which made the reasoning-effort tooltip jump near step boundaries.

This PR uses the slider root as the tooltip trigger so its geometry stays stable while the per-step copy still updates.

## Tests

- Updated the story.
- Tested locally + preview.

## Risk

- Low.

## Deploy Plan

- Deploy front.

### DIFF
--- a/sparkle/src/components/SliderSteps.tsx
+++ b/sparkle/src/components/SliderSteps.tsx
@@ -122,7 +122,7 @@ export function SliderSteps({
       ? (stepTooltips?.[hoveredIndex] ?? null)
       : null;
 
-  const root = (
+  const slider = (
     <SliderPrimitive.Root
       className={cn(
         "relative flex h-5 w-full touch-none select-none items-center",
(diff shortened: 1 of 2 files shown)


radjakahoul:
All good

What can still be slightly frustrating is to have a slider that actually does not exactly follow my mouse as I move.
I think it is okay that it goes all the way to the next closest level, but not 100% smooth that it goes faster than the mouse move. Clearly minor detail though

radjakahoul (approved):
good, maybe further small improvements later - see comment :)

aubin-tchoi:
yes this slider is really not great in terms of interaction. I want to borrow a few things from https://www.fluidfunctionalism.com/docs/slider

aubin-tchoi:
but mostly want to fix the real bugs first (the tooltip blinking was hard to miss)

PR #31385, opened 2026-08-28
PostHog session persistence + pageviews fix

theogz:
## Description

PostHog stored its session id in memory-only persistence, so every page load started a new session and almost every sessions ended up with a single pageview. This picks the persistence store from cookie consent at init instead, and stops overwriting PostHog's own distinct_id at init, which was resetting identified users to their anonymous id on every load.

## Tests

Will try to test locally after a first review

## Risk

Low

## Deploy Plan

Very safe to rollback


frankaloia (approved):

tdraier:
LGTM ✅

PR #31387, opened 2026-08-28
[front] feat: add action_id to tool.executed audit event

zmarouf:
## Description

Adds `action_id` (the MCP action sId) to the `tool.executed` audit event so it can be joined to `tool.approval_requested` / `tool.approval_resolved`, which already emit that field.

- Schema: `front/admin/audit_log_schemas/tool.executed.json`
- Emit sites: agent-loop `run_tool.ts` and pod-function `run_sandbox_function_tool.ts`
- Version map: `tool.executed` v8 in `schema_versions.json` (bumped from main's v7 during rebase)

Approval fields stay on `tool.approval_resolved`; this is only the join key.

## Tests

- `npm -w front run test -- lib/api/audit/audit_log_schemas.test.ts`

## Risk

Low. Additive optional metadata on an existing event. Events emitted before deploy omit the field; WorkOS treats declared metadata as optional.

## Deploy Plan

Register the updated schema with WorkOS (expect v8), then deploy `front` workers that emit `tool.executed` so they send version 8. No DB migration. Rollback: revert the emit/schema PR; if workers already send v8, keep the version map or re-register as needed.

### DIFF
--- a/front/admin/audit_log_schemas/tool.executed.json
+++ b/front/admin/audit_log_schemas/tool.executed.json
@@ -14,6 +14,7 @@
     "mcp_server_name": "string",
     "conversation_id": "string",
     "agent_message_id": "string",
+    "action_id": "string",
     "trigger_id": "string",
     "pod_id": "string",
     "pod_function_id": "string",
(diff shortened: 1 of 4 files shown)


davidebbo (approved):
LGTM

PR #31382, opened 2026-08-28
Declare Frames v2 SQLite databases

fontanierh:
## Description

Adds the foundational Frames v2 SQLite contract. Frame manifests can declare uniquely named databases and their schema source paths. Publication rejects missing schema files and preserves declared schemas in the immutable source snapshot. Stable database replica prefixes are keyed by workspace, Frame identity, and database name outside publication storage.

This slice does not yet mount, reconcile, or replicate live SQLite databases; those runtime lifecycle changes follow separately.

Stacked on #31380.

## Tests

- 43 focused manifest, GCS path, and publication storage tests
- `npm run tsgo` in `front`

## Risk

Existing manifests default to an empty database list. No runtime or existing Pod database behavior changes. Safe to roll back.

## Deploy Plan

Normal deployment after #31380.

PR #31388, opened 2026-08-28
Activate published Frame functions atomically

fontanierh:
## Description

Materialize one immutable sandbox function row per Frames v2 publication function. Activation updates the function set, authorized-file snapshot, and active publication pointer in one transaction under the Frame publish lock.

Frame deletion and workspace scrub now remove Frame-owned function invocations, actions, and rows before deleting the owning file.

Stacked on #31382.

## Tests

Added coverage for function materialization, transactional rollback, direct Frame deletion, and workspace scrub. Ran the focused publication, sharing allowlist, FileResource, Sandbox Function, invocation, action, and build/publish suites locally.

## Risk

Activation depends on valid stored function contracts. A failed activation leaves the prior publication and allowlist active. Frame-owned cleanup is batched for workspace scrub.

## Deploy Plan

Merge after #31382, then standard deploy.

### DIFF
--- a/front/lib/resources/file_resource.ts
+++ b/front/lib/resources/file_resource.ts
@@ -37,6 +37,7 @@ import { isGCSNotFoundError } from "@app/lib/file_storage/types";
 import { BaseResource } from "@app/lib/resources/base_resource";
 import { ConversationResource } from "@app/lib/resources/conversation_resource";
 import { DataSourceResource } from "@app/lib/resources/data_source_resource";
+import { SandboxFunctionInvocationResource } from "@app/lib/resources/sandbox_function_invocation_resource";
 import { SpaceResource } from "@app/lib/resources/space_resource";
 import {
   AuthorizedFileAccessModel,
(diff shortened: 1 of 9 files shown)


front/lib/resources/file_resource.ts
```diff
@@ -500,6 +506,71 @@ export class FileResource extends BaseResource<FileModel> {
...
+  ): Promise<number> {
+    if (sandboxFunctionModelIds.length === 0) {
+      return 0;
+    }
+
+    await SandboxFunctionInvocationResource.deleteAllForSandboxFunctionModelIds(
+      { workspaceModelId, sandboxFunctionModelIds }
+    );
+
+    return SandboxFunctionModel.destroy({
```
davidebbo, line 521: nit: Maybe should be in SandboxFunctionResource

front/lib/resources/file_resource.ts
```diff
@@ -500,6 +506,71 @@ export class FileResource extends BaseResource<FileModel> {
...
+      where: {
+        id: sandboxFunctionModelIds,
+        workspaceId: workspaceModelId,
+      },
+    });
+  }
+
+  private static async deleteAllFrameFunctionsForWorkspace(
```
davidebbo, line 529: Same here, this whole function feels like it belongs in SandboxFunctionResource. It's not related to Files.

davidebbo (commented):
LGTM with some comments

PR #31389, opened 2026-08-28
Add: cell metadata, lookup and backfill script

Fraggle:
## Description

WorkOS organization metadata now records both `cell` and `region`, with `lookupWorkspaceCell` resolving local workspaces first then using a WorkOS fallback for remote workspaces and finally looking up other cells manually as last resort. Workspace relocation flows update and invalidate cell-aware lookup metadata, while a guarded backfill script populates existing organizations without overwriting relocation stubs or metadata owned by another deployment.

## Tests

No automated tests were added or updated. The changes were reviewed against the relocation and WorkOS organization flows; the backfill script supports a dry-run before execution.

## Risk

Medium risk: relocation and backfill paths mutate global WorkOS organization metadata, and incorrect `cell` values can misroute workspace lookups. The backfill skips relocation stubs and organizations pointing to another deployment, supports dry-run execution, and relocation invalidates both region and cell caches. Rollback is safe by reverting the deploy and restoring metadata with `update_workspace_workos_metadata`.

## Deploy Plan

Deploy `front` to all cells. Run the new WorkOS organization metadata backfill once per cell, starting with a dry run and then executing it. No SQL migration or infrastructure change is required.


Fraggle:

- #31389 ◀

front/migrations/20260828_backfill_workos_organization_cell_metadata.ts
```diff
@@ -0,0 +1,215 @@
...
+  }
+
+  try {
+    await getWorkOS().organizations.updateOrganization({
+      organization: organization.id,
+      metadata: {
+        region: targetRegion,
+        cell: targetCell,
+      },
+    });
```
tdraier, line 205: we're limited to 6000 req/min/key , maybe this would need some throttling and/or run on a dedicated API key

tdraier (approved):

front/migrations/20260828_backfill_workos_organization_cell_metadata.ts
Fraggle: Good idea, i'll use a dedicated API KEY to manually run the script (by shadowing the env var).

front/lib/api/cells/lookup.ts
```diff
@@ -0,0 +1,152 @@
...
+}
+
+async function lookupWorkspaceInCell(
+  wId: string,
+  cell: CellInfo
+): Promise<boolean> {
+  const body = { workspace: wId };
+
+  // eslint-disable-next-line no-restricted-globals
+  const response = await fetch(
```
flvndvd, line 61: So fallback if WorkOS is unreachable, we do N calls to get the cell?

front/lib/api/cells/lookup.ts
```diff
@@ -0,0 +1,152 @@
...
+    }
+
+    throw error;
+  }
+}
+
+async function lookupWorkspaceInCell(
+  wId: string,
+  cell: CellInfo
+): Promise<boolean> {
```
flvndvd, line 57: Why not using `Result` here?

front/lib/api/cells/lookup.ts
```diff
@@ -0,0 +1,152 @@
...
+    if (result.status === "fulfilled" && result.value) {
+      return result.value;
+    }
+  }
+
+  const errors = results.filter(
+    (result): result is PromiseRejectedResult => result.status === "rejected"
+  );
+  if (errors.length === results.length) {
+    throw normalizeError(errors[0].reason);
```
flvndvd, line 106: Same here, why throwing?

flvndvd (commented):
LGTM

front/lib/api/cells/lookup.ts
Fraggle: Copied the code from region's version to replace easily.
I agree that Result is better, will refactor later :)

front/lib/api/cells/lookup.ts
Fraggle: AND if the cache is cold.

I discovered that It's also possible that the user has no organization (free plan for example).

front/lib/api/cells/lookup.ts
Fraggle: Same, same code as the region one.

PR #31391, opened 2026-08-28
Add Frame-owned sandbox lifecycle

fontanierh:
## Description

- Add a Frame sandbox adapter keyed by the stable Frames v2 FileResource identity.
- Re-resolve the Frame source scope under the lifecycle lock and inherit Pod-scoped runtime configuration when applicable.
- Add FRAME_ID runtime metadata while preserving existing conversation and Pod behavior.
- Delete Frame sandboxes before Frame or workspace deletion, with batched workspace cleanup, and include them in the sandbox reaper.
- Keep the slice internal: publication mounts and invocation routes follow in stacked PRs.

Stacked on #31388.

## Tests

- Added Frame sandbox identity, Pod configuration, direct deletion, batched workspace scrub, environment manifest, and reaper coverage.
- Ran 123 focused tests across Frame, FileResource, sandbox lifecycle, egress, environment manifest, and reaper suites.
- Ran npm run tsgo.
- Ran Biome on all changed files and git diff --check.

## Risk

The adapter is not reachable from a user invocation path yet. Existing Frames v1 and Pod Functions paths are unchanged. Runtime Pod lookup is workspace-scoped and intentionally independent of caller visibility, so a shared invocation cannot silently lose inherited configuration. The shared lifecycle adjustment only allows an absent owner link to resolve under lock before requiring a configured provider; deleting an existing sandbox still fails closed when no provider is configured.

## Deploy Plan

Merge #31388 first, then merge this PR. No migration or configuration change is required.

PR #31392, opened 2026-08-28
Mount canonical Frame publication artifacts

fontanierh:
## Description

- Store immutable Frame artifacts under the canonical publication keys and keep authoring source out of publication storage.
- Mount the stable Frame publication root read-only in its Frame-owned sandbox, with no agent-visible source mounts.
- Stack on #31391.

Follow-up before SQLite reconciliation: replace the raw manifest payload in `publication.json` with the versioned publication descriptor, including immutable database schema sources and hashes.

## Tests

Added coverage for publication keys, source exclusion, Frame filesystem scoping, mount configuration, and lifecycle setup. Tested the focused 142-test suite locally.

## Risk

Low. The path changes are behind Frames v2 and are not user-reachable until invocation wiring lands. Existing draft publications need republishing. Frames v1 and Pod Functions are unchanged.

## Deploy Plan

Merge #31391 first, then standard deploy. No migration.


smb2268 (commented):
LGTM 🚢

PR #31393, opened 2026-08-28
[front] fix: stop blocked-action prompts from getting permanently stuck

davidebbo:
## Problem

`POST .../messages/:mId/retry?blocked_only=true` returns **500 whenever there is nothing left to resume**, and leaves the stale blocked-action events in Redis. The result is a prompt the user can never clear.

`findUserMessageForRetry` returns `DustError("no_blocked_actions")` when the blocked actions were already consumed, and `DustError("agent_message_not_resumable")` when the agent message reached a terminal state. Neither is a failure — the code says so itself (*"Not a failure: the message simply has nothing waiting on user input"*), and `resumeAncestorConversations` already lists both in `NON_BLOCKING_RETRY_ERROR_CODES`. But the handler special-cased only `agent_loop_already_running` and mapped everything else to a 500.

The purge of blocked-action events from the message stream sat *after* those two early returns, so it never ran for precisely the messages that needed it. Clients replay the whole stream from `"0-0"` on load, so:

1. User connects Slack / clicks retry → 500, no purge, no agent loop.
2. `useRetryMessage` swallows non-limit errors and returns `Ok`, so no error is surfaced — the button just does nothing.
3. The banner is only removed by an incoming stream event (`AgentMessage.tsx:428-443`); none arrives.
4. On reload the stale events replay and the prompt is back.

Forever. This is what 1Password reported as "3 (or more) manual actions are needed, only shows one" — the queue holds several, `getFirstBlockedActionForMessage` renders the head.

**Scale:** 103,609 of these 500s in 7 days across prod (~15k/day). 3,037 of them are the 1Password workspace, whose Slack integration is otherwise healthy (26,123 tool executions in the same window). So this is not a Slack problem and not a 1Password problem.

## Change

- **Purge blocked-action events before the early returns** in `retry_blocked_actions.ts`. A message with nothing left to resume is exactly the case where stale events keep re-prompting.
- **Treat both codes as success** in the handler and return the message. The prompt is gone, which is what the caller asked for. `agent_loop_already_running` still maps to 400.
- **Log the underlying error before any 500.** The handler discarded `retryBlockedActionsRes.error` entirely, which is why a five-figure daily error count went unnoticed and why I can't say from logs which of the two codes dominates.
- **One definition of "nothing to resume"**, shared with `resumeAncestorConversations` rather than each caller deciding for itself.

## Test

`retry.test.ts` covers the reported case: a blocked-only retry on a message with no blocked actions. Verified it fails on the old code with `expected 500 to be 200`, and passes here.

- `useRetryMessage` swallowing non-limit errors is a landmine beyond this endpoint — it should return `Err` and notify.
- Clear the banner on a successful blocked-only retry, so the fix doesn't need a reload.
- A monitor on retry 5xx.
- Longer term, blocked-action prompts should derive from server state rather than a replayed event stream whose purge is on the success path only.

### DIFF
--- a/front-api/routes/w/[wId]/assistant/conversations/[cId]/messages/[mId]/retry.test.ts
+++ b/front-api/routes/w/[wId]/assistant/conversations/[cId]/messages/[mId]/retry.test.ts
@@ -1,0 +1,64 @@
+import { ConversationResource } from "@app/lib/resources/conversation_resource";
+import { AgentConfigurationFactory } from "@app/tests/utils/AgentConfigurationFactory";
+import { ConversationFactory } from "@app/tests/utils/ConversationFactory";
+import { createPrivateApiMockRequest } from "@app/tests/utils/generic_private_api_tests";
+import { honoApp } from "@front-api/app";
+import { describe, expect, it } from "vitest";
+
+async function setupAgentMessage() {
+  const { auth, workspace } = await createPrivateApiMockRequest({
+    role: "user",
+    method: "POST",
+  });
+
+  const agentConfiguration = await AgentConfigurationFactory.createTestAgent(
+    auth,
+    { name: "Retry endpoint" }
+  );
+
+  const createdConversation = await ConversationFactory.create(auth, {
+    agentConfigurationId: agentConfiguration.sId,
+    messagesCreatedAt: [],
+  });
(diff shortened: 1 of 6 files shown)


davidebbo:
This is not the right fix. New one coming.

PR #31394, opened 2026-08-28
Version Frame publication descriptors

fontanierh:
## Description

- Replace the raw manifest commit marker with a versioned `publication.json` descriptor.
- Record publisher/time, normalized manifest, source hashes, UI/function hashes, build-derived function contracts, and exact database schema source.
- Verify database, UI, and function hashes before activation. Remove function schema sidecars now covered by the descriptor.
- Stack on #31392.

## Tests

Added descriptor validation, manifest-to-source integrity checks, immutable database-schema capture, corruption rejection, artifact ordering, and activation regression coverage. Tested the focused 45-test suite locally.

## Risk

Low. Frames v2 remains feature-flagged and invocation is not user-reachable yet. Draft publications from earlier slices must be republished. Frames v1 and Pod Functions are unchanged.

## Deploy Plan

Merge #31392 first, then standard deploy. No migration.


smb2268 (commented):
LGTM!

PR #31395, opened 2026-08-28
Run Frame functions from exact publications

fontanierh:
## Description

- Execute Frame functions in the Frame-owned sandbox from the immutable publication selected by the function row.
- Store invocation payloads under the Frame GCS prefix and bind invocation tokens to the Frame and its lifecycle-locked runtime space.
- Require workspace membership before a Frame sandbox can be woken, then enforce scope-dependent identity policies against the locked location.
- Keep immutable Frame function declarations unchanged when a fast function attempts a tool call; the error requires republishing as durable.
- Keep Pod Function mounts, state, tokens, authorization, and fast-function self-healing unchanged.
- Stack on #31394.

## Tests

Added Frame token, runtime scope, move-race authorization, membership, exact mount, GCS path, immutable execution-mode, and Pod regression coverage. Ran 106 focused tests, Front and Front API typechecks, Biome, audit-schema lint, and `git diff --check`.

## Risk

Low. Frames v2 remains feature-flagged and no caller-facing Frame invocation route is added in this slice. Frame functions intentionally receive no Pod database environment; Frame-owned SQLite follows separately. Frames v1 and Pod Functions retain their existing execution path.

## Deploy Plan

Merge #31394 first. After merge and before deploy, run `npx tsx front/admin/register_audit_log_schemas.ts --execute --changed`, then deploy normally. No database migration.


front/lib/api/sandbox_functions/create_sandbox_function_mcp_action.ts
```diff
@@ -186,7 +186,7 @@ export async function createSandboxFunctionMCPAction(
     return new Err(
       new SandboxFunctionMCPActionError(
         "invocation_not_found",
-        "Pod function not found."
+        "Sandbox function not found."
```
davidebbo, line 189: If we called them 'pod functions' before, shouldn't they logically be called 'frame functions' now, and not 'sandbox functions'? 🤔

front/lib/api/sandbox_functions/self_heal_execution_mode.ts
```diff
@@ -39,6 +39,13 @@ export async function selfHealSandboxFunctionExecutionMode(
     );
     return;
   }
+  if (sandboxFunction.frame) {
+    logger.error(
+      { sandboxFunctionId, invocationId, frameId: sandboxFunction.frame.sId },
+      "Refusing to mutate an immutable Frame publication function"
```
davidebbo, line 45: So we lose self healing?

fontanierh: yes, I prefer that over having a mismatch with the manifest

front/lib/api/sandbox_functions/create_sandbox_function_mcp_action.ts
fontanierh: no because currently they are both pod functions and frame functions. Once we retire pod functions they can indeed become "frame functions"

davidebbo (commented):
LGTM

PR #31397, opened 2026-08-28
Persist Frame-owned SQLite runtime state

fontanierh:
## Description

- mount a private writable Litestream replica at `w/{workspaceId}/frames/{frameId}/state/databases`
- restore Frame SQLite state before a fresh sandbox can execute functions
- flush Frame databases before sandbox sleep or kill, including credential refresh
- expose the existing DSBX database runtime ABI to Frame functions without mounting source files
- delete all canonical publication and state objects on Frame deletion or workspace scrub
- stack on #31396

## Tests

- 143 focused Front tests covering Frame paths, mounts, cold start, reaper sleep, invocation, individual deletion, and workspace scrub
- Front `tsgo`
- Biome on all touched files
- `git diff --check`

## Risk

Low and feature-gated by `frames_v2`. The Pod path keeps the same mount prefix, runtime env, and lifecycle behavior. Rollback removes the Frame state mount and runtime wiring; explicit deletion remains safe.

## Deploy Plan

Merge #31396 first, then deploy normally. No migration or manual production step.

### DIFF
--- a/front/lib/api/sandbox/db.ts
+++ b/front/lib/api/sandbox/db.ts
@@ -26,13 +26,13 @@ import { normalizeError } from "@app/types/shared/utils/error_utils";
 /**
  * Pod state runtime plumbing: SQLite databases under /pod-state/databases,
  * continuously replicated by a litestream daemon (running as dust-state) to a
- * gcsfuse-mounted GCS prefix at /pod-state/replica.
+ * gcsfuse-mounted GCS prefix at /sandbox-state/replica.
  *
  * The daemon runs litestream's directory watcher over /pod-state/databases
  * (static /etc/litestream.yml baked at image build): databases created at any
  * point — including publish-time `dsbx db reconcile` — are discovered and
  * replicated automatically within seconds. Replica subdirectories are named
- * by database FILENAME: /pod-state/replica/{db}.db/ltx/...
+ * by database FILENAME: /sandbox-state/replica/{db}.db/ltx/...
  *
  * Lifecycle:
  *  - Cold start (`setupPodStateOnColdStart`, after the gcsfuse mounts): restore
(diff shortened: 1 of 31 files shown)


id13 (commented):
LGTM 

@fontanierh should we keep the pod-function code alive knowing we are not shipping it ? 

At least for the db part, I don't see any use for it.

front/lib/api/sandbox/image/pod_package.ts
```diff
@@ -2,14 +2,14 @@ import { runCachedBunBuild } from "@app/lib/api/sandbox/image/bun_build";
 import fs from "fs";
 import path from "path";
 
-// @dust/pod is the pod-state runtime package exposed to sandbox function
-// code: db(name) returns a handle on one of the pod's SQLite databases
+// @dust/pod is the sandbox-state runtime package exposed to sandbox function
```
id13, line 5: maybe rename @dust/pod to @dust/sandbox or @dust/frame ?

front/lib/api/sandbox/db.ts
```diff
@@ -597,17 +607,59 @@ export async function ensurePodStateHealthOnSleep(
...
+  if (rootMissingResult.isErr()) {
+    return rootMissingResult;
+  }
+  if (rootMissingResult.value.exitCode !== 0) {
+    return validateReplicaMountStat(result);
+  }
+
+  return validateReplicaMountStat(
+    await statReplicaMount(auth, sandbox, LEGACY_POD_STATE_REPLICA_MOUNT_POINT)
+  );
```
id13, line 645: do we really care about all this machinery ? i.e is it worth to be backward compatible with pod-functions ? 
Long term no, short term debatable but I would just go scorched earth on this one.

front/lib/api/sandbox/image/pod_package.ts
fontanierh: Yes but won't do now

fontanierh:
@id13
> should we keep the pod-function code alive knowing we are not shipping it ?

i'm going to remove it eventually, but not before Frames functions are implemented end to end

front/lib/api/sandbox/db.ts
fontanierh: not sure I understand what you mean ?

PR #31400, opened 2026-08-28
fix(sandbox): Pod network scope-badge color and domain removal

smb2268:
## Description

Follow-ups from testing the merged Pod network editor (`30996`). Three fixes:

1. **Scope-badge color.** The Workspace/Pod badges used Chip `color="info"`, which in the sparkle palette aliases to `--color-golden-*` — so they rendered gold, not blue. Switched to `color="highlight"` (`--color-blue-*`), which is also the only option with real contrast against the rows (`bg-muted-background` = `stone-50`; a `primary` chip is also `stone-50`).

2. **Removal targeted the wrong scopes.** Pressing the trash on a Pod-owned domain removed it from *every selected scope* (e.g. Workspace **and** the Pod), not just the scopes that own the row. `handleRemoveDomain` now takes the `DomainRow` and targets only `includeWorkspace && row.inWorkspace` plus `row.ownedByPods` — matching what `removableScopeCount` already computes.

3. **Wildcards couldn't be removed.** `removeOwnerPolicyDomain` / `removeWorkspacePolicyDomain` ran the input through `parseExactEgressDomain`, which rejects wildcards — but the admin whole-list write can store wildcards (`*.example.com`), so a stored wildcard was impossible to remove. Removal now matches the stored entry as-is (the domain comes straight from the allowlist, so there's nothing to normalize); an absent domain stays a no-op.

## Tests

Added: backend unit test that a stored wildcard is removed from an owner policy; frontend render test that removing a Pod-owned row targets only that Pod (`includeWorkspace: false`, `pods: [pod]`). Existing `buildDomainRows` coverage already exercises the `ownedByPods` / `removableScopeCount` data the removal now relies on.

## Risk

Low. Cosmetic color change; removal now correctly scoped and no longer rejects wildcards on delete (the exact-domain add path is unchanged, so wildcards still can't be *added* via the bulk editor). Safe to roll back.

## Deploy Plan

Standard front deploy. No migration.

### DIFF
--- a/front/components/sandbox/MultiPodNetworkSection.test.tsx
+++ b/front/components/sandbox/MultiPodNetworkSection.test.tsx
@@ -10,16 +10,26 @@ import { render, screen, waitFor } from "@testing-library/react";
 import userEvent from "@testing-library/user-event";
 import { beforeEach, describe, expect, it, vi } from "vitest";
 
-const { bulkUpdateMock, mutatePodPoliciesMock, mutateWorkspaceMock } =
-  vi.hoisted(() => ({
+const {
+  bulkUpdateMock,
+  mutatePodPoliciesMock,
+  mutateWorkspaceMock,
+  podPoliciesState,
+} = vi.hoisted(() => {
+  const podPoliciesState: {
+    current: { podId: string; policy: EgressPolicy }[];
+  } = { current: [] };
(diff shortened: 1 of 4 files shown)


frankaloia:
LGTM

PR #31401, opened 2026-08-28
Route Frames v2 calls through Frame identity

fontanierh:
## Description

Route Frames v2 `callFunction` messages through the stable Frame identity and active publication. Keep the existing Pod-relative resolver for legacy Frames and Pod Functions, and fail closed when a Pod-hosted Frame cannot be classified from trusted metadata.

Reuse the existing path HEAD response for the linked FileResource id and MIME type, avoiding a second metadata request for Pod tabs and pinned banners. The raw mounted object keeps its standard `Content-Type`; the linked FileResource MIME is returned through an explicit CORS-exposed Dust header. Metadata loading stays disabled while the Pod Frame sheet is closed.

Based directly on `main`.

No visual changes.

## Tests

- focused Front Vitest: 9 files / 36 tests passed
- focused Front API Vitest: 2 files / 29 tests passed
- `npx tsgo --noEmit` in `front-api`
- `npm run format:changed`
- `git diff --check`
- Storybook MCP was unavailable in this session; no Sparkle properties changed

## Risk

Low and feature-gated. Frames v2 receive an identity-qualified reference; legacy Frames, raw visualizations, and Pod Functions retain their existing reference resolution. Unknown Pod Frame metadata fails closed. The new response header is additive and raw `Content-Type` behavior is preserved. Rollback is reverting this PR.

## Deploy Plan

Deploy normally. No migration or backfill.

### DIFF
--- a/front-api/middlewares/cors.test.ts
+++ b/front-api/middlewares/cors.test.ts
@@ -1,5 +1,8 @@
 import { FRAME_SHARE_TOKEN_HEADER } from "@app/types/api/sandbox_functions";
-import { DUST_FILE_ID_HEADER } from "@app/types/files";
+import {
+  DUST_FILE_CONTENT_TYPE_HEADER,
+  DUST_FILE_ID_HEADER,
+} from "@app/types/files";
 import { cors } from "@front-api/middlewares/cors";
 import { Hono } from "hono";
 import { describe, expect, it } from "vitest";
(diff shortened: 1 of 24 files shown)


fontanierh:
Superseded by the focused renderer stack: #31518, #31519, #31520, #31521, and #31522. Each behavior slice has focused tests, and the synthetic final tree matches this PR across all original files.

PR #31402, opened 2026-08-28
Delete Frames v2 packages

fontanierh:
## Description

- Preserve the generic private file DELETE contract: Frames v2 delete their source package, runtime data, sharing records, and identity, then return 204.
- For Pod Frames, delete the source first, then clean project metadata and content fragments before deleting runtime state and identity. Missing source folders are treated as an idempotent retry.
- Reject direct or legacy project deletion before any mutation, and keep the Poke action legacy-only.
- Reject deleting a package that contains another registered Frame, preventing orphaned nested identities.
- Emit `frame.deleted` after successful package deletion.

Scope: 15 files, +671/-69. No dsbx CLI or sandbox endpoint changes.

## Tests

- Generic DELETE coverage for the 204 response and full source/runtime/identity cleanup.
- Pod source-failure and retry coverage for orphaned and referenced fragments.
- Direct-delete, legacy project-delete, nested-package, and Poke guards.
- Existing publication and sandbox deletion regressions.
- Locally: 76 focused Front tests and 90 file-route Front API tests; both typechecks, Biome, Swagger checks, and commit hooks pass.

## Risk

Deletion is irreversible. The flow holds the existing bounded Frame publication lock across source and owned-resource cleanup, and retains the Frame identity when a required cleanup step fails. Large-installation batching and renewable lease checks are intentionally deferred to a separate hardening change.

## Deploy Plan

Register the WorkOS `frame.deleted` schema before deploy. Standard application deploy.

### DIFF
--- a/front-api/public/swagger.json
+++ b/front-api/public/swagger.json
@@ -9199,7 +9199,7 @@
       },
       "delete": {
         "summary": "Delete a file",
-        "description": "Delete a file from the workspace. Files referenced by a skill or its version history cannot be deleted.",
+        "description": "Delete a file from the workspace. Frames v2 are deleted with their source package and owned runtime data. Files referenced by a skill or its version history cannot be deleted.",
         "tags": [
           "Private Files"
         ],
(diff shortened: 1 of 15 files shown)


fontanierh:
Replaced by smaller reviewable slices:

- #31502 package-aware FileResource deletion
- #31503 project cleanup ordering
- #31504 package deletion HTTP/service
- #31507 audit emission
- #31505 independent legacy Poke guard

PR #31403, opened 2026-08-28
Configure Frame use rights through dsbx

fontanierh:
## Description

- Add `dsbx frame share` for workspace, invite-only and public Frame use rights.
- Resolve the stable Frame from its writable source package and enforce existing workspace sharing permissions.
- Commit scope, authorized-file access and new email grants in one transaction. Audits and notifications run only after commit.
- Reuse the current sharing records and permission checks. Legacy Frame sharing is unchanged.

## Tests

Added sandbox-token coverage for a published v2 Frame, idempotent grants and validation rollback. Re-ran all existing share-scope and email-grant route tests.

## Risk

Authorized-file validation completes before mutation. Scope, allowlist and grants then commit atomically, so a failed share leaves prior use rights unchanged. The existing 28 web-route tests pass unchanged. The new endpoint is internal, feature-flagged and restricted to registered v2 packages in writable mounts.

## Deploy Plan

### DIFF
--- a/cli/dust-sandbox/src/api/client.rs
+++ b/cli/dust-sandbox/src/api/client.rs
@@ -9,8 +9,9 @@ use tokio::time::sleep;
 use super::error::DustApiError;
 use super::types::{
     parse_action_poll_response, ActionPollResponse, CallToolPostResponse, CallToolRequest,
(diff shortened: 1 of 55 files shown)


fontanierh:
Superseded by the recut sharing lane #31506 -> #31508 -> #31497 -> #31498, with #31493 as a hardening tip.

PR #31404, opened 2026-08-28
Reuse safe Frame source copying

fontanierh:
## Description

- Extract the current bounded, generation-pinned Frame source copier and exact-generation cleanup for clone reuse.
- Keep Frame move behavior unchanged while preserving manifest completeness, create-only destination ownership, concurrent-source-edit safety, and recovery bookkeeping.

## Tests

- 14 focused Frame move lifecycle tests.
- Front typecheck.
- Biome checks and git diff check for all changed files.

## Risk

Refactor only. The move path delegates to the extracted helper without changing its copy, recovery, or cleanup behavior.

## Deploy Plan


fontanierh:
Superseded by the recut clone foundations #31513 and #31515.

PR #31405, opened 2026-08-28
Clone Frames into fresh runtime ownership

fontanierh:
## Description

- Add the feature-flagged internal Frame clone lifecycle used by DSBX.
- Copy source without overwriting, register a new Frame identity, and publish a new immutable publication across authorized conversation or Pod mounts.
- Start from a distinct sharing record with no inherited sandbox or SQLite state.
- Remove copied objects and the new identity if the first publication fails.
- Emit a registered `frame.cloned` audit event after the complete lifecycle succeeds.

## Tests

- Twelve sandbox Frame route tests cover same-conversation and cross-conversation clone, fresh identity/publication/sharing/runtime ownership, audit emission, publication rollback, and legacy Frame publish behavior.
- Two direct registration tests.
- Nine audit schema consistency tests.
- Front and Front API typechecks.

## Risk

Clone spans GCS and Postgres. Destination writes use conditional creation and the exact generation returned by GCS for cleanup. A failed cleanup preserves the registered partial clone so it remains explicitly deletable instead of risking newer destination data.

## Deploy Plan

### DIFF
--- a/front-api/routes/v1/w/[wId]/sandbox/frames/clone.ts
+++ b/front-api/routes/v1/w/[wId]/sandbox/frames/clone.ts
@@ -1,0 +1,123 @@
+import {
+  type CloneFrameV2SourceError,
+  type CloneFrameV2SourceResult,
+  cloneFrameV2Source,
+  isFrameSourceCloneError,
+} from "@app/lib/api/frames/clone_source";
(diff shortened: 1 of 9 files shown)


fontanierh:
Superseded by the recut clone lane #31513 -> #31515 -> #31496 -> #31500.

PR #31406, opened 2026-08-28
Expose Frame clone through dsbx

fontanierh:
## Description

- Add `dsbx frame clone <source> <destination>` backed by the internal Frame clone lifecycle.
- Print the fresh Frame and publication identity returned by the server.
- Document that registered Frames v2 packages clone source into fresh sharing, sandbox ownership, and empty SQLite state.
- Tell agents to use the lifecycle command instead of raw filesystem copy.
- Release `dsbx` 0.1.53 and pin it in `dust-base` 0.8.101 so the documented command exists in live sandboxes.

## Tests

- 405 Rust unit tests plus the local forwarder end-to-end test.
- Rust clippy across all targets with warnings denied.
- Twenty-one sandbox image registry tests and ten Frames skill tests.
- Front typecheck and Biome checks.

## Risk

The command is additive and calls the feature-gated internal route from the parent PR. Legacy Frames and Pod Functions are unchanged. The sandbox image change is acknowledged with `sandbox-image-ack`.

## Deploy Plan

1. Publish the `dsbx-v0.1.53` release artifact.
2. Build and deploy `dust-base:0.8.101`.
3. Deploy front with the new image registry pin and Frames skill.
4. Roll existing sandboxes through `/poke/kill` so they recreate on the new base image before agents use clone.

### DIFF
--- a/cli/dust-sandbox/Cargo.lock
+++ b/cli/dust-sandbox/Cargo.lock
@@ -313,7 +313,7 @@ checksum = "92773504d58c093f6de2459af4af33faa518c13451eb8f2b5698ed3d36e7c813"
 
 [[package]]
 name = "dust-sandbox"
-version = "0.1.52"
+version = "0.1.53"
 dependencies = [
  "anyhow",
  "base64",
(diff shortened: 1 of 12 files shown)


fontanierh:
Superseded by the recut clone endpoint and CLI PRs #31496 and #31500.

PR #31407, opened 2026-08-28
Clean up failed Frame publications

fontanierh:
## Description\n\n- Delete the fresh publication prefix when artifact storage, database reconciliation, or activation fails.\n- Drain concurrent artifact uploads before cleanup and refuse to delete an active publication.\n- Keep cleanup best effort so it never masks the publication error.\n\n## Tests\n\nAdded focused coverage for partial artifact and descriptor writes, cleanup failure, active-pointer races, reconciliation failures, and activation failures while preserving the previous active publication.\n\n## Risk\n\nCleanup is limited to the newly generated inactive publication ID and rechecks the active pointer before deletion.\n\n## Deploy Plan\n\nDeploy after #31406, then standard deploy.


fontanierh:
Closed as deferred failed-publication cleanup hardening. It will be recut from current main after the core end-to-end lane is merged.

PR #31408, opened 2026-08-28
[front] fix: stop autoscroll after scrolling up

aubin-tchoi:
## Description

Fixes dust-tt/tasks#10209.

Long agent responses currently keep pulling the conversation to the bottom when someone scrolls up to read, which is especially disruptive on mobile because the screen is usually quite narrow, and when you try reading the message you are fighting against the scroll.

This PR makes auto-scroll distinguish streamed content growth from the user's upward viewport movement. If you don't do anything, content keeps following, but scrolling up "detaches" auto-scroll. Reaching the bottom re-attaches.

This was already what `updateAutoScrollEnabledFromLocation` was intended to do. It only worked if the user scrolled between content updates. In practice, during streaming `scrollHeight` changes constantly.

## Tests

- Added tests.
- Tested locally + preview.

## Risk

- Low.

## Deploy Plan

- Deploy front.

### DIFF
--- a/front/components/assistant/conversation/utils.ts
+++ b/front/components/assistant/conversation/utils.ts
@@ -18,6 +18,7 @@ import {
   isReinforcedSkillNotificationMetadata,
 } from "@app/types/assistant/conversation";
 import type { ContentFragmentType } from "@app/types/content_fragment";
+import { assertNever } from "@app/types/shared/utils/assert_never";
 import { truncate } from "@app/types/shared/utils/string_utils";
 import type { PodListItemType } from "@app/types/space";
 import moment from "moment";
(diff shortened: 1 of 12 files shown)


front/components/assistant/conversation/utils.ts
```diff
@@ -19,12 +21,39 @@ import type { ContentFragmentType } from "@app/types/content_fragment";
...
+  location,
+  previousLocation,
+}: {
+  isAutoScrollEnabled: boolean;
+  location: AutoScrollLocation;
+  previousLocation: AutoScrollLocation;
+}): boolean {
+  // Re-attach as soon as the reader reaches the bottom, including while the
+  // streamed message is still growing.
+  if (location.bottomOffset === 0) {
```
PopDaph: While I agree this is the logic behavior, iirc, in the past, we had some issues with re-attaching, it caused some weird behavior. 

I think we can still merge this and maybe just post a short message (no need to ping everyone) saying we changed the scrolling behavior and we want feedback 🙏🏻

PopDaph (approved):

PopDaph:
(I did not spent much time testing on the review app since it was approved by Alex!)

front/components/assistant/conversation/utils.ts
aubin-tchoi: agreed! there are indeed lots of potential edge cases

PopDaph:
@aubin-tchoi it keeps growing and growing 🤯 😅

aubin-tchoi:
Yeah this is not going so well aha

aubin-tchoi:
I think this is not a good look and it's easy to stumble onto it, but I'm struggling to find a working 80/20 and have no idea how close I am to finding a solution

aubin-tchoi:
I will unfortunately drop this one, I didn't fix one original goal on mobile (if you use touch actions to go up and hold it, it will start scrolling down incontrollably after a little bit) and introduced a few other issues
will wait for astra on codex and retry then

PopDaph:
🫡  that was quite a fight

PR #31409, opened 2026-08-28
Clean up stale Frame publication artifacts

fontanierh:
## Description

- Remove publication prefixes left without a publication.json commit marker before the next publish.
- Run cleanup under the Frame publish lock and preserve the active publication plus all committed inactive publications.
- Keep recovery best effort so storage discovery or deletion cannot block a valid publish.

## Tests

Added focused coverage for active and committed publication retention, stale prefix deletion, discovery and deletion failures.

## Risk

Cleanup targets only uncommitted prefixes for the same Frame. The active publication is always skipped. A writer is considered abandoned after the existing 10-minute publish-lock lease expires.

## Deploy Plan

Deploy after #31407, then standard deploy.

### DIFF
--- a/front/lib/api/frames/publication_storage.test.ts
+++ b/front/lib/api/frames/publication_storage.test.ts
@@ -899,6 +899,60 @@ describe("publishFramePublication", () => {
     );
   });
 
+  it("cleans only stale uncommitted publication artifacts", async () => {
+    const { auth, frame } = await setupFrame();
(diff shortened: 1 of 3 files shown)


fontanierh:
Closed as deferred stale-publication cleanup hardening. It will be rebuilt as a small child of the cleanup tip after end-to-end validation.

PR #31410, opened 2026-08-28
[front] enh: add button to copy link to skill

aubin-tchoi:
## Description

This PR adds a button in Manage skills to copy the link to a skill.
This link points to the Manage skills page, opened on the skill.

## Tests

- Tested locally + preview.

## Risk

- Low.

## Deploy Plan

- Deploy front.

### DIFF
--- a/front/components/skills/SkillDetailsButtonBar.tsx
+++ b/front/components/skills/SkillDetailsButtonBar.tsx
@@ -43,10 +43,6 @@ export function SkillDetailsButtonBar({
   const [showArchiveDialog, setShowArchiveDialog] = useState(false);
   const [isSkillLinkCopied, copySkillLink] = useCopyToClipboard();
 
-  if (!skill.canAdministrate && !onFavoriteChange) {
-    return null;
-  }
-
   return (
     <>
       <ArchiveSkillDialog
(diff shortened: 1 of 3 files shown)


front/components/skills/SkillDetailsButtonBar.tsx
```diff
@@ -70,6 +74,18 @@ export function SkillDetailsButtonBar({
             icon={Edit04}
           />
         )}
+        <Button
+          size="sm"
+          tooltip={isSkillLinkCopied ? "Copied!" : "Copy link"}
+          variant="outline"
+          icon={isSkillLinkCopied ? ClipboardCheck : Clipboard}
```
fabiencelier, line 81: is the copy icon the best ? I think a "share" icon would make sense here too ?
Debatable because you don't actually open a share menu and just copy the link 🤷 
I'm fine with both

fabiencelier (approved):

front/components/skills/SkillDetailsButtonBar.tsx
aubin-tchoi: yes share implies that you open a menu, typically with socials, here I just took an icon that we already had and that had a "check" version

PR #31412, opened 2026-08-29
Expose canonical Frames v2 invocation endpoint

fontanierh:
## Description

Add the canonical POST endpoint for invoking a Frames v2 function by stable Frame id and bare function name. Resolution is pinned to the active immutable publication, reuses the existing Frame use-rights check, and preserves the existing caller-identity policy and invocation response shape.

This private API is additive and feature-gated. Legacy Frame and Pod Function routes are unchanged. The client switch and canonical SSE route are intentionally separate so both servers can land independently. Based directly on `main`.

## Tests

- focused Front API Vitest: 1 file / 4 tests passed
- `npm run docs` in `front-api`
- pre-commit Biome, Swagger docs check, and Front typecheck
- synthetic POST + SSE final tree: POST 4/4 and SSE 4/4 tests passed after regenerating Swagger
- `git diff --check`

## Risk

Low and feature-gated. The route is additive, fails closed for a missing Frame, inactive publication, missing function, or missing use right, and delegates identity enforcement to the existing invocation authorization path. Rollback is reverting this PR.

## Deploy Plan

Deploy normally. No migration or backfill. Enable the client switch only after both canonical POST and SSE server routes are available.

### DIFF
--- a/front-api/public/swagger.json
+++ b/front-api/public/swagger.json
@@ -9335,6 +9335,78 @@
         }
       }
     },
+    "/api/w/{wId}/frames/{frameId}/functions/{name}/invocations": {
+      "post": {
+        "summary": "Invoke an active Frames v2 function",
+        "description": "Resolves a bare function name from the Frame's active immutable publication, checks Frame use rights, and starts an invocation pinned to that publication.",
+        "tags": [
+          "Private Frames"
+        ],
+        "parameters": [
+          {
+            "in": "path",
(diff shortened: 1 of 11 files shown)


fontanierh:
Superseded by the focused canonical invocation stack: #31523, #31538, #31524, and #31534. The contract, resolver, and transport are isolated with focused tests and strict Swagger validation.

PR #31415, opened 2026-08-29
Convert legacy Frames to v2 in place

fontanierh:
## Description

Add `dsbx frame convert <legacy-entry> <manifest>` for the explicit one-way v1 to v2 migration. Conversion preserves the FileResource identity, share URL, ACL rows, and legacy rendering until the first v2 publication activates. A persisted transition marker makes interrupted conversions retryable, while the shared legacy edit lock prevents queued v1 publishes from overwriting v2 metadata. Source and manifest must stay in the same conversation or Pod mount; Frame-owned databases can be added after conversion.

Register and emit the `frame.converted` WorkOS audit event.

## Tests

- `front-api`: Frames sandbox route test, 16 tests
- `front`: conversion, publication, rendering, and audit-schema tests, 82 tests
- `dust-sandbox`: 406 unit tests plus local forwarder e2e
- `dust-sandbox`: clippy with warnings denied
- Biome and rustfmt checks

## Risk

Medium and feature-gated. Conversion changes the stable FileResource binding in place. The transition is serialized with legacy and v2 publishing, keeps the legacy artifact renderable until activation, restores the complete legacy binding when publication fails, and resumes safely after interruption.

## Deploy Plan

Deploy normally behind the existing `frames_v2` feature flag. The `frame.converted` WorkOS schema is registered at version 1. No database migration or backfill is required.


fontanierh:
Superseded by the focused conversion stack:\n\n- #31539 conversion primitives\n- #31541 conversion service\n- #31542 HTTP surface\n- #31543 audit tip\n- #31544 CLI tip\n- #31546 legacy publish fence\n- #31547 recovery marker and legacy rendering\n- #31548 crash-safe recovery service\n\nAll replacements are live, tested, and each is under 350 total changed lines.

PR #31417, opened 2026-08-29
Garbage collect retired Frame publications

fontanierh:
## Description

Schedule durable cleanup whenever a Frames v2 publication is retired. Invocation admission revalidates the active publication and persists its invocation row under the same per-Frame lock used by publishing and cleanup, so cleanup cannot delete artifacts needed by an admitted invocation. Cleanup waits one hour, retries while any invocation is running, removes immutable GCS artifacts, deletes only unreferenced function rows, preserves invocation history, and never removes the active publication. Existing committed publications are discovered on the next publish. The workflow continues as new every 24 attempts to bound history.

## Tests

- `front`: 44 relevant files, 422 tests passed across Frames v2, legacy Frames, Pod Functions, publication storage, invocation execution, sharing, mounts, UI rendering, and Temporal activities
- `front-api`: 9 relevant files, 124 tests passed across public and authenticated Frame routes, canonical and legacy function invocation, and SSE
- `front` and `front-api`: `npx tsgo --noEmit`
- `front`: all 30 Temporal bundles built
- `npm run format:changed` and Biome on changed files

## Risk

Moderate. This deletes retired publication artifacts. Deletion runs only after a grace period, under the Frame operation lock, after confirming the publication is inactive and has no running invocation. Invocation admission creates its durable reference under that same lock. Cleanup retries durably, workflow history is bounded, and scheduling failures do not break publishing.

## Deploy Plan

Deploy the front service and sandbox-functions Temporal worker normally behind the existing Frames v2 feature flag.

### DIFF
--- a/front-api/routes/w/[wId]/frames/[frameId]/functions/[name]/invocations.ts
+++ b/front-api/routes/w/[wId]/frames/[frameId]/functions/[name]/invocations.ts
@@ -107,7 +107,7 @@ app.post(
     if (invocationResult.isErr()) {
       if (isSandboxFunctionInvocationError(invocationResult.error)) {
         return apiError(ctx, {
-          status_code: 401,
+          status_code: invocationResult.error.status,
           api_error: {
             type: invocationResult.error.code,
             message: invocationResult.error.message,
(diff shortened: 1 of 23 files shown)


fontanierh:
Closed as deferred publication-GC hardening. The old branch also contains lock and invocation work now replaced by the smaller lock and canonical-route lanes.

PR #31420, opened 2026-08-30
[front] Fix Metronome seat-sync edit storm: serialize full reconcile, narrow immediate seat writes

tdraier:
## Description

Fixes the Metronome seat-sync rate-limit incident (workspace Vanta: ~300 contract edits over ~25h, Metronome blocked contract edits for the customer).

Root cause was **concurrent, uncoordinated full seat reconciles** — not a formula bug (the reconcile is correct and idempotent when run serially, confirmed by local repro). Two families of full reconcile ran with no shared lock: the debounced Temporal workflow, and direct in-request `syncMetronomeSeatCountForWorkspace` calls. `membership.ts` fired **both** a workflow launch and a direct full sync for a single seat transition; `business_activation`, `usage_configuration`, and the poke plugin also ran direct full syncs. Overlapping runners each read the same pre-edit unassigned count and issued their own **open-ended** `add_unassigned_seats` delta (Metronome seat edits have `starting_at` only, no `ending_before`), so the unassigned pool **stacked** to multiples of the floor gap and never converged.

Fix — eliminate the concurrency by construction:
- The **full workspace reconcile runs only on the single serialized per-workspace Temporal workflow path**.
- New **`assignSeatForUser`** for the immediate-unblock need (auto-upgrade, business activation): writes only that one user's seat. Never touches the unassigned pool, other users, or alerts, so it is safe to overlap the workflow.
- **`moveSeatWithCreditCarry`** backs `assignSeatForUser` and does the per-user **seat-credit consumption carry** across a recurring-credit change (pro↔max: empty the old credit → move the seat → carry consumed AWU onto the new credit). This is required because the immediate path moves the seat right away, which would otherwise hide the pro→max transition from the debounced workspace sync and drop the carry (leaving the user a full fresh max allowance on top of what they already spent). The empty-origin loop is extracted into a shared helper reused by both the workspace sync and the single-user move.
- Repoint callers: membership immediate path → `assignSeatForUser` (workflow launch stays as backstop); `business_activation` → `isDirectSync` (single-seat + workflow backstop); `usage_configuration` + poke plugin → `launchMetronomeSeatCountSyncWorkflow`.
- Bound the activity retry policy (maxAttempts 5, scheduleToClose 1h) so a stuck sync can't retry unbounded and pile on self-overlap.

Admin-driven / debounced seat changes are unchanged — they already carry consumption correctly via `syncSeatCount`.

Also includes ops tooling: `scripts/audit_metronome_seat_state.ts` (`--probe`) to diff Dust vs Metronome seat state and dump raw seat-history segments, and `scripts/fix_metronome_future_seat_segment.ts` (dry-run one-off) to repair an inflated future segment.

## Tests

- New `lib/metronome/seats_credit_carry.test.ts`: pro→max empties the origin credit, moves the seat, then carries consumption onto the new credit — in that order; a plain add does no ledger transfer; a non-billable target no-ops.
- New `lib/api/metronome/seat_sync.test.ts`: `assignSeatForUser` delegates the move to `moveSeatWithCreditCarry` with the correct from→to seat types, reconciles the user, and no-ops when not on Metronome.
- Existing `lib/metronome/seats_sync.test.ts` and `lib/api/metronome/reconcile_credit_state.test.ts` still pass (25 tests total). `tsgo --noEmit` clean.
- Reproduced the healthy/idempotent reconcile locally via the poke plugin against a workspace with a minSeats floor + scheduled future floor change (serial runs converge to a no-op).

## Risk

- Medium — billing-facing seat reconcile + seat-credit ledger. Immediate seat updates no longer run a full workspace reconcile inline: they do a narrow single-seat write + per-user credit carry, and rely on the debounced workflow for the workspace-wide reconcile (unassigned floor, other members, alerts). Worst case if the narrow op overlaps the workflow is an off-by-one on the unassigned pool, self-corrected on the next full sync.
- The poke "Sync Metronome Seat Count" plugin is now async (schedules the workflow, ~15s) instead of returning a synchronous summary — operators read the outcome from logs.
- Safe to roll back: revert the commits; the debounced workflow path is unchanged in behavior.

## Deploy Plan

- Standard deploy; no migration.
- Post-deploy operational follow-up (separate from this deploy): once Metronome re-enables contract edits for Vanta, run `npx tsx front/scripts/fix_metronome_future_seat_segment.ts --workspaceId HzXDVQGUQF` (dry-run) then `--execute` to correct its inflated future Pro segment (1141 → 730) **before 2026-11-13**, else that segment bills 1141 seats.

### DIFF
--- a/front/lib/api/checkout/business_activation.ts
+++ b/front/lib/api/checkout/business_activation.ts
@@ -1,5 +1,4 @@
 import { updateMembershipSeatAndTrack } from "@app/lib/api/membership";
-import { syncMetronomeSeatCountForWorkspace } from "@app/lib/api/metronome/seat_sync";
 import {
   isMetronomeBillingEnabled,
   restoreWorkspaceAfterSubscription,
@@ -601,11 +600,17 @@ export async function handleSubscriptionActivationSuccess({
       checkoutPayment.seatType,
       checkoutPayment.billingPeriod
     );
+    // `isDirectSync` makes this immediately assign the target user's single seat
+    // (and reconcile their credit state) so they land correctly without waiting
+    // for the 15 s debounce — while the workspace-wide reconcile still goes
+    // through the debounced Temporal workflow it launches. This avoids running a
+    // second full reconcile here in parallel with the workflow.
     const seatResult = await updateMembershipSeatAndTrack({
       user: targetUser,
       workspace: lightWorkspace,
       newSeatType: membershipSeatType,
       author: "no-author",
+      isDirectSync: true,
     });
     if (seatResult.isErr()) {
       logger.error(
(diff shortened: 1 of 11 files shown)


tdraier:

- #31420 ◀

tdraier:
Superseded — split into #31422 (incident fix: retry cap + heartbeat + repair scripts) and #31423 (narrow immediate seat writes + pro↔max credit carry).

PR #31422, opened 2026-08-30
[front] Bound Metronome seat-sync retries + heartbeat to stop the edit storm

tdraier:
## Description

Focused fix for the Metronome seat-sync edit storm.

**Incident (from Metronome's contract-edit audit log):** one workspace's debounced seat-sync workflow issued **481 contract edits** on a single subscription over ~25h, piling **~548 phantom unassigned seats** into a future (scheduled-floor) segment. Mechanism, now certain:

- `add/remove_unassigned_seats` are **cumulative deltas**, but the reconcile computes them as "set unassigned to `floor − assigned`" from a read (`getSubscriptionSeatsHistory`, a reporting endpoint).
- The activity had **no retry cap** (Temporal default = unlimited attempts) → the single workflow activity ran **~240 attempts**. Each attempt re-read the seat state **before the read model reflected the prior attempt's edit** (the reads came back as multiples of the 87-seat floor — `0, 174, 261, …` — **never 87**) and re-applied the full delta, which **stacked**.
- A tight **1-min `heartbeatTimeout`** on a slow, rate-limited sync produced *false* timeouts that drove many of those retries.
- The scheduled future floor (Nov-13) added a second reconcile pass writing open-ended into a future segment that every now-pass edit also bled into, so the residue concentrated and froze there (`total 1141 / 548 unassigned` vs the correct `730 / 137`).

This PR bounds the retry/timeout behavior only — three guards:
- **`maximumAttempts: 5`** — a stuck sync can no longer re-apply its edits ~240 times.
- **`initialInterval` 30s → 5m** (`maximumInterval` → 10m) — the key one: spacing retries several minutes apart lets the seat-history read catch up, so the next attempt reads the converged value and computes a **zero delta** (converges) instead of re-adding. Seat-count sync is a billing reconcile, not a user-facing path (immediate effects go through `assignSeatForUser` in the follow-up PR), so minutes between retries is fine.
- **`heartbeatTimeout` 1m → 5m, `startToCloseTimeout` 10m → 20m** — a healthy-but-slow big-customer sync isn't falsely timed out. `scheduleToCloseTimeout` 1h → 2h to fit the bounded retries.

Also adds ops tooling used to diagnose/repair the incident:
- `scripts/audit_metronome_seat_state.ts` (`--probe`): diff Dust vs Metronome seat state per subscription + per-user balances; dump raw seat-history segments.
- `scripts/fix_metronome_future_seat_segment.ts`: dry-run-by-default one-off to correct an inflated future segment's unassigned pool to `minSeats − assigned`.

The deeper reconcile idempotency (delta-as-set → set-absolute, no-progress guard) and the direct-sync narrowing + credit-carry are in follow-up PR #31423.

## Tests

- `tsgo --noEmit` clean; existing seat-sync suites unaffected (this is a Temporal timeout/retry config change).
- The scripts are dry-run/read-only by default; used against the affected workspace to confirm current state and the pending future-segment correction.

## Risk

- Low. Changes only the retry policy / timeouts of `syncMetronomeSeatCountActivity`, plus two read-only/dry-run scripts. No behavior change to the sync logic itself. Retries are now spaced minutes apart, delaying a failed *reconcile* by a few minutes — acceptable (not user-facing; a real membership change re-signals the workflow anyway). Safe to roll back by reverting.
- Note: the retry cap + spacing is a strong mitigation, not a guarantee — it relies on the read lag being shorter than the 5-min interval. The delta-as-set root is removed properly in a later PR (set-absolute / no-progress guard).

## Deploy Plan

- Standard deploy; no migration.
- Post-deploy operational follow-up: once Metronome re-enables contract edits for the affected workspace, run `npx tsx front/scripts/fix_metronome_future_seat_segment.ts --workspaceId HzXDVQGUQF` (dry-run) then `--execute` to correct its inflated future Pro segment (1141 → 730) **before 2026-11-13**, else that segment bills 1141 seats.

### DIFF
--- a/front/scripts/audit_metronome_seat_state.ts
+++ b/front/scripts/audit_metronome_seat_state.ts
@@ -1,0 +1,490 @@
+/**
+ * Read-only audit of a workspace's seat state: compares what Dust believes
+ * (DB membership seat types) against what Metronome actually holds (assigned
+ * seat IDs + unassigned pool per seat subscription) and the per-user seat
+ * credit balances.
+ *
+ * Written to diagnose the seat-sync loop on workspaces where `syncSeatCount`
+ * keeps re-issuing the same seat edit because some users desired in Dust never
+ * land in Metronome's assigned set (so the reconcile never converges). It
+ * surfaces exactly which users are missing / stale per seat type, and each
+ * assigned seat's live balance, so we can figure out why they won't assign
+ * before touching credit state.
+ *
+ * Purely diagnostic — makes only READ calls to Metronome and the DB. There is
+ * nothing to --execute; it always just reports.
+ *
+ *   npx tsx scripts/audit_metronome_seat_state.ts --workspaceId <wId>
+ */
+import config from "@app/lib/api/config";
+import {
+  getMetronomeClient,
+  getMetronomeSubscriptionSeatState,
+  listMetronomeSeatBalances,
+} from "@app/lib/metronome/client";
+import { getCreditTypeAwuId } from "@app/lib/metronome/constants";
+import { getActiveContract } from "@app/lib/metronome/plan_type";
+import {
+  getProductSeatTypes,
+  getSeatSubscriptionsFromContract,
+} from "@app/lib/metronome/seat_types";
+import { MembershipResource } from "@app/lib/resources/membership_resource";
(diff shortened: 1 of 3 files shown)


tdraier:

- #31422 ◀

fabiencelier (commented):
LGTM

PR #31423, opened 2026-08-30
[front] Narrow immediate seat writes: assignSeatForUser + pro↔max credit carry

tdraier:
## Description

Part of the seat-sync follow-ups (incident fix is #31422; poke/usage-config → workflow routing is #31568). This PR narrows the **immediate** (non-debounced) seat path so it no longer runs a full workspace reconcile inline.

- The full `syncSeatCount` (which manages the unassigned-seat pool that stacked in the incident) now runs **only on the single serialized per-workspace Temporal workflow**.
- The immediate path (auto-upgrade `free→pro`/`pro→max`, business activation — the only `isDirectSync` callers) uses a new narrow **`assignSeatForUser`** that writes just that one user's seat:
  - membership immediate path → `assignSeatForUser` (+ workflow backstop, launched unconditionally)
  - `business_activation` → `isDirectSync` (single-seat + workflow backstop)
- **`moveSeatWithCreditCarry`** carries the seat-credit consumption on a `pro↔max` move: `assignSeatForUser` pre-moves the seat_id, which would hide the transition from the later debounced sync (it detects moves by diffing Metronome vs the DB) and drop the ledger transfer. So it does the per-user carry directly — **empty origin credit → move seat → carry consumed AWU onto the new credit** (same order/logic as the workspace sync). The empty-origin loop is extracted into a shared `emptyOriginCreditsForTransfers` reused by both paths.

**Scope notes:**
- Admin-driven / debounced seat changes are unchanged (they already carry via `syncSeatCount`).
- Free-seat credit grant/revoke stays **workflow-owned**; the immediate path only handles pro/max (the `isDirectSync` upgrade targets — it never assigns a free seat). A `free→pro` upgrade's free-credit *revoke* lags to the debounced workflow (~15s) — a brief over-allowance, never a block.
- The seat move on the immediate path is load-bearing (the new tier's credit only exists once the seat is assigned, and that's what un-caps the user) — a creditState-only update can't unblock an upgraded user.

## Tests

- New `lib/metronome/seats_credit_carry.test.ts`: `pro→max` empties the origin credit, moves the seat, then carries consumption — in that order; a plain add does no ledger transfer; a non-billable target no-ops.
- New `lib/api/metronome/seat_sync.test.ts`: `assignSeatForUser` delegates to `moveSeatWithCreditCarry` with the correct from→to seat types, reconciles the user, no-ops when not on Metronome.
- Existing `seats_sync.test.ts` / `reconcile_credit_state.test.ts` still pass. `tsgo --noEmit` clean.

## Risk

- Medium — touches the immediate seat path + seat-credit ledger. Immediate seat updates do a narrow single-seat write + per-user credit carry, relying on the debounced workflow for the workspace-wide reconcile. Worst case if the narrow op overlaps the workflow is an off-by-one on the unassigned pool, self-corrected next sync. Follow-up worth considering: drop `maxRetries` to 0 for seat edits so an immediate-path edit can't stack on a 504 (see incident notes). Safe to roll back by reverting.

## Deploy Plan

- Standard deploy; no migration. Merge after #31422; independent of #31568.

### DIFF
--- a/front/lib/api/checkout/business_activation.ts
+++ b/front/lib/api/checkout/business_activation.ts
@@ -1,5 +1,4 @@
 import { updateMembershipSeatAndTrack } from "@app/lib/api/membership";
-import { syncMetronomeSeatCountForWorkspace } from "@app/lib/api/metronome/seat_sync";
 import {
   isMetronomeBillingEnabled,
   restoreWorkspaceAfterSubscription,
@@ -601,11 +600,17 @@ export async function handleSubscriptionActivationSuccess({
       checkoutPayment.seatType,
       checkoutPayment.billingPeriod
     );
+    // `isDirectSync` makes this immediately assign the target user's single seat
+    // (and reconcile their credit state) so they land correctly without waiting
+    // for the 15 s debounce — while the workspace-wide reconcile still goes
+    // through the debounced Temporal workflow it launches. This avoids running a
+    // second full reconcile here in parallel with the workflow.
     const seatResult = await updateMembershipSeatAndTrack({
       user: targetUser,
       workspace: lightWorkspace,
       newSeatType: membershipSeatType,
       author: "no-author",
+      isDirectSync: true,
     });
     if (seatResult.isErr()) {
       logger.error(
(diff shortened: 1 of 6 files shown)


tdraier:

- #31423 ◀

PR #31424, opened 2026-08-31
[front] feat: name the group behind an inherited pool credit spend limit

avervaet:
The spend-limit tooltip on a member's pool credit usage said "Limit
from a group" whenever the cap was inherited from a group, without
saying which one. Track the winning group's name alongside its cap so
the tooltip can name it; override and workspace-default limits are
unaffected.


avervaet:

- #31125
    - #31126
        - #31134
            - #31155
                - #31241
                    - #31339
                        - #31349
                            - #31424 ◀

PR #31429, opened 2026-08-31
[marketing] feat: promo banner for "Take control of your AI spend"

martindust:
Replace the promo schedule with the September 17th online event and drop the two August entries, which had already expired and could no longer render.

Also render the promo image, which the config carried but the card never displayed. It sits full-bleed at the top of the card on `sm` and up, and is hidden below that so the banner stays small on a phone.

## Deploy Plan

Marketing


martindust:
@matteotrab @id13 @matthieuhaguen can someone validate my PR pls?

matteotrab (approved):

PR #31430, opened 2026-08-31
Admin governance: drop the builder role from tests

fabiencelier:
## Description

Remove "builder" usage in tests

 - replace by "users" when possible, sometimes by admin when necessary
 - Uses MembershipRoleType to type "any roles" instead of listing them all
 - fixes some tests doing "all non admin roles" with [builder, user] to [manager, user]

## Deploy Plan

deploy front
deploy front-sse


ykmsd (approved):
ty!

PR #31436, opened 2026-08-31
chore: gate file_generation to conversations without the sandbox

adrsimon:
## Description

The sandbox already generates and converts files, so keeping `file_generation` next to it gives the model two ways to do the same thing. It is now dropped from the resolved server list whenever the sandbox is part of a run (in `deduplicateMCPServerConfigurations`, so the sandbox child-action path gets it too), and hidden from the tools picker when the computer feature is enabled. Existing agents that have it configured still resolve and render fine, and conversations without the sandbox keep the tool.

## Tests

By hand

## Risk

Low. First step to deprecation of this tool until we fully release sbx.

## Deploy Plan

Deploy front.


fontanierh (approved):
thank you

PR #31438, opened 2026-08-31
[front] consumption chart: paint today label above bars

PopDaph:
On the analytics consumption chart, the "Today (partial)" pill was painted behind the bars: the `ReferenceLine` was declared before the `Bar` series, and recharts paints SVG children in JSX order. When the tallest bar of the period sits next to the today line and reaches near the top of the axis, it covers the label.

Fix: declare the `ReferenceLine` after the bars so the line and its label paint on top.

Before / After:
<kbd>Screenshot 2026-08-31 at 11 29 44</kbd>
<kbd>Screenshot 2026-08-31 at 11 34 02</kbd>


adrsimon (approved):
LGTM

PR #31440, opened 2026-08-31
[front] agents table: show model tier chip

PopDaph:
On the manage agents page, the Model column now shows the model's tier (Basic / Standard / Premium) next to the provider icon and model name, reusing `ModelTierChip`.

The tier resolves from the agent's model and stored reasoning effort. Some stored efforts map to no tier for their model (e.g. `medium` kept from a previous model after switching to the `auto` stream, which is only tiered at `none`): in that case resolution falls back to the model's default effort, so the chip always shows a tier. This lives in `getTierForModelConfiguration` and `getTieredReasoningEffort`, next to `getTierForModel`, with tests.

Since two agents on the same model can be on different tiers because of their reasoning effort, the cell tooltip shows the effort the tier resolves at (e.g. "Claude Opus 4.8 High"), following the model picker's naming.

Streams (`auto`, `auto_fast`, `auto_complex`) show only the chip, no model name: they are named after their tier, so the name would repeat the chip.

Long model names truncate with an ellipsis before the chip does, so the chip is always fully visible.

Test it here: https://pr-31440-merge--app.preview.dust.tt/w/0ec9852c2f/builder/agents#?selectedTab=all

### DIFF
--- a/front/components/assistant/manager/AssistantsTable.tsx
+++ b/front/components/assistant/manager/AssistantsTable.tsx
@@ -394,7 +394,7 @@ const getTableColumns = ({
         );
       },
       meta: {
-        className: "hidden @sm:w-28 @sm:table-cell @xl:w-60",
+        className: "hidden @sm:w-28 @sm:table-cell @xl:w-56",
       },
     },
     {
--- a/front/components/assistant/manager/AssistantsTable.tsx
+++ b/front/components/assistant/manager/AssistantsTable.tsx
@@ -25,6 +25,7 @@ import type {
   AgentUsageType,
   LightAgentConfigurationType,
 } from "@app/types/assistant/agent";
+import { isModelStreamId } from "@app/types/assistant/models/auto";
 import { getTieredReasoningEffort } from "@app/types/assistant/models/model_tiers";
 import { getModelMaker } from "@app/types/assistant/models/providers";
 import type {
@@ -360,6 +361,11 @@ const getTableColumns = ({
             ? `${modelName} ${capitalize(reasoningEffort)}`
             : modelName;
 
+        // Streams are named after their tier: the chip alone carries the info.
+        const isStreamModel = modelConfig
+          ? isModelStreamId(modelConfig.modelId)
+          : false;
+
         return (
           <Tooltip
             tooltipTriggerAsChild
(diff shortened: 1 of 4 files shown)


front/components/assistant/manager/AssistantsTable.tsx
```diff
@@ -348,16 +358,28 @@ const getTableColumns = ({
...
                   iconClassName="mr-0 @xl:mr-2"
                 >
-                  <span className="hidden @xl:inline">{modelName}</span>
-                  {!modelIcon && <span className="@xl:hidden">-</span>}
+                  <div className="flex min-w-0 items-center gap-2">
+                    <span className="hidden min-w-0 truncate @xl:inline">
+                      {modelName}
+                    </span>
+                    {!modelIcon && <span className="@xl:hidden">-</span>}
+                    {modelConfig && (
```
aubin-tchoi, line 389: on a laptop screen this will probably only show the chip and not the model name

front/components/assistant/manager/AssistantsTable.tsx
```diff
@@ -348,16 +358,28 @@ const getTableColumns = ({
...
+                      <div className="shrink-0">
+                        <ModelTierChip
```
aubin-tchoi, line 391: since the reasoning effort is not exposed anywhere we could have an issue where two agents on Claude Sonnet 5 are on different tiers with no visible explanation

PopDaph: Then the explanation could be on a tooltip?

front/components/assistant/manager/AssistantsTable.tsx
PopDaph: Approved by Designer :D

front/components/assistant/manager/AssistantsTable.tsx
PopDaph: I can add the reasoning effort on the tooltip, adding a commit

front/components/assistant/manager/AssistantsTable.tsx
aubin-tchoi: wfm, counter uno then: can we add a little horizontal space?

PopDaph: 🫡

aubin-tchoi (approved):
LGTM ✨

PR #31441, opened 2026-08-31
Skills: Store manually requested spaces 2.5/4: backfill archived and suggested skills

fabiencelier:
## Description:

Follow-up to #31341. The backfill logged `Some skills could not be fetched and are left untouched` and skipped archived and suggested skills: `baseFetch` filters to `status: "active"` unless a caller overrides it, and `fetchByModelIds` did not expose that option. Those skills still hold their requested spaces and can be restored and saved, so they need the provenance stored too — otherwise 3/4 would drop their hand-picked spaces on the next save. `fetchByModelIds` now takes a `status` option, like the other fetchers, and the backfill asks for all three statuses.

## Tests
Exercised against a local database seeded with an active, an archived and a suggested skill, each holding requested spaces: before the fix the script reported 3 requested / 1 fetched and processed only the active one, after the fix it processes all three and the warning is gone.

## Deploy plan
Deploy front. No migration to run, but the backfill must be re-run afterwards so the skills it skipped are covered:
`npx tsx migrations/20260828_backfill_skill_manually_requested_spaces.ts --execute`

### DIFF
--- a/front/lib/resources/skill/skill_resource.test.ts
+++ b/front/lib/resources/skill/skill_resource.test.ts
@@ -2127,6 +2127,34 @@ describe("SkillResource", () => {
     });
   });
 
+  describe("fetchByModelIds", () => {
+    it("returns active skills only unless a status is given", async () => {
+      const activeSkill = await SkillFactory.create(testContext.authenticator, {
+        name: "Active Skill",
+      });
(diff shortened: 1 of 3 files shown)


philipperolet:
LGTM

PR #31444, opened 2026-08-31
[Admin Governance] Batch agent identity backfill

philipperolet:
## Description

Follows https://github.com/dust-tt/dust/pull/31330.

The original backfill loads every agent version and opens one transaction per logical agent. This becomes slow for workspaces with many agents, including reruns where most rows are already attached.

This change scans each workspace once to collect distinct logical agents with at least one version missing an identity, keeps only those short sIds in memory, then processes configurable write batches of 1,000. Each batch bulk-creates missing identities and attaches every version with one joined update. Dry-run reporting, idempotency, link validation, and the final completeness check remain in place.

## Risks

Blast radius: the one-time agent identity backfill script.

Risk: standard.

## Deploy Plan

- Deploy front.
- Run the backfill in dry-run mode for target workspaces and inspect the counts.
- Execute it with the default batch size of 1,000, lowering --batchSize if needed.
- Verify that no agent configuration is missing agentId before #31331.

## Tests

- [x] NODE_ENV=test npm test -- migrations/20260828_backfill_agent_identities.test.ts

### DIFF
--- a/front/migrations/20260828_backfill_agent_identities.ts
+++ b/front/migrations/20260828_backfill_agent_identities.ts
@@ -12,14 +12,14 @@ import { Op, QueryTypes } from "sequelize";
 
 const DEFAULT_BATCH_SIZE = 1_000;
 
-const SELECT_BATCH_SQL = `
-  SELECT DISTINCT "sId"
+// Applying the null check after grouping makes PostgreSQL scan one workspace once instead of
+// walking a global index to find matching configurations for every batch.
+const SELECT_MISSING_AGENT_IDS_SQL = `
+  SELECT "sId"
   FROM agent_configurations
   WHERE "workspaceId" = :workspaceId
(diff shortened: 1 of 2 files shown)


fabiencelier (approved):

front/migrations/20260828_backfill_agent_identities.ts
```diff
@@ -19,123 +42,152 @@ export type AgentIdentityBackfillStats = {
...
-  return identity ?? null;
+  return {
+    logicalAgentCount,
+    identitiesToCreate: 0,
+    versionsToAttach,
+    orphanIdentityCount: Number(count),
+  };
 }
 
 export async function backfillAgentIdentities({
```
fabiencelier, line 114: maybe worth adding a small doc to explain the process:  it first fills the agents, then use the generated id to fill the agent_configurations back

PR #31439, opened 2026-08-31
[sparkle] add Databricks platform logo

FlagBenett:
## Summary

Adds a `DatabricksLogo` platform logo to the Sparkle design system, so it can be used as the icon for the upcoming Databricks managed MCP server presets (SQL, Genie).

## Changes
- `src/logo/src/platforms/Databricks.svg` — source SVG (brand red `#ee3d2c`).
- `src/logo/platforms/Databricks.tsx` — generated component (svgr; `1em`, viewBox preserved).
- `src/logo/platforms/index.ts` — export `DatabricksLogo`.
- `src/logo/platforms/registry.ts` — add to `PLATFORM_LOGOS` (so it appears in the Assets → Platform Logos catalog).

## Test plan
- `tsgo` clean; biome clean.
- Visible under Storybook → Assets → Platform Logos.
- tested locally

## deploy
- sparkle only

PR #31445, opened 2026-08-31
[front] Databricks: official managed MCP servers as remote presets

FlagBenett:
## Summary

Replaces the internal Databricks MCP proof-of-concept (a single `list_warehouses` tool hitting the REST API) with the **official Databricks-managed MCP servers**, exposed as **remote MCP server presets** — the same pattern as Asana/Notion/etc. (`DEFAULT_REMOTE_MCP_SERVERS`, `oauth-static`). Tools are discovered + cached by the existing remote-MCP path, so we don't hand-maintain tool lists.

Depends on #31439 (adds the `DatabricksLogo` used here), already merged to `main`.

## Changes

**New presets** (`lib/actions/mcp_internal_actions/remote_servers.ts`):
- **Databricks SQL** and **Databricks Genie**, `authMethod: "oauth-static"`, `featureFlag: "databricks_tool".
- Their URL is per-workspace, so `url: ""` the add dialog shows the URL field for the user to enter `https://<workspace-host>/api/2.0/mcp/sql` 
- No changes to shared MCP dialog/provider files.

**Removed the legacy internal server** (`lib/api/actions/servers/databricks/`):
- Deleted the server, reserved id 45 in `LEGACY_INTERNAL_MCP_SERVER_IDS`, and cleaned it from `constants.ts`, `servers/index.ts`, `tool_display_labels.ts`, the bm25 corpus/test, and the metadata + availability snapshots.

**Icon registry**
- registered `DatabricksLogo` as an allowed internal icon.

## Deployment

SDK and front


Fraggle (approved):
LGTM, let's notify eng runner / support just in case we broke something.

PR #31446, opened 2026-08-31
feat(workspace-analytics): move get_usage_timeseries onto the consumption index

achilleburah:
Last of the stack — on #31303, itself on #31300, #31290 and #31273. Review those first; this diff is against #31303's branch.

## Description

`get_usage_timeseries` is the last reader of `front.agent_message_analytics` in `workspace_analytics`; re-pointed at the consumption index, same name.

With no legacy tool left, two pieces of scaffolding go:

- `usageFilterSchema`, which existed only because the old tools took a singular `source` over `context_origin`.
- `toConsumptionPeriod`, the +1ms conversion from the old inclusive `lte` bound. `get_credit_timeseries` (#31303) and the ranking/overview tools switch to the raw resolved window at their call sites.

`workspace_analytics` now runs entirely on the consumption index: six tools, and `get_agent_details` which only ever read Postgres.

## Tests

- `activity_timeseries.test.ts`: distinct messages rather than documents, tool executions counted from tool documents only, a skill execution requiring an attributed skill, and bucketing in the caller's timezone.
- Review fix: date labels used a bare UTC slice instead of `formatDateFromMillis`, same timezone bug as #31303.

## Risk

Medium. The tool changes index, and message counts become billed messages only — the consumption indexer skips failed and unbilled messages. Caller is `@analyst` only.

## Deploy Plan

Merge after #31303. Last of five: credits ranking → count rankings → overview → credit timeseries → usage timeseries.

### DIFF
--- a/front/lib/api/actions/servers/workspace_analytics/tools/index.ts
+++ b/front/lib/api/actions/servers/workspace_analytics/tools/index.ts
@@ -1,4 +1,5 @@
 import { MCPError } from "@app/lib/actions/mcp_errors";
+import { SKILL_MANAGEMENT_SERVER_NAME } from "@app/lib/actions/mcp_internal_actions/constants";
 import type {
   ToolHandlerResult,
   ToolHandlers,
@@ -6,7 +7,11 @@ import type {
 import { buildTools } from "@app/lib/actions/mcp_internal_actions/tool_definition";
 import { workspaceManagerGuard } from "@app/lib/actions/mcp_internal_actions/utils";
 import { WORKSPACE_ANALYTICS_TOOLS_METADATA } from "@app/lib/api/actions/servers/workspace_analytics/metadata";
-import type { ResolvedTimeWindow } from "@app/lib/api/actions/servers/workspace_analytics/query_input";
+import type {
+  ConsumptionFilterInput,
+  ResolvedTimeWindow,
+  TimeWindowInput,
+} from "@app/lib/api/actions/servers/workspace_analytics/query_input";
 import {
   DEFAULT_CREDIT_GROUPS,
   DEFAULT_RESULTS,
@@ -14,6 +19,10 @@ import {
   toConsumptionPeriod,
   toConsumptionScope,
 } from "@app/lib/api/actions/servers/workspace_analytics/query_input";
+import type {
+  ConsumptionTopDimension,
+  ConsumptionTopRankBy,
+} from "@app/lib/api/analytics/consumption/scope";
 import { CONSUMPTION_TOP_DIMENSION_UNIT } from "@app/lib/api/analytics/consumption/scope";
 import {
   fetchConsumptionTopGroups,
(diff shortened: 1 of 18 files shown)


front/lib/api/actions/servers/workspace_analytics/tools/index.ts
```diff
@@ -383,103 +308,68 @@ const handlers: ToolHandlers<typeof WORKSPACE_ANALYTICS_TOOLS_METADATA> = {
...
     const deniedError = workspaceManagerGuard(auth);
     if (deniedError) {
       return new Err(deniedError);
     }
 
-    const window = resolveTimeWindow(
-      { period, startDate, endDate, timezone },
-      "last_30_days"
-    );
+    const window = resolveTimeWindow(input, "last_30_days");
```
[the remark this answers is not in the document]
achilleburah: @ZeroPath false-positive because this data is made accessible to managers on purpose.

front/lib/api/analytics/consumption/activity_timeseries.ts
```diff
@@ -0,0 +1,220 @@
...
+    case "messages":
+      return {
+        messages: Math.round(bucket.messages?.value ?? 0),
+        conversations: Math.round(bucket.conversations?.value ?? 0),
+        activeUsers: Math.round(bucket.activeUsers?.value ?? 0),
+      };
+    case "tools":
+    case "skills":
+      return {
+        executions: bucket.executions?.doc_count ?? 0,
```
avervaet, line 157: Could skill activity count the number of tool.attributed_skill_ids values instead of tool documents? One invocation can be attributed to multiple skills, so this undercounts skill executions versus the ranking semantics; test one document with two skill IDs.

front/lib/api/actions/servers/workspace_analytics/tools/index.ts
```diff
@@ -420,102 +357,64 @@ const handlers: ToolHandlers<typeof WORKSPACE_ANALYTICS_TOOLS_METADATA> = {
...
       return new Err(new MCPError(window.error, { tracked: false }));
     }
+    const { filter, agentTagIds } = toConsumptionScope(input);
 
-    const baseQuery = scopedBaseQuery(auth, window.value, {
-      source,
-      agentIds,
-      userIds,
+    const result = await fetchConsumptionActivityTimeseries(auth, {
+      period: window.value,
```
avervaet: Could we preserve the +1ms conversion, or make resolveTimeWindow return an exclusive end? Explicit endDate is documented as inclusive but the consumption query uses < endDate, omitting records at 23:59:59.999Z; please add an endpoint-boundary test.

avervaet: Low impact but the API promises an inclusive date range, while the query silently treats it as exclusive..

avervaet (approved):
LGTM

achilleburah:
Tool was deleted in the end. Migration not deemed necessary.

PR #31447, opened 2026-08-31
chore(spa): tag browser logs and RUM with region and cell

adrsimon:
## Description

Browser logs and RUM events from US and EU both go to the same Datadog org (`site` is `datadoghq.eu` for both), and nothing on the event says where the user was. Only `env` (prod/dev) and `service` are set, so a query can't be scoped to one region or one deployment.

`useDatadogLogs` now puts two things in the global context:

- `region`, from `RegionContext`. Follows the live region, which matters because one SPA build serves both and a user can switch mid-session through `setRegionInfo`.
- `cell`, from a new build-time `VITE_DUST_CELL` set per region in `.github/configs/*/.env.{prod,edge}`, mirroring how `VITE_DUST_REGION` is already wired. The `CELL` env var the server reads isn't reachable from the browser.

Both go into the existing `wId` effect rather than their own: `setGlobalContext` replaces the whole context, so separate effects would clobber each other.

`cell` carries no information beyond `region` while the two map 1:1. It's there so the tag exists before cells stop lining up with regions. Since it's fixed at build time it describes the deployment that served the assets, not the backend being called, so it can disagree with `region` after a mid-session switch.

Events emitted before `AppRootLayout` mounts still have neither tag.

## Tests

Not tested. Worth a look in Datadog after deploy to confirm the tags land, and that `workspaceId` still appears on workspace pages and is absent elsewhere.

## Risk

Low. One hook plus config, no behavior outside telemetry. Worst case a tag is missing or wrong, which is where we already are.

## Deploy Plan

Deploy front, then the SPA to both regions. `VITE_DUST_CELL` is baked in at build time, so the cell tag only shows up once each region rebuilds.


front/hooks/useDatadogLogs.ts
```diff
@@ -26,20 +30,12 @@ export function useDatadogLogs() {
...
-        });
-      });
-    } else {
-      datadogLogs.setGlobalContext({});
-      window.DD_RUM?.onReady(() => {
-        window.DD_RUM?.setGlobalContext({});
-      });
-    }
-  }, [wId]);
+    const globalContext: Record<string, string> =
```
flvndvd: As we work on single tenant, mind adding the cell as well?

flvndvd (approved):

PR #31448, opened 2026-08-31
Admin governance: Remove some builder role usage

fabiencelier:
## Description

related to https://github.com/dust-tt/tasks/issues/9755

Follow-up to #31430 (builder role dropped from tests). Now that every builder membership has been migrated to `user`, a few leftovers in production code are unreachable or actively misleading. No observable behavior change.

Each commit stands alone:

  - **`a61eaecb` Drop builder from the member search role filter** — `resolveRoleFilterUserIds` expanded a `user` filter to `["user", "builder"]`. It only queries active memberships, none of which carry the role any more, so the expansion is dead. Its test asserted exactly that, so it now covers the member-role filter with two regular members rather than being deleted.
  - **`e2af3b35` Correct stale builder-sync documentation** — `createMembership` / `updateMembershipRole` still told callers to invoke `syncBuilderGroupMembership` after every membership write; that stopped being true when the role-driven call sites left `lib/api/membership.ts`, so following the instruction today is a no-op at best. The sync's own doc claimed the role was its source of truth — its only remaining driver is the `dust-builders` SCIM group. Docs only.
  - **`85edff79` Rename `AGENT_IDS_RESTRICTED_TO_BUILDER`** — gates `canShowAgentConversationActions` and never had anything to do with the role; the name just made it look like role logic. Now `AGENT_IDS_WITHOUT_CONVERSATION_ACTIONS`.
  - **`73faf0af` Fix the marketing `RoleType`** — listed the deprecated `builder` and was missing `manager` entirely, so it was wrong in both directions. Marketing only forwards `role` to PostHog.

## Tests

No new tests; this is dead-code and documentation cleanup.

## Risk

Low, safe to roll back (revert any commit independently, no migration or data change)

## Deploy Plan

deploy front

### DIFF
--- a/front/lib/resources/group_resource.ts
+++ b/front/lib/resources/group_resource.ts
@@ -2670,24 +2670,19 @@ export class GroupResource extends BaseResource<GroupModel> {
    * Transitional — builder role deprecation, see
    * https://github.com/dust-tt/tasks/issues/9459.
    *
-   * Keeps a per-workspace "Builders" group (`regular_manual`) in sync with the `builder`
-   * role so that when the role is removed, the group can be granted the builders' governance
-   * capabilities (create agents / skills) and former builders keep their rights. Until then
-   * the role is the source of truth: the group is created lazily and manual edits may be
-   * undone by the sync. Once the role is removed, the sync goes away and the group becomes
-   * fully admin-managed.
+   * Maintains the per-workspace "Builders" group (`regular_manual`), which holds the governance
+   * capabilities (create agents / skills) that the `builder` role used to confer, so former
+   * builders kept their rights when the role was migrated away.
    *
-   * The group is deliberately independent from SCIM provisioning: provisioned
-   * "dust-builders" groups proved unreliable (drifted membership, stale groups after
-   * deprovisioning). Workspaces provisioning builders get both groups — IdP changes flow
-   * through role assignment, which keeps this group in sync automatically.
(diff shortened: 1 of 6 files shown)


front/lib/resources/group_resource.ts
```diff
@@ -2670,24 +2670,19 @@ export class GroupResource extends BaseResource<GroupModel> {
...
-   * fully admin-managed.
+   * Maintains the per-workspace "Builders" group (`regular_manual`), which holds the governance
+   * capabilities (create agents / skills) that the `builder` role used to confer, so former
+   * builders kept their rights when the role was migrated away.
    *
-   * The group is deliberately independent from SCIM provisioning: provisioned
-   * "dust-builders" groups proved unreliable (drifted membership, stale groups after
-   * deprovisioning). Workspaces provisioning builders get both groups — IdP changes flow
-   * through role assignment, which keeps this group in sync automatically.
+   * The `builder` role is no longer the source of truth and no longer drives this sync — every
```
ykmsd, line 2677: I don't really know how it works but do we need to keep this forever? 🤔

ykmsd (approved):

front/lib/resources/group_resource.ts
fabiencelier: Don't know what's the plan for stopping the sync between dust-builders provisionned group and the "builders" manual group

PR #20360, opened 2026-01-21
Remove support for `allowUserOutsideCurrentWorkspace`

spolu:
## Description

This is now unused. Removing as this is basically security debt.

## Tests

Typechecking + tested locally

## Risk

None

## Deploy Plan

- deploy `front`

PR #6648, opened 2024-08-05
Add x-dust-user-id header to run dust apps on public api endpoint

flvndvd:
## Description

See thread [here](https://dust4ai.slack.com/archives/C050SM8NSPK/p1722612328124899).

This PR introduces a new `x-dust-user-id` header on the run endpoint. When used with a system key, it will be used to fetch the user's group within the key's workspace. If the header is not provided, the default will be the key's workspace group. This header is supported exclusively on this public API endpoint and serves as a temporary solution until we refactor our authentication mechanism.

This PR also refactors the `Authenticator.fromKey()` function to returns two authenticator, the one associated with the provided workspace (workspaceAuth) and the one associated with the key (keyAuth).

PR #31449, opened 2026-08-31
Admin governance: Remove the unused keyAuth from Authenticator.fromKey

fabiencelier:
## Description

`fromKey` returned two Authenticators: `workspaceAuth` for the target workspace, and `keyAuth` for the key's own workspace. `keyAuth` was added in #6648 (Aug 2024) for cross-workspace Dust app runs: the app runs endpoint used it to read the key workspace's providers and feature flags, and to check `app.canRead(keyAuth)`.

That went away in Jan 2026: https://github.com/dust-tt/dust/pull/20357 ("Remove ability to execute dust apps cross workspacee") rewrote every use in that endpoint to `auth`, and https://github.com/dust-tt/dust/pull/20360 dropped the `keyAuth` parameter from `withPublicAPIAuthentication` altogether. Since then nothing in production has read it — the only caller, `public_api_auth.ts`, destructures `workspaceAuth` and discards the rest. Two tests asserted on it.

Dropping it removes unnecessary work `fromKey` did on public API request.

This also removes one of the last `role: "builder"` references in production: `keyAuth` was hardcoded to the deprecated role.

## Deploy Plan

deploy front

### DIFF
--- a/front/lib/auth.ts
+++ b/front/lib/auth.ts
@@ -910,16 +910,14 @@ export class Authenticator {
    *                                   possible with a system key).
    * @param requestedRole optional role to assign the auth in place of the key role (only possible
    *                               with a system key).
-   * @returns Promise<{ workspaceAuth: Authenticator }>
+   * @returns Promise<Authenticator>
    */
   static async fromKey(
     key: KeyResource,
     wId: string,
     requestedGroupIds?: string[],
     requestedRole?: RoleType
-  ): Promise<{
-    workspaceAuth: Authenticator;
-  }> {
+  ): Promise<Authenticator> {
     const [workspace, keyWorkspace] = await Promise.all([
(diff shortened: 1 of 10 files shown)


front/lib/auth.ts
```diff
@@ -918,7 +918,6 @@ export class Authenticator {
     requestedRole?: RoleType
   ): Promise<{
     workspaceAuth: Authenticator;
-    keyAuth: Authenticator;
```
fabiencelier, line 999: Note: this could now return `Promise<Authenticator>` directly
can do it in a follow up PR if you think it's desirable

flvndvd: Yeah probably nice to simplify return type here.

flvndvd (approved):

spolu (approved):
LGTM

PR #31451, opened 2026-08-31
Add: Poke auth via CloudFlare.

Fraggle:
## Description

- Authenticate Poke requests with Cloudflare Access JWTs from `Cf-Access-Jwt-Assertion` or `CF_Authorization`, validating the issuer, audience, signature, and Dust-internal email. Keep WorkOS super-user sessions as the fallback.
- Add `Authenticator.fromDustSuperUser()` and `toPokeUserJSON()` so Poke routes, audit logs, and plugin runs work without a provisioned Dust user. Harden the local `front-api` restart loop and expose port `3010` for Poke.

## Tests

Added Vitest coverage for Cloudflare Access configuration, JWT claim validation and failures, JWKS reuse, and synthetic Poke super-user identities. Updated test scripts to set `NODE_ENV=test`.

## Risk

Medium security-sensitive risk. A misconfigured Cloudflare Access issuer, audience, or internal-email policy can block Poke access; an authentication regression could expose privileged routes. JWTs require a valid signature, issuer, audience, and Dust-internal identity. Unset the Cloudflare Access variables to return to the WorkOS session path.

## Deploy Plan

Configure `CLOUDFLARE_ACCESS_TEAM_DOMAIN` and `CLOUDFLARE_ACCESS_AUD` on `front-api`, and configure the Cloudflare Access application to protect Poke and inject the assertion header or authorization cookie. Deploy `front-api` and `front` together. No SQL migration or repository infrastructure change is required; rebuild the development container to use the exposed Poke port `3010`.


Fraggle:

- #31451 ◀

zmarouf (approved):

front-api/routes/poke/plugins/index.ts
```diff
@@ -32,8 +32,11 @@ app.get(
 
     let auth = ctx.get("auth");
     if (workspaceId) {
-      const session = ctx.get("session");
-      auth = await Authenticator.fromSuperUserSession(session, workspaceId);
+      auth = await Authenticator.fromDustSuperUser({
```
flvndvd, line 35: Seeing all those if, IMHO this should be handled by the poke mixin. This made sense in the past as we had shared memberships between app and poke, but as we split them, poke users have no default workspace anymore.

front/lib/api/poke/cloudflare_access.ts
```diff
@@ -0,0 +1,101 @@
...
+    const { payload } = await jwtVerify(
+      token,
+      getJwks(accessConfig.teamDomain),
+      {
+        issuer: accessConfig.teamDomain,
+        audience: accessConfig.aud,
+      }
+    );
+
+    const email =
```
flvndvd, line 71: Can we use zod here please?

flvndvd (approved):
Are we sure we cover all call sites here? Wondering if part such as `emitAuditLogEvent` are not being left on the side without using the right identity?

front/lib/auth.ts
```diff
@@ -102,6 +107,12 @@ export type AuthMethodType =
   | "sandbox_token"
   | "internal";
 
+/** Identity used by poke when there is no provisioned Dust user (e.g. Cloudflare Access). */
+export type PokeIdentity = {
```
flvndvd: `nit` but speaking identity, we usually go with principal, e.g. `PokePrincipal`

front-api/routes/poke/plugins/index.ts
Fraggle: As seen IRL, the wID is coming from the post body.
Will TAL (another PR to keep it short) to automatically do it based on poke/.../[wId]/... in the path.

front/lib/auth.ts
Fraggle: This + emitAuditLogEvent and I merge

front-api/routes/poke/plugins/index.ts
Fraggle: FYI, it's actually already the case, we wrap with withPokeWorkspace

PR #31452, opened 2026-08-31
fix: invalidate relocation caches before resume

flvndvd:
## Description

This PR fixes workspace relocation occasionally loading a stale subscription and treating the workspace as FREE.
When resuming in the destination cell, we now clear the workspace and subscription caches before creating the authenticator, ensuring it loads the relocated data from the destination database. As this write goes through the raw SQL queries, it currently by passes most of the resource logic. Only those two are required, as others are not loaded until workspace is resumed.


davidebbo (approved):
LGTM

id13:
LGTM

PR #31453, opened 2026-08-31
Enabled workspace analytics for analyst global agent

flvndvd:
## Description

This PR enables the workspace analytics skill on global agent `@.analyst` while still keeping it available as a regular standalone skill that can be pinned in the input bar.

### DIFF
--- a/front/lib/api/actions/servers/workspace_analytics/metadata.ts
+++ b/front/lib/api/actions/servers/workspace_analytics/metadata.ts
@@ -128,6 +128,7 @@ export const WORKSPACE_ANALYTICS_TOOLS_METADATA = [
       "most active. Admin-only.",
     schema: getTopAgentsSchema,
     stake: "never_ask",
+    eager: true,
     displayLabels: {
       running: "Retrieving top agents",
       done: "Retrieved top agents",
@@ -145,6 +146,7 @@ export const WORKSPACE_ANALYTICS_TOOLS_METADATA = [
       "are, rank members by usage, or find your top contributors. Admin-only.",
     schema: getTopUsersSchema,
     stake: "never_ask",
+    eager: true,
     displayLabels: {
       running: "Retrieving top users",
       done: "Retrieved top users",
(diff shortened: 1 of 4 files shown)


front/lib/api/actions/servers/workspace_analytics/metadata.ts
```diff
@@ -128,6 +128,7 @@ export const WORKSPACE_ANALYTICS_TOOLS_METADATA = [
       "most active. Admin-only.",
     schema: getTopAgentsSchema,
     stake: "never_ask",
+    eager: true,
```
flvndvd, line 131: Needed so they are exposed directly when "enabled" still stripped away when adding though regular skill use.

front/lib/resources/skill/code_defined/global/workspace_analytics.ts
```diff
@@ -70,4 +71,10 @@ export const workspaceAnalyticsSkill = {
     }
     return !isWorkspaceAnalyticsEnabled(auth.getNonNullableWorkspace());
   },
+  getAutoEnabledOrEquippedForAgentLoop: ({ agentConfiguration }) => {
+    return agentConfiguration.scope === "global" &&
+      agentConfiguration.sId === GLOBAL_AGENTS_SID.ANALYST
+      ? "enabled"
+      : "equipped";
```
aubin-tchoi: I don't think we want it equipped for all agents other than analyst, we probably need a new `null` case here

flvndvd: Oh yeah good catch, my read was wrong here. Let me fix.

flvndvd: Updated.

aubin-tchoi (approved):
LGTM

front/lib/resources/skill/code_defined/global/workspace_analytics.ts
fabiencelier: does it still mean `@dust` can enable it ?

flvndvd: Yes

PR #31455, opened 2026-08-31
Expose poke port 3010 in Cursor/dev and fix ES index init under set -e

Fraggle:
## Description

Expose `front-poke` on port `3010` in Cursor Cloud and the local `docker-run.sh` environment. Fix Elasticsearch index initialization in `init-elasticsearch-indices.sh` so `set -e` does not abort before the script captures and handles the “already exists” status.

## Tests

Reviewed the shell control flow for both index creation paths. No automated tests cover these development scripts or Cursor configuration.

## Risk

Low. The changes affect development tooling only and have no production runtime impact. Rollback is a simple revert.

## Deploy Plan

No production deployment or SQL migration. Rebuild the shared development image and refresh the Cursor Cloud snapshot so the new port mapping and script behavior take effect.

### DIFF
--- a/.cursor/environment.json
+++ b/.cursor/environment.json
@@ -15,6 +15,7 @@
   ],
   "ports": [
     { "name": "front-api", "port": 3000 },
+    { "name": "front-poke", "port": 3010 },
     { "name": "front-spa", "port": 3011 },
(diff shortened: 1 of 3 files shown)


zmarouf (commented):
LGTM

PR #31457, opened 2026-08-31
Add GLM-5.3 Flash on Fireworks

fontanierh:
## Description

Add GLM-5.3 Flash through the Fireworks model stack. Dust exposes 256k context and 64k output, with rollout gated by the existing Fireworks preview flag.

## Tests

Added model configuration and endpoint contract coverage. Ran the 46-case live Fireworks suite and a live image-input probe.

## Risk

Limited to US Fireworks traffic and preview workspaces. Named forced tools stay disabled because the deployment emits an incompatible finish reason.

## Deploy Plan

Standard deploy. Enable `fireworks_new_model_feature` for preview workspaces.


aubin-tchoi (approved):
LGTM

PR #31460, opened 2026-08-31
Restyle notifications as an inverted, stacking toast

pinotalexandre:
Reworks `Notification` — the sonner-backed toast — around the [`<Toast.Sonner>`](https://www.figma.com/design/R7vT7SUjyTSosoiTOABonk/Sparkle-Components?node-id=1818-9734) spec, then follows Radja's feedback away from it on colour.

One file: `sparkle/src/components/Notification.tsx`. `Notification.stories.tsx` and the Storybook config are untouched.

## Colour

The Figma node specced an inverted `bg-primary` surface. Radja's read: that works in light mode, but in dark mode a large near-white block is too much, and the common pattern is a green/orange/red surface per outcome with lighter tints in light mode.

So each type now takes the surface and foreground of its **`ContentMessage` counterpart, verbatim** — success/green, error/red, warning/orange, info/blue, hello/neutral. Reusing that mapping rather than inventing one keeps the two status surfaces from drifting apart.

Contrast measured on the rendered output, not estimated: titles 6.9–15.1 in light and 13.2–19.0 in dark; descriptions 4.66–8.04 and 8.6–12.1. Everything clears 4.5:1, though the info and warning descriptions have little room in light mode. No axe violations under `VITE_A11Y_STRICT=1`.

**This leaves the Figma spec** — worth a design confirmation before it ships.

## Stack

Queued toasts collapse into a stack with the geometry sonner uses on [its own reference site](https://sonner.emilkowal.ski): three visible, each peeking 14px above the front one at a further 5% scale down, expanding on hover. Sonner only hides stacked toasts' content for its own *styled* toasts and ours are `unstyled`, so the card fades its own children out via the `[data-expanded=false][data-front=false]` attributes. `visibleToasts` goes 5 → 3 to match sonner's default.

## Motion

The card was running `animate-in fade-in-0 zoom-in-95` (200ms, ease-out quint) *on top of* sonner's own entrance transition (400ms, `ease`) — computed styles showed the `li` at `opacity: 0` while the card inside it was also at `opacity: 0, scale(.95)`, so every toast faded twice through the product of two curves. Dropping the keyframe leaves sonner's `translateY(100%) → 0` as the single entrance, and removes a keyframe from an element mounted in rapid succession (keyframes restart from zero when interrupted; transitions retarget).

The stacked-content fade uses sonner's own duration and curve (`duration-400 ease-enter`) so it stays in step with the transform it accompanies, and no longer opts out under `prefers-reduced-motion`: it is opacity only, with no movement to remove.

## Spacing and type

Padding 8px → 12px, icon/text gap 4px → 8px, 2px between title and description, width 246px → 264px so the extra padding comes out of the card rather than the text column. Title and description move to the `label-sm` / `copy-xs` semantic utilities. The action button uses the `outline` variant, which inverts correctly on both sides with no override.

## Checks

`tsgo --noEmit`, biome, and the Notification story tests pass, including under `VITE_A11Y_STRICT=1`.

Known follow-up, not addressed here: the toast has no elevation shadow, only its border. `Popover`, `Dialog` and `Dropdown` do carry `shadow-md`/`shadow-lg` but with no dark handling, so their shadows are invisible in dark mode.


radjakahoul (commented):
lgtm

radjakahoul (approved):
lgtm

PR #31461, opened 2026-08-31
feat: use spaces when creating api keys

matteotrab:
## Description

This PR lets API keys be scoped to restricted spaces and pods instead of selecting groups directly.

When a space used manual management mode, we previously displayed the provisioned and regular_manual groups the space relied on, which was wrong. Resolving through the space first means the key is always linked to the space's regular_auto groups.

It also ensures regular_auto groups are always reached through their space rather than fetched directly, which is needed for the follow-up permission refactor of group_resource.

## Tests

Added tests.

## Risks
Low.

## Deploy Plan

Standard deployment.

### DIFF
--- a/front-api/routes/w/[wId]/keys/index.ts
+++ b/front-api/routes/w/[wId]/keys/index.ts
@@ -3,6 +3,7 @@ import {
   emitAuditLogEvent,
   getAuditLogContext,
 } from "@app/lib/api/audit/workos_audit";
+import { listKeyScopableGroups } from "@app/lib/api/keys/scopable_groups";
 import {
   MAX_API_KEY_SPEND_LIMIT_AWU_CREDITS,
   MIN_API_KEY_SPEND_LIMIT_AWU_CREDITS,
@@ -33,6 +34,8 @@ const MAX_API_KEY_CREATION_PER_DAY = 30;
 const CreateKeyPostBodySchema = z.object({
   name: z.string(),
   space_ids: z.array(z.string()).optional(),
+  group_ids: z.array(z.string()).optional(),
+  group_id: z.string().optional(),
   monthly_cap_micro_usd: z.number().nullish(),
   // Per-key credit cap in AWU credits (credit-priced plans only). null/omitted
   // = unlimited.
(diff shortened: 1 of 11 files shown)


matteotrab:

- #31461 ◀
    - #31464

front-api/routes/w/[wId]/keys/index.ts
```diff
@@ -154,7 +158,8 @@ app.post(
       });
     }
 
-    // Resolve groups: prefer group_ids (new), fall back to group_id (retro-compatibility).
+    // A key always carries the workspace global group, so it can reach everything every workspace
+    // member can reach; the spaces it is scoped to add to that.
```
Comment (philipperolet) [inline front-api/routes/w/[wId]/keys/index.ts]: nit:
```suggestion
    // member can reach; in addition, it can access the spaces it is scoped to
```

front-api/routes/w/[wId]/keys/index.ts
```diff
@@ -154,7 +158,8 @@ app.post(
       });
     }
 
-    // Resolve groups: prefer group_ids (new), fall back to group_id (retro-compatibility).
```
Comment (philipperolet) [inline front-api/routes/w/[wId]/keys/index.ts:157]: IIRC now we prefer spaces, and we have 2 legacy fields for retro compatibility?
if it's that, we might want a follow-up to remove at least one of them?

front-api/routes/w/[wId]/keys/index.ts
```diff
@@ -168,6 +173,38 @@ app.post(
     const globalGroup = globalGroupRes.value;
 
     const resolvedGroups: GroupResource[] = [globalGroup];
+    const requestedSpaceIds = [...new Set(space_ids ?? [])];
```
Comment (philipperolet) [inline front-api/routes/w/[wId]/keys/index.ts]: logic is becoming too much for a handler, could we extract in lib/api? (cf BACK16 rule)

philipperolet (commented):
Thanks 🙏 
Thinking about it, keys that are only "user", when scoped to a pod, should likely not get the pod editor group which grants admin access power (editing pod admins, etc.)
Left a few other coms

front-api/routes/w/[wId]/keys/index.ts
Comment (philipperolet) [reply front-api/routes/w/[wId]/keys/index.ts:157]: Ooh just saw the follow-up PR in the stack 👍

philipperolet (approved):
LGTM 👍

PR #31463, opened 2026-08-31
[admin governance] migrate remaining builder role

ykmsd:
## Description
This PR is to migrate the remaining builder role users.

- My previous script skipped the users with the future date of startAt. I will re-run the previous one for the users who started between my migration and today (since they need to invalidate the cache)
- The new script is updating all the past and future builders to users (from what I understood, the future users don't need to update workOS role)

It's around 6,700 users in US and 2,700 in EU.

### DIFF
--- a/front/migrations/20260831_migrate_remaining_builder_role_to_user.ts
+++ b/front/migrations/20260831_migrate_remaining_builder_role_to_user.ts
@@ -1,0 +1,62 @@
+import { Op } from "sequelize";
+
+import { MembershipModel } from "@app/lib/resources/storage/models/membership";
+import { makeScript } from "@app/scripts/helpers";
+
+/**
+ * The previous migration only handled currently-active `builder` memberships.
+ * This script flips the remaining rows: not-yet-started ones (`startAt` in the
+ * future) and already-ended ones (`endAt` in the past).
+ *
+ * None of these rows are currently active, so their cached role is never
+ * `builder` and no cache invalidation is needed.
+ */
+
+makeScript({}, async ({ execute }, logger) => {
+  const now = new Date();
+
+  const where = {
+    role: "builder" as const,
+    [Op.or]: [{ startAt: { [Op.gt]: now } }, { endAt: { [Op.lt]: now } }],
+  };
+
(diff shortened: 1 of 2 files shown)


front/migrations/20260831_migrate_remaining_builder_role_to_user.ts
```diff
@@ -0,0 +1,62 @@
...
+    logger.info(
+      { total: builderMemberships.length },
+      "Remaining builder -> user migration dry run complete"
+    );
+    return;
+  }
+
+  const [migrated] = await MembershipModel.update(
+    { role: "user" },
+    { where: { id: { [Op.in]: builderMemberships.map((m) => m.id) } } }
```
fabiencelier: can you batch by ~100 ids to avoid doing a big update at once ?

front/migrations/20260831_migrate_remaining_builder_role_to_user.ts
```diff
@@ -0,0 +1,62 @@
...
+/**
+ * The previous migration only handled currently-active `builder` memberships.
+ * This script flips the remaining rows: not-yet-started ones (`startAt` in the
+ * future) and already-ended ones (`endAt` in the past).
+ *
+ * None of these rows are currently active, so their cached role is never
+ * `builder` and no cache invalidation is needed.
+ */
+
+makeScript({}, async ({ execute }, logger) => {
```
fabiencelier, line 13: I'm not sure if it is mandatory but we generally use `runOnAllWorkspaces` util to iterate and on all workspaces and apply the workspaceId filter in all the DB queries so we don't merge query the data from different workspaces at once.

If you do that, no need to batch as I think update per workspace is small enough

ykmsd: okie okie

ykmsd: discussed IRL, TIL it's better to not mix update with different workspaces and we prefer to use runOnAllWorkspaces 🫡

matteotrab (approved):

PR #31464, opened 2026-08-31
feat: drop group_ids and group_id when creating api keys

matteotrab:
## Description

This PR removes the legacy `group_ids` and `group_id` fields from the API key creation endpoint.

## Tests
Manually.

## Risks

Low risk. Clients still sending the removed fields may no longer have those fields honored, so this should be deployed only after clients using the previous bundle have cycled out.

## Deploy Plan

Deploy after confirming that clients on the previous bundle have cycled out.


matteotrab:

- #31461
    - #31464 ◀
        - #31563
            - #31565

philipperolet (approved):
👍

PR #31465, opened 2026-08-31
feat: close the user credit pool to non credit-priced plans

adrsimon:
## Description

Choosing user pool on new triggers should not be available in the legacy plan anymore. `availableTriggerExecutionModes` now decides which pools a caller may pick, from the plan, the `legacy_trigger_limits` flag and the `use_workspace_pool` grant, and both `TriggerResource` and the new `useTriggerExecutionModes` hook read from it.

A workspace with `legacy_trigger_limits` keeps the legacy trigger limits and can still charge automations to personal credits. Use it to hold back a legacy workspace that is not ready to move.

In the trigger sheets the selector disables the pools the user cannot pick and says why, and Save stays disabled while the selected pool is unavailable.

Existing `user_pool` triggers on legacy plans keep running and stay editable. They cannot move back once they leave.

## Tests

`trigger_resource.test.ts` covers the legacy-plan transitions: creation on `user_pool` refused, allowed again with `legacy_trigger_limits`, an existing `user_pool` trigger moving to the workspace pool, no move back. `front-api/routes/w/[wId]/triggers/index.test.ts` covers a legacy-plan PATCH that edits other fields while leaving the pool alone.

The `plan: "creditPriced"` additions elsewhere are tests that build a trigger through `TriggerFactory` and would otherwise fail. Each was added after it failed without it.

## Risk

On a non credit-priced workspace only admins hold `use_workspace_pool` by default, so managers and members cannot create any trigger. Enable `legacy_trigger_limits` on the workspace to unblock them.

The form still opens on personal credits, so a legacy-plan admin has to switch the pool before saving.

Existing legacy-plan `user_pool` triggers keep being charged as they are today. No migration moves them.

Revert to roll back, no schema change.

## Deploy Plan

Deploy `front`.

### DIFF
--- a/front-api/routes/v1/w/[wId]/triggers/hooks/[webhookSourceId]/[webhookSourceUrlSecret]/index.test.ts
+++ b/front-api/routes/v1/w/[wId]/triggers/hooks/[webhookSourceId]/[webhookSourceUrlSecret]/index.test.ts
@@ -251,7 +251,9 @@ describe("POST /api/v1/w/[wId]/triggers/hooks/[webhookSourceId]/[webhookSourceUr
   });
 
   it("stores payload when a matched trigger includes payload", async () => {
-    const { workspace } = await createPublicApiMockRequest();
+    const { workspace } = await createPublicApiMockRequest({
+      plan: "creditPriced",
+    });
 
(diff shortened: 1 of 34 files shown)


id13 (approved):
LGTM

PR #31467, opened 2026-08-31
[front] Fix Pod group cleanup during deletion

davidebbo:
**Issue**: Trying to recreate a deleted pod (re-using same name) was failing with "Error: Space with that name already exists". I reproed it multiple times, and then multiple times verified that it worked after the fix.

## Description

- Propagate the Space deletion transaction to the remaining group-grants lookup.
- Ensure a permanently deleted Pod's auto-group is removed, allowing the Pod name to be reused.
- Keep archive behavior unchanged.

Fixes dust-tt/tasks#10245.

Existing orphaned groups created before this fix still require targeted cleanup.

## Tests

- NODE_ENV=test npm run test -- lib/resources/space_resource.test.ts lib/resources/group_permission_resource.test.ts lib/api/spaces.test.ts: 163 passed.
- Regression test confirmed failing before the fix and passing after it.
- Pre-commit Biome and front-typecheck hooks passed.
- Full front suite: 9,033 passed and 6 unrelated Frames tests failed. The same six failures reproduce on unchanged origin/main because sandbox-function fixtures omit the required spaceId.

## Risk

Low. The transaction parameter is optional for existing callers and is only supplied by the Space deletion path. The change can be rolled back by reverting these commits.

## Deploy Plan

Deploy front normally. Clean up the existing orphaned auto-group for the affected EQT workspace separately so the historical Pod name can be reused.

### DIFF
--- a/front/lib/resources/group_permission_resource.ts
+++ b/front/lib/resources/group_permission_resource.ts
@@ -860,13 +860,15 @@ export class GroupPermissionResource extends BaseResource<GroupPermissionModel> 
   // Every grant held by one group, across resource types and instances.
   static async listForGroup(
     auth: Authenticator,
-    group: GroupResource
+    group: GroupResource,
+    transaction?: Transaction
   ): Promise<GroupPermissionResource[]> {
     const rows = await GroupPermissionModel.findAll({
       where: {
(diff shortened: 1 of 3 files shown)


tdraier (approved):
🙏

PR #31468, opened 2026-08-31
Normalize relocation IDs in destination region

flvndvd:
## Description

During a workspace relocation, we copy database rows from one region to another. The problem is that database IDs are local to a region: the same user can already exist in both regions under different numeric IDs, and the same is true for plans. For example, a user may have ID 123 in the source and ID 456 in the destination. If we copy a membership containing `userId: 123`, it would refer to the source ID, which may not exist in the destination. A subscription can break in exactly the same way when its `planId` comes from the source region.

We already had special handling for users, but the approach was awkward. The destination detected user conflicts, then the workflow carried that information back through Temporal so the source side could rewrite the exported data. This meant the source needed to know about IDs chosen by the destination, and the workflow had to transport implementation details such as mapping files and user-reference columns. Adding plans would have required repeating and expanding that plumbing, making the abstraction increasingly leaky.

The new approach keeps the source simple and independent. The source exports rows exactly as they exist, without knowing anything about the destination database. Once the data reaches the destination, the destination identifies equivalent existing records using stable identities: users are matched by their WorkOS user ID, and plans are matched by their code. It then builds a mapping from each source database ID to the corresponding destination-local ID.

Immediately before executing the relocation SQL, the destination applies those mappings to the incoming statement parameters. References such as `memberships.userId` are changed to the destination user ID, while `subscriptions.planId` is changed to the destination plan ID. The source IDs therefore never reach the destination database in those reference columns.


flvndvd:
PMRR.

PR #31469, opened 2026-08-31
[front] fix: show configured model in agent builder picker

aubin-tchoi:
## Description

Context [here](https://dust4ai.slack.com/archives/C0BJ3T5SA0L/p1788187137875059).

The agent builder could sometimes show "Standard" (it's the fallback option) instead of the configured model when that model was not present in the current viewer model list, notably for admins who are not agent editors.

Showing "Standard" is misleading, it's different from "unknown".

This PR fixes this by always showing the actual model stored in the builder form. Also fixes the short blink where we see "Standard" for a split second before hydrating with the actual configured model.

Can be tested [here](https://app.dust.tt/w/0ec9852c2f/builder/agents/Hzyy2eXV5k) (prod) vs [here](https://pr-31469-merge--app.preview.dust.tt/w/0ec9852c2f/builder/agents/Hzyy2eXV5k) (preview).

## Tests

- Tested locally + preview.

## Risk

- Low.

## Deploy Plan

- Deploy front.


front/components/agent_builder/instructions/AdvancedSettings.tsx
```diff
@@ -153,36 +153,15 @@ export function AdvancedSettings() {
     name: "generationSettings",
   });
 
-  const [modelSelection, setModelSelection] = React.useState<
```
aubin-tchoi, line 156: this just removes an extra state (already in the form), which fixes the blink issue

front/components/agent_builder/instructions/AdvancedSettings.tsx
```diff
@@ -197,20 +176,16 @@ export function AdvancedSettings() {
         onOpenChange={setIsResponseFormatDialogOpen}
       />
       <ModelPicker
-        // Set these 2 as we are in the context of the agent builder, not a conversation
         agentId={null}
-        agentModel={null}
```
aubin-tchoi, line 202: that's the actual fix

aubin-tchoi: will cleanup a bit, we don't actually need all of these props

front/components/agent_builder/instructions/AdvancedSettings.tsx
```diff
@@ -153,36 +153,15 @@ export function AdvancedSettings() {
...
-    }
-  }, [
-    generationSettingsField.value.modelSettings?.modelId,
-    generationSettingsField.value.modelSettings?.providerId,
-    generationSettingsField.value.reasoningEffort,
-    modelSelection?.modelId,
-    modelSelection?.providerId,
-    modelSelection?.reasoningEffort,
-  ]);
+  const agentModel: AgentModelConfigurationType | null = generationSettingsField
```
PopDaph: Could we useMemo this? 
Also, feel free to ignore but it could be easier to read if it was not inlined. 


```
let agentModel: AgentModelConfigurationType | null = null;
const agentData = generationSettingsField.value;

if (agentData.modelSettings) {
   agentData = // blablabla how cute is soupinou
}
```

front/components/agent_builder/instructions/AdvancedSettings.tsx
```diff
@@ -197,20 +176,16 @@ export function AdvancedSettings() {
...
+        lastRequestedModel={null}
         owner={owner}
         buttonSize="sm"
         buttonVariant="outline"
         showLabel={true}
         side="top"
         disabled={false}
         onSelectionChange={(newModelSelection) => {
-          // Keep both in sync to avoid extra re-renders
-          setModelSelection(newModelSelection);
```
PopDaph, line 213: As someone traumatized by agent builder dark bugs, have you checked why we had that extra state initially? Was it "never needed complexity" or "necessary dark magic"?

front/components/agent_builder/instructions/AdvancedSettings.tsx
aubin-tchoi: We can indeed memo here to get a stable ref
Will also address the refactor!

front/components/agent_builder/instructions/AdvancedSettings.tsx
aubin-tchoi: From the looks of if, it was used for something else that was removed since (will double check)

front/components/agent_builder/instructions/AdvancedSettings.tsx
aubin-tchoi: fixed

front/components/agent_builder/instructions/AdvancedSettings.tsx
aubin-tchoi: was introduced in https://github.com/dust-tt/dust/commit/6b7606c7a94f548a231669a4513b5139a0a46264, it was really an additional state that copies exactly what was in RHF 🤷🏻

PopDaph (approved):
Thank you 🙏🏻

PR #31470, opened 2026-08-31
feat(poke): add the degraded models poke API

Nils-Fedrigo:
## Description

Third of the degraded-model stack (on top of #31137): the poke API for reading and setting the degraded set. Split out of #31145, which is now the Kill page UI only.

- **`GET /api/poke/degraded_models`** returns the degradable catalog plus the endpoints currently degraded.
- **`POST /api/poke/degraded_models`** degrades or restores the endpoints it names — `{ endpoints: [{ modelId, providerId, host, degraded }] }` — and leaves every other endpoint untouched. It is a per-endpoint update rather than a whole-set replace on purpose: two people working the same incident from their own poke tabs would otherwise clobber each other, since a whole-set write computed from a stale read silently restores whatever the other person just degraded.
- The **degradable catalog** is derived from the router's `DUST_STREAM_ENDPOINTS`, de-duplicated across regions, rather than from `SUPPORTED_MODEL_CONFIGS`. So an operator can only degrade an endpoint that actually serves traffic, and what reaches the database is always a catalog entry with its narrow types rather than strings off the wire. An endpoint that is not in the catalog is a 400, not a silent no-op.
- `listDegradableEndpoints` lives in `front/lib/api/poke/` rather than in the route, because the dialog in #31145 needs its type: the endpoint registries pull in provider SDKs and cannot be imported into the browser bundle, so the catalog has to come over the wire.

Nothing calls this yet — the UI lands in #31145.

## Tests

`npx tsgo --noEmit` clean on `front` and `front-api`.

Seven functional tests in `front-api/routes/poke/degraded_models.test.ts`: the catalog shape, degrading and restoring **only** the named endpoints (the property the whole per-endpoint design exists for), repeat requests being no-ops, an endpoint named twice in one request settling on its last requested state, a 400 for an endpoint that is not in the catalog, and the two super-user checks.

One thing worth not undoing: they reset the table through the resource rather than through the model. `model_degradations` is global and written in its own transaction, so it escapes vitest's per-test rollback and has to be reset explicitly — but a bare model call joins the CLS transaction that `front/vite.setup.ts` wraps each test in, and the row locks it takes there deadlock against the resource's own transaction. The comment in the file says so.

## Risk

Low. Poke-only surface, super-user gated, and no caller until #31145. The rows it writes have no reader until #31137 ships.

## Deploy Plan

No migration of its own — #31136 creates the table. Plain code deploy, merged in stack order (#31136 → #31137 → this one → #31145).

1. Deploy `front` / `front-api`.
2. Smoke-test with a super-user session: `GET /api/poke/degraded_models` returns a non-empty `degradableEndpoints` and an empty `degradedEndpoints`.
3. Nothing to roll back beyond a revert; no rows are written until someone uses the UI in #31145.


front/lib/api/poke/degraded_models.ts
```diff
@@ -0,0 +1,90 @@
...
+import { Err, Ok } from "@app/types/shared/result";
+import { z } from "zod";
+
+export type DegradableModelEndpointType = DegradedModelEndpointType & {
+  displayName: string;
+};
+
+export type GetDegradedModelsResponseBody = {
+  degradableEndpoints: DegradableModelEndpointType[];
+  degradedEndpoints: DegradedModelEndpointType[];
```
frankaloia: Should we simplify this and just have one endpointtype that returns status?

frankaloia (approved):
LGTM, will trust you to manage rebases

PR #31476, opened 2026-08-31
fix(conversation): avoid duplicate final answer in activity steps

davidebbo:
## Description

Fixes dust-tt/tasks#10241.

Prevent the final answer from appearing twice when an identical text fragment was already included in the Work timeline. The renderer now removes only content activity steps that match the final body after trimming, while preserving reasoning, actions, and distinct intermediate text.

## Tests

- Added a regression test for identical intermediate and final text.
- NODE_ENV=test npm test: 701 files passed, 9,048 tests passed.
- npx tsgo --noEmit.
- Targeted Biome check.
- Sparkle Storybook: 115 files passed, 439 tests passed.

## Risk

Low. The change only filters exact duplicate content from the rendered activity-step list. It is safe to roll back.

## Deploy Plan

Deploy through the standard front release process. No migration or configuration change is required.

### DIFF
--- a/front/lib/api/assistant/activity_steps.test.ts
+++ b/front/lib/api/assistant/activity_steps.test.ts
@@ -88,6 +88,50 @@ describe("renderAgentMessageContentView", () => {
     });
   });
 
+  it("does not repeat the final answer in the activity steps", async () => {
+    expect(
+      await render([
+        {
+          step: 0,
+          content: {
+            type: "reasoning",
+            value: {
+              reasoning: "I need the user to confirm the application",
+              metadata: "",
+              tokens: 0,
+              provider: "anthropic",
+            },
+          },
+        },
+        {
+          step: 0,
+          content: {
+            type: "text_content",
+            value: "Please confirm which application this is for.",
+          },
+        },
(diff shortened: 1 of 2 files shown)


PopDaph:
I'm a bit bearish to merge this without understanding the duplication 🙏🏻 
 Is it content that was render once as thinking and then the same as the generation tokens?

davidebbo:
> I'm a bit bearish to merge this without understanding the duplication 🙏🏻
> Is it content that was render once as thinking and then the same as the generation tokens?

Yep, I hear you. I'm not able to repro this myself, making it hard to root cause. Since it's a render issue, I'm not sure we can get the info we need from the Convo history.

This fix is more about "something bad happened, we don't know what, let's try to limit its damage", with the idea that it probably can't do harm. Ideally, we'd root cause, but not confident that I can do this.

PopDaph:
I'll take some time to dig 🙏🏻 
You can close this one and assign me the runner card 🙏🏻

davidebbo:
Let's look for a fix that targets the true root cause, per discussion above.

PR #31480, opened 2026-08-31
Move Frames within one mount

fontanierh:
## Description

- add the bounded backend operation for moving a registered Frames v2 source folder within the same conversation or Pod mount
- preserve the stable FileResource identity, publication metadata, sharing state, and source ownership metadata
- reject nested registered files and occupied destinations before copying
- raw-list at most 1,024 GCS objects / 100 MiB so hidden, processed, and placeholder objects are included
- copy objects, commit the FileResource path, best-effort delete the source, and emit the existing file.moved audit event

This is deliberately non-crash-safe. It does not include reservations, recovery, generation fencing, rollback, or renewable locks. The HTTP route is split into a child PR.

## Tests

Tested the move operation locally, including hidden objects, placeholders, size bounds, nested registered Frames, copy failures, commit failures, and source cleanup failures.

## Risk

The operation has no production caller until the child HTTP PR merges. Once called, a crash or typed failure can leave partial destination objects. The FileResource path remains authoritative.

## Deploy Plan

### DIFF
--- a/front/lib/api/files/gcs_mount/files.ts
+++ b/front/lib/api/files/gcs_mount/files.ts
@@ -422,7 +422,7 @@ async function moveGCSMountFile({
   return new Ok(undefined);
 }
 
-async function emitGCSMountFileMovedAuditLog(
+export async function emitGCSMountFileMovedAuditLog(
   auth: Authenticator,
   scope: GCSMountPoint,
   {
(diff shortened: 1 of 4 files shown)


fontanierh:
Superseded by the focused core PR #31510. Extended edge coverage is in #31511, while the existing HTTP and GCS generation children are now rebased as #31494 and #31495.

PR #31481, opened 2026-08-31
LLM stream errors: shared transport classifier (OpenAI/Fireworks)

frankaloia:
## Summary

Follow-up to #31149. Adds the shared transport classifier and wires it into the Fireworks chat-completions adapter.

- Classify transport errors from stable `code` / `name` / cause-chain values, never free-form English.
- Reuse `classifyHttpStatus` / `buildHttpStatusErrorEvent` for typed HTTP. No second status table.
- Treat leftover 4xx (`413`, `418`) as `unknown`, not Dust.
- Statusless SSE `APIError` stays provider + non-retryable `unknown_error`.
- Fireworks `terminated` + `UND_ERR_SOCKET` becomes retryable `network_error` with `errorSource: unknown`.

Remaining adapters move in #31482.

## Tests

Unit tests

## Risk

Low. Streaming happy-path events are unchanged. Retryability still comes from error type, not `errorSource`.

## Deploy Plan

1. Deploy front


front/lib/model_constructors/sdk/openai_shared/stream_error.ts
```diff
@@ -0,0 +1,93 @@
...
+    metadata,
+    error,
+    classifyStreamError({ error, providerName })
+  );
+}
+
+// OpenAI-family adapters serve more than one host. Attribute the error to the
+// host actually in use so a Fireworks or xAI failure is not labelled "OpenAI".
+export function openaiProviderNameForHost(
+  host: EndpointMetadata["host"]
```
Nils-Fedrigo: maybe a good place to have an assertNever? 

also I'm not sure where this is used, maybe next PR

front/lib/model_constructors/utils/classify_stream_error.test.ts
Nils-Fedrigo, line 1: this test is a bit redundant with the one above no?

front/lib/model_constructors/utils/classify_stream_error.ts
```diff
@@ -0,0 +1,288 @@
...
+
+// Wraps a classification into the `ErrorEvent` every endpoint emits, attaching
+// the metadata and the original error for logging. Adapters classify with plain
+// control flow (see each `streamErrorToErrorEvent`) and pass the result here.
+export function classificationToErrorEvent(
+  metadata: EndpointMetadata,
+  error: unknown,
+  classification: StreamErrorClassification
+): ErrorEvent {
+  return buildErrorEvent({ ...classification, metadata, originalError: error });
```
Nils-Fedrigo: instead of having this adapter, directly return the good type?

Nils-Fedrigo (approved):
Nit : the logic feels a bit intricate to me, so it can probably be simplified a bit -- or I'm tired. 

Feels weird to have this 2-step logic with classificationToErrorEvent to classifyStreamError, for example

front/lib/model_constructors/sdk/openai_shared/stream_error.ts
frankaloia: Moved out of this PR — nothing here called it. The adapter PR will add it with an exhaustive `assertNever` when Responses/xAI actually need the host label.

front/lib/model_constructors/utils/classify_stream_error.test.ts
frankaloia: Yes. Folded the extra network-code cases into the first table, and merged the two unclassified / abort cases into `it.each`.

front/lib/model_constructors/utils/classify_stream_error.ts
frankaloia: Agreed — `classifyStreamError` now returns the `ErrorEvent` directly via `buildErrorEvent`. The classify-then-wrap helper is gone.

frankaloia:
> Nit : the logic feels a bit intricate to me, so it can probably be simplified a bit -- or I'm tired.
> 
> Feels weird to have this 2-step logic with classificationToErrorEvent to classifyStreamError, for example

@Nils-Fedrigo I simplified, and stripped it down to just the shared classifyStreamError component + the shared open AI specific component

PR #31482, opened 2026-08-31
LLM stream errors: migrate remaining adapters onto the shared classifier

frankaloia:
## Summary

Rebased directly onto `main` after #31149 and #31481 merged. This PR contains only the remaining adapter migrations:

- Anthropic preserves its provider-specific invalid-tool-JSON and file-download handling, while SDK aborts, timeouts, connections, and untyped transport errors use the shared classifier.
- Google preserves its existing typed `ApiError` handling, including the invalid-key 400 exception; only untyped transport errors move to the shared classifier.
- Mistral preserves typed HTTP handling and delegates untyped transport errors to the shared classifier.
- Fireworks completions calls the shared OpenAI-SDK error helper directly without a local wrapper.
- OpenAI Responses uses a small converter factory so OpenAI, Fireworks, and xAI endpoints explicitly provide their provider identity.
- OpenAI `response.failed` results are classified from stable in-band error codes rather than being treated as thrown transport errors.

The rebase drops duplicate HTTP-status logic and the older intermediate classification API.

## Behavior

- Retryability continues to be determined by error type, not `errorSource`.
- Ambiguous transport failures use `errorSource: "unknown"`.
- Anthropic malformed tool JSON remains a retryable `model_output_error` with unknown attribution.
- Anthropic file-download failures remain retryable Dust `server_error`s, so they do not count as provider outages.
- OpenAI content-policy failures are non-retryable `refusal_error`s with neutral/unknown attribution.
- Generic transport classification never matches free-form message text.

## Testing

- `cd front && npx tsgo --noEmit`
- Explicit Biome checks for all changed files
- `cd front && npm test -- lib/model_constructors/sdk/openai_completions/converters/output/utils.test.ts lib/model_constructors/sdk/openai_responses/converters/output/utils.test.ts lib/model_constructors/stream/clients/fireworks_responses.test.ts` (47 tests)
- Anthropic/OpenAI focused converter suite from the rebase (102 tests)

## Deploy Plan

1. Deploy front.

### DIFF
--- a/front/lib/model_constructors/sdk/anthropic_ai/converters/output/utils.ts
+++ b/front/lib/model_constructors/sdk/anthropic_ai/converters/output/utils.ts
@@ -1,7 +1,9 @@
 import {
   AnthropicError,
   APIConnectionError,
+  APIConnectionTimeoutError,
   APIError,
+  APIUserAbortError,
 } from "@anthropic-ai/sdk";
 import type { BetaMessageBatchResult } from "@anthropic-ai/sdk/resources/beta/messages/batches";
 import type {
@@ -45,12 +47,12 @@ import {
   buildHttpStatusErrorEvent,
   httpErrorMessage,
 } from "@app/lib/model_constructors/utils/classify_http_status";
+import { classifyStreamError } from "@app/lib/model_constructors/utils/classify_stream_error";
 import logger from "@app/logger/logger";
 import {
   assertNever,
   assertNeverAndIgnore,
 } from "@app/types/shared/utils/assert_never";
-import { normalizeError } from "@app/types/shared/utils/error_utils";
 import { isRecord } from "@app/types/shared/utils/general";
 import { safeParseJSON } from "@app/types/shared/utils/json_utils";
 
@@ -414,29 +416,18 @@ function isApiConnectionError(err: unknown): err is APIConnectionError {
   return err instanceof APIConnectionError;
 }
 
-function isApiError(err: unknown): err is APIError {
-  return err instanceof APIError;
+function isApiConnectionTimeoutError(
+  err: unknown
+): err is APIConnectionTimeoutError {
+  return err instanceof APIConnectionTimeoutError;
 }
 
-// A stream error classified into the categories we surface. `APIConnectionError`
-// is checked before `APIError` since the former extends the latter.
-type ClassifiedStreamError =
-  | { kind: "invalid_tool_json" }
-  | { kind: "connection"; error: APIConnectionError }
-  | { kind: "api"; error: APIError }
-  | { kind: "unknown" };
+function isApiUserAbortError(err: unknown): err is APIUserAbortError {
+  return err instanceof APIUserAbortError;
+}
 
-function classifyStreamError(error: unknown): ClassifiedStreamError {
-  if (getInvalidToolJsonMessage(error) !== null) {
-    return { kind: "invalid_tool_json" };
-  }
-  if (isApiConnectionError(error)) {
-    return { kind: "connection", error };
-  }
-  if (isApiError(error)) {
-    return { kind: "api", error };
-  }
-  return { kind: "unknown" };
+function isApiError(err: unknown): err is APIError {
+  return err instanceof APIError;
 }
 
 function apiErrorToErrorEvent(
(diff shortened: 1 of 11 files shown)


front/lib/model_constructors/sdk/anthropic_ai/converters/output/utils.ts
```diff
@@ -531,170 +434,80 @@ export function streamErrorToErrorEvent(
...
-      metadata,
-      type,
-      message: text,
-      originalError: error,
+  // Invalid tool-call JSON aborted the stream with no tool_use block to recover
+  // from. Surface a distinct, retryable type so the agent loop re-samples
+  // instead of treating it as a terminal invalid_request_error.
+  if (getInvalidToolJsonMessage(error) !== null) {
+    return classificationToErrorEvent(metadata, error, {
+      errorSource: "unknown",
```
Nils-Fedrigo, line 475: this is an "expected" error no? Perhaps we should classify it accordingly. 
Even this afternoon I got 2/ 20 wrong calls with A\ which have nothing to do with outages

front/lib/model_constructors/sdk/anthropic_ai/converters/output/utils.ts
```diff
@@ -531,170 +434,80 @@ export function streamErrorToErrorEvent(
...
-    return build(
-      "authentication_error",
-      `Authentication failed for Anthropic: ${message}`,
-      "dust"
-    );
+  // A file-download failure surfaces as an `APIError` but is a provider-side
+  // fetch failure, not a bad request.
+  if (isApiError(error) && isAnthropicFileDownloadError(error)) {
+    return classificationToErrorEvent(metadata, error, {
+      errorSource: "provider",
```
Nils-Fedrigo: should this count towards "provider outage" too?

front/lib/model_constructors/sdk/google_genai/converters/output/utils.ts
```diff
@@ -227,84 +233,8 @@ export function finishReasonToErrorEvent(
...
-      }
-      return buildErrorEvent({
-        errorSource: "provider",
-        metadata,
-        type: "unknown_error",
-        message: `Error from Google (${status}): ${error.message}`,
-        originalError: error,
-      });
-  }
+function isApiError(error: unknown): error is ApiError {
```
Nils-Fedrigo, line 267: mmm is this function not already existing above?

front/lib/model_constructors/sdk/anthropic_ai/converters/output/utils.ts
Nils-Fedrigo: i tought we said no

front/lib/model_constructors/sdk/google_genai/converters/output/utils.ts
```diff
@@ -313,16 +243,36 @@ export function streamErrorToErrorEvent(
   metadata: EndpointMetadata,
   error: unknown
 ): ErrorEvent {
-  if (error instanceof ApiError) {
-    return apiErrorToErrorEvent(metadata, error);
+  // Google returns a 400 (not a 401) for an invalid API key; disambiguate by
+  // message so it is attributed as an authentication failure.
```
Nils-Fedrigo: I thought we did not base on the message in the previous PR, was that for a particular reason?

front/lib/model_constructors/sdk/openai_responses/converters/output/utils.test.ts
```diff
@@ -159,6 +161,80 @@ function completedResponse(
...
+        }),
+      }),
+    ]);
+  });
+
+  it.each([
+    ["server_error", "server_error", "provider"],
+    ["rate_limit_exceeded", "rate_limit_error", "dust"],
+    ["invalid_prompt", "invalid_request_error", "dust"],
+    ["bio_policy", "refusal_error", "dust"],
```
Nils-Fedrigo: we should maybe have a third type of error "not important" or something?

Nils-Fedrigo: this is not really on us

front/lib/model_constructors/sdk/openai_responses/converters/output/utils.test.ts
```diff
@@ -386,3 +462,89 @@ describe("rawOutputToEvents", () => {
     });
   });
 });
+
+describe("streamErrorToErrorEvent", () => {
```
Nils-Fedrigo, line 466: is it not redundant with the previous test?

Nils-Fedrigo (approved):
ok, I'll leave it to you for the logic of error attributions, a few comments, maybe I'm confused

front/lib/model_constructors/sdk/anthropic_ai/converters/output/utils.ts
frankaloia: Yea so it was marked as unknown because it technically could be an anthropic issue if they are just sending us malformed tool jsons OR it could be us if we send bad argument shape or bad tool schema or something along those lines. "unknown" is safe in this case because they means we won't consider it an outage

front/lib/model_constructors/sdk/google_genai/converters/output/utils.ts
frankaloia: This is specifically the google_genai ApiError type which we don't have a function for yet

front/lib/model_constructors/sdk/google_genai/converters/output/utils.ts
frankaloia: For google gen AI, this is the same exact logic, just a refactor. It was previously:

```

  if (status === 401 || (status === 400 && isAuthMessage)) {
    return buildErrorEvent({
      errorSource: "dust",
      metadata,
      type: "authentication_error",
      message: `Authentication failed for Google: ${error.message}`,
      originalError: error,
    });
  }
```

front/lib/model_constructors/sdk/openai_responses/converters/output/utils.test.ts
frankaloia: Slightly different, this is parsing events from the SSE stream while the above is parsing API errors. So it's the difference between errors return from API vs when the POST already succeeded (HTTP 200) and we are iterating SSE events

PR #31484, opened 2026-08-31
Renew long-lived distributed locks

fontanierh:
## Description

Add owner-token-based lease renewal to the shared Redis lock abstraction. Renewal extends only the caller-owned lock, tracks a conservative monotonic deadline, retries bounded transient refresh failures, and exposes a guard that fails when ownership can no longer be proven.

This PR contains only generic lock infrastructure. Frame adoption is a separate hardening tip.

## Tests

- focused shared lock tests
- Front typecheck
- Biome and pre-commit checks
- exact final tree previously passed the full Front CI suite

## Risk

Low. Existing non-renewing lock callers retain their current behavior. Renewal is opt-in.

## Deploy Plan

Deploy normally. No migration or backfill.

### DIFF
--- a/front/lib/lock.test.ts
+++ b/front/lib/lock.test.ts
@@ -1,0 +1,276 @@
+import type { RedisClientType } from "@app/lib/api/redis";
+import {
+  distributedRefresh,
+  executeWithRenewingLockResult,
+  isLockLeaseLostError,
+} from "@app/lib/lock";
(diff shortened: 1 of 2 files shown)


fontanierh:
Replaced by smaller slices:

- #31509 atomic refresh primitive
- #31512 core renewable lock runner
- #31514 timing and transient-error hardening

Frame adoption is separately split across #31489, #31516, and #31517.

PR #31485, opened 2026-08-31
Migrate Fireworks models to Responses API

fontanierh:
## Description

- Move DeepSeek V4 Flash, Kimi K2.6, GLM 5.2, GLM 5.3 Flash, and Inkling to the Fireworks Responses API.
- Preserve optional tool parameters in the remaining Chat Completions path instead of forcing every schema property to be required under strict decoding.
- Keep Responses error messages attributed to Fireworks rather than the underlying OpenAI-compatible SDK.
- Retire the unavailable Kimi K2.5 serving endpoint while retaining its historical model metadata for stored runs and token attribution.
- Keep DeepSeek V4 Pro on Chat Completions until its active configurations can be moved to an approved replacement model ID.

The production failure came from the Chat converter rewriting optional tool fields as required under strict constrained generation. Fireworks then emitted values such as the string `null` for an omitted cursor, which failed downstream validation.

## Tests

Live-tested all five migrated model IDs against Fireworks across 230 endpoint cases, including optional cursor omission, reasoning, automatic and forced tools, output schemas, caching, and tool-result replay. Also exercised parallel multi-tool replay through the final response. Vision requests completed through Responses for Kimi K2.6, GLM 5.3 Flash, and Inkling, and each identified the dog in the test image.

Added focused coverage for optional schemas, Inkling output limits, historical Kimi K2.5 attribution, and provider-specific Fireworks errors.

## Risk

Moderate transport change across five endpoints. Each model passed the live Responses compatibility suite. Inkling keeps Dust's 64k output cap while omitting the invalid native 1M cap. Kimi K2.5 had no active production configuration before retirement. Rollback is the revert commit.

## Deploy Plan

Standard deploy. Monitor Fireworks provider errors and tool argument validation after rollout.


aubin-tchoi (approved):
LGTM

PR #31486, opened 2026-08-31
Move Frames across same-runtime mounts

fontanierh:
## Description

Stacked on #31480.

Allows Frames v2 sources to move between conversation and Pod mounts only when both resolve to the same runtime space. Both owners are resolved and authorized before mutation. Immediately before the final FileResource update, conversation owners are locked by stable sId order using their existing sandbox lifecycle serialization and their runtime spaces are resolved again. GCS copy stays outside those locks. If eligibility changed, copied generations are cleaned and the reservation is restored.

Lifecycle lock acquisition and release errors use the same final Frame reconciliation as an ambiguous database update. A committed destination returns success, the exact pending reservation is cleanup-safe, and missing or foreign state preserves destination objects.

## Tests

- `npm test -- move_source.test.ts sandbox_resource.test.ts frame_sandbox_adapter.test.ts` (83 tests)
- `npm run tsgo -- --noEmit`
- `npx biome check front/lib/api/frames/move_source.ts front/lib/api/frames/move_source.test.ts front/lib/resources/conversation_sandbox_adapter.ts front/lib/resources/sandbox_resource.ts front/lib/resources/sandbox_resource.test.ts`
- `git diff --check`

## Risk

The move state machine is sensitive to partial GCS, database, and lock failures. This change preserves the renewable locks, operation and attempt fencing, pinned generations, exact-generation recovery, and post-commit reconciliation from #31480. It does not destroy conversation sandboxes or use FrameSandboxAdapter scope transitions. Rollback is a revert of this PR.

## Deploy Plan

Merge after #31480, then use the standard Front deployment. No migration or backfill is required. Monitor Frame move errors and audit events after deploy.

### DIFF
--- a/cli/dust-sandbox/src/api/client.rs
+++ b/cli/dust-sandbox/src/api/client.rs
@@ -9,8 +9,9 @@ use tokio::time::sleep;
 use super::error::DustApiError;
 use super::types::{
     parse_action_poll_response, ActionPollResponse, CallToolPostResponse, CallToolRequest,
-    CallToolResponse, CallToolResult, FramePublishRequest, FramePublishResponse,
-    FrameRegisterRequest, FrameRegisterResponse, MCPServerView, SandboxServerViewsResponse,
+    CallToolResponse, CallToolResult, FrameMoveRequest, FrameMoveResponse, FramePublishRequest,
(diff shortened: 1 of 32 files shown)


fontanierh:
Cross-mount/runtime Frame moves remain deferred. The current scope is the smaller same-mount lane being recut under #31531/#31532 and its safety child.

PR #31489, opened 2026-08-31
Add renewable Frame lock wrappers

fontanierh:
## Description

Add renewable Frame publish-lock wrappers on top of #31514. The wrapper passes a lease guard to callers and maps acquisition/lease failures through typed Results.

Because importing Frame resources now reaches the renewable wrapper, this slice also completes the existing `FrameSandboxAdapter` test mock with `executeWithRenewingLockResult`; that is the earliest point where the mock becomes required.

Stack: #31509 -> #31512 -> #31514 -> this PR. Generic sandbox cleanup fencing follows in #31516. Retry/logging hardening #31551 remains an independent sibling from #31514.

## Tests

- Frame operation-lock test: 1 passed.
- Frame sandbox adapter tests: 4 passed.
- Biome and diff checks pass.

## Risk

No production caller changes in this slice. Existing zero-argument callbacks remain compatible.

## Deploy Plan

Merge after #31514. No migration is required.

### DIFF
--- a/front/lib/api/frames/operation_lock.test.ts
+++ b/front/lib/api/frames/operation_lock.test.ts
@@ -1,0 +1,43 @@
+import { withFramePublishLock } from "@app/lib/api/frames/operation_lock";
+import type { RedisClientType } from "@app/lib/api/redis";
+import { Ok } from "@app/types/shared/result";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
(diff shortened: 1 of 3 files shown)


fontanierh:
Closing with the renewable-lock approach: Frames v2 will use the existing fixed 10-minute lock instead of renewable wrappers.

PR #31491, opened 2026-08-31
Expose canonical Frames v2 invocation event stream

fontanierh:
## Description

Add the canonical Frames v2 invocation event stream keyed by stable Frame identity and invocation id. The stream rechecks current Frame use rights, resolves the invocation through its owning function row instead of the active publication, and therefore remains available after a Frame republishes.

The authenticated private route redirects to the SSE service, which repeats the feature gate and authorization checks. The private API is additive; legacy Frame and Pod Function event routes are unchanged. Swagger includes the approval and personal-auth event variants that the shared sandbox invocation stream can emit.

This PR is a true child of #31412 because both endpoints share the Frames route skeleton and API schemas. Its incremental diff contains only the SSE path.

## Tests

- Front API event tests: 6 files / 24 tests passed, including 4 canonical Frame SSE tests
- Front and Front API typechecks
- Swagger generation and strict validation
- Biome and git diff check

## Risk

Low and feature-gated. The endpoint is read-only and additive, fails closed when the Frame, current use right, or Frame-owned invocation is missing, and preserves Last-Event-ID replay through the existing stream helper. Rollback is reverting this PR.

## Deploy Plan

Merge and deploy after #31412. No migration or backfill. Enable the client switch only after both canonical POST and SSE server routes are available.

### DIFF
--- a/front-api/lib/api/sse/frame_function_invocation_events.ts
+++ b/front-api/lib/api/sse/frame_function_invocation_events.ts
@@ -1,0 +1,50 @@
+import { getSandboxFunctionInvocationEvents } from "@app/lib/api/sandbox_functions/events";
+import type { Authenticator } from "@app/lib/auth";
+import { FileResource } from "@app/lib/resources/file_resource";
+import { SandboxFunctionResource } from "@app/lib/resources/sandbox_function_resource";
+import { streamEvents } from "@front-api/lib/api/sse/stream_events";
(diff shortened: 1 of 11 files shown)


fontanierh:
Superseded by the focused canonical event stack: #31523, #31538, #31526, #31527, and optional documentation hardening #31528. The POST and SSE final trees merge cleanly and pass combined validation.

PR #31493, opened 2026-09-01
Block legacy sharing mutations for Frames v2

fontanierh:
## Description

Reject Frames v2 share-scope and email-grant mutations through the legacy generic FileResource routes. Frames v2 sharing is configured by users in the Dust UI, while agents can only retrieve existing links through the read-only flow merged in #31498.

Legacy read routes and Frames v1 mutation behavior are unchanged. The obsolete sandbox share-route change has been removed from this PR.

## Tests

- Legacy share scope and grant suites: 30 passed, including three Frames v2 rejection cases that preserve prior state
- Front-api `tsgo` passed
- Biome passed on all four changed files

## Risk

Low. This only rejects legacy mutation requests for Frames v2 and leaves the existing read paths intact.

## Deploy Plan

Standard deploy.


fontanierh:
Closing because this guard is obsolete after #31498. Frames v2 UI sharing still uses the generic FileResource share and grant mutation routes through ShareFrameSheet, so rejecting Frames v2 there would break the only current UI sharing controls. #31498 already makes the sandbox agent surface read-only by replacing mutation with existing-link retrieval. UI mutations must remain available.

PR #31497, opened 2026-09-01
Expose Frames v2 use rights endpoint

fontanierh:
## Description

Expose the source-authorized Frames v2 use-rights operation through the feature-gated sandbox endpoint.

Stacked on #31508. The reusable sharing primitives are in #31506.

## Tests

- Frames sandbox endpoint suite: 9 tests passed, including scope and grant updates, atomic allowlist failure, source write access, and feature gating

## Risk

Feature-gated. Frames v1 keeps its existing endpoints and behavior.

## Deploy Plan

Merge after #31508, then standard deploy.

### DIFF
--- a/front-api/routes/v1/w/[wId]/sandbox/frames/share.ts
+++ b/front-api/routes/v1/w/[wId]/sandbox/frames/share.ts
@@ -1,0 +1,122 @@
+import {
+  FrameSharingError,
+  type ShareFrameV2FromSourceError,
+  type ShareFrameV2FromSourceResult,
+  shareFrameV2FromSource,
+} from "@app/lib/api/frames/share_from_source";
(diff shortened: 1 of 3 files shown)


front-api/routes/v1/w/[wId]/sandbox/frames/share.ts
```diff
@@ -0,0 +1,122 @@
...
+      return apiError(ctx, {
+        status_code: 403,
+        api_error: {
+          type: "invalid_request_error",
+          message: "Frames v2 is not enabled for this workspace.",
+        },
+      });
+    }
+
+    const conversation = await ConversationResource.fetchById(auth, claims.cId);
```
Comment (davidebbo) [inline front-api/routes/v1/w/[wId]/sandbox/frames/share.ts:82]: Why does a frame need a conversation? I would have expected to be decoupled from that.

davidebbo (commented):
LGTM, w/ 1 question

PR #31499, opened 2026-09-01
Add Frame source operation locks

fontanierh:
## Description

Add a distinct distributed lock namespace, `frame:source:<frameId>`, for operations that read or mutate a Frames v2 source package.

A Frame identity is stable while its source path can change. Publishing concurrently with a move, clone, delete, or conversion could otherwise validate one path and then act on another. Callers acquire the source lock, re-fetch the Frame identity under that lock, and only then inspect or mutate its source. If publication is also involved, the existing `frame:publish` lock is acquired inside the source lock.

This PR only adds the fixed-lock primitive; it does not add call sites or acquire a `file:edit` lock.

The 10-minute TTL is a crash-expiry ceiling, not the expected lock duration. The lock is released immediately when the callback completes. A contender waits at most 30 seconds before returning a conflict. The conservative TTL prevents a slow build or GCS operation from outliving the lease, while ensuring a crashed process cannot leave a permanent lock.

## Tests

- Added focused coverage for the distinct source namespace and 10-minute TTL.
- Focused operation-lock test, Biome, and full `front` typecheck pass.

## Risk

Low. This adds an unused lock primitive. Later call sites can temporarily reject concurrent source operations, but successful operations release the lock immediately.

## Deploy Plan

Deploy normally. No migration or coordinated rollout is required.

### DIFF
--- a/front/lib/api/frames/operation_lock.ts
+++ b/front/lib/api/frames/operation_lock.ts
@@ -3,9 +3,14 @@ import { executeWithLockResult } from "@app/lib/lock";
 import type { Result } from "@app/types/shared/result";
 
 const FRAME_OPERATION_LOCK_TTL_MS = 10 * 60_000;
+const FRAME_OPERATION_LOCK_ACQUISITION_TIMEOUT_MS = 30_000;
 
 export function getFramePublishLockName(frameId: string): string {
   return `frame:publish:${frameId}`;
+}
+
+export function getFrameSourceLockName(frameId: string): string {
+  return `frame:source:${frameId}`;
 }
 
 export function withFramePublishLock<T, E>(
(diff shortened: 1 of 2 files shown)


front/lib/api/frames/operation_lock.ts
```diff
@@ -19,3 +23,15 @@ export function withFramePublishLock<T, E>(
...
 }
+
+export function withFrameSourceLock<T, E>(
+  frameId: string,
+  callback: () => Promise<Result<T, E>>
+): Promise<Result<T, E | LockAcquisitionTimeoutError>> {
+  return executeWithLockResult(
+    getFrameSourceLockName(frameId),
+    callback,
+    30_000,
```
flvndvd: No magic variable please 🙏

flvndvd (approved):

front/lib/api/frames/operation_lock.ts
fontanierh: fixing

PR #31502, opened 2026-09-01
Make FileResource own Frames v2 deletion

fontanierh:
## Description

Keep `FileResource` as the canonical Frames v2 identity and make `FileResource.delete(auth)` own the complete v2 lifecycle internally.

- Validate the registered manifest path, GCS-backed writable scope, stable identity, and absence of nested Frames under the publication lock.
- Delete source first, then project metadata/fragments, sandbox, functions/invocations, publication artifacts, sharing state, and the file row.
- Delegate project-file removal to this path before mutating project fragments.
- Keep generic file deletion and workspace scrubbing behavior unchanged except for Frame-owned runtime cleanup.
- Expose no Frame deletion callback, wrapper resource, service, or new data model.

First deletion slice for #31402.

## Tests

97 focused tests pass across file deletion, sandbox lifecycle serialization, publication serialization, and project fragment failure ordering. Biome and `tsgo --noEmit` pass.

## Risk

A partial v2 deletion after source removal is retryable: missing source is accepted and the stable file identity remains until runtime and database cleanup succeeds. Database-backed Frame source deletion is rejected explicitly in this GCS-only slice.

## Deploy Plan

Standard deploy. No migration or coordinated rollout.


front/lib/resources/file_resource.ts
```diff
@@ -472,6 +478,32 @@ export class FileResource extends BaseResource<FileModel> {
...
 
+  static async fetchFrameV2Descendants(
+    auth: Authenticator,
+    parent: FileResource
+  ): Promise<FileResource[]> {
+    if (!parent.mountFilePath) {
+      return [];
+    }
+    const owner = auth.getNonNullableWorkspace();
+    const directoryPrefix = parent.mountFilePath.slice(
```
davidebbo, line 489: nit: feels like we should have helpers for these common GCS path operations.

front/lib/resources/file_resource.ts
```diff
@@ -717,16 +753,142 @@ export class FileResource extends BaseResource<FileModel> {
...
+      const frame = await this.fetchFreshFrameV2(auth);
+      if (!frame || frame.toScopedPath(auth) !== manifestPath) {
+        return new Err(
+          new Error(
+            "The Frame source changed while it was being deleted; retry from its current path."
+          )
+        );
+      }
+
+      const descendantFrames = await FileResource.fetchFrameV2Descendants(
```
davidebbo, line 813: In what scenario would we have nested frames?

davidebbo (commented):
LGTM. A bit confused about nested frames.

front/lib/resources/file_resource.ts
fontanierh: Technically there is nothing that prevents this:
```
Parent/manifest.json
Parent/Child/manifest.json
```

Registration accepts any writable manifest.json, so an agent could register both paths.
So if we blindly delete we could end up with corruption

front/lib/resources/file_resource.ts
fontanierh: directionally agreed, though this one is very specific and would only have a single caller I think

front/lib/resources/file_resource.ts
davidebbo: I see. Hopefully not a mainstream case.

fontanierh: ideally we'll figure a way to block it, but it's not trivial as you need to prevent moves etc..

fontanierh: at the very least we'll prompt against it becase it will look pretty broken (our UI isn't made for that)

PR #31509, opened 2026-09-01
Refresh owned distributed locks

fontanierh:
## Description

Add the atomic Redis primitive needed to extend a distributed lock only while the caller still owns it. This is the first small replacement slice for #31484.

## Tests

- Focused ownership and stale-owner test passes.
- Biome and diff checks pass.

## Risk

No existing call path changes. This only exposes the refresh primitive.

## Deploy Plan

Deploy normally. No migration or ordering dependency.


fontanierh:
Closing this approach: Frames v2 operations are expected to complete well within the existing 10-minute lock TTL, so lease refresh/renewal complexity is not justified for the initial scope.

PR #31510, opened 2026-09-01
Move Frames within one mount

fontanierh:
## Description

- add the bounded backend operation for moving a registered Frames v2 source folder within the same conversation or Pod mount
- preserve the stable FileResource identity, publication metadata, sharing state, and source ownership metadata
- reject nested registered files and occupied destinations before copying
- raw-list at most 1,024 GCS objects and 100 MiB so hidden, processed, and placeholder objects are included
- copy objects, commit the FileResource path, best-effort delete the source, and emit the existing file.moved audit event

This is deliberately non-crash-safe and has no production caller until the HTTP child merges. Reservations, recovery, generation fencing, rollback, and renewable locks remain out of scope.

## Tests

Added a focused end-to-end test proving the folder moves while the stable Frame identity and active publication are preserved. Also ran Biome and the front typecheck.

## Risk

Once called, a crash or typed failure can leave partial destination objects. The FileResource path remains authoritative. Generation-safe copy and cleanup stay isolated in a separate child PR.

## Deploy Plan

### DIFF
--- a/front/lib/api/files/gcs_mount/files.ts
+++ b/front/lib/api/files/gcs_mount/files.ts
@@ -422,7 +422,7 @@ async function moveGCSMountFile({
   return new Ok(undefined);
 }
 
-async function emitGCSMountFileMovedAuditLog(
+export async function emitGCSMountFileMovedAuditLog(
   auth: Authenticator,
   scope: GCSMountPoint,
   {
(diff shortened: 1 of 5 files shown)


fontanierh:
Superseded by #31536 and #31537 for the two foundations, plus the focused +284 move operation in #31532. Children #31511, #31494, and #31495 now target #31532.

PR #31512, opened 2026-09-01
Renew distributed lock leases

fontanierh:
## Description

Add a result-based lock runner that renews ownership while a callback runs and exposes a lease guard for callers to check before committing writes. A refresh error or ownership change fails closed.

Stack: #31509 -> this PR. Timing and transient-error retry hardening remain a separate child.

## Tests

- All 6 focused core renewal tests pass.
- Biome and diff checks pass.

## Risk

No existing caller changes. The new runner fails the operation if renewal becomes uncertain.

## Deploy Plan

Merge after #31509. No migration is required.


fontanierh:
Closing with the renewable-lock approach: the existing 10-minute fixed lease is sufficient for the initial Frames v2 operations.

PR #31514, opened 2026-09-01
Measure renewable leases from request start

fontanierh:
## Description\n\nMeasure renewable lease safety from the start of the successful acquisition request, not from its delayed response. The same request-start rule already used for refreshes now has direct regression coverage.\n\nMandatory lock correctness stack: #31509 -> #31512 -> this PR. Frame lock wrappers may build on this head. Transient refresh retries and recovery logging move to a separate late hardening tip.\n\n## Tests\n\n- Renewable lock suite: 8 tests passed, including delayed acquisition and delayed refresh responses\n- Biome and diff checks\n\n## Risk\n\nLow. A delayed Redis response can only shorten the locally trusted lifetime; it can no longer extend trust beyond the Redis TTL.\n\n## Deploy Plan\n\nMerge after #31512, before renewable Frame wrapper adoption.

### DIFF
--- a/front/lib/lock.test.ts
+++ b/front/lib/lock.test.ts
@@ -55,6 +55,32 @@ describe("renewing locks", () => {
     ).resolves.toBe(false);
   });
 
+  it("measures the initial lease from the start of lock acquisition", async () => {
+    const redisClient = makeRedisClient();
+    let finishAcquisition: (value: string) => void = () => {};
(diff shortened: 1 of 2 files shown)


fontanierh:
Closing with the renewable-lock approach: request-start lease accounting is unnecessary without lease renewal.

PR #31516, opened 2026-09-01
Guard sandbox cleanup with renewable leases

fontanierh:
## Description

Add the generic pre-cleanup lease boundary required by Frame deletion: the callback is checked before provider destruction, before database cleanup, and again before transaction commit. A lost lease stops cleanup and rolls database changes back. FrameSandboxAdapter forwards the optional guard to the shared sandbox cleanup primitive.

Stack: #31509 -> #31512 -> #31514 -> #31489 -> this PR. Source-lock and publication adoption follow this shared cleanup primitive. Retry/logging hardening #31551 is an independent sibling.

## Tests

- Focused SandboxResource coverage proves lease loss before commit rolls back owner and sandbox rows.
- Frame sandbox adapter tests cover the forwarded callback contract.
- 53 focused tests pass.
- Biome and diff checks pass.

## Risk

Existing callers are unchanged because the guard is optional. Guarded cleanup now fails instead of committing after ownership is lost.

## Deploy Plan

Merge after #31489. No migration is required.

### DIFF
--- a/front/lib/resources/frame_sandbox_adapter.ts
+++ b/front/lib/resources/frame_sandbox_adapter.ts
@@ -261,14 +261,16 @@ export class FrameSandboxAdapter {
     frame: FrameSandboxOwner,
     {
       afterSandboxCleanup,
+      beforeSandboxCleanup,
     }: {
       afterSandboxCleanup?: () => Promise<Result<undefined, Error>>;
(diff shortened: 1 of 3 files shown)


fontanierh:
Closing with the renewable-lock approach: sandbox cleanup will use the existing fixed 10-minute lock without renewal infrastructure.

PR #31517, opened 2026-09-01
Renew Frame publication operation locks

fontanierh:
## Description

Adopt renewable lease checks for Frame publication, activation, and deletion before durable state changes. The inherited #31516 primitive carries the lease guard through sandbox cleanup. This slice uses the publish-lock domain; the following slice adds the outer source lock for publication.

Stack: #31516 -> #31499 -> this PR. Retry/logging hardening #31551 is independent and is not inherited.

## Tests

- 88 focused publication and FileResource tests pass.
- Biome and diff checks pass.

## Risk

Publication or deletion now fails rather than committing after losing publish-lock ownership. Existing timeout errors remain typed.

## Deploy Plan

Merge after #31499. No migration is required.


fontanierh:
Closing with the renewable-lock approach: publication fencing will be recut on the existing fixed 10-minute lock.

PR #31518, opened 2026-09-01
Resolve Frame function references by renderer kind

fontanierh:
## Description

Classify renderer function references by Frame kind. Frames v2 qualify bare manifest function names with the stable Frame id, while legacy Frames keep Pod-relative resolution.

This is the shared base for the renderer replacements of #31401. No visual or Sparkle property changes.

## Tests

- Front Vitest: 1 file, 5 tests passed
- `git diff --check`

## Risk

Low. This adds a typed resolver and tests without changing a renderer call site. Rollback is reverting this PR.

## Deploy Plan

Deploy normally. Merge before its renderer children.

### DIFF
--- a/front/types/api/frame_function_reference.test.ts
+++ b/front/types/api/frame_function_reference.test.ts
@@ -1,0 +1,59 @@
+import {
+  getFrameFunctionReferenceKind,
+  resolveFrameFunctionReference,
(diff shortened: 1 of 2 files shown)


smb2268 (commented):
LGTM

PR #31519, opened 2026-09-01
Bind renderer invocations to Frame identity

fontanierh:
## Description

Pass the stable Frames v2 identity through conversation and public renderers, then bind `callFunction` messages to that identity in the trusted iframe host. Legacy Frames keep Pod-relative resolution.

This is a focused renderer child replacing part of #31401. No visual or Sparkle property changes.

## Tests

- Front Vitest: 3 files, 14 tests passed
- `git diff --check`

## Risk

Low and feature-gated. Raw visualizations and legacy Frames do not receive a Frame id. Rollback is reverting this PR.

## Deploy Plan

Merge after the Frame function reference base. Deploy normally.


davidebbo (commented):
LGTM

PR #31520, opened 2026-09-01
Expose linked file metadata on path HEAD

fontanierh:
## Description

Return the linked FileResource MIME type from the existing path HEAD request through a dedicated CORS-exposed Dust header. Keep the mounted object's raw `Content-Type` unchanged and extend the existing SWR metadata lookup without adding a second request.

This is a metadata foundation replacing part of #31401. The private API change is additive and backward compatible.

## Tests

- Front Vitest: 1 file, 6 tests passed
- Front API Vitest: 2 files, 29 tests passed
- `git diff --check`

## Risk

Low. Existing file-id consumers and raw content metadata remain unchanged. Rollback is reverting this PR.

## Deploy Plan

Merge after the Frame function reference base. Deploy normally before the Pod renderer child.


flvndvd (commented):
LGTM

PR #31522, opened 2026-09-01
Bind Pod sheet invocations to Frame identity

fontanierh:
## Description

Load Pod Frame sheet metadata only while the sheet is open, fail closed when the resource is unavailable, and pass the stable identity only for Frames v2.

This optional renderer tip replaces the remaining Pod sheet portion of #31401. No visual or Sparkle property changes.

## Tests

- Front Vitest: 1 file, 1 test passed
- `git diff --check`

## Risk

Low and feature-gated. Closed sheets stop an unnecessary metadata request. Rollback is reverting this PR.

## Deploy Plan

Merge after the Frame function reference base. It can deploy independently of the other renderer children.


smb2268 (commented):
LGTM

PR #31524, opened 2026-09-01
Resolve active Frame functions for use

fontanierh:
## Description

Resolve one named function from a Frame's active immutable publication and recheck current Frame use rights. Resolution fails closed if the publication changes during lookup.

This is the POST resolver foundation replacing part of #31412.

## Tests

- Front Vitest: 1 file, 3 tests passed
- `git diff --check`

## Risk

Low and feature-gated at its callers. The helper is additive and the existing reference resolver keeps its public contract. Rollback is reverting this PR.

## Deploy Plan

Merge after #31538 and #31545, then retarget from the temporary integration base to main. Merge before the POST transport.


flvndvd (commented):
LGTM

PR #31526, opened 2026-09-01
Resolve invocations through Frame ownership

fontanierh:
## Description

Resolve a sandbox function invocation through its owning Frame and function row. The lookup is independent of the active publication, so an in-flight stream survives a Frame republish while cross-Frame access fails closed.

This is the SSE resolver foundation replacing part of #31491.

## Tests

- Front Vitest: 1 file, 2 tests passed
- `git diff --check`

## Risk

Low. The helper is additive and scopes every lookup to the current workspace and owning Frame. Rollback is reverting this PR.

## Deploy Plan

Merge after #31523 and #31545, then retarget from the temporary integration base to main. Merge before the SSE endpoint.

### DIFF
--- a/front/lib/resources/sandbox_function_resource.ts
+++ b/front/lib/resources/sandbox_function_resource.ts
@@ -590,6 +590,60 @@ export class SandboxFunctionResource extends BaseResource<SandboxFunctionModel> 
     });
   }
 
+  /**
+   * Fetch an invocation through its stable Frame identity. The function row is resolved from the
+   * invocation itself rather than from the active publication, so an in-flight stream survives a
+   * republish or removal of the function from a later publication.
+   */
+  static async fetchInvocationByFrameAndId(
+    auth: Authenticator,
+    {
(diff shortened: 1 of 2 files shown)


front/lib/resources/sandbox_function_resource.ts
```diff
@@ -590,6 +590,60 @@ export class SandboxFunctionResource extends BaseResource<SandboxFunctionModel>
...
+      !isResourceSId("sandbox_function_invocation", invocationId)
+    ) {
+      return null;
+    }
+    const invocationModelId = getResourceIdFromSId(invocationId);
+    if (invocationModelId === null) {
+      return null;
+    }
+
+    const invocation = await SandboxFunctionInvocationModel.findOne({
```
flvndvd, line 619: Feels weird to circumvent the resource here to fetch it a few lines below. Ideally, all the boilerplate code above lives exclusively in the associated resource.

PR #31529, opened 2026-09-01
Add a Frame sandbox cleanup hook

fontanierh:
## Description

Allow Frame deletion to run a guarded callback immediately before sandbox cleanup. This small contract is shared by the otherwise independent publication/resource and sandbox lock-adoption lanes.

Stack: #31489 -> this PR.

## Tests

- All 4 focused Frame sandbox adapter tests pass.
- Biome and diff checks pass.

## Risk

Existing callers are unchanged because the hook is optional. Callback failure stops cleanup.

## Deploy Plan

Merge after #31489. No migration is required.


fontanierh:
Folded back into #31516: the adapter hook is not independently valid without the SandboxResource fencing behavior. The combined PR is still only 113 additions/10 deletions.

PR #31413, opened 2026-08-29
Present Frame packages in the file explorer

fontanierh:
## Description

- Collapse registered Frames v2 folders into one package entry in conversation and Pod file explorers behind `frames_v2`.
- Open the Frame on primary click and expose raw source through `View source`.
- Keep generic rename, move, and delete actions off package entries.

## Tests

- `npm test components/file_explorer/FileExplorer.test.tsx`
- `npm run tsgo`
- `npx biome check --error-on-warnings` on all six changed files

## Risk

The presentation is feature-flagged. Source listing metadata and package construction are isolated in the two preceding PRs.

## Deploy Plan

Standard deploy after #31533 and #31535.

PR #31533, opened 2026-09-01
Expose Frame package file metadata

fontanierh:
## Description

Annotate file-system entries that are registered Frames v2 with stable Frame identity and source-package path. This is the backend/type foundation extracted from #31413; explorer aggregation and rendering remain separate.

## Tests

- 26 focused space-file route tests pass.
- Biome and diff checks pass.

## Risk

The internal file-list response gains optional fields. Existing consumers remain compatible.

## Deploy Plan

Deploy normally. The route is internal and ignored by Swagger. No migration is required.


smb2268 (commented):
LGTM!

PR #31535, opened 2026-09-01
Build Frame packages in the file explorer

fontanierh:
## Description

Aggregate a registered Frames v2 manifest and its source descendants into one package node while preserving previewable source entries. This is the pure explorer-pipeline slice extracted from #31413; component rendering remains separate.

Stack: #31533 -> this PR.

## Tests

- All 10 focused pipeline tests pass.
- Biome and diff checks pass.

## Risk

Only entries carrying the optional Frame metadata from #31533 are grouped. Ordinary files and folders retain the existing pipeline.

## Deploy Plan

Merge after #31533. No migration is required.


smb2268 (commented):
looks good!

PR #31536, opened 2026-09-01
Resolve same-mount Frame move paths

fontanierh:
## Description

- normalize Frames v2 source and destination directories
- reject identical, nested, user, and cross-mount moves
- derive manifest paths, destination mount scope, and audit-relative paths
- expose the shared typed move error used by the orchestration layer

This pure path foundation has no production caller until the move operation merges.

## Tests

Added 5 focused tests covering conversation paths and audit metadata, Pod scope resolution, trailing-slash normalization, trailing-slash nesting regressions, and cross-mount rejection.

## Risk

Low. This adds a pure validation helper without a production caller.

## Deploy Plan

### DIFF
--- a/front/lib/api/frames/move_source_paths.test.ts
+++ b/front/lib/api/frames/move_source_paths.test.ts
@@ -39,6 +39,19 @@ describe("resolveFrameSourceMovePaths", () => {
     });
   });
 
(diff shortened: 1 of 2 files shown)


flvndvd (commented):
LGTM

PR #31537, opened 2026-09-01
Add bounded Frame source storage copies

fontanierh:
## Description

Add an operation-neutral Frame source-storage foundation that inspects an empty destination, enumerates the complete raw source tree, enforces the 1,024-object and 100 MiB limits, validates the manifest, and copies with bounded concurrency.

The shared FileStorage mock now models baseline unconditional copy and reads copied object content. This fidelity belongs with the source-storage contract and is exercised directly by its copy test. Generation preconditions remain in child #31540 and generation-safe source copying remains in #31549.

## Tests

- Frame source-storage suite: 4 tests passed, including baseline copy and read fidelity.
- Biome on changed files.
- Front tsgo.

## Risk

Low. This direct-main foundation is not wired to a public endpoint. Typed failures preserve the source as authoritative.

## Deploy Plan

Deploy normally. Consumers may merge after this foundation.

### DIFF
--- a/front/lib/api/frames/source_storage.ts
+++ b/front/lib/api/frames/source_storage.ts
@@ -1,0 +1,157 @@
+import path from "node:path";
+import { getPrivateUploadBucket } from "@app/lib/file_storage";
+import { concurrentExecutor } from "@app/lib/utils/async_utils";
+import type { Result } from "@app/types/shared/result";
+import { Err, Ok } from "@app/types/shared/result";
+import { normalizeError } from "@app/types/shared/utils/error_utils";
+
+const FRAME_SOURCE_COPY_CONCURRENCY = 4;
+const MAX_FRAME_SOURCE_FILE_COUNT = 1024;
+const MAX_FRAME_SOURCE_BYTES = 100 * 1024 * 1024;
+
+export class FrameSourceStorageError extends Error {
+  constructor(
+    readonly code: "conflict" | "copy_failed" | "invalid_source",
+    message: string
+  ) {
+    super(message);
+    this.name = "FrameSourceStorageError";
+  }
+}
+
+function storageError(code: FrameSourceStorageError["code"], message: string) {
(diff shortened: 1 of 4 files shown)


front/lib/api/frames/source_storage.ts
```diff
@@ -0,0 +1,157 @@
...
+  } catch (error) {
+    return storageError(
+      "copy_failed",
+      `Failed to inspect the Frame source; the source remains authoritative: ${normalizeError(error).message}`
+    );
+  }
+  if (sourceObjects.length > MAX_FRAME_SOURCE_FILE_COUNT) {
+    return storageError(
+      "invalid_source",
+      "Frame source exceeds the copy size or file count limit."
```
davidebbo, line 77: Only 'file count' here

front/lib/api/frames/source_storage.ts
```diff
@@ -0,0 +1,157 @@
...
+        "invalid_source",
+        `Frame source object has invalid size metadata: ${sourceObject.name}`
+      );
+    }
+    sourceSizeBytes += sizeBytes;
+  }
+  if (sourceSizeBytes > MAX_FRAME_SOURCE_BYTES) {
+    return storageError(
+      "invalid_source",
+      "Frame source exceeds the copy size or file count limit."
```
davidebbo, line 95: Only 'copy size' here

front/lib/api/frames/source_storage.ts
```diff
@@ -0,0 +1,157 @@
...
+  }
+
+  return new Ok({
+    destinationMountPrefix,
+    sourceMountPrefix,
+    sourceObjectNames: sourceObjects.map((entry) => entry.name),
+  });
+}
+
+export async function copyFrameSourceStorage(
```
davidebbo, line 112: Could use a comment here. This is about copying from where to where?

davidebbo (commented):
LGTM with minor comments

PR #31538, opened 2026-09-01
Define canonical Frames v2 invocation contract

fontanierh:
## Description

Define the canonical Frames v2 invocation contract before transport wiring. This adds the strict parameter and request-body parsers, shared private Swagger schemas, generated endpoint documentation, and parser coverage.

This contract foundation replaces the documentation portion of #31412. It has no runtime handler.

## Tests

- Front API Vitest: 1 file, 2 parser tests passed
- `npm run docs` in `front-api`
- `git diff --check`

## Risk

Very low. Runtime behavior is unchanged. Private Swagger can describe the additive route before the transport child merges.

## Deploy Plan

Merge after the canonical route foundation, then merge either resolver lane.

### DIFF
--- a/front-api/routes/w/[wId]/frames/[frameId]/functions/[name]/index.ts
+++ b/front-api/routes/w/[wId]/frames/[frameId]/functions/[name]/index.ts
@@ -2,4 +2,52 @@ import { workspaceApp } from "@front-api/middlewares/ctx";
 
 const app = workspaceApp();
 
+/**
+ * @swagger
+ * /api/w/{wId}/frames/{frameId}/functions/{name}/invocations:
+ *   post:
+ *     summary: Invoke an active Frames v2 function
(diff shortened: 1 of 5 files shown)


front-api/routes/w/[wId]/frames/[frameId]/functions/[name]/index.ts
```diff
@@ -2,4 +2,52 @@ import { workspaceApp } from "@front-api/middlewares/ctx";
 
 const app = workspaceApp();
 
+/**
```
Comment (davidebbo) [inline front-api/routes/w/[wId]/frames/[frameId]/functions/[name]/index.ts:5]: Do we really need swagger for non-public?

davidebbo (commented):
LGTM, though I don't know why we have swagger for private methods.

fontanierh:
>LGTM, though I don't know why we have swagger for private methods.

you are right, probably not useful actually...

PR #31551, opened 2026-09-01
Retry transient lock refresh failures

fontanierh:
## Description\n\nRetry transient Redis refresh failures while the last confirmed lease lifetime is still provably safe. Retries use bounded exponential backoff, fail closed at the safety boundary, and log first failure, recovery, or final loss.\n\nThis is a late hardening tip on #31514. Functional Frame lock adoption does not depend on it.\n\n## Tests\n\n- Renewable lock suite: 9 tests passed, including transient recovery and exhausted safe lifetime\n- Biome and diff checks\n\n## Risk\n\nLow to moderate. Definitive owner changes still fail immediately; only transient Redis errors are retried.\n\n## Deploy Plan\n\nMerge after #31514 when convenient. No functional stack depends on this tip.

### DIFF
--- a/front/lib/lock.test.ts
+++ b/front/lib/lock.test.ts
@@ -152,7 +152,7 @@ describe("renewing locks", () => {
     expect(result.isErr() && isLockLeaseLostError(result.error)).toBe(true);
   });
 
-  it("fails the lease when refresh errors", async () => {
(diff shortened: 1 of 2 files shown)


fontanierh:
Closing with the renewable-lock approach: refresh retries are unnecessary without lease renewal.

PR #31558, opened 2026-09-01
[front] Fix workspace deletion after agent identity migration

davidebbo:
## Description

Delete stable `AgentModel` rows at the end of `deleteAgentsActivity`, after all agent configurations are removed. This unblocks workspace hard deletion when `agents_workspaceId_fkey` still references the workspace and makes retries converge after a partial previous run.

## Tests

- `npm run test -- poke/temporal/activities.test.ts`
- `npx tsgo --noEmit`
- `npx biome check front/poke/temporal/activities.ts front/poke/temporal/activities.test.ts`

## Risk

Low. The change affects only the Poke workspace hard-deletion activity. It deletes identity rows only after their configuration rows have been deleted. Re-running the activity is safe when no identity rows remain.

## Deploy Plan

Deploy Front workers normally. Monitor or retry the affected workspace deletion workflow after the new activity code is live.

### DIFF
--- a/front/poke/temporal/activities.test.ts
+++ b/front/poke/temporal/activities.test.ts
@@ -1,10 +1,14 @@
 import { Authenticator } from "@app/lib/auth";
+import { AgentModel } from "@app/lib/models/agent/agent";
 import { ActivationPodResource } from "@app/lib/resources/activation_pod_resource";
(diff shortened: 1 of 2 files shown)


philipperolet (approved):
LGTM

PR #31560, opened 2026-09-01
[core] Fix data source deletion with orphaned relational rows

davidebbo:
## Description

`DataSource::delete` listed tables and folders through `data_sources_nodes`. A table or folder row without its node was skipped, so the final `data_sources` delete failed on `tables_data_source_fkey` or `data_sources_folders_data_source_fkey`.

Add deletion-specific relational candidate queries that read `tables` and `data_sources_folders` directly. Table candidates still use the existing table lifecycle, including transient database invalidation and GCS cleanup for local tables. Folder candidates use the existing transactional folder deletion. Normal table and folder listing behavior is unchanged.

Add PostgreSQL regression tests for both missing-node cases.

## Tests

- `cargo fmt --all -- --check`
- `cargo test --all`

## Risk

Low. The new queries are used only during full data source deletion. Existing table and folder deletion paths still own cleanup, and retries remain idempotent after partial deletion.

## Deploy Plan

Deploy Core normally. After deployment, retry `poke-jaeLj27yy8-scrub-data-source-dts_NuzxXDRsVegj` and `poke-jaeLj27yy8-scrub-data-source-dts_j9wupsiOuhbpj`.

### DIFF
--- a/core/src/data_sources/data_source.rs
+++ b/core/src/data_sources/data_source.rs
@@ -2135,9 +2135,10 @@ impl DataSource {
         );
 
         // Delete tables (concurrently).
-        let (tables, total) = store
-            .list_data_source_tables(&self.project, &self.data_source_id, &None, &None, None)
+        let tables = store
+            .list_data_source_table_deletion_candidates(&self.project, &self.data_source_id)
             .await?;
+        let total = tables.len();
 
         info!(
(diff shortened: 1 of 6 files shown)


davidebbo:
Was hand fixed instead.

PR #31563, opened 2026-09-01
feat: return spaces of api keys instead of groups

matteotrab:
## Description

This PR returns the spaces accessible to each API key instead of requiring the frontend to resolve and display its underlying groups.

## Tests

Manually.

## Risks

The new response serialization performs additional group-permission and space lookups when API keys are returned. Incorrect permission mapping could result in incomplete or inaccurate space labels in the UI, but authorization behavior remains unchanged because the derived spaces are display-only.

## Deploy Plan

Standard deployment.

### DIFF
--- a/front/lib/resources/key_resource.ts
+++ b/front/lib/resources/key_resource.ts
@@ -3,7 +3,9 @@
 
 import type { Authenticator } from "@app/lib/auth";
 import { BaseResource } from "@app/lib/resources/base_resource";
-import type { GroupResource } from "@app/lib/resources/group_resource";
+import { GroupPermissionResource } from "@app/lib/resources/group_permission_resource";
+import { GroupResource } from "@app/lib/resources/group_resource";
+import { SpaceResource } from "@app/lib/resources/space_resource";
 import { KeyModel } from "@app/lib/resources/storage/models/keys";
 import { UserModel } from "@app/lib/resources/storage/models/user";
 import type { ReadonlyAttributesType } from "@app/lib/resources/storage/types";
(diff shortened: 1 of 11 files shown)


matteotrab:

- #31461
    - #31464
        - #31563 ◀
            - #31565

front/lib/resources/key_resource.ts
```diff
@@ -376,13 +379,113 @@ export class KeyResource extends BaseResource<KeyModel> {
...
+
+        return [
+          key.id,
+          [...spaces].sort((a, b) => a.name.localeCompare(b.name)),
+        ];
+      })
+    );
+  }
+
+  static async toJSONWithSpaces(
```
fabiencelier, line 466: can we add unit tests on this to make sure we return the expected spaces resources

fabiencelier (commented):
So now the user can add a space A to the key, then we save group X, and when we display it we might see that group X grant access to key A and B and show 2 spaces in the UI 
Is it possible ?

fabiencelier:
> So now the user can add a space A to the key, then we save group X, and when we display it we might see that group X grant access to key A and B and show 2 spaces in the UI
Is it possible ?

Seen IRL: for new keyw we will only store the regular auto group of the space (and then match back the space based on it)
But for historical keys with any group related to it: there can be regular manual groups stored for the key, which will map several spaces.
LGTM then
Maybe just add the unit test on the resource + integration test on the endpoint: creating a key with a space then fetching the key doed return the space

fabiencelier (approved):

PR #31566, opened 2026-09-01
Swap the input bar "+" menu in place instead of opening a second dropdown

pinotalexandre:
### Description
- Improve hover timing on dropdown items
- Swap the input bar "+" menu in place instead of opening a second dropdown 

https://github.com/user-attachments/assets/92431caf-26de-465e-862e-b749aa7f073e


front/components/assistant/conversation/input_bar/InputBarPlusMenu.tsx
```diff
@@ -90,197 +157,232 @@ export function InputBarPlusMenu({
...
+            isLoading={!!isSpacesLoading}
+            onBack={goBack}
+            selectedSpaceIds={selectedSpaceIds}
+            onSelectedSpaceIdsChange={onSelectedSpaceIdsChange}
+            spaces={spaces ?? []}
+          />
+        );
+      case "root":
+        return rootPage;
+    }
```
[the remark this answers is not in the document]
pinotalexandre: Added in 5b58750f51 — with one deviation from the suggestion: [REACT5] requires `assertNeverAndIgnore` rather than `assertNever` in React components, so the branch is:

```ts
default:
  assertNeverAndIgnore(page);
  return null;
```

Compile-time exhaustiveness is identical; this just avoids throwing in the menu if the union ever grows.

front/components/assistant/conversation/input_bar/InputBarPlusMenu.test.tsx
```diff
@@ -0,0 +1,100 @@
...
+  metadata: null,
+  metronomeCustomerId: null,
+  sharingPolicy: "all_scopes",
+  regionalModelsOnly: false,
+} satisfies WorkspaceType;
+
+const spaces = [
+  { sId: "space-1", name: "Engineering", kind: "regular" },
+  { sId: "space-2", name: "Marketing", kind: "regular" },
+] as SelectableConversationSpaceType[];
```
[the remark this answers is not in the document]
pinotalexandre: Fixed in 5b58750f51 — the fixture now carries the full `SelectableConversationSpaceType` shape (`managementMode`, `createdAt`, `updatedAt`, `groupIds`, `isRestricted`, `selected`) and is checked with `satisfies` instead of cast with `as`.

front/components/assistant/conversation/input_bar/InputBarPlusMenu.tsx
```diff
@@ -21,24 +21,86 @@ import type { UserType, WorkspaceType } from "@app/types/user";
...
   DropdownMenuItem,
   DropdownMenuTrigger,
+  Icon,
   Planet,
   Plus,
   ShapesPlus,
 } from "@dust-tt/sparkle";
-import { useRef, useState } from "react";
+import type { Transition, Variants } from "framer-motion";
+import { AnimatePresence, motion, useReducedMotion } from "framer-motion";
```
pinotalexandre: Not applying this one — the ~30 kb is already on the wire for this route regardless.

`InputBarPlusMenu` only ever renders inside `AgentInputBar` (`AgentInputBar` → `InputBar` → `InputBarContainer` → `InputBarButtons` → `InputBarPlusMenu`), and `AgentInputBar.tsx` itself does `import { AnimatePresence, motion, useReducedMotion } from "framer-motion"` eagerly. Switching this one component to `LazyMotion` + `m` would add a provider and a second import style without saving a byte, and it would diverge from its immediate neighbour ([GEN1]).

Worth doing properly as its own change if we want the conversation route on `LazyMotion` — that means converting `AgentInputBar` and its siblings together, which is out of scope here.

front/components/assistant/conversation/input_bar/InputBarPlusMenu.tsx
```diff
@@ -8,37 +8,104 @@ import {
...
```
pinotalexandre: Applied in 65cda44c82 — `LazyMotion` + `m`, with `domAnimation` rather than the `domMax` used elsewhere in the repo, since the page swap needs variants and exit animations but not drag or layout projection.

Recording the caveat from the earlier thread so it is not lost: this does not shrink the conversation route today. `AgentInputBar` — the direct ancestor of this component — and `SidebarMenu` both still `import { AnimatePresence, motion }` eagerly, so the full bundle is on the wire regardless. Landing it because it costs four lines and matches the existing `LazyMotion` usage in `UserAnalyticsPopover` / `AgentInsightsTab` / the analytics tables; the actual ~30 kb needs those two converted as their own change.

front/components/assistant/conversation/input_bar/DropdownPanel.tsx
```diff
@@ -0,0 +1,72 @@
...
+
+interface DropdownPanelContentProps {
+  children: React.ReactNode;
+  className?: string;
+  dropdownHeaders?: React.ReactNode;
+  onBack?: () => void;
+  title?: string;
+}
+
+export function DropdownPanelContent({
```
aubin-tchoi: feels like this should be in sparkle

front/components/assistant/conversation/input_bar/DropdownPanel.tsx
```diff
@@ -0,0 +1,72 @@
...
+          />
+        )}
+        <DropdownMenuSeparator />
+        {dropdownHeaders}
+      </div>
+      <ScrollArea
+        className="w-full flex-1"
+        hideScrollBar={false}
+        orientation="vertical"
+        viewportClassName="flex-1 overscroll-contain"
```
aubin-tchoi: nice!

front/components/assistant/conversation/input_bar/InputBarAttachmentsPicker.tsx
```diff
@@ -91,7 +93,7 @@ interface InputBarAttachmentsPickerProps {
   onNodeSelect: (node: DataSourceViewContentNode) => void;
   onNodeUnselect: (node: DataSourceViewContentNode) => void;
   attachedNodes: DataSourceViewContentNode[];
-  type: "dropdown" | "subdropdown";
+  type: "dropdown" | "panel";
```
aubin-tchoi: we remove the subdropdown mode entirely?

front/components/assistant/conversation/input_bar/InputBarPlusMenu.tsx
```diff
@@ -8,37 +8,110 @@ import {
...
 );
 
-const MOBILE_PICKERS = ["capabilities", "attachments", "spaces"] as const;
-type MobilePicker = (typeof MOBILE_PICKERS)[number];
+const DRILL_IN_CHEVRON = (
+  <Icon size="xs" visual={ChevronRight} className="text-primary-400" />
+);
+
+const PANEL_SWAP_OFFSET_PX = 8;
+const PANEL_EASE_ENTER = [0.215, 0.61, 0.355, 1] as const;
```
aubin-tchoi: the durations should be stored in sparkle, some of the values from sparkle's theme.css are ported to values that can be used in framer motion in https://github.com/dust-tt/dust/blob/main/sparkle/src/lib/motion.ts#L7

front/components/assistant/conversation/input_bar/InputBarPlusMenu.tsx
```diff
@@ -8,37 +8,110 @@ import {
...
+const PANEL_ENTER_TRANSITION: Transition = {
```
aubin-tchoi, line 58: I find the animation to be a bit "bouncy"

front/components/assistant/conversation/input_bar/DropdownPanel.tsx
```diff
@@ -0,0 +1,72 @@
...
+                size="xs"
+                visual={ChevronLeft}
+                className="text-muted-foreground"
+              />
+            }
+            onSelect={(event) => event.preventDefault()}
+            onClick={onBack}
+          />
+        )}
+        <DropdownMenuSeparator />
```
aubin-tchoi: nit: we should hide this separator if there's no title

front/components/assistant/conversation/input_bar/InputBarPlusMenu.tsx
```diff
@@ -90,197 +161,228 @@ export function InputBarPlusMenu({
...
+            <Icon
+              size="xs"
+              visual={ShapesPlus}
+              className="text-muted-foreground"
+            />
+          }
+          disabled={disabled}
+          endComponent={DRILL_IN_CHEVRON}
+          onSelect={(event) => event.preventDefault()}
+          onClick={() => openPage("capabilities")}
```
aubin-tchoi: nit: I think that a single onSelect that calls onBack or smth is better in terms of a11y

front/components/assistant/conversation/input_bar/InputBarPlusMenu.tsx
```diff
@@ -90,197 +161,228 @@ export function InputBarPlusMenu({
...
+            className={PLUS_BUTTON_CLASSNAME}
+          />
+        </DropdownMenuTrigger>
+        <DropdownMenuContent
+          align="start"
+          collisionPadding={8}
+          className="w-80 max-w-[calc(100vw-1rem)] overflow-hidden p-0"
+        >
+          <div
+            className="relative w-full overflow-hidden transition-[height] duration-150 ease-enter motion-reduce:transition-none"
```
aubin-tchoi: shouldn't this be a duration 140? would be even better to use an existing token if there's one

front/components/assistant/conversation/input_bar/InputBarPlusMenu.tsx
```diff
@@ -8,37 +8,110 @@ import {
...
+interface PanelSizerProps {
+  children: React.ReactNode;
+  onHeightChange: (page: PlusMenuPage, height: number) => void;
+  page: PlusMenuPage;
+}
+
+function PanelSizer({ children, onHeightChange, page }: PanelSizerProps) {
+  const nodeRef = useRef<HTMLDivElement>(null);
+
+  useLayoutEffect(() => {
```
aubin-tchoi: nit: I think that during the popLayout you render 2 resize observers at once (incoming and outgoing)

front/components/assistant/conversation/input_bar/DropdownPanel.tsx
pinotalexandre: Done

front/components/assistant/conversation/input_bar/InputBarPlusMenu.tsx
pinotalexandre: Done!

PR #31567, opened 2026-09-01
fix(poke): make favorite star icon reflect toggle state

thomasvicaire:
## Summary
- `PokeFavoriteButton` rendered `icon={isCurrentlyFavorite ? Star01 : Star01}` — both branches resolved to the same outline icon, so the star never visually updated when toggled (only the tooltip text changed).
- Now uses `StarFilled` when favorited and `Star01` (outline) otherwise, matching the existing pattern in `SkillFavoriteButton.tsx`.
- Toggle behavior, accessibility/tooltip labels, and styling (`variant`, `size`) are unchanged.

## Test plan
- [x] Added `front/components/poke/PokeFavorites.test.tsx` covering: outline icon when unfavorited, filled icon when favorited, click calls `toggleFavorite`.
- [x] `npm run test -- PokeFavorites.test.tsx` passes (3/3).


front/components/poke/PokeFavorites.test.tsx
```diff
@@ -0,0 +1,88 @@
+import { PokeFavoriteButton } from "@app/components/poke/PokeFavorites";
```
adrsimon: we do not need 88 lines of test to check that the star is the right one this feels overkill, also we usually don't test pure front-end rendering stuff
(proof is we don't have the right tooling for that and you end end mocking a whole lot of functions just to test we have the right icon)
could you remove it ? 🙏 tyty

adrsimon (approved):
LGTM but unsure we need this test tbh

front/components/poke/PokeFavorites.test.tsx
thomasvicaire: fair point lol, thanks for the review

PR #13361, opened 2025-06-11
`run_agent` origin and participation skip

spolu:
## Description

Removes `mcp` origin in favor of `run_agent`.
Skips upserting participation for user when user message origin is `run_agent`

Fixes https://github.com/dust-tt/tasks/issues/3105

## Tests

Tested locally

## Risk

Low

## Deploy Plan

- deploy `front`

PR #31574, opened 2026-09-01
[front] fix: exclude sub-conversations (depth > 0) from conversation search

PopDaph:
## Fix

Sub-conversations created by `run_agent` (e.g. deep-dive calling dust_task agents) show up in the sidebar conversation search.

<kbd>
Screenshot 2026-09-01 at 11 50 05
</kbd>

## Description

Since #28348 child conversations are created in the same pod as their parent, so they get indexed into the pod's `dust_project` data source and surface in semantic search. #28679 hid depth > 0 conversations from the pod lists but missed the search paths.

- `semantic_search` (sidebar search): drop depth > 0 results after fetch.
- `spaces/[spaceId]/search_conversations` (pod page search): same filter, consistent with the pod conversation list which already hides them.
- `searchByTitlePaginated` (sidebar title search): add `depth = 0` to the where clause. Current sub-conversations have no participation row so they don't surface here, but sub-conversations created before the participation skip (#13361) can.

Sub-conversations stay indexed in the pod data source (agent search may rely on them); only the user-facing endpoints filter them out.

Follow up of #28679.

## Tests

One regression test per endpoint: a depth 1 conversation matching the query is not returned.

## Risk

Worst case, a legitimate root conversation is hidden from search results. Safe to rollback.

### DIFF
--- a/front/lib/api/projects/search.ts
+++ b/front/lib/api/projects/search.ts
@@ -26,6 +26,9 @@ interface ConversationSearchResult {
   spaceId: string;
 }
 
+// TODO(2026-09-01 SEARCH): sub-conversations (depth > 0) are indexed and consume
+// the topK budget; callers filter them out post-fetch. Exclude them at sync time
+// instead (listSpaceConversationsForSync) and clean up already-indexed documents.
 export async function searchProjectConversations(
   auth: Authenticator,
   options: SearchProjectConversationsOptions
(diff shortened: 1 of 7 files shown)


front/lib/api/projects/search.ts
```diff
@@ -26,6 +26,9 @@ interface ConversationSearchResult {
   spaceId: string;
 }
 
+// TODO(2026-09-01 SEARCH): sub-conversations (depth > 0) are indexed and consume
+// the topK budget; callers filter them out post-fetch. Exclude them at sync time
+// instead (listSpaceConversationsForSync) and clean up already-indexed documents.
```
Fraggle, line 31: If you remove them from the endpoint that fetch the conversations to sync, they will be automatically cleaned up.

Fraggle (approved):
LGTM

PR #31575, opened 2026-09-01
[front] poke: add reset fair-use button to members usage table

tdraier:
## Description

Adds a "Reset" button shortcut to the poke Usage tab members table (`PokeMembersUsageTable`) so support can reset a member's current AWU fair-use consumption directly from the row, next to the existing Grant / Reconcile actions.

Screenshot 2026-09-01 at 11 51 45

The button reuses the existing `reset-message-rate-limit` poke plugin with its `user_awu_fair_use` target (which expires the per-user fair-use Redis counter via `resetFairUseAwuCreditsRateLimitForUser`) — no new plugin was needed. It only renders for rows that carry a fair-use limit and a known email, and refreshes the table on success.

## Tests

Manually verified type-checking (`npx tsgo --noEmit`) and formatting. The underlying plugin path is unchanged; this is a UI shortcut into an existing, already-registered plugin (gated on the `support` poke role).

## Risk

Low. UI-only change in poke (internal admin). No API/schema changes; the reset action goes through the existing plugin with its existing role gating. Safe to roll back by reverting.

## Deploy Plan

Standard front deploy. No migrations or coordination required.

### DIFF
--- a/front/components/poke/credits/ResetFairUseButton.tsx
+++ b/front/components/poke/credits/ResetFairUseButton.tsx
@@ -1,0 +1,73 @@
+import { useSendNotification } from "@app/hooks/useNotification";
+import { useRunPokePlugin } from "@app/poke/swr/plugins";
+import type { WorkspaceType } from "@app/types/user";
+import { Button } from "@dust-tt/sparkle";
+import { useState } from "react";
+
+const RESET_MESSAGE_RATE_LIMIT_PLUGIN_ID = "reset-message-rate-limit";
+
+interface ResetFairUseButtonProps {
+  owner: WorkspaceType;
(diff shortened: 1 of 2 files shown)


front/components/poke/credits/ResetFairUseButton.tsx
```diff
@@ -0,0 +1,73 @@
...
+      resourceType: "workspaces",
+      resourceId: owner.sId,
+      workspace: owner,
+    },
+  });
+
+  const handleClick = async () => {
+    setIsRunning(true);
+    const result = await doRunPlugin({
+      resetTarget: ["user_awu_fair_use"],
```
PopDaph, line 39: awoooo 🐺

PopDaph (approved):

PR #31576, opened 2026-09-01
Harden cache with race prevention.

Fraggle:
## Description

`cacheWithRedis` now uses per-key generation fences to prevent an in-flight database read from repopulating Redis with stale data after invalidation. Lua scripts atomically delete payloads and increment generations, while cache loads and `warmCacheWithRedis` write only when their observed generation still matches. Batch invalidation and migrated keys use the same atomic path.

## Tests

Added Vitest coverage for generation-guarded writes, stale-load invalidation races, warm-cache behavior, batch invalidation, key migrations, and feature-flag cache invalidation. Extended the Redis test mock to execute the new Lua paths.

## Risk

Low risk. The change only affects Redis cache population and invalidation; stale writes are rejected rather than served. Rollback is safe, and generation keys are harmless if older code resumes managing payload keys.

## Deploy Plan

Deploy `front`. No SQL migration or infrastructure change required.


Fraggle:

- #31576 ◀

id13 (approved):
LGTM

PR #31578, opened 2026-09-01
[front] Don't SDK-retry the seat contracts.edit (deltas stack on 504)

tdraier:
## Description

The Metronome SDK client uses `maxRetries: 5`, so any 5xx is retried up to 5 times. **Two** of our contract edits are cumulative **deltas**, and the edit endpoint returns **504 gateway timeouts that still apply server-side** — so an SDK retry re-applies the delta and **stacks** it. (Vanta cleanup: a single "add 137 unassigned" that 504'd landed **4× → 548**.)

Fix, per what each endpoint supports:

- **`updateSubscriptionSeats` (`contracts.edit`)** — attach a fresh per-edit **`uniqueness_key`**. Metronome rejects a duplicate key with a **409**, so an SDK retry of an already-applied edit is a no-op we catch (`ConflictError`) and treat as success. This keeps the SDK's transient-5xx recovery *and* makes the edit idempotent against retries — strictly better than disabling retries. (Thanks @tdraier for the pointer to `uniqueness_key`.)
- **`addManualBalanceEntry`** (per-seat ledger deltas, the pro↔max credit carry) — the v1 endpoint has **no** `uniqueness_key`, so a retry can't be deduped. There we set **`maxRetries: 0`** and let the caller retry at its own bounded/spaced level (the seat-sync Temporal activity, #31422, which re-reads first).

**Not** applied to the other `contracts.edit` callers, on purpose — those are idempotent and *should* keep retrying:
- `add_credits` / `add_commits` already carry a `uniqueness_key`.
- `updateSubscriptionQuantity`, `editCredit`, billing-provider config, contract-name — absolute *sets*.

## Tests

- `tsgo --noEmit` clean. Adds a `uniqueness_key` body field + a `ConflictError` branch on one call, and a per-request `maxRetries: 0` on another; no change to the edit payloads' semantics. Existing seat-sync suites unaffected.

## Risk

- Low. `contracts.edit` gains a fresh uniqueness_key per edit (each edit is still a distinct key, so nothing is wrongly deduped across separate reconciles); a 409 on our own key means the edit already applied → success. `addManualBalanceEntry` now surfaces a transient 5xx as `Err` instead of silently retrying — the activity retries it (bounded + spaced). Net: no more silent double-applies on either delta. Safe to roll back by reverting.

## Deploy Plan

- Standard deploy; no migration. Independent of the other seat-sync PRs.

### DIFF
--- a/front/lib/metronome/client.test.ts
+++ b/front/lib/metronome/client.test.ts
@@ -301,16 +301,21 @@ describe("adjustSeatCreditBalances", () => {
     });
 
     expect(result.isOk()).toBe(true);
-    expect(mockAddManualBalanceEntry).toHaveBeenCalledWith({
-      id: "credit-1",
-      customer_id: "cust-1",
-      contract_id: "contract-1",
-      segment_id: "segment-1",
-      amount: -1500,
-      per_group_amounts: { seatA: -1000, seatB: -500 },
-      reason: "test adjustment",
-      timestamp: "2026-06-11T15:00:00.000Z",
-    });
+    expect(mockAddManualBalanceEntry).toHaveBeenCalledWith(
+      {
+        id: "credit-1",
+        customer_id: "cust-1",
+        contract_id: "contract-1",
+        segment_id: "segment-1",
+        amount: -1500,
+        per_group_amounts: { seatA: -1000, seatB: -500 },
+        reason: "test adjustment",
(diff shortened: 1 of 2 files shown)


tdraier:

- #31578 ◀

fabiencelier (approved):

PR #31580, opened 2026-09-01
[front] pod sync: exclude sub-conversations from the conversations data source

PopDaph:
## Fix

Sub-conversations (`run_agent` children, depth > 0) are indexed into the pod's `dust_project` data source: they show up in search and their citations link to conversations that are hidden from every list.

Stop indexing them, front only:

- `listSpaceConversationsForSync`: exclude depth > 0, the connector stops upserting them.
- `conversation_ids` (GC endpoint): exclude depth > 0, the connector GC deletes the already-indexed documents after each sync. No connector deploy, no manual cleanup.

Both use a new `onlyRootConversations` option on `FetchConversationOptions`.

Follow up of #31574 (whose read-time filters stay in place as guards).

## Tests

A depth 1 conversation is absent from the sync list and from the GC id list.

## Risk

Worst case, a root conversation's documents get GC'd from the pod index. Revert stops the exclusion; a removed document comes back on the conversation's next update or a manual full sync.


Fraggle (approved):
LGTM, thanks !

PR #31583, opened 2026-09-01
[front] enh: surface webhook source id and GCS bucket link in poke

adrsimon:
## Description

Poke only showed webhook source *views*, so getting to a source's stored payloads meant looking up its model id by hand. The trigger page now links to the webhook source (name, sId, id) and to its GCS folder, and the source page shows the id plus the same bucket link.

## Tests

Manual: opened a webhook trigger and its webhook source in poke, checked both links resolve.

## Risk

Low, poke-only reads.

## Deploy Plan

Deploy front and front-api.


davidebbo (approved):
LGTM. Nice!

PR #31607, opened 2026-09-01
Prevent legacy Frames from using the v2 CLI

fontanierh:
## Description

When `frames_v2` is disabled, tell agents with Computer enabled to load `Create Frames`, publish with `publish_interactive_content_file`, and never use `dsbx frame`. This guard lives in the Computer instructions because Frame-customizing skills do not automatically load the base Frames skill or its MCP server.

## Tests

Added coverage for both flag states in `sandbox.test.ts`.

## Risk

Low. Prompt-only change for unflagged workspaces; flagged workspaces retain the CLI workflow.

## Deploy Plan

Standard Front deployment. Safe to roll back.


davidebbo (approved):
LGTM

PR #31565, opened 2026-09-01
feat: remove scopable group endpoint and groupIds from keyType

matteotrab:
## Description

This PR removes group-based API key scoping from the API, making spaces the only exposed scoping representation.

- Removes the `/keys/groups` endpoint and related implementation.
- Removes `groupIds` from serialized API keys and `KeyType`.
- Updates key creation tests to verify group scoping directly in the database.

## Tests

Manually.

## Risks

Low. We are not reading the groupIds returned by the API anymore.

## Deploy Plan

Wait for old clients still expecting the groupIds to cycle out.

PR #31612, opened 2026-09-01
feat: fetch only visible groups when listing workspace groups

matteotrab:
## Description

This PR updates workspace group listing to return only user-visible groups by default. This is needed for the refactor of the group_resource permissions, which will throw errors when checking permissions for regular_auto groups. In fact, regular_auto groups should only be fetched through their resource.

This PR also restricts the groups shown in the poke tab to visible groups only. This removes the noise coming from the regular_auto groups linked to spaces, skills and agents, which can be reached directly through their resource.

## Tests

Manually.

## Risks

Low. The only call sites where we were fetching regular_auto groups using this function were in poke.

## Deploy Plan

Standard deployment.


matteotrab:

- #31461
    - #31464
        - #31563
            - #31565
                - #31612 ◀

front-api/routes/poke/workspaces/[wId]/groups/index.ts
```diff
@@ -17,11 +12,7 @@ const app = pokeApp();
 app.get("/", async (ctx): HandlerResult<PokeListGroups> => {
   const auth = ctx.get("auth");
 
-  const groups = await GroupResource.listAllWorkspaceGroups(auth, {
-    groupKinds: GROUP_KINDS.filter(
-      (kind) => !isSystemGroupKind(kind) && !isAgentEditorGroupKind(kind)
-    ),
-  });
+  const groups = await GroupResource.listAllWorkspaceGroups(auth);
```
Comment (matteotrab) [inline front-api/routes/poke/workspaces/[wId]/groups/index.ts:15]: I think it makes sense to only have visible groups in the poke groups tab, considering that skill and agent editors group are now regular_auto groups.

Regular auto groups are still reachable from their resource (space, skill or agent)

front/lib/api/slack/summoning_whitelist.ts
```diff
@@ -57,6 +57,27 @@ async function fetchSlackBotDataSource(
   return new Ok({ dataSource, connectorId: dataSource.connectorId });
 }
 
+// The groups a Slack bot may be whitelisted with: the workspace groups a caller can reference
+// directly (global, provisioned, regular_manual), plus the `regular_auto` groups backing the
+// workspace spaces — a space's group is how the bot gets access to that space's agents. The other
+// `regular_auto` groups are agent and skill editor groups, which are never whitelistable.
+export async function listSlackWorkflowGroups(
```
matteotrab: This also removes the skill and agent editor groups that were included before

fabiencelier:
```suggestion
export async function listSlackWorkflowWhitelistableGroups(
```
?

fabiencelier (approved):

front/migrations/20241218_delete_dangling_groups.ts
philipperolet, line 1: nit: ideally we do not delete migrations, but there's a comment line for them to be skipped by linter / compiler

front/lib/api/slack/summoning_whitelist.ts
```diff
@@ -57,6 +57,27 @@ async function fetchSlackBotDataSource(
   return new Ok({ dataSource, connectorId: dataSource.connectorId });
 }
 
+// The groups a Slack bot may be whitelisted with: the workspace groups a caller can reference
+// directly (global, provisioned, regular_manual), plus the `regular_auto` groups backing the
+// workspace spaces — a space's group is how the bot gets access to that space's agents. The other
+// `regular_auto` groups are agent and skill editor groups, which are never whitelistable.
```
philipperolet, line 63: very nit: we might check if any agent / skill group had been used before on an agent whitelist workflow (might break at some point if so); i'd say very unlikely but who knows

philipperolet (approved):
LGTM 🙏 Left minor nits

front/lib/api/slack/summoning_whitelist.ts
matteotrab: Good point, just checked and there are only regular auto groups for spaces

PR #31614, opened 2026-09-01
[front] Show confirmation content in ask_user_question (Forms) card

tdraier:
## Description

Fixes dust-tt/tasks#10196.

When an agent uses the `ask_user_question` (Forms) tool to confirm something ("Does this look right? Any corrections before I post?"), the content being confirmed was frequently never shown to the user. The tool had no field for it, so the subject lived only in the model's reasoning (collapsed/hidden) or was materialized *after* the answer — the confirmation card showed just the question and its options, so "this" had no visible referent.

This adds an optional `content` field to `UserQuestionSchema`. The tool description now instructs the model to put the exact content it asks the user to review/confirm (draft message, email, Slack post, payload) into that field. The live answer card (`UserAnswerRequired`) and the read-only action details (`MCPAskUserQuestionActionDetails`) render it as markdown, in a scrollable muted block above the question. The field flows through the pause resource, resume state, SSE event and the blocked-actions reload endpoint automatically, since they pass the whole `question` object through.

Screenshot 2026-09-01 at 14 57 22

## Tests

- `npm run test -- UserAnswerRequired` — added two tests (renders the content block when present; no block when absent). 14/14 pass.
- `npx tsgo --noEmit` clean.

## Risk

Low. `content` is optional (`.nullish()`) and append-only: old stored tool params and older clients are unaffected, and `UserQuestionSchema.safeParse` on historical actions still succeeds. Purely additive, no breaking API/schema change. Safe to roll back.

## Deploy Plan

Standard front deploy. No migration and no ordering constraints.

### DIFF
--- a/front/lib/actions/types/index.ts
+++ b/front/lib/actions/types/index.ts
@@ -65,6 +65,18 @@ export const UserQuestionSchema = z.object({
   question: z
     .string()
     .describe("The question text. Should be clear and specific."),
+  content: z
+    .string()
+    .nullish()
+    .describe(
+      "The exact content this question is about, rendered as markdown, so " +
+        "the user can review it before answering. Set this whenever the " +
+        "question asks the user to review, confirm, or validate specific " +
+        "content you have prepared (e.g. a draft message, email, post, or " +
+        "payload) — put the full content here rather than only referring to " +
+        "it in the question. Leave null for pure clarification questions " +
+        "that have no content to show."
+    ),
   options: z
     .array(UserQuestionOptionSchema)
     .describe(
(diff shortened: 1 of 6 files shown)


tdraier:

- #31614 ◀

front/lib/actions/types/index.ts
```diff
@@ -65,6 +65,18 @@ export const UserQuestionSchema = z.object({
...
+    .string()
+    .nullish()
+    .describe(
+      "The exact content this question is about, rendered as markdown, so " +
+        "the user can review it before answering. Set this whenever the " +
+        "question asks the user to review, confirm, or validate specific " +
+        "content you have prepared (e.g. a draft message, email, post, or " +
+        "payload) — put the full content here rather than only referring to " +
+        "it in the question. Leave null for pure clarification questions " +
+        "that have no content to show."
```
aubin-tchoi: let's make this a lot shorter (ideally 1 line) and less prescriptive. if we need to add instructions we can add them in `constructToolsSection`

PR #31615, opened 2026-09-01
[front] chore: add legacy fair use ff script migration

id13:

### DIFF
--- a/front/migrations/20260901_enable_fair_use_for_legacy_plan_workspaces.ts
+++ b/front/migrations/20260901_enable_fair_use_for_legacy_plan_workspaces.ts
@@ -1,0 +1,187 @@
+import { parse } from "csv-parse";
+import * as fs from "fs";
+
+import { PlanModel } from "@app/lib/models/plan";
+import { FeatureFlagResource } from "@app/lib/resources/feature_flag_resource";
+import { frontSequelize } from "@app/lib/resources/storage";
+import { SubscriptionResource } from "@app/lib/resources/subscription_resource";
+import { WorkspaceResource } from "@app/lib/resources/workspace_resource";
+import { makeScript } from "@app/scripts/helpers";
+import type { WhitelistableFeature } from "@app/types/shared/feature_flags";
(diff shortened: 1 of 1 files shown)


front/migrations/20260901_enable_fair_use_for_legacy_plan_workspaces.ts
```diff
@@ -0,0 +1,144 @@
...
+          ROW_NUMBER() OVER (
+            PARTITION BY subscriptions."workspaceId"
+            ORDER BY
+              (subscriptions.status = 'active') DESC,
+              subscriptions."startDate" DESC,
+              subscriptions."createdAt" DESC,
+              subscriptions.id DESC
+          ) AS rank
+        FROM subscriptions
+        INNER JOIN plans ON plans.id = subscriptions."planId"
```
tdraier, line 73: aren't you missing a `subscriptions."endDate" is null` ? this seems to return a lot of workpaces

front/migrations/post-deploy/20260901100000_enable_legacy_plan_fair_use.sql
```diff
@@ -0,0 +1,17 @@
...
+WHERE SUBSTRING("code" FROM 1 FOR 3) <> 'CP_'
+  AND (
+    "maxAwuCredits" IS DISTINCT FROM 20000
+    OR "maxAwuCreditsTimeframe" IS DISTINCT FROM 'week'
+  )
+  AND EXISTS (
+    SELECT 1
+    FROM "public"."subscriptions"
+    WHERE "subscriptions"."planId" = "plans"."id"
+  );
```
tdraier: seen IRL , rather do a script to handle cache

front/migrations/20260901_enable_fair_use_for_legacy_plan_workspaces.ts
id13: good point

tdraier (approved):

PR #31616, opened 2026-09-01
[front] chore: read source breakdown from consumption analytics

aubin-tchoi:
## Description

`fetchContextOriginBreakdown` still reads from the legacy message analytics ES index, which we plan to retire in favor of the analytics consumption index.

This PR switches the call site in `fetchContextOriginBreakdown` to the new index.
Also contains a script that computes the current diff between the two queries.

## Tests

- Added tests.

## Risk

- Low.

## Deploy Plan

- Run script.
- Deploy front.


aubin-tchoi:
superseded by the upcoming removal of the analytics page

PR #31618, opened 2026-09-01
Sidekick: always run in standard even if user restricted to basics

fabiencelier:
## Description

closes https://github.com/dust-tt/tasks/issues/10246

Always allow standard tier for sidekick which is the harcoded tier.

sidekick is free for now so no impact for customers. 
Avoid downgrading to a lower tier where sidekick would not work.

## Tests

 - Added unit tests
 - tested locally: even when restricted to basics I can now use sidekick

## Risk

low
add new flags to bypass the model tier check but there are only usable from sidekick and sidekick does not allow to override model (blocked on purpose)

## Deploy Plan

deploy front

### DIFF
--- a/front/lib/model_tiers/access.test.ts
+++ b/front/lib/model_tiers/access.test.ts
@@ -131,7 +131,7 @@ describe("getModelTierAccessErrorForAgentConfiguration", () => {
 
     const error = await getModelTierAccessErrorForAgentConfiguration(auth, {
       agentSId,
-      agentName: "test-agent",
+      agentName: agentSId,
       model: streamModel,
       modelResolutionMethod: AUTO_MODEL_ID,
     });
--- a/front/lib/model_tiers/access.test.ts
+++ b/front/lib/model_tiers/access.test.ts
@@ -1,14 +1,18 @@
 import { Authenticator } from "@app/lib/auth";
+import { getModelConfigByModelId } from "@app/lib/llms/model_configurations";
 import { getModelTierAccessErrorForAgentConfiguration } from "@app/lib/model_tiers/access";
 import { setUserMaxAllowedTier } from "@app/lib/model_tiers/allowed_tiers";
 import { WorkspaceResource } from "@app/lib/resources/workspace_resource";
 import { MembershipFactory } from "@app/tests/utils/MembershipFactory";
 import { UserFactory } from "@app/tests/utils/UserFactory";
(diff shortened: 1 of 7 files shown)


front/lib/model_tiers/access.test.ts
```diff
@@ -104,6 +112,30 @@ describe("getModelTierAccessErrorForAgentConfiguration", () => {
...
+    const streamModel = getModelConfigByModelId(
+      MODEL_STREAMS[AUTO_MODEL_ID][0].modelId
+    );
+    if (!streamModel) {
+      throw new Error("Standard stream's first candidate is not supported.");
+    }
+
+    const error = await getModelTierAccessErrorForAgentConfiguration(auth, {
+      agentSId,
+      agentName: "test-agent",
```
Nils-Fedrigo: worth renaming the agentName to make it clearer to understand

front/lib/model_tiers/agent_tier_overrides.ts
```diff
@@ -0,0 +1,19 @@
...
+// own tier cap is. The Standard stream's candidates never exceed the `balanced`
+// tier, so this can never hand a member a premium model.
+const SIDEKICK_MAX_MODEL_TIER: ModelsTierName = "balanced";
+
+// Tiers an agent may use regardless of the member's own tier grants, or null when
+// the member's grants apply as usual.
+export function getAgentAllowedTierNamesOverride(
+  agentSId: string
+): ModelsTierName[] | null {
+  return agentSId === GLOBAL_AGENTS_SID.SIDEKICK
```
Nils-Fedrigo, line 16: Just to check:
is there a way for a user to modify the sId of an agent?

Nils-Fedrigo (approved):
lgtm but is there a way for a user to modify the sId of an agent?

This seems semi-risky so want to be sure

fabiencelier:
> lgtm but is there a way for a user to modify the sId of an agent?

nope
And there is a limit of usage per day on sidekick anyway

front/lib/model_tiers/agent_tier_overrides.ts
fabiencelier: no way to do it

front/lib/model_tiers/access.test.ts
fabiencelier: done

PR #31620, opened 2026-09-01
[front] Render persisted content for messages paused on a user question

tdraier:
## Description

Follow-up to #31614 (independent). Fixes the reload/hydration half of dust-tt/tasks#10196.

A message paused on `ask_user_question` (Forms) has message status `created` — the blocked status lives on the action, not the message. On a fresh load, `makeInitialMessageStreamState` deliberately discarded the persisted `activitySteps` (setting `inlineActivitySteps: []`) and marked the message `agentState: "thinking"`, delegating all reconstruction to the live SSE replay. The message body is additionally gated on `agentState === "done"`.

That replay only exists for the Redis stream TTL (10 min). A message paused waiting for a user answer stops publishing, so once the stream expires a reload rebuilds nothing: empty timeline, non-`done` state, gated body → the content the user was meant to review renders nowhere (while it is intact server-side / in poke).

This change:
- **Hydrates persisted `activitySteps` for `created` messages too** (`types.ts`), so a reloaded paused message shows its steps even with no live stream.
- **De-dupes against the live replay** (`useAgentMessageStream.ts`): when a stream *is* available it replays from the start and would rebuild the same steps, so the hydrated copy is cleared exactly once, on the first replayed step-bearing event (`generation_tokens` tokens/CoT, or `tool_params`). An expired stream never emits such an event, so the hydrated steps are preserved. This mirrors the existing fresh-mount content/CoT reset.
- **Renders the body when the message is blocked** (`AgentMessage.tsx`): the body gate now also opens when a blocked action is present, so a paused message's persisted body (the last text fragment) shows even though `agentState !== "done"`.

## Tests

- `npm run test -- useAgentMessageStream` — added two tests: `makeInitialMessageStreamState` hydrates steps for a paused `created` message, and the hook clears hydrated steps on the first replayed event (de-dup). 18/18 pass.
- Existing `AgentMessage.compactUIView` + stream + markdown suites still green (63 tests).
- `npx tsgo --noEmit` clean.

## Risk

Low–moderate; the de-dup touches the live streaming reducer. The reset is guarded by a fresh-mount ref that is only true when persisted steps were hydrated on a `created` message (i.e. a reload of an existing message) — a brand-new live message has empty `activitySteps`, so behavior there is unchanged. The body-gate change is additive (blocked messages only). Safe to roll back.

## Deploy Plan

Standard front deploy. No migration, no ordering constraints.

### DIFF
--- a/front/components/assistant/conversation/AgentMessage.tsx
+++ b/front/components/assistant/conversation/AgentMessage.tsx
@@ -1473,7 +1473,8 @@ function AgentMessageContent({
 
         {agentMessage.content !== null &&
           agentMessage.content !== "" &&
-          agentMessage.streaming.agentState === "done" && (
+          (agentMessage.streaming.agentState === "done" ||
+            blockedAction !== null) && (
             <div>
               <AgentMessageMarkdown
                 content={sanitizeVisualizationContent(agentMessage.content)}
--- a/front/components/assistant/conversation/types.ts
+++ b/front/components/assistant/conversation/types.ts
@@ -239,9 +239,15 @@ export const makeInitialMessageStreamState = (
     streaming: {
       actionProgress: new Map(),
       agentState: message.status === "created" ? "thinking" : "done",
-      // Live messages rebuild inline steps from the SSE replay on mount.
-      inlineActivitySteps:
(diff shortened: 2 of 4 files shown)


tdraier:

- #31620 ◀

PR #31622, opened 2026-09-01
feat: serve feedback distrib from pg

sfriquet:
## Description

Refs https://github.com/dust-tt/tasks/issues/10262

Update route `/api/w/:wId/assistant/agent_configurations/:aId/observability/feedback-distribution` to fetch data from pg instead of es.

## Tests

- added test for the endpoint

## Risk

Low, safe to rollback until the ES index is deleted.

## Deploy Plan

front only.


front/lib/resources/agent_message_feedback_resource.ts
```diff
@@ -591,6 +599,52 @@ export class AgentMessageFeedbackResource extends BaseResource<AgentMessageFeedb
...
+    const rows = await replicaDb.query<{
+      day: string;
+      positive: string;
+      negative: string;
+    }>(
+      `
+      SELECT
+        "createdAt"::date AS day,
+        SUM(CASE WHEN "thumbDirection" = 'up' THEN 1 ELSE 0 END) AS positive,
+        SUM(CASE WHEN "thumbDirection" = 'down' THEN 1 ELSE 0 END) AS negative
```
aubin-tchoi: OOC how different would it be with a `COUNT(*) FILTER (WHERE "thumbDirection" = 'up')`?
to avoid the case on every row

aubin-tchoi (approved):
LGTM

front/lib/resources/agent_message_feedback_resource.ts
sfriquet: it'll execute the same  but agree `count filter` is less verbose 👍

PR #31623, opened 2026-09-01
[front] Complete sandbox child workflows after activity failures

davidebbo:
## Description

Sandbox child tool workflows currently finalize the action as errored and then rethrow terminal activity failures. A worker-drain race can therefore leave a correctly finalized child action behind a failed Temporal workflow.

This change treats the errored action as the in-band result for new histories. It logs the activity failure, keeps finalization non-cancellable, and returns successfully. A new Temporal patch marker preserves the recorded failure outcome for older histories.

## Tests

- npm run test -- temporal/agent_loop/workflows.test.ts temporal/agent_loop/activities/finalize_sandbox_child_tool.test.ts
- npx tsgo --noEmit
- Pre-commit Biome and front type-check hooks

The full front test suite was started but not run to completion.

## Risk

Risk is low. The sandbox action status remains errored and stays the result consumed by polling clients. No-retry activities remain single-attempt. Workflow failure alerts become structured workflow error logs.

The patch marker makes forward replay safe. A rollback must keep patch-aware workers available until marked executions close. The marked branch completes immediately after logging.

## Deploy Plan

Use the standard forward rolling deploy. Monitor sandbox child workflow failures, the Sandbox child tool activity failed log, and child actions stuck in running. For a rollback, let marked executions close before removing patch-aware workers.

### DIFF
--- a/front/temporal/agent_loop/workflows.ts
+++ b/front/temporal/agent_loop/workflows.ts
@@ -47,6 +47,7 @@ import {
   ActivityCancellationType,
   CancellationScope,
   deprecatePatch,
+  log,
   patched,
   proxyActivities,
   proxySinks,
@@ -599,6 +600,17 @@ export async function runSandboxChildToolWorkflow({
           actionModelId,
         })
       );
+
+      // Preserve the recorded failed outcome when replaying histories created before terminal
+      // activity failures became action results.
+      if (patched("sandbox-child-tool-terminal-failure-as-result")) {
+        log.error("Sandbox child tool activity failed.", {
+          actionModelId,
+          error,
+        });
+        return;
+      }
+
       throw error;
     }
   } else {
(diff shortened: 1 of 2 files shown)


front/temporal/agent_loop/workflows.ts
```diff
@@ -599,6 +600,17 @@ export async function runSandboxChildToolWorkflow({
           actionModelId,
         })
       );
+
+      // Preserve the recorded failed outcome when replaying histories created before terminal
+      // activity failures became action results.
+      if (patched("sandbox-child-tool-terminal-failure-as-result")) {
```
davidebbo, line 606: Once you remove the patch, this basically just replaces the throw by a return, to avoid failing the workflow.

fontanierh (approved):
LGTM

PR #31627, opened 2026-09-01
Improve sign-in flow for shared Frames

flvndvd:
## Description

This PR fixes https://github.com/dust-tt/tasks/issues/10171.

Workspace-shared Frames return a 404 to unauthenticated users to prevent resource enumeration. This creates a poor experience for users who simply need to sign in. This PR detects that logged-out state and shows a sign-in prompt while preserving the existing 404 behavior for authenticated users without access.


fontanierh (approved):
LGTM

PR #31637, opened 2026-09-01
feat: fetch feedback counts from pg

sfriquet:
## Description

Refs https://github.com/dust-tt/tasks/issues/10262

- Stop aggregating feedback counts from ES in `fetchAgentOverview()`, instead, rely on the already existing `AgentMessageResource.getFeedbackCountForAssistants`.
- Added a shortcut `AgentMessageResource.getFeedbackCountForAssistant`.

There should be no behavior change.

## Tests

- Added test for `AgentMessageResource.getFeedbackCountForAssistant`

## Risk

Low. Safe to rollback until the ES index `front.agent_message_analytics` is removed.

## Deploy Plan

Front only

### DIFF
--- a/front/lib/resources/agent_message_feedback_resource.ts
+++ b/front/lib/resources/agent_message_feedback_resource.ts
@@ -361,6 +361,28 @@ export class AgentMessageFeedbackResource extends BaseResource<AgentMessageFeedb
     }[];
   }
 
+  static async getFeedbackCountForAssistant(
+    auth: Authenticator,
+    agentConfigurationId: string,
+    daysOld?: number
+  ): Promise<{ positive: number; negative: number }> {
+    const feedbackCounts = await this.getFeedbackCountForAssistants(
+      auth,
+      [agentConfigurationId],
+      daysOld
+    );
+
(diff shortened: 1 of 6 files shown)


front/lib/resources/agent_message_feedback_resource.ts
```diff
@@ -361,6 +361,28 @@ export class AgentMessageFeedbackResource extends BaseResource<AgentMessageFeedb
     }[];
   }
 
+  static async getFeedbackCountForAssistant(
+    auth: Authenticator,
+    agentConfigurationId: string,
+    daysOld?: number
+  ): Promise<{ positive: number; negative: number }> {
+    const feedbackCounts = await this.getFeedbackCountForAssistants(
```
aubin-tchoi, line 369: why not use the SQL query from the other PR?

aubin-tchoi (approved):
LGTM

front/lib/resources/agent_message_feedback_resource.ts
sfriquet: the other PR performs a group by this one is just a regular count

PR #31641, opened 2026-09-01
[core] Add diagnostic script for orphaned relational rows

tdraier:
## Description

Adds a read-only migration binary (`core/bin/migrations/20260901_find_orphan_relational_rows.rs`) that finds `tables`, `data_sources_folders` and (latest) `data_sources_documents` rows with **no matching `data_sources_nodes` row**.

Motivation: these orphans are why full data source deletion can fail (e.g. the Front workspace scrub). `DataSource::delete` enumerates tables/folders through `data_sources_nodes` (via an `INNER JOIN`), so a node-less row is never deleted and the final `DELETE FROM data_sources` fails on `tables_data_source_fkey` / `data_sources_folders_data_source_fkey`. Document orphans do **not** block deletion (`delete_data_source` sweeps all doc rows by `data_source`) but are surfaced for completeness — scoped to `status = 'latest'`, since superseded/deleted document rows legitimately have no node.

The orphans are a historical relic of the Nov–Dec 2024 introduction of `data_sources_nodes` and its one-shot backfill (`20241204_create_nodes.rs`); the current write paths create/delete the row and its node in a single transaction, so no new orphans are produced. This script quantifies how broadly the historical gap affected existing data sources.

Design notes:
- Batched with keyset pagination over bounded id windows, so no single statement touches more than `--batch-size` rows — a whole-table anti-join over `tables` / `data_sources_documents` would be far too heavy. Node membership is checked with `WHERE <fk_col> = ANY($ids)`, riding the existing FK indices.
- Read-only. Reports total orphans per table, count of affected data sources, and a per-data-source breakdown resolved to `project=… data_source_id=…`.
- The `[[bin]]` entry is kept commented out in `core/Cargo.toml`, matching the existing migration convention (uncomment to build/run).

Usage:
```
CORE_DATABASE_URI=... cargo run --bin find_orphan_relational_rows -- [--batch-size 5000] [--data-source <row_id>]
```

## Tests

- `cargo check --bin find_orphan_relational_rows` (compiles clean).
- `cargo fmt`.
- Not run against a live database in this change (no DB access from the authoring environment); it is read-only and intended to be run manually by an operator.

## Risk

Low. Read-only script, not compiled by default (bin entry commented out), no runtime/service impact. Each SQL statement is bounded to `--batch-size` rows.

## Deploy Plan

No deploy. Run ad-hoc against Core PG by uncommenting the `[[bin]]` entry and executing with `CORE_DATABASE_URI` set.

### DIFF
--- a/core/Cargo.toml
+++ b/core/Cargo.toml
@@ -109,6 +109,10 @@ path = "bin/salesforce.rs"
 # name = "migrate_slack_to_slack_tools_platform_actions"
 # path = "bin/migrations/20251208_migrate_slack_to_slack_tools_platform_actions.rs"
 
+# [[bin]]
+# name = "find_orphan_relational_rows"
+# path = "bin/migrations/20260901_find_orphan_relational_rows.rs"
+
 [[bin]]
 name = "check_table"
 path = "bin/check_table.rs"
--- a/core/bin/migrations/20260901_find_orphan_relational_rows.rs
+++ b/core/bin/migrations/20260901_find_orphan_relational_rows.rs
@@ -1,0 +1,279 @@
+use std::collections::{HashMap, HashSet};
+
+use anyhow::{anyhow, Result};
+use clap::Parser;
+use dust::stores::{postgres, store};
+
+/*
+ * Diagnostic for orphaned relational rows: `tables`, `data_sources_folders` and (latest)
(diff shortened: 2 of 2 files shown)


tdraier:

- #31641 ◀

PR #31644, opened 2026-09-01
[core] Add --delete to orphan relational rows script

tdraier:
## Description

Stacked on #31641. Adds a `--delete` mode to the orphan diagnostic that cleans up the **deletion-blocking** orphans (tables and folders) for a single data source, so a stuck data source scrub can complete (e.g. the Front workspace — 10 orphan folders + 1 orphan table on `data_sources.id in (21636, 862473)`).

Deletion reuses the production lifecycle rather than raw SQL:
- Orphan **tables** go through `Table::delete` → transient database invalidation + GCS CSV cleanup for local tables + transactional row/node delete. A minimal `Table` is reconstructed from the `tables` row; the node-derived display fields (title/mime_type/parents) are unknown for an orphan but are not read during delete. No `search_store` is passed — an orphan has no node and therefore no search-index entry.
- Orphan **folders** go through `Store::delete_data_source_folder` (already transactional).
- Document orphans are intentionally **not** deleted here — they do not block deletion and carry versioned GCS blobs.

Safety guards (orphan-row deletion is data loss, only appropriate for a data source being fully scrubbed — for a live data source the correct fix is to backfill the missing node, not delete the row):
- `--delete` requires `--data-source` (no global orphan deletion).
- Dry-run by default; only acts with `--execute`.

Usage:
```
# dry-run
cargo run --bin find_orphan_relational_rows -- --data-source 21636 --delete
# execute
cargo run --bin find_orphan_relational_rows -- --data-source 21636 --delete --execute
```

## Tests

- `cargo check --bin find_orphan_relational_rows` (compiles clean).
- `cargo fmt`.
- Not run against a live database from the authoring environment. Intended to be run manually by an operator, dry-run first.

## Risk

Medium when actually executed (`--delete --execute` deletes rows and, for local tables, their GCS data), mitigated by: single-data-source scoping, mandatory dry-run-by-default, and reuse of the existing deletion lifecycle (same code paths as normal table/folder deletion). Zero risk at rest — the bin stays commented out in `core/Cargo.toml`.

## Deploy Plan

No deploy. Run ad-hoc against Core PG by uncommenting the `[[bin]]` entry. For the stuck Front data sources: dry-run `--delete` for `21636` and `862473`, verify the counts, then re-run with `--execute`, then retry the scrub jobs.

### DIFF
--- a/core/bin/migrations/20260901_find_orphan_relational_rows.rs
+++ b/core/bin/migrations/20260901_find_orphan_relational_rows.rs
@@ -2,11 +2,17 @@ use std::collections::{HashMap, HashSet};
 
 use anyhow::{anyhow, Result};
 use clap::Parser;
-use dust::stores::{postgres, store};
+use dust::{
+    databases::table::Table,
+    databases_store::{self, gcs::GoogleCloudStorageDatabasesStore},
+    project::Project,
+    stores::{postgres, store},
+};
 
 /*
- * Diagnostic for orphaned relational rows: `tables`, `data_sources_folders` and (latest)
- * `data_sources_documents` rows that have no matching `data_sources_nodes` row.
+ * Diagnostic (and optional cleanup) for orphaned relational rows: `tables`,
+ * `data_sources_folders` and (latest) `data_sources_documents` rows that have no matching
+ * `data_sources_nodes` row.
  *
  * Why this matters:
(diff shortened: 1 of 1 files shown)


tdraier:

- #31641
    - #31644 ◀

PR #31646, opened 2026-09-01
Add: cell in webhook routing config

Fraggle:
## Description

Add `CELL`-based webhook routing configuration alongside the existing region mapping. `WebhookRouterConfigService` now writes `cells`, derives cell-to-region mappings from `CELL_TO_REGION`, and provides a backfill script for existing GCS entries; the Firebase sync path validates the expanded schema while retaining region compatibility.

## Tests

No automated tests were added. The backfill script supports dry-run validation before writing existing entries.

## Risk

Low-to-medium risk. Webhook synchronization now depends on a valid `CELL` environment variable, and malformed configuration can block routing. The legacy `regions` data remains populated, `cells` is backward-compatible and optional, and the backfill can be dry-run before execution.

## Deploy Plan

Deploy `connectors` with `CELL` configured to a supported value, then run `20260901_backfill_webhook_router_config_cells` in dry-run mode and execute it to backfill existing GCS entries. Deploy `firebase-functions/webhook-router` afterward. No SQL migration or infrastructure change is required.

### DIFF
--- a/connectors/migrations/20251120_migrate_slack_connection_credentials.ts
+++ b/connectors/migrations/20251120_migrate_slack_connection_credentials.ts
@@ -124,7 +124,7 @@ async function migrateSlackConnectionCredential(
 
   // Configure webhook router
   const webhookService = new WebhookRouterConfigService();
-  const currentRegion = connectorsConfig.getCurrentRegion();
+  const currentCell = connectorsConfig.getCurrentCell();
 
   // Get all connectors for this team to build the connector IDs list
   const allTeamSlackConfigs = await SlackConfigurationResource.listForTeamId(
@@ -137,12 +137,12 @@ async function migrateSlackConnectionCredential(
     PROVIDER,
     teamId,
     slackSigningSecret,
-    currentRegion,
+    currentCell,
     connectorIds
   );
 
   logger.info(
-    { teamId, region: currentRegion, connectorIds },
+    { teamId, cell: currentCell, connectorIds },
     `Updated webhook router configuration for team ${teamId} with ${connectorIds.length} connector(s).`
   );
 
--- a/connectors/src/connectors/shared/config.ts
+++ b/connectors/src/connectors/shared/config.ts
@@ -3,6 +3,13 @@ import { EnvironmentConfig } from "@connectors/types";
 export const SUPPORTED_REGIONS = ["europe-west1", "us-central1"] as const;
 export type RegionType = (typeof SUPPORTED_REGIONS)[number];
 
+export const SUPPORTED_CELLS = ["cell-00000", "cell-00001"] as const;
+export type CellType = (typeof SUPPORTED_CELLS)[number];
+export const CELL_TO_REGION: Record<CellType, RegionType> = {
+  "cell-00000": "us-central1",
+  "cell-00001": "europe-west1",
+};
+
 export const connectorsConfig = {
   getDustTmpSyncBucketName: (): string => {
     return EnvironmentConfig.getEnvVariable("DUST_TMP_SYNC_BUCKET_NAME");
@@ -15,5 +22,8 @@ export const connectorsConfig = {
   },
   getCurrentRegion: (): RegionType => {
     return EnvironmentConfig.getEnvVariable("REGION") as RegionType;
+  },
+  getCurrentCell: (): CellType => {
+    return EnvironmentConfig.getEnvVariable("CELL") as CellType;
   },
 };
--- a/connectors/src/lib/webhook_router_config.ts
+++ b/connectors/src/lib/webhook_router_config.ts
@@ -1,42 +1,35 @@
-import { connectorsConfig } from "@connectors/connectors/shared/config";
+import {
+  CELL_TO_REGION,
+  type CellType,
+  connectorsConfig,
+  SUPPORTED_CELLS,
+  SUPPORTED_REGIONS,
+} from "@connectors/connectors/shared/config";
 import logger from "@connectors/logger/logger";
 import { isDevelopment } from "@connectors/types";
 import { Storage } from "@google-cloud/storage";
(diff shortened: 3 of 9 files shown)


Fraggle:

- #31646 ◀

connectors/migrations/20251120_migrate_slack_connection_credentials.ts
```diff
@@ -124,7 +124,7 @@ async function migrateSlackConnectionCredential(
 
   // Configure webhook router
   const webhookService = new WebhookRouterConfigService();
-  const currentRegion = connectorsConfig.getCurrentRegion();
+  const currentCell = connectorsConfig.getCurrentCell();
```
davidebbo, line 127: Probably obsolete script, but doesn't hurt to fix it up.

connectors/src/connectors/shared/config.ts
```diff
@@ -3,6 +3,13 @@ import { EnvironmentConfig } from "@connectors/types";
 export const SUPPORTED_REGIONS = ["europe-west1", "us-central1"] as const;
 export type RegionType = (typeof SUPPORTED_REGIONS)[number];
 
+export const SUPPORTED_CELLS = ["cell-00000", "cell-00001"] as const;
+export type CellType = (typeof SUPPORTED_CELLS)[number];
+export const CELL_TO_REGION: Record<CellType, RegionType> = {
```
davidebbo, line 8: I don't have all the context, but first reaction is that ideally, the list of cell would not be hard coded.

davidebbo: Or maybe at least derived from the one in lib/api/cells/config.ts?

connectors/src/lib/webhook_router_config.ts
```diff
@@ -1,42 +1,35 @@
...
-export interface WebhookRouterEntry {
-  signingSecret: string;
-  regions: {
-    [region: string]: number[]; // region name -> connector IDs
-  };
-}
+const WebhookRouterEntrySchema = z.object({
+  signingSecret: z.string(),
+  regions: z.record(z.enum(SUPPORTED_REGIONS), z.array(z.number())),
+  cells: z.record(z.enum(SUPPORTED_CELLS), z.array(z.number())).optional(),
```
davidebbo, line 18: Why optional? And why have both regions and cells, which seems redundant.

Fraggle: Because of the migration:
- Step 1, double write, do not fail parsing if missing cells.
- Step 2, run migration script.
- Step 3, start reading from cells data.
- Step 4, drop regions from format & file.

connectors/src/connectors/shared/config.ts
Fraggle: Can't do, connectors do not have access to front/

connectors/src/lib/webhook_router_config.ts
```diff
@@ -328,10 +329,73 @@ export class WebhookRouterConfigService {
...
+          }
+        }
+      }
+    }
+
+    if (entriesUpdated === 0) {
+      return { entriesUpdated: 0 };
+    }
+
+    const config = WebhookRouterConfigSchema.parse(raw);
```
davidebbo, line 379: Risk of dropping unknown fields that are not in the schema. Though hopefully, there are no such fields, so probably ok!

Fraggle: I checked the file :)

davidebbo (approved):
LGTM. Biggest FUD is having both regions and cells side by side in the router entries, but maybe I don't understand how it's used.

connectors/src/connectors/shared/config.ts
flvndvd: To me they should only know about their own cell, through the env var added from the pod tag.

connectors/src/lib/webhook_router_config.ts
```diff
@@ -278,12 +266,18 @@ export class WebhookRouterConfigService {
...
+          // No connectors for this cell - remove the region & cell
           if (existingEntry) {
             delete existingEntry.regions[region];
+            if (existingEntry.cells) {
+              delete existingEntry.cells[cell];
+            }
 
-            // If no regions left, delete the entire entry
-            if (Object.keys(existingEntry.regions).length === 0) {
+            // If no regions & cells left, delete the entire entry
```
flvndvd, line 276: To be honest, we could keep an empty array.

connectors/src/lib/webhook_router_config.ts
```diff
@@ -328,10 +329,73 @@ export class WebhookRouterConfigService {
...
+    if (!exists) {
+      return { entriesUpdated: 0 };
+    }
+
+    const [contents] = await file.download();
+    const [metadata] = await file.getMetadata();
+    const raw = JSON.parse(contents.toString("utf-8"));
+
+    let entriesUpdated = 0;
+    if (typeof raw === "object" && raw !== null) {
```
flvndvd, line 358: Why not using the schema to parse it here?

flvndvd, line 358: Not clear what this first loop does 🤔. Can't we do one simple loop that adds the cell projection?

Fraggle: Overlooked.

Fraggle: Yep, weird.

connectors/src/connectors/shared/config.ts
davidebbo: Yes, for connectors, this would make sense. For firebase, it obviously needs to know about all.

davidebbo: Though that could be a configuration file that comes from GCS?

PR #31649, opened 2026-09-01
[Admin Governance] Add agent permission resource

philipperolet:
## Description

Context: https://app.notion.com/p/dust-tt/Design-Doc-Admin-Governance-38a28599d941817292d0e94f8dfafe48

Today, editor access for an agent is represented by membership in a dedicated group. The governance migration will instead store editor access in `group_permissions`. A permission must target the logical agent, while each saved version currently has a different `agent_configurations.id`.

This PR introduces `AgentResource`, a common interface for checking whether a user may read an agent, edit it, or manage its editors. For custom agents, permissions target the new stable `agents.id` shared by all versions.

The resource represents the current permission behavior:

- workspace admins can read any custom agent and manage its editors, but must add themselves as an editor before changing the agent;
- the user who created a custom agent keeps the same access while the legacy permission checks are being migrated;
- users with an editor grant can read, edit, and manage editors for that agent;
- agents shared with the workspace can be read by all workspace members, but sharing alone does not grant edit access;
- built-in global agents cannot be edited, and can be read by the roles configured for each agent: everyone, managers and admins, or admins only.

This PR only adds the permission abstraction and its tests. Existing agent endpoints do not use it yet, so it does not change end-user behavior by itself.

## Risks

Blast radius: the new agent permission abstraction and the existing global-agent visibility check; no production agent endpoint uses `AgentResource` yet.
Risk: low

## Deploy Plan

- deploy front; no migration required

## Tests

- [x] `cd front && NODE_ENV=test npm test lib/resources/agent_resource.test.ts lib/api/assistant/global_agents/global_agent_metadata.test.ts`

### DIFF
--- a/front/lib/resources/agent_resource.ts
+++ b/front/lib/resources/agent_resource.ts
@@ -1,0 +1,134 @@
+import { globalAgentReaderRoles } from "@app/lib/api/assistant/global_agents/global_agent_metadata";
+import type { Authenticator } from "@app/lib/auth";
+import type { AgentConfigurationModel } from "@app/lib/models/agent/agent";
+import type { AgentConfigurationScope } from "@app/types/assistant/agent";
+import type { GLOBAL_AGENTS_SID } from "@app/types/assistant/assistant";
+import type { GrantVerb } from "@app/types/group_permissions";
+import type {
+  AccessControlList,
+  RoleGrant,
+  WithAccessControl,
+} from "@app/types/resource_permissions";
+import type { ModelId } from "@app/types/shared/model_id";
+import { assertNever } from "@app/types/shared/utils/assert_never";
+
+// Legacy `canEdit` also allows changing the editor set, so the author fallback mirrors the full
+// editor role rather than granting write alone.
+const AGENT_EDITOR_VERBS: GrantVerb[] = ["read", "write", "admin"];
+
+const HIDDEN_AGENT_ROLE_GRANTS: RoleGrant[] = [
+  { role: "admin", permissions: AGENT_EDITOR_VERBS },
+];
+
+const VISIBLE_AGENT_ROLE_GRANTS: RoleGrant[] = [
+  ...HIDDEN_AGENT_ROLE_GRANTS,
+  { role: "manager", permissions: ["read"] },
+  { role: "builder", permissions: ["read"] },
+  { role: "user", permissions: ["read"] },
+  { role: "none", permissions: ["read"] },
+];
+
+type CustomAgentAccess = {
+  kind: "custom";
+  agentModelId: ModelId;
+  authorModelId: ModelId;
+  scope: Exclude<AgentConfigurationScope, "global">;
+};
+
+type GlobalAgentAccess = {
+  kind: "global";
+  agentId: GLOBAL_AGENTS_SID;
+};
+
+type AgentAccess = CustomAgentAccess | GlobalAgentAccess;
+
+type AgentConfigurationAccess = Pick<
+  AgentConfigurationModel,
+  "agentId" | "authorId" | "sId" | "scope" | "workspaceId"
+>;
+
+export class AgentResource implements WithAccessControl {
+  private constructor(
+    readonly id: ModelId | null,
+    readonly sId: string,
+    readonly workspaceId: ModelId,
+    private readonly access: AgentAccess
+  ) {}
+
+  static fromAgentConfiguration(
+    configuration: AgentConfigurationAccess
+  ): AgentResource {
+    return new AgentResource(
+      configuration.agentId,
+      configuration.sId,
+      configuration.workspaceId,
+      {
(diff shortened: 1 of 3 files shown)


front/lib/resources/agent_resource.ts
```diff
@@ -0,0 +1,134 @@
...
+            })),
+            workspaceId: this.workspaceId,
+          },
+        ];
+      default:
+        return assertNever(this.access);
+    }
+  }
+
+  canRead(auth: Authenticator): boolean {
```
philipperolet: no need for canXXX helpers

front/lib/resources/agent_resource.ts
```diff
@@ -0,0 +1,134 @@
...
+} from "@app/types/resource_permissions";
+import type { ModelId } from "@app/types/shared/model_id";
+import { assertNever } from "@app/types/shared/utils/assert_never";
+
+// Legacy `canEdit` also allows changing the editor set, so the author fallback mirrors the full
+// editor role rather than granting write alone.
+const AGENT_EDITOR_VERBS: GrantVerb[] = ["read", "write", "admin"];
+
+const HIDDEN_AGENT_ROLE_GRANTS: RoleGrant[] = [
+  { role: "admin", permissions: AGENT_EDITOR_VERBS },
```
philipperolet: admin has read permission on hidden agents, but not sure they have write? IIRC we have a text asking them to become editor when they need to make a change on an agent they're not an editor of, check that. And explain why admins were considered to have "write" & admin access?
However they can add themselves to editor so that'd be an admin right? check that too

front/lib/resources/agent_resource.ts
```diff
@@ -0,0 +1,134 @@
...
+
+const VISIBLE_AGENT_ROLE_GRANTS: RoleGrant[] = [
+  ...HIDDEN_AGENT_ROLE_GRANTS,
+  { role: "manager", permissions: ["read"] },
+  { role: "builder", permissions: ["read"] },
+  { role: "user", permissions: ["read"] },
+  { role: "none", permissions: ["read"] },
+];
+
+type CustomAgentAccess = {
```
philipperolet: let's not create those intermediary types
down the road, the agentResource will become big. We should do the access check on this.id, this.kind, etc. not introduce an intermediary abstraction. We should follow resource patterns, look at recent new resources and write this one in an idomatic way for this codebase

philipperolet: Removed the intermediary access types. `AgentResource` now stores `id`, `sId`, `workspaceId`, `kind`, `authorId`, and `scope` directly, and `getAccessControlLists` switches on `this.kind` before checking the other fields.

front/lib/resources/agent_resource.ts
philipperolet: Confirmed: `BuilderEditorGateMessage` tells workspace admins they can view the agent but must become an editor before saving. The existing agent-editor group ACL follows the same rule. I changed the admin role to `read + admin`, without `write`; `admin` lets them manage the editor list and add themselves. The test and PR description now make this distinction explicit.

front/lib/resources/agent_resource.ts
philipperolet: Removed the three wrappers. Callers and tests now use `Authenticator.hasPermission` directly with the resource.

front/lib/resources/agent_resource.ts
```diff
@@ -0,0 +1,119 @@
...
+
+const VISIBLE_AGENT_ROLE_GRANTS: RoleGrant[] = [
+  ...HIDDEN_AGENT_ROLE_GRANTS,
+  { role: "manager", permissions: ["read"] },
+  { role: "builder", permissions: ["read"] },
+  { role: "user", permissions: ["read"] },
+  { role: "none", permissions: ["read"] },
+];
+
+export class AgentResource implements WithAccessControl {
```
tdraier, line 33: 🎉

front/lib/resources/agent_resource.ts
```diff
@@ -0,0 +1,119 @@
...
+          },
+        ];
+      case "custom": {
+        assert(this.id !== null);
+        assert(this.authorId !== null);
+
+        const grants = auth.getGrantedVerbs("agent", this.id);
+        const isAuthor =
+          auth.workspace()?.id === this.workspaceId &&
+          auth.user()?.id === this.authorId;
```
tdraier, line 100: this isAuthor looks awkward, author is stored in agent_configuration and is different from every version. I also don't understand why this author has special permissions, I thought we were only relying on agent_editors ?

philipperolet: yes, it's to comply with a legacy behaviour. It will be removed by PR 17 :) The idea was to separate the risky refactor from the change of behaviour by removing legacy author fallback.

tdraier:
✅

PR #31652, opened 2026-09-01
[front] fix: redirect legacy analytics page

aubin-tchoi:
## Description

The legacy analytics URL currently opens the legacy page. This PR changes that to instead redirect to the new `/analytics/consumption` page. We don't remove it entirely because some users may have kept bookmarks or shared direct links internally.

A follow up PR will remove the front-end component.

## Tests

- Tested locally.

## Risk

- Low.

## Deploy Plan

- Deploy front.

### DIFF
--- a/front-spa/src/app/routes/adminRoutes.tsx
+++ b/front-spa/src/app/routes/adminRoutes.tsx
@@ -4,10 +4,6 @@ import { withSuspense } from "@spa/app/routes/withSuspense";
 import type { RouteObject } from "react-router-dom";
 import { Navigate } from "react-router-dom";
 
-const AnalyticsPage = withSuspense(
-  () => import("@dust-tt/front/components/pages/workspace/AnalyticsPage"),
-  "AnalyticsPage"
-);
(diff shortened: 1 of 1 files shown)


front-spa/src/app/routes/adminRoutes.tsx
adrsimon, line 124: couldn't it be the other way around ? a redirect from /analytics/consumption to /analytics ? and have the main page hosted at /analytics which would be its final URL

adrsimon (approved):
lgtm code wise, will let u advise on the question

flvndvd (approved):

front-spa/src/app/routes/adminRoutes.tsx
aubin-tchoi: I think that in the comms we sent we put `analytics/consumption`. not a big issue but weird looks to switch it so soon. definitely worth doing in a few weeks that said

flvndvd: Yep we indeed sent `analytics/consumption`

PR #31654, opened 2026-09-01
feat: move version-markers to new es index

sfriquet:
## Description

Refs https://github.com/dust-tt/tasks/issues/10185

Migrate `/api/w/:wId/assistant/agent_configurations/:aId/observability/version-markers` to `searchConsumptionAnalytics`.

## Tests

- added tests
- tested locally by navigating to an agent > insights > feedback > by version and see the dropdown correctly populated

## Risk

Low. Safe to rollback while the legacy es index still exists.

## Deploy Plan

Front only


aubin-tchoi (approved):
LGTM

PR #31657, opened 2026-09-01
[Email Agents] Retry transient regional relays

philipperolet:
## Description

Fixes https://github.com/dust-tt/tasks/issues/10266.

Production logs show that the source region correctly tried to relay the reported email to the region containing the workspace, but that request returned a transient 502. With only one relay attempt, the email was not processed and the source region sent its local workspace error.

Retry a cross-region relay up to two times after a network or server error. The receiving region records the existing email `Message-ID` for five minutes before processing and returns success without processing the same email twice.

## Risks

Blast radius: inbound Email Agents requests that require a cross-region relay.
Risk: low. During a rolling `front-api` deployment, a new source pod could retry against an old target pod that does not yet deduplicate by `Message-ID`, which could process one email up to three times. We accept this risk because it requires both a cross-region email and a transient relay failure during the short mixed-version window, and inbound email volume is low at the planned deployment time.

## Deploy Plan

- deploy `front-api` normally in both regions during low inbound email traffic
- monitor regional email relay errors after deployment

## Test

- [ ] `NODE_ENV=test npm -w front-api test -- routes/email/webhook.test.ts` (local hive test services are unavailable; covered by CI)


front/lib/api/assistant/email/webhook_helpers.ts
```diff
@@ -45,6 +47,12 @@ export const EMAIL_WEBHOOK_RELAY_SOURCE_ERROR_HEADER =
   "x-dust-email-webhook-source-error";
 export const EMAIL_WEBHOOK_RELAY_HEADER_VALUE = "1";
 
+const EMAIL_RELAY_KEY_PREFIX = "email-webhook-relay";
+const EMAIL_RELAY_DEDUPE_TTL_SECONDS = 60 * 60;
```
philipperolet: so 1 hour? isn't that very long? especially if we retry only once?

front/lib/api/assistant/email/webhook_helpers.ts
```diff
@@ -45,6 +47,12 @@ export const EMAIL_WEBHOOK_RELAY_SOURCE_ERROR_HEADER =
...
+const EMAIL_RELAY_MAX_RETRIES = 1;
```
philipperolet: why only 1 retry and not 3 as is common?

front-api/routes/email/webhook.ts
```diff
@@ -141,6 +142,17 @@ app.post("/", async (ctx): HandlerResult<PostResponseBody> => {
 
   const email = emailRes.value;
 
+  if (
+    isRelayRequest &&
+    !(await recordEmailRelay(email.threadingHeaders.messageId))
```
philipperolet, line 147: I thought we used message id to answer new emails from the same thread in the same dust conversation;  are we sure this won't mess with that logic?  is it true that Message-Id is different for each new email trigger?

front/lib/api/assistant/email/webhook_helpers.ts
```diff
@@ -193,22 +221,53 @@ export async function relayEmailToOtherRegion(
...
+          );
+        }
+        return response;
       },
-      body: formData,
-    });
+      {
+        maxRetries: email.threadingHeaders.messageId
+          ? EMAIL_RELAY_MAX_RETRIES
+          : EMAIL_RELAY_NO_RETRIES,
```
philipperolet: in which cases don't we have a messageId?

davidebbo (commented):
LGTM

PR #31658, opened 2026-09-01
Add direct Frames v2 function calls

fontanierh:
## Description

Add `dsbx frame call` so agents can invoke a named function from a registered Frame's active publication.

The preferred form takes the stable Frame ID and checks Frame use rights directly. For authoring convenience, the existing source-folder or manifest-path form remains supported and additionally requires read access to that mounted source.

Both forms reuse the existing active-publication resolver and sandbox function invocation path. The private action-token endpoints are behind `frames_v2` and ignored by Swagger.

Bump dsbx to 0.1.55 and dust-base to 0.8.104 so the command ships in the sandbox image.

## Tests

Added functional coverage for calls by Frame ID and source path, active-publication scoping, runtime errors, JSON input, Frame-ID validation, and the full CLI argument contract for both target forms.

## Risk

Low. This adds private action-token endpoints and a CLI command behind `frames_v2`; existing Frame and Pod Function invocation paths are unchanged.

## Deploy Plan

1. Deploy Front with both private invocation endpoints.
2. Release dsbx 0.1.55.
3. Build and release dust-base 0.8.104.
4. Use `/poke/kill` to replace sandboxes running older dust-base versions.

### DIFF
--- a/cli/dust-sandbox/Cargo.lock
+++ b/cli/dust-sandbox/Cargo.lock
@@ -313,7 +313,7 @@ checksum = "92773504d58c093f6de2459af4af33faa518c13451eb8f2b5698ed3d36e7c813"
 
 [[package]]
 name = "dust-sandbox"
-version = "0.1.54"
+version = "0.1.55"
 dependencies = [
  "anyhow",
  "base64",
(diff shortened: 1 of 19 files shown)


davidebbo (commented):
LGTM

PR #31660, opened 2026-09-02
[Email Agents] Retry transient regional relay failures

philipperolet:
## Description

Fixes https://github.com/dust-tt/tasks/issues/10266.

Stacked on https://github.com/dust-tt/dust/pull/31657.

Production logs for the reported email show that the US lookup did attempt the EU relay. That relay returned a transient 502 and had no matching target-region receipt; a relay for the same sender had succeeded nine minutes earlier.

Retry a cross-region email relay once after a network or server failure, but only when the target explicitly confirms that it did not receive the first request. Once receipt is confirmed, the source treats the target as the owner and never replays the email because Email Agents processing is not idempotent.

## Risks

Blast radius: inbound Email Agents requests that require a cross-region relay.
Risk: low after the dependency is fully deployed; retries require an explicit negative receipt from the target.

## Deploy Plan

- do not merge or deploy this PR until #31657 is deployed to every `front-api` pod in both regions
- after that rollout is confirmed, deploy normally to both regions

## Test

- [ ] `NODE_ENV=test npm -w front-api test -- routes/email/webhook.test.ts` (local shared test containers unavailable; covered by CI)
- [x] `npx tsgo --noEmit -p front/tsconfig.json`
- [x] `npx tsgo --noEmit -p front-api/tsconfig.json`
- [x] `npm -w front-api run lint:swagger-annotations`

### DIFF
--- a/front-api/routes/email/webhook.test.ts
+++ b/front-api/routes/email/webhook.test.ts
@@ -140,6 +140,84 @@ describe("POST /api/email/webhook", () => {
     }
   });
 
+  it("retries a relay after a transient server error", async () => {
+    const { user } = await createResourceTest({ role: "admin" });
+    const fetchMock = vi
+      .fn()
+      .mockResolvedValueOnce(new Response(null, { status: 502 }))
+      .mockResolvedValueOnce(
+        Response.json({ received: false }, { status: 200 })
+      )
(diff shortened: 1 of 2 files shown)


philipperolet:
Superseded by #31657, which now contains the simpler Message-ID deduplication and retry implementation.

PR #31662, opened 2026-09-02
[front] Handle 422 duplicate uniqueness_key on seat edit as idempotent success

tdraier:
## Description

Follow-up hotfix to #31578. That PR added a fresh per-edit `uniqueness_key` on
the seat-count `v2.contracts.edit` so that an SDK auto-retry (after a lost/slow
response) can't blindly re-apply the cumulative seat delta and stack it (the
548-phantom-unassigned-seat incident). The retry-after-apply is rejected by
Metronome as a duplicate key, and the catch treated that rejection as an
idempotent success.

The catch only handled `ConflictError` (409) — the status the docs imply. In
production the Metronome SDK actually maps a duplicate `uniqueness_key` to a
**422 `UnprocessableEntityError`** ("Uniqueness key already exists: <key>"), so
the rejection fell through to `Err` and surfaced as a hard sync failure on the
`metronome-seat-count-sync` workflow (workspace `HzXDVQGUQF`).

The Metronome audit log for that run confirms the mechanism: the edit
**succeeded** server-side at 20:07:03, the SDK retried the same key ~29s later,
and Metronome correctly rejected the duplicate (422) at 20:07:32 — the
idempotency guard doing its job. Only our 409-vs-422 mismatch turned that
successful edit into a reported failure.

This change accepts the 422 (and 409 defensively) as an idempotent success via a
small `isDuplicateUniquenessKeyError` guard, gated on the
`"uniqueness key already exists"` message so genuine 422 validation errors still
surface as errors. Stacking remains impossible regardless: Metronome rejects the
duplicate key at the server level, independent of how we classify the error.

## Tests

- `front/lib/metronome/client.test.ts`: the duplicate-key test now throws a 422
  `UnprocessableEntityError` with the real message; added a defensive 409 case
  and a "non-duplicate 422 surfaces as an error" case. 25/25 pass.
- `npx tsgo --noEmit` clean.

## Risk

Low. Narrows a false-failure path to success; the message guard prevents
swallowing real 422 validation errors. No change to the happy path. The
anti-stacking guarantee is unaffected (server-side duplicate-key rejection).

## Deploy Plan

Standard `front` deploy. Resolves an active production error on the
`metronome-seat-count-sync` workflow introduced by the already-merged #31578.

### DIFF
--- a/front/lib/metronome/client.test.ts
+++ b/front/lib/metronome/client.test.ts
@@ -39,9 +39,13 @@ const {
   mockContractsEdit,
   mockSetCustomFieldValues,
   MockConflictError,
+  MockUnprocessableEntityError,
 } = vi.hoisted(() => {
   class MockConflictError extends Error {
     status = 409;
+  }
+  class MockUnprocessableEntityError extends Error {
+    status = 422;
   }
   return {
     mockCreate: vi.fn(),
@@ -51,6 +55,7 @@ const {
     mockContractsEdit: vi.fn(),
     mockSetCustomFieldValues: vi.fn(),
     MockConflictError,
+    MockUnprocessableEntityError,
   };
 });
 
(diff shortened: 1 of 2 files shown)


tdraier:

- #31662 ◀

avervaet (approved):
LGTM

PR #31664, opened 2026-09-02
[front] chore: remove components from legacy Analytics page

aubin-tchoi:
## Description

Follow-up to #31652. Since `/analytics` now redirects to the consumption analytics page, the old workspace analytics page is unreachable.

This PR remove the legacy page and the components that were only used there.

## Tests

- Tested locally.

## Risk

- Low.

## Deploy Plan

- Deploy front.


flvndvd (approved):
\o/

PR #31665, opened 2026-09-02
style(analytics): clarify unit used on graphs and cards

achilleburah:
## Description

**Units on the analytics**
Feedback from users that they were confused about the analytics because the unit is unclear (they were trying to guess if the numbers were credits, tokens, messages, conversations).

Added "credits" units in the first card and update graph title.

image

## Tests

Previewed locally. 

## Risk

None. Wording only.

## Deploy Plan

Normal deploy.


ClementAupiais (approved):
LGTM

PR #31666, opened 2026-09-02
fix(pricing): add long-context billing tiers for GPT 5.6 and Gemini Pro

Nils-Fedrigo:
## Description

Several models were billed at a single flat rate even though their providers price prompts above a
context threshold at a higher tier for the whole request.

OpenAI GPT 5.6 (Sol, Terra, Luna) had no `long_context` entry, so prompts above 272K input tokens
were billed at the short-context rate. Added the published long-context tiers (2x input / cached
input / cache writes, 1.5x output), matching the tables at
https://developers.openai.com/api/docs/pricing:

| model | short (in/cached/write/out) | long (in/cached/write/out) |
| --- | --- | --- |
| gpt-5.6-sol | 4 / 0.4 / 5 / 20 | 8 / 0.8 / 10 / 30 |
| gpt-5.6-terra | 2 / 0.2 / 2.5 / 12 | 4 / 0.4 / 5 / 18 |
| gpt-5.6-luna | 0.2 / 0.02 / 0.25 / 1.2 | 0.4 / 0.04 / 0.5 / 1.8 |

Sol, Terra and Luna are all configured with a 272K context window today, so the new tier is not
reachable through those model IDs — it makes the table correct if the window is ever raised, and
keeps the family consistent with `gpt-5.6-terra-long-context`, `gpt-5.5` and `gpt-5.4`, which
already carry the tier.

Gemini 3 Pro and 3.1 Pro were priced flat at the *high* tier (4/18) with a comment noting the real
structure is 2/12 up to 200K input tokens and 4/18 above. Both are configured with a 1M context
window, so this over-billed every prompt below 200K tokens. They now use the tiered entry, so
short-context prompts are billed at 2/12 and long-context ones stay at 4/18.

EU pricing picks up all of the new tiers automatically through `applyRegionalUplift`.

Not changed: `gpt-5.4-mini`, `gpt-5.4-nano` and the Flash/Flash-Lite families have no long-context
SKU, and `gpt-5.5-pro` / `gpt-5.4-pro` are not in our model list.

## Tests

Added threshold-boundary cases in `token_pricing/index.test.ts` for Sol, Terra, Luna, Gemini 3 Pro
and Gemini 3.1 Pro (one at the threshold, one just above), and updated the Sol pricing-sync
assertion in `gpt_five_dot_six.test.ts`.

## Risk

Gemini 3 Pro / 3.1 Pro prompts under 200K input tokens are now billed at half the input rate and
2/3 of the output rate they were before — that is the correction, but it does lower recorded cost
for those models going forward. Rollback is a revert of this commit; no data migration involved.

## Deploy Plan

Nothing specific — pricing is read from code at usage-computation time.

### DIFF
--- a/front/lib/api/assistant/token_pricing/global.ts
+++ b/front/lib/api/assistant/token_pricing/global.ts
@@ -27,6 +27,13 @@ const CURRENT_MODEL_PRICING: Record<StaticModelIdType, PricingEntry> = {
     output: 20.0,
     cache_creation_input_tokens: 5.0,
     cache_read_input_tokens: 0.4,
+    long_context: {
+      prompt_token_threshold: 272_001,
+      input: 8.0,
+      output: 30.0,
+      cache_creation_input_tokens: 10.0,
+      cache_read_input_tokens: 0.8,
+    },
   },
   // https://openai.com/api/pricing
   "gpt-5.6-terra": {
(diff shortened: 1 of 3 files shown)


fontanierh (approved):
LGMT thanks for fixing

flvndvd (commented):
LGTM

PR #31667, opened 2026-09-02
feat: key the Slack workflow summoning whitelist on spaces

adrsimon:
## Description

The slack workflow whitelist stores group ids, so `whitelist-bot --groupId` can name any group at all. It causes a few problem, for example, a group later attached to a new space silently widens an already-allowed workflow.

It now stores the space ids support/admins picks, and connectors asks front which groups represent those spaces at run time. A space is represented by its own `regular_auto` member group, the Company Space by the workspace global group, and that mapping reaches connectors through `GET /api/v1/w/:wId/spaces/groups`, a system-key-only endpoint. Connectors caches the answer in redis for 5 minutes, keyed on the whitelist row, so a summon does not pay a front round trip every time.

#31701 added the column. This one writes space ids and keeps reading group ids for rows written before it. #31668 drops the group ids once every row has spaces.

`20260901_backfill_slack_workflow_spaces.ts` records, for the workspaces named in a file of ids, the spaces a row's groups stand for.

## Tests

Checked the production whitelist before writing the backfill: 115 global and 64 `regular_auto` groups map mechanically to spaces. 7 slack workflows are whitelisted to non `regular_auto` groups. Checked by hand, those give no particular space grant, so ignoring them loses no specific reach.

Endpoint tests cover the system-key requirement and the space to group mapping, including the Company Space.

## Risk

Rollback is a revert. Rows written with space ids keep their group ids, which the reverted code reads.

Editing a workflow's reach drops its cache entry, so it takes effect right away. A change to a space's own membership takes up to 5 minutes, the redis TTL.

A workflow whose spaces have all been deleted now fails to summon instead of resolving to no group. An empty `X-Dust-Group-Ids` reads as the whole workspace to a system key, so failing closed is the point.

## Deploy Plan

1. Deploy front and connectors in any order. No row carries space ids until the new front writes one, and an old connectors ignores `space_ids` while front still sends `group_ids`.
2. Run the backfill per region, dry-run first: `npx tsx front/migrations/20260901_backfill_slack_workflow_spaces.ts --filePath <ids>`


connectors/src/connectors/slack/bot.ts
```diff
@@ -773,6 +773,35 @@ async function processErrorResult(
   }
 }
 
+// A whitelisted workflow reaches the spaces it was allowed on. Front owns which group stands for a
+// space, so ask it at run time instead of keeping group ids here.
+async function getWorkflowGroupIds(
+  connector: ConnectorResource,
+  slackConfig: SlackConfigurationResource,
+  botName: string
```
flvndvd: `nit`

```suggestion
  {botName}: {botName: string}
```

connectors/src/resources/slack_configuration_resource.ts
```diff
@@ -244,8 +254,9 @@ export class SlackConfigurationResource extends BaseResource<SlackConfigurationM
     });
   }
 
-  // Get the Dust group IDs that the bot is whitelisted for.
-  async getBotGroupIds(botName: string): Promise<string[]> {
+  // What the bot is whitelisted for: spaces, or group ids for a row written before the whitelist
+  // moved to spaces.
+  async getBotReach(botName: string): Promise<BotReachType | null> {
```
flvndvd: `nit` it's usually call a "principal".

front/lib/api/spaces.ts
```diff
@@ -65,6 +72,55 @@ import { Op, UniqueConstraintError } from "sequelize";
...
+      const memberGroup = space.isGlobal()
+        ? globalGroup
+        : memberGroupByGrantKey.get(grantKey(spaceMemberGrant(space)));
+
+      return memberGroup ? { space, memberGroup } : null;
+    })
+  );
+}
+
+export async function listSpaceMemberGroupIds(
```
flvndvd: This name is very confusing, member is somewhat of a reserved word to users. Also reading the code it only gives back the global group, right?

flvndvd: Surprised that this is not an instance method on the resource or at least `listSpaceMemberGroups` should be.

sdks/js/src/index.ts
```diff
@@ -1908,6 +1909,24 @@ export class DustAPI {
     return new Ok(r.value.apps);
   }
 
+  async getSpaceGroupIds({ spaceIds }: { spaceIds: string[] }) {
```
flvndvd: Looks like it only return the global group

adrsimon: hmmm no ? spaces/groups returns every group 
can u elaborate ?

adrsimon: as per irl we return every auto group as it's intended, renamed eevrything to make it clearer

connectors/src/connectors/slack/bot.ts
```diff
@@ -773,6 +776,54 @@ async function processErrorResult(
...
+async function resolveWorkflowGroupIds(
+  connector: ConnectorResource,
+  entry: SlackBotWhitelistEntry
+): Promise<Result<string[], Error>> {
+  const { whitelistModelId, groupIds, spaceIds } = entry;
+  if (!spaceIds) {
+    return new Ok(groupIds);
+  }
+
+  try {
```
flvndvd: No need for a try/catch here. DustAPI already gets you covered.

connectors/src/connectors/slack/bot.ts
```diff
@@ -773,6 +776,54 @@ async function processErrorResult(
...
+    return new Ok(
```
flvndvd: Why not just returning`await getAutoGroupIdsForSpaces()`?

connectors/src/connectors/slack/bot.ts
```diff
@@ -943,7 +994,13 @@ async function answerMessage(
     if (!botName) {
       throw new Error("Failed to get bot name. Should never happen.");
     }
-    requestedGroups = await slackConfig.getBotGroupIds(botName);
+    const groupIdsRes = await getWorkflowGroupIds(connector, slackConfig, {
+      botName,
+    });
+    if (groupIdsRes.isErr()) {
+      return new Err(groupIdsRes.error);
```
flvndvd:
```suggestion
      return groupIdsRes;
```

connectors/src/lib/slack/space_auto_groups.ts
```diff
@@ -0,0 +1,55 @@
...
+  const dustAPI = new DustAPI(
+    { url: apiConfig.getDustFrontAPIUrl() },
+    { workspaceId, apiKey: workspaceAPIKey },
+    logger
+  );
+
+  const groupIdsRes = await dustAPI.getAutoGroupIdsForSpaces({ spaceIds });
+  if (groupIdsRes.isErr()) {
+    // Thrown so the cache keeps no failure, turned back into an Err by the caller.
+    throw new Error(groupIdsRes.error.message);
```
flvndvd: 😱

connectors/src/lib/slack/space_auto_groups.ts
```diff
@@ -0,0 +1,55 @@
...
+  }
+
+  return groupIdsRes.value;
+}
+
+const autoGroupIdsCacheKey = (whitelistModelId: ModelId) =>
+  `${whitelistModelId}`;
```
flvndvd: mind adding a few more tokens here? At least something like `auto-group-ids-${whitelistModelId}`

flvndvd: Forget about this, `cacheWithRedis` already does it.

front-api/routes/poke/workspaces/[wId]/slack-workflows.test.ts
```diff
@@ -48,6 +51,13 @@ async function setupTest({
   return setup;
 }
 
+async function createSpaceWithMemberGroup(workspace: LightWorkspaceType) {
+  const auth = await Authenticator.internalAdminForWorkspace(workspace.sId);
+  const space = await SpaceFactory.regular(workspace);
+
+  return { space, memberGroup: await space.fetchManualMemberGroup(auth) };
```
Comment (flvndvd) [inline front-api/routes/poke/workspaces/[wId]/slack-workflows.test.ts]: `await` in return 🤮

front-api/routes/v1/w/[wId]/spaces/groups.test.ts
```diff
@@ -0,0 +1,86 @@
...
+    ]);
+
+    expect(response.status).toBe(200);
+    expect(await response.json()).toEqual({ groupIds: [memberGroup.sId] });
+  });
+
+  it("returns 400 without space ids", async () => {
+    const { workspace, key } = await setupTest();
+
+    const response = await getAutoGroupIdsForSpaces(workspace, key);
```
Comment (flvndvd) [inline front-api/routes/v1/w/[wId]/spaces/groups.test.ts]: This API feels misleading, the function's name is `ForSpaces` but spaces is optional 🤯

front/lib/resources/space_resource.ts
```diff
@@ -1766,6 +1787,33 @@ export class SpaceResource extends BaseResource<SpaceModel> {
     return autoGroups.filter((group) => !editorGroupIds.has(group.sId));
   }
 
+  // The group that stands for a space: its own regular_auto member group, which lives and dies with
+  // the space and cannot be attached to another one. The Company Space has no such group — everyone
+  // in the workspace belongs to it — so the workspace global group stands for it. A provisioned
+  // group would drift the moment an admin attaches it elsewhere.
```
flvndvd: ✂️

front/lib/resources/space_resource.ts
```diff
@@ -1766,6 +1787,33 @@ export class SpaceResource extends BaseResource<SpaceModel> {
...
+    auth: Authenticator,
+    spaces: SpaceResource[]
+  ): Promise<SpaceAutoGroup[]> {
+    const [autoGroupByGrantKey, globalGroupRes] = await Promise.all([
+      GroupPermissionResource.findRegularAutoGroupsForGrants(auth, {
+        grants: spaces.filter((space) => !space.isGlobal()).map(memberGrant),
+      }),
+      GroupResource.fetchWorkspaceGlobalGroup(auth),
+    ]);
+    const globalGroup = globalGroupRes.isOk() ? globalGroupRes.value : null;
```
flvndvd: I don't see a world where a workspace can be functional on Dust if it's missing a global group, right? Can we enforce this here as well.

front/lib/resources/space_resource.ts
```diff
@@ -1766,6 +1787,33 @@ export class SpaceResource extends BaseResource<SpaceModel> {
...
+
+    return removeNulls(
+      spaces.map((space) => {
+        const autoGroup = space.isGlobal()
+          ? globalGroup
+          : autoGroupByGrantKey.get(grantKey(memberGrant(space)));
```
flvndvd: Don't we have a contract where each space **has at least** one auto group? If so, let's assert this here, over swallowing errors.

front/lib/api/slack/summoning_whitelist.ts
```diff
@@ -123,40 +133,54 @@ export async function listSlackWorkflows(
...
+  const [autoGroupIdBySpaceId, globalGroupRes, globalSpace] = await Promise.all(
+    [
+      SpaceResource.listAutoGroupIdsBySpaceId(
+        auth,
+        spaces.map((space) => space.sId)
+      ),
+      GroupResource.fetchWorkspaceGlobalGroup(auth),
+      SpaceResource.fetchWorkspaceGlobalSpace(auth),
+    ]
+  );
```
adrsimon, line 154: the fetch then map to sID is transitory and is only here to double write spaces column and groups column

this will disappear in 31668

connectors/src/resources/slack_configuration_resource.ts
```diff
@@ -15,14 +16,66 @@ import type {
...
+): Promise<Result<string[], Error>> {
+  const dustAPI = new DustAPI(
+    { url: apiConfig.getDustFrontAPIUrl() },
+    { workspaceId, apiKey: workspaceAPIKey },
+    logger
+  );
+
+  const groupIdsRes = await dustAPI.getAutoGroupIdsForSpaces({ spaceIds });
+  if (groupIdsRes.isErr()) {
+    return new Err(new Error(groupIdsRes.error.message));
```
flvndvd, line 47:
```suggestion
    return groupIdsRes;
```

connectors/src/resources/slack_configuration_resource.ts
```diff
@@ -15,14 +16,66 @@ import type {
...
+  }
+
+  return new Ok(groupIdsRes.value);
```
flvndvd:
```suggestion
  return groupIdsRes;
```

front/lib/resources/space_resource.ts
```diff
@@ -1766,6 +1782,40 @@ export class SpaceResource extends BaseResource<SpaceModel> {
...
   }
 
+  static async listAutoGroupIdsBySpaceId(
+    auth: Authenticator,
+    spaceIds: string[]
+  ): Promise<Map<string, string>> {
+    const spaces = await SpaceResource.fetchByIds(auth, spaceIds);
+    // System and conversations spaces have no auto group of their own.
+    const membershipSpaces = spaces.filter(
+      (space) => space.isGlobal() || space.isRegular() || space.isProject()
```
flvndvd: Would be nice to align on the UI logic, which might be include all spaces **BUT** system?

flvndvd (approved):

connectors/src/resources/slack_configuration_resource.ts
adrsimon: not doing the err part we receive an API error not an error 👍

PR #31668, opened 2026-09-02
feat: drop the group ids of Slack workflows

adrsimon:
## Description

Cleanup step after #31667.

Spaces are the only thing the whitelist needs now that reach is resolved from them at run time, so the stored group ids go. A workflow left with no space now fails instead of summoning, since an empty group header reads as the whole workspace. `whitelist-bot --groupId` becomes `--spaceIds` for `summon_agent`.
It writes straight to the whitelist, so it validates no space, adds no Company Space and emits no audit event, unlike `allowSlackWorkflow`. The automations page and the Poke plugin stay the way to allow a workflow.

The backfill script from #31667 is kept commented out rather than deleted, as the record of how the spaces of already-allowed workflows were derived.

The `groupIds` column stays for now. Dropping it will come in a follow-up.

## Tests

Poke slack-workflows suite.

## Risk

This is what makes un-migrated rows fail, so it must not merge before the backfill from the previous PR has run in both regions.

## Deploy Plan

1. Confirm no `slack_bot_whitelist` row with `whitelistType = 'summon_agent'` still has a null `spaceIds`, in EU and US.
2. Deploy connectors before front.


front/lib/api/slack/summoning_whitelist.ts
```diff
@@ -141,28 +127,15 @@ export async function allowSlackWorkflow(
...
-  );
-  if (globalGroupRes.isErr()) {
-    return new Err(
-      new SlackWorkflowError(
-        "invalid_groups",
-        "Failed to fetch the workspace group."
-      )
-    );
-  }
+  const [spaces, globalSpace] = await Promise.all([
```
aubin-tchoi: we can do this one sequentially

aubin-tchoi (approved):
LGTM

front/migrations/20260901_backfill_slack_workflow_spaces.ts
```diff
@@ -1,236 +1,239 @@
-import { createReadStream } from "fs";
-import { createInterface } from "readline";
+// Ran once per region before the group ids it reads were dropped, and kept commented for the
```
flvndvd: Let's delete it, no need to keep commented code

flvndvd (approved):

PR #31670, opened 2026-09-02
feat: move agent_configurationt endpoint to new es index

sfriquet:
## Description

Refs https://github.com/dust-tt/tasks/issues/10185

- `/api/w/:wId/assistant/agent_configurations/:aId/observability/overview` used to call `fetchAgentOverview` and `fetchAgentCostStats` which call `searchAnalytics`.
- Turns out this endpoint is only used to display feedback count, so removed those calls from this endpoint
- Turns out `fetchAgentCostStats` result is never used, so removed this function entirely
- Migrated `fetchAgentOverview` which is use elsewhere to `searchConsumtptionAnalytics`

## Tests

- added tests for `fetchAgentOverview`
- lint + browse agent config locally to check nothing is broken

## Risk

Low. Safe to rollback while the es index `front.agent_message_analytics` exists

## Deploy Plan

Front only

### DIFF
--- a/front/lib/api/assistant/observability/overview.ts
+++ b/front/lib/api/assistant/observability/overview.ts
@@ -1,8 +1,13 @@
-import { buildAgentAnalyticsBaseQuery } from "@app/lib/api/assistant/observability/utils";
-import type { ElasticsearchError } from "@app/lib/api/elasticsearch";
-import { bucketsToArray, searchAnalytics } from "@app/lib/api/elasticsearch";
+import { resolveConsumptionPeriod } from "@app/lib/api/analytics/consumption/period";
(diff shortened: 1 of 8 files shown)


front/lib/api/assistant/observability/overview.ts
```diff
@@ -20,140 +25,68 @@ type OverviewAggs = {
 };
 
 export async function fetchAgentOverview(
```
aubin-tchoi, line 27: outside the scope of the PR but this should not be in the `/observability` folder since it's not used for observability anymore

aubin-tchoi (approved):
LGTM

PR #31672, opened 2026-09-02
[front] chore: remove role switchers from user menu

aubin-tchoi:
## Description

The buttons "Become Admin/Manager/User" are currently visible but the corresponding API calls will always fail since we can't have super admin sessions in the app anymore.

Since it was also a goal to implement a proper role separation on our workspace, this PR hides the buttons outside of a dev env.

Roles can still be changed from poke.

## Tests

- Tested locally.

## Risk

- Low.

## Deploy Plan

- Deploy front.


Fraggle (approved):
LGTM ! Thanks

PR #31682, opened 2026-09-02
[front] feat: add message consumption breakdown in Poke

aubin-tchoi:
## Description

We used to have some indications of the cost on a message from the Poke conversation view.
This was dropped in https://github.com/dust-tt/dust/pull/31642, it relied on the legacy message analytics index.

This PR reintroduces a breakdown that relies on the same data as the in-app message breakdown (pg consumption items table).
Displays a few more details that are hidden in-app to ease the debugging.

image

## Tests

- Tested locally.

## Risk

- Low.

## Deploy Plan

- Deploy front.


front/components/poke/conversation/message_consumption_inspector.tsx
```diff
@@ -0,0 +1,541 @@
...
+            )}
+          </div>
+        </div>
+        <p className="shrink-0 text-xs font-medium tabular-nums text-muted-foreground">
+          {formatShare(tool.attributedCredits, totalCredits)}
+        </p>
+      </div>
+      <dl className="grid grid-cols-2 gap-2 border-t border-border pt-3">
+        <div>
+          <dt className="text-xs text-muted-foreground">Tokens</dt>
```
avervaet: I thinks those are credits not tokens as it comes from tool.attributedCredits

avervaet (approved):

front/components/poke/conversation/message_consumption_inspector.tsx
aubin-tchoi: meant "credits from tokens" but that's indeed very unclear

aubin-tchoi: replacing with "input/output"

PR #31684, opened 2026-09-02
[front] rename input bar to composer in UI and use curly quote

ykmsd:
## Description
We want to say composer instead of input bar for user facing communication so I'm renaming it 🫡 (will do the same for the doc)

Also I use curly quote for greeting messages 🫡


matteotrab (approved):

ClementAupiais (approved):
LGTM

PR #31693, opened 2026-09-02
[front] usage: open the page to all workspaces, hide credit plan sections for the others

PopDaph:
The Usage page is now available to every workspace. The sidebar entry no longer depends on the plan or the `usage_page_read_only` flag, which is removed. We needed it to allow user change the Model Tiers.

Workspaces that are not on a credit plan get the page without:
- the Workspace credit pool block (title, counter, progress bar),
- the Seat and Credits usage columns, the seat filter, and the seat / spend limit row actions,
- the credit-only settings cards (default spend limit, programmatic limit, notifications) and the top-ups tab.

Model tiers are editable for them: per member, per group, and the workspace default in Settings. Credit actions (top up, invite, seat changes, spend limits) stay disabled as before.

The credit pool summary and seat plan endpoints are no longer called for those workspaces.

To test: check on Dust then on your perso workspace on free-friends family

### DIFF
--- a/front/components/pages/workspace/UsagePage.tsx
+++ b/front/components/pages/workspace/UsagePage.tsx
@@ -32,11 +32,7 @@ import {
   formatConsumptionDate,
 } from "@app/lib/analytics/consumption_period";
 import type { MemberUsageType } from "@app/lib/api/credits/members_usage";
-import {
-  useAuth,
-  useFeatureFlags,
-  useWorkspace,
-} from "@app/lib/auth/AuthContext";
+import { useAuth, useWorkspace } from "@app/lib/auth/AuthContext";
 import { formatCredits } from "@app/lib/client/credits";
 import type { UserModelTierSelection } from "@app/lib/client/model_tier_options";
 import { INHERIT_MODEL_TIER } from "@app/lib/client/model_tier_options";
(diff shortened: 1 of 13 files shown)


front/types/shared/feature_flags.ts
```diff
@@ -183,12 +183,6 @@ export const WHITELISTABLE_FEATURES_CONFIG = {
     stage: "self_serve",
     owner: "flvndvd",
   },
-  usage_page_read_only: {
```
PopDaph, line 186: Confirmed with Thomas we can kill

front/components/pages/workspace/UsagePage.tsx
```diff
@@ -1038,7 +987,8 @@ export function UsagePage() {
       creditsResetAt={creditsResetAt}
       isLoading={isMembersUsageLoading}
       isRefreshing={isMembersUsageRefreshing}
-      readOnly={isReadOnly}
+      readOnly={!isCreditPriced}
```
aubin-tchoi: why are legacy plans in readonly here?

front/components/pages/workspace/UsagePage.tsx
```diff
@@ -1038,7 +987,8 @@ export function UsagePage() {
...
+      showSeatAndCredits={isCreditPriced}
       seatActionsDisabled={isSubscriptionCancelled}
       showSpendLimit={!isFreePlanWorkspace}
```
aubin-tchoi: shouldn't we hide spent limits for legacy plans? (unless spend limits have nothing to do with credits, honestly don't know)

aubin-tchoi (approved):
LGTM

PR #31686, opened 2026-09-02
feat: avoid fetching regular_auto groups with fetchById

matteotrab:
## Description

This PR makes sure that the `fetchById` method in groupResource is not being used to fetch regular_auto groups. This is needed as permissions will be removed from regular_auto groups. Regular_auto groups should be fetched only from their resources with `fetchRegularAutoGroupForResource` or dangerously fetched from poke.

## Tests

Manually.

## Risks

Limited to Poke group lookup and display paths.

## Deploy Plan

Standard deployment.

PR #31685, opened 2026-09-02
fix: display company data in api keys table

matteotrab:
## Description

This PR fixes the API keys table to display the global company space alongside each key’s explicitly scoped spaces.

Capture d’écran 2026-09-02 à 14 15 17

## Tests

Manually.

## Risks

Low.

## Deploy Plan

Standard deployment.

PR #31704, opened 2026-09-02
feat: remove builder group sync

matteotrab:
## Description

This PR removes the sync of the dust-builder provisioned group with the Builders regular_manual group. Additionally all references to the dust-builder provisioned group are removed as it is now a classic provisioned group with no special behavior.

## Tests
Manually.

## Risk
Low. Members of the dust-builders provisioned group are not added automatically anymore to the Builders regular_manual group. We did not communicate that we would keep this group in sync, we just told admins on July 22 that 

> Existing builders will be migrated to a dedicated Builders group with the corresponding governance permissions

## Deploy Plan

Standard deployment.


matteotrab:

- #31461
    - #31464
        - #31563
            - #31565
                - #31612
                    - #31685
                        - #31686
                            - #31704 ◀
                                - #31726

front-api/routes/w/[wId]/provisioning-status.ts
```diff
@@ -21,7 +20,6 @@ app.get("/", async (ctx): HandlerResult<GetProvisioningStatusResponseBody> => {
   return ctx.json({
     hasAdminGroup: groups.some((g) => g.name === ADMIN_GROUP_NAME),
     hasManagerGroup: groups.some((g) => g.name === MANAGER_GROUP_NAME),
-    hasBuilderGroup: groups.some((g) => g.name === BUILDER_GROUP_NAME),
```
Comment (philipperolet) [inline front-api/routes/w/[wId]/provisioning-status.ts:24]: nit: legacy client bundles might fail temporarily, but not very important since the route is likely seldom used + refresh solves it.

front/lib/api/permissions/governance_seeding.ts
philipperolet: nit: maybe rephrase this comment to avoid "builder"?

philipperolet (approved):
Thanks 👍 Left 2 nits

front-api/routes/w/[wId]/provisioning-status.ts
Comment (matteotrab) [reply front-api/routes/w/[wId]/provisioning-status.ts:24]: In theory they should not fail because they'll receive undefined instead of false and in the way they use it it should not make a difference. In any case I agree that it's seldom used and easily solvable

front/lib/api/permissions/governance_seeding.ts
matteotrab: We can even remove this seeder completely as when a capability is granted to admins-only we don't do anything

PR #31709, opened 2026-09-02
[front] feat: flag over-allocated seats in metronome seat-state audit

tdraier:
## Description

Adds a read-only over-allocation (credit-stacking) detection pass to `front/scripts/audit_metronome_seat_state.ts`.

For every assigned seat, it compares the AWU `starting_balance` (total granted this cycle, independent of consumption) against the seat type's contract allocation (`getAwuAllocationForSeatType`), and logs any seat granted more than its allocation under `"[SeatAudit] over-allocated seats (credit stacking)"` with the affected seat count, total over-granted AWU, and per-seat detail.

Context: when a member changes seat type (e.g. pro→max) during a seat-sync failure/retry storm, the seat assignment can converge on the new type while the origin (pro) seat credit is never emptied — leaving the pro 8000 grant stacked on top of the max 40000 (`starting_balance` 48000). Once the seat has converged, `syncSeatCount`'s `emptyOriginSeatCreditsForTransfers` sees no transfer candidate and can no longer self-heal it, so the stacked credit is invisible to the existing seat-assignment audit. Using `starting_balance` (not the live balance) makes detection exact regardless of how much the user has already consumed.

## Tests

Manual: ran against the affected Enterprise workspace in production (read-only). The existing seat-assignment diff showed a fully-converged workspace (0 stale / 0 missing across pro/max/free), confirming the stacking is only visible at the credit-ledger level — which this new pass surfaces.

No automated tests: this is an operator diagnostic script (`makeScript`), consistent with the rest of the file. It makes only READ calls to Metronome and the DB; there is nothing to `--execute`.

## Risk

Very low. Read-only additive change to a diagnostic script; no production code path imports it. Metronome reads reuse the existing rate-limited (`paceMetronome`) balance fetch already performed by the audit — no additional API calls. Trivially safe to roll back.

## Deploy Plan

No deploy required — operator script run on demand via `npx tsx front/scripts/audit_metronome_seat_state.ts --workspaceId <wId>`. Merges with the next `front` release.

### DIFF
--- a/front/lib/metronome/stacked_seat_credits.ts
+++ b/front/lib/metronome/stacked_seat_credits.ts
@@ -38,16 +38,26 @@ import { Err, Ok } from "@app/types/shared/result";
  * each individual materialized credit id. We map each materialized credit id to a
  * seat type once per contract, then for each seat:
  *
- *  - empty every STRAY credit fully (a credit whose seat type ≠ the seat's tier),
- *  - carry the consumption that was charged to those strays onto the HOME credit,
- *    so the seat nets `homeAllocation − totalConsumed` instead of a full home
- *    allowance plus forgiven stray usage.
+ *  - empty every STRAY credit fully (a credit whose seat type ≠ the seat's tier);
+ *  - drive the HOME credit to `max(0, homeAllocation − currentPeriodUsage)`, using
+ *    Metronome's authoritative per-user AWU usage.
  *
- * Without the carry, consumption billed to the stray credit is forgiven when the
- * stray is emptied and the seat ends over-credited by that amount. Both figures
- * are exact — observed per-seat credit balances vs. deterministic tier
- * allocations, not an ES/usage estimate. Idempotent by construction: a re-run
- * reads every stray at 0 (no stray consumption left to carry) and emits nothing.
+ * The home step is what keeps the seat correct: consumption billed to a stray is
+ * otherwise forgiven when the stray is emptied, leaving the seat over-credited by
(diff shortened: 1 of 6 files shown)


tdraier:

- #31709 ◀
    - #31932

front/lib/metronome/stacked_seat_credits.ts
```diff
@@ -0,0 +1,499 @@
...
+ * `credit.segment.start`-triggered reconcile go through this module so detection
+ * can never drift between the one-off cleanup and the ongoing guard.
+ *
+ * The key primitive is the per-seat `credits[]` breakdown on `MetronomeSeatBalance`
+ * (populated with `include_credits_and_commits`): it gives each seat's balance on
+ * each individual materialized credit id. We map each materialized credit id to a
+ * seat type once per contract, then for each seat:
+ *
+ *  - empty every STRAY credit fully (a credit whose seat type ≠ the seat's tier);
+ *  - drive the HOME credit to `max(0, homeAllocation − currentPeriodUsage)`, using
```
fabiencelier, line 42: what does "HOME" credit mean ?

fabiencelier (approved):

front/lib/metronome/stacked_seat_credits.ts
tdraier: "home" seat is the one where we are right now, home credit is the credit which matches this "home" seat.
so it adjust the current credit, and empty all others

PR #31710, opened 2026-09-02
[front] feat: add max seats column to poke switch-contract form

tdraier:
## Description

Adds a **Max** (hard assignment cap) column beside the seat commitment floor in the poke **Switch contract** dialog, and renames the **Min seats** header to **Commitment** to match the existing "Seat commitments and limit" modal.

`maxSeats` is threaded through `SwitchContractBodySchema` (new optional field) and persisted server-side via `WorkspaceSeatLimitResource.upsert`, so it writes the same open-ended `workspace_seat_limits` row that seat assignment already reads. The persist condition now also fires when only a cap is set (`minSeats === 0 && maxSeats != null`) so a max-only cap isn't dropped.

No new enforcement logic: the cap is already honored on assignment — `updateMembershipSeatAndTrack` hard-blocks a seat-type change once `currentCount >= maxSeats` (`seat_limit_reached` → 400), and new-member/invite resolution falls back to free/none rather than exceeding the cap.

## Risk

Low. Poke-only admin form. Additive optional schema field (no breaking change to the private API). Existing `maxSeats >= minSeats` validation in `WorkspaceSeatLimitResource` surfaces as a switch-contract error if violated.

## Tests

- `npx tsgo --noEmit` passes.
- `biome check --changed` clean.
- Manual verification of enforcement paths (existing behavior unchanged).

## Deploy Plan

Standard deploy. No migration, no config, no feature flag.

### DIFF
--- a/front/types/poke/switch_contract.ts
+++ b/front/types/poke/switch_contract.ts
@@ -52,12 +52,13 @@ const recurringFreeCreditSchema = z
   .min(1, "Recurring free credit must be at least 1 credit");
 
 // Per-seat-type settings for a contract switch, keyed by seat type (see
-// `SwitchContractBodySchema.seats`). `minSeats` is the billing floor
-// persisted to `workspace_seat_limits`. `rate` is the per-seat rate in the
-// currency's MAJOR units (dollars / euros); the server converts it to
-// Metronome's fiat unit via `metronomeAmount`. When `commitmentPrice` is set
-// (also in major units), a contract prepaid commit is created granting
-// `minSeats * rate` of contract credit, invoiced at `commitmentPrice`.
+// `SwitchContractBodySchema.seats`). `minSeats` is the billing floor and
+// `maxSeats` the hard assignment cap (omitted = no cap), both persisted to
+// `workspace_seat_limits`. `rate` is the per-seat rate in the currency's MAJOR
+// units (dollars / euros); the server converts it to Metronome's fiat unit via
+// `metronomeAmount`. When `commitmentPrice` is set (also in major units), a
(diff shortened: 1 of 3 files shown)


tdraier:

- #31710 ◀

front/types/poke/switch_contract.ts
```diff
@@ -52,19 +52,21 @@ const recurringFreeCreditSchema = z
...
+// contract prepaid commit is created granting `minSeats * rate` of contract
+// credit, invoiced at `commitmentPrice`.
 const seatEntrySchema = z.object({
   // Whether the seat is entitled on the new contract. `true` (the default,
   // for backward compatibility) entitles and configures the seat; `false`
   // disables a seat the package would otherwise sell. The dialog submits
   // every known seat so deselections can be turned into disable overrides.
   selected: z.boolean().default(true),
   minSeats: z.number().int().min(0, "Min seats must be ≥ 0"),
+  maxSeats: z.number().int().min(1, "Max seats must be ≥ 1").optional(),
```
PopDaph: Could we validate the relationship between these fields in the Zod schema?

```ts
    maxSeats: z.number().int().min(1, "Max seats must be ≥ 1").optional(),
  }).refine(
    ({ minSeats, maxSeats }) =>
      maxSeats === undefined || maxSeats >= minSeats,
    {
      path: ["maxSeats"],
      message: "Max seats must be greater than or equal to commitment",
    }
  );
```

Right now { minSeats: 10, maxSeats: 5 } passes both the form and endpoint validation it seems, and fails at the resource level!

PopDaph (approved):

PR #31712, opened 2026-09-02
Drop NEXT_PUBLIC_ENABLE_BOT_CRAWLING from the front images

flvndvd:
## Description

The `front-api` and `front-workers` images take `NEXT_PUBLIC_ENABLE_BOT_CRAWLING` as a build arg and bake it into the runtime env, but nothing running in those images reads it. The only readers are the marketing sitemap config and the marketing document head, and the marketing image already gets the value from its own Dockerfile. This removes the four dead lines.

Small on its own, this is the first step towards building the backend once and shipping the same image to every cell. The US and EU front images differ today only by a handful of baked `NEXT_PUBLIC_*` values, and this flag was one of them: true in the US config, absent in the EU one. The US config keeps the entry because marketing still needs it, so the front build will still receive it as a build arg and Docker will print an unused-arg warning. Harmless, and it goes
away once the build config is split from marketing.

## Risk

None expected, the variable was never read by these processes. Rollback is fine.

## Deploy Plan

Ships with the next regular front deploy. Nothing to do in dust-infra.


davidebbo (approved):
LGTM

PR #31715, opened 2026-09-02
[front] Create WorkOS organization for every workspace

Fraggle:
## Description

Create a WorkOS organization for every workspace in `createWorkspaceInternal` and the admin `workspace create` command, regardless of plan or signup path. Refresh the returned workspace so immediate membership creation can sync to WorkOS, while preserving the existing soft-failure behavior when WorkOS is unavailable. Add a dry-run backfill for legacy workspaces, excluding `relocation-done` stubs and supporting per-cell execution and bounded concurrency.

## Tests

Added Vitest coverage for organization creation on new workspaces and the non-blocking WorkOS failure path in `front/lib/iam/workspaces.test.ts`.

## Risk

Medium risk from increased WorkOS API traffic and the backfill operating across existing workspaces. Creation remains idempotent and non-blocking; the backfill defaults to dry-run and skips relocated stubs. Rollback is a code revert; created WorkOS organizations require separate cleanup if needed.

## Deploy Plan

Deploy `front` (the `front-api` diff only updates comments). Run the backfill dry-run per cell, review its counts, then rerun with `--execute` as needed:

````bash
npx tsx front/migrations/20260902_backfill_workos_organizations.ts --execute
````

No SQL migration or infrastructure change.

### DIFF
--- a/front/admin/cli.ts
+++ b/front/admin/cli.ts
@@ -8,6 +8,7 @@ import {
 } from "@app/lib/api/data_sources";
 import { garbageCollectGoogleDriveDocument } from "@app/lib/api/poke/plugins/data_sources/garbage_collect_google_drive_document";
 import { getLlmCredentials } from "@app/lib/api/provider_credentials";
+import { getOrCreateWorkOSOrganization } from "@app/lib/api/workos/organization";
 import { Authenticator } from "@app/lib/auth";
 import { getModelConfigByModelId } from "@app/lib/llms/model_configurations";
 import { FREE_UPGRADED_PLAN_CODE } from "@app/lib/plans/plan_codes";
@@ -71,6 +72,14 @@ const workspace = async (command: string, args: parseArgs.ParsedArgs) => {
         systemGroup,
         globalGroup,
       });
+
+      const orgRes = await getOrCreateWorkOSOrganization(lightWorkspace);
+      if (orgRes.isErr()) {
+        logger.error(
+          { error: orgRes.error, workspaceId: w.sId },
+          "Failed to create WorkOS organization during workspace creation"
+        );
+      }
 
       args.wId = w.sId;
       return;
--- a/front/admin/cli.ts
+++ b/front/admin/cli.ts
@@ -8,18 +8,16 @@ import {
 } from "@app/lib/api/data_sources";
 import { garbageCollectGoogleDriveDocument } from "@app/lib/api/poke/plugins/data_sources/garbage_collect_google_drive_document";
 import { getLlmCredentials } from "@app/lib/api/provider_credentials";
-import { getOrCreateWorkOSOrganization } from "@app/lib/api/workos/organization";
 import { Authenticator } from "@app/lib/auth";
+import { createWorkspaceInternal } from "@app/lib/iam/workspaces";
 import { getModelConfigByModelId } from "@app/lib/llms/model_configurations";
 import { FREE_UPGRADED_PLAN_CODE } from "@app/lib/plans/plan_codes";
 import { ConversationResource } from "@app/lib/resources/conversation_resource";
 import { DataSourceResource } from "@app/lib/resources/data_source_resource";
-import { GroupResource } from "@app/lib/resources/group_resource";
 import { KeyResource } from "@app/lib/resources/key_resource";
 import { LabsTranscriptsConfigurationResource } from "@app/lib/resources/labs_transcripts_resource";
 import { MembershipResource } from "@app/lib/resources/membership_resource";
 import { SpaceResource } from "@app/lib/resources/space_resource";
-import { generateRandomModelSId } from "@app/lib/resources/string_ids_server";
 import { SubscriptionResource } from "@app/lib/resources/subscription_resource";
 import { UserResource } from "@app/lib/resources/user_resource";
 import { WebhookRequestResource } from "@app/lib/resources/webhook_request_resource";
@@ -30,7 +28,6 @@ import {
   getWebhookRequestPayloadFromGCS,
   processWebhookRequest,
 } from "@app/lib/triggers/webhook";
-import { renderLightWorkspaceType } from "@app/lib/workspace";
 import logger from "@app/logger/logger";
 import { launchScrubSpaceWorkflow } from "@app/poke/temporal/client";
 import {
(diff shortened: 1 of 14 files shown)

### CODE REVIEW

File: front/admin/cli.ts [hunk 1]
```diff
@@ -72,6 +73,14 @@ const workspace = async (command: string, args: parseArgs.ParsedArgs) => {
         globalGroup,
       });
 
+      const orgRes = await getOrCreateWorkOSOrganization(lightWorkspace);
```
flvndvd: Don't we have a canonical way to create a workspace? No matter if it's coming from the admin/or a new user logging-in?

front/admin/cli.ts
```diff
@@ -72,6 +73,14 @@ const workspace = async (command: string, args: parseArgs.ParsedArgs) => {
...
+      if (orgRes.isErr()) {
```
flvndvd: IMHO this should hardly fail if we plan to rely on it to route traffic.

front/lib/api/workos/get_or_create_organization.test.ts
```diff
@@ -0,0 +1,178 @@
...
+
+    // Ensure factory workspace has no WorkOS org id.
+    await WorkspaceResource.updateWorkOSOrganizationId(workspace.id, null);
+    const lightWorkspace = {
+      ...renderLightWorkspaceType({ workspace }),
+      workOSOrganizationId: null,
+    };
+
+    mockGetOrganizationByExternalId.mockRejectedValue(
+      Object.assign(new Error("not found"), {
```
flvndvd: Really no need for the `Object.assign` here

front/lib/api/workos/organization.ts
```diff
@@ -22,6 +23,44 @@ import {
...
+
+  await concurrentExecutor(
+    memberships,
+    async (membership) => {
+      if (!membership.user?.workOSUserId) {
+        return;
+      }
+
+      await MembershipResource.updateWorkOSMembershipRole({
+        user: new UserResource(UserModel, membership.user),
```
flvndvd, line 48: Sad that we need this

front/lib/api/workos/organization.ts
```diff
@@ -32,7 +71,14 @@ export async function getOrCreateWorkOSOrganization(
       return new Err(organizationRes.error);
     }
 
-    let organization = organizationRes.value;
+    const existingOrganization = organizationRes.value;
```
flvndvd: Not sure this first assignment brings lots of value.

front/lib/api/workos/organization.ts
```diff
@@ -84,6 +105,13 @@ export async function getOrCreateWorkOSOrganization(
       );
     }
 
+    if (shouldSyncMemberships) {
+      await syncActiveMembershipsToWorkOSOrganization({
+        workspace,
+        organizationId: organization.id,
```
flvndvd: Why passing `organizationId` here, workspace is already up-to-date at this point. It feels confusing to update it internally in `syncActiveMembershipsToWorkOSOrganization` then.

front/lib/iam/workspaces.ts
```diff
@@ -107,26 +103,32 @@ export async function createWorkspaceInternal({
...
+      });
     }
   }
 
-  return workspace;
+  // Every workspace gets a WorkOS organization (SSO/SCIM/domains/audit).
+  // Failures are logged but do not block workspace creation — callers can
+  // still complete signup, and getOrCreateWorkOSOrganization is idempotent.
+  const orgRes = await getOrCreateWorkOSOrganization(lightWorkspace);
+  if (orgRes.isErr()) {
```
flvndvd: Current approach that does not fail, has basically no recover path which isn't ideal. Why not making it bubble up the error?

front/admin/cli.ts
Fraggle: This is iso with past behavior but I also wondered (i checked, no logs in the past 2 weeks).
Can change it but as we already have a workspace at that point so we would fail for the user but no guarantee that all workspaces in the DB will have a workOS org.

What I planned to change is to not have any fallback in the lookup method and fail hard if no workOS org.

front/lib/api/workos/organization.ts
Fraggle: Yep...

front/admin/cli.ts
Fraggle: @flvndvd WDYT ?

front/admin/cli.ts
Fraggle: createWorkspaceInternal(), not sure why it isn't used here.

Fraggle: Looking

front/lib/api/workos/organization.ts
Fraggle: Fair

front/lib/iam/workspaces.ts
Fraggle: See my message above, the workspace would already exists, the user would have an error without abilty to do anything.

Solution is to just run the backfill script on our side.

front/admin/cli.ts
flvndvd: I'd also enforce it here.

front/lib/iam/workspaces.ts
```diff
@@ -78,25 +74,62 @@ export async function createWorkspaceInternal({
...
+      auth,
+      {
+        systemGroup,
+        globalGroup,
+      },
+      transaction
+    );
 
-  const lightWorkspace = renderLightWorkspaceType({ workspace });
+    const orgRes = await getOrCreateWorkOSOrganization(lightWorkspace, {
```
flvndvd: Hum, can we avoid doing a HTTP call while holding the SQL transaction? If WorkOS struggles we end up with idle connections.

Fraggle: I need the workspace ID i think to create the org, so it needs to be created in the DB, which means I need to be in the transaction. Or I have to write reversal code.
I think it's not worth it and will just make sure that workOS request has a short timeout.

id13: If it’s from a front api pod, let’s do it without a transaction and properly handle work os failure / retries. 

If it’s from a temporal workflow though… not ideal, requires a comment above to say this is strongly against best practices but could be done since blast radius is much smaller

id13 (approved):
lgtm

PR #31716, opened 2026-09-02
Build connectors once, push to the global registry

flvndvd:
## Description

Connectors is built twice per deploy today, once per region, with the two legs pushing identical images to two regional registries. This adds a single build job for connectors that authenticates against the dust-infra project and pushes one image to the new global `dust-images` repository from https://github.com/dust-tt/dust-infra/pull/1010. The regional build matrix is skipped for connectors and unchanged for every other component.

For now the same build is also tagged into both regional registries, through a new `extra_image_registries` input on the build-image action. That keeps the regional values files in dust-infra, `revert.yml` and `list-tags` working without any ordering constraints. The regional tags go away once dust-infra points the connectors values at the global repository.

Connectors goes first because its Dockerfile takes no region-specific build args. The front images still bake three region-specific URLs into the runtime env, which is why they cannot follow yet.

## Tests

I gave up testing GH action a long time ago 🙈 

## Risk

Only the connectors component is affected. If the new job fails, the deploy job does not run and the currently deployed image stays. Rollback is reverting this PR, which puts connectors back on the regional matrix.

## Deploy Plan

1. dust-infra applies the registry, the build identity and the pull grants (https://github.com/dust-tt/dust-infra/pull/1010).
2. Merge this PR.
3. Deploy connectors as usual. The image lands in the global registry and both regional ones under the same tag.

### DIFF
--- a/.github/actions/build-image/action.yml
+++ b/.github/actions/build-image/action.yml
@@ -51,6 +51,10 @@ inputs:
   contentful_access_token:
     description: "Contentful access token"
     required: false
+  extra_image_registries:
+    description: "Space-separated <host>/<project>/<repo> prefixes that also receive the image, on top of the primary registry"
+    required: false
+    default: ""
 
 outputs:
   current_version:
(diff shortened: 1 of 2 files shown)


id13 (approved):

flvndvd:
Build worked as expected see [here](https://console.cloud.google.com/artifacts/docker/dust-infra/us-central1/dust-images/connectors?hl=en&project=dust-infra)

PR #31717, opened 2026-09-02
ci: trigger the SPA preview deploy from a /deploy PR comment

achilleburah:
## Description

The `deploy-app-preview` label only fires on the label being added, or on a push while it is
already there. Redeploying an unchanged branch means removing the label and adding it back, which
is why the trigger feels unreliable. Commenting `/deploy` on the PR now does the same thing.

`issue_comment` runs on the default branch, so the comment itself cannot call `deploy-spa.yaml`
directly, it would check out and deploy `main`. The new `dispatch` job instead resolves the PR head
branch and re-runs this workflow on it through `workflow_dispatch`. That run has `github.ref_name`
set to the PR branch, so `deploy-spa.yaml` is unchanged and still gets `skip_slack: true` and
posts the preview URL. The command is restricted to OWNER, MEMBER and COLLABORATOR comments,
because the run carries our Cloudflare and Datadog secrets.

Two known limits, both acceptable for a preview:

- `/deploy` only works on PRs whose branch already contains this workflow, since the dispatched run
  uses the branch's own copy of the file. Older branches need a merge of `main` first.
- The alias comes from the branch name, so a PR deployed through both the label and a comment ends
  up with two preview URLs.

## Tests

`actionlint` passes. Traced the three event payloads through the job conditions: a labeled or
synchronized PR runs `deploy-app` only, a `/deploy` comment runs `dispatch` only, and the resulting
`workflow_dispatch` runs `deploy-app` and the comment job with the PR number from the input. The
command cannot be exercised before this lands, `workflow_dispatch` has to be on the default branch
to be dispatchable at all.

## Risk

Low. The label and push paths keep their exact conditions, and the comment path is additive. If the
dispatch misbehaves the label still works. Rollback is a revert of this file.

## Deploy Plan

Nothing to deploy, it takes effect on merge to `main`.

### DIFF
--- a/.github/workflows/deploy-spa-preview.yaml
+++ b/.github/workflows/deploy-spa-preview.yaml
@@ -1,18 +1,60 @@
-name: Deploy SPA preview on PR label
+name: Deploy SPA preview on PR
 
 on:
   pull_request:
     types: [labeled, synchronize]
+  issue_comment:
+    types: [created]
+  workflow_dispatch:
+    inputs:
+      pr_number:
+        description: "PR number to deploy and comment the preview URL on"
+        type: string
+        required: true
 
 permissions:
   contents: read
   pull-requests: write
 
 jobs:
+  dispatch:
+    if: >-
+      github.event_name == 'issue_comment'
+      && github.event.issue.pull_request
+      && startsWith(github.event.comment.body, '/deploy')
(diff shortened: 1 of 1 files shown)


achilleburah:
seems to work fine now... closing

PR #31720, opened 2026-09-02
Fail loudly on unattributed user usage in Metronome events

avervaet:
## Description

A `usage_type: "user"` event with no `userId` used to ship to Metronome as `user_id: "unknown"` instead of surfacing the caller's attribution bug. This now throws instead, so the bug fails loudly (a Temporal retry/alert) rather than shipping silently misattributed spend. Programmatic and free-origin usage are unaffected.

Base for #31721, which fixes one caller this check surfaces.

## Tests

Updated `lib/metronome/events.test.ts`: one test asserts the throw fires for a null-userId "user" event, another confirms programmatic usage still falls back to sane placeholder values.

## Risk

Low — no user-facing path touches this code. Worst case, an already-buggy event that used to ship as "unknown" now fails and retries in Temporal instead, surfacing the bug via alerting.

## Deploy Plan

- Front

### DIFF
--- a/front/lib/metronome/events.test.ts
+++ b/front/lib/metronome/events.test.ts
@@ -79,11 +79,26 @@ describe("Metronome aggregated usage event", () => {
     });
   });
 
-  it("always emits the required billable shape with valid values", () => {
-    // Even with no user / api key / agent, the required properties fall back to
-    // valid, non-empty values so the billable metric's `exists` filters match.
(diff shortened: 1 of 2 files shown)


avervaet:

- #31720 ◀
    - #31722

flvndvd (approved):

fabiencelier (approved):

PR #31721, opened 2026-09-02
Classify system API keys as programmatic Metronome usage

avervaet:
## Description

`isProgrammaticUsageFromContext` only recognized `authMethod === "api_key"`, missing `"system_api_key"`. 

That usage was falling into the generic "user" bucket with no user to attribute it to, which is exactly what #31720 now catches loudly. Fixing this

**Warning:** Workspace that used `system_api_key` auth this cycle will see its programmatic spend-cap counter jump upward on next resync, since usage previously misbucketed as "user" now counts as "programmatic".   

If a workspace has a configured `monthlyCapAwuCredits`, this could push it over the cap mid-cycle and start blocking programmatic calls that weren't being capped before. Not a bug, but a behavior change that lands the moment this deploys, with no gradual rollout.

## Tests

Covered by existing `lib/api/programmatic_usage/tracking.test.ts` suite (37 tests, all passing).

## Risk

Low — reclassifies a subset of previously-mislabeled "user" usage as "programmatic". No change to billing totals, only to which credit pool an event drains and how it's attributed.

## Deploy Plan

- Front

### DIFF
--- a/front/lib/api/programmatic_usage/common.ts
+++ b/front/lib/api/programmatic_usage/common.ts
@@ -55,17 +55,23 @@ const PROGRAMMATIC_USAGE_ORIGINS = Object.keys(
     USAGE_ORIGINS_CLASSIFICATION[origin as UserMessageOrigin] === "programmatic"
 );
 
+// Auth methods with no human behind them: a workspace API key, or the workspace's
+// own system key used internally (e.g. sub-agent runs). Both must classify as
+// programmatic, otherwise their usage falls back to the "user" bucket with no
+// real user to attribute it to.
+const PROGRAMMATIC_AUTH_METHODS = ["api_key", "system_api_key"];
+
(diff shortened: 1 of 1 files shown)


avervaet:

- #31720
    - #31721 ◀
        - #31722

flvndvd (commented):
Actually system key associated with a `USAGE_TYPE_USER` should be properly assigned to the underlying users. As mentioned on Slack this path is used through 2 (at least) known ways:
1. when you post a message in Slack and mention `@.Dust`
2. sub agent as we use our API to create the sub conversation

I'm not comfortable assigning this as programmatic, they are not, they are using the associated credit pool (at least, that the expectation I'd have)./

avervaet:
Closing — no longer needed, see https://dust4ai.slack.com/archives/C09GELMTTRT/p1788362364425619

PR #31722, opened 2026-09-02
Trace usage attribution back through parent conversations

avervaet:
## Description

Some messages still have no directly attributable user even after #31721 — e.g. a system-key hidden helper sub-agent calling pod_manager on behalf of a real user's conversation.

When a message's own user and the current auth are both empty, walk `agenticOriginMessageId` back to the parent conversation's message, follow `parentId` to the user message it replied to, and repeat (bounded to 5 hops) until a real human is found.

## Risk

Low-medium — new DB lookups on an already-async, fire-and-forget path; bounded hop count prevents runaway queries. No behavior change for the common case (direct user or same-auth fallback still resolve first).

## Deploy Plan

Front


avervaet:

- #31720
    - #31722 ◀

fabiencelier (approved):

PR #31723, opened 2026-09-02
feat: export messages with new es index

sfriquet:
## Description

Refs https://github.com/dust-tt/tasks/issues/10162

Add an alternative way to fetch messages from ES, using `agent_message_consumption` index. This is done by grouping documents by `message_id` on the fly.
Page size is decreased from 10K to 5K to avoid hitting the bucket size limits.
Add a feature flag to toggle between the old and new index.
Note: we still need to add `parent_message_id` in the new index to be iso.

## Tests

tested manually: downloaded a csv with and without the feature flag. 

## Risk

Low. No behavior change yet.

## Deploy Plan

Front only

### DIFF
--- a/front/lib/api/analytics/messages_export.ts
+++ b/front/lib/api/analytics/messages_export.ts
@@ -13,17 +13,15 @@ import {
   searchConsumptionAnalytics,
 } from "@app/lib/api/elasticsearch";
 import type { Authenticator } from "@app/lib/auth";
-import { MICRO_CREDITS_PER_CREDIT } from "@app/lib/credits/units";
+import { microCreditsToCredits } from "@app/lib/credits/units";
 import type {
   AgentMessageAnalyticsCost,
   AgentMessageAnalyticsModel,
 } from "@app/types/assistant/analytics";
-import type {
-  ModelIdType,
-  ModelProviderIdType,
-  ModelResolutionMethodType,
-  ReasoningEffort,
-} from "@app/types/assistant/models/types";
+import { isModelId } from "@app/types/assistant/models/models";
+import { isModelProviderId } from "@app/types/assistant/models/providers";
+import { isReasoningEffort } from "@app/types/assistant/models/reasoning";
+import { isModelResolutionMethod } from "@app/types/assistant/models/types";
 import type { Result } from "@app/types/shared/result";
 import { Err, Ok } from "@app/types/shared/result";
 import type { WorkspaceType } from "@app/types/user";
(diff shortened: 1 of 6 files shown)


front/lib/api/analytics/messages_export.ts
```diff
@@ -123,26 +133,252 @@ async function fetchAllMessageDocuments(
...
+        skills: { terms: { field: "tool.attributed_skill_ids", size: 100 } },
+        models: {
+          multi_terms: {
+            terms: [
+              { field: "model.model_id" },
+              { field: "model.provider_id" },
+              { field: "model.resolution_method" },
+              { field: "model.reasoning_effort" },
+            ],
+            size: 10,
```
flvndvd, line 247: Hard to keep track of all those sizes, should we split this query a bit with small comment above to explain what each piece does?

front/lib/api/analytics/messages_export.ts
```diff
@@ -123,26 +133,252 @@ async function fetchAllMessageDocuments(
...
+
+  const skillBuckets = bucketsToArray(bucket.skills?.buckets);
+
+  const modelBuckets = bucketsToArray(bucket.models?.buckets);
+  const primaryModel = modelBuckets[0];
+
+  const creditMicro = bucket.total_credit_micro.value ?? 0;
+
+  return {
+    workspace_id: "",
```
flvndvd: 🤔 Why not populating it?

front/lib/api/analytics/messages_export.ts
```diff
@@ -123,26 +133,252 @@ async function fetchAllMessageDocuments(
...
+      tool_name: String(b.key[1]),
+    })),
+    skills_used: skillBuckets.map((b) => ({
+      skill_name: b.key,
+    })),
+    cost: {
+      full_awu: 0,
+      llm_awu: 0,
+      tool_awu: 0,
+      billable_awu: Math.round(microCreditsToCredits(creditMicro)),
```
flvndvd, line 326: This won't match exactly to the actual billing, should we fetch the source of truth in DB?

front/lib/api/analytics/messages_export.ts
sfriquet: it's a bit misleading because we never need it, and I didn't bother passing a parameter just to ignore it later.
It's required because the function returns the existing `AgentMessageDocument` type which maps to the old index schema.
I'll just pass the workspaceId so it's less misleading for now.
Once we get rid of the old export I can get rid of this type entirely and map the result of the ES aggregation to the expected csv schema without this proxy.

front/lib/api/analytics/messages_export.ts
```diff
@@ -123,26 +133,252 @@ async function fetchAllMessageDocuments(
...
+    agent_id: firstBucketKey(bucket.agent_id),
+    agent_tag_ids: bucketsToArray(bucket.agent_tag_ids?.buckets).map(
+      (b) => b.key
+    ),
+    conversation_id: firstBucketKey(bucket.conversation_id),
+    ancestor_message_ids: [],
+    user_id: firstBucketKey(bucket.user_id),
+    context_origin: firstBucketKey(bucket.context_origin),
+    status: statusHit?.status ?? "",
+    tools_used: toolBuckets.map((b) => ({
```
flvndvd, line 315: Safe to remove. No associated header.

flvndvd (approved):
Let's add a log as well to track export type.

PR #31719, opened 2026-09-02
Add custom chrome extension in cors

tdraier:
## Description

fixes https://github.com/dust-tt/tasks/issues/10275

Add allowed origin for custom extension

## Deploy Plan

deploy front

PR #31724, opened 2026-09-02
[front-api] Serve /mcp as public credential-less CORS endpoint

tdraier:
## Description

The Dust MCP server endpoint (`POST /mcp`) authenticates **strictly via a Bearer JWT** in the `Authorization` header (`front/lib/api/mcp_server/auth.ts`) — it never reads cookies or any other ambient credential. The global CORS middleware (`front-api/middlewares/cors.ts`) nonetheless ran `/mcp` through the Origin allowlist and returned `403 { X-CORS-Reason: origin }` for any origin not on the list.

For a Bearer-only endpoint the Origin allowlist protects nothing (the JWT is the security boundary; a malicious page can't obtain the token, and non-browser clients ignore CORS anyway) while it blocks legitimate third-party MCP clients — browser extensions and other apps that register dynamically via DCR, which is exactly the audience MCP + OAuth DCR is built for. A customer hit this from a Chrome extension: OAuth/DCR/token exchange all succeeded, but the `initialize` POST was rejected with `403 x-cors-reason: origin`.

This change serves `/mcp` (and `/mcp/*`) as a public, credential-less CORS endpoint:

- `Access-Control-Allow-Origin: *`
- **no** `Access-Control-Allow-Credentials` (browsers forbid `*` + credentials anyway, and it's unnecessary since the token is an explicit header, not sent via credentials mode)
- preflight echoes the client's requested headers

This supersedes per-client origin allowlisting (e.g. #31719) — no extension ID needs whitelisting now or in the future. Scope is `/mcp` only; the `.well-known` OAuth discovery routes and every other endpoint keep the existing allowlist behavior untouched.

Note ([GEN1]): this introduces a second, path-scoped CORS pattern alongside the existing single global allowlist. Justified by MCP's open-client model, but flagging for reviewers.

## Tests

`front-api/middlewares/cors.test.ts` — added:
- non-allowlisted origin still `403`s with `X-CORS-Reason: origin` on regular endpoints (regression guard)
- `/mcp` allows any origin with `ACAO: *` and no `Allow-Credentials`
- `/mcp` preflight answers any origin and echoes requested headers

All 6 tests pass; `tsgo --noEmit` clean.

## Risk

Low. The change only widens CORS for `/mcp`, which is Bearer-JWT-only — no cookie/session auth path exists there, so there is no CSRF/credential-theft surface to open. No behavior change for any other route (allowlist + `x-cors-reason` unchanged). Fully rollback-safe (revert the commit).

## Deploy Plan

Standard front-api deploy. No migration, no config/env change, no ordering constraints. Effective as soon as front-api ships.

### DIFF
--- a/front-api/middlewares/cors.test.ts
+++ b/front-api/middlewares/cors.test.ts
@@ -13,8 +13,11 @@ function createApp() {
   const app = new Hono();
   app.use("*", cors);
   app.get("/", (ctx) => ctx.text("ok"));
+  app.all("/mcp", (ctx) => ctx.text("ok"));
   return app;
 }
+
+const EXTENSION_ORIGIN = "chrome-extension://adoiifkpgaibbkgbicbdhpeoffmblbeb";
 
 function getExposedHeaders(response: Response): string[] {
   return (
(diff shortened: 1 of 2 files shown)


tdraier:

- #31724 ◀

Fraggle (approved):
LGTM

matteotrab (approved):

PR #31726, opened 2026-09-02
feat: prefer auth-scoped group fetches and make group acl bypasses explicit

matteotrab:
## Description

This PR makes sure that the internal fetch methods of the group resource are used only when there is no authenticator. Moreover it removes the canRead check from the `fetchByName` method and renames it to `dangerouslyFetchByName`. This is needed because its callers also need to fetch regular_auto groups.

## Tests
Manually.

## Risks
Low. Same behavior is maintained.

## Deploy Plan
Standard deployment.

### DIFF
--- a/front/lib/resources/group_resource.ts
+++ b/front/lib/resources/group_resource.ts
@@ -973,6 +973,11 @@ export class GroupResource extends BaseResource<GroupModel> {
     return !!group;
   }
 
+  // Fetches a group by name across every kind, without an ACL check: the result
+  // may be a group the caller cannot read, so never use it to grant access.
+  // Only for name-conflict handling, where (workspaceId, name) being unique
+  // forces us to see the colliding row whatever its kind.
+  // Prefer groupExistsByName when a boolean is enough.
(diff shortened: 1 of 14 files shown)


matteotrab:

- #31461
    - #31464
        - #31563
            - #31565
                - #31612
                    - #31685
                        - #31686
                            - #31704
                                - #31726 ◀
                                    - #31754
                                        - #31762
                                            - #31771
                                                - #31774
                                                    - #31808

front/lib/resources/group_resource.ts
```diff
@@ -963,7 +963,7 @@ export class GroupResource extends BaseResource<GroupModel> {
     return !!group;
   }
 
-  static async fetchByName(
+  static async dangerouslyFetchByName(
```
fabiencelier, line 981: let explain in the doc of the function what is dangerous and must be done/checked outside of it

fabiencelier (approved):

philipperolet (approved):

PR #31727, opened 2026-09-02
Webhook router now uses cells information instead of regions.

Fraggle:
## Description

I migrate webhook-router routing metadata from regions to cells across Slack, Notion, and Teams. `WebhookRouterConfigService` now stores and forwards connector IDs by cell, webhook verification selects configured cells directly, and `front/lib/api/cells/config.ts` exposes connector URLs. A connectors migration removes legacy `regions` fields after the existing cells backfill.

## Tests

No new tests added. CI runs lint, TypeScript checks, and repository test shards; React Doctor correctly skipped because no React files changed.

## Risk

Medium risk: webhook delivery depends on the cell mappings, and the config cleanup is not compatible with an older regions-based router. Run the cells backfill before removing `regions`, verify webhook delivery in both cells, and retain a config backup for rollback.

## Deploy Plan

Deploy `connectors`, the Firebase `webhook-router`, and `front`. Run `connectors/migrations/20260902_remove_webhook_router_config_regions.ts` in dry-run mode, then execute it after confirming the cells backfill is complete. No SQL migration or infrastructure change.

### DIFF
--- a/connectors/migrations/20260902_remove_webhook_router_config_regions.ts
+++ b/connectors/migrations/20260902_remove_webhook_router_config_regions.ts
@@ -1,0 +1,25 @@
+import { WebhookRouterConfigService } from "@connectors/lib/webhook_router_config";
+import type { Logger } from "@connectors/logger/logger";
+import { makeScript } from "scripts/helpers";
+
+makeScript({}, async ({ execute }, logger: Logger) => {
+  const service = new WebhookRouterConfigService();
+  const { entriesUpdated } = await service.removeRegions(execute);
+
+  if (entriesUpdated === 0) {
+    logger.info("No webhook router entries have a regions field to remove.");
+    return;
+  }
+
+  if (execute) {
+    logger.info(
+      { entriesUpdated },
+      "Removed regions from webhook router config entries."
+    );
+  } else {
+    logger.info(
+      { entriesUpdated },
+      "[DRY RUN] Would remove regions from webhook router config entries."
+    );
+  }
+});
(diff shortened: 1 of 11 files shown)


Fraggle:

- #31727 ◀

connectors/migrations/20260902_remove_webhook_router_config_regions.ts
```diff
@@ -0,0 +1,25 @@
+import { WebhookRouterConfigService } from "@connectors/lib/webhook_router_config";
+import type { Logger } from "@connectors/logger/logger";
+import { makeScript } from "scripts/helpers";
+
+makeScript({}, async ({ execute }, logger: Logger) => {
+  const service = new WebhookRouterConfigService();
+  const { entriesUpdated } = await service.removeRegions(execute);
```
flvndvd, line 7: Let's maybe keep region around for a few days, first? In case we need to rollback.

connectors/src/lib/webhook_router_config.ts
```diff
@@ -12,10 +10,10 @@ import { z } from "zod";
 
 const WEBHOOK_ROUTER_CONFIG_FILE = "webhook-router-config.json";
 
+// Unknown keys (e.g. legacy `regions`) are stripped by Zod's default object behavior.
 const WebhookRouterEntrySchema = z.object({
```
flvndvd: As this one is used on the update path, it will drop regions from the GCS file anyway.

flvndvd (approved):

connectors/migrations/20260902_remove_webhook_router_config_regions.ts
Fraggle: Yep. FYI, i also saved a copy of the router file prior my changes.

connectors/src/lib/webhook_router_config.ts
Fraggle: Good catch, I dont want that.

PR #31728, opened 2026-09-02
[front] usage members table: drop secondary columns before Name on narrow screens

PopDaph:
The members table on the Usage page is fixed layout and Name is the only column without a set width, so on narrow windows Name shrank to nothing while every other column kept its width.

Secondary columns now drop first as the window narrows, keeping Name readable:
- Groups below 2xl (1536px),
- Models tier and Seat below xl (1280px),
- Credits usage below lg (1024px).

Seat goes from `w-36` to `w-32` and Credits usage from `w-64` to `w-56` so Name gets ~200px at each step with the sidebar open.

Test on https://pr-31728-merge--app.preview.dust.tt/w/0ec9852c2f/usage versus https://app.dust.tt/w/0ec9852c2f/usage

### DIFF
--- a/front/components/workspace/MembersUsageTable.tsx
+++ b/front/components/workspace/MembersUsageTable.tsx
@@ -526,7 +526,7 @@ const seatTypeColumn: ColumnDef<RowData, string> = {
     );
   },
   meta: {
-    className: "w-36",
+    className: "w-32",
   },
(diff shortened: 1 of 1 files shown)


aubin-tchoi (approved):
LGTM

PR #31730, opened 2026-09-02
[Admin Governance] Mirror agent editor additions to grants

philipperolet:
## Description

Follows https://github.com/dust-tt/dust/pull/31649, which introduced the agent permission resource.

Agent editors are currently stored in legacy `agent_editors` groups. As we migrate to `group_permissions`, editor additions made after the backfill must be written to both systems so the data remains comparable.

This PR mirrors additions in the three workflows that can add an editor:

- creating a pending agent grants its creator editor access;
- publishing or updating an agent grants every submitted editor;
- the dedicated editor endpoint grants newly added editors.

The new grants use the stable numeric agent ID and add multiple editors in one batch. Legacy memberships and grants share a database transaction, so either both writes succeed or neither does. Permission reads remain on the legacy system; editor removals and agent-deletion cleanup are left to the next PR.

## Risks

Blast radius: Creating or publishing custom agents and adding agent editors.
Risk: standard

## Deploy Plan

- deploy front

## Tests

- [x] `npm test --workspace front -- lib/resources/agent_resource.test.ts lib/resources/group_permission_resource.test.ts lib/api/assistant/configuration/agent.test.ts` (80 tests)

### DIFF
--- a/front/lib/api/assistant/configuration/agent.ts
+++ b/front/lib/api/assistant/configuration/agent.ts
@@ -140,7 +140,7 @@ export async function createPendingAgentConfiguration(
       transaction: t,
       authorId: user.id,
     });
-    await AgentResource.fromAgentConfiguration(agent).grantEditors(auth, {
+    await AgentResource.fromAgentConfigurationModel(agent).grantEditors(auth, {
       editors: [user.toJSON()],
       transaction: t,
     });
@@ -943,7 +943,7 @@ export async function createAgentConfiguration(
           }
         }
 
-        await AgentResource.fromAgentConfiguration(
+        await AgentResource.fromAgentConfigurationModel(
           agentConfigurationInstance
         ).grantEditors(auth, { editors, transaction: t });
       }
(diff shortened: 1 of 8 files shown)


front/lib/api/assistant/configuration/agent.ts
```diff
@@ -83,6 +85,30 @@ const PENDING_AGENT_PLACEHOLDER_DESCRIPTION = "";
 const PENDING_AGENT_PLACEHOLDER_PICTURE_URL =
   "https://dust.tt/static/systemavatar/dust_avatar_full.png";
 
+async function grantAgentEditors(
```
philipperolet: could we add that as a method of the resource?
Then when needed to use, construct the agent resource and call the method?
this is because we plan to transition slowly towards the resource

philipperolet: Done in 45b3242c04. The grant operation now lives on AgentResource, and each write path constructs the logical-agent resource before calling grantEditors().

matteotrab (approved):

PR #31731, opened 2026-09-02
[sparkle] DataTable: allow every breakpoint in columnsBreakpoints

PopDaph:
`columnsBreakpoints` only accepted `xs` to `xl` while `breakpoints` also defines `2xl` and `xxs`. The type now derives from the `breakpoints` map, so `2xl` can be used and the two stay in sync.

Need this to update a table on the Usage page 🙏🏻


tdraier (approved):

ykmsd:
👍

PR #31732, opened 2026-09-02
fix(tools): load 1Password-managed .env correctly in start-mprocs.sh

avervaet:
## Description

Fixing the broken 1-password manage env and mprocs caused by GCP_SERVICE_ACCOUNT format.

## Tests

tested localy

## Risk

None — dev tooling only, not shipped code. Affects local Mac dev environment setup exclusively.

## Deploy Plan

- None (local dev script, no deploy)

### DIFF
--- a/tools/start-mprocs.sh
+++ b/tools/start-mprocs.sh
@@ -3,6 +3,34 @@
 # Where the script is defined, absolute path
 SCRIPT_DIR=$(cd "$(dirname "$0")" && pwd)
 
+# Load the repo-root .env into the environment without `source`-ing it as shell
+# code: the file is a 1Password-managed KEY=VALUE dump, not valid shell (it can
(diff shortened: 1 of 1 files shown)


tools/start-mprocs.sh
```diff
@@ -3,6 +3,34 @@
 # Where the script is defined, absolute path
 SCRIPT_DIR=$(cd "$(dirname "$0")" && pwd)
 
+# Load the repo-root .env into the environment without `source`-ing it as shell
```
Fraggle, line 6: That is specific to your setup (mine no longer use the .env file, I was sick of validating all the time).
Sending you an alternative.

avervaet:
Closing this in favor of a cleaner approach — see https://dust4ai.slack.com/archives/C050SM8NSPK/p1788251282643479. We'll load 1Password secrets via the `op` CLI directly into the shell instead of parsing the repo's `.env` file.

PR #31734, opened 2026-09-02
Track blocked trigger executions in PostHog

theogz:
## Description

Instrument blocked trigger executions with a `trigger_blocked` PostHog event. The event is emitted when trigger-created message posting is rejected by a usage or access limit, after the orphan conversation has been created but before any user or agent message is persisted.

## Risk

Low. The PostHog call is best-effort and is wrapped in its own error handling. It does not change trigger execution, limit enforcement, or retry behavior. The event is only emitted for limit-related and no-seat errors.

## Deploy Plan

Normal. Verify that the events are sent post-deploy

### DIFF
--- a/front/lib/tracking/posthog/server.ts
+++ b/front/lib/tracking/posthog/server.ts
@@ -23,30 +23,14 @@ function getClient(): PostHog | null {
 }
 
 export class PostHogServerSideTracking {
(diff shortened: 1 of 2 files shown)


aubin-tchoi (approved):
LGTM

PR #31741, opened 2026-09-02
[dust-hive] add opt-in storybook service forwarded on 6006

PopDaph:
Adds a `storybook` service to dust-hive so Sparkle's Storybook can run per hive instead of by hand on a fixed 6006.

- New port offset 8 (`base + 8`), forwarded from 6006 so `.mcp.json` and the Storybook MCP keep working and follow the forwarded env.
- Not started by `warm` (same as `viz`). Start it with `dust-hive restart <env> storybook`, stop with `dust-hive stop <env> storybook`.
- `dust-hive forward` and `status` now list it.
- README forward table updated (also adds the viz and front-spa ports that were missing).
- Root `AGENTS.md` mentions the dust-hive way of starting Storybook, only relevant when using dust-hive locally.

Note: 6006 is now a forwarder port, so `dust-hive warm` will refuse to start the forwarder if a manual `npm run storybook` already holds it. Stop it or use `--no-forward`.


zmarouf (approved):
LGTM

PR #31749, opened 2026-09-03
Front builds once

flvndvd:
## Description

Front, front-edge and front-sse move to the build-once job: one Depot bake per deploy, pushed to the global `dust-infra/dust-images` registry and, for now, to both regional registries so the current values files keep resolving. Marketing is the only component left on the regional matrix (will see with @id13 for this one exception).

This is possible because nothing region-specific is baked into the front images anymore. front-api and the workers are esbuild bundles, so every `NEXT_PUBLIC_*` reference is a runtime `process.env` lookup, and the pods already receive their region config through the front secret bundle (DUST_REGION, DUST_US_URL, VIZ_PUBLIC_URL and friends). The Dockerfile stops declaring the six values that differed between regions (the API and static website URLs, the viz URL and the three Novu endpoints). The three still read server-side, `NEXT_PUBLIC_DUST_API_URL`, `NEXT_PUBLIC_DUST_STATIC_WEBSITE_URL` and `NEXT_PUBLIC_NOVU_API_URL`, now come from Secret Manager like the rest of the region config. The Novu identifier and websocket URL are only read by the SPA, which has its own build.

## Tests

Not tested end to end: the first `front` deploy after merge is the real test, see the deploy plan.

## Deploy Plan

1. Create the region secrets in both projects:
- `front-NEXT_PUBLIC_DUST_API_URL`
- `front-NEXT_PUBLIC_DUST_STATIC_WEBSITE_URL`
- `front-NEXT_PUBLIC_NOVU_API_URL`
- and the two `front-edge-` URL overrides
2. Merge, then deploy `front-edge` first as the canary of the new path, then `front` and `front-sse`.
3. Follow-up in dust-infra: point the six front image repositories at the global registry.


flvndvd:
Secrets added ✅

id13:
> Secrets added ✅

can you add it to the new cell ? :tear

id13 (approved):

flvndvd:
> > Secrets added ✅
> 
> can you add it to the new cell ? :tear

Yes, sir 👍

PR #31752, opened 2026-09-03
[front] Fix accent-insensitive consumption search

flvndvd:
## Description

Make analytics consumption attribution search ignore diacritics in both the query and catalog labels. This lets searches such as `developpeur` match `Développeur`, and vice versa.

## Tests

- Added regression coverage for accented labels and accented queries.
- `npm run test -- lib/api/analytics/consumption/top.test.ts` (26 tests pass).
- Biome check passes on both changed files.
- The full front typecheck currently fails on unrelated existing Sparkle prop/type mismatches on `main` (`grow`, `2xl`, and `xs`).

## Risk

Low. The change only normalizes strings before the existing in-memory catalog substring match. The Elasticsearch query and stored values remain unchanged.

## Deploy Plan

Deploy with the regular front release. No migration or backfill is required.

### DIFF
--- a/front/lib/api/analytics/consumption/top.test.ts
+++ b/front/lib/api/analytics/consumption/top.test.ts
@@ -260,12 +260,26 @@ describe("consumption top rankings", () => {
   });
 
   it.each([
-    ["AGENT 080", { terms: { "agent.attributed_id": ["agent80"] } }],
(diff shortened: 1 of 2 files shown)


adrsimon (approved):
LGTM

PR #31753, opened 2026-09-03
feat(frames): expose Delete for Frame v2 packages in file explorers

davidebbo:
## Description

Frame v2 packages in the file explorer only offered "View source" in their "..." menu, so there was no user-facing way to delete a Frame v2. This adds a Delete item for Frame packages in both the Pod file explorer and the conversation file explorer.

Deleting a package deletes its `manifest.json` path through the existing `DELETE /api/w/{wId}/files/path/{canonicalPath}` route. The server resolves that path to the Frame file and runs the package-aware `deleteFrameV2`, which removes the source folder, sandbox, functions and invocations, publications, database replicas and sharing records. No server changes.

- `FileExplorer` adds a Delete item on `frame_package` entries when `onDelete` is set, and gains an optional `canDelete` filter so a caller can restrict which entries are deletable.
- The Pod explorer's delete handler gets a `frame_package` branch with a Frame-specific confirmation.
- The conversation explorer gets a delete handler restricted to Frame packages, gated on the `frames_v2` feature flag, refreshing both conversation and Pod file lists afterwards.
- `useDeletePodFile` moves from `lib/swr/pods.ts` to `lib/swr/files.ts` as `useDeleteFileByPath`, since it never depended on the Pod id.

Not included: deleting a plain folder that contains a Frame still bypasses the Frame cleanup and orphans its identity and sandbox. That is pre-existing and worth a separate change.

## Tests

- `FileExplorer.test.tsx` updated to assert Delete is offered on a Frame package and called with the manifest path.
- Typecheck and biome pass.

## Risk

Low. Client-only change reusing an existing deletion endpoint. Frames in a database-backed file system still fail server-side and surface an error notification.

## Deploy Plan

Deploy normally.

### DIFF
--- a/front/components/assistant/conversation/files_panel/ConversationFileExplorer.tsx
+++ b/front/components/assistant/conversation/files_panel/ConversationFileExplorer.tsx
@@ -105,8 +105,8 @@ export function ConversationFileExplorer({
     [openPanel]
   );
 
-  // Only Frame packages are deletable from the conversation explorer; the Pod explorer handles
-  // the rest of a Pod's files.
+  // Only Frame packages get a Delete item here (see `canDelete`); other conversation files stay
+  // non-deletable as before.
   const onDelete = useCallback(
     async (entry: FileExplorerEntry) => {
       if (entry.kind !== "frame_package") {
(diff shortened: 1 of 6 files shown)


front/components/assistant/conversation/files_panel/ConversationFileExplorer.tsx
```diff
@@ -87,6 +105,34 @@ export function ConversationFileExplorer({
     [openPanel]
   );
 
+  // Only Frame packages get a Delete item here (see `canDelete`); other conversation files stay
+  // non-deletable as before.
+  const onDelete = useCallback(
```
davidebbo, line 110: Not sure why regular convo file didn't have a delete, so didn't change that.

fontanierh (approved):

PR #31754, opened 2026-09-03
feat: filter user group listing on canRead and make bypass explicit

matteotrab:
## Description

This PR adds canRead checks when listing user groups and makes internal bypasses explicit.

## Tests

Manually.

## Risks
We are adding a canRead check that was not present before. There should be no change as currently every role can read every group kind different from system and agent_editors, which are not group kinds fetched by the updated method.

## Deploy Plan

Standard deployment.

### DIFF
--- a/front/lib/resources/group_resource.ts
+++ b/front/lib/resources/group_resource.ts
@@ -1125,14 +1125,14 @@ export class GroupResource extends BaseResource<GroupModel> {
   private static async listUserGroupModelIdsInWorkspace({
     user,
     workspace,
-    groupKinds = GROUP_KINDS.filter((k) => k !== "system"),
+    groupKinds,
     transaction,
     dangerouslySkipMembershipCheck = false,
     at = new Date(),
   }: {
     user: UserResource;
     workspace: LightWorkspaceType;
-    groupKinds?: Exclude<GroupKind, "system">[];
+    groupKinds: Exclude<GroupKind, "system">[];
     transaction?: Transaction;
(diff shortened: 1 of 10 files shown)


matteotrab:

- #31461
    - #31464
        - #31563
            - #31565
                - #31612
                    - #31685
                        - #31686
                            - #31704
                                - #31726
                                    - #31754 ◀
                                        - #31762
                                            - #31771
                                                - #31774
                                                    - #31808

front/lib/resources/group_resource.ts
```diff
@@ -1188,38 +1188,51 @@ export class GroupResource extends BaseResource<GroupModel> {
...
-    groupKinds = GROUP_KINDS.filter((k) => k !== "system"),
+    groupKinds,
     transaction,
     at,
   }: {
+    auth: Authenticator;
     user: UserResource;
-    workspace: LightWorkspaceType;
-    groupKinds?: Exclude<GroupKind, "system">[];
+    groupKinds: Exclude<GroupKind, "system">[];
```
fabiencelier, line 1232: can this be UserVisibleGroupKind ?

fabiencelier (approved):

front/lib/resources/group_resource.ts
matteotrab: No because we need to fetch all the groups when called by `revokeWorkOSUserMembership`. We have the user visible version `listUserGroupsInWorkspace` for all other callers

PR #31755, opened 2026-09-03
feat(pods): offer Add as Pod tab on Frame packages instead of manifest.json

davidebbo:
## Description

In a Pod's Files tab, a Frame v2 package had no "Add as Pod tab" item. The item only appeared after opening "View source" and using it on `manifest.json`. This moves it to the Frame package entry and removes it from the manifest file.

The package entry already carries the manifest path, so the created tab points at the same path as before. The default tab title now comes from the Frame folder name rather than `manifest`.

The "Pin as Pod banner" item is unaffected: Frame v2 manifests were never an interactive content type, so it did not show on them before and still does not.

## Tests

- Typecheck and biome pass.
- Manual: open a Pod's Files tab with a Frame v2, check the "..." menu on the Frame card shows "Add as Pod tab" and the manifest inside "View source" no longer does.

## Risk

Low. Client-only change to which explorer entries expose an existing action.

## Deploy Plan

Deploy normally.

### DIFF
--- a/front/components/pod/files/PodFileExplorer.tsx
+++ b/front/components/pod/files/PodFileExplorer.tsx
@@ -47,6 +47,7 @@ import type {
   DataSourceViewType,
 } from "@app/types/data_source_view";
 import {
+  frameV2ContentType,
   getSupportedFileExtensions,
   isInteractiveContentType,
 } from "@app/types/files";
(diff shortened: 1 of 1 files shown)


fontanierh (commented):
LGTM

PR #31757, opened 2026-09-03
feat(ConversationPicker): feat(ConversationPicker): new conversation picker component

achilleburah:
## Description

ConversationPicker component to allow selecting a previous conversation. 
Conversations items should be passed to the component and therefore, the filtering responsibility falls to the caller / parent component. 

Added in the side conversation panel of the Analytics view. 

#### Component
image

#### Analytics View
image

## Tests

Tested locally. 

## Risk

Low. 

## Deploy Plan

Classic deploy.


achilleburah:
Changed direction.

PR #31759, opened 2026-09-03
[front] enh: add active user count in timeseries endpointt

aubin-tchoi:
## Description

This PR exposes the per-bucket active user count in the consumption timeseries endpoint.

One backend change adds a cardinality aggregation per date bucket and returns the count alongside credits. Active users remain a per-bucket count in cumulative mode, and future buckets are zeroed (won't be shown anyway).

The chart overlay is split into the stacked frontend follow-up #31809.

## Tests

- Added tests.
- Tested locally (with the next PR containing the UI).

## Risk

- Low.

## Deploy Plan

- Deploy front.


avervaet (approved):
LGTM

PR #31763, opened 2026-09-03
fix(pods): render Frame v2 tabs as Frames instead of manifest previews

davidebbo:
## Description

Adding a Frame v2 as a Pod tab showed the raw `manifest.json` instead of launching the Frame. `PodFileTabContent` decided whether a tab is a Frame with `isInteractiveContentType`, which only covers the legacy Frame content types, so Frame v2 fell through to the generic file preview.

Switch to `isFrameContentType`, which also matches the Frame v2 manifest content type. The Frame visualization branch already resolves the renderable content through `usePodFrameRenderableContent` and passes the v2 `frameId` through, the same way `PodPinnedBanner` does, so no other change is needed.

## Tests

- Typecheck and biome pass.
- Manual: add a Frame v2 as a Pod tab, open the tab, the Frame renders instead of the manifest.

## Risk

Low. One-line predicate change in the tab content component.

## Deploy Plan

Deploy normally.


fontanierh (commented):
LGMT

PR #31765, opened 2026-09-03
feat(frames): download a Frame v2 package as a zip from the file explorer

davidebbo:
## Description

Regular files have a Download item in the file explorer's "..." menu, but Frame v2 packages did not. This adds Download to Frame package entries. It downloads the whole source folder as a zip named after the folder, in both the Pod file explorer and the conversation file explorer.

Server: `GET /api/w/{wId}/files/path/{canonicalPath}?archive=1` zips every file under the folder at that path, with entry paths relative to the folder. The listing and reads go through `DustFileSystem`, so mount read access applies as for any other read. Empty or missing folders return 404, and folders over 100 MB uncompressed return 413 before any bytes are read. The route is `@ignoreswagger`, as before.

Client: `useFileDownload` accepts Frame package entries and fetches the archive of the manifest's folder, saving it as `<folder>.zip`. `FileExplorerFramePackageCard` gets the same `onDownload` wiring as file cards.

## Tests

- Route tests: archive zips the listed files with relative paths and the right headers; a missing folder returns 404. The test file now pins `@vitest-environment node`, as the Pod app export test does, since adm-zip needs Node builtins.
- Explorer test asserts Download appears on a Frame package.
- Typecheck for front and front-api, biome pass.

## Risk

Low. New opt-in query parameter on an existing read route; downloads are bounded by a size limit.

## Deploy Plan

Deploy normally.

### DIFF
--- a/front-api/routes/w/[wId]/files/path/[...canonicalPath].test.ts
+++ b/front-api/routes/w/[wId]/files/path/[...canonicalPath].test.ts
@@ -1,3 +1,6 @@
+// @vitest-environment node: adm-zip requires Node builtins (Buffer, zlib).
+// This directive makes them available in the test environment.
+
 import { createConversation } from "@app/lib/api/assistant/conversation";
 import { getPrivateUploadBucket } from "@app/lib/file_storage";
 import { FileResource } from "@app/lib/resources/file_resource";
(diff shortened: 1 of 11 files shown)


davidebbo:
Redid this differently

PR #31769, opened 2026-09-03
feat(frames): import a Frame v2 zip into a Pod folder from the Files UI

davidebbo:
## Description

Stacked on #31765 (Download on Frame packages); retarget to main once that merges.

This adds the inverse of the download: a Frame zip can be imported into a Pod, including a different Pod from the one it was downloaded from, from the Files tab's Add menu.

Server: `POST /api/w/{wId}/files/path/{canonicalFolder}?archive=1` takes the zip as a multipart `file` field and creates the folder from it. It refuses when the folder already has files (409), rejects zip-slip entries, empty archives, and archives over the size and entry-count limits (400 / 413), and writes each entry through `DustFileSystem` so mount write access applies. When the archive root holds a `manifest.json`, the Frame is registered through `registerFrameV2FromSourceUsingFileSystem`, so it appears as a Frame package right away. Entries are stored with a content type derived from their extension; `.json` is pinned to plain JSON because an internal Dust format claims that extension first in the format table.

Client: "Import Frame" in the Pod Add menu, gated on the `frames_v2` flag. The dialog takes the zip and a folder name defaulting to the zip's basename, imports into the current folder, and refreshes the explorer.

Publishing: the imported Frame is published as part of the import, so it renders right away. The UI bundle is built in-process with esbuild and needs no sandbox, so the build path now accepts a null conversation and only requires one when the manifest declares functions, which `dsbx function build` compiles inside a sandbox (it also runs the module's top-level code to extract schemas). For a Pod import whose manifest declares functions, a short-lived conversation is created in the Pod with `test` visibility, which Pod conversation lists and new-conversation notifications already exclude, its sandbox is used for the build, and it is torn down afterwards: sandbox destroyed, conversation soft-deleted. A failed publication does not fail the import, since the files and the Frame identity are in place; it is reported on the result and surfaced in the notification.

## Tests

- Route tests: import extracts files with the expected paths and content types, registers a Frame v2 at the manifest path and publishes it (active publication set on the Frame); a plain folder imports without registration; an existing folder returns 409 with nothing written; an unsafe `..` entry returns 400 with nothing written; POST without `?archive=1` returns 400.
- Typecheck for front and front-api, biome pass.

## Risk

Low to moderate. New write endpoint, but it reuses the existing file system write path and Frame registration, and validates the archive before writing anything.

## Deploy Plan

Deploy normally, after #31765.

### DIFF
--- a/front-api/routes/w/[wId]/files/path/[...canonicalPath].test.ts
+++ b/front-api/routes/w/[wId]/files/path/[...canonicalPath].test.ts
@@ -242,6 +242,146 @@ describe("GET /api/w/:wId/files/path/:canonicalPath?archive=1", () => {
   });
 });
 
+describe("POST /api/w/:wId/files/path/:canonicalPath?archive=1", () => {
+  beforeEach(() => {
+    vi.clearAllMocks();
+  });
+
+  function makeArchive(files: Record<string, string>): Buffer {
+    const zip = new AdmZip();
+    for (const [relPath, content] of Object.entries(files)) {
+      zip.addFile(relPath, Buffer.from(content, "utf-8"));
+    }
+    return zip.toBuffer();
+  }
+
+  function importRequest(
+    workspace: { sId: string },
+    canonicalPath: string,
+    archive: Buffer
+  ) {
+    const body = new FormData();
+    body.append(
+      "file",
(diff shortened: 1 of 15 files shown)


davidebbo:
Will do differently

PR #31762, opened 2026-09-03
feat: filter member group name fetch on canRead

matteotrab:
## Description

This PR filters member group names by the requesting user’s `canRead` permissions and requires callers to explicitly provide the group kinds to fetch.

## Tests

Manually.

## Risks
Low. Only user visible groups are fetched and everyone has permission to read them.

## Deploy Plan

Standard deployment.

PR #31771, opened 2026-09-03
feat: explicit group acl bypasses and resource-owned membership changes

matteotrab:
## Description

This PR centralizes project leave behavior in `SpaceResource.leave`, makes group membership operations explicitly privileged through `dangerously*` methods, and filters provisioned groups by `canRead` before returning them.

## Tests

Manually.

## Risks
Low. Same behavior is mantained.

## Deploy Plan

Standard deployment.

### DIFF
--- a/front/lib/resources/group_resource.ts
+++ b/front/lib/resources/group_resource.ts
@@ -1314,7 +1314,7 @@ export class GroupResource extends BaseResource<GroupModel> {
    * query to the caller's group ids keeps this to a single row-bounded query
    * regardless of how many groups the workspace has.
    */
-  static async listGroupModelIdsByUserModelIdInWorkspace({
+  static async dangerouslyListGroupModelIdsByUserModelIdInWorkspace({
     workspace,
     userModelIds,
     groupModelIds,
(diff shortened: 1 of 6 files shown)


matteotrab:

- #31461
    - #31464
        - #31563
            - #31565
                - #31612
                    - #31685
                        - #31686
                            - #31704
                                - #31726
                                    - #31754
                                        - #31762
                                            - #31771 ◀
                                                - #31774
                                                    - #31808

front/lib/resources/space_resource.ts
```diff
@@ -1691,6 +1691,72 @@ export class SpaceResource extends BaseResource<SpaceModel> {
...
   }
 
+  /**
+   * Removes the authenticated user from this space's auto-created groups.
+   *
+   * Unlike `removeMembers`, no admin permission is required: a member may
+   * always leave on their own. Projects only, as membership in the other space
+   * kinds is not opt-out.
+   */
+  async leave(
```
fabiencelier: can we rename it leavePod or leaveProject for clarity ?

front/lib/resources/group_resource.ts
```diff
@@ -1299,7 +1299,7 @@ export class GroupResource extends BaseResource<GroupModel> {
    * query to the caller's group ids keeps this to a single row-bounded query
    * regardless of how many groups the workspace has.
    */
-  static async listGroupModelIdsByUserModelIdInWorkspace({
+  static async dangerouslyListGroupModelIdsByUserModelIdInWorkspace({
```
fabiencelier, line 1321: please say in the doc why it is dangerous for all these dangerouslyXXX methods

fabiencelier (approved):

philipperolet (approved):
LGTM modulo indeed explaining in all dangerous method (or maybe in one, and others reference it) why they are dangerous: they don't check permissions, and are for reg auto groups that should go through the resource etc.

front/lib/resources/space_resource.ts
```diff
@@ -1691,6 +1691,72 @@ export class SpaceResource extends BaseResource<SpaceModel> {
...
+      return new Err(
+        new DustError("user_not_member", "You are not a member of this Pod.")
+      );
+    }
+
+    const user = auth.getNonNullableUser();
+
+    const editorGroup = await this.fetchManualEditorGroup(auth);
+    if (editorGroup) {
+      const activeEditors = await editorGroup.getActiveMembers(auth);
```
philipperolet, line 1751: ok so leaving a group-managed Pod no longer clears suspended manual membership? likely fine but worth checking e.g. when reg auto members are suspended when moving to a group management, there's an edge case there, check whether we can ignore or not

matteotrab: Currently pods cannot have a management mode group so this is not an issue. 
When we will introduce a group management mode for pods this should work because you will not be able to autonomously leave a pod in management mode groups, as it would mean leaving the group not the pod.

PR #31773, opened 2026-09-03
Fix duplicate Frame package switch case

fontanierh:
## Description

Remove the duplicate, unreachable `frame_package` switch case that triggers a Vite duplicate-case warning.

## Tests

Ran the focused FileExplorer component test locally (4/4).

## Risk

Low. The remaining `frame_package` branch already renders the package card.

## Deploy Plan

Standard deploy.


smb2268 (approved):
LGTM!

PR #31778, opened 2026-09-03
[front] Frames v2: rename Frames from the file explorer

davidebbo:
## Description

Adds a **Rename** action on Frame v2 package entries in the file explorer (Pod files and conversation files). Until now a Frame folder could only be renamed as a plain folder, which moved the bytes in GCS but left the manifest \`FileResource.mountFilePath\`, the Pod file tabs and the pinned Frame pointing at the old path.

- \`renameCanonicalFile\` detects a registered Frame at \`<path>/manifest.json\` and delegates to the Frame source move (\`moveFrameV2Source\`, merged in #31532 / #31552 without a production caller). No new API: the explorer keeps calling \`PATCH /files/path/... {action:"rename"}\`, and the route maps Frame move errors to 400/409/500.
- \`moveFrameV2SourceUsingFileSystem\` is split out so callers that already hold an authorized \`DustFileSystem\` (the Pod explorer has no conversation) can use it; \`moveFrameV2Source\` keeps its conversation-based signature.
- The move now repoints Pod metadata (\`pinnedFramePath\`, \`frameTabs\`, \`tabsOrder\`) via a new \`ProjectMetadataResource.renameFramePath\`, so Pod navigation keeps working after a rename or move.
- UI: \`RenameFileDialog\` moves next to the explorer with a \`frame\` kind and a shared \`useRenameFileByPath\` hook (replacing the Pod-only \`useRenamePodFile\`). \`FileExplorer\` gets a \`canRename\` filter mirroring \`canDelete\`; the conversation panel only offers Rename on Frame packages, like Delete.

Not covered here: renaming a parent folder that contains a Frame, and the MCP \`move\` tool on a Frame folder, still leave the Frame stale. Both would need the same delegation.

## Tests

- \`renameCanonicalFile\` on a Frame source folder: mount path updated, storage moved, separator rejected, same-name no-op.
- \`moveFrameV2SourceUsingFileSystem\` on a Pod Frame repoints tabs and the pinned Frame.
- \`ProjectMetadataResource.renameFramePath\` unit coverage.
- Existing move, edge-case and \`/files/path\` route suites pass. \`tsgo\` passes for front, front-api and front-spa.

## Risk

Low. The Frame path only triggers when a registered Frame manifest sits directly under the renamed folder; plain files and folders keep the previous behavior. The Frame move itself remains intentionally non-crash-safe (see #31532).

## Deploy Plan

Standard deploy.

### DIFF
--- a/front-api/routes/w/[wId]/files/path/[...canonicalPath].ts
+++ b/front-api/routes/w/[wId]/files/path/[...canonicalPath].ts
@@ -11,6 +11,8 @@ import {
   WriteCanonicalFileContentError,
   writeCanonicalFileContent,
 } from "@app/lib/api/files/file_system_ops";
+import type { FrameSourceMoveError } from "@app/lib/api/frames/move_source_paths";
+import { isFrameSourceMoveError } from "@app/lib/api/frames/move_source_paths";
 import { requestDustProjectIncrementalSyncForScopedPath } from "@app/lib/api/projects/request_incremental_sync";
 import type { DustFileSystemError } from "@app/types/file_system";
 import {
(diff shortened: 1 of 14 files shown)


davidebbo:
Will do differently

PR #31780, opened 2026-09-03
dust-hive: add front alias to restart front-api and front-spa-app

PopDaph:
## Description

`dust-hive restart <env> <service>` only accepts a single service, so restarting the front stack means several commands. This adds a `front` alias that restarts every `front-*` service in start order: `front-api`, `front-workers`, `front-spa-poke`, `front-spa-app`.

Aliases live in a `SERVICE_ALIASES` map in `services.ts`, so adding another one is a one-line entry. Unknown names now print both the services and aliases lists. The interactive picker is unchanged.

## Tests

`bun run check` passes (typecheck, lint, tests).

## Risk

None, dev tooling only.

## Deploy Plan

None.


fontanierh (approved):
LGTM

PR #31788, opened 2026-09-03
Revert and list-tags read the global registry only, canary build removed

flvndvd:
## Description

Two loose ends from the move to the global `dust-infra/dust-images` registry.

`revert.yml` and `list-tags.yaml` still looked at the regional repos, picking a project from the region. Every component they offer is now built once into the global registry, so both read from there only, authenticated as the dust-infra build invoker. Revert validates the tag once instead of once per region, the regions input still drives the deploy dispatch to dust-infra. list-tags loses its region input. While there they also validate the image the front releases actually run, `front-api` and `front-api-edge`, instead of the legacy `front` and `front-edge` names that stopped receiving tags when the bake split the images.

Marketing, marketing-edge and metabase leave both choice lists. Metabase has no release, no values file and no image anymore. Marketing is the one component still built per region, and the decision is to not support revert or tag listing for it rather than keep a regional code path alive for one component.

`build-canary-image.yml` is deleted rather than ported. It built per region into the regional repos, nothing in dust-infra references a canary image, and the last canary tags in the registry are old. The `-canary` suffix handling in the build action goes with it, it had no other caller.

## Tests

`actionlint` passes on both workflows. Not exercised end to end, both are manual dispatches and the change is the registry path and the image name.

## Risk

A revert to a tag built before a component moved to build-once fails validation because the tag is not in the global repo. That is the correct answer today since the deployment could not pull it either, and a `crane copy` from the regional repo is the escape hatch. Marketing can no longer be reverted through the workflow, a manual dispatch of the dust-infra revert with the regional tag still works. Rollback is a revert of this PR.

## Deploy Plan

Merge. Both workflows read their definition from main at dispatch time.


.github/workflows/build-canary-image.yml
```diff
@@ -1,162 +0,0 @@
-name: Build canary image
```
flvndvd, line 1: Will bring it back under a more polished version

.github/workflows/revert.yml
```diff
@@ -73,57 +70,38 @@ jobs:
...
-          service_account: "github-build-invoker@${{ steps.project.outputs.PROJECT_ID }}.iam.gserviceaccount.com"
+          service_account: "github-build-invoker@dust-infra.iam.gserviceaccount.com"
 
       - name: "Set up Cloud SDK"
         uses: "google-github-actions/setup-gcloud@aa5489c8933f4cc7a4f7d45035b3b1440c9c10db" # v3.0.1
 
       - name: Validate image tag exists
         run: |
-          echo "Validating image tag ${{ inputs.image_tag }} exists for ${{ inputs.component }} in ${{ matrix.region }}"
+          # The front releases run the front-api images.
```
flvndvd, line 96: We had no rollback support for `front` since Slim API work.

flvndvd:
PMRR.

PR #31789, opened 2026-09-03
Build-once stops pushing to the regional registries

flvndvd:
## Description

The build-once job stops pushing every image to the two regional registries. Since dust-tt/dust-infra#1014, #1015 and #1016 every release on the build-once path pulls from the global `dust-infra/dust-images` repo, and with dust-tt/dust#31788 revert and list-tags read from it too, so the regional copies have no reader left. The `extra_image_registries` input, the extra tag loops in the build action and the extra tag variables in the front bake file were only there for that transition and go away. The Docker auth step goes back to the single registry host.

Marketing is untouched, it still builds per region through the regional job with its own project.

## Tests

`actionlint` passes on the workflow. `docker buildx bake --print` on the bake file shows one tag per target again. Not exercised end to end, the first build-once deploy after merge is the check.

## Risk

None on the pull side, nothing reads the regional repos for these components anymore. Merge after dust-tt/dust#31788 so revert does not look for new tags in a repo that stopped receiving them. Rollback is a revert of this PR.

## Deploy Plan

Merge after #31788. Then dust-infra drops the transitional writer roles on the two regional projects, which only existed for these extra tags.

### DIFF
--- a/.github/actions/build-image/action.yml
+++ b/.github/actions/build-image/action.yml
@@ -51,10 +51,6 @@ inputs:
   contentful_access_token:
     description: "Contentful access token"
     required: false
-  extra_image_registries:
-    description: "Space-separated <host>/<project>/<repo> prefixes that also receive the image, on top of the primary registry"
-    required: false
-    default: ""
 
 outputs:
   current_version:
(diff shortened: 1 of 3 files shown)


davidebbo (approved):
LGTM

id13 (approved):

PR #31770, opened 2026-09-03
Admin access: hide prompts/skills/knowledge from admin when they cannot see an agent

fabiencelier:
## Description

Even admin should not be able to see automatically the agents details (prompt, skills, knowledge are considered private).

This is done by returning a redacted details of the agent on the endpoint `front-api/routes/w/[wId]/assistant/agent_configurations/[aId]/index.ts`

This apply everywhere we show the agent details side panel, so in the manage agent page, in the analytics, in conversation...

image

image

https://github.com/user-attachments/assets/acefacf0-dc02-4fef-9aba-3bd7d0d3fb2b

## Follow ups:
 - add audit log event for the update of skill editors: https://github.com/dust-tt/dust/pull/31802
 - Harden the skills, suggestions and YAML export endpoints: #31801.
 - Add an endpoint to simply add a user to a space: https://github.com/dust-tt/dust/pull/31805
 - Update the `manage_workspace` MCP to be consistent: https://github.com/dust-tt/dust/pull/31813

## Tests

- tested locally
- added unit tests

## Risk

Hide some content currently visible to the admins

## Deploy Plan

deploy front

PR #31801, opened 2026-09-03
agent permissions: require read access on skills, suggestions and YAML export

fabiencelier:
## Description

Three agent endpoints returned private data of agents the caller cannot read:

- `GET .../agent_configurations/[aId]/skills` had no read check: any workspace member knowing an unpublished agent's ID could list its skills, instructions included.
- `GET` and `PATCH .../agent_configurations/[aId]/suggestions` let non-editor admins read and apply suggestions, which carry instruction replacements.
- `GET .../agent_configurations/[aId]/export/yaml` let non-editor admins export the full agent (prompt, tools, knowledge, skills).

They now behave like the agent details endpoint: 404 when the agent is not readable, 403 for suggestions unless the caller is an editor. The shared agent context keeps its admin bypass because the batch model update relies on it.

Extracted from #31770, which makes the admin experience on such agents explicit.

## Tests

Added route tests for each endpoint: non-editor admin on an unpublished agent is rejected, editors still get their data.

## Risk

Medium. Permission tightening on shared endpoints. Admins lose access to the skills, suggestions and YAML export of agents they do not edit until they become editors.

## Deploy Plan

- Deploy front


philipperolet (approved):
LGTM

PR #31802, opened 2026-09-03
audit log: emit agent.editors_updated when an agent's editors change

fabiencelier:
## Description

Becoming an editor of an agent grants access to its private data (prompt, skills, knowledge), and #31770 (merged) lets admins add themselves as editors of agents they cannot read from the details panel. Editor changes were not audited.

- New `agent.editors_updated` audit event, emitted after every editors update. Metadata: agent name, scope, added and removed editor IDs, and `actor_added_self` to flag an admin granting themselves access.
- The redacted details panel now warns that becoming an editor gives access to private data and is logged.

image

## Tests

 - added unit test for audit log event
 - tested manually the new warning pop up

## Risk

Low.

## Deploy Plan

- Register the audit log schema with WorkOS before merging
- Deploy front

### DIFF
--- a/front/lib/api/assistant/configuration/agent.ts
+++ b/front/lib/api/assistant/configuration/agent.ts
@@ -1685,6 +1685,30 @@ export async function updateAgentPermissions(
       return transactionResult;
     }
 
+    // Editors get access to the agent's private data (prompt, skills, knowledge), so editor changes
+    // are audited as soon as they are committed, whatever happens to the triggers below.
+    // `actor_added_self` flags an admin granting themselves that access.
+    const actorUserId = auth.user()?.sId;
+    void emitAuditLogEvent({
+      auth,
+      action: "agent.editors_updated",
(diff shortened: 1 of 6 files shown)


front/lib/api/assistant/configuration/agent.ts
```diff
@@ -1712,6 +1712,29 @@ export async function updateAgentPermissions(
       }
     }
 
+    // Editors get access to the agent's private data (prompt, skills, knowledge), so editor changes
+    // are audited. `actor_added_self` flags an admin granting themselves that access.
+    const actorUserId = auth.user()?.sId;
+    void emitAuditLogEvent({
```
smb2268: should we move this up to ~1685? the log should probably emit if the editors are updated correctly even if trigger update doesn't go through

smb2268 (approved):
One comment, otherwise LGTM!

PR #31803, opened 2026-09-03
chore(workspace_analytics): revamp tools and skills metadata

achilleburah:
## Description

Dry the `workspace_analytics` tool descriptions and clean the `workspace-analytics` skill so it matches the current consumption index.
Removed the lists that leaked dimensions and filters into the prompt: they added maintenance effort and would drift as new ones get added.

Keeping the tool names in the skill's routing section for now, since the tool description were pruned.

Also in this PR:
- `workspace_analytics` becomes a shared constant across the metadata, the server and the registry.
- BM25 tool-search cases for `workspace_analytics` are removed: its tools are all eager, so they never go through tool search. One unrelated near-tie (`data_sources_file_system.semantic_search`) gets `maxRank: 2`.

## Tests

Tested locally.

## Risk

Low. Weaker descriptions could make the model pick the wrong tool. The tools are eager, so they always load and cannot be hidden by their metadata.

## Deploy Plan

Standard deploy. Skill version bumps to 9.

### DIFF
--- a/front/lib/actions/mcp_internal_actions/constants.ts
+++ b/front/lib/actions/mcp_internal_actions/constants.ts
@@ -100,7 +100,10 @@ import {
   WEB_SEARCH_BROWSE_SERVER,
   WEB_SEARCH_BROWSE_SERVER_NAME,
 } from "@app/lib/api/actions/servers/web_search_browse/metadata";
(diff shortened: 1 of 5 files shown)


flvndvd (approved):
Propre \o/!

PR #31809, opened 2026-09-03
[front] enh: overlay active users on consumption chart

aubin-tchoi:
## Description

This PR adds active users as an overlay of the existing consumption chart.

Example with fake data (data not very authentic past 2d ago):

image

The line chart is golden, complementary to the credits in blue. It is a straight line on a separate right axis. It has 20% headroom, which keeps it from visually sitting on top of the bars. It is not part of the cumulative view and is hidden when filtering to a single user.

The tooltip puts active users first and separates them from the credit values. The legend keeps the credit labels and active users. Hovering a point shows a white-centered golden marker with a short animation.

Also removes the "Credits over time" title as it is not accurate anymore.

## Tests

- Added tests.
- Tested locally.

## Risk

- Low.

## Deploy Plan

- Deploy front.

### DIFF
--- a/front/components/workspace/analytics/consumption/ConsumptionChart.tsx
+++ b/front/components/workspace/analytics/consumption/ConsumptionChart.tsx
@@ -463,7 +463,10 @@ export function ConsumptionDailyChart({
             strokeWidth={2}
             dot={{ r: 3, fill: "currentColor", strokeWidth: 0 }}
             activeDot={{
-              className: cn(ACTIVE_USERS_COLOR, "animate-active-users-dot-in"),
+              className: cn(
+                ACTIVE_USERS_COLOR,
+                "origin-center animate-in zoom-in-90 duration-[50ms] ease-out motion-reduce:animate-none [transform-box:fill-box]"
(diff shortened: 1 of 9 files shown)


ClementAupiais (approved):
LGTM

front/components/workspace/analytics/consumption/ConsumptionChart.tsx
```diff
@@ -175,21 +215,35 @@ function ConsumptionDailyTooltip({
...
     <ChartTooltipCard
       title={formatConsumptionDate(datum.timestamp)}
-      rows={rows.map((row) => ({
-        key: row.key,
-        label: row.label,
-        value: formatCredits(row.credits),
-        colorClassName: row.colorClassName,
-      }))}
+      rows={[
+        ...(showActiveUsers && activeUsers
```
avervaet: nit: Could we include the row when activeUsers === 0 so the tooltip don't appear empty when credits are also zero ?

avervaet (approved):
Top !

PR #31810, opened 2026-09-03
[Admin Governance] Mirror agent editor removals to grants

philipperolet:
## Description

Follows https://github.com/dust-tt/dust/pull/31730, which began writing editor additions to the new grants system.

Agent editor changes still write to the legacy `agent_editors` group while the migration is being verified. This PR mirrors removals to the `agent.editor` grant group in both workflows that can remove an editor: saving the complete editor list and using the dedicated editor update path. Multiple removals use targeted, batched membership queries, and both storage systems remain in the same transaction.

When the final configuration version of a logical agent is hard-deleted, its permission rows and grant-backed group are now deleted with its stable agent identity. Deleting only one version leaves them intact, and archiving continues to preserve editor access. Expired pending-agent cleanup removes permission rows, grant groups, and identities in set-based batches. Permission reads remain on the legacy system.

## Risks

Blast radius: Removing custom-agent editors and hard-deleting custom agents.
Risk: standard. The legacy and grant writes are transactional, and permission reads have not switched to grants yet.

## Deploy Plan

- deploy front

## Tests

- [x] `npm run test --workspace=front -- lib/resources/group_permission_resource.test.ts lib/api/assistant/configuration/agent.test.ts temporal/hard_delete/activities.test.ts` (84 tests)
- [x] `NODE_ENV=test npm run test --workspace=front-api -- editors.test.ts` (18 agent-editor tests passed; the local command later exited on missing AWS region configuration)

### DIFF
--- a/front/lib/api/assistant/configuration/agent.ts
+++ b/front/lib/api/assistant/configuration/agent.ts
@@ -31,6 +31,7 @@ import { GroupAgentModel } from "@app/lib/models/agent/group_agent";
 import { TagAgentModel } from "@app/lib/models/agent/tag_agent";
 import { AgentResource } from "@app/lib/resources/agent_resource";
 import { AgentUserRelationResource } from "@app/lib/resources/agent_user_relation_resource";
+import { GroupPermissionResource } from "@app/lib/resources/group_permission_resource";
 import { GroupResource } from "@app/lib/resources/group_resource";
 import { canReadRequestedSpaces } from "@app/lib/resources/permission_utils";
 import { SpaceResource } from "@app/lib/resources/space_resource";
@@ -934,15 +935,6 @@ export async function createAgentConfiguration(
               "Unexpected: agent should have exactly one editor group."
             );
           }
-          const newEditorIds = new Set(editors.map((editor) => editor.id));
-          removedEditors = (
-            await group.getActiveMembers(auth, {
-              transaction: t,
-            })
-          )
-            .filter((editor) => !newEditorIds.has(editor.id))
-            .map((editor) => editor.toJSON());
-
           // For pending agents updated in place, the group is already linked to the same agent ID
           // For regular updates, we need to link the group to the new agent configuration
           if (existingAgent.id !== agentConfigurationInstance.id) {
@@ -979,6 +971,7 @@ export async function createAgentConfiguration(
             );
             throw setMembersRes.error;
           }
+          removedEditors = setMembersRes.value.removedUsers;
         }
 
         const agentResource = AgentResource.fromAgentConfigurationModel(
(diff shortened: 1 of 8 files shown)


front/lib/api/assistant/configuration/agent.ts
```diff
@@ -587,8 +588,7 @@ export async function createAgentConfiguration(
 
   let userFavorite = false;
 
-  // For hidden agents, track previous editors to disable triggers when editors are removed.
```
philipperolet, line 630: keep this useful comment

front/lib/api/assistant/configuration/agent.ts
```diff
@@ -599,16 +599,6 @@ export async function createAgentConfiguration(
     });
     if (existingAgent) {
       currentScope = existingAgent.scope;
-      if (scope === "hidden") {
```
philipperolet, line 642: why is that code not needed anymore vs was needed before? could triggers have used setmembersres.value.removedusers also before?

philipperolet: surprised that the logic was gated behind a "scope === hidden", what's that about?

front/lib/resources/agent_resource.ts
```diff
@@ -135,6 +135,54 @@ export class AgentResource implements WithAccessControl {
...
+      resourceType: "agent",
+      resourceId: this.id,
+      transaction,
+    });
+    if (revokeResult.isErr()) {
+      throw revokeResult.error;
+    }
+  }
+
+  async deletePermissions(
```
philipperolet: maybe call that destroyPermissionsAndGroups?

philipperolet: or something showing that we remove the regular auto group so it's not used for other purposes than fully deleting an agent (since a live agent without a regular auto group would be a broken state i think?)

front/lib/resources/group_resource.ts
```diff
@@ -1835,17 +1836,22 @@ export class GroupResource extends BaseResource<GroupModel> {
...
     }
 
-    // Check if the users are already a member of the group.
-    const activeMembers = await this.getActiveMembers(auth);
-    const activeMembersIds = activeMembers.map((m) => m.sId);
-    const notActiveUserIds = userIds.filter(
-      (userId) => !activeMembersIds.includes(userId)
-    );
-    if (notActiveUserIds.length > 0) {
+    const groupMembershipCount = await GroupMembershipModel.count({
```
philipperolet, line 1844: why this change and why did you remove the comment?

front/lib/resources/agent_resource.ts
```diff
@@ -135,6 +135,58 @@ export class AgentResource implements WithAccessControl {
     }
   }
 
+  async revokeEditors(
```
fabiencelier, line 138: I find it good practice to add direct unit test on the methods of a resource as it code that will be used in many places so we want to make sure it behaves as expected and do not change un-expectedly later

fabiencelier (commented):
LGTM ✅

PR #31813, opened 2026-09-03
workspace_management MCP: redact private agent fields for admins in get_agent_details

fabiencelier:
## Description

In the `workspace_management` MCP, `list_agents` with `all_unrestricted` already lists every agent for admins, but `get_agent_details` was inconsistent with it:

- agents built on a space the admin is not a member of came back as "no agent found";
- unpublished agents the admin does not edit came back with a hand-written partial view that hid the description, which admins may see.

`get_agent_details` now uses `getAgentConfigurationForDetails`, the same fetcher as the agent details endpoint (#31770, merged). Admins get name, description, scope and model for agents they cannot read, with instructions, skills, tools and knowledge withheld and a line explaining why. There is no way to override this from the tool, and the tool description says so. Non-admins keep the not-found answer.

## Tests

 - Added three tool tests

## Risk

Low. Admin-only behavior change in one MCP tool, reusing the redaction already reviewed in #31770.

## Deploy Plan

- Deploy front

### DIFF
--- a/front/lib/api/actions/servers/workspace_management/metadata.ts
+++ b/front/lib/api/actions/servers/workspace_management/metadata.ts
@@ -196,7 +196,11 @@ export const WORKSPACE_MANAGEMENT_TOOLS_METADATA = [
     description:
       "Return an agent's full configuration: name, description, scope, model, " +
       "equipped skills and capabilities, and its complete system prompt and " +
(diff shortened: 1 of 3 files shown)


achilleburah (approved):
LGTM !

PR #31814, opened 2026-09-03
fix(front): restore inline model makers on mobile clients

Nils-Fedrigo:
## Description

Screenshot 2026-09-03 at 17 16 14

On mobile the model picker's "More models" list can't be used: each maker renders as a
`DropdownMenuSub`, and a Radix submenu only opens on pointer enter, which touch input never
produces. Worse, when the submenu does open on a narrow viewport it lands off-screen to the
right (a 336px menu plus a 256px submenu), and Radix focusing it makes the browser scroll the
document sideways — the whole layout appears shifted left.

This is a regression from #31378, which replaced `ModelPickerMoreModels` (that had a dedicated
`useIsWidthConstrained()` branch expanding make
rs inline) with submenus for every client.

Restores the inline behaviour, gated on the capability the affordance actually needs rather than
on width alone: makers expand in place when the viewport is narrow **or** the pointer can't hover
(`(hover: hover) and (pointer: fine)`, the query already used in `sparkle/src/styles/tailwind.css`).
The hover gate also covers touch on wide screens — a tablet, or a phone in landscape — which the
768px breakpoint missed. Desktop keeps the hover submenus.

- `useModelPickerMenuState` now holds the expanded maker (one at a time, as before #31378) and
  clears it on open, so all three call sites (input bar picker, `SetModelAssistantsDialog`,
  `ModelsFilterMenu`) get the fix through `ModelPickerContent`.
- Model-row rendering is shared between both branches, so lock reasons, degradation badges,
  selection indicators and the effort slider behave identically either way.

## Tests
Manual 

## Deploy plan 

Standard


front/components/model_picker/ModelPickerContent.tsx
```diff
@@ -45,6 +46,10 @@ interface ModelPickerContentProps {
   streams: ModelStreamResolutionsType | null;
   isMakersExpanded: boolean;
   onToggleMakers: () => void;
+  // Inline maker expansion, used on width-constrained clients where makers
+  // can't open as submenus.
```
Nils-Fedrigo:
```suggestion
```

front/components/model_picker/ModelPickerMakersView.tsx
```diff
@@ -46,65 +57,105 @@ export function ModelPickerMakersView({
...
-        const hasDegradedModel = maker.models.some((model) =>
-          degradedModelIds.has(model.modelId)
-        );
-        return (
-          <DropdownMenuSub key={maker.makerId}>
-            <DropdownMenuSubTrigger
+  // Submenus open on pointer enter, so touch input can never reach them, and on
+  // a narrow viewport they would open off-screen (focusing them then scrolls the
+  // page sideways). Either way, makers expand inline instead.
+  const isWidthConstrained = useIsWidthConstrained();
```
aubin-tchoi, line 70: not really tied to having a narrow viewport, no one does this but if you open the page in landscape mode or on a tablet you may still be stuck

front/components/model_picker/ModelPickerMakersView.tsx
```diff
@@ -46,65 +57,105 @@ export function ModelPickerMakersView({
...
+  const useInlineMakers = isWidthConstrained;
+
+  const makerIcon = (maker: MakerGroup) => {
+    const makerLogo = getModelMakerLogo(maker.makerId, isDark);
+    const hasDegradedModel = maker.models.some((model) =>
+      degradedModelIds.has(model.modelId)
+    );
+    return hasDegradedModel ? (
+      <DegradedModelIcon icon={makerLogo} surface="overlay-background" />
```
aubin-tchoi: what is the `surface` prop (seems new)?

front/components/model_picker/useModelPickerMenuState.ts
```diff
@@ -1,16 +1,25 @@
...
 
   const resetMenu = () => {
     setIsMakersExpanded(false);
+    setExpandedMakerId(null);
   };
 
   return {
     menuStateProps: {
       isMakersExpanded,
       onToggleMakers: () => setIsMakersExpanded((expanded) => !expanded),
```
aubin-tchoi: `onToggleMakers` should also reset `expandedMakerId`

front/components/model_picker/ModelPickerMakersView.tsx
```diff
@@ -46,65 +57,105 @@ export function ModelPickerMakersView({
...
+    ) : (
+      makerLogo
+    );
+  };
+
+  const renderModelRows = (maker: MakerGroup) =>
```
aubin-tchoi, line 86: we need a left padding on each row in the mobile branch here to align the name of the model with the name of the provider

front/components/model_picker/ModelPickerMakersView.tsx
```diff
@@ -46,65 +57,105 @@ export function ModelPickerMakersView({
...
+          onRevert={isSelected ? selection.onRevert : undefined}
+        />
+      );
+    });
+
+  if (useInlineMakers) {
+    return (
+      <>
+        {makerGroups.map((maker) => (
+          <Fragment key={maker.makerId}>
```
aubin-tchoi, line 123: OOC no width here?

aubin-tchoi (commented):
thanks 🙏🏻

front/components/model_picker/ModelPickerMakersView.tsx
Nils-Fedrigo: needed for the icon to be properly following the theme

front/components/model_picker/ModelPickerMakersView.tsx
Nils-Fedrigo: const [isMobile, setIsMobile] = useState<boolean>(
    () => !!window && window.innerWidth < MOBILE_BREAKPOINT
  );


the width constraint is actual width constraint, so we're good here I think

front/components/model_picker/ModelPickerMakersView.tsx
Nils-Fedrigo: That's a possibility, but do we really need that?

front/components/model_picker/ModelPickerMakersView.tsx
Nils-Fedrigo: ok I think I understand, a prev version of this PR introduced a useHover, lemme restore it

front/components/model_picker/ModelPickerMakersView.tsx
Nils-Fedrigo: for now merging without it

front/components/model_picker/ModelPickerMakersView.tsx
Nils-Fedrigo: yep it's part of the main menu so does not need a width

front/components/model_picker/ModelPickerMakersView.tsx
Nils-Fedrigo: did it in the end

front/components/model_picker/ModelPickerMakersView.tsx
aubin-tchoi: the "surface" prop describes who the consumer is here?

aubin-tchoi: wfm, the confusion came from the fact that it's a name typically used to describe whether the content will be raised, overlaid, inverse, etc..

PR #31815, opened 2026-09-03
[front] enh: name the conflicting tool in MCP name-conflict error

tdraier:
## Description

Fixes dust-tt/tasks#10265

When saving an MCP server connection fails with a tool-name conflict, Dust only echoed the name the user typed and never surfaced which existing tool it collided with. 

The conflict check (`MCPServerViewResource.hasNameConflictInSpaceByName`) already computed both sides of every collision but discarded them, returning only `{ hasConflict, name }`. This PR returns `conflictDetails` — the existing connection's display name, plus (for the cropped-tool case) the shared model-facing tool name both servers resolve to after prefix truncation — and threads it through the create and edit flows to the UI.

Message before → after (the hodor-shield cropped-tool case):
- Before: `This name conflicts with an existing Tool. Enter a different name.`
- After: `This name produces the tool "hodor_shield_product__shield_run_full_security_policy_evaluation", which already exists on the connection "hodor-shield-...-alpha". Enter a different name.`

Screenshot 2026-09-03 at 16 58 36

## Tests

- Updated the two name-conflict assertions in `front-api/routes/w/[wId]/mcp/index.test.ts` to cover both the cropped-tool-name collision (asserts `conflictDetails.conflictingServerName` and `conflictingToolName`) and the plain same-name collision.
- `npm run test` green for: mcp POST handler (11), submit-form (4), rename/view/server-view edit paths (22).
- `tsgo --noEmit` clean on `front` and `front-api`; `biome check --changed` clean.
- Verified live against a local remote MCP test server by connecting two similarly-named "hodor-shield" instances and confirming the create dialog now names the colliding tool.

## Risk

Low. No schema or API-contract removals — the change only enriches an existing error path. The HTTP `nameConflict` payload gains an optional `conflictDetails` field (`name` unchanged), and all new fields are optional, so older clients keep working. Fully rollback-safe.

## Deploy Plan

Standard deploy. front and front-api ship together; no migration, no ordering constraints.

### DIFF
--- a/front-api/routes/w/[wId]/mcp/index.test.ts
+++ b/front-api/routes/w/[wId]/mcp/index.test.ts
@@ -313,7 +313,18 @@ describe("POST /api/w/:wId/mcp/ — name conflict", () => {
     expect(response.status).toBe(400);
     const body = await response.json();
     expect(body.error.message).toContain(candidateName);
-    expect(body.nameConflict).toEqual({ name: candidateName });
+    // Cropped tool-name collision: the response names the existing connection
+    // and the shared model-facing tool name.
+    expect(body.nameConflict.name).toBe(candidateName);
+    expect(body.nameConflict.conflictDetails.conflictingServerName).toBe(
+      existingName
+    );
+    expect(body.nameConflict.conflictDetails.conflictingToolName).toContain(
+      "test_tool"
+    );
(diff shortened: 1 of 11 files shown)


tdraier:

- #31815 ◀

Fraggle (approved):
LGTM

PR #31816, opened 2026-09-03
Prodbox prompt shows the cell

flvndvd:
## Description

The prodbox prompt shows the cell next to the region, `prodbox(europe-west1 cell-00002):~$`. With cells arriving there will be several prodboxes per region and the prompt is the one thing you look at before typing into the wrong one. The pod already receives `CELL` from its topology label, same as `REGION`, so this is the prompt line only. When the label is absent the prompt stays as it is today, `prodbox(us-central1):~$`.

## Tests

Rendered the prompt line locally with and without `CELL` set. Both legacy prodbox pods carry the cell label (cell-00000, cell-00001), checked on the clusters.

## Risk

A prompt string. Rollback is a revert.

## Deploy Plan

Next prodbox deploy.

### DIFF
--- a/prodbox/init.sh
+++ b/prodbox/init.sh
@@ -21,7 +21,7 @@ git remote set-url origin git@github.com:dust-tt/dust.git
 
 git pull origin main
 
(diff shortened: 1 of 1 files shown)


Fraggle (approved):
Lovely :)

PR #31818, opened 2026-09-03
Poke: hide credit-pool cards and previous-cycles table for legacy no-pool workspaces

avervaet:
Workspaces with no active credit pool, no Metronome contract, and a legacy
(non credit-priced) plan only enforce a premium-message rate limit — there's
no pool or Metronome usage to show, so the header cards and previous-cycles
table just rendered empty for them.

PR #31819, opened 2026-09-03
Poke: replace credit usage with premium message usage for legacy no-pool workspaces

avervaet:
## Description

For workspaces with no pool, no Metronome contract, and a legacy plan, the members table showed a credit-usage column and a pace-warning column that have no meaning without a credit pool. Swap the credit column for a new
premium message usage column.

**Visual:**
Capture d’écran 2026-09-04 à 10 18 24

## Tests

Tested on front-edge

## Risk

low additons only scoped to poke

## Deploy Plan

- Deploy front

### DIFF
--- a/front/components/poke/pages/PoolUsagePage.tsx
+++ b/front/components/poke/pages/PoolUsagePage.tsx
@@ -198,12 +198,6 @@ export function PoolUsagePage() {
 
   const { data: workspaceInfo } = usePokeWorkspaceInfo({ owner });
   const activeSubscription = workspaceInfo?.activeSubscription;
-  // Many legacy (non credit-priced) workspaces still carry a Metronome
-  // contract used only for usage diagnostics (Stripe stays the billing
-  // source of truth), so a raw `metronomeContractId !== null` check matches
-  // almost every legacy workspace. `isSubscriptionMetronomeBilled` excludes
-  // those Stripe-billed shadow contracts and only reports a real,
-  // billing-active Metronome contract.
   const hasMetronomeContract =
     !!activeSubscription && isSubscriptionMetronomeBilled(activeSubscription);
   const isLegacyPremiumMessagePlan =
@@ -213,9 +207,6 @@ export function PoolUsagePage() {
   // cards and previous-cycles table would just render empty for them.
   const isLegacyWithoutPoolOrMetronome =
     isLegacyPremiumMessagePlan && !hasMetronomeContract;
-  // Default to hidden until the subscription loads, so the pool-cycle
-  // endpoints (which 400 for workspaces with no real Metronome contract)
-  // aren't fetched before we know whether this workspace has one.
   const showPoolSection =
     !!activeSubscription && !isLegacyWithoutPoolOrMetronome;
 
@@ -225,9 +216,6 @@ export function PoolUsagePage() {
   });
   const hasPool = (awuPoolCurrentCycle?.totalActiveCredits ?? 0) > 0;
 
-  // Apply the "premium message usage" default sort once, the first time we
-  // learn this workspace is legacy-without-pool — without clobbering a sort
-  // the user picks afterwards.
   const hasAppliedDefaultSortRef = useRef(false);
   useEffect(() => {
     if (hasAppliedDefaultSortRef.current || !activeSubscription) {
--- a/front/components/poke/pages/PoolUsagePage.tsx
+++ b/front/components/poke/pages/PoolUsagePage.tsx
@@ -155,9 +155,6 @@ export function PoolUsagePage() {
     pageIndex: 0,
     pageSize: DEFAULT_PAGE_SIZE,
   });
-  // Empty until the user picks a column: the default depends on
-  // `isLegacyWithoutPoolOrMetronome`, which is derived below once workspace
-  // info has loaded.
   const [sorting, setSorting] = useState<SortingState>([]);
 
   // Debounce the search input, and reset to the first page on a new query.
--- a/front/components/poke/pages/PoolUsagePage.tsx
+++ b/front/components/poke/pages/PoolUsagePage.tsx
@@ -33,7 +33,10 @@ import {
   SEAT_TYPE_ORDER,
   toBaseSeatType,
 } from "@app/types/memberships";
-import { isCreditPricedPlan } from "@app/types/plan";
+import {
+  isCreditPricedPlan,
+  isSubscriptionMetronomeBilled,
+} from "@app/types/plan";
 import {
   AlertCircle,
   Button,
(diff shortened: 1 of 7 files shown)


avervaet:

- #31818
    - #31819 ◀
        - #31833
            - #31834
            - #31847

front/lib/api/credits/members_usage.ts
```diff
@@ -2037,6 +2055,22 @@ async function resolveMembersUsagePageUsers({
       }
       break;
     }
+    case "premiumMessageUsage": {
+      const entries = await concurrentExecutor(
+        allUsers,
```
PopDaph: Quick question on the sort case: iiuc allUsers is the full ES scan of the workspace (searchAllUsers pages until total), so this reads one Redis key per member, 8 at a time. The other concurrentExecutors in getMembersUsage run over the page (≤ 150), but this one runs over every member, so on a large workspace that's thousands of round-trips each time this column is the sort key.

Do we need sorting on this column for poke? If not, dropping enableSorting would remove the full-set read and the page-level block below would be enough on its own. If sorting is required, would a batched read work?

front/components/poke/pages/PoolUsagePage.tsx
```diff
@@ -187,25 +190,42 @@ export function PoolUsagePage() {
...
+  const showPoolSection =
+    !!activeSubscription && !isLegacyWithoutPoolOrMetronome;
+
+  const { awuPoolCurrentCycle } = usePokeAwuPoolCurrentCycle({
+    owner,
+    disabled: !showPoolSection,
+  });
+  const hasPool = (awuPoolCurrentCycle?.totalActiveCredits ?? 0) > 0;
+
+  const hasAppliedDefaultSortRef = useRef(false);
```
PopDaph: Could the default sort be derived instead of set in an effect? For example, start `sorting` as `null` (nothing picked yet) and fall back to the plan's default when reading it:

```ts
const defaultSortId = isLegacyWithoutPoolOrMetronome
  ? "premiumMessageUsage"
  : "consumedFromPoolAwuCredits";
const sort = sorting?.[0] ?? { id: defaultSortId, desc: true };
```

No ref, no effect, and it avoids the double fetch: today the first members request goes out sorted by pool credits, then a second one once the effect flips it.

front/components/poke/pages/PoolUsagePage.tsx
```diff
@@ -187,25 +190,42 @@ export function PoolUsagePage() {
...
       : "name";
   const orderDirection = sort?.desc ? "desc" : "asc";
 
-  const { awuPoolCurrentCycle } = usePokeAwuPoolCurrentCycle({ owner });
-  const hasPool = (awuPoolCurrentCycle?.totalActiveCredits ?? 0) > 0;
-
   const { data: workspaceInfo } = usePokeWorkspaceInfo({ owner });
   const activeSubscription = workspaceInfo?.activeSubscription;
-  const hasMetronomeContract = activeSubscription?.metronomeContractId != null;
+  const hasMetronomeContract =
```
PopDaph, line 189: Is this change intentional?

front/components/workspace/MembersUsageTable.tsx
```diff
@@ -747,6 +749,74 @@ function buildPoolCreditUsageColumn(
...
+      );
+    }
+    const { usedMessages, limitMessages, nextRefill } = usage;
+    const percentage =
+      limitMessages > 0
+        ? Math.min(100, (usedMessages / limitMessages) * 100)
+        : usedMessages > 0
+          ? 100
+          : 0;
+    const isOverLimit = usedMessages > limitMessages;
```
PopDaph: Can `usedMessages` be > `limitMessages`

front/lib/api/credits/members_usage.ts
avervaet: We want it as one of the main goal of the page is to allow admin to grasp quickly who is stucked. 

Will implement the batch read.

front/components/poke/pages/PoolUsagePage.tsx
avervaet: Yes, we missed legacy plan in the initial framing of the page.

front/lib/api/credits/members_usage.ts
```diff
@@ -2037,6 +2048,17 @@ async function resolveMembersUsagePageUsers({
       }
       break;
     }
+    case "premiumMessageUsage": {
+      // Count-only and pipelined into a single Redis round-trip
+      const usedCountByUserId = await getPremiumModelMessageUsedCountsByUser({
+        workspace,
+        users: allUsers.map((u) => u.toJSON()),
```
PopDaph: NIT: do we need to serialize all users (can be a lot) just to get their id and sId?

front/lib/utils/rate_limiter.ts
```diff
@@ -249,6 +249,40 @@ export async function getRateLimiterCount({
...
+    const uniqueKeys = Array.from(new Set(keys));
+    const pipeline = redis.multi();
+    for (const key of uniqueKeys) {
+      pipeline.zCount(makeRateLimiterKey(key), trimBeforeMs, "+inf");
+    }
+    const counts = (await pipeline.exec()) as number[];
+
+    return new Ok(
+      new Map(uniqueKeys.map((key, index) => [key, counts[index]]))
+    );
```
PopDaph: NIT: Could we drop the `as number[]`? We try to avoid `as` casts. Checking each reply instead would look like:

```ts
const replies = await pipeline.exec();
return new Ok(
  new Map(
    uniqueKeys.map((key, index) => {
      const reply = replies[index];
      return [key, typeof reply === "number" ? reply : 0];
    })
  )
);
```

front/lib/utils/rate_limiter.ts
```diff
@@ -249,6 +249,40 @@ export async function getRateLimiterCount({
...
+    return new Ok(new Map());
+  }
+
+  try {
+    const redis = await getRedisStreamClient({ origin: "rate_limiter" });
+    const windowMs = timeframeSeconds * 1000;
+    const trimBeforeMs = Date.now() - windowMs;
+
+    const uniqueKeys = Array.from(new Set(keys));
+    const pipeline = redis.multi();
```
PopDaph: Should we chunk the keys? One `MULTI` with thousands of `ZCOUNT` is fine for Redis, but a small constant (a few hundred keys per batch, like the other bulk helpers I think) would keep the payload bounded on very large workspaces.

front/components/poke/pages/PoolUsagePage.tsx
```diff
@@ -183,29 +184,42 @@ export function PoolUsagePage() {
...
+    !!activeSubscription && isSubscriptionMetronomeBilled(activeSubscription);
   const isLegacyPremiumMessagePlan =
     !!activeSubscription && !isCreditPricedPlan(activeSubscription.plan);
   // Users with no pool and no Metronome contract on a legacy plan only have
```
PopDaph: Is this comment still accurate? The condition below no longer checks `hasPool`.

front/components/workspace/MembersUsageTable.tsx
```diff
@@ -747,6 +750,92 @@ function buildPoolCreditUsageColumn(
...
 
+const premiumMessageUsageColumn: ColumnDef<RowData, string> = {
+  id: "premiumMessageUsage" as const,
+  header: () => (
+    <span className="flex items-center gap-1">
+      <Icon visual={CoinsStacked03} size="xs" />
+      Premium messages
+      <Tooltip
+        trigger={<Icon visual={InfoCircle} size="xs" />}
+        label="Usage is capped over a rolling 7-day window: each message frees up its slot 7 days after it was sent, not all at once."
```
PopDaph: nit: should we use `windowDays` from the payload here instead of hardcoding 7 days? Same for the "this week" in the aria text below.

PopDaph (commented):
Approving already!
(I did not test is locally I only reviewed the code 🫡 )

PopDaph (approved):

PR #31824, opened 2026-09-03
fix(workspace_analytics): fix linting error in metadata

achilleburah:
## Description

Fix linting error on main introduced by chore(workspace_analytics): revamp tools and skills metadata (#31803)

## Risk

Lint only.


aubin-tchoi (approved):
LGTM

PR #31826, opened 2026-09-03
feat(model_picker): knock the degraded badge out of its surface

Nils-Fedrigo:
## Description
Screenshot 2026-09-03 at 18 08 43
Screenshot 2026-09-03 at 18 12 26

The degraded-model badge (`InfoCircle` filled with `info`) knocked its glyph out in
plain white. That only holds on a light surface: in dark mode the white ring the
glyph draws around itself read as a bright halo pasted onto the icon, and the badge
sat visually on top of the provider logo rather than beside it.

`DoubleIcon` now takes the `surface` it is rendered on and knocks a filled badge's
glyph out in that surface's color, so the halo matches whatever is behind the icon in
both themes. Surfaces are the existing background tokens (`overlay-background`,
`panel-background`, …) plus `current`, for callers whose surface paints its own color
— they set it as a `text-*` class on the `DoubleIcon`.

The expanded input bar is one such caller: it painted a raw `oklch(...)` pair inline.
That value moves to a `--color-input-bar-background` token in `theme-extras.css`
(the one entry every app's Tailwind pass imports), so the composer and the glyph
knocked out of it read the same value.

Also on the badge: `secondarySize` lets the badge keep its full size instead of
scaling down from `size`. On the 16px composer trigger the badge carries the entire
degradation signal, so it now matches the provider icon rather than shrinking below it.

- `DegradedModelIcon` takes a `surface` prop (`composer` | `menu`) and maps it to the
  right knockout.
- `DegradedInfoIcon` (the picker row) knocks out in `overlay-background` instead of
  white — nothing sits behind it, so the halo correctly disappears and it reads as a
  plain filled disc.
- Storybook: `surface` / `secondarySize` controls, updated guidance, and a
  `StatusBadges` row showing the composer badge over a non-default surface.

## Tests

Visual — check the degraded badge in the model picker button and in the makers list,
in both light and dark mode, and confirm the badge halo matches the surface behind it
(input bar surface for the composer trigger, menu background inside the picker).

### DIFF
--- a/front/components/assistant/conversation/input_bar/InputBar.tsx
+++ b/front/components/assistant/conversation/input_bar/InputBar.tsx
@@ -718,8 +718,8 @@ export const InputBar = React.memo(function InputBar({
             : classNames(
                 "rounded-squircle-40 w-full overflow-hidden",
                 "border",
-                "has-[.tiptap:focus]:bg-stone-25 bg-[oklch(0.988_0_89.876)]",
-                "dark:has-[.tiptap:focus]:bg-[oklch(0.310_0.007_75)] dark:bg-[oklch(0.294_0.008_84.593)]",
+                "bg-input-bar-background",
+                "has-[.tiptap:focus]:bg-stone-25 dark:has-[.tiptap:focus]:bg-[oklch(0.310_0.007_75)]",
                 isFloating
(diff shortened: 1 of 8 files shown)


frankaloia (approved):
LGTM

ClementAupiais (approved):
LGTM

PR #31827, opened 2026-09-03
[front] fix: make analytics filter search accent-insensitive

flvndvd:
## Description

- Make the analytics Filters picker search accent-insensitive.
- Normalize both the search query and option names before matching.
- Add regression coverage for accented, unaccented, case-insensitive, and unrelated searches.

## Tests

- npm run test -- components/workspace/analytics/usageFilter.test.ts components/workspace/analytics/filterPanel/filterState.test.ts (34 passed)
- Biome on the three changed files

## Notes

The repository-wide front type-check is currently blocked by unrelated existing errors outside these changed files.


aubin-tchoi (approved):
LGTM

PR #31828, opened 2026-09-03
Fix marketing auth context cross-region redirect

smb2268:
## Description

A user whose session cookie points at a workspace in the other region was shown the logged-out landing page on the marketing site. Typically a user with accounts in both regions who logs into one, then visits the other's marketing site. Follow up of `27268` / `27389`.

**Root cause:** marketing calls `/api/auth-context` on its own region's front-api. When the sealed `workspaceId` in the shared `workos_session` cookie lives in the other region, front-api answers `400 workspace_in_different_region` with `redirect.url`. Marketing treated any non-2xx as anonymous.

**Fix:** `resolveAuthContext`, shared by SSR and the landing SWR fetcher, follows that redirect exactly once. Direction-agnostic: US marketing → EU and EU marketing → US use the same path, since both front-apis run the same lookup and the resolver follows whatever URL comes back. `redirect.url` is trusted as-is since front-api builds it from its own region config. SSR now also warns on 5xx so a degraded front-api is distinguishable from an anonymous visitor.

## Tests

- `Lint & Build (marketing)` (tsgo + next build).
- Manual: US and EU sessions on both `dust.tt` and `eu.dust.tt`, anonymous, expired.

## Risk

Worst case, the regional retry fails or times out and we fall back to today's behavior (landing page). SSR worst case is 2 × 1.5s. Safe to rollback.

## Deploy Plan

- [ ] Deploy marketing (both regions)
- [ ] Verify cross-region session on each marketing site: `400` then `200` from the other region, header shows logged-in state

### DIFF
--- a/marketing/lib/api/authContext.ts
+++ b/marketing/lib/api/authContext.ts
@@ -35,6 +35,88 @@ const AuthContextResponseSchema = z.object({
   defaultWorkspaceId: z.string().nullable().optional(),
 });
 
+const RegionRedirectResponseSchema = z.object({
+  error: z.object({
+    type: z.literal("workspace_in_different_region"),
+    message: z.string(),
+    redirect: z.object({
+      region: z.string(),
+      url: z.string().url(),
+    }),
+  }),
+});
+
+export type MarketingRegionRedirect = z.infer<
+  typeof RegionRedirectResponseSchema
+>["error"]["redirect"];
+
+type AuthContextResponse =
+  | { type: "success"; authContext: MarketingAuthContext }
+  | { type: "region_redirect"; redirect: MarketingRegionRedirect }
+  | null;
+
+/**
+ * Parse the auth-context response while preserving the cross-region routing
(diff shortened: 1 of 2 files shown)


frankaloia (approved):

marketing/lib/api/authContext.ts
```diff
@@ -35,6 +41,100 @@ const AuthContextResponseSchema = z.object({
   defaultWorkspaceId: z.string().nullable().optional(),
 });
 
+// front-api answers 400 with this body when the session's workspace lives in
+// the other region. Mirrors `RegionRedirectError` in front/types/error.ts.
+const RegionRedirectSchema = z.object({
```
Fraggle, line 46: FYI, now we work with "cells" instead of region. Can you update this ?

Fraggle: (note: the front-api PR that will update workspace_in_different_region => workspace_in_different_cell) has not been merged)
You can still follow the same `url`property but the schema also has a `name` property (the cell name).

PR #31831, opened 2026-09-03
Fix tracking extra/properties

theogz:
## Description

Follow-up from https://github.com/dust-tt/dust/pull/31734. The `extra` properties need to be formatted a certain way for Snowflake and PostHog to be able to consume them properly.
Also automatically passes workspace sId when available.

## Tests

Will look at new events post merge

## Risk

Low.

## Deploy Plan

Normal


aubin-tchoi (approved):
LGTM

PR #31822, opened 2026-09-03
feat(core): support BigQuery auth via ADC/WIF

Fraggle:
## Description

Addresses Flavien's request to reduce hard dependency on mounted service-account JSON in Rust GCP paths.

### `core` (BigQuery)
- Upgrades `gcp-bigquery-client` from `0.25.1` to `0.28.0`.
- Updates BigQuery remote DB initialization to:
  - keep the existing service-account JSON path when SA fields are present
  - fall back to Application Default Credentials (ADC) when they are absent, enabling WIF/metadata-based auth

### `egress-proxy` (GCS policy reads)
- Upgrades `gcp_auth` from `0.12.6` to `0.12.7`.
- Tightens static-token handling so blank `GOOGLE_CLOUD_ACCESS_TOKEN` values do not disable provider setup.
- Clarifies missing-credentials error messaging to explicitly include metadata/ADC/WIF.

### Cloud Storage crate audit
- I also checked `core`'s `cloud-storage = 0.11.1` usage (table/document GCS operations).
- This crate version still expects service-account style env credentials and does **not** expose first-class metadata/ADC/WIF behavior in its default path, so I did **not** silently bump/swap it in this PR.
- If desired, we should do a follow-up migration PR for `core` GCS operations (either custom token cache wiring or crate migration) to fully align those paths with WIF.

## Tests

- `cargo check --manifest-path core/Cargo.toml`
- `cargo check --manifest-path egress-proxy/Cargo.toml`
- `cargo test --manifest-path egress-proxy/Cargo.toml healthz_returns_ok -- --nocapture`

## Risk

Low-to-medium:
- `core` has a dependency bump plus a targeted auth-path fallback.
- `egress-proxy` bump is small and behavior-preserving, with stricter handling of blank static tokens.
- Legacy service-account JSON paths remain supported.

## Deploy Plan

1. Deploy `core` and `egress-proxy` normally.
2. Validate runtime identities have expected GCP permissions.
3. For WIF operation, avoid relying on mounted SA JSON files where possible; use workload identity/ADC paths.

[Slack Thread](https://dust4ai.slack.com/archives/C0BRB4T824R/p1788449443774299?thread_ts=1788449443.774299&cid=C0BRB4T824R)

[...]

PR #31832, opened 2026-09-03
feat(core): authenticate GCS via ADC/WIF

Fraggle:
## Description

I route all `core` GCS operations through a shared `gcs_client()` backed by `cloud_storage::Client::with_cache(...)` and `gcp_auth` ADC. This supports service-account JSON, gcloud ADC, GCE/GKE metadata, and WIF, while retaining `GOOGLE_CLOUD_ACCESS_TOKEN` for local and test overrides. I replace global `Object::*` calls across table CSV storage, document storage, SQLite streaming, background cleanup, and migration tooling, and remove `cloud-storage`'s `global-client` feature. Related: dust-tt/dust#31822.

## Tests

- Ran `cargo check --manifest-path core/Cargo.toml`.
- Ran `cargo test --manifest-path core/Cargo.toml --lib databases::csv`.

## Risk

Medium: production GCS reads and writes now use the shared ADC-backed client. Existing `GOOGLE_APPLICATION_CREDENTIALS` support remains; a misconfigured ADC or WIF environment can block table or document access. Rollback is a standard `core` deploy.

## Deploy Plan

Deploy `core`. Ensure the runtime identity has GCS access and ADC/WIF is configured for the tables and data-source buckets. Keep the mounted service-account JSON until WIF is verified; remove it afterward if desired. No SQL or infrastructure migration.

### DIFF
--- a/core/src/gcs_client.rs
+++ b/core/src/gcs_client.rs
@@ -1,0 +1,135 @@
+use anyhow::{Context, Result};
+use async_trait::async_trait;
+use cloud_storage::{Client, TokenCache};
+use gcp_auth::TokenProvider;
+use std::sync::Arc;
+use tokio::sync::{OnceCell, RwLock};
+
+const GCS_SCOPE: &str = "https://www.googleapis.com/auth/devstorage.full_control";
+
+static GCS_CLIENT: OnceCell<Client> = OnceCell::const_new();
+
+/// Shared GCS client authenticated via Application Default Credentials.
+///
+/// Supports:
+/// - `GOOGLE_CLOUD_ACCESS_TOKEN` (tests / local overrides)
+/// - `GOOGLE_APPLICATION_CREDENTIALS` service-account JSON
+/// - gcloud application-default credentials
+/// - GCE/GKE metadata server (including Workload Identity Federation)
+pub async fn gcs_client() -> Result<&'static Client> {
+    GCS_CLIENT
+        .get_or_try_init(|| async { build_gcs_client().await })
+        .await
+}
+
+async fn build_gcs_client() -> Result<Client> {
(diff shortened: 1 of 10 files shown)


core/src/gcs_client.rs
```diff
@@ -0,0 +1,135 @@
...
+        .get_or_try_init(|| async { build_gcs_client().await })
+        .await
+}
+
+async fn build_gcs_client() -> Result<Client> {
+    if let Some(token) = get_static_access_token_from_env() {
+        return Ok(Client::with_cache(StaticAccessTokenCache::new(token)));
+    }
+
+    let provider = gcp_auth::provider()
```
flvndvd: `gcp_auth::provider()` only reads service-account files through `GOOGLE_APPLICATION_CREDENTIALS`. The previous crate also accepted `SERVICE_ACCOUNT` and JSON variants. Both dev and prod exposes `SERVICE_ACCOUNT` but not `GOOGLE_APPLICATION_CREDENTIALS` can we stick with `SERVICE_ACCOUNT`?

core/src/gcs_client.rs
```diff
@@ -0,0 +1,135 @@
...
+        GCS_SCOPE.to_string()
+    }
+
+    async fn fetch_token(
+        &self,
+        _client: &reqwest011::Client,
+    ) -> cloud_storage::Result<(String, u64)> {
+        let token = self
+            .provider
+            .token(&[GCS_SCOPE])
```
flvndvd, line 186: Don't we risk many callers to queue refresh tokens, IMHO we should wrap with a mutex here. So one refresh occurs and waiters reuse it.

core/src/gcs_client.rs
flvndvd: + can we add some tests around this?

PR #31834, opened 2026-09-03
Share the credit-pool card logic between Poke and the compact usage page

avervaet:
Poke's PoolCreditCard and the customer-facing CompactCreditPoolCards had
grown into near-identical copies of the same showPoolCard/isVisible
derivation. Extract that into CreditPoolCardsFromCycleData so the two
pages can't silently drift apart; only the SWR hooks that fetch the
cycle data still differ between them.


avervaet:

- #31868
    - #31869
        - #31870
            - #31834 ◀

PR #31835, opened 2026-09-03
Add tracking when users hit credit limits

theogz:
## Description

Sending PostHog events when users hit limits or have messages unable to send because of such limits.

## Tests

None

## Risk

Low, safe to rollback.


tdraier (approved):
lgtm

PR #31839, opened 2026-09-04
Poke: add a filter to only show Frames that have a sandbox

davidebbo:
## Description

Adds an "Only with a sandbox" checkbox to the Frames table on the Poke workspace page.

The list is server-paginated with "Load more", so a client-side filter would hide matching frames behind later pages. The filter is applied in the database instead: a new `hasSandbox` query param on the Poke frames endpoint adds an `IN (subquery)` clause on the `sandbox_owners` link table (same literal-subquery pattern as `run_resource.ts`, workspace id asserted numeric). Toggling the checkbox resets the page limit and refetches.

## Tests

Type-checked front and front-api, verified the generated SQL runs against the local database. Not verified in the browser.

## Risk

Low, Poke-only.

## Deploy Plan

Deploy front-api then front-spa.

### DIFF
--- a/front-api/routes/poke/workspaces/[wId]/frames/index.ts
+++ b/front-api/routes/poke/workspaces/[wId]/frames/index.ts
@@ -32,11 +32,13 @@ app.get("/", async (ctx): HandlerResult<PokeListFrames> => {
   }
 
   const { limit, lastValue, orderDirection } = paginationRes.value;
+  const { hasSandbox } = ctx.req.query();
 
   const frames = await listWorkspaceFrames(auth, {
     limit,
     lastValue,
     orderDirection,
+    hasSandbox: hasSandbox === "true",
   });
 
   return ctx.json(frames);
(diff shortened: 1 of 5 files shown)


front-api/routes/poke/workspaces/[wId]/frames/index.ts
```diff
@@ -32,11 +32,13 @@ app.get("/", async (ctx): HandlerResult<PokeListFrames> => {
   }
 
   const { limit, lastValue, orderDirection } = paginationRes.value;
+  const { hasSandbox } = ctx.req.query();
 
   const frames = await listWorkspaceFrames(auth, {
     limit,
     lastValue,
     orderDirection,
+    hasSandbox: hasSandbox === "true",
```
Comment (fontanierh) [inline front-api/routes/poke/workspaces/[wId]/frames/index.ts]: nit "has functions" would maybe be better because the sandbox is implementation details. Could also not have a sandbox yet but still be a backend-enabled frame

PR #31849, opened 2026-09-04
[front] enh: replace group consumption share with member usage metrics in attribution table

aubin-tchoi:
## Description

Context in [thread](https://dust4ai.slack.com/archives/C0BJ3T5SA0L/p1788508787255659).

In the attribution table of the new analytics page, the group tab currently shows the consumption share of the group.
Since groups are not MECE, the consumption share is hard to interpret as the user needs to constantly account for group overlap when looking at this one, unlike most other tabs where it's straightforward.

Groups are also hard to identify from here, so the user needs to have in mind precisely which group is which if not named clearly, and the information displayed cannot be normalized by group size so large groups will be advertised with no way to account for it.

This PR removes the consumption share column and adds the active user/total members and a comparison with the workspace's average.

image

## Tests

- Added tests.
- Tested locally + preview.

## Risk

- Low.

## Deploy Plan

- Deploy front.

### DIFF
--- a/front/components/poke/analytics/PokeConsumptionAttributionTable.tsx
+++ b/front/components/poke/analytics/PokeConsumptionAttributionTable.tsx
@@ -82,6 +82,7 @@ function PokeConsumptionAttributionRows(
   const {
     rows,
     totalCredits,
+    totalActiveMembers,
     totalCount,
     isTopLoading,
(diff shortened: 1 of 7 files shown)


avervaet (approved):
LGTM

PR #31851, opened 2026-09-04
Surface Frames v2 in conversations

davidebbo:
## Description

- Resolve a Frames v2 manifest path through conversation_side_panel.open_frame, so the agent does not parse CLI output for a file ID.
- Return the linked Frame as a generated-file resource and use its Frame name as the card title.
- Treat Frames v1 and v2 consistently for the Frame card, citations, file panel behavior, and automatic side-panel opening.
- Add conversation_side_panel to the Frames skill and instruct the agent to open the published manifest path once.

## Tests

- Focused suite passed: 6 files and 84 tests.
- npx tsgo --noEmit passed before the final branch packaging.
- The full front suite was stopped after it reported 2 failures in 1 file, per request.
- The pre-commit front type-check later failed with broad repository errors, including unresolved @dust-tt/client imports. The commit bypassed the hook per request.

## Risk

The existing file_id input remains supported. The new path input resolves only files linked into the current agent-loop file system. The main risk is a published Frame path that is not linked into that scope; open_frame then returns a file-not-found error.

Rollback by reverting this commit.

## Deploy Plan

No special deployment steps.

### DIFF
--- a/front/lib/api/actions/servers/conversation_side_panel/tools/index.ts
+++ b/front/lib/api/actions/servers/conversation_side_panel/tools/index.ts
@@ -9,9 +9,12 @@ import {
   SET_FILES_SIDE_PANEL_TOOL_NAME,
 } from "@app/lib/api/actions/servers/conversation_side_panel/metadata";
 import { buildInteractiveContentFileNotification } from "@app/lib/api/actions/servers/interactive_content/helpers";
+import { DustFileSystem } from "@app/lib/api/file_system";
+import { fetchLinkedFileResource } from "@app/lib/api/files/file_system_ops";
 import { FileResource } from "@app/lib/resources/file_resource";
-import { isInteractiveContentType } from "@app/types/files";
+import { getFileDisplayName, isFrameContentType } from "@app/types/files";
 import { Err, Ok } from "@app/types/shared/result";
+import { INTERNAL_MIME_TYPES } from "@dust-tt/client";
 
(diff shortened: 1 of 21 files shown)


front/lib/api/actions/servers/conversation_side_panel/tools/index.ts
```diff
@@ -51,17 +54,55 @@ const handlers: ToolHandlers<typeof CONVERSATION_SIDE_PANEL_TOOLS_METADATA> = {
...
+          tracked: false,
+        })
+      );
+    }
+
+    let fileResource: FileResource | undefined;
+    if (file_id) {
+      fileResource = (await FileResource.fetchById(auth, file_id)) ?? undefined;
+    } else if (path) {
+      const pathWithoutMountPrefix = path.startsWith("/files/")
```
davidebbo, line 69: This is where most of the logic is, to get the file id when we only have the path.

front/lib/resources/skill/code_defined/global/frames_v2.ts
```diff
@@ -351,6 +351,10 @@ stored, and activated atomically:
 dsbx frame publish /files/<scope>/<frame-folder>/manifest.json
 \`\`\`
 
+After a successful publish, call \`conversation_side_panel.open_frame\` exactly once with \`path\`
```
fontanierh, line 354: oh ok!

fontanierh (commented):
LGTM

PR #31857, opened 2026-09-04
LLMs: add Gemini 3.8 Flash

spolu:
## Description

Add Gemini 3.8 Flash to the user-facing model catalog.

- Register Google AI Studio and global Agent Platform stream endpoints.
- Keep the EU Agent Platform endpoint defined but unregistered: a live request through the native `@google/genai` transport returns `Invalid hostname: eu-aiplatform.googleapis.com`.
- Add model capabilities, introductory pricing, reasoning tiers, SDK typing, and auto-fast routing.
- Mark Gemini 3.8 Flash as latest while retaining Gemini 3.7 Flash as supported.

## Tests

- Google AI Studio live endpoint suite: 46/46 passed.
- Global Agent Platform live endpoint suite: 46/46 passed.
- EU Agent Platform live probe confirmed the endpoint is unavailable through the native transport.
- Model registry, SDK drift, pricing tier, user availability, and routing tests.

## Risk

Low. The change is additive. Both registered endpoints were validated live, while the failing EU endpoint remains unregistered.

## Deploy Plan

- deploy `front`


fontanierh (approved):

spolu:
(will thoroughly test locally before merging)

spolu:
Note 3.7 and 3.8 are aligned here:
```
 - Global Agent Platform: registered
  - Global Google AI Studio: registered
  - EU Agent Platform class: defined but unregistered
  - regionalAvailability["europe-west1"]: false
```

Based on tests I discovered this which requires more investigation:

```
EU is available. Direct REST results:

  - aiplatform.googleapis.com + locations/eu
      - Gemini 3.7 Flash: HTTP 200
      - Gemini 3.8 Flash: HTTP 200

  - eu-aiplatform.googleapis.com + locations/eu
      - Both fail with HTTP 400: Invalid hostname

  - europe-west1-aiplatform.googleapis.com
      - Both return HTTP 404

  - Default hostname + locations/europe-west1
      - Both return HTTP 200

  So the model supports EU; @google/genai is constructing the wrong hostname. We should not describe this as missing native API support. EU can likely be registered if we configure the SDK to retain aiplatform.googleapis.com while sending location: "eu".
```

PR #31858, opened 2026-09-04
governance seed: add a skill on the private space

fabiencelier:
## Description

Test data for #31830 (admins listing skills built on spaces they are not a member of).

- New governance seed skill "Space Restricted Procedures", owned by Alfred and requiring the Governance Private Space the current user is not a member of.
- The skill seeding lookup used the context user's view, so a re-run could not see skills built on spaces the user is not a member of and failed on the unique name. It now looks the row up directly and loads it with permission filtering skipped, like the agent seeding does.

## Tests

Seed test covers the new skill and its requested space.

## Risk

Low. Dev seed only.

## Deploy Plan

- Nothing to deploy


matteotrab (commented):
LGTM

PR #31859, opened 2026-09-04
Core GCS access in cell-00002

flvndvd:
## Description

Core can now access GCS from cell-00002 through Workload Identity, while existing regions continue to use their configured service-account credentials. All GCS operations share one client that keeps the current `cloud-storage` authentication whenever any of its four credential variables is present, and otherwise resolves Application Default Credentials through the GKE metadata server.

This keeps object operations and error behavior on the existing storage crate instead of combining the authentication rollout with a storage client migration. Concurrent ADC token requests share one refresh and use the same 20-second expiry margin as `gcp_auth`.

> [!NOTE]
> BigQuery authentication and removal of legacy service-account secrets are deliberately out of scope. This supersedes #31832.

## Tests

Added six focused tests covering authentication selection, blank legacy variables, token reuse, near-expiry refresh, and concurrent refreshes.

```
cargo check --manifest-path core/Cargo.toml
Finished `dev` profile [unoptimized + debuginfo] target(s)

cargo test --manifest-path core/Cargo.toml --lib gcs_client
test result: ok. 6 passed; 0 failed; 0 ignored; 0 measured; 96 filtered out
```

`cargo fmt --manifest-path core/Cargo.toml --all -- --check` and `cargo clippy --manifest-path core/Cargo.toml --lib` also pass. The full library run passed 96 tests with one ignored; five unrelated MCP tests could not run in the local macOS proxy and socket environment.

## Risk

The main risk is selecting the wrong authentication source. Existing regions take the unchanged `Client::new()` path whenever one of the four credentials supported by the current crate is configured, while cell-00002 uses ADC only when none is present. Rollback is safe because this changes no infrastructure and removes no secrets.

## Deploy Plan

Deploy to a legacy region first and verify GCS reads and writes still use its service account. Then deploy to cell-00002 without the four legacy credential variables, verify a GCS read and write through Workload Identity, and watch authentication errors. Roll back the Core deployment if either check fails; keep all legacy secrets in place.

### DIFF
--- a/core/src/gcs_client.rs
+++ b/core/src/gcs_client.rs
@@ -1,0 +1,263 @@
+use anyhow::{Context, Result};
+use async_trait::async_trait;
+use cloud_storage::{Client, TokenCache};
+use gcp_auth::TokenProvider;
+use std::sync::Arc;
+use tokio::sync::{Mutex, OnceCell, RwLock};
+
+const GCS_SCOPE: &str = "https://www.googleapis.com/auth/devstorage.full_control";
+// Keep this aligned with gcp_auth::Token::has_expired().
+const TOKEN_EXPIRY_SKEW_SECONDS: u64 = 20;
+// These are the four credential sources supported by cloud-storage::Client::new().
+// Keep their authentication behavior unchanged in existing deployments.
+const LEGACY_SERVICE_ACCOUNT_ENV_VARIABLES: [&str; 4] = [
+    "SERVICE_ACCOUNT",
+    "GOOGLE_APPLICATION_CREDENTIALS",
+    "SERVICE_ACCOUNT_JSON",
+    "GOOGLE_APPLICATION_CREDENTIALS_JSON",
+];
+
+static GCS_CLIENT: OnceCell<Client> = OnceCell::const_new();
+
+#[derive(Debug, PartialEq)]
+enum GcsAuthenticationMode {
+    LegacyServiceAccount,
+    ApplicationDefaultCredentials,
+}
+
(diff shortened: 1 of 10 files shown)


core/src/gcs_client.rs
```diff
@@ -0,0 +1,263 @@
...
+            .unwrap_or(false)
+    });
+
+    if has_legacy_credentials {
+        GcsAuthenticationMode::LegacyServiceAccount
+    } else {
+        GcsAuthenticationMode::ApplicationDefaultCredentials
+    }
+}
+
```
id13, line 70: OOC why aren't we simply doing ?

```suggestion
fn authentication_mode_from_environment(
    mut read_environment_variable: impl FnMut(&str) -> Option<String>,
) -> GcsAuthenticationMode {
    let has_legacy_credentials = LEGACY_SERVICE_ACCOUNT_ENV_VARIABLES.iter().any(|name| {
        std::env::var(name).ok()
            .map(|value| !value.trim().is_empty())
            .unwrap_or(false)
    });

    if has_legacy_credentials {
        GcsAuthenticationMode::LegacyServiceAccount
    } else {
        GcsAuthenticationMode::ApplicationDefaultCredentials
    }
}

```

id13 (approved):

core/src/gcs_client.rs
flvndvd: I went this way so we could test properly without mutating env.

flvndvd:
PR is live and confirmed no issues on legacy regions.

PR #31860, opened 2026-09-04
[front] Enable inline editing for Frames v2

davidebbo:
## Description

Enable inline text editing for Frames v2 in the conversation side panel.

- Gate editable iframe mode on the existing Frame v2 source-write permission.
- Route legacy and Frame v2 edits through the existing private file edit endpoint.
- Update canonical source under the Frame source lock, rebuild, and activate a new immutable publication.
- Roll source back if build, storage, reconciliation, or activation fails.
- Preserve legacy Frame inline-edit routing.

## Tests

- 51 focused tests covering Frame v2 editing, rollback on build/storage failure, permission gating, legacy editing, publication, and iframe behavior.
- Front and front-api TypeScript checks.
- Swagger annotation lint.
- Changed-file formatting and lint.

## Risk

Moderate and limited to inline Frame editing. The shared endpoint rechecks source write access server-side, rejects unsafe relative paths, serializes edits with the Frame source lock, and restores prior source bytes when publication fails. Read-only viewers remain non-editable.

## Deploy Plan

Standard deploy. No migration or backfill required.


front/components/assistant/conversation/interactive_content/FrameRenderer.tsx
```diff
@@ -154,50 +154,35 @@ export function FrameRenderer({
...
-      newText: string;
-      oldText: string;
-      targetFileId?: string;
-      source?: string;
-    }) => {
-      try {
-        // Location-based edits address the published entry Frame (the clicked bundle); the source
-        // path inside `source` resolves within its build root. Legacy context edits route to the
-        // nested target file.
-        const editFileId = source ? fileId : (targetFileId ?? fileId);
```
davidebbo, line 173: Logic moved to swr

fontanierh (approved):

PR #31863, opened 2026-09-04
skills: permission filtering modes on the skill fetchers, with a redacted mode for admins

fabiencelier:
## Description

Groundwork for listing to admins the skills built on spaces they are not a member of (follow-up PR on top of this one: https://github.com/dust-tt/dust/pull/31830).

- The skill fetchers (`fetchById`, `fetchByIds`, `fetchByModelIds`, `listByWorkspace`, `listByAgentConfiguration`, ...) take a `permissionFiltering` mode instead of the `dangerouslySkipPermissionFiltering` flag:
  - `strict` (default): drop the skills the caller cannot read, as today.
  - `redact_unreadable`: keep them, redacted. The resource flags them, its permission methods answer false and `toJSON` drops instructions, tools and files. Admins only, the resource throws otherwise.
  - `dangerously_skip`: keep them as is, the former flag.
- `baseFetch` centralizes the readability check in `filterReadable` (row ACL plus requested spaces) and applies the mode there, so every fetcher behaves the same.
- New `canRead` field on the skill type and the public Swagger schema, false only for redacted skills.
- Existing callers of the flag switch to `dangerously_skip`. Nothing uses `redact_unreadable` yet.

## Tests

 - new resource tests

## Risk

Low

## Deploy Plan

- Deploy front

### DIFF
--- a/front/lib/resources/skill/skill_resource.ts
+++ b/front/lib/resources/skill/skill_resource.ts
@@ -654,17 +654,11 @@ export class SkillResource extends BaseResource<SkillConfigurationModel> {
    * referencing one are dropped too. This is what the fetch path applies (see
    * `SkillPermissionFilteringMode`).
    */
-  static async filterReadable<
-    T extends {
-      id: ModelId;
-      workspaceId: ModelId;
-      requestedSpaceIds: ModelId[];
-    },
-  >(
+  private static async filterReadable(
     auth: Authenticator,
-    skills: T[],
+    skills: SkillConfigurationModel[],
     { transaction }: { transaction?: Transaction } = {}
-  ): Promise<T[]> {
+  ): Promise<SkillConfigurationModel[]> {
     const uniqueRequestedSpaceIds = uniq(
       skills.flatMap((skill) => skill.requestedSpaceIds)
     );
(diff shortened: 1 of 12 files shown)


front/lib/resources/skill/skill_resource.ts
```diff
@@ -632,21 +647,57 @@ export class SkillResource extends BaseResource<SkillConfigurationModel> {
...
+    T extends {
+      id: ModelId;
+      workspaceId: ModelId;
+      requestedSpaceIds: ModelId[];
+    },
+  >(
+    auth: Authenticator,
+    skills: T[],
+    { transaction }: { transaction?: Transaction } = {}
+  ): Promise<T[]> {
```
aubin-tchoi: I think that `T` is always `SkillConfigurationModel` here

front/lib/resources/skill/skill_resource.ts
```diff
@@ -2253,7 +2330,7 @@ export class SkillResource extends BaseResource<SkillConfigurationModel> {
   // skill the caller cannot read is never hydrated.
   private static canReadRow(
     auth: Authenticator,
-    skill: SkillConfigurationModel
+    skill: { id: ModelId; workspaceId: ModelId }
```
aubin-tchoi: let's keep the more restrictive type here

front/lib/resources/skill/skill_resource.ts
```diff
@@ -632,21 +647,57 @@ export class SkillResource extends BaseResource<SkillConfigurationModel> {
...
+    const uniqueRequestedSpaceIds = uniq(
+      skills.flatMap((skill) => skill.requestedSpaceIds)
+    );
+    const spaces =
+      uniqueRequestedSpaceIds.length > 0
+        ? await SpaceResource.fetchByModelIds(auth, uniqueRequestedSpaceIds, {
+            transaction,
+          })
+        : [];
+    const spaceById = new Map(spaces.map((s) => [s.id, s]));
```
aubin-tchoi: nit
```suggestion
    const spaceByModelId = new Map(spaces.map((s) => [s.id, s]));
```

aubin-tchoi (approved):
LGTM

PR #31864, opened 2026-09-04
[front] cleanup: remove unused hooks and endpoints

aubin-tchoi:
## Description

The workspace datasource retrieval treemap was removed, but its SWR hook and private Poke API path were left behind. This removes that unused workspace-wide path end to end, including route registration, response types, and the Elasticsearch aggregation.

Also removes `/api/w/:wId/assistant/agent_configurations/:aId/mcp_configurations` as it's not used either.

## Tests

- typecheck

## Risk

- Low.

## Deploy Plan

- Deploy front.


adrsimon (approved):
🧹

PR #31868, opened 2026-09-04
Add customer-facing AWU pool current-cycle / cycle-history app routes

avervaet:
## Description

Registers the awu-pool-current-cycle and awu-pool-cycle-history endpoints under the customer-facing /api/w/:wId/credits app routes, alongside SWR hooks to consume them from the front.

This is preliminary work toward exposing the new usage page already available in poke to front.

## Risk

- Low not consumed anywhere yet mirror existing behavior powering poke

## Deploy Plan

- Deploy front

### DIFF
--- a/front-api/routes/w/[wId]/credits/awu-pool-cycle-history.ts
+++ b/front-api/routes/w/[wId]/credits/awu-pool-cycle-history.ts
@@ -1,0 +1,67 @@
+import type { AwuPoolSummaryError } from "@app/lib/api/credits/awu_pool_summary";
+import {
+  AwuPoolSummaryQuerySchema,
+  getAwuPoolCycleHistory,
+} from "@app/lib/api/credits/awu_pool_summary";
+import type { AwuPoolCycleHistoryResponseBody } from "@app/types/api/credits/awu_pool_summary";
+import { assertNever } from "@app/types/shared/utils/assert_never";
+import { workspaceApp } from "@front-api/middlewares/ctx";
+import { ensureIsManager } from "@front-api/middlewares/ensure_role";
+import type { HandlerResult } from "@front-api/middlewares/utils";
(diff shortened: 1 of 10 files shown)


avervaet:

- #31868 ◀
    - #31869
        - #31870
            - #31834

front-api/routes/w/[wId]/credits/awu-pool-cycle-history.ts
```diff
@@ -0,0 +1,67 @@
...
+import type { AwuPoolCycleHistoryResponseBody } from "@app/types/api/credits/awu_pool_summary";
+import { assertNever } from "@app/types/shared/utils/assert_never";
+import { workspaceApp } from "@front-api/middlewares/ctx";
+import { ensureIsManager } from "@front-api/middlewares/ensure_role";
+import type { HandlerResult } from "@front-api/middlewares/utils";
+import { apiError } from "@front-api/middlewares/utils";
+import { validate } from "@front-api/middlewares/validator";
+import type { Context } from "hono";
+
+function cycleHistoryErrorToApi(ctx: Context, err: AwuPoolSummaryError) {
```
Comment (PopDaph) [inline front-api/routes/w/[wId]/credits/awu-pool-cycle-history.ts]: This is the same for both routes? Could it be extracted and shared? (maybe also duplicated elsewhere)?

PopDaph (approved):
Quick question: `awu-pool-summary` already returns both halves in one call. Is the split here so the cards can render before the slower history comes back?

Approving already since I trust your call 🫡

avervaet:
> Quick question: `awu-pool-summary` already returns both halves in one call. Is the split here so the cards can render before the slower history comes back?
> 
> Approving already since I trust your call 🫡

Yes the split is here so card related to the current cycle consumption, probably the first things people want to check loads before we have all the data required to display the previous cycle table.

PR #31869, opened 2026-09-04
Expose a shared member spend-limit modal for the front usage page

avervaet:
## Description

Make the MemberSpendLimitModal shareable by Poke and the app. 

This prepares the modal to be wired into the customer- facing usage page in a follow-up.

## Risk

Low

## Deploy Plan

- Deploy front


avervaet:

- #31868
    - #31869 ◀
        - #31870
            - #31834
            - #31982

front/components/workspace/MemberSpendLimitModal.tsx
```diff
@@ -0,0 +1,518 @@
...
+  onClose,
+  onSavingChange,
+  onSaved,
+}: MemberSpendLimitFormProps) {
+  const sendNotification = useSendNotification();
+  const notifySaveUnavailable = () =>
+    sendNotification({
+      title: "Not available from Poke yet",
+      description: "Editing spend limits from Poke isn't supported yet.",
+      type: "info",
```
PopDaph: Could the poke-specific copy stay out of this shared component? Since it now lives under `components/workspace/`, maybe `readOnly` could just disable Validate, and `PoolUsagePage` pass its own handler if poke wants the "not available yet" toast?

front/components/workspace/MemberSpendLimitModal.tsx
```diff
@@ -0,0 +1,518 @@
+import { useSendNotification } from "@app/hooks/useNotification";
```
PopDaph: Could we split into multiple files?

PopDaph (commented):
I see there's a EditSpendLimitModal already existing, do we need both? what's the difference?

avervaet:
> I see there's a EditSpendLimitModal already existing, do we need both? what's the difference?

We have on but he is ugly and does not support group edit at the moment here are the visuals: 
**now:**
Capture d’écran 2026-09-04 à 16 44 51

**existing one:**
Capture d’écran 2026-09-04 à 16 45 23

PopDaph:
But do we need both modals? Since I don't have the full context it's unclear to me what is the end goal 🙏🏻

avervaet:
> But do we need both modals? Since I don't have the full context it's unclear to me what is the end goal 🙏🏻

We will remove those later on when the paged is fully rolled-in so we don't display the new modal before the usage page is released.

front/components/workspace/member_spend_limit_helpers.ts
```diff
@@ -0,0 +1,71 @@
+import type { MemberUsageType } from "@app/lib/api/credits/members_usage";
+import type { GroupType } from "@app/types/groups";
+import { removeNulls } from "@app/types/shared/utils/general";
+
+export const MIN_AWU_CREDITS = 0;
```
PopDaph: Looks like we already have both in lib/api/users/spend_limit.ts, ideally we centralize that somewhere that's importable both server and client side 🙏🏻 

export const MIN_USER_SPEND_LIMIT_AWU_CREDITS = 0;
export const MAX_USER_SPEND_LIMIT_AWU_CREDITS = 2_000_000;

front/components/workspace/EditMemberSpendLimitModal.tsx
```diff
@@ -0,0 +1,307 @@
...
+          disabled: isSaving || readOnly,
+          isLoading: isSaving,
+          onClick: handleValidate,
+        }}
+      />
+    </>
+  );
+}
+
+export function EditMemberSpendLimitModal({
```
PopDaph, line 258: We now have EditSpendLimitModal, BulkEditSpendLimitModal and EditMemberSpendLimitModal side by side, two of them with their own parser. I get the rollout reason, but can you add some #TODO so we know what is deprecated when? 🙏🏻

front/components/workspace/EditMemberSpendLimitModal.tsx
```diff
@@ -0,0 +1,307 @@
...
+  onSaved?: () => void;
+}
+
+interface MemberSpendLimitFormProps {
+  member: MemberUsageType | null;
+  owner: LightWorkspaceType;
+  groups: GroupType[];
+  readOnly: boolean;
+  onClose: () => void;
+  onSavingChange?: (memberId: string, isSaving: boolean) => void;
```
PopDaph: Do onSavingChange & onSaved have callers?

front/components/workspace/MemberGroupLimitTable.tsx
```diff
@@ -0,0 +1,90 @@
...
+}
+
+export function MemberGroupLimitTable({
+  rows,
+  readOnly,
+  groupLimitInputs,
+  groupValidationMessages,
+  onChange,
+}: MemberGroupLimitTableProps) {
+  const groupColumns: ColumnDef<GroupRow, string>[] = useMemo(
```
PopDaph: Are you sure this useMemo is doing anything? 

iiuc groupLimitInputs is a new object each time a value changes, and handleGroupLimitChange is a new function on every render of the form. So the deps change on every render and the columns get rebuilt anyway.

front/components/workspace/member_spend_limit_helpers.ts
```diff
@@ -0,0 +1,71 @@
...
+  }, null);
+
+  return memberGroups.map((g) => ({
+    groupId: g.sId,
+    name: g.name,
+    poolCapAwuCredits: g.poolCapAwuCredits,
+    memberCount: g.memberCount,
+    isHighest: !hasPersonalOverride && g.sId === highestGroupId,
+  }));
+}
```
PopDaph, line 74:
```suggestion
export function groupRowsForMember(
  member: MemberUsageType | null,
  groups: GroupType[]
): GroupRow[] {
  const hasPersonalOverride = member?.spendLimitSource === "override";

  const groupsByName = new Map(groups.map((g) => [g.name, g]));
  const memberGroups = removeNulls(
    (member?.groups ?? []).map((name) => groupsByName.get(name))
  );

  // When there's no personal override, the highest group cap the member is
  // part of is the one currently granting their extra credits.
  let highest: GroupType | null = null;
  for (const g of memberGroups) {
    if (
      g.poolCapAwuCredits !== null &&
      g.poolCapAwuCredits > (highest?.poolCapAwuCredits ?? -1)
    ) {
      highest = g;
    }
  }

  return memberGroups.map((g) => ({
    groupId: g.sId,
    name: g.name,
    poolCapAwuCredits: g.poolCapAwuCredits,
    memberCount: g.memberCount,
    isHighest: !hasPersonalOverride && g.sId === highest?.sId,
  }));
}
```

PopDaph: I think the reduce with a find inside is hard to follow. A Map by name and a plain loop for the highest cap reads better (and drops the nested lookups, though sizes are small so that part does not matter much?). Feel free to ignore

PopDaph (approved):
Approving already. 
I did not test locally I only reviewed the code!

PR #31871, opened 2026-09-04
webhook-router: forward to cell-00002

flvndvd:
## Description

The webhook router forwards to cell-00002. The router is cell-based since #31727 but only knows the two legacy cells. cell-00002 joins the cell list with a third forwarding target: `https://cell-00002.cells.dust.tt/connectors`, where the cell gateway serves the connectors API under a `/connectors` prefix (dust-infra #1049), and the cell's own `connectors-DUST_CONNECTORS_WEBHOOKS_SECRET`, read from project `prj-dust-cell-00002` through a new `GCP_CELL_00002_PROJECT_ID` parameter. The router service account already has read access on that secret from the dust-infra global root.

Once connectors on cell-00002 registers a Slack or Notion workspace, its config entry lists `cell-00002` and the router reaches it. Microsoft forwards to every cell as before, now three.

Pairs with #31862, which lets connectors itself accept `cell-00002`.

## Tests

`tsc` passes with the same pre-existing zod resolution errors as main, nothing new. Same shape as the two existing targets.

## Risk

The router refuses to start if `GCP_CELL_00002_PROJECT_ID` is missing at deploy time, the same guard as the other project ids, and `deploy.sh` checks it up front. A wrong secret only fails forwarding to cell-00002, the legacy cells are untouched.

## Deploy Plan

Merge, then deploy the function with `GCP_CELL_00002_PROJECT_ID=prj-dust-cell-00002` exported next to the existing project ids.

### DIFF
--- a/firebase-functions/webhook-router/README.md
+++ b/firebase-functions/webhook-router/README.md
@@ -37,6 +37,7 @@ verification.
    - `GCP_GLOBAL_PROJECT_ID`
    - `GCP_US_PROJECT_ID`
    - `GCP_EU_PROJECT_ID`
+   - `GCP_CELL_00002_PROJECT_ID`
    - `GCP_WEBHOOK_ROUTER_CONFIG_BUCKET`
 
 ### Project Configuration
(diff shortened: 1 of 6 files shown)


Fraggle (approved):
LGTM !

PR #31867, opened 2026-09-04
Poke: introduce the fair-use credits column for premium message plans

avervaet:
## Description

Premium message plan members also carry a fixed AWU credit allowance (separate from the rolling premium-message limit) for usage on non-premium models, fetched the same way as the personal analytics cards fair-use credits block. Surfaces it as its own column, with a progress bar and per-row reset countdown on hover. Sorting comes in the next PR in this stack. Stacked on #31866.

## Tests

Tested on front-edge

## Risk

- Poke only for now

## Deploy Plan

- Deploy front

PR #31866, opened 2026-09-04
Poke: remove seat columns for premium message plans

avervaet:
## Description

Premium message plans have no seats: every member is billed per message, so the seat and seat-usage columns had nothing to show, and the row menu's "Assign seat" action didn't apply either. Stacked on #31853.

## Tests

Tested on front-edge

## Risk

- Poke only for now

## Deploy Plan

- Deploy front

PR #31853, opened 2026-09-04
Poke: drop seat columns and add reset caption for premium message plans

avervaet:
## Description

Replace the Premium messages tooltip by a "Resets on a rolling 7-day basis" sub column header

## Tests

Tested on front-edge

## Risk

- Poke only for now

## Deploy Plan

- Deploy front

PR #31872, opened 2026-09-04
Poke: sort by the fair-use credits column, and default to it

avervaet:
## Description

Adds server-side sorting on the fair-use credits column, batched into a single pipelined Redis round-trip so a full-table sort does not cost one round-trip per member — same pattern already used for the premium-messages sort. 

Premium message workspaces now default-sort by this column instead of premium message usage. 

Stacked on #31867.

## Tests

Tested on front-edge

## Risk

- Poke only for now

## Deploy Plan

- Deploy front

### DIFF
--- a/front/lib/api/assistant/rate_limits.ts
+++ b/front/lib/api/assistant/rate_limits.ts
@@ -81,8 +81,8 @@ export const makeAgentMentionsRateLimitKeyForWorkspace = (
 };
 
 export const makeFairUseAwuCreditsRateLimitKeyForUser = (
-  owner: LightWorkspaceType,
-  user: UserType,
+  owner: Pick<LightWorkspaceType, "id">,
+  user: Pick<UserType, "id">,
   maxAwuCreditsTimeframe: MaxAwuCreditsTimeframeType
 ) => {
   // `:v2_microcredits` marks the switch to weighted amount-carrying entries
--- a/front/lib/api/assistant/rate_limits.ts
+++ b/front/lib/api/assistant/rate_limits.ts
@@ -81,8 +81,8 @@ export const makeAgentMentionsRateLimitKeyForWorkspace = (
 };
 
 export const makeFairUseAwuCreditsRateLimitKeyForUser = (
-  owner: Pick<LightWorkspaceType, "id">,
-  user: Pick<UserType, "id">,
+  owner: LightWorkspaceType,
+  user: UserType,
   maxAwuCreditsTimeframe: MaxAwuCreditsTimeframeType
 ) => {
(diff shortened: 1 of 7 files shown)


avervaet:

- #31853
    - #31866
        - #31867
            - #31872 ◀

avervaet:
Addressed the review comment: `getWeightedRateLimiterUsageForKeys` now validates the eval reply with a zod schema (`WeightedRateLimiterUsageForKeysReplySchema`) instead of an unsafe `as` cast.

front/lib/metronome/user_block.ts
```diff
@@ -303,6 +304,45 @@ export async function getFairUseAwuCreditsStatus({
   };
 }
 
+// Bulk variant of `getFairUseAwuCreditsStatus`
+export async function getFairUseAwuCreditsUsedCountsByUser({
+  workspace,
+  users,
+  plan,
+}: {
+  workspace: Pick<LightWorkspaceType, "id">;
```
aubin-tchoi: why not the entire workspace? this type is extremely permissive as is (any object that has a numeric `id` field)

front/lib/metronome/user_block.ts
```diff
@@ -303,6 +304,45 @@ export async function getFairUseAwuCreditsStatus({
...
+  users: Pick<UserType, "id" | "sId">[];
```
aubin-tchoi: same

front/lib/utils/rate_limiter.ts
```diff
@@ -400,6 +402,96 @@ export async function getWeightedRateLimiterUsage({
...
+      const parsedReplies =
+        WeightedRateLimiterUsageForKeysReplySchema.safeParse(rawReplies);
+      if (!parsedReplies.success) {
+        return new Err(
+          new Error(
+            `Unexpected reply shape from getWeightedRateLimiterUsageForKeys: ${fromError(parsedReplies.error).toString()}`
+          )
+        );
+      }
+      const replies = parsedReplies.data;
```
aubin-tchoi, line 524: we could do a length check here to make sure we have the expected number of replies

front/lib/api/assistant/rate_limits.ts
```diff
@@ -81,8 +81,8 @@ export const makeAgentMentionsRateLimitKeyForWorkspace = (
 };
 
 export const makeFairUseAwuCreditsRateLimitKeyForUser = (
-  owner: LightWorkspaceType,
-  user: UserType,
+  owner: Pick<LightWorkspaceType, "id">,
```
aubin-tchoi: we don't need to widen the type here

aubin-tchoi (approved):
LGTM once comments are addressed

PR #31873, opened 2026-09-04
[sparkle] Markdown: add step variant for task lists

PopDaph:
Adds a `taskListVariant` prop to `Markdown` (`"checkbox"` by default, `"step"`).

In step mode, GFM task items render read-only circles instead of checkboxes: numbered inside ordered lists, empty in bulleted ones, highlight check when done. Intended for the plan mode panel, where the agent ticks progress and the user cannot.

New `TaskLists` story shows both variants.

Small thing about the animation: when a task gets ticked the circle changes color and the check pops in. Since the pop-in fires when the icon mounts, already done tasks also pop once when the plan first shows up. I'm fine with it, it looks like a reveal. Only animating real changes would mean knowing what the plan looked like before, and Markdown doesn't, only the caller does.

<kbd>
Screenshot 2026-09-04 at 14 54 23
</kbd>

### DIFF
--- a/sparkle/src/components/markdown/InputBlock.tsx
+++ b/sparkle/src/components/markdown/InputBlock.tsx
@@ -1,6 +1,7 @@
 import { Checkbox } from "@sparkle/components/Checkbox";
 import { useMarkdownStyle } from "@sparkle/components/markdown/MarkdownStyleContext";
 import { sameNodePosition } from "@sparkle/components/markdown/utils";
+import { assertNever } from "@sparkle/lib/utils";
 import React, { memo } from "react";
 import type { ReactMarkdownProps } from "react-markdown/lib/ast-to-react";
 
@@ -11,7 +12,7 @@ type InputProps = Omit<React.InputHTMLAttributes<HTMLInputElement>, "ref"> &
 
 /**
  * Renders `<input>` elements inside Markdown output; checkboxes (as produced
- * by GFM task lists) are swapped for the Sparkle Checkbox component, or
+ * by GitHub Flavored Markdown task lists) are swapped for the Sparkle Checkbox component, or
  * dropped when the "step" task-list variant draws its own badge; other input
  * types pass through unchanged.
  * @summary Input renderer for Markdown task-list checkboxes.
@@ -34,8 +35,13 @@ export const InputBlock = memo(
       );
     }
 
-    if (taskListVariant === "step") {
-      return null;
+    switch (taskListVariant) {
+      case "step":
+        return null;
+      case "checkbox":
+        break;
+      default:
+        assertNever(taskListVariant);
     }
 
     const handleCheckedChange = (isChecked: boolean) => {
--- a/sparkle/src/components/markdown/InputBlock.tsx
+++ b/sparkle/src/components/markdown/InputBlock.tsx
@@ -1,4 +1,5 @@
 import { Checkbox } from "@sparkle/components/Checkbox";
+import { useMarkdownStyle } from "@sparkle/components/markdown/MarkdownStyleContext";
 import { sameNodePosition } from "@sparkle/components/markdown/utils";
 import React, { memo } from "react";
 import type { ReactMarkdownProps } from "react-markdown/lib/ast-to-react";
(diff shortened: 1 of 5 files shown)


sparkle/src/components/markdown/InputBlock.tsx
```diff
@@ -10,14 +11,16 @@ type InputProps = Omit<React.InputHTMLAttributes<HTMLInputElement>, "ref"> &
 
 /**
  * Renders `<input>` elements inside Markdown output; checkboxes (as produced
- * by GFM task lists) are swapped for the Sparkle Checkbox component, other
- * input types pass through unchanged.
+ * by GFM task lists) are swapped for the Sparkle Checkbox component, or
```
ykmsd: what is GFM btw? 👶

sparkle/src/components/markdown/List.tsx
```diff
@@ -1,12 +1,17 @@
...
   sameNodePosition,
 } from "@sparkle/components/markdown/utils";
+import { Check } from "@sparkle/icons/v2-stroke";
 import { cn } from "@sparkle/lib";
 import { cva } from "class-variance-authority";
-import React, { memo } from "react";
+import React, { createContext, memo, useContext } from "react";
+
+// Ordered lists share their `start` so step badges can number task items.
+const OlStartContext = createContext(1);
```
ykmsd, line 15: do we plan to use something that doesn't start with 1?

sparkle/src/components/markdown/List.tsx
```diff
@@ -54,54 +59,162 @@ export const UlBlock = memo(
...
+ * from the source Markdown; GFM task lists (detected via the `contains-task-list`
+ * class) drop the decimal markers and indent, like UlBlock does for bullets.
  * @summary Ordered-list renderer for Markdown.
  */
 export const OlBlock = memo(
-  ({ children, start }: OlBlockProps) => {
+  ({ children, className, start }: OlBlockProps) => {
     const { textColor, forcedTextSize } = useMarkdownStyle();
     const textSize = forcedTextSize ?? markdownParagraphSize;
+    const isTaskList = className?.includes("contains-task-list");
```
ykmsd: it's not really a good practice to detect based on classnames (easy to break) but is this classname coming from the external library? (I couldn't find it in the codebase) ideally we want to control with js but I guess we cannot do it here?

sparkle/src/components/markdown/List.tsx
```diff
@@ -54,54 +59,162 @@ export const UlBlock = memo(
...
+    prev.start === next.start &&
+    prev.className === next.className
 );
 OlBlock.displayName = "OlBlock";
 
+// mt-[3px] centers the 20px badge on the 26px leading-relaxed text line. The border stays on
+// both states so ticking a task only transitions colors.
+const taskStepBadgeVariants = cva(
+  cn(
+    "mt-[3px] flex size-5 shrink-0 items-center justify-center rounded-full border",
```
ykmsd, line 124: we should avoid using custom pixel like this if possible `mt-[3px]` and I think we can align them with align-items: center by adding it to the li of LiBlock?

ykmsd: okay we cannot use it when the text is multi line 🙃 let's leave as it is

ykmsd (approved):

sparkle/src/components/markdown/InputBlock.tsx
PopDaph: Great question, and totally fair to ask! GFM stands for GitHub Flavored Markdown. It's the Markdown dialect GitHub uses, and it adds things plain Markdown doesn't have: task lists (- [ ] / - [x]), tables, strikethrough, autolinks and footnotes.

In Sparkle the Markdown component uses react-markdown with the remark-gfm plugin, which is the library that implements that spec. It's also where the contains-task-list and task-list-item classnames come from, so they're not something we made up in this codebase.

I spelled out "GitHub Flavored Markdown" in the JSDoc so the next person doesn't have to wonder. Thanks for flagging it! 🙏🏻

sparkle/src/components/markdown/List.tsx
PopDaph: Yes, those classes come from remark-gfm (the library behind our Markdown rendering), not from us. For list items react-markdown also passes a boolean checked prop, so I switched item detection to that, which is more robust. For the ul/ol elements the class is the only signal the library gives, so that one stays, with a comment pointing to the source. Good catch!

sparkle/src/components/markdown/List.tsx
PopDaph: We don't plan to, but we don't fully control it either: the plan is Markdown written by the agent, so technically it could write a list that starts at 3, or split tasks across two lists after a paragraph, where Markdown restarts the numbering. 

react-markdown already hands us the start value and OlBlock was already forwarding it to the element, so the badge just reads the same value to stay in sync with what a regular numbered list would show. 

It's a few lines, and it avoids a case where the badge says 1 and the Markdown says 3. Happy to drop it if you'd rather keep things minimal!

PR #31874, opened 2026-09-04
[front] plan mode: render tasks as numbered steps

PopDaph:
## Description 

**Goal: improve plan mode delight to ship it 💅🏻** 

Follow-up to #31873, which added the `taskListVariant` prop to Sparkle's `Markdown`.

The plan panel now uses it, so tasks show up as circles instead of checkboxes nobody can click. I also removed the `key` that remounted the whole Markdown on every edit and passed `optimizeForStreaming={false}` instead. Items re-render in place, which is what lets the circle animate when the agent ticks a task.

The plan template and the skill instructions ask for numbered tasks (`1. [ ]`) so the circles carry the task number. The done/total count on the plan card moved to `utils.ts` and accepts both `1. [ ]` and `- [ ]`, so plans created before this keep counting.

While testing I hit the `[!]` marker we were teaching the agent for blocked tasks. It's not GFM, so it renders as plain text and the item loses its number. The skill now says a blocked task stays `[ ]` with the reason in its text. Old `[!]` items still count as not done.

<kbd>
Screenshot 2026-09-04 at 15 57 14
</kbd>

## Test

See it here: https://pr-31874-merge--app.preview.dust.tt/w/0ec9852c2f/conversation/HE3JRv9LzO#?spid=plan&spt=plan

## Deploy 

Deploy front.


front/components/assistant/conversation/plan_mode/PlanCard.tsx
```diff
@@ -23,24 +24,6 @@ interface PlanCardProps {
   workspaceId: string;
 }
 
-// Total counts every task marker (open, done, blocked); done counts only checked boxes.
```
PopDaph, line 26: just moved in front/components/assistant/conversation/plan_mode/utils.tsx

aubin-tchoi (approved):
LGTM

PR #31875, opened 2026-09-04
core: backfill prompts name the cell and region from the pod topology

flvndvd:
## Description

The Elasticsearch backfill binaries in core asked for `DUST_REGION` only to print it in their confirmation prompt. Every core pod already carries `CELL` and `REGION` from its topology labels, so the prompt now reads them and names the deployment precisely: "backfill the index X in cell-00002 (europe-west1)?". The generic backfill keeps its "local" shortcut in development.

Two of the four binaries are commented out of the build. They get the same change so they do not rot. Once this is deployed, `core-DUST_REGION` can leave the legacy bundles, nothing else reads it.

## Tests

No behaviour change.

## Risk

None at runtime, core-api and the sqlite worker never read these variables. An operator running a backfill from a pod without `CELL` set gets a clear panic message instead of a prompt.

## Deploy Plan

Merge, ships with the next core deploy.

### DIFF
--- a/core/bin/migrations/20250226_backfill_elasticsearch_text_size.rs
+++ b/core/bin/migrations/20250226_backfill_elasticsearch_text_size.rs
@@ -33,6 +33,14 @@ struct Args {
  * cargo run --bin backfill_elasticsearch_text_size -- --index-version <version> [--skip-confirmation] [--start-cursor <cursor>] [--batch-size <batch_size>]
  *
  */
+// The pods carry CELL and REGION from their topology labels, so the prompt names the
+// deployment the operator is about to touch.
+fn deployment_target() -> String {
+    let cell = std::env::var("CELL").expect("CELL must be set");
+    let region = std::env::var("REGION").expect("REGION must be set");
+    format!("{} ({})", cell, region)
+}
+
 #[tokio::main]
 async fn main() {
     if let Err(e) = run().await {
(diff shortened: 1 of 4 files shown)


core/bin/migrations/20250226_backfill_elasticsearch_text_size.rs
```diff
@@ -33,6 +33,14 @@ struct Args {
  * cargo run --bin backfill_elasticsearch_text_size -- --index-version <version> [--skip-confirmation] [--start-cursor <cursor>] [--batch-size <batch_size>]
  *
  */
+// The pods carry CELL and REGION from their topology labels, so the prompt names the
+// deployment the operator is about to touch.
+fn deployment_target() -> String {
+    let cell = std::env::var("CELL").expect("CELL must be set");
+    let region = std::env::var("REGION").expect("REGION must be set");
+    format!("{} ({})", cell, region)
+}
```
adrsimon, line 42: isn't it defined two times ?

adrsimon: ah no ok it's a migration, we could have imported the one from above but ok, prob better to not make it public i guess

adrsimon (approved):

core/bin/migrations/20250226_backfill_elasticsearch_text_size.rs
flvndvd: Yep exactly.

PR #31697, opened 2026-09-02
Add free-seat lifetime rate-limiter

tdraier:
## Description

Follow-up to #31625. Free seats have a **lifetime** credit balance — an admin-editable, contract-surviving per-user Metronome credit (`startingBalanceAwu`, toppable via the `grant-user-free-credits` poke plugin) — not a per-cycle cap, so #31625 left them on the Metronome balance-threshold alert. This adds a synchronous **lifetime rate-limiter** for free seats behind `enforce_user_spend_limit_rate_cap`: under the flag, free-seat blocking and near-limit are driven **entirely by the rate limiter** (no Metronome credit-state fallback), matching how pool seats moved off Metronome.

### Enforcement
- New per-user **lifetime** fixed-window counter (`makeFreeSeatLifetimeAwuCreditsRateLimitKeyForUser` + never-rolling `makeSpendLimitLifetimeWindowBounds`), seeded from all-time `is_free_seat` ES consumption, recorded per message in `credit_cost.ts` for free-seat users.
- Threshold = the per-user free-seat allowance from the cached Metronome credit (`startingBalanceAwu`); admin top-ups just raise it (no new store, poke top-up flow unchanged).
- `isUserSpendLimitRate{Cap,Warning}Reached` delegate free seats to `isFreeSeatLifetime{Cap,Warning}Reached`. Under the flag these always return a boolean — **no Metronome credit-state fallback**; if the allowance can't be resolved they fail open (`false`) and log an error.
- The Metronome free-seat balance alerts keep running as the **flag-off** enforcement source, and still drive the free→pro auto-upgrade (`dispatchSeatBalanceExhausted`, not flag-gated).

### Poke usage tab (under the flag)
- **RL** column shows the lifetime counter for free seats (per-cycle counter otherwise).
- **near-limit** mirrors enforcement: lifetime counter vs the free-seat allowance for free seats, per-cycle cap otherwise.
- The Metronome **alert badges** (per-user cap / 80% warning / free-seat low+empty) are dropped, since they no longer drive enforcement.

## Tests

- tsgo clean on the changed backend files (two pre-existing sparkle `ProgressRing`/icon-size errors in unrelated component files are environmental).
- Passing: credit_check, members_usage, metronome/user_block, programmatic_usage_limit, keys/spend_limit, groups/spend_limit, spend_limits/effective, default_user_spend_limit, alerts/spend_limits.

## Risk

[...]

PR #31876, opened 2026-09-04
API keys: flag-aware capped verdict, never read creditState under the rate-cap flag

tdraier:
## Description

Stacked on #31697. Applies the same treatment as the members table to API keys, and — since API keys have no seat/pool distinction — makes the flag a clean cutover: under `enforce_user_spend_limit_rate_cap` the persisted Metronome `creditState` is **never read** for API keys.

**Customer path (flag-aware "capped" status):**
- New `KeyType.isSpendCapped`, computed in `KeyResource.toJSONWithSpaces`. Flag on → per-key Redis fixed-window counter vs the key's `monthlyCapAwuCredits` (microcredit comparison, exact parity with `isApiKeySpendLimitRateCapReached`); flag off → `creditState === "capped"`. Fails open when the billing cycle can't be resolved or Redis errors. Mirrors the enforcement switch in `isApiKeyBlocked`.
- The keys table status (`getKeyStatus`) and the "at the monthly cap" count now key off `isSpendCapped` instead of `key.creditState === "capped"`.

**Poke path (new column):**
- New `ApiKeyUsageType.rateLimiterState` (`ok`/`near limit`/`capped`), from the already-fetched counter vs the key cap — flag-independent, surfaced beside the Metronome "Credit state" column to spot divergence.
- New "Rate limiter state" column in `PokeApiKeysUsageTable`, mirroring `PokeMembersUsageTable`.

## Tests

- `tsgo --noEmit` clean in `front` and `front-api`.
- biome clean.
- `key_resource.test.ts` (25) and `front-api` `keys/index.test.ts` (11) pass.

## Risk

Low. UI/serialization only; no enforcement change (enforcement was already flag-aware via `isApiKeyBlocked`). When the flag is off, behavior is unchanged (`isSpendCapped === (creditState === "capped")`). Under the flag, the customer "capped" badge now reflects the same signal that actually blocks. Adds one Redis read per key on keys-list/create/update/disable responses under the flag (small key counts, not a hot path).

## Deploy Plan

Standard deploy. No migration, no config. Behavior only changes for workspaces with `enforce_user_spend_limit_rate_cap` enabled.

### DIFF
--- a/front/lib/api/credits/api_keys_usage.ts
+++ b/front/lib/api/credits/api_keys_usage.ts
@@ -1,7 +1,9 @@
 import { makeApiKeySpendLimitAwuCreditsRateLimitKey } from "@app/lib/api/assistant/rate_limits";
+import type { RateLimiterState } from "@app/lib/api/credits/members_usage";
 import { fetchConsumedAwuCreditsByApiKeyName } from "@app/lib/api/credits/members_usage";
 import type { Authenticator } from "@app/lib/auth";
 import { microCreditsToCredits } from "@app/lib/credits/units";
+import { USER_AWU_WARNING_PERCENTAGE } from "@app/lib/metronome/alerts/spend_limits";
 import { fetchPerApiKeyAwuUsage } from "@app/lib/metronome/per_api_key_usage";
 import { KeyResource } from "@app/lib/resources/key_resource";
 import { resolveSpendLimitCycleBounds } from "@app/lib/spend_limits/cycle";
@@ -24,6 +26,12 @@ export type ApiKeyUsageType = {
   // figures this one is per key, not per name. Null when the billing period
   // can't be resolved.
   rateLimiterSpendAwuCredits: number | null;
+  // The rate-limiter's verdict for this key's cap: "capped" (counter ≥ cap),
+  // "near_limit" (≥ 80%), or "ok", from `rateLimiterSpendAwuCredits` vs
+  // `monthlyCapAwuCredits`. Null when the key has no cap or the counter couldn't
+  // be read. Independent of the enforcement flag (surfaced beside the Metronome
+  // "Credit state" column to spot divergence).
+  rateLimiterState: RateLimiterState | null;
   // Metronome-side per-API-key AWU consumption for the current billing cycle
   // (the value reconcile and the cap alert read). Null when Metronome isn't
   // configured or the read failed.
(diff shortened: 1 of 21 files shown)


tdraier:

- #31697
    - #31876 ◀
        - #31890
            - #31893
                - #31914
                    - #31916
                        - #31928

tdraier:
> Please add the sse-ack label to acknowledge that a front-sse deploy is required alongside the front deploy.

New methods only, no deploy required

front/lib/spend_limits/api_key_cap_status.ts
```diff
@@ -0,0 +1,68 @@
...
+
+  // Redis reads (not database queries), one per key of the workspace.
+  const entries = await concurrentExecutor(
+    keys,
+    async (key) => {
+      const cap = key.monthlyCapAwuCredits;
+      if (cap === null) {
+        return [key.id, false] as const;
+      }
+      const result = await getFixedWindowCount({
```
avervaet, line 55: Nit: We might want to reuse the readApiKeySpendLimitCountWithLazySeed called by ront/lib/api/keys/spend_limit.ts:408-423 
Because now we can be in a state where the user see an ok state but request are blocked if ES and redis diverges.

front/lib/api/credits/api_keys_usage.ts
```diff
@@ -132,15 +140,31 @@ export async function getApiKeysUsage(
...
-    rateLimiterSpendAwuCredits: rateLimiterSpendByKeyId.get(key.id) ?? null,
-    metronomeConsumedAwuCredits: metronomeConsumedByName.get(key.name) ?? null,
-    monthlyCapAwuCredits: key.monthlyCapAwuCredits,
-  }));
+  const apiKeys = keys.map((key) => {
+    const rateLimiterSpendAwuCredits =
+      rateLimiterSpendByKeyId.get(key.id) ?? null;
+    const cap = key.monthlyCapAwuCredits;
+    let rateLimiterState: RateLimiterState | null = null;
+    if (cap !== null && cap > 0 && rateLimiterSpendAwuCredits !== null) {
```
avervaet, line 148: nit: Redis read failures to 0 so we will show an ok state in this case instead of a failure. We might want to preserve read failures as null or an explicit error ?

avervaet (approved):

front/lib/spend_limits/api_key_cap_status.ts
tdraier: we don't want to seed here just for display, so it can indeed shows 0 while the user will be capped once data is seeded ( so when they try to send a message)

front/lib/api/credits/api_keys_usage.ts
tdraier: Intentional fail-open :  a Redis read error falls to 0 , which matches the check in the user path. still the es/mt/rl column allow us to see the divergence

PR #31877, opened 2026-09-04
feat(llm): expose the Dust-managed LLM credentials on their own

Nils-Fedrigo:
## Description

Factors the Dust-managed key lookups out of `getLlmCredentials` into a standalone
`getDustManagedLlmCredentials`, resolved from the environment alone.

Pure refactor: the three local consts become module-level functions and the non-BYOK
branch returns the new helper, which builds the exact same object it built inline. No
behavior change, for BYOK or non-BYOK workspaces.

The extraction is what lets a caller with no `Authenticator` reach Dust's own
credentials without inventing a workspace — the model health probe runs in a Temporal
worker and needs exactly that.

First of a 5-PR stack adding a shadow-phase model health circuit breaker:
this → #31878 → #31879 → #31880 → #31881.

## Tests

Existing suites and typecheck only — there is no observable change to test. Worth a
reviewer's eye on the object spread order in `getDustManagedLlmCredentials`, which is
preserved from the original (base variables, then other providers, then BYOK
providers, so `GOOGLE_AI_STUDIO_API_KEY` resolves the same way it did before).

### DIFF
--- a/front/lib/api/provider_credentials.ts
+++ b/front/lib/api/provider_credentials.ts
@@ -21,6 +21,60 @@ const BYOK_TRANSITION_BYOK_KEYS_RATIO = 1; // 100%
 export const MISSING_EMBEDDING_API_KEY_ERROR_MESSAGE =
   "An OpenAI API key is required to perform this action. Please configure it in your workspace settings or contact an admin.";
 
+const env = (key: string) =>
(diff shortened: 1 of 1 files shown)


Nils-Fedrigo:

- #31877 ◀
    - #31878
        - #31879
            - #31880
                - #31881

frankaloia (approved):

PR #31878, opened 2026-09-04
feat(llm): add the model health counter keys and window read

Nils-Fedrigo:
## Description

First slice of the model health circuit breaker. Nothing calls it yet.

Per-endpoint counters live in Redis hashes bucketed by UTC minute, one key per
(provider, model, host, minute). `readEndpointWindow` sums the five buckets ending at
now into an `{ attempts, providerErrors }` window, pipelined into a single round trip.
The current minute is included even though it is still being written to: a partial
bucket contributes to both sides of the ratio, so leaving it out would cost a minute of
detection latency for nothing.

The endpoint — not the model — is the unit of health, matching `DegradedModelEndpointType`:
the same model is served from several hosts (Claude via `anthropic` and `agent-platform`,
Gemini via `google-ai-studio` and `agent-platform`) and an incident usually hits one of them.

Also teaches the Redis test mock `hIncrBy`/`expire`, and has the hash commands returned
rather than assigned in place so tests can call them off the client without a cast.

PR 2 of 5: #31877 → this → #31879 → #31880 → #31881.

## Tests

window.test.ts

## Risk 

Low

## Deploy plan

Standard


Nils-Fedrigo:

- #31877
    - #31878 ◀
        - #31879
            - #31880
                - #31881

front/lib/api/llm/health/config.ts
```diff
@@ -0,0 +1,9 @@
+/**
+ * Model health circuit breaker parameters.
```
Nils-Fedrigo, line 2: we'll add more in later PRs

front/lib/api/llm/health/keys.ts
```diff
@@ -0,0 +1,41 @@
...
+import type { DegradedModelEndpointType } from "@app/lib/model_constructors/types/degradations";
+import assert from "assert";
+
+const KEY_PREFIX = "mh";
+
+export const ATTEMPTS_FIELD = "attempts";
+
+// Only provider-attributed errors land here, so the name says so: a reader that
+// took this for "all failures" would compute a ratio the threshold is not
+// calibrated against.
```
Nils-Fedrigo:
```suggestion
```

front/lib/model_constructors/types/degradations.ts
```diff
@@ -4,6 +4,15 @@ import type {
...
 
+/**
+ * The unit of degradation, and therefore of health: the same model is served
+ * from several hosts (Claude via `anthropic` and `agent-platform`, Gemini via
+ * `google-ai-studio` and `agent-platform`) and an incident usually hits one of
+ * them.
+ *
+ * When logging one of these, rename `host` to `modelHost`: `host` is reserved by
+ * our log infrastructure and gets dropped or reinterpreted.
+ */
```
Nils-Fedrigo:
```suggestion
// if logging, rename host->modelHost (reserved keyword)
```

front/lib/api/llm/health/config.ts
```diff
@@ -0,0 +1,9 @@
...
+ *
+ * Kept free of any Temporal import so the detection rails can be unit tested
+ * without a workflow environment.
+ */
+
+// Length of the sliding window, in whole UTC minutes.
+export const WINDOW_MINUTES = 5;
```
frankaloia, line 9: nit: It's more like this is MAX_WINDOW_LIMITS as the window will always be 4 minutes + the amount of seconds elapsed in current minute

front/lib/model_constructors/types/degradations.ts
```diff
@@ -4,6 +4,7 @@ import type {
   ModelProviderIdType,
 } from "@app/types/assistant/models/types";
 
+// if logging, rename host->modelHost (reserved keyword)
```
frankaloia, line 7: nit: likely to get missed?

front/lib/api/llm/health/keys.ts
```diff
@@ -0,0 +1,38 @@
...
+export const ATTEMPTS_FIELD = "attempts";
+
+export const PROVIDER_ERRORS_FIELD = "error_provider";
+
+/** `yyyymmddHHMM` in UTC. */
+export function minuteBucket(date: Date): string {
+  return date.toISOString().slice(0, 16).replace(/[-:T]/g, "");
+}
+
+/** The `WINDOW_MINUTES` buckets ending at (and including) `now`. */
```
frankaloia: I think it's worth a comment in here on why we are using minute buckets + the downside.

frankaloia (approved):
Logic LGTM

PR #31879, opened 2026-09-04
feat(llm): record every model attempt into the health counters

Nils-Fedrigo:
## Description

Hooks `recordLLMAttempt` into the stream telemetry path, where the terminal outcome of a
run is already known. Write-only for now: the counters accumulate, nothing reads them
back until #31881.

Fire and forget on purpose — the call is `void`ed, wrapped in a try/catch, and reports
failures to statsd rather than the logs (this runs once per attempt, so a Redis outage
would otherwise put one line per attempt in the logs; the connection error itself is
already logged once by the client's `error` handler). `HINCRBY` is additive, so N pods
need no coordination and no read-modify-write, and a dropped increment cannot move a
threshold measured over hundreds of attempts.

Only provider-attributed errors count towards the numerator, matching the
`error_source:provider` filter our Datadog monitor already applies to `llm_error.count`.

Batch attempts are deliberately left out: a batch is submitted then polled over hours,
so its error lands at a timestamp unrelated to when the provider was unhealthy, and
folding that into a five-minute window would smear it.

PR 3 of 5: #31877 → #31878 → this → #31880 → #31881.

## Tests

Unit tests covering the aggregation, that a Dust-attributed failure counts as an attempt
but not as an error, the TTL, additivity against what another pod already wrote,
per-minute bucketing, the noop endpoint being skipped, and that a failed write resolves
rather than rejecting (callers `void` this, so a rejection would be an unhandled one on
the request path). `npm run test -- llm/health`.

## Risk

This is the only PR in the stack that touches the inference request path. The exposure is
one extra `void`ed Redis pipeline per attempt, after the outcome is already settled: it
cannot throw into the caller, cannot delay the stream, and degrades to a statsd counter
if Redis is unavailable. Safe to roll back on its own — nothing depends on the counters yet.

Worth watching after deploy: `model_health.write_error.count` should stay flat, and Redis
memory should rise by roughly (endpoints × 15 live minute-buckets) small hashes, each
expiring on its own via the 900s TTL.

### DIFF
--- a/front/lib/api/llm/llm.ts
+++ b/front/lib/api/llm/llm.ts
@@ -84,8 +84,6 @@ export abstract class LLM<
   protected responseFormat: string | null;
   protected bypassFeatureFlag: boolean;
   protected metadata: LLMClientMetadata;
-  // Kept alongside modelId/providerId because model health is tracked per
-  // endpoint, see `DegradedModelEndpointType`.
   protected host: Host;
   // Temporary during the router migration; "new" is set by BaseTransition.
   protected readonly router: "legacy" | "new" = "legacy";
@@ -202,10 +200,6 @@ export abstract class LLM<
     const baseTags = this.getTelemetryTags({ surface: "stream" });
     const latencyTags = this.getLatencyTelemetryTags({ surface: "stream" });
 
-    // Feeds the model health circuit breaker. Fired and forgotten: nothing here
-    // waits on Redis or fails with it. The endpoint is always concrete at this
-    // point -- `auto`, `auto_fast` and `auto_complex` are collapsed to a real
-    // model before the LLM is built.
     void recordLLMAttempt({
       endpoint: {
         modelId: this.modelId,
--- a/front/lib/api/llm/llm.ts
+++ b/front/lib/api/llm/llm.ts
@@ -3,6 +3,7 @@ import {
   contributeFreeUsageCostForUser,
   isFreeUsageContext,
 } from "@app/lib/api/llm/free_usage";
+import { recordLLMAttempt } from "@app/lib/api/llm/health/counters";
 import { LLMRunLifecycle } from "@app/lib/api/llm/run_lifecycle";
 import type {
   LLMAttemptOutcome,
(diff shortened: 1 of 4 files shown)


Nils-Fedrigo:

- #31877
    - #31878
        - #31879 ◀
            - #31880
                - #31881

front/lib/api/llm/llm.ts
```diff
@@ -196,6 +202,19 @@ export abstract class LLM<
     const baseTags = this.getTelemetryTags({ surface: "stream" });
     const latencyTags = this.getLatencyTelemetryTags({ surface: "stream" });
 
+    // Feeds the model health circuit breaker. Fired and forgotten: nothing here
+    // waits on Redis or fails with it. The endpoint is always concrete at this
+    // point -- `auto`, `auto_fast` and `auto_complex` are collapsed to a real
+    // model before the LLM is built.
```
Nils-Fedrigo:
```suggestion
```

front/lib/api/llm/llm.ts
```diff
@@ -82,6 +84,9 @@ export abstract class LLM<
   protected responseFormat: string | null;
   protected bypassFeatureFlag: boolean;
   protected metadata: LLMClientMetadata;
+  // Kept alongside modelId/providerId because model health is tracked per
+  // endpoint, see `DegradedModelEndpointType`.
```
Nils-Fedrigo:
```suggestion
```

front/lib/api/llm/llm.ts
```diff
@@ -808,6 +821,11 @@ export abstract class LLM<
       }
       statsDMetrics.increment("llm_interaction.count", 1, metricTags);
 
+      // TODO(detect_outage): batch attempts deliberately do not feed the model
```
frankaloia: Is this a TODO? Or just a comment for clarity?

frankaloia (approved):

front/lib/api/llm/llm.ts
Nils-Fedrigo: maybe a todo but let's keep it as a comment for now in deed!

PR #31880, opened 2026-09-04
feat(llm): add the model health recovery workflow and its probe

Nils-Fedrigo:
## Description

Second half of the breaker, and inert on its own: nothing starts these workflows until
#31881, so the worker sits idle once deployed.

A recovery workflow *is* the degraded state in this phase. Its id is derived from the
endpoint, so Temporal's workflow-id uniqueness doubles as cross-pod dedup and concurrent
starts collapse into one run. It holds for `MIN_DEGRADED_DURATION_MS` on a durable timer
(a worker restart mid-hold costs nothing), then probes until the endpoint answers,
continuing as new after 100 rounds to keep history bounded on an outage that drags on.

A probe is one synthetic inference that bypasses `LLM` entirely — no `Run` row, no usage
row, no Langfuse trace, no counter write — so it can never feed the signal it measures.
It resolves on the first token and returns early to stop the generation rather than
paying for a full response. Three sequential probes must all pass before a recovery is
declared, so a provider that is only intermittently serving cannot pass on a lucky burst.

Transitions are logged and counted, nothing is persisted: in this shadow phase the logs
and the matching Datadog series are the whole record of what the breaker would have done.
The endpoint's host goes out as `modelHost` — `host` is reserved by our log infrastructure.

PR 4 of 5: #31877 → #31878 → #31879 → this → #31881.

## Tests

Typecheck and the existing suites; the probe itself is exercised against live providers
rather than mocked. Before merging #31881, I plan to deploy this worker and start one
recovery workflow by hand against a healthy endpoint, to confirm the probe path, the
early return on first token, and the transition logs.

## Risk

Probes spend real provider credits under the Dust-managed keys, but the prompt is a
single character with reasoning off and we abort at the first token, so a round is three
near-minimal calls. The activity is capped at 5 minutes with `maximumAttempts: 1` — the
probes inside it are the retry, and retrying the activity on top would multiply the calls
and blur what "3 probes passed" means.

Inert until #31881, so this is safe to land and observe on its own.

## Deploy Plan

Adds a new `model_health` Temporal worker. It needs its deployment before #31881 merges,
otherwise breaches would start workflows with no worker to pick them up. The
`shutdownGraceTime` is set to 70s and must stay at the deployment's
`terminationGracePeriodSeconds` minus a 10s buffer.

Nothing to do for this PR itself — no schedule is registered, and no workflow is started
until the next one.

### DIFF
--- a/front/lib/api/llm/health/transitions.ts
+++ b/front/lib/api/llm/health/transitions.ts
@@ -1,0 +1,58 @@
+import type { ModelHealthWindowType } from "@app/lib/api/llm/health/types";
+import type { DegradedModelEndpointType } from "@app/lib/model_constructors/types/degradations";
+import { statsDMetrics } from "@app/lib/utils/statsd";
+import logger from "@app/logger/logger";
+
+export type ModelHealthTransitionType =
+  | "degraded"
+  | "recovered"
+  | "probe_failed";
+
+/**
+ * The only output of the shadow phase. Nothing is persisted, so these logs and
+ * the matching Datadog series are the whole record of what the breaker would
+ * have done.
+ */
+export function logModelHealthTransition({
+  endpoint,
+  transition,
+  window,
+  degradedForMs,
+}: {
+  endpoint: DegradedModelEndpointType;
+  transition: ModelHealthTransitionType;
+  window?: ModelHealthWindowType;
+  degradedForMs?: number;
+}): void {
+  const { modelId, providerId, host } = endpoint;
+
+  logger.info(
+    {
+      modelId,
+      providerId,
+      // `host` is reserved by our log infrastructure, so the endpoint's host
+      // goes out under a name of our own.
+      modelHost: host,
+      transition,
+      ...(window
+        ? {
+            attempts: window.attempts,
+            providerErrors: window.providerErrors,
+            errorRatio:
+              window.attempts > 0
+                ? window.providerErrors / window.attempts
+                : null,
+          }
+        : {}),
+      ...(degradedForMs !== undefined ? { degradedForMs } : {}),
+    },
+    "Model health transition"
+  );
+
+  statsDMetrics.increment("model_health.transition.count", 1, [
+    `transition:${transition}`,
+    `model_id:${modelId}`,
+    `provider_id:${providerId}`,
+    `model_host:${host}`,
+  ]);
+}
--- a/front/temporal/model_health/client.ts
+++ b/front/temporal/model_health/client.ts
@@ -1,0 +1,45 @@
+import type { DegradedModelEndpointType } from "@app/lib/model_constructors/types/degradations";
+import { getTemporalClientForFrontNamespace } from "@app/lib/temporal";
+import {
+  QUEUE_NAME,
+  recoveryWorkflowId,
+} from "@app/temporal/model_health/config";
+import { modelHealthRecoveryWorkflow } from "@app/temporal/model_health/workflows";
+import type { Result } from "@app/types/shared/result";
+import { Err, Ok } from "@app/types/shared/result";
+import { normalizeError } from "@app/types/shared/utils/error_utils";
+import { WorkflowExecutionAlreadyStartedError } from "@temporalio/client";
+
+export type LaunchRecoveryOutcome = "started" | "already_degraded";
+
+/**
+ * Declares an endpoint degraded by starting its recovery workflow.
+ *
+ * Idempotent by construction: the workflow id is derived from the endpoint, so
+ * every pod that detects the same breach in the same instant lands on the same
+ * id and all but one get `WorkflowExecutionAlreadyStartedError` back. That
+ * rejection is not a failure -- it is how we learn the endpoint was already
+ * degraded -- so it comes back as an outcome rather than an error.
+ */
+export async function launchModelHealthRecovery(
+  endpoint: DegradedModelEndpointType
+): Promise<Result<LaunchRecoveryOutcome, Error>> {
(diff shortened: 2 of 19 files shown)


Nils-Fedrigo:

- #31877
    - #31878
        - #31879
            - #31880 ◀
                - #31881

front/lib/api/llm/health/transitions.ts
```diff
@@ -0,0 +1,58 @@
...
+  degradedForMs?: number;
+}): void {
+  const { modelId, providerId, host } = endpoint;
+
+  logger.info(
+    {
+      modelId,
+      providerId,
+      // `host` is reserved by our log infrastructure, so the endpoint's host
+      // goes out under a name of our own.
```
Nils-Fedrigo, line 34:
```suggestion
```

front/lib/api/llm/health/transitions.ts
```diff
@@ -0,0 +1,58 @@
...
+export type ModelHealthTransitionType =
+  | "degraded"
+  | "recovered"
+  | "probe_failed";
+
+/**
+ * The only output of the shadow phase. Nothing is persisted, so these logs and
+ * the matching Datadog series are the whole record of what the breaker would
+ * have done.
+ */
```
Nils-Fedrigo, line 15:
```suggestion
```

front/temporal/model_health/client.ts
```diff
@@ -0,0 +1,45 @@
...
+
+/**
+ * Declares an endpoint degraded by starting its recovery workflow.
+ *
+ * Idempotent by construction: the workflow id is derived from the endpoint, so
+ * every pod that detects the same breach in the same instant lands on the same
+ * id and all but one get `WorkflowExecutionAlreadyStartedError` back. That
+ * rejection is not a failure -- it is how we learn the endpoint was already
+ * degraded -- so it comes back as an outcome rather than an error.
+ */
```
Nils-Fedrigo, line 69:
```suggestion
// idempotent by construction, rejection is not a failure
```

front/temporal/model_health/config.ts
```diff
@@ -0,0 +1,28 @@
+import type { DegradedModelEndpointType } from "@app/lib/model_constructors/types/degradations";
+
+const QUEUE_VERSION = 1;
+export const QUEUE_NAME = `model-health-queue-v${QUEUE_VERSION}`;
+
+// Probe rounds before the workflow continues as new, to keep history bounded on
+// an outage that drags on.
```
Nils-Fedrigo:
```suggestion
```

front/temporal/model_health/config.ts
```diff
@@ -0,0 +1,28 @@
...
+export const MAX_PROBE_ROUNDS_PER_RUN = 100;
+
+/**
+ * The workflow id *is* the state in this phase: while a recovery workflow with
+ * this id is running, the endpoint is degraded. Temporal's workflow-id
+ * uniqueness therefore doubles as the cross-pod dedup, and the default
+ * `ALLOW_DUPLICATE` reuse policy frees the id once recovery completes so a later
+ * breach can open a fresh one.
+ */
```
Nils-Fedrigo, line 19:
```suggestion
```

front/temporal/model_health/worker.ts
```diff
@@ -0,0 +1,46 @@
...
+          return {
+            inbound: new ActivityInboundLogInterceptor(ctx, logger),
+          };
+        },
+      ],
+    },
+  });
+
+  // No schedule: recovery workflows are started on demand by the pod that
+  // detects the breach.
```
Nils-Fedrigo, line 44:
```suggestion
```

front/temporal/model_health/workflows.ts
```diff
@@ -0,0 +1,64 @@
...
+import { continueAsNew, proxyActivities, sleep } from "@temporalio/workflow";
+
+const {
+  probeEndpointActivity,
+  logModelHealthProbeFailedActivity,
+  logModelHealthRecoveryActivity,
+} = proxyActivities<typeof activities>({
+  // A round is `PROBES_PER_RECOVERY` sequential provider calls, so give it room
+  // for slow-but-alive providers. This is the probe's only timeout: it runs
+  // outside the agent loop, so the stream watchdog never sees it.
```
Nils-Fedrigo, line 14:
```suggestion
```

front/temporal/model_health/workflows.ts
```diff
@@ -0,0 +1,64 @@
...
+  startToCloseTimeout: "5 minutes",
+  retry: {
+    // The probes inside the activity are the retry. Retrying the activity on top
+    // would multiply the calls and blur what "3 probes passed" means.
```
Nils-Fedrigo, line 18:
```suggestion
```

front/temporal/model_health/workflows.ts
```diff
@@ -0,0 +1,64 @@
...
+ * Started by whichever pod detected the breach; the deterministic workflow id
+ * makes concurrent starts collapse into this single run, and its existence is
+ * what "degraded" means while it lasts.
+ *
+ * Holds for `MIN_DEGRADED_DURATION_MS` on a durable Temporal timer -- a worker
+ * restart mid-hold costs nothing -- then probes until the endpoint answers.
+ *
+ * There is no sleep between failed rounds. A round is three sequential provider
+ * calls, so the loop paces itself on real provider latency rather than spinning.
+ */
```
Nils-Fedrigo, line 41:
```suggestion
```

frankaloia (approved):
Looks clean to me

PR #31881, opened 2026-09-04
feat(llm): close the model health loop with breach detection

Nils-Fedrigo:
## Description

Closes the loop. The only PR in the stack that changes what the system does.

An endpoint is declared degraded when at least `MIN_ATTEMPTS_IN_WINDOW` (100) attempts
landed in the five-minute window and the provider-attributed share crossed
`ERROR_RATIO_THRESHOLD` (20%). The volume floor is what keeps a handful of failures on a
low-traffic endpoint from tripping it, and it is low enough that we fail back quickly
after reopening an endpoint to full traffic; a real outage routinely sits at 60%.

Detection is driven by the write path rather than by a sweep. Only an error can push the
ratio over the threshold, so an error write evaluates the one endpoint it just served —
never a scan of the catalog — throttled to once per `MIN_EVALUATION_INTERVAL_MS` (5s) per
pod. During an outage those writes land hundreds of times a second on the same endpoint,
and evaluating each one would re-read the same window and hand Temporal a rejected
duplicate start hundreds of times to learn the same thing.

Several pods can reach the same conclusion in the same instant, which is fine: the
recovery workflow's deterministic id makes the start idempotent, and only a start that
actually created the workflow is logged as a transition.

Still shadow-only: nothing is persisted and no traffic is steered away from a degraded
endpoint. Declaring one degraded starts its recovery workflow and logs the transition.

The whole path sits behind a `pause_model_health_detection` kill switch, toggled from
Poke. Enabled, `recordLLMAttempt` returns immediately: no counter write, no evaluation, no
workflow start. It exists because the one thing this path must survive is an incident, and
shipping a revert is exactly what we cannot do during one. It is read on every attempt, so
it comes out of an in-process cache refreshed at most every 60s rather than out of Redis --
consulting Redis to decide whether to write to Redis would be self-defeating on the failure
it most needs to cover. That pattern already existed for `use_legacy_acls`, so it moved to
a shared `makeCachedKillSwitch` and both read through it.

PR 5 of 5: #31877 → #31878 → #31879 → #31880 → this.

## Tests

Unit tests for the threshold rails (volume floor, at/just-below the threshold, idle
endpoint) and for `evaluateEndpoint` — that a breaching endpoint is declared degraded,
that a healthy one is left alone, that an already-degraded endpoint is not re-logged as a
transition, and that a failed workflow start does not claim one. Plus the counter-side
tests for the evaluation trigger: only on a provider error, at most once per interval,
and never when the write failed, and that the kill switch takes out both the write and the
detection it triggers. `npm run test -- llm/health`.

## Risk

The thresholds (20% over 200 attempts in 5 minutes, 5s evaluation throttle, 10 minutes
degraded before the first probe) are the tunable part and are calibrated against what the
existing Datadog monitor sees, not against a replay of past incidents. A too-sensitive
setting produces spurious "degraded" logs and one recovery workflow per endpoint; it
cannot affect serving, because nothing consumes the degraded state yet. That is the point
of the shadow phase.

Small and self-contained enough to revert on its own, which leaves the stack in the
write-only state of #31879 -- and the kill switch means we do not have to: enabling
`pause_model_health_detection` in Poke takes the path out within 60s per pod, without a
deploy. Recovery workflows already running are left to finish on their own; they only probe
and log.

## Deploy Plan

Merge after the `model_health` worker from #31880 is actually deployed — this is what
starts recovery workflows, and without a worker they would sit unpicked.

Then watch the `model_health.transition.count` series (tagged by transition, model,
provider and host) for a few days and compare its `degraded` transitions against the
incidents the Datadog `llm_error.count` monitor fires on, before wiring the state into
routing in a follow-up.

### DIFF
--- a/front/lib/api/llm/health/config.ts
+++ b/front/lib/api/llm/health/config.ts
@@ -5,14 +5,29 @@
  * without a workflow environment.
  */
 
+// Share of provider-attributed errors over attempts, within the window, at
+// which an endpoint is declared degraded. A real outage is routinely 60%.
+export const ERROR_RATIO_THRESHOLD = 0.2;
+
 // Length of the sliding window, in whole UTC minutes.
 export const WINDOW_MINUTES = 5;
+
+// Volume floor. Below this the ratio is noise, and it is low enough that we
+// fail back quickly after reopening an endpoint to 100% of traffic.
+export const MIN_ATTEMPTS_IN_WINDOW = 200;
 
 // How long an endpoint stays degraded before the first recovery probe.
 export const MIN_DEGRADED_DURATION_MS = 10 * 60 * 1000;
 
 // Consecutive synthetic probes that must all succeed to declare a recovery.
 export const PROBES_PER_RECOVERY = 3;
+
+// How often one endpoint may be evaluated, per pod. A breach can only begin
+// with an error write, and during an outage those land hundreds of times a
+// second on the same endpoint: evaluating every one would re-read the same
+// window, and hand Temporal a rejected duplicate start, hundreds of times to
+// learn the same thing.
+export const MIN_EVALUATION_INTERVAL_MS = 5_000;
 
 // Counter keys are only ever read across `WINDOW_MINUTES`; the extra headroom
 // covers clock skew between pods.
--- a/front/lib/api/llm/health/config.ts
+++ b/front/lib/api/llm/health/config.ts
@@ -5,15 +5,11 @@
  * without a workflow environment.
  */
 
-// Share of provider-attributed errors over attempts, within the window, at
-// which an endpoint is declared degraded. A real outage is routinely 60%.
 export const ERROR_RATIO_THRESHOLD = 0.2;
 
 // Length of the sliding window, in whole UTC minutes.
 export const WINDOW_MINUTES = 5;
 
-// Volume floor. Below this the ratio is noise, and it is low enough that we
-// fail back quickly after reopening an endpoint to 100% of traffic.
 export const MIN_ATTEMPTS_IN_WINDOW = 200;
 
 // How long an endpoint stays degraded before the first recovery probe.
--- a/front/lib/api/llm/health/config.ts
+++ b/front/lib/api/llm/health/config.ts
@@ -10,7 +10,7 @@ export const ERROR_RATIO_THRESHOLD = 0.2;
 // Length of the sliding window, in whole UTC minutes.
 export const WINDOW_MINUTES = 5;
 
-export const MIN_ATTEMPTS_IN_WINDOW = 200;
+export const MIN_ATTEMPTS_IN_WINDOW = 100;
 
 // How long an endpoint stays degraded before the first recovery probe.
 export const MIN_DEGRADED_DURATION_MS = 10 * 60 * 1000;
--- a/front/lib/api/llm/health/config.ts
+++ b/front/lib/api/llm/health/config.ts
@@ -18,21 +18,12 @@ export const MIN_DEGRADED_DURATION_MS = 10 * 60 * 1000;
 // Consecutive synthetic probes that must all succeed to declare a recovery.
 export const PROBES_PER_RECOVERY = 3;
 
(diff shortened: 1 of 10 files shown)


Nils-Fedrigo:

- #31877
    - #31878
        - #31879
            - #31880
                - #31881 ◀

front/lib/api/llm/health/config.ts
```diff
@@ -5,15 +5,30 @@
  * without a workflow environment.
  */
 
+// Share of provider-attributed errors over attempts, within the window, at
+// which an endpoint is declared degraded. A real outage is routinely 60%.
```
Nils-Fedrigo:
```suggestion
```

front/lib/api/llm/health/config.ts
```diff
@@ -5,15 +5,30 @@
...
+export const ERROR_RATIO_THRESHOLD = 0.2;
+
 // Length of the sliding window, in whole UTC minutes.
 export const WINDOW_MINUTES = 5;
 
+// Volume floor. Below this the ratio is noise, and it is low enough that we
+// fail back quickly after reopening an endpoint to 100% of traffic.
```
Nils-Fedrigo:
```suggestion
```

front/lib/api/llm/health/detect.ts
```diff
@@ -0,0 +1,67 @@
...
+import type { DegradedModelEndpointType } from "@app/lib/model_constructors/types/degradations";
+import logger from "@app/logger/logger";
+import { launchModelHealthRecovery } from "@app/temporal/model_health/client";
+import { normalizeError } from "@app/types/shared/utils/error_utils";
+
+/**
+ * An endpoint is degraded when enough attempts landed in the window and the
+ * share attributed to the provider crossed the threshold. The volume floor is
+ * what keeps a handful of failures on a low-traffic endpoint from tripping it.
+ */
```
Nils-Fedrigo:
```suggestion
```

front/lib/api/llm/health/detect.ts
```diff
@@ -0,0 +1,67 @@
...
+ * Runs on the pod that just served the endpoint, right after it wrote its
+ * counter, so only the endpoint in hand is ever read -- never a sweep of the
+ * catalog. Several pods can reach the same conclusion at the same instant,
+ * which is fine: starting the recovery workflow is idempotent, and its
+ * deterministic workflow id is what makes it so.
+ *
+ * There is no state to consult first. In this phase a running recovery workflow
+ * *is* the degraded state, so re-declaring an endpoint that is already degraded
+ * costs one rejected start and nothing else.
+ */
```
Nils-Fedrigo, line 29:
```suggestion
```

front/lib/api/llm/health/detect.ts
```diff
@@ -0,0 +1,67 @@
...
+        providerId: endpoint.providerId,
+        modelHost: endpoint.host,
+      },
+      "Failed to start the model health recovery workflow"
+    );
+    return;
+  }
+
+  // Only a launch that actually created the workflow is a transition: a
+  // rejected duplicate means the endpoint was already degraded.
```
Nils-Fedrigo:
```suggestion
```

front/lib/api/llm/health/config.ts
```diff
@@ -5,15 +5,26 @@
  * without a workflow environment.
  */
 
+export const ERROR_RATIO_THRESHOLD = 0.2;
+
 // Length of the sliding window, in whole UTC minutes.
 export const WINDOW_MINUTES = 5;
 
+export const MIN_ATTEMPTS_IN_WINDOW = 200;
```
frankaloia: This is inconsistent with what we have with DDOG configuration. Should make consistent one way or another.

frankaloia (approved):
LGTM

front/lib/api/llm/health/config.ts
Nils-Fedrigo: yeah lol, let's put it to 100 everywhere for now

PR #31886, opened 2026-09-04
Enable EU routing for Gemini 3.6-3.8

spolu:
## Description

- Upgrade @google/genai to a version that correctly routes the Vertex AI eu location through aiplatform.eu.rep.googleapis.com.
- Register the existing EU Agent Platform endpoints for Gemini 3.6 Flash, 3.7 Flash, and 3.8 Flash in both stream registries and the shared test setup registry.
- Mark europe-west1 availability as enabled for all three models and update the non-BYOK routing expectation.

This fixes the invalid eu-aiplatform.googleapis.com hostname generated by the older SDK. EU availability was also verified directly against Agent Platform.

## Tests

- 138/138 live Agent Platform EU endpoint cases passed across Gemini 3.6, 3.7, and 3.8.
- 26/26 focused routing and model configuration tests passed.
- Biome and git diff checks passed.
- Full TypeScript checking still reports unrelated pre-existing errors in Google Calendar, Metronome, Mistral, OpenAI, and Anthropic code.

## Risk

Low. The new endpoints use existing constructors and pricing configuration. The main risk is provider-side EU availability; this is mitigated by the complete live endpoint test suites for all three models.

## Deploy Plan

Deploy normally. The SDK upgrade and endpoint registration ship together, so EU routing cannot select these endpoints before the corrected hostname logic is available.

### DIFF
--- a/front/lib/api/llm/index.test.ts
+++ b/front/lib/api/llm/index.test.ts
@@ -47,7 +47,7 @@ describe("getWorkspaceFilter", () => {
     expect(flashEndpoints.every((e) => e.host === AGENT_PLATFORM_HOST)).toBe(
       true
     );
-    expect(flashEndpoints.map((e) => e.region)).toEqual(["global"]);
+    expect(flashEndpoints.map((e) => e.region)).toEqual(["eu", "global"]);
 
     const proEndpoints = getStreamEndpoints(workspaceConfig, {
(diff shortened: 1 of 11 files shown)


spolu:
cc @pmilliotte

flvndvd (approved):

PR #31887, opened 2026-09-04
[front] enh: show active user percentage in consumption tooltip

aubin-tchoi:
## Description

The consumption chart tooltip now shows each bucket’s active-user share next to the absolute count: DAU if the granularity is daily, WAU if weekly and MAU if monthly.

It reads the total user count from the existing consumption overview for the current analytics scope, which is most likely already in SWR cache. If the overview is unavailable, the absolute active-user count remains visible.

Screenshot 2026-09-04 at 3 51 49 PM

## Tests

- Updated existing tests.
- Tested locally.

## Risk

- Low.

## Deploy Plan

- Deploy front.

### DIFF
--- a/front/components/workspace/analytics/consumption/ConsumptionChart.tsx
+++ b/front/components/workspace/analytics/consumption/ConsumptionChart.tsx
@@ -178,6 +178,7 @@ interface ConsumptionDailyTooltipProps
   partialTimestamp: number | undefined;
   currentBucketLabel: string;
   showActiveUsers: boolean;
+  totalUsers: number | null;
 }
 
 function ConsumptionDailyTooltip({
@@ -188,6 +189,7 @@ function ConsumptionDailyTooltip({
   partialTimestamp,
   currentBucketLabel,
   showActiveUsers,
+  totalUsers,
 }: ConsumptionDailyTooltipProps) {
   const datum = payload?.[0]?.payload;
   if (!active || !isConsumptionTimeseriesPoint(datum)) {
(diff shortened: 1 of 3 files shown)


front/components/workspace/analytics/consumption/ConsumptionChart.tsx
```diff
@@ -514,6 +524,8 @@ function WorkspaceConsumptionDailyChart({
   analyticsScope,
   disabled,
 }: ConsumptionChartProps) {
+  const showActiveUsers =
+    analyticsScope?.kind !== "personal" && filter?.users?.length !== 1;
```
avervaet, line 528: nit: we could reuse showActiveUsers here

avervaet (approved):

front/components/workspace/analytics/consumption/ConsumptionChart.tsx
```diff
@@ -526,16 +538,22 @@ function WorkspaceConsumptionDailyChart({
...
     <ConsumptionDailyChart
       timeseries={timeseries}
       isTimeseriesLoading={isTimeseriesLoading}
       isTimeseriesError={Boolean(isTimeseriesError)}
       emptyMessage="No consumption over this period."
-      showActiveUsers={
-        analyticsScope?.kind !== "personal" && filter?.users?.length !== 1
-      }
+      showActiveUsers={showActiveUsers}
+      totalUsers={overview?.members.total ?? null}
```
avervaet, line 556: members.total is the selected period’s active-member count, not the workspace’s total members. 

This would mean daily/weekly percentages use the wrong denominator.

avervaet: Or maybe this is intended behavior as we are zooming on a period

aubin-tchoi: ah I meant to omit the period and filter from `useConsumptionOverview`
good catch, will fix

PR #31889, opened 2026-09-04
[Admin Governance] Shadow agent editor lists (11a)

philipperolet:
## Description

Part 11a of the agent permission migration. When `group_permissions_shadow` is enabled, editor-list reads compare legacy group membership with editors found through grants and log differences. Responses continue to use the legacy editor lists.

Adds the single and batch resource lookups and covers editor endpoints and agent configuration context.

PR11 parts: **11a (this PR)** → [11b: permissions and usage](https://github.com/dust-tt/dust/pull/31919). [11c: agent views](https://github.com/dust-tt/dust/pull/31920) is independent and targets `main`.

## Risks

Blast radius: Agent editor lists and agent configuration context when shadowing is enabled.
Risk: standard

## Deploy Plan

- Deploy front.
- After PR9's dual writes and PR10's editor-grant backfill are complete, enable `group_permissions_shadow` progressively.
- Monitor editor mismatch/error logs; disable the flag if needed. Finish all PR11 parts and verify parity before PR12.

## Tests

- [x] Agent resource and editor shadow tests (6 tests).
- [x] Agent editor endpoint tests (20 tests).

### DIFF
--- a/front/lib/api/assistant/editors.ts
+++ b/front/lib/api/assistant/editors.ts
@@ -1,9 +1,54 @@
+import { shadowCompare } from "@app/lib/api/permissions/shadow";
 import type { Authenticator } from "@app/lib/auth";
+import { AgentResource } from "@app/lib/resources/agent_resource";
 import { GroupResource } from "@app/lib/resources/group_resource";
 import { UserResource } from "@app/lib/resources/user_resource";
 import type { LightAgentConfigurationType } from "@app/types/assistant/agent";
 import { removeNulls } from "@app/types/shared/utils/general";
 import type { UserType } from "@app/types/user";
+import assert from "assert";
+
+function sortedUserModelIds(users: { id: number }[]): number[] {
+  return [...new Set(users.map((user) => user.id))].sort((a, b) => a - b);
+}
+
+function sameModelIds(left: number[], right: number[]): boolean {
+  return (
+    left.length === right.length &&
+    left.every((modelId, index) => modelId === right[index])
+  );
+}
+
+export async function shadowAgentEditors(
+  auth: Authenticator,
+  agentConfiguration: LightAgentConfigurationType,
+  legacyEditors: UserResource[],
+  callSite: string
+): Promise<UserResource[]> {
+  await shadowCompare({
+    auth,
+    legacy: sortedUserModelIds(legacyEditors),
+    candidate: async () => {
+      const resource = await AgentResource.fetchByAgentConfiguration(
+        auth,
+        agentConfiguration
+      );
+      const editors = await resource.listEditors(auth);
+      assert(editors !== null);
+      return sortedUserModelIds(editors);
+    },
+    context: {
+      check: "agent_editors",
+      callSite,
+      agentId: agentConfiguration.sId,
+      agentConfigurationModelId: agentConfiguration.id,
+      workspaceId: auth.getNonNullableWorkspace().sId,
+    },
+    equals: sameModelIds,
+  });
+
+  return legacyEditors;
+}
 
 export const getAuthors = async (
   agentConfigurations: LightAgentConfigurationType[]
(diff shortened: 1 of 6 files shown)


front/lib/api/assistant/editors.ts
```diff
@@ -1,9 +1,54 @@
...
+}
+
+function sameModelIds(left: number[], right: number[]): boolean {
+  return (
+    left.length === right.length &&
+    left.every((modelId, index) => modelId === right[index])
+  );
+}
+
+export async function shadowAgentEditors(
```
philipperolet: nit: getAgentEditorsShadowed

front/lib/api/assistant/editors.ts
```diff
@@ -25,15 +70,71 @@ export const getEditors = async (
...
   return memberUsers;
 };
 
+async function shadowAgentEditorsBatch(
+  auth: Authenticator,
+  agents: LightAgentConfigurationType[],
+  legacyEditors: Record<string, UserType[]>
+): Promise<void> {
+  const customAgents = agents.filter((agent) => agent.scope !== "global");
+  const canonicalLegacy = customAgents
```
philipperolet: what does canonicalLegacy mean?
Does it mean "normalizedLegacyAgents"?

front/lib/api/assistant/editors.ts
```diff
@@ -25,15 +70,71 @@ export const getEditors = async (
...
+      const resources = await AgentResource.fetchByAgentConfigurations(
+        auth,
+        customAgents
+      );
+      const editorsByAgentId = await AgentResource.batchListEditors(
+        auth,
+        resources
+      );
+
+      return customAgents
```
philipperolet: the return is 9 lines. More than ~5-7 lines of code with multiple operations, nested calls, multiple indentation levels are hard to grasp. It's better to provide a top-doewn explanation in that case.  Either explain in a 1-line comment or wrap in a variable name
// normalize candidates similarly to legacy [or something like that]

front/lib/resources/agent_resource.ts
```diff
@@ -115,6 +120,143 @@ export class AgentResource implements WithAccessControl {
...
+    return configurations.map((configuration) => {
+      assert(configuration.scope !== "global");
+      assert(
+        configuration.versionAuthorId !== null,
+        "Unexpected: custom agent author is missing"
+      );
+      const agent = agentById.get(configuration.sId);
+      assert(agent, "Unexpected: agent identity is missing");
+
+      return new AgentResource(
```
philipperolet: here it's 8 lines but it's completely fine because it's 1 call with 1 level of indention, easy to understand

front/lib/resources/agent_resource.ts
```diff
@@ -115,6 +120,143 @@ export class AgentResource implements WithAccessControl {
...
+    const agents = await AgentModel.findAll({
+      where: {
+        sId: configurations.map((configuration) => configuration.sId),
+        workspaceId: auth.getNonNullableWorkspace().id,
+      },
+      attributes: ["id", "sId", "workspaceId"],
+    });
+    const agentById = new Map(agents.map((agent) => [agent.sId, agent]));
+
+    return configurations.map((configuration) => {
```
philipperolet: return inside a return with assert; here same rule as above

front/lib/resources/agent_resource.ts
philipperolet: side note: codebase sometimes uses makeNew rather than constructor; check why, and check if it applies in this case and explain.
Question: why not try and reuse `fromAgentConfigurationModel` here? it seems there's duplication?

front/lib/resources/agent_resource.ts
```diff
@@ -115,6 +120,143 @@ export class AgentResource implements WithAccessControl {
...
+        grantType: "editor",
+        resourceType: "agent",
+        resourceId: this.id,
+      }
+    );
+
+    return group ? group.getActiveMembers(auth) : [];
+  }
+
+  static async batchListEditors(
```
philipperolet, line 178: if we have this, could the function above be simplified by a call to batchListEditors with agents: [this] ?

front/lib/resources/agent_resource.ts
```diff
@@ -115,6 +120,143 @@ export class AgentResource implements WithAccessControl {
...
+      )
+    );
+    const membershipsByGroupId =
+      await GroupResource.getActiveMembershipsForGroups(auth, [
+        ...groupByAgentModelId.values(),
+      ]);
+    const userModelIds = [
+      ...new Set(Object.values(membershipsByGroupId).flat()),
+    ];
+    if (userModelIds.length === 0) {
```
philipperolet: is that needed?
principle: zero-length check are useful only if:
- the code for non-zero length does not simply produce the correct output for 
- or, the code for non-zero length check would perform a costly operation (IO, etc.) even with a 0-lenght input
Otherwise no need to have redundant specific code for 0 length
Assume callees respect this rule, so e.g. fetchByModelIds below is not costly with a 0-length input.

philipperolet: separately, check the callees do respect the rule (if not, we'll make a PR so they do respect it)

philipperolet:

- #31889 ◀
    - #31919

front/lib/api/assistant/editors.ts
philipperolet: Addressed in 6343f5214b: renamed the helper to `getAgentEditorsShadowed` and updated all callers.

front/lib/api/assistant/editors.ts
philipperolet: Yes: this is the normalized legacy editor list, not a separate domain concept. Renamed it to `normalizedLegacyEditors` in 6343f5214b. Its entries are `[agentId, sorted editor model IDs]`.

front/lib/api/assistant/editors.ts
philipperolet: Addressed in 6343f5214b: the expression now builds `normalizedCandidateEditors`, then returns it. This gives the reader the intent before the nested mapping/sorting, and mirrors `normalizedLegacyEditors` without introducing another helper.

front/lib/resources/agent_resource.ts
philipperolet: Agreed: a single factory call with named fields remains readable over several lines. Addressed the follow-up in 6343f5214b by reusing `fromAgentConfigurationModel`, as the single-item fetch already does. I checked `UserResource.makeNew` and `SpaceResource.makeNew`: they create database rows and perform creation-side work (indexing or grants). Here we are wrapping an existing agent identity, so introducing or calling a persistence factory would be the wrong lifecycle operation.

front/lib/resources/agent_resource.ts
philipperolet: Addressed in 6343f5214b: the mapping is now introduced as `const resources = ...`, followed by `return resources`. The name explains the result before the callback's assertions and resource construction.

front/lib/resources/agent_resource.ts
philipperolet: Addressed in 6343f5214b: `listEditors` now delegates to `batchListEditors(auth, [this])` and reads its entry. The batch method owns the distinction between global agents (`null`) and custom agents without editors (`[]`). An assertion checks the promised map entry rather than silently defaulting a missing entry. The resource/editor tests and all 20 editor endpoint tests pass.

front/lib/resources/agent_resource.ts
philipperolet: I agree with the principle, but the requested callee audit found that its I/O assumption is not true yet. `UserResource.fetchByModelIds([])` and `MembershipResource.getActiveMemberships({ workspace, users: [] })` both reach `findAll`; `GroupResource.getActiveMembershipsForGroups([])` also issues a query. `GroupPermissionResource.findRegularAutoGroupsForGrants` already short-circuits empty grants. I retained this guard for now, with a short explanation, because removing it would add two needless database calls for the no-editors case. The separate follow-up should add empty-input returns to those three resource methods, preserving the distinction between omitted `users` (all workspace members) and `users: []` (none), then remove the redundant caller guard.

philipperolet: Removed in efba0aa570 as requested, ahead of the separate callee optimization. I verified installed Sequelize 6.37.8 by intercepting query dispatch: findAll with an empty id array still dispatches SELECT ... WHERE id IN (NULL). The remaining path returns the same empty editor lists; the editor tests pass.

tdraier:
LGTM ✅

PR #31892, opened 2026-09-04
[front] fix: use workspace member total for active user percentages

aubin-tchoi:
## Description

Follow up on https://github.com/dust-tt/dust/pull/31887#discussion_r3934810655

The percentage of active users shown in the hover card of the consumption chart should use the workspace's absolute member count as its denominator instead of the active user count scoped to the period.

This PR fixes that.

Note: in `countActiveMembersForWorkspace`, "active" means the user has an active membership, not that they were active in the same sense that we're using everywhere else.

## Tests

- Updated existing tests.
- Tested locally.

## Risk

- Low.

## Deploy Plan

- Deploy front.

### DIFF
--- a/front/lib/api/analytics/consumption/timeseries.ts
+++ b/front/lib/api/analytics/consumption/timeseries.ts
@@ -148,9 +148,10 @@ export async function fetchConsumptionTimeseries(
     return timeseriesResult;
   }
 
-  const workspace = auth.getNonNullableWorkspace();
   const workspaceMemberCount =
-    await MembershipResource.countActiveMembersForWorkspace({ workspace });
+    await MembershipResource.countActiveMembersForWorkspace({
+      workspace: auth.getNonNullableWorkspace(),
+    });
 
   return new Ok({
     ...timeseriesResult.value,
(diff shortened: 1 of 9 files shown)


front/lib/api/analytics/consumption/timeseries.ts
```diff
@@ -127,22 +134,36 @@ export async function fetchConsumptionTimeseries(
...
 
-  return fetchTimeseriesBreakdown(auth, query, scope, {
-    breakdownBy,
-    breakdownCount,
+  if (timeseriesResult.isErr()) {
+    return timeseriesResult;
+  }
+
+  const workspaceMemberCount =
+    await MembershipResource.countActiveMembersForWorkspace({
```
avervaet: fetchConsumptionTimeseries is shared by the manager-only workspace route, the member-accessible /me route, and the agent analytics route. So rn we are leaking workspace level info (not the most critical one i agree).

We could pass an includeWorkspaceContext flag to avoid leaking workspace level infos here. 

Same approach as for fetchConsumptionOverview ?

avervaet (approved):
Pre-approving all good once the comment is addressed

front/lib/api/analytics/consumption/timeseries.ts
aubin-tchoi: hmm thought you had various easy ways to have the total count but maybe not

aubin-tchoi: you're right, I'll filter out just in case

PR #31895, opened 2026-09-04
[front] plan mode: move the plan card into the title bar

PopDaph:
## Description

Goal: make the plan mode pretty so we can ship 💅🏻 
Follow-up to #31873 and #31874.

The plan card above the composer is gone. The title bar gets a `Plan 4/5` chip next to Credits and Files that toggles the plan panel, like in the design prototype. The three of them are `FilterChip variant="secondary"` (#31903), selected while their panel is open.

The chip stayed selected during the whole closing animation because `currentPanel` only clears once the panel has collapsed. The side panel context now exposes `isPanelClosing` and the chips use it to unselect right away.

The card was the only way to close a plan, so the trash moved to the plan panel header. The auto open/close logic moved to the chip as is, keyed by conversation so opening a conversation that already has a plan no longer pops the panel.

## Tests

Locally, and on a review app https://pr-31895-merge--app.preview.dust.tt/w/0ec9852c2f/conversation/HE3JRv9LzO#?spid=plan&spt=plan

## Deployments

Deploy front.

### DIFF
--- a/front/components/assistant/UserMessageMarkdown.integration.test.tsx
+++ b/front/components/assistant/UserMessageMarkdown.integration.test.tsx
@@ -61,6 +61,7 @@ vi.mock(
     ...(await importOriginal()),
     useConversationSidePanelContext: () => ({
       currentPanel: undefined,
+      isPanelClosing: false,
       openPanel: vi.fn(),
       togglePanel: togglePanelMock,
       closePanel: vi.fn(),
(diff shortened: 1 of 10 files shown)


smb2268 (approved):
Nice, I think this looks really good :) 

One note that is tangentially related but not a result of these changes - when the agent renders the list and has a checkbox in the markdown, we render the sparkle checkbox and the cursor state makes it look clickable when it isn't. Maybe we update the checkbox component so there's no cursor change or hover state if there's no on click or something?

Image

PopDaph:
Yes fair since the agent markdown is never editable I could also use the "pretty version" on the agent message, will do!

PR #31896, opened 2026-09-04
[front] fix: empty origin seat credit's next-period pre-grant on seat change

tdraier:
## Description

Recurring per-seat AWU credits are materialized **one period ahead**: when period N starts, Metronome creates the credit's N+1 segment and grants the origin seat's allocation to whoever is assigned to it. A mid-period `pro→max` move only empties/carries the **current (N)** segment — both `emptyOriginSeatCreditsForTransfers` and `carryConsumptionToNewSeatCredits` resolve the credit segment covering *now* — so the origin's **N+1 pre-grant survives untouched**. When N+1 begins, the user holds that orphaned `pro` 8000 on top of the `max` 40000 → a fresh, no-usage **48000** stack.

This is the root cause of stacked credits on workspaces whose seat-sync does **not** fail and whose seat assignments are clean (no stray/double assignment): the leak is purely the un-emptied future-period credit segment.

Adds `emptyOriginNextPeriodCredits`, invoked in `syncSeatCount` right after `carryConsumptionToNewSeatCredits`. For each transfer it resolves the origin credit's next segment (`getNextSeatCreditRenewalDate` → `findSeatCreditSegmentForPeriod`), reads the seat's balance at that covering date, and removes it from the origin credit's next segment — capped at the origin allocation so it can never drive the balance negative, which also makes it idempotent (once removed, the next read is 0 → 0 delta). Best-effort per user.

## Tests

Manual root-cause validation on production data (workspace `BEjN1UpVix`, user on `max`, not assigned to `pro`, balance 48000 with zero usage) confirmed the mechanism; the cleanup scripts on PR #31709 already reconcile the *current* period. This PR stops the *future* period from re-stacking. A follow-up unit test around `emptyOriginNextPeriodCredits` in `seats.test.ts` is worth adding.

⚠️ One assumption to verify before/at rollout: that `listMetronomeSeatBalances` / `findSeatCreditSegmentForPeriod` with a **future** `coveringDate` resolve the pre-materialized N+1 segment (they should, since it already exists at sync time). Validate on one workspace in a controlled run.

## Risk

Scoped to the seat-change credit-transfer path (only runs when there are pending transfers). Best-effort and idempotent, capped at the origin allocation so it cannot over-remove or go negative. On rollback, stacks simply resume regenerating each period (the pre-fix behavior); the cleanup scripts remain the mitigation.

## Deploy Plan

Deploy with `front`. After deploy, re-run the cleanup audit/fix (PR #31709) once more to clear the accumulated backlog; going forward, new `pro→max` moves should no longer leave a next-period stack.

### DIFF
--- a/front/lib/metronome/seats.ts
+++ b/front/lib/metronome/seats.ts
@@ -1405,6 +1405,149 @@ async function carryConsumptionToNewSeatCredits({
   }
 }
 
+/**
+ * Reclaim the origin seat credit's NEXT-period pre-grant after a seat change.
+ *
+ * Recurring per-seat credits are materialized one period AHEAD: when period N
+ * starts, Metronome creates the credit's N+1 segment and grants the origin
+ * seat's allocation to whoever is on it. A mid-period origin→new move only
+ * empties/carries the CURRENT (N) segment (`emptyOriginSeatCreditsForTransfers`
+ * / `carryConsumptionToNewSeatCredits` both read the segment covering "now"), so
+ * the origin's N+1 pre-grant survives untouched. When N+1 begins, the user holds
+ * that orphaned origin allocation ON TOP of the new seat's allocation — a fresh,
+ * no-usage stack (e.g. pro 8000 + max 40000) that needs no sync failure and no
+ * stray seat assignment to appear. This empties that pre-grant too.
+ *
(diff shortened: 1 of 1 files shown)


tdraier:

- #31896 ◀

PR #31899, opened 2026-09-04
admin governance: end space memberships instead of suspending them

fabiencelier:
## Description

First of 4 PRs merging the two ways a space gets its members: the manual member list and the attached groups (provisioned or manual). Today `vaults.managementMode` makes them exclusive.

Switching a space to group mode suspends its manual memberships (`status = 'suspended'`, kept on record but inert) so switching back can restore them. The merged model has nowhere to hide an inert member list, and restoring one would silently re-grant access to people an admin dropped by switching to groups.

So: end those memberships instead of suspending them — the same mutation as removing every member (`endAt` in the past). Plus the backfill that ends the rows already suspended in prod, to run after this deploys. Once it has, no membership row is suspended-and-open any more, and nothing can create one, which is what lets the next PRs drop the mode without any chance of resurrecting access.

- `GroupResource.dangerouslySuspendMembers` → `dangerouslyEndAllMemberships`
- `SpaceResource.suspendManualGroupMembers` → `endManualGroupMembers`
- `dangerouslyRestoreMembers` / `restoreManualGroupMembers` are kept on purpose: until the backfill runs, a group → manual switch must still restore what the old behaviour suspended. They go away with the `status` column.

The backfill is scoped to the `regular_auto` groups holding a `space` grant, i.e. exactly the groups a mode switch could suspend. Agent-editor and skill-editor memberships were suspended by their own migrations and have their own restore backfills; ending them here would make them unrestorable.

User-visible: switching a space from manual to group access now drops its member list for good. The confirm dialog says so.

## Tests

 - updated unit tests

## Risk

Medium. Permission-adjacent, but it only makes access strictly narrower, never wider, and the backfill grants nothing: suspended memberships are already excluded from every active-membership query, so ending them changes no one's access — it moves the "inert" state from `status` onto `endAt`.

Not reversible for the spaces it touches: a space switching to group mode loses its manual member list, where before it could be restored. That is the intended product change.

## Deploy Plan

- Deploy front
- Run the backfill: `front/migrations/20260904_end_suspended_group_memberships.ts` (dry run
  first, then `--execute`)

Check afterwards, expected to stay at 0:

```sql
SELECT count(*) FROM group_memberships WHERE status = 'suspended' AND "endAt" IS NULL;
```

### DIFF
--- a/front/components/spaces/RestrictedAccessBody.tsx
+++ b/front/components/spaces/RestrictedAccessBody.tsx
@@ -78,7 +78,8 @@ export function RestrictedAccessBody({
         title: "Switch to groups",
         message:
           "This switches from manual member to group-based access. " +
-          "Your current member list will be saved but no longer active.",
+          "Your current member list will be removed; you'll need to re-add " +
+          "members to switch back.",
         validateLabel: "Confirm",
         validateVariant: "primary",
       });
--- a/front/components/spaces/CreateOrEditSpaceModal.tsx
+++ b/front/components/spaces/CreateOrEditSpaceModal.tsx
@@ -90,7 +90,7 @@ export function CreateOrEditSpaceModal({
   const { spaceInfo, mutateSpaceInfo, isSpaceInfoLoading } = useSpaceInfo({
     workspaceId: owner.sId,
     spaceId: space?.sId ?? null,
-    includeAllMembers: true, // Always include all members so we can see suspended ones when switching modes
+    includeAllMembers: true, // Include members whose membership is not active yet.
   });
 
   const { groups } = useGroups({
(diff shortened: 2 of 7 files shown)


front/components/spaces/RestrictedAccessBody.tsx
```diff
@@ -78,7 +78,8 @@ export function RestrictedAccessBody({
         title: "Switch to groups",
         message:
           "This switches from manual member to group-based access. " +
-          "Your current member list will be saved but no longer active.",
+          "Your current member list will be removed; you'll need to re-add " +
+          "members to switch back.",
```
fabiencelier, line 82: this is temporary, will be dropped when we can maange both at the same time

tdraier (approved):
lgtm

PR #31900, opened 2026-09-04
[skills] take out reference from the input

ykmsd:
## Description
This PR is to take out skill's references from the input, and add an option to show more/less. Show more/less button only appears when the content height when we show three lines of chips.

CleanShot 2026-09-07 at 12 35 20@2x

### DIFF
--- a/front/components/skill_builder/SkillBuilderInstructionsEditor.tsx
+++ b/front/components/skill_builder/SkillBuilderInstructionsEditor.tsx
@@ -222,10 +222,6 @@ function sanitizeSkillInstructionsHtml(html: string): string {
 }
 
 const INSTRUCTIONS_EDITOR_SIZE = "min-h-60 max-h-[50vh]";
-const INSTRUCTIONS_EDITOR_REFERENCE_SUMMARY_CONTAINER_SIZE =
-  "h-80 min-h-80 max-h-[50vh]";
-const INSTRUCTIONS_EDITOR_REFERENCE_SUMMARY_SIZE =
-  "h-full min-h-0 max-h-none resize-none rounded-b-none border-b-0 pb-44";
 
 interface SkillBuilderInstructionsEditorProps {
   onAddKnowledge?: (addKnowledge: () => void) => void;
@@ -736,8 +732,6 @@ export function SkillBuilderInstructionsEditor({
               readOnly: isInstructionsReadOnly,
             }),
             INSTRUCTIONS_EDITOR_SIZE,
-            hasInstructionReferenceSummary &&
-              INSTRUCTIONS_EDITOR_REFERENCE_SUMMARY_SIZE
           ),
         },
       },
@@ -840,8 +834,6 @@ export function SkillBuilderInstructionsEditor({
         <div
           className={cn(
             "group relative overflow-hidden rounded-xl",
-            hasInstructionReferenceSummary &&
-              INSTRUCTIONS_EDITOR_REFERENCE_SUMMARY_CONTAINER_SIZE,
             hasInstructionReferenceSummary && !isDiffMode && "resize-y"
           )}
         >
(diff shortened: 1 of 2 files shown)


front/components/skill_builder/SkillBuilderInstructionsEditor.tsx
```diff
@@ -735,21 +724,12 @@ export function SkillBuilderInstructionsEditor({
...
     });
-  }, [
-    editor,
-    displayError,
-    isDiffMode,
-    isInstructionsReadOnly,
-    isReadOnly,
-    hasInstructionReferenceSummary,
-  ]);
+  }, [editor, displayError, isDiffMode, isInstructionsReadOnly, isReadOnly]);
```
ykmsd, line 732: seems to be fp 🤔

front/components/skill_builder/SkillBuilderInstructionsReferenceSummary.tsx
```diff
@@ -202,32 +221,58 @@ export function SkillBuilderInstructionsReferenceSummary({
...
+      getComputedStyle(document.documentElement).fontSize
+    );
+
+    if (rootFontSize !== BROWSER_DEFAULT_FONT_SIZE) {
+      setOverflowThresholdHeight(rootFontSize * THRESHOLD_HEIGHT_REM);
+    }
+  }, []);
+
+  // biome-ignore lint/correctness/useExhaustiveDependencies: referenceItems.length triggers re-measurement
+  useEffect(() => {
```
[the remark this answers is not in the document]
ykmsd: we cannot write it inside the event handler since the dependencies are computed values from somewhere else

front/components/skill_builder/SkillBuilderInstructionsReferenceSummary.tsx
```diff
@@ -109,14 +116,24 @@ function renderReferenceSummaryItem({
   }
 }
 
+const BROWSER_DEFAULT_FONT_SIZE = 16;
+const THRESHOLD_HEIGHT_REM = 3.5; // max-h-14 => 3.5 rem
```
ykmsd, line 120: not the best but not sure what else I can do 🥲

front/components/skill_builder/SkillBuilderInstructionsReferenceSummary.tsx
```diff
@@ -202,32 +219,52 @@ export function SkillBuilderInstructionsReferenceSummary({
...
+      getComputedStyle(document.documentElement).fontSize
+    );
+
+    if (rootFontSize !== BROWSER_DEFAULT_FONT_SIZE) {
+      setOverflowThresholdHeight(rootFontSize * THRESHOLD_HEIGHT_REM);
+    }
+  }, []);
+
+  // biome-ignore lint/correctness/useExhaustiveDependencies: referenceItems.length triggers re-measurement
+  useEffect(() => {
```
ykmsd, line 234: we cannot write it inside the event handler since the dependencies are computed values from somewhere else

ykmsd: also computation doesn't work neither because we need the element size after painting

front/components/skill_builder/SkillBuilderInstructionsEditor.tsx
```diff
@@ -222,10 +222,6 @@ function sanitizeSkillInstructionsHtml(html: string): string {
 }
 
 const INSTRUCTIONS_EDITOR_SIZE = "min-h-60 max-h-[50vh]";
-const INSTRUCTIONS_EDITOR_REFERENCE_SUMMARY_CONTAINER_SIZE =
```
ykmsd, line 225: removing out from the input so we don't need these logic 🧹

front/components/skill_builder/SkillBuilderInstructionsReferenceSummary.tsx
```diff
@@ -202,32 +219,51 @@ export function SkillBuilderInstructionsReferenceSummary({
     [knowledgeReferences, skillReferences, toolReferences]
   );
 
+  // compute root font size only once
```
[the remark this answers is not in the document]
ykmsd: I thought about it but we don't wanna trigger getComputedStyle many times and I think it's okay to have two useEffects (not ideal but okayish)

smb2268 (approved):
LGTM!

PR #31901, opened 2026-09-04
Poke: paginate the Frames v2 list server-side like the other tables

davidebbo:
## Description

The poke Frames v2 table showed two pagination systems at once: a "Load more" button that re-fetched with a growing `limit`, plus the shared data table's own client-side prev/next controls over the loaded rows.

This switches it to the pattern used by the agent conversations and triggers tables: server-side limit/offset pagination with a total count, driven through `serverSideRowCount` on `PokeDataTable`. The "Load more" button is gone.

- `FileResource.listFrameV2ForWorkspacePaginated` now takes `limit`/`offset` and returns `totalCount` (via `findAndCountAll`). The keyset `lastValue` path is removed since nothing used it; `id` breaks `updatedAt` ties so rows cannot drift between pages.
- `listWorkspaceFrames` returns `{ items, totalCount }` instead of `hasMore`/`lastValue`.
- The poke route validates `limit`, `offset`, `orderDirection` and `hasSandbox` with a zod query schema, like the conversations route.
- `usePokeFrames` takes `offset` and exposes `totalCount` and `isValidating`.
- The table holds a `PaginationState`; toggling the sandbox filter resets to the first page.

The endpoint is private and this table is its only consumer, so dropping `hasMore`/`lastValue` breaks no other client.

## Tests

- `front-api` poke frames route tests updated and passing, including a two-page limit/offset assertion.
- Type-checks for `front` and `front-api` pass.

## Risk

Low: poke-only UI.

## Deploy Plan

Deploy front-api and front-spa.

### DIFF
--- a/front-api/routes/poke/workspaces/[wId]/frames/index.test.ts
+++ b/front-api/routes/poke/workspaces/[wId]/frames/index.test.ts
@@ -31,7 +31,7 @@ describe("GET /api/poke/workspaces/:wId/frames", () => {
       functionCount: 1,
       sandboxStatus: null,
     });
-    expect(data.hasMore).toBe(false);
+    expect(data.totalCount).toBe(1);
   });
 
   it("counts only the active publication's functions", async () => {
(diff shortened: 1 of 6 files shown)


fontanierh (commented):
LGTM

PR #31882, opened 2026-09-04
Upg: use CellContext instead of RegionContext except for extension (another PR)

Fraggle:
## Description

I replace browser-facing `RegionContext` with `CellContext` across the front SPA, Poke, OAuth, email flows, MCP/connectors, and browser extensions (later). I add a public `GET /api/cells` catalog, resolve the active cell from URL/storage/current-cell metadata, and propagate `cell` through login, redirects, approval links, and cell-aware API clients. Workspace lookup now relies on WorkOS cell metadata instead of probing other cells; legacy region redirects keep a compatibility fallback.

## Tests

Added `GET /api/cells` front-api coverage and updated email validation and MCP form tests for cell-aware bootstrap and payloads.

## Risk

Medium risk. Cell resolution now gates every SPA entrypoint, and login routing depends on complete WorkOS cell metadata. Missing or stale metadata can misroute users or reject workspace lookup. The change remains rollback-safe before any data cleanup.

## Deploy Plan

Deploy `front-api` first to expose `GET /api/cells`, then deploy `front` and rebuild or publish browser extension artifacts. Confirm WorkOS cell metadata is populated before rollout. No SQL migration or infrastructure change.

PR #31902, opened 2026-09-04
Use cell for workspace redirect lookups

Fraggle:
## Description

Route workspace redirect resolution through `getWorkspaceCellRedirect` and `CellInfo` instead of the legacy region cache. Auth-context, join, Poke, shared-frame, SPA login, and frame metadata handling now use cell-aware redirects and `workspace_in_different_cell`; remove the redundant region lookup, invalidation, registry entry, and warm-up script.

## Tests

Updated `front-api/routes/w/[wId]/join.test.ts` to mock `getWorkspaceCellRedirect`. CI lint and CodeQL checks pass; remaining test and type checks are running.

## Risk

Low-to-medium. Redirect payloads and error types change across `front` and `front-api`; both deploy together to keep producers and consumers aligned. The change is rollback-safe, but reverting restores the legacy region lookup path and cache behavior.

## Deploy Plan

Deploy `front` and `front-api` together. No SQL migration or cache warm-up is required.


Fraggle:

- #31882
    - #31902 ◀

front/lib/api/cells/lookup.ts
```diff
@@ -82,11 +148,241 @@ export async function invalidateWorkspaceCellCache(wId: string): Promise<void> {
...
+  const otherCells = config.getOtherCells();
+
+  const body: UserLookupRequestBodyType = {
+    user: userLookup,
+  };
+
+  try {
+    // eslint-disable-next-line no-restricted-globals
+    for (const cell of otherCells) {
+      const res = await fetch(`${cell.url}/api/lookup/user`, {
```
flvndvd, line 193: Can't we serve this directly from WorkOS?

front/lib/api/cells/lookup.ts
```diff
@@ -82,11 +148,241 @@ export async function invalidateWorkspaceCellCache(wId: string): Promise<void> {
...
+      }
+
+      if (data.exists) {
+        return new Ok(cell);
+      }
+    }
+  } catch (error) {
+    if (error instanceof Error) {
+      return new Err(error);
+    }
```
flvndvd: To avoid falsy return, what about:

```suggestion
    return new Err(normalize(error));
```

flvndvd (approved):

front/lib/api/cells/lookup.ts
Fraggle: Let me think about it. I'm not sure since the user in workOS can be in multiple orgs in multiple cells.

front/lib/api/cells/lookup.ts
Fraggle: Done

PR #31903, opened 2026-09-04
[sparkle] FilterChips: secondary selected variant and standalone FilterChip

PopDaph:
Design feedback on #31895: the filled primary chip is too heavy for the conversation title bar, so `FilterChips` gets a `variant` prop. `primary` is the current filled look and stays the default. `secondary` gives the selected chip the `bg-selected` token instead, the same one NavTabPill uses for its active pill.

The single chip is now exported as `FilterChip`, controlled by the caller (`isSelected`, `onClick`), with an optional `icon`, a `tooltip` for icon-only chips, and `aria-pressed`. `FilterChips` renders it internally, its API is unchanged apart from the new `variant`.

Two stories added: `Secondary` for the row, `SingleChips` for the standalone chip with icons.

Once merged, #31895 switches the Credits / Files / Plan toggles to `FilterChip variant="secondary"`.

<kbd>
Screenshot 2026-09-04 at 18 19 00
</kbd>

### DIFF
--- a/sparkle/src/components/FilterChips.tsx
+++ b/sparkle/src/components/FilterChips.tsx
@@ -1,6 +1,58 @@
+import { cn } from "@sparkle/lib/utils";
 import React, { useCallback, useState } from "react";
-
 import { Button } from "./Button";
+
+export const FILTER_CHIP_VARIANTS = ["primary", "secondary"] as const;
+export type FilterChipVariant = (typeof FILTER_CHIP_VARIANTS)[number];
+
+interface FilterChipProps {
+  /** Chip text; omit it (with an `icon`) for an icon-only chip. */
+  label?: string;
+  /** Leading icon component. */
+  icon?: React.ComponentType<{ className?: string }>;
+  /** Whether this chip is the active one. */
+  isSelected?: boolean;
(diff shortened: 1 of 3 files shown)


sparkle/src/components/FilterChips.tsx
```diff
@@ -1,6 +1,68 @@
...
+  /** Leading icon. */
+  icon?: ButtonProps["icon"];
+  /** Whether this chip is selected. */
+  isSelected?: boolean;
+  /** Selected look: `primary` fills the chip, `secondary` uses the lighter selected background. */
+  variant?: FilterChipVariant;
+  /** Tooltip label; required for icon-only chips. */
+  tooltip?: string;
+  /** Called on click; the caller owns the selection and toggles it. */
+  onClick?: () => void;
```
smb2268: Is it right that this is optional? We have a different `Chip` component that is just a badge so I think this one should likely only be used if it is clickable

smb2268 (approved):
One question but overall LGTM!

sparkle/src/components/FilterChips.tsx
PopDaph: Agreed 👍🏻 
Adding a commit

PR #31905, opened 2026-09-04
[sparkle] Storybook: paint every docs canvas in dark mode

PopDaph:
In dark mode, docs pages only painted the first story canvas black: the theme decorator used `querySelector(".docs-story")`. Every other story on the page kept a white canvas while its content switched to dark colors, so ghost text became invisible (see FilterChips docs, "No Default Selection" and below).

The decorator now paints every `.docs-story` on the page.

Ex here: https://sparkle.dust.tt/?path=/docs/forms-inputs-filterchips--docs&globals=theme:dark


PopDaph:
PMRR 🙏🏻

PR #31906, opened 2026-09-04
[sparkle] Storybook: paint the canvas with the background token

PopDaph:
Follow-up to #31905. Storybook painted the canvas pure black in dark mode, but nothing in the product is black: dark surfaces are stone. Subtle things like a `bg-selected` chip looked much fainter in Storybook than in the app.

The canvas now uses `var(--color-background)` in both themes, so components sit on the surface they ship on. The dark swatch in the toolbar matches, and I removed the `backgrounds` config since that addon isn't loaded.

### DIFF
--- a/sparkle/.storybook/preview.ts
+++ b/sparkle/.storybook/preview.ts
@@ -61,29 +61,9 @@ const preview: Preview = {
     themes: {
       default: "light",
       list: [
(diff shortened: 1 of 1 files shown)


PopDaph:
PMRR 🙏🏻

PR #31917, opened 2026-09-05
[front] enh: sort group attribution by workspace average

aubin-tchoi:
## Description

The Groups attribution tab can now sort the `Vs workspace avg` column in either direction.

The group endpoint ranks this derived per-active-member value server-side before applying pagination. Normal credit sorting keeps the existing bounded metric query; only the explicit workspace-average sort enumerates group buckets so the derived order stays correct across pages.

## Tests

- Added focused coverage for UI sort wiring, API forwarding, and server-side ordering across composite pages.

## Risk

- Low, limited to group attribution sorting.

## Deploy Plan

- Deploy front.

### DIFF
--- a/front/lib/api/analytics/consumption/schema.ts
+++ b/front/lib/api/analytics/consumption/schema.ts
@@ -5,6 +5,7 @@ import {
   CONSUMPTION_METRICS,
   CONSUMPTION_SCOPE_DIMENSIONS,
   CONSUMPTION_SCOPE_FILTER_KEYS,
+  CONSUMPTION_TOP_GROUP_SORT_BY,
   CONSUMPTION_TOP_SORT_ORDER,
   DEFAULT_CONSUMPTION_METRIC,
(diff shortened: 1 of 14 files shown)


front/lib/api/analytics/consumption/schema.ts
```diff
@@ -83,13 +84,21 @@ export const ConsumptionTopBodySchema = ConsumptionBodySchema.extend({
     .transform((limit) => limit ?? DEFAULT_CONSUMPTION_TOP_LIMIT),
   offset: z.number().int().nonnegative().default(0),
   search: z.string().trim().optional(),
-  // Always ranks by gross credits; see the comment on
-  // CONSUMPTION_TOP_SORT_ORDER for why other metrics aren't supported yet.
```
aubin-tchoi, line 87: could not find this comment anywhere in the history

aubin-tchoi:
can't sort in the ES query

PR #31921, opened 2026-09-05
Remove unused v2-stroke icons from Sparkle

Duncid:
The v2-stroke set had 1,203 icons but only 185 were ever imported by a product app (front, front-spa, extension, marketing) or by a Sparkle component. Delete the other 1,018 along with their SVG sources.

Rewrite the index by filtering its export lines rather than running build_icons.sh: that script wipes src/icons/v2-stroke before regenerating from SVGs, which would drop the 19 kept icons that are hand-authored TSX with no SVG source.

Stories and the playground were the only in-repo consumers of the deleted icons, so swap those to the closest kept equivalent. ICONS_USED_IN_PRODUCT was stale (listed 4 unused icons, missed 41 used ones) and now matches the kept set.

This breaks the public @dust-tt/sparkle export surface for consumers importing a deleted icon.


Duncid:
Hey @Yutcam !
This is just removing what seem to be un-used icons.
AI generated though…

Duncid:
Hey @ykmsd !
This PR is removing icons from Untitled that are **supposedly not in use**.
Could you double check it for me please?

ykmsd (approved):

PR #31922, opened 2026-09-05
[front] fix: detach conversation scrolling during streaming

aubin-tchoi:
## Description

Fixes dust-tt/tasks#10209.

Long agent responses currently keep pulling the conversation to the bottom when someone scrolls up to read, which is especially disruptive on mobile because the screen is usually quite narrow, and when you try reading the message you are fighting against the scroll.

This PR makes auto-scroll distinguish streamed content growth from the user's upward viewport movement. If you don't do anything, content keeps following, but scrolling up "detaches" auto-scroll. Reaching the bottom re-attaches.

The main catch here is that Virtuoso resizes the message rows during the streaming: updates the CSS first, then scrolls into it. We unfortunately can't reliably opt-out from this entirely, so what we're doing here is compensating for this resize when detached.

## Tests

- Added tests extensively (checks the detach, the re-attach, the normal scroll, etc..).
- Tested using preview on desktop + mobile.

## Risk

- Low.

## Deploy Plan

- Deploy front.

### DIFF
--- a/front/components/assistant/conversation/AgentInputBar.tsx
+++ b/front/components/assistant/conversation/AgentInputBar.tsx
@@ -122,7 +122,7 @@ export const AgentInputBar = ({ context }: AgentInputBarProps) => {
     workspaceId: context.owner.sId,
   });
   const methods = useVirtuosoMethods<VirtuosoMessage>();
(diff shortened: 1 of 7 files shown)


aubin-tchoi:
🥲

PR #31924, opened 2026-09-06
Show the published Frame name on the shared Frame v2 page

davidebbo:
## Problem

Opening a shared Frame v2 showed "Manifest" as the page header and tab title. The share page derived the title from the file name, and a Frame v2's file is always `manifest.json`.

## Fix

- The share metadata endpoint (`front-api/routes/share/frame/[token].ts`) now returns a display-ready `title`: the published Frame name from `useCaseMetadata.frameName` for a Frame v2, and the formatted file name for legacy Frames (same output as the client-side formatting it replaces).
- `SharedFramePage` uses that title directly and passes it down through `PublicInteractiveContentContainer` to `PublicFrameRenderer`, which now takes a `title` prop instead of re-deriving one from `fileName`.

## Tests

- Added share route tests covering the legacy and Frame v2 titles.
- Updated the container and renderer tests for the new `title` prop.

## Risk

Low. Public share page only; legacy Frame titles render identically.


smb2268 (commented):
LGTM!

PR #31926, opened 2026-09-06
Stream file explorer folders as ZIP archives

davidebbo:
## Description

- Add a storage-backend-neutral folder archive planner with limits of 5,000 entries and 1 GiB of declared uncompressed data.
- Stream ZIP entries through DustFileSystem one file at a time, preserving nested files and explicit empty directories.
- Add `archive=zip` GET support and HEAD preflight support to the existing private canonical-path route.
- Use Archiver because the existing `adm-zip` package buffers source files and the completed ZIP. Archiver streams one source file at a time and avoids holding the completed archive in memory.

This is PR 2 of 3. PR #31925 has merged. PR #31927 adds the explorer UI and browser preflight.

## Tests

- `npm run test -- lib/api/files/folder_archive.test.ts` in `front`
- `NODE_ENV=test npm run test -- canonicalPath` in `front-api`
- `npm run check:bundled-deps` in `front-api`
- `npx tsgo --noEmit` in `front-api`

The route test parses the streamed response as a real ZIP and verifies its entries and contents.

## Risk

The main risk is server resource use for large folders. Planning enforces entry and declared-size limits, and streaming opens only one backing file at a time. No UI calls this endpoint until PR 3 lands.

## Deploy Plan

Deploy normally after PR #31925 and before the explorer UI PR.


front/lib/api/files/folder_archive.ts
```diff
@@ -0,0 +1,236 @@
...
+    return new Err(
+      new FolderArchiveError(
+        "not_directory",
+        "Only folders can be downloaded as ZIP archives."
+      )
+    );
+  }
+
+  const listResult = await fileSystem.list(normalizedFolderPath, {
+    maxFiles: limits.maxEntries + 2,
```
flvndvd, line 93: +2 ?

front/lib/api/files/folder_archive.ts
```diff
@@ -0,0 +1,236 @@
...
+  });
+  if (listResult.isErr()) {
+    return new Err(listResult.error);
+  }
+
+  const isMountRoot = fileSystem
```
flvndvd, line 99: Does this let you export the whole conversation or pod folder?

front-api/routes/w/[wId]/files/path/[...canonicalPath].ts
```diff
@@ -129,6 +136,79 @@ async function resolveFs(
...
 
+function mapFolderArchiveError(error: FolderArchivePlanError) {
+  if (!isFolderArchiveError(error)) {
+    return mapDustFsError(error);
+  }
+
+  const { code } = error;
+  switch (code) {
+    case "not_found":
+      return {
```
Comment (flvndvd) [inline front-api/routes/w/[wId]/files/path/[...canonicalPath].ts:147]: Let's type function's return type to avoid all the `as const` here.

front-api/routes/w/[wId]/files/path/[...canonicalPath].ts
```diff
@@ -129,6 +136,79 @@ async function resolveFs(
...
+    "Cache-Control": "private, no-store",
+    "X-Content-Type-Options": "nosniff",
+  };
+
+  if (headOnly) {
+    return new Response(null, { status: 200, headers });
+  }
+
+  return new Response(
+    readableToReadableStream(streamFolderArchive(dustFs, planResult.value)),
```
Comment (flvndvd) [inline front-api/routes/w/[wId]/files/path/[...canonicalPath].ts:207]: `readableToReadableStream` ignores Web-stream backpressure, so a slow client can queue the entire ~1 GiB ZIP in memory. 

You should use `Readable.toWeb(output)`.

front/lib/api/files/folder_archive.ts
```diff
@@ -0,0 +1,236 @@
...
+    }
+    failed = true;
+    activeReadStream?.destroy();
+    void archive.abort();
+    output.destroy(normalizeError(error));
+  };
+
+  archive.on("warning", fail);
+  archive.on("error", fail);
+  archive.pipe(output);
```
flvndvd, line 195: This guys fires output’s "pipe" event synchronously. This should happens within `readableToReadableStream` boundary, otherwise the server keeps reading GCS and compressing even after the client has disconnected.

front/lib/api/files/folder_archive.ts
```diff
@@ -0,0 +1,236 @@
...
+    }
+  });
+
+  const writeArchive = async () => {
+    for (const directory of plan.directories) {
+      archive.append(Buffer.alloc(0), { name: directory });
+    }
+
+    for (const file of plan.files) {
+      const readResult = await fileSystem.read(file.canonicalPath);
```
flvndvd, line 209: If my read is correct, a cancellation may set failed while this read is pending. If it then resolves with a stream, we append it to an already-aborted archive without destroying it.

front/lib/api/files/folder_archive.ts
davidebbo: Yes, that's deliberate. e.g. http://localhost:3000/api/w/BHWrN3HaUa/files/path/conversation-0m9lsJyVIq?archive=zip

flvndvd: Solide 💪 !

front-api/routes/w/[wId]/files/path/[...canonicalPath].ts
Comment (davidebbo) [reply front-api/routes/w/[wId]/files/path/[...canonicalPath].ts:147]: Good catch. Do you think we have coding rules for this already and they're just getting ignored? This is the type of things that we should expect agents to get right on the first try. 😠

PR #31929, opened 2026-09-06
admin governance: merge a space's manual members and its groups

fabiencelier:
## Description

Second of 4 PRs merging the two ways a space gets its members. Stacked on #31899.

A space's members were either a manual list or the members of the groups attached to it, never both: `vaults.managementMode` picked one, and the other was dropped. They are now merged: a space's members are its manual list **plus** every member of every group attached to it.

`updatePermissions` takes each dimension on its own (`memberIds`, `editorIds`, `groupIds`, `editorGroupIds`): the ones a request carries are overwritten, the ones it leaves out keep their current value. So the UI can keep sending only members, or only groups.

Nothing switches modes any more. 

UI still switch between both mode, when saved it overrides everything

## Tests

 - added new unit tests to cover this

## Risk

Large blast radius: it rewrites the space permission path, which is what every space read is checked against.

## Deploy Plan

- Deploy front

Requires #31899 and its backfill to have run first: this PR removes the guards that kept a suspended member list from coming back.

### DIFF
--- a/front-api/public/swagger.json
+++ b/front-api/public/swagger.json
@@ -10394,8 +10394,7 @@
                 "required": [
                   "isRestricted",
                   "name",
-                  "spaceKind",
-                  "managementMode"
+                  "spaceKind"
                 ],
                 "properties": {
                   "isRestricted": {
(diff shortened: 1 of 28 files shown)


fabiencelier:
Note for review: this PR orphans two now-unreachable helpers — `SpaceResource.endManualGroupMembers` and `GroupResource.dangerouslyEndAllMemberships` — because nothing switches modes any more. They are deleted in #31931 (the cleanup PR of the stack) rather than here, to keep the stack's history stable.

front-api/public/swagger.json
```diff
@@ -10411,26 +10410,27 @@
...
-                    "description": "Required when managementMode is group"
+                    "description": "The groups given access to the space. Left untouched when omitted."
+                  },
+                  "managementMode": {
+                    "type": "string",
+                    "enum": [
+                      "manual",
+                      "group"
+                    ],
+                    "description": "Deprecated. When set, the dimension the mode does not cover is cleared."
```
fabiencelier: to map th UI behaviour until changed

tdraier (approved):
looks reasonable

PR #31932, opened 2026-09-06
[front] feat: empty stacked seat credits on credit.segment.start

tdraier:
## Description

Wires the shared stacked-seat-credit core (added in the parent PR #31709) into the debounced credit-state reconcile so **new** stacked per-seat AWU credits are cleaned automatically, at the source, instead of needing the backfill script.

Why `credit.segment.start`: a per-seat recurring credit is materialized one period ahead, so a mid-period seat change strands a stray grant on the **next** segment that only becomes a live balance when that segment starts. The 4 concurrent segment-start events per contract (2 pro + 2 max) already collapse to one workspace-scoped execution via `WorkflowIdConflictPolicy.USE_EXISTING`, so the new `reconcileStackedSeatCreditsForWorkspace`:

- reads seat balances **once** (with the per-seat `credits[]` breakdown),
- maps each materialized credit id → seat type,
- empties any credit whose type ≠ the seat's current tier and carries its consumption onto the home credit,
- busts the per-user balance caches.

It runs right after the existing per-user reconcile in `reconcileWorkspaceUserCreditStatesActivity`. Idempotent and self-healing: each stray is cleaned when its segment becomes current; a re-run reads strays at 0 and emits nothing.

## Tests

- `npx tsgo --noEmit` clean; `biome check` clean.
- Detection/correction core validated on real workspaces via the backfill script (same code path): exact stray-emptying + consumption carry (e.g. a seat that spent 69 AWU against its stray nets `8000 − 69 = 7931`, not a full 8000).
- Manual verification of the debounce on a seat-type change pending in a test workspace.

## Risk

Low. Additive: a new lib function called after the existing reconcile on an already-debounced, low-frequency path (period boundaries / seat changes). Apply is two-phase (empty strays, then carry only for seats whose strays emptied), so a partial Metronome failure can never double-debit a user — worst case is the pre-existing small over-credit. Manual ledger entries don't fire Metronome webhooks, so there is no reconcile feedback loop.

## Deploy Plan

Ship after #31709 (depends on the `lib/metronome/stacked_seat_credits` core). No migration. Backfill existing strays with the script; this PR prevents new ones going forward.

### DIFF
--- a/front/lib/api/metronome/reconcile_credit_state.ts
+++ b/front/lib/api/metronome/reconcile_credit_state.ts
@@ -17,6 +17,7 @@ import {
 import { fetchPerApiKeyAwuUsage } from "@app/lib/metronome/per_api_key_usage";
 import { fetchPerUserAwuUsage } from "@app/lib/metronome/per_user_usage";
 import type { CachedContract } from "@app/lib/metronome/plan_type";
+import { getActiveContract } from "@app/lib/metronome/plan_type";
 import { getWorkspacePoolAwuBalance } from "@app/lib/metronome/pool_balance";
 import { fetchProgrammaticAwuSpend } from "@app/lib/metronome/programmatic_awu_usage";
 import {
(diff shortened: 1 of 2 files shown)


tdraier:

- #31709
    - #31932 ◀

fabiencelier (approved):
LGTM ✅

PR #31933, opened 2026-09-06
Remove sandbox_functions feature flag

davidebbo:
## Summary

Removes the `sandbox_functions` feature flag and every code path that existed only to serve it — the legacy "Pod Functions" mechanism, superseded by Frames v2.

- Deletes the `sandbox_functions` internal MCP server (publish/unpublish/list/get/call/db_* tools) and its skill, since `isRestricted` on both was tied directly to this flag and becomes permanently unreachable once it's gone.
- Deletes the tool-approval UI cards for publish/unpublish, and the now-orphaned `publish_sandbox_function`/`unpublish_sandbox_function` engine functions (their only caller was the deleted MCP tool).
- Simplifies `withSandboxFunctionInvocationFeature` and `resolveSandboxFunctionWithCapability` to key off `frames_v2` only, dropping the legacy Pod-only invocation path and the frame-share-token capability mechanism — both were flag-gated and only used by the pre-Frames-v2 flow.
- The shared invocation engine that Frames v2 also depends on (Temporal workers, DB models/resources, `/sandbox-functions` and `/sandbox/frames` HTTP routes) is untouched — those are generic infrastructure, not specific to this flag.

## Test plan

- [x] `tsgo --noEmit` passes clean in both `front` and `front-api`
- [x] Ran the full test suites for every touched area (~300+ tests across the invocation routes, SSE events, resources, skills/instructions) — all passing
- [x] Regenerated the MCP server availability snapshot (`npm run test:update-internal-mcp-availability-snapshot`)
- [x] `biome check --write` on all changed files


PopDaph (commented):
Looks reasonable 👍🏻 
(didn't review the full code :D)

PR #31937, opened 2026-09-07
fix(onboarding): make Pro checkout page responsive on mobile

tdraier:
## Description

Fixes the Pro subscription checkout page not rendering correctly on Safari mobile (dust-tt/tasks#10312).

The page in `front/components/pages/onboarding/CheckoutPage.tsx` used a fixed two-pane layout with **no responsive breakpoints**: the root `<main>` was `flex h-screen overflow-hidden` and both panes were locked to `w-1/2` with `p-24` (6rem) padding. On a narrow Safari-mobile portrait viewport each pane was squeezed to 50% width while still carrying that padding, so the Stripe payment iframe overflowed — and because the container was `overflow-hidden`, there was no way to swipe/scroll to reach the payment fields. Landscape "worked" only because the wider viewport left room for the 50% panes.

Changes (desktop `md:` ≥768px stays pixel-identical):
- **Layout**: `<main>` stacks vertically on mobile (`flex min-h-screen flex-col md:flex-row md:overflow-hidden`); both panes are `w-full md:w-1/2` with `p-6 md:p-24`. The desktop-only `pt-[296px]` alignment offset is gated behind `md:`.
- **Mobile scroll**: enable the app's existing `useDocumentScrollMode(isMobile)` mechanism (already used by `AppContentLayout`) so the stacked panes participate in one natural document scroll instead of two cramped per-pane scroll areas. This is what makes `min-h-screen` meaningful — it frees the `html/body/#root` height chain (`height:auto; min-height:100dvh`) so `main` grows with content.
- Dropped the redundant `md:h-screen`: the unlayered global `main { height:100% }` rule (`front/styles/global.css`) wins over the layered Tailwind utility regardless, and already equals the viewport via the `h-full` chain.

## Tests

Type-check (`tsgo --noEmit`) and biome format pass. Verified the DOM chain from `#root` down to this page's `<main>` is context-providers/`<Outlet>` only (no intermediate fixed-height wrapper), so `document-scroll-mode` propagates correctly. Not visually verified on a physical Safari-mobile device — worth a quick manual check on a phone. No automated tests exist for this page and none added (layout-only change).

## Risk

Low. Layout-only change scoped to one component. Desktop rendering is unchanged (all original classes preserved behind `md:`, and `document-scroll-mode` is mobile-only via `useIsMobile`). Trivially rollback-able.

## Deploy Plan

Standard front-spa deploy, no migrations or coordination required.

### DIFF
--- a/front/components/pages/onboarding/CheckoutPage.tsx
+++ b/front/components/pages/onboarding/CheckoutPage.tsx
@@ -467,9 +467,9 @@ export function CheckoutPage() {
   }
 
   return (
-    <main className="flex h-screen overflow-hidden">
+    <main className="flex min-h-screen flex-col md:h-screen md:flex-row md:overflow-hidden">
       {/* Left pane: order summary + coupon */}
-      <div className="flex w-1/2 flex-col gap-14 overflow-y-auto bg-muted-background p-24">
+      <div className="flex w-full flex-col gap-14 overflow-y-auto bg-muted-background p-6 md:w-1/2 md:p-24">
         <div>
           <Icon visual={DustLogoSquare} size="lg" />
         </div>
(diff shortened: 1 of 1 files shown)


tdraier:

- #31937 ◀

front/components/pages/onboarding/CheckoutPage.tsx
```diff
@@ -467,9 +467,9 @@ export function CheckoutPage() {
   }
 
   return (
-    <main className="flex h-screen overflow-hidden">
+    <main className="flex min-h-screen flex-col md:h-screen md:flex-row md:overflow-hidden">
```
ykmsd: h-screen doesn't work and we need min-h-screen for mobile? 🤔

ykmsd (approved):

front/components/pages/onboarding/CheckoutPage.tsx
tdraier: actually h-screen is inert in all cases ( height is defined on main in global.css , and has precedence over this css rule ), removing. Also adding useDocumentScrollMode()

PR #31938, opened 2026-09-07
[front] plan mode: confirm before archiving a plan

PopDaph:
The plan panel header has a trash icon labeled "Close plan" right next to the X that closes the panel. Easy to mix up, and archiving can't be undone from the UI.

The action is now "Archive plan" with the archive icon (that's what the endpoint does, it moves `plan.md` to `archived_plans/`), and it asks for confirmation first, using the shared `ConfirmContext` dialog like file deletion does. The dialog says the file stays available in the conversation's archived files (wording from uxWriter). The error toast says "archive" too.

Follow-up to #31895.

### DIFF
--- a/front/components/assistant/conversation/plan_mode/ConversationPlanModePanel.tsx
+++ b/front/components/assistant/conversation/plan_mode/ConversationPlanModePanel.tsx
@@ -38,7 +38,7 @@ export function ConversationPlanModePanel({
       message:
         "This plan will be moved to the conversation's archived files. You can still open it later, and the agent can create a new plan if needed.",
       validateLabel: "Archive plan",
-      validateVariant: "warning",
+      validateVariant: "primary",
     });
     if (confirmed) {
       await closePlan();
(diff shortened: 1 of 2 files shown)


front/components/assistant/conversation/plan_mode/ConversationPlanModePanel.tsx
```diff
@@ -27,6 +29,21 @@ export function ConversationPlanModePanel({
...
+  const confirm = useContext(ConfirmContext);
+
+  // Sits next to the panel close button, so ask before archiving.
+  const archivePlan = async () => {
+    const confirmed = await confirm({
+      title: "Archive this plan?",
+      message:
+        "This will hide the current plan from the side panel, but keep it in the conversation's files. The agent can create a new one.",
+      validateLabel: "Archive plan",
+      validateVariant: "primary",
```
PopDaph, line 41: I decided against warning since this is not really a destructive action, we do not loose the plan.

adrsimon (approved):
LGTM

PR #31940, opened 2026-09-07
[front] fix: keep SWR mutation keys when disabled

aubin-tchoi:
## Description

We have a few `mutate` that actually do nothing. In some SWR hooks we sometimes pass `null` as a key to `useSWRWithDefaults` when disabled, in this case our implementation of `useSWRWithDefaults` makes it impossible to identify the cache entry and the mutate becomes a no-op.

This PR fixes that by aligning on always passing the key + disabled as a config. This allows `mutate` and `mutateRegardlessOfQueryParams` to target mounted cache entries without starting requests.

We currently sometimes do it and sometimes not, my guess is that we sometimes do this because agents are not always aware of the fact that we override a bunch of things in `useSWRWithDefaults` and rely on the more classic SWR patterns.

## Tests

- Added tests.

## Risk

- Low.

## Deploy Plan

- Deploy front.

### DIFF
--- a/front/lib/swr/data_source_view_tables.ts
+++ b/front/lib/swr/data_source_view_tables.ts
@@ -92,11 +92,9 @@ export function useDataSourceViewTables({
   const tablesFetcher: Fetcher<
     ListTablesResponseBody | SearchTablesResponseBody
   > = fetcher;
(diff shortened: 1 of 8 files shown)


ykmsd (approved):
> We have a few mutate that actually do nothing

😱

PR #31943, opened 2026-09-07
Don't fail loudly on unattributed Slack Metronome usage

avervaet:
## Description

#31720, added "Fail loudly on unattributed user usage in Metronome events" behavior. This allows us to catch cases where Slack users could talk with Dust even if we can't reconcile them as a Dust user. We want to keep this behavior but record their usage as programmatic rather than user_usage with no user.

This branch fixes usage attribution for Slack messages that Metronome can't tie to a real Dust user, so they're billed as programmatic usage instead of silently falling through as unattributed/free.

**Core change** : isProgrammaticUsageFromContext ([common.ts](vscode-webview://1kbbslh48ss885vmbv2sfm0unevqnu9843dvuctk7ubendum6kcf/front/lib/api/programmatic_usage/common.ts)) gains a fallback rule: if a message's userId is null and its auth method is system_api_key on a Slack-originated message, it's now classified as programmatic usage. Previously this only checked the caller's auth method and the message origin's static classification.

**Plumbing to support it**:  the triggering user's userId and messageAuthMethod are threaded through everywhere usage type gets computed:

LLMTraceContext ([types.ts](vscode-webview://1kbbslh48ss885vmbv2sfm0unevqnu9843dvuctk7ubendum6kcf/front/lib/api/llm/traces/types.ts)) now carries userId (nullable) and userMessageAuthMethod.
runModel ([run_model.ts](vscode-webview://1kbbslh48ss885vmbv2sfm0unevqnu9843dvuctk7ubendum6kcf/front/temporal/agent_loop/lib/run_model.ts)) resolves userId from the user message first, falling back to the authenticator's user, explicitly null if neither resolves.
ConversationResource ([conversation_resource.ts](vscode-webview://1kbbslh48ss885vmbv2sfm0unevqnu9843dvuctk7ubendum6kcf/front/lib/resources/conversation_resource.ts)) now also fetches the parent user message's associated UserModel to get triggeringUserId/triggeringUserMessageAuthMethod.
credit_cost.ts, llm.ts, and usage_queue/activities.ts all pass these two new fields into isProgrammaticUsage/isProgrammaticUsageFromContext.

**Analytics-side mirror**: getProgrammaticUsageFilterClause in the same file gets a matching Elasticsearch clause (Slack origin + system_api_key auth + user_id: "unknown") so query-time programmatic/user splitting stays consistent with the emission-time rule.

## Tests

Added/updated in common.test.ts and events.test.ts to cover the new fallback path.

## Risk

TBD

## Deploy Plan

- Deploy front

### DIFF
--- a/front/lib/api/assistant/credit_cost.ts
+++ b/front/lib/api/assistant/credit_cost.ts
@@ -5,6 +5,10 @@ import { makeFairUseAwuCreditsRateLimitKeyForUser } from "@app/lib/api/assistant
 import { recordProgrammaticSpendLimitUsage } from "@app/lib/api/credits/programmatic_usage_limit";
 import { recordApiKeySpendLimitUsage } from "@app/lib/api/keys/spend_limit";
 import { PostHogServerSideTracking } from "@app/lib/api/posthog";
+import {
+  getUsageType,
+  resolveUsageTypeForAttribution,
+} from "@app/lib/api/programmatic_usage/common";
 import { isProgrammaticUsage } from "@app/lib/api/programmatic_usage/tracking";
 import {
   recordFreeSeatLifetimeUsage,
@@ -20,7 +24,6 @@ import {
   microCreditsToCredits,
   roundCreditsToMicroCredits,
 } from "@app/lib/credits/units";
-import { getUsageType } from "@app/lib/metronome/events";
 import { AgentMCPActionResource } from "@app/lib/resources/agent_mcp_action_resource";
 import { ConversationResource } from "@app/lib/resources/conversation_resource";
 import { MembershipResource } from "@app/lib/resources/membership_resource";
@@ -113,6 +116,8 @@ export async function computeAndStoreAgentMessageCredits(
     status,
     runIds,
     triggeringUserMessageOrigin,
+    triggeringUserId,
+    triggeringUserMessageAuthMethod,
     previousCostCredits,
   } = creditContext;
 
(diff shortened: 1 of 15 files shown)


front/lib/metronome/events.ts
```diff
@@ -58,6 +58,24 @@ export function getUsageType(
...
+): UsageType {
+  if (
+    usageType === USAGE_TYPE_USER &&
+    !userId &&
+    PROGRAMMATIC_FALLBACK_ORIGINS.has(origin)
+  ) {
+    return USAGE_TYPE_PROGRAMMATIC;
+  }
+  return usageType;
+}
```
tdraier: this must be done directly in isProgrammaticUsage , otherwise metronome will be inconsistent with all other accounting ( especially ES )

front/lib/api/assistant/credit_cost.ts
```diff
@@ -141,9 +146,16 @@ export async function computeAndStoreAgentMessageCredits(
...
+      getUsageType(
+        isProgrammaticUsage(auth, { userMessageOrigin: messageOrigin }),
+        messageOrigin
+      ),
+      {
+        userId: triggeringUserId,
+        origin: messageOrigin,
+        authMethod: triggeringUserMessageAuthMethod,
+      }
     ),
```
tdraier: I don't understand the need of adding `resolveUsageTypeForAttribution` on top of `isProgrammaticUsage` ? All calls to isProgrammaticUsage() should return true for slack / no user / system key ?

avervaet:

- #31943 ◀
    - #31970

front/lib/api/programmatic_usage/common.ts
```diff
@@ -111,6 +133,19 @@ export function getProgrammaticUsageFilterClause(): estypes.QueryDslQueryContain
...
+                terms: {
+                  context_origin: [...PROGRAMMATIC_FALLBACK_ORIGINS],
+                },
+              },
+              { term: { auth_method: SYSTEM_API_KEY_AUTH_METHOD } },
+              { term: { user_id: "unknown" } },
+            ],
+          },
+        },
       ],
```
tdraier, line 156: ooc, but this is only for the "old" index we're removing right ? new index contains the isProgrammatic value ?

tdraier (approved):

PR #31825, opened 2026-09-03
[Admin Governance] Backfill agent editor grants

philipperolet:
## Description

Related to https://github.com/dust-tt/tasks/issues/9481

Context: [Admin Governance design doc](https://app.notion.com/p/dust-tt/Design-Doc-Admin-Governance-38a28599d941817292d0e94f8dfafe48)

Follows https://github.com/dust-tt/dust/pull/31810, which keeps legacy and grant-backed editor removals in sync.

Adds an idempotent, dry-run-by-default migration that makes each custom agent grant editor set match its active legacy editor set, including archived agents. It uses the latest non-draft configuration so stale groups linked only to obsolete versions are ignored. Additions and removals go through AgentResource in one transaction, preserving grant locking, membership lifecycle, cleanup, and cache invalidation. The script reports any remaining editor-set differences after execution.

The all-workspace path first selects the workspaces containing non-draft agents. On the US production read replica, the resulting 15.9k-workspace predicate planned in 20 ms and executed in 594 ms; this avoids constructing auth for roughly 457k workspaces.

## Risks

Blast radius: Manually executing the migration adds and removes grant-backed editor memberships for existing custom agents so they match the legacy source of truth. Runtime permission reads remain unchanged.
Risk: standard

## Deploy Plan

- deploy front
- run the migration without --execute and inspect planned additions, removals, and editor differences
- run the migration with --execute, optionally staged by workspace
- rerun without --execute and verify there are no editor differences

## Tests

- [x] NODE_ENV=test npm test -- migrations/20260903_backfill_agent_editor_grants.test.ts lib/resources/agent_resource.test.ts
- [x] NODE_OPTIONS="--max-old-space-size=8192" npx tsgo --noEmit
- [x] npx biome check . (10,258 files)

PR #31947, opened 2026-09-07
[Admin Governance] Fix editor backfill connection pool starvation

philipperolet:
## Description

Follows https://github.com/dust-tt/dust/pull/31825.

The all-workspace editor-grant backfill can time out acquiring a database connection. Transactions hold one connection, but some group-member reads omit the transaction and request another. Concurrent agents can therefore exhaust the 10-connection pool while waiting for additional connections.

Pass the transaction through these reads and reduce concurrency from 4 workspaces × 8 agents to **2 × 8**, preserving throughput within large workspaces without increasing the pool or its timeout.

## Risks

Blast radius: group membership reads/additions/removals inside transactions and the editor-grant backfill.
Risk: low — existing queries now use their caller's transaction; membership semantics are unchanged.

## Deploy Plan

- Deploy the updated front code before rerunning the backfill.
- Rerun `migrations/20260903_backfill_agent_editor_grants.ts --execute`; already matching agents need no writes.

## Tests

- [x] Regression test with automatic transaction propagation disabled: grant, read members, revoke, and remove the empty group. Confirmed it fails before the fix and passes after.
- [x] Backfill, group-permission, group, and user resource suites: 119 tests pass.

### DIFF
--- a/front/migrations/20260903_backfill_agent_editor_grants.ts
+++ b/front/migrations/20260903_backfill_agent_editor_grants.ts
@@ -16,7 +16,7 @@ import assert from "assert";
 import { col, fn, Op } from "sequelize";
 
 const AGENT_CONCURRENCY = 8;
-const WORKSPACE_CONCURRENCY = 4;
+const WORKSPACE_CONCURRENCY = 2;
 
 const AgentConfigModel: ModelStaticWorkspaceAware<AgentConfigurationModel> =
   AgentConfigurationModel;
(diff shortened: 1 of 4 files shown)


front/migrations/20260903_backfill_agent_editor_grants.ts
```diff
@@ -16,7 +16,7 @@ import assert from "assert";
 import { col, fn, Op } from "sequelize";
 
 const AGENT_CONCURRENCY = 8;
-const WORKSPACE_CONCURRENCY = 4;
```
fabiencelier, line 19: could be a good idea to make these as parameters of the script so you don't need another PR to adjust

fabiencelier (commented):
LGTM ✅

PR #31949, opened 2026-09-07
Invitations now uses cell.

Fraggle:
## Description

`PendingInvitationOption` now carries the target `cell` instead of a `region`. Cross-cell invitation lookup preserves `cell.name`, and onboarding passes that exact cell to `handleInvitationSelection`, removing the region-based fallback.

## Tests

Not run; frontend-only type and routing update.

## Risk

Low. The change routes invitations to the cell returned by the lookup instead of inferring it from the region. Rollback is a single frontend revert.

## Deploy Plan

Deploy `front`.


Fraggle:

- #31949 ◀

flvndvd (approved):

PR #31950, opened 2026-09-07
Add script to copy Pod database state to Frames

davidebbo:
## Description

Adds a dry-run-by-default script that mirrors Litestream database replicas from a Pod GCS state prefix into a Frame database state prefix.

The execute path copies every source object before deleting destination-only objects. The Pod source remains unchanged.

## Tests

Not run per request.

## Risk

The script deletes Frame database objects that do not exist in the Pod source when run with `--execute`. It rejects empty or invalid Pod state before mutation and requires the Pod and Frame sandboxes to stop writing state during the copy.

## Deploy Plan

Merge the script. Run it manually with the workspace, Pod, and Frame identifiers, first as a dry run and then with `--execute`.

### DIFF
--- a/front/scripts/copy_pod_databases_to_frame.ts
+++ b/front/scripts/copy_pod_databases_to_frame.ts
@@ -1,0 +1,261 @@
+/**
+ * Mirror one Pod's Litestream replicas into one Frame's database state.
+ *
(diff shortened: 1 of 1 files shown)


Fraggle (commented):
LGTM

PR #31953, opened 2026-09-07
refact: consumption analytics: propagate and log error

sfriquet:
## Description

Propagate errors and log it once if we can't create agent message consumption analytics document, so the assert has hopefully more details about why we couldn't generate the documents.

## Tests

existing tests

## Risk

Low

## Deploy Plan

Front only


aubin-tchoi (approved):
LGTM

flvndvd (approved):

PR #31956, opened 2026-09-07
fix(webhook-sources): block save until provider connection is ready

adrsimon:
## Description

The webhook source creation modal let users click Save before authenticating with the provider. 
The readiness flag (`isPresetReadyToSubmit`) defaulted to `true` regardless of whether a provider was selected, and nothing set it to `false` while the OAuth connection step hadn't completed yet.

Now the flag starts `false` whenever a provider is selected (only the "custom"/no-provider path, which has no OAuth step, starts ready).

## Tests

Nothing ^^

## Risk

Low.

## Deploy Plan

Deploy front.


aubin-tchoi (approved):
LGTM

PR #31961, opened 2026-09-07
fix(useAgentFromSearchParam): remove param stripping on new conversation

achilleburah:
## Description

Gh issue: https://github.com/dust-tt/tasks/issues/10317

### Context
`?agent=` parameter was getting stripped from the URL as soon as the composer picked it up, so
`/w/{wId}/conversation/new?agent={sId}` could never be copied out of the address bar.

Removing the strip alone was not enough though. Switching agent from the composer never updated the URL, while starting a conversation from the details sheet or Cmd+K did, so the URL only told the truth depending on the last entry point used. And pushing an `?agent=` already in the URL was a no-op the hook could not see.

### Solution
No more strip, and `useAgentFromSearchParam` now syncs both ways on the new conversation page. The URL wins on arrival, the picker wins afterwards and is mirrored back into `?agent=`.

## Tests
- Tested locally.
- New unit tests in [useAgentFromSearchParam.test.ts](https://github.com/dust-tt/dust/pull/31961/changes#diff-8b8c23cc8baa7ba0cc2670f1b655dbdbcd84306299c60a1ba3fe73279f6dc645):
  - selects the URL agent on arrival and leaves the URL alone
  - does not write the URL while the URL agent is still loading
  - mirrors a picker change to the URL without re-selecting
  - applies a new URL agent over the current selection
  - writes the URL when an agent is selected and the param is absent
  - does not touch the URL on an existing conversation

Before: 

https://github.com/user-attachments/assets/265fae00-1336-4583-a0f5-4990479003b9

After:
https://github.com/user-attachments/assets/8c7c2711-9328-4639-8278-ae636a694bdc

## Risk

Low. 

## Deploy Plan

Standard deploy.

### DIFF
--- a/front/hooks/useAgentFromSearchParam.test.ts
+++ b/front/hooks/useAgentFromSearchParam.test.ts
@@ -1,0 +1,109 @@
+import { useAgentFromSearchParam } from "@app/hooks/useAgentFromSearchParam";
+import type { LightAgentConfigurationType } from "@app/types/assistant/agent";
+import { renderHook, waitFor } from "@testing-library/react";
+import { beforeEach, describe, expect, it, vi } from "vitest";
+
+const { agentConfigurationHolder, searchParamHolder, setSelectedAgent } =
+  vi.hoisted(() => ({
+    agentConfigurationHolder: {
+      current: null as LightAgentConfigurationType | null,
+    },
+    searchParamHolder: { current: null as string | null },
+    setSelectedAgent: vi.fn(),
+  }));
+
+vi.mock("@app/lib/platform", () => ({
+  useSearchParam: () => searchParamHolder.current,
+}));
+
+vi.mock("@app/lib/swr/assistants", () => ({
+  useAgentConfiguration: () => ({
+    agentConfiguration: agentConfigurationHolder.current,
+  }),
+}));
+
+vi.mock(
+  "@app/components/assistant/conversation/input_bar/InputBarContext",
+  async () => {
+    const { createContext } = await import("react");
+    return { InputBarContext: createContext({ setSelectedAgent }) };
+  }
+);
+
+function makeAgentConfiguration(sId: string): LightAgentConfigurationType {
+  return {
+    id: 1,
+    versionCreatedAt: null,
+    sId,
+    version: 1,
+    versionAuthorId: null,
+    instructions: null,
+    model: {
+      providerId: "openai",
+      modelId: "gpt-4o",
+      temperature: 0.7,
+    },
+    status: "active",
+    scope: "visible",
+    userFavorite: false,
+    name: `agent-${sId}`,
+    description: "desc",
+    pictureUrl: "https://example.com/p.png",
+    maxStepsPerRun: 8,
+    tags: [],
+    templateId: null,
+    requestedGroupIds: [],
+    requestedSpaceIds: [],
+    canRead: true,
+    canEdit: true,
+  };
+}
+
+describe("useAgentFromSearchParam", () => {
+  beforeEach(() => {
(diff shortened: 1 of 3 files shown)


front/hooks/useAgentFromSearchParam.test.ts
```diff
@@ -0,0 +1,90 @@
+import { useAgentFromSearchParam } from "@app/hooks/useAgentFromSearchParam";
+import type { LightAgentConfigurationType } from "@app/types/assistant/agent";
+import { renderHook, waitFor } from "@testing-library/react";
+import { beforeEach, describe, expect, it, vi } from "vitest";
+
+const { agentConfigurationHolder, searchParamHolder, setSelectedAgent } =
+  vi.hoisted(() => ({
+    agentConfigurationHolder: {
+      current: null as LightAgentConfigurationType | null,
```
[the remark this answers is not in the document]
achilleburah: Fixed in 78eaf56. Went with a fully typed fixture rather than `satisfies Partial<...> as ...`, since that suggestion still relies on an unsafe `as` for the missing fields. `makeAgentConfiguration` now declares `: LightAgentConfigurationType` and provides every required field, matching the `makeAgent` helper in AgentBuilderSimilarAgentsSection.test.tsx.

front/hooks/useAgentFromSearchParam.ts
```diff
@@ -20,21 +22,14 @@ export function useAgentFromSearchParam(workspaceId: string) {
...
   useEffect(() => {
-    if (!agentConfiguration) {
+    if (
+      !agentConfiguration ||
+      appliedAgentIdRef.current === agentConfiguration.sId
+    ) {
       return;
     }
 
+    appliedAgentIdRef.current = agentConfiguration.sId;
```
PopDaph: The ref never resets, so re-applying the same agent through in-app navigation is skipped. 🙏🏻 

Repro on this branch:

1. Load /w/{wId}/conversation/new?agent=dust (fresh load). @dust is selected.
2. Switch to @deep-dive in the picker.
3. Click "New" in the sidebar. URL becomes /conversation/new, no reload.
4. Open @dust's details and click "New conversation". URL becomes /conversation/new?agent=dust via router.push.

The composer still shows @deep-dive, because the ref still holds dust from step 1. 

On main I think the same steps re-select @dust (the strip made the config go null and back, which retriggered the effect). 

Editing the address bar or refreshing does not show it, since a full load resets the ref, which is probably why it was easy to miss. 🙏🏻

achilleburah: > The composer still shows @deep-dive, because the ref still holds dust from step 1.

I'm not reproducing this. when clicking new conversation, the agent is correctly updated in the composer.

Is this the repro, or am I missing something ?  

https://github.com/user-attachments/assets/b88f0427-fe5d-4d63-8654-0440b9329813

PopDaph: Start from this url: https://pr-31961-merge--app.preview.dust.tt/w/0ec9852c2f/conversation/new?agent=dust

achilleburah: Resolved by uniformizing the url parameter behavior.

frankaloia (approved):
LGTM

PopDaph:
@achilleburah please check the attached PR from https://dust4ai.slack.com/archives/C050A0S2Z7F/p1789064943635739 

Just to make sure this does not break what you were trying to fix 🙏🏻

achilleburah:
Thanks for the ping 🙏  
No it does not break what I did but it does handle a case I seem to have missed

PR #31965, opened 2026-09-07
Gate Poke pool-usage premium-message columns by feature flag

avervaet:
## Description

The Poke pool-usage members table decided whether to show the legacy premium-message columns from subscription/contract state instead of the feature flags that actually govern that billing mode:

- Whether the workspace is on legacy premium-message billing at all now follows the `legacy_billing` feature flag (via `hasMetronomeFeature`) instead of `isSubscriptionMetronomeBilled(activeSubscription)`, so it stays in sync with the flag instead of trailing contract provisioning.
- The "Premium messages" column only renders when `enforce_premium_model_message_limit` is on for the workspace.
- The "Fair Usage Credits" column hides when `disable_fair_use_awu_limit` is on for the workspace.

Both new flags are decoupled from the existing `showPremiumMessageUsage` prop (which still governs the seat-vs-premium-message column layout) via two new optional `MembersUsageTable` props, `showPremiumMessageColumn` / `showFairUseCreditsColumn`, defaulting to `showPremiumMessageUsage` so no other caller changes behavior.

Poke-only (`variant=\"compact\"`); the customer-facing usage table is unaffected.

## Tests

Verified with `npx tsgo --noEmit` and `biome check` on the changed files. Not exercised end-to-end in the browser (would require toggling these flags on a real workspace via Poke).

## Risk

Low — scoped to the Poke admin pool-usage page only.

### DIFF
--- a/front/components/poke/pages/PoolUsagePage.tsx
+++ b/front/components/poke/pages/PoolUsagePage.tsx
@@ -185,11 +185,6 @@ export function PoolUsagePage() {
   const activeSubscription = workspaceInfo?.activeSubscription;
   const isLegacyPremiumMessagePlan =
     !!activeSubscription && !isCreditPricedPlan(activeSubscription.plan);
-  // Whether the pool/Metronome columns apply is driven by the `legacy_billing`
-  // feature flag (`hasMetronomeFeature`) rather than the subscription's actual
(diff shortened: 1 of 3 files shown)


avervaet:

- #31965 ◀

tdraier (approved):

PR #31967, opened 2026-09-07
[connectors] google_drive: retry 403 "User rate limit exceeded" instead of skipping the drive

tdraier:
## Description

Fixes intermittent "Google Drive connector not syncing" ([tasks#10311](https://github.com/dust-tt/tasks/issues/10311)), where syncs stall with `User rate limit exceeded.` errors.

**Root cause:** Google Drive returns **HTTP 403** (not 429) for user rate-limit/quota exhaustion, with `reason: "userRateLimitExceeded"` (message `"User rate limit exceeded."`). The `incrementalSync` activity's top-level `catch` treated **any** 403 as *"Looks like we lost access to this drive. Skipping"* and returned `undefined`. The workflow reads `undefined` as "no more pages" and stops **without advancing the sync page token** — so a transient rate-limit silently halts the drive's incremental sync and leaves it stale. Separately, the activity interceptor only mapped **429** to a retryable `rate_limit_error`, so 403 rate-limits bypassed it entirely.

The spreadsheets path already documents this exact trap (`spreadsheets.ts:745`), but the incremental-sync path never applied the same disambiguation.

**Changes:**
- `utils.ts`: new `isGoogleDriveRateLimitError()` helper matching 403/429 with rate-limit reasons (`userRateLimitExceeded`, `rateLimitExceeded`, `sharingRateLimitExceeded`, `dailyLimitExceeded`, `quotaExceeded`).
- `incremental_sync.ts`: re-throw rate-limit errors (bare `GaxiosError` and the `WithRetriesError` wrapper) **before** the "lost access → skip" branch, so Temporal retries with backoff instead of silently skipping the drive.
- `cast_known_errors.ts`: map 403 rate-limit errors to a retryable `rate_limit_error`, consistent with the existing 429 handling (clean classification + metrics).

## Tests

- Added 8 focused unit tests for `isGoogleDriveRateLimitError` (403/429 detection, permission-loss vs. rate-limit disambiguation, non-Gaxios inputs) in `utils.test.ts`. All pass.
- `tsc --noEmit` clean on the touched files.

## Risk

Low. The change narrows an existing overly-broad 403 handler: genuine permission-loss 403s still skip the drive as before; only rate-limit/quota 403s now retry instead of being swallowed. Retries use Temporal's default exponential backoff. Interceptor ordering (cast before log) is unchanged and matches the existing 429 path.

## Deploy Plan

Standard connectors deploy. No migration, no config change.

### DIFF
--- a/connectors/src/connectors/google_drive/temporal/cast_known_errors.ts
+++ b/connectors/src/connectors/google_drive/temporal/cast_known_errors.ts
@@ -1,3 +1,4 @@
+import { isGoogleDriveRateLimitError } from "@connectors/connectors/google_drive/temporal/utils";
 import {
   ExternalOAuthTokenError,
   ProviderWorkflowError,
@@ -72,6 +73,17 @@ export class GoogleDriveCastKnownErrorsInterceptor
     } catch (err: unknown) {
       if (isGoogleDriveInsufficientPermissionsError(err)) {
         throw new ExternalOAuthTokenError(err);
+      }
+
+      // Google returns 403 (not 429) for user rate-limit/quota exhaustion, e.g. "User
+      // rate limit exceeded.". Map it to a retryable rate_limit_error like the 429 case.
+      if (isGoogleDriveRateLimitError(err)) {
+        throw new ProviderWorkflowError(
+          "google_drive",
+          "403: Rate Limit Error",
+          "rate_limit_error",
+          err
+        );
       }
 
       if (err instanceof GaxiosError) {
--- a/connectors/src/connectors/google_drive/temporal/cast_known_errors.ts
+++ b/connectors/src/connectors/google_drive/temporal/cast_known_errors.ts
@@ -75,12 +75,13 @@ export class GoogleDriveCastKnownErrorsInterceptor
         throw new ExternalOAuthTokenError(err);
       }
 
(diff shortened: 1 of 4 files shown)


tdraier:

- #31967 ◀

connectors/src/connectors/google_drive/temporal/cast_known_errors.ts
```diff
@@ -74,6 +75,17 @@ export class GoogleDriveCastKnownErrorsInterceptor
         throw new ExternalOAuthTokenError(err);
       }
 
+      // Google returns 403 (not 429) for user rate-limit/quota exhaustion, e.g. "User
+      // rate limit exceeded.". Map it to a retryable rate_limit_error like the 429 case.
+      if (isGoogleDriveRateLimitError(err)) {
```
philipperolet, line 81: nit: The helper also matches 429 responses, so they now enter this branch and receive "403: Rate Limit Error", bypassing the existing 429 handler. Retries still work, but logs become misleading. What about restricting this branch to 403 or use the actual response status?

philipperolet (approved):
LGTM 🙏 left a minor com

connectors/src/connectors/google_drive/temporal/cast_known_errors.ts
tdraier: rather fixing the log to see the proper error code

PR #31969, opened 2026-09-07
feat(analytics): update side panel entry point

achilleburah:
## Description

New side conversation entry point in the Analytics View. 

## Tests

Tested Locally. 

### Screeshots

image

## Risk

Low. UI only. 

## Deploy Plan

Standard Deploy.


radjakahoul (changes requested):
To quickly unlock discoverability : can we keep the label you had "Ask analyst" with the back button aligned with other buttons (as is, just add label)?

We can have better discoverability later?

radjakahoul (approved):

PR #31970, opened 2026-09-07
TEMP: self-heal stale user_usage rows once Metronome activities unstick

avervaet:
## Description

emitMetronomeUsageEventsActivity already recomputes usage_type fresh on every run (with the new unattributed-Slack fallback), so once this deploy unblocks the currently-stuck activities they'll bill Metronome
correctly as programmatic — but RunUsageModel.usageType for those same runs stays frozen at the old wrong "user" classification, since nothing else touches an already-classified row.

Piggyback the correction on the same activity: it already knows the freshly-recomputed usage_type and the exact runs it's billing, so promote user -> programmatic there once Metronome ingestion succeeds. Scoped narrowly (only that one transition, only the runs just reclassified) so it's safe to drop once the backlog has drained — see the TODO on RunResource.selfHealSlackUsageTypeForRuns.

This only self-heals runs tied to currently-stuck retries (plus all new messages going forward, where it's a no-op). Messages that already succeeded under the old bug, before this release, aren't touched.

## Deploy Plan

**IMPORTANT:** Should be deployed alongside https://github.com/dust-tt/dust/pull/31943

### DIFF
--- a/front/lib/resources/run_resource.ts
+++ b/front/lib/resources/run_resource.ts
@@ -3,6 +3,10 @@ import { computeTokensCostForUsageInMicroUsd } from "@app/lib/api/assistant/toke
 import type { TokenUsage } from "@app/lib/api/llm/types/events";
 import type { Authenticator } from "@app/lib/auth";
 import { getModelConfigByModelId } from "@app/lib/llms/model_configurations";
+import {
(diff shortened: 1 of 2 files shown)


avervaet:

- #31943
    - #31970 ◀

tdraier (approved):

PR #31972, opened 2026-09-07
Spend caps: free seats never blocked by workspace pool depletion

tdraier:
## Description

A free seat spends from its **own lifetime credit allowance** (enforced by the per-user rate cap, `isFreeSeatLifetimeCapReached`) and **never** from the shared workspace pool. But `isUserBlockedByMetronome` only carved pool depletion out for personal-seat *credit states* (`user_seat` / `user_seat_low_balance`), so a free seat whose `creditState` was `capped` or `on_pool` was still returned as `credits_exhausted` whenever `poolCreditState` showed `depleted`.

Under the `enforce_user_spend_limit_rate_cap` flag this stranded free users after a credit regrant:
- The per-user cap comes from the fresh Redis rate-limiter counter → the input-bar warning clears and the user appears able to post.
- But `deriveBlockedReason` falls through to `workspacePoolDepleted`, which reads the **webhook-updated** `poolCreditState`. It's still `depleted`, so message posting stays blocked on `credits_exhausted` until the webhook eventually clears the stale state. The existing carve-out can't rescue it because the free seat's `creditState` is the equally-stale `capped`.

This extends the carve-out to also exempt free seats **by `seatType`** (structural, not credit-state), so a free seat's block status depends solely on the rate-limiter counter — a regrant unblocks immediately, with no webhook dependency.

The fix is correct in both flag modes: with the flag off, a truly-exhausted free seat is still blocked via `user_cap_reached` (`creditState === "capped"`), so there is no regression; a free seat with remaining lifetime allowance is (correctly) no longer blocked by unrelated pool depletion.

## Tests

Manually reproduced: a capped free user, after regranting credits, had the input bar clear but message posting still returned `credits_exhausted` until the Metronome webhook fired. With this change, posting is unblocked as soon as the rate-limiter counter reflects the regrant. Type-checked with `tsgo`.

## Risk

Low, narrowly scoped to one condition in `isUserBlockedByMetronome`. It only ever *removes* a block for free seats that never consume the pool; it cannot newly block anyone. Rollback is a trivial revert of the single hunk.

## Deploy Plan

Standard front deploy. No migration, no config, no coordinated ordering required.

### DIFF
--- a/front/lib/metronome/user_block.test.ts
+++ b/front/lib/metronome/user_block.test.ts
@@ -186,6 +186,42 @@ describe("isUserBlockedByMetronome", () => {
     expect(blocked).toBe("credits_exhausted");
   });
 
+  it("does not block a 'free' seat when the pool is depleted, even with a stale 'capped' state, once the rate cap clears", async () => {
+    // Regression: a free seat spends from its own lifetime allowance, never the
+    // pool. After a credit regrant the rate limiter clears
+    // (userCapBlockedOverride=false), but the webhook-driven creditState/pool
+    // status can still read stale ("capped"/"depleted"). The seat-type carve-out
+    // must unblock the free seat without waiting on the webhook.
+    mockGetActiveMembershipOfUserInWorkspace.mockResolvedValue({
+      seatType: "free",
+    });
+    redisValues.set("metronome:user_credit_state:ws_test:u_test", "capped");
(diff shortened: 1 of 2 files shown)


tdraier:

- #31972 ◀

avervaet (approved):
All good for me. We might want to add a regression test tho.

PR #31975, opened 2026-09-07
Pause Google Drive connectors on admin policy errors

spolu:
## Description

Google Drive OAuth refresh failures with `admin_policy_enforced` currently surface as generic errors, leaving sync activities retrying for days. Classify this Google-specific refresh error as `ExternalOAuthTokenError` so the existing handler marks the connector `oauth_token_revoked`, pauses it, and stops its workflows.

Google documents [`admin_policy_enforced`](https://developers.google.com/identity/protocols/oauth2/web-server#authorization-errors-admin-policy-enforced) as:

> The Google Account is unable to authorize one or more scopes requested due to the policies of their Google Workspace administrator.

The [refresh token documentation](https://developers.google.com/identity/protocols/oauth2#expiration) also explicitly lists this error when an administrator restricts a service requested by the app.

## Tests

Ran the OAuth and Temporal monitoring regression tests: 9 tests passed. The captured Google policy-error test fails before the fix and passes afterward. Coverage includes existing account restrictions, provider/error scoping, transient failures, and the automatic pause handler.

## Risk

Low. Uses the existing OAuth failure and pause path. Google Workspace policy must be corrected before the connection can be restored.

## Deploy Plan

Deploy `connectors`, including the Google Drive workers. Affected running connectors will pause on their next `admin_policy_enforced` failure after rollout.

### DIFF
--- a/connectors/src/lib/oauth.ts
+++ b/connectors/src/lib/oauth.ts
@@ -4,9 +4,11 @@ import type { OAuthConnectionType, OAuthProvider } from "@connectors/types";
 import { getOAuthConnectionAccessToken } from "@connectors/types";
 import type { LoggerInterface } from "@dust-tt/client";
 
-// Most connectors are built on the assumption that errors are thrown with special handling of
-// selected errors such as ExternalOauthTokenError. This function is used to retrieve an OAuth
-// connection access token and throw an ExternalOauthTokenError if the token is revoked.
+/**
+ * @cc [label:error-handling] google-drive-admin-policy-error
+ * Google Drive token refresh failures containing `admin_policy_enforced` must throw
+ * `ExternalOAuthTokenError` to trigger the connector's existing pause-and-stop handling.
+ */
 export async function getOAuthConnectionAccessTokenWithThrow({
   logger,
   provider,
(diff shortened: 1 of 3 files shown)


connectors/src/lib/oauth.ts
```diff
@@ -44,9 +48,11 @@ export async function getOAuthConnectionAccessTokenWithThrow({
       (tokRes.error.code === "provider_access_token_refresh_error" &&
         (tokRes.error.message.includes("invalid_grant") ||
           tokRes.error.message.includes("invalid_client"))) ||
-      // Happens with google drive
+      // Happens with Google Drive.
       (tokRes.error.code === "provider_access_token_refresh_error" &&
-        tokRes.error.message.includes("Account Restricted"))
+        (tokRes.error.message.includes("Account Restricted") ||
+          (provider === "google_drive" &&
+            tokRes.error.message.includes("admin_policy_enforced"))))
```
PopDaph, line 55: Suggest moving this to core instead. google_drive.rs already maps invalid_grant on a 400 to TokenRevokedError and connectors already handles token_revoked_error in the first branch of this if.

 WDYT?

PopDaph (approved):
Suggestion but feel free to discard and move forward with this way 👍🏻

connectors/src/lib/oauth.ts
spolu: Yep agreed that's what we'd want here :) but won't tackle it here :)

spolu:
cc @frankaloia for failed workflows retrying / eng on call

PR #31976, opened 2026-09-07
[front] fix: surface Outlook event categories and recurrence

matthieuhaguen:
## Description

Raised by **Orange Cyberdefense** (200-user pilot, potential 4,500-seat deal) : the Outlook calendar tools never surfaced an event's category, nor whether it was part of a recurring series, even though Microsoft Graph returns both. The same fields were visible to them through Copilot Studio, which is what prompted the report.

Nothing needed to change on the Microsoft tenant side, and we do not filter the Graph request:
no `$select` is applied on the calendar path, so the full event resource comes back. The data was
being dropped on our side, in two distinct places:

- **Schema.** `categories`, `type` and `seriesMasterId` were not declared in `OutlookEventSchema`.
  Zod strips undeclared keys, so they were discarded at parse time.
- **Rendering.** `recurrence` *was* declared, but `renderOutlookEvent()` never printed it. Since
  the tools return rendered markdown rather than raw JSON, it was invisible to the model anyway.

This declares the missing fields and renders categories plus a recurrence line. `recurrence` moves
from `z.any()` to a typed subset of Graph's `patternedRecurrence` now that we actually read it.

Rendering the pattern only reads the fields its type gives meaning to. Graph populates the whole
pattern object whatever the type is, defaulting the ones it does not use. A weekly pattern still
carries `index: "first"` and `dayOfMonth: 0`. Reading `index` unconditionally turns "every monday"
into "the first monday"; ignoring it turns "the second thursday of every third month" into "every
thursday". Both are wrong, so the renderer switches on the pattern type.

Occurrences returned by `/calendarView` carry `type` but a null `recurrence`, so a listed recurring
meeting reliably reports *that* it recurs but not its pattern. `seriesMasterId` is rendered so the
agent can fetch the master with `get_event` to get the pattern, without extra plumbing.

All four calendar tools render through `renderOutlookEvent()`, so `list_events`, `get_event`,
`create_event` and `update_event` are covered by the single change.

One limitation worth flagging for this account specifically: all calendar endpoints are `/me`-scoped
and we request `Calendars.ReadWrite.Shared`, so a shared mailbox is only reachable once its calendar
has been added to the signed-in user's own calendar list and is then targeted via `calendarId`.
There is no `/users/{upn}/calendar` path in our implementation. This PR does not change that, and
it is likely to come up next given the shared-mailbox requirement in the use case.

Sample output for an occurrence of a series:

```
Categories: Customer
Recurring: yes, occurrence of a series
Event ID: AAMkAGI2
Series Master Event ID: AAMkAGI2MASTER
```

## Tests

Added `rendering.test.ts` (19 tests), following the colocated `rendering.test.ts` pattern already
used by `skill_management` and `skill_authoring`. There was no test file for the Outlook servers
before. Coverage:

- The three new fields survive parsing, and the six recurrence pattern types render correctly —
  including that a weekly pattern must ignore the `index` Graph defaults in, and that
  `relativeMonthly` must keep it.
- Graph sends `recurrence: null` on non-recurring events, which must still parse.
- Resilience: each new field degrades to `undefined` on an unexpected shape instead of failing the
  event, and one bad event in a list no longer takes the whole list down.
  

**Not tested:** this has never run against a real Microsoft account. The fixtures encode Graph's
documented v1.0 behaviour, which I checked against the docs, not observed behaviour. Worth a manual
check on a mailbox with a categorised recurring meeting before merge.

## Risk

**Low, with one behavioural caveat worth a second opinion.**

Declaring these fields is what makes them reachable, but it also makes them able to *fail* the
parse, which `z.any()` and undeclared keys never could. That matters more than it looks: `listEvents`
parses through `z.array(OutlookEventSchema)`, so a single event failing validation fails the whole
batch and the tool returns no events at all. Since nothing branches on these fields, they are only
rendered. Each one carries `.catch(undefined)`. Worst case is now that a field silently does not
appear, which is exactly today's behaviour.

The caveat: exposing `Series Master Event ID` means the model now sees two identifiers on an
occurrence. If it passes the series master ID to `delete_event` or `update_event`, it would affect
the **whole series** instead of one occurrence. That was impossible before, since the ID was never
shown. The label is explicit and `Event ID` is still present, but no test can cover model behaviour.
Happy to drop that line if reviewers judge the trade-off not worth it, the cost is losing the
ability to retrieve a series' pattern from one of its occurrences.

Rollback is a plain revert: no migration, no persisted state, no config.

## Deploy Plan

Standard deploy, nothing to sequence.

- No DB migration, no schema change, no config or environment variable.
- No new OAuth scope and no new Graph request: the payload was already being fetched in full, we
  just stop discarding parts of it. No extra latency, no extra API call.
- No API contract change. Tool names, descriptions and input schemas are untouched, so the MCP
  metadata snapshots are unaffected.
- Output is append-only: the new lines land after `Show as:` and before `Status: Cancelled`, so
  nothing existing shifts.

Once deployed, worth confirming with Orange Cyberdefense that categories and recurrence now show up
as expected before closing the loop with them.

Unrelated follow-ups spotted while working on this, worth their own tickets:
`google_calendar/helpers.ts` declares `recurrence` and never renders it either — the same bug on
the other connector; and `create_event`/`update_event` still cannot *set* categories, which this
PR deliberately leaves out of scope.


front/lib/api/actions/servers/outlook/rendering.test.ts
```diff
@@ -0,0 +1,188 @@
+import { describe, expect, it } from "vitest";
```
aubin-tchoi: let's remove these tests

aubin-tchoi (approved):
LGTM once comment is addressed

PR #31977, opened 2026-09-07
Data sources carry their Qdrant shard key, collections declare their layout

flvndvd:
## Description

A new cell runs one workspace on a three node Qdrant cluster, and the way core routes vectors today was designed for hundreds of workspaces on a big one. Every data source hashes into one of 24 shard keys, a constant compiled into core, and every collection is created with 2 shards behind each key. On the cell that layout means 96 physical shards per collection to hold a handful of data sources, each of them pinned to two shards. We want fewer keys and more shards per key there, and the constant makes that impossible: the key count has to be the same in the binary and in the collection, for every cluster the binary talks to.

So the layout moves to where it belongs, the collection, and the data source remembers where it lives. The creation binary now takes the shard key count, the shards per key and the replication factor as flags, defaulting to what it always did. When a data source is registered, core asks Qdrant which shard keys its collection has, hashes the internal id into that count, and stores the resulting key in the data source's `qdrant_config`, one entry per cluster it writes to. From then on every upsert, search, scroll, count, delete and payload update routes with the stored key. Data sources created before this change have no stored key and keep hashing modulo 24, so nothing moves for them and no backfill is needed. The tenant filter on the data source id is untouched, isolation never depended on the key.

The migrator assigns the shadow cluster's key when a shadow write is set and drops it when removed. The ops tools that computed keys from the constant now read them from the row. Front only learns about the optional field on its type.

## Tests

Unit tests cover the three things that could silently go wrong: the hash stays balanced and in range for 24, 3 and 1 keys, the key list is read correctly from Qdrant's cluster info with local and remote shards and comes back empty for a plain collection, and a config without the new field parses and serializes exactly as before.

Against a local Qdrant in distributed mode, a collection created with 3 keys and 4 shards per key reports 12 shards, 4 under each key. Six fake data sources written with the test fill tool landed on keys 0, 2, 1, 2, 0 and 0, all 120 points counted under those keys, and one data source's 20 points are found under its key and nowhere else. The same id hashed the legacy way would have targeted `key_23`, which that collection does not have. Writing to a standalone Qdrant without shard keys fails at assignment with an explicit message rather than misrouting.

End to end, core-api running locally against a scratch Postgres, the local Elasticsearch and that 3 key Qdrant: registering a data source through the API returned its config with `"shard_keys": {"cluster-0": "key_1"}`, the row in Postgres carries the same, and `key_1` is what the internal id hashes to modulo 3 (modulo 24 would have given `key_4`). A search without a query on that data source scrolled Qdrant through the stored key and answered, the log line for the search reports `qdrant_shard_key="key_1"`, and deleting the data source removed its points through the same key without error.

## Risk

On legacy, no data source changes key, requests are identical, and the collections are created with the same parameters. The one new behaviour is a cluster info call at data source registration, cached per process per collection, and a registration now fails if Qdrant is unreachable, where it used to succeed and fail on the first upsert instead.

Rollback is a plain revert. Data sources registered in between carry a `shard_keys` entry the old code ignores, and since a legacy collection has 24 keys, the stored key and the legacy hash are the same key for them.

The shadow embedder collection must share the layout of the main one on a cluster, the assignment checks that and refuses otherwise.

## Deploy Plan

1. Deploy core to the legacy regions. No behaviour change there.
2. Pin the image on cell-00002 and create its two collections from prodbox with `--shard-key-count 3 --shards-per-key 4`, 24 physical shards per collection instead of 96.
3. First data source on the cell: its row shows `shard_keys: {"cluster-0": "key_<n>"}` with n in 0..2, and its points count under that key in the Qdrant cluster info.

### DIFF
--- a/core/bin/qdrant/create_collection.rs
+++ b/core/bin/qdrant/create_collection.rs
@@ -34,6 +34,18 @@ struct Args {
     /// Name of the cluster.
     #[arg(short, long)]
     cluster: QdrantCluster,
+
+    /// Number of shard keys. Data sources hash into them, one key per data source.
+    #[arg(long, default_value_t = SHARD_KEY_COUNT)]
+    shard_key_count: u64,
+
+    /// Number of shards behind each shard key, spread across the nodes.
+    #[arg(long, default_value_t = 2)]
+    shards_per_key: u32,
+
+    /// Number of copies of each shard.
+    #[arg(long, default_value_t = 2)]
+    replication_factor: u32,
 }
 
 async fn create_indexes_for_collection(
@@ -89,11 +101,10 @@ async fn create_indexes_for_collection(
     Ok(())
 }
 
-async fn create_qdrant_collection(
-    cluster: QdrantCluster,
-    provider_id: ProviderID,
-    model_id: SupportedEmbedderModels,
-) -> Result<()> {
+async fn create_qdrant_collection(args: &Args) -> Result<()> {
+    let cluster = args.cluster;
+    let provider_id = args.provider;
+    let model_id = args.model.clone();
     let qdrant_clients = QdrantClients::build().await?;
     let client = qdrant_clients.client(cluster);
     let raw_client = client.raw_client();
@@ -151,8 +162,8 @@ async fn create_qdrant_collection(
     if use_sharding {
         builder = builder
             .sharding_method(qdrant::ShardingMethod::Custom.into())
-            .shard_number(2)
-            .replication_factor(2)
+            .shard_number(args.shards_per_key)
+            .replication_factor(args.replication_factor)
             .write_consistency_factor(1);
     }
 
@@ -172,8 +183,8 @@ async fn create_qdrant_collection(
 
     // Only create shard keys when sharding is enabled
     if use_sharding {
-        // Then, we create the 24 shard_keys.
-        for i in 0..SHARD_KEY_COUNT {
+        // Then, we create the shard keys.
+        for i in 0..args.shard_key_count {
             let shard_key = format!("{}_{}", client.shard_key_prefix(), i);
 
             let operation_result = raw_client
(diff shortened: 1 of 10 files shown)


core/src/data_sources/qdrant.rs
```diff
@@ -166,31 +193,121 @@ impl DustQdrantClient {
...
+        let collection = self.collection_name(embedder_config);
+        if let Some(keys) = self.shard_key_names.lock().get(&collection) {
+            return Ok(keys.clone());
+        }
+        let info = self
+            .client
+            .collection_cluster_info(collection.clone())
+            .await
+            .map_err(|e| anyhow!("Error getting collection cluster info: {}", e))?;
+        let keys = Self::shard_key_names_from_cluster_info(&info);
```
flvndvd, line 258: Here we retrieve the collection config from Qdrant to get the actual sharding config.

core/bin/qdrant/create_collection.rs
```diff
@@ -34,6 +34,18 @@ struct Args {
     /// Name of the cluster.
     #[arg(short, long)]
     cluster: QdrantCluster,
+
+    /// Number of shard keys. Data sources hash into them, one key per data source.
+    #[arg(long, default_value_t = SHARD_KEY_COUNT)]
```
Fraggle: nit: SHARD_KEY_COUNT => DEFAULT_SHARD_KEY_COUNT ?

core/src/data_sources/qdrant.rs
```diff
@@ -18,13 +18,16 @@ use serde::{Deserialize, Serialize};
...
 pub enum QdrantCluster {
     #[serde(rename = "cluster-0")]
     Cluster0,
 }
 
 // See: https://app.notion.com/p/dust-tt/Design-Doc-Qdrant-re-arch-d0ebdd6ae8244ff593cdf10f08988c27
+// Key count of the collections created before data sources stored their shard key. A data source
+// without a stored key still hashes into these. Newer collections declare their own count and the
+// data sources created against them carry their key in `QdrantDataSourceConfig::shard_keys`.
 pub const SHARD_KEY_COUNT: u64 = 24;
```
Fraggle: Actually: LEGACY_XX ?

Fraggle (approved):
LGTM

core/src/data_sources/qdrant.rs
flvndvd: Updated.

PR #31978, opened 2026-09-07
[front] feat: add pagination on skill search

aubin-tchoi:
## Description

This PR implements pagination on the skills search. 

Fetch 1 more item in the index to know whether we have more elements and returns a cursor.
More or less the same interface as always: boolean `hasMore` and nullable string `nextCursor`.

Also moves the search file from lib to lib/api since it's backend-only.

## Tests

- Added tests.
- Tested locally.

## Risk

- Low.

## Deploy Plan

- Deploy front.


front/lib/skill_search/query.ts
```diff
@@ -6,9 +6,15 @@ import type {
   SkillSearchPermissionFiltering,
 } from "@app/types/api/skills";
 import type { estypes } from "@elastic/elasticsearch";
+import { z } from "zod";
 
 export const MAX_SKILL_SEARCH_RESULTS = 100;
 
+export const SkillSearchSortSchema = z.array(
+  z.union([z.string(), z.number(), z.boolean(), z.null()])
```
aubin-tchoi: im fixing this to keep a string and only decode way closer to the query

adrsimon (commented):
LGTM

PR #31983, opened 2026-09-07
Pause BigQuery connectors on policy violations

spolu:
## Description

BigQuery sync activities keep retrying when VPC Service Controls rejects requests with HTTP 403 and the structured `policyViolation` reason. The BigQuery interceptor currently misses this Google Cloud `ApiError` shape.

Classify these failures as `ThirdPartyConfigurationError` so the existing activity handler marks the connector `third_party_internal_error`, pauses it, and stops its workflows. Preserve the original error as the cause, including its VPC Service Controls unique identifier, for troubleshooting.

The [Google VPC Service Controls troubleshooting guide](https://docs.cloud.google.com/vpc-service-controls/docs/retrieve-troubleshoot-errors) identifies this BigQuery error:

> 403: VPC Service Controls: Request is prohibited by organization's policy.

The guide also explains how an administrator can use the unique identifier to locate the denial in Cloud Audit Logs.

## Tests

Ran the BigQuery error-classification and Temporal monitoring tests: 13 tests passed. Both the captured-error classification test and the interceptor-chain pause test fail before the fix and pass afterward. Coverage includes preserving rate-limit and other unrecognized failures, existing OAuth handling, successful activity results, and the original error cause.

## Risk

Low. The new classification requires HTTP 403 and a structured `policyViolation` reason and uses the existing configuration-error pause path. The customer must correct the Google Cloud policy before synchronization can resume.

## Deploy Plan

Deploy `connectors`, including the BigQuery workers. Affected running connectors will pause on their next matching activity failure after rollout.

### DIFF
--- a/connectors/src/connectors/bigquery/temporal/cast_known_errors.test.ts
+++ b/connectors/src/connectors/bigquery/temporal/cast_known_errors.test.ts
@@ -1,0 +1,97 @@
+import {
+  ExternalOAuthTokenError,
+  ThirdPartyConfigurationError,
+} from "@connectors/lib/error";
+import type {
+  ActivityExecuteInput,
+  ActivityInboundCallsInterceptor,
+  Next,
+} from "@temporalio/worker";
+import { describe, expect, it, vi } from "vitest";
+
+import { BigQueryCastKnownErrorsInterceptor } from "./cast_known_errors";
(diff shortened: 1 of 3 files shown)


spolu:
PMRR

PR #31989, opened 2026-09-07
Handle late sandbox child consumption attribution

flvndvd:
## Description

Improve agent-message consumption attribution diagnostics and support securely linked sandbox child actions that finish after an earlier attribution snapshot.

This PR:

- adds a read-only diagnostic script that reconstructs canonical LLM/tool billing, attributed items, run chronology, cache behavior, tool-result footprints, offload metadata, action coverage gaps, and Temporal lifecycle attempts;
- supports OpenAI and Gemini token-count comparisons for rendered tool results;
- allows a direct-charge-only sandbox child action to be appended after its parent attribution item has completed;
- preserves the invariant failure for ordinary late model-visible actions.

The production failure occurred when a sandbox child action was persisted after an earlier attribution pass. Its flat tool charge was valid, but the evidence-selection guard only accepted the child while the parent attribution item was pending. Once the parent item was complete, the final attribution pass rejected the new child and retried the same assertion indefinitely.

The exception is now based on the durable parent relationship and same producing run/step. Because sandbox child rows are direct-charge-only with a zero model-token footprint, appending one does not alter the model-token partition already stored for the parent run.

## Tests

- Added coverage for appending a late sandbox child after its parent evidence completed.
- Preserved coverage that unsupported late model-visible actions fail attribution.
- Focused agent-message consumption attribution suite: 16 tests passed.
- Biome passed on all changed files.

## Risk

The late-child exception is narrowly constrained to a durable sandbox parent/child relationship on the same producing run and step. Ordinary late actions still fail the existing assertion.

The diagnostic script is read-only unless explicitly executed and does not print tool-result contents.

## Deploy Plan

- Standard frontend deployment; no migration, backfill, or feature flag is required.
- Monitor agent-message attribution workflow retries and reconciliation failures.
- Use the diagnostic script on remaining failures to distinguish lifecycle gaps, caching, offloading, and token-estimation differences.


front/scripts/debug_agent_message_tool_footprints.ts
```diff
@@ -0,0 +1,1572 @@
+/**
```
flvndvd, line 1: Vibe coding to ease debugging.

sfriquet (approved):

adrsimon (approved):

PR #31990, opened 2026-09-07
feat: persist automations page tab in the URL

adrsimon:
## Description

The automations page (Triggers / Slack workflows) kept its active tab in local component state only. Reloading the page, using browser back/forward, or sharing a link to the Slack workflows tab always dropped you back on Triggers.

Switched the tab state to the `?tab=` query param, using the same convention already in place on the members page (`useQueryParams`). Default (Triggers) omits the param; selecting Slack workflows sets `?tab=slack-workflows`.

## Tests

Manually verified: switching tabs updates the URL, reloading on `?tab=slack-workflows` restores that tab, and browser back/forward moves between tabs.

## Risk

Low. Purely additive URL state on an existing client-only tab switch; no API or data changes.

## Deploy Plan

Deploy front.


aubin-tchoi (approved):
LGTM

PR #31935, opened 2026-09-07
feat: add seat_type to es consumption index

sfriquet:
## Description

Closes https://github.com/dust-tt/tasks/issues/10309

Add `seat_type` to `front.agent_message_consumption_analytics` so we can figure out what kind of seat the user had at the time the message was sent.
Useful to track free usage and be iso with `front.agent_message_analytics`.

## Tests

```
POST /front.agent_message_consumption_analytics/_search
{
  "query": {
    "exists": { "field": "user.seat_type"}
  },
  "_source": ["user.seat_type"]
}
```

## Risk

Low.

## Deploy Plan

Apply ES migration first then deploy front

PR #31991, opened 2026-09-07
[front] fix: add seat_type to consumption analytics test expectations

tdraier:
## Description

PR #31935 ("add seat_type to es consumption index") added a `seat_type` field to the consumption-analytics user object (`load.ts`) but did not update two tests that assert that object with exact `toEqual` matching. This broke the `front` test suite on `main`.

CI only surfaced the `llm_documents` failure (shard 4) because the workflow cancels the remaining shards on first failure — the `load.test.ts` failure was hiding in a cancelled shard.

Fix: add the missing `seat_type` field to both expected user objects (`null` where the user has no seat, `"workspace"` where they do).

## Tests

Recreated the test DB fresh and ran all 6 front shards locally — all green. The two affected files pass.

## Risk

None. Test-only change (2 lines), aligns expectations with already-shipped behavior.

## Deploy Plan

Merge to main. No migration or config.

### DIFF
--- a/front/lib/analytics/agent_message_consumption/llm_documents.test.ts
+++ b/front/lib/analytics/agent_message_consumption/llm_documents.test.ts
@@ -155,6 +155,7 @@ describe("buildLlmConsumptionDocuments", () => {
         user: {
           id: auth.getNonNullableUser().sId,
           group_ids: [],
+          seat_type: null,
         },
       }),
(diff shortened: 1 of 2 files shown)


tdraier:

- #31991 ◀

sfriquet (approved):

PR #31994, opened 2026-09-07
feat(front): use Claude Sonnet 5 as second candidate in auto and auto_fast streams

Nils-Fedrigo:
## Description

See https://dust4ai.slack.com/archives/C050A0S2Z7F/p1788337077502159 for context

Promotes Claude Sonnet 5 to the second candidate of the `auto` (Standard) and `auto_fast` (Basic) model streams, replacing Claude Sonnet 4.6 at those positions (`medium` effort for `auto`, `light` for `auto_fast`).

Sonnet 5 carries the same tier mapping as Sonnet 4.6 (`light` → cost_efficient, `medium` → balanced), so tier-capped resolution behaves the same. Sonnet 4.6 remains the Basic-tier floor at the end of the `auto` and `auto_complex` streams.

## Risk

Low — pure candidate reordering in the stream tables; instant rollback by reverting the commit. Workspaces where Anthropic is whitelisted and OpenAI is not will now route to Sonnet 5 instead of Sonnet 4.6.

### DIFF
--- a/front/lib/model_tiers/enabled_models.test.ts
+++ b/front/lib/model_tiers/enabled_models.test.ts
@@ -14,6 +14,7 @@ import {
   CLAUDE_OPUS_4_8_DEFAULT_MODEL_CONFIG,
   CLAUDE_OPUS_5_DEFAULT_MODEL_CONFIG,
   CLAUDE_SONNET_4_6_DEFAULT_MODEL_CONFIG,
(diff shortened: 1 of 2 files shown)


pmilliotte (approved):
Should replace also auto complex from sonnet 4.6 to 5 right ?

Nils-Fedrigo:
normally we never reach the Basic floor, so I think it's good

PR #31995, opened 2026-09-07
[viz] dev frame-ancestors include ALLOWED_VISUALIZATION_ORIGIN

PopDaph:
Frames never load in a dust-hive env: viz's dev `frame-ancestors` CSP is a fixed list (`localhost:3000`, `3011`, `3012`, `16011`), so a front on `localhost:11011` gets `ERR_BLOCKED_BY_RESPONSE` and the panel spins forever. Going through the forwarder on 3011 works, which is why nobody noticed.

In development the list now also includes the origins from `ALLOWED_VISUALIZATION_ORIGIN`, the variable the content page already uses to filter `postMessage` senders and which dust-hive sets to its ports. Production is unchanged. Added a `viz/CONTRACTS` entry tying the two uses of the variable together.

Verified: after restarting viz in the hive, the header lists `http://localhost:11011`.

### DIFF
--- a/viz/next.config.mjs
+++ b/viz/next.config.mjs
@@ -1,8 +1,25 @@
 /** @type {import('next').NextConfig} */
 const isDev = process.env.NODE_ENV === "development";
 
+// Dev fronts that may embed viz. dust-hive envs run on other ports and list them in
+// ALLOWED_VISUALIZATION_ORIGIN (the same variable the content page checks), so include those too.
+const DEV_FRAME_ANCESTORS = [
+  "http://localhost:3000",
+  "http://localhost:3011",
+  "http://localhost:3012",
+  "chrome-extension://okjldflokifdjecnhbmkdanjjbnmlihg",
+  "http://localhost:16011",
+  ...(process.env.ALLOWED_VISUALIZATION_ORIGIN ?? "")
+    .split(",")
+    .map((origin) => origin.trim())
+    .filter(Boolean),
+];
+
+const PROD_FRAME_ANCESTORS =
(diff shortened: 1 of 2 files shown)


viz/next.config.mjs
```diff
@@ -1,9 +1,35 @@
...
+];
+
+const PROD_FRAME_ANCESTORS = [
+  "https://dust.tt",
+  "https://app.dust.tt",
+  "https://eu.dust.tt",
+  "https://front-edge.dust.tt",
+  "https://eu.front-edge.dust.tt",
+  "https://*.preview.dust.tt",
+  "chrome-extension://okjldflokifdjecnhbmkdanjjbnmlihg",
```
PopDaph, line 24: Wait this one is duplicated (was also duplicated on the diff, checking if it's the dev or prod one to de-duplicate)

aubin-tchoi (approved):
LGTM

viz/next.config.mjs
zmarouf: @flvndvd Do we still need the regional endpoints here?
If yes we need `eu2`

zmarouf: Ah I guess these are for the landings. Never mind then

zmarouf (approved):
LGTM post dedup

viz/next.config.mjs
PopDaph: ok it's cause we want both the prod & dev extension to access prod frames!

PopDaph: (sorry was answering to myslef before seeing your messages)

aubin-tchoi: yes we only need `app.dust.tt` AFAICT

PR #31952, opened 2026-09-07
Fix: NOVU now uses cells instead of regions.

Fraggle:
## Description

Map Novu application identifiers to `cell-00000`, `cell-00001`, and `cell-00002` in the frontend environment configs, and update `useNovuClient` to select the identifier from `CellInfo.name` while keeping API and WebSocket endpoints region-based. This aligns Novu client initialization with the current cell topology.

## Tests

Not run; this PR updates frontend environment configuration and Novu client selection only.

## Risk

Low risk. A missing or incorrect cell-specific identifier prevents Novu initialization for that cell; reverting the frontend deployment restores the previous mapping.

## Deploy Plan

Deploy `front`. No SQL migrations or infrastructure changes required.

PR #31997, opened 2026-09-07
dust-hive: Novu per-cell placeholders in generated env

PopDaph:
## Description

Since #31952 front reads Novu config per cell (`NEXT_PUBLIC_NOVU_APPLICATION_IDENTIFIER_CELL_00000`, `NEXT_PUBLIC_NOVU_API_URL_US`, `NEXT_PUBLIC_NOVU_WEBSOCKET_API_URL_US`) and throws when they're missing, so a dust-hive front crashes on load with "NEXT_PUBLIC_NOVU_APPLICATION_IDENTIFIER for cell is not set".

The generated env now exports placeholders for the three so the app boots. Notifications won't connect locally, same as before.

## Tests

`bun run check` passes. Applied to the aug-2 env, restarted front-spa-app, the app loads again.

## Risk

None, dev tooling only.

## Deploy Plan

None.


PopDaph:
Sorry my agent is drunk

PR #31998, opened 2026-09-07
Admin access: admin_can_see_private_entities feature flag lifts the agent and skill redaction

fabiencelier:
## Description

add a new flag `admin_can_see_private_entities` to bypass the visibility restrictions for admin when the workspace requires it.
Asmin can always see anything by adding themself as editor or agent and members of the space, this flag allow them to see everything without these requirements which may not scale for large organisations requiring to check all agents.

## Tests

 - added unit test
 - tested locally

## Risk

low, all behind a feature flag which can be removed

## Deploy Plan

deploy front


aubin-tchoi (approved):
LGTM but I think there are other downstream consumers that also need to check this, for instance I think you'll lose the skills in the agent info because `/api/w/:wId/assistant/agent_configurations/:aId/skills` still gates access

PR #31999, opened 2026-09-07
[front] side panel: closing goes back to the previous panel

PopDaph:
## Description

Open a frame, then the agent opens a plan mode and if you close the plan mode panel you lost your frame and have to get it back manually. Or the other way around.

The side panel now keeps a small history. Opening a panel of another type on top pushes the current one, closing (X or re-clicking the chip) goes back to the previous panel when there is one, otherwise collapses. Chips still open their own panel. The X stays a plain cross.

Panels of the same type replace each other (browsing tool steps or Frames doesn't pile up), a panel never appears twice in the history, an archived plan is dropped from it, deep-linked panels are the bottom of the stack, browser back/forward is followed, and switching conversation or dismissing the whole panel clears it. A Frame refreshed during streaming updates the shown panel instead of stacking a copy.

Follow-up to #31895.

## Risks

Mess up with the side panel state, can be rolled back safely.

## Tests

Provider tests cover the history rules. Tested locally and on a review app.

## Deploy

Deploy front.


front/components/assistant/conversation/ConversationSidePanelContext.tsx
```diff
@@ -76,6 +156,10 @@ interface ConversationSidePanelContextType {
   data: string | undefined;
 }
 
+// Defensive only: same-type panels replace each other and a panel appears once, so the history
+// is bounded by the number of panel types.
+const MAX_PANEL_HISTORY = 20;
```
ykmsd: do we really want to store 20? It's hard for me to imagine that you want to go back to that deep, maybe 5 or even 3?

front/components/assistant/conversation/plan_mode/PlanPanelChip.tsx
```diff
@@ -56,13 +62,18 @@ export function PlanPanelChip({
     } else if (action === "close") {
       closePanel();
     }
+    // An archived plan must not come back through the side panel history either.
+    if (next === "empty") {
+      forgetPanels(PLAN_SIDE_PANEL_TYPE);
```
ykmsd: ultra nit but `forgetPanels` is kind of weird name no? Maybe more like `removeFromPanelHistory`?

front/components/assistant/conversation/ConversationSidePanelContext.tsx
ykmsd: discussed here, we will iterate based on the feedback https://dust4ai.slack.com/archives/C09GELMTTRT/p1789048376689089

ykmsd (approved):

front/components/assistant/conversation/ConversationSidePanelContext.tsx
PopDaph: Will set 3 👍🏻

front/components/assistant/conversation/plan_mode/PlanPanelChip.tsx
PopDaph: done!

PR #32001, opened 2026-09-07
[Resources] Skip empty user and membership lookups

philipperolet:
## Description

Follow-up to https://github.com/dust-tt/dust/pull/31889, based independently on main.

Sequelize still sends SQL for empty `IN` arrays. Return early in `UserResource.fetchByModelIds`, `GroupResource.getActiveMembershipsForGroups`, and `MembershipResource.getActiveMemberships` when their input lists are empty, so callers do not need their own guards.

Results and validation stay unchanged: omitted `users` still lists workspace memberships, while `users: []` returns none and still requires a workspace. Paginated empty-user lookups skip both the fetch and count queries.

## Risks

Blast radius: User and membership lookups with empty input lists.
Risk: low

## Deploy Plan

- Deploy front-api and front workers. No migration required.

## Tests

- [x] User, group, membership, and agent resource suites: 122 tests passed.
- [x] New no-query tests fail without the guards and pass with them.

### DIFF
--- a/front/lib/resources/membership_resource.ts
+++ b/front/lib/resources/membership_resource.ts
@@ -142,6 +142,10 @@ export class MembershipResource extends BaseResource<MembershipModel> {
     return { memberships: orderedResourcesFromModels(rows), total: count };
   }
 
+  /**
+   * @cc [owner:philipperolet,label:performance] empty-users-skip-membership-queries
+   * With a workspace, empty users return no memberships without querying; omitted users do not filter.
+   */
   static async getActiveMemberships({
     users,
     workspace,
@@ -155,6 +159,10 @@ export class MembershipResource extends BaseResource<MembershipModel> {
   }): Promise<MembershipsWithTotal> {
     if (!workspace && !users?.length) {
       throw new Error("At least one of workspace or userIds must be provided.");
+    }
+
+    if (users?.length === 0) {
+      return { memberships: [], total: 0, nextPageParams: undefined };
     }
 
     const whereClause: WhereOptions<InferAttributes<MembershipModel>> = {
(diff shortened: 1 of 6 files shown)


front/lib/resources/membership_resource.ts
```diff
@@ -142,6 +142,10 @@ export class MembershipResource extends BaseResource<MembershipModel> {
     return { memberships: orderedResourcesFromModels(rows), total: count };
   }
 
+  /**
+   * @cc [owner:philipperolet,label:performance] empty-users-skip-membership-queries
+   * With a workspace, empty users return no memberships without querying; omitted users do not filter.
```
spolu: Not sure I parse the end or it, "omitted users do not filter" ?

philipperolet: it meant that if users is not passed or null, we don't return early (so we don't break the function logic). I agree it's poorly written. Hadn't reviewed it yet.

do you have time to discuss contracts? a bit worried here that this kind of minor change (adding a small if clause to exit on 0 length input) triggers a 2-line comment on the function with a specific ownership. The cc clauses would end up bigger than the functions no?

Would there be a way in your opinion that contracts focus on the function's core purpose vs on the various implementation details?

philipperolet:
```suggestion
   * With a workspace, empty users return no memberships without querying. This does not 
   *  apply to undefined/null users.
```

front/lib/resources/membership_resource.ts
```diff
@@ -142,6 +142,11 @@ export class MembershipResource extends BaseResource<MembershipModel> {
     return { memberships: orderedResourcesFromModels(rows), total: count };
   }
 
+  /**
+   * @cc [owner:philipperolet,label:performance] empty-users-skip-membership-queries
+   * With a workspace, empty users return no memberships without querying. This does not 
+   *  apply to undefined/null users.
```
philipperolet:
```suggestion
   * apply to undefined/null users.
```

spolu (commented):
LGTM

PR #25865, opened 2026-05-20
[REGION] EU-only models toggle

pmilliotte:
## Description

Adds an admin "EU only" toggle on the Model Providers page that filters models to those available in the workspace's region. Enabling is blocked if any active agent still uses a non-regional model. Agent builder's model picker shows a 🇪🇺 suffix on regional models in `europe-west1`.

**In admin settings:**

image

**Agent builder:**

image

## Tests

Manual.

## Deploy Plan

- Standard deploy, no migration
- Deploy `main` on `front`

---

<a href="https://stagereview.app/dust-tt/dust/pull/25865">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://stagereview.app/assets/gh-open-in-stage-dark.svg">
    Open in Stage
  </picture>
</a>

PR #32002, opened 2026-09-07
fix(front): restore the EU-hosted flag in the model picker

Nils-Fedrigo:
## Description

Screenshot 2026-09-08 at 10 16 27

The tiered model picker shipped in #29436 stopped rendering `RegionalFlag`, so EU
workspaces lost the 🇪🇺 badge that the previous agent-builder list showed next to
EU-hosted models (added in #25865, extended in #28592). The "EU-hosted models only"
toggle still hard-filters the catalog, so this was never a residency regression —
customers simply could no longer self-verify residency from the UI and escalated
instead. Fixes dust-tt/tasks#10208.

The flag is back, on the same gate the old list used
(`use_vertex_for_supported_models` and a non-BYOK plan — BYOK models run on the
customer's own provider keys, so we make no hosting claim there):

- `useModelPickerModels` exposes `hostingRegion` through `modelProps`, so every
  surface spreading it (input bar, agent builder, bulk "Set model", models filter)
  picks the flag up.
- Each model row under "More models" carries the flag when the model's
  `regionalAvailability` is `true` for that region — including locked rows, next to
  the padlock.
- The Basic / Standard / Premium rows carry it too, based on the model they
  currently resolve to. Those rows name a concrete model, so flagging only the
  "More models" rows would have read as "the recommendations are *not* EU-hosted".

## Tests

- Unit coverage for the two new predicates in `modelPickerUtils.test.ts`
  (`isModelHostedInRegion`, `isTierResolvedModelHostedInRegion`, including the
  in-flight-resolutions case), plus a render assertion in
  `ModelPickerModelRow.test.tsx` that the row surfaces the flag it is given.
- `npm run test -- components/model_picker` → 26 passing.
- Worth a manual pass on an EU cell with `use_vertex_for_supported_models` on: the
  flag should appear on every EU-hosted model row and on the tier rows whose
  resolved model is EU-hosted, and nowhere at all on a US cell or a BYOK plan.

BTW : The flag now appears on *every* EU-hosted model, so a
workspace with "EU-hosted models only" enabled sees it on every row.

### DIFF
--- a/front/components/model_picker/ModelPickerContent.tsx
+++ b/front/components/model_picker/ModelPickerContent.tsx
@@ -11,8 +11,10 @@ import {
   getModelLockTooltip,
   getTierLockReason,
   getTierResolvedModelLabel,
+  isTierResolvedModelHostedInRegion,
   isTierSelected,
 } from "@app/components/model_picker/modelPickerUtils";
+import { RegionalFlag } from "@app/components/shared/RegionalFlag";
 import type {
   EnabledModelConfigurationType,
   ModelStreamResolutionsType,
@@ -22,6 +24,7 @@ import type {
   ModelMakerIdType,
   ReasoningEffort,
 } from "@app/types/assistant/models/types";
+import type { RegionType } from "@app/types/region";
 import {
   Button,
   ChevronDown,
@@ -41,6 +44,9 @@ interface ModelPickerContentProps {
   ignoreTierRestrictions: boolean;
   tiers: ModelTierDefinition[];
   degradedModelIds: ReadonlySet<string>;
+  // Region whose hosting flag is shown next to the models hosted there; `null`
+  // when the workspace has no hosting region to advertise.
+  hostingRegion: RegionType | null;
   makerGroups: MakerGroup[];
   streamModels: EnabledModelConfigurationType[];
   streams: ModelStreamResolutionsType | null;
@@ -70,6 +76,7 @@ export function ModelPickerContent({
   ignoreTierRestrictions,
   tiers,
   degradedModelIds,
+  hostingRegion,
   makerGroups,
   streamModels,
   streams,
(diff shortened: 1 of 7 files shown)


front/components/model_picker/ModelPickerContent.tsx
```diff
@@ -41,6 +44,9 @@ interface ModelPickerContentProps {
   ignoreTierRestrictions: boolean;
   tiers: ModelTierDefinition[];
   degradedModelIds: ReadonlySet<string>;
+  // Region whose hosting flag is shown next to the models hosted there; `null`
+  // when the workspace has no hosting region to advertise.
```
Nils-Fedrigo:
```suggestion
```

front/components/model_picker/ModelPickerContent.tsx
```diff
@@ -116,6 +123,13 @@ export function ModelPickerContent({
             />
           );
         }
+        // A tier row names the model it currently resolves to, so it carries
+        // the same hosting flag as that model's own row would.
```
Nils-Fedrigo:
```suggestion
```

front/components/model_picker/ModelPickerMakersView.tsx
```diff
@@ -40,6 +43,9 @@ interface ModelPickerMakersViewProps {
   // credit-based plan).
   lockPremiumEfforts: boolean;
   degradedModelIds: ReadonlySet<string>;
+  // Region whose hosting flag is shown next to the models hosted there; `null`
+  // when the workspace has no hosting region to advertise.
```
Nils-Fedrigo:
```suggestion
```

front/components/model_picker/ModelPickerModelRow.tsx
```diff
@@ -24,6 +24,9 @@ interface ModelPickerModelRowProps {
   isDefault: boolean;
   lockReason: ModelLockReason | null;
   isDegraded: boolean;
+  // The hosting flag shown next to the model name (e.g. 🇪🇺 for an EU-hosted
+  // model), or `null` when the model has no hosting region to advertise.
```
Nils-Fedrigo:
```suggestion
```

Fraggle (approved):

PR #32003, opened 2026-09-07
feat(models): add a feature flag to remove access to GPT 6 Astra

Nils-Fedrigo:
## Description

See context here https://app.plain.com/workspace/w_01KJG5G96PGBWD5YG61BFEDSJ3/thread/th_01M1XBDKZWG5YAXDCV9N6W1388/

Adds a `disable_gpt_6_astra` feature flag that removes GPT 6 Astra from a workspace: once
enabled, the model disappears from the model picker and is refused at message time.

To do this without hardcoding a model anywhere in the access logic, model configurations gain
an optional `unavailableIfFeatureFlag` field. It is the mirror image of the existing
`availableIfOneOf`: an opt-out switch checked first in `checkModelSpecificAccessRules`, which
vetoes every other grant (upgraded plan, credit-priced plan, `plansWithAdvancedModels` and
`claude_4_5_opus_feature`). `GPT_6_ASTRA_MODEL_CONFIG` is the first model to declare it, and
the next model kill switch is a one-liner.

Because the veto sits in `isModelAvailable`, it applies everywhere model access is decided
rather than only in the UI:

- the picker (`/api/w/:wId/models` → `getAvailableModelsForWorkspace` → `filterEnabledModels`)
  stops returning the model,
- posting a message on an agent configured on Astra is rejected in `postUserMessage`,
- `resolveModel` / `selectEnabledModel` skip a user selection of Astra and fall through to the
  next candidate.

Astra is not part of any `MODEL_STREAMS` candidate pool nor of `PREFERRED_LARGE_MODEL_CONFIGS`,
so no auto/stream tier can route to it behind the flag.

Also documents the veto as a code contract on `isModelAvailable`, and mirrors the new flag into
the `@dust-tt/client` `WhitelistableFeature` union (additive, it is a `FlexibleEnumSchema`).

## Tests

Three cases added to `front/lib/assistant.test.ts`: the flag vetoes a model that another
condition would otherwise grant, absence of the flag leaves the model available, and the real
`GPT_6_ASTRA_MODEL_CONFIG` flips from available to unavailable on a credit-priced plan.
`lib/assistant.test.ts` and the model availability / tier suites pass locally.

## Risk

Low, and the flag is off everywhere by default so the change is a no-op until it is enabled on
a workspace. Worth knowing before enabling it: an agent already configured on Astra in that
workspace will fail to answer with "The model is not supported." — the same behavior as any
model becoming unavailable through a plan downgrade or a provider being un-whitelisted. Agents
should be moved off Astra before, or right after, flipping the flag.

## Deploy Plan

Nothing specific — no migration, and the flag is enabled per workspace from Poke.

### DIFF
--- a/front/lib/assistant.ts
+++ b/front/lib/assistant.ts
@@ -27,7 +27,21 @@ function checkModelSpecificAccessRules(
     plan: PlanType | null;
   }
 ): boolean {
-  const { availableIfOneOf, largeModel } = modelConfiguration;
+  const { availableIfOneOf, largeModel, unavailableIfOneOf } =
+    modelConfiguration;
+
+  // Opt-out check: matching a model-specific unavailability rule vetoes every
+  // grant below.
+  if (unavailableIfOneOf) {
+    const { featureFlag } = unavailableIfOneOf;
+
+    const matchesFeatureFlagCondition =
+      featureFlag !== undefined && featureFlags.includes(featureFlag);
+
+    if (matchesFeatureFlagCondition) {
+      return false;
+    }
+  }
 
   // First check: downgraded plans only have access to the small models.
   if (largeModel && !isUpgraded(plan)) {
@@ -61,6 +75,12 @@ function checkModelSpecificAccessRules(
   return true;
 }
 
+/**
+ * @cc [owner:Nils-Fedrigo,label:product] unavailable-feature-flag-vetoes-availability
+ * A model must be reported unavailable whenever it satisfies one of its
+ * `unavailableIfOneOf` conditions, whatever `availableIfOneOf`, `plan`, `region` or
+ * `regionalModelsOnly` would otherwise grant.
+ */
 // Returns true if the model is available to the workspace for build.
 export function isModelAvailable(
   m: ModelConfigurationType,
--- a/front/types/assistant/models/types.ts
+++ b/front/types/assistant/models/types.ts
@@ -172,13 +172,7 @@ export type ModelConfigurationType = Omit<
     // If set, model is available if feature flag is enabled.
     featureFlag?: WhitelistableFeature;
   };
-  // Mirror image of `availableIfOneOf`: an opt-out that pulls a model back from
-  // a workspace without touching its availability conditions.
-  // If undefined or empty, no opt-out applies.
-  // If defined, model is unavailable as soon as it satisfies one of the
-  // conditions, whatever `availableIfOneOf` and the plan would grant.
   unavailableIfOneOf?: {
-    // If set, model is unavailable if feature flag is enabled.
     featureFlag?: WhitelistableFeature;
   };
   // Pre-requisite: must be available.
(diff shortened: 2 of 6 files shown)


front/lib/assistant.ts
```diff
@@ -61,6 +70,12 @@ function checkModelSpecificAccessRules(
   return true;
 }
 
+/**
+ * @cc [label:product] unavailable-feature-flag-vetoes-availability
```
spolu: Please add yourself as owner (we have too many examples without owner I presume which make agents not add it by default)

spolu:
Goal is to release astra to all credit based but no legacy?

Nils-Fedrigo:
@spolu one customer (mano mano) is having second thoughts about adding Astra to their workspace as they don't want to blow their entire credit allocation (risk management)
Even though it's probably unjustified (probably same cost per task), we don't have data yet to counter argument efficiently, so to alleviate the pain we remove it for this customer only.

There are multiple product solutions we could envision if we have more signal on this (higher tier above premium with only Astra in, ability for admisn to remove models,...)! We don't want to rush to a solution for a n of 1 problem tough.

spolu:
Gotcha. Do you confirm we did not roll out Astra to legacy non-credit-based workspaces? Or is the policy to roll it out but be capped by the 25 messages thing for large models?

Nils-Fedrigo:
> Gotcha. Do you confirm we did not roll out Astra to legacy non-credit-based workspaces? Or is the policy to roll it out but be capped by the 25 messages thing for large models?

We DID roll out Astra to legacy pricing, the gate is the same as Sol for instance:
```
  availableIfOneOf: {
    creditPricedPlan: true,
    plansWithAdvancedModels: true,
    featureFlag: "claude_4_5_opus_feature",
  },
```

It's capped by fair use limits.

We can always revisit though!

front/types/assistant/models/types.ts
```diff
@@ -136,6 +136,10 @@ export const ModelConfigurationSchema = z.object({
   // Specify if the model is available in specific regions.
   regionalAvailability: z.record(z.enum(SUPPORTED_REGIONS), z.boolean()),
   availableIfOneOf: AvailabilityConditionSchema.optional(),
+  // If set, the model is unavailable to workspaces that have this feature flag
+  // enabled, whatever `availableIfOneOf` grants. Opt-out switch used to pull a
+  // model back from a workspace without touching its availability conditions.
```
Nils-Fedrigo:
```suggestion
```

front/types/assistant/models/types.ts
```diff
@@ -167,6 +172,15 @@ export type ModelConfigurationType = Omit<
     // If set, model is available if feature flag is enabled.
     featureFlag?: WhitelistableFeature;
   };
+  // Mirror image of `availableIfOneOf`: an opt-out that pulls a model back from
+  // a workspace without touching its availability conditions.
+  // If undefined or empty, no opt-out applies.
+  // If defined, model is unavailable as soon as it satisfies one of the
+  // conditions, whatever `availableIfOneOf` and the plan would grant.
+  unavailableIfOneOf?: {
+    // If set, model is unavailable if feature flag is enabled.
```
Nils-Fedrigo:
```suggestion
  unavailableIfOneOf?: {
```

Fraggle (approved):
LGTM

PR #28834, opened 2026-07-10
[front] feat: add Kimi K2.6 (Fireworks), deprecate K2.5

adrsimon:
## Description

Adds **Kimi K2.6** (served via Fireworks) to both the legacy and new (`model_constructors`) LLM routers, and deprecates **Kimi K2.5** in its favor.

- K2.6 mirrors K2.5's config (262k context, vision, native reasoning, gated behind `fireworks_new_model_feature`) with updated pricing (0.95 / 4.0, cache-read 0.16).
- `dust-kimi`, `dust-kimi-medium`, `dust-kimi-high` global agents now run K2.6.
- K2.5 is marked legacy and removed from the model picker, but kept in the supported-model registry, whitelist and pricing so existing agents keep running and historical token accounting stays intact.

## Tests

New-router live integration suite passes 40/40 against the Fireworks API (`fireworks_global_kimi_k2_dot_six.test.ts`, `RUN_LLM_TEST=true`). `tsgo` clean, formatted.

## Risk

Low. Additive model; K2.5 stays supported for existing agents.

## Deploy Plan

After merge, migrate existing agents off K2.5:

```
NODE_OPTIONS="--max-old-space-size=8192" npx tsx scripts/update_agent_models.ts --fromModel accounts/fireworks/models/kimi-k2p5 --toModel accounts/fireworks/models/kimi-k2p6 --onlyActive --execute
```

PR #26958, opened 2026-06-08
Remove decommissioned DeepSeek R1 models

fontanierh:
## Description

DeepSeek R1 was decommissioned across every provider that served it (Fireworks `deepseek-r1-0528`, TogetherAI `DeepSeek-R1`, and the DeepSeek API `deepseek-reasoner`). This removes the three model configs, the `deepseek-r1` global agent, the `deepseek_r1_global_agent_feature` flag, and the related pricing/parser/route entries. `deepseek-chat` and the other DeepSeek/Fireworks models are untouched.

A migration (`front/migrations/20260608_migrate_deepseek_r1_models.ts`) repoints active agents on any of the three ids to the latest DeepSeek model, DeepSeek V4 Pro (Fireworks). All model id strings (sources and target) are hardcoded in the migration so it stays a frozen snapshot independent of later config changes.

Drive-by: fixed stale `deepseek-reasoner` comments in `core/src/providers/deepseek.rs` (the gate is `MODEL_IDS_WITH_TOOLS_SUPPORT`, the named model is gone).

fixes https://github.com/dust-tt/tasks/issues/8624

## Tests

Ran `assistant.test.ts` and `global_agents.test.ts` locally, both green. tsc clean across front, front-api, and sdks/js. Migration is dry-run by default (no `--execute`); it has not been run against a database yet, the dry-run + execute happen at deploy.

## Risk

Low. The migration runs before front rolls out (see deploy plan), so no active agent is left pointing at a removed model id. A removed `modelId` does not resolve in `SUPPORTED_MODEL_CONFIGS` at runtime, which is why the order matters. Historical conversations and orphaned global-agent settings rows render fine since the old sId is filtered out.

## Deploy Plan

- [ ] Run the agent migration first, in both regions, so no active agent points at a removed id mid-deploy:
      `npx tsx front/migrations/20260608_migrate_deepseek_r1_models.ts --execute`
- [ ] Deploy front.
- [ ] Clear orphaned flag rows:
      `npx tsx front/scripts/delete_legacy_feature_flag.ts --featureFlag deepseek_r1_global_agent_feature --execute`

PR #32005, opened 2026-09-07
docs(dust-llm): fix drifted paths and add model deprecation/removal guide

Nils-Fedrigo:
## Description

The `dust-llm` skill had drifted from the codebase in two places, and was silent on half of the task it is meant to cover.

**Drift (flagged in review):**

- It pointed at `front/lib/api/assistant/token_pricing/static_model_reasoning_efforts.ts` and `.../tiers.ts`. Both now live under `front/types/assistant/models/`, and the test is `model_tiers.test.ts`, not `tiers.test.ts`. Fixed in the §B table, the Verify command block, and Troubleshooting.
- It listed a `marketing/` mirror for models and token pricing. Those files no longer exist — `marketing` fetches `/api/marketing/model-credits`, which `front/lib/api/marketing/model_credits.ts` derives at request time from `SUPPORTED_MODEL_CONFIGS` + `MODEL_PRICING`. The dead rows are replaced with a note on what actually gates a model appearing there (`MODEL_PRICING` entry, not `isLegacy`, no feature flag).

**Missing half — deprecating/removing the superseded model.** Reconstructed from the Kimi K2.5/K2.6 (`f2824da5c5e`, #28834) and DeepSeek R1 (`3ca8d834527`, #26958) commits, then re-verified against current code:

- *Path 1, superseded but still served*: `isLegacy`/`isLatest`, drop from `USED_MODEL_CONFIGS`, repoint `MODEL_STREAMS` and the `dust-*` global agents. Where the old model had its own global agent, retire it via `RETIRED_GLOBAL_AGENTS_SID` rather than deleting the `GLOBAL_AGENTS_SID` member — the enum values are the `sId`s persisted in historical messages. That mechanism postdates the DeepSeek commit, which deleted the member; the skill now says to prefer retirement.
- *Path 2, decommissioned*: full removal, with the pricing entry **moved** into `LEGACY_MODEL_PRICING` rather than deleted, so historical runs still cost out. Narrowing `KnownModelLLMId` is called out as a breaking public-API change needing sign-off.
- *The agent-config repoint migration*: hardcoded ids when the source config is deleted in the same PR (the script must stay a frozen snapshot), the workspace-isolation bypass on the scan only, and a single batched `update` instead of one per row.

Also corrected a pre-existing inaccuracy surfaced while verifying: `sdk_drift.test.ts` is a compile-time assertion whose `it()` body always passes, so a green vitest run proves nothing — `tsgo` is what enforces front ⊆ SDK, and the SDK types must be rebuilt before it runs.

## Tests

Docs-only; no code changed. Verification done while writing:

- Every fully-qualified path in the skill resolves on disk, and every symbol it names (`LEGACY_MODEL_PRICING`, `RETIRED_GLOBAL_AGENTS_SID`, `isRetiredGlobalAgent`, `STATIC_MODEL_TIERS`, `ModelStaticWorkspaceAware`, …) still exists.
- The corrected Verify command runs green: `sdk_drift.test.ts`, `types.test.ts`, `model_tiers.test.ts` — 22 tests passing.
- One command in an early draft (`admin/cli.sh migrate …`) was invented and has been removed; the skill now only states `makeScript`'s verified `--execute` behavior.

### DIFF
--- a/.claude/skills/dust-llm/SKILL.md
+++ b/.claude/skills/dust-llm/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: dust-llm
-description: Step-by-step guide for adding support for a new LLM in Dust. Use when adding a new model, or updating a previous one.
+description: Step-by-step guide for adding support for a new LLM in Dust, and for deprecating or removing the model it supersedes (including the agent-config repoint migration). Use when adding a new model, updating a previous one, or retiring a decommissioned one.
 ---
 
 # Adding Support for a New LLM Model
@@ -8,6 +8,10 @@ description: Step-by-step guide for adding support for a new LLM in Dust. Use wh
 This skill guides you through adding a newly released LLM to the **model_constructors +
 llms** stack (the endpoint-class router). It replaces the legacy `lib/api/llm/clients/*`
 router, which no longer exists.
+
+Adding a model is usually only half the task: the model it supersedes has to be retired in
+the same PR, and a model the provider has switched off needs its agents repointed. See
+[Deprecating or removing an old model](#deprecating-or-removing-an-old-model).
 
 ## Mental model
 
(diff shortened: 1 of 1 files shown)


.claude/skills/dust-llm/SKILL.md
```diff
@@ -276,12 +284,111 @@ npx tsgo --noEmit                        # whole-project type check
...
+- **Write once, batched**: a single
+  `update({ providerId, modelId }, { where: { id: agents.map((a) => a.id) } })` over the ids
+  already gathered — not `agent.update()` per row. Scoped to those ids, the update needs no
+  isolation bypass of its own (the cross-workspace scan already happened in the `findAll`).
+- Set `providerId` alongside `modelId` — the replacement often sits on a different provider.
+  Reset `reasoningEffort` too if the target does not support the effort the agent was on.
+
+`makeScript` defaults `execute` to `false`, so running the script with no flag is the dry
+run — read the log, confirm the agent count and the from → to ids, then re-run with
+`--execute` (`-e`) to write.
```
Nils-Fedrigo:
```suggestion
```

pmilliotte (approved):

PR #24934, opened 2026-04-28
[LLM Router] Add support for DeepSeek V4 Pro (Fireworks) + dust-deepseek

fontanierh:
## Description

Adds DeepSeek V4 Pro served via Fireworks.

- Model ID: `accounts/fireworks/models/deepseek-v4-pro`
- 1M context, 64k output budget, function calling, no vision
- Pricing: \$1.74/M input, \$3.48/M output, \$0.14/M cache read
- Reasoning kept at `none` for now. The model exposes `thinking` / `reasoning_effort` per DeepSeek API docs but our LLM router does not translate those yet (same TODO as V3.2). Follow-up.

Also:
- Demotes V3.2: still defined for backward compatibility (whitelist, pricing, registry kept) but removed from `USED_MODEL_CONFIGS` and flagged `isLegacy: true`. New users get V4 Pro instead.
- Adds `dust-deepseek` internal global agent (single variant for now, no -medium/-high until reasoning is wired up).

## Tests

`npx tsgo --noEmit` passes. LLM integration test entry added with `runTest: false`, can be flipped on locally to validate against Fireworks.

## Risk

Low. V3.2 stays functional for existing usage. New agent is in `DUST_INTERNAL_AGENTS`, only visible to workspaces that already see the other dust-* internal agents.

## Deploy Plan

Standard deploy, no migrations.

PR #32009, opened 2026-09-07
feat(llm): add DeepSeek V4 Pro 0813 on Fireworks, deprecate the preview

Nils-Fedrigo:
## Description

Fireworks pulled `accounts/fireworks/models/deepseek-v4-pro` from serverless — its model page now offers on-demand dedicated deployment only — so every agent pointing at it has been failing for about a week. DeepSeek shipped `deepseek-v4-pro-0813` as the official replacement for that preview.

- Add DeepSeek V4 Pro 0813 across the model config, `model_constructors` and `llms` layers, plus pricing, tiers, reasoning efforts, the SDK union, the core tool-support whitelist and the model picker. Specs and pricing verified 2026-09-07 against the Fireworks model page and DeepSeek's own docs (1040k context, 131k max output per response on Fireworks, $1.32/$3.96/$0.044 per 1M).
- Serve it over the Fireworks **Responses** API. #31485 moved the rest of the Fireworks fleet across but deliberately kept V4 Pro on Chat Completions "until its active configurations can be moved to an approved replacement model ID". 0813 is that id, so it starts on the fixed transport and picks up the optional-tool-args fix rather than needing a second migration later.
- Deprecate the preview: `isLegacy`, renamed "DeepSeek V4 Pro (preview)", dropped from the picker. It stays in the registry, priced, with its endpoint registered, so historical token accounting still resolves and any straggler row fails as a model error rather than an unknown-model crash.
- Point `dust-deepseek` at the new model.
- Migrate agent configurations off the dead model (`front/migrations/20260907_migrate_deepseek_v4_pro_0813.ts`).

Two decisions worth a look:

**Reasoning ladder.** The preview exposed `none` as its only effort, which was never a product statement — it was a placeholder for "reasoning not wired up yet" (#24934) that omitted `reasoning_effort` and let Fireworks fall back to its `high` default. DeepSeek now documents low/high/max for Pro identically to V4 Flash, all three confirmed live, so 0813 exposes light/medium/high via `mapReasoningEffortToLowHighMax`, matching Kimi K3 and GLM-5.3 Flash. `forceHighReasoningEffort` is no longer needed and has only the preview left as a consumer.

**Where migrated agents land.** On this ladder, DeepSeek's `high` — what the preview actually ran at — is our **`medium`**, so that is what the migration writes. `high` would map to DeepSeek `max` and silently buy every migrated agent more reasoning tokens at $3.96/1M output.

Also adds `front/lib/llms/CONTRACTS` with one contract: a replacement model must preserve the effort surface of the model it replaces, or ship a migration rewriting the stored `reasoningEffort`. That is the trap this PR had to steer around, and it is not otherwise written down anywhere.

## Tests

Characterized live against Fireworks on 2026-09-07 — the full 46-case endpoint suite, 46/46 green. It confirms DeepSeek's documented effort mapping applies to Pro exactly as to Flash: `low`, `high` and `maximal` all reason, and the gateway-only `minimal`/`medium`/`xhigh` are rejected by our schema before any request goes out.

One case is asserted as `SUCCESS` rather than `TOOL_CALL_CALCULATOR`: at temperature 0.1 with tools offered and no forced `tool_choice`, 0813 usually does the arithmetic itself instead of calling the calculator, but not consistently (2 of 3 runs skipped the tool, 1 used it). An intermittent model choice cannot be asserted either way, so the case asserts only that the call succeeds — same treatment as Kimi K3, which does this deterministically. The four `force-tool` cases cover the tool-calling path itself and all pass.

Note the suite's `r-default` exercises the endpoint schema default (`high`), while the product default is now `light` → DeepSeek `low`, so there is no signal on unforced tool-choice at the effort most traffic will actually use.

Also green: `tsgo`, `sdk_drift.test.ts`, `types.test.ts`, `model_tiers.test.ts`, and the Fireworks Chat Completions client test (still exercised by the preview endpoint).

## Risk

Moderate — a transport change bundled with a model swap, on agents that are currently broken anyway. The 46-case live suite is the mitigation; every reasoning effort, forced and unforced tool use, JSON-schema output and prompt caching passed over Responses.

The migration is not scoped to `status: "active"`, unlike the DeepSeek R1 one. `restoreAgentConfiguration` un-archives the latest row **in place** rather than writing a new one, so a deleted agent restored after this ships would otherwise come back pointing at a model Fireworks no longer serves; `draft` rows back the builder's "try" button and have the same problem. The tradeoff is that superseded version rows get rewritten too, so version history will show the new model id for versions that ran on the preview. Token accounting is unaffected — that comes from run and message records, not `agent_configurations`.

Rollback is the revert commit plus repointing the migrated rows back, though the preview will still be dead upstream.

## Deploy Plan

Standard deploy, then run the migration:

```
npx tsx front/migrations/20260907_migrate_deepseek_v4_pro_0813.ts            # dry run
npx tsx front/migrations/20260907_migrate_deepseek_v4_pro_0813.ts --execute
```

It logs a per-status breakdown and one line per row before writing, so the dry run shows exactly what will move. One batched UPDATE, no schema change.

Afterwards, watch Fireworks provider errors and confirm `dust-deepseek` responds. Once the preview has no rows left it can be deleted outright, which also removes the last users of `forceHighReasoningEffort` and the Fireworks Chat Completions client.


core/src/providers/fireworks.rs
```diff
@@ -29,6 +29,7 @@ const MODEL_IDS_WITH_TOOLS_SUPPORT: &[&str] = &[
     "accounts/fireworks/models/kimi-k2-instruct-0905",
     "accounts/fireworks/models/deepseek-v3p2",
     "accounts/fireworks/models/deepseek-v4-pro",
+    "accounts/fireworks/models/deepseek-v4-pro-0813",
```
pmilliotte: I don't think we need this

pmilliotte (approved):

adrsimon (approved):

PR #32014, opened 2026-09-07
[front] fix: report monthly active users from consumption analytics

aubin-tchoi:
## Description

Context in [thread](https://dust4ai.slack.com/archives/C09M7UQBLHM/p1788793233988219).

There is an edge case where the billed MAU count differs from the active user count displayed in the analytics. The ES index does not record Sidekick usage, whereas the seat count report does (it counts mentions). There is an edge case where if a user only uses Sidekick during a cycle it will only be counted in MAUs but won't appear in analytics.

This PR moves the MAU reporting to rely on the ES index rather than counting mentions. The goal is to align on a single source of truth and make sure we provide observability on what is actually billed.

There is another theoretical difference: if a user initiated a message before the end of the cycle but ended it after the cycle (e.g. waited on a tool approval), its billing will move to the next cycle.

## Tests

- Tested locally.

## Risk

- Medium. Changes Stripe MAU reporting.

## Deploy Plan

- Deploy front.


front/lib/plans/usage/mau.ts
```diff
@@ -22,33 +42,74 @@ async function countActiveUsersForPeriodInWorkspace({
...
+          range: {
+            [COMPLETED_AT_FIELD]: {
+              gte: since.toISOString(),
+              lt: (to ?? new Date()).toISOString(),
+            },
+          },
+        },
+      ],
+      // Exclude subagent responses so they don't inflate the user's message count.
+      must_not: [{ exists: { field: "parent_message_id" } }],
```
aubin-tchoi, line 59: not great in terms of performances but hard to avoid here. we can monitor and change the approach if too costly

tdraier (approved):
LGTM

PR #32017, opened 2026-09-08
[front,connectors] Exclude originating Slack thread from retrieval

tdraier:
## Description

Fixes https://github.com/dust-tt/tasks/issues/10294

Asked from inside a Slack thread to find an *earlier* thread, the bot returned a link to the **current** thread. The active Slack thread is indexed into the data source with the user's question text verbatim (as a document tagged `threadId:<ts>`), so it self-matches on its own query and gets served back as the answer. Its content is already provided to the conversation as a content fragment, so retrieving the indexed copy only ever hurts.

This adds a generic, channel-agnostic retrieval-scope hint on the user message context: **`excludedRetrievalTags`** — opaque `key:value` tag strings a run asks retrieval to skip (merged into each data source's `filter.tags.not`). The conversation/message models stay entirely Slack-agnostic; the **only** component that knows the Slack→tag mapping is the connectors Slack bot, which sets `excludedRetrievalTags: ["threadId:<threadTs>"]`.

- `sdks/js` + `front` `UserMessageContext`: new optional `excludedRetrievalTags`.
- `front-api`: forwarded on post-message and create-conversation; swagger schemas updated.
- `front`: persisted via new nullable column `userContextExcludedRetrievalTags` on `user_messages` (pre-deploy migration), read back into the message context.
- `front`: both query-based search tools (`search` and `data_sources_file_system/search`) exclude the tags. New `@cc` contract `excluded-retrieval-tags-honored` documents the invariant for future retrieval tools.
- `connectors`: Slack bot stamps the originating thread's tag.

## Tests

- `tsgo --noEmit` clean on front, front-api, connectors; SDK build passes.
- `messages.test.ts` + `conversation/messages.test.ts` — 40/40 pass, exercising the new column's persist→readback roundtrip.
- Biome clean on all changed files.

## Risk

Low. The field is optional everywhere and defaults to no exclusions, so non-Slack runs are unaffected. The exclusion is applied post-`checkConflictingTags` at `finalTagsNot`, and the tags only exist on documents that carry them, so excluding them on other data sources is a no-op. Safe to roll back (additive nullable column, expand-only).

## Deploy Plan

1. Run the pre-deploy migration `20260907175148_add_excluded_retrieval_tags_to_user_messages.sql` (adds nullable column) **before** deploying code.
2. Deploy front / front-api / connectors (rebuilt `@dust-tt/client` is bundled via the workspace file dependency). Field is optional in all directions, so relative deploy order does not matter.

### DIFF
--- a/front/lib/models/agent/conversation.ts
+++ b/front/lib/models/agent/conversation.ts
@@ -313,6 +313,7 @@ export class UserMessageModel extends WorkspaceAwareModel<UserMessageModel> {
   declare userContextLastTriggerRunAt: Date | null;
   declare userContextApiKeyId: ForeignKey<KeyModel["id"]> | null;
   declare userContextAuthMethod: string | null;
+  declare userContextExcludedRetrievalTags: string[] | null;
 
   declare userId: ForeignKey<UserModel["id"]> | null;
   // Denormalized from messages for conversation-scoped fetches (plain column, no FK — the value
@@ -390,6 +391,11 @@ UserMessageModel.init(
     userContextAuthMethod: {
       type: DataTypes.STRING(50),
       allowNull: true,
+    },
+    userContextExcludedRetrievalTags: {
+      type: DataTypes.ARRAY(DataTypes.STRING),
+      allowNull: true,
+      defaultValue: null,
     },
     agenticMessageType: {
       type: DataTypes.STRING(16),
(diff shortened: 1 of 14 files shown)


tdraier:

- #32017 ◀

front/lib/models/agent/conversation.ts
```diff
@@ -392,6 +393,11 @@ UserMessageModel.init(
       type: DataTypes.STRING(50),
       allowNull: true,
     },
+    userContextExcludedRetrievalTags: {
```
flvndvd: Do we really need to store this in DB 🤔?

tdraier: mm not really - we should be able to pass it through AgentLoopArgs

PR #32020, opened 2026-09-08
Remove otherConsumedCredits from AWU pool summary

avervaet:
## Description
Drop the derived "other" bucket (and the pool-consumption-by-member computation that only existed to power it) from the AWU pool current-cycle response, type, and UI hint.

Other was use for mapping gaps between real user usage + programmatic usage vs billed usage. It allowed us to find case for which we were billing user_usage with no user. This have been patched.

## Risk

Low this page was never reachable by an user

## Deploy Plan

- Deploy front


avervaet:

- #32020 ◀

aubin-tchoi (approved):
LGTM

PR #32023, opened 2026-09-08
[front] Spend caps: enforce default free-seat allowance when grant not provisioned yet

tdraier:
## Description

Under the rate-limiter spend caps, a free seat's lifetime allowance threshold is read from its Metronome per-user free-seat credit grant. That grant is provisioned asynchronously (via the `credit.segment.start` webhook / backfill), so a brand-new free seat can hit enforcement **before** its grant exists. Today the resolver returns `null` in that window and enforcement **fails open** — while the members-usage "Your Credits" display already falls back to the seat-type constant (`FREE_SEAT_LIFETIME_AWU_CREDITS = 500`). So a not-yet-granted free seat shows a 500 cap in the UI but is enforced at nothing, and the fail-open path logs `logger.error` (`[FreeSeatLifetime] … no free-seat allowance resolved; failing open`), which was polluting the fail-open error signal watched during the FF rollout.

This PR:

- **`getCachedCustomerPerUserCreditBalances`** now returns `{ balances, locked }`. `locked: true` distinguishes the `skipIfLocked` empty-map case (allowance momentarily *unknown*) from a genuinely-empty result. Display call sites updated to `.value.balances` (behavior unchanged).
- **`getFreeSeatLifetimeAllowanceAwuCredits`** returns a discriminated `FreeSeatAllowanceResolution` (`resolved` / `no-metronome` / `locked` / `no-grant` / `read-error`). A shared `resolveFreeSeatAllowance` centralizes the fail-open logging so cap and warning level each case identically:
  - `locked` → `debug` (transient, self-heals — off the rollout error signal)
  - `no-grant` → `warn`
  - `read-error` → `error` (this case previously logged **nothing** — the actually-dangerous silent one)
  - `no-metronome` → silent
- **`no-grant` now enforces the default `FREE_SEAT_LIFETIME_AWU_CREDITS`** instead of failing open, mirroring the display fallback, so a free seat whose grant hasn't landed yet is still capped at the default rather than spending unbounded. `locked` and `read-error` keep failing open (we don't know the true grant — could be an admin-raised amount above the default — so we don't over-enforce).

## Tests

- New `front/lib/api/users/spend_limit.test.ts` covering `resolveFreeSeatAllowance`: `resolved` passthrough, `no-grant` → default constant + `warn`, `locked` → `null` + `debug`, `read-error` → `null` + `error`, `no-metronome` → `null` + silent. 5/5 green.
- `tsgo --noEmit` clean; `biome check` clean.

## Risk

Low. Behavior changes only for the `no-grant` case, and only ever adds a cap where there was none (default 500) — matching what users already see in the UI. `locked`/`read-error` still fail open, so no new over-enforcement on transient/failed reads. Fully isolated diff (3 files + 1 test), based on `main`.

## Deploy Plan

Standard. Reduces `logger.error` noise from the free-seat fail-open path during the `enforce_user_spend_limit_rate_cap` rollout and closes the display/enforcement divergence for not-yet-provisioned free seats.

### DIFF
--- a/front/lib/api/credits/members_usage.ts
+++ b/front/lib/api/credits/members_usage.ts
@@ -889,7 +889,7 @@ async function fetchFreeSeatCreditsForMembersTable({
     for (const [
       userId,
       { balanceAwu, startingBalanceAwu },
-    ] of perUserCreditBalances.value) {
+    ] of perUserCreditBalances.value.balances) {
       freeBalanceByUserId.set(userId, balanceAwu);
       freeStartingByUserId.set(userId, startingBalanceAwu);
     }
@@ -1656,7 +1656,7 @@ export async function getMemberUsage({
       contractCreditType: CONTRACT_CREDIT_TYPE_FREE_SEAT,
     });
     if (balances.isOk()) {
-      const entry = balances.value.get(userId);
+      const entry = balances.value.balances.get(userId);
       if (entry) {
         freeSeatBalanceAwu = entry.balanceAwu;
         freeSeatAllowanceAwu = entry.startingBalanceAwu;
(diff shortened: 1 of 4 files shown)


fabiencelier (approved):
LGTM 
Needs some formatting before merging

PR #32024, opened 2026-09-08
[front] fix: report agent edit permissions for API keys

aubin-tchoi:
## Description

Closes https://github.com/dust-tt/tasks/issues/9336.

Agents created or edited with a workspace API key return `canEdit: false` on subsequent reads because the field is populated as follows `isAuthor || isMember`, which reflects well the access of a user authenticated via the webapp, but not the access granted by an API key (you can read the agent even if you're not an author or a member).

This PR computes `canEdit` for API keys using the same logic as in the PATCH endpoint for agents.

## Tests

- Added tests.

## Risk

- Low.

## Deploy Plan

- Deploy front.


front/lib/api/assistant/configuration/helpers.ts
```diff
@@ -196,7 +206,12 @@ export async function enrichAgentConfigurations<V extends AgentFetchVariant>(
     const isMember = editorIds.includes(agent.id);
 
     const canRead = isAuthor || isMember || agent.scope === "visible";
-    const canEdit = isAuthor || isMember;
+    const canEdit = auth.isKey()
+      ? auth.isBuilder() &&
+        agent.status === "active" &&
+        (canRead || auth.isAdmin()) &&
+        canReadRequestedSpaces(auth, spaceById, agent.requestedSpaceIds)
```
davidebbo, line 227: So iiuc, if it's a key from a non-admin builder that just has read access, we treat this is `canEdit=true`? Not very intuitive why, but I'm probably misreading it.

front/lib/api/assistant/configuration/helpers.ts
```diff
@@ -196,7 +206,12 @@ export async function enrichAgentConfigurations<V extends AgentFetchVariant>(
     const isMember = editorIds.includes(agent.id);
 
     const canRead = isAuthor || isMember || agent.scope === "visible";
-    const canEdit = isAuthor || isMember;
+    const canEdit = auth.isKey()
+      ? auth.isBuilder() &&
+        agent.status === "active" &&
```
davidebbo, line 225: Why don't we need to check `agent.status === "active"` in the non-key case?

davidebbo (approved):
LGTM

front/lib/api/assistant/configuration/helpers.ts
aubin-tchoi: all builder API keys can actually update agents AFAICT, indeed somewhat surprising but I assume we don't want to break flows here

front/lib/api/assistant/configuration/helpers.ts
aubin-tchoi: we actually do rely on the canEdit in the restore endpoint (can't restore if you can't edit)

PR #32027, opened 2026-09-08
Show and edit the workspace default limit in EditMemberSpendLimitModal

avervaet:
## Description

Stack 4/4, on #32151.
- When the workspace default is the member's applicable limit, the modal shows a "Workspace default limit" field. Only for pool-bearing seats: the "default" source also labels free and none seats, whose cap is the seat allowance (or 0), not the workspace default.
- Admins edit it. Managers see it locked with a tooltip explaining only admins can change it, while personal and group limits stay editable for them. Poke stays read-only.
- Value is fetched by the caller (usage page or poke) and passed in as a single state: loading, error, unavailable, or ready. Saving is blocked while it loads only for viewers who can edit it, so a manager is never held back by that field. A failed fetch locks only that field with a message, the other limits can still be saved. Poke passes "unavailable" for workspaces without a default pool limit (not Metronome-only billed), which hides the field.
- The form no longer remounts when the default finishes loading, so values typed in the other fields meanwhile are kept.

## Tests

`EditMemberSpendLimitModal.test.tsx` (14 tests): admin editable, manager locked with personal limit still editable, read-only viewer, loading placeholder, loading blocks admin save but not manager save, typed value survives loading→ready, error locks only the default field, hidden when unavailable, hidden for free/none seats, hidden when a personal or group cap applies. Manual: edit the default as admin, row updates; locked with tooltip as manager.

## Risk

Low, additive and edit scoped to admins.

## Deploy Plan

Deploy front.

### DIFF
--- a/front/components/pages/workspace/UsagePage.tsx
+++ b/front/components/pages/workspace/UsagePage.tsx
@@ -9,6 +9,7 @@ import {
   seatTypeDisplayName,
 } from "@app/components/workspace/billing/seatTypeUtils";
 import { ChangeSeatModal } from "@app/components/workspace/ChangeSeatModal";
+import type { DefaultUserSpendLimitState } from "@app/components/workspace/EditMemberSpendLimitModal";
 import { EditMemberSpendLimitModal } from "@app/components/workspace/EditMemberSpendLimitModal";
 import { EditSpendLimitModal } from "@app/components/workspace/EditSpendLimitModal";
 import { GroupModelTierPickerDropdown } from "@app/components/workspace/GroupModelTierPickerDropdown";
@@ -313,11 +314,17 @@ export function UsagePage() {
     useState<MemberUsageType | null>(null);
   const [spendLimitRecapMember, setSpendLimitRecapMember] =
     useState<MemberUsageType | null>(null);
-  const { defaultUserSpendLimit, isDefaultUserSpendLimitLoading } =
+  const { defaultUserSpendLimit, isDefaultUserSpendLimitError } =
     useDefaultUserSpendLimit({
       workspaceId: owner.sId,
       disabled: spendLimitRecapMember === null,
     });
+  const defaultUserSpendLimitState: DefaultUserSpendLimitState =
+    defaultUserSpendLimit
+      ? { status: "ready", awuCredits: defaultUserSpendLimit.awuCredits }
+      : isDefaultUserSpendLimitError
+        ? { status: "error" }
+        : { status: "loading" };
   const [
     totalAllowedUsagePendingMemberIds,
     setTotalAllowedUsagePendingMemberIds,
@@ -1453,8 +1460,7 @@ export function UsagePage() {
           groups={groups}
           readOnly={!isManager(owner)}
           canEditDefaultLimit={isWorkspaceAdmin}
-          defaultUserSpendLimitAwuCredits={defaultUserSpendLimit?.awuCredits}
-          isDefaultUserSpendLimitLoading={isDefaultUserSpendLimitLoading}
+          defaultUserSpendLimit={defaultUserSpendLimitState}
         />
 
         <BulkEditSpendLimitModal
--- a/front/components/pages/workspace/UsagePage.tsx
+++ b/front/components/pages/workspace/UsagePage.tsx
@@ -77,7 +77,10 @@ import {
   useResolveUpgradeRequest,
   useUpgradeRequests,
 } from "@app/lib/swr/upgrade_requests";
-import { useUsageSettings } from "@app/lib/swr/usage_settings";
+import {
+  useDefaultUserSpendLimit,
+  useUsageSettings,
+} from "@app/lib/swr/usage_settings";
 import type { ModelsTierName } from "@app/types/assistant/models/model_tiers";
 import { CAP_ELIGIBLE_GROUP_KINDS } from "@app/types/groups";
 import type {
@@ -310,6 +313,11 @@ export function UsagePage() {
     useState<MemberUsageType | null>(null);
   const [spendLimitRecapMember, setSpendLimitRecapMember] =
     useState<MemberUsageType | null>(null);
+  const { defaultUserSpendLimit, isDefaultUserSpendLimitLoading } =
+    useDefaultUserSpendLimit({
+      workspaceId: owner.sId,
+      disabled: spendLimitRecapMember === null,
+    });
   const [
     totalAllowedUsagePendingMemberIds,
     setTotalAllowedUsagePendingMemberIds,
@@ -1444,6 +1452,8 @@ export function UsagePage() {
           owner={owner}
           groups={groups}
           readOnly={!isManager(owner)}
+          defaultUserSpendLimitAwuCredits={defaultUserSpendLimit?.awuCredits}
+          isDefaultUserSpendLimitLoading={isDefaultUserSpendLimitLoading}
         />
 
         <BulkEditSpendLimitModal
--- a/front/components/pages/workspace/UsagePage.tsx
+++ b/front/components/pages/workspace/UsagePage.tsx
@@ -97,6 +97,7 @@ import {
 import {
   isCreditPricedPlan,
   isSubscriptionCancellationScheduled,
+  isSubscriptionMetronomeBilled,
 } from "@app/types/plan";
 import { isAdmin, isManager } from "@app/types/user";
 import {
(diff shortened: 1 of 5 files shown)


avervaet:

- #32020
    - #32027 ◀

front/components/workspace/EditMemberSpendLimitModal.tsx
```diff
@@ -275,18 +345,26 @@ export function EditMemberSpendLimitModal({
     (groupName) => groups.some((g) => g.name === groupName)
   );
 
+  const { defaultUserSpendLimit, isDefaultUserSpendLimitLoading } =
+    useDefaultUserSpendLimit({
+      workspaceId: owner.sId,
+      disabled: !isOpen || readOnly,
```
PopDaph: Quick question on the read-only path: the fetch is skipped with `disabled: !isOpen || readOnly`, and the value falls back with `?? 0`. Doesn't that render "Workspace default limit: 0" for read-only viewers (poke `PoolUsagePage`, non-admins on `UsagePage`)? 

For a member whose `spendLimitSource` is `default`, the row also gets the "Highest" chip, so it reads as "the limit that applies is 0"?

front/components/workspace/EditMemberSpendLimitModal.tsx
```diff
@@ -210,6 +245,41 @@ function MemberSpendLimitForm({
       </DialogHeader>
       <DialogContainer>
         <div className="flex flex-col gap-5">
+          <Page.Vertical gap="xs" align="stretch">
```
PopDaph: This looks very similar to MemberPersonalLimitInput, shoudln't we do a common component?

front/components/workspace/EditMemberSpendLimitModal.test.tsx
```diff
@@ -0,0 +1,94 @@
...
+}));
+
+vi.mock("@app/lib/swr/groups", () => ({
+  useUpdateGroupSpendLimit: () => ({ doUpdateGroupSpendLimit: vi.fn() }),
+}));
+
+const owner = { sId: "wId_1" } as LightWorkspaceType;
+const member = {
+  sId: "user_1",
+  name: "Ada Lovelace",
```
PopDaph: <3

front/components/workspace/EditMemberSpendLimitModal.test.tsx
```diff
@@ -0,0 +1,94 @@
...
+vi.mock("@app/lib/swr/memberships", () => ({
+  useUpdateUserSpendLimit: () => ({ doUpdateSpendLimit: vi.fn() }),
+}));
+
+vi.mock("@app/lib/swr/groups", () => ({
+  useUpdateGroupSpendLimit: () => ({ doUpdateGroupSpendLimit: vi.fn() }),
+}));
+
+const owner = { sId: "wId_1" } as LightWorkspaceType;
+const member = {
```
PopDaph: Can you reuse your own helper makeMemberUsage?

front/components/workspace/member_spend_limit_helpers.ts
```diff
@@ -37,6 +41,26 @@ export function parseCreditsInput(
   return { ok: true, awuCredits: parsed };
 }
 
+// Unlike `parseCreditsInput`, the workspace default has no "unlimited" state:
+// it's always a concrete number, and an empty input means 0 (no pool access).
+export function parseDefaultLimitInput(
```
PopDaph: Nice to have this extracted. While you are at it, `handleSaveDefaultLimit` in `UsageSettingsCard.tsx` does the same integer/min/max check inline. Could it call this helper so the two default-limit editors stay in sync?

PopDaph:
Can you fill the test section so I know exactly how to test it? I deployed on a review app 🙏🏻

front-api/routes/w/[wId]/usage_settings/default_user_spend_limit.ts
```diff
@@ -75,9 +75,10 @@ function mapErrorToApiError(
 // Mounted at /api/w/:wId/usage_settings/default_user_spend_limit.
 const app = workspaceApp();
 
+// Any workspace member can read the default limit that applies to them;
+// only admins can change it (enforced on the PUT below).
```
Comment (PopDaph) [inline front-api/routes/w/[wId]/usage_settings/default_user_spend_limit.ts]: Dropping `ensureIsAdmin()` here opens the GET to every workspace member (`user` and `builder` included, since the workspace middleware only checks `isUser()`). I get why: managers now see the read-only modal and need the value, and admin-only would 403 them.

But the only non-admins who can reach the Usage page are managers (the SPA route sits under `RequireRoleLayout requiredRole="manager"` and the nav entry is `disabled: !hasManagerRole`), and we already have `ensureIsManager()` in `ensure_role.ts` that accepts admins + managers. Can we use that instead of removing the guard entirely?

[Meta] Worth a line in the PR description when you change permissions on an existing endpoint, the Risk section currently says "scoped to admins". 😄

front/components/workspace/member_spend_limit_helpers.ts
```diff
@@ -37,6 +41,26 @@ export function parseCreditsInput(
...
+  raw: string
+): { ok: true; awuCredits: number } | { ok: false; message: string } {
+  const trimmed = raw.trim();
+  const parsed = trimmed === "" ? 0 : Number(trimmed);
```
PopDaph: Clearing this field sets the workspace default to `0`, which means "no pool access" for every member on the default (not "no limit"). Right next to the personal input, where blank means "unlimited", that's an easy trap, especially in a modal titled "Edit spend limit for <member>".

Could we make empty invalid here? Admins who really want `0` can type it, and `UsageSettingsCard` gets the same protection since it shares the helper.

front/components/workspace/EditMemberSpendLimitModal.tsx
```diff
@@ -280,13 +336,15 @@ export function EditMemberSpendLimitModal({
...
-          key={`${displayedMember?.sId ?? "none"}:${isOpen}:${memberGroupsResolved}`}
+          // targeted member changes, once the member's groups resolve, or
+          // once the workspace default limit finishes loading, instead of
+          // syncing state from props.
+          key={`${displayedMember?.sId ?? "none"}:${isOpen}:${memberGroupsResolved}:${isDefaultUserSpendLimitLoading}`}
           member={displayedMember}
           owner={owner}
           groups={groups}
           readOnly={readOnly}
+          defaultLimitAwuCredits={defaultUserSpendLimitAwuCredits ?? 0}
```
PopDaph: Do we really want to display 0?

front/components/pages/workspace/UsagePage.tsx
```diff
@@ -1432,11 +1441,18 @@ export function UsagePage() {
 
         <EditMemberSpendLimitModal
           isOpen={spendLimitRecapMember !== null}
-          onClose={() => setSpendLimitRecapMember(null)}
+          onClose={() => {
+            setSpendLimitRecapMember(null);
+            // Refreshes the member row (limit, source, isSpendCapped) so the
+            // table and the Unblock button reflect the edit without a reload.
+            void mutateMembersUsage();
```
PopDaph: Why do we mutate onClose? If I click on cancel or if I'm read only I can make any change so no need to mutate unless I submitted the form no?

front/components/workspace/EditMemberSpendLimitModal.tsx
```diff
@@ -125,7 +144,16 @@ function MemberSpendLimitForm({
       )
     );
 
-    if (!personalResult.ok || groupResults.some(({ result }) => !result.ok)) {
+    const defaultLimitResult = parseDefaultLimitInput(defaultLimitInput);
```
PopDaph: Pre-existing, but this now repeats the same parse / validate / diff / task shape three times. Two quick wins: a `toSpendLimit(awuCredits: number | null)` helper (the `unlimited | limited` ternary is written twice), and the `if (!result.ok) continue` in the group loop is unreachable.

Longer term this wants react-hook-form + zod (see the `dust-react-hook-form-writer` skill). Fine as a follow-up.

avervaet:
@PopDaph splitting this pr into multiple so it is easier to review and the changes here stays scoped at adding the workspace default limit to the edit spend limit modal.

front/components/pages/workspace/UsagePage.tsx
```diff
@@ -310,6 +314,17 @@ export function UsagePage() {
     useState<MemberUsageType | null>(null);
   const [spendLimitRecapMember, setSpendLimitRecapMember] =
     useState<MemberUsageType | null>(null);
+  const { defaultUserSpendLimit, isDefaultUserSpendLimitError } =
+    useDefaultUserSpendLimit({
+      workspaceId: owner.sId,
+      disabled: spendLimitRecapMember === null,
```
PopDaph: Question: why not disabled on isSubscriptionMetronomeBilled?

front/components/pages/workspace/UsagePage.tsx
```diff
@@ -310,6 +314,17 @@ export function UsagePage() {
...
+    });
+  const defaultUserSpendLimitState: DefaultUserSpendLimitState =
```
PopDaph, line 327: Do we also need "unavailable" if !isSubscriptionMetronomeBilled?

PopDaph (approved):
TBH I am getting a bit lost in all this code, looks reasonable so approving already!
Make sure to get approval from design before we ship (ie before remove the ff) 🙏🏻

PR #31934, opened 2026-09-07
feat: add micro_usd to es consumption index

sfriquet:
## Description

Closes https://github.com/dust-tt/tasks/issues/10310

Add `micro_usd` field to `front.agent_message_consumption_analytics`. Index the per-run cost, similar to how we have `tokens. cost_micro_usd` in index `front.agent_message_analytics` that represents the per-message dollar cost.

The sum of `micro_usd` for a given `message_id` equals `tokens. cost_micro_usd`.

## Tests

Tested locally comparing a message with several llm and tool call

```
POST /front.agent_message_analytics/_search
{
  "query": {
      "bool": {
        "filter": [
          { "term": { "message_id":"jvU13pThFE" } }
        ]
      }
    },
  "_source": ["tokens.cost_micro_usd"]
}
 "hits": {
    "total": {
      "value": 1,
      "relation": "eq"
    },
    "max_score": 0,
    "hits": [
      {
        "_index": "front.agent_message_analytics_2",
        "_id": "AYjP6ks68g_jvU13pThFE_0",
        "_score": 0,
        "_source": {
          "tokens": {
            "cost_micro_usd": 1943
          }
        }
      }
    ]
  }
```

```
POST /front.agent_message_consumption_analytics/_search
{
  "query": {
      "bool": {
        "filter": [
          { "term": { "agent_message_id":"jvU13pThFE" } },
          { "term": { "consumption_type": "llm" } }
        ]
      }
    },
    "aggs": {
      "cost": {
        "sum": {
          "field": "micro_usd"
        }
        
      }
    }, 
  "_source": ["micro_usd"]
}
"hits": {
    "total": {
      "value": 4,
      "relation": "eq"
    },
    "max_score": 0,
    "hits": [
      {
        "_index": "front.agent_message_consumption_analytics_1",
        "_id": "AYjP6ks68g_jvU13pThFE_run-usage:491",
        "_score": 0,
        "_source": {
          "micro_usd": 446
        }
      },
      {
        "_index": "front.agent_message_consumption_analytics_1",
        "_id": "AYjP6ks68g_jvU13pThFE_run-usage:492",
        "_score": 0,
        "_source": {
          "micro_usd": 492
        }
      },
      {
        "_index": "front.agent_message_consumption_analytics_1",
        "_id": "AYjP6ks68g_jvU13pThFE_run-usage:493",
        "_score": 0,
        "_source": {
          "micro_usd": 406
        }
      },
      {
[...]

PR #32028, opened 2026-09-08
feat: hide microUsd in consumption export

sfriquet:
## Description

field added in #31934 that we have no reason to expose in export at this point. removing it from the export.

## Tests

downloaded export locally

## Risk

Low

## Deploy Plan

Front


aubin-tchoi (approved):
LGTM

PR #32033, opened 2026-09-08
fix(front): stop cropping Name/Group columns in Usage tables

tdraier:
## Description

Fixes dust-tt/tasks#10335. In **Admin > Usage**, the primary text column was getting cropped:

- **Members tab**: long names/emails were clipped (e.g. on a 1932px monitor the table container still caps at ~1072px, and every secondary column rendered into it, crushing Name to ~230px — or ~0 on the new/compact usage page).
- **Groups tab**: the Group name was cropped to "Gr…" on narrower containers (the fixed Members/Spend limit/Models tier columns left the Group column ~50px).

Root cause: responsive column dropping was keyed to **window width** (`columnsBreakpoints`), but these tables live in a max-width content area whose container is far narrower than the viewport. So columns never dropped and the flexible text column got squeezed.

Fix: switch both tables to **container-query-based** column visibility on each column's `meta.className` (`hidden @Nxl:table-cell`), driven by the `@container/table` context on `DataTable.Root` — the same pattern already used by `AssistantsTable`, `AdminActionsList`, and `APIKeysTable`. Columns now drop by the table's *actual* width, so the primary column always has room to display fully. Also removed the backwards `columnsBreakpoints={{ name: "md" }}` on the Groups table that hid the primary Group column below the md window.

## Tests

Manual, via Playwright against the local app at several widths (measured column widths directly):

- Members tab: Name goes from ~230px (legacy) / ~0px (compact) to **~430–450px** at the 1072px container; secondary columns drop cleanly as the container narrows and Name absorbs the space.
- Groups tab: "group test" now renders fully — Group column **309px** at a 429px container (was 53px, "Gr…"); Models tier drops, Members stays.
- Validated the crowded credit-priced layouts (which the dev workspace can't populate) with a synthetic render using the real compiled CSS at the 1072px cap.
- `tsgo --noEmit` clean; biome clean.

No unit tests exist for these presentational table components.

## Risk

Low. Presentational-only (Tailwind class changes); no logic, data, or API changes. Behavior is easy to eyeball and trivially reversible. Container-query utilities are already used by sibling tables in the codebase.

## Deploy Plan

Standard front deploy. No migrations, flags, or ordering constraints.

### DIFF
--- a/front/components/workspace/GroupsUsageTable.tsx
+++ b/front/components/workspace/GroupsUsageTable.tsx
@@ -78,7 +78,7 @@ export function GroupsUsageTable({
             {
               id: "cap",
               header: "Spend limit",
-              meta: { className: "w-64" },
+              meta: { className: "hidden @3xl:table-cell @3xl:w-64" },
               cell: (info: GroupInfo) => (
                 <GroupSpendLimitCell
                   group={info.row.original}
@@ -105,7 +105,7 @@ export function GroupsUsageTable({
                   <ModelTiersInfoButton />
                 </span>
               ),
-              meta: { className: "w-64" },
+              meta: { className: "hidden @2xl:table-cell @2xl:w-64" },
               cell: (info: GroupInfo) => (
                 <GroupModelTierPickerDropdown
                   owner={owner}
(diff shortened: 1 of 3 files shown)


tdraier:

- #32033 ◀

ykmsd (approved):

PR #32034, opened 2026-09-08
Move top-up button between cards and cycle history, cap history to 3 rows

avervaet:
## Description

In the new usage page only show up to 3 previous cycle so the table doesn't consume too much space.

## Risk

Low

## Deploy Plan

Deploy front


avervaet:

- #32034 ◀

PR #32036, opened 2026-09-08
Fix Unblock action never showing on the customer usage page

avervaet:
## Description

`getMembersUsage` only computed the `isSpendCapped`/`rateLimiterState` off-pace verdict when `includeAlertLinks` was true, which only poke ever passes. So the Unblock action on the new usage page was structurally impossible on the customer usage page. It's now gated on `includeAlertLinks || spendCapEnabled` instead.

## Risk

Low risk: the change only widens when an existing verdict is computed and leaves poke's behavior untouched. Rollback is a plain revert.

## Deploy Plan

Deploy front

### DIFF
--- a/front/components/pages/workspace/UsagePage.tsx
+++ b/front/components/pages/workspace/UsagePage.tsx
@@ -1113,7 +1113,20 @@ export function UsagePage() {
             />
           ) : (
             <div className="flex items-center justify-between">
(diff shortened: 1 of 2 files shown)


avervaet:

- #32036 ◀
    - #32040

tdraier (approved):

PR #25647, opened 2026-05-15
[Scrub] Jitter workspace scrub wake-up

philipperolet:
## Description
Related to https://github.com/dust-tt/tasks/issues/8102.

Adds deterministic 0-24h jitter before scheduled workspace scrubs call `scrubWorkspaceData`, so retention wake-ups do not all execute the heavy delete path at the same time. The new timer is guarded with a Temporal patch marker for replay safety. Immediate workspace scrubs remain immediate.

This does not completely solve the issue, other PRs will follow.

## Risks
Blast radius: scheduled workspace data scrubbing after retention.
Risk: low.

## Deploy Plan
- pmrr
- deploy front workers

---

<a href="https://stagereview.app/dust-tt/dust/pull/25647">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://stagereview.app/assets/gh-open-in-stage-dark.svg">
    Open in Stage
  </picture>
</a>

PR #32038, opened 2026-09-08
[Scrub] Deprecate workspace scrub jitter patch

philipperolet:
## Description

Related to https://github.com/dust-tt/tasks/issues/8122. Follows https://github.com/dust-tt/dust/pull/25647.

First of two cleanup deployments: replace `patched("workspace-scrub-jitter")` with `deprecatePatch`, keeping the existing delay of up to 24 hours and the eligibility check after it. Temporal still needs the marker to replay histories recorded with the current code.

Draft pending verification of production workflow histories. The May merge date alone does not establish readiness: workspace retention can be 1,000 days, and activities can keep retrying without a workflow execution timeout.

## Risks

Blast radius: scheduled workspace scrubs and replay of their workflow histories.
Risk: standard

Deploying while a replayable history still takes the pre-jitter path can cause Temporal replay failures.

## Deploy Plan

- Before merging, verify in every production region that no relevant history still takes the pre-jitter path. Use the actual rollout of #25647, inspect old running/stuck executions, and account for retained closed histories that may be replayed or reset. This check has not been performed.
- Deploy front workers (`scrub_workspace_queue`).
- In a later PR and deployment, remove `deprecatePatch` once all workers have this change and no history that may still be replayed contains the old, non-deprecated jitter marker. Use workflow histories rather than a fixed waiting period. See [Temporal's patch lifecycle](https://docs.temporal.io/develop/typescript/workflows/versioning#deprecating-patches).

## Tests

- [x] Existing jitter tests: 3 passed (`npm -w front run test -- temporal/scrub_workspace/jitter.test.ts`).
- [ ] Verify production history compatibility before merging.


PopDaph (commented):
👍🏻

PR #32045, opened 2026-09-08
Fix seat usage column sort on customer and poke usage pages

avervaet:
## Description

Seats without a personal usage percentage (workspace/none seat types) no longer outrank free/pro/max seats when sorting by seat usage, and the customer usage page now forwards `seatUsage` clicks to the server instead of silently falling back to sorting by name.

Extracted the seat-usage arithmetic (`computeSeatUsage`, `splitConsumedAwuCredits`) into a shared `front/lib/api/credits/seat_usage.ts` module, used by the response builders, the server-side sort, and the client table, so the three can't drift apart. The server-side sort by `seatUsage` now reuses the exact same inputs the response builder emits for each row, tiebreaks on pool/overage consumption before overage limit, and groups rows without a comparable percentage last regardless of sort direction.

## Tests

tested on front-edge

## Risk

Low

## Deploy Plan

Deploy front

### DIFF
--- a/front/lib/api/credits/seat_usage.ts
+++ b/front/lib/api/credits/seat_usage.ts
@@ -1,14 +1,12 @@
 import type { MembershipSeatType } from "@app/types/memberships";
+import { isSeatBased } from "@app/types/memberships";
 
 // Pure seat-usage arithmetic. Every place that derives a seat percentage or an
 // allowance/pool split must go through here so the numbers never drift apart.
 
-/**
- * @cc [owner:avervaet,label:product] allowance-first-split
- * `consumedFromAllowanceAwuCredits` is `totalConsumedAwuCredits` capped at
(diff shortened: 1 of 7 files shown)


avervaet:

- #32045 ◀

front/lib/api/credits/seat_usage.ts
```diff
@@ -0,0 +1,71 @@
...
+  const isFreeWithBalance =
+    seatType === "free" &&
+    typeof seatBalanceAwu === "number" &&
+    typeof memberUsageLimit === "number";
+  const consumed = isFreeWithBalance
+    ? Math.max(0, memberUsageLimit - seatBalanceAwu)
+    : consumedFromAllowanceAwuCredits;
+  return {
+    percent:
+      isSeatBased(seatType) && allowance > 0
```
aubin-tchoi: I think this one is not safe to use in the front-end as it `assertNever`s

aubin-tchoi (approved):
LGTM

PR #32047, opened 2026-09-08
[front] chore: remove dead sandbox_raw helper

avervaet:
## Summary
- `front/lib/api/files/sandbox_raw.ts` (`isSandboxRawDelimitedConversationFile`) has been unreferenced since "Remove JIT upsert in conversations" stripped the JIT-upsert-to-datasource code path that was its only caller, without deleting the helper itself.

Found via a manual dead-code sweep (grep-verified, cross-checked against git history), not an automated tool.

## Test plan
- [x] Confirmed no other file in the repo imports `isSandboxRawDelimitedConversationFile` or `sandbox_raw`


Fraggle (approved):
LGTM

PR #32041, opened 2026-09-08
Remove poke workOS fallback outside of dev.

Fraggle:
## Description

Restrict Poke's WorkOS session fallback to development environments. Production requests without a Cloudflare Access token now return `401`; development fallback sessions use `ALL_ROLES` and continue through `Authenticator.fromDustSuperUser`. Remove the GCS-backed per-user role lookup, cache, and related error handling from `front/lib/poke/roles.ts`.

## Tests

No automated tests were added or updated. Validate the middleware manually in development and confirm missing-token requests are rejected outside development.

## Risk

Medium security impact: production requests that previously fell back to WorkOS now fail closed. Rollback is straightforward by reverting the middleware and role-loader changes.

## Deploy Plan

Deploy `front` and `front-api`. Verify Cloudflare Access authentication in production and WorkOS fallback behavior in development after deployment.

PR #32048, opened 2026-09-08
Fix poke coupon creation 500 under Cloudflare Access auth

tdraier:
## Description

Creating a coupon from Poke in production fails with:

```
500 {"error":{"type":"internal_server_error","message":"Failed to create coupon: Unexpected unauthenticated call to `getNonNullableUser`."}}
```

Root cause: since #32041 ("Remove poke workOS fallback outside of dev"), Poke
requests in prod authenticate exclusively via the Cloudflare Access path, which
builds the `Authenticator` with a `pokePrincipal` (email/name) but **no Dust
`UserResource`** attached. `CouponResource.makeNew` called
`auth.getNonNullableUser()` to populate `createdByUserId`, which throws when no
user is attached. It worked in dev only because the dev-only WorkOS fallback
attaches a real user.

The `coupons.createdByUserId` column is already nullable, so the fix records the
user id when one is present and `null` otherwise (Cloudflare Access operators).

## Tests

Existing `front-api/routes/poke/coupons/index.test.ts` covers creation. No new
behavior beyond allowing a null author when the authenticator carries only a
poke principal.

## Risk

Low. Single-line change on a nullable column. Attribution (`createdByUserId`) is
now null for Cloudflare-Access-authenticated operators, which is acceptable
given the column already allows null. Other poke write paths calling
`getNonNullableUser()` may have the same latent issue and are out of scope here.

## Deploy Plan

Standard deploy. No migration required (column already nullable).

### DIFF
--- a/front/lib/resources/coupon_resource.ts
+++ b/front/lib/resources/coupon_resource.ts
@@ -66,11 +66,11 @@ export class CouponResource extends BaseResource<CouponModel> {
     { transaction }: { transaction?: Transaction } = {}
   ): Promise<Result<CouponResource, Error>> {
     try {
-      const user = auth.getNonNullableUser();
+      const user = auth.user();
       const coupon = await CouponModel.create(
         {
           ...body,
-          createdByUserId: user.id,
+          createdByUserId: user?.id ?? null,
         },
         { transaction }
       );


tdraier:

- #32048 ◀

Fraggle (approved):
Thx

PR #26241, opened 2026-05-27
Harden agent loop finalization when conversation or message is inaccessible.

Fraggle:
## Description

When `getAgentLoopData` hits an `isAgentLoopDataSoftDeleteError` (conversation or message soft-deleted while the agent loop is still running), `finalizeCancellation`, `finalizeInterruption`, and `finalizeGracefulStop` previously bailed out with a log line — leaving `AgentMessageModel.status` stuck at `"created"` and `ConversationModel.isRunningAgentLoop` stuck at `true`.

The fix adds `writeAgentMessageFinalStatusWhenInaccessible`: when the resource layer can't load the conversation, it resolves the workspace via `WorkspaceResource.fetchById`, looks up the `MessageModel` by `sId + version + workspaceId` to get the `agentMessageId`, then directly updates `AgentMessageModel.status + completedAt` and clears `ConversationModel.isRunningAgentLoop` (with `silent: true`). All three finalization paths are now wired to this function.

Also adds a one-off migration script (`20260526_cancel_stuck_created_agent_messages.ts`) to bulk-cancel existing stuck messages older than 7 days.

## Tests

Tested manually by tracing the code paths. The update queries both have a `status: "created"` guard to ensure idempotency. The migration script supports `--scanOnly` dry-run mode.

## Risk

Low. The new code path is only triggered when `isAgentLoopDataSoftDeleteError` is returned — the happy path is completely untouched. The `AgentMessageModel.update` is guarded by `status: "created"` so it won't overwrite already-finalized messages. Migration script is batched (1000/page) and idempotent.

## Deploy Plan

1. Deploy front.
2. Run migration script in dry-run: `ts-node front/migrations/20260526_cancel_stuck_created_agent_messages.ts`
3. Run with `--execute` once output looks correct.

PR #32049, opened 2026-09-08
[Agent Loop] Clean up unfinished runs after conversation deletion

philipperolet:
## Description

Fixes https://github.com/dust-tt/tasks/issues/8402.

If a user deletes a conversation while an agent is answering, the agent can notice the deletion and exit, while its saved message still says `created` and the conversation is still marked as running. The code previously just logged the deletion and returned.

This change marks that unfinished message as cancelled and clears the conversation's running flag when no other agent message is running. It covers both deletion noticed between steps and deletion noticed while stopping or handling an error. Repeated cleanup leaves finished messages and other running agents alone. The same cleanup applies when the loop loses access to its conversation.

The separate bug where deleting a message made later messages stay queued was fixed by https://github.com/dust-tt/dust/pull/29418. This PR addresses the remaining cleanup gap discussed in the closed https://github.com/dust-tt/dust/pull/26241. It does not sweep historical messages.

Context: [original Slack thread](https://dust4ai.slack.com/archives/C09GELMTTRT/p1779872037151629).

## Risks

Blast radius: agent loops exiting after their conversation or message becomes unavailable.
Risk: low. Cleanup targets the original message version and uses the conversation lock to preserve other running messages.

## Deploy Plan

- Deploy front

## Tests

- [x] Existing agent message finalization tests and new deletion regression tests: 29 tests passed.
- [x] Deletion noticed between steps and during cancellation finalizes the unfinished message.
- [x] Repeated cleanup preserves the completion time and another running message.
- [x] Cleanup cannot change a message in another workspace.

### DIFF
--- a/front/lib/resources/conversation_resource.ts
+++ b/front/lib/resources/conversation_resource.ts
@@ -1,3 +1,4 @@
+import { getConversationRankVersionLock } from "@app/lib/api/assistant/conversation/lock";
 import type { Authenticator } from "@app/lib/auth";
 import { ConversationMCPServerViewModel } from "@app/lib/models/agent/actions/conversation_mcp_server_view";
 import {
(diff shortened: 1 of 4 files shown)


front/lib/resources/conversation_resource.ts
```diff
@@ -3002,6 +3003,72 @@ export class ConversationResource extends BaseResource<ConversationModel> {
...
+      includeDeleted: true,
+      dangerouslySkipPermissionFiltering: true,
+    });
+    if (!conversation) {
+      return;
+    }
+
+    const workspaceId = auth.getNonNullableWorkspace().id;
+    await withTransaction(async (transaction) => {
+      await getConversationRankVersionLock(auth, conversation, transaction);
```
id13, line 3036: Is this lock really needed ?

id13:
LGTM @philipperolet, just one comment about the necessity of the lock

front/lib/resources/conversation_resource.ts
philipperolet: I presume so, to avoid races on setting the conversation run flag since this code can happen in a variety of edge cases; didn't dig deeper but since the lock was relatively short and rare, thought it was fine

PR #32050, opened 2026-09-08
Inline tool reference

ykmsd:


front/components/editor/extensions/shared/slash_suggestion/useSlashCommandCapabilities.ts
```diff
@@ -95,12 +95,8 @@ export function useInputBarSlashCommandCapabilities({
         query,
         skills,
         tools: serverViews,
-        toolFilter: (serverView) =>
```
ykmsd, line 98: you can inline tools/skills multiple times if you want (same as skills)

front/components/assistant/conversation/input_bar/InputBarContainer.tsx
```diff
@@ -185,6 +186,25 @@ function readDefaultSkillEditorState(editor: Editor): {
...
 }
 
+function hasAnotherAttachedNode(
+  editor: Editor,
+  nodeType: string,
+  attrName: string,
+  value: string
+): boolean {
+  let found = false;
+  editor.state.doc.descendants((node) => {
```
ykmsd, line 196: this is prose mirror thing and apparently you cannot stop the loop even if you find one

PR #32051, opened 2026-09-08
Fix Stripe cancellation for schedule-managed subscriptions

tdraier:
## Description

When switching a Stripe-billed workspace to a Metronome contract in Poke, the post-provision step that schedules the old Stripe subscription to cancel could fail with:

> The subscription is managed by the subscription schedule `sub_sched_...`, and updating any cancelation behavior directly is not allowed. Please update the schedule instead.

`scheduleSubscriptionCancellation` (`front/lib/plans/stripe.ts`) set `cancel_at` directly via `stripe.subscriptions.update`, which Stripe rejects for subscriptions driven by a subscription schedule. 

Adds a `@cc` code contract documenting the schedule-managed behavior.

## Risk

Low

## Deploy Plan

Standard `front` deploy. No migration, no config, no ordering constraints.


tdraier:

- #32051 ◀

PopDaph (approved):
Looks reasonable 👍🏻

PR #32052, opened 2026-09-08
Pods: alllow groups for members and editors of pods

fabiencelier:
## Description

closes https://github.com/dust-tt/tasks/issues/10299

Add groups to the way members can be manage in pods. Old section has been renamed "Individual members" and a new "Group members" section has been added.
Managed the same way as individual members: groups can be either a member or an editor of the pod.

image

image

## Tests

 - updated unit tests
 - tested locally

## Risk

blast radius: accessing and administrating skills
this work with the same mecanism as spaces, so risk is limited 

## Deploy Plan

deploy front

### DIFF
--- a/front/lib/spaces_utils.ts
+++ b/front/lib/spaces_utils.ts
@@ -65,11 +65,11 @@ export const isPrivateSpacesLimitReached = (
     plan.limits.vaults.maxVaults;
 
 /**
- * @cc [owner:fabiencelier,label:product] membership-dimensions-mirror-space
- * Returns the space's four membership dimensions as the update endpoint takes them, so a caller
- * that spreads the result and overrides only the dimensions it edits leaves the others unchanged.
+ * @cc [owner:fabiencelier,label:product] membership-properties-mirror-space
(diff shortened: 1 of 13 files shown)


front/lib/spaces_utils.ts
```diff
@@ -62,3 +63,17 @@ export const isPrivateSpacesLimitReached = (
   plan.limits.vaults.maxVaults !== -1 &&
   spaces.filter((s) => s.kind === "regular").length >=
     plan.limits.vaults.maxVaults;
+
+/**
+ * @cc [owner:fabiencelier,label:product] membership-dimensions-mirror-space
+ * Returns the space's four membership dimensions as the update endpoint takes them, so a caller
+ * that spreads the result and overrides only the dimensions it edits leaves the others unchanged.
+ */
+export const spaceMembershipDimensions = (space: RichSpaceType) => ({
```
Fraggle: very nit: the naming is weird to me (i would have Dimensions => Properties) but you can ignore me.

Fraggle (approved):
LGTM and thanks for carrying this one all the way!

PR #32053, opened 2026-09-08
Add CIMD support for remote MCP OAuth

tdraier:
## Description

fixes: https://github.com/dust-tt/tasks/issues/9873

Enterprise MCP servers that refuse Dynamic Client Registration (ServiceNow, Azure DevOps, Meta Ads, ZoomInfo) currently fail Dust's automatic OAuth and force every customer into per-instance Static OAuth setup (create an OAuth app, whitelist Dust's callback, paste client id/secret). 

This adds support for **CIMD (Client ID Metadata Documents)**, the MCP 2025-11-25 auth mechanism. When a server advertises `client_id_metadata_document_supported`, the client presents an HTTPS URL as its `client_id`; the server fetches that URL, reads the client metadata, and validates the redirect URI against it — no registration, no secret, no admin app. Servers like ServiceNow (Zurich P7+) accept CIMD where they reject DCR.

Changes:
- **New static document** `front/public/.well-known/oauth-client.json`, served as a static asset at `https://app.dust.tt/.well-known/oauth-client.json` (same path mechanism as `security.txt`; `.json` yields `application/json`). Its `client_id` is self-referential; `redirect_uris` lists the go-forward `app.dust.tt/oauth/mcp/finalize` plus the two legacy override hosts (`dust.tt`, `eu.dust.tt`).
- **`MCP_CLIENT_ID_METADATA_DOCUMENT_URL`** constant (one global value — CIMD requires a single stable client identity across all regions/cells, so it is deliberately not derived from per-region config), guarded by a `security;mcp` code contract requiring it to equal the served document's `client_id`.
- **`discoverOAuthMetadata()`** gains a CIMD branch that runs *before* DCR when the flag is advertised (and outside development, where the finalize redirect is localhost and absent from the published document): it returns a public-client metadata (`token_endpoint_auth_method: "none"`, no `client_secret`) with `client_id` set to the document URL and skips registration. DCR and the existing Static OAuth error remain the fallbacks. A `mcp` code contract documents this precedence.
- **No core change**: the Rust token exchange already forwards `client_id`.

The SDK (`@modelcontextprotocol/sdk@^1.29.0`) already parses `client_id_metadata_document_supported`, so no SDK bump and no raw metadata re-fetch are needed.

## Tests

- `cc-check format`/`list` pass for both new code contracts.
- JSON document validated.
- ⚠️ Typecheck/unit tests were not run in the authoring worktree (deps not installed there) — needs `npx tsgo --noEmit` (front) and MCP OAuth tests in a fully-installed env before merge.
- Post-deploy: verify end-to-end against the official ServiceNow MCP server (document fetch content-type + `redirect_uri` match).

## Risk

Low. The CIMD branch only triggers for servers that explicitly advertise the flag; all other servers keep the exact current DCR → Static OAuth behavior. No changes to `core` or to the Static OAuth path. The one correctness invariant (constant must equal the served document's `client_id`) is captured by a code contract.

## Deploy Plan

Standard. The static document ships with `front` and is served by front-spa at `app.dust.tt`. Follow-up (non-blocking): when PR #31913's legacy redirect-URI override is removed, trim `redirect_uris` in the document down to just `app.dust.tt`.

### DIFF
--- a/front/lib/api/mcp_server/urls.ts
+++ b/front/lib/api/mcp_server/urls.ts
@@ -6,6 +6,23 @@ import { EnvironmentConfig } from "@app/types/shared/utils/config";
 const MCP_OAUTH_PROXY_ENABLED = true;
 
 /**
+ * @cc [owner:tdraier,label:security;mcp] cimd-client-id-matches-document
+ * `MCP_CLIENT_ID_METADATA_DOCUMENT_URL` must be byte-for-byte identical to the
+ * `client_id` field of the document served at
+ * `front/public/.well-known/oauth-client.json`, and that document must remain
+ * reachable at this exact URL. Authorization servers fetch the URL and reject a
+ * document whose `client_id` differs from the URL it was fetched from, so any
+ * divergence silently breaks every CIMD-based MCP connection.
+ *
+ * The value is a single global constant (not derived from per-region config)
+ * because CIMD requires one stable client identity across all regions and
+ * cells; the redirect URI a given region sends is validated against the
+ * document's `redirect_uris`, which lists every region's finalize callback.
+ */
+export const MCP_CLIENT_ID_METADATA_DOCUMENT_URL =
+  "https://app.dust.tt/.well-known/oauth-client.json";
+
+/**
  * Whether MCP OAuth metadata/token/registration should be proxied through the
  * Dust MCP host. Enabled in development only; never active in production.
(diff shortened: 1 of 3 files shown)


tdraier:

- #32053 ◀

Fraggle (approved):
LGTM
I thought we already supported it.

PR #32059, opened 2026-09-08
Remove invite members button from usage page

avervaet:
## Description

Remove the add user button from the usage page. We would like user to use this functionality from the people tab and leave the scope of the usage page as tight as possible.

## Risk
- Low the feature already exit in the people page.

## Deploy Plan

- Deploy front


front/components/pages/workspace/UsagePage.tsx
PopDaph: Nit: can rename to searchRow 💅🏻

PopDaph: NIT: since there now just one children we could drop the wrapper?

PopDaph (approved):
🚀

PR #32060, opened 2026-09-08
[front] feat: add optional You.com web search provider

mouse-value-add:
## Description

This adds [You.com](https://you.com) as an optional web search provider, following the existing provider pattern in `front/lib/utils/websearch.ts` (SerpAPI / Serper / Firecrawl / Exa). I noticed the `webSearch()` switch already had a clean per-provider shape, so this reuses it rather than introducing anything new:

- `YouComParams { provider: "you_com"; api_key?: string }` added to the `SearchParams` union
- `youComSearch()` — plain `fetch` POST to `https://ydc-index.io/v1/search` with the `X-API-Key` header, mapping `results.web[]` to `SearchResultItem` (`title` / `link` / `snippet`). No new dependency, same style as `serpapiSearch` / `serperSearch`.
- `"you_com"` added to `WEB_SEARCH_PROVIDERS`, so a workspace can select it through the existing `webSearchProvider` metadata, and to the `set-web-providers` poke plugin enum
- `YDC_API_KEY` credential + `DUST_MANAGED_YDC_API_KEY` env var, mirroring the `EXA_API_KEY` / `DUST_MANAGED_EXA_API_KEY` pattern

Fully opt-in: the default provider is unchanged (`firecrawl` fallback in the websearch tool stays as-is); You.com is only used when a workspace explicitly sets `webSearchProvider: "you_com"` or a caller passes `provider: "you_com"`. A missing key fails closed with the same error shape as the other providers. No secrets are exposed; the key only ever comes from the dust-managed credentials or an explicit `api_key` param.

For context on where You.com also ships agent-side: the You.com MCP server is available at `https://api.you.com/mcp` (and a keyless `?profile=free` variant), packaged in [youdotcom-oss/agent-skills](https://github.com/youdotcom-oss/agent-skills) — not needed for this PR, just noting the ecosystem in case it's useful for docs elsewhere.

## Tests

- `biome check --error-on-warnings` on all 5 touched files: clean
- `tsgo --noEmit -p front/tsconfig.json`: 421 errors on this branch vs 421 on unmodified `main` (pre-existing, mostly `@dust-tt/sparkle` build artifacts and unrelated type noise); zero errors reference the touched files. Verified by stashing the change and re-running.
- Runtime sanity check of the `results.web` → `SearchResultItem` mapping (normal payload, missing `results`, non-array `results.web`, result without `url` dropped, missing title/snippet fallbacks): 3/3 cases pass
- Live API validation requires a key: `curl -X POST https://ydc-index.io/v1/search -H "X-API-Key: $YDC_API_KEY" -H 'Content-Type: application/json' -d '{"query":"...","count":5}'` with a key from https://you.com/platform/api-keys

## Risk

Low — purely additive union member + switch case + enum entries. Default behavior for existing providers is untouched; a workspace that never selects `you_com` sees no difference. Rollback is a plain revert.

## Deploy Plan

Standard front deploy. Self-hosted instances that want to use the provider set `DUST_MANAGED_YDC_API_KEY`; nothing else changes.

### DIFF
--- a/front/lib/api/poke/plugins/workspaces/set_web_providers.ts
+++ b/front/lib/api/poke/plugins/workspaces/set_web_providers.ts
@@ -21,6 +21,7 @@ export const setWebProvidersPlugin = createPlugin({
         values: [
           { label: "Firecrawl", value: "firecrawl" },
           { label: "Exa", value: "exa" },
+          { label: "You.com", value: "you_com" },
         ],
         multiple: false,
       },
--- a/front/lib/api/workspace.ts
+++ b/front/lib/api/workspace.ts
@@ -540,7 +540,7 @@ type WorkspaceKillSwitchValue =
   | typeof WorkspaceResource.FULL_WORKSPACE_KILL_SWITCH_VALUE
   | WorkspaceConversationKillSwitchValue;
 
-export const WEB_SEARCH_PROVIDERS = ["exa", "firecrawl"] as const;
+export const WEB_SEARCH_PROVIDERS = ["exa", "firecrawl", "you_com"] as const;
 export type WebSearchProvider = (typeof WEB_SEARCH_PROVIDERS)[number];
 
 export const WEB_BROWSE_PROVIDERS = ["exa", "firecrawl", "spider"] as const;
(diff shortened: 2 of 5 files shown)


mouse-value-add:
The Vercel previews (playground / storybook) are currently failing because the deploy is waiting on team authorization for this fork PR — could a maintainer authorize the Vercel deploy when you get a chance so CI can run? Everything else on my side (GitGuardian, Snyk) is green, and the branch is up to date with the base.

PR #32063, opened 2026-09-08
feat: show progress during conversation compaction

adrsimon:
## Description

Replace the static compaction spinner with a progress card docked above the composer. The card uses a fitted time curve, changes its status text for long runs, and shows success or retry states when the run ends.

Disable the composer while compaction runs so the control matches the blocked state. Keep completion state in the message cache and poll active compactions every five seconds. This lets a reloaded conversation recover when the completion event was missed.

Sometimes the bar is gonna end at 25%, sometimes it's gonna stay for long at 99%, but this is better than a bland "compacting..." message i believe.

## Tests

Locally, over a 25% full convo, triggered with /compact

Screenshot 2026-09-08 at 15 42 16

Screenshot 2026-09-08 at 15 42 28

Screenshot 2026-09-08 at 15 42 54

## Risk

The percentage estimates elapsed time rather than backend work. It stays below 100% until the backend reports completion. Polling runs only while a loaded compaction has `created` status, and reconciliation only advances that status to `succeeded` or `failed`.

## Deploy Plan

Deploy through the standard frontend release process. No migration or configuration change is required.

### DIFF
--- a/front/components/assistant/conversation/AgentInputBar.tsx
+++ b/front/components/assistant/conversation/AgentInputBar.tsx
@@ -171,10 +171,15 @@ export const AgentInputBar = ({ context }: AgentInputBarProps) => {
     .filter(isCompactionMessage)
     .findLast((message) => message.status === "created");
   const lastMessage = allMessages.at(-1);
-  const latestCompactionMessage =
-    lastMessage && isCompactionMessage(lastMessage) ? lastMessage : undefined;
+  // Keep the terminal state docked until the next message moves it into the transcript.
+  const latestTerminalCompactionMessage =
+    lastMessage &&
+    isCompactionMessage(lastMessage) &&
+    lastMessage.status !== "created"
+      ? lastMessage
+      : undefined;
   const dockedCompactionMessage =
-    activeCompactionMessage ?? latestCompactionMessage;
+    activeCompactionMessage ?? latestTerminalCompactionMessage;
   const isCompactionInProgress = activeCompactionMessage !== undefined;
   const { compact: retryCompaction, isCompacting: isRetryingCompaction } =
     useCompactConversation({
--- a/front/components/assistant/conversation/AgentInputBar.tsx
+++ b/front/components/assistant/conversation/AgentInputBar.tsx
@@ -1,4 +1,5 @@
 import { useBlockedActionsContext } from "@app/components/assistant/conversation/BlockedActionsProvider";
+import { CompactionProgress } from "@app/components/assistant/conversation/CompactionProgress";
 import { ContextUsageWarningBanner } from "@app/components/assistant/conversation/ContextUsageWarningBanner";
 import { useGenerationContext } from "@app/components/assistant/conversation/GenerationContextProvider";
 import { InputBar } from "@app/components/assistant/conversation/input_bar/InputBar";
@@ -21,6 +22,7 @@ import { WakeUpBanner } from "@app/components/assistant/conversation/WakeUpBanne
 import { PodJoinCTA } from "@app/components/pod/conversation/PodJoinCTA";
 import {
   useCancelMessage,
+  useCompactConversation,
   useConversation,
   useConversationContextUsage,
 } from "@app/hooks/conversations";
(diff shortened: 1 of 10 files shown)


front/components/assistant/conversation/CompactionProgress.tsx
```diff
@@ -0,0 +1,230 @@
...
+  message: Pick<CompactionMessageType, "created" | "status">;
+  canRetry: boolean;
+  isRetrying: boolean;
+  onRetry: () => void;
+}
+
+type CompactionProgressTone = "default" | "success" | "warning";
+
+const PROGRESS_CLASSES = {
+  default: "bg-muted-background [&>div]:bg-foreground",
```
aubin-tchoi: let's avoid these selectors and instead handle in the component

front/components/assistant/conversation/AgentInputBar.tsx
```diff
@@ -165,11 +167,27 @@ export const AgentInputBar = ({ context }: AgentInputBarProps) => {
...
-    (message) => isCompactionMessage(message) && message.status === "created"
-  );
+  const activeCompactionMessage = allMessages
+    .filter(isCompactionMessage)
+    .findLast((message) => message.status === "created");
+  const lastMessage = allMessages.at(-1);
+  const latestCompactionMessage =
+    lastMessage && isCompactionMessage(lastMessage) ? lastMessage : undefined;
+  const dockedCompactionMessage =
+    activeCompactionMessage ?? latestCompactionMessage;
```
aubin-tchoi: OOC when will we have the latest message be a compaction message with status !== created?

front/components/assistant/conversation/CompactionProgress.tsx
```diff
@@ -0,0 +1,230 @@
...
+  const curvePosition = Math.pow(
+    elapsedSeconds / CURVE_SCALE_SECONDS,
+    CURVE_POWER
+  );
+  return (CEILING_PERCENT * curvePosition) / (1 + curvePosition);
+}
+
+export function getCompactionProgressTier(
+  elapsedSeconds: number,
+  slowThresholdSeconds = SLOW_THRESHOLD_SECONDS
```
aubin-tchoi: nit: let's pass those two arguments within a structured object

front/components/assistant/conversation/CompactionProgress.tsx
```diff
@@ -0,0 +1,230 @@
...
+}: {
+  fillPercentage: number;
+  displayPercentage?: number;
+  stage: string;
+  stageAction?: ReactNode;
+  footer?: string;
+  tone: CompactionProgressTone;
+}) {
+  return (
+    <section className="mb-2 rounded-2xl border border-border bg-background px-4 py-3 md:px-5 md:py-4">
```
aubin-tchoi: can't we have the same padding horizontally and vertically?

front/components/assistant/conversation/CompactionProgress.tsx
```diff
@@ -0,0 +1,230 @@
...
+        label="Conversation compaction progress"
+        percentage={fillPercentage}
+        className={classNames(
+          "h-1 w-full [&>div]:transition-[width] [&>div]:duration-1000 [&>div]:ease-linear motion-reduce:[&>div]:transition-none",
+          PROGRESS_CLASSES[tone]
+        )}
+      />
+      <div
+        className={classNames(
+          "mt-2 flex items-start gap-2 text-sm",
```
aubin-tchoi: instead of handling margins on every div can we use a single gap so that the spacing is by default consistent?

front/components/assistant/conversation/CompactionProgress.tsx
```diff
@@ -0,0 +1,230 @@
...
+          <span className="shrink-0 text-sm tabular-nums text-muted-foreground">
+            {displayPercentage}%
+          </span>
+        )}
+      </div>
+      <ProgressBar
+        label="Conversation compaction progress"
+        percentage={fillPercentage}
+        className={classNames(
+          "h-1 w-full [&>div]:transition-[width] [&>div]:duration-1000 [&>div]:ease-linear motion-reduce:[&>div]:transition-none",
```
aubin-tchoi: let's avoid those selectors and instead handle the styling in the `ProgressBar` component itself (using an additional prop if we need one)

front/components/assistant/conversation/CompactionProgress.tsx
```diff
@@ -0,0 +1,230 @@
...
+      return (
+        <CompactionProgressCard
+          fillPercentage={fillPercentage}
+          stage="Compaction didn't finish. The conversation is unchanged."
+          stageAction={
+            <Button
+              label="Try again"
+              variant="ghost"
+              size="xs"
+              className="-mr-2 shrink-0"
```
aubin-tchoi: nit: this `-mr-2` looks hacky

front/components/assistant/conversation/ConversationViewer.tsx
```diff
@@ -559,6 +560,37 @@ export const ConversationViewer = ({
 
     const messagesFromBackend = messages.flatMap((m) => m.messages);
 
+    const refreshedCompactionsById = new Map(
+      messagesFromBackend
+        .filter(isCompactionMessageType)
+        .map((message) => [message.sId, message])
+    );
+
+    if (
```
aubin-tchoi, line 575: seems like we never update SWR here nor dispatch a `CompactionCompletedEvent` event, which could cause some side effects to be missed (for instance unblocking the input bar potentially)

PR #32064, opened 2026-09-08
feat: add read-only public API for agent triggers

adrsimon:
## Description

People asked for public API access to agent triggers. Adds admin-only, read-only `GET /w/{wId}/triggers` (list, filterable by kind/status/agent) and `GET /w/{wId}/triggers/{tId}`, mirroring the read access already granted on the automations page.

## Tests

Route tests cover the 403 for non-admin keys, default enabled-only filtering, kind filtering, and the webhookSource label enrichment.

## Risk

Low. New endpoints only, admin-gated, no writes.

## Deploy Plan

Deploy front-api.

### DIFF
--- a/front/lib/resources/trigger_resource.ts
+++ b/front/lib/resources/trigger_resource.ts
@@ -374,7 +374,18 @@ export class TriggerResource extends BaseResource<TriggerModel> {
     {
       kinds,
       executionModes,
-    }: { kinds?: TriggerKind[]; executionModes?: TriggerExecutionMode[] }
+      statuses,
+      agentConfigurationId,
+      limit,
+      offset,
+    }: {
+      kinds?: TriggerKind[];
+      executionModes?: TriggerExecutionMode[];
+      statuses?: TriggerStatus[];
+      agentConfigurationId?: string;
+      limit?: number;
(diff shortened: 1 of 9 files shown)


front/lib/resources/trigger_resource.ts
```diff
@@ -372,7 +374,14 @@ export class TriggerResource extends BaseResource<TriggerModel> {
...
       kinds,
       executionModes,
-    }: { kinds?: TriggerKind[]; executionModes?: TriggerExecutionMode[] }
+      limit,
+      offset,
+    }: {
+      kinds?: TriggerKind[];
+      executionModes?: TriggerExecutionMode[];
+      limit?: number;
+      offset?: number;
```
fabiencelier, line 383: maybe add a simple unit test that limit + offset work as expected on the resource ?

fabiencelier (approved):
LGTM ✅

front/lib/resources/trigger_resource.ts
adrsimon: isn't it testing sequelize actual logic ? i'm all in for tests but unsure this one would make sense

PR #32065, opened 2026-09-08
feat: expose creator id when listing skills/agents

sfriquet:
## Description

Closes https://github.com/dust-tt/tasks/issues/8867

Expose `creatorId` (sid) in:
- `api/v1/w/{{wId}}/assistant/agent_configurations?`
- `/api/v1/w/{{wId}}/skills`

## Tests
added uts + tested locally:

```
$ curl  'http://localhost:3000/api/v1/w/AYjP6ks68g/assistant/agent_configurations?withAuthors=true' | jq '[.agentConfigurations[] | {sId, name, creatorId}]'
[
  {
    "sId": "helper",
    "name": "help",
    "creatorId": null
  },
  {
    "sId": "dust",
    "name": "dust",
    "creatorId": null
  },
  {
    "sId": "deep-dive",
    "name": "deep-dive",
    "creatorId": null
  },
  {
    "sId": "094fe2d61b",
    "name": "TestAgent01",
    "creatorId": "ZWmML49W5v"
  }
]
```

```
$ curl 'http://localhost:3000/api/v1/w/AYjP6ks68g/skills' | jq '[.skills[] | {sId, name, creatorId}]'
[
  {
    "sId": "skl_Q8Ac4uf35L",
    "name": "Test",
    "creatorId": "ZWmML49W5v"
  },
  {
    "sId": "skl_M3noynRxIM",
    "name": "Test 2",
    "creatorId": "ZWmML49W5v"
  }
]
```

## Risk

Low

## Deploy Plan

Front only


front/types/assistant/skill_configuration.ts
```diff
@@ -68,6 +68,7 @@ export const SkillWithoutInstructionsAndToolsSchema = z.object({
   createdAt: z.number().nullable(),
   updatedAt: z.number().nullable(),
   editedBy: z.number().nullable(),
+  creatorId: z.string().nullable().optional(),
```
aubin-tchoi: FYI in zod ̀nullish() is equivalent to nullable().optional()

front/lib/api/assistant/configuration/creators.ts
```diff
@@ -0,0 +1,50 @@
...
+  const rows = await AgentConfigurationModel.findAll({
+    attributes: ["sId", "authorId"],
+    where: {
+      workspaceId: auth.getNonNullableWorkspace().id,
+      sId: { [Op.in]: sIds },
+      version: 0,
+    },
+  });
+
+  const creatorModelIdBySId = new Map<string, number>();
```
aubin-tchoi: nit (applies to a few other places)
```suggestion
  const creatorModelIdById = new Map<string, number>();
```

front/lib/resources/skill/skill_resource.ts
```diff
@@ -3083,6 +3083,60 @@ export class SkillResource extends BaseResource<SkillConfigurationModel> {
     return result;
   }
 
+  static async batchGetCreatorIds(
+    auth: Authenticator,
+    skills: SkillResource[]
+  ): Promise<Map<number, string | null>> {
```
aubin-tchoi: nit: we could return the entire UserResource | null to make the signature even more self explanatory

aubin-tchoi (approved):
LGTM

front-api/routes/w/[wId]/skills/index.ts
```diff
@@ -236,6 +236,8 @@ app.get(
               skill.canAdministrate(auth))
         );
 
+    const creatorIds = await SkillResource.batchGetCreatorIds(auth, skills);
```
Comment (aubin-tchoi) [inline front-api/routes/w/[wId]/skills/index.ts]: A bit wary of adding one more sql query in this path since it's already quite slow. Not very clean but what about only adding it in the public API?

aubin-tchoi (approved):

front-api/routes/w/[wId]/skills/index.ts
Comment (sfriquet) [reply front-api/routes/w/[wId]/skills/index.ts]: Not sure it's good to introduce discrepancies, if it's too slow we could also store the creator in the skill_configuration. A bit more work to do + requires a migration but that'd be the cleanest implementation imo.
We could do the same for agents when we introduce a proper agent model.

Comment (aubin-tchoi) [reply front-api/routes/w/[wId]/skills/index.ts]: agreed, we didn't do it initially because the product does not show it and that was intentional: the last editors matters as much as the creator given that even small edits can change the behavior notably. We could argue the same and push back on the original issue and ask why they need the creator vs the full list of users who made edits. they could definitely rely on the editor list for this

sfriquet:
Closing for now until we get more feedback

PR #32066, opened 2026-09-08
[front] fix: keep generated text visible during inline activity

aubin-tchoi:
## Description

Context [here](https://dust4ai.slack.com/archives/C050SM8NSPK/p1788869543248569), trace [here](https://langfuse.dust.tt/project/cmi4ft2qm00061707q4a9azyc/traces?filter=metadata%3BstringObject%3BconversationId%3B%3D%3Bmw37Htt2hW&peek=93db2d263f53eb99fee408a4b7853aec&timestamp=2026-09-08T06%3A02%3A40.959Z&observation=40a51bfa22ed85ea).

We have an example where the final generation tokens just before a tool call (actually part of the same LLM output to be precise) are not displayed in a conversation where the agent is waiting for the user to answer a question.

This PR fixes that by keeping available content visible in that case.

## Tests

- Added a test.
- Tested locally + preview on the problematic convo

## Risk

- Low.

## Deploy Plan

- Deploy front.

### DIFF
--- a/front/components/assistant/conversation/actions/inline/InlineActivitySteps.tsx
+++ b/front/components/assistant/conversation/actions/inline/InlineActivitySteps.tsx
@@ -59,11 +59,6 @@ function getTerminalLabel(status: LightAgentMessageType["status"]): string {
  *
  * Steps are accumulated by useAgentMessageStream — this component is a pure render.
  */
-/**
- * @cc [owner:aubin-tchoi,label:product] preserve-active-message-content
- * Before completion, non-empty message content must remain visible outside the
- * collapsible activity section, including while thinking or executing tools.
- */
 export function InlineActivitySteps({
   agentMessage,
   lastAgentStateClassification,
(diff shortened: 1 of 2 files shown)


front/components/assistant/conversation/actions/inline/InlineActivitySteps.tsx
```diff
@@ -203,12 +204,12 @@ export function InlineActivitySteps({
       : undefined;
 
   const extraBelowCollapse =
-    showActiveWriting && agentMessage.content ? (
+    !isDone && agentMessage.content ? (
```
aubin-tchoi, line 206: `showActiveWriting` is defined as `!isDone && isWriting`
here we have a case where the model generates both a tool call and some content in one step and we hide the content. usually not a big deal but here the tool call is `ask_user_question` and the content explain what the question is about 🙃

PopDaph: Tricky one!

PopDaph (approved):

PR #32068, opened 2026-09-08
[sparkle] re-import platform logos and retire White/Mono variants

Duncid:
## Description

Re-imports every platform logo from Figma (`Sparkle Assets` → page `Platforms` → section `Used`) and retires the per-theme variants. The 103 sources now map 1:1 to the design file — same names, same casing, in both directions.

- **Retired** `GithubMonoLogo`, `GithubWhiteLogo`, `LinearWhiteLogo`, `ZendeskWhiteLogo` and the legacy `OutlookLogo`. In-repo callers use the canonical mark; `@deprecated` aliases in `registry.ts` keep published consumers compiling for one release.
- **Colour fix.** svgr's `replaceAttrValues` (black → `currentColor`) was leaking into platform logos. They are brand marks on a fixed chip, so the substitution left 24 of them invisible against their own background in dark mode. The platforms svgr run now passes `--no-runtime-config`. Stroke icons and the Dust mono logos still get the substitution, which is correct for them.
- **Pipeline fix.** `build_icons.sh` used to `rm -rf src/logo/platforms`, destroying the hand-maintained `registry.ts` and with it `PLATFORM_LOGOS` and the aliases. It now clears generated files individually, and the index template re-exports the registry so the line survives a rebuild.
- **Added** `Atlassian`, `GoogleCloud` and `Togetherai`, plus the seven logos that had generated TSX but no SVG source (`Clari`, `Lemlist`, `Minimax`, `Modjo`, `Moonshot`, `ThinkingMachines`, `Zai`) and would have been destroyed by the next rebuild. `YoutrustLogo` was exported and allowlisted but missing from `PLATFORM_LOGOS`; it's now registered.
- **Docs.** New `# Platform logos` section in `sparkle/AGENTS.md` covering the Figma source of truth, the PascalCase naming chain, the colour rule, what `registry.ts` is for, and the deprecation dance required before removing a public export.
- **Storybook.** Logo catalogs grouped under `Assets/Logo` as `Dust Logo` and `Platform Logos`, matching how `Assets/Icons` and `Assets/Avatars` are already structured.

## Tests

- `npx tsgo --noEmit` clean in `sparkle`, `front` and `marketing`.
- `npm run build-storybook` completes; the generated story index confirms the new `Assets/Logo/*` hierarchy.
- Scripted parity audit: 103 Figma symbols ↔ 103 SVG sources ↔ 103 exports ↔ `PLATFORM_LOGOS` entries, with no case drift in either direction. All three wire-format allowlists (`front`, `sdks`, `marketing`) resolve through the registry.
- `build_icons.sh` re-run several times to confirm it is now idempotent and preserves `registry.ts`.

Visual verification is via Storybook (`Assets/Logo/Platform Logos`) rather than a screenshot, since the change covers the whole catalog.

## Risk

Low to moderate, and almost entirely visual. Every platform logo's artwork changes, so the connector and integration surfaces deserve a pass — the 24 logos previously hit by the `currentColor` substitution are the ones worth checking in dark mode specifically.

No public export is removed: all five retired names remain as aliases, so consumers of `@dust-tt/sparkle` keep compiling. Wire-format payloads (`skill.icon`, `remote_mcp_server.icon`, `webhook_sources_view.icon`) still resolve, including for the retired names, because the aliases are also present in `PLATFORM_LOGOS`.

No schema or data changes, so this is safe to roll back.

## Deploy Plan

Standard — no ordering constraints and no migrations.

Follow-up for the release *after* the next Sparkle publish: delete the five `@deprecated` aliases and their matching `PLATFORM_LOGOS` entries at the bottom of `sparkle/src/logo/platforms/registry.ts`.


ykmsd (approved):
I don't have a solution but the radius is not matching when it's inside a chip and feels a bit weird
Image

Duncid:
@ykmsd agreed—actually looks more like a problem of positioning the icon than radius itself no?
image

PR #32071, opened 2026-09-08
company space: give Company Data its own member group

fabiencelier:
## Description

Give the company space (`Company Data`, the `global` space) a member list that decides who can modify its content. Today write on it comes only from the `admin` / `manager` / legacy `builder` workspace roles, because `spaceGroupRoles` grants `reader` to every group attached to the global space.

This PR makes room for that list, and nothing else:

- The global space is now granted like an open regular space: the workspace global group stays a viewer (`reader`, so everyone keeps reading Company Data) and every other attached group holds `member` (read + write).
- Every global space gets its own `regular_auto` member group: `makeDefaultsForWorkspace` creates it with the space, `ensureGlobalSpaceMemberGroup` is the idempotent repair the backfill uses for existing workspaces, and `SpaceFactory.global` matches production.

Nobody gains access: the group is created empty, and nothing can add members to it yet: `updatePermissions` / `addMembers` still reject the global space (that comes in PR 3).

## Tests

 - new unit tests

## Risk

Medium. Touches how the global space's grants are computed, which is the read path for every workspace's company data — but the only behavioral change is that a group attached to the global space would confer write, and the sole such group is created empty.

## Deploy Plan

- Deploy front
- Run the post-deploy migration `front/migrations/20260908_backfill_global_space_member_group.ts` (dry-run first, then `--execute`; `--wId` for a single workspace)

### DIFF
--- a/front/lib/resources/space_resource.test.ts
+++ b/front/lib/resources/space_resource.test.ts
@@ -3019,4 +3019,121 @@ describe("SpaceResource group_permissions enforcement", () => {
     // read on a space the caller just opened, in the same request.
     expect(adminAuth.getGrantedVerbs("space", space.id)).toContain("read");
   });
+
+  // The global space (Company Data): everyone reads it, and write comes from the admin/manager
+  // roles plus the members of its member group (see `spaceRoleGrants` / `spaceGroupRoles`).
+  describe("global space member group", () => {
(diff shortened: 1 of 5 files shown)


tdraier (approved):
lgtm

PR #32076, opened 2026-09-08
[front] enh: reveal truncated tool names in message breakdown

aubin-tchoi:
## Description

Long tool names in the message consumption breakdown are truncated with no way to read their full name.

This PR makes it so hovering a truncated tool row expands the name into the space occupied by the usage chip and credit count. Both return when the pointer leaves. Names longer than the full row wrap.

https://github.com/user-attachments/assets/a297e561-f745-4e62-a2d4-aa964f92aa60

## Tests

- Tested locally + preview.

## Risk

- Low.

## Deploy Plan

- Deploy front.

### DIFF
--- a/front/components/assistant/conversation/CreditCostPopover.tsx
+++ b/front/components/assistant/conversation/CreditCostPopover.tsx
@@ -65,10 +65,9 @@ function CreditDetailRow({
           labelElement &&
           labelElement.scrollWidth > labelElement.clientWidth
         ) {
-          const labelBounds = labelElement.getBoundingClientRect();
           labelElement.style.setProperty(
             "--credit-label-collapsed-width",
-            `${labelBounds.width}px`
+            `${labelElement.clientWidth}px`
           );
           setIsLabelExpanded(true);
         }
--- a/front/components/assistant/conversation/CreditCostPopover.tsx
+++ b/front/components/assistant/conversation/CreditCostPopover.tsx
@@ -74,51 +74,54 @@ function CreditDetailRow({
       }}
       onPointerLeave={() => setIsLabelExpanded(false)}
     >
-      <dt className="col-start-1 row-start-1 flex min-w-0 items-center gap-2 font-medium text-foreground">
-        {icon && (
-          <Icon
-            visual={icon}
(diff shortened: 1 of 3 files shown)


front/components/assistant/conversation/CreditCostPopover.tsx
```diff
@@ -37,37 +38,96 @@ function toolDescription(tool: AgentMessageConsumptionToolDetails): string {
...
+        if (
+          expandLabelOnHover &&
+          event.pointerType === "mouse" &&
+          labelElement &&
+          labelElement.scrollWidth > labelElement.clientWidth
+        ) {
+          event.currentTarget.style.setProperty(
+            "--credit-label-collapsed-width",
+            `${labelElement.clientWidth}px`
+          );
```
ykmsd: I don't think we have to run this on every pointer enter event no? shouldn't we set `--credit-label-collapsed-width` only once on mount with useEffect?

front/components/assistant/conversation/CreditCostPopover.tsx
```diff
@@ -37,37 +38,96 @@ function toolDescription(tool: AgentMessageConsumptionToolDetails): string {
...
+      >
+        <span className="col-start-1 row-start-1 flex items-center gap-2">
+          {icon && (
+            <Icon visual={icon} size="xs" className="text-muted-foreground" />
+          )}
+          <span
+            ref={labelRef}
+            className={cn(
+              "overflow-hidden whitespace-nowrap",
+              isLabelExpanded ? "z-10 bg-overlay-background" : "text-ellipsis"
```
ykmsd: only minor problem of this approach is you cannot select the whole tool name on expand:

Image

I'm not sure how to fix it though, maybe we swap z index and show the underneath one on top? 🤔

ykmsd (approved):

front/components/assistant/conversation/CreditCostPopover.tsx
aubin-tchoi: ah good catch, we can swap z index at the end of the animation yeah

PR #32079, opened 2026-09-08
fix(front): apply template seats when the package is unchanged

tdraier:
## Description

Fixes the "Free pilot" template applying with **no seat enabled at all**.

Selecting a template resets the form (clearing seats + package) and then re-selects the template's package. When the resolved package equals the one already selected — e.g. re-applying a template, or **switching between the two pilots, which both use the Enterprise Pooled package** — `metronomePackageId` never actually changes, so React never sees `selectedSeats` change. The seat-reset effect and the deferred template-apply effect therefore don't fire, and the seats stay `{}` (empty). Applying from a blank form the first time worked; re-applying/switching didn't.

Fix:
- Extract the package's default-seat builder into `buildDefaultSeats` (shared with the seat-reset effect so they can't drift).
- In `applyTemplate`, when the package is unchanged, rebuild the seats and restore the default start **synchronously** (and still apply the template fields), instead of relying on effects that won't fire.
- Also surface an error if a template's package can't be resolved for the current currency, instead of silently applying no package/seats (the other way "no seats" could happen).

## Tests

- `tsgo` and biome pass.
- Manual: apply Free pilot from blank → workspace seat enabled; switch Free pilot ↔ Pilot — 2 months repeatedly → seats stay populated each time; re-apply the same template → seats populated.

## Risk

Low, internal Poke tooling. Behavior only changes the template-apply path (no server/billing changes). Rollback is a straight revert. No migration or feature flag.

## Deploy Plan

Standard `front` deploy. No migration or ordering constraints.

### DIFF
--- a/front/components/poke/shadcn/ui/form/fields.tsx
+++ b/front/components/poke/shadcn/ui/form/fields.tsx
@@ -57,8 +57,14 @@ export function SelectField<T extends FieldValues>({
       name={name}
       render={({ field }) => {
         const selectedOption = flatOptions.find((o) => o.value === field.value);
+        // Fall back to the title or a neutral placeholder — never the raw field
+        // `name` (a dotted path like `seats.free.paymentSchedule.frequency`),
+        // which leaks when the current value matches no option.
         const displayLabel =
-          selectedOption?.display ?? selectedOption?.value ?? title ?? name;
+          selectedOption?.display ??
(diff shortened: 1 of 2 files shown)


tdraier:

- #32079 ◀

fabiencelier (commented):
LGTM

PR #27558, opened 2026-06-18
Restore group membership when a user rejoins

tdraier:
## Description

When a user is removed from a workspace (via SCIM or manually), their group memberships are ended at the same time. On rejoin, only the workspace membership was restored — group memberships were not, causing users to lose editor access on agents they created.

Fix: when creating a workspace membership for a user who previously had one revoked,  now looks for group memberships ended within ±60 seconds of the workspace revocation and restores them. The 60s window matches the co-termination window in `handleDeleteWorkOSUser` (groups removed just before workspace membership). Applied to all rejoin paths, not just SCIM.

Fixes dust-tt/tasks#8984, dust-tt/tasks#8420.

## Tests

Manually verified the logic — no automated test added in this PR (flow is exercised through the SCIM Temporal activity path).

## Risk

Low. The restore path only runs when a previously revoked membership exists. It creates new group membership rows without touching existing ones, and skips groups that no longer exist or where the user is already active.

## Deploy Plan

Standard deploy, no migration needed.

PR #8420, opened 2024-11-04
Fix: multiple-upload allow re-uploading the same file, can be re-opened after cancelling the dialog

Fraggle:
## Description

Fixes:
- Couldn't open the dialog twice without reloading the page.
- Couldn't re-upload twice the same file without reloading the page (upload A, delete A, upload A)

## Risk

None

## Deploy Plan

Deploy `front`

PR #32081, opened 2026-09-08
[Admin Governance] Exclude former members from agent editor lists

philipperolet:
## Description

Related to https://github.com/dust-tt/dust/pull/32056; fixes a separate `agent_editors_batch` shadow mismatch.

Revoked workspace members can still appear in batched agent editor lists because revocation preserves group memberships. Require active workspace membership as well, matching the existing single-agent lookup and the new grant-backed lookup. Stored memberships and grants are unchanged.

## Risks

Blast radius: Batched agent editor listings in the UI and reporting scripts.
Risk: low

## Deploy Plan

- Deploy front; no migration.

### DIFF
--- a/front/lib/api/assistant/editors.ts
+++ b/front/lib/api/assistant/editors.ts
@@ -2,6 +2,7 @@ import { shadowCompare } from "@app/lib/api/permissions/shadow";
 import type { Authenticator } from "@app/lib/auth";
 import { AgentResource } from "@app/lib/resources/agent_resource";
 import { GroupResource } from "@app/lib/resources/group_resource";
+import { MembershipResource } from "@app/lib/resources/membership_resource";
 import { UserResource } from "@app/lib/resources/user_resource";
 import type { LightAgentConfigurationType } from "@app/types/assistant/agent";
 import { removeNulls } from "@app/types/shared/utils/general";
@@ -137,6 +138,10 @@ async function shadowAgentEditorsBatch(
   });
 }
 
+/**
+ * @cc [owner:philipperolet,label:product] active-agent-editors
+ * Returned editors must have active membership in the workspace and their agent's editor group.
+ */
 export const getAgentsEditors = async (
   auth: Authenticator,
   agentConfigurations: LightAgentConfigurationType[]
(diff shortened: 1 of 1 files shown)


philipperolet:
@PopDaph by any chance, do you know or have an opinion on whether we should actually revoke all group memberships when we revoke a user?

philipperolet:
Follow-up on whether workspace revocation should also revoke every group membership:

TLDR: there's a rationale, to resurrect users easily; while questionable technically, not in scope of this PR to revisit the decision

AI details below:

**Current behavior and rationale**
- The ordinary workspace-revocation path leaves group memberships intact; [SCIM deprovisioning already ends them](https://github.com/dust-tt/dust/blob/19f344ac9fb8959b176c17d6177e9e4e6c7990b2/front/temporal/workos_events_queue/activities.ts#L1317).
- [PR #27558](https://github.com/dust-tt/dust/pull/27558) deliberately added restoration on rejoin after SCIM removal/recreation caused users to lose agent-editor and pod access, requiring manual repairs ([#8984](https://github.com/dust-tt/tasks/issues/8984), [#8420](https://github.com/dust-tt/tasks/issues/8420)).
- The comment in `MembershipResource.revokeMembership` explains why group-cache invalidation is unnecessary once workspace access is revoked; it is not itself a product rationale for retaining membership rows.
- Restoration matches group memberships ended within ±60 seconds of the workspace revocation. That window concerns the original termination timestamps, **not the duration of the user's absence**: old assignments can therefore be restored much later.

**Recommendation — separate product decision, not implemented here**
Distinguish reversible deactivation from permanent removal:
- **Deactivate/suspend:** deny workspace access but preserve assignments for reactivation, accommodating temporary removal and provisioning churn.
- **Remove/offboard:** end explicit group memberships while retaining history. Rejoining should use current IdP assignments and newly approved permissions; restoring old permissions should be explicit.

The key policy is what happens **on rejoin**, not just whether revocation sets group-membership `endAt`. I would not introduce blanket group revocation as a follow-up to this fix without deciding those semantics.

This PR remains appropriate either way: preserved assignments or incomplete cleanup must not make former workspace members appear as current agent editors.

PopDaph:
👋🏻 I haven’t checked the data model, but my intuition is that memberships should be fully historized. For example, if I’m a regular user from May 1 to May 15 and then become a Builder, I’d expect those to be represented as two separate memberships, each with its own validity period.

More generally, if a user’s workspace membership is revoked, I’d expect all of their active group memberships to be ended as well, while preserving the historical records. We could then make a separate product decision about whether those memberships should be restored if the user rejoins, or whether they should be granted again explicitly.

PR #32082, opened 2026-09-08
feat(analytics): author the analyst panel's opening message server-side

achilleburah:
## Description

Currently the Analytics page's "Ask @analyst" panel opens on a hardcoded greeting bubble with its own input bar, and only creates a conversation once the user submits

PR stack: Directly start a conversation with the from the Ask Analyst side conversation panel instead of first presenting a empty state "fake message". 

- PR#1 implements the backend necessary to do that.
- PR#2 connects the front

## Tests

Tested Locally.

## Risk

Low. Still behind Dust Only FF. 

## Deploy Plan

Standard deploy.

### DIFF
--- a/front-api/routes/swagger_private_schemas.ts
+++ b/front-api/routes/swagger_private_schemas.ts
@@ -1489,7 +1489,7 @@
  *           type: string
  *         origin:
  *           type: string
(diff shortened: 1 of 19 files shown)


achilleburah:
changed PR split

PR #32083, opened 2026-09-08
[Admin Governance] Support admin API keys in agent ACLs

philipperolet:
## Description

Follows https://github.com/dust-tt/dust/pull/32024.

Admin API keys can edit agents without belonging to a user editor group. The served `canEdit` reflects that behavior, but the permission shadow still checks only the agent ACL, producing false write/admin mismatches.

Grant regular admin API keys `write` in `AgentResource`'s ACL and make the shadow candidate use `auth.can()` for all three verbs. Keep active-status and requested-space checks outside the ACL, using the already-loaded spaces. The legacy admin-key condition remains only until the read flip; the migration plan records this boundary and the distinction between grant enumeration and effective ACL permissions.

Require admin rather than builder role, since builder is being retired: builder keys now report `canEdit: false`. Human and system-key rules, global-agent restrictions, and stored grants remain unchanged.

The PATCH endpoint's existing acceptance of legacy builder keys is unchanged; tightening that authorization is separate from permission reporting.

## Risks

Blast radius: Agent ACL checks, permission reporting, and shadow comparisons for regular API keys.
Risk: low

## Deploy Plan

- Deploy front; no migration or backfill.

## Tests

- [x] Existing API-key endpoint suites with shadow-parity assertions: 16 passed, including admin-only edit reporting on single-agent and list responses.
- [x] Agent configuration, AgentResource, and shadow suites: 47 passed, including admin/builder keys, workspace scoping, read-only global agents, and human/system-key restrictions.

### DIFF
--- a/front-api/routes/v1/w/[wId]/assistant/agent_configurations.test.ts
+++ b/front-api/routes/v1/w/[wId]/assistant/agent_configurations.test.ts
@@ -80,7 +80,7 @@ describe("GET /api/v1/w/[wId]/assistant/agent_configurations", () => {
     expect(response.status).toBe(200);
     expect(
       agentConfigurations.find((a) => a.name === "Published Agent")?.canEdit
-    ).toBe(role !== "user");
+    ).toBe(role === "admin");
     expect(
       agentConfigurations
         .filter((a) => a.scope === "global")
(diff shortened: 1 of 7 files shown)


aubin-tchoi (commented):
LGTM

PR #31973, opened 2026-09-07
Use cells for agent email relay

Fraggle:
## Description

Inbound email relay now uses `cellsConfig` instead of region configuration and iterates through configured cells when forwarding SendGrid webhooks. Relay authorization, source metadata, error selection, logging, retries, and multipart attachments now use cell-aware handling.

## Tests

No automated tests were added or updated in this PR.

## Risk

Medium risk because this changes production email routing from a single-region target to sequential cell delivery. Existing retry and error handling remain in place. No schema changes; rollback is safe through a front/front-api deployment rollback.

## Deploy Plan

Deploy `front` and `front-api` together. No SQL migrations or infrastructure changes are required.

PR #32084, opened 2026-09-08
[Email Agents] Relay workspace misses in cell order

philipperolet:
## Description

Fixes email routing introduced in https://github.com/dust-tt/dust/pull/31973: a successful webhook response acknowledges receipt before workspace lookup, so email never reached the second remote cell.

On a workspace lookup miss, forward to the next cell in alphabetical name order: US → EU1 → EU2. Each cell derives later destinations from the shared configuration; no remaining-cells header is needed. Assert that US comes first. The receiving cell owns further lookup and forwarding, and the last cell sends one error reply using the most informative lookup error.

Keep fallback to later cells after an explicit HTTP rejection. Stop on acceptance or exhausted transport retries, because the target may already be processing.

## Risks

Blast radius: inbound Email Agents routing between cells.
Risk: standard

## Deploy Plan

- Deploy front-api from the last cell to the first: EU2, EU1, then US. No migrations.

## Tests

- [x] Email webhook suite: 12 tests, including US → EU1 and EU1 → EU2 routing with an unsorted cell configuration, HTTP rejection fallback, stopping after uncertain receipt, duplicate detection, and one final error reply without relaying back.

### DIFF
--- a/front-api/routes/email/webhook.test.ts
+++ b/front-api/routes/email/webhook.test.ts
@@ -1,13 +1,10 @@
 import { randomUUID } from "node:crypto";
-import { makeEmailAgentsDisabledEmailTriggerError } from "@app/lib/api/assistant/email/email_trigger";
 import {
   EMAIL_WEBHOOK_RELAY_HEADER,
   EMAIL_WEBHOOK_RELAY_HEADER_VALUE,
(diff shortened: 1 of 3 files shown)


Fraggle (approved):
🙏🏼

PR #32085, opened 2026-09-08
fix(oauth): accept custom Freshworks organization domains

frankaloia:
## Summary
- https://github.com/dust-tt/tasks/issues/7864
- Freshservice connection finalize rejected custom Freshworks org domains (e.g. `it.example.com`) because the hostname regex only allowed `*.myfreshworks.com`. OAuth authorize already succeeded; the failure was silent at token exchange.
- Accept any valid hostname, strip scheme/trailing slash, and send token requests through the untrusted egress proxy now that the host is user-provided.

## Risk
Low

## Testing
- new tests added

## Deploy
1. Deploy oauth
2. Deploy front


smb2268 (approved):
LGTM

PR #32086, opened 2026-09-08
fix(slack): block auto-respond without mention on private channels

frankaloia:
## Summary
- https://github.com/dust-tt/tasks/issues/10157
- Auto-respond without mention does not work in private Slack channels because the bot app is not subscribed to `message.groups`. The agent builder was still letting admins enable it, so the setting appeared to work and then silently did nothing.
- The Slack prefs sheet now disables the toggle when the selection is only private channels, and never writes the flag onto private channels when the selection is mixed.

## Testing
Screenshot 2026-09-08 at 1 12 37 PM

## Risk
Low

## Deploy Plan
1. Deploy front


smb2268 (approved):
LGTM but I think connectors deploy also needed?

PR #32088, opened 2026-09-08
Track activity steps toggle clicks

PopDaph:
## Description

Track clicks on the inline activity steps toggle ("Thinking…" while generating, "Completed in …" once done) so we can see whether people open or close the steps, and whether they do it while the message is still generating.

Event: `conversation:activity_steps:open` / `conversation:activity_steps:close`

| property | values |
|---|---|
| `source` | `message`, `run_agent_details` |
| `message_state` | `generating`, `done` |

`source` is a required prop on `ActivityTimeline` so any new consumer has to declare where it lives.

## Risk

Low, tracking only.

## Deploy Plan

- Deploy front


theogz (approved):
👌

PR #32090, opened 2026-09-08
chore: add scoped empty trigger conversation cleanup script

smb2268:
## Description

Adds a reusable front admin script for cleaning up empty conversations created by a scheduled trigger after a terminal failure.

The script requires all cleanup dimensions explicitly:

- Workspace sId
- User sId
- Trigger sId
- Inclusive start timestamp
- Exclusive end timestamp

It is dry-run by default. Candidates must belong to the exact workspace and trigger, be within the requested time window, have zero messages, and have no participant other than the requested user. Empty failed-trigger conversations commonly have no participant because creation happens before participant insertion, so the script validates that the requested user is the trigger editor rather than requiring a participant row. Execution revalidates the message and participant conditions before removing participants and soft-deleting the conversations.

## Tests

- `bun build front/scripts/cleanup_empty_trigger_conversations.ts --target=node --no-bundle --external:'*'`
- `git diff --check`
- No production cleanup was executed.
- Biome was not run because the repository does not have dependencies installed in the Computer and the registry is blocked by the sandbox network policy.

## Risk

Low and bounded. The script is dry-run by default, requires explicit identifiers and timestamps, refuses selections above the configurable safety limit, skips conversations that gain messages or other participants, and uses soft deletion rather than hard deletion. The main operational risk is selecting an overly broad time window, mitigated by the exact trigger and user-editor validation plus the dry run.

## Deploy Plan

1. Run the script without `--execute` and review the candidate conversation IDs and count.
2. Confirm the workspace, user, trigger, and timestamps match the intended incident.
3. Rerun with `--execute` to soft-delete the confirmed empty conversations.
4. Re-run the dry-run query to verify no matching non-deleted conversations remain.

Follow-up application fixes should prevent these empty conversations from being created in the first place.

### DIFF
--- a/front/scripts/cleanup_empty_trigger_conversations.ts
+++ b/front/scripts/cleanup_empty_trigger_conversations.ts
@@ -1,0 +1,264 @@
+import { Authenticator } from "@app/lib/auth";
+import {
+  ConversationParticipantModel,
+  MessageModel,
+} from "@app/lib/models/agent/conversation";
+import { ConversationResource } from "@app/lib/resources/conversation_resource";
+import { frontSequelize } from "@app/lib/resources/storage";
+import { UserModel } from "@app/lib/resources/storage/models/user";
+import { TriggerResource } from "@app/lib/resources/trigger_resource";
+import { WorkspaceResource } from "@app/lib/resources/workspace_resource";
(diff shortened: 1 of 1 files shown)


smb2268:
Not going to clean up for now bc they have no participants

PR #32092, opened 2026-09-09
Use localStorage instead of sessionStorage so sessions survive new tabs

Zeph-Pln05:
## Description

 For non-consented visitors, PostHog stored its `session_id` in per-tab sessionStorage, so opening any link in a new tab (incl. cmd+click) started a new session and fragmented visits. This switches pre-consent persistence to cookieless localStorage, which is shared across same-origin tabs, so the session survives new tabs. Consented users are unaffected. Follows on from #31385.
  

## Tests

Manually verified in-browser: reproduced the per-tab fragmentation (reject cookies → open link in new tab → new session_id), and confirmed localStorage users keep the same session_id across tabs and a full browser restart.

## Risk

Low

### DIFF
--- a/marketing/components/app/PostHogTracker.tsx
+++ b/marketing/components/app/PostHogTracker.tsx
@@ -108,7 +108,8 @@ function PostHogTrackerInner({ authenticated }: PostHogTrackerInnerProps) {
 
   // Phase 1: Initialize PostHog. This captures events for all visitors,
   // including anonymous ad traffic. Visitors who have not accepted cookies get
-  // sessionStorage persistence, which is cleared when the tab closes.
+  // cookieless localStorage persistence, which is shared across same-origin
+  // tabs so opening a link in a new tab keeps the same session.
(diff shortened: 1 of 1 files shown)


Zeph-Pln05:
@zmarouf For non-consented visitors we keep PostHog's `session_id` in sessionStorage (per-tab), so opening any link in a new tab (incl. opening Blog page or using cmd+click) starts a fresh session and fragments the visit (tracking wise). This PR switches the pre-consent store to localStorage so the session is shared across tabs. Tradeoff: sessionStorage clears on tab close, localStorage persists, but it stays cookieless & raw IP not stored. 

Was sessionStorage a deliberate privacy choice here, or just a default? And is **persistent-but-cookieless, no-PII localStorage pre-consent fine privacy-wise**?

zmarouf (approved):
PostHog default: 30 minutes idle / 24 hours maximum.
LGTM

PR #32077, opened 2026-09-08
Remove connected data from pods.

Fraggle:
## Description

I retire admin-controlled Pods and Pod-level connected data. This removes `isAdminControlled` from `ProjectMetadataModel`, Pod metadata APIs, SDK types, feature flags, and membership transitions; removes the Connected Data tab and Pod search inclusion; and restores standard editor/member management. `NavTabPill` now keeps labels visible on active tabs and shows them as tooltips on collapsed tabs.

UI changes: please attach a screenshot or GIF showing Pod navigation and settings after the Connected Data and admin-controlled options are removed.

## Tests

Updated `front-api` and resource tests to remove admin-controlled Pod expectations and align metadata payloads. Updated `NavTabPill` Storybook documentation.

## Risk

Medium risk. The retired feature and its connected-data path become unavailable. The post-deploy migration drops `project_metadata.isAdminControlled`, so deploy the application changes before running the migration; rollback requires restoring the column and code. Standard Pod membership and editor flows remain unchanged.

## Deploy Plan

Deploy `front` and `front-api`. After the application deploy, run `front/migrations/post-deploy/20260908151717_Remove_isAdminControlled_field.sql`. Verify Pod navigation, member/editor management, and Pod settings after deployment.

PR #32093, opened 2026-09-09
Add Pod customization settings

Fraggle:
## Description

Adds the `Pod Customization` settings section for `pod_frame_tabs`. Editors can rename pinned file tabs, choose custom icons, remove tabs, and reorder file tabs while preserving system-tab slots. The UI uses `usePodFileTabs` and optimistic `useUpdatePodMetadata` updates so settings and Pod navigation stay in sync. 

image

## Tests

Added Vitest coverage for `reorderFileTabsInTabsOrder` and `orderedPodFileTabs`, including system-slot preservation, no-op/missing paths, and navigation ordering.

## Risk

Low. The UI is feature-gated and reuses the existing Pod metadata PATCH endpoint. The main risk is an optimistic cache mismatch after a failed metadata update; SWR rolls back the live space-info cache and the mutation returns an error state. Rollback is safe because the change is front-end only.

## Deploy Plan

Deploy `front`. No SQL migrations or infrastructure changes.

### DIFF
--- a/front/lib/swr/pods.ts
+++ b/front/lib/swr/pods.ts
@@ -50,6 +50,7 @@ import type {
 } from "@app/types/api/sandbox/egress_policy";
 import type {
   CheckNameResponseBody,
+  GetSpaceResponseBody,
   PatchPodMetadataBodyType,
 } from "@app/types/api/spaces";
 import type {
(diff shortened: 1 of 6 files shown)


Fraggle:

- #32077
    - #32093 ◀

front/lib/swr/pods.ts
```diff
@@ -1078,59 +1079,163 @@ export function useUpdatePodMetadata({
...
     updates: PatchPodMetadataBodyType
   ): Promise<PodMetadataType | null> => {
     const url = `/api/w/${owner.sId}/spaces/${podId}/project_metadata`;
 
-    const res = await clientFetch(url, {
-      method: "PATCH",
-      headers: { "Content-Type": "application/json" },
-      body: JSON.stringify(updates),
-    });
+    const mirrorsOnSpace =
```
fabiencelier: not obvious what this does from the name, maybe add a small comment or find more relevant name ?

fabiencelier (approved):

front/lib/swr/pods.ts
Fraggle: Changed, hopefully better :)

PR #32095, opened 2026-09-09
Remove sanitization process and use base64

Fraggle:
## Description

Moves local 1Password environment loading away from the Python sanitizer and stores `GCP_SERVICE_ACCOUNT_B64` as base64-encoded JSON. `materialize_dev_environment` now sources the raw dotenv output, while `write_gcp_service_account_file` decodes the key into `SERVICE_ACCOUNT`, validates base64, and enforces mode `0600`. Removes `dev/scripts/sanitize-op-env.py`.

## Tests

Not run; this change only affects local development environment bootstrapping.

## Risk

Medium. Invalid base64 or a non-shell-compatible 1Password dotenv value can break local shell initialization or prevent Google clients from finding credentials. The decoder fails closed, removes the invalid credential file, and logs an actionable error. Rollback is safe by reverting the change and restoring the previous `GCP_SERVICE_ACCOUNT` entry.

## Deploy Plan

No application deploy, database migration, or infrastructure change. Update the 1Password entry to provide `GCP_SERVICE_ACCOUNT_B64` with base64-encoded service-account JSON, then restart or reload local development shells so `SERVICE_ACCOUNT` is regenerated.


Fraggle:

- #32095 ◀

zmarouf (commented):
LGTM

PR #32103, opened 2026-09-09
[front] fix: remove `dd.internal.entity_tag` from StatsD metrics

aubin-tchoi:
## Description

We have seen examples where some dd metrics have a high cardinality due to the `dd.internal.entity_tag` tag (example [here](https://app.datadoghq.eu/metric/volume?metric=llm_cache.token_hit_ratio&sort=-volume_total)). This tag contains the pod UID as value.

This PR renames it into `dd.internal.entity_id`, it's the name used by the Datadog agent to add the tags for `podLabelsAsTags`. This tag is stripped from the tags that end up in the metric so does not contribute to the billed cardinality.

Example of metric emitted with the renamed tag [here](https://app.datadoghq.eu/metric/summary?filter=statsd.dummy.count&metric=statsd.dummy.count&sort=name).

## Tests

- Added a script that increments a dummy metric so that we can check in datadog what tags on it.

## Risk

- Low.

## Deploy Plan

- Deploy front.

### DIFF
--- a/front/lib/utils/statsd.ts
+++ b/front/lib/utils/statsd.ts
@@ -3,13 +3,15 @@ import { StatsD } from "hot-shots";
 
 let statsDClient: StatsD | undefined = undefined;
 
+/**
+ * @cc [owner:aubin-tchoi,label:performance] statsd-pod-origin-tag
+ * When `DD_ENTITY_ID` is set, the client must identify the pod through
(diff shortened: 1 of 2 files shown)


flvndvd:
Seen IRL, we actually need this tag to ensure custom metrics used by our monitors are properly tagged with the pod labels.

flvndvd (approved):

PR #32104, opened 2026-09-09
chore: simplify table config fetching error handling

aubin-tchoi:
## Description

This PR simplifies the error handling in table configuration fetching for MCP tools.
Instead of having `fetchTableDataSourceConfigurations` return an Error that needs to be wrapped into an MCP error, it directly supplies an MCPError that is then just passed through.

## Risk

- Low.

## Deploy Plan

- Deploy front.


aubin-tchoi:
PMRR

PR #32105, opened 2026-09-09
[front] fix: defer nested transaction effects until outer commit

aubin-tchoi:
## Description

Search indexation must not run when only an inner savepoint has committed. Add shared transaction helpers that defer effects until the outer commit, suppress them on rollback, and support atomic operations returning `Result`.

Cache invalidation uses the same outer-commit boundary. Nested savepoints share a unique physical counter while retaining their logical parent for effect ordering.

Foundation for the revised shared skill/agent search, stacked after #32046.

## Tests

- Added real-transaction coverage for nested commits, rollback, and deferred cache invalidation.

## Risk

- Medium. Changes nested transaction and cache-invalidation timing.

## Deploy Plan

- Deploy front.

### DIFF
--- a/front/lib/utils/cache.ts
+++ b/front/lib/utils/cache.ts
@@ -1,5 +1,6 @@
 import { getRedisCacheClient } from "@app/lib/api/redis";
 import { distributedLock, distributedUnlock } from "@app/lib/lock";
+import { runAfterTransactionCommit } from "@app/lib/utils/sql_utils";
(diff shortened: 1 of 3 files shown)


aubin-tchoi:
No longer needed after removing the agent refactor. #32233 retains a small skill-specific after-commit helper without the transaction/savepoint framework; #32232 contains the query changes. The original branch is retained for recovery.

PR #32106, opened 2026-09-09
[front] refactor: own global agent settings in a resource

aubin-tchoi:
## Description

Global-agent availability needs a resource-owned database boundary before it joins shared search. Introduce workspace-scoped settings reads, atomic upserts, and transactional cleanup, with admin-only writes.

Global-agent constructors now consume plain settings instead of Sequelize models. The existing settings endpoint and three-second Deep Dive cache retain their behavior.

## Tests

- Added workspace isolation, idempotent upsert, rollback, and admin-permission coverage.

## Risk

- Low. Preserves the existing global-agent settings API.

## Deploy Plan

- Deploy front.


aubin-tchoi:
Closing because global-agent settings are outside the reduced skills-only scope. The replacement stack is #32232 followed by #32233. The original branch is retained for recovery.

PR #32115, opened 2026-09-09
feat(analytics): hide an opening message the user did not write

achilleburah:
## Description

Stack, merges in order:

**This PR (1).**
Registers an `analytics_panel` message origin and hides the rank 0 message carrying it, the way `agent_sidekick` already is. 

The rest is the fan-out a new origin always costs: one entry each in the exhaustive tables for message validation, agent loop queue, unread notifications, usage classification, analytics labels, plus the SDK enum and swagger. 

## Tests

`conversation-unread` and `agent_loop/config` suites pass, 95 tests. Nothing produces the origin yet, so behaviour is unchanged.

## Risk

Inert on its own, since nothing carries the origin until PR#3.

The origin is client-claimable, like `agent_sidekick`. The hiding is gated on rank 0, so the most a caller can do is hide the first message of a conversation it just created, which is the exposure Sidekick already carries. It is deliberately kept out of `HIDDEN_MESSAGE_ORIGINS`, which hides at any rank and also drives pod conversation previews.

One behaviour to note for when PR#3 lands: the origin classifies as `user` usage, like `agent_sidekick`, so opening the panel counts toward DAU.

## Deploy Plan

Standard, no migration.

### DIFF
--- a/front/components/assistant/conversation/types.ts
+++ b/front/components/assistant/conversation/types.ts
@@ -171,11 +171,15 @@ export const isTriggeredOrigin = (origin?: UserMessageOrigin | null) => {
 
 // Central helper to control which user message should be hidden in the UI.
 // Extend this list as we introduce more bootstrap/system user messages.
-export const isHiddenMessage = (message: VirtuosoMessage): boolean => {
+export const isHiddenMessage = (
+  message: VirtuosoMessage,
+  conversation: ConversationWithoutContentType | undefined
+): boolean => {
   return (
     (isUserMessage(message) &&
       (isHiddenMessageOrigin(message.context.origin) ||
-        isSidekickBootstrapMessage(message))) ||
+        isSidekickBootstrapMessage(message) ||
+        isBootstrappedOpeningMessage(message, conversation))) ||
     isHandoverUserMessage(message)
   );
 };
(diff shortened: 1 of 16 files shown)


front/components/assistant/conversation/types.ts
```diff
@@ -173,11 +173,15 @@ export const isTriggeredOrigin = (origin?: UserMessageOrigin | null) => {
...
+export const isHiddenMessage = (
+  message: VirtuosoMessage,
+  conversation: ConversationWithoutContentType | undefined
+): boolean => {
   return (
     (isUserMessage(message) &&
       (isHiddenMessageOrigin(message.context.origin) ||
-        isSidekickBootstrapMessage(message))) ||
+        isSidekickBootstrapMessage(message) ||
+        isBootstrappedOpeningMessage(message, conversation))) ||
```
Fraggle: Did you consider the same approach as sidekick bootstraped message instead of creating a new metadata thing on the conversation ?

If we can avoid introducing more complexity / ways to do things, the better :)

I might be missing some context / reason that led to this choice 'tho.

Fraggle (approved):
LGTM (but check my comment)

front/components/assistant/conversation/types.ts
achilleburah: Yes I did but Sidekick has a very specific comment on it stating that this is not the good approach and should be migrated toward this new approach. 
So it's probably best to try and refactor Sidekick instead.

Fraggle: Consistently bad is better than having two differents system in place.
Imho, either you align on what was done for sidekick or you come up with a better solution and you switch both.

Taking a step back, isHiddenMessage should be at the message level, the fact that you put something on the whole conversation but then look for the message rank == 0 doesn't seem super principled.

Questions that comes to mind:
- What is a "bootstraped" conversation ?
- What if you want to hide more messages in this conversation but at different (non deterministic) ranks ?

achilleburah: I hear your points. I was planning to move the Sidekick origin afterwards but I can align first and refacto both afterward. 

Context, for me a "bootstraped" conversation was one initiated by the system. The name could be more transparent. But I don't think the system can send prompt messages during a conversation ?

achilleburah: Extended CLIENT_MESSAGE_ORIGINS to match sidekick's implementation

Fraggle (approved):
THanks !

PR #32128, opened 2026-09-09
[front-api] docs: clarify public trigger serialization contracts

spolu:
## Description

Split the public trigger serialization contract into explicit requirements for excluded internal fields and webhook source metadata, including missing views and the `"custom"` fallback for nullish providers.

## Tests

- Verified both contracts against the serializer and its callers, and checked contract syntax and discoverability.
- Attempted both existing trigger endpoint suites; neither could run because Vite could not resolve `@novu/framework`.

## Risk

None. Comment-only change.

## Deploy Plan

No deployment required.

### DIFF
--- a/front-api/routes/v1/w/[wId]/triggers/serialize.ts
+++ b/front-api/routes/v1/w/[wId]/triggers/serialize.ts
@@ -6,10 +6,15 @@ import type { TriggerType as PublicTriggerType } from "@dust-tt/client";
 
 /**
  * @cc [owner:adrien,label:api] public-api-trigger-serialization
- * Serializes a workspace-wide `TriggerResource` list for the read-only public API: internal
- * fields with no meaning outside the workspace (`editor`, `origin`, `spaceId`,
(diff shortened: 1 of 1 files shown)


spolu:
r? @adrsimon PMRR better contract writing

I'm updating the skill to nudge in that direction.

Note that your function breaks 3 existing contracts;

```
Existing directory-contract findings:

  - Numeric id is exposed (front-api/routes/v1/w/[wId]/triggers/serialize.ts:38), violating string-ids-in-api-interfaces. Removing it would conflict with public API compatibility rules.
  - Serialization is assembled outside the Resource (front-api/routes/v1/w/[wId]/triggers/serialize.ts:37), violating resource-owned-serialization.
  - Kind branching is non-exhaustive (front-api/routes/v1/w/[wId]/triggers/serialize.ts:49), contrary to the directory’s preferred pattern.
```

adrsimon (commented):
LGTM

adrsimon:
will check 👍 tyty

PR #32129, opened 2026-09-09
[front] Clarify per-user pool limit copy when seats carry no allowance

tdraier:
## Description

The spending-policy copy on the Usage page hard-coded *"This limit is added on top of each seat's built-in allowance."* That is inaccurate for pooled workspaces whose seats carry no built-in allowance (e.g. pooled enterprise): there the per-user pool limit **is** the user's whole monthly budget, not a top-up.

This branches the copy on **real seat data** rather than the plan code — enterprise contracts exist both *with* and *without* a per-seat allowance, so the plan prefix is not a valid signal:

- **Workspace-level surfaces** (`UsageSettingsCard` default-limit setting, `BulkEditSpendLimitModal`) derive `seatsHaveBuiltInAllowance` from `seatPlans` (`SeatTypeInfo.awuCredits > 0` for any seat type), computed in `UsagePage` and threaded down.
- **Per-member modals** (`EditSpendLimitModal`, `EditMemberSpendLimitModal`) use the member's own seat allowance (`memberUsageLimit`), which is more precise per member.

When no allowance is present, the copy drops the "on top of the seat allowance" wording and states the pool limit is the whole budget. The Poke plugin's static field description (covers all plans) is reworded to state both cases.

Covers both usage-page variants (`enable_new_usage_page`) — the affected components render in both.

## Tests

Manual: no logic change beyond display copy. `npx tsgo --noEmit` and `biome check` pass on all changed files. No automated test covers this static copy.

## Risk

Low. Display-copy-only change gated on an already-fetched, contract-derived seat-allowance signal; no data mutation or API surface change.

## Deploy Plan

Standard deploy. No migration, feature flag, or config change required.

### DIFF
--- a/front/components/pages/workspace/UsagePage.tsx
+++ b/front/components/pages/workspace/UsagePage.tsx
@@ -830,6 +830,9 @@ export function UsagePage() {
   const plan = subscription.plan;
   const isEnterprise = isEnterprisePlanPrefix(plan.code);
   const isFreePlanWorkspace = isFreePlan(plan.code);
+  const seatsHaveBuiltInAllowance = Object.values(seatPlans).some(
+    (info) => (info?.awuCredits ?? 0) > 0
+  );
 
   const poolConsumedCredits = Math.max(
     0,
@@ -1388,6 +1391,7 @@ export function UsagePage() {
                     <UsageSettingsCard
                       workspaceId={owner.sId}
                       hasPool={hasPool}
+                      seatsHaveBuiltInAllowance={seatsHaveBuiltInAllowance}
                     />
                   )}
                   <ModelTiersSettingsCard owner={owner} />
@@ -1446,6 +1450,7 @@ export function UsagePage() {
           isOpen={isBulkSpendLimitOpen}
           onClose={() => setIsBulkSpendLimitOpen(false)}
           memberCount={selection.selectedCount}
+          seatsHaveBuiltInAllowance={seatsHaveBuiltInAllowance}
           onValidate={handleBulkSpendLimitValidate}
         />
         <BulkChangeSeatModal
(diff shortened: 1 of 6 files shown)


tdraier:

- #32129 ◀

front/components/pages/workspace/UsagePage.tsx
```diff
@@ -826,6 +826,15 @@ export function UsagePage() {
   const plan = subscription.plan;
   const isEnterprise = isEnterprisePlanPrefix(plan.code);
   const isFreePlanWorkspace = isFreePlan(plan.code);
+  // Whether any seat sold on this workspace's contract carries a built-in AWU
+  // allowance. Derived from the seats themselves (`SeatTypeInfo.awuCredits`),
```
avervaet: nit: avoid naming specific fields/components in comments.

front/components/pages/workspace/UsagePage.tsx
```diff
@@ -826,6 +826,15 @@ export function UsagePage() {
...
+  // not from the plan code: enterprise contracts exist both with and without a
+  // per-seat allowance. When no seat carries an allowance (e.g. pooled
+  // enterprise), the pool limit is the user's whole budget rather than a top-up,
+  // and the spending-policy copy drops the "on top of the seat allowance" wording.
+  const seatsHaveBuiltInAllowance = Object.values(seatPlans).some(
```
avervaet, line 833: nit: we might want to avoid computing these while isSeatPlanLoading

avervaet (approved):
LGTM

front/components/pages/workspace/UsagePage.tsx
tdraier: while loading seatPlans is empty, seatsHaveBuiltInAllowance will be false - message will update once loaded, acceptable

PR #32131, opened 2026-09-09
Store skill attribution on consumption items

flvndvd:
## Description

Tool skill attribution was only materialized while building Elasticsearch documents, so the database attribution items did not retain the skill IDs behind each tool call. This stores `attributedSkillIds` (for our data friends!) on pending and completed tool items using the same rule as Elasticsearch, including skills that expose the tool and skills enabled by the action result.

The new column is nullable so existing rows remain valid. Elasticsearch reads the stored value for new rows and recomputes it for legacy null rows, which avoids a backfill or reindex requirement.

## Tests

The focused attribution integration suite passed all 48 tests. It covers exposed tool skills, enable-skill actions, pending and completed row transitions, database persistence, model validation, and the legacy Elasticsearch fallback.

The local repository-wide type check is blocked by unrelated existing Sparkle UI errors and local debug scripts. A filtered check reports no errors in the changed attribution paths.

## Risk

Risk is low. The migration only adds a nullable column, old application code ignores it, and existing rows keep their current Elasticsearch behavior. Rolling back the application code is safe while leaving the unused column in place.

## Deploy Plan

Apply the pre-deploy migration before deploying the application code. No backfill or reindex is required.


sfriquet (approved):

front/migrations/pre-deploy/20260909141830_add_attributed_skill_ids_to_agent_message_consumption_items.sql
```diff
@@ -0,0 +1,6 @@
+SET SESSION statement_timeout = 3000;
+SET SESSION lock_timeout = 3000;
+
+/* Existing attribution rows remain null and use the legacy projection fallback. */
+ALTER TABLE "public"."agent_message_consumption_items"
+ADD COLUMN "attributedSkillIds" character varying(255)[];
```
aubin-tchoi, line 6: we store the string ids, not the model IDs? isn't the model ID more practical for our data friends?

aubin-tchoi (approved):
LGTM

front/migrations/pre-deploy/20260909141830_add_attributed_skill_ids_to_agent_message_consumption_items.sql
flvndvd: Given that all global skills only have a `sId` and no PK, will stick with `sId`.

PR #32132, opened 2026-09-09
feat(analytics): add the analyst conversation to history on first write

achilleburah:
## Description

Stack, merges in order:

**This PR (2).**
Posting a message to one of these conversations moves it from `test` to `unlisted`, so it reaches the user's history the first time they actually write in it. It recognises them by the origin of their opening message, through the existing `ConversationResource.openingUserMessageOrigin`, which is also how `project-new-conversation` tells a conversation Dust opened from one the user started. The conversation itself carries no new metadata.

## Tests

Two endpoint tests on the messages route: a conversation the panel opened gets promoted, and a `test` conversation it did not open is left alone.

## Risk

Low. Fired without awaiting so it cannot fail a message that was already posted, and the lookup is an indexed read behind the `visibility === "test"` guard, so it stops running once the conversation is promoted.

## Deploy Plan

Standard, no migration.

### DIFF
--- a/front-api/routes/w/[wId]/assistant/conversations/[cId]/messages/index.ts
+++ b/front-api/routes/w/[wId]/assistant/conversations/[cId]/messages/index.ts
@@ -13,7 +13,6 @@ import { SkillResource } from "@app/lib/resources/skill/skill_resource";
 import { extractUniqueSkillIds } from "@app/lib/skills/format";
 import { concurrentExecutor } from "@app/lib/utils/async_utils";
 import { statsDMetrics } from "@app/lib/utils/statsd";
-import logger from "@app/logger/logger";
 import { InternalPostMessagesRequestBodySchema } from "@app/types/api/assistant";
 import type { PostMessagesResponseBody } from "@app/types/api/assistant/messages";
 import type {
@@ -408,12 +407,7 @@ app.post(
     // so it must not fail a message that was already posted.
     void promoteAnalyticsPanelConversation(auth, {
       conversation: conversationResource,
-    }).catch((err) =>
-      logger.error(
-        { err, conversationId: conversation.sId },
-        "Failed to promote Analytics-panel conversation"
-      )
-    );
+    });
 
     const contentFragments = await fetchPrecedingContentFragments(auth, {
       conversationResource,
--- a/front-api/routes/w/[wId]/assistant/conversations/[cId]/messages/index.ts
+++ b/front-api/routes/w/[wId]/assistant/conversations/[cId]/messages/index.ts
@@ -2,6 +2,7 @@ import { validateMCPServerAccess } from "@app/lib/api/actions/mcp/client_side_re
 import { isSidekickConversation } from "@app/lib/api/actions/servers/helpers";
 import { fetchPrecedingContentFragments } from "@app/lib/api/assistant/content_fragments";
 import { postUserMessage } from "@app/lib/api/assistant/conversation";
+import { promoteAnalyticsPanelConversation } from "@app/lib/api/assistant/conversation/analytics_panel";
 import { addSelectedConversationSpaces } from "@app/lib/api/assistant/conversation/selected_spaces";
 import { fetchConversationMessages } from "@app/lib/api/assistant/messages";
 import { getAuditLogContext } from "@app/lib/api/audit/workos_audit";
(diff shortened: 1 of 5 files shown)


front/types/assistant/conversation.ts
```diff
@@ -525,12 +525,18 @@ export type ConversationUrlAccessMode =
...
 export type ConversationMetadata = Record<string, unknown> & {
   urlAccessMode?: ConversationUrlAccessMode;
   projectTaskId?: string;
   useFileSystem?: boolean;
   /** Selects the database-backed filesystem for a fresh standalone conversation. */
   useDatabaseFileSystem?: boolean;
+  /** The surface that created the conversation, when it was not a plain user conversation. */
+  origin?: ConversationOrigin;
   /** Set when the conversation's first message was written by that surface, not by the user. */
   bootstrapped?: boolean;
```
Fraggle: I'm a bit confused, now instead of one we have 2 new metadata on the convos ?

achilleburah: Well, I could have added only one and rest all the logic on it but I feel like this could lead to tricky convoluted logic as those two fields do not represent the same thing. 

I could just "bootstrap" the conversation on a check of the origin === analytics_panel but then that would make the whole thing less flexible. 

The `origin` indicates "where" the conversation was started and can handle some specific logic based on. (ie, will also allow filtering previous conversations in a follow up PR")
While a conversation can still be "bootstrapped" and never need any logic based on the origin. 

I tend to like clear business separation, but I could merge those two if you think it's best.

achilleburah: Followed on the user origin from PR1

front/lib/api/assistant/conversation/analytics_panel.ts
```diff
@@ -0,0 +1,18 @@
...
+  if (conversation.visibility !== "test") {
+    return;
+  }
+
+  const openingOrigin = await conversation.openingUserMessageOrigin(auth);
+  if (openingOrigin !== "analytics_panel") {
+    return;
+  }
+
+  await conversation.updateVisibilityToUnlisted(auth);
```
Fraggle: nit: it silently do nothing if not applicable. It's not a big deal but what about returning a Promise<Result<undefined, Error>> ? Even if you do not really care about it today, it's good practice and not too much more code.

front-api/routes/w/[wId]/assistant/conversations/[cId]/messages/index.ts
```diff
@@ -402,6 +404,17 @@ app.post(
       return apiError(ctx, messageRes.error);
     }
 
+    // Fire and forget: this only decides whether the conversation shows up in the user's history,
+    // so it must not fail a message that was already posted.
+    void promoteAnalyticsPanelConversation(auth, {
```
Comment (Fraggle) [inline front-api/routes/w/[wId]/assistant/conversations/[cId]/messages/index.ts:408]: super nit: Do you know what's the behavior of ignoring an async in the backend if the handling of the request is shorter than the handling of the async ?

Fraggle (approved):
LGTM

front/lib/api/assistant/conversation/analytics_panel.ts
achilleburah: matched `Promise<Result<undefined, Error>> ` to return an err, but now discarding it at the api handler level.
Do you prefer i handle the result rather than fire-and-forget ?

front-api/routes/w/[wId]/assistant/conversations/[cId]/messages/index.ts
Comment (achilleburah) [reply front-api/routes/w/[wId]/assistant/conversations/[cId]/messages/index.ts:408]: should work fine on a long live deployed env, isn't it ?

Comment (Fraggle) [reply front-api/routes/w/[wId]/assistant/conversations/[cId]/messages/index.ts:408]: Likely, but I was wondering.

front/lib/api/assistant/conversation/analytics_panel.ts
Fraggle: Nope, fire and forget is fine.

front-api/routes/w/[wId]/assistant/conversations/[cId]/messages/index.ts
Comment (achilleburah) [reply front-api/routes/w/[wId]/assistant/conversations/[cId]/messages/index.ts:408]: Issue would be on a lambda or short live env.

PR #32134, opened 2026-09-09
feat(analytics): open the analyst panel on a live conversation

achilleburah:
## Description

Stack, merges in order:

**This PR (3).**
The panel creates its conversation when it first opens, sending the opening message the way the agent builder sidekick does and tagging it with the `analytics_panel` origin so PR#1 hides it. The empty state, its input bar and its sticky mentions are deleted.

`ResizableSidePanel` keeps the panel mounted while closed, so creation is keyed on the panel being open rather than on mount. The `disabled` prop becomes `isOpen` to kill the double negative.

## Tests

No automated coverage, this is UI wiring. Checked manually:

- loading the Analytics page without opening the panel issues no `POST /assistant/conversations`
- opening it issues exactly one, then shows @analyst's greeting with no gap above it
- the conversation stays out of the sidebar until the first real question
- closing and reopening reuses the same conversation

## Risk

Connects the rest of the stack but still behind the `analytics_conversation_panel` flag.

The regression it is most exposed to is the first item above: the panel stays mounted while closed, so a mount-keyed effect would create a conversation on every Analytics page load.

## Deploy Plan

Standard, merges last in the stack.

### DIFF
--- a/front/components/assistant/conversation/ConversationViewer.tsx
+++ b/front/components/assistant/conversation/ConversationViewer.tsx
@@ -125,6 +125,8 @@ interface ConversationViewerProps {
   owner: WorkspaceType;
   user: UserType;
   clientSideMCPServerIds?: string[];
+  /** Set while the viewer is mounted but not shown, to skip its fetches. */
+  disabled?: boolean;
 }
 
 function easeOutQuint(x: number): number {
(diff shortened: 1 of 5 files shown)


Fraggle (approved):
LGTM

PR #32138, opened 2026-09-09
Clarify Result and caught-error contracts

spolu:
## Description

Clarify the `no-catching-own-errors` and `normalize-caught-errors` contracts: expected failures use `Result`, unexpected throws propagate, and caught values require normalization or a runtime type guard before being treated as errors. Remove the unconditional HTTP 500 and normalization-success claims, and retain the example using an external-library call.

## Tests

N/A. Contract prose and example changes only.

## Risk

None.

## Deploy Plan

Merge to main.

### DIFF
--- a/CONTRACTS
+++ b/CONTRACTS
@@ -486,13 +486,11 @@ attempts to write the attacker-controlled file as the workload user.
 Repository functions MUST return failures that callers are expected to handle as `Err<E>` values in
 `Result<T, E>`, rather than throw them.
 
-External-service adapters whose local contract requires throwing recognized provider failures are
-exempt: they MAY throw documented errors, and callers MAY catch them using runtime type guards.
-`catch` is also permitted around external-library calls whose exceptions we do not control.
-
 Applicable local contracts MAY define explicit exceptions to this rule, limited to the boundaries
-and failure cases they specify. Outside these exceptions, errors thrown by repository code MUST
-propagate to the runtime's error handler.
+and failure cases they specify; callers MAY catch those documented failures using runtime type
+guards. `catch` is also permitted around external-library calls whose exceptions we do not control.
+Outside these exceptions, errors thrown by repository code MUST propagate to the runtime's error
+handler.
 
 @cc [owner:spolu,label:error-handling] normalize-caught-errors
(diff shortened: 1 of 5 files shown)


spolu:
(Mostly to test `cc` automation)

CONTRACTS
```diff
@@ -483,31 +483,28 @@ hardening that makes the path root-owned and not group/other writable, plus a re
...
 
 @cc [owner:spolu,label:error-handling] no-catching-own-errors
-Never catch your own errors. `catch` is authorized around external libraries (whose error handling
-we don't control), but otherwise errors that may alter the execution upstream should be returned
-using our `Result<>` pattern. It is OK to throw errors, since we can't catch them these are
-guaranteed to trigger an internal error (and return a 500).
+Repository functions MUST return failures that callers are expected to handle as `Err<E>` values in
+`Result<T, E>`, rather than throw them. Repository code MUST NOT catch errors thrown by repository
+code; `catch` is permitted around external-library calls whose exceptions we do not control.
+Unexpected errors MAY be thrown and MUST propagate to the runtime's error handler.
```
flvndvd: What about Temporal activities, only allowed place, right?

flvndvd (approved):

CONTRACTS
spolu: Yes fair we could use an exception there. The exception being that it's the only way to communicate errors between activities and workflows?

flvndvd: yep, that's correct! So only activities are allowed but all helpers need to use `Result` until they reach the root of the activity (often limited to files in `/activities/`).

CONTRACTS
```diff
@@ -483,31 +483,28 @@ hardening that makes the path root-owned and not group/other writable, plus a re
 attempts to write the attacker-controlled file as the workload user.
 
 @cc [owner:spolu,label:error-handling] no-catching-own-errors
```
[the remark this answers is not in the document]
spolu: handling this one with a contract for temporal activities

spolu:
Will make a pass of `normalize-caught-errors` fixes since these are easier to handle in a subsequent PR. `no-catching-own-errors` hopefully will get handled progressively as agents stumble on them.

PR #32139, opened 2026-09-09
Sort members usage by the same consumed source the table renders

avervaet:
## Description

The pool and seat usage in-app sorts read per-user usage from the Metronome cache while the rendered values come from the analytics index, so ties and cold-cache pages could order rows against the displayed numbers. Both sorts now use the analytics fetcher the response uses.

## Deploy Plan

- Deploy front


avervaet:

- #32139 ◀

avervaet:
pmrr @tdraier

tdraier:
✅

tdraier (approved):

PR #32142, opened 2026-09-09
Admin access: show the restricted access message instead of the editor for redacted agents and skills

fabiencelier:
## Description

Prevent seeing the editor page when trying to access a private agent/skill as an admin
Instead show a call to action to become an editor and join spaces.
This make it consistent with what admin can see all over the app.
Before this PR they would just see a half-filled editor

image

image

## Tests

tested locally

https://github.com/user-attachments/assets/2537b393-f240-441f-b0fd-b0cabd4f50e4

## Risk

low

## Deploy Plan

deploy front


aubin-tchoi (commented):
The page looks very barebone, is there a way to instead have all the components in a disabled state or have the page be blurred or smth?

fabiencelier:
@aubin added the actual editor with blur in the background 

image

aubin-tchoi (approved):
LGTM if performances are good (blur is usually quite heavy on CPU)

fabiencelier:
I think perf is OK + it's pretty niche as no link point to this page
it's just admin guessing the edit page for an agent or getting a link from a user
Non admin people will keep seeing a 404

PR #32143, opened 2026-09-09
Show transferred-asset counts after merging user identities

tdraier:
## Description

Fixes https://github.com/dust-tt/tasks/issues/8920.

After running the **Merge user identities** poke plugin (e.g. to fix duplicate accounts from an SSO/email change), support had no quick way to confirm the merge actually worked — the only signal was a periodically-refreshed Snowflake `DIM_USERS` snapshot, which is stale right after the merge.

The plugin now reports, right after the merge, **how many of each asset were transferred from the secondary user onto the primary**, as a markdown table:

Screenshot 2026-09-09 at 15 54 19

Counts reflect rows genuinely re-pointed to the primary user; assets the primary already owned are deduped/deleted, so they are correctly **not** counted as transferred.

Implementation: `mergeUserIdentities` now returns a `transferCounts` object built from the affected-row count of each migration. To make those counts available, three helpers (each with only `mergeUserIdentities` as a non-test caller) now surface their transfer count:
- `ConversationResource.mergeUserParticipations` → `Promise<number>`
- `GroupResource.migrateUserMemberships` → `Promise<number>`
- `TriggerResource.transferEditor` → `Result<number, Error>` (was `Result<undefined, Error>`)

## Tests

- Extended the existing `migrateUserMemberships` and `transferEditor` resource tests to assert the returned counts, including the dedup case (primary already owned the asset → count 0). `group_resource.test.ts` + `trigger_resource.test.ts`: 52/52 pass.
- `tsgo --noEmit` clean; Biome format clean.

## Risk

Low. The change is additive — it only adds a return value (counts) to functions and enriches the poke plugin output; the merge behavior itself is unchanged. The three helpers changed signature but all non-test callers were updated in this PR, and callers only ever checked `isOk()`/ignored the return. Safe to roll back.

## Deploy Plan

Standard `front` deploy. No migration, no config, no ordering constraints.

### DIFF
--- a/front/lib/api/poke/plugins/workspaces/user_identity_merge.ts
+++ b/front/lib/api/poke/plugins/workspaces/user_identity_merge.ts
@@ -3,8 +3,25 @@ import {
   emitAuditLogEvent,
 } from "@app/lib/api/audit/workos_audit";
 import { createPlugin } from "@app/lib/api/poke/types";
+import type { UserIdentityMergeTransferCounts } from "@app/lib/iam/users";
 import { mergeUserIdentities } from "@app/lib/iam/users";
 import { Err, Ok } from "@app/types/shared/result";
+
+// Labels for the transferred-asset summary shown after a merge, in display order.
+const TRANSFER_COUNT_LABELS: [keyof UserIdentityMergeTransferCounts, string][] =
+  [
+    ["conversations", "Conversations"],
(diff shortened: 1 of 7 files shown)


tdraier:

- #32143 ◀

adrsimon (approved):
LGTM

PR #32144, opened 2026-09-09
Rename top up to "Add credits" on the new usage page and refresh the credit cards after a purchase

avervaet:
## Description

- Rename the "Top up" button to "Add credits" on the new usage page and render it above the credit cards. It is now greyed out rather than hidden when top-up is disabled for the workspace.
- Revalidate the pool current-cycle and cycle-history endpoints from the purchase dialog success handler so the new credit cards refresh after a purchase or coupon redemption.
- Invalidate the Redis-cached current-cycle and cycle-history reads when a payment-gated commit is paid or a pool top-up coupon is redeemed. Without this the client refetch got the pre-grant balance back for up to 60s.
- Remove the "Credit consumption" and "Previous cycles" titles from the shared credit pool section (seen with @ClementAupiais). This also applies to the Poke credits page.

## Tests

- `lib/api/metronome/process_webhook.test.ts` and `lib/metronome/coupons.test.ts` pass.
- Manual: purchase flow on the new usage page not exercised end to end in a browser.

## Risk

Low

## Deploy Plan

Deploy front

### DIFF
--- a/front/components/pages/workspace/UsagePage.tsx
+++ b/front/components/pages/workspace/UsagePage.tsx
@@ -290,9 +290,7 @@ export function UsagePage() {
     sort?.id === "email" || sort?.id === "seatUsage"
       ? sort.id
       : sort?.id === "consumedFromPoolAwuCredits"
-        ? isNewUsagePage
-          ? "consumedFromPoolAwuCredits"
-          : "consumedAwuCredits"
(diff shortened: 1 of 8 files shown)


avervaet:

- #32144 ◀

tdraier (approved):
lgtm

PR #32145, opened 2026-09-09
Deploys and reverts name cells, not regions

flvndvd:
## Description

**Today.** Deploy and Revert ask for a region, `us-central1`, `europe-west1` or `all`, and send that to dust-infra, which runs the legacy clusters. cell-00002 is not reachable from here.

**After.** They ask for cells. `cell-00000` and `cell-00001` are the legacy clusters, `cell-00002` the first customer cell, `all` every one of them. The payload to dust-infra carries `cells`, and dust-infra (dust-tt/dust-infra#1081) turns each into a deploy target, records the tag in git and applies. The Slack thread names cells too.

The marketing build is the one job that still thinks in regions, its bundle inlines region specific values, so its matrix maps the two legacy cells to their region and skips any other. Nothing else in the build path changes.

## Tests

Read through and linted with actionlint. The cells to regions mapping for marketing was run locally for `all`, a single legacy cell, a mix with `cell-00002` and `cell-00002` alone, which yields an empty matrix as intended since marketing does not run on cells. A real deploy after both PRs are merged is the proof.

## Risk

dust-infra#1081 must merge first, it understands `cells` and refuses a payload without them. Merged in that order, the only window is between the two merges, during which a deploy from here stops in dust-infra before touching any cluster. Rollback is a revert of both.

## Deploy Plan

1. Merge dust-tt/dust-infra#1081.
2. Merge this.
3. Deploy any component to `cell-00001` from the Actions tab: one line in the thread for that cell, a commit in dust-infra under `versions/`.

### DIFF
--- a/.github/actions/slack-notify/action.yml
+++ b/.github/actions/slack-notify/action.yml
@@ -98,7 +98,7 @@ runs:
           channel: ${{ inputs.channel }}
           text: |
             ${{ inputs.is_revert == 'true' && '⏮️' || '🚀' }} ${{ inputs.is_revert == 'true' && 'Starting revert of' || 'Starting deployment of' }} ${{ inputs.component }} `${{ inputs.image_tag }}`
-            • Region: `${{ inputs.region }}`
+            • Cells: `${{ inputs.region }}`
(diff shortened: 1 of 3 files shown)


flvndvd:
PMRR.

PR #32147, opened 2026-09-09
[LLM] Open Grok models to all workspaces and deprecate Grok 4.5

pmilliotte:
## Description

We now have ZDR on xai.
Removes the `xai_feature` feature flag so Grok models are available to all workspaces (EU-inference-only workspaces stay excluded via `regionalAvailability`, since inference runs in the US), and deprecates Grok 4.5 (`isLegacy: true`, dropped from `USED_MODEL_CONFIGS`) now that 4.6 is the flagship.

## Tests

Local

## Risk

Low on the code path — but Grok becomes visible to all workspaces and Grok 4.6 appears on the public credits table, so the exposure is a product/comms decision rather than a technical one.

## Deploy Plan

- Deploy `main` on `front`

r? @Nils-Fedrigo PMRR 🙏


Nils-Fedrigo (commented):
lgtm, I don't understand why 
we replace "[Dust only] xai_feature",
 with "[Dust only] noop_model_feature",


but you probably know better!

PR #32149, opened 2026-09-09
Extract shared CreditLimitInput and spend limit parsing helpers

avervaet:
## Description

Stack 1/4 for #32027. Unifies the spend limit inputs and fixes a few rough edges found along the way.

- Shared `CreditLimitInput` for the personal limit field and the group limit table.
- Single bounded credit parser behind `parseCreditsInput` and `parseDefaultLimitInput`, reused by the usage settings card. `toSpendLimit` returns the API `UserSpendLimit` type.
- Default limit field shows the real value, so 0 renders as "0" and saving 0 applies 0. Placeholder is "--" while the value loads. Clearing the field is rejected instead of being coerced to 0.
- Personal and group limit inputs use a "--" placeholder. "No limit" was misleading since an empty group cap does not mean uncapped.
- Group limit inputs keep focus while typing. The table columns are now defined at module scope, since tanstack renders each `cell` function as a component type and a new identity per render remounted the input.

## Tests

Manual: member spend limit modal (typing in group cells keeps focus, placeholders) and usage settings default limit (0 displays and saves as 0).

## Risk

Low. Small UI behavior changes on admin-only spend limit fields.

## Deploy Plan

Deploy front.


front/components/workspace/usage/UsageSettingsCard.tsx
```diff
@@ -57,20 +51,15 @@ export function UsageSettingsCard({
     });
   };
 
-  const currentDefaultLimit = defaultUserSpendLimit?.awuCredits ?? 0;
+  const currentDefaultLimit = defaultUserSpendLimit?.awuCredits ?? null;
```
PopDaph, line 59: Now that clearing the field is rejected, can we surface it instead of silently reverting? Today an empty or out-of-range value just snaps back to the previous number, so it's hard to tell whether the save happened. 🙏🏻

front/components/workspace/MemberGroupLimitTable.tsx
```diff
@@ -10,77 +11,80 @@ interface MemberGroupLimitTableProps {
...
-      meta: { headerAlign: "right" },
-      cell: ({ row }) => (
-        <DataTable.BasicCellContent
-          label={row.original.memberCount.toLocaleString()}
-          className="justify-end"
-        />
-      ),
-    },
-  ];
+  const data: GroupLimitRow[] = rows.map((row) => ({
```
PopDaph, line 77: nit: should we memoize this?

PopDaph (approved):

PR #32152, opened 2026-09-09
chore: move requester at beginning of line in review action

aubin-tchoi:
## Description

This PR edits the message output by the review bot in the review channel.
Puts the requester name at the beginning instead of the end.

Rationale is that the message is easier to parse at a glance if we know where to expect finding the requester name.

## Tests

- N/A

## Risk

- N/A

## Deploy Plan

- No deploy.


spolu:
LGTM

PR #32154, opened 2026-09-09
feat(extension): ephemeral per-conversation tool approval

tdraier:
## Description

Adds an **"Allow all inputs this session"** option to the tool approval prompt. Approving it grants the tool an ephemeral, conversation-scoped approval: it runs for the rest of the conversation **whatever the inputs**, without re-prompting — then resets when the conversation changes or the page reloads.

Fixes dust-tt/tasks#9381 — Mexens needs massive web form-filling across many URLs, but medium-stake tools re-prompt on every new input value (browser tools scope approval by `domainToFetch`/`tabId`). Rather than a permanent, workspace-wide per-agent stake grant, this scopes the trust to the conversation the user is actively driving: broad across inputs, but bounded in time.

Details:
- **Ephemeral**: in-memory only in `BlockedActionsProvider` (keyed by `mcpServerName`+`toolName`), never persisted, reset when the conversation changes.
- **Scoped to where it helps**: the button only shows for a `medium`-stake tool with a non-empty `argumentsRequiringApproval`. At other stakes it is redundant (`never_ask`/`low`) or contradicts the "never savable" intent (`high`), and a medium tool without approval args is already covered permanently by "Always allow".
- **No server change**: the approval prompt is server-driven; the client auto-answers the streamed `tool_approve_execution` with a one-off `approved` (never `always_approved`, which would persist).
- Available on all surfaces (web app + extension).

Three code contracts added: `conversation-approvals-ephemeral`, `conversation-approval-auto-submit`, `conversation-approval-medium-only`.

## Tests

Manual testing (in progress): approve a medium tool "for this session", confirm subsequent calls with different input values are not re-prompted, and that switching conversation / reloading resets it. `tsgo` and biome pass on the changed files.

## Risk

Low. Purely additive and client-side; nothing is persisted and the core stake model is unchanged. Known limitation: if the auto-submit validate request fails, the action stays blocked silently until the component remounts/reload. Safe to roll back (revert the commit).

## Deploy Plan

Standard front + extension deploy. No migration, no server/API change.


tdraier:

- #32154 ◀

adrsimon (commented):
do we want this only in the extension ? or should we do it in the main product too ? 

it feels weird to always approve for the conversation client side, it means we get an event asking us for an approval and we instantly approve without the user clicking anything right ?

tdraier:
> do we want this only in the extension ? or should we do it in the main product too ?
> 
> it feels weird to always approve for the conversation client side, it means we get an event asking us for an approval and we instantly approve without the user clicking anything right ?

i'm not sure why the card was explicit about only doing it in extension, and still wondering if it should not be everywhere.  otherwise, could be only on some tools ? ( browser interaction ? )

and yes we get the events and it's instantly approved , only ui/client - didnt want to add a new stake level here

adrsimon:
@tdraier i believe we could add it as an always approve option for the low stake ? either you always approve or always approve for convo ?

adrsimon:
this could benefit all tools actually

tdraier:
> @tdraier i believe we could add it as an always approve option for the low stake ? either you always approve or always approve for convo ?

i'm not sure about the different stakes but i think this one makes sense only for medium stake, since the goal is to bypass the `argumentsRequiringApproval` (ignored for low stake) ?

adrsimon (approved):
LGTM

PR #32155, opened 2026-09-09
remove databricks oauth provider and mcp feature flag

FlagBenett:
## Description

Removes the now-unused dedicated databricks OAuth provider (core Rust + front TS + SDK) and the databricks_tool feature flag. 
The provider was scaffolded for the internal Databricks MCP server (ID 45), which was already removed in favor of the official Databricks-managed MCP servers (Databricks SQL / Genie) shipped as remote presets. https://github.com/dust-tt/dust/pull/31445

Those presets authenticate through the generic mcp_static provider (host-derived OAuth from the workspace URL), not the dedicated databricks provider, so this code was orphaned.

## Tests

locally. 

## Risk

impact is 2 workspaces that had the feature-flag on while testing last year.

## Deploy Plan

core and front


aubin-tchoi (approved):
LGTM

PR #32156, opened 2026-09-09
agents: rename the global-agent `skills` field to `codeDefinedSkillIds`

fabiencelier:
## Description

`AgentConfigurationType.skills` holds the ids of the skills a **global** agent ships with, set
only by the in-code global agent definitions. Workspace agents attach skills through
`AgentSkillModel` rows and never populate it. So the name reads like the agent -> skills mapping
while being neither the mapping nor skill objects.

- Rename it to `codeDefinedSkillIds`, matching the `codeDefined` naming already used for
  code-defined skills (`code_defined/`, `codeDefinedFiles`, `CodeDefinedSkillFile`). No client
  reads the field, so this is internal only.
- Fix the two MCP tools that were relying on the misleading name: `get_agent_details` and the
  `workspace_analytics` agent-details tool both did `(agent.skills ?? []).join(", ")`, which
  reported `Skills: none` for **every** workspace agent since only global agents ever set it.
  They now resolve skills through `SkillResource.listByAgentConfiguration`, which handles both
  global and workspace agents.

Groundwork for #32157, where `skills` becomes the public API field name for the resolved skill
objects.

## Tests

Updated the existing tests that assert on the field: the global agent tests
(`global_agents.test.ts`, `analyst.test.ts`) and the private agent endpoint redaction tests
(`routes/w/[wId]/assistant/agent_configurations/[aId]/index.test.ts`, which read the field off an
untyped JSON body so the compiler could not catch them). Ran `skill_resource.test.ts` (84 tests)
since the rename touches `getSkillReferencesForAgent`.

The two MCP tool changes are covered by hand: both handlers return early for an agent the caller
cannot read, so the new call is only reached for readable agents.

## Risk

Low. Internal rename, compiler-enforced, no client reads the field and no serialized public
contract changes. The MCP tool change makes two tools report skills they previously always
reported as `none`.

## Deploy Plan

- Deploy front and front-api


aubin-tchoi (approved):
LGTM modulo renaming discussed in slack thread

PR #32157, opened 2026-09-09
public API: serialize an agent's skills on the agent endpoints

fabiencelier:
## Description

related to https://github.com/dust-tt/tasks/issues/9379 

The public `agent_configurations` endpoints serialize `actions`, `tags` and `requestedSpaceIds` but never the skills attached to an agent, so there is no way to build an agent -> skills mapping from JSON.

- Every public endpoint returning an agent configuration now serializes `skills` as`{ sId, name }` objects.
- `SkillResource.listByAgentConfigurationsBySId` resolves them in at most 2 queries regardless of agent count, and handles global agents, whose skills come from their in-code `globalSkillIds` rather than from `AgentSkillModel`.
- Agents whose details were redacted for the caller (`canRead` false) get an empty list, consistently with `redactPrivateAgentConfigurationFields`.

## Tests

 - added and updated unit test on public endpoint an resource

## Risk

Low. Additive on the response.

## Deploy Plan

- Deploy front

### DIFF
--- a/front/lib/api/assistant/configuration/helpers.ts
+++ b/front/lib/api/assistant/configuration/helpers.ts
@@ -8,17 +8,30 @@ import { AgentConfigurationModel } from "@app/lib/models/agent/agent";
 import { AgentResource } from "@app/lib/resources/agent_resource";
 import { GroupResource } from "@app/lib/resources/group_resource";
 import { canReadRequestedSpaces } from "@app/lib/resources/permission_utils";
+import { SkillResource } from "@app/lib/resources/skill/skill_resource";
+import type { SkillHydrationOptions } from "@app/lib/resources/skill/types";
 import { SpaceResource } from "@app/lib/resources/space_resource";
 import { TagResource } from "@app/lib/resources/tags_resource";
 import { TemplateResource } from "@app/lib/resources/template_resource";
 import { tagsSorter } from "@app/lib/utils";
 import type {
   AgentConfigurationType,
+  AgentConfigurationWithSkillsType,
   AgentFetchVariant,
   AgentModelConfigurationType,
   LightAgentConfigurationType,
 } from "@app/types/assistant/agent";
+import { isGlobalAgentId } from "@app/types/assistant/assistant";
 import type { ModelId } from "@app/types/shared/model_id";
+import { removeNulls } from "@app/types/shared/utils/general";
+import partition from "lodash/partition";
+import uniq from "lodash/uniq";
+
+const LABELS_ONLY_FETCH_OPTIONS: SkillHydrationOptions = {
+  withInstructions: false,
+  withTools: false,
+  withFileAttachments: false,
+};
 
 export function getModelForAgentConfiguration(
   agent: AgentConfigurationModel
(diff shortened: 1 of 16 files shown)


front/lib/resources/skill/skill_resource.ts
```diff
@@ -1507,6 +1511,31 @@ export class SkillResource extends BaseResource<SkillConfigurationModel> {
     );
   }
 
+  /**
+   * Resolves the code-defined skills a global agent declares, in one SQL query. Global agents
+   * hold no `AgentSkillModel` row, so they cannot go through `listByAgentConfigurations`.
+   * Skills the caller cannot access are filtered out, so the result can be shorter than the
+   * requested ids.
+   */
+  static async listByCodeDefinedSkillIds(
```
aubin-tchoi: in this function the naming is hard to understand without reading the comment

front/lib/api/assistant/configuration/helpers.ts
```diff
@@ -311,3 +324,68 @@ export function redactPrivateAgentConfigurationFields(
...
+      )
+    );
+  }
+
+  // `codeDefinedSkillIds` does not reach the wire: the resolved `skills` replace it.
+  return agents.map(
+    ({ codeDefinedSkillIds: _codeDefinedSkillIds, ...agent }) => ({
+      ...agent,
+      skills: (skillsByAgentSId[agent.sId] ?? []).map((skill) =>
+        skill.toAgentSkillJSON()
```
aubin-tchoi: nit: ideally we only use the toJSON methods in API handlers to avoid using the output types in the backend. here we could even inline it?

front/lib/resources/skill/skill_resource.ts
aubin-tchoi: OOC can't we do fetchByIds instead?

front/lib/resources/skill/skill_resource.ts
```diff
@@ -1435,17 +1435,16 @@ export class SkillResource extends BaseResource<SkillConfigurationModel> {
 
   /**
    * Batched version of listByAgentConfiguration. Performs 2 SQL queries.
-   * Does not support global agents as we rely on the ID for mapping.
+   * Does not support global agents as we rely on the ID for mapping: they all share the same
+   * model id and hold no `AgentSkillModel` row. Use `listByCodeDefinedSkillIds` for those.
    */
-  static async listByAgentConfigurations(
+  static async listByAgentConfigurations<
+    T extends Pick<LightAgentConfigurationType, "id" | "sId">,
```
aubin-tchoi: this type is very permissive, is there a more specific type we could use?

front/lib/api/assistant/configuration/helpers.ts
```diff
@@ -311,3 +324,68 @@ export function redactPrivateAgentConfigurationFields(
...
+
+/**
+ * @cc [owner:fabiencelier,label:security] no-skills-for-redacted-agents
+ * An agent whose details were redacted (`canRead === false`) MUST get an empty `skills` array:
+ * its skills are private, consistently with `redactPrivateAgentConfigurationFields`.
+ */
+export async function serializeAgentConfigurationsWithSkills<
+  // `codeDefinedSkillIds` is declared on the full configuration schema, but `getGlobalAgents`
+  // puts it on global agents in every variant, so light configurations carry it too.
+  T extends LightAgentConfigurationType & { codeDefinedSkillIds?: string[] },
```
aubin-tchoi: OOC are there cases where this type is different from `LightAgentConfigurationType & { codeDefinedSkillIds?: string[] }`? if not we can simplify a bit

aubin-tchoi (approved):
LGTM

front/lib/resources/skill/skill_resource.ts
fabiencelier: removed indeed

front/lib/api/assistant/configuration/helpers.ts
fabiencelier: This whole function serializeAgentConfigurationsWithSkills is the "toJson" equivalent
It is only used by the different endpoints 
There is no agent resource to put it in though
Renamed `toAgentConfigurationsWithSkills` to make it slightly better

PR #32158, opened 2026-09-09
Add group-to-role mapping (governance) replacing hardcoded dust-admins

tdraier:
## Description

Lets admins map any workspace group to the **Admin** or **Manager** role from **Settings & Governance**, replacing the hardcoded `dust-admins`/`dust-managers` name-based role provisioning. Members of a mapped group inherit the role; a user always holds the highest role granted across their groups (admin > manager). Groups need not be provisioned — manually-managed groups work too.

in governance (admin only) :

Screenshot 2026-09-09 at 17 01 58

in people/groups : 

Screenshot 2026-09-09 at 17 03 24

Fixes dust-tt/tasks#8601.

Key points:
- **Data:** nullable `grantedRole` column on `groups` (+ pre-deploy migration and a backfill that maps existing `dust-admins`/`dust-managers` provisioned groups).
- **Sync in `GroupResource`:** `computeUserRoleFromGroups` + `recomputeAndSyncWorkspaceRolesForUsers`, hooked into the member-mutation methods and `setGrantedRole`, so both SCIM/directory sync and manual edits converge. Deleted the old `determineUserRoleFromGroups` and the WorkOS `handleRoleAssignmentForGroup`.
- **Security:** managers cannot manage members of an admin-granting group (privilege escalation to admin) — enforced server-side (`canManageMembersGivenGrantedRole`) and reflected in the group dialog (read-only + warning for managers). Screenshot 2026-09-09 at 17 01 44

- **Last-group safety:** removing the last group granting a role keeps that role on its current holders (`protectRoles`); a role still granted by another group downgrades affected members normally.
- **API:** `PUT /api/w/:wId/groups/:groupId/granted_role` (admin only); `provisioning-status` now returns `grantedRoles[]`.
- **UI:** governance **Roles** section, People → Groups role badge, updated member/invitation role-lock messaging.

UI changes (governance Roles section, groups role badge, group dialog lock) — screenshots to add.

## Tests

- Resource tests (`group_resource.test.ts`): role computation precedence, upgrade on add / downgrade on remove, `setGrantedRole` backfill, orphaned-role protection (last group keeps role; two-admin-groups removes only the orphaned member), and the admin-only membership guard (admin can add, manager blocked, manager still allowed on manager-granting groups).
- Front-api route tests (`granted_role.test.ts`): admin sets/clears, manager 403, invalid role/kind 400, cross-workspace 404.
- Existing group + member-group route tests pass (66). `tsgo` clean on front/front-api/front-spa; biome clean; `cc-check format` clean.

## Risk

- Behavior change: role provisioning is now driven by the `grantedRole` column instead of hardcoded group names; the backfill migration maps legacy `dust-admins`/`dust-managers` groups so existing behavior is preserved.
- The WorkOS `membership.role_updated` audit *event* is no longer emitted for group-driven role changes (GroupResource can't import the audit module without an import cycle); the change is still recorded via `updateMembershipRole`'s structured `auditLog` and the causal `scim.group_user_*` events.
- Rollback-safe: the column is nullable and additive; reverting the code leaves the column harmless.

## Deploy Plan

1. Pre-deploy: run `migrations/pre-deploy/20260909161843_add_granted_role_to_groups.sql` (adds nullable column).
2. Deploy code.
3. Run the backfill: `migrations/20260909_backfill_group_granted_role.ts` with `--execute` (idempotent; dry-run first without the flag).


tdraier:

- #32158 ◀

front/components/members/EditInvitationModal.tsx
```diff
@@ -61,18 +61,17 @@ export function EditInvitationModal({
...
     workspaceId: owner.sId,
   });
 
-  // Check if this invitation's role would be managed by provisioning groups
-  const isRoleManagedByProvisioning =
-    (roleProvisioningStatus.hasAdminGroup && selectedRole === "admin") ||
-    (roleProvisioningStatus.hasManagerGroup && selectedRole === "manager");
+  // Check if this invitation's role would be managed by a role-granting group.
+  const isRoleManagedByGroup =
+    selectedRole !== undefined && grantedRoles.some((r) => r === selectedRole);
```
fabiencelier, line 70: so as soon as you have one group granting admin role you can no longer maange it by hand ?

front/lib/resources/group_resource.ts
```diff
@@ -58,8 +62,33 @@ import type {
...
+ * ┃  IMPORTANT: GroupResource DOES NOT and SHOULD NOT have permissions      ┃
+ * ┃  management of its own.                                                 ┃
+ * ┃                                                                         ┃
+ * ┃  Groups are designed to be used within the context of other resources   ┃
+ * ┃  (e.g., SpaceResource, AgentConfigurationResource). The permissions     ┃
+ * ┃  should be managed at the junction with parent resource level,          ┃
+ * ┃  not at the group level.                                                ┃
+ * ┃                                                                         ┃
+ * ┗━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━┛
+ */
```
fabiencelier: wrongly re-introduce I think

front/lib/resources/group_resource.ts
```diff
@@ -58,8 +62,33 @@ import type {
...
+
+// Privilege ordering used to decide whether a role sync would upgrade or
+// downgrade a member. Higher is more privileged; `builder` is deprecated and
+// ranks with `user`.
+const ROLE_RANK: Record<MembershipRoleType, number> = {
+  admin: 3,
+  manager: 2,
+  builder: 1,
+  user: 1,
+};
```
fabiencelier: We already use the fact that roles are ordered in here:

https://github.com/dust-tt/dust/blob/95f675bb3b48dddbcd30b4a781334633a274550c/front/types/user.ts#L47-L50

Maybe let's unify these to be constent and not have 2 ways of determining role rank

front/lib/resources/group_resource.ts
```diff
@@ -2648,33 +2738,254 @@ export class GroupResource extends BaseResource<GroupModel> {
...
+
+    let role: MembershipRoleType = "user";
+    for (const group of userGroups) {
+      if (group.grantedRole === "admin") {
+        return "admin";
+      }
+      if (group.grantedRole === "manager") {
+        role = "manager";
+      }
+    }
```
fabiencelier: maybe make it more generic and support and get a set of granted roles, then return the max rank one ?

front/lib/resources/group_resource.ts
```diff
@@ -2648,33 +2738,254 @@ export class GroupResource extends BaseResource<GroupModel> {
...
+          workspace,
+          transaction,
+        });
+      if (!currentMembership) {
+        continue;
+      }
+
+      const newRole = await this.computeUserRoleFromGroups(auth, user, {
+        transaction,
+      });
```
fabiencelier: This will refetch all the groups for all the members.
Ok as there should not be too many members in admin/managers group
But you can consider loading all the groups and pre-build a map groupId -> granted role to fetch them only once

front/lib/resources/group_resource.ts
tdraier: weird , messed up with the rebase :-/

front/components/members/EditInvitationModal.tsx
tdraier: that's how it was done before - if managed by groups, can't be manually managed anymore, otherwise groups/roles cannot be in sync

front/lib/resources/group_resource.ts
tdraier: indeed, missed that one

front/lib/resources/group_resource.ts
tdraier: yeah not optimal, fixing

front/lib/resources/group_resource.ts
```diff
@@ -2648,33 +2724,254 @@ export class GroupResource extends BaseResource<GroupModel> {
...
+        continue;
+      }
+
+      const updateResult = await MembershipResource.updateMembershipRole({
+        user,
+        workspace,
+        newRole,
+        allowLastAdminRemoval: true,
+        transaction,
+        author,
```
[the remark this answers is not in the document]
tdraier: @ZeroPath accepted-risk. The exploitable part is fixed: the sync now (1) refuses to change the `admin` role unless the actor is a workspace admin or Poke super user — a manager can no longer demote an admin (or the last admin) via a manager-granting group — and (2) never downgrades the acting user, so no self-lockout by one's own group change.

The residual — an admin- or SSO-authorized group change reconciling a non-mapped admin down to their group-derived role — is intended: roles are group-managed here (the members UI locks manual editing when a group grants the role), matching the prior `dust-admins` behavior.

tdraier:
@ZeroPath the single finding (Issue 1 — Broken Access Control) is triaged on the inline `group_resource.ts` thread. Summary: the exploitable path — a manager demoting an admin (or the last admin) via a manager-granting group, bypassing the direct role-change guard — is fixed with an actor guard on the `admin` role plus a self-lockout guard (a user can't downgrade themselves via their own group action). The residual (an admin- or SSO-authorized group change reconciling a non-mapped admin down to their group-derived role) is accepted-risk: intended group-managed-roles behavior, matching the prior `dust-admins` sync.

fabiencelier (approved):
LGTM ✅

PR #32159, opened 2026-09-09
[connections] Surface the authorizing Microsoft account for managed connections

tdraier:
## Description

Admins cannot verify which Microsoft account authorizes an existing managed Microsoft/SharePoint connection: the flow reports success, but Dust only exposes the *Dust* user who edited the connection, not the Microsoft identity. This has repeatedly forced support to delete + recreate connections (once detaching a data source from 23 agents) just to change or verify the authorizing account.

Fixes dust-tt/tasks#10172.

This extracts the authorizing account at OAuth **finalize** and stores it as `connected_account` in the connection metadata (same `extra_metadata` pattern as Slack/Discord/Gong), then surfaces it:

- **Core** — `MicrosoftConnectionProvider`: decode the Graph access-token JWT and read the account (`upn` → `preferred_username` → `unique_name` → `email`). Extraction is fully non-fatal: any non-JWT/opaque/claim-less token degrades to `None` and finalize proceeds unchanged.
- **Poke** — the data-source details route fetches the OAuth connection metadata for any managed connector and returns `oauthConnectedAccount`, rendered as a "Connected account" row under "Edited by/at" (provider-agnostic; only Microsoft populates it today).
- **Manage Connection modal** — shows `Authorized with Microsoft account <email>`.

App-only tokens (Microsoft service principal) carry no user identity, so `connected_account` is simply absent there.

## Tests

- New Rust unit tests for the extractor: claim precedence, blank-value skipping, app-only case (no account), and opaque/non-JWT/non-object token formats that must degrade to `None` without erroring (`cargo test -p dust --lib oauth::providers::microsoft`, 7 tests, all green).
- `tsgo --noEmit` clean on `front` and `front-api`; `cargo check` clean; biome/rustfmt applied.
- Poke route reuses the existing `getConnectionMetadata` path; the modal reuses the existing `useOAuthMetadata` hook.

## Risk

Low. Additive only — a new optional metadata field and read-only UI. No schema/migration, no change to token exchange, refresh, or scopes (no re-consent). The JWT is decoded without signature verification, but it is obtained directly from Microsoft over TLS and used only for display, never for authorization (mirrors the existing tenant-id decode in connectors). Safe to roll back.

The account is captured at finalize, so **existing connections show nothing until their next re-authorization** ("Edit permissions") — intended; no backfill in this PR.

## Deploy Plan

Standard. Deploy `core` (oauth) first so the metadata is populated on new/re-authorized connections, then `front`/`front-api`. No ordering constraints beyond that; nothing to run.

### DIFF
--- a/front-api/routes/poke/workspaces/[wId]/data_sources/[dsId]/details.ts
+++ b/front-api/routes/poke/workspaces/[wId]/data_sources/[dsId]/details.ts
@@ -11,6 +11,8 @@ import type { InternalConnectorType } from "@app/types/connectors/connectors_api
 import { ConnectorsAPI } from "@app/types/connectors/connectors_api";
 import { isSlackAutoReadPatterns } from "@app/types/connectors/slack";
 import { CoreAPI } from "@app/types/core/core_api";
+import { OAuthAPI } from "@app/types/oauth/oauth_api";
+import { isString } from "@app/types/shared/utils/general";
 import { safeParseJSON } from "@app/types/shared/utils/json_utils";
 import { pokeApp } from "@front-api/middlewares/ctx";
 import { apiError, type HandlerResult } from "@front-api/middlewares/utils";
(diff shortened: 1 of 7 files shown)


front-api/routes/poke/workspaces/[wId]/data_sources/[dsId]/details.ts
```diff
@@ -104,6 +106,31 @@ app.get(
       }
     }
 
+    // Surface the external account authorizing the OAuth connection (e.g. the Microsoft account
+    // behind a SharePoint connection) so admins can verify which identity backs it.
+    let oauthConnectedAccount: string | null = null;
+    if (connector?.connectionId) {
+      const oauthAPI = new OAuthAPI(config.getOAuthAPIConfig(), logger);
+      const metadataRes = await oauthAPI.getConnectionMetadata({
+        connectionId: connector.connectionId,
```
Comment (aubin-tchoi) [inline front-api/routes/poke/workspaces/[wId]/data_sources/[dsId]/details.ts:115]: I think that this will fail for snowflake and bigquery

aubin-tchoi (approved):
LGTM, FYI for Notion there's a concept of `privateIntegrationCredentialId`, which can cause you to display the wrong info I believe

tdraier:
> LGTM, FYI for Notion there's a concept of `privateIntegrationCredentialId`, which can cause you to display the wrong info I believe

:ah: - should I completely remove the info from notion ?

aubin-tchoi:
I don't remember how it works, removing for now could be the simples?

PR #32161, opened 2026-09-09
feat: add v1/wId/analytics/consumption/export

sfriquet:
## Description

Refs https://github.com/dust-tt/tasks/issues/9988, https://github.com/dust-tt/tasks/issues/10156

Add a public API to fetch consumption analytics data, to get granular and filtered export.
Supports `csv` and `ndjson` format to ease programmatic usage.
Unlike the private `export-raw` endpoint, this endpoint is synchronous and meant to be consumed programmatically. 
Results are streamed. The time range is arbitrarily capped at 30 days and there's a 10s timeout.
Under feature flag for now.

## Tests

added tests + tested locally for now.

## Risk

Low while under feature flag.

## Deploy Plan

Front + Docs

### DIFF
--- a/front-api/routes/v1/w/[wId]/analytics/consumption/export.ts
+++ b/front-api/routes/v1/w/[wId]/analytics/consumption/export.ts
@@ -1,0 +1,99 @@
+/** @ignoreswagger */
+
+import {
+  fetchConsumptionExportRows,
+  rowsToCsvString,
+  rowsToNdjson,
+} from "@app/lib/api/analytics/consumption/export_lines";
+import logger from "@app/logger/logger";
+import { PostConsumptionExportRequestSchema } from "@dust-tt/client";
+import { publicApiApp } from "@front-api/middlewares/ctx";
+import { ensureIsAdmin } from "@front-api/middlewares/ensure_role";
+import { apiError } from "@front-api/middlewares/utils";
+import { validate } from "@front-api/middlewares/validator";
+
+// Mounted at /api/v1/w/:wId/analytics/consumption/export.
+const app = publicApiApp();
+
+// No swagger doc while this endpoint is under feature flag.
+app.post(
+  "/",
+  ensureIsAdmin(),
+  validate("json", PostConsumptionExportRequestSchema),
+  async (ctx) => {
+    const auth = ctx.get("auth");
+
+    if (!(await auth.hasFeatureFlag("consumption_export_api"))) {
+      return apiError(ctx, {
+        status_code: 403,
+        api_error: {
+          type: "workspace_auth_error",
+          message:
+            "The workspace does not have access to the consumption export API.",
+        },
+      });
+    }
+
+    const body = ctx.req.valid("json");
+    const format = body.format ?? "csv";
+
+    const startDate = new Date(body.startDate).toISOString();
+    const endDate = new Date(body.endDate).toISOString();
+
+    const owner = auth.getNonNullableWorkspace();
+
+    logger.info(
+      {
+        workspaceId: owner.sId,
+        startDate,
+        endDate,
+        format,
+        hasFilter: !!body.filter,
+      },
+      "Consumption export requested."
+    );
+
+    const result = await fetchConsumptionExportRows(auth, {
+      period: { startDate, endDate },
+      filter: body.filter,
+    });
+
+    if (result.isErr()) {
+      return apiError(ctx, {
+        status_code: 500,
+        api_error: {
+          type: "internal_server_error",
+          message: result.error.message,
+        },
+      });
+    }
+
+    if (format === "ndjson") {
+      ctx.header("Content-Type", "application/x-ndjson");
+      ctx.header(
+        "Content-Disposition",
+        `attachment; filename="dust_consumption_${body.startDate}_${body.endDate}.ndjson"`
+      );
+      return ctx.body(rowsToNdjson(result.value));
+    }
+
+    ctx.header("Content-Type", "text/csv");
+    ctx.header(
+      "Content-Disposition",
+      `attachment; filename="dust_consumption_${body.startDate}_${body.endDate}.csv"`
+    );
+    return ctx.body(rowsToCsvString(result.value));
+  }
+);
+
+app.all("/", (ctx) =>
+  apiError(ctx, {
+    status_code: 405,
+    api_error: {
+      type: "method_not_supported_error",
+      message: "The method passed is not supported, POST is expected.",
+    },
+  })
+);
+
+export default app;
--- a/front-api/routes/v1/w/[wId]/analytics/consumption/export.ts
+++ b/front-api/routes/v1/w/[wId]/analytics/consumption/export.ts
@@ -1,14 +1,12 @@
-import {
-  fetchConsumptionExportRows,
-  rowsToCsvString,
-  rowsToNdjson,
-} from "@app/lib/api/analytics/consumption/export_lines";
+import { streamConsumptionExport } from "@app/lib/api/analytics/consumption/export_lines";
 import logger from "@app/logger/logger";
 import { PostConsumptionExportRequestSchema } from "@dust-tt/client";
 import { publicApiApp } from "@front-api/middlewares/ctx";
 import { ensureIsAdmin } from "@front-api/middlewares/ensure_role";
 import { apiError } from "@front-api/middlewares/utils";
 import { validate } from "@front-api/middlewares/validator";
+
+const EXPORT_TIMEOUT_MS = 10_000;
 
 // Mounted at /api/v1/w/:wId/analytics/consumption/export.
 const app = publicApiApp();
@@ -22,7 +20,9 @@ const app = publicApiApp();
  *       Export per-call consumption analytics for the workspace identified by {wId}.
  *       Each row represents one unit of billed credit consumption (an LLM call or a tool call).
  *       The export can be filtered by various dimensions (agents, users, API keys, groups, models, tools, skills, sources, tags).
- *       The export is limited to a maximum of 30 days per request.
+ *       The export is limited to a maximum of 30 days per request and times out after 10 seconds: reduce the time range
+ *       or apply filters to reduce the number of rows if you encounter a timeout.
+ *       Results are streamed, if an error occurs, an error message is appended and the stream is closed.
  *     tags:
  *       - Analytics
  *     security:
@@ -143,6 +143,17 @@ app.post(
       });
     }
 
+    if (!auth.isKey()) {
+      return apiError(ctx, {
+        status_code: 403,
+        api_error: {
+          type: "workspace_auth_error",
+          message:
+            "Workspace analytics export requires API key authentication.",
+        },
+      });
+    }
+
     const body = ctx.req.valid("json");
     const format = body.format ?? "csv";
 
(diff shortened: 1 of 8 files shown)


front-api/routes/v1/w/[wId]/analytics/consumption/export.ts
```diff
@@ -0,0 +1,208 @@
...
+ *       400:
+ *         description: Invalid request body (missing fields, invalid dates, range exceeds 30 days)
+ *       403:
+ *         description: Requires an API key with admin scope
+ *       500:
+ *         description: Internal Server Error
+ */
+app.post(
+  "/",
+  ensureIsAdmin(),
```
Comment (avervaet) [inline front-api/routes/v1/w/[wId]/analytics/consumption/export.ts:141]: ensureIsAdmin() also authorizes OAuth workspace-admin sessions, although this public endpoint is documented as requiring an admin API key and the existing analytics export explicitly enforces API-key authentication. We could reuse what front-api/routes/v1/w/[wId]/analytics/export.ts:95-106 is doing

front-api/routes/v1/w/[wId]/analytics/consumption/export.ts
```diff
@@ -0,0 +1,208 @@
...
+        workspaceId: owner.sId,
+        startDate,
+        endDate,
+        format,
+        hasFilter: !!body.filter,
+      },
+      "Consumption export requested."
+    );
+
+    const result = await fetchConsumptionExportRows(auth, {
```
Comment (avervaet) [inline front-api/routes/v1/w/[wId]/analytics/consumption/export.ts]: did we try these with Dust workspace ? Loading every matching document for a 30-day period into memory before serializing can put a lot of pressure on ES

Comment (avervaet) [reply front-api/routes/v1/w/[wId]/analytics/consumption/export.ts]: Maybe we want to add a failsafe or a limit cap

sdks/js/src/types.ts
```diff
@@ -3241,6 +3242,50 @@ export type GetAnalyticsExportRequestType = z.infer<
...
+    message: "Must be a valid ISO 8601 datetime (e.g. 2026-01-15T00:00:00Z)",
+  });
+
+export const PostConsumptionExportRequestSchema = z
+  .object({
+    startDate: IsoDateTimeSchema,
+    endDate: IsoDateTimeSchema,
+    format: z.enum(["csv", "ndjson"]).optional(),
+    filter: z
+      .record(ConsumptionExportFilterKeySchema, z.string().array())
```
avervaet: every filter dimension accepts an unlimited number of arbitrary-length strings that are passed into Elasticsearch terms clauses, allowing oversized requests and expensive queries. 

We could cap each dimension using the existing 500-value limit, bound individual string lengths, and enforce a total filter-size limit ?

front-api/routes/v1/w/[wId]/analytics/consumption/export.ts
```diff
@@ -0,0 +1,208 @@
...
+      period: { startDate, endDate },
+      filter: body.filter,
+    });
+
+    if (result.isErr()) {
+      return apiError(ctx, {
+        status_code: 500,
+        api_error: {
+          type: "internal_server_error",
+          message: result.error.message,
```
Comment (avervaet) [inline front-api/routes/v1/w/[wId]/analytics/consumption/export.ts]: Do we really want to exposes Elasticsearch or other backend implementation details to API consumers ?

sdks/js/src/types.ts
sfriquet: added limit on number of values across filter + individual filter length

front-api/routes/v1/w/[wId]/analytics/consumption/export.ts
Comment (sfriquet) [reply front-api/routes/v1/w/[wId]/analytics/consumption/export.ts]: now streaming with a 10s timeout

front/lib/api/analytics/consumption/export_lines.ts
```diff
@@ -287,12 +316,127 @@ export async function fetchConsumptionExportBucketCsv(
...
+  return new ReadableStream({
+    async pull(controller) {
+      try {
+        const { value, done } = await gen.next();
+        if (done) {
+          controller.close();
+        } else {
+          controller.enqueue(value);
+        }
+      } catch (err) {
```
avervaet, line 493: The stream's pull wraps the generator in a try/catch. On any exception it logs and calls controller.close(), which tells the client "end of file, all good". Only Elasticsearch failures come back as a Result and hit the explicit error branch. I don't think we want to return OK if only a portion of the data has been fetched due to an error.

front/lib/api/analytics/consumption/export_lines.ts
avervaet, line 1: The route starts a timer that flips an AbortSignal after 10s. But, the generator only looks at that signal at the top of each page iteration. The signal is never handed to searchConsumptionAnalytics, whose options accept no signal, and the ES client is built without a request timeout, so its default of 30s applies. 

A single slow 10k-doc page therefore runs to completion regardless. The abort is only noticed on the next iteration, after that page has already been flushed. Since the stream is pull-based, time spent by a slow client draining chunks also counts against the budget. 

Also, the timer is never cleared, so it fires uselessly after every fast export.

front/lib/api/analytics/consumption/export_lines.ts
```diff
@@ -287,12 +316,127 @@ export async function fetchConsumptionExportBucketCsv(
...
+    let pageSize: number;
+    let isFirst = true;
+
+    do {
+      if (cancelled) {
+        // Cancelled by the client, stop streaming.
+        return;
+      }
+      if (signal?.aborted) {
+        yield encoder.encode("ERROR: Export timed out.");
```
avervaet, line 449: When ES fails or the timeout fires, the generator yields the text ERROR: Internal server error. into a body already declared as text/csv or application/x-ndjson. A CSV loader reads that as a one-column row.

front-api/routes/v1/w/[wId]/analytics/consumption/export.ts
```diff
@@ -0,0 +1,209 @@
...
+          message:
+            "Workspace analytics export requires API key authentication.",
+        },
+      });
+    }
+
+    const body = ctx.req.valid("json");
+    const format = body.format ?? "csv";
+
+    const startDate = new Date(body.startDate).toISOString();
```
Comment (avervaet) [inline front-api/routes/v1/w/[wId]/analytics/consumption/export.ts:171]: Here we are normalizing startDate and endDate to ISO strings. But Line 193 ignores those and interpolates body.startDate and body.endDate into the Content-Disposition header. 

Because the schema only checks !isNaN(new Date(s)), values like 06/01/2024 or 1 Jun 2024 (") pass validation. The resulting header contains slashes, spaces, or an unbalanced quote, which browsers and curl -J mangle or reject.

sdks/js/src/types.ts
```diff
@@ -3241,6 +3242,64 @@ export type GetAnalyticsExportRequestType = z.infer<
...
+  "api_keys",
+  "groups",
+  "models",
+  "tools",
+  "skills",
+  "sources",
+  "tags",
+]);
+
+const IsoDateTimeSchema = z
```
avervaet, line 3268: "Must be a valid ISO 8601 datetime" but the check accepts anything V8 can parse. Non-ISO forms like 06/01/2024 are interpreted in the server's local timezone, so the same day spelled two ways yields two different windows on a non-UTC host. Very large years like +275760-09-13T00:00:00Z are accepted, normalize to a form ES cannot parse, and produce a 200 with an error body instead of a 400.

front-api/routes/v1/w/[wId]/analytics/consumption/export.ts
```diff
@@ -0,0 +1,209 @@
...
+ *         description: Internal Server Error
+ */
+app.post(
+  "/",
+  ensureIsAdmin(),
+  validate("json", PostConsumptionExportRequestSchema),
+  async (ctx) => {
+    const auth = ctx.get("auth");
+
+    if (!(await auth.hasFeatureFlag("consumption_export_api"))) {
```
Comment (avervaet) [inline front-api/routes/v1/w/[wId]/analytics/consumption/export.ts]: The handler checks the flag inline and returns 403 workspace_auth_error. The middleware withFeatureFlag in [with_feature_flag.ts](vscode-webview://1kbbslh48ss885vmbv2sfm0unevqnu9843dvuctk7ubendum6kcf/front-api/middlewares/with_feature_flag.ts) does the same thing and returns feature_flag_not_found. 

With the inline version a client cannot tell "flag not enabled" from "not admin" since both are 403 workspace_auth_error. The check also runs after validate, so a workspace without the flag gets 400s that reveal the schema shape before learning the endpoint is unavailable.

sdks/js/src/types.ts
```diff
@@ -3241,6 +3242,64 @@ export type GetAnalyticsExportRequestType = z.infer<
...
+  .refine(
+    (d) => {
+      if (!d.filter) {
+        return true;
+      }
+      const total = Object.values(d.filter).reduce(
+        (sum, arr) => sum + arr.length,
+        0
+      );
+      // Aligned with CONSUMPTION_FILTER_MAX_VALUES_PER_DIMENSION in front.
```
avervaet, line 3304: here we are capping at 500 per dimension not 500 total.

front/lib/api/analytics/consumption/export_lines.ts
```diff
@@ -274,6 +277,32 @@ export async function fetchConsumptionExportBucketCsv(
...
+      includeHeader: false,
+    })
+  );
+}
+
+export function buildConsumptionLineExportCsvHeader(): string {
+  return rowsToCsv(CONSUMPTION_LINE_EXPORT_HEADERS, []);
+}
+
+export async function fetchConsumptionExportRows(
```
avervaet: exported but not consumed anywhere

front/lib/api/analytics/consumption/export_lines.ts
```diff
@@ -287,12 +316,127 @@ export async function fetchConsumptionExportBucketCsv(
...
+      const docs: AgentMessageConsumptionAnalyticsData[] = [];
+      // Accumulate hits to call buildConsumptionLineExportRows once per page,
+      // since it needs to resolve labels.
+      for (const hit of hits) {
+        if (hit._source) {
+          docs.push(hit._source);
+        }
+      }
+
+      const rows = await buildConsumptionLineExportRows(auth, docs);
```
avervaet: Each 10k-row page calls buildConsumptionLineExportRows, which fans out to seven label lookups, five of them hitting Postgres. The group branch additionally lists all workspace groups and computes member counts, which loads every active membership, to produce a count the export never uses. 

A 30-day export of a busy workspace runs tens of pages, so this is hundreds of queries for keys that are nearly identical page to page.

front/lib/api/analytics/consumption/export_lines.ts
```diff
@@ -287,12 +316,127 @@ export async function fetchConsumptionExportBucketCsv(
...
+          },
+          "[Consumption Export] Failed to stream consumption lines"
+        );
+        yield encoder.encode("ERROR: Internal server error.");
+        return;
+      }
+
+      const { hits } = result.value.hits;
+      const docs: AgentMessageConsumptionAnalyticsData[] = [];
+      // Accumulate hits to call buildConsumptionLineExportRows once per page,
```
avervaet: we tend to avoid adding fucntions names into comment thoses ages badly

front/lib/api/analytics/consumption/export_lines.ts
sfriquet: done on purpose, we can't change the http status once we start streaming, the error may break formating which makes it harder to miss imo.

sdks/js/src/types.ts
sfriquet: +275760-09-13T00:00:00Z works, but updating the PR to have the first page throwing an actual error so we can't start streaming with invalid parameters

sdks/js/src/types.ts
sfriquet: it's 500 across dimensions, adding another test

front/lib/api/analytics/consumption/export_lines.ts
sfriquet: not sure how big of a problem that'll be... adding a cache for good measures

front/types/shared/feature_flags.ts
```diff
@@ -431,6 +431,12 @@ export const WHITELISTABLE_FEATURES_CONFIG = {
     stage: "ask_owner",
     owner: "avervaet",
   },
+  consumption_export_api: {
+    description:
+      "Enable the public API endpoint for raw consumption analytics export (POST /api/v1/w/:wId/analytics/consumption/export).",
+    stage: "ask_owner",
+    owner: "sylvain",
```
avervaet: I thing the convention is to use your github handle here

avervaet (approved):
LGTM

PR #32163, opened 2026-09-09
feat(frames): add an open-in-new-tab button next to the full screen button

Nils-Fedrigo:
## Description

Adds an "Open in a new tab" button to the frame preview action buttons, right below the full screen
toggle. It opens the current conversation URL in a new tab with the interactive content side panel
pointed at the same frame and full screen mode on, so a frame can be moved to its own browser tab
without going through sharing.

The button is hidden in the extension client, like the full screen toggle, since the extension side
panel has no dust.tt location to derive the URL from.

## Tests

Unit test in `FrameRenderer.test.tsx` covering the generated URL (origin/pathname/search preserved,
`spid` / `spt` / `fullScreen` hash params set).

Manually: open a frame in a conversation, click the new button, and check that the new tab lands on
the same frame, full screen.

### DIFF
--- a/front/components/assistant/conversation/interactive_content/FrameRenderer.tsx
+++ b/front/components/assistant/conversation/interactive_content/FrameRenderer.tsx
@@ -15,7 +15,11 @@ import { useSendNotification } from "@app/hooks/useNotification";
 import { useAuth, useFeatureFlags } from "@app/lib/auth/AuthContext";
 import { useClientType } from "@app/lib/context/clientType";
 import { clientFetch } from "@app/lib/egress/client";
-import { useFileContent, useFileMetadata } from "@app/lib/swr/files";
+import {
+  useFileContent,
+  useFileMetadata,
+  useShareInteractiveContentFile,
+} from "@app/lib/swr/files";
 import { useEditFrameText, useFramePermissions } from "@app/lib/swr/frames";
 import { usePodFiles } from "@app/lib/swr/pods";
 import { useSpaceInfo } from "@app/lib/swr/spaces";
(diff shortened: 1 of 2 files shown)


front/components/assistant/conversation/interactive_content/FrameRenderer.tsx
```diff
@@ -143,6 +148,14 @@ export function FrameRenderer({
     owner,
   });
 
+  // Same share URL the share sheet hands out; the sheet in the header already fetches it, so
+  // this reuses that SWR entry rather than issuing a second request.
```
ykmsd: I find this comment useless 🙃

front/components/assistant/conversation/interactive_content/FrameRenderer.tsx
```diff
@@ -505,6 +519,8 @@ interface PreviewActionButtonsProps {
   isFullScreen: boolean;
   enterFullScreen: () => void;
   exitFullScreen: () => void;
+  // The frame's share URL, once loaded. Absent while it loads.
```
ykmsd: this too

front/components/assistant/conversation/interactive_content/FrameRenderer.tsx
```diff
@@ -535,6 +552,24 @@ function PreviewActionButtons({
           }
         />
       )}
+      {clientType !== "extension" && shareUrl && (
```
ykmsd: I'm not a big fan that the height changes when url is ready, what do you think of showing this all the time but either disabled on click or do nothing while it's loading? I imagine loading time is short and people rarely click that fast?

ykmsd (approved):

front/components/assistant/conversation/interactive_content/FrameRenderer.tsx
Nils-Fedrigo: > do nothing while it's loading

yeah indeed this is better!

PR #32165, opened 2026-09-09
Normalize caught errors and guard provider failures

spolu:
## Description

Caught values were cast to error types or accessed without runtime validation, which could mask the original failure with another exception. Fix 34 direct `normalize-caught-errors` violations across 28 files: 23 sites use `normalizeError`, and 11 use runtime guards. Provider-specific handlers validate Google, Axios, and Slack errors before inspecting their fields.

## Tests

- Five Slack error guard tests passed, including recognized sync/permission errors and malformed inputs, using an isolated Node test configuration.
- Re-scanned all 34 targeted sites to verify normalization or runtime guards.
- The WebhookRequestResource suite could not load because the local environment lacks `@novu/framework`. Full hive validation is also blocked by missing `bun-types`.

## Risk

Low. Changes are limited to error handling. Known provider failures retain their existing handling; unexpected values are normalized or propagated.

## Deploy Plan

Deploy `connectors`, `front-api`, `front-spa`, and front workers. Tooling changes take effect when those tools are next run.

### DIFF
--- a/cli/dust-sandbox/e2e/smoke.ts
+++ b/cli/dust-sandbox/e2e/smoke.ts
@@ -184,9 +184,8 @@ async function startForward(): Promise<ForwardHandle> {
     await waitForListen(LISTEN_ADDR, FORWARD_READY_TIMEOUT_MS);
   } catch (err) {
     const stderr = await stop();
-    throw new Error(
-      `${(err as Error).message}\ndsbx forward stderr:\n${stderr}`
-    );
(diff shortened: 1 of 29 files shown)


spolu:
(Tweaking the skill to not report all pre-existing violation only ones related to the PR work)

spolu:
Happening here: https://github.com/spolu/code-contracts/pull/36/changes

PR #32166, opened 2026-09-09
[sparkle] DataTable: add a "Load more" footer with an animated row reveal

radjakahoul:
## Description

Adds a "Load more" footer to `DataTable`, as an alternative to `pagination` for tables that append rows instead of swapping pages.

```tsx
<DataTable
  data={rows}
  totalRowCount={total}
  onLoadMore={handleLoadMore}
  isLoadingMore={isLoadingMore}
  columns={columns}
/>
```

`onLoadMore` renders the control in the same footer slot `Pagination` uses, and is ignored when `pagination` is set, so the two can't double up.

### `LoadMore`

A new component that mirrors `Pagination`'s layout and typography:

- the label carries the same `text-xs font-medium` as the page numbers, in `text-primary-400` with a `hover:text-foreground`
- the item count sits on the right, in the same spot and the same muted style
- once every row is loaded the control hides and the count remains — the way `Pagination` behaves when there's only one page

While loading, the label animates three dots (new `--animate-loading-dot`) instead of showing a spinner, which sat too heavily next to the small type. It falls back to a static `Loading...` under `prefers-reduced-motion`.

### Row reveal

Appended rows are revealed by animating the **table's height**, so they slide into view at exactly the rate the footer moves down. Animating the footer instead — the first thing I tried — let the rows pop in at full height while the line took 300ms to catch up.

Two details worth a reviewer's eye:

- A `ResizeObserver` keeps the measured height current through every layout change, not just row changes. `DataTable` renders no cells on the first commit (`useWindowSize` hasn't measured yet), so a height recorded only on mount is the height of a table with **no columns** — the first reveal then animates up from nothing. `ResizeObserver` callbacks run after layout effects, so the value read during the reveal is reliably the height from before the new rows landed.
- The reveal is keyed on the *rendered* row count, not `data.length`, so filtering keeps the measurement in sync.

It's gated on `!!onLoadMore && !pagination`, so paginated and plain tables are untouched. `DataTable.Root` gains a `containerRef` prop to carry the ref — this avoids a wrapper div, and kept the diff off ~200 lines of re-indented JSX.

### Risk

- `overflow: hidden` sits on the table container for the 300ms the animation runs, cleared on `transitionend`. A row dropdown opened during that window would clip.
- `ScrollableDataTable` already had its own `onLoadMore` meaning "infinite scroll on intersection". Its props now `Omit` the two new ones so the meanings can't collide.

## Tests

Typecheck and biome pass. New story: `DataTable / Data Table Load More Example` — loads 2 rows at a time behind a simulated 600ms round-trip.

Checked by hand in Storybook: first click and subsequent clicks (the first was the case the stale baseline broke), and the fully-loaded end state.

### DIFF
--- a/sparkle/src/components/LoadMore.tsx
+++ b/sparkle/src/components/LoadMore.tsx
@@ -1,0 +1,86 @@
+import { cn } from "@sparkle/lib/utils";
+import React from "react";
+
+const LOADING_DOT_DELAYS = ["0ms", "250ms", "500ms"];
+
+interface LoadMoreProps {
+  showDetails?: boolean;
+  /** Number of rows currently loaded. */
+  rowCount: number;
+  /** Total number of rows available, when known. */
+  totalRowCount?: number;
+  totalRowCountIsCapped?: boolean;
+  isLoading?: boolean;
+  onLoadMore: () => void;
+  label?: string;
+  loadingLabel?: string;
+}
+
+export function LoadMore({
+  showDetails = true,
+  rowCount,
+  totalRowCount,
+  totalRowCountIsCapped = false,
+  isLoading = false,
+  onLoadMore,
+  label = "Load more",
+  loadingLabel = "Loading",
+}: LoadMoreProps) {
+  // When the total is known and everything is loaded, there is nothing left to
+  // fetch: keep the details, hide the control (same behavior as Pagination).
+  const controlIsHidden =
+    totalRowCount !== undefined &&
+    !totalRowCountIsCapped &&
+    rowCount >= totalRowCount;
+
+  return (
+    <div
+      className={cn(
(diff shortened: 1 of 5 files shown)


avervaet (approved):
LGTM

sparkle/src/components/LoadMore.tsx
```diff
@@ -0,0 +1,86 @@
...
+    rowCount >= totalRowCount;
+
+  return (
+    <div
+      className={cn(
+        "flex w-full items-center",
+        controlIsHidden ? "justify-end" : "justify-between"
+      )}
+    >
+      <button
```
avervaet, line 43: Fixed in 6bd03de18b — added explicit `type="button"` to the LoadMore control.

PR #32168, opened 2026-09-09
feat(dusthivecat): add Pompon

radjakahoul:
## Description

Adds **Pompon** as a selectable pet in DustHiveCat, and makes the status bar icon follow the selected pet.

**Pompon** is a fluffy purple ball, and original artwork — no character design or sprite from any existing property is involved. (This PR previously added Ditto; the branch has been rebuilt from `main`, so that art is not in its history.)

- One entry in `CatType.allCats`, so it shows up in the status-bar Pet menu automatically. The README's pet list is updated to match — it was also missing Lou, so that is corrected in passing.
- 13 sprites under `Resources/pompon/`: 6 walk frames, 6 notification frames, 1 sleep frame. Plus an idle frame and 6 animation frames for the status bar.

**Status bar icon** — the icon was read from a fixed `statusbar/` folder, so every pet showed the cat regardless of what was selected in the Pet menu. It now looks for `statusbar/<petId>/` first and falls back to the existing shared icons, leaving pets without their own set unaffected. The icon already re-renders on `.catPreferencesChanged`, which `catType`'s setter posts, so switching pet updates the status bar immediately rather than at next launch.

**Speed menu** gains `0.1x` and `0.25x`. The `speed` preference was never clamped in code (`value > 0 ? value : 1.0`) — `0.5` was simply the lowest entry in the menu's hardcoded list, so anything slower was unreachable from the UI and would be reset by the next menu click.

## On the sprites

They are generated by `Scripts/generate_pompon.py`, committed alongside them. The other pets ship as bare PNGs and I would normally have kept that convention, but here the script is the provenance: every pixel is a function of the maths in it, so the drawing can be re-derived and audited rather than taken on trust. Running it reproduces the committed PNGs byte for byte.

The silhouette is a union of overlapping circles — a small core plus a ring of seven tufts that scallop past it — which is what gives the rim its fluff. Shading is a crescent that follows the silhouette rather than a band across the body. Walk and notification are a 6-frame squash-and-stretch hop with the bottom edge pinned to a baseline, so Pompon squashes against the ground instead of floating.

Notification widens the eyes, opens the mouth and adds a yellow lightning bolt with small pink sparkles twinkling around it. While walking, Pompon lets go of a single pink sparkle that drifts up and away — it appears on three of the six frames, so it reads as an occasional puff rather than a permanent decoration. Sleep flattens the body, closes the eyes and adds three rising z's. Status bar frames use a coarser grid and carry no bolt or sparkles, since those turn to noise at 22pt.

Each frame is drawn directly at its final squash rather than being scaled from a neutral pose. The eyes carry a 1px glint and the rim is one cell wide, so nearest-neighbour resampling would drop whole rows and columns — deleting a glint or exposing body colour at the edge depending on the frame.

## Tests

Automated / scripted:
- `swift build` from a fresh worktree off `main` — clean, and all 20 PNGs land in the built bundle.
- Re-running `Scripts/generate_pompon.py` reproduces every committed PNG byte for byte.
- Over the rendered PNGs: all are 512×512, none touches the canvas edge, none is empty.
- Frame naming matches `SpriteManager`'s `%02d_<id>_<anim>.png`, and the walk / notification / sleep set matches what every pet except `lou` ships.

Manual, on macOS:
- Installed via `dustcat update` with Pompon selected, and confirmed at runtime that walk, sleep and notification frames all resolve — the app logs `No frames loaded for <pet> <anim>, using placeholder` when they don't, and it logs nothing.
- Triggered `dustcat://notify` and `dustcat://reset`.

- Checked status bar icon resolution for all seven pets: Pompon resolves its own `statusbar/pompon/` set, and the six pets without one fall back to the shared icons. No pet has a gap in its seven-icon set.

## Risk

Low. Confined to `x/daph/dust-hive-cat`, which isn't part of any deployed service. Two changes touch existing behaviour: two extra entries in the Speed menu, and the status bar icon lookup — the latter falls back to the current shared icons, so every existing pet keeps the icon it has today. The default pet stays Soupinou. Rollback is a revert.

## Deploy Plan

None — nothing here is deployed. Existing users pick it up on their next `dustcat update`, and the pet is opt-in from the Pet menu.


radjakahoul:
Closing: this was on the wrong repo. DustHiveCat under `x/daph/` is a stale vendored snapshot of [Soupinou/miaou](https://github.com/Soupinou/miaou), which is where the app actually lives.

Redone properly there: **https://github.com/Soupinou/miaou/pull/18**

The port was not a copy — the vendored copy was old enough that moving the sprites across unchanged would have shipped a broken pet:
- the art was undersized for that repo's ~400x350 convention (338x286 against Soupinou's 408x353), so the ball was regenerated at 429x364
- `lifted` and `lifted_idle` don't exist in this snapshot; without them the pet goes stiff when dragged by the scruff, so both were added (25 sprites rather than 13)

The status bar and Speed menu changes carried over to that PR too.

PR #32169, opened 2026-09-09
fix(marketing): server-render marketing pages so meta tags reach crawlers

adrsimon:
## Description

No marketing page was server-rendered. Crawlers got `next-head-count=2` and an empty `<div id=\"__next\">` on `/`, `/home`, `/pricing`, `/blog`, `/customers`, `/security`, so none of the OG, Twitter or `article:*` tags we write in the page components ever reached the HTML. Link previews and indexing worked off nothing.

Cause: `_app.tsx` loaded `PostHogTracker` via `dynamic(..., { ssr: false })` and wrapped the whole app in it. That combination does not hide children on the server, it discards them, so every page, layout and `<Head>` was dropped from the server output. Introduced in d67604229a (#26962).

Fix: the tracker no longer wraps anything. `PostHogProvider` renders normally, and the client-only effects render as a null sibling next to the page tree.

While in there, seven pages built `<title>` from several JSX children. `<title>` is RCDATA, so React's `<!-- -->` separators are parsed as literal text and end up visible in the tab and in search results. Each is now a single template string.

## Tests

Ran the marketing app locally and checked the served HTML (not the hydrated DOM) on the six routes above: head count went from 2 to 21-35 and the `__next` body from 0 to 55-119 KB. The customer story page serves 6 OG and 4 Twitter tags. No `<!-- -->` left in any title. `next build` prerenders 39/39 pages and First Load JS shared is unchanged at 339 kB.

PostHog capture is not verified locally: `NEXT_PUBLIC_POSTHOG_KEY` is unset in my environment so `posthog.init` short-circuits. Worth a look at pageview events on the preview deploy.

## Risk

Low on the rendering side, the build prerenders every page and the pages themselves are untouched. The one thing to watch is PostHog: the provider now mounts during SSR instead of being skipped, and the effects mount as a sibling rather than a parent. Any regression shows up as missing pageview events, not as a broken page.

## Deploy Plan

Deploy marketing.

### DIFF
--- a/marketing/components/app/PostHogTracker.tsx
+++ b/marketing/components/app/PostHogTracker.tsx
@@ -15,7 +15,6 @@ import {
 } from "@marketing/lib/utils/utm";
 import { isString } from "@marketing/types/shared/utils/general";
 import posthog from "posthog-js";
-import { PostHogProvider } from "posthog-js/react";
 import { useEffect, useMemo, useRef } from "react";
 import { useCookies } from "react-cookie";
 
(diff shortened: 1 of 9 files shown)


tdraier (approved):
looks reasonable

spolu:
(I eased the review process to only pick up diff relevant contract violations, should be less noisy on anything rebased on main)

PR #32171, opened 2026-09-09
A deploy thread that reads at a glance

flvndvd:
## Description

**Today** the deploy thread opens with a four bullet header that names the commit twice, and the build message says "Build completed for all", "all" being the cell list.

**After** the header is one line, `🚀 front b6cf5d9 → all · @flavien · workflow`, the tag linking to the commit, and the build line is `🔨 b6cf5d9 built · 8384b67 → b6cf5d9, 3 commits`. Region builds, marketing, still say which region. The cell lines and the recap that follow come from dust-tt/dust-infra#1082.

## Tests

Read through, actionlint clean. Visual proof on the next deploy.

## Risk

Slack text only.

## Deploy Plan

Merge with dust-tt/dust-infra#1082, any order.

### DIFF
--- a/.github/actions/slack-notify/action.yml
+++ b/.github/actions/slack-notify/action.yml
@@ -96,12 +96,7 @@ runs:
         token: ${{ inputs.slack_token }}
         payload: |
           channel: ${{ inputs.channel }}
-          text: |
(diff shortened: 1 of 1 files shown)


Fraggle:
Love that

PR #32172, opened 2026-09-09
Disable triggers targeting a Pod when it is archived

frankaloia:
## Summary
- When a Pod is archived, disable enabled triggers that target it so they stop opening new conversations there.
- Leave those triggers attached to the Pod (do not retarget to personal conversations) and do not re-enable them on unarchive.
- Archive confirmation copy now tells editors that those triggers will be disabled.
- https://github.com/dust-tt/tasks/issues/10343

r? aubin-tchoi

## Testing
Validated disable locally

## Risk
Low

## Deploy Plan
1. Deploy front


aubin-tchoi (approved):
LGTM

PR #32177, opened 2026-09-09
[core] Delete a project's runs when deleting the project

smb2268:
## Description

Core's `delete_project` cascaded datasets, cache and specifications but not runs, so deleting
any project with run history failed with:

```
Failed to delete project (error: db error: ERROR: update or delete on table "projects" violates foreign key constraint "runs_project_fkey" on table "runs")
```

`hardDeleteApp` deletes the Core project before the front rows, so this blocked the space scrub
workflow for every space containing an app that had ever been run.

This adds a `delete_project_runs` SQL function (mirroring `delete_project_datasets`) that removes
the project's `runs_joins`, the referenced `block_executions` and the `runs`, and calls it from
`delete_project`. Block executions are reached through `runs_joins` only, since
`block_executions` has no index on `project`.

## Tests

- `cargo check` and `cargo fmt --check` on core.
- Exercised the new function against a local Core database: a project with two runs, two block
  executions, joins and a specification was fully deleted; a control project's runs, joins and
  block executions were untouched.

## Risk

Low. SQL functions are registered with `CREATE OR REPLACE` at Core startup, so no migration is
needed. Deleting a project with a very large run history does more work inside one transaction
than before.

## Deploy Plan

Deploy core. Then relaunch the stuck space scrub with a new Temporal run via
`./admin/cli.sh space scrub`; the terminated run will not resume. If it fails on
`block_executions_project_fkey`, that indicates orphaned block executions with no join row, which
this change deliberately does not scan for.


smb2268:
@frankaloia r?

frankaloia (approved):

PR #32178, opened 2026-09-09
Make missing upsert queue payloads non-retryable

smb2268:
## Description

Supersedes `32089` (same change, rebranched off main).

Upsert queue workflows retry forever when their GCS payload has disappeared from the staging bucket: `download()` throws a 404, Temporal retries the activity, and the 404 repeats indefinitely. Two examples:
- https://cloud.temporal.io/namespaces/dust-front-prod.gmnlm/workflows/upsert-queue-document-HFFSyjxbZ5-dts_bnZYniqvhA4QK-cc36cf8f-dff7-4ded-a6ba-bf0bb6a661c5/01a0438d-3b99-75a4-94b1-0ef184d22139/history
- https://cloud.temporal.io/namespaces/dust-front-prod.gmnlm/workflows/upsert-queue-document-1616590054-dts_kQKMTAfAKGCSP-b5950c30-6d98-4ec8-8996-896987146677/01a045a9-7f64-7b63-95ff-f50f40af8693/history

**Root cause:** _[TODO: why payloads disappear from the bucket — lifecycle rule vs. late execution vs. external deletion]_. Once the object is gone, no retry can succeed.

**What this PR does:**
1. Adds `fetchUpsertQueuePayload` in `lib/upsert_queue.ts`, next to the write, reusing the existing `isGCSNotFoundError` typeguard. It returns `null` for a missing object and rethrows everything else.
2. `upsertDocumentActivity` and `upsertTableActivity` both use it. On a missing payload they log `upsertQueueId` + `delaySinceEnqueueMs`, emit `upsert_queue_payload_missing.count` (tagged `kind:document` / `kind:table`), and throw a non-retryable `ApplicationFailure` of type `upsert_queue_payload_missing`.
3. Other download errors keep the existing retry behavior.

🚗 Drive-by: both activities now go through `getUpsertQueueBucket()` instead of constructing their own `Storage` client from `process.env`, which also removes the now-unused `temporal/config.ts`.

## Tests

- New `lib/upsert_queue.test.ts` covers the three `fetchUpsertQueuePayload` paths (buffer, 404 → `null`, other error rethrown).
- `npm test -- lib/upsert_queue.test.ts temporal/upsert_queue/activities.test.ts`, `tsgo`, `biome check` all pass locally.

## Risk

Worst case, a transient GCS error misreported as a 404 drops one upsert instead of retrying it. All other failures are unchanged. Safe to rollback.

## Deploy Plan

- [ ] Deploy front worker
- [ ] Monitor `upsert_queue_payload_missing.count`; workflows with missing payloads should fail after one attempt
- [ ] Cancel the currently stuck workflows and re-enqueue them from their source system


smb2268:
Opened by mistake, changes live on `32089`.

PR #32180, opened 2026-09-09
[Admin Governance] Preserve revoked editors in the grant backfill

philipperolet:
## Description

Follows https://github.com/dust-tt/dust/pull/31825.

The agent editor backfill currently copies only active editors. If a user is revoked when it runs, their ended editor memberships are missing from the new groups, so rejoining the workspace restores legacy access but not grant-backed access.

Also copy ended, non-suspended memberships from each agent's latest legacy editor group, preserving their timestamps. Create the grant group if all editors are revoked. Copies stay ended; the existing rejoin flow decides which memberships to restore. Dry runs report missing history, and reruns do not duplicate it.

Runtime authorization is unchanged. Backfill-only helpers and database access stay in the migration. Existing Resource APIs still handle group creation and grants, and the migration uses the same lock as live editor changes.

The generic runtime/non-runtime coding-rule clarification is in the independent https://github.com/dust-tt/dust/pull/32181; this PR only changes the backfill and its tests.

## Risks

Blast radius: Agent editor membership history when manually running the backfill.
Risk: standard

The existing last-active-editor cleanup still deletes its grant group and history; changing that lifecycle is a separate follow-up.

## Deploy Plan

- Make the updated front migration available.
- Run `migrations/20260903_backfill_agent_editor_grants.ts` without `--execute`, initially scoped with `--wId`, and inspect active-editor differences and `endedMembershipsToAdd`.
- After approval, rerun with `--execute` in both regions, then dry-run again to verify no remaining differences or missing history.
- No production migration was run as part of this PR.

## Tests

- [x] Regression: revoke → backfill → rejoin, with and without an existing grant group. Revoked memberships stay ended; an editor removed before revocation is not restored.
- [x] Dry-run and rerun idempotence.
- [x] Backfill, group-permission, group, and agent resource suites: 92 tests pass.


philipperolet:
Superseded by https://github.com/dust-tt/dust/pull/32205: a standalone repair targeting legacy editor groups with ended memberships, without rerunning the active-editor backfill. The contract change is handled separately.

PR #31968, opened 2026-09-07
Enable code contracts

spolu:
## Description

Enable code-contracts on our code-base and moves CODING_RULES => CONTRACTS.

Large scale experimentation of code-contracts + this will allow us to create automations against coding rules. Please add yourself as owner 

cc @dust-tt/engineering-team

- No immediate action required, agents will start emitting contract to explicit invariants and assumptions governing the code they write.  Ask is to review these thoroughly.
- Feel free to add yourself as owner on existing coding-related contracts.

**Next steps**
- re-work our agent reviewer around this

## Tests

N/A

## Risk

None, easy to revert with this PR as blueprint.

## Deploy Plan

Merge normally. Contract documentation and local seed tooling changes require no production deployment or database migration.

PR #32181, opened 2026-09-09
[Coding Rules] Distinguish runtime and non-runtime database access

philipperolet:
## Description

Extracted from https://github.com/dust-tt/dust/pull/32180 into an independent, documentation-only PR.

Scope the Resources-only database access rule to runtime application code. Tests, test factories, and one-off scripts/backfills still prefer existing Resource APIs, but may access models directly when those APIs cannot express the needed operation. Keep their dedicated code local instead of adding runtime APIs just for them.

Align the testing rule with this boundary, preserving the preference for Resources and factory-based setup. `CODING_RULES.md` was renamed to `CONTRACTS` in https://github.com/dust-tt/dust/pull/31968; checked the former BACK3, TEST2, and TEST5 rules and their current equivalents for consistency. API and worker access restrictions remain unchanged.

## Risks

Blast radius: Front development and review guidance only.
Risk: none

## Deploy Plan

- Merge normally; no deployment or migration required.

### DIFF
--- a/front/CONTRACTS
+++ b/front/CONTRACTS
@@ -511,7 +511,7 @@ Direct use of Sequelize models in tests should be avoided in favor of Resources.
 setup and assertions.
 
 When existing Resource APIs cannot express a required fixture or assertion, direct model access
-is allowed under [models-behind-resources]; do not add test-only Resource APIs.
+is allowed under the `models-behind-resources` contract; do not add test-only Resource APIs.
 
 @cc [label:react] component-props-interfaces
 Components props should always be typed using an `interface`.
--- a/front/CONTRACTS
+++ b/front/CONTRACTS
@@ -18,10 +18,18 @@ function doWorkspace({ workspace }: { workspace: WorkspaceType }) { }
 ```
 
 @cc [owner:flvndvd,label:backend] models-behind-resources
-Sequelize models must not be used directly outside Resources.
+Runtime application code must access Sequelize models through Resources. This includes API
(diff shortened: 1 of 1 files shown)


front/CONTRACTS
```diff
@@ -502,6 +510,9 @@ possible.
 Direct use of Sequelize models in tests should be avoided in favor of Resources. This includes test
 setup and assertions.
 
+When existing Resource APIs cannot express a required fixture or assertion, direct model access
+is allowed under [models-behind-resources]; do not add test-only Resource APIs.
```
spolu: The [] is not quite valid markdown is it?

spolu (approved):

spolu:
Let maybe try to fix pre existing violations of your new rule `non-runtime-code-stays-local`? 🙏

front/CONTRACTS
philipperolet: likely not indeed; what about renaming to CONTRACTS.md to nudge agentic (& human) coders into validity?

philipperolet:
> Let maybe try to fix pre existing violations of your new rule `non-runtime-code-stays-local`? 🙏

Good point, will send subsequent PRs

front/CONTRACTS
philipperolet: TIL: The brackets are technically valid Markdown, but without a reference-link definition they render as literal brackets—not a link. Backticks make the intent clear.

PR #29485, opened 2026-07-24
[Content Tree] Ignore non-selectable nodes in Select All

philipperolet:
## Description

Related to https://github.com/dust-tt/tasks/issues/9778

`Select All` selected every displayed content node, including nodes explicitly marked as non-selectable. For Microsoft connections this could persist a site container, making the connector keep syncing all libraries under that site even after an individual library was deselected.

Bulk selection now skips non-selectable nodes.

## Risks

Blast radius: Bulk selection in connector content trees.
Risk: low

## Deploy Plan

- Deploy front

## Tests

- [x] `NODE_ENV=test npm test components/ContentNodeTree.test.tsx`

PR #32182, opened 2026-09-09
Fix Microsoft Select All skipping SharePoint libraries

frankaloia:
## Summary
- Eng runner ticket: https://github.com/dust-tt/tasks/issues/10351
- Select All on Microsoft connections was a no-op because the root list is SharePoint sites marked `preventSelection`, after [#29485](https://github.com/dust-tt/dust/pull/29485).
- Select All now loads children of those non-selectable site containers and selects the libraries/folders underneath, without persisting the site itself.
- Unselect All clears those selected descendants as well.

## Testing
Validated locally

## Risk
This can be slow because we are now traversing the tree of nested subsites. This will only effect connectors with top level nodes taht are not selectable

## Deploy Plan
Deploy front

r? @philipperolet Addressing eng runner ticket

### DIFF
--- a/front/components/ContentNodeTree.tsx
+++ b/front/components/ContentNodeTree.tsx
@@ -1,3 +1,7 @@
+import {
+  collectSelectableNodesForSelectAll,
+  unselectVisibleNodesAndDescendants,
+} from "@app/components/ContentNodeTreeSelection";
 import { InfiniteScroll } from "@app/components/InfiniteScroll";
 import { useSendNotification } from "@app/hooks/useNotification";
 import { getVisualForContentNode } from "@app/lib/content_nodes";
@@ -79,6 +83,7 @@ type ContextType = {
   emptyComponent: ReactNode;
   defaultExpandedIds?: string[];
   getLabel?: (node: ContentNode) => string;
+  fetchChildResources?: (parentId: string) => Promise<ContentNode[]>;
 };
 
 const ContentNodeTreeContext = React.createContext<ContextType | undefined>(
@@ -184,6 +189,7 @@ function ContentNodeTreeChildren({
     emptyComponent,
     defaultExpandedIds,
     getLabel,
+    fetchChildResources,
   } = useContentNodeTreeContext();
 
   const sendNotification = useSendNotification();
@@ -192,6 +198,7 @@ function ContentNodeTreeChildren({
   // If the user pressed "select all", we want to display "unselect all" and vice versa.
   // But if the user types in the search bar, we want to reset the button to "select all".
   const [selectAllClicked, setSelectAllClicked] = useState(false);
+  const [isSelectAllLoading, setIsSelectAllLoading] = useState(false);
 
   const {
     resources,
@@ -353,6 +360,48 @@ function ContentNodeTreeChildren({
     </Tree>
   );
 
+  const handleSelectAllClick = async () => {
+    if (!setSelectedNodes) {
+      return;
+    }
+    if (selectAllClicked) {
+      setSelectAllClicked(false);
+      setSelectedNodes((prev) =>
+        unselectVisibleNodesAndDescendants(prev, filteredNodes)
+      );
+      return;
+    }
+
+    setIsSelectAllLoading(true);
+    try {
+      const nodesToSelect = await collectSelectableNodesForSelectAll({
+        nodes: filteredNodes,
+        parentIds,
+        fetchChildResources,
+      });
+      setSelectAllClicked(true);
+      setSelectedNodes((prev) => {
+        const newState = { ...prev };
+        for (const { node, parents } of nodesToSelect) {
+          newState[node.internalId] = {
+            isSelected: true,
+            node,
+            parents,
+          };
+        }
+        return newState;
+      });
+    } catch {
+      sendNotification({
+        type: "error",
+        title: "Failed to select all",
+        description: "Could not load folders to select. Please try again.",
+      });
+    } finally {
+      setIsSelectAllLoading(false);
+    }
+  };
+
   return (
     <>
       {isTitleFilterEnabled && setSelectedNodes && (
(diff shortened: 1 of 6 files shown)


front/lib/swr/connectors.ts
```diff
@@ -51,36 +83,71 @@ export function useConnectorPermissions<T extends ConnectorPermission | null>({
...
+    parentId,
+    filterPermission,
+    viewType,
+  });
 
   const { data, error } = useSWRWithDefaults(url, permissionsFetcher, {
     disabled,
   });
 
+  const includePrivateSlackChannels = featureFlags.includes(
```
philipperolet, line 98: nit: not sure side this refactoring adds value, it adds lines without increasing clarity as the initial featureFlag.includes was already clear (it's a nit though)

front/components/ContentNodeTreeSelection.ts
philipperolet, line 1: nit: this seems like a lib file; I thought it was a component because of the capitalization, i'd uncapitalize

front/components/ContentNodeTreeSelection.ts
```diff
@@ -0,0 +1,115 @@
...
+export function unselectVisibleNodesAndDescendants(
+  prev: Record<string, SelectionStatus>,
+  filteredNodes: ContentNode[]
+): Record<string, SelectionStatus> {
+  const filteredIds = new Set(filteredNodes.map((n) => n.internalId));
+  const newState: Record<string, SelectionStatus> = { ...prev };
+
+  for (const [id, status] of Object.entries(prev)) {
+    const isVisible = filteredIds.has(id);
+    const isDescendant = (status.parents ?? []).some((parentId) =>
```
philipperolet, line 94: IUUC the unselect relies on parent list to the site
so it can leave previously saved folders selected. nodes loaded when opening the modal have either an empty parents list or only their immediate parent (see getNodeParents in ConnectorPermissionsModal and getMicrosoftNodeAsContentNode). 

For a previously selected folder inside a library, Select All adds the library but leaves that folder entry untouched, since collection stops at selectable nodes. Unselect All then cannot match the folder to a visible site, so the folder stays selected and Save continues sending it

front/components/ContentNodeTreeSelection.ts
```diff
@@ -0,0 +1,115 @@
...
+    return selected;
+  }
+
+  const nested = await concurrentExecutor(
+    containersToExpand,
+    async (node) => {
+      const nextVisited = new Set(visitedInternalIds);
+      nextVisited.add(node.internalId);
+      const children = await fetchChildren(node.internalId);
+      return collectSelectableNodesForSelectAll({
```
philipperolet, line 60: The concurrency limit should apply across the traversal. Each recursive call starts another executor with concurrency 8. Eight sites with eight subsites can therefore schedule 64 child fetches concurrently, with further multiplication at deeper level; could trigger Microsoft Graph throttling, and a rejected fetch aborts the entire bulk selection.  What about using one shared queue or concurrency limit for all child fetches in the traversal?

front/components/ContentNodeTree.tsx
```diff
@@ -353,6 +360,48 @@ function ContentNodeTreeChildren({
...
+    }
+
+    setIsSelectAllLoading(true);
+    try {
+      const nodesToSelect = await collectSelectableNodesForSelectAll({
+        nodes: filteredNodes,
+        parentIds,
+        fetchChildResources,
+      });
+      setSelectAllClicked(true);
```
philipperolet, line 382: [P2] Keep a stale completion from overriding the search reset

The search input remains enabled while Select All loads. If I start Select All with filter A and change the filter to B before it completes, onChange resets selectAllClicked, but this completion sets it back to true and selects the libraries under A. The button now says Unselect All for B; clicking it clears B selections and leaves A selected. Please disable search during loading or invalidate the pending operation when the filter changes.

front/components/ContentNodeTree.tsx
```diff
@@ -353,6 +360,48 @@ function ContentNodeTreeChildren({
...
+    }
+    if (selectAllClicked) {
+      setSelectAllClicked(false);
+      setSelectedNodes((prev) =>
+        unselectVisibleNodesAndDescendants(prev, filteredNodes)
+      );
+      return;
+    }
+
+    setIsSelectAllLoading(true);
```
philipperolet, line 375: [P2] Disable Save while bulk selection is pending

This loading state stays inside the tree, so ConnectorPermissionsModal does not include it in the Save button disabled condition. After making a manual selection change, I can click Select All and then Save while children are still loading. Save persists only the earlier selection; the collected libraries arrive afterward and are never saved. Please expose the pending state to the modal and prevent saving until the operation completes.

front/lib/swr/connectors.ts
```diff
@@ -51,36 +83,71 @@ export function useConnectorPermissions<T extends ConnectorPermission | null>({
...
+}): (parentId: string) => Promise<ContentNode[]> {
+  const { fetcher } = useFetcher();
+  const { featureFlags } = useFeatureFlags();
+  const includePrivateSlackChannels = featureFlags.includes(
+    "index_private_slack_channel"
+  );
+
+  return useCallback(
+    async (parentId: string) => {
+      const data: GetDataSourcePermissionsResponseBody = await fetcher(
```
philipperolet, line 135: [P3] Follow the SWR fetching contract

The network-operations-in-swr-hooks contract in front/CONTRACTS requires data fetching through SWR hooks. This callback calls fetcher directly, bypassing the cache, deduplication, and retry handling used by useConnectorPermissions for the same URL. Please use the SWR fetching path for this on-demand load as well.

front/components/ContentNodeTree.tsx
```diff
@@ -353,6 +360,48 @@ function ContentNodeTreeChildren({
...
+        for (const { node, parents } of nodesToSelect) {
+          newState[node.internalId] = {
+            isSelected: true,
+            node,
+            parents,
+          };
+        }
+        return newState;
+      });
+    } catch {
```
philipperolet, line 394: [P3] Handle expected load failures through the repository error contract

This catch handles exceptions propagated through collectSelectableNodesForSelectAll and our fetcher, with no documented local exception to the root no-catching-own-errors contract. That contract requires expected failures from repository functions to be returned as Result/Err values, while unexpected exceptions propagate. Please handle the expected load failure through that mechanism and narrow the error handling so selection/state errors are not also reported as folder-loading failures.

front/components/ContentNodeTreeSelection.ts
```diff
@@ -0,0 +1,115 @@
...
+  return [...selected, ...nested.flat()];
+}
+
+/**
+ * @cc [owner:frankaloia,label:product] unselect-descendants
+ * Unselect All at a tree level MUST unselect the currently visible nodes and any
+ * already selected descendants of those nodes, including previously persisted
+ * selections on non-selectable containers.
+ */
+type SelectionStatus = {
```
philipperolet, line 79: [P3] Attach unselect-descendants to the function it governs

This contract comment currently attaches to the SelectionStatus type, so contract discovery for unselectVisibleNodesAndDescendants will not include it. Please move the comment immediately above the function declaration.

philipperolet (commented):
Thanks 🙏 Left a few coms, most importantly the unselect one
+ a few of them from an AI ([Pnumber]) that seemed legit

philipperolet (approved):
Oops sorry for the delay again 🙇 
Thanks for the fix 🙏

PR #32184, opened 2026-09-09
[front] Slash model picker: /model <query> shortcut with effort selection

PopDaph:
## Description

Pick a model from the composer's `/` menu in one line of typing.

- `/mo ` (any prefix of a command word, or the full `/pick model `) opens the model list. `/att <search>` does the same for Attach.
- Type the model name loosely, in order, spaces and hyphens optional: `fab`, `gpt6`, `glm3`, `claudehaiku`. Best match first.
- End with `l`, `m` or `h` to pick the effort: `/mo fa h` is Claude Fable 5 High. Without it, rows stay light, medium, high and the model's default effort is highlighted, like the model picker.
- Enter or Tab selects, Escape goes back to the `/` list, Backspace edits the text.

Why it is built this way:

- The space acts as Enter once the command word matches a label start. Matching labels only avoids opening sub-menus on words like `/or `. The sub-menu is derived from the text, so Backspace and Escape need no extra state.
- Model names use the agent search's `subFilter` and `compareForFuzzySort` instead of a new fuzzy scorer. Efforts are a closed set of three words, so the trailing letter is matched exactly while the name stays loose.

The skill builder's slash menu is unchanged.

## Tests

Unit tests for sub-menu resolution, name matching and ranking, effort selection, tier matching and the highlighted row. Manually checked `/mo fa h`, `/mo gpt6 h`, `/pick model gpt6 h`, `/mo op`, `/att report`, Escape and Backspace, and that `/or x` stays on the top-level list.

## Risk

Frontend only. Model search moves from substring to subsequence with ranking, so short queries show more rows with the best match on top. Safe to rollback.

## Deploy Plan

Deploy front.


front/components/editor/extensions/shared/slash_suggestion/buildPickModelSlashCommandItems.ts
```diff
@@ -47,14 +49,79 @@ export function getSelectableEffortsForSlashMenu(
...
+  // Hyphens split like spaces so a displayed name such as "GPT-5.4 Mini" can be typed as is.
+  const queryWords = query
+    .toLowerCase()
+    .split(/[\s-]+/)
+    .filter((word) => word.length > 0);
+  const lastWord = queryWords.at(-1);
+  const effort =
+    lastWord === undefined
+      ? undefined
+      : SLIDER_EFFORTS.find((candidate) => candidate.startsWith(lastWord));
```
aubin-tchoi, line 95: nit: if a model has either "light", "medium" or "high" in its name this could mess up the logic (I think there's Mistral medium). not a big deal atm but if such a model is released we could struggle finding it

aubin-tchoi (approved):
LGTM

front/components/editor/extensions/shared/slash_suggestion/buildPickModelSlashCommandItems.ts
PopDaph: ooooh nasty cat 😅

PopDaph: Added a commit, not proud of how to handle but I think okay enough. 
Thanks for catching!

PR #32187, opened 2026-09-09
[SANDBOX] Compute step context for sandbox child actions

smb2268:
## Description

Fixes the recurring `web_search_browse` Firecrawl 400 errors (`path: ["limit"], minimum: 0, inclusive: false`) from sandbox child actions: 18 failures on a single agent message in `cell-00001` (Europe West) on 2026-09-14, same signature first seen on 2026-09-09.

Sandbox child actions (an LLM inside the `sandbox` bash tool calling another tool through `dsbx`) copied the parent bash action's step context. That context was computed for a step containing only the sandbox tool, so `websearchResultCount` (and `retrievalTopK`, `citationsCount`) were 0. A child `websearch` call then forwarded `num: 0` to Firecrawl, which rejects `limit: 0`; a child `search` call likewise ran with a zero topK.

The child's step context is now computed with `computeStepContexts` from its own tool configuration, as a single-action step starting at the parent's citation offset. The model is resolved through `resolveAgentMessageModelConfig`, extracted from the agent loop so both paths share the stored-resolution, legacy-stream and noop-pin logic. Direct agent-loop calls and the `/run` command path are unchanged. As a safety net, the websearch handler falls back to its agent-less default with a warning when a step context carries a non-positive result count.

The `AgentMCPServerConfigurationFactory` wrote a placeholder `internalMCPServerId` where production copies the view's id, which made every internal tool look external in tests and hid this. It now mirrors production.

## Tests

- Added a regression test in `create_child_action.test.ts`: a websearch child of a sandbox action gets the allocation of a websearch-only step (16 results, 16 citations at the parent's non-zero offset) rather than the parent's counts. It fails without the fix.
- Added a websearch handler test asserting the step-context count is forwarded to the provider and that a zero count falls back to the agent-less default.
- Ran the child-action test file, the agent loop context provider test, `models.test.ts`, and the five test files using `AgentMCPServerConfigurationFactory` (all passing).

## Risk

Low. Only affects step contexts of sandbox child actions; the loop's model resolution is moved, not changed. Sandbox children are hidden from the conversation timeline and excluded from citation offset sums, so the new non-zero `citationsCount` on children does not shift later steps (their references may overlap later steps', but child outputs feed the sandbox LLM rather than conversation citations). Safe to roll back.

## Deploy Plan

Deploy front. No migration.


smb2268:
PMRR - seeing more flurries of this error so want to kill it @Fraggle

Fraggle (commented):
Lgtm

PR #32188, opened 2026-09-10
Allow conversation sandboxes to query Frame databases

davidebbo:
## Description

- Add `--frame <frame-id>` to `dsbx db list` and `dsbx db query` in conversation sandboxes.
- Forward those commands through Front to the target Frame v2 sandbox. Frame sandboxes keep local database behavior without `--frame`.
- Require a conversation action token and write access to the Frame source before Front starts or queries the target sandbox.
- Support SELECT and data-changing SQL while preserving the existing single-statement and no-DDL checks.
- Return bounded, ordered previews for large remote result sets instead of inaccessible Frame-local spill files.
- Bump dsbx to 0.1.58 and dust-base to 0.8.108.

## Tests

- `cargo test` in `cli/dust-sandbox`: 413 unit tests and 1 e2e test
- `bun test functions-runner/db/query.test.ts`: 22 tests
- Focused `front` tests: 31 tests
- Focused `front-api` Frame route tests: 21 tests
- `npx tsgo --noEmit` in `front` and `front-api`
- Biome, Rust formatting, Swagger annotation, code-contract, and diff checks

## Risk

The endpoint can mutate Frame databases. Front requires write access to the Frame source, rejects function invocation tokens, and resolves the Frame within the token workspace before starting its sandbox. Cross-workspace requests return 404, and same-workspace callers without write access receive 403.

Large results stay within the existing inline limits. Reverting this change removes remote access, but it cannot undo database writes that already ran.

## Deploy Plan

1. Deploy Front and Front API with the internal Frame database endpoint.
2. Run the manual `Release dsbx CLI` workflow to publish `dsbx-v0.1.58`.
3. Build and release `dust-base:0.8.108`.
4. Replace sandboxes running older dust-base versions so both conversation and Frame sandboxes receive the new CLI.

### DIFF
--- a/cli/dust-sandbox/Cargo.lock
+++ b/cli/dust-sandbox/Cargo.lock
@@ -313,7 +313,7 @@ checksum = "92773504d58c093f6de2459af4af33faa518c13451eb8f2b5698ed3d36e7c813"
 
 [[package]]
 name = "dust-sandbox"
-version = "0.1.57"
+version = "0.1.58"
 dependencies = [
  "anyhow",
  "base64",
--- a/cli/dust-sandbox/Cargo.toml
+++ b/cli/dust-sandbox/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "dust-sandbox"
-version = "0.1.57"
+version = "0.1.58"
 edition = "2021"
 
 [[bin]]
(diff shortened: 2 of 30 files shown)


davidebbo:
Replaced by a stack.

PR #32189, opened 2026-09-10
Report peak concurrent database queries for Hono requests

flvndvd:
## Description

Peak concurrent database query tracking was lost when API traffic moved from Next.js to Hono. Hono requests now create their own query tracker context and include peakConcurrentQueries in the standard processed-request log, restoring the signal used to spot handlers that put pressure on the connection pool.

## Tests

The request instrumentation test runs two database queries concurrently and checks that the processed-request log reports a peak of 2. The focused suite passes with all three tests. The full workspace type check was not clean because the current worktree contains unrelated type errors in existing and untracked files.

## Risk

Low. The extra AsyncLocalStorage context is request-local and only affects query counting and one log field. Rolling back the PR removes the field and returns to the current behavior.

## Deploy Plan

Deploy front-api normally. No migration or ordering requirement.

### DIFF
--- a/front-api/middlewares/request_instrumentation.test.ts
+++ b/front-api/middlewares/request_instrumentation.test.ts
@@ -1,4 +1,5 @@
 import { frontSequelize } from "@app/lib/resources/storage";
+import logger from "@app/logger/logger";
 import {
   getRequestContext,
   RequestCachedQuery,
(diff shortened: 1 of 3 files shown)


ykmsd (approved):

adrsimon (commented):
tyty!

PR #32191, opened 2026-09-10
feat(input-bar): shorten knowledge attach/search labels to Attach and Search

Nils-Fedrigo:
## Description
Screenshot 2026-09-10 at 10 07 17

Shortens the knowledge labels in the composer: the attachment entry point now reads **Attach** instead of "Attach knowledge", and the attachments picker empty state reads **Search** instead of "Search knowledge".

Touched all three composer entry points (`InputBarAttachmentsPicker`, `InputBarPlusMenu`, `InputBarContainer`), the `/attach-knowledge` slash command in the editor, and the Sparkle `Composer` story so the mockup stays in sync with the product.

The slash command's own description line ("Search knowledge and reference conversation or pod files") and its tooltip are unchanged — they are prose, not labels.

## Tests

Copy-only change, verified by inspection. In the composer: the `+` menu / attachment sub-trigger shows "Attach", the attachments picker with an empty search shows "Search", and typing `/` lists "Attach".

## Risk

None — no behavior change.

## Deploy Plan

Standard deploy.

### DIFF
--- a/front/components/assistant/conversation/input_bar/InputBarAttachmentsPicker.tsx
+++ b/front/components/assistant/conversation/input_bar/InputBarAttachmentsPicker.tsx
@@ -577,7 +577,7 @@ export const InputBarAttachmentsPicker = ({
         <DropdownAnchorTrigger anchorRef={anchorRef} />
       ) : (
         <DropdownMenuSubTrigger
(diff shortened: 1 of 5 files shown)


Fraggle (approved):
LGTM

PR #32190, opened 2026-09-10
Add date-fns helpers to cover moment's .fromNow(), .calendar(), and date bucketing

avervaet:
## Description

Decision: https://github.com/dust-tt/decisions/issues/1009 phase out `moment` in favor of `date-fns`. This PR ships step 2 of the phased plan: extends `front/lib/utils/timestamps.ts` with `date-fns`-based helpers covering the patterns `moment` is used for in the codebase, so that migrating a call site becomes a drop-in swap.

- `formatRelativeTime`: replaces `.fromNow()`. Uses date-fns's own `formatDistance` thresholds and wording rather than porting moment's exactly, the exact phrasing isn't load-bearing, and matching it exactly would mean maintaining a parallel implementation of date-fns's own logic.
- `formatCalendarDateTime`: replaces the `.calendar()` sameDay/lastDay/lastWeek pattern. Matches moment's exact output here, including its asymmetry between past (seconds shown) and future (seconds omitted) formats, since this one changes what's rendered in the UI, not just wording.
- `getRelativeDateBucket`: replaces the Today/Yesterday/Last Week/Last Month/Last 12 Months/Older bucketing used for sidebar conversation lists. Boundaries are computed by shifting the reference instant first and taking the start of that day second, so bucketing doesn't drift an hour in timezones whose DST switch happens at midnight (e.g. Chile, Cuba).

## Tests

Adding test for the describe helpers.

## Risk

Low. Additive helpers only, no existing call site is migrated or touched.

## Deploy Plan

None

PR #32192, opened 2026-09-10
Migrate easy moment usages to date-fns

avervaet:
## Description

Point 3 of decisions https://github.com/dust-tt/decisions/issues/1009 (phase out `moment` in favor of `date-fns`).

Swaps the `moment` call sites that don't need timezone-aware handling for the `date-fns` helpers (#32190).

The `date-fns-over-moment` contract is reworded to allow small, reviewable migration batches
like this one, in addition to opportunistic swaps.

**Behavior note:** `formatRelativeTime` (date-fns `formatDistance`) words things differently
from moment's `fromNow()`: "about 1 hour ago" instead of "an hour ago", "26 days ago" instead of
"a month ago", "less than a minute ago" instead of "a few seconds ago". Exact parity would need a
custom locale; the new wording is accepted as-is.

## Tests

Covered by the existing `timestamps.test.ts`. No new tests; every hunk is a one-line swap.

## Risk

Low. Relative-time strings change wording as described above; everything else is a like-for-like
swap behind helpers that already have tests.

## Deploy Plan

Deploy front.


aubin-tchoi (approved):
LGTM
do we have tests that confirm that these do the same?

PR #32195, opened 2026-09-10
Remove pod-owned sandboxes and Pod functions

davidebbo:
Pods no longer own sandboxes: conversation and Frame are the only owner kinds left. Nothing could create a pod-owned sandbox or publish a Pod function any more — `SandboxFunctionResource.makeNew` had no production caller, and every `.invoke()` entry point resolves through `resolveSandboxFunctionWithCapability`, which is Frames-v2-only — but the machinery for both was still wired throughout. This removes it.

## What goes with the pod owner kind

`PodSandboxAdapter`, `ensurePodSandboxReady`, the pod sandbox-only mounts with their GCS prefixes and profiles, `DustFileSystem.forPodSandboxProvisioning`, and the `"pod"` arm of `SandboxRuntimeOwner` — which is what removes the **`SPACE_ID`** env var, along with its reserved-name entry in `SandboxResource` and its read in the Fluent Bit enrichment (now `FRAME_ID`).

Also the reaper's pod branch, and on the Poke side the wake-sandbox plugin, the pod-databases page (route, SWR hook, table) and the sandbox row on the space-details page.

## What goes with Pod functions

The pod branch of `SandboxFunctionInvocationResource.execute` and `SandboxFunctionResource.invoke`, the `"pod"` owner of `authorizeSandboxFunctionInvocation` and of the invocation token, and the fast-function self-heal — a published Frame function's mode is immutable, so the refusal is the whole outcome and only a republish changes how it runs.

Two features whose subject was Pod functions go too: the publish-time validation of Frame-to-Pod-function references (`validate_frame_pod_functions.ts`, a ~560-line TypeScript-compiler pass, plus `pod_function_reference.ts` — a legacy Frame's `callFunction` no longer resolves against a Pod app folder), and the Pod restriction-impact warning, which counted Pod-function invocations and so would have reported zero forever.

That let the rest of `SandboxFunctionResource` collapse to Frames: `makeNew`, `listBySpace`, `deleteAllForSpace`, the `space`/`ownerSpace` getter, and `baseFetch`'s space hydration with its three flags. `delete` now refuses — a Frame function's row belongs to its Frame's publication history, and `FileResource` already cascades the set.

## Kept deliberately

Pod-scoped env vars, HTTPS secrets and egress policy are untouched: they still apply to every conversation and Frame sandbox running *inside* a Pod. `buildPodScopedEnvVars` moved to `SandboxEnvVarResource.loadPodScopedEnv` now that its owner adapter is gone.

## One fix rather than a deletion

Removing `sandboxFunction.space` forced six live call sites off Pod ownership: tool outputs, `generate_file`, upserted files, the `pod_manager` tools, audit metadata and the invocation log context. **Those were already broken for Frame functions on `main`** — the getter asserts on a null owner space, so any Frame function whose tool wrote a file threw. The Pod is now resolved once from the Frame's own scope when the tool run context is built, and a Frame outside any Pod gets a clear refusal instead of an assertion. This is the one behavior change in the PR and the part most worth a close read.

## Production cleanup

`front/scripts/cleanup_pod_functions_and_sandboxes.ts` scrubs what all of this leaves behind, per workspace: pod-owned sandboxes (destroyed at the provider, then `sandbox_owners` and `sandboxes` rows dropped), pod functions with their MCP actions, invocations and published bundle files, and the two pod GCS prefixes (litestream state replica, read-only bundle mount). Idempotent, resumable via `--wId` / `--fromWorkspaceModelId`, dry-run by default.

**It has to run once after this deploys.** Until it does, a pod that booted a sandbox keeps it alive at the provider forever: the reaper no longer has an owner adapter for it and skips it. `deleteLegacyPodOwnedSandbox` is the only path left that can reach one, and pod deletion also calls it.

The two `spaceId` columns themselves are untouched — `sandbox_functions.spaceId` is now only ever written as `null` and read nowhere, and `sandbox_owners.spaceId` is read only by the space-delete cleanup and that helper. Dropping them needs a migration, and that should follow the cleanup run.

## Testing

`front` 236 files / 3033 tests and `front-api` 292 files / 2177 tests pass; `tsgo --noEmit` clean in `front`, `front-api` and `front-spa`. Net **101 files, +928 / −6837**.


front/lib/api/sandbox/legacy_pod_sandbox.ts
```diff
@@ -0,0 +1,52 @@
...
+ * holds the provider sandbox and its `sandbox_owners` row. Destroying the provider sandbox is what
+ * stops it being stranded: with no owner adapter left, the reaper skips it forever.
+ *
+ * `dropOwnerLink` says who removes the `sandbox_owners` row. Pod deletion leaves it to
+ * `SpaceResource.delete`, which has to drop it anyway (the FK is `onDelete: "RESTRICT"`); the
+ * backfill script keeps the pod, so it drops the row itself.
+ *
+ * Delete this module once production holds no `sandbox_owners` row with a `spaceId`.
+ */
+export async function deleteLegacyPodOwnedSandbox(
```
davidebbo, line 20: This can be deleted later

davidebbo:
Superseded by a stack of six smaller PRs, each independently reviewable and green: #32213 → #32214 → #32215 → #32216 → #32217 → #32218. Same end state, except that `getDatabaseSchemaOnSandbox` and `queryDatabaseOnSandbox` are kept and moved onto the ready-sandbox shape in #32213 rather than deleted, and the backfill in #32218 deletes rows only and never GCS objects.

PR #32196, opened 2026-09-10
The deployment card

flvndvd:
## Description

Today a deploy opens the channel with a four bullet header, a build notice, and lines from dust-infra as cells complete. The only person told anything is the one who clicked deploy. The engineers whose commits are going live find out when someone asks.

After this PR a deploy opens with one card, and the authors of the commits shipping are mentioned on it, once, when it starts:

```
🚀 front is going live on all cells
3 commits by @arthur, @jd and @flavien

⏳ cell-00000  waiting
⏳ cell-00001  waiting
⏳ cell-00002  waiting

started by @arthur at 09:25 · ccd48c3 · run
```

The commits are the difference between what each target cell runs today, read from the versions files dust-infra keeps on main, and the commit being deployed, the widest range across the cells. Authors resolve to Slack through the `.authors` file, unknown ones keep their name, and messages are escaped so a commit subject cannot break the card. The list itself goes into the thread as "What is shipping", newest first, twenty at most with a count of the rest. The cells are resolved the way dust-infra resolves them, the legacy clusters from the Novu cell list and the descriptor cells that list the component, so the card and the deploy agree on who is in. dust-infra (dust-tt/dust-infra#1084) then moves the cell lines and the header, the card facts travel in the dispatch payload.

Revert opens the same card with a rewind header and lists what is being rolled back. The two build notices are gone, the card and its thread carry what they said, and the build step's commit diff outputs are left in place for a later cleanup.

## Tests

The script passes a syntax check and actionlint is clean on both workflows. The descriptor parsing was run against the real cell-00002 descriptor and returns its twelve applications and the egress proxy flag. Two independent reviews found and fixed: the octokit for dust-infra was built through a method that does not exist in github-script, a revert compared the wrong pair of commits and would have mentioned the wrong people, the compare API caps at a hundred commits per page, and a Slack failure let the changelog land at top level. The Slack and GitHub calls themselves need a real run, the first deploy after both PRs merge is the proof.

## Risk

Slack text, and a read of dust-infra through the App token the deploy already uses. If the card cannot be posted the job fails before building, as it does today when the start notice fails.

## Deploy Plan

Merge this first, then dust-tt/dust-infra#1084. In between, dust-infra ignores the card and keeps posting its own cell lines in the thread.

### DIFF
--- a/.github/actions/deploy-card/action.yml
+++ b/.github/actions/deploy-card/action.yml
@@ -22,7 +22,7 @@ inputs:
     description: "Slack bot token"
     required: true
   infra_token:
-    description: "Token that reads dust-infra, for the tags currently deployed"
+    description: "Token that reads dust-infra, for the cells and the tags currently deployed"
     required: true
 
 outputs:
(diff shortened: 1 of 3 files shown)


Fraggle (approved):
LGTM

PR #32197, opened 2026-09-10
[LLM] Claude Opus 5 EU agent-platform endpoint

pmilliotte:
## Description

Adds the Claude Opus 5 EU agent-platform endpoint (Vertex `eu`), and fixes a bug the endpoint surfaced: the Anthropic stream converter emitted the stop-reason error and then a `success`, so consumers — which stop at the first terminal event — dropped the `token_usage` that followed and double-counted the call as both a failure and a success.

## Tests

Added a converter regression test (a refused turn ends on the error, after usage, with no `success`) and pointed the Opus 5 EU cache case at a refusal error; updated the Opus 4.8 EU expectations for temperature 1 and forced tool use. Live endpoint suite (`RUN_LLM_TEST=true`) not run.

## Risk

Refusal and `max_tokens` turns on Anthropic now record the token usage that was previously dropped, so expect a small uptick in recorded usage and a drop in `llm_success.count` for turns that were being counted twice.

## Deploy Plan

- Deploy `main` on `front`

### DIFF
--- a/front/CONTRACTS
+++ b/front/CONTRACTS
@@ -448,15 +448,6 @@ schema file, re-register with WorkOS after merge and before deploy:
 `npx tsx front/admin/register_audit_log_schemas.ts --execute --changed`. The CI test enforces
 that the `AuditAction` union and schema files stay consistent.
 
-@cc [label:error-handling] single-terminal-model-response-event
-A streamed `ModelResponseEvent` sequence must end with exactly one terminal event — either
-`success` or `error`, never both — and every `token_usage` event must be emitted before it.
-
-Consumers (`LLM.streamWithTracing`, `completeStream`) stop at the first terminal event: anything
-yielded after it is dropped, and a `success` following an `error` double-counts the call as both a
-failure and a success in telemetry. When a provider stop/finish reason maps to an error, hold the
-error back until the usage has been yielded, then yield the error and end the stream.
-
 @cc [label:mcp] internal-mcp-server-structure
 Internal MCP servers must live in `lib/api/actions/servers/<server_name>/` and separate metadata,
 tool implementation, and server wiring:
--- a/front/CONTRACTS
+++ b/front/CONTRACTS
@@ -448,6 +448,15 @@ schema file, re-register with WorkOS after merge and before deploy:
 `npx tsx front/admin/register_audit_log_schemas.ts --execute --changed`. The CI test enforces
 that the `AuditAction` union and schema files stay consistent.
 
+@cc [label:error-handling] single-terminal-model-response-event
+A streamed `ModelResponseEvent` sequence must end with exactly one terminal event — either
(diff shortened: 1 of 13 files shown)


front/lib/llms/stream/endpoints/anthropic_claude_opus_five_eu_agent_platform.ts
```diff
@@ -0,0 +1,18 @@
...
+
+export class DustAnthropicClaudeOpusFiveEuropeAgentPlatformStream extends WithDustClaudeOpusFiveConfig(
+  AnthropicClaudeOpusFiveEuropeAgentPlatformStream
+) {
+  static readonly endpointFilter = {
+    or: [
+      {
+        featureFlags: { contains: "use_vertex_for_supported_models" as const },
+      },
+      { isCreditPriced: { eq: true } },
```
Nils-Fedrigo, line 13: did not check if we have the same guard but I expect us to?

front/lib/model_constructors/sdk/anthropic_ai/converters/output/utils.test.ts
```diff
@@ -1239,6 +1239,38 @@ describe("rawOutputToEvents", () => {
     });
   });
 
+  it("ends a refused turn on the error event, after usage and with no success", async () => {
```
Nils-Fedrigo, line 1242: and in practise, it's supposed to be an error right? Faulty behaviour?

front/CONTRACTS
```diff
@@ -448,6 +448,15 @@ schema file, re-register with WorkOS after merge and before deploy:
 `npx tsx front/admin/register_audit_log_schemas.ts --execute --changed`. The CI test enforces
 that the `AuditAction` union and schema files stay consistent.
 
+@cc [label:error-handling] single-terminal-model-response-event
```
Nils-Fedrigo: put it closer to the function instead of the general file?

Nils-Fedrigo (approved):
ty!

front/lib/model_constructors/sdk/anthropic_ai/converters/output/utils.test.ts
pmilliotte: Yes, user sees a message like "Anthropic refused to complete your request. Please rephrase your message and try again."

Nils-Fedrigo: Thanks!

front/lib/llms/stream/endpoints/anthropic_claude_opus_five_eu_agent_platform.ts
pmilliotte: confirmed

PR #32198, opened 2026-09-10
[MODELS] Default internal Gemini Flash usage to 3.8

pmilliotte:
## Description

Repoints internal Gemini Flash defaults (conversation titles, the Dust global agent, small/fast model selection) to 3.8, marks 3.5 and 3.6 legacy and drops them from the model picker, and documents the rule of keeping at most two versions per model family in `USED_MODEL_CONFIGS`.

## Tests

Not needed — model configuration and documentation only.

## Risk

Low.

## Deploy Plan

- Deploy `main` on `front`


Nils-Fedrigo (approved):
thanks

PR #32199, opened 2026-09-10
[BYOK] Exclude agent-platform endpoints from BYOK workspaces

pmilliotte:
## Description

Excludes `agent-platform` endpoints from BYOK workspaces in `getWorkspaceFilter`: those endpoints run on Dust's own Vertex project (`AGENT_PLATFORM_PROJECT_ID`) and ignore the credentials the workspace provided, and until now nothing in the router prevented a BYOK workspace from reaching one — it was kept out only by each endpoint's `endpointFilter` (`use_vertex_for_supported_models` flag **or** `isCreditPriced`), neither of which mentions BYOK.

## Tests

Added a test that grants a BYOK workspace both the vertex feature flag and credit pricing and asserts agent-platform stays excluded for Gemini and Claude; verified it fails without the fix.

## Risk

BYOK workspaces now route Anthropic and Gemini to the labs' own APIs rather than Vertex; every affected model keeps a lab-hosted endpoint, so nothing becomes unavailable.

## Deploy Plan

- Deploy `main` on `front`

### DIFF
--- a/front/CONTRACTS
+++ b/front/CONTRACTS
@@ -448,6 +448,16 @@ schema file, re-register with WorkOS after merge and before deploy:
 `npx tsx front/admin/register_audit_log_schemas.ts --execute --changed`. The CI test enforces
 that the `AuditAction` union and schema files stay consistent.
 
+@cc [label:security;product] byok-never-routes-to-dust-hosted-inference
+A workspace on a BYOK plan (`plan.isByok`) must only reach endpoints served by the model lab's own
+API, using the credentials the workspace provided. Endpoints hosted on Dust's infrastructure —
+`agent-platform` (Vertex, keyed by `AGENT_PLATFORM_PROJECT_ID`) today — must be filtered out in
+`getWorkspaceFilter`, not merely made unreachable by an endpoint's `endpointFilter`: plan and
+feature-flag conditions can change, the BYOK guarantee cannot.
+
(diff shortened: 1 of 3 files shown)


front/CONTRACTS
```diff
@@ -448,6 +448,16 @@ schema file, re-register with WorkOS after merge and before deploy:
 `npx tsx front/admin/register_audit_log_schemas.ts --execute --changed`. The CI test enforces
 that the `AuditAction` union and schema files stay consistent.
 
+@cc [label:security;product] byok-never-routes-to-dust-hosted-inference
```
Nils-Fedrigo: prehaps put it closer to code?

Nils-Fedrigo (approved):
agree with the code but lack product rationale 

Why not allow BYOK on vertex AI instead of blocking it? 

Seems like we would need to do a 2 step process: if we have a BYOK, we take the API that the BYOK is using, then if not and credit pricing we take the region one etc

But you probably have more context and we probably don't have appetite on bringing BYOK to vertex 🙏

pmilliotte:
Yes does not change the behavior but its an added security check. The problem is that we authenticate through our service account for vertex, not an api key. If a byok workspace went to be hosted in eu (which is not the case rn), then anthropic eu models (through vertex) would take precedence over us ones (through anthropic host), and we would pay the inference cost.

This is a small protection for now which would be revised if we have additional appetite for BYOK

PR #32200, opened 2026-09-10
The card mentions people again

flvndvd:
## Description

The card went out this morning with names instead of mentions: "3 commits by Flavien David and Thomas Draier", "started by flvndvd". The point of the card is that the people whose work ships get pinged, so this is the one thing it had to do right.

The cause is a Slack API detail. Its read methods, `users.lookupByEmail` here, refuse a JSON body and want a query string, while the write methods take JSON. The card's helper sent everything as JSON, every lookup failed silently, and every author fell back to a name. Lookups now go as GET with query parameters, the card mentions the authors and the person deploying, and the "What is shipping" list in the thread names each commit's author the same way.

## Tests

The script passes a syntax check and the query string encoding was checked on an address with a plus sign. The next deploy is the proof: the authors line reads `by @flavien and @thomas`.

## Risk

Slack text only.

## Deploy Plan

Merge, deploy anything.

### DIFF
--- a/.github/actions/deploy-card/action.yml
+++ b/.github/actions/deploy-card/action.yml
@@ -56,11 +56,15 @@ runs:
           const tag = process.env.IMAGE_TAG;
           const revert = process.env.IS_REVERT === 'true';
           const esc = (s) => String(s).replace(/&/g, '&amp;').replace(/</g, '&lt;').replace(/>/g, '&gt;');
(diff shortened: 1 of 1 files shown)


aubin-tchoi (approved):
LGTM

adrsimon (commented):
LGTM

PR #32201, opened 2026-09-10
Wake-ups: add a tab in automations to list your wakeups

fabiencelier:
## Description

closes https://github.com/dust-tt/tasks/issues/9378

Add a new tab in the personal "automations" screen to list wakeups.
This uses a new endpoint to list the wakeups, with pagination.
Wakeups are filtered by "scheduled" as other status means the wakeup already fired or has been canceled.
Not adding the option to cancel the trigger, this can be done by going to the conversation page and click the obvious cancel button (this can be reconsidered later if really needed)

image

## Tests

- unit tests for endpoint and resource method
- manually tested the feature
- manually tested the pagination 

https://github.com/user-attachments/assets/9e69470f-582a-458d-b5ee-07179408f66c

## Risk

low, people can only see their wakeups

## Deploy Plan

deploy front

### DIFF
--- a/front-api/routes/w/[wId]/me/wakeups.ts
+++ b/front-api/routes/w/[wId]/me/wakeups.ts
@@ -1,0 +1,34 @@
+import type { UserWakeUps } from "@app/lib/api/assistant/wakeups";
+import { listUserWakeUps } from "@app/lib/api/assistant/wakeups";
+import { WakeUpStatusSchema } from "@app/types/assistant/wakeups";
+import { workspaceApp } from "@front-api/middlewares/ctx";
+import type { HandlerResult } from "@front-api/middlewares/utils";
+import { validate } from "@front-api/middlewares/validator";
+import { z } from "zod";
+
+const QuerySchema = z.object({
+  limit: z.coerce.number().int().min(1).max(100).optional().default(25),
+  offset: z.coerce.number().int().min(0).optional().default(0),
+  // Only the still-pending wake-ups are of interest by default; terminal ones are history.
+  status: WakeUpStatusSchema.optional().default("scheduled"),
+});
+
+export type GetUserWakeUpsResponseBody = UserWakeUps;
+
+// Mounted at /api/w/:wId/me/wakeups.
+const app = workspaceApp();
+
+/** @ignoreswagger */
+// Internal endpoint backing the "Wake-Ups" tab of the personal automations view.
+app.get(
+  "/",
+  validate("query", QuerySchema),
+  async (ctx): HandlerResult<GetUserWakeUpsResponseBody> => {
+    const auth = ctx.get("auth");
+    const { limit, offset, status } = ctx.req.valid("query");
+
+    return ctx.json(await listUserWakeUps(auth, { limit, offset, status }));
+  }
+);
+
+export default app;
--- a/front-api/routes/w/[wId]/me/wakeups.ts
+++ b/front-api/routes/w/[wId]/me/wakeups.ts
@@ -27,7 +27,9 @@ app.get(
     const auth = ctx.get("auth");
     const { limit, offset, status } = ctx.req.valid("query");
 
-    return ctx.json(await listUserWakeUps(auth, { limit, offset, status }));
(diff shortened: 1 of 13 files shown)


front-api/routes/w/[wId]/me/wakeups.ts
```diff
@@ -0,0 +1,34 @@
...
+/** @ignoreswagger */
+// Internal endpoint backing the "Wake-Ups" tab of the personal automations view.
+app.get(
+  "/",
+  validate("query", QuerySchema),
+  async (ctx): HandlerResult<GetUserWakeUpsResponseBody> => {
+    const auth = ctx.get("auth");
+    const { limit, offset, status } = ctx.req.valid("query");
+
+    return ctx.json(await listUserWakeUps(auth, { limit, offset, status }));
```
Comment (adrsimon) [inline front-api/routes/w/[wId]/me/wakeups.ts]: nit: i don't think this is a common pattern on the codebase, we usually 

const result = await func()  
return ctx.json(result) no ?

front/lib/api/assistant/wakeups.ts
```diff
@@ -0,0 +1,79 @@
...
+/**
+ * @cc [owner:fabiencelier,label:product;security] only-caller-wake-ups
+ * The returned wake-ups MUST all be owned by `auth`'s user and in `status`. A wake-up owned by
+ * another member of the workspace MUST never be returned, whatever the caller's role.
+ */
+/**
+ * @cc [owner:fabiencelier,label:product] drop-unreachable-conversations
+ * A wake-up whose conversation the caller cannot read (deleted conversation, revoked space access)
+ * MUST be omitted from `wakeUps` rather than returned without its conversation. `totalCount` still
+ * counts it, so it can exceed the number of returned rows.
```
adrsimon: do we really want totalCount to still count this ?

front/lib/resources/wakeup_resource.test.ts
```diff
@@ -0,0 +1,159 @@
...
+    await AgentConfigurationFactory.createTestAgent(authenticator);
+  const conversation = await ConversationFactory.create(authenticator, {
+    agentConfigurationId: agentConfiguration.sId,
+    messagesCreatedAt: [new Date()],
+  });
+
+  return { agentConfiguration, authenticator, conversation, workspace };
+}
+
+describe("WakeUpResource.listByUserWithTotalCount", () => {
```
adrsimon: aren't we basically testing sequelize here for most of the test ?

adrsimon (approved):
LGTM, few nits

front/lib/api/assistant/wakeups.ts
fabiencelier: cost for not counting them is high, would need to resolve them all
I think it will surface rarely 

A cheap alternative would be to just do an "hasNext" and not ahve the total count

adrsimon: if the totalCount is wrong in the table i believe it's better to not display it right ?

fabiencelier: Ok, then let's drop the server side pagination entirely and return all wakeups then paginate UI side.
The list should not be too large
I think count is a nice to have when you have more than 10 so you have an overview of how many you have.

front/lib/resources/wakeup_resource.test.ts
fabiencelier: removed most of them
just kept one as sanity check that the resource method works

front/lib/api/assistant/wakeups.ts
```diff
@@ -0,0 +1,79 @@
+import type { Authenticator } from "@app/lib/auth";
+import { ConversationResource } from "@app/lib/resources/conversation_resource";
+import { WakeUpResource } from "@app/lib/resources/wakeup_resource";
+import type { ConversationRefType } from "@app/types/assistant/conversation";
+import type { WakeUpStatus, WakeUpType } from "@app/types/assistant/wakeups";
+import { removeNulls } from "@app/types/shared/utils/general";
+
+export interface UserWakeUpWithConversation {
+  conversation: ConversationRefType;
+  wakeUp: WakeUpType;
```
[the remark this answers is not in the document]
fabiencelier: fixed

PR #32207, opened 2026-09-10
[front] support tool markdown

ykmsd:
## Description
This PR is a part of supporting inline tool reference in the composer. We actually need to change a lot of things so to make diffs small here are the minor changes to support a tool directive in user message, and strip away icon from user messages.

In the next PR I will add actual logic to inline tools: 
CleanShot 2026-09-10 at 12 26 55@2x


front/lib/api/assistant/conversation_rendering/helpers.ts
```diff
@@ -461,7 +463,9 @@ export function renderUserMessage(
   conversation: ConversationWithoutContentType,
   m: UserMessageType
 ): UserMessageTypeModel {
-  const content = replaceMentionsWithAt(m.content);
+  const content = stripSkillTagPresentationAttributes(
```
ykmsd: am I correct to put it here like this? 👶

front/components/editor/extensions/skill_builder/ToolChip.tsx
```diff
@@ -6,7 +6,7 @@ import {
 import { AlertCircle, Chip, ShapesPlus } from "@dust-tt/sparkle";
 import type React from "react";
 
-function getToolIcon(toolIcon: string | null) {
+export function getToolIcon(toolIcon: string | null) {
```
ykmsd, line 9: we need to move some files (this should not live under skill_builder) but I will do it once I finish all the changes

front/components/assistant/UserMessageMarkdown.tsx
```diff
@@ -57,16 +95,28 @@ export const UserMessageMarkdown = ({
...
       pasted_attachment: PastedAttachmentBlock,
       file_preview: getFilePreviewPlugin(),
-      skill: ({ skillIcon, skillId, skillName }: SkillDirectiveProps) => (
-        <SkillBlock
-          skillIcon={skillIcon ?? null}
-          skillId={skillId}
-          skillName={skillName}
+      skill: ({ skillId, skillIcon, skillName }: SkillDirectiveProps) => {
+        return (
+          <AttachmentChipDirectiveBlock
```
ykmsd, line 73: delete SkillBlock and create a reusable one since it's the same as others

front/components/assistant/UserMessageMarkdown.tsx
```diff
@@ -57,16 +95,28 @@ export const UserMessageMarkdown = ({
...
+            icon={skillIcon ?? null}
+            getIcon={getSkillIcon}
+            onClick={() =>
+              togglePanel({ type: SKILL_SIDE_PANEL_TYPE, skillId })
+            }
+          />
+        );
+      },
+      tool: ({ toolIcon, toolName }: ToolDirectiveProps) => (
+        <AttachmentChipDirectiveBlock
```
ykmsd, line 84: no onClick handler yet

front/lib/api/assistant/conversation_rendering/helpers.ts
aubin-tchoi: we can do this, IMHO we don't really need to

aubin-tchoi (approved):
LGTM

front/lib/api/assistant/conversation_rendering/helpers.ts
ykmsd: ah then I will remove it 🫡

PR #32212, opened 2026-09-10
[front] perf: reduce agent mention count query cost

aubin-tchoi:
## Description

Counting agent mentions puts heavy load on Elasticsearch (see [monitor](https://dust4ai.slack.com/archives/C05F84CFP0E/p1788980486912309)). Restrict the ranking query to LLM consumption documents and lower its three cardinality aggregations from a precision threshold of 40,000 to Elasticsearch's default of 3,000 (we don't need a perfect precision, and this matches the actual order of magnitude expected), reducing aggregation work and per-agent counter memory.

## Tests

- Updated existing.

## Risk

- Low.

## Deploy Plan

- Deploy front.


smb2268 (approved):
lgtm!

PR #32213, opened 2026-09-10
dsbx db: take a ready sandbox, drop the pod-only paths

davidebbo:
**Stack 1/6** — groundwork for removing pod-owned sandboxes. Nothing here has a caller, before or after.

`dsbx_db.ts` carries two shapes for the same `dsbx db` commands: a `*OnReadySandbox` one that the Frame reconcile uses, and a pod-ensuring wrapper (`execDbCommand` → `ensurePodSandboxReady`) behind the rest. Pods are losing the ability to own a sandbox, so that wrapper's owner is going away. This moves every helper that only existed in wrapper form onto the ready-sandbox shape, and deletes the pod-only remainder.

### Moved onto the ready-sandbox shape

| before | after |
|---|---|
| `getDatabaseSchemaOnSandbox(auth, { space, database })` | `getDatabaseSchemaOnReadySandbox(auth, { sandbox, database })` |
| `queryDatabaseOnSandbox(auth, { space, database, sql })` | `queryDatabaseOnReadySandbox(auth, { sandbox, database, sql })` |
| `deleteDatabaseOnSandbox(auth, { space, database })` | `deleteDatabaseOnReadySandbox(auth, { sandbox, database })` |
| `deletePodDatabaseReplica(auth, space, { database })` | `deleteFrameDatabaseReplica(auth, frame, { database })` |

This matches `reconcileDatabaseOnReadySandbox` and `listDatabasesOnReadySandbox`, and it is what lets one `ensure` cover several databases — see [database_reconciliation.ts](front/lib/api/frames/database_reconciliation.ts), which ensures once and then loops the manifest.

The bodies were already owner-neutral: `SANDBOX_STATE_DATABASES_DIR` is shared, and `FRAME_DATABASE_NAME_REGEX` and `POD_DATABASE_NAME_REGEX` are both `SANDBOX_DATABASE_NAME_REGEX`, so the name guards needed nothing. Only the `space` parameter and a `podId` log field were pod-shaped.

**The query spill dir does change.** It pointed at the pod files mount (`/files/pod-{id}/.tool_outputs/db`), and a Frame-owned sandbox has no agent-visible files mount at all — `forFrameSandboxProvisioning` mounts only publications and state. Oversized results now spill to a non-mounted scratch root, alongside the one the regenerated schema file already used. `resultsFile` is still a sandbox path read back off the same sandbox, so it is scratch that does not outlive the sandbox rather than a pod file.

### Deleted

`reconcileDatabaseFromPodPath` and the private pod-ensuring reconcile behind it, plus the reconcile-side prefix helpers in `db_naming.ts` that only they used. None had a caller.

### One gap worth knowing about

The delete pair still has no caller, and there is a real one waiting: `reconcileFramePublicationDatabases` only ever adds or alters, so **a Frame publication that drops a database from its manifest leaves the live file and its replica behind indefinitely.** Wiring the pair into reconcile would mean publish starts destroying data with no undo — a manifest typo silently dropping a table's worth of rows — so that wants its own decision and its own PR rather than riding along here.

### Testing

`front` sandbox db + frames suites: 14 files / 97 tests pass. Typecheck clean in all three workspaces; biome clean.

### DIFF
--- a/front/lib/api/sandbox_functions/dsbx_db.ts
+++ b/front/lib/api/sandbox_functions/dsbx_db.ts
@@ -1,14 +1,7 @@
 import { randomUUID } from "node:crypto";
 import path from "node:path";
-import { DustFileSystem } from "@app/lib/api/file_system/dust_file_system";
-import { getRedisStreamClient } from "@app/lib/api/redis";
-import { isValidPodDatabaseName } from "@app/lib/api/sandbox/db";
 import { ensurePodSandboxReady } from "@app/lib/api/sandbox/lifecycle";
 import { shellEscape } from "@app/lib/api/sandbox/shell";
-import {
-  podDatabasePrefixFromPodPath,
-  resolvePodDatabaseName,
-} from "@app/lib/api/sandbox_functions/db_naming";
 import type { SandboxFunctionErrorCode } from "@app/lib/api/sandbox_functions/errors";
 import { SandboxFunctionError } from "@app/lib/api/sandbox_functions/errors";
 import type { StagingHashes } from "@app/lib/api/sandbox_functions/staging_integrity";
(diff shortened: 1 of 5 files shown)


front/lib/api/sandbox_functions/dsbx_db.ts
```diff
@@ -614,25 +454,24 @@ export interface QueryDatabaseResult {
...
-    // set becomes a pod file. Mirrors the pod mount in dust_file_system.ts. The var name must
-    // match POD_QUERY_SPILL_DIR_ENV in cli/dust-sandbox/src/commands/db/query.rs.
-    envVars: {
-      DUST_POD_QUERY_SPILL_DIR: `/files/${SCOPED_PREFIX_POD}${space.sId}/${TOOL_OUTPUTS_FOLDER_NAME}/db`,
-    },
+    // Oversized results spill here rather than into a files mount: a Frame-owned sandbox carries
+    // no agent-visible one. The spill is a plain sandbox path, so `resultsFile` is read back off
+    // the same sandbox the caller passed in. The var name must match POD_QUERY_SPILL_DIR_ENV in
+    // cli/dust-sandbox/src/commands/db/query.rs.
+    envVars: { DUST_POD_QUERY_SPILL_DIR: DB_QUERY_SPILL_ROOT },
```
davidebbo, line 474: This is probably broken. I opened https://github.com/dust-tt/tasks/issues/10365 to track

pmilliotte (commented):
LGTM

PR #32214, opened 2026-09-10
Poke: drop the pod sandbox surfaces

davidebbo:
**Stack 2/5** — base is #32213.

Three Poke views existed only to inspect or wake a pod-owned sandbox, which the rest of the stack removes.

### Pod databases table (project page)

Goes with its route, SWR hook and columns. Worth noting *why* it cannot survive rather than just that it is unused: it had no database-backed source. It ran `dsbx db list` against the pod on demand, **waking or cold-starting the sandbox to answer** — see the comment on the route. That is precisely the capability the owner kind is losing.

It was also the last caller of `listDatabasesOnSandbox`, so the pod-ensuring list wrapper goes with it. The `listDatabasesOnReadySandbox` variant that the Frame functions view uses stays.

### Sandbox rows (space Overview)

`Sandbox Status` and `Sandbox Connect` go from `ViewSpaceViewTable`, along with the `sandbox` field on the space-details response that fed them — only pods ever populated it, and the field was documented as `null` for every other space kind.

### Wake sandbox plugin

Goes. There is no pod sandbox left to wake.

Frame sandbox Poke surfaces are untouched.

### Testing

`front` sandbox_functions + poke suites: 28 files / 181 tests pass. Typecheck clean in `front`, `front-api` and `front-spa`; biome clean.


Fraggle (commented):
LGTM

PR #32215, opened 2026-09-10
Remove the two Pod functions specific features

davidebbo:
**Stack 3/6**

These are the only two *product* removals in the stack, so they are worth reviewing on their own rather than buried in a mechanical cleanup. Both can only ever have Pod functions, which are gone.

### Publish-time validation of Frame → Pod function references

It's dead code now, so removing. I opened https://github.com/dust-tt/tasks/issues/10366 to track bringing back some equivalent for Frames v2.

### Pod restriction-impact warning

The dialog that told an editor how many Pod function calls would break if they restricted an open Pod. It counted invocations of functions from `listBySpace`, so post-cleanup it reports zero forever and the warning silently never fires. `countByUserSince` existed for this alone and goes too.

If either intent is worth keeping for Frames, that is a new feature rather than a rescue. The validator especially: a Frame-function equivalent is better positioned than this one was — contracts (`inputSchema` **and** `outputSchema`) are already extracted at publish time by the same pipeline, nothing type-checks a v2 Frame's UI source today since esbuild strips types without checking them, and `validateFrameV2FromSource` already has a `warnings` channel so it need not block publish. The machinery here stays recoverable from history if we build it.

### Testing

`front` viz + types/api + assistant + pod + swr + invocation suites: 46 files / 377 tests pass. Typecheck clean in all three workspaces; biome clean.


PopDaph:
👍🏻

PR #32216, opened 2026-09-10
More Pod functions cleanup

davidebbo:
**Stack 4/5**

Mostly cleaning up obsolete Pod Function things.

Other notable:
- Flow a pod into the runContext when the frame lives in a pod
- When there is no pod, there is no place to write files, which is something we'll need to think about (https://github.com/dust-tt/tasks/issues/10367)
- The self-healing that automatically changes the function's executionMode no longer makes sense

### Testing

`front`: 183 files / 2639 tests pass. `front-api` sandbox routes: 6 files / 44 tests. Typecheck clean in all three workspaces; biome clean.


front/lib/actions/mcp_utils.ts
```diff
@@ -296,11 +296,18 @@ export async function handleBase64Upload(
...
-        auth,
-        runContext.invocation.sandboxFunction.space,
-        writeArgs
-      );
+      if (!runContext.pod) {
+        return {
+          content: {
+            type: "text",
+            text:
+              "This Frame does not run in a Pod, so there is nowhere to save the generated " +
```
davidebbo, line 304: See https://github.com/dust-tt/tasks/issues/10367

Fraggle (approved):
I lack most context but it looks gtm.

PR #32217, opened 2026-09-10
Another round of Pod sandbox removal

davidebbo:
**Stack 5/5** — base is #32216.

Conversation and Frame are the only owner kinds left. With every pod consumer gone in the earlier PRs, the owner itself can go.

### What goes

`PodSandboxAdapter`, `ensurePodSandboxReady`, the pod sandbox-only mounts (published bundles, litestream state replica) with their GCS prefixes and mount profile, and `DustFileSystem.forPodSandboxProvisioning`.

The `"pod"` arm of `SandboxRuntimeOwner` — which is what removes **`SPACE_ID`**: from the owner env vars, from the env manifest, from the reserved-name set in `SandboxResource`, and from the Fluent Bit enrichment, which now tags `FRAME_ID` instead. `sandboxOwnerHasPersistentState` loses its pod case; Frames still have durable state.

The reaper's pod branch — owner kind, auth map, adapter wiring. Its `requireRunning` suite moves to the Frame adapter, the other owner that passes the flag through, so that coverage is kept rather than dropped.

`scripts/ensure_pod_sandbox.ts`, and the `setupPodStateOnColdStart` / `ensurePodStateHealthOnSleep` compat aliases now that nothing calls the Pod-named exports.

### What deliberately stays

`buildPodScopedEnvVars` moves to `SandboxEnvVarResource.loadPodScopedEnv` rather than going away — its owner adapter is leaving, but the two remaining adapters still need it. **Pod-scoped env vars, HTTPS secrets and egress policy still apply to every conversation and Frame sandbox running inside a Pod**, and so do Pod file mounts. None of that is touched.

### The one thing that needs an operator

`deleteLegacyPodOwnedSandbox` is what is left to reach a pod-owned sandbox that already exists. Pod deletion calls it, because with no owner adapter the reaper skips such a sandbox forever and it would sit running at the provider indefinitely.

**A follow-up PR adds the backfill that drains them, and it needs to run once soon after this ships.** Until then every pod that ever booted a sandbox is paying for an unreapable VM.

### Still to do after the stack

The two `spaceId` columns (`sandbox_owners`, `sandbox_functions`) are untouched. `sandbox_functions.spaceId` is now only ever written as `null` and read nowhere; `sandbox_owners.spaceId` is read only by the space-delete cleanup and the legacy helper. Dropping them needs a migration, and that should follow the backfill run.

### Testing

`front`: 109 files / 1080 tests pass. Typecheck clean in all three workspaces; biome clean.


front/lib/api/sandbox/legacy_pod_sandbox.ts
```diff
@@ -0,0 +1,52 @@
...
+ * holds the provider sandbox and its `sandbox_owners` row. Destroying the provider sandbox is what
+ * stops it being stranded: with no owner adapter left, the reaper skips it forever.
+ *
+ * `dropOwnerLink` says who removes the `sandbox_owners` row. Pod deletion leaves it to
+ * `SpaceResource.delete`, which has to drop it anyway (the FK is `onDelete: "RESTRICT"`); the
+ * backfill script keeps the pod, so it drops the row itself.
+ *
+ * Delete this module once production holds no `sandbox_owners` row with a `spaceId`.
+ */
+export async function deleteLegacyPodOwnedSandbox(
```
davidebbo, line 20: Temporary during transition, will remove this later

davidebbo: https://github.com/dust-tt/tasks/issues/10373

pmilliotte (approved):

PR #32219, opened 2026-09-10
fix(front): keep agent picker row actions above the dropdown scrollbar

achilleburah:
## Description

In the agent picker, the scroll area draws is on top of the "agent details" button and prevent click on a big part of the button. 
Adding a bigger z-index fixes this. 

https://github.com/user-attachments/assets/1429da09-6655-4118-9854-9f6f0fab2675

## Tests

Previewed locally.

## Risk

Low. z-index only

## Deploy Plan

Regular front deploy.

### DIFF
--- a/front/components/assistant/AgentPicker.tsx
+++ b/front/components/assistant/AgentPicker.tsx
@@ -140,7 +140,7 @@ export function AgentPicker({
                   isSelected ? "bg-primary-100" : ""
                 }`}
                 endComponent={
-                  <div className="z-10 flex items-center gap-1">
+                  <div className="flex items-center gap-1">
                     {isSelected && (
                       // Show a tick by default; on hover swap it for an X to
                       // signal that clicking will deselect the agent.
--- a/front/components/assistant/AgentPicker.tsx
+++ b/front/components/assistant/AgentPicker.tsx
@@ -140,7 +140,7 @@ export function AgentPicker({
                   isSelected ? "bg-primary-100" : ""
                 }`}
                 endComponent={
-                  <div className="flex items-center gap-1">
+                  <div className="z-10 flex items-center gap-1">
                     {isSelected && (
                       // Show a tick by default; on hover swap it for an X to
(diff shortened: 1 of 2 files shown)


front/components/assistant/AgentPicker.tsx
```diff
@@ -140,7 +140,7 @@ export function AgentPicker({
                   isSelected ? "bg-primary-100" : ""
                 }`}
                 endComponent={
-                  <div className="flex items-center gap-1">
+                  <div className="z-10 flex items-center gap-1">
```
ykmsd: but it's weird to hide the scroll bar behind the button no? 

Image

ykmsd: I think we should rather move the scroll bar so it doesn't overwrap, let me think

achilleburah: It's a little weird I agree but that only occurs when hovering the first button. When currently even lower buttons (not directly behind the scroll handle) are also hidden.

ykmsd: so the problem is we apply p-1 spacing to the dropdown container, but the scroll bar is located inside this spacing: 

Image

Image

I think we should move this container spacing to individual elements (I guess header and content?) so that scroll will be outside of this spacing box, so should look something like that? 

Image

achilleburah: deployed a new version, let me know if that's what you had in mind.

ykmsd, line 143: so we don't need this anymore right?

ykmsd (approved):
ty!

front/components/assistant/AgentPicker.tsx
achilleburah: yes we still do actually because the scroll bar area is actually mush bigger than it looks. Without the z-index, you won't "see" the overlap, but the button will still be masked.

PR #32220, opened 2026-09-10
chore(sandbox-reaper): bound authenticator fan-out to the connection pool

adrsimon:
## Description

`sandboxReaperWorkflow` peaked at 19 concurrency over the last few days.

A batch holds up to 128 sandboxes spanning as many workspaces. Authenticators were built in three separate executors wrapped in one `Promise.all`, one per owner kind, so the real ceiling was 48 concurrent builds. Each build issues about four queries and has its own nested `Promise.all`. 

Frame and pod requested the same authenticator (`internalAdminForWorkspace` with `dangerouslyRequestAllGroups`), so a workspace with both kinds in one batch built it twice, concurrently.

The three executors are now one over a flat list. `REAPER_CONCURRENCY` stays at 16 for the lifecycle executor so capacity reclamation in the `kill_requested` and `running` phases keeps its current speed.

## Tests

Worth watching `peakConcurrentQueries` for `sandboxReaperWorkflow` after deploy. Expect around 8 for the auth phase, with the lifecycle phase as the new ceiling.

## Risk

Low. 

## Deploy Plan

Deploy front.

### DIFF
--- a/front/temporal/sandbox_reaper/activities.ts
+++ b/front/temporal/sandbox_reaper/activities.ts
@@ -23,6 +23,7 @@ import {
 } from "./config";
 
 const REAPER_CONCURRENCY = 16;
+const REAPER_AUTH_CONCURRENCY = 4;
 
 export type ReaperPhase =
   | "kill_requested"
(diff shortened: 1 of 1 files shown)


front/temporal/sandbox_reaper/activities.ts
```diff
@@ -77,97 +78,113 @@ type SandboxOwnerMaps = {
...
+type ReaperAuthMaps = Record<ReaperAuthKind, Map<ModelId, Authenticator>>;
+
+function getAuthKindForOwnerKind(
+  kind: ReaperSandboxLifecycleOwner["kind"]
+): ReaperAuthKind {
+  switch (kind) {
+    case "conversation":
+      return "conversation";
+    case "frame":
+    case "pod":
```
davidebbo, line 94: I'll probably get a conflict as I'm getting rid of pods, but I'll deal with that.

davidebbo (approved):
LGTM

PR #32221, opened 2026-09-10
Only front changelogs credit per-commit authors

flvndvd:
## Description

Every deploy posts a changelog to Slack crediting the author of each commit in the range. That attribution only carries meaning on front deploys: we work in a monorepo and front is where the whole team's work ships, so seeing your name there tells you your commit reached production. On every other component (front-edge, connectors, core, ...) the range is just whatever landed on main since the last run, and the credit reads as if a crowd shipped a deploy none of them started.

The changelog now credits per-commit authors on front only. Everywhere else the commit list stays and the person who triggered the deploy remains named (and mentioned) in the start message. The author names were plain text from git log, not Slack mentions, so nobody was being pinged either way, this is about attribution.

## Tests

None, the change is a format string picked by component in the commit-diff step. The next deploys are the proof.

## Risk

Notification formatting only, nothing in the build or deploy path.

## Deploy Plan

Merge, takes effect on the next deploys.

### DIFF
--- a/.github/actions/build-image/action.yml
+++ b/.github/actions/build-image/action.yml
@@ -280,12 +280,13 @@ runs:
 
         echo "📋 Getting commit diff: $CURRENT_TAG → $TARGET_TAG"
 
-        # front-edge changelogs skip per-commit authors, the run initiator is
-        # already named in the start message.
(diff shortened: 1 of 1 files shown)


adrsimon (approved):

avervaet (approved):
LGTM

PR #32222, opened 2026-09-10
[front] chore: split `ConversationPage` component

aubin-tchoi:
## Description

`ConversationPage` has grown to ~1.5k lines. It handles many things: page layout, message rendering, tool details, and debugging actions.

This PR split it into a few focused components and move the render request into a Poke hook

Included 2 minor QoL improvement: for tools the entire row is clickable to expand it vs only the small icon, and when the timestamp is not available we just leave it blank instead of showing an em dash.

## Tests

- Tested locally.

## Risk

- Low.

## Deploy Plan

- Deploy front.


ykmsd (approved):
👍

PR #32223, opened 2026-09-10
Normalize connector errors and guard Snowflake failures

spolu:
## Description

Replace all 20 `errorFromAny` call sites with `normalizeError` and delete the helper, which could fail on null or undefined and returned a plain object instead of an `Error`.

Validate Snowflake error fields before inspecting them. Preserve recognized provider failures, normalize plain payloads before wrapping them, and rethrow unrecognized values unchanged instead of masking them with a `TypeError`. This fixes three more guard candidates from the `normalize-caught-errors` inventory.

## Tests

23 Snowflake interceptor cases pass with an isolated Vitest configuration, without database setup. Nine regression cases failed before the fix. Coverage includes numeric/string role codes, recognized errors, plain provider payloads, and malformed or primitive thrown values.

## Risk

Low. 

## Deploy Plan

- Deploy `connectors`.

### DIFF
--- a/connectors/src/api/create_connector.ts
+++ b/connectors/src/api/create_connector.ts
@@ -3,7 +3,6 @@ import type {
   ConnectorManagerError,
   CreateConnectorErrorCode,
 } from "@connectors/connectors/interface";
-import { errorFromAny } from "@connectors/lib/error";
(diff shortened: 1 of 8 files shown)


philipperolet (commented):
LGTM

PR #32226, opened 2026-09-10
fix(conversations): commit each attachment upload as soon as it settles

Nils-Fedrigo:
## Description

https://github.com/user-attachments/assets/048acdd7-a947-495d-bffc-06b5cd55a4b7

Instead of waiting the whole batch to run processing on a file, do it as you go. 

Previously `uploadFiles` ran the files through `concurrentExecutor` and `handleFilesUpload` called `processResults` on the returned array
afterwards.

`onResult` is optional, so the other ~455 `concurrentExecutor` call sites are unaffected.

Side effect: error notifications now appear as each file fails rather than batched at the end. :)

## Tests

`npm run test -- lib/utils/async_utils.test.ts` — 4 passing, including a new case pinning that a
fast item is reported while a slow one is still pending, and that the returned array still follows
input order. `npx tsgo --noEmit` clean across all call sites.


front/hooks/useFileUploaderService.ts
```diff
@@ -284,7 +287,7 @@ export function useFileUploaderService({
             path: fileUploaded.path,
           });
         },
-        { concurrency: 4 }
+        { concurrency: 4, onResult: options?.onFileSettled }
```
PopDaph: Could we do this here instead of extending `concurrentExecutor`? It's used in hundreds of places, and the hook already owns the iterator, so the same effect is reachable locally:

```ts
const uploadOne = async (
  fileBlob: FileBlob
): Promise<Result<FileBlob, FileBlobUploadError>> => {
  // current iterator body, unchanged
};

return concurrentExecutor(
  newFileBlobs,
  async (fileBlob) => {
    const result = await uploadOne(fileBlob);
    options?.onFileSettled?.(result);
    return result;
  },
  { concurrency: 4 }
);

handleFilesUpload stays as you have it. The async_utils change, its new test and the contract would then be unneeded.

Not against onResult as a feature, just wondering if the smaller diff is enough for this fix. Wdyt?
```


WDYT @flvndvd ?

Nils-Fedrigo: yeah indeed we can do this!!

Nils-Fedrigo: Just checking if Flavien has time to share his position, but else I can proceed with no concurrentExecutor modification instead.

front/lib/utils/async_utils.ts
```diff
@@ -13,12 +13,21 @@ import { normalizeError } from "@app/types/shared/utils/error_utils";
...
+ * `onResult` MUST run once per item, as soon as that item's iterator resolves, and MUST NOT be
+ * batched until the end: callers use it to release per-item state while the batch is still running.
+ */
 export async function concurrentExecutor<T, V>(
   items: T[] | readonly T[],
   iterator: (item: T, idx: number) => Promise<V>,
-  { concurrency = 8 }: { concurrency: number }
+  {
+    concurrency = 8,
+    onResult,
```
flvndvd: `onResult` feels somewhat misleading as a name, maybe it's more `onItemDone`?

front/hooks/useFileUploaderService.ts
flvndvd: Aligned with @PopDaph on the proposal.

PopDaph (approved):

PR #32228, opened 2026-09-10
feat(connectors): sync SharePoint text files reported as octet-stream

tdraier:
## Description

When SharePoint reports a `.txt` file's content-type via Microsoft Graph as `application/octet-stream` (the generic binary default) instead of `text/plain`, the Microsoft connector silently skips it. At ICF Habitat, 31 of 38 legacy COBOL `.txt` source files never sync for this reason, with no user-facing error. Re-encoding/re-uploading doesn't help — the customer can't control the per-file content-type SharePoint assigns.

This adds an **extension-based MIME fallback** (`resolveMicrosoftMimeType`): when Graph reports `application/octet-stream` (or no type) **and** the filename ends in a known text extension (`.txt`, `.md`, `.markdown`), the file is treated as `text/plain`/`text/markdown`. Any other type is returned unchanged, so unknown-extension binaries are still skipped exactly as before.

The resolver is used by **both** mime-type sync gates — the `syncFiles` pre-filter (`activities.ts`) and the per-file `syncOneFile` gate (`file.ts`) — so the accept/skip decision and the downstream handler routing agree, and the resolved type is what gets stored on the node and upserted to core.

We deliberately do **not** whitelist `application/octet-stream` broadly (rejected on dust-tt/tasks#9014): only explicit text extensions are reclassified, and `handleTextFile`'s existing letter-ratio heuristic still rejects any binary that happens to carry a text extension.

Relates to dust-tt/tasks#9122 (follow-up to dust-tt/tasks#9014).

## Tests

- New unit tests `connectors/src/connectors/microsoft/temporal/mime_types.test.ts` (6 cases): real types pass through, octet-stream/absent + known extension → text, case-insensitive extension, unknown extension + octet-stream stays octet-stream, absent type + no match → undefined.
- `npx tsgo --noEmit` clean; `biome check` clean; all 6 tests pass.

## Risk

Low. Purely widens which files are accepted; no schema or API changes. Only files that (a) are reported as `octet-stream`/untyped and (b) carry a `.txt`/`.md`/`.markdown` extension are newly considered — and they still pass through the same content heuristics before upsert. Fully rollback-safe (revert only); no data migration. Worst case a genuinely-binary `.txt` is downloaded and then rejected by `handleTextFile`.

## Deploy Plan

Standard connectors deploy. Newly-eligible files sync on the next incremental/batch sync of affected Microsoft connections; no backfill step required.

### DIFF
--- a/connectors/src/connectors/microsoft/temporal/activities.ts
+++ b/connectors/src/connectors/microsoft/temporal/activities.ts
@@ -52,7 +52,10 @@ import {
   updateDescendantsParentsInCore,
   // biome-ignore lint/suspicious/noImportCycles: ignored using `--suppress`
 } from "@connectors/connectors/microsoft/temporal/file";
-import { getMimeTypesToSync } from "@connectors/connectors/microsoft/temporal/mime_types";
+import {
+  getMimeTypesToSync,
+  resolveMicrosoftMimeType,
+} from "@connectors/connectors/microsoft/temporal/mime_types";
 import { connectorsConfig } from "@connectors/connectors/shared/config";
 import { dataSourceConfigFromConnector } from "@connectors/lib/api/data_source_config";
 import {
@@ -748,10 +751,10 @@ export async function syncFiles({
       pdfEnabled: providerConfig.pdfEnabled || false,
       csvEnabled: providerConfig.csvEnabled || false,
     });
-    const filesToSync = children.filter(
-      (item) =>
-        item.file?.mimeType && mimeTypesToSync.includes(item.file.mimeType)
-    );
+    const filesToSync = children.filter((item) => {
+      const mimeType = resolveMicrosoftMimeType(item);
+      return mimeType != null && mimeTypesToSync.includes(mimeType);
+    });
 
     const concurrency = getAdaptiveConcurrency(
       filesToSync,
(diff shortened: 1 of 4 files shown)


tdraier:

- #32228 ◀

philipperolet (approved):
LGTM 🙏 (will let you check the cc violations)

connectors/src/connectors/microsoft/temporal/activities.ts
```diff
@@ -748,10 +751,10 @@ export async function syncFiles({
       pdfEnabled: providerConfig.pdfEnabled || false,
       csvEnabled: providerConfig.csvEnabled || false,
     });
-    const filesToSync = children.filter(
-      (item) =>
-        item.file?.mimeType && mimeTypesToSync.includes(item.file.mimeType)
-    );
+    const filesToSync = children.filter((item) => {
```
philipperolet, line 754: Oooh actually, I approved too fast; A folder named notes.txt now resolves to text/plain, reaches syncOneFile, and throws Item is not a file, preventing full sync from advancing
Edge case but will likely happen

tdraier: a folder will never send application/octet-stream as mimetype, so will not be catched here. Only application/octet-stream + .txt -> text/plain

philipperolet: doesn't the fallback also applies when the mimetype is absent? which could be the case for folders no? i had an agent check the doc, no indication that they return a mimetype for folders

tdraier: folders have a ~"microsoft/folder" mimetype like . I removed the fallback on empty mimetype anyway to avoid other edge cases (it was in contradiction with m own cc 😆  ), we only want  to handle `application/octet-stream`

philipperolet: 👍 thanks 🙏

PR #32229, opened 2026-09-10
Guard caught HubSpot and Notion error payloads

spolu:
## Description

Guard caught provider payloads at the four HubSpot sites and the Notion garbage-collection site from the `normalize-caught-errors` inventory. Null or undefined failures no longer cause another error while their fields are inspected.

HubSpot keeps its numeric-404 fallbacks and normalizes other failures. Notion keeps recognizing its existing unhealthy error codes and numeric 5xx statuses, preserving resources after the existing retry threshold and propagating other failures as `Error` instances. Add local contracts for the 404 detection and garbage-collection retention behavior.

## Tests

N/A

## Risk

Low

## Deploy Plan

- Deploy `front`.
- Deploy `connectors`.


philipperolet (commented):
LGTM

PR #32238, opened 2026-09-10
[front] Inline tool reference 

ykmsd:
## Description
This PR is to support inline tool reference in the composer. Previously we create a conversation with selected tools, but for existing conversations we make a post request on the moment we modify the tool selection instead of submit time. In this PR you can no longer delete the tool once you attach, and we send it to alongside with a message instead of updating live when it's added. 

Note that clicking a chip does nothing right now, this will be supported in my next PR 🫡

If you want to test it, you can enable the ff `inline_tool_knowledge_reference` from my dev console tool and most likely you need to reload the page.

### DIFF
--- a/front/components/assistant/conversation/input_bar/InputBar.tsx
+++ b/front/components/assistant/conversation/input_bar/InputBar.tsx
@@ -270,13 +270,13 @@ export const InputBar = React.memo(function InputBar({
     workspaceId: owner.sId,
   });
 
-  const IsInlineReferenceEnabled = featureFlags.includes("inline_tool_knowledge_reference");
+  const isInlineReferenceEnabled = featureFlags.includes("inline_tool_knowledge_reference");
 
   useEffect(() => {
-    if (!IsInlineReferenceEnabled) {
+    if (!isInlineReferenceEnabled) {
       setSelectedMCPServerViews(conversationTools);
     }
-  }, [conversationTools, IsInlineReferenceEnabled]);
+  }, [conversationTools, isInlineReferenceEnabled]);
 
   const { addTool, deleteTool } = useAddDeleteConversationTool({
     conversationId: conversation?.sId,
@@ -413,7 +413,7 @@ export const InputBar = React.memo(function InputBar({
 
   const handleMCPServerViewSelect = useCallback(
     (serverView: MCPServerViewLightType) => {
-      if (!IsInlineReferenceEnabled && selectedMCPServerViewIds.has(serverView.sId)) {
+      if (!isInlineReferenceEnabled && selectedMCPServerViewIds.has(serverView.sId)) {
         return;
       }
 
@@ -422,17 +422,17 @@ export const InputBar = React.memo(function InputBar({
           ? prev
           : [...prev, serverView]
       );
-      if (!IsInlineReferenceEnabled) {
+      if (!isInlineReferenceEnabled) {
         void addTool(serverView.sId);
       }
     },
-    [addTool, IsInlineReferenceEnabled, selectedMCPServerViewIds]
+    [addTool, isInlineReferenceEnabled, selectedMCPServerViewIds]
   );
 
   const handleMCPServerViewDeselect = useCallback(
     (serverView: MCPServerViewLightType) => {
       if (
-        !IsInlineReferenceEnabled &&
+        !isInlineReferenceEnabled &&
         !selectedMCPServerViewIds.has(serverView.sId)
       ) {
         return;
@@ -441,11 +441,11 @@ export const InputBar = React.memo(function InputBar({
       setSelectedMCPServerViews((prev) =>
         prev.filter((sv) => sv.sId !== serverView.sId)
       );
-      if (!IsInlineReferenceEnabled) {
+      if (!isInlineReferenceEnabled) {
         void deleteTool(serverView.sId);
       }
     },
-    [deleteTool, IsInlineReferenceEnabled, selectedMCPServerViewIds]
+    [deleteTool, isInlineReferenceEnabled, selectedMCPServerViewIds]
   );
 
   const clearSideChannelSelections = useCallback(async () => {
@@ -455,12 +455,12 @@ export const InputBar = React.memo(function InputBar({
     setSelectedMCPServerViews([]);
     setAttachedNodes([]);
 
-    if (!IsInlineReferenceEnabled) {
+    if (!isInlineReferenceEnabled) {
       await Promise.all(
         serverViewIds.map((serverViewId) => deleteTool(serverViewId))
       );
     }
-  }, [deleteTool, IsInlineReferenceEnabled, selectedMCPServerViews]);
+  }, [deleteTool, isInlineReferenceEnabled, selectedMCPServerViews]);
 
   const handleSelectedSpaceIdsChange = useCallback(
     async (spaceIds: string[]): Promise<string[] | null> => {
@@ -561,12 +561,12 @@ export const InputBar = React.memo(function InputBar({
       mentions.some((m) => m.id === a.sId && m.type === "agent")
     );
 
-    const messageTools = IsInlineReferenceEnabled ? extractToolTags(markdown) : [];
+    const messageTools = isInlineReferenceEnabled ? extractToolTags(markdown) : [];
     const toolIdsToAttach = getToolIdsToAttach(
       messageTools,
       new Set(conversationTools.map((serverView) => serverView.sId))
     );
-    const trackedTools = IsInlineReferenceEnabled
+    const trackedTools = isInlineReferenceEnabled
       ? messageTools.map((t) => t.name)
       : selectedMCPServerViews.map((t) => t.server.name);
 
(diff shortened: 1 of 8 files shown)


front/components/assistant/conversation/input_bar/InputBarContainer.tsx
```diff
@@ -1098,50 +1141,85 @@ const InputBarContainer = ({
...
-  }, []);
+    },
+    [onNodeUnselect, onMCPServerViewDeselect, selectedMCPServerViews]
+  );
 
   // Update the editor ref when the editor is created and listen for updates to the editor.
   useEffect(() => {
     if (editorRef.current) {
       editorRef.current.off("update", handleEditorUpdate);
+      editorRef.current.off("delete", handleContentDeleted);
```
ykmsd, line 1206: I don't think we need them but don't wanna break things right now so will keep it

front/components/assistant/conversation/input_bar/InputBar.tsx
```diff
@@ -549,23 +566,34 @@ export const InputBar = React.memo(function InputBar({
       mentions.some((m) => m.id === a.sId && m.type === "agent")
     );
 
+    const messageTools = isInlineReferenceEnabled
+      ? extractToolTags(markdown)
+      : [];
+    const toolIdsToAttach = getToolIdsToAttach(
```
ykmsd, line 572: am I correct to understand that we only need send diffs?

front/components/assistant/conversation/input_bar/InputBarContainer.tsx
```diff
@@ -185,6 +188,25 @@ function readDefaultSkillEditorState(editor: Editor): {
...
 }
 
+function hasAnotherAttachedNode(
+  editor: Editor,
+  nodeType: string,
+  attrName: string,
+  value: string
+): boolean {
+  let found = false;
+  editor.state.doc.descendants((node) => {
```
ykmsd, line 198: so this is prose mirror thing and apparently you cannot stop the iteration in the middle so we keep track of found variable on top

front/components/assistant/conversation/input_bar/InputBarContainer.tsx
```diff
@@ -1098,50 +1140,85 @@ const InputBarContainer = ({
       userMentioned,
       editorStartsWithUserMention
     );
+  }, []);
+
+  const handleContentDeleted = useCallback(
```
ykmsd, line 1145: I think instead of handling everything inside `handleEditorUpdate` it's better to have a separate delete handler so I moved out the content node deletion logic and added tool node deletion logic

front/components/assistant/conversation/input_bar/InputBarContainer.tsx
```diff
@@ -1098,50 +1140,85 @@ const InputBarContainer = ({
...
     editorRef.current = editor;
 
     return () => {
       if (editor) {
         editor.off("update", handleEditorUpdate);
+        editor.off("delete", handleContentDeleted);
       }
     };
-  }, [editor, handleEditorUpdate]);
+  }, [editor, handleEditorUpdate, handleContentDeleted]);
```
ykmsd, line 1221: IIUC editor on/off doesn't trigger re-render so it's safe to run on handleContentDeleted reference changes

front/components/assistant/conversation/input_bar/InputBar.tsx
```diff
@@ -641,8 +670,16 @@ export const InputBar = React.memo(function InputBar({
...
+        if (isInlineReferenceEnabled) {
+          setSelectedMCPServerViews([]);
+        }
 
-        await submitPromise;
+        const r = await submitPromise;
+        if (r.isOk() && toolIdsToAttach.length > 0) {
+          // Newly attached tools are now part of the conversation's server-side set; refresh it
+          // so the next submit's diff doesn't re-send them.
+          void mutateConversationTools();
```
ykmsd, line 681: I can also just update the cache but I thiiink it's safer to fetch it every time?

front/components/editor/input_bar/useCustomEditor.tsx
```diff
@@ -14,6 +14,7 @@ import { URLDetectionExtension } from "@app/components/editor/extensions/input_b
 import { URLStorageExtension } from "@app/components/editor/extensions/input_bar/URLStorageExtension";
 import { MentionExtension } from "@app/components/editor/extensions/MentionExtension";
 import type { SlashCommand } from "@app/components/editor/extensions/shared/slash_suggestion/SlashCommandDropdown";
+import { ToolNodeWithView } from "@app/components/editor/extensions/skill_builder/ToolNodeWithView";
```
ykmsd, line 17: this is under /skill_builder right now but I will move everything once the feature is done

front/components/assistant/conversation/input_bar/InputBar.tsx
aubin-tchoi: yes I agree

aubin-tchoi: wait but why do you need a local state? you could even remove the call to the endpoint and fold it in the postUserMessage

ykmsd: but to send it we need to keep the current state somewhere? or you mean we should parse it from the message when we send it?

aubin-tchoi: ah wait I though this was a separate endpoint, re-reading

ykmsd: so there are two endpoints indeed, and before we send selected views in message endpoint for new conversation , and then for existing conversation we use another endpoint to add/delete, and I think it's simpler to attach to message endpoint? 🤔

front/components/assistant/conversation/input_bar/InputBar.tsx
```diff
@@ -549,23 +566,34 @@ export const InputBar = React.memo(function InputBar({
       mentions.some((m) => m.id === a.sId && m.type === "agent")
     );
 
+    const messageTools = isInlineReferenceEnabled
+      ? extractToolTags(markdown)
```
aubin-tchoi, line 570: not a fan of doing this in the front-end, whenever we have a tiptap editor we should only rely on tiptap to deserialize the nodes and should never have to parse regexes ourselves

ykmsd: I'm a bit lost in your comment, I initially thought what you meant is something like that https://github.com/dust-tt/dust/pull/32238/changes/5dcf566b6ee2954bedf6ee84c9bad5165d3dd958 but actually are you suggesting to do that completely backend and remove passing them as `selectedMCPServerViewIds`? (seems like this is how we do for skills?)

front/components/assistant/conversation/input_bar/InputBarContainer.tsx
```diff
@@ -1098,50 +1140,85 @@ const InputBarContainer = ({
...
-      // Update the tracked set and unselect nodes whose chip was removed.
-      const prevIds = dataSourceLinkNodeIdsRef.current;
-      for (const prevId of prevIds) {
-        if (!chipNodeIds.has(prevId)) {
-          const node = attachedNodesRef.current.find(
-            (n) => n.internalId === prevId
+      if (node.type.name === TOOL_NODE_TYPE) {
+        const mcpServerViewId = node.attrs.mcpServerViewId;
+        if (
+          typeof mcpServerViewId === "string" &&
```
aubin-tchoi: nit: we could use `isString` here

front/components/assistant/conversation/input_bar/InputBarContainer.tsx
```diff
@@ -1098,50 +1140,85 @@ const InputBarContainer = ({
...
+      }
 
-      // Update the tracked set and unselect nodes whose chip was removed.
-      const prevIds = dataSourceLinkNodeIdsRef.current;
-      for (const prevId of prevIds) {
-        if (!chipNodeIds.has(prevId)) {
-          const node = attachedNodesRef.current.find(
-            (n) => n.internalId === prevId
+      if (node.type.name === TOOL_NODE_TYPE) {
+        const mcpServerViewId = node.attrs.mcpServerViewId;
```
aubin-tchoi: nit:
```suggestion
        const { mcpServerViewId } = node.attrs;
```

front/components/assistant/conversation/input_bar/InputBar.tsx
aubin-tchoi: you should send `messageTools` in its entirety (and ignore `conversationTools`) IMHO, since it's an upsert it's fine to re-send the same ones, and this way we follow one simple logic of always sending the message's tools

front/components/assistant/conversation/input_bar/InputBar.tsx
aubin-tchoi: yeah we should always send them in the message endpoint alongside the message, new logic could be: when submitting the message we get the tools from tiptap and send them in the payload, and that's it: no need to call `useConversationTools` at all

aubin-tchoi (approved):
LGTM

front/components/assistant/conversation/input_bar/InputBar.tsx
ykmsd: okok

PR #32244, opened 2026-09-10
Size the members usage off-pace column to its content

avervaet:
## Description

On the poke Pool Usage page, the members usage table's off-pace column (pace icon or "Unblock" button) had a fixed 7rem width, wasting space when nobody is blocked.

The column now starts at icon width (3rem) and switches to the button width (7rem) the first time a spend-capped member is rendered on any page, then keeps it. Whether a member is blocked is only known page by page (the members-usage endpoint computes it per page), so latching avoids the column resizing on every page change.

## Tests

-Local manual testing

## Risk

Low. Layout change only for the members usage table 

## Deploy Plan

Deploy front

### DIFF
--- a/front/components/workspace/MembersUsageTable.tsx
+++ b/front/components/workspace/MembersUsageTable.tsx
@@ -70,7 +70,7 @@ import type {
   RowSelectionState,
   SortingState,
 } from "@tanstack/react-table";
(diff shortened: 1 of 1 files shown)


aubin-tchoi (commented):
LGTM

PR #32245, opened 2026-09-10
Keep the usage page scroll offset when switching tabs

PopDaph:
## Description

On `/w/<wId>/usage`, switching tab after scrolling down jumps the page. Radix unmounts the previous tab panel and the new one starts as a spinner (or is just shorter), so the scroll container shrinks and the browser clamps the scroll offset. When the data lands the page grows back but the offset stays clamped.

Each `TabsContent` now gets `min-h-panel` (new utility next to `h-panel`, the height of the scrolling panel), so the content under the tab list is never shorter than the viewport and the offset is never clamped. Sparkle's `TabsContent` renders as `display: contents`, hence the added `block`.

Trade-off: short tabs (Groups with few groups) get up to one screen of empty space below the table, only visible when scrolling past the content.

## Tests

Manual, locally with 41 members: scroll offset stays identical across Members / Groups / Top-ups / Settings, both while loading and after.

## Risk

Low, styling only on the usage page.

## Deploy Plan

Deploy front.

### DIFF
--- a/front/components/pages/workspace/UsagePage.tsx
+++ b/front/components/pages/workspace/UsagePage.tsx
@@ -216,7 +216,8 @@ const DEFAULT_PAGE_SIZE = 25;
 
 // Keep every tab panel at least as tall as the scrolling panel so switching to a
 // shorter (or still loading) tab never shrinks the page and clamps the scroll offset.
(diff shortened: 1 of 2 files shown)


avervaet (approved):
🙏

PR #32247, opened 2026-09-10
fix: normalize caught errors across result and callback handlers

spolu:
## Description

Replace ad hoc caught-error conversions with `normalizeError` at 20 result, MCP/sidekick, and text-extraction callback sites. Existing errors retain their original stack, cause, subtype, and fields; structured non-Error failures retain their details instead of becoming `[object Object]` or a generic message. Document the helper's identity-preservation contract.

## Tests

- Four existing SDK tests passed.
- Sixteen local stream checks passed across the front and connectors text/CSV transforms, covering transform and flush failures with Error and structured-object rejections.
- Billing endpoint suites could not start because the local `@novu/framework` dependency is missing.

## Risk

Low. Uses existing normalization helpers. Non-Error diagnostic messages become more descriptive. Local front validation is limited by missing or outdated dependencies.

## Deploy Plan

Deploy affected front services and connectors. Include SDK changes in the next SDK release.

### DIFF
--- a/connectors/src/connectors/bigquery/lib/bigquery_api.ts
+++ b/connectors/src/connectors/bigquery/lib/bigquery_api.ts
@@ -307,7 +307,7 @@ export const fetchTables = async ({
 
     return new Ok(removeNulls(remoteDBTables));
   } catch (error) {
-    return new Err(error instanceof Error ? error : new Error(String(error)));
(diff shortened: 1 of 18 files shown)


Fraggle (approved):
nice cleanup !

PR #32248, opened 2026-09-10
Only reveal cycles with consumption in the usage cycle history and add a show less control

avervaet:
## Description

On the usage page and the poke pool usage page, the previous cycles table fetched N invoices then dropped the ones without consumption. "Load more" could therefore append fewer than a page of rows, or none at all, and the footer offered "load more" whenever older invoices existed even if none of them would render.

The backend now scans a fixed window of finalized invoices and counts only cycles with consumption toward the requested limit. A workspace only ever renders one of the two breakdowns (pool vs. excess), so `hasMoreCycleHistory` is returned as `{ cycleBreakdown, excessCycleBreakdown }` instead of a single flag — the frontend checks the one it actually displays. Otherwise "load more" could show up when only the *other*, unrendered breakdown had more, reproducing the same bug for that case.

Also add a show less property to the Sparkle loading table component.

## Tests

- tested locally and on front-edge
- added a regression test asserting the pool and excess "more history" flags stay independent of each other

## Risk

Low. Each cycle history request now lists up to 48 invoices instead of `limit + 1` (one or two Metronome pages), cached per workspace and limit for a minute as before. `hasMoreCycleHistory` changes shape from a boolean to `{ cycleBreakdown, excessCycleBreakdown }` — consumed only by this PR's own frontend changes.

## Deploy

front, sparkle

### DIFF
--- a/front/lib/api/credits/awu_pool_summary.ts
+++ b/front/lib/api/credits/awu_pool_summary.ts
@@ -439,11 +439,11 @@ export async function getAwuPoolCycleHistory(
  */
 /**
  * @cc [owner:avervaet,label:product] has-more-is-per-breakdown
- * `hasMoreCycleBreakdown` and `hasMoreExcessCycleBreakdown` MUST each be `true` only when
- * `cycleHistoryLimit` is below `MAX_CYCLE_HISTORY_LIMIT` and at least one more cycle with
- * consumption was found beyond the returned ones in that specific breakdown. Neither MUST be
- * `true` merely because older invoices exist, or because the *other* breakdown has more —
- * callers only ever render one of the two breakdowns for a given workspace.
+ * Each field of `hasMoreCycleHistory` MUST be `true` only when `cycleHistoryLimit` is below
+ * `MAX_CYCLE_HISTORY_LIMIT` and at least one more cycle with consumption was found beyond the
+ * returned ones in that specific breakdown. Neither field MUST be `true` merely because older
+ * invoices exist, or because the *other* breakdown has more — callers only ever render one of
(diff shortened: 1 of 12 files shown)


front/lib/api/credits/awu_pool_summary.ts
```diff
@@ -463,32 +466,35 @@ async function getAwuPoolCycleHistoryUncached(
     return new Ok({
       cycleBreakdown: [],
       excessCycleBreakdown: [],
-      hasMoreCycleHistory: false,
+      hasMoreCycleHistory: {
```
aubin-tchoi: don't you need to change the cache key?

sparkle/src/components/LoadMore.tsx
```diff
@@ -23,51 +32,55 @@ export function LoadMore({
...
+          onClick={onLoadMore}
+          disabled={isLoading || loadMoreIsHidden}
+        >
+          {isLoading ? (
+            <span>
+              {loadingLabel}
+              {LOADING_DOT_DELAYS.map((delay) => (
+                <span
+                  key={delay}
+                  className="animate-loading-dot opacity-100 motion-reduce:animate-none"
```
aubin-tchoi: fancy!

sparkle/src/components/LoadMore.tsx
```diff
@@ -23,51 +32,55 @@ export function LoadMore({
...
+                  style={{ animationDelay: delay }}
```
aubin-tchoi: nit: could be passed as a tailwind classname

aubin-tchoi (approved):
LGTM

PR #31914, opened 2026-09-04
Spend caps cleanup: remove dead Metronome credit-state machinery

tdraier:
## Description

Stacked on #31893. Follow-up cleanup now that spend caps are enforced only from the Redis rate limiter. This is the **code** PR; the DB migrations (drop `keys.creditState` + `workspaces.programmaticCreditState`, backfill `memberships.creditState`) land in #31916.

Removes the dead Metronome credit-state machinery:
- **api-key `creditState`** — FSM, Redis cache, dispatcher, reconcile, webhook branch, poke chip, wire type; the reconcile Temporal activity is gutted to a no-op (shell kept to drain in-flight executions), trigger removed.
- **programmatic `creditState`** — FSM/dispatcher/reconcile/Redis-cache/webhook/poke removal (keeps `expectedProgrammaticCreditStateFromUsage` + `WorkspaceProgrammaticCreditState` for the rate-limiter concurrency band).
- **`memberships.creditState`** — narrowed to `on_pool` / `user_seat` (the personal-seat pool carve-out); per-user-cap FSM transitions/dispatchers removed.
- **Unused Metronome spend alerts** — stop creating/updating per-user cap+warning, per-API-key cap, and programmatic cap alerts; add a script to delete the orphaned ones. Pool / seat-balance / balance-threshold / usage-cap / free-seat-credit-balance / default alerts stay.

The two model `declare` columns + physical columns are **kept** here (schema-drift-safe); #31916 drops them.

## Tests

- `tsgo --noEmit` clean in `front` and `front-api`; biome clean.
- Updated suites pass: `process_webhook` (26), `reconcile_credit_state` (3), `credit_state_dispatcher` (2), `user_block` (19), `user_credit_state_machine` (15), `keys/spend_limit` (2), `groups/spend_limit` (3), `default_user_spend_limit` (8). The FSM/alert suites for removed code were deleted alongside it.

## Risk

Medium — large mechanical removal of code that no longer drives enforcement (the rate limiter does). No enforcement behavior change.

## Deploy Plan

Deploy this (code) PR first; then #31916 (post-deploy `DROP COLUMN` + membership backfill) after it bakes.

PR #32253, opened 2026-09-10
feat(triggers): notify admins when a trigger is blocked by the programmatic cap

smb2268:
## Description

Follow up of `#32179`. Scheduled `workspace_pool` triggers now stop cleanly when the workspace programmatic monthly cap blocks them, but nobody is told. This emails workspace admins once per cap state through the existing `programmatic-cap-reached` Novu workflow, which had no live producer since `#31914`.

The payload's `isBlocked` flag becomes a `reason`. A cap of 0 gets new copy ("Your programmatic triggers are paused", addressed to the admin reading it, generic across schedule and webhook triggers, no mention of consumed credits); a positive cap that was used up keeps the existing exhausted copy, and the 80% warning keeps its own. A new `notifyAdminsTriggerBlockedByProgrammaticCap` in `lib/api/credits/` runs from the trigger activity gate and the webhook ingest gate via `fireAndForgetNotification`, so a failure is logged and never fails the run. It only acts for enabled `workspace_pool` triggers on credit-priced workspaces.

Deduplication keys on workspace, reason, the cap value and, for a positive cap, the billing cycle. That key is the Novu `transactionId` and a Redis sent marker written only after a successful send, so a failed send retries on the next run, a cap change re-arms the email, and unrelated usage-configuration edits do not. A 60 second per-workspace claim throttles the check itself, so a burst of blocked webhook requests costs one Redis call per event. Logs and the `trigger_blocked` PostHog event carry `block_reason=programmatic_monthly_cap`; the webhook request status stays `credits_exhausted`.

## Tests

28 tests across 4 files: copy per reason, `transactionId` stability, one email per cap state with retries and repeated runs staying silent, re-arm on cap change but not on unrelated configuration edits, the webhook ingest path notifying for a blocked `workspace_pool` trigger, and no email for disabled, `user_pool`, non-credit-priced or pool-depleted cases. A blocked schedule run still creates no conversation and completes without retry.

## Risk

Low. Worst case an admin gets one reminder per quarter while a workspace stays at a 0 cap. Safe to roll back.

## Deploy Plan

Deploy `front`. The Novu bridge picks up the new payload schema on its next sync.

### DIFF
--- a/front/lib/api/credits/access_control.ts
+++ b/front/lib/api/credits/access_control.ts
@@ -31,6 +31,11 @@ import type { ModelId } from "@app/types/shared/model_id";
 export const PROGRAMMATIC_CAP_REACHED_MESSAGE =
   "Your workspace has reached its programmatic monthly spending cap. An admin can raise the cap in the workspace's usage settings.";
 
+// Structured reason attached to logs and analytics when a run is blocked by the
+// programmatic monthly cap, as opposed to pool depletion or a generic rate limit.
+export const PROGRAMMATIC_MONTHLY_CAP_BLOCK_REASON =
+  "programmatic_monthly_cap" as const;
+
 /**
  * Whether the workspace is on a credit-priced (Metronome) plan, i.e. whether
  * the pool and programmatic-cap readers below apply to it.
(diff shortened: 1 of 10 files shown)


tdraier (approved):
LGTM

PR #32254, opened 2026-09-10
[front] enh: match new usage page skeletons to loaded layouts

aubin-tchoi:
## Description

This PR replaces the spinners on the new Usage page with skeletons matching the target layout (credit cards (not an actual credit card, a Card component for the credits), cycle history, members, requests, groups, and top-ups tables). 

Table skeletons reuse the actual column definitions to pre-shoot the display of the headers and keep the same widths

There's only the Top up button that makes the layout shift downwards, but this button will move around anyway so not changing for now.

## Tests

- Tested locally + preview.

## Risk

- Low.

## Deploy Plan

- Deploy front.


front/components/workspace/MembersUsageTable.tsx
```diff
@@ -126,6 +131,61 @@ type RowData = {
 
 type Info = CellContext<RowData, string>;
 
+function MemberUsageSkeletonCell({
```
aubin-tchoi, line 134: went super custom to have really the exact same layout. we could choose to simplify a bit

avervaet:
The visual effect is really nice 🔥

avervaet (approved):
Maybe UsageTableSkeleton and MemberNameSkeleton could be added to Sparkle ?

aubin-tchoi:
I want to keep them close to their "non-loading" counterparts since they will evolve together

PR #32256, opened 2026-09-10
Persist file explorer layout and sort across tab switches

PopDaph:
## Description

To fix: https://github.com/dust-tt/tasks/issues/8226

The file explorer kept its layout (grid/list) and sort mode in local component state, so leaving the Pod Files tab (or closing the conversation files panel) and coming back reset them to the defaults. Radix tabs unmount inactive content, so the state died on every switch.

Both are now stored through the existing `useScopedPodUiPreferences` localStorage hook, under a new `fileExplorer` scope. One preference for the whole app: the Pod Files tab and the conversation files panel share it.

## Tests

- Added two tests in `FileExplorer.test.tsx`: preferences survive an unmount/remount when a key is set, and stay in memory when none is set.
- Manually on a Pod: switched to Grid + Name Z → A, went to Conversations and back, then reloaded the page. Both settings held.

## Risk

Worst case, a stored blob fails to parse and the explorer falls back to the defaults. Front-end only, safe to rollback.

## Deploy Plan

- [ ] Deploy front

### DIFF
--- a/front/components/assistant/conversation/files_panel/ConversationFileExplorer.tsx
+++ b/front/components/assistant/conversation/files_panel/ConversationFileExplorer.tsx
@@ -282,6 +282,9 @@ export function ConversationFileExplorer({
           onOpenInteractive={onOpenInteractive}
           onOpenInPanel={onOpenInPanel}
           owner={owner}
+          preferencesResourceId={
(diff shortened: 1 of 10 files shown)


aubin-tchoi (approved):
LGTM

PR #32257, opened 2026-09-10
[Admin Governance] Fix agent permission mismatches from Poke MCP

philipperolet:
## Description

Follows https://github.com/dust-tt/dust/pull/32146.

Poke MCP reads conversations using an internal admin identity with no user attached. Its groups grant agent edit access, but the old permission check only recognizes system keys and Poke web sessions in this case. It therefore reports that agents cannot be edited while the new check says they can. A large batch of Poke reads produced over 27,000 mismatch logs on September 9.

Use the existing agent write-permission check for internal callers without a user too. Callers without editor grants still cannot edit; human permissions and regular API-key behavior stay unchanged. No new grants or backfill are needed.

## Risks

Blast radius: Agent read/edit permission reporting for internal callers without a user.
Risk: low

## Deploy Plan

- Deploy front-api and front workers.

## Tests

- [x] Internal admins with and without editor grants; existing agent configuration, ACL and shadow tests (50 tests).
- [x] Public API agent list and details tests, including unchanged hidden-agent read reporting for regular keys (19 tests).

### DIFF
--- a/front-api/routes/v1/w/[wId]/assistant/agent_configurations/[sId]/index.test.ts
+++ b/front-api/routes/v1/w/[wId]/assistant/agent_configurations/[sId]/index.test.ts
@@ -208,6 +208,7 @@ describe("GET /api/v1/w/[wId]/assistant/agent_configurations/[sId]", () => {
     const data = await response.json();
 
     expect(response.status).toBe(200);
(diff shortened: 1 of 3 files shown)


flvndvd (commented):
LGTM

PR #32258, opened 2026-09-10
[front] fix: stop the new usage page from double-fetching the credit pool

avervaet:
## Description

- The new usage page's credit cards (`CreditPoolCards`) fetch the pool current-cycle and cycle-history endpoints themselves. `UsagePage` was separately fetching the same current-cycle figures again through the `awu-pool-summary` endpoint, only to feed the legacy pool section and the top-up dialog.
- Each of those extra requests still resolves a full session/auth context against Postgres before the handler runs, so the redundant fetch added real DB load on every pageview of the new usage page (surfaced as connection-pool acquisition delay warnings).
- Read the current-cycle figures through the same `useAwuPoolCurrentCycle` hook `CreditPoolCards` already calls, instead of the separate `awu-pool-summary` call. SWR dedupes identical keys, so the two call sites now share a single request instead of issuing two.
- With `UsagePage` no longer mounting a live `useAwuPoolSummary`, its post-purchase refresh-polling machinery (`awuPostPurchaseRefreshState` / `resetAwuPostPurchaseRefreshCount` and the hook's `refreshInterval`) had no subscriber left to poll — removed it, along with the calls to it from `useAwuPurchase` and `useRedeemPoolTopupCoupon`.
- Fixed a loading-state regression introduced by the swap: the legacy pool section's outer guard hid the whole section while `awu-pool-current-cycle` was loading, making its own nested spinner unreachable. It now shows during loading too.
- Applied the same fix to the poke path: `PokeCreditPoolCard` (poke's workspace usage tab) fetched the same heavier `awu-pool-summary` endpoint just to read fields already returned by `awu-pool-current-cycle`. Swapped it to `usePokeAwuPoolCurrentCycle` and removed the now-unused `usePokeAwuPoolSummary` hook.

## Tests

- `npx tsgo --noEmit` and Biome pass on all changed files.
- Manual: not exercised end to end in a browser; behavior is unchanged, only the number of requests changes.

## Risk

Low-to-moderate. The data-source swap itself is equivalent (same backend computation, same defaulting), but two assumptions aren't independently verified:
- The removed 5×5s post-purchase polling on `awu-pool-summary` assumed the balance is already consistent by the time a purchase reports "succeeded." That polling existed for a reason (likely billing eventual-consistency); removing it trades a retry safety net for a single refetch on `awu-pool-current-cycle`, which has no equivalent retry.
- No manual/browser verification of the purchase flow (buy credits, confirm the pool numbers and locked/unlocked state update correctly) — only type-checking and lint were run.

Recommend a quick manual check on staging (purchase credits, watch the legacy pool section and top-up dialog update; check poke's workspace usage tab) before merging.

## Deploy Plan

Deploy front


front/hooks/useAwuPurchase.ts
```diff
@@ -75,7 +71,6 @@ export function useAwuPurchase({ workspaceId }: { workspaceId: string }) {
         await response.json();
 
         void mutateAwuPurchaseInfo();
-        resetAwuPostPurchaseRefreshCount(workspaceId);
         void mutateAwuPoolSummary();
```
tdraier, line 74: dead now ? can be removed ?

front/lib/swr/credits.ts
```diff
@@ -460,7 +430,6 @@ export function useRedeemPoolTopupCoupon({
           return { status: "error", message };
         }
 
-        resetAwuPostPurchaseRefreshCount(workspaceId);
         void mutateAwuPoolSummary();
```
tdraier, line 433: same here (dead ?)

front/components/pages/workspace/UsagePage.tsx
```diff
@@ -1275,12 +1269,13 @@ export function UsagePage() {
           {!isNewUsagePage &&
           isCreditPriced &&
           !showConsumptionAnalytics &&
-          !isAwuPoolSummaryLoading &&
-          (isAwuPoolSummaryError || hasPool) ? (
+          (isAwuPoolCurrentCycleLoading ||
+            isAwuPoolCurrentCycleError ||
+            hasPool) ? (
```
tdraier, line 1274: this shows the section while loading (at least the header) and flash/disappear if no pool after load

front/poke/swr/credits.ts
```diff
@@ -156,28 +155,6 @@ export function usePokeAwuUsageFromAnalytics({
...
 
-export function usePokeAwuPoolSummary({
-  owner,
-  disabled,
-}: PokeConditionalFetchProps) {
-  const { fetcher } = useFetcher();
-  const fetcherFn: Fetcher<AwuPoolSummaryResponseBody> = fetcher;
-
-  const { data, error, isValidating, mutate } = useSWRWithDefaults(
-    `/api/poke/workspaces/${owner.sId}/credits/awu-pool-summary`,
```
tdraier, line 167: can the route be removed ?

tdraier (approved):
minor nits, looks safe

front/hooks/useAwuPurchase.ts
avervaet: Yep, dead — removed in c723bf9ba2.

front/lib/swr/credits.ts
avervaet: Same, removed alongside the other one in c723bf9ba2.

front/components/pages/workspace/UsagePage.tsx
avervaet: Fixed in cceeb51f71: the section now shows the spinner/error while the current-cycle request is pending instead of hiding, and only renders once loaded when there's actually a pool, so no more flash.

front/poke/swr/credits.ts
avervaet: Confirmed nothing else called it after the PokeUsageTab switch, removed in 78ea8e084f.

PR #32261, opened 2026-09-10
refactor: extract useMarkdownFileEditor from FilePreviewDialog

adrsimon:
## Description

`FilePreviewDialog` owned markdown editing inline: draft content, view mode switch, dirty tracking, save and revert. The side panel needs the same behaviour, and copying it would leave two implementations to keep in sync.

This lifts it into `useMarkdownFileEditor` with no behaviour change. The dialog now reads everything through `markdown.*`.

Stack: 2 of 5, see #32260.

## Tests

Existing `file_explorer` suite passes unchanged (73 tests). The hook has no new behaviour of its own to cover yet; the panel that consumes it lands in PR 3.

## Risk

Low, pure extraction. `FilePreviewDialog` is still used by the Pod file explorer, so a regression would show up there.

## Deploy Plan

Deploy front.


Nils-Fedrigo (approved):
looks good, have you checked front/components/editor/MarkdownFileEditor.tsx ? Worth using the same implementation there too?

PR #32263, opened 2026-09-10
feat: preview files in the side panel

adrsimon:
## Description

Conversation file previews now open in the resizable side panel instead of the centered modal, so the conversation stays readable while you look at a file. `FilePreviewContext` becomes the single place that routes a citation click: side panel when previewable, download otherwise. Path-less content fragments are addressed by file id (`id:` prefix in the hash param) since they have no sandbox path. This drops the `opensInSidePanel` format flag, which existed only to send presentations down the panel path.

`useFilePreviewContext` now throws when no provider is mounted, so a surface that renders citations without one fails loudly instead of silently degrading to downloads. The agent builder right panel and the pod composer, which both render citations, get the provider.

## Tests

Updated the markdown and citation integration tests to assert side panel routing, plus a new `FilePreviewContext` test covering the missing-provider case. Full `components/assistant`, `components/markdown`, `components/file_explorer`, `components/agent_builder` and `components/pod` suites pass.

## Risk

Touches every surface that renders file citations: conversations, the agent builder preview and sidekick, the pod composer, and the analytics conversation panel. A surface missing the provider now throws rather than degrading, so the three mount sites above are the thing to check. `FilePreviewDialog` stays for the pod file explorer, haven't touched this surface yet.

## Deploy Plan

Deploy front.


front/components/assistant/conversation/FilePreviewContext.tsx
```diff
@@ -44,91 +34,73 @@ export function FilePreviewProvider({
...
     </FilePreviewContext.Provider>
   );
 }
 
 export function useFilePreviewContext() {
-  return useContext(FilePreviewContext);
+  const context = useContext(FilePreviewContext);
+  if (!context) {
+    throw new Error(
+      "useFilePreviewContext must be used within a FilePreviewProvider"
```
Nils-Fedrigo, line 106: ooc did you try in the extension? Like do we always have a FilePreviewProvider there?

front/components/assistant/conversation/files_panel/FilePreviewPanel.tsx
```diff
@@ -51,40 +86,55 @@ export function FilePreviewPanel({
...
+    category: preview.category,
+    entryPath: filePath,
+    fileUrl: urls?.baseUrl ?? null,
+    isActive: !!entry,
+    isContentLoading: preview.isContentLoading,
+    owner,
+    processedContent: preview.processedContent,
+  });
+
+  if (!entry || !urls) {
```
Nils-Fedrigo, line 136: if there is an error we should display the ConversationSidePanelHeader too no?

front/components/assistant/conversation/files_panel/FilePreviewPanel.tsx
```diff
@@ -37,8 +53,27 @@ export function FilePreviewPanel({
...
+    ? (filePath.split("/").pop() ?? filePath)
+    : (fileMetadata?.fileName ?? "");
+  const urls = filePath
+    ? {
+        baseUrl: getFilePathViewUrl(owner, filePath),
+        downloadUrl: getFilePathDownloadUrl(owner, filePath),
+      }
+    : fileId
+      ? {
+          baseUrl: getFileViewUrl(owner, fileId),
```
Nils-Fedrigo, line 83: I don't think getFileViewUrl works with Excels

Nils-Fedrigo: actually I confirm it does not

Nils-Fedrigo: Image

front/components/assistant/conversation/FilePreviewContext.test.tsx
```diff
@@ -0,0 +1,86 @@
...
+          data: undefined,
+        }}
+      >
+        {children}
+      </ConversationSidePanelContext.Provider>
+    ),
+  };
+}
+
+describe("useFilePreviewContext", () => {
```
Nils-Fedrigo, line 57: worth adding a test for excels lol

Nils-Fedrigo (approved):
Huge work, thanks! A few comments

PR #32264, opened 2026-09-10
[front] feat: back button in the side panel

adrsimon:
## Description

Going from the files list into a preview, or from one preview to another, left no way back. The panel now keeps an in-memory stack of what it showed and renders a back arrow in the shared header. Browser history can't serve here: panel state lives in hash params written with `replaceState`, so no entries are pushed. The stack is capped at 50 entries and cleared when the panel closes.

## Tests

Provider tests cover the stack, walking back through several panels, no entry on reselect, and clearing on close.

## Risk

Low. The arrow only shows when there is something to go back to.

## Deploy Plan

Deploy front.


adrsimon:
Superseded by #31999, which landed panel history on main: closing the panel now goes back to the previous one, so a separate back arrow is redundant. Dropped from the stack.

PR #32267, opened 2026-09-10
[front] Fix the editor link button doing nothing when clicked

rfrenoy:
## Description

Fixes https://github.com/dust-tt/tasks/issues/10354

The link dialog was rendered inside EditorSelectionToolbar, which returns null unless the editor is focused with a non-empty selection. Opening the dialog moves focus into it, so ProseMirror fired blur, the toolbar hid, and the dialog unmounted before it was painted. The keyboard shortcut faces the same issue: its listener lived in ToolBarContent, so it only existed while the toolbar was visible.

Fix: Split sparkle's ToolbarLink into the button and a separate ToolbarLinkDialog, and render the dialog as a sibling of the selection toolbar instead of a descendant, so its lifetime no longer depends on the editor keeping focus.
## Tests

test file + local env

## Risk

Low

## Deploy Plan

Front

### DIFF
--- a/front/components/editor/EditorFormattingToolbar.tsx
+++ b/front/components/editor/EditorFormattingToolbar.tsx
@@ -15,6 +15,8 @@ interface EditorFormattingToolbarProps {
 /**
  * The formatting toolbar shown above an editor's selection, with the link dialog it opens.
  *
+ */
+/**
  * @cc [owner:rfrenoy,label:react] link-dialog-sibling-of-selection-toolbar
  * The link dialog MUST stay a sibling of `EditorSelectionToolbar`, never a descendant: opening it
  * blurs the editor, which hides the selection toolbar and would unmount the dialog with it.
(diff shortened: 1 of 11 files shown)


front/components/editor/EditorFormattingToolbar.tsx
```diff
@@ -0,0 +1,40 @@
...
+  /** Extra actions appended to the toolbar, after the formatting ones. */
+  extraActions?: ReactNode;
+}
+
+/**
+ * The formatting toolbar shown above an editor's selection, with the link dialog it opens.
+ *
+ * @cc [owner:rfrenoy,label:react] link-dialog-sibling-of-selection-toolbar
+ * The link dialog MUST stay a sibling of `EditorSelectionToolbar`, never a descendant: opening it
+ * blurs the editor, which hides the selection toolbar and would unmount the dialog with it.
```
rfrenoy, line 22: I'm not sure I understand what makes for a good "cc" comment, these were automatically generated, apparently following some skills we have on the repo, so I kept them untouched

rfrenoy:
Closing in favour of https://github.com/dust-tt/dust/pull/32290

PR #26933, opened 2026-06-07
Remove all the Next APIs

davidebbo:
## Description

Now that `front-api` (Hono) fully owns the API surface and OpenAPI generation, this PR
  deletes the Next.js API and everything that only existed to support it.

  **Removed**
  - The entire `front/pages/api/` tree (757 files) — all routes are served by
    `front-api` (Hono).
  - Dead Next-handler infrastructure in `front/lib`:
    - `lib/api/auth_wrappers.ts`, `lib/api/sse.ts`, `lib/api/resource_wrappers.ts`
      (all exports had zero remaining references).
    - `logger/withlogging.ts` — dropped `withLogging` and its now-orphaned
      helpers/types; kept `apiError`, `getClientIp`, `getSequelizeErrorDetails`,
      `withGetServerSidePropsLogging` (still used by SSR pages / front-api).
    - `lib/api/.../resolve_authentication.ts` — removed
      `makeResolveAuthenticationHandler` + `ResolveAuthenticationResponse`; kept
      `resolveAuthentication` and `ResolveAuthenticationSchema` (front-api imports
    - `tests/utils/generic_public_api_tests.ts` — removed the two dead Next
      auth-test generators + `NextHandler`; kept `createPublicApiMockRequest`
      (used by ~28 front-api tests).
  - `next-swagger-doc` dependency from `front` (only `pages/api/doc.ts` used it;
    `front-api` keeps its own copy). `package-lock.json` re-synced.

  **Migration tooling cleanup (the Next→Hono migration is complete)**
  - `dangerfile.ts`: removed the `@migration-status` / `checkHonoMigrationSync`
    machinery and the `skip-migration-check` label; repointed the front-sse
    deploy check to a `front-api/routes/sse/` prefix match.
  - `front/CODING_RULES.md`: removed `BACK17` (Next↔Hono sync rule).
    `front-api/CODING_RULES.md`: reworded `[API1]` to drop the Next-mirror
    framing.
  - `run-dangerjs.yml`: dropped the `front/pages/api/` monitored path.

  **Comment cleanup**
  - Updated 13 `front/lib/api/*` shared-contract comments that described the code
    as "shared by the Next handlers and their Hono counterparts" → now just the
    front-api routes.
  - Removed dangling `pages/api` path references and one deleted-symbol reference
    from `front-api/routes/**` comments. Left behavioral-rationale comments
    (e.g. "matching the Next handler") intact.

## Tests

All passing

## Risk

That something somehow still depends on those, but not so likely.

[...]

PR #26888, opened 2026-06-05
Attribute sub-agent usage to the originating API key (attribution only)

tdraier:
https://github.com/dust-tt/tasks/issues/8571

## Description

> **Stacked on #26790** (base branch `use-parent-agent-name-key`). Review/merge that one first; this PR's diff is only the API-key attribution.

Fix API key attribution for sub-agent runs (`run_agent` / `agent_handover`) — e.g. the `dust-task` / `dust-planning` runs spawned by the **Go Deep** agent — which were all surfacing under the workspace system key (`DustSystemKey`) in usage analytics, with no link to the originating API key.

**Root cause.** Sub-agents re-authenticate to the Dust API with the workspace system key. Usage analytics resolves `api_key_name` from the persisted key id (`userContextApiKeyId`), which is always the system key for these flows. The pre-existing `x-dust-api-key-name` header (#19280) only overrode the *in-memory* key name, never the persisted id — so once analytics switched to id-based resolution (#20746) the header became a no-op and everything collapsed to `DustSystemKey`.

**Changes — attribution-only key propagation, never touching `auth.key()`:**
- `Authenticator` gains an attribution-only key reference (`withAttributionKey` / `attributionKeyId`), separate from `_key` and persisted through `toJSON`/`fromJSON`. It **never** affects authorization — role, caps, and system-key checks all keep reading `_key`.
- `withPublicAPIAuthentication` resolves the `x-dust-api-key-name` header to a key **within the same workspace** (`KeyResource.fetchByName`) and attaches it as the attribution key, instead of overriding the auth key.
- User messages persist `apiKeyId = attributionKeyId ?? key.id`, so id-based analytics resolve the real originating key.
- `getApiKeyNameHeader` forwards the attribution name, so attribution propagates **transitively** through nested sub-agents and across the Temporal boundary.
- Metronome events never surface the system key as `api_key_name` (keyless internal runs → `unknown`).

**Result:** API-rooted Go Deep runs attribute to the real customer key at every depth; web/session-rooted runs (no key) resolve to `unknown` instead of `DustSystemKey`. Only the key *name* travels over the wire (no ModelId), preserving SEC2.

## Tests

- `tsgo --noEmit` clean; `biome check` clean.
- `lib/api/assistant/conversation.test.ts` — 82/82 pass.
- Verified `auth.key()` (and therefore authorization) is untouched, and that `_attributionKey` round-trips through `toJSON`/`fromJSON`.

## Risk

[...]

PR #32268, opened 2026-09-10
Auth: correctly attribute child agent run to the parent key

fabiencelier:
## Description

closes https://github.com/dust-tt/tasks/issues/9537

run agent calls originating from API calls were not correctly assigned to the api key.
This fixes the issue by correctly attributing the key in the auth + using the attribution key everywhere

Tested locally: before the fix child run where wrongly billed on no key and the rate limiter was not hit.
After the fix, the billing is done on correct key + rate limiter is used.

Note that attribution key was already added in #26888 but removed in #26933

## Tests

- Tested locally: before the fix child run where wrongly billed on no key and the rate limiter was not hit. After the fix, the billing is done on correct key + rate limiter is used.
- Added some small unit tests

## Risk

medium, this impact authentication

## Deploy Plan

deploy front

### DIFF
--- a/front-api/routes/v1/w/[wId]/assistant/conversations/index.ts
+++ b/front-api/routes/v1/w/[wId]/assistant/conversations/index.ts
@@ -184,10 +184,9 @@ app.post(
               },
             });
           }
-          // Must resolve the same key as the authoritative gate in
-          // `postUserMessage`, or a capped originating key behind an uncapped
-          // system key would pass here and be rejected there — persisting the
-          // empty conversation this precheck exists to avoid.
+          // Same key as the authoritative gate in `postUserMessage`, or this
+          // precheck would let through a message that gate rejects — persisting
+          // the empty conversation it exists to avoid.
           const key = auth.keyForUsageAttribution();
           if (key && (await isApiKeyBlocked(auth, { keyModelId: key.id }))) {
             return apiError(ctx, {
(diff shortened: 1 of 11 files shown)


front-api/routes/v1/w/[wId]/assistant/conversations/index.ts
```diff
@@ -184,7 +184,11 @@ app.post(
               },
             });
           }
-          const key = auth.key();
+          // Must resolve the same key as the authoritative gate in
```
Comment (flvndvd) [inline front-api/routes/v1/w/[wId]/assistant/conversations/index.ts]: Can we DRY a bit those comments?

front/lib/auth.ts
```diff
@@ -1733,6 +1733,24 @@ export class Authenticator {
...
+  // run_dust_app), the request's own key otherwise. Distinct from
+  // `attributionKey()`, which exposes only the forwarded reference and is null
+  // on a direct call.
+  /**
+   * @cc [owner:fabiencelier,label:product;performance] one-usage-key-identity
+   * Every per-API-key spend-cap read, gate and increment, every persisted usage attribution, and
+   * every api-key field on an observability trace MUST resolve the key through this method.
+   * Authorization checks are the exception and MUST keep using `key()`.
+   */
+  keyForUsageAttribution(): { id: ModelId; name: string } | null {
```
flvndvd, line 1747: It feels weird that we have to do this, ideally we have just one api key and we stick with it, right? I guess the issue stems from us i. using API for internal actions (e.g. sub agent) and ii. using a system key instead of using the original api key we got.

tdraier:
Shouldn't `getMessageRateLimitActor` also use the keyForUsageAttribution ?

flvndvd (approved):

PR #32269, opened 2026-09-10
poke: stop double-fetching the credit pool summary

avervaet:
PokeCreditPoolCard fetched the heavier awu-pool-summary endpoint
(current-cycle + full cycle history) just to read fields already
returned by the lighter awu-pool-current-cycle endpoint the pool
cards elsewhere already use. Same pattern fixed for the customer-
facing usage page in #32258.


avervaet:
Moving this change into #32258 instead.

PR #32270, opened 2026-09-10
[front] Preserve Gemini function response boundaries

frankaloia:
## Summary

Fix Gemini request construction after `skill_management__enable_skill` without adding synthetic conversation content.

- Keep a `functionResponse` user Content separate from an adjacent ordinary user Content, such as the enabled skill instructions.
- Continue merging adjacent function responses so they match the preceding function-call count.
- Continue merging adjacent model parts so thought signatures stay attached to their logical turn.
- Remove the previous synthetic `\".\"` user-turn closer entirely.

No messages or parts are dropped; this only changes the boundary between two adjacent user Contents.

## Production evidence

Conversation `1ttRJZ6zHe` failed three times on `gemini-3.8-flash` with Google returning `Requests ending with a model turn are not supported.` In every version:

1. Gemini produced reasoning plus a signed `skill_management__enable_skill` call.
2. The MCP action succeeded and persisted one output item.
3. The next inference started after that output was committed and immediately returned the Google 400.
4. Rendering selected the full conversation (`prunedContext: false`), including the tool result and the following enabled-skill instruction message.

The Gemini converter then merged those adjacent user messages into one Content containing both `functionResponse` and plain text. The repeated failure shape and timing point to that lost function-response boundary, rather than a genuinely missing tool result or context pruning.

Direct A/B verification against the production agent-platform/Vertex route was not available. Google AI Studio accepted both hand-built shapes, so the provider error text alone is not treated as proof. This change instead preserves the semantic tool boundary represented in the source conversation and common to both production failures investigated.

## Test plan

- [x] Reproduction test for `enable_skill`: signed model tool call, function result, then skill-instruction user message remain three logical Contents.
- [x] Adjacent function responses still merge.
- [x] Adjacent plain user messages still merge.
- [x] Existing provider-passthrough coverage still passes.
- [x] Focused converter suite: 4/4 passed.
- [x] Biome check with warnings as errors on both changed files: clean.
- [x] Code-contract format check: clean.
- [x] PR diff whitespace check: clean.
- [ ] Full local typecheck is currently blocked by unrelated workspace dependency errors (`archiver` declaration in the active hive; stale Sparkle types in the isolated worktree). Neither reports an error in the changed files.

## Risk

Low. All converted parts remain present and in order; only incompatible user Contents are prevented from merging. Remaining uncertainty is end-to-end validation on the production Vertex-backed route.

## Deploy Plan

1. Deploy front.
2. Verify new `enable_skill` turns no longer produce the Google 400.

## Review

r? @davidebbo

Solving eng runner ticket: https://github.com/dust-tt/tasks/issues/10341

### DIFF
--- a/front/lib/model_constructors/sdk/google_genai/converters/input/utils.test.ts
+++ b/front/lib/model_constructors/sdk/google_genai/converters/input/utils.test.ts
@@ -43,9 +43,57 @@ describe("conversationToContents — provider_passthrough", () => {
     const contents = await conversationToContents(conversation, converters);
 
     // The two text turns merge into one model Content; the passthrough produces
-    // no part (no empty `{ text: "" }` slips in).
+    // no part (no empty `{ text: "" }` slips in). The trailing model turn is
+    // closed by the synthetic user turn.
     expect(contents).toEqual([
       { role: "model", parts: [{ text: "before" }, { text: "after" }] },
+      { role: "user", parts: [{ text: "." }] },
     ]);
+  });
+});
+
+describe("conversationToContents — trailing model turn", () => {
+  it("appends a user turn when the conversation ends on an assistant turn", async () => {
+    const conversation: BaseConversation = {
+      system: [],
+      messages: [
+        { role: "user", type: "text", content: { value: "hello" } },
+        { role: "assistant", type: "text", content: { value: "hi" } },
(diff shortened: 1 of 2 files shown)


davidebbo (approved):
LGTM

PR #32273, opened 2026-09-10
Self improving skill: entirely disable on mobile

fabiencelier:
## Description

closes https://github.com/dust-tt/tasks/issues/7742

Before this PR, the editor was in a read only mode without the ability to approve suggestions.
This disables the self improving skills suggestion entirely on mobile.
Long term, people will be able to accept suggestions without entering the builder anyway

r? @frankaloia 

## Tests

tested locally: I can edit skill even with existing suggestions on mobile

## Risk

low, purely UI + only for user on mobile, with self improving suggestion, in the builder so really not common.

## Deploy Plan

deploy front


frankaloia (approved):
LGTM

PR #32274, opened 2026-09-10
perf(user-metadata): cut per-request DB queries on the metadata route and notification reads

smb2268:
## Description

`GET/POST /api/user/metadata/:key` serves personal settings (sound preferences, extension menu, persisted navigation selection, notification preferences) and is called about four times on every app load. Each call resolved the user through `getUserFromSession`, which loads memberships and workspaces to build a workspace list the route only used to map an optional `workspaceId` query param, then fetched the same user again by id. That is four uncached queries per call to read one row. Resolve the caller through the Redis-cached `fetchUserFromSession` instead, and resolve `workspaceId` through the cached workspace and membership role lookups, so the metadata row is the only uncached query.

The notification path read the same table for the notify condition and email delay preferences with a filter on `userId` and `key` only. Both indexes on `user_metadata` are partial on `workspaceId`, so those queries could not use either one and scanned the table on every conversation-unread and new-project-conversation fan-out. Route them through the existing user-scoped `UserResource` helpers, which filter on `workspaceId IS NULL` and hit the index. This also removes the last direct `UserMetadataModel` imports from `lib/notifications`.

The GET handler now returns `{ key, value }`, matching the declared `UserMetadataType`, rather than the raw Sequelize row with its internal ids and timestamps. Clients only read `value`.

Code contracts on the route helper document that it must only use cached lookups and that workspace scoping requires membership. The index constraint on `user_metadata` reads is not encoded as a contract: the durable fix is an index, listed below.

Possible next steps, in separate PRs: confirm against `pg_stat_statements` which of the two query shapes dominated, since this analysis is from code rather than measured cost; return the always-needed keys inline from `/api/user` to remove the boot-time requests entirely; cache the memberships and workspaces lookup that `/api/user` still runs on every load; add a `userId` index on `user_metadata` so reads that cannot constrain `workspaceId`, and scrub deletes, stop scanning.

## Tests

- New functional tests for the route: user scope, workspace scope, and the non-member `workspaceId` fallback for GET, POST and DELETE. Workspace-scoped deletion itself is covered by the tests added in `32246`.
- One test per notification read (`filterMembersByNotifyCondition`, `filterParticipantsByNotifyCondition`, `getUserNotificationDelay`) asserting that a workspace-scoped row is ignored and the user-scoped preference is selected.
- Existing `project-new-conversation`, `conversation-unread` and `user_resource` suites pass unchanged.

## Risk

Low. A non-member `workspaceId` falls back to user-scoped metadata, as before. Safe to rollback.

## Deploy Plan

- Deploy front-api and front workers.


front-api/routes/user/metadata/[key]/index.ts
```diff
@@ -21,18 +24,35 @@ const ParamsSchema = z.object({
...
+ * metadata rather than failing.
+ */
 async function loadUserAndWorkspace(
-  ctx: Context
+  ctx: Context,
+  workspaceId: string | undefined
 ): Promise<
-  | { u: UserResource; workspaceModelId: number | undefined }
+  | { user: UserResource; workspaceModelId: number | undefined }
   | { err: Response & TypedResponse<APIErrorResponse> }
```
Comment (aubin-tchoi) [inline front-api/routes/user/metadata/[key]/index.ts:52]: outside of the scope of the PR but this should be a `Result`

aubin-tchoi (approved):

PR #32276, opened 2026-09-10
[sparkle] chore: pass down skeleton cells to `DataTableLoadingSkeleton`

aubin-tchoi:
## Description

Table skeletons should match the content of each cell. Having a generic component `DataTableLoadingSkeleton` actually encourages putting a generic placeholder but since the content of each cell is highly dynamic (we can put whatever component we want), it cannot reasonably match details such as avatars, badges, etc...

This PR replaces this component with a more modular one that receives a skeleton cell as prop.

Replaces its five usages with custom skeletons based on each table's actual cell contents.

Will add some lower level components (same level as `LoadingBlock`) for the loading blocks that map each main component (avatar, progress bar, etc..)

## Tests

- Tested locally + preview.

## Risk

- Low.

## Deploy Plan

- Deploy front.


front/components/members/InvitationsList.tsx
```diff
@@ -4,6 +4,8 @@ import {
   normalizeDisplayRole,
   ROLES_DATA,
 } from "@app/components/members/Roles";
+import type { UsageTableSkeletonCellProps } from "@app/components/workspace/UsageTableSkeleton";
```
avervaet: maybe we should rename this component if we start using it in different pages.

front/components/members/InvitationsList.tsx
```diff
@@ -4,6 +4,8 @@ import {
...
+import { UsageTableSkeleton } from "@app/components/workspace/UsageTableSkeleton";
```
avervaet: same

front/components/members/InvitationsList.tsx
aubin-tchoi: good point, forgot to do this
will also move it to sparkle near `DataTable`, this is actually kind of the new `DataTableLoadingSkeleton` except you have to pass a loading component for each cell

avervaet (approved):
LGTM but we definitely need to move and rename UsageTableSkeletonCellProps and UsageTableSkeleton as Sparkle component or somewhere else in the repo as they starts being used in multiple places.

front/components/members/InvitationsList.tsx
aubin-tchoi: actually I'll just rename it to `DataTableLoadingSkeleton` 😬

PR #32280, opened 2026-09-10
[front] Reset premium-model weekly message limit from poke

tdraier:
## Description

Adds a `user_premium_model_limit` reset target to the **Reset Message Rate Limits** poke plugin so support can clear a user's premium-model weekly message counter (25 messages / rolling 7 days, keyed `rate_limiter:workspace:<id>:user:<id>:premium_model_message_count`). Previously the plugin only reset the workspace message limit and a user's AWU fair-use limit — there was no reset path for the premium-model limit.

- `rate_limits.ts`: new `resetPremiumModelMessageRateLimitForUser({ auth, user })`, mirroring the AWU reset. Expires the premium-model key; no plan-limit guard since the cap is a fixed weekly constant (no-op when no key exists).
- `reset_message_rate_limit.ts`: new reset target + case; extracted the shared user-lookup / active-membership check into `resolveActiveWorkspaceUser`; generalized the Zod refinement to require `userEmail` for either per-user target.

Note: the `userEmail` field's `dependsOn` gate was removed because `dependsOn` arrays are AND-ed (no OR), so it can't show for either per-user target. The field now always renders; its description states it is required for the per-user resets.

## Tests

Type-checks clean (`npx tsgo --noEmit`). No behavior change to existing reset targets; new path reuses the established rate-limiter expiry helper.

## Risk

Low. Support-scoped poke plugin; only expires a Redis rate-limiter key.

## Deploy Plan

### DIFF
--- a/front/lib/api/assistant/rate_limits.ts
+++ b/front/lib/api/assistant/rate_limits.ts
@@ -381,6 +381,28 @@ export async function resetFairUseAwuCreditsRateLimitForUser({
   });
 }
 
+export async function resetPremiumModelMessageRateLimitForUser({
+  auth,
+  user,
+}: {
+  auth: Authenticator;
+  user: UserType;
+}) {
+  const workspace = auth.getNonNullableWorkspace();
+
+  const resetResult = await expireRateLimiterKey({
+    key: makePremiumModelMessageRateLimitKeyForUser(workspace, user),
+  });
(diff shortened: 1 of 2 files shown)


tdraier:

- #32280 ◀

fabiencelier (commented):
LGTM 🤙

PR #25480, opened 2026-05-11
[sessions] Sticky human mentions

ykmsd:
## Description
This PR is to add sticky human mentions. Note that this still doesn't work perfectly and we sometimes auto select last agent when you go back to an existing conversation. I will spend 30 mins more time tomorrow morning if I can fix it.  

## Deploy Plan

---

<a href="https://stagereview.app/dust-tt/dust/pull/25480">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://stagereview.app/assets/gh-open-in-stage-dark.svg">
    Open in Stage
  </picture>
</a>

PR #32281, opened 2026-09-10
[front] fix: stop sticky human mention re-inserting in the composer

frankaloia:
## Summary

Fixes [dust-tt/tasks#10353](https://github.com/dust-tt/tasks/issues/10353).

In a multi-user conversation, the user's handle kept popping back into the composer after every message/answer and couldn't be dismissed.

### Root cause

The "sticky human mentions" feature (#25480). The sticky mention is derived from **your own last sent message**:

```ts
const lastUserMessage = allMessages.filter(isUserMessage).findLast(
  (m) => ... && m.user?.id === context.user.id && ...   // your messages only
);
// autoMentions = lastUserMessage.richMentions, iff they are all human mentions
```

`autoMentions` becomes `stickyMentions`, and the draft-restore effect in `InputBarContainer.tsx` inserts them via `editorService.resetWithMentions(...)`. That effect re-runs whenever `stickyMentions` changes reference — and `stickyMentions` derives from `lastUserMessage`, which gets a **new object reference on every conversation store update** (another participant's message, an agent answer, each streaming tick). Its only guard was `editorService.isEmpty()`, so whenever the composer was empty it re-inserted the mention.

So `stickyMentions` = "the human mentions in the last message *I* sent," and whether it reappears is entirely a function of **what you last sent**:

- **You send with `@B` still in it** — the default, because the composer pre-fills `@B`, so unless you delete the chip whatever you type goes out as "@B …". Your new last message still mentions `@B` → `autoMentions` is still `[@B]` → after the composer clears it re-inserts `@B`. This is the vicious loop: the prefill guarantees your next message keeps the mention, which keeps it sticky, which prefills again.
- **You send without `@B`** (deleted the chip, sent a plain message) — your last message has no mention → `autoMentions` returns `[]` → nothing inserted. Cleared.

Two consequences:
- It reappears after every message/answer (your last message still carries the mention, and the churn re-inserts it on every store update while the box is empty).
- You can't dismiss it: deleting the chip empties the composer, and the next update (or the clear/re-render during your own send) re-inserts it before you can get a mention-less message out. Once it's back, the "editor not empty" guard stops the later-empty derivation from removing it — so the only escape is to send a message without the mention.

### Fix

Gate the insertion on an idempotency key `(conversation.sId, source message sId)` held in a ref — insert **at most once per source message**:
- **Deleting the chip now sticks** — same source id, already offered → not re-inserted.
- **Streaming / other participants' messages** don't re-insert — your last message's `sId` is unchanged.
- **Re-prefill still works** — sending a new message that keeps the mention changes the source id, so it re-arms (preserving the intended stickiness; the submit handler only clears the editor and relies on this effect to re-prefill).
- Sending a message with no mention clears it naturally (`autoMentions` returns `[]`).

Plumbs `stickyMentionsSourceId` (the current user's last message `sId`) through `AgentInputBar → InputBar → InputBarContainer`, documented with a `@cc [label:react]` contract on the prop.

### Pattern

The guard is the codebase's standard "run once per id" ref idiom — same shape as `useAutoOpenSidePanel.ts` (`autoOpenedFilesForRef === agentMessage.sId`), the composite-key `initKey` guards (`MCPServerAuthConnection.tsx`), and the sibling `useHandleMentions.tsx` (`prevConversationIdRef`) that handles agent sticky mentions.

## Testing

- `npx tsgo --noEmit` (front): no new errors (pre-existing unrelated `archiver` decl only).
- `biome check` + `cc-check format`/`list` on the changed file: clean; the new contract is discoverable.

Behavior reasoned through the effect's re-run conditions; there's no automated harness for this tiptap+Virtuoso composer.

## Risk

Low, scoped to the sticky human-mention prefill. Agent-mention selection (`useHandleMentions`) is untouched. Agent builder passes no source id and only has agent sticky mentions, so it's unaffected.

## Deploy Plan
1. Deploy fron

## Reviewer
r? @ykmsd Eng runner ticket https://github.com/dust-tt/tasks/issues/10353

### DIFF
--- a/front/components/assistant/conversation/AgentInputBar.tsx
+++ b/front/components/assistant/conversation/AgentInputBar.tsx
@@ -712,6 +712,7 @@ export const AgentInputBar = ({ context }: AgentInputBarProps) => {
                 user={context.user}
                 onSubmit={context.handleSubmit}
                 stickyMentions={autoMentions}
+                stickyMentionsSourceId={lastUserMessage?.sId ?? null}
                 lastRequestedModel={lastRequestedModel}
                 conversation={context.conversation}
                 draftKey={context.draftKey}
--- a/front/components/assistant/conversation/input_bar/InputBar.tsx
+++ b/front/components/assistant/conversation/input_bar/InputBar.tsx
@@ -90,6 +90,7 @@ interface InputBarProps {
   conversation?: ConversationWithoutContentType;
   space?: SpaceType;
   stickyMentions?: RichMention[];
+  stickyMentionsSourceId?: string | null;
   defaultAgentId?: string | null;
   isDefaultAgentLoading?: boolean;
   lastRequestedModel?: ModelSelectionType | null;
@@ -121,6 +122,7 @@ export const InputBar = React.memo(function InputBar({
   draftKey,
   space,
   stickyMentions,
+  stickyMentionsSourceId,
   defaultAgentId,
   isDefaultAgentLoading,
   lastRequestedModel = null,
@@ -767,6 +769,7 @@ export const InputBar = React.memo(function InputBar({
             pendingInputText={pendingInputText}
             onEnterKeyDown={handleSubmit}
             stickyMentions={stickyMentions}
+            stickyMentionsSourceId={stickyMentionsSourceId}
             defaultAgentId={defaultAgentId}
             isDefaultAgentLoading={isDefaultAgentLoading}
             lastRequestedModel={lastRequestedModel}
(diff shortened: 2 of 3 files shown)


Fraggle (approved):
LGTM

ykmsd (commented):
We just need to remove the sticky restore logic so it's much simpler to go this way no? https://github.com/dust-tt/dust/pull/32308

frankaloia:
> We just need to remove the sticky restore logic so it's much simpler to go this way no? #32308

@ykmsd I'm cool with removing if we want to go that route. Can close this one.

ykmsd:
ah! sorry I misunderstood the issue, I thought we were talking about human mentions always being sticky 😅 Seems like Seb is okay to go this path so let's remove it for now and see 🫡

PR #32282, opened 2026-09-10
Check Frame function calls at publish validation

davidebbo:
Fixes https://github.com/dust-tt/tasks/issues/10366

The Frames v2 equivalent of the publish-time check that `validate_frame_pod_functions.ts` used to do for Pod functions, but this was lost in the transition. This brings it back for Frames v2.

Tested by adding a bad `usePodFunction("list-todos-bad", {})` and calling validate, which got:

```json
{
  "timestamp": "2026-09-11T08:12:12.908360Z",
  "level": "ERROR",
  "fields": {
    "message": "dsbx command failed",
    "error": "POST https://localhost/api/v1/w/BHWrN3HaUa/sandbox/frames/validate: API error 400 (invalid_request): Frame UI calls a function that is not declared in its manifest:\nindex.tsx:21:32: usePodFunction('list-todos-bad')\nDeclared functions: 'list-todos', 'add-todo', 'toggle-todo'."
  },
  "target": "dsbx"
}
```

### DIFF
--- a/front/package.json
+++ b/front/package.json
@@ -166,7 +166,6 @@
     "jose": "^6.2.10",
     "jsdom": "^26.0.0",
     "jsforce": "^3.8.2",
-    "json-schema-to-typescript": "^15.0.4",
     "jsonwebtoken": "^9.0.0",
     "jwks-rsa": "^3.1.0",
     "libphonenumber-js": "^1.12.33",
--- a/front/lib/api/files/content_validation.ts
+++ b/front/lib/api/files/content_validation.ts
@@ -4,7 +4,7 @@ import { normalizeError } from "@app/types/shared/utils/error_utils";
 import * as ts from "typescript";
 
 export interface ValidationWarning {
-  type: "tailwind" | "typescript";
+  type: "frame_function" | "tailwind" | "typescript";
   message: string;
   oldString?: string;
   suggestion?: string;
--- a/front/lib/api/files/content_validation.ts
+++ b/front/lib/api/files/content_validation.ts
@@ -4,7 +4,7 @@ import { normalizeError } from "@app/types/shared/utils/error_utils";
 import * as ts from "typescript";
 
 export interface ValidationWarning {
-  type: "frame_function" | "tailwind" | "typescript";
+  type: "tailwind" | "typescript";
   message: string;
   oldString?: string;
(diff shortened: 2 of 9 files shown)


davidebbo:
Pushed `d44028972c`: an undeclared function reference now **blocks the publish**, not just validation.

The gap you hit: `dsbx frame publish` never ran validation at all. It goes through `buildAndPublishFramePublication`, which collects no warnings — Tailwind included, so that had the same hole. Its result type is `{ publicationId }`, with nowhere for a warning to go.

References are now checked in `buildFramePublication`, which the publish and validate paths share, and a literal naming nothing in the manifest is a `FramePublicationError("invalid_function_reference")`. The check is syntactic and runs before the UI build and before any sandbox work, so a rejected publish costs nothing. Computed names are still skipped, so a UI that builds names at run time stays publishable.

Why this one blocks while the input check does not: the reference check is exact. `resolveFrameFunctionReference` pins a v2 reference to `<frameId>/<name>`, so a name outside the manifest cannot resolve for any input or any viewer — there is no false positive to protect against. The input check compares against JSON Schema derived from Zod contracts whose runtime-only refinements TypeScript cannot express, so it stays a validate-only warning. The `frame-function-check-stays-advisory` contract is narrowed to the input check accordingly, and now says why the reference check is the exception; it was over-claiming by covering both under one "heuristic" rationale.

New test asserts the publish path itself refuses, and that nothing was bundled or stored and no sandbox was started.

front/package.json
```diff
@@ -166,7 +166,6 @@
     "jose": "^6.2.10",
     "jsdom": "^26.0.0",
     "jsforce": "^3.8.2",
-    "json-schema-to-typescript": "^15.0.4",
```
flvndvd, line 169: This guy is still needed to validate the contract of the pod function.

flvndvd (commented):
LGTM, OOC why aren't we using existing brick and redo-ing the whole code? Instantiating `ts` here as quite some cost, so ideally we do it only once like in the former publishing path.

flvndvd:
> LGTM, OOC why aren't we using existing brick and redo-ing the whole code? Instantiating `ts` here as quite some cost, so ideally we do it only once like in the former publishing path.

Seen IRL, will rework a bit.

PR #32284, opened 2026-09-10
Fix "Ask agent to fix" in Frames v2

davidebbo:
Fixes https://github.com/dust-tt/tasks/issues/10386

## Summary

It was broken because we were not setting`lastEditedByAgentConfigurationId` in the `useCaseMetadata` field for v2 frames. And this is what `useVisualizationRetry` looks for to trigger this UI.


flvndvd (commented):
LGTM

PR #32285, opened 2026-09-10
docs: restrict swagger to v1 endpoints, add private swagger

sfriquet:
## Description

Closes https://github.com/dust-tt/tasks/issues/9257

Move private endpoint out of `public/swagger.json` (used by mintlify to generate https://docs.dust.tt/api-reference) so that our public api reference only shows public endpoints.

Keep a private swagger in `routes/swagger-private.json` that contains all endpoints so docs on private endpoints can still be leveraged.


front-api/swagger.json
```diff
@@ -22,30 +22,32 @@
       }
     ],
     "tags": [
```
sfriquet, line 24: reordering + added Analytics

tdraier (approved):

PR #32286, opened 2026-09-10
[Agents] Default agent queries to alphabetical ordering

philipperolet:
## Description

Give active-agent queries a stable alphabetical order when no sort is requested. We noticed this missing default while comparing the old and new permission checks, so this PR adds it using the existing name index.

## Risks

Blast radius: Active-agent lists without an explicit sort in the API and agent tools.
Risk: low

## Deploy Plan

- Deploy front-api and front workers. No migration.

## Tests

- [x] Agent views, agent configuration, and Sidekick context tests (116 tests).
- [x] Default alphabetical ordering with and without a limit.

### DIFF
--- a/front/lib/api/assistant/configuration/views.ts
+++ b/front/lib/api/assistant/configuration/views.ts
@@ -25,7 +25,6 @@ import type { ModelId } from "@app/types/shared/model_id";
 import { assertNever } from "@app/types/shared/utils/assert_never";
 import { normalizeError } from "@app/types/shared/utils/error_utils";
 import type { WorkspaceType } from "@app/types/user";
-import type { Order } from "sequelize";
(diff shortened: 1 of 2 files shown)


front/lib/api/assistant/configuration/views.ts
```diff
@@ -187,9 +192,14 @@ async function fetchWorkspaceAgentConfigurationsWithoutActions(
       ? { attributes: { exclude: [...new Set(attributesToExclude)] } }
       : {};
 
+  // A unique order keeps legacy and grant queries on the same page despite different query plans.
```
philipperolet:
```suggestion
```

adrsimon (commented):
LTGM

